Issue #574, opened 2016-04-26
vmomi simulator property collector


dougm:
The simulator needs an implementation of the PropertyCollector to inspect the inventory (#573)

http://pubs.vmware.com/vsphere-60/index.jsp#com.vmware.wssdk.apiref.doc/vmodl.query.PropertyCollector.html

dougm:
Demo'd using my branch yesterday at the hackup: https://github.com/dougm/vic/tree/vcsim
Need to implement a bit more and cleanup the code, but should have a PR by tomorrow.

dougm:
Running into a issue with XML encode/decode for interface types returned within a PropertySet.  Will get it sorted out, just taking some unexpected time to debug.

dougm:
I can see the problem, order of the XML start elements are different between the real host/vc and govmomi.  For example against ESX, the attribute "type" is first, the xsi "type" is second:

```go
[]xml.Attr{
    {
        Name:  xml.Name{Space:"", Local:"type"},
        Value: "Folder",
    },
    {
        Name:  xml.Name{Space:"http://www.w3.org/2001/XMLSchema-instance", Local:"type"},
        Value: "ManagedObjectReference",
    },
}
```

Opposite order returned by govmomi:
```go
[]xml.Attr{
    {
        Name:  xml.Name{Space:"http://www.w3.org/2001/XMLSchema-instance", Local:"type"},
        Value: "ManagedObjectReference",
    },
    {
        Name:  xml.Name{Space:"", Local:"type"},
        Value: "Folder",
    },
}
```

On the client side in the case, the object is unmarshaled as a 'ManagedObjectReference', when it should be a 'Folder' type.

The xsi:type is only used for interface types, so I hadn't run into this issue until now.  Finding where to fix this properly will be a bit of a needle in a haystack of the xml package.  Will probably end up with a quick hack first.

dougm:
Using this hack for now to workaround: https://github.com/vmware/govmomi/compare/master...dougm:xsi-type

dougm:
Ran into a couple more XML encoding issues that held up the PR.  When DynamicProperty.Val is a string enum, for example:
https://github.com/vmware/govmomi/blob/master/vim25/types/enum.go#L1988

The xsi:type is 'string' rather than ManagedEntityStatus.  Working around this for now with:
https://github.com/dougm/govmomi/commit/3c0f97b936b27f4dc0f91fb9582ad4757d3442c6

[...]

PR #616, opened 2016-05-02
Support decoding of interface types

dougm:
XML decoder instance needs a handle to the type table, otherwise
interface types are decoded as nil.

Required for #574


caglar10ur:
LGTM :sheep: :it:

PR #617, opened 2016-05-02
create portlayer/exec and move handler code there

caglar10ur:


dougm:
LGTM

portlayer/exec/exec.go
```diff
@@ -0,0 +1,163 @@
...
+
+	return e, nil
+}
+
+// Create creates a container vm with the given params
+func (e *Executor) Create(ctx context.Context, config *ExecutorCreateConfig) error {
+	defer trace.End(trace.Begin("Create"))
+
+	// Convert the management hostname to IP
+	ips, err := net.LookupIP(managementHostName)
```
hickeng: torn about whether this should be done at initialization - any idea what the latency is on a hostname lookup from /etc/hosts?

PR #619, opened 2016-05-03
Fix go vet errors

dougm:
Don't know why this isn't happening in CI, via 'make check':

```
./cmd/imagec/docker.go:193: declaration of err shadows declaration at ./cmd/imagec/docker.go:154: 
./cmd/imagec/fetcher.go:154: declaration of err shadows declaration at ./cmd/imagec/fetcher.go:117: 
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ GOLINT ?= $(GOPATH)/bin/golint$(BIN_ARCH)
 GVT ?= $(GOPATH)/bin/gvt$(BIN_ARCH)
 
 .PHONY: all tools clean test check \
-	goversion goimports gvt gopath \
+	goversion goimports gvt gopath govet \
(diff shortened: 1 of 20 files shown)


fdawg4l:
:eyes: :+1: :v: :meat_on_bone:

dougm:
@caglar10ur pointed out the govet target was broken.. included his fix and got rid of all the vet warnings.

emlin:
Looks like we cannot write like following
a, err := XXX, if a is not defined, but err is already there. Is that a best practice?

caglar10ur:
LGTM :sheep: :it: 👍 👯

dougm:
@emlin yes, shadow checker is a fairly new feature of go vet: https://golang.org/cmd/vet/#hdr-Shadowed_variables

PR #620, opened 2016-05-03
Change default value of generate-cert in vic-machine

### LINKED ISSUE #615: vic-machine: default to generate certificate
Also support non-tls deployment

emlin:
Fixes #615 
Also fix non-tls VCH deployment issues.

### DIFF
--- a/usage.md
+++ b/usage.md
@@ -16,12 +16,12 @@ Installation will be required for capabilities such as [self-provisioning](doc/d
 
 Replace the `<fields>` in the example with values specific to your environment - this will install VCH to the specified resource pool of ESXi or vCenter, and the container VMs will be created under that resource pool. Bug #539 is tracking the container VM creation issue.
 
-Here is one resource pool path sample: /ha-datacenter/host/localhost/Resources/test 
--generate-cert flag is to generate certificates and configure TLS
--force flag is to remove an existing datastore folder or VM with the same name:
+- Here is one resource pool path sample: `/ha-datacenter/host/localhost/Resources/test`
+- -generate-cert flag is to generate certificates and configure TLS. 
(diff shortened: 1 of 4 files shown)


caglar10ur:
Do we need generate-cert flag at all? We can just generate one by default if -key flag is missing

emlin:
That's what I confirmed yesterday. Matt said that we still need to provide that option.
I think it's reasonable for a test env, otherwise, user has to have certificate files to access VCH.

mdubya66:
To clarify I said we need to confirm with @karthik-narayan that we need the non-tls scenario.

emlin:
OK, got it. 
@karthik-narayan, any idea from you?
 
In general, I think we're trying to collect feedback from customer through vic-machine, I'd prefer to keep it, so user can try that. If we found that's not necessary later on, it's easy to remove.

usage.md
```diff
@@ -16,9 +16,9 @@ Installation will be required for capabilities such as [self-provisioning](doc/d
 
 Replace the `<fields>` in the example with values specific to your environment - this will install VCH to the specified resource pool of ESXi or vCenter, and the container VMs will be created under that resource pool. Bug #539 is tracking the container VM creation issue.
 
-Here is one resource pool path sample: /ha-datacenter/host/localhost/Resources/test 
--generate-cert flag is to generate certificates and configure TLS
--force flag is to remove an existing datastore folder or VM with the same name:
+- Here is one resource pool path sample: /ha-datacenter/host/localhost/Resources/test.
```
hickeng: suggest you add \`'s around the path rather than put a . at the end - it makes it easier to differentiate between value and simple punctuation.

hickeng:
One comment on docs, otherwise LGTM.

PR #622, opened 2016-05-03
Update list image params to use the new format

jzt:
This brings the docker personality image backend up to date with recent go-swagger changes regarding how params are created.


sflxn:
lgtm

PR #626, opened 2016-05-03
Pass correct ResourcePool to VCH for container VMs

### LINKED ISSUE #539: Create container VM in resource pool
Now the installer will create resource pool for VCH, but the container VM is still created out of that resource pool

gigawhitlocks:
Fixes #539

### DIFF
--- a/install/management/appliance.go
+++ b/install/management/appliance.go
@@ -170,8 +170,9 @@ func (d *Dispatcher) createApplianceSpec(conf *configuration.Configuration) (*ty
 			&types.OptionValue{Key: "guestinfo.vch/components", Value: "/sbin/docker-engine-server /sbin/port-layer-server /sbin/vicadmin"},
 			&types.OptionValue{Key: "guestinfo.vch/sbin/imagec", Value: "-debug -logfile=/var/log/vic/imagec.log -insecure"},
 			&types.OptionValue{Key: "guestinfo.vch/sbin/port-layer-server",
-				Value: fmt.Sprintf("--host=localhost --port=8080 --insecure --sdk=%s --datacenter=%s --cluster=%s --pool=%s --datastore=%s --network=%s --vch=%s",
-					conf.TargetPath, conf.DatacenterName, conf.ClusterPath, conf.ResourcePoolPath, conf.ImageStorePath, conf.ExtenalNetworkPath, conf.DisplayName)},
+				Value: fmt.Sprintf("--host=localhost --port=8080 --insecure --sdk=%s --datacenter=%s --cluster=%s --pool=%s/%s --datastore=%s --network=%s --vch=%s",
+					conf.TargetPath, conf.DatacenterName, conf.ClusterPath, conf.ResourcePoolPath, conf.DisplayName,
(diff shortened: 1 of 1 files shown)


install/management/appliance.go
```diff
@@ -171,7 +171,8 @@ func (d *Dispatcher) createApplianceSpec(conf *configuration.Configuration) (*ty
 			&types.OptionValue{Key: "guestinfo.vch/sbin/imagec", Value: "-debug -logfile=/var/log/vic/imagec.log -insecure"},
 			&types.OptionValue{Key: "guestinfo.vch/sbin/port-layer-server",
 				Value: fmt.Sprintf("--host=localhost --port=8080 --insecure --sdk=%s --datacenter=%s --cluster=%s --pool=%s --datastore=%s --network=%s --vch=%s",
-					conf.TargetPath, conf.DatacenterName, conf.ClusterPath, conf.ResourcePoolPath, conf.ImageStorePath, conf.ExtenalNetworkPath, conf.DisplayName)},
+					conf.TargetPath, conf.DatacenterName, conf.ClusterPath, conf.ResourcePoolPath+`/`+conf.DisplayName,
```
emlin: Nice job! 
For this change, I would suggest to leverage the fmt.Sprintf previously.
from "--pool=%s" to "--pool=%s/%s", and just put conf.ResourcePoolPath and conf.DisplayName separately.

gigawhitlocks: Oh yes, this is better. My Pythonic tendencies were showing. I'll make this change momentarily

emlin:
LGTM

install/management/appliance.go
```diff
@@ -170,8 +170,9 @@ func (d *Dispatcher) createApplianceSpec(conf *configuration.Configuration) (*ty
 			&types.OptionValue{Key: "guestinfo.vch/components", Value: "/sbin/docker-engine-server /sbin/port-layer-server /sbin/vicadmin"},
 			&types.OptionValue{Key: "guestinfo.vch/sbin/imagec", Value: "-debug -logfile=/var/log/vic/imagec.log -insecure"},
 			&types.OptionValue{Key: "guestinfo.vch/sbin/port-layer-server",
-				Value: fmt.Sprintf("--host=localhost --port=8080 --insecure --sdk=%s --datacenter=%s --cluster=%s --pool=%s --datastore=%s --network=%s --vch=%s",
-					conf.TargetPath, conf.DatacenterName, conf.ClusterPath, conf.ResourcePoolPath, conf.ImageStorePath, conf.ExtenalNetworkPath, conf.DisplayName)},
+				Value: fmt.Sprintf("--host=localhost --port=8080 --insecure --sdk=%s --datacenter=%s --cluster=%s --pool=%s/%s --datastore=%s --network=%s --vch=%s",
+					conf.TargetPath, conf.DatacenterName, conf.ClusterPath, conf.ResourcePoolPath, conf.DisplayName,
```
hickeng: I don't recommend it here as it's potentially fragile, but be aware there is an explicit index mechanism for fmt.
From the docs:
`fmt.Sprintf("%[2]d %[1]d\n", 11, 22)```
will yield "22 11"

gigawhitlocks: Good to know :+1:

Issue #602, opened 2016-04-29
extraconfig requires special code to handle net.IPNet


caglar10ur:
Types requires some special handling as they may have unexported fields but in the IPNet case the struct is already flat

```
    30	type IP []byte
    31	
    32	// An IP mask is an IP address.
    33	type IPMask []byte
    34	
    35	// An IPNet represents an IP network.
    36	type IPNet struct {
    37		IP   IP     // network number
    38		Mask IPMask // network mask
    39	}
```
yet we still required https://github.com/vmware/vic/pull/600

hickeng:
Added it to the migration epic as that's where we'll likely hit these problems.

Issue #402, opened 2016-04-02
Serialize entire struct to/from guestinfo


hickeng:
Write an Encoder/Decoder for guestinfo storage of structures.
We wish to be able to serialize structs with:
maps
embedded types
the exec.Cmd struct

Serialization should be able to be sparse, with only values that are non-default being serialized.
Deserialization should be able to update an existing object

The intent here is that we be able to:
1. populate an exec.Cmd object with the command details
2. serialize the structure
3. deserialize the structure
4. launch the command
5. wait for it to complete
6. serialize the structure, including exit code
7. deserialize structure and retrieve exit code

Do this by adding annotations to the struct with the explicit extraConfig path at which to store the value - this will be replaced eventually by calculated paths derived from other annotations, however it's a closer match than making everything one giant blob.

hickeng:
Max extraConfig (and therefore guestinfo) value is 64k iirc. This should include a check to avoid value truncation.
64k is big given the datatypes we're dealing with, but for completeness later passes could add chunking. Shouldn't spend time on it this time around however.

hickeng:
@caglar10ur not a fan of the Squash and Merge button? ;) I think this can now be closed - concur? I've moved it into Done as it's definitely hit the review stage.

PR #627, opened 2016-05-03
Extraconfig update for untagged types

hickeng:
Adds support for serializing untagged types in extraconfig. This allows types that belong to non VIC libraries to be serialized without duplication.
Also adds direct guestinfo encode/decode mechanisms (untested) and error checking on that encode path.

Fixes: #602 #402


hickeng:
@caglar10ur take another look.

caglar10ur:
LGTM :sheep: :it:

PR #629, opened 2016-05-04
Resolve path to imagec on startup

fdawg4l:
Hardcoded `/sbin/imagec` was making local testing difficult.


caglar10ur:
Is there a reason not to assume imagec binary will be somewhere in the PATH? Couldn't we call it as "imagec"?

fdawg4l:
Good point!

Issue #352, opened 2016-03-31
Extracted dependencies for Go in Makefile


hickeng:
To keep the build short the Makefile tries to only rebuild what is needed. However the dependencies are currently updated by hand, rather than dumped from the Go imports.

Update the Makefile to use generated dependencies using the following or similar to list the packages
```
go list -f '{{join .Deps "\n"}}' ./tether |  xargs go list -f '{{if not .Standard}}{{.ImportPath}}{{end}}'
```

This should significantly reduce the odds of a failure to rebuild when source has changed.

hickeng:
Got sufficient irritated with tracking down a missing dependency that this is getting done now.

PR #633, opened 2016-05-04
Adding handle based spec config

### LINKED ISSUE #458: convert create to use construct/commit model
Create needs to be refactored to construct and return a config spec - this should be an opaque blob from the personality perspective - there's some work to decide how to represent this in swagger.

### LINKED ISSUE #459: convert portlayer network to use construct/commit model
Address the VM creation path first, not adding a network to an existing container.

This should be modifying a config spec, passed in as the container handle. This will entail:
* addition of vNIC(s) specification
* updating extraConfig with metadata.NetworkEndpoint data

The metadata.NetworkEnpoint sturcture is simply a placeholder from a while back for the data that was likely to be required. The separation between NetworkEndpoint and ContainerNetwork wasn't arbitrary but adds no value for now so they can be conflated if it simplifies implementation.

The reason for having the network config structs outside of the network component is because the following components need to be able to parse/modify them:
* vch - network component
* containerVM - tether
* vic-machine - diagnostics & introspection

### LINKED ISSUE #465: port-layer commit
This will live as part of the execution component for now as it's integral to the create and start workflows.

The commit call takes a config spec, possibly for a completely new VM, possibly for reconfiguring an existing VM, and actually performs the vSphere operation.

It's probably that most calls mapping to a reconfigure are going to be updating only the ExtraConfig array.

hmahmood:
Fixes #465 
Fixes #459 
Fixes #458 

Adding handle-based container create/commit/start model
Adding scope bind/unbind
Modifying scope container add to create ethernet cards when necessary
Using pci slot of ethernet card for identity in tether config


Makefile
```diff
@@ -247,7 +247,7 @@ $(portlayerapi-server): $(PORTLAYER_DEPS) $(SWAGGER)
 	@echo regenerating swagger models and operations for Portlayer API server...
 	@$(SWAGGER) generate server -A PortLayer -t $(realpath $(dir $<)) -f $<
 
-$(portlayerapi): $(portlayerapi-server) $(shell find pkg/ apiservers/engine/ -name '*.go') metadata/*.go
+$(portlayerapi): $(portlayerapi-server) $(shell find pkg/ apiservers/engine/ portlayer/ -name '*.go') metadata/*.go
```
hickeng: should be superceeded by #352

Makefile
```diff
@@ -201,7 +201,7 @@ docker-integration-tests:
 
 $(tether-linux): $(shell find cmd/tether -name '*.go') metadata/*.go
 	@echo building tether-linux
-	@CGO_ENABLED=1 GOOS=linux GOARCH=amd64 $(GO) build $(RACE) -tags netgo -installsuffix netgo --ldflags '-extldflags "-static"' -o ./$@ ./cmd/tether
+	@GOOS=linux GOARCH=amd64 $(GO) build $(RACE) -tags netgo -installsuffix netgo --ldflags '-extldflags "-static"' -o ./$@ ./cmd/tether
```
hickeng: I think CGO was explicitly enabled in be658a65a0675daa3bcbe0b2f7ca88bcf5a7d056 to ensure we had a static tether binary - does this still produce a static binary?

apiservers/engine/backends/container.go
```diff
@@ -173,22 +174,39 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
...
+		return types.ContainerCreateResponse{}, derr.NewErrorWithStatusCode(err, http.StatusInternalServerError)
+	}
+
+	id := createResults.Payload.ID
+	h := createResults.Payload.Handle
+
+	// configure networking
+	netConf := toModelsNetworkConfig(config)
+	if netConf != nil {
+		_, err := client.Scopes.AddContainer(
```
hickeng: For the handle model to work, we should be using the handle returned by AddContainer to pass into commit.

apiservers/engine/backends/container.go
```diff
@@ -231,18 +249,50 @@ func (c *Container) ContainerStart(name string, hostConfig *container.HostConfig
...
+		return derr.NewErrorWithStatusCode(err, http.StatusServiceUnavailable)
+	}
+
+	h = bindRes.Payload
+
+	// change the state of the container
 	// TODO: We need a resolved ID from the name
-	plStartParams := exec.NewContainerStartParams().WithID(name)
-	_, err := client.Exec.ContainerStart(plStartParams)
+	_, err = client.Containers.StateChange(containers.NewStateChangeParams().WithHandle(h).WithState("RUNNING"))
```
hickeng: If StateChange doesn't actually power on the VM directly then we should be using the handle that's returned from this call to pass into create.

apiservers/engine/backends/network.go
```diff
@@ -114,7 +114,7 @@ func (n *Network) CreateNetwork(name, driver string, ipam apinet.IPAM, options m
 		IPAM:      pools,
 	}
 
-	created, err := PortLayerClient().Scopes.Create(scopes.NewCreateParams().WithConfig(cfg))
+	created, err := PortLayerClient().Scopes.CreateScope(scopes.NewCreateScopeParams().WithConfig(cfg))
```
hickeng: Minor naming quirk - we've just altered the ContainerCreate call to be containers.Create - why are we going the other way with CreateScope?

apiservers/portlayer/restapi/configure_port_layer.go
```diff
@@ -90,19 +88,15 @@ func configureAPI(api *operations.PortLayerAPI) http.Handler {
...
-	netCtx, err := network.NewContext(
-		net.IPNet{
-			IP:   net.IPv4(172, 16, 0, 0),
-			Mask: net.CIDRMask(12, 32),
-		},
-		net.CIDRMask(16, 32),
-		sess)
 	if err != nil {
-		log.Fatalf("failed to create network context: %s", err)
+		log.Fatalf("ExecHandler ERROR: %s", err)
```
hickeng: I think this is unreachable as err is not set by any of the calls between here and the check after session.NewSession

apiservers/portlayer/restapi/configure_port_layer.go
```diff
@@ -90,19 +88,15 @@ func configureAPI(api *operations.PortLayerAPI) http.Handler {
...
 	}
 
+	handlerCtx := &handlers.HandlerContext{
+		Session: sess,
```
hickeng: FYI - session will go away when I get time to do so, or become global if that's not current feasible. No need to change anything, but so you know the plan.

Session is only really useful when you don't know the morefs of the networks/datastores/clusters/et al that you're working with and those will be globally accessible through the extraconfig object once the PL is migrated to that.

apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -0,0 +1,188 @@
...
+		Common: metadata.Common{
+			ID:   id,
+			Name: name,
+		},
+		Sessions: map[string]metadata.SessionConfig{
+			id: metadata.SessionConfig{
+				Common: metadata.Common{
+					ID: id,
+				},
+				Tty: false,
```
hickeng: tty should be coming from the params as well

apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -0,0 +1,188 @@
...
+				Attach: true,
+				Cmd: metadata.Cmd{
+					Env:  params.CreateConfig.Env,
+					Dir:  *params.CreateConfig.WorkingDir,
+					Path: *params.CreateConfig.Path,
+					Args: append([]string{*params.CreateConfig.Path}, params.CreateConfig.Args...),
+				},
+			},
+		},
+		Networks: make(map[string]*metadata.NetworkEndpoint),
```
hickeng: Might be neater to add this in the scopes.AddContainer impl - there's no guarantee we have any networks to put in it at this point

apiservers/engine/backends/container.go
```diff
@@ -33,7 +33,8 @@ import (
 	"github.com/docker/engine-api/types/container"
 	"github.com/docker/engine-api/types/strslice"
 
-	"github.com/vmware/vic/apiservers/portlayer/client/exec"
+	"github.com/vmware/vic/apiservers/portlayer/client/containers"
```
hickeng: Why change client/exec -> client/containers when also creating portlayer/exec/.. and the exec.NewContainer function?
Deliberate so that the packages/functions don't reflect each others naming on either side of the api boundary?

apiservers/portlayer/swagger.yml
```diff
@@ -281,27 +364,91 @@ paths:
...
+        '200':
+          description: "OK"
+          schema:
+            type: string
+        default:
+          description: "Error"
+          schema:
+            $ref: "#/definitions/Error"
+  /containers/{handle}:
+    post:
```
hickeng: should probably be PUT as this had better be idempotent for the same handle.

apiservers/portlayer/swagger.yml
```diff
@@ -281,27 +364,91 @@ paths:
...
+        '404':
+          description: "not found"
+        '200':
+          description: "OK"
+        default:
+          description: "Error"
+          schema:
+            $ref: "#/definitions/Error"
+  /containers/{handle}/state:
+    post:
```
hickeng: put

cmd/tether/tether_linux.go
```diff
@@ -117,7 +117,7 @@ func (t *osopsLinux) setup() error {
 		log.Error(detail)
 		return errors.New(detail)
 	}
-	log.SetOutput(io.MultiWriter(out, os.Stdout))
+	log.SetOutput(io.MultiWriter(os.Stdout, out))
```
hickeng: curious why this was necessary - it will change the order of writing between VM console and debug log - was that somehow significant?

portlayer/exec/container.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	return nil
+}
+
+func (c *Container) newHandle() *Handle {
+	return newHandle(c)
+}
+
+func (c *Container) CommitHandle(ctx context.Context, sess *session.Session, h *Handle) error {
```
hickeng: Why commit a handle instead of just passing the configspec in directly? This looks like we're blending the handle/cache code with the foundational PL functions.

hickeng:
Apologies - didn't manage to finish reviewing it tonight... I'm going to need to look at the code in place when I'm back in the office.

apiservers/engine/backends/network.go
hmahmood: I actually tried Scopes.Create, but swagger doesn't seem to like generating that (it would generate ScopesPost instead for some reason). I think there may be a conflict between Scopes.Create and Container.Create in the swagger code generation.

apiservers/portlayer/restapi/handlers/containers_handlers.go
hmahmood: Hm... I don't see that in the swagger file.

apiservers/portlayer/restapi/handlers/containers_handlers.go
hmahmood: Context.AddContainer does that actually; I will take this one away.

apiservers/portlayer/swagger.yml
hmahmood: Right

cmd/tether/tether_linux.go
hmahmood: Apparently io.MultiWriter quits on the first stream that has an error. In my testing on my linux box, writing to ttySx was failing, but I still wanted to see the log output on stdout.

apiservers/engine/backends/container.go
hmahmood: Trying to follow REST here. exec exposes a collection of objects from the PL, just like network. REST operates on resources/objects, so we have /containers and /scopes there instead of /exec/containers and /network/scopes.

apiservers/engine/backends/network.go
hickeng: can you raise an issue against go-swagger for that - and a comment in the code as to why the name, referencing that issue.

apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -0,0 +1,187 @@
...
+	id := exec.GenerateID().String()
+	// Autogenerate a name if client doesn't specify one
+	if params.Name == nil {
+		name = namesgenerator.GetRandomName(0)
+	} else {
+		name = *params.Name
+	}
+
+	// Init key for tether
+	privateKey, err := rsa.GenerateKey(rand.Reader, 512)
```
andrewtchin: Is this 512 for a reason? I realize this is internal and tether might not be permanent, but there's almost no point in using encryption at all if it's with a 512 bit RSA key

cmd/tether/utils_linux.go
```diff
@@ -92,87 +93,60 @@ func (t *osopsLinux) SetHostname(hostname string) error {
 	return nil
 }
 
-func slotToPCIAddr(pciSlot int32) (string, error) {
-	slotPath := path.Join("/sys/bus/pci/slots/", strconv.Itoa(int(pciSlot)), "/address")
-	f, err := os.Open(slotPath)
-	if err != nil {
-		return "", fmt.Errorf("unable to open slotPath %s: %s", slotPath, err)
+func slotToPCIPath(pciSlot int32) (string, error) {
```
andrewtchin: 👏

apiservers/portlayer/restapi/handlers/containers_handlers.go
hickeng: to some extent it depends on what you're trying to do. At the moment there's no validation of the key - it's present solely because the crypto/ssh code requires a host key to function at all.

We're using ssh primarily for the muxing, out-of-band control channels, and potential for scp rather than for the encryption. Getting the key management/distribution correct requires significantly more thread modelling and thought about the trust domains - for now we assume that the ESX infrastructure is mostly trustworthy.

apiservers/engine/backends/network.go
hmahmood: operationId is unique for a swagger spec, so using Create for both /scopes/ and /containers/ does not work.

apiservers/engine/backends/container.go
```diff
@@ -173,22 +174,41 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 	}
 
 	plCreateParams := c.dockerContainerCreateParamsToPortlayer(config, layer.ID, host)
-	createResults, err := client.Exec.ContainerCreate(plCreateParams)
-
+	createResults, err := client.Containers.Create(plCreateParams)
 	// transfer port layer swagger based response to Docker backend data structs and return to the REST front-end
 	if err != nil {
-		if _, isa := err.(*exec.ContainerCreateNotFound); isa {
+		if _, isa := err.(*containers.CreateNotFound); isa {
```
jzt: Can you rename `isa` to `ok`?

apiservers/engine/backends/container.go
```diff
@@ -173,22 +174,41 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
...
+	// commit the create op
+	_, err = client.Containers.Commit(containers.NewCommitParams().WithHandle(h))
+	if err != nil {
+		return types.ContainerCreateResponse{}, derr.NewErrorWithStatusCode(err, http.StatusInternalServerError)
 	}
 
 	// Success!
-	log.Printf("container.ContainerCreate succeeded.  Returning container id %s", *createResults.Payload.ContainerID)
-	return types.ContainerCreateResponse{ID: *createResults.Payload.ContainerID}, nil
+	log.Printf("container.ContainerCreate succeeded.  Returning container handle %s", *createResults.Payload)
```
jzt: Should this print the container ID, or are you wanting to print the entire Payload struct?

apiservers/engine/backends/container.go
```diff
@@ -231,18 +251,52 @@ func (c *Container) ContainerStart(name string, hostConfig *container.HostConfig
...
+	h = bindRes.Payload
+
+	// change the state of the container
 	// TODO: We need a resolved ID from the name
-	plStartParams := exec.NewContainerStartParams().WithID(name)
-	_, err := client.Exec.ContainerStart(plStartParams)
+	stateChangeRes, err := client.Containers.StateChange(containers.NewStateChangeParams().WithHandle(h).WithState("RUNNING"))
 	if err != nil {
-		if _, isa := err.(*exec.ContainerStartNotFound); isa {
+		if _, isa := err.(*containers.StateChangeNotFound); isa {
```
jzt: `s/isa/ok`

apiservers/engine/backends/container.go
hmahmood: The payload is just a string, the handle, so it should print out the handle.

apiservers/engine/backends/container.go
hmahmood: fixed

apiservers/engine/backends/container.go
hmahmood: fixed

pkg/vsphere/spec/spec.go
```diff
@@ -248,3 +251,42 @@ func (s *VirtualMachineConfigSpec) generateNextKey() int32 {
...
+		return i.PciSlotNumber
+	}
+
+	return NilSlot
+}
+
+// CollectSlotNumbers returns a collection of all the PCI slot numbers for devices in the spec
+func (s *VirtualMachineConfigSpec) CollectSlotNumbers() []int32 {
+	// collect all the already assigned slot numbers
+	var slots map[int32]bool
```
hickeng: This is going to panic - it's a nil map

hmahmood: I mainly had it in there for slot number duplicates. The code is getting exercised, and the reason it doesn't panic is because it never gets to line 280. 

It doesn't really make sense the way the map is used here, since it is not really detecting duplicates. I am going to take it and assume there are no duplicate slot numbers in the spec.

hmahmood:
Got an LGTM from George over email ... merging.

PR #637, opened 2016-05-05
Dynamic go dependencies in Makefile

### LINKED ISSUE #352: Extracted dependencies for Go in Makefile

hickeng:
This adds dynamically generated Go dependencies in the Makefile to avoid
the case where a dependency is missed from a target.
Adds seconds to initial Makefile parse as that's when the dependencies are
actually generated - could look into a deferred execution pass if the time
increment becomes a problem, but kept it simple initially.

Fixes #352

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ ifeq ($(USER),vagrant)
 endif
 
 BASE_DIR := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
+BASE_PKG := github.com/vmware/vic/
 
 BIN ?= bin
 IGNORE := $(shell mkdir -p $(BIN))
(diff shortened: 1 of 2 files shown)


caglar10ur:
LGTM :sheep: :it:

PR #644, opened 2016-05-06
Tether connection handling fixes and add attach server to PL

fdawg4l:
Retry failed connections faster than 1s per attempt.
Add test to mess with network alignment of bytes between serial attach
server and test client.
Add Purge function before handshake on the PL side to purge the line of
any (buffered) data.
Fix stacked connection close defers in session attach connector when
connection fails.

### DIFF
--- a/cmd/tether/main_linux.go
+++ b/cmd/tether/main_linux.go
@@ -34,8 +34,6 @@ func main() {
 		halt()
 	}()
 
-	log.SetLevel(log.DebugLevel)
-
 	// where to look for the various devices and files related to tether
 	pathPrefix = "/.tether"
 
--- a/cmd/tether/main_linux.go
+++ b/cmd/tether/main_linux.go
@@ -34,6 +34,8 @@ func main() {
 		halt()
 	}()
 
+	log.SetLevel(log.DebugLevel)
+
 	// where to look for the various devices and files related to tether
 	pathPrefix = "/.tether"
 
--- a/cmd/tether/tether_linux.go
+++ b/cmd/tether/tether_linux.go
@@ -176,7 +176,7 @@ func (t *osopsLinux) backchannel(ctx context.Context) (net.Conn, error) {
 	}
 
 	// HACK: currently RawConn dosn't implement timeout so throttle the spinning
-	ticker := time.NewTicker(1000 * time.Millisecond)
+	ticker := time.NewTicker(10 * time.Millisecond)
 	for {
 		select {
 		case <-ticker.C:
--- a/cmd/tether/tether_linux.go
+++ b/cmd/tether/tether_linux.go
@@ -189,6 +189,10 @@ func (t *osopsLinux) backchannel(ctx context.Context) (net.Conn, error) {
 	}
 
 	// HACK: currently RawConn dosn't implement timeout so throttle the spinning
+
+	// This needs to tick *faster* than the ticker in connection.go on the
+	// portlayer side.  The PL sends the first syn and if this isn't waiting,
+	// alignment will take a few rounds (or it may never happen).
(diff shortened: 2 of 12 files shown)


portlayer/attach/connector.go
```diff
@@ -167,21 +175,30 @@ func (c *Connector) processIncoming(conn net.Conn) {
...
+	var ids []string
+	ids, err = SSHls(client)
 	if err != nil {
 		log.Errorf("SSH connection could not be established: %s", errors.ErrorStack(err))
 		return
 	}
 
 	for _, id := range ids {
-		si, err := SSHAttach(client, id)
+		var si SessionInteraction
```
caglar10ur: not sure what is the expected length of ids but maybe you may want to consider moving the declaration out of the loop so that we create one SessionInteraction and reuse multiple times.

portlayer/attach/connector.go
```diff
@@ -137,22 +137,30 @@ func (c *Connector) Remove(id string) {
 
 // takes the base connection, determines the ID of the source and stashes it in the map
 func (c *Connector) processIncoming(conn net.Conn) {
+	var err error
+	defer func() {
+		if err != nil && conn != nil {
```
hmahmood: So if `err == nil`, we don't close `conn`? This seems to be a change from before where `conn.Close()` was called unconditionally.

fdawg4l: If err == nil and conn is not nil, we dont need to do anything.  if err is nil and conn is nil, again we don't need to do anything.

portlayer/attach/connector.go
fdawg4l: :+1:

fdawg4l:
Updated this change.  Included a pending fix upstream for the handshake issues we were seeing upstream.  I'll revert the specific commit for vendor when (if?) upstream accepts my PR.  See the issue for details.

cmd/tether/main_linux.go
```diff
@@ -34,6 +34,8 @@ func main() {
 		halt()
 	}()
 
+	log.SetLevel(log.DebugLevel)
+
```
caglar10ur: We do that via (4-5 lines below)

```
 40     if strings.HasSuffix(os.Args[0], "-debug") {
 41         extraconfig.DecodeLogLevel = log.DebugLevel
 42         extraconfig.EncodeLogLevel = log.DebugLevel
 43         log.SetLevel(log.DebugLevel)
 44     }
 45
```

cmd/tether/tether_linux.go
```diff
@@ -176,7 +189,7 @@ func (t *osopsLinux) backchannel(ctx context.Context) (net.Conn, error) {
 	}
 
 	// HACK: currently RawConn dosn't implement timeout so throttle the spinning
-	ticker := time.NewTicker(1000 * time.Millisecond)
+	ticker := time.NewTicker(10 * time.Millisecond)
```
caglar10ur: just wondering, why it was 1 sec and why it's not 10  ms?

cmd/tether/main_linux.go
fdawg4l: Yeah, we do that *now* after your commit.  Will fix asap.  Thanks for noticing.

cmd/tether/tether_linux.go
fdawg4l: 1s was a bug.  The tether should retry more frequently than the PL connection side.  @hickeng spotted this issue and couldnt tell me why it was set to 1s.

fdawg4l: Helpfully there's a comment there as to why.... LOL!  :trollface: 

I'll add one now.

caglar10ur:
LGTM :sheep: :it: :)

PR #653, opened 2016-05-07
Block start until tether confirms

### LINKED ISSUE #426: start
Start a containerVM and launch the main container process

Should start the main container process - either specified on command line or from image metadata (depends on #423)
Should publish the success/failure launching the container process
Should publish the return value of exited container process after successful launch
Docker start command should block on the cli until success/failure of launch is confirmed

caglar10ur:
Fixes #426

### DIFF
--- a/cmd/tether/tether.go
+++ b/cmd/tether/tether.go
@@ -195,7 +195,7 @@ func handleSessionExit(session *SessionConfig) error {
 	// record exit status
 	// FIXME: we cannot have this embedded knowledge of the extraconfig encoding pattern, but not
 	// currently sure how to expose it neatly via a utility function
-	extraconfig.EncodeWithPrefix(dataSink, session.ExitStatus, fmt.Sprintf("sessions|%s/status", session.ID))
+	extraconfig.EncodeWithPrefix(dataSink, session.ExitStatus, fmt.Sprintf(".sessions|%s.status", session.ID))
 	log.Infof("%s exit code: %d", session.ID, session.ExitStatus)
 
 	// check for executor behaviour
@@ -250,11 +250,21 @@ func launch(session *SessionConfig) error {
 		if err != nil {
 			detail := fmt.Sprintf("failed to start container process: %s", err)
 			log.Error(detail)
+
+			// Set the Started key to the detailed error message
+			session.Started = detail
+			// save the modified value back to the vmx file
+			extraconfig.EncodeWithPrefix(dataSink, session.Started, fmt.Sprintf(".sessions|%s.started", session.ID))
+
 			return errors.New(detail)
 		}
 
 		// ChildReaper will use this channel to inform us the wait status of the child.
 		config.pids[session.Cmd.Process.Pid] = session
+		// Set the Started key to "true"
+		session.Started = "true"
+		// save the modified value back to the vmx file
+		extraconfig.EncodeWithPrefix(dataSink, session.Started, fmt.Sprintf(".sessions|%s.started", session.ID))
 
 		log.Debugf("Launched command with pid %d", session.Cmd.Process.Pid)
 
--- a/pkg/vsphere/vm/vm.go
+++ b/pkg/vsphere/vm/vm.go
@@ -116,3 +116,18 @@ func (vm *VirtualMachine) FetchExtraConfig(ctx context.Context) (map[string]stri
 	}
 	return info, nil
 }
+
+// WaitForExtraConfig waits until key shows up with the expected value inside the ExtraConfig
+func (vm *VirtualMachine) WaitForExtraConfig(ctx context.Context, waitFunc func(pc []types.PropertyChange) bool) error {
+	// Get the default collector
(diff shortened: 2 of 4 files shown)


cmd/tether/tether.go
```diff
@@ -156,12 +156,16 @@ func run(src extraconfig.DataSource, sink extraconfig.DataSink) error {
 
 			// check if session has never been started
 			if proc == nil {
+				// save the modified config back to the vmx file
+				defer extraconfig.Encode(sink, config)
```
hickeng: I'm not sure this will function as desired. I think this is serializing the Started key only when the primary process exits, thereby causing the tether to exit.

Look at line 204 for exit status seralization in the handleSessionExit function - we should be able to do something similar in the launch function for Started.

caglar10ur: Ah, correct. Will change it to that...

caglar10ur:
Pushed a new revision, this also fixes the code @hickeng was referring as it wasn't working :)

caglar10ur:
Hmm just realized that EncodeWithPrefix never returns and causes tether to stuck just after setting the started value.

caglar10ur:
OK, ignore the last comment and review please

pkg/vsphere/vm/vm.go
```diff
@@ -116,3 +116,34 @@ func (vm *VirtualMachine) FetchExtraConfig(ctx context.Context) (map[string]stri
 	}
 	return info, nil
 }
+
+// WaitForExtraConfig waits until key shows up with the expected value inside the ExtraConfig
+func (vm *VirtualMachine) WaitForExtraConfig(ctx context.Context, key string, expected string) error {
```
hickeng: This should take a function to do the validation rather than an expected string.
This currently hangs indefinitely in the error case - the Start call needs to block until this is non-nil, then return success or error based on the contents.

cmd/tether/tether.go
```diff
@@ -157,11 +157,17 @@ func run(src extraconfig.DataSource, sink extraconfig.DataSink) error {
 			// check if session has never been started
 			if proc == nil {
 				log.Infof("Launching process for session %s\n", session.ID)
+
+				// launch the session
 				err := launch(session)
+				// save the modified value back to the vmx file
+				extraconfig.EncodeWithPrefix(dataSink, session.Started, fmt.Sprintf(".sessions|%s.started", session.ID))
```
hickeng: would prefer to have this in the launch function and keep this main loop as free from clutter as possible; it's already pretty long.

caglar10ur: Will move there

pkg/vsphere/vm/vm.go
caglar10ur: That's why we are passing context and it - to be able to define deadlines/timeouts or cancelation on top of it. We can key/expected string to a func that does the check so that we can do more enriched checks but not sure how it will help your original concern

caglar10ur:
Rebased & pushed a new revision (but haven't tested)

dougm:
lgtm

portlayer/exec/container.go
```diff
@@ -135,5 +138,29 @@ func (c *Container) Start(ctx context.Context) error {
...
+
+		for _, c := range pc {
+			if c.Op != types.PropertyChangeOpAssign {
+				continue
+			}
+
+			values := c.Val.(types.ArrayOfOptionValue).OptionValue
+			for _, value := range values {
+				// check the status of the key and return true if it's "true"
+				if key == value.GetOptionValue().Key && value.GetOptionValue().Value.(string) == "true" {
```
hickeng: We get potentially valuable error info in the Started field when it's not set to true. It also indicates that the command is never going to start as we've tried, failed, and there's no retry mechanism.

I like the context timeout for systemic failure issues such as vSphere communication breaking down, but the presence of this key with a non-zero value is sufficient to break from this wait, and doubles as error detail.

portlayer/exec/container.go
```diff
@@ -135,5 +138,29 @@ func (c *Container) Start(ctx context.Context) error {
...
+			}
+		}
+		return false
+	}
+
+	// Wait some before giving up...
+	ctx, cancel := context.WithTimeout(ctx, propertyCollectorTimeout)
+	defer cancel()
+
+	return c.vm.WaitForExtraConfig(ctx, waitFunc)
```
hickeng: I don't know how error return from Start gets handled at this point, but if we fail to launch we should be returning an error, so return value is based around `value == "true"`

hickeng:
@caglar10ur there's a PR against your branch that adds error handling - tested it with /bin/date2 and ensured that we get a "fork/exec: no such file or directory" error (or whatever the text is).

hickeng:
LGTM

fdawg4l:
Lgtm

hickeng:
Couldn't figure out how to merge this without Caglar available so it's gone in as #704

PR #654, opened 2016-05-09
Set timezone to US/Pacific-New

caglar10ur:


mdubya66:
That's not very i18n? Should we not use GMT?

caglar10ur:
I believe the default is UTC already. I guess we can keep it that way (and I can change on my local vm). Dropping this...

PR #657, opened 2016-05-10
Add a top level panic handler

caglar10ur:


cmd/tether/main_darwin.go
```diff
@@ -26,7 +27,12 @@ import (
 )
 
 func main() {
-	defer halt()
+	defer func() {
```
fdawg4l: Do `defer` instructions get run when something PANICs?  Also, logrus has a panic handler.  See [here](https://github.com/deferpanic/dp-logrus) for an example.

caglar10ur: defer is the only way to catch a panic http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#panic_recover

Not sure why we want logrus middleware?

caglar10ur: Oh I see what you are saying. Yeah not sure if it catches a panic from a function that's called from main

caglar10ur: https://play.golang.org/p/45Ctcn3cgb unless I misunderstood what you are asking

fdawg4l:
LGTM.  Good work.  Seems useful!

PR #659, opened 2016-05-11
disable cgo for darwin tether build

caglar10ur:
Fixes the integration tests as they fail with following;

```
building tether-darwin
# crypto/x509
gcc: error: unrecognized command line option '-mmacosx-version-min=10.6'
Makefile:217: recipe for target 'bin/tether-darwin' failed
make: *** [bin/tether-darwin] Error 2
Build step 'Execute shell' marked build as failure
```


gigawhitlocks:
LGTM

PR #661, opened 2016-05-11
Move portlayer/vsphere/storage into storage/vsphere pkg

fdawg4l:
I wanted to keep all the vsphere specific implementation stuff for all
of our portlayer handlers in the vsphere package but turns out no other
module was doing this which led to some confusion.  Might was well keep
this vsphere specific storage implementation in the portlayer storage
package.


cgtexmex:
LGTM

PR #668, opened 2016-05-12
Rename storage package for vsphere impl to vsphere

fdawg4l:
Renamed the directory in an earlier change without renaming the package.
Fixed for parity;  directory and package now have the same name
(vsphere).


cgtexmex:
LGTM

PR #669, opened 2016-05-12
Use bats assertions

dougm:
assert_success in particular is useful as bats will include the failed
command output.


fdawg4l:
Commit message- "will output will include".  LGTM otherwise.

jzt:
lgtm

gigawhitlocks:
lgtm

PR #670, opened 2016-05-12
Support resizing of pty windows

fdawg4l:
Exports the SystemInteraction interface.  Was previously exporting Connection which had all of its members private.  

Also removed some `payload` reply cruft which isn't/can't be used on an ssh channel.  

Moved the resize message to the `attach` package.


dougm:
lgtm

PR #671, opened 2016-05-13
Deleting overview from tech-preview branch.

stuclem:
Since the Overview doc applies to the official VIC release more than to tech preview, this PR is to delete the overview files from the tech-preview branch in vmware/vic.


dougm:
lgtm

PR #674, opened 2016-05-13
Add signal to attach server and cleanup ssh messages to come out of attach package

fdawg4l:
The majority of this code is cleanup.  We had messages defined in (the main package) of tether but were being used in both tether and PL.

### DIFF
--- a/portlayer/attach/messages.go
+++ b/portlayer/attach/messages.go
@@ -51,3 +51,44 @@ func (wc *WindowChangeMsg) Marshal() []byte {
 func (wc *WindowChangeMsg) Unmarshal(payload []byte) error {
 	return ssh.Unmarshal(payload, wc)
 }
+
+var (
+	Signals = map[ssh.Signal]int{
+		ssh.SIGABRT: 6,
+		ssh.SIGALRM: 14,
+		ssh.SIGFPE:  8,
+		ssh.SIGHUP:  1,
+		ssh.SIGILL:  4,
+		ssh.SIGINT:  2,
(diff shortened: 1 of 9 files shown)


portlayer/attach/messages.go
```diff
@@ -14,17 +14,100 @@
...
 
+import "golang.org/x/crypto/ssh"
+
 // All of the messages passed over the ssh channel/global mux are (or will be)
 // defined here.
 
-const (
-	WindowChangeReq = "window-change"
-)
+type Messager interface {
```
hmahmood: Should this just be `Message`?

fdawg4l: You don't like back-rubs?  :laughing: 

https://golang.org/doc/effective_go.html#interface-names

It's an interface for sending messages.  So messagER seems right to me.

hmahmood: It seems to be just serializing/deserializing messages and storing its type. Are you planning on adding more methods that would actually send/receive messages?

fdawg4l: I just realized messager != massager.  Please ignore my nonfunny joke.

fdawg4l: You're right;  the convention is for single method interfaces.  

Maybe we'll add more. It does serialize/deserialize and store data in messages we pass over ssh.  Actually sending is tricky.  I may add that later.

I don't feel strongly about the name and can change it to `Message`.

fdawg4l: Done.

hmahmood:
LGTM

PR #675, opened 2016-05-13
Remove govet shadow check

dougm:
There are still issues with using vet from go 1.6.2 on Mac vs Linux.
Let's remove the check until go vet shadow check functions as expected.


jzt:
LGTM 💯

hmahmood:
LGTM

PR #676, opened 2016-05-14
move apiservers into cmd/servers

caglar10ur:


fdawg4l:
I like this change less than what we have now. 'Cmd' should have a directories for each binary we're building and an accompanying 'main.go'. I don't like Cmd being the root of everything.

dougm:
We should avoid any renames/package moves until Loc get his PR submitted to avoid possible merge conflicts for now.

Maybe we should add a little naming conventions section to the contrib doc first too.

cgtexmex:
Agree with @fdawg4l I like what we have currently over this change.  If anything I'd refactor the "apiservers" to something like this:

/api
  /engine
    main.go (server)
  /portlayer
     main.go
 / etc..

caglar10ur:
@fdawg4l if I were up to me I would put apiservers/engine into cmd/engine and apiservers/portlayer into cmd/portlayer but I believe @hickeng suggested using a top level servers directory under cmd when we discuss this structure in person.

@dougm sure, I sent this cause it was sitting in my local git repo, not urgent at all.

@cgtexmex I don't personally like lots of directories in the top level. If it was up to me, I would put everything that has a main under cmd/, shared libraries under pkg/ and find different homes for the rest to end up with something like this;

cmd/...
pkg/....
vendor/...
docs/...
hacks/... (for scripts etc.)
contrib/... (for machines)
metadata/ (if it really need to sit on the top level instead of pkg)
isos/... 
templates/.. (assuming that's what swagger wants)

hickeng:
The reason for the server suggestion is that if everything with a main and it's dependencies ends up in cmd then everything ends up in cmd.
@caglar10ur's mention of pkg for shared dependencies explains some things - we've not all been using that in the same manner. I have been treating it as a location for packages that are _not_ VIC specific and therefore could be consumable by other projects.

I'm not a fan of the current layout for portalyer & engine api servers - I find it confusing. It's not clear to me how much of that structure is imposed by swagger and how much is evolved by us.
I'll take this opportunity to say that I also don't like `engine` as folder in apiservers at the same level as docker and portlayer as it implies to me that we have 3 APIs being served.

fdawg4l:
> The reason for the server suggestion is that if everything with a main and it's dependencies ends up in cmd then everything ends up in cmd.

If and only if all of your code is in `package main`.

fdawg4l:
```

**Comment by fdawg4l** [comment]
I think we all have an opinion here and there is no _the right way_.  I just don't like what we have now and this change doesn't get us closer to anywhere I'm happy with.   But I'll live with whatever the majority decides.

My suggestion is as follows.  Feel free to ignore/optimize.
```
/portlayer/{storage, network, metadata, other nouns that make up the port layer}/
/tether/{packages which compose tether}
/vicmachine/{packages which compose vic machine}
/{new component binary}/{its packages}

/common/{common packages used by > 1 component}

/cmd/{portlayer-daemon, docker-api-daemon, tether, vicmachine, other binaries we own and require in an vic deployment}/main.go

# For useful scripts.·
/scripts/{test, machines, other intended use cases}/*.sh

/docs/

# swagger dependencies
templates/{

isos/

# pubicly consumable independent packages
pkg/

# vendor
vendor/
```

hmahmood:
I agree with @fdawg4l on this one ... I would like to see cmd/portlayer, cmd/docker-engine, etc., and not cmd/servers/...

hickeng:
@caglar10ur @hmahmood @fdawg4l Take a look at #762 and let me know what you think. It doesn't have cmd/portlayer as I've not spent time to figure out how to get swagger to use a provided main instead of generating it.

caglar10ur:
Closing this as #762 is in now...

PR #677, opened 2016-05-14
ESX vm updates

dougm:
- Install host client

- Enable MOB


cgtexmex:
This will require the ESX host to have internet access -- probably not an issue in most cases, but curious how the esx cli will react if that's not the case

dougm:
We are assuming it does.  The old bonneville scripts installed the vib this way, but via the PXE install script.

cgtexmex:
LGTM

PR #679, opened 2016-05-14
Make go-deps silent by default

dougm:
These messages were included in 'gopath' target output, breaking GOPATH
when set via go-projectile.

- go-deps.sh is now silent unless make is given the '-d' flag (or error)

- script is now shellcheck clean


caglar10ur:
LGTM :sheep: :it:

Issue #680, opened 2016-05-14
Docker API:  Docker Resize (TTY)


sflxn:
Implement the docker container TTY resize operation in the engine-api server and integrate with the interaction port layer

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.23/#resize-a-container-tty

Please note, the engine-api's rest server already packaging these json struct into golang structs. The url above is just for reference on what the operation should do.

sflxn:
This operation is required for Docker Attach #362 (from the CLI perspective).  I've hooked up the code on the docker API and portlayer's interaction servers.  It's also now calling the ssh server to resize the container tty.  However, I am not yet able to get the session for the container (for connection attach) so there is no way to know whether resize works.

Docker Attach from the CLI is a composite of 3 docker API (important to distinguish CLI from API) oeprations, docker inspect, docker attach, and docker resize.  The CLI calls those APIs in that order.  Since we're currently hanging on docker attach right now, resize has yet been called.

Issue #647, opened 2016-05-06
Portlayer REST infrastructure for attach


sflxn:
Add the Swagger/REST code to support attach.  This is different from the code that connects to the tether sessions.

This is setting up the Swagger server, performing the hijack, and attaching the streams.  Other than the Swagger code, most of this will be similar to what the docker api server is doing with the CLI.

sflxn:
I've added Swagger infrastructure for Attach and Resize to the new interaction handler.  I'm now hooking up the hijack code.  In the process of doing this, I realize our Attach handler (via Swagger) will not work.  The current version of Swagger requires us to handle hijacking via an HTTP middleware.  However, we need to pass on the streams to the handlers, but Swagger provides no method of handling this.  Therefore, we're going to have to bypass the standard Swagger handling for this milestone and handle Attach completely in the middleware.

@casualjim has told me the latest version of Swagger adds a few parameters to every handler that may allow us to perform the hijack within the handler itself.  Right now, we just have no bandwidth to migrate to the current version of go-swagger.  We'll just have to implement attach via a middleware now, perform a go-swagger update (post milestone), and then reevaluate.

hickeng:
@sflxn Should this be closed?

Issue #388, opened 2016-03-31
Docker API: Image Inspect


sflxn:
Implement the docker images operation in the engine-api server and integrate with storage port layer.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#inspect-an-image
https://docs.docker.com/engine/reference/commandline/inspect/

Please note, the engine-api's rest server already packaging these json struct into golang structs. The url above is just for reference on what the operation should do.

sflxn:
This operation was part of docker attach so it had to be implemented.  Currently, we are just returning a cached version of the metadata that is a union of the metadata we get from docker and the metadata overrides from the CLI during docker create.

We are not yet going to the portlayer to rebuild that data.  This is a pass we must address to make it more robust.  We also need to update this metadata with events (docker events, vch events, etc).  That will come later.

hickeng:
@sflxn please open an additional issue for the reconstruction from portlayer - there is an entire milestone for dealing with appliance restart as it has impact across the entire codebase.

sflxn:
I am splitting this task into two task.  One for Inspect Image and one for Inspect Container.  Container Inspect relies on portlayer work that involves @cgtexmex.

sflxn:
This is done.  PR and waiting for comments.

Issue #362, opened 2016-03-31
Docker API: Docker Attach


sflxn:
Implement the ability to stream an interactive session with the running process in a container.  This work is integrating our implementation of the engine-api servr to our interaction port layer.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#attach-to-a-container
https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#attach-to-a-container-websocket

Please note, the engine-api's rest server already packaging these json struct into golang structs. The url above is just for reference on what the operation should do.

hickeng:
@sflxn please add an estimate

sflxn:
I've ported the Docker code to do attach.  The original code is vendored, but the modified version will sit in vic/pkg/docker.  The code is actually very simple, and I considered just writing it myself, but they seem to have a lot more code than seems to be necessary.  I must assume they have all that code to handle edge cases.  That's the primary reason for reusing their code instead of rewriting well known code patterns.

sflxn:
The attach code in the docker api server has some issue with tty input.  This is the tty input that's set up by the client side CLI.  I suspect the issue may be around either the resize operation or setting to raw mode.

I've taken a break from this to work on code to make a connection to the port layer.  I haven't yet decided to reuse docker's code for this or just write my own.  Again, this is really simple code and Docker's code in the CLI takes a very circuitous path.  Way more involved than it seems like it should.  It should just be an HTTP call and grabbing the conn, but their code makes dozens and dozens of call.  Perhaps it's all needed.  I don't know.  I'm going to take a second look at it.

sflxn:
One last update, the attach code on the api server side was more involved than originally thought.  Attach on the CLI side is actually a composition of calls to inspect, resize, and attach.  Inspect wasn't originally in the scope of this milestone.  I know from looking at docker's code that the CLI doesn't expect much from the metadata sent back from inspect.  We do not need to fully implement inspect to demonstrate attach.  For the first pass, I am dummying data in inspect.  If I have time today, I will add inspect to the port layer side, but that is very optimistic.  I'll see if I can't add that over the weekend.

sflxn:
[...]

PR #681, opened 2016-05-14
Docker Attach (API and Portlayer rest servers)

sflxn:
**NOTE!!!! - This PR contains two commits.  One is very large because it involves consolidating vendoring of some Docker sources.  When reviewing, click on the SECOND link below for the "Implemented Docker Attach..." and you will see VIC's code changes and updates!**

Implemented docker attach on the Docker API server and Port Layer's
REST server.  Attach involved,

1. Docker API Inspect, Resize, Attach
2. Portlayer hijack middleware (bypasses Swagger for attach)
3. Portlayer hookup for the docker API equivalent above

Also, Sync'd the VIC backend code to match Docker's v1.11 sources
that was recently vendored.  They were mostly function signatures.

#362, #647, #388, #680


sflxn:
An easier view of VIC's actual changes can be seen from my fork,

https://github.com/sflxn/vic/commit/9c1672f441956c51ebf78519ddc196cdd9271cac

sflxn:
Currently, stuck here https://github.com/sflxn/vic/blob/9c1672f441956c51ebf78519ddc196cdd9271cac/apiservers/portlayer/restapi/middleware/attach_hijacker.go#L87

@fdawg4l @hickeng 

```INFO[0000] Started reaping child processes
INFO[0000] [ END ] [main.(*osopsLinux).setup] [914.587µs] run OS specific tether setup
INFO[0000] Loaded executor config
INFO[0000] [BEGIN] [main.(*osopsLinux).SetHostname] setting hostname to 7988d4784b02
INFO[0000] [ END ] [main.(*osopsLinux).SetHostname] [37.951µs] setting hostname to 7988d4784b02
INFO[0000] [BEGIN] [main.(*attachServerSSH).start] start attach server
INFO[0000] [ END ] [main.(*attachServerSSH).start] [133.52µs] start attach server
INFO[0000] Launching process for session 7988d4784b02b125bf6f89c2ff8f8f3b75447cf8ba3b201db83bbc8184d2524d

INFO[0000] [BEGIN] [main.launch] launching session 7988d4784b02b125bf6f89c2ff8f8f3b75447cf8ba3b201db83bbc8184d2524d
INFO[0000] [BEGIN] [main.(*osopsLinux).sessionLogWriter] configure tether session log writer
INFO[0000] opening ttyS2 for session logging
INFO[0000] [ END ] [main.(*osopsLinux).sessionLogWriter] [1.40713ms] configure tether session log writer
INFO[0000] Launching command []string{"/bin/top"}

INFO[0000] [BEGIN] [main.(*attachServerSSH).run] main attach server loop
INFO[0000] [BEGIN] [main.(*osopsLinux).backchannel] establish tether backchannel
INFO[0000] opening ttyS0 for backchannel
INFO[0000] s = &terminal.State{termios:syscall.Termios{Iflag:0x500, Oflag:0x5, Cflag:0xcbd, Lflag:0x8a3b, Line:0x0, Cc:[32]uint8{0x3, 0x1c, 0x7f, 0x15, 0x4, 0x0, 0x1, 0x0, 0x11, 0x13, 0x1a, 0x0, 0x12, 0xf, 0x17, 0x16, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}, Pad_cgo_0:[3]uint8{0x0, 0x0, 0x0}, Ispeed:0x0, Ospeed:0x0}}
INFO[0000] creating raw connection from ttyS0 (fd=17)

INFO[0000] [ END ] [main.launch] [31.014159ms] launching session 7988d4784b02b125bf6f89c2ff8f8f3b75447cf8ba3b201db83bbc8184d2524d```

The process is running in the VM.  I've verified it from the console.

docker create -it busybox /bin/top  (which there is a bug here I may have to write up a bug against the tether)
docker start <id>
docker attach <id>

**Comment by fdawg4l** [comment]
Did you use vicmachine to install the vch? Also place post the port layer log.

**Comment by fdawg4l** [comment]
*please. I'm afk until tomm and can be more useful then.

**Comment by fdawg4l** [comment]
Also 'hijack'* is terrible naming IMHO.

**Comment by fdawg4l** [comment]
We're looking for output on the PL side that looks similar to [this](https://github.com/vmware/vic/issues/440#issuecomment-218292284)

**Comment by hickeng** [comment]
You will find that there is currently a timing hole between start returning and the process actually having been kicked off. That's what #653 is addressing - if this is blocking then I can push a corrected version of it tomorrow sometime but doesn't look like it is.
Is the containerVM correctly configured with a network serial port that connects back to the appliance?
@sflxn **do you have the ESX firewall disabled or configured to allow outbound 2377?**

**Comment by dougm** [comment]
@sflxn are you using a manually created ESX vm?  You'll find this config in our various Vagrantfiles for esx:

https://github.com/vmware/govmomi/blob/master/govc/test/esxbox/Vagrantfile#L21-L22

And our automated cluster config:
https://github.com/vmware/vic/blob/master/machines/vcsa/create-cluster.sh#L105-L106

If that is the issue you're having, you can use either of those commands to fix.

File: apiservers/portlayer/restapi/handlers/interaction_handlers.go [hunk 1]
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/vmware/vic/portlayer/network"
+)
+
+// ExecHandlersImpl is the receiver for all of the exec handler methods
+type InteractionHandlersImpl struct {
+	attachServer *attach.Server
+}
+
+var (
+	interactionSession = &session.Session{}
```
fdawg4l: Minor- I only see this used once and it's in this file.  Likely you don't need this package global.

fdawg4l:
I pulled this change, built it from a clean workspace, and installed it using `vicmachine`.  I brought up a `busybox` container running `/bin/top` and confirmed the ssh over serial attach channel works as expected.  The handshake passed on both sides (the tether and the PL) and the connection was established.  

Also looking at this change, I don't see any glaring errors other than things we can address during the code review (when the WIP prefix is dropped).  So, as I see it and have verified, there aren't any connection issues to speak of.

What's left is to get the docker api to PL attach workflow to do what we need it to.  I tried this using a `docker` 1.9 client and got the following.

```
$ docker -H 192.168.8.198:2376 --tls --tlscert='../cert.pem' --tlskey='../key.pem' start -a 2191f33fd1d1e054866da84cc2acef47119d7ff3953f887d583e785d696ec65e
Unrecognized input header

$ docker -H 192.168.8.198:2376 --tls --tlscert='../cert.pem' --tlskey='../key.pem' attach 2191f33fd1d1e054866da84cc2acef47119d7ff3953f887d583e785d696ec65e
 
Unrecognized input header
```

I'll try this again with this change and an up to date docker client tomorrow.

fdawg4l:
It built faster than I expected.  I reran this change with a 1.11 client and got the following

```
$ docker -H 192.168.8.196:2376 --tls --tlscert='../cert.pem' --tlskey='../key.pem' create busybox /bin/top                                                
ec106f9b9990704d4179fbb773c931398d95c3799162de4f41ad338e48b6c640

$ docker -H 192.168.8.196:2376 --tls --tlscert='../cert.pem' --tlskey='../key.pem' start -a ec106f9b9990704d4179fbb773c931398d95c3799162de4f41ad338e48b6c640
Unrecognized input header: 27

$ docker -H 192.168.8.196:2376 --tls --tlscert='../cert.pem' --tlskey='../key.pem' attach  ec106f9b9990704d4179fbb773c931398d95c3799162de4f41ad338e48b6c640
Unrecognized input header: 27

$ docker version
Client:
 Version:      1.11.1
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   5604cbe
 Built:        Tue Apr 26 23:30:23 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.11.1
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   5604cbe
 Built:        Tue Apr 26 23:30:23 2016
 OS/Arch:      linux/amd64

```

sflxn:
I cleaned out my entire datastore, re-installed, and I am still having connection problem.  We tried on @matthewavery machine this morning (first time he used vic-machine to do so), and he experienced the exact same issue.

dougm:
Didn't see any response to the firewall question, just to double check, can you paste the output of:

```
govc firewall.ruleset.find -direction outbound -port 2377
```

sflxn:
Both ESX and appliance VM has firewall disabled.

fdawg4l:
Just keeping this updated-  @hickeng and I got @sflxn up and running with MASTER so we know the environment is fine.  The tether connection to the PL is fine.

sflxn:
After we figured out how to get my portlayer server to successfully get the ssh session, I made some progress.

I figured out why output was not working correctly.  I forgot about Docker's multiplexed stdout/stderr handling.  Docker CLI looks for the multiplexed stream, and I wasn't using the multiplexed stream so it closed the pipe.

After I added the multiplex streaming, output works.  I can attach with /bin/top and see the results.  Trying out stdin now.

fdawg4l:
@sflxn this commit needs some rebase love.  There are merge conflict headers in the last commit.

sflxn:
I got the input stream working.  During docker create, we need to modify the ContainerConfig struct so we can return those values during docker inspect (1st operation of docker attach).  Some of those params are attachStdin, ..., openstdin, etc.  The original image's containerconfig has all these values set to false.  If you execute docker create -it busybox /bin/top, the CLI will request overrides on those values.

I also got resize working.  Swagger consumes octet-streaming tripped me up again.  Without this mime type, the resize call wasn't getting to the portlayer.

pkg/attachutils/simpleattach.go
```diff
@@ -0,0 +1,385 @@
...
+		conn.Write([]byte{})
+
+		if upgrade {
+			fmt.Fprintf(conn, "HTTP/1.1 101 UPGRADED\r\nContent-Type: application/vnd.vmware.raw-stream\r\nConnection: Upgrade\r\nUpgrade: tcp\r\n\r\n")
+		} else {
+			fmt.Fprintf(conn, "HTTP/1.1 200 OK\r\nContent-Type: application/vnd.vmware.raw-stream\r\n\r\n")
+		}
+
+		closer := func() error {
+			httputils.CloseStreams(conn)
```
fdawg4l: Why not just `conn.Close()`?  Looking at `httputils.CloseStreams` with @hickeng and we're not even sure if the `TCPConn` is being closed?  On first pass, it looks like it's only calling `CloseWrite()` and we're not sure if the whole connection is being closed.  Just call `Close()`.

pkg/attachutils/simpleattach.go
```diff
@@ -0,0 +1,385 @@
...
+	conn, _, err := hijacker.Hijack()
+	if err != nil {
+		return nil
+	}
+
+	// GetStreams hijacks the the connection from the http handler. This code originated
+	// in the docker engine-api image router.
+	setupStreams := func() (io.ReadCloser, io.Writer, io.Writer, error) {
+		// set raw mode
+		conn.Write([]byte{})
```
fdawg4l: Please clarify.  What is this setting and how?  It's writing with length 0; which we believe to be a NOOP.

apiservers/portlayer/restapi/middleware/attach_hijacker.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+
+	code, err := a.GetParams(r)
+	if err != nil {
+		log.Errorf("interaction attach container failed %#v\n", err)
+		a.WritePortlayerError(rw, code, err.Error())
+		return
+	}
+
+	getStream := attachutils.HijackConnection(rw, r)
```
fdawg4l: There's only one reference to this and the result (`getStream`) can only be used once.  Just take the resulting closure and embed it here.  Or flatten the `HijackConnection` method to return the right response to the client and return the streams directly.

pkg/attachutils/simpleattach.go
```diff
@@ -0,0 +1,385 @@
...
+
+	// Connect stdin of container to the http conn.
+	go func() {
+		if clientStdin == nil {
+			return
+		}
+		log.Debugf("attach: stdin: begin")
+
+		var err error
+		if tty {
```
fdawg4l: The `tty` being allocated on the container happens during `create` in the exec layer.  In effect, we never need to worry about this in the portlayer attach code. 

This means copyEscapable can be removed the PL.

apiservers/portlayer/restapi/middleware/attach_hijacker.go
```diff
@@ -0,0 +1,193 @@
...
+	// Get the stdio streams from the hijacked connection
+	hijackedStdin, hijackedStdout, hijackedStderr, err := getStream()
+	if err != nil {
+		a.WritePortlayerError(rw, http.StatusInternalServerError, "Error getting hijacked streams")
+		return
+	}
+
+	log.Printf("Attempting to get ssh session for container %s", a.containerID)
+	// Get the ssh session streams
+	sshConn, err := a.attachServer.Get(context.Background(), a.containerID, 600*time.Second)
```
fdawg4l: This shouldn't have a timeout.  We should wait forever.  Really the context being passed in (`Background()` currently) should have a `WithCancel()` so the client can drop and cancel the `Get()`.

pkg/attachutils/simpleattach.go
```diff
@@ -0,0 +1,385 @@
...
+		}
+	}
+	return nil
+}
+
+// Code c/c from io.Copy() modified to handle escape sequence
+func copyEscapable(dst io.Writer, src io.ReadCloser, keys []byte) (written int64, err error) {
+	if len(keys) == 0 {
+		// Default keys : ctrl-p ctrl-q
+		keys = []byte{16, 17}
```
fdawg4l: The only reason this function exists in the docker code is to handle this condition.  In our case, we can do better.  If we were to create a REST endpoint on the PL which handles "detach", then we simply need to capture the cntrl-p cntrl-q in the docker personality and call the REST detach to close the connection without closing `stdin`, `stdout`, `stderr` on the process in the container.

pkg/attachutils/simpleattach.go
```diff
@@ -0,0 +1,385 @@
...
+
+// CloseWriter is an interface that implement structs
+// that close input streams to prevent from writing.
+type CloseWriter interface {
+	CloseWrite() error
+}
+
+// Code copied from Docker and modified.  stdin, stdout, stderr are the client
+// (hijacked connection) and conn, bufReader are the server
+func AttachStreamsToConn(openStdin, stdinOnce, tty bool, stdin io.ReadCloser, stdout io.Writer, stderr io.Writer, conn net.Conn, bufReader *bufio.Reader, keys []byte) error {
```
fdawg4l: This function and the one below it are sufficiently similar that they can be made more generic and condensed into 1 function.

fdawg4l:
Chatting w/ @hickeng and we have a few design questions and a few alternatives to reusing docker's code in this way.

Currently, it looks like the docker/hijack stuff is happening in both the personality AND the portlayer.  However, we own both pieces-  the portlayer and the personality.    There's really no need to reuse docker's implementation between our own components.  And we can simplify some of their design decisions.  Most of the complexity here is trying to multiplex 3 streams across a single connection and implement effectively 3 verbs (attach, detach, close)

Here's an alternative-
On the portlayer, we create 3 REST verb endpoints- 
```
# Get the output stream from the given file descriptor
GET {containerid}/attach?fd={STDOUT, STDERR}

# Close the given output stream.  
DELETE {containerid}/attach?fd={STDIN, STDOUT, STDERR}

# POST the input stream on STDIN
POST {containerid}/attach
```

The GET and POST are streaming actions and need to be handled (`octet/stream`) accordingly.  Other than that, we get out of 1) using docker's multiplexing mechanism and their code, and 2) simplifies our implementation without tying ourselves to someone else's technical debt.

On the docker personality side, you would create an appropriate request for each stream.  Specifically on `docker attach`, you'd implement a GET on STDOUT and another on STDERR and pass the resulting streams to the docker stream apis like you are now.  Likewise, you'd need to open a POST to the container for STDIN or explicitly close it (via `DELETE`).

`detach` becomes `conn.Close()` on each REST request which is *different* from closing `STDIN`, `STDERR`, or `STDOUT`.  We need to support both operations (`detach` and close) independently.

sflxn:
Sure, we can use 3 separate connection for this.  I am not really tied to Docker's approach.  I simply reused their code for expediency.  My thoughts on this is, let's wait till post 0.2 to change this.  There are actually many approaches to solving this problem.  Reusing their code was simpler because the stream handling isn't the only thing I had to figure out with their Attach operation.  Reverse engineering the composition of their call, the metadata handling, figuring out their container state management, etc.  Reusing their stream attach freed up some time.

Let's get this into 0.2 and then clean up the technical debt after.  The attach functionality is basic.  There's still more work to be done, such as key handling, getting running state of VIC containers, and correct use of their metadata.  We can change the API <-> PL stream attachment while we address the other issues.  This is no different than how we're doing pull/create/start.  Start out with basic functionality and iterate after.

apiservers/portlayer/restapi/middleware/attach_hijacker.go
sflxn: I also thought of doing that.  I left it as is so we have more flexibility when we upgrade our Swagger version.  It would provide us with a mechanism to get at these streams in functions that don't have direct access to the http responsewriter.  However, I do see the point of keeping it simple now and have no issue changing it.

pkg/attachutils/simpleattach.go
sflxn: Yeah, I see your point.  If we handle this in the API server, there's no reason to do so in the PL.  We can just pass false as tty into this attach function.

pkg/attachutils/simpleattach.go
sflxn: Yes.  See my comments to your previous comments.

sflxn:
Updated PR with the suggested changes.  Also worked around the docker create regression (commit isn't returning no container error) so now you can do docker create and cause it to pull images if it hasn't been pulled again.  The start still has problems.  I left that to Hasan to figure out.

fdawg4l:
> Let's get this into 0.2 and then clean up the technical debt after.

I'll leave this for @hickeng to make that call.  My opinion is the changes asked for here aren't that large but I'm sure easier said than done. 

I'll continue reviewing in the case we want to move forward with this approach.  Bear in mind you'll be making changes to code that, IMHO, we shouldn't be using.

apiservers/engine/backends/container.go
```diff
@@ -20,90 +20,111 @@ import (
...
 func (c *Container) ContainerExecInspect(id string) (*backend.ExecInspect, error) {
 	return nil, fmt.Errorf("%s does not implement container.ContainerExecInspect", c.ProductName)
 }
 
+// ContainerExecResize changes the size of the TTY of the process
+// running in the exec with the given name to the given height and
+// width.
 func (c *Container) ContainerExecResize(name string, height, width int) error {
-	return fmt.Errorf("%s does not implement container.ContainerExecResize", c.ProductName)
+	return nil
```
fdawg4l: This looks like a NOOP.  Is that intentional?

apiservers/engine/backends/container.go
```diff
@@ -445,3 +607,67 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+
+	imageMetadata := &viccontainer.VicContainer{
+		ID:     v1.ID,
+		Config: &v1.Config,
+	}
+
+	return imageMetadata, nil
+}
+
+func (c *Container) getContainerConfigFromExecPL(name string) (*container.Config, error) {
```
hickeng: This doesn't appear to be called from anywhere.

apiservers/engine/backends/container/viccontainer.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func NewVicContainer() *VicContainer {
+	return &VicContainer{
+		StreamConfig: runconfig.NewStreamConfig(),
+		Config:       &containertypes.Config{},
+	}
+}
+
+func GetCache() *Cache {
```
fdawg4l: https://golang.org/doc/effective_go.html#Getters

apiservers/engine/backends/container.go
```diff
@@ -445,3 +607,67 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+	// Get an API client to the portlayer
+	client := PortLayerClient()
```
hickeng: We're not initializing the client lazily and this doesn't perform additional checking. Would be cleaner to use a global var directly and avoid the unreachable nil check.

apiservers/engine/backends/container/viccontainer.go
```diff
@@ -0,0 +1,84 @@
...
+	if cache == nil {
```
fdawg4l: Is there an advantage to this lazy instantiation?  Also this doesn't look threadsafe.  If i call GetCache() from 2 go routines, may get 2 different caches (because of the race on the nil check).

apiservers/portlayer/restapi/handlers/interaction_handlers.go
hickeng: The interaction portion of the portlayer doesn't touch vSphere at all, so should not need a session object.

apiservers/engine/backends/container/viccontainer.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (cc *Cache) SaveContainer(name string, container *VicContainer) {
+	cc.m.Lock()
+	defer cc.m.Unlock()
+
+	cc.containerStore[name] = container
+}
+
+func (cc *Cache) Lock() {
```
fdawg4l: Every member in this struct is private to the package.  Why does the write lock need to be public?

Also I don't see this or the `Unlock` method being used anywhere.

apiservers/engine/backends/container.go
sflxn: We're not doing docker exec yet.

fdawg4l:
@sflxn looks like tests are passing on this change.  Is it ready to be reviewed or should we hold off looking at it?

apiservers/engine/backends/container.go
fdawg4l: If we don't support it yet, then should we revert this line and return the error or is that error getting in the way somewhere else?

sflxn:
@fdawg4l I'm going to look at the ReadFrom to get rid of the double copy and look at the buffering.  You're welcome to look at it, but it's not fully working yet.

fdawg4l:
@sflxn I know it's a big ask, but can you squash your commits?  The changelist is too big to look at it all in one go, and looking at individual commits is proving problematic since it includes your previous design.

Can you split the commit into the following-
* keep the "Updated docker..." vendor commit.  I think this is where the bulk of the diff comes from.
* squash Ivan's changes into 1 new commit
* then squash your attach changes into 1 new commit.

I think doing the above would make reviewing slightly less painful since we don't get distracted with vendor changes, swagger changes, and changes on your previous design plus your previous design changes.  Know what I mean?

caglar10ur:
why is some commits (like https://github.com/vmware/vic/pull/681/commits/2752c23899e64a050daf57d6d0cf7e2cd383418a) changing vendored code?

caglar10ur:
same for https://github.com/vmware/vic/pull/681/commits/a890c4a8d2f277f3421efc585863de93ae5a9c8d

casualjim:
@caglar10ur because the swagger codebase has diverged and packages have been reorganized. The changes to the generated code are quite substantial and for now the fixes that are required have been implemented in both your vendor as well as the upstream version of swagger.   Having updates made to vendored code which are also made upstream doesn't seem like too big of an issue because you're checking it in.

Until I close ticket go-swagger/go-swagger#476 upgrading would still be a moving target because the way pointers are treated. Currently everything that used to be a pointer would not be one anymore and pretty everything that wasn't a pointer would become one. After 476 there are no pointer struct fields anymore except for complex objects.  After go-swagger/go-swagger#476 the models should be in a stable state and not change their public appearance.

I was hoping I'd be able to finish this in a few days but the change proves to be much larger, because it requires me to implement a new serialization/deserialization mechanism for the generated code (implementation detail).

The benefit you'd get out of it besides faster and less memory hungry serialization is the ability to actually use the PATCH verb.

lib/apiservers/engine/backends/container/viccontainer.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/docker/docker/runconfig"
+	containertypes "github.com/docker/engine-api/types/container"
+)
+
+// This is VIC's abridged version of Docker's container object.
+type VicContainer struct {
+	*runconfig.StreamConfig
+
+	ID          string
+	ContainerID string
```
andrewtchin: What's the difference between ID and ContainerID?

sflxn:
@caglar10ur:  I agree that's a cleaner approach, but for this one time, I'd vote for letting it slide.  @casualjim have some major changes to go-swagger and we're going to upgrade to it in a week or two.  It's now up to a calculation of where to best spend our time.  Ivan can spend his time getting the big go-swagger changes and refactor in.  That seems like better use of time.  If others disagree, we can can revisit.

lib/apiservers/engine/backends/container/viccontainer.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func GetCache() *Cache {
+	if cache == nil {
+		cache = &Cache{containerStore: make(map[string]*VicContainer)}
+	}
+	return cache
+}
+
+func (cc *Cache) GetContainerByName(name string) *VicContainer {
```
hickeng: Go conventions frown on using `Get` prefix for getters (https://golang.org/doc/effective_go.html getters section)

lib/apiservers/engine/backends/container/viccontainer.go
```diff
@@ -0,0 +1,84 @@
...
+// Tracks our container info from calls
+type Cache struct {
+	m sync.RWMutex
+
+	containerStore map[string]*VicContainer
+}
+
+var cache *Cache
+
+func init() {
```
hickeng: Is this a placeholder for somthing else? Go will initialize the pointer to nil by default.

If we're using a global var, why provide the GetCache method? If this is to allow mocking or alternative implementations then we need an interface as well.

lib/apiservers/engine/backends/container/viccontainer.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (cc *Cache) SaveContainer(name string, container *VicContainer) {
+	cc.m.Lock()
+	defer cc.m.Unlock()
+
+	cc.containerStore[name] = container
+}
+
+func (cc *Cache) Lock() {
```
hickeng: Doesn't look like these need to be exported - they're only called from SaveContainer/GetContainerByName.

caglar10ur:
@sflxn sure, I'm fine with letting it slide if this is really for short time. I'll take do a first pass on the changeset

lib/apiservers/engine/backends/container.go
```diff
@@ -20,88 +20,113 @@ import (
...
 
+// ContainerExecStart starts a previously set up exec instance. The
+// std streams are set up.
 func (c *Container) ContainerExecStart(name string, stdin io.ReadCloser, stdout io.Writer, stderr io.Writer) error {
 	return fmt.Errorf("%s does not implement container.ContainerExecStart", c.ProductName)
 }
 
+// ExecExists looks up the exec instance and returns a bool if it exists or not.
+// It will also return the error produced by `getConfig`
 func (c *Container) ExecExists(name string) (bool, error) {
```
fdawg4l: `ContainerExecInspect` uses an id, whereas `ContainerExecResize, `ContainerExecStart, and `ExecExists` use names.  First is `name` the human readable name returned by the daemon?  Second is name interchangeable with id?  Please provide a comment in the three mentioned functions on what `name` means to clarify.

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +653,235 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+	for err := range errors {
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// Code c/c from io.Copy() modified by Docker to handle escape sequence
+func copyEscapable(dst io.Writer, src io.ReadCloser, keys []byte) (written int64, err error) {
```
hickeng: This is copied docker code - can we have this is a separate file or package? I know it's inconvenient but it helps keep the provenance of code clear.

cmd/docker/main.go
```diff
@@ -168,7 +168,7 @@ func startServerWithOptions(cli *CliOptions) *apiserver.Server {
 
 func setAPIRoutes(api *apiserver.Server) {
 	imageHandler := &vicbackends.Image{ProductName: productName}
-	containerHandler := &vicbackends.Container{ProductName: productName, HackMap: make(map[string]vicbackends.V1Compatibility)}
```
caglar10ur: yihuu, finally getting rid of this!

lib/apiservers/engine/backends/container.go
```diff
@@ -20,88 +20,113 @@ import (
...
 	return nil, fmt.Errorf("%s does not implement container.ContainerExecInspect", c.ProductName)
 }
 
+// ContainerExecResize changes the size of the TTY of the process
+// running in the exec with the given name to the given height and
+// width.
 func (c *Container) ContainerExecResize(name string, height, width int) error {
-	return fmt.Errorf("%s does not implement container.ContainerExecResize", c.ProductName)
+	return nil
+	//	return fmt.Errorf("%s does not implement container.ContainerExecResize", c.ProductName)
```
caglar10ur: Why we are returning nil here instead of calling ContainerResize?

lib/apiservers/engine/backends/container.go
```diff
@@ -118,43 +143,18 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
...
-		var v1 struct {
-			ID string `json:"id"`
-			// https://github.com/docker/engine-api/blob/master/types/container/config.go
-			Config container.Config `json:"config"`
-		}
-		if err := json.Unmarshal(out, &v1); err != nil {
-			return types.ContainerCreateResponse{},
-				derr.NewErrorWithStatusCode(fmt.Errorf("Failed to unmarshall image history: %s", err),
-					http.StatusInternalServerError)
+			return types.ContainerCreateResponse{}, err
```
fdawg4l: Is `err` appropriate here?  Looks like above this where we return an error, we use `derr.NewErrorWithStatusCode`.

lib/apiservers/engine/backends/container.go
```diff
@@ -217,38 +217,103 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
...
+	layer.Config.Env = config.Config.Env
+	layer.Config.AttachStdin = config.Config.AttachStdin
+	layer.Config.AttachStdout = config.Config.AttachStdout
+	layer.Config.AttachStderr = config.Config.AttachStderr
+	layer.Config.Tty = config.Config.Tty
+	layer.Config.OpenStdin = config.Config.OpenStdin
+	layer.Config.StdinOnce = config.Config.StdinOnce
+	layer.ContainerID = createResults.Payload.ID
+
+	viccontainer.GetCache().SaveContainer(createResults.Payload.ID, layer)
```
caglar10ur: Do we need a lock for serializing this for concurrent requests?

lib/apiservers/engine/backends/container.go
```diff
@@ -367,48 +438,124 @@ func (c *Container) ContainerStop(name string, seconds int) error {
...
+
+	defer trace.End(trace.Begin("ContainerInspect"))
+
+	// Look up the container info in the metadata cache
+	vc := viccontainer.GetCache().GetContainerByName(name)
+	if vc == nil {
+		return nil, derr.NewRequestNotFoundError(fmt.Errorf("No such container: %s", name))
+	}
+
+	// HACK: Just set a bunch of dummy info needed to get docker attach working
```
caglar10ur: I think we need a FIXME here to indicate required future work, HACK sounds like we are done with it :)

lib/apiservers/engine/backends/container.go
```diff
@@ -217,38 +217,103 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
...
+	plResizeParam := interaction.NewContainerResizeParams().WithID(name).WithHeight(plHeight).WithWidth(plWidth)
+
+	_, err := client.Interaction.ContainerResize(plResizeParam)
+	if err != nil {
+		if _, isa := err.(*interaction.ContainerResizeNotFound); isa {
+			return derr.NewRequestNotFoundError(fmt.Errorf("No such container: %s", name))
+		}
+
+		// If we get here, most likely something went wrong with the port layer API server
+		return derr.NewErrorWithStatusCode(fmt.Errorf("Unknown error from the interaction port layer"),
```
fdawg4l: Can you log the error and append it's string to this `fmt` string?  `fmt.Errorf("Unknown error {....}: %s", err)`.

lib/apiservers/engine/backends/container.go
```diff
@@ -367,48 +438,124 @@ func (c *Container) ContainerStop(name string, seconds int) error {
...
+	base := &types.ContainerJSONBase{
+		State: &types.ContainerState{Status: "running",
+			Running: true,
+			Paused:  false,
+		},
+	}
+
+	conJSON := &types.ContainerJSON{ContainerJSONBase: base, Config: vc.Config}
+
+	log.Printf("ContainerInspect json config = %+v\n", conJSON.Config)
```
fdawg4l: Maybe this is appropriate for `Debugf`?

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +653,235 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+	}
+
+	if clStdin != nil {
+		// Pipe for stdin so we can interject and watch the input streams for detach keys.
+		stdinReader, stdinWriter := io.Pipe()
+
+		defer clStdin.Close()
+		defer stdinReader.Close()
+
+		transport := httptransport.New(PortLayerServer(), "/", []string{"http"})
```
caglar10ur: Do we use http for this or that  []string{"http"} means something else?

lib/apiservers/engine/backends/container.go
```diff
@@ -367,48 +438,124 @@ func (c *Container) ContainerStop(name string, seconds int) error {
...
+
+	if vc == nil {
+		//FIXME: If we didn't find in the cache, we should goto the port layer and
+		//see if it exists there.  API server might have been bounced.  For now,
+		//just return error
+
+		return derr.NewRequestNotFoundError(fmt.Errorf("No such container: %s", prefixOrName))
+	}
+
+	clStdin, clStdout, clStderr, err := ca.GetStreams()
```
fdawg4l: Looks like `err` isn't being caught.

lib/apiservers/engine/backends/container.go
```diff
@@ -367,48 +438,124 @@ func (c *Container) ContainerStop(name string, seconds int) error {
...
+		// replace the stdout/stderr with Docker's multiplex stream
+		clStderr = stdcopy.NewStdWriter(clStderr, stdcopy.Stderr)
+		clStdout = stdcopy.NewStdWriter(clStdout, stdcopy.Stdout)
+	}
+
+	if !ca.UseStdin {
+		clStdin = nil
+	}
+	if !ca.UseStdout {
+		clStdout = nil
```
fdawg4l: But on line `543`, we now have a new `NewStdWriter`.  Is that intentional?  If so, please add a comment here or refactor the `if` on line 540 to test for this condition before calling the New... routine.

lib/apiservers/engine/backends/container.go
```diff
@@ -367,48 +438,124 @@ func (c *Container) ContainerStop(name string, seconds int) error {
...
+	}
+	if !ca.UseStderr {
+		clStderr = nil
```
fdawg4l: Same comment as above.

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +653,235 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+		ID string `json:"id"`
+		// https://github.com/docker/engine-api/blob/master/types/container/config.go
+		Config container.Config `json:"config"`
+	}
+	if err := json.Unmarshal(out, &v1); err != nil {
+		return nil,
+			derr.NewErrorWithStatusCode(fmt.Errorf("Failed to unmarshall image history: %s", err),
+				http.StatusInternalServerError)
+	}
+	log.Printf("v1 = %+v", v1)
```
fdawg4l: Looks like debugging code.

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +653,235 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+	}
+
+	return imageMetadata, nil
+}
+
+func attachStreams(name string, tty, stdinOnce bool, clStdin io.ReadCloser, clStdout, clStderr io.Writer, keys []byte) error {
+	var wg sync.WaitGroup
+
+	errors := make(chan error, 3)
+	//FIXME: Swagger will timeout on us.  We need to either have an infinite timeout or the timeout should
```
fdawg4l: Prefer we don't time out ever if we're attached.  I don't like it when terminals thing they're smarter than me ;-).

lib/apiservers/engine/backends/container.go
caglar10ur: Never mind this as I just saw that SaveContainer uses a mutex- see this wouldn't happen if we were using gerrit :)

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -0,0 +1,233 @@
...
+
+	if err := i.attachServer.Start(); err != nil {
+		log.Fatalf("Attach server unable to start: %s", err)
+		return
+	}
+}
+
+func (i *InteractionHandlersImpl) ContainerResizeHandler(params interaction.ContainerResizeParams) middleware.Responder {
+	// Get the ssh session to the container
+	connContainer, err := i.attachServer.Get(context.Background(), params.ID, 600*time.Second)
```
caglar10ur: can we make this 600 sec. a const?

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -0,0 +1,233 @@
...
+		retErr := &models.Error{Message: "SSH Session failed to resize the container's TTY"}
+		return interaction.NewContainerResizeInternalServerError().WithPayload(retErr)
+	}
+
+	return interaction.NewContainerResizeOK()
+}
+
+func (i *InteractionHandlersImpl) ContainerSetStdinHandler(params interaction.ContainerSetStdinParams) middleware.Responder {
+	log.Printf("Attempting to get ssh session for container %s stdin", params.ID)
+	sshConn, err := i.attachServer.Get(context.Background(), params.ID, 600*time.Second)
```
caglar10ur: ^ ditto

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +653,235 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+	// start after some inactivity?
+	inContext, inCancel := context.WithTimeout(context.Background(), 2*time.Hour)
+	outContext, outCancel := context.WithTimeout(context.Background(), 2*time.Hour)
+	errContext, errCancel := context.WithTimeout(context.Background(), 2*time.Hour)
+
+	// Create writers that will redirect copying to the writer we really want to
+	// write to.  These redirect writers will allow us to close
+	if clStdin != nil {
+		wg.Add(1)
```
fdawg4l: `Add()` in the go routine itself.  That's the idiom i tend to see and it makes for more readable code.  Also it removes this `if` condition.  Same comment for the 2 other instances of this below.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -0,0 +1,233 @@
...
+
+func NewFlushingReader(rdr io.Reader) *FlushingReader {
+	return &FlushingReader{Reader: rdr, flusher: nil}
+}
+
+func (d *FlushingReader) AddFlusher(flusher GenericFlusher) {
+	d.flusher = flusher
+}
+
+// Derived from go's io.Copy
```
caglar10ur: Can we add a comment explaining why we needed this?

lib/apiservers/portlayer/swagger.yml
```diff
@@ -527,28 +527,129 @@ paths:
...
           required: true
+        - name: height
+          in: query
+          type: integer
+          format: int32
+          required: true
+        - name: width
+          in: query
+          type: integer
+          format: int32
```
caglar10ur: would it be possible to declare those two as int64 and remove the casting from the code?

lib/apiservers/engine/backends/container/viccontainer.go
sflxn: Making the change.

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +653,235 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+
+	imageMetadata := &viccontainer.VicContainer{
+		ID:     v1.ID,
+		Config: &v1.Config,
+	}
+
+	return imageMetadata, nil
+}
+
+func attachStreams(name string, tty, stdinOnce bool, clStdin io.ReadCloser, clStdout, clStderr io.Writer, keys []byte) error {
```
fdawg4l: Please add a comment as to what this function does, how it does it, and what the individual arguments to it are.  I think I understand how it works, but I'm sure I'll forget and a comment about it would be helpful.

lib/apiservers/engine/backends/container/viccontainer.go
sflxn: Removing init().  Not allowing access to global var is just my own practice.  I never like giving access to global vars.

lib/apiservers/engine/backends/container/viccontainer.go
sflxn: Removing.

lib/apiservers/engine/backends/container.go
sflxn: Yes, name encompasses Id, friendly name, partial id, etc.  These are the same signature that Docker uses for their backend interfaces.  We can make a pass and clean up the method signature to make it clear, but I'd like to do that in another PR, outside of this one.

lib/apiservers/engine/backends/container.go
sflxn: I can move copyEscapable into a package.  I just pulled it in based on the idea that the code is mostly from the go libraries and they also just inline the code in their sources, but if it makes the division clearer, I'll move it.

lib/apiservers/engine/backends/container.go
sflxn: This is a mistake.  We haven't done anything with Exec yet.  I'll go back to reporting "... does not implement..."

lib/apiservers/engine/backends/container.go
sflxn: I agree, we shouldn't be returning portlayer errors.  We should map the errors.  I believe @caglar10ur is working on that so we should defer to his future PR to fix this.

lib/apiservers/engine/backends/container.go
sflxn: The locking is done in SaveContainer.

lib/apiservers/engine/backends/container.go
sflxn: Will make the changes.

lib/apiservers/engine/backends/container.go
sflxn: Making the changes.

lib/apiservers/engine/backends/container.go
sflxn: Yes, we use http for this.

lib/apiservers/engine/backends/container.go
sflxn: Fixing.

lib/apiservers/engine/backends/container.go
sflxn: It's intentional.  We can check these options in 543.  I debated on which would be cleaner to look at.  Maybe I'll move it into the code above.

lib/apiservers/engine/backends/container.go
sflxn: I'm not sure that is a good practice.  I agree moving the Add() would be good, but let's assume we don't know when the go routine is schedule, there is a potential for a race condition where we hit the wg.Wait() before the Add().  I'll move the Add() to the ifs below though.

lib/apiservers/portlayer/swagger.yml
sflxn: The casting is between signed ints (Swagger) and unsigned in the portlayer.  As I recall, Swagger doesn't have unsigned.

lib/apiservers/engine/backends/container.go
sflxn: sure.

lib/apiservers/portlayer/swagger.yml
casualjim: have you tried using: `format: uint32` 

Swagger swagger doesnt support uint so it would fall back to int, but platforms (like go-swagger) that do support uint will us the format as type.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
sflxn: done

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
sflxn: Added.

jzt:
So github is being flaky and won't show all files, but I noticed in your branch that the vendor manifest contains merge conflicts:

https://github.com/sflxn/vic/blob/container-attach/vendor/manifest#L173
https://github.com/sflxn/vic/blob/container-attach/vendor/manifest#L594
https://github.com/sflxn/vic/blob/container-attach/vendor/manifest#L658

I found them because I was discussing with @hickeng whether or not we should be vendoring the entire `docker/docker` tree. I'm seeing build problems because of a missing dependency (see https://ci.vmware.run/vmware/vic/1612 for more info) and I'm wondering if it wouldn't be easier to maintain if we just vendored the entire docker repo.

I remember you discussing vendoring docker code - can you refresh my memory as to why we only vendor select files?

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +663,185 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+
+	errors := make(chan error, 3)
+	//FIXME: Swagger will timeout on us.  We need to either have an infinite timeout or the timeout should
+	// start after some inactivity?
+	inContext, inCancel := context.WithTimeout(context.Background(), 2*time.Hour)
+	outContext, outCancel := context.WithTimeout(context.Background(), 2*time.Hour)
+	errContext, errCancel := context.WithTimeout(context.Background(), 2*time.Hour)
+
+	if clStdin != nil {
+		wg.Add(1)
```
fdawg4l: Please move this to line 732.  It is not possible to hit the `Wait()` before the Add inside the go routine because you start the go routine before `Wait` is called.

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +663,185 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+		defer stdinReader.Close()
+
+		transport := httptransport.New(PortLayerServer(), "/", []string{"http"})
+		plClient := client.New(transport, nil)
+		transport.Consumers["application/octet-stream"] = httpkit.ByteStreamConsumer()
+		transport.Producers["application/octet-stream"] = httpkit.ByteStreamProducer()
+
+		// Swagger wants an io.reader so give it the reader pipe.  Also, the swagger call
+		// to set the stdin is synchronous so we need to run in a goroutine
+		go func() {
```
fdawg4l: This doesn't need to be added to the `wg`?

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +663,185 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+				// Shutdown the client's request for stdout, stderr
+				errCancel()
+				outCancel()
+			}
+			wg.Done()
+		}()
+	}
+
+	if clStdout != nil {
+		wg.Add(1)
```
fdawg4l: Same `wg` comment.   Move this to line 776.

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +663,185 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+				return
+			}
+
+			errors <- nil
+			wg.Done()
+		}()
+	}
+
+	if clStderr != nil {
+		wg.Add(1)
```
fdawg4l: Same`wg` comment.  Please Move this to line 811.

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +663,185 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+		transport.Consumers["application/octet-stream"] = httpkit.ByteStreamConsumer()
+		transport.Producers["application/octet-stream"] = httpkit.ByteStreamProducer()
+
+		// Swagger -> pipewriter.  Synchronous, blocking call
+		go func() {
+			getStdoutParams := interaction.NewContainerGetStdoutParamsWithContext(outContext).WithID(name)
+			_, err := plClient.Interaction.ContainerGetStdout(getStdoutParams, clStdout)
+			if clStdin != nil {
+				// Close the client stdin connection (e.g. CLI)
+				clStdin.Close()
```
fdawg4l: This is already done on 721.

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +663,185 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+				log.Errorf(err.Error())
+			}
+			errors <- err
+
+			if stdinOnce && !tty {
+				// Close the stdin connection.  Mimicing Docker's behavior.
+				// FIXME: If we close this stdin connection and request it again,
+				// does the port layer reconnects the stdin?  Is this diff from
+				// current Docker behavior?
+				stdinReader.Close()
```
fdawg4l: This is already done on 722.  I'm actually surprised you aren't hitting a panic in your testing here.

lib/apiservers/engine/backends/container/viccontainer.go
sflxn: Reverted changes back.  I read the article, it refers to fields in a struct not the struct itself.  If I tried to change it to Cache(), it fails to build.  If I remove Get and rename the function to something like ContainerCache(), golint fails.  I now remember why I named it this way in the first place.  I encountered this a few weeks back.

lib/apiservers/engine/backends/container.go
fdawg4l: Looks like it won't panic.  In any case, this returns an error I think.  Can you catch it?

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +663,185 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+
+		transport := httptransport.New(PortLayerServer(), "/", []string{"http"})
+		plClient := client.New(transport, nil)
+		transport.Consumers["application/octet-stream"] = httpkit.ByteStreamConsumer()
+		transport.Producers["application/octet-stream"] = httpkit.ByteStreamProducer()
+
+		// Swagger wants an io.reader so give it the reader pipe.  Also, the swagger call
+		// to set the stdin is synchronous so we need to run in a goroutine
+		go func() {
+			setStdinParams := interaction.NewContainerSetStdinParamsWithContext(inContext).WithID(name)
```
fdawg4l: Can we do this outside the `go` routine?  It doesn't need to be in the go routine does it?

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +663,185 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+	if clStdout != nil {
+		wg.Add(1)
+		transport := httptransport.New(PortLayerServer(), "/", []string{"http"})
+		plClient := client.New(transport, nil)
+		transport.Consumers["application/octet-stream"] = httpkit.ByteStreamConsumer()
+		transport.Producers["application/octet-stream"] = httpkit.ByteStreamProducer()
+
+		// Swagger -> pipewriter.  Synchronous, blocking call
+		go func() {
+			getStdoutParams := interaction.NewContainerGetStdoutParamsWithContext(outContext).WithID(name)
```
fdawg4l: Same go routine comment.  Please take this out of the routine and do this in the parent function.

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +663,185 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+	if clStderr != nil {
+		wg.Add(1)
+		transport := httptransport.New(PortLayerServer(), "/", []string{"http"})
+		plClient := client.New(transport, nil)
+		transport.Consumers["application/octet-stream"] = httpkit.ByteStreamConsumer()
+		transport.Producers["application/octet-stream"] = httpkit.ByteStreamProducer()
+
+		// Swagger -> pipewriter.  Synchronous, blocking call
+		go func() {
+			getStderrParams := interaction.NewContainerGetStderrParamsWithContext(errContext).WithID(name)
```
fdawg4l: Same go routine comment.  Please take this out of the routine and do this in the parent function.

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +663,185 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+			_, err := plClient.Interaction.ContainerGetStderr(getStderrParams, clStderr)
+			if clStdin != nil {
+				// Close the client stdin connection (e.g. CLI)
+				clStdin.Close()
+				inCancel()
+			}
+			if err != nil {
+				if _, ok := err.(*interaction.ContainerGetStderrNotFound); ok {
+					errors <- derr.NewRequestNotFoundError(fmt.Errorf("No such container: %s", name))
+					wg.Done()
```
fdawg4l: Nuke this and add `defer wg.Done()` to line 812.

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +663,185 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+					return
+				}
+
+				// If we get here, most likely something went wrong with the port layer API server
+				errors <- derr.NewErrorWithStatusCode(fmt.Errorf("Unknown error from the interaction port layer: %s", err),
+					http.StatusInternalServerError)
+				wg.Done()
```
fdawg4l: Same `wg` comment.  Use `defer`.

lib/apiservers/engine/backends/container.go
```diff
@@ -506,3 +663,185 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+				return
+			}
+
+			errors <- nil
+			wg.Done()
```
fdawg4l: Same `wg` comment. Use `defer`.

lib/apiservers/engine/backends/container/viccontainer.go
```diff
@@ -0,0 +1,72 @@
...
+
+func NewVicContainer() *VicContainer {
+	return &VicContainer{
+		StreamConfig: runconfig.NewStreamConfig(),
+		Config:       &containertypes.Config{},
+	}
+}
+
+func Cache() *ContainerCache {
+	if cache == nil {
```
fdawg4l: I think I made this comment previously, but it may have gotten lost.  This is not threadsafe.  If more than one routine calls `Cache()`, some or all of them may fall through this `nil` check and create a `ContainerCache`.

lib/apiservers/portlayer/restapi/configure_port_layer.go
```diff
@@ -37,12 +37,12 @@ type handler interface {
 	Configure(api *operations.PortLayerAPI, handlerCtx *handlers.HandlerContext)
 }
 
-var portlayerhandlers = []handler{
-	&handlers.StorageHandlersImpl{},
-	&handlers.MiscHandlersImpl{},
-	&handlers.ScopesHandlersImpl{},
-	&handlers.AttachHandlersImpl{},
-	&handlers.ContainersHandlersImpl{},
+var portlayerhandlers = map[string]handler{
```
fdawg4l: Does this compile?  I think this refers to previous revisions.  You may want to revert this back to a list.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -0,0 +1,240 @@
...
+	interactionSession, err = session.NewSession(sessionconfig).Create(ctx)
+	if err != nil {
+		log.Fatalf("InteractionHandler ERROR: %s", err)
+	}
+
+	i.attachServer = attach.NewAttachServer("", 0)
+
+	if err := i.attachServer.Start(); err != nil {
+		log.Fatalf("Attach server unable to start: %s", err)
+		return
```
fdawg4l: `Fatal` does not need a return.

lib/apiservers/engine/backends/container.go
sflxn: Can we do this at a later date?  Our copyright script doesn't have an ignore function.  I separated the file, but the build fails on copyright check.  I don't want to add vmware's copyright to this file, and I don't want to make the changes to the copyright script as part of this PR.

lib/apiservers/portlayer/swagger.yml
sflxn: So, I would like to keep this yml as close to the official swagger spec as possible, even if go-swagger support something it doesn't.  Casting isn't really a big deal.

sflxn:
@jzt: It's a merge problem.  I'll fix it.  One of the intent of this PR is to vendor the entire docker repo and others from docker.  Git rebase always messes up with the manifest file, and I usually have to go hand modify it.

caglar10ur:
@jzt I think I'm a fan of vendoring entire docker/docker repo

sflxn:
@caglar10ur @jzt Yeah, I am too.  Vendoring individual folder can mask hidden bugs.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
sflxn: Yeah.  I blindly copied that from your original handler.  Removed.

lib/apiservers/engine/backends/container/viccontainer.go
sflxn: Moved creation to an init().

lib/apiservers/portlayer/restapi/configure_port_layer.go
sflxn: Sure, it compiles.  It's how you initialize a golang map.  You're right.  It was used earlier when we were using a custom middleware and needed the map.  Reverting.

lib/apiservers/engine/backends/container.go
sflxn: I'll log the error, but at this point there's nothing else we can do.

lib/apiservers/engine/backends/container.go
sflxn: really?  Isn't that a little pedantic?  Besides, I like to keep code usage together.  Unless you have a really good reason,...

fdawg4l: It is really minor.  It's up to you.  I like to keep my anonymous functions small but its really not important.

lib/apiservers/engine/backends/container.go
fdawg4l: The error is important, but I was more concerned about the double `Close()` of the same reader.

fdawg4l:
@sflxn Great work!  It was a long road but this, I think, is much less complicated than using Docker's stream mux'ing code.  Everything looks good from my perspective.  Just niggles that come down to style in the end, but nothing major (other than that defer `Close` which you may or may not have already addressed.  I wish there were tests here, however.  We have a test framework in the attach server as well as the tether which you can adapt to do an end to end test using your own (or a generated) client vs the docker client.  But, we can get to that in a different issue. It *is* possible, to have a unit test for this and it's complex enough to warrant one.  But, again, we can get to it later.

Great work!

LGTM

fdawg4l:
@hickeng when you get the chance, can you please look at this change?  Thanks!

sflxn:
We're going to have docker integration tests as soon as we get docker ps complete.  Docker already have a suite of tests to check our compliance, but every single test relies on docker ps working.

lib/apiservers/engine/backends/container.go
caglar10ur: Not really, I briefly looked at what docker does. They do ad-hoc error reporting without using any libraries or pkgs so I ended up doing something similar in imagec. 

I think error reporting should be an epic that needs to be worked on.

caglar10ur: And I agree with @fdawg4l we should use derr - in the name of consistency

lib/apiservers/engine/backends/container.go
```diff
@@ -526,3 +683,236 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+// keys are the keys that are used for detaching streams without closing them
+func attachStreams(name string, tty, stdinOnce bool, clStdin io.ReadCloser, clStdout, clStderr io.Writer, keys []byte) error {
+	var wg sync.WaitGroup
+
+	errors := make(chan error, 3)
+	//FIXME: Swagger will timeout on us.  We need to either have an infinite timeout or the timeout should
+	// start after some inactivity?
+	inContext, inCancel := context.WithTimeout(context.Background(), 2*time.Hour)
+	outContext, outCancel := context.WithTimeout(context.Background(), 2*time.Hour)
+	errContext, errCancel := context.WithTimeout(context.Background(), 2*time.Hour)
```
caglar10ur: can we make those const so that we can change them in single place when needed?

lib/apiservers/engine/backends/container.go
```diff
@@ -526,3 +683,236 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+				return
+			}
+
+			errors <- nil
+		}()
+	}
+
+	// Wait for all stream copy to exit
+	wg.Wait()
+	log.Printf("Attach stream closed")
```
caglar10ur: Debugf?

caglar10ur:
Great work @sflxn ! It LGTM but would like following 3 fixed before merge;

- https://github.com/vmware/vic/pull/681/commits/45f27d198cb2592bcf6da551f0074f33ae6912c9#r66112576
- https://github.com/vmware/vic/pull/681/commits/45f27d198cb2592bcf6da551f0074f33ae6912c9#r66113156
- https://github.com/vmware/vic/pull/681/commits/45f27d198cb2592bcf6da551f0074f33ae6912c9#r65945039

fdawg4l:
@sflxn those are "integration tests".  I would like to see a unit test.  See `/cmd/tether/attach_test.go` and `lib/portlayer/attach/server_test.go` for examples.

lib/apiservers/engine/backends/container.go
sflxn: I just looked at that.  The code was broken out to getImageMetadataFromStoragePL(), which returns derr.

sflxn:
@caglar10ur Took care of 1&2.  3rd issue isn't an issue.  I added init() back in to resolve another review issue.

caglar10ur:
LGTM :sheep: :it:

Issue #682, opened 2016-05-15
Integration test often fails connecting to VCH


dougm:
AFAICT, the installer does not wait for VCH docker listener on port 2376 to come up.  Nor does the integration test, so we see often these failures:

```
not ok 18 vic-machine deployment and docker commands
# (from function `assert_success' in file ./helpers/bats-assert/src/assert.bash, line 114,
#  in test file ./vic-machine.bats, line 128)
#   `assert_success' failed
# <http://ci:8080/job/vic-integration-tests/ws/tests/../bin/vic-machine> -target 10.17.109.150 -name VCH-18804 -user root -compute-resource /ha-datacenter/host/localhost.eng.vmware.com/Resources -image-store datastore
# INFO[2016-05-15T10:25:26Z] docker -H 10.17.109.128:2376 --tls --tlscert='./VCH-18804-cert.pem' --tlskey='./VCH-18804-key.pem' info
# INFO[2016-05-15T10:25:26Z] Installer completed successfully...
# docker_cmd= -H 10.17.109.128:2376 --tls --tlscert='./VCH-18804-cert.pem' --tlskey='./VCH-18804-key.pem'
# Containers: 0
# <http://ci:8080/job/vic-integration-tests/ws/tests/../bin/vic-machine> -target 10.17.109.150 -name VCH-18804 -user root -compute-resource /ha-datacenter/host/localhost.eng.vmware.com/Resources -image-store datastore -force -key ./VCH-18804-key.pem -cert ./VCH-18804-cert.pem
# INFO[2016-05-15T10:25:58Z] Installer completed successfully...
# 
# -- command failed --
# status : 1
# output : Cannot connect to the Docker daemon. Is the docker daemon running on this host?
# --
# 
# Clear VCH deployments if exists
Makefile:178: recipe for target 'integration-tests' failed
make: *** [integration-tests] Error 1
Build step 'Execute shell' marked build as failure
```

dougm:
I'm going to fix within the test for now.  Doing some refactoring of the vic-machine.bats test while I'm in there.

dougm:
I'll finish up later today, refactoring first: https://github.com/vmware/vic/compare/master...dougm:bats

PR #683, opened 2016-05-16
Refactor vic-machine tests

dougm:
- Add retry for docker info, as listener may not be up yet post-install

- Remove container vm datastore directory

- More use of bats assertions

- Simplify govc usage

- Make use of GOVC_* variables

- Fix shellcheck warnings/errors

Fixes issue #682

### DIFF
--- a/tests/vic-machine.bats
+++ b/tests/vic-machine.bats
@@ -1,3 +1,4 @@
+#!/usr/bin/env bats
 # Copyright 2016 VMware, Inc. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


fdawg4l:
LGTM

tests/vic-machine.bats
```diff
@@ -12,135 +13,126 @@
...
-    govc pool.destroy "${POOL_PATH}/${vch_name}" || echo "no need to delete resource pool"
-    govc datastore.rm -ds ${IMAGE_STORE_NAME} "${vch_name}" || echo "no need to delete datastore files"
-    govc datastore.rm -ds ${IMAGE_STORE_NAME} "VIC" || echo "no need to delete datastore file VIC"
-    govc host.vswitch.remove "${vch_name}" || echo "no need to remove vswitch"
-    rm *.pem || echo "no pem file found"
-    rm *.log || echo "no log file found"
+    if [ -n "$(govc vm.info $vch_name)" ] ; then
+        govc vm.destroy $vch_name
+        govc pool.destroy "$GOVC_RESOURCE_POOL/$vch_name"
+        govc datastore.rm -f $vch_name
```
hickeng: I think this is broken on vSAN as the name is not the definitive location.

hickeng:
LGTM

tests/vic-machine.bats
```diff
@@ -12,135 +13,126 @@
...
+    if [ -z "$GOVC_RESOURCE_POOL" ] ; then
+        GOVC_RESOURCE_POOL=$(govc ls /*/host/*/Resources | head -1)
+    fi
+
+    if [ -z "$GOVC_DATASTORE" ] ; then
+        GOVC_DATASTORE=$(govc ls /*/datastore | head -1 | xargs basename)
     fi
+
+    # split GOVC_URL into GOVC_{URL,USERNAME,PASSWORD}
+    eval "$(govc env | sed -e 's/\$/\\$/g')"
```
emlin: I got govc usage error while execute "govc env", is there any change in govc recently?

tests/vic-machine.bats
```diff
@@ -12,135 +13,126 @@
...
-    govc vm.destroy "${vch_name}" || echo "no need to delete VCH"
-    govc pool.destroy "${POOL_PATH}/${vch_name}" || echo "no need to delete resource pool"
-    govc datastore.rm -ds ${IMAGE_STORE_NAME} "${vch_name}" || echo "no need to delete datastore files"
-    govc datastore.rm -ds ${IMAGE_STORE_NAME} "VIC" || echo "no need to delete datastore file VIC"
-    govc host.vswitch.remove "${vch_name}" || echo "no need to remove vswitch"
-    rm *.pem || echo "no pem file found"
-    rm *.log || echo "no log file found"
+    if [ -n "$(govc vm.info $vch_name)" ] ; then
+        govc vm.destroy $vch_name
+        govc pool.destroy "$GOVC_RESOURCE_POOL/$vch_name"
```
emlin: The resource pool might technically exist, even if the vch vm is not created. Maybe we can move this line out of the if condition?

tests/vic-machine.bats
```diff
@@ -12,135 +13,126 @@
...
+    run "$installer"
     assert_failure
-    echo ${lines[0]}
-    [[ ${lines[0]} =~ "-target argument must be specified" ]]
+    assert_line -e "-target argument must be specified"
 }
 
 @test "vic-machine user is missing" {
-    run "${installer}" -target ${TEST_URL}
+    run "$installer" -target "$GOVC_URL"
```
emlin: I didn't successfully run your command in setup, so not sure if GOVC_URL contains user name or not here, but if yes, it's not as expected, the target should contain the VC/ESXi ip or hostname only.

tests/vic-machine.bats
```diff
@@ -12,135 +13,126 @@
...
+      # retry as listener may not be up yet
+      sleep 1
+    done
+
     assert_success
-    len=${#lines[@]}
-    echo ${lines[len-1]}
-    [[ ${lines[len-1]} =~ "Installer completed successfully..." ]]
-    docker_cmd=`echo ${lines[len-2]} | awk '{$1=$2=""; $(NF--)=""; print$0}'`
+
```
emlin: Looks like you removed the second installation in this test case. I intentionally rerun the installation, to make sure -force and -generate-cert=false work here.

emlin:
Thanks for the enhancement. This looks much more cleaner. I'm not used to code with script.
Anyway, a few comments, please have a look.

tests/vic-machine.bats
dougm: I've been meaning to add a switch to govc vm.destroy to remove the vm folder too (directory in VirtualMachine.Config.Files.VmPathName), will do that in any case.

tests/vic-machine.bats
dougm: Yes, please watch the vic PRs for govmomi updates and/or the govmomi project itself.  Looks like the problem is a bug here: https://github.com/vmware/vic/blob/master/tests/run_test.sh#L82

This does not install from our vendored sources, nor does it update.  So you're currently pinned to the govc version when you ran the script, unless you rm $GOPATH/bin/govc
I'll open an issue in a bit, we shouldn't be doing any provisioning in that script anyhow.

tests/vic-machine.bats
dougm: So it is possible for the installer to create the resource pool, but not remove it if vm create fails?  Sounds like an installer issue.

tests/vic-machine.bats
dougm: This is what 'govc env' does, so GOVC_URL will not contain credentials at this point.

tests/vic-machine.bats
dougm: The second installation is still there, at the end:
```
    # test reinstall with same name
    params+=(-force -key ./${vch_name}-key.pem -cert ./${vch_name}-cert.pem)

    run "$installer" "${params[@]}"
    assert_success
    assert_line -e "Installer completed successfully"
```

tests/vic-machine.bats
emlin: Do you mean to remove the resource pool if VCH VM creation failed?
Cause we don't throw error if VCH resource pool already exists, if VM creation failed, I think we should not remove resource pool as well.
But in test env, it's good to delete resource pool anyway.

tests/vic-machine.bats
emlin: great

emlin:
LGTM

tests/vic-machine.bats
emlin: My understanding is that run_test will install any dependency, if it does not exist. Otherwise, we cannot run test through make integration-tests.
But in integration test, we can prepare environment before we run make integration-tests, and then the script will simply detect if the dependencies are already there.

For the issues, it's my own env, not the test.

dougm: I'll fix in #687

tests/vic-machine.bats
dougm: Not sure I follow.  We are removing a pool with the $vch_name (VCH-$RANDOM), that pool would have been created by the installer.

emlin: Installer is leaving resource there if VCH creation failed. But what I mean is that we might leave the resource pool not cleared in some edge case.
Never mind, it's a small issue.

Issue #672, opened 2016-05-13
Capture log output on test failure in Jenkins or new Drone


gigawhitlocks:
The integration tests need to be modified to capture the log output from the vicadmin server after execution but before the VCH is destroyed in order to provide context when the tests fail. Modifications will be made in addition to artifacting that output (a tarball) in Jenkins on test failure.

Additionally this will be exercising our log collection facilities which, to quote George, "has its own value"

mdubya66:
fyi, fabio is in austin today and we are looking at moving back to drone for Integration tests and to leverage the HaaS infrastructure in the new DC.

fdawg4l:
Do we need to use drone in order to use HaaS?  Can we put jenkins on HaaS and make it public?

dougm:
Let's not get into Jenkins vs Drone here.  The main change we need from @gigawhitlocks here is for the integration test to capture the VCH logs.  This will both test some of the vicadmin code paths and provide debugging info when running the tests locally, independent of CI.  The CI portion will just archive these logs as job artifacts.

gigawhitlocks:
Yep @dougm has it -- biggest change is actually capturing the logs inside of tests. Whether we use Drone or Jenkins in the long run, it doesn't really matter as far as this story is concerned.

gigawhitlocks:
Changes needed to test are done but changes will be made to Drone after https://github.com/vmware/vic/pull/707 is merged, rather than spending any more time on Jankins

gigawhitlocks:
Placing back in TODO until #707 is merged since #698 is merged so the last part of this will be configuring Drone

hickeng:
@gigawhitlocks Can this be closed now #752 is closed?

PR #698, opened 2016-05-17
Exercise log collection in vic-machine tests

gigawhitlocks:
This adds a log collection step to grab install logs out of the VCH before it is destroyed in the vic-machine tests.

Once this is committed, Jenkins will be configured (by me) to save these artifacts off if there is a test failure. Tests will also fail if for some reason `wget` fails to get logs.

Addresses #672

### DIFF
--- a/tests/vic-machine.bats
+++ b/tests/vic-machine.bats
@@ -41,6 +41,11 @@ setup () {
 }
 
 teardown() {
+    if [ -n "$installer_output" ] ; then
+        run get_logs "$installer_output"
+        installer_output=""
+    fi
+
     if [ -n "$(govc vm.info $vch_name)" ] ; then
         govc vm.destroy $vch_name
         govc pool.destroy "$GOVC_RESOURCE_POOL/$vch_name"
@@ -52,6 +57,14 @@ teardown() {
     fi
 
     rm -f ./$vch_name-*.pem
+}
+
+# pass this the installer output & it will download the logs from the vicadmin server
+get_logs () {
+    vicadmin_url=$(grep -n1 'Log server' <<<"$1" | grep http | cut -d' ' -f2)
+    time="$(date -Iseconds)"
+    wget $vicadmin_url/container-logs.tar.gz --no-check-certificate -O install-container-logs-"$time".tar.gz
+    wget $vicadmin_url/logs.tar.gz --no-check-certificate -O install-logs-"$time".tar.gz
 }
 
 @test "vic-machine usage" {
@@ -89,6 +102,7 @@ teardown() {
 
     local output status
     run "$installer" "${params[@]}"
+    installer_output="$output" # save this before checking for success
     assert_success
     assert_line -e "Installer completed successfully"
     docker_cmd=($(grep 'docker -H' <<<"$output" | cut -d' ' -f3-7))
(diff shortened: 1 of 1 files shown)


tests/vic-machine.bats
```diff
@@ -125,6 +126,10 @@ teardown() {
     run "$docker" "${docker_cmd[@]}" start "$name"
     assert_success
 
+    wget "$vicadmin_ip"/container-logs.tar.gz --no-check-certificate -O install-container-logs.tar.gz
+    wget "$vicadmin_ip"/logs.tar.gz --no-check-certificate -O install-logs.tar.gz
+    assert_success
```
dougm: You need to use 'run' to set '$status' before 'assert_success'.  And you need an 'assert_success' for each 'run wget ...'.

https://github.com/sstephenson/bats#run-test-other-commands

https://github.com/ztombol/bats-assert#assert_success

gigawhitlocks: Ah, I see. I'll update this. I assume that as it is, the assert_success just does nothing? Would've expected them to fail, heh

dougm:
lgtm

tests/vic-machine.bats
```diff
@@ -123,6 +124,12 @@ teardown() {
     run "$docker" "${docker_cmd[@]}" start "$name"
     assert_success
 
+    run wget "$vicadmin_ip"/container-logs.tar.gz --no-check-certificate -O install-container-logs.tar.gz
```
dougm: Realizing you'll probably need to capture logs in the teardown() instead.  Otherwise if a prior assertion fails, we won't reach this point and no logs are captured for debugging.

gigawhitlocks: Hmm, you're right. I'll have to break this test into two tests, then, since the first VM is destroyed in the middle of the test. Really those vm destruction steps should be in teardown too..

gigawhitlocks:
@dougm I've addressed your concerns
@emlin can you take a look at this too? Need two LGTMs anyway and you two are the most familiar with these tests

tests/vic-machine.bats
```diff
@@ -123,16 +137,21 @@ teardown() {
...
     assert_success
 
+    # if we're here we should grab the logs cause we're gonna wipe out the VCH
+    get_logs "$installer_output"
+
     # test reinstall with same name
     params+=(-force -key ./${vch_name}-key.pem -cert ./${vch_name}-cert.pem)
 
     run "$installer" "${params[@]}"
+    installer_output="$output"
```
hmahmood: Don't understand why you have this line

emlin:
LGTM

tests/vic-machine.bats
gigawhitlocks: Saves off the installer output so that the correct logs are grabbed when teardown is called. Notice that it happens before assert_success so that the logs will be grabbed regardless of failure/success of the test. Without it, the first installation's logs would be grabbed by teardown instead. I do realize that it does not appear straightforward, as I'm using global state to pass information so the flow of the test is not terribly obvious.. wish there was a better way but given the constraints of testing with Bash I don't think there is

dougm:
lgtm

andrewtchin:
lgtm again

PR #704, opened 2016-05-17
Startblock

### LINKED ISSUE #426: start

hickeng:
Blocks until the container process has been launched. This will either return:
* nil: no error - process launched successfully
* err: process failed to launch - err contains the error message from the tether.

Fixes #426

### DIFF
--- a/portlayer/exec/container.go
+++ b/portlayer/exec/container.go
@@ -18,12 +18,13 @@ import (
 	"fmt"
 	"math/rand"
 	"sync"
+	"time"
 
-	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/session"
 	"github.com/vmware/vic/pkg/vsphere/tasks"
+	"github.com/vmware/vic/pkg/vsphere/vm"
 	"golang.org/x/net/context"
 )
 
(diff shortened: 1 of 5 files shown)


portlayer/exec/container.go
```diff
@@ -135,5 +139,36 @@ func (c *Container) Start(ctx context.Context) error {
...
+			}
+		}
+		return false
+	}
+
+	// Wait some before giving up...
+	ctx, cancel := context.WithTimeout(ctx, propertyCollectorTimeout)
+	defer cancel()
+
+	c.vm.WaitForExtraConfig(ctx, waitFunc)
```
dougm: Looks like you should still check err here, in case property.Wait() fails:
```go
if err != nil {
   return err
}
```

fdawg4l:
Is it possible to add a test?  I don't think it would be trivial if at all but thought it worth asking.

hickeng:
@fdawg4l not until @dougm has the sim up and running - it depends on property collector. I have added a test for the field that we block on into the tether tests, specifically in this case for the missing binary test.

dougm:
lgtm

fdawg4l:
lgtm

hickeng:
LGTM

PR #712, opened 2016-05-18
Path resolution for tether binaries

### LINKED ISSUE #708: Add path resolution
Currently we require fully qualified paths that will run in the container - we need to add PATH based lookup.

The test already exists and simply needs enabling, see cmd/tether/cmd_test.go:TestRelativePath.

hickeng:
This adds path resolution for tether binaries. Also reshuffles where
logging and launch status reporting happens to reduce the time the pid
mutex is held for.

Fixes #708

### DIFF
--- a/cmd/tether/tether.go
+++ b/cmd/tether/tether.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"os/signal"
 	"syscall"
 
@@ -215,10 +216,17 @@ func handleSessionExit(session *SessionConfig) error {
 func launch(session *SessionConfig) error {
 	defer trace.End(trace.Begin("launching session " + session.ID))
 
+	// encode the result whether success or error
+	defer func() {
+		extraconfig.EncodeWithPrefix(dataSink, session.Started, fmt.Sprintf("guestinfo..sessions|%s.started", session.ID))
+	}()
+
 	logwriter, err := utils.sessionLogWriter()
 	if err != nil {
 		detail := fmt.Sprintf("failed to get log writer for session: %s", err)
 		log.Error(detail)
+		session.Started = detail
+
 		return errors.New(detail)
 	}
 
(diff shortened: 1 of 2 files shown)


cmd/tether/tether.go
```diff
@@ -215,10 +216,17 @@ func handleSessionExit(session *SessionConfig) error {
 func launch(session *SessionConfig) error {
 	defer trace.End(trace.Begin("launching session " + session.ID))
 
+	// encode the result whether success or error
+	defer func() {
```
dougm: No need for the extra func:
```
defer extraconfig.EncodeWithPrefix(...)
```

cmd/tether/tether.go
```diff
@@ -233,9 +241,20 @@ func launch(session *SessionConfig) error {
 	session.Cmd.Stderr = session.errwriter
 	session.Cmd.Stdin = session.reader
 
+	resolved, err := exec.LookPath(session.Cmd.Path)
+	if err != nil {
+		pretty := fmt.Errorf("%s: no such file or directory", session.Cmd.Path)
+		log.Errorf("Path lookup failed for %s: %s", session.Cmd.Path, err)
+		session.Started = pretty.Error()
+		return err
```
dougm: Did you want to return `pretty` here instead of `err`?  Otherwise, pretty could just be a string (fmt.Sprintf intead of fmt.Errorf)

dougm:
lgtm

cmd/tether/tether.go
hickeng: If you do that then the session.Started value on entry to Launch is used and not it's value on exit. Made this mistaken initially ;)

cmd/tether/tether.go
hickeng: I did mean to return pretty given the full error is already logged.

cmd/tether/tether.go
hmahmood: Does this code still run on error?

hmahmood: Never mind ... I guess it has to.

hmahmood:
lgtm

PR #720, opened 2016-05-18
vic-machine usage examples

hickeng:
This adds some examples of vic-machine CLI usage. The logic around specifying FROM/TO with manifest files is still very much WIP and this doesn't add *-ipam arguments for the appliance networks (client, et al).

[skip-ci]

Fixes # 547


doc/design/vic-machine.md
```diff
@@ -63,6 +63,111 @@ To create a VCH, disambiguation of which resource to use from the available set
...
+These are various examples intended to illustrate how vic-machine is intended to be used. These are illustrations rather than the finalized naming so if something doesn't make sense ask about it. 
+
+### Fully specified command line
+
+```
+vic-machine create
+-TO=https://root@****:vcenter/
+-name=edw-web
+-compute-resource=/example-datacenter/host/mycluster/Resources/edw-pool/web
+-image-store=ds://datastore/containers/edw
```
emlin: What does this format mean? Is the containers/edw the directory in datastore? How does this work in vSAN?

doc/design/vic-machine.md
```diff
@@ -63,6 +63,111 @@ To create a VCH, disambiguation of which resource to use from the available set
...
+-container-store=ds://datastore/edw/edw/web
+-client-network="VM Network"
+-management-network="management-net"
+-volume-store=ds://edw-san/reports:reports
+-volume-store=ds://edw-san/data:data
+-docker-network=private-vlan12d7f2:bridge
+-docker-network=edw-corp-net:backend
+-docker-network-ipam=edw-corp-net:10.118.78.128-192/24,10.118.78.1
+-container=oracledb.edw.corp.net:db.backend
```
emlin: So the container created in db.backend network, should have domain oracledb.edw.corp.net?

doc/design/vic-machine.md
```diff
@@ -63,6 +63,111 @@ To create a VCH, disambiguation of which resource to use from the available set
...
+* -docker-network-ipam - this is an optional argument furnishing additional information for controlling IP address management on the network, in the form `ipaddress-range/mask,gateway`. If not specified, DHCP is used.
+* -container - this allows containers to address a specific FQDN as if it were itself a container managed by the VCH, specified in the form of `FQDN:alias.network`.  
+
+
+### Manifest file - create
+Note that the various resource paths end in a `/` - this signifies that it's acceptable for additional refinement to be provided when using this manifest file:
+
+```
+vic-machine create
+-TO=file::///home/joe/vch/edw.manifest
```
emlin: Here is to reuse the parameter -TO, to create manifest. Would it be more clear to have a new command? For example vic-machine manifest create ...

In addition, are we using different URL format for the file?

doc/design/vic-machine.md
```diff
@@ -63,6 +63,111 @@ To create a VCH, disambiguation of which resource to use from the available set
...
+-container-store=web
+-volume-store=reports/web:report
+-docker-network=bridge
+-name=web-team-private
+```
+
+The exact consumption mechanic of manifests is still quite uncertain. The example above is working on the premise that all mappings in the manifest are optional and are bound only if referenced, whether simply by label or as part of a refinement. The example above does not reference the _frontend_ network, so no containers can be attached to it.
+
+### List existing VCHs
+List the existing VCHs under the specified compute resource. The example below will list all VCHs in the target vCenter.
```
emlin: What is supposed to be shown during VCH listing, or inspection

doc/design/vic-machine.md
hickeng: You can create top level directories in vSAN using the DatastoreNamespaceManager.CreateDir call - those are then embedded vmfs blobs that can have subdirectories.

doc/design/vic-machine.md
hickeng: I've been unsure about this myself. I've been trying to figure out a way to maintain the symmetry of to/from so that it's possible to, for example, extract the config of an existing VCH and save it to a manifest, or the opposite.

I'm not happy with the current overloading, but am also reluctant to add yet another argument or command as we will rapidly lose the ease of use we're after. The best thought I've had so far is a simplification of your suggestion here, simply adding `manifest` as a new command to differentiation deploying of VCH from creating of manifest:
`vic-machine manifest ...`

doc/design/vic-machine.md
hickeng: Still thinking through this. My inclination is towards basic informational listing:
```
MBUsed MHzUsed adminEmail /inventory/path/to/vch
```

This should probably go to UX to find out what viadmin's actually want to know about systems in this fashion.

doc/design/vic-machine.md
```diff
@@ -73,7 +73,8 @@ These are various examples intended to illustrate how vic-machine is intended to
 vic-machine create
 -TO=https://root@****:vcenter/
```
emlin: Just find that this target URL is different with what we're using with govc. Is that by design?
I'm just wondering if we're introducing too many resource format to user. Cause user will need govc to figure out the resource pool format.

hickeng:
@corrieb @dougm Please take a look

emlin:
LGTM

hickeng:
LGTM

This being still in PR is meaning it's difficult to point people at the doc so pushing it through after 30 days.

hickeng:
LGTM

This being still in PR is meaning it's difficult to point people at the doc so pushing it through after 30 days.

Issue #665, opened 2016-05-12
Refactor vic-machine: use metadata/VirtualContainerHostConfigSpec


emlin:
replace existing install/configuration package

hickeng:
The VirtualContainerHostConfigSpec is WIP and was used to reason about what aspects are common between the appliance and containerVMs. As such it needs some changes to be directly useful for us:

1. remove ComponentConfig - it duplicates Sessions in the embedded ExecutorConfig
2. change `BridgeNetwork object.Network` to `DefaultNetwork *object.Network` - that allows flexibility in the default network and also permits "none" as a choice
3. Change the type of ContainerNetworks to something like the following so we can express subnet fragments. Nil values for first/last can be used to indicate DHCP - no interest at this time in providing richer DHCP config than on/off:
```
type PartialNetwork struct {
   // network configuration
    ContainerNetwork `vic:"0.1" scope:"read-only" key:"netconfig"`

   // first address of the available IP range
   First *net.IP `vic:"0.1" scope:"read-only" key:"first"`
   // last address of the available IP range
   Last net.IP `vic:"0.1" scope:"read-only" key:"last"`
}
```

hickeng:
I ended up doing the remainder of #665, #686 and #666 in PR #1127 for #746. Updating the estimate appropriately.

PR #721, opened 2016-05-18
Fix error caused by NetworkReference type cast

emlin:
There is NetworkReference cast to Network, but it's wrong in dvSwitch case.
To see if #665 can be fixed by it.


dougm:
lgtm

fdawg4l:
:sheep: :it:

PR #724, opened 2016-05-18
relative path support

### LINKED ISSUE #708: Add path resolution

hickeng:
Use cmd.Env path for lookup as this needs to use the PATH provided for the container, not the system environment inherited by tether.

Fixes #708

### DIFF
--- a/cmd/tether/tether_linux.go
+++ b/cmd/tether/tether_linux.go
@@ -24,6 +24,7 @@ import (
 	_ "net/http/pprof"
 	"os"
 	"os/signal"
+	"path/filepath"
 	"strings"
 	"syscall"
 	"time"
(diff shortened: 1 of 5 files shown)


dougm:
lgtm

cmd/tether/tether_linux.go
```diff
@@ -228,6 +229,55 @@ func (t *osopsLinux) processEnvOS(env []string) []string {
...
+			return file, nil
+		}
+		return "", err
+	}
+
+	// extract path from the env
+	var pathenv string
+	for _, value := range env {
+		if strings.HasPrefix(value, "PATH=") {
+			pathenv = value
```
fdawg4l: break

fdawg4l:
lgtm

hickeng:
LGTM

PR #730, opened 2016-05-19
Refactor vic-machine to use extraconfig and metadata packages

### LINKED ISSUE #665: Refactor vic-machine: use metadata/VirtualContainerHostConfigSpec

emlin:
Fixes #665


cmd/vic-machine/validator.go
```diff
@@ -0,0 +1,210 @@
...
+	sessionconfig := &session.Config{
+		Service:        v.TargetPath,
+		Insecure:       input.insecure,
+		DatacenterPath: v.DatacenterName,
+		ClusterPath:    v.ClusterPath,
+		DatastorePath:  v.ImageStorePath,
+		PoolPath:       v.ResourcePoolPath,
+	}
+
+	v.Session, err = session.NewSession(sessionconfig).Create(v.Context)
```
hickeng: I suggest ditching session entirely - it was written as a temporary convenience while we were getting up and running and doesn't handle multiple networks or datastores well.
In bonneville having the defaults in the case of only one network/datastore/etc was convenient, but it's also a degenerative scenario for a production deployment. If need be we can add selection/recommendation of defaults in at a later date. For now we're better off requiring all options that map to implemented features be specified.


You can use SearchIndex.FindByInventoryPath to go straight from the ComputeResource path to the moref.

You cannot do so for the various datastore paths as the portion of the path after the datastore name is not part of the vsphere inventory. The host part of the datastore URL is the datastore name.

cmd/vic-machine/validator.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	vchConfig.ComputeResources = append(vchConfig.ComputeResources, v.Session.Pool)
+	vchConfig.ImageStores = append(vchConfig.ImageStores, v.ImageStorePath)
+	//TODO: Add more configuration validation
+	return nil
+}
+
+func (v *Validator) getNetworkPath(net object.NetworkReference) (string, string, error) {
+	switch t := net.(type) {
+	case *object.DistributedVirtualPortgroup:
```
hickeng: Use `t` instead of `net` inside the case statements. `t` is already of the appropriate type as per the type switch semantics:
https://golang.org/doc/effective_go.html#type_switch

```
switch t := net.(type) {
   case *object.DistributedVirtualPortgroup:		
      return t.InventoryPath, t.Name(), nil
   case *object.Network:
      return t.InventoryPath, t.Name(), nil
   default:
      return "", "", errors.Errorf("Unknown network card type: %s", reflect.TypeOf(t))
}
```

You can also add a test for the raw dvswitch types - I think this is likely to be a common enough mistake that it warrants a custom error message:
DistributedVirtualSwitch
VmwareDistributedVirtualSwitch

cmd/vic-machine/validator.go
emlin: After discussion with George, will keep as it is right now, cause the objects in Session is leveraged in both validator and dispatcher.
We'll revisit it after govmomi gateway is ready.

cmd/vic-machine/validator.go
emlin: Didn't know that. Let me try it

metadata/virtual_container_host.go
```diff
@@ -36,13 +37,18 @@ type VirtualContainerHostConfigSpec struct {
 	// The base config for the appliance. This includes the networks that are to be attached
 	// and disks to be mounted.
 	// Networks are keyed by interface name
-	ExecutorConfig
-
-	// The set of components to launch
-	ComponentConfig []SessionConfig
-
+	ExecutorConfig `vic:"0.1" scope:"read-only" recurse:"depth=0"`
```
gigawhitlocks: Is this `0.1` appearing throughout referring to the VIC version? Or is it the metadata schema version? At the least we should probably make this value a `const`, right?

emlin: Good question. But looks like they are using this version all of the world. @hickeng for comments on it.

install/management/appliance.go
```diff
@@ -348,12 +330,19 @@ func (d *Dispatcher) setMacToGuestInfo() error {
 			err = errors.Errorf("Timeout to get VM MAC address")
 			return err
 		}
-		netName := d.nics[key]
-		spec.ExtraConfig = append(spec.ExtraConfig, &types.OptionValue{Key: fmt.Sprintf("guestinfo.vch/networks/%s/portgroup", netName), Value: key})
-		spec.ExtraConfig = append(spec.ExtraConfig, &types.OptionValue{Key: fmt.Sprintf("guestinfo.vch/networks/%s/mac", netName), Value: value})
-		keys = append(keys, netName)
+		for nicName, netInfo := range conf.Networks {
```
hickeng: This entire nested for loop construct can be replaced with the following - you should never need nested loops like this to discover if a key is in a hashmap:
```
	for name, info := range conf.Networks {
		value, ok := m[info.PortGroupName]
		if !ok || value == "" {
			// timeout to wait MAC address, so empty mac address is returned
			return errors.Errorf("Timeout getting MAC address for %s on %s", name, info.PortGroupName)
		}

		info.Mac = value
	}
```

I've left out the extraconfig manipulation from that loop for clarity.

metadata/virtual_container_host.go
hickeng: This is an annotation rather than compiled code - I don't think you can use a constant here, but I'll confess I've never tried it.... and brief experimentation shows that consts are not interpreted in annotations:
https://play.golang.org/p/7l8EWR2XuJ

It's also worth noting that while all the versions are the same currently they will not increment globally - it's tied to the semantic meaning of each field.

install/management/appliance.go
emlin: @hickeng the map conf.Networks is one mapping between nic name and networkinfo, but mac address mapping is between network name and mac address. 
The key is different, that's why I loop the map and compare there key with following sentence:
		for nicName, netInfo := range conf.Networks {
			if netInfo.PortGroupName == netName {
You can see that's not the key comparison.

But you do find an issue for the mac address waiting. I assume we should wait until mac address of all nics are ready, instead of any of them is there. So I changed the WaitForMAC method.

hickeng: The key for the maps is different, yes, but the key for `m` is part of the value for `Networks` (assuming this is the comparison you actually want to make - you say netName in the reply above):
````
if netInfo.PortGroupName == key
```
See https://play.golang.org/p/ks6tpd7vdL

install/management/appliance.go
```diff
@@ -301,18 +279,22 @@ func (d *Dispatcher) createAppliance(conf *configuration.Configuration) error {
 			Value: fmt.Sprintf("-serveraddr=0.0.0.0 -port=%s -port-layer-port=8080 %s", d.DockerPort, d.dockertlsargs)})
```
hmahmood: Shouldn't all of these also be in the vch config, so that they can be serialized with the extraconfig package?

install/management/appliance.go
```diff
@@ -301,18 +279,22 @@ func (d *Dispatcher) createAppliance(conf *configuration.Configuration) error {
...
 			Value: fmt.Sprintf("-serveraddr=0.0.0.0 -port=%s -port-layer-port=8080", d.DockerPort)})
 		spec.ExtraConfig = append(spec.ExtraConfig, &types.OptionValue{Key: "guestinfo.vch/sbin/vicadmin",
 			Value: fmt.Sprintf("-docker-host=unix:///var/run/docker.sock -insecure -sdk=%s -ds=%s -vm-path=%s -cluster=%s -pool=%s -tls=%t",
-				conf.TargetPath, conf.ImageStorePath, vm.InventoryPath, conf.ClusterPath, conf.ResourcePoolPath, false)})
+				conf.Target, conf.ImageStores[0], vm.InventoryPath, conf.ClusterPath, d.vchPoolPath, false)})
 	}
 	spec.ExtraConfig = append(spec.ExtraConfig, &types.OptionValue{Key: "guestinfo.vch/files", Value: files})
 
+	cfg := make(map[string]string)
+	extraconfig.EncodeWithPrefix(extraconfig.MapSink(cfg), conf, "vch")
```
hmahmood: Isn't this already happening in `createApplianceSpec` above (line 179)?

install/management/appliance.go
```diff
@@ -342,18 +324,24 @@ func (d *Dispatcher) setMacToGuestInfo() error {
...
-		keys = append(keys, netName)
+
+		spec.ExtraConfig = append(spec.ExtraConfig, &types.OptionValue{Key: fmt.Sprintf("guestinfo.vch/networks/%s/portgroup", nicName), Value: netInfo.PortGroupName})
+		spec.ExtraConfig = append(spec.ExtraConfig, &types.OptionValue{Key: fmt.Sprintf("guestinfo.vch/networks/%s/mac", nicName), Value: mac})
+		keys = append(keys, nicName)
+		netInfo.Mac = mac
 	}
 	spec.ExtraConfig = append(spec.ExtraConfig, &types.OptionValue{Key: "guestinfo.vch/networks", Value: strings.Join(keys, " ")})
+	cfg := make(map[string]string)
+	extraconfig.EncodeWithPrefix(extraconfig.MapSink(cfg), conf, "vch")
```
hmahmood: Is there a need to do this if `conf` hasn't changed?

install/management/appliance.go
```diff
@@ -342,18 +324,24 @@ func (d *Dispatcher) setMacToGuestInfo() error {
...
 			// timeout to wait MAC address, so empty mac address is returned
 			err = errors.Errorf("Timeout to get VM MAC address")
 			return err
 		}
-		netName := d.nics[key]
-		spec.ExtraConfig = append(spec.ExtraConfig, &types.OptionValue{Key: fmt.Sprintf("guestinfo.vch/networks/%s/portgroup", netName), Value: key})
-		spec.ExtraConfig = append(spec.ExtraConfig, &types.OptionValue{Key: fmt.Sprintf("guestinfo.vch/networks/%s/mac", netName), Value: value})
-		keys = append(keys, netName)
+
+		spec.ExtraConfig = append(spec.ExtraConfig, &types.OptionValue{Key: fmt.Sprintf("guestinfo.vch/networks/%s/portgroup", nicName), Value: netInfo.PortGroupName})
```
hmahmood: These should be in the vch config again.

install/management/appliance.go
emlin: Eventually, they should be part of executor config, but so far I'm not clear for how exectutor config will be used, so do not change that part.
Do you that config this time?

install/management/appliance.go
emlin: This time the conf is changed.

install/management/appliance.go
emlin: What I'm thinking is the conf will change for VM reconfiguration, but right now, there is nothing happened in VCH, so will remove this duplication.

install/management/appliance.go
emlin: This is already in VCH config, https://github.com/vmware/vic/pull/730/files#diff-f8dda6a8e934bfbbf6e19f4868adef44R110
Just I think you'll remove the old code in launcher.sh, I just keep the existing format here, instead of change launcher.sh to read new guestinfo format.
So while you change to read VirtualContainerHostConfigSpec.Networks, these configuration should be removed as well.

hmahmood:
LGTM

emlin:
Fixed issue to work with dvSwitch. Originally, it cannot correct get Mac address if deploy VCH with dvSwitch portgroup

emlin:
@hickeng test passed, is it OK for you now?

metadata/network_interface.go
```diff
@@ -47,4 +47,11 @@ type ContainerNetwork struct {
 
 	// The set of nameservers associated with this network - may be empty
 	Nameservers []net.IP `vic:"0.1" scope:"read-only" key:"dns"`
+
+	// Subnet for the network
+	Subnet net.IPNet `vic:"0.1" scope:"read-only" key:"subnet"`
```
hickeng: The subnet is already represented in the Gateway (IPNet is an IP & Mask pair)

pkg/vsphere/vm/vm.go
```diff
@@ -69,12 +69,35 @@ func (vm *VirtualMachine) FolderName(ctx context.Context) (string, error) {
 }
 
 func (vm VirtualMachine) WaitForMAC(ctx context.Context) (map[string]string, error) {
+	devices, err := vm.Device(ctx)
+	if err != nil {
+		log.Errorf("Unable to get device listing for VM")
+		return nil, err
+	}
+
+	nics := devices.SelectByType(&types.VirtualEthernetCard{})
```
hickeng: add a comment about what is being mapped to what. I presume this is mapping an arbitrary label to "portgroupname" but only because of the variable names where this is used.

pkg/vsphere/vm/vm.go
```diff
@@ -87,17 +110,41 @@ func (vm VirtualMachine) WaitForMAC(ctx context.Context) (map[string]string, err
 					if mac == "" {
 						return false
 					}
-					summary := nic.GetVirtualEthernetCard().DeviceInfo.GetDescription().Summary
-					macs[summary] = mac
+					netName := nicMappings[nic.GetVirtualEthernetCard().DeviceInfo.GetDescription().Summary]
```
hickeng: don't return false in this loop if mac == "" - that means we will not gather the values from other devices reported in this propertychange set. `Return false` suffices for a binary done/not done test but not for building a mapping.

install/management/appliance.go
```diff
@@ -153,26 +126,27 @@ func (d *Dispatcher) getNetworkDevices(conf *configuration.Configuration) ([]typ
...
+
 	spec := &types.VirtualMachineConfigSpec{
-		Name:     conf.DisplayName,
+		Name:     conf.Name,
 		GuestId:  "other3xLinux64Guest",
-		Files:    &types.VirtualMachineFileInfo{VmPathName: fmt.Sprintf("[%s]", conf.ImageDatastoreName)},
-		NumCPUs:  conf.NumCPUs,
-		MemoryMB: conf.MemoryMB,
+		Files:    &types.VirtualMachineFileInfo{VmPathName: fmt.Sprintf("[%s]", conf.ImageStoreName)},
+		NumCPUs:  int32(conf.VCHSize.CPU.Limit),
```
hickeng: This should be conf.ApplianceSize.CPU - the VCH size is for the resource pool that gets created.

hickeng:
In addition to the new comments can you encode the full config object to the ExtraConfig. This will mean that we can start putting in the appliance side changes that pull from guestinfo directly, whereas at the moment there's no way to use it inguest without jumping through hoops.

emlin:
@hickeng the whole VirtualContainerHost object is in ExtraConfig.

hickeng:
LGTM

emlin:
As discussed with George, his last comment is not an issue at this moment.

PR #735, opened 2016-05-19
fixes HTTP status codes for Container Start

matthewavery:
Fixes HTTP codes that I found were incorrect. 

Previously we were returning 503 errors in certain cases with Container Start
That is not in line with docker's Remote API documentation and has been
fixed to reflect what the Docker CLI expects.

@cgtexmex @jzt


andrewtchin:
lgtm

sflxn:
lgtm

PR #753, opened 2016-05-20
Unmount ESXi datastores for use with vSAN

dougm:
The old PXE install of ESXi VMs did not claim virtual disks as local
datastores, but stateless ESXi does.  Unmount these datastores by
default, otherwise they cannot be autoclaimed by vSAN.


dougm:
cc @fdawg4l ran into this yesterday.  And @emlin was trying this today, I believe.

emlin:
LGTM

dougm:
lgtm

PR #754, opened 2016-05-20
(WIP) Calculate and store sha256 of uncompressed layer data

jzt:


portlayer/storage/vsphere/store.go
```diff
@@ -254,8 +256,24 @@ func (v *ImageStore) WriteImage(ctx context.Context, parent *portlayer.Image, ID
...
-		cerr = archive.Untar(r, dir, &archive.TarOptions{})
+		// Decompress the archive into memory (this would happen anyway within the call to Untar)
+		decompressedTar, cerr := archive.DecompressStream(r)
+		if cerr != nil {
+			return nil, cerr
+		}
+
+		// Get the SHA256 sum of the decompressed archive (layer diffID)
+		h := sha256.New()
+		if _, cerr := io.Copy(h, decompressedTar); cerr != nil {
```
fdawg4l: This will copy all of the data presented by the reader (`decompressedTar`) to h.  You will get your sha that way, but the untar below will likely get nothing.  If you use a TeeReader (https://golang.org/pkg/io/#TeeReader), you can avoid the `Copy` of the whole of the data and get the sha inline.  Pretty sure we already do exactly that in `store_cache.go`.  Either move that bit of code (TeeReader for the xfer sha sum) here or do this there.  In either case, `io.Copy` may not be what you want.

jzt: @fdawg4l Good call, thanks. I wonder if we can apply the same optimization to imagec.

Also, this currently results in two metadata files being written to the `<id>/imageMetadata/` directory. Are we ok with that or do we prefer keeping a single blob on disk?

Another related thing to consider is whether or not we actually want/need the diff_id to live under the corresponding layer's directory on disk, versus all diff_ids for a particular image being stored in a single file at the top level layer for that image (since it is only needed by the image metadata JSON and not by the layers themselves).

jzt:
Closing in favor of #781.

PR #762, opened 2016-05-23
Repo reorganisation

hickeng:
A degree of restructuring to the repo.
cmd: contains almost all of the main packages in the repo
pkg: intended to contain the standalone packages that could be reused outside of the repo
lib: the packages intended for use by multiple binaries
infra: anything related to dev infrastructure
doc/bundle: doc files for inclusion into published bundles.


Makefile
```diff
@@ -179,33 +178,30 @@ vendor: $(GVT)
...
-TEST_DIRS+=github.com/vmware/vic/install
 
 
 test:
 	@echo Running unit tests
 	# test everything but vendor
 ifdef DRONE
 	@echo Generating coverage data
-	@scripts/coverage.sh $(TEST_DIRS)
+	@infra/scripts/coverage.sh $(TEST_DIRS)
```
fdawg4l: Minor and need not be addressed in this PR, but if we have a `tests` directory, we might as well move the coverage script there.

fdawg4l:
Looks fine.  As soon as the build is happier, I'd be happy, I'll give it the thumbs up.

Makefile
```diff
@@ -242,31 +238,32 @@ $(imagec): $(call godeps,cmd/imagec/*.go) $(portlayerapi-client)
...
+	@$(SWAGGER) generate client -A PortLayer --template-dir lib/apiservers/templates  -t $(realpath $(dir $<)) -f $<
 
 
 $(portlayerapi-server): $(PORTLAYER_DEPS) $(SWAGGER)
 	@echo regenerating swagger models and operations for Portlayer API server...
-	@$(SWAGGER) generate server -A PortLayer -t $(realpath $(dir $<)) -f $<
+	@$(SWAGGER) generate server -A PortLayer --template-dir lib/apiservers/templates -t $(realpath $(dir $<)) -f $<
 
-$(portlayerapi): $(call godeps,apiservers/portlayer/cmd/port-layer-server/*.go) $(portlayerapi-server) $(portlayerapi-client)
+$(portlayerapi): $(call godeps,lib/apiservers/portlayer/cmd/port-layer-server/*.go) $(portlayerapi-server) $(portlayerapi-client)
```
hmahmood: Should `cmd/port-layer-server` not be under lib?

Makefile
hickeng: /tests is for the integration tests whereas coverage.sh is used to process the `go test` output.

Makefile
hickeng: Ideally, but I've not spent the time to determine how that's to be accomplished. The main package in there is generated by swagger.

hickeng:
@fdawg4l @hmahmood this is odd... looks like the PR didn't update with the most recent. I've got both of those changes locally already and they're what I pushed last night. github is definitely having problems.
Please hold off on further review until the build passes as it'll be a good indicator that the PR is actually up-to-date with the commit.

hickeng:
@caglar10ur @fdawg4l @hmahmood @dougm build is now running and passing.

Makefile
hmahmood: I would have moved the whole apiservers/ folder under cmd/ . None of the code under apiservers is being used by any other component, so it doesn't make sense to me to put it in lib.

fdawg4l:
LGTM

dougm:
LGTM

hickeng:
LGTM

(do we really need this after every rebase?)

fdawg4l:
I think you just need to comment.

PR #772, opened 2016-05-23
Implements docker stop - ticket 363

### LINKED ISSUE #363: Docker API: Docker Container Stop
Implement docker stop operation in our docker engine-api server.  Integrate it with the exec portlayer server.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#stop-a-container
https://docs.docker.com/engine/reference/commandline/stop/

matthewavery:
Fixes #363 

This is our minimal lifecycle implementation of docker stop. it
does not cleanly shut down VMs at the moment. It simply calls
poweroff through govmomi as a first pass. Goals in the future
should involve an interaction with the tether which will allow
us to signal the vm to shutdown all of the services it is
running before a poweroff call. @sflxn @jzt @hmahmood @cgtexmex

### DIFF
--- a/apiservers/engine/backends/container.go
+++ b/apiservers/engine/backends/container.go
@@ -245,7 +245,7 @@ func (c *Container) ContainerStart(name string, hostConfig *container.HostConfig
 			http.StatusInternalServerError)
 	}
 
-	// handle legancy hostConfig
+	// handle legacy hostConfig
 	if hostConfig != nil {
 		// hostConfig exist for backwards compatibility.  TODO: Figure out which parameters we
 		// need to look at in hostConfig
@@ -255,10 +255,9 @@ func (c *Container) ContainerStart(name string, hostConfig *container.HostConfig
 	getRes, err := client.Containers.Get(containers.NewGetParams().WithID(name))
 	if err != nil {
 		if _, ok := err.(*containers.GetNotFound); ok {
-			return derr.NewRequestNotFoundError(err)
+			return derr.NewRequestNotFoundError(fmt.Errorf("No such container: %s", name))
 		}
-
-		return derr.NewErrorWithStatusCode(err, http.StatusInternalServerError)
+		return derr.NewErrorWithStatusCode(fmt.Errorf("server error from portlayer"), http.StatusInternalServerError)
 	}
 
 	h := getRes.Payload
@@ -267,10 +266,9 @@ func (c *Container) ContainerStart(name string, hostConfig *container.HostConfig
 	bindRes, err := client.Scopes.BindContainer(scopes.NewBindContainerParams().WithHandle(h))
 	if err != nil {
 		if _, ok := err.(*scopes.BindContainerNotFound); ok {
-			return derr.NewRequestNotFoundError(err)
+			return derr.NewRequestNotFoundError(fmt.Errorf("server error from portlayer"))
 		}
-
-		return derr.NewErrorWithStatusCode(err, http.StatusInternalServerError)
+		return derr.NewErrorWithStatusCode(fmt.Errorf("server error from portlayer"), http.StatusInternalServerError)
 	}
 
 	h = bindRes.Payload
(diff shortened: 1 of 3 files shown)


apiservers/engine/backends/container.go
```diff
@@ -303,7 +303,46 @@ func (c *Container) ContainerStart(name string, hostConfig *container.HostConfig
...
+	client := PortLayerClient()
+	if client == nil {
+		return derr.NewErrorWithStatusCode(fmt.Errorf("container.ContainerCreate failed to create a portlayer client"),
+			http.StatusInternalServerError)
+	}
+
+	getResponse, err := client.Containers.Get(containers.NewGetParams().WithID(name))
+	if err != nil {
+		if _, ok := err.(*containers.GetNotFound); ok {
+			return derr.NewRequestNotFoundError(err)
```
matthewavery: Not sure what to output here for a more defined errror. Any suggestions?

apiservers/engine/backends/container.go
```diff
@@ -303,7 +303,46 @@ func (c *Container) ContainerStart(name string, hostConfig *container.HostConfig
...
+	}
+
+	handle := getResponse.Payload
+
+	// change the state of the container
+	// TODO: We need a resolved ID from the name
+	stateChangeResponse, err := client.Containers.StateChange(containers.NewStateChangeParams().WithHandle(handle).WithState("STOPPED"))
+	if err != nil {
+		if _, ok := err.(*containers.StateChangeNotFound); ok {
+			return derr.NewRequestNotFoundError(err)
```
matthewavery: ditto.

apiservers/engine/backends/container.go
```diff
@@ -303,7 +303,46 @@ func (c *Container) ContainerStart(name string, hostConfig *container.HostConfig
...
+		// If we get here, most likely something went wrong with the port layer API server
+		return derr.NewErrorWithStatusCode(err, http.StatusInternalServerError)
+	}
+
+	handle = stateChangeResponse.Payload
+
+	_, err = client.Containers.Commit(containers.NewCommitParams().WithHandle(handle))
+	if err != nil {
+		if _, ok := err.(*containers.CommitNotFound); ok {
+			return derr.NewRequestNotFoundError(err)
```
matthewavery: ditto.

portlayer/exec/container.go
```diff
@@ -176,3 +176,23 @@ func (c *Container) Start(ctx context.Context) error {
...
+	defer trace.End(trace.Begin("Container.Stop"))
+
+	if c.vm == nil {
+		return fmt.Errorf("vm not set")
+	}
+
+	//TODO: make the shutdown much cleaner, right now we just pull the plug on the vm.(may need corresponding work in tether.)
+
+	// Power off
+	_, err := tasks.WaitForResult(ctx, func(ctx context.Context) (tasks.ResultWaiter, error) {
```
fdawg4l: Does it matter if the container is already off?

fdawg4l:
Is it possible to provide a unit test for this change?

portlayer/exec/container.go
matthewavery: It looks like govmomi already covers this case. same for start if you try to start a container that is already started then you will get a change state error.

fdawg4l: Sounds good, but what about the docker client?  Does it get some useful and known error or does the error returned confuse it?

matthewavery:
are there unit tests for the docker personality calls that exist now? These are api calls so an integration test is more suited for the task. Once the Docker Tests are working, we should have the same tests running against this call as docker has against theirs.

portlayer/exec/container.go
matthewavery: testing it now, i just changed the error codes section so you are likely to get a "error occurred in the portlayer" error with a 500 status code.

matthewavery: yes, you will get the error `Error response from daemon: server error from portlayer` at the moment.

portlayer/exec/container.go
```diff
@@ -176,3 +176,23 @@ func (c *Container) Start(ctx context.Context) error {
...
+		return c.vm.PowerOff(ctx)
```
jzt: Would a Suspend be better in this case?

matthewavery: Eventually yes, since the docker documentation mentions that the `docker stop` sends a sigterm and then a sigkill we will eventually want to implement something similar in the tether. For minimal functionality PowerOff is being used.

hmahmood:
LGTM

portlayer/exec/container.go
hickeng: Testing with local docker shows that stopping a stopped container doesn't return an error - the operation is idempotent and I like that behaviour for a mostly declarative API.

hickeng:
LGTM

This is very bare bones - when we flesh it out it needs to encompass things such as stop signal mapping, signal delivery, handling of exec containers (where powering down the VM isn't the appropriate response), etc.

jzt:
LGTM

fdawg4l:
LGTM

PR #774, opened 2016-05-23
Fix vSAN with stateless ESXi

dougm:
Commit 1cc8ebb was not good enough when used in a fresh environment.
So, rather than unmount the ESX datastores, add virtual disks after ESXi
has booted.  This requires 2 additional changes:

- rescan HBA to pickup the new devices

- mark disk as SSD (extraConfig trick does not work after VM has booted)

### DIFF
--- a/machines/vcsa/README.md
+++ b/machines/vcsa/README.md
@@ -20,7 +20,7 @@ But, you should already be using and loving jq for other tasks: http://stedolan.
 ### create-esxi-vm.sh
 
 This script creates a VM running stateless ESXi, booted via cdrom/iso.
-It will create 2 disks:
+It will create 2 disks by default:
 
 * vSAN cache disk (Virtual SSD)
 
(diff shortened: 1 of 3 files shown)


dougm:
Depends on: https://github.com/vmware/govmomi/pull/511

dougm:
cc @fdawg4l @emlin 

Note that I haven't updated govmomi/govc in our vendor/ yet.  For the moment, just do:
```
go get -u github.com/vmware/govmomi/govc
```

fdawg4l:
LGTM

hickeng:
LGTM

fdawg4l:
Still wonky.  Each host now shows 2 disks vsphere claims is "Not In Use".  With Auto claim configured, capacity is still 0 bytes.  Looks like the disks aren't being added to vsan unless I'm not waiting long enough.  It's been about 10m.

fdawg4l:
Feel free to poke around.  Will slack host details.

dougm:
Can't see why it would happen, but all your disks were marked as SSD.  Its only supposed to be the 1st disk (cache), and 2nd disk (store) must be HHD.  After I changed your 2nd to HHD (via govc host.storage.mark -ssd=false $disk), your vSAN datastore has formed.

Going to merge, we can revisit if this happens in your env again.

PR #781, opened 2016-05-24
Calculate diffID for each layer in imagec

### LINKED ISSUE #775: Calculate layer diffIDs in imagec
For all the layers that compose an image, the diffIDs are required in order to calculate the ID of the image. Since this is docker-specific behavior, it should occur in either imagec or the docker personality. imagec seems like the right place, as it knows all about the layers in an image while doing a docker pull.

jzt:
Currently this just logs the diffID and stores it in the `ImageWithMeta` struct for each layer after it is calculated.

Fixes #775

### DIFF
--- a/cmd/imagec/docker.go
+++ b/cmd/imagec/docker.go
@@ -27,6 +27,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 
+	"github.com/docker/docker/pkg/archive"
 	"github.com/docker/docker/pkg/progress"
 
 	"github.com/vmware/vic/pkg/trace"
@@ -144,16 +145,17 @@ func FetchToken(url *url.URL) (*Token, error) {
 }
 
 // FetchImageBlob fetches the image blob
-func FetchImageBlob(options ImageCOptions, image *ImageWithMeta) error {
+func FetchImageBlob(options ImageCOptions, image *ImageWithMeta) (string, error) {
 	defer trace.End(trace.Begin(options.image + "/" + image.layer.BlobSum))
 
 	id := image.ID
 	layer := image.layer.BlobSum
 	history := image.history.V1Compatibility
+	diffID := ""
 
 	url, err := url.Parse(options.registry)
 	if err != nil {
-		return err
+		return diffID, err
 	}
 	url.Path = path.Join(url.Path, options.image, "blobs", layer)
 
@@ -170,7 +172,7 @@ func FetchImageBlob(options ImageCOptions, image *ImageWithMeta) error {
 	})
 	imageFileName, err := fetcher.FetchWithProgress(url, image.String())
 	if err != nil {
-		return err
+		return diffID, err
 	}
 
 	// Cleanup function for the error case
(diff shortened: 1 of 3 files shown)


cmd/imagec/docker.go
```diff
@@ -144,16 +145,17 @@ func FetchToken(url *url.URL) (*Token, error) {
 }
 
 // FetchImageBlob fetches the image blob
-func FetchImageBlob(options ImageCOptions, image *ImageWithMeta) error {
+func FetchImageBlob(options ImageCOptions, image *ImageWithMeta) (*string, error) {
```
fdawg4l: Why a `*string` vs `string`?

jzt: I just didn't like returning "" in the error cases. I'm happy to change it if it's not proper go convention.

fdawg4l: I prefer "" vs a pointer to a string.  I think we do that in other places as well.

cmd/imagec/docker.go
```diff
@@ -183,45 +185,89 @@ func FetchImageBlob(options ImageCOptions, image *ImageWithMeta) error {
...
+
+	r, w := io.Pipe()
+	t := io.TeeReader(in, w)
+
+	// validate the sha256 sum of the compressed data
+	go func() {
+		defer w.Close()
+		h := sha256.New()
+		progress.Update(po, image.String(), "Verifying Checksum")
+		if _, cerr := io.Copy(h, t); cerr != nil {
```
fdawg4l: This looks sufficiently complicated that a nice comment of what is wired to what reader/writer would be appreciated.  

The single letter variables is making the logic pretty difficult to understand at first pass.  Please make things a bit easier to read if you can.  `piper, pipew, tr, xferSum`, etc.

cmd/imagec/docker.go
jzt: Fixed

cmd/imagec/docker.go
```diff
@@ -183,45 +185,89 @@ func FetchImageBlob(options ImageCOptions, image *ImageWithMeta) error {
...
 	defer in.Close()
 
-	h := sha256.New()
+	done := make(chan bool)
+	errs := make(chan error)
+	defer close(done)
+	defer close(errs)
+
+	r, w := io.Pipe()
+	t := io.TeeReader(in, w)
```
fdawg4l: This looks lopsided.  One of the go funcs uses `t` while the other uses `r`.  So, can you reduce by only using one of them (either Tee or Pipe)?

fdawg4l: And if you can reduce them to 1 (either Tee or Pipe), do we need 2 go funcs?

cmd/imagec/docker.go
```diff
@@ -183,45 +185,89 @@ func FetchImageBlob(options ImageCOptions, image *ImageWithMeta) error {
...
+		d := fmt.Sprintf("%x", h.Sum(nil))
+		diffID = &d
+		done <- true
+	}()
 
-	// Calculate the sum
-	sum := fmt.Sprintf("sha256:%x", h.Sum(nil))
-	if sum != layer {
-		return fmt.Errorf("Failed to validate layer checksum. Expected %s got %s", layer, sum)
+	// wait for both goroutines to finish
```
fdawg4l: This will work, but `sync.WaitGroup` does this without the need for the `for` with `select`.  You do  the following
```
wg := &sync.WaitGroup{}
go func() { wg.Add(1) }()
go func() { wg.Add(1) }()
wg.Wait()
```

cmd/imagec/docker.go
jzt: Fixed, how's it look now?

jzt:
@fdawg4l I refactored out the goroutines and pipes and switched to using chained `io.TeeReader`s. What do you think?

cmd/imagec/docker.go
```diff
@@ -181,47 +183,64 @@ func FetchImageBlob(options ImageCOptions, image *ImageWithMeta) error {
...
+	diffIDSum := sha256.New()
+
+	// blobTr is an io.TeeReader that writes bytes to blobSum that it reads from imageFile
+	// see https://golang.org/pkg/io/#TeeReader
+	blobTr := io.TeeReader(imageFile, blobSum)
 
 	progress.Update(po, image.String(), "Verifying Checksum")
-	if _, cerr := io.Copy(h, in); cerr != nil {
-		return cerr
+	tar, err := archive.DecompressStream(blobTr)
```
fdawg4l: Shadow declaration.  `vet` may complain. Please ignore if otherwise.

cmd/imagec/docker.go
```diff
@@ -181,47 +183,64 @@ func FetchImageBlob(options ImageCOptions, image *ImageWithMeta) error {
...
-	sum := fmt.Sprintf("sha256:%x", h.Sum(nil))
-	if sum != layer {
-		return fmt.Errorf("Failed to validate layer checksum. Expected %s got %s", layer, sum)
+	// Copy bytes from decompressed layer into diffIDSum to calculate diffID
+	if _, cerr := io.Copy(diffIDSum, tar); cerr != nil {
+		return diffID, cerr
 	}
 
+	bs := fmt.Sprintf("sha256:%x", blobSum.Sum(nil))
+	if bs != layer {
```
fdawg4l: Just curios.  Does this test work?  I couldnt tell if docker hub was doing a sum over the files or a sum over the tar and wasn't sure what to check.

fdawg4l:
LGTM.  Nice work!

cmd/imagec/docker.go
jzt: Yeah, it's doing a sum over the compressed layer. So we do the same and just check it against the one in the manifest to make sure they match.

This 1.10 design doc was an enlightening read for me, especially this part:  https://gist.github.com/aaronlehmann/b42a2eaf633fc949f93b#id-definitions-and-calculations

hickeng:
LGTM

Was debating commenting on the double decompression (here and in the PL) but couldn't think of a way to do it that also reduced copying, so decided to leave it. Don't spend time thinking about it now, but it's an area for potential future optimization.

fdawg4l:
@hickeng The pl IIRC also uses the 'archive' package and that checks for compression headers before unpacking using a compressor. I think we're fine unless I'm missing a step.

PR #784, opened 2016-05-24
Mark vSAN store disk as non-SSD

dougm:
If the underlying disk is SSD, the virtual disks are reported as SSD.
vSAN requires 1 SSD and 1 HDD.


fdawg4l:
:sheep: :it:

:eyes: :+1: :v: :meat_on_bone:

dougm:
lgtm

PR #785, opened 2016-05-25
Implement network connect

### LINKED ISSUE #703: Implement docker network connect
`docker create` does not allow creating a container that is connected to more than one network, which is required for a container to act as a bridge between two networks. The only way to do that in docker is the `network connect` command.

hmahmood:
Fixes #703

### DIFF
--- a/cmd/tether/config_test.go
+++ b/cmd/tether/config_test.go
@@ -54,8 +54,7 @@ func TestToExtraConfig(t *testing.T) {
 		},
 		Networks: map[string]*metadata.NetworkEndpoint{
 			"eth0": &metadata.NetworkEndpoint{
-				IP:  net.IPNet{IP: localhost, Mask: lmask.Mask},
-				MAC: "a-mac-address",
+				IP: net.IPNet{IP: localhost, Mask: lmask.Mask},
 				Network: metadata.ContainerNetwork{
 					Name:        "notsure",
 					Gateway:     net.IPNet{IP: gateway, Mask: gmask.Mask},
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -104,6 +104,8 @@ func (c *Container) ContainerStatPath(name string, path string) (stat *types.Con
 func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.ContainerCreateResponse, error) {
 	defer trace.End(trace.Begin("ContainerCreate"))
 
+	var err error
+
 	//TODO: validate the config parameters
 	log.Printf("config.Config = %+v", config.Config)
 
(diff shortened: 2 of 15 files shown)


dougm:
lgtm

portlayer/exec/container.go
```diff
@@ -80,44 +86,78 @@ func (c *Container) newHandle() *Handle {
...
-		if err != nil {
-			return err
-		}
-	} else {
-		// Find the Virtual Machine folder that we use
-		folders, err := sess.Datacenter.Folders(ctx)
-		if err != nil {
-			return err
+func (c *Container) Commit(ctx context.Context, sess *session.Session, h *Handle) error {
+	c.Lock()
```
jzt: I remember @caglar10ur and @fdawg4l had a good reason for naming your mutexes inside of structs rather than embedding them. It may not apply here, but I CC'd them just in case. Maybe one of them can refresh my memory.

apiservers/engine/backends/network.go
```diff
@@ -123,7 +129,64 @@ func (n *Network) CreateNetwork(name, driver string, ipam apinet.IPAM, options m
 }
 
 func (n *Network) ConnectContainerToNetwork(containerName, networkName string, endpointConfig *apinet.EndpointSettings) error {
-	return fmt.Errorf("%s does not implement network.ConnectContainerToNetwork", n.ProductName)
+	client := PortLayerClient()
+	getRes, err := client.Containers.Get(containers.NewGetParams().WithID(containerName))
```
andrewtchin: What is "Res" short for?

hmahmood: Result

portlayer/exec/container.go
hmahmood: I believe this is only a problem if you are passing the struct by value, so of course the lock state will be invalid in the copy. Here we are not doing that, `Container` is always used as a pointer.

portlayer/network/context.go
```diff
@@ -54,34 +50,28 @@ type Context struct {
...
-	c := context.Background()
-	vch, err := guest.GetSelf(c, sess)
-	if err != nil {
-		return "", fmt.Errorf("unable to get VCH ref")
-	}
-
-	vchVM := vm.NewVirtualMachine(c, sess, vch.Reference())
-	guestInfo, err := vchVM.FetchExtraConfig(c)
+var getBridgeNetworkName = func() (string, error) {
+	config := rpcvmx.NewConfig()
```
andrewtchin: 👍

portlayer/network/context.go
```diff
@@ -508,51 +498,61 @@ func findSlotNumber(slots map[int32]bool) int32 {
...
 	}
 
-	var devices object.VirtualDeviceList
-	backing := &types.VirtualEthernetCardNetworkBackingInfo{
-		VirtualDeviceDeviceBackingInfo: types.VirtualDeviceDeviceBackingInfo{
-			DeviceName: s.NetworkName,
-		},
-	}
+	if d == nil {
+		backing := &types.VirtualEthernetCardNetworkBackingInfo{
```
andrewtchin: indent?

andrewtchin:
lgtm brah

portlayer/network/context.go
```diff
@@ -639,6 +639,90 @@ func (c *Context) AddContainer(h *exec.Handle, scope string, ip *net.IP) (*Endpo
...
+			return
+		}
+
+		var ip *net.IP
+		if e.static {
+			i := e.IP()
+			ip = &i
+		}
+
+		s.addContainer(con, ip)
```
hmahmood: This is a rollback in case of error.

hmahmood:
Lgtm

Issue #777, opened 2016-05-23
Persist image metadata into the child-most layer of said image


jzt:
Once the image metadata is generated and sha256'd, that data can be included into the layer metadata blob that is sent to the portlayer for the child-most layer of the image. The docker personality can later use this metadata to construct a cache of available images that have been pulled. Depends on #776.

jzt:
I had to change the way I calculate image size as docker has (for some unknown reason) removed the `Size` attribute from the `v1Compatibility` metadata in the manifest.

PR #792, opened 2016-05-26
Construct image metadata in imagec and use it to calculate image iD

### LINKED ISSUE #776: Construct image metadata in imagec
Once all layers have been pulled by imagec, it now has enough information to construct the image's metadata JSON structure. Once this is created, a sha256 sum can be taken of this json struct which will represent the ID of the actual image. Depends on #775.

jzt:
Currently, these changes can create everything we need in the image metadata JSON. It currently uses this metadata to calculate the sha256 sum that will serve as the Image ID. Currently, this ID is logged to imagec's output, but will be persisted to the storage portlayer in the next task (#777).

Fixes #776

### DIFF
--- a/cmd/imagec/imagec.go
+++ b/cmd/imagec/imagec.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"crypto/sha256"
 	"encoding/json"
 	"flag"
 	"fmt"
@@ -23,6 +24,7 @@ import (
 	"os"
 	"path"
 	"runtime/trace"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


dougm:
Regarding the map, you can pass a map by reference but can make things hard to read.  Another option is to create a struct type with map as a field and pass the struct by reference.

jzt:
@dougm I was able to modify the existing code to pass around a `[]*ImageWithMeta` instead and that fixed it. Thanks @cgtexmex.

cmd/imagec/imagec.go
```diff
@@ -370,6 +375,60 @@ func WriteImageBlobs(images []ImageWithMeta) error {
 	return nil
 }
 
+// CreateImageConfig constructs the image metadata from layers that compose the image
+func CreateImageConfig(images []*ImageWithMeta) error {
+
+	image := docker.Image{}
+	rootFS := docker.NewRootFS()
+	history := []docker.History{}
```
dougm: Since you know what the length of history will be:
```go
history := make([]docker.History, 0, len(image))
```

dougm:
lgtm

andrewtchin:
lgtm brah

cmd/imagec/imagec.go
jzt: Fixed!

PR #793, opened 2016-05-26
Add repo structure comments after reorg

hickeng:
Should have been in the reorg itself but forgot.

[skip ci]


fdawg4l:
LGTM. 

Why does a doc change cause the build to fail??  Weird.

dougm:
lgtm

hickeng:
@fdawg4l I killed the build because it is just a doc change. I forgot to add [skip ci] to the PR description ahead of time - also not sure that's the correct tag but remember @frapposelli mentioning it.

PR #797, opened 2016-05-27
Clean up vic-machine VM creation

### LINKED ISSUE #571: Refactor vic-machine: use vsphere/spec to create VM spec

emlin:
vic-machine is creating and configuring appliance VMs through direct VM spec construction, so looks pretty ugly.
This pull request used VirtualDeviceList to create device and generate VM configure spec, to avoid some weird vSphere parameters. And make the logic looks clear.

I didn't use spec library to create VM spec, is because that library contains some attributes not used by appliance VM, and right now they have some assumption not fit for appliance VM, for example only one nic etc.  
Fixes #571


andrewtchin:
lgtm

dougm:
lgtm

I find the ExtraConfig related code hard to read.  In a future change, maybe we could build up the config with a simpler structure, such as `map[string]string`, then convert to `[]types.OptionValue` at the end.  Or even just adding newlines to the existing code.  For example, instead of:
```go
&types.OptionValue{Key: "guestinfo.vch/components", Value: "/sbin/docker-engine-server /sbin/port-layer-server /sbin/vicadmin"},
```
this:
```go
&types.OptionValue{
    Key: "guestinfo.vch/components",
    Value: "/sbin/docker-engine-server /sbin/port-layer-server /sbin/vicadmin",
},
```

emlin:
Yeah, it is.
I reformatted the code as your suggestion, instead of convert them from a map, cause eventually, suppose we'll serialize the VirtualContainerHost only, instead of set these guestinfo manually.

dougm:
lgtm

PR #804, opened 2016-05-27
Fix wait for mac issue while connect to dvSwitch

### LINKED ISSUE #800: if set external-network or management-network to vDS, vic-machine failed for cannot get MAC address

emlin:
If multiple networks in same dvSwitch is configured, waitForMac can only return one nic's mac address. 
Also make sure no more nics created for same network.
Fixes #800


dougm:
lgtm

caglar10ur:
LGTM :sheep: :it:

Issue #919, opened 2016-05-31
vic-machine delete volume interaction


hickeng:
Initially we’ll go with this:
1. the directories specified for volume-store's must be empty during VCH create
2. when deleting the VCH, require a --force flag if the volume-store directories are *not* empty
3. if –force is specified then delete the vmdk’s in the volume store directories

It’s very simple behavior (same behavior as rmdir) and it provides basic protection against accidental deletion.

To confirm, please can you add a section to vic-machine.md that describes the behavior with CLI example.

hickeng:
@emlin @fdawg4l Emma, please can you make a vic-machine.md PR with the example as a readback.

PR #921, opened 2016-05-31
Description about vic-machine delete

emlin:
This PR is for #919 to confirm vic-machine delete design.

### DIFF
--- a/doc/design/vic-machine.md
+++ b/doc/design/vic-machine.md
@@ -129,30 +129,30 @@ below this point is working notes.
 
 ### Inputs
 
-VCH user (existing or new) **
-resource lists:
- pool **
- imagestore datastore paths **
- contianer datastore paths **
- volume datastore paths (restriction)
- network mappings:
-  one network minimum for VCH comms**
-  other network mappings
-resource allotments:
- cpu
- memory
- network IO
- disk IO
- datastore quotas (per datastore path)
-certificates
- users - for access to VCH
- hosts - for container access to external hosts
- network - for VCH/container access to networks (gating proxies)
-registry lists
- whitelist
- blacklist
-default container resource reservations and limits *
-containerVM naming convention (displayName for vSphere) *
+* VCH user (existing or new) **
+* resource lists:
+ - pool **
+ - imagestore datastore paths **
+ - container datastore paths **
+ - volume datastore paths (restriction)
+ - network mappings:
+  - one network minimum for VCH comms**
+  - other network mappings
+* resource allotments:
+ - cpu
+ - memory
+ - network IO
+ - disk IO
+ - datastore quotas (per datastore path)
+* certificates
+ - users - for access to VCH
+ - hosts - for container access to external hosts
+ - network - for VCH/container access to networks (gating proxies)
+* registry lists
+ - whitelist
+ - blacklist
+* default container resource reservations and limits *
+* containerVM naming convention (displayName for vSphere) *
 
 ### Actions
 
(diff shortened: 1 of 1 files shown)


doc/design/vic-machine.md
```diff
@@ -129,30 +129,30 @@ below this point is working notes.
...
+* resource allotments:
+ - cpu
+ - memory
+ - network IO
+ - disk IO
+ - datastore quotas (per datastore path)
+* certificates
+ - users - for access to VCH
+ - hosts - for container access to external hosts
+ - twork - for VCH/container access to networks (gating proxies)
```
hickeng: looks like this got mangled.

doc/design/vic-machine.md
```diff
@@ -179,6 +179,47 @@ Some of the elevated privilege operations could be delegated during self-provisi
...
 
+
+## Deleting - per vSphere target
+
+### Inputs
+
+1. vSphere SDK endpoint
+2. vSphere administrative credentials
+3. VCH identification
+ - VCH resource pool path in govc format, e.g. /ha-datacenter/vm/vch1
```
hickeng: the moref is a shorter form but doesn't really matter so long as it matches the output in `vic-machine ls` so copy-paste can be used.

doc/design/vic-machine.md
```diff
@@ -179,6 +179,47 @@ Some of the elevated privilege operations could be delegated during self-provisi
...
+
+### Actions
+
+1. Get VCH VM
+2. Read back VCH configuration from VM guestinfo
+3. Delete following resources based on VCH configuration.
+ - Container VMs managed by the VCH
+   - Contianer datastore paths and resource pool path configured during installation will be used here, to detect if the VM belongs to this VCH. If yes, these VMs will be removed.
+   - Container VMs will be removed even -force is not specified
```
hickeng: require --force if there are containerVMs that are powered on.

doc/design/vic-machine.md
```diff
@@ -179,6 +179,47 @@ Some of the elevated privilege operations could be delegated during self-provisi
...
+ - volumes managed by the VCH
+   - -force option is required to delete volumes together with VCH uninstallation.
+    - Volume datastore path (sample: ds://datastore1/volume/vch1) configured during installation will be used here, to detect if the volumes are created by this VCH.
+     - If volume directory (sample: ds://datastore1/volume/vch1) is created by vic-machine create, vic-machine delete should delete this directory as well (how to know if the directory is created by vic-machine create?).
```
hickeng: I don't think it matters if the directory was created _by_ vic-machine as the directory was created _for_ or assigned to the VCH. We can just delete it if it's empty.

doc/design/vic-machine.md
```diff
@@ -179,6 +179,47 @@ Some of the elevated privilege operations could be delegated during self-provisi
...
+ - images managed by the VCH
+   - Images are shared by VCHs. Currently no reference or metadata to specify which image is referenced by which VCH, so for TP3, vic-machine delete will not touch images.
```
hickeng: Images are not currently shared by VCHs unless you point them at the same image-store. The code does not currently support this so while it will sort of function there's not protection for concurrent pulls/rmi/et al.

For now assume that images are deleted with the VCH. The way we will share images is likely via a hierarchical model:
```
vic-machine -image-store=ds://path/to/shared images -image-store=ds://path/to/vch/specific/images
```
In that case we'd only delete the images in the leaf location (last image-store specified).

doc/design/vic-machine.md
```diff
@@ -179,6 +179,47 @@ Some of the elevated privilege operations could be delegated during self-provisi
...
+ - vSphere networks created
+   - vic-machine create or VCH port-layer-server will create network, so vic-machine delete should delete any new networks. (Still need to confirm for how to get created networks)
```
hickeng: If necessary we can add a boolean to the VCH config for whether the various resources were created by the VCH - there is a 'managed-by' property in vmodl but I don't know if it can be set to anything other than an extension id.

doc/design/vic-machine.md
```diff
@@ -179,6 +179,47 @@ Some of the elevated privilege operations could be delegated during self-provisi
...
+ - VCH specific metadata from vsphere objects if any
+   - Cause image will not be removed at this time, image metadata will not touched as well. But for container metadata, if any, will be deleted. (Need to finalize where it is)
```
hickeng: Container metadata is in the VM extraconfig - that will be deleted with the VM. I am not aware of any other metadata at this point - if we create some to support the portlets for example then that needs to be cleaned up.

doc/design/vic-machine.md
```diff
@@ -179,6 +179,47 @@ Some of the elevated privilege operations could be delegated during self-provisi
...
+ - the resource pool and appliance VM
+
+### Samples
+```
+vic-machine delete 
+-target 192.168.1.1
```
hickeng: Looking at the vic-machine.md PR I've got in progress the target specifier is:
`<user:password>@fqdn/path/to/datacenter/`

This is because the datacenter is the top level namespace and is necessary to disambiguate all other resources. In the case of ESX there can be only one, but in general:
`-target=https://root@****:vcenter.fqdn/path/to/datacenter`

My apologies for the shifting of argument specifications - it's needed quite a few iterations to get to something sane(ish). I still need to revise the examples with the -FROM argument.

doc/design/vic-machine.md
emlin: And cause it's the format of govc resources, it's easy to figure out what's that. But for mob-id, user will have to look up that through VC mob api, before we have vic-machine ls.
So here I plan to support path first.

doc/design/vic-machine.md
emlin: As our discussion today, unlikely we'll have this file directory ready before TP3, then please let me know where is current image directory while it's ready, and then will remove that directory during delete.

doc/design/vic-machine.md
emlin: till now, the portlet might only rely on VM guestinfo. We can figure out if anything else happens.

doc/design/vic-machine.md
emlin: I understand that we need to tuning the parameter name and format, to make it better. Everything related to use experience should be designed carefully.

But for the target, I'm still confused. 
```
<user:password>@fqdn/path/to/datacenter/ 
```
and 
```
https://root@****:vcenter.fqdn/path/to/datacenter
```
Are they same? user:password and user@****

hickeng:
LGTM

doc/design/vic-machine.md
```diff
@@ -129,30 +129,30 @@ below this point is working notes.
...
-registry lists
- whitelist
- blacklist
-default container resource reservations and limits *
-containerVM naming convention (displayName for vSphere) *
+* VCH user (existing or new) **
+* resource lists:
+ - pool **
+ - imagestore datastore paths **
+ - contianer datastore paths **
```
andrewtchin: container

andrewtchin:
lgtm

emlin:
♻️

PR #931, opened 2016-06-01
Refactor vcsim moref generation

dougm:
Generating a ManagedObjectReference.Value was a bit cumbersome.
Refactor so the Registry takes care of this in most cases.  The
exception is the case where we want to use an object template,
such as esx.HostSystem, where the caller still needs to set the new
Value.


fdawg4l:
LGTM :santa:

caglar10ur:
LGTM :sheep: :it:

Issue #436, opened 2016-04-05
Milestone: docker rm


hickeng:
Delete an existing, stopped container.

The low level port layer implementation should not delete attached named volumes, but likely should delete attached anonymous volumes - this maps much more closely to an intuitive behaviour than the orphaned volume behaviour with docker.

For a conservative initial implementation we don't delete any directly.
We can construct the docker semantics in the personality.

#688 will implement a minimal subset of this function

corrieb:
Since #364 is the only remaining issue for this epic, right now, assigning to @matthewavery

matthewavery:
this is complete, all issues related to this epic have been completed.

Issue #688, opened 2016-05-16
Milestone: minimal container lifecycle


hickeng:
This provides minimal implementation for the remainder of the core container lifecycle functions. This is specifically to support work such as the performance tests and should be targeted at allowing those to proceed.

This adds placeholder implementations for:
- [ ] docker stop - implement with raw vm power off and no concern at this point with filesystem sync.
- [ ] docker rm - again skeletal, with no care taken for preservation of volumes, etc, at this point.

Issue #364, opened 2016-03-31
Docker API: Docker Container Remove


sflxn:
Implement the docker container remove operation (docker rm from CLI) in our docker engine-api server and integrate that with the storage port layer.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#remove-a-container

Please note, the engine-api's rest server already packaging these json struct into golang structs. The url above is just for reference on what the operation should do.

matthewavery:
Minimal work has been completed on this Issue, all that is left is providing support for the options.

mdubya66:
@matthewavery is this complete?

matthewavery:
This is just pending the flag support for container removal. Currently there is no way to gracefully remove the containers.

matthewavery:
According to ticket #436 we will need a way to discern between anonymous and regular volumes. Then we will need `docker volume rm` to work for us to remove anonymous volumes with the container that is associated with it. additionally for the `docker rm -f` we should be using kill to send signals to the tether. So this is dependent on the `docker kill` functionality being in place.

hickeng:
@matthewavery is this still TBD?

mdubya66:
Closing. @matthewavery if this is incorrect please talk to @hickeng and reopen

PR #939, opened 2016-06-01
Minimal docker rm

### LINKED ISSUE #782: Port layer: Initial Container Removal of stopped containers
Swagger definition for exec portlayer's container remove.
Basic implementation of removal code.  Ignore volumes.  Check for container state.

matthewavery:
Fixes #782 and is the beginning of #364 

This also addresses the final remaining command missing for #688.

Right now this will not work from a direct pull and build due to a bug with our current vendored version of go-swagger and the delete function. one proposition that I would like to discuss in this PR is moving the call to `/containers/delete/{id}` and posing the call as a `POST` instead. This can be used to circumvent the `DELETE` content type bug in go-swagger. 

The exec layer portion removes containers from the container map which is in the exec layer. the handler for now only calls the vmremove function. future functionality will have us commiting a spec change for detaching the volume's before removal. 

We only currently support the most minimal `rm` call the only parameter which is supported is the `name` parameter. the parameters left for future implementation are `force` and `v` which is a bool for whether or not to remove associated volumes before deletion.

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -165,7 +165,6 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 		config.Config.Entrypoint = layer.Config.Entrypoint
 	}
 	config.Config.Env = append(config.Config.Env, layer.Config.Env...)
-
 	log.Printf("config.Config' = %+v", config.Config)
 
 	// Call the Exec port layer to create the container
@@ -246,7 +245,28 @@ func (c *Container) ContainerRestart(name string, seconds int) error {
 }
 
 func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) error {
-	return fmt.Errorf("%s does not implement container.ContainerRm", c.ProductName)
+	defer trace.End(trace.Begin("ContainerRm"))
+
+	// Get the portlayer Client API
+	client := PortLayerClient()
+	if client == nil {
+		return derr.NewErrorWithStatusCode(fmt.Errorf("container.ContainerRm failed to create a portlayer client"),
+			http.StatusInternalServerError)
+	}
+
+	//TODO: verify params that are passed in(like -f) and then pass them along.
+
+	//FIXME: currently the portlayer yaml does not support more params than the simple name param
+
+	//call the remove directly on the name. No need for using a handle.
+	_, err := client.Containers.ContainerRemove(containers.NewContainerRemoveParams().WithID(name))
+	if err != nil {
+		if _, ok := err.(*containers.ContainerRemoveNotFound); ok {
+			return derr.NewRequestNotFoundError(fmt.Errorf("No such container: %s", name))
+		}
+		return derr.NewErrorWithStatusCode(fmt.Errorf("server error from portlayer"), http.StatusInternalServerError)
+	}
+	return nil
 }
 
 func (c *Container) ContainerStart(name string, hostConfig *container.HostConfig) error {
@@ -486,9 +506,9 @@ func toModelsNetworkConfig(cc types.ContainerCreateConfig) *models.NetworkConfig
 
 func (c *Container) imageExist(imageID string) (storeName string, err error) {
 	// Call the storage port layer to determine if the image currently exist
-	host, err := os.Hostname()
+	host, err := guest.UUID()
 	if err != nil {
-		return "", derr.NewBadRequestError(fmt.Errorf("container.ContainerCreate got unexpected error getting hostname"))
+		return "", derr.NewBadRequestError(fmt.Errorf("container.ContainerCreate got unexpected error getting VCH UUID"))
 	}
 
 	getParams := storage.NewGetImageParams().WithID(imageID).WithStoreName(host)
--- a/lib/apiservers/portlayer/restapi/configure_port_layer.go
+++ b/lib/apiservers/portlayer/restapi/configure_port_layer.go
@@ -17,18 +17,17 @@ package restapi
 import (
 	"net/http"
 
-	"golang.org/x/net/context"
-
 	log "github.com/Sirupsen/logrus"
 
 	errors "github.com/go-swagger/go-swagger/errors"
 	httpkit "github.com/go-swagger/go-swagger/httpkit"
 	"github.com/go-swagger/go-swagger/swag"
-
 	"github.com/vmware/vic/lib/apiservers/portlayer/restapi/handlers"
 	"github.com/vmware/vic/lib/apiservers/portlayer/restapi/operations"
 	"github.com/vmware/vic/lib/apiservers/portlayer/restapi/options"
 	"github.com/vmware/vic/pkg/vsphere/session"
+
+	"golang.org/x/net/context"
 )
 
 // This file is safe to edit. Once it exists it will not be overwritten
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -54,7 +54,7 @@ func (handler *ContainersHandlersImpl) Configure(api *operations.PortLayerAPI, h
 	api.ContainersGetHandler = containers.GetHandlerFunc(handler.GetHandler)
 	api.ContainersCommitHandler = containers.CommitHandlerFunc(handler.CommitHandler)
 	api.ContainersGetStateHandler = containers.GetStateHandlerFunc(handler.GetStateHandler)
-
+	api.ContainersContainerRemoveHandler = containers.ContainerRemoveHandlerFunc(handler.RemoveContainerHandler)
 	handler.handlerCtx = handlerCtx
 }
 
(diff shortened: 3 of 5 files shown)


matthewavery:
I will squash commits after I am sure all my issues from the /lib move are resolved. right now after my rebase I cannot build because there are issues with import paths and swagger.

lib/apiservers/portlayer/restapi/configure_port_layer.go
```diff
@@ -103,7 +103,7 @@ func configureAPI(api *operations.PortLayerAPI) http.Handler {
 // The middleware configuration is for the handler executors. These do not apply to the swagger.json document.
 // The middleware executes after routing but before authentication, binding and validation
 func setupMiddlewares(handler http.Handler) http.Handler {
-	return handler
+	return curler.New(handler, os.Stdout)
```
hmahmood: What is this about?

lib/portlayer/exec/container.go
```diff
@@ -166,7 +166,8 @@ func (c *Container) Commit(ctx context.Context, sess *session.Session, h *Handle
 // Start starts a container vm with the given params
 func (c *Container) Start(ctx context.Context) error {
 	defer trace.End(trace.Begin("Container.Start"))
-	//no need to grab the lock, there is no state change to the container
+	c.Lock()
```
hmahmood: Why do we need this?

lib/portlayer/exec/container.go
```diff
@@ -220,3 +221,27 @@ func (c *Container) Stop(ctx context.Context) error {
 
 	return nil
 }
+
+func (c *Container) VmRemove(ctx context.Context) error {
```
hmahmood: I would just name this Remove.

lib/apiservers/portlayer/restapi/configure_port_layer.go
matthewavery: ah that was left over from when I was debugging the swagger issue. I was going to ask @sflxn if we should just leave it so that logs in production have some curl output as well. I guess not. it is a debug thing.

lib/portlayer/exec/container.go
```diff
@@ -220,3 +221,27 @@ func (c *Container) Stop(ctx context.Context) error {
...
+	containersLock.Lock()
+	c.Lock()
+	defer containersLock.Lock()
```
fdawg4l: Lock in order, unlock in reverse order.  This seems ripe for deadlocks.

fdawg4l: The `defer` keryword calls the deferred item in LIFO order.  Disregard my comment above.  TIL!!

matthewavery: defer occurs like a stack as far as I know. Which means that the unlocks will occur in the correct order. Unfortunately upong reading this I realized that that defer is actually calling a lock... fixing this now. I just had a conversation about how to move the locks around with hasan so this is changing anyways.

lib/apiservers/engine/backends/container.go
```diff
@@ -245,7 +244,29 @@ func (c *Container) ContainerRestart(name string, seconds int) error {
...
+	}
+
+	//TODO: verify params that are passed in(like -f) and then pass them along.
+
+	//FIXME: currently the portlayer yaml does not support more params than the simple name param
+
+	//call the remove directly on the name. No need for using a handle.
+	_, err := client.Containers.RemoveContainer(containers.NewRemoveContainerParams().WithID(name))
+	if err != nil {
+		fmt.Printf("this is the error that we are getting %#v", err)
```
hmahmood: Debugging leftover?

lib/portlayer/exec/container.go
```diff
@@ -202,6 +203,8 @@ func (c *Container) Start(ctx context.Context) error {
 
 func (c *Container) Stop(ctx context.Context) error {
 	defer trace.End(trace.Begin("Container.Stop"))
+	c.Lock()
+	defer c.Unlock()
 	//no need to grab the lock, there is no state change to the container
```
hmahmood: Comment is not accurate anymore

hmahmood:
lgtm except for the two comments above.

mdubya66:
We are tracking #436 in the GA plan. Is this resolving that?

matthewavery:
it is not resolved #436 it does affect it though since it is the start of that ticket. What is left is just volume detaching and the functionality for force.

lib/apiservers/portlayer/restapi/configure_port_layer.go
```diff
@@ -15,16 +15,14 @@
 package restapi
 
 import (
-	"net/http"
-
 	"golang.org/x/net/context"
+	"net/http"
```
caglar10ur: this chance is failing with goimports. Any reason to have it?

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -210,3 +210,22 @@ func (handler *ContainersHandlersImpl) CommitHandler(params containers.CommitPar
...
 	return containers.NewCommitOK()
 }
+
+func (handler *ContainersHandlersImpl) RemoveContainerHandler(params containers.ContainerRemoveParams) middleware.Responder {
+	defer trace.End(trace.Begin("Containers.RemoveContainerHandler"))
+
+	//get the indicated container for removal
+	cID := exec.ParseID(params.ID)
+	h := exec.GetContainer(cID)
+
```
caglar10ur: remove the extra empty line

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -210,3 +210,22 @@ func (handler *ContainersHandlersImpl) CommitHandler(params containers.CommitPar
...
+	if h == nil {
+		return containers.NewContainerRemoveNotFound()
+	}
+
+	rmErr := h.Container.Remove(context.Background())
```
caglar10ur: s/rmErr/err/g

lib/apiservers/portlayer/swagger.yml
```diff
@@ -440,6 +440,30 @@ paths:
           description: "Error"
           schema:
             $ref: "#/definitions/Error"
+    delete:
+      description: "Remove a container from existence"
+      operationId: ContainerRemove
+      tags: ["containers"]
+      consumes:
+        - application/octet-stream
```
caglar10ur: why is this octet-stream?

lib/portlayer/exec/container.go
```diff
@@ -166,7 +166,8 @@ func (c *Container) Commit(ctx context.Context, sess *session.Session, h *Handle
 // Start starts a container vm with the given params
 func (c *Container) Start(ctx context.Context) error {
 	defer trace.End(trace.Begin("Container.Start"))
-	//no need to grab the lock, there is no state change to the container
```
caglar10ur: what changed here to cause us to grab the lock?

lib/portlayer/exec/container.go
```diff
@@ -202,7 +203,8 @@ func (c *Container) Start(ctx context.Context) error {
 
 func (c *Container) Stop(ctx context.Context) error {
 	defer trace.End(trace.Begin("Container.Stop"))
-	//no need to grab the lock, there is no state change to the container
+	c.Lock()
```
caglar10ur: ^ ditto

lib/apiservers/portlayer/swagger.yml
matthewavery: Swagger requires a consumes function to function properly. it doesnt really use the octet/stream. we do not stream any bytes through the body.

lib/portlayer/exec/container.go
matthewavery: hasan and I talked about the possibility of something in that container getting commit and changing  mid call due to a reconfigure. or if c.vm changed.

lib/apiservers/portlayer/swagger.yml
caglar10ur: can't we use application/json?

lib/apiservers/engine/backends/container.go
```diff
@@ -165,14 +164,13 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 		config.Config.Entrypoint = layer.Config.Entrypoint
 	}
 	config.Config.Env = append(config.Config.Env, layer.Config.Env...)
-
 	log.Printf("config.Config' = %+v", config.Config)
 
 	// Call the Exec port layer to create the container
-	host, err := guest.UUID()
+	host, err := os.Hostname()
```
caglar10ur: I think this is wrong as it reverts @fdawg4l's change (#79df40b586ba5191844068c624814dfde0a2d016)

lib/apiservers/portlayer/swagger.yml
matthewavery: I think we can, @sflxn asked me to add that in with octet-stream. So it should be ok either way.

lib/apiservers/engine/backends/container.go
matthewavery: Thanks for the catch, must have happened during rebase. that is odd.

matthewavery: it is fixed for some reason the PR is not updating... 

please look here.

https://github.com/matthewavery/vic/blob/docker-rm/lib/apiservers/engine/backends/container.go#L509

matthewavery:
All comments should be addressed thus far.

mdubya66:
lgtm

caglar10ur:
pr still shows old code

lib/apiservers/engine/backends/container.go
matthewavery: I noted that I fixed this issue in the wrong place. I will be adding that fix into this PR as well though. another changeset is coming to fix this.

caglar10ur:
lgtm

PR #941, opened 2016-06-01
Persist image metadata to portlayer storage

### LINKED ISSUE #777: Persist image metadata into the child-most layer of said image

jzt:
This refactors imagec to allow for image metadata to be persisted to portlayer storage.

Fixes #777

### DIFF
--- a/cmd/imagec/docker.go
+++ b/cmd/imagec/docker.go
@@ -27,11 +27,17 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 
+	docker "github.com/docker/docker/image"
+	dlayer "github.com/docker/docker/layer"
 	"github.com/docker/docker/pkg/archive"
 	"github.com/docker/docker/pkg/progress"
 
 	"github.com/vmware/vic/pkg/trace"
 )
+
+// DigestSHA256EmptyTar is the canonical sha256 digest of empty tar file -
+// (1024 NULL bytes)
+const DigestSHA256EmptyTar = string(dlayer.DigestSHA256EmptyTar)
 
 // FSLayer is a container struct for BlobSums defined in an image manifest
 type FSLayer struct {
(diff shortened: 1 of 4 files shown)


cmd/imagec/docker.go
```diff
@@ -55,11 +61,14 @@ type Manifest struct {
 
 // V1Compatibility represents some parts of V1Compatibility
 type V1Compatibility struct {
-	ID        string    `json:"id"`
-	Parent    string    `json:"parent,omitempty"`
-	Comment   string    `json:"comment,omitempty"`
-	Created   time.Time `json:"created"`
-	Container string    `json:"container,omitempty"`
+	docker.V1Image
```
caglar10ur: Hmm now that you embedded the whole V1Image I guess the comment is no longer true (it's not some parts of V1Compatibility, it's V1 now :P) and maybe we can rename this struct to something more meaningful .

jzt: @caglar10ur good call, how's `ImageConfig`? I cringed at the thought of putting "metadata" in the name, and docker calls it "ImageConfigJSON" in documentation that I've seen.

cmd/imagec/imagec.go
```diff
@@ -46,6 +46,9 @@ import (
 	"github.com/pkg/profile"
 )
 
+// ImageLayerIndex is the index of the child most layer in the manifest
+const ImageLayerIndex = 0
+
```
caglar10ur: [minor] as far as I see we only use this const once so I suggest simply using 0 in the code below :)

cmd/imagec/imagec.go
```diff
@@ -395,7 +402,10 @@ func CreateImageConfig(images []*ImageWithMeta) error {
 			Comment:   image.Comment,
 		}
 		history = append(history, h)
-		rootFS.DiffIDs = append(rootFS.DiffIDs, dockerLayer.DiffID(layer.diffID))
+		diffID := layer.diffID
+		rootFS.DiffIDs = append(rootFS.DiffIDs, dockerLayer.DiffID(diffID))
+		diffIDs[layer.diffID] = layer.ID
```
caglar10ur: either use diffID here or remove that variable and use layer.diffID on line above

caglar10ur:
minor nitpicks, overall LGTM

cgtexmex:
LGTM

cmd/imagec/imagec.go
jzt: Fixed

cmd/imagec/imagec.go
jzt: Ahh, good catch. That was a remnant of a previous change that didn't get fully refactored. Fixed.

andrewtchin:
lgtm

PR #944, opened 2016-06-01
Grab bag of fixes

fdawg4l:
* Add a comment in the attach GET method.

* WriteImage in the vsphere implementation of the storage interface wasn't checking if the image already existed.  Instead it would attempt to write the image, and the caller would get some obscure vsphere error about scsi attach of some disk address.  Since the image already exists, just move on.

* Move the root of the image store from the hostname of the VCH to the UUID of the vch.  We use the UUID to get the MoRef of the vm the VCH is running on, so it is safe to use this same UUID as the root of our image store.  Also using the UUID (which can be gathered via the `/sys`) means we don't need to infect the docker api server or imagec with unnecessary govmomi calls.  We can easily replace this `UUID()` call to something gathered from `guestinfo`.  But we can punt on this for now.

### DIFF
--- a/cmd/imagec/imagec.go
+++ b/cmd/imagec/imagec.go
@@ -41,6 +41,7 @@ import (
 	"github.com/docker/docker/reference"
 
 	"github.com/vmware/vic/lib/apiservers/portlayer/models"
+	"github.com/vmware/vic/lib/guest"
 	"github.com/vmware/vic/pkg/i18n"
 
 	"github.com/pkg/profile"
@@ -221,7 +222,7 @@ func DestinationDirectory() string {
 }
 
 // ImagesToDownload creates a slice of ImageWithMeta for the images that needs to be downloaded
-func ImagesToDownload(manifest *Manifest, hostname string) ([]*ImageWithMeta, error) {
+func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, error) {
 	images := make([]*ImageWithMeta, len(manifest.FSLayers))
 
 	v1 := docker.V1Image{}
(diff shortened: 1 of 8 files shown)


caglar10ur:
LGTM :sheep: :it:

cgtexmex:
So an image store to VCH is a 1:1 mapping?  I thought there was discussion that an image store could be shared across VCH instances -- or am I missing something with regard to the naming?

fdawg4l:
@cgtexmex, @hickeng can clarify but the idea is to `write` an image to a VCH, there is a 1 to 1 mapping of VCH to image store directory in the datastore.  To `get` (or `read`/`attach`) an image, we'll create a "shared" image store one can put images (via an offline tool or datastore operations) and `get` these images from many VCHs.  That work is TBD.

We will never have 1 instance of the VCH (portlayer) writing to many image stores.  We *will* have many VCHs writing to their own imagestores possibly on the same datastore.  This work provides facilities for the latter.  

Really, all I did here was move from using the hostname of the VCH to the UUID of the VCH.  This stops the collision that was happening when people were starting many VCH instances on the same esx box.

cgtexmex:
ok -- previously didn't realize that the image store name was the hostname of the VCH...just thought it was named VIC to represent a shared image store to multiple VCHs.  So one Image Store many VCH -- reading / writing / sharing that VMDK goodness.

LGTM

fdawg4l:
Ah, sort of.  The path structure should be opaque to the user, but for clarity it's `[datastore] VIC/<UUID>/<image name>/<image name>.vmdk`

cgtexmex:
Thanks for the explanation -- I thought photon was there because it had photon as it's base.....

Issue #746, opened 2016-05-20
Migrate appliance to tether


hickeng:
The tether is used to:
* configure networks
* launch processes with specific arguments and environment
* handle process exit, relaunch, and status reporting

This is the same set of requirements that we have for the appliance, currently handled by the launcher.sh script.

This issue covers:
- [x] update of the appliance iso to package and launch the tether
- [x] update of the tether to handle component relaunch on death
- [ ] consolidation between bootstrap and appliance isos where available
- [ ] launching of the appliance processes via the tether, and status reporting to vic-machine (see portlayer start blocking)
- [ ] trigger of configuration reload via vsphere event on extraconfig (see portlayer start blocking). Depends on #742 
- [ ] update of the tether to handle component configuration change - this one is tentative as I'm unsure which of the implementation approaches is preferable at this time.

hickeng:
This touches on a lot on milestone #747 but not with the same level of structure required. We may find that #747 is partially addressed by this.

hickeng:
Using the tether makes sense, but setting up the serial devices and so on does not - logging is handled differently for the appliance.
This is tracked in #756

dougm:
Opened a PR to support hydrating a ManagedObjectReference to object.Reference with InventoryPath set: https://github.com/vmware/govmomi/pull/527

hickeng:
The core of this is in place in [my branch](https://github.com/hickeng/vic/tree/746) - however the remaining items need to be addressed:
- [ ] component relaunch
- [ ] update to state/status reporting
- [ ] unit tests for vic-machine changes
- [ ] testing of vic-machine against vCenter

hickeng:
This has taken significantly longer than anticipated due to having to complete additional work needed in vic-machine (see #665, #666, #686)

hickeng:
The update to status/state from the less concrete started/exitstatus values will not go in in with this fix.

PR #946, opened 2016-06-02
Separation of tether into lib and cmd

### LINKED ISSUE #756: Split tether into library and mains
Splitting the tether out into library with different mains for:
* containerVM
* appliance
* testing

This is essentially what is being done in the tether tests anyway, but made explicit so that the appliance can consume it with custom config.

hickeng:
This split tether into a lib component that has only the core init
behaviours (lib) and a main package that is specific to it's use in
contianerVMs that has the attach server, backchannel establishment, et al.
In addition this adds vch-init which is an initial placeholder at the cmd
package necessary to use the tether to configure the appliance and launch
its components (746).

Fixes #756


cmd/tether/main_darwin.go
```diff
@@ -69,7 +66,11 @@ func main() {
 		return
 	}
 
-	err = run(src, sink)
+	// create the tether and register the attach extension
+	tthr := tether.New(src, sink, &operations{})
+	tthr.Register("Attach", sshserver)
```
hmahmood: What other "extensions" do you expect to add?

lib/tether/tether_linux.go
```diff
@@ -0,0 +1,184 @@
...
+	if m := d.Mode(); !m.IsDir() && m&0111 != 0 {
+		return nil
+	}
+	return os.ErrPermission
+}
+
+// lookPath searches for an executable binary named file in the directories
+// specified by the path argument.
+// This is a direct modification of the unix os/exec core libary impl
+func lookPath(file string, env []string, dir string) (string, error) {
```
hmahmood: Can we rename this please; `lookPath` is a little odd. How about `findFile` or something similar?

cmd/tether/attach.go
```diff
@@ -328,3 +364,11 @@ func (t *attachServerSSH) channelMux(in <-chan *ssh.Request, process *os.Process
 
 	detach()
 }
+
```
caglar10ur: Why we (and everyone else) keep copying this struct? I know it's just 4 lines and not going to change but it's just keep coming :) Docker one seems to have an exported version so maybe we should use it.

```
[vagrant@devbox:/opt/go/src/github.com/vmware/vic(fix)] grep winsize | grep struct
cmd/tether/tether_linux.go:324:type winsize struct {
vendor/github.com/docker/docker/pkg/term/term.go:27:type Winsize struct {
vendor/github.com/docker/docker/pkg/term/term_windows.go:23:type Winsize struct {
vendor/github.com/docker/libnetwork/Godeps/_workspace/src/github.com/docker/docker/pkg/term/term.go:27:type Winsize struct {
vendor/github.com/docker/libnetwork/Godeps/_workspace/src/github.com/docker/docker/pkg/term/term_windows.go:24:type Winsize struct {
vendor/github.com/go-swagger/go-swagger/vendor/github.com/jessevdk/go-flags/termsize.go:10:type winsize struct {
vendor/github.com/jessevdk/go-flags/termsize.go:10:type winsize struct {
vendor/github.com/kr/pty/util.go:17:type winsize struct {
```

pkg/vsphere/extraconfig/decode.go
```diff
@@ -332,7 +332,7 @@ func decodeMap(src DataSource, dest reflect.Value, prefix string, depth recursio
 	}
```
caglar10ur: Should we merge these two in a different PR as they are fixing issues?

lib/tether/interfaces.go
```diff
@@ -12,37 +12,41 @@
...
-type utilities interface {
-	setup() error
-	cleanup()
-	sessionLogWriter() (dio.DynamicMultiWriter, error)
-	processEnvOS(env []string) []string
-	establishPty(session *SessionConfig) error
-	resizePty(pty uintptr, winSize *attach.WindowChangeMsg) error
-	signalProcess(process *os.Process, sig ssh.Signal) error
-	backchannel(ctx context.Context) (net.Conn, error)
+type Extension interface {
```
fdawg4l: Please add a comment above this and other interfaces about what they are and why they are needed.

lib/tether/interfaces.go
```diff
@@ -12,37 +12,41 @@
...
 	"golang.org/x/net/context"
 )
 
-// ops is here so we can switch the OS Ops impls for test mocking
-var ops osops
-var utils utilities
+type Operations interface {
+	Setup() error
+	Cleanup() error
+	// Log returns the tether debug log writer
```
fdawg4l: + `which writes to the vm directory in the datastore`

cmd/tether/attach.go
```diff
@@ -37,18 +38,55 @@ const (
...
+// Reload - tether.Extension implementation
+func (t *attachServerSSH) Reload(config *tether.ExecutorConfig) error {
+	defer trace.End(trace.Begin("attach reload"))
+
+	t.config = config
+	// process the sessions and launch if needed
+	for id, session := range config.Sessions {
+		log.Debugf("Processing config for session %s", id)
+		if session.Attach {
+			log.Debugf("Session %s is configured for attach", id)
```
fdawg4l: ^^ `Infof`.  Too much debugging info would be a problem I'd welcome.

cmd/tether/attach.go
```diff
@@ -37,18 +38,55 @@ const (
...
+			// this will return nil if already running - calling server.start not t.start so that
+			// test impl gets invoked
+			err := server.start()
+			if err != nil {
+				detail := fmt.Sprintf("unable to start attach server: %s", err)
+				log.Error(detail)
+				return errors.New(detail)
```
fdawg4l: The caller can do the logging in this error condition.

cmd/tether/attach.go
```diff
@@ -37,18 +38,55 @@ const (
...
+
+	t.config = config
+	// process the sessions and launch if needed
+	for id, session := range config.Sessions {
+		log.Debugf("Processing config for session %s", id)
+		if session.Attach {
+			log.Debugf("Session %s is configured for attach", id)
+			// this will return nil if already running - calling server.start not t.start so that
+			// test impl gets invoked
+			err := server.start()
```
fdawg4l: `server` is a global for an `attachServer`, yet in the impl, we refer to the global.  How about just calling `t.start()`.

cmd/tether/attach.go
```diff
@@ -37,18 +38,55 @@ const (
...
+
+	// none of the sessions allows attach, so stop the server - calling server.start not t.start so that
+	// test impl gets invoked
+	server.stop()
+	return nil
+}
+
+// Stop needed for tether.Extensions interface
+func (t *attachServerSSH) Stop() error {
+	return nil
```
fdawg4l: `Stop()` is NOOP?  why not call `t.stop()`?

cmd/tether/attach.go
```diff
@@ -37,18 +38,55 @@ const (
...
+				return errors.New(detail)
+			}
+
+			return nil
+		}
+	}
+
+	// none of the sessions allows attach, so stop the server - calling server.start not t.start so that
+	// test impl gets invoked
+	server.stop()
```
fdawg4l: `s/server/t/g`

cmd/tether/attach_linux.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/vmware/vic/pkg/serial"
+	"github.com/vmware/vic/pkg/trace"
+)
+
+var backchannelMode = os.ModePerm
+
+func backchannel(ctx context.Context, conn *net.Conn) error {
+	defer trace.End(trace.Begin("establish tether backchannel"))
+
+	// HACK: currently RawConn dosn't implement timeout so throttle the spinning
```
fdawg4l: I still don't understand this comment and I think I've read this code more than a dozen times.  I think it's safe to nuke it.

fdawg4l:
My comments were just about missing comments for the most part.  LGTM otherwise.

cmd/tether/main_darwin.go
hickeng: I'm not expecting any currently but I've also not thought about what else could be added. This is here because:
a. it provides a mechanism for dependency injection during testing
b. it means we can use the core of the tether code without needing to implement backchannel and corresponding supporting functions when it's not needed (#746)

lib/tether/tether_linux.go
hickeng: This is called lookPath because the function it's derived from in the standard library is called lookPath and it has identical intent, only it uses an array for input instead of os.Environ

cmd/tether/attach.go
hickeng: The correct solution is to have this in Go standard libaries and exposed as part of the syscall libraries. The docker one is for Windows.

lib/tether/interfaces.go
hickeng: only for the tether when running in an actual container. For tests it goes into /tmp/*/dev/ttyS1 and when used in the appliance it goes to /var/log/vic/init.log

cmd/tether/attach.go
```diff
@@ -37,18 +38,57 @@ const (
...
+				detail := fmt.Sprintf("unable to start attach server: %s", err)
+				log.Error(detail)
+				return errors.New(detail)
+			}
+
+			return nil
+		}
+	}
+
+	// none of the sessions allows attach, so stop the server - calling server.start not t.start so that
```
caglar10ur: calling server.start not t.start -> calling server.stop not t.stop

cmd/tether/attach.go
```diff
@@ -37,18 +38,57 @@ const (
...
+	// test impl gets invoked
+	server.stop()
+	return nil
+}
+
+// Stop needed for tether.Extensions interface
+func (t *attachServerSSH) Stop() error {
+	// calling server.start not t.start so that test impl gets invoked
```
caglar10ur: calling server.start not t.start -> calling server.stop not t.stop

cmd/tether/attach.go
caglar10ur: vendor/github.com/docker/docker/pkg/term/term.go seems have the definitions as well as resize functionality

cmd/tether/main_linux.go
```diff
@@ -97,7 +98,7 @@ func createDevices() error {
 
 	// make an access to urandom
 	path := fmt.Sprintf("%s/urandom", pathPrefix)
-	err = syscall.Mknod(path, syscall.S_IFCHR|uint32(os.FileMode(0444)), Mkdev(1, 9))
+	err = syscall.Mknod(path, syscall.S_IFCHR|uint32(os.FileMode(0444)), tether.Mkdev(1, 9))
```
caglar10ur: should we consider relying on /dev instead of creating our own?

caglar10ur:
minor nitpicks, LGTM otherwise

hickeng:
LGTM

PR #947, opened 2016-06-02
Add vcsim Datastore and Cluster support

### LINKED ISSUE #734: vcsim cluster implementation
We currently have the ability to create Datacenter and Folder objects, along with navigating the inventory and collecting object properties.  Next steps are to further populate the inventory with HostSystem, default ResourcePool, Network, Datastore and ClusterComputeResource types.

dougm:
Fixes #734

### DIFF
--- a/pkg/vsphere/simulator/datastore.go
+++ b/pkg/vsphere/simulator/datastore.go
@@ -1,0 +1,62 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package simulator
+
+import (
+	"os/exec"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/vmware/govmomi/vim25/methods"
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/soap"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+type Datastore struct {
+	mo.Datastore
+}
+
+func (ds *Datastore) RefreshDatastore(*types.RefreshDatastore) soap.HasFault {
+	r := &methods.RefreshDatastoreBody{}
+
+	info := ds.Info.GetDatastoreInfo()
+
+	buf, err := exec.Command("df", "-k", info.Url).Output()
+
+	if err != nil {
+		r.Fault_ = Fault(err.Error(), &types.HostConfigFault{})
+		return r
+	}
+
+	lines := strings.Split(string(buf), "\n")
+	columns := strings.Fields(lines[1])
+
+	used, _ := strconv.ParseInt(columns[2], 10, 64)
+	info.FreeSpace, _ = strconv.ParseInt(columns[3], 10, 64)
+
+	info.FreeSpace *= 1024
+	used *= 1024
+
+	ds.Summary.FreeSpace = info.FreeSpace
+	ds.Summary.Capacity = info.FreeSpace + used
+
+	now := time.Now()
+
+	info.Timestamp = &now
+
+	return r
+}
--- a/pkg/vsphere/simulator/host_datastore_system.go
+++ b/pkg/vsphere/simulator/host_datastore_system.go
@@ -1,0 +1,140 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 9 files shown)


pkg/vsphere/simulator/registry.go
```diff
@@ -105,3 +111,31 @@ func (r *Registry) Remove(item types.ManagedObjectReference) {
...
 }
+
+func (r *Registry) GetEntityParent(item mo.Entity, kind string) mo.Entity {
+	for {
+		parent := item.Entity().Parent
+
+		item = Map.Get(*parent).(mo.Entity)
+
+		if item.Reference().Type == kind {
+			return item
```
hmahmood: Is the only way we exit loop? What if parent == nil?

pkg/vsphere/simulator/host_system.go
```diff
@@ -34,9 +34,24 @@ func NewHostSystem(host mo.HostSystem) *HostSystem {
 	host.Summary.Runtime = &host.Runtime
 	host.Summary.Runtime.BootTime = &now
 
-	return &HostSystem{
+	hs := &HostSystem{
 		HostSystem: host,
 	}
+
+	config := []struct {
```
hmahmood: Only `obj` is used below in the loop; why have this slice then?

pkg/vsphere/simulator/host_datastore_system.go
```diff
@@ -0,0 +1,140 @@
...
+
+	return r
+}
+
+func (dss *HostDatastoreSystem) CreateNasDatastore(c *types.CreateNasDatastore) soap.HasFault {
+	r := &methods.CreateNasDatastoreBody{}
+
+	ds := &Datastore{}
+	ds.Name = path.Base(c.Spec.LocalPath)
+	ds.Self.Value = c.Spec.RemoteHost + ":" + c.Spec.RemotePath
```
hmahmood: Can `RemoteHost` or `RemotePath` be empty?

pkg/vsphere/simulator/registry.go
dougm: Yes, it'll panic if parent == nil.  But that's only possible if there's a bug in the simulator.  User input can't produce such a state.

pkg/vsphere/simulator/host_system.go
dougm: If you look at HostSystem.ConfigManager, there are 36 of these fields, we'll end up populating a few more of them.  The slice is just a table for these mappings so we don't have to duplicate the code inside the for loop x HostSystem.ConfigManager fields we want to populate.

pkg/vsphere/simulator/registry.go
hmahmood: So we are guaranteed that this loop will exit on this line at some point? There is no risk the loop will iterate indefinitely?

pkg/vsphere/simulator/host_datastore_system.go
dougm: It would actually work if they were empty in this case, just a funny looking moref.  Didn't do much validation here, since these methods to create datastores will only be used by the test code.  We'll do more validation in the methods that application code hits, such as CreateVm, CreateResourcePool, etc.

pkg/vsphere/simulator/registry.go
dougm: Yes, it will exit via panic if parent == nil and the root Folder of the inventory always has a nil Parent field.

hmahmood:
lgtm

pkg/vsphere/simulator/datastore.go
```diff
@@ -0,0 +1,62 @@
...
+type Datastore struct {
+	mo.Datastore
+}
+
+func (ds *Datastore) RefreshDatastore(*types.RefreshDatastore) soap.HasFault {
+	r := &methods.RefreshDatastoreBody{}
+
+	info := ds.Info.GetDatastoreInfo()
+
+	buf, err := exec.Command("df", "-k", info.Url).Output()
```
caglar10ur: Maybe something like following instead of parsing df?
```
package main

import (
	"fmt"
	"syscall"
)

const (
	B  = 1
	KB = 1024 * B
	MB = 1024 * KB
	GB = 1024 * MB
)

func main() {
	fs := syscall.Statfs_t{}
	err := syscall.Statfs("/", &fs)
	if err != nil {
		return
	}

	all := fs.Blocks * uint64(fs.Bsize)
	free := fs.Bfree * uint64(fs.Bsize)
	used := all - free

	fmt.Printf("All: %.2f GB\n", float64(all)/float64(GB))
	fmt.Printf("Used: %.2f GB\n", float64(used)/float64(GB))
	fmt.Printf("Free: %.2f GB\n", float64(free)/float64(GB))
}
```

dougm: Yes, I have done that before, in gosigar, avoided it here for now since darwin is a different syscall: https://github.com/cloudfoundry/gosigar/blob/master/sigar_darwin.go#L166

I could also just remove the RefreshDatastore method for now, we don't need it for testing yet, just nice to see with govc datastore.info against vcsim.  Later we could consider using gosigar or something like to fill in some of the other data, such as HostSystem cpu, mem, etc.

dougm: Was actually messing around with it yesterday too: https://github.com/dougm/gosigar/commit/758c241e3e95b63ba11c6f2f1358a3455ef36a4b

caglar10ur:
LGTM :sheep: :it:

dougm:
rebased

PR #959, opened 2016-06-02
Use correct flag in installer

### LINKED ISSUE #956: imagec failing with new insecure flags
The insecure flag in imagec was split into two new flags:

insecure-skip-verify
insecure-allow-http

The dashes in the flag name cause flag.parse to stop at the first dash instance and seeing a flag of "insecure" that no longer exists.

caglar10ur:
Fixes #956


fdawg4l:
LGTM

jzt:
LGTM

Issue #976, opened 2016-06-04
imagec and our metadata design need to support multiple tags per image, including image digests


jzt:
We need a way to store a new tag in the ImageConfig metadata of an existing image. For example, when we already have `busybox:latest` and we pull `busybox:1.24.2` (also the latest busybox at the time of this writing), we should update the tags to reflect that `1.24.2` and `latest` are pointing to the same image.

Currently, imagec will realize that it has all the layers for a particular image and essentially perform a NOP (other than echoing that the images already exist).

**Expected behavior:**

Docker:
```
$ docker pull busybox
Using default tag: latest
latest: Pulling from library/busybox
385e281300cc: Pull complete
a3ed95caeb02: Pull complete
Digest: sha256:4a731fb46adc5cefe3ae374a8b6020fc1b6ad667a279647766e9a3cd89f6fa92
Status: Downloaded newer image for busybox:latest

$ docker pull busybox:1.24.2
1.24.2: Pulling from library/busybox
385e281300cc: Already exists
a3ed95caeb02: Already exists
Digest: sha256:b820bb67cb60307e900f747edfca526fda133cf1c0868c516d9e695ef1f32067
Status: Downloaded newer image for busybox:1.24.2

$ docker images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
busybox             1.24.2              47bcc53f74dc        11 weeks ago        1.113 MB
busybox             latest              47bcc53f74dc        11 weeks ago        1.113 MB
```
**Actual behavior:**

VIC:
```
$ docker $DOCKER_OPTS pull busybox
Using default tag: latest
latest: Pulling from library/busybox
a3ed95caeb02: Pull complete
385e281300cc: Pull complete
Status: Downloaded newer image for library/busybox:latest

$ docker $DOCKER_OPTS pull busybox:1.24.2
1.24.2: Pulling from library/busybox
385e281300cc: Already exists
a3ed95caeb02: Already exists
Status: Image is up to date for library/busybox:1.24.2

$ docker $DOCKER_OPTS images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
busybox             latest              47bcc53f74dc        11 weeks ago        1.311 MB
```

hickeng:
@cgtexmex @jzt Clint, you said you were persisting the tag data along with the image metadata - does that mean that "latest" gets special handling (it does seem more like a synthesized tag to me than the others - but that's possibly because I don't know if other tags can be updated in the same way).
[...]

Issue #975, opened 2016-06-04
Figure out how to calculate image size


jzt:
Since docker has removed layer `Size` attribute from the `v1Compatibility` metadata, we need a way to calculate image size. My first attempt was to calculate each layer size in `imagec/docker.go` during the step that calculates the `diffID`:

```go
	// Copy bytes from decompressed layer into diffIDSum to calculate diffID
	bytesRead, cerr := io.Copy(diffIDSum, tar)
	if cerr != nil {
		return diffID, cerr
	}
```

If the layer being processed represents an empty layer (meaning it has not changed the filesystem at all and thus has no size), I skip it. Otherwise, `bytesRead` becomes the size of the layer.

```go
	if diffID != string(DigestSHA256EmptyTar) {
		// bytesRead represents the size of the decompressed layer data
		image.size = bytesRead
	}
```

However, when summing up the layer sizes in `imagec/imagec.go:CreateImageConfig()`, it results in an incorrect image size:

```go
func CreateImageConfig(images []*ImageWithMeta, manifest *Manifest) error {

	imageLayer := images[0] // the layer that represents the actual image
	image := docker.V1Image{}
	rootFS := docker.NewRootFS()
	history := make([]docker.History, 0, len(images))
	diffIDs := make(map[string]string)
	var size int64

	// step through layers to get command history and diffID from oldest to newest
	for i := len(images) - 1; i >= 0; i-- {
		layer := images[i]
		if err := json.Unmarshal([]byte(layer.meta), &image); err != nil {
			return fmt.Errorf("Failed to unmarshall layer history: %s", err)
		}
		h := docker.History{
			Created:   image.Created,
			Author:    image.Author,
			CreatedBy: strings.Join(image.ContainerConfig.Cmd, " "),
			Comment:   image.Comment,
		}
		history = append(history, h)
		rootFS.DiffIDs = append(rootFS.DiffIDs, dockerLayer.DiffID(layer.diffID))
		diffIDs[layer.diffID] = layer.ID
		size += layer.size
	}
```

Expected:
```
$ docker images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
hello-world         latest              94df4f0ce8a4        5 weeks ago         967 B
busybox             latest              47bcc53f74dc        11 weeks ago        1.113 MB
```

Actual:
```
$ docker -H 192.168.218.219:2376 --tls --tlscert='./vic-docker-cert.pem' --tlskey='./vic-docker-key.pem' images
[...]

PR #977, opened 2016-06-04
Reimplementation of docker list images backend code

### LINKED ISSUE #387: Docker API: Images List
Implement the docker images operation in the engine-api server and integrate with storage port layer.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#list-images
https://docs.docker.com/engine/reference/commandline/images/

jzt:
imagec still needs to update an image's tags when we pull using a different tag for the same image (see #976).

Fixes #387, #975

### DIFF
--- a/cmd/imagec/docker.go
+++ b/cmd/imagec/docker.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"archive/tar"
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
@@ -27,7 +28,6 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 
-	docker "github.com/docker/docker/image"
 	dlayer "github.com/docker/docker/layer"
 	"github.com/docker/docker/pkg/archive"
 	"github.com/docker/docker/pkg/progress"
@@ -57,18 +57,6 @@ type Manifest struct {
 	FSLayers []FSLayer `json:"fsLayers"`
 	History  []History `json:"history"`
 	// ignoring signatures
-}
-
-// ImageConfig contains configuration data describing images and their layers
-type ImageConfig struct {
-	docker.V1Image
-
-	// image specific data
-	ImageID string            `json:"image_id,omitempty"`
-	Tag     string            `json:"tag,omitempty"`
-	Name    string            `json:"name,omitempty"`
-	DiffIDs map[string]string `json:"diff_ids,omitempty"`
-	History []docker.History  `json:"history,omitempty"`
 }
 
 // LearnRegistryURL returns the registry URL after making sure that it responds to queries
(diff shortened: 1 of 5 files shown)


caglar10ur:
LGTM :sheep: :it:

jzt:
Size is now calculated correctly. I'm going to try to get this in now and I'll add support for multiple tags in a followup PR.

cmd/imagec/docker.go
```diff
@@ -269,9 +257,35 @@ func FetchImageBlob(options ImageCOptions, image *ImageWithMeta) (string, error)
...
+				break
+			}
+			if err != nil {
+				return "", err
+			}
+			layerSize += tarHeader.Size
+		}
+
+		image.size = layerSize
 	}
```
caglar10ur: Correct me if I'm wrong but this block uncompresses the tarball and keeps that in memory right?

cgtexmex:
LGTM

cmd/imagec/docker.go
jzt: `archive.DecompressStream` just wraps the compressed data with another Reader. Looking at the `archive/tar` Reader, it doesn't appear to keep the uncompressed tarball in memory. `Next()` skips from file to file using `io.CopyN` from the embedded `Reader` into `ioutil.Discard` (`/dev/null`). It reads actual header data into a 512 byte blocks. 

It would probably be worth having a second set of eyes make sure I'm understanding it correctly.

Code is here:  https://github.com/golang/go/blob/master/src/archive/tar/reader.go#L104

caglar10ur: Yeah but that's tar file itself. I believe tar has a header that contains the original file size (which you use above) but to reach the tar we need to decompress gz part, right? So if we download a 50MB tar.gz file, and if unpacking that ends up creating a 75MB tar file, then we will have 75MB in memory to iterate over its header. 

By the way I'm just trying to understand and not opposing or suggesting something, to be honest I don't even know whether it would be possible to get tar header without decompressing the whole gzip part.

jzt: `DecompressStream` wraps the gzipped `File` in a `gzip.Reader`, so (as I understand Readers & Writers) the bytes are streaming from the compressed file Reader, through `gzip.Reader` where they are decompressed and into the tar Reader, where bytes are streamed into `/dev/null` until the next file header in the tar is hit.

Does that sound crazy? I'm like you.. trying to understand and not confident to assert with 100% certainty that what I suspect is happening is actaully happening. :)

caglar10ur: According to my google-fu you can't decompress a gzip file partially (unless something changed) (http://comp.compression.narkive.com/WVnnqT3M/uncompress-a-partial-gzip-file) and I'll take his word as he is the Adler of Adler-32 as well as the author of zlib/gzip :)

I was/am just curious to learn how it works.

caglar10ur:
LGTM :sheep: :it:

cmd/imagec/docker.go
jzt: That particular case definitely wouldn't work as there is no gzip file header available (it was presumably in backup.tgz.a that he deleted). But in our case, with the header available, wouldn't the bytes be decompressed in a stream starting from the beginning?

caglar10ur: (copying my comment from slack to here for the reference)

Yep, I think you are right. Gzip reader says following;

```
// A Reader is an io.Reader that can be read to retrieve
// uncompressed data from a gzip-format compressed file.
//
// In general, a gzip file can be a concatenation of gzip files,
// each with its own header. Reads from the Reader
// return the concatenation of the uncompressed data of each.
// Only the first header is recorded in the Reader fields.
//
// Gzip files store a length and checksum of the uncompressed data.
// The Reader will return a ErrChecksum when Read
// reaches the end of the uncompressed data if it does not
// have the expected length or checksum. Clients should treat data
// returned by Read as tentative until they receive the io.EOF
// marking the end of the data.
```
and the you wrap that with tar reader which skips anything that is not tar header so at any given time we are bounded by the max size of gzip chunk instead of the total gzip size

jzt:
I updated the `TestFetchImageBlob` unit test to use a tar archive as dummy data.

PR #978, opened 2016-06-05
contains a number of fixes to the swagger runtime

casualjim:
This contains a number backported fixes from go-swagger. This should fix things like deleting without a content-type present etc.
When go-swagger/go-swagger#476 is fixed we can discard those backports, but that will still take a week or 2 to complete.


casualjim:
build passes, i think this is good for merge now

jzt:
lgtm pending @hmahmood 's manual testing

caglar10ur:
Like I said @ https://github.com/vmware/vic/pull/681#issuecomment-224027045 I prefer vendor/ code come from a branch so that we can control and see what's happening. Right now I'm at loss about what we are changing and why...

caglar10ur:
@sflxn convinced me not to create a branch as this is only temporarily. I believed him :) hence here comes LGTM

hmahmood:
LGTM ... start/stop, network ops work, including `network rm` and `docker rm`

casualjim:
SWIM can merge this, I have no write access

Issue #945, opened 2016-06-02
vic-machine `image-store` location used by images and by vms


fdawg4l:
```
Usage of ./vic-machine:
<snip />
  -image-store string
         Image datastore name
<snip />
```

The flag `-image-store` is confusing.  Lets rename this to something like `-image-datastore`.

The datastore specified by this flag defines where images are downloaded to (via `docker pull`) AND where the container VMs are created.  There should likely be another flag to specify what datastore the container VMs should be created on.  If not a second flag, then the flag should be renamed to `-datastore`.

emlin:
@fdawg4l There is another flag: **-container-store**, is meant to store container VMs, but that's not consumed by port-layer yet.
Is that what you want?

btw, will rename both flags to **-datastores.

Issue #897, opened 2016-05-31
vic-machine: delete command processor


emlin:
Target of this task is to create vic-machine delete command, and process the input parameters.
The function to delete resources is not in this scope.

```
vic-machine delete
-target=<VC/ESXi URL>
-user=<user.
-passwd=<password>
-path=mycluster/edw-pool/web/edw-web // VCH resource pool path in govc format
```

emlin:
Target of this task is to support subcommands in vic-machine, to make it support both create and delete commands, and in the future, to support more commands.

We might not leverage govc cli framework directly, cause there is no much value if flags are not reused, and it's unlikely to do that, cause vic-machine and govc should have different customer set.
As @dougm suggested, I'm looking at other CLI frameworks.

https://github.com/urfave/cli is used by docker machine, so it will be easier for us in the future to integrate with docker machine.
And it has all functions required by us right now, for example, custom flag types to support empty password, auto-completion, sub-commands, etc.

Will discuss with @dougm and @hickeng later today to finalize the solution, and likely this task will delay a couple of days.

dougm:
Right, we don't want to adopt govc for just the cli framework itself.  There are several general purpose Go cli frameworks that are much better suited. At some point in the distant future will likely convert the govc cli to use one of them.  Personally I prefer https://github.com/spf13/cobra , used by the docker cli, etcd and kubernetes to name a few.  But urfave/cli (aka codegansta/cli) would be fine too, whichever meets your needs.  I don't want to delay moving forward, just wanted to avoid building new code on the govc cli framework and understand your needs better for a cli framework.

emlin:
We decided go with [github.com/urfave/cli](http://github.com/urfave/cli)

emlin:
Still need two days including today.

Issue #948, opened 2016-06-02
Document vic-machine create, long and short cmds, and updates to image-store flag


stuclem:
Per https://github.com/vmware/vic/issues/945.

stuclem:
Looks like https://github.com/vmware/vic/issues/945 is being worked on right now.

stuclem:
Fix for https://github.com/vmware/vic/issues/945 is in PR https://github.com/vmware/vic/pull/995.

PR #995, opened 2016-06-07
Multiwords vic machine

emlin:
Move vic-machine based on "github.com/urfave/cli", which is simple to use, and support multiple words command.
So vic-machine will follow common command format, to have "--" for long argument, and "-" for short argument.
Also changed "image-store" to full name "image-datastore", "passwd" to "password".

Here is one sample command line: 
```
./bin/vic-machine-linux create --target 10.17.109.50 --user root --password ca\$hc0w --compute-resource /Datacenter/host/VSAN/Resources/line --image-datastore vsanDatastore -name line001 --bridge-network InternalNetwork
 --force 
```
Related to: #897 Fixes: #945


emlin:
@stuclem this change will impact user document, thanks

Makefile
```diff
@@ -307,15 +307,15 @@ $(bootstrap-staging-debug): isos/bootstrap-staging.sh $(iso-base)
 
 $(vic-machine-linux): $(call godeps,cmd/vic-machine/*.go)
 	@echo building vic-machine linux...
-	@GOARCH=amd64 GOOS=linux $(TIME) $(GO) build $(RACE) --ldflags '-extldflags "-static"' -o ./$@ ./$(dir $<)
+	@GOARCH=amd64 GOOS=linux time -f "t%E" $(GO) build $(RACE) -ldflags "-X github.com/vmware/vic/cmd/vic-machine.BuildID=$(BUILD_NUMBER)" -o ./$@ ./$(dir $<)
```
caglar10ur: these changes are wrong, guessing you branch is not up2date

caglar10ur:
It will LGTM after you revert your Makefile changes :)

Makefile
emlin: I made wrong code merge, thanks for find it.

stuclem:
@emlin thanks for the heads-up. I am tracking this in https://github.com/vmware/vic/issues/948.

cmd/vic-machine/create/create.go
```diff
@@ -0,0 +1,398 @@
...
+	if c.computeResourcePath == "" {
+		return cli.NewExitError("--compute-resource Compute resource path must be specified", 1)
+	}
+
+	if c.imageDatastoreName == "" {
+		return cli.NewExitError("--image-datastore Image datastore name must be specified", 1)
+	}
+
+	if c.cert != "" && c.key == "" {
+		return cli.NewExitError("key cert should be specified at the same time", 1)
```
andrewtchin: "key and cert should be specified at the same time"

andrewtchin:
👍 lgtm

dougm:
@emlin you need to update gvt - with the current version, there would only be *.go files added.

We also agreed a while back that vendor/ changes should be in their own PR.

dougm:
Regarding @caglar10ur 's Makefile comment, I think you just need to add '-static' back to the flags.  Or @caglar10ur , did you have an issue with injecting the build number?

Otherwise, lgtm

caglar10ur:
Oh my makefile comment was for something else. @emlin's first revision was reverting the time cmd related changes back to a previous version by mistake. It's no longer the case in the latest revision

emlin:
rebased

PR #998, opened 2016-06-07
Add a whilelist of log files

### LINKED ISSUE #552: vicadmin is not showing all log files
as it gets the list of the files when it starts and then keep using it. New log files that appears in /var/log/vic (like imagec.log) is not visible on the web interface

caglar10ur:
vicadmin does a ioutil.ReadDir to get the file names once it starts.
Since this function gets called once some log files end up missing
as they may not exists at that time.

Current approach works right now because we start vicadmin after
starting docker-server and port-layer but that's because we start
services in alphabetical order.

This minor change simply adds a list hard coded whitelist of log files
that we are interested in. I opted in to this approach thinking that
we should only show known log files in vicadmin interface instead of
everything in that directory.

Fixes #552

### DIFF
--- a/cmd/vicadmin/vicadm.go
+++ b/cmd/vicadmin/vicadm.go
@@ -49,8 +49,16 @@ import (
 )
 
 var (
-	logFileDir = "/var/log/vic/"
-	config     struct {
+	logFileDir  = "/var/log/vic/"
+	logFileList = []string{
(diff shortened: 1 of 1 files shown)


dougm:
lgtm

andrewtchin:
lgtm

caglar10ur:
♻️

PR #1011, opened 2016-06-08
Pass an io.Writer to NewErrorHook

caglar10ur:


fdawg4l:
LGTM

cmd/imagec/hook.go
```diff
@@ -16,25 +16,25 @@ package main
...
 import (
 	"fmt"
-	"os"
+	"io"
 
 	"github.com/Sirupsen/logrus"
 )
 
 // ErrorHook is a simple logrus hook to duplicate Fatalf leveled error messages to stderr
 type ErrorHook struct {
```
fdawg4l: LGTM but do you think we should make this a pkg?  This looks pretty useful.

dougm:
lgtm

caglar10ur:
♻️

cmd/imagec/hook.go
caglar10ur: We can promote it to a pkg if someone else decides to use it, for now it's only used by imagec

PR #1017, opened 2016-06-08
Add support for VSAN in the storage layer

### LINKED ISSUE #710: Milestone: vSAN support
Add support & integration tests for vSAN datastores.

fdawg4l:
Fixes #710

VSAN uses namespaces for top-level directories which themselves are VMFS
filesystems.  The top-level directory is a symlink to the VMFS
(represented by a uuid).  This UUID is the stable path to the namespace
and should be used instead of the directory symlink.  This change uses
that and simplifies the datastore URI convetion so we don't have to
compute the URI format for each datastore or fm govmomi API call.  The
`datastore` struct wraps all of the datastore methods and supplies the
right arguments in the appropriate datastore URI format for the given
vsphere datastore.

Tested against VC + VSAN, ESX + VSAN, and ESX.

### DIFF
--- a/lib/portlayer/storage/vsphere/datastore.go
+++ b/lib/portlayer/storage/vsphere/datastore.go
@@ -1,0 +1,266 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package vsphere
+
+import (
+	"io"
+	"path"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/soap"
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/pkg/vsphere/session"
+	"github.com/vmware/vic/pkg/vsphere/tasks"
+	"golang.org/x/net/context"
+)
+
+// Datastore gives access to the datastore regardless of type (esx, esx + vc,
+// or esx + vc + vsan).  Also wraps paths to a given root directory
+type datastore struct {
+	// The Datastore API likes everything in "path/to/thing" format.
+	ds *object.Datastore
+
+	s *session.Session
+
+	// The FileManager API likes everything in "[dsname] path/to/thing" format.
+	fm *object.FileManager
+
+	// the name of the datastore path (in case of vsan, this is the human readable name)
+	rootname string
+
+	// The datastore url (including root)
+	rooturl string
+
+	// datastore path (not in url form)
+	rootdir string
+}
+
+// newDatastore returns a Datastore.
+// ctx is a context,
+// s is an authenticated session
+// ds is the vsphere datastore
+// root is the top level directory to root all data.  If root does not exist,
+// it will be created.  If it already exists, NOOP. This cannot be empty.
+func newDatastore(ctx context.Context, s *session.Session, ds *object.Datastore, root string) (*datastore, error) {
+	d := &datastore{
+		ds:       ds,
+		s:        s,
+		rootname: root,
+		fm:       object.NewFileManager(s.Vim25()),
+	}
+
+	if err := d.mkRootDir(ctx, root); err != nil {
+		log.Infof("error creating root directory %s: %s", root, err)
+		return nil, err
+	}
+
+	log.Infof("Datastore path is %s", d.rooturl)
+	return d, nil
+}
+
+func (d *datastore) Summary(ctx context.Context) (*types.DatastoreSummary, error) {
+
+	var mds mo.Datastore
+	if err := d.ds.Properties(ctx, d.ds.Reference(), []string{"info", "summary"}, &mds); err != nil {
+		return nil, err
+	}
+
+	return &mds.Summary, nil
+}
+
+// Mkdir creates directories.
+func (d *datastore) Mkdir(ctx context.Context, createParentDirectories bool, dirs ...string) (string, error) {
+	// XXX TODO check if exists
+
+	upth := path.Join(dirs...)
+	upth = path.Join(d.rooturl, upth)
+
+	log.Infof("Creating directory %s", upth)
+
+	if err := d.fm.MakeDirectory(ctx, upth, d.s.Datacenter, createParentDirectories); err != nil {
+		return "", err
+	}
+
+	return upth, nil
+}
+
+// Ls returns a list of dirents at the given path (relative to root)
+//
+// A note aboutpaths and the datastore browser.
+// None of these work paths work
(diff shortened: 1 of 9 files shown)


fdawg4l:
Once I can get this to work against VC, I'll remove the WIP tag.  I'm having trouble with `vic-machine` against VC w/ networking.  Will update when I have more details.

fdawg4l:
Got `vic-machine-linux` to play along.  Pulled busybox, created a VM, and started the VM.  The VM got stuck and it's unclear where but the vmkd config looks fine and it references the image we pulled.  I'm calling that success.

lib/portlayer/storage/vsphere/parent.go
```diff
@@ -48,29 +42,20 @@ const parentMFile = "parentMap"
...
 // Starts here.  Tries to create a new parentM or load an existing one.
-func restoreParentMap(ctx context.Context, s *session.Session) (*parentM, error) {
-
-	// Use the UUID of VCH as the root in the datastore.
-	host, err := guest.UUID()
-	if err != nil {
-		return nil, err
-	}
-
+// basepath is the directory path (not in url form) to root the map
```
andrewtchin: Is basepath supposed to be an arg or where does this refer to?

andrewtchin:
⭐ for commented code

lib/portlayer/storage/vsphere/datastore.go
```diff
@@ -0,0 +1,243 @@
...
+	var mds mo.Datastore
+	if err := d.ds.Properties(ctx, d.ds.Reference(), []string{"info", "summary"}, &mds); err != nil {
+		return nil, err
+	}
+
+	return &mds.Summary, nil
+}
+
+// Mkdir creates directories.  The force flag creates the dir even if it exists. Returns the datastore url of the created directory
+func (d *datastore) Mkdir(ctx context.Context, force bool, dirs ...string) (string, error) {
```
andrewtchin: is `force` supposed to be used in this func?

fdawg4l: Good catch!  It should be `createParentDirectories` (not `force`) and it should be passed through to the `fm.MakeDirectory` call.  Will fix.

lib/portlayer/storage/vsphere/datastore.go
```diff
@@ -0,0 +1,243 @@
...
+	// swallow the error if it's already there.
+	if d.IsVSAN(ctx) {
+		nm := object.NewDatastoreNamespaceManager(d.s.Vim25())
+
+		// This returns the vmfs path (including the datastore and directory
+		// UUIDs).  Use the directory UUID in future operations because it is
+		// the stable path which we can use regardless of vsan state.
+		uuid, err := nm.CreateDirectory(ctx, d.ds, rootdir, "")
+
+		if err != nil {
```
fdawg4l: I missed this.  This should check if it's a `SoapFault` before descending.  Will fix.

lib/portlayer/storage/vsphere/parent.go
fdawg4l: Stale comment.  It used to be a param, but I found later that I didn't need it since we're passing in a *datastore to this.  It itself has a `rootdir` so `basepath` isn't necessary.  Will remove.

pkg/vsphere/disk/disk_manager_test.go
```diff
@@ -19,21 +19,41 @@ import (
...
+		Service: env.URL(t),
+
+		DatastorePath: env.DS(t),
+
+		Insecure:  true,
+		Keepalive: time.Duration(5) * time.Minute,
+	}
+
+	s, err := session.NewSession(config).Create(ctx)
+	if err != nil || s.IsVSAN(ctx) {
```
fdawg4l: This should have a big fat comment as to why we're skipping this test if it's VSAN.  Will add.  The TL;DR is it's too complicated to make it work on vsan without the `datastore` struct doing all of the URI mangling.  And importing that package leads to a circular dependency.

lib/portlayer/storage/vsphere/datastore.go
```diff
@@ -0,0 +1,243 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	task, err := b.SearchDatastore(ctx, path.Join(d.rooturl, p), &spec)
+	if err != nil {
+		return nil, err
+	}
+
+	info, err := task.WaitForResult(ctx, nil)
```
caglar10ur: you could potentially use https://github.com/vmware/vic/tree/master/pkg/vsphere/tasks

lib/portlayer/storage/vsphere/datastore.go
```diff
@@ -0,0 +1,243 @@
...
+	"github.com/vmware/govmomi/vim25/soap"
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/pkg/vsphere/session"
+	"github.com/vmware/vic/pkg/vsphere/tasks"
+	"golang.org/x/net/context"
+)
+
+// Datastore gives access to the datastore regardless of type (esx, esx + vc,
+// or esx + vc + vsan).  Also wraps paths to a given root directory
+type datastore struct {
```
caglar10ur: Why we are not exporting the type (but have exported functions on it) ? Just asking, no objections :)

lib/portlayer/storage/vsphere/datastore_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+	"github.com/vmware/vic/pkg/vsphere/session"
+	"github.com/vmware/vic/pkg/vsphere/tasks"
+	"github.com/vmware/vic/pkg/vsphere/test/env"
+	"golang.org/x/net/context"
+)
+
+func Session(ctx context.Context, t *testing.T) *session.Session {
```
caglar10ur: should we consider making pkg/vsphere/test to handle this config too?

caglar10ur:
LGTM :sheep: :it:

andrewtchin:
lgtm brah

lib/portlayer/storage/vsphere/datastore.go
fdawg4l: Oh Good call!  Will update.

lib/portlayer/storage/vsphere/datastore.go
fdawg4l: Yeah, i struggled with this a bit.  I didn't want this to be confused with the govmomi `object.Datastore` and I didn't want to export it to be used outside this package.  So, I made the struct private to the package, made the `new` method private, and only use the exported functions outside this file (except for the 1 place which I commented).  

Yeah TL;DR-  i just didnt want to export this since it doesn't do much and was painful to get right.

lib/portlayer/storage/vsphere/datastore_test.go
fdawg4l: I tried using the `test` package but none of the default inventory paths (i.e datacenter name, network name, etc.) mapped to the environment I was testing against (VC with non standard cluster, datacenter, and datastore names).  And apparently calling `Create()` validates all of those entries so just making my own Session seemed logical.  I think going forward we can replace all the inventory paths in `test.Session` with params extracted from environment variables or some default.  This was a quick hack.

dougm:
lgtm

lib/portlayer/storage/vsphere/datastore.go
fdawg4l: Looks like we can't use this here because we need to check if the error type asserts to a `types.FileNotFound` govmomi error.  The `tasks` package returns the result of `errors.Errorf()` which fails the assertion.

fdawg4l:
:facepunch:

PR #1029, opened 2016-06-09
Release network resources on container stop

### LINKED ISSUE #992: docker rm and stop do not release network
Currently `docker rm` and `docker stop` do not do a network disconnect and unbind respectively. This results in `network rm` not being able to delete a user defined network.

hmahmood:
Fixes #992


lib/portlayer/network/context.go
```diff
@@ -391,49 +395,60 @@ func (c *Context) Scopes(idName *string) ([]*Scope, error) {
...
+		s, ok := c.scopes[ne.Network.Name]
+		if !ok {
+			return nil, &ResourceNotFoundError{}
 		}
 
-		for _, s := range bound {
-			s.unbindContainer(con)
-		}
-	}()
+		defer func() {
```
caglar10ur: Why do we have this function?

hmahmood: To roll back the `addContainer` calls (below) that have happened to date in case of error.

caglar10ur: Ahh I thought 
```
defer func() {
		if err == nil {
 				return
 			}
}
```

is the whole function...

caglar10ur:
LGTM :sheep: :it:

matthewavery:
LGTM

hmahmood:
lgtm

Issue #1050, opened 2016-06-10
vic-machine: no detail message while usage error happens


emlin:
For example if flag --password is specified incorrectly, the error message is as following:
```
line@ubuntu:/home/line2/gopath/src/github.com/vmware/vic$ ./bin/vic-machine-linux create --passwd
Incorrect Usage.

NAME:
   vic-machine-linux create - Deploy VCH

USAGE:
   vic-machine-linux create [command options] [arguments...]

OPTIONS:
   --target value               ESXi or vCenter FQDN or IPv4 address
   --user value                 ESX or vCenter user
   --password value, -p value   ESX or vCenter password (default: <nil>)
   --cert value                 Virtual Container Host x509 certificate file
   --key value                  Virtual Container Host private key file
   --compute-resource value     Compute resource path, e.g. /ha-datacenter/host/myCluster/Resources/myRP
   --image-datastore value      Image datastore name
   --container-datastore value  Container datastore name - defaults to image datastore
   --name value                 The name of the Virtual Container Host (default: "docker-appliance")
   --external-network value     The external network (can see hub.docker.com)
   --management-network value   The management network (can see target)
   --bridge-network value       The bridge network
   --appliance-iso value        The appliance iso
   --bootstrap-iso value        The bootstrap iso
   --force                      Force the install, removing existing if present
   --generate-cert              Generate certificate for Virtual Container Host
   --timeout value              Time to wait for appliance initialization (default: 3m0s)
   --appliance-memory value     Memory for the appliance VM, in MB (default: 2048)
   --appliance-cpu value        vCPUs for the appliance VM (default: 1)
   
```
Without correct information for which flag is wrong. 
For this error, should include message like following:
" flag provided but not defined: -passwd"

PR #1051, opened 2016-06-10
Print error message from command execution

emlin:
urface/cli does not print out usage error in the general usage error message, but will return the detail message to caller. So modify main function to print out detailed message.

Fixes: #1050


caglar10ur:
LGTM

andrewtchin:
lgtm

emlin:
rebased

PR #1074, opened 2016-06-11
testing the new ci system and cleaning up some files

mdubya66:
https://ci2.vmware.run is setup and will start taking the CI builds. Eventually it will switch back to ci.vmware.run.

This PR is cleaning out 2 lines at the end of the contrib file while also kicking off a testci build


mdubya66:
Closing this as the test failed. @mhagen-vmware  is taking it on

PR #1081, opened 2016-06-13
Switch back to the original URLs

mhagen-vmware:
We went back to the original URLs so need to re-encode the .sec files again.


emlin:
Just to confirm, where did you switch back the URLs? Cause in this PR, you only changed the sec file.

mhagen-vmware:
When you encode the .sec files, you need to specify the URLs.  It doesn't show up in any file change, just in the procedure to encode the file.

emlin:
LGTM

andrewtchin:
lgtm

PR #1087, opened 2016-06-13
Implement ContainerRestart

### LINKED ISSUE #805: Support restart of containers via docker restart
`docker restart` should restart the container vm.

(I don't know if this should be an "epic".  Please modify as you see fit.)

caglar10ur:
Restart is a stop followed by start without touching the network state.

Fixes #805

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -51,11 +51,14 @@ import (
 	"github.com/vmware/vic/pkg/trace"
 )
 
+// Container struct represents the Container
 type Container struct {
 	ProductName string
 }
 
-const attachRequestTimeout time.Duration = 2 * time.Hour
+const (
+	attachRequestTimeout time.Duration = 2 * time.Hour
+)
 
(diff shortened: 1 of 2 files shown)


dougm:
lgtm

fdawg4l:
lgtm

hmahmood:
I would prefer you make bind idempotent ... look at https://github.com/vmware/vic/blob/master/lib/portlayer/network/context.go#L418 and

caglar10ur:
@hmahmood do you suggest adding restart as new state and returning early from bind/unbound when Handle.State is Restart? I thought about doing that but adding restart as a state sounded strange to me.

hmahmood:
No I was suggesting make bind and unbind be idempotent always. So the line I referenced above, instead of returning an error we would just log a warning maybe, and return nil.

hmahmood:
Idempotent bind and unbind will be tackled in a separate PR. LGTM

caglar10ur:
♻️

caglar10ur:
♻️

PR #1095, opened 2016-06-14
Fix vic-machine create example in README

gigawhitlocks:
I just noticed that the example for `vic-machine create` in the README had not been updated since the `create` subcommand was added. This change corrects that.


andrewtchin:
lgtm 🍰

PR #1097, opened 2016-06-14
Add support for adding NICs attached to distributed port groups

### LINKED ISSUE #1002: Distributed vSwitch support in portlayer
The network backing info for a VM is supposed to be the govmomi structure, not the string name - the latter may work for a regular vSwitch, but a DVS requires a richer structure. It results in the following additional entries for the vnic in the vmx file:
ethernet1.dvs.switchId = "aa bb cc dd ..."
ethernet1.dvs.portId = "1"
ethernet1.dvs.portgroupId = "dvportgroup-1"
ethernet1.dvs.connectionId = "123456789"

hmahmood:
Fixes #1002

### DIFF
--- a/lib/metadata/virtual_container_host.go
+++ b/lib/metadata/virtual_container_host.go
@@ -112,7 +112,7 @@ type VirtualContainerHostConfigSpec struct {
 
 	// FIXME: remove following attributes after change to launch through tether
 	// Networks represents mapping between nic name and network info object. For example: bridge: vmomi object
-	Networks map[string]*NetworkInfo `vic:"0.1" scope:"read-only" key:"networks"`
+	Networks map[string]*NetworkInfo `vic:"0.1" scope:"read-only" key:"networks2"`
 
 	ImageFiles []string `vic:"0.1" scope:"read-only" recurse:"depth=0"`
 }
--- a/lib/portlayer/portlayer.go
+++ b/lib/portlayer/portlayer.go
@@ -14,8 +14,41 @@
 
 package portlayer
 
-import "github.com/vmware/vic/lib/portlayer/storage"
+import (
+	log "github.com/Sirupsen/logrus"
+	"github.com/vmware/govmomi/find"
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/vic/lib/metadata"
+	"github.com/vmware/vic/lib/portlayer/storage"
+	"github.com/vmware/vic/pkg/vsphere/extraconfig"
+	"github.com/vmware/vic/pkg/vsphere/session"
+	"golang.org/x/net/context"
+)
+
+var Config metadata.VirtualContainerHostConfigSpec
 
 type API interface {
 	storage.ImageStorer
+}
+
+func Init(ctx context.Context, sess *session.Session) error {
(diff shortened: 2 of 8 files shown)


lib/metadata/virtual_container_host.go
```diff
@@ -112,7 +112,7 @@ type VirtualContainerHostConfigSpec struct {
 
 	// FIXME: remove following attributes after change to launch through tether
 	// Networks represents mapping between nic name and network info object. For example: bridge: vmomi object
-	Networks map[string]*NetworkInfo `vic:"0.1" scope:"read-only" key:"networks"`
+	Networks map[string]*NetworkInfo `vic:"0.1" scope:"read-only" key:"networks2"`
```
caglar10ur: Why are we doing this?

hmahmood: Currently there is a conflict with key names, with the old key names that vic-machine is using conflict with this one. This is just a workaround for now; the old keys in vic-machine will be discarded eventually once we have the tether running appliance finished.

lib/portlayer/portlayer.go
```diff
@@ -14,8 +14,41 @@
...
+
+func Init(ctx context.Context, sess *session.Session) error {
+	source, err := extraconfig.GuestInfoSource()
+	if err != nil {
+		return err
+	}
+
+	extraconfig.DecodeWithPrefix(source, &Config, "guestinfo.vch")
+	log.Debugf("decoded vch config: %#v", Config)
+	f := find.NewFinder(sess.Vim25(), false)
```
caglar10ur: Should we add this type of finder into session for common usage?

hmahmood: Not sure what you mean. In any case, I don't want to touch session until @corrieb is done with his portlayer redo.

lib/portlayer/portlayer.go
```diff
@@ -14,8 +14,41 @@
...
 	storage.ImageStorer
 }
+
+func Init(ctx context.Context, sess *session.Session) error {
+	source, err := extraconfig.GuestInfoSource()
+	if err != nil {
+		return err
+	}
+
+	extraconfig.DecodeWithPrefix(source, &Config, "guestinfo.vch")
```
caglar10ur: Should we check the return code?

lib/portlayer/portlayer.go
caglar10ur: https://github.com/vmware/vic/blob/master/pkg/vsphere/session/session.go#L85 is a finder instance but its 2nd argument is set to true (aka all = true). I was just suggesting adding another Finder (like FinderNoRecurse) into session and using it instead of creating one. I'm fine with both ways.

caglar10ur:
LGTM - just minor nitpicks

lib/portlayer/portlayer.go
hmahmood: There is none; it just returns the decoded object.

caglar10ur: Hmm can we compare it against empty inferface? We are using Config.Networks just a line below and I think it will panic if decode fails for some reason.

emlin:
LGTM

andrewtchin:
lgtm

PR #1107, opened 2016-06-15
Update README for the devbox

corrieb:
I had a pull request open, but it got closed. Presumably because we deleted the branch. Still learning the process here fellas. This is a v simple change...

Fixes #1105 
[skip-ci]

### DIFF
--- a/infra/machines/devbox/README.md
+++ b/infra/machines/devbox/README.md
@@ -2,7 +2,7 @@
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


dougm:
lgtm

Regarding nfs-common, it was included in the original 15.04 base box we were using iirc.  Probably worth adding nfs-common to the list of packages in provision.sh

hickeng:
LGTM

PR #1111, opened 2016-06-15
Implement layer caching in docker persona

### LINKED ISSUE #778: Update docker personality image backend to accommodate new caching structure
The docker images backend needs to be brought inline with the new caching structure we are using to support docker 1.11 behaviors. Depends on #777 .

jzt:
This implements a layer cache on the docker persona. Any time the image store is mutated, a call to `Update()` is made, which will add new layers to the cache (and eventually remove them once we implement `docker rmi`. When `docker images` is called, this cache is accessed within the persona to build a list of images to present to the docker client.

I started to add all the typical methods like `Get()`, `Delete()`, etc, but realized none of those can sensibly be called in the persona because there is no knowledge about the layer(s) available to the persona without specifically calling the portlayer's `ListImages()` function to get a full list.

Looking for feedback on the initial approach, so tagging @corrieb, @hickeng, @cgtexmex, @sflxn, @caglar10ur.

Fixes #778

### DIFF
--- a/lib/apiservers/engine/backends/image.go
+++ b/lib/apiservers/engine/backends/image.go
@@ -15,16 +15,13 @@
 package vicbackends
 
 import (
-	"encoding/json"
 	"fmt"
 	"io"
-	"net/http"
 	"os"
 	"os/exec"
 	"sort"
 
 	log "github.com/Sirupsen/logrus"
-	derr "github.com/docker/docker/errors"
 
 	"golang.org/x/net/context"
 
@@ -32,10 +29,7 @@ import (
 	"github.com/docker/engine-api/types"
 	"github.com/docker/engine-api/types/container"
 	"github.com/docker/engine-api/types/registry"
-	"github.com/vmware/vic/lib/apiservers/portlayer/client/storage"
 	"github.com/vmware/vic/lib/metadata"
-
-	"github.com/vmware/vic/lib/guest"
 )
 
 // byCreated is a temporary type used to sort a list of images by creation
@@ -68,53 +62,18 @@ func (i *Image) ImageHistory(imageName string) ([]*types.ImageHistory, error) {
 
 func (i *Image) Images(filterArgs string, filter string, all bool) ([]*types.Image, error) {
 
-	result := []*types.Image{}
+	imageCache := ImageCache()
+	images := imageCache.GetImages()
+	result := make([]*types.Image, 0, len(images))
 
-	host, err := guest.UUID()
-	if host == "" {
-		host, err = os.Hostname()
-	}
-	if err != nil {
-		return result,
-			derr.NewErrorWithStatusCode(fmt.Errorf("image.Images got unexpected error getting hostname"),
-				http.StatusInternalServerError)
-	}
-
-	params := storage.NewListImagesParams().WithStoreName(host)
-
-	client := PortLayerClient()
-	if client == nil {
-		return result,
-			derr.NewErrorWithStatusCode(fmt.Errorf("image.Images failed to create a portlayer client"),
-				http.StatusInternalServerError)
-	}
-
-	layers, err := client.Storage.ListImages(params)
-	if err != nil {
-		log.Warning(err)
-		return result, nil
-	}
-
-	// TODO(jzt): avoid this by having portlayer send back only images (not layers)
-	// find actual image layers
-	for _, layer := range layers.Payload {
-		imageConfig := &metadata.ImageConfig{}
-		if err := json.Unmarshal([]byte(layer.Metadata["metaData"]), imageConfig); err != nil {
-			derr.NewErrorWithStatusCode(fmt.Errorf("image.Images failed to create a portlayer client"),
-				http.StatusInternalServerError)
-		}
-
-		// HACK(jzt): we could add a marker to models.Image indicating whether or not it represents an image or a layer
-		if imageConfig.ImageID != "" {
-			// convert to dockert image format
-			result = append(result, convertV1ImageToDockerImage(imageConfig))
-			log.Debugf("Found image %#v", imageConfig)
-		}
+	for _, image := range images {
+		result = append(result, convertV1ImageToDockerImage(image))
 	}
 
 	// sort on creation time
 	sort.Sort(sort.Reverse(byCreated(result)))
 
+	log.Infof("Returning %d elements to docker client", len(result))
 	return result, nil
 }
 
@@ -184,6 +143,8 @@ func (i *Image) PullImage(ctx context.Context, ref reference.Named, metaHeaders 
 		return err
 	}
 
+	client := PortLayerClient()
+	ImageCache().Update(client)
 	return nil
 }
 
--- a/lib/apiservers/engine/backends/vicbackends.go
+++ b/lib/apiservers/engine/backends/vicbackends.go
@@ -15,9 +15,11 @@
 package vicbackends
 
 import (
+	"fmt"
 	"net"
 
 	httptransport "github.com/go-swagger/go-swagger/httpkit/client"
+	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client"
 )
 
@@ -28,6 +30,8 @@ const (
 var (
 	portLayerClient     *client.PortLayer
 	portLayerServerAddr string
+
+	imageCache *cache.ImageCache
 )
 
 func Init(portLayerAddr string) error {
@@ -39,6 +43,13 @@ func Init(portLayerAddr string) error {
 	t := httptransport.New(portLayerAddr, "/", []string{"http"})
 	portLayerClient = client.New(t, nil)
 	portLayerServerAddr = portLayerAddr
+
+	imageCache = cache.NewImageCache()
+	// update the image cache at startup
+	if err := imageCache.Update(portLayerClient); err != nil {
+		return fmt.Errorf("Error refreshing image cache: %s", err)
+	}
+
 	return nil
 }
 
@@ -48,4 +59,8 @@ func PortLayerClient() *client.PortLayer {
 
 func PortLayerServer() string {
 	return portLayerServerAddr
+}
+
+func ImageCache() *cache.ImageCache {
+	return imageCache
 }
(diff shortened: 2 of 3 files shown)


lib/apiservers/engine/backends/layer_cache.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	// attempt to create the image store if it doesn't exist
+	store := &models.ImageStore{Name: host}
+	_, err = client.Storage.CreateImageStore(
+		storage.NewCreateImageStoreParams().WithBody(store),
+	)
+
+	if err != nil {
+		if _, ok := err.(*storage.CreateImageStoreConflict); ok {
```
jzt: I probably should change this to just overwrite existing layers in the cache to cover the case of an image having been updated.

lib/apiservers/engine/backends/layer_cache.go
```diff
@@ -0,0 +1,98 @@
...
+	params := storage.NewListImagesParams().WithStoreName(host)
+
+	layers, err := client.Storage.ListImages(params)
+	if err != nil {
+		return fmt.Errorf("Failed to retrieve image list from portlayer: %s", err)
+	}
+
+	for _, layer := range layers.Payload {
+		// TODO(jzt): This assumes that layer ID and their data will remain consistent going forward
+		if _, ok := c.cache[layer.ID]; !ok {
```
jzt: I probably should change this to just overwrite existing layers in the cache to cover the case of an image having been updated.

caglar10ur:
LGTM :sheep: :it:

we should document that this is not getting persistent so restart or crash will make system in a weird state

lib/apiservers/engine/backends/image.go
```diff
@@ -184,6 +155,7 @@ func (i *Image) PullImage(ctx context.Context, ref reference.Named, metaHeaders
 		return err
 	}
 
+	layerCache.Update()
```
cgtexmex: Why do you update the layer cache on images which is a read function?  Why wouldn't we update images on the pull only?

lib/apiservers/engine/backends/image.go
```diff
@@ -67,44 +64,18 @@ func (i *Image) ImageHistory(imageName string) ([]*types.ImageHistory, error) {
...
-		log.Warning(err)
-		return result, nil
-	}
-
-	// TODO(jzt): avoid this by having portlayer send back only images (not layers)
-	// find actual image layers
-	for _, layer := range layers.Payload {
+	for _, layer := range layerCache.cache {
 		imageConfig := &metadata.ImageConfig{}
 		if err := json.Unmarshal([]byte(layer.Metadata["metaData"]), imageConfig); err != nil {
```
cgtexmex: Why not have an image cache?  Then the marshaling could be done during cache update and not on read...

lib/apiservers/engine/backends/image.go
jzt: That's actually at the bottom of the `PullImages` function (I think your diff collapsed it making it appear like it was in `Images`).

cgtexmex:
A general question -- why not create a package in the backends for the caches / stores to reside?  We could then refactor the container store into that cache / store location?

lib/apiservers/engine/backends/image.go
cgtexmex: Ahh..thanks!

lib/apiservers/engine/backends/vicbackends.go
```diff
@@ -28,6 +29,9 @@ const (
 var (
 	portLayerClient     *client.PortLayer
 	portLayerServerAddr string
+
+	// TODO(jzt): should this be defined in backends/image.go?
+	layerCache = NewLayerCache()
```
sflxn: I would really like to have some kind of a getter function to get this.  Since this file is separate but in the same package, addressing this via direct name might be confusing to some people because they don't know where this var is.  Same as we do with PortLayerClient.

lib/apiservers/engine/backends/image.go
jzt: This is actually a really good reason to have an Image cache. Between this and @sflxn's reasons on slack, I'm convinced. Update incoming.

sflxn:
I think we need to review the CLI and API and see if the personality server ever needs to access the layer info.  I cannot think of an instance, but if there is, then the cache is ok.  If we never need access to the layer, then we should not store layers in the cache.  Instead we should just store image data.  I'll repeat, we should research this need before we remove it.

sflxn:
We should add getter and setter to the cache.  If we lock the cache for update and then allow direct access to the underlying data structure, then the lock is not really doing anything.  Also, we shouldn't assume this cache will always remain in the vicbackends namespace.  If we ever decide to move it, we'll have to add these getters and setters anyway.

jzt:
@sflxn It's so trivial to add back that I think it may be a waste of time to comb through APIs looking for an outlying case that might need it. I'd rather just ditch it for now, and add it back if/when we find we're going to need it.

sflxn:
I would rather we do not cache a dictionary of Swagger based structs.  Why don't we unmarshall the data once into imageconfig and store it in the cache instead?  I see Swagger as part of the transport layer and the personality server needs to use it but shouldn't assume Swagger is there.  I know we have to have Swagger info around, but we need to strive and to keep the Swagger out of the backend.  Although, that is a longer term desire.

jzt:
@sflxn can you comment in the actual code so that I can reply to comments individually rather than trying to address them in this general conversation?

jzt:
@sflxn Agreed with not caching swagger structs. I think that falls in line with @cgtexmex's comment about unmarshalling in the image cache.

lib/apiservers/engine/backends/image.go
sflxn: Yes, I agree with Clint's comments.  Also, my earlier comments below elaborates why we shouldn't cache the blobs around since they're swagger blobs.

lib/apiservers/engine/backends/layer_cache.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+// LayerCache is an in-memory cache of image layers. It is refreshed at startup
+// by a call to the portlayer. It is updated when new layers are pulled or
+// layers are deleted.
+type LayerCache struct {
+	sync.Mutex
+
+	// cache maps layer ID to layer data
+	cache map[string]*models.Image
```
sflxn: We really need getters and setters to this cache.  It's the only way we can lock the cache.  My earlier out non inline comments explains this more and also adds other reasons.

lib/apiservers/engine/backends/image.go
sflxn: Ok, since it doesn't show up on my view either, I will add the comment here.  You're doing a full update of the cache on every pull.  That seems rather excessive.  Can't we just get the metadata for that one image and save it to the cache?

lib/apiservers/engine/backends/vicbackends.go
fdawg4l: I think this instantiation and usage is fine.

lib/apiservers/engine/backends/image.go
jzt: No, that's the problem with imagec, remember? The docker persona knows absolutely nothing about an image it just pulled other than the reference that was used to pull it, and nowhere in the storage API is there support for getting a single layer based on this information.

sflxn: Yeah, I see the problem.  I guess once we have the persistent key/value store, we can optimize this.  Ok, the full update is a must for now.

lib/apiservers/engine/backends/vicbackends.go
sflxn: I'm referring not to the instantiation but to the fact that we're accessing layerCache directly, with the assumption that layerCache is in the same namespace.  When I first looked at the image.go code, I wondered where layerCache was declared.

lib/apiservers/engine/backends/layer_cache.go
jzt: I think I addressed this in the changes I just pushed. Can you take a gander and let me know if this is more in line with what you were thinking?

sflxn:
LGTM.  BTW, I do agree with @cgtexmex that it should be in another package, but we can reorg it later.  Also, apparently Go adherents likes their getters to be named a certain way, ImageCache() instead of GetImageCache(), but I could care less.

lib/apiservers/engine/backends/image_cache.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/vmware/vic/lib/apiservers/portlayer/models"
+	"github.com/vmware/vic/lib/guest"
+	"github.com/vmware/vic/lib/metadata"
+)
+
+// ImageCache is an in-memory cache of image layers. It is refreshed at startup
+// by a call to the portlayer. It is updated when new layers are pulled or
+// layers are deleted.
+type ImageCache struct {
+	sync.Mutex
```
caglar10ur: Don't embed mutex into a struct. This makes ImageCache to have Lock/Unlock methods on it, instead use 

```
type ImageCache struct {
    m sync.Mutex
...
}
```
to keep it internal to the struct

caglar10ur: Alternatively, if you think read is the hot path, you can convert this to RWMutex and grab the write lock in Update and read lock in Get

lib/apiservers/engine/backends/image_cache.go
```diff
@@ -15,43 +15,48 @@
...
 // by a call to the portlayer. It is updated when new layers are pulled or
 // layers are deleted.
-type LayerCache struct {
+type ImageCache struct {
 	sync.Mutex
 
-	// cache maps layer ID to layer data
-	cache map[string]*models.Image
+	// cache maps image ID to image metadata
+	cache map[string]*types.Image
```
sflxn: One comment.  We chatted about it over Slack, but I'll add it here for posterity.  types.Image doesn't have all the data we need.  ImageConfig does, but not types.Image.

caglar10ur:
LGTM :sheep: :it:

jzt:
@cgtexmex, @sflxn: I've addressed all your comments with my last change:
- now caching `ImageConfig` structs instead of `types.Image`
- refactored the cache into its own package
- changed `GetImageCache()` to `ImageCache()` (couldn't do that before because image cache lived in the `vicbackends` package
- Added a getter for images

lib/apiservers/engine/backends/image_cache.go
jzt: Fixed, and thanks for the heads up on `RWMutex`, I learned something new today. :)

jzt:
Rebased, will merge after successful build.

jzt:
♻️

PR #1123, opened 2016-06-16
Move directory for images in datastore to ./images

fdawg4l:
We need to differentiate between volumes and images so we can enumerate
them independently from each other.  Moving them from <imageStore(which
is the vch uuid)>/<image id> to <imagestore>/images/<image id>.


dougm:
lgtm

caglar10ur:
LGTM :sheep: :it:

fdawg4l:
:telephone_receiver:

Issue #809, opened 2016-05-29
Update to govmomi/govc 0.8.0 release


dougm:
Planning a new release of govmomi/govc the week of 6/27 to incorporate the various VIC related enhancements.

dougm:
Putting some time into this yesterday/today, mainly for:

- Changes in support of govcsim

- Starting to test govmomi/govc against govcsim, so there are some misc changes coming to pkg/vsphere/simulator

- Recursive Finder support (traverse into Folders)

- Poking at SDRS/Datastore Cluster (via nimbus)

hickeng:
@dougm are you aware of any significant addition items to go into govmomi for TP3 timeframe (thursday)? If not, can we get this done by 6/28 so we've got a couple of days with it before release.

dougm:
No major changes planned, but a minor one to revert the use of service ticket hostname for datastore access, mentioned in #1157 

I'll get that change in and do a tag+release of govmomi, followed by vendor update in vic (by 6/28)

dougm:
#1230 should be the last code update for vendor/govmomi before making a 0.8.0 tag/release.  This will only change the vic/vendor/manifest so we point to the tagged commit.

PR #1124, opened 2016-06-16
vcsim: Update vendor/ govmomi

dougm:
Pulling in recent vcsim related fixes and enhancements

Towards issue #809

Issue #990, opened 2016-06-07
We are using the bits from old iso images when you redeploy using the same name


hagenmt:
Issue that we saw while trying to prep for the demo:
@hickeng can provide more detail.

Something related to us putting the new isos in docker-appliance_1 and then not using that path when we deploy containers.

hickeng:
This is a dup of #711 - I'll add a note to that specifically dealing with the bootstrap iso image.

Issue #711, opened 2016-05-18
Reference VMs by cannonical paths not assumed paths


hickeng:
The path of a VM is not guaranteed to be [datastore] displayname/displayname.vmx. In the case of vSAN it is *never* that path because the symbolic reference is only a best effort.

The cannonical path for a VM can be determined only _after_ it is created from the config.Files.VmPathName property.

This is particularly relevant when we configure containerVMs to boot from an ISO file that's stored with the applianceVMs files - the current path construction is not always correct.

hickeng:
See pkg/vsphere/vm/vm.go:FolderName

fdawg4l:
@hickeng it looks like we're already getting the `VmPathName` for the path.

```
 54 // FolderName returns the name of the namespace(vsan) or directory(vmfs) that holds the VM
 55 // this equates to the normal directory that contains the vmx file, stripped of any parent path
 56 func (vm *VirtualMachine) FolderName(ctx context.Context) (string, error) {
 57 »···var mvm mo.VirtualMachine
 58 
 59 »···if err := vm.Properties(ctx, vm.Reference(), []string{"runtime.host", "config"}, &mvm); err != nil {
 60 »···»···log.Errorf("Unable to get managed config for VM: %s", err)
 61 »···»···return "", err
 62 »···}
 63 
 64 »···path := path.Base(path.Dir(mvm.Config.Files.VmPathName))
 65 »···if path[0] == '[' {
 66 »···»···path = strings.Split(path, "] ")[1]
 67 »···}
 68 »···return path, nil
 69 }
```

fdawg4l:
Do we need to do anything else?

hickeng:
@fdawg4l this path should also be used to determine the location of bootstrap.iso when creating a VM.

fdawg4l:
@hickeng, looking at #990, I can't make the connection between that and this issue.  Can you (briefly) explain the connection?

hickeng:
@fdawg4l #990 results from the following:
1. create VCH with appliance VM named my-vch in folder `[datastore1] my-vch`
2. 'delete' VCH - the `[datastore1] my-vch` folder will still be there because it's not empty after deleting the VM files - it still contains the two ISO files.
3. create new VCH, again named my-vch. Appliance VM will be in folder `[datastore1] my-vch_1`
4. create a container with the new VCH and it will be booting from `[datastore1] my-vch/bootstrap.iso`, not `[datastore1] my-vch_1/bootstrap.iso`

[...]

PR #1125, opened 2016-06-16
vcsim fixes and enhancements

dougm:
In support of #809 - much of these are related to testing govmomi/govc against vcsim, but also of use with vic.

### DIFF
--- a/pkg/vsphere/simulator/virtual_machine.go
+++ b/pkg/vsphere/simulator/virtual_machine.go
@@ -15,11 +15,15 @@
 package simulator
 
 import (
+	"fmt"
+	"io"
+	"log"
 	"os"
 	"path"
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/methods"
 	"github.com/vmware/govmomi/vim25/mo"
@@ -30,6 +34,8 @@ import (
 
 type VirtualMachine struct {
 	mo.VirtualMachine
+
+	log *log.Logger
 }
 
 func NewVirtualMachine(spec *types.VirtualMachineConfigSpec) (*VirtualMachine, types.BaseMethodFault) {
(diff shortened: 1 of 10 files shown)


dougm:
CI will fail until #1124 (govmomi update) is merged.

caglar10ur:
LGTM :sheep: :it:

pkg/vsphere/simulator/virtual_machine.go
```diff
@@ -129,39 +138,88 @@ func (vm *VirtualMachine) configure(spec *types.VirtualMachineConfigSpec) types.
...
+
+		file = path.Join(file, name)
+	}
 
-	// Assuming local datastore for now.  TODO: should be via datastore browser / file manager
-	dir := ds.Info.GetDatastoreInfo().Url
+	dir := path.Dir(file)
 
-	vmx := path.Join(dir, p.Path)
+	_ = os.MkdirAll(dir, 0700)
```
hickeng: The vSphere behavior is actually to append a suffix if the directory already exists, so MyVM would become MyVM_1 if directory MyVM already existed. This has caused us issues in the past and there's #990 which is a current issue relating to this behaviour

dougm: Currently vcsim will return a FileAlreadyExists fault with either format for Config.Files.VmPathName if the .vmx file already exists:
```
[DS_Name] vm-name
[DS_Name] vm-name/vm-name.vmx
```

Real VC/ESX will only rename in the case of '[DS_Name] vm-name' and will fault with FileAlreadyExists in the latter case '[DS_Name] vm-name/vm-name.vmx' if the file already exists.

I assumed we were already use the latter form (full path to .vmx) and intentionally chose not to support the duplicate name (w/ rename) to avoid this trouble with real VC/ESX.

emlin: But sometimes, after the VM is deleted, the VM folder still exists, if some files are not removed, e.g. the uploaded iso files.
In that case, the simulator will be different, right?

dougm: Yes, in this case I think we want the simulator to be different, to help us avoid the VmPathName auto-rename (see #711).  We could also change the simulator to return FileAlreadyExists here if the directory itself already exists.

hickeng: @dougm Perhaps we make this configurable - I agree that we should try to avoid the rename at all, but it's not completely within our control. Having the simulator deliberately from known behavior won't help us ensure that we correctly handle the case where rename does happen.

hickeng:
LGTM
My other comments should not be considered blocking.

pkg/vsphere/simulator/virtual_machine.go
dougm: Sounds good, we can do both.  I will submit that change as separate PR.

dougm:
rebased

PR #1126, opened 2016-06-17
delete containers VMs together with appliance vm

### LINKED ISSUE #1088: Handle delete of container VMs
vic-machine delete should delete container VMs created/managed by removed VCH

emlin:
Fixes #1088 

vic-machine delete will delete all container VMs in the resource pool, and new created vSwitch in ESX.


caglar10ur:
What happens if I create a vm manually in the same RP? I think this nukes that too. Maybe we should start marking VCH created vms somehow

emlin:
Good question, but right now, we don't have other mechanism to query out VCH created container VMs.
If we put container VM to VCH VM folder, this will be more accurate.
I personally think that user should not operate the resource pool we created for VCH. Maybe change to vApp, this will be better, but for ESXi, vApp is not supported.

hickeng:
The containerVMs are pretty easily identified.... do they have a key `guestinfo./common/id`?
I do think that people shouldn't mix and match and that a vApp will help make that evident... and as @emlin said the Folders will help as well. There's only so much we can do however to protect people from themselves.

cmd/vic-machine/delete/delete.go
```diff
@@ -114,13 +116,36 @@ func (d *Uninstall) Run(cli *cli.Context) error {
...
-	rp, err := validator.GetResourcePool(d.Data)
+	defer func() {
+		if validator.Session != nil {
+			validator.Session.Client.Logout(validator.Context)
+		}
+	}()
+
+	vch, err := validator.GetVCH(d.Data)
+	if err != nil {
+		err = errors.Errorf("%s. Exiting...", err)
```
hickeng: Unless we have  a structured method for marking whether an error is fatal or not, adding "Exiting..." onto the end makes it harder to have clean user focused output.
Consider adding log.Debug messages along with the pass through error check that explain whether the error is fatal or not

cmd/vic-machine/validate/network.go
```diff
@@ -22,36 +22,37 @@ import (
 	"github.com/vmware/vic/pkg/errors"
 )
 
-func (v *Validator) createBridgeNetwork() error {
+func (v *Validator) createBridgeNetwork() (bool, error) {
```
hickeng: Sure you saw but I've moved the network create into the lib portion - that way it's only created once we've actually validated the config for the appliance.

cmd/vic-machine/validate/validator.go
```diff
@@ -119,7 +122,64 @@ func (v *Validator) GetResourcePool(input *data.Data) (*object.ResourcePool, err
...
+			//this is the appliance vm
+			mapConfig, err := vm.FetchExtraConfig(v.Context)
+			if err != nil {
+				err = errors.Errorf("Failed to get VM extra config of %s, %s", vmName, err)
+				log.Errorf("%s", err)
+				return nil, err
+			}
+			data := extraconfig.MapSource(mapConfig)
+			vchConfig := &metadata.VirtualContainerHostConfigSpec{}
+			result := extraconfig.DecodeWithPrefix(data, vchConfig, "guestinfo.vch")
```
hickeng: You'll need to update this to simply `Decode(...)` if it's merged after my PR.
We should also check that vchConfig.ID contains the moref of the VM - it's a good additional check that this is a VCH appliance.

lib/install/management/appliance.go
```diff
@@ -93,6 +77,35 @@ func (d *Dispatcher) removeApplianceIfForced(conf *metadata.VirtualContainerHost
 	return nil
 }
 
+func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine) error {
+	var err error
+	// ignore power off error cause VM can be powered off already
+	if _, err = tasks.WaitForResult(d.ctx, func(ctx context.Context) (tasks.ResultWaiter, error) {
```
hickeng: This is not a fast path, so let's be neat and check the VM power state to avoid errors in the Task view of UIs as much as possible. Not needed for this PR.

lib/install/management/appliance.go
```diff
@@ -93,6 +77,35 @@ func (d *Dispatcher) removeApplianceIfForced(conf *metadata.VirtualContainerHost
...
+		return vm.Destroy(ctx)
+	})
+	if err != nil {
+		err = errors.Errorf("Failed to destroy vm %s: %s", vm.Reference(), err)
+		return err
+	}
+	return nil
+}
+
+func (d *Dispatcher) deleteDatastoreFiles(conf *metadata.VirtualContainerHostConfigSpec) error {
```
hickeng: We should log the full `[ds] path` of the directory we're attempting to delete so that the viadmin knows where to do manual cleanup.

lib/install/management/dispatcher.go
```diff
@@ -245,3 +254,47 @@ func (d *Dispatcher) CollectDiagnosticLogs() {
...
+	var errs []string
+
+	var err error
+	var vmm *vm.VirtualMachine
+	var children []*vm.VirtualMachine
+
+	if vmm, err = d.findAppliance(conf); err != nil {
+		return err
+	}
+	if vmm == nil {
```
hickeng: if we cannot find the appliance we should continue deleting the VMs we can find if -force is specified.

lib/install/management/dispatcher.go
```diff
@@ -245,3 +254,47 @@ func (d *Dispatcher) CollectDiagnosticLogs() {
...
+	}
+
+	for _, child := range children {
+		if err = d.deleteVM(child); err != nil {
+			errs = append(errs, err.Error())
+		}
+	}
+
+	if len(errs) > 0 {
+		log.Debugf("Error to delete container VMs %s", errs)
```
hickeng: "Error deleting container VMs", or "Error while deleting container VMs"

lib/install/management/network.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/vmware/vic/lib/metadata"
+)
+
+func (d *Dispatcher) removeNetwork(conf *metadata.VirtualContainerHostConfigSpec) error {
+	if !conf.CreateBridgeNetwork {
+		log.Infof("Bridge network is not created for this VCH, leave it there.")
```
hickeng: "Bridge network _was_ not created for this VCH"

lib/install/management/network.go
```diff
@@ -0,0 +1,49 @@
...
+		return nil
+	}
+
+	if d.session.Host == nil {
```
hickeng: All VC checks should be d.Session.IsVC()

lib/install/management/network.go
```diff
@@ -0,0 +1,49 @@
...
+		log.Warnf("Remove network is not supported in VC")
+		return nil
+	}
+	log.Infof("Removing Portgroup")
```
hickeng: Add the name of the switch in here so that if it fails a viadmin knows what to manually delete.

lib/install/management/resource_pool.go
```diff
@@ -79,18 +80,25 @@ func (d *Dispatcher) createResourcePool(conf *metadata.VirtualContainerHostConfi
 	return vrp, nil
 }
 
-func (d *Dispatcher) destroyResourcePool(conf *metadata.VirtualContainerHostConfigSpec) error {
-	log.Infof("Destroying the Resource Pool")
+func (d *Dispatcher) destroyResourcePoolIfEmpty(conf *metadata.VirtualContainerHostConfigSpec) error {
+	log.Infof("Removing Resource Pool")
```
hickeng: Let's have the name of the pool as with switch and pg, given it's the outer edge of the VCH from a compute/folder hierarchy perspective.

cmd/vic-machine/validate/validator.go
```diff
@@ -119,7 +122,64 @@ func (v *Validator) GetResourcePool(input *data.Data) (*object.ResourcePool, err
...
+	}
+	return vchPool, nil
+}
+
+func (v *Validator) GetVCHConfig(vch *object.ResourcePool, name string) (*metadata.VirtualContainerHostConfigSpec, error) {
+	var err error
+	var vms []*vm.VirtualMachine
+
+	rp := compute.NewResourcePool(v.Context, v.Session, vch.Reference())
+	if vms, err = rp.GetChildrenVMs(v.Context, v.Session); err != nil {
```
hickeng: Why isn't this using GetChildVM?

pkg/vsphere/tasks/waiter.go
```diff
@@ -77,7 +77,7 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
 			cerr = errors.Errorf("%s - (%s)", cerr, info.Error)
 		}
 
-		log.Errorf(cerr.Error())
+		//		log.Errorf(cerr.Error())
```
hickeng: Please make this debug - we can look into package level log configuration later, but this shouldn't be squashed completely for the other components.

pkg/vsphere/vm/vm.go
```diff
@@ -208,3 +208,15 @@ func (vm *VirtualMachine) WaitForKeyInExtraConfig(ctx context.Context, key strin
 	}
 	return detail, nil
 }
+
+func (vm *VirtualMachine) LoadName(ctx context.Context) (string, error) {
```
hickeng: This should be just `Name()`  - that's the Go convention and is consistent with govmomi

lib/install/management/dispatcher.go
```diff
@@ -245,3 +254,47 @@ func (d *Dispatcher) CollectDiagnosticLogs() {
...
+}
+
+func (d *Dispatcher) DeleteVCHInstances(conf *metadata.VirtualContainerHostConfigSpec) error {
+	var errs []string
+
+	var err error
+	var vmm *vm.VirtualMachine
+	var children []*vm.VirtualMachine
+
+	if vmm, err = d.findAppliance(conf); err != nil {
```
hickeng: This should be `d.VMFromID(id string)` or similar and should use the encoded moref, not the name:
```
moref := new(types.ManagedObjectReference)
ok := moref.FromString(id)
if !ok {
 ...
}
ref, err := v.Session.Finder.ObjectReference(ctx, *moref)
vm, ok := ref.(*object.VirtualMachine)
if !ok {
  ...
}
```

cmd/vic-machine/validate/network.go
emlin: yeah, will rebase with that

cmd/vic-machine/validate/validator.go
emlin: sure

lib/install/management/dispatcher.go
emlin: That means all VMs under the specified resource pool will be removed, even that resource pool is not created for VCH. Is that expected?

cmd/vic-machine/validate/validator.go
emlin: I think we need more test for if that VM is a VCH or not. Instead of just check the VM name.
But with GetChildVM, this check can be done after the VM is found. Will change it.

lib/install/management/dispatcher.go
emlin: Added one method to search appliance VM through ID, instead of name. But will wait your PR, and then switch to that one.
I don't want to have more conflict with you.

hickeng:
@emlin #1127 is merged

lib/install/management/dispatcher.go
hickeng: Apologies - you're entirely correct. We would need a --really-force flag for that :)

We should ensure that the appliance is the last VM deleted in that case - that would mean that if delete fails for part way through deleting children and some are left behind we should still be able to run delete again against that VCH.

emlin:
@hickeng Merged with your change

cmd/vic-machine/validate/validator.go
```diff
@@ -350,6 +354,62 @@ func (v *Validator) network(ctx context.Context, input *data.Data, conf *metadat
...
+	// if the moref of the target matches where we expect to find it for a VCH, run with it
+	if vchConfig.ExecutorConfig.ID != vm.Reference().String() {
+		err = errors.Errorf(
+			"The VCH configuration VM ID %s is inconsistent with appliance VM ID %s",
+			vchConfig.ExecutorConfig.ID, vm.Reference().String())
+		log.Errorf("%s", err)
+		return nil, err
+	}
+
+	vchConfig.ComputeResources = append(vchConfig.ComputeResources, v.Session.Pool.Reference())
```
hickeng: you can call vchConfig.AddComputeResource(...) to add a compute resource. While append handles ComputeResources being nil cleanly, it doesn't work for maps so the functions are there for consistency.

I think `v.Session.Pool.Reference() == vch.Reference()` in which case why are we adding it twice, or v.Session.Pool refers to the parent pool of the VCH, in which case we shouldn't be adding it. Not yet worked through the initial CLI arg parsing so may need to update this when done there.

cmd/vic-machine/validate/validator.go
```diff
@@ -350,6 +354,62 @@ func (v *Validator) network(ctx context.Context, input *data.Data, conf *metadat
 	}
 }
 
+func (v *Validator) GetVCH(input *data.Data) (*object.ResourcePool, error) {
```
hickeng: Validator is not the right place for this function - the only reason to have this in validator is because of the compute resource lookup code which should be moved under pkg/vsphere/ at some point anyway.
Validator could and should expose an IsAppliance(object.VirtualMachine) function that will check assertions about the VM config, but that's it.

This is the some of the same function we'll need for `ls` and `inspect` so let's get it in the right place. I think that's in lib/install/management. These functions can locate an appliance VM if possible, then pass that to validator for checking:

```
NewFromID(moref string) (*management.VCH, error)
NewFromCompute(path string) (*management.VCH, err) // specified by compute resource path
NewFromPath(path string) (*management.VCH, err) // specified by folder path when we add folder support.
```

We can then add functions to management.VCH for:
* inspect
* delete
* update

cmd/vic-machine/validate/validator.go
```diff
@@ -350,6 +354,62 @@ func (v *Validator) network(ctx context.Context, input *data.Data, conf *metadat
...
+	vchConfig := &metadata.VirtualContainerHostConfigSpec{}
+	result := extraconfig.Decode(data, vchConfig)
+	if result == nil {
+		err = errors.Errorf("Failed to decode VM configuration %s, %s", name, err)
+		log.Errorf("%s", err)
+		return nil, err
+	}
+
+	// if the moref of the target matches where we expect to find it for a VCH, run with it
+	if vchConfig.ExecutorConfig.ID != vm.Reference().String() {
```
hickeng: This should differentiate between ID == "" which suggests there's no ExecutorConfig in the ExtraConfig at all, and a mismatch which suggests something else (I'm not sure what).
This should be in validator.IsVCH(...).

cmd/vic-machine/validate/validator.go
```diff
@@ -350,6 +354,62 @@ func (v *Validator) network(ctx context.Context, input *data.Data, conf *metadat
...
+	vchPath := fmt.Sprintf("%s/%s", rp.InventoryPath, input.DisplayName)
+	vchPool, err := v.Session.Finder.ResourcePool(v.Context, vchPath)
+	if err != nil {
+		log.Errorf("Failed to get VCH resource pool %s: %s", vchPath, err)
+		return nil, err
+	}
+	return vchPool, nil
+}
+
+func (v *Validator) GetVCHConfig(vch *object.ResourcePool, name string) (*metadata.VirtualContainerHostConfigSpec, error) {
```
hickeng: This definitely doesn't belong in Validator - I think this would be part of the New... methods in the prior comment.

cmd/vic-machine/validate/validator.go
emlin: This is a mistake during integration with your code. No need to add twice.

emlin:
@hickeng Simple refactor for validator.

cmd/vic-machine/delete/delete.go
```diff
@@ -61,6 +61,12 @@ func (d *Uninstall) Flags() []cli.Flag {
 			Usage:       "Compute resource path, e.g. /ha-datacenter/host/myCluster/Resources/myRP",
 			Destination: &d.ComputeResourcePath,
 		},
+		cli.StringFlag{
+			Name:        "name",
+			Value:       "",
+			Usage:       "The name of the Virtual Container Host",
```
hickeng: "The name of the Virtual Container Host to delete; requires that compute-resource argument be supplied"

cmd/vic-machine/delete/delete.go
```diff
@@ -61,6 +61,12 @@ func (d *Uninstall) Flags() []cli.Flag {
 			Usage:       "Compute resource path, e.g. /ha-datacenter/host/myCluster/Resources/myRP",
```
hickeng: "The compute resource containing the Virtual Container Host; requires the 'name' argument be supplied"

lib/install/management/delete.go
```diff
@@ -0,0 +1,84 @@
...
+func (d *Dispatcher) DeleteVCH(conf *metadata.VirtualContainerHostConfigSpec) error {
+	log.Infof("Removing VMs")
+
+	if err := d.DeleteVCHInstances(conf); err != nil {
+		return err
+	}
+	if err := d.destroyResourcePoolIfEmpty(conf); err != nil {
+		return err
+	}
+	return d.removeNetwork(conf)
```
hickeng: The appliance VM should be the last thing deleted other than the resource pool so that a rerun of delete will function. We should also be allowing --force to override the return on error behaviour

lib/install/management/delete.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	for _, child := range children {
+		name, err := child.Name(d.ctx)
+		if err != nil {
+			errs = append(errs, err.Error())
+			continue
+		}
+		if name == conf.Name {
+			continue
```
hickeng: comment about why we're skipping it

lib/install/management/finder.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/vmware/vic/pkg/vsphere/vm"
+)
+
+func (d *Dispatcher) NewVCHFromID(id string) (*vm.VirtualMachine, error) {
+	var err error
+	var vmm *vm.VirtualMachine
+
+	moref := new(types.ManagedObjectReference)
+	if ok := moref.FromString(id); !ok {
+		message := "Failed to get appliance VM mob reference"
```
hickeng: We need to start moving these into a message file as per the conversation in the engineering meeting yesterday. Not for this PR, but please look at doing so for `ls`. @andrewtchin wrote the catalogue package so can help get that started fast.

lib/install/management/finder.go
```diff
@@ -0,0 +1,131 @@
...
+		log.Errorf(message)
+		return nil, errors.New(message)
+	}
+	ref, err := d.session.Finder.ObjectReference(d.ctx, *moref)
+	if err != nil {
+		if _, ok := err.(*find.NotFoundError); !ok {
+			err = errors.Errorf("Failed to query appliance (%s): %s", moref, err)
+			return nil, err
+		}
+		log.Debugf("Appliance is not found")
```
hickeng: Given this is debug, let's log the ID supplied as well. Can do this with an opening trace call at the top of the function (which we should have for almost all functions):
```
defer trace.End(trace.Begin(id))
```

hickeng:
LGTM
There are comments for changes, but they're all non critical and can go in as a followup change.

emlin:
Updated the message error, and will handle the delete sequence issue together with image/volume delete in next PR.

andrewtchin:
lgtm

Issue #594, opened 2016-04-29
Migrate vic-machine to pkg/vsphere/extraconfig


hickeng:
vic-machine currently has to construct the extraconfig/guestinfo configuration for the VCH through a manual process of mapping data to guestinfo keys.

pkg/vsphere/extraconfig is now sufficiently functional to serialize the primary config structures directly to the ExtraConfig array in a config spec. Until the other _migration_ issues in this epic are completed this cannot work directly, so may entail an update to the launcher.sh script that generates the CLI arguments if not done as a single PR.

emlin:
Not started yet. Need more discussion for it, so move back to TODO

PR #1127, opened 2016-06-17
Experimental tether based appliance

hickeng:
My sincere apologies for the size and scope of this PR - it turned out to be difficult to do vic-machine and tether parts independently and I simply couldn't face splitting out the minor supporting work into separate items.

* Moves vic-machine to use extraconfig completely
* vic-machine -target now takes the datacenter (top level scope) as the path component of the URI
* vic-machine -compute-resource is now relative to datacenter and shouldn't require the internal structure of `host` and `Resources`
* Validation takes human name, translates to moref where appropriate and populates config struct
* Minor change to compute recommendations to avoid insanely long lists in large clusters (some people will have 100s of resource pools)
* Defer creation of network until after validation has completed and appliance is being created
* Slightly better blocking for appliance start status
* IP address reporting via guestinfo as for containerVMs
* vicadmin and docker personality load certs from extraconfig
* network interfaces are renamed and IPs are aliased with role name (client, management, external, bridge)

Outstanding items are in #746 and include:
* process restart behaviours
* finer grained status and state reporting of processes
* many additional tests, particularly around negative cases

This should really be a WIP given the outstanding test work, but it touches too many areas of the code that are currently undergoing work not to put it up for merging now.

Tested again:
* nested local ESX
* @corrieb's vCenter

Towards #746 #594


hickeng:
@hmahmood The most major changes you'll note in the networks side of things are:
* Replacing PCISlot with ID
* Move of findSlot code to spec package for reuse by vic-machine
* Move to StaticIP and AssignedIP for address reporting.

The PCI slot is the most contentious - mostly done because that ID field was/is specifically intended to hold the unique identifier for _thingX_ within _environmentY_ and I wanted to be consistent across structures. That said there's a couple of hoops due to the difference in types for which I apologize.

hickeng:
@emlin There's a swath of changes to vic-machine.
* url target - had to stop using soap.Parse to get viable Path element
* validator - I'd needed almost all of this last week and it was must easier to keep the consistent code than to attempt to blend. The only blending I did was the compute resource recommendation which just has some minor tweaks for large environments.
* dispatcher - moved the network creation to here - conceptually this is where it belongs and it avoids creation of bridge networks in the case where validation fails.
* obviously major changes to appliance given a lot of that relates to configuration of the appliance components. Still more changes to do here once the component migrations are complete.

cmd/vic-machine/validate/validator.go
```diff
@@ -45,13 +48,183 @@ type Validator struct {
...
+	})
+
+	// Management net
+	if input.managementNetworkName != "" {
+		managementMoref, err := v.Network(ctx, input.managementNetworkName)
+		v.NoteIssue(err)
+		conf.AddNetwork(&metadata.NetworkEndpoint{
+			Network: metadata.ContainerNetwork{
+				Common: metadata.Common{
+					Name: "bridge",
```
emlin: Two bridge networks?

cmd/vic-machine/validate/validator.go
```diff
@@ -45,13 +48,183 @@ type Validator struct {
...
+	// Perform the higher level compatibility and consistency checks
+	err = v.CompatibilityChecks(ctx, conf)
+	v.NoteIssue(err)
+
+	// TODO: determine if this is where we should turn the noted issues into message
+	return conf, v.ListIssues()
+}
+
+func (v *Validator) Network(ctx context.Context, path string) (*types.ManagedObjectReference, error) {
+	nets, err := v.Session.Finder.NetworkList(ctx, path)
```
emlin: Why not just use Finder.Network()? It will return error message for not found and multiple found.

lib/install/management/appliance.go
```diff
@@ -147,12 +147,16 @@ func (d *Dispatcher) createApplianceSpec(conf *metadata.VirtualContainerHostConf
 	var devices object.VirtualDeviceList
 	var err error
 
+	cfg := make(map[string]string)
+	extraconfig.Encode(extraconfig.MapSink(cfg), conf)
```
emlin: So change to use extraconfig directly? Is that available through rpctool inside guest os?

lib/install/management/appliance.go
```diff
@@ -394,14 +398,8 @@ func (d *Dispatcher) reconfigureApplianceSpec(vm *vm.VirtualMachine, conf *metad
 	spec.ExtraConfig = d.getPresetExtraconfig(conf)
 
 	cfg := make(map[string]string)
-	extraconfig.EncodeWithPrefix(extraconfig.MapSink(cfg), conf, "guestinfo.vch")
+	extraconfig.Encode(extraconfig.MapSink(cfg), conf)
 	spec.ExtraConfig = append(spec.ExtraConfig, extraconfig.OptionValueFromMap(cfg)...)
```
emlin: I ever removed the extraconfig settings during VM creation, is because this step will add duplicate value into extraconfig.

cmd/vic-machine/validate/validator.go
```diff
@@ -45,314 +51,577 @@ type Validator struct {
...
+	sessionconfig.Service = tURL.String()
+
+	v.Session = session.NewSession(sessionconfig)
+	v.Session, err = v.Session.Connect(v.Context)
+	if err != nil {
 		return nil, err
 	}
-	if _, err := v.getConnection(input); err != nil {
+
+	v.Session, err = v.Session.Populate(ctx)
```
emlin: This method will load default resources for cluster, datastore, etc. So you have to set value into session config, to avoid ambiguous.
I added one method getResources, to avoid session class, and that method  loads set resources only, so is OK to be invoked by all vic-machine commands. Is it good to invoke getResources here?

lib/install/management/appliance.go
hickeng: It's not duplicated, but it is set twice - the only reason it's in the initial spec for the applianceVM creation is so that the VM can be identified as a VCH appliance even if it's never powered on.
I'll add a comment.

cmd/vic-machine/validate/validator.go
```diff
@@ -45,314 +51,577 @@ type Validator struct {
...
+	pools, err = v.Session.Finder.ResourcePoolList(ctx, cpath)
 	if err != nil {
-		errs = append(errs, fmt.Sprintf("Failure finding pool (%s): %s", v.ResourcePoolPath, err.Error()))
-		if v.Session.Datacenter == nil {
-			log.Info("Invalid datacenter. Unable to suggest valid --compute-resource")
-		} else {
-			v.suggestComputeResource()
+		log.Debugf("failed to look up compute resource as cluster path %s: %s", cpath, err)
+		if _, ok := err.(*find.NotFoundError); !ok {
+			// we return err directly here so we can check the type
```
emlin: DO we need to suggest path here?

cmd/vic-machine/validate/validator.go
```diff
@@ -45,314 +51,577 @@ type Validator struct {
...
+			return nil, err
```
emlin: do we need to suggest pool here?

cmd/vic-machine/validate/validator.go
hickeng: @emlin `getResources` looks identical to `session.Populate` and I still see session in use throughout the code in validator.go so it's not intended to remove our dependency on the session package - are you in the process of removing session? If not what's the logic behind the duplication? It's not clear to me from your comment.

cmd/vic-machine/validate/validator.go
```diff
@@ -45,314 +51,577 @@ type Validator struct {
...
+	}
+
+	return numPools
+}
+
+func (v *Validator) listResourcePools(path string) int {
+	pools, _ := v.Session.Finder.ResourcePoolList(v.Context, path+"/*")
+	if len(pools) > 0 {
+		for _, p := range pools {
+			log.Infof("    %s", p.InventoryPath)
```
emlin: Assume we'll accept relative resource pool path, instead inventory path, do we need to change the suggested resource pool format here?

lib/install/management/appliance.go
```diff
@@ -69,7 +77,7 @@ func (d *Dispatcher) removeApplianceIfForced(conf *metadata.VirtualContainerHost
 			return vm.PowerOff(ctx)
 		})
 		if err != nil {
-			log.Warnf("Failed to power off existing appliance for %s, try to remove anyway", err)
+			log.Warnf("Failed to power off existing appliance for %s: err", conf.Name, err)
```
emlin: plan to remove this error message, as it doesn't really impact the result.

lib/portlayer/network/config.go
```diff
@@ -0,0 +1,50 @@
...
+var Config Configuration
+
+type Configuration struct {
+	// The default bridge network supplied for the Virtual Container Host
+	BridgeNetwork string `vic:"0.1" scope:"read-only" key:"bridge_network"`
+	// Published networks available for containers to join, keyed by consumption name
+	ContainerNetworks map[string]*ContainerNetwork `vic:"0.1" scope:"read-only" key:"container_networks"`
+}
+
+type ContainerNetwork struct {
```
hmahmood: Why not just reuse the struct in the metadata package? You can add PortGroup to that.

lib/metadata/virtual_container_host.go
```diff
@@ -56,73 +56,53 @@ type VirtualContainerHostConfigSpec struct {
...
 	// Certificates for specific system access, keyed by FQDN
-	HostCertificates map[string]tls.Certificate
+	HostCertificates map[string]*RawCertificate
 
 	// Port Layer - storage
-	// Datastore used for the ImageStore
-	// TODO this will change, probably to url.URL, as needed by the port layer
-	ImageStore string `vic:"0.1" scope:"read-only" key:"image_store"`
+	// Datastore URLs for image stores - the top layer is [0], the bottom layer is [len-1]
+	ImageStores []url.URL `vic:"0.1" scope:"read-only" key:"image_stores"`
```
emlin: When will we have multiple image stores layers configured?

emlin:
Well, there is lot of code refactor in vic-machine. It looks good, with a few comments. 👍

lib/install/management/appliance.go
emlin: make sense

cmd/vic-machine/validate/validator.go
emlin: so far, the dependency comes from object reference. And we still rely on the session to establish connection.
But populate method is not that helpful for me, cause they will load many objects, even we don't set value, e.g. for cluster, datastore, e.g.
If you didn't set cluster, and there is multiple clusters in VC, populate will throw exception. For the getResources method, is just meant to be used for our own resource validation. No matter for create, and also for other commands. For example for delete command, will only have resource pool set.
As you already have different methods for different kind of resources, why do you still need populate method?
Maybe we should remove both populate and getResources.

cmd/vic-machine/validate/validator.go
hickeng: Really good point - yes we should :)

lib/install/management/appliance.go
hickeng: moving it to debug instead of removing it.

emlin:
btw, do you have ETA for this change? Cause there are quite some changes for image store, container store. I think it would be better to rely on this for images and containers datastore clean up.
But that will have some logistic issue if you cannot have this RP ready for merge soon.

lib/portlayer/network/config.go
hickeng: If the core config struct becomes everything to everyone for runtime state as well as config then it will balloon significantly.

I would like to make the core structure more composition based so that we can easily specialize portions of it for runtime state without needing to care about having key names match up but not yet got around to it.

cmd/vic-machine/validate/validator.go
hickeng: Whoops - was supposed to be ignoring errors from Populate - I'd added that error clause to try and track down an issue. Very much looking forward to having the vcsim tests for this - didn't want to duplicate the work you've got in the delete PR though.

I'm using session simply as a convenient way to get defaults if available (hence deliberately _not_ populating it's config), but you're correct, the individual validation methods are similar in purpose and should replace that eventually. Longer term I'd like to move the recommendations to a package as it's definitely not tightly coupled to VIC, and replace session with the `vmomi gateway` component which is purely about connection handling, management, and authentication than about the inventory.

The individual validation methods don't replace session currently because session is used throughout and I'd have had to pull it out of appliance.go as well - the scope of this PR was already too large. The changes are the best balance I could find between keeping it small and providing the necessary extraconfig

cmd/vic-machine/validate/validator.go
hickeng: no, that's the error return if the finder returned an error that _wasn't_ NotFoundError - that's a different problem and indicates that there's an issue with our vmomi connection so subsequent operations are likely to fail.

lib/portlayer/network/config.go
hmahmood: So maybe something like

```go
type ContainerNetwork struct {
  metadata.ContainerNetwork
  PortGroup object.NetworkReference
}
```

I am not sure duplicating some of the elements of the struct here is a good idea.

lib/metadata/virtual_container_host.go
hickeng: post-GA most likely - I moved this to an array because where we've just had string values there's assumptions being made that the value is a) simple, b) always going to be a single value.
Using an array is a clear indicator that we expect to have more than one value for this in the future.

cmd/vic-machine/validate/validator.go
emlin: Sure

cmd/vic-machine/validate/validator.go
emlin: ok, misread that.

lib/install/management/appliance.go
emlin: ok

hickeng:
@emlin it's almost ready to go - there are additional pieces before I consider 746 complete, however I don't think we can delay this PR until then as it touches too many areas and is necessary for too many other TP3 items.

lib/portlayer/network/config.go
hickeng: Agreed, composition is preferred. This requires and addition to the extraConfig code of an `inline` annotation to suppress the extra key layer that would appear from reflection. I don't think it's in scope for this but have opened #1145 to track it.

hickeng:
Minor changes to the trace package to allow toggling of the trace through means other than adjusting the log level.
Ideally this would be inverted so we can tell the trace package what level we want it to log at, but there wasn't an obvious way to do so in logrus - no `func Log(Level, interface{}...)`

caglar10ur:
Let's not have PR's like this big in the future, it took me couple of hours to go over it and to be honest I constantly lost where I was and what I was thinking :)

LGTM

hickeng:
@caglar10ur agreed! It's too big - as with Mr. Twain I did not have time to make it shorter ;) Seriously, I tried splitting it up but didn't find a way to do so sensibly. Could and should have broken out the isolated supporting work.

emlin:
LGTM

caglar10ur:
@hickeng I think it's time to squash those commits, rebase and merge :)

hickeng:
LGTM

forcing check update for LGTMs from @caglar10ur and @emlin

PR #1132, opened 2016-06-17
Improve build speed, chain the builds so publish only happens on inte…

mhagen-vmware:
Original build time locally - just build+unit tests:
real	26m46.253s
user	0m0.092s
sys	    0m0.076s

Proposed build time locally - just build+unit tests:
real	18m49.260s
user	0m0.092s
sys	    0m0.080s

Reasons for including a custom image that builds on top of golang:1.6.2:

1. Improve build speed by not having to download the same dependencies every time
    - Without custom image - just the setup phase of integration tests:  
        real	18m4.377s  
        user	0m0.084s  
        sys	0m0.056s  
    - With custom image - just the setup phase of the intergation tests:  
        real	12m35.116s  
        user	0m0.060s  
        sys	0m0.040s  
2. Maintain a static environment for testing, so that if dependencies get updated it doesn't break our tests
3. Narrow down the number of external dependencies from 20+ to only 1 (bintray)
    - In the last two weeks, both docker hub and google dependencies broke our builds when they had problems/introduced a bug into their code
    - The above took us down for a full day in both cases
4. Simpler and more readable .drone file/more maintainable


infra/integration-image/Dockerfile.test
```diff
@@ -0,0 +1,37 @@
...
+RUN go get github.com/godbus/dbus
+RUN go get github.com/kr/pty
+RUN go get github.com/vishvananda/netlink
+RUN go get google.golang.org/grpc
+RUN go get github.com/docker/docker; echo 0
+
+RUN curl -sSL https://github.com/vmware/govmomi/releases/download/v0.7.0/govc_linux_amd64.gz | gzip -d > /usr/local/bin/govc
+RUN chmod +x /usr/local/bin/govc
+
+# Fix because swarm daemon started causing problems for some reason
```
hickeng: link to the associated issue if we're having to drop tests in this manner.

hickeng:
@mhagen-vmware how much of that time delta comes from the removal of the repeated apt-gets and how much is from the custom image?

infra/integration-image/package.list
```diff
@@ -0,0 +1,274 @@
+Desired=Unknown/Install/Remove/Purge/Hold
```
hickeng: The package list of what's in the build image shouldn't be a manually generated file checked into the repo - while this could be a useful reference for intent it doesn't capture what was actually used - it should be a build artifact generated each run.

infra/integration-image/check_test.go.bak
```diff
@@ -0,0 +1,216 @@
+package main
```
hickeng: If we duplicating/modifying docker's code like this it needs to be very explicit - note where it comes from and why it's modified as it is.

infra/integration-image/Dockerfile.test
```diff
@@ -0,0 +1,37 @@
...
+RUN apt-get update
+RUN apt-get install -y jq time
+RUN curl -sSL https://get.docker.com/ | sh
+
+RUN apt-get install -y gcc python-dev python-setuptools python-pip libffi-dev libssl-dev
+RUN pip install pyasn1 gsutil --upgrade
+RUN pip install robotframework
+RUN pip install robotframework-sshlibrary
+
+# Libraries required to run the docker integration tests
```
hickeng: Putting the core Go dependencies in the image definition like this means that this isn't just a base image for running tests in, it's an image that captures the tests to be run at a point in time.
That actually has value of it's own (we can assert that we pass/fail tests as of a specific date) - however:
1. these aren't pulled from a tagged release so it's luck as to what we get when the image is built
2. we *want* early exposure to changes in these packages - we'll get some of it by our updates to vendor, but this is another vector
3. these packages are already in the repo vendor directory.

infra/integration-image/Dockerfile.test
```diff
@@ -0,0 +1,37 @@
...
+RUN go get google.golang.org/grpc
+RUN go get github.com/docker/docker; echo 0
+
+RUN curl -sSL https://github.com/vmware/govmomi/releases/download/v0.7.0/govc_linux_amd64.gz | gzip -d > /usr/local/bin/govc
+RUN chmod +x /usr/local/bin/govc
+
+# Fix because swarm daemon started causing problems for some reason
+RUN rm /go/src/github.com/docker/docker/integration-cli/docker_api_swarm_test.go
+RUN rm /go/src/github.com/docker/docker/integration-cli/daemon_swarm.go
+COPY check_test.go.bak /go/src/github.com/docker/docker/integration-cli/check_test.go
```
hickeng: Assuming we need to have our own check_test.go for some reason (separate doc comment), copying it in as part of the image means we cannot update it and run tests without also rebuilding _and pushing_ the image. That means this image is not a general base for integration testing, it *is* the integration test for a given point in time. Was that the intent?

Is it viable to use something like the following in the drone.yml (not the best way to do this, just a direct translation from Dockerfile to drone.yml):
```
     commands:
      - cp infra/integration-image/check_test.go.bak /go/src/github.com/docker/docker/integration-cli/check_test.go
      - tests/integration-test.sh
```

Makefile
```diff
@@ -182,10 +182,6 @@ vendor: $(GVT)
 	@echo restoring vendor
 	$(GVT) restore
 
-integration-tests: $(GOVC) components isos vic-machine
```
emlin: Do we need a new task for current integration-tests?

infra/integration-image/Dockerfile.test
```diff
@@ -0,0 +1,37 @@
+# docker build -t vic-test -f Dockerfile.test .
+FROM golang:1.6.2
+
+RUN apt-get update
+RUN apt-get install -y jq time
+RUN curl -sSL https://get.docker.com/ | sh
```
caglar10ur: This looks wrong as it returns the latest version, we should request a specific version (or use distro package)

infra/integration-image/Dockerfile.test
```diff
@@ -0,0 +1,37 @@
...
+RUN go get github.com/docker/libtrust
+RUN go get github.com/gorilla/mux
+RUN go get github.com/docker/libnetwork/driverapi
+RUN go get github.com/godbus/dbus
+RUN go get github.com/kr/pty
+RUN go get github.com/vishvananda/netlink
+RUN go get google.golang.org/grpc
+RUN go get github.com/docker/docker; echo 0
+
+RUN curl -sSL https://github.com/vmware/govmomi/releases/download/v0.7.0/govc_linux_amd64.gz | gzip -d > /usr/local/bin/govc
```
caglar10ur: s/0.7.0/0.7.1/g

mhagen-vmware:
@caglar10ur @hickeng 
2 Issues:
1. The swarm hack that I had to put in I was able to remove, turns out it was an actual build problem on dockers part, waiting a couple of days and it fixed itself.
2. I was able to eliminate the go get package lines as well, I thought that GOPATH would be recursive but that was not working so explicitly specifying a multi-part GOPATH with the vendor folder fixed the issue.

caglar10ur:
LGTM :sheep: :it:

mdubya66:
Lgtm

.drone.yml
```diff
@@ -112,15 +74,6 @@ build:
     when:
       success: true
 
-notify:
```
hickeng: This could be really useful if it's on vic-dev and:
```
when:
   repo: vmware/vic
   branch: master
   success: false
```

Makefile
mhagen-vmware: @emlin I don't think this is feasible at this point.  The integration tests require much more orchestrating now, with the custom image, environment variables and drone secrets.  I think at this point the only mechanism we should support is the drone methodology.

hickeng:
LGTM

mhagen-vmware:
.

PR #1134, opened 2016-06-17
RFC: Registering a vSphere Extension for eventual use w/ RBAC

gigawhitlocks:
_This is currently WIP but it's time to start taking feedback_

This PR adds a routine to vic-machine to register a vSphere extension for us to do certificate-based authentication against. This is in lieu of creating a 'service account' or having a VI admin do this. Presumably more advanced RBAC may involve SSO-based authentication & authorization but this change allows the VCH to have the permissions allowed to it by the vSphere extension API without storing the VI admin's credentials in plain text.

It modifies the shared vSphere Session object to prefer certificates to username/password, and uses the extraconfig library to store the generated cert/key added to the extension.

TODO:
* More extensive testing (so far I've only tested against vSphere and behavior against ESX is undefined -- this will _obviously_ be addressed before merge) & test cases need to be written

* Strip the VI admin credentials from anywhere they might appear _outside of this change_ and outside of vic-machine (which needs them to bootstrap the install process).

* Anything y'all suggest

* If @emlin merges `delete` before this gets in I'll have to add a routine to that to unregister the extension also.

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -27,8 +27,11 @@ import (
 	"github.com/vmware/vic/lib/install/data"
 	"github.com/vmware/vic/lib/install/management"
 	"github.com/vmware/vic/lib/install/validate"
+	"github.com/vmware/vic/pkg/certificate"
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/trace"
+
+	"bytes"
 
 	"golang.org/x/net/context"
 )
@@ -223,16 +226,16 @@ func (c *Create) processParams() error {
 	return nil
 }
 
-func (c *Create) loadCertificate() (*Keypair, error) {
-	var keypair *Keypair
+func (c *Create) loadCertificate() (*certificate.Keypair, error) {
+	var keypair *certificate.Keypair
 	if c.cert != "" && c.key != "" {
 		log.Infof("Loading certificate/key pair - private key in %s", c.key)
-		keypair = NewKeyPair(false, c.key, c.cert)
+		keypair = certificate.NewKeyPair(false, c.key, c.cert)
 	} else if c.tlsGenerate && c.DisplayName != "" {
 		c.key = fmt.Sprintf("./%s-key.pem", c.DisplayName)
 		c.cert = fmt.Sprintf("./%s-cert.pem", c.DisplayName)
 		log.Infof("Generating certificate/key pair - private key in %s", c.key)
-		keypair = NewKeyPair(true, c.key, c.cert)
+		keypair = certificate.NewKeyPair(true, c.key, c.cert)
 	}
 	if keypair == nil {
 		log.Warnf("Configuring without TLS - to enable use -generate-cert or -key/-cert parameters")
@@ -316,7 +319,7 @@ func (c *Create) Run(cli *cli.Context) error {
 
 	log.Infof("### Installing VCH ####")
 
-	var keypair *Keypair
+	var keypair *certificate.Keypair
 	if keypair, err = c.loadCertificate(); err != nil {
 		log.Error("Creation cannot continue: unable to load certificate")
 		return err
@@ -342,8 +345,22 @@ func (c *Create) Run(cli *cli.Context) error {
 		return err
 	}
 
+	if keypair != nil {
+		vchConfig.UserKeyPEM = string(keypair.KeyPEM)
+		vchConfig.UserCertPEM = string(keypair.CertPEM)
+	}
+
 	vConfig := validator.AddDeprecatedFields(ctx, vchConfig, c.Data)
 	vConfig.ImageFiles = images
+
+	{ // create certificates for VCH extension
+		var certbuffer, keybuffer bytes.Buffer
+		if certbuffer, keybuffer, err = certificate.CreateRawKeyPair(); err != nil {
+			return errors.Errorf("Failed to create certificate for VIC vSphere extension: %s", err)
+		}
+		vchConfig.ExtensionCert = certbuffer.String()
+		vchConfig.ExtensionKey = keybuffer.String()
+	}
 
 	executor := management.NewDispatcher(ctx, validator.Session, vchConfig, c.Force)
 	if err = executor.Dispatch(vchConfig, vConfig); err != nil {
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/lib/install/data"
 	"github.com/vmware/vic/lib/metadata"
@@ -301,6 +302,26 @@ func (d *Dispatcher) findAppliance(conf *metadata.VirtualContainerHostConfigSpec
 	return newVM, nil
 }
 
+func (d *Dispatcher) generateExtensionName(conf *metadata.VirtualContainerHostConfigSpec) error {
+	// must be called after appliance VM is created
+	vm, err := d.findAppliance(conf)
+
+	if err != nil {
+		return errors.Errorf("Could not find appliance at extension creation time; failed with error: %s", err)
+	}
+
+	var o mo.VirtualMachine
+	err = vm.Properties(d.ctx, vm.Reference(), []string{"config.uuid"}, &o)
+	if err != nil {
+		return errors.Errorf("Could get VM UUID from appliance VM")
+	}
+
+	// identifier for extension, which vSphere calls a 'key' but we want to avoid confusion w/ privkey
+	extensionIDKey := "com.vmware.vic." + o.Config.Uuid
+	conf.ExtensionName = extensionIDKey
+	return nil
+}
+
 func (d *Dispatcher) configIso(conf *metadata.VirtualContainerHostConfigSpec, vm *vm.VirtualMachine) (object.VirtualDeviceList, error) {
 	var devices object.VirtualDeviceList
 	var err error
@@ -368,6 +389,18 @@ func (d *Dispatcher) createAppliance(conf *metadata.VirtualContainerHostConfigSp
 	}
 	log.Debugf("vm folder name: %s", d.vmPathName)
 	log.Debugf("vm inventory path: %s", vm2.InventoryPath)
+
+	// create an extension to register the appliance as
+	d.generateExtensionName(conf)
+	settings.Extension = types.Extension{
+		Description: &types.Description{
+			Label:   "VIC",
+			Summary: "vSphere Integrated Containers Virtual Container Host",
+		},
+		Company: "VMware, Inc.",
+		Version: "0.0",
+		Key:     conf.ExtensionName,
+	}
 
 	conf.AddComponent("vicadmin", &metadata.SessionConfig{
 		Cmd: metadata.Cmd{
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -302,6 +302,7 @@ func (d *Dispatcher) findAppliance(conf *metadata.VirtualContainerHostConfigSpec
 	return newVM, nil
 }
 
+// retrieves the uuid of the appliance vm to create a unique vsphere extension name
 func (d *Dispatcher) generateExtensionName(conf *metadata.VirtualContainerHostConfigSpec) error {
 	// must be called after appliance VM is created
 	vm, err := d.findAppliance(conf)
@@ -313,12 +314,10 @@ func (d *Dispatcher) generateExtensionName(conf *metadata.VirtualContainerHostCo
 	var o mo.VirtualMachine
 	err = vm.Properties(d.ctx, vm.Reference(), []string{"config.uuid"}, &o)
 	if err != nil {
-		return errors.Errorf("Could get VM UUID from appliance VM")
+		return errors.Errorf("Could not get VM UUID from appliance VM due to error: %s", err)
 	}
 
-	// identifier for extension, which vSphere calls a 'key' but we want to avoid confusion w/ privkey
-	extensionIDKey := "com.vmware.vic." + o.Config.Uuid
-	conf.ExtensionName = extensionIDKey
+	conf.ExtensionName = "com.vmware.vic." + o.Config.Uuid
 	return nil
 }
 
--- a/lib/install/management/dispatcher.go
+++ b/lib/install/management/dispatcher.go
@@ -185,10 +185,11 @@ func (d *Dispatcher) Dispatch(conf *metadata.VirtualContainerHostConfigSpec, set
 		return errors.Errorf("Uploading images failed with %s. Exiting...", err)
 	}
 
-	if err = d.registerExtension(conf, settings.Extension); err != nil {
-		return errors.Errorf("Error registering VCH vSphere extension: %s", err)
+	if d.session.IsVC() {
+		if err = d.registerExtension(conf, settings.Extension); err != nil {
+			return errors.Errorf("Error registering VCH vSphere extension: %s", err)
+		}
 	}
-
 	_, err = tasks.WaitForResult(d.ctx, func(ctx context.Context) (tasks.ResultWaiter, error) {
 		return d.appliance.PowerOn(ctx)
 	})
--- a/lib/install/management/dispatcher.go
+++ b/lib/install/management/dispatcher.go
@@ -133,7 +133,7 @@ func (d *Dispatcher) initDiagnosticLogs(conf *metadata.VirtualContainerHostConfi
 	}
 }
 
-func (d *Dispatcher) registerExtension(conf *metadata.VirtualContainerHostConfigSpec, settings *data.InstallerData) error {
+func (d *Dispatcher) registerExtension(conf *metadata.VirtualContainerHostConfigSpec, extension types.Extension) error {
 	log.Infoln("Registering VCH as a vSphere extension")
 
 	// vSphere confusingly calls the 'name' of the extension a 'key'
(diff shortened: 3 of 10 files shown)


lib/install/management/dispatcher.go
```diff
@@ -125,12 +139,108 @@ func (d *Dispatcher) initDiagnosticLogs(conf *metadata.VirtualContainerHostConfi
...
+			Summary: "vSphere Integrated Containers Virtual Container Host",
+		},
+		Company: "VMware, Inc.",
+		Version: "0.2",
+		Key:     "com.vmware.vic",
+	}
+
+	// anonymous function generates a keypair
+	var err error
+	if conf.ExtensionCert, conf.ExtensionKey, err = func() (string, string, error) {
```
caglar10ur: didn't we have this in cmd/vic-machine/create/certificate.go?

dougm:
Looking good @gigawhitlocks 

One issue that jumps out, there should be a VC extension per VCH instance.  The code currently supports only 1 extension, with the name "com.vmware.vic".  Just changing that to something like "com.vmware.vic.${vch_name}" to should do the trick.

lib/install/management/dispatcher.go
dougm: Don't see a need to use an anonymous func here.. can this func be moved elsewhere to its own named function or method?

lib/install/management/appliance.go
```diff
@@ -402,6 +402,19 @@ func (d *Dispatcher) reconfigureApplianceSpec(vm *vm.VirtualMachine, conf *metad
 			Key:   "guestinfo.vch/networks",
 			Value: " ",
 		})
+
+	spec.ExtraConfig = append(spec.ExtraConfig,
```
hmahmood: You should not need these if you're using the extraconfig package.

lib/install/management/dispatcher.go
gigawhitlocks: Yeah, it's pretty much a copypasta -- I didn't just call that code for two reasons but I'm obviously open to input on this:
1) We don't write the keypair out to a file here, so the routine in certificate would have to be refactored somewhat as it doesn't fit the use-case exactly and
2) I'd need to break it into its own package in order to use it from here and from `create` or there's a cyclical import. I didn't want to waste time belaboring it so I copied over the part of the routine that I needed (stripping out the write-to-file parts), wrapped it in an anonymous function, and called it a day.

gigawhitlocks:
Oh @dougm I actually thought about that and was under the impression that right now we only support one VCH per vSphere installation. I can easily make this change, however.

lib/install/management/dispatcher.go
caglar10ur: You can  easily split the original function into 2 where the first one returns keys and second one writes them to a file. vic-machine can call both, you can call only first one.

gigawhitlocks: Yeah I'll go ahead and do that.

gigawhitlocks:
OK @caglar10ur that should be more to your liking. Do note where I put the new `certificate` package as I'm not sure that's the right/best place to put a new utility package.

lib/install/management/dispatcher.go
```diff
@@ -125,12 +131,60 @@ func (d *Dispatcher) initDiagnosticLogs(conf *metadata.VirtualContainerHostConfi
...
+	extensionManager := object.NewExtensionManager(d.session.Vim25())
+	extension := types.Extension{
+		Description: &types.Description{
+			Label:   "VIC",
+			Summary: "vSphere Integrated Containers Virtual Container Host",
+		},
+		Company: "VMware, Inc.",
+		Version: "0.2",
+		Key:     "com.vmware.vic",
+	}
```
caglar10ur: Can you move them to const section so that we can see as soon as we open the file. Also where is that 0.2 coming from :) ?

cmd/vic-machine/create/create.go
```diff
@@ -27,6 +27,7 @@ import (
 	"github.com/vmware/vic/cmd/vic-machine/data"
 	"github.com/vmware/vic/cmd/vic-machine/validate"
 	"github.com/vmware/vic/lib/install/management"
+	"github.com/vmware/vic/pkg/certificate"
```
caglar10ur: maybe vsphere/certificate, not sure

lib/install/management/dispatcher.go
```diff
@@ -125,12 +131,60 @@ func (d *Dispatcher) initDiagnosticLogs(conf *metadata.VirtualContainerHostConfi
...
+		}
+	}
+
+	if err := extensionManager.Register(d.ctx, extension); err != nil {
+		log.Errorf("Could not register the vSphere extension")
+		return err
+	}
+
+	// TODO: set com.vmware.vic as a const
+	if err := extensionManager.SetCertificate(d.ctx, "com.vmware.vic", conf.ExtensionCert); err != nil {
```
caglar10ur: Wonderfull suggestion in the TODO comment 👍 :)

pkg/certificate/certificate.go
```diff
@@ -46,11 +46,12 @@ func NewKeyPair(tlsGenerate bool, keyFile, certFile string) *Keypair {
...
 
-func (k *Keypair) generate() error {
-	org := ""
+func CreateRawKeyPair() (bytes.Buffer, bytes.Buffer, error) {
+	var cert, key bytes.Buffer
+	org := "VMware"
 	priv, err := rsa.GenerateKey(rand.Reader, 2048)
 	if err != nil {
-		return err
+		return cert, key, err
```
caglar10ur: Use ```return nil, nil, err ```as they are much easier to spot and suggests an error case

pkg/certificate/certificate.go
```diff
@@ -60,7 +61,7 @@ func (k *Keypair) generate() error {
 	serialNumber, err := rand.Int(rand.Reader, serialNumberLimit)
 	if err != nil {
 		err = errors.Errorf("Failed to generate random number: %s", err)
-		return err
+		return cert, key, err
```
caglar10ur: ^ ditto

pkg/certificate/certificate.go
```diff
@@ -77,21 +78,30 @@ func (k *Keypair) generate() error {
 	derBytes, err := x509.CreateCertificate(rand.Reader, &template, &template, &priv.PublicKey, priv)
 	if err != nil {
 		err = errors.Errorf("Failed to generate x509 certificate: %s", err)
-		return err
+		return cert, key, err
 	}
 
-	var cert bytes.Buffer
```
caglar10ur: ^ ditto

pkg/certificate/certificate.go
```diff
@@ -77,21 +78,30 @@ func (k *Keypair) generate() error {
...
 	err = pem.Encode(&cert, &pem.Block{Type: "CERTIFICATE", Bytes: derBytes})
 	if err != nil {
-		return err
+		err = errors.Errorf("Failed to encode x509 certificate: %s", err)
+		return cert, key, err
```
caglar10ur: ^ ditto

pkg/vsphere/session/session.go
```diff
@@ -177,7 +188,7 @@ func (s *Session) Connect(ctx context.Context) (*Session, error) {
 	if !s.HasCertificate() {
 		err = s.Client.Login(ctx, user)
 	} else {
-		err = s.LoginExtensionByCertificate(ctx, user.Username(), "")
+		err = s.LoginExtensionByCertificate(ctx, "com.vmware.vic", "")
```
caglar10ur: make constants great again

pkg/certificate/certificate.go
gigawhitlocks: Go actually doesn't allow this. I had it initially and it gave me a type error. Am I missing something @caglar10ur ?

lib/install/management/dispatcher.go
gigawhitlocks: ahaha yeah that was for myself, meant to do that before checking it in, whoops!

lib/install/management/dispatcher.go
gigawhitlocks: oh ha the 0.2 is nonsense isn't it, my bad, that's from where I was looking at some of Doug's Bonneville code and for some reason didn't change it. Do we have a version constant living somewhere?

cmd/vic-machine/create/create.go
gigawhitlocks: yeah it's not vSphere specific code though, it's a general certificate utility package

lib/install/management/dispatcher.go
```diff
@@ -125,12 +131,60 @@ func (d *Dispatcher) initDiagnosticLogs(conf *metadata.VirtualContainerHostConfi
...
+		log.Errorf("Could not set the certificate on the vSphere extension")
```
fdawg4l: Might be useful to print the error.

pkg/certificate/certificate.go
```diff
@@ -46,11 +46,12 @@ func NewKeyPair(tlsGenerate bool, keyFile, certFile string) *Keypair {
 	}
 }
 
-func (k *Keypair) generate() error {
-	org := ""
+func CreateRawKeyPair() (bytes.Buffer, bytes.Buffer, error) {
```
fdawg4l: Please add a comment as to what this function does and what the returned values are specifically.

pkg/vsphere/session/session.go
```diff
@@ -156,9 +167,9 @@ func (s *Session) Connect(ctx context.Context) (*Session, error) {
 		}
 
 		// load the certificates
-		cert, err2 := tls.LoadX509KeyPair(s.CertFile, s.KeyFile)
+		cert, err2 := tls.X509KeyPair([]byte(s.ExtensionCert), []byte(s.ExtensionKey))
```
fdawg4l: I know `err2` isn't your addition, but do we need it?  Can we use the same err?

fdawg4l:
Change looks fine.  Minor stuff.  Up to you to deal with anything i mentioned since it's all minor.

fdawg4l:
LGTM

fdawg4l:
But please do wait for @dougm's approval before checking in since he's the vsphere expert.

lib/install/management/dispatcher.go
dougm: I think 0.2 is actually from govc/test/extension.bats, in bonneville the extension was registered by the WebClient UI plugin.  It is just cosmetic, so "0.0.0" would be fine for now and later we get a version variable in place to use both here and vic-machine/main.go

dougm:
@fdawg4l @gigawhitlocks vsphere stuff looks good, except we do need to support multiple VCH instances per VC.  So we don't want to be using the same "com.vmware.vic" extension id.

gigawhitlocks:
Sounds good guys. I'll address these ASAP.

hickeng:
@gigawhitlocks if you need a value to append to the end of the extension ID I suggest using the moid of the VCH appliance VM or the UUID (I think that's just guest.UUID).

lib/install/management/dispatcher.go
```diff
@@ -125,12 +131,60 @@ func (d *Dispatcher) initDiagnosticLogs(conf *metadata.VirtualContainerHostConfi
...
+			Summary: "vSphere Integrated Containers Virtual Container Host",
+		},
+		Company: "VMware, Inc.",
+		Version: "0.2",
+		Key:     "com.vmware.vic",
+	}
+
+	var err error
+	var certbuffer, keybuffer bytes.Buffer
+	if certbuffer, keybuffer, err = certificate.CreateRawKeyPair(); err != nil {
```
emlin: I would prefer to see the certificate is generated in cmd/vic-machine/create pkg, instead of here. There already has certificate generation for docker service.
This generation should happen before other VC resource creation, e.g. resource pool, bridge network, etc.

lib/install/management/dispatcher.go
```diff
@@ -125,12 +131,60 @@ func (d *Dispatcher) initDiagnosticLogs(conf *metadata.VirtualContainerHostConfi
...
+	log.Infoln("Registering VCH as a vSphere extension")
+	extensionManager := object.NewExtensionManager(d.session.Vim25())
+	extension := types.Extension{
+		Description: &types.Description{
+			Label:   "VIC",
+			Summary: "vSphere Integrated Containers Virtual Container Host",
+		},
+		Company: "VMware, Inc.",
+		Version: "0.2",
+		Key:     "com.vmware.vic",
```
emlin: would suggest to append appliance VM id, to make the extension key unique.

gigawhitlocks:
The appliance VM doesn't exist at the time that the extension is registered (just realized/confirmed this), so the MOID doesn't exist yet either. ${vch_name} makes the extension's key human readable anyway.

lib/install/management/dispatcher.go
gigawhitlocks: yep working on this now

lib/install/management/dispatcher.go
gigawhitlocks: @emlin I'm not sure that I understand what you mean. I've moved the certificate routines from cmd/vic-machine/create to their own package so that I can call into them. Right now the generation does take place before the rest of the resource creation, except the resource pool, but I can move the routine to be called before that too. I believe I put it after the resource pool since the creation of the resource pool needs to have a unique name and I plan to reuse that name to make the extension's key unique.

pkg/certificate/certificate.go
gigawhitlocks: Yeah I agree, this totally needs comments. I know you called this minor further down the page but I actually disagree, especially since I didn't use named return values, it's pretty unclear.

lib/install/management/dispatcher.go
emlin: @gigawhitlocks what I mean is not for where to register the extension, which should happen after the appliance VM creation, then you can get the VM mob-id or UUID.
But for the certificate generation, cause theoretically it is still possible to have one certificate generation error, so it will be better to have the generation finished before we start VC connection.
That's why we have user (or docker) certificate generated in create package.
I assume what you're doing is a similar process.

lib/metadata/virtual_container_host.go
```diff
@@ -96,6 +100,26 @@ type VirtualContainerHostConfigSpec struct {
 
 	// Allow custom naming convention for containerVMs
 	ContainerNameConvention string
+
```
gigawhitlocks: @hickeng I did a rebase against master after your PR went in and I'm seeing some things showing up now as green in the diff which leads me to think you removed some things and I've unintentionally re-added them -- can you confirm that e.g. this code segment was stripped by your change?

gigawhitlocks:
All present comments should be addressed at this time.

For anyone following along, things still missing:
# Still working out where the ServiceURL contains credentials and making sure those don't get passed to the appliance, but are still usable by the installer
# Testing

gigawhitlocks:
Just needs a rebase and tests now and it's ready to go.

lib/install/management/dispatcher.go
```diff
@@ -147,6 +149,31 @@ func (d *Dispatcher) initDiagnosticLogs(conf *metadata.VirtualContainerHostConfi
 	}
 }
 
+func (d *Dispatcher) registerExtension(conf *metadata.VirtualContainerHostConfigSpec, settings *InstallerData) error {
```
gigawhitlocks: note to self: change this to take the extension pointer instead of installerdata pointer and pass the struct at method call time instead, so that this method could be reused to register another extension

lib/install/management/dispatcher.go
```diff
@@ -165,13 +192,19 @@ func (d *Dispatcher) Dispatch(conf *metadata.VirtualContainerHostConfigSpec, set
...
+
 	if err = d.createAppliance(conf, settings); err != nil {
 		return errors.Errorf("Creating the appliance failed with %s. Exiting...", err)
 	}
+
 	if err = d.uploadImages(settings.ImageFiles); err != nil {
 		return errors.Errorf("Uploading images failed with %s. Exiting...", err)
 	}
 
+	if err = d.registerExtension(conf, settings); err != nil {
```
gigawhitlocks: per my last comment to myself, this should be registerExtension(conf, settings.Extension)

lib/install/management/appliance.go
```diff
@@ -254,6 +255,26 @@ func (d *Dispatcher) findAppliance(conf *metadata.VirtualContainerHostConfigSpec
...
+	vm, err := d.findAppliance(conf)
+
+	if err != nil {
+		return errors.Errorf("Could not find appliance at extension creation time; failed with error: %s", err)
+	}
+
+	var o mo.VirtualMachine
+	err = vm.Properties(d.ctx, vm.Reference(), []string{"config.uuid"}, &o)
+	if err != nil {
+		return errors.Errorf("Could get VM UUID from appliance VM")
```
dougm: typo.  Should also include the 'err' in your Errorf.

lib/install/management/dispatcher.go
```diff
@@ -147,6 +149,31 @@ func (d *Dispatcher) initDiagnosticLogs(conf *metadata.VirtualContainerHostConfi
...
+	// vSphere confusingly calls the 'name' of the extension a 'key'
+	// This variable is named IdKey as to not confuse it with its private key
+	if conf.ExtensionCert == "" {
+		return errors.Errorf("Extension certificate does not exist")
+	}
+
+	extensionManager := object.NewExtensionManager(d.session.Vim25())
+
+	if err := extensionManager.Register(d.ctx, settings.Extension); err != nil {
+		log.Errorf("Could not register the vSphere extension")
```
dougm: same here: Should also include the 'err' in your Errorf.

lib/install/management/dispatcher.go
```diff
@@ -147,6 +149,31 @@ func (d *Dispatcher) initDiagnosticLogs(conf *metadata.VirtualContainerHostConfi
...
+		return err
+	}
+
+	// TODO: set com.vmware.vic as a const
+	if err := extensionManager.SetCertificate(d.ctx, conf.ExtensionName, conf.ExtensionCert); err != nil {
+		log.Errorf("Could not set the certificate on the vSphere extension")
```
dougm: ditto.

dougm:
lgtm

just a few log message nits

lib/install/management/dispatcher.go
```diff
@@ -147,6 +149,31 @@ func (d *Dispatcher) initDiagnosticLogs(conf *metadata.VirtualContainerHostConfi
...
+	}
+
+	extensionManager := object.NewExtensionManager(d.session.Vim25())
+
+	if err := extensionManager.Register(d.ctx, settings.Extension); err != nil {
+		log.Errorf("Could not register the vSphere extension")
+		return err
+	}
+
+	// TODO: set com.vmware.vic as a const
```
gigawhitlocks: clean up irrelevant TODO

lib/install/management/appliance.go
gigawhitlocks: oh good, "Could get". Good catch

lib/install/management/appliance.go
```diff
@@ -254,6 +255,26 @@ func (d *Dispatcher) findAppliance(conf *metadata.VirtualContainerHostConfigSpec
...
+	}
+
+	// identifier for extension, which vSphere calls a 'key' but we want to avoid confusion w/ privkey
+	extensionIDKey := "com.vmware.vic." + o.Config.Uuid
```
gigawhitlocks: rename this

gigawhitlocks:
OK this is ready. I added a unit test for the single self-contained piece. The rest should be covered by existing vic-machine integration tests, as they would fail if this stuff broke the authentication.

gigawhitlocks:
Well it looks like my two LGTMs aren't valid anymore so I'll merge this as soon as those checks are satisfied.

PR #1137, opened 2016-06-17
Create tools to help in auditing the vendor manifest

### LINKED ISSUE #1136: Create a way of auditing our vendor manifest
Auditing the vendor manifest is a painful process. There are two main goals here:
1) Making sure that for any given repository, we are using consistent dependency versions
2) Making sure that we are using tagged versions of a dependency, where possible

In some cases there will be reasons why a specific revision must be used and these should now be documented in the exceptions json file next to manifest.

I've created some tools that use the Github REST API to create a human-readable report on our tags and dependencies, grouped by repository. This tool will be used to do a root and branch review of what dependencies we have and whether they meet the requirements above.

The source for the tool should be compiled on demand by the user. There's no need for this to be part of our build.

corrieb:
Note that this is not part of the build. Rather it should be considered scripts that can be compiled and run if we need to audit our manifest.

Fixes #1136
[skip-ci]


caglar10ur:
I think vendor/ is not the correct place to put this tool. Maybe infra/ or contrib?

mdubya66:
i concur with caglar10ur, vendor should be just the vendored code. otherwise. lgtm.

hickeng:
concur with @caglar10ur - this should be in infra. Maybe rename scripts to utils so it's not misleading as to the type of the content and put them in there.

dougm:
lgtm

seems generally useful, maybe consider folding these features into the gvt tool itself at some point.

caglar10ur:
LGTM :sheep: :it:

hickeng:
LGTM

PR #1139, opened 2016-06-17
Update image cache in background

### LINKED ISSUE #1138: Docker persona should update image cache in background to avoid restarts
**Expected behavior:**
The persona should not exit if the update does not succeed. It should retry some number of times and then warn.

**Actual behavior:**
Currently the persona exits when the cache is not updated at startup. This happens a couple of times at startup until the portlayer is up and serving requests.

jzt:
This updates the image cache in the docker persona in a goroutine. Previously it would exit every time the cache failed to update, causing restarts by the watchdog. It now retries 5 times, ultimately logging a warning upon failure to update.

This change also includes some minor log cleanup.

Fixes #1138.

### DIFF
--- a/lib/apiservers/engine/backends/cache/image_cache.go
+++ b/lib/apiservers/engine/backends/cache/image_cache.go
@@ -16,6 +16,7 @@ package cache
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -31,6 +32,15 @@ import (
 	"github.com/vmware/vic/lib/metadata"
 )
 
+var (
+	// ErrCacheNotUpdated is returned when an operation is attempted while the cache has
+	// not yet hydrated.
+	ErrCacheNotUpdated = errors.New("Image cache not updated")
+
+	// CacheNotUpdated indicates whether or not the cache has been successfully hydrated
+	CacheNotUpdated = true
+)
+
 // ImageCache is an in-memory cache of image metadata. It is refreshed at startup
 // by a call to the portlayer. It is updated when new images are pulled or
 // images are deleted.
@@ -41,16 +51,19 @@ type ImageCache struct {
 	cache map[string]*metadata.ImageConfig
 }
 
+// NewImageCache creates and returns a new ImageCache
 func NewImageCache() *ImageCache {
 	return &ImageCache{
 		cache: make(map[string]*metadata.ImageConfig),
 	}
 }
 
-// Update adds new layers to the cache
+// Update adds new images to the cache
 func (c *ImageCache) Update(client *client.PortLayer) error {
 	c.m.Lock()
 	defer c.m.Unlock()
+
+	CacheNotUpdated = true
 
 	log.Debugf("Updating image cache...")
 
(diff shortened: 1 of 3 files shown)


sflxn:
LGTM

sflxn:
Hey, I gave you an ok, but if you want to add the hydrating status to the cache like we chatted about, I wouldn't mind.  It should help with the race condition where users attempt to get data before it's finished hydrating.

caglar10ur:
Asking to learn, why is exiting is a bad thing in that situation? Not being able to reach the portlayer sounds like a problem we shouldn't ignore :)

hickeng:
@caglar10ur It exiting requires either:
1. strict order of start between components and blocking until a component is *fully* initialized
2. a watchdog to do the restarting once the dependency is up.

Some tolerance significantly simplifies matters - it does mean we should figure out a formal way of reporting component status.

lib/apiservers/engine/backends/vicbackends.go
```diff
@@ -45,10 +49,23 @@ func Init(portLayerAddr string) error {
...
+			if err := imageCache.Update(portLayerClient); err == nil {
+				log.Info("Image cache updated successfully")
+				return
+			}
+			log.Info("Failed to refresh image cache, retrying...")
+		}
+		log.Warn("Failed to refresh image cache. Is the portlayer server down?")
+	}()
 
 	return nil
```
hickeng: If the image cache isn't updated, how functional is the system?

jzt: Not very.. the persona would have no knowledge of what lives on the portlayer. I've refactored the `Images` function to rely on the cache. I considered going around the cache directly to the portlayer in the event that the cache wasn't updated, but if the cache isn't updated, it's because the portlayer is either down or unreachable, which should demand some attention.

jzt:
@caglar10ur I am not opposed to exiting in the event that the portlayer is unreachable. My mindset while writing this was that it may just not be up yet. I don't know why it would ever need more than 10 seconds to come up, but if it did, the first/next call to pull an image would trigger another update. If it was still unreachable in that case, imagec would make it known that something was wrong.

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -41,17 +51,20 @@ type ImageCache struct {
...
 	}
 }
 
-// Update adds new layers to the cache
+// Update adds new images to the cache
 func (c *ImageCache) Update(client *client.PortLayer) error {
 	c.m.Lock()
 	defer c.m.Unlock()
 
+	CacheNotUpdated = true
```
sflxn: You should also initialize this in the var section above or put in init() just to avoid any potential for race condition where we try to get before update.  It'll only happen in the event that there is a change in how we rehydrate in the future.  Never know.

jzt: Good catch, fixed

caglar10ur:
LGTM :sheep: :it:

jzt:
♻️

lib/apiservers/engine/backends/vicbackends.go
sflxn: The portlayer has a ping request.  We can keep trying to check if the portlayer's rest server is up before bringing up the personality's rest server.

sflxn:
lgtm

PR #1140, opened 2016-06-17
Move images url namespace to /images

fdawg4l:
The storage layer API for images used urls to specifiy datastore
locations.  These URLs are different than vsphere datastore URLs.  We
use URLs to avoid embedding any vsphere semantics in the APIs including
using datastore URLs.

Images used the form `http://<host>/storage/<image store name>/<image
name>`.  To differentiate between images and volumes, Images are being
moved to `http://<host>/storage/images/<image store name>/<image name>`.


caglar10ur:
LGTM :sheep: :it:

hickeng:
LGTM

fdawg4l:
🎱

PR #1141, opened 2016-06-18
Implement docker inspect image

### LINKED ISSUE #388: Docker API: Image Inspect

sflxn:
This is docker image inspect not docker container inspect.  We've split up docker inspect into two operations (as they are in the docker daemon).  Container inspect is still blocked on metadata support.

Resolves #388.


lib/apiservers/engine/backends/image.go
```diff
@@ -179,3 +193,76 @@ func convertV1ImageToDockerImage(image *metadata.ImageConfig) *types.Image {
...
+		}
+	}
+
+	return nil, derr.NewRequestNotFoundError(fmt.Errorf("No such image: %s", image))
+}
+
+// Converts the data structure retrieved from the portlayer.  This src datastructure
+// represents the unmarshalled data saved in the storage port layer.  The return
+// data is what the Docker CLI understand and returns to user.
+func imageConfigToDockerImageInspect(imageConfig *metadata.ImageConfig, productName string) *types.ImageInspect {
```
fdawg4l: This should probably have some validation in it to make sure all of the fields of the `imageConfig` struct are 1) set and 2) valid, no?  I would check for the fields being set and return an error if they aren't there sane.

fdawg4l:
This change looks fine.  The data being inspected looks like it's primarily coming from a cache.  And that cache has a global `getter`.  So it should be possible to write a unit test which populates the cache, then creates an `Image` struct, calls the inspect method and validates the result.

cmd/docker/main.go
```diff
@@ -46,7 +46,7 @@ type CliOptions struct {
 	proto         string
 }
 
-const productName = "vSphere Integrated Containers"
+const productName = "vSphere Integrated Containers 0.3"
```
caglar10ur: Let's create a version file or a build tag for this and either read or inject into at compile time

caglar10ur:
LGTM :sheep: :it:

cmd/docker/main.go
hickeng: agree

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -115,3 +119,51 @@ func (c *ImageCache) GetImages() []*metadata.ImageConfig {
...
 }
+
+func (c *ImageCache) GetImageByDigest(digest digest.Digest) *metadata.ImageConfig {
+	c.m.RLock()
+	defer c.m.RUnlock()
+
+	var imageID string
+
+	for _, image := range c.cache {
+		if strings.Contains(image.ImageID, "sha") {
```
hickeng: Should this be `strings.HasPrefix`? Currently we will search the entire string if it's not a sha digest and I've no idea if those characters could occur in the actual digest sum.
A comment about the format this is attempting to assert would be good.

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -115,3 +119,51 @@ func (c *ImageCache) GetImages() []*metadata.ImageConfig {
...
+			imageID = image.ImageID
+		} else {
+			imageID = "sha256:" + image.ImageID
```
hickeng: Is this an assumption that it's a sha256? Is that enforced else where in the code - if so comment that this must be a sha256.
Are there other types it could be? If so should we iterate over those?

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -115,3 +119,51 @@ func (c *ImageCache) GetImages() []*metadata.ImageConfig {
...
+		}
+
+		if imageID == string(digest) {
+			newImage := new(metadata.ImageConfig)
```
hickeng: If this is attempting to isolate the cache from changes then it needs to do a deep copy. At the moment changing elements in the history array in the returned structure will alter the entry in the cache:
https://play.golang.org/p/By21rCOeg1

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -115,3 +119,51 @@ func (c *ImageCache) GetImages() []*metadata.ImageConfig {
...
 	return result
 }
+
+func (c *ImageCache) GetImageByDigest(digest digest.Digest) *metadata.ImageConfig {
+	c.m.RLock()
+	defer c.m.RUnlock()
+
+	var imageID string
+
+	for _, image := range c.cache {
```
hickeng: This loop is currently O(n) for the number of images we have. If lookup by digest is needed in anything other than very rare cases, consider having an additional index keyed by digest.

lib/apiservers/engine/backends/cache/image_cache.go
hickeng: Should this be `strings.HasPrefix(image.ImageID, "sha")`? At the moment it will both:
a. check the entire string rather than the start
b. match on the string incorrectly if "sha" can be part of the digest (I don't know if it can - is the digest a hex string?).

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -115,3 +119,51 @@ func (c *ImageCache) GetImages() []*metadata.ImageConfig {
...
+		if namedtag, ok := ref.(reference.NamedTagged); ok {
+			if named.RemoteName() == image.Name && namedtag.Tag() == image.Tag {
+				newImage := new(metadata.ImageConfig)
+				*newImage = *image
+				return newImage
+			}
+		} else {
+			if named.Name() == image.ImageID {
+				newImage := new(metadata.ImageConfig)
+				*newImage = *image
```
hickeng: ditto for prior deep copy comment

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -115,3 +119,51 @@ func (c *ImageCache) GetImages() []*metadata.ImageConfig {
...
+	for _, image := range c.cache {
+		ref := reference.WithDefaultTag(named)
+		if namedtag, ok := ref.(reference.NamedTagged); ok {
+			if named.RemoteName() == image.Name && namedtag.Tag() == image.Tag {
+				newImage := new(metadata.ImageConfig)
+				*newImage = *image
+				return newImage
+			}
+		} else {
+			if named.Name() == image.ImageID {
```
hickeng: Is this where we would implement lookup by short ID? If so is it viable to do so now:
1. use strings.HasPrefix instead of ==
2. continue iterating after the first match and abort with error if there's a second match

lib/apiservers/engine/backends/image.go
sflxn: We can add some validation.  The question is where?  My personal preference is to add the validation in the image cache on update.  That way, all operations that require image data can trust that source data.

cmd/docker/main.go
sflxn: I have a GI for changing the way we do strings in the personality server.  I'd rather do that there than in this PR.

lib/apiservers/engine/backends/cache/image_cache.go
sflxn: I'll add the comment in both places.

lib/apiservers/engine/backends/cache/image_cache.go
sflxn: Yeah, I realized that when I was putting it in, with the thought of returning to it to optimize it.  I'm going to go ahead and change this now.

lib/apiservers/engine/backends/cache/image_cache.go
sflxn: Sha is a hex string except the prefix, but I'm going to make the change.

lib/apiservers/engine/backends/cache/image_cache.go
sflxn: Changing that.  There's no reason to make a copy.

lib/apiservers/engine/backends/cache/image_cache.go
sflxn: I guess you found the GI for the partial id lookup.

sflxn:
@hickeng @fdawg4l @caglar10ur Made all the requested changes.

lib/apiservers/engine/backends/container.go
```diff
@@ -755,7 +755,7 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
 	return host, nil
 }
 
-func (c *Container) getImageMetadataFromStoragePL(image string) (*viccontainer.VicContainer, error) {
+func (c *Container) getImageMetadataFromImageC(image string) (*viccontainer.VicContainer, error) {
```
cgtexmex: Why is this still needed?  The persona has an image cache that should have the image information you seek...

lib/apiservers/engine/backends/container.go
```diff
@@ -151,7 +151,7 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 	var layer *viccontainer.VicContainer
 	container := viccontainer.GetCache().GetContainerByName(config.Config.Image)
 	if container == nil {
-		layer, err = c.getImageMetadataFromStoragePL(config.Config.Image)
+		layer, err = c.getImageMetadataFromImageC(config.Config.Image)
```
jzt: If you're using this just to get the layer ID, it's in `ImageConfig.V1Image.ID`.

lib/apiservers/engine/backends/container.go
sflxn: That was just a function name change.  Create still uses that.  This PR wasn't really meant to touch everything that uses image metadata.  I'll leave the changes for Create to @caglar10ur.  I think he's slated for that.

lib/apiservers/engine/backends/container.go
sflxn: see above.  I was in that file when I started inspect and noticed the function was incorrectly named.  It's not related to the purpose of the PR.

jzt:
LGTM

lib/apiservers/engine/backends/image.go
```diff
@@ -102,6 +114,8 @@ func (i *Image) ExportImage(names []string, outStream io.Writer) error {
 }
 
 func (i *Image) PullImage(ctx context.Context, ref reference.Named, metaHeaders map[string][]string, authConfig *types.AuthConfig, outStream io.Writer) error {
+	defer trace.End(trace.Begin("PullImage"))
```
hickeng: This will log the name of the fuction anyway. If supplying context, data from `ref` makes more sense.

lib/apiservers/engine/backends/image.go
```diff
@@ -81,8 +84,17 @@ func (i *Image) Images(filterArgs string, filter string, all bool) ([]*types.Ima
 	return result, nil
 }
 
+// Docker Inspect.  LookupImage looks up an image by name and returns it as an
+// ImageInspect structure.
 func (i *Image) LookupImage(name string) (*types.ImageInspect, error) {
-	return nil, fmt.Errorf("%s does not implement image.LookupImage", i.ProductName)
+	defer trace.End(trace.Begin("LookupImage (docker inspect)"))
```
hickeng: Could just pass `name` to trace being - allows us to differentiate concurrent requests easily.

lib/apiservers/engine/backends/vicbackends.go
```diff
@@ -28,6 +28,7 @@ const (
 	Imagec           = "imagec"
 	Retries          = 5
 	RetryTimeSeconds = 2
+	PortlayerName    = "Backend Engine"
```
hickeng: Given this is used to identify the graph driver that's associated with an image, "vSphere" is a better user facing term.

hickeng:
LGTM
Minor comments - they can go in now if you've got to rebase anyway, but don't mind if it's done in a followup PR.

Issue #1099, opened 2016-06-14
Docker API Server: Cleanup strings handling


sflxn:
We're currently adding in strings ad hoc, much like how Docker does it.  This is bad practice.  We should look into handling strings in a more structured way.

hickeng:
@sflxn can you give an example? We have a message catalogue mechanism that should be used for most user facing strings - this includes the error messages returned from the Docker API - they are a soft part of the API as people _will_ search for specific strings in command output.

mdubya66:
making low. is this still needed? @sflxn

PR #1142, opened 2016-06-18
Docker info without VCH info or container count

sflxn:
This is a WIP for Docker Info.  It is still waiting on container metadata in the portlayer.  It also needs portlayer support to get resource pool info.

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -50,7 +50,6 @@ import (
 
 // Container struct represents the Container
 type Container struct {
-	ProductName string
 }
 
 const (
@@ -61,32 +60,32 @@ const (
 
 // ContainerExecCreate sets up an exec in a running container.
 func (c *Container) ContainerExecCreate(config *types.ExecConfig) (string, error) {
-	return "", fmt.Errorf("%s does not implement container.ContainerExecCreate", c.ProductName)
+	return "", fmt.Errorf("%s does not implement container.ContainerExecCreate", ProductName())
 }
 
 // ContainerExecInspect returns low-level information about the exec
 // command. An error is returned if the exec cannot be found.
 func (c *Container) ContainerExecInspect(id string) (*backend.ExecInspect, error) {
-	return nil, fmt.Errorf("%s does not implement container.ContainerExecInspect", c.ProductName)
+	return nil, fmt.Errorf("%s does not implement container.ContainerExecInspect", ProductName())
 }
 
 // ContainerExecResize changes the size of the TTY of the process
 // running in the exec with the given name to the given height and
 // width.
 func (c *Container) ContainerExecResize(name string, height, width int) error {
-	return fmt.Errorf("%s does not implement container.ContainerExecResize", c.ProductName)
+	return fmt.Errorf("%s does not implement container.ContainerExecResize", ProductName())
 }
 
 // ContainerExecStart starts a previously set up exec instance. The
 // std streams are set up.
 func (c *Container) ContainerExecStart(name string, stdin io.ReadCloser, stdout io.Writer, stderr io.Writer) error {
-	return fmt.Errorf("%s does not implement container.ContainerExecStart", c.ProductName)
+	return fmt.Errorf("%s does not implement container.ContainerExecStart", ProductName())
 }
 
 // ExecExists looks up the exec instance and returns a bool if it exists or not.
 // It will also return the error produced by `getConfig`
 func (c *Container) ExecExists(name string) (bool, error) {
-	return false, fmt.Errorf("%s does not implement container.ExecExists", c.ProductName)
+	return false, fmt.Errorf("%s does not implement container.ExecExists", ProductName())
 }
 
 // docker's container.copyBackend
@@ -95,19 +94,19 @@ func (c *Container) ExecExists(name string) (bool, error) {
 // specified path in the container identified by the given name. Returns a
 // tar archive of the resource and whether it was a directory or a single file.
 func (c *Container) ContainerArchivePath(name string, path string) (content io.ReadCloser, stat *types.ContainerPathStat, err error) {
-	return nil, nil, fmt.Errorf("%s does not implement container.ContainerArchivePath", c.ProductName)
+	return nil, nil, fmt.Errorf("%s does not implement container.ContainerArchivePath", ProductName())
 }
 
 // ContainerCopy performs a deprecated operation of archiving the resource at
 // the specified path in the container identified by the given name.
 func (c *Container) ContainerCopy(name string, res string) (io.ReadCloser, error) {
-	return nil, fmt.Errorf("%s does not implement container.ContainerCopy", c.ProductName)
+	return nil, fmt.Errorf("%s does not implement container.ContainerCopy", ProductName())
 }
 
 // ContainerExport writes the contents of the container to the given
 // writer. An error is returned if the container cannot be found.
 func (c *Container) ContainerExport(name string, out io.Writer) error {
-	return fmt.Errorf("%s does not implement container.ContainerExport", c.ProductName)
+	return fmt.Errorf("%s does not implement container.ContainerExport", ProductName())
 }
 
 // ContainerExtractToDir extracts the given archive to the specified location
@@ -117,13 +116,13 @@ func (c *Container) ContainerExport(name string, out io.Writer) error {
 // be an error if unpacking the given content would cause an existing directory
 // to be replaced with a non-directory and vice versa.
 func (c *Container) ContainerExtractToDir(name, path string, noOverwriteDirNonDir bool, content io.Reader) error {
-	return fmt.Errorf("%s does not implement container.ContainerExtractToDir", c.ProductName)
+	return fmt.Errorf("%s does not implement container.ContainerExtractToDir", ProductName())
 }
 
 // ContainerStatPath stats the filesystem resource at the specified path in the
 // container identified by the given name.
 func (c *Container) ContainerStatPath(name string, path string) (stat *types.ContainerPathStat, err error) {
-	return nil, fmt.Errorf("%s does not implement container.ContainerStatPath", c.ProductName)
+	return nil, fmt.Errorf("%s does not implement container.ContainerStatPath", ProductName())
 }
 
 // docker's container.stateBackend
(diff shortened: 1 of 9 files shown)


lib/apiservers/engine/backends/system.go
```diff
@@ -15,31 +15,148 @@
...
 }
 
+type ContainerStatus struct {
+	count      int
+	numRunning int
+	numStopped int
+	numPaused  int
+}
+
+var (
```
caglar10ur: const?

lib/apiservers/engine/backends/system.go
```diff
@@ -15,31 +15,148 @@
...
+
 func (s *System) SystemInfo() (*types.Info, error) {
-	Driver := "Portlayer Storage"
-	IndexServerAddress := "https://index.docker.io/v1/"
-	ServerVersion := "0.0.1"
-	Name := "VIC"
+	// Set VIC specific data
+	Driver := s.ProductName + " Backend Engine"
+	IndexServerAddress := "registry-1.docker.io"
+	ServerVersion := "0.3"
```
caglar10ur: Same for my comment on the other PR, we should make this a file or a build tag

lib/apiservers/engine/backends/system.go
```diff
@@ -15,31 +15,148 @@
...
+)
+
 func (s *System) SystemInfo() (*types.Info, error) {
-	Driver := "Portlayer Storage"
-	IndexServerAddress := "https://index.docker.io/v1/"
-	ServerVersion := "0.0.1"
-	Name := "VIC"
+	// Set VIC specific data
+	Driver := s.ProductName + " Backend Engine"
+	IndexServerAddress := "registry-1.docker.io"
```
caglar10ur: make this const

lib/apiservers/engine/backends/system.go
```diff
@@ -15,31 +15,148 @@
...
+		ContainersRunning:  containerStatus.numRunning,
+		ContainersPaused:   containerStatus.numPaused,
+		ContainersStopped:  containerStatus.numStopped,
+		Images:             numImages,
+		SystemStatus:       systemStatus,
+		Debug:              (os.Getenv("DEBUG") != ""),
+		NGoroutines:        runtime.NumGoroutine(),
+		SystemTime:         time.Now().Format(time.RFC3339Nano),
+		ExecutionDriver:    Driver,
+		LoggingDriver:      "json-file",
```
caglar10ur: What is json-file?

lib/apiservers/engine/backends/system.go
```diff
@@ -15,31 +15,148 @@
...
+		CgroupDriver:       "N/A",
+		// FIXME: Get this info once we have event listening service
+		//	NEventsListener    int
+		DockerRootDir:    "N/A",
+		ClusterStore:     "",
+		ClusterAdvertise: "",
+		HTTPProxy:        sockets.GetProxyEnv("http_proxy"),
+		HTTPSProxy:       sockets.GetProxyEnv("https_proxy"),
+		NoProxy:          sockets.GetProxyEnv("no_proxy"),
```
caglar10ur: We don't need a package for this, just call os.GetEnv :)

isos/base.sh
```diff
@@ -68,7 +68,7 @@ cp $DIR/base/*.repo $(rootfs_dir $PKGDIR)/etc/yum.repos.d/
 cp $DIR/base/yum.conf $(rootfs_dir $PKGDIR)/etc/yum/
 
 # install the core packages
-yum_cached -c $cache -u -p $PKGDIR install filesystem coreutils bash linux-esx --nogpgcheck -y
+yum_cached -c $cache -u -p $PKGDIR install filesystem coreutils bash linux-esx photon-release --nogpgcheck -y
```
hickeng: This should be added into the appliance-staging.sh, not the base. The entire initramfs is blown away after tether launches in the bootstrap.

lib/apiservers/engine/backends/system.go
```diff
@@ -15,31 +15,148 @@
...
+	systemStatus := make([][2]string, 1)
+	systemStatus[0][0] = s.ProductName + " Backend Engine"
+	if pingPortlayer(plClient) {
+		systemStatus[0][1] = "RUNNING"
+	} else {
+		systemStatus[0][1] = "STOPPED"
+	}
+
+	// Retrieve info about the VCH from Exec port layer
+	//FIXME: Write the portlayer code to do this
```
hickeng: This doesn't need portlayer code, it's part of the tether/extraconfig. You can reload a portion of extraconfig and get the current status for all components.

lib/apiservers/engine/backends/system.go
```diff
@@ -15,31 +15,148 @@
...
+		//	MemoryLimit        bool
+		//	SwapLimit          bool
+		KernelMemory: false,
+		//	IPv4Forwarding     bool
+		//	BridgeNfIptables   bool
+		//	BridgeNfIP6tables  bool `json:"BridgeNfIp6tables"`
+	}
+
+	info.Plugins.Network = append(info.Plugins.Network, "N/A")
+	info.Plugins.Volume = append(info.Plugins.Volume, "N/A")
```
hickeng: We should use this to identify the possible `locations` that can be optionally specified for volume create and eventually the free space. These are the map keys in VirtualContainerHostConfigSpec.VolumeLocations.

lib/apiservers/engine/backends/system.go
hickeng: Why do this via the environment at all - it's not likely to be set in the appliance.
If we need a proxy then it should be in the VCH config struct so that imagec can make use of it as well.

lib/apiservers/engine/backends/system.go
hickeng: Suggest we change this - logging driver doesn't really exist for us but is one of the first places where we should consider extensibility - being able to log to syslogd (which allows LogInsight) is very useful enterprise function.

lib/apiservers/engine/backends/system.go
sflxn: Good point.

lib/apiservers/engine/backends/system.go
sflxn: We do.  We need to clean up both errors and strings for both the personality and port layer server.  I don't want to do that with this PR or operation.  I have another GI to clean up strings in the personality server.

lib/apiservers/engine/backends/system.go
sflxn: Yeah.

lib/apiservers/engine/backends/system.go
sflxn: json-file appears to be the default Docker logging driver.

lib/apiservers/engine/backends/system.go
sflxn: I was returning just the default docker info.  I'll move those to the "need to get from portlayer" section.

lib/apiservers/engine/backends/system.go
sflxn: We could get the info ourselves, but this is the personality server.  Should it even know about any VMW specific APIs?

lib/apiservers/engine/backends/system.go
sflxn: I thought we didn't want to leak information about the vsphere environment.

lib/apiservers/engine/backends/system.go
hickeng: @sflxn It's not a vSphere API, it's a VIC package that's used for component configuration and status publishing.

lib/apiservers/engine/backends/container.go
```diff
@@ -64,32 +63,32 @@ const (
 
 // ContainerExecCreate sets up an exec in a running container.
 func (c *Container) ContainerExecCreate(config *types.ExecConfig) (string, error) {
-	return "", fmt.Errorf("%s does not implement container.ContainerExecCreate", c.ProductName)
+	return "", fmt.Errorf("%s does not implement container.ContainerExecCreate", ProductName())
```
hickeng: @sflxn All of these should be pulled out into a message file - it shouldn't take long. The intent is that there is an actual github issue link included for unimplemented features.

lib/apiservers/engine/backends/system.go
```diff
@@ -87,5 +198,73 @@ func (s *System) UnsubscribeFromEvents(chan interface{}) {
...
+	return "", "", fmt.Errorf("%s does not implement System.AuthenticateToRegistry", ProductName())
+}
+
+// Utility functions
+
+// Returns the name of the OS on the appliance vm.  Very likely Photon OS.  This
+// function takes the place of Docker's function that relies on other packages.
+// Since VIC will likely run on either Photon OS or devbox (ubuntu), we can
+// make some assumptions that avoid having to pull in other packages.
+func getOperatingSystem() string {
```
hickeng: This will do for now, but reporting the OS on the appliance is pointless in reality - it tells the user nothing about what they can actually run in containers. Luckily the bootstrap image and appliance are currently on the same version, but we cannot expect that to remain the case, so this will need to become significantly more involved in the future as it'll be reporting what the possible operating systems/variants are.

lib/apiservers/engine/backends/system.go
```diff
@@ -15,31 +15,142 @@
...
+		//	IPv4Forwarding     bool
+		//	BridgeNfIptables   bool
+		//	BridgeNfIP6tables  bool `json:"BridgeNfIp6tables"`
+		HTTPProxy:  "N/A",
+		HTTPSProxy: "N/A",
+		NoProxy:    "N/A",
+	}
+
+	info.Plugins.Network = append(info.Plugins.Network, "N/A")
+	info.Plugins.Volume = append(info.Plugins.Volume, "N/A")
```
hickeng: I think this is the correct location to report the available volume locations that can be specified.

hickeng:
@sflxn I'm okay with this going in as is for TP3 - I'd prefer it to have the volume locations (this is simply the keys from the VirtualContainerHostConfigSpec.VolumeLocations map) if possible.

lib/apiservers/engine/backends/container.go
sflxn: #1099.  It won't take long, but I do have another higher priority issues to work on.  Also, if we're buying into the small commits regime, then we should have the strings be another PR.  It's on my list of items to work on immediately after our TP3 work.

cgtexmex:
Help me to understand all the error message changes -- that's not required for info is it?  Is this change addressing two things?  Also...if the error messages aren't the way we want to go longer term do we really need this interim step?

sflxn:
The error message thing does not address info.  I just wanted to centralize how we got productflags the way we do constants and globals in vicbackends to follow the same pattern we've been doing.  It started with trying to address comments about constants, which kinda balloon while I was waiting on our CI to work again.

mdubya66:
lgtm

hickeng:
LGTM
Viable placeholder data.

matthewavery:
lgtm

PR #1143, opened 2016-06-18
Adding mapping for vsphere networks

### LINKED ISSUE #454: Map existing vsphere network in installer
It is up to the viadmin to determine which vsphere networks are available to a VCH for direct use - meaning that containers can be attached to them directly. All configuration of this type is done via vic-machine, aka the installer.
To reflect default docker behaviour we should enable mapping of the VCH 'external' network by default - this is as close as we can come to -net=host.

On completion:
* the VCH 'external' network is present in 'docker network ls' and containers can reference it via '-net=host'
  * net=host is simply the obvious naming to me - feel free to object/propose alternatives

hmahmood:
Fixes #454 

Some of this conflicts with #1127 , which I will merge into this PR once that is closed.


cmd/vic-machine/create/create.go
```diff
@@ -114,18 +121,33 @@ func (c *Create) Flags() []cli.Flag {
...
 		cli.StringFlag{
+			Name:        "client-network",
+			Value:       "",
+			Usage:       "The client network (restricts DOCKER_API access to this network)",
+			Destination: &c.ClientNetworkName,
+		},
+		cli.StringSliceFlag{
+			Name:  "container-network",
+			Value: &c.containerNetworks,
+			Usage: "Networks that containers can use",
```
hickeng: If there's a Description field that should be used for the current string. The Usage should include the format specification and not make people experiment to get error output with that detail.

cmd/vic-machine/create/create.go
```diff
@@ -176,6 +198,166 @@ func (c *Create) Flags() []cli.Flag {
...
+	vnet := mapped[0]
+	if vnet == "" {
+		if vnet == "" {
+			return "", nil, cli.NewExitError(fmt.Sprintf("Container network not specified"), 1)
+		}
+	}
+
+	gw, subnet, err := net.ParseCIDR(mapped[1])
+	if err != nil {
+		return "", nil, cli.NewExitError(fmt.Sprintf("Invalid gateway %s specified for container network %s", mapped[1], vnet), 1)
```
hickeng: This should mention that it expects the gateway in CIDR format, and an example.

cmd/vic-machine/create/create.go
```diff
@@ -176,6 +198,166 @@ func (c *Create) Flags() []cli.Flag {
...
+}
+
+func parseGatewaySpec(cg string) (string, *net.IPNet, error) {
+	if cg == "" {
+		return "", nil, cli.NewExitError("container network gateway is empty", 1)
+	}
+
+	mapped := strings.Split(cg, ":")
+	if len(mapped) != 2 {
+		return "", nil, cli.NewExitError(fmt.Sprintf("Container network gateway %s not in VSPHERE-NETWORK:GATEWAY format", cg), 1)
```
hickeng: Worth noting that gateway is IPv4 here otherwise people will wonder why their IPv6 is causing this message.

cmd/vic-machine/create/create.go
```diff
@@ -176,6 +198,166 @@ func (c *Create) Flags() []cli.Flag {
...
+
+func (c *Create) processContainerNetworkGateways() (map[string]net.IPNet, error) {
+	gws := make(map[string]net.IPNet)
+	for _, cg := range c.containerNetworksGateway {
+		vnet, gw, err := parseGatewaySpec(cg)
+		if err != nil {
+			return nil, err
+		}
+
+		if _, ok := gws[vnet]; ok {
```
hickeng: Is this actually a problem if we allow for IP ranges. Is it viable to have:
```
netA: 192.168.0.100-110
netB: 192.168.0.110-120
netC: 192.168.0.120-130
```

all with the gateway `192.168.0.1`?

cmd/vic-machine/create/create.go
```diff
@@ -176,6 +198,166 @@ func (c *Create) Flags() []cli.Flag {
...
+
+func parseIPRangeSpec(cp string) (vnet string, ipr metadata.IPRange, err error) {
+	if cp == "" {
+		err = cli.NewExitError("container network ip range is empty", 1)
+		return
+	}
+
+	mapped := strings.Split(cp, ":")
+	if len(mapped) != 2 {
+		err = cli.NewExitError(fmt.Sprintf("Container network ip range %s not in VSPHERE-NETWORK:IP-RANGE format", cp), 1)
```
hickeng: what is IP-RANGE format?

cmd/vic-machine/create/create.go
```diff
@@ -176,6 +198,166 @@ func (c *Create) Flags() []cli.Flag {
 	return flags
 }
 
+func zeroIPNet(n net.IPNet) bool {
```
hickeng: There is a reflect Value.IsNil function - suggest naming this similarly as the intent is identical, just not applied to a reflect.Value

cmd/vic-machine/create/create.go
```diff
@@ -176,6 +198,166 @@ func (c *Create) Flags() []cli.Flag {
...
+
+	mapped := strings.Split(cp, ":")
+	if len(mapped) != 2 {
+		err = cli.NewExitError(fmt.Sprintf("Container network ip range %s not in VSPHERE-NETWORK:IP-RANGE format", cp), 1)
+		return
+	}
+
+	vnet = mapped[0]
+	if vnet == "" {
+		err = cli.NewExitError(fmt.Sprintf("Container network not specified"), 1)
```
hickeng: vSphere network not specified.

cmd/vic-machine/create/create.go
```diff
@@ -176,6 +198,166 @@ func (c *Create) Flags() []cli.Flag {
...
+	if r == nil {
+		err = cli.NewExitError(fmt.Sprintf("Invalid ip range for container network %s", vnet), 1)
+		return
+	}
+
+	ipr = *r
+	return
+}
+
+func (c *Create) processContainerNetworkIPRanges() (map[string][]metadata.IPRange, error) {
```
hickeng: Sort of thing we should definitely have a unit test for.

cmd/vic-machine/create/create.go
```diff
@@ -176,6 +198,166 @@ func (c *Create) Flags() []cli.Flag {
...
+			return cli.NewExitError(fmt.Sprintf("Container network mapping %s not in VSPHERE-NETWORK[:VIC-NETWORK] format", cn), 1)
+		}
+
+		vnet := mapped[0]
+		vicnet := vnet
+		if len(mapped) > 1 {
+			vicnet = mapped[1]
+		}
+
+		if vicnet == "bridge" {
```
hickeng: How do we allow specification of IP range/gateway for bridge?

hickeng: how do we specify IP range/gateway for bridge, and how do we avoid conflict with other addresses? Do we need to calculate a non-conflicting scope automatically?

cmd/vic-machine/create/create.go
```diff
@@ -176,6 +198,166 @@ func (c *Create) Flags() []cli.Flag {
...
+			return cli.NewExitError("Cannot use reserved name bridge for container network", 1)
+		}
+
+		gw := gws[vnet]
+		r := pools[vnet]
+
+		if len(r) != 0 && zeroIPNet(gw) {
```
hickeng: Good choice - considered if we should try to generate a sane default for a gateway based on range, but no can do. Easier to generate range based on gateway!

hickeng: r & gw could be nil

cmd/vic-machine/create/create.go
```diff
@@ -176,6 +198,166 @@ func (c *Create) Flags() []cli.Flag {
...
+			return cli.NewExitError(fmt.Sprintf("IP range specified without gateway for container network %s", vnet), 1)
```
hickeng: Good choice - considered if we should try to generate a sane default for a gateway based on range, but no can do. Easier to generate range based on gateway!

cmd/vic-machine/create/create.go
```diff
@@ -176,6 +198,166 @@ func (c *Create) Flags() []cli.Flag {
...
+			if gw.Contains(i.FirstIP) && gw.Contains(i.LastIP) {
+				continue
+			}
+
+			return cli.NewExitError(fmt.Sprintf("IP range %s is not in subnet %s", i, gw), 1)
+		}
+
+		c.MappedNetworks[vicnet] = &metadata.ContainerNetwork{
+			Name:    vnet,
+			Gateway: gw,
```
hickeng: What does it mean for both gw and r to be nil? Is that DHCP?

cmd/vic-machine/create/create.go
```diff
@@ -196,8 +378,8 @@ func (c *Create) processParams() error {
 		return cli.NewExitError("key cert should be specified at the same time", 1)
 	}
 
-	if c.ExternalNetworkName == "" {
-		c.ExternalNetworkName = "VM Network"
+	if c.ClientNetworkName == "" {
```
hickeng: Let's have this in the argument definition as the default value rather than inline here.

lib/metadata/ip.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"bytes"
+	"fmt"
+	"math"
+	"net"
+	"strconv"
+	"strings"
+)
+
+type IPRange struct {
```
hickeng: You could remove these annotations and move this into /pkg - there's nothing here that's unique to VIC and you don't need the annotations for the fields to be serialized. They will inherit their parent scope.

cmd/vic-machine/create/create.go
```diff
@@ -17,8 +17,10 @@ package create
 import (
```
hickeng: Most of these changes should really live in validator.go rather than in create.go. They will be relevant for later operations such as `update`

lib/metadata/network_interface.go
```diff
@@ -54,7 +59,10 @@ type ContainerNetwork struct {
...
 
-	// The IP range for this network
-	FirstIP net.IP `vic:"0.1" scope:"read-only" key:"first_ip"`
-	LastIP  net.IP `vic:"0.1" scope:"read-only" key:"last_ip"`
+	// The IP ranges for this network
+	Pools []IPRange `vic:"0.1" scope:"read-only" key:"pools"`
+
+	// A reference to the vsphere network object
+	NetworkRef types.ManagedObjectReference `vic:"1.0" scope:"read-only" key:"network"`
+	Network    object.NetworkReference
```
hickeng: Not a good idea to have this here - it will serialize, but when it's deserialized it won't have a valid Client.

lib/portlayer/network/context.go
```diff
@@ -68,9 +68,28 @@ func NewContext(bridgePool net.IPNet, bridgeMask net.IPMask) (*Context, error) {
 	if err != nil {
 		return nil, err
 	}
-
 	s.builtin = true
 	ctx.defaultScope = s
+
+	// add any external networks
+	for nn, n := range portlayer.Config.ContainerNetworks {
+		if nn == "bridge" {
```
hickeng: With my current PR the tether takes care of bringing the bridge interface up and setting it's IP - does the PL need to do specific handling for it, or can it be considered just another container network at this point?
(not considered the port forwarding with this comment).

lib/portlayer/network/context.go
```diff
@@ -191,23 +215,27 @@ func (c *Context) newBridgeScope(id, name string, subnet *net.IPNet, gateway net
...
+	if ipam != nil && len(ipam.pools) > 0 {
+		if isUnspecifiedSubnet(subnet) || gateway.IsUnspecified() {
+			return nil, fmt.Errorf("ipam cannot be specified without gateway and subnet for external network")
+		}
 	}
 
-	if isUnspecifiedSubnet(subnet) || gateway.IsUnspecified() {
-		return nil, fmt.Errorf("neither subnet nor gateway specified for external network")
+	if !isUnspecifiedSubnet(subnet) {
+		// cannot overlap with the default bridge pool
```
hickeng: What happens if it overlaps with other scopes and we connect the container to those overlapping scopes?

lib/tether/ops_linux.go
```diff
@@ -213,7 +213,7 @@ func assignIP(link netlink.Link, endpoint *metadata.NetworkEndpoint) (bool, erro
 	}
 
 	// Set IP address if it's specified - this is now named for for the network role rather than the nic
-	if len(endpoint.Static.IP) > 0 {
+	if endpoint.Static.IP != nil && !endpoint.Static.IP.IsUnspecified() {
```
hickeng: This is actually an issue for me - Static could be nil. Actually surprised I'm not seeing issues with that in the tether already.

hickeng:
@hmahmood I've run through this, but was less able to comment during the network context portion and scope handling. If you need me to go through that with a fine tooth comb let me know - I can but it'll take longer.

cmd/vic-machine/create/create.go
hmahmood: IP range specification for bridge networks is not part of vic-machine; that is done with `docker network create`.

cmd/vic-machine/create/create.go
hmahmood: Yes that would be dhcp

lib/metadata/ip.go
hmahmood: Yes we probably should remove them, although in some cases we need custom keys.

I will move this to pkg.

lib/metadata/network_interface.go
hmahmood: This, along with code in portlayer.go, I expect to see replaced with code from your PR.

lib/portlayer/network/context.go
hmahmood: See https://github.com/hmahmood/vic/blob/3a4a029784c94a94d3b5bbfcad8f339722f65fed/lib/portlayer/network/context.go#L245, which is called from `newScopeCommon`.

lib/tether/ops_linux.go
hmahmood: I was having some difficulty with that -- Static would always been non-nil even when the network was using DHCP. I will put it back since it is a safe bet anyway.

cmd/vic-machine/create/create.go
hickeng: I mean for the default bridge that is created automatically

lib/tether/ops_linux.go
hickeng: Dug into this - it's actually something I'd looked at at the beginning of #1127. It's related to handling in extraconfig that I've not hand time to fix. There's now a test being skipped to assert that unspecified structures remain nil.
I tried switching to IsUnspecified but in the end I had to go back the len(IP) > 0.

hmahmood: You have to do both, i.e. len(IP) > 0 and isUnspecified(). isUnspecified will check for zero ip addresses as well.

cmd/vic-machine/create/create.go
hmahmood: So I made some headway into implementing port mapping over the weekend, and added some command line arguments for that to vic-machine. Have not added overlap checks in vic-machine for this, but the portlayer will catch this when it tries to create the networks. Maybe too late, so probably better to have some checks in the vic-machine as well.

cmd/vic-machine/create/create.go
hmahmood: Yes they could all have the same gateway.

hickeng: @hmahmood in that case, should this be an error?

cmd/vic-machine/validate/validator.go
```diff
@@ -301,10 +301,15 @@ func (v *Validator) network(ctx context.Context, input *data.Data, conf *metadat
...
+	// FIXME: The gateway for the bridge network
+	//        should be the gateway address of the
+	//        default bridge network. The bridge
+	//        network pool should be specified
+	//        on the vic-machine command line.
+	gateway, ok := input.MappedNetworksGateways["bridge"]
 	if !ok {
-		ip, mask, _ := net.ParseCIDR("172.16.0.1/24")
-		gateway = &net.IPNet{
+		ip, mask, _ := net.ParseCIDR("172.16.0.1/16")
```
hickeng: I only discovered this recently. There is a netlink.ParseIPNet that will preserve the IP address in the resulting net so we could use this directly without lines 312-315

lib/portlayer/network/context.go
```diff
@@ -99,51 +119,52 @@ func isUnspecifiedSubnet(n *net.IPNet) bool {
...
 
-	subSpaces, err := reservePools(space, ipam)
-	if err != nil {
-		return nil, err
-	}
-
-	ipam.spaces = subSpaces
-
-	if gateway.IsUnspecified() {
+	if gateway.IsUnspecified() && len(ipam.spaces) > 0 {
```
hickeng: gateway.IsUnspecified will not work currently - the IP address unpacked from Config will be an empty array rather than {0, 0, 0, 0} - see https://play.golang.org/p/gU9_3_0y-n
I suggest for now that we write our own IsUnspecified as with IsUnspecifiedSubnet to check for nil, len(x) == 0, etc.

lib/portlayer/network/context.go
```diff
@@ -430,9 +459,13 @@ func (c *Context) BindContainer(h *exec.Handle) ([]*Endpoint, error) {
 	endpoints := con.Endpoints()
 	for _, e := range endpoints {
 		ne := h.ExecConfig.Networks[e.Scope().Name()]
-		ne.Static = &net.IPNet{
-			IP:   e.IP(),
-			Mask: e.Scope().Subnet().Mask,
+		ne.Static = nil
+		ip := e.IP()
```
hickeng: not for this PR, but we may be able to make consumption simpler by returning an IPNet that preserves the full IP address.

hickeng:
LGTM

caglar10ur:
Still going over the code but I would be really happy if we make "bridge" a const and use it instead of hardcoding it everywhere.

hickeng:
@caglar10ur not entirely sure if it addresses the issue - but there's a config field BridgeNetwork which is the string name of the NetworkEndpoint entry to use for the bridge.

hmahmood:
There is only one instance of "bridge" in the network PL code (tests have more). Using that BridgeNetwork is different.

caglar10ur:
Only nitpicks otherwise LGTM

Issue #915, opened 2016-05-31
docker volume create


fdawg4l:
Support creation of anonymous volumes per container.

matthewavery:
@hickeng I am not aware of separated issues. I believe this is the only one for Volume create at the moment.

PR #1152, opened 2016-06-20
Docker volume Create stubbing

matthewavery:
referencese #915 

This adds stubbing to the docker engine side of volume create. It does not involve random naming yet. This is in conjunction to the storage portlayer additions that @fdawg4l is making.

### DIFF
--- a/lib/apiservers/engine/backends/volume.go
+++ b/lib/apiservers/engine/backends/volume.go
@@ -15,32 +15,77 @@
 package vicbackends
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
+	"strconv"
 
 	derr "github.com/docker/docker/errors"
 
 	"github.com/docker/engine-api/types"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/storage"
+	"github.com/vmware/vic/lib/apiservers/portlayer/models"
 	"github.com/vmware/vic/pkg/trace"
 )
 
+//NOTE: These might be moved to a utility package once there are multiple personalities
+const optsVolumeStoreKey string = "VolumeStore"
+const optsCapacityKey string = "Capacity"
+const dockerMetadataModelKey string = "DockerMetaData"
+
+//Volume : struct which defines the docker personalities view of a Volume
 type Volume struct {
 	ProductName string
 }
 
+type volumeMetadata struct {
+	Driver     string
+	DriverOpts map[string]string
+	Name       string
+	Labels     map[string]string
+}
+
+//Volumes : docker personality implementation for VIC
 func (v *Volume) Volumes(filter string) ([]*types.Volume, []string, error) {
 	return nil, make([]string, 0), fmt.Errorf("%s does not implement volume.Volumes", v.ProductName)
 }
 
+//VolumeInspect : docker personality implementation for VIC
 func (v *Volume) VolumeInspect(name string) (*types.Volume, error) {
 	return nil, fmt.Errorf("%s does not implement volume.VolumeInspect", v.ProductName)
 }
 
+//VolumeCreate : docker personality implementation for VIC
 func (v *Volume) VolumeCreate(name, driverName string, opts, labels map[string]string) (*types.Volume, error) {
-	return nil, fmt.Errorf("%s does not implement volume.VolumeCreate", v.ProductName)
+	defer trace.End(trace.Begin("Volume.VolumeCreate"))
+	result := &types.Volume{}
+
+	//TODO: design a way to have better error returns.
+
+	client := PortLayerClient()
+	if client == nil {
+		return nil, derr.NewErrorWithStatusCode(fmt.Errorf("Failed to get a portlayer client"), http.StatusInternalServerError)
+	}
+
+	//TODO: support having another driver besides vsphere.
+	//assign the values of the model to be passed to the portlayer handler
+	model, varErr := translateInputsToPortlayerRequestModel(name, driverName, opts, labels)
+	if varErr != nil {
+		return result, derr.NewErrorWithStatusCode(fmt.Errorf("Bad Driver Arg: %s", varErr), http.StatusBadRequest)
+	}
+
+	//TODO: setup name randomization if name == nil
+
+	res, err := client.Storage.CreateVolume(storage.NewCreateVolumeParams().WithVolumeRequest(&model))
+	if err != nil {
+		return result, derr.NewErrorWithStatusCode(fmt.Errorf("Server error from Portlayer: %s", err), http.StatusInternalServerError)
+	}
+
+	result = fillDockerVolumeModel(res.Payload, labels)
+	return result, nil
 }
 
+//VolumeRm : docker personality for VIC
 func (v *Volume) VolumeRm(name string) error {
 	defer trace.End(trace.Begin("Volume.VolumeRm"))
 
@@ -58,7 +103,68 @@ func (v *Volume) VolumeRm(name string) error {
 		if _, ok := err.(*storage.RemoveVolumeConflict); ok {
 			return derr.NewRequestConflictError(fmt.Errorf("Volume is in use"))
 		}
-		return derr.NewErrorWithStatusCode(fmt.Errorf("Server error form portlayer"), http.StatusInternalServerError)
+		return derr.NewErrorWithStatusCode(fmt.Errorf("Server error from portlayer: %s", err), http.StatusInternalServerError)
 	}
 	return nil
+}
+
+//Utility Functions
+
+func fillDockerVolumeModel(volume *models.VolumeResponse, labels map[string]string) *types.Volume {
+	dockerVol := types.Volume{
+		Driver:     volume.Driver,
+		Name:       volume.Name,
+		Labels:     labels,
+		Mountpoint: volume.SelfLink,
+	}
+	return &dockerVol
+}
+
+func validateDriverArgs(args map[string]string, model *models.VolumeRequest) error {
+	//volumestore name validation
+	storeName, ok := args[optsVolumeStoreKey]
+	if !ok {
+		storeName = "default"
+	}
+	model.Store = storeName
+
+	//capacity validation
+	capstr, ok := args[optsCapacityKey]
+	if !ok {
+		model.Capacity = -1
+		return nil
+	}
+	capacity, convErr := strconv.ParseInt(capstr, 10, 64)
+	if convErr != nil {
+		model.Capacity = -1
+		return fmt.Errorf("Capacity must be an integer value. The unit is GB.: %s", convErr)
+	}
+	model.Capacity = int64(capacity)
+	return nil
+}
+
+func translateInputsToPortlayerRequestModel(name, driverName string, opts, labels map[string]string) (models.VolumeRequest, error) {
+	model := models.VolumeRequest{
+		Driver:     driverName,
+		DriverArgs: opts,
+		Name:       name,
+	}
+	metadata := createVolumeMetadata(&model, labels)
+	model.Metadata = make(map[string]string)
+	model.Metadata["dockerMetaData"] = metadata
+	if err := validateDriverArgs(opts, &model); err != nil {
+		return model, err
+	}
+	return model, nil
+}
+
+func createVolumeMetadata(model *models.VolumeRequest, labels map[string]string) string {
+	metadata := volumeMetadata{
+		Driver:     model.Driver,
+		DriverOpts: model.DriverArgs,
+		Name:       model.Name,
+		Labels:     labels,
+	}
+	result, _ := json.Marshal(metadata)
+	return string(result)
 }
--- a/lib/apiservers/portlayer/swagger.yml
+++ b/lib/apiservers/portlayer/swagger.yml
@@ -193,6 +193,66 @@ paths:
           description: "error"
           schema:
             $ref: "#/definitions/Error"
+  /storage/volumes/:
+    get:
+      description: "Get a list of available volumes"
+      operationId: ListVolumes
+      tags: ["storage"]
+      consumes:
+        - application/json
+        - application/octet-stream
+      produces:
+        - application/json
+      parameters:
+        - name: id
+          type: string
+          in: query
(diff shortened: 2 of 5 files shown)


lib/apiservers/portlayer/swagger.yml
```diff
@@ -193,6 +193,64 @@ paths:
...
+        - application/json
+        - application/octet-stream
+      produces:
+        - application/json
+      parameters:
+        - name: Volume
+          required: true
+          in: body
+          schema:
+            $ref: "#/definitions/Volume"
```
matthewavery: Change this to a list of paramters. Specifically. ID, Metdata, driver.

lib/apiservers/engine/backends/volume.go
```diff
@@ -17,30 +17,70 @@ package vicbackends
...
+	defer trace.End(trace.Begin("Volume.VolumeCreate"))
+	result := &types.Volume{}
+
+	host, err := guest.UUID()
+	if host == "" {
+		host, err = os.Hostname()
+	}
+	if err != nil {
+		return result,
+			derr.NewErrorWithStatusCode(fmt.Errorf("Volume.CreateVolume got unexpected error getting hostname"),
```
jzt: Should you also display the error?

```go
if err != nil {
 		return result,
 			derr.NewErrorWithStatusCode(fmt.Errorf("Volume.CreateVolume got unexpected error getting hostname: %s", err),
				http.StatusInternalServerError)
 }
```
Also, we should start making a concerted effort to make output more friendly, ie: "Unexpected error getting hostname" instead of "Volume.CreateVolume ..."

matthewavery: Yeah I was also going to advocate for that 👍 . The general errors we have from some of the other commands now really make it hard to know exactly what happened. I think we should look at expanding the responses of our swagger calls since our personality servers are really in charge of interpretting portlayer error codes to the user. So naturally I think we need more to know what is happening. LMKWYT

lib/apiservers/portlayer/swagger.yml
```diff
@@ -859,3 +920,46 @@ definitions:
...
+  VolumeRequest:
+    type: object
+    required:
+      - Name
+      - Store
+      - Driver
+    properties:
+      Name:
+        type: string
+      Driver:
```
matthewavery: the portlayer does not care about the driver name. Remove this from the inputs. Since the portlayer is our driver. Per conversation with george (6/22)

lib/apiservers/portlayer/swagger.yml
```diff
@@ -859,3 +920,46 @@ definitions:
...
+        type: object
+        additionalProperties:
+          type: string
+  VolumeResponse:
+    type: object
+    required:
+      - Name
+      - SelfLink
+      - Store
+      - Driver
```
matthewavery: this should essentially just return "vsphere" as the value for driver. Might be worthwhile just to have interpretation code for that in the personality and remove driver from here as well. The portlayer does not need to know that a driver specification is necessary.

lib/apiservers/engine/backends/volume.go
```diff
@@ -15,32 +15,79 @@
...
 
+type volumeMetadata struct {
+	Driver string
+	Name   string
+	Labels map[string]string
+}
+
+//Volumes : docker personality implementation for VIC
 func (v *Volume) Volumes(filter string) ([]*types.Volume, []string, error) {
 	return nil, make([]string, 0), fmt.Errorf("%s does not implement volume.Volumes", v.ProductName)
```
jzt: Can you not return `nil` in place of this slice? Or will the client crash or something? Just curious.

lib/apiservers/engine/backends/volume.go
```diff
@@ -58,7 +105,43 @@ func (v *Volume) VolumeRm(name string) error {
...
+	model := models.VolumeRequest{
+		Driver:     driverName,
+		DriverArgs: opts,
+		Name:       name,
+		Store:      host,
+	}
+
+	metadata := createVolumeMetadata(&model, labels)
+	model.Metadata = make(map[string]string)
+	model.Metadata["dockerMetaData"] = metadata
```
jzt: can we make `dockerMetaData` a constant?

lib/apiservers/engine/backends/volume.go
```diff
@@ -58,7 +105,43 @@ func (v *Volume) VolumeRm(name string) error {
...
+	return model
+}
+
+func createVolumeMetadata(model *models.VolumeRequest, labels map[string]string) string {
+	metadata := volumeMetadata{
+		Driver: model.Driver,
+		Name:   model.Name,
+		Labels: labels,
+	}
+	result, _ := json.Marshal(metadata)
```
jzt: Any reason we ignore the error here?

lib/apiservers/portlayer/swagger.yml
matthewavery: On second thought, I will avoid this for now. the refactor would not be too bad. But in the event that we want it eventually it can still be there. It does not adversely effect anything in the design.

lib/apiservers/engine/backends/volume.go
```diff
@@ -58,7 +105,43 @@ func (v *Volume) VolumeRm(name string) error {
...
+	}
+	return &dockerVol
+}
+
+func translateInputsToPortlayerRequestModel(name, host, driverName string, opts, labels map[string]string) models.VolumeRequest {
+	model := models.VolumeRequest{
+		Driver:     driverName,
+		DriverArgs: opts,
+		Name:       name,
+		Store:      host,
```
fdawg4l: Store is the name of the datastore which we want to create a volume on.  Not the hostname or the uuid of the vch.

Think of it this way-  I have 3 datastores in my vsphere environment.  One of them is called `Fast`, the other `Backedup`, and the last `Junk`.

On `Fast`, I (as a VI admin) want my volumes created in `[FastDatastore] /path/to/some/place/relevant/that/i/decided`.
On `Backedup`, I want my volumes created on `[BackedupDatastore] /some/other/location`.
On `Junk`, I want my volumes created on `[JunkDatastore] /junk/folder`.

Anything *beyond* those paths are irrelevant to me.

How does a docker user now specify that stuff?  Im now a docker user and I just want to create a volume and know nothing of paths or datastores or whatever. 

We want to use `volumeStore` (or `Store` in this case) to encompass the datastore + path so the docker user doesnt need to know about the details.  

The VI admin would do the following when setting up the vch

`vic-admin create <....> -volumestore Fast="[FastDatastore] /path/to/where/i/want/volumes`

And the docker user would do the following.
`docker volume create --driver=vsphere --name="nameRelevantToMeTheDockerUser" -o "volumestore=Fast"`

Make sense?

fdawg4l: In case it wasn't clear.  THe VI Admin could specify many volume stores and the docker user could create volumes on any one of them.

fdawg4l: To extend our example, given the following docker user input

> docker volume create --driver=vsphere --name="nameRelevantToMeTheDockerUser" -o "volumestore=Fast" capacity=8G foo=bar something=somethingelse

The call to the portlayer should be 

`VolumeCreate(ctx, "nameRelelvantToMeTheDockerUser", util.VolumeStore("Fast"), 8 * 1024, map[string][]byte{"foo": []byte("bar"), "something":[]byte("somethingelse")})`

lib/apiservers/engine/backends/volume.go
matthewavery: specifically, I am not sure what I would do with an error there. I would not want to panic the server. Perhaps write an error to the logs and have an additional error return on the function? Suggestions?

lib/apiservers/engine/backends/volume.go
matthewavery: yeah, sorry about the magic number.

lib/apiservers/engine/backends/volume.go
matthewavery: That is something left over from the beginning of the file. I do not know for certain if the client will crash or report something nonsensical. Since I did not add this code, I  avoided speculating on it for now. There are no integration tests around it at the moment either.

lib/apiservers/engine/backends/volume.go
matthewavery: Write and all of those things would get passed into the DriverArgs section when being translated to the portlayer hander. Which will negotiate the data before making the VolumeCreate call. So my question is how do we want to handle those args in the personality? George has expressed that no value should be passed into the portlayer as blank. So I talked to him about listing "default" as the value of any required args that are not set by the user. Then those defaults would be specified at deployment time with vic-machine. If not set the handler would talk to the storage layer(in the portlayer) and a "No default set, operation not permitted" type error should be returned and conveyed to the Personality.

matthewavery: tl:dr version -> add checking for the `VolumeStore` and `Capacity` driver args that are needed for the vsphere volume driver.

lib/apiservers/engine/backends/volume.go
```diff
@@ -15,32 +15,92 @@
...
 	"github.com/docker/engine-api/types"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/storage"
+	"github.com/vmware/vic/lib/apiservers/portlayer/models"
+	"github.com/vmware/vic/lib/guest"
 	"github.com/vmware/vic/pkg/trace"
 )
 
+//NOTE: These might be moved to a utility package once there are multiple personalities
+const optsVolumeStoreKey string = "VolumeStore"
+const optsCapacityKey string = "Capacity"
```
hickeng: Should the personality know about Capacity? That seems like something that should be blindly passed through to the PL

lib/apiservers/engine/backends/volume.go
```diff
@@ -15,32 +15,92 @@
...
 
 	"github.com/docker/engine-api/types"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/storage"
+	"github.com/vmware/vic/lib/apiservers/portlayer/models"
+	"github.com/vmware/vic/lib/guest"
 	"github.com/vmware/vic/pkg/trace"
 )
 
+//NOTE: These might be moved to a utility package once there are multiple personalities
+const optsVolumeStoreKey string = "VolumeStore"
```
hickeng: are these the user facing parameters to --opt? ie. --opt VolumeStore=someStoreName?
If so the camel case isn't consistent with the other docker CLI argument styles.

lib/apiservers/engine/backends/volume.go
```diff
@@ -58,7 +118,62 @@ func (v *Volume) VolumeRm(name string) error {
...
+}
+
+func validateDriverArgs(args map[string]string) error {
+	//volumestore name validation
+	if _, ok := args[optsVolumeStoreKey]; !ok {
+		args[optsVolumeStoreKey] = "default"
+	}
+
+	//capacity validation
+	if _, ok := args[optsCapacityKey]; !ok {
```
hickeng: why not write this as below instead of two accesses to the map?:
```
cap, ok := args[optsCapacityKey]
if !ok {
  ...
  return nil
}

if _, convErr := ...

```

lib/apiservers/portlayer/swagger.yml
```diff
@@ -193,6 +193,66 @@ paths:
...
+          required: true
+          in: body
+          schema:
+            $ref: "#/definitions/VolumeRequest"
+      responses:
+        '201':
+          description: "Created"
+          schema:
+            $ref: "#/definitions/VolumeResponse"
+        '500':
```
hickeng: 409 Conflict error for already exists.

lib/apiservers/engine/backends/volume.go
matthewavery: This check can also be moved to the portlayer handler. However, if we do not want to pass anything blank to the portlayer we should be doing the checking on this side.

lib/apiservers/engine/backends/volume.go
matthewavery: They are I will take a look at the cli options and make them look more like docker.

lib/apiservers/engine/backends/volume.go
matthewavery: Good point, I was looking at this and couldn't decide which way to write it.

lib/apiservers/engine/backends/volume.go
```diff
@@ -58,7 +103,68 @@ func (v *Volume) VolumeRm(name string) error {
...
+func validateDriverArgs(args map[string]string, model *models.VolumeRequest) error {
+	//volumestore name validation
+	storeName, ok := args[optsVolumeStoreKey]
+	if !ok {
+		storeName = "default"
+	}
+	model.Store = storeName
+
+	//capacity validation
+	capstr, _ := args[optsCapacityKey]
```
fdawg4l: Missing the `ok`.

lib/apiservers/engine/backends/volume.go
```diff
@@ -58,7 +103,68 @@ func (v *Volume) VolumeRm(name string) error {
...
+	dockerVol := types.Volume{
+		Driver:     volume.Driver,
+		Name:       volume.Name,
+		Labels:     labels,
+		Mountpoint: volume.SelfLink,
+	}
+	return &dockerVol
+}
+
+func validateDriverArgs(args map[string]string, model *models.VolumeRequest) error {
```
fdawg4l: Go lets you return more than one value.  Consider returning `(*models.VolumeRequest, error)`.

lib/apiservers/engine/backends/volume.go
matthewavery: looks like I need to make my unit test better. It should have caught something like this.

lib/apiservers/engine/backends/volume.go
matthewavery: The function that calls this function has that return signature.

fdawg4l:
LGTM

andrewtchin:
lgtm

matthewavery:
♻️

PR #1153, opened 2016-06-20
Add name resolution support to port-layer

### LINKED ISSUE #84: Container name resolution
There is a fundamental requirement for containers to be able to obtain a handle to another container via symbolic name. We are going to use DNS resolution because it's a completely pervasive standard, the implementation of which could vary (plug in consol or kubernetes name resolution for example). This has been the mechanism of choice since inception given there's no shared filesystem or easy way to manipulate per container resolution in a black box fashion.

Eventually we may wish to serve up a different response based on the origin of the inquiry to provide contextual lookup, without having to to configure default search domains correctly. This would allow us to provide individual dynamic resolution without having to ensure global consistency & lack of conflicts between names.

Initial implementation should be relatively straightforward, simply a dns server within the VCH. Given the potentially very short lifetimes of containers, existing DNS servers are not really ideal - they are not indented for very short lived entries and potentially high churn so we may need to use an embedded DNS server that we can drive directly via APIs.

- [ ] define tests for varied container network topologies that define source, target request, and expected return value

- [ ] look at docker 1.10 and see if the embedded DNS there is usefully abstracted into a consumable package
- [x] look at the other dns packages available as a base

- [x] if docker 1.10 package isn't suitable, how viable is it to extract it into a standalone package
- [ ] select appropriate base package, wrap a main method around it and write the tests to ensure its behaviour meets our requirements

caglar10ur:
This adds a forwarding domain name serving capability to port-layer. The
forwarded queries are cached locally with a pre-defined TTL. The cache
lazily removes the expired items and evicts items randomly when it is
full.

This change set also moves the port-layer-server main function from
swagger generated area to under cmd/ and adds --exclude-main parameter
to swagger call to instruct it to not to re-generate main.go file.

Container name resolution is not implemented and will follow as a
separate PR.

Partially fixes #84

### DIFF
--- a/cmd/port-layer-server/main.go
+++ b/cmd/port-layer-server/main.go
@@ -1,0 +1,85 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"log"
+	"os"
+	"os/signal"
+	"syscall"
+
+	spec "github.com/go-swagger/go-swagger/spec"
+	flags "github.com/jessevdk/go-flags"
+
+	"github.com/vmware/vic/lib/apiservers/portlayer/restapi"
+	"github.com/vmware/vic/lib/apiservers/portlayer/restapi/operations"
+
+	"github.com/vmware/vic/lib/dns"
+)
+
+var (
+	options = dns.ServerOptions{}
+)
+
+func main() {
+	swaggerSpec, err := spec.New(restapi.SwaggerJSON, "")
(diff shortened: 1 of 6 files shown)


dougm:
lgtm

cmd/port-layer-server/main.go
```diff
@@ -0,0 +1,85 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
```
hickeng: Yes!!! Finally this reorg is closer to complete.

caglar10ur: And how about the rest? This pr still needs another lgtm :)

lib/dns/cache.go
```diff
@@ -0,0 +1,155 @@
...
+	return fmt.Sprintf("%s:%s", q.Name, mdns.TypeToString[q.Qtype])
+}
+
+// Add adds dns.Msg to the cache
+func (c *Cache) Add(msg *mdns.Msg) {
+	c.Lock()
+	defer c.Unlock()
+
+	if len(c.m) >= c.capacity {
+		// pick a random key and remove it
```
hmahmood: Why don't we use the ttl to figure out which one to delete?

caglar10ur: That requires a data structure other than map or traversing the whole map. Go randomly walks over maps and their shuffling is O(1) so I just sticked with that. On that note, we are removing the expired TTLs lazily on the Get path.

hmahmood:
lgtm

caglar10ur:
♻️

PR #1155, opened 2016-06-21
Use metadata tempfile to optimize image caching

### LINKED ISSUE #1154: Improve image cache availability and locking

Additionally, access to the cache is disallowed during this process as it remains locked until the update completes. This is problematic in situations where a read access to the cache is attempted which must block until the cache update is complete. In the case of a docker host with hundreds or thousands of images, this wait would be evident.

This process can be improved with two changes: 

1. imagec could be updated to provide information to the docker persona about the image it just pulled. The docker persona can then add this information (metadata) to the cache without having to trigger another request to the portlayer that will have to iterate over the entire layer list to find anything it needs to cache.

2. The full cache update can now be run just once when the docker persona starts, and can lock the cache only when it actually finds something in the list of layers that it needs to add to the cache. This would allow read/write access to the cache while the update is in progress.

jzt:
I could use some feedback in a couple of areas:

1. Is using a temp file for relaying image information from imagec to the docker persona acceptable? Using stdout didn't seem as straightforward to me, as imagec's stdout is shown in the output while images are pulling. If it's possible to use golang's `exec` package to do something like this, and it makes more sense, I'll be glad to do it that way.

2. Right now the persona reads the tempfile as a `[]byte` and then unmarshals it into a `metadata.ImageConfig` struct that it then passes to `AddImage()` in the cache. Since the cache already has to do this unmarshaling at startup, I was thinking about just passing the `[]byte` directly into `AddImage()` and basically let all unmarshaling of bytes into structs occur in the cache layer. 

Fixes #1154, #1158

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -14,7 +14,7 @@ environment:
 ```
 2. Execute drone from the projects root directory:
 
-  `drone exec --trusted -E "secrets.yml" --yaml ".drone-e2e.yml"`
+  `drone exec --trusted -E "secrets.yml" --yaml ".drone.yml"`
 
 Find the documentation for each of the tests here:
 -


lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -60,8 +58,6 @@ func NewImageCache() *ImageCache {
 
 // Update adds new images to the cache
 func (c *ImageCache) Update(client *client.PortLayer) error {
-	c.m.Lock()
-	defer c.m.Unlock()
 
 	CacheNotUpdated = true
```
jzt: @sflxn Do we still need this with the updated locking behavior?

cmd/imagec/imagec.go
```diff
@@ -379,6 +382,17 @@ func WriteImageBlobs(images []*ImageWithMeta) error {
 	if err := os.RemoveAll(destination); err != nil {
 		return fmt.Errorf("Failed to remove download directory: %s", err)
 	}
+
+	if len(images) > 0 && options.imagefile != "" {
```
jzt: This needs to be fixed to cover the case of pulling the same image by a different tag or by the digest. In this current implementation, the image metadata would not get updated with the new tag information. I will consult with @cgtexmex and see if I can come up with something to address this.

lib/apiservers/engine/backends/image.go
```diff
@@ -147,8 +155,22 @@ func (i *Image) PullImage(ctx context.Context, ref reference.Named, metaHeaders
 		return err
 	}
 
-	client := PortLayerClient()
-	ImageCache().Update(client)
+	imageJSON, err := ioutil.ReadFile(imageFile)
```
dougm: Looks like instead of the temp file, you could use io.MultiWriter:
```go
var buf bytes.Buffer
cmd.Stdout = io.MultiWriter(outStream, &buf)
...
json.Unmarshal(buf.Bytes(), image)
```

jzt: @dougm So there's other stuff coming through stdout (progress, digest, etc). Would this provide any way of ignoring the stuff I don't want?

cmd/imagec/imagec.go
jzt: I've talked with @cgtexmex and I've decided that this will not support multiple tags per image until his changes go in. Once they're in, I can refactor this to take advantage of them.

lib/apiservers/engine/backends/image.go
fdawg4l: @jzt i think in the multiwriter case, (please correct me if I'm wrong), you'd have to deal with alignment of the stream yourself.  It's not impossible, but it's not trivial either.

fdawg4l:
While chatting with @caglar10ur, we have an alternative solution. For sure this implementation gets you to where you need to be so that's, by definition, good enough for now.  However, if you want to try something, we have a few options-

We could move imagec from its own binary to having the docker engine calling the imagec functions directly.  Then you can modify PULL to return the `Image` struct once it's done successfully.

The other option is to buffer STDERR (4k or something meaningfully small but large enough to fit your blob), wait for the 0 exit code in the docker daemon, then retrieve the marshaled `Image` struct from that, and then unmarshal it.  Otherwise on non-zero exit, dump the buffered STDERR back to the client.

jzt:
@fdawg4l @caglar10ur Of the two alternatives you propose, I'd definitely prefer the first. A third alternative that @hickeng and I discussed may suit you better if you are opposed to changing imagec: If we could ask the portlayer "What has changed since the last time I asked you what has changed?", we could get incremental updates from the portlayer after a pull without having to get everything the portlayer has and having to sift through it all. There's a similar notion in VC that inspired @hickeng to suggest this as a solution.

That being said, I have always longed for imagec to be able to provide info back to whatever persona is using it. I do think using a tempfile, and even stdout, is not ideal. Treating imagec as a library or an actual service would be a more elegant solution in my opinion (although perhaps way more work), but I still have not fully wrapped my head around the reasons for why it is a standalone binary. I imagine there are very good reasons for that design decision and my lack of understanding there is not reason enough to change it in such a major way. So, if keeping it a standalone binary is the desired outcome, we could investigate using the third solution so that everyone wins. :)

lib/apiservers/engine/backends/cache/image_cache.go
sflxn: No.

sflxn:
I think using stderr for actual data instead of errors is a bad idea.  I would normally be fine with stdout, but as you've said, it's being used for progress so it's out.  A file works but tightly couples you to having both the personality server and imagec on the same machine.

Getting deltas is fine.  The question is how does imagec know what your specific personality server have?  If you're sharing images and there are several personality server, what then?  The way to get around that is to do something similar to GIT.  Every update has an identifier.  "Hey, imagec, give me all updates since this sha".  Then you need to synchronize the data on the personality server side.

We're making it harder than it needs to be just to keep imagec a standalone binary.  We all see the path of least resistance, but we keep pushing the end of that string.

fdawg4l:
@jzt It's your call in the end how you'd like to proceed.  If you want to continue with a tmpfile, i'm fine with it.  If you want take on merging the imagec calls into the docker daemon, I think that's a great idea.

As for diffing between 2 calls, I have a solution but you're not going to like it :-).  I could hand back a handle every time you ask for the list of images.  To find out what changed, you'd reissue the `List` but using the handle as a reference point.  The rub is the handle is the totallity of images :-).  I'd just be diffing the list in the port layer.

lib/apiservers/engine/backends/cache/image_cache.go
jzt: I have confuse. I thought in our slack discussion we decided we needed to keep this?

sflxn: Oh, I'm sorry, I was looking at the highlighted code, the removal of the lock.

jzt:
@fdawg4l @sflxn I've closed this PR for now as the tempfile is simply an optimization at this point. I will open another PR in the near future that updates tests & docker errors, as well as improves locking in the cache. I will talk to @hickeng about when it would be appropriate to implement an alternative solution to the tempfile that we discussed previously. I'll probably open that PR after @sflxn's `docker inspect` changes have been merged, presumably after this CI armageddon dies down.

jzt:
@fdawg4l @sflxn @caglar10ur https://github.com/vmware/vic/pull/1238

Issue #917, opened 2016-05-31
docker volume ls


fdawg4l:

fdawg4l:
@matthewavery `VolumesList` in the volume cache works as expected, returning a list of `Volume`s.  Just create a list of ID from the returned `Volumes`, and that should be all you need to make the changes to the personality to support `docker volume ls`.

matthewavery:
Yes this one is pretty much wrapped up opening a PR on it now. was very simple to get working. Just lacks some unit tests and this ticket is good to go.

PR #1170, opened 2016-06-21
Add VolumeStorer API to the storage port layer

fdawg4l:
Includes a VolumeStorer interface definition to store/retrieve/remove
volumes.

Implements a cache implementation of the interface so as to avoid
unnecessary datastore operations.

This code isn't being called yet but is in support of #915 #916 and #917.

### DIFF
--- a/lib/portlayer/storage/volume_cache.go
+++ b/lib/portlayer/storage/volume_cache.go
@@ -1,0 +1,137 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package storage
+
+import (
+	"net/url"
+	"os"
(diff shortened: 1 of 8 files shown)


caglar10ur:
LGTM :sheep: :it:

lib/portlayer/storage/volume_cache.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (v *VolumeLookupCache) VolumesList(ctx context.Context) ([]*Volume, error) {
+	v.vlcLock.Lock()
+	defer v.vlcLock.Unlock()
+
+	// look in the cache, return the list
+	var i int
+	l := make([]*Volume, len(v.vlc))
+	for _, vol := range v.vlc {
+		// this is idiotic
```
dougm: Could just return a '[]Volume' instead?  In any case, can you use the range index (currently ignored '_'), rather than the 'var i' and 'i++' ?  Or even drop the index and use 'l = append(l, &e)'

dougm:
lgtm

seems like we could make VolumesList a little happier though

lib/portlayer/storage/volume_cache.go
fdawg4l: `v.vlc` is a map so the `_` is the key.  We can't use that as an index.  But i *can* use `append`.  I'll go ahead and do that.  Thanks!

fdawg4l:
@hickeng i think i addressed all of the comments you had in my branch if you want to take a look.  The only issue is that now we're stuck behind a lock for every call but i guess that's unavoidable.

fdawg4l:
:recycle:

fdawg4l:
:hankey:

caglar10ur:
LGTM :sheep: :it:

fdawg4l:
.

PR #1172, opened 2016-06-22
Add integration testing for PRs into drone yml file

mhagen-vmware:
Unfortunately, the only way to support this currently is to just duplicate the entire section.  When we upgrade to version 0.5 of drone we might be able to clean this up and make it a bit more elegant.


andrewtchin:
lgtm

matthewavery:
LGTM

mhagen-vmware:
.

PR #1177, opened 2016-06-22
promote context to a pkg level global

caglar10ur:
so that other port-layer pkgs can use.

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/scopes_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/scopes_handlers.go
@@ -48,18 +48,12 @@ func (handler *ScopesHandlersImpl) Configure(api *operations.PortLayerAPI, handl
 	api.ScopesBindContainerHandler = scopes.BindContainerHandlerFunc(handler.ScopesBindContainer)
 	api.ScopesUnbindContainerHandler = scopes.UnbindContainerHandlerFunc(handler.ScopesUnbindContainer)
 
-	netCtx, err := network.NewContext(
-		net.IPNet{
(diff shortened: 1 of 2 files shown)


caglar10ur:
Forgot to mention, this is originally @hmahmood's work, I just copied over his branch as we need it for the DNS work.

andrewtchin:
lgtm

hmahmood:
lgtm

caglar10ur:
♻️

lib/apiservers/portlayer/restapi/handlers/scopes_handlers.go
```diff
@@ -48,18 +48,12 @@ func (handler *ScopesHandlersImpl) Configure(api *operations.PortLayerAPI, handl
...
-		},
-		net.CIDRMask(16, 32),
-	)
+	err := network.Init()
 	if err != nil {
 		log.Fatalf("failed to create network context: %s", err)
 	}
 
-	handler.netCtx = netCtx
+	handler.netCtx = network.DefaultContext
```
hickeng: How about we make this portlayer.DefaultContext? lib/portlayer/portlayer.go if it's a default setup - I'm open to reasons not to.
To confirm, this is only used by portions that are _not_ triggered in response to a user request on the API, correct?

caglar10ur: I'm not sure about other usages but right now it holds all the network information that dns needs and since handler was creating it as a singleton it wasn't possible

caglar10ur:
♻️

caglar10ur:
♻️  (merging without waiting borked ci)

PR #1181, opened 2016-06-23
Use network backing info for debug network

### LINKED ISSUE #1180: Debug network not correct for vCenter
The debug network is not being configured reliably on vCenter - it's still got the moref name mangling work around that was put into place a long time back.
We still need the debug network until basic #747 and either #454 or #938 are compete to deal with dhcp address handling and access to the containers.

This was broken by #1127 and the move to extraconfig & moref based identifcation.

hickeng:
The debug network was using name mangling based on the moref which is utterly horrific and doesn't work reliably in vCenter. We cannot yet get rid of the debug network (see #1180) so this gets the backing info via the extraconfig mechanisms.

Fixes #1180

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .vagrant/
+.vscode/
 .cover/
 bin/
 binary/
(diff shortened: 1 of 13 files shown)


hickeng:
@hasan @andrewtchin this should address the odd vCenter config on the debug NIC for now.

andrewtchin:
lgtm

andrewtchin:
verified this works on vCenter <3

gigawhitlocks:
lgtm

hickeng:
lgtm
force update

hickeng:
lgtm force update

PR #1185, opened 2016-06-23
Photon 1.0

### LINKED ISSUE #1041: Update build to use Photon 1.0 GA
In order to go ahead with VIC GA, the system should be built against the Photon 1.0 GA release.

kjplatz:
Fixes #1041.
It turns out that when moving from Photon RC2 to Photon-1.0 GA, the naming convention of the network interfaces changed from eno* to eth*.  Therefore, the bootstrap script would spin forever waiting for /dev/eno1 to get an address, when /dev/eno1 does not exist.  This updates both  repo files and the bootstrap script to reflect the Photon-1.0 changes.


isos/bootstrap/bootstrap
```diff
@@ -59,21 +59,21 @@ if mount -t ext4 /dev/disk/by-label/containerfs ${MOUNTPOINT}; then
         cp /bin/tether ${MOUNTPOINT}/.tether/tether-bin
 
         # temporary hack for waiting the network
-        while ! ip addr show eno1 | grep " inet "; do sleep 1;done
+        while ! ip addr show eth1 | grep " inet "; do sleep 1;done
```
matthewavery: looks like there are several places where "eth1" is used. It might be best to add a variable that references that string so that if we have to change that value again it will be a one line quick fix. Even though this is a hack until we can add our own nics.

andrewtchin:
lgtm

isos/bootstrap/bootstrap
```diff
@@ -59,21 +59,21 @@ if mount -t ext4 /dev/disk/by-label/containerfs ${MOUNTPOINT}; then
         cp /bin/tether ${MOUNTPOINT}/.tether/tether-bin
 
         # temporary hack for waiting the network
-        while ! ip addr show eno1 | grep " inet "; do sleep 1;done
```
mlh78750: Consider scanning the system for the correct ethernet name and they putting that into a variable that we use for all cases below.

matthewavery: I believe that George will be adding a simple dhcp server to the tether to address this issue in the near future.

kjplatz: Per @hickeng this is a temporary solution until he can get a basic DHCP client built into the tether.

matthewavery:
lgtm

kjplatz:
This is a comment

caglar10ur:
LGTM :sheep: :it:

hickeng:
@kjplatz looks like you need a rebase and maybe some tweaks - not sure where/why your build is referencing integration-test2.sh

kjplatz:
This is a comment

PR #1187, opened 2016-06-23
Multiple image tag support

### LINKED ISSUE #976: imagec and our metadata design need to support multiple tags per image, including image digests

cgtexmex:
This change will provide support for multiple tags per docker image.  For example two pull requests were initiated which resulted in one image being downloaded -- a single Image ID with multiple tags.  

```
busybox             1.24.2              47bcc53f74dc        3 months ago        1.113 MB
busybox             latest              47bcc53f74dc        3 months ago        1.113 MB
```

This is considered a temporary fix until a more robust key / value store is implemented in the future.  This is only addressing multiple tags and not multiple digests, so it partially fixes #976

### DIFF
--- a/cmd/imagec/imagec.go
+++ b/cmd/imagec/imagec.go
@@ -123,6 +123,10 @@ const (
 
 	// DefaultTokenExpirationDuration specifies the default token expiration
 	DefaultTokenExpirationDuration = 60 * time.Second
+
+	// attribute update actions
+	Add = iota + 1
+	Remove
 )
 
 func init() {
@@ -226,7 +230,7 @@ func DestinationDirectory() string {
 }
 
 // ImagesToDownload creates a slice of ImageWithMeta for the images that needs to be downloaded
-func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, error) {
+func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, *ImageWithMeta, error) {
 	images := make([]*ImageWithMeta, len(manifest.FSLayers))
 
 	v1 := docker.V1Image{}
@@ -237,7 +241,7 @@ func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, e
 
 		// unmarshall V1Compatibility to get the image ID
 		if err := json.Unmarshal([]byte(history.V1Compatibility), &v1); err != nil {
-			return nil, fmt.Errorf("Failed to unmarshall image history: %s", err)
+			return nil, nil, fmt.Errorf("Failed to unmarshall image history: %s", err)
 		}
 
 		// if parent is empty set it to scratch
@@ -257,28 +261,32 @@ func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, e
 			layer:  layer,
 			diffID: "",
 		}
-		log.Debugf("Manifest image: %#v", images[i])
+		log.Debugf("ImagesToDownload Manifest image: %#v", images[i])
 	}
 
 	// return early if -standalone set
 	if options.standalone {
-		return images, nil
+		return images, nil, nil
 	}
 
 	// Create the image store just in case
 	err := CreateImageStore(storeName)
 	if err != nil {
-		return nil, fmt.Errorf("Failed to create image store: %s", err)
+		return nil, nil, fmt.Errorf("Failed to create image store: %s", err)
 	}
 
 	// Get the list of known images from the storage layer
 	existingImages, err := ListImages(storeName, images)
 	if err != nil {
-		return nil, fmt.Errorf("Failed to obtain list of images: %s", err)
+		return nil, nil, fmt.Errorf("Failed to obtain list of images: %s", err)
 	}
+
 	for i := range existingImages {
 		log.Debugf("Existing image: %#v", existingImages[i])
 	}
+
+	// grab the imageLayer for use in later evaluation of metadata update
+	imageLayer := images[0]
 
 	// iterate from parent to children
 	// so that we can delete from the slice
@@ -296,7 +304,117 @@ func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, e
 		}
 	}
 
-	return images, nil
+	return images, imageLayer, nil
+}
+
+// Will updated metadata based on the current image being requested
+//
+// Currently on tags are updated, but once imagec supports pulling by digest
+// then digest will need to be updated as well
+func updateImageMetadata(imageLayer *ImageWithMeta, manifest *Manifest) error {
+	// if standalone no need to continue
+	if options.standalone {
+		return nil
+	}
+	// updated Image slice
+	var updatedImages []*ImageWithMeta
+
+	// assuming calls to the portlayer are cheap, lets get all existing images
+	existingImages, err := ListImages(imageLayer.Store, nil)
+	if err != nil {
+		return fmt.Errorf("updateImageMetadata failed to obtain list of images: %s", err)
+	}
+
+	// iterate overall the images and remove the tag that was just downloaded
+	for id := range existingImages {
+		// utlizing index to avoid copy
+		existingImage := existingImages[id]
+
+		imageMeta := &metadata.ImageConfig{}
+		// tag / digest info only resides in the metadata so must unmarshall meta to determine
+		if err := json.Unmarshal([]byte(existingImage.Metadata[MetaDataKey]), imageMeta); err != nil {
+			return fmt.Errorf("updateImageMetadata failed to get existing metadata: Layer(%s) %s", id, err)
+		}
+
+		// if this isn't an image we can skip..
+		if imageMeta.ImageID == "" {
+			continue
+		}
+
+		// is this the same repo (busybox, etc)
+		if imageMeta.Name == options.image {
+
+			// default action to remove and update if necessary
+			action := Remove
+			if id == imageLayer.ID {
+				// add to the array if this layer is the requested layer
+				action = Add
+			}
+
+			// TODO: imagec doesn't support pulling by digest - add digest support once
+			// issue 1186 is implemented
+			if newTags, ok := arrayUpdate(manifest.Tag, imageMeta.Tags, action); ok {
+				imageMeta.Tags = newTags
+
+				// We need the parent id this is the layer ID of the parent disk
+				// Parent is returned from the PL as a full URL, so we only need the base
+				layerParent := path.Base(*existingImage.Parent)
+
+				// marshall back to dumb byte array
+				meta, err := json.Marshal(imageMeta)
+				if err != nil {
+					return fmt.Errorf("updateImageMetadata unable to marshal modified metadata: %s", err.Error())
+				}
+				updatedImage := &ImageWithMeta{Image: &models.Image{
+					ID:     id,
+					Parent: &layerParent,
+					Store:  imageLayer.Store,
+				},
+					meta: string(meta),
+					// create a dummy sum as this is required by the portlayer,
+					// but since we are only updating metadata it will be ignored
+					layer: FSLayer{BlobSum: "this-is-required"},
+				}
+				updatedImages = append(updatedImages, updatedImage)
+			}
+
+		}
+	}
+
+	log.Debugf("%d images require metadata updates", len(updatedImages))
+	for id := range updatedImages {
+		// utlizing index to avoid copy
+		update := updatedImages[id]
+		log.Debugf("updating ImageMetadata: layerID(%s)", update.ID)
+		//Send the Metadata to the PL to write to disk
+		err = WriteImage(update, nil)
+		if err != nil {
+			return fmt.Errorf("updateImageMetadata failed to writeMeta to the portLayer: %s", err.Error())
+		}
+
+	}
+	return nil
+}
+
+// Will conditionally update array based on presence of value
+func arrayUpdate(val string, list []string, action int) ([]string, bool) {
+
+	for index, item := range list {
+		if val == item {
+			if action == Remove {
+				// remove item requested
+				list = append(list[:index], list[index+1:]...)
+				return list, true
+			}
+			// item already in array
+			return list, false
+		}
+	}
+	if action == Add {
+		list = append(list, val)
+		return list, true
+	}
+	return list, false
 }
 
 // DownloadImageBlobs downloads the image blobs concurrently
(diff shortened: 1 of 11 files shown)


cmd/imagec/imagec.go
```diff
@@ -296,7 +304,107 @@ func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, e
...
+				action = Add
+			} else {
+				// remove from array if this layer isn't the requested layer
+				action = Remove
+			}
+			// TODO: imagec doesn't support pulling by digest - add digest support once
+			// issue 1186 is implemented
+			if newTags, ok := arrayUpdate(manifest.Tag, imageMeta.Tags, action); ok {
+				imageMeta.Tags = newTags
+				layerParent := (*existingImage.Parent)[strings.LastIndex(*existingImage.Parent, "/"):len(*existingImage.Parent)]
```
jzt: Minor: It's somewhat difficult to see what's happening here at first glance. Might be worth commenting or slightly refactoring.

cgtexmex: comment added - let me know if that provides the necessary clarity

lib/apiservers/engine/backends/image.go
```diff
@@ -280,3 +272,29 @@ func imageConfigToDockerImageInspect(imageConfig *metadata.ImageConfig, productN
 
 	return inspectData
 }
+
+func clientFriendlyTags(imageName string, tags []string) []string {
```
jzt: A comment about what this func is doing here would be useful for people who don't yet have the 2000 yard stare from digging through docker code. :)

lib/apiservers/engine/backends/image.go
```diff
@@ -280,3 +272,29 @@ func imageConfigToDockerImageInspect(imageConfig *metadata.ImageConfig, productN
...
+			clientTags[index] = fmt.Sprintf("%s:%s", imageName, tag)
+		}
+	} else {
+		clientTags = append(clientTags, fmt.Sprintf("%s:%s", "<none>", "<none>"))
+
+	}
+	return clientTags
+}
+
+func clientFriendlyDigests(imageName string, digests []string) []string {
```
jzt: Ditto here

lib/apiservers/engine/backends/image.go
cgtexmex: bastard - but agree

lib/apiservers/engine/backends/image.go
cgtexmex: ditto

cmd/imagec/storage.go
```diff
@@ -140,3 +138,39 @@ func WriteImage(image *ImageWithMeta, data io.ReadCloser) error {
...
+	*key = MetaDataKey
+	*blob = image.meta
+
+	r, err := client.Storage.WriteImage(storage.NewWriteImageParams().
+		WithImageID(image.ID).
+		WithParentID(*image.Parent).
+		WithStoreName(image.Store).
+		WithMetadatakey(key).
+		WithMetadataval(blob).
+		WithSum("required"))
```
jzt: I'm confused about why the blobsum for this layer would be "required". Can you shine some light?

jzt:
LGTM

cmd/imagec/storage.go
cgtexmex: I didn't create a new method on the API -- I'm using the existing WriteImage, so blob sum is required for that API call...but not necessary when only metadata is updated

cmd/imagec/imagec.go
```diff
@@ -123,6 +123,10 @@ const (
 
 	// DefaultTokenExpirationDuration specifies the default token expiration
 	DefaultTokenExpirationDuration = 60 * time.Second
+
+	// attribute update actions
+	Add = iota
```
caglar10ur: [minor] we can do ```iota + 1``` here if we want to differentiate Add from unset

cmd/imagec/imagec.go
```diff
@@ -226,7 +230,7 @@ func DestinationDirectory() string {
 }
 
 // ImagesToDownload creates a slice of ImageWithMeta for the images that needs to be downloaded
-func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, error) {
+func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, *ImageWithMeta, error) {
```
caglar10ur: confused, why we are returning an additional ImageWithMeta here? As far as I see it's alwasy the first item on the list so we can drop all the changed in this function and let caller to get first item.
```
eg; 
```
if err := updateImageMetadata(images[0], manifest); err != nil {
```
instead of
```
if err := updateImageMetadata(imageLayer, manifest); err != nil {
```

File: cmd/imagec/imagec.go [hunk 7]
```diff
@@ -296,7 +304,115 @@ func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, e
...
+func updateImageMetadata(imageLayer *ImageWithMeta, manifest *Manifest) error {
+	// if standalone no need to continue
+	if options.standalone {
+		return nil
+	}
+	// updated Image slice
+	var updatedImages []*ImageWithMeta
+
+	// assuming calls to the portlayer are cheap, lets get all existing images
+	existingImages, err := ListImages(imageLayer.Store, []*ImageWithMeta{})
```
caglar10ur: Can't we use nil here instead of []*ImageWithMeta{}?

cmd/imagec/imagec.go
```diff
@@ -296,7 +304,115 @@ func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, e
...
+	if err != nil {
+		return fmt.Errorf("updateImageMetadata failed to obtain list of images: %s", err)
+	}
+
+	// iterate overall the images and remove the tag that was just downloaded
+	for id, existingImage := range existingImages {
```
caglar10ur: existingImage will be a copy, instead we can do

```
for id := range existingImages {
    existingImage = existingImages[id]
...
```

cmd/imagec/imagec.go
```diff
@@ -296,7 +304,115 @@ func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, e
...
+		// if this isn't an image we can skip..
+		if imageMeta.ImageID == "" {
+			continue
+		}
+
+		// is this the same repo (busybox, etc)
+		if imageMeta.Name == options.image {
+			var action int
+			// determine action
+			if id == imageLayer.ID {
```
caglar10ur:
```
action = Remove
if id == imageLayer.ID {
    action = Add
}
```

cmd/imagec/imagec.go
```diff
@@ -296,7 +304,115 @@ func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, e
...
+					layer: FSLayer{BlobSum: "this-is-required"},
+				}
+				updatedImages = append(updatedImages, updatedImage)
+			}
+
+		}
+	}
+
+	log.Debugf("%d images require metadata updates", len(updatedImages))
+	for _, update := range updatedImages {
```
caglar10ur: Same as above, we can avoid copying by using indexed access

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -161,7 +161,9 @@ func (handler *StorageHandlersImpl) ListImages(params storage.ListImagesParams)
 
 // WriteImage writes an image to an image store
 func (handler *StorageHandlersImpl) WriteImage(params storage.WriteImageParams) middleware.Responder {
+
 	u, err := util.ImageStoreNameToURL(params.StoreName)
+
```
caglar10ur: Why?

caglar10ur:
couple nitpicks otherwise LGTM

cmd/imagec/imagec.go
```diff
@@ -262,24 +266,28 @@ func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, e
 
 	// return early if -standalone set
 	if options.standalone {
-		return images, nil
+		return images, nil, nil
 	}
 
 	// Create the image store just in case
 	err := CreateImageStore(storeName)
```
hickeng: We should restructure this, and anything similar, so that it's triggered as remediate/retry based on error returns when time allows. Not for this PR.

cmd/imagec/imagec.go
```diff
@@ -296,7 +304,115 @@ func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, e
...
+			// TODO: imagec doesn't support pulling by digest - add digest support once
+			// issue 1186 is implemented
+			if newTags, ok := arrayUpdate(manifest.Tag, imageMeta.Tags, action); ok {
+				imageMeta.Tags = newTags
+
+				// We need the parent id this is the layer ID of the parent disk
+				// Parent is returned from the PL as a full URL, so we need
+				// to slice the string (also a slice) get the ID -- since the parent string is a pointer
+				// we need to dereference
+				layerParent := (*existingImage.Parent)[strings.LastIndex(*existingImage.Parent, "/"):len(*existingImage.Parent)]
```
hickeng: why are we doing this instead of using existingImage.Parent directly? I presume there's a reason but I cannot determine what it is, from inspection or the comment.

cmd/imagec/imagec.go
```diff
@@ -296,7 +304,115 @@ func ImagesToDownload(manifest *Manifest, storeName string) ([]*ImageWithMeta, e
...
+				}
+				updatedImage := &ImageWithMeta{Image: &models.Image{
+					ID:     id,
+					Parent: &layerParent,
+					Store:  imageLayer.Store,
+				},
+					meta: string(meta),
+					// create a dummy sum as this is required by the portlayer,
+					// but since we are only updating metadata it will be ignored
+					layer: FSLayer{BlobSum: "this-is-required"},
```
hickeng: This seems like something that should be fixed on the PL side if there are updates that do not require a sum. Ok to leave as is if this is something that will cease to be necessary with move to k/v store.

cmd/imagec/storage.go
```diff
@@ -121,16 +121,14 @@ func WriteImage(image *ImageWithMeta, data io.ReadCloser) error {
...
-		storage.NewWriteImageParams().
-			WithImageID(image.ID).
-			WithParentID(*image.Parent).
-			WithStoreName(image.Store).
-			WithMetadatakey(key).
-			WithMetadataval(blob).
-			WithImageFile(data).
-			WithSum(image.layer.BlobSum),
-	)
+	r, err := client.Storage.WriteImage(storage.NewWriteImageParams().
```
hickeng: This is a very minor formatting change and that's the only thing in the file.
I'm not averse to formatting improvements going in along with other changes, but when it's the only change in the file it confuses the PR and commit diff.

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -130,17 +130,19 @@ func (c *ImageCache) Update(client *client.PortLayer) error {
...
-				c.cacheByName[ref.Name()] = imageConfig
+			for _, tag := range imageConfig.Tags {
+				ref, err = reference.WithTag(ref, tag)
+				if err != nil {
+					log.Errorf("Tried to create tagged reference from %s and tag %s: %s", imageConfig.Name, tag, err.Error())
+					continue
+				}
+
+				if tagged, ok := ref.(reference.NamedTagged); ok {
+					taggedName := tagged.Name() + ":" + tagged.Tag()
```
hickeng: While I hope the compiler optimizes this construct I prefer fmt.Sprintf("%s:%s", tagged.Name(), tagged.Tag()) when chaining

hickeng:
LGTM
Minor comments, but no functional issues.

cmd/imagec/imagec.go
cgtexmex: If all images (layers) already exist then images will be an empty slice. So the need for the additional ImageWithMeta -- in some cases it will be duplicated, but in others it won't.

cmd/imagec/imagec.go
cgtexmex: Will most likely cease with advent of k/v -- so I'll leave.

cmd/imagec/storage.go
cgtexmex: Reverted to original

lib/apiservers/engine/backends/cache/image_cache.go
cgtexmex: Agree -- I only added the loop...which I'll correct to avoid a copy (thanks @caglar10ur ) and I'll fix up the string concats

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
cgtexmex: to see who's awake!  ;-)  Seriously, it was some logging I added and removed, but failed to revert to the original file.  Corrected now...thanks!

cmd/imagec/imagec.go
cgtexmex: That is nil at that point...unless I'm misunderstanding the question..

caglar10ur: was asking whether calling this function like following would be possible

```
existingImages, err := ListImages(imageLayer.Store, nil)
```

instead of 

```
existingImages, err := ListImages(imageLayer.Store, []*ImageWithMeta{})
```

cgtexmex: yeah..we can..probably should as it could be confused currently...

caglar10ur:
LGTM :sheep: :it:

jzt:
♻️

PR #1191, opened 2016-06-23
Move parentMap file to VCH namespace

emlin:
To avoid concurrent write between multiple VCH, move the parentMap file to VCH directory.


dougm:
lgtm

emlin:
Updated GetImageStores method to restore parent map as well.
If reinstalled VCH, and occasionally the UUID is not changed, that happens in ESXi with same VM path, the GetImageStore will return a non nil value, and then CreateImageStore is not invoked.

hickeng:
LGTM

PR #1193, opened 2016-06-23
Check firewall rules for tether traffic, update port

### LINKED ISSUE #801: Change backchannel port
The backchannel port is currently 2377 - this requires us to open outgoing ports on the firewall of all ESXi hosts in a cluster. This is only truly viable until we consider dynamic add/remove of hosts. Ideally the appliance should not have credentials to deploy VIBs onto hosts and unconfiguring the firewall when they leave is just as problematic. 

The port should be changed to one that is open by default (vpxa backchannel for example) and a check added in the appliance to validate new hosts when they join the cluster.

andrewtchin:
Switch serialOverLANPort to 8080 which is open outbound by default. 
Provide firewall check in vic-machine and warn if the firewall is misconfigured. 
If the firewall would prevent VIC from running properly as configured, validation would fail and install would halt.

```
INFO[2016-06-23T12:41:59-07:00] ### Installing VCH ####                      
INFO[2016-06-23T12:41:59-07:00] Generating certificate/key pair - private key in ./chin-vic-38-key.pem 
INFO[2016-06-23T12:42:00-07:00] Validating supplied configuration            
INFO[2016-06-23T12:42:01-07:00] Firewall status: DISABLED on /Datacenter/host/Austin/patty.eng.vmware.com 
INFO[2016-06-23T12:42:01-07:00] Firewall configuration OK on hosts:          
INFO[2016-06-23T12:42:01-07:00]   /Datacenter/host/Austin/patty.eng.vmware.com 
```

Fixes #801

### DIFF
--- a/lib/install/validate/validator.go
+++ b/lib/install/validate/validator.go
@@ -27,6 +27,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 
 	"github.com/vmware/govmomi/find"
+	"github.com/vmware/govmomi/govc/host/esxcli"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/lib/install/data"
@@ -156,6 +157,7 @@ func (v *Validator) Validate(ctx context.Context, input *data.Data) (*metadata.V
 	v.compute(ctx, input, conf)
 	v.storage(ctx, input, conf)
 	v.network(ctx, input, conf)
+	v.firewall(ctx)
 
 	v.certificate(ctx, input, conf)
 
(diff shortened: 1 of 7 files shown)


lib/install/validate/validator.go
```diff
@@ -357,6 +359,86 @@ func (v *Validator) generateBridgeName(ctx, input *data.Data, conf *metadata.Vir
...
+			return
+		}
+
+		disabled := false
+		esxfw, err := esxcli.GetFirewallInfo(host)
+		if err == nil && !esxfw.Enabled {
+			disabled = true
+			log.Infof("Firewall status: DISABLED on %s", host.InventoryPath)
+		} else {
+			log.Infof("Firewall status: ENABLED on %s", host.InventoryPath)
```
emlin: Does this mean if we got error here, we'll assume the firewall is enabled?

Can we warn user for this error, even if we don't stop installation? Or maybe if forced, we continue, otherwise, just stop here?

andrewtchin: Right here it's just reporting the status - if the firewall is enabled it's still OK as long as the correct outbound port is open, which is open by default. We check if the port is open further down and then warn if it's not open.

lib/install/validate/validator.go
```diff
@@ -357,6 +359,86 @@ func (v *Validator) generateBridgeName(ctx, input *data.Data, conf *metadata.Vir
...
+			correct = append(correct, host.InventoryPath)
+		}
+	}
+
+	if len(correct) > 0 {
+		log.Info("Firewall configuration OK on hosts:")
+		for _, h := range correct {
+			log.Infof("  %s", h)
+		}
+	}
```
emlin: if no firewall is enabled, need to stop installation

andrewtchin: Fixed - on line 436 if we find the firewall misconfigured we NoteIssue which will cause validation to fail

lib/install/validate/validator.go
```diff
@@ -357,6 +359,93 @@ func (v *Validator) generateBridgeName(ctx, input *data.Data, conf *metadata.Vir
...
+	if len(misconfiguredEnabled) > 0 {
+		log.Error("Firewall configuration incorrect on hosts:")
+		for _, h := range misconfiguredEnabled {
+			log.Errorf("  %s", h)
+		}
+		msg := "Firewall must permit 8080/tcp outbound to use VIC"
+		log.Error(msg)
+		v.NoteIssue(errors.New(msg))
```
emlin: Would suggest to noteIssue while no any host is correctly configured, instead of to have all of them enable this policy.
In the future, user might complain that I have some hosts works with VIC, but you requires all of them match your rules.

emlin: It's fine to keep it at this moment, cause we still do not have filter in port-layer.

emlin:
LGTM

matthewavery:
LGTM

andrewtchin:
♻️

Issue #719, opened 2016-05-18
Milestone: portlet for containerVM


hickeng:
Adds a portlet describing the details of a containerVM to the VM detail view in the webclient

PR #1194, opened 2016-06-23
Initial vic-ui plugin with container portlet 

jak-atx:
This is the initial VIC-UI plugin with a container portlet. Currently we have container name available and placeholders for Image and Port mapping until we get the data from config. This is related to https://github.com/vmware/vic/issues/719


andrewtchin:
https://i.imgur.com/oI6yPja.jpg

dougm:
ui/vic-ui-bundles/register-plugin.jar - would prefer to add binary files to the git repo.  Is this a dependency we can pull in via maven or similar instead?

andrewtchin:
@dougm that's used to register the vsphere plugin - I don't think it's publicly available so I don't think we'd be able to pull it in from anywhere

dougm:
If register-plugin.jar isn't available to the public, are we even allowed to distribute it? cc @mdubya66 
If it isn't already available (via maven repo), could drop it on bintray for now as we do for the photon assets, rather than putting the binary .jar in our git repo?

jooskim:
Sorry I jumped in a little late, but @dougm, if we put the file somewhere online then users will have to somehow download it, when Karthik said many of our customers are not going to have internet connection. Having only that file hosted somewhere might overcomplicate stuff.

jooskim:
In the .gitignore file target/ and *.zip could possibly be prefixed with folder ui/

dougm:
@jooskim the .jar can be included in our release bundle.  We shouldn't need to include a .jar file in our source repo.  Again, we should be able to deal with this as we do with the photon assets (iso).

jooskim:
@dougm Ok. We recently discussed that we might require users to go into the /ui folder and run its platform-dependent script file - either install shell script file (VCSA) or install.bat (vCenter Windows), as with the current limitations it does not make much sense to integrate the plugin installation process into vic machine command. So assuming that the user downloaded the final tar ball file and ran the vic machine command, would we see be able to see the jar file where we want it to be? I'm curious how photon assets could be retrieved through which command.

dougm:
Not sure how that relates.  All I'm asking is to have the build system download the .jar, rather than add it to git.  Once the file has been downloaded, you can package it however you need to.

jooskim:
Oh ok. I just wasn't sure about it exactly. I'll check with @jak-atx and @andrewtchin for details.

andrewtchin:
i pointed jkim to where we would add this in `.drone.yml`

jak-atx:
@dougm we have removed the register-plugin.jar. It turns out there are other hoops involved in the CI build and we plan to address those in a separate PR

jooskim:
lgtm

dougm:
@jak-atx ok great.  Please do squash the commits for this PR, otherwise we'll still have the register-plugin.jar binary in our git repo (history).

jak-atx:
@dougm squashed

mdubya66:
@dougm Good point about the permission to distribute. @jak-atx and @jooskim can you confirm this jar is part of the https://developercenter.vmware.com/web/sdk/60/web-client?

jooskim:
@mdubya66 I'll check with the author of the tool and keep you in the loop

jooskim:
@dougm @mdubya66 @jak-atx Turns out this jar file should be used internally only. However, I have the source code from which I could write our own registration utility. Confirmed with Vladimir, the original author. And I'm already working on that.

emlin:
I would prefer to see a separate PR for plugin register. If that's the case, this one is LGTM.

jak-atx:
Yes plugin registration and anything related to CI build integration will be handled in separate CR

mdubya66:
I've asked @jooskim to hold off on pushing this until after TP3 is baked. Probably the week of July 11.  My rationale is that the UI plugins are now an Onsite Beta deliverable and that our build system is already flakey and we need more stability. This will give us 2 weeks of soak before the Onsite.

jak-atx:
Holding off on this PR until post TP3

Issue #857, opened 2016-05-31
docker info


hickeng:
This should return the docker specific data such as:
```
Containers: 14
 Running: 3
 Paused: 1
 Stopped: 10
Images: 52
Server Version: 1.9.0
```

It should also return some very basic capacity information: 
* memory available (resource pool limit or cluster capacity)
* cpu available (resource pool limit or cluster capacity)

Some customers expressly wanted information such as datastore hidden (pretty sure this was a service provider) so other data needs market research for inclusion.

It would be interesting to add the following elements as time allows:
* maximum cpu/memory for unreserved pools modified based on shares and current cluster utilization
* % available

#861 is open for exposing dynamic info about the execution environment but, as that issue notes, the static data in VCH config struct can be used as interim values.

corrieb:
@sflxn please close if this is completed

sflxn:
This is done but #1391 was opened for some of the remaining work.  There is a PR for 1391, but we held off on merging it due to CI issues at the time.

hickeng:
and #1390 for the correct values of cpu and memory

Issue #369, opened 2016-03-31
Docker API: Docker Container List


sflxn:
Implement the docker container list operation (docker images from the CLI) and integrate with the storage port layer.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#list-containers

Please note, the engine-api's rest server already packaging these json struct into golang structs.  The url above is just for reference on what the operation should do.

jzt:
This issue is currently blocked until we get something in place that stores container metadata. See #498, #457.

PR #1195, opened 2016-06-23
Swagger spec for container info and list

### LINKED ISSUE #1122: Port Layer: Create swagger definitions for getting container metadata and list of containers
We need some portlayer definitions that allows us to get container metadata (for docker inspect container) and get a list of containers (for docker ps).

related to #798 and #388

sflxn:
**WIP PR to allow collaboration between docker ps and docker inspect container**

Add in swagger spec to retrieve information about containers and get
a list of all containers in the system.  This supports #857 & #369.

Resolves #1122.


sflxn:
@cgtexmex @corrieb

caglar10ur:
LGTM

corrieb:
Seems like a good place to start. With the caveat that the REST interface doesn't have to reflect the golang interface exactly, it would be good (eventually) to consider having the separation between container status and container config reflected in different API calls. Apart from reflecting different kinds of state (static config data vs dynamic runtime state data), the two calls will have different costs on the implementation side. The config call will only ever hit port layer caches, whereas the status call will have to hit tether(s) and potentially even the vmomi gateway. I hope to avoid the latter though with the event mechanism

sflxn:
lgtm.  Haven't changed anything other then rebase.  Shouldn't require all new status.

PR #1196, opened 2016-06-24
Volumes on vsphere

### LINKED ISSUE #915: docker volume create

fdawg4l:
Github is dumb and won't let me create 2 PRs, one on top of another, without creating 2 separate branches and including the diff from the first in the second.  Go figure.  #imissgerrit #gerrit4ever

Anyway, this complete `the volume create` and `volume ls` port layer work on vsphere.

Fixes #915 #917

### DIFF
--- a/lib/portlayer/storage/volume.go
+++ b/lib/portlayer/storage/volume.go
@@ -25,6 +25,26 @@ import (
 	"golang.org/x/net/context"
 )
 
+type Disk interface {
+	MountPath() (string, error)
+	DiskPath() string
+}
+
+// VolumeStorer is an interface to create, remove, enumerate, and get Volumes.
+type VolumeStorer interface {
+	// Creates a volume on the given volume store, of the given size, with the given metadata.
+	VolumeCreate(ctx context.Context, ID string, store *url.URL, capacityMB uint64, info map[string][]byte) (*Volume, error)
+
+	// Get an existing volume via it's ID.
+	VolumeGet(ctx context.Context, ID string) (*Volume, error)
+
+	// Destroys a volume
+	VolumeDestroy(ctx context.Context, ID string) error
+
+	// Lists all volumes
+	VolumesList(ctx context.Context) ([]*Volume, error)
+}
+
 // Volume is the handle to identify a volume on the backing store.  The URI
 // namespace used to identify the Volume in the storage layer has the following
 // path scheme:
@@ -43,21 +63,31 @@ type Volume struct {
 
 	// Namespace in the storage layer to look up this volume.
 	SelfLink *url.URL
+
+	// Backing device
+	Device Disk
 }
 
-// VolumeStorer is an interface to create, remove, enumerate, and get Volumes.
-type VolumeStorer interface {
-	// Creates a volume on the given volume store, of the given size, with the given metadata.
-	VolumeCreate(ctx context.Context, ID string, store *url.URL, capacityMB uint64, info map[string][]byte) (*Volume, error)
+// NewVolume creates a Volume
+func NewVolume(store *url.URL, ID string, device Disk) (*Volume, error) {
+	storeName, err := util.VolumeStoreName(store)
+	if err != nil {
+		return nil, err
+	}
 
-	// Get an existing volume via it's ID.
-	VolumeGet(ctx context.Context, ID string) (*Volume, error)
+	selflink, err := util.VolumeURL(storeName, ID)
+	if err != nil {
+		return nil, err
+	}
 
-	// Destroys a volume
-	VolumeDestroy(ctx context.Context, ID string) error
+	vol := &Volume{
+		ID:       ID,
+		Store:    store,
+		SelfLink: selflink,
+		Device:   device,
+	}
 
-	// Lists all volumes
-	VolumesList(ctx context.Context) ([]*Volume, error)
+	return vol, nil
 }
 
 func (v *Volume) Parse(u *url.URL) error {
--- a/lib/portlayer/storage/volume.go
+++ b/lib/portlayer/storage/volume.go
@@ -85,17 +85,24 @@ func NewVolume(store *url.URL, ID string, device Disk) (*Volume, error) {
 	}
 
 	// Set the label to the md5 of the ID
-	h := md5.New()
 
 	vol := &Volume{
 		ID:       ID,
-		Label:    string(h.Sum([]byte(ID))),
+		Label:    label(ID),
 		Store:    store,
 		SelfLink: selflink,
 		Device:   device,
 	}
 
 	return vol, nil
+}
+
+// given an ID, compute the volume's label
+func label(ID string) string {
+	h := md5.New()
+
+	// e2label's manpage says the label size is 16 chars
+	return string(h.Sum([]byte(ID)))[:16]
 }
 
 func (v *Volume) Parse(u *url.URL) error {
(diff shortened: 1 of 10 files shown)


fdawg4l:
cc @matthewavery

fdawg4l:
Just pushed the 1st pr.   The diff should be clean now.

caglar10ur:
LGTM :sheep: :it:

lib/portlayer/storage/vsphere/volume.go
```diff
@@ -0,0 +1,217 @@
...
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/lib/portlayer/storage"
+	"github.com/vmware/vic/lib/portlayer/util"
+	"github.com/vmware/vic/pkg/vsphere/disk"
+	"github.com/vmware/vic/pkg/vsphere/session"
+
+	"golang.org/x/net/context"
+)
+
+const volumesDir = "volumes"
```
hickeng: not for this PR but I'm thinking of having a `lib/config/const.go` and moving the core VCH config files into that dir from metadata. This is because we're starting too need access to this sort of constant in vic-machine and vicadmin as well (for install and diagnostics respectively) and including the portlayer package is not a nice way to do it.

lib/portlayer/storage/vsphere/volume.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if _, ok := v.ds[*u]; ok {
+		return nil, fmt.Errorf("volumestore (%s) already added", u.String())
+	}
+
+	// Root our datastore by the directory structured above.
+	p := path.Join(parentDir, storageParentDir, volumesDir)
```
matthewavery: I do not see where "storageParentDir" is defined. should this be parentDir?

fdawg4l: Modified this piece of the change since we're getting vic-machine support sooner than expected.

fdawg4l:
Fixes #1213

lib/portlayer/storage/volume.go
```diff
@@ -43,21 +67,35 @@ type Volume struct {
...
-	VolumeGet(ctx context.Context, ID string) (*Volume, error)
+	selflink, err := util.VolumeURL(storeName, ID)
+	if err != nil {
+		return nil, err
+	}
 
-	// Destroys a volume
-	VolumeDestroy(ctx context.Context, ID string) error
+	// Set the label to the md5 of the ID
+	h := md5.New()
```
hickeng: If I remember correctly disk labels are capped at 15 characters (could be wrong) and md5sum is 16. Worth noting, but might not be worth doing anything about it at this point in time.

lib/portlayer/storage/vsphere/datastore.go
```diff
@@ -232,35 +255,23 @@ func (d *datastore) mkRootDir(ctx context.Context, rootdir string) error {
...
-	// Handle regular local datastore
-	// check if it already exists
-	_, err := d.Ls(ctx, d.ds.Path(rootdir))
-	if err != nil && !types.IsFileNotFound(err) {
-		return err
-	}
-
-	d.rootdir = rootdir
-	d.rooturl = d.ds.Path(rootdir)
+		d.RootURL = d.ds.Path(path.Base(uuid))
```
hickeng: Am I correct in thinking this is not actually a URL given the rootDir implementation? Otherwise I'd expect it to be used directly as `d.RootURL.Path`:
```
func (d *Datastore) rootDir() string {
 	return strings.SplitN(d.RootURL, " ", 2)[1]
 }
```

lib/portlayer/storage/vsphere/store.go
```diff
@@ -99,7 +99,7 @@ func (v *ImageStore) imageDirPath(storeName, imageName string) string {
 // Returns the path to the vmdk itself
 func (v *ImageStore) imageDiskPath(storeName, imageName string) string {
 	// XXX this could be hidden in a helper.  We shouldn't use rooturl outside the datastore struct
-	return path.Join(v.ds.rooturl, v.imageDirPath(storeName, imageName), imageName+".vmdk")
+	return path.Join(v.ds.RootURL, v.imageDirPath(storeName, imageName), imageName+".vmdk")
```
hickeng: Is the comment above this still valid given we've exported RootURL.

lib/portlayer/storage/vsphere/volume.go
```diff
@@ -0,0 +1,212 @@
...
+
+// AddStore adds a volumestore by uri.
+//
+// ds is the Datastore (+ path) volumes will be created under.
+//       The resulting path will be parentDir/VIC/volumes.
+// storeName is the name used to refer to the datastore + path (ds above).
+//
+// returns the URL used to refer to the volume store
+func (v *VolumeStore) AddStore(ctx context.Context, ds *Datastore, storeName string) (*url.URL, error) {
+	u, err := util.VolumeStoreNameToURL(storeName)
```
hickeng: What was the reasoning for the "http://<appliance_name_name>" base in the ServiceURL creation step? Does this URL get persisted anywhere or is it purely for internal reference?

lib/portlayer/storage/vsphere/volume.go
```diff
@@ -0,0 +1,212 @@
...
+
+	// Create the disk
+	vmdisk, err := v.dm.CreateAndAttach(ctx, volDiskDsURL, "", int64(capacityMB), os.O_RDWR)
+	if err != nil {
+		return nil, err
+	}
+	defer v.dm.Detach(ctx, vmdisk)
+
+	// Make the filesystem and set its label to the volume ID
+	if err = vmdisk.Mkfs(ID); err != nil {
```
hickeng: Shouldn't this be set to Label instead so we can identify the disk in the containerVMs?

lib/portlayer/storage/vsphere/datastore.go
fdawg4l: It is not a real URL.  It's the format used by the vsphere api to denote a datastore in the form of `[datastoreName] /path/`.  I think I read somewhere in the SDK this was referred to as a URL but I can change the name if you have a more descriptive identifier for the format.  I use the convention of variable name including URL and keeping the type as a string.

lib/portlayer/storage/vsphere/store.go
fdawg4l: LOL!  no, its not.  Good catch.  Will fix.

lib/portlayer/storage/vsphere/volume.go
fdawg4l: Yeah, good catch.  Will fix.

lib/portlayer/storage/vsphere/volume.go
fdawg4l: It's just internal reference and not persisted anywhere.  We do validate these URLs (here and everywhere else we need to parse it via helpers), but they're purely for the internal representation of in-memory objects.

lib/portlayer/storage/volume.go
fdawg4l: Close.  Looks like it's 16 characters.

> If the optional argument new-label is present, then e2label will set the filesystem label to be new-label.  Ext2 filesystem labels can be at most 16 characters long; if new-label is longer than 16 characters, e2label will truncate it and print a warning message.

I'll update this with a comment referring to the `e2label` manpage.

jzt:
LGTM

matthewavery:
LGTM

fdawg4l:
.

PR #1209, opened 2016-06-24
[dns] Introduce network scopes

caglar10ur:
- we now have a seperate vic-dns binary to create standalone DNS service
- context now inserts/removed truncated id and common/name into
the map of containers (using SCOPE:NAME as a key)
- dns part of the PL now iterates over scopes to map requester IP adress
and returns the SCOPE:NAME key as a response
- tether now generates /etc/resolv.conf for containers

### DIFF
--- a/lib/dns/dns.go
+++ b/lib/dns/dns.go
@@ -25,15 +25,19 @@ import (
 	"time"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/vmware/vic/lib/portlayer/exec"
+	"github.com/vmware/vic/lib/portlayer/network"
 	"github.com/vmware/vic/pkg/trace"
 
 	mdns "github.com/miekg/dns"
 )
 
 const (
+	DefaultIP        = "127.0.0.1"
 	DefaultPort      = 53
 	DefaultTTL       = 600 * time.Second
 	DefaultCacheSize = 1024
+	DefaultTimeout   = 4 * time.Second
 )
 
 var (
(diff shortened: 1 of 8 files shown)


lib/dns/dns.go
```diff
@@ -284,88 +298,85 @@ Redo:
...
+	defer trace.End(trace.Begin(r.String()))
 
-	return nil
-}
+	question := r.Question[0]
 
-// ServeDNS implements the handler interface
-func (s *Server) ServeDNS(w mdns.ResponseWriter, r *mdns.Msg) {
-	defer trace.End(trace.Begin(r.String()))
+	ctx := network.DefaultContext
```
hmahmood: Maybe pass in the context?

lib/portlayer/network/context.go
```diff
@@ -480,7 +497,24 @@ func (c *Context) UnbindContainer(h *exec.Handle) error {
...
+	var key string
+	// network scoped
+	for _, e := range con.Endpoints() {
+		sname := e.Scope().Name()
+
+		key = fmt.Sprintf("%s:%s", sname, tid)
+		delete(c.containers, exec.ParseID(key))
+
+		key = fmt.Sprintf("%s:%s", sname, cname)
+		delete(c.containers, exec.ParseID(key))
```
hmahmood: Do we also remove from the dns cache?

caglar10ur: Nope, but we don't cache internal names

hmahmood:
lgtm

dougm:
lgtm

PR #1211, opened 2016-06-24
vic-machine delete - delete images, volumes

emlin:
This PR finished vic-machine delete functions. 
Notes:
  Always force delete images.
  Leave volumes there if --force is not specified in command line, so it's possible to keep user data even if VCH is deleted.
  Dettach network devices from appliance VM, so it's possible to remove created bridge network before delete appliance VM.

  Delete appliance VM in the last, in case anything wrong happens, it's still possible to run vic-machine delete once again.

  Do not support --force to reinstall VCH, cause currently vic-machine delete function well. Force reinstall might remove user existing work by mistake. Clearly delete will be preferred. So removed the test case for force reinstall.

This PR also fixed one bug in previous delete function, which will fail if env has multiple datastore and hosts.


dougm:
lgtm

hickeng:
@cormachogan @emlin The overloaded nature of --force is awkward seeing it listed in the description. Can we have a separate flag to determine whether to delete volumes, e.g.
```
delete                                                                  # preserves volumes
delete --all                                                           # deletes everything
delete --preserve-volumes --preserve-images    # only delete the appliance/containers/networks
```
didn't put much thought into the example above - it's just to illustrate.

lib/install/management/dispatcher.go
```diff
@@ -181,8 +196,8 @@ func (d *Dispatcher) Dispatch(conf *metadata.VirtualContainerHostConfigSpec, set
 		return err
 	}
 
-	if err = d.removeApplianceIfForced(conf); err != nil {
-		return errors.Errorf("%s", err)
+	if err = d.checkExistence(conf); err != nil {
```
hickeng: We should change the name of Dispatch to Create or Deploy - doesn't need to be done in this PR, but at the moment it's single purpose with a general purpose name which is confusing

lib/install/management/appliance.go
```diff
@@ -73,26 +73,15 @@ func (d *Dispatcher) removeApplianceIfForced(conf *metadata.VirtualContainerHost
...
-	return d.deleteDatastoreFiles(conf, folder)
+	err = errors.Errorf("Appliance %s exists, to install with same name, please delete it first.", conf.Name)
+	return err
 }
 
-func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine) (string, error) {
+func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) (string, error) {
 	var err error
 	power, err := vm.PowerState(d.ctx)
 	if err != nil {
```
hickeng: While it's likely to fail if we cannot get the power state, we should continue our attempt if force == true. Assume that the VM is powered on and continue.

lib/install/management/delete.go
```diff
@@ -66,7 +100,7 @@ func (d *Dispatcher) DeleteVCHInstances(conf *metadata.VirtualContainerHostConfi
 		if name == conf.Name {
 			continue
 		}
-		if _, err = d.deleteVM(child); err != nil {
+		if _, err = d.deleteVM(child, d.force); err != nil {
```
hickeng: This is ignoring the extra files that will be in those containerVM directories - the containerVM folder will not be deleted.
Please move the folder deletion for VMs into the deleteVM function - I don't think there's any case at this time where we don't want to delete the folder as well - can you identify any?

lib/install/management/delete.go
```diff
@@ -76,10 +110,82 @@ func (d *Dispatcher) DeleteVCHInstances(conf *metadata.VirtualContainerHostConfi
...
+		return err
+	}
+	if info.State != types.TaskInfoStateSuccess {
+		log.Errorf("Removing network devices reported: %s", info.Error.LocalizedMessage)
+		return err
+	}
 	return nil
 }
+
+func (d *Dispatcher) deleteNetworkSpec(vmm *vm.VirtualMachine, conf *metadata.VirtualContainerHostConfigSpec) (*types.VirtualMachineConfigSpec, error) {
```
hickeng: There is a govmomi operation VirtualMachine.RemoveDevice and you can use VirtualDeviceList to get the network cards as it doesn't look like you're using the spec for anything other than device deletion

lib/install/management/delete.go
```diff
@@ -15,41 +15,75 @@
...
+		log.Infof("Specify --force to force delete")
 		return err
 	}
-	return d.removeNetwork(conf)
+
+	if err = d.DeleteDataStores(vmm, conf); err != nil {
+		errs = append(errs, err.Error())
+	}
+
+	if err = d.deleteNetworkDevices(vmm, conf); err != nil {
```
hickeng: Is it actually necessary to delete the network devices before deleting the network? I may be recalling incorrectly but I thought it was possible to do so and associated NICs would simply become disconnected.

lib/install/management/store_files.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/vmware/vic/pkg/vsphere/vm"
+
+	"golang.org/x/net/context"
+)
+
+const (
+	volumeRoot = "volumes"
+)
+
+func (d *Dispatcher) DeleteDataStores(vchVM *vm.VirtualMachine, conf *metadata.VirtualContainerHostConfigSpec) error {
```
hickeng: This is a terrifying function name - we should **never** be deleting datastores - let's call this DeleteStores instead as it's less likely to be confused... the imageStore and volumeStore do not have _data_ in them so it's not an obvious mapping.

lib/install/management/store_files.go
```diff
@@ -0,0 +1,162 @@
...
+	var emptyVolumes bool
+	log.Infof("Removing images")
+	if emptyImages, err = d.deleteImages(ds, path); err != nil {
+		errs = append(errs, err.Error())
+	}
+	log.Infof("Removing volumes")
+	if emptyVolumes, err = d.deleteVolumes(ds, path); err != nil {
+		errs = append(errs, err.Error())
+	} else if !emptyVolumes {
+		log.Infof("Volumes directory %s is not empty, to delete with --force specified", path)
```
hickeng: nice

lib/install/management/store_files.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	if emptyImages && emptyVolumes {
+		// if not empty, don't try to delete parent directory here
+		log.Infof("Removing root directory")
```
hickeng: `Removing stores directory` - let's not alarm users... the person running vic-machine and reading this may have a very different interpretation of that message that results in a reflective Ctrl-C to kill whatever says it.

lib/install/management/store_files.go
```diff
@@ -0,0 +1,162 @@
...
+	return d.deleteDatastoreFiles(ds, p, true)
+}
+
+func (d *Dispatcher) deleteVolumes(ds *object.Datastore, root string) (bool, error) {
+	p := path.Join(root, volumeRoot)
+	// if not forced delete, leave volumes there. Cause user data can be persisted in volumes
+	return d.deleteDatastoreFiles(ds, p, d.force)
+}
+
+func (d *Dispatcher) deleteDatastoreFiles(ds *object.Datastore, path string, force bool) (bool, error) {
```
hickeng: This and deleteVMFSFiles seem like methods that should be pushed into govmomi
@dougm comments?

hickeng:
In addition to the inline comments, can you also add entry/exit trace - it'll become rather verbose, but for TP3 we really do need to be able to grab good diagnostic logs.

lib/install/management/dispatcher.go
emlin: As we plan to separate the modules to different services in the future, I assume dispatcher will be the service name, and it provides different API for create, delete, etc.
But yeah, single purpose function will be more clear. Will do it in next PR.

lib/install/management/delete.go
emlin: My concern is that container VM will be in different datastore with appliance VM. Right now, I could assume they are all in the image datastore, but in the future, this still need to be changed.

lib/install/management/delete.go
emlin: didn't know that. let me try it.

lib/install/management/store_files.go
emlin: too many delete there, and we have to have different name for each method.

lib/install/management/store_files.go
emlin: I think govmomi should have those methods. 
And have talked with @fdawg4l about those storage methods. He will have some common methods in pkg/vsphere. While that's ready, we have refactor vic-machine to use them.

emlin:
> delete                                                                  # preserves volumes
> delete --all                                                           # deletes everything
> delete --preserve-volumes --preserve-images    # only delete the appliance/containers/networks
> 
> didn't put much thought into the example above - it's just to illustrate.

Not quite sure about it. Wouldn't that becomes too many parameters?
If we do need these many different options, I think we might need to go with interactive installation solution. 
And provide silent install option for users who already know every options.

emlin:
> In addition to the inline comments, can you also add entry/exit trace - it'll become rather verbose, but for TP3 we really do need to be able to grab good diagnostic logs.
So that will be log out in debug level?

I agree to do it, but is it OK to make the change in next PR, together with the previous comment for command option change? 
This one is good to have, but is not related to the function change.

lib/install/management/delete.go
emlin: Looks like if VM is powered on, delete network will have problem, but if VM is powered off, it is OK. 
Right now, I think delete device is more safe then delete network directly.

lib/install/management/delete.go
```diff
@@ -19,43 +19,80 @@ import (
...
+		// stop here, leave vch appliance there for next time delete
+		return errors.New(strings.Join(errs, "\n"))
+	}
+	err = d.deleteVM(vmm, true)
+	if err != nil {
+		log.Debugf("Error deleting appliance VM %s", err)
 		return err
 	}
-	return d.removeNetwork(conf)
+	if err = d.destroyResourcePoolIfEmpty(conf); err != nil {
```
hickeng: are you aware of the object.ResourcePool.DestroyChildren method in govmomi? I don't know how aggressive it is, but it may be worth trying here.

lib/install/management/delete.go
```diff
@@ -19,43 +19,80 @@ import (
...
 	var children []*vm.VirtualMachine
 
-	if vmm, err = d.findApplianceByID(conf); err != nil {
-		return err
-	}
-	if vmm == nil {
-		return nil
-	}
 	rpRef := conf.ComputeResources[len(conf.ComputeResources)-1]
 	rp := compute.NewResourcePool(d.ctx, d.session, rpRef)
```
hickeng: @emlin NewResourcePool is returning a broken resource pool from the govmomi perspective - it doesn't have the InventoryPath set.
Please update this to use the find.ObjectReference method that was added to retrieve valid govmomi objects from moref values.

lib/install/management/delete.go
emlin: DestroyChildren will delete all children resource pools, do we need that function?
We don't create child resource pool for VCH now.

lib/install/management/delete.go
emlin: I know that object does not provide govmomi functions well, but actually, I only need their moRef later on.
But anyway, find.ObjectReference could verify if the mob-id exists in vsphere, so changed to that.

lib/install/management/delete.go
hickeng: I don't think we need that function yet - we will if/when we add a cgroup-parent translation as that's essentially identical to nested resource pools from the resource management perspective.
Just wanted to make sure you were aware of the function as it wasn't one I knew.

hickeng:
LGTM

@emlin yup - we can and should leave some of the enhancements for another PR. I add notes when issues occur to me, but I normally remember to note whether it should be in a later PR or not.

Issue #709, opened 2016-05-18
vic-machine: debug flag


emlin:
This should configure all components to run in debug mode:
1. govmomi soap recordings
2. enable pprof server in all components
3. vicadmin collects pprof data
4. log level set to debug
5. port layer listens on client IP as well as of localhost for direct connect

This implies that we have debug modes for all of the components.

emlin:
Closed for the debug portion is finished. Portlayer is still not available at this moment, but that is not a requirement yet.

PR #1217, opened 2016-06-26
Fixes container log collection in vicadmin

### LINKED ISSUE #1150: container-logs.tar.gz does not contain container logs
#1075 was opened to collect container logs - turns out that's because vicadmin is not putting the container logs into the container-logs.tar.gz bundle any more.
This is a significant reduction in diagnostic capability and should be addressed asap.

@dougm Assigned to you as you're the original author - not intending this to take much of your time so if it's more than an hour or two let me know and we'll assign it to someone else to dig into.

hickeng:
Fixes #1150 
Towards #709

vicadmin was not collecting the logs for containers - the assumptions
carried over from bonneville were not valid (and would have been broken on
vsan anyway).

This code also adds some restructuring in anticipation of adding pprof
collection to the other appliance components and, ideally, the containerVMs


cmd/vicadmin/vicadm.go
```diff
@@ -300,56 +325,76 @@ type datastoreReader struct {
...
+		path, err := child.DSPath(ctx)
+		if err != nil {
+			log.Errorf("Unable to get datastore path for child VM %s: %s", child.Reference(), err)
+			// we need to get as many logs as possible
+			continue
+		}
 
-	if info.Error != nil {
-		return nil, errors.New(info.Error.LocalizedMessage)
+		log.Debugf("Adding VM for log collection: %s", path.String())
```
hickeng: If anyone can tell me why:
a. the log package qualifier isn't visible here but is in the full file view
b. why this line doesn't print out, but does compile and the line below executes...
I will owe them thanks. Same for line 397

dougm: Not sure what you mean by a ?  Regarding b, I assume you're seeing other log.Debug's ?  Do you see the message if you change to another log level?

lib/metadata/virtual_container_host.go
```diff
@@ -24,6 +24,19 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 )
 
+// PatternToken is a set of tokens that can be placed into string constants
```
dougm: Not seeing this referenced anywhere, other than the const below?

dougm:
lgtm

Just not seeing the PatternToken type being used anywhere.

cmd/vicadmin/vicadm.go
hickeng: By a, I mean in the diff above I see:
`+		Debugf("Adding VM for log collection: %s", path.String())`

but when I paste from the diff:
`+		log.Debugf("Adding VM for log collection: %s", path.String())`

I have tried deleting and recreating the line in case of non-printable characters, but it does compile and that had no effect.

lib/metadata/virtual_container_host.go
hickeng: https://github.com/vmware/vic/pull/1217/files#diff-227bcd7fc9072cf62ac2d3ed69ba4071R391

cmd/vicadmin/vicadm.go
dougm: Oh strange, I see 'log.Debugf' the diff above.

matthewavery:
lgtm

hickeng:
poke

hickeng:
poke

Issue #828, opened 2016-05-30
Document vic-machine delete


stuclem:
Per https://github.com/vmware/vic/issues/702, https://github.com/vmware/vic/issues/652, https://github.com/vmware/vic/issues/897, and https://github.com/vmware/vic/issues/919.

stuclem:
See https://github.com/vmware/vic/blob/master/doc/design/vic once PR https://github.com/vmware/vic/pull/921/ is merged.

stuclem:
Issue https://github.com/vmware/vic/issues/897 to create the vic-machine delete processor is in progress, so moving this one to To Do.

stuclem:
Pending the implementation of vic-machine delete, I added a topic on how to manually remove a VCH:
https://github.com/stuclem/vic/blob/master/doc/user_doc/vic_admin/remove_vch.md

emlin:
@stuclem Merged vic-machine delete code yesterday. There will not have functional change, but still might update flags based on usability requirement.

stuclem:
Thanks @emlin.

See also PR https://github.com/vmware/vic/pull/1253, which contains the integration test for vic-machine delete.

stuclem:
See also PR https://github.com/vmware/vic/pull/1302, which fixes issue https://github.com/vmware/vic/issues/1296.

stuclem:
PR https://github.com/vmware/vic/pull/1290 grabs VCH logs before deleting the appliance. This might just be for internal testing, though. TBC with @mhagen-vmware.

stuclem:
Added a topic on vic-machine delete in admin.

stuclem:
Still need to write up full descriptions of the options for vic-machine delete.

stuclem:
Added new topic that describes the vic-machine inspect options:

https://github.com/stuclem/vic/blob/master/doc/user_doc/vic_admin/remove_vch.md
https://github.com/stuclem/vic/blob/master/doc/user_doc/vic_admin/delete_vch_options.md

@emlin are these two topics OK? Thanks!

stuclem:
Actually, I missed two options: force and timeout. @emlin please wait before reviewing.

stuclem:
Have now added force and timeout to https://github.com/stuclem/vic/blob/master/doc/user_doc/vic_admin/delete_vch_options.md

emlin:
@stuclem this looks good.

stuclem:
Thanks!

PR #1227, opened 2016-06-27
Add short name for common vic-machine arguments

emlin:
Also add trace to vic-machine functions, and update usage document for vic-machine delete command.
Reference #828


andrewtchin:
lgtm

caglar10ur:
LGTM (it's minor but I think we should pass meaningful strings to trace package when possible instead of just saying "")

emlin:
ah, make sense. Didn't know that.
It will be useful to have parameter printed out.

emlin:
refresh

PR #1232, opened 2016-06-27
Print issue number in error messages that have open issues

sgairo:
This also cleans up and gets rid of the TIME variable


andrewtchin:
lgtm

caglar10ur:
Looks like your branch is not up2date and shows unrelated commits, can you fix that?

PR #1238, opened 2016-06-27
(WIP) Make imagec work as a library

### LINKED ISSUE #1154: Improve image cache availability and locking

jzt:
This change moves most of the imagec functionality into `lib/imagec`. What it leaves behind in `cmd/imagec/` imports `github.com/vmware/vic/lib/imagec` and makes use of it that way for the standalone binary. There are some extra changes to facilitate that as well.

The purpose of these changes is to allow for imagec to cleanly provide feedback to the docker persona. This improves the caching mechanism for images by allowing the full cache update to happen only once at persona startup, and by allowing the persona to cache image metadata about pulled images without having to make extra calls to the portlayer. There are also locking/availability improvements around the cache as well.

Feedback is certainly welcome. Criticism is expected. :)

Fixes #1154, #1269


jzt:
Also, apologized for all the red. I moved a lot of stuff around.

jzt:
@hickeng We discussed making things like `docker images` (aka [`GetImages()`](https://github.com/jzt/vic/blob/32b1c80c4c1a7a8d857a5ed51dc0db8db5ab8b4e/lib/apiservers/engine/backends/cache/image_cache.go#L121) in the cache) block and wait if the update is in progress. During `Update()`, we only lock when [adding an image to the cache](https://github.com/jzt/vic/blob/32b1c80c4c1a7a8d857a5ed51dc0db8db5ab8b4e/lib/apiservers/engine/backends/cache/image_cache.go#L179), but until the update is entirely complete, we need a clean way for `GetImages()` to block and wait. What would be the proper idiomatic Go way to do this?

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -209,3 +174,43 @@ func (c *ImageCache) GetImageByNamed(named reference.Named) (*metadata.ImageConf
...
+
+// AddImage adds an image to the image cache
+func (c *ImageCache) AddImage(imageConfig *metadata.ImageConfig) {
+	c.m.Lock()
+	defer c.m.Unlock()
+	var imageID string
+
+	// Don't assume the image id in image has "sha256:<id> as format.  We store it in
+	// this fomat to make it easier to lookup by digest
+	if strings.HasPrefix(imageConfig.ImageID, "sha") {
```
andrewtchin: Could the prefix be something other than sha256? (why not check for `sha256` as prefix?)

jzt: It can, see [here](https://github.com/docker/docker/blob/3d13fddd2bc4d679f0eaa68b0be877e5a816ad53/vendor/src/github.com/docker/distribution/digest/digester.go#L16).

lib/imagec/imagec.go
```diff
@@ -0,0 +1,665 @@
...
+	Stdout bool
+	Debug  bool
+
+	Outstream io.Writer
+
+	InsecureSkipVerify bool
+	InsecureAllowHTTP  bool
+
+	Standalone bool
+	Resolv     bool
```
caglar10ur: You can nuke this now

lib/imagec/imagec.go
```diff
@@ -0,0 +1,665 @@
...
+		options.Token = token
+	}
+
+	// Get the manifest
+	manifest, err := FetchImageManifest(*options)
+	if err != nil {
+		return nil, fmt.Errorf("Error while pulling image manifest: %s", err)
+	}
+
+	// HACK: Required to learn the name of the vmdk from given reference
```
caglar10ur: resolv related stuff could be dropped

cmd/imagec/imagec.go
```diff
@@ -139,451 +50,33 @@ func init() {
...
-	flag.BoolVar(&options.insecureAllowHTTP, "insecure-allow-http", false, i18n.T("Uses unencrypted connections when fetching images"))
-	flag.BoolVar(&options.standalone, "standalone", false, i18n.T("Disable port-layer integration"))
+	flag.BoolVar(&options.Stdout, "stdout", false, i18n.T("Enable writing to stdout"))
+	flag.BoolVar(&options.Debug, "debug", false, i18n.T("Show debug logging"))
+	flag.BoolVar(&options.InsecureSkipVerify, "insecure-skip-verify", false, i18n.T("Don't verify certificates when fetching images"))
+	flag.BoolVar(&options.InsecureAllowHTTP, "insecure-allow-http", false, i18n.T("Uses unencrypted connections when fetching images"))
+	flag.BoolVar(&options.Standalone, "standalone", false, i18n.T("Disable port-layer integration"))
 
-	flag.BoolVar(&options.resolv, "resolv", false, i18n.T("Return the name of the vmdk from given reference"))
+	flag.BoolVar(&options.Resolv, "resolv", false, i18n.T("Return the name of the vmdk from given reference"))
```
caglar10ur: gone with the wind

caglar10ur:
Just a minor resolv allergy :) Otherwise LGTM

jzt:
@caglar10ur: removed all `resolv` related stuff. Thanks!

lib/imagec/imagec.go
```diff
@@ -613,28 +612,12 @@ func Pull(opt *Options) (*metadata.ImageConfig, error) {
 		return nil, fmt.Errorf("Error while pulling image manifest: %s", err)
 	}
 
-	// HACK: Required to learn the name of the vmdk from given reference
-	// Used by docker personality until metadata support lands
-	if !options.Resolv {
-		progress.Message(po, options.Tag, "Pulling from "+options.Image)
```
caglar10ur: We still need this message :)

jzt: D'OH! Thanks for that.

caglar10ur:
Can't we use a channel to signal GetImages (and serialize access)?

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -66,8 +66,6 @@ func NewImageCache() *ImageCache {
 
 // Update adds new images to the cache
 func (c *ImageCache) Update(client *client.PortLayer) error {
-	c.m.Lock()
-	defer c.m.Unlock()
 
 	CacheNotUpdated = true
```
caglar10ur: By the way, removing this lock caused this var to become a data race :) You can something else here to protect that (channel, atomic var etc (for atomic see https://github.com/vmware/vic/blob/master/lib/dns/cache.go))

jzt: @caglar10ur does it make a difference that `Update()` is now only called one single time, at persona startup?

caglar10ur:
Another quick note, maybe this PR also should remove imagec binary from the ISO (but keep building it via Makefile) to shave some MB @hickeng was not happy with the ISO sizes

jzt:
@caglar10ur: saved us about 7 MB :)

lib/imagec/imagec.go
```diff
@@ -0,0 +1,651 @@
...
+	}
+
+	// calculate image ID
+	sum := fmt.Sprintf("%x", sha256.Sum256(bytes))
+	log.Infof("Image ID: sha256:%s", sum)
+
+	// prepare metadata
+	result.V1Image.Parent = image.Parent
+	result.Size = size
+	result.V1Image.ID = imageLayer.ID
```
jzt: Fixes #1269

lib/apiservers/engine/backends/container.go
```diff
@@ -154,7 +154,7 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 
 	// provide basic container config via the image
 	container := &viccontainer.VicContainer{
-		ID:     image.Parent,
+		ID:     image.ID,
```
jzt: Fixes #1269

caglar10ur:
LGTM :sheep: :it:

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -207,5 +146,45 @@ func (c *ImageCache) GetImageByNamed(named reference.Named) (*metadata.ImageConf
...
+	c.cacheByID[imageID] = imageConfig
+
+	// Normalize the name stored in imageConfig using Docker's reference code
+	ref, err := reference.WithName(imageConfig.Name)
+	if err != nil {
+		log.Errorf("Tried to create reference from %s: %s", imageConfig.Name, err.Error())
+		return
+	}
+
+	for id := range imageConfig.Tags {
```
hickeng: why not use?
`for id, tag := range imageConfig.Tags`

lib/apiservers/engine/backends/image.go
```diff
@@ -117,51 +114,43 @@ func (i *Image) PullImage(ctx context.Context, ref reference.Named, metaHeaders
...
-
-	log.Printf("PullImage: cmd = %s %+v\n", Imagec, cmdArgs)
-
-	cmd := exec.Command(Imagec, cmdArgs...)
-	cmd.Stdout = outStream
-	cmd.Stderr = outStream
-
-	// Execute
-	err := cmd.Start()
+	log.Printf("PullImage: reference: %s, username: %s, password: %s, portlayer: %#v",
```
hickeng: log.Infof, log.Debugf, ...

lib/apiservers/engine/backends/cache/image_cache.go
jzt: Good question, that was @sflxn's code I moved over, but I'll update it.

lib/imagec/imagec.go
```diff
@@ -0,0 +1,651 @@
...
+
+	Stdout bool
+	Debug  bool
+
+	Outstream io.Writer
+
+	InsecureSkipVerify bool
+	InsecureAllowHTTP  bool
+
+	Standalone bool
```
hickeng: Does this still have any meaning?

jzt: I thought about removing it, but if someone is using the binary they should still have that option.

jzt:
Found a critical bug today. Marking this WIP for now.

jzt:
Closing this to reduce noise. Will revisit later when time allows.

PR #1245, opened 2016-06-28
Process restart on failure

### LINKED ISSUE #746: Migrate appliance to tether

hickeng:
This adds restart behaviour to the tether. Because that then requires the ability to do some diagnostics on what occurred I've also added very basic rolling log for restarted components and a resurrection count.
This also includes finer control over the debugging via the vic-machine --debug flag - this still needs to be properly glued through into all the components however.

Fixes #746


lib/tether/tether.go
```diff
@@ -245,14 +264,29 @@ func (t *tether) handleSessionExit(session *SessionConfig) {
...
+		logs = logs[logCount-MaxDeathRecords+1:]
+	}
+
+	session.Diagnostics.ExitLogs = append(logs, metadata.ExitLog{
+		Time:       time.Now(),
+		ExitStatus: session.ExitStatus,
+		// We don't have any message for now
+	})
+
+	f := t.ops.HandleSessionExit(t.config, session)
```
fdawg4l: `f` is the shutdown func (essentially). Probably good to comment that fact.

cmd/tether/tether_test.go
```diff
@@ -109,9 +109,15 @@ func (t *Mocker) SessionLog(session *tether.SessionConfig) (dio.DynamicMultiWrit
 	return dio.MultiWriter(&t.SessionLogBuffer, os.Stdout), nil
 }
 
-func (t *Mocker) HandleSessionExit(config *tether.ExecutorConfig, session *tether.SessionConfig) bool {
+func (t *Mocker) HandleSessionExit(config *tether.ExecutorConfig, session *tether.SessionConfig) func() {
 	// check for executor behaviour
-	return session.ID == config.ID
+	if session.ID == config.ID {
```
fdawg4l: It may be cleaner to do this check inside the lambda so this `HandleSessionExit` never returns nil.

lib/tether/tether.go
```diff
@@ -263,7 +297,7 @@ func (t *tether) launch(session *SessionConfig) error {
 
 	// encode the result whether success or error
 	defer func() {
-		extraconfig.EncodeWithPrefix(t.sink, session.Started, fmt.Sprintf("guestinfo..sessions|%s.started", session.ID))
+		extraconfig.EncodeWithPrefix(t.sink, session, fmt.Sprintf("guestinfo..sessions|%s", session.ID))
```
fdawg4l: double dot.  You can't see me, but I'm shaking my head.  #intuitive.

fdawg4l:
LGTM

lib/tether/tether.go
hickeng: One of the few places where this happens where I don't have the comment:
```
	// FIXME: we cannot have this embedded knowledge of the extraconfig encoding pattern, but not
	// currently sure how to expose it neatly via a utility function

```

hickeng:
@caglar10ur @dougm @hmahmood Need another review to close out #746

dougm:
lgtm

cmd/vic-machine/common/debug.go
```diff
@@ -17,15 +17,15 @@ package common
...
 }
 
 func (d *Debug) DebugFlags() []cli.Flag {
 	return []cli.Flag{
-		cli.BoolFlag{
+		cli.IntFlag{
 			Name:        "debug, v",
 			Destination: &d.Debug,
-			Usage:       "Enable debug level logging",
+			Usage:       "[0(default),1...n], 0 is disabled, 1 is enabled, >= 1 may alter behaviour",
```
caglar10ur: How is this altering the behaviour?

hickeng: At the moment it's only suppressing relaunch of components in the tether, but I expect we may want to have pprof enable, and at high levels listen on external interfaces instead of loopback, etc.

hickeng:
poke

caglar10ur:
LGTM :sheep: :it:

caglar10ur:
don't forget to squash your commits or remove the WIP prefix from them :P

hickeng:
@caglar10ur yup - they always get squashed during merge.

hickeng:
poke

PR #1255, opened 2016-06-29
Introduce --link/--net-alias support

caglar10ur:
- Add -debug flag to the docker engine
- Change various log levels to Debugf
- Set -debug flag in docker-machine by default
- Start passing links to the portlayer
- Start passing aliases to the portlayer (after flattening them)
- Tidy up .gitignore/Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -341,9 +341,5 @@ clean:
 	rm -rf ./lib/apiservers/portlayer/models/
 	rm -rf ./lib/apiservers/portlayer/restapi/operations/
 
-	rm -f lib/apiservers/docker/restapi/doc.go
-	rm -f lib/apiservers/docker/restapi/embedded_spec.go
-	rm -f lib/apiservers/docker/restapi/server.go
-	rm -fr lib/apiservers/docker/cmd
-	rm -fr lib/apiservers/docker/models
-	rm -fr lib/apiservers/docker/restapi/operations
+	rm -f *.log
+	rm -f *.pem
--- a/cmd/docker/main.go
+++ b/cmd/docker/main.go
@@ -106,6 +106,8 @@ func handleFlags() (*CliOptions, bool) {
 	portLayerAddr := flag.String("port-layer-addr", "127.0.0.1", "Port layer server address")
 	portLayerPort := flag.Uint("port-layer-port", 9001, "Port Layer server port")
 
+	debug := flag.Bool("debug", false, "Enable debuglevel logging")
+
 	flag.Parse()
 
 	if *enableTLS && (len(*certfile) == 0 || len(*keyfile) == 0) {
@@ -120,6 +122,10 @@ func handleFlags() (*CliOptions, bool) {
 			fmt.Fprintf(os.Stderr, "tlsverfiy requested, but tls-ca-certificate, tls-certificate, tls-key were all not specified\n")
 			return nil, false
 		}
+	}
+
+	if *debug || vchConfig.Diagnostics.DebugLevel > 0 {
+		log.SetLevel(log.DebugLevel)
 	}
 
 	cli := &CliOptions{
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -136,7 +136,7 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 	var err error
 
 	//TODO: validate the config parameters
-	log.Printf("config.Config = %+v", config.Config)
+	log.Debugf("config.Config = %+v", config.Config)
 
 	// Get an API client to the portlayer
 	client := PortLayerClient()
@@ -184,7 +184,7 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 		config.Name = namesgenerator.GetRandomName(0)
 	}
 
-	log.Printf("ContainerCreate config' = %+v", config)
+	log.Debugf("ContainerCreate config' = %+v", config)
 	// Call the Exec port layer to create the container
 	host, err := guest.UUID()
 	if err != nil {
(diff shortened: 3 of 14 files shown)


lib/portlayer/network/context.go
```diff
@@ -478,15 +486,53 @@ func (c *Context) BindContainer(h *exec.Handle) ([]*Endpoint, error) {
...
 		c.containers[exec.ParseID(key)] = con
+
+		ne, ok := h.ExecConfig.Networks[e.Scope().Name()]
+		if !ok {
+			log.Errorf("Failed to find Network %s", e.Scope().Name())
+			continue
+		}
+
+		// Aliases/Links
+		for i := range ne.Network.Aliases {
```
hmahmood: Could you add some unit test coverage for this? Ditto for the code in Unbind.

caglar10ur: Yep, will do!

hmahmood:
lgtm

Makefile
```diff
@@ -341,9 +341,5 @@ clean:
...
 	rm -rf ./lib/apiservers/portlayer/restapi/operations/
 
-	rm -f lib/apiservers/docker/restapi/doc.go
-	rm -f lib/apiservers/docker/restapi/embedded_spec.go
-	rm -f lib/apiservers/docker/restapi/server.go
-	rm -fr lib/apiservers/docker/cmd
-	rm -fr lib/apiservers/docker/models
-	rm -fr lib/apiservers/docker/restapi/operations
+	rm -f *.log
+	rm -f *.pem
```
hickeng: nice :) should have done this a long time ago.

cmd/docker/main.go
```diff
@@ -106,6 +106,8 @@ func handleFlags() (*CliOptions, bool) {
 	portLayerAddr := flag.String("port-layer-addr", "127.0.0.1", "Port layer server address")
 	portLayerPort := flag.Uint("port-layer-port", 9001, "Port Layer server port")
 
+	debug := flag.Bool("debug", false, "Enable debuglevel logging")
```
hickeng: We should start pulling this data from vchConfig. There are two sources, one is appliance wide, and the other is session specific.
```
vchConfig.ExecutorConfig.Diagnostics.DebugLevel
vchConfig.ExecutorConfig.Sessions["docker-personality"].Diagnostics.DebugLevel
```

lib/apiservers/engine/backends/container.go
```diff
@@ -818,7 +841,7 @@ func attachStreams(name string, tty, stdinOnce bool, clStdin io.ReadCloser, clSt
 				// behavior where you connect to stdin on the first time only.
 				// If we really want to add this behavior, we need to add support
 				// in the ssh tether in the portlayer.
-				log.Printf("Attach stream has stdinOnce set.  VIC does not yet support this.")
+				log.Debugf("Attach stream has stdinOnce set.  VIC does not yet support this.")
```
hickeng: This should be Warnf or Errorf - could be significant if attempting to diagnose behavioural issue for customer.

lib/apiservers/engine/backends/image.go
```diff
@@ -149,14 +149,14 @@ func (i *Image) PullImage(ctx context.Context, ref reference.Named, metaHeaders
 	err := cmd.Start()
 
 	if err != nil {
-		log.Printf("Error starting %s - %s\n", Imagec, err)
+		log.Debugf("Error starting %s - %s\n", Imagec, err)
```
hickeng: This and the one below should be errors - this is in the context of the appliance logs, so logging recoverable errors as errors is okay. We do need a much better contextual logging story.

lib/apiservers/engine/backends/network.go
```diff
@@ -178,6 +178,18 @@ func (n *Network) ConnectContainerToNetwork(containerName, networkName string, e
 		nc.Address = &endpointConfig.IPAMConfig.IPv4Address
 	}
 
+	log.Debugf("EndpointsConfig: %#v", endpointConfig)
```
hickeng: looks identical to the backends/container.go fragment - do these need to be consistent if changes are made? If so this should be a util function

lib/metadata/network_interface.go
```diff
@@ -58,4 +58,7 @@ type ContainerNetwork struct {
 
 	// The IP ranges for this network
 	Pools []ip.Range `vic:"0.1" scope:"read-only" key:"pools"`
+
+	// set of network wide links and aliases for this container on this network
+	Aliases []string `vic:"0.1" scope:"hidden" key:"aliases"`
```
hickeng: where do we store the per-container aliases?

cmd/docker/main.go
caglar10ur: hmm how about keeping this and checking whether the debug flag or vchConfig is set? Otherwise we need guestinfo values all the time

lib/apiservers/engine/backends/container.go
caglar10ur: Done

lib/apiservers/engine/backends/image.go
caglar10ur: Done

lib/apiservers/engine/backends/network.go
caglar10ur: Done

lib/metadata/network_interface.go
caglar10ur: We create a new ContainerNetwork per container so we are storing them there

caglar10ur:
♻️

lib/metadata/network_interface.go
hickeng: In that case how do we differentiate network wide alias from container specific alias? It looks like it's
`target:alias` and therefore a global will be `:alias`?

hickeng:
LGTM

PR #1256, opened 2016-06-29
Add DHCP client to tether. Fix other DHCP issues.

### LINKED ISSUE #747: Milestone: container dhcp support
When attaching containers to a mapped vSphere network, one of the likely IPAM mechanisms available to us is DHCP. This comes with two requirements:

1. dhcp client in bootstrap (perhaps https://github.com/d2g/dhcp4client)
2. reporting of the assigned address (perhaps https://github.com/davecheney/mdns/, or simple update to guestinfo)

There is no need to expose the external mapped network IP via the normal container name resolution as it's fundamentally outside of that managed environment.

#745 is related but is not directly required.

### LINKED ISSUE #1214: Basic address acquisition
This covers basic address acquisition from a DHCP server.
This does not cover renew and release operations.

hmahmood:
Fixes #1214 
Fixes #747


cmd/tether/ops.go
```diff
@@ -36,9 +42,17 @@ func (t *operations) Setup() error {
 }
 
 func (t *operations) Cleanup() error {
+	if err := t.BaseOperations.Cleanup(); err != nil {
```
hickeng: `return t.BaseOperations.Cleanup()`
but I don't think you need this at all if `t` is embedding `BaseOperations`

cmd/tether/ops.go
```diff
@@ -36,9 +42,17 @@ func (t *operations) Setup() error {
...
+		return err
+	}
+
 	return nil
 }
 
+func (t *operations) Apply(endpoint *metadata.NetworkEndpoint) error {
```
hickeng: Same as above

cmd/vch-init/ops.go
```diff
@@ -52,6 +58,10 @@ func (t *operations) Setup() error {
 }
 
 func (t *operations) Cleanup() error {
+	if err := t.BaseOperations.Cleanup(); err != nil {
```
hickeng: same comments as for cmd/tether/ops.go

lib/metadata/network_interface.go
```diff
@@ -59,3 +59,7 @@ type ContainerNetwork struct {
 	// The IP ranges for this network
 	Pools []ip.Range `vic:"0.1" scope:"read-only" key:"pools"`
 }
+
+func (e *NetworkEndpoint) IsDynamic() bool {
```
hickeng: Is this materially different from IsUnspecified?

lib/tether/net_linux_test.go
```diff
@@ -69,7 +69,7 @@ func TestSetIpAddress(t *testing.T) {
 						Name: "bridge",
 					},
 					Default: true,
-					Gateway: gwIP,
+					Gateway: *gwIP,
```
hickeng: found this myself - no idea how this got committed as I always run the package tests directly as well as make test when making tether changes.

hickeng:
@hmahmood Took a look through this - probably easiest to talk through it with you tomorrow.

isos/bootstrap/bootstrap
```diff
@@ -58,24 +58,9 @@ if mount -t ext4 /dev/disk/by-label/containerfs ${MOUNTPOINT}; then
...
-
-    echo "Setting IP addr to " \$addr > /dev/ttyS1
-    echo "Setting route to" \$gateway > /dev/ttyS1
-
-    ip addr add \$addr dev eth1
-    ip link set dev eth1 up
-    ip route add default via \$gateway dev eth1
-fi
-
 /.tether/tether-bin
```
caglar10ur: I think we can drop this wrapper and start using binary as is

lib/dhcp/client.go
```diff
@@ -0,0 +1,174 @@
...
+import (
+	"fmt"
+	"net"
+	"sync"
+	"syscall"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/d2g/dhcp4"
+	"github.com/d2g/dhcp4client"
```
caglar10ur: Not that I've any experience with them but we have https://github.com/vmware/godhcpv4

lib/tether/ops_linux.go
```diff
@@ -442,20 +438,115 @@ func apply(t Netlink, endpoint *metadata.NetworkEndpoint) error {
 			}
 			log.Infof("Added nameserver: %s", server.String())
 		}
-	} else {
+	} else if !ip.IsUnspecifiedIP(endpoint.Network.Gateway.IP) {
```
caglar10ur: Who wins here? Our internal DNS or dhcp provided one?

caglar10ur:
Great work, some nitpicks but overall LGTM

lib/tether/ops_linux.go
hmahmood: For endpoints connected to the bridge network, it will be our internal DNS (i.e. vch). For DHCP endpoints, which should never be the same as the bridge endpoint, it most likely will not go into this branch as we will use the DNS entries from the DHCP server.

hmahmood:
There is a problem with this PR: the dhcp loop does not update endpoints when renewing the dhcp lease. I will fix this, but it will require adding more robust resolv.conf and hosts handling. I will add basic implementations of these.

isos/bootstrap/bootstrap
```diff
@@ -33,14 +33,10 @@ if mount -t ext4 /dev/disk/by-label/containerfs ${MOUNTPOINT}; then
...
-        touch ${MOUNTPOINT}/.tether-init/resolv.conf ${MOUNTPOINT}/.tether-init/resolv.conf
+        touch ${MOUNTPOINT}/etc/hostname
         touch ${MOUNTPOINT}/etc/hosts
-        echo "127.0.0.1 localhost localhost.localdomain" > ${MOUNTPOINT}/.tether-init/hosts
-
-        mount --bind ${MOUNTPOINT}/.tether-init/hostname ${MOUNTPOINT}/etc/hostname
-        mount --bind ${MOUNTPOINT}/.tether-init/resolv.conf ${MOUNTPOINT}/etc/resolv.conf
-        mount --bind ${MOUNTPOINT}/.tether-init/hosts ${MOUNTPOINT}/etc/hosts
+        touch ${MOUNTPOINT}/etc/resolv.conf
+        echo "127.0.0.1 localhost localhost.localdomain" > ${MOUNTPOINT}/etc/hosts
```
caglar10ur: Should we add ipv6 defaults here? My ubuntu box has following defaults

```
127.0.0.1       localhost

# The following lines are desirable for IPv6 capable hosts
::1     localhost ip6-localhost ip6-loopback
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
```

isos/bootstrap/bootstrap
caglar10ur: Now that we have lib/etcconf/hosts.go, I think prefer tether writing the default config files and then dropping this shell wrapper and calling the binary directly.

lib/dhcp/client.go
```diff
@@ -0,0 +1,175 @@
...
+
+// The default timeout for the client
+const defaultTimeout = 10 * time.Second
+
+// NewClient creates a new DHCP client. Note the returned object is not thread-safe.
+func NewClient() (Client, error) {
+	return &client{timeout: defaultTimeout}, nil
+}
+
+func (c *client) SetTimeout(t time.Duration) error {
```
caglar10ur: Why we are returning error here? I think we can just drop that and return nothing

lib/dhcp/packet.go
```diff
@@ -0,0 +1,104 @@
...
+
+	var dns []net.IP
+	for i := 0; i < len(b); i += 4 {
+		dns = append(dns, net.IP(b[i:i+4]))
+	}
+
+	return dns
+}
+
+func (p *Packet) ServerIP() net.IP {
```
caglar10ur: Needs a comment - exported function

lib/metadata/network_interface.go
```diff
@@ -33,7 +33,7 @@ type NetworkEndpoint struct {
 	Static *net.IPNet `vic:"0.1" scope:"read-only" key:"staticip"`
 
 	// Actual IP address assigned
-	Assigned net.IP `vic:"0.1" scope:"read-write" key:"ip"`
+	Assigned net.IPNet `vic:"0.1" scope:"read-write" key:"ip"`
```
caglar10ur: should we change the key to ipnet as well?

hmahmood: I think ip is fine, matches `staticip` above.

isos/bootstrap/bootstrap
hmahmood: i will add those entries.

isos/bootstrap/bootstrap
hmahmood: I will drop the wrapper. However, I am not sure what happens if we boot with an empty hosts file. So I more inclined to have a default there.

lib/tether/config.go
```diff
@@ -88,3 +89,32 @@ type SessionConfig struct {
...
+	// The network in which this information should be interpreted. This is embedded directly rather than
+	// as a pointer so that we can ensure the data is consistent
+	Network metadata.ContainerNetwork `vic:"0.1" scope:"read-only" key:"network"`
+
+	// DHCP runtime info
+	DHCP *DHCPInfo `vic:"0.1" scope:"read-only" recurse:"depth=0"`
+}
+
+func (e *NetworkEndpoint) IsDynamic() bool {
+	return len(e.Static.IP) == 0
```
caglar10ur: ```return e.Static.IPIsUnspecifiedIP()``` maybe?

caglar10ur:
Have some minor questions but overall LGTM

cmd/vic-init/ops.go
```diff
@@ -42,7 +42,12 @@ type operations struct {
 	tether.BaseOperations
 }
 
-func (t *operations) Setup() error {
+func (t *operations) Setup(sink tether.ConfigSink) error {
+	if err := t.BaseOperations.Setup(sink); err != nil {
+		return err
+	}
+
+	// TODO: enabled for initial dev debugging only
```
hickeng: No longer just for dev debugging - now it's restricted to loopback it's there for vicadmin to gather as part of diagnostic logs.

isos/bootstrap/bootstrap
```diff
@@ -32,15 +32,9 @@ if mount -t ext4 /dev/disk/by-label/containerfs ${MOUNTPOINT}; then
         mkdir -p ${MOUNTPOINT}/dev ${MOUNTPOINT}/proc ${MOUNTPOINT}/sys ${MOUNTPOINT}/etc
         # ln -sf /proc/mounts ${MOUNTPOINT}/etc/mtab
 
-        # ensure both the source and destination files exist for bind mounting
-        touch ${MOUNTPOINT}/etc/hostname ${MOUNTPOINT}/.tether-init/hostname
-        touch ${MOUNTPOINT}/.tether-init/resolv.conf ${MOUNTPOINT}/.tether-init/resolv.conf
+        touch ${MOUNTPOINT}/etc/hostname
         touch ${MOUNTPOINT}/etc/hosts
-        echo "127.0.0.1 localhost localhost.localdomain" > ${MOUNTPOINT}/.tether-init/hosts
```
hickeng: Even if there are no network endpoints defined, we still want the loopback entry in hosts.

lib/install/validate/validator.go
```diff
@@ -327,27 +328,12 @@ func (v *Validator) network(ctx context.Context, input *data.Data, conf *metadat
...
-		}
-	}
-
 	bridgeNet := &metadata.NetworkEndpoint{
 		Common: metadata.Common{
 			Name: "bridge",
 			ID:   endpointMoref,
 		},
-		Static: &gateway,
+		Static: &net.IPNet{IP: net.IPv4zero}, // static but managed externally
```
hickeng: nice

lib/portlayer/network/context.go
```diff
@@ -912,6 +953,24 @@ func (c *Context) DeleteScope(name string) error {
 		return fmt.Errorf("scope has bound endpoints")
 	}
 
+	// remove gateway ip from bridge interface
+	link, err := getBridgeLink()
```
hickeng: Should we stash this somewhere on initialization? It seems like something that should be done during portlayer init and likely a fatal failure (without it we cannot do name resolution for containers or port forwarding).

lib/tether/config.go
```diff
@@ -88,3 +89,32 @@ type SessionConfig struct {
...
+
+	// Actual IP address assigned
+	Assigned net.IPNet `vic:"0.1" scope:"read-write" key:"ip"`
+
+	// The network in which this information should be interpreted. This is embedded directly rather than
+	// as a pointer so that we can ensure the data is consistent
+	Network metadata.ContainerNetwork `vic:"0.1" scope:"read-only" key:"network"`
+
+	// DHCP runtime info
+	DHCP *DHCPInfo `vic:"0.1" scope:"read-only" recurse:"depth=0"`
```
hickeng: Are you intending that this be serialized so we can preserve DHCP info across reboots? If so then it needs to be read-write and present before the VM is powered on (i.e. in the lib/metadata structures). If not then we can simply omit the annotation and it won't get serialized in guestinfo as it is read-only by default.

hmahmood: Actually I don't want to serialize it at all. Without the annotation I was getting was not DHCP == nil, which I need in the code.

isos/bootstrap/bootstrap
hmahmood: The loopback entries are now done in BaseOperations.Setup as per @caglar10ur suggestion.

lib/tether/ops_linux.go
```diff
@@ -539,3 +643,61 @@ func (t *BaseOperations) Fork() error {
...
+	for _, e := range entries {
+		h.SetHost(e.hostname, e.addr)
+	}
+
+	if err = h.Save(); err != nil {
+		return err
+	}
+
+	// start with empty resolv.conf
+	os.Remove("/etc/resolv.conf")
```
hickeng: Pull this path from the const in lib/etcconf?

lib/tether/ops_linux.go
```diff
@@ -539,3 +643,61 @@ func (t *BaseOperations) Fork() error {
...
+
+	rc := etcconf.NewResolvConf("")
+	if err = rc.Load(); err != nil {
```
hickeng: Why do we call load having just deleted the source file?

hickeng:
LGTM
minor comments and questions - built it but not yet done any testing with it. Would like to sit down with you next time I'm in Austin or before if necessary and spend a day going over tether restructure to make this an easier integration.

hmahmood:
Poke

hmahmood:
Poke

Issue #742, opened 2016-05-20
Tether: vsphere guest operations support


hickeng:
There is a mechanism available in guest tools called "guest operations" - this causes a process to be executed within the guestOS. This implies there is a mechanism of some kind available to trigger operations within the guest as I refuse to believe vmtoolsd polls continuously.

I suspect this is related to #406 and #407, but regardless this is necessary for picking up reconfigure operations within the containerVMs.

dougm:
I have this working (in sloppyjoe.go form) where we can receive RPC calls from the vmx.  There's a few RPCs we need to support before its possible to run in-guest commands: reset, ping, set option, get tools status, etc.  With proper responses to these RPCs in place to make the vmx happy, I can use the API to trigger a start-program RPC in guest like so:

```
govc guest.start -vm dougm-ubuntu-1604 -l user:pass -C /tmp -e FOO=bar /bin/date --date=@2147483647
```

Within sloppyjoe.go, we are able to decode the RPC request (credentials, current directory, env vars, program path, program args) - in the case of tether we can support pseudo commands with this RPC path, to reconfigure the container VM for example.

Also have the path to report the guest IP address when requested, and heartbeat (for HA).

It is actually a "poll" on the guest side, via the backdoor (vmtoolsd does this too).  I see plumbing in vmtoolsd to use a vsocket for the transport, not being used by default with 6.0 afaict.  But regardless if transport is the backdoor or vsocket, the RPC messages are in the same format.  So, any decode/encode we implement would apply to either transport.

I'm expecting to open a PR by tomorrow with the basic framework in decent form.

hickeng:
@dougm awesome news doug! the only thing we're using this channel for as of TP3 is the signal delivery, so please prioritize that over other RPC responses. I was initially thinking of doing this via a read-write key that the tether would erase once the signal was delivered, and using a reload to pick it up, but if you think there's a cleaner way that doesn't involve a guestinfo key directly, then feel free.

dougm:
I still have some testing/cleanup/doc/etc todo before opening a PR, which should be by tomorrow for sure.  In the meantime, my WIP is here: https://github.com/vmware/vic/compare/master...dougm:toolbox

mdubya66:
is this still required for tp3? @dougm @hickeng?

dougm:
[...]

Issue #407, opened 2016-04-03
Add IP address reporting


hickeng:
vSphere relies on guest tools to report the IP addresses of a VM - we do not want to have the full guest tools installed as it is way too large, but we do want the option to report and IP address that will show up in the standard location in vSphere.

This is likely to be implemented in the same library as rpctool as I believe they rely on the same communication mechanism

hickeng:
@caglar10ur @dougm there was an rpctool command added to force a guest info refresh as it turns out there's a refresh period.
Implemented by "vmware-toolbox-cmd info update network"

hickeng:
The following is the line in question in open-vm-tools for reporting the IP.
https://github.com/vmware/open-vm-tools/blob/master/open-vm-tools/services/plugins/guestInfo/guestInfoServer.c#L1594

hickeng:
@mdubya66 we should try for this for onsite. It's almost free once we get #1267 done.

corrieb:
Assigning to @dougm given the dependency on #1267

dougm:
Implemented most of what we need for this as part of #742 
We just need to integrate toolbox with tether and we get the IP Address reporting "for free".  This step needs to happen on the way to #834 anyhow.

dougm:
Have this working: https://github.com/vmware/vic/compare/master...dougm:tether-toolbox

I'm going to add integration tests before submitting a PR.

Issue #406, opened 2016-04-03
Add heartbeat support


hickeng:
vSphere HA depends on a heartbeat mechanism, which I believe is supplied by guest tools.
We wish to leverage HA to implement container restart policies - for that we need heartbeat.

I would like to add this into the library that also provides rpctool function as I believe the low level comms mechanisms are the same - if I'm incorrect perhaps this can be a separate library.

My intent is that this library be used by tether - tether will check for the liveliness of the processes it's responsible for, and if those are present and responsive the heartbeat continues. 'checking' the liveliness may be as simple as watching for SIGCHLD, or a more involved check.

hickeng:
github.com/vmware/open-vm-tools has the code to support heartbeat - this needs to extract it from the rest of tools and consume it. If at all possible let's keep it pure Go so that it's portable.

hickeng:
@dougm this is a direct follow on from the push to tether work (i.e. it's the channel in the other direction).

dougm:
Not sure we need to add anything to tether/toolbox for VM HA to work.  Let's try the test on shared storage first, see: https://github.com/vmware/vic/issues/3784#issuecomment-276741268

mdubya66:
removing the high priority based on @dougm's assessment. setting to low

corrieb:
Note that Dockerfile now has a HEALTHCHECK option and you can specify a command to be run periodically to determine if a container is healthy. Docker engine will generate an event if it becomes unhealthy.

hmahmood:
I can confirm that HA works without the heartbeat (and shared storage). Should we close this issue and track support for health check in another?

corrieb:
HA works in a limited set of circumstances. It works in the event of host failure, but doesn't in the event of workload failure. Unless there's something I'm misunderstanding. That's what this is designed to address, right?

hmahmood:
If by workload failure you mean tether failure (panic), then we reboot the VM currently, and we could modify that behavior to conform to whatever restart policy the user wants. I am not sure we need  heartbeat for this, since vsphere will monitor network traffic and cpu (?) if no heartbeat support is present in the toolbox. @dougm could you confirm that is the case since you stated above heartbeat is not needed.

If by workload failure you mean the container process crashes, then I believe we will restart the process, and again we can modify that behavior to support what the user wants.

dougm:
@hmahmood my comment was related to VM HA when a host fails, and referenced a comment that the HA test was broken (using local storage).

[...]

PR #1257, opened 2016-06-29
Add guest tools framework

dougm:
Initial version to support vmware-tools "lite" in pure Go.

Towards:

Issue #742
Issue #407
Issue #406

### DIFF
--- a/pkg/vsphere/toolbox/backdoor.go
+++ b/pkg/vsphere/toolbox/backdoor.go
@@ -1,0 +1,73 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !windows
+
+package toolbox
+
+import (
+	"errors"
+
+	"github.com/vmware/vmw-guestinfo/message"
+	"github.com/vmware/vmw-guestinfo/vmcheck"
+)
+
+const (
+	rpciProtocol uint32 = 0x49435052
+	tcloProtocol uint32 = 0x4f4c4354
+)
+
+var (
+	ErrNotVirtualWorld = errors.New("not in a virtual world")
+)
+
+type backdoorChannel struct {
+	protocol uint32
+
+	*message.Channel
+}
+
+func (b *backdoorChannel) Start() error {
+	if !vmcheck.IsVirtualWorld() {
+		return ErrNotVirtualWorld
+	}
+
+	channel, err := message.NewChannel(b.protocol)
+	if err != nil {
+		return err
+	}
+
+	b.Channel = channel
+
+	return nil
+}
+
+func (b *backdoorChannel) Stop() error {
+	err := b.Channel.Close()
+	return err
+}
+
+// NewBackdoorChannelOut creates a Channel for use with the RPCI protocol
+func NewBackdoorChannelOut() Channel {
+	return &backdoorChannel{
+		protocol: rpciProtocol,
+	}
+}
+
+// NewBackdoorChannelOut creates a Channel for use with the TCLO protocol
+func NewBackdoorChannelIn() Channel {
+	return &backdoorChannel{
+		protocol: tcloProtocol,
+	}
+}
--- a/pkg/vsphere/toolbox/service.go
+++ b/pkg/vsphere/toolbox/service.go
@@ -1,0 +1,208 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package toolbox
+
+import (
+	"bytes"
+	"fmt"
+	"log"
+	"net"
+	"os"
+	"sync"
+	"time"
+)
+
+// Service receives and dispatches incoming RPC requests from the vmx
+type Service struct {
+	name     string
+	in       Channel
+	out      Channel
+	handlers map[string]Handler
+	stop     chan struct{}
+	wg       *sync.WaitGroup
+
+	Interval  time.Duration
+	PrimaryIP func() string
+}
+
+// NewService initializes a Service instance
+func NewService(rpcIn Channel, rpcOut Channel) *Service {
+	s := &Service{
+		name:     "toolbox", // Same name used by vmtoolsd
+		in:       NewTraceChannel(rpcIn),
+		out:      NewTraceChannel(rpcOut),
+		handlers: make(map[string]Handler),
+		wg:       new(sync.WaitGroup),
+		stop:     make(chan struct{}, 1),
+
+		Interval:  time.Second,
+		PrimaryIP: DefaultIP,
+	}
+
+	s.RegisterHandler("reset", s.Reset)
+	s.RegisterHandler("ping", s.Ping)
+	s.RegisterHandler("Set_Option", s.SetOption)
+	s.RegisterHandler("Capabilities_Register", s.CapabilitiesRegister)
+
+	return s
+}
+
+// Start initializes the RPC channels and starts a goroutine to listen for incoming RPC requests
+func (s *Service) Start() error {
+	err := s.in.Start()
+	if err != nil {
+		return err
+	}
+
+	err = s.out.Start()
+	if err != nil {
+		return err
+	}
+
+	ticker := time.NewTicker(s.Interval)
+
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
+
+		for {
+			select {
+			case <-ticker.C:
+				_ = s.in.Send(nil) // POKE
+
+				request, _ := s.in.Receive()
+
+				if len(request) > 0 {
+					response := s.Dispatch(request)
+
+					_ = s.in.Send(response)
+				}
+			case <-s.stop:
+				ticker.Stop()
+				return
+			}
+		}
+	}()
+
+	return nil
+}
+
+// Stop cancels the RPC listener routine created via Start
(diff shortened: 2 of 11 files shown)


caglar10ur:
👯 👯 👯

pkg/vsphere/toolbox/service.go
```diff
@@ -0,0 +1,201 @@
...
+	err = s.out.Start()
+	if err != nil {
+		return err
+	}
+
+	ticker := time.NewTicker(s.Interval)
+
+	s.wg.Add(1)
+	go func() {
+		defer s.wg.Done()
```
casualjim: image

pkg/vsphere/toolbox/service.go
```diff
@@ -0,0 +1,201 @@
...
+func (s *Service) Dispatch(request []byte) []byte {
+	msg := bytes.SplitN(request, []byte{' '}, 2)
+	name := msg[0]
+
+	// Trim NULL byte terminator
+	if len(name) > 0 && name[len(name)-1] == 0 {
+		name = name[:len(name)-1]
+	}
+
+	handler, ok := s.handlers[string(name)]
```
casualjim: I think you may want to guard this, reads from maps are not safe for concurrent use in golang

dougm: That's true if you have concurrent writes to the map, but we do not in this case.  All writes to this map (registering handlers) happen before starting the service.

dougm:
I haven't tried within tether/container VM yet.  Have tested cmd/toolbox on Ubuntu (devbox on ESX), just requires that you turn off the existing tools first:
```
sudo service open-vm-tools stop
```

I used debian to test a VM that doesn't have tools pre-installed:
```
govc datastore.upload debian-live-8.5.0-amd64-standard.iso debian-live.iso
govc vm.create -iso debian-live.iso debian-live
govc vm.vnc -enable -port -1 -password 6b33546b debian-live | awk '{print $NF}' | xargs open
```

Within the VNC console, boot live debian (user/pass is user/live).
```
/sbin/ifconfig # see your current IP
ssh-keygen
cat .ssh/id_rsa.pub > .ssh/authorized_keys
scp .ssh/id_rsa your-machine:debian-id
```

On your-machine:
```
ssh-add debian-id
scp $GOPATH/bin/toolbox debian-ip
ssh debian-ip sudo ./toolbox # start tools-lite
```

See tools-lite in action:
```
govc vm.ip debian-live
192.168.247.243

govc guest.start -vm debian-live -l user:pass kill SIGHUP
-1
```

You'll see the following toolbox trace output:
```
ignoring credentials: 'user:pass'
guest-command: kill SIGHUP
SEND 10...
00000000  4f 4b 20 30 20 30 20 2d  31 00                    |OK 0 0 -1.|
```

pkg/vsphere/toolbox/vix_property.go
```diff
@@ -0,0 +1,223 @@
...
+	_ = iota // ANY type not supported
+	vixPropertyTypeInt32
+	vixPropertyTypeString
+	vixPropertyTypeBool
+	_ // HANDLE type not supported
+	vixPropertyTypeInt64
+	vixPropertyTypeBlob
+)
+
+const (
```
caglar10ur: Should we add some comments to explain where are those coming from?

pkg/vsphere/toolbox/vix_command.go
```diff
@@ -0,0 +1,342 @@
...
+	c.handlers[op] = handler
+}
+
+func (c *VixRelayedCommandHandler) GetToolsState(_ string, _ VixCommandRequestHeader, _ []byte) ([]byte, error) {
+	hostname, _ := os.Hostname()
+	osname := fmt.Sprintf("%s-%s", runtime.GOOS, runtime.GOARCH)
+
+	// Note that vmtoolsd sends back 40 or so of these properties, sticking with the minimal set for now.
+	props := VixPropertyList{
+		NewStringProperty(VixPropertyGuestOsVersion, osname),
```
caglar10ur: Maybe we should make them const and also add a comment where to find them?

pkg/vsphere/toolbox/trace_channel.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"encoding/hex"
+	"fmt"
+	"io"
+	"os"
+)
+
+var (
+	Trace = true // TODO: make optional
```
caglar10ur: Instead of doing that maybe you could add two files one with functions doing nothing and have a build flag 
```
// +build !trace
...
func (d *TraceChannel) Send(buf []byte) error {
    return nil
}
...
```
and another with 
```
// +build trace
...
+func (d *TraceChannel) Send(buf []byte) error {
 +	if len(buf) > 0 {
 +		fmt.Fprintf(d.log, "SEND %d...\n%s\n", len(buf), hex.Dump(buf))
 +	}
 +
 +	err := d.Channel.Send(buf)
 +
 +	return err
 +}
```

and enable disable tracing with a build tag (go build -tags trace ...) and call those functions without decorating if statements?

caglar10ur:
LGTM :sheep: :it:

pkg/vsphere/toolbox/backdoor.go
```diff
@@ -0,0 +1,71 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
```
hickeng: This is linux specific given the vmw-guestinfo package - should it be backdoor_linux.go?

pkg/vsphere/toolbox/service.go
```diff
@@ -0,0 +1,210 @@
...
+		out:      NewTraceChannel(rpcOut),
+		handlers: make(map[string]Handler),
+		wg:       new(sync.WaitGroup),
+		stop:     make(chan struct{}, 1),
+
+		Interval:  time.Second,
+		PrimaryIP: DefaultIP,
+	}
+
+	s.RegisterHandler("reset", s.Reset)
```
hickeng: Is this the guest reset action? Same action for shutdown?

pkg/vsphere/toolbox/service.go
```diff
@@ -0,0 +1,210 @@
...
+type Handler func([]byte) ([]byte, error)
+
+// RegisterHandler for the given RPC name
+func (s *Service) RegisterHandler(name string, handler Handler) {
+	s.handlers[name] = handler
+}
+
+// Dispatch an incoming RPC request to a Handler
+func (s *Service) Dispatch(request []byte) []byte {
+	msg := bytes.SplitN(request, []byte{' '}, 2)
```
hickeng: Is it possible to get nil from the channel in the same way we Send(nil) as a poke?

pkg/vsphere/toolbox/service.go
```diff
@@ -0,0 +1,210 @@
...
+	name := msg[0]
+
+	// Trim NULL byte terminator
+	if len(name) > 0 && name[len(name)-1] == 0 {
```
hickeng: Not sure which is more performant but this is concise:
`name = strings.TrimRight(name, "\x00")`

pkg/vsphere/toolbox/service.go
```diff
@@ -0,0 +1,210 @@
...
+
+	return nil, nil
+}
+
+// DefaultIP is used by default when responding to a Set_Option broadcastIP request
+// It can be overridden with the Service.PrimaryIP field
+func DefaultIP() string {
+	addrs, err := net.InterfaceAddrs()
+	if err == nil {
+		for _, addr := range addrs {
```
hickeng: Any idea how this is supposed to be determined? We could vet whether the default route is associated with the interface.

hickeng:
LGTM
We need to be careful not to allow direct command execution in the containerVMs.
Do you have a plan for how to register a callback on message so tether can validate and perform the kill?

pkg/vsphere/toolbox/vix_property.go
dougm: All of the [Vv]ix constants are from vix.h, I was thinking about moving them all to a vix.go.  Maybe even generate vix.go from vix.h, but there's a bunch that we don't need.

pkg/vsphere/toolbox/vix_command.go
dougm: You mean the values such as "VMware Tools (Go)"?  In any case, I agree, I'll add some comments/references to where we derive these from open-vm-tools.

pkg/vsphere/toolbox/trace_channel.go
dougm: I could just move the TraceChannel usage to the example cmd/toolbox/main.go for now.  Was thinking it would be a useful runtime option, at least while developing the toolbox further.

pkg/vsphere/toolbox/backdoor.go
dougm: This file does compile and run on Darwin as well, tho if you run cmd/toolbox/main.go it'll bail with an error "not in a virtual world" - but in theory that wouldn't happen if run within a VM.  Likewise for Windows.

pkg/vsphere/toolbox/service.go
dougm: No, the "reset" RPC is for the guest<->vmx channel itself, not related to VM power operations.  You'll see a "reset" request at startup, with cmd/toolbox/main.go for example.

pkg/vsphere/toolbox/service.go
dougm: It is possible, and common, for Channel.Receive() to return nil - but in that case we'd skip the path that calls into this Service.Dispatch method.  But when Channel.Receive() does return a []byte, we'll at least get a msg[0] here and there's a guard below checking for 'if len(msg) == 2', as there may or may not be additional arguments depending on the RPC.

pkg/vsphere/toolbox/service.go
dougm: Yes, in general I'd like to make the []byte mangling more concise - so I'll make that change here.

pkg/vsphere/toolbox/service.go
dougm: tools/vmomi doesn't care what value we send back, so it is up to us to decide which IP would be the most useful to report.  There is also another RPC to report an array of NIC info, I haven't looked at yet - but in any case the IP value reported here is the most prominent in the UI and most commonly read via the API (e.g. govc vm.ip)

dougm:
@hickeng this PR doesn't support execution of arbitrary commands.  See cmd/toolbox/main.go, the example sets vix.ProcessStartCommand to just print the command path and args.  So the tether can set this func to only handle a pseudo 'kill' command.  We could also make use of the name/pass credentials to validate, but that is currently ignored.

pkg/vsphere/toolbox/backdoor.go
hickeng: however the bdoor package doesn't build on Windows.

hickeng:
Not sure why LGTM isn't picked up

pkg/vsphere/toolbox/backdoor.go
dougm: Ah right, the code in sigma/bdoor is unix only.  I've added a tag for now so backdoor.go isn't built on windows:

```
// +build !windows
```

We can look at porting bdoor to windows in the future and/or use Go asm instead of cgo.

dougm:
rebased

Issue #1189, opened 2016-06-23
Resource pool for VCH being created with reservation


hickeng:
The resource pools for VCHs are being created with a reservation of 42. There should be no reservation on the pool unless it's explicitly set by the user as a way of putting a floor on VCH resources.

PR #1260, opened 2016-06-29
Remove resource pool limitation

emlin:


andrewtchin:
lgtm

cgtexmex:
Would be nice to provide a description -- maybe a brief "why" of this change...otherwise LGTM

emlin:
We still didn't expose parameters to set VCH limitation at this time, so we should not have a default reservation, although that is not a big number, empty resource pool will still consume resources.

In the future, we'll expose those parameters, and user could change this limitation manually, to limit VCH resources, and all container VMs created by this VCH.

emlin:
> Would be nice to provide a description -- maybe a brief "why" of this change
@hickeng think this might be related to our current CI issue, so I had this quick PR.

mhagen-vmware:
@cgtexmex When we kill jobs in the CI part of the problem is this reservation.  It hangs around and consumes 42MB of memory and never gets cleaned up.  So that builds up rather quickly and crashes all of our CI testing.  According to George, it isn't needed at all and is just an excessive reservation that doesn't help anything in later on.

emlin:
Fixes #1189

PR #1270, opened 2016-06-29
Cleanup any VMs that are not related to running drone jobs [skip ci]

mhagen-vmware:
Wipes out any VMs that are not associated with running drone jobs and cleans up any datastore entry that doesn't have an active vmx file in it.


mdubya66:
lgtm

matthewavery:
lgtm

PR #1272, opened 2016-06-29
Use image vmdk leaf layer for container config

### LINKED ISSUE #1269: VIC is not using leaf layer to create the image
For tutumcloud/dnsutils image(https://github.com/tutumcloud/dnsutils/blob/master/Dockerfile)

Docker
```[vagrant@devbox:~] docker run -it tutum/dnsutils
root@cfd419d0a610:/# which dig
/usr/bin/dig
root@cfd419d0a610:/# exit
exit
[vagrant@devbox:~]
```
VIC
[vagrant@devbox:/opt/go/src/github.com/vmware/vic(link)] docker -H 10.143.2.191:2376 --tls attach d62e84a18270e74ae450ea7b9f200c296bf9cd6d42476b084a
a3f9d04aae2562
root@d62e84a18270:/#
root@d62e84a18270:/# which dig
root@d62e84a18270:/#
```

Looks like we are not using the leaf layer but one before it.

cgtexmex:
Image metadata was missing the layer id for the image layer (leaf layer).  Additionally the container create was configuring the container with the leaf layer's parent.  Both of those issues have been corrected.

Fixes #1269

### DIFF
--- a/cmd/imagec/imagec.go
+++ b/cmd/imagec/imagec.go
@@ -561,6 +561,7 @@ func CreateImageConfig(images []*ImageWithMeta, manifest *Manifest) error {
 	// prepare metadata
 	result.V1Image.Parent = image.Parent
 	result.Size = size
+	result.V1Image.ID = imageLayer.ID
 	metaData := metadata.ImageConfig{
 		V1Image: result.V1Image,
 		ImageID: sum,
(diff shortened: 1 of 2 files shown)


caglar10ur:
#1238 already have this snippet, either @jzt needs to drop those bits or you need to drop this PR. Let us know which way to go. If you prefer this PR than it LGTM

cgtexmex:
@caglar10ur I'm fine with either -- @jzt ran into some issues with his change, so I was pushing this in case that didn't show up for a bit.

hickeng:
LGTM

caglar10ur:
♻️

PR #1273, opened 2016-06-29
Minor Volumes changes

fdawg4l:
Moved `storage*` files to `image_` since they are primarily for images.  We now have `volume_` files for volumes, so this is now consistent.

Set the volume label to the the computed label (the first 16 characters of the md5).

Also fix some asserts in the volumes REST API tests.


matthewavery:
lgtm

hickeng:
LGTM

PR #1274, opened 2016-06-29
Volume PL +  Docker Engine Server integration

matthewavery:
This is the Handler code which connects the Volume portion of the storage layer to the docker engine server. The only thing missing is the extraconfig for populating the volumestore. This work will be included in a separate PR. The only major changes involve parameter validation and cache setup in the portlayer storage handlers.


fdawg4l:
LGTM

lib/apiservers/engine/backends/volume.go
```diff
@@ -28,9 +28,9 @@ import (
 	"github.com/vmware/vic/pkg/trace"
 )
 
-//NOTE: These might be moved to a utility package once there are multiple personalities
-const optsVolumeStoreKey string = "VolumeStore"
-const optsCapacityKey string = "Capacity"
+//NOTE: FIXME: These might be moved to a utility package once there are multiple personalities
+const OptsVolumeStoreKey string = "VolumeStore"
+const OptsCapacityKey string = "Capacity"
 const dockerMetadataModelKey string = "DockerMetaData"
```
caglar10ur: use
```
const (
...
) 
```
format please

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -237,3 +281,16 @@ func convertImage(image *spl.Image) *models.Image {
...
 	}
 }
+
+//utility functions
+
+func volumeToCreateResponse(volume *spl.Volume, model *models.VolumeRequest) models.VolumeResponse {
+	response := models.VolumeResponse{
+		Driver:   model.Driver,
+		Name:     volume.ID,
+		SelfLink: volume.SelfLink.String(), //FIXME: is this right?
```
caglar10ur: This is a very scary FIXME right there. Why do we have it, what's not clear and what can we do to clarify?

matthewavery: I am renaming that to label. and passing the label into it. There was some decision making to do on what to return from the portlayer.

lib/apiservers/portlayer/swagger.yml
```diff
@@ -249,6 +249,8 @@ paths:
             $ref: "#/definitions/VolumeResponse"
         '500':
           description: "Error"
+          $schema:
+            $ref: "#/definitions/Error"
         default:
           description: "error"
```
sflxn: I know this wasn't added as part of this PR, but is a total waste.  It won't do what it was originally meant to do (capture generated client errors).  Defining a catch all error without an http status code feels very wrong to me.

fdawg4l: The status code is 500 @sflxn.  And as long as the response is relevant, I don't see the code as being a problem unless you want to map all possible error codes (http://www.gnu.org/software/libc/manual/html_node/Error-Codes.html) to an http error code.

In any case, we're not going to address this here.  Please create an issue and describe an alternative.

gigawhitlocks:
LGTM

matthewavery:
♻️

Issue #1279, opened 2016-06-30
Milestone: non-root users inside containers


jzt:
Currently we only support root users inside containers. @stuclem: We should document why this is not a high priority. Ping myself or @hickeng for info.

mdubya66:
Moving to Backlog. This needs more details and needs to be documented.

stuclem:
This issue is included as a known issue in the release notes for TP3.

compendius:
Is  --cap-drop supported? For example  --cap-drop ALL . This would limit root capabilities in a container

hickeng:
@compendius The containers are isolated VMs so --cap-drop doesn't add additional security from the point of view of container isolation as that's provided by the hypervisor, not the kernel.

You could use it _within_ the scope of a container if wanting to isolate processes further for some reason.

stuclem:
Already in the release notes, so removing the kind/note tag.

hickeng:
This was closed by #4385

Issue #1032, opened 2016-06-09
[tether] should use a default and sane PATH


caglar10ur:
**VIC Version:**

master

**Code**

```ts
[vagrant@devbox:/opt/go/src/github.com/vmware/vic(master)] $Z create -it busybox
148e86e8327c9e4c3a21f98074a1ec53d2bc79547340877b8f349259aca1475b
[vagrant@devbox:/opt/go/src/github.com/vmware/vic(master)] $Z start 148e86e8327c9e4c3a21f98074a1ec53d2bc79547340877b8f349259aca1475b
Error response from daemon: server error from portlayer
Error: failed to start containers: 148e86e8327c9e4c3a21f98074a1ec53d2bc79547340877b8f349259aca1475b
[vagrant@devbox:/opt/go/src/github.com/vmware/vic(master)] $Z create -it busybox /bin/sh
5ea8e1f9ab808677221759a9505878689c71683a8bfcac7952bf9a49fffe7979
[vagrant@devbox:/opt/go/src/github.com/vmware/vic(master)] $Z start 5ea8e1f9ab808677221759a9505878689c71683a8bfcac7952bf9a49fffe7979
5ea8e1f9ab808677221759a9505878689c71683a8bfcac7952bf9a49fffe7979
[vagrant@devbox:/opt/go/src/github.com/vmware/vic(master)]
```

**Expected behavior:**

Both should work

**Actual behavior:**

First one fails with 
```
INFO[0000] Loaded executor config
INFO[0000] [BEGIN] [github.com/vmware/vic/lib/tether.(*BaseOperations).SetHostname] setting hostname to
INFO[0000] [ END ] [github.com/vmware/vic/lib/tether.(*BaseOperations).SetHostname] [91.116µs] setting hostname to
INFO[0000] [BEGIN] [github.com/vmware/vic/lib/tether.(*BaseOperations).Apply] applying endpoint configuration for bridge
INFO[0000] [ END ] [github.com/vmware/vic/lib/tether.(*BaseOperations).Apply] [555.819µs] applying endpoint configuration for bridge
INFO[0000] Launching process for session 148e86e8327c9e4c3a21f98074a1ec53d2bc79547340877b8f349259aca1475b
INFO[0000] [BEGIN] [github.com/vmware/vic/lib/tether.(*tether).launch] launching session 148e86e8327c9e4c3a21f98074a1ec53d2bc79547340877b8f349259aca1475b
[...]

PR #1280, opened 2016-06-30
Reject non-root users

### LINKED ISSUE #725: Add default docker environment
Docker adds default environment to every container. This needs to be merged with the config specified on the command line and in the layer config before calling create on the port layer. This interacts with #411.

The defaults [are detailed here](https://docs.docker.com/engine/reference/run/#env-environment-variables):
```
#Set based on the value of USER
HOME=... 

#The truncated id of the container
HOSTNAME=... 

PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"

# if the container is allocated a pseudo-TTY
TERM="xterm" 
```

jzt:
This change rejects container create for non-root supplied `-user` names (see #1279).

Fixes #725, #1032

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -185,6 +185,14 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 	}
 	config.Config.Env = append(config.Config.Env, container.Config.Env...)
 
+	// TODO(jzt): users other than root are not currently supported
+	// We should check for USER in config.Config.Env once we support Dockerfiles.
+	if config.Config.User != "" && config.Config.User != "root" {
+		return types.ContainerCreateResponse{},
+			derr.NewErrorWithStatusCode(fmt.Errorf("Failed to create container - users other than root are not currently supported"),
+				http.StatusInternalServerError)
+	}
+
 	// Was a name provided - if not create a friendly name
 	if config.Name == "" {
 		//TODO: Assume we could have a name collison here : need to


caglar10ur:
Don't we need to export the Hostname value as a HOSTNAME env. var for the container?

andrewtchin:
lgtm

jzt:
@caglar10ur Fixed.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -71,7 +71,11 @@ func (handler *ContainersHandlersImpl) CreateHandler(params containers.CreatePar
 	log.Debugf("Env: %#v", params.CreateConfig.Env)
 	log.Debugf("WorkingDir: %#v", params.CreateConfig.WorkingDir)
 
-	id := exec.GenerateID().String()
+	containerID := exec.GenerateID()
+	id := containerID.String()
+
+	// inject hostname into environment
+	params.CreateConfig.Env = append(params.CreateConfig.Env, fmt.Sprintf("HOSTNAME=%s", containerID.TruncateID()))
```
caglar10ur: I think we need to move this into  container.go, next to this guy https://github.com/vmware/vic/blob/master/lib/apiservers/engine/backends/container.go#L172

caglar10ur: [hit comment too early]

As this is not a portlayer feature but a personality

jzt: Agreed that it's a personality thing, but we don't get the ID back until the container is already created (and `ENV` already injected). I tossed around the idea of creating the ID in the personality itself, but @hmahmood objected on the grounds that it should be generated in the portlayer.

This is why I originally didn't have it in the environment vars, but only in the local shell vars. :)

caglar10ur:
Hmmm, didn't like doing that in port-layer at all. Maybe tether can export that env var while setting the hostname (https://github.com/vmware/vic/blob/73b7ad8161ea9ffa595bee6ad1eb4c870b258937/lib/tether/ops_linux.go#L112). @hickeng any objections to that?

jzt:
@caglar10ur @hickeng @hmahmood It seems like everyone is in agreement that this should go into the persona. I know @hickeng doesn't want this to happen in the tether for this reason. What would stop us from creating the container ID in the persona instead of the portlayer? This is otherwise a trivial issue and I don't really see a reason to punt on it.

caglar10ur:
Just wondering why we don't want that in tether? we are already setting the hostname via tether inside the container.

hmahmood:
I agree that this is a docker thing, but I don't see how it hurts us to put this in the port layer. Besides putting this in the tether is injecting docker behavior even deeper into the runtime. HOSTNAME is actually not set in env even on ubuntu, and I think tether should mimic that. Putting this in the port layer is a good compromise, imho.

caglar10ur:
How about separating ID generation from the container creation. We could first call portlayer.ReserveID (with very short expiration time) and then calling portlayer.Create with that reservedID.

hickeng:
I'm not seeing this debate get resolved in a TP3 timeframe. Let's drop the HOSTNAME portion and put in the check for non-root users alone.

jzt:
@hickeng Done. Apologies for putting this at a lower priority the last few days.

hickeng:
LGTM

jzt:
♻️

PR #1288, opened 2016-06-30
Adding pprof data collection to all cmds

### LINKED ISSUE #1243: Add pprof logging for all appliance components
#1217 adds an easy mechanism for collecting pprof debug data from all components, however we need to:
a. enable pprof on a know port for each component
b. add those addresses to the array in vicadmin

kjplatz:
Fixes #1243

This involved refactoring the pprof data collection to use
more user-friendly names (instead of http://127.0.0.1:6060/
for example).  Pprof support has now been added to vicadmin,
vch-init (aka vic-init), the portlayer server, the docker
personality, and the imagec server


lib/pprof/pprof.go
```diff
@@ -0,0 +1,59 @@
...
+	DockerPort
+	PortlayerPort
+	maxPort
+)
+
+func StartPprof(name string, component PprofPort) error {
+	port := component + basePort
+	log.Info(fmt.Sprintf("Launching %s server on port %d", name, port))
+	go func() {
+		log.Info(http.ListenAndServe(fmt.Sprintf("127.0.0.1:%d", port), nil))
```
fdawg4l: It's only listening on `lo`?  Is that intentional?  How do we get to this from outside the vch?

lib/pprof/pprof.go
```diff
@@ -0,0 +1,59 @@
...
+	log "github.com/Sirupsen/logrus"
+)
+
+type PprofPort int
+
+const basePort = 6060
+
+const (
+	VchInitPort PprofPort = iota
+	ImagecPort
```
caglar10ur: You can drop imagec as we are not running it as a daemon

lib/pprof/pprof.go
caglar10ur: Can we make that a value based on the debug level @hickeng introduced? (Like listen * if debug is enabled)

lib/pprof/pprof.go
kjplatz: @hickeng requested this be added for imagec in the event that it were to run into an infinite loop and/or hang.  We do not expect imagec to generate any logs, but if it is behaving abnormally, it may.

lib/pprof/pprof.go
kjplatz: It only listens on lo; the logs are accessed via the vicadmin web page.

cmd/imagec/imagec.go
```diff
@@ -44,6 +44,7 @@ import (
 	"github.com/vmware/vic/lib/apiservers/portlayer/models"
 	"github.com/vmware/vic/lib/guest"
 	"github.com/vmware/vic/lib/metadata"
+	"github.com/vmware/vic/lib/pprof"
```
caglar10ur: Same as below, you can drop imagec bits

cmd/vicadmin/vicadm.go
```diff
@@ -135,51 +138,55 @@ func logFiles() []string {
...
-		commandReader("ip route"),
-		commandReader("lsmod"),
+		"concise": "/debug/pprof/goroutine?debug=1",
+		"block":   "/debug/pprof/block?debug=1",
+		"heap":    "/debug/pprof/heap?debug=1",
+	}
+
+	pprofSources := map[string]string{
+		"docker":    pprof.GetPprofEndpoint(pprof.DockerPort).String(),
+		"imagec":    pprof.GetPprofEndpoint(pprof.ImagecPort).String(),
```
caglar10ur: drop imagec

cmd/vicadmin/vicadm.go
```diff
@@ -311,8 +318,11 @@ func httpEntry(name string, res *http.Response) (entry, error) {
 
 func (path urlReader) open() (entry, error) {
 	defer trace.End(trace.Begin(string(path)))
-
-	res, err := http.Get(string(path))
+	timeout := time.Duration(2 * time.Second)
+	client := http.Client{
+		Timeout: timeout,
```
caglar10ur: Will 2 seconds be enough to read everything?

caglar10ur: Also please make this a const

lib/pprof/pprof.go
caglar10ur: Yeah but https://github.com/vmware/vic/pull/1238 will deprecate imagec so instead of adding here, merging that and removing these bits with another PR, I suggest not having them in the first place

cmd/vicadmin/vicadm.go
kjplatz: It's only a few K or megs of log files over a network internal to the vSphere environment.  The alternative is for users to wait potentially a very long time if every collector hangs.

cmd/vicadmin/vicadm.go
```diff
@@ -311,8 +318,11 @@ func httpEntry(name string, res *http.Response) (entry, error) {
 
 func (path urlReader) open() (entry, error) {
 	defer trace.End(trace.Begin(string(path)))
-
-	res, err := http.Get(string(path))
+	const timeout = time.Duration(2 * time.Second)
```
caglar10ur: Uhh I meant moving this to the top of the file, eg;

```
const (
    timeout = 2*time.Second
)
```

cmd/vicadmin/vicadm.go
caglar10ur: I'm just saying 2 seconds is too optimistic as a timeout, and I' m guessing it's just a random number you picked. Note that timeout is not a connection establishment time but that includes everything including reading the body.

Bump it to some sane worst case (yet another random number but bigger than 2), like 60 or 600 as  it won't affect the normal/expected case, it's just for bounding the worst case.

cmd/vic-init/ops.go
```diff
@@ -43,12 +43,9 @@ type operations struct {
 }
 
 func (t *operations) Setup() error {
-	log.Info("Launching pprof server on port 6060")
-	go func() {
-		log.Info(http.ListenAndServe("127.0.0.1:6060", nil))
-	}()
+	err := pprof.StartPprof("vch-init", pprof.VchInitPort)
```
hickeng: Doesn't need fixing, but for future reference: Go idiom is for acronyms to be capitalized, so this would be VCHInitPort

cmd/vicadmin/vicadm.go
```diff
@@ -135,51 +142,55 @@ func logFiles() []string {
...
+		"proc-self-fd": commandReader("ls -l /proc/self/fd"),
 	}
 
 	// add the pprof collection
-	for _, source := range pprofSources {
-		for _, paths := range pprofPaths {
-			readers = append(readers, urlReader(source+paths))
+	for sname, source := range pprofSources {
+		for pname, paths := range pprofPaths {
+			rname := sname + "/" + pname
```
hickeng: This should use either:
`fmt.Sprintf("%s/%s", sname, pname)`, or
`path.Join(sname, pname)`

Strings in Go are immutable and avoiding unnecessary allocations is key to performant code. Not fundamentally necessary for this PR.

lib/pprof/pprof.go
hickeng: #1238 is not going in yet - @jzt has hit some issues that need resolving. I'm not sure it'll go in for TP3, so let's leave this as is for now unless it has the potential to cause imagec functional issues - @caglar10ur ?

caglar10ur: Yeah, sure if that's the case

cmd/vic-init/ops.go
caglar10ur: go vet might cry about it so we may very well fix it while the code is in review

lib/pprof/pprof.go
hickeng: Getting to this is a little more involved that it should be (error checking is omitted):
```
ip := "127.0.0.1"
self := vchConfig.ExecutorConfig.Sessions["vicadmin"]
if vchConfig.ExecutorConfig.Diagnostics.DebugLevel > 1 || (self != nil && self.Diagnostics.DebugLevel > 1) {
   ip = net.LookupIP("client.localhost")
}
```

Unfortunately that would conflict with docker port forwarding in the case client network role is shared with external.

cmd/vic-init/ops.go
```diff
@@ -43,12 +43,9 @@ type operations struct {
...
 
 func (t *operations) Setup() error {
-	log.Info("Launching pprof server on port 6060")
-	go func() {
-		log.Info(http.ListenAndServe("127.0.0.1:6060", nil))
-	}()
+	err := pprof.StartPprof("vch-init", pprof.VCHInitPort)
 
-	return nil
+	return err
```
caglar10ur: you can directly return, no need to create an intermediate variable
```
return pprof.StartPprof("vch-init", pprof.VCHInitPort)
```

cmd/vicadmin/vicadm.go
```diff
@@ -46,15 +46,20 @@ import (
...
+	"github.com/vmware/vic/lib/pprof"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/compute"
 	"github.com/vmware/vic/pkg/vsphere/extraconfig"
 	"github.com/vmware/vic/pkg/vsphere/session"
 	"github.com/vmware/vic/pkg/vsphere/vm"
 )
 
+const (
+	timeout = time.Duration(2 * time.Second)
```
caglar10ur: I still think we need to bump it, 2 seconds will cause issues with large vc's with esxi's on slow networks.

caglar10ur:
LGTM (except 2 sec. timeout which I hope will be bumped to something bigger (a minute etc))

cmd/vicadmin/vicadm.go
hickeng: This is internal to the appliance so network latency not an issue - we end up blocking with log bundle collection for `components * timeout` seconds and to keep it responsive I'd prefer to have this reasonably low.
Wish there was a way to detect nested to automatically x10 but no.

We can bump it to 5 to allow for massive CPU ready time in nested envs - but would prefer to externalize it eventually. We just don't have a structured place to put it for now (I don't want to just keep dumping finely granular config into the metadata/VirtualContainerHostConfigSpec without more structure as it'll be awkward to change.

caglar10ur: Oh this is not how we download logs from ESXi and used only for pprof logs? If that's the case yeah, it doesn't matter that much...

lib/pprof/pprof.go
```diff
@@ -0,0 +1,87 @@
...
+var (
+	vchConfig metadata.VirtualContainerHostConfigSpec
+)
+
+func init() {
+	// load the vch config
+	// TODO: Optimize this to just pull the fields we need...
+	src, err := extraconfig.GuestInfoSource()
+	if err != nil {
+		log.Errorf("Unable to load configuration from guestinfo")
```
hickeng: return after this line - I've added code into extraconfig.Decode to address unit test failures, but if we have an error getting src there's no way Decode will function.

lib/pprof/pprof.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func StartPprof(name string, component PprofPort) error {
+	port := component + basePort
+	if component != ImagecPort {
+		log.Info(fmt.Sprintf("Launching %s server on port %d", name, port))
+	}
+
+	ip := "127.0.0.1"
+	self, ok := vchConfig.ExecutorConfig.Sessions["vicadmin"]
```
hmahmood: A comment here would help. Same goes for the special case for imagec above.

Also rename `self` to something less generic.

lib/pprof/pprof.go
```diff
@@ -0,0 +1,90 @@
...
+
+	return nil
+}
+
+func GetPprofEndpoint(component PprofPort) *url.URL {
+	if component >= maxPort {
+		return nil
+	}
+	port := component + basePort
+	endpoint, err := url.Parse(fmt.Sprintf("http://127.0.0.1:%d", port))
```
hmahmood: Shouldn't this match what you spawned in the go routine above?

lib/pprof/pprof.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		log.Errorf("Unable to load configuration from guestinfo")
+		return
+	}
+	extraconfig.Decode(src, &vchConfig)
+}
+
+func StartPprof(name string, component PprofPort) error {
+	port := component + basePort
+	if component != ImagecPort {
```
hickeng: I we're dropping imagec because it was causing issues, then let's just remove it from the set of components for which we're launching pprof. Filtering it here makes little sense.

lib/pprof/pprof.go
hickeng: Despite me having written this sample as reference, on a second read embedding "vicadmin" like this is very ugly. Let's drop the self portion and just check `vchConfig.ExecutorConfig.Diagnostics.DebugLevel`

lib/pprof/pprof.go
```diff
@@ -0,0 +1,90 @@
...
+		log.Info(fmt.Sprintf("Launching %s server on port %d", name, port))
+	}
+
+	ip := "127.0.0.1"
+	self, ok := vchConfig.ExecutorConfig.Sessions["vicadmin"]
+	if vchConfig.ExecutorConfig.Diagnostics.DebugLevel > 1 || (ok && self.Diagnostics.DebugLevel > 1) {
+		ips, err := net.LookupIP("client.localhost")
+		if err != nil {
```
hickeng: `if err != nil || len(ips) == 0` - let's not assume that no error means an address was found as the doc doesn't specify those semantics.

lib/pprof/pprof.go
```diff
@@ -0,0 +1,99 @@
...
+
+func StartPprof(name string, component PprofPort) error {
+	url := GetPprofEndpoint(component)
+	if url == nil {
+		err := errors.New("Unable to get pprof endpoint")
+		log.Error(err.String())
+		return err
+	}
+
+	if component != ImagecPort {
```
hmahmood: Why the check?

lib/pprof/pprof.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return nil
+	}
+	return endpoint
+}
+
+func StartPprof(name string, component PprofPort) error {
+	url := GetPprofEndpoint(component)
+	if url == nil {
+		err := errors.New("Unable to get pprof endpoint")
```
hmahmood: Additional detail about which endpoint would help here.

kjplatz: Added in the name of the endpoint.

lib/pprof/pprof.go
kjplatz: Fixed.

lib/pprof/pprof.go
hmahmood: Where?

hmahmood:
LGTM

kjplatz:
Comment

kjplatz:
Comment

Issue #1285, opened 2016-06-30
vic-machine-linux failed:  Creating resource pool failed: ServerFaultCode: A specified parameter was not correct:


cormachogan:
**VIC Version:**

0.1 / master / 2416

// A self-contained demonstration of the problem follows...

Trying to create a new VCH using vic-machine-linux from #2416

```ts
root@photon-NaTv5i8IA [ /workspace/vic ]# ./vic-machine-linux create  --bridge-network Bridge-DPG --image-datastore isilion-nfs-01 -t "administrator@vsphere.local@10.27.51.103" --compute-resource Mgmt
INFO[2016-06-30T10:37:12Z] Please enter ESX or vCenter password:
INFO[2016-06-30T10:37:16Z] ### Installing VCH ####
INFO[2016-06-30T10:37:16Z] Generating certificate/key pair - private key in ./docker-appliance-key.pem
INFO[2016-06-30T10:37:16Z] Validating supplied configuration
INFO[2016-06-30T10:37:16Z] Firewall status: DISABLED on /CNA-DC/host/Mgmt/10.27.51.8
INFO[2016-06-30T10:37:16Z] Firewall configuration OK on hosts:
INFO[2016-06-30T10:37:16Z]   /CNA-DC/host/Mgmt/10.27.51.8
INFO[2016-06-30T10:37:17Z] License check OK on hosts:
INFO[2016-06-30T10:37:17Z]   /CNA-DC/host/Mgmt/10.27.51.8
INFO[2016-06-30T10:37:17Z] DRS check OK on:
INFO[2016-06-30T10:37:17Z]   /CNA-DC/host/Mgmt/Resources
INFO[2016-06-30T10:37:18Z] Creating Resource Pool docker-appliance
INFO[2016-06-30T10:37:18Z] Collecting a0522635-224e-4171-92b1-7edaf726b39f vpxd.log
vic-machine-linux failed:  Creating resource pool failed: ServerFaultCode: A specified parameter was not correct:
```

Here is the vpxd.log:

```ts
root@photon-NaTv5i8IA [ /workspace/vic ]# cat vpxd.log
[...]

PR #1291, opened 2016-06-30
Revert "Remove resource pool limitation"

andrewtchin:
This was a breaking change - reverting while we fix
This reverts commit 81aae035218a7e56e945c6cc2df7c0cd13a9dbda.


matthewavery:
lgtm

hickeng:
LGTM
@emlin @mhagen-vmware if this passed testing, how did that occur?

emlin:
> @emlin @mhagen-vmware if this passed testing, how did that occur?

This is because the problem only happens in VC, and we don't test VC in any integration test yet.

andrewtchin:
♻️

emlin:
Fixes #1285

Issue #1129, opened 2016-06-17
Attach hangs when container has stopped


hickeng:
We do not abort the wait for a connection for a prolonged period in the case that the containerVM has already stopped before attach starts. There are two issues illustrated here:
1. the use of a closed network connection doesn't either:
  a. log a message about reconnecting
  b. cause us to disconnect
2. even on the PL side we don't discover the dropped connection for approximately 90s

From personality while running `docker -H 192.168.78.138:2376 --tls run -i busybox /bin/dmesg`
```
time="2016-06-17T07:57:20Z" level=info msg="[BEGIN] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).ContainerCreate] ContainerCreate" 
time="2016-06-17T07:57:20Z" level=info msg="config.Config = &{Hostname: Domainname: User: AttachStdin:true AttachStdout:true AttachStderr:true ExposedPorts:map[] PublishService: Tty:false OpenStdin:true StdinOnce:true Env:[] Cmd:[/bin/dmesg] ArgsEscaped:false Image:busybox Volumes:map[] WorkingDir: Entrypoint:[] NetworkDisabled:false MacAddress: OnBuild:[] Labels:map[] StopSignal:}" 
time="2016-06-17T07:57:25Z" level=info msg="v1 = {ID:437595becdebaaaf3a4fc3db02c59a980f955dee825c153308c670610bb694e1 Config:{Hostname:156e10b83429 Domainname: User: AttachStdin:false AttachStdout:false AttachStderr:false ExposedPorts:map[] PublishService: Tty:false OpenStdin:false StdinOnce:false Env:[] Cmd:[sh] ArgsEscaped:false Image:56ed16bd6310cca65920c653a9bb22de6b235990dcaa1742ff839867aed730e5 Volumes:map[] WorkingDir: Entrypoint:[] NetworkDisabled:false MacAddress: OnBuild:[] Labels:map[] StopSignal:}}" 
time="2016-06-17T07:57:25Z" level=info msg="ContainerCreate config' = {Name:hungry_hypatia Config:0xc8203aeb40 HostConfig:0xc820336380 NetworkingConfig:0xc82001c3d0 AdjustCPUShares:false}" 
[...]

PR #1295, opened 2016-06-30
Attempt to detect EOF from Swagger during docker attach

sflxn:
Swagger is catching EOF but returns an untyped error to us with a
string that contains a string with 'EOF' in it.  We now check for
this substring.

This should help with #1129 a little.  We still need to address
the timeouts and checking if the container's process has stopped.

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


caglar10ur:
Uhhhh, that means I can return an error message "This is not EOF" and caused a termination, no? Why we are not adding a new error type (EOF etc.) and returning that? That way we can type assert instead of string matching.

sflxn:
No, you can't send a "This is not EOF" message.  All returnable errors in go-swagger are checked first.  The 'unknown' errors are internal go-swagger client errors that it returns.  There is no way for you to return an unknown error with that message or any message.

caglar10ur:
OK, lgtm then

jzt:
lgtm

sflxn:
lgtm.  no changes since the previous two lgtm other than rebase and wait for a successful CI.

PR #1302, opened 2016-06-30
Fix vic-machine delete to parse compute resource

### LINKED ISSUE #1296: vic-machine delete does not take resource pool path <cluster path>/<pool>
The parameter --compute-resource Austin/line --name line001 works for vic-machine-linux create, but not for delete.

emlin:
Fixes #1296

  this fix will have vic-machine delete support relative path and 'cluster'/'pool' format for compute-resource, which usually works for vic-machine create.


andrewtchin:
lgtm

PR #1304, opened 2016-06-30
adds storage layer implementation of VolumeJoin, handler code for Vol… [ci skip]

matthewavery:
Implements VolumeJoin and provides functionality for a `docker create -v` call to succeed. This PR will involve changes to the docker engine server and the portlayer. Mostly it involves modifying the config spec of the container so that on commit the vmdk is mounted appropriately. It also involves logic for creating a default volume if an anonymous volume is the `-v` target.

### DIFF
--- a/doc/design/volumes.md
+++ b/doc/design/volumes.md
@@ -39,7 +39,7 @@ type Volume struct {
 // VolumeStorer is an interface to create, remove, enumerate, and get Volumes.
 type VolumeStorer interface {
 	// Creates a volume on the given volume store, of the given size, with the given metadata.
-	VolumeCreate(ctx context.Context, ID string, store *url.URL, capacityMB uint64, info map[string][]byte) (*Volume, error)
+	VolumeCreate(ctx context.Context, ID string, store *url.URL, capacityKB uint64, info map[string][]byte) (*Volume, error)
 
 	// Get an existing volume via it's ID.
 	VolumeGet(ctx context.Context, ID string) (*Volume, error)
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -295,8 +295,9 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 		//NOTE: for now we are passing the flags directly through. This is NOT SAFE and only a stop gap.
 		flags["Mode"] = fields.VolumeFlags
 		joinParams := storage.NewVolumeJoinParams().WithJoinArgs(&models.VolumeJoinConfig{
-			Flags:  flags,
-			Handle: h,
+			Flags:     flags,
+			Handle:    h,
+			MountPath: fields.VolumeDest,
 		}).WithName(fields.VolumeID)
 
 		res, err := client.Storage.VolumeJoin(joinParams)
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -40,6 +40,8 @@ import (
 	"github.com/docker/engine-api/types/container"
 	"github.com/docker/engine-api/types/strslice"
 
+	"github.com/google/uuid"
+
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	viccontainer "github.com/vmware/vic/lib/apiservers/engine/backends/container"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client"
@@ -55,6 +57,12 @@ import (
 
 // Container struct represents the Container
 type Container struct {
+}
+
+type volumeFields struct {
+	VolumeID    string
+	VolumeDest  string
+	VolumeFlags string
 }
 
 const (
@@ -142,8 +150,7 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 	var err error
 
 	//TODO: validate the config parameters
-	log.Debugf("config.Config = %+v", config.Config)
-
+	log.Printf("config.Config = %+v", config.Config)
 	// Get an API client to the portlayer
 	client := PortLayerClient()
 	if client == nil {
@@ -263,6 +270,25 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 		}()
 
 		h = addContRes.Payload
+	}
+
+	//Volume Attachment Section
+	for i, v := range config.HostConfig.Binds {
+		fields, err, shouldExist := processVolumeParam(v)
+		var volumeResponse models.VolumeResponse
+		if err != nil {
+			return nil, derr.NewErrorWithStatusCode(fmt.Errorf("Server error from Portlayer: %s", err), http.StatusBadRequest)
+		}
+		if !shouldExist {
+			//FIXME: in the future a default Capacity will be expected from ExtraConfig<vic-machine>
+			volumeRquest := models.VolumeRequest{
+				Capacity: 1024,
+				Driver:   "vsphere",
+				Store:    "default",
+			}
+			client.Storage.CreateVolume(storage.NewCreateVolumeParams().WithVolumeRequest(volumeRequest * models.VolumeRequest))
+		}
+
 	}
 
 	// commit the create op
@@ -1430,4 +1456,33 @@ func copyEscapable(dst io.Writer, src io.ReadCloser, keys []byte) (written int64
 // to the docker client approved format
 func clientFriendlyContainerName(name string) string {
 	return fmt.Sprintf("/%s", name)
+}
+
+func processVolumeParam(volString string) (volumeFields, error, bool) {
+	volumeStrings := strings.Split(volString, ":")
+	fields := volumeFields{}
+	created := true
+	switch len(volumeStrings) {
+	case 1:
+		VolumeID, err := uuid.NewUUID()
+		if err != nil {
+			return volumeFields{}, nil, false
+		}
+		fields.VolumeID = VolumeID.String()
+		fields.VolumeDest = volumeStrings[0]
+		fields.VolumeFlags = "rw"
+		created = false
+	case 2:
+		fields.VolumeID = volumeStrings[0]
+		fields.VolumeDest = volumeStrings[1]
+		fields.VolumeFlags = "rw"
+	case 3:
+		fields.VolumeID = volumeStrings[0]
+		fields.VolumeDest = volumeStrings[1]
+		fields.VolumeFlags = volumeStrings[2]
+	default:
+		//NOTE: the docker cli should cover this case. This is here for posterity.
+		return volumeFields{}, fmt.Errorf("Volume bind input is invalid : -v %s", volString), false
+	}
+	return fields, nil, created
 }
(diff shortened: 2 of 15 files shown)


lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -18,6 +18,8 @@ import (
 	"net/http"
 	"os"
 
+	// "github.com/google/uuid"
```
fdawg4l: Can probably remove this.

fdawg4l:
LGTM.  Just a minor comment.

andrewtchin:
lgtm

doc/design/volumes.md
```diff
@@ -39,7 +39,7 @@ type Volume struct {
 // VolumeStorer is an interface to create, remove, enumerate, and get Volumes.
 type VolumeStorer interface {
 	// Creates a volume on the given volume store, of the given size, with the given metadata.
-	VolumeCreate(ctx context.Context, ID string, store *url.URL, capacityMB uint64, info map[string][]byte) (*Volume, error)
+	VolumeCreate(ctx context.Context, ID string, store *url.URL, capacityKB uint64, info map[string][]byte) (*Volume, error)
```
caglar10ur: IMHO one way to enforce this is to define a kb type and passing that instead of uint64.That way caller had to create/cast a kb instance so will know that you expect KB. Alternatively you can convert this into a VolumeCreateOptions struct with named members.

fdawg4l: The type gets obfuscated by swagger in this case, so creating a `KB` type will require a swagger model and then a named parameter that the user will need to set.  Eventually we devolve to using names in one way (the model) or another (the parameter name).

lib/apiservers/engine/backends/container.go
```diff
@@ -57,6 +59,12 @@ import (
 type Container struct {
 }
 
+type volumeFields struct {
```
fdawg4l: I'm surprised you aren't getting vet/lint warnings on this about "volume" stuttering.  I suggest removing "Volume" from the member names.  Also consider renaming the struct to something like `volumeParams`.  Your call, just suggestions.

lib/apiservers/engine/backends/container.go
```diff
@@ -142,8 +150,8 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 	var err error
 
 	//TODO: validate the config parameters
-	log.Debugf("config.Config = %+v", config.Config)
-
+	log.Printf("Image fetch section - Container Create")
```
fdawg4l: What does this message actually mean?  I.e. is it useful in debugging?  Can it be made clearer?

lib/apiservers/engine/backends/container.go
```diff
@@ -142,8 +150,8 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
...
+	log.Printf("config.Config = %+v", config.Config)
```
fdawg4l: This looks useful.  But is it redundant?  I thought we print out all this stuff somewhere already when we create the container.  Could be wrong.

lib/apiservers/engine/backends/container.go
```diff
@@ -212,6 +220,7 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 		return types.ContainerCreateResponse{}, derr.NewRequestNotFoundError(fmt.Errorf("No command specified"))
 	}
 
+	log.Printf("ContainerCreate config' = %+v", config)
```
fdawg4l: This is redundant, no?  Is this debug code?

lib/apiservers/engine/backends/container.go
matthewavery: I added the log.printf(maybe it should be log.Infof) to help break down the massive `ContainerCreate` call. so now you can see the individual sections of the create call you are passing through in the logs. We don't actually need it. But that function is massive and makes several calls to the portlayer. I found it useful to atleast have some logging around which section the call might be failing in.

lib/apiservers/engine/backends/container.go
```diff
@@ -264,7 +274,39 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 
 		h = addContRes.Payload
 	}
+	//Volume Attachment Section
+	msg, name, time := trace.Begin("Container.ContainerCreate - VolumeSection")
+	log.Infof("Raw Volume arguments : binds:  %#v : volumes : %#v", config.HostConfig.Binds, config.Config.Volumes)
```
andrewtchin: Should this be debug?

lib/apiservers/engine/backends/container.go
matthewavery: Yes it looks like there is a redundant print. I will fix that.

lib/apiservers/engine/backends/container.go
matthewavery: Yeah probably should

lib/apiservers/engine/backends/container.go
matthewavery: Most likely was, I will remove all of these except for the original one.

lib/apiservers/engine/backends/container.go
```diff
@@ -1431,3 +1473,72 @@ func copyEscapable(dst io.Writer, src io.ReadCloser, keys []byte) (written int64
 func clientFriendlyContainerName(name string) string {
 	return fmt.Sprintf("/%s", name)
 }
+
+func processVolumeParam(volString string) (volumeFields, error) {
```
fdawg4l: Nit. I know there's a convention, but a 1500 line file is a bit long.  Please consider moving these `*Volume*` functions to their own file in this package.  `container_volumes.go` or something like that.  Your call.

lib/apiservers/engine/backends/container.go
```diff
@@ -1431,3 +1473,72 @@ func copyEscapable(dst io.Writer, src io.ReadCloser, keys []byte) (written int64
...
+		//NOTE: This should be the guard for the case of an anonymous volume.
+		//NOTE: we should not expect any driver args if the drive is anonymous.
+		log.Infof("anonymous volume being created - Container Create - volume mount section")
+		metadata := make(map[string]string)
+		metadata["flags"] = fields.VolumeFlags
+		volumeRequest := models.VolumeRequest{
+			Capacity: -1,
+			Driver:   "vsphere",
+			Store:    "default",
+			Name:     fields.VolumeID,
```
fdawg4l: Can we add the VolumeID to the `Infof` above.

lib/apiservers/engine/backends/container.go
```diff
@@ -1431,3 +1473,72 @@ func copyEscapable(dst io.Writer, src io.ReadCloser, keys []byte) (written int64
...
+			Metadata: metadata,
+		}
+		_, err = client.Storage.CreateVolume(storage.NewCreateVolumeParams().WithVolumeRequest(&volumeRequest))
+		if err != nil {
+			return volumeFields, err
```
fdawg4l: `return nil, err`

lib/apiservers/engine/backends/container.go
```diff
@@ -1431,3 +1473,72 @@ func copyEscapable(dst io.Writer, src io.ReadCloser, keys []byte) (written int64
...
+}
+
+func processAnonymousVolumes(h *string, volumes map[string]struct{}, client *client.PortLayer) ([]volumeFields, error) {
+	var volumeFields []volumeFields
+
+	for v := range volumes {
+		fields, err := processVolumeParam(v)
+		log.Infof("Processed Volume arguments : %#v", fields)
+		if err != nil {
+			return volumeFields, err
```
fdawg4l: I'm not sure what to do here.  We tried to process some args, some args *were* processed, but we have an error.  Is the semantic to continue the container `Create` with what worked and ignore the error or stop everything?  I prefer stopping everything by returning nil, err.

lib/apiservers/engine/backends/container.go
```diff
@@ -1431,3 +1473,72 @@ func copyEscapable(dst io.Writer, src io.ReadCloser, keys []byte) (written int64
 func clientFriendlyContainerName(name string) string {
 	return fmt.Sprintf("/%s", name)
 }
+
+func processVolumeParam(volString string) (volumeFields, error) {
+	volumeStrings := strings.Split(volString, ":")
+	fields := volumeFields{}
+	switch len(volumeStrings) {
```
fdawg4l: I think this deserves a comment at least.  Just include what the expected invocations are and the resulting params.

lib/apiservers/engine/backends/container.go
fdawg4l: Returning a value instead of a reference. Is that intentional?

lib/apiservers/engine/backends/container.go
matthewavery: Ah not redundant I deleted the original for some reason... I fixed it.

lib/apiservers/engine/backends/container.go
matthewavery: I was thinking about that actually. I have been unhappy with how large this file is as well. I don't like having to memorize line numbers to navigate a file. Since these are part of the engine I would rather move all container files into the container directory of the repo as well. Perhaps that can be a cleanup for another time though.

lib/apiservers/engine/backends/container.go
matthewavery: Yeah, that would be much more useful.

lib/apiservers/engine/backends/container.go
matthewavery: renamed the members of the struct.

fdawg4l:
LGTM again.  Nice work!

lib/apiservers/engine/backends/container.go
```diff
@@ -264,7 +273,39 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 
 		h = addContRes.Payload
 	}
+	//Volume Attachment Section
+	msg, name, time := trace.Begin("Container.ContainerCreate - VolumeSection")
```
hickeng: If you want to use trace here instead of log.Debug then this should really be a separate helper function or wrapped in an inline function (with comment as to why) and use defer for trace.End. At the moment we have error paths that would result in no matching End output.
```
func() {
  defer trace.End(trace.Begin("Container.Create - volumes"))
  // code goes here
}()
```

lib/portlayer/storage/vsphere/vm.go
```diff
@@ -0,0 +1,75 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
```
hickeng: not sure this warrants it's own file but keep it for now.

lib/portlayer/storage/vsphere/vm.go
```diff
@@ -0,0 +1,75 @@
...
+	defer trace.End(trace.Begin("vsphere.VolumeJoin"))
+
+	if _, ok := handle.ExecConfig.Mounts[volume.ID]; ok {
+		return nil, fmt.Errorf("Volume with ID %s is already in container %s's mountspec'", volume.ID, handle.Container.ExecConfig.ID)
+	}
+
+	newMountSpec := metadata.MountSpec{
+		Source: url.URL{
+			Scheme: "label",
+			Host:   volume.Label,
```
hickeng: This should have been in my PR:
```
Path: volume.Label
```

hickeng:
LGTM

PR #1293, opened 2016-06-30
Initial support for volume stores in vic-machine

gigawhitlocks:
This change allows the creation of volume stores at install time.

PR #1305, opened 2016-06-30
Get volume stores from the vch's extra config

### LINKED ISSUE #429: Milestone: volume lifecycle commands
Use the docker volume command to create named volumes.

This encompasses volumes that live past the life of one container and are used by many in serial.
This is not a concurrent or non-persistent volume.

This will be implemented using attached VMDKs as in Bonneville, and should leverage the portlayer storage component.

After implementation the following commands should function:
```
docker volume create
docker volume ls
docker volume rm
docker run -v name:/mount/path ...
```

fdawg4l:
Fixes #429

vic-machine will add the volume store names and their datastore + path
to the extra config of the vch when it instantiates it.  The portlayer
will now retrieve this, iterate it, and call `AddStore` for each of
these so the docker user can refer to the store by the given name.  The
resulting volume will be created in the path above.

With #1293 and this change, things work.

```
$ ./vic-machine-linux create ... --image-datastore datastore1 --volume-store doobie:datastore1/voloom
...
$ docker -H 192.168.8.223:2376 --tls volume create --name MyVolume -o VolumeStore=doobie -o Capacity=1024
MyVolume
...
[root@esxbox:/vmfs/volumes/56dcdc1c-ce4398df-7d12-000c29fc9a54] ls -lah voloom/VIC/volumes/MyVolume/
...
-rw-------    1 root     root        1.0G Jun 30 21:17 MyVolume-flat.vmdk
-rw-------    1 root     root         494 Jun 30 21:17 MyVolume.vmdk

```

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
@@ -31,7 +31,7 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/session"
 
 	spl "github.com/vmware/vic/lib/portlayer/storage"
-	"github.com/vmware/vic/lib/portlayer/storage/vsphere"
+	vsphereSpl "github.com/vmware/vic/lib/portlayer/storage/vsphere"
 	"github.com/vmware/vic/lib/portlayer/util"
 
 	"golang.org/x/net/context"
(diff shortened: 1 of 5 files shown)


gigawhitlocks:
lgtm

caglar10ur:
LGTM :sheep: :it:

fdawg4l:
.

PR #1323, opened 2016-07-01
Return deep copy of image config from cache

### LINKED ISSUE #1265: Command line for container persists in subsequent creates with the same image
`docker create -it busybox /bin/ip addr`

And then run the subsequently created container.

Now create another container:
`docker create -it busybox`
(sans the command)

Run the container, and it runs the command from the previous run of the first container.

jzt:
The docker `V1Image` struct embedded in `ImageConfig` contains a pointer to a `container.Config`. Container create was overwriting members of that `Config` struct and that was corrupting the cache. The image cache now returns an `ImageConfig` with a deep copy of the `container.Config` struct to prevent consumers from having this issue.

Fixes #1265

### DIFF
--- a/lib/apiservers/engine/backends/cache/image_cache.go
+++ b/lib/apiservers/engine/backends/cache/image_cache.go
@@ -28,6 +28,7 @@ import (
 	"github.com/docker/distribution/digest"
 	derr "github.com/docker/docker/errors"
 	"github.com/docker/docker/reference"
+	"github.com/docker/engine-api/types/container"
 
 	"github.com/vmware/vic/lib/apiservers/portlayer/client"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/storage"
@@ -163,9 +164,7 @@ func (c *ImageCache) GetImages() ([]*metadata.ImageConfig, error) {
 
 	result := make([]*metadata.ImageConfig, 0, len(c.cacheByID))
 	for _, image := range c.cacheByID {
-		newImage := new(metadata.ImageConfig)
-		*newImage = *image
-		result = append(result, newImage)
+		result = append(result, copyImageConfig(image))
 	}
 
 	return result, nil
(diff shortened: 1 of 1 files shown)


andrewtchin:
lgtm

caglar10ur:
Since you are copying the same object I suggest using reflect.Copy here to shorten the code

caglar10ur:
Hmm wait ignore this as I think Copy cannot handle structs

caglar10ur:
But I think we can still do this via regular assignment, unless I miss something. See https://play.golang.org/p/qMNY_cOryW

jzt:
@caglar10ur We are having a slightly different problem: Copying the struct makes a copy of the pointer (so it still points to the same memory). Container create writes data to the struct that is pointed to, which is actually the original.

Example:
https://play.golang.org/p/Mez5Sq0n0m

jzt:
BTW, I was sad that there doesn't seem to be a built-in for deep-copying a struct. Maybe there's a good reason and I'm not thinking that far into it.

caglar10ur:
LGTM :sheep: :it:

*thanks for the playground explanation

hickeng:
@jzt Hey Zach - where in the code was the actual changing of the command happening? While I'm not a huge fan of returning structures with a rule of "You must not modify this value" I'm also not a fan of speculatively copying all of the images each time they're returned.

jzt:
@hickeng https://github.com/vmware/vic/blob/master/lib/apiservers/engine/backends/container.go#L250

The cache could contain `metadata.ImageConfig` rather than `*metadata.ImageConfig`, but the problem still remains:  https://play.golang.org/p/5RBSuBdMmi

We could have that code linked above deal with making the copy, but that requires your first situation of having to enforce the "do not modify" rule everywhere the cache is used. I don't know of a better way to do this than making the cache give copies as a form of self-preservation, considering the cache is supposed to be the source of truth.

hickeng:
@jzt yup. I see you found a better way of doing the copy - still not a deep copy, but hopefully enough.
It is viable to have that sort of rule - Go maps have it for concurrent access - but it's easy to have mistakes occur.

jzt:
@hickeng yeah, it works since there are no deeper pointers in the structure, but it may become an issue if we unpin from 1.11.0 and docker has changed things.

PR #1324, opened 2016-07-01
Initial ps implementation

### LINKED ISSUE #796: Milestone: docker ps
Implement docker ps command.
This should support:
```
-a
-q
```

cgtexmex:
Implements docker ps and portlayer container list.  This is done without caching, so a call to the infrastructure is required for each request.  

Fixes #796 #369

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -607,7 +607,35 @@ func (c *Container) ContainerTop(name string, psArgs string) (*types.ContainerPr
 
 // Containers returns the list of containers to show given the user's filtering.
 func (c *Container) Containers(config *types.ContainerListOptions) ([]*types.Container, error) {
-	return nil, fmt.Errorf("%s does not implement container.Containers", ProductName())
+
+	// Get an API client to the portlayer
+	portLayerClient := PortLayerClient()
+	if portLayerClient == nil {
+		return nil, derr.NewErrorWithStatusCode(fmt.Errorf("container.Containers failed to create a portlayer client"),
+			http.StatusInternalServerError)
+	}
+
+	containme, err := portLayerClient.Containers.GetContainerList(containers.NewGetContainerListParams().WithAll(&config.All))
+	if err != nil {
+		return nil, fmt.Errorf("Error invoking GetContainerList: %s", err.Error())
+	}
+	// TODO: move to conversion function
+	containers := make([]*types.Container, 0, len(containme.Payload))
+	for _, t := range containme.Payload {
+		cmd := strings.Join(t.ExecArgs, " ")
+		c := &types.Container{
+			ID:      *t.ContainerID,
+			Image:   *t.RepoName,
+			Created: *t.Created,
+			Status:  *t.Status,
+			Names:   t.Names,
+			Command: cmd,
+			SizeRw:  *t.StorageSize,
+		}
+		containers = append(containers, c)
+	}
+
+	return containers, nil
 }
 
 // docker's container.attachBackend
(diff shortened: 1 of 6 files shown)


lib/portlayer/exec/container.go
```diff
@@ -244,3 +250,77 @@ func (c *Container) Remove(ctx context.Context) error {
...
+		container := &Container{ExecConfig: &metadata.ExecutorConfig{}}
+		source := extraconfig.OptionValueSource(vms[i].Config.ExtraConfig)
+		extraconfig.Decode(source, container.ExecConfig)
+
+		// set state & friendly status
+		if vms[i].Runtime.PowerState == types.VirtualMachinePowerStatePoweredOn {
+			container.State = StateRunning
+			container.Status = "running"
+		} else {
+			container.State = StateRunning
```
jzt: Asking to learn: is `StateRunning` still valid for a container with status `stopped`?

jzt:
LGTM 👍

lib/apiservers/engine/backends/container.go
```diff
@@ -607,7 +607,35 @@ func (c *Container) ContainerTop(name string, psArgs string) (*types.ContainerPr
...
+	// Get an API client to the portlayer
+	portLayerClient := PortLayerClient()
+	if portLayerClient == nil {
+		return nil, derr.NewErrorWithStatusCode(fmt.Errorf("container.Containers failed to create a portlayer client"),
+			http.StatusInternalServerError)
+	}
+
+	containme, err := portLayerClient.Containers.GetContainerList(containers.NewGetContainerListParams().WithAll(&config.All))
+	if err != nil {
+		return nil, fmt.Errorf("Containers Error: %s", err.Error())
```
corrieb: I'd prefer something like "Error invoking GetContainerList: %s"

lib/portlayer/exec/container.go
cgtexmex: No..that needs to be changed....

lib/apiservers/engine/backends/container.go
```diff
@@ -668,6 +696,10 @@ func (c *Container) dockerContainerCreateParamsToPortlayer(cc types.ContainerCre
 	config.Image = new(string)
 	*config.Image = layerID
 
+	// Repo Requested
+	config.RepoName = new(string)
```
corrieb: In this line - and in line 696 - I don't understand why we're setting config.Image to a pointer to an empty string and then resetting it to a different string pointer in the line below. Isn't the empty string assignment unnecessary? Or am I missing something?

lib/portlayer/exec/config.go
```diff
@@ -38,9 +38,13 @@ type Configuration struct {
 
 	// Resource pools under which all containers will be created
 	ComputeResources []types.ManagedObjectReference `vic:"0.1" scope:"read-only"`
+
 	// Resource pool is the working version of the compute resource config
 	ResourcePool *object.ResourcePool
 
+	// uuid of portlayer vm
+	PortlayerUUID string
```
corrieb: I'm not clear what purpose this serves. I would imagine everything in this struct will be vsphere-specific, immutable and at the scope of the VCH. Is this so we can distinguish between container VMs created by different VCHs? Is the resource pool they're in not a reliable enough filter?

lib/portlayer/exec/container.go
corrieb: Shouldn't this be StateStopped?

corrieb:
Good to see more than one of us caught the typo. Other than that, seems fine for the purpose it serves.
LGTM

lib/portlayer/exec/config.go
cgtexmex: This was added to help in filtering out the VCH appliance from the ps list.  I'm going to approach that a different way.  So it will be removed.

mdubya66:
LGTM

hickeng:
LGTM

cgtexmex:
rebased

PR #1332, opened 2016-07-01
Convert between vsphere datastore format to URLs and back

fdawg4l:
Added helpers to convert between the two formats.

### DIFF
--- a/lib/portlayer/storage/vsphere/datastore.go
+++ b/lib/portlayer/storage/vsphere/datastore.go
@@ -15,9 +15,11 @@
 package vsphere
 
 import (
+	"fmt"
 	"io"
 	"net/url"
 	"path"
+	"regexp"
 	"strings"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


lib/portlayer/storage/vsphere/datastore.go
```diff
@@ -15,9 +15,11 @@
 package vsphere
 
 import (
+	"fmt"
 	"io"
 	"net/url"
 	"path"
+	"regexp"
```
caglar10ur: https://imgs.xkcd.com/comics/perl_problems.png :)

caglar10ur:
LGTM

lib/portlayer/storage/vsphere/datastore.go
fdawg4l: @mavery helped me with the expression. Also, we know a Perl celebrity. True story.

lib/portlayer/storage/vsphere/datastore.go
```diff
@@ -301,3 +303,36 @@ func (d *Datastore) mkRootDir(ctx context.Context, rootdir string) error {
...
+
+// Converts `[datastore] /path` to URL
+func DatastoreToURL(ds string) (*url.URL, error) {
+	matches := datastoreFormat.FindAllStringSubmatch(ds, -1)
+
+	if len(matches) != 1 || len(matches[0]) != 3 {
+		return nil, fmt.Errorf("invalid datastore format: %s", ds)
+	}
+
+	datastoreName := matches[0][1]
```
hmahmood: Is empty datastore name allowed? Your regexp allows that ('*' vs. '+').

hmahmood: Also, spaces allowed in datastore path? Can you omit it altogether? Your regex allows those cases too.

lib/portlayer/storage/vsphere/datastore.go
```diff
@@ -301,3 +303,36 @@ func (d *Datastore) mkRootDir(ctx context.Context, rootdir string) error {
 func (d *Datastore) rootDir() string {
 	return strings.SplitN(d.RootURL, " ", 2)[1]
 }
+
+// Parse the datastore format ([datastore1] /path/to/thing) to groups.
+var datastoreFormat = regexp.MustCompile(`\[([\w\d]*)\]\ (.*)?`)
```
hickeng: Is this deliberately allowing `[] /path/to/thing` and `[]`?

`(.*)?` - ? is redundant as .* will greedily consume everything up to the end of the line so it's not possible to have more than one group here.

lib/portlayer/storage/vsphere/datastore_test.go
```diff
@@ -127,6 +127,36 @@ func TestDatastoreMkdirAndLs(t *testing.T) {
...
 }
 
+func TestDatastoreToURLParsing(t *testing.T) {
+	expected := "ds://datastore1/path/to/thing"
+
+	input := []string{
+		"[datastore1] /path/to/thing",
+		"[datastore1] path/to/thing",
+		"[datastore1] ///path////to/thing",
+	}
```
hickeng: Add the following at a minimum if intended by expression:
```
"[]",
"[] /path/to/thing",

lib/portlayer/storage/vsphere/datastore.go
fdawg4l: You're regex is stronger than mine.  Lets chat offline and you can help me craft something smarter.

fdawg4l:
Going to drop this and allow @gigawhitlocks to grab this, fix it, and use it in his vic-machine changes since it'd be really useful there and i'm out of cycles to drive this to completion.  Thanks Ian!

PR #1353, opened 2016-07-06
[integration test] Add datastore parameter

caglar10ur:
also introduce timeout parameter for the vic-machine


caglar10ur:
cc @mhagen-vmware as we need to change secrets.yml as well

emlin:
lgtm

andrewtchin:
lgtm

caglar10ur:
♻️

PR #1362, opened 2016-07-06
Add UT for install/management

emlin:
The test is running on VC simulator, but there are some functions missing or inconsistent with real vSphere env, so some cases are comment off.

Also a few changes in production code, to make it easy to test.


andrewtchin:
lgtm

dougm:
lgtm

I can spend some time on getting the simulator to behave as expected in the cases where you needed to comment out tests.

emlin:
♻️

PR #1388, opened 2016-07-07
Only attach if container is running

### LINKED ISSUE #1254: attach should check vm state
Otherwise portlayer attempts to get ssh session and gets stuck

sflxn:
Check if container is running before attempting an attach.  Also update container inspect to use container running state.

Fixes #1254.  Helps #1129.


lib/apiservers/engine/backends/container.go
```diff
@@ -824,6 +826,37 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+	}
+
+	all := new(bool)
+	*all = true
+	containList, err := portLayerClient.Containers.GetContainerList(containers.NewGetContainerListParams().WithAll(all))
+	if err != nil {
+		return false, derr.NewErrorWithStatusCode(fmt.Errorf("Failed to get container list: %s", err), http.StatusInternalServerError)
+	}
+
+	for _, t := range containList.Payload {
```
caglar10ur: This is not going to scale with number of containers at all, this is O(n) for no good reason. Why are we asking all and then iterating over them one by one? Can't we ask the individual container to portlayer? If not sounds like a design problem to me...

lib/apiservers/engine/backends/container.go
```diff
@@ -671,13 +673,13 @@ func (c *Container) ContainerAttach(prefixOrName string, ca *backend.ContainerAt
...
 
 	if vc == nil {
-		//FIXME: If we didn't find in the cache, we should goto the port layer and
-		//see if it exists there.  API server might have been bounced.  For now,
-		//just return error
-
 		return derr.NewRequestNotFoundError(fmt.Errorf("No such container: %s", prefixOrName))
 	}
 
+	if running, err := IsContainerRunning(vc.ContainerID); !running {
```
fdawg4l: This seems ripe for racing.  This is a test/set problem.  You might want to at a minimum do this in a loop for some duration of time (`>= ` time to start a container) before deciding if the container is indeed not running.

Is there not a subscription mechanism we can use here instead of polling?  That could be a later optimization but would be the preferable implementation.

lib/apiservers/engine/backends/container.go
sflxn: No, it will not scale.  This is a stopgap solution.  Until we have the portlayer support for container inspect, there is no other way to obtain container status for a specific container.  Line 844 has a comment to replace this implementation once we have that support.

lib/apiservers/engine/backends/container.go
sflxn: There will be.  That will be the event monitoring in the portlayer that will be forthcoming.  Clint is adding that for ps and will propogate it to inspect, and then we'll use it.

The race condition will still persist even with the event monitoring.  This is just a sanity check so we don't go try to attach to a container that hasn't started.  From the user perspective, they will get a message that it is stopped and to try and start it before attaching.  They will simply check status and retry again.  If this ends up being a field problem for us, we can always tie it more tightly to the forthcoming event subscription.

lib/apiservers/engine/backends/container.go
```diff
@@ -824,6 +826,37 @@ func (c *Container) imageExist(imageID string) (storeName string, err error) {
...
+		return false, derr.NewErrorWithStatusCode(fmt.Errorf("Failed to get container list: %s", err), http.StatusInternalServerError)
+	}
+
+	for _, t := range containList.Payload {
+		if *t.ContainerID == containerID {
+			if *t.Status == "Running" {
+				return true, nil
+			}
+
+			return false, nil
```
sflxn: Hmm... Good point.  Let me look at the docker sources to see what errors it expects.

sflxn: Now returning the appropriate error in ContainerAttach() if container is not running.

lib/apiservers/engine/backends/container.go
```diff
@@ -602,10 +602,12 @@ func (c *Container) ContainerInspect(name string, size bool, version version.Ver
 	// FIXME: Just set a bunch of dummy info needed to get docker attach working.
 	// Once we have event stream working (part of docker ps work), we need to go
 	// back and replace this.
+	running, _ := IsContainerRunning(vc.ContainerID)
```
fdawg4l: Why are we doing this twice now?  See line 693.

sflxn: Two different API calls.  Sure, the CLI will call both inspect and attach during an attach, but the API doesn't know that.  We could reuse the value in the cache in ContainerAttach, but we can't know when inspect was last called.  We can't trust that value.  We need to check the portlayer again.

lib/apiservers/engine/backends/container.go
hickeng: What does this do in the case of `attach` followed by `start` as is seen with run and start -ai ?

sflxn:
Updated with the latest merged code from Clint that implements container info in the portlayer.  Now, we're using that to check for container status instead of the ps infrastructure.

caglar10ur:
LGTM :sheep: :it:

hmahmood:
lgtm

lib/apiservers/engine/backends/container.go
```diff
@@ -606,10 +606,12 @@ func (c *Container) ContainerInspect(name string, size bool, version version.Ver
...
+
 	base := &types.ContainerJSONBase{
-		State: &types.ContainerState{Status: "running",
-			Running: true,
-			Paused:  false,
+		State: &types.ContainerState{
+			Running: running,
+			Paused:  !running,
```
hickeng: We don't support pause - if it's not running then it's stopped, not paused.

caglar10ur: Hmmm so we are not going to suspend/resume?

hickeng: Not at the moment - whether we do in the future or not is a deferred question.

sflxn: Updated.  Set Paused to false and added comment.

hickeng:
LGTM

sflxn:
just clicked the update branch and github asking for approvals again.

PR #1391, opened 2016-07-07
Get container count for docker info

### LINKED ISSUE #1389: Docker Info should get container count
Docker info should use the docker ps infrastructure to get container count info.

sflxn:
Use the docker ps infrastructure to retrieve container
count data for docker info.

Resolves #1389.

### DIFF
--- a/lib/apiservers/engine/backends/system.go
+++ b/lib/apiservers/engine/backends/system.go
@@ -17,6 +17,7 @@ package vicbackends
 import (
 	"bufio"
 	"fmt"
+	"net/http"
 	"os"
 	"regexp"
 	"runtime"
@@ -29,9 +30,11 @@ import (
 
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client"
+	"github.com/vmware/vic/lib/apiservers/portlayer/client/containers"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/misc"
 	"github.com/vmware/vic/pkg/trace"
 
+	derr "github.com/docker/docker/errors"
 	"github.com/docker/docker/pkg/parsers/kernel"
 	"github.com/docker/docker/pkg/platform"
 	"github.com/docker/docker/pkg/system"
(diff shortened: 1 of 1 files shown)


lib/apiservers/engine/backends/system.go
```diff
@@ -262,6 +265,25 @@ func getImageCount(plClient *client.PortLayer) int {
...
 }
 
+// Use the Portlayer's support for docker ps to get the container count
 func getContainerStatus(plClient *client.PortLayer) (ContainerStatus, error) {
-	return ContainerStatus{0, 0, 0, 0}, nil
+	var status ContainerStatus
+
+	all := new(bool)
+	*all = true
+	containList, err := plClient.Containers.GetContainerList(containers.NewGetContainerListParams().WithAll(all))
```
jzt: Minor nit: Any reason we can't go the extra inch and call this `containerList`? :)

lib/apiservers/engine/backends/system.go
```diff
@@ -262,6 +265,25 @@ func getImageCount(plClient *client.PortLayer) int {
...
+		return status, derr.NewErrorWithStatusCode(fmt.Errorf("Failed to get container list: %s", err), http.StatusInternalServerError)
+	}
+
+	for _, t := range containList.Payload {
+		if *t.Status == "Running" {
+			status.numRunning++
+		} else if *t.Status == "Stopped" {
+			status.numStopped++
+		}
+		status.count++
```
jzt: Do we care about paused containers? We'd be counting them as well here, although I'm not sure we care about them right now. `numPaused` in the `ContainerStatus` struct seems to suggest so.

jzt:
LGTM

hmahmood:
lgtm

lib/apiservers/engine/backends/system.go
sflxn: yeah, it appears we're not supporting paused container for now.

lib/apiservers/engine/backends/system.go
sflxn: lol.  I didn't even notice that.  That's so weird.  Why did I do that?  lol

sflxn:
poke

Issue #1301, opened 2016-06-30
URL <-> vSphere path identifier translator


gigawhitlocks:
Need to add a routine to convert between the vSphere standard path identifier for e.g. specifying a Datastore + folder path for storage and the one we use internally, `ds://datastore_name/path/to/folder`

gigawhitlocks:
This issue was addressed in #1410

PR #1410, opened 2016-07-07
Adds volume store deletion to vic-machine delete

gigawhitlocks:
Does what it says on the tin -- vic-machine will now remove volume stores upon VCH destruction.

If the user does not specify `--force`, the volume stores used by the VCH are printed for their reference and optional manual removal.

Pretty straightforward and self-contained.

### DIFF
--- a/lib/install/management/delete.go
+++ b/lib/install/management/delete.go
@@ -15,10 +15,11 @@
 package management
 
 import (
+	"bytes"
+	"fmt"
 	"strings"
 
 	log "github.com/Sirupsen/logrus"
-
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/lib/metadata"
@@ -85,6 +86,12 @@ func (d *Dispatcher) DeleteVCH(conf *metadata.VirtualContainerHostConfigSpec) er
 	if err = d.destroyResourcePoolIfEmpty(conf); err != nil {
 		log.Warnf("VCH resource pool is not removed, %s", err)
 	}
+
+	log.Infoln("Removing volume stores...")
+	if err = d.deleteVolumeStoreIfForced(conf); err != nil {
+		log.Warnf("Error while deleting volume store: %s", err)
+		return err
+	}
 	return nil
 }
 
@@ -140,6 +147,46 @@ func (d *Dispatcher) DeleteVCHInstances(vmm *vm.VirtualMachine, conf *metadata.V
 		return errors.New(strings.Join(errs, "\n"))
 	}
 
+	return nil
+}
+
+func (d *Dispatcher) deleteVolumeStoreIfForced(conf *metadata.VirtualContainerHostConfigSpec) error {
+	if d.force {
+
+		for label, url := range conf.VolumeLocations {
+
+			// separate the host from the path in the provided path URL
+			pathAry := strings.SplitN(url.Path, " ", 2)
+			if len(pathAry) != 2 {
+				return errors.New("Didn't receive an expected volume store path format")
+			}
+
+			// convert the URL path to vSphere style pathing, and omit everything from the path except the root of the volume store
+			vSpherePath := fmt.Sprintf("%s %s", pathAry[0], strings.Split(pathAry[1], "/")[0])
+
+			// connect to vSphere and do the actual deletion
+			m := object.NewFileManager(d.session.Vim25())
+			log.Infof("Deleting volume store %s at path %s", label, vSpherePath)
+			task, err := m.DeleteDatastoreFile(d.ctx, vSpherePath, d.session.Datacenter)
+			if err != nil {
+				return errors.Errorf("Failed to start delete of %s due to error: %s", vSpherePath, err)
+			}
+
+			if err = task.Wait(d.ctx); err != nil {
+				return errors.Errorf("Failed to finish delete of %s due to error: %s", vSpherePath, err)
+			}
+
+		}
+
+	} else {
+		volumeStores := new(bytes.Buffer)
+		for label, url := range conf.VolumeLocations {
+			if _, err := volumeStores.WriteString(fmt.Sprintf("\t%s: %s\n", label, url.Path)); err != nil {
+				return err
+			}
+		}
+		log.Warnf("Since --force was not specified, the following volume stores will not be removed. Use the vSphere UI to delete content you do not wish to keep.\n%s", volumeStores.String())
+	}
 	return nil
 }
 
--- a/lib/install/management/delete.go
+++ b/lib/install/management/delete.go
@@ -55,6 +55,8 @@ func (d *Dispatcher) DeleteVCH(conf *metadata.VirtualContainerHostConfigSpec) er
 		errs = append(errs, err.Error())
 	}
 
+	d.deleteVolumeStoreIfForced(conf) // logs errors but doesn't ever bail out if it has an issue
+
 	if err = d.deleteNetworkDevices(vmm, conf); err != nil {
 		errs = append(errs, err.Error())
 	}
@@ -84,10 +86,6 @@ func (d *Dispatcher) DeleteVCH(conf *metadata.VirtualContainerHostConfigSpec) er
 	if err = d.destroyResourcePoolIfEmpty(conf); err != nil {
 		log.Warnf("VCH resource pool is not removed, %s", err)
 	}
-
-	log.Infoln("Removing volume stores...")
-	d.deleteVolumeStoreIfForced(conf) // logs errors but doesn't ever bail out if it has an issue
-
 	return nil
 }
 
--- a/lib/install/management/delete.go
+++ b/lib/install/management/delete.go
@@ -15,8 +15,6 @@
 package management
 
 import (
-	"bytes"
-	"fmt"
 	"strings"
 
 	log "github.com/Sirupsen/logrus"
@@ -88,10 +86,8 @@ func (d *Dispatcher) DeleteVCH(conf *metadata.VirtualContainerHostConfigSpec) er
 	}
 
 	log.Infoln("Removing volume stores...")
-	if err = d.deleteVolumeStoreIfForced(conf); err != nil {
-		log.Warnf("Error while deleting volume store: %s", err)
-		return err
-	}
+	d.deleteVolumeStoreIfForced(conf) // logs errors but doesn't ever bail out if it has an issue
+
 	return nil
 }
 
@@ -147,46 +143,6 @@ func (d *Dispatcher) DeleteVCHInstances(vmm *vm.VirtualMachine, conf *metadata.V
 		return errors.New(strings.Join(errs, "\n"))
 	}
 
-	return nil
-}
-
-func (d *Dispatcher) deleteVolumeStoreIfForced(conf *metadata.VirtualContainerHostConfigSpec) error {
-	if d.force {
-
-		for label, url := range conf.VolumeLocations {
-
-			// separate the host from the path in the provided path URL
-			pathAry := strings.SplitN(url.Path, " ", 2)
-			if len(pathAry) != 2 {
-				return errors.New("Didn't receive an expected volume store path format")
-			}
-
-			// convert the URL path to vSphere style pathing, and omit everything from the path except the root of the volume store
-			vSpherePath := fmt.Sprintf("%s %s", pathAry[0], strings.Split(pathAry[1], "/")[0])
-
-			// connect to vSphere and do the actual deletion
-			m := object.NewFileManager(d.session.Vim25())
-			log.Infof("Deleting volume store %s at path %s", label, vSpherePath)
-			task, err := m.DeleteDatastoreFile(d.ctx, vSpherePath, d.session.Datacenter)
-			if err != nil {
-				return errors.Errorf("Failed to start delete of %s due to error: %s", vSpherePath, err)
-			}
-
-			if err = task.Wait(d.ctx); err != nil {
-				return errors.Errorf("Failed to finish delete of %s due to error: %s", vSpherePath, err)
-			}
-
-		}
-
-	} else {
-		volumeStores := new(bytes.Buffer)
-		for label, url := range conf.VolumeLocations {
-			if _, err := volumeStores.WriteString(fmt.Sprintf("\t%s: %s\n", label, url.Path)); err != nil {
-				return err
-			}
-		}
-		log.Warnf("Since --force was not specified, the following volume stores will not be removed. Use the vSphere UI to delete content you do not wish to keep.\n%s", volumeStores.String())
-	}
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


lib/install/management/delete.go
```diff
@@ -143,6 +150,46 @@ func (d *Dispatcher) DeleteVCHInstances(vmm *vm.VirtualMachine, conf *metadata.V
...
+			pathAry := strings.SplitN(url.Path, " ", 2)
+			if len(pathAry) != 2 {
+				return errors.New("Didn't receive an expected volume store path format")
+			}
+
+			// convert the URL path to vSphere style pathing, and omit everything from the path except the root of the volume store
+			vSpherePath := fmt.Sprintf("%s %s", pathAry[0], strings.Split(pathAry[1], "/")[0])
+
+			// connect to vSphere and do the actual deletion
+			m := object.NewFileManager(d.session.Vim25())
```
emlin: There is method in store_files, deleteDatastoreFiles, which will handle file not found scenario, and return correct error message.
Please change to call that method here.

lib/install/management/delete.go
```diff
@@ -143,6 +150,46 @@ func (d *Dispatcher) DeleteVCHInstances(vmm *vm.VirtualMachine, conf *metadata.V
 	return nil
 }
 
+func (d *Dispatcher) deleteVolumeStoreIfForced(conf *metadata.VirtualContainerHostConfigSpec) error {
```
emlin: Also please move this method to store_files.go.

And, there is logic to delete volumes directory previously, which is used before you add volume creation related logic into vic-machine.
It would be appreciated if you can clean up those logic.

lib/install/management/delete.go
gigawhitlocks: I hate it when this happens, ha, wish I'd seen this method

lib/install/management/store_files.go
```diff
@@ -189,3 +191,27 @@ func (d *Dispatcher) createVolumeStores(conf *metadata.VirtualContainerHostConfi
...
+	if d.force {
+		for label, url := range conf.VolumeLocations {
+			// separate the host (datastore name) from the path in the provided path URL
+			pathComponents := strings.SplitN(url.Path, " ", 2)
+			if len(pathComponents) != 2 {
+				log.Warnf("Didn't receive an expected volume store path format: %s", url.Path)
+			}
+
+			log.Infof("Deleting volume store %s on Datastore %s at path %s", label, pathComponents[0], pathComponents[1])
+			if _, err := d.deleteDatastoreFiles(d.session.Datastore, pathComponents[1], d.force); err != nil {
```
fdawg4l: Without having seen the definition for `deleteDatastoreFiles`, it looks like you're using the datastore passed in via the session.  Specifically `d.session.Datastore`.  This means if there is more than 1 datastore, you may be attempting to delete files from the wrong one, right?

gigawhitlocks: Oh, you're right. I need to use finder on each datacenter from the paths in conf.. good catch Faiyaz

gigawhitlocks: Datastore I mean. Apparently I can't edit comments on mobile github

emlin:
@gigawhitlocks the old deleteVolumes method should be removed

gigawhitlocks:
@emlin is there no scenario in which a user might want to keep a volume store but delete one or more volumes? I realize that's a very similar method.. if it's not being used anywhere though I agree that it should be stripped out

gigawhitlocks:
Also I'm going to replace some of my `split` calls by merging in and cleaning up #1332 into this PR before I get ready to merge to master

emlin:
> is there no scenario in which a user might want to keep a volume store but delete one or more volumes? I realize that's a very similar method.. if it's not being used anywhere though I agree that it should be stripped out

I think after you add volume delete, that method is not useful anymore, cause that's assume the volumes is in the same datastore with image datastore.
But for what you said, to keep some volumes, but delete others. In the future, if we want, we can add them. Right now, user can rely on docker enginer api to delete volumes.

emlin:
> Also I'm going to replace some of my split calls by merging in and cleaning up #1332 into this PR before I get ready to merge to master

nice

gigawhitlocks:
Sounds good @emlin

lib/install/management/store_files.go
```diff
@@ -189,3 +191,44 @@ func (d *Dispatcher) createVolumeStores(conf *metadata.VirtualContainerHostConfi
 	}
 	return nil
 }
+
+func (d *Dispatcher) deleteVolumeStoreIfForced(conf *metadata.VirtualContainerHostConfigSpec) {
+	if d.force {
```
hickeng: Go idom is to address the short path first and return. This avoids large swathes of indented code.

lib/install/management/store_files.go
```diff
@@ -189,3 +191,44 @@ func (d *Dispatcher) createVolumeStores(conf *metadata.VirtualContainerHostConfi
...
+		for label, url := range conf.VolumeLocations {
+			// separate the host (datastore name) from the path in the provided path URL
+			pathComponents := strings.SplitN(url.Path, " ", 2)
+			if len(pathComponents) != 2 {
+				log.Warnf("Didn't receive an expected volume store path format: %s", url.Path)
+			}
+			log.Infof("Deleting volume store %s on Datastore %s at path %s", label, pathComponents[0], pathComponents[1])
+
+			// pathComponents[0] holds the datastore in format [foo] where foo is the name of the datastore. We need to pass just foo to DatastoreList
+			datastores, err := d.session.Finder.DatastoreList(d.ctx, pathComponents[0][1:len(pathComponents[0])-1])
```
hickeng: This isn't a fast path so let's be defensive:
```
ds := strings.TrimPrefix(pathComponents[0], "[")
ds = strings.TrimSuffix(pathComponents[0], "]")
datastores, err := d.session.Finder.DatastoreList(d.ctx, ds)
```

lib/install/management/store_files.go
```diff
@@ -189,3 +191,44 @@ func (d *Dispatcher) createVolumeStores(conf *metadata.VirtualContainerHostConfi
...
+				log.Errorf("Failed to delete volume store %s on Datastore %s at path %s", label, pathComponents[0], pathComponents[1])
+			}
+		}
+
+	} else { // user didn't specify --force so we're just going to print out some useful info for them
+		volumeStores := new(bytes.Buffer)
+		for label, url := range conf.VolumeLocations {
+			volumeStores.WriteString(fmt.Sprintf("\t%s: %s\n", label, url.Path))
+		}
+		log.Warnf("Since --force was not specified, the following volume stores will not be removed. Use the vSphere UI to delete content you do not wish to keep.\n%s", volumeStores.String())
```
hickeng: This message alone makes me think that we should have an explicit `-v` or similar argument to delete. Well, let's see what people think in onsite testing.

lib/install/management/store_files.go
```diff
@@ -189,3 +191,44 @@ func (d *Dispatcher) createVolumeStores(conf *metadata.VirtualContainerHostConfi
 	}
 	return nil
 }
+
+func (d *Dispatcher) deleteVolumeStoreIfForced(conf *metadata.VirtualContainerHostConfigSpec) {
+	if d.force {
+		for label, url := range conf.VolumeLocations {
+			// separate the host (datastore name) from the path in the provided path URL
+			pathComponents := strings.SplitN(url.Path, " ", 2)
```
hickeng: This format surprises me given our discussion earlier today - is this PR based off another that changes this or am I mis-remembering the current format?

lib/install/management/store_files.go
gigawhitlocks: Ah, good to know. And re: the rest of your comments, I'm refactoring that whole section to get rid of hackish `Split`s etc.

gigawhitlocks:
I pulled in Faiyaz's change, made some significant changes to that, namely included more restrictive regexes for parsing the Datastore format (`[datastoreFoo] foo/path/foo`) and replaced all of my hacky string manipulation with slightly-less-hacky string manipulation.

Note the FIXME -- we should be using one of my new helpers to encode the URL in `conf.VolumeLocations` but maybe that's not critical right now? I suspect also that during the `validate` step in vic-machine we could be using one of these helpers instead of some specific code that exists there now. I'll have to take a look into that but I'd like to declare that out of scope for this PR or the scope of this PR is just going to keep exploding.

lib/install/management/store_files.go
gigawhitlocks: Yeah I agree. It's unclear to me what the best usage pattern is, but what exists is definitely suboptimal.

gigawhitlocks:
With the addition of the translation code this PR actually also addresses #1301

lib/install/management/delete.go
```diff
@@ -85,6 +84,10 @@ func (d *Dispatcher) DeleteVCH(conf *metadata.VirtualContainerHostConfigSpec) er
 	if err = d.destroyResourcePoolIfEmpty(conf); err != nil {
 		log.Warnf("VCH resource pool is not removed, %s", err)
 	}
+
+	log.Infoln("Removing volume stores...")
```
emlin: You cannot remove volume after the appliance vm is removed. Cause we need the capability to rerun delete if anything wrong, and appliance vm keeps all the information for next run.

lib/install/management/store_files.go
```diff
@@ -189,3 +191,46 @@ func (d *Dispatcher) createVolumeStores(conf *metadata.VirtualContainerHostConfi
...
+		return
+	}
+
+	for label, url := range conf.VolumeLocations {
+		// FIXME: url is being encoded by the portlayer incorrectly, so we have to convert url.Path to the right url.URL object
+		log.Debugf("Provided datastore URL %s", url.Path)
+		dsURL, err := vsphere.DatastoreToURL(url.Path)
+		log.Debugf("Parsed volume store path %s", dsURL.Path)
+		if err != nil {
+			log.Warnf("Didn't receive an expected volume store path format: %s", url.Path)
```
emlin: even parse failed, you still want to try datastorelist here?

lib/install/management/store_files.go
```diff
@@ -189,3 +191,46 @@ func (d *Dispatcher) createVolumeStores(conf *metadata.VirtualContainerHostConfi
...
+		log.Infof("Deleting volume store %s on Datastore %s at path %s", label, dsURL.Host, dsURL.Path)
+
+		datastores, err := d.session.Finder.DatastoreList(d.ctx, dsURL.Host)
+
+		if err != nil {
+			log.Errorf("Error finding datastore %s: %s", dsURL.Host, err)
+			continue
+		}
+		if len(datastores) != 1 {
+			foundDatastores := new(bytes.Buffer)
```
emlin: I think you might use similar error message with the one in validator, for example, 
```
	if len(stores) > 1 {
		// TODO: error about required disabmiguation and list entries in nets
		log.Errorf("ambiguous datastore %s" + dsURL.Host)
	}
```

lib/install/management/store_files.go
```diff
@@ -189,3 +191,46 @@ func (d *Dispatcher) createVolumeStores(conf *metadata.VirtualContainerHostConfi
...
+			for _, d := range datastores {
+				foundDatastores.WriteString(fmt.Sprintf("\n%s\n", d))
```
emlin: if you prefer to print out all datastore names, might try strings.join()

emlin:
Also, you might want to add few cases in create_test and delete_test, they are using simulator. The datastore operations works well now.

lib/install/management/delete.go
gigawhitlocks: Yeah, I wasn't sure about this, I'll move it to earlier in the deletion routine.

lib/install/management/store_files.go
gigawhitlocks: Ah you're right, I need to add a `continue` so that the `for` loop continues to the other found volumes

gigawhitlocks:
Good to know re: tests & the vsim @emlin -- I'll take a look in those files especially since this can't be merged today anyway

lib/install/management/store_files.go
gigawhitlocks: I'll take a look at this method at least. Wasn't sure the best way to replicate the Java `StringBuilder` pattern in Go, which is essentially what I'm doing here.

lib/install/management/store_files.go
gigawhitlocks: Ironically I think I had that error message at one point and must've removed it..

lib/install/management/store_files.go
gigawhitlocks: Well, the slice returned by `DatastoreList` is a `[]object.Datastore*` so actually I found a bug here in that I was printing `d` and I wanted to print `d.InventoryPath` but in any case since I'm printing a member of an object and `datastores` is not a `[]string` I cannot use `strings.Join()`. Similar problem on line 202-204. 

There might be a slightly more concise way to express this, but the way I have it is straightforward, easy to understand, and fast.

lib/portlayer/storage/vsphere/datastore.go
```diff
@@ -301,3 +303,26 @@ func (d *Datastore) mkRootDir(ctx context.Context, rootdir string) error {
...
 }
+
+// Parse the datastore format ([datastore1] /path/to/thing) to groups.
+var datastoreFormat = regexp.MustCompile(`^\[([\w\d\(\)\s]+)\]`)
+var pathFormat = regexp.MustCompile(`\s[\/\w]+$`)
+
+// Converts `[datastore] /path` to URL
+func DatastoreToURL(ds string) (*url.URL, error) {
+	u := new(url.URL)
+	u.Host = strings.TrimSuffix(strings.TrimPrefix(datastoreFormat.FindString(ds), "["), "]")
```
fdawg4l: `Trim`?  Why can't we match on `[` and `]` and use the result?

fdawg4l: Reading further, I think you need the `Trim` because you're using `FindString`.  Why not use one of the `Match` calls which returns the result of the match.

emlin:
btw, don't forget to remove existing deleteVolumes method

lib/portlayer/storage/vsphere/datastore.go
gigawhitlocks: The `[]`s are part of the pattern because they indicate, of course, where the datastore name is. As far as I could tell (maybe I'm wrong), all of the Go regex methods include them as part of the match because they're part of the pattern. So I trim them off.

Same with the `\s` in `pathFormat`. Needs to be there, because that's what makes the rest of the pattern unique from `datastoreFormat`, but then must be trimmed.

gigawhitlocks: Yeah @fdawg4l I did a little experimenting -- using FindStringSubmatch also includes the square brackets: https://play.golang.org/p/6HB6sLucmA

fdawg4l:
lgtm

emlin:
lgtm

fdawg4l:
.

PR #1412, opened 2016-07-07
NAT forwarding in the appliance

hickeng:
This provides containers on the bridge network with a route to the wider
world via the appliance.

```
vagrant@devbox:~/vic$ docker -H 192.168.78.245:2376 --tls run -it busybox
a3ed95caeb02: Pull complete
8ddc19f16526: Pull complete
/ # ping www.google.com
PING www.google.com (172.217.3.196): 56 data bytes
64 bytes from 172.217.3.196: seq=0 ttl=127 time=22.380 ms
64 bytes from 172.217.3.196: seq=1 ttl=127 time=23.645 ms
```


isos/appliance/nat-setup
```diff
@@ -40,32 +43,29 @@ echo 2 > /proc/sys/net/ipv4/ip_dynaddr
...
-iptables -A OUTPUT -j ACCEPT
-
-# Permit answers on already established connections
-# and permit new connections related to established ones
-# (e.g. port mode ftp)
-iptables -A INPUT -m conntrack --ctstate ESTABLISHED,RELATED -j ACCEPT
+# wait for the bridge interface to exist - 100ms wait
+until ip link show dev bridge;do sleep 0.1;done
+
+iptables-restore <<RULES
```
hmahmood: How does this interact with the iptables service? As I recall, it runs the iptables script that comes photon (the one just renamed).

hickeng: It doesn't. It needs to run after vic-init has renamed the interfaces which means it cannot run at the same time as iptables service (which is part of network-online).
It seemed less prone to confusion to rename the script and invoke it from a specific later target than to change the iptables service to depend on vic.target.

caglar10ur:
need to drop pkt counters, other than that LGTM

fdawg4l:
LGTM

hickeng:
Looks like this has some issues running the nat-setup script currently. Don't know why it works in local testing.

hickeng:
poke

isos/appliance/nat-setup
hmahmood: The new setup is equally confusing to me, fwiw. I would prefer we keep the dependencies simple. You are waiting for the bridge interface to come up anyway. Did you consider running the rules from within vic-init? There is this: https://github.com/docker/libnetwork/blob/master/iptables/iptables.go .

caglar10ur:
👍 using https://github.com/docker/libnetwork/blob/master/iptables/iptables.go

hickeng:
poke

PR #1414, opened 2016-07-08
Preserve extraconfig on stop

### LINKED ISSUE #1337: Stopping container removes extraconfig information
When issuing a stop command for a container VM the extraconfig information is removed from the VMX file.  This extraconfig information (guestinfo fields) is used by the portlayer and specifically the exec.List command to identify container VMs.  

Since this information is removed on container stop the exec.List is unable to identify container VMs and subsequently they do not appear in the docker ps output.  

Either some extraconfig information should be left upon container stop or we need to identify container vms via an alternative mechanism.

hickeng:
This preserves the extraconfig data on stop, and also removes the unnecessary host specification that restricts DRS placement when creating a VM.

Fixes #1337

### DIFF
--- a/lib/portlayer/exec/container.go
+++ b/lib/portlayer/exec/container.go
@@ -106,67 +106,74 @@ func (c *Container) Commit(ctx context.Context, sess *session.Session, h *Handle
 	c.Lock()
 	defer c.Unlock()
 
-	if h.Spec != nil {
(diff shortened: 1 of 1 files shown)


caglar10ur:
LGTM :sheep: :it:

cgtexmex:
LGTM

PR #1419, opened 2016-07-08
Fix pprof data collection

### LINKED ISSUE #1417: pprof files missing from log bundle
The bundle contains no pprof files (see https://ci.vmware.run/vmware/vic/2698 as an example).

This was visible via warning messages in the unit tests - however the way vicadmin is written is to be resilient to errors and to continue in any way possible so this doesn't easily present as a failure.

Fixing the lack may be as simple as adding "http://" scheme to address the unknown scheme errors and should be done asap (TP3), however this issue shouldn't be closed until there are:
a. an integration test that downloads the logs and checks the archive listing for the expected names.
b. ideally a unit test that will actually fail if there's a collection issue.

This can be moved out of TP3 once the collection is corrected.

kjplatz:
Fixes #1417

During the course of PR review, the pprof code got refactored several times which involved dropping the leading "http://" from the pprof endpoints.  This PR addresses that problem.

### DIFF
--- a/lib/pprof/pprof.go
+++ b/lib/pprof/pprof.go
@@ -70,7 +70,7 @@ func GetPprofEndpoint(component PprofPort) *url.URL {
 		}
 	}
 
-	endpoint, err := url.Parse(fmt.Sprintf("%s:%d", ip, port))
+	endpoint, err := url.Parse(fmt.Sprintf("http://%s:%d", ip, port))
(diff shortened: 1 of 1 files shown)


andrewtchin:
lgtm

jzt:
lgtm

kjplatz:
Bing

PR #1307, opened 2016-07-01
Container inspect in the personality server

sflxn:
Add container inspect to the docker personality server.

resolves #370

Issue #370, opened 2016-03-31
Docker API: Docker Inspect Container


sflxn:
Implement the docker inspect operation (docker inspect from CLI) in the engine-api server and integrate with the interaction port layer.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#inspect-a-container
https://docs.docker.com/engine/reference/commandline/inspect/

Please note, the engine-api's rest server already packaging these json struct into golang structs.  The url above is just for reference on what the operation should do.

hickeng:
Portions of this are necessary for attach to work properly

mdubya66:
@sflxn is this being worked on under another issue?

sflxn:
It is. #1144.  I will close the other one as duplicate, as there is more info here.

sflxn:
We now have a swagger definition, and I just updated it. I've implemented a first rev of container inspect on the personality server side. Until the portlayer returns those swagger definitions I created, container inspect is still blocked. I'm going to create some dummy code on the portlayer side to test with the personality server while I wait for @cgtexmex to populate those data fields.

Related to #1276

sflxn:
I submitted a PR yesterday, but it shouldn't be merged until the portlayer side is done.  The PR will override the work that was done for the inspect for attach.  If this PR gets approved and merged without the portlayer side, attach will no longer work.

cgtexmex:
@sflxn do we have a portlayer story regarding the data to return or the API that will return it?  I implemented the GetContainerListHandler for ps, but am not aware of another issue for the remaining data needed for inspect...

PR #1420, opened 2016-07-08
ContainerInfo API

cgtexmex:
This provides the initial implementation of container info and will return the container information available to the portlayer.  

Much like the current implementation of containerlistinfo, this API will query the infrastructure for accurate container state / attribute information.

Fixes # 1393

### DIFF
--- a/lib/apiservers/portlayer/swagger.yml
+++ b/lib/apiservers/portlayer/swagger.yml
@@ -1082,12 +1082,14 @@ definitions:
     properties:
       containerId:
         type: string
-      imageId:
+      layerId:
+        type: string
+      repoName:
         type: string
       created:
-        type: string
-        format: datetime
-      name:
+        type: integer
+        format: int64
+      names:
         type: array
         items:
           type: string
@@ -1147,15 +1149,18 @@ definitions:
     properties:
       pid:
         type: integer
-        format: int32
       execPath:
         type: string
       execArgs:
-        type: string
+        type: array
+        items:
+          type: string
       workingDir:
         type: string
       env:
-        type: string
+        type: array
+        items:
+          type: string
       status:
         type: string
       started:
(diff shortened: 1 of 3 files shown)


jzt:
LGTM

lib/apiservers/portlayer/swagger.yml
```diff
@@ -1082,12 +1082,14 @@ definitions:
...
+      layerId:
         type: string
-      created:
+      repoName:
         type: string
-        format: datetime
-      name:
+      created:
+        type: integer
+        format: int64
```
sflxn: What's the interpretation of this date value?  Or better yet, how do I turn it into a DateTime?

lib/portlayer/exec/container.go
```diff
@@ -250,6 +251,51 @@ func (c *Container) Remove(ctx context.Context) error {
 	return nil
 }
 
+// Grab the info for the requested container
+// TODO:  Possibly change so that hanlder requests a handle to the
```
caglar10ur: [cosmetic] s/hanlder/handler/g

caglar10ur:
LGTM :sheep: :it:

lib/apiservers/portlayer/swagger.yml
cgtexmex: That's the unix time -- it's placed in the vmx file on container vm creation.  Using time.Unix(n,n) you can covert to time.  i.e. https://play.golang.org/p/Dx2asGut5Y

sflxn:
Some of these changes to the swagger spec break PR #1307, but we're delaying issue #370 anyway so I can update that PR to reflect your swagger changes.

caglar10ur:
@sflxn I have #1307 working here on top of this as I'm testing docker-compose integration :)

cgtexmex:
rebase

PR #1424, opened 2016-07-09
Restart on unhandled signal

hickeng:
This triggers a component restart on unhandled signals as well as "clean" exits.
SIGTERM and SIGQUIT are both handled by Go so the waitstatus is Exited.

SIGTERM and others are unhandled so require us to check for Signalled as well.

Upping the default debug level as this was a pain to debug with Info and TP3 will have issue to diagnose.


dougm:
lgtm

caglar10ur:
LGTM :sheep: :it:

hickeng:
poke

hickeng:
poke

PR #1431, opened 2016-07-11
Guest tools framework improvements

dougm:
- Implement Capabilities_Register RPC

- Use the same polling interval and backoff logic as vmtoolsd

- Add support for guest power operations

- Trace disabled by default

- Add ChannelOut.Request wrapper

- Add ESX tests, additional unit tests

Closes issue #742


dougm:
Not required for TP3, will hold off on updates/merge, but ready for review.

pkg/vsphere/toolbox/service.go
```diff
@@ -59,6 +78,22 @@ func NewService(rpcIn Channel, rpcOut Channel) *Service {
 	return s
 }
 
+// backoff exponentially increases the RPC poll delay up to maxDelay
+func (s *Service) backoff() {
+	if s.delay < maxDelay {
+		if s.delay > 0 {
+			d := s.delay * 2
```
fdawg4l: Nit.  y = 2x is linear, not exponential.

fdawg4l:
LGTM.  

Out of curiosity, is the channel 2-way?  I.e. can the guest write to it and have the host read from it?  I ask because I'm baffled by the broadcast/backoff logic.  There must be a reason for doing it that way because otherwise it seems so error prone.

caglar10ur:
LGTM :sheep: :it:

dougm:
@fdawg4l In general, we are using the same logic as vmtoolsd, as of this PR.

broadcast?  You mean the Send(nil) "to let the VMX know we are here"?  The VMX simply won't send us any RPCs until that happens.  Note that there is still message metadata sent in that case, even though this buffer is 'nil'.

We could remove backoff() and it would still function the same.  We (and vmtoolsd) just start polling at a higher rate.  At startup for example, VMX will send a quick series of RPCs.  Once we've responded, there's a a less often set of ping/tell-me-your-ip-address/etc.  Other RPCs would be triggered by API calls, such as to start an in-guest command or graceful shutdown.

RPC requests are serialized in a queue on the VMX side, if we don't respond in time (which varies depending on the type of RPC), they are removed from the queue and it moves on to the next message in the queue.  With the longer polling interval, we were missing a call at startup (Capabilities_Register), that was never retried by the VMX.

There are 2 channels, TCLO (VMX request to guest) and RPCI (guest request to VMX).  The Service.Start loop handles incoming (TCLO) RPC requests and can only send a response to those RPCs.  If the guest wants to send an RPC request to the VMX, it uses the RPCI channel, and will receive a response to that RPC from the VMX.

pkg/vsphere/toolbox/service.go
dougm: will fix that

hickeng:
LGTM

dougm:
rebase

PR #1432, opened 2016-07-11
Add two more ESX servers

mhagen-vmware:
Add an additional 2 more ESX servers so that we can increase the number of drone workers eventually.


mdubya66:
Lgtm

andrewtchin:
lgtm

PR #1433, opened 2016-07-11
Add a deadline parameter to the attach streams in port layer

sflxn:
A deadline can now be passed into each attach stream request.  The
deadline is a hard set time instead of a timeout.  This allows the
caller to better manage total timeout for each attach.  Also make
a call to flush in WriteResponse() to ensure the header goes out
before the body does.

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/go-swagger/go-swagger/httpkit"
 	httptransport "github.com/go-swagger/go-swagger/httpkit/client"
+	strfmt "github.com/go-swagger/go-swagger/strfmt"
 	"github.com/mreiferson/go-httpclient"
 
 	log "github.com/Sirupsen/logrus"
@@ -66,10 +67,12 @@ type volumeFields struct {
 }
 
 const (
-	attachConnectTimeout time.Duration = 15 * time.Second //timeout for the connection
-	attachAttemptTimeout time.Duration = 30 * time.Second //timeout before we ditch an attach attempt
-	attachRequestTimeout time.Duration = 2 * time.Hour    //timeout to hold onto the attach connection
-	swaggerSubstringEOF                = "EOF"
+	attachConnectTimeout   time.Duration = 15 * time.Second //timeout for the connection
+	attachAttemptTimeout   time.Duration = 40 * time.Second //timeout before we ditch an attach attempt
+	attachPLAttemptDiff    time.Duration = 10 * time.Second
+	attachPLAttemptTimeout time.Duration = attachAttemptTimeout - attachPLAttemptDiff //timeout for the portlayer before ditching an attempt
+	attachRequestTimeout   time.Duration = 2 * time.Hour                              //timeout to hold onto the attach connection
+	swaggerSubstringEOF                  = "EOF"
 )
 
 // docker's container.execBackend
@@ -177,10 +180,9 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 	}
 
 	// provide basic container config via the image
-	container := &viccontainer.VicContainer{
-		ID:     image.ID,
-		Config: image.Config,
-	}
+	container := viccontainer.NewVicContainer()
+	container.ID = image.ID
+	container.Config = image.Config
 
 	// Overwrite or append the image's config from the CLI with the metadata from the image's
 	// layer metadata where appropriate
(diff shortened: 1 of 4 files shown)


lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -99,8 +99,24 @@ func (i *InteractionHandlersImpl) ContainerResizeHandler(params interaction.Cont
...
+	var ctxDeadline time.Time
+	var timeout time.Duration
+
+	// Calculate the timeout for the attach if the caller specified a deadline
+	if params.Deadline != nil {
+		ctxDeadline = time.Time(*params.Deadline)
+		timeout = ctxDeadline.Sub(time.Now())
+		log.Printf("Attempting to get ssh session for container %s stdin with deadline %s", params.ID, ctxDeadline.Format(time.UnixDate))
+		if timeout < 0 {
+			e := &models.Error{Message: fmt.Sprintf("Deadline for stdin already passed for container %s", params.ID)}
```
fdawg4l: You're calculating from wall clock time to derive a timeout from a deadline?   That sounds like a lot of work just to set a duration on a timeout.  Why not just pass/use a timeout duration vs a deadline?

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -118,25 +134,59 @@ func (i *InteractionHandlersImpl) ContainerSetStdinHandler(params interaction.Co
...
 func (i *InteractionHandlersImpl) ContainerGetStderrHandler(params interaction.ContainerGetStderrParams) middleware.Responder {
-	log.Printf("Attempting to get ssh session for container %s stderr", params.ID)
-	sshConn, err := i.attachServer.Get(context.Background(), params.ID, interactionTimeout)
+	var ctxDeadline time.Time
+	var timeout time.Duration
+
+	// Calculate the timeout for the attach if the caller specified a deadline
+	if params.Deadline != nil {
+		ctxDeadline = time.Time(*params.Deadline)
+		timeout = ctxDeadline.Sub(time.Now())
```
fdawg4l: And you're doing the inverse here.  This just requires a paramter to set the timeout.  Deadlines are nice if you need somethign done by a certain wall clock time but i don't see that requirement here.  Just pass a timeout.

Also, the timeout duration itself looks like it's coming from a constant.  So I'm kind of wondering why we need this change at all.  Am I missing something?

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
sflxn: The idea was that we wanted to start a "timer" from when a caller performs an attach on the personality server side.  That is total time we're going to allow, inclusive of the portlayer.  That includes the time it takes for the ssh server to finish.  Since this is total overall time, defined by the personality server side, we need the portlayer to also abide by that time so giving it a hard deadline is the only way to remove any other overhead.

fdawg4l: Where is the timer being used?

The arithmetic I'm seeing says you can do all of the things you're describing with a duration VS a deadline. 

Deferring to @hickeng. Sounds like you're both in favor of this approach but I fail to see why this is necessary.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
sflxn: It's a constant, but there can be other operations occurring before.  As from the explanation above, the total time allotted to the portlayer to timeout is within the total time the personality server define.  A deadline works better for this.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
sflxn: It wasn't originally my idea, but I really like it.  If you're trying to force a 2nd party to adhere to your timeouts, a deadline removes all timing uncertainties.

caglar10ur:
The code LGTM, not clear about logic as I'm not familiar with the attach part

fdawg4l:
Can this change wait until after #1440 goes in?

sflxn:
Sure

corrieb:
#1440 is merged. We need to get this in today. What more do we need in order to expedite?

lib/apiservers/engine/backends/container.go
```diff
@@ -913,7 +916,8 @@ func attachStreams(name string, tty, stdinOnce bool, clStdin io.ReadCloser, clSt
 		// Swagger wants an io.reader so give it the reader pipe.  Also, the swagger call
 		// to set the stdin is synchronous so we need to run in a goroutine
 		go func() {
-			setStdinParams := interaction.NewContainerSetStdinParamsWithContext(inContext).WithID(name)
+			var setStdinParams *interaction.ContainerSetStdinParams
+			setStdinParams = interaction.NewContainerSetStdinParamsWithContext(inContext).WithID(name)
```
hickeng: misleading variable name, `name`

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
hickeng: Obviously this is only viable when sharing a common time source. It's less prone to drift if there's a significant delay in between client side marshaling and server side receipt.

hickeng:
@corrieb this needs rebasing onto 1440 which should significantly reduce the amount of code as a single context is used.

hickeng:
LGTM
The deadline -> timeout code looks like it could do with a utility function and we should be merging the interaction timeout with the deadline but works for now.

PR #1434, opened 2016-07-11
Container inspect in the personality server

caglar10ur:
This is simply Loc's original work #1307 + adjustments required for the
ContainerInfo API.

Also removes the containers from the cache after a docker rm

Fixes #370

Requires #1427 and #1428

PR #1438, opened 2016-07-11
Fix container tty resize for docker attach by name

### LINKED ISSUE #1436: Docker API: docker resize fails if short name is used with docker attach
docker create -ps busybox /bin/top
docker start <id>
docker ps .... to look up short name (trusting_saha)
docker attach trusting_saha

Results:  it attaches but does not resize the tty window.

sflxn:
Fix resize issue when using docker attach on shorten name (e.g. frustrating_saha).

Resolves #1436.

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -317,6 +317,11 @@ func (c *Container) ContainerRename(oldName, newName string) error {
 func (c *Container) ContainerResize(name string, height, width int) error {
 	defer trace.End(trace.Begin("ContainerResize"))
 
+	vc := cache.ContainerCache().GetContainer(name)
+	if vc == nil {
+		return derr.NewRequestNotFoundError(fmt.Errorf("No such container: %s", name))
+	}
+
 	// Get an API client to the portlayer
 	client := PortLayerClient()
 	if client == nil {
@@ -327,7 +332,7 @@ func (c *Container) ContainerResize(name string, height, width int) error {
 	// Call the port layer to resize
 	plHeight := int32(height)
 	plWidth := int32(width)
-	plResizeParam := interaction.NewContainerResizeParams().WithID(name).WithHeight(plHeight).WithWidth(plWidth)
+	plResizeParam := interaction.NewContainerResizeParams().WithID(vc.ContainerID).WithHeight(plHeight).WithWidth(plWidth)
 
 	_, err := client.Interaction.ContainerResize(plResizeParam)
 	if err != nil {


lib/apiservers/engine/backends/container.go
```diff
@@ -858,6 +856,30 @@ func createNewAttachClientWithTimeouts(connectTimeout, responseTimeout, response
 	return plClient, transport
 }
 
+func IsContainerRunning(containerID string) (bool, error) {
```
fdawg4l: Doesn't look like you need the bool.  Just return an error and check if err != nil.

lib/apiservers/engine/backends/container.go
```diff
@@ -858,6 +856,30 @@ func createNewAttachClientWithTimeouts(connectTimeout, responseTimeout, response
...
+
+	ci := resp.Payload
+
+	if ci != nil && ci.ContainerConfig != nil {
+		if ci.ContainerConfig.State != nil && strings.ToLower(*ci.ContainerConfig.State) == "running" {
+			return true, nil
+		}
+	}
+
+	return false, nil
```
fdawg4l: Re: comment above, return an error in this case.

lib/apiservers/engine/backends/container.go
sflxn: this was part of another PR.   This branch was based on that branch.

lib/apiservers/engine/backends/container.go
sflxn: ditto above

sflxn:
just squashed the commits.

lib/apiservers/engine/backends/container.go
```diff
@@ -317,6 +317,11 @@ func (c *Container) ContainerRename(oldName, newName string) error {
 func (c *Container) ContainerResize(name string, height, width int) error {
 	defer trace.End(trace.Begin("ContainerResize"))
 
+	vc := cache.ContainerCache().GetContainer(name)
+	if vc == nil {
+		return derr.NewRequestNotFoundError(fmt.Errorf("No such container: %s", name))
```
caglar10ur: Do not return err here, as you check its existence later via PL. Instead just do
```
vc := cache.ContainerCache().GetContainer(name)
if vc != nil {
name = vic
}
```

lib/apiservers/engine/backends/container.go
```diff
@@ -327,7 +332,7 @@ func (c *Container) ContainerResize(name string, height, width int) error {
 	// Call the port layer to resize
 	plHeight := int32(height)
 	plWidth := int32(width)
-	plResizeParam := interaction.NewContainerResizeParams().WithID(name).WithHeight(plHeight).WithWidth(plWidth)
+	plResizeParam := interaction.NewContainerResizeParams().WithID(vc.ContainerID).WithHeight(plHeight).WithWidth(plWidth)
```
caglar10ur: No need to do this if you use the if above

caglar10ur:
Closing this in as I went over all the function on that file @https://github.com/vmware/vic/pull/1434

lib/apiservers/engine/backends/container.go
sflxn: Good point.

PR #1440, opened 2016-07-12
Fix attach panic in port layer server

### LINKED ISSUE #1130: port layer crash during attach error path
[container-logs.tar.gz](https://github.com/vmware/vic/files/320044/container-logs.tar.gz)

The port layer crashed during the workflow described in #1129 - the error message about deadline exceeded is the last thing we see from it.

fdawg4l:
When the client (the DPS in this case) goes away, there is nothing to write a body to from the "interaction" handlers.  In that case, set the error code and return.  If a client *is* infact still connection, the header is all it cares about anyway.  This fix is a workaround for the issue in WriteResponse() in the swagger generated server.

Also refactored attachStreams for readability and debugability.  There was too much redundancy and not enough logging or error path validation.

Fixes #1130

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -155,9 +155,9 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 
 	// bail early if container name already exists
 	if exists := cache.ContainerCache().GetContainer(config.Name); exists != nil {
-		return types.ContainerCreateResponse{},
-			derr.NewErrorWithStatusCode(fmt.Errorf("Conflict. The name \"%s\" is already in use by container %s. You have to remove (or rename) that container to be able to re use that name.", config.Name, exists.ContainerID),
-				http.StatusConflict)
+		err := fmt.Errorf("Conflict. The name \"%s\" is already in use by container %s. You have to remove (or rename) that container to be able to re use that name.", config.Name, exists.ContainerID)
+		log.Errorf("%s", err.Error())
+		return types.ContainerCreateResponse{}, derr.NewErrorWithStatusCode(err, http.StatusConflict)
 	}
 
 	// get the image from the cache
@@ -165,6 +165,7 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 	if err != nil {
 		// if no image found then error thrown and a pull
 		// will be initiated by the docker client
+		log.Errorf("ContainerCreate: image not found %s", config.Config.Image)
 		return types.ContainerCreateResponse{}, err
 	}
 
@@ -225,7 +226,9 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 	// transfer port layer swagger based response to Docker backend data structs and return to the REST front-end
 	if err != nil {
 		if _, ok := err.(*containers.CreateNotFound); ok {
-			return types.ContainerCreateResponse{}, derr.NewRequestNotFoundError(fmt.Errorf("No such image: %s", container.ID))
+			err = fmt.Errorf("No such image: %s", container.ID)
+			log.Errorf(err.Error())
+			return types.ContainerCreateResponse{}, derr.NewRequestNotFoundError(err)
 		}
 
 		// If we get here, most likely something went wrong with the port layer API server
@@ -246,6 +249,7 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 			}))
 
 		if err != nil {
+			log.Errorf("ContainerCreate: Scopes error: %s", err.Error())
 			return types.ContainerCreateResponse{}, derr.NewErrorWithStatusCode(err, http.StatusInternalServerError)
 		}
 
@@ -265,10 +269,12 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 	// commit the create op
 	_, err = client.Containers.Commit(containers.NewCommitParams().WithHandle(h))
 	if err != nil {
+		err = fmt.Errorf("No such image: %s", container.ID)
+		log.Errorf("%s", err.Error())
 		// FIXME: Containers.Commit returns more errors than it's swagger spec says.
 		// When no image exist, it also sends back non swagger errors.  We should fix
 		// this once Commit returns correct error codes.
-		return types.ContainerCreateResponse{}, derr.NewRequestNotFoundError(fmt.Errorf("No such image: %s", container.ID))
+		return types.ContainerCreateResponse{}, derr.NewRequestNotFoundError(err)
 	}
 
 	// Container created ok, overwrite the container params in the container store as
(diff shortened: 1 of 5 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -153,16 +152,17 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
...
+		log.Errorf("%s", err.Error())
+		return types.ContainerCreateResponse{}, derr.NewErrorWithStatusCode(err, http.StatusConflict)
 	}
 
 	// get the image from the cache
 	image, err := cache.ImageCache().GetImage(config.Config.Image)
 	if err != nil {
 		// if no image found then error thrown and a pull
 		// will be initiated by the docker client
+		log.Errorf("ContainerCreate: image not found %s", config.Config.Image)
```
hickeng: Is this the correct message to get the docker client to then pull the image?

fdawg4l: No idea. I just wanted to log the error since it was completely silent previously.

fdawg4l: I should probably actually print the error. :-). Oops.

lib/apiservers/engine/backends/container.go
```diff
@@ -882,198 +877,174 @@ func IsContainerRunning(containerID string) (bool, error) {
...
+				log.Errorf("container attach: stdin (%s): %s", vc.ContainerID, err.Error())
 			} else {
-				_, err = io.Copy(stdinWriter, clStdin)
+				log.Infof("container attach: stdin (%s) done: %s", vc.ContainerID)
 			}
 
-			if err != nil {
-				log.Errorf(err.Error())
-			}
+			cancel()
```
hickeng: Let's have a comment about the divergence from dockers behaviour and an issue to investigate further.

lib/apiservers/engine/backends/container.go
```diff
@@ -882,198 +877,174 @@ func IsContainerRunning(containerID string) (bool, error) {
...
+
+	go func() {
+		defer stdinReader.Close()
+		// Copy the stdin from the CLI and write to a pipe.  We need to do this so we can
+		// watch the stdin stream for the detach keys.
+		var err error
+		if vc.Config.Tty {
+			_, err = copyEscapable(stdinWriter, clStdin, keys)
+		} else {
+			_, err = io.Copy(stdinWriter, clStdin)
```
hickeng: Don't care for this PR, but is there scope here for us to pass the clStdin directly to setStdinPara,s.WithRawStream? It looks like we've an extra copy here for no benefit.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -91,8 +91,7 @@ func (i *InteractionHandlersImpl) ContainerResizeHandler(params interaction.Cont
 
 	err = connContainer.Resize(cWidth, cHeight, 0, 0)
 	if err != nil {
-		retErr := &models.Error{Message: "SSH Session failed to resize the container's TTY"}
-		return interaction.NewContainerResizeInternalServerError().WithPayload(retErr)
+		return interaction.NewContainerResizeInternalServerError()
```
hickeng: Do we lose potentially useful info here or does the PL log it?

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -102,14 +101,17 @@ func (i *InteractionHandlersImpl) ContainerSetStdinHandler(params interaction.Co
 	log.Printf("Attempting to get ssh session for container %s stdin", params.ID)
 	sshConn, err := i.attachServer.Get(context.Background(), params.ID, interactionTimeout)
 	if err != nil {
-		e := &models.Error{Message: fmt.Sprintf("No stdin found for %s", params.ID)}
-		return interaction.NewContainerSetStdinNotFound().WithPayload(e)
+		err = fmt.Errorf("No stdin found (id:%s): %s", params.ID, err.Error())
+		log.Errorf("%s", err.Error())
+
+		return interaction.NewContainerSetStdinNotFound()
```
hickeng: presume the payload is dropped because it adds no value beyond the base error? (assuming the caller knows which container the connection corresponds to..)

hickeng:
LGTM assuming minor comments are address (namely followup issue for behavioural divergence)

lib/apiservers/engine/backends/container.go
fdawg4l: I don't know what you're referring to.  The PR doesn't change behavior.  If we've diverged, it's not because of these changes.

lib/apiservers/engine/backends/container.go
fdawg4l: I preserved previous behavior here but moved around the anon routine.  You're right though;  without `vc.Config.Tty` being set, the pipe/copy is irrelevant.  I'll fix it possibly in a later PR.  Great catch!

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
fdawg4l: Ah, I'll add a log line.

This is the crux of the panic fix, btw.  The `WithPayload` calls when there is no client to receive the payload will panic the server.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
fdawg4l: Removed to avoid the panic.  The client (if connected) will still get an error code, but no error.  In my testing, thats never been necessary however.

lib/apiservers/engine/backends/container.go
```diff
@@ -153,16 +152,17 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 
 	// bail early if container name already exists
 	if exists := cache.ContainerCache().GetContainer(config.Name); exists != nil {
-		return types.ContainerCreateResponse{},
-			derr.NewErrorWithStatusCode(fmt.Errorf("Conflict. The name \"%s\" is already in use by container %s. You have to remove (or rename) that container to be able to re use that name.", config.Name, exists.ContainerID),
-				http.StatusConflict)
+		err := fmt.Errorf("Conflict. The name \"%s\" is already in use by container %s. You have to remove (or rename) that container to be able to re use that name.", config.Name, exists.ContainerID)
```
caglar10ur: You can use %q instead of escaping from " with \ :)

caglar10ur:
LGTM :sheep: :it:

lib/apiservers/engine/backends/container.go
fdawg4l: Good call.  Will update.  I carried this over from the previous version and simply assigned it to err.  Will take care of it in the next PR.

fdawg4l:
:recycle:

lib/apiservers/engine/backends/container.go
hickeng: There was a prior comment associated with this block - it's been moved to copyStdin so nevermind.

corrieb:
Build failed with the Link test failure that borked other builds

fdawg4l:
Doh!

Issue #1262, opened 2016-06-29
vic-machine create.log not providing the right information on failure


mhagen-vmware:
In the case when I put the wrong IP address in and vic-machine cannot connect to ESX system at all:

Expected error (what is printed to screen):
vic-machine-linux failed:  Failed to connect to https://192.168.1.1/sdk: Post https://192.168.1.1/sdk: dial tcp 192.168.1.1:443: i/o timeout' does not contain 'Installer completed successfully'

Actual create.log error:
INFO[2016-06-29T09:41:47-05:00] ### Installing VCH ####                      
INFO[2016-06-29T09:41:47-05:00] Generating certificate/key pair - private key in ./VCH-local-9017-key.pem 
ERRO[2016-06-29T09:42:17-05:00] Creation cannot continue: failed to create validator

emlin:
Interesting. We didn't actually log out every errors we got before return to main method.

PR #1453, opened 2016-07-12
remove timeout and combine logs

### LINKED ISSUE #1303: vic-machine timeout
timeout usage message is wrong in vic-machine delete
timeout parameter is not necessary for vic-machine inspect

emlin:
Fixes #1303 #1262 

This request will remove timeout from vic-machine respect, cause that command just query objects from vsphere.

Also combine all vic-machine command log to one single log file, vic-machine.log. Separated logs files does not make the story easier. And will lose some log information, if that's returned to vic-machine main file.

### DIFF
--- a/cmd/vic-machine/main.go
+++ b/cmd/vic-machine/main.go
@@ -16,8 +16,11 @@ package main
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
+
+	log "github.com/Sirupsen/logrus"
 
 	"github.com/urfave/cli"
 	"github.com/vmware/vic/cmd/vic-machine/create"
@@ -30,6 +33,10 @@ var (
 	Version  string
 	BuildID  string
 	CommitID string
+)
+
+const (
+	LogFile = "vic-machine.log"
 )
 
 func main() {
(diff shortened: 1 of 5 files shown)


andrewtchin:
lgtm

cmd/vic-machine/main.go
```diff
@@ -72,9 +79,21 @@ func main() {
 	} else {
 		app.Version = fmt.Sprintf("%s-%s", BuildID, CommitID)
 	}
+	// Open log file
+	f, err := os.OpenFile(LOG_FILE, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0644)
+	if err != nil {
+		fmt.Printf("Error opening logfile %s: %v", LOG_FILE, err)
+	}
+	defer f.Close()
```
dougm: This should be inside an `} else {`, otherwise will panic() in the case where we cannot OpenFile() above.

cmd/vic-machine/main.go
```diff
@@ -72,9 +79,21 @@ func main() {
...
+
+	// Initiliaze logger with default TextFormatter
+	log.SetFormatter(&log.TextFormatter{ForceColors: true, FullTimestamp: true})
+	// SetOutput to io.MultiWriter so that we can log to stdout and a file
+	log.SetOutput(io.MultiWriter(os.Stdout, f))
```
dougm: Let's also avoid writing to 'f' when nil.  Above you can start with:
```go 
logs := []io.Writer{io.Stdout}
```

When os.OpenFile is ok, then:

```go
logs = append(f, logs)
```
And here:

```go
log.SetOutput(io.MultiWriter(logs...))
```

cmd/vic-machine/main.go
```diff
@@ -72,9 +79,21 @@ func main() {
 	} else {
 		app.Version = fmt.Sprintf("%s-%s", BuildID, CommitID)
 	}
+	// Open log file
+	f, err := os.OpenFile(LOG_FILE, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0644)
+	if err != nil {
+		fmt.Printf("Error opening logfile %s: %v", LOG_FILE, err)
```
dougm: This message should be written to os.Stderr

cmd/vic-machine/main.go
```diff
@@ -72,9 +79,21 @@ func main() {
...
+	}
+	defer f.Close()
+
+	// Initiliaze logger with default TextFormatter
+	log.SetFormatter(&log.TextFormatter{ForceColors: true, FullTimestamp: true})
+	// SetOutput to io.MultiWriter so that we can log to stdout and a file
+	log.SetOutput(io.MultiWriter(os.Stdout, f))
+
 	if err := app.Run(os.Args); err != nil {
 		fmt.Println("--------------------")
```
dougm: This line should also be switched to log.Errorf, or removed

lib/install/data/data.go
```diff
@@ -89,6 +89,7 @@ func NewData() *Data {
 		MappedNetworksGateways: make(map[string]net.IPNet),
 		MappedNetworksIPRanges: make(map[string][]ip.Range),
 		MappedNetworksDNS:      make(map[string][]net.IP),
+		Timeout:                3 * time.Minute,
```
dougm: I can't see what this timeout is for, 3 minutes seems too short if it is install related.

emlin: This is a default value. In general, default value can be set in argument default, but if there is no timeout option at all, the default value cannot be set, and then it will be 0 second timeout.
So this is to solve the issue in commands without timeout option.

In addition, the timeout is set for application initialization waiting, not for the whole process, so 3 minutes by default for install is ok as well.

dougm:
lgtm

emlin:
♻️

PR #1456, opened 2016-07-12
Temporarily remove default SSH access from appliance for TP3

corrieb:
This will be re-enabled once TP3 has been tagged. This is for bug 1218, but it should not be closed by this PR as this is simply a workaround


hickeng:
We should keep this disabled in the main appliance and add an appliance-dev target for generating development appliance ISOs with it enabled.

fdawg4l:
:sheep: :it:

hickeng:
:sheep: :it:

dougm:
sheep it

PR #1463, opened 2016-07-13
trivial test case fixes

corrieb:


corrieb:
Note that the increase in sleep time is because on some VC clusters, 5 seconds is not long enough for the container to have stopped

dougm:
lgtm

hickeng:
LGTM - especially that increased timeout... ;)

hickeng:
poke

tests/test-cases/Group1-Docker-Commands/1-11-Docker-RM.robot
```diff
@@ -26,7 +26,7 @@ Remove a stopped container
     Should Be Equal As Integers  ${rc}  0
     ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} start ${container}
     Should Be Equal As Integers  ${rc}  0
-    Sleep  5 seconds
+    Sleep  10 seconds
```
caglar10ur: Can't we do better here instead of sleeping some random seconds? Like docker ps --filter status=running?

corrieb: Yes, that would be much better. I think we should have a general issue to cover removing sleeps and replacing with tests

corrieb:
ping

tests/test-cases/Group1-Docker-Commands/1-11-Docker-RM.robot
mhagen-vmware: Friendly addition:
[patch.txt](https://github.com/vmware/vic/files/363015/patch.txt)

hickeng: I like this change from @mhagen-vmware - this can easily be switched to using docker wait once implemented.

corrieb:
ping

Issue #1459, opened 2016-07-13
Link/Alias resolution not working in bridge network


hickeng:
First there is a panic in the personality during `network connect` - I was using an external network but didn't seem to matter whether it was an external network or the bridge that was first or second:
```
2016/07/13 10:13:58 http: panic serving 192.168.78.202:52582: runtime error: invalid memory address or nil pointer dereference
goroutine 44 [running]:
net/http.(*conn).serve.func1(0xc820363d80)
	/usr/local/go/src/net/http/server.go:1389 +0xc1
panic(0x1579a40, 0xc82000e080)
	/usr/local/go/src/runtime/panic.go:443 +0x4e9
github.com/vmware/vic/lib/apiservers/engine/backends.EP2Alias(0x0, 0x0, 0x0, 0x0)
	/C/Users/georg/Documents/VMware/VIC/src/github.com/vmware/vic/lib/apiservers/engine/backends/network.go:163 +0x5da
github.com/vmware/vic/lib/apiservers/engine/backends.(*Network).ConnectContainerToNetwork(0x26e4278, 0xc82034c0c0, 0x40, 0xc82012a270, 0x8, 0x0, 0x0, 0x0)
	/C/Users/georg/Documents/VMware/VIC/src/github.com/vmware/vic/lib/apiservers/engine/backends/network.go:198 +0x602
github.com/vmware/vic/vendor/github.com/docker/docker/api/server/router/network.(*networkRouter).postNetworkConnect(0xc820233cb0, 0x7fbcbf0d60d8, 0xc82038cab0, 0x7fbcbf0d1f20, 0xc82078c680, 0xc8207815e0, 0xc82038c990, 0x0, 0x0)
	/C/Users/georg/Documents/VMware/VIC/src/github.com/vmware/vic/vendor/github.com/docker/docker/api/server/router/network/network_routes.go:124 +0x361
github.com/vmware/vic/vendor/github.com/docker/docker/api/server/router/network.(*networkRouter).(github.com/vmware/vic/vendor/github.com/docker/docker/api/server/router/network.postNetworkConnect)-fm(0x7fbcbf0d60d8, 0xc82038cab0, 0x7fbcbf0d1f20, 0xc82078c680, 0xc8207815e0, 0xc82038c990, 0x0, 0x0)
	/C/Users/georg/Documents/VMware/VIC/src/github.com/vmware/vic/vendor/github.com/docker/docker/api/server/router/network/network.go:32 +0x74
[...]

PR #1465, opened 2016-07-13
fix panic in network connect case

hickeng:
This addresses a panic during network connect of a second network. I think this would have only occurred if the two networks were bridge & an external but unconfirmed.

This doesn't address the underlying issue that is in https://github.com/vmware/vic/issues/1459 but does allow network connect to work in this scenario.

Towards #1459


corrieb:
LGTM. This seems trivial enough that merge should be OK

emlin:
lgtm

PR #1480, opened 2016-07-13
Restore personality server's attach attempt timeout

### LINKED ISSUE #1478: After executing invalid command on container all future containers with tty hang the terminal and do not attach
Build 2840.

Installed in empty nested ESX6u2 on fusion.
installed with => ./vic-machine-darwin create --target 172.16.252.131 --image-datastore datastore1 --name vic-demo --user root --password 'Vmware!23'  --compute-resource /ha-datacenter/host/*/Resources

then in one terminal:
docker -H 172.16.252.132:2376 --tls run -it --name foo busybox

In another I ran:
docker -H 172.16.252.132:2376 --tls run -it --name cat busybox "/bin/sh -c 'ping -c 40 foo'"
and got:
docker: Error response from daemon: server error from portlayer.

Then ran:
docker -H 172.16.252.132:2376 --tls run -it --name dog busybox

And it started the container but did not attach. 

output from ps -a:
def05b2d7d7f        busybox             "sh"                     3 weeks ago         Running                                 foo
75776b6316ed        busybox             "/bin/sh -c 'ping -c "   3 weeks ago         Stopped                                 cat
d305b2345fae        busybox             "sh"                     3 weeks ago         Running                                 dog

sflxn:
Fixes #1478

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -1258,7 +1258,7 @@ func attachStreams(ctx context.Context, vc *viccontainer.VicContainer, clStdin i
 
 	// For stdin, we only have a timeout for connection.  There can be a long duration before
 	// the first entry so there is no timeout for response.
-	plClient, transport := createNewAttachClientWithTimeouts(attachConnectTimeout, 0, 0)
+	plClient, transport := createNewAttachClientWithTimeouts(attachConnectTimeout, 0, attachAttemptTimeout)
 	defer transport.Close()
 
 	if ca.UseStdin {


hickeng:
LGTM - I want this in TP3

caglar10ur:
LGTM :sheep: :it:

sflxn:
poke

lib/apiservers/engine/backends/container.go
```diff
@@ -1258,7 +1258,7 @@ func attachStreams(ctx context.Context, vc *viccontainer.VicContainer, clStdin i
 
 	// For stdin, we only have a timeout for connection.  There can be a long duration before
 	// the first entry so there is no timeout for response.
-	plClient, transport := createNewAttachClientWithTimeouts(attachConnectTimeout, 0, 0)
+	plClient, transport := createNewAttachClientWithTimeouts(attachConnectTimeout, 0, attachAttemptTimeout)
```
fdawg4l: The name of the cost is misleading.  This is apparently a `responseHeaderTimeout`.  Can we name it as such?  

Also the ordering of the args to `createNewAttachClientWithTimeouts` is massively counterintuitive.  Can the parameter order match the order they're being assigned to the `Transport` struct?  Connect, Header, Response seems intuitive.

hickeng: I agree with the renaming, but not for TP3...

hickeng:
poke

lib/apiservers/engine/backends/container.go
sflxn: I'll do it post-TP3.  Let's not delay this tag.  It's on my list of post-TP3 clean up and refactoring to do.

PR #1481, opened 2016-07-13
Fix deadlock in connector.Get

### LINKED ISSUE #1478: After executing invalid command on container all future containers with tty hang the terminal and do not attach

hickeng:
This corrects an incorrect placement of RLock/RUnlock paired with a cond.Wait. Paired with the Go fairness behaviour this causes deadlock if a Writer attempts to acquire the lock if we do not succeed in locating a connection on the first iteration.

Fixes #1478


fdawg4l:
LGTM

dougm:
lgtm

caglar10ur:
LGTM :sheep: :it:

PR #1485, opened 2016-07-13
Unsqash error messages and better error handling

matthewavery:
This removes the useless nature of the dreaded "error from portlayer server"

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -305,11 +305,15 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 		if err != nil {
 			switch err := err.(type) {
 			case *storage.VolumeJoinInternalServerError:
-				return nil, derr.NewErrorWithStatusCode(err.Payload.Message, http.StatusInternalServerError)
+				return types.ContainerCreateResponse{}, derr.NewErrorWithStatusCode(fmt.Errorf(err.Payload.Message), http.StatusInternalServerError)
+
 			case *storage.VolumeJoinDefault:
-				return nil, derr.NewErrorWithStatusCode(err.Payload.Message, http.StatusInternalServerError)
+				return types.ContainerCreateResponse{}, derr.NewErrorWithStatusCode(fmt.Errorf(err.Payload.Message), http.StatusInternalServerError)
+			default:
+				return types.ContainerCreateResponse{}, derr.NewErrorWithStatusCode(err, http.StatusInternalServerError)
 			}
 		}
+
 		h = res.Payload
 	}
 	// commit the create op
@@ -548,6 +552,9 @@ func (c *Container) containerStart(name string, hostConfig *container.HostConfig
 
 		case *containers.StateChangeDefault:
 			return derr.NewRequestNotFoundError(fmt.Errorf("server error from portlayer : %s", err.Payload.Message))
+
+		default:
+			return derr.NewRequestNotFoundError(fmt.Errorf("server error from portlayer : %s", err))
 		}
 	}
 
@@ -563,6 +570,8 @@ func (c *Container) containerStart(name string, hostConfig *container.HostConfig
 
 		case *containers.CommitDefault:
 			return derr.NewRequestNotFoundError(fmt.Errorf("server error from portlayer : %s", err.Payload.Message))
+		default:
+			return derr.NewRequestNotFoundError(fmt.Errorf("server error from portlayer : %s", err))
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -303,7 +303,12 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (types.C
 
 		res, err := client.Storage.VolumeJoin(joinParams)
 		if err != nil {
-			return types.ContainerCreateResponse{}, derr.NewErrorWithStatusCode(fmt.Errorf("Server error from Portlayer: %s", err), http.StatusInternalServerError)
+			switch err := err.(type) {
+			case *storage.VolumeJoinInternalServerError:
+				return nil, derr.NewErrorWithStatusCode(err.Payload.Message, http.StatusInternalServerError)
+			case *storage.VolumeJoinDefault:
+				return nil, derr.NewErrorWithStatusCode(err.Payload.Message, http.StatusInternalServerError)
+			}
```
hickeng: default clause just to be safe

lib/apiservers/engine/backends/volume.go
```diff
@@ -98,13 +98,21 @@ func (v *Volume) VolumeRm(name string) error {
...
+
+		switch err := err.(type) {
+
+		case *storage.RemoveVolumeNotFound:
 			return derr.NewRequestNotFoundError(fmt.Errorf("Get %s: no such volume", name))
+
+		case *storage.RemoveVolumeConflict:
+			return derr.NewRequestConflictError(fmt.Errorf("Volume '%s' is in use", name))
+
+		case *storage.RemoveVolumeInternalServerError:
```
hickeng: This is the one we get if the volume store isn't specified and there's no default? What's the actual message that's sent back in that case?

lib/apiservers/engine/backends/container.go
matthewavery: ah there are going to be a lot of these.... I will look through them and fix it.

lib/apiservers/engine/backends/volume.go
matthewavery: they should now get a "volumestore (<name of store>) not found"

lib/apiservers/engine/backends/container.go
matthewavery: So that will be unreachable code if I put a default in. I will be adding a catch all return below the switch instead.

lib/apiservers/engine/backends/volume.go
hickeng: What happens if they simply haven't specified a store - would (name of store) be "default"?

lib/apiservers/engine/backends/container.go
```diff
@@ -530,24 +541,29 @@ func (c *Container) containerStart(name string, hostConfig *container.HostConfig
...
 	if err != nil {
-		if _, ok := err.(*containers.CommitNotFound); ok {
-			return derr.NewRequestNotFoundError(fmt.Errorf("server error from portlayer"))
+		switch err := err.(type) {
+
+		case *containers.CommitNotFound:
+			return derr.NewRequestNotFoundError(fmt.Errorf("server error from portlayer : %s", err.Payload.Message))
+
+		case *containers.CommitDefault:
+			return derr.NewRequestNotFoundError(fmt.Errorf("server error from portlayer : %s", err.Payload.Message))
```
jzt: If neither of these `case`s are matched, it falls through and returns nil. Other error switching in this file is handled the same way (no `default`). Are these known to be the only possible error types you can get back?

Also, since these two cases do exactly the same thing, why not just make it a general case? Or are you being explicit here for the sake of people looking at the code?

matthewavery: The explicit repetition is there in case we want to handle individual errors with more docker like responses and it makes it easier to add more error types. As for the defaults. they are now fixed.

lib/apiservers/engine/backends/volume.go
matthewavery: yes that would be the case.

lib/apiservers/engine/backends/container.go
jzt: LGTM, no further questions your honor. 👍

fdawg4l:
lgtm

matthewavery:
♻️

jzt:
LGTM

lib/apiservers/portlayer/swagger.yml
```diff
@@ -249,7 +249,7 @@ paths:
             $ref: "#/definitions/VolumeResponse"
         '500':
           description: "Error"
-          $schema:
+          schema:
```
matthewavery: this error should be noted as `swagger validate` does not catch it however it does generate code still. Because of this it can actually affect expected response codes.

matthewavery:
♻️

matthewavery:
♻️

matthewavery:
♻️

PR #1494, opened 2016-07-13
Adding endpoint info to network inspect

hmahmood:


hickeng:
Will need to get the tether tests compiling and passing before we put any else into it.

caglar10ur:
LGTM

lib/apiservers/portlayer/restapi/handlers/scopes_handlers.go
```diff
@@ -259,17 +293,48 @@ func (handler *ScopesHandlersImpl) ScopesUnbindContainer(params scopes.UnbindCon
...
+		sc.Endpoints[i] = &models.EndpointConfig{
+			Gateway:   e.Gateway().String(),
+			ID:        e.ID(),
+			Name:      e.Container().Name(),
+			Network:   e.Scope().Name(),
+			Container: e.Container().ID().String(),
+		}
+
+		ip := e.IP()
+		if ip != nil {
```
hickeng: If you move this block above line 324 and assign Address in the creation it's easier to read. addr can be "" by default for the case ip == nil.

lib/apiservers/portlayer/restapi/handlers/scopes_handlers.go
```diff
@@ -259,17 +293,48 @@ func (handler *ScopesHandlersImpl) ScopesUnbindContainer(params scopes.UnbindCon
...
+		// use subnet as pool
+		sc.IPAM = []string{subnet}
+	}
+
+	eps := scope.Endpoints()
+	sc.Endpoints = make([]*models.EndpointConfig, len(eps))
+	for i, e := range eps {
+		sc.Endpoints[i] = &models.EndpointConfig{
+			Gateway:   e.Gateway().String(),
+			ID:        e.ID(),
```
hickeng: I should know already - what's is the value of this ID given we're not tracking endpoints in the same way docker is?

hickeng:
Minor comments, LGTM.

lib/apiservers/portlayer/restapi/handlers/scopes_handlers.go
hmahmood: This is the PL endpoint struct, so we do track those like docker. The id is just a randomly generated string.

hmahmood:
♻️

PR #1518, opened 2016-07-14
The location of the bootstrap iso image

### LINKED ISSUE #1351: Incorrect formulation of bootstrap iso path
The location of the bootstrap iso image should be set in the VCH config data:
https://github.com/vmware/vic/blob/master/lib/metadata/virtual_container_host.go#L111

and pulled from there when creating containers:
https://github.com/vmware/vic/blob/master/lib/portlayer/exec/handle.go#L221

caglar10ur:
should be set in the VCH config data.

Fixes #1351


caglar10ur:
Pushed it by mistake, please ignore it for now

fdawg4l:
I think this fixes #711 as well.  See https://github.com/vmware/vic/issues/711#issuecomment-226153713

lib/install/data/data.go
```diff
@@ -68,6 +71,8 @@ type InstallerData struct {
 	KeyPEM  string
 	CertPEM string
 
+	BootstrapISO string
```
emlin: You don't have to add value in this deprecated structure, cause you have new value in VirtualContainerHost spec.

lib/install/management/appliance.go
```diff
@@ -223,6 +223,8 @@ func (d *Dispatcher) createApplianceSpec(conf *metadata.VirtualContainerHostConf
 	var devices object.VirtualDeviceList
 	var err error
 
+	conf.BootstrapImagePath = fmt.Sprintf("[%s] %s/%s", conf.ImageStores[0].Host, conf.Name, vConf.BootstrapISO)
```
emlin: Move this to validator, method basics(), and use input.BootstrapISO.

lib/install/validate/validator.go
```diff
@@ -1185,6 +1185,8 @@ func (v *Validator) AddDeprecatedFields(ctx context.Context, conf *metadata.Virt
 	dconfig.ClusterPath = v.Session.Cluster.InventoryPath
 
 	dconfig.ResourcePoolPath = v.ResourcePoolPath
+	// This is the only way to move data from Data to InstallerData
+	dconfig.BootstrapISO = input.BootstrapISO
```
emlin: Remove it.

lib/install/management/appliance.go
caglar10ur: done

lib/install/validate/validator.go
caglar10ur: done

caglar10ur:
@emlin you could give this a another shot :)

emlin:
LGTM 🐮

dougm:
lgtm

caglar10ur:
♻️

PR #1533, opened 2016-07-15
Move log message out of for loop

gigawhitlocks:
This informational log message should only get printed once, and I erroneously put it inside of the for loop below where it should be.

This PR fixes that. Also adds some traces for debug.

### DIFF
--- a/lib/portlayer/storage/vsphere/datastore.go
+++ b/lib/portlayer/storage/vsphere/datastore.go
@@ -28,6 +28,7 @@ import (
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/soap"
 	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/session"
 	"github.com/vmware/vic/pkg/vsphere/tasks"
 	"golang.org/x/net/context"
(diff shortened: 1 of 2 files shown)


emlin:
lgtm

lib/portlayer/storage/vsphere/datastore.go
```diff
@@ -311,6 +312,7 @@ var pathFormat = regexp.MustCompile(`\s([\/\w]+$)`)
 
 // Converts `[datastore] /path` to URL
 func DatastoreToURL(ds string) (*url.URL, error) {
+	defer trace.End(trace.Begin(""))
```
caglar10ur: use ds instead of ""

lib/portlayer/storage/vsphere/datastore.go
```diff
@@ -329,6 +331,7 @@ func DatastoreToURL(ds string) (*url.URL, error) {
 
 // Converts URL for datastores to datastore format ([datastore1] /path/to/thing)
 func URLtoDatastore(u *url.URL) (string, error) {
+	defer trace.End(trace.Begin(""))
```
caglar10ur: use u.String() instead of ""

caglar10ur:
just cosmetic nitpicks LGTM otherwise

gigawhitlocks:
Thanks @caglar10ur

gigawhitlocks:
Poke

PR #1536, opened 2016-07-15
vapp for VCH in VC

### LINKED ISSUE #1350: vic-machine: vapp for VCH on vCenter
vCenter provides the vApp construct as a mechanism for grouping a collection of VMs that work together as an application. It extends resource pool mechanisms to add high level power operations, changes the deletion behaviour to delete child VMs instead of re-parent, allows for disabling of operations on child VMs, etc.

This issue covers basic vApp creation:
- [ ] VCH resource pool is replaced with a vApp in clusters with compatible DRS config

emlin:
Fixes #1350 

  Create vApp in VC while deploy VCH, and then create every VM under vApp. In ESX, still keep resource pool.
  As vApp is one special resource pool, except creation, all the other operation based on resource pool is left, for example, to  query child VMs under the resource pool.

### DIFF
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -16,6 +16,7 @@ package management
 
 import (
 	"fmt"
+	"path"
 	"strconv"
 	"time"
 
@@ -53,7 +54,7 @@ func (d *Dispatcher) isVCH(vm *vm.VirtualMachine) (bool, error) {
 
 	info, err := vm.FetchExtraConfig(d.ctx)
 	if err != nil {
-		err = errors.Errorf("Failed to fetch guest info of appliance vm, %s", err)
+		err = errors.Errorf("Failed to fetch guest info of appliance vm: %s", err)
 		return false, err
 	}
 
@@ -72,7 +73,7 @@ func (d *Dispatcher) checkExistence(conf *metadata.VirtualContainerHostConfigSpe
 	defer trace.End(trace.Begin(""))
 
 	var err error
-	d.vchPoolPath = fmt.Sprintf("%s/%s", settings.ResourcePoolPath, conf.Name)
+	d.vchPoolPath = path.Join(settings.ResourcePoolPath, conf.Name)
 	var orp *object.ResourcePool
 	var vapp *object.VirtualApp
 	if d.isVC {
@@ -100,8 +101,8 @@ func (d *Dispatcher) checkExistence(conf *metadata.VirtualContainerHostConfigSpe
 	}
 	if vm == nil {
 		if vapp != nil {
-			err = errors.Errorf("virtual app %s is found, but is not VCH, please choose different name", d.vchPoolPath)
-			log.Errorf("%s", err)
+			err = errors.Errorf("virtual app %q is found, but is not VCH, please choose different name", d.vchPoolPath)
+			log.Error(err)
 			return err
 		}
 		return nil
@@ -109,10 +110,10 @@ func (d *Dispatcher) checkExistence(conf *metadata.VirtualContainerHostConfigSpe
 
 	log.Debugf("Appliance is found")
 	if ok, verr := d.isVCH(vm); !ok {
-		verr = errors.Errorf("VM %s is found, but is not VCH appliance, please choose different name", conf.Name)
+		verr = errors.Errorf("VM %q is found, but is not VCH appliance, please choose different name", conf.Name)
 		return verr
 	}
-	err = errors.Errorf("Appliance %s exists, to install with same name, please delete it first.", conf.Name)
+	err = errors.Errorf("Appliance %q exists, to install with same name, please delete it first.", conf.Name)
 	return err
 }
 
@@ -123,7 +124,7 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
 	power, err := vm.PowerState(d.ctx)
 	if err != nil || power != types.VirtualMachinePowerStatePoweredOff {
 		if err != nil {
-			log.Warnf("Failed to get vm power status %s: %s", vm.Reference(), err)
+			log.Warnf("Failed to get vm power status %q: %s", vm.Reference(), err)
 		}
 		if !force {
 			if err != nil {
@@ -131,12 +132,12 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
 			}
 			name, err := vm.Name(d.ctx)
 			if err != nil {
-				log.Errorf("VM name is not found, %s", err)
+				log.Errorf("VM name is not found: %s", err)
 			}
 			if name != "" {
-				err = errors.Errorf("VM %s is powered on", name)
+				err = errors.Errorf("VM %q is powered on", name)
 			} else {
-				err = errors.Errorf("VM %s is powered on", vm.Reference())
+				err = errors.Errorf("VM %q is powered on", vm.Reference())
 			}
 			return err
 		}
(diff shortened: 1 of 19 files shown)


gigawhitlocks:
this LGTM

lib/install/management/appliance.go
```diff
@@ -67,14 +68,42 @@ func (d *Dispatcher) isVCH(vm *vm.VirtualMachine) (bool, error) {
 	return false, nil
 }
 
-func (d *Dispatcher) checkExistence(conf *metadata.VirtualContainerHostConfigSpec) error {
+func (d *Dispatcher) checkExistence(conf *metadata.VirtualContainerHostConfigSpec, settings *data.InstallerData) error {
 	defer trace.End(trace.Begin(""))
 
-	vm, err := d.findAppliance(conf)
+	var err error
+	d.vchPoolPath = fmt.Sprintf("%s/%s", settings.ResourcePoolPath, conf.Name)
```
dougm: Using `path.Join()` is a bit more concise for constructing inventory paths, no need for a format string.

lib/install/management/appliance.go
```diff
@@ -67,14 +68,42 @@ func (d *Dispatcher) isVCH(vm *vm.VirtualMachine) (bool, error) {
...
+	}
+
+	rp := compute.NewResourcePool(d.ctx, d.session, orp.Reference())
+	vm, err := rp.GetChildVM(d.ctx, d.session, conf.Name)
 	if err != nil {
 		return err
 	}
 	if vm == nil {
+		if vapp != nil {
+			err = errors.Errorf("virtual app %s is found, but is not VCH, please choose different name", d.vchPoolPath)
```
dougm: Would be good in general if we use `%q` rather than `%s`, the quotes help to see if user input happens to include whitespace.

lib/install/management/appliance.go
```diff
@@ -67,14 +68,42 @@ func (d *Dispatcher) isVCH(vm *vm.VirtualMachine) (bool, error) {
...
+			log.Errorf("%s", err)
```
dougm: No need for the format string here, `log.Error(err)` would do.

lib/install/management/create.go
```diff
@@ -38,20 +38,38 @@ func (d *Dispatcher) CreateVCH(conf *metadata.VirtualContainerHostConfigSpec, se
...
+		}
+	} else {
+		if d.vchPool, err = d.createResourcePool(conf, settings); err != nil {
+			detail := fmt.Sprintf("Creating resource pool failed: %s", err)
+			if !d.force {
+				return errors.New(detail)
+			}
+
+			log.Error(detail)
+			log.Errorf("Deploying vch under parent pool %s, , cause --force is provided", settings.ResourcePoolPath)
```
dougm: extra ' , ' in that message.  Maybe change to:
```
"Deploying vch under parent pool %q (--force=true)"
```

lib/install/management/delete.go
```diff
@@ -104,11 +104,14 @@ func (d *Dispatcher) DeleteVCHInstances(vmm *vm.VirtualMachine, conf *metadata.V
...
 		return err
 	}
-	_, ok := ref.(*object.ResourcePool)
-	if !ok {
-		log.Errorf("Failed to find resource pool %s, %s", rpRef, err)
-		return err
+	switch ref.(type) {
+	case *object.VirtualApp:
+	case *object.ResourcePool:
+		break
```
dougm: No need for the 'break' here, since Go switch does not fall through by default.  You could even just do:
```
switch ref.(type) {
    case *object.VirtualApp, *object.ResourcePool:
        // ok
    default:
        log.Errorf(...)
}
```

lib/portlayer/portlayer.go
```diff
@@ -64,11 +65,22 @@ func Init(ctx context.Context, sess *session.Session) error {
 	cr := exec.VCHConfig.ComputeResources[0]
 	r, err := f.ObjectReference(ctx, cr)
 	if err != nil {
-		detail := fmt.Sprintf("could not get resource pool reference from %s: %s", cr.String(), err)
+		detail := fmt.Sprintf("could not get resource pool or vitual app reference from %s: %s", cr.String(), err)
```
dougm: typo 'vitual'

lib/portlayer/portlayer.go
```diff
@@ -64,11 +65,22 @@ func Init(ctx context.Context, sess *session.Session) error {
...
-	exec.VCHConfig.ResourcePool = r.(*object.ResourcePool)
+	switch o := r.(type) {
+	case *object.VirtualApp:
+		exec.VCHConfig.VirtualApp = o
+		exec.VCHConfig.ResourcePool = o.ResourcePool
+	case *object.ResourcePool:
+		exec.VCHConfig.ResourcePool = o
+	default:
+		detail := fmt.Sprintf("could not get resource pool or virtual app from reference %s: object type is wrong", cr.String())
+		log.Errorf(detail)
```
dougm: `log.Error(detail)`

dougm:
A few minor nits, more cases of those I mentioned in there too.  Otherwise, LGTM

emlin:
@dougm would you mind to check it again?
I think %q will be slightly better for user message, so changed a little bit more in vic-machine files.

lib/install/management/delete.go
```diff
@@ -101,22 +101,25 @@ func (d *Dispatcher) DeleteVCHInstances(vmm *vm.VirtualMachine, conf *metadata.V
...
-	_, ok := ref.(*object.ResourcePool)
-	if !ok {
-		log.Errorf("Failed to find resource pool %s, %s", rpRef, err)
-		return err
+	switch ref.(type) {
+	case *object.VirtualApp:
+	case *object.ResourcePool:
+		//		ok
+	default:
+		log.Errorf("Failed to find virtual app or resource pool %q, %s", rpRef, err)
```
dougm: Looks like this should return an error here too.

Minor change for the format string, we are using colon as a delimiter, so "... %q: %s" instead of comma.

lib/install/management/finder.go
```diff
@@ -46,67 +47,85 @@ func (d *Dispatcher) NewVCHFromID(id string) (*vm.VirtualMachine, error) {
...
+			err = errors.Errorf("Failed to query appliance (%q): %s", moref, err)
 			return nil, err
 		}
 		log.Debugf("Appliance is not found")
 		return nil, nil
 	}
 	ovm, ok := ref.(*object.VirtualMachine)
 	if !ok {
-		log.Errorf("Failed to find VM %s, %s", moref, err)
+		log.Errorf("Failed to find VM %q, %s", moref, err)
```
dougm: comma should be colon in format string

lib/install/management/finder.go
```diff
@@ -46,67 +47,85 @@ func (d *Dispatcher) NewVCHFromID(id string) (*vm.VirtualMachine, error) {
...
-	vchPool, err := v.ResourcePoolHelper(d.ctx, vchPath)
-	if err != nil {
-		log.Errorf("Failed to get VCH resource pool %s: %s", vchPath, err)
-		return nil, vchPath, err
+	d.vchPoolPath = path.Join(parent.InventoryPath, name)
+	var vchPool *object.ResourcePool
+	if d.isVC {
+		vapp, err := d.findVirtualApp(d.vchPoolPath)
+		if err != nil {
+			log.Errorf("Failed to get VCH virtual app %q, %s", d.vchPoolPath, err)
```
dougm: looks like there are several of these, where comma should be colon in the format string.

emlin:
Alright, I didn't know the format. Now it should be better. :)

lib/install/management/delete.go
emlin: yeah, it's an issue

dougm:
lgtm

emlin:
♻️

PR #1537, opened 2016-07-15
Make datastore path parsing more robust

### LINKED ISSUE #1535: Add special characters & error handling to DatastoreToURL
We found an issue with some special characters, namely `-`, causing Datastore paths to not parse properly. The regex needs to be expanded to allow vSphere-permitted special characters in paths, the unit test needs to be expanded to cover these cases, and the call to `DatastoreToURL` found in `DeleteVolumeStoresIfForced` in store_files.go needs to handle errors output by this method properly so that even if a path is not parsed properly, the user gets a real error instead of a nil pointer.

gigawhitlocks:
Adds legality for - and _ in datastore paths. **Please comment if there are other special characters that should be allowed** as there does not seem to be a canonical list of vSphere-allowed characters for datastore names & paths

Also moves a call to log.Debug to _after_ the error checking when calling DatastoreToURL so that we get an error message instead of a nil pointer if DatastoreToURL returns nil.

Fixes #1535

(also at this time this includes some changes from another PR I have open, in the first commit, those are (mostly) unrelated)

### DIFF
--- a/lib/install/management/store_files.go
+++ b/lib/install/management/store_files.go
@@ -203,12 +203,13 @@ func (d *Dispatcher) deleteVolumeStoreIfForced(conf *metadata.VirtualContainerHo
 
 		dsURL, err := vsphere.DatastoreToURL(url.Path)
 
-		log.Debugf("Provided datastore URL: %s\nParsed volume store path: %s", url.Path, dsURL.Path)
-
 		if err != nil {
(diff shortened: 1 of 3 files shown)


dougm:
lgtm

andrewtchin:
lgtm

emlin:
lgtm

gigawhitlocks:
poke

gigawhitlocks:
ROBUST

PR #1553, opened 2016-07-18
Rename Datastore helper to Helper and move to pkg

fdawg4l:
This struct is fairly useful in that it gives you a datastore object
relative to a given path.  And the methods attached to it do the URI
munging for the caller (relative to that root path).  Might as well move it to /pkg.


dougm:
The name, 'datastore.DSWrapper' - `DS` stutters a bit, seems `datastore.Wrapper` would do?
We also have some methods named `FooHelper`, can we stick with just using `Helper` instead of also having `Wrapper`?

There's a number of these: `uuid.New().String()[0:16]`.  Some use a `-` delim, some don't.
Might be nice to have `datastore.ID(name string)` to concat the random bits + delim + name.

fdawg4l:
Struggled with the name.  Will rename to `Helper`.    Will also add a utility function to build the name.  It's for tests only which happen to live in other packages.

fdawg4l:
Updated.  Fire when read.  Thanks!

lib/portlayer/storage/vsphere/parent_test.go
```diff
@@ -101,3 +109,49 @@ func TestParentSaveRestore(t *testing.T) {
 		return
 	}
 }
+
+func Session(ctx context.Context, t *testing.T) *session.Session {
```
dougm: Looks like this and DSsetup below are not used?  Tests above are using datastore.DSsetup, which looks about the same.

pkg/vsphere/datastore/datastore_util.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	return s
+}
+
+func DSsetup(t *testing.T) (context.Context, *Helper, func()) {
+	ctx := context.Background()
+	sess := Session(ctx, t)
+
+	ds, err := NewHelper(ctx, sess, sess.Datastore, uuid.New().String()[0:16]+"dstests")
```
dougm: Looks like you could use TestName("dstests") here instead

lib/portlayer/storage/vsphere/parent_test.go
fdawg4l: Oops.  Done.

pkg/vsphere/datastore/datastore_util.go
fdawg4l: Done.

dougm:
lgtm

caglar10ur:
LGTM :sheep: :it:

Do you need to change github PR title as it still refers the old name?

fdawg4l:
Thanks!  Updated title.  Will squash too!

PR #1556, opened 2016-07-19
Integrate toolbox with tether

### LINKED ISSUE #407: Add IP address reporting

dougm:
The toolbox service is started as a tether extension.
The VCH and container VM IP addresses are now displayed in WebClient and
via 'govc vm.ip'.

Fixes #407


caglar10ur:
LGTM :sheep: :it:

infra/machines/vcsa/create-esxi-vm.sh
```diff
@@ -17,19 +17,22 @@
...
 }
 
 disk=48
 mem=16
 iso=VMware-VMvisor-6.0.0-3634798.x86_64.iso
-vib=http://download3.vmware.com/software/vmw-tools/esxui/esxui-signed-3843236.vib
+vib=http://download3.vmware.com/software/vmw-tools/esxui/esxui-signed-3976049.vib
 
-while getopts c:d:i:m:s flag
+while getopts c:d:hi:m:s flag
```
corrieb: shouldn't there be a colon between h and i?

corrieb:
LGTM

infra/machines/vcsa/create-esxi-vm.sh
dougm: The colon is only for flags that take an argument, '-h' and '-s' do not take an argument.

infra/machines/vcsa/create-esxi-vm.sh
```diff
@@ -17,19 +17,22 @@
...
+
+GOVC_* environment variables also apply, see https://github.com/vmware/govmomi/tree/master/govc#usage
+If GOVC_USERNAME is set, it is used to create an account on the ESX vm.  Default is to use the existing root account.
+If GOVC_PASSWORD is set, the account password will be set to this value.  Default is to use the given ESX_URL password.
+EOF
 }
 
 disk=48
 mem=16
 iso=VMware-VMvisor-6.0.0-3634798.x86_64.iso
```
caglar10ur: should we also switch VMware-VMvisor-6.0.0-3620759.x86_64.iso while at it?

dougm: yes, will do.

dougm:
👊

PR #1572, opened 2016-07-19
Fill in VCH info and real host for docker info

### LINKED ISSUE #1390: Docker info should get reasonable values for vch info
Docker info currently does not populate core count or memory limits.  The OS values are also data for the appliance VM itself.  We should get this info from guestinfo and govmomi.

depends on #861

sflxn:
Retrieve the following parameters for docker info

1. Core count = VCH mhz (same as Bonneville)
2. MemInfo = VCH memory limit (same as Bonneville)
3. Operating System = real host OSType (from vmomi)
4. Additional info in SystemStatus
   - real host product name (ESXi, VMware vCenter...) from AboutInfo
   - real host OS (from real host OSType in AboutInfo)
   - real host OS version (from Version in AboutInfo)
   
resolves #1390.

### DIFF
--- a/lib/apiservers/engine/backends/system.go
+++ b/lib/apiservers/engine/backends/system.go
@@ -14,10 +14,20 @@
 
 package vicbackends
 
+//****
+// system.go
+//
+// Rules for code to be in here:
+// 1. No remote or swagger calls.  Move those code to system_portlayer.go
+// 2. Always return docker engine-api compatible errors.
+//		- Do NOT return fmt.Errorf()
+//		- Do NOT return errors.New()
+//		- DO USE the aliased docker error package 'derr'
+//		- It is OK to return errors returned from functions in system_portlayer.go
+
 import (
 	"bufio"
 	"fmt"
-	"net/http"
 	"os"
 	"regexp"
 	"runtime"
@@ -29,12 +39,9 @@ import (
 	log "github.com/Sirupsen/logrus"
 
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
-	"github.com/vmware/vic/lib/apiservers/portlayer/client"
-	"github.com/vmware/vic/lib/apiservers/portlayer/client/containers"
-	"github.com/vmware/vic/lib/apiservers/portlayer/client/misc"
 	"github.com/vmware/vic/pkg/trace"
 
-	derr "github.com/docker/docker/errors"
+	//	derr "github.com/docker/docker/errors"
 	"github.com/docker/docker/pkg/parsers/kernel"
 	"github.com/docker/docker/pkg/platform"
 	"github.com/docker/docker/pkg/system"
@@ -44,13 +51,6 @@ import (
 )
 
 type System struct {
-}
-
-type ContainerStatus struct {
-	count      int
-	numRunning int
-	numStopped int
-	numPaused  int
 }
 
 const (
@@ -78,24 +78,22 @@ func (s *System) SystemInfo() (*types.Info, error) {
 	}
 
 	// Check if portlayer server is up
-	plClient := PortLayerClient()
-
 	systemStatus := make([][2]string, 1)
 	systemStatus[0][0] = PortLayerName()
-	if pingPortlayer(plClient) {
+	if VicPingPortlayer() {
 		systemStatus[0][1] = "RUNNING"
 	} else {
 		systemStatus[0][1] = "STOPPED"
 	}
 
 	// Retrieve number of images from storage port layer
-	numImages := getImageCount(plClient)
+	numImages := getImageCount()
 	if err != nil {
 		log.Infof("System.SytemInfo unable to get image count: %s.", err.Error())
 	}
 
 	// Retieve container status from port layer
-	containerStatus, err := getContainerStatus(plClient)
+	containerStatus, err := VicGetContainerCount()
 	if err != nil {
 		log.Infof("System.SytemInfo unable to get global status on containers: ", err.Error())
 	}
@@ -244,46 +242,7 @@ func getOperatingSystem() string {
 	return "Linux"
 }
 
-func pingPortlayer(plClient *client.PortLayer) bool {
-	if plClient != nil {
-		pingParams := misc.NewPingParams()
-		_, err := plClient.Misc.Ping(pingParams)
-		if err != nil {
-			log.Info("Ping to portlayer failed")
-			return false
-		}
-		return true
-	}
-
-	log.Errorf("Portlayer client is invalid")
-	return false
-}
-
-func getImageCount(plClient *client.PortLayer) int {
-
+func getImageCount() int {
 	images := cache.ImageCache().GetImages()
 	return len(images)
-}
-
-// Use the Portlayer's support for docker ps to get the container count
-func getContainerStatus(plClient *client.PortLayer) (ContainerStatus, error) {
-	var status ContainerStatus
-
-	all := new(bool)
-	*all = true
-	containList, err := plClient.Containers.GetContainerList(containers.NewGetContainerListParams().WithAll(all))
-	if err != nil {
-		return status, derr.NewErrorWithStatusCode(fmt.Errorf("Failed to get container list: %s", err), http.StatusInternalServerError)
-	}
-
-	for _, t := range containList.Payload {
-		if *t.Status == "Running" {
-			status.numRunning++
-		} else if *t.Status == "Stopped" {
-			status.numStopped++
-		}
-		status.count++
-	}
-
-	return status, nil
 }
--- a/lib/apiservers/engine/backends/system.go
+++ b/lib/apiservers/engine/backends/system.go
@@ -26,12 +26,8 @@ package vicbackends
 //		- It is OK to return errors returned from functions in system_portlayer.go
 
 import (
-	"bufio"
 	"fmt"
-	"os"
-	"regexp"
 	"runtime"
-	"strings"
 	"time"
 
 	"golang.org/x/net/context"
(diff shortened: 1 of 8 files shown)


lib/apiservers/engine/backends/system.go
```diff
@@ -14,90 +14,55 @@
...
-
-	// Retrieve number of images from storage port layer
-	numImages := getImageCount(plClient)
-	if err != nil {
-		log.Infof("System.SytemInfo unable to get image count: %s.", err.Error())
-	}
-
 	// Retieve container status from port layer
-	containerStatus, err := getContainerStatus(plClient)
+	containerStatus, err := VicGetContainerCount()
```
cgtexmex: Why is the function called "VicGetContainerCount" -- when you're returning a ContainerStatus struct?  Why not call the function ContainerStatus()?

lib/apiservers/engine/backends/system_portlayer.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+type ContainerStatus struct {
+	count      int
+	numRunning int
+	numStopped int
+	numPaused  int
+}
+
+func VicPingPortlayer() bool {
```
cgtexmex: Why are all the functions prefixed with Vic?  This is the Vic project not sure what good it accomplishes to prefix the functions with it...

lib/portlayer/exec/vchinfo.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"golang.org/x/net/context"
+
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+// NCPU returns the CPU limit (MHZ)
+func NCPU(ctx context.Context, moref ...types.ManagedObjectReference) int64 {
```
cgtexmex: These resource pool functions should be combined into one call -- no need to go to the infra twice for the info.

lib/portlayer/exec/vchinfo.go
```diff
@@ -0,0 +1,84 @@
...
+	if err := VCHConfig.ResourcePool.Properties(ctx, r, []string{"parent", "config.memoryAllocation"}, &p); err != nil {
+		return 0
+	}
+
+	limit := p.Config.MemoryAllocation.GetResourceAllocationInfo().Limit
+	if limit == -1 {
+		return MemTotal(ctx, *p.Parent)
+	}
+	// Docker API expects B govmomi returns MB
+	return limit * 1024 * 1024
```
cgtexmex: Any docker specific formatting should be handled at the persona and not "leak" into the portlayer

lib/apiservers/engine/backends/system_portlayer.go
```diff
@@ -0,0 +1,126 @@
...
+		} else if *t.Status == "Stopped" {
+			status.numStopped++
+		}
+		status.count++
+	}
+
+	return status, nil
+}
+
+func VicGetVCHInfo() (*models.VCHInfo, error) {
```
cgtexmex: In addition to the previous "Vic" comment -- I'm not sure why you put Get in the function name -- I'd say it's more idiomatic from a go perspective to have "VCHInfo" as the function name.

lib/apiservers/engine/backends/system.go
sflxn: Sounds reasonable.

lib/portlayer/exec/vchinfo.go
sflxn: That was cut and pasted from Bonneville.  However, it's reasonable to move that to personality side.

lib/portlayer/exec/vchinfo.go
sflxn: True, I can move this call to portlayer init() instead of calling this in the rest server.

lib/apiservers/engine/backends/system_portlayer.go
sflxn: Removed 'Vic'.  Made things a little clearer by adding an interface.  Should also make unit testing easier.

lib/apiservers/engine/backends/system_portlayer.go
sflxn: Removed 'Vic'. Made things a little clearer by adding an interface. Should also make unit testing easier.

lib/apiservers/engine/backends/system_portlayer.go
```diff
@@ -0,0 +1,135 @@
...
+type VicSystemProxy interface {
+	PingPortlayer() bool
+	ContainerCount() (ContainerCounts, error)
+	VCHInfo() (*models.VCHInfo, error)
+}
+
+type SystemProxy struct{}
+
+func (s *SystemProxy) PingPortlayer() bool {
+	defer trace.End(trace.Begin("VicPingPortlayer"))
```
cgtexmex: since the functions changed need to update the trace string

lib/apiservers/engine/backends/system.go
```diff
@@ -128,19 +99,14 @@ func (s *System) SystemInfo() (*types.Info, error) {
...
-		MemTotal:        meminfo.MemTotal,      //stubbed
-		CPUCfsPeriod:    false,
-		CPUCfsQuota:     false,
-		CPUShares:       false,
-		CPUSet:          false,
-		OomKillDisable:  true,
+		Name:           VchConfig().Name,
+		KernelVersion:  "",
+		Architecture:   platform.Architecture, //stubbed
+		CPUCfsPeriod:   false,
```
caglar10ur: can we make those true to get rid of warnings?

lib/apiservers/engine/backends/system_portlayer.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// Interface to make unit testing viable
+type VicSystemProxy interface {
+	PingPortlayer() bool
+	ContainerCount() (ContainerCounts, error)
+	VCHInfo() (*models.VCHInfo, error)
+}
+
+type SystemProxy struct{}
```
cgtexmex: You might consider having SystemProxy encapsulate the ContainerCount struct (or just do away with it and have systemProxy have those attributes) and the VCHInfo model -- so you'd call the proxy to get the the info and then interrogate the proxy for the data. 

IMO that would help to increase the readability of these sections.

lib/apiservers/engine/backends/system.go
sflxn: I'll look into it, but I think the CLI does that if the values are false.

caglar10ur: yep, that's why I'm suggesting setting all of those true, you can also add https://github.com/docker/docker/blob/master/daemon/info.go#L127 and set them true too

lib/apiservers/engine/backends/system_portlayer.go
sflxn: Only problem with that is that there the system backend has one systemproxy, and if we add counts to it, then dev needs to remember to call ContainerCount() before using the values.  By returning all the counts in a separate struct, the contract is this returns a point in time count of containers.

lib/apiservers/engine/backends/system.go
sflxn: Well, I silenced them for the CLI, but API users need to beware.  Real solution is for us to submit a PR to Docker to ignore these value as they do for Windows once we GA.

lib/apiservers/engine/backends/system_portlayer.go
cgtexmex: Ok - I'm not really following the reason fro SystemProxy then as an empty struct -- I realize you add the interface and thus need a struct...but an empty struct to me seems silly...

If you want to leave as it so be it, but from my perspective there's added complexity with limited benefit.

cgtexmex:
LGTM

lib/apiservers/engine/backends/system_portlayer.go
sflxn: I added an interface so we can mock it in unit testing.  I added the empty struct so that the real code followed the interface.  In unit testing, we can replace it with a mock system proxy to fake out calls.

lib/apiservers/engine/backends/system.go
```diff
@@ -14,90 +14,62 @@
...
 
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
-	"github.com/vmware/vic/lib/apiservers/portlayer/client"
-	"github.com/vmware/vic/lib/apiservers/portlayer/client/containers"
-	"github.com/vmware/vic/lib/apiservers/portlayer/client/misc"
 	"github.com/vmware/vic/pkg/trace"
 
-	derr "github.com/docker/docker/errors"
-	"github.com/docker/docker/pkg/parsers/kernel"
+	//	derr "github.com/docker/docker/errors"
```
fdawg4l: Can remove.

lib/apiservers/engine/backends/system.go
```diff
@@ -14,90 +14,62 @@
...
-	operatingSystem := getOperatingSystem()
-
-	meminfo, err := system.ReadMemInfo()
-	if err != nil {
-		log.Errorf("Could not read system memory info: %v", err)
-	}
-
-	// Check if portlayer server is up
-	plClient := PortLayerClient()
+// Override this for unit testing
```
fdawg4l: Probably worth a comment for what this is when *not* unit testing vs when unit testing.

lib/apiservers/engine/backends/system_portlayer.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (s *SystemProxy) PingPortlayer() bool {
```
fdawg4l: Looks like you're basically doing a `return (err != nil)` and making the prototype return a bool.  Why not return an error?  The the caller can type assert the error and do something useful.

lib/apiservers/engine/backends/system.go
```diff
@@ -14,90 +14,62 @@
...
 
-	systemStatus := make([][2]string, 1)
-	systemStatus[0][0] = PortLayerName()
-	if pingPortlayer(plClient) {
-		systemStatus[0][1] = "RUNNING"
-	} else {
-		systemStatus[0][1] = "STOPPED"
-	}
+func init() {
+	systemProxy = &SystemProxy{}
```
fdawg4l: I understand `SystemProxy` is used to fulfill an interface to be mocked out for tests.  But SystemProxy has no members, therefore you don't need a global.  

I would, in the `System` struct, include a member to the interface `VicSystemProxy`.  Then you don't need this global and you can still fullfill your testing requirement using mocks.

lib/apiservers/engine/backends/system.go
```diff
@@ -162,6 +131,50 @@ func (s *System) SystemInfo() (*types.Info, error) {
...
+		info.SystemStatus = append(info.SystemStatus, status)
+	} else {
+		status := [2]string{PortLayerName(), "STOPPED"}
+		info.SystemStatus = append(info.SystemStatus, status)
+	}
+
+	// Add in vch information
+	vchInfo, err := systemProxy.VCHInfo()
+	if err != nil || vchInfo == nil {
+		log.Infof("System.SystemInfo unable to get vch info from port layer: %s", err.Error())
```
caglar10ur: you can add return info, nil here, lose the else and indent the whole block

lib/apiservers/engine/backends/system_portlayer.go
```diff
@@ -0,0 +1,135 @@
...
+	log "github.com/Sirupsen/logrus"
+	derr "github.com/docker/docker/errors"
+
+	"github.com/vmware/vic/lib/apiservers/portlayer/client/containers"
+	"github.com/vmware/vic/lib/apiservers/portlayer/client/misc"
+	"github.com/vmware/vic/lib/apiservers/portlayer/models"
+	"github.com/vmware/vic/pkg/trace"
+)
+
+type ContainerCounts struct {
```
caglar10ur: isn't count == numRunning + numStopped + numPaused?

You could even drop this struct all together and return 3 ints directly and use them on the info

lib/apiservers/portlayer/restapi/handlers/misc_handlers.go
```diff
@@ -16,18 +16,45 @@ package handlers
...
+	osVer := new(string)
+	hostProduct := new(string)
+
+	*n = exec.VCHConfig.VCHMhz
+	*m = exec.VCHConfig.VCHMemoryLimit
+	*os = exec.VCHConfig.HostOS
+	*osVer = exec.VCHConfig.HostOSVersion
+	*hostProduct = exec.VCHConfig.HostProductName
+
+	vchInfo := &models.VCHInfo{
```
caglar10ur: why we are not using exec.VCHConfig values, am i missing something (eg; CPUMhz: &exec.VCHConfig.VCHMhz)?

caglar10ur:
LGTM with some nitpicks

lib/apiservers/engine/backends/system_portlayer.go
sflxn: That operation doesn't actually return any error.  It's just a ping.  Either it's up (server returns success) or it's down (client returns error).

lib/apiservers/engine/backends/system.go
sflxn: That's a good idea.  Changing that.

lib/apiservers/engine/backends/system.go
sflxn: Sure.

lib/apiservers/engine/backends/system_portlayer.go
sflxn: I could.  That was the original way, but I changed it because returns were getting long.  I have no preference.  I'll change it back.

lib/apiservers/portlayer/restapi/handlers/misc_handlers.go
sflxn: sure

sflxn:
poke

lib/apiservers/engine/backends/system_portlayer.go
fdawg4l: Right up (returns nil) or not (returns error).  I don't like losing errors for bools.

fdawg4l:
lgtm.  thanks for taking care of the global.  this looks nicer.

Issue #1547, opened 2016-07-18
PortLayer container cache to include MoRef ID


cgtexmex:
Currently the PortLayer container cache is a map with the container ID as the key.  This needs to be refactored to allow for the lookup of container vms via the vSphere MoRef ID of the container VM.  This will be in addition to look up via container ID.  

The MoRef ID is the unique identifier of a vSphere managed object (per cluster) and is returned as part of the event stream.  

Related to #1329

PR #1573, opened 2016-07-19
Refactors exec layer container cache with addition of InfraID

cgtexmex:
Change allows the container to be retrieved from the cache
by ID or InfraID.  In the case of vSphere the infraID is the
MoRef.  This is being added to support event streaming.

Fixes: #1547

### DIFF
--- a/lib/portlayer/exec/container.go
+++ b/lib/portlayer/exec/container.go
@@ -33,11 +33,8 @@ import (
 	"golang.org/x/net/context"
 )
 
-var containers map[ID]*Container
-var containersLock sync.Mutex
-
 func init() {
-	containers = make(map[ID]*Container)
+	NewContainerCache()
 }
 
 type State int
@@ -54,6 +51,7 @@ type Container struct {
 
 	ExecConfig *metadata.ExecutorConfig
 	State      State
+
 	// friendly description of state
 	Status string
 
@@ -69,28 +67,27 @@ func NewContainer(id ID) *Handle {
 	}
 	con.ExecConfig.ID = id.String()
 
-	containersLock.Lock()
-	containers[id] = con
-	containersLock.Unlock()
+	// add to cache
+	containers.Put(con)
 
 	return con.newHandle()
 }
 
 func GetContainer(id ID) *Handle {
-	containersLock.Lock()
-	defer containersLock.Unlock()
-
-	if con, ok := containers[id]; ok {
-		return con.newHandle()
+	// get from the cache
+	container := containers.Container(id.String())
+	if container != nil {
+		return container.newHandle()
 	}
-
 	return nil
+
 }
 
 func (c *Container) newHandle() *Handle {
 	return newHandle(c)
 }
 
+// TODO: Is this used anywhere?
 func (c *Container) cacheExecConfig(ec *metadata.ExecutorConfig) {
 	c.Lock()
 	defer c.Unlock()
@@ -177,6 +174,8 @@ func (c *Container) Commit(ctx context.Context, sess *session.Session, h *Handle
 
 	c.ExecConfig = &h.ExecConfig
 
+	// add or overwrite the container in the cache
+	containers.Put(c)
 	return nil
 }
 
@@ -237,6 +236,7 @@ func (c *Container) stop(ctx context.Context) error {
 }
 
 func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
+	defer trace.End(trace.Begin("Container.Remove"))
 	c.Lock()
 	defer c.Unlock()
 
@@ -271,15 +271,13 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
 		log.Debugf("Failed to delete %s, %s", dsPath, err)
 	}
 
-	//removes container from map
-	containersLock.Lock()
-	delete(containers, ParseID(c.ExecConfig.ID))
-	containersLock.Unlock()
-
+	//remove container from cache
+	containers.Remove(c.ExecConfig.ID)
 	return nil
 }
 
 func (c *Container) Update(ctx context.Context, sess *session.Session) (*metadata.ExecutorConfig, error) {
+	defer trace.End(trace.Begin("Container.Update"))
 	c.Lock()
 	defer c.Unlock()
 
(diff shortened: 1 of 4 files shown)


lib/portlayer/exec/container.go
```diff
@@ -33,18 +33,16 @@ import (
...
-	containers = make(map[ID]*Container)
+	NewContainerCache()
 }
 
 type State int
 
 const (
 	StateRunning = iota + 1
 	StateStopped
+	StateSuspended
```
corrieb: Does this have anything to do with the change set, or has this crept in by mistake?

lib/portlayer/exec/container_cache.go
```diff
@@ -0,0 +1,90 @@
...
+* ContainerCache will provide an in-memory cache of containerVMs.  It will
+* be refreshed on portlayer start and updated via container lifecycle
+* operations (start, stop, rm) and well as in response to infrastructure
+* events
+ */
+type containerCache struct {
+	m sync.RWMutex
+
+	// cache maps containerVM to container ID && infrastructure ID
+	cacheByID      map[string]*Container
```
hmahmood: Are we also going to cache by truncated it? This has sometimes tripped us up in the past.

corrieb: I wonder if the two maps is more complicated than it needs to be. Is there a disadvantage to having a single map and having two keys point to the same container? it would make the lookup more simple. In the remove case, it would be a single lookup followed by an optional cleanup.

sflxn: @hmahmood If I understand your question correctly, I don't think the id are truncated.  Before ids get to portlayer, the personality server gets the full id and use that for calls to portlayer.

lib/portlayer/exec/container.go
cgtexmex: It's got nothing to do with this change, but will have something to do with the next change -- so I can remove until then if you prefer

lib/portlayer/exec/container_cache.go
cgtexmex: @hmahmood what @sflxn said -- the portlayer will always have the full container id

cgtexmex: @corrieb I don't see the two maps as being more complicated -- from my perspective it provides clarity of what's in the cache since there's a segregation.  Also this follows a similar pattern in the persona code for the image cache.  With that said I'm fine with either implementation.

lib/portlayer/exec/container.go
cgtexmex: For the sake of clarity I removed from this change

lib/portlayer/exec/container_cache.go
```diff
@@ -0,0 +1,90 @@
...
+	// find by id
+	container := cache.cacheByID[idOrRef]
+	// if not found by id find by ref
+	if container == nil {
+		container = cache.cacheByInfraID[idOrRef]
+	}
+	return container
+}
+
+func (cache *containerCache) Add(container *Container) {
```
hmahmood: This really should be called Put since we are overwriting what may be previously cached.

lib/portlayer/exec/container_cache.go
hmahmood: I kind of like @corrieb suggestion, it simplifies the code, and there is next to nil chance of a clash between id and reference string.

lib/portlayer/exec/container_cache.go
cgtexmex: Done

lib/portlayer/exec/container_cache.go
cgtexmex: I've converted the cache to a single map..

lib/portlayer/exec/container_cache.go
hmahmood: I am trying to look at it sideways, but the diff still says `Add` :-)

cgtexmex: hmmm...apparently jackaroo'd something...I shall return!

lib/portlayer/exec/container_cache.go
corrieb: @hmahmood "next to nil chance of a clash between id and reference string". It's unlikely, but getting ill-defined or incorrect behavior if you name a container "vm-15" is a hole we don't need to intentionally open up.

@cgtexmex Wasn't requesting you change the code, rather just wanting to understand the choice behind the design. The point Hasan raises may be the very reason to have the two hashtables.

I would like to make sure that if someone does choose to name a container the same string as an existing moref - however unlikely - that we don't return the wrong container. In some respects, the double hashtable implementation makes it easier to avoid that kind of clash.

@cgtexmex can you please add a unit test to verify that naming a container the same as an existing moref gives the expected result. Thx

hmahmood: I based my comment on us caching ids -- which are randomly generated so have almost zero chance of collision with a moref string.

cgtexmex: We are storing id -- which is randomly generated and follows the the regex of [a-z0-9] -- originally I was checking via regex to classify as either id or ref and store in the appropriate map.  This is not the friendly name that a user can specify...

The moref will be the full moref so Object:ID -- "VirtualMachine:vm-100" -- again I think that a collision is doubtful.

hmahmood:
lgtm

sflxn:
lgtm

lib/portlayer/exec/container_cache.go
corrieb: Sure - ID is a different story than name. Thanks for clarifying

lib/portlayer/exec/container_cache.go
```diff
@@ -0,0 +1,81 @@
...
+
+// puts a container in the cache and will overwrite an existing container
+func (conCache *containerCache) Put(container *Container) {
+	conCache.m.Lock()
+	defer conCache.m.Unlock()
+
+	// add pointer to cache by container ID
+	conCache.cache[container.ExecConfig.ID] = container
+	//if we have a moref then add to infra ID
+	if container.vm != nil {
```
corrieb: Is there any circumstance in which a container will only be cached under its ID (the nil VM case above), or is that an error scenario? Reason I ask is that Put() has two behaviors that is not under user control - cache under ID, or cache under ID and Ref. If Remove() is called with a Ref, it won't work for containers that were cached under ID only.

cgtexmex: A long term nil vm would be an error scenario -- so the ID would be cached and a moref would never exist for the failed create operation.  

I think this is fine for now, but when I actually plumb the event stream in we'll need to push on it a bit more.

PR #1576, opened 2016-07-20
Add VCH resource limits

### LINKED ISSUE #1538: vic-machine: resource pool limits
Provide arguments to configure the VCH size (resource pool reservations, limits, and shares) when deploying a VCH:

```
--memory-limits=reservation:limit
--cpu-limits=reservation:limit
```
e.g. specify both reservation and limit, and shares
```
--memory-limits=800MB:8GB
--memory-shares=2000
```

e.g. specify only limit (note the leading ':')
```
--cpu-limits=:20GHz
```

e.g. specify only reservation (note the trailing ':')
```
--cpu-limits=500MHz:
```

On completion vic-machine should configure the resource pool created for the VCH with reservation and limits specified. It does *NOT* need to sanity check the figures against the parent pool configuration initially.

emlin:
Fixes #1538 

  Allow user to specify reservations/limits/shares for VCH CPU and
  Memory.

  Cpu should end with MHz or GHz, case-insensitive.
  Memory should end with MB or GB, case-insensitive.
  Shares only accept numbers, and custom level will always be used.


cmd/vic-machine/create/reservations.go
```diff
@@ -0,0 +1,169 @@
...
+
+func (c *Create) handleCPUReservation(m string) error {
+	defer trace.End(trace.Begin(m))
+	if m == "" {
+		return nil
+	}
+
+	var errs []string
+
+	elements := strings.Split(m, ":")
```
hmahmood: Possible to do a regex for these?

emlin: regex will be more efficient to find error, but error message will be only format error.
let me try that way to see the result.

andrewtchin:
lgtm

dougm:
For your consideration, the govc pool.change command uses separate flags:
```
  -cpu.expandable=<nil>                            CPU expandable reservation
  -cpu.limit=0                                     CPU limit in MHz
  -cpu.reservation=0                               CPU reservation in MHz
  -cpu.shares=                                     CPU shares level or number
  ...
  -mem.expandable=<nil>                            Memory expandable reservation
  -mem.limit=0                                     Memory limit in MB
  -mem.reservation=0                               Memory reservation in MB
  -mem.shares=                                     Memory shares level or number
```

Should be simpler for the code and the user to have separate flags.

dougm:
Why do we support multiple units here (MB, GB), when appliance memory for example only support MB?
For now, I suggest we only support MB and MHz for these flags.  Later we could add generic flags (implementing the flag.Value interface) that can support unit parsing and apply them to these flags, appliance-memory flag, etc.

emlin:
> Why do we support multiple units here (MB, GB), when appliance memory for example only support MB?
> For now, I suggest we only support MB and MHz for these flags. Later we could add generic flags (implementing the flag.Value interface) that can support unit parsing and apply them to these flags, appliance-memory flag, etc.

I think as @hickeng said, we plan to remove those two flags for appliance memory and cpu.
I didn't do that in this PR, is because no one confirm on that yet.

For the combined flag design, talk to @hickeng

dougm:
ok, I hadn't seen the spec in the linked issue.  We can discuss the UX later I suppose, if this functionality is required for onsite then lgtm.

In any case, would still prefer to separate the unit related flags, though it can happen in a future PR.  We can use the existing docker/go-units or govmomi/units for memory.  Not sure if there's one already for hz, but we could add one.  Again, to help keep our code simpler and UX (flags) consistent.

emlin:
@dougm Only thing available currently is the description in #1538, will have these comments tracked in that issue. thanks

caglar10ur:
I second @dougm comment, we should stick to single unit for every parameter (M is granular enough) or use *units packages instead of doing regex and also have seperate flags
@hickeng is there a reason for you to want them this way?

emlin:
As you two both think of this way, and I totally agree with you. 
@dougm @caglar10ur @hickeng, so I gonna change based on this comment tomorrow

emlin:
Based on the above comments, I change the command line to following format:
```
   --pool-memory-reservation value     VCH Memory reservation in MB (default: 0)
   --pool-memory-limit value           VCH Memory limit in MB (default: 0)
   --pool-memory-shares value          VCH Memory shares in level or share number, e.g. high, normal, low, or 163840 (default: normal)
   --pool-cpu-reservation value        VCH vCPUs reservation in MHz (default: 0)
   --pool-cpu-limit value              VCH vCPUs limit in MHz (default: 0)
   --pool-cpu-shares value             VCH vCPUs shares, in level or share number, e.g. high, normal, low, or 4000 (default: high)
```

The name was added with "pool" prefix to differentiate with --appliance-memory and --appliance-cpu. Those two parameters might be removed eventually, but before that happens, this prefix can avoid confusion.

emlin:
@dougm @caglar10ur @andrewtchin could you please have a look again?

lib/install/management/resource_pool.go
```diff
@@ -58,20 +58,45 @@ func (d *Dispatcher) createResourcePool(conf *metadata.VirtualContainerHostConfi
...
+		memory.Limit = settings.VCHSize.Memory.Limit
+	} else {
+		memory.Limit = -1
+	}
+	if settings.VCHSize.Memory.Reservation != 0 {
+		memory.Reservation = settings.VCHSize.Memory.Reservation
+	} else {
+		// FIXME: govmomi omitempty
+		memory.Reservation = 1
+	}
```
caglar10ur: [cosmetic]

you can use

```
X = -1 
if ... != {
  X = ...
}
```
and got rid of all those else blocks

lib/install/management/virtual_app.go
```diff
@@ -38,20 +38,45 @@ func (d *Dispatcher) createVApp(conf *metadata.VirtualContainerHostConfigSpec, s
...
+		memory.Limit = settings.VCHSize.Memory.Limit
+	} else {
+		memory.Limit = -1
+	}
+	if settings.VCHSize.Memory.Reservation != 0 {
+		memory.Reservation = settings.VCHSize.Memory.Reservation
+	} else {
+		// FIXME: govmomi omitempty
+		memory.Reservation = 1
+	}
```
caglar10ur: same as ^

caglar10ur:
some cosmetic nitpicks otherwise LGTM, thanks Emma!

lib/install/management/resource_pool.go
emlin: Good suggestion. thanks

dougm:
nice, lgtm

Issue #854, opened 2016-05-31
Docker API: docker port


hickeng:
Implementation of the docker port command. This data should be pulled from the network port layer, not from the container metadata - while the two should match (ports - IPs are not held in the metadata) the port layer pulls from the source of truth.

This ties in with #692

corrieb:
@hickeng I very much doubt this will make onsite

sflxn:
It will.  It's free.  Once we fill in the container info data structure on the portlayer side, the personality server's container inspect will return the correct values.  Docker port gets all the port data from container inspect.  It simply calls docker inspect.

sflxn:
I was mistaken.  It is not free.  This is now done.  When container inspect was first started, I thought we needed to get this information in the port layer.  After @hmahmood implemented port mapping, we had a discussion about how the mapping metadata was used, and I also looked at the docker code for inspect again.  We don't really need to get any of this data from the portlayer at all.

mdubya66:
GA feature from SSOD. needs to be tested.

mdubya66:
I see this is closed. So docker port feature is working? is there a test? @sflxn @hickeng

hmahmood:
Seems to be working:

```
hmahmood@ubuntu:~/go/src/github.com/vmware/vic$ docker -H 172.16.127.243:2376 --tls port insane_swirles
80/tcp -> 0.0.0.0:5003
```

PR #1577, opened 2016-07-20
Implement port mapping

### LINKED ISSUE #938: Implement port mapping for containers
This would enable the docker use case where containers map ports through the docker server. This can be done by implementing the correct NAT rules on the VCH and container side. This would also mean assigning an IP address to the VCH on the relevant bridge network(s).

Acceptance:
* VCH has an IP address on the relevant bridge network(s)
* `docker create -p/-P` works

hmahmood:
Fixes #938


jooskim:
Do we now have port mapping info available to UI? Or can I access port mapping info through mob interface?

isos/appliance/nat-setup
```diff
@@ -58,15 +56,30 @@ iptables-restore <<RULES
...
+-A INPUT -p tcp -m tcp --dport 22 -j ACCEPT
+-A INPUT -p icmp -m icmp --icmp-type 8 -j ACCEPT
+-A INPUT -p udp -m udp --dport 68 -j ACCEPT
+-A INPUT -p tcp -m tcp --dport 2378 -j ACCEPT
+-A INPUT -p tcp -m tcp --dport 2376 -j ACCEPT
+-A INPUT -i bridge -p udp -m udp --dport 53 -j ACCEPT
+-A INPUT -i bridge -p tcp -m tcp --dport 53 -j ACCEPT
+-A INPUT -p tcp -m tcp --dport 8080 -j ACCEPT
+-A INPUT -p udp -j REJECT --reject-with icmp-port-unreachable
+-A INPUT -p tcp -j REJECT --reject-with tcp-reset
```
andrewtchin: 👏

lib/apiservers/engine/backends/portmap/portmap.go
```diff
@@ -0,0 +1,170 @@
...
+
+type Operation int
+
+const (
+	Map Operation = iota
+	Unmap
+)
+
+type PortMapper interface {
+	MapPort(op Operation, ip net.IP, port int, proto string, destIP string, destPort int, strcIface, destIface string) error
```
andrewtchin: is `strcIface` a typo?

hmahmood:
@jooskim port mapping info is not available in the guestinfo, only the exposed ports are. So, for instance, if a user specifies `-p 8080:80` during `docker create/run`, only `80` would be available in guestinfo. The command `docker ports` will give you the full info though, and I think that is still under development here #854

jooskim:
@hmahmood Ok thanks. This is just out of curiosity, but what happens if more than one port is exposed to host? like 80 and 443. Do we get both ports in guestinfo?

hmahmood:
@jooskim yes you will get both ports in guestinfo.

lib/apiservers/engine/backends/portmap/portmap.go
hmahmood: Yes, I thought I had fixed that :-/

Makefile
```diff
@@ -282,7 +282,8 @@ $(iso-base): isos/base.sh isos/base/*.repo isos/base/isolinux/** isos/base/xorri
 	@$(TIME) $< -c $(BIN)/yum-cache.tgz -p $@
 
 # appliance staging - allows for caching of package install
-$(appliance-staging): isos/appliance-staging.sh $(iso-base)
+$(appliance-staging): isos/appliance-staging.sh $(i
+so-base)
```
caglar10ur: I think this commit escaped from a jail :)

isos/appliance/nat-setup
```diff
@@ -43,11 +43,9 @@ echo 2 > /proc/sys/net/ipv4/ip_dynaddr
 # too many routers are still ignorant
 echo 0 > /proc/sys/net/ipv4/tcp_ecn
 
-# wait for the bridge interface to exist - 100ms wait
-until ip link show dev bridge;do sleep 0.1;done
-
 # blow away any existing rules with iptables-restore
 iptables-restore <<RULES
+# Generated by iptables-save v1.6.0 on Tue Jul 19 22:57:40 2016
```
caglar10ur: do we care about those headers ^v?

caglar10ur:
LGTM :sheep: :it:

andrewtchin:
lgtm

hmahmood:
:recycle:

PR #1587, opened 2016-07-20
Move to keyword so that we can use the regression test as a quick val…

mhagen-vmware:


emlin:
Well, just curious, can't we make the test case one keyword, in just it's original place?

emlin:
LGTM to the code

mhagen-vmware:
1. Then any test that wants to use it would need to pull in the regression robot file as a resource.
2. You can't have a test cases table in a resource.

andrewtchin:
lgtm

PR #1589, opened 2016-07-20
Adding option to vic-machine to specify default bridge network range

hmahmood:


andrewtchin:
lgtm

lib/portlayer/network/network.go
```diff
@@ -40,19 +40,27 @@ func (e DuplicateResourceError) Error() string {
 func Init() error {
 	var err error
 
-	bridgeRange := net.IPNet{
-		IP:   net.IPv4(172, 17, 0, 0),
-		Mask: net.CIDRMask(12, 32),
+	bridgeRange := Config.BridgeIPRange
+	if bridgeRange == nil || len(bridgeRange.IP) == 0 || bridgeRange.IP.IsUnspecified() {
+		_, bridgeRange, err = net.ParseCIDR("172.16.0.0/12")
```
caglar10ur: We go from .17 to .16, guessing that was intentional?

caglar10ur:
LGTM :sheep: :it:

lib/portlayer/network/network.go
hmahmood: Actually in this case it doesn't matter since the mask is /12; the network will still be 172.16.0.0.

mdubya66:
lgtm

hmahmood:
♻️

PR #1605, opened 2016-07-21
Only cache container when it's committed

### LINKED ISSUE #1603: Container shouldn't be added to PL cache until committed
The container creation process is a multi-step / call process between the docker persona and the PortLayer.  During the initial call of this process a container is added to the PL container cache prior to it being created in the infrastructure.  If the creation process fails on a later step the container will be orphaned in the cache.

A container should only be added to the cache when the creation process is complete via a successful commit call.

cgtexmex:
Change will only place a container into the container cache
when the containerVM has been created.

Fixes #1603

### DIFF
--- a/lib/portlayer/exec/container.go
+++ b/lib/portlayer/exec/container.go
@@ -66,10 +66,6 @@ func NewContainer(id ID) *Handle {
 		State:      StateStopped,
 	}
 	con.ExecConfig.ID = id.String()
-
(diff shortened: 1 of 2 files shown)


caglar10ur:
LGTM :sheep: :it:

jzt:
lgtm 👍

cgtexmex:
Rebase

PR #1606, opened 2016-07-21
The results of ps command should be sorted by create date

### LINKED ISSUE #1334: ps command should sort by create date
The ps command should sort by create date in reverse order.

cgtexmex:
Change will reverse sort results of ps -- newest will appear at the
top of the list.

Fixes #1334

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -19,6 +19,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -77,6 +78,14 @@ func init() {
 	portMapper = portmap.NewPortMapper()
 }
 
+//TODO: gotta be a better way...
```
caglar10ur: nope, that's the best way :)

caglar10ur:
LGTM :sheep: :it:

lib/apiservers/engine/backends/container.go
jzt: That's The Docker Way™!

jzt:
LGTM

cgtexmex:
REBASE

PR #1617, opened 2016-07-21
Add VolumeStoresList to volume layer

### LINKED ISSUE #1506: Add ListStores to volume layer of port layer
Return a list of volume store names.

fdawg4l:
NOTE:  TO BE MERGED POST BETA.

Fixes #1506 

* Added handler for port layer rest API
* Also fix sync() issues in unit test as well as detach semantics.


lib/apiservers/portlayer/swagger.yml
```diff
@@ -207,6 +207,29 @@ paths:
...
+      responses:
+        '200':
+          description: "OK"
+          schema:
+            $ref: "#/definitions/VolumeStoresListResponse"
+        '404':
+          description: "Not found"
+          schema:
+            $ref: "#/definitions/Error"
+        default:
```
hmahmood: I would prefer a status code here, and not a catch all. '500' should do the trick.

fdawg4l: Sounds good.  Will do.

lib/portlayer/storage/vsphere/volume.go
```diff
@@ -87,6 +87,28 @@ func (v *VolumeStore) AddStore(ctx context.Context, ds *datastore.Helper, storeN
 	return u, nil
 }
 
+func (v *VolumeStore) VolumeStoresList(ctx context.Context) (map[string]url.URL, error) {
+	m := make(map[string]url.URL)
+	for u, ds := range v.ds {
+
+		// Get the ds:// URL given the datastore url ("[datastore] /path")
+		dsurl, err := datastore.ToURL(ds.RootURL)
```
matthewavery: In order to avoid copies of "ds" you could replace `ds.RootURL` with v[u].RootURL

matthewavery:
lgtm, only one small comment. Not really a big issue.

lib/portlayer/storage/vsphere/volume.go
fdawg4l: @matthewavery are you concerned about a pointer?  ds is a map  of the following type

```
39 »···// maps datastore uri (volume store) to datastore
40 »···ds map[url.URL]*datastore.Helper 
```Copying of a pointer is trivial unless youre concerned about something else.  

Also, about your proposed solution;  in general, avoid iterating maps whenever possible. But if that can't be avoided (like here), don't then use the result of the iteration to once again look the iterated key up in the map.

fdawg4l:
.

caglar10ur:
LGTM :sheep: :it:

PR #1622, opened 2016-07-22
Enable sshd again on VCH (with default pw: changeme)

caglar10ur:
This reverts @corrieb's commit which titled as 

```
Temporarily remove default SSH access from appliance for TP3 (#1456)
```

### DIFF
--- a/isos/appliance-staging.sh
+++ b/isos/appliance-staging.sh
@@ -110,11 +110,8 @@ yum_cached -p $PKGDIR clean all
 # configure us for autologin of root
 #COPY override.conf $ROOTFS/etc/systemd/system/getty@.service.d/
 # HACK until the issues with override.conf above are dealt with
-pwhash=$(openssl passwd -1 -salt vic password)
+pwhash=$(openssl passwd -1 -salt vic changeme)
 sed -i -e "s/^root:[^:]*:/root:${pwhash}:/" $(rootfs_dir $PKGDIR)/etc/shadow
-
-# 1218: Temporarily disable SSH for TP3
-rm $(rootfs_dir $PKGDIR)/usr/lib/systemd/system/sshd@.service
 
 # Allow root login via ssh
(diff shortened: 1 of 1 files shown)


kjplatz:
:+1: LGTM

hickeng:
A password of changeme without a mechanism for making a persistent change is going to lead to more than a little frustration. Why are we re-enabling this in base builds?
Is it viable to add an appliance-debug target that has ssh access configured by default?

caglar10ur:
@hickeng to enable debugging, I know you will say one need to use appliance-debug but that's not we build by default. It doesn't help when you want to grab a build from bintray and test it.

I was thinking we keep enable this for now, then add a Makefile var like PRODUCTION=true to disable it and build releases with that flag.

hmahmood:
Is debugging via the console not viable? I am leaning towards not enabling this. Primarily because the VCH can be on an external network, and we really should not have a guessable password without some mechanism to change it.

mdubya66:
would it be possible to have a flag in vic-machine to the VCH to enable debugging?

caglar10ur:
vic-machine is an option, alternatively @dougm suggested using toolbox for this. We could have "enable-debug" command which enables ssh, sets logging level to debug, enables pprof etc.

hickeng:
@caglar10ur ssh is currently re-enabled on the appliance as part of the change which drops systemd-networkd. Let's leave this PR open for how to do this safely/sanely in the shipping version.

isos/appliance-staging.sh
```diff
@@ -110,12 +110,9 @@ yum_cached -p $PKGDIR clean all
 # configure us for autologin of root
 #COPY override.conf $ROOTFS/etc/systemd/system/getty@.service.d/
 # HACK until the issues with override.conf above are dealt with
-pwhash=$(openssl passwd -1 -salt vic password)
+pwhash=$(openssl passwd -1 -salt vic changeme)
```
vburenin: Is this default password documented somewhere?

caglar10ur:
dropping this in lieu of #2555

Issue #925, opened 2016-06-01
vic-machine create: change datastore parameter to accept file directory


emlin:
Currently the image store parameter only accept vsphere datastore name, so images for different VCHs will share same directory.
This is good for image sharing, but we're still not ready for that function, e.g no reference count, no permission restriction, so we'd like to keep them separated, until the imagec are ready for sharing.
So far, the suggested configuration format is as following:
-image-store ds://datastore/dev/vch1

vic-machine should check if the directory exists or not. If the leaf directory does not exist, vic-machine will create that, but not for all directory hierarchy.
If directory exists, vic-machine need to check if the directory is empty. Otherwise, return failure.

This issue includes changes for image store, volume store and container store.

hickeng:
@fdawg4l to confirm what's needed on the portlayer side to treat this as a datastore path (i.e. datastore id and path) and not just a datastore name.

fdawg4l:
The bulk of the work to support this on the PL side is done.  It needs to be wired in where the image storer is instantiated.  This is identical to what we did for volumes.  Not a lot of work.  May not happen in the next 4 days.

hickeng:
vic-machine currently accepts paths, but while we can pull images with that config we cannot create containers from them.

fdawg4l:
@gigawhitlocks this is another one that #1678, i think, closes.

PR #1631, opened 2016-07-22
Allow custom paths for image store

gigawhitlocks:
Allows users to specify a specific directory in a datastore to use for their image store.
Only supports the first one specified, for the time being. TODO in place to that effect. Addresses #925 -- note that despite the title of the issue in question, the changes needed in vic-machine were already in place & so I just did the wiring in the port layer.


fdawg4l:
LGTM

hmahmood:
lgtm

gigawhitlocks:
Spoke with Emma out of band about this, needs some changes to `vic-machine delete` and one usage in `vic-machine create` to be complete. Those are pending.

PR #1636, opened 2016-07-22
Add port mapping to container inspect

### LINKED ISSUE #854: Docker API: docker port

sflxn:
Add port mapping to ContainerInspect so that 'docker port' will
display port mapping.  Also, docker inspect should display the
right port mapping data in HostConfig, Config, and NetworkSettings
section of the inspect json.

Resolves #854

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -41,6 +41,7 @@ import (
 	"github.com/docker/docker/pkg/version"
 	"github.com/docker/engine-api/types"
 	"github.com/docker/engine-api/types/container"
+	dnetwork "github.com/docker/engine-api/types/network"
 	"github.com/docker/engine-api/types/strslice"
 	"github.com/docker/go-connections/nat"
 
@@ -893,10 +894,13 @@ func (c *Container) ContainerInspect(name string, size bool, version version.Ver
 	defer trace.End(trace.Begin("ContainerInspect"))
 
 	// Look up the container name in the metadata cache to get long ID
-	if vc := cache.ContainerCache().GetContainer(name); vc != nil {
-		log.Debugf("Found %q in cache as %q", name, vc.ContainerID)
-		name = vc.ContainerID
+	vc := cache.ContainerCache().GetContainer(name)
+	if vc == nil {
+		return nil, derr.NewRequestNotFoundError(fmt.Errorf("No such container: %s", name))
 	}
+
+	log.Debugf("Found %q in cache as %q", name, vc.ContainerID)
+	name = vc.ContainerID
 
 	client := PortLayerClient()
 	if client == nil {
@@ -915,13 +919,18 @@ func (c *Container) ContainerInspect(name string, size bool, version version.Ver
 		}
 	}
 
-	inspectJSON, err := containerInfoToDockerContainerInspect(name, results.Payload)
+	inspectJSON, err := containerInfoToDockerContainerInspect(vc, results.Payload)
 	if err != nil {
 		log.Errorf("containerInfoToDockerContainerInspect failed with %s", err)
 		return nil, err
 	}
 
 	log.Debugf("ContainerInspect json config = %+v\n", inspectJSON.Config)
+	if inspectJSON.NetworkSettings != nil {
+		log.Debugf("Docker inspect - network settings = %#v", inspectJSON.NetworkSettings)
+	} else {
+		log.Debugf("Docker inspect - network settings = null")
+	}
 
 	return inspectJSON, nil
 }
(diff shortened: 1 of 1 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -1243,35 +1256,39 @@ func containerInfoToDockerContainerInspect(id string, info *models.ContainerInfo
...
+// hostConfigFromContainerInfo() gets the hostconfig that is passed to the backend during
+// docker create and updates any needed info
+func hostConfigFromContainerInfo(vc *viccontainer.VicContainer, info *models.ContainerInfo) *container.HostConfig {
+	if vc == nil || info == nil {
 		return nil
 	}
 
+	// Create a copy of the created container's hostconfig.  This is passed in during
+	// container create
+	hostConfig := *vc.HostConfig
```
hmahmood: Can vc.HostConfig be nil?

lib/apiservers/engine/backends/container.go
```diff
@@ -1384,16 +1358,13 @@ func mountsFromContainerInfo(id string, info *models.ContainerInfo) []types.Moun
...
 		return nil
 	}
 
-	container := &container.Config{
-		Domainname: "",    // Domainname
-		User:       "",    // User that will run the command(s) inside the container
-		StdinOnce:  false, // If true, close stdin after the 1 attached client disconnects.
-	}
+	// Copy the working copy of our container's config
+	container := *vc.Config
```
hmahmood: Same comment here: can vc.Config be nil?

lib/apiservers/engine/backends/container.go
```diff
@@ -1466,32 +1437,78 @@ func containerConfigFromContainerInfo(id string, info *models.ContainerInfo) *co
...
 }
 
+// portMapFromVicContainer() constructs a docker portmap from both the container's
+// hostconfig and config (both stored in VicContainer).  They are added and modified
+// during docker create.  This function creates a new map that is adhere's to docker's
+// structure for types.NetworkSettings.Ports.
+func portMapFromVicContainer(vc *viccontainer.VicContainer) nat.PortMap {
+	var portMap nat.PortMap
+
+	if vc != nil {
```
hmahmood:
```
if vc == nil {
  return portMap
}
```
will get rid of the indent further down.

lib/apiservers/engine/backends/container.go
```diff
@@ -915,13 +917,18 @@ func (c *Container) ContainerInspect(name string, size bool, version version.Ver
 		}
 	}
 
-	inspectJSON, err := containerInfoToDockerContainerInspect(name, results.Payload)
+	inspectJSON, err := containerInfoToDockerContainerInspect(vc, results.Payload)
```
hmahmood: I know `vc` is checked in `containerInfoToDockerContainerInspect`, but I think it would be better to check here as well, in case someone in the future adds code further down that uses `vc`.

lib/apiservers/engine/backends/container.go
```diff
@@ -1210,23 +1217,29 @@ func containerInfoToDockerContainerInspect(id string, info *models.ContainerInfo
...
-		NetworkSettings: networkFromContainerInfo(id, info),
+		Mounts:          mountsFromContainerInfo(vc, info),
+		Config:          containerConfigFromContainerInfo(vc, info),
+		NetworkSettings: networkFromContainerInfo(vc, info),
+	}
+
+	if inspectJSON.NetworkSettings != nil {
+		log.Info("Docker inspect - network settings = %#v", inspectJSON.NetworkSettings)
+	} else {
+		log.Info("Docker inspect - network settings = null")
```
jzt: Should these be `log.Debug`?

lib/apiservers/engine/backends/container.go
```diff
@@ -1243,35 +1256,39 @@ func containerInfoToDockerContainerInspect(id string, info *models.ContainerInfo
...
-			inpsectJSON.ID = *info.ContainerConfig.ContainerID
+			inspectJSON.ID = *info.ContainerConfig.ContainerID
 		}
 		if info.ContainerConfig.Created != nil {
-			inpsectJSON.Created = time.Unix(*info.ContainerConfig.Created, 0).String()
+			inspectJSON.Created = time.Unix(*info.ContainerConfig.Created, 0).String()
 		}
 		if len(info.ContainerConfig.Names) > 0 {
-			inpsectJSON.Name = info.ContainerConfig.Names[0]
+			inspectJSON.Name = info.ContainerConfig.Names[0]
```
jzt: Good catch! :)

lib/apiservers/engine/backends/container.go
sflxn: It really can't be nil, but I went ahead and put a check in.  If a container is in the cache, it had to have gone through ContainerCreate.

lib/apiservers/engine/backends/container.go
sflxn: sure.

lib/apiservers/engine/backends/container.go
sflxn: I added a check near the top of the function where we get the container.  It doesn't make sense for us to get down here if vc was nil.

lib/apiservers/engine/backends/container.go
sflxn: Changed to log.Debug.

hmahmood:
lgtm

jzt:
lgtm

caglar10ur:
LGTM :sheep: :it:

sflxn:
poke

PR #1639, opened 2016-07-25
Add pkg/version and use it with some cmd/

caglar10ur:
pkg/version adds -version flag to the caller so that we can get
detailed build info;

$ bin/imagec -version
v0.4.0 git:e422f47-dirty build:2016/07/25@05:13:55 runtime:go1.7rc3

$ bin/vicadmin -version
v0.4.0 git:e422f47-dirty build:2016/07/25@05:18:25 runtime:go1.7rc3

$ rpctool -version
v0.4.0 git:e422f47-dirty build:2016/07/25@05:15:07 runtime:go1.7rc3

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ define godeps
 endef
 
 define ldflags
-	$(shell $(BASE_DIR)/infra/scripts/version-linker-flags.sh)
+	$(shell BUILD_NUMBER=${BUILD_NUMBER} $(BASE_DIR)/infra/scripts/version-linker-flags.sh)
 endef
 
 # target aliases - environment variable definition
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ define godeps
 	$(wildcard $1) $(shell $(BASE_DIR)/infra/scripts/go-deps.sh $(dir $1) $(MAKEFLAGS))
 endef
 
-define flags
+define ldflags
 	$(shell $(BASE_DIR)/infra/scripts/version-linker-flags.sh)
 endef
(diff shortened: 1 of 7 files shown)


dougm:
Very nice.  Would be good to put the flags in a single place so it'll be easier to update.  Makefile variable or even a scripts/version-linker-flags.sh that outputs the flags?

caglar10ur:
I think reorganized it as you suggested @dougm

dougm:
lgtm

tho maybe 'ldflags' or 'version-ldflags' instead of 'flags'

caglar10ur:
done (flags -> ldflags)

dougm:
:sheep: :it:

emlin:
The change looks good, then if trigger build after version tag, we'll get a new version release in the binary.
Just a few questions.
1. Do we need drone build ID? For example 3308.
2. Do we think about consolidate this module version with vic-machine version? I understand that might have different requirement, but probably will introduce confusion if they are different.

caglar10ur:
@emlin we could add drone id here. We could also consolidate this with vic-machine, I just didn't want to touch it for now but I can take a look at it.

emlin:
@caglar10ur it's fine to leave vic-machine version command there.
Just if you have drone build included, I can make changes in vic-machine to leverage the flags and versions in vic-machine

caglar10ur:
@emlin please see the new revision

emlin:
nice. LGTM 👍

caglar10ur:
♻️

PR #1660, opened 2016-07-25
[imagec] set --insecure-skip-verify flag

caglar10ur:


mdubya66:
if i followed this correctly this was a regression. is there a test for this to prevent that from happening?

caglar10ur:
nope, we don't have one now. We can bring a registry on vic but the issue is lack of push like we discussed last week. Will work with @mhagen-vmware to figure out a way

fdawg4l:
LGTM

jzt:
lgtm

cgtexmex:
LGTM

PR #1662, opened 2016-07-25
unmount /run

### LINKED ISSUE #1643: unable to run postgres:9.4
For some reason we're missing `/run/postgresql`

A quick hunt suggests that 60a4daae58264b9e52dbadafdc809e2d27a899fc2a5840f67a610cdbcf234289 is the ID in which this directory can be found, however it doesn't exist in our containers.

My hypothesis on the basis of a quick inspection is that the systemctl switch-root that we're using to move into the container filesystem is purging /run (or /var/run - which is a symlink to /run).

caglar10ur:
systemd is not purging but just propagating the mount points into the new
environment so we end up with a tmpfs /run instead of Dockerfile created
(static) one.

IMHO that's docker bug. Their images should not assume /run/XYZ
directories existence as systemd clearly says "Applications must ensure
to recreate their own files/dirs on startup, and cannot rely that doing
this at package installation will suffice. It" but since fixing all
those images are not possible we simply unmount /run in tether.

Fixes #1643

### DIFF
--- a/lib/tether/ops_linux.go
+++ b/lib/tether/ops_linux.go
@@ -31,6 +31,7 @@ import (
 	"golang.org/x/net/context"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/docker/docker/pkg/mount"
 	"github.com/vishvananda/netlink"
 	"github.com/vmware/vic/lib/dhcp"
 	"github.com/vmware/vic/lib/dhcp/client"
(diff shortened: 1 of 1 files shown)


dougm:
lgtm

Assuming the syscall just returns an errno, EBUSY for example, might be good to add some context there: return fmt.Errorf("unmount /run: %s", err)

caglar10ur:
in theory it shouldn't return EBUSY as we are telling it to perform a lazy unmount. According to man page MNT_DETACH does following

```
Perform a lazy unmount: make the mount point unavailable for
              new accesses, immediately disconnect the filesystem and all
              filesystems mounted below it from each other and from the
              mount table, and actually perform the unmount when the mount
              point ceases to be busy.
```

will add fmt.Errorf in every case :)

hickeng:
What happens here if /run isn't mounted? That's an acceptable condition for us and is likely to be the case once we ditch systemd.

caglar10ur:
I guess we error out if /run is not mounted, I'll add a check to confirm we have /run before trying to unmount

caglar10ur:
@hickeng take a look again

caglar10ur:
@hickeng ping

hickeng:
LGTM

caglar10ur:
♻️

caglar10ur:
♻️

Issue #1612, opened 2016-07-21
Backend volume tests are hosed


jzt:
```
$ go test -v github.com/vmware/vic/lib/apiservers/engine/backends/...
# github.com/vmware/vic/lib/apiservers/engine/backends
lib/apiservers/engine/backends/volume_test.go:129: undefined: testCap
lib/apiservers/engine/backends/volume_test.go:133: undefined: testCap
lib/apiservers/engine/backends/volume_test.go:136: undefined: trestLabelMap
lib/apiservers/engine/backends/volume_test.go:146: undefined: metadataBefore
lib/apiservers/engine/backends/volume_test.go:146: multiple-value json.Marshal() in single-value context
FAIL    github.com/vmware/vic/lib/apiservers/engine/backends [build failed]
?       github.com/vmware/vic/lib/apiservers/engine/backends/cache      [no test files]
?       github.com/vmware/vic/lib/apiservers/engine/backends/container  [no test files]
?       github.com/vmware/vic/lib/apiservers/engine/backends/portmap    [no test files]
```

They aren't being caught by CI because the backend tests aren't running in CI. See #1613.

jzt:
Closed via #1664.

PR #1664, opened 2016-07-25
Fix docker daemon volumes unit test and persist docker volumes metadata

fdawg4l:
This test wasn't running because the directory was being skipped due to the Makefile.  Also the unit test itself didn't compile.  Fixed.

We weren't persisting the volumes metadata when creating the volume.  Fixed.


fdawg4l:
@matthewavery, can you take a look at this change when you get the chance?  Thanks!

caglar10ur:
LGTM :sheep: :it:

matthewavery:
lgtm, thanks for cleaning that up for me. I never liked that function name. Also yeah somehow that test made it into master in a weird state... I am thinking I forgot to push a final commit at some point since I use `go test` to check all my unit tests before a merge.

fdawg4l:
.

jzt:
Hey!  This fixes #1612!

PR #1666, opened 2016-07-25
Tidy up bootstrap script

caglar10ur:
Also decrease tmpfs size to 64MB from 128 and start copying kernel
module directory as bind mount is not working as expected after a
switch-root.


hmahmood:
lgtm

isos/bootstrap/bootstrap
```diff
@@ -43,7 +38,7 @@ if mount -t ext4 /dev/disk/by-label/containerfs ${MOUNTPOINT}; then
     # enable full system functionality in the container
     echo "Publishing modules within container"
     mkdir -p ${MOUNTPOINT}/lib/modules
-    mount --bind /lib/modules ${MOUNTPOINT}/lib/modules
+    cp -pr /lib/modules/* ${MOUNTPOINT}/lib/modules/
```
fdawg4l: Once the modules are `insmod`'ed, do we need them accessible on the filesystem anymore?

caglar10ur: We do if we want to provide more (like nfs/cifs mounting, enabling some other kernel feature etc)

fdawg4l:
lgtm

hickeng:
LGTM

caglar10ur:
♻️

Issue #1210, opened 2016-06-24
vic-machine: create parent directory for images and volumes


emlin:
Parse the file directory configured by user, and create root directory if does not exist.
Persist the root directory to guestinfo, so that vic-machine delete can use that.

If on VSAN, and the root directory is name space, persist the name space uuid, to make sure next time we can delete the directory successfully.

hickeng:
@stuclem @npakrasi For TP3 we should document behaviour when specifying a directory that doesn't exit, and one that does but isn't empty.
The critical point is what occurs when `vic-machine delete` is run as that cleans up images and volumes with the assumption that all files in the directories are fair game.

gigawhitlocks:
`vic-machine delete` feature is in PR #1410 -- parsing and persistence to GuestInfo at install time are already on master at this time. VSAN features I'm not sure about -- I think those are still TBD

fdawg4l:
We've abstracted the datastore type (vsan or otherwise) so that is no longer a concern.  For volumes, the path indicated for the volume store will either be used if it exists or created.  For images, this work is still TBD.

stuclem:
This issue is included as a known issue in the release notes for TP3.

corrieb:
Latest update from Faiyaz: "for clarity, the images directory on the datastore should be specified via vic-machine and used in the port layer like we are for volumes. volumes is A-OK". Sounds like this needs to remain open and move to onsite.

gigawhitlocks:
Is this issue able to be closed after my PR here https://github.com/vmware/vic/pull/1678 is merged?

fdawg4l:
@gigawhitlocks yes.

stuclem:
This is fixed. Removing kind/note.

PR #1678, opened 2016-07-26
Allow custom image store paths

gigawhitlocks:
Creating a new PR because this one is more expansive. Previous reviewers ( @hmahmood @emlin @fdawg4l ) may wish to take another gander.

This modifies the portlayer as well as vic-machine to allow custom image store locations. If one is not provided, a directory called images/ inside of the normal vch storage directory (where the ISOs go) is used

During delete, only the final directory in the path is removed -- during create, the path is created if it does not exist.

Comments more than welcome. Emma I expect you specifically to have some input about some functions / variables I've renamed.

Addresses the last of the functionality described in #1210

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -117,8 +117,8 @@ func (c *Create) Flags() []cli.Flag {
 		cli.StringFlag{
 			Name:        "image-datastore, i",
 			Value:       "",
-			Usage:       "REQUIRED. Image datastore name",
-			Destination: &c.ImageDatastoreName,
+			Usage:       "Image datastore path",
+			Destination: &c.ImageDatastorePath,
 		},
 		cli.StringFlag{
 			Name:        "container-datastore, cs",
@@ -309,8 +309,8 @@ func (c *Create) processParams() error {
 		return err
 	}
 
-	if c.ImageDatastoreName == "" {
-		return cli.NewExitError("--image-datastore Image datastore name must be specified", 1)
+	if c.ImageDatastorePath == "" {
+		return cli.NewExitError("--image-datastore Image datastore path must be specified", 1)
 	}
 
 	if c.cert != "" && c.key == "" {
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -115,9 +115,9 @@ func NewCreate() *Create {
 func (c *Create) Flags() []cli.Flag {
 	flags := []cli.Flag{
 		cli.StringFlag{
-			Name:        "image-datastore, i",
+			Name:        "image-store, i",
 			Value:       "",
-			Usage:       "Image datastore path",
+			Usage:       "Image datastore path in format \"datastore/path\"",
 			Destination: &c.ImageDatastorePath,
 		},
 		cli.StringFlag{
@@ -293,7 +293,7 @@ func (c *Create) processVolumeStores() error {
 	for _, arg := range c.volumeStores {
 		splitMeta := strings.SplitN(arg, ":", 2)
 		if len(splitMeta) != 2 {
-			return errors.New("Volume store input must be in format datastore-path:label")
+			return errors.New("Volume store input must be in format datastore-name/path:label")
 		}
 		c.VolumeLocations[splitMeta[1]] = splitMeta[0]
 	}
@@ -310,7 +310,7 @@ func (c *Create) processParams() error {
 	}
 
 	if c.ImageDatastorePath == "" {
-		return cli.NewExitError("--image-datastore Image datastore path must be specified", 1)
+		return cli.NewExitError("--image-store Image datastore path must be specified", 1)
 	}
 
 	if c.cert != "" && c.key == "" {
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -293,7 +293,7 @@ func (c *Create) processVolumeStores() error {
 	for _, arg := range c.volumeStores {
 		splitMeta := strings.SplitN(arg, ":", 2)
 		if len(splitMeta) != 2 {
-			return errors.New("Volume store input must be in format datastore-name/path:label")
+			return errors.New("Volume store input must be in format datastore/path:label")
 		}
 		c.VolumeLocations[splitMeta[1]] = splitMeta[0]
 	}
@@ -310,7 +310,7 @@ func (c *Create) processParams() error {
 	}
 
 	if c.ImageDatastorePath == "" {
-		return cli.NewExitError("--image-store Image datastore path must be specified", 1)
+		return cli.NewExitError("--image-store Image datastore path must be specified; use format datastore/path", 1)
 	}
 
 	if c.cert != "" && c.key == "" {
--- a/doc/user_doc/vic_installation/vch_installer_examples.md
+++ b/doc/user_doc/vic_installation/vch_installer_examples.md
@@ -21,7 +21,7 @@ You can install vSphere Integrated Containers directly on an ESXi host that is n
 <pre>vic-machine<i>-darwin</i><i>-linux</i><i>-windows</i> create
 --target <i>esxi_host_IPv4_address_or_FQDN</i>
 --user root
---image-datastore <i>datastore_name</i>
+--image-store <i>datastore_name</i>
 </pre>
 
 <a name="cluster"></a>
@@ -38,7 +38,7 @@ In addition to the mandatory options for deployment to a cluster, this example s
 --user Administrator@vsphere.local
 --password <i>vcenter_sso_password</i>
 --compute-resource <i>cluster_name</i>
---image-datastore <i>datastore_name</i>
+--image-store <i>datastore_name</i>
 --bridge-network <i>distributed_port_group_name</i>
 --name vch1
 </pre>
@@ -61,7 +61,7 @@ In addition to the mandatory options for deployment to a cluster, this example s
 --user Administrator@vsphere.local
 --password <i>vcenter_sso_password</i>
 --compute-resource <i>cluster_name</i>
---image-datastore <i>datastore_name</i>
+--image-store <i>datastore_name</i>
 --bridge-network <i>network_1_name</i>
 --management-network <i>network_2_name</i>
 --external-network <i>network_3_name</i>
@@ -80,7 +80,7 @@ In addition to the mandatory options for deployment to a cluster, this example s
 --user Administrator@vsphere.local
 --password <i>vcenter_sso_password</i>
 --compute-resource <i>cluster_name</i>
---image-datastore <i>datastore_1_name</i>
+--image-store <i>datastore_1_name</i>
 --container-datastore <i>datastore_2_name</i>
 --bridge-network <i>network_1_name</i>
 --name vch1
@@ -98,7 +98,7 @@ In addition to the mandatory options, this example specifies the vCenter Single 
 --user Administrator@vsphere.local
 --password <i>vcenter_sso_password</i>
 --compute-resource <i>host_IPv4_address_or_FQDN</i>
---image-datastore <i>datastore_name</i>
+--image-store <i>datastore_name</i>
 --bridge-network <i>network_name</i>
 --name vch1
 </pre>
@@ -114,7 +114,7 @@ This example uses the minimum required options.
 --target <i>esxi_host_IPv4_address_or_FQDN</i>
 --user root
 --compute-resource <i>resource_pool_name</i>
---image-datastore <i>datastore_name</i>
+--image-store <i>datastore_name</i>
 </pre>
 
 <a name="rp_cluster"></a>
@@ -129,7 +129,7 @@ In addition to the mandatory options, this example sets the vCenter Single Sign-
 --user Administrator@vsphere.local
 --password <i>vcenter_sso_password</i>
 --compute-resource <i>cluster_name</i>/<i>resource_pool_name</i>
---image-datastore <i>datastore_name</i>
+--image-store <i>datastore_name</i>
 --bridge-network <i>network_name</i>
 --name vch1
 </pre>
@@ -146,7 +146,7 @@ In addition to the mandatory options for deployment to a cluster, this example s
 --user Administrator@vsphere.local
 --password <i>vcenter_sso_password</i>
 --compute-resource <i>cluster_name</i>
---image-datastore <i>datastore_name</i>
+--image-store <i>datastore_name</i>
 --bridge-network <i>network_name</i>
 --cert <i>path_to_certificate_file</i>
 --key <i>path_to_certificate_file</i>
@@ -165,7 +165,7 @@ In addition to the mandatory options for deployment to a cluster, this example s
 --user Administrator@vsphere.local
 --password <i>vcenter_sso_password</i>
 --compute-resource <i>cluster_name</i>
---image-datastore <i>datastore_name</i>
+--image-store <i>datastore_name</i>
 --bridge-network <i>network_name</i>
 --name vch1
 --no-tls
@@ -183,7 +183,7 @@ In addition to the mandatory options for deployment to a cluster, this example s
 --user Administrator@vsphere.local
 --password <i>vcenter_sso_password</i>
 --compute-resource <i>cluster_name</i>
---image-datastore <i>datastore_name</i>
+--image-store <i>datastore_name</i>
 --bridge-network <i>network_name</i>
 --name vch1
 --appliance-cpu <i>number_of_CPUs</i>
--- a/doc/user_doc/vic_installation/vch_installer_examples.md
+++ b/doc/user_doc/vic_installation/vch_installer_examples.md
@@ -86,6 +86,19 @@ In addition to the mandatory options for deployment to a cluster, this example s
 --name vch1
 </pre>
 
+It is also possible to specify a specific folder in which to store your images, instead of just a datastore. This can be done by providing a path following the `image-store` option:
+
+<pre>vic-machine<i>-darwin</i><i>-linux</i><i>-windows</i> create
+--target <i>vcenter_server_IPv4_address_or_FQDN</i>
+--user Administrator@vsphere.local
+--password <i>vcenter_sso_password</i>
+--compute-resource <i>cluster_name</i>
+--image-store <i>datastore_1_name/path/to/some/specific/location</i>
+--container-datastore <i>datastore_2_name</i>
+--bridge-network <i>network_1_name</i>
+--name vch1
+</pre>
+
 <a name="standalone"></a> 
 ## Deploy a Virtual Container Host on a Standalone Host in vCenter Server ##
 
--- a/doc/user_doc/vic_installation/vch_installer_options.md
+++ b/doc/user_doc/vic_installation/vch_installer_options.md
@@ -46,13 +46,12 @@ You can also specify the username in the URL that you pass to `vic-machine creat
 
 Short name: `-i`
 
-The datastore in which to store container image files. When you deploy a virtual container host, `vic-machine` creates a folder named `VIC` on the target datastore,  in which to store all of the container images that you pull into a virtual container host. The `vic-machine` utility also places the VM files for the virtual container host in the datastore that you designate as the image store, in a folder that has the same name as the virtual container host.
+The datastore in which to store container image files. When you deploy a virtual container host, `vic-machine` creates a folder named `VIC` on the target datastore,  in which to store all of the container images that you pull into a virtual container host. The `vic-machine` utility also places the VM files for the virtual container host in the datastore that you designate as the image store, in a folder that has the same name as the virtual container host. If you specify an image store in the format `datastore/some/path`, `/some/path` will be used as your image store instead of the folder with the same name as the virtual container host.
 
 You can designate the same datastore as the image store for multiple virtual container hosts. In this case, only one `VIC` folder is created in the datastore and the container image files are made available to all of the virtual container hosts that use that image store. 
 
 **NOTES**: 
 - vSphere Integrated Containers supports all alphanumeric characters, hyphens, and underscores in datastore paths and datastore names.
-- In the current builds the `image-store` option does not support datastore folders. If you specify a datastore folder in the `image-store` option, `vic-machine` does not return an error, but creates all of the necessary folders at the root level of the datastore.
 - In the current builds the `container-datastore` option is not enabled. As a consequence, container VM files are also stored in the datastore that you designate as the image store.
 
 <pre>--image-store <i>datastore_name</i></pre> 
--- a/doc/user_doc/vic_installation/vch_installer_options.md
+++ b/doc/user_doc/vic_installation/vch_installer_options.md
@@ -42,7 +42,7 @@ If you are deploying vSphere Integrated Containers on vCenter Server, specify a 
 You can also specify the username in the URL that you pass to `vic-machine create` in the `target` option.
 
 <a name="image"></a>
-### `image-datastore` ###
+### `image-store` ###
 
 Short name: `-i`
 
@@ -52,10 +52,10 @@ You can designate the same datastore as the image store for multiple virtual con
 
 **NOTES**: 
 - vSphere Integrated Containers supports all alphanumeric characters, hyphens, and underscores in datastore paths and datastore names.
-- In the current builds the `image-datastore` option does not support datastore folders. If you specify a datastore folder in the `image-datastore` option, `vic-machine` does not return an error, but creates all of the necessary folders at the root level of the datastore.
+- In the current builds the `image-store` option does not support datastore folders. If you specify a datastore folder in the `image-store` option, `vic-machine` does not return an error, but creates all of the necessary folders at the root level of the datastore.
 - In the current builds the `container-datastore` option is not enabled. As a consequence, container VM files are also stored in the datastore that you designate as the image store.
 
-<pre>--image-datastore <i>datastore_name</i></pre> 
+<pre>--image-store <i>datastore_name</i></pre> 
 
 <a name="bridge"></a>
 ### `bridge-network` ###
@@ -208,9 +208,9 @@ The `vic-machine` utility allows you to specify the datastores in which to store
 - vSphere Integrated Containers fully supports VMware Virtual SAN datastores. 
 - vSphere Integrated Containers supports all alphanumeric characters, hyphens, and underscores in datastore paths and datastore names.
 
-### `image-datastore` ###
+### `image-store` ###
 
-See [image-datastore](#image) in the section on mandatory options.
+See [image-store](#image) in the section on mandatory options.
 
 ### `container-datastore` ###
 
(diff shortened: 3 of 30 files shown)


lib/install/management/store_files.go
```diff
@@ -53,15 +53,37 @@ func (d *Dispatcher) DeleteStores(vchVM *vm.VirtualMachine, conf *metadata.Virtu
 	var emptyImages bool
 	var emptyVolumes bool
 	log.Infof("Removing images")
-	if emptyImages, err = d.deleteImages(ds, p); err != nil {
-		errs = append(errs, err.Error())
+	for _, imageDir := range conf.ImageStores {
```
emlin: These logic all delete images function related, please move them to deleteImages method.

lib/install/management/store_files.go
```diff
@@ -53,15 +53,37 @@ func (d *Dispatcher) DeleteStores(vchVM *vm.VirtualMachine, conf *metadata.Virtu
...
+			errs = append(errs, err.Error())
+			continue
+		}
+
+		if len(imageDSes) != 1 {
+			errs = append(errs, fmt.Sprintf("Invalid or ambiguous datastore name %s provided while attempting to remove image stores", imageDir.Host))
+			continue
+		}
+
+		if emptyImages, err = d.deleteParent(imageDSes[0], imageDir.Path); err != nil {
```
emlin: As the name show, deleteParent is supposed to delete parent directory only.
Here you should use deleteDatastoreFiles directly

lib/install/management/store_files.go
```diff
@@ -53,15 +53,37 @@ func (d *Dispatcher) DeleteStores(vchVM *vm.VirtualMachine, conf *metadata.Virtu
...
+			errs = append(errs, err.Error())
+		}
+
+		if !emptyImages {
+			log.Infof("Not deleting [%s] %s as it still contains files after removing images", imageDir.Host, imageDir.Path)
+			continue
+		}
+
+		if _, err = d.deleteParent(imageDSes[0], imageDir.Path); err != nil {
```
emlin: As you already tried to delete everything under image store path, you should not call deleteParent here again, it will generate delete failure error message for file not found.

emlin:
Another thing is about test for vic-machine. The simulator function is still not ideal at this time, so more integration test is expected.
Could you please look at test plan https://github.com/vmware/vic/blob/master/tests/test-cases/Group6-VIC-Machine/6-6-Create-Datastore.md for datastore related functions in vic-machine create/delete, and add few more test cases?
That would help to make sure all of those functions works as expectation.
thanks

lib/install/management/store_files.go
```diff
@@ -53,15 +53,37 @@ func (d *Dispatcher) DeleteStores(vchVM *vm.VirtualMachine, conf *metadata.Virtu
 	var emptyImages bool
 	var emptyVolumes bool
 	log.Infof("Removing images")
-	if emptyImages, err = d.deleteImages(ds, p); err != nil {
-		errs = append(errs, err.Error())
+	for _, imageDir := range conf.ImageStores {
+		imageDSes, err := d.session.Finder.DatastoreList(d.ctx, imageDir.Host)
```
hickeng: @fdawg4l has some utility functions to create a URL from ds & path, and visa versa. I'd recommend using those - extending them if necessary.
The specific reason for this is the case of special characters in the datastore name (something I'd overlooked when suggesting using the Host field for the datastore) - at some point we may well change how dspaths are packed so using a utility function to extract datastore from URL and path from URL means only a single location to change if/when we change packing.

gigawhitlocks: @hickeng at this point in the code we already have the URL extracted and stored. It's extracted using a helper method, but not one of Faiyaz's (which actually I wrote the majority of) because Faiyaz's code goes from URL format to vSphere datastore formate( `ds://host/path`  to `[host] path` ) as strings on either end or "[host] path" string to `url.URL` -- the helper that goes from `ds://host/path` (string) to url.URL is a vic-machine specific one -- the one you helped me edit (`dsHelper` I believe is the name) so I'm not entirely sure to what you're referring in this case.

gigawhitlocks: @hickeng I actually wrote (most) of those functions that Faiyaz started -- I haven't changed how datastore URLs are parsed in vic-machine, although I do agree that those should be updated to use generic parsing functions (if not the ones that I wrote, another one that uses the same regexes to go from `ds://host/path` format to `url.URL` because that function doesn't exist currently) but if that needs to be done I'd like to do it separately from this PR since as it stands the parsing is unchanged at the moment.

At the point in the code where you're commenting, the URL is already parsed into a proper url.URL object.

fdawg4l: >  another one that uses the same regexes to go from ds://host/path format to url.URL because that function doesn't exist currently)
https://github.com/vmware/vic/blob/master/pkg/vsphere/datastore/datastore.go#L331

lib/portlayer/storage/vsphere/image.go
```diff
@@ -70,7 +71,16 @@ func NewImageStore(ctx context.Context, s *session.Session) (*ImageStore, error)
 	// Currently using the datastore associated with the session which is not
```
fdawg4l: Might as well update this comment since after this change it's no longer relevant.

fdawg4l:
LGTM

lib/portlayer/storage/vsphere/image.go
gigawhitlocks: good catch

gigawhitlocks:
Build's going to fail because I haven't updated tests yet but @emlin do the other changes look good?

emlin:
Everything else LGTM. 👍 
Test can be added in separate PR. 😄

gigawhitlocks:
Cool, I'll just fix the ones that are failing in Drone for the time being then

lib/install/management/store_files.go
gigawhitlocks: That one goes from `url.URL` to string "[datastore] /path" @fdawg4l

gigawhitlocks:
poke

gigawhitlocks:
1

gigawhitlocks:
build start please

gigawhitlocks:
poke

gigawhitlocks:
re: last commit message, this is working now, I do believe. Please take a look at the diff & keep an eye out for e.g. log messages that shouldn't make it into `master`. Thanks to everybody who helped me out on this issue.

cmd/vic-machine/create/create.go
```diff
@@ -116,8 +116,8 @@ func (c *Create) Flags() []cli.Flag {
 		cli.StringFlag{
 			Name:        "image-datastore, i",
```
hickeng: let's change this to `image-store` to be consistent with `volume-store`

cmd/vic-machine/create/create.go
```diff
@@ -308,8 +308,8 @@ func (c *Create) processParams() error {
 		return err
 	}
 
-	if c.ImageDatastoreName == "" {
-		return cli.NewExitError("--image-datastore Image datastore name must be specified", 1)
+	if c.ImageDatastorePath == "" {
+		return cli.NewExitError("--image-datastore Image datastore path must be specified", 1)
```
hickeng: Let's supply a message that avoids the need to refer to the documentaiton for format.
`--image-store must be specified, e..g datastore1 or datastore1/store`

lib/install/management/store_files.go
```diff
@@ -51,18 +51,17 @@ func (d *Dispatcher) DeleteStores(vchVM *vm.VirtualMachine, conf *config.Virtual
 	}
 
 	var errs []string
-	var emptyImages bool
 	var emptyVolumes bool
 	log.Infof("Removing images")
-	if emptyImages, err = d.deleteImages(ds, p); err != nil {
-		errs = append(errs, err.Error())
+	if err = d.deleteImages(conf); err != nil {
+		return err
```
hickeng: why has the behaviour changed here, i.e. why have we stopped appending the errors and instead bail?

lib/install/management/store_files.go
```diff
@@ -51,18 +51,17 @@ func (d *Dispatcher) DeleteStores(vchVM *vm.VirtualMachine, conf *config.Virtual
...
 	}
 	emptyVolumes, err = d.deleteDatastoreFiles(ds, path.Join(p, volumeRoot), d.force)
 
-	if emptyImages && emptyVolumes {
+	if emptyVolumes {
 		// if not empty, don't try to delete parent directory here
 		log.Debugf("Removing stores directory")
-		if err = d.deleteParent(ds, p); err != nil {
+		if _, err = d.deleteParent(ds, p); err != nil {
```
hickeng: I'm unable to figure out how this works - as best I can tell `p == "VIC/<uuid>"` but I cannot see anywhere that an additional prefix is added in front of "VIC" meaning we'll be attempting to delete `[datastore] VIC/<uuid>`
Is that path even used any more?

lib/install/validate/storage.go
```diff
@@ -32,14 +33,26 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
 	defer trace.End(trace.Begin(""))
 
 	// Image Store
-	imageDSpath, ds, err := v.DatastoreHelper(ctx, input.ImageDatastoreName, "", "--image-datastore")
+	imageDSpath, ds, err := v.DatastoreHelper(ctx, input.ImageDatastorePath, "", "--image-datastore")
+
+	// provide a default path if only a DS name is provided
+	if imageDSpath.Path == "" {
+		imageDSpath.Path = path.Join(input.DisplayName, "images")
```
hickeng: If this is the same datastore as the VCH appliance, this will be within the appliance VM folder - is that intentional?

lib/install/validate/storage.go
```diff
@@ -32,14 +33,26 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
...
 	v.NoteIssue(err)
 	if ds != nil {
 		v.SetDatastore(ds, imageDSpath)
 		conf.AddImageStore(imageDSpath)
 	}
 
 	// Volume Store(s)
+	conf.BootstrapImagePath = fmt.Sprintf("[%s] %s/%s",
+		imageDSpath.Host,
+		input.DisplayName,
```
hickeng: This is broken - it won't work if there's a directory conflict as ESX will create `[datastore] displayName_1` for example. The ISOs get uploaded to the correct location, but the values in the vmx are wrong.
This can be fixed in a different PR and a regression test added.

lib/install/management/store_files.go
gigawhitlocks: This is in error, we should not bail.

cmd/vic-machine/create/create.go
gigawhitlocks: sure

lib/install/validate/storage.go
gigawhitlocks: This is the same behavior as before, I believe. I did not intentionally change the behavior. There should be logic to make sure the directory doesn't exist in the routine that creates the directory (regardless of its location) but I will double-check

lib/install/validate/storage.go
gigawhitlocks: Did this issue exist before I moved this piece of code? @emlin do you know? This isn't actually new code, I just moved it to a different execution time.

gigawhitlocks:
OK I've got passing tests so I'm logging for the night but in the AM I'll take a closer look at @hickeng 's comments & give better responses/fix things and I'll be sure that the integration tests are also passing.

lib/install/management/store_files.go
emlin: should not bail here.

lib/install/validate/storage.go
emlin: No, there is no such issue before. Where is the code from?
I think you need to keep the original execution sequence, but only change the image store path there.

gigawhitlocks: OK so looking at this, I only moved it from `basics` to `storage` so that we have the image datastore path parsed before we do anything with the bootstrapimagepath.

If the directory already existed previously, I still don't see any kind of checks surrounding this code, so when this is created (not during validate time, note I'm just assigning these values to the VirtualContainerHostConfigSpec) if there are not checks I can add them -- in any case I don't see how moving these assignments from `basics` to `storage` would expose this issue, but I'm happy to resolve it either way.

lib/portlayer/storage/vsphere/image.go
```diff
@@ -63,16 +65,22 @@ type ImageStore struct {
...
 		return nil, err
 	}
 
-	// Currently using the datastore associated with the session which is not
-	// ideal.  This should be passed in via the config.  The datastore need not
-	// be the same datastore used for the rest of the system.
-	ds, err := datastore.NewHelper(ctx, s, s.Datastore, StorageParentDir)
+	datastores, err := s.Finder.DatastoreList(ctx, u.Host)
+	if err != nil {
+		// TODO: real error beyond whatever DatastoreList returns
```
gigawhitlocks: I like how no one spotted this 😛

lib/install/validate/storage.go
emlin: I found where it is. Thanks.
Then unfortunately, this problem exists before. The correct file path is set here:
https://github.com/vmware/vic/blob/master/lib/install/management/appliance.go#L414

Reference here https://github.com/vmware/vic/blob/master/lib/install/management/appliance.go#L362 for how to set correct file path, bootstrap.iso is located in the same place with appliance.iso.

lib/install/management/store_files.go
gigawhitlocks: At this point `p` should just be the root of the VCH directory, in your example `VIC` but if a user passes in a custom directory with `--image-store` it will be that value and it gets deleted if it's empty.

cmd/vic-machine/create/create.go
gigawhitlocks: roger

lib/install/management/store_files.go
```diff
@@ -72,19 +71,44 @@ func (d *Dispatcher) DeleteStores(vchVM *vm.VirtualMachine, conf *config.Virtual
...
+		}
+
+		if len(imageDSes) != 1 {
+			errs = append(errs, fmt.Sprintf("Invalid or ambiguous datastore name %s provided while attempting to remove image stores", imageDir.Host))
+			continue
+		}
 
-	_, err := d.deleteDatastoreFiles(ds, root, true)
-	return err
+		if emptyImages, err = d.deleteDatastoreFiles(imageDSes[0], imageDir.Path, true); err != nil {
```
fdawg4l: I don't think this will work consistently on VSAN.  VSAN has complicated and inconsistent behavior when deleting a directory full of child directories with `vmdk` files.  You'll need to delete the children first before deleting the files in most cases or this will return an error.  @emlin has a helper to do this in `vic-machine` somewhere.

emlin: Yeah, it will be invoked by this method if datastore is vSAN. 😄

lib/install/management/store_files.go
```diff
@@ -72,19 +71,44 @@ func (d *Dispatcher) DeleteStores(vchVM *vm.VirtualMachine, conf *config.Virtual
...
+
+	for _, imageDir := range conf.ImageStores {
+		imageDSes, err := d.session.Finder.DatastoreList(d.ctx, imageDir.Host)
+		if err != nil {
+			errs = append(errs, err.Error())
+			continue
+		}
+
+		if len(imageDSes) != 1 {
+			errs = append(errs, fmt.Sprintf("Invalid or ambiguous datastore name %s provided while attempting to remove image stores", imageDir.Host))
```
fdawg4l: I'm not a UX expert, but these kinds of error messages are infuriating because they tell me little about the actual condition that's throwing the error.  How about we let the user know that the datastore provided maps to >1 datastores found?

lib/portlayer/storage/vsphere/image.go
fdawg4l: What's to spot?  If error, return error.  Seems fine, no?

lib/install/validate/storage.go
hickeng: @gigawhitlocks don't address this here. I've reopened #1351 to address this.

doc/user_doc/vic_installation/vch_installer_options.md
```diff
@@ -52,10 +52,10 @@ You can designate the same datastore as the image store for multiple virtual con
 
 **NOTES**: 
 - vSphere Integrated Containers supports all alphanumeric characters, hyphens, and underscores in datastore paths and datastore names.
-- In the current builds the `image-datastore` option does not support datastore folders. If you specify a datastore folder in the `image-datastore` option, `vic-machine` does not return an error, but creates all of the necessary folders at the root level of the datastore.
+- In the current builds the `image-store` option does not support datastore folders. If you specify a datastore folder in the `image-store` option, `vic-machine` does not return an error, but creates all of the necessary folders at the root level of the datastore.
```
hickeng: @gigawhitlocks Thank you! I think you're the first person (including me) I've seen update the docs concurrently with the code change. That could be because I'm not reviewing every PR, but still :+1:

lib/install/management/store_files.go
gigawhitlocks: That's why I said "ambiguous", that's what that word means to me, but my developer tendencies are probably leaking so I'll word this a bit more explicitly

hickeng: I concur - it's worth the extra code to separate "unknown datastore" from "ambiguous datastore". @fdawg4l are you objecting to the use of ambiguous as a term?

doc/user_doc/vic_installation/vch_installer_options.md
gigawhitlocks: Oh! Yeah I see that, I'll change this so that it says we do support them now 😃

lib/install/management/store_files.go
fdawg4l: It's too ambiguous.

gigawhitlocks: OMG I laughed at that for real @fdawg4l

lib/install/management/store_files.go
```diff
@@ -72,19 +71,47 @@ func (d *Dispatcher) DeleteStores(vchVM *vm.VirtualMachine, conf *config.Virtual
...
+
+	for _, imageDir := range conf.ImageStores {
+		imageDSes, err := d.session.Finder.DatastoreList(d.ctx, imageDir.Host)
+		if err != nil {
+			errs = append(errs, err.Error())
+			continue
+		}
+
+		if len(imageDSes) != 1 {
+			errs = append(errs, fmt.Sprintf("Found %d datastores with provided datastore path %s. Provided datastore path must identify exactly one datastore.",
```
gigawhitlocks: omg lol @fdawg4l I just actually laughed IRL at that response.

 Back to the topic though, is this error message a bit better?

fdawg4l: Looks great.  Thanks!

lib/install/management/store_files.go
hickeng: Remove this entire function and replace it with just the call to `deleteImages`. This
a. is trying to delete directories that don't exist
b. shouldn't be deleting the volumes even if it was looking in the correct location.

doc/user_doc/vic_installation/vch_installer_examples.md
```diff
@@ -183,9 +183,9 @@ In addition to the mandatory options for deployment to a cluster, this example s
 --user Administrator@vsphere.local
 --password <i>vcenter_sso_password</i>
 --compute-resource <i>cluster_name</i>
---image-datastore <i>datastore_name</i>
+--image-store <i>datastore_name</i>
```
hickeng: let's have one of these examples with a path portion

lib/install/validate/validator_test.go
```diff
@@ -250,25 +250,77 @@ func testStorage(v *Validator, input *data.Data, conf *config.VirtualContainerHo
...
+			true,
+			"",
+			nil},
+
+		{"ds://",
+			map[string]string{"volume1": "",
+				"volume2": "ds://"},
+			true,
+			"",
+			nil},
```
hickeng: can we have a datastore with a space in the name - given I failed to anticipate that initially having a test to confirm we don't break it seems a good idea.

lib/spec/disk.go
```diff
@@ -80,7 +80,9 @@ func (s *VirtualMachineConfigSpec) AddVirtualDisk(device *types.VirtualDisk) *Vi
 			VirtualDeviceFileBackingInfo: types.VirtualDeviceFileBackingInfo{
 				// XXX This needs to come from a storage helper in the future
 				// and should not be computed here like this.
-				FileName: s.Datastore.Path(fmt.Sprintf("VIC/%s/images/%s/%[2]s.vmdk",
+
+				FileName: s.Datastore.Path(fmt.Sprintf("%s/VIC/%s/images/%s/%[3]s.vmdk",
```
hickeng: Just a note - this is the hack that will be moved into Storage.Join

tests/test-cases/Group6-VIC-Machine/6-4-Create-Basic.robot
```diff
@@ -53,7 +53,7 @@ Create VCH - full params
     Set Test VCH Name
 
     Log To Console  \nInstalling VCH to test server...
-    ${output}=  Run  bin/vic-machine-linux create --name=${vch-name} --target=%{TEST_URL} --user=%{TEST_USERNAME} --image-datastore=%{TEST_DATASTORE} --appliance-iso=bin/appliance.iso --bootstrap-iso=bin/bootstrap.iso --password=%{TEST_PASSWORD} --force=true --bridge-network=network --compute-resource=%{TEST_RESOURCE} --timeout %{TEST_TIMEOUT} --volume-store=%{TEST_DATASTORE}/test:default
+    ${output}=  Run  bin/vic-machine-linux create --name=${vch-name} --target=%{TEST_URL} --user=%{TEST_USERNAME} --image-store=%{TEST_DATASTORE} --appliance-iso=bin/appliance.iso --bootstrap-iso=bin/bootstrap.iso --password=%{TEST_PASSWORD} --force=true --bridge-network=network --compute-resource=%{TEST_RESOURCE} --timeout %{TEST_TIMEOUT} --volume-store=%{TEST_DATASTORE}/test:default
```
hickeng: Let's have another that tests both a datastore path as well as just datastore. Longer term it should also be testing with different datastores between images and volumes - we'll need to ask @mhagen-vmware for additional datastores on the hosts.

lib/spec/disk.go
gigawhitlocks: Yup :)

lib/install/validate/validator_test.go
gigawhitlocks: Absolutely. Surprised I didn't add one w/ a space already actually (there's a unit test that does that, so that might be why I overlooked it here)

lib/install/management/store_files.go
```diff
@@ -276,7 +258,7 @@ func (d *Dispatcher) deleteVolumeStoreIfForced(conf *config.VirtualContainerHost
 		for label, url := range conf.VolumeLocations {
 			volumeStores.WriteString(fmt.Sprintf("\t%s: %s\n", label, url.Path))
 		}
-		log.Warnf("Since --force was not specified, the following volume stores will not be removed. Use the vSphere UI to delete content you do not wish to keep.\n%q", volumeStores.String())
+		log.Warnf("Since --force was not specified, the following volume stores will not be removed. Use the vSphere UI to delete content you do not wish to keep.\n%s", volumeStores.String())
```
gigawhitlocks: Hey @hickeng -- I remember in a PR seeing you tell Emma to use `%q` at the end of this log string, so I wanted to bring your attention to this and make sure it's OK that I replaced it with `%s`. When it was `%q` it was printing out like this:
`WARN[2016-08-10T12:08:50-05:00] Since --force was not specified, the following volume stores will not be removed. Use the vSphere UI to delete content you do not wish to keep.
"\tdefault: [datastore1] ians_volume_store_fo_sho\n" `

Now it looks like this:
`WARN[2016-08-10T12:29:07-05:00] Since --force was not specified, the following volume stores will not be removed. Use the vSphere UI to delete content you do not wish to keep.
	default: [datastore1] ians_volume_store_fo_sho
`

gigawhitlocks:
Final commit & concerns regarding it: added tests re: https://github.com/vmware/vic/pull/1678#discussion_r74122812

Needed to add a path format check to DatastoreHelper in validate/datastore.go -- calls the one in vsphere/datastore that I wrote a while back, also modified that to not bail when there's no path specified (since just having a datastore w/ no path is allowed in many contexts and I hit that while testing) and DatastoreHelper now bails successfully if the user tries to use an emoji or other non-ascii character as their ds name.

@hickeng @fdawg4l please take a look at those changes specifically in the most recent commit.

tests/test-cases/Group6-VIC-Machine/6-4-Create-Basic.robot
gigawhitlocks: Done

doc/user_doc/vic_installation/vch_installer_examples.md
gigawhitlocks: Done

gigawhitlocks:
poke

lib/install/management/store_files.go
hickeng: Looks good - was not expecting the string to contain formatting elements at this point.

lib/install/management/store_files.go
```diff
@@ -40,51 +38,47 @@ const (
...
+		if err != nil {
+			errs = append(errs, err.Error())
+			continue
+		}
 
-	p, err := d.getVCHRootDir(vchVM) // p would be path but there's an imported package called path
-	if err != nil {
-		return err
-	}
+		if len(imageDSes) != 1 {
```
hickeng: I don't think there's anything the user can do to address this as we're pulling the input from an existing config.
Will -force let the user progress past this error?

lib/install/validate/validator_test.go
```diff
@@ -102,6 +104,11 @@ func TestMain(t *testing.T) {
 		if err != nil {
 			t.Errorf("Failed to new validator: %s", err)
 		}
+		ds, _ := validator.Session.Finder.Datastore(validator.Context, "LocalDS_1")
+		simulator.Map.Get(ds.Reference()).(mo.Entity).Entity().Name = "Local DS_0"
+
+		ds, _ = validator.Session.Finder.Datastore(validator.Context, "LocalDS_2")
+		simulator.Map.Get(ds.Reference()).(mo.Entity).Entity().Name = `😗`
```
hickeng: nice

lib/portlayer/storage/vsphere/image.go
```diff
@@ -60,16 +62,22 @@ type ImageStore struct {
...
-	// ideal.  This should be passed in via the config.  The datastore need not
-	// be the same datastore used for the rest of the system.
-	ds, err := datastore.NewHelper(ctx, s, s.Datastore, StorageParentDir)
+	datastores, err := s.Finder.DatastoreList(ctx, u.Host)
+	if err != nil {
+		return nil, errors.New(fmt.Sprintf("Host returned error when trying to locate provided datastore %s: %s", u.String(), err.Error()))
+	}
+
+	if len(datastores) != 1 {
+		return nil, errors.New(fmt.Sprintf("Found %d datastores with provided datastore path %s. Installation will not continue, as provided datastore path must identify exactly one datastore.", len(datastores), u.String()))
```
hickeng: Is this really called during installation?

gigawhitlocks: Yeah, this is where the image store is actually created

gigawhitlocks: OK this is fixed now

lib/install/management/store_files.go
gigawhitlocks: Because we `continue` and just append errs, it should continue and just print out the error so that the user can do manual cleanup if necessary.

lib/install/validate/validator_test.go
gigawhitlocks: (thanks @dougm)

hickeng:
nice job @gigawhitlocks - LGTM

gigawhitlocks:
YAYYYYY

PR #1705, opened 2016-07-27
Adds VolumeStore to the name portion of Volume ls output

### LINKED ISSUE #1679: add volume store name to the volume ls output.
currently the output only displays `driver` and `volume name` 

Since our implementation also relies on multiple volume-stores. We want the `volume ls` command to actually report the name of a volume along with the store it belongs too. 

**Expected behavior:**

[test-store]:VolumeName

**Actual behavior:**
VolumeName

matthewavery:
Fixes #1679 

Cormac brought up the fact that after making a volume there is no way to tell what volume store a specfic volume belonged to. He recommended providing a prefix to the name field in `volume ls` so the output of the said function was much more useful. 

sample output:
```
 % docker -H 10.17.109.84:2376 --tls volume ls                                                                   
DRIVER              VOLUME NAME
vsphere             [default]:newTestVol
vsphere             [test-store]:testVol
```

### DIFF
--- a/lib/apiservers/engine/backends/volume.go
+++ b/lib/apiservers/engine/backends/volume.go
@@ -81,6 +81,7 @@ func (v *Volume) Volumes(filter string) ([]*types.Volume, []string, error) {
 			return nil, nil, fmt.Errorf("error unmarshalling docker metadata: %s", err)
 		}
 		volume := NewVolumeModel(vol, volumeMetadata.Labels)
+		volume.Name = fmt.Sprintf("[%s]:%s", vol.Store, vol.Name)
 		volumes = append(volumes, volume)
 	}
 	return volumes, nil, nil


jzt:
lgtm

caglar10ur:
how is this problem handled in docker world with different volume drivers?

caglar10ur:
https://docs.docker.com/engine/tutorials/dockervolumes/#/locating-a-volume so I think we should rely on inspect and not change volume names like this. I personally don't like name a thing pikachu and found out that it's no longer called X but [Pokemon]:pikachu

matthewavery:
@caglar10ur volume ls already displays the driver as a separate column. As for other drivers, I am not sure. We can rely on `inspect` for this, I put this up since it was a comment that Cormac had that he though would be beneficial.

caglar10ur:
I understand that's a useful thing but I don't think our solution should involve changing user entered data. In the above scenario, how am I going to use newTestVol? Are we going to refer it  as [default]:newTestVol or just newTestVol? If latter than this change just broke all the scripts on the wild since they need to split the volume name from : and use the second part. If we are going to use the first one then I doubt : will work as expected as it means a remote host, also we need to escape from [ and ] while writing shell commands.

So overall - as you can tell - I really don't like this change :)

matthewavery:
@caglar10ur fair enough good point :) I put this PR up for conversation, no worries. It was a quick change. I will look at the `docker inspect` for this information instead. I will probably look at putting the volume store name in the `source` field. pending further conversation with faiyaz and george as well. The unfortunate part is that `volume ls` does not give us a great ability to add columns to the output...

matthewavery:
Closing this issue as it has been deemed not needed/desired for design reasons.

PR #1707, opened 2016-07-27
Return nil when no volume metadata files are found

fdawg4l:
TestVolumeCreateListAndRestart was failing on this.  Also added metadata
to the test to check metadata is handled properly.  That is making sure
it gets read on startup and returned appropriately.


dougm:
lgtm

caglar10ur:
LGTM :sheep: :it:

fdawg4l:
.

PR #1708, opened 2016-07-27
Show correct output while pulling fake image (#757)

### LINKED ISSUE #757: Incorrect output when attempting to pull a non-existant image
Docker:

```
vagrant@devbox ~/g/s/g/v/vic ❯❯❯ docker pull ibm/os2warp
Using default tag: latest
Pulling repository docker.io/ibm/os2warp
Error: image ibm/os2warp not found
```

VIC:

```
vagrant@devbox /U/z/g/s/g/v/vic ❯❯❯ docker $DOCKER_OPTS pull ibm/os2warp
Using default tag: latest
Error response from daemon: exit status 1
```

anchal-agrawal:
* Fixes output while pulling a non-existent image

Vanilla docker:
```
$ sudo docker pull foo/bar
Using default tag: latest
Pulling repository docker.io/foo/bar
Error: image foo/bar not found
```

VIC:
```
$ docker -H <host> --tls pull foo/bar
Using default tag: latest
latest: Pulling from foo/bar
Error: image foo/bar not found
```
* Enables related integration test for `docker pull`

Fixes #757

### DIFF
--- a/cmd/imagec/imagec.go
+++ b/cmd/imagec/imagec.go
@@ -694,16 +694,20 @@ func main() {
 		options.token = token
 	}
 
-	// Get the manifest
-	manifest, err := FetchImageManifest(options)
-	if err != nil {
-		log.Fatalf("Error while pulling image manifest: %s", err)
-	}
-
 	// HACK: Required to learn the name of the vmdk from given reference
 	// Used by docker personality until metadata support lands
 	if !options.resolv {
 		progress.Message(po, options.tag, "Pulling from "+options.image)
+	}
+
+	// Get the manifest
+	manifest, err := FetchImageManifest(options)
+	if err != nil {
+		if strings.Contains(err.Error(), "image not found") {
+			log.Fatalf("Error: image %s not found", options.image)
+		} else {
+			log.Fatalf("Error while pulling image manifest: %s", err)
+		}
 	}
 
 	// Create the ImageWithMeta slice to hold Image structs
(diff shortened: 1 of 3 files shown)


cmd/imagec/imagec.go
```diff
@@ -694,18 +694,22 @@ func main() {
...
 	// HACK: Required to learn the name of the vmdk from given reference
 	// Used by docker personality until metadata support lands
 	if !options.resolv {
 		progress.Message(po, options.tag, "Pulling from "+options.image)
 	}
 
+	// Get the manifest
+	manifest, err := FetchImageManifest(options)
+	if err != nil {
+		if strings.Contains(err.Error(), "image not found") {
```
caglar10ur: this is ugly :) Why don't you just return the desired error message from the above function in the first place?

anchal-agrawal: Thanks for your input! I figured I didn't have the image name in `fetcher.go`, that's why I propagate `"image not found"` and add the name in `imagec.go`.

I believe the image name would be in the `url` struct there, so let me see if I can parse it out.

caglar10ur: options.image should be the image name in FetchImageManifest since we are passing the whole struct to it

caglar10ur: Hmm I see what you are saying now, it's fetch function that returning the error not FetchImageManifest

caglar10ur: One other possibility is starting to return typed errors and use assertion instead of parsing the error message itself but that could be more involved

I'll create an issue for that.

LGTM

caglar10ur: https://github.com/vmware/vic/issues/1710 is created to track that work

matthewavery: lgtm :sheep: :it:

matthewavery:
lgtm :sheep:  :it:

caglar10ur:
LGTM :sheep: :it: (

anchal-agrawal:
Triggered another build, it had failed while downloading test logs.

PR #1726, opened 2016-07-28
Unify metadata functions for images and volumes

fdawg4l:
This is just code cleanup.  We need to store metadata for volumes as
well as images.  Might as well use the same routines for both.


caglar10ur:
LGTM :sheep: :it:

emlin:
lgtm

fdawg4l:
.

PR #1740, opened 2016-07-29
Add container stop signal support

### LINKED ISSUE #1268: Dockerfile STOPSIGNAL support
Use the dockerfile STOPSIGNAL value as the signal to deliver to the container.
If the container doesn't stop after `docker personality defined timeout` (I believe that was 15s) then send SIGTERM.

dougm:
The stop signal can be specified in the Dockerfile or docker run cli.

The toolbox now has support for the VirtualMachine.ShutdownGuest API
and 'docker stop' behaves the same as guest shutdown from the vSphere
side.

Fixes #1268

### DIFF
--- a/cmd/tether/msgs/messages.go
+++ b/cmd/tether/msgs/messages.go
@@ -14,7 +14,13 @@
 
 package msgs
 
-import "golang.org/x/crypto/ssh"
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	"golang.org/x/crypto/ssh"
+)
 
 // All of the messages passed over the ssh channel/global mux are (or will be)
 // defined here.
@@ -91,6 +97,28 @@ func (s *SignalMsg) Unmarshal(payload []byte) error {
 
 func (s *SignalMsg) Signum() int {
 	return Signals[s.Signal]
+}
+
+func (s *SignalMsg) FromString(name string) error {
+	num, err := strconv.Atoi(name)
+	if err == nil {
+		for sig, val := range Signals {
+			if num == val {
+				s.Signal = sig
+				return nil
+			}
+		}
+	}
+
+	name = strings.TrimPrefix(strings.ToUpper(name), "SIG")
+
+	s.Signal = ssh.Signal(name)
+	_, ok := Signals[s.Signal]
+	if !ok {
+		return fmt.Errorf("unsupported signal name: %q", name)
+	}
+
+	return nil
 }
 
 // ContainersMsg
--- a/cmd/tether/msgs/messages_test.go
+++ b/cmd/tether/msgs/messages_test.go
@@ -42,6 +42,20 @@ func TestSignal(t *testing.T) {
 	out.Unmarshal(tmp)
 
 	assert.Equal(t, s, out)
+
+	for _, name := range []string{"SIGQUIT", "QUIT", "quit", "3"} {
+		err := out.FromString(name)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	for _, name := range []string{"SIGNOPE", "nope", "0", "-1"} {
+		err := out.FromString(name)
+		if err == nil {
+			t.Errorf("expected error parsing %q", name)
+		}
+	}
 }
 
 func TestContainers(t *testing.T) {
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -107,9 +107,10 @@ func (handler *ContainersHandlersImpl) CreateHandler(params containers.CreatePar
 				},
 			},
 		},
-		Key:      pem.EncodeToMemory(&privateKeyBlock),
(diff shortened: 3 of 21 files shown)


cmd/tether/msgs/messages.go
```diff
@@ -93,6 +99,28 @@ func (s *SignalMsg) Signum() int {
...
 
+func (s *SignalMsg) FromString(name string) error {
+	num, err := strconv.Atoi(name)
+	if err == nil {
+		for sig, val := range Signals {
+			if num == val {
+				s.Signal = sig
+				return nil
+			}
+		}
```
hickeng: What do we do if num is not in the list?

cmd/tether/msgs/messages_test.go
```diff
@@ -42,6 +42,20 @@ func TestSignal(t *testing.T) {
...
+			t.Fatal(err)
+		}
+	}
+
+	for _, name := range []string{"SIGNOPE", "nope", "0", "-1"} {
+		err := out.FromString(name)
+		if err == nil {
+			t.Errorf("expected error parsing %q", name)
+		}
+	}
```
hickeng: Add a test for numeric signal

cmd/tether/msgs/messages.go
dougm: Will fallthrough to the error below: `return fmt.Errorf("unsupported signal name: %q", name)`

cmd/tether/msgs/messages_test.go
dougm: There are numeric tests in there "3" for the success case and "0", "-1" for the error case.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -105,9 +105,10 @@ func (handler *ContainersHandlersImpl) CreateHandler(params containers.CreatePar
 				},
 			},
 		},
-		Key:      pem.EncodeToMemory(&privateKeyBlock),
-		LayerID:  *params.CreateConfig.Image,
-		RepoName: *params.CreateConfig.RepoName,
+		Key:        pem.EncodeToMemory(&privateKeyBlock),
+		LayerID:    *params.CreateConfig.Image,
+		RepoName:   *params.CreateConfig.RepoName,
+		StopSignal: *params.CreateConfig.StopSignal,
```
hickeng: I think we should associate this per Session rather than per Executor. If we can have multiple Sessions per Executor then we should allow those sessions to have distinct stop signals.

lib/tether/toolbox.go
```diff
@@ -56,10 +61,63 @@ func (t *Toolbox) Stop() error {
...
+
+	if err := session.Cmd.Process.Signal(num); err != nil {
+		return fmt.Errorf("failed to signal %s: %s", session.ID, err)
+	}
+
 	return nil
 }
 
+func (t *Toolbox) halt() error {
+	session := t.config.Sessions[t.config.ID]
```
fdawg4l: Is it possible this map is empty?

fdawg4l:
LGTM

lib/tether/tether.go
```diff
@@ -349,6 +352,8 @@ func (t *tether) launch(session *SessionConfig) error {
 	session.Errwriter = logwriter
 	session.Reader = dio.MultiReader()
 
+	session.exit = make(chan struct{})
```
hickeng: why?

lib/tether/tether.go
```diff
@@ -290,6 +290,9 @@ func (t *tether) Register(name string, extension Extension) {
 func (t *tether) handleSessionExit(session *SessionConfig) {
 	defer trace.End(trace.Begin("handling exit of session " + session.ID))
 
+	// notify guest shutdown waiter this process has gone away
+	close(session.exit)
```
hickeng: can we move this after the IO is shut down - if were to hang anything else off the channel, whether all output has been written could be important.

lib/tether/toolbox.go
dougm: AFAICT it would only be empty if someone manually removed it from the .vmx, but can add a check anyways.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
dougm: ok.  The StopSignal is currently only applied session where SessionConfig.ID == ExecutorConfig.ID

lib/tether/tether.go
dougm: sure, will do.

dougm:
I'll address the review comments after #1727 is merged, there will be merge conflicts that are easier for me to resolve in this PR than vice versa.

lib/tether/toolbox.go
fdawg4l: Nah, it's cool.  I just didnt understand the behavior and wanted clarification.  Sounds good as is.  Thanks!

caglar10ur:
LGTM :sheep: :it:

dougm:
👊

PR #1742, opened 2016-07-29
Fix nil pointer dereferences in vicadm_test.go

kjplatz:
The associated panic() occurs periodically when running "go test":
2016/07/28 17:08:32 GET https://localhost:33160/logs/vicadmin.log:
time="2016-07-28T17:08:32Z" level=debug msg="[ END ] [github.com/vmware/vic/cmd/vicadmin.(*server).tailFiles] [28.04363ms] " 
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x20 pc=0xb0685a]

goroutine 69 [running]:
panic(0x1026cc0, 0xc820012110)
    /usr/local/go/src/runtime/panic.go:481 +0x3e6
bufio.(*Writer).flush(0xc8201cc200, 0x0, 0x0)
    /usr/local/go/src/bufio/bufio.go:562 +0xda
bufio.(*Writer).Flush(0xc8201cc200, 0x0, 0x0)
    /usr/local/go/src/bufio/bufio.go:551 +0x2d
net/http.(*response).Flush(0xc8201d1e10)
    /usr/local/go/src/net/http/server.go:1301 +0x45
github.com/vmware/vic/cmd/vicadmin.(*flushWriter).Write(0xc8203cfa00, 0xc820396000, 0x183, 0x8000, 0x183, 0x0, 0x0)
    github.com/vmware/vic/cmd/vicadmin/_test/_obj_test/vicadm.go:458 +0xcd
io.copyBuffer(0x7fa4f9b81bc0, 0xc8203cfa00, 0x7fa4f9b79590, 0xc820026028, 0xc820396000, 0x8000, 0x8000, 0x2023, 0x0, 0x0)
    /usr/local/go/src/io/io.go:382 +0x2c9
io.Copy(0x7fa4f9b81bc0, 0xc8203cfa00, 0x7fa4f9b79590, 0xc820026028, 0x0, 0x0, 0x0)
    /usr/local/go/src/io/io.go:350 +0x64
os/exec.(*Cmd).writerDescriptor.func1(0x0, 0x0)
    /usr/local/go/src/os/exec/exec.go:236 +0x8b
os/exec.(*Cmd).Start.func1(0xc8201dcc80, 0xc8203cfb80)
    /usr/local/go/src/os/exec/exec.go:344 +0x1d
created by os/exec.(*Cmd).Start
    /usr/local/go/src/os/exec/exec.go:345 +0x967
exit status 2
FAIL    github.com/vmware/vic/cmd/vicadmin    0.248s
Makefile:217: recipe for target 'test' failed
make: *** [test] Error 1
[info] build failed (exit code 2)

It turns out that there is currently a race condition in tailFiles, where the following sequence of events happens:
    The http connection is closed and the runtime begins tearing down the connection
    The http CloseNotifier sends a signal on a channel
    The main thread receives from the channel and exits
    The deferred cmd.Kill() kills the "tail -f" command

The race condition occurs when the "tail -f" command writes (and flushes) to the flushWriter after the runtime has torn down the endpoint, resulting in a nil pointer dereference.  This PR moves the "tail -f" functionality into a "tailFile" function which runs as a goroutine and explicitly accepts that there may be data races, and does an explicit recover() from any panic() which may arise from this.
    
Fixes nil pointer dereferences in vicadm_test.go.

### DIFF
--- a/cmd/vicadmin/logs.go
+++ b/cmd/vicadmin/logs.go
@@ -36,6 +36,7 @@ import (
 
 const (
 	uint32max = (1 << 32) - 1
+	tailLines = 8
 )
 
 type dlogReader struct {
@@ -301,7 +302,7 @@ func zipEntries(readers map[string]entryReader, out *zip.Writer) error {
 }
 
 func tailFile(wr io.Writer, file string, done *chan bool) error {
-	nlines := 8
+	nlines := tailLines
 	f, err := os.Open(file)
 	if err != nil {
 		return err
--- a/cmd/vicadmin/logs.go
+++ b/cmd/vicadmin/logs.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"os"
 	"path"
 	"sync"
 	"time"
@@ -295,6 +296,60 @@ func zipEntries(readers map[string]entryReader, out *zip.Writer) error {
 			continue
 		}
 		log.Infof("Wrote %d bytes to %s", sz, header.Name)
+	}
+	return nil
+}
+
+func tailFile(wr io.Writer, file string, done *chan bool) error {
+	nlines := 8
+	f, err := os.Open(file)
+	if err != nil {
+		return err
+	}
+
+	pos, err := f.Seek(0, 2)
+	if err != nil {
+		return err
+	}
+	b := make([]byte, 1)
+	for pos > 0 && nlines >= 0 {
+		pos--
+		_, err := f.ReadAt(b, pos)
+		if err != nil && err != io.EOF {
+			return err
+		}
+		if b[0] == '\n' {
+			nlines--
+		}
+	}
+
+	pos++
+	if _, err = f.Seek(pos, 0); err != nil {
+		return err
+	}
+
+	b = make([]byte, 1024)
+	// We KNOW there's a data race here.
+	// But it doesn't break anything, so we just trap it.
+	defer func() {
+		_ = recover()
+	}()
+	var n int
+	for true {
+		for err != io.EOF {
+			n, err = f.Read(b)
+			if err != nil && err != io.EOF {
+				return err
+			}
+			fmt.Fprint(wr, string(b[:n]))
+		}
+		select {
+		case _ = <-*done:
+			return nil
+		default:
+		}
+		time.Sleep(50 * time.Millisecond)
+		err = nil
 	}
 	return nil
 }
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -23,7 +23,7 @@ import (
 	"html/template"
 	"net"
 	"net/http"
-	"os/exec"
+	// "os/exec"
 	"path/filepath"
 
 	"golang.org/x/net/context"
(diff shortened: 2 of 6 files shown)


cmd/vicadmin/logs.go
```diff
@@ -298,3 +299,57 @@ func zipEntries(readers map[string]entryReader, out *zip.Writer) error {
 	}
 	return nil
 }
+
+func tailFile(wr io.Writer, file string, done *chan bool) error {
+	nlines := 8
```
matthewavery: maybe make this magic number a constant up at the top?

cmd/vicadmin/server.go
```diff
@@ -23,7 +23,7 @@ import (
 	"html/template"
 	"net"
 	"net/http"
-	"os/exec"
+	// "os/exec"
```
jzt: Remove commented code.

cmd/vicadmin/server.go
```diff
@@ -248,21 +248,25 @@ func (s *server) tailFiles(res http.ResponseWriter, req *http.Request, names []s
 		w: res,
 	}
 
-	// TODO: tail in Go, rather than shelling out
+	/* // TODO: tail in Go, rather than shelling out
 	cmd := exec.Command("tail", append([]string{"-F"}, names...)...)
 	cmd.Stdout = fw
 	cmd.Stderr = fw
+	*/
```
jzt: Ditto

cmd/vicadmin/server.go
```diff
@@ -248,21 +248,25 @@ func (s *server) tailFiles(res http.ResponseWriter, req *http.Request, names []s
...
 
-	if err := cmd.Start(); err != nil {
+	done := make(chan bool)
+	go tailFile(fw, names[0], &done)
+
+	/* if err := cmd.Start(); err != nil {
```
jzt: Tritto

jzt:
LGTM

cmd/vicadmin/server.go
```diff
@@ -248,21 +248,25 @@ func (s *server) tailFiles(res http.ResponseWriter, req *http.Request, names []s
 		w: res,
 	}
 
-	// TODO: tail in Go, rather than shelling out
+	/* // TODO: tail in Go, rather than shelling out
```
matthewavery: is this /* needed?

matthewavery:
LGTM

cmd/vicadmin/logs.go
```diff
@@ -298,3 +300,57 @@ func zipEntries(readers map[string]entryReader, out *zip.Writer) error {
 	}
 	return nil
 }
+
+func tailFile(wr io.Writer, file string, done *chan bool) error {
+	nlines := tailLines
+	f, err := os.Open(file)
```
hickeng: If the log file hasn't been created when this is first run then it will fail. This could easily be the case, particularly for imagec.log (in it's current form).

cmd/vicadmin/logs.go
```diff
@@ -298,3 +300,57 @@ func zipEntries(readers map[string]entryReader, out *zip.Writer) error {
...
+	for true {
+		for err != io.EOF {
+			n, err = f.Read(b)
+			if err != nil && err != io.EOF {
+				return err
+			}
+			fmt.Fprint(wr, string(b[:n]))
+		}
+		select {
+		case _ = <-*done:
```
hickeng: This is explicitly what the context package is for - that allows for explicit Cancellation, timeouts etc.

cmd/vicadmin/logs.go
```diff
@@ -298,3 +300,57 @@ func zipEntries(readers map[string]entryReader, out *zip.Writer) error {
...
+	}
+
+	b = make([]byte, 1024)
+	// We KNOW there's a data race here.
+	// But it doesn't break anything, so we just trap it.
+	defer func() {
+		_ = recover()
+	}()
+	var n int
+	for true {
```
hickeng: no need for `true` here - go idom is:
`for {`

hickeng:
Consider using a package such as https://github.com/hpcloud/tail which handles all the various edge conditions such as deletion, truncation, et al. It also uses inotify which is a significantly less impactful method of checking for new content than loop/sleep.

kjplatz:
.

kjplatz:
.

PR #1748, opened 2016-07-29
Adding encoding/decoding for slice of net.IP's

hmahmood:


hmahmood:
@caglar10ur coud you please review?

pkg/vsphere/extraconfig/encode.go
```diff
@@ -135,6 +136,15 @@ func encodeStruct(sink DataSink, src reflect.Value, prefix string, depth recursi
 	}
 }
 
+func isEncodableSliceElemType(t reflect.Type) bool {
+	switch t {
+	case reflect.TypeOf((net.IP)(nil)):
```
caglar10ur: I think I prefer using reflect.Zero here instead of casting nil to net.IP

caglar10ur:
a small comment otherwise LGTM

pkg/vsphere/extraconfig/encode.go
hmahmood: Not sure how to do that since [reflect.Zero](https://golang.org/pkg/reflect/#Zero) takes in a Type, and I am trying to get a Type in the first place.

dougm:
lgtm

hmahmood:
♻️

PR #1750, opened 2016-07-29
Fix test logic

mhagen-vmware:
Alleviate a race condition in the ps test, when the sleep and the other creates are not enough to let the first container finish it's output.


kjplatz:
:+1: :sheep:

matthewavery:
LGTM

andrewtchin:
lgtm

mhagen-vmware:
.

Issue #1710, opened 2016-07-28
Structured errors


caglar10ur:
https://github.com/vmware/vic/pull/1708/files#diff-20cac0652b997c8f1ee1d0fb89688480R706 is an example what happens when we don't define errors as types. 

We should implement a set of types like net package (https://golang.org/pkg/net/) and start using type assertions

#3636 should be defining these errors as an integral part of the interface work for the portlayer.

caglar10ur:
@mdubya66 not sure why is this assigned to me and estimated as 2. This was just a placeholder issue to for the overall changes needed in our components. I think this is way bigger than 2 and requires multiple man efforts so I'm removing the milestone and putting it into icebox

PR #1753, opened 2016-07-29
Show correct output during rm on a running container

### LINKED ISSUE #1311: docker rm to a running container returns incorrect output
Actual:
Error response from daemon: server error from portlayer

Expected:
Error response from daemon: Conflict, You cannot remove a running container. Stop the container before attempting removal or use -f

anchal-agrawal:
* Fixes output during a `docker rm` on a running container

Vanilla docker:
```
$ sudo docker create busybox /bin/top
<container-id>
$ sudo docker start <container-id>
<container-id>
$ sudo docker rm <container-id>
Error response from daemon: You cannot remove a running container <container-id>. Stop the container before attempting removal or use -f
```

VIC:
```
$ docker -H <host> --tls create busybox /bin/top
<container-id>
$ docker -H <host> --tls start <container-id>
<container-id>
$ docker -H <host> --tls rm <container-id>
Error response from daemon: You cannot remove a running container. Stop the container before attempting removal or use -f
```

* Enables the related integration test in the `docker rm` suite

Fixes #1311

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -409,20 +409,19 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 	//call the remove directly on the name. No need for using a handle.
 	_, err := client.Containers.ContainerRemove(containers.NewContainerRemoveParams().WithID(name))
 	if err != nil {
-		if _, ok := err.(*containers.ContainerRemoveNotFound); ok {
+		switch err := err.(type) {
+		case *containers.ContainerRemoveNotFound:
 			return derr.NewRequestNotFoundError(fmt.Errorf("No such container: %s", name))
+		case *containers.ContainerRemoveDefault:
+			return derr.NewErrorWithStatusCode(fmt.Errorf("server error from portlayer : %s", err.Payload.Message), http.StatusInternalServerError)
+		case *containers.ContainerRemoveConflict:
+			return derr.NewErrorWithStatusCode(fmt.Errorf(err.Payload.Message), http.StatusConflict)
+		default:
+			return derr.NewErrorWithStatusCode(fmt.Errorf("server error from portlayer : %s", err), http.StatusInternalServerError)
 		}
-		if errModel, ok := err.(*containers.ContainerRemoveDefault); ok {
-			return derr.NewErrorWithStatusCode(fmt.Errorf("server error from portlayer : %s", errModel.Payload.Message), http.StatusInternalServerError)
-		}
-		if _, ok := err.(*containers.ContainerRemoveConflict); ok {
-			return derr.NewErrorWithStatusCode(fmt.Errorf("You cannot remove a running container. Stop the container before attempting removal or use -f"), http.StatusConflict)
-		}
-		return derr.NewErrorWithStatusCode(fmt.Errorf("server error from portlayer : %s", err), http.StatusInternalServerError)
 	}
 	// delete container from the cache
 	cache.ContainerCache().DeleteContainer(name)
-
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


mhagen-vmware:
lgtm

anchal-agrawal:
@corrieb @caglar10ur @matthewavery I'd really appreciate your input on this. I added an HTTP 409 in swagger to mirror [docker's remote API for `rm`](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.23/#/remove-a-container).

@caglar10ur, I do a string comparison on the error [here](https://github.com/vmware/vic/pull/1753/files#diff-b26a5f740e99039b0c7fd95ab80ce082R227), which is related to #1710.

lib/apiservers/engine/backends/container.go
```diff
@@ -515,6 +515,9 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 		if errModel, ok := err.(*containers.ContainerRemoveDefault); ok {
 			return derr.NewErrorWithStatusCode(fmt.Errorf("server error from portlayer : %s", errModel.Payload.Message), http.StatusInternalServerError)
 		}
+		if _, ok := err.(*containers.ContainerRemoveConflict); ok {
+			return derr.NewErrorWithStatusCode(fmt.Errorf("You cannot remove a running container. Stop the container before attempting removal or use -f"), http.StatusConflict)
+		}
 		return derr.NewErrorWithStatusCode(fmt.Errorf("server error from portlayer : %s", err), http.StatusInternalServerError)
```
matthewavery: this error handling section should be moved to a `type switch` instead of a series of `if` statements see lib/apiservers/engine/backends/volume.go for an example.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -223,6 +224,9 @@ func (handler *ContainersHandlersImpl) RemoveContainerHandler(params containers.
 
 	err := h.Container.Remove(context.Background(), handler.handlerCtx.Session)
 	if err != nil {
+		if strings.Contains(err.Error(), "The attempted operation cannot be performed in the current state (Powered on)") {
```
dougm: That message is localized, so it is best to check the error type instead.  In this case you could use the following helper:

```go
func isInvalidPowerStateError(err error) bool {
	if terr, ok := err.(task.Error); ok {
		_, invalid := terr.Fault().(*types.InvalidPowerState)
		return invalid
	}
	return false
}
```

hmahmood: You should instead return a typed error instead. See https://blog.golang.org/error-handling-and-go

lib/portlayer/exec/container.go
```diff
@@ -258,6 +266,16 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
...
 	dsPath := fmt.Sprintf("[%s] %s", url.Host, url.Path)
 
+	state, err := c.vm.PowerState(ctx)
+	if err != nil {
+		return fmt.Errorf("Failed to get vm power status %q: %s", c.vm.Reference(), err)
+	}
+
+	// don't remove the containerVM if it is powered on
+	if state == types.VirtualMachinePowerStatePoweredOn {
+		return RemovePowerError{fmt.Errorf("You cannot remove a running container. Stop the container before attempting removal or use -f")}
```
hmahmood: This error string is docker specific and should be in the docker personality. The port layer does not know what `-f` is. You should view the port layer as a library that can stand on its own, and so the errors should reflect that. Over here just "container is running," or something similar would suffice.

hmahmood:
lgtm

anchal-agrawal:
♻️

Issue #1724, opened 2016-07-28
Docker start time is inconsistent and avg 2x equivalent Bonneville (no vmfork)


corrieb:
I decided to run some tests today to see where we stand with respect to like-for-like performance with Bonneville. By like-for-like, I mean same kind of datastore, vmfork disabled etc.

On a single ESX host with SSD datastore, here are the numbers from one run:

pull busybox   VIC 4.4  Bonn 5.3
create   VIC 0.54  Bonn 0.77
start   VIC 13.2  Bonn 4.2
delete   VIC 0.45  Bonn 0.25

I'll raise a separate issue for the difference in delete performance.

Additionally, start time with VIC seems very inconsistent. 4 runs on VIC give:
16s, 9s, 16s, 7s
With Bonneville:
4s, 3.7s, 3.8s, 3.8s

corrieb:
I did a few manual boots and it seems like boot time is slow (7-8s), but consistent. I suspect there's two issues here. Boot time slowness (likely to be addressed as part of the work to optimize the container ISO) which can be addressed for GA. Then the additional inconsistencies being introduced in our control plane. This latter issue is a high priority to fix IMHO

corrieb:
Here's some debug from an example. I start the same container 3 times in succession. It takes 16s, then 12s, then 8s. This is the output from the port-layer:
```
time="2016-07-29T00:49:18Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/portlayer/exec.(*Container).start] Container.start" 
time="2016-07-29T00:49:18Z" level=info msg="attach connector: Received incoming connection" 
time="2016-07-29T00:49:23Z" level=debug msg="HandshakeClient: did not receive synack: 0x061a != 0x0001" 
time="2016-07-29T00:49:23Z" level=debug msg="discarding following 14 bytes from input channel\n" 
time="2016-07-29T00:49:23Z" level=debug msg="[3 4 5 6 7 8 9 10 11 12 13 14 15 16]\n" 
time="2016-07-29T00:49:24Z" level=debug msg="caught EOF" 
time="2016-07-29T00:49:34Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/portlayer/exec.(*Container).start] [16.056826898s] Container.start" 
time="2016-07-29T00:49:34Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/portlayer/exec.(*Container).Commit] [16.294369189s] Committing handle" 
[...]

PR #1756, opened 2016-07-29
Remove transitive govmomi inclusion from guest.UUID

hickeng:
DHCP id.go pulled in guest.UUID which inflated the size of the tether to 44MB from 12MB

Towards: #1724

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,11 +1,11 @@
-LICENSE
+ï»¿LICENSE
 
-vSphere Integrated Containers version 0.4.0
+vSphere Integrated Containers version 0.5.0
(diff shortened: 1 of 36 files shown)


dougm:
lgtm

hickeng:
@caglar10ur can I get a review? This one actually managed to pass integration ;)

hmahmood:
If its going to be just one function, maybe we can name it something other than guest?

fdawg4l:
lgtm

caglar10ur:
Lgtm

hickeng:
poke

Issue #366, opened 2016-03-31
Docker API: Docker Container Logs


sflxn:
Add docker log operation to our engine-api server and integrate that with the interaction portlayer.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#get-container-logs
https://docs.docker.com/engine/reference/commandline/logs/

Please note, the engine-api's rest server already packaging these json struct into golang structs. The url above is just for reference on what the operation should do.

dougm:
@jzt is going to take on the API/PortLayer piece, I'm taking the vSphere datastore file streaming piece (#1682)

sflxn:
Separated the last few docker operations into personality and portlayer work to follow the pattern we've been using on this project.

sflxn:
This will use a similar mechanism on the portlayer rest side as attach.  The same goes for the personality server side.  My estimates for this includes both the personality server and portlayer rest server.

In addition to container logs, docker logs also requires getting the stdout and stderr.  We should be able to attach to the tether and get those as well as get the container logs from the datastore log streaming in #1682.

sflxn:
@hickeng @dougm and I just had a good conversation about docker logs. Docker logs can do 2 things that we cannot currently do at this point. The first is to tag all log lines with a timestamp. The second is to be able to grab all log since a timestamp. The importance of the latter may not be apparent immediately. Docker logs interleaves containers logs, stdout, stderr into the output. The tether returns the stdout and stderr streams at the point that we ask for data. To fully support the docker logs behavior, we will need to keep stdout and stderr around as we do the container logs, and we need to do so with timestamp.

It was decided at this meeting that we would go with what we have right now. Timestamps and historical stdout & stderr requires significant plumbing and will need to come later.

@stuclem What this means is that the docker logs options --timestamps (also -t) and --since options will not work initially.

sflxn:
I noticed something very peculiar with docker logs.  I was trying to figure out how they delimit their log lines.  They are delimiting their logs with either /n or /r/n.

If you run

```
docker run -it busybox /bin/sh -c 'a=1; while [ $a -gt 0 ]; do echo "line $a"; a=`expr $a + 1`; sleep 1; done;'
```

their json log file looks normal.  However, if instead of echo, I use printf (no newline), their json log files never gets updated nor can I get the logs of the container until I quit the running session.  At that point, it just dumps out everything from stdout as one log line with the timestamp of that time.

```
[...]

PR #1757, opened 2016-07-29
Initial docker log portlayer spec and rest handler

sflxn:
Contributes to #366.

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -53,6 +53,7 @@ func (handler *ContainersHandlersImpl) Configure(api *operations.PortLayerAPI, h
 	api.ContainersGetContainerInfoHandler = containers.GetContainerInfoHandlerFunc(handler.GetContainerInfoHandler)
 	api.ContainersGetContainerListHandler = containers.GetContainerListHandlerFunc(handler.GetContainerListHandler)
 	api.ContainersContainerSignalHandler = containers.ContainerSignalHandlerFunc(handler.ContainerSignalHandler)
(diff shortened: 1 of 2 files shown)


dougm:
lgtm

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -269,6 +270,14 @@ func (handler *ContainersHandlersImpl) GetContainerListHandler(params containers
 	return containers.NewGetContainerListOK().WithPayload(vmList)
 }
 
+func (handler *ContainersHandlersImpl) GetContainerLogsHandler(params containers.GetContainerLogsParams) middleware.Responder {
+	defer trace.End(trace.Begin("Containers.PostContainerSignal"))
```
hmahmood: Is this the right trace message?

hmahmood:
Minor comment ... LGTM

lib/apiservers/portlayer/swagger.yml
```diff
@@ -819,6 +819,51 @@ paths:
...
+          in: query
+          type: boolean
+          default: false
+          required: false
+        - name: since
+          in: query
+          type: integer
+          format: int64
+          required: false
+        - name: timestamp
```
fdawg4l: This is a docker'ism, right?  For our PL, we don't have to use their (docker's naming).  I'd rather call this "raw".  We could then JSONify the log output and include timestamps in the body unless raw=1.

Just a thought.

fdawg4l:
LGTM

caglar10ur:
LGTM :sheep: :it:

sflxn:
poke

Issue #1720, opened 2016-07-28
Volume rm not removing the volume


mhagen-vmware:
Running command 'docker -H 192.168.254.172:2375 volume rm test2
${rc} = 0 	
${output} = test2

Running command 'docker -H 192.168.254.172:2375 volume ls
${rc} = 0 	
${output} = DRIVER              VOLUME NAME
vsphere             test
vsphere             test2

fdawg4l:
This is a dupe of #1511.  I'd like to close this unless you want to track the integration test separately @mhagen-vmware .

Issue #1511, opened 2016-07-14
Support `docker volume delete`


fdawg4l:

matthewavery:
This should be resolved. I believe this was among the first of the volume calls to be implemented. I am now hand testing everything to make sure that it works. I will look at writing integration tests now since mike got Volume Create tests running.

mdubya66:
@matthewavery this is an Epic that seems to have two issues remaining, #1508 and #1510 It's not resolved until those are.

fdawg4l:
@matthewavery Sort of.  VolumeDestroy on the vsphere side will return an error "TBD. Not supported yet".  FYI.  If you have the docker client -> docker server -> pl stuff done, then I'm ok closing this issue if we can create an issue to capture filling in the functionality on the vsphere side.

mdubya66:
Moving to Doc for GA as that is the only remaining issue.

stuclem:
Doc is done. Closed https://github.com/vmware/vic/issues/1521.

matthewavery:
@fdawg4l This is ready for closing right?

fdawg4l:
Yup!

Issue #1508, opened 2016-07-14
Add support for `volume delete`


fdawg4l:
This needs to pass the ID given to `docker volume delete <ID>` to `VolumeDestroy` in the port layer.

fdawg4l:
This likely requires work in both the DPS and the PL to shuttle that ID from the user's docker invocation down to the volume layer in the PL.

PR #1760, opened 2016-08-01
Docker volume rm storage handler implementation

matthewavery:
references #1720 #1511 #1508 

1720 - Volume rm not removing the volume
1511 - Support `docker volume delete`
1508 - Add support for `volume delete`

This is the implementation of the handler which is to call `VolumeDestroy` right now that function is `TBD` and once that is complete further work can be done on this PR.

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
@@ -295,7 +295,7 @@ func (handler *StorageHandlersImpl) CreateVolume(params storage.CreateVolumePara
 func (handler *StorageHandlersImpl) RemoveVolume(params storage.RemoveVolumeParams) middleware.Responder {
 	defer trace.End(trace.Begin("storage_handlers.RemoveVolume"))
 
-	err := storageVolumeLayer.VolumeGet(context.TODO(), params.Name)
+	_, err := storageVolumeLayer.VolumeGet(context.TODO(), params.Name)
 	if err != nil {
 		return storage.NewRemoveVolumeNotFound().WithPayload(&models.Error{
 			Message: err.Error(),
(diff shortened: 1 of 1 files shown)


lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -292,9 +292,23 @@ func (handler *StorageHandlersImpl) CreateVolume(params storage.CreateVolumePara
 }
 
 //RemoveVolume : Remove a Volume from existence
-func (handler *StorageHandlersImpl) RemoveVolume(storage.RemoveVolumeParams) middleware.Responder {
+func (handler *StorageHandlersImpl) RemoveVolume(params storage.RemoveVolumeParams) middleware.Responder {
 	defer trace.End(trace.Begin("storage_handlers.RemoveVolume"))
-	return storage.NewRemoveVolumeOK() //TODO: this is just a stub for now.
+
+	_, err := storageVolumeLayer.VolumeGet(context.TODO(), params.Name)
```
fdawg4l: Won't the `destroy` give you an error if the given volume doesn't exist?  

I don't think we need this check since the destroy call itself will have to do more than just a check for existence anyway.

caglar10ur:
LGTM :sheep: :it:

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
hmahmood: +1

matthewavery: alright yeah, I will change this. This will really just be calling `VolumeDestroy` and then type switching over the errors. pretty simple fix.

jzt:
LGTM 🚢 🎱

matthewavery:
♻️

PR #1761, opened 2016-08-01
Change container VM name to prettyname-ID

### LINKED ISSUE #787: Change VM name to prettyName-UUID
Given a docker run command of:
`docker run -name sleepy_hibbert busybox`
resulting in a container ID of:
`DEADBEEF`

the containerVM name should be:
`sleepy-hibbert-DEADBEEF`

ghicken update: There are tokens already defined in virtual_container_host.go for replacement of id, name, and vm name.

emlin:
Fixes #787 

  This will significantly increase the docker inspect time if the container is not in container cache. Cause we cannot search VM by it's ID only.

  Did not change portlayer API to pass in container name at this time, cause in theory we should be able to find container in it's cache. If not, we might need to populate cache while port layer is started.


emlin:
The change in childVM is not easy to be hit in general workflow. 
We don't have container cache update in docker personality server at this time, so I cannot test it by restart VCH.
But after VCH reboot is supported, this function will be useful.

caglar10ur:
@hickeng why do we prefer this format?

andrewtchin:
lgtm

hickeng:
@caglar10ur we had repeated requests from management and customers to change the VM name to the pretty name (SHA based IDs not being considered readable). The more compelling argument I have had from customers is that they use process mandated naming conventions for VMs - this scenario is why the `ContainerNameConvention` field has been in the VCH config since inception.

emlin:
@hickeng looks like we also want to have container vm name convention as well. But what is that looks like?
My PR actually does not include the vic-machine create change for it.

lib/portlayer/exec/container.go
```diff
@@ -393,16 +393,24 @@ func infraContainers(ctx context.Context, sess *session.Session) ([]mo.VirtualMa
 
 // find the childVM for this resource pool by name
 func childVM(ctx context.Context, sess *session.Session, name string) (*vm.VirtualMachine, error) {
-	searchIndex := object.NewSearchIndex(sess.Client.Client)
-	child, err := searchIndex.FindChild(ctx, VCHConfig.ResourcePool.Reference(), name)
```
dougm: I suggest using the container ID in the VmPathName, along the lines of:

```go
spec.Files = &types.VirtualMachineFileInfo{
		VmPathName: fmt.Sprintf("[%s] %s/%d.vmx", datastore, id, id),
}
```

Then here, you can use `SearchIndex.FindByDatastorePath` with the value of VmPathName.

The VM can have any name you choose, can be renamed, etc., and VmPathName will stay the same.

emlin: What if the vm path is renamed by vsphere? for example for vm test, vsphere might use test_1 as the folder name, if it found duplicate.
And do you know if that works for vSAN?

dougm: vSphere won't rename it.  When you specify the full path to .vmx, vSphere will raise an error if the .vmx already exists.
Yes, it works for vSAN.

emlin: @dougm Here is the spec to create container VM:
https://github.com/vmware/vic/blob/master/lib/spec/spec.go#L94
While it's vSAN, looks like we can not specify full vmx path. so It's still possible to rename the folder path, right?

emlin: Actually, it's not renaming issue on vSAN. We didn't have vmx file path configured on vSAN as I mentioned above, so the file path will be prettyname-ID, so it's not possible to query back that VM with ID only.
I tried the datastore path search, which does not support partial match as well.

dougm: Wasn't aware of that issue with vSAN, but not surprised I guess.  Another possible option, if we can use the ID to set the vm Config.InstanceUuid, then use SearchIndex.FindByUuid

We'd may need to convert the ID to uuid using for example: https://godoc.org/github.com/google/uuid#NewSHA1

emlin:
@dougm The code is changed to use searchByUuid, could you please help to have a look?

lib/portlayer/exec/container.go
```diff
@@ -423,16 +425,39 @@ func infraContainers(ctx context.Context, sess *session.Session) ([]mo.VirtualMa
 	return vms, nil
 }
 
+func instanceUUID(id string) (string, error) {
+	// generate VM instance uuid, which will be used to query back VM
+	u, err := sys.UUID()
+	if err != nil {
+		return "", err
+	}
+	NameSpaceDNS, err := uuid.Parse(u)
```
dougm: Let's name this variable 'space' or 'namespace'.

lib/portlayer/exec/container.go
```diff
@@ -423,16 +425,39 @@ func infraContainers(ctx context.Context, sess *session.Session) ([]mo.VirtualMa
...
+	if err != nil {
+		return "", errors.Errorf("uable to parse VCH uuid: %s", err)
```
dougm: typo

lib/spec/spec.go
```diff
@@ -34,12 +34,15 @@ const (
 	pciSlotNumberBegin int32 = 0xc0
 	pciSlotNumberEnd   int32 = 1 << 10
 	pciSlotNumberInc   int32 = 1 << 5
+
+	MaxVMNameLength = 80
```
dougm: Let's not export this if it isn't used outside of the package ('M'->'m')

lib/spec/spec.go
```diff
@@ -99,8 +102,17 @@ func NewVirtualMachineConfigSpec(ctx context.Context, session *session.Session,
...
+	nameMaxLen := MaxVMNameLength - len(config.ID)
+	prettyName := config.Name
+	if len(prettyName) > nameMaxLen-1 {
+		prettyName = prettyName[:nameMaxLen-1]
+	}
+	fullName := fmt.Sprintf("%s-%s", prettyName, config.ID)
 	s := &types.VirtualMachineConfigSpec{
-		Name: config.ID,
+		Name:         fullName,
+		InstanceUuid: config.InstanceUUID,
```
dougm: Note that @hickeng would prefer we use 'Uuid' here instead of 'InstanceUuid'

emlin: @dougm Do you mean set Uuid, instead of InstanceUuid?
But the search is based on instance uuid, instead of uuid, right? How can that work?

dougm: Yes, I meant Uuid.  See: http://pubs.vmware.com/vsphere-60/index.jsp#com.vmware.wssdk.apiref.doc/vim.SearchIndex.html#findByUuid

> If specified, search for virtual machines whose instance UUID matches the given uuid. 
> Otherwise, search for virtual machines whose BIOS UUID matches the given uuid.

emlin: I see. The search interface can query both Bios uuid and instance uuid.
OK, so let me set bios uuid.
thanks

lib/portlayer/exec/container.go
```diff
@@ -423,16 +425,39 @@ func infraContainers(ctx context.Context, sess *session.Session) ([]mo.VirtualMa
...
+	uuid, err := instanceUUID(name)
+	if err != nil {
+		detail := fmt.Sprintf("unable to get instance UUID: %s", err)
+		log.Error(detail)
+		return nil, errors.New(detail)
+	}
+
 	searchIndex := object.NewSearchIndex(sess.Client.Client)
-	child, err := searchIndex.FindChild(ctx, VCHConfig.ResourcePool.Reference(), name)
+	child, err := searchIndex.FindByUuid(ctx, sess.Datacenter, uuid, true, types.NewBool(true))
```
dougm: The last param needs to be 'nil' or 'false' to find by the bios uuid.

dougm:
lgtm

emlin:
♻️

PR #1764, opened 2016-08-01
Abandon images if checksums don't match on pull

### LINKED ISSUE #933: WriteImage vmdks to tmp directory, then move to final location
To guard against half writes or other inconsistencies, we should create the vmdk in `WriteImage` in a (unique) temp directory, then move the result after the data has been extracted and disk unmounted to the final location.

fdawg4l:
Fixes #933 

This change writes the image to a temporary directory, compares the
computed checksum with the expected checksum, and moves the disk to the
image store if they match.  Otherwise the temporary directory is
removed.  Subsequent pulls should complete without an issue.

### DIFF
--- a/lib/portlayer/storage/image.go
+++ b/lib/portlayer/storage/image.go
@@ -72,9 +72,9 @@ type ImageStorer interface {
 	// parent - The parent image to create the new image from.
 	// ID - textual ID for the image to be written
 	// meta - metadata associated with the image
+	// sum - expected sha266 sum of the image content.
 	// r - the image tar to be written
-	WriteImage(ctx context.Context, parent *Image, ID string, meta map[string][]byte, r io.Reader) (*Image,
-		error)
+	WriteImage(ctx context.Context, parent *Image, ID string, meta map[string][]byte, sum string, r io.Reader) (*Image, error)
 
 	// GetImage queries the image store for the specified image.
 	//
--- a/lib/portlayer/storage/vsphere/image.go
+++ b/lib/portlayer/storage/vsphere/image.go
@@ -15,6 +15,7 @@
 package vsphere
 
 import (
+	"crypto/sha256"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -24,12 +25,14 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/docker/pkg/archive"
+	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/types"
 	portlayer "github.com/vmware/vic/lib/portlayer/storage"
 	"github.com/vmware/vic/lib/portlayer/util"
 	"github.com/vmware/vic/pkg/vsphere/datastore"
 	"github.com/vmware/vic/pkg/vsphere/disk"
 	"github.com/vmware/vic/pkg/vsphere/session"
+	"github.com/vmware/vic/pkg/vsphere/tasks"
 	"golang.org/x/net/context"
 )
 
@@ -187,7 +190,7 @@ func (v *ImageStore) ListImageStores(ctx context.Context) ([]*url.URL, error) {
 // ID - textual ID for the image to be written
 // meta - metadata associated with the image
 // Tag - the tag of the image to be written
-func (v *ImageStore) WriteImage(ctx context.Context, parent *portlayer.Image, ID string, meta map[string][]byte,
+func (v *ImageStore) WriteImage(ctx context.Context, parent *portlayer.Image, ID string, meta map[string][]byte, sum string,
 	r io.Reader) (*portlayer.Image, error) {
 
 	storeName, err := util.ImageStoreName(parent.Store)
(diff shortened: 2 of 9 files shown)


fdawg4l:
I need to test this on VSAN yet, but things look fine on standalone esx.

lib/portlayer/storage/vsphere/image.go
```diff
@@ -283,6 +244,119 @@ func (v *ImageStore) WriteImage(ctx context.Context, parent *portlayer.Image, ID
...
+// the disk, unpack the tar, check the checksum.  If the data doesn't match the
+// expected checksum, abort by nuking the tempdir.  If everything matches, move
+// the tmpdir to the expected location (with the vmdk inside it).  The unwind
+// path is a bit convoluted here;  we need to clean up on the way out in the
+// error case (using the tmpdir).
+func (v *ImageStore) writeImage(ctx context.Context, storeName, parentID, ID,
+	sum string, r io.Reader) error {
+
+	// Create the image directory in the store.
+	tmpImageDir := v.imageDirPath(storeName, ID+".inprogress")
```
hmahmood: Multiple downloading clients would not overwrite each other here?

fdawg4l: It would have to be multiple clients downloading the same image. In that case, the right thing to do is for this call to fail for one of the clients. But, that's not very elegant. I'll make an issue to add a map of IDs in flight to check against for new clients to guard against that eventuality. Thanks for bringing it up!

caglar10ur: I think docker is blocking other callers saying download in progress, not sure how...

fdawg4l: The daemon, IIRC, enforces that (probably via the graph driver). In any case, we should guard against it so other personalities get the same protection.

jzt: Docker uses [this](https://github.com/docker/docker/tree/master/distribution/xfer) to basically share the progress across multiple clients during concurrent download of the same layer(s). It has a collection of layers that are in flight, as well as a collection of `watcher`s that are concerned about one or more layers in flight. When a layer's progress gets updated, that `Progress` update struct is sent to each `watcher` watching that particular layer via a channel. 

We will definitely want something like this in the future.

fdawg4l: Created  #1772 to track this.  It's an open issue and not a regression from this change.  We've just never encountered it.

fdawg4l:
Vsan doesn't like moves.  Apparently moves arent actually consistent across stat operations.

mv dir0 dir1 && ls -l dir1  # not happy

I can't use a temp directory with vsan.  So this PR needs some massaging.  I'll use (this)[https://github.com/vmware/govmomi/blob/master/object/virtual_disk_manager.go#L97] instead.

fdawg4l:
OK, using the virtual disk manager did the trick.  It's one more level of indirection, but we're consistent throughout.  Please take another peak when you get the chance.

Tested on VSAN (with vc) and stand alone esx.

cgtexmex:
LGTM

lib/portlayer/storage/vsphere/image.go
```diff
@@ -283,6 +246,124 @@ func (v *ImageStore) WriteImage(ctx context.Context, parent *portlayer.Image, ID
...
+// the tmpdir to the expected location (with the vmdk inside it).  The unwind
+// path is a bit convoluted here;  we need to clean up on the way out in the
+// error case (using the tmpdir).
+func (v *ImageStore) writeImage(ctx context.Context, storeName, parentID, ID,
+	sum string, r io.Reader) error {
+
+	// Create a temp image directory in the store.
+	tmpImageDir := v.imageDirPath(storeName, ID+"inprogress")
+	_, err := v.ds.Mkdir(ctx, true, tmpImageDir)
+	if err != nil {
```
hickeng: How do we address the scenario where we've left the `ID+inprogress` directory around from a previous failed attempt?

lib/portlayer/storage/vsphere/image.go
```diff
@@ -283,6 +246,124 @@ func (v *ImageStore) WriteImage(ctx context.Context, parent *portlayer.Image, ID
...
+	}
+
+	if err = v.dm.Detach(ctx, vmdisk); err != nil {
+		return err
+	}
+
+	// If we've gotten here, prepare the final location for the image
+	imageDir := v.imageDirPath(storeName, ID)
+	_, err = v.ds.Mkdir(ctx, true, imageDir)
+	if err != nil {
```
hickeng: what happens if we previously died at this point, having created the directory but not moved the image in?

lib/portlayer/storage/vsphere/image.go
fdawg4l: The only way for the `ID+inprogress` directory to exist after a `writeImage` call is when the VCH is rebooted in the middle of `writeImage`.  And in that case the image will be added to the image list and one could `rmi` it.

fdawg4l:
Now we write to a tmp vmdk  and move it once the checksum matches.  Also added appropriate cleanup code at startup.

lib/portlayer/storage/vsphere/image.go
```diff
@@ -366,3 +456,41 @@ func (v *ImageStore) ListImages(ctx context.Context, store *url.URL, IDs []strin
...
+
+	for _, fdir := range res.HostDatastoreBrowserSearchResults {
+		if len(fdir.File) == 0 {
+			break
+		}
+
+		log.Infof("Removing in progress image %s", fdir.FolderPath)
+
+		err = tasks.Wait(context.TODO(), func(ctx context.Context) (tasks.Waiter, error) {
+			return fm.DeleteDatastoreFile(ctx, fdir.FolderPath, v.s.Datacenter)
```
matthewavery: Possibly learning experience for me. Should you try to remove all the in progress files before reporting any errors trying to remove any of them? Then only the in progress images that you had problems removing would be left.

fdawg4l: Ah, so you're asking about continuing if there is an error during the `DeleteDatastoreFile`.  That's true and a very good point.  I'll do that.

fdawg4l:
Force pushed.  Instead of nuking image directories with `inprogress.vmdk` files, we're nuking image directories which don't have the expected <ID>.vmdk.  Since moving the vmdk from the `inprogress.vmdk` file to <ID>.vmdk is the last thing we do, the existence of this file means the image is consistent.

lib/portlayer/storage/image.go
```diff
@@ -72,9 +72,9 @@ type ImageStorer interface {
 	// parent - The parent image to create the new image from.
 	// ID - textual ID for the image to be written
 	// meta - metadata associated with the image
+	// sum - expected sha266 sum of the image content.
 	// r - the image tar to be written
-	WriteImage(ctx context.Context, parent *Image, ID string, meta map[string][]byte, r io.Reader) (*Image,
-		error)
+	WriteImage(ctx context.Context, parent *Image, ID string, meta map[string][]byte, sum string, r io.Reader) (*Image, error)
```
caglar10ur: just a general comment, it doesn't need to be addressed at all. 

I personally like using structs as params approach. We could create a WriteImageOptions struct with those field and start passing that. This way we don't end up changing function signatures at all.

caglar10ur:
LGTM

fdawg4l:
.

PR #1765, opened 2016-08-02
Implement docker log in personality and portlayer rest

### LINKED ISSUE #366: Docker API: Docker Container Logs

sflxn:
Implment the streaming support in the docker personality server and portlayer
rest server.  Added some unit testing for container logs.  Also moved some
docker inspect code to container_proxy.go so we can fully mock container logs.
Logs needs to make a container inspect call to check on the container's
running state.  We can't mock that state test without moving some of the
inspect code to container_proxy.go.

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -36,14 +36,14 @@ import (
 	"github.com/docker/docker/api/types/backend"
 	derr "github.com/docker/docker/errors"
 	"github.com/docker/docker/pkg/archive"
+	"github.com/docker/docker/pkg/ioutils"
 	"github.com/docker/docker/pkg/namesgenerator"
 	"github.com/docker/docker/pkg/stdcopy"
-	"github.com/docker/docker/pkg/stringid"
 	"github.com/docker/docker/pkg/version"
 	"github.com/docker/engine-api/types"
 	"github.com/docker/engine-api/types/container"
 	containertypes "github.com/docker/engine-api/types/container"
-	dnetwork "github.com/docker/engine-api/types/network"
+	timetypes "github.com/docker/engine-api/types/time"
 	"github.com/docker/go-connections/nat"
 
 	"github.com/vishvananda/netlink"
@@ -115,7 +115,7 @@ const (
 
 func NewContainerBackend() *Container {
 	return &Container{
-		containerProxy: portlayer.NewContainerProxy(PortLayerClient()),
+		containerProxy: portlayer.NewContainerProxy(PortLayerClient(), PortLayerServer(), PortLayerName()),
 	}
 }
 
@@ -817,7 +817,7 @@ func (c *Container) ContainerInspect(name string, size bool, version version.Ver
 		}
 	}
 
-	inspectJSON, err := containerInfoToDockerContainerInspect(vc, results.Payload)
+	inspectJSON, err := portlayer.ContainerInfoToDockerContainerInspect(vc, results.Payload, PortLayerName())
 	if err != nil {
 		log.Errorf("containerInfoToDockerContainerInspect failed with %s", err)
 		return nil, err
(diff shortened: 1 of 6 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -1699,3 +1391,42 @@ func ContainerSignal(containerID string, sig uint64) error {
...
+	if config.Since != "" {
+		s, n, err := timetypes.ParseTimestamps(config.Since, 0)
+		if err != nil {
+			return 0, 0, false, err
+		}
+		since = time.Unix(s, n)
+	}
+
+	// Check if container is running
+	running, _ := c.containerProxy.ContainerRunning(vc)
```
cgtexmex: Why have this as a check and not an exception on the portlayer side?  So if you request the logs of a stopped container then throw an error instead of checking before hand?

lib/apiservers/engine/backends/portlayer/container_proxy.go
```diff
@@ -248,6 +274,97 @@ func (c *ContainerProxy) CommitContainerHandle(handle, imageID string) error {
 	return nil
 }
 
+// StreamContainerlogs() reads the log stream from the portlayer rest server and writes
+// it directly to the io.Writer that is passed in.
+func (c *ContainerProxy) StreamContainerLogs(name string, out io.Writer, started chan struct{}, showTimestamps bool, followLogs bool, since int64, tailLines int64) error {
+	defer trace.End(trace.Begin("ContainerProxy.StreamContainerLogs"))
```
cgtexmex: I thought @caglar10ur previously mentioned that trace will automatically provide the function call so this shouldn't include the function but other potentially helpful information.... possibly I misunderstood

lib/apiservers/engine/backends/container.go
```diff
@@ -934,347 +967,6 @@ func (c *Container) ContainerAttach(name string, ca *backend.ContainerAttachConf
 	return err
 }
 
-//-------------------------------------
```
cgtexmex: Assuming this is clean up -- just moving chairs around on the deck?  if so why included here and not a PR specifically addressing that goal?

lib/apiservers/engine/backends/container.go
```diff
@@ -836,7 +836,40 @@ func (c *Container) ContainerInspect(name string, size bool, version version.Ver
...
+	close(started)
+	wf.Flush()
+
+	var outStream io.Writer = wf
+	if !vc.Config.Tty {
+		outStream = stdcopy.NewStdWriter(outStream, stdcopy.Stdout)
+	}
+
+	// Make a call to our proxy to handle the remoting
+	err = c.containerProxy.StreamContainerLogs(name, outStream, started, config.Timestamps, follow, since, tailLines)
```
jzt: Two things here that I've been noticing in the container code that are worth mentioning:

1. When we use the cache to convert from a container name -> container ID before making a backend call, we lose the name and are thus unable to present the error to the client in the same way docker does (displaying the name instead of the ID if that's what was supplied by the CLI). This especially obvious in cases where a containerVM has been removed out-of-band (using the GUI, for example). I've been storing off the `name` param in its original state and using that when returning `Not Found` type errors to the cli.

2. If the `err` we get back here happens to some type of container `NotFound` error (again, caused by some out-of-band container removal), we should delete the container from the container cache so later calls to `docker create --name mycontainer` won't return with `Already exists` type errors.

Don't worry about it here, I have a task open right now to address this, but I thought it was worth bringing to your attention.

lib/apiservers/engine/backends/container.go
sflxn: Moving it is described in the PR description.  Had to move to our new container_proxy, which implements an interface that makes unit test mocking possible.  Any operation that needs to be mocked has to go into that file and function proto added to the interface.  That also includes utility function it relies on.

lib/apiservers/engine/backends/container.go
sflxn: It wasn't always this way.  At one point, we were using vc.ContainerID, but a few PR back, there was a suggestion that we overwrite name for stylistic reason.  I didn't care so I started adopting it, but if it taking away functionality, then we should stop doing it.

lib/apiservers/engine/backends/container.go
sflxn: Sure, I'll probably change this since we are following that pattern of blind calling portlayer and letting it determine whether the operation is possible or not elsewhere in the personality server.

lib/apiservers/engine/backends/container.go
jzt: LGTM

jzt:
LGTM

lib/apiservers/engine/backends/container.go
sflxn: btw, my goal is to eventually refactor all the personality code and move all remoting code to _proxy.go.  Right now, only remoting code for Create and some of inspect is there.  Once I am done with logs and fix some attach issues, I'm going to go back and go through all of container.go and do the same, and attempt to add unit tests for the other operations.

lib/apiservers/engine/backends/portlayer/container_proxy.go
sflxn: Just looked at the trace code.  You're right.  Guess, we'll need to remove all those everywhere also.

dougm:
lgtm

sflxn:
poke

PR #1766, opened 2016-08-02
Unbind containers if they are no longer running

### LINKED ISSUE #1749: network ls crashes docker personality after running container
The sequence of commands:

```
docker run -it busybox
docker network ls
```
crashes docker personality.

hmahmood:
Fixes #1749

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -235,20 +235,23 @@ func (handler *ContainersHandlersImpl) GetContainerInfoHandler(params containers
 	defer trace.End(trace.Begin("Containers.GetContainerInfoHandler"))
 
 	// get the container id for interogation
-	containerID := exec.ParseID(params.ID)
-	cc, err := exec.ContainerInfo(context.Background(), handler.handlerCtx.Session, containerID)
+	h, err := exec.GetInfraContainer(context.Background(), handler.handlerCtx.Session, exec.ParseID(params.ID))
 	if err != nil {
 		log.Debugf("GetContainerInfoHandler Error: %s", err.Error())
 		return containers.NewGetContainerInfoNotFound().WithPayload(&models.Error{Message: err.Error()})
 	}
-	info := convertContainerToContainerInfo(cc)
+	info := convertContainerToContainerInfo(h)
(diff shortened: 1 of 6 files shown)


lib/portlayer/exec/container.go
```diff
@@ -280,69 +301,65 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
...
-		return nil, fmt.Errorf("%s does not appear to be a container", containerID)
+		return fmt.Errorf("not a container")
 	case 1:
 		// we have a winner
-		return cc[0], nil
-	default:
-		// we manged to find multiple vms
-		return nil, fmt.Errorf("multiple containers named %s found", containerID)
+		cc[0].vm = c.vm
+		c.swap(cc[0])
```
emlin: why do you need this swap? Is this to update the cached container information based on VM information?

hmahmood: Yes; we update the config and the state, as well as the vm moref.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -261,7 +264,8 @@ func (handler *ContainersHandlersImpl) GetContainerListHandler(params containers
 		info.ContainerID = &container.ExecConfig.ID
 		info.LayerID = &container.ExecConfig.LayerID
 		info.Created = &container.ExecConfig.Created
-		info.Status = &container.Status
+		info.Status = new(string)
+		*info.Status = container.State.String()
```
hickeng: Strings are immutable so:
```
info.Status = &container.State.String()
```

dougm: @hickeng Don't think this needs to change, it is consistent with our existing pattern where we deal with 'string *' fields.  And `... = &container.State.String()` won't compile.

dougm:
lgtm

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
hickeng: Sorry - apparently I'm idiotic today.

caglar10ur:
LGTM :sheep: :it:

hmahmood:
This does not really fix the problem as further testing has revealed. This will have to be fixed after events are implemented in the port layer.

PR #1773, opened 2016-08-02
Add container signal delivery

### LINKED ISSUE #1267: Port layer: signal
Deliver a signal to a specific container/session.
The signal should be specified as an ssh.Signal to avoid confusion as to interpretation.

dougm:
Port layer now uses the vSphere API to send the "kill" command to the
container VM guest toolbox.

Closes #1267

### DIFF
--- a/lib/portlayer/exec/container.go
+++ b/lib/portlayer/exec/container.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/vmware/govmomi/guest"
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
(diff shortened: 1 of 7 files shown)


caglar10ur:
LGTM :sheep: :it:

lib/portlayer/exec/container.go
```diff
@@ -240,6 +241,37 @@ func (c *Container) stop(ctx context.Context) error {
...
+}
+
+func (c *Container) Signal(ctx context.Context, num int64) error {
+	defer trace.End(trace.Begin("Container.Signal"))
+
+	if c.vm == nil {
+		return fmt.Errorf("vm not set")
+	}
+
+	return c.startGuestProgram(ctx, "kill", fmt.Sprintf("%d", num))
```
sflxn: Huh, so this is one of the toolbox's program and not linux's kill, right?   The reason I ask is you're calling kill with the args == signal value.

dougm: That's right, it is a toolbox method call, not shelling out to the kill command.

sflxn:
lgtm

dougm:
♻️

Issue #1229, opened 2016-06-27
Add "healthy" setup checks into vic-admin web page.


mlh78750:
I recently setup a new ESX and forgot to change my firewall rules so VIC was not working properly. I would be nice to be able to go to the vicadmin web page and get a "red/yellow/green" indicator of the setup of VIC in that environment. I'm thinking of something where it shows if all the hosts have the firewall rules set properly, the available space in the datastore being used, ability to connect to public repos, etc.. 

So in my case: If I'd gone to the vicadmin page to pull logs because attach was not working I would have seen an ESX listed as not having a correct firewall configuration and I would have known my issue was my configuration. 

Having this information would also be useful for the log bundles to provide context of how "healthy" the environment is at the time.

lweitzman:
this sounds like great data for the VIC overview page. but from what i understand, the overview page is not part of v1.0.

mdubya66:
we've asked our crack group of interns to take a look at vicadmin. we will, opportunistically, improve the interface for 1.0 but will not hold the release.

PR #1775, opened 2016-08-02
Enable Firewall Checking in vicadmin

kjplatz:
Enables the "Firewall" checkbox on the vicadmin portal.  In the event the firewall is enabled and misconfigured on a host, the checkbox will change from a green check to orange warning triangle, and an appropriate warning message will be posted.  Partially addresses #1229.

### DIFF
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -255,7 +255,8 @@ func (s *server) tailFiles(res http.ResponseWriter, req *http.Request, names []s
 func (s *server) index(res http.ResponseWriter, req *http.Request) {
 	defer trace.End(trace.Begin(""))
 	ctx := context.Background()
-	v := vicadmin.NewValidator(ctx, vchConfig)
+	sess, err := client()
+	v := vicadmin.NewValidator(ctx, &vchConfig, sess)
 
 	tmpl, err := template.ParseFiles("dashboard.html")
 	err = tmpl.ExecuteTemplate(res, "dashboard.html", v)
--- a/lib/install/validate/config.go
+++ b/lib/install/validate/config.go
@@ -28,10 +28,9 @@ import (
 	"golang.org/x/net/context"
 )
 
-func (v *Validator) CheckFirewall(ctx context.Context) []error {
-	fwissues := []error{}
+func (v *Validator) CheckFirewall(ctx context.Context) {
 	if v.DisableFirewallCheck {
-		return fwissues
+		return
 	}
 	defer trace.End(trace.Begin(""))
 
@@ -47,12 +46,13 @@ func (v *Validator) CheckFirewall(ctx context.Context) []error {
 
 	errMsg := "Firewall check SKIPPED"
 	if !v.sessionValid(errMsg) {
-		return fwissues
+		return
 	}
 
 	if hosts, err = v.Session.Datastore.AttachedClusterHosts(ctx, v.Session.Cluster); err != nil {
 		log.Errorf("Unable to get the list of hosts attached to given storage: %s", err)
-		return append(fwissues, err)
+		v.NoteIssue(err)
+		return
 	}
 
 	var misconfiguredEnabled []string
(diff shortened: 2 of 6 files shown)


cmd/vicadmin/server.go
```diff
@@ -255,7 +255,8 @@ func (s *server) tailFiles(res http.ResponseWriter, req *http.Request, names []s
 func (s *server) index(res http.ResponseWriter, req *http.Request) {
 	defer trace.End(trace.Begin(""))
 	ctx := context.Background()
-	v := vicadmin.NewValidator(ctx, vchConfig)
+	sess, err := client()
```
emlin: Actually, install/validator.NewValidator also did more than to create a session only. It will figure out the problem if datacenter is not correctly set.
Would you need that function?

lib/install/validate/validator.go
```diff
@@ -56,13 +58,25 @@ type Validator struct {
 	Session *session.Session
 	Context context.Context
 
-	isVC   bool
-	issues []error
+	isVC           bool
+	firewallIssues []error
```
emlin: I would rather not to add a specific issues type here. Following comments will show how to do.

lib/install/validate/validator.go
```diff
@@ -178,6 +200,11 @@ func (v *Validator) Validate(ctx context.Context, input *data.Data) (*config.Vir
 
 }
 
+func (v *Validator) ReValidate(ctx context.Context) {
```
emlin: You could just write ValidateFirewall if you need additional check here. Otherwise, just public firewall method.
Then you could call the method anytime you want, and then get out issues listed in Validator.

lib/install/validate/validator.go
```diff
@@ -178,6 +200,11 @@ func (v *Validator) Validate(ctx context.Context, input *data.Data) (*config.Vir
...
+	v.firewall(ctx)
+	v.license(ctx)
```
emlin: Same thing for license check

andrewtchin:
lgtm

cmd/vicadmin/server.go
kjplatz: install/validator.NewValidator actually takes the input.Data (as specified on the command-line) and constructs a vch config spec from that.  In this case I already have the VCH config and only need to run certain segments of the install validation.

emlin: Then it's fine.

lib/install/validate/config.go
```diff
@@ -46,13 +47,12 @@ func (v *Validator) firewall(ctx context.Context) {
 
 	errMsg := "Firewall check SKIPPED"
 	if !v.sessionValid(errMsg) {
-		return
+		return issues
```
emlin: don't need to return issues here. You can call v.ListIssues() to get the issues generated in this method

lib/install/validate/validator.go
```diff
@@ -34,6 +34,8 @@ import (
 )
 
 type Validator struct {
+	config.VirtualContainerHostConfigSpec
```
emlin: it's not used

lib/install/validate/config.go
kjplatz: Changed per discussion with @emlin

lib/install/validate/validator.go
kjplatz: Removed.

lib/install/validate/validator.go
```diff
@@ -222,6 +240,10 @@ func (v *Validator) target(ctx context.Context, input *data.Data, conf *config.V
 			v.NoteIssue(fmt.Errorf("Error processing target after transformation to SOAP endpoint: %q: %s", v.Session.Service, err))
 			return
 		}
+
+		// ESXi requires user/password to be encoded in the Target URL
```
emlin: The connection parameter is set here: https://github.com/vmware/vic/blob/master/lib/install/management/appliance.go#L435
Please reference what port-layer did: https://github.com/vmware/vic/blob/master/lib/install/management/appliance.go#L494
The URL does not lose password information if it's there before serialize.

emlin: As portlayer is gonna change it's behavior, it's good here. Thanks @kjplatz for clarify this.

emlin:
lgtm 👍

PR #1780, opened 2016-08-02
fix TestConcurrency test

caglar10ur:
/cc @hickeng


hickeng:
LGTM
much prefer the Add(1) in the loop - much less error prone.

dougm:
lgtm

PR #1793, opened 2016-08-03
Implements uuid naming for volumes without a specified name on create…

### LINKED ISSUE #1548: Add support for random naming of volumes
Currently `docker volume create` requires that the `--name` option be specified for a successful interaction to occur. Docker behavior is that a volume with `--name` when created will get a random uuid as a name.

matthewavery:
Fixes #1548 

This adds a simple uuid generation for the name of a volume if `docker volume create` is invoked without the `--name` option. Currently we fail the operation and return a swagger error. Instead we will return a uuid. Or if we want a random name if we so choose.

### DIFF
--- a/lib/apiservers/engine/backends/volume.go
+++ b/lib/apiservers/engine/backends/volume.go
@@ -24,6 +24,7 @@ import (
 	derr "github.com/docker/docker/errors"
 
 	"github.com/docker/engine-api/types"
+	"github.com/google/uuid"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/storage"
 	"github.com/vmware/vic/lib/apiservers/portlayer/models"
 	"github.com/vmware/vic/pkg/trace"
@@ -96,8 +97,6 @@ func (v *Volume) VolumeCreate(name, driverName string, opts, labels map[string]s
 	defer trace.End(trace.Begin("Volume.VolumeCreate"))
 	result := &types.Volume{}
 
-	// TODO: design a way to have better error returns.
-
 	client := PortLayerClient()
 	if client == nil {
 		return nil, derr.NewErrorWithStatusCode(fmt.Errorf("Failed to get a portlayer client"), http.StatusInternalServerError)
@@ -110,7 +109,10 @@ func (v *Volume) VolumeCreate(name, driverName string, opts, labels map[string]s
 		return result, derr.NewErrorWithStatusCode(fmt.Errorf("Bad Driver Arg: %s", varErr), http.StatusBadRequest)
 	}
 
-	// TODO: setup name randomization if name == nil
+	if model.Name == "" {
+		volumeUUID, _ := uuid.NewUUID()
+		model.Name = volumeUUID.String()
+	}
 
 	res, err := client.Storage.CreateVolume(storage.NewCreateVolumeParams().WithVolumeRequest(model))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


matthewavery:
small pr, requesting input. cc @jzt  @fdawg4l @sflxn @anchal-agrawal @cgtexmex

lib/apiservers/engine/backends/volume.go
```diff
@@ -110,21 +109,24 @@ func (v *Volume) VolumeCreate(name, driverName string, opts, labels map[string]s
...
 		switch err := err.(type) {
 
 		case *storage.CreateVolumeInternalServerError:
 			// FIXME: right now this does not return an error model...
-			return result, derr.NewErrorWithStatusCode(fmt.Errorf("Server error from Portlayer: %s", err.Error()), http.StatusInternalServerError)
+			return result, derr.NewErrorWithStatusCode(fmt.Errorf("%s", err.Error()), http.StatusInternalServerError)
 
 		case *storage.CreateVolumeDefault:
-			return result, derr.NewErrorWithStatusCode(fmt.Errorf("Server error from Portlayer: %s", err.Payload.Message), http.StatusInternalServerError)
+			return result, derr.NewErrorWithStatusCode(fmt.Errorf("%s", err.Payload.Message), http.StatusInternalServerError)
```
hmahmood: Do these match the errors we get from docker?

lib/apiservers/engine/backends/volume.go
```diff
@@ -110,21 +109,24 @@ func (v *Volume) VolumeCreate(name, driverName string, opts, labels map[string]s
 		return result, derr.NewErrorWithStatusCode(fmt.Errorf("Bad Driver Arg: %s", varErr), http.StatusBadRequest)
 	}
 
-	// TODO: setup name randomization if name == nil
+	if model.Name == "" {
+		volumeUUID, _ := uuid.NewUUID()
+		model.Name = volumeUUID.String()
```
jzt: Why not just `model.Name = uuid.NewUUID().String()`? :)

jzt:
LGTM

lib/apiservers/engine/backends/volume.go
matthewavery: it has a multiple return sadly. so there is an err var that must be ignored.

lib/apiservers/engine/backends/volume.go
matthewavery: I removed the "error from port layer server" since we talked about the error reporting and we don't really want to see `error from daemon server: error from portlayer server :`

hmahmood: I mean these errors get propagated to the docker client command line; do we match the errors we get from regular docker?

matthewavery: Unfortunately according to api v1.23 we the volume api only returns a `201` for `no error` and a `500` for `error from daemon server` and the 500 return string can change from time to time since it is meant as a custom return i believe. I will take a look at the what docker returns, it is likely we will have different messages since we are only dealing with block devices.

hmahmood:
lgtm

lib/apiservers/engine/backends/volume.go
dougm: Can we not use random uuid?  `uuid.New().String()`

Or does this need to be a version 1 uuid?

matthewavery: we can use that. I would rather use that as well.

matthewavery:
♻️

PR #1804, opened 2016-08-03
Feature/disable cleanup temporarily

mhagen-vmware:
Need to keep VCHs around if we hit the issue of not being able to gather logs.


dougm:
lgtm

fdawg4l:
lgtm

PR #1805, opened 2016-08-04
Remove systemd-networkd from appliance

### LINKED ISSUE #1791: network unreachable on appliance
This is the original error that caused #1788.

The vic-machine basic create test failed due to a failed pull subsequent to an install. We've been having issues with VCH not responding and a corresponding failure to collect log bundles so hopefully this explains why. Due to the #1788 failure this VCH was left lying around and I was able to collect the logs after the fact:

[vic-logs.zip](https://github.com/vmware/vic/files/398648/vic-logs.zip)

hickeng:
Removes the dependency on systemd-networkd which seems to have been
disrupting our link management in the appliance and causing "network
unreachable" errors

Fixes #1791


fdawg4l:
LGTM

dougm:
lgtm

PR #1811, opened 2016-08-04
Fix tailFiles in vicadmin (and several other fixes)

### LINKED ISSUE #1744: vic-machine create: update installation output to show vic-admin portal address

### LINKED ISSUE #1754: The Tailf function does not check if the file exists before trying to open it
A check is needed so that we do not try to open and seek a non existent file. Which could happen in certain use case scenarios of VIC. 

line 306 of `cmd/vicadmin/logs.go 
`f, err := os.Open(file)`

### LINKED ISSUE #1810: Vic Admin not showing Font Awesome visuals
image

kjplatz:
Fixes #1810
Fixes #1754 
Fixes #1744 

This also adds the following functionality/enhancements:
* Includes the hpcloud/tail package for use in tailFiles (per @hickeng) as a vendored package
* Tweaks the Makefile to fix dependencies to detect changes in HTML files under isos/vicadmin
* Moves the vicadmin user creation from appliance-staging.sh to appliance.sh to speed builds
* Changes hardcoded references to docker ports to use docker/docker/opts.DefaultHTTPPort


cmd/vicadmin/logs.go
```diff
@@ -302,55 +303,75 @@ func zipEntries(readers map[string]entryReader, out *zip.Writer) error {
...
-			nlines--
+
+	// If the file exists, we want to go back tailLines lines
+	// and pass that new offset into the TailFile() constructor
+	f, err := os.Open(file)
+	if err == nil {
+		nlines := tailLines
+		pos, err := f.Seek(0, 2)
+		if err == nil {
+			b := make([]byte, 1)
```
fdawg4l: If I'm reading this right, if the file is open you seek/read 1 byte at a time until you get to `\n`.  As written, this works, but using [Index](https://golang.org/pkg/bytes/#Index) with a larger buffer may do the same with less seeking/reading.

cmd/vicadmin/server.go
```diff
@@ -120,6 +115,7 @@ func (s *server) serve() error {
 
 	s.mux.Handle("/css/", http.StripPrefix("/css/", http.FileServer(http.Dir("css/"))))
 	s.mux.Handle("/images/", http.StripPrefix("/images/", http.FileServer(http.Dir("images/"))))
+	s.mux.Handle("/fonts/", http.StripPrefix("/fonts/", http.FileServer(http.Dir("fonts/"))))
```
fdawg4l: Not a fan of this and definitely not a fan of checking in a binary file.  Do we strictly need this?

isos/appliance.sh
```diff
@@ -83,6 +83,14 @@ ln -s /etc/systemd/system/vic.target $(rootfs_dir $PKGDIR)/etc/systemd/system/mu
 rm -f $(rootfs_dir $PKGDIR)/etc/systemd/network/*
 cp ${DIR}/base/no-dhcp.network $(rootfs_dir $PKGDIR)/etc/systemd/network/
 
+#
+# Set up vicadmin user
+#
+
+chroot $(rootfs_dir $PKGDIR) groupadd -g 1000 vicadmin
```
fdawg4l: `chroot` just to add a line to the groups file sounds awkward.  How bout just using `echo >>`

isos/appliance.sh
```diff
@@ -83,6 +83,14 @@ ln -s /etc/systemd/system/vic.target $(rootfs_dir $PKGDIR)/etc/systemd/system/mu
...
+chroot $(rootfs_dir $PKGDIR) useradd -u 1000 -g 1000 -m -d /home/vicadmin -s /bin/false vicadmin
```
fdawg4l: Same comment.  This isn't necessary. Just edit /etc/passwd directly.

cmd/vicadmin/logs.go
kjplatz: @fdawg4l Thanks for the suggestion.  I'll add that in there.

cmd/vicadmin/server.go
kjplatz: Yeah, we do.  If the user doesn't have already have the fontawesome fonts installed locally, then they will get colored boxes instead of the desired icons

isos/appliance.sh
kjplatz: Funny thing, @caglar10ur insisted that I use groupadd and useradd for this instead of using echo.

cmd/vicadmin/logs.go
```diff
@@ -302,55 +304,105 @@ func zipEntries(readers map[string]entryReader, out *zip.Writer) error {
...
-		}
-		if b[0] == '\n' {
-			nlines--
+
+	// If the file exists, we want to go back tailLines lines
+	// and pass that new offset into the TailFile() constructor
+	// Per @fdawg4l, use bytes.LastIndex() and a 1k buffer to reduce
+	// seeks/reads
+	f, err := os.Open(file)
+	if err == nil {
```
hmahmood: You can just say
```
if err != nil {
  return err 
}
```
to avoid that indent that runs all the way down.

cmd/vicadmin/logs.go
```diff
@@ -302,55 +304,105 @@ func zipEntries(readers map[string]entryReader, out *zip.Writer) error {
...
+						bufend = off
+					}
+				}
+				lastnl = readPos + int64(bufend)
+			}
+			spos = tail.SeekInfo{
+				Offset: lastnl,
+				Whence: 0,
+			}
+/*
```
hmahmood: Not taking this out?

kjplatz: Took out the commented code.  Should have done that already.

cmd/vicadmin/logs.go
kjplatz: Actually it's not returning an error.  In the event the file doesn't exists or we can't seek to EOF, we still want to do something with it.  We just don't calculate a seek position and start our tailing at beginning-of-file.

hmahmood: Then I would say move this code to a separate function. It is pretty confusing code to boot; please add comments. A unit test would also be great.

isos/appliance.sh
hmahmood: +1 for sticking with useradd and groupadd.

lib/vicadmin/validate.go
```diff
@@ -69,5 +74,14 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
 	log.Info(fmt.Sprintf("FirewallStatus set to: %s", v.FirewallStatus))
 	log.Info(fmt.Sprintf("FirewallIssues set to: %s", v.FirewallIssues))
 
+	//Retrieve Host IP Information and Set socker Endpoint
```
hmahmood: s/socker/socket

cmd/vicadmin/logs.go
hmahmood: There is also this: https://github.com/landley/toybox/blob/master/toys/posix/tail.c#L95

kjplatz: @hmahmood moved this to a function and commented per request.  There is already unit testing in place for the "tail" functionality.

isos/appliance.sh
caglar10ur: Noooo... Use the tools as intended :)

lib/vicadmin/validate.go
kjplatz: @hmahmood correction: s/socker/Docker/

cmd/vicadmin/server.go
caglar10ur: Can't we use some other font or something else (like words with colors, Success/Fail etc.)?

cmd/vicadmin/logs.go
```diff
@@ -302,55 +304,99 @@ func zipEntries(readers map[string]entryReader, out *zip.Writer) error {
 }
 
 func tailFile(wr io.Writer, file string, done *chan bool) error {
-	nlines := tailLines
-	f, err := os.Open(file)
-	if err != nil {
-		return err
-	}
+	defer trace.End(trace.Begin(""))
```
caglar10ur: defer trace.End(trace.Begin(file)) could be more usefull

isos/vicadmin/dashboard.html
```diff
@@ -68,7 +68,7 @@ <h3 class="card-title">Status</h3>
               </div>
               <div class="card card-block">
                 <h3 class="card-title">Virtual Container Host Info</h3>
-                <p class="card-text">Docker Endpoint: 183.28.75.125:2376</p>
+                <p class="card-text">Docker Endpoint: {{.HostIP}}:{{.DockerPort}}</p>
```
caglar10ur: 👍

cmd/vicadmin/server.go
kjplatz: Using these icons keeps us consistent with the Clarity UI standards.

lib/install/management/appliance.go
```diff
@@ -458,10 +459,10 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 
 	if conf.HostCertificate != nil {
```
caglar10ur: you can do 
```
a=1
b=2
c=3
if somecond {
    a=x 
    b=y  
    c=z
} ```

and remove the else altogether

File: lib/install/management/inspect.go [hunk 11]
```diff
@@ -49,10 +50,10 @@ func (d *Dispatcher) InspectVCH(vch *vm.VirtualMachine, conf *config.VirtualCont
 	log.Debug("IP address for client interface: %s", d.HostIP)
 	if !conf.HostCertificate.IsNil() {
 		d.VICAdminProto = "https"
-		d.DockerPort = "2376"
+		d.DockerPort = fmt.Sprintf("%d", opts.DefaultTLSHTTPPort)
```
caglar10ur: ^ditto

caglar10ur:
minor comments otherwise LGTM

hmahmood:
lgtm

kjplatz:
Bing

PR #1814, opened 2016-08-04
Remove systemd-networkd from bootstrap

hickeng:
Given we've removed systemd-networkd from the appliance because the tether
is doing link management, there's no need for it in bootstrap either. This
also removes the multi-user target as the login services are entirely
unnecessary for containerVMs.

Towards: #1724


dougm:
lgtm

caglar10ur:
LGTM :sheep: :it:

One question, we can still start ssh after this change right?

hickeng:
@caglar10ur This is for the containerVM - the change for the appliance has already gone in and actually enables ssh by default again (for now - see my comment on your related PR).

hickeng:
poke

Issue #1799, opened 2016-08-03
make dns tests more resilient to network problems and enable them


caglar10ur:
```
--- FAIL: TestForwarding (2.00s)
    dns_test.go:67: Exchange failed: read udp 127.0.0.1:47725->127.0.0.1:5354: i/o timeout
=== RUN   TestVIC
--- FAIL: TestVIC (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
    panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x4878fa]

goroutine 4 [running]:
panic(0x10852e0, 0xc820012120)
    /usr/local/go/src/runtime/panic.go:481 +0x3e6
testing.tRunner.func1(0xc8201ee360)
    /usr/local/go/src/testing/testing.go:467 +0x192
panic(0x10852e0, 0xc820012120)
    /usr/local/go/src/runtime/panic.go:443 +0x4e9
github.com/vmware/vic/lib/dns.(*Server).Addr(0x0, 0x0, 0x0)
    github.com/vmware/vic/lib/dns/_test/_obj_test/dns.go:183 +0x1ca
github.com/vmware/vic/lib/dns.TestVIC(0xc8201ee360)
    /drone/src/github.com/vmware/vic/lib/dns/dns_test.go:143 +0xa78
testing.tRunner(0xc8201ee360, 0x1e168d0)
    /usr/local/go/src/testing/testing.go:473 +0x98
created by testing.RunTests
    /usr/local/go/src/testing/testing.go:582 +0x892
exit status 2
FAIL    github.com/vmware/vic/lib/dns    2.016s
make: *** [test] Error 1
Makefile:223: recipe for target 'test' failed
[info] build failed (exit code 2)
```

caglar10ur:
Just to make it clear, there is nothing wrong with the current tests. This bug is adding some resiliency for the network problems so that tests won't panic. But they will continue to fail under those conditions.

PR #1820, opened 2016-08-04
Disable DNS tests as they fail with CI

caglar10ur:


mhagen-vmware:
lgtm

lib/dns/dns_test.go
```diff
@@ -80,6 +82,8 @@ func TestForwarding(t *testing.T) {
 }
 
 func TestVIC(t *testing.T) {
+	t.Skipf("Failing with CI")
```
hickeng: can you instead check the error from `bind` and report that - we cannot currently tell if this is a cascade failure from prior test (cannot reuse port?) or a separate problem.

hickeng:
LGTM - please raise a second PR immediately with these re-enabled so we can iterate on a fix for them

mdubya66:
Did an issue get raised to re-enable?

caglar10ur:
https://github.com/vmware/vic/issues/1799

PR #1824, opened 2016-08-04
Feature/trap kill signal

mhagen-vmware:
Trap the drone kill signal and attempt clean up the VCH that might be in the middle of being used still.


mhagen-vmware:
This doesn't appear to work and is actually implemented in drone 0.5.  Closing for now and will implement when we upgrade drone.

PR #1826, opened 2016-08-05
Wait for tools initialization in tether

dougm:
This avoid races where we start a container then issue a
'docker kill -s SIG', as we are unable to service such a request until
the vmx considers the toolbox to be running.

### DIFF
--- a/pkg/vsphere/toolbox/service_test.go
+++ b/pkg/vsphere/toolbox/service_test.go
@@ -279,7 +279,7 @@ func TestServiceRunESX(t *testing.T) {
 	ping := sync.NewCond(new(sync.Mutex))
 
 	service.RegisterHandler("ping", func(b []byte) ([]byte, error) {
-		go ping.Broadcast()
+		ping.Broadcast()
 		return service.Ping(b)
 	})
 
--- a/pkg/vsphere/toolbox/service_test.go
+++ b/pkg/vsphere/toolbox/service_test.go
@@ -24,6 +24,8 @@ import (
 	"net"
 	"sync"
 	"testing"
+
+	"github.com/vmware/govmomi/vim25/types"
 )
 
 func TestDefaultIP(t *testing.T) {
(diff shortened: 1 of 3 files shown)


pkg/vsphere/toolbox/service_test.go
```diff
@@ -288,6 +298,17 @@ func TestServiceRunESX(t *testing.T) {
 		}
 	}
 
+	if *testOn == string(types.VirtualMachinePowerStatePoweredOff) {
+		wg.Add(1)
```
hickeng: are we assured of receiving a power on event irrespective of any timing issues if the VM wasn't powered on initially? In the main path there's a timeout but not in the test.

pkg/vsphere/toolbox/service_test.go
```diff
@@ -273,10 +276,17 @@ func TestServiceRunESX(t *testing.T) {
 
 	service := NewService(in, out)
 
+	ping := sync.NewCond(new(sync.Mutex))
+
+	service.RegisterHandler("ping", func(b []byte) ([]byte, error) {
+		go ping.Broadcast()
```
hickeng: why is broadcast done as a goroutine?

hickeng:
Couple of questions for clarification, but unless there's a need for timeout in the test, LGTM

pkg/vsphere/toolbox/service_test.go
dougm: I've received the PowerOn RPC in every run of that test so far.  But we only need to wait to handle the case of a quick start+kill, so added the timeout just in case.

pkg/vsphere/toolbox/service_test.go
dougm: Since the Wait() call blocks the calling goroutine, but this will be called from the toolbox.Service's goroutine not the test's - so it isn't needed, I'll remove that.

mdubya66:
lgtm

dougm:
:recycle:

PR #1835, opened 2016-08-05
Fix out-of-band integration tests

dougm:
PR #1761 changed how VMs are named, we need to use the new name format
for govc based out-of-band operations.

### DIFF
--- a/tests/test-cases/Group6-VIC-Machine/6-3-Delete.robot
+++ b/tests/test-cases/Group6-VIC-Machine/6-3-Delete.robot
@@ -13,7 +13,7 @@ Initial load
     Should Be Equal As Integers  ${rc}  0
     Should Not Contain  ${output}  Error
     ${name}=  Generate Random String  15
-    ${rc}  ${container-id}=  Run And Return Rc And Output  docker ${params} create --name ${name} busybox
+    ${rc}  ${container-id}=  Run And Return Rc And Output  docker ${params} create --name ${name} busybox /bin/top
     Should Be Equal As Integers  ${rc}  0
     Should Not Contain  ${container-id}  Error
     ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} start ${container-id}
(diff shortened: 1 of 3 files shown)


mdubya66:
lgtm

tests/test-cases/Group6-VIC-Machine/6-3-Delete.robot
```diff
@@ -12,19 +12,20 @@ Initial load
...
+    Set Suite Variable  ${containerName}  ${name}
     
 Delete VCH and verify
     # Get VCH uuid and container VM uuid, to check if resources are removed correctly
     Run Keyword And Ignore Error  Gather Logs From Test Server
     ${uuid}=  Run  govc vm.info -json\=true ${vch-name} | jq -r '.VirtualMachines[0].Config.Uuid'
-    Run  govc vm.power -on=true ${containerID}
+    ${rc}  ${output}=  Run And Return Rc And Output  govc vm.power -on=true ${containerName}-*
     ${ret}=  Run  bin/vic-machine-linux delete --target %{TEST_URL} --user %{TEST_USERNAME} --password=%{TEST_PASSWORD} --compute-resource=%{TEST_RESOURCE} --name ${vch-name}
     Should Contain  ${ret}  is powered on
```
hickeng: I don't think this is reliable - given we're creating busybox without arguments it'll be running /bin/ash. Without specifying -i stdin should be closed and the container should have exited.
If that's not what's happening then it's a different behavioural problem and this test is still failure prone

tests/test-cases/Group8-vSphere-Integration/8-1-GuestTools.robot
```diff
@@ -14,32 +14,34 @@ Verify container VM guest IP is reported
...
 Stop container VM using guest shutdown
     ${rc}=  Run And Return Rc  docker ${params} pull busybox
     Should Be Equal As Integers  ${rc}  0
-    ${rc}  ${id}=  Run And Return Rc And Output  docker ${params} run -d busybox /bin/top
+    ${name}=  Generate Random String  15
+    ${rc}  ${id}=  Run And Return Rc And Output  docker ${params} run --name ${name} -d busybox /bin/top
     Should Be Equal As Integers  ${rc}  0
-    Run  govc vm.ip ${id}
-    ${rc}=  Run And Return Rc  govc vm.power -s ${id}
+    ${rc}  ${output}=  Run And Return Rc And Output  govc vm.power -s ${name}-*
```
hickeng: does this check the resulting power state or just whether the shutdown was delivered correctly?

dougm: Just testing that we are able to issue a GuestShutdown (-s), which would fail if toolbox wasn't running and properly configured.  The 'docker stop' flavor of this test does wait for the power state to be off.  But no reason we can't do that here too, I'll add it.

tests/test-cases/Group6-VIC-Machine/6-3-Delete.robot
dougm: I'll add the -i

hmahmood:
lgtm

dougm:
poke

PR #1854, opened 2016-08-08
Fixing pointer serialization/deserialization

### LINKED ISSUE #1853: extraconfig package pointer serialization/deserialization is not symmetric
When serializing a `nil` pointer, the extraconfig package will skip writing its key to guestinfo. However, when deserializing the corresponding pointer, if the key does not exist in guestinfo, extraconfig will still allocate an object to assign to the pointer.

hmahmood:
Fixes #1853 

Not required for 0.5.0


caglar10ur:
LGTM, @hickeng you might want to take a look at this PR

pkg/vsphere/extraconfig/decode.go
```diff
@@ -334,7 +331,7 @@ func decodeMap(src DataSource, dest reflect.Value, prefix string, depth recursio
 	mapkeys, err := src(prefix)
 	if mapkeys == "" || err != nil {
 		log.Debugf("No value found in data source for maps keys \"%s\"", prefix)
-		return this
+		return reflect.ValueOf(nil)
```
hickeng: This requires every reference to a map element, pointer or otherwise, to check for nil. I'd prefer to have an empty map returned. If you must absolutely have a nil pointer in the case of a pointer to a map, then check for map size explicitly.

hmahmood: This is returned from an unexported function, and I tried to make all of the `decodeX` functions follow the same pattern, i.e. return reflect.ValueOf(nil) in case of error. Returning a nil map would disrupt this pattern, since we can't return for every type that we decode in case of error.

I will add some tests for the error cases.

hickeng: That precise issue is why I hadn't implemented the nil pointer logic yet. Perhaps we can add a boolean field to decode that denotes whether it's hung from a pointer or not. That would allow us to control on a type by type basis whether you get zero value or nil pointer.

hmahmood:
@caglar10ur @hickeng can you guys take a look at this again please? Thanks.

pkg/vsphere/extraconfig/store.go
```diff
@@ -43,7 +42,7 @@ func (t *MapStore) Get(key string) (string, error) {
 
 	val, ok := t.store[key]
 	if !ok {
-		return "", os.ErrNotExist
+		return "", ErrKeyNotFound
```
caglar10ur: Do we have any callers type asserting for os.ErrNotExist?

caglar10ur:
Overall LGTM, https://github.com/vmware/vic/pull/1854/files#r74861264 is the only comment I have

pkg/vsphere/extraconfig/store.go
hmahmood: I made an attempt to find all callers and fix them. Don't think I missed any in our source at least.

hickeng:
Thanks! LGTM

hmahmood:
:recycle:

PR #1858, opened 2016-08-08
vic-machine ls

### LINKED ISSUE #1550: vic-machine ls
Usage sample:
```
vic-machine ls
-FROM=https://root@****:vcenter/

ID        Path                                Notes
vm-239    mycluster/edw-pool/web/edw-web      Notes from the VCH config
vm-2372   mycluster/random/someother          Another VCH in the random pool
+vm-15    mycluster/xyz                       VCH in the root of the cluster
```
The following example lists VCHs under a specific resource
```
vic-machine ls
-FROM=https://root@****:vcenter/example-datacenter/host/mycluster/Resources/edw-pool/
```

emlin:
Fixes #1550 
This PR add command 'vic-machine ls', also unit test, integration test for it.
And also update inspect command to accept --id got from vic-machine ls command.

### DIFF
--- a/cmd/vic-machine/common/id.go
+++ b/cmd/vic-machine/common/id.go
@@ -28,7 +28,7 @@ func (i *VCHID) IDFlags() []cli.Flag {
 		cli.StringFlag{
 			Name:        "id",
 			Value:       "",
-			Usage:       "The ID of the Virtual Container Host",
+			Usage:       "The ID of the Virtual Container Host, e.g. vm-220",
 			Destination: &i.ID,
 		},
 	}
--- a/cmd/vic-machine/common/id.go
+++ b/cmd/vic-machine/common/id.go
@@ -16,8 +16,6 @@ package common
 
 import (
 	"github.com/urfave/cli"
-
-	log "github.com/Sirupsen/logrus"
 )
 
 type VCHID struct {
@@ -30,16 +28,12 @@ func (i *VCHID) IDFlags() []cli.Flag {
 		cli.StringFlag{
 			Name:        "id",
 			Value:       "",
-			Usage:       "The ID of the Virtual Container Host - not supported until vic-machine ls is ready",
+			Usage:       "The ID of the Virtual Container Host",
 			Destination: &i.ID,
 		},
 	}
 }
 
 func (i *VCHID) ProcessID() error {
-	if i.ID != "" {
-		log.Warnf("ID of Virtual Container Host is not supported until vic-machine ls is ready. For details, please refer github issue #810")
-		return cli.NewExitError("Can specify --compute-resource and --name", 1)
-	}
 	return nil
 }
--- a/cmd/vic-machine/common/id.go
+++ b/cmd/vic-machine/common/id.go
@@ -33,7 +33,3 @@ func (i *VCHID) IDFlags() []cli.Flag {
 		},
 	}
 }
-
-func (i *VCHID) ProcessID() error {
-	return nil
-}
--- a/cmd/vic-machine/list/list.go
+++ b/cmd/vic-machine/list/list.go
@@ -43,7 +43,7 @@ const templ = `{{range .}}
 {{.ID}}	{{.Path}}	{{.Name}}{{end}}
 `
 
-// Delete has all input parameters for vic-machine delete command
+// List has all input parameters for vic-machine ls command
 type List struct {
 	*data.Data
 
@@ -56,7 +56,7 @@ func NewList() *List {
 	return d
 }
 
-// Flags return all cli flags for delete
+// Flags return all cli flags for ls
 func (l *List) Flags() []cli.Flag {
 	flags := append(l.TargetFlags(), l.ComputeFlagsNoName()...)
 	flags = append(flags, l.DebugFlags()...)
--- a/cmd/vic-machine/list/list.go
+++ b/cmd/vic-machine/list/list.go
@@ -38,6 +38,7 @@ type items struct {
 	Name string
 }
 
+// templ is parsed by text/template package
 const templ = `{{range .}}
 {{.ID}}	{{.Path}}	{{.Name}}{{end}}
 `
@@ -82,7 +83,7 @@ func (l *List) prettyPrint(cli *cli.Context, ctx context.Context, vchs []*vm.Vir
 		parentPath := path.Dir(path.Dir(vch.InventoryPath))
 		name := path.Base(vch.InventoryPath)
 		data = append(data,
-			items{vch.Reference().String(), parentPath, name})
+			items{vch.Reference().Value, parentPath, name})
 	}
 	t := template.New("vic-machine ls")
 	t, _ = t.Parse(templ)
@@ -105,7 +106,6 @@ func (l *List) Run(cli *cli.Context) error {
 	}
 
 	if len(cli.Args()) > 0 {
-		log.Error("List cannot continue: invalid CLI arguments")
 		log.Errorf("Unknown argument: %s", cli.Args()[0])
 		return errors.New("invalid CLI arguments")
 	}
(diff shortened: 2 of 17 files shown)


emlin:
The ls command output looks like following:
```
INFO[2016-08-08T13:26:10-05:00] ### Listing VCHs ####                        

ID                           PATH                                                   NAME
VirtualMachine:vm-189        /dc1/host/cluster1/Resources/line/line1/line1-2        line1-2-1
```

cmd/vic-machine/common/id.go
```diff
@@ -30,16 +28,12 @@ func (i *VCHID) IDFlags() []cli.Flag {
 		cli.StringFlag{
 			Name:        "id",
 			Value:       "",
-			Usage:       "The ID of the Virtual Container Host - not supported until vic-machine ls is ready",
+			Usage:       "The ID of the Virtual Container Host",
```
hickeng: Is this the full `VirtualMachine:vm-xyz` or just the last portion? If it's just the last portion that that's what we should output in the `ls` output for the ID field.
If this is the full : separated string then I think we should change it to be just `vm-xyz` as it will _always_ be prefixed by VirtualMachine

cmd/vic-machine/list/list.go
```diff
@@ -0,0 +1,141 @@
...
+	"golang.org/x/net/context"
+)
+
+type items struct {
+	ID   string
+	Path string
+	Name string
+}
+
+const templ = `{{range .}}
```
hickeng: This format string needs commenting - with either the details of the formatting or the pkg that does the processing

cmd/vic-machine/list/list.go
```diff
@@ -0,0 +1,141 @@
...
+	if err := l.HasCredentials(); err != nil {
+		return err
+	}
+
+	l.Insecure = true
+	return nil
+}
+
+func (l *List) prettyPrint(cli *cli.Context, ctx context.Context, vchs []*vm.VirtualMachine) {
+	data := []items{
```
hickeng: I think it'd be extremely useful to have a comment field that could be set during create - just a brief freeform description. Not for this PR but with any significant number of VCHs to manage it could become critical

cmd/vic-machine/list/list.go
```diff
@@ -0,0 +1,141 @@
...
+		return err
+	}
+
+	if l.Debug.Debug > 0 {
+		log.SetLevel(log.DebugLevel)
+		trace.Logger.Level = log.DebugLevel
+	}
+
+	if len(cli.Args()) > 0 {
+		log.Error("List cannot continue: invalid CLI arguments")
```
hickeng: This line adds no value - the only reason to keep this would be for consistency with the other commands where a similar comment would apply.

cmd/vic-machine/common/id.go
emlin: Currently is using full name prefixed by VirtualMachine, in ls and inspect command.
Will change both of them.

cmd/vic-machine/list/list.go
emlin: I know that in your design, there is one Notes column, but as we still didn't support that argument in create, I didn't add that column in ls command as well.
Should we just add one column here, but leave it empty? Or you'd like to give one suggested argument, to add into create command?

cmd/vic-machine/list/list.go
emlin: yeah, it is text/template

andrewtchin:
lgtm

cmd/vic-machine/list/list.go
hickeng: Ignore this comment - I have a bad habit of referencing future work.

cmd/vic-machine/common/id.go
```diff
@@ -30,16 +28,12 @@ func (i *VCHID) IDFlags() []cli.Flag {
...
 			Name:        "id",
 			Value:       "",
-			Usage:       "The ID of the Virtual Container Host - not supported until vic-machine ls is ready",
+			Usage:       "The ID of the Virtual Container Host, e.g. vm-220",
 			Destination: &i.ID,
 		},
 	}
 }
 
 func (i *VCHID) ProcessID() error {
```
hmahmood: Do we need this anymore?

emlin: Not really at this time.

cmd/vic-machine/list/list.go
```diff
@@ -0,0 +1,141 @@
...
+	Path string
+	Name string
+}
+
+// templ is parsed by text/template package
+const templ = `{{range .}}
+{{.ID}}	{{.Path}}	{{.Name}}{{end}}
+`
+
+// Delete has all input parameters for vic-machine delete command
```
hmahmood: Need to update this

cmd/vic-machine/list/list.go
```diff
@@ -0,0 +1,141 @@
...
+	executor *management.Dispatcher
+}
+
+func NewList() *List {
+	d := &List{}
+	d.Data = data.NewData()
+	return d
+}
+
+// Flags return all cli flags for delete
```
hmahmood: Ditto

lib/install/management/finder.go
```diff
@@ -150,3 +153,155 @@ func (d *Dispatcher) GetVCHConfig(vm *vm.VirtualMachine) (*config.VirtualContain
...
+		return nil, err
+	}
+
+	var vchs []*vm.VirtualMachine
+	for _, dc := range dcs {
+		dcVCHs, err := d.searchVCHsPerDC(dc)
+		if err != nil {
+			return nil, err
+		}
+		vchs = append(vchs, dcVCHs...)
```
hmahmood: Any chance we may have duplicates?

lib/install/management/finder.go
```diff
@@ -150,3 +153,155 @@ func (d *Dispatcher) GetVCHConfig(vm *vm.VirtualMachine) (*config.VirtualContain
...
+		return nil, err
+	}
+	pools = append(pools, children...)
+	var vchs []*vm.VirtualMachine
+	for _, parent := range pools {
+		chidren, err := d.getChildVCHs(parent, true)
+		if err != nil {
+			return nil, err
+		}
+		vchs = append(vchs, chidren...)
```
hmahmood: Any chance of dups?

pkg/vsphere/compute/rp.go
```diff
@@ -84,26 +84,16 @@ func (rp *ResourcePool) GetChildrenVMs(ctx context.Context, s *session.Session)
...
-		if vmName, err = vmm.Name(ctx); err != nil {
-			err = errors.Errorf("Failed to get VM name of %s, %s", vmm.Reference(), err)
-			log.Errorf("%s", err)
-			return nil, err
-		}
-		if vmName == name {
-			return vmm, nil
-		}
+	if child == nil {
+		return nil, nil
```
hmahmood: This is weird; why would be here if err == nil?

hmahmood:
Minor comments ... lgtm

lib/install/management/finder.go
emlin: VCH is VM object, which is unique in whole VC.

lib/install/management/finder.go
emlin: ditto^

pkg/vsphere/compute/rp.go
emlin: We're searching VM has same name with parent resource pool, that's the pre-requisition for a VCH.
If such VM does not exist, searchIndex returns nil result, with nil err. So we do same thing.

emlin:
♻️

emlin:
♻️

Issue #1837, opened 2016-08-05
Tether nil pointer during shutdown


hickeng:
Failure during vic-machine create tests (https://ci.vmware.run/vmware/vic/3703)

From containerVM console:
![image](https://cloud.githubusercontent.com/assets/3923729/17451801/8a7ed89e-5b1e-11e6-9662-7f52e7bef103.png)

No log bundle available due to #1836 so contianer debug log is:
```

ISOLINUX 4.05 0x4f92e181  Copyright (C) 1994-2011 H. Peter Anvin et al
INFO[0000] Started reaping child processes
INFO[0000] Launching pprof server on port 6060
INFO[0000] Starting extension Attach
INFO[0000] opening ttyS0 for backchannel
INFO[0000] s = &terminal.State{termios:syscall.Termios{Iflag:0x500, Oflag:0x5, Cflag:0xcbd, Lflag:0x8a3b, Line:0x0, Cc:[32]uint8{0x3, 0x1c, 0x7f, 0x15, 0x4, 0x0, 0x1, 0x0, 0x11, 0x13, 0x1a, 0x0, 0x12, 0xf, 0x17, 0x16, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}, Pad_cgo_0:[3]uint8{0x0, 0x0, 0x0}, Ispeed:0x0, Ospeed:0x0}}
INFO[0000] creating raw connection from ttyS0 (fd=9)

INFO[0000] Starting extension Toolbox
INFO[0000] toolbox: waiting for initialization
INFO[0000] toolbox: service is ready (power on event received)
INFO[0000] Loading main configuration
INFO[0000] Loaded executor config
INFO[0000] setting ip address 172.16.0.2/16 for link eth0
INFO[0000] Setting default gateway to 172.16.0.1
INFO[0000] updated default route to bridge interface, gateway: 172.16.0.1
INFO[0000] Added nameservers: [172.16.0.1]
INFO[0000] Launching process for session 39539aa2299e5540b1483071eedd234fc7a82a3d4c9da72d161e8ec79da2d17f
INFO[0000] opening ttyS2 for session logging
INFO[0000] Passing config to Attach
INFO[0000] Processing config for session 39539aa2299e5540b1483071eedd234fc7a82a3d4c9da72d161e8ec79da2d17f
[...]

Issue #1836, opened 2016-08-05
nil pointer in vicadmin


hickeng:
Unable to get a log bundle due to nil pointers in vicadmin:

```
root@VCH-3703-1037 [ /var/log/vic ]# cat vicadmin.log
time="2016-08-05T21:35:36Z" level=info msg="Launching vicadmin server on 127.0.0.1:6061"
time="2016-08-05T21:35:36Z" level=info msg="Current UID/GID = 1000/1000"
time="2016-08-05T21:35:36Z" level=debug msg="[ END ] [main.init.1] [278.254908ms] "
time="2016-08-05T21:35:36Z" level=debug msg="[BEGIN] [main.(*server).listen]"
time="2016-08-05T21:35:36Z" level=debug msg="[ END ] [main.(*server).listen] [2.890002ms] "
time="2016-08-05T21:35:36Z" level=info msg="listening on :2378"
time="2016-08-05T21:35:36Z" level=debug msg="[BEGIN] [main.(*server).serve]"
time="2016-08-05T21:35:36Z" level=debug msg="[BEGIN] [main.(*server).handleFunc]"
time="2016-08-05T21:35:36Z" level=debug msg="[ END ] [main.(*server).handleFunc] [1.13µs] "
time="2016-08-05T21:35:36Z" level=debug msg="[BEGIN] [main.(*server).handleFunc]"
time="2016-08-05T21:35:36Z" level=debug msg="[ END ] [main.(*server).handleFunc] [728ns] "
time="2016-08-05T21:35:36Z" level=debug msg="[BEGIN] [main.logFiles]"
time="2016-08-05T21:35:36Z" level=debug msg="[ END ] [main.logFiles] [4.19µs] "
time="2016-08-05T21:35:36Z" level=debug msg="[BEGIN] [main.(*server).handleFunc]"
time="2016-08-05T21:35:36Z" level=debug msg="[ END ] [main.(*server).handleFunc] [635ns] "
time="2016-08-05T21:35:36Z" level=debug msg="[BEGIN] [main.(*server).handleFunc]"
time="2016-08-05T21:35:36Z" level=debug msg="[ END ] [main.(*server).handleFunc] [707ns] "
time="2016-08-05T21:35:36Z" level=debug msg="[BEGIN] [main.(*server).handleFunc]"
time="2016-08-05T21:35:36Z" level=debug msg="[ END ] [main.(*server).handleFunc] [686ns] "
[...]

Issue #1823, opened 2016-08-04
Error response from daemon: server error from portlayer : [DELETE /containers/{id}][500] containerRemoveInternalServerError


mhagen-vmware:
Looks like failed to stop the container properly before trying to remove failed:
[integration_test_logs_3662_1470347704.zip](https://github.com/vmware/vic/files/402800/integration_test_logs_3662_1470347704.zip)

```
time="2016-08-04T21:53:51Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/portlayer/exec.(*Container).stop] Container.stop" 
time="2016-08-04T21:53:51Z" level=error msg="HandshakeClient: failed to read expected SYN-ACK: n=2, err=read tcp 192.168.31.148:8080->192.168.31.12:10735: i/o timeout buf=[0x1515]" 
time="2016-08-04T21:53:52Z" level=error msg="HandshakeClient: failed to read expected SYN-ACK: n=1, err=read tcp 192.168.31.148:8080->192.168.31.12:10735: i/o timeout buf=[0x15]" 
time="2016-08-04T21:53:52Z" level=error msg="HandshakeClient: failed to read expected SYN-ACK: n=1, err=read tcp 192.168.31.148:8080->192.168.31.12:10735: i/o timeout buf=[0x15]" 
time="2016-08-04T21:53:52Z" level=debug msg="HandshakeClient: did not receive synack: 0x0697 != 0x1506" 
time="2016-08-04T21:53:52Z" level=debug msg="discarding following 1 bytes from input channel\n" 
time="2016-08-04T21:53:52Z" level=debug msg="[48]\n" 
time="2016-08-04T21:53:52Z" level=debug msg="caught EOF" 
time="2016-08-04T21:53:52Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/portlayer/exec.(*Container).stop] [698.372953ms] Container.stop" 
time="2016-08-04T21:53:52Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/portlayer/exec.(*Container).Commit] [901.171181ms] Committing handle" 
time="2016-08-04T21:53:52Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/apiservers/portlayer/restapi/handlers.(*ContainersHandlersImpl).CommitHandler] [902.095679ms] Containers.CommitHandler" 
[...]

Issue #1822, opened 2016-08-04
Error response from daemon: server error from portlayer : context deadline exceeded 


mreferre:
I am using a recent build (1 week) old and I keep hitting what appears to be timeout issues. 

This is a good example of it: 

     root@ubuntu:/go/src/github.com/vmware/vic/bin# docker --tls run -d vmwarecna/nginx
     561914bca648562c43df58c5b513693be422c13776bd9f1d6192c0b15cd3ec56
     root@ubuntu:/go/src/github.com/vmware/vic/bin# docker --tls run -d vmwarecna/nginx
     f646d5b158813f77af2a4ac0fe13bd36dc616397d35af515aed54ede7f0da64b
     root@ubuntu:/go/src/github.com/vmware/vic/bin# docker --tls run -d vmwarecna/nginx
     4578cfe14f4e8e1dd279a2ff445b5d5d3090ffb34aaad226db65f91bf7b6b474
     docker: Error response from daemon: server error from portlayer : context deadline exceeded.

When this happens (randomly) that container appears to be running (ps) but you can't do much with it (can't stop, can't rm, etc). The only workaround I found is to power it off in the vSphere UI, at which point you can rm it (or even start it!). 

Gut feeling is that this happens more often with bigger images (perhaps the deployment time takes longer than expected and it times out?). 

Everything else keeps working fine (docker info, docker ps, docker images, etc). Only container operations on containers that experienced that issue will fail. 

I am using a nested environment (with a VM providing shared storage via NFS). I know it could be slow but I am not looking for performance, just functionalities.

mreferre:
After further testing in my (nested) environment I am leaning to conclude that the problem is due to using the NFS shared storage (provided by a VM running alongside the ESXi hosts). 

To exercise the setup I am using a script that deploys 10 containers: 

      for i in {0..9}
      do 
      echo -n "instantianting container: nginx$i - " & docker --tls run -d -p 8$i:80 --name nginx$i vmwarecna/nginx
      done


This is the result of a text script that deployes 10 x nginx when the VCH is on the NFS shared storage:

      root@ubuntu:~# ./multiple-containers-create.sh 
      instantianting container: nginx0 - c730cc6f18ded6584f1171afbf202922982db9e2e033a310a808f27109e6123a
      instantianting container: nginx1 - f869a1ff2ee46d75402baa60211ee53fc2772be711a4e12c8fc37e979d425a32
      docker: Error response from daemon: server error from portlayer : context deadline exceeded.
[...]

PR #1869, opened 2016-08-08
Just big sleep to allow VM to shut off and hopefully avoid intermitte…

mhagen-vmware:
We believe that there is a race between docker stop and docker rm. Because we have not implemented containerWait yet, our docker rm implementation immediately tries to rm a vm that might still be powered on.

### DIFF
--- a/tests/resources/Util.robot
+++ b/tests/resources/Util.robot
@@ -267,6 +267,7 @@ Run Regression Tests
     ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} ps -a
     Should Be Equal As Integers  ${rc}  0
     Should Contain  ${output}  Stopped
+    Sleep  30
     ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} rm ${container}
     Should Be Equal As Integers  ${rc}  0
     ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} ps -a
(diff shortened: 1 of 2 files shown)


mdubya66:
@hickeng : this requires your review before merge.

mdubya66:
potential workaround for #1822 and #1823

hickeng:
Firstly that's a VERY long timeout. I'm not sure how many minutes that'll add to the integration tests but it'll be a lot.
Secondly I don't think it'll actually work - I believe there are two problems: #1870 and another one in the containerVM itself, possibly #1837). The latter is hard to confirm as we need the container logs which is currently blocked by #1836)

mhagen-vmware:
1. It will add 30 sec to each PR run
2. It will add 2 minutes to a merge to master run

I don't like the big sleep either, but it doesn't add that much to our development process and unless all those issues are getting fixed today, I really don't want to sit through another day of CI failures. I think this is a relatively easy and limited potential fix that has little cost and either improves things or costs us a small amount of time.  And it is really easy to remove once we have the other fixes in place.  @hickeng

mhagen-vmware:
@hickeng How about this: https://github.com/vmware/vic/pull/1869/commits/77d68ae036bef6775d95d5a807f41aae36a3c543

Poll govc?

mdubya66:
@hickeng @mhagen-vmware where are we with this?

caglar10ur:
LGTM :sheep: :it:

PR #1882, opened 2016-08-09
Fix incorrect assumption for bootstrap.iso path

### LINKED ISSUE #1351: Incorrect formulation of bootstrap iso path

caglar10ur:
The ISO path that's generated is based off the display name, not the
actual VM folder path. This pr fixes that also removes the hardcoded
appliance.iso file name.

Fixes #1351 2nd time

### DIFF
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -338,7 +338,7 @@ func (d *Dispatcher) GenerateExtensionName(conf *config.VirtualContainerHostConf
 	return nil
 }
 
-func (d *Dispatcher) configIso(conf *config.VirtualContainerHostConfigSpec, vm *vm.VirtualMachine) (object.VirtualDeviceList, error) {
+func (d *Dispatcher) configIso(conf *config.VirtualContainerHostConfigSpec, vm *vm.VirtualMachine, settings *data.InstallerData) (object.VirtualDeviceList, error) {
 	defer trace.End(trace.Begin(""))
 
 	var devices object.VirtualDeviceList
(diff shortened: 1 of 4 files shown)


caglar10ur:
Created https://github.com/vmware/vic/issues/1883 for the integration test part

fdawg4l:
LGTM

caglar10ur:
```
[conur@vir:/opt/go/src/github.com/vmware/vic(iso)] govc datastore.ls ZzZ/*.iso
b.iso
a.iso
[conur@vir:/opt/go/src/github.com/vmware/vic(iso)] govc datastore.download ZzZ/ZzZ.vmx ZzZ.vmx; grep iso ZzZ.vmx
[09-08-16 10:34:30] Downloading... OK
48:ide0:0.fileName = "a.iso"
102:guestinfo./bootstrap_image_path = "[vsanDatastore] 980faa57-97c1-5f89-4578-000c298a9889/b.iso"
[conur@vir:/opt/go/src/github.com/vmware/vic(iso)]
```

lib/install/management/appliance.go
```diff
@@ -359,7 +359,7 @@ func (d *Dispatcher) configIso(conf *config.VirtualContainerHostConfigSpec, vm *
 		log.Errorf("Failed to create Cdrom device for appliance: %s", err)
 		return nil, err
 	}
-	cdrom = devices.InsertIso(cdrom, fmt.Sprintf("[%s] %s/appliance.iso", conf.ImageStores[0].Host, d.vmPathName))
+	cdrom = devices.InsertIso(cdrom, fmt.Sprintf("[%s] %s/%s", conf.ImageStores[0].Host, d.vmPathName, settings.ApplianceISO))
```
hickeng: Should we be using the actual path of the appliance VM rather than composing in this fashion. It'll currently work because we use ImageStores[0] as the applianceVM datastore, but if that ever changes then this will break.

hickeng:
LGTM other than future proofing.

lib/install/management/appliance.go
caglar10ur: conf.ImageStores[0] used everywhere in the vic-machine code so will open a issue to address that if needed

emlin:
Caglar, this change is good. 
Just wondering if we should custom the iso files in datastore.
I mean we uploaded the file here https://github.com/vmware/vic/blob/master/lib/install/management/create.go#L122, with the filename from local file system, would it be better to use "appliance.iso" and "bootstrap.iso" during uploading, no matter how user name them?
then we could assume them in our code as a constant name.

PR #1895, opened 2016-08-09
Use WithContext variant of the swagger params

caglar10ur:
Uses context.TODO() for now. We should do one more pass to properly pass ctx around.


fdawg4l:
LGTM

dougm:
lgtm

PR #1903, opened 2016-08-10
Switch to the more github friendly zip and remove the ESX log gathering

mhagen-vmware:
I think we have the reproduction of the ESX issue and logs that we need, so removing for now and switching to the zip bundle so we can upload directly to github now.


matthewavery:
lgtm

dougm:
lgtm

PR #1913, opened 2016-08-10
Make sure that env variable is set before using it

mhagen-vmware:
Just need to make sure that the env variables are set properly before trying to use them


gigawhitlocks:
LGTM

emlin:
lgtm

Issue #1832, opened 2016-08-05
Relocate port layer container type and container cache


cgtexmex:
Currently the Container type and cache resides in the exec portion of the port layer.  Refactor so that they are both accessible from all portions of the port layer.

This will provide a common Container Type for the port layer and a centralized cache.

hmahmood:
This is not done yet. I just exported the cache object; this calls for moving it into its own package where other PL components can use it.

Issue #1701, opened 2016-07-27
docker ps does not show ports


hmahmood:
```
hmahmood@ubuntu:~/go/src/github.com/vmware/vic$ docker -H 172.16.127.133:2376 --tls ps
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS               NAMES
074002e44685        nginx               "nginx -g daemon off;"   16 hours ago        Running                                 happy_boyd
hmahmood@ubuntu:~/go/src/github.com/vmware/vic$ docker -H 172.16.127.133:2376 --tls port happy_boyd
80/tcp -> 0.0.0.0:8081
```

stuclem:
This is fixed. Removing kind/note.

PR #1914, opened 2016-08-10
PortLayer event stream

### LINKED ISSUE #1329: Integrate VC event stream with Exec Layer
We need to integrate the VC event stream for the following:
* network
  * unbind of IP address on power off
  * update of DNS
* exec
  * implementation of Wait
  * efficient reporting of container state
* interaction
  * reject pending requests for attach sessions

cgtexmex:
Initial event stream for PortLayer

Change is focused on exec and will update the container cache with out of band activity.  ContainerList (utilized by docker ps) & ContainerInfo (utilized by docker inspect) have been refactored to take advantage of an up to date container cache.  

This change does not encompass network and will not release network resources
for containers are acted on out of band.  That will be addressed in a future
change.

Fixes: #1329

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -242,26 +242,23 @@ func (handler *ContainersHandlersImpl) RemoveContainerHandler(params containers.
 func (handler *ContainersHandlersImpl) GetContainerInfoHandler(params containers.GetContainerInfoParams) middleware.Responder {
 	defer trace.End(trace.Begin("Containers.GetContainerInfoHandler"))
 
-	// get the container id for interogation
-	containerID := uid.Parse(params.ID)
-	cc, err := exec.ContainerInfo(context.Background(), handler.handlerCtx.Session, containerID)
-	if err != nil {
-		log.Debugf("GetContainerInfoHandler Error: %s", err.Error())
-		return containers.NewGetContainerInfoNotFound().WithPayload(&models.Error{Message: err.Error()})
+	container := exec.ContainerInfo(params.ID)
+	if container == nil {
+		info := fmt.Sprintf("GetContainerInfoHandler ContainerCache miss for container(%s)", params.ID)
+		log.Error(info)
+		return containers.NewGetContainerInfoNotFound().WithPayload(&models.Error{Message: info})
 	}
-	info := convertContainerToContainerInfo(cc)
-	return containers.NewGetContainerInfoOK().WithPayload(info)
+
+	containerInfo := convertContainerToContainerInfo(container)
+	return containers.NewGetContainerInfoOK().WithPayload(containerInfo)
 }
 
 func (handler *ContainersHandlersImpl) GetContainerListHandler(params containers.GetContainerListParams) middleware.Responder {
 	defer trace.End(trace.Begin("Containers.GetContainerListHandler"))
 
-	containerVMs, err := exec.List(context.Background(), handler.handlerCtx.Session, params.All)
-	if err != nil {
-		return containers.NewGetContainerListInternalServerError()
-	}
-
+	containerVMs := exec.Containers(*params.All)
 	vmList := make([]models.ContainerListInfo, 0, len(containerVMs))
+
 	for i := range containerVMs {
 		// convert to return model
 		container := containerVMs[i]
@@ -269,7 +266,8 @@ func (handler *ContainersHandlersImpl) GetContainerListHandler(params containers
 		info.ContainerID = &container.ExecConfig.ID
 		info.LayerID = &container.ExecConfig.LayerID
 		info.Created = &container.ExecConfig.Created
-		info.Status = &container.Status
+		state := container.State.String()
+		info.Status = &state
 		info.Names = []string{container.ExecConfig.Name}
 		info.ExecArgs = container.ExecConfig.Sessions[*info.ContainerID].Cmd.Args
 		info.StorageSize = &container.VMUnsharedDisk
@@ -323,9 +321,8 @@ func convertContainerToContainerInfo(container *exec.Container) *models.Containe
 	ccid := container.ExecConfig.ID
 	info.ContainerConfig.ContainerID = &ccid
 
-	// TODO: need to determine an appropriate state model
-	// for now leveraging the status used in ps
-	info.ContainerConfig.State = &container.Status
+	s := container.State.String()
+	info.ContainerConfig.State = &s
 	info.ContainerConfig.LayerID = &container.ExecConfig.LayerID
 	info.ContainerConfig.RepoName = &container.ExecConfig.RepoName
 	info.ContainerConfig.Created = &container.ExecConfig.Created
--- a/lib/portlayer/event/event_manager.go
+++ b/lib/portlayer/event/event_manager.go
@@ -1,0 +1,52 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package event
+
+import (
+	"github.com/vmware/vic/pkg/vsphere/session"
+)
+
+type EventManager interface {
+
+	// AddMonitoredObject will add the object for event listening
+	AddMonitoredObject(ref string) error
+
+	// RemoveMonitoredObject will remove the object from event listening
+	RemoveMonitoredObject(ref string)
+
+	// Register will for event callbacks
+	Register(caller string, callback func(Event, *session.Session))
+
+	// Unregsiter from event callbacks
+	Unregister(caller string)
+
+	// Registry will return the callback map
+	Registry() map[string]func(Event, *session.Session)
+
+	// RegistryCount returns the count of callbacks
+	RegistryCount() int
+
+	// Start listening for events
+	Start() error
+
+	// Stop listening for events
+	Stop()
+
+	// Blacklist will identify an object to be omitted from event callbacks
+	Blacklist(ref string) error
+
+	// Unblacklist will remove an object that was previously blacklisted
+	Unblacklist(ref string)
+}
--- a/lib/portlayer/event/events.go
+++ b/lib/portlayer/event/events.go
@@ -1,0 +1,37 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package event
+
+const (
+	ContainerCreated      = "Created"
+	ContainerShutdown     = "Shutdown"
+	ContainerPoweredOn    = "PoweredOn"
+	ContainerPoweredOff   = "PoweredOff"
+	ContainerSuspended    = "Suspended"
+	ContainerResumed      = "Resumed"
+	ContainerRemoved      = "Removed"
+	ContainerReconfigured = "Reconfigured"
+)
+
+type Event interface {
+	// id of event
+	EventID() int
+	// event (PowerOn, PowerOff, etc)
+	String() string
+	// reference evented object
+	Reference() string
+	// event message
+	Message() string
+}
--- a/lib/portlayer/event/vsphere/manager.go
+++ b/lib/portlayer/event/vsphere/manager.go
@@ -1,0 +1,225 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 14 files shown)


lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -277,7 +273,8 @@ func (handler *ContainersHandlersImpl) GetContainerListHandler(params containers
 		info.ContainerID = &container.ExecConfig.ID
 		info.LayerID = &container.ExecConfig.LayerID
 		info.Created = &container.ExecConfig.Created
-		info.Status = &container.Status
+		friendly := container.FriendlyStatus()
```
hmahmood: How about `conatiner.Status.String()`, i.e. implement the [Stringer](https://github.com/vmware/vic/issues/1701) interface for `Status`?

lib/portlayer/event/vsphere/manager.go
```diff
@@ -0,0 +1,232 @@
...
+	"github.com/vmware/vic/lib/portlayer/event"
+	"github.com/vmware/vic/pkg/vsphere/session"
+
+	vmwEvents "github.com/vmware/govmomi/event"
+	"github.com/vmware/govmomi/vim25/types"
+
+	"golang.org/x/net/context"
+)
+
+type vSphereManager struct {
```
hmahmood: Does this object need to be thread safe?

lib/portlayer/event/vsphere/manager.go
```diff
@@ -0,0 +1,232 @@
...
+	return nil
+}
+
+// function will process the event and call the necessary registered functions
+//
+// Initial implmentation will only act on certain events -- future implementations
+// may provide more flexibilty
+func evented(mgr *vSphereManager, page []types.BaseEvent) error {
+	var err error
+	if mgr.RegistryCount() > 0 {
```
hmahmood:
```
if mgr.RegistryCount() > 0 {
  log.Debugf(...)
  return nil
}
```

lib/portlayer/event/vsphere/manager.go
cgtexmex: possibly an addition for the future.  This is a pretty controlled launch with only one managed object (resource pool) being added.  I think it's fine for now....

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -243,23 +243,21 @@ func (handler *ContainersHandlersImpl) GetContainerInfoHandler(params containers
...
 	// get the container id for interogation
 	containerID := exec.ParseID(params.ID)
-	cc, err := exec.ContainerInfo(context.Background(), handler.handlerCtx.Session, containerID)
-	if err != nil {
-		log.Debugf("GetContainerInfoHandler Error: %s", err.Error())
-		return containers.NewGetContainerInfoNotFound().WithPayload(&models.Error{Message: err.Error()})
+	container := exec.ContainerInfo(context.Background(), containerID)
+	if container == nil {
+		info := fmt.Sprintf("GetContainerInfoHandler ContainerCache miss for container(%s)", containerID.String())
+		log.Debug(info)
```
caglar10ur: log.Error?

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -268,7 +266,8 @@ func (handler *ContainersHandlersImpl) GetContainerListHandler(params containers
 		info.ContainerID = &container.ExecConfig.ID
 		info.LayerID = &container.ExecConfig.LayerID
 		info.Created = &container.ExecConfig.Created
-		info.Status = &container.Status
+		s := container.State.String()
```
caglar10ur: Can you declare the string out of the loop and reuse the same variable instead of creating one for each vm?

lib/portlayer/event/event_manager.go
```diff
@@ -0,0 +1,46 @@
...
+type EventManager interface {
+
+	// add / remove managed objects for listening
+	AddMonitoredObject(ref string) error
+	RemoveMonitoredObject(ref string)
+
+	// register / unregister for callbacks
+	// TODO: hate to muddy the interface w/the session, but currently required
+	Register(caller string, callback func(Event, *session.Session))
+	UnRegister(caller string)
```
caglar10ur: Unregister

lib/portlayer/event/event_manager.go
```diff
@@ -0,0 +1,46 @@
...
+	RegistryCount() int
+
+	// start listening for events
+	Start() error
+	// stop listening for events
+	Stop()
+
+	// blacklist / unblacklist object from event listening
+	Blacklist(ref string) error
+	UnBlacklist(ref string)
```
caglar10ur: Unblacklist

lib/portlayer/event/events.go
```diff
@@ -0,0 +1,39 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package event
+
+type Container string
```
hmahmood: Is this used anywhere?

lib/portlayer/event/vsphere/manager.go
```diff
@@ -0,0 +1,234 @@
...
+
+	"golang.org/x/net/context"
+)
+
+type vSphereManager struct {
+	vmwManager *vmwEvents.Manager
+	callbacks  map[string]func(event.Event, *session.Session)
+	blacklist  map[types.ManagedObjectReference]string
+	objects    map[string]types.ManagedObjectReference
+	sess       *session.Session
```
caglar10ur: let's not save 4 letters and call this session (unless we get a compile error or something). sess sounds like a session slice

lib/portlayer/event/vsphere/manager.go
```diff
@@ -0,0 +1,234 @@
...
+
+func (mgr *vSphereManager) monitoredObjects() []types.ManagedObjectReference {
+	refs := make([]types.ManagedObjectReference, 0, len(mgr.objects))
+	for k := range mgr.objects {
+		refs = append(refs, mgr.objects[k])
+	}
+	return refs
+}
+
+// blacklist object from callbacks
```
caglar10ur: vet will whine (and our default script will ignore) because comments are starting with lower case letters. Can you match with fn names?

lib/portlayer/exec/container_cache.go
```diff
@@ -48,11 +61,35 @@ func NewContainerCache() {
...
 	defer conCache.m.RUnlock()
-	// find by id
+	// find by id or moref
 	container := conCache.cache[idOrRef]
 	return container
 }
 
+// return an array of pointers to the containers in the cache
+// TODO: implement filtering here?
+func (conCache *containerCache) Containers(allStates *bool) []*Container {
```
caglar10ur: Not sure why we need this in cache pkg. AFAIS there is only one caller so this code very well could live in the caller. Also please bool not *bool.

lib/portlayer/exec/container_cache.go
```diff
@@ -79,3 +117,109 @@ func (conCache *containerCache) Remove(idOrRef string) {
 	}
 
 }
+
+func isContainerID(id string) bool {
```
caglar10ur: Why we are checking the IDs and why using regex? We put those ids there in the first place right?

lib/portlayer/exec/container_cache.go
```diff
@@ -79,3 +117,109 @@ func (conCache *containerCache) Remove(idOrRef string) {
...
+	// ID should only be lower case chars & numbers
+	match, _ := regexp.Match("^[a-z0-9]*$", []byte(id))
+	return match
+}
+
+// call back for the event stream
+func callme(ie event.Event, sess *session.Session) {
```
caglar10ur: I suggest calling this callback unless there is another function maybe and you want to do :)

```
defer callme(maybe(...))
```

lib/portlayer/exec/container_cache.go
```diff
@@ -79,3 +117,109 @@ func (conCache *containerCache) Remove(idOrRef string) {
...
+	// grab the container from the cache
+	container := containers.Container(ie.Reference())
+	if container != nil {
+
```
caglar10ur: extra space

lib/portlayer/exec/container_cache.go
```diff
@@ -79,3 +117,109 @@ func (conCache *containerCache) Remove(idOrRef string) {
...
+func isContainerID(id string) bool {
+	// ID should only be lower case chars & numbers
+	match, _ := regexp.Match("^[a-z0-9]*$", []byte(id))
+	return match
+}
+
+// call back for the event stream
+func callme(ie event.Event, sess *session.Session) {
+	// grab the container from the cache
+	container := containers.Container(ie.Reference())
```
caglar10ur: do you need to check whether ie is nil before dereferencing it?

lib/portlayer/exec/container_cache.go
```diff
@@ -79,3 +117,109 @@ func (conCache *containerCache) Remove(idOrRef string) {
...
+				// remove from cache
+				containers.Remove(container.ExecConfig.ID)
+
+			}
+		}
+	} else {
+		// TODO: do we need to worry about this?  any further evaluation, etc?
+		log.Warnf("ContainerCache miss for VM(%s) in VCH Resource Pool: Event(%s)", ie.Reference(), ie.Event())
+	}
+	return
```
caglar10ur: no need for this

lib/portlayer/exec/container_cache.go
```diff
@@ -79,3 +117,109 @@ func (conCache *containerCache) Remove(idOrRef string) {
...
+}
+
+// best effort container udpate -- if there's an issue we'll warn and return
+func containerUpdate(sess *session.Session, container *Container, start string, exitStatus int) {
+	// first update exec from vsphere
+	ctx, cancel := context.WithTimeout(context.TODO(), propertyCollectorTimeout)
+	defer cancel()
```
caglar10ur: Not sure I understand the purpose of this. This will never timeout and no one calls cancel so why we bother to create?

lib/portlayer/exec/container_cache.go
cgtexmex: we could, but the nil check is in the event manager -- the only time we callback is if an event is present

lib/portlayer/event/vsphere/manager.go
cgtexmex: I think the issue is that it's a package name -- I can provide an alias for that package and then call it session -- was just being lazy and following the other areas of the portlayer that do the same thing

lib/portlayer/exec/container_cache.go
cgtexmex: we have a single map[string]*Container that has both IDs and Managed Object References as the key -- this allows us to distinguish between the two

lib/portlayer/exec/container_cache.go
cgtexmex: sure we could put it there -- I placed here so that others could use if desired -- in a near term change the cache will be moved out of exec and accessible by other areas.  So the thought being those other areas might need to filter -- this would provide them a standard way to do that...

cgtexmex: ok - so I THOUGHT because it was the package name that's why the other areas used sess instead of session, but apparently that's not the case.  A package alias wan't required, so I've changed it to session

lib/portlayer/exec/container_cache.go
cgtexmex: Why could it not timeout?

lib/portlayer/event/events.go
cgtexmex: removed

lib/portlayer/event/vsphere/manager.go
cgtexmex: ok - so I THOUGHT because it was the package name that's why the other areas used sess instead of session, but apparently that's not the case. A package alias wan't required, so I've changed it to session

lib/portlayer/exec/container_cache.go
cgtexmex: Can we agree to leave here for now? -- we can revisit when I implement #1832

I did remove the pointer

hmahmood:
lgtm

lib/portlayer/exec/container_cache.go
caglar10ur: Cause I can't read properly :) I didn't see propertyCollectorTimeout

caglar10ur:
LGTM :sheep: :it:

lib/portlayer/exec/container_cache.go
```diff
@@ -27,32 +34,55 @@ import (
 type containerCache struct {
 	m sync.RWMutex
 
-	// cache maps containerVM to container ID && infrastructure ID
+	// cache by container id
```
hickeng: If this is only containerID as a key then should we use `uid.UID` as the type instead of string?

lib/portlayer/event/events.go
```diff
@@ -0,0 +1,39 @@
...
+	ContainerResumed      = "Resumed"
+	ContainerRemoved      = "Removed"
+	ContainerReconfigured = "Reconfigured"
+)
+
+type Event interface {
+	// id of event
+	EventID() int
+	// event (PowerOn, PowerOff, etc)
+	Event() string
```
hickeng: Should this be an enum of the types rather than freeform string?
If freedform string, why not satisfy the Stringer interface and use String()

lib/portlayer/exec/container_cache.go
cgtexmex: The single map has moids or UIDs -- so typically 2 entries for each containerVM -- one keyed off of moid another via UID.

Originally I had this as two maps -- that might come back to fruition in a near term change.

lib/portlayer/event/events.go
cgtexmex: Agreed -- Event should be replaced with the Stringer interface.

hickeng:
LGTM

cgtexmex:
rebase

lib/portlayer/exec/container_cache.go
```diff
@@ -79,3 +109,106 @@ func (conCache *containerCache) Remove(idOrRef string) {
...
+			return StateStopped
+		}
+	case event.ContainerSuspended:
+		// are we in the process of suspending
+		if current != StateSuspending {
+			return StateSuspended
+		}
+	case event.ContainerRemoved:
+		if current != StateRemoving {
+			return StateRemoved
```
mhagen-vmware: Is this essentially asking after an event when the vm is no long in the transition state, assume it is in the final state?  Isn't that inherently quite risky?  If you attempt power on, it is in starting state, but then fails and enters poweredoff state, would we then report it is in the running state?

cgtexmex: This is used to determine who initiated the activity that lead to the event.  If it's initiated by the PL then we trust that the PL will set the final state -- if it's attributed to the PL then we set the state to the event received as it's considered OOB activity.

Issue #1578, opened 2016-07-20
Improve docker run for quick ending containers by delaying attach


sflxn:
When users use the CLI to call attach with individual operations, they usually use the following sequence:
1. docker create
2. docker start
3. docker attach

When users use the CLI to call docker run, the actual sequence occurs
1. docker create
2. docker attach
3. docker start

For quick ending containers, start will end quick but attach will continue to wait till it times out.  If we delay attach for start to provide some results, we can improve the user experience.  We'll need to try and detect docker run.

mdubya66:
@hickeng : please triage for prioritization

sflxn:
Since @hickeng is sick, I'll mentioned he said he wanted this of onsite during our meeting on Monday.  I looked for a pre-existing issue and created this new one when I didn't find one.

hickeng:
I don't think we'll be able to get this in for onsite given we're going with a build tomorrow, but I'll still work on it as time allows.
My approach is very simple for now conceptually:
1. in the personality, check if there's an active attach for the container about to be started. This likely requires the addition of a map to record active attach connections in the personality side.
2. when starting a container check for an active attach session and, if so, modify the session configuration to enable a blocking mode
3. in the tether, launch the session in a deferred Go routine that blocks on a channel. That channel will be closed either by establishing an attach connection or via timeout

From a concrete perspective my initial implementation will be:
1. modify tether so that the attach server is started before sessions are launched rather than after.
2. launch the session in a deferred fashion if attach is configured for the session

This approach is because the problem interacts with the on-demand attach behaviour we need to enable vMotion.  Attach enabled/disabled is already present in the tether - what we need to add is to determine if the serial connection is enabled. There is a toolbox mechanism for doing this connect/disconnect within the guest (@dougm) however we can infer this from whether the base ssh connection can be established for now.

hickeng:
Let's do this:

attach.Join(...) - configures the VM for attach at all (e.g. add serial port configured as network client) 
attach.Bind(...) - configures the VM for interaction (configures the serial port for connect)

This is exactly the model that the other port layer components use and also allows us to perform the on-demand attach via a call to attach.Bind and disconnect via a call to attach.Unbind.

hickeng:
@caglar10ur

hickeng:
Ran through the logic with @caglar10ur. This requires:
1. move extension initialization ahead of session launch (lib/tether)
2. modify the session reader/writers to be plain Writer interface (lib/tether)
[...]

PR #1918, opened 2016-08-10
[WIP] On-demand serial port connection [full ci]

### LINKED ISSUE #858: On-demand serial port connection
To support vMotion while:
* not requiring the applianceVM be running
* not requiring a native agent on the ESX host

the only real option we have available is to toggle the connected state of the network serial port.

This will require code in the following components:
* portlayer/interaction - toggle the port on when an attach request comes in
* tether - toggle disconnection of the port if no connection established within timeout (allows for vMotion of VM in the case of appliance failure during attach flow).

To keep attach responsive under normal circumstances we should try having every VM connected as they start, dropping that connection after a couple of minutes. The primary drawback of this will be if starting large numbers of VMs that should immediately be vMotioned - should make this configuration, but with programmatic coupling with rate controls for VM start in VMfork scenarios (where single host start point is more likely),

caglar10ur:
On-demand serial port connection

    - Tether always starts SSH server
    - spec/serial.go no longer exists
    - ContainerCreate calls AddInteractionToContainer to add Serial ports which
    calls interaction.InteractionJoin
    - ContainerCreate calls AddLoggingToContainer to add Serial ports which
    calls logging.LoggingJoin
    - VMs start with disconnected ports if 'run -d ...' used
    - ContainerAttach toggles the connected bit via interaction.Bind
    - We tear down the serial connection when we detach from the container.
    This logic assumes that we let only 1 client to attach. Need to revisit
    this and implement a reference counter.

    - Commit now uses ChangeVersion to guard against updates happened
    between when vm's config is read and when it is applied. To support that
    we call Refresh in every time we modify the state. So each state change
    now may contain one or more property collector call.

    - Cosmetic changes all over
fixes #858


caglar10ur:
@hickeng can you take a quick look as a 1st pass?

lib/apiservers/engine/backends/container.go
```diff
@@ -117,6 +119,38 @@ const (
 	DefaultEnvPath         = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 )
 
+func NotFoundError(msg string) error {
+	return derr.NewRequestNotFoundError(fmt.Errorf("No such container: %s", msg))
+}
+
+func UnknownError(msg string) error {
```
hickeng: Why do we have both unknownerror and server error? They seem to serve much the same role

lib/apiservers/engine/backends/container.go
```diff
@@ -333,17 +370,15 @@ func (c *Container) ContainerResize(name string, height, width int) error {
...
 	}
+	name = vc.ContainerID
 
 	// Get an API client to the portlayer
-	client := PortLayerClient()
-	if client == nil {
-		return derr.NewErrorWithStatusCode(fmt.Errorf("container.ContainerResize failed to create a portlayer client"),
-			http.StatusInternalServerError)
-	}
+	client := c.containerProxy.Client()
```
hickeng: thank you!

lib/apiservers/engine/backends/container.go
```diff
@@ -478,50 +503,35 @@ func (c *Container) containerStart(name string, hostConfig *container.HostConfig
 	}
 
 	// get a handle to the container
-	var getRes *containers.GetOK
-	getRes, err = client.Containers.Get(containers.NewGetParamsWithContext(ctx).WithID(name))
+	handle, err := c.Handle(name, paramName)
```
hickeng: The model for the handles is that there is one per operation and the handle is only valid for that one operation.
This allows separation of the port layer components in the future without major changes to the callers.

lib/apiservers/engine/backends/container.go
caglar10ur: The original code was using those, the only difference is the string that we are returning. Server error from vs Unknown error from. I could remove Unknown error and use ServerError everywhere if Unknown error messages are not required for some reason

lib/apiservers/engine/backends/container.go
```diff
@@ -1014,6 +997,30 @@ func (c *Container) ContainerAttach(name string, ca *backend.ContainerAttachConf
 		}
 	}
 
+	client := c.containerProxy.Client()
+	handle, err := c.Handle(name, paramName)
```
hickeng: same comment as previously regarding acquisition of handle

lib/apiservers/engine/backends/container.go
```diff
@@ -1014,6 +997,30 @@ func (c *Container) ContainerAttach(name string, ca *backend.ContainerAttachConf
...
+	if err != nil {
+		return err
+	}
+
+	_, err = client.Interaction.Bind(interaction.NewBindParams().WithID(handle))
+	if err != nil {
+		return ServerError(err.Error())
+	}
+
+	// commit the handle; this will reconfigure and start the vm
```
hickeng: pretty sure this comment is a copy/paste error

lib/apiservers/engine/backends/container.go
caglar10ur: Not sure I understand your comment. We are still doing what the previous code is doing. I just created a function https://github.com/vmware/vic/pull/1918/files#diff-6e989557b980355774bd4187364245fbR138 to make it DRY

lib/apiservers/engine/backends/portlayer/container_proxy.go
```diff
@@ -259,8 +269,29 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
 	return handle, nil
 }
 
-// CommitContainerHandle() commits any changes to container handle.
+// AddPortsToContainer adds ports to a container, referenced by handle.
+// If an error is return, the returned handle should not be used.
 //
+// returns:
+//	modified handle
+func (c *ContainerProxy) AddPortsToContainer(handle string, config types.ContainerCreateConfig) (string, error) {
```
hickeng: What's the logic on the naming here? I had read this to now thinking it was adding exposed ports for some refactoring reason. Only now is it evident it's adding the interaction mechanism - this name is going to make less sense when it's a VMCI connection and not a serial port.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -78,6 +84,72 @@ func (i *InteractionHandlersImpl) Configure(api *operations.PortLayerAPI, _ *Han
 	}
 }
 
+func (i *InteractionHandlersImpl) JoinHandler(params interaction.JoinParams) middleware.Responder {
+	defer trace.End(trace.Begin(params.ID))
+
+	handle := exec.GetHandle(params.ID)
```
hickeng: this should be params.Config.Handle (see ScopesAddContainer, but ignore the naming) - think of the case when the port layer components are not co-located... the handle in that scenario is the full textual representation of the config spec, which this call will modify.

lib/apiservers/portlayer/swagger.json
```diff
@@ -1411,6 +1411,137 @@
 				}
 			}
 		},
+		"/interaction/{id}/serial": {
+			"post": {
+				"description": "Add serial port",
```
hickeng: Let's avoid leaking the implementation detail (serial port) in the description. This call should remain viable when we switch to VMCI based backchannel.

lib/apiservers/engine/backends/container.go
hickeng: You're right - strike this comment as a mistake.

lib/apiservers/engine/backends/container.go
caglar10ur: Yep, fixing it

lib/apiservers/engine/backends/portlayer/container_proxy.go
caglar10ur: I just followed the other naming convention in that file. I could change it, what should we use?

lib/apiservers/portlayer/swagger.json
caglar10ur: I thought we could use "/interaction/{id}/vmci" when the time comes

lib/apiservers/engine/backends/container.go
caglar10ur: @sflxn can you comment whether we need UnknownError for some reason?

lib/apiservers/portlayer/swagger.json
hickeng: While this is an internal API for now it is supposed to be exposed at some point - I don't think this is a level of detail we should be exposing. The caller of this API should know little to nothing about the infrastructure so how are they supposed to determine which are available or desired.
If we have different end points for interaction it should be because they provide different user facing semantics.

lib/portlayer/attach/handle.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	config := &types.VirtualDeviceConfigSpec{
+		Device:    serial,
+		Operation: types.VirtualDeviceConfigSpecOperationAdd,
+	}
+	s.handle.Spec.DeviceChange = append(s.handle.Spec.DeviceChange, config)
+
+	VMPathName := s.handle.Spec.VMPathName()
+	VMName := s.handle.Spec.Spec().Name
+	for _, suffix := range []string{"debug", "log"} {
```
hickeng: I'm not comfortable having the logging tied to the attach mechanism. We should chat in person about why these are rolled in here.

lib/portlayer/attach/handle.go
```diff
@@ -0,0 +1,179 @@
...
+	return nil
+}
+
+// Join adds 1 network backed, 2 file backed serial port to the caller and configures them
+func (s *SerialHandler) Join() error {
+	defer trace.End(trace.Begin(""))
+
+	// FIXME: THERE MUST BE ANOTHER WAY
+	// following is from Create@exec.go
+	ips, err := net.LookupIP(exec.ManagementHostName)
```
hickeng: We need to have the serial port configured during Join, however we may need the IP configured during Bind. What occurs after appliance restart and/or if the DHCP address changes?

emlin:
One comment for your comments: 😄 

> ContainerCreate calls AddPortsToContainer to add Serial ports which calls interaction.Bind

which calls interaction.Join

emlin:
Another question is you had Unbind method, but not called from docker engine. Should that be invoked after attach?

caglar10ur:
@hickeng I would love to get a new review

lib/apiservers/engine/backends/portlayer/container_proxy.go
```diff
@@ -427,6 +474,10 @@ func dockerContainerCreateParamsToPortlayer(cc types.ContainerCreateConfig, laye
 	config.WorkingDir = new(string)
 	*config.WorkingDir = cc.Config.WorkingDir
 
+	// attach
+	config.Attach = new(bool)
+	*config.Attach = cc.Config.AttachStdin || cc.Config.AttachStdout || cc.Config.AttachStderr
```
caglar10ur: Compare it with vanilla docker (docker create/start/attach case)

lib/portlayer/logging/logging.go
```diff
@@ -0,0 +1,89 @@
...
+import (
+	"fmt"
+
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/lib/portlayer/exec"
+	"github.com/vmware/vic/pkg/trace"
+)
+
+// Handler defines the interface
+type Handler interface {
```
caglar10ur: Convert them to return handle

cmd/tether/attach.go
```diff
@@ -87,6 +75,7 @@ func (t *attachServerSSH) Stop() error {
 	defer trace.End(trace.Begin("stop attach server"))
 	// calling server.start not t.start so that test impl gets invoked
 	server.stop()
+	t.running = false
```
hickeng: server.stop() sets running to false

cmd/tether/attach.go
```diff
@@ -51,34 +51,22 @@ type attachServerSSH struct {
 	sshConfig *ssh.ServerConfig
 
 	enabled bool
+
+	running bool
```
hickeng: I cannot see why we have running as well as enabled - as best I can tell they are both used to make start & stop idempotent and are effectively identical.

lib/apiservers/engine/backends/container.go
```diff
@@ -549,9 +559,7 @@ func (c *Container) containerStart(name string, hostConfig *container.HostConfig
 	if bind {
 		e := c.findPortBoundNetworkEndpoint(hostConfig, endpoints)
 		if err = c.mapPorts(portmap.Map, hostConfig, e); err != nil {
-			err = fmt.Errorf("error mapping ports: %s", err)
-			log.Error(err)
-			return InternalServerError(err.Error())
+			return InternalServerError(fmt.Sprintf("error mapping ports: %s", err))
```
hickeng: Why no logging here - even if mapPorts logs the initial error I don't think it has the requisite knowledge to record which container had the problem.
Admittedly the original code didn't add that useful bit of data either, and the context based logging @fdawg4l has in WIP would address it in a cleaner fashion.

lib/apiservers/engine/backends/container.go
```diff
@@ -988,6 +996,33 @@ func (c *Container) ContainerAttach(name string, ca *backend.ContainerAttachConf
 		}
 	}
 
+	client := c.containerProxy.Client()
+	handle, err := c.Handle(name, paramName)
+	if err != nil {
+		return err
+	}
+
+	_, err = client.Interaction.InteractionBind(interaction.NewInteractionBindParamsWithContext(ctx).
```
hickeng: We should be getting the returned handle here and using that in the call to Commit

lib/apiservers/engine/backends/portlayer/container_proxy.go
```diff
@@ -68,17 +70,21 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/sys"
 )
 
+// VicContainerProxy interface
 type VicContainerProxy interface {
 	CreateContainerHandle(imageID string, config types.ContainerCreateConfig) (string, string, error)
```
hickeng: For my own sanity - is this Get to get an existing container or Create to create a new container?

lib/apiservers/engine/backends/portlayer/container_proxy.go
```diff
@@ -265,14 +273,58 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
...
+// AddLoggingToContainer adds logging capability to a container, referenced by handle.
+// If an error is return, the returned handle should not be used.
+//
+// returns:
+//	modified handle
+func (c *ContainerProxy) AddLoggingToContainer(handle string, config types.ContainerCreateConfig) (string, error) {
+	defer trace.End(trace.Begin(handle))
+
+	if c.client == nil {
+		return "", InternalServerError("ContainerProxy.AddLoggingToContainer failed to create a portlayer client")
```
hickeng: Pedantic - this isn't creating a portlayer client, but instead attempting to use a pre-created one.

For future work - we shouldn't even get to these calls if there's no port layer client created.

lib/apiservers/engine/backends/portlayer/container_proxy.go
```diff
@@ -265,14 +273,58 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
...
+
+	_, err := c.client.Logging.LoggingJoin(logging.NewLoggingJoinParamsWithContext(ctx).
+		WithConfig(&models.LoggingJoinConfig{
+			Handle: handle,
+		}))
+	if err != nil {
+		return "", InternalServerError(err.Error())
+	}
+
+	return handle, nil
```
hickeng: Should be returning the handle returned by the LoggingJoin call.

lib/apiservers/engine/backends/portlayer/container_proxy.go
```diff
@@ -265,14 +273,58 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
...
+
+	_, err := c.client.Interaction.InteractionJoin(interaction.NewInteractionJoinParamsWithContext(ctx).
+		WithConfig(&models.InteractionJoinConfig{
+			Handle: handle,
+		}))
+	if err != nil {
+		return "", InternalServerError(err.Error())
+	}
+
+	return handle, nil
```
hickeng: Should be returning the handle returned by the InteractionJoin call.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -99,9 +98,8 @@ func (handler *ContainersHandlersImpl) CreateHandler(params containers.CreatePar
 					ID:   id,
 					Name: *params.CreateConfig.Name,
 				},
-				Tty: *params.CreateConfig.Tty,
-				// FIXME: default to true for now until we can have a more sophisticated approach
-				Attach: true,
+				Tty:    *params.CreateConfig.Tty,
+				Attach: *params.CreateConfig.Attach,
```
hickeng: I think this may need to be updated for #1578

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -71,13 +79,99 @@ func (i *InteractionHandlersImpl) Configure(api *operations.PortLayerAPI, _ *Han
...
+func (i *InteractionHandlersImpl) JoinHandler(params interaction.InteractionJoinParams) middleware.Responder {
+	h, ok := params.Config.Handle.(string)
+	if !ok {
+		return interaction.NewInteractionJoinInternalServerError().WithPayload(
+			&models.Error{Message: "Handle type assertion failed"},
+		)
+	}
+	defer trace.End(trace.Begin(h))
+
+	handle := exec.GetHandle(h)
```
hickeng: I hate this. This has to change (not for this PR)

lib/portlayer/attach/attach.go
```diff
@@ -0,0 +1,148 @@
...
+)
+
+// Handler defines the interface
+type Handler interface {
+	Join(handle interface{}) (interface{}, error)
+	Bind(handle interface{}) (interface{}, error)
+	Unbind(handle interface{}) (interface{}, error)
+}
+
+func toggle(h interface{}, connected bool) (interface{}, error) {
```
hickeng: This is not a part of the Port Layer API so could actually specify the type of the handle we require.

lib/apiservers/engine/backends/container.go
caglar10ur: the caller is logging the error but I can add the logging back here if you want. It's just this is the only log call in the entire function :)

lib/apiservers/engine/backends/portlayer/container_proxy.go
caglar10ur:
```
// CreateContainerHandle creates a new VIC container by calling the portlayer
```

what it says and it returns id, handle (as string) and error

lib/portlayer/attach/attach.go
```diff
@@ -0,0 +1,151 @@
...
+
+		if serial.GetVirtualDevice().Connectable.Connected == connected {
+			log.Debugf("Already in the desired state (connected: %t)", connected)
+			return handle, nil
+		}
+
+		log.Debugf("Setting the connected state to %t", connected)
+		// set the values
+		serial.GetVirtualDevice().Connectable.Connected = connected
+		if connected && handle.ExecConfig.Sessions[handle.ExecConfig.ID].Attach {
```
hickeng: This will almost always start the container connected. The only time that's not the case is with `run -d`.
This should be addressed in the follow up PR we're been discussing that adds run-blocking.

lib/portlayer/attach/attach.go
```diff
@@ -0,0 +1,151 @@
...
+	handle, ok := h.(*exec.Handle)
+	if !ok {
+		return nil, fmt.Errorf("type assertion failed")
+	}
+	// make sure a spec exists
+	handle.SetSpec(nil)
+
+	defer trace.End(trace.Begin(""))
+
+	// FIXME: THERE MUST BE ANOTHER WAY
```
hickeng: What is your objection here?

lib/portlayer/attach/attach.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, fmt.Errorf("Multiple IPs found on %s: %#v", constants.ManagementHostName, ips)
+	}
+
+	// Tether serial port - backed by network
+	serial := &types.VirtualSerialPort{
+		VirtualDevice: types.VirtualDevice{
+			Backing: &types.VirtualSerialPortURIBackingInfo{
+				VirtualDeviceURIBackingInfo: types.VirtualDeviceURIBackingInfo{
+					Direction:  string(types.VirtualDeviceURIBackingOptionDirectionClient),
+					ServiceURI: fmt.Sprintf("tcp://%s:%d", ips[0], constants.SerialOverLANPort),
```
hickeng: We could, and probably should, do this in bind.

caglar10ur:
Pushed the new version as it was sitting on my local disk but this is still WIP (missing reconnect logic) so no need to review it at this point.

caglar10ur:
closing this as I started to split this into multiple PRs

PR #1920, opened 2016-08-11
Handle container vm stop/power off race

dougm:
In container stop() it is possible for the container vm to have reached
powered off state in between graceful shutdown and hard power off.
Ignore the PowerOff_Task error in this case.

### DIFF
--- a/lib/portlayer/exec/container.go
+++ b/lib/portlayer/exec/container.go
@@ -23,6 +23,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/vmware/govmomi/guest"
 	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/task"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/soap"
 	"github.com/vmware/govmomi/vim25/types"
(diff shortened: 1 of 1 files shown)


dougm:
Looks like the cause of this test failure: https://ci.vmware.run/vmware/vic/3842

mhagen-vmware:
lgtm

mhagen-vmware:
Are we elsewhere checking first before we attempt the power off in the first place? If not, it would be nice to do that to avoid forcing vsphere constantly down error paths.

dougm:
@mhagen-vmware this is the only place we stop/power off a container vm.  We do the following here:

* send stop signal to the container process (SIGTERM by default)

* wait for vm power state == off or timeout (10 second timeout by default)

* if power state is not off, send SIGKILL to the container process

* wait for vm power state == off or timeout (10 second timeout by default)

* if power state is still not off, PowerOff the container vm

Normally we'd be done after the first SIGTERM + a short wait for power state off and not hit the remaining steps.  The test that failed intentionally forces us to the SIGKILL step.  Combined with a shorter timeout (2sec) and a busy system, it is possible the vm power state won't be set to "off" until after we've entered the call to PowerOff().

cgtexmex:
LGTM

PR #1932, opened 2016-08-11
Fix Util.robot to reflect change in vic-machine command-line options

kjplatz:
vic-machine was recently changed to use --image-store as opposed to --image-datastore, however both cases of "Install VCH Appliance to Test Server" were not updated to reflect this change.  This PR fixes the ${cert}=true case.


hmahmood:
lgtm

PR #1934, opened 2016-08-11
[WIP] [NEEDSLOVE] StartedAt/FinishedAt for docker personality

### LINKED ISSUE #1899: Empty start/finish date in inspected container
**VIC Version:**

0.4 / build #3411

**Code**
I have running container, but it's started/finished time values are empty :
```
root@docker ~ $ docker -H vic:2375 ps -a
CONTAINER ID    IMAGE                           COMMAND             CREATED          STATUS     NAMES
4710627dd31d    kitematic/hello-world-nginx     "sh /start.sh"      15 hours ago     Running    hello-world-nginx
root@docker ~ $ docker -H vic:2375 inspect 4710627dd31d
[
...
        "State": {
            "Status": "running",
            "Running": true,
            "Paused": false,
            "Restarting": false,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 0,
            "ExitCode": 0,
            "Error": "",
            "StartedAt": "",
            "FinishedAt": ""
        },
...
```
same docker inspect executed in docker env have the correct start time and 'zero'-ed finish time (the container is still running)
```
root@docker ~ $ docker inspect 3fcd
[
...
        "State": {
            "Status": "running",
            "Running": true,
            "Paused": false,
            "Restarting": false,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 9012,
            "ExitCode": 0,
            "Error": "",
            "StartedAt": "2016-08-09T06:45:38.881606877Z",
            "FinishedAt": "0001-01-01T00:00:00Z"
        },
...
```
Noted that Pid is not populated too, but maybe this is fine since there's no real process on the host itself.

Similarly to #1874 that breaks clients that expect to have correct time set in the fields

caglar10ur:
fixes #1899

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -577,6 +577,9 @@ func (c *Container) containerStart(name string, hostConfig *container.HostConfig
 		}
 	}
 
+	// set StartedAt time
+	vc.StartedAt = time.Now().UTC()
+
 	return nil
 }
 
@@ -807,6 +810,9 @@ func (c *Container) containerStop(name string, seconds int, unbound bool) error 
 		}
 
 	}
+
+	// set FinishedAt time
+	vc.FinishedAt = time.Now().UTC()
 
 	return nil
 }
--- a/lib/apiservers/engine/backends/container/viccontainer.go
+++ b/lib/apiservers/engine/backends/container/viccontainer.go
@@ -15,6 +15,8 @@
 package container
 
 import (
+	"time"
+
 	containertypes "github.com/docker/engine-api/types/container"
 )
 
@@ -25,6 +27,10 @@ type VicContainer struct {
 	ContainerID string
 	Config      *containertypes.Config //Working copy of config (with overrides from container create)
 	HostConfig  *containertypes.HostConfig
+
+	// non-persistent last start and finish time
+	StartedAt  time.Time
+	FinishedAt time.Time
 }
 
 // NewVicContainer returns a reference to a new VicContainer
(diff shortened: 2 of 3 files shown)


cgtexmex:
Why are we not really fixing the issue by adding start / finished in the ContainerVM?  I think no time is better than wrong times....am I missing something?

hmahmood:
Will this work if the container stops out of band?

caglar10ur:
@cgtexmex cause we really don't need them in the portlayer, and we can get those via events. This is - I think - docker specific data

caglar10ur:
@hmahmood it won't and to be honest not sure how to handle that. We can ask esxi to provide those to us - if there is a way to get them without asking all the events

cgtexmex:
@caglar10ur I think process start / stop times would be useful for other things as well.  From my perspective the PL is the source of truth for ContainerVM lifecycle events...so to exclude this data is a miss IMO.

cgtexmex:
FWIW it would be easiest to get these start / stop times if they are persisted in the VMX like we do created.  The event stream on standalone ESX is not persistent and to rebuild that data via the VC event stream would be challenging...not impossible, but just more work than benefit

hmahmood:
I am not sure putting the times in the vmx is required here. Can we just add this info to what is returned here: https://github.com/vmware/vic/blob/master/lib/apiservers/engine/backends/container.go#L864

cgtexmex:
@hmahmood sure..but if we don't persist it somewhere then it won't survive restarts, etc...again we can probably pull from the event history, but that doesn't scale particularly well

hmahmood:
@cgtexmex yes that makes sense.

caglar10ur:
OK, will look into persisting the values in vmx file

hickeng:
This ties into a change to the process status reporting when factoring in any out-of-band or hard power off operations. Currently we have no way of determining if a process exited with status 0 vs exit status not being recorded.

It's just about viable to have Started time be stored in the VMX as the tether can enter that on launch, but we then have to deal with the scenario of a powered off VM that _doesn't_ have the stopped time. We can infer that the container was "killed" (i.e. power off) in that circumstance if we can also tell that the reported state for the process is still "running".

cgtexmex:
closing this as a new PR will be submitted

Issue #1924, opened 2016-08-11
Failure in CI test Stop a container with specific grace period


mhagen-vmware:
This test has now failed 3 times in a row with 3 different errors:
3846: 125 != 0
[integration_logs_3846_936a10a4e99665dd7521fe7fbd8d6482296e0621_1470925723.zip](https://github.com/vmware/vic/files/413758/integration_logs_3846_936a10a4e99665dd7521fe7fbd8d6482296e0621_1470925723.zip)

3856: 1 != 0
[integration_logs_3856_cb88305b25486548b7829c822926f4c61b63f808_1470934272.zip](https://github.com/vmware/vic/files/413760/integration_logs_3856_cb88305b25486548b7829c822926f4c61b63f808_1470934272.zip)

3860: '' does not contain 'StopSignalHUP'
[integration_logs_3860_ef590de88f0aaa0a03e439e214e3b8476e87c63a_1470935256.zip](https://github.com/vmware/vic/files/413784/integration_logs_3860_ef590de88f0aaa0a03e439e214e3b8476e87c63a_1470935256.zip)

I am going to quarantine this test now, until this issue is resolved.

dougm:
Seen with the 10s grace period too:

```
Gathering logs from the test server...
Deleting the VCH appliance...
Test-Cases.Group1-Docker-Commands.1-14-Docker-Kill :: Test 1-14 - ... | FAIL |
3 critical tests, 2 passed, 1 failed
3 tests total, 2 passed, 1 failed

Signal a container with default kill signal                           | PASS |
------------------------------------------------------------------------------
Signal a container with SIGHUP                                        | FAIL |
Keyword 'Assert Kill Signal' failed after retrying 5 times. The last error was: '' does not contain 'KillSignalHUP'
```

https://ci.vmware.run/vmware/vic/3916

dougm:
Looks like the test may just need to wait a bit.. possible we are sending the stop signal before the signal handler (shell trap) is registered within the container process.  In which case, it will exit on the stop signal, without having trapped/echo'd to the ${container}.log file.

PR #1940, opened 2016-08-12
Avoid race in docker stop test

dougm:
It was possible for the test to call 'docker stop' before the container
process had reached the point of trapping the given signal.  Resulting
in the test failure: '' does not contain 'StopSignalHUP'

Issue #1924


mhagen-vmware:
lgtm

dougm:
lgtm

Issue #1778, opened 2016-08-02
No such image error on successful docker pull


fdawg4l:
Reported by @mreferre.  We suspect it's the datastore name, datastore path, or both being passed to `vic-machine create`

```
docker --tls run -it busybox
Unable to find image 'busybox:latest' locally
latest: Pulling from library/busybox
a3ed95caeb02: Pull complete 
8ddc19f16526: Pull complete 
Digest: sha256:65ce39ce3eb0997074a460adfb568d0b9f0f6a4392d97b6035630c9d7bf92402
Status: Downloaded newer image for library/busybox:latest
docker: Error: No such image: busybox.
```

`/vic-machine create --target https://'administrator@vsphere.local:Vmware123!'@192.168.8.211 -i  sharedstorage --name VCH1 --external-network External --bridge-network BridgeVCH1”`

fdawg4l:
cc @gigawhitlocks 

Ian, I think this issue goes away with your change (#1678).  I'm going to assign this to you to try it with your change and close it if the issue goes away.  If not, feel free to debug or bounce it back to me to take a look at.

mlh78750:
I'm seeing same for create command on VC 6.5 beta 3:

`Bauxite:~/vicbin/vic-3538 $ docker -H 10.160.19.62:2376 --tls create busybox
Unable to find image 'busybox:latest' locally
latest: Pulling from library/busybox
a3ed95caeb02: Pull complete
8ddc19f16526: Pull complete
Digest: sha256:65ce39ce3eb0997074a460adfb568d0b9f0f6a4392d97b6035630c9d7bf92402
Status: Downloaded newer image for library/busybox:latest
Error: No such image: busybox`

Install command was: 
`Bauxite:~/vicbin/vic-3538 $ ./vic-machine-darwin create --target 10.160.64.44  --image-datastore datastore1 --name vic-demo --user Administrator@vsphere.local --password 'Admin!23'  --compute-resource /ha-datacenter/host/10.160.6.233/Resources --bridge-network bridge --external-network vm-network`

mlh78750:
Also, after you do this on VC 6.5 beta 3, you cannot delete the files in the datastore over the API. I have to login to the esx and rm -rf. Something is getting messed up on the host side. I can provide the logins/ip's for my setup on nimbus if someone wants to look at this.

mlh78750:
After the failure a vic-machine delete has the following error:
[...]

PR #1949, opened 2016-08-12
Add protection against regression #1778

gigawhitlocks:
Small change; expected it to take more but this should address the final bit in #1778 which is to add a regression test to protect against that issue arising again; namely, where the image pull appears to succeed but finally states `no such image`

Instead of writing a new test, I've modified the `pull` integration tests to ensure that this message is not output at the end of a pull


anchal-agrawal:
LGTM

fdawg4l:
LGTM

gigawhitlocks:
poke

PR #1955, opened 2016-08-12
[skip ci]: draft upgrade related ideas

### LINKED ISSUE #2167: upgrade design
Design first support of vic-machine upgrade

emlin:
@hickeng @karthik-narayan
This is my understanding for current upgrade story. And a few implementation related details. Create a PR for it for easy to get comments.
Fixes #2167


doc/design/upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+
+For any running VCH, VIC admin can downgrade existing VCH to an old version. (I assume this function is only used for upgrade roll back, we might not want to recommend downgrade for anything working well, then the downgrade scope will be same to upgrade versions)
+
+Verified docker-engine upgrade/downgrade, which works well between 1.11.2 and 1.12. The running container will be stopped after upgrade/downgrade, but is good to start again in new version. And all images and containers information is not lost. So from technical point of view, we should support same story.
+
+## Implementation/Issues
+
+### Versioning
+
+#### Forzen must to have attributes in VCH configuration and container VM configuration
```
andrewtchin: s/Forzen/Version ?

doc/design/upgrade.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Transactional Operation/Roll Back
+
+VCH old status and configuration need to be remembered, in case any upgrade failure happens. The information should be in VM guestinfo to make sure if any vSphere connection related issue happens, vic-machine is still able to fetch out enough information for roll back.
+
+Based on this idea, we don't need configuration data downgrade utility here.
+
+# Plan
+
+## Pre-requisition
```
andrewtchin: Prerequisites

andrewtchin:
lgtm

doc/design/upgrade.md
emlin: hm.. typo. 
I meant to say 'Freeze' the attributes definition

andrewtchin:
I just thought of this, but maybe it's worth mentioning that some users will require to be able to upgrade without an internet connection

emlin:
> I just thought of this, but maybe it's worth mentioning that some users will require to be able to upgrade without an internet connection

You're right. Need to mention that.

emlin:
@hickeng Updated document as our discussion last Friday. Please help to review it.
thanks

doc/design/upgrade.md
```diff
@@ -0,0 +1,118 @@
...
+
+vic-machine upgrade will rely on build id change only. User need to provide newer binary with bigger build id, to upgrade existing VCH.
+
+We could also use github commit hash for version diff, but that will be difficult to check if one version is newer than another one, so at this time, we'll leverage build id only. This requires our build system to keep increase build id no matter any system change. 
+
+vic-machine upgrade cannot upgrade VCH to a version newer than itself.
+
+### Impact
+
+The VCH will have downtime during upgrade, suppose container lifecyle is not closely coupled with VCH up time, for VIC has good network isolation. User could run appliacation and connect to their container without VCH is running. But there still have some impact, including docker attach, docker log and docker port mapping, which can be resumed after upgrade is finished.
```
hickeng: The VCH _appliance_ (control plane) will have downtime during the upgrade. Container lifecycles are not coupled with that of the appliance and, when using mapped vsphere networks (vic-machine _--container-network_ argument) instead of port forwarding, the network data paths are not dependent on the appliance either.
There will be impact on container interaction while the appliance is down:
* no direct access to container logs
* no attach ability
* network outage for NAT based port forwarding
All of those facets should resume normal operation after upgrade is complete

doc/design/upgrade.md
```diff
@@ -0,0 +1,118 @@
...
+
+In the future, will consider to add live migration, so user does not experience service broken during upgrade. Integration docker log with syslogd is another option to reduce impact of VCH downtime.
+
+### VCH status
+
+vic-machine ls and inspect should be able to show the VCH versions and upgrade status.
```
hickeng: Does vic-machine inspect need to be able to show the bootstrap versions in use? I'd say yes, but only as a stretch goal.

doc/design/upgrade.md
```diff
@@ -0,0 +1,118 @@
...
+
+vic-machine need to identify the iso file version before and after upgrade. Two options here:
+- Add version into iso file Primary Volume Descriptor, and read back from vic-machine
+- Leverage iso file name including version, e.g. appliance-0.5.0.iso
+
+vic-machine should check iso file version during deployment, and after upload to datastore, version should be appended to iso file name, to make sure mutilple iso files could co-exist in the same VCH, and used for different version's container.
+
+Image file name should not be changed in datastore, cause the file path is used to create container VM. And then vic-machine will leverage file name for version checking during upgrade, to avoid download iso file from datastore back to where vic-machine is running.
+
+Note: No feasible golang library found for this function, so will write our own library to read iso metadata.
```
dougm: https://github.com/hooklift/iso9660

doc/design/upgrade.md
```diff
@@ -0,0 +1,118 @@
...
+
+In the future, will consider to add live migration, so user does not experience service broken during upgrade. Integration docker log with syslogd is another option to reduce impact of VCH downtime.
+
+### VCH status
+
+vic-machine ls and inspect should be able to show the VCH versions and upgrade status.
+
+### Roll Back
+
+For anything wrong happens during upgrade, vic-machine will rollback any changes made to have VCH return to original version and status. If VC/vSphere connection is broken during upgrade, the upgrade or rollback will fail, but user should be able to rollback previously broken upgrade, to recover VCH.
```
hickeng: Roll back should be implemented via snapshot mechanisms as that will preserve the VM hardware state and VMX configuration without us needing to add additional capabilities.

doc/design/upgrade.md
```diff
@@ -0,0 +1,118 @@
...
+- Snapshot VCH appliance VM. - Should quiesce VCH during upgrade here (how?)
+- Poweroff VCH appliance VM if it's not
+- Upload iso files
+- Generate new VCH configuration from old one
+- Update VCH configuration version
+- Update container bootstrap iso file path
+- Update container VM configuration for anything required to support VCH restart(e.g. vch ip change, vch serial port number change, etc.)
+- Attach new appliance iso to VCH appliance VM
+- Power on VCH, and wait VCH initialization
+- Anything wrong in the above steps, switch appliance iso file back to old one, and then roll back to upgrade snapshot
```
hickeng: I think the following ordering will reduce our downtime:
 +- Find existing VCH appliance
 +- Query VCH configuration
 +- Verify new version and existing VCH's version
 +- Generate new VCH configuration from old one
 +- Valiate VCH configuration with new version's validator. If anything incompatible found, stop the upgrade. As no data migration yet.
 +- Upload iso files
 +- Prepare combined applianceVM configuration, both hardware spec (e.g. ISO path) and the extraconfig portion (e.g. VCH version)

--- everything until this point can be done without interrupting ongoing operation ---

 +- Snapshot VCH appliance VM - I don't think we need to quiesce it, we _could_ snapshot memory but that's only useful in the very short term for failure case, not rollback
 +- Poweroff VCH appliance VM
 +- Update container VM configuration with new config
 +- Power on VCH, and wait VCH initialization
 +- Anything wrong in the above steps, roll back to upgrade snapshot (this includes the ISO path so no need to explicitly revert that).
 +- Ensure appliance is powered on after rollback to snapshot.

doc/design/upgrade.md
emlin: The first goal is to show the upgrade is running or not, based on if upgrade snapshot is created.
Can do more in the future.

doc/design/upgrade.md
emlin: The iso9660 is to extract iso files, not read iso file itself. 
The iso9660 standard defines where to put some metadata, and we've used the publisher field for our information. So I reused that for vic version information. But no go library found to read the field out.
But it's easy to do that ourselves.

doc/design/upgrade.md
```diff
@@ -0,0 +1,118 @@
...
+To make data migration easier, we'll need to refactor current VCH and container VM configuration structure.
+
+VirtualContainerHostConfigSpec is too big, which includes everything in one structure. We're not be able to update part of configuration structure, it will help to update part of data with separated structures based on functionality, e.g. network, storage, and common attributes.
+
+For ContainerVM configuration, there are a few attributes not used for container VM setup, e.g. in ExecutorConfig, Key, LayerID, RepoName, are all not container VM configuration related. Move out irrelated attributes can help the structure stability in the future. 
+
+### Transactional Operation/Roll Back
+VM snapshot is one good option for upgrade roll back. Before upgrade, vic-machine will create one snapshot of VCH. If anything wrong happens, vic-machine can switch back to the pre-upgrade snapshot.
+
+As snapshot does not change iso files, vic-machine still need to switch the appliance iso files back to old one.
```
hickeng: No need to switch ISOs as the VMX is backed up by the snapshot. That means restoring the snapshot will restore the old path of the ISO.
We only need to have the ISOs have distinct names and to leave the old one in place.

emlin: yeah, forgot this one.

doc/design/upgrade.md
```diff
@@ -0,0 +1,118 @@
...
+
+Both VCH and container configuration will have version and the value is same with what vic-machine version command shown. After upgrade, VCH configuration version should be updated to new version, but container VM configuration version will still be old one.
+
+New version's VCH will work with both old and new versions containers.
+
+#### Embed iso file version
+
+vic-machine need to identify the iso file version before and after upgrade. Two options here:
+- Add version into iso file Primary Volume Descriptor, and read back from vic-machine
+- Leverage iso file name including version, e.g. appliance-0.5.0.iso
```
hickeng: We should also add a product identifier as well, i.e. `vice`

This is because we already have Harbor wanting to make use of the same ISO/appliance build process meaning we'll need to be able to filter based on appliance type as well as the simple presence of the guestinfo.

We could make `vice` part of the key prefix for all of the keys - it's a better match for the namespace mechanism when we move to that.

emlin: then we should change the Makefile to generate iso files as vice-XXX.iso right?

hickeng: I didn't mean for the ISO names (although I'm struggling to think of a reason _not_ to do it there), but for the guestinfo configuration - that's what the UI plugin looks at, and how vic-machine determines that a VM is a VCH.
If the same config mechanism is used for other products then we'll need a way to filter.

emlin: I see what you think. I can change the configuration key first, otherwise, we'll have problem to detect if it's VCH or not.

emlin:
lgtm, cause we've implemented most of it.

PR #1960, opened 2016-08-12
grab ID.log file from datastore

caglar10ur:


anchal-agrawal:
LGTM 🐑 🇮🇹

hmahmood:
lgtm

caglar10ur:
dropping this PR. {id}.log could contain sensitive information so we shouldn't be grabbing it as a part of support bundle...

PR #1961, opened 2016-08-12
Abandon inconsistent images and purge images without a manifest file

### LINKED ISSUE #933: WriteImage vmdks to tmp directory, then move to final location

fdawg4l:
Write manifest file to signify doneness for images
    
    Write a manifest file (currently empty but free to use for future
    stashing of important stuff) for each image after the checksum matches.
    This is the last step in a WriteImage so we can use this to know when an
    image is complete and consistent.

    Abandon images if checksums don't match on pull
    
    This change writes the image compares the computed checksum with the
    expected checksum, and moves the disk to the image store if they match.
    Otherwise the image's directory is removed.  Subsequent pulls should
    complete without an issue.

Fixes #933

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers_test.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers_test.go
@@ -145,7 +145,7 @@ func (c *MockDataStore) ListImageStores(ctx context.Context) ([]*url.URL, error)
 	return nil, nil
 }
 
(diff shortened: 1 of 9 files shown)


caglar10ur:
LGTM :sheep: :it:

matthewavery:
lgtm

fdawg4l:
.

Issue #1781, opened 2016-08-02
Allow admin to set capacity for base disk (scratch) via vic-machine


fdawg4l:
We need to allow the admin to set the base disk capacity via `vic-machine`.  Currently we use 8gb for the base disk and it's not tunable by the user.

matthewavery:
cc @fdawg4l can you comment briefly on what reopened this issue?

fdawg4l:
I believe the change was reverted. Also, I believe @sgairo reopened it. She may have more details.

sgairo:
Correct, Matt W. asked that I revert and reopen last Thursday.



On Aug 29, 2016, at 9:26 AM, Faiyaz Ahmed <notifications@github.com<mailto:notifications@github.com>> wrote:


I believe the change was reverted. Also, I believe @sgairo<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_sgairo&d=CwMCaQ&c=Sqcl0Ez6M0X8aeM67LKIiDJAXVeAw-YihVMNtXt-uEs&r=tCZwlfDcXCdCpz3_ppVAqoiM28k06Nbbnu9QsMcIJfo&m=od5_QCsq3rd9C6nMJOUcnA9o4l86u31S8172RP1bj-c&s=-S43djhZ74SA7bxsQmHLl_xk8aR5RRJ-fbihdkbAfSQ&e=> reopened it. She may have more details.

-

gigawhitlocks:
This was reopened due to a failing integration test that I checked in. The test is fixed and the feature has been in master for a few days now, so I'm closing this.

PR #1982, opened 2016-08-15
Allow custom scratch disk size

gigawhitlocks:
Changes to the portlayer and vic-machine to support a custom scratch disk size as described in #1781

Thoughts on testing this change are appreciated; I've only been able to verify that this code works by manually changing the spec in `disk_manager.go:createDiskSpec()` to thickly provision the disk so that its true size is displayed in the UI -- I'm not exactly sure the best approach for testing this with that in mind.


lib/install/validate/storage.go
```diff
@@ -175,3 +177,34 @@ func (v *Validator) suggestDatastore(path string, label string, flag string) {
 		}
 	}
 }
+
+func parseScratchSize(input string) int {
+
+	sizeFormat := regexp.MustCompile(`(\d+)([TMGK])`)
```
hmahmood: Can we use https://github.com/docker/go-units?

cmd/vic-machine/create/create.go
```diff
@@ -239,6 +241,12 @@ func (c *Create) Flags() []cli.Flag {
 			Usage:       "Virtual Container Host x509 certificate file",
 			Destination: &c.cert,
 		},
+		cli.StringFlag{
+			Name:        "base-image-size",
```
hmahmood: Is this accurate? I am a little confused by the terminology here. Aren't we specifying the storage for images here?

cmd/vic-machine/create/create.go
```diff
@@ -239,6 +241,12 @@ func (c *Create) Flags() []cli.Flag {
...
+			Value:       "",
```
hmahmood: There is no default?

lib/portlayer/storage/config.go
```diff
@@ -31,4 +31,6 @@ type Configuration struct {
 	// Keyed by the volume store name (which is used by the docker user to
 	// refer to the datstore + path), valued by the datastores and the path.
 	VolumeLocations map[string]url.URL `vic:"0.1" scope:"read-only"`
+
+	ScratchSize int `vic:"0.1" scope:"read-only" key:"scratch_size"`
```
hmahmood: We should set a default for this somewhere, or else error out if its 0.

cmd/vic-machine/create/create.go
```diff
@@ -491,6 +499,7 @@ func (c *Create) Run(cliContext *cli.Context) error {
 	ctx, cancel := context.WithTimeout(context.Background(), c.Timeout)
 	defer cancel()
 
+	c.Data.ScratchSize = c.scratchSize
```
gigawhitlocks: @emlin is there a better place for me to do this assignment? This moves the input from the user from VirtualContainerHostConfigSpec to InstallerData and while this location _works_ I don't know that it's the best place, as it's a seemingly-random assignment in this location (to someone unfamiliar with the code at least)

lib/install/validate/storage.go
```diff
@@ -175,3 +177,34 @@ func (v *Validator) suggestDatastore(path string, label string, flag string) {
...
+	var matches []string
+	if matches = sizeFormat.FindStringSubmatch(input); len(matches) != 3 {
+		log.Errorln("Incorrect size format provided for default image size. Should be an integer followed by a unit, such as 10T, 10G, 10M, or 10K")
+		return -1
+	}
+
+	unit := matches[2]
+	var multiplier int
+	switch unit {
+	case "M":
```
gigawhitlocks: Didn't want to mess with Math.pow() cause it gives me `float`s which then have to be cast -- if someone sees a more elegant way to do this, though, I'm happy to change it

lib/install/validate/storage.go
```diff
@@ -175,3 +177,34 @@ func (v *Validator) suggestDatastore(path string, label string, flag string) {
 		}
 	}
 }
+
+func parseScratchSize(input string) int {
```
gigawhitlocks: Also I'm doing C-style error reporting here; I can make this function more verbose & use a Go-style error instead if that's more appropriate.

cmd/vic-machine/create/create.go
emlin: c.ScratchSize could be used directly after you define that in Data structure

cmd/vic-machine/create/create.go
emlin: +1, as well as the variable ScratchSize

fdawg4l: @hmahmood Kind of, not really.  Storage size for "images" implies the size of the image vmdk.  Since we map images to layers on the vmdk, the changing size of the delta vmdk is possible, but not the intent.  This size is effectively the size of the root disk which images are derived from.  Effectively, it's the size of `scratch`.

lib/install/validate/validator.go
```diff
@@ -205,6 +205,20 @@ func (v *Validator) basics(ctx context.Context, input *data.Data, conf *config.V
 	conf.SetName(input.DisplayName)
 	conf.SetDebug(input.Debug.Debug)
 	conf.Name = input.DisplayName
+
+	// Set default size of scratch disk
+	if input.ScratchSize == "" {
+		log.Warnln("Using default scratch size of 8GB")
+		input.ScratchSize = "8G"
```
emlin: looks like it's the default value, should specify that in the flag

lib/install/validate/validator.go
```diff
@@ -205,6 +205,20 @@ func (v *Validator) basics(ctx context.Context, input *data.Data, conf *config.V
 	conf.SetName(input.DisplayName)
 	conf.SetDebug(input.Debug.Debug)
 	conf.Name = input.DisplayName
+
+	// Set default size of scratch disk
+	if input.ScratchSize == "" {
```
emlin: this logic should be in storage file, in well named method

lib/config/virtual_container_host.go
```diff
@@ -125,6 +125,9 @@ type VirtualContainerHostConfigSpec struct {
 
 	// Allow custom naming convention for containerVMs
 	ContainerNameConvention string
+
+	// default size for root image of image B tree
```
fdawg4l: Not to be pedantic, but it's a B+tree.  Anyway, the impl is irrelevant.  Just nuke `of...`.

lib/install/data/data.go
```diff
@@ -70,6 +70,8 @@ type Data struct {
 
 	Force bool
 	UseRP bool
+
+	ScratchSize string
```
fdawg4l: Just out of curiousity, why is this a string while it's an `int` in `VirtualContainerHostConfigSpec`?

lib/install/validate/validator.go
```diff
@@ -205,6 +205,20 @@ func (v *Validator) basics(ctx context.Context, input *data.Data, conf *config.V
...
+		log.Warnln("Using default scratch size of 8GB")
+		input.ScratchSize = "8G"
+	}
+	scratchSize := parseScratchSize(input.ScratchSize)
+	if scratchSize < 0 { // TODO set minimum size of scratch disk
+		v.NoteIssue(errors.Errorf("Invalid default image size %s provided", input.ScratchSize))
+	} else {
+		conf.ScratchSize = scratchSize
```
emlin: Should validate if the datastore still has enough storage for this size.

lib/config/virtual_container_host.go
```diff
@@ -125,6 +125,9 @@ type VirtualContainerHostConfigSpec struct {
...
+	ScratchSize int `vic:"0.1" scope:"read-only" key:"scratch_size"`
```
fdawg4l: Do you intend for the size to be negative ever?  You may also want to mention the units expected.

lib/install/validate/storage.go
fdawg4l: Why make life hard?  `Error` is a useful thing vs implicit and/or explicit negative values.

lib/install/validate/storage.go
```diff
@@ -175,3 +177,34 @@ func (v *Validator) suggestDatastore(path string, label string, flag string) {
...
+
+	sizeFormat := regexp.MustCompile(`(\d+)([TMGK])`)
+	var matches []string
+	if matches = sizeFormat.FindStringSubmatch(input); len(matches) != 3 {
+		log.Errorln("Incorrect size format provided for default image size. Should be an integer followed by a unit, such as 10T, 10G, 10M, or 10K")
```
fdawg4l: And here's your `fmt.Errorf`

lib/install/validate/storage.go
fdawg4l: Yeah, please use `go-units`.  For one thing, @matthewavery's volume size changes uses base10 values while this expects base2.  go-units, i think, does the right thing regardless.  @matthewavery would have more details.

lib/install/validate/validator.go
gigawhitlocks: yeah I wasn't 100% sure where this should go (a common problem with me) so I was expecting this feedback thank you @emlin

lib/install/validate/validator.go
gigawhitlocks: Yep, you're right, and then I can eliminate this logic.

cmd/vic-machine/create/create.go
emlin: @fdawg4l I thought it's the image datastore size.
Is there any data in the scratch disk? I'm just not clear what does this mean for end user

hmahmood: The option name is just a little confusing to me. Do you guys have any other suggestions?

lib/install/validate/validator.go
gigawhitlocks: Sure, I'll emit a warning too.

lib/portlayer/storage/config.go
gigawhitlocks: Yeah Emma caught this here https://github.com/vmware/vic/pull/1982/files#r74835090 <- I'll alter the flag to provide a default there so that 8G will be provided if nothing else is.

lib/install/data/data.go
gigawhitlocks: Comes in as a string from the user on the command line so that they can use G/K/T/P to specify units. Gets parsed into an int (in kilobytes, because that's what vSphere takes) before it's stored in the `VirtualContainerHostConfigSpec`

lib/install/validate/storage.go
gigawhitlocks: SRSLY lol

lib/install/validate/storage.go
gigawhitlocks: Didn't know it existed. In any case, I was half expecting something like this to be brought up to me in review, so I didn't spend long throwing together this parser. I'll strip it out & use the library -- thanks @fdawg4l

lib/config/virtual_container_host.go
gigawhitlocks: roger LOL

lib/config/virtual_container_host.go
gigawhitlocks: It should definitely never be negative. I'll add a comment.

cmd/vic-machine/create/create.go
gigawhitlocks: Honestly it's confusing to me too but I have been struggling to think of a better name -- paging @hickeng

cmd/vic-machine/create/create.go
gigawhitlocks: Default will be set to 8G per Emma's comment below

lib/install/validate/storage.go
gigawhitlocks: Yeah you're right lol idk why I left it like this

cmd/vic-machine/create/create.go
gigawhitlocks: I'm going to ping you about this out-of-band

cmd/vic-machine/create/create.go
fdawg4l: Base is fine.  I'd rather not inherit `docker`'s naming convention.  It's the base disk images are written to.  That's it.

lib/config/virtual_container_host.go
fdawg4l: If it should never be negative, use use `unsigned` types.

lib/install/validate/validator.go
fdawg4l: I disagree we need to validate free space.  Our disks are sparse.  It is possible to have 8 8GB disks on a 1GB datastore and happily run as long as the data written didn't exceed the datastore size.  This qualifies as a quota check and that's outside this change and our intent.

gigawhitlocks:
Comments are addressed except for warning the user if they don't have enough space for the size of the scratch layer, and I need to add tests and docs.

emlin:
> Comments are addressed except for warning the user if they don't have enough space for the size of the scratch layer

As it's only the size of scratch disk, it's not that necessary to warn the datastore capacity for itself. For we plan to put images to this datastore, only bigger than scratch disk is not enough. Just wondering why the scratch disk is 8GB, isn't that too big?

As @matthewavery said, the disk is thin provisioned, it does not consume more spaces if we don't write data into it, why should user care about this configuration at this time?
So to avoid confusion here, please add hidden attribute into the flag, then this option will not be shown in general help -h, instead it will be in -x, advanced options.

lib/install/validate/validator.go
```diff
@@ -205,6 +208,15 @@ func (v *Validator) basics(ctx context.Context, input *data.Data, conf *config.V
 	conf.SetName(input.DisplayName)
 	conf.SetDebug(input.Debug.Debug)
 	conf.Name = input.DisplayName
+
+	scratchSize, err := units.FromHumanSize(input.ScratchSize)
+	if err != nil { // TODO set minimum size of scratch disk
+		v.NoteIssue(errors.Errorf("Invalid default image size %s provided", input.ScratchSize))
+	} else {
+		conf.ScratchSize = scratchSize / bytesToKilobytes
```
hmahmood: Use units.KB instead of bytesToKilobytes

lib/portlayer/storage/vsphere/image.go
```diff
@@ -222,7 +222,8 @@ func (v *ImageStore) WriteImage(ctx context.Context, parent *portlayer.Image, ID
 	// will be descended from this created and prepared fs.
 	if ID == portlayer.Scratch.ID {
 		// Create the disk
-		vmdisk, cerr := v.dm.CreateAndAttach(ctx, imageDiskDsURI, "", defaultDiskSize, os.O_RDWR)
+		log.Debugf("Default disk size reported as %d", portlayer.Config.ScratchSize)
+		vmdisk, cerr := v.dm.CreateAndAttach(ctx, imageDiskDsURI, "", int64(portlayer.Config.ScratchSize), os.O_RDWR)
```
hmahmood: Do you need the cast to int64?

lib/install/validate/validator.go
gigawhitlocks: makes sense

lib/portlayer/storage/vsphere/image.go
gigawhitlocks: not anymore, thanks for noticing it, the output from the `units` conversion is an int64

gigawhitlocks:
@emlin, @matthewavery and I were discussing why the user would want to configure this yesterday and I was planning to talk to you about the warning before this went in so I'm happy to see you're actually on the same page as us already. I will absolutely make this a hidden flag.

emlin:
lgtm 👍

andrewtchin:
lgtm

gigawhitlocks:
Working on adding an integration test before I check in but thanks @emlin and @andrewtchin

lib/install/validate/validator.go
matthewavery: Ha thanks for this, I also made this change for unit handling in `VolumeCreate` 👍

lib/portlayer/storage/vsphere/image.go
```diff
@@ -366,7 +366,7 @@ func (v *ImageStore) scratch(ctx context.Context, storeName string) error {
 	log.Infof("Creating image %s (%s)", portlayer.Scratch.ID, imageDiskDsURI)
 
 	// Create the disk
-	vmdisk, err := v.dm.CreateAndAttach(ctx, imageDiskDsURI, "", defaultDiskSize, os.O_RDWR)
+	vmdisk, err := v.dm.CreateAndAttach(ctx, imageDiskDsURI, "", portlayer.Config.ScratchSize, os.O_RDWR)
```
fdawg4l: Purely for convenience, can we check that this is non zero and if it is zer, use the defaultDIskSize?  The reason I ask is the `vsphere/image_test.go` file expects scratch to be larger than 0.  And since the tests work without vic machine creating that config struct (`portlayer.Config`), they'll fail unless you want to create the config struct in the test `setup()` function and set the size.

gigawhitlocks: I can do this during `vic-machine create` time @fdawg4l so that we bail & do not create the VCH if 0 is specified, or we can log a warning and supply the default at that time. We could also impose a sane lower limit instead of 0.

fdawg4l:
lgtm. Try running the image tests on a nested test setup just to make sure we aren't breaking them.  Thanks!

```
VIC_ESX_TEST_DATASTORE=datastore1 VIC_ESX_TEST_URL=root:ca\$hc0w@10.118.67.200 sudo -E go test -v github.com/vmware/vic/lib/portlayer/storage/vsphere/...
```

gigawhitlocks:
I'm currently doing my testing against a nested ESX -- I can run the unit tests specifically too

gigawhitlocks:
OK an actual functional test is added. @mhagen will be updating the integration test image to include unzip so this will pass; the references and everything except the Dockerfile update are included here in this PR

MHagen:
Think you've added the wrong MHagen. I'm not part of this work.

gigawhitlocks:
.

PR #1984, opened 2016-08-15
WIP: Feature/new build process [skip ci]

mhagen-vmware:
Alright, kind of a mess with drone and github but the general idea is:
1. Create two drone servers, with two yml files: drone-master and drone-pr
2. drone-pr.yml should run unit and regression on a pr from a developer, then take the patch from the PR and create a new temp branch and make a PR on that branch
3. Once that PR is created, drone-master should pick it up and run all integration tests on it.

TODO:
1. close the original PR? do we want to do this?
2. somehow merge the new PR - not sure if this is even possible using github API, also we need to ask if we want to automatically do this?
3. need to somehow check for lgtms... again, not potentially super easy to do...


hickeng:
@mhagen-vmware 
This looks sane, however I'm not sure how we use it without 2 & 3 from your TODOs.
Without the LGTM check we cannot tell whether we should be running integration on in
Without the automatic merge on pass it's not a pipeline and we end up with the same race-to-merge issues that we have currently.

mhagen-vmware:
@hickeng Actually by virtue of pipelining a single job through the final step we have eliminated the race to merge, but I was able to figure out number 2 and number 3 with the help of Fabio and Ivan. I am still not sure if we ultimately want number 1 or not, but on quick inspection, it doesn't look like github api supports a close pr call.

mhagen-vmware:
oh, looks like I can actually do a update call in order to close the original PR. Let me know if we want that or not, and I can add that also fairly easily.

PR #1988, opened 2016-08-15
WIP trace contexts

fdawg4l:
Fixes #

I'm not convinced we shouldn't just implement the context interface.
Then we can embed whatever we want (stack traces, deltas, etc) in the
Err().

### DIFF
--- a/pkg/trace/context.go
+++ b/pkg/trace/context.go
@@ -17,7 +17,6 @@ package trace
 import (
 	"fmt"
 	"sync/atomic"
-	"time"
 
 	"github.com/Sirupsen/logrus"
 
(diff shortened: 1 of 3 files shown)


fdawg4l:
```
=== RUN   TestDeadlineLogging
DEBU[0000] [BEGIN] [github.com/vmware/vic/pkg/trace.Start] testmsg 
ERRO[0000] op=101 (delta:2.582µs): github.com/vmware/vic/pkg/trace.Start testmsg: error: context deadline exceeded 
--- PASS: TestDeadlineLogging (0.00s)
```

fdawg4l:
Hmm...   It occurs to me know the `op` is only being printed on `Log` messages and not on `Start`.

PR #1994, opened 2016-08-16
Bump to golang 1.7

caglar10ur:


mdubya66:
I approve this change now that Go Lang has GA'd.

cmd/docker/main.go
```diff
@@ -176,7 +176,28 @@ func startServerWithOptions(cli *CliOptions) *apiserver.Server {
 		Version: "1.22", //dockerversion.Version,
 	}
 
-	tlsConfig := tlsconfig.ServerDefault
+	copyTLSConfig := func(c *tls.Config) *tls.Config {
```
fdawg4l: Does this need to be in an anonymous func?  I only see it used once immediately after declaring.

vendor/manifest
```diff
@@ -34,7 +34,7 @@
 			"importpath": "github.com/HewlettPackard/gas",
 			"repository": "https://github.com/HewlettPackard/gas",
 			"vcs": "git",
-			"revision": "12d370b11b9e6a94368ebd229b020bb55b51fe69",
+			"revision": "37205e9afa56150a634a1d0d19347446e9b27f17",
```
fdawg4l: Is the bump for gas required for go1.7?

cmd/docker/main.go
caglar10ur: Could be, will do that. I added this as a temporary workaround until the library we use fixes their code :)

cmd/docker/main.go
```diff
@@ -176,7 +176,27 @@ func startServerWithOptions(cli *CliOptions) *apiserver.Server {
 		Version: "1.22", //dockerversion.Version,
 	}
 
-	tlsConfig := tlsconfig.ServerDefault
+	tlsConfig := func(c *tls.Config) *tls.Config {
```
hickeng: ?

caglar10ur: go vet is/was whining about ServerDefault copying locks

caglar10ur: cmd/docker/main.go:179: assignment copies lock value to tlsConfig: crypto/tls.Config contains sync.Once contains sync.Mutex

hickeng:
LGTM - but why the need to explicitly expand the tls config?

fdawg4l:
LGTM

caglar10ur:
♻️

Issue #812, opened 2016-05-30
vic-machine: first support of upgrade


emlin:
vic-machine upgrade should be able to update existing VCH to latest version. 

This issue includes the upgrade mechanism design, and prepare for upgrade from 1.0 to future version.
If necessary, should be able to update binary of existing TP customers to latest GA version. 

At this time, it's not possible to know which meta-data or configuration has to be changed to have GA binaries running in TP3 VM. But for any modules, after TP3, need to consider the binaries upgrade path.

hickeng:
@emlin The scope of this is too broad - all we need to support for GA is that vic-machine be able to update itself. That allows a user to easily get a version of vic-machine that *can* perform an upgrade to the latest bits.

emlin:
@hickeng so as my understanding, this command is just to help user find new VIC versions, and download to local.
We're not going to replace anything runtime, e.g. update existing VCH, etc.

emlin:
Start to work on design

emlin:
@hickeng Here is the draft based on my understanding, and with some assumption or questions inside. https://github.com/emlin/vic/blob/upgrade/doc/design/upgrade.md
As not be able to confirm with you for all issues, please have a look while you're working on the upgrade document.

emlin:
As the upgrade feature scope is not clarified yet, move this issue back to TODO.
The #2031 is just pre upgrade support.

emlin:
Convert this issue to Epic, to split upgrade related tasks.

emlin:
@karthik-narayan We'll have upgrade ready for GA, do we have vic-e upgrade matrix? Or is there common definition for vSphere?
We'll need to run integration test based on that, and of course user document as well. @stuclem

emlin:
@hickeng @mdubya66 ^^^

PR #2031, opened 2016-08-17
upgrade versioning

emlin:
This PR add version information to iso files, and check the version during vic-machine create.
If the version does not match, create will fail, to ignore this failure, need to specify --force

For #812 

Note: this force installation should be fore developer only. Customer should be careful if they have this version mismatch issue, that means they are using binaries from different build.


caglar10ur:
LGTM :sheep: :it:

andrewtchin:
lgtm

emlin:
♻️

Issue #2028, opened 2016-08-17
Error while pulling image manifest: Unexpected http code: 401, URL: http://registry-1.docker.io/v2/library/busybox/manifests/latest	


mhagen-vmware:
Saw this on one of the recent builds:
[integration_logs_4108_874f5840ed1e42f06dca17d602cdc3c68a7ec0e1_1471448360.zip](https://github.com/vmware/vic/files/423120/integration_logs_4108_874f5840ed1e42f06dca17d602cdc3c68a7ec0e1_1471448360.zip)

mhagen-vmware:
from imagec.log:
```
time="2016-08-17T15:13:11Z" level=debug msg="URL: http://registry-1.docker.io/v2/library/busybox/manifests/latest" 
[...]

Issue #2011, opened 2016-08-16
Failed to fetch image blob


mhagen-vmware:
Saw this on a recent CI build:
[integration_logs_4070_fe089a13384da6e9e62355813a33948ccb8da951_1471380846.zip](https://github.com/vmware/vic/files/421275/integration_logs_4070_fe089a13384da6e9e62355813a33948ccb8da951_1471380846.zip)

mhagen-vmware:
```
time="2016-08-16T19:59:06Z" level=info msg="Image cache updated successfully" 
time="2016-08-16T19:59:16Z" level=error msg="imagec exit code: exit status 1" 
time="2016-08-16T19:59:16Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Image).PullImage] [11.582279128s] PullImage" 
time="2016-08-16T19:59:16Z" level=error msg="Handler for POST /v1.21/networks/test-network/connect returned error: container fakeContainer not found" 
time="2016-08-16T19:59:16Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).ContainerCreate] ContainerCreate" 
time="2016-08-16T19:59:16Z" level=error msg="ContainerCreate: image busybox error: No such image: busybox" 
time="2016-08-16T19:59:16Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).ContainerCreate] [94.328µs] ContainerCreate" 
time="2016-08-16T19:59:16Z" level=error msg="Handler for POST /v1.21/containers/create returned error: No such image: busybox" 
```

mhagen-vmware:
```
time="2016-08-16T19:59:06Z" level=debug msg="Creating a store from input 564d6a5d-f2ee-3ae4-c00f-073a534116b9" 
time="2016-08-16T19:59:06Z" level=debug msg="Store already exists" 
time="2016-08-16T19:59:06Z" level=debug msg="URL: https://registry-1.docker.io/v2/library/busybox/blobs/sha256:a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4\n " 
[...]

PR #2032, opened 2016-08-17
[imagec] introduce retry logic with backoff

### LINKED ISSUE #2015: Failed to fetch OAuth token: Get https://auth.docker.io/token?scope=repository%!A(MISSING)library%!F(MISSING)busybox%!A(MISSING)pull&service=registry.docker.io: dial tcp:
Saw this failure on a recent CI build:

[integration_logs_4080_5685722d6a53cec1a6f3a82c7d2fc5664b06a35b_1471382669.zip](https://github.com/vmware/vic/files/421318/integration_logs_4080_5685722d6a53cec1a6f3a82c7d2fc5664b06a35b_1471382669.zip)

caglar10ur:
Hopefully fixes #2015, #2011 and #2028

### DIFF
--- a/cmd/imagec/fetcher.go
+++ b/cmd/imagec/fetcher.go
@@ -37,10 +37,13 @@ import (
 	"github.com/vmware/vic/pkg/trace"
 )
 
+const (
+	maxDownloadAttempts = 5
+)
+
 // Fetcher interface
 type Fetcher interface {
-	Fetch(url *url.URL) (string, error)
-	FetchWithProgress(url *url.URL, ID string) (string, error)
+	Fetch(url *url.URL, id ...string) (string, error)
 
 	Head(url *url.URL) (http.Header, error)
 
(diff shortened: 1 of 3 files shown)


hmahmood:
I am not sure this is fixing the problem. The code also looks pretty hairy, and I am nervous to just introduce it into the already unstable mess that we have.

caglar10ur:
Not sure why you think the code is hairy, it's pretty straightforward and exactly what docker does and also I'm not sure what will be the harm. At worst case this will fail 5 times instead of giving up on 1st try.

hmahmood:
I am a little surprised that occasional 401's from the docker registry are routine. Nonetheless, retries don't hurt.

lgtm

caglar10ur:
I'm nearly sure that they are keeping those oauth tokens in memory (as there is no real reason to persist them) so any failure on their side will end up returning a 401 to the caller. It's just a design choice they made to make their life easier/

cmd/imagec/fetcher.go
```diff
@@ -98,24 +101,69 @@ func NewURLFetcher(options FetcherOptions) Fetcher {
...
-	return u.fetch(ctx, url, "")
-}
-
-// FetchWithProgress fetches a web page from url and stores in a temporary file while showing a progress bar.
-func (u *URLFetcher) FetchWithProgress(url *url.URL, ID string) (string, error) {
-	defer trace.End(trace.Begin(url.String()))
+	var name string
+	var err error
+	var retries int
+	for {
```
hickeng: not that I fundamentally object, but why not use the loop variables?
```
for attempts := 1; attempts < maxDownloadAttempts; attempts++ {
  ...
}
log.Debugf("Download failed: %v", err)
return "", err
```

cmd/imagec/fetcher.go
```diff
@@ -98,24 +101,69 @@ func NewURLFetcher(options FetcherOptions) Fetcher {
...
+				}
+			case <-ctx.Done():
+				ticker.Stop()
+				return "", fmt.Errorf("download cancelled during retry delay")
+			}
+		}
+	}
 }
 
 func (u *URLFetcher) fetch(ctx context.Context, url *url.URL, ID string) (string, error) {
```
hickeng: Would like a function comment explaining the behaviour with/without ID==""

hickeng:
questions, but mostly LGTM

cmd/imagec/fetcher.go
caglar10ur: done

cmd/imagec/fetcher.go
caglar10ur: done

Issue #1991, opened 2016-08-15
vicadmin server consumes 100% CPU in 0.5 tag candidate


corrieb:
I have tried to reproduce the steps I went through initially, but have been unable to. However, I have a log bundle with pprof enabled which I will attach here. Logging into the appliance confirms with top that it's the admin server looping, not being restarted.

I will keep the VCH alive as long as I can so that it can be poked if necessary.

[logs.zip](https://github.com/vmware/vic/files/419166/logs.zip)

sgairo:
@corrieb were you running multiple VCH off of the same host?

corrieb:
This is against VC. Yes I think there was another VCH on there at the time

hickeng:
I grabbed a profile via :6061/debug/pprof/profile (which we really should include in the bundles!).
[vic-admin.zip](https://github.com/vmware/vic/files/419188/vic-admin.zip)

The `top` from profile is:
```
vagrant@devbox:~$ go tool pprof ./vicadmin ./vic-admin.profile
Entering interactive mode (type "help" for commands)
(pprof) top
29.59s of 29.96s total (98.77%)
Dropped 10 nodes (cum <= 0.15s)
Showing top 10 nodes out of 15 (cum >= 0.16s)
      flat  flat%   sum%        cum   cum%
    22.74s 75.90% 75.90%     22.74s 75.90%  bytes.LastIndex
     4.77s 15.92% 91.82%      6.62s 22.10%  syscall.Syscall6
     0.66s  2.20% 94.03%      0.66s  2.20%  runtime/internal/atomic.Cas
     0.32s  1.07% 95.09%      0.83s  2.77%  runtime.reentersyscall
     0.22s  0.73% 95.83%      0.66s  2.20%  runtime.casgstatus
     0.20s  0.67% 96.50%     29.94s 99.93%  main.findSeekPos
     0.18s   0.6% 97.10%      0.91s  3.04%  runtime.exitsyscall
     0.18s   0.6% 97.70%      0.40s  1.34%  runtime.exitsyscallfast
     0.16s  0.53% 98.23%      6.89s 23.00%  os.(*File).ReadAt
     0.16s  0.53% 98.77%      0.16s  0.53%  runtime/internal/atomic.Store
(pprof)
```

and cross-referencing that with the stack traces from the bundle then this thread is maxing CPU:
```
goroutine 23 [runnable]:
[...]

PR #2035, opened 2016-08-17
Test findSeekPos in vicadm.go

kjplatz:
Fixes and includes unit tests #1991 

This cleans up the findSeekPos() routine in cmd/vicadm/logs.go and includes unit tests.  This includes 16 randomized unit tests, plus one unit test that includes extremely long lines (the condition that should cause the #1991 issue)

### DIFF
--- a/cmd/vicadmin/logs.go
+++ b/cmd/vicadmin/logs.go
@@ -363,42 +363,43 @@ func tailFile(wr io.Writer, file string, done *chan bool) error {
 func findSeekPos(f *os.File) int64 {
 	defer trace.End(trace.Begin(""))
 	nlines := tailLines
-	pos, err := f.Seek(0, 2)
-	// If for some reason we can't seek, we will just start tailing from
-	// beginning-of-file
+	readPos, err := f.Seek(0, 2)
+	// If for some reason we can't seek, we will just start tailing from beginning-of-file
 	if err != nil {
 		return int64(0)
 	}
 
-	// Last newline we've seen (from EOF).  Initialize to EOF
-	lastnl := pos
 	// Buffer so we can seek nBytes (default: 1k) at a time
 	buf := make([]byte, nBytes)
-	for lastnl > 0 && nlines > 0 {
-		// Go back nBytes from the last newline we've seen (or beginning-of-file),
+
+	for readPos > 0 {
+		// Go back nBytes from the last readPos we've seen (stopping at beginning-of-file)
 		// and read the next nBytes
-		readPos := lastnl - int64(len(buf))
-		if pos < 0 {
+		readPos -= int64(len(buf))
+		if readPos < 0 {
+			// We don't want to overlap our read with previous reads...
+			buf = buf[:(int(readPos)+nBytes)]
 			readPos = 0
 		}
 		bufend, err := f.ReadAt(buf, readPos)
 
 		// It's OK to get io.EOF here.  Anything else is bad.
 		if err != nil && err != io.EOF {
-			log.Errorf("Error reading from logfile: %s", err)
+			log.Errorf("Error reading from file %s: %s", f.Name(), err)
 			return 0
 		}
 
 		// Start from the end of the buffer and start looking for newlines
-		off := bufend
-		for nlines > 0 && off >= 0 {
-			off = bytes.LastIndex(buf[:bufend], []byte("\n"))
-			if off >= 0 {
-				nlines--
-				bufend = off
+		for bufend > 0 {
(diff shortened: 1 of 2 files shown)


cmd/vicadmin/logs.go
```diff
@@ -378,14 +378,15 @@ func findSeekPos(f *os.File) int64 {
 		// Go back nBytes from the last newline we've seen (or beginning-of-file),
 		// and read the next nBytes
 		readPos := lastnl - int64(len(buf))
-		if pos < 0 {
+		// Fix here.  Previously was "pos < 0".
+		if readPos < 0 {
 			readPos = 0
 		}
 		bufend, err := f.ReadAt(buf, readPos)
```
hickeng: bufend isn't the end of the buffer, it's the number of bytes read. The index in the buffer should probably be `readPos + bufend`

cmd/vicadmin/logs.go
```diff
@@ -378,14 +378,15 @@ func findSeekPos(f *os.File) int64 {
 		// Go back nBytes from the last newline we've seen (or beginning-of-file),
 		// and read the next nBytes
 		readPos := lastnl - int64(len(buf))
-		if pos < 0 {
+		// Fix here.  Previously was "pos < 0".
```
hickeng: this comment isn't useful in the code. My assumption is that you're trying to ensure that we don't go beyond the beginning of the file?

kjplatz: Yeah, we don't want to try reading from a negative offset.  That was generating an error. I'll nix that comment after the rest of the comments come through.

cmd/vicadmin/logs.go
```diff
@@ -378,14 +378,15 @@ func findSeekPos(f *os.File) int64 {
 		// Go back nBytes from the last newline we've seen (or beginning-of-file),
 		// and read the next nBytes
 		readPos := lastnl - int64(len(buf))
```
hickeng: why are we reading from the location of the last newline? We should be using the last value of readPos.

Work through the current logic when you've got a buffer sized 1000, a file sized 2000 and no newline in it:
1. readPos = 2000 - 1000 = 1000
2. bufend = 1000
3. off = bufend = 1000
3a. in loop: off = -1 (not found)
4. lastnl = readPos + bufend = 1000 + 1000 = 2000

And that's an infinite loop.

cmd/vicadmin/logs.go
```diff
@@ -378,14 +378,15 @@ func findSeekPos(f *os.File) int64 {
...
 			readPos = 0
 		}
 		bufend, err := f.ReadAt(buf, readPos)
 
 		// It's OK to get io.EOF here.  Anything else is bad.
 		if err != nil && err != io.EOF {
-			log.Errorf("Error reading from logfile: %s", err)
+			log.Errorf("Error reading from file %s: %s", f.Name(), err)
 			return 0
 		}
```
hickeng: Why are we referencing bufend in the loop and changing it? Why not use the single byte version of index for a single byte? If you're going to do it this via ReadAt I'd phrase it this way (pseudo-code):
```
for readPos := pos; readPos >0; {
  readPos-= len(buf)
  if readPos < 0 {
    readPos = 0
  }
  bufend, err := f.ReadAt(buf, readPos)
  if err != nil ...

  // scan buffer for newlines
  for end := bufend; end >0; {
    end = bytes.LastIndexByte(buf[:end], '\n')
    if end < 0 {
      // go idiom is to check err/exit conditions first, then return/break
      break
    }

    // logic body after validation
    nlines--
    if nlines == 0 {
      // return directly with file offset
      return readPos + end
    }
  }

  // didn't find enough newlines
  return 0
}

```

cmd/vicadmin/logs.go
```diff
@@ -378,14 +378,15 @@ func findSeekPos(f *os.File) int64 {
 		// Go back nBytes from the last newline we've seen (or beginning-of-file),
 		// and read the next nBytes
 		readPos := lastnl - int64(len(buf))
-		if pos < 0 {
+		// Fix here.  Previously was "pos < 0".
+		if readPos < 0 {
```
hickeng: This could easily be the issue as Go source makes it very unclear what the behaviour would be with a negative position. Looks like it gets passed unmodified to syscall.pread.

hickeng:
read it through and haven't spotted other infinite loop conditions - LGTM

hmahmood:
lgtm

kjplatz:
bing

kjplatz:
Bing

kjplatz:
Bing

PR #2041, opened 2016-08-18
introduce DoNotRetry type

caglar10ur:
to signal fetcher to not to retry.

Fixes the latest CI problem and is a starting point for #1710

### DIFF
--- a/cmd/imagec/errors.go
+++ b/cmd/imagec/errors.go
@@ -1,0 +1,25 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+// DoNotRetry is an error wrapper indicating that the error cannot be resolved with a retry.
+type DoNotRetry struct {
+	Err error
+}
+
+// Error returns the stringified representation of the encapsulated error.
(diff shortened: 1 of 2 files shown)


andrewtchin:
lgtm

cmd/imagec/errors.go
```diff
@@ -0,0 +1,25 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+// DoNotRetry is an error wrapper indicating that the error cannot be resolved with a retry.
+type DoNotRetry struct {
+	Err error
```
hmahmood: Do you need this? I think if you just say `error`, you don't have to implement `Error()` below.

hmahmood:
lgtm

cmd/imagec/errors.go
caglar10ur: Yep, otherwise you cannot create one via

```
 DoNotRetry{Err: fmt.Errorf("image not found")}
```

hmahmood: you can do
```
DoNotRetry{error: fmt.Errorf("image not found")}
```

caglar10ur: This only works if they are in the same pkg, also error is a reserved word :) I have a PR moving this (and some more) into errors pkg

hmahmood: I don't understand why it has to be the same package. You are just embedding `error`; see the example in https://blog.golang.org/error-handling-and-go

caglar10ur: Because only the exported members of Embedded types are exported. You cannot do what you suggest as compiler will throw an error like ```...DoNotRetry field 'error' in struct literal```. You will need a NewDoNoRetry function to achieve what you want but since you are trying to save a function this won't work as it will introduce one.

hmahmood: Well normally I would not have users of the package create errors that are in the package. They are only exported so that callers can type assert or inspect the error. In this case if you want callers to get to the actual underlying error then yes you can't just embed error and what you have is appropriate. I don't see that is your PR right now but perhaps it would be useful in the future.

PR #2052, opened 2016-08-19
Revert "We shouldn't retry if we don't have a token and received 401"

mdubya66:
Reverts vmware/vic#2049

PR #2054, opened 2016-08-19
Revert "Revert "We shouldn't retry if we don't have a token and received 401""

caglar10ur:
Reverts vmware/vic#2052

local testing showed that the latest failure is something new and not caused by this and since this is fixing another bug I'm reverting the revert


hickeng:
LGTM

fdawg4l:
LGTM

Issue #1986, opened 2016-08-15
[6.5] vic-machine port check fails


caglar10ur:
6.5 comes with 8080 disabled. We should find a default that works with 5.5/6.0 and 6.5

```
INFO[2016-08-15T21:47:12Z] DRS check SKIPPED - target is standalone host
ERRO[2016-08-15T21:47:12Z] --------------------
ERRO[2016-08-15T21:47:12Z] Firewall must permit 8080/tcp outbound to use VIC
ERRO[2016-08-15T21:47:12Z] Create cannot continue: configuration validation failed
ERRO[2016-08-15T21:47:12Z] --------------------
ERRO[2016-08-15T21:47:12Z] vic-machine-linux failed: validation of configuration failed
```

hmahmood:
Don't believe this is a problem anymore since we switched to using port 2377.

PR #2056, opened 2016-08-19
Changing attach and port layer default port to 2377 from 8080

### LINKED ISSUE #2006: Cannot map container port to 8080 (not available)
root@ubuntu-1604-vmware:~/vic/src/github.com/vmware/vic-n# docker run -d -p 8080:80 nginx
4c85c24058372f10334a419c077c333bbd70546f2abbef069ac01f93ea4b5c71
docker: Error response from daemon: error mapping ports: port 8080 is not available.

Speculation on vic-dev is that the port layer is published on port 8080. Many customer demos will expect to be able to use port 8080 to publish on, so this is a major UX issue.

Also related to #1986 which is a different problem but has the same solution

Our internal services should use obscure port numbers that we document as being unavailable. Would be nice if we can detect these port numbers and say "this is in use internally by the VCH" rather than just "not available".

hmahmood:
Fixes #2006

### DIFF
--- a/lib/install/validate/config.go
+++ b/lib/install/validate/config.go
@@ -23,6 +23,7 @@ import (
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/lib/portlayer/constants"
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/trace"
 	"golang.org/x/net/context"
@@ -38,7 +39,7 @@ func (v *Validator) CheckFirewall(ctx context.Context) {
 	var err error
 
 	rule := types.HostFirewallRule{
-		Port:      8080, // serialOverLANPort
+		Port:      constants.SerialOverLANPort, // serialOverLANPort
 		PortType:  types.HostFirewallRulePortTypeDst,
 		Protocol:  string(types.HostFirewallRuleProtocolTcp),
 		Direction: types.HostFirewallRuleDirectionOutbound,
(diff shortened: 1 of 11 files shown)


lib/install/validate/config.go
```diff
@@ -38,7 +39,7 @@ func (v *Validator) CheckFirewall(ctx context.Context) {
 	var err error
 
 	rule := types.HostFirewallRule{
-		Port:      8080, // serialOverLANPort
+		Port:      exec.SerialOverLANPort, // serialOverLANPort
```
hickeng: we really should try to avoid pulling the entire portlayer into the installer just for constants.
Can we split this into a seperate portlayer/exec/consts package if this is necessary.

hickeng:
The port layer listening on localhost:8080 should be completely irrelevant to this issue so no need for the changes there.

The only reason we're seeing this is because the client interface and the management interface are shared. @hmahmood didn't you mention adding a duplicate interface on the same network specifically for the port forwarding work?

hmahmood:
@hickeng yes it is because we can share the management and external networks that we are seeing this issue. So I tried adding separate interfaces, but that is generally frowned upon (and difficult to get right): http://serverfault.com/questions/336021/two-network-interfaces-and-two-ip-addresses-on-the-same-subnet-in-linux . So I never added the interface.

Given that, there really no choice but to change the port for at least the attach server.

tests/resources/Util.robot
```diff
@@ -70,6 +70,8 @@ Get Docker Params
 Install VIC Appliance To Test Server
     [Arguments]  ${certs}=${false}  ${vol}=default
     Set Test Environment Variables  ${certs}  ${vol}
+    # disable firewall
+    Run  govc host.esxcli network firewall set -e false
```
mhagen-vmware: lgtm

hmahmood:
@hickeng ideally we can keep the port for the port layer at 8080. The problem is that when we check if that port is available, we don't use the client interface's ip address to do a socket connect. The reason I didn't do this was because the ip address of the client interface can change if its using DHCP. The iptables rule will have the ip address in it as well, so if it changes, we won't be able to remove it once the container stops (since the rule won't match).

cgtexmex:
LGTM

emlin:
As in 6.5, 8080 is not usable as well #1986, I think it's fine to fix this issue with a changed port first. We'll need to figure out one port common for both 6.5 and 6.0.

LGTM

hmahmood:
:recycle:

hmahmood:
:recycle:

Issue #1896, opened 2016-08-09
Another task is already in progress


hickeng:
This is the first time I've actually seen this but we've known it will occur for a while.

This should be fixed via a retry mechanism in the tasks package - this and extraction of error messages is explicitly what that package was intended for.

The problem here is that scratch.vmdk preparation is still underway when we try to create a volume, causing a failure in the appliance reconfigure.
```
time="2016-08-10T05:39:15Z" level=info msg="Refreshing image cache from datastore." 
time="2016-08-10T05:39:15Z" level=info msg="Creating directory [datastore1] VIC/564d3686-b4d2-1f65-6b6e-5c19852997e9/images" 
time="2016-08-10T05:39:15Z" level=info msg="Checking for inconsistent images on http://VCH-0-9650/storage/images/564d3686-b4d2-1f65-6b6e-5c19852997e9" 
time="2016-08-10T05:39:15Z" level=info msg="Creating directory [datastore1] VIC/564d3686-b4d2-1f65-6b6e-5c19852997e9/images/scratch" 
time="2016-08-10T05:39:16Z" level=info msg="Creating directory [datastore1] VIC/564d3686-b4d2-1f65-6b6e-5c19852997e9/images/scratch/imageMetadata" 
time="2016-08-10T05:39:16Z" level=info msg="Creating image scratch ([datastore1] VIC/564d3686-b4d2-1f65-6b6e-5c19852997e9/images/scratch/scratch.vmdk)" 
time="2016-08-10T05:39:16Z" level=debug msg="[BEGIN] [github.com/vmware/vic/pkg/vsphere/disk.(*Manager).CreateAndAttach] [datastore1] VIC/564d3686-b4d2-1f65-6b6e-5c19852997e9/images/scratch/scratch.vmdk" 
time="2016-08-10T05:39:16Z" level=info msg="Create/attach vmdk [datastore1] VIC/564d3686-b4d2-1f65-6b6e-5c19852997e9/images/scratch/scratch.vmdk from parent " 
time="2016-08-10T05:39:19Z" level=debug msg="Mapping vmdk to pci device [datastore1] VIC/564d3686-b4d2-1f65-6b6e-5c19852997e9/images/scratch/scratch.vmdk" 
time="2016-08-10T05:39:19Z" level=debug msg="[BEGIN] [github.com/vmware/vic/pkg/vsphere/disk.findDisk] VirtualMachine:547" 
[...]

Issue #1725, opened 2016-07-28
Debug why "Another task is already in progress" when detaching/attaching disks quickly


fdawg4l:
```
=== RUN   TestListImageStore
time="2016-07-28T18:33:38Z" level=info msg="Creating directory [datastore1] dd024095-be66-47-imageTests" 
time="2016-07-28T18:33:38Z" level=info msg="Datastore path is [datastore1] dd024095-be66-47-imageTests" 
time="2016-07-28T18:33:38Z" level=info msg="Refreshing image cache from datastore." 
time="2016-07-28T18:33:38Z" level=info msg="Creating directory [datastore1] dd024095-be66-47-imageTests/storeName0/images" 
time="2016-07-28T18:33:38Z" level=info msg="Creating directory [datastore1] dd024095-be66-47-imageTests/storeName0/images/scratch" 
time="2016-07-28T18:33:38Z" level=info msg="Creating image scratch ([datastore1] dd024095-be66-47-imageTests/storeName0/images/scratch/scratch.vmdk)" 
time="2016-07-28T18:33:38Z" level=info msg="Create/attach vmdk [datastore1] dd024095-be66-47-imageTests/storeName0/images/scratch/scratch.vmdk from parent " 
time="2016-07-28T18:33:38Z" level=error msg="vmdk storage driver failed to attach disk: Another task is already in progress." 
--- FAIL: TestListImageStore (1.14s)
        Error Trace:    image_test.go:129
        Error:          Received unexpected error "Another task is already in progress."
```

```
=== RUN   TestCreateFS
time="2016-07-28T18:33:36Z" level=debug msg="Trying to log in with username/password in lieu of cert" 
time="2016-07-28T18:33:36Z" level=debug msg="Session Environment Info: " API Type=HostAgent API Version=6.0 Build=3634798 Name="VMware ESXi" OS Type=vmnix-x86 Product ID=embeddedEsx UUID= Vendor="VMware, Inc." Version=6.0.0 
time="2016-07-28T18:33:36Z" level=debug msg="vSphere resource cache populating..." 
time="2016-07-28T18:33:36Z" level=debug msg="Cached dc: " 
time="2016-07-28T18:33:36Z" level=debug msg="Cached cluster: " 
time="2016-07-28T18:33:36Z" level=debug msg="Cached ds: datastore1" 
time="2016-07-28T18:33:36Z" level=debug msg="Cached host: " 
[...]

Issue #1560, opened 2016-07-19
VIC does not support creating a volume with auto-generated name


mhagen-vmware:
Actual behavior:
Error response from daemon: Server error from Portlayer: Name in body is required

Expected behavior:
anon@anon-Latitude-E7450:~/vic$ docker volume create 
d3a0986406196224a06f4592c3ef730f97646a0cc920c31d9656ee16b3ecfd20

matthewavery:
This was closed in a previous PR but the ticket was missed in the actual pr. pr #2058 is reenabling the test for this case which shows that we do in fact support uuid random names of volumes now.

matthewavery:
This has been resolved.

stuclem:
This is fixed. Removing kind/note.

PR #2058, opened 2016-08-19
Wait and Retry implementation for TaskInProgressFaults during calls to vsphere

### LINKED ISSUE #2013: Error response from daemon: [POST /storage/volumes/][500] createVolumeInternalServerError  &{Code:0xc82000f700 Message:Another task is already in progress.}
Saw this in one of the latest CI:
[integration_logs_4076_fb4a743b76b95b99029f483c596efe1638984ce3_1471381988.zip](https://github.com/vmware/vic/files/421311/integration_logs_4076_fb4a743b76b95b99029f483c596efe1638984ce3_1471381988.zip)

matthewavery:
Fixes #2013 #1725 #1896 #1560 

This PR will be a simple implementation of a wait and retry to solve issues where we get a `TaskInProgressFault` from vsphere when trying to attempt multiple "long running" tasks. Currently we have only seen this on `attach` and `detach` requests for disk devices to the same vm(our appliance). 

Additionally the test for a randomly named volume has been fixed up to reflect that the issue for that has been addressed in the past. this has also been noted in #1560

### DIFF
--- a/lib/apiservers/engine/backends/volume.go
+++ b/lib/apiservers/engine/backends/volume.go
@@ -35,7 +35,6 @@ const (
 	OptsVolumeStoreKey     string = "VolumeStore"
 	OptsCapacityKey        string = "Capacity"
 	dockerMetadataModelKey string = "DockerMetaData"
-	bytesToMegabyte               = int64(1000000)
 )
 
 func NewVolumeModel(volume *models.VolumeResponse, labels map[string]string) *types.Volume {
@@ -216,8 +215,7 @@ func validateDriverArgs(args map[string]string, model *models.VolumeRequest) err
 	if err != nil {
 		return err
 	}
-	model.Capacity = int64(capacity) / bytesToMegabyte
-
+	model.Capacity = int64(capacity) / int64(units.MB)
 	return nil
 }
 
--- a/pkg/vsphere/disk/disk_manager.go
+++ b/pkg/vsphere/disk/disk_manager.go
@@ -223,10 +223,18 @@ func (m *Manager) Create(ctx context.Context, newDiskURI string,
 //	return nil
 // }
 
-func (m *Manager) Attach(ctx context.Context, spec *types.VirtualDisk) error {
-	machineSpec := configureDeviceSpec(ctx, *m.vm, types.VirtualDeviceConfigSpecOperationAdd, types.VirtualDeviceConfigSpecFileOperationCreate, spec)
+func (m *Manager) Attach(ctx context.Context, disk *types.VirtualDisk) error {
+
+	deviceList := object.VirtualDeviceList{}
+	deviceList = append(deviceList, disk)
 
 	_, err := tasks.Retry(ctx, func(ctx context.Context) (tasks.ResultWaiter, error) {
+		changeSpec, err := deviceList.ConfigSpec(types.VirtualDeviceConfigSpecOperationAdd)
+		if err != nil {
+			return nil, err
+		}
+		machineSpec := types.VirtualMachineConfigSpec{}
+		machineSpec.DeviceChange = append(machineSpec.DeviceChange, changeSpec...)
 		return m.vm.Reconfigure(ctx, machineSpec)
 	})
 
@@ -295,38 +303,4 @@ func (m *Manager) devicePathByURI(ctx context.Context, datastoreURI string) (str
 	}
 
 	return fmt.Sprintf(m.byPathFormat, *disk.UnitNumber), nil
-}
-
-//Utility Functions
-
-// configureDeviceSpec is a function that provides the spec for our new volume. We do this because we need to be able to retry on the actual reconfigure task. Whic is hidden in govmomi calls unless we call the reconfigure operation  ourselves.
-func configureDeviceSpec(ctx context.Context, v object.VirtualMachine, op types.VirtualDeviceConfigSpecOperation, fop types.VirtualDeviceConfigSpecFileOperation, devices ...types.BaseVirtualDevice) types.VirtualMachineConfigSpec {
-	spec := types.VirtualMachineConfigSpec{}
-
-	for _, device := range devices {
-		config := &types.VirtualDeviceConfigSpec{
-			Device:    device,
-			Operation: op,
-		}
-
-		if disk, ok := device.(*types.VirtualDisk); ok {
-			config.FileOperation = fop
-
-			// Special case to attach an existing disk
-			if op == types.VirtualDeviceConfigSpecOperationAdd && disk.CapacityInKB == 0 {
-				childDisk := false
-				if b, ok := disk.Backing.(*types.VirtualDiskFlatVer2BackingInfo); ok {
-					childDisk = b.Parent != nil
-				}
-
-				if !childDisk {
-					config.FileOperation = "" // existing disk
-				}
-			}
-		}
-
-		spec.DeviceChange = append(spec.DeviceChange, config)
-	}
-
-	return spec
 }
--- a/pkg/vsphere/disk/disk_manager.go
+++ b/pkg/vsphere/disk/disk_manager.go
@@ -224,7 +224,6 @@ func (m *Manager) Create(ctx context.Context, newDiskURI string,
 // }
 
 func (m *Manager) Attach(ctx context.Context, disk *types.VirtualDisk) error {
-
 	deviceList := object.VirtualDeviceList{}
 	deviceList = append(deviceList, disk)
 
--- a/pkg/vsphere/disk/disk_manager.go
+++ b/pkg/vsphere/disk/disk_manager.go
@@ -228,13 +228,15 @@ func (m *Manager) Attach(ctx context.Context, disk *types.VirtualDisk) error {
 	deviceList := object.VirtualDeviceList{}
 	deviceList = append(deviceList, disk)
 
-	_, err := tasks.Retry(ctx, func(ctx context.Context) (tasks.ResultWaiter, error) {
-		changeSpec, err := deviceList.ConfigSpec(types.VirtualDeviceConfigSpecOperationAdd)
-		if err != nil {
-			return nil, err
-		}
-		machineSpec := types.VirtualMachineConfigSpec{}
-		machineSpec.DeviceChange = append(machineSpec.DeviceChange, changeSpec...)
+	changeSpec, err := deviceList.ConfigSpec(types.VirtualDeviceConfigSpecOperationAdd)
+	if err != nil {
+		return err
+	}
+
+	machineSpec := types.VirtualMachineConfigSpec{}
+	machineSpec.DeviceChange = append(machineSpec.DeviceChange, changeSpec...)
+
+	_, err = tasks.Retry(ctx, func(ctx context.Context) (tasks.ResultWaiter, error) {
 		return m.vm.Reconfigure(ctx, machineSpec)
 	})
 
--- a/pkg/vsphere/disk/disk_manager.go
+++ b/pkg/vsphere/disk/disk_manager.go
@@ -226,7 +226,7 @@ func (m *Manager) Create(ctx context.Context, newDiskURI string,
 func (m *Manager) Attach(ctx context.Context, spec *types.VirtualDisk) error {
 	machineSpec := configureDeviceSpec(ctx, *m.vm, types.VirtualDeviceConfigSpecOperationAdd, types.VirtualDeviceConfigSpecFileOperationCreate, spec)
 
-	_, err := tasks.WaitAndRetryForResult(ctx, func(ctx context.Context) (tasks.ResultWaiter, error) {
+	_, err := tasks.Retry(ctx, func(ctx context.Context) (tasks.ResultWaiter, error) {
 		return m.vm.Reconfigure(ctx, machineSpec)
 	})
 
@@ -269,7 +269,7 @@ func (m *Manager) Detach(ctx context.Context, d *VirtualDisk) error {
 
 	spec.DeviceChange = config
 
-	_, err = tasks.WaitAndRetryForResult(ctx, func(ctx context.Context) (tasks.ResultWaiter, error) {
+	_, err = tasks.Retry(ctx, func(ctx context.Context) (tasks.ResultWaiter, error) {
 		return m.vm.Reconfigure(ctx, spec)
 	})
 	if err != nil {
--- a/pkg/vsphere/disk/disk_manager.go
+++ b/pkg/vsphere/disk/disk_manager.go
@@ -92,8 +92,8 @@ func (m *Manager) CreateAndAttach(ctx context.Context, newDiskURI,
 
 	log.Infof("Create/attach vmdk %s from parent %s", newDiskURI, parentURI)
 
-	if err := m.vm.AddDevice(ctx, spec); err != nil {
-		log.Errorf("vmdk storage driver failed to attach disk: %s", errors.ErrorStack(err))
+	err = m.Attach(ctx, spec)
+	if err != nil {
 		return nil, errors.Trace(err)
 	}
 
@@ -223,6 +223,20 @@ func (m *Manager) Create(ctx context.Context, newDiskURI string,
 //	return nil
 // }
 
+func (m *Manager) Attach(ctx context.Context, spec *types.VirtualDisk) error {
+	machineSpec := configureDeviceSpec(ctx, *m.vm, types.VirtualDeviceConfigSpecOperationAdd, types.VirtualDeviceConfigSpecFileOperationCreate, spec)
+
+	_, err := tasks.WaitAndRetryForResult(ctx, func(ctx context.Context) (tasks.ResultWaiter, error) {
+		return m.vm.Reconfigure(ctx, machineSpec)
+	})
+
+	if err != nil {
+		log.Errorf("vmdk storage driver failed to attach disk: %s", errors.ErrorStack(err))
+		return errors.Trace(err)
+	}
+	return nil
+}
+
 func (m *Manager) Detach(ctx context.Context, d *VirtualDisk) error {
 	defer trace.End(trace.Begin(d.DevicePath))
 	log.Infof("Detaching disk %s", d.DevicePath)
@@ -255,7 +269,7 @@ func (m *Manager) Detach(ctx context.Context, d *VirtualDisk) error {
 
 	spec.DeviceChange = config
 
-	err = tasks.Wait(ctx, func(ctx context.Context) (tasks.Waiter, error) {
+	_, err = tasks.WaitAndRetryForResult(ctx, func(ctx context.Context) (tasks.ResultWaiter, error) {
 		return m.vm.Reconfigure(ctx, spec)
 	})
 	if err != nil {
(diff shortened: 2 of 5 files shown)


pkg/vsphere/disk/disk_manager.go
```diff
@@ -223,6 +223,20 @@ func (m *Manager) Create(ctx context.Context, newDiskURI string,
 //	return nil
 // }
 
+func (m *Manager) Attach(ctx context.Context, spec *types.VirtualDisk) error {
+	machineSpec := configureDeviceSpec(ctx, *m.vm, types.VirtualDeviceConfigSpecOperationAdd, types.VirtualDeviceConfigSpecFileOperationCreate, spec)
+
+	_, err := tasks.WaitAndRetryForResult(ctx, func(ctx context.Context) (tasks.ResultWaiter, error) {
```
fdawg4l: I would rather integration of wait/retry happen in a different PR.  These storage changes need to be looked at after we know wait/retry is usable;  not at the same time.

fdawg4l: Also, this happens at the wrong layer. But, lets chat about that in a different PR which is solely integration of this helper.

pkg/vsphere/tasks/waiter.go
```diff
@@ -82,3 +87,29 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
 	}
 	return info, nil
 }
+
+func WaitAndRetryForResult(ctx context.Context, f func(context.Context) (ResultWaiter, error)) (*types.TaskInfo, error) {
+	rand.NewSource(time.Now().UnixNano()) //creates a more unique random
```
fdawg4l: rand?  What not exponential backoff?

matthewavery: I previously had an exponential backoff type solution. but in reality we should not need exponential back off here. That might even be overkill since this is not a truly permanent solution.

pkg/vsphere/tasks/waiter.go
```diff
@@ -82,3 +87,29 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
 	}
 	return info, nil
 }
+
+func WaitAndRetryForResult(ctx context.Context, f func(context.Context) (ResultWaiter, error)) (*types.TaskInfo, error) {
```
fdawg4l: Since this function takes a ctx, it should abide by the context interface and cancel when the `context.CancelFunc` is called.  Just my opinion.

pkg/vsphere/disk/disk_manager.go
matthewavery: The PR itself is out against a p0 issue blocking volume ls tests. Which means that the best way to test and make for sure that function is working as expected is to see if the fault we are seeing is now blocking the tests.

matthewavery: The storage changes are actually just sections of code that are somewhat extracted from govmomi since the reconfigure and the spec creation needed to be done separately in ordered to perform the WaitAndRetry on the actual Reconfigure. We needed the exposed Task object which was hidden in the old call path.

pkg/vsphere/tasks/waiter.go
fdawg4l: Why isn't a permanent solution?  These retry-able errors are normal in distributed systems.   

Also I'm not getting why rand is better than expo backoff.

pkg/vsphere/disk/disk_manager.go
fdawg4l: This is not the only path which would trigger the p0 volume error.  This is why I'd rather the PRs be separate.

fdawg4l: If reconfigure is the operation which needs to retry, then shouldnt the vsphere pkg in vic implement the reconfigure operation (if it doesnt already) and use the retry there?

matthewavery: That is a good point, we should probably do that instead.

matthewavery: You are right, the reconfigure on a volume create should also be guarded by this... hmmm, in this case creating a packaged reconfigure option is best. I will make that refactor after I confirm some of the other issues.

pkg/vsphere/tasks/waiter.go
matthewavery: meaning do not have a max retry count?

pkg/vsphere/tasks/waiter.go
hmahmood: Its 50-50 in my book. Both are OK approaches here.

pkg/vsphere/tasks/waiter.go
matthewavery: Heeded your advice :) there is now no longer a collision limit. Only an infinite loop and a possible CDE. we should note though, that if no deadline is set we now have an infinite loop.

matthewavery:
cc @dougm

pkg/vsphere/disk/disk_manager.go
```diff
@@ -282,3 +296,37 @@ func (m *Manager) devicePathByURI(ctx context.Context, datastoreURI string) (str
 
 	return fmt.Sprintf(m.byPathFormat, *disk.UnitNumber), nil
 }
+
+//Utility Functions
+
+// configureDeviceSpec is a function that provides the spec for our new volume. We do this because we need to be able to retry on the actual reconfigure task. Whic is hidden in govmomi calls unless we call the reconfigure operation  ourselves.
+func configureDeviceSpec(ctx context.Context, v object.VirtualMachine, op types.VirtualDeviceConfigSpecOperation, fop types.VirtualDeviceConfigSpecFileOperation, devices ...types.BaseVirtualDevice) types.VirtualMachineConfigSpec {
```
hmahmood: Don't like copying this from govmomi, but I think of any other way of retrying `AddDevice`.

dougm: Looks like you should be able to use VirtualDeviceList.ConfigSpec() instead.  Not sure why VirtualMachine.configureDevice() isn't using VirtualDeviceList.ConfigSpec() too, might just be a missed cleanup - I'll look into it.

matthewavery: Cool thanks! I definitely did not want to to copy that part of the code

fdawg4l:
NACK

Lets chat offline, but this is happening at the wrong layer.  Don't check this in, please.

pkg/vsphere/disk/disk_manager.go
```diff
@@ -223,6 +223,28 @@ func (m *Manager) Create(ctx context.Context, newDiskURI string,
 //	return nil
 // }
 
+func (m *Manager) Attach(ctx context.Context, disk *types.VirtualDisk) error {
+
+	deviceList := object.VirtualDeviceList{}
+	deviceList = append(deviceList, disk)
+
+	_, err := tasks.Retry(ctx, func(ctx context.Context) (tasks.ResultWaiter, error) {
+		changeSpec, err := deviceList.ConfigSpec(types.VirtualDeviceConfigSpecOperationAdd)
```
hmahmood: This needs to be outside of the `Retry`, but otherwise LGTM

fdawg4l:
Ok, after Doug reading the code to me and @matthewavery patiently describing what's happening, I'm better now.  I take back my NACK.

Would really rather this retry call be wrapped around the reconfigure call in the vm package that we have in pkg.  We can leave this here for now, but i do think implementing the ReconfigureWithRetry (or whatever you feel like calling it) is the right place to call the new task function.  But, fine as-is.

LGTM

pkg/vsphere/tasks/waiter.go
```diff
@@ -82,3 +89,49 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
...
+		<-timer.C
+	}
+
+	for {
+		taskInfo, err = WaitForResult(ctx, f)
+		if err == nil && taskInfo.Error == nil {
+			break
+		}
+
+		if _, ok := taskInfo.Error.Fault.(types.TaskInProgressFault); !ok {
```
dougm: If err != nil, then taskInfo will be nil and would result in a panic here.  if err != nil, you should just return it after the WaitForResult call.

tests/test-cases/Group1-Docker-Commands/1-19-Docker-Volume-Create.robot
```diff
@@ -2,16 +2,12 @@
 Documentation  Test 1-19 - Docker Volume Create
 Resource  ../../resources/Util.robot
 Suite Setup  Install VIC Appliance To Test Server
-Suite Teardown  Cleanup VIC Appliance On Test Server
+suite Teardown  Cleanup VIC Appliance On Test Server
```
dougm: why 'S' -> 's' here?

pkg/vsphere/disk/disk_manager.go
```diff
@@ -223,6 +223,30 @@ func (m *Manager) Create(ctx context.Context, newDiskURI string,
 //	return nil
 // }
 
+func (m *Manager) Attach(ctx context.Context, disk *types.VirtualDisk) error {
+
+	deviceList := object.VirtualDeviceList{}
+	deviceList = append(deviceList, disk)
+
+	changeSpec, err := deviceList.ConfigSpec(types.VirtualDeviceConfigSpecOperationAdd)
```
dougm: :+1:

tests/test-cases/Group1-Docker-Commands/1-19-Docker-Volume-Create.robot
matthewavery: oh good catch, this should not be different

matthewavery:
.

matthewavery:
LGTM, carrying @hmahmood's lgtm out of code comments

pkg/vsphere/tasks/waiter.go
```diff
@@ -82,3 +89,54 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
...
+		if err != nil {
+			return nil, err
+		}
+
+		if taskInfo.Error == nil {
+			break
+		}
+
+		if _, ok := taskInfo.Error.Fault.(types.TaskInProgressFault); !ok {
+			break
```
dougm: Looks like the 2 callers of Retry() ignore the returned TaskInfo and only check the returned error.  So this Error would be swallowed.  At the very least, this taskInfo.Error should be logged here.  Returning the error in this case or having Retry() callers check taskInfo.Error would be better.

pkg/vsphere/tasks/waiter.go
```diff
@@ -82,3 +89,54 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
...
+
+	//setting up timer
+	timer := time.NewTimer(time.Duration(100) * time.Millisecond)
+	if !timer.Stop() {
+		<-timer.C
+	}
+
+	for {
+		taskInfo, err = WaitForResult(ctx, f)
+
```
dougm: cosmetic: remove the newline here (in general, no newline between func call and 'if err != nil' check)

pkg/vsphere/disk/disk_manager.go
```diff
@@ -223,6 +223,30 @@ func (m *Manager) Create(ctx context.Context, newDiskURI string,
 //	return nil
 // }
 
+func (m *Manager) Attach(ctx context.Context, disk *types.VirtualDisk) error {
+
```
dougm: cosmetic: remove this empty line

pkg/vsphere/tasks/waiter.go
matthewavery: what should happen if we find a task error from the retry in attach and detach? since it will be a type `Fault` should an error with that fault object be returned back?

pkg/vsphere/tasks/waiter.go
matthewavery: 👍

pkg/vsphere/tasks/waiter.go
```diff
@@ -82,3 +89,54 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
 	}
 	return info, nil
 }
+
+func Retry(ctx context.Context, f func(context.Context) (ResultWaiter, error)) (*types.TaskInfo, error) {
+	rand.NewSource(time.Now().UnixNano()) //creates a more unique random
```
dougm: You're not using the return value of NewSource() here, so this should be removed.

matthewavery: ah yeah, I need a `rand.New` to wrap that

matthewavery: alright changed this so I am actually using a custom random object. I am surprised at my self for how bad that was.

pkg/vsphere/tasks/waiter.go
dougm: Any non-nil error would do.  Leave it to the caller to check taskInfo.Error if they care about a specific type.
```go
return taskInfo, errors.New(taskInfo.Error.LocalizedMessage)
```

pkg/vsphere/tasks/waiter.go
dougm: I suggested removing the line also because I don't see why we'd need to do that here.

lib/apiservers/engine/backends/volume.go
```diff
@@ -216,8 +215,7 @@ func validateDriverArgs(args map[string]string, model *models.VolumeRequest) err
 	if err != nil {
 		return err
 	}
-	model.Capacity = int64(capacity) / bytesToMegabyte
-
+	model.Capacity = int64(capacity) / int64(units.MB)
```
hickeng: This looks like it's for a different PR?

pkg/vsphere/tasks/waiter.go
```diff
@@ -82,3 +89,54 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
 	}
 	return info, nil
 }
+
+func Retry(ctx context.Context, f func(context.Context) (ResultWaiter, error)) (*types.TaskInfo, error) {
```
hickeng: Worth some function comments to describe the semantics of this call. I'd be inclined to leave it called Wait as well, given we're not retrying on any errors, only the single case where there's an in-flight collision.

pkg/vsphere/tasks/waiter.go
```diff
@@ -82,3 +89,54 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
...
+			break
+		}
+
+		if _, ok := taskInfo.Error.Fault.(types.TaskInProgressFault); !ok {
+			break
+		}
+		sleepValue := backoffFactor * (rand.Int63n(100) + int64(50))
+		timer.Reset(time.Duration(sleepValue) * time.Millisecond)
+		select {
+		case <-timer.C:
```
hickeng: This can be simplified to the following with no need to manage a timer:
```
case <- time.After(sleepValue * time.Millisecond)
```

pkg/vsphere/tasks/waiter.go
hickeng: concur with Doug. This isn't crypto so having slightly less deterministic rand shouldn't be an issue.

pkg/vsphere/tasks/waiter.go
matthewavery: There is already a function above called wait. I figured I would just keep that as `Retry` so that the option to cancel a task if there is an in flight collision is there.

pkg/vsphere/tasks/waiter.go
matthewavery: turned out much simpler thanks. I have already learned a lot about the time package now. The only thing I have not used to implement this is a `ticket`

pkg/vsphere/tasks/waiter.go
matthewavery: If we find we want this in the future a new ticket should be opened. There have been other comments surrounding this function as well. However, right now other functions also use the wait functions that already existed and still operate under certain expectations potentially.

hickeng: In which case we need a follow up exercise to move all calls to Wait to Retry, having checked they have an appropriate context configured to provide cancellation.

pkg/vsphere/tasks/waiter.go
```diff
@@ -82,3 +89,51 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
...
 }
+
+func Retry(ctx context.Context, f func(context.Context) (ResultWaiter, error)) (*types.TaskInfo, error) {
+	r := rand.New(rand.NewSource(time.Now().UnixNano())) //creates a more unique random
+	var err error
+	var taskInfo *types.TaskInfo
+	backoffFactor := int64(1)
+
+	//setting up timer
+	timer := time.NewTimer(time.Duration(100) * time.Millisecond)
```
hickeng: This doesn't need to be here any more.

pkg/vsphere/tasks/waiter.go
```diff
@@ -82,3 +89,51 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
...
+	}
+
+	for {
+		taskInfo, err = WaitForResult(ctx, f)
+		if err != nil {
+			return nil, err
+		}
+
+		if taskInfo.Error == nil {
+			break
```
hickeng: why not `return taskInfo, nil` inline here, the way we do for the clause above?

pkg/vsphere/tasks/waiter.go
matthewavery: right, Do you want to make that ticket or should I?

pkg/vsphere/tasks/waiter.go
matthewavery: hmm that was somehow leftover... that is odd. it is not on my branch maybe it got cherry picked out by accident.

matthewavery: yes looks like that was the case

pkg/vsphere/tasks/waiter.go
```diff
@@ -82,3 +89,51 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
...
+		}
+
+		if _, ok := taskInfo.Error.Fault.(types.TaskInProgressFault); !ok {
+			log.Debugf("Task failed during a retry operation : %#v", taskInfo.Task)
+			log.Debugf("Failed TaskInfo Object : %#v", taskInfo)
+			return taskInfo, errors.New(taskInfo.Error.LocalizedMessage)
+		}
```
hickeng: can we have a newline after this `if` block to separate an exit path from the retry logic

pkg/vsphere/tasks/waiter.go
```diff
@@ -26,6 +29,10 @@ import (
 	"github.com/vmware/vic/pkg/errors"
 )
 
+const (
+	maxBackoffFactor = int64(8)
```
hickeng: If I read this correctly we max out at a retry of 1sec (150ms * 8)  - I'd let this grow to at least 3sec given potential completion times in busy environments.

pkg/vsphere/tasks/waiter.go
```diff
@@ -82,3 +89,51 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
...
+		sleepValue := time.Duration(backoffFactor * (r.Int63n(100) + int64(50)))
+		select {
+		case <-time.After(sleepValue * time.Millisecond):
+			if backoffFactor*2 > maxBackoffFactor {
+				backoffFactor = maxBackoffFactor
+			} else {
+				backoffFactor *= 2
+			}
+		case <-ctx.Done():
+			return nil, ctx.Err()
```
hickeng: Log that we've stopped retrying the task due to context cancellation.

pkg/vsphere/tasks/waiter.go
```diff
@@ -82,3 +89,51 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
...
+		}
+		log.Infof("Retrying Task due to TaskInProgressFault: %s", taskInfo.Task.Reference())
+	}
+
+	if err != nil {
```
hickeng: This clause cannot be hit

pkg/vsphere/tasks/waiter.go
matthewavery: alright, I  have increased it to 16. which has a max of 2400 but that is only after several failures.

hickeng:
LGTM

matthewavery:
♻️

PR #2061, opened 2016-08-19
[WIP] [DEBUG] [CI] Enable integration tests for retry PR

caglar10ur:

PR #2063, opened 2016-08-19
Return os.ErrExist when making a directory that already exists

### LINKED ISSUE #2053: docker pull fails with error 500 from port layer
Get this error in the test suite for several tests for PR #2049 (see attached logs for build [4170](https://ci.vmware.run/vmware/vic/4170)):

```
time="2016-08-19T00:04:22Z" level=debug msg="Creating a store from input 564de3f6-7c77-c6d6-b0a1-d54cc62eedcf" 
time="2016-08-19T00:04:22Z" level=debug msg="Creating a store failed: [POST /storage][500] CreateImageStore default  &{Code:0xc420234380 Message:ServerFaultCode: Cannot complete the operation because the file or folder [datastore1] VCH-4170-9224/images/VIC/564de3f6-7c77-c6d6-b0a1-d54cc62eedcf/images already exists}" 
time="2016-08-19T00:04:22Z" level=fatal msg="Failed to create image store: [POST /storage][500] CreateImageStore default  &{Code:0xc420234380 Message:ServerFaultCode: Cannot complete the operation because the file or folder [datastore1] VCH-4170-9224/images/VIC/564de3f6-7c77-c6d6-b0a1-d54cc62eedcf/images already exists}" 
```

[integration_logs_4170_7c25b65f0beb59332358ab785fabeba1b3c8bd75_1471566347.zip](https://github.com/vmware/vic/files/426175/integration_logs_4170_7c25b65f0beb59332358ab785fabeba1b3c8bd75_1471566347.zip)

fdawg4l:
Looks like the docker image cache and a docker client are both attempting to create an image store at the same time.  Guard against this.

Fixes #2053

### DIFF
--- a/pkg/vsphere/datastore/datastore.go
+++ b/pkg/vsphere/datastore/datastore.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"io"
 	"net/url"
+	"os"
 	"path"
 	"regexp"
 	"strings"
@@ -79,7 +80,10 @@ func NewHelper(ctx context.Context, s *session.Session, ds *object.Datastore, ro
 	if len(root) > 1 {
 		r, err := d.Mkdir(ctx, true, root[1])
 		if err != nil {
-			return nil, err
+			if !os.IsExist(err) {
+				return nil, err
+			}
+			log.Debugf("%s already exists", d.RootURL)
 		}
 		d.RootURL = r
 	}
@@ -127,23 +131,23 @@ func (d *Helper) Summary(ctx context.Context) (*types.DatastoreSummary, error) {
 func (d *Helper) Mkdir(ctx context.Context, createParentDirectories bool, dirs ...string) (string, error) {
 
 	upth := path.Join(dirs...)
-
-	// check if it already exists
-	_, err := d.Ls(ctx, upth)
-	if err != nil && !types.IsFileNotFound(err) {
-		return "", err
-	}
-
 	upth = path.Join(d.RootURL, upth)
-
-	// dir already exists
-	if err == nil {
-		return upth, nil
-	}
 
 	log.Infof("Creating directory %s", upth)
 
 	if err := d.fm.MakeDirectory(ctx, upth, d.s.Datacenter, createParentDirectories); err != nil {
+
+		log.Debugf("Creating %s error: %s", upth, err)
+
+		if err != nil {
+			if soap.IsSoapFault(err) {
+				soapFault := soap.ToSoapFault(err)
+				if _, ok := soapFault.VimFault().(types.FileAlreadyExists); ok {
+					return "", os.ErrExist
+				}
+			}
+		}
+
 		return "", err
 	}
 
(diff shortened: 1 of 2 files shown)


caglar10ur:
LGTM :sheep: :it:

lib/portlayer/storage/image_cache.go
```diff
@@ -113,14 +113,14 @@ func (c *NameLookupCache) CreateImageStore(ctx context.Context, storeName string
 		return nil, os.ErrExist
 	}
 
+	c.storeCacheLock.Lock()
+	defer c.storeCacheLock.Unlock()
+
 	u, err = c.DataStore.CreateImageStore(ctx, storeName)
```
hmahmood: Is CreateImageStore idempotent? If not, you can still get into situation where two callers can end up calling this function (if they are both waiting on the lock). You would need to move the lock further up to the beginning of the function to avoid this.

hickeng:
Don't think this is the right way to go - @hmahmood is correct about the lock placement but @fdawg4l notes that GetImageStore also obtains that lock.

I would contend that CreateImageStore is an initialization time operation and shouldn't be called for every op.

hmahmood:
Having GetImageStore acquire the lock does not solve the problem either. The Get and Create have to be called while holding onto the lock, without releasing in the intermediate.

I do agree with @hickeng about the initialization. If this is a one time thing, I think this is an attempt at lazily creating the store, but maybe an explicit (early) initialization is called for here.

fdawg4l:
Agree with all.  Will rearrange hydration of the cache vs creation of the store and solve the locking hole.

fdawg4l:
Ah, now i remember.  We load the cache for a given image store on `Create` because we don't know what image stores exist on startup.  We can crawl around the directory tree and load all available stores.  I'll add that.

fdawg4l:
Ok, so the crux of this issue is that imagec was `Fatalf`'ing on a vsphere error.  The vsphere error was because `CreateImageStore` resultes in an `mkdir` on the datastore if it does not exist.  `imagec` *will* ignore already exists errors from `CreateImageStore` iff the error asserts to `os.ErrExists` (indirectly by way of the REST API).  Our call to `mkdir` didn't do that.  I'm taking care of this in my next push to this PR.

The rest of this (cache cleanup, locking, etc) are all red-herrings.  `mkdir` is atomic.  It's all we need.

fdawg4l:
> I would contend that CreateImageStore is an initialization time operation and shouldn't be called for every op.

@hickeng now that we have vic-machine giving us some config at startup, we can have the docker daemon call this when it starts up.  Won't happen in this PR, but I'll create an issue to track it.

hmahmood:
lgtm

hickeng:
This is currently blocking #2054 / #2061

hickeng:
LGTM - we should look at including this helper code into govmomi longer term as the differing semantics of vSAN vs vmfs are not something govmomi users should really need to care about.

PR #2082, opened 2016-08-22
Don't provide redundant 'images' default for image store path

### LINKED ISSUE #2081: Remove redundant "images" directory from path when vic-machine -i only includes a datastore
`./vic-machine-linux create ... -i datastore1`

**Expected behavior:**
`./VIC/564d9a86-068d-dfad-823e-0d0a2b4c76df/images/scratch/scratch.vmdk`

**Actual behavior:**
**./images**`/VIC/564d9a86-068d-dfad-823e-0d0a2b4c76df/images/scratch/scratch.vmdk`

gigawhitlocks:
Initially I provided a default path for images ("images") in `validator.go` but this winds up being redundant, so an empty path is actually fine.

Fixes #2081


emlin:
lgtm

lib/install/validate/storage.go
```diff
@@ -41,11 +40,6 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
 		return
 	}
 
-	// provide a default path if only a DS name is provided
-	if imageDSpath.Path == "" {
```
fdawg4l: So we used to do `path.Join( display name, "images" )`.  After this change, is the path just "display name"?  Because I would've expected line 46 to just be `imageDSpath.Path = input.DisplayName`

fdawg4l: To clarify-  we just want to remove the first "images" directory from the path.

gigawhitlocks: After this, path adds nothing special, so we get something like (in vsphere format) `[datastore] /VIC/portlayer-provided-UUID/images/imageName` -- the path as far as the validator is concerned is empty.

gigawhitlocks: No wait, I'm wrong. We should be keeping the display name as the default. Backing this PR out.

PR #2084, opened 2016-08-22
Use ctx timeout in portlayer init

### LINKED ISSUE #2059: ensureComponentsInitialize is not using ctx
it should either be longer than 10sec or needs to honor timeout option

anchal-agrawal:
Changes logic in  `vic-machine`'s `ensureComponentsInitialize` to use the context timeout.

Fixes #2059

### DIFF
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -617,6 +617,7 @@ func (d *Dispatcher) ensureComponentsInitialize(conf *config.VirtualContainerHos
 		client *http.Client
 		res    *http.Response
 		err    error
+		req    *http.Request
 	)
 
 	if conf.HostCertificate.IsNil() {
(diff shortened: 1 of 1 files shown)


lib/install/management/appliance.go
```diff
@@ -634,18 +635,38 @@ func (d *Dispatcher) ensureComponentsInitialize(conf *config.VirtualContainerHos
...
+	req, err = http.NewRequest("GET", dockerInfoURL, nil)
+	if err != nil {
+		return errors.New("invalid HTTP request for docker info")
+	}
+	req = req.WithContext(d.ctx)
+
+	timer := time.NewTimer(time.Second)
+	if !timer.Stop() {
+		<-timer.C
+	}
```
caglar10ur: ticker := time.NewTicker(time.Second)

caglar10ur: [Hit send too early]

that way you don't need  reset

anchal-agrawal: Fixed, thank you!

lib/install/management/appliance.go
```diff
@@ -634,18 +635,31 @@ func (d *Dispatcher) ensureComponentsInitialize(conf *config.VirtualContainerHos
...
+	for {
+		res, err = client.Do(req)
 		if err == nil && res.StatusCode == http.StatusOK {
 			if isPortLayerRunning(res) {
-				return nil
+				break
 			}
 		}
-		time.Sleep(time.Second)
+
```
emlin: I think you need to check if err is not nil. If timeout, it should return error, I assume.
Did you test timeout in your env?

emlin: I see, you checked the ctx.Done. Never mind.

lib/install/management/appliance.go
```diff
@@ -634,18 +635,31 @@ func (d *Dispatcher) ensureComponentsInitialize(conf *config.VirtualContainerHos
...
+	}
+	req = req.WithContext(d.ctx)
+
+	ticker := time.NewTicker(time.Second)
+	for {
+		res, err = client.Do(req)
 		if err == nil && res.StatusCode == http.StatusOK {
 			if isPortLayerRunning(res) {
-				return nil
+				break
```
hmahmood: Need to call `ticker.Stop` here, other LGTM

caglar10ur: or call it via defer and remove the call from below :)

caglar10ur:
LGTM after you fix the ticker leak

lib/install/management/appliance.go
anchal-agrawal: Yep, Hasan suggested that offline. Fix incoming.

emlin: So basically here, we cannot query out error message from docker personality API. 
If there is anything wrong to initialize those services, user can only wait until timeout, without useful error message.

emlin:
LGTM. 👍 
Just wondering if it's possible to get out more useful information while real error happens.

anchal-agrawal:
Ping

Issue #253, opened 2016-03-15
Handle deletes of images


fdawg4l:
On a standard docker based setup, deleting of images is protected by the graph driver which has a representation of which images are in use (either directly or via inheritance) by containers.  For instance

```
scratch -> layer1 -> layer2 -> layer3 -> RW_layer -> container
                          \--> layer2' -> layer3' 
```

Since `container` is live, none of the layers in its hierarchy can be deleted.  However, `layer2'` and `layer3'` are good to be reclaimed.

VIC doesn't have a graph driver.  In fact, all we have is a list of VMDK disks and their immediate parents.  We could walk the list of running VMs, introspect their attached disks, identify their hierarchies, then select the set of VMDKs which are not related and allow those to be deleted for _every delete image request_ (since we're stateless and dont want to do any runtime accounting with transaction semantics).

Alternatively, we could mark each leaf image we create as a leaf.  Then when we go to delete a given image, we simply check if the given VMDK is marked and proceed if it's not marked.  We would then have to unmark each image when they become parents of images or are used (via RW_layer inheritance) by a VM.  

This avoids needing to construct a graph driver in memory and maintains the stateless nature of the port layer.  Also, since we cannot rely on ESX to give us this level of introspection, maintains correctness.  

CC: @hickeng

hickeng:
This is not part of `vic-machine delete` - that is being accomplished with zero knowledge of the disk structures by deleting the entire image store.

PR #2085, opened 2016-08-23
Implement an index to track tree nodes

fdawg4l:
This index will replace the map in the storage layer used to cache images.  By replacing the map with this index, we'll be able to support `rmi`.

### DIFF
--- a/pkg/index/index.go
+++ b/pkg/index/index.go
@@ -103,6 +103,19 @@ func (i *Index) Get(nodeId string) (Node, error) {
 	return n.Copy(), nil
 }
 
+func (i *Index) List() ([]Node, error) {
+	i.l.Lock()
+	defer i.l.Unlock()
+
+	nodes := make([]Node, 0, len(i.lookupTable))
+
+	for _, v := range i.lookupTable {
+		nodes = append(nodes, v.Copy())
+	}
+
+	return nodes, nil
+}
+
 // Delete deletes a leaf node
 func (i *Index) Delete(nodeId string) (Node, error) {
 	i.l.Lock()
--- a/pkg/index/index.go
+++ b/pkg/index/index.go
@@ -1,0 +1,192 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package index
+
+import (
+	"container/list"
+	"fmt"
+	"sync"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+type Node interface {
+	// Returns the string representation (usually the ID) of the node
+	String() string
+
+	// Deep copy of the node
+	Copy() Node
+}
+
+type node struct {
+	Node
+	parent   *node
+	children []*node
+	mask     uint32
+}
+
+func (n *node) addChild(child *node) {
+	n.children = append(n.children, child)
+}
+
+type Index struct {
+	root        *node
+	lookupTable map[string]*node
+	l           sync.Mutex
+}
+
+func NewIndex() *Index {
+	i := &Index{
+		lookupTable: make(map[string]*node),
+	}
+
+	return i
+}
+
+// Insert inserts a copy of the given node to the tree under the given parent.
+func (i *Index) Insert(parent string, n Node) error {
+	i.l.Lock()
+	defer i.l.Unlock()
+
+	_, ok := i.lookupTable[n.String()]
+	if ok {
+		return fmt.Errorf("node %s already exists in index", n.String())
+	}
+
+	newNode := &node{
+		Node: n.Copy(),
+	}
+
+	if parent == n.String() {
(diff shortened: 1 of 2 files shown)


fdawg4l:
Implements needed functionality for #253

pkg/index/index.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func (n *node) addChild(child *node) {
+	n.children = append(n.children, child)
+}
+
+type Index struct {
+	root        *node
+	lookupTable map[string]*node
+	l           sync.Mutex
```
emlin: is it better to use sync.RWMutex?

pkg/index/index.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil, fmt.Errorf("Node %s not found", nodeId)
+	}
+
+	if len(n.children) != 0 {
+		return nil, fmt.Errorf("Node %s has children %#q", nodeId, n.children)
+	}
+
+	// remove the reference to the node from its parent
+	var deleted bool
+	err := i.bfsworker(i.root, func(needle *node) (iterflag, error) {
```
emlin: confused here. Why not use node.parent here to find out it's parent?

pkg/index/index.go
fdawg4l: There are no cases where we need to atomically read the index.  If we find we'll need it in the future, we can change this.

pkg/index/index.go
fdawg4l: Good point!!  You're right!  I don't need to bfs walk the tree to remove a node in that case.  Will fix this and update it.  Thank you!!

fdawg4l:
LGTM

fdawg4l:
.

pkg/index/index.go
```diff
@@ -0,0 +1,206 @@
...
+	// Returns the string respresentation of the nodes parent (usually it's ID)
+	Parent() string
+
+	// Deep copy of the node
+	Copy() Node
+}
+
+type node struct {
+	Node
+	parent   *node
```
caglar10ur: this upper case lower case Node vs node look really confusing. Can you name the interface with something else? Preferably something that ends up with er like the other interfaces (Stringer, Reader, Writer etc.)

pkg/index/index.go
caglar10ur: Can we use m for the mutex instead of l in the name of readability? i.l v.s i.m :)

pkg/index/index.go
```diff
@@ -0,0 +1,206 @@
...
+	lookupTable map[string]*node
+	l           sync.Mutex
+}
+
+func NewIndex() *Index {
+	i := &Index{
+		lookupTable: make(map[string]*node),
+	}
+
+	return i
```
caglar10ur: there is no need the intermediate variable, you could directly return without assigning to i

pkg/index/index.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	if n.Parent() == n.String() {
+		if i.root != nil {
+			return fmt.Errorf("node cannot point to self unless it's root")
+		}
+
+		// set root
+		i.root = newNode
+
```
caglar10ur: extra space :)

pkg/index/index.go
```diff
@@ -0,0 +1,206 @@
...
+		return fmt.Errorf("node %s already exists in index", n.String())
+	}
+
+	log.Debugf("Inserting %s (parent: %s) in index", n.String(), n.Parent())
+
+	newNode := &node{
+		Node: n.Copy(),
+	}
+
+	if n.Parent() == n.String() {
```
caglar10ur: For a second I thought Parent is returning the Node and was trying to understand why we are comparing Parent with String. How about naming the String as Self or something else to make it clear?

caglar10ur: Or maybe Parent returning Node and then calling n.Parent().String()?

pkg/index/index.go
```diff
@@ -0,0 +1,206 @@
...
+	i.l.Lock()
+	defer i.l.Unlock()
+
+	// XXX Look into parallelizing this without breaking API boundaries.
+	return i.bfsworker(root, func(n *node) (iterflag, error) { return visitFunc(n.Node) })
+}
+
+func (i *Index) bfsworker(root *node, visitFunc func(*node) (iterflag, error)) error {
+
+	queue := list.New()
```
caglar10ur: now that I see we use list, how about using it for addChild as well? That way you don't need to call append (hence no realloc and copy)

caglar10ur:
Mostly cosmetic or readability comments, otherwise LGTM

emlin:
lgtm 👍

pkg/index/index.go
fdawg4l: Thats a good point.  Will rename to something better.  `Noder` sounds too much like neuter.  Maybe `Elementer` since the package and struct is called `Index`.

pkg/index/index.go
fdawg4l: Parent can't return the `Node` because our `Node` interface is used to keep string references.  The `Index` uses these to build the tree using actual pointers ( to`node`s).

fdawg4l: Will rename String() to Self().  I had originally wanted anything that satisfied the stringer interface to be able to be added to the index, but the parent relationships broke all that.  Self makes more sense.

pkg/index/index.go
fdawg4l: I tried using a `List` here, and this definitely works, but the extra pointer (because `list.Element` has a `Prev`,`Next`, and Value pointer) is an extra 16bytes per element.  This vs the 1 reference (8byte) for each element in the array.  I'd rather take the realloc overhead of copying vs keeping more references in memory.

fdawg4l:
.

PR #2048, opened 2016-08-18
vic-ui allows HTTPS plugin URL

andrewtchin:
Fixes #2045

PR #2094, opened 2016-08-23
Improve UI installer

jooskim:
This PR improves the UI installer in that:

1. It does not ask whether or not the VC version is 5.5 when installing VIC UI through the standard plugin deployment procedure (using a web server) The reason for checking the VC version is that 5.5 has a different folder structure for storing Web Client plugin contents than in 6.0+, and the user doesn't need to care about this when deploying the plugin with the standard method.
2. The flag used for specifying a server thumbprint flag in VIC UI registration is changed from --serverThumbprint to --server-thumbprint. (See #2048)
3. Removed an unnecessary check on whether the plugin extension was registered successfully with VC

### DIFF
--- a/ui/installer/VCSA/install.sh
+++ b/ui/installer/VCSA/install.sh
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash
 # Copyright 2016 VMware, Inc. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


andrewtchin:
lgtm

jooskim:
Paging @stuclem to make sure documentation is consistent with this change. Could you review what I left in the comment above excluding the last one?

stuclem:
@jooskim I'll make the doc update once this is merged. Doc update is tracked in https://github.com/vmware/vic/issues/2136.

jooskim:
LGTM. Can't wait too much on this, as UI integration tests need to be implemented.

Issue #2069, opened 2016-08-21
Remove call to CreateImageStore from docker image cache and imagec


fdawg4l:
From #2053

`CreateImageStore` should be called from the docker personality on startup and not on every interaction with the storage layer involving images.  Right now, `imagec` and the image cache call this function (mostly due to legacy). 

I don't understand why the image cache is calling this at all.  If you guys can clarify, I can remove this first.

hickeng:
This is currently blocking #2054 / #2061

jzt:
I should be able to refactor the image store creation step out of the cache `Update`.

fdawg4l:
Rawk.  Thanks @jzt!

fdawg4l:
Looks like we're still racing on the image cache creation logic.  Can you guys remove the create call from the image cache asap?  #2092 seems to manifest this.

jzt:
PR should be up momentarily, just waiting on these tests to finish!

PR #2097, opened 2016-08-23
Move image store creation to docker persona startup

### LINKED ISSUE #2092: Failed to obtain list of images: [GET /storage/{store_name}][404] listImagesNotFound
https://ci.vmware.run/vmware/vic/4271

[integration_logs_4271_a331e5a85fe9269e8f3b306535442f15990fb500_1471962085.zip](https://github.com/vmware/vic/files/432662/integration_logs_4271_a331e5a85fe9269e8f3b306535442f15990fb500_1471962085.zip)

jzt:
This removes the image store creation step from imagec and the image cache's `Update()` and places it at the persona startup. It will retry 5 times for a total of 10 seconds, and will move on if the store already exists (in the case of a persona restart, for example). The image store must be created before the persona can start serving requests.

Paging @fdawg4l, @mhagen-vmware 

Fixes #2092, #2069

### DIFF
--- a/lib/apiservers/engine/backends/cache/image_cache.go
+++ b/lib/apiservers/engine/backends/cache/image_cache.go
@@ -34,7 +34,6 @@ import (
 
 	"github.com/vmware/vic/lib/apiservers/portlayer/client"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/storage"
-	"github.com/vmware/vic/lib/apiservers/portlayer/models"
 	"github.com/vmware/vic/lib/metadata"
 	"github.com/vmware/vic/pkg/vsphere/sys"
 )
@@ -71,7 +70,7 @@ func ImageCache() *ICache {
 
 // Update runs only once at startup to hydrate the image cache
 func (ic *ICache) Update(client *client.PortLayer) error {
-	log.Debugf("Updating image cache...")
+	log.Debugf("Updating image cache")
 
 	host, err := sys.UUID()
 	if host == "" {
@@ -79,21 +78,6 @@ func (ic *ICache) Update(client *client.PortLayer) error {
 	}
 	if err != nil {
 		return fmt.Errorf("Unexpected error getting hostname: %s", err)
-	}
-
-	// attempt to create the image store if it doesn't exist
-	store := &models.ImageStore{Name: host}
-	_, err = client.Storage.CreateImageStore(
-		storage.NewCreateImageStoreParamsWithContext(ctx).WithBody(store),
-	)
-
-	if err != nil {
-		if _, ok := err.(*storage.CreateImageStoreConflict); ok {
-			log.Debugf("Store already exists")
-		} else {
-			log.Debugf("Creating a store failed: %#v", err)
-			return err
-		}
 	}
 
 	params := storage.NewListImagesParamsWithContext(ctx).WithStoreName(host)
@@ -105,7 +89,7 @@ func (ic *ICache) Update(client *client.PortLayer) error {
 
 	for _, layer := range layers.Payload {
 		imageConfig := &metadata.ImageConfig{}
-		if err := json.Unmarshal([]byte(layer.Metadata["metaData"]), imageConfig); err != nil {
+		if err := json.Unmarshal([]byte(layer.Metadata[metadata.MetaDataKey]), imageConfig); err != nil {
 			derr.NewErrorWithStatusCode(fmt.Errorf("Failed to unmarshal image config: %s", err),
 				http.StatusInternalServerError)
 		}
--- a/lib/apiservers/engine/backends/vicbackends.go
+++ b/lib/apiservers/engine/backends/vicbackends.go
@@ -15,6 +15,7 @@
 package vicbackends
 
 import (
+	"context"
 	"net"
 	"time"
 
@@ -22,16 +23,17 @@ import (
 	httptransport "github.com/go-swagger/go-swagger/httpkit/client"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client"
+	"github.com/vmware/vic/lib/apiservers/portlayer/client/misc"
+	"github.com/vmware/vic/lib/apiservers/portlayer/client/storage"
+	"github.com/vmware/vic/lib/apiservers/portlayer/models"
 	"github.com/vmware/vic/lib/config"
+	"github.com/vmware/vic/pkg/vsphere/sys"
 )
 
 const (
 	Imagec             = "imagec"
 	PortlayerName      = "Backend Engine"
 	IndexServerAddress = "registry-1.docker.io"
-
-	// Retries defines how many times to attempt refreshing the image cache at startup
-	Retries = 5
 
 	// RetryTimeSeconds defines how many seconds to wait between retries
 	RetryTimeSeconds = 2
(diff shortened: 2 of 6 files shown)


mhagen-vmware:
I don't think we want another arbitrary 10 second deadline again do we? Is there any other way to avoid that? If we have a loaded system, will it consistently exceed that deadline like we doing in other places as well?

jzt:
If the appropriate place to create the image store is indeed in the persona during startup, then there's no way around having the persona wait on the portlayer to come up and start serving before it can actually do that. If the persona starts serving before the store is created, that's just moving the race condition somewhere else.

Whether or not the image store should be created by way of the persona (instead of imagec) is another question entirely, that should probably be answered by @fdawg4l.

Perhaps we could reconfigure starting our services to allow for serially ordered bringup and configurable timeouts, then maybe the CI system can just wait forever? That would probably cause the VCH to take longer to come up.

fdawg4l:
@jzt, you're right-  moving the creation code from imagec to the persona or the persona to imagec without knowing whether the PL is up, just moves the condition and doesn't solve it.  So, the problem of knowing whether the PL is up should be done elsewhere.  Lets assume in this change that it is up.  

We can create an issue and add a change to implement a barrier in the docker daemon on startup which waits for connectivity to the PL before coming up.  (I thought one already exists by way of a REST `Ping()` but it shouldn't be difficult to implement if it doesnt exist.)

lib/apiservers/engine/backends/vicbackends.go
```diff
@@ -71,8 +76,13 @@ func Init(portLayerAddr, product string, config *config.VirtualContainerHostConf
 	portLayerClient = client.New(t, nil)
 	portLayerServerAddr = portLayerAddr
 
+	log.Info("Creating image store")
+	if err := createImageStore(); err == nil {
```
fdawg4l: If you move this inside the `for` loop on line 87, then you won't need to do the loop in `createImageStore()`

lib/apiservers/engine/backends/vicbackends.go
```diff
@@ -114,3 +124,40 @@ func ProductVersion() string {
...
 	return vchConfig
 }
+
+func createImageStore() error {
+
+	// TODO(jzt): we should move this to a utility package or something
+	host, err := sys.UUID()
+	if host == "" {
+		host, err = os.Hostname()
+	}
```
fdawg4l: This condition should never happen.  It's safe to remove 132:134 and just check for and return err if not nil.

fdawg4l: ^^^ unless you run the persona on a non nested env.  I don't know how you prefer to develop.  Your call if you want to support running this daemon on a non-nested setup.

lib/apiservers/engine/backends/vicbackends.go
```diff
@@ -71,8 +76,13 @@ func Init(portLayerAddr, product string, config *config.VirtualContainerHostConf
...
 
+	log.Info("Creating image store")
+	if err := createImageStore(); err == nil {
+		return err
+	}
+
 	// attempt to update the image cache at startup
-	log.Info("Refreshing image cache...")
+	log.Info("Refreshing image cache")
 	go func() {
```
fdawg4l: While we're here, this go routine could use some TLC.  First, lets do this synchronously and not in a go routine if it can be avoided.  Second, instead of a loop over retries, lets use[Ticker](https://golang.org/pkg/time/#Ticker).  How about something like https://play.golang.org/p/WsJ9EI8M7l.  Points if you can get rid of the `var` declaration.

fdawg4l: PS.  This loop is just a workaround since we dont have that barrier mentioned earlier.

lib/apiservers/engine/backends/vicbackends.go
jzt: I do run nested, but I believe this pattern came from imagec - you may be right, it may be deprecated. Perhaps it was so that imagec could run in standalone mode (in which case, we should probably still keep it).

lib/apiservers/engine/backends/vicbackends.go
jzt: Yeah, its in a goroutine because @hickeng did not want a cache update to block the persona from serving requests. If the PL has a large number of images, it could easily block a docker client from doing stuff while it updates.  Your thoughts?

I'll switch to using a `Ticker` in the meantime.

lib/apiservers/engine/backends/vicbackends.go
jzt: That loop is running inside a goroutine and we don't want to run in a goroutine because the race condition will move (assuming we don't ditch the goroutine). Also, it seems like it would be kinda clunky to have to account for retrying image store creation or cache updating or both within the same loop, given that one could succeed while the other could fail. Your thoughts?

jzt:
@fdawg4l I went ahead and gated everything on a ping to the portlayer that actually does the retry. The image store is created synchronously, after which the persona begins serving requests and the cache starts updating.

I ditched the `Ticker`, though. The retry code is simpler and prettier as a result, I think.

jzt:
@mhagen-vmware does this work better for you?

jzt:
Ok, I've removed the timeout altogether as this operation is already wrapped in a timeout by vic-machine.

hickeng:
@jzt this isn't actually wrapped in a timeout via vic-machine. vic-machine ceases waiting, but that has zero impact on the appliance. That said, I'm fine with this model - this is supposed to be structured as a distributed system that just happens to be co-located for now.

Ideally we could respond to requests that don't require us to hit the port layer (which would be _version_ and _login_ iirc) but no need to go that far for now.

LGTM

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -71,7 +70,7 @@ func ImageCache() *ICache {
 
 // Update runs only once at startup to hydrate the image cache
 func (ic *ICache) Update(client *client.PortLayer) error {
-	log.Debugf("Updating image cache...")
+	log.Debugf("Updating image cache")
```
fdawg4l: Minor and feel free to ignore.  This could probably be made an `Infof` message.   Up to you.

lib/apiservers/engine/backends/vicbackends.go
```diff
@@ -114,3 +118,40 @@ func ProductVersion() string {
 func VchConfig() *config.VirtualContainerHostConfigSpec {
 	return vchConfig
 }
+
+func pingPortLayer() {
+	params := misc.NewPingParamsWithContext(context.TODO())
+	ticker := time.NewTicker(RetryTimeSeconds * time.Second)
+	defer ticker.Stop()
```
fdawg4l: You may want to add an `Infof` message here that we're waiting for the PL to come up.  Otherwise the log will be empty and we won't know what the deamon is doing.

fdawg4l:
Nice work @jzt!

LGTM.  Thanks for taking care of this!

lib/apiservers/engine/backends/cache/image_cache.go
jzt: There's actually an `Infof` before the call to this guy over in the persona init.

lib/apiservers/engine/backends/vicbackends.go
jzt: Fixed

jzt:
♻️

PR #2098, opened 2016-08-23
Populate VCH status on VIC admin page

### LINKED ISSUE #1229: Add "healthy" setup checks into vic-admin web page.

kjplatz:
Fixes #1229

This populates the "VCH Status" health check on the VICadmin page by verifying that "vic-init" and all proceses defined in the VirtualContainerHostConfigSpec.ExecutorConfig are currently running.  If any processes are not running, the green check changes to an orange warning triangle, and a warning message indicating that that particular process is not running is published on the vicadmin page.

### DIFF
--- a/lib/tether/tether.go
+++ b/lib/tether/tether.go
@@ -18,9 +18,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"os/signal"
+	"path"
 	"syscall"
 	"time"
 
@@ -131,6 +133,15 @@ func (t *tether) setup() error {
 			log.Errorf("Failed to start extension %s: %s", name, err)
 			return err
 		}
+	}
+
+	// Create PID file for tether
+	tname := path.Base(os.Args[0])
+	err = ioutil.WriteFile(fmt.Sprintf("%s.pid", path.Join(PIDFileDir, tname)),
+		[]byte(fmt.Sprintf("%d", os.Getpid())),
+		0644)
+	if err != nil {
+		log.Errorf("Unable to open PID file for %s : %s", os.Args[0], err)
 	}
 
 	return nil
@@ -305,6 +316,10 @@ func (t *tether) handleSessionExit(session *SessionConfig) {
 		logs = logs[logCount-MaxDeathRecords+1:]
 	}
 
+	// Remove associated PID file
+	cmdname := path.Base(session.Cmd.Path)
+	_ = os.Remove(fmt.Sprintf("%s.pid", path.Join(PIDFileDir, cmdname)))
+
 	session.Diagnostics.ExitLogs = append(logs, executor.ExitLog{
 		Time:       time.Now(),
 		ExitStatus: session.ExitStatus,
@@ -406,6 +421,14 @@ func (t *tether) launch(session *SessionConfig) error {
 
 	// Set the Started key to "true" - this indicates a successful launch
 	session.Started = "true"
+	// Write the PID to the associated PID file
+	cmdname := path.Base(session.Cmd.Path)
+	err = ioutil.WriteFile(fmt.Sprintf("%s.pid", path.Join(PIDFileDir, cmdname)),
+		[]byte(fmt.Sprintf("%d", session.Cmd.Process.Pid)),
+		0644)
+	if err != nil {
+		log.Errorf("Unable to write PID file for %s: %s", cmdname, err)
+	}
 	log.Debugf("Launched command with pid %d", session.Cmd.Process.Pid)
 
 	return nil
--- a/lib/vicadmin/validate.go
+++ b/lib/vicadmin/validate.go
@@ -17,8 +17,10 @@ package vicadmin
 import (
 	"fmt"
 	"html/template"
+	"io/ioutil"
 	"net"
 	"os"
+	"path"
 	"sort"
 	"strings"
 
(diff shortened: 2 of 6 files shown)


lib/tether/tether.go
```diff
@@ -133,6 +136,15 @@ func (t *tether) setup() error {
 		}
 	}
 
+	// Create PID file for tether
+	paths := strings.Split(os.Args[0], "/")
+	err = ioutil.WriteFile(fmt.Sprintf("%s/%s.pid", PIDFileDir, paths[len(paths)-1]),
```
hmahmood: There is no need to create a pid file for the tether. You should only create pid files for the processes the tether launches.

lib/tether/tether.go
```diff
@@ -39,6 +41,7 @@ const (
 
 	// the length of a truncated ID for use as hostname
 	shortLen = 12
+	PIDFileDir = "/var/run"
```
hmahmood: This should be in tether_linux.go .

lib/tether/tether.go
```diff
@@ -305,6 +317,10 @@ func (t *tether) handleSessionExit(session *SessionConfig) {
 		logs = logs[logCount-MaxDeathRecords+1:]
 	}
 
+	// Remove associated PID file
+	paths := strings.Split(session.Cmd.Path, "/")
+	_ = os.Remove(fmt.Sprintf("%s/%s.pid", PIDFileDir, paths[len(paths)-1]))
```
hmahmood: If there is no `/` in the path, paths will be empty and `len(paths)-1` will be `-1`.

lib/tether/tether.go
```diff
@@ -305,6 +317,10 @@ func (t *tether) handleSessionExit(session *SessionConfig) {
 		logs = logs[logCount-MaxDeathRecords+1:]
 	}
 
+	// Remove associated PID file
+	paths := strings.Split(session.Cmd.Path, "/")
```
hmahmood: You can use [path.Base](https://golang.org/pkg/path/#Base)

lib/tether/tether.go
hmahmood: ... and [path.Join](https://golang.org/pkg/path/#Join) here

lib/vicadmin/validate.go
```diff
@@ -191,3 +202,55 @@ func (v *Validator) QueryDatastore(ctx context.Context, vch *config.VirtualConta
...
+
+	procs := []string { "vic-init" }
+
+	// Extract required components from vchConfig
+	// Only report on components with Restart set to true
+	for _, sess := range vch.ExecutorConfig.Sessions {
+		if !sess.Restart {
+			continue
+		}
+		parsedcmd := basename.FindStringSubmatch(sess.Cmd.Path)
```
hmahmood: Maybe you need to write a helper function that return the pid file path given a session object. That way you can consistently use it everywhere in the tether and vidadmin.

hmahmood: A pidfile type/package perhaps?

lib/vicadmin/validate.go
```diff
@@ -191,3 +202,55 @@ func (v *Validator) QueryDatastore(ctx context.Context, vch *config.VirtualConta
...
+		// Field 3 is the current state as reported by the kernel
+		switch fields[2][0] {
+		// We're good
+		case 'R', 'S', 'D': break
+		// Process has been killed, is dying, or a zombie
+		case 'K', 'X', 'x', 'Z':
+			v.VCHIssues = template.HTML(fmt.Sprintf("%s<span class=\"error-message\">Process %s has failed</span>\n",
+				v.VCHIssues, err))
+		}
+		log.Infof("Process %s running as PID %s", proc, pid)
```
hmahmood: We are going to fall through here in either case from above; is this message accurate?

lib/vicadmin/validate.go
```diff
@@ -191,3 +202,55 @@ func (v *Validator) QueryDatastore(ctx context.Context, vch *config.VirtualConta
...
+		}
+
+		fields := strings.Split(string(status), " ")
+		// Field 3 is the current state as reported by the kernel
+		switch fields[2][0] {
+		// We're good
+		case 'R', 'S', 'D': break
+		// Process has been killed, is dying, or a zombie
+		case 'K', 'X', 'x', 'Z':
+			v.VCHIssues = template.HTML(fmt.Sprintf("%s<span class=\"error-message\">Process %s has failed</span>\n",
```
hmahmood: If this is going to be displayed to the user, I would use the name of the service whose process if failing.

lib/tether/tether.go
kjplatz: @hmahmood Per our discussion, we may be moving vicadmin to be its own service not spawned by the tether.  In that event, it does make sense to create a PID file for the tether.

hmahmood:
lgtm

matthewavery:
lgtm

kjplatz:
Bing

PR #2105, opened 2016-08-23
Docker login [full ci]

### LINKED ISSUE #368: Docker API: Docker Registry Authenticate
Implement the docker registry authentication operation in the engine-api server.  Right now, imagec does all of our image pulling.  I would assume this auth has some TTL attribute.

https://docs.docker.com/engine/reference/commandline/login/
https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#check-auth-configuration

Please note, this functionality exist in the CLI but not the API.  For all we know, this might be handled in the docker CLI.  If that is the case, there might not be any work for us to do.

sflxn:
Implements docker login

- Support insecure registries
- Support docker login
- Fixed docker pull
- Fixed imagec's OAuth2 workflow

Resolves #368


matthewavery:
is it possible for you to put the vendored code in a separate pr? It is a little difficult to see the exact changes

sflxn:
@matthewavery Just click on the specific commit to only see the changes.  You will not see the vendored code.

matthewavery:
nvm, with the squash I can view the commit. Thanks!

lib/apiservers/engine/backends/vicbackends.go
```diff
@@ -45,9 +47,12 @@ var (
 	productVersion      string
 
 	vchConfig *config.VirtualContainerHostConfigSpec
+
+	insecureRegistries []string
+	RegistryService    *registry.Service
 )
 
-func Init(portLayerAddr, product string, config *config.VirtualContainerHostConfigSpec) error {
+func Init(portLayerAddr, product string, config *config.VirtualContainerHostConfigSpec, registries []url.URL) error {
```
hickeng: These registries are used as insecure registries - call the arg that to make it explicit

lib/apiservers/engine/backends/image.go
```diff
@@ -112,6 +112,7 @@ func (i *Image) PullImage(ctx context.Context, ref reference.Named, metaHeaders
 	defer trace.End(trace.Begin("PullImage"))
 
 	log.Debugf("PullImage: ref = %+v, metaheaders = %+v\n", ref, metaHeaders)
+	log.Printf("* PullImage - authConfig = %#v", authConfig)
```
hickeng: This should be Debugf, and only if authConfig doesn't contain sensitive data such as username/password/auth token

lib/apiservers/engine/backends/system.go
```diff
@@ -240,7 +241,8 @@ func (s *System) UnsubscribeFromEvents(chan interface{}) {
 }
 
 func (s *System) AuthenticateToRegistry(ctx context.Context, authConfig *types.AuthConfig) (string, string, error) {
-	return "", "", fmt.Errorf("%s does not implement System.AuthenticateToRegistry", ProductName())
+	defer trace.End(trace.Begin(""))
+	return RegistryService.Auth(authConfig, dockerversion.DockerUserAgent(ctx))
```
hickeng: This is a simple approach, but as @fdawg4l says it pulls in a lot of docker/docker. As best I can tell the docker/docker/registry/... is mainly to abstract between v1 and v2. Given we don't support v2 can we use docker/distribution/registry/client directly?

cmd/docker/main.go
```diff
@@ -76,7 +78,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := vicbackends.Init(cli.portLayerAddr, productName, &vchConfig); err != nil {
+	if err := vicbackends.Init(cli.portLayerAddr, productName, &vchConfig, cli.insecureRegistries); err != nil {
```
sflxn: It's not getting pulled from the command line.  cli is just a structure we use for command line flags, but the insecure registry flag is pulled from the vchconfig.

lib/apiservers/engine/backends/image.go
sflxn: Removing.  That was left there by mistake.

lib/apiservers/engine/backends/system.go
sflxn: I can look into reducing the footprint.  We can also abstract the authen/authoriz code from imagec out and implement our own.  I used the docker code as the most expedient path to implementation for the upcoming deliverable.  Let me write some robot scripts to test the login/pull workflow, and then I can revisit this.

caglar10ur:
Can you please fix the conflicts with the master? I'm seeing commits that's coming from other authors

lib/apiservers/engine/backends/system.go
sflxn: @hickeng I looked into using the docker/distribution/registry/client directly.  There isn't a login call.  The registry client in distribution provides a rountripper to add to an http client that will handle the authentication and authorization, but it doesn't actually provide a function that does the authentication.  The docker/docker/registry/auth is effectively the login client.  If we don't want to import docker/docker and import docker/distribution only for login, we would basically duplicate the loginV2() function in docker/docker/registry/auth.go.

RegistryService.Auth() that we use in our code doesn't just auth and differentiate between V1 or V2.  It handles the login option of ip:port or cidr format.  It also handles the insecure registry and does some work with a mirror registry option.

If we want to avoid importing code, we'll have to reimplement the same code.  If all we want to do is to avoid V1, we can set V2Only to true when creating the RegistryService.

Lastly, we can also import some of the imagec code and modify it to do what we need.

At this point, I think we should leave everything as is, with the exception of disabling the V1 registries.

hickeng: We don't want or need CIDR format at this point - we're specifying registries via fqdn or IP.
Mirrors are irrelevant from a login perspective and likely a security issue given you then need to trust all of the mirrors.

Longer term we're going to want the following services in a VCH:
a. port layer API exposed to provide basic container ops
b. registry service that can, given an image tag, obtain the image and pass it to the port layer
c. personality services on top of a & b

We pretty much have to go with whatever is expedient at this point given the bundle timelines, but we do need to remember that this isn't supposed to be a direct re-implementation of the docker daemon function.

sflxn: Ok, I'll remove the WIP tag.  Re-implementing the docker daemon was never the point.  Given the short period to do this, it was the most expedient method.  Most of this time was spent on discovery -- how does login works, how does it affect pull, how to setup harbor (secure, and insecure), how does this work with docker.io.  The work on the personality server is a little different from the rest of VIC in that there is always some time spent on discovery of how docker actually works, and as we all see, their documentation doesn't fully detail everything in one place.

mdubya66:
We should change the title of this to [full ci] to get some runs on this change.

cmd/docker/main.go
```diff
@@ -126,6 +126,17 @@ func handleFlags() (*CliOptions, bool) {
 		}
 	}
 
+	// load the vch config
+	src, err := extraconfig.GuestInfoSource()
+	if err != nil {
+		log.Errorf("Unable to load configuration from guestinfo")
```
dougm: I realize this was existing code, but shouldn't we include the original error in that log message and use log.Fatalf to exit(1) rather than continue?

cmd/imagec/docker.go
```diff
@@ -158,8 +158,10 @@ func FetchToken(url *url.URL) (*Token, error) {
 	})
 	tokenFileName, err := fetcher.Fetch(url)
 	if err != nil {
+		log.Debugf("*** FetchToken FAILED")
```
dougm: instead: return nil, log.Formatf("FetchToken (%s) failed: %s", url, err)

cmd/imagec/docker.go
```diff
@@ -158,8 +158,10 @@ func FetchToken(url *url.URL) (*Token, error) {
...
 		return nil, err
 	}
+	log.Debugf("*** FetchToken SUCCEEDED")
```
dougm: Lets remove that or add some more context:
```
log.Debugf("FetchToken (%s) succeeded", url)
```

dougm:
A few minor comments otherwise lgtm

cmd/imagec/docker.go
sflxn: sounds good.

cmd/imagec/docker.go
sflxn: Just removed the line.  Success isn't important to log.

cmd/docker/main.go
sflxn: sounds good.

tests/test-cases/Group1-Docker-Commands/1-27-Docker-Login.robot
```diff
@@ -0,0 +1,26 @@
...
+Resource  ../../resources/Util.robot
+Suite Setup  Install VIC Appliance To Test Server  ${false}
+Suite Teardown  Cleanup VIC Appliance On Test Server
+
+*** Variables ***
+${yml}  version: "2"\nservices:\n${SPACE}web:\n${SPACE}${SPACE}image: python:2.7\n${SPACE}${SPACE}ports:\n${SPACE}${SPACE}- "5000:5000"\n${SPACE}${SPACE}volumes:\n${SPACE}${SPACE}- ".:/code"\n${SPACE}${SPACE}depends_on:\n${SPACE}${SPACE}- redis\n${SPACE}redis:\n${SPACE}${SPACE}image: redis\n${SPACE}${SPACE}ports:\n${SPACE}${SPACE}- "5001:5001"
+
+*** Test Cases ***
+Docker login and pull from docker.io
+    ${status}=  Get State Of Github Issue  368
```
hickeng: These 3 lines blocking this test need to be removed.

hickeng:
LGTM once test is re-enabled

mdubya66:
lgtm per slack conversation with @hickeng

PR #2109, opened 2016-08-23
Initial event stream for portlayer

cgtexmex:
Change is focused on exec and will update the container cache. container-list
has also been refactored to take advantage of an up to date container cache.

This change does not encompass network and will not release network resources
for containers acted on out of band.  That will be addressed in a future
change.

Fixes: #1329


mdubya66:
LGTM

hmahmood:
lgtm

PR #2110, opened 2016-08-24
Check if target is managed by VC

### LINKED ISSUE #1998: Running vic-machine ls on a host that is managed by VC returns an empty list
When you run `vic-machine ls` and specify an ESXi host in the `target ` option, if that ESXi host is managed by vCenter Server, `vic-machine ls` returns an empty list. This happens even if there are VCHs running on that ESXi host in VC. This is potentially misleading.

If possible, `vic-machine ls` should return an error stating that the target host is managed by VC XXX, as is the case when you try to run `vic-machine create` on an ESXi host that is managed by VC.

emlin:
Fixes #1998

### DIFF
--- a/lib/install/validate/validator.go
+++ b/lib/install/validate/validator.go
@@ -17,12 +17,12 @@ package validate
 import (
 	"crypto/tls"
 	"fmt"
-	"net"
 	"net/url"
 	"strings"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/vmware/govmomi/find"
+	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/lib/install/data"
 	"github.com/vmware/vic/pkg/errors"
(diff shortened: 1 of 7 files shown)


hmahmood:
lgtm

lib/install/validate/validator.go
```diff
@@ -280,19 +289,40 @@ func (v *Validator) target(ctx context.Context, input *data.Data, conf *config.V
...
+
+	var mh mo.HostSystem
+
+	if err = host.Properties(ctx, host.Reference(), []string{"summary.managementServerIp"}, &mh); err != nil {
+		v.NoteIssue(fmt.Errorf("Failed to get host properties: %s", err))
+		return
+	}
+
+	if ip := mh.Summary.ManagementServerIp; ip != "" {
+		v.NoteIssue(fmt.Errorf("Target is managed by vCenter server %q, please change --target to vCenter or standalone ESXi address", ip))
```
andrewtchin: maybe `Target is managed by vCenter server %q, please change --target to the vCenter address or select a standalone ESXi`

andrewtchin:
lgtm

lib/install/validate/validator.go
emlin: nice, thanks

emlin:
♻️

PR #2121, opened 2016-08-24
Work around github by just querying drone itself, add full ci support [full ci]

mhagen-vmware:
Finally (I think) came up with at least a work around solution until we upgrade to drone .5

### DIFF
--- a/tests/integration-test.sh
+++ b/tests/integration-test.sh
@@ -20,11 +20,9 @@ gsutil version -l
 
 dpkg -l > package.list
 
-echo $DRONE_BUILD_URL
-echo $DRONE_BUILD_DIR
-echo $DRONE_PULL_REQUEST
-
 if [ $DRONE_BRANCH = "master" ] && [ $DRONE_REPO = "vmware/vic" ]; then
+    pybot --removekeywords TAG:secret tests/test-cases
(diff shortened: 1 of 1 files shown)


tests/integration-test.sh
```diff
@@ -20,12 +20,10 @@ gsutil version -l
 
 dpkg -l > package.list
 
-echo $DRONE_BUILD_URL
-echo $DRONE_BUILD_DIR
-echo $DRONE_PULL_REQUEST
-
 if [ $DRONE_BRANCH = "master" ] && [ $DRONE_REPO = "vmware/vic" ]; then
     pybot --removekeywords TAG:secret tests/test-cases
+elif grep -q "\[full ci\]" <(drone build info vmware/vic $DRONE_BUILD_NUMBER); then
```
hickeng: let's be consistent with `[full-ci]` given it's `[skip-ci]`

mhagen-vmware: It is actually not skip-ci... it is skip ci.  Loc continues to use skip-ci and it doesn't work, per:
https://ci.vmware.run/vmware/vic/4344

hickeng:
Okay, LGTM then

kjplatz:
:+1:

PR #2122, opened 2016-08-24
Do not call info-set guestinfo.ip if ip is empty

caglar10ur:


dougm:
lgtm

pkg/vsphere/toolbox/service.go
```diff
@@ -223,6 +223,10 @@ func (s *Service) SetOption(args []byte) ([]byte, error) {
 	case "broadcastIP": // TODO: const-ify
 		if val == "1" {
 			ip := s.PrimaryIP()
+			if ip == "" {
```
hickeng: what is the value we get back in `netInterfaceAddrs()` in the DefaultIP path - do we need a corresponding code change there as well?

caglar10ur: I don't think so, we just end up interface with no ip

pkg/vsphere/toolbox/service_test.go
```diff
@@ -185,7 +185,7 @@ func TestServiceErrors(t *testing.T) {
 
 	in.rpc = []*testRPC{
 		{"Capabilities_Register", "OK "},
-		{"Set_Option broadcastIP 1", "ERR "},
+		{"Set_Option broadcastIP 1", "OK "},
```
dougm: We want to keep this to cover that error path.  I think to cover the new empty ip check/return path, would need add something different that sets service.PrimaryIP to a func that returns empty.

fdawg4l:
I don't really understand this code but it looks fine.

LGTM

pkg/vsphere/toolbox/service.go
```diff
@@ -223,6 +223,10 @@ func (s *Service) SetOption(args []byte) ([]byte, error) {
...
+				log.Printf("failed to find primary IP\n")
```
hmahmood: You should probably make this a warning.

caglar10ur: That's is golang's log package so we don't have levels there. Not sure whether @dougm prefers it over logrus.

dougm: Right, vsphere/toolbox package currently uses stdlib log.. if we have a good reason to switch it to logsrus, let's do that in a separate PR that changes the rest of the logging too.

mdubya66:
Open for 1 month, is it still valid?

caglar10ur:
Happened again so here is the rebased version with test changes

PR #2125, opened 2016-08-24
Allow volume stores which already exist to be added

fdawg4l:
This is a fix for a regression.  We should be able to add volume stores with existing volumes.


mdubya66:
Do we need a regression test? Is there no bug filed for the regression it is fixing?

fdawg4l:
@mdubya66 there is no regression test yet.  We also don't support VCH restart because of an open issue.  We can't test this without supporting VCH restart.

On the plus side, there are unit tests which tickle this.  That's how I found it.  I'll talk to @mhagen-vmware about getting CI to run them since they require a VM with access to the esx box its on.

hmahmood:
lgtm

caglar10ur:
LGTM :sheep: :it:

fdawg4l:
.

Issue #2000, opened 2016-08-16
Ensure volumes are writeable when creating VIC appliance


kjplatz:
When attempting to create a VIC appliance, it will quietly succeed when assigning a volume store that the VCH does not have write permissions to.  For example, using the following command will succeed in creating a volume on datastore2:bing (assuming the directory bing exists on datastore2):

```
bin/vic-machine-linux create --user=root --password='password' --force=true --compute-resource=/ha-datacenter/host/localhost.localdomain/Resources --appliance-iso=bin/appliance.iso --bootstrap-iso=bin/bootstrap.iso --container-datastore=datastore1 --image-store=datastore1 --force=true --bridge-network=bridge --target=192.168.113.128 --volume-store=datastore2:myvol --volume-store=datastore2/bing:myvol2
```

However, the user will not be able to write anything to either the myvol or myvol2 datastores, and when attempting to remove appliance, the user will see the following cryptic error messages:

```
bin/vic-machine-linux delete --password='password' --user=root --force=true --compute-resource=/ha-datacenter/host/localhost.localdomain/Resources --target=192.168.113.128
INFO[2016-08-16T11:09:25-05:00] ### Removing VCH ####                        
INFO[2016-08-16T11:09:25-05:00]                                              
INFO[2016-08-16T11:09:25-05:00] VCH ID: VirtualMachine:31                    
INFO[2016-08-16T11:09:25-05:00] Removing VMs                                 
INFO[2016-08-16T11:09:25-05:00] Removing volume stores...                    
INFO[2016-08-16T11:09:25-05:00] Deleting volume store "myvol2" on Datastore "datastore2" at path "bing" 
ERRO[2016-08-16T11:09:25-05:00] Operation failed: Error caused by file /vmfs/volumes/8adb3c9c-e6017b23/bing - (&{{} %!!(MISSING)s(*types.FileFault=&{{{<nil> []}} /vmfs/volumes/8adb3c9c-e6017b23/bing}) Error caused by file /vmfs/volumes/8adb3c9c-e6017b23/bing}) 
INFO[2016-08-16T11:09:25-05:00] Deleting volume store "myvol" on Datastore "datastore2" at path "." 
[...]

PR #2126, opened 2016-08-24
Makes sure valid volume/image stores provided

### LINKED ISSUE #1974: Error response from daemon: [POST /storage/volumes/][500] createVolumeInternalServerError &{Code:0xc82000f540 Message:volumestore (http://VCH-0-8191/storage/volumes/default) not found}
Saw this issue on the latest attempt to test the build on VC in nimbus.  Here is the log bundle:
[VCH-0-8191-container-logs.zip](https://github.com/vmware/vic/files/418319/VCH-0-8191-container-logs.zip)

gigawhitlocks:
Right now, `vic-machine` will happily allow the user to create an image or volume store on a datastore that is not actually writable by the compute-resource. This change adds checks to `validator` to make sure that the datastore can be seen and written to by the compute-resource.

The bulk of the work here is done but I'm still in the process of verifying that everything works as expected in the three cases: VC w/o a cluster (works), VC w/ a cluster (pending), ESX (works)

Comments welcome while I work on testing :)

Fixes #1974 #2000

### DIFF
--- a/lib/install/validate/storage.go
+++ b/lib/install/validate/storage.go
@@ -58,14 +58,10 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
 
 	// TODO: add volume locations
 	for label, volDSpath := range input.VolumeLocations {
-		dsURL, vds, err := v.DatastoreHelper(ctx, volDSpath, label, "--volume-store")
+		dsURL, _, err := v.DatastoreHelper(ctx, volDSpath, label, "--volume-store")
 		v.NoteIssue(err)
 		if dsURL != nil {
 			conf.VolumeLocations[label] = dsURL
-		}
-
-		if !v.targetDatastoreIsWriteable(ctx, vds) {
-			return
 		}
 	}
 }
--- a/lib/install/validate/storage.go
+++ b/lib/install/validate/storage.go
@@ -52,11 +52,6 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
 		v.SetDatastore(ds, imageDSpath)
 		conf.AddImageStore(imageDSpath)
 	}
-
-	if !v.targetDatastoreIsWriteable(ctx, ds) {
-		return
-	}
-
 	if conf.VolumeLocations == nil {
 		conf.VolumeLocations = make(map[string]*url.URL)
 	}
@@ -183,62 +178,4 @@ func (v *Validator) suggestDatastore(path string, label string, flag string) {
 			log.Infof("  %q", d)
 		}
 	}
-}
-
-// checks if a given datastore ds is writeable by compute resource specified on cmd line --compute-resource
-func (v *Validator) targetDatastoreIsWriteable(ctx context.Context, ds *object.Datastore) bool {
-	defer trace.End(trace.Begin(ds.String()))
-
-	// get compute hosts if we haven't run this method on this object before
-	if len(v.computeHosts) == 0 {
-		var err error
-		v.computeHosts, err = v.Session.Cluster.Hosts(ctx)
-		if err != nil {
-			v.NoteIssue(err)
-			return false
-		}
-		// v.computeHosts shouldn't be zero anymore; if it still is, there's a problem
-		if len(v.computeHosts) == 0 {
-			v.NoteIssue(errors.New("Couldn't get list of hosts connected to target cluster"))
-			return false
-		}
-	}
-
-	// now get hosts connected to ds
-	attachedHosts, err := ds.AttachedClusterHosts(ctx, v.Session.Cluster)
-	if err != nil {
-		v.NoteIssue(err)
-		return false
-	}
-
-	if len(attachedHosts) == 0 {
-		// this could mean that we're not operating on a cluster..
-		log.Debugf("AttachedClusterHosts returned nothing; searching for unclustered hosts connected to datastore %s", ds.String())
-		attachedHosts, err := ds.AttachedHosts(ctx)
-
-		if err != nil {
-			v.NoteIssue(err)
-			return false
-		}
-
-		if len(attachedHosts) == 0 {
-			// shouldn't be possible to wind up here, but we should get an error message if we do ¯\_(ツ)_/¯
-			v.NoteIssue(errors.Errorf("No attached hosts found for datastore %s", ds.String()))
-			return false
-		}
-	}
-
-	// find a common host in the two lists
-	for _, c := range v.computeHosts {
-		for _, a := range attachedHosts {
-			if c.Reference() == a.Reference() {
-				return true
-			}
-		}
-	}
-
-	// no common host found -- erroneous user input
-	v.NoteIssue(errors.Errorf("Datastore %s is not writeable from the compute resource specified under --compute-resource", ds.String()))
-	return false
-
 }
--- a/lib/install/validate/storage.go
+++ b/lib/install/validate/storage.go
@@ -39,16 +39,22 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
 		v.NoteIssue(err)
 		return
 	}
+	if err != nil {
+		v.NoteIssue(err)
+	}
 
 	// provide a default path if only a DS name is provided
 	if imageDSpath.Path == "" {
 		imageDSpath.Path = input.DisplayName
 	}
 
-	v.NoteIssue(err)
 	if ds != nil {
 		v.SetDatastore(ds, imageDSpath)
 		conf.AddImageStore(imageDSpath)
+	}
+
+	if !v.targetDatastoreIsWriteable(ctx, ds) {
+		return
 	}
 
 	if conf.VolumeLocations == nil {
@@ -57,10 +63,14 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
 
 	// TODO: add volume locations
 	for label, volDSpath := range input.VolumeLocations {
-		dsURL, _, err := v.DatastoreHelper(ctx, volDSpath, label, "--volume-store")
+		dsURL, vds, err := v.DatastoreHelper(ctx, volDSpath, label, "--volume-store")
 		v.NoteIssue(err)
 		if dsURL != nil {
 			conf.VolumeLocations[label] = dsURL
+		}
+
+		if !v.targetDatastoreIsWriteable(ctx, vds) {
+			return
 		}
 	}
 }
@@ -173,4 +183,62 @@ func (v *Validator) suggestDatastore(path string, label string, flag string) {
 			log.Infof("  %q", d)
 		}
 	}
+}
+
+// checks if a given datastore ds is writeable by compute resource specified on cmd line --compute-resource
+func (v *Validator) targetDatastoreIsWriteable(ctx context.Context, ds *object.Datastore) bool {
+	defer trace.End(trace.Begin(ds.String()))
+
+	// get compute hosts if we haven't run this method on this object before
+	if len(v.computeHosts) == 0 {
+		var err error
+		v.computeHosts, err = v.Session.Cluster.Hosts(ctx)
+		if err != nil {
+			v.NoteIssue(err)
+			return false
+		}
+		// v.computeHosts shouldn't be zero anymore; if it still is, there's a problem
+		if len(v.computeHosts) == 0 {
+			v.NoteIssue(errors.New("Couldn't get list of hosts connected to target cluster"))
+			return false
+		}
+	}
+
+	// now get hosts connected to ds
+	attachedHosts, err := ds.AttachedClusterHosts(ctx, v.Session.Cluster)
+	if err != nil {
+		v.NoteIssue(err)
+		return false
+	}
+
+	if len(attachedHosts) == 0 {
+		// this could mean that we're not operating on a cluster..
+		log.Debugf("AttachedClusterHosts returned nothing; searching for unclustered hosts connected to datastore %s", ds.String())
+		attachedHosts, err := ds.AttachedHosts(ctx)
+
+		if err != nil {
+			v.NoteIssue(err)
+			return false
+		}
+
+		if len(attachedHosts) == 0 {
+			// shouldn't be possible to wind up here, but we should get an error message if we do ¯\_(ツ)_/¯
+			v.NoteIssue(errors.Errorf("No attached hosts found for datastore %s", ds.String()))
+			return false
+		}
+	}
+
+	// find a common host in the two lists
+	for _, c := range v.computeHosts {
+		for _, a := range attachedHosts {
+			if c.Reference() == a.Reference() {
+				return true
+			}
+		}
+	}
+
+	// no common host found -- erroneous user input
+	v.NoteIssue(errors.Errorf("Datastore %s is not writeable from the compute resource specified under --compute-resource", ds.String()))
+	return false
+
 }
--- a/lib/install/validate/storage.go
+++ b/lib/install/validate/storage.go
@@ -39,19 +39,18 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
 		v.NoteIssue(err)
 		return
 	}
-	if err != nil {
-		v.NoteIssue(err)
-	}
 
 	// provide a default path if only a DS name is provided
 	if imageDSpath.Path == "" {
 		imageDSpath.Path = input.DisplayName
 	}
 
+	v.NoteIssue(err)
 	if ds != nil {
 		v.SetDatastore(ds, imageDSpath)
 		conf.AddImageStore(imageDSpath)
 	}
+
 	if conf.VolumeLocations == nil {
 		conf.VolumeLocations = make(map[string]*url.URL)
 	}
--- a/lib/install/validate/validator.go
+++ b/lib/install/validate/validator.go
@@ -369,32 +369,31 @@ func (v *Validator) compatibility(ctx context.Context, conf *config.VirtualConta
 	_, err := v.Session.Datastore.AttachedClusterHosts(v.Context, v.Session.Cluster)
 	v.NoteIssue(err)
 
-	type Stringable interface {
-		String()
-	}
-	getDatastores := func(s []Stringable) []*object.Datastore {
-		datastores := new([]*object.Datastore)
-		for _, i := range s {
-			_, ds, err := v.DatastoreHelper(i.String())
-			if err != nil {
-				v.NoteIssue(err)
-				continue
-			}
-			datastores = append(datastores, ds)
+	// closure to look up a datastore based on a URL from the config & check to see if it's writeable
+	checkDS := func(u *url.URL) {
+		datastores, err := v.Session.Finder.DatastoreList(ctx, fmt.Sprintf("[%s] %s", u.Host, u.Path))
+		if err != nil {
+			v.NoteIssue(err)
 		}
-		return datastores
+		if len(datastores) != 1 {
+			// this is basically an assert, at this point DatastoreList should absolutely only return one result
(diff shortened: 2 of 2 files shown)


lib/install/validate/storage.go
```diff
@@ -40,29 +40,40 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
...
+		dsURL, vds, err := v.DatastoreHelper(ctx, volDSpath, label, "--volume-store")
 		v.NoteIssue(err)
 		if dsURL != nil {
 			conf.VolumeLocations[label] = dsURL
 		}
+
+		if !v.targetDatastoreIsWriteable(ctx, vds) {
+			return
+		}
+
```
gigawhitlocks: strip newline

gigawhitlocks:
Looks like tests are failing in `storage_test.go` likely because I added some govmomi calls  -- @dougm does vcsim offer me a way to stub out those calls?

dougm:
@gigawhitlocks that's what vcsim is all about, mocking vmomi method calls - but there needs to be some implementation of the method(s).  If a method isn't implemented, you'd see this error message:
```
return serverFault(fmt.Sprintf("%s does not implement: %s", method.This, method.Name))
```

If there are methods implemented but causing tests to fail, then there are some properties/state/etc we need to wire up.  I'll have to dig in to get an idea of what's going on here.

lib/install/validate/storage.go
```diff
@@ -39,29 +39,39 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
...
+	}
+
 	if conf.VolumeLocations == nil {
 		conf.VolumeLocations = make(map[string]*url.URL)
 	}
 
 	// TODO: add volume locations
 	for label, volDSpath := range input.VolumeLocations {
-		dsURL, _, err := v.DatastoreHelper(ctx, volDSpath, label, "--volume-store")
+		dsURL, vds, err := v.DatastoreHelper(ctx, volDSpath, label, "--volume-store")
```
dougm: looks like you need to check if err != nil here, else you are calling targetDatastoreIsWriteable below with vds == nil, which panics.

dougm:
@gigawhitlocks I fetched your PR branch locally, test fails the same way locally.. don't see an issue with missing vcsim methods.  Get your error checking fixed up and we can take a closer look then.

lib/install/validate/storage.go
```diff
@@ -39,29 +39,39 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
 		v.NoteIssue(err)
 		return
 	}
+	if err != nil {
```
emlin: no need to check err != nil, which is already done inside of v.NoteIssue

lib/install/validate/storage.go
```diff
@@ -39,29 +39,39 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
...
 		imageDSpath.Path = input.DisplayName
 	}
 
-	v.NoteIssue(err)
 	if ds != nil {
 		v.SetDatastore(ds, imageDSpath)
 		conf.AddImageStore(imageDSpath)
 	}
 
+	if !v.targetDatastoreIsWriteable(ctx, ds) {
```
emlin: this method is no necessary. we already have code to test host/datastore connectivity, which is in the v.compatibility method. 
But currently it only checked session datastore, that one is for image datastore I think.
Please extend that method to include volume datastore validation.

lib/install/validate/storage.go
emlin: It's not enough to check if there is one host connected to this datastore, cause if two datastores connected two different hosts, vch deployment will have problem as well.
So you should do the checking after all datastore validation is finished.

lib/install/validate/storage.go
```diff
@@ -174,3 +184,61 @@ func (v *Validator) suggestDatastore(path string, label string, flag string) {
...
+		if err != nil {
+			v.NoteIssue(err)
+			return false
+		}
+		// v.computeHosts shouldn't be zero anymore; if it still is, there's a problem
+		if len(v.computeHosts) == 0 {
+			v.NoteIssue(errors.New("Couldn't get list of hosts connected to target cluster"))
+			return false
+		}
+	}
```
fdawg4l: Can this whole `if` block be made a helper?  It smells like it's probably used elsewhere and a helper called in the constructor is probably the right thing to do vs doing it everywhere we need a host.

lib/install/validate/storage.go
```diff
@@ -174,3 +184,61 @@ func (v *Validator) suggestDatastore(path string, label string, flag string) {
...
+			v.NoteIssue(err)
+			return false
+		}
+
+		if len(attachedHosts) == 0 {
+			// shouldn't be possible to wind up here, but we should get an error message if we do ¯\_(ツ)_/¯
+			v.NoteIssue(errors.Errorf("No attached hosts found for datastore %s", ds.String()))
+			return false
+		}
+	}
```
fdawg4l: Same comment as above.  Lets make a helper for this and the above and hide all of it somewhere else.

lib/install/validate/storage.go
```diff
@@ -174,3 +184,61 @@ func (v *Validator) suggestDatastore(path string, label string, flag string) {
...
+		}
+		// v.computeHosts shouldn't be zero anymore; if it still is, there's a problem
+		if len(v.computeHosts) == 0 {
+			v.NoteIssue(errors.New("Couldn't get list of hosts connected to target cluster"))
+			return false
+		}
+	}
+
+	// now get hosts connected to ds
+	attachedHosts, err := ds.AttachedClusterHosts(ctx, v.Session.Cluster)
```
fdawg4l: It may be obvious, but it wasn't to me on first glance.  Can you add a comment that this method checks that the host can write to the [datastore](https://github.com/vmware/govmomi/blob/master/object/datastore.go#L316)?

lib/install/validate/storage.go
gigawhitlocks: We spoke about this in person but I'm going to leave a comment here just for context -- the current check looks only at datastores connected to the host the session is connected to. This may or may not be the imagestore, and it may or may not be the volumestore. Instead, my function finds all of the hosts in the computed cluster and compares those to the hosts that each of the user-specified datastores uses and makes sure that those datastores can each be written to by at least one host in the cluster.

lib/install/validate/storage.go
gigawhitlocks: Yup, I can break this out. Ain't nothing wrong with that!

lib/install/validate/storage.go
gigawhitlocks: Definitely -- it's only obvious if you open AttachedClusterHosts or AttachedHosts and look at the comment on _those_ methods, which is an indirection for someone reading this code, so I agree that a comment would be good.

lib/install/validate/storage.go
gigawhitlocks: We check to see if at least one clustered host is connected to each datastore. So if hosts `A` and `B` have datastores `x` and `y`, and `A` and `B` are in a cluster, and `x` is connected to `A` and `y` is connected to `B`, we'll find `y` in the list of both `B.AttachedClusterHosts` and in `Cluster.Hosts` (so we can write, and helper returns `true`) and `x` will be in `A.AttachedClusterHosts` and `Cluster.Hosts` so again helper returns `true` because we can write.

lib/install/validate/validator.go
```diff
@@ -482,3 +512,68 @@ func (v *Validator) AddDeprecatedFields(ctx context.Context, conf *config.Virtua
...
+	defer trace.End(trace.Begin(ds.String()))
+
+	// get compute hosts if we haven't run this method on this object before
+	if len(v.computeHosts) == 0 {
+		var err error
+		v.computeHosts, err = v.getHostsInCluster(ctx)
+		if err != nil {
+			v.NoteIssue(err)
+		}
+	}
```
hmahmood: Should you continue here if `v.computeHosts` is empty?

lib/install/validate/validator.go
```diff
@@ -361,8 +363,36 @@ func (v *Validator) compatibility(ctx context.Context, conf *config.VirtualConta
...
 	}
 
+	// check session's datastore(s) exist
 	_, err := v.Session.Datastore.AttachedClusterHosts(v.Context, v.Session.Cluster)
 	v.NoteIssue(err)
+
+	// closure to look up a datastore based on a URL from the config & check to see if it's writeable
+	checkDS := func(u *url.URL) {
+		datastores, err := v.Session.Finder.DatastoreList(ctx, fmt.Sprintf("[%s] %s", u.Host, u.Path))
+		if err != nil {
```
caglar10ur: https://github.com/vmware/vic/pull/2126/files#r76264344

lib/install/validate/validator.go
```diff
@@ -482,3 +512,68 @@ func (v *Validator) AddDeprecatedFields(ctx context.Context, conf *config.Virtua
...
+
+// checks if a given datastore ds is writeable by compute resource specified on cmd line --compute-resource
+func (v *Validator) isTargetDatastoreWriteable(ctx context.Context, ds *object.Datastore) bool {
+	defer trace.End(trace.Begin(ds.String()))
+
+	// get compute hosts if we haven't run this method on this object before
+	if len(v.computeHosts) == 0 {
+		var err error
+		v.computeHosts, err = v.getHostsInCluster(ctx)
+		if err != nil {
```
caglar10ur: https://github.com/vmware/vic/pull/2126/files#r76264344

lib/install/validate/validator.go
```diff
@@ -482,3 +512,68 @@ func (v *Validator) AddDeprecatedFields(ctx context.Context, conf *config.Virtua
...
+			v.NoteIssue(err)
+		}
+	}
+
+	// now get hosts connected to ds
+	attachedHosts, err := ds.AttachedClusterHosts(ctx, v.Session.Cluster)
+	if err != nil {
```
caglar10ur: https://github.com/vmware/vic/pull/2126/files#r76264344

lib/install/validate/validator.go
```diff
@@ -361,8 +363,36 @@ func (v *Validator) compatibility(ctx context.Context, conf *config.VirtualConta
 		return
 	}
 
+	// check session's datastore(s) exist
 	_, err := v.Session.Datastore.AttachedClusterHosts(v.Context, v.Session.Cluster)
 	v.NoteIssue(err)
+
+	// closure to look up a datastore based on a URL from the config & check to see if it's writeable
+	checkDS := func(u *url.URL) {
```
emlin: Looks like this func is used by all datastore hosts checking, it'll be clearer to have one named method.

lib/install/validate/validator.go
```diff
@@ -482,3 +512,68 @@ func (v *Validator) AddDeprecatedFields(ctx context.Context, conf *config.Virtua
 
 	return &dconfig
 }
+
+// helper to grab HostSystem references out of the Session cluster
+func (v *Validator) getHostsInCluster(ctx context.Context) ([]*object.HostSystem, error) {
```
emlin: Didn't see much value to have this method. 
What you want is actually to find hosts in cluster. if the result of Cluster.Hosts is empty, the later test will find that. 
And you need to check error in the caller no matter how, so suggest to remove this method.

lib/install/validate/validator.go
```diff
@@ -482,3 +512,68 @@ func (v *Validator) AddDeprecatedFields(ctx context.Context, conf *config.Virtua
...
+	}
+	return computeHosts, nil
+}
+
+// checks if a given datastore ds is writeable by compute resource specified on cmd line --compute-resource
+func (v *Validator) isTargetDatastoreWriteable(ctx context.Context, ds *object.Datastore) bool {
+	defer trace.End(trace.Begin(ds.String()))
+
+	// get compute hosts if we haven't run this method on this object before
+	if len(v.computeHosts) == 0 {
```
emlin: This v.computeHosts should be initialized only once, and then while it's empty, this method should return false, cause this method will be called for all datastores.
We should figure out at least one datastore connected to all of them.

lib/install/validate/validator.go
```diff
@@ -482,3 +512,68 @@ func (v *Validator) AddDeprecatedFields(ctx context.Context, conf *config.Virtua
...
+		var err error
+		v.computeHosts, err = v.getHostsInCluster(ctx)
+		if err != nil {
+			v.NoteIssue(err)
+		}
+	}
+
+	// now get hosts connected to ds
+	attachedHosts, err := ds.AttachedClusterHosts(ctx, v.Session.Cluster)
```
emlin: The AttachedClusterHosts method will check if this datastore is connected to cluster hosts, which does not provide value for your current function. 
ds.AttachedHosts is good enough.

lib/install/validate/validator.go
```diff
@@ -482,3 +512,68 @@ func (v *Validator) AddDeprecatedFields(ctx context.Context, conf *config.Virtua
...
+			v.NoteIssue(errors.Errorf("No attached hosts found for datastore %s", ds.String()))
+			return false
+		}
+	}
+
+	// find a common host in the two lists
+	for _, c := range v.computeHosts {
+		for _, a := range attachedHosts {
+			if c.Reference() == a.Reference() {
+				return true
```
emlin: this loop should filter out all hosts not connected to target datastore.
to make sure the v.computeHosts will contains all hosts connected to this datastore.
while the v.computeHosts is empty, your method should return failure, for no hosts connected to all configured datastores.

lib/install/validate/validator.go
```diff
@@ -361,8 +363,36 @@ func (v *Validator) compatibility(ctx context.Context, conf *config.VirtualConta
...
+		checkDS(&u)
+	}
+	for _, u := range conf.ContainerStores {
+		checkDS(&u)
+	}
+
+	for _, u := range conf.VolumeLocations {
+		checkDS(u)
+	}
 }
```
emlin: After all datastores are verified, check if v.computeHosts has only one host left. if yes, warning user that DRS/HA will not work.

lib/install/validate/validator.go
```diff
@@ -361,8 +363,36 @@ func (v *Validator) compatibility(ctx context.Context, conf *config.VirtualConta
...
+		datastores, err := v.Session.Finder.DatastoreList(ctx, fmt.Sprintf("[%s] %s", u.Host, u.Path))
+		if err != nil {
+			v.NoteIssue(err)
+		}
+		if len(datastores) != 1 {
+			// this is basically an assert; at this point DatastoreList should absolutely only return one result
+			v.NoteIssue(errors.Errorf("Looking up datastore %s returned %d results.", u.String(), len(datastores)))
+		}
+		if !v.isTargetDatastoreWriteable(ctx, datastores[0]) {
+			v.NoteIssue(errors.Errorf("Datastore %s is not writeable by the compute resource provided by --compute-resource", u.String()))
```
emlin: Basically, I think the logic is that datastore not connected to host, so the datastore is not writable. But I assume "not writable" might not a VI admin term.

lib/install/validate/storage.go
hickeng: yup - just pass it straight through to NoteIssue and that'll do the right thing.

hickeng:
@gigawhitlocks The current comments indicate a clarity issue with the current approach.
I recommend doing the checks in two phases (forgive the ad-hoc pseudo-code) :

First validate that a specified datastore is accessible to the specified compute resource at all - this can be done for each datastore argument in isolation in simple fashion:
```
var compute object.ComputeResource
datastores := compute.Datastores()
for _, i := range datastoreArgs
  if !contains(i, datastores)
    v.NoteIssue("Datastore %s specified for %s is not accessible to compute resource %s", i.Name(), cliArg.Details(), compute.Name())
```

Second, check for permissible intersects. The first possible approach is to require that all specified datastores be accessible from all hosts (simple), the second is to ensure that there exists viable host placements for any given configuration and provide appropriate warnings (more involved).

Simple:
```
hosts := cluster.Hosts()
numHosts := len(hosts)
for _, i := range datastores
  dsHosts := i.Hosts()
  hosts = intersect(hosts, dsHosts)
  if len(hosts) != numHosts 
    v.NoteWarning("Use of datastore %s reduces number of viable hosts in %s to %d", i.Name(), cluster.Name())
    numHosts = len(hosts)

if len(hosts) < 2
  v.NoteWarning("Only one host can access all data stores - potential point of contention/failure for containers")
```

More complex (the only datastores that must to be accessible to all hosts are the image datastores)
```
ihosts := imagestore.AttachedClusterHosts()
vhosts := cluster.Hosts()

for _, v := range volumeDatastores
  dsHosts := v.Hosts()
  hosts := intersect(ihosts, dsHosts)
  vhosts := intersect(vhosts, dsHosts)

  if len(hosts) != len(ihosts)
    v.NoteWarning("Volume store %s cannot be accessed on all compute hosts", i.Name())

if len(vhosts) == 1 && len(ihosts) > 1
  v.NoteWarning("Only one host can access all volume stores - potential point of contention/failure for containers")

if len(vhosts) == 0
  v.NoteWarning("No host can access all volume stores - potential for invalid container configurations")
```

This requires two utility methods implemented to keep things clear - given these the core code becomes very concise and mostly related to error handling for the govmomi calls and user messages about config issues:
* contains(entity, set) bool
* intersect(set, set) set

gigawhitlocks:
Thank you for the feedback, @hickeng, I do think your approach is clearer, I'll get a refactor out ASAP.

lib/install/validate/validator.go
gigawhitlocks: Literally just broke this out into a method because @fdawg4l asked me to.

lib/install/validate/validator.go
gigawhitlocks: No, I introduced a bug here during the refactor when I broke `getHostsInCluster` into a helper method

gigawhitlocks:
Just an update on this, to anyone who's paying attention (probably talking to vacuum but you never know) I have a refactor saved locally that I'm starting to test, hope to get an updated PR out tomorrow

gigawhitlocks:
Also when this is in it will likely fix #2127 as well

gigawhitlocks:
Last commit at least doesn't break anything on standalone ESXi. Still testing on other environments.

lib/install/validate/validator.go
```diff
@@ -363,8 +364,103 @@ func (v *Validator) compatibility(ctx context.Context, conf *config.VirtualConta
...
+}
+
+// looks up a datastore and adds it to the set
+func (v *Validator) getDatastore(ctx context.Context, u *url.URL, datastoreSet map[*object.Datastore]bool) {
+	if datastoreSet == nil {
+		datastoreSet = make(map[*object.Datastore]bool)
+	}
+
+	datastores, err := v.Session.Finder.DatastoreList(ctx, u.Host)
+	if err != nil {
```
emlin: no need to check here. Is done in NoteIssue

lib/install/validate/validator.go
```diff
@@ -363,8 +364,103 @@ func (v *Validator) compatibility(ctx context.Context, conf *config.VirtualConta
...
+			delete(requestedDatastores, cds)
+			validStores[cds] = true
+		}
+	}
+	// if requestedDatastores is not empty, some requested datastores are not writable
+	for store := range requestedDatastores {
+		v.NoteIssue(errors.Errorf("Datastore %s is not accessible by the compute resource", store.String()))
+	}
+
+	v.NoteIssue(err)
```
emlin: is this one necessary? didn't see where err is assigned

gigawhitlocks: looks like a duplicate of 412 that got left behind

emlin:
nice and clean. LGTM 👍

matthewavery:
LGTM

gigawhitlocks:
poke

Issue #2038, opened 2016-08-17
${output} = Error response from daemon: server error from portlayer : Operation failed: The attempted operation cannot be performed in the current state (Powered off). - (&{{} %!s(*types.InvalidPowerState=&{{{{<n...	


mhagen-vmware:
Saw this in a recent CI build:
[integration_logs_4112_c61b35292f46740fc318eab050a8fde3a426853b_1471463307.zip](https://github.com/vmware/vic/files/423662/integration_logs_4112_c61b35292f46740fc318eab050a8fde3a426853b_1471463307.zip)

hmahmood:
The code to handle this error is already in `Container.stop`, but is not run because `tasks.WaitForResult` does not preserve the error from govmomi. 

@matthewavery is already working on fixing `WaitForResult` as part of #2013, so assigning it to him.

matthewavery:
Solved in pr #2141 due to the previous implementation of `WaitForResult` the fault's were getting squashed and never made it back to the callers. This has been fixed, and the fault checks on errors are now valid is this area.

PR #2141, opened 2016-08-25
[full ci] P0 volume ls

### LINKED ISSUE #2013: Error response from daemon: [POST /storage/volumes/][500] createVolumeInternalServerError  &{Code:0xc82000f700 Message:Another task is already in progress.}

matthewavery:
Fixes #2013 #2038

### DIFF
--- a/pkg/vsphere/datastore/datastore.go
+++ b/pkg/vsphere/datastore/datastore.go
@@ -239,7 +239,7 @@ func (d *Helper) Mv(ctx context.Context, fromPath, toPath string) error {
 	from := path.Join(d.RootURL, fromPath)
 	to := path.Join(d.RootURL, toPath)
 	log.Infof("Moving %s to %s", from, to)
-	err := tasks.Wait(ctx, func(context.Context) (tasks.ResultWaiter, error) {
+	err := tasks.Wait(ctx, func(context.Context) (tasks.Task, error) {
 		return d.fm.MoveDatastoreFile(ctx, from, d.s.Datacenter, to, d.s.Datacenter, true)
 	})
 
@@ -249,7 +249,7 @@ func (d *Helper) Mv(ctx context.Context, fromPath, toPath string) error {
 func (d *Helper) Rm(ctx context.Context, pth string) error {
 	f := path.Join(d.RootURL, pth)
 	log.Infof("Removing %s", pth)
-	return tasks.Wait(context.TODO(), func(ctx context.Context) (tasks.ResultWaiter, error) {
+	return tasks.Wait(context.TODO(), func(ctx context.Context) (tasks.Task, error) {
 		return d.fm.DeleteDatastoreFile(ctx, f, d.s.Datacenter)
 	})
 }
--- a/pkg/vsphere/datastore/datastore.go
+++ b/pkg/vsphere/datastore/datastore.go
@@ -239,7 +239,7 @@ func (d *Helper) Mv(ctx context.Context, fromPath, toPath string) error {
 	from := path.Join(d.RootURL, fromPath)
 	to := path.Join(d.RootURL, toPath)
 	log.Infof("Moving %s to %s", from, to)
-	err := tasks.Wait(ctx, func(context.Context) (tasks.Waiter, error) {
+	err := tasks.Wait(ctx, func(context.Context) (tasks.ResultWaiter, error) {
 		return d.fm.MoveDatastoreFile(ctx, from, d.s.Datacenter, to, d.s.Datacenter, true)
 	})
 
@@ -249,7 +249,7 @@ func (d *Helper) Mv(ctx context.Context, fromPath, toPath string) error {
 func (d *Helper) Rm(ctx context.Context, pth string) error {
 	f := path.Join(d.RootURL, pth)
 	log.Infof("Removing %s", pth)
-	return tasks.Wait(context.TODO(), func(ctx context.Context) (tasks.Waiter, error) {
+	return tasks.Wait(context.TODO(), func(ctx context.Context) (tasks.ResultWaiter, error) {
 		return d.fm.DeleteDatastoreFile(ctx, f, d.s.Datacenter)
 	})
 }
--- a/pkg/vsphere/tasks/waiter.go
+++ b/pkg/vsphere/tasks/waiter.go
@@ -104,7 +104,12 @@ func Retry(ctx context.Context, f func(context.Context) (ResultWaiter, error)) (
 
 	for {
 		taskInfo, err = WaitForResult(ctx, f)
-		if err == nil && taskInfo.Error == nil {
+
+		if err != nil {
+			return nil, err
+		}
+
+		if taskInfo.Error == nil {
 			break
 		}
 
--- a/pkg/vsphere/tasks/waiter.go
+++ b/pkg/vsphere/tasks/waiter.go
@@ -17,6 +17,7 @@
 package tasks
 
 import (
+	"fmt"
 	"math/rand"
 	"time"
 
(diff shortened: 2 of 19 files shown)


pkg/vsphere/tasks/waiter.go
```diff
@@ -65,20 +73,80 @@ func Wait(ctx context.Context, f func(context.Context) (Waiter, error)) error {
 func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, error)) (*types.TaskInfo, error) {
 	task, err := f(ctx)
 	if err != nil {
-		cerr := errors.Errorf("Failed to invoke operation: %s", errors.ErrorStack(err))
-		log.Errorf(cerr.Error())
-		return nil, cerr
+		terr := &TaskError{
```
hmahmood: Why don't you just return `err` here?

pkg/vsphere/tasks/waiter.go
```diff
@@ -65,20 +73,80 @@ func Wait(ctx context.Context, f func(context.Context) (Waiter, error)) error {
...
+			taskError: nil,
+		}
+		log.Errorf(terr.Error())
+		return nil, terr
 	}
 
 	info, err := task.WaitForResult(ctx, nil)
 	if err != nil {
-		cerr := errors.Errorf("Operation failed: %s", errors.ErrorStack(err))
+		terr := &TaskError{
```
hmahmood: Same comment here

pkg/vsphere/tasks/waiter.go
matthewavery: because the current way other functions rely on this function they do not expect an info object back on a failure. if we want to see the fault our two choices are to return the info even on error, or create a new error type that has the LocalizedMethodFault packed into it.

matthewavery: We also want to log the error before returning it. atleast for now.

pkg/vsphere/tasks/waiter.go
```diff
@@ -78,7 +85,44 @@ func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, e
...
+
+func Retry(ctx context.Context, f func(context.Context) (ResultWaiter, error)) (*types.TaskInfo, error) {
+	r := rand.New(rand.NewSource(time.Now().UnixNano())) //creates a more unique random
+	var err error
+	var taskInfo *types.TaskInfo
+	backoffFactor := int64(1)
+
+	for {
+		taskInfo, err = WaitForResult(ctx, f)
+		if err != nil && taskInfo == nil {
```
matthewavery: I am pretty sure the logic here can be simplified to `if taskInfo == nil` thoughts?

dougm: Just use Wait(ctx, f) instead of WaitForResult() - just seeing now that taskInfo isn't used for anything other than the error.  See: https://github.com/vmware/govmomi/blob/master/task/wait.go#L97-L111

If Wait() returns err != nil:
```go
if terr, ok:= err.(*task.Error); ok {
   if _, ok := terr.Fault().(types.TaskInProgressFault); ...
} else {
   return err // not a task.Error
}
```

matthewavery: awesome this is great!

dougm: For reference, this is the example I had pointed you at a while back: https://github.com/vmware/vic/commit/936a10a4e99665dd7521fe7fbd8d6482296e0621

pkg/vsphere/tasks/waiter.go
```diff
@@ -63,22 +62,45 @@ func Wait(ctx context.Context, f func(context.Context) (Waiter, error)) error {
 //       return vm.Reconfigure(ctx, config)
 //    })
 func WaitForResult(ctx context.Context, f func(context.Context) (ResultWaiter, error)) (*types.TaskInfo, error) {
-	task, err := f(ctx)
+	r := rand.New(rand.NewSource(time.Now().UnixNano())) //creates a more unique random
```
dougm: mentioned earlier, no need for this.  Just use rand.Int63n() below.

pkg/vsphere/tasks/waiter.go
```diff
@@ -63,22 +62,45 @@ func Wait(ctx context.Context, f func(context.Context) (Waiter, error)) error {
...
+				sleepValue := time.Duration(backoffFactor * (r.Int63n(100) + int64(50)))
+				select {
+				case <-time.After(sleepValue * time.Millisecond):
+					if backoffFactor*2 > maxBackoffFactor {
+						backoffFactor = maxBackoffFactor
+					} else {
+						backoffFactor *= 2
+					}
+				case <-ctx.Done():
+					log.Errorf("Context Deadline Exceeded while trying to Retry task : %#v", taskInfo)
```
dougm: ctx.Done() could also be due to cancellation (vs deadline).  Let's make that:
```go
err = fmt.Errorf("%s while retrying task %#v", ctx.Error(), taskInfo)
log.Error(err)
return nil, err
```

pkg/vsphere/tasks/waiter.go
```diff
@@ -63,22 +62,45 @@ func Wait(ctx context.Context, f func(context.Context) (Waiter, error)) error {
...
+					return nil, ctx.Err()
+				}
+				log.Infof("Retrying Task due to TaskInProgressFault: %s", taskInfo.Task.Reference())
```
dougm: log.Warnf instead?  Rather than "TaskInProgressFault" literal, you could %#v terr.Fault(), that'll include the fault type name and the moref of the task that is currently in progress.

matthewavery: makes sense, makes more of a distinction since it isnt really `info`

pkg/vsphere/tasks/waiter.go
matthewavery: @dougm we may not need it. but it does provide more randomness per run. or does the default `rand` already do that?

matthewavery: looks like the default sournce always calls `NewSource(1)` does it hurt us to set our own?

dougm: '1' is just the seed, as time.Now().  With the use of 'r.Int63n(100)' here, you're picking a random number in the range of 0..100, so the seed won't make much difference in this case.  Why add code we don't need?

pkg/vsphere/datastore/datastore.go
```diff
@@ -239,7 +239,7 @@ func (d *Helper) Mv(ctx context.Context, fromPath, toPath string) error {
 	from := path.Join(d.RootURL, fromPath)
 	to := path.Join(d.RootURL, toPath)
 	log.Infof("Moving %s to %s", from, to)
-	err := tasks.Wait(ctx, func(context.Context) (tasks.Waiter, error) {
+	err := tasks.Wait(ctx, func(context.Context) (tasks.ResultWaiter, error) {
```
hmahmood: Why did you make this change? If you are going to change this, might as well use `object.Task` from govmomi.

matthewavery: I had to for the tests. They were passing `waiter` objects I to the `wait` tests and we call `WaitForResult` in the wait call now. I thought about making the object.task change now. Wanted to clear this up first before potentially using the task object.

pkg/vsphere/tasks/waiter.go
```diff
@@ -39,21 +46,11 @@ type ResultWaiter interface {
...
-
-	err = task.Wait(ctx)
-	if err != nil {
-		cerr := errors.Errorf("Operation failed: %s", errors.ErrorStack(err))
-		log.Errorf(cerr.Error())
-		return cerr
-	}
-	return nil
+func Wait(ctx context.Context, f func(context.Context) (Task, error)) error {
+	_, err := WaitForResult(ctx, func(context.Context) (Task, error) {
```
hmahmood: Don't need the extra func now.

pkg/vsphere/tasks/waiter.go
```diff
@@ -62,23 +59,47 @@ func Wait(ctx context.Context, f func(context.Context) (Waiter, error)) error {
...
+				}
+				log.Warnf("Retrying Task due to TaskInProgressFault: %s", taskInfo.Task.Reference())
+				continue
+			}
+		}
+		err = werr
+		break
 	}
-	return info, nil
+	log.Errorf("Task failed with non fault error : %s", err)
```
hmahmood: It could be a fault too.

pkg/vsphere/tasks/waiter.go
matthewavery: ah yeah, fixing

hmahmood:
lgtm

dougm:
:sheep: :it:

matthewavery:
♻️

PR #2157, opened 2016-08-26
vic-machine delete cleans up VMs in invalid state

### LINKED ISSUE #1787: vic-machine delete leaving containers behind
I'm seeing container remnants left on several of the CI systems. Some of them cannot easily be identified as they are tagged as _invalid_, meaning that the corresponding datastore files have been deleted directly.

I'm also seeing empty VCH-X-Y resource pools left lying around on the systems, and containerVMs that have been orphaned (resource pool deleted before the VMs)

However some of them are named _webserver-xxx_ from the pretty name PR. This suggests to me that something related to the network port mapping tests could be the culprit. A quick check of that test suite showed nothing obvious however.

image

image

image

andrewtchin:
Fixes #1787

If we fail to destroy a VM, unregister it and attempt to delete its files from the datastore.

Regression tests pass on nimbus esx


lib/install/management/appliance.go
```diff
@@ -168,7 +168,14 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
 	})
 	if err != nil {
 		err = errors.Errorf("Failed to destroy vm %q: %s", vm.Reference(), err)
-		return err
+		// Unregister then do we still want to deleteDatastoreFiles? something about attached volumes
```
emlin: Attached volumes are ok here, cause only for vm delete, we'll delete any attached volumes. 
Delete datastore files will not delete volumes cause they are in different datastore and directories

lib/install/management/appliance.go
```diff
@@ -127,8 +127,17 @@ func (d *Dispatcher) checkExistence(conf *config.VirtualContainerHostConfigSpec,
...
+	if err != nil {
+		log.Errorf("VM name is not found: %s", err)
+		return ""
+	}
+	return name
+}
+
 func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
-	defer trace.End(trace.Begin(""))
+	defer trace.End(trace.Begin(vm.String()))
```
emlin: it's good. what about add force variable as well?

lib/install/management/appliance.go
```diff
@@ -160,18 +166,30 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
...
-		err = errors.Errorf("Failed to destroy vm %q: %s", vm.Reference(), err)
-		return err
+		err = errors.Errorf("Failed to destroy VM %q: %s", vm.Reference(), err)
+		err2 := vm.Unregister(d.ctx)
+		if err2 != nil {
+			return errors.Errorf("%s then failed to unregister VM: %s", err, err2)
+		}
+		log.Infof("Unregistered VM to cleanup after failed destroy: %q", vm.Reference())
 	}
 	if _, err = d.deleteDatastoreFiles(d.session.Datastore, folder, true); err != nil {
```
emlin: Looks like it's possible to have "" as folder name to delete datastore files, suggest to have a detection in the deleteDatastoreFiles. 
If the folder is "", return directly. Cause that method might try to remove everything in the datastore.

emlin:
Nice job. 👍 
lgtm

jzt:
LGTM

andrewtchin:
♻️

PR #2201, opened 2016-08-30
Assign error for cleanup lambda to remove inconsistent image

fdawg4l:
Errors thrown by the checksum not matching weren't triggering the cleanup lambda since `err` was unset.


matthewavery:
lgtm

fdawg4l:
LGTM

Issue #437, opened 2016-04-05
Milestone: docker rmi


hickeng:
Delete images, failing if they are not leaves of the tree or are in an image store that the VCH isn't directly managing.

corrieb:
NOTE: CLOSING THIS ISSUE MAY CAUSE AN INTEGRATION TEST FAILURE. Please re-enable the test in 1-12-Docker-RMI.robot as part of a PR to fix this issue. Please also ensure that the test passes cleanly. If the test needs improving or updating, please address it. Thx

corrieb:
@fdawg4l assigning this to you to co-ordinate with @sflxn. Is it realistic to be targeting onsite with this? I note that #253 is targeted at GA

fdawg4l:
@corrieb I don't see this happening in the next 5days.

corrieb:
@hickeng @mdubya66 note this is moving out of onsite to GA

mdubya66:
GA feature from SSOD

mdubya66:
Only Documentation remains, moving to Doc milestone

stuclem:
Doc is done. Closed https://github.com/vmware/vic/issues/835.

hickeng:
Re-opening until robot rmi tests are re-enabled.
@fdawg4l @vburenin 

I tried simply un-commenting them and deleting the check however they fail. It doesn't seem like it's a simple matter of updated expected strings (although the error messages are quite horrific as they stand).

The primary functional failure is `rmi -f <id>` which returns exactly the same message as `rmi <id>`:
```
<msg timestamp="20161009 08:12:39.261" level="INFO">Running command 'docker -H 192.168.78.230:2376 --tls rmi -f busybox 2&gt;&amp;1'.</msg>
<msg timestamp="20161009 08:12:39.280" level="INFO">${rc} = 1</msg>
<msg timestamp="20161009 08:12:39.280" level="INFO">${output} = Error response from daemon: [DELETE /storage/{store_name}/info/{id}][423] deleteImageLocked  &amp;{Code:0xc420314550 Message:image 9967c5ad88de8c101809f7f22d4774b6791fe46ac3033d57abf7ebb1dd8e36ee in use b...</msg>
```

fdawg4l:
```
$ ag -r deleteImageLocked *
lib/apiservers/portlayer/restapi/operations/storage/delete_image_responses.go
89:swagger:response deleteImageLocked

lib/apiservers/portlayer/client/storage/delete_image_responses.go
[...]

PR #2215, opened 2016-08-31
[full ci] Use pkg/index for storage image cache

fdawg4l:
This is in support of `rmi` (#437).  We need to keep images in b+trees so we can tell what can be deleted and what can't.

### DIFF
--- a/lib/portlayer/storage/image.go
+++ b/lib/portlayer/storage/image.go
@@ -24,29 +24,8 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/vmware/vic/lib/portlayer/util"
+	"github.com/vmware/vic/pkg/index"
 )
-
-// Image is the handle to identify an image layer on the backing store.  The
-// URI namespace used to identify the Image in the storage layer has the
-// following path scheme:
-//
-// `/storage/<image store identifier, usually the vch uuid>/<image id>`
-//
-type Image struct {
-	// Identifer for this layer.  Usually a SHA
-	ID string
-
-	// location of the layer.  Filled in by the runtime.
-	SelfLink *url.URL
-
-	// Parent's location.  It's the VMDK this snapshot inerits from.
-	Parent *url.URL
-
-	Store *url.URL
-
-	// Metadata associated with the image.
-	Metadata map[string][]byte
-}
 
 // ImageStorer is an interface to store images in the Image Store
 type ImageStorer interface {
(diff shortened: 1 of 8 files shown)


hmahmood:
Is there a go package out there that can do this? Maybe a graph library?

fdawg4l:
@hmahmood Probably.  But why does it matter? What's your concern?

fdawg4l:
FYI `pkg/index` was merged yesterday.  This PR is to simply use it.

cgtexmex:
LGTM

lib/portlayer/storage/image.go
```diff
@@ -87,6 +66,72 @@ type ImageStorer interface {
...
+// following path scheme:
+//
+// `/storage/<image store identifier, usually the vch uuid>/<image id>`
+//
+type Image struct {
+	// Identifer for this layer.  Usually a SHA
+	ID string
+
+	// location of the layer.  Filled in by the runtime.
+	SelfLink *url.URL
```
dougm: // SelfLink ... - make go lint happy

lib/portlayer/storage/image.go
```diff
@@ -87,6 +66,72 @@ type ImageStorer interface {
...
+
+	// Parent's location.  It's the VMDK this snapshot inerits from.
```
dougm: // ParentLink ... - make go lint happy

lib/portlayer/storage/image.go
```diff
@@ -87,6 +66,72 @@ type ImageStorer interface {
...
+
+	c := &Image{
+		ID:         i.ID,
+		SelfLink:   selflink,
+		ParentLink: parent,
+		Store:      store,
+	}
+
+	if i.Metadata != nil {
+
```
dougm: RNL

lib/portlayer/storage/image_cache.go
```diff
@@ -193,57 +219,83 @@ func (c *NameLookupCache) GetImage(ctx context.Context, store *url.URL, ID strin
...
+		indx := c.storeCache[*store]
+		c.storeCacheLock.Unlock()
+
+		images, err := indx.List()
+		if err != nil {
+			return nil, err
+		}
+
+		for _, v := range images {
+
```
dougm: RNL

hmahmood:
@fdawg4l I didn't realize the index package was already. I just wanted to see if there was already a well known graph package out there that we can utilize, instead of writing our own which of course we would have to maintain and test properly.

fdawg4l:
.

PR #2216, opened 2016-08-31
Refactor event manager

### LINKED ISSUE #2206: Refactor Event Manager to pub / sub
Refactor the event manager to provide event publish / subscribe for portlayer components.

cgtexmex:
Refactoring will focus on the event manager providing pub / sub of event activity for the various Portlayer components.  Additionally, it will create the notion of a Collector that can be registered with the event manager for event publishing and notification to subscribers.

Fixes #2206


cgtexmex:
x

lib/portlayer/event/collector/vsphere/collector.go
```diff
@@ -0,0 +1,177 @@
...
+	callback   func(events.Event)
+}
+
+type monitoredCache struct {
+	mu sync.RWMutex
+
+	mos map[string]types.ManagedObjectReference
+}
+
+func NewCollector(session *session.Session, objects ...string) *EventCollector {
```
hickeng: Can we use Client directly here instead of session? This appears to be the only dependency on session in the collector.

cgtexmex: sure

cgtexmex:
@hickeng updated the collector to use client directly -- other concerns / questions?

lib/portlayer/event/manager.go
```diff
@@ -0,0 +1,158 @@
...
+	if collector == nil {
+		return
+	}
+
+	mgr.cos.mu.Lock()
+	defer mgr.cos.mu.Unlock()
+
+	collector.Register(mgr.Publish)
+
+	mgr.cos.collectors[collector.Name()] = collector
```
hickeng: This could be an issue (or not) currently - the collector's name is the same for all collectors.
This doesn't unregister an existing collect if it's overwritten, which means we'd still get even callbacks from it?

lib/portlayer/event/manager.go
```diff
@@ -0,0 +1,158 @@
...
+// Subscribe to the event manager for callback
+func (mgr *Manager) Subscribe(eventTopic interface{}, caller string, callback func(events.Event)) {
+	mgr.subs.mu.Lock()
+	defer mgr.subs.mu.Unlock()
+	t := topic(eventTopic)
+	if _, ok := mgr.subs.subscribers[t]; !ok {
+
+		mgr.subs.subscribers[t] = make(map[string]func(events.Event))
+	}
+	mgr.subs.subscribers[t][caller] = callback
```
hickeng: Should we check whether caller already exists and return the old entry that we're replacing?

lib/portlayer/event/manager.go
```diff
@@ -0,0 +1,158 @@
...
+	t := topic(e)
+
+	// subscribers for this event
+	subs := mgr.subs.subscribers[t]
+
+	log.Debugf("Found %d subscribers to %s", len(subs), reflect.TypeOf(t))
+
+	for sub, f := range subs {
+		//TODO: need to introduce timeout -- either with callbacks
+		// or when channels are introduced
```
hickeng: or run the callback in a goroutine?

hickeng:
Could we have a package comment that describes the various Collector, Manager, Subscriber interactions, along with the permitted multiplicities (I thought it was single collector only having read through the collector code until I got to the manager). Something to give people a conceptual framework to hang the details off.

Some comments to be addressed (code fixes or refute), but then LGTM.

lib/portlayer/event/event_manager.go
```diff
@@ -15,38 +15,31 @@
...
 
-	// RemoveMonitoredObject will remove the object from event listening
-	RemoveMonitoredObject(ref string)
+	// Collectors returns registered collectors
+	Collectors() map[string]collector.Collector
 
-	// Register will for event callbacks
-	Register(caller string, callback func(Event, *session.Session))
+	// Subscribe for event callbacks
+	Subscribe(event interface{}, caller string, callback func(events.Event))
```
dougm: Can we use a more concrete type `interface{}` here?

lib/portlayer/event/manager.go
cgtexmex: Currently I don't think it's an issue as we are only getting VM events -- but I agree that it might need to be changed to allow multiple collectors of type vsphere event collector.  I'd prefer to tackle it in a subsequent change if that's agreeable.

lib/portlayer/event/manager.go
cgtexmex: we could -- we don't have a use for it currently, but I could see it being beneficial in the future.  Would you like me to change that as part of this change?

lib/portlayer/event/manager.go
```diff
@@ -0,0 +1,158 @@
...
+		log.Debugf("Event manager calling back to %s", sub)
+		f(e)
+	}
+}
+
+// topic is helper function that will retun the default
+// interface of the passed event
+func topic(e interface{}) interface{} {
```
dougm: Related to the `Subscribe(interface{})` question.. looks like this creates a new instance of the given arg type?  Not sure what you mean by "default interface" in the comment.  Again, seems like we should use a more concrete type for a "topic" rather than `interface{}`.

lib/portlayer/event/manager.go
cgtexmex: Potentially -- I'm working on the next change (emitting events from the exec layer) and thought about running the publish in a goroutine, but looked at it as premature optimization.  I don't believe we'll hit contention issues now -- but I do believe it needs to be addressed.  Ultimately I believe the callbacks / publishes, etc..should be fire and forget.  

If agreed I'll open an issue discussing the need and we can address in the near future.

lib/portlayer/event/event_manager.go
cgtexmex: The intent was to allow flexibility for any concrete event type, so used interface{}

lib/portlayer/event/manager.go
cgtexmex: Yes this does create a the unaltered base type -- that's what is used to subscribe.  Again an attempt at flexibility.

Open to suggestions -- possibly a discussion is more in order than comments?

lib/portlayer/event/event_manager.go
dougm: Maybe just add a `Topic() string` method to the `Event` interface then?  In addition to having `Subscribe` typed, you can use that as your map key too (instead of interface{}) and avoid the runtime reflection overhead of the `topic()` helper.  Also avoids the garbage created by reflect.New(), since the topic() return value is only used as a map key.

cgtexmex: I thought about that -- but was concerned with the definition and maintenance of a topic.  I wasn't happy about the reflection overhead, but since it's used minimally I thought it to be acceptable.

If you think it's worth it I can move to a string topic -- I'll define the topics as a constant in the event package.  Then as new events are created, removed, etc...the constant and the identification method (this string == this event, etc) will need to be updated.  Is that what you're thinking?

cgtexmex:
@hickeng basic doc.go added providing an overview.  Let me know if you'd like more detail.

lib/portlayer/event/event_manager.go
dougm: You can still use the type-as-topic and just do reflection once, at init time:

```go
func topicName(kind interface{}) string {
	t := reflect.TypeOf(kind).Elem()
	return fmt.Sprintf("%s.%s", t.PkgPath(), t.Name())
}

var topic = topicName((*types.VmPoweredOffEvent)(nil))
```

Could also have a helper like so:

```go
type EventTopic interface {
	Topic() string
}

type eventTopic string

func (t eventTopic) Topic() string {
	return string(t)
}

func NewEventTopic(kind interface{}) EventTopic {
	t := reflect.TypeOf(kind).Elem()
	return eventTopic(fmt.Sprintf("%s.%s", t.PkgPath(), t.Name()))
}
```

And embed EventTopic interface in the existing Events interface.

Using this `topicName` at runtime would be ok too, at least avoids the New() garbage.  And you can use string keys for the map(s) instead of interface{}.

cgtexmex: Ok -- gotcha -- more of a hybrid approach with some reflection...very nice...will do!

lib/portlayer/event/manager.go
hickeng: We can do it in a followup. Many of my other comments also relate to the handling of name collisions and can be deferred to the same followup.

lib/portlayer/event/manager.go
hickeng: Open an issue. I think we'll settle on goroutine given how simple it is to fire. If we want to be really proper we can pass a clone of the event to avoid concurrent modification issues.

cgtexmex: #2240 created...

cgtexmex:
@hickeng @dougm requested change implemented

lib/portlayer/event/event_manager.go
hmahmood: How would a caller subscribe to all events?

cgtexmex:
@hmahmood currently you cannot subscribe to all events in a single call.  I had planned on doing that with a nil interface...we'll need to add that functionality.  I'll open an issue.

lib/portlayer/event/manager.go
```diff
@@ -0,0 +1,146 @@
...
+type collectorCache struct {
+	mu sync.RWMutex
+
+	collectors map[string]collector.Collector
+}
+
+type subscriberCache struct {
+	mu sync.RWMutex
+
+	subscribers map[string]map[string]func(events.Event)
```
hmahmood: I wonder if you could use a some string based scheme here to key the subscribers instead of using a map of maps? Maybe topic+caller?

lib/portlayer/event/event_manager.go
```diff
@@ -15,38 +15,31 @@
...
 
-	// RemoveMonitoredObject will remove the object from event listening
-	RemoveMonitoredObject(ref string)
+	// Collectors returns registered collectors
+	Collectors() map[string]collector.Collector
 
-	// Register will for event callbacks
-	Register(caller string, callback func(Event, *session.Session))
+	// Subscribe for event callbacks
+	Subscribe(eventTopic string, caller string, callback func(events.Event))
```
hmahmood: I would say a more flexible approach is to have a filter interface/callback that can be passed in so that callers can fine tune the events they want. For instance, I may want to listen to events for only a particular object. 

```go
Subscribe(filter func(events.Event) bool, caller string, callback func(events.Event)
```
You could also provide helper functions for stock filters, like for topic or object.

cgtexmex: Yeah...agree would be more flexible...I'd open an issue as a suggested enhancement

lib/portlayer/event/events/base_event.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func (be *BaseEvent) Reference() string {
+	return be.Ref
+}
+
+// NewEventType utility function that uses reflection to return
+// the event type
+func NewEventType(kind interface{}) EventType {
+	var t reflect.Type
```
dougm: Looks like you could simplify with:
```go
t := reflect.TypeOf(kind)
if t.Kind() == reflect.Ptr {
     t = t.Elem()
}
```

lib/portlayer/event/manager.go
dougm: That'd be great if we can simplify and collapse the map-of-maps to a single map.

dougm:
lgtm

hickeng:
LGTM - I'm sure you've already noted them, but please do open issue or a tracking issue for the followups.

PR #2224, opened 2016-09-01
Restructure to create root datastore directory on start/restart

hmahmood:
Found this when implementing vch restart. On a restart the datastore root url would be empty.


fdawg4l:
@hmahmood lets chat before you check this in.

pkg/vsphere/datastore/datastore.go
```diff
@@ -268,54 +252,56 @@ func (d *Helper) IsVSAN(ctx context.Context) bool {
...
-			if os.IsExist(err) {
-				log.Debugf("%s already exists", d.RootURL)
-				return nil
-			}
-			return err
-		}
+	// create the rest of the root dir in case of vSAN, otherwise
+	// create the full path
+	if _, err := mkdir(ctx, d.s, d.fm, true, rooturl); err != nil && !os.IsExist(err) {
+		return err
```
fdawg4l: By my reading if the root exists, we fall out here, not setting RootURL.  No?

pkg/vsphere/datastore/datastore.go
```diff
@@ -268,54 +252,56 @@ func (d *Helper) IsVSAN(ctx context.Context) bool {
...
+		uuid, err := nm.CreateDirectory(ctx, d.ds, comps[0], "")
 		if err != nil {
 			if !soap.IsSoapFault(err) {
 				return err
 			}
 
 			soapFault := soap.ToSoapFault(err)
-			_, ok := soapFault.VimFault().(types.FileAlreadyExists)
-			if ok {
+			if _, ok := soapFault.VimFault().(types.FileAlreadyExists); ok {
```
hickeng: if we're restructuring, change this test to `!ok` so that we can directly return and drop the else clause.

pkg/vsphere/datastore/datastore.go
fdawg4l: George pointed out we're checking for the exist err.  Ignore my previous question.

pkg/vsphere/datastore/datastore.go
```diff
@@ -268,54 +252,56 @@ func (d *Helper) IsVSAN(ctx context.Context) bool {
...
+	}
 
-		// Handle regular local datastore
-		// check if it already exists
+	rooturl := d.ds.Path(rootdir)
 
-		d.RootURL = d.ds.Path(rootdir)
-		if _, err := d.Mkdir(ctx, true); err != nil {
-			if os.IsExist(err) {
-				log.Debugf("%s already exists", d.RootURL)
```
fdawg4l: It would be nice to know if the datastore already exists.  Can we log this condition?

hickeng:
This change makes it more readable, however I cannot identify a path by which d.RootURL would not have  been set in the prior code. What is the specific circumstance?

hmahmood:
@hickeng the actual problem was in MkDir (not returning the actual path in case it existed) and was fixed by @fdawg4l earlier this week unbeknown to me. So I guess this is a restructure not a fix.

fdawg4l:
LGTM

hickeng:
@hmahmood good, means I didn't misunderstand the change. LGTM

hmahmood:
:recycle:

Issue #2203, opened 2016-08-31
TestLogTail ran too long


fdawg4l:
```
2016/08/31 00:00:38 Testing TestLogTail
2016/08/31 00:00:38 GET https://localhost:40694/logs/tail/vicadmin.log:
time="2016-08-31T00:00:38Z" level=debug msg="[BEGIN] [github.com/vmware/vic/cmd/vicadmin.(*server).serve]" 
time="2016-08-31T00:00:38Z" level=debug msg="[BEGIN] [github.com/vmware/vic/cmd/vicadmin.(*server).handleFunc]" 
time="2016-08-31T00:00:38Z" level=debug msg="[ END ] [github.com/vmware/vic/cmd/vicadmin.(*server).handleFunc] [1.222µs] " 
time="2016-08-31T00:00:38Z" level=debug msg="[BEGIN] [github.com/vmware/vic/cmd/vicadmin.(*server).handleFunc]" 
time="2016-08-31T00:00:38Z" level=debug msg="[ END ] [github.com/vmware/vic/cmd/vicadmin.(*server).handleFunc] [740ns] " 
time="2016-08-31T00:00:38Z" level=debug msg="[BEGIN] [github.com/vmware/vic/cmd/vicadmin.logFiles]" 
time="2016-08-31T00:00:38Z" level=debug msg="[ END ] [github.com/vmware/vic/cmd/vicadmin.logFiles] [3.905µs] " 
time="2016-08-31T00:00:38Z" level=debug msg="[BEGIN] [github.com/vmware/vic/cmd/vicadmin.(*server).handleFunc]" 
time="2016-08-31T00:00:38Z" level=debug msg="[ END ] [github.com/vmware/vic/cmd/vicadmin.(*server).handleFunc] [570ns] " 
time="2016-08-31T00:00:38Z" level=debug msg="[BEGIN] [github.com/vmware/vic/cmd/vicadmin.(*server).handleFunc]" 
time="2016-08-31T00:00:38Z" level=debug msg="[ END ] [github.com/vmware/vic/cmd/vicadmin.(*server).handleFunc] [701ns] " 
time="2016-08-31T00:00:38Z" level=debug msg="[BEGIN] [github.com/vmware/vic/cmd/vicadmin.(*server).handleFunc]" 
time="2016-08-31T00:00:38Z" level=debug msg="[ END ] [github.com/vmware/vic/cmd/vicadmin.(*server).handleFunc] [482ns] " 
[...]

Issue #2096, opened 2016-08-23
Unit test failure in vicadmin


mhagen-vmware:
Was this build before @cgtexmex restarted the build:
https://ci.vmware.run/vmware/vic/4290
```
time="2016-08-23T17:59:19Z" level=warning msg="Unable to connect: SDK URL () could not be parsed: %!s(<nil>)" 
time="2016-08-23T17:59:19Z" level=debug msg="[ END ] [github.com/vmware/vic/cmd/vicadmin.client] [12.52µs] " 
time="2016-08-23T17:59:19Z" level=debug msg="[ END ] [github.com/vmware/vic/cmd/vicadmin.findDatastore] [34.926µs] " 
time="2016-08-23T17:59:19Z" level=warning msg="SDK URL () could not be parsed: %!s(<nil>)" 
time="2016-08-23T17:59:19Z" level=debug msg="[ END ] [github.com/vmware/vic/cmd/vicadmin.configureReaders] [181.409µs] " 
2016/08/23 17:59:19 Testing TestLogTail
2016/08/23 17:59:19 GET https://localhost:36946/logs/tail/vicadmin.log:
time="2016-08-23T17:59:19Z" level=debug msg="[BEGIN] [github.com/vmware/vic/cmd/vicadmin.(*server).tailFiles]" 
time="2016-08-23T17:59:19Z" level=debug msg="[BEGIN] [github.com/vmware/vic/cmd/vicadmin.tailFile] ./vicadmin.log" 
time="2016-08-23T17:59:19Z" level=debug msg="[BEGIN] [github.com/vmware/vic/cmd/vicadmin.findSeekPos]" 
time="2016-08-23T17:59:19Z" level=debug msg="[ END ] [github.com/vmware/vic/cmd/vicadmin.findSeekPos] [8.665µs] " 
2016/08/23 17:59:19 Seeked ./vicadmin.log - &{Offset:0 Whence:0}
SIGQUIT: quit
PC=0x4a4a1a m=0

goroutine 95 [syscall, 9 minutes]:
syscall.Syscall6(0xe8, 0x13, 0xc4200efc14, 0x7, 0xffffffffffffffff, 0x0, 0x0, 0x0, 0x101020203030303, 0x0)
	/usr/local/go/src/syscall/asm_linux_amd64.s:44 +0x5 fp=0xc4200efb40 sp=0xc4200efb38
[...]

PR #2225, opened 2016-09-01
WIP: Use poll instead of inotify for tailing logs

anchal-agrawal:
There are known issues with the hpcloud/tail pkg that we're using for tailing vicadmin logs: 
hpcloud/tail#90
hpcloud/tail#21
hpcloud/tail#97

Using `poll` instead of `inotify` resolves the issue in some cases.

I can't repro #2096 locally. I'm using the WIP tag here to see if I can repro the issue on CI with this change. I'll update the PR once I've root-caused the issue.

Addresses #2096 #2203

### DIFF
--- a/cmd/vicadmin/logs.go
+++ b/cmd/vicadmin/logs.go
@@ -329,6 +329,7 @@ func tailFile(wr io.Writer, file string, done *chan bool) error {
 		Location:  &spos,
 		ReOpen:    true,
 		MustExist: false,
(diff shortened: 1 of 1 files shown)


anchal-agrawal:
Closing this PR as the issue was a race condition in the unit test itself (thanks @jzt!), not in the `hpcloud/tail` pkg. I'll open a new PR with the fix.

Issue #2210, opened 2016-08-31
Document workarounds for cross-platform installation of client plug-in


hickeng:
This applies to all of the UI plugin installation paths, but using this one as an example:
https://vmware.github.io/vic/assets/files/html/vic_installation/plugin_vcsa_no_web.html

The install script (install.sh or install.bat) is run on the client system where the VIC package was unpacked. This system could be linux, OSX, or Windows and that has no correlation with the type of vCenter server we're deploying to.

The documentation is written with the assumption that if you're deploying to a Windows based vCenter then the client system is Windows based and therefore we use `install.bat`, `SET`, et al.

stuclem:
Hi @hickeng. You're quite right that the doc assumes that if you're running a Windows VC then you need to run the Windows installer script on a Windows machine, and that if you're running VCSA then you need to run the Mac/Linux installer script on Mac/Linux. However, having played about a bit this morning, I think that the doc actually reflects the reality of the UI installer scripts as they are currently. Basically:

- If your VC is a VCSA, you have to update the `vic/ui/VCSA/configs` file with the VC info.
- If your VC is Windows-based, you have to update the `vic/ui/vCenterForWindows/configs` file with the VC info.
- The `vic/ui/VCSA/install.sh` script gets its VC info from the `vic/ui/VCSA/configs` file.
- The `vic/ui/vCenterForWindows/install.bat` script gets its VC info from the `vic/ui/vCenterForWindows/configs` file.
- The contents of the `vic/ui/VCSA/configs` and `vic/ui/vCenterForWindows/configs` and the `install.sh` and `install.bat` files are obviously different.

So, for example, you cannot point `install.bat` to `vic/ui/VCSA/configs`. I tested this by copying `install.bat` into the `vic/ui/VCSA/` folder and running it. I also tried configuring `vic/ui/VCSA/configs` with info about a VCSA that I deployed in Nimbus, ran `vic/ui/VCSA/install.sh` in a bash shell on Windows, and got the following error:
`install.sh: line 137: ../../vic-ui-linux: cannot execute binary file: Exec format error`

So, I edited `install.sh` to change 

```
if [[ $(echo $OS | grep -i "darwin") ]] ; then
    PLUGIN_MANAGER_BIN="../../vic-ui-darwin"
else
    PLUGIN_MANAGER_BIN="../../vic-ui-linux"
```

to 

```
if [[ $(echo $OS | grep -i "darwin") ]] ; then
[...]

Issue #2207, opened 2016-08-31
UI plugin install: confusing doc around running install.bat


hickeng:
Reading through the docs for how to install the UI plugin:
```
9. Run the installer.
       install.bat
    Make sure that you use the correct account to run install.bat.
       * If vCenter Server uses the local system account, run install.bat with the local system account.
       * If vCenter Server uses a different user account, run install.bat with that account.
10. Enter the user name and password for the vCenter Server administrator account.
```

This reads as if install.bat needs to be run on the vCenter system as the administrator account.
The note about using the correct account is associated with point 10, not point 9.

This is an issue in all of the plugin install sections

stuclem:
Removed the lines about the VC account from the "Run the installer" step in the topics on deployment to a Windows VC from :
https://github.com/stuclem/vic/blob/master/doc/user_doc/vic_installation/plugin_vc_no_sftp_or_web.md
https://github.com/stuclem/vic/blob/master/doc/user_doc/vic_installation/plugin_vc_sftp.md
https://github.com/stuclem/vic/blob/master/doc/user_doc/vic_installation/plugin_vc_web.md

This does not affect the topics on deployment to VCSA.

@hickeng and @jooskim, is this OK?

stuclem:
Incidentally, I am dealing with the cross-platform deployment issue in https://github.com/vmware/vic/issues/2210, so let's keep that discussion there.

stuclem:
Merged in https://github.com/vmware/vic/issues/2227.

Issue #1530, opened 2016-07-15
Document the location and content of the vicadmin portal


stuclem:
The vicadmin portal will include basic health status, logs, VCH information and resource usage. See PR https://github.com/vmware/vic/pull/1515, and epic https://github.com/vmware/vic/issues/1372.

stuclem:
https://github.com/vmware/vic/issues/1372 is for GA.

stuclem:
PR https://github.com/vmware/vic/pull/1515 is now merged.

stuclem:
The VCH Admin portal is visible at https://`vch_address`:2378 after you deploy a VCH. 

![vicadmin](https://cloud.githubusercontent.com/assets/16718369/17250941/d1874f28-55a6-11e6-84e6-f8cd9a25d151.png)

stuclem:
@kjplatz and @sgairo, I see in the screencap above that my network connectivity is showing up as orange. I assume that this means that these squares can be green (OK), orange (something's dodgy), or red (it's dead). It would be good if in the docs we could point people to some clues as to what they need to do to make their squares green, if some of them are orange/red. 

So, are there any design docs anywhere that you can point us to that list the conditions under which each of these statuses appears as either orange or red? 

Thanks!

stuclem:
https://github.com/vmware/vic/issues/1744 will change the output of vic-machine create to include the vicadmin address (currently says `Log Server`). Requires an update in the topic on deploying a VCH.

sgairo:
@stuclem We are currently still working out what categorizes a status as green/orange/red for each health report. We will surely send it your way once we have it.

stuclem:
@sgairo thanks!

stuclem:
https://github.com/vmware/vic/issues/1744 is merged in PR https://github.com/vmware/vic/issues/1811. Install output now says `vic-admin portal` instead of `Log server`.

stuclem:
Marking this one as done. Will include info about what turns statuses from green to yellow in https://github.com/vmware/vic/issues/1317.

PR #2227, opened 2016-09-01
[skip ci] More doc fixes for 0.5.5

stuclem:
Fixes https://github.com/vmware/vic/issues/2210, https://github.com/vmware/vic/issues/2207, is a start on https://github.com/vmware/vic/issues/1530.


stuclem:
This LGTM, but I'd appreciate @jooskim taking a look too. Thanks!

jooskim:
Step 9 in https://github.com/stuclem/vic/blob/master/doc/user_doc/vic_installation/plugin_vcsa_web.md contains a duplicate sentence "Do not use Windows 10 native Bash."

jooskim:
Step 5 in https://github.com/stuclem/vic/blob/master/doc/user_doc/vic_installation/plugin_vcsa_web.md and https://github.com/stuclem/vic/blob/master/doc/user_doc/vic_installation/plugin_vc_web.md tells the user to enter the URL of the "ZIP" file which is not correct.

It should just be the folder where the ZIP file is, and the actual zip filename will be populated during the installation process.

jooskim:
Other than those two, LGTM.

doc/user_doc/vic_installation/plugin_vc_sftp.md
```diff
@@ -6,10 +6,11 @@ If your vCenter Server instance runs on a Windows system that has access to an S
 
 - You deployed at least one virtual container host to a vCenter Server instance that runs on Windows.
 - Your vCenter Server instance is running on a Windows system that has an SFTP service.
+- You must use a Windows system to run the script to install the plug-in on a vCenter Server that runs on Windows. If you used a Linux or Mac OS system to deploy the virtual container host, download and unpack the vSphere Integrated Containers Engine package on a Windows system.
```
hickeng: @jooskim Is this correct? I can see that there's only a `.bat` file in the installer/vCenterForWindows directory however it would seem that the type of the target vCenter is the important criteria, not what OS we're running the script on.

jooskim: @hickeng Yes. I talked with Stuart about this after he tried running the `.bat` file in Linux using Wine to no avail. And the configurations will not work as expected under whatever is the choice for the Windows cmd emulator. I see that Wine is available for both Linux and macOS, but I'm not sure if we want the users to install it for this. Plus that, I need to verify the `install.bat` works in the emulator as well.

jooskim:
I played with Wine in Ubuntu and couldn't get the `install.bat` script to work properly using `wineconsole`, because the command prompt did not work the same way I expected a real `cmd` would; the script did not register the user variables defined inside the `configs` file, as the Wine Command Prompt did not understand some of the syntaxes used in the `install.bat` file. This relates to Stuart's reporting that the script complained about an empty IP address. `wineconsole` also did not have Windows Powershell which is needed to take user input for vCenter password.

stuclem:
:recycle:

stuclem:
Thanks @jooskim, I made those two fixes. The other comments in this PR relate to future work, rather than to the current version of the UI installer, so merging.

PR #2229, opened 2016-09-01
[full ci] Allow custom scratch size (revert revert)

gigawhitlocks:
This change got reverted due to a failing test. Apparently one of the helper methods I was using got updated upstream and I didn't notice during my rebase (the PR sat and rotted in the queue while we were having CI issues), which was causing the new integration test I wrote to fail upon merge to master.

Old PR was #1982


cgtexmex:
LGTM

dougm:
lgtm

PR #2230, opened 2016-09-01
Unregister container during removal if state invalid

### LINKED ISSUE #1823: Error response from daemon: server error from portlayer : [DELETE /containers/{id}][500] containerRemoveInternalServerError

jzt:
This change allows a container VM to be unregistered during `ContainerRm` in the event that `stop` left it in an invalid state.

I used a script to get a container into an invalid state:
```bash
#!/bin/bash
while :
do
docker -H 192.168.218.235:2376 --tls create --name test busybox
docker -H 192.168.218.235:2376 --tls start test
docker -H 192.168.218.235:2376 --tls stop test
docker -H 192.168.218.235:2376 --tls rm test
# once state is invalid, we will have a non-zero exit status
if [ $? -ne 0 ]
then
        exit $STATUS
fi
done
```

Once the fix was added, I ran the script again for 30 minutes without issue. Containers with invalid states were successfully removed:

```
root@vic-docker [ /var/log/vic ]# grep "invalid state" port-layer.log
time="2016-05-22T03:37:48Z" level=debug msg="container VM is in invalid state, unregistering"
time="2016-05-22T03:41:29Z" level=debug msg="container VM is in invalid state, unregistering"
```

Fixes #1823

### DIFF
--- a/lib/portlayer/exec/container.go
+++ b/lib/portlayer/exec/container.go
@@ -421,6 +421,7 @@ func (r RemovePowerError) Error() string {
 	return r.err.Error()
 }
 
+// Remove removes a containerVM after detaching the disks
 func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
 	defer trace.End(trace.Begin("Container.Remove"))
 	c.Lock()
@@ -466,8 +467,23 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
 		return c.vm.DeleteExceptDisks(ctx)
 	})
 	if err != nil {
-		c.State = existingState
-		return err
+		f, ok := err.(types.HasFault)
+		if !ok {
+			c.State = existingState
+			return err
+		}
+		switch f.Fault().(type) {
+		case *types.InvalidState:
+			log.Warnf("container VM is in invalid state, unregistering")
+			if err := c.vm.Unregister(ctx); err != nil {
+				log.Errorf("Error while attempting to unregister container VM: %s", err)
+				return err
+			}
+		default:
+			log.Debugf("Fault while attempting to destroy vm: %#v", f.Fault())
+			c.State = existingState
+			return err
+		}
 	}
 
 	// remove from datastore


jzt:
Tagging @hickeng

lib/portlayer/exec/container.go
```diff
@@ -451,8 +452,16 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
 		return c.vm.DeleteExceptDisks(ctx)
 	})
 	if err != nil {
-		c.State = existingState
-		return err
+		if f, ok := err.(types.HasFault); ok {
+			switch f.Fault().(type) {
+			case *types.InvalidState:
+				log.Debugf("container VM is in invalid state, unregistering")
+				unregisterContainerVM(c, ctx)
```
hmahmood: Not checking for error here?

lib/portlayer/exec/container.go
```diff
@@ -451,8 +452,16 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
 		return c.vm.DeleteExceptDisks(ctx)
 	})
 	if err != nil {
-		c.State = existingState
-		return err
+		if f, ok := err.(types.HasFault); ok {
+			switch f.Fault().(type) {
+			case *types.InvalidState:
+				log.Debugf("container VM is in invalid state, unregistering")
```
hickeng: bump this to a Warnf as it's definitely not standard workflow

lib/portlayer/exec/container.go
```diff
@@ -451,8 +452,16 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
...
+				unregisterContainerVM(c, ctx)
+			}
+		} else {
```
hickeng: put this clause first so we don't get the if/else nesting

lib/portlayer/exec/container.go
```diff
@@ -451,8 +452,16 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
 		return c.vm.DeleteExceptDisks(ctx)
 	})
 	if err != nil {
-		c.State = existingState
-		return err
+		if f, ok := err.(types.HasFault); ok {
+			switch f.Fault().(type) {
```
hickeng: can we have a `default:` target that logs the fault type - currently we lose it completely.

hickeng:
@jzt can you confirm what happens when this is run against VC, particularly when there's volumes attached to the VM. I'm wondering whether fm.Delete cleans up the disk chain in the case where DeleteExceptDisks didn't succeed.

lib/portlayer/exec/container.go
jzt: Fixed

lib/portlayer/exec/container.go
jzt: D'doy... thanks, fixed

lib/portlayer/exec/container.go
```diff
@@ -451,8 +452,21 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
...
+		if !ok {
+			c.State = existingState
+			return err
+		}
+		switch f.Fault().(type) {
+		case *types.InvalidState:
+			log.Warnf("container VM is in invalid state, unregistering")
+			if err := c.vm.Unregister(ctx); err != nil {
+				log.Errorf("Error while attempting to unregister container VM: %s", err)
+				return err
```
hickeng: If we see an error here can we retry the delete now c.State == StateRemoving? I don't want us to end up with a zombie container where we cannot retry the operation.

lib/portlayer/exec/container.go
```diff
@@ -451,8 +452,21 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
...
+			}
+		default:
+			log.Debugf("Fault while attempting to destroy vm: %#v", f.Fault())
```
hickeng: Not sure what Faults could apply here - should we do the follow as in the non fault error case?:
```
c.State = existingState
return err
```

jzt:
@hickeng So I created a volume on VC (skinner) and I attached it to a container I made, and then I unregistered the container by way of the VC gui. The volume stayed around, and the image vmdk's were left alone.

The only way the `Remove` path will continue when `DeleteExceptDisks` returns an error is if it's an `InvalidState` error, in which case the unregister followed by `DeleteDatstoreFile` functions are called. Any other error will be returned from `Remove` and `DeleteDatastoreFile` will not be called.

In order to ensure that a govmomi call to `UnregisterVM` doesn't affect the disks/volumes, I'm going to run a test by forcing the invalid state (as mentioned in the PR description) and I'll report back.

lib/portlayer/exec/container.go
jzt: Yeah, that's probably a safe bet.

lib/portlayer/exec/container.go
jzt: There's no logic preventing attempted removal of a container whose current state is `StateRemoving`. Would seeing the container state as `Removing` in the output of `docker ps` be useful? If we reset it to `existingState`, it may not be evident that something went wrong (or is wrong with the container). What do you think?

jzt:
@hickeng I've been running this script on VC for almost an hour straight and haven't been able to get the container into an invalid state. Is it possible this bug doesn't surface within VC? @hmahmood did you mention this was a VC bug?

mhagen-vmware:
This bug really appears in the CI, which so far is still only nested ESX

hmahmood:
@jzt I am only aware of it happening in the CI, as @mhagen-vmware points out.

lib/portlayer/exec/container.go
hickeng: I don't think we should reset the state. We don't currently know at which point in DeleteExceptDisks we failed. It's possible the VM doesn't have a root disk any more if they were successfully detached.
The case we're attempting to handle (InvalidState due to guestinfo update during shutdown) should mean that the failure is all or nothing, but unless we rework DeleteExceptDisks to allow us to be sure of that we shouldn't assume it here.
Stick with the current Removing state.

hickeng:
@jzt This should impact both ESX and VC. The concern with volumes and images isn't the unregister, it's the follow on operation where we delete the files. @fdawg4l was going to verify what filemanager delete does in the VC disk chain case which is what I was primarily concerned about.
I *think* we're okay, but I'd like validation.

jzt:
@hickeng:

I can show that if `DeleteExceptDisks` successfully unmounts the disks, but fails to destroy the VM, the volumes are preserved on disk (I started to describe it here with logs and info but it was way long, ask me if you'd like to see it).

The next question I have is this:

```go
// DeleteExceptDisks destroys the VM after detaching all virtual disks
func (vm *VirtualMachine) DeleteExceptDisks(ctx context.Context) (*object.Task, error) {
	devices, err := vm.Device(ctx)
	if err != nil {
		return nil, err
	}

	disks := devices.SelectByType(&types.VirtualDisk{})

	err = vm.RemoveDevice(ctx, true, disks...)
	if err != nil {
		return nil, err
	}

	return vm.Destroy(ctx)
}
```

What happens here if/when either `vm.Device()` or `vm.RemoveDevice()` manage to return an `InvalidState` fault (if that's even possible)? This would imply that the disks were not unmounted successfully. Returning an `InvalidStateFault` would not reset container state and return err, but would follow the `UnregisterVM` path, followed by `fm.DeleteDatastoreFile`.

According to the API docs, `UnregisterVM` doesn't delete anything from disk.

Does the call to `fm.DeleteDatastoreFile(ctx, dsPath, sess.Datacenter)` delete the volumes? Does `dsPath` simply contain the path to the VM directory in the datastore?

hickeng:
@jzt It shouldn't, unless VC is doing something 'intelligent' on our behalf; you can delete a disk on ESX without the disk chain being cleaned, but the same operation via VC cleans the entire chain (don't recall specifically which method that was - think it was one of the Datastore ones).
As mentioned @fdawg4l was going to confirm the FileManager behaviour as part of the `rmi` implementation so I suggest opening an issue to check it, assign it to him and then this PR can be merged given we have a followup if needed.

jzt:
@hickeng With help from @emlin, I was able to run my own test on both VC and standalone ESX using `govc datastore.rm`, which also uses `DeleteDatastoreFile_Task` to delete files/directories from the datastore. I pulled `golang` and found the child-most image layer by way of the `parentMap`, and deleted its vmdk and delta disks and all the parent image vmdks were still present and unaffected.

`fa28a7` is the child-most image layer:
```
$ govc datastore.ls zach-vch1/VIC/422334c0-4cc6-a35b-5917-5fe25bd4e4aa/images/fa28a7d8725fc6adc5b24643527f350cbe029b2a1d3943cbbe1a29fd163b26e3
imageMetadata
fa28a7d8725fc6adc5b24643527f350cbe029b2a1d3943cbbe1a29fd163b26e3.vmdk
manifest
fa28a7d8725fc6adc5b24643527f350cbe029b2a1d3943cbbe1a29fd163b26e3-delta.vmdk
$ govc datastore.rm zach-vch1/VIC/422334c0-4cc6-a35b-5917-5fe25bd4e4aa/images/fa28a7d8725fc6adc5b24643527f350cbe029b2a1d3943cbbe1a29fd163b26e3/
$ govc datastore.ls zach-vch1/VIC/422334c0-4cc6-a35b-5917-5fe25bd4e4aa/images/fa28a7d8725fc6adc5b24643527f350cbe029b2a1d3943cbbe1a29fd163b26e3
govc: File [DatastoreSSD] zach-vch1/VIC/422334c0-4cc6-a35b-5917-5fe25bd4e4aa/images/fa28a7d8725fc6adc5b24643527f350cbe029b2a1d3943cbbe1a29fd163b26e3 was not found
$ govc datastore.ls zach-vch1/VIC/422334c0-4cc6-a35b-5917-5fe25bd4e4aa/images/
c0276f282936a2e39006d976ffee6d80764fb57655caadfe75659114121bd23e
a63c621debe7286555cb07d1b5b55fc5d8d15a2d438a7a60457e036b2781d4dd
scratch
993f155600d89074fc8932e252bd4c85401f5e1740e030b78d39649f533c5d08
e80cf2815218a1495d82df64d63432e80dca9a9de4cb1689c76ca69cc05d2d0b
f825c53868c695b6ea639efe612a92f08e8ca991e0c55125525c14dfd7f9ae30
c289280e4ecc29506dea24a1f11222d427b187a49a0da7046ea788b01ad39440
58dad7e7831979e92340ee7cca0df19477f3091ab3072daf8a5a3f5c3f34c290
7a252027522adaffb432fbd0a7eaa778238cbde9df5f2271f74e97c4a1084560
98052a526e4e2b77a0d5c49f734bea79c202ee7945c63fa9024016a41cbb5c3c
f89af82b270bc5876b493fcb5e6bf8608fad59865b22124821375d4e8f1675e7
ea5fa6b84e55265127c0cb538e030010cbd06f739de2c827c76ec2e57b5838c6
307524089b335f318b3c2ab298b39cf0328644271f2ee4456f806f219b658841
396b505407b4b46d60943127d3e0e347aa68d4b23d6e1aa03973185c56a835a9
```
A peek at random parent images shows that their vmdks are still alive:
```
$ govc datastore.ls zach-vch1/VIC/422334c0-4cc6-a35b-5917-5fe25bd4e4aa/images/58dad7e7831979e92340ee7cca0df19477f3091ab3072daf8a5a3f5c3f34c290
imageMetadata
manifest
58dad7e7831979e92340ee7cca0df19477f3091ab3072daf8a5a3f5c3f34c290.vmdk
58dad7e7831979e92340ee7cca0df19477f3091ab3072daf8a5a3f5c3f34c290-delta.vmdk
```

Would you consider this adequate to unblock this merge, or do we need more definitive proof, like API docs or a lifecycle expert signing off?

emlin:
lgtm

hickeng:
@jzt LGTM - that is sufficient. It aligns with expectations, but wanted confirmation.

jzt:
♻️

fdawg4l:
@jzt are you using the parentmap directly? If so, we need to augment the api so you don't need to. IIRC GetImage should do that for you as is, but we can chat about it offline.

PR #2231, opened 2016-09-01
Fix version.ShortVersion formatting

caglar10ur:
now shows
```
Server Version: v0.5.5-c775037
```
instead of
```
Server Version: v0.5.5--c775037
```
when the BuildNumber is omitted

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -75,5 +75,8 @@ func (v *Build) String() string {
 }
 
 func (v *Build) ShortVersion() string {
+	if v.BuildNumber == "" {
+		v.BuildNumber = "0"
+	}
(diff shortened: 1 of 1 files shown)


cgtexmex:
LGTM

dougm:
Can we make build number '0' in that case?
```
v0.5.5-0-c775037
```

Just to have the same format in either case, should a script want to parse the version.  But if we don't care about that, then LGTM

emlin:
> Can we make build number '0' in that case?
> 
> v0.5.5-0-c775037
> 
> Just to have the same format in either case, should a script want to parse the version. 

 +1 with it.
vic-machine upgrade will check version, and assume the same format.

caglar10ur:
Sure thing, will update the PR with 0 as the placeholder

emlin:
lgtm

caglar10ur:
♻️

Issue #1642, opened 2016-07-25
generate /etc/mtab in container


hickeng:
Currently we do not construct /etc/mtab in the container so the df command does not function:
VIC:
```
root@ae7b8f307ee3:/# cat /etc/mtab
cat: /etc/mtab: No such file or directory
root@ae7b8f307ee3:/#
```

docker:
```
root@a1722c1e6629:/# cat /etc/mtab
none / aufs rw,relatime,si=f1cde4db423deea6,dio,dirperm1 0 0
proc /proc proc rw,nosuid,nodev,noexec,relatime 0 0
tmpfs /dev tmpfs rw,nosuid,mode=755 0 0
devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=666 0 0
sysfs /sys sysfs ro,nosuid,nodev,noexec,relatime 0 0
tmpfs /sys/fs/cgroup tmpfs ro,nosuid,nodev,noexec,relatime,mode=755 0 0
cgroup /sys/fs/cgroup/systemd cgroup ro,nosuid,nodev,noexec,relatime,xattr,release_agent=/lib/systemd/systemd-cgroups-agent,name=systemd,nsroot=/docker/a1722c1e66292e36453918865d643130db128c8e7f256b085e743d83f712e46b 0 0
cgroup /sys/fs/cgroup/perf_event cgroup ro,nosuid,nodev,noexec,relatime,perf_event,nsroot=/docker/a1722c1e66292e36453918865d643130db128c8e7f256b085e743d83f712e46b 0 0
cgroup /sys/fs/cgroup/net_cls,net_prio cgroup ro,nosuid,nodev,noexec,relatime,net_cls,net_prio,nsroot=/docker/a1722c1e66292e36453918865d643130db128c8e7f256b085e743d83f712e46b 0 0
cgroup /sys/fs/cgroup/cpu,cpuacct cgroup ro,nosuid,nodev,noexec,relatime,cpu,cpuacct,nsroot=/docker/a1722c1e66292e36453918865d643130db128c8e7f256b085e743d83f712e46b 0 0
cgroup /sys/fs/cgroup/memory cgroup ro,nosuid,nodev,noexec,relatime,memory,nsroot=/docker/a1722c1e66292e36453918865d643130db128c8e7f256b085e743d83f712e46b 0 0
[...]

PR #2232, opened 2016-09-02
Revert: symlink /proc/mounts to /etc/mtab

hmahmood:
I left this out in my previous commit for the tether unit test fixes. The `df` seems to work without the symlink though -- something that this was supposed to fix. See #1642 . This only needs to happen in the container VM. Same goes for unmounting `/run`.


dougm:
lgtm

hmahmood:
@caglar10ur @hickeng could you please review?

anchal-agrawal:
LGTM

hmahmood:
♻️

PR #2243, opened 2016-09-03
Add DeleteImage to storage image cache

fdawg4l:
This change implements image deletion from the image cache.  The vsphere impl which actually removes it from the datastore (after checking if it is attached) will come in next.  

Only allows deletes of images which have no children.  Deletes of
unused branches will come in at a later date if at all.

### DIFF
--- a/lib/portlayer/storage/image_cache.go
+++ b/lib/portlayer/storage/image_cache.go
@@ -15,6 +15,7 @@
 package storage
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"net/url"
@@ -31,6 +32,8 @@ import (
 var Scratch = Image{
 	ID: "scratch",
 }
+
+var ErrCorruptImageStore = errors.New("Corrupt image store")
 
 // NameLookupCache the global view of all of the image stores.  To avoid unnecessary
 // lookups, the image cache keeps an in memory map of the store URI to the map
@@ -68,7 +71,7 @@ func (c *NameLookupCache) GetImageStore(ctx context.Context, storeName string) (
 	_, ok := c.storeCache[*store]
 
 	if !ok {
-		log.Info("ImageCache: Refreshing image cache from datastore.")
+		infof("Refreshing image cache from datastore.")
 		// Store isn't in the cache.  Look it up in the datastore.
 		storeName, err := util.ImageStoreName(store)
 		if err != nil {
@@ -89,7 +92,7 @@ func (c *NameLookupCache) GetImageStore(ctx context.Context, storeName string) (
 		scratch, err := c.DataStore.GetImage(ctx, store, Scratch.ID)
 		if err != nil {
 			log.Errorf("ImageCache Error: looking up scratch on %s: %s", store.String(), err)
-			return nil, fmt.Errorf("Scratch does not exist.  Imagestore is corrrupt.")
+			return nil, ErrCorruptImageStore
 		}
 
 		if err = indx.Insert(scratch); err != nil {
@@ -306,7 +309,8 @@ func (c *NameLookupCache) DeleteImage(ctx context.Context, image *Image) error {
 	// Check the image exists.  This will rehydrate the cache if necessary.
 	_, err := c.GetImage(ctx, image.Store, image.ID)
 	if err != nil {
-		return errorf("DeleteImage: %s", err)
+		errorf("DeleteImage: %s", err)
+		return err
 	}
 
 	// get the relevant cache
@@ -316,21 +320,24 @@ func (c *NameLookupCache) DeleteImage(ctx context.Context, image *Image) error {
 
 	hasChildren, err := indx.HasChildren(image.Self())
 	if err != nil {
-		return errorf("DeleteImage: %s", err)
+		errorf("DeleteImage: %s", err)
+		return err
 	}
 
 	if hasChildren {
-		return errorf("deleting image %s: in use by child images", image.Self())
+		return &ErrImageInUse{image.Self() + " in use by child images"}
 	}
 
 	// The datastore will tell us if the image is attached
 	if err = c.DataStore.DeleteImage(ctx, image); err != nil {
-		return errorf("%s", err)
+		errorf("%s", err)
+		return err
 	}
 
 	// Remove the image from the cache
 	if _, err = indx.Delete(image.Self()); err != nil {
-		return errorf("%s", err)
+		errorf("%s", err)
+		return err
 	}
 
 	return nil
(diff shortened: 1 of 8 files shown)


lib/portlayer/storage/image_cache_test.go
```diff
@@ -306,3 +313,89 @@ func TestImageStoreRestart(t *testing.T) {
...
+		return
+	}
+
+	scratch, err := imageCache.GetImage(ctx, storeURL, Scratch.ID)
+	if !assert.NoError(t, err) {
+		return
+	}
+
+	// create a 3 level tree with 9 branches
+	branches := 4
```
fdawg4l: 4 branches, not 9.

lib/portlayer/storage/image_cache.go
```diff
@@ -68,7 +68,7 @@ func (c *NameLookupCache) GetImageStore(ctx context.Context, storeName string) (
 	_, ok := c.storeCache[*store]
 
 	if !ok {
-		log.Info("Refreshing image cache from datastore.")
+		log.Info("ImageCache: Refreshing image cache from datastore.")
```
hickeng: Should be using the `infof` function?

lib/portlayer/storage/image_cache.go
```diff
@@ -88,7 +88,7 @@ func (c *NameLookupCache) GetImageStore(ctx context.Context, storeName string) (
 		// Add Scratch
 		scratch, err := c.DataStore.GetImage(ctx, store, Scratch.ID)
 		if err != nil {
-			log.Errorf("Error looking up scratch on %s: %s", store.String(), err)
+			log.Errorf("ImageCache Error: looking up scratch on %s: %s", store.String(), err)
```
hickeng: should be using the `errorf` function?

lib/portlayer/storage/image_cache.go
```diff
@@ -300,3 +298,54 @@ func (c *NameLookupCache) ListImages(ctx context.Context, store *url.URL, IDs []
 
 	return imageList, nil
 }
+
+// DeleteImage deletes an image from the image store.  If it is in use or is being inheritted from, then this will return an error.
+func (c *NameLookupCache) DeleteImage(ctx context.Context, image *Image) error {
+	infof("DeleteImage: deleting %s", image.Self())
```
hickeng: Is there a reason not to use the entry/exit trace here for consistency with the rest of the logging?

lib/portlayer/storage/image_cache.go
```diff
@@ -300,3 +298,54 @@ func (c *NameLookupCache) ListImages(ctx context.Context, store *url.URL, IDs []
...
+
+	// Check the image exists.  This will rehydrate the cache if necessary.
+	_, err := c.GetImage(ctx, image.Store, image.ID)
+	if err != nil {
+		return errorf("DeleteImage: %s", err)
```
hickeng: If `err` is ErrNotExist, should we be preserving that so callers can map errors to detailed responses.

lib/portlayer/storage/image_cache.go
```diff
@@ -300,3 +298,54 @@ func (c *NameLookupCache) ListImages(ctx context.Context, store *url.URL, IDs []
...
+	}
+
+	// get the relevant cache
+	c.storeCacheLock.Lock()
+	indx := c.storeCache[*image.Store]
+	c.storeCacheLock.Unlock()
+
+	hasChildren, err := indx.HasChildren(image.Self())
+	if err != nil {
+		return errorf("DeleteImage: %s", err)
```
hickeng: Same comment as for ErrNotExist - how does a caller distinguish between "in use", "not found", "permission denied", or the fallback "internal error"?

lib/portlayer/storage/image_cache.go
```diff
@@ -300,3 +298,54 @@ func (c *NameLookupCache) ListImages(ctx context.Context, store *url.URL, IDs []
...
+	}
+
+	if hasChildren {
+		return errorf("deleting image %s: in use by child images", image.Self())
+	}
+
+	// The datastore will tell us if the image is attached
```
hickeng: Only true if the VM with it attached is powered on, unless there's logic in the datastore call?

lib/portlayer/storage/image_cache.go
fdawg4l: Done.

lib/portlayer/storage/image_cache.go
fdawg4l: Done.  i still want a separate log line with the error received since that includes the vsphere error.

lib/portlayer/storage/image_cache.go
fdawg4l: Do you mean in addition to the `infof` above or in place of?  

I can add trace calls in a later PR.  You'll notice it's not being used anywhere in this package.

lib/portlayer/storage/image_cache.go
fdawg4l: Yeah, you're right.  Damn.  I really hate errors in go.

lib/portlayer/storage/image_cache.go
fdawg4l: In the vsphere image store implementation (`c.DataStore`), that check will be made in the `Delete` call.  This is just a cache and I don't want to add logic which is specific to vsphere here.

The `Delete` call in the `DataStore` will come in my next PR.  Lots of testing is required there and its best to keep these PRs separate.

hickeng:
If we add in your WIP for contextual logging and call stacks does it change how you'd approach the infof, debugf functions?

fdawg4l:
Including it would make the change kind of huge.  We'd need to change the storage handlers on the REST side to pass this new context implementation and plumb them all the way down to the `govmomi` calls to be useful.

As far as changing the approach-  not really.  The context WIP change is to help debug from the client side where the operation went belly up.  The logging here is help debug why they went belly up from the point of view of the PL.

I just need to clean up / omit the errorf function in order for the REST API to type assert the appropriate error.

lib/portlayer/storage/image_cache.go
fdawg4l: in use -  will need to define that error.  Will update this change with a new error struct for that use case if it doesnt already exist in the `io` package.

not found - will use the `io` package's variant.

permission denied - we don't support mode bits or uid/gid of vmdk images.  

internal error - aren't all undefined errors internal errors?

fdawg4l: > in use

Looks like there is no EBUSY in the `os` package.  Lame.

matthewavery:
lgtm

fdawg4l:
Poke @hickeng.  I think i addressed all of your comments.  Please double check.

hickeng:
LGTM - turns out that plugin doesn't recognise the comments from the new github review process.

fdawg4l:
.

Issue #2245, opened 2016-09-06
VCH appliance prevents the host from entering maintenance mode


mhagen-vmware:
After installing VCH into a nested ESX host environment, running:

govc host.maintenance.enter

now hangs at 100% and doesn't complete.  When I ctrl-c out of the command, it causes the host to be in some sort of invalid state that I cannot exit or leave maintenance mode nor do anything with VMs.  Power cycling the host recovers and the host ends up in maintenance mode when it comes back up.

VIC support bundle:
[container-logs.zip](https://github.com/vmware/vic/files/457052/container-logs.zip)

ESX bundle:

[esx-bundle.zip](https://github.com/vmware/vic/files/457067/esx-bundle.zip)

cgtexmex:
In the case of a standalone host you must first power down the VCH (or any other VMs) before entering maintenance mode.  The govc library could be improved to handle this more gracefully, but it's not an issue w/vic or the VCH.

In a clustered vSphere environment with DRS set to automatic I was able to put a host into maintenance mode which hosted a running VCH.  DRS will first migrate the VCH to another host in the cluster and then enter maintenance mode.  Additionally, I tested with a host that was hosting a running container VM -- same result a successful migration and entry to maintenance mode.

If we want to address the govc improvement then we should open an issue against that project.

stuclem:
Thanks @cgtexmex. We will include the above info in the doc, as part of https://github.com/vmware/vic/issues/864.

Issue #2244, opened 2016-09-06
Clean up error when installing VIC into maintenance mode


mhagen-vmware:
INFO[2016-09-06T10:02:28-05:00] Collecting ha-host hostd.log                 
ERRO[2016-09-06T10:02:28-05:00] --------------------                         
ERRO[2016-09-06T10:02:28-05:00] vic-machine-linux failed: Creating the appliance failed with Operation failed: The operation is not allowed in the current state. - (&{{} %!s(*types.InvalidState=&{{{<nil> []}}}) The operation is not allowed in the current state.}). Exiting...

The host is in maintenance mode in this scenario, it would be nice if we could provide a more useful error indicating as such.

PR #2247, opened 2016-09-06
Initial maintenance mode test plan [skip ci]

### LINKED ISSUE #860: host maintenance mode
This is primarily a testing exercise as the implementation is dependent on #770. This is pulled out as a separate issue because of how important it is that we not inhibit this pervasive day2 operation.

This test asserts that a host, with containerVMs on it, can be put into maintenance mode without manual intervention. There *will* be a case where this is problematic - the case where there is an attach operation in progress to a containerVM on that host.

There is one implementation component to it that may or may not be viable:
* if there is an attach session open to a containerVM on the quiescing host and
* the applianceVM can receive a notification about the hosts change of status
  * then we can consider dropping the attach connection to allow the containerVM to vMotion.

This is getting very close to the behaviour implemented via the vSPC and if we wish to support this we should seriously consider doing so leveraging that mechanism (see the [communications document](https://github.com/vmware/vic/blob/master/doc/design/communications.md#vspc-relay) for vSPC details - there is scope for implementing it directly in the attach server rather than as an agent).

mhagen-vmware:
Just some initial support for maintenance mode testing, we currently do not support this yet and I have filed the following issues related to it:
#2245 
#2244 
Fixes #860

### DIFF
--- a/tests/manual-test-cases/Group12-Maintenance-Mode/12-1-VIC-Machine-Maintenance-Mode.md
+++ b/tests/manual-test-cases/Group12-Maintenance-Mode/12-1-VIC-Machine-Maintenance-Mode.md
@@ -1,0 +1,26 @@
+Test 12-1 VIC Machine Maintenance Mode
+=======
+
+#Purpose:
+To verify the VIC appliance provides a reasonable error message when installing or deleting it from a host in maintenance mode
+
+#References:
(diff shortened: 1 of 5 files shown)


rajanashok:
LGTM

matthewavery:
LGTM

PR #2251, opened 2016-09-06
Initial vMotion test plan [skip ci]

### LINKED ISSUE #2250: Write vMotion for VIC test plan
Need to write a new test plan to test how VIC behaves in the face of vMotion happening.

mhagen-vmware:
Fixes #2250


mdubya66:
Good first step. LGTM.

dougm:
lgtm

mhagen-vmware:
.

hickeng:
@mhagen-vmware couple of items of note
1. It's not clear to me whether we're installing 3 concurrent VCHs, or cleaning up prior to subsequent install.
2. The initial test (powered off vch) isn't testing vmotion, but is testing some of VCH restart.
3. We should be specific about which commands are crucial in expected results. Saying all docker commands complete without error is already wider than viable unless we link a doc of supports commands and sub-options.
We should confirm attach and logs as a priority after the vmotion.
4. We should confirm that a container that has previously been attached to can be vmotioned after the fact.
5. We should confirm our current behaviour that a container with an _active_ attach connection cannot be vmotioned.

mhagen-vmware:
Thanks @hickeng #2270

PR #1815, opened 2016-08-04
Show server and build info in docker version output

anchal-agrawal:
Populates the build time, commit and version fields for the server portion of the `docker version` output

Server portion output:

Vanilla docker
```
Server:
 Version:      1.11.2
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   b9f10c9
 Built:        Wed Jun  1 22:00:43 2016
 OS/Arch:      linux/amd64
```
VIC
```
Server:
 Version:      0.4.0
 API version:  1.23
 Go version:   go1.6.3
 Git commit:   5622735
 Built:        Thu Aug  4 17:14:53 2016
 OS/Arch:      linux/amd64
 Experimental: true
```
Fixes #856

PR #2257, opened 2016-09-06
[full ci] Fix the wrong version info in vApp

### LINKED ISSUE #2158: vApp showing VIC version as 0.0.1 not 0.5.0
image

This file seems the one where the version is set:
https://github.com/vmware/vic/blob/af71181d82f0bff404f7109cd3176ed0ecc21ca7/lib/install/management/virtual_app.go

jooskim:
This simple PR fixes #2158 in that the version of vApp in vSphere Web Client currently showing 0.0.1 is corrected to 0.5.5.

After fix:
image

### DIFF
--- a/lib/install/management/virtual_app.go
+++ b/lib/install/management/virtual_app.go
@@ -24,6 +24,7 @@ import (
 	"github.com/vmware/vic/lib/install/data"
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/version"
 )
 
 func (d *Dispatcher) createVApp(conf *config.VirtualContainerHostConfigSpec, settings *data.InstallerData) (*object.VirtualApp, error) {
@@ -79,7 +80,7 @@ func (d *Dispatcher) createVApp(conf *config.VirtualContainerHostConfigSpec, set
 			Name:      "vSphere Integrated Containers",
 			Vendor:    "VMware",
 			VendorUrl: "http://www.vmware.com/",
-			Version:   "0.0.1",
+			Version:   version.Version,
 		},
 		ArrayUpdateSpec: types.ArrayUpdateSpec{
 			Operation: types.ArrayUpdateOperationAdd,


lib/install/management/virtual_app.go
```diff
@@ -79,7 +79,7 @@ func (d *Dispatcher) createVApp(conf *config.VirtualContainerHostConfigSpec, set
 			Name:      "vSphere Integrated Containers",
 			Vendor:    "VMware",
 			VendorUrl: "http://www.vmware.com/",
-			Version:   "0.0.1",
+			Version:   "0.5.5",
```
anchal-agrawal: Using the `Version` var from the `version` pkg would be a better alternative to hardcoding the value. https://github.com/vmware/vic/blob/master/pkg/version/version.go#L25

jooskim: Great. Thanks for letting me know.

sgairo: I agree about the version package, however, it currently doesn't show the correct version. Right now it shows as 0.4.0, so you might not always have the most up-to-date version by using the package.

jooskim: Hmm.. I am building vic-machine after making the change to see if it shows the right version. I have a couple older binaries on my laptop and I just tried running vic-machine with build 4430 and it seems to fetch the right version. (v0.5.0-4430-f5c412b) I thought it was getting from the same version package. I'll see if I get 0.4.0 as well once I get the binary!

sgairo: I'm going to see why mine is showing 0.4.0

anchal-agrawal: @jooskim Yes, it should show the latest version.

jooskim: It seems to work nice for me. Not sure where the version info is set explicitly though. Or it might be set automatically from a git tag.. Anybody know about this?

anchal-agrawal: See #1815

lib/install/management/virtual_app.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/vmware/vic/lib/install/data"
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/trace"
+  "github.com/vmware/vic/pkg/version"
```
anchal-agrawal: `go fmt` :)

jooskim: yeah i was wondering why spacemacs was not allowing me to do indentation the way i wanted. another new stuff learned

anchal-agrawal: The [spacemacs go layer](https://github.com/syl20bnr/spacemacs/tree/master/layers/%2Blang/go) runs `go fmt` on every file save.

anchal-agrawal:
Thanks @jooskim! LGTM

lib/install/management/virtual_app.go
jooskim: yup, i just realized my spacemacs doesn't have go layer installed and that was why. now it does :-)

jooskim:
@andrewtchin Mind taking a look? Thanks!

andrewtchin:
lgtm

andrewtchin:
not sure how granular you want, but to differentiate between different builds you could use `version.GetBuild().ShortVersion()` that's the version the update process will look at vs `version.Version` is just the latest git tag so wouldn't necessarily change between builds

jooskim:
@andrewtchin i tended to display only the major minor patch digits so `version.Version` should be fine for me. thanks!

jooskim:
♻️

Issue #2261, opened 2016-09-07
Update UI docs to reflect that WinSCP has been removed


stuclem:
Per https://github.com/vmware/vic/issues/2204

stuclem:
WinSCP was removed in PR https://github.com/vmware/vic/pull/2275.

stuclem:
Removed topic on SFTP deployment, removed references to SFTP from the other topics.

PR #2275, opened 2016-09-07
[skip ci] Remove WinSCP from package

### LINKED ISSUE #2204: remove winscp from shipping bundle
The winscp binary that is pulled into the bundle uses OpenSSL 1.01 which has published CVEs.
While this is only used for a single point-in-time setup it's still not acceptable and nor is download of the binary without checking thumbprints.

There is a wider question about whether this should be in the main bundle at all. While we need to have a lowest common denominator solution as a fallback for deploying the plugins I am of the opinion that:
a. it doesn't need to be shipped in the main bundle
b. we should not be shipping winscp regardless

jooskim:
This PR fixes #2204 in that it removes the WinSCP binary and updates the affected scripts for vCenter Windows. This will require @stuclem to update the current documentation on UI installation via PR #2261.

### DIFF
--- a/ui/installer/vCenterForWindows/install.bat
+++ b/ui/installer/vCenterForWindows/install.bat
@@ -49,20 +49,7 @@ IF /I %vic_ui_host_url% NEQ NOURL (
 
     IF %vic_ui_host_url:~-1,1% NEQ / (
         SET vic_ui_host_url=%vic_ui_host_url%/
-    )    
-) ELSE (
-    IF %sftp_supported% EQU 1 (
-        ECHO Copying plugins...
-        "%utils_path%winscp.com" /command "open -hostkey=* sftp://%sftp_username%:%sftp_password%@%target_vcenter_ip%" "put -filemask=|*.zip ..\vsphere-client-serenity\* %target_vc_packages_path%" "exit"
-    ) ELSE (
-        ECHO SFTP not enabled. You have to manually copy the com.vmware.vicui.* folder in \ui\vsphere-client-serenity to %VMWARE_CFG_DIR%\vsphere-client\vc-packages\vsphere-client-serenity
-        ECHO Note: If you are running vCenter 5.5 Windows, copy the folder to %PROGRAMDATA%\VMware\vSphere Web Client\vc-packages\vsphere-client-serenity
     )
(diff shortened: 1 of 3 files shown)


ui/installer/vCenterForWindows/install.bat
```diff
@@ -92,4 +79,12 @@ FOR /F "tokens=*" %%A IN (..\vCenterForWindows\_scratch_flags.txt) DO (
 cd ..\vCenterForWindows
 DEL _scratch_flags.txt
 
-ECHO VIC UI was successfully installed. Make sure to log out of vSphere Web Client if you are logged in, and log back in.
+IF /I %vic_ui_host_url% EQU NOURL (
+    ECHO =============================
+    ECHO With the current version of VIC, you have to manually copy the com.vmware.vicui.* folder in \ui\vsphere-client-serenity to %VMWARE_CFG_DIR%\vsphere-client\vc-packages\vsphere-client-serenity
+    ECHO Note: If you are running vCenter Server 5.5, copy the folder to %PROGRAMDATA%\VMware\vSphere Web Client\vc-packages\vsphere-client-serenity instead
```
andrewtchin: can the script detect the version and display the correct message?

andrewtchin:
lgtm

ui/installer/vCenterForWindows/install.bat
jooskim: yeah. good suggestion. this is already part of the ui build process where it gets the version from `vic-machine`

jooskim: Never mind. I mistook the VIC version for VC version. Not sure how to get vCenter version. Will research if there's a way. It is still possible by running `vic-machine inspect` and extracting field `VMware OS version`, but that would require the user to put in the VCH's endpoint information which could be cumbersome

caglar10ur:
LGTM :sheep: :it:

jooskim:
♻️ 
Thanks @andrewtchin @caglar10ur

jooskim:
@stuclem The change is now merged, so you can update the documentation.

stuclem:
@jooskim will do. Thanks for the heads-up.

PR #2280, opened 2016-09-08
Fix erroneous file deletion output during successful vic-machine delete

### LINKED ISSUE #2120: vic-machine delete throws error, but delete completes successfully
When deleting the VCH via vic-machine via the following command:
` vic-machine-linux delete -t 'root:vagrant@192.168.175.131' -n vickie-esx" `

The following output occurs:
```
INFO[2016-08-24T18:04:57Z] ### Removing VCH ####
INFO[2016-08-24T18:04:57Z]
INFO[2016-08-24T18:04:57Z] VCH ID: VirtualMachine:10
INFO[2016-08-24T18:04:57Z] Removing VMs
ERRO[2016-08-24T18:04:57Z] Operation failed: Error caused by file /vmfs/volumes/567a44c7-67ea0133-584e-000c2956ea35/vickie-esx - (&{{} %!!(MISSING)s(*types.FileFault=&{{{<nil> []}} /vmfs/volumes/567a44c7-67ea0133-584e-000c2956ea35/vickie-esx}) Error caused by file /vmfs/volumes/567a44c7-67ea0133-584e-000c2956ea35/vickie-esx})
INFO[2016-08-24T18:04:57Z] Removing appliance VM network devices
INFO[2016-08-24T18:04:57Z] Bridge network was not created during VCH deployment, leaving it there
INFO[2016-08-24T18:04:57Z] Removing Resource Pool "vickie-esx"
INFO[2016-08-24T18:04:57Z] Completed successfully
```

The error is regarding the appliance vm, but appears to be a false error as everything is deleted and there are no items "orphaned" on the datastore.

Environment Details - Standalone ESXi:
```
Version:      6.0.0
Build:        3029758
OS type:      vmnix-x86
API type:     HostAgent
API version:  6.0
Product ID:   embeddedEsx
```

This is backed by a local datastore -- no vsan or network attached storage.  This occurs every time I delete from ESXi.

gigawhitlocks:
Fixes #2120 

Does not change other behavior, namely, still deletes the images parent directory if it's a non-default directory ( @emlin ) and in the case where the image directory is default, the images get wiped out when we delete the VCH directory.

We can change the behavior to leave the image directory around after deletion if it's non-default, but it's a little more involved and if we'd like to do that I'd like to squash this bug and get it out of master first and tackle that as a separate thing.

### DIFF
--- a/lib/install/management/delete_test.go
+++ b/lib/install/management/delete_test.go
@@ -88,10 +88,10 @@ func TestDelete(t *testing.T) {
 
 		createAppliance(ctx, validator.Session, conf, installSettings, false, t)
 		// FIXME: cannot check if it's VCH or not
-		//		testNewVCHFromCompute(input.ComputeResourcePath, input.DisplayName, validator, t)
-		testDeleteVCH(validator, conf, t)
+		// testNewVCHFromCompute(input.ComputeResourcePath, input.DisplayName, validator, t)
+		//	testDeleteVCH(validator, conf, t)
 		// FIXME: ServerFaultCode: HostDatastoreBrowser:hostdatastorebrowser-34 does not implement: SearchDatastoreSubFolders_Task
-		//		testDeleteDatastoreFiles(validator, t)
+		// testDeleteDatastoreFiles(validator, t)
 	}
 }
 
--- a/lib/install/management/store_files.go
+++ b/lib/install/management/store_files.go
@@ -39,7 +39,7 @@ const (
 )
 
 func (d *Dispatcher) deleteImages(conf *config.VirtualContainerHostConfigSpec) error {
-	var emptyImages bool
+	defer trace.End(trace.Begin(""))
 	var errs []string
 
 	for _, imageDir := range conf.ImageStores {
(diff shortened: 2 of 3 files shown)


emlin:
Is there any reason to delete image store parent directory?

gigawhitlocks:
If it's empty after images are removed we really should delete it. Otherwise it's clutter left behind by the installer and my opinion with uninstallers is that it's annoying when they don't fully clean up after themselves. It's also a directory that we (usually) create & we assume it's safe for writing arbitrary files in throughout runtime, so unless there's value in leaving images (in which case we should skip deleting images altogether), I don't know what the value is in leaving the directory there. 

And currently we delete everything in the directory, so if a user stored something besides images in that directory for some reason during the runtime of the VCH, their data would get wiped anyway. I don't see any reason to leave it around.

gigawhitlocks:
The unified diff makes it a little hard to see, but all I've done to `getSortedChildren` is break out the bulk of it into a method called `getChildren` so that I can call that in a case where I do not care if the results are sorted, and then `getSortedChildren` sorts the result of a call to the new function `getChildren`, which contains everything from the original `getSortedChildren` except the sort call.

This was a shortcut for testing to see if the directory was empty, but seems a good solution to me.

emlin:
lgtm

andrewtchin:
lgtm

gigawhitlocks:
Tests are failing due to a missing method in vcsim. Looking into it.

gigawhitlocks:
poke

Issue #701, opened 2016-05-17
Milestone: [composite] VCH appliance restart


hickeng:
This is a composite of the component parts of restarting a VCH:
- [x] Running containers need to be discovered and indexed
- [x] Name resolution needs to be populated with container names and aliases.
- [x] Existing networks need to be retrieved, and current IPAM state synthesized from network config and running containers.
- [x] Image stores need to be processed and indexed

This adds support for restarting the VCH appliance and retrieving system state from vsphere - given the appliance has no persistent state, being able to restart it is very close to being able to hand control off to a second appliance which is necessary for zero downtime upgrade.

The appliance should be responsive to requests immediately after boot where those requests don't rely on state that has yet to be restored.

mdubya66:
GA feature from SSOD

hickeng:
Closing this epic - the general behaviour is now implemented. Any deviation are now bugs.

PR #2283, opened 2016-09-08
[full ci] VCH Restart (partial)

### LINKED ISSUE #1830: Container cache to support restart

### LINKED ISSUE #1832: Relocate port layer container type and container cache

hmahmood:
Fixes #1830 

Partial fixes towards #701 

With this PR containers get indexed in the port layer's exec and network layers, but the docker personality still does not have an containers indexed. Coming soon in a follow up PR.

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -358,7 +358,7 @@ func (handler *ContainersHandlersImpl) ContainerWaitHandler(params containers.Co
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 
-	c := exec.ContainerInfo(uid.Parse(params.ID).String())
+	c := exec.Containers.Container(uid.Parse(params.ID).String())
 	if c == nil {
 		return containers.NewContainerWaitNotFound().WithPayload(&models.Error{
 			Message: fmt.Sprintf("container %s not found", params.ID),
@@ -376,7 +376,7 @@ func (handler *ContainersHandlersImpl) ContainerWaitHandler(params containers.Co
 		return containers.NewContainerWaitInternalServerError().WithPayload(&models.Error{Message: err.Error()})
 	}
 
-	c = exec.ContainerInfo(uid.Parse(params.ID).String())
+	c = exec.Containers.Container(uid.Parse(params.ID).String())
 	containerInfo := convertContainerToContainerInfo(c)
 	return containers.NewContainerWaitOK().WithPayload(containerInfo)
 }
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -248,7 +248,7 @@ func (handler *ContainersHandlersImpl) RemoveContainerHandler(params containers.
 func (handler *ContainersHandlersImpl) GetContainerInfoHandler(params containers.GetContainerInfoParams) middleware.Responder {
 	defer trace.End(trace.Begin("Containers.GetContainerInfoHandler"))
 
-	container := exec.ContainerInfo(params.ID)
+	container := exec.Containers.Container(params.ID)
 	if container == nil {
 		info := fmt.Sprintf("GetContainerInfoHandler ContainerCache miss for container(%s)", params.ID)
 		log.Error(info)
@@ -262,7 +262,13 @@ func (handler *ContainersHandlersImpl) GetContainerInfoHandler(params containers
 func (handler *ContainersHandlersImpl) GetContainerListHandler(params containers.GetContainerListParams) middleware.Responder {
 	defer trace.End(trace.Begin("Containers.GetContainerListHandler"))
 
-	containerVMs := exec.Containers(*params.All)
+	var state *exec.State
+	if params.All != nil && !*params.All {
+		state = new(exec.State)
+		*state = exec.StateRunning
+	}
+
+	containerVMs := exec.Containers.Containers(state)
 	containerList := make([]models.ContainerListInfo, 0, len(containerVMs))
 
 	for i := range containerVMs {
--- a/lib/portlayer/exec/config.go
+++ b/lib/portlayer/exec/config.go
@@ -19,11 +19,12 @@ import (
 
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/lib/config/executor"
 	"github.com/vmware/vic/lib/portlayer/event"
 )
 
-var config Configuration
+var Config Configuration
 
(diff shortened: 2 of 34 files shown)


lib/portlayer/exec/container.go
```diff
@@ -47,7 +47,7 @@ import (
 type State int
 
 const (
-	StateStarting = iota + 1
+	StateStarting State = iota
```
cgtexmex: iirc @caglar10ur specifically asked for the iota + 1 (in this or another change) --> I can't recall the reasoning...might double check

lib/portlayer/exec/config.go
```diff
@@ -24,7 +24,7 @@ import (
 	"github.com/vmware/vic/lib/portlayer/event"
 )
 
-var VCHConfig Configuration
+var Config Configuration
```
cgtexmex: iirc @caglar10ur had a change to add the "VCH" prefix -- I believe he was concerned about the proliferation of "config" vars in the codebase -- might double check

lib/portlayer/exec/exec.go
```diff
@@ -15,89 +15,117 @@
...
+		if err != nil {
+			err = fmt.Errorf("%s failed to start: %s", ec.Name(), err)
+			log.Error(err)
+			return
+		}
 
-	// subscribe the exec layer to the event stream for Vm events
-	VCHConfig.EventManager.Subscribe(events.NewEventType(vsphere.VmEvent{}).Topic(), "exec", eventCallback)
+		// create the event manager &  register the existing collector
+		Config.EventManager = event.NewEventManager(ec)
```
cgtexmex: The subscription to vsphere events has been removed, so OOB activity isn't being reflected in the container cache -- this needs to be added, so events are processed

hmahmood: Oops ... let me fix that

cgtexmex:
LGTM

hmahmood:
This is a seemingly voluminous PR. However, a lot of the changes are from the `VCHConfig -> Config` rename (which I think is important for consistency across port layers). Here are the files that actually form the crux of the restart changes:
```
lib/apiservers/portlayer/restapi/handlers/scopes_handlers.go
lib/portlayer/exec/container.go
lib/portlayer/exec/container_cache.go
lib/portlayer/exec/exec.go
lib/portlayer/network/config.go 
lib/portlayer/network/context.go
lib/portlayer/network/network.go
lib/portlayer/portlayer.go
```

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -244,7 +244,7 @@ func (handler *ContainersHandlersImpl) RemoveContainerHandler(params containers.
 func (handler *ContainersHandlersImpl) GetContainerInfoHandler(params containers.GetContainerInfoParams) middleware.Responder {
 	defer trace.End(trace.Begin("Containers.GetContainerInfoHandler"))
 
-	container := exec.ContainerInfo(params.ID)
+	container := exec.Containers.Container(params.ID)
```
fdawg4l: Kind of stutters, no?  I like the previous incarnation of this method name.

fdawg4l: Or maybe `exec.Containers.Info`

hmahmood: Yeah point taken. I initially changed it to `Get()` but didn't want to introduce even more changes to this PR. I can follow up with another PR that changes the names of the methods on the container cache.

lib/portlayer/exec/container_cache.go
```diff
@@ -32,45 +35,41 @@ type containerCache struct {
 	cache map[string]*Container
 }
 
-var containers *containerCache
+var Containers *containerCache
```
fdawg4l: This poses a startup problem, no?  Someone is going to call `Containers.Container` and PANIC the PL because `Containers` is `nil`.

lib/portlayer/exec/container_cache.go
```diff
@@ -32,45 +35,41 @@ type containerCache struct {
...
-
 }
 
 func (conCache *containerCache) Container(idOrRef string) *Container {
 	conCache.m.RLock()
 	defer conCache.m.RUnlock()
 	// find by id or moref
-	container := conCache.cache[idOrRef]
-	return container
+	return conCache.cache[idOrRef]
```
fdawg4l: I think this is another bug which this PR didn't implement, but still exists.

The map is a map of strings to pointers.  You're returning a reference to a value in the map which can be mutated outside of the cache (or this lock).  Is that intentional?

lib/portlayer/exec/container_cache.go
```diff
@@ -99,11 +102,25 @@ func (conCache *containerCache) Remove(idOrRef string) {
...
+func (conCache *containerCache) sync(ctx context.Context, sess *session.Session) error {
+	cons, err := infraContainers(ctx, sess)
+	if err != nil {
+		return err
+	}
+
+	conCache.m.Lock()
+	defer conCache.m.Unlock()
+
+	conCache.cache = make(map[string]*Container)
```
fdawg4l: `sync` is called from Init, as is NewContainerCache().  Can this be moved to NewContainerCache() or a comment added as to why this needs to be here?

lib/portlayer/exec/container_cache.go
hmahmood: Yes, but we do initialize this in exec.Init(). I will move this to package init() when we have a new package for the container cache.

lib/portlayer/exec/container_cache.go
hmahmood: Yes this is intentional. The Container object is thread safe.

lib/portlayer/exec/container_cache.go
```diff
@@ -32,45 +35,41 @@ type containerCache struct {
...
-			} else if container.State == StateRunning {
-				// only include running...
-				containers = append(containers, container)
-
-			}
+		if !isContainerID(id) {
+			continue
+		}
+
+		if state == nil || *state == con.State {
```
hmahmood: We are looking for containers with a particular state, e.g. StateRunning.

lib/portlayer/exec/container_cache.go
hmahmood: This is to clear out the map; is there another way to do this in go?

lib/portlayer/exec/container_cache.go
emlin: Hasan, is there any reason to return container with state==nil here?

emlin:
lgtm

lib/portlayer/exec/container_cache.go
hmahmood: state == nil means any state, so we return all the containers in the cache.

hmahmood:
♻️

PR #2284, opened 2016-09-08
Enable debug mode to the VCH to allow more logging during our pre-GA …

mhagen-vmware:
Just enabling debug mode on our VCH for now, should be beneficial as we head in this period of more difficult to reproduce bugs.


jzt:
LGTM

andrewtchin:
lgtm

mhagen-vmware:
.

PR #2288, opened 2016-09-09
vcsim: Implement SearchDatastoreSubFolders_Task

### LINKED ISSUE #2287: Need SearchDatastoreSubFolders_Task in VC simulator
We have two unit tests commented out due to the lack of this method in the VC simulator, in `/lib/install/management/delete_test.go` at the bottom of `TestDelete` you will find this:
```
		//	testDeleteVCH(validator, conf, t)
		// FIXME: ServerFaultCode: HostDatastoreBrowser:hostdatastorebrowser-34 does not implement: SearchDatastoreSubFolders_Task
		// testDeleteDatastoreFiles(validator, t)
```

`SearchDatastoreSubFolders_Task` needs to be implemented in vcsim so that `testDeleteVCH` and `testDeleteDatastoreFolders` can be re-enabled. 

This issue can be closed once those two tests are re-enabled.

dougm:
Fixes #2287

### DIFF
--- a/lib/install/management/delete_test.go
+++ b/lib/install/management/delete_test.go
@@ -90,8 +90,8 @@ func TestDelete(t *testing.T) {
 		// FIXME: cannot check if it's VCH or not
 		// testNewVCHFromCompute(input.ComputeResourcePath, input.DisplayName, validator, t)
 		//	testDeleteVCH(validator, conf, t)
-		// FIXME: ServerFaultCode: HostDatastoreBrowser:hostdatastorebrowser-34 does not implement: SearchDatastoreSubFolders_Task
-		// testDeleteDatastoreFiles(validator, t)
+
+		testDeleteDatastoreFiles(validator, t)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


emlin:
lgtm

lib/install/management/delete_test.go
```diff
@@ -90,8 +90,8 @@ func TestDelete(t *testing.T) {
 		// FIXME: cannot check if it's VCH or not
 		// testNewVCHFromCompute(input.ComputeResourcePath, input.DisplayName, validator, t)
 		//	testDeleteVCH(validator, conf, t)
```
gigawhitlocks: uncomment line 92 too, it should work now :)

gigawhitlocks:
One comment, otherwise LGTM :) :+1:

lib/install/management/delete_test.go
dougm: Looks like we need to implement SearchIndex in vcsim for that to work, so I'll leave it for now.

PR #2292, opened 2016-09-09
Sync UI plugin version with the latest tag version

jooskim:
This PR makes changes to the `.drone.yml` file so that the UI plugin version is always synchronized with the `vmware/vic` repo's latest tag version. Addresses the oral requirement from @mdubya66 about versioning.


hickeng:
This makes me a little twitchy... the UI plugin operates on a different cycle as far as versioning is concerned because one UI plugin has to deal with hetrogeneous VIC/E versions. If we're forcing a UI version update each build then what's the update strategy for the plugin?

Do we allow multiple plugin versions to exist and a plugin version only renders if it can validate all the data it expects and a more recent version doesn't "claim" the VCH?

jooskim:
@hickeng `If we're forcing a UI version update each build then what's the update strategy for the plugin?` I am pulling only major, minor and patch digits not build number, which are coming from the latest tag we have. So unless we make another tag the UI version won't change. Also there's a script called `upgrade.sh` that will basically uninstall the existing extension and install the new one, so the upgrade process is quite straightforward.

In addition, we are not going to make any more change to the Flex code because we're moving towards H5 client, and the plugin codebase has been pretty much the same for the past one-two months. So it's more for the sake of consistency such that when we finally release VIC v1.0.0 then the plugin version will also be v1.0.0, by when the code really will not have changed from what it is now.

jooskim:
@hickeng It'd be great if you could share your further concerns or questions here, as versioning will be quite important when we start building an HTML5 plugin. For example, we'll have to discuss whether we want to distribute the HTML5 plugin and Flex plugin in one bundle (pro: easier version control and distribution, con: increasing version of the Flex plugin will not make sense), or separate bundles (pro: accurate version control, con: a lot more effort to maintain).

andrewtchin:
lgtm pending george's comments

jooskim:
@hickeng Do you still think we shouldn't version the UI according to that of VIC Engine? If you think there are more risks associated with having it follow the latest tag version I could take this PR down and do versioning manually.

hickeng:
@jooskim LGTM given we're not investing much more into the flex plugin.

For H5, we had a brief conversation months ago with the webclient team about whether it's viable to have a more generic plugin that's commonly available for rendering application specific data in a structured fashion. Basically a richer version of the Notes or Annotations that already exist.
If that's viable then we should consider having the dependency the other way around and it's actually the plugin that exposes a data-model that VIC conforms to.

jooskim:
@hickeng Thanks for your input. We will soon confirm the specific plan for UI / UX on H5 Client with @karthik-narayan. As the first step we are looking to port the existing Flex portlets into the H5 client.

jooskim:
♻️

PR #2300, opened 2016-09-12
Follow up to vMotion test plan[skip ci]

### LINKED ISSUE #2270: Follow on work for vmotion test plan
@mhagen-vmware couple of items of note
1. It's not clear to me whether we're installing 3 concurrent VCHs, or cleaning up prior to subsequent install.
2. The initial test (powered off vch) isn't testing vmotion, but is testing some of VCH restart.
3. We should be specific about which commands are crucial in expected results. Saying all docker commands complete without error is already wider than viable unless we link a doc of supports commands and sub-options.
We should confirm attach and logs as a priority after the vmotion.
4. We should confirm that a container that has previously been attached to can be vmotioned after the fact.
5. We should confirm our current behaviour that a container with an active attach connection cannot be vmotioned.

mhagen-vmware:
Fixes #2270 

@hickeng I have addressed all comments I think except for:
3. We should be specific about which commands are crucial in expected results. Saying all docker commands complete without error is already wider than viable unless we link a doc of supports commands and sub-options.

The expected results statement that references "all docker commands" is not an execution step, it is only the expected results of elsewhere defined execution steps, which is only an indication that whatever docker commands you may have run should not fail.


matthewavery:
lgtm

andrewtchin:
lgtm

mhagen-vmware:
.

PR #2301, opened 2016-09-12
Upgrade configuration

### LINKED ISSUE #2172: vic-machine validator: validate existing VCH configuration
Validate VirtualContainerHostConfigSpec data, to make sure all required attributes is there, and all values are in expected format.
This function will be used in upgrade, to verify if existing VCH configuration is compatible with latest VirtualContainerHostConfigSpec version.
If not, return correct error message.
This function should not print out message only, cause eventually, we'll need to migrate data from old version's VirtualContainerHostConfigSpec based on the error. Print out is not a good way to consume the error.
In the meanwhile, this issue does not include data migration.

emlin:
Fixes #2172 

Check required options are there, and change configuration version.

In the meanwhile, update upgrade process to align with design doc changes.

### DIFF
--- a/cmd/vic-machine/upgrade/upgrade.go
+++ b/cmd/vic-machine/upgrade/upgrade.go
@@ -56,7 +56,7 @@ func (u *Upgrade) Flags() []cli.Flag {
 		},
 		cli.BoolFlag{
 			Name:        "force, f",
-			Usage:       "Force the upgrade -- ignore version issue",
+			Usage:       "Force the upgrade (ignores version checks)",
 			Destination: &u.Force,
 		},
 	}
--- a/cmd/vic-machine/upgrade/upgrade.go
+++ b/cmd/vic-machine/upgrade/upgrade.go
@@ -54,6 +54,11 @@ func (u *Upgrade) Flags() []cli.Flag {
 			Usage:       "Time to wait for upgrade",
 			Destination: &u.Timeout,
 		},
+		cli.BoolFlag{
+			Name:        "force, f",
+			Usage:       "Force the upgrade -- ignore version issue",
+			Destination: &u.Force,
+		},
 	}
 	flags = append(
 		append(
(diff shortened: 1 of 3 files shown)


emlin:
At this time, upgrade could finish successfully, cause all services are running and docker info works well, but the vch restart feature is not ready yet, so even with these conditions checked, the VCH upgrade does not actually pick up container information created before upgrade.
So the upgrade command is till hidden, to avoid confusion to user.

lib/install/management/upgrade.go
```diff
@@ -58,14 +59,28 @@ func (d *Dispatcher) Upgrade(vch *vm.VirtualMachine, conf *config.VirtualContain
...
+		d.VICAdminProto = "http"
+		d.DockerPort = fmt.Sprintf("%d", opts.DefaultHTTPPort)
+	}
+
+	if err = d.uploadImages(settings.ImageFiles); err != nil {
+		return errors.Errorf("Uploading images failed with %s. Exiting...", err)
+	}
+	conf.BootstrapImagePath = fmt.Sprintf("[%s] %s/%s", conf.ImageStores[0].Host, d.vmPathName, settings.BootstrapISO)
 
 	snapshotName := fmt.Sprintf("upgrade for %s", conf.Version.BuildNumber)
```
andrewtchin: can this use the const?

lib/install/validate/update.go
```diff
@@ -0,0 +1,95 @@
...
+	if len(conf.ImageStores) == 0 {
+		v.NoteIssue(errors.New("Image store is not set"))
+	}
+}
+
+func (v *Validator) assertTarget(conf *config.VirtualContainerHostConfigSpec) {
+	defer trace.End(trace.Begin(""))
+	if conf.Target.User != nil {
+		if _, set := conf.Target.User.Password(); set {
+			v.NoteIssue(errors.New("Password should not set in target URL"))
```
andrewtchin: `Password should not be set in target URL`

cmd/vic-machine/upgrade/upgrade.go
```diff
@@ -54,6 +54,11 @@ func (u *Upgrade) Flags() []cli.Flag {
 			Usage:       "Time to wait for upgrade",
 			Destination: &u.Timeout,
 		},
+		cli.BoolFlag{
+			Name:        "force, f",
+			Usage:       "Force the upgrade -- ignore version issue",
```
andrewtchin: maybe change this to `Force the upgrade (ignores version checks)`

andrewtchin:
lgtm

lib/install/validate/update.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (v *Validator) assertVersion(conf *config.VirtualContainerHostConfigSpec) {
+	defer trace.End(trace.Begin(""))
+	if conf.Version == nil {
+		v.NoteIssue(errors.Errorf("Unknown version of VCH %q", conf.Name))
+		return
+	}
+	installerBuild := version.GetBuild()
+	if installerBuild.Equal(conf.Version) {
+		v.NoteIssue(errors.Errorf("%q has same version as installer. No upgrade is avaialble.", conf.Name))
```
anchal-agrawal: s/avaialble/available

lib/install/management/upgrade.go
```diff
@@ -22,6 +22,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 
+	"github.com/docker/docker/opts"
 	"github.com/vmware/govmomi/object"
 	//	"github.com/vmware/govmomi/task"
```
anchal-agrawal: Let's delete this commented line. 😃

anchal-agrawal:
Just a couple nitpicks. LGTM

emlin:
♻️

PR #2304, opened 2016-09-12
ContainerWait

### LINKED ISSUE #384: Docker API: Container Wait
Implement the docker wait operation in the engine-api server and integrate with interaction port layer.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#wait-a-container
https://docs.docker.com/engine/reference/commandline/wait/

### LINKED ISSUE #2310: Out of band guest shutdown results in wrong container state
When issuing an out of band shutdown to a container VM the expected state via `docker ps` is "Stopped".  Currently the container will be set to "Stopping" and won't reach "Stopped".

To recreate simply create a container, start the container and then perform a guest shutdown via govc or a ui client.

cgtexmex:
The primary change is to implement ContainerWait, but a few
other changes were also implemented:

* inspect now has the appropriate exit code
* ps shows the exit code for stopped containers
* fixed issue with state of containers shutdown out of band

fixes #384, fixes #2310

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -793,7 +793,84 @@ func (c *Container) ContainerUpdate(name string, hostConfig *containertypes.Host
 // timeout, an error is returned. If you want to wait forever, supply
 // a negative duration for the timeout.
 func (c *Container) ContainerWait(name string, timeout time.Duration) (int, error) {
-	return 0, fmt.Errorf("%s does not implement container.ContainerWait", ProductName())
+	defer trace.End(trace.Begin(fmt.Sprintf("name(%s):timeout(%s)", name, timeout)))
+
+	// Look up the container name in the metadata cache to get long ID
+	vc := cache.ContainerCache().GetContainer(name)
+	if vc == nil {
+		return -1, NotFoundError(name)
+	}
+	id := vc.ContainerID
+
+	//retrieve client to portlayer
+	client := c.containerProxy.Client()
+	results, err := client.Containers.ContainerWait(containers.NewContainerWaitParamsWithContext(ctx).WithTimeout(int64(timeout.Seconds())).WithID(id))
+	if err != nil {
+		switch err := err.(type) {
+		case *containers.ContainerWaitNotFound:
+			// since the container wasn't found on the PL lets remove from the local
+			// cache
+			cache.ContainerCache().DeleteContainer(id)
+			return -1, NotFoundError(name)
+		case *containers.ContainerWaitInternalServerError:
+			return -1, InternalServerError(err.Payload.Message)
+		default:
+			return -1, InternalServerError(err.Error())
+		}
+	}
+
+	containerInfo := results.Payload
+
+	// call to the dockerStatus function to retrieve the docker friendly exitCode
+	// TODO: once started / finished time are available replace time.Now()
+	exitCode, _ := dockerStatus(int(*containerInfo.ProcessConfig.ExitCode),
+		*containerInfo.ProcessConfig.Status,
+		*containerInfo.ContainerConfig.State,
+		time.Now(), time.Now())
+
+	return exitCode, nil
+}
+
+// dockerStatus will evaluate the container state, exit code and
+// process status to return a docker friendly status
+//
+// exitCode is the container process exit code
+// status is the container process status -- stored in the vmx file as "started"
+// started / finished are timestamps, but are not yet implemented
+func dockerStatus(exitCode int, status string, state string, started time.Time, finished time.Time) (int, string) {
+	// TODO: once started / finished are available then github.com/docker/go-units must be imported for use in
+	// the status string
+	switch state {
+	case "Created":
+		status = state
+	case "Running":
+		//TODO: once we have started use commented out status
+		// status = fmt.Sprintf("Up %s", units.HumanDuration(time.Now().UTC().Sub(started)))
+		status = state
+	case "Stopped":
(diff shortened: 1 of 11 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -801,7 +801,71 @@ func (c *Container) ContainerUpdate(name string, hostConfig *containertypes.Host
...
+	switch state {
+	case "Created":
+		status = state
+	case "Running":
+		//TODO: once we have started use commented out status
+		// status = fmt.Sprintf("Up %s", units.HumanDuration(time.Now().UTC().Sub(started)))
+		status = state
+	case "Stopped":
+
+		if strings.Contains(status, "permission denied") {
```
cgtexmex: status in this context is the started string from the vmx file -- so in the case of a permission / path /file issue it will have error text.  Unfortunately, in those cases the exit code in the vmx is still 0...possibly that can be corrected in the tether /tools...but I understood that there were some challenges with that.

Agree the use of status as a func variable is confusing -- I'll change that and add some additional explanation.

lib/apiservers/engine/backends/container.go
```diff
@@ -793,7 +793,81 @@ func (c *Container) ContainerUpdate(name string, hostConfig *containertypes.Host
...
+		return -1, NotFoundError(name)
+	}
+	id := vc.ContainerID
+
+	//retrieve client to portlayer
+	client := c.containerProxy.Client()
+	results, err := client.Containers.ContainerWait(containers.NewContainerWaitParamsWithContext(ctx).WithTimeout(int64(timeout.Seconds())).WithID(id))
+	if err != nil {
+		switch err := err.(type) {
+		case *containers.ContainerWaitNotFound:
```
jzt: Do we need to sync the cache in this case? Or is that taken care of by an event?

cgtexmex: I'm assuming you're referring to the persona container cache -- if so, event activity is not streamed to the persona yet, so that cache wouldn't be mutated based on events.

So -- in the case that the container isn't found on the PL side then yes...we should eject it from the persona side.  I can add that when I rebase.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -315,6 +323,45 @@ func (handler *ContainersHandlersImpl) GetContainerLogsHandler(params containers
 	return NewContainerOutputHandler("logs").WithPayload(detachableOut, params.ID)
 }
 
+func (handler *ContainersHandlersImpl) ContainerWaitHandler(params containers.ContainerWaitParams) middleware.Responder {
+	defer trace.End(trace.Begin(fmt.Sprintf("%s:%d", params.ID, params.Timeout)))
+
+	// default context timeout in seconds
+	default_timeout := int64(180)
```
hmahmood: Violates go style I believe: https://golang.org/doc/effective_go.html#mixed-caps

hmahmood: You should make this is a const.

cgtexmex: apologies that's the ruby in me...

dougm: How did this pass CI?  That should be an error from golint.

lib/portlayer/exec/exec.go
```diff
@@ -108,9 +109,24 @@ func eventCallback(ie events.Event) {
...
 				StateStopped,
 				StateSuspended:
+
 				log.Debugf("Container(%s) state set to %s via event activity", container.ExecConfig.ID, newState.String())
 				container.State = newState
-				publishContainerEvent(container.ExecConfig.ID, ie.Created(), ie.String())
+				// container state has changed so we need to update the container attributes
+				// we'll do this in a go routine to avoid blocking
+				go func() {
+
```
dougm: What up with the empty newlines at the beginning and end of this and some of your other, but not all, functions?

lib/portlayer/exec/exec_test.go
```diff
@@ -21,7 +21,7 @@ import (
 	"github.com/stretchr/testify/assert"
 
 	"github.com/vmware/vic/lib/portlayer/event"
-	"github.com/vmware/vic/lib/portlayer/event/collector/vsphere"
+	// "github.com/vmware/vic/lib/portlayer/event/collector/vsphere"
```
dougm: remove

lib/portlayer/exec/exec.go
```diff
@@ -171,3 +184,35 @@ func publishContainerEvent(id string, created time.Time, eventType string) {
...
+	VCHConfig.EventManager.Subscribe(topic, sub, watch)
+	defer VCHConfig.EventManager.Unsubscribe(topic, sub)
+
+	// wait for the event to be pushed on the channel or
+	// the context to be complete
+	select {
+	case <-listen:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
```
dougm: Let's wrap that error here or in the caller, otherwise we have another possible CDE error without any context.

dougm:
lgtm

lib/portlayer/exec/exec.go
cgtexmex: I'll work on my consistency...I removed them from here

cgtexmex:
addressed concerns & rebased

cgtexmex:
rebase

jzt:
LGTM ♻️

PR #2308, opened 2016-09-13
Add tail and follow support to logs [full ci]

### LINKED ISSUE #1682: Datastore file streaming
In support of #366 - we need to figure out a way to stream log files from the datastore.  We are currently able to download a complete file, but we'll want to "tail" the file to support docker logs streaming.

dougm:
Fixes #1682

### DIFF
--- a/lib/portlayer/exec/container.go
+++ b/lib/portlayer/exec/container.go
@@ -70,6 +70,8 @@ type Container struct {
 	VMUnsharedDisk int64
 
 	vm *vm.VirtualMachine
+
+	logFollowers []io.Closer
 }
 
 func NewContainer(id uid.UID) *Handle {
@@ -315,6 +317,9 @@ func (c *Container) stop(ctx context.Context, waitTime *int32) error {
 	if c.vm == nil {
 		return fmt.Errorf("vm not set")
 	}
+
+	defer c.onStop()
+
 	// get existing state and set to stopping
 	// if there's a failure we'll revert to existing
 	existingState := c.State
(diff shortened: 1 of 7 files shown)


lib/portlayer/exec/container.go
```diff
@@ -377,15 +382,23 @@ func (c *Container) Signal(ctx context.Context, num int64) error {
 	return c.startGuestProgram(ctx, "kill", fmt.Sprintf("%d", num))
 }
 
-func (c *Container) LogReader(ctx context.Context) (io.ReadCloser, error) {
+func (c *Container) stopped() {
```
sflxn: Looks like this only deals with the log reader.  Why not call it stopLogReaders() or something?  Not a big deal.

sflxn:
Just a naming comment.  Other then that, LGTM.

lib/portlayer/exec/container.go
cgtexmex: agree that a different function name would be preferable

dougm: Just wanted a generic method that'd be called when an container is stopped, to
avoid sandwiching duplicate code into the same two places in the future.  I can rename it.

cgtexmex: @dougm I like the idea of a generic stop method that could be called for any clean up routines....I'd name it onStop and have the state evaluation in the function

dougm: ok onStop it is.  Why would 'onStop' evaluate the state?  eventCallback could probably be refactored a bit, but that's where we are already evaluating state changes and dispatching based on the state change.

cgtexmex: eventCallback will only execute onStop if the event is stopping the container -- so out of band stops / power offs, etc...or a short lived container that exits and powers down.  Executing a client side stop will not.  

So thinking potentially others might call onStop and didn't want to recreate the if statement in those places...no biggie...so whatever you feel best about

dougm: Right, I added to call there to handle out-of-band power off and container shutting down on its own.  The Container.stop() method has an explicit call to Container.onStop to handle client stop call, in which case there's no need to check the state.

cgtexmex:
LGTM

caglar10ur:
LGTM :sheep: :it:

dougm:
♻️

PR #2309, opened 2016-09-13
Fix "Container stop does not work"

### LINKED ISSUE #2271: Stopped Containers appear to not actually be stopped
When attemping to reproduce another issue. I made a busybox /bin/top container. started it. then stopped it(supposedly) and tried to remove it. The removal fails with a "power of the container first" type error, this is after a reportedly successful stop call. 

```
matthewavery@ubuntu ~/go/src/github.com/vmware/vic
 % docker -H 10.160.60.105:2376 --tls stop dced210b3c98596bb03447f7cd1cd50661528004ef2d7c6f359f8e69e8faa53a    
dced210b3c98596bb03447f7cd1cd50661528004ef2d7c6f359f8e69e8faa53a
matthewavery@ubuntu ~/go/src/github.com/vmware/vic
 % docker -H 10.160.60.105:2376 --tls rm dced210b3c98596bb03447f7cd1cd50661528004ef2d7c6f359f8e69e8faa53a
Error response from daemon: You cannot remove a running container. Stop the container before attempting removal or use -f
matthewavery@ubuntu ~/go/src/github.com/vmware/vic
 % docker -H 10.160.60.105:2376 --tls stop dced210b3c98596bb03447f7cd1cd50661528004ef2d7c6f359f8e69e8faa53a
dced210b3c98596bb03447f7cd1cd50661528004ef2d7c6f359f8e69e8faa53a
matthewavery@ubuntu ~/go/src/github.com/vmware/vic
 % docker -H 10.160.60.105:2376 --tls rm dced210b3c98596bb03447f7cd1cd50661528004ef2d7c6f359f8e69e8faa53a
Error response from daemon: You cannot remove a running container. Stop the container before attempting removal or use -f
```

gigawhitlocks:
Fixes #2271 

This bug was caused by an unreachable block of code in the Docker persona causing the container cache in the persona to remain valid after a lookup miss. The second commit contains the fix, which is clearer if it's viewed in isolation from the first commit. Note that if execution reaches line 723 (in the original version) that `err` will always be `nil` at that time. Assigning the variable on line 720 (in the new version) allows cache invalidation to occur properly in the case that a miss occurs.

The first commit is a small refactor to remove an unneeded and confusing parameter rename (the parameter was renamed from `name` to `paramName` so that the variable `name` could be reused to contain a container id. I've removed `paramName` so that `name` remains the function parameter throughout the method, and assigned the container id to a variable called `id` instead.)

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -716,8 +716,8 @@ func (c *Container) containerStop(name string, seconds int, unbound bool) error 
 
 	// we have a container on the PL side lets check the state before proceeding
 	// ignore the error  since others will be checking below..this is an attempt to short circuit the op
-	// TODO: can be replaced with simple cache check once power events are propigated to persona
-	infoResponse, _ := client.Containers.GetContainerInfo(containers.NewGetContainerInfoParamsWithContext(ctx).WithID(name))
+	// TODO: can be replaced with simple cache check once power events are propagated to persona
+	infoResponse, err := client.Containers.GetContainerInfo(containers.NewGetContainerInfoParamsWithContext(ctx).WithID(id))
 	if err != nil {
 		cache.ContainerCache().DeleteContainer(id)
 		return NotFoundError(name)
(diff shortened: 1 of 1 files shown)


jzt:
LGTM

cgtexmex:
LGTM

PR #2315, opened 2016-09-13
Remove start context, change commit timeout to 3 minutes [full ci]

jzt:
I've removed the timeouts used in container commit and start. These should hopefully clear up any CDE's during CI builds. Users are now responsible for cancelling long-running requests themselves.

See #1822 for discussion around this.

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -108,8 +108,7 @@ type Container struct {
 }
 
 const (
-	commitTimeout  = 1 * time.Minute
-	DefaultEnvPath = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+	defaultEnvPath = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 )
 
 func NotFoundError(msg string) error {
@@ -555,7 +554,7 @@ func (c *Container) containerStart(name string, hostConfig *containertypes.HostC
 	}
 
 	// commit the handle; this will reconfigure and start the vm
-	_, err = client.Containers.Commit(containers.NewCommitParamsWithTimeout(commitTimeout).WithHandle(handle))
+	_, err = client.Containers.Commit(containers.NewCommitParamsWithContext(ctx).WithHandle(handle))
 	if err != nil {
 		switch err := err.(type) {
 		case *containers.CommitNotFound:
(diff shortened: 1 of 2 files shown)


sflxn:
As far as changing the timeouts, it LGTM, but these timeouts are really hard to manage.  Longer term, we should do something different.  For now, this is as good a try as any.

cgtexmex:
LGTM

lib/apiservers/engine/backends/container.go
```diff
@@ -109,14 +109,14 @@ type Container struct {
 }
 
 const (
-	commitTimeout          = 1 * time.Minute
+	commitTimeout          = 3 * time.Minute
```
hmahmood: I am in favor of getting rid of this timeout altogether. This is a client side timeout, and given that the caller of the docker command can hit Ctrl-C to back out of this, having an arbitrary timeout seems superfluous. The port layer server is going to (eventually) queue up an operation on VC, and that will result in a response in a finite amount of time. Any concerns about outstanding operations taking up resources indefinitely on the server side on the VCH should be mitigated by this.

caglar10ur:
LGTM :sheep: :it:

hmahmood:
lgtm

lib/apiservers/engine/backends/container.go
```diff
@@ -108,8 +108,7 @@ type Container struct {
 }
 
 const (
-	commitTimeout  = 1 * time.Minute
-	DefaultEnvPath = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
+	defaultEnvPath = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
```
caglar10ur: nice change (just testing review thingy)

lib/apiservers/engine/backends/container.go
```diff
@@ -555,7 +554,7 @@ func (c *Container) containerStart(name string, hostConfig *containertypes.HostC
 	}
 
 	// commit the handle; this will reconfigure and start the vm
-	_, err = client.Containers.Commit(containers.NewCommitParamsWithTimeout(commitTimeout).WithHandle(handle))
+	_, err = client.Containers.Commit(containers.NewCommitParams().WithHandle(handle))
```
caglar10ur: WithContext? Otherwise it will default to 30 sec.

lib/apiservers/engine/backends/container.go
jzt: I'm honored to be the guinea pig in this experiment. :)

lib/apiservers/engine/backends/container.go
jzt: Fixed

jzt:
♻️

PR #2319, opened 2016-09-14
Rename vicbackends package to backends

fdawg4l:
Go convention is to name the directory the same as the pacakge.  Also
vicbackends stutters.


cgtexmex:
LGTM

anchal-agrawal:
LGTM

PR #2320, opened 2016-09-14
Call VolumeCreate for anonymous volumes

### LINKED ISSUE #2313: docker volume ls returns EOF in some cases
1. docker volume create --name victest
2. docker volume ls

- success

Then do

3\. docker create -v /test busybox
4\. docker volume ls

- An error occurred trying to connect: Get http://192.168.218.128:2375/v1.23/volumes: EOF

Actually, after operation 3, docker volume operations no longer seem to work.

The only way to recover is to uninstall vic, using vic-machine, with the --force flag and reinstall vic.

fdawg4l:
The AddVolumesToContainer() call was calling processAnonymousVolumes
which implemented its own call to VolumeCreate on the PL.  Call the
VolumeCreate used by the volume create method instead of this out of
band version.

Moved container proxy into backends.  I didn't see the need to have this
in its own package only to import the package above it.  This avoids the
cyclic dependency.

Fixes #2313

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -51,7 +51,6 @@ import (
 
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	viccontainer "github.com/vmware/vic/lib/apiservers/engine/backends/container"
-	"github.com/vmware/vic/lib/apiservers/engine/backends/portlayer"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/portmap"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/containers"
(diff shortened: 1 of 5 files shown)


matthewavery:
lgtm, there are so many things named `Volume` haha took me a moment to track down where everything was that was getting called still.

anchal-agrawal:
LGTM

fdawg4l:
@matthewavery There really shouldnt be that many things named `Volume`.  There's the `Volume` struct on the PL side.  That should be it.  There are `VolumeCreate` calls on the docker daemon side (which I think is named incorrectly), and a VolumeCreate on the PL REST server side which calls VolumeCreate on the storage layer, which calls VolumeCreate on vsphere.  

Yes, they are redundant, but that's the nature of interfaces.  It's important, therefore, to name the struct the methods belong to appropriately.  And we need to do a better job at that.   Ping with a list of weirdly named structs and we can rename them as we go along.

PR #2322, opened 2016-09-14
Add host type and differentiate between behavior based on it

### LINKED ISSUE #1963: Docker RM test govc call needs modifying for VC
I'm raising an issue because I don't know the cleanest way to fix this, so delegating.

Simple problem is that the VCH vApp on VC means that govc ls /ha-datacenter/vm doesn't show the test containerVM. You have to prefix the vApp name.

This means that for the govc command in "Remove a container deleted out of band", you need the following: ${rc}  ${output}=  Run And Return Rc And Output  govc vm.destroy ${vch-name}"/test*"

This works for VC, but breaks on ESX. We need to figure out a way to try one and if it doesn't succeed, try the other.

mhagen-vmware:
Fixes #1963

### DIFF
--- a/tests/resources/Util.robot
+++ b/tests/resources/Util.robot
@@ -35,6 +35,11 @@ Set Test Environment Variables
     Set Environment Variable  GOVC_RESOURCE_POOL  %{TEST_RESOURCE}
     Set Environment Variable  GOVC_DATASTORE  %{TEST_DATASTORE}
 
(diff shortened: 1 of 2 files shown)


matthewavery:
lgtm

andrewtchin:
lgtm

mhagen-vmware:
.

PR #1441, opened 2016-07-12
Correct handling of disks on container delete

hickeng:
This corrects the handling of disks (both vmfs and vsan) when deleting container VMs.
This is an inefficient implementation in terms of number of remote API operations but uses only public calls.

Fixes #1383

Issue #1313, opened 2016-07-01
Removal of a stopped container results in the datastore not being cleaned


mhagen-vmware:
docker ${params} pull busybox
docker ${params} create busybox ls
docker ${params} start ${container}
docker ${params} rm ${container}

Results in the vm being removed but the datastore doesn't get cleaned up.

cgtexmex:
This appears to be caused by the fact that we create a container VM specific log and debug file upon container start -- I believe these are created and managed by the tether.  esx is deleting from disk all the files it's aware of and since it's unaware of those it leaves them behind.

corrieb:
NOTE: CLOSING THIS ISSUE MAY CAUSE AN INTEGRATION TEST FAILURE. Please re-enable the test in 1-11-Docker-RM.robot as part of a PR to fix this issue. Please also ensure that the test passes cleanly. If the test needs improving or updating, please address it. Thx

hickeng:
@corrieb could you try re-enabling this in your local integration build? I'm in the middle of redeploying dev box but this should now be fixed by #1441

corrieb:
@hickeng will do

matthewavery:
@corrieb I am closing this after confirming that it is resolved by #1441.

andrewtchin:
the build is broken! :trollface:

matthewavery:
The test has now be re-enabled in master.

stuclem:
This is fixed. Removing kind/note.

PR #2323, opened 2016-09-14
Enables tests which were disabled due to tickets 1313 and 1312

### LINKED ISSUE #1312: force remove of a running container fails
Trying to force remove a running container:
docker ${params} rm -f ${container}

Actual:
Error response from daemon: server error from portlayer

Expected:
Container should be removed

matthewavery:
Fixes #1312 #1313 

Both #1312 and #1313 were solved when other p0s were solved. 

#1312 : this was solved by @gigawhitlocks fix for the `container stop` issue where stop would not work and was not updating state. The fix was #2309 since rebasing past this PR. 

#1313 : this was solved by @hickeng with pr #1441

After re enabling the test, it was confirmed that the tests now pass on both VC and ESX.


andrewtchin:
lgtm

mdubya66:
lgtm

mhagen-vmware:
lgtm

matthewavery:
♻️

PR #2325, opened 2016-09-14
Add vice prefix for vic engine configuration

emlin:
As @hickeng  mentioned in #1955, we'd like to have a prefix like "vice" in the guestinfo key, to differentiate with other modules in vic bundle.

So this PR change the default guestinfo prefix from "guestinfo." to "guestinfo.vice.". With this change, vic-machine will not recognize old VCH created by early version, cause the vch identification key in guestinfo is changed.
And this is also the reason, for we need to have this merged before GA.


jzt:
LGTM 👍

andrewtchin:
lgtm

emlin:
♻️

emlin:
@jooskim This PR changed the vch appliance configuration key, which might be used in UI to query VCH appliance information, e.g. the ip key is changed to "guestinfo.vice..init.networks|client.ip.IP"
That means all key is inserted with "vice."

jooskim:
@emlin Thanks for letting me know. I'll follow up with this PR soon.

mdubya66:
what drove the move to vice? that is not something we've agreed upon, as an acronym?

emlin:
@mdubya66 the discussion about this change is in this PR: https://github.com/vmware/vic/pull/1955

The key changed is inside of guestinfo, which is not accessible by end user, but UI rely on that to get VCH appliance information.

jooskim:
FYI, here's what changed and how:

from `guestinfo./init/common/name` to `guestinfo.vice./init/common/name`
from `guestinfo./common/name` to `guestinfo.vice./common/name`
from `guestinfo./repo` to `guestinfo.vice./repo`
from `guestinfo./networks|bridge/ports~` to `guestinfo.vice./networks|bridge/ports~`
from `guestinfo..init.networks|client.ip.IP` to `guestinfo.vice..init.networks|client.ip.IP`

PR #2333, opened 2016-09-15
[skip ci] bump to 1.7.1 - the rest refers to golang:1.7 so no change is needed

caglar10ur:


dougm:
lgtm

cgtexmex:
lgtm

mdubya66:
should we wait to stabilize the CI before merging this? also given we are trying to tag 0.6.0 should we wait until after that happens? @hickeng

mdubya66:
Add [full ci] to this and when it passes it should be fine to merge.

dougm:
This one should have been skip ci, since it only runs in the Vagrant devbox.

mdubya66:
ah.

caglar10ur:
@mdubya66 we already use 1.7.1 since it's released as we use golang:1.7 in our scripts which points to latest 1.7 version

caglar10ur:
♻️

caglar10ur:
♻️

PR #2336, opened 2016-09-15
Default build number to 0 in version

### LINKED ISSUE #2324: Build number not set in ISO version on local build
Build number not defaulted to 0 in `isos/base/utils.sh`

andrewtchin:
Fixes #2324 
@gigawhitlocks @jzt this fixes `vic-machine create` version issues you were seeing

### DIFF
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -55,17 +55,14 @@ func String() string {
 }
 
 func GetBuild() *Build {
-	var b string
 	if BuildNumber == "" {
-		b = "0"
-	} else {
-		b = BuildNumber
+		BuildNumber = "0"
 	}
 	return &Build{
 		Version:     Version,
 		GitCommit:   GitCommit,
 		BuildDate:   BuildDate,
-		BuildNumber: b,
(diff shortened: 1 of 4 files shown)


pkg/version/version.go
```diff
@@ -55,11 +55,17 @@ func String() string {
 }
 
 func GetBuild() *Build {
+	var b string
+	if BuildNumber == "" {
+		b = "0"
```
caglar10ur: Just set BuildNumber to "0", no need to have another variable or an else statement

gigawhitlocks: Also, now that you point this out, where does BuildNumber get defined? It looks like this function should be a method w/ BuildNumber a receiver member.

andrewtchin: good point BuildNumber is set here: https://github.com/vmware/vic/blob/master/Makefile#L57 based on env var which usually comes from drone

jzt:
LGTM

andrewtchin:
♻️

andrewtchin:
♻️

andrewtchin:
lgtm

Issue #1718, opened 2016-07-28
Implement volume ls filtering


mhagen-vmware:
Expected:
docker volume ls -f name=test
Error response from daemon: Invalid filter 'name'

Actual:
DRIVER VOLUME NAME
 vsphere test
1. Not sure if we want to actually block filtering based on name or not, it would seem to be a nice feature that docker doesn't actually support.
2. We seem to be ignoring the filter completely, as no matter what I put in there it actually returns all volumes.

1 is a choice we can make, 2 is more egregious.

stuclem:
Already in the release notes, so removing the kind/note tag.

mhagen-vmware:
lgtm

PR #2339, opened 2016-09-15
Support creating named volumes

### LINKED ISSUE #2317: docker create doesn't handle creation of named data volume
Docker-compose cannot work with compose files that uses named data volumes because we are not creating the named data volume in ContainerCreate.  This was an unexpected path.

anchal-agrawal:
* Adds support for creating named volumes with `docker create -v` and `docker run -v`
* Adds an integration test in `1-4-Docker-Create`

Also cleans up unneeded params in `processAnonymousVolumes` and typos.

`1-4-Docker-Create` and `1-19-Docker-Volume-Create` pass locally.

Fixes #2317

### DIFF
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -220,7 +220,7 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
 	var joinList []volumeFields
 	var err error
 
-	joinList, err = processAnonymousVolumes(&handle, config.Config.Volumes, c.client)
+	joinList, err = processAnonymousVolumes(config.Config.Volumes)
 	if err != nil {
 		return handle, derr.NewErrorWithStatusCode(fmt.Errorf("%s", err), http.StatusBadRequest)
 	}
@@ -474,7 +474,7 @@ func toModelsNetworkConfig(cc types.ContainerCreateConfig) *models.NetworkConfig
 }
 
 //This function is used to turn any call from docker create -v <stuff> into a volumeFields object.
-//the -v has 3 forms. 1: -v <anonymouse mount path>, -v <Volume Name>:<Destination Mount Path>, and -v <Volume Name>:<Destination Mount Path>:<mount flags>
+//the -v has 3 forms. 1: -v <anonymous mount path>, -v <Volume Name>:<Destination Mount Path>, and -v <Volume Name>:<Destination Mount Path>:<mount flags>
 func processVolumeParam(volString string) (volumeFields, error) {
 	volumeStrings := strings.Split(volString, ":")
 	fields := volumeFields{}
@@ -504,7 +504,9 @@ func processVolumeParam(volString string) (volumeFields, error) {
 	return fields, nil
 }
 
-func processAnonymousVolumes(h *string, volumes map[string]struct{}, client *client.PortLayer) ([]volumeFields, error) {
+// processAnonymousVolumes parses fields for implicit volume mappings in a
+// create/run (such as -v /dir) and creates volumes for them.
+func processAnonymousVolumes(volumes map[string]struct{}) ([]volumeFields, error) {
 	var volumeFields []volumeFields
 
 	for v := range volumes {
@@ -516,7 +518,7 @@ func processAnonymousVolumes(h *string, volumes map[string]struct{}, client *cli
 		//NOTE: This should be the guard for the case of an anonymous volume.
 		//NOTE: we should not expect any driver args if the drive is anonymous.
 		log.Infof("anonymous volume being created - Container Create - volume mount section ID: %s ", fields.ID)
-		//
+
 		driverArgs := make(map[string]string)
 		driverArgs["flags"] = fields.Flags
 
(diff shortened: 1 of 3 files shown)


anchal-agrawal:
Ping @matthewavery @mhagen-vmware @sflxn

tests/test-cases/Group1-Docker-Commands/1-4-Docker-Create.robot
```diff
@@ -33,6 +33,19 @@ Create with volume
...
 
+Create with named volume
+    ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} create -v test-named-vol:/testdir busybox
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${output}  Error
+    ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} start ${output}
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${output}  Error
+
+    # TODO: this can be changed to `volume ls -f name=test-named-vol` once filtering by name is supported
```
mhagen-vmware: instead of a TODO here, preference would be for a github check so that we get notified to update when filtering issue is closed.

tests/test-cases/Group1-Docker-Commands/1-4-Docker-Create.md
```diff
@@ -16,31 +16,35 @@ This test requires that a vSphere server is running and available
...
 * Steps 3-7 should all return without error and printing the container ID on return
 * Step 8 should show that the contents of the containers /var/log matches the contents of the hosts /var/log
-* Step 10 should return with the error message - Error: image library/fakeimage not found
-* Step 11 should return with the error message - Error parsing reference: "fakeImage" is not a valid repository/tag
-* Step 14 should result in success and the busy2 container should exist
-* Step 16 should show that busy2 was able to successfully ping busy1 just using the linked name
-* Step 17 should result in success for all three parts
-* Step 18 should return without error
-* Step 19 should return with the following error message - Error response from daemon: No command specified
+* Steps 9, 10 and 11 should return without errors and should successfully create a new volumed called `test-named-vol`
```
mhagen-vmware: volumed -> volume

anchal-agrawal: Good catch! Fixed.

tests/test-cases/Group1-Docker-Commands/1-4-Docker-Create.robot
anchal-agrawal: Addressed. Since #1718 doesn't necessarily involve name filtering, I've added a log instead of failing the test if #1718 is closed.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -530,12 +530,27 @@ func processAnonymousVolumes(h *string, volumes map[string]struct{}, client *cli
 
 func processSpecifiedVolumes(volumes []string) ([]volumeFields, error) {
```
sflxn: Hey, I know you didn't name this, but can you add a function comment explaining a 'specified' volume?  The terminology difference with what docker actually uses may confuse people.

anchal-agrawal: Fixed. I've also added a comment for `processAnonymousVolumes`.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -528,14 +530,31 @@ func processAnonymousVolumes(h *string, volumes map[string]struct{}, client *cli
...
+
+		// Error out if the intended volume is a directory on the client filesystem.
+		if strings.HasPrefix(fields.ID, "/") {
+			return nil, fmt.Errorf("%s does not support mounting directories as a data volume.", ProductName())
+		}
+
+		driverArgs := make(map[string]string)
+		driverArgs["flags"] = fields.Flags
+		_, err = vol.VolumeCreate(fields.ID, "vsphere", driverArgs, nil)
+		if err == nil {
```
gigawhitlocks: So the error isn't handled here if `err != nil`?

anchal-agrawal: That is correct. This implicitly handles the case when a volume with the same name already exists and we don't need to create it - `VolumeCreate` will not create that volume and return an error. I can add error handling for the `VolumeCreate` call to differentiate b/w the error scenarios, but it either succeeds or returns an error with `http.StatusInternalServerError` in all but one cases.

tests/test-cases/Group1-Docker-Commands/1-4-Docker-Create.robot
```diff
@@ -33,6 +33,21 @@ Create with volume
...
 
+Create with named volume
+    ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} create -v test-named-vol:/testdir busybox
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${output}  Error
+    ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} start ${output}
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${output}  Error
+
+    ${status}=  Get State Of Github Issue  1718
```
mhagen-vmware: The problem with warn is that it doesn't really get any attention. We have something like 40 or 50 warns on any ci run. I think it is important to have a fail in order actually get attention paid. If 1718 isnt exactly a perfect fit, maybe create a new issue specifically for it?

anchal-agrawal: Sure, sounds good. I'll create a new issue for filtering volumes by name because a fix for #1718 may not include name filtering since Docker API v1.23 doesn't support it.

anchal-agrawal:
Made the integration test more explicit. Props to @matthewavery and @gigawhitlocks for the awesome test 👍

mhagen-vmware:
LGTM

tests/test-cases/Group1-Docker-Commands/1-4-Docker-Create.robot
```diff
@@ -34,19 +34,8 @@ Create with anonymous volume
...
-    Should Not Contain  ${output}  Error
-
-    ${status}=  Get State Of Github Issue  1718
-    Run Keyword If  '${status}' == 'closed'  Log  Test `Create with named volume` can be improved with volume name filtering, now that #1718 has been resolved  WARN
-    #${rc}  ${output}=  Run And Return Rc And Output  docker ${params} volume ls -f name=test-named-vol
-    ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} volume ls
-    Should Be Equal As Integers  ${rc}  0
-    Should Contain  ${output}  test-named-vol
+    ${disk-size}=  Run  docker ${params} logs $(docker ${params} start $(docker ${params} create -v test-named-vol:/testdir busybox /bin/df -Ph) && sleep 10) | grep by-label | awk '{print $2}'
+    Should Be Equal As Strings  ${disk-size}  975.9M
```
sflxn: This is a very specific value.  Where is this coming from?

sflxn:
LGTM

tests/test-cases/Group1-Docker-Commands/1-4-Docker-Create.robot
anchal-agrawal: https://github.com/vmware/vic/blob/master/tests/test-cases/Group1-Docker-Commands/1-19-Docker-Volume-Create.robot#L11

anchal-agrawal:
♻️

PR #2341, opened 2016-09-15
[full ci] Remove paramName anti-pattern

gigawhitlocks:
While fixing `docker stop` I found an anti-pattern in `backends/container.go`

For whatever reason, the parameter `name` is passed in as a method argument, but then saved off as `paramName` so that the `name` variable can be reused to contain a ContainerID, in a number of methods throughout the file.

I've removed this confusing anti-pattern so that the container ID is simply named `id` at assignment time instead of overwriting `name`, and `name` remains the same as the argument `name` throughout each method. `paramName` is eliminated. The result is code that's easier to follow, since `name` doesn't change meanings halfway through.

I'm not sure how this anti-pattern proliferated, but I've removed all instances of it from this file.

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -349,16 +349,13 @@ func (c *Container) ContainerRename(oldName, newName string) error {
 func (c *Container) ContainerResize(name string, height, width int) error {
 	defer trace.End(trace.Begin(name))
 
-	// save the name provided to us so we can refer to it if we have to return an error
-	paramName := name
-
 	// Look up the container name in the metadata cache to get long ID
 	vc := cache.ContainerCache().GetContainer(name)
 	if vc == nil {
 		return NotFoundError(name)
 
 	}
-	name = vc.ContainerID
+	id := vc.ContainerID
 
 	// Get an API client to the portlayer
 	client := c.containerProxy.Client()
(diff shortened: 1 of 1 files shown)


jzt:
LGTM

stan_small

lib/apiservers/engine/backends/container.go
```diff
@@ -349,30 +349,27 @@ func (c *Container) ContainerRename(oldName, newName string) error {
 func (c *Container) ContainerResize(name string, height, width int) error {
 	defer trace.End(trace.Begin(name))
 
-	// save the name provided to us so we can refer to it if we have to return an error
-	paramName := name
```
caglar10ur: looks like this was intentional and by removing that we are expanding names from short ones to long ones. eg;
docker stop 1 should say 1 is already stopped instead of 1231euayfghhjsegfqw236rgq is already stopped

gigawhitlocks: The name is still stored in `name` and used wherever `paramName` was used

gigawhitlocks: That's the case everywhere, the value in `name` was just saved off so the `name` identifier could be reused -- don't reuse identifiers! It's really confusing!

caglar10ur: Ah yeah, missed the id

caglar10ur:
LGTM

lib/apiservers/engine/backends/container.go
gigawhitlocks: omg @caglar10ur I just busted a gut really loudly in the office at that edit

gigawhitlocks:
:recycle:

Issue #365, opened 2016-03-31
Docker API: Docker Image Remove


sflxn:
Implement docker image remove (docker rmi from CLI) in our docker engine-api server and integrate that with the storage portlayer.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#remove-an-image

Please note, the engine-api's rest server already packaging these json struct into golang structs. The url above is just for reference on what the operation should do.

corrieb:
Moving out of onsite to GA, given that #437 got moved

PR #2346, opened 2016-09-16
[full ci] Rmi

### LINKED ISSUE #253: Handle deletes of images

fdawg4l:
Add vsphere implementation for delete image
    
    * Also includes supporting handlers in the daemon and the PL REST API.
    
    We currently only remove leaf images.  It is possible to remove whole
    branches if they are not in use (see pkg/index for marking of the
    branch), but that can be done later.
    
    This should conclude the work necessary to support rmi
    
    Fixes #253 #365

Also fix broken storage unit tests.

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
@@ -115,6 +115,7 @@ func (handler *StorageHandlersImpl) Configure(api *operations.PortLayerAPI, hand
 	api.StorageGetImageTarHandler = storage.GetImageTarHandlerFunc(handler.GetImageTar)
 	api.StorageListImagesHandler = storage.ListImagesHandlerFunc(handler.ListImages)
 	api.StorageWriteImageHandler = storage.WriteImageHandlerFunc(handler.WriteImage)
+	api.StorageDeleteImageHandler = storage.DeleteImageHandlerFunc(handler.DeleteImage)
 
 	api.StorageVolumeStoresListHandler = storage.VolumeStoresListHandlerFunc(handler.VolumeStoresList)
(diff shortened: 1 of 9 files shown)


lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -167,6 +168,44 @@ func (handler *StorageHandlersImpl) GetImage(params storage.GetImageParams) midd
 	return storage.NewGetImageOK().WithPayload(result)
 }
 
+// GetImage retrieves an image from a store
```
anchal-agrawal: `DeleteImage`

dougm:
lgtm

cgtexmex:
lgtm

fdawg4l:
.

PR #2359, opened 2016-09-16
Update LEMP and voting app tests

sflxn:
Added new Github issues to fix the LEMP and voting app tests.
Updated the test to watch those issues instead of basic compose
test issue.


anchal-agrawal:
LGTM

dougm:
lgtm

PR #2364, opened 2016-09-17
Add docker wait to docker kill [full ci]

### LINKED ISSUE #378: Docker API: Container Kill
Implement the docker kill operation (docker kill from CLI) in the engine-api server and integrate with the exec port layer.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#kill-a-container
https://docs.docker.com/engine/reference/commandline/kill/

sflxn:
Finish up docker kill implementation by adding support for
docker wait if the signal is 0 or sigkill.  Also moved all the
swagger related code for docker kill and docker wait to
container_proxy.go.

Resolves #378

### DIFF
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -37,6 +37,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"syscall"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -78,6 +79,8 @@ type VicContainerProxy interface {
 	CommitContainerHandle(handle, imageID string) error
 	StreamContainerLogs(name string, out io.Writer, started chan struct{}, showTimestamps bool, followLogs bool, since int64, tailLines int64) error
 	ContainerRunning(vc *viccontainer.VicContainer) (bool, error)
+	Wait(vc *viccontainer.VicContainer, timeout time.Duration) (exitCode int32, processStatus string, containerState string, reterr error)
+	Signal(vc *viccontainer.VicContainer, sig uint64) error
 
 	Client() *client.PortLayer
 }
@@ -104,6 +107,7 @@ const (
 	swaggerSubstringEOF                  = "EOF"
 	forceLogType                         = "json-file" //Use in inspect to allow docker logs to work
 	annotationKeyLabels                  = "docker.labels"
+	killWaitForExit        time.Duration = 2 * time.Second
 )
 
 // NewContainerProxy creates a new ContainerProxy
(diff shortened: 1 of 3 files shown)


lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -352,6 +356,99 @@ func (c *ContainerProxy) ContainerRunning(vc *viccontainer.VicContainer) (bool,
...
+	}
+
+	// Get an API client to the portlayer
+	client := PortLayerClient()
+	if client == nil {
+		reterr = InternalServerError("Wait failed to create a portlayer client")
+		return
+	}
+
+	results, err := client.Containers.ContainerWait(containers.NewContainerWaitParamsWithContext(ctx).WithTimeout(int64(timeout.Seconds())).WithID(vc.ContainerID))
```
anchal-agrawal: Consider breaking this up for readability.

sflxn: Yeah, I just moved the code, but I agree, it's a little bit much.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -352,6 +356,102 @@ func (c *ContainerProxy) ContainerRunning(vc *viccontainer.VicContainer) (bool,
...
+	}
+
+	// Get an API client to the portlayer
+	client := PortLayerClient()
+	if client == nil {
+		return InternalServerError("Signal failed to create a portlayer client")
+	}
+
+	if running, err := c.ContainerRunning(vc); !running {
+		if err == nil {
```
hickeng: idomatic would be  collapsing this to a single test clause: `!running && err == nil`

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -352,6 +356,102 @@ func (c *ContainerProxy) ContainerRunning(vc *viccontainer.VicContainer) (bool,
...
+	if sig == 0 || syscall.Signal(sig) == syscall.SIGKILL {
+		params := containers.NewContainerSignalParamsWithContext(ctx).WithID(vc.ContainerID).WithSignal(int64(syscall.SIGKILL))
+		if _, err := client.Containers.ContainerSignal(params); err != nil {
+			switch err.(type) {
+			case *containers.ContainerSignalNotFound:
+				return NotFoundError(vc.ContainerID)
+			case *containers.ContainerSignalInternalServerError:
+				// Container may have received the signal but we got an error.
+				// Give it some time and check again.
+				c.Wait(vc, killWaitForExit)
```
hickeng: If we're attempting to ensure that the container is stopped in this path, then calling Stop on it with a very short delay period is a better way to go. Stop will eventually call PowerOff if all in-guest avenues fail.

sflxn: I see your point, and I started to make the changes, but the problem is this, container.go and container_proxy.go isn't fully refactored yet.  I need to move the Stop() swagger code to container_proxy.go, and it's actually not a simple move.  It would make this PR overly large.  I want to do that in another PR.

hickeng:
LGTM

mdubya66:
lgtm

sflxn:
ping

PR #2365, opened 2016-09-17
enable expvar support and open ports 6062 and 6063

caglar10ur:
when the debug is not 0. This way we can use expvarmon to monitor
services. See https://github.com/divan/expvarmon

```
expvarmon -ports="http://192.168.218.140:6062-6063"
```


dougm:
lgtm

anchal-agrawal:
LGTM

PR #2381, opened 2016-09-19
[full ci] Refactor volume create/join error handling

anchal-agrawal:
Follow-up of #2339

* Moves invalid volume create check to `processVolumeParam` func
* Condenses `processAnonymousVolumes` and `processSpecifiedVolumes` into one func, `processVolumeFields`
* Per @hmahmood, adds a helper func `volumeCreate` to separate the error handling between the `create`/`run` and `volume create` paths
* Adds a 404 response in swagger for a `VolumeJoin` to check if volume was not found
* Adds unit test improvements

`1-4-Docker-Create` and `1-19-Docker-Volume-Create` pass locally.

### DIFF
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -191,7 +191,7 @@ func (c *ContainerProxy) AddContainerToScope(handle string, config types.Contain
 }
 
 // AddVolumesToContainer adds volumes to a container, referenced by handle.
-// If an error is return, the returned handle should not be used.
+// If an error is returned, the returned handle should not be used.
 //
 // returns:
 //	modified handle
@@ -205,21 +205,42 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
 	//Volume Attachment Section
 	log.Debugf("ContainerProxy.AddVolumesToContainer - VolumeSection")
 	log.Debugf("Raw Volume arguments : binds:  %#v : volumes : %#v", config.HostConfig.Binds, config.Config.Volumes)
-	var joinList []volumeFields
-	var err error
 
-	joinList, err = processAnonymousVolumes(config.Config.Volumes)
+	// Collect all volume mappings. In a docker create/run, they
+	// can be anonymous (-v /dir) or specific (-v vol-name:/dir).
+	volumes := config.HostConfig.Binds
+	for anonVol := range config.Config.Volumes {
+		volumes = append(volumes, anonVol)
+	}
+
+	volList, err := processVolumeFields(volumes)
 	if err != nil {
 		return handle, BadRequestError(err.Error())
 	}
 
-	volumeSubset, err := processSpecifiedVolumes(config.HostConfig.Binds)
-	if err != nil {
-		return handle, BadRequestError(err.Error())
-	}
-	joinList = append(joinList, volumeSubset...)
+	// Create and join volumes.
+	for _, fields := range volList {
 
-	for _, fields := range joinList {
+		driverArgs := make(map[string]string)
+		driverArgs["flags"] = fields.Flags
+
+		vol := &Volume{}
+		_, req, err := vol.volumeCreate(fields.ID, "vsphere", driverArgs, nil)
+		if err != nil {
+			switch err := err.(type) {
+			case *storage.CreateVolumeConflict:
+				// Implicitly ignore the error where a volume with the same name
+				// already exists. We can just join the said volume to the container.
+				log.Infof("a volume with the name %s already exists", fields.ID)
+			case *storage.CreateVolumeNotFound:
+				return handle, VolumeCreateNotFoundError(req.Store) //derr.NewErrorWithStatusCode(fmt.Errorf("No volume store named (%s) exists", volRequest.Store), http.StatusInternalServerError)
+			default:
+				return handle, InternalServerError(err.Error()) //derr.NewErrorWithStatusCode(fmt.Errorf("%s", err), http.StatusInternalServerError)
+			}
+		} else {
+			log.Infof("volumeCreate succeeded. Volume mount section ID: %s", fields.ID)
+		}
+
 		flags := make(map[string]string)
 		//NOTE: for now we are passing the flags directly through. This is NOT SAFE and only a stop gap.
 		flags["Mode"] = fields.Flags
@@ -236,6 +257,8 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
 				return handle, InternalServerError(err.Payload.Message)
 			case *storage.VolumeJoinDefault:
 				return handle, InternalServerError(err.Payload.Message)
+			case *storage.VolumeJoinNotFound:
+				return handle, VolumeJoinNotFoundError(err.Payload.Message) //derr.NewErrorWithStatusCode(fmt.Errorf(err.Payload.Message), http.StatusNotFound)
 			default:
 				return handle, InternalServerError(err.Error())
 			}
@@ -248,7 +271,6 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
 }
 
 // CommitContainerHandle() commits any changes to container handle.
-//
 func (c *ContainerProxy) CommitContainerHandle(handle, imageID string) error {
 	defer trace.End(trace.Begin(handle))
 
@@ -455,14 +477,21 @@ func toModelsNetworkConfig(cc types.ContainerCreateConfig) *models.NetworkConfig
 	return nc
 }
 
-//This function is used to turn any call from docker create -v <stuff> into a volumeFields object.
-//the -v has 3 forms. 1: -v <anonymous mount path>, -v <Volume Name>:<Destination Mount Path>, and -v <Volume Name>:<Destination Mount Path>:<mount flags>
+// processVolumeParam is used to turn any call from docker create -v <stuff> into a volumeFields object.
+// The -v has 3 forms. -v <anonymous mount path>, -v <Volume Name>:<Destination Mount Path> and
+// -v <Volume Name>:<Destination Mount Path>:<mount flags>
 func processVolumeParam(volString string) (volumeFields, error) {
 	volumeStrings := strings.Split(volString, ":")
 	fields := volumeFields{}
 
-	//This switch determines which type of -v was invoked.
-	switch len(volumeStrings) {
+	// Error out if the intended volume is a directory on the client filesystem.
+	numVolParams := len(volumeStrings)
+	if numVolParams > 1 && strings.HasPrefix(volumeStrings[0], "/") {
+		return volumeFields{}, InvalidVolumeError{}
+	}
+
+	// This switch determines which type of -v was invoked.
+	switch numVolParams {
 	case 1:
 		VolumeID, err := uuid.NewUUID()
 		if err != nil {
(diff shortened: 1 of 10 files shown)


anchal-agrawal:
Needs review from @matthewavery.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -473,14 +481,22 @@ func toModelsNetworkConfig(cc types.ContainerCreateConfig) *models.NetworkConfig
...
 func processVolumeParam(volString string) (volumeFields, error) {
 	volumeStrings := strings.Split(volString, ":")
 	fields := volumeFields{}
 
-	//This switch determines which type of -v was invoked.
-	switch len(volumeStrings) {
+	// Error out if the intended volume is a directory on the client filesystem.
+	volStrLen := len(volumeStrings)
+	if volStrLen > 1 && strings.HasPrefix(volumeStrings[0], "/") {
+		invalidVolumeErr := fmt.Errorf("%s does not support mounting directories as a data volume.", ProductName())
```
fdawg4l: This smells like it should be its own type that satisfies the `error` interface which we can then type assert.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -473,14 +481,22 @@ func toModelsNetworkConfig(cc types.ContainerCreateConfig) *models.NetworkConfig
...
+// The -v has 3 forms. -v <anonymous mount path>, -v <Volume Name>:<Destination Mount Path> and
+// -v <Volume Name>:<Destination Mount Path>:<mount flags>
 func processVolumeParam(volString string) (volumeFields, error) {
 	volumeStrings := strings.Split(volString, ":")
 	fields := volumeFields{}
 
-	//This switch determines which type of -v was invoked.
-	switch len(volumeStrings) {
+	// Error out if the intended volume is a directory on the client filesystem.
+	volStrLen := len(volumeStrings)
```
fdawg4l: This variable name is misleading.  It's not the `strlen` of a string;  it's the len of the slice.  `numVarParam` or `numElem` or something like that would be more appropriate.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -532,32 +548,31 @@ func processAnonymousVolumes(volumes map[string]struct{}) ([]volumeFields, error
...
 // processSpecifiedVolumes parses fields for volumes specified explicitly during a
 // create/run (such as -v volume-name:/dir) and creates them if they don't already exist.
-func processSpecifiedVolumes(volumes []string) ([]volumeFields, error) {
-	var volumeFields []volumeFields
+func processSpecifiedVolumes(volumes []string) ([]volumeFields, error, error) {
+	var (
+		volumeFields []volumeFields
+		createErr    error
+	)
 	vol := &Volume{}
```
fdawg4l: This variable isn't returned and can be scoped to where it's used.  Can you move this closer to where we need it?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -532,32 +548,31 @@ func processAnonymousVolumes(volumes map[string]struct{}) ([]volumeFields, error
 
 // processSpecifiedVolumes parses fields for volumes specified explicitly during a
 // create/run (such as -v volume-name:/dir) and creates them if they don't already exist.
-func processSpecifiedVolumes(volumes []string) ([]volumeFields, error) {
-	var volumeFields []volumeFields
+func processSpecifiedVolumes(volumes []string) ([]volumeFields, error, error) {
```
fdawg4l: Just repeating the above;  lets try to avoid returning more than one `error`.  If you need to find the condition the error is being thrown from, then it sounds like you need a specific error type to assert.

lib/apiservers/engine/backends/container_proxy.go
fdawg4l: Also, the variable itself looks like its only used once.  You could simply remove the variable all together and evaluate the len where it's needed.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -225,9 +225,9 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
 		return handle, derr.NewErrorWithStatusCode(fmt.Errorf("%s", err), http.StatusBadRequest)
 	}
 
-	volumeSubset, err := processSpecifiedVolumes(config.HostConfig.Binds)
-	if err != nil {
-		return handle, derr.NewErrorWithStatusCode(fmt.Errorf("%s", err), http.StatusBadRequest)
+	volumeSubset, processErr, createErr := processSpecifiedVolumes(config.HostConfig.Binds)
```
anchal-agrawal: Thanks @fdawg4l! I'm currently refactoring the `AddVolumesToContainer` func so I don't need to return multiple errors from `processSpecifiedVolumes` while still keeping track of the errors returned from `VolumeCreate`.

lib/apiservers/engine/backends/container_proxy.go
anchal-agrawal: Fixed!

lib/apiservers/engine/backends/container_proxy.go
anchal-agrawal: Done.

lib/apiservers/engine/backends/container_proxy.go
anchal-agrawal: Fixed.

lib/apiservers/engine/backends/container_proxy_test.go
```diff
@@ -42,27 +43,7 @@ func TestProcessVolumeParams(t *testing.T) {
...
 	assert.Equal(t, "r", processedTestVolumes[2].Flags)
-}
-
-func TestProcessSpecifiedVolumes(t *testing.T) {
-	rawTestVolumes := []string{"masterVolume:/blah", "testVolume:/mount:r", "specVol:/mount/path:r"}
-	var processedTestVolumes []volumeFields
-
-	processedFields, err := processSpecifiedVolumes(rawTestVolumes)
-	assert.Nil(t, err)
-	processedTestVolumes = append(processedTestVolumes, processedFields...)
```
hmahmood: Can we incorporate these test cases in the other tests? Or rewrite this test so it uses the new functions?

lib/apiservers/engine/backends/volume.go
```diff
@@ -98,14 +98,14 @@ func (v *Volume) VolumeInspect(name string) (*types.Volume, error) {
 	return nil, fmt.Errorf("%s does not implement volume.VolumeInspect", ProductName())
 }
 
-// VolumeCreate : docker personality implementation for VIC
-func (v *Volume) VolumeCreate(name, driverName string, driverArgs, labels map[string]string) (*types.Volume, error) {
-	defer trace.End(trace.Begin("Volume.VolumeCreate"))
+// volumeCreate issues a CreateVolume request to the portlayer
+func (v *Volume) volumeCreate(name, driverName string, driverArgs, labels map[string]string) (*types.Volume, *models.VolumeRequest, error) {
```
hmahmood: Why are we returning the request object?

lib/apiservers/engine/backends/volume.go
```diff
@@ -116,12 +116,24 @@ func (v *Volume) VolumeCreate(name, driverName string, driverArgs, labels map[st
...
 	req, varErr := newVolumeCreateReq(name, driverName, driverArgs, labels)
 	if varErr != nil {
-		return result, derr.NewErrorWithStatusCode(varErr, http.StatusBadRequest)
+		return result, req, derr.NewErrorWithStatusCode(varErr, http.StatusBadRequest)
 	}
 	log.Infof("Finalized model for volume create request to portlayer: %#v", req)
 
 	res, err := client.Storage.CreateVolume(storage.NewCreateVolumeParamsWithContext(ctx).WithVolumeRequest(req))
 	if err != nil {
+		return result, req, err
```
hmahmood: In the error case you should return `nil, err`

lib/apiservers/engine/backends/volume.go
anchal-agrawal: It is needed by both callers for their error messages.

lib/apiservers/engine/backends/container_proxy_test.go
anchal-agrawal: The new func, `processVolumeFields`, simply calls `processVolumeParam` on a slice. There's already a test for the latter, and I added a new test case where an error is expected.

anchal-agrawal: Also, the cases in `TestProcessSpecifiedVolumes` are part of  `TestProcessVolumeParams`.

lib/apiservers/engine/backends/volume.go
anchal-agrawal: In an error case, `result` would be a ptr to an empty `types.Volume` struct.

matthewavery:
LGTM

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -217,21 +217,51 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
...
+			switch err := err.(type) {
+			case *storage.CreateVolumeConflict:
+				// Implicitly ignore the error where a volume with the same name
+				// already exists. We can just join the said volume to the container.
+				log.Infof("a volume with the name %s already exists", fields.ID)
+
+			case *storage.CreateVolumeNotFound:
+				return handle, derr.NewErrorWithStatusCode(fmt.Errorf("No volume store named (%s) exists", req.Store), http.StatusInternalServerError)
+
+			// Handle non-portlayer error
```
hmahmood: Forgot to remove this?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -217,21 +217,51 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
...
+	for _, fields := range volList {
+
+		driverArgs := make(map[string]string)
+		driverArgs["flags"] = fields.Flags
+
+		// NOTE: calling volumeCreate regardless of whether the volume is already
+		// present can be avoided by adding an extra optional param to VolumeJoin,
+		// which would then call volumeCreate if the volume does not exist.
+		vol := &Volume{}
+		_, req, err := vol.volumeCreate(fields.ID, "vsphere", driverArgs, nil)
```
hmahmood: I thought you were going to remove `req`?

lib/apiservers/engine/backends/container_proxy.go
anchal-agrawal: Sorry about that! Fixed.

lib/apiservers/engine/backends/container_proxy.go
anchal-agrawal: After chatting with @matthewavery, I decided to keep it. The raw error from swagger isn't CLI-friendly and keeping `req` helps us show the exact parameters used in the volume create request.

hmahmood: Again those parameters might not match up with the command line, and confuse the caller.

matthewavery: He either has to have request or he has to access the driverargs again when returning that error. However, the exact arguments passed through to the portlayer are in the request and could be scrubbed during the personality process before the call to the portlayer. The `req`'s values are the most accurate for what the failuer could have been.

hmahmood: Log the generated params for the request, and just return the error from the port layer.

mdubya66:
This will need a [full ci] before merged.

hmahmood:
lgtm

anchal-agrawal:
♻️

PR #2383, opened 2016-09-19
No need to waste time installing on commented out test

mhagen-vmware:
Just noticed we were still installing the appliance in this test even though we don't run the test. Waste of time, let's not do that right now...


jzt:
LGTM

mdubya66:
lgtm

mhagen-vmware:
.

PR #2399, opened 2016-09-20
[full ci] [WIP] Strip vmware.log of sensitive information

### LINKED ISSUE #2164: vic-admin: strip vmware.log
vmware.log for the appliance contains the extraconfig entries for the VM - this needs to be sanitized by vic-admin prior to bundling.

https://bugzilla.eng.vmware.com/show_bug.cgi?id=1727636

sgairo:
Fixes #2164

Reads the log line by line and writes to temporary File if the line contains no sensitive data.

### DIFF
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -16,12 +16,17 @@ package main
 
 import (
 	"archive/zip"
+	"bufio"
+	"bytes"
 	"compress/gzip"
 	"crypto/tls"
 	"html/template"
+	"io"
 	"net"
 	"net/http"
+	"os"
 	"path/filepath"
+	"strings"
 
 	"golang.org/x/net/context"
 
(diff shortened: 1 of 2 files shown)


hickeng:
@sgairo This is heading in a direction I did not anticipate. I was expecting it to apply to:
a. just the vmware.log associated with the appliance VM, not including the containerVMs.
b. be a simple stream filter - once we introduce tmp files things get significantly more involved.

Let's go for a simpler option - don't collect the vmware.log file for the applianceVM. We can get GSS to request it explicitly if we need it.

sgairo:
@hickeng alright, will adjust.

jzt:
@hickeng The temp file was my idea for an iterative approach building up to the point where we would get a stream filter in place. Overall this is a rather non-trivial problem that we would really like to solve in a way that is pluggable and reusable for different needs. I asked @sgairo to throw up a WIP PR so that @hmahmood and I could give feedback as we iterate.  Apologies for any confusion.

Issue #2372, opened 2016-09-19
Property collector error: context deadline exceeded on container start


mhagen-vmware:
Saw this issue on a recent functional run on a Nimbus VC cluster with 3 ESX hosts. Probably related to #1822 , but I am going to leave that closed for the time being.

From portlayer.log:
```
time="2016-09-19T15:14:58Z" level=debug msg="Found 1 subscribers to vsphere.VmEvent: fervent_bhaskar-1379b2830c2c417e926a0fa924bb6c4f60d432443a4dae6c5085eae25a924648 on  10.160.120.204 in ha-datacenter is powered off" 
time="2016-09-19T15:14:58Z" level=debug msg="Event manager calling back to exec" 
time="2016-09-19T15:14:58Z" level=debug msg="Container(1379b2830c2c417e926a0fa924bb6c4f60d432443a4dae6c5085eae25a924648) state set to Stopped via event activity" 
time="2016-09-19T15:14:58Z" level=debug msg="Container(1379b2830c2c417e926a0fa924bb6c4f60d432443a4dae6c5085eae25a924648) closing 0 log followers" 
time="2016-09-19T15:14:58Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/portlayer/exec.(*Container).Update] Container.Update" 
time="2016-09-19T15:17:44Z" level=error msg="Property collector error: context deadline exceeded" 
```

From docker-personality.log:
```
time="2016-09-19T15:14:42Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).ContainerStart] 1379b2830c2c417e926a0fa924bb6c4f60d432443a4dae6c5085eae25a924648" 
time="2016-09-19T15:17:44Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).ContainerStart] [3m2.034466734s] 1379b2830c2c417e926a0fa924bb6c4f60d432443a4dae6c5085eae25a924648" 
time="2016-09-19T15:17:44Z" level=error msg="Handler for POST /v1.23/containers/1379b2830c2c417e926a0fa924bb6c4f60d432443a4dae6c5085eae25a924648/start returned error: Server error from portlayer: unable to wait for process launch status: context deadline exceeded" 
```

[...]

Issue #2327, opened 2016-09-15
Certain tether start failures may not propagate to port layer


dougm:
Container.start powers on the vm and waits for the `$id.started` key to be set in the vm extra config.  It is expected the value of this key will be "true" for success, anything else is considered an error:
https://github.com/vmware/vic/blob/0f6cb8929a99500f9ce634450480a47d417993c7/lib/portlayer/exec/container.go#L241-L244

`WaitForKeyInExtraConfig` ignores updates where the value of this key is set to "" or "&lt;nil&gt;".  If the tether fails to start without setting this value, the vm powers off and portlayer continues to wait.  In the past this surfaced as "Context Deadline Exceeded", but as of #2315 removing the timeout, will block forever.

The specific error seen in the tether debug log:

```
failed to apply network endpoint config: unable to acquire reference to link 192: more than one eth interface matches /sys/bus/pci/devices/0000:00:16.0/0000:*:00.0/net/*"
```

There are multiple issues here:

* We need to use a timeout in WaitForKeyInExtraConfig, just in case.

* Tether needs to set the SessionConfig.Started field in case(s) such as this

* The specific PCI path mapping error above

* Reduce mutex locking in the port layer; without the timeout, the mutex continues to hold, making it impossible to docker rm the container as it will block waiting to acquire the mutex

dougm:
Timeout is covered by #2332 

Let's focus next on fixing the cases where the tether SessionConfig.Started field isn't being set.

We can create separate issues for the other 2 items later.

dougm:
Regarding "more than one eth interface matches", the message is misleading.. the problem is that there are *zero* matches.  Short term fix would be to retry as we do for disk device PCI paths:

https://github.com/vmware/vic/blob/fde9102c937986487e8735e79a34f9a1a12abee5/pkg/vsphere/disk/disk_manager.go#L108

mhagen-vmware:
Note: This error results in a failure to start a container. To workaround, wait until the system is less loaded and try again.

dougm:
Retry checking for the ethernet device PCI path did not help.  Looking at the output of `lsmod` (via tether init), the vmxnet3 driver is *not* loaded when this happens.  Adding a call to `modprobe vmxnet3` in the tether seems to avoid the problem.

dougm:
Root of this issue is fixed in #2407 - where the fix is to explicitly load the vmxnet3 driver.

stuclem:
This is fixed. Removing kind/note.

PR #2405, opened 2016-09-20
Avoid waiting on guestinfo if vm powered off

dougm:
If the tether fails to initialize, for example when no ethernet card is
found, the container VM will power off.  With this change we stop
waiting on guestinfo keys in such an event.

Issue #2327


jzt:
LGTM

jzt:
This PR also fixes #2372. 👍

hmahmood:
LGTM

dougm:
:recycle:

PR #2406, opened 2016-09-21
Establish stdin connection immediately after attach [full ci]

### LINKED ISSUE #2353: Change attach in personality and portlayer to establish stdin connection without waiting on client inputs
Currently, our attach connection for stdin between the personality server and the portlayer rest server is established lazily.  Swagger won't make that first call to the PL until there is data in the io.writer we pass to the Swagger client.  We want to update this link so the stdin connection gets made immediately after ContainerAttach is called.  We can do this by injecting some initialization bytes into the pipe we set up on the personality side.  On the portlayer rest side, we should ignore these bytes and not send them onto the tether.

sflxn:
Formerly, the attach performed a lazy connection between the personality
and the portlayer servers for the stdin stream.  The reason is the
Swagger client will not make the initial connection unless there is data
present in the ReaderCloser given to the client.  We force the connection
by sending over an init byte sequence, which causes an immediate connection
to the portlayer.  The interaction portlayer was updated to ignore the initial byte
sequence.

Also, since I touched the attach code again, I went ahead and moved
all the attach and resize swagger related code to the _proxy.go file.

Resolves #2353

**NOTE!!!: A lot of this code is the moving swagger code for resize and attach over to the _proxy.go file.  The relevant code is in container_proxy.go:copyStdIn() and in interaction_handler.go.**

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -26,11 +26,6 @@ import (
 
 	"golang.org/x/net/context"
 
-	"github.com/go-swagger/go-swagger/httpkit"
-	httptransport "github.com/go-swagger/go-swagger/httpkit/client"
-	strfmt "github.com/go-swagger/go-swagger/strfmt"
-	"github.com/mreiferson/go-httpclient"
-
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/docker/api/types/backend"
 	derr "github.com/docker/docker/errors"
@@ -51,7 +46,6 @@ import (
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	viccontainer "github.com/vmware/vic/lib/apiservers/engine/backends/container"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/portmap"
-	"github.com/vmware/vic/lib/apiservers/portlayer/client"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/containers"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/interaction"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/scopes"
@@ -343,29 +337,13 @@ func (c *Container) ContainerResize(name string, height, width int) error {
 	vc := cache.ContainerCache().GetContainer(name)
 	if vc == nil {
 		return NotFoundError(name)
-
 	}
-	id := vc.ContainerID
-
-	// Get an API client to the portlayer
-	client := c.containerProxy.Client()
 
 	// Call the port layer to resize
 	plHeight := int32(height)
 	plWidth := int32(width)
-	plResizeParam := interaction.NewContainerResizeParamsWithContext(ctx).WithID(id).WithHeight(plHeight).WithWidth(plWidth)
 
-	_, err := client.Interaction.ContainerResize(plResizeParam)
-	if err != nil {
-		if _, isa := err.(*interaction.ContainerResizeNotFound); isa {
-			return ResourceNotFoundError(name, "interaction connection")
-		}
-
-		// If we get here, most likely something went wrong with the port layer API server
-		return InternalServerError(err.Error())
-	}
-
-	return nil
+	return c.containerProxy.Resize(vc, plHeight, plWidth)
 }
 
 // ContainerRestart stops and starts a container. It attempts to
@@ -1089,7 +1067,7 @@ func (c *Container) ContainerAttach(name string, ca *backend.ContainerAttachConf
 		}
 	}
 
-	err = attachStreams(context.Background(), vc, clStdin, clStdout, clStderr, ca)
+	err = c.containerProxy.AttachStreams(context.Background(), vc, clStdin, clStdout, clStderr, ca)
 	if err != nil {
 		if _, ok := err.(DetachError); ok {
 			log.Infof("Detach detected, tearing down connection")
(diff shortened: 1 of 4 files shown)


sflxn:
@caglar10ur Here is the behavior you were requesting.

caglar10ur:
Oh, I know I requested that but wasn't expected that much of code moving around. This conflicts very badly with my branch so my review/test will take some time.

mdubya66:
This should be [full ci] before merged

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -218,17 +222,66 @@ type FlushingReader struct {
...
+		total := nr
+		for {
+			if total < initLen {
+				if er != nil {
+					// Got an error, zero out len and break out to return error back
+					nr = 0
+					break
+				}
+				nr, er = d.Read(buf[total:])
+				total += nr
```
hickeng: This is going to prevent us from ever sending `< len(d.initBytes)` as the first input.

This is written with an assumption that the other end of the data channel is sending the correct data, with basic precautions. We can either:
a. assume that the source is aware of how to behave and simply discard the first X bytes
b. code defensively in the expectation that it doesn't.

In the latter case we should be comparing what we read in each iteration of the loop as that allows us to detect, in a timely fashion, unexpected usage.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -238,8 +291,9 @@ func (d *FlushingReader) AddFlusher(flusher GenericFlusher) {
 func (d *FlushingReader) WriteTo(w io.Writer) (written int64, err error) {
 	buf := make([]byte, 64)
 
+	nr, er := d.readDetectInit(buf)
```
hickeng: It's permitted for Read/WriteTo/ReadFrom to return `0, nil` - I'm not aware of when that will occur with stdlib functions.
Placing this here means we will attempt to filter initial bytes for every call to WriteTo, instead of the initial bytes for the Reader as a whole - is this intentional?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -368,6 +376,119 @@ func (c *ContainerProxy) createNewAttachClientWithTimeouts(connectTimeout, respo
...
+	log.Infof("container attach:  cleaned up connections to %s.", vc.ContainerID)
+	for err := range errors {
+		if err != nil {
+			// If we get here, most likely something went wrong with the port layer API server
+			// These errors originate within the go-swagger client itself.
+			// Go-swagger returns untyped errors to us if the error is not one that we define
+			// in the swagger spec.  Even EOF.  Therefore, we must scan the error string (if there
+			// is an error string in the untyped error) for the term EOF.
+
+			log.Errorf("container attach error: %s", err.Error())
```
vburenin: It looks like any error that happened is already logged. I wonder if there any reason to log first error that occurred and return it?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -895,3 +1016,167 @@ func portMapFromVicContainer(vc *viccontainer.VicContainer) nat.PortMap {
...
+
+	log.Debugf("* stdout attach start %s", time.Now().Format(time.UnixDate))
+	getStdoutParams := interaction.NewContainerGetStdoutParamsWithContext(ctx).WithID(name).WithDeadline(&swaggerDeadline)
+	_, err := pl.Interaction.ContainerGetStdout(getStdoutParams, clStdout)
+	log.Debugf("* stdout attach end %s", time.Now().Format(time.UnixDate))
+	if err != nil {
+		if _, ok := err.(*interaction.ContainerGetStdoutNotFound); ok {
+			return NotFoundError(name)
+		}
+		if _, ok := err.(*interaction.ContainerGetStdoutInternalServerError); ok {
```
vburenin: Does it make sense to check this type if it returns InternalServerError that is the same as return after this if?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -895,3 +1016,167 @@ func portMapFromVicContainer(vc *viccontainer.VicContainer) nat.PortMap {
...
+	getStderrParams := interaction.NewContainerGetStderrParamsWithContext(ctx).WithID(name)
+	_, err := pl.Interaction.ContainerGetStderr(getStderrParams, clStderr)
+
+	if err != nil {
+		if _, ok := err.(*interaction.ContainerGetStderrNotFound); ok {
+			return NotFoundError(name)
+		}
+
+		if _, ok := err.(*interaction.ContainerGetStderrInternalServerError); ok {
+			return NotFoundError(name)
```
vburenin: The similar case was returning InternalServerError, is it a copy/paste accidental issue?

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
sflxn: I'm still not clear what scenario you're trying to address.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
sflxn: WriteTo() is only called once for the stream.  io.copy() only calls WriteTo() once.

lib/apiservers/engine/backends/container_proxy.go
sflxn: Good point.

lib/apiservers/engine/backends/container_proxy.go
sflxn: good point.

caglar10ur:
would it be possible to split the work into multiple commits so that we could see the changes instead of big file moves around?

caglar10ur:
actually nevermind my previous comment. Looked at it with a fresh brain and LGTM

lib/apiservers/engine/backends/container.go
```diff
@@ -344,29 +338,14 @@ func (c *Container) ContainerResize(name string, height, width int) error {
...
-			return ResourceNotFoundError(name, "interaction connection")
-		}
 
-		// If we get here, most likely something went wrong with the port layer API server
-		return InternalServerError(err.Error())
-	}
-
-	return nil
+	err := c.containerProxy.Resize(vc, plHeight, plWidth)
+	return err
```
caglar10ur: [cosmetic] you could return directly

lib/apiservers/engine/backends/container.go
```diff
@@ -1090,7 +1069,8 @@ func (c *Container) ContainerAttach(name string, ca *backend.ContainerAttachConf
 		}
 	}
 
-	err = attachStreams(context.Background(), vc, clStdin, clStdout, clStderr, ca)
+	err = c.containerProxy.AttachStreams(context.Background(), vc, clStdin, clStdout, clStderr, ca)
+
```
caglar10ur: [cosmetic] drop the extra line

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -548,6 +559,110 @@ func (c *ContainerProxy) createNewAttachClientWithTimeouts(connectTimeout, respo
 	return plClient, transport
 }
 
+func (c *ContainerProxy) Resize(vc *viccontainer.VicContainer, height, width int32) error {
+	defer trace.End(trace.Begin(vc.ContainerID))
+
+	if c.client == nil {
+		return derr.NewErrorWithStatusCode(fmt.Errorf("ContainerProxy failed to create a portlayer client"),
```
caglar10ur: Please use InternalServerError helper

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1177,3 +1292,304 @@ func labelsFromAnnotations(config *container.Config, annotations map[string]stri
...
 }
+
+//------------------------------------
+// ContainerAttach() Utility Functions
+//------------------------------------
+
+// attacheStreams takes the the hijacked connections from the calling client and attaches
+// them to the 3 streams from the portlayer's rest server.
+// clStdin, clStdout, clStderr are the hijacked connection
+func attachStreams(ctx context.Context, vc *viccontainer.VicContainer, clStdin io.ReadCloser, clStdout, clStderr io.Writer, ca *backend.ContainerAttachConfig) error {
```
caglar10ur: you can drop this fn as it's already above

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -548,6 +559,110 @@ func (c *ContainerProxy) createNewAttachClientWithTimeouts(connectTimeout, respo
...
+			}
+
+			cancel()
+			errors <- err
+		}()
+	}
+
+	// Wait for all stream copy to exit
+	wg.Wait()
+
```
caglar10ur: This is not the same function and not reflects the new code, please use the function below

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
sflxn: Added code to kick out of monitoring the primer/init bytes if it doesn't look like we're getting them.

lib/apiservers/engine/backends/container.go
sflxn: done.

lib/apiservers/engine/backends/container.go
sflxn: done.

lib/apiservers/engine/backends/container_proxy.go
sflxn: I'd rather wait on this.  The InternalServerError message is wrong.  We need to go through and clean those up again.  This isn't an internal error from portlayer.  We need finer grain error messages in errors.go.

lib/apiservers/engine/backends/container_proxy.go
sflxn: thanks for finding this rebase issue

lib/apiservers/engine/backends/container_proxy.go
sflxn: thanks for finding this rebase issue

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
sflxn: Modified this so that we only check for init bytes if this reader is created with init bytes.  Otherwise, it uses normal d.read.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -339,17 +340,75 @@ type FlushingReader struct {
...
 }
 
+// readDetectInit() is used by WriteTo() which is used by io.Copy.  It attempts
+// to detect a init byte buffer.  If it finds that init byte sequence, it is
+// ignored.  This reader does not care about the init sequeunce.  The init sequence
+// maybe used by the higher level interaction, which in this case is the Swagger
+// establishing initial connection for stdin.
+func (d *FlushingReader) readDetectInit(buf []byte) (int, error) {
+	nr, er := d.Read(buf)
+	if d.initBytes != nil {
```
hickeng: This duplicates the d.initBytes != nil check before calling this method. You could place both of those checks in here.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -339,17 +340,75 @@ type FlushingReader struct {
...
+				}
+				nr, er = d.Read(buf[total:])
+				total += nr
+			} else {
+				// Got enough bytes to compare init bytes so break out
+				break
+			}
+		}
+
+		if er == nil {
```
hickeng: I think we should be checking this irrespective of `err == nil`

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -339,17 +340,75 @@ type FlushingReader struct {
...
+					nr = 0
+					break
+				}
+				if bytes.Compare(d.initBytes[0:total], buf[0:total]) != 0 {
+					// First bytes aren't part of init bytes so client must not be
+					// the docker personality so break and ignore looking for the
+					// init bytes.  Set nr to total and break out.
+					log.Debugf("Did not find primer bytes, stopping watch")
+					nr = total
+					break
```
hickeng: `return total, er`

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
sflxn: yeah

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -339,17 +340,75 @@ type FlushingReader struct {
...
+func (d *FlushingReader) readDetectInit(buf []byte) (int, error) {
+	nr, er := d.Read(buf)
+	if d.initBytes != nil {
+		initLen := len(d.initBytes)
+		total := nr
+		for {
+			if total < initLen {
+				if er != nil {
+					// Got an error, zero out len and break out to return error back
+					nr = 0
```
sflxn: Not clear from docs, but ok.  If this was like most c/c++ calls, nr would be undetermined, but ok.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
sflxn: yeah.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -339,17 +340,70 @@ type FlushingReader struct {
...
 	d.flusher = flusher
 }
 
+// readDetectInit() is used by WriteTo() which is used by io.Copy.  It attempts
+// to detect a init byte buffer.  If it finds that init byte sequence, it is
+// ignored.  This reader does not care about the init sequeunce.  The init sequence
+// maybe used by the higher level interaction, which in this case is the Swagger
+// establishing initial connection for stdin.
+func (d *FlushingReader) readDetectInit(buf []byte) (int, error) {
+	nr, er := d.Read(buf)
```
hickeng: Sorry, it bugs me when I'm sure there's a clear and more concise way of writing something, so:
```

// fast path - len(nil) return 0
if len(d.initBytes) == 0 {
    return d.Read(buf)
}

initLen := len(d.initBytes)
total := 0
for total < initLen {
    nr, er := d.Read(buf) 
    total += nr

    if bytes.Compare(d.initBytes[0:total], buf[0:total]) != 0 {
        // First bytes aren't part of init bytes so client must not be
        // the docker personality so break and ignore looking for the
        // init bytes.
        log.Debugf("Did not find primer bytes, stopping watch")
        return total, er
    }

    if er != nil && total < initLen{
        // you're right - we've got no idea if these were init bytes or
        // not at this point so don't return data
        return 0, err
    }    
}

// would have returned in the compare clause if not matching init bytes
copy(buf[0:], buf[initLen:])
log.Debugf("Found primer bytes, port layer client might be personality server")

// no risk of returning <0
return total - initLen, er
```

sflxn: Alright, I agree it is more concise.

hickeng:
LGTM ;)

sflxn:
ping

sflxn:
ping

PR #2421, opened 2016-09-21
Fix guestinfo wait poweroff shortcut

dougm:
Shortcut was added in PR #2405 - possible for the update callback
included both properties.  In that case, we'd miss the powerState change.


cgtexmex:
lgtm

anchal-agrawal:
LGTM

PR #2426, opened 2016-09-22
Container Start / Stop time

### LINKED ISSUE #1899: Empty start/finish date in inspected container

cgtexmex:
Container start / stop time persisted to the container vmx.  The following docker commands have been updated with this information:

- ps
```
vic ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS                            PORTS               NAMES
55ab3a4adfc5        busybox             "/bin/top"          28 seconds ago      Up 25 seconds                                         insane_poitras
df7d0a04de20        busybox             "/bin/top"          3 minutes ago       Exited (143) About a minute ago                       elegant_albattani
```
- inspect
```
{
        "Id": "df7d0a04de20fa47eb6371a7189ec3013673ab12625e65728c658109e5b8e46f",
        "Created": "2016-09-22T15:12:40Z",
        "Path": "/bin/top",
        "Args": [],
        "State": {
            "Status": "exited",
            "Running": false,
            "Paused": false,
            "Restarting": false,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 0,
            "ExitCode": 143,
            "Error": "",
            "StartedAt": "2016-09-22T15:13:06Z",
            "FinishedAt": "2016-09-22T15:14:32Z"
        }
```

Known issue:

- Out of Band powerOn, powerOff result in incorrect time, so no time will be shown

fixes #1899

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -45,6 +45,7 @@ import (
 	dnetwork "github.com/docker/engine-api/types/network"
 	timetypes "github.com/docker/engine-api/types/time"
 	"github.com/docker/go-connections/nat"
+	"github.com/docker/go-units"
 	"github.com/docker/libnetwork/portallocator"
 
 	"github.com/vishvananda/netlink"
@@ -94,7 +95,6 @@ func init() {
 	clientIfaceName = l.Attrs().Name
 }
 
-//TODO: gotta be a better way...
 // type and funcs to provide sorting by created date
 type containerByCreated []*types.Container
 
@@ -811,13 +811,12 @@ func (c *Container) ContainerWait(name string, timeout time.Duration) (int, erro
 	}
 
 	containerInfo := results.Payload
-
 	// call to the dockerStatus function to retrieve the docker friendly exitCode
-	// TODO: once started / finished time are available replace time.Now()
+	// we are only calling for the exitCode, thus the zero time
 	exitCode, _ := dockerStatus(int(*containerInfo.ProcessConfig.ExitCode),
 		*containerInfo.ProcessConfig.Status,
 		*containerInfo.ContainerConfig.State,
-		time.Now(), time.Now())
+		time.Time{}, time.Time{})
 
 	return exitCode, nil
 }
@@ -827,41 +826,43 @@ func (c *Container) ContainerWait(name string, timeout time.Duration) (int, erro
 //
 // exitCode is the container process exit code
 // status is the container process status -- stored in the vmx file as "started"
-// started / finished are timestamps, but are not yet implemented
+// started & finished are the process start / finish times
 func dockerStatus(exitCode int, status string, state string, started time.Time, finished time.Time) (int, string) {
-	// TODO: once started / finished are available then github.com/docker/go-units must be imported for use in
-	// the status string
-	switch state {
-	case "Created":
-		status = state
-	case "Running":
-		//TODO: once we have started use commented out status
-		// status = fmt.Sprintf("Up %s", units.HumanDuration(time.Now().UTC().Sub(started)))
-		status = state
-	case "Stopped":
-		// interrogate the process status returned from the portlayer
-		// and based on status text and exit codes set the appropriate
-		// docker exit code
-		if strings.Contains(status, "permission denied") {
-			exitCode = 126
-		} else if strings.Contains(status, "no such") {
-			exitCode = 127
-		} else if status == "true" && exitCode == -1 {
-			// most likely the process was killed via the cli
-			// or recieved a sigkill
-			exitCode = 137
-		} else if status == "" && exitCode == 0 {
-			// the process was stopped via the cli
-			// or recieved a sigterm
-			exitCode = 143
-		}
 
-		// TODO: once we have finished use commented out status
-		// status = fmt.Sprintf("Exited (%d) %s ago", exitCode, units.HumanDuration(time.Now().UTC().Sub(finished)))
-		status = fmt.Sprintf("Exited (%d)", exitCode)
+	// set docker status to state and we'll change if needed
+	dockStatus := state
+
+	switch state {
+	case "Running":
+		// if we don't have a start date leave the status as the state
+		if !started.IsZero() {
+			dockStatus = fmt.Sprintf("Up %s", units.HumanDuration(time.Now().UTC().Sub(started)))
+		}
+	case "Stopped":
+		// if we don't have a finished date then don't process exitCode and return "Stopped" for the status
+		if !finished.IsZero() {
+			// interrogate the process status returned from the portlayer
+			// and based on status text and exit codes set the appropriate
+			// docker exit code
+			if strings.Contains(status, "permission denied") {
+				exitCode = 126
+			} else if strings.Contains(status, "no such") {
+				exitCode = 127
+			} else if status == "true" && exitCode == -1 {
+				// most likely the process was killed via the cli
+				// or recieved a sigkill
+				exitCode = 137
+			} else if status == "" && exitCode == 0 {
+				// the process was stopped via the cli
+				// or recieved a sigterm
+				exitCode = 143
+			}
+
+			dockStatus = fmt.Sprintf("Exited (%d) %s ago", exitCode, units.HumanDuration(time.Now().UTC().Sub(finished)))
+		}
 	}
 
-	return exitCode, status
+	return exitCode, dockStatus
 }
 
 // docker's container.monitorBackend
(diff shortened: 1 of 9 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -94,7 +95,6 @@ func init() {
 	clientIfaceName = l.Attrs().Name
 }
 
-//TODO: gotta be a better way...
```
caglar10ur: resistance is futile :P

lib/apiservers/engine/backends/container.go
```diff
@@ -813,13 +813,13 @@ func (c *Container) ContainerWait(name string, timeout time.Duration) (int, erro
 	}
 
 	containerInfo := results.Payload
-
+	started := time.Unix(*containerInfo.ProcessConfig.StartTime, 0)
```
caglar10ur: do we need to check containerInfo != nil && containerInfo.ProcessConfig != nil here?

lib/apiservers/engine/backends/container.go
```diff
@@ -902,12 +895,14 @@ func (c *Container) ContainerInspect(name string, size bool, version version.Ver
 			return nil, InternalServerError(err.Error())
 		}
 	}
+
+	started := time.Unix(*results.Payload.ProcessConfig.StartTime, 0)
```
caglar10ur: same question about nil's?

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -277,7 +278,11 @@ func (handler *ContainersHandlersImpl) GetContainerListHandler(params containers
 		info := models.ContainerListInfo{}
 		info.ContainerID = &container.ExecConfig.ID
 		info.LayerID = &container.ExecConfig.LayerID
-		info.Created = &container.ExecConfig.Created
+		info.CreateTime = &container.ExecConfig.CreateTime
+		started := container.ExecConfig.Sessions[*info.ContainerID].StartTime
```
caglar10ur: Do we need to check the existence or they are guaranteed to be in the map?

lib/config/executor/container_vm.go
```diff
@@ -185,6 +188,8 @@ type SessionConfig struct {
 	// Diagnostics holds basic diagnostics data
 	Diagnostics Diagnostics `vic:"0.1" scope:"read-only" key:"diagnostics"`
 
+	// Details Status `vic:"0.1" scope:"read-write" key:"details"`
```
caglar10ur: Debug artifact?

lib/portlayer/exec/container.go
```diff
@@ -192,7 +190,6 @@ func (c *Container) Commit(ctx context.Context, sess *session.Session, h *Handle
 		}
 
 		c.State = *h.State
-		commitEvent = events.ContainerStopped
```
caglar10ur: Just wondering why do we remove those?

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
cgtexmex: it will be there..worse case it will be zero (default value)

lib/apiservers/engine/backends/container.go
cgtexmex: Probably a good idea -- I'll make the change

lib/apiservers/engine/backends/container.go
cgtexmex: Ditto

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
cgtexmex: If we've created a container -- which we have at this point -- the session will always exist, so the StartTime will at a minimum be 0

lib/portlayer/exec/container.go
cgtexmex: This change introduced a "Creating" state -- so the need to update the handle state in commit was eliminated.  

The elimination of the stopped was a mistake -- THANKS!

lib/portlayer/exec/container.go
```diff
@@ -174,9 +175,6 @@ func (c *Container) Commit(ctx context.Context, sess *session.Session, h *Handle
 		c.vm = vm.NewVirtualMachine(ctx, sess, res.Result.(types.ManagedObjectReference))
 		c.State = StateCreated
 
-		// align the handle state w/the container
```
hmahmood: Why did we take this out?

lib/portlayer/exec/handle.go
```diff
@@ -147,6 +147,19 @@ func (h *Handle) Commit(ctx context.Context, sess *session.Session, waitTime *in
 	// make sure there is a spec
 	h.SetSpec(nil)
 	cfg := make(map[string]string)
+
+	// Set timestamps based on target state
+	switch *h.State {
```
hmahmood: Should this be after the `Commit` call below?

lib/portlayer/exec/container.go
cgtexmex: I introduced the "Creating" state so it wasn't needed anymore -- previously when you created a containerVM the state was "Stopped" thus causing the issue.

lib/portlayer/exec/handle.go
cgtexmex: No - I have it specifically before the commit so it will be added to the VM spec -- then during the reconfigure ops for start / stop it will make it to the VMX

lib/portlayer/exec/handle.go
```diff
@@ -179,8 +192,6 @@ func (h *Handle) Create(ctx context.Context, sess *session.Session, config *Cont
 
 	// update the handle with Metadata
 	h.ExecConfig = config.Metadata
-	// add create time to config
```
hmahmood: Why did we take this out?

lib/tether/tether.go
```diff
@@ -315,28 +314,16 @@ func (t *tether) handleSessionExit(session *SessionConfig) {
...
-	logCount := len(logs)
-	if logCount >= MaxDeathRecords {
-		logs = logs[logCount-MaxDeathRecords+1:]
-	}
-
 	// Remove associated PID file
 	cmdname := path.Base(session.Cmd.Path)
 	_ = os.Remove(fmt.Sprintf("%s.pid", path.Join(PIDFileDir(), cmdname)))
 
-	session.Diagnostics.ExitLogs = append(logs, executor.ExitLog{
```
hmahmood: Do we not need these anymore?

lib/portlayer/exec/handle.go
cgtexmex: It moved to the createHandler -- so when the config is created it's populated there

lib/portlayer/exec/handle.go
hmahmood: Could `h.State == nil` here?

lib/tether/tether.go
cgtexmex: It wasn't being populated and after discussion w/ @hickeng we agreed to remove and use a single attribute

lib/portlayer/exec/handle.go
cgtexmex: No -- when the handle is created it's state is set -- so worst case it will be unKnown

cgtexmex: Yes -- they won't be exact...but the time difference should be minimal and acceptable.

hmahmood: Why isn't the tether updating these?

cgtexmex: The order of operations creates a challenge -- that's what I originally started to do, but the reconfigures on start / stop would overwrite the tether data.  So the path of least resistance was this -- we can look at it later to change, but this is a good starting point

hmahmood: Alright; can we open an issue to track this?

cgtexmex: Yep

hmahmood:
lgtm

cgtexmex:
rebase

jzt:
♻️  LGTM ♻️

PR #2433, opened 2016-09-22
Sanitize all the govc calls to support either VC or ESXi

### LINKED ISSUE #2371: Docker ps Remove container OOB to support VC
Probably add ${vm}/${vm}

mhagen-vmware:
Fixes #2371


dougm:
lgtm

Another option would be to just have a suite variable, "vch-path".  On ESX set it to ".", on VC set to "vch-name".  Then use `govc foo-command ${vch-path}/${container-name}`

matthewavery:
LGTM

mhagen-vmware:
.

Issue #2416, opened 2016-09-21
Add a warning in the UI installer for vCenter 5.5 users


jooskim:
Now that we're not supporting vSphere 5.5 in MVP GA any warning or comment in scripts as well as UI code regarding vCenter 5.5 should be removed.

mdubya66:
are they harming anything? what if we decided we wanted to add support again in the future?

jooskim:
There isn't much that will break or harm to any existing environment we have here, as the only difference I had to accommodate for v5.5 was the installation path of the Web Client plugin. The plugin code itself will remain the same. If we want to bring this back in the future for VC 5.5 (by when it would look and behave exactly the same still), I can put the existing version into a separate branch that I could freeze.

jooskim:
@mdubya66 Following up with your latest email, we can keep the scripts as they are. I was wondering if we needed to remove anything that slightly makes the user feel v5.5 is supported, say, by having the installer script ask "Are you running vSphere 5.5?".

mdubya66:
let's play it by ear. if you want to remove the UI plugins that should be ok but i'd like to keep vic-machine working on 5.5 since we don't have anything to prevent it from operating. Customers just won't get support.

jooskim:
Ok, what I could do is leave the code (actual plugin code and installer) as it is, but add a warning that VC 5.5 is not officially supported and that it should be used with their own risk. As you may know the plugin itself actually works anyway. It's just the lack of automated test framework that was being a problem.

mdubya66:
whatever you are comfortable with. thanks.

PR #2434, opened 2016-09-22
Move vCenter version check to configs

jooskim:
Made a change to the UI installer in a way that vCenter 5.5 support is still there, but needs to be configured in the `configs` file which comes with a warning message that VC 5.5 is not officially supported.

Addresses #2416.

### DIFF
--- a/ui/installer/VCSA/configs
+++ b/ui/installer/VCSA/configs
@@ -9,4 +9,8 @@ VCENTER_IP=""
 VIC_UI_HOST_URL="NOURL"
 
 # If VIC_UI_HOST_URL starts with "https", enter SHA-1 thumbprint of the web server hosting the vic-ui.zip file
+# The fingerprint should be in the format of A0:B1:C2:... In case you retrieved the fingerprint from Google Chrome make sure to replace spaces with colons (:)
 VIC_UI_HOST_THUMBPRINT=""
+
+# WARNING: VIC UI can run on a vCenter 5.5 setup, but is not officially supported. Use it with your own risk. If you wish to continue change the value from 0 to 1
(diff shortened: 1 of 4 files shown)


ui/installer/VCSA/configs
```diff
@@ -10,3 +10,6 @@ VIC_UI_HOST_URL="NOURL"
 
 # If VIC_UI_HOST_URL starts with "https", enter SHA-1 thumbprint of the web server hosting the vic-ui.zip file
 VIC_UI_HOST_THUMBPRINT=""
```
andrewtchin: side note if it's not documented anywhere else, what format does this expect as i've seen fingerprints with spaces, no spaces, or `:`

andrewtchin:
lgtm

ui/installer/VCSA/configs
jooskim: @andrewtchin great point. i think the space character is the expected delimiter, but i'll double check. also i'll give it try with `:` to see if VC doesn't like it. thanks for your input

jooskim: turns out that VC only takes a fingerprint with `:`s. i'll make that clear in the configs file and also possibly let Stuart be aware of it.

jooskim: @stuclem it would be great if you could also make a note in the documentation that the fingerprint should be delimited by colons not space. google chrome gives a fingerprint that has spaces in it, which doesn't work with VC. so i've made a note in the `configs` file both for VCSA and vCenter for Windows.

anchal-agrawal:
:shipit:

jooskim:
♻️

PR #2440, opened 2016-09-22
swallow the error if there are no readers

caglar10ur:
Introduce a new SSH request type named CloseStdinReq and use it to signal tether and force it to send EOF to the process.

Introduce a portlayer handler for closing the stdin.

Refactor the overall code a bit, fix the tests add a new reattach test.

Lastly fix some of the races reported by the golang race detector.


pkg/dio/multi.go
```diff
@@ -192,6 +192,11 @@ func (t *multiReader) Read(p []byte) (int, error) {
 		}
 	}
 
+	// return 0, nil if there are no readers
+	if len(t.readers) == 0 {
```
hmahmood: How would this be true given the for loop above at line 164?

caglar10ur: On the detach, the SSH connection sends a EOF and drops the connection, since t.err is io.EOF on that case we don't block but we also don't want to propagate that error to the caller hence we swallow it

hmahmood: Hm.. not sure if I follow. The only way you get out of the for loop above is if `t.readers != 0 || t.err != nil`. The `t.err != nil` is handled right after the for loop. That leaves `t.readers != 0`, in which case you go through the other for loop at line 174, at the end of which isn't `t.readers != 0`? Or am I seeing things?

dougm: I'm having trouble seeing how this is possible too.

hickeng: @hmahmood t.readers can be modified by Remove - this will likely occur while Read is blocked on line 182.

hmahmood: @hickeng is that safe to do? We are in the middle of iterating over it.

hickeng: yes, but @caglar10ur was going to alter this to use a tmpvar anyway, as with the Writer impl.

caglar10ur:
@hickeng @hmahmood can I get another pass?

pkg/dio/multi.go
```diff
@@ -43,9 +43,11 @@ func (t *multiWriter) Write(p []byte) (int, error) {
 	var n int
 	var err error
 
+	t.mutex.Lock()
 	// stash a local copy of the slice as we never want to write twice to a single writer
 	// if remove is called during this flow
 	wTmp := t.writers
```
hmahmood: Is this a copy?

pkg/dio/multi.go
```diff
@@ -174,14 +175,17 @@ func (t *multiReader) Read(p []byte) (int, error) {
 	for len(t.readers) == 0 && t.err == nil {
 		t.cond.Wait()
 	}
+	// stash a copy of the readers slie to iterate later
+	rTmp := t.readers
```
hmahmood: ditto

hmahmood:
never mind the copy comments; that works.

lgtm

pkg/dio/multi.go
caglar10ur: should be, https://play.golang.org/p/m-Mp9Y0VpY

cmd/tether/attach.go
```diff
@@ -382,6 +387,16 @@ func (t *attachServerSSH) channelMux(in <-chan *ssh.Request, process *os.Process
 					log.Errorf("Failed to dispatch signal to process: %s\n", err)
 				}
 			}
+		case msgs.CloseStdinReq:
+			msg := msgs.CloseStdinMsg{}
```
hickeng: There's no data sent over this channel so we can drop the unmarshalling and structure definitions.

cmd/tether/msgs/messages.go
```diff
@@ -121,6 +121,25 @@ func (s *SignalMsg) FromString(name string) error {
 	return nil
 }
 
+// CloseStdinMsg
+const CloseStdinReq = "close-stdin"
+
+type CloseStdinMsg struct {
```
hickeng: No need for this - we send nil for the payload.

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -216,6 +220,23 @@ func (i *InteractionHandlersImpl) ContainerSetStdinHandler(params interaction.Co
 	return interaction.NewContainerSetStdinOK()
 }
 
+// ContainerCloseStdinHandler closes the stdin
+func (i *InteractionHandlersImpl) ContainerCloseStdinHandler(params interaction.ContainerCloseStdinParams) middleware.Responder {
+	defer trace.End(trace.Begin(params.ID))
+
+	contConn, err := i.attachServer.Get(context.Background(), params.ID, 0)
+	if err != nil {
```
hickeng: Let's have invalid or illegal state - must be attached to call CloseStdin

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -216,6 +220,23 @@ func (i *InteractionHandlersImpl) ContainerSetStdinHandler(params interaction.Co
 	return interaction.NewContainerSetStdinOK()
 }
 
+// ContainerCloseStdinHandler closes the stdin
```
hickeng: Prerequisites in the godoc please.

lib/portlayer/attach/client.go
```diff
@@ -111,6 +113,22 @@ func (t *attachSSH) Signal(signal ssh.Signal) error {
 	return nil
 }
 
+func (t *attachSSH) CloseStdin() error {
+	defer trace.End(trace.Begin(""))
+
+	msg := msgs.CloseStdinMsg{Noop: true}
+	ok, err := t.channel.SendRequest(msgs.CloseStdinReq, false, msg.Marshal())
```
hickeng: send `nil` instead of the noop struct

lib/portlayer/attach/client.go
```diff
@@ -111,6 +113,22 @@ func (t *attachSSH) Signal(signal ssh.Signal) error {
...
+	if err == nil && !ok {
+		return fmt.Errorf("unknown error")
```
hickeng: let's change this to "unknown error closing stdin" - otherwise if the string appears many levels of call stack from here and is logged it's going to be hard to determine where/why.

pkg/dio/multi.go
```diff
@@ -191,20 +205,25 @@ func (t *multiReader) Read(p []byte) (int, error) {
...
-		} else if err != io.EOF {
-			// if there was an actual error, return that
-			// we cannot handle multiple not EOF errors, so return now
-			t.err = err
-			return n, err
+		if err != nil {
+			if err == io.EOF {
+				// increment the EOF counter and remove the reader that retured EOF
+				log.Debugf("[%p] returned EOF", r)
+				t.Remove(r)
```
hickeng: Comment that remove takes locks

pkg/dio/multi.go
```diff
@@ -191,20 +205,25 @@ func (t *multiReader) Read(p []byte) (int, error) {
...
+				return n, err
+			}
 		}
 	}
 
-	// we'd have returned anything other than EOF/nil inline
-	t.err = eof
-	return n, eof
+	// This means readers closed/removed while we iterate
+	if n == 0 && t.err == nil && eof == len(rTmp) {
```
hickeng: `if eof != 0 && ....` - otherwise we'll get this even when no eof was returned every time we remove the last writer.

cmd/tether/attach.go
caglar10ur: Done

cmd/tether/attach.go
```diff
@@ -382,6 +387,16 @@ func (t *attachServerSSH) channelMux(in <-chan *ssh.Request, process *os.Process
...
+			if err = msg.Unmarshal(req.Payload); err != nil {
+				ok = false
+				log.Errorf(err.Error())
+			} else {
+				log.Infof("Received CloseStdinMsg. Closing the dio")
```
caglar10ur: Done

cmd/tether/attach_test.go
```diff
@@ -711,3 +677,110 @@ func TestMockAttachTetherToPL(t *testing.T) {
...
+			done <- true
+		}()
+
+		// write something to echo
+		log.Debug("sending test data")
+		sshSession.Stdin().Write(testBytes)
+		log.Debug("sent test data")
+	}
+
+	// attach
```
caglar10ur: Done

cmd/tether/attach_test.go
```diff
@@ -711,3 +677,110 @@ func TestMockAttachTetherToPL(t *testing.T) {
...
+	testBytes = []byte("\x1b[32mhello world - again\x1b[39m!\n")
+
+	// attach
+	attach()
+
+	// wait for the close to propagate
+	<-done
+	// exit
+	sshSession.CloseStdin()
+
```
caglar10ur: Done

cmd/tether/msgs/messages.go
caglar10ur: Done

lib/portlayer/attach/client.go
caglar10ur: Done

lib/portlayer/attach/client.go
caglar10ur: Done

pkg/dio/multi.go
caglar10ur: Done

pkg/dio/multi.go
caglar10ur: Done

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
```diff
@@ -154,68 +139,106 @@ func (i *InteractionHandlersImpl) UnbindHandler(params interaction.InteractionUn
...
-	_, err = io.Copy(contConn.Stdin(), detachableIn)
+	_, err = io.Copy(session.Stdin(), detachableIn)
 	if err != nil {
-		err = fmt.Errorf("Error copying stdin (id:%s): %s", params.ID, err.Error())
 		log.Errorf("%s", err.Error())
+
+		e := &models.Error{
+			Message: fmt.Sprintf("Error copying stdin (id: %s): %s", params.ID, err.Error()),
+		}
+		return interaction.NewContainerSetStdinInternalServerError().WithPayload(e)
```
caglar10ur: @sflxn any idea why we weren't returning an error here? Just an oversight or an intentional choice?

caglar10ur:
♻️

caglar10ur:
This PR also fixes (some of) the races detected by golang race detector.

mdubya66:
lgtm

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
caglar10ur: Done

lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go
caglar10ur: Done

fdawg4l:
lgtm

PR #2443, opened 2016-09-22
populate container cache in docker personality during init

### LINKED ISSUE #2382: Repopulate container cache in docker personality after VCH restart

emlin:
Fixes #2382 

In this change, I unified the response object of container list and container inspect in portlayer API. So container list will return enough information to rebuild container cache.

Another thing is that the container repopulated method end up in backend.go, instead of container cache file. The reason is that @sflxn is refactoring container_proxy file, I don't want to move that file around at this moment.

### DIFF
--- a/lib/apiservers/engine/backends/backends.go
+++ b/lib/apiservers/engine/backends/backends.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"net"
 	"net/url"
+	"sync"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -25,10 +26,12 @@ import (
 	httptransport "github.com/go-swagger/go-swagger/httpkit/client"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client"
+	"github.com/vmware/vic/lib/apiservers/portlayer/client/containers"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/misc"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/storage"
 	"github.com/vmware/vic/lib/apiservers/portlayer/models"
 	"github.com/vmware/vic/lib/config"
+	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/vsphere/sys"
 )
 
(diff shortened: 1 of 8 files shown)


lib/apiservers/engine/backends/backends.go
```diff
@@ -90,13 +93,34 @@ func Init(portLayerAddr, product string, config *config.VirtualContainerHostConf
...
+	results := make(chan error, initNum)
+
 	log.Info("Refreshing image cache")
 	go func() {
+		defer wg.Done()
 		if err := cache.ImageCache().Update(portLayerClient); err != nil {
-			log.Warn("Failed to refresh image cache: %s", err)
+			err = errors.Errorf("Failed to refresh image cache: %s", err)
+			log.Warn(err)
+			results <- err
```
hmahmood: Would this not promote this warning to an error (on line 146)? Previously the code did not treat this as an error.

hmahmood:
LGTM except for promoting that warning to error, but that should be easily rectified.

lib/apiservers/engine/backends/backends.go
emlin: @hmahmood so you think it should still be a warning, instead of error?
But in theory, the error only happens while portlayer has problem, e.g. query from vsphere. At this moment, I think it might be better to stop the initialization.

emlin: Confirmed with @hmahmood and @jzt, will keep existing behavior. thanks

lib/apiservers/engine/backends/backends.go
```diff
@@ -93,12 +95,20 @@ func Init(portLayerAddr, product string, config *config.VirtualContainerHostConf
...
 			return
 		}
 		log.Info("Image cache updated successfully")
 	}()
 
+	log.Info("Refreshing container cache")
+	go func() {
+		if err := syncContainerCache(portLayerClient); err != nil {
+			log.Warnf("Failed to refresh container cache: %s", err)
+		}
```
andrewtchin: does this need a return or an else to prevent the success message?

emlin: good point, thanks

jzt:
LGTM

andrewtchin:
lgtm

emlin:
♻️

PR #2446, opened 2016-09-22
detect the detach operation and propogate it

caglar10ur:
This will be used for on-demand serial work later...

/cc @sflxn

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -123,13 +123,13 @@ func BadRequestError(msg string) error {
 	return derr.NewErrorWithStatusCode(fmt.Errorf("Bad request error from portlayer: %s", msg), http.StatusBadRequest)
 }
 
-func (c *Container) Handle(name, paramName string) (string, error) {
-	resp, err := c.containerProxy.Client().Containers.Get(containers.NewGetParamsWithContext(ctx).WithID(name))
+func (c *Container) Handle(id, name string) (string, error) {
+	resp, err := c.containerProxy.Client().Containers.Get(containers.NewGetParamsWithContext(ctx).WithID(id))
 	if err != nil {
 		switch err := err.(type) {
 		case *containers.GetNotFound:
-			cache.ContainerCache().DeleteContainer(name)
-			return "", NotFoundError(paramName)
+			cache.ContainerCache().DeleteContainer(id)
+			return "", NotFoundError(name)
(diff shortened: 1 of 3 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -534,9 +534,7 @@ func (c *Container) containerStart(name string, hostConfig *containertypes.HostC
 	if bind {
 		e := c.findPortBoundNetworkEndpoint(hostConfig, endpoints)
 		if err = c.mapPorts(portmap.Map, hostConfig, e); err != nil {
-			err = fmt.Errorf("error mapping ports: %s", err)
-			log.Error(err)
```
fdawg4l: This looked somewhat useful.  Is the err still being logged somewhere?

caglar10ur: Hmm not really. I'll move that logging into the mapPorts function though as the rest of the errors gets logged there

caglar10ur: Hmm wait, we log it when we get the InternalServer error on the other side like the rest of the errors.

lib/apiservers/engine/backends/container.go
```diff
@@ -1119,36 +1133,53 @@ func attachStreams(ctx context.Context, vc *viccontainer.VicContainer, clStdin i
...
 
-	log.Infof("container attach:  cleaned up connections to %s.", vc.ContainerID)
+	// close the channel so that we don't leak (if there is an error)/or get blocked (if there are no errors)
+	close(errors)
+
+	log.Infof("cleaned up connections to %s. Checking errors", vc.ContainerID)
 	for err := range errors {
 		if err != nil {
+			// check if we got DetachError
+			if _, ok := err.(DetachError); ok {
```
fdawg4l: Will we get a `DetachError` from all of the `go func`s above or is it possible only 1 returns this error, while the other 2 return CDE?  Just asking out of curiosity.

caglar10ur: Nope, we are only intercepting stdin

lib/apiservers/engine/backends/container.go
fdawg4l: Cool.  Thanks!

fdawg4l:
LGTM

dougm:
lgtm

caglar10ur:
♻️

PR #2454, opened 2016-09-23
[full ci] Introduce logging and interaction subsystems

caglar10ur:
for the portlayer. They use file/network backed serial ports right now.

Also remove serial related code from the spec package


lib/apiservers/portlayer/swagger.json
```diff
@@ -2408,6 +2686,138 @@
...
 			}
+		},
+		"LoggingJoinConfig": {
+			"type": "object",
+			"required": [
+				"handle"
+			],
+			"properties": {
+				"handle": {
+					"type": "object"
```
hmahmood: Should this be a string? I know we wanted to pass around a serialized version of the handle. Plus other swagger endpoints use `handle` as a string.

lib/portlayer/attach/attach.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/lib/portlayer/constants"
+	"github.com/vmware/vic/lib/portlayer/exec"
+	"github.com/vmware/vic/pkg/trace"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// Handler defines the interface
+type Handler interface {
```
hmahmood: Maybe just called it `Attach`? Handler is very HTTPy

lib/portlayer/attach/attach.go
```diff
@@ -0,0 +1,169 @@
...
+	if len(serials) > 1 {
+		return nil, fmt.Errorf("Multiple matches found with desired backing")
+	}
+
+	ip, err := lookupVCHIP()
+	if err != nil {
+		return nil, err
+	}
+
+	for i := range serials {
```
hmahmood: Go question: I have seen you use this form of iterating over a slice before. What's wrong with 
```go
for _, serial := range serials
```

lib/portlayer/attach/attach.go
```diff
@@ -0,0 +1,169 @@
...
+			Operation: types.VirtualDeviceConfigSpecOperationEdit,
+		}
+		handle.Spec.DeviceChange = append(handle.Spec.DeviceChange, config)
+	}
+
+	return handle, nil
+}
+
+// Join adds network backed serial port to the caller and configures them
+func Join(h interface{}) (interface{}, error) {
```
hmahmood: Why not just use `exec.Handle`?

lib/portlayer/attach/attach.go
```diff
@@ -0,0 +1,169 @@
...
+	handle.SetSpec(nil)
+	// get the virtual device list
+	devices := object.VirtualDeviceList(handle.Container.Config.Hardware.Device)
+
+	// select the virtual serial ports
+	serials := devices.SelectByBackingInfo((*types.VirtualSerialPortURIBackingInfo)(nil))
+	if len(serials) == 0 {
+		return nil, fmt.Errorf("Unable to bind a device with desired backing")
+	}
+	if len(serials) > 1 {
```
dougm: Looks like we could either remove this check or just use serials[0] below instead of the for loop?

lib/apiservers/portlayer/swagger.json
caglar10ur: I think the other  endpoints also need to change to object type so that we end up with interface{} instead of string. Assuming the string ID won't have any meaning when we split the services from each other and start moving full spec around

cc @hickeng

lib/portlayer/attach/attach.go
caglar10ur: In the latter case serial is a copy of serials[i]

lib/portlayer/attach/attach.go
caglar10ur: Same reason above

lib/portlayer/attach/attach.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	return handle, nil
+}
+
+// Join adds network backed serial port to the caller and configures them
+func Join(h interface{}) (interface{}, error) {
+	handle, ok := h.(*exec.Handle)
+	if !ok {
+		return nil, fmt.Errorf("type assertion failed")
```
dougm: Maybe in these cases we could include `%T` in the message, to see the actual type of the handle.

lib/portlayer/exec/container.go
```diff
@@ -119,6 +127,27 @@ func (c *Container) NewHandle() *Handle {
...
+	}
+	ctx := context.TODO()
+
+	if err := c.vm.Properties(ctx, c.vm.Reference(), []string{"config", "runtime"}, &o); err != nil {
+		return err
+	}
+
+	c.Config = o.Config
+	c.Runtime = &o.Runtime
+
```
hmahmood: Not updating `c.ExecConfig`?

lib/apiservers/portlayer/swagger.json
hickeng: @hmahmood we started with string because a serialized form can be represented as a string - after asking Ivan about how to return an untyped value he pointed us at this form.
If the Go API returns `interface{}` then it's simple to have the base layer returning rich objects, have another layer returning serialized forms, have another layer returning "pointers" as we are atm.
If we use string directly then we lose either the flexibility or the ability to layer shims of various types over the basic impl.

lib/portlayer/attach/attach.go
hmahmood: Is this interface used anywhere? Apologies if I missed this.

caglar10ur: Yeah I'm open to naming suggestion and nope it's not used anywhere. Initially there was a helper that returns a struct that holds those methods (hence the interface) but then @hickeng wanted them as directly exported methods. 

Long story short, I can remove the definition if you want :)

lib/portlayer/attach/attach.go
caglar10ur: Ah yeah, will do that

lib/portlayer/exec/container.go
caglar10ur: c.ExecConfig is ExecutorConfig right, not sure why should I touch it. Am I missing something?

lib/portlayer/attach/attach.go
caglar10ur: Removed

lib/portlayer/attach/attach.go
caglar10ur: Done

lib/portlayer/attach/attach.go
caglar10ur: Used %#+v to dump the whole thing

lib/portlayer/exec/container.go
hmahmood: You are basically sync'ing the container's state with what the port layer has cached. Things like assigned ip addresses may have changed in the ExecutorConfig object on the container. These need to returned as output to several commands, like `network inspect` and `inspect`.

lib/apiservers/portlayer/swagger.json
hmahmood: So I get enabling rich handle types. What I don't quite like is passing in `interface{}` to port layer api's. Can we have a base type for handles (or can we assume the handle types share behavior/attributes, like they must be serializable)? Also we are doing `object -> interface{} -> string -> exec.Handle -> interface{} -> exec.Handle` (from http request to port layer api call). That seems a little ridiculous to me. Is `object -> interface{} -> Handle` not possible?

lib/portlayer/exec/container.go
caglar10ur: I understand that things can change but I'm not sure how this new code is changing what we return in those cases? Whoever was handling those should be still doing it. 

We are calling this 2 get two things, the runtime.PowerState so that we can modify ExtraConfig and later we will start to get ChangeVersion attribute for making API calls and ensuring nothing changed since we last checked.

hmahmood: Its not changing anything because we do not currently updating this before returning the ip address for instance. For DHCP networks there is a chance these ip addresses will be stale unless we fetch the current ExecutorConfig. I was going to add something like this function to fix these stale ips.

lib/apiservers/portlayer/swagger.json
caglar10ur: Removed some of the assertions

caglar10ur: Actually I'll introduce them back and instead and do that exec.GetHandle surgery in a different PR. They are related but Handle refactoring has nothing to do with 2 new subsystems.

dougm:
lgtm

lib/portlayer/exec/container.go
caglar10ur: Done

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -277,12 +274,12 @@ func (handler *ContainersHandlersImpl) GetContainerListHandler(params containers
 	}
 
 	containerVMs := exec.Containers.Containers(state)
-	containerList := make([]models.ContainerInfo, 0, len(containerVMs))
```
caglar10ur: This change is here because of the swagger validation change this PR introduces

lib/apiservers/engine/backends/backends.go
```diff
@@ -207,7 +207,7 @@ func syncContainerCache(client *client.PortLayer) error {
 	}
 	cc := cache.ContainerCache()
 	for _, info := range containme.Payload {
-		container := ContainerInfoToVicContainer(info)
+		container := ContainerInfoToVicContainer(*info)
```
caglar10ur: This change is here because of the swagger validation change this PR introduces

caglar10ur:
♻️

caglar10ur:
♻️

caglar10ur:
: recycle

PR #2469, opened 2016-09-26
Fixes the failing test that looks for the tether debug logs

matthewavery:
This fixes a small integration test failure that was introduced due to the incorrect robot keyword being used when attempting to apply a regex match. Now the lookup on the name of the tether debug file for robot test group 9-1 should pass. This has been tested on VC and ESX. Not on VSAN though.


anchal-agrawal:
LGTM

mhagen-vmware:
LGTM

PR #2471, opened 2016-09-26
Fixes vic-machine delete removing attached disks to powered off VMs

### LINKED ISSUE #2114: vic-machine delete should not delete volume vmdk attached to container VM
@matthewavery @hickeng 
Currently, vic-machine delete will delete container VMs if VM status is powered off, so the attached volume VMDK will be removed as well, even if --force is not specified. (--force is not good to define volume delete, but that's another story)

vic-machine delete should do similar thing with container rm in portlayer.

matthewavery:
Previously we were calling the heavy handed vm.Destroy which removes the vm and any devices attached to it. Now we will be calling DeleteExceptDisks which removes disks from the VM before calling Destroy on it. 

Fixes #2114


andrewtchin:
lgtm

emlin:
cool. lgtm 👍

PR #2478, opened 2016-09-26
Refactor the ps test suite so that each test is independent of the pr…

### LINKED ISSUE #2370: Docker ps tests need to not rely on previous steps
Go through each step and re-write to avoid dependencies on previous tests that might have failed.

mhagen-vmware:
Fixes #2370 

This is the best way I could think without using ps in it's own test and without greatly increasing test time, I think it still maintains the goal of the tests.


anchal-agrawal:
LGTM

cgtexmex:
lgtm

mhagen-vmware:
.

PR #2479, opened 2016-09-26
Implement KeyValueStore pkg

### LINKED ISSUE #2080: Implement KV store in PL for generic personality specific data
This KV store will implement

- CRUD semantics
- There will be no cache for it in the PL unless we find usage pattern to require it.  
-  Also this data should be cleaned up when vic-machine removes the VCH.
-  Should follow the same URI based image store mechanism to namespace the k/v data (in case we fan in personalities each with their own image store).

fdawg4l:
Fixes #2080 

This pkg can be used to store values based on keys in a vsphere
datastore, local filesystem, or in memory since the backing storage is
just an interface.


pkg/kvstore/kvstore.go
```diff
@@ -0,0 +1,186 @@
...
+
+	rc, err := p.b.Download(op, p.fileName)
+	if err != nil {
+		// We need to check for 404 vs something else here.
+		op.Errorf("KeyValueStore(%s) ignoring error: %s", p.fileName, err)
+		return nil
+	}
+	defer rc.Close()
+
+	buf, err := ioutil.ReadAll(rc)
```
caglar10ur: maybe err = json.NewDecoder(rc).Decode(&p.kv) instead of this (and the rest), streaming decoding is the bonus :)

fdawg4l: Good call!!!  WIll do. Thanks!

pkg/kvstore/kvstore.go
```diff
@@ -0,0 +1,180 @@
...
+
+	// get the old value in case we need to roll back
+	oldvalue, ok := p.kv[key]
+
+	if ok && bytes.Compare(oldvalue, value) == 0 {
+		// NOOP
+		return nil
+	}
+
+	p.kv[key] = value
```
vburenin: Doc-string says "it will sync the data to the backend before updating the in-memory state"

but what I see in the code that value is updated in memory fist and only then saved to back end.
Am I missing something? 

You could just store a value into map if "save" was successful, so no need to restore an old value.

pkg/kvstore/kvstore.go
```diff
@@ -0,0 +1,180 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// Add a key to the KeyValueStore with the given value.  The Add operation is
+// write-through consistent, which means it will sync the data to the backend
+// before updating the in-memory state and returning to the caller.
+func (p *KeyValueStore) Add(op trace.Operation, key string, value []byte) error {
```
hmahmood: Is this really a "Set"?

pkg/kvstore/kvstore.go
```diff
@@ -0,0 +1,180 @@
...
+	defer p.l.Unlock()
+
+	oldvalue, ok := p.kv[key]
+	if !ok {
+		return ErrKeyNotFound
+	}
+
+	delete(p.kv, key)
+
+	if err := p.save(op); err != nil {
```
vburenin: I think if you delete a value from a map after you saved changes into backend it should make this code a little simpler.

pkg/kvstore/kvstore.go
hmahmood: We could also check if values implement the [BinaryMarshaler](https://golang.org/pkg/encoding/#BinaryMarshaler) interface, and marshal the value ourselves.

pkg/kvstore/kvstore.go
```diff
@@ -0,0 +1,180 @@
...
+// caller to provision the namespace.
+
+type KeyValueStore struct {
+	b Backend
+
+	kv map[string][]byte
+
+	fileName string
+
+	l sync.Mutex
```
hmahmood: Would a r/w lock be more appropriate?

pkg/kvstore/kvstore.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func (p *KeyValueStore) save(op trace.Operation) error {
+	buf, err := json.Marshal(p.kv)
+	if err != nil {
+		return err
+	}
+
+	// upload to an ephemeral file
+	tmpfile := p.fileName + ".tmp"
```
hmahmood: Use [ioutil.TempFile](https://golang.org/pkg/io/ioutil/#TempFile)?

pkg/kvstore/kvstore.go
```diff
@@ -0,0 +1,180 @@
...
+	fileName string
+
+	l sync.Mutex
+}
+
+type Backend interface {
+	// Creates path and ovewrites whatever is there
+	Upload(ctx context.Context, r io.Reader, pth string) error
+
+	Download(ctx context.Context, pth string) (io.ReadCloser, error)
```
hmahmood: Why doesn't `Download` take an `io.Writer`?

pkg/kvstore/kvstore.go
```diff
@@ -0,0 +1,180 @@
...
+
+	Mv(ctx context.Context, fromPath, toPath string) error
```
hmahmood: `io.Reader` -> `io.Writer` instead of `fromPath` to `toPath`?

pkg/kvstore/kvstore.go
fdawg4l: That assumes access to vfs. This is over http.

pkg/kvstore/kvstore.go
fdawg4l: Why? The map is private. What are you locking for read and why?

pkg/kvstore/kvstore.go
fdawg4l: The data is marshaled and written to disk. Without modifying the map, how can you write the updated field to disk?

fdawg4l: Unless you're implying writing a copy of the map.  In which case we have to remember we're memory constrained on the vch.

pkg/kvstore/kvstore.go
fdawg4l: See comment about save.

pkg/kvstore/kvstore.go
fdawg4l: The json marshaller already does this for json unless you're implying we should move to binary encoding of our data. That's fine in my opinion, but I think there are cases where json is preferred. Specifically when the blob's structure implementation changes over time.

fdawg4l: Re set/put. It's both. I'm fine changing the name of it clarifies any ambiguity.

pkg/kvstore/kvstore.go
vburenin: There is no way without making a copy. The main confusion here is just a doc string.
Do you have an idea how frequently anybody is going to write into this key/value storage?
It feels like "save" operation can be slow.

pkg/kvstore/kvstore.go
```diff
@@ -0,0 +1,180 @@
...
+	b Backend
+
+	kv map[string][]byte
+
+	fileName string
+
+	l sync.Mutex
+}
+
+type Backend interface {
```
hmahmood: So for the Backend wouldn't this be sufficient:

```go
NewReader(name string) io.ReadCloser
NewWriter(name string) io.WriteCloser
```

fdawg4l: Sure, if you had access to VFS.  We don't.  Remember, this goes to VC, then to the datastore, which may or may not be eventually consistent.  Where are you writing the data, how do you know if your last write was complete?

fdawg4l: Also, this `Backend` is implemented by the `datastore.Helper` that we have.

pkg/kvstore/kvstore.go
fdawg4l: The doc string,Ii agree, deviates from the implementation.  But what about the behavior?  I'm a caller, I'm calling `Add`, what's my contract?  I think in that sense the doc string works fairly well unless you have a different suggestion.

fdawg4l: Save can be very slow.  We've seen IO latencies measured in minutes to VC.

pkg/kvstore/kvstore.go
fdawg4l: Ah, Gets.  Sure.  Parallel gets are fine.  I can change this.

pkg/kvstore/kvstore.go
fdawg4l: I think I'll just nuke the consistency line.  No one cares from the point of view of the caller how things are organized.

pkg/kvstore/kvstore.go
hmahmood: I was implying that we pass in interface{} instead of []byte to Add.

pkg/kvstore/kvstore.go
hmahmood: That assumes we are doing http? I thought the backend was supposed to abstract that away?

fdawg4l: It does.  But ioutil.TempFIle does not use this abstraction.  It will create a file in /tmp of the vch.

pkg/kvstore/kvstore.go
fdawg4l: I'm not sure how the BinaryMarshaller implies interface{}, but that's fine.  No issue with changing the value type from []byte to interface{}.

hmahmood:
This meets our needs for now, but long term we need to come up with a better implementation.

lgtm

fdawg4l:
.

fdawg4l:
lgtm

PR #2481, opened 2016-09-27
Fix the path on bin-dir so that it works both locally and with drone

### LINKED ISSUE #2374: Imagec tests rely on bin-dir which is not set correctly when running locally
Some sort of switch to determine if you are not in drone.

mhagen-vmware:
Fixes #2374 

Needed for functional testing


anchal-agrawal:
:shipit:

rajanashok:
lgtm

mhagen-vmware:
.

PR #2486, opened 2016-09-27
Implement imagec private registry test[skip ci]

### LINKED ISSUE #250: Testing imagec against a private registry
VIC needs to be able to interact with a private Docker Registry:
https://docs.docker.com/registry/deploying/

This was originally included in the task with the rest of the integration tests but setting up a private registry requires a few steps, and this needs to be automated so that when a developer calls `make integration-tests` a Docker Registry is pulled down from Docker Hub, configured w/ a self-signed certificate & a user created to accept password auth, and `tests/imagec.bats` needs to be updated run some tests against it.

This is enough of a task that I've broken this out into its own story.

mhagen-vmware:
Fixes #250

This needs to be a manual test case as well, until we can get a private registry hooked up in the CI. Then we can move this over.


rajanashok:
LGTM.

anchal-agrawal:
:shipit:

mhagen-vmware:
.

PR #2489, opened 2016-09-27
Implements PORTS output in `docker ps`

### LINKED ISSUE #1701: docker ps does not show ports

gigawhitlocks:
Adds functionality to show ports mapped to containers when `docker ps` is called.

<pre>~/go/src/github.com/vmware/vic ps-ports* 19s
❯ docker -H 192.168.218.212:2376 --tls create -p 9000:80 -p 8443:443 nginx && docker -H 192.168.218.212:2376 --tls ps -a
Unable to find image 'nginx:latest' locally
Pulling from library/nginx
a3ed95caeb02: Pull complete 
6a5a5368e0c2: Pull complete 
4aceccff346f: Pull complete 
c8967f302193: Pull complete 
Digest: sha256:0d4dac4abf4f5217f878974a6cffea83b9102ed49b1f9f4c5216db0fd6b0af99
Status: Downloaded newer image for library/nginx:latest
b1358a9d60a4390c938327f086e7947d7c936a08836abf2c165a508afdb2fa67
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                                                         NAMES
b1358a9d60a4        nginx               "nginx -g daemon off;"   6 weeks ago         Created             192.168.218.212:9000->80/tcp, 192.168.218.212:8443->443/tcp   insane_galileo

~/go/src/github.com/vmware/vic ps-ports* 26s
❯ docker -H 192.168.218.212:2376 --tls create -p 8000:80 nginx && docker -H 192.168.218.212:2376 --tls create busybox && docker -H 192.168.218.212:2376 --tls ps -a
75701a6bf7a13c0876901b3a6cb734424431a288774579646fdb4a187a5b980e
Pulling from library/busybox
a3ed95caeb02: Pull complete 
8ddc19f16526: Pull complete 
1d0359ea0e392566b943bd78ad4d65179db96bc204fd5d5b4cdc48d6bf57aa26
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                                                         NAMES
1d0359ea0e39        busybox             "sh"                     6 weeks ago         Created                                                                           evil_leavitt
75701a6bf7a1        nginx               "nginx -g daemon off;"   6 weeks ago         Created             192.168.218.212:8000->80/tcp                                  sad_banach
b1358a9d60a4        nginx               "nginx -g daemon off;"   6 weeks ago         Created             192.168.218.212:9000->80/tcp, 192.168.218.212:8443->443/tcp   insane_galileo
</pre>

Fixes #1701

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -1019,7 +1019,7 @@ func (c *Container) Containers(config *types.ContainerListOptions) ([]*types.Con
 			Names:   names,
 			Command: cmd,
 			SizeRw:  *t.ContainerConfig.StorageSize,
-			Ports:   getPortInformation(&t),
+			Ports:   portInformation(&t),
 		}
 		containers = append(containers, c)
 	}
@@ -1544,7 +1544,7 @@ func ContainerSignal(containerID string, sig uint64) error {
 	return nil
 }
 
-func getClientIPv4Addrs() ([]netlink.Addr, error) {
+func clientIPv4Addrs() ([]netlink.Addr, error) {
 	l, err := netlink.LinkByName(clientIfaceName)
 	if err != nil {
 		return nil, fmt.Errorf("Could not look up link from client interface name %s due to error %s",
@@ -1560,11 +1560,11 @@ func getClientIPv4Addrs() ([]netlink.Addr, error) {
 
 // returns port bindings as a list of Docker Ports for return to the client
 // returns empty slice on error
-func getPortInformation(t *models.ContainerInfo) []types.Port {
+func portInformation(t *models.ContainerInfo) []types.Port {
 	// create a port for each IP on the interface (usually only 1, if netlink.FAMILY_ALL then usually 2)
 	var ports []types.Port
 
-	ips, err := getClientIPv4Addrs()
+	ips, err := clientIPv4Addrs()
 	if err != nil {
 		log.Errorf("Problem getting client IP address: %s", err.Error())
 		return ports
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -15,10 +15,12 @@
 package backends
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
 	"net"
 	"net/http"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -50,7 +52,6 @@ import (
 
 	"github.com/vishvananda/netlink"
 
-	"encoding/json"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	viccontainer "github.com/vmware/vic/lib/apiservers/engine/backends/container"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/portmap"
@@ -61,7 +62,6 @@ import (
 	"github.com/vmware/vic/lib/apiservers/portlayer/models"
 	"github.com/vmware/vic/lib/metadata"
 	"github.com/vmware/vic/pkg/trace"
-	"regexp"
 )
 
 const (
@@ -1549,12 +1549,11 @@ func ContainerSignal(containerID string, sig uint64) error {
 func getPortInformation(t *models.ContainerInfo, names []string) []types.Port {
 	var ports []types.Port
 	if len(names) == 0 {
-		log.Errorf("Couldn't find container while looking up port bindings; No container names provided")
+		log.Infof("Couldn't find container while looking up port bindings; No container names provided")
 		return ports
 	}
 
-	var container *viccontainer.VicContainer
-	container = cache.ContainerCache().GetContainer(names[0])
+	container := cache.ContainerCache().GetContainer(names[0])
 	if container == nil {
 		if match := regexp.MustCompile(`[\w_]+`).FindString(names[0]); match != "" {
 			container = cache.ContainerCache().GetContainer(match)
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -1570,7 +1570,7 @@ func getPortInformation(t *models.ContainerInfo) []types.Port {
 	}
 	container := cache.ContainerCache().GetContainer(*t.ContainerConfig.ContainerID)
 	if container == nil {
-		log.Errorf("Could not find container based on incorrectly formatted name")
+		log.Errorf("Could not find container with ID %s", *t.ContainerConfig.ContainerID)
 		return ports
 	}
 
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -50,6 +50,7 @@ import (
 
 	"github.com/vishvananda/netlink"
 
+	"encoding/json"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	viccontainer "github.com/vmware/vic/lib/apiservers/engine/backends/container"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/portmap"
@@ -1019,6 +1020,7 @@ func (c *Container) Containers(config *types.ContainerListOptions) ([]*types.Con
 			Names:   names,
 			Command: cmd,
 			SizeRw:  *t.ContainerConfig.StorageSize,
+			Ports:   getPortInformation(&t),
 		}
 		containers = append(containers, c)
 	}
@@ -1541,6 +1543,31 @@ func ContainerSignal(containerID string, sig uint64) error {
 	}
 
 	return nil
+}
+
+func getPortInformation(t *models.ContainerListInfo) []types.Port {
+	portBindings := cache.ContainerCache().GetContainer(t.Names[0]).HostConfig.PortBindings
+	var ports []types.Port
+	for _, jsonPort := range t.Ports {
+		port := new(types.Port)
+		err := json.Unmarshal([]byte(jsonPort), port)
+		if err != nil {
+			log.Errorf("Failed to unmarshal %+v due to error %s", jsonPort, err)
+			continue
+		}
+		privatePort, err := nat.NewPort(port.Type, strconv.Itoa(port.PrivatePort))
+		if err != nil {
+			log.Errorf("Got error %s while trying to create a new Port from type %s and number %d",
+				err.Error(), port.Type, port.PrivatePort)
+			continue
+		}
+		for i := 0; i < len(portBindings[privatePort]); i++ {
+			newport := port
+			newport.PublicPort, err = strconv.Atoi(portBindings[privatePort][i].HostPort)
+			ports = append(ports, *newport)
+		}
+	}
+	return ports
 }
 
 //----------------------------------
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -19,7 +19,6 @@ import (
 	"io"
 	"net"
 	"net/http"
-	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -77,8 +76,7 @@ var (
 
 	portMapper portmap.PortMapper
 
-	ctx                          = context.TODO()
-	containerFriendlyNamePattern = regexp.MustCompile(`[\w_]+`)
+	ctx = context.TODO()
 )
 
 func init() {
@@ -1021,7 +1019,7 @@ func (c *Container) Containers(config *types.ContainerListOptions) ([]*types.Con
 			Names:   names,
 			Command: cmd,
 			SizeRw:  *t.ContainerConfig.StorageSize,
-			Ports:   getPortInformation(&t, names),
+			Ports:   getPortInformation(&t),
 		}
 		containers = append(containers, c)
 	}
@@ -1547,7 +1545,6 @@ func ContainerSignal(containerID string, sig uint64) error {
 }
 
 func getClientIPAddrs() []netlink.Addr {
-
 	l, err := netlink.LinkByName(clientIfaceName)
 	if err != nil {
 		log.Errorf("Could not look up link from client interface name %s due to error %s",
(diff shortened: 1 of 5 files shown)


gigawhitlocks:
Two things to note: for some reason after I rebased in @caglar10ur and @cgtexmex 's changes, the `names` field of the `Container` passed around, which holds friendly names, now prepends a `/` onto the friendly names ( why? ) so I threw in a regex to find the friendly name in this field so that 
1) container lookup from the cache works and 2) if the prepended `/` is a bug, this code will not break should that bug be fixed.

2) Not sure why `CREATED` shows `6 weeks ago` ?? But I can reproduce on `master` so that doesn't seem to be related to this change:
<pre>~/go/src/github.com/vmware/vic master 20s
❯ docker -H 192.168.218.212:2376 --tls create busybox
Pulling from library/busybox
a3ed95caeb02: Pull complete 
8ddc19f16526: Pull complete 
56ddad7972584c55786eab1234c026b03119e528296f6bd6fd54320b2d5d26c7

~/go/src/github.com/vmware/vic master
❯ docker -H 192.168.218.212:2376 --tls ps -a
56ddad797258        busybox             "sh"                6 weeks ago         Created                                 cocky_bohr

</pre>

lib/apiservers/portlayer/swagger.json
```diff
@@ -2251,7 +2251,13 @@
 						"type": "string"
 					}
 				},
-				"state": {
+        "ports": {
```
anchal-agrawal: Indentation is off. I've had json indentation issues with spacemacs before.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -29,6 +29,8 @@ import (
 
 	"net/http"
 
+	"encoding/json"
+	"github.com/docker/engine-api/types"
```
hmahmood: Should not be referencing docker types in the port layer.

lib/apiservers/engine/backends/container.go
```diff
@@ -1543,6 +1546,48 @@ func ContainerSignal(containerID string, sig uint64) error {
 	return nil
 }
 
+func getPortInformation(t *models.ContainerInfo, names []string) []types.Port {
+	var ports []types.Port
+	if len(names) == 0 {
+		log.Errorf("Couldn't find container while looking up port bindings; No container names provided")
```
gigawhitlocks: roger

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
gigawhitlocks: hrmm.. okay.. it's a simple enough type I guess I can just redefine it..

lib/apiservers/portlayer/swagger.json
gigawhitlocks: yeah.. I hand-indented this and it still hates me

lib/apiservers/engine/backends/container.go
```diff
@@ -1543,6 +1546,48 @@ func ContainerSignal(containerID string, sig uint64) error {
...
+			return ports
+		}
+	}
+
+	portBindings := container.HostConfig.PortBindings
+
+	for _, jsonPort := range t.ContainerConfig.Ports {
+		port := new(types.Port)
+		err := json.Unmarshal([]byte(jsonPort), port)
+		if err != nil {
```
vburenin: Does it mean that if error happens you are really interest just to log it? No handling, etc?

lib/apiservers/engine/backends/container.go
```diff
@@ -50,6 +50,7 @@ import (
 
 	"github.com/vishvananda/netlink"
 
+	"encoding/json"
```
gigawhitlocks: we do; I lazily allowed emacs to add this import. Usually `goimports` complains at me if something got put in the wrong place but it missed this one.

lib/apiservers/engine/backends/container.go
gigawhitlocks: Hasan caught this. Should be `Infof`

lib/apiservers/engine/backends/container.go
```diff
@@ -1543,6 +1546,48 @@ func ContainerSignal(containerID string, sig uint64) error {
...
+	var ports []types.Port
+	if len(names) == 0 {
+		log.Errorf("Couldn't find container while looking up port bindings; No container names provided")
+		return ports
+	}
+
+	var container *viccontainer.VicContainer
+	container = cache.ContainerCache().GetContainer(names[0])
+	if container == nil {
+		if match := regexp.MustCompile(`[\w_]+`).FindString(names[0]); match != "" {
```
gigawhitlocks: Then it'd have to be global or package scope. Seems like a bad plan.
Also: how would it catch a RE expression at "first run" instead of "runtime"? Do you mean an error compiling the regex? Or an error performing a match? Because if you mean the latter, that needs to happen at runtime.

lib/apiservers/engine/backends/container.go
gigawhitlocks: Yes, there's a `continue` on the next line. We just report the error and move on. It's not fatal, to the rest of the operation. It may not even be fatal to the rest of the ports in the slice being iterated over.

lib/apiservers/engine/backends/container.go
```diff
@@ -1543,6 +1546,48 @@ func ContainerSignal(containerID string, sig uint64) error {
...
 }
 
+func getPortInformation(t *models.ContainerInfo, names []string) []types.Port {
+	var ports []types.Port
+	if len(names) == 0 {
+		log.Errorf("Couldn't find container while looking up port bindings; No container names provided")
+		return ports
+	}
+
+	var container *viccontainer.VicContainer
```
gigawhitlocks: I'm an idiot? 😝

lib/apiservers/engine/backends/container.go
vburenin: it is not a bad plan if it is unexported. it is a normal practice.
Catching error on first run would happen due to compilation attempt when you run binary.

lib/apiservers/engine/backends/container.go
vburenin: I am afraid I am not qualified to answer this question ;)

lib/apiservers/engine/backends/container.go
gigawhitlocks: The "error" you mean is if there's a regex compilation error? The value being matched against is not present anywhere until runtime.

lib/apiservers/engine/backends/container.go
vburenin: it is definitely not a fatal error, but it is an error that in my opinion would be suspicious. I am not really familiar with the details of the data sources in this case. If it is an input generated by machine - it is a bad error, if it is a human's error and not really an error, I would make it warning only.

gigawhitlocks: Refactored this out completely.

lib/apiservers/engine/backends/container.go
gigawhitlocks: Moved variable definition outside of the function per our discussion in person

lib/apiservers/engine/backends/container.go
gigawhitlocks: 😆

lib/apiservers/engine/backends/container.go
```diff
@@ -1543,6 +1546,72 @@ func ContainerSignal(containerID string, sig uint64) error {
...
+
+	// create a port for each IP on the interface (usually only 1, if netlink.FAMILY_ALL then usually 2)
+	ips := getClientIPAddrs()
+	var ports []types.Port
+	for _, ip := range ips {
+		ports = append(ports, types.Port{IP: ip.IP.String()})
+	}
+	container := cache.ContainerCache().GetContainer(names[0])
+	if container == nil {
+		if match := containerFriendlyNamePattern.FindString(names[0]); match != "" {
```
hmahmood: Not sure we need to do this (just err out in this case).

lib/apiservers/engine/backends/container.go
```diff
@@ -1543,6 +1546,72 @@ func ContainerSignal(containerID string, sig uint64) error {
...
+		return nil
+	}
+
+	// create a port for each IP on the interface (usually only 1, if netlink.FAMILY_ALL then usually 2)
+	ips := getClientIPAddrs()
+	var ports []types.Port
+	for _, ip := range ips {
+		ports = append(ports, types.Port{IP: ip.IP.String()})
+	}
+	container := cache.ContainerCache().GetContainer(names[0])
```
hmahmood: If we are ever going to use name[0], why have an array?

lib/apiservers/portlayer/swagger.json
gigawhitlocks: This change to the swagger def was removed entirely.

lib/apiservers/engine/backends/container.go
gigawhitlocks: It actually always runs this piece. During the rebase yesterday the friendly name got a `/` prepended to it. Seems like a bug but I'm not sure. I did this rather than stripping the `/` to be robust in case the preceding `/` is removed.

lib/apiservers/engine/backends/container.go
gigawhitlocks: I asked the same thing. Apparently the Docker spec requires one, although as best we can tell there's only ever one.

lib/apiservers/engine/backends/container.go
gigawhitlocks: The part I could remove is the `if container == nil` and just always use the regex.

lib/apiservers/engine/backends/container.go
gigawhitlocks: I guess I could've passed in `name[0]` as an argument instead of passing the array, but it's kind of six to one half-dozen the other

emlin: You can use ContainerID instead of name here. Then don't need to rely on name pattern as well.

gigawhitlocks: :+1: Thank you @emlin

gigawhitlocks:
Comments are addressed -- using `ContainerID` also allowed me to simplify my helper method to take only one argument, regex hack is gone, everything's lookin' good.

lib/apiservers/engine/backends/container.go
```diff
@@ -1543,6 +1544,64 @@ func ContainerSignal(containerID string, sig uint64) error {
...
+func getPortInformation(t *models.ContainerInfo) []types.Port {
+	// create a port for each IP on the interface (usually only 1, if netlink.FAMILY_ALL then usually 2)
+	ips := getClientIPAddrs()
+	var ports []types.Port
+	for _, ip := range ips {
+		ports = append(ports, types.Port{IP: ip.IP.String()})
+	}
+	container := cache.ContainerCache().GetContainer(*t.ContainerConfig.ContainerID)
+	if container == nil {
+		log.Errorf("Could not find container based on incorrectly formatted name")
```
emlin: might update the error message to log out container id?

emlin:
minor message comments. lgtm 👍

lib/apiservers/engine/backends/container.go
gigawhitlocks: yes that makes perfect sense; will do

lib/apiservers/engine/backends/container.go
```diff
@@ -1543,6 +1544,64 @@ func ContainerSignal(containerID string, sig uint64) error {
 	return nil
 }
 
+func getClientIPAddrs() []netlink.Addr {
+	l, err := netlink.LinkByName(clientIfaceName)
+	if err != nil {
+		log.Errorf("Could not look up link from client interface name %s due to error %s",
+			clientIfaceName, err.Error())
+		return nil
```
caglar10ur: why we are returning nil but reporting an error? If returning nil means err then I strongly suggest returning a pair ([]netlink.Addr, error) to make the intention of the function clear

lib/apiservers/engine/backends/container.go
```diff
@@ -1543,6 +1544,64 @@ func ContainerSignal(containerID string, sig uint64) error {
...
+	}
+	ips, err := netlink.AddrList(l, netlink.FAMILY_V4)
+	if err != nil {
+		log.Errorf("Could not get IP addresses of link due to error %s", err.Error())
+		return nil
```
caglar10ur: ^ ditto

lib/apiservers/engine/backends/container.go
```diff
@@ -1543,6 +1544,64 @@ func ContainerSignal(containerID string, sig uint64) error {
 	return nil
 }
 
+func getClientIPAddrs() []netlink.Addr {
+	l, err := netlink.LinkByName(clientIfaceName)
+	if err != nil {
+		log.Errorf("Could not look up link from client interface name %s due to error %s",
+			clientIfaceName, err.Error())
```
caglar10ur: also you can make this a single line, year is 2016 /s :P

lib/apiservers/engine/backends/container.go
```diff
@@ -1543,6 +1544,64 @@ func ContainerSignal(containerID string, sig uint64) error {
...
+		return nil
+	}
+	ips, err := netlink.AddrList(l, netlink.FAMILY_V4)
```
caglar10ur: only v4? If so let's rename the function to clientIPv4Addrs

lib/apiservers/engine/backends/container.go
```diff
@@ -1543,6 +1544,64 @@ func ContainerSignal(containerID string, sig uint64) error {
 	return nil
 }
 
+func getClientIPAddrs() []netlink.Addr {
```
caglar10ur: lose the get prefix, this is not java :)

andrewtchin:
lgtm

lib/apiservers/engine/backends/container.go
gigawhitlocks: Yeah I suppose so. Gotta say, not a huge fan of Go's multiple return value pattern, as it breaks function composition :(

lib/apiservers/engine/backends/container.go
gigawhitlocks: Renamed.

lib/apiservers/engine/backends/container.go
gigawhitlocks: Switched to return `ipaddr[], error`

caglar10ur:
lgtm

gigawhitlocks:
♻️

PR #2497, opened 2016-09-28
modify trace logging in network and exec for easier debugging

### LINKED ISSUE #2494: Lack of consistent container exec logging increases troubleshooting time
While investigating #2482 the ability to follow a container from handler to exec, across various calls is challenging.  The use of trace isn't consistent and the alignment of handle to a respective container can be mysterious.  Since a handle is temporary the logging of that shouldn't be necessary, but currently we log it in some places and there's little correlation to a specific container.

At a minimum the ContainerHandler and the exec/container.go should be updated to use trace consistently.   That should provide better visibility of the operations being performed on a specific container.

cgtexmex:
- added trace logging in network / exec where missing
- modified trace logging to include attributes
- if no attributes added then func name removed from trace

fixes #2494


anchal-agrawal:
LGTM

matthewavery:
LGTM

emlin:
lgtm 👍

cgtexmex:
rebase

PR #2500, opened 2016-09-28
Use common configuration definition between vic-machine and portlayer

### LINKED ISSUE #2177: Refactor VirtualContainerHostConfigSpec
To separate this big configuration structure to different functional structures. So any changes in one structure will not impact others.
This will help the data migration in vic-machine upgrade

emlin:
Clean up VirtualContainerHost configuration, and use the same configuration struct in portlayer modules.
Also a minor change for container VM configuration, which is to move session related attributes to SessionConfig.

Fixes #2177 
@hickeng @cgtexmex @hmahmood


lib/config/virtual_container_host.go
```diff
@@ -54,82 +54,119 @@ type VirtualContainerHostConfigSpec struct {
...
+	// vSphere connection configuration
+	Connection `vic:"0.1" scope:"read-only" key:"connect"`
 
-	////////////// basic contact information
-	// Administrative contact for the Virtual Container Host
-	Admin []mail.Address
-	// Administrative contact for hosting infrastructure
-	InfrastructureAdmin []mail.Address
+	// basic contact information
+	Contact `vic:"0.1" scope:"read-only" key:"contact"`
```
hmahmood: Contact_s_

lib/config/virtual_container_host.go
```diff
@@ -54,82 +54,119 @@ type VirtualContainerHostConfigSpec struct {
...
+	// certificate configuration, for both inbound and outbound access
+	CertificateConfig `vic:"0.1" scope:"read-only" key:"cert"`
 
-	////////////// Port Layer - storage
-	// Datastore URLs for image stores - the top layer is [0], the bottom layer is [len-1]
-	ImageStores []url.URL `vic:"0.1" scope:"read-only" key:"image_stores"`
-	// Permitted datastore URL roots for volumes
-	VolumeLocations map[string]*url.URL `vic:"0.1" scope:"read-only"`
+	// Port Layer - storage
+	StorageConfig `vic:"0.1" scope:"read-only" key:"storage"`
```
hmahmood: Just `Storage` maybe?

lib/config/virtual_container_host.go
```diff
@@ -54,82 +54,119 @@ type VirtualContainerHostConfigSpec struct {
...
-	////////////// Port Layer - network
-	// The network to use by default to provide access to the world
-	BridgeNetwork       string `vic:"0.1" scope:"read-only" key:"bridge_network"`
-	CreateBridgeNetwork bool   `vic:"0.1" scope:"read-only" key:"create_bridge_network"`
-	// Published networks available for containers to join, keyed by consumption name
-	ContainerNetworks map[string]*executor.ContainerNetwork `vic:"0.1" scope:"read-write" key:"container_networks"`
-	// The IP range for the bridge networks
-	BridgeIPRange *net.IPNet `vic:"0.1" scope:"read-only" key:"bridge-ip-range"`
+	// Port Layer - network
+	NetworkConfig `vic:"0.1" scope:"read-only" key:"network"`
```
hmahmood: "Config" suffix is a little redundant.

emlin:
@jak-atx @jooskim 
With this change, the appliance configuration change is for all other modules, except keys with this prefix: "guestinfo.vice..init."
I think you just use keys with prefix "guestinfo.vice..init", but if not, please let me know. That means UI is impacted by this change.

For the container VM configuration, only guestinfo.vice./stopSignal is change to inside of SessionConfig, which means the key will vary for different instance. But I assume you're not rely on that property.

hmahmood:
I should point out that these changes make a little harder to "cherry pick" the config elements that the individual layers need. For example, before the network layer could just redefine the  `ContainerNetwork` struct and the `Networks` field in its Config. Now it has to define `NetworkConfig` as well as `ContainerNetwork`.

jooskim:
@emlin there's only one config element we use that starts with "guestinfo.vice..init" as seen below. Does this PR will change the rest keys? If so how would those change them?

guestinfo.vice./common/name
guestinfo.vice./repo
guestinfo.vice./networks|bridge/ports~
guestinfo.vice./init/common/name
guestinfo.vice..init.networks|client.ip.IP

emlin:
> guestinfo.vice./common/name
> guestinfo.vice./repo
> guestinfo.vice./networks|bridge/ports~

Are they for vch appliance or container VM? I assume they are for container VM only.

> guestinfo.vice./init/common/name
> guestinfo.vice..init.networks|client.ip.IP

These two are not changed as they are in init module

jooskim:
The following three are used by VCH appliance:
guestinfo.vice./init/common/name
guestinfo.vice./repo
guestinfo.vice./networks|bridge/ports~

VCH's name is stored in the init module.

emlin:
Josh, There should be something wrong. There is no following two keys in VCH appliance configuration:
guestinfo.vice./repo
guestinfo.vice./networks|bridge/ports~

They are for container VM, as the port mapping is defined for container, which is not part of vch configuration.
And the repo is same.

jooskim:
Sorry. I completely mixed stuff up here. `guestinfo.vice./init/common/name` goes with `guestinfo.vice..init.networks|client.ip.IP` for VCH and the rest for Container VM. Things are actually in good shape. My apologies. I was just curious how the config keys would be impacted by this PR.

emlin:
That's great. Then no UI changes for it.

As backend is actually encoding the configuration structure to guestinfo key/values, so if I refactor the structure definition, the serialized key will be changed, although the value is still same.
As long as you didn't reference the key through string type, you're not infected, but unfortunately, UI does not have the utility support now.

emlin:
@jooskim although there should not have any impact on UI for this PR, I would highly recommend you to pick up my PR, and try UI functions, to avoid any potential misunderstanding here.
thanks

jooskim:
@emlin Thanks, I'll try it now.

jooskim:
@emlin Just tested UI with vic-machine based on your PR, and it worked just fine. In fact, the key-value pairs needed in UI were not changed.

emlin:
great, thanks

cgtexmex:
don't really care for the container, network, etc structs in the VCH config file -- makes it appear (in my mind) that they are limited to VCH configuration. 

lgtm

lib/portlayer/exec/config.go
```diff
@@ -32,25 +31,13 @@ type Configuration struct {
...
 	// Parent resource will be a VirtualApp on VC
 	VirtualApp *object.VirtualApp
 
-	// Path of the ISO to use for bootstrapping containers
-	BootstrapImagePath string `vic:"0.1" scope:"read-only" key:"bootstrap_image_path"`
-
-	// Allow custom naming convention for containerVMs
-	ContainerNameConvention string
-
 	// FIXME: temporary work around for injecting network path of debug nic
```
hickeng: May as well remove Networks and DebugNetwork as well. I've submitted a PR to your branch for that.

cgtexmex:
lgtm

lib/config/executor/network_interface.go
```diff
@@ -66,4 +67,6 @@ type ContainerNetwork struct {
 
 	// set of network wide links and aliases for this container on this network
 	Aliases []string `vic:"0.1" scope:"hidden" key:"aliases"`
+
+	PortGroup object.NetworkReference `vic:"0.1" scope:"read-only" recurse:"depth=0"`
```
emlin: wouldn't this recurse:"depth=0" work? I assume with this, we don't persist this attribute at all.

emlin: this is a runtime variable, but as it's referenced in the map object. It's not easy to address that issue without hook that in here.

lib/config/executor/container_vm.go
```diff
@@ -186,6 +183,9 @@ type SessionConfig struct {
 
 	Restart bool `vic:"0.1" scope:"read-only" key:"restart"`
 
+	// StopSignal is the signal name or number used to stop container session
+	StopSignal string `vic:"0.1" scope:"read-only" key:"stopSignal"`
```
emlin: got it. thanks. and looks like it's not caught by full CI as well.

lib/portlayer/exec/config.go
emlin: merged

lib/config/executor/container_vm.go
cgtexmex: oops..I should've caught that...apologies

lib/config/executor/network_interface.go
emlin: Didn't think this change will impact the iso file size. Then it's definitely not good.
In theory, the PortGroup is not part of configuration, it's just stay inside of configuration structure for internal cache only.
So I will change the network context structure to separate it.

lib/portlayer/network/context.go
```diff
@@ -85,18 +89,23 @@ func NewContext(bridgePool net.IPNet, bridgeMask net.IPMask, config *Configurati
...
 		scopes:            make(map[string]*Scope),
 		containers:        make(map[string]*Container),
+		PortGroups:        decodeNetworkReferences(&vctx, sess, config),
 	}
 
-	s, err := ctx.newBridgeScope(uid.New(), DefaultBridgeName, nil, net.IPv4(0, 0, 0, 0), nil, &IPAM{})
+	return ctx, nil
+}
+
+func (c *Context) Init() error {
```
hmahmood: Why the need for a separate Init()?

emlin: So it's possible to set any value of the context, before actually initialize any in memory cache, actually to initialize the scopes.
Otherwise, the NewContext method will resolve portgroups before create scopes, which means unit test will need govmomi connection. 
It's ok to change unit test to connect vsphere, but as the majority function is not related to vsphere, I assume this will make it easier.

lib/portlayer/network/network.go
```diff
@@ -94,8 +74,13 @@ func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSou
 		}
 
 		var netctx *Context
-		netctx, err = NewContext(*bridgeRange, *bridgeWidth, &config)
+		netctx, err = NewContext(ctx, sess, *bridgeRange, *bridgeWidth, &config)
 		if err != nil {
+			log.Warn(err)
```
emlin: well, previous it didn't return anything. I don't want to have those errors swallowed without even one message out.

hmahmood: Its _supposed_ to return err (line 120), but I think right now it always return `nil`. I have a fix for this in vch restart changes.

emlin: then I'll remove the log here.

lib/portlayer/network/context.go
```diff
@@ -59,6 +61,9 @@ type Context struct {
 	scopes       map[string]*Scope
 	containers   map[string]*Container
 	defaultScope *Scope
+
+	portgroups              map[string]object.NetworkReference
+	DecodeNetworkReferences func(vctx context.Context, sess *session.Session, config *Configuration) map[string]object.NetworkReference
```
hmahmood: This is really weird. I would have just replaced the global function in the unit tests.

hmahmood:
lgtm

mdubya66:
can we get this merged?

emlin:
♻️

PR #2506, opened 2016-09-28
Implement Docker Volume Inspect [full ci]

### LINKED ISSUE #2493: Docker API: Volume Inspect
Volume is the only API primitive that we support that currently does not have inspect support.  We need this for docker-compose.  It's used to determine existence.

sflxn:
Implement docker volume inspect for docker-compose usage.

Resolves #2493

### DIFF
--- a/lib/apiservers/engine/backends/errors.go
+++ b/lib/apiservers/engine/backends/errors.go
@@ -47,6 +47,11 @@ func VolumeCreateNotFoundError(msg string) error {
 	return derr.NewErrorWithStatusCode(fmt.Errorf("No volume store named (%s) exists", msg), http.StatusInternalServerError)
 }
 
+// VolumeNotFoundError returns a 404 docker error for a volume get request.
+func VolumeNotFoundError(msg string) error {
+	return derr.NewErrorWithStatusCode(fmt.Errorf("No such volume: %s", msg), http.StatusNotFound)
+}
+
 // NotFoundError returns a 404 docker error when a container is not found.
 func NotFoundError(msg string) error {
 	return derr.NewRequestNotFoundError(fmt.Errorf("No such container: %s", msg))
--- a/lib/apiservers/engine/backends/volume.go
+++ b/lib/apiservers/engine/backends/volume.go
@@ -95,7 +95,35 @@ func (v *Volume) Volumes(filter string) ([]*types.Volume, []string, error) {
 
 // VolumeInspect : docker personality implementation for VIC
 func (v *Volume) VolumeInspect(name string) (*types.Volume, error) {
-	return nil, fmt.Errorf("%s does not implement volume.VolumeInspect", ProductName())
+	defer trace.End(trace.Begin(name))
+
+	client := PortLayerClient()
+	if client == nil {
+		return nil, fmt.Errorf("failed to get a portlayer client")
+	}
+
+	if name == "" {
+		return nil, nil
+	}
+
+	param := storage.NewGetVolumeParamsWithContext(ctx).WithName(name)
+	res, err := client.Storage.GetVolume(param)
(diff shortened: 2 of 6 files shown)


sflxn:
@fdawg4l @matthewavery

lib/apiservers/engine/backends/errors.go
```diff
@@ -47,6 +47,11 @@ func VolumeCreateNotFoundError(msg string) error {
 	return derr.NewErrorWithStatusCode(fmt.Errorf("No volume store named (%s) exists", msg), http.StatusInternalServerError)
 }
 
+// VolumeNotFoundError returns a 404 docker error for a volume create request.
```
anchal-agrawal: `... for a volume inspect request.`

matthewavery: Maybe actually `... for a volume get request` since this could also be return by `VolumeJoin`

lib/apiservers/engine/backends/volume.go
```diff
@@ -95,7 +95,30 @@ func (v *Volume) Volumes(filter string) ([]*types.Volume, []string, error) {
...
+	}
+
+	if name == "" {
+		return nil, nil
+	}
+
+	param := storage.NewGetVolumeParamsWithContext(ctx).WithName(name)
+	res, err := client.Storage.GetVolume(param)
+	if err != nil {
+		return nil, VolumeNotFoundError(name)
```
hmahmood: We return VolumeNotFound for all errors from the port layer?

matthewavery:
Other than the small comment. LGTM 😄

lib/apiservers/engine/backends/volume.go
sflxn: Good point.

lib/apiservers/engine/backends/errors.go
sflxn: Updating.

lib/apiservers/engine/backends/volume.go
fdawg4l: The only error returned by the portlayer in this path is the not found error.

fdawg4l:
LGTM

fdawg4l:
And thanks for adding this.

lib/apiservers/engine/backends/volume.go
hmahmood: There is another case where the port layer returns 500.

hmahmood:
lgtm

lib/apiservers/engine/backends/errors.go
```diff
@@ -47,6 +47,11 @@ func VolumeCreateNotFoundError(msg string) error {
 	return derr.NewErrorWithStatusCode(fmt.Errorf("No volume store named (%s) exists", msg), http.StatusInternalServerError)
 }
 
+// VolumeNotFoundError returns a 404 docker error for a volume get request.
```
hickeng: Does this trigger volume creation in the client in the same way image 404 triggers pull?
IIRC that on-demand creation was implemented explicitly, but don't recall hearing that we checked for implicit behaviours.

lib/apiservers/portlayer/swagger.json
```diff
@@ -493,7 +493,7 @@
 		},
 		"/storage/volumes/{name}": {
 			"get": {
-				"description": "Get a Volume Handle",
+				"description": "Get info about a volume",
```
hickeng: Assuming that this change means this was never used anywhere until now?

sflxn:
ping

lib/apiservers/portlayer/swagger.json
sflxn: yup

PR #2480, opened 2016-09-26
Fix port-mapping integration test

anchal-agrawal:
The first integration test in [`1-25-Docker-Port-Map`](https://github.com/vmware/vic/blob/master/tests/test-cases/Group1-Docker-Commands/1-25-Docker-Port-Map.robot#L8) has a transient failure where nginx isn't running when we hit its endpoint with `curl`. This PR fixes the issue by using `wget` instead, which supports retries when the server isn't fully up.

More details in the commit message.

Fixes #2354

PR #2507, opened 2016-09-28
Improve port-mapping integration test

### LINKED ISSUE #2354: CI integration test failure: port mapping
A recent master build (https://ci.vmware.run/vmware/vic/4863) failed with:

```
Test-Cases.Group1-Docker-Commands.1-25-Docker-Port-Map :: Test 1-25 - Docke...
192.168.31.13

Installing VCH to test server...
Installer completed successfully: VCH-4863-1473...
Create container with port mappings                                   | FAIL |
7 != 0
```
From `log.html`:

```
Running command 'curl 192.168.31.77:10000 --connect-timeout 10 2>&1'.	
12:36:24.115	INFO	${rc} = 7
```

This is after a `start` on a container created with `create -it -p 10000:80 -p 10001:80 --name webserver nginx`

[integration_logs_4863_dbfaf282c1ef3d0cd395bb54ddee6d4e3bb458fd_1474049824.zip](https://github.com/vmware/vic/files/477515/integration_logs_4863_dbfaf282c1ef3d0cd395bb54ddee6d4e3bb458fd_1474049824.zip)

anchal-agrawal:
The first integration test in [1-25-Docker-Port-Map](https://github.com/vmware/vic/blob/master/tests/test-cases/Group1-Docker-Commands/1-25-Docker-Port-Map.robot#L8) has an intermittent failure where nginx isn't fully initialized when we hit its endpoint. #2480 addressed this by retrying the connection request if we get a `Connection refused` error, but we've seen this test fail in CI again.

This PR aims to resolve the issue by increasing the number of connection retries to give nginx more time to finish initializing. It also logs the output of `wget` during the test run.

Fixes #2354

### DIFF
--- a/tests/test-cases/Group1-Docker-Commands/1-25-Docker-Port-Map.robot
+++ b/tests/test-cases/Group1-Docker-Commands/1-25-Docker-Port-Map.robot
@@ -12,9 +12,11 @@ Create container with port mappings
     ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} start webserver
     Should Be Equal As Integers  ${rc}  0
     Should Not Contain  ${output}  Error
-    ${rc}  ${output}=  Run And Return Rc And Output  wget ${vch-ip}:10000 --retry-connrefused --timeout=10 --tries=10
+    ${rc}  ${output}=  Run And Return Rc And Output  wget ${vch-ip}:10000 --retry-connrefused --timeout=10
+    Log To Console  ${output}
     Should Be Equal As Integers  ${rc}  0
-    ${rc}  ${output}=  Run And Return Rc And Output  wget ${vch-ip}:10001 --retry-connrefused --timeout=10 --tries=10
+    ${rc}  ${output}=  Run And Return Rc And Output  wget ${vch-ip}:10001 --retry-connrefused --timeout=10
+    Log To Console  ${output}
     Should Be Equal As Integers  ${rc}  0
     ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} stop webserver
     Should Be Equal As Integers  ${rc}  0
--- a/tests/test-cases/Group1-Docker-Commands/1-25-Docker-Port-Map.robot
+++ b/tests/test-cases/Group1-Docker-Commands/1-25-Docker-Port-Map.robot
@@ -4,6 +4,12 @@ Resource  ../../resources/Util.robot
 Suite Setup  Install VIC Appliance To Test Server
 Suite Teardown  Cleanup VIC Appliance On Test Server
 
+*** Keywords ***
+Hit Nginx Endpoint
+    [Arguments]  ${vch-ip}  ${port}
+    ${rc}  ${output}=  Run And Return Rc And Output  wget ${vch-ip}:${port}
+    Should Be Equal As Integers  ${rc}  0
+
 *** Test Cases ***
 Create container with port mappings
     ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} create -it -p 10000:80 -p 10001:80 --name webserver nginx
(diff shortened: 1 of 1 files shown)


hmahmood:
lgtm

tests/test-cases/Group1-Docker-Commands/1-25-Docker-Port-Map.robot
```diff
@@ -12,9 +12,11 @@ Create container with port mappings
     ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} start webserver
     Should Be Equal As Integers  ${rc}  0
     Should Not Contain  ${output}  Error
-    ${rc}  ${output}=  Run And Return Rc And Output  wget ${vch-ip}:10000 --retry-connrefused --timeout=10 --tries=10
```
dougm: With --tries=10 removed, it'll go back to the default of 20, which still may not be enough time.  How about using the robot built-in to have a little time in between tries, we use something like this in several tests:

```
Wait Until Keyword Succeeds  20x  200 milliseconds ...
```

tests/test-cases/Group1-Docker-Commands/1-25-Docker-Port-Map.robot
```diff
@@ -12,9 +12,11 @@ Create container with port mappings
...
+    ${rc}  ${output}=  Run And Return Rc And Output  wget ${vch-ip}:10000 --retry-connrefused --timeout=10
+    Log To Console  ${output}
```
dougm: No need for this Log, the output.xml / log.html reports include the ${output}.

tests/test-cases/Group1-Docker-Commands/1-25-Docker-Port-Map.robot
mhagen-vmware: Agreed on the using wait until keyword succeeds, but we also do need to keep this reasonable. Just masking underlying problems with massive timeouts is not the way to go.. if timeout=10 and default 20 tries equals over 3 minutes (I am just guessing, don't know those wget parameters all that well), then something is not right.   I can docker run -d -p 80:80 nginx locally and instantly hit localhost in my browser. It seems like a lot of these "random" CI failures end up being internal component crashes/panics/races that we need to fix and not mask with looser tests.

dougm: I read the wget man page, the timeout and retries params do nothing in this case, so let's remove them both.  Not sure about retry-connrefused either, doesn't say how many times it retries - just get rid of that too.  The test does need to wait a bit as there will be time in between start returning and nginx listening on the port, we can use Robot Wait for that.

tests/test-cases/Group1-Docker-Commands/1-25-Docker-Port-Map.robot
anchal-agrawal: `log.html` truncates `${output}` in this case. I added `Log To Console` to display the complete output and see if nginx isn't up even after retrying several times - which will probably be the only reason this test fails. I can, however, take it out altogether - the logging statement is purely for diagnostic purposes.

dougm: Is it truncated in output.xml too?  If you remove the wget options and use robot's Wait, there's less output for a given 'Run', so it shouldn't get truncated.

tests/test-cases/Group1-Docker-Commands/1-25-Docker-Port-Map.robot
anchal-agrawal: Good call, @dougm. I'll refactor this test. `wget` doesn't retry if it gets a `connection refused` error unless `--retry-connrefused` is specified. I'll use robot's wait and retry mechanism instead.

@mhagen-vmware, nginx takes just a few seconds to come up when I run this test locally. `start` maps the ports and commits the container, and there could be latency/network issues in some cases.

tests/test-cases/Group1-Docker-Commands/1-25-Docker-Port-Map.robot
anchal-agrawal: Yes, it's truncated in `output.xml`. I'm refactoring the test to use robot's wait and retry.

anchal-agrawal:
I've updated the test to use robot's retry logic with 20 tries and a 5 second gap.

Looks like the root cause for this failure is a container IP conflict. I've posted an update to this PR's corresponding issue (#2354).

@mhagen-vmware has created #2524 to resolve bridge network issues in CI.

mhagen-vmware:
lgtm

anchal-agrawal:
🏁

PR #2510, opened 2016-09-29
[full ci] Set ChangeVersion

caglar10ur:
This property is useful because it guards against updates that have happened
between when the VM’s config is read and when it is applied.

Will return "Cannot complete operation due to concurrent modification by another
operation.." on failure.

Right now we are calling Refresh() in every stage, following PRs will
remove that and introduce retry logic.

### DIFF
--- a/lib/portlayer/exec/container.go
+++ b/lib/portlayer/exec/container.go
@@ -215,6 +215,11 @@ func (c *Container) Commit(ctx context.Context, sess *session.Session, h *Handle
 
 		// clear the spec as we've acted on it
 		h.Spec = nil
+
+		// refresh the struct with what propery collector provides
+		if err = c.Refresh(); err != nil {
+			return err
+		}
 	}
 
 	// if we're stopping the VM, do so before the reconfigure to preserve the extraconfig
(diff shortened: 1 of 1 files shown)


dougm:
lgtm

lib/portlayer/exec/container.go
```diff
@@ -258,6 +272,11 @@ func (c *Container) Commit(ctx context.Context, sess *session.Session, h *Handle
 
 		c.State = *h.State
 		commitEvent = events.ContainerStarted
+
+		// refresh the struct with what propery collector provides
+		if err := c.Refresh(); err != nil {
```
emlin: The refresh happens after reconfigure. Does that mean if there is inconsistency between portlayer and VC, the operation will always fail?
Currently, the VM power on will change the changeVersion, so out of band operation might cause failure with it.

caglar10ur: True out of band operations will cause failure. That's why we will introduce retry logic to the personality.

emlin:
lgtm

cgtexmex:
What's the benefit of having the new container attributes of Config & Runtime and the new Refresh method vs. the modifying the existing update method and utilizing the existing vm type?

caglar10ur:
@cgtexmex to be honest I didn't want to touch Update as there is no comment what it supposed to be doing and why. Looks like it got introduces via 4e5b80197a0dda731c75274563a164d17fac7af8
 and since than that's the only caller

cgtexmex:
lgtm

PR #2511, opened 2016-09-29
Conflict error type for portlayer/docker persona

caglar10ur:
So that we can detect conflict errors. Also changes stop and start logic
to check Runtime.PowerState before taking an action.

Depends on #2510


dougm:
lgtm

jzt:
LGTM

PR #2523, opened 2016-09-29
Fix 'Equivalent path already exists: /containers/{handle}'

caglar10ur:
```
       "/containers/{id}": {
and
        "/containers/{handle}": {
```

are the same endpoint and validators not happy about it.

Reindent the file while at it...

### DIFF
--- a/lib/apiservers/portlayer/swagger.json
+++ b/lib/apiservers/portlayer/swagger.json
@@ -1207,7 +1207,7 @@
 				}
 			}
 		},
-		"/containers/{handle}": {
+		"/containers/handle/{handle}": {
 			"put": {
 				"description": "Commit and close a container handle",
 				"operationId": "Commit",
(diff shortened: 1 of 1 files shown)


hickeng:
LGTM - worth noting that I've not really reviewed this as the indentation makes it excruciating and I don't have diff -w to hand. Rather, if the integration passed for this, then it's not regressed.

caglar10ur:
@hickeng there 2 commits here to ease the pain. The first one has the real change (https://github.com/vmware/vic/pull/2523/commits/ac5f421ab52aaf256beb4a020fc4044a83e64ae5) the other is purely indentation change :)

hickeng:
@caglar10ur I took a long time over it, being exceptionally diligent. My LGTM stands ;)

anchal-agrawal:
:shipit:

caglar10ur:
@hmahmood if I remember correctly you objected that. What do you want? Should I drop this?

hmahmood:
I think drop it for now, and open an issue.

Issue #1510, opened 2016-07-14
Implement VolumeDestroy in the PL


fdawg4l:
The function is there, it just needs to be implemented.

fdawg4l:
```
$ docker -H 192.168.8.131:2376 --tls create -it -v aaa:/poop  busybox /bin/sh                                                                             
Unable to find image 'busybox:latest' locally
latest: Pulling from library/busybox
a3ed95caeb02: Pull complete 
8ddc19f16526: Pull complete 
Digest: sha256:65ce39ce3eb0997074a460adfb568d0b9f0f6a4392d97b6035630c9d7bf92402
Status: Downloaded newer image for library/busybox:latest
81d88ebe57ef6970b5fe698d66d33870260ac800284c5545750582977a560248

$ govc datastore.ls
volooms
virtual-container-host
81d88ebe57ef6970b5fe698d66d33870260ac800284c5545750582977a560248
VIC

$ govc datastore.ls volooms/VIC/volumes
aaa
aaad
aaada

$ govc datastore.rm volooms/VIC/volumes/aaa                                                                                                               

$ govc datastore.ls volooms/VIC/volumes/aaa
govc: File [datastore1] volooms/VIC/volumes/aaa was not found
```

That shouldn't have worked.  What this means is the vsphere/filemanager does not check if a directory contains a `vmdk` and if that `vmdk` is referenced by a container vm.

So we either need to create a graph and figure out the set of un-attached volumes before deleting, or find another vsphere API which does this for us.  I find it hard to believe there isn't a vsphere "safe" way of deleting vmdks which throws the appropriate error if the vmdk is referenced.

fdawg4l:
cc @hickeng

fdawg4l:
Maybe we should be using (this)[ https://github.com/vmware/govmomi/blob/master/object/virtual_disk_manager.go#L128] API.  I'll try it.

PR #2526, opened 2016-09-29
Implement volume rm

### LINKED ISSUE #916: docker volume rm

fdawg4l:
Implements volume rm as well as some code clean up.

Fixes #916 #1511 #1510

### DIFF
--- a/lib/apiservers/engine/backends/volume.go
+++ b/lib/apiservers/engine/backends/volume.go
@@ -165,13 +165,17 @@ func (v *Volume) VolumeCreate(name, driverName string, driverArgs, labels map[st
 		switch err := err.(type) {
 		case *storage.CreateVolumeConflict:
 			return result, derr.NewErrorWithStatusCode(fmt.Errorf("A volume named %s already exists. Choose a different volume name.", name), http.StatusInternalServerError)
+
 		case *storage.CreateVolumeNotFound:
 			return result, derr.NewErrorWithStatusCode(fmt.Errorf("No volume store named (%s) exists", volumeStore(driverArgs)), http.StatusInternalServerError)
+
 		case *storage.CreateVolumeInternalServerError:
 			// FIXME: right now this does not return an error model...
 			return result, derr.NewErrorWithStatusCode(fmt.Errorf("%s", err.Error()), http.StatusInternalServerError)
+
 		case *storage.CreateVolumeDefault:
 			return result, derr.NewErrorWithStatusCode(fmt.Errorf("%s", err.Payload.Message), http.StatusInternalServerError)
+
 		default:
 			return result, derr.NewErrorWithStatusCode(fmt.Errorf("%s", err), http.StatusInternalServerError)
 		}
@@ -198,7 +202,7 @@ func (v *Volume) VolumeRm(name string) error {
 			return derr.NewRequestNotFoundError(fmt.Errorf("Get %s: no such volume", name))
 
 		case *storage.RemoveVolumeConflict:
-			return derr.NewRequestConflictError(fmt.Errorf("Volume '%s' is in use", name))
+			return derr.NewRequestConflictError(fmt.Errorf(err.Payload.Message))
 
 		case *storage.RemoveVolumeInternalServerError:
 			return derr.NewErrorWithStatusCode(fmt.Errorf("Server error from portlayer: %s", err.Payload.Message), http.StatusInternalServerError)
--- a/lib/portlayer/storage/vsphere/image.go
+++ b/lib/portlayer/storage/vsphere/image.go
@@ -589,9 +589,16 @@ func inUse(ctx context.Context, ID string) error {
 
 	for _, cont := range conts {
 		layerID := cont.ExecConfig.LayerID
-		if layerID == ID {
+
+		// check if the id is a volume
+		var mounted bool
+		if cont.ExecConfig.Mounts != nil {
+			_, mounted = cont.ExecConfig.Mounts[ID]
+		}
+
+		if layerID == ID || mounted {
 			return &portlayer.ErrImageInUse{
-				Msg: fmt.Sprintf("image %s in use by %s", layerID, cont.ExecConfig.ID),
+				Msg: fmt.Sprintf("image %s in use by %s", ID, cont.ExecConfig.ID),
 			}
 		}
 	}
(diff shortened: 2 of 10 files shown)


lib/portlayer/storage/vsphere/image.go
```diff
@@ -578,20 +578,20 @@ func (v *ImageStore) verifyImage(ctx context.Context, storeName, ID string) erro
 
 // XXX TODO This should be tied to an interface so we don't have to import exec
 // here (or wherever the cache lives).
-func inUse(ctx context.Context, ID string) error {
-	// XXX why doesnt this ever return an error?  Strange.
-	// Gather all the running containers and the images they are refering to.
-	state := exec.StateRunning
-	conts := exec.Containers.Containers(&state)
+func imagesInUse(ctx context.Context, ID string) error {
```
hmahmood: Should this just return a `bool`?

lib/portlayer/storage/vsphere/volume.go
```diff
@@ -189,8 +203,26 @@ func (v *VolumeStore) VolumeCreate(ctx context.Context, ID string, store *url.UR
 	return vol, nil
 }
 
-func (v *VolumeStore) VolumeDestroy(ctx context.Context, ID string) error {
-	return fmt.Errorf("TBD.  Not supported yet")
+func (v *VolumeStore) VolumeDestroy(ctx context.Context, vol *storage.Volume) error {
+	if err := volumesInUse(ctx, vol.ID); err != nil {
```
hmahmood: So after this call and the call to `Rm` below, is it still guaranteed that the volume is not in use?

fdawg4l: You're implying compare and swap.  No, there is no guarantee that the image you're going to RM below is not in use after exec tells us what images there are because the exec has no lock to speak of.  But, you're the authority here.  Is there a lock we can hold?

However, popping up a level, vsphere will keep things consistent for us in any case.  If we iterate the images in use, find that the image we're going to RM is not used, then RM the image as a new container is spinning up,  vsphere will either return an error on the container creation path, or prevent the image from being removed because the container is running (and return an error to the caller of `rm`).

In my view, that's sufficient.  Thoughts @hickeng ?

hmahmood: Yes I was hoping the `Rm` would fail in this case. So then the question is do we need the call to `volumesInUse`?

lib/portlayer/storage/vsphere/image.go
fdawg4l: Good point.  Also the function name could use some TLC.

lib/portlayer/storage/vsphere/volume.go
fdawg4l: Yes, we do.  In the case of the VM being off but still referring to an image, vsphere will happily let you nuke the image.  If the VM is then powered on, vsphere will return an error.

Likewise if have a VM that is on (during the in use check) and you attempt to remove that image before RM is called, and then the VM is powered off, NOOP-  user gets an error.

If a VM is off and turned on pre RM, the in use check will fail and the user will get the well crafted error.

If the VM

hickeng: @fdawg4l 
vsphere will only prevent the deletion of the disk if the containerVM is powered on (iirc). You will get an error during creation if the volume disk doesn't exist, which is the desired behaviour.

The case we may not address here is:
1. call volumesInUse - no containers
2. containerVM created referencing volume
3. volume deleted
4. containerVM powered on - will fail due to missing disk

One approach is to attempt _pressimistic_ reference counting in the PL:
1. set a reference count on initialization having checked existing containers
2. each time volumejoin is called, increment
3. check for zero reference count on rm
4. allow force deletion to bypass inconsistency.

There are a couple of other approaches that could reduce, but not close, this hole. My inclination at this time is to provide a good error message for the failing power on and then leverage the `deletable` flag once we have disklib access (or FCD appears).

lib/portlayer/storage/vsphere/volume.go
```diff
@@ -189,8 +203,26 @@ func (v *VolumeStore) VolumeCreate(ctx context.Context, ID string, store *url.UR
...
+	}
+
+	volDir := v.volDirPath(vol.ID)
+
+	dstore, err := v.getDatastore(vol.Store)
+	if err != nil {
+		return err
+	}
+
+	log.Infof("VolumeStore: Deleting %s", volDir)
```
hickeng: we could perform the in use check immediately before deletion instead of on entry to the function. Tiny difference, but still moves time of check closer to time of action.

lib/portlayer/storage/vsphere/volume.go
hmahmood: Ok it makes sense to keep the call to `volumesInUse` for the first case. The other two cases are really the same.

lib/portlayer/storage/vsphere/image.go
vburenin: is context needed here?

fdawg4l: Not necessarily.  The call to the container cache is a blocking one (or should be) and should therefore take a `ctx`.  Likewise, it should return an error (but isn't, ergo why this function returns an error).   I'll remove the `ctx` and the `error`.  The container cache could use some eyes TBH.

lib/portlayer/storage/vsphere/volume.go
fdawg4l: Not a fan of doing that.  I don't know how the go-runtime does branch detection or inlining and this just makes the code messy.

lib/portlayer/storage/vsphere/image.go
fdawg4l: I take it back.  I'd rather leave the error where it is.  The `err` string references the container using the given vol ID and that's useful.

hmahmood:
lgtm

lib/apiservers/engine/backends/volume.go
```diff
@@ -198,7 +202,7 @@ func (v *Volume) VolumeRm(name string) error {
 			return derr.NewRequestNotFoundError(fmt.Errorf("Get %s: no such volume", name))
 
 		case *storage.RemoveVolumeConflict:
-			return derr.NewRequestConflictError(fmt.Errorf("Volume '%s' is in use", name))
+			return derr.NewRequestConflictError(fmt.Errorf(err.Payload.Message))
```
matthewavery: Why switch this to the message we get from the Portlayer? Since this error return will need to be specific to docker.

fdawg4l: Because the message returned by the PL includes the container ID of the container using the image.  Squashing the error and returning the generic error above lacks detail the PL error has.

matthewavery: sounds good to me as long as it does not conflict with normal docker behavior. The volume stuff is almost free form though.

matthewavery:
LGTM

PR #2534, opened 2016-09-30
Persist docker port mapping in guestinfo

### LINKED ISSUE #2452: Map port for existing containers after VCH restart
For containers used port mapping, need to setup ip table for those containers during container cache population.

emlin:
Fixes #2452 

Persist docker port mapping information to make it survive VCH restart.
The information is in guestinfo right now, to avoid performance issue caused by current kv store.

### DIFF
--- a/lib/apiservers/engine/backends/backends.go
+++ b/lib/apiservers/engine/backends/backends.go
@@ -23,7 +23,6 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/docker/registry"
-	"github.com/docker/go-connections/nat"
 	httptransport "github.com/go-swagger/go-swagger/httpkit/client"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/container"
@@ -218,7 +217,7 @@ func syncContainerCache() error {
 	for _, info := range containme.Payload {
 		container := ContainerInfoToVicContainer(*info)
 		cc.AddContainer(container)
-		if err = setPortMapping(backend, container); err != nil {
+		if err = setPortMapping(info, backend, container); err != nil {
 			errs = append(errs, err.Error())
 		}
 	}
@@ -228,22 +227,26 @@ func syncContainerCache() error {
 	return nil
 }
 
-func setPortMapping(backend *Container, container *container.VicContainer) error {
-	log.Debugf("Set port mapping for container %q", container.Name)
+func setPortMapping(info *models.ContainerInfo, backend *Container, container *container.VicContainer) error {
+	if info.ContainerConfig.State == nil {
+		log.Infof("container state is nil")
+		return nil
+	}
+	if *info.ContainerConfig.State != "Running" || len(container.HostConfig.PortBindings) == 0 {
+		log.Infof("Container info state: %s", *info.ContainerConfig.State)
+		log.Infof("container portbinding: %+v", container.HostConfig.PortBindings)
+		return nil
+	}
 
+	log.Debugf("Set port mapping for container %q, portmapping %+v", container.Name, container.HostConfig.PortBindings)
 	client := backend.containerProxy.Client()
-	endpointsOK, err := client.Scopes.GetContainerEndpoints(scopes.NewGetContainerEndpointsParamsWithContext(ctx).WithHandleOrID(container.ContainerID))
+	endpointsOK, err := client.Scopes.GetContainerEndpoints(
+		scopes.NewGetContainerEndpointsParamsWithContext(ctx).WithHandleOrID(container.ContainerID))
 	if err != nil {
 		return err
 	}
 	for _, e := range endpointsOK.Payload {
 		if len(e.Ports) > 0 {
-			log.Infof("Map ports %s", e.Ports)
-			_, container.HostConfig.PortBindings, err = nat.ParsePortSpecs(e.Ports)
-			if err != nil {
-				return errors.Errorf("Failed to parse port mapping %s: %s", e.Ports, err)
-			}
-			log.Infof("host config port binding: %+v", container.HostConfig.PortBindings)
 			if err = backend.mapPorts(portmap.Map, container.HostConfig, e); err != nil {
 				return err
 			}
--- a/lib/apiservers/engine/backends/backends.go
+++ b/lib/apiservers/engine/backends/backends.go
@@ -18,15 +18,20 @@ import (
 	"context"
 	"net"
 	"net/url"
+	"strings"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/docker/registry"
+	"github.com/docker/go-connections/nat"
 	httptransport "github.com/go-swagger/go-swagger/httpkit/client"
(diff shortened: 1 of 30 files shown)


tests/test-cases/Group10-VCH-Restart/Group10-VCH-Restart.robot
```diff
@@ -28,6 +28,18 @@ Created Network Persists And Containers Are Discovered With Correct IPs
...
     ${id1}  ${ip1}=  Launch Container  vch-restart-test1  bridge
     ${id2}  ${ip2}=  Launch Container  vch-restart-test2  bridge
+
+    ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} create -it -p 10000:80 -p 10001:80 --name webserver nginx
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${output}  Error
+    ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} start webserver
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${output}  Error
+    ${rc}  ${output}=  Run And Return Rc And Output  wget ${vch-ip}:10000 --retry-connrefused --timeout=10 --tries=10
```
anchal-agrawal: These can be refactored using a recent change: https://github.com/vmware/vic/blob/master/tests/test-cases/Group1-Docker-Commands/1-25-Docker-Port-Map.robot#L8

lib/portlayer/network/endpoint.go
```diff
@@ -31,14 +31,15 @@ type alias struct {
...
-	static    bool
-	ports     map[Port]interface{} // exposed ports
-	aliases   map[string][]alias
+	container   *Container
+	scope       *Scope
+	ip          net.IP
+	gateway     net.IP
+	subnet      net.IPNet
+	static      bool
+	ports       map[Port]interface{} // exposed ports
```
hmahmood: Is there a need for this with `portmapping` below?

emlin: This map is still used to check exposed ports, to see if port is already consumed by other mapping.
The new attribute below is just for persistence. I don't think portlayer should care too much about the conversion from nat.Portbinding to string.

lib/apiservers/engine/backends/container_proxy_test.go
```diff
@@ -47,3 +48,31 @@ func TestProcessVolumeParams(t *testing.T) {
 	invalidFields, _ := processVolumeParam(invalidVolume)
 	assert.Equal(t, volumeFields{}, invalidFields)
 }
+
+func TestPort(t *testing.T) {
+
+	portMap, bindingMap, err := nat.ParsePortSpecs([]string{"1236:1235/tcp", "1237:1235/tcp", "2345/udp", "80", "127.0.0.1::8080", "127.0.0.1:5279:8080"})
+	if err != nil {
+		t.Errorf("Failed to part ports: %s", err.Error())
```
anchal-agrawal: Should this be `parse ports`?

tests/resources/Util.robot
```diff
@@ -436,3 +436,8 @@ Put Host Into Maintenance Mode
 Remove Host From Maintenance Mode
     ${rc}  ${output}=  Run And Return Rc And Output  govc host.maintenance.exit -host.ip=%{TEST_URL}
     Should Contain  ${output}  exiting maintenance mode... OK
+
+Hit Nginx Endpoint
+    [Arguments]  ${vch-ip}  ${port}
+    ${rc}  ${output}=  Run And Return Rc And Output  wget ${vch-ip}:${port}
+Should Be Equal As Integers ${rc} 0
```
anchal-agrawal: Needs indentation.

lib/apiservers/engine/backends/backends.go
```diff
@@ -197,18 +202,52 @@ func InsecureRegistries() []string {
...
+		return errors.Errorf("Failed to set port mapping: %s", strings.Join(errs, "\n"))
+	}
+	return nil
+}
+
+func setPortMapping(backend *Container, container *container.VicContainer) error {
+	log.Debugf("Set port mapping for container %q", container.Name)
+
+	client := backend.containerProxy.Client()
+	endpointsOK, err := client.Scopes.GetContainerEndpoints(scopes.NewGetContainerEndpointsParamsWithContext(ctx).WithHandleOrID(container.ContainerID))
```
anchal-agrawal: Consider breaking this up into separate lines for readability.

lib/apiservers/engine/backends/container_proxy_test.go
```diff
@@ -47,3 +48,31 @@ func TestProcessVolumeParams(t *testing.T) {
 	invalidFields, _ := processVolumeParam(invalidVolume)
 	assert.Equal(t, volumeFields{}, invalidFields)
 }
+
+func TestPort(t *testing.T) {
+
+	portMap, bindingMap, err := nat.ParsePortSpecs([]string{"1236:1235/tcp", "1237:1235/tcp", "2345/udp", "80", "127.0.0.1::8080", "127.0.0.1:5279:8080"})
```
anchal-agrawal: Minor comment - separating this into multiple lines would improve readability.

lib/portlayer/network/endpoint.go
emlin: There is no 1X1 mapping for the exposed ports and portbinding string, for example, the string looks like "80:80/tcp", "8080:80/tcp", but the exposed ports will be "80/tcp" only.

portlayer will check "80/tcp" only, but persistence need to keep "80:80/tcp", "8080:80/tcp". 
Ideally portlayer should not care about whole portmapping, if we merge them two, we'll need to have portlayer convert the nat.Map back to string, and in the future, if we move the portmapping to kv store, we'll need to separate them again.

emlin: OK, get rid of this, cause it doesn't work for not started container.

lib/apiservers/engine/backends/backends.go
```diff
@@ -197,18 +201,56 @@ func InsecureRegistries() []string {
...
+	}
+	if *info.ContainerConfig.State != "Running" || len(container.HostConfig.PortBindings) == 0 {
+		log.Infof("Container info state: %s", *info.ContainerConfig.State)
+		log.Infof("container portbinding: %+v", container.HostConfig.PortBindings)
+		return nil
+	}
+
+	log.Debugf("Set port mapping for container %q, portmapping %+v", container.Name, container.HostConfig.PortBindings)
+	client := backend.containerProxy.Client()
+	endpointsOK, err := client.Scopes.GetContainerEndpoints(
```
hmahmood: Do we need this call, given that the full port mapping info is in HostConfig?

emlin: this call is to get back endpoint information, including the containerIP, which is not persisted in portmapping.
Need this information if container is running

hmahmood: Why do you need IP information at this stage? This should only be collected when `inspect` is called.

emlin: But the method backend.mapPorts require that information to map port.

hmahmood:
lgtm

anchal-agrawal:
:shipit:

emlin:
⏺

PR #2535, opened 2016-09-30
Fix 0->0 docker ps port binding bug

gigawhitlocks:
Yesterday @hmahmood found a bug where sometimes when there's no ports bound to a container, docker ps shows `0->0` as being bound. This makes zero sense, so I've added a check so that we don't report an impossible `0->0` binding & also added a unit test for the method .

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -1597,6 +1597,7 @@ func ContainerSignal(containerID string, sig uint64) error {
 	return nil
 }
 
+// pulls out the client IP(s) from program state
 func clientIPv4Addrs() ([]netlink.Addr, error) {
 	l, err := netlink.LinkByName(clientIfaceName)
 	if err != nil {
@@ -1611,10 +1612,11 @@ func clientIPv4Addrs() ([]netlink.Addr, error) {
 	return ips, nil
 }
 
-// returns port bindings as a list of Docker Ports for return to the client
+// returns port bindings as a slice of Docker Ports for return to the client
 // returns empty slice on error
 func portInformation(t *models.ContainerInfo, ips []netlink.Addr) []types.Port {
-	// create a port for each IP on the interface (usually only 1, if netlink.FAMILY_ALL then usually 2)
+	// create a port for each IP on the interface (usually only 1, but could be more)
+	// (works with both IPv4 and IPv6 addresses)
 	var ports []types.Port
 
 	container := cache.ContainerCache().GetContainer(*t.ContainerConfig.ContainerID)
@@ -1640,7 +1642,6 @@ func portInformation(t *models.ContainerInfo, ips []netlink.Addr) []types.Port {
 				continue
 			}
 			port.Type = portAndType[1]
-			log.Infof("%d", len(hostPortBindings))
 
 			for i := 0; i < len(hostPortBindings); i++ {
 				newport := port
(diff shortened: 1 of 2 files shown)


hmahmood:
lgtm

lib/apiservers/engine/backends/container.go
```diff
@@ -1024,6 +1024,14 @@ func (c *Container) Containers(config *types.ContainerListOptions) ([]*types.Con
 		// get the docker friendly status
 		_, status := dockerStatus(int(*t.ProcessConfig.ExitCode), *t.ProcessConfig.Status, *t.ContainerConfig.State, started, stopped)
 
+		ips, err := clientIPv4Addrs()
+		var ports []types.Port
+		if err == nil {
```
dougm: Looks like that should be `err != nil`.

lib/apiservers/engine/backends/container_test.go
```diff
@@ -580,3 +582,65 @@ func TestContainerLogs(t *testing.T) {
...
+	}
+	portBindings := []nat.PortBinding{portBinding}
+	portMap[port] = portBindings
+	mockHostConfig.PortBindings = portMap
+
+	mockContainerInfo.ContainerConfig = mockContainerConfig
+
+	ip, _ := netlink.ParseAddr("192.168.1.1/24")
+	ips := []netlink.Addr{*ip}
+	// ports := portInformation(mockContainerInfo, ips)
```
anchal-agrawal: 2dd

lib/apiservers/engine/backends/container_test.go
```diff
@@ -580,3 +582,65 @@ func TestContainerLogs(t *testing.T) {
...
+	// assert.Empty(t, ports, "There should be no bound IPs")
+
+	co := viccontainer.NewVicContainer()
+	co.HostConfig = mockHostConfig
+	co.ContainerID = containerID
+	co.Name = "bar"
+	cache.ContainerCache().AddContainer(co)
+
+	ports := portInformation(mockContainerInfo, ips)
+	// assert.Empty(t, ports, "There should still be no bound IPs")
```
anchal-agrawal: Here as well.

lib/apiservers/engine/backends/container.go
gigawhitlocks: :+1:

lib/apiservers/engine/backends/container_test.go
gigawhitlocks: :+1:

lib/apiservers/engine/backends/container_test.go
gigawhitlocks: :+1:

lib/apiservers/engine/backends/container_test.go
```diff
@@ -580,3 +582,61 @@ func TestContainerLogs(t *testing.T) {
 		}
 	}
 }
+
+func TestPortInformation(t *testing.T) {
+	// I could refactor this to have a loop and an array or whatever but eh
```
anchal-agrawal: 👀

anchal-agrawal:
lgtm

lib/apiservers/engine/backends/container_test.go
gigawhitlocks: LOL WHOOPS probably shouldn't leave that kind of comment around thanks for spotting it

gigawhitlocks:
♻️

PR #2536, opened 2016-09-30
Creates new block device package to remove our reliance on udev for volume mounting WIP

### LINKED ISSUE #2472: Server error from portlayer: unable to wait for process launch status: context deadline exceeded
Seen in:
https://ci.vmware.run/vmware/vic/5139

[integration_logs_5139_6ee9c6b83d2fe921fc6f0eaa72310b3a53414f1b.zip](https://github.com/vmware/vic/files/494051/integration_logs_5139_6ee9c6b83d2fe921fc6f0eaa72310b3a53414f1b.zip)

### LINKED ISSUE #2485: Volumes: volume disks not presenting in containers
I have seen, only once that I'm aware of, a container fail to start due to the volume it expects not being present in /dev/disk/by-label when it comes to time to mount it.

I do not have the log bundle for this failure unfortunately. My estimation is that switch-root occurred before systemd-udev had created the link. While `[systemd-udev]` appears in the process list for the containerVM after switchroot but is defunct, so isn't going to be processing netlink messages after the switch.

One possible option is to wait to switch-root until there's the same number of entries in /dev/disk/by-label as in /sys/block - **SPECULATIVE FIX - definitely needs confirmation of problem and validation of solution**

#blkid #udev

matthewavery:
This Commit uses a new vendored package to inspect attached block devices and create a map which is then used to look up(by label) all of the mounts that are expected for the container. We need this in order to strip out udev and more of systemd. It will also keeps us from having the race condition where udev has not finished sym linking block devices when we `switch-root` which then breaks udev causing us to lose any labeled devices that we intend to mount. 

Fixes #2485 
Fixes #2472

### DIFF
--- a/lib/tether/interfaces.go
+++ b/lib/tether/interfaces.go
@@ -18,7 +18,6 @@ import (
 	"io"
 
 	"github.com/vmware/vic/pkg/dio"
-	"github.com/vmware/vic/pkg/fs"
 	"golang.org/x/net/context"
 
 	"github.com/vmware/vic/pkg/dio"
@@ -36,7 +35,7 @@ type Operations interface {
 
 	SetHostname(hostname string, aliases ...string) error
 	Apply(endpoint *NetworkEndpoint) error
-	MountLabel(ctx context.Context, devicePath, target string) error
+	MountLabel(ctx context.Context, devicePath string, target string) error
 	Fork() error
 
 	SessionLog(session *SessionConfig) (dio.DynamicMultiWriter, error)
--- a/lib/tether/interfaces.go
+++ b/lib/tether/interfaces.go
@@ -17,6 +17,8 @@ package tether
 import (
 	"io"
 
+	"github.com/vmware/vic/pkg/dio"
+	"github.com/vmware/vic/pkg/fs"
 	"golang.org/x/net/context"
 
 	"github.com/vmware/vic/pkg/dio"
@@ -34,7 +36,7 @@ type Operations interface {
 
 	SetHostname(hostname string, aliases ...string) error
 	Apply(endpoint *NetworkEndpoint) error
-	MountLabel(ctx context.Context, label, target string) error
+	MountLabel(ctx context.Context, label, target string, BlockDeviceMap map[string]fs.Fsinfo) error
 	Fork() error
 
 	SessionLog(session *SessionConfig) (dio.DynamicMultiWriter, error)
--- a/lib/tether/interfaces.go
+++ b/lib/tether/interfaces.go
@@ -36,7 +36,7 @@ type Operations interface {
 
 	SetHostname(hostname string, aliases ...string) error
 	Apply(endpoint *NetworkEndpoint) error
-	MountLabel(ctx context.Context, label, target string, BlockDeviceMap map[string]fs.Fsinfo) error
+	MountLabel(ctx context.Context, devicePath, target string) error
 	Fork() error
 
 	SessionLog(session *SessionConfig) (dio.DynamicMultiWriter, error)
--- a/lib/tether/ops_linux.go
+++ b/lib/tether/ops_linux.go
@@ -584,25 +584,20 @@ func (t *BaseOperations) dhcpLoop(stop chan bool, e *NetworkEndpoint, ack *dhcp.
 
 // MountLabel performs a mount by looking up a fsinfo from the device map using the volume label
 // the value of the map is the fsinfo object which has the device path.
-func (t *BaseOperations) MountLabel(ctx context.Context, label, target string, BlockDeviceMap map[string]fs.Fsinfo) error {
-	defer trace.End(trace.Begin(fmt.Sprintf("Mounting %s on %s", label, target)))
+func (t *BaseOperations) MountLabel(ctx context.Context, devicepath, target string) error {
+	defer trace.End(trace.Begin(fmt.Sprintf("Mounting %s on %s", devicepath, target)))
 
 	if err := os.MkdirAll(target, 0600); err != nil {
 		return fmt.Errorf("unable to create mount point %s: %s", target, err)
 	}
 
-	blockInfo, ok := BlockDeviceMap[label]
-	if !ok {
-		return fmt.Errorf(fmt.Sprintf("Could not find block device (%s) when attempting to perform mount operation", label))
-	}
-
-	sourcePath, err := filepath.EvalSymlinks(blockInfo.DevicePath)
+	sourcePath, err := filepath.EvalSymlinks(blockInfo.DevPath())
 	if err != nil {
-		return fmt.Errorf("error attempting to follow the symlink in (%s)", blockInfo.DevicePath)
+		return fmt.Errorf("error attempting to follow the symlink in (%s)", devicepath)
 	}
 
 	if err := Sys.Syscall.Mount(sourcePath, target, "ext4", syscall.MS_NOATIME, ""); err != nil {
-		detail := fmt.Sprintf("mounting %s on %s failed: %s", label, target, err)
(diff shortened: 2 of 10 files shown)


lib/tether/interfaces.go
```diff
@@ -33,7 +34,7 @@ type Operations interface {
 
 	SetHostname(hostname string, aliases ...string) error
 	Apply(endpoint *NetworkEndpoint) error
-	MountLabel(ctx context.Context, label, target string) error
+	MountLabel(ctx context.Context, label, target string, BlockDeviceMap map[string]fs.Fsinfo) error
```
fdawg4l: `BlockDeviceMap` as a map?  That's redundant or named incorrectly.  How about creating a struct `BlockDevices`, and hiding the map implementation inside it?

lib/tether/ops_linux.go
```diff
@@ -580,35 +581,26 @@ func (t *BaseOperations) dhcpLoop(stop chan bool, e *NetworkEndpoint, ack *dhcp.
 	}
 }
 
-// MountLabel performs a mount with the label and target being absolute paths
-func (t *BaseOperations) MountLabel(ctx context.Context, label, target string) error {
+// MountLabel performs a mount by looking up a fsinfo from the device map using the volume label
+// the value of the map is the fsinfo object which has the device path.
+func (t *BaseOperations) MountLabel(ctx context.Context, label, target string, BlockDeviceMap map[string]fs.Fsinfo) error {
```
fdawg4l: Same comment.  Hide the map in the struct.

lib/tether/tether.go
```diff
@@ -212,6 +217,15 @@ func (t *tether) Start() error {
 		}
 		extraconfig.Encode(t.sink, t.config)
 
+		log.Info("Populating ext4 block device list now.")
+		deviceManager := fs.Ext4DeviceManager{}
```
fdawg4l: Very `ext` specific.  How do you intend to extend this to support NFS or NTFS?  We should hide these implementations in `BlockDevices`.

pkg/fs/block.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+//this contains the fs block info that we are interested in.
+type Fsinfo struct {
+	DevicePath  string
+	UUID        string
+	VolumeLabel string
+}
+
+type fsDeviceManager interface {
```
fdawg4l: Non-public interface that doesn't seem to be referenced.

lib/tether/ops_linux.go
```diff
@@ -580,35 +581,26 @@ func (t *BaseOperations) dhcpLoop(stop chan bool, e *NetworkEndpoint, ack *dhcp.
...
-	if timeout {
-		detail := fmt.Sprintf("timed out waiting for %s to appear", label)
-		return errors.New(detail)
+	sourcePath, err := filepath.EvalSymlinks(blockInfo.DevicePath)
+	if err != nil {
+		return fmt.Errorf("error attempting to follow the symlink in (%s)", blockInfo.DevicePath)
 	}
 
-	if err := Sys.Syscall.Mount(label, target, "ext4", syscall.MS_NOATIME, ""); err != nil {
+	if err := Sys.Syscall.Mount(sourcePath, target, "ext4", syscall.MS_NOATIME, ""); err != nil {
```
vburenin: is ext4 the only FS we support?

pkg/fs/block.go
```diff
@@ -0,0 +1,92 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fs
+
+import (
+	"fmt"
+	"github.com/ctdk/sbinfo"
```
vburenin: Keep system, vic and third-party imports separated.

lib/tether/ops_linux.go
matthewavery: For now it is, I think the next one we aim to support is NFS

lib/tether/tether.go
```diff
@@ -212,6 +217,15 @@ func (t *tether) Start() error {
 		}
 		extraconfig.Encode(t.sink, t.config)
 
+		log.Info("Populating block device list now.")
+		BlockDevices, err := fs.NewBlockDevices(linuxBlockDevicePath)
+		DeviceMap := BlockDevices.DevicesByLabel()
```
caglar10ur: You are shadowing DeviceMap here so either make this an assignment or remove the declaration

lib/tether/interfaces.go
```diff
@@ -33,7 +34,7 @@ type Operations interface {
 
 	SetHostname(hostname string, aliases ...string) error
 	Apply(endpoint *NetworkEndpoint) error
-	MountLabel(ctx context.Context, label, target string) error
+	MountLabel(ctx context.Context, devicePath, target string) error
```
caglar10ur: Please add the type (devicePath string) there to ease reading instead of saving some keypresses.

lib/tether/ops_linux.go
```diff
@@ -580,36 +581,22 @@ func (t *BaseOperations) dhcpLoop(stop chan bool, e *NetworkEndpoint, ack *dhcp.
 	}
 }
 
-// MountLabel performs a mount with the label and target being absolute paths
-func (t *BaseOperations) MountLabel(ctx context.Context, label, target string) error {
-	defer trace.End(trace.Begin(fmt.Sprintf("Mounting %s on %s", label, target)))
+// MountLabel performs a mount by looking up a fsinfo from the device map using the volume label
+// the value of the map is the fsinfo object which has the device path.
+func (t *BaseOperations) MountLabel(ctx context.Context, devicepath, target string) error {
```
caglar10ur: same as above, let's add the type

pkg/fs/block.go
```diff
@@ -0,0 +1,144 @@
...
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/ctdk/sbinfo"
+)
+
+//this contains the fs block info that we are interested in.
+type Fsinfo struct {
```
caglar10ur: FSInfo?

matthewavery: Changed this to `DeviceInfo` which seemed much more appropriate.

lib/tether/interfaces.go
```diff
@@ -34,7 +33,7 @@ type Operations interface {
 
 	SetHostname(hostname string, aliases ...string) error
 	Apply(endpoint *NetworkEndpoint) error
-	MountLabel(ctx context.Context, devicePath, target string) error
+	MountLabel(ctx context.Context, devicePath string, target string) error
```
hickeng: why changing the signature - it's a noop.

hickeng:
If you want a work around until this is in you could add the following just before we exec tether in `isos/bootstrap/bootstrap`:
```
until [[ $(ls -1 /dev/disk/by-label | wc -l) -eq $(ls -1 /sys/block | wc -l) ]]; do :;done
```
or if we want a bit of yielding:
```
```
until [[ $(ls -1 /dev/disk/by-label | wc -l) -eq $(ls -1 /sys/block | wc -l) ]]; do sleep 0.1;done
```

lib/tether/interfaces.go
matthewavery: This was requested by @caglar10ur

caglar10ur: Yep, easier to read this way.

matthewavery:
@hickeng @faijaz @vburenin I am closing this PR since we are tabling this for the time being. The branch is backed up in my fork with the name `tether-mount-bug`

PR #2548, opened 2016-10-03
[full ci] VCH static IPs for client, external, and management networks

### LINKED ISSUE #1184: vic-machine: static ip addresses for appliance
We should allow users to specify static IP addresses for the appliance networks.

@hmahmood has code to validate that networks do not overlap/conflict/et al, but if that's not easy to consume then we can skip the validation for TP3 timeframe and open a second issue to add that.

andrewtchin:
Allows the user to specify static IP address and gateway for client, external, and management (CEM) networks.
Allows the user to specify DNS servers to use for CEM networks.
If the same port group is specified for more than one of the CEM networks, only one static IP may be specified for the networks sharing the port group. 

Fixes #1184 

passed local regression tests

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -79,6 +79,16 @@ type Create struct {
 	containerNetworksDNS      cli.StringSlice
 	volumeStores              cli.StringSlice
 	insecureRegistries        cli.StringSlice
+	dns                       cli.StringSlice
+	clientNetworkName         string
+	clientNetworkGateway      string
+	clientNetworkIP           string
+	externalNetworkName       string
+	externalNetworkGateway    string
+	externalNetworkIP         string
+	managementNetworkName     string
+	managementNetworkGateway  string
+	managementNetworkIP       string
 
 	memoryReservLimits string
 	cpuReservLimits    string
@@ -107,7 +117,7 @@ func (c *Create) Flags() []cli.Flag {
 		cli.StringFlag{
 			Name:        "container-store, cs",
 			Value:       "",
-			Usage:       "Container datastore path - not supported yet, default to image datastore",
+			Usage:       "Container datastore path - not supported yet, defaults to image datastore",
 			Destination: &c.ContainerDatastoreName,
 		},
 		cli.StringSliceFlag{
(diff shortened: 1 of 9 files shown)


lib/install/validate/network.go
```diff
@@ -74,40 +82,157 @@ func (v *Validator) checkNetworkConflict(bridgeNetName, otherNetName, otherNetTy
...
+		}
+	}
+}
+
+// checkPortGroups checks that network config is valid
+// prevent assigning > 1 static IP to the same port group
+func (v *Validator) checkPortGroups(counts map[string]int, ips map[string][]data.NetworkConfig) error {
+	defer trace.End(trace.Begin(""))
+
+	for pg := range ips {
```
hickeng: As a forward looking note, we should be able to handle multiple IPs on one interface. This could be leveraged by customers to get separation between client & other networks without having separate port groups.

This is related again to source routing support.

lib/install/validate/network.go
```diff
@@ -74,40 +82,157 @@ func (v *Validator) checkNetworkConflict(bridgeNetName, otherNetName, otherNetTy
...
+	if input.ManagementNetwork.Name == "" {
+		input.ManagementNetwork.Name = input.ClientNetwork.Name
+	}
+
+	c := make(map[string]int)                  // number of VCH networks using portgroup
+	i := make(map[string][]data.NetworkConfig) // user configured IPs for portgroup
+	v.portGroupConfig(input, c, i)
+
+	err = v.checkPortGroups(c, i)
+	if err != nil {
```
hickeng: no need for this check, just pass the err to NoteIssue directly. Same for other NoteIssue locations

lib/install/validate/network.go
```diff
@@ -74,40 +82,157 @@ func (v *Validator) checkNetworkConflict(bridgeNetName, otherNetName, otherNetTy
...
+
+	err = v.configureSharedPortGroups(input, c, i)
+	if err != nil {
+		v.NoteIssue(err)
+	}
+
 	// External net
 	// external network is default for appliance
-	err := v.addNetworkHelper(ctx, conf, input.ExternalNetworkName, "external", "external", true)
+	if !input.ExternalNetwork.Empty() {
```
hickeng: Can we collapse this to a common call of:
```
	e, err = v.getEndpoint(ctx, conf, input.ExternalNetwork.Name, "external", "external", true, input.VCHDNS)
```

and move the log message and check for Empty/Static into getEndpoint? We're unpacking every element of `NetworkConfig` in the arguments to `getStaticEndpoint` and I don't see it being called from anywhere else.

IMO this would make this portion of the code significantly more readable

tests/test-cases/Group6-VIC-Machine/6-7-Create-Network.md
```diff
@@ -159,3 +159,35 @@ This test requires that a vSphere server is running and available
...
+   networks to be on the same port group
+2. Verify debug output shows specified static IP address correctly assigned and copied to client
+   and external management networks
+
+# Test Steps
+1. Create with static IP address for external network and specify client and management networks to
+   be on different port group
+2. Verify debug output shows specified static IP address correctly assigned
+3. Verify debug output shows client and management networks set to DHCP
+
```
andrewtchin: added one for this

pkg/ip/ip.go
```diff
@@ -159,6 +159,24 @@ func (i *Range) UnmarshalText(text []byte) error {
 	return nil
 }
 
+// ParseIPandMask parses a CIDR format address (e.g. 1.1.1.1/8)
+func ParseIPandMask(s string) (net.IPNet, error) {
```
andrewtchin: I tried to replace usage with `netlink.ParseIPNet`, but that gave me compile errors for the windows exe

cmd/vic-machine/create/create.go
```diff
@@ -118,32 +128,68 @@ func (c *Create) Flags() []cli.Flag {
...
 		cli.StringFlag{
+			Name:        "client-network, cln",
+			Value:       "",
+			Usage:       "The client network port group name (restricts DOCKER_API access to this network)",
+			Destination: &c.clientNetworkName,
+		},
+		cli.StringFlag{
+			Name:        "client-network-gateway",
+			Value:       "",
+			Usage:       "Gateway for the VCH on the client network, e.g. 172.16.0.1/24",
```
andrewtchin: fixed

cmd/vic-machine/create/create.go
```diff
@@ -79,6 +79,16 @@ type Create struct {
 	containerNetworksDNS      cli.StringSlice
 	volumeStores              cli.StringSlice
 	insecureRegistries        cli.StringSlice
+	vchDNS                    cli.StringSlice
```
andrewtchin: changed to `dns`

lib/install/validate/network.go
```diff
@@ -34,38 +34,46 @@ import (
...
-		conf.AddNetwork(&executor.NetworkEndpoint{
-			Network: executor.ContainerNetwork{
-				Common: executor.Common{
-					Name: contNetName,
-					ID:   moid,
-				},
-			},
-		})
+			Default:     def,
+			Nameservers: ns,
```
andrewtchin: * With DHCP on all networks, the user specified nameserver is currently ignored and only the DHCP provided nameserver ends up in `resolv.conf`. 

If this behavior is acceptable, I will add a warning that user input nameservers are ignored. If not, I'll see how `resolv.conf` is populated since we are currently passing the nameserver setting in the `NetworkEndpoint` but it's getting ignored in this case.

* With a static IP and user specified DNS servers and all networks on a single port group, we `resolv.conf` has only the user specified nameserver. 

* With a static IP on one network, DHCP on other networks, and a user specified nameserver, `resolv.conf` has both the user specified and DHCP provided nameservers.

lib/install/validate/network.go
andrewtchin: 👍

lib/install/validate/network.go
andrewtchin: A+ good refactor

cmd/vic-machine/create/create.go
```diff
@@ -118,32 +128,68 @@ func (c *Create) Flags() []cli.Flag {
...
+			Value:       "",
+			Usage:       "IP address for the VCH on the client network, e.g. 172.16.0.2/24",
+			Destination: &c.clientNetworkIP,
+		},
+		cli.StringFlag{
 			Name:        "external-network, en",
 			Value:       "",
-			Usage:       "The external network (can see hub.docker.com)",
-			Destination: &c.ExternalNetworkName,
+			Usage:       "The external network port group name (can see hub.docker.com). Defaults to 'VM Network'",
```
andrewtchin: fixed

cmd/vic-machine/create/create.go
```diff
@@ -414,6 +481,116 @@ func (c *Create) processContainerNetworks() error {
 	return nil
 }
 
+// processClientNetwork parses client network args if present
+func (c *Create) processClientNetwork() error {
```
hmahmood: Aren't `processClientNetwork`, `processExternalNetwork`, and `processManagementNetwork` essentially the same but with different names? Looks like a DRY violation.

andrewtchin: A+ refactored

hmahmood:
lgtm

lib/install/validate/network.go
andrewtchin: added warning from (1)

anchal-agrawal:
LGTM

tests/test-cases/Group6-VIC-Machine/6-7-Create-Network.md
```diff
@@ -159,3 +159,35 @@ This test requires that a vSphere server is running and available
...
+# Test Steps
+1. Create with static IP address for each external network and client network and specify both to be
+   on the same port group
+2. Verify output shows configuration error and install does not proceed
```
andrewtchin: added a test case and a check for this to warn for unsupported config

lib/install/validate/network.go
```diff
@@ -74,40 +76,148 @@ func (v *Validator) checkNetworkConflict(bridgeNetName, otherNetName, otherNetTy
...
+			log.Error("All VCH networks on the same port group must have the same IP address")
+			log.Error("To resolve this, configure static IP for one network and assign other networks to same port group.")
+			log.Error("The static IP will be automatically configured for networks sharing the port group.")
+			return fmt.Errorf("Incorrect static IP configuration for networks on port group %q", pg)
+		}
+	}
+
+	// check if same subnet assigned to multiple portgroups - this can cause routing problems
+	networks := make(map[string]string)
+	for n, config := range ips {
```
hmahmood: This and the previous loop can't be combined?

andrewtchin: fixed

hmahmood:
lgtm

andrewtchin:
self lgtm ♻️

PR #2553, opened 2016-10-04
[full ci] Certificate generation and TLS verify

hickeng:
Fixes #2197 

This expands the certificate generation in vic-machine to include creation
or use of a CA, and creation or use of signed certificates.
This enables user authentication via client certificates

This does require either a known IP address or a known DNS name (can use
wildcards for a single FQDN element, e.g. *.vmware.com, but not
*.*.vmware.com)

Still pending are all of the updates to Robot to make use of this - blocked
by either predictable DNS names or static IPs

PR #2554, opened 2016-10-04
Add vSphere certificate verification support [full ci]

dougm:
- Un-hardcode insecure flag for ESX/VC certificate verification support

- Add support for certificate thumbprint based verification

### DIFF
--- a/cmd/vcsim/main.go
+++ b/cmd/vcsim/main.go
@@ -15,7 +15,9 @@
 package main
 
 import (
+	"crypto/tls"
 	"flag"
+	"fmt"
 	"log"
 	"runtime"
 
@@ -35,12 +37,15 @@ func main() {
 	flag.IntVar(&model.Pool, "pool", model.Pool, "Number of resource pools per compute resource")
 
 	isESX := flag.Bool("esx", false, "Simulate standalone ESX")
+	isTLS := flag.Bool("tls", false, "Enable TLS")
+	cert := flag.String("tlscert", "", "Path to TLS certificate file")
+	key := flag.String("tlskey", "", "Path to TLS key file")
 
 	flag.Parse()
 
 	f := flag.Lookup("httptest.serve")
 	if f.Value.String() == "" {
-		f.Value.Set("localhost:8989")
+		_ = f.Value.Set("127.0.0.1:8989")
 	}
 
 	if *isESX {
@@ -64,5 +69,21 @@ func main() {
 		log.Fatal(err)
 	}
 
-	model.Service.NewServer()
+	if *isTLS {
+		model.Service.TLS = new(tls.Config)
+		if *cert != "" {
+			c, err := tls.LoadX509KeyPair(*cert, *key)
+			if err != nil {
+				log.Fatal(err)
+			}
+
+			model.Service.TLS.Certificates = []tls.Certificate{c}
+		}
+	}
+
+	s := model.Service.NewServer()
+
+	fmt.Printf("GOVC_URL=%s", s.URL)
+
+	select {}
 }
--- a/cmd/vic-machine/common/target.go
+++ b/cmd/vic-machine/common/target.go
@@ -31,8 +31,9 @@ import (
 type Target struct {
 	URL *url.URL
 
-	User     string
-	Password *string
+	User       string
+	Password   *string
+	Thumbprint string
 }
 
 func NewTarget() *Target {
@@ -56,6 +57,11 @@ func (t *Target) TargetFlags() []cli.Flag {
 			Name:  "password, p",
 			Value: flags.NewOptionalString(&t.Password),
 			Usage: "ESX or vCenter password",
+		},
+		cli.StringFlag{
+			Name:        "thumbprint",
+			Destination: &t.Thumbprint,
+			Usage:       "ESX or vCenter host certificate thumbprint",
 		},
 	}
 }
--- a/lib/apiservers/portlayer/restapi/options/options.go
+++ b/lib/apiservers/portlayer/restapi/options/options.go
@@ -20,7 +20,7 @@ type PortLayerOptionsType struct {
 	SDK       string        `long:"sdk" description:"SDK URL or proxy" env:"VC_URL" required:"true"`
 	Cert      string        `long:"cert" description:"Client certificate" env:"VC_CERTIFICATE"`
 	Key       string        `long:"key" description:"Private key file" env:"VC_PRIVATE_KEY"`
-	Insecure  bool          `long:"insecure" default:"true" description:"Skip verification of server certificate" env:"VC_INSECURE"`
+	Insecure  bool          `long:"insecure" default:"false" description:"Skip verification of server certificate" env:"VC_INSECURE"`
 	Keepalive time.Duration `long:"keepalive" default:"5m" description:"Session timeout" env:"VC_KEEPALIVE"`
 
 	DatacenterPath string `long:"datacenter" default:"/ha-datacenter" description:"Datacenter path" env:"DC_PATH" required:"true"`
--- a/lib/config/virtual_container_host.go
+++ b/lib/config/virtual_container_host.go
@@ -155,6 +155,8 @@ type Connection struct {
 	ExtensionName string `vic:"0.1" scope:"read-only" key:"extension_name"`
 	// Whether the session connection is secure
 	Insecure bool `vic:"0.1" scope:"read-only" key:"insecure"`
+	// TargetThumbprint is the SHA-1 digest of the Target's public certificate
+	TargetThumbprint string `vic:"0.1" scope:"read-only" key:"target_thumbprint"`
 	// The session timeout
 	Keepalive time.Duration `vic:"0.1" scope:"read-only" key:"keepalive"`
 }
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -477,7 +477,6 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 				"/sbin/vicadmin",
 				"-docker-host=unix:///var/run/docker.sock",
 				// FIXME: hack during config migration
-				"-insecure",
 				"-ds=" + conf.ImageStores[0].Host,
 				"-cluster=" + settings.ClusterPath,
 				"-pool=" + settings.ResourcePoolPath,
--- a/pkg/vsphere/session/session.go
+++ b/pkg/vsphere/session/session.go
@@ -25,7 +25,6 @@
 package session
 
 import (
-	// "bytes"
 	"crypto/tls"
 	"fmt"
 	"strings"
@@ -53,6 +52,8 @@ type Config struct {
 	Service string
 	// Allow insecure connection to Service
 	Insecure bool
+	// Target thumbprint
+	Thumbprint string
 	// Keep alive duration
 	Keepalive time.Duration
 
@@ -129,6 +130,7 @@ func (s *Session) Create(ctx context.Context) (*Session, error) {
 	s.ExtensionKey = vchExtraConfig.ExtensionKey
 	s.ExtensionCert = vchExtraConfig.ExtensionCert
 	s.ExtensionName = vchExtraConfig.ExtensionName
+	s.Thumbprint = vchExtraConfig.TargetThumbprint
 
 	_, err = s.Connect(ctx)
 	if err != nil {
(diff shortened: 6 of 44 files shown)


pkg/vsphere/session/session.go
```diff
@@ -150,60 +151,85 @@ func (s *Session) Create(ctx context.Context) (*Session, error) {
...
+	// TODO: get rid of govmomi.Client usage, only provides a few helpers we don't need.
+	s.Client = &govmomi.Client{
+		Client:         vimClient,
+		SessionManager: session.NewManager(vimClient),
 	}
+
+	err = login(ctx)
 	if err != nil {
-		return nil, errors.Errorf("Failed to log in to %s: %s", soapURL.String(), err)
+		return nil, errors.Errorf("Failed to log in to %s: %s", s.URL().Host, err)
```
hickeng: I've had this crash with s.Client being nil. Think that was before the login call was added above, but fyi

dougm: Reproduced it implementing Session tests against the simulator, will wrap those up after lunch.

cmd/vic-machine/common/target.go
```diff
@@ -56,6 +58,16 @@ func (t *Target) TargetFlags() []cli.Flag {
 			Value: flags.NewOptionalString(&t.Password),
 			Usage: "ESX or vCenter password",
 		},
+		cli.BoolTFlag{ // TODO: s/BoolTFlag/BoolFlag/ - should be false by default
+			Name:        "insecure",
+			Destination: &t.Insecure,
+			Usage:       "Skip verification of ESX or vCenter host certificate",
+		},
+		cli.GenericFlag{
+			Name:  "thumbprint",
```
hickeng: what does this flag mean if the value string is omitted? Should this be cli.StringFlag?

lib/install/management/appliance.go
```diff
@@ -465,8 +465,8 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 			Args: []string{
 				"/sbin/vicadmin",
 				"-docker-host=unix:///var/run/docker.sock",
-				// FIXME: hack during config migration
-				"-insecure",
+				fmt.Sprintf("-insecure=%t", conf.Insecure),
```
hickeng: These args are here primarily as a legacy effect.
Any new parameters that have been added are consumed directly from the Decoded config in the process instead of being passed in on the cmd line.

These are still here because I didn't have time to migrate all of the uses in the appliance components when adding the extraconfig mechanism to vic-machine

pkg/vsphere/session/session.go
```diff
@@ -53,6 +52,8 @@ type Config struct {
 	Service string
 	// Allow insecure connection to Service
 	Insecure bool
+	// Target thumbprint
+	Thumbprint string
```
hickeng: Thumbprint can be set directly, see line 130-132

pkg/vsphere/session/session.go
```diff
@@ -150,60 +151,85 @@ func (s *Session) Create(ctx context.Context) (*Session, error) {
...
+
+	vimClient, err := vim25.NewClient(ctx, soapClient)
+	if err != nil {
+		return nil, errors.Errorf("Failed to connect to %s: %s", soapURL.Host, err)
 	}
 
 	if s.Keepalive != 0 {
-		// now that we've verified everything, enable keepalive
-		s.RoundTripper = session.KeepAlive(s.Client.RoundTripper, s.Keepalive)
+		// TODO: add login() to the keep alive handler
```
hickeng: Yes!!! Assuming login does what I think it does...

pkg/vsphere/simulator/simulator_test.go
```diff
@@ -266,6 +271,74 @@ func TestServeHTTP(t *testing.T) {
...
+		t.Fatal(err)
+	}
+
+	// Test custom RootCAs list
+	sc = soap.NewClient(ts.URL, false)
+	caFile, err := ts.CertificateFile()
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err = sc.SetRootCAs(caFile); err != nil {
```
hickeng: No objection to this, but wondering if it'd be easier just to pass around the certificate directly instead of dealing with tmp files.

tests/test-cases/Group6-VIC-Machine/6-4-Create-Basic.robot
```diff
@@ -74,6 +74,27 @@ Create VCH - target URL
     Run Regression Tests
     Cleanup VIC Appliance On Test Server
 
+Create VCH - thumbprint verification
+    Log To Console  \nRunning vic-machine create - thumbprint verification
+    Set Test Environment Variables
+    # TODO: need an integration/vic-test update to include the about.cert command
+    #${rc}  ${thumbprint}=  Run And Return Rc And Output  govc about.cert -k | jq -r .ThumbprintSHA1
+    ${rc}  ${thumbprint}=  Run And Return Rc And Output  openssl s_client -connect $(govc env -x GOVC_URL_HOST):443 </dev/null 2>/dev/null | openssl x509 -fingerprint -noout | cut -d= -f2
```
hickeng: You should be able to use `%{TEST_URL}` instead of the inlined `govc env` comment

cmd/vic-machine/common/target.go
```diff
@@ -56,6 +58,16 @@ func (t *Target) TargetFlags() []cli.Flag {
 			Value: flags.NewOptionalString(&t.Password),
 			Usage: "ESX or vCenter password",
 		},
+		cli.BoolTFlag{ // TODO: s/BoolTFlag/BoolFlag/ - should be false by default
+			Name:        "insecure",
```
hickeng: If we specify a thumbprint at all, doesn't that imply --insecure=false? This could be added to the processCertificates method.

lib/apiservers/portlayer/restapi/options/options.go
```diff
@@ -17,11 +17,12 @@ package options
 import "time"
 
 type PortLayerOptionsType struct {
-	SDK       string        `long:"sdk" description:"SDK URL or proxy" env:"VC_URL" required:"true"`
-	Cert      string        `long:"cert" description:"Client certificate" env:"VC_CERTIFICATE"`
-	Key       string        `long:"key" description:"Private key file" env:"VC_PRIVATE_KEY"`
-	Insecure  bool          `long:"insecure" default:"true" description:"Skip verification of server certificate" env:"VC_INSECURE"`
-	Keepalive time.Duration `long:"keepalive" default:"5m" description:"Session timeout" env:"VC_KEEPALIVE"`
+	SDK        string        `long:"sdk" description:"SDK URL or proxy" env:"VC_URL" required:"true"`
+	Thumbprint string        `long:"thumbprint" description:"SDK thumbprint" required:"false"`
```
hickeng: if the thumbprint is loaded directly in session.go then you can drop all of the changes related to argument passing

cmd/vic-machine/common/target.go
dougm: Yes, specifying thumbprint implies insecure=false. I'm not seeing a method named processCertificates anywhere tho?

cmd/vic-machine/common/target.go
dougm: It should be a cli.StringFlag, I'll fix that.

lib/install/management/appliance.go
dougm: Ah I didn't think of that, and passing the flags around was the most painful part of this whole thing ;)  I'll undo the flag passing.

pkg/vsphere/session/session.go
dougm: I see what you mean, will do.

pkg/vsphere/session/session.go
dougm: Indeed it does

pkg/vsphere/simulator/simulator_test.go
dougm: It was meant for testing govc, which now has a -tls-known-hosts option for thumbprint checking.  vcsim/main.go would output the path in env var form (as it does for GOVC_URL).  Can be used for other out-of-process clients too, but we won't need for the vic in-process clients.

tests/test-cases/Group6-VIC-Machine/6-4-Create-Basic.robot
dougm: TEST_URL contains more than the hostname, s_client fails with 'getservbyname failure' if you do that.  It'll get changed to the line above anyhow.

lib/apiservers/portlayer/restapi/options/options.go
dougm: Will do.

cmd/vic-machine/common/target.go
hickeng: drat, must have added it in my PR.
https://github.com/vmware/vic/pull/2553/files#diff-5a99f2371e79e7b3ef78eefdc3419581R443

dougm:
Another change I need to make, in the VC case we're using host service tickets for datastore HTTP access.  The service ticket response includes the host's thumbprint, but I'll need to refactor on the govmomi side a bit as this should be transparent.

cmd/vic-machine/common/target.go
hickeng: @dougm PR is merged

cmd/vcsim/main.go
```diff
@@ -35,12 +37,15 @@ func main() {
 	flag.IntVar(&model.Pool, "pool", model.Pool, "Number of resource pools per compute resource")
 
 	isESX := flag.Bool("esx", false, "Simulate standalone ESX")
+	isTLS := flag.Bool("tls", false, "Enable TLS")
```
vburenin: I see we are heavily using builtin flag module for some reason. Is there really any need for that? I always feel built-in golang flag module is a little clunky. I think go-flags package is much nicer.

dougm: Yeah we're using too many flags packages stdlib, go-flags, urfave/cli and govmomi has its own.  The only user facing one in VIC is urfave/cli (via vic-machine).  We should discuss the topic outside of this PR.

dougm:
vmware/govmomi#603 is pending.  I'm in the middle of the VIC side of the changes + tests, will need till tomorrow to wrap those up.

dougm:
Not sure how to deal with Test 11-1 - Upgrade, it downloads a build from bintray, runs vic-machine create using the Util.robot helpers - but the downloaded build does not have the new `-thumbprint` flag.

emlin:
The upgrade test is just try to make sure it can work from a earlier version. After GA, we gonna have more limitation there. But right now just update the version to a close one with you expected feature.

dougm:
@emlin the test can't upgrade since there's not a version on bintray that supports the new flag, until this change is merged.

emlin:
That's fine. Just disable the test, and then add back as Hasan did for his change.
Basically we cannot migrate data at this time, any changes related to configuration will impact upgrade. Btw, can you have default value for thumbprint in guestinfo? So that even without that option, vch still works well?

dougm:
ok I will disable the test then.  Regarding thumbprint value, there's a TODO in there where we can prompt to the user to accept the thumbprint if not specified.  And `-force` will also accept the host thumbprint without prompt.

pkg/vsphere/session/session.go
```diff
@@ -157,53 +159,60 @@ func (s *Session) Connect(ctx context.Context) (*Session, error) {
...
+
+	vimClient, err := vim25.NewClient(ctx, soapClient)
+	if err != nil {
+		return nil, errors.Errorf("Failed to connect to %s: %s", soapURL.Host, err)
+	}
+
 	if s.Keepalive != 0 {
-		// now that we've verified everything, enable keepalive
-		s.RoundTripper = session.KeepAlive(s.Client.RoundTripper, s.Keepalive)
+		// TODO: add login() to the keep alive handler
```
hickeng: What's the implication of not having this done?

pkg/vsphere/simulator/simulator_test.go
```diff
@@ -266,6 +271,82 @@ func TestServeHTTP(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if pinfo.ThumbprintSHA1 != sinfo.ThumbprintSHA1 {
+		t.Error("thumbprint mismatch")
+	}
+
+	// Test custom RootCAs list
+	sc = soap.NewClient(ts.URL, false)
+	caFile, err := ts.CertificateFile()
```
hickeng: You could also use the pkg/certificate to generate a CA and server certs for the IP addresses vcsim is currently listening on.

tests/test-cases/Group6-VIC-Machine/6-5-Create-Validation.robot
```diff
@@ -26,10 +26,22 @@ Suggest resources - Invalid target path
...
+Create VCH - target thumbprint verification
+    Log To Console  \nRunning vic-machine create - thumbprint verification
+    Set Test Environment Variables
+    # Attempt to cleanup old/canceled tests
+    Run Keyword And Ignore Error  Cleanup Dangling VMs On Test Server
+    Run Keyword And Ignore Error  Cleanup Datastore On Test Server
+    Set Test VCH Name
+
+    Log To Console  \nInstalling VCH to test server...
+    ${output}=  Run  bin/vic-machine-linux create --thumbprint=NOPE --name=${vch-name} --target="%{TEST_USERNAME}:%{TEST_PASSWORD}@%{TEST_URL}" --image-store=%{TEST_DATASTORE} ${vicmachinetls}
```
hickeng: could specify other invalid arguments here (such as invalid image store) to ensure that thumbprint is the first failure,

hickeng:
LGTM - minor comments

pkg/vsphere/simulator/simulator_test.go
dougm: Don't want to depend on vic/pkg here since we plan to move pkg/vsphere/simulator to the govmomi repo at some point.

tests/test-cases/Group6-VIC-Machine/6-5-Create-Validation.robot
dougm: Good call, I'll just change the existing -image-store value for now.

cmd/vcsim/main.go
```diff
@@ -64,5 +69,21 @@ func main() {
...
+				log.Fatal(err)
+			}
+
+			model.Service.TLS.Certificates = []tls.Certificate{c}
+		}
+	}
+
+	s := model.Service.NewServer()
+
+	fmt.Printf("GOVC_URL=%s", s.URL)
```
vburenin: Is it supposed to be print, not log?

lib/config/virtual_container_host.go
```diff
@@ -155,6 +155,8 @@ type Connection struct {
 	ExtensionName string `vic:"0.1" scope:"read-only" key:"extension_name"`
 	// Whether the session connection is secure
 	Insecure bool `vic:"0.1" scope:"read-only" key:"insecure"`
+	// TargetThumbprint is the SHA-1 digest of the Target's public certificate
```
vburenin: Why SHA-1? It is no longer considered secure. We should go at least with SHA-256.

hmahmood:
lgtm

lib/config/virtual_container_host.go
dougm: That's what vCenter/ESX uses (see SDK related comments in the govmomi change).  I suppose we could support using either for the target host, but vCenter SDK will still be giving us SHA-1 to verify the ESX hosts.

cmd/vcsim/main.go
dougm: Yes, stdout will only contain env variables, for use by out-of-process clients.  For example, you can run with `-httptest.serve 127.0.0.1:0` and the value of GOVC_URL contains the randomly chosen port.

dougm:
@vmware/vic-maintainers 

With this change merged, you'll need use `-thumbprint` or `-force` if your VC/ESX cert is not signed by a trusted CA

See Util.robot:Set Test Environment Variables for cli examples on obtaining a thumbprint, or look w/ your web browser

Will follow up with a small change to prompt / save the thumbprint similar to ssh known_hosts

dougm:
♻️

PR #2555, opened 2016-10-04
[full ci] allow ssh and shell enable via vic-machine

### LINKED ISSUE #1218: Appliance SSH status and password
Should we, or should we not, leave SSH installed and enabled on the appliance for TP3? If we do it should be a backup to the vicadmin log collection otherwise we will not improve that where it's lacking.

Irrespective, the root password for the appliance needs to be altered and we need to document how to determine what the password is - standard seems to be randomly generate and echo on the VM console.

**TP3 update:**
SSH is disabled by default on the appliance. To re-enable:
1. log into the appliance VM console using root/password
2. run `systemctl start sshd`

This will persist until appliance is rebooted.

hickeng:
This disables console and ssh access to the appliance VM by default.
Access can be restored by running vic-machine debug, with password and/or
ssh authorized key

This is something of a kludge on the vic-machine side at the moment as I
fully expect it to be merged into `vic-machine configure` when that's available.

Fixes #1218

### DIFF
--- a/cmd/vic-init/toolbox.go
+++ b/cmd/vic-init/toolbox.go
@@ -1,0 +1,178 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build !windows,!darwin
+
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/toolbox"
+)
+
+// startCommand is the switch for the synthetic commands that are permitted within the appliance.
+// This is not intended to allow arbitrary commands to be executed.
+func startCommand(r *toolbox.VixMsgStartProgramRequest) (int, error) {
+	switch r.ProgramPath {
+	case "enable-ssh":
+		return -1, enableSSH(r.Arguments)
+	case "passwd":
+		return -1, passwd(r.Arguments)
+	default:
+		return -1, fmt.Errorf("unknown command %q", r.ProgramPath)
+	}
+}
+
+// enableShell changes the root shell from /bin/false to /bin/bash
+func enableShell() error {
+	defer trace.End(trace.Begin(""))
+
+	chsh := exec.Command("/bin/chsh", "-s", "/bin/bash", "root")
+	err := chsh.Start()
+	if err != nil {
+		err := fmt.Errorf("Failed to launch chsh: %s", err)
+		log.Error(err)
+		return err
+	}
+
+	// ignore the error - it's likely raced with child reaper, we just want to make sure
+	// that it's exited by the time we pass this point
+	chsh.Wait()
+
+	// confirm the change
+	file, err := os.Open("/etc/passwd")
+	if err != nil {
+		err := fmt.Errorf("Failed to open file to confirm change: %s", err)
+		log.Error(err)
+		return err
+	}
+
+	reader := bufio.NewReader(file)
+	line, err := reader.ReadString('\n')
+	if err != nil {
+		err := fmt.Errorf("Failed to read line from file to confirm change: %s", err)
+		log.Error(err)
+		return err
+	}
+
+	// assert that first line is root
+	if !strings.HasPrefix(line, "root") {
+		err := fmt.Errorf("Expected line to start with root: %s", line)
+		log.Error(err)
+		return err
+	}
+
+	// assert that first line is root
+	if !strings.HasSuffix(line, "/bin/bash\n") {
+		err := fmt.Errorf("Expected line to end with /bin/bash: %s", line)
+		log.Error(err)
+		return err
+	}
+
+	return nil
+}
+
+// passwd sets the password for the root user to that provided as an argument
+func passwd(pass string) error {
+	defer trace.End(trace.Begin(""))
+
+	err := enableShell()
+	if err != nil {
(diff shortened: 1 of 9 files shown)


cmd/vic-init/toolbox.go
```diff
@@ -0,0 +1,178 @@
...
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/toolbox"
+)
+
+// startCommand is the switch for the synthetic commands that are permitted within the appliance.
+// This is not intended to allow arbitrary commands to be executed.
+func startCommand(r *toolbox.VixMsgStartProgramRequest) (int, error) {
+	switch r.ProgramPath {
```
fdawg4l: May want to log that there's a command being processed here.

cmd/vic-init/toolbox.go
```diff
@@ -0,0 +1,178 @@
...
+		err := fmt.Errorf("Failed to launch chsh: %s", err)
+		log.Error(err)
+		return err
+	}
+
+	// ignore the error - it's likely raced with child reaper, we just want to make sure
+	// that it's exited by the time we pass this point
+	chsh.Wait()
+
+	// confirm the change
```
fdawg4l: Is this really required?  It's not enough to check for non-negative exit status from `chsh`?

fdawg4l: Ok, now I read something about a child reaper below and I think I understand; you can't rely on exit status.  I get it.   Disregard my comment.

lib/install/management/debug.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/vm"
+)
+
+func (d *Dispatcher) DebugVCH(vch *vm.VirtualMachine, conf *config.VirtualContainerHostConfigSpec, password, authorized_key string) error {
+	defer trace.End(trace.Begin(conf.Name))
+
+	op, err := trace.FromContext(d.ctx)
+	if err != nil {
+		op = trace.NewOperation(d.ctx, "enable appliance debug")
```
fdawg4l: \m/

cmd/vic-init/toolbox.go
```diff
@@ -0,0 +1,178 @@
...
+		log.Error(err)
+		return err
+	}
+
+	_, err = stdin.Write([]byte("root:" + pass))
+
+	// so that we're actively waiting when the process exits, or we'll race (and lose) to child reaper
+	var waitErr error
+	go func() {
+		waitErr = setPasswd.Wait()
```
fdawg4l: Doesn't look like `waitErr` is needed.  How about just `_ = setPasswd.Wait()`?

cmd/vic-init/toolbox.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/toolbox"
+)
+
+// startCommand is the switch for the synthetic commands that are permitted within the appliance.
+// This is not intended to allow arbitrary commands to be executed.
+func startCommand(r *toolbox.VixMsgStartProgramRequest) (int, error) {
+	switch r.ProgramPath {
+	case "enable-ssh":
+		return -1, enableSSH(r.Arguments)
```
fdawg4l: Not really a comment.  Looking at `VixCommandHandler`, it's not immediately obvious what the `int` is used for.  Can you clarify the `-1`?

cmd/vic-machine/debug/debug.go
```diff
@@ -0,0 +1,163 @@
...
+	executor *management.Dispatcher
+
+	enableSSH     bool
+	password      string
+	authorizedKey string
+}
+
+func NewDebug() *Debug {
+	d := &Debug{}
+	d.Data = data.NewData()
```
fdawg4l: We have the best package names.

fdawg4l:
:shipit:

cmd/vic-init/toolbox.go
hickeng: @dougm help me out here - this was borrowed from you code.

cmd/vic-init/toolbox.go
```diff
@@ -0,0 +1,178 @@
...
+			err := fmt.Errorf("unable to create authorized_keys: %s", err)
+			log.Error(err)
+			return err
+		}
+	}
+
+	// because init is explicitly reaping child processes we cannot use simple
+	// exec commands to gather status
+	startSSH := exec.Command("/usr/bin/systemctl", "start", "sshd")
+	out, err := startSSH.CombinedOutput()
```
dougm: looks like you'd want to check and/or return err here

dougm:
lgtm

Keep in mind wrt StartProgramInGuest, RPC calls are serialized on the VMX side as well as within the toolbox.   If we're doing anything in a start program impl that may take longer than a few hundred millis, we may want to consider making it async.  We'd kick off a process and/or go routine and return a pid (real or pseudo) before the work is done, to prevent blocking the RPC queue.  Downside is the client then needs to poll for status, such as with GuestProcessManager.ListProcessesInGuest and of course we'd need to maintain some sort of state table within the toolbox.  Doesn't look like we need to at this point, just wanted to make a note of it.

cmd/vic-init/toolbox.go
hickeng: done

cmd/vic-machine/debug/debug.go
hickeng: and so much stuttering

cmd/vic-init/toolbox.go
hickeng: Can't - there's no guarantee that Wait gets the exit value for the process given the child reaper. We just have to hope (or attempt to confirm as with the prior operations)

dougm: Let's not wait for Wait here then, how about:
```go
func startSSH() error {
	c := exec.Command("/usr/bin/systemctl", "start", "sshd")

	var b bytes.Buffer
	c.Stdout = &b
	c.Stderr = &b

	if err := c.Start(); err != nil {
		return err
	}

	go func() {
		_ = c.Wait()
		log.Info("Attempted to start ssh service:\n %s", b.String())
	}()

	return nil
}
```

cmd/vic-init/toolbox.go
```diff
@@ -0,0 +1,182 @@
...
+	log.Info("Attempted to start ssh service:\n %s", out)
+
+	return nil
```
dougm: then `return startSSH()` here

hickeng:
poke

hickeng:
poke

PR #2566, opened 2016-10-04
Remove full datastore path from vm spec attribute VmPathName

### LINKED ISSUE #1888: docker create fails on 6.5_RC1 VC
build 3538

installed 6.5b3 VC and two 6.5b3 hosts. 

Installed to VC with 

`./vic-machine-darwin create --target 10.161.245.177 --image-datastore datastore1 --name vic-demo --user Administrator@vsphere.local --password 'Admin!23'  --compute-resource /ha-datacenter/host/10.160.97.176/Resources --bridge-network bridge --external-network vm-network`

Then tried create:

$ docker -H 10.162.13.107:2376 --tls create busybox
a3ed95caeb02: Pull complete
8ddc19f16526: Pull complete
Error: No such image: busybox

This also leaves files on the host that are marked as "deletable=false" and are unable to be deleted via API.

cgtexmex:
Removes the full datastore path from VmPathName in
the VM Spec.  Will now ALWAYS be [datastore name] -- i.e.
[datastore1]

fixes #1888

### DIFF
--- a/lib/spec/spec.go
+++ b/lib/spec/spec.go
@@ -108,17 +108,11 @@ func NewVirtualMachineConfigSpec(ctx context.Context, session *session.Session, 
 	fullName := fmt.Sprintf("%s-%s", prettyName, config.ID)
 	config.VMFullName = fullName
 
-	VMPathName := config.VMPathName
-	if !session.IsVSAN(ctx) {
-		// VMFS requires the full path to vmx or everything but the datastore is ignored
-		VMPathName = fmt.Sprintf("%s/%s/%s.vmx", config.VMPathName, config.VMFullName, config.ID)
-	}
-
 	s := &types.VirtualMachineConfigSpec{
 		Name: fullName,
 		Uuid: config.BiosUUID,
 		Files: &types.VirtualMachineFileInfo{
-			VmPathName: VMPathName,
+			VmPathName: config.VMPathName,
 		},
 		NumCPUs:             config.NumCPUs,
 		CpuHotAddEnabled:    &config.VMForkEnabled, // this disables vNUMA when true


lib/spec/spec.go
```diff
@@ -109,10 +109,6 @@ func NewVirtualMachineConfigSpec(ctx context.Context, session *session.Session,
 	config.VMFullName = fullName
 
 	VMPathName := config.VMPathName
```
dougm: You could get rid of that var too and just use `config.VMPathName` below.

dougm:
lgtm

fdawg4l:
lgtm

cgtexmex:
rebase -- removed unneeded variable

hickeng:
LGTM - however I don't recall why we were doing this in the first place, so @fdawg4l is the authoritative source.

dougm:
@hickeng it was the old trick to avoid creating a vm with the same name, avoiding the foo_1.vmx, foo_2.vmx, etc., renames.  I recall discovering this via source opengrok in the enatai/dcp days, but not finding any related docs.  Also recall sharing it with you in the early bonneville days, in the old kitchen, while you were on the couch enjoying a bowl of soup you crafted from canned V8 juice.
```
commit cb475c33d6fac2e12a1eb4e2d9050405a38da96a
Author: Doug MacEachern <dougm@vmware.com>
Date:   Fri Jan 9 10:55:12 2015 -0800

    Handle VM create/delete races

    Fixes enatai/dcp#157

    Change-Id: I4acf0ba951d30afc012e93d1ded4c963b563b42e

diff --git a/go/src/dcp/provisioning/esx/client.go b/go/src/dcp/provisioning/esx/client.go
index 41b85c2..370f21e 100644
--- a/go/src/dcp/provisioning/esx/client.go
+++ b/go/src/dcp/provisioning/esx/client.go
...
@@ -368,9 +359,13 @@ func (c *Client) CreateVM() (*govmomi.VirtualMachine, error) {
                return nil, err
        }

+       name := c.state.ID
+       // Use a full path to the config file to avoid creating a VM with the same name
+       cfgPath := fmt.Sprintf("[%s] %s/%s.vmx", ds.Name(), name, name)
+
        spec := types.VirtualMachineConfigSpec{
-               Name:  c.state.ID,
-               Files: &types.VirtualMachineFileInfo{VmPathName: fmt.Sprintf("[%s]", ds.Name())},
+               Name:  name,
+               Files: &types.VirtualMachineFileInfo{VmPathName: cfgPath},
```

caglar10ur:
```
bowl of soup you crafted from canned V8 juice
```

need to open a p0 issue for this :) LGTM

cgtexmex:
rebase

PR #2577, opened 2016-10-05
expose vic-machine upgrade command

### LINKED ISSUE #2545: Upgrade integration and integration test
Make upgrade command visible to user, and run upgrade end-2-end to fix any issues in this process.
Add integration test for upgrade

emlin:
Fixes #2545 
Rely on #2452, upgrade test can only work after the build with issue fixed.

### DIFF
--- a/lib/install/management/upgrade.go
+++ b/lib/install/management/upgrade.go
@@ -69,14 +69,6 @@ func (d *Dispatcher) Upgrade(vch *vm.VirtualMachine, conf *config.VirtualContain
 	if err = d.uploadImages(settings.ImageFiles); err != nil {
 		return errors.Errorf("Uploading images failed with %s. Exiting...", err)
 	}
-	var deleteImage bool
-	defer func() {
-		if deleteImage {
-			// do clean up aggressively, even the previous operation failed with context deadline excceeded.
-			d.ctx = context.Background()
-			d.deleteUpgradeImages(ds, settings)
-		}
-	}()
 
 	conf.BootstrapImagePath = fmt.Sprintf("[%s] %s/%s", conf.ImageStores[0].Host, d.vmPathName, settings.BootstrapISO)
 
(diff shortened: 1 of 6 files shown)


andrewtchin:
lgtm

lib/install/management/upgrade.go
```diff
@@ -96,10 +114,8 @@ func (d *Dispatcher) Upgrade(vch *vm.VirtualMachine, conf *config.VirtualContain
 		return err
 	}
 
-	// do clean up aggressively, even the previous operation failed with context deadline excceeded.
-	d.ctx = context.Background()
-	d.cleanSnapshot(*snapshotRefID, snapshotName, conf.Name)
-	d.deleteUpgradeImages(ds, settings)
+	deleteSnapshot = true
```
gigawhitlocks: though with these assignments here, is there a scenario in which the deleteSnapshot and deleteImage lambdas would _not_ be called at the bottom here? Because if they're always called, which seems to be the case, it would make more sense to me to break them out into helper functions & call them here directly, instead of using `defer` and testing to see if `deleteSnapshot` & `deleteImage` are `true`, since they always seem to be `true` at the end of the method here

gigawhitlocks:
Left some comments & then figured out how it worked & deleted them -- apologies if you saw them in the meantime!

LGTM

emlin:
> Left some comments & then figured out how it worked & deleted them -- apologies if you saw them in the meantime!

no, didn't see anything. thanks for your review.

lib/install/management/upgrade.go
```diff
@@ -69,17 +69,35 @@ func (d *Dispatcher) Upgrade(vch *vm.VirtualMachine, conf *config.VirtualContain
 	if err = d.uploadImages(settings.ImageFiles); err != nil {
 		return errors.Errorf("Uploading images failed with %s. Exiting...", err)
 	}
+	var deleteImage bool
+	defer func() {
```
vburenin: it seems like defer just complicates the code. I would rather call an appropriate function right before return, since it is clear where it should happens.

emlin: emm... I don't think defer itself make it complicated. delete or not itself is not a straightforward thing. And it combines with snapshot, which should be removed as well.

To add this is because I actually forgot to delete image in some branches before. I think this will make us more careful to add bail branch.

emlin:
♻️

emlin:
upgrade test passed, so cut off [full ci]

PR #2583, opened 2016-10-05
Remove Image Config from local cache on "docker rmi" command.

### LINKED ISSUE #2070: Remove image from image cache when rmi is called
`docker rmi` will remove images from the image store.  However, something needs to purge the entry from the image cache (in the docker daemon) if the image store call is successful.

vburenin:
When Image is removed it is still displayed upon image list request, it happened dues to it still in the cache. This change adds necessary logic to remove image from cache as well as an actually call to remove image config from the cache.

Fixes #2070


lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -220,7 +220,32 @@ func (ic *ICache) AddImage(imageConfig *metadata.ImageConfig) {
 		}
 		ic.cacheByName[imageConfig.Reference] = imageConfig
 	}
+}
+
+// RemoveImage removes image from the cache. The call return
```
fdawg4l: "The call return"

fdawg4l:
LGTM

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -181,7 +180,19 @@ func (ic *ICache) getImageByNamed(named reference.Named) (*metadata.ImageConfig,
...
 	}
 
-	return copyImageConfig(config), nil
+	return copyImageConfig(config)
+}
+
+// Don't assume the image id in image has "sha256:<id> as format.  We store it in
+// this format to make it easier to lookup by digest
+func prefixImageID(imageID string) string {
+	// Prefix was just "sha" which seems weird to be. Following general logic,
```
anchal-agrawal: `// Add the "sha256:" prefix to the image ID if missing`

would be more suitable.

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -181,7 +180,19 @@ func (ic *ICache) getImageByNamed(named reference.Named) (*metadata.ImageConfig,
...
+	// I have update it to "sha256:"
```
jzt: Docker allows for different algorithms (see [here](https://github.com/docker/docker/blob/b248de7e332b6e67b08a8981f68060e6ae629ccf/vendor/src/github.com/docker/distribution/digest/digester.go#L16)), but since we're pinned to 1.11.2, we know we will never need another one so this is fine.

Just FYI!

jzt:
LGTM 👍

lib/apiservers/engine/backends/cache/image_cache.go
vburenin: Yah, I was really hoping to get comment on that. Thanks for confirming.

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -181,7 +180,18 @@ func (ic *ICache) getImageByNamed(named reference.Named) (*metadata.ImageConfig,
...
+	return copyImageConfig(config)
+}
+
+// Add the "sha256:" prefix to the image ID if missing.
+// Don't assume the image id in image has "sha256:<id> as format.  We store it in
+// this format to make it easier to lookup by digest
+func prefixImageID(imageID string) string {
+	if strings.HasPrefix(imageID, "sha256:") {
+		return imageID
+	} else {
```
caglar10ur: you could drop the else and return directly

caglar10ur:
LGTM :sheep: :it:

lib/apiservers/engine/backends/cache/image_cache.go
vburenin: I could, it just looks nicer for some reason. I was not using else before some moment of time, but last several years I tend to write like this.

dougm: golint agrees with @caglar10ur :

```
if block ends with a return statement, so drop this else and outdent its block
```

Our CI used to require our code is go-lint, go-vet, etc. clean, we really need to fix this.  @vburenin do you not have go-lint integrated with your editor?  Maybe you could take a look?

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -220,7 +219,32 @@ func (ic *ICache) AddImage(imageConfig *metadata.ImageConfig) {
...
+	defer ic.m.Unlock()
 
+	// If we get here we definitely want to remove image config from any data structure
+	// where it can be present. So that, if there is something is wrong
+	// it could be tracked on debug level.
+	if err := ic.idIndex.Delete(imageConfig.ImageID); err != nil {
+		log.Debugf("Not found in image cache index: %v", err)
+	}
+
+	prefixedId := prefixImageID(imageConfig.ImageID)
```
dougm: Another golint warning that should have caused CI to fail:
```
var prefixedId should be prefixedID
```

PR #2584, opened 2016-10-05
 On demand serial port

caglar10ur:
Tether now starts the SSH server unconditionally.

This PR replaces hardcoded attach value with the computed one.

Personality now tracks the DetachError and calls unbind. This assumes that
we will have only 1 client attached at any given time and calls unbind
on exit. The multiple client case needed to be handled in portlayer
(mux) and requires a reference counting.

Still need

"Establish stdin connection immediately after attach #2406"

otherwise first re-attach fails
```
[] docker run -it busybox
/ #
[] docker ps
3dd5bb954d07        busybox             "sh"                18 hours ago        Up 6 seconds                            nostalgic_roentgen
[] docker attach nostalgic_roentgen
[] docker attach nostalgic_roentgen
/ #
```
This enables vmotion but doesn't make it %100 functional


caglar10ur:
Please ignore the commit starting with [dio] prefix as that's identical to #2440 and review the other

dougm:
lgtm

cmd/tether/attach.go
```diff
@@ -179,131 +168,147 @@ func (t *attachServerSSH) stop() {
...
 
-	var sConn *ssh.ServerConn
+	var serverConn *ssh.ServerConn
 	var chans <-chan ssh.NewChannel
 	var reqs <-chan *ssh.Request
 	var err error
 
-	// keep waiting for the connection to establish
-	for t.enabled && sConn == nil {
-		conn := t.conn
```
hickeng: was specifically using a copy of conn as it was getting nil'd elsewhere (presumably by stop) - doesn't look like we've got a lock here, so that possibility still exists. We cannot just hold the lock here while we access the conn because we cannot yield it during all the blocking operations, hence the copy.

cmd/tether/attach.go
```diff
@@ -179,131 +168,147 @@ func (t *attachServerSSH) stop() {
...
-			// never be nil otherwise
-			err := fmt.Errorf("connection provided for backchannel is nil")
-			log.Debug(err.Error())
-			return err
-		}
+	for t.enabled {
+		// keep waiting for the connection to establish
+		for serverConn == nil {
+			// tests are passing their own connections so do not create connections when testing is set
+			if !t.testing {
```
hickeng: Can we add a mock for this, or even add it into backchannel?

backchannel was originally opening the connection as well as running the handshake, but when this was split out as an extension that was moved into Start.
If we need to reinitialize the connection in this way for serial then putting it into backchannel is a much cleaner approach.

cmd/tether/attach.go
```diff
@@ -179,131 +168,147 @@ func (t *attachServerSSH) stop() {
...
-		if conn == nil {
-			// Stop has probably been called as t.conn is set in Start and should
-			// never be nil otherwise
-			err := fmt.Errorf("connection provided for backchannel is nil")
-			log.Debug(err.Error())
-			return err
-		}
+	for t.enabled {
+		// keep waiting for the connection to establish
+		for serverConn == nil {
```
hickeng: normally this loop should spend all it's time in backchannel waiting for a conn in the case where we're not attached.
There is a possibility that we spin fast on this if we hit error conditions, and that we max cpu if it's a persistent error cond.

cmd/tether/attach_windows.go
```diff
@@ -29,12 +29,37 @@ import (
...
+	// TODO: sort out the named port impl so that we can transparently switch from that to/from
+	// regular files for testing
+	// t.conn, err := serial.NewTypedConn(f, "file")
+	return nil, nil
+}
+
+func backchannel(ctx context.Context, conn *net.Conn) error {
 	// HACK: currently RawConn dosn't implement timeout so throttle the spinning
-	ticker := time.NewTicker(50 * time.Millisecond)
+	ticker := time.NewTicker(10 * time.Millisecond)
```
hickeng: why leave backchannel in Windows, but remove it in darwin? I appreciate you leaving the rawConnection logic as that's a good record for when we add windows support again.

lib/apiservers/engine/backends/container.go
```diff
@@ -1115,7 +1114,46 @@ func (c *Container) ContainerAttach(name string, ca *backend.ContainerAttachConf
...
+
+			unbind, err := client.Interaction.InteractionUnbind(interaction.NewInteractionUnbindParamsWithContext(ctx).
+				WithConfig(&models.InteractionUnbindConfig{
+					Handle: handle,
+				}))
+			if err != nil {
+				return InternalServerError(err.Error())
+			}
+
+			handle, ok = unbind.Payload.Handle.(string)
```
hickeng: This is just while we have a mix of handle types, correct?

cmd/tether/attach_windows.go
caglar10ur: I assumed that darwin code never used hence removed it not to create an illusion :)

cmd/tether/attach.go
caglar10ur: I can create a copy and use it through out in the loop

cmd/tether/attach.go
caglar10ur: Will try to move it into backchannel then

lib/apiservers/engine/backends/container.go
caglar10ur: hmm not sure, it's been a while let me check :)

cmd/tether/attach.go
```diff
@@ -179,131 +168,147 @@ func (t *attachServerSSH) stop() {
...
+			// tests are passing their own connections so do not create connections when testing is set
+			if !t.testing {
+				// close the connection if required
+				if t.conn != nil {
+					(*t.conn).Close()
```
vburenin: Why do you dereference it here? Is it a pointer to a pointer?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -491,6 +491,10 @@ func dockerContainerCreateParamsToPortlayer(cc types.ContainerCreateConfig, laye
 	config.WorkingDir = new(string)
 	*config.WorkingDir = cc.Config.WorkingDir
 
+	// attach
+	config.Attach = new(bool)
```
vburenin: config.Attach = swag.Bool(cc.Config.AttachStdin || cc.Config.AttachStdout || cc.Config.AttachStderr) - this should make things a little more clear. I spent several seconds trying to figure out what's going on. I also wonder if there are any references to this value from any other place.

lib/apiservers/engine/backends/container.go
caglar10ur: Remembered now

```
cannot use handle (type interface {}) as type string in argument to containers.NewCommitParamsWithContext(ctx).WithHandle: need type assertion
```

Commit expects them as strings, we need to change all portlayer handle usage to interface{} but that's another day's work :)

lib/apiservers/engine/backends/container_proxy.go
caglar10ur: hmm I think I leave it as it is to keep the consistency with the rest of the code but I agree we should start to switch swag types.

This Attach value gets serialized into guestinfo and then used on the other side (tether)

cmd/tether/attach.go
caglar10ur: https://github.com/vmware/vic/pull/2584/files/34734f29e4f5ec5dad73097c16a41c56f835ea9e#diff-76607e475c2cd3318bf43a8550e85a92R54

lib/apiservers/engine/backends/container_proxy.go
caglar10ur: Actually changed my mind again, replacing others with swag :)

caglar10ur:
@hickeng addressed some of your comments. My suggestion is - unless you think otherwise - is to merge this and then addressing remaining comments as bugs. What do you think?

cmd/tether/attach.go
hickeng: This is not new - I mention it because I'm not sure what happens when the port is disconnected. I *think* we should just block as normal, but ....

cmd/tether/attach_windows.go
hickeng: It doesn't - that code is only there to enforce somewhat portable coding between the possible container platforms.

hickeng:
@caglar10ur sure - let's get this in. Please add a comment in the code around refactoring the !testing into backchannel if that doesn't make it - it's not a model I want to see picked up elsewhere for 'consistency' ;)

caglar10ur:
♻️

hickeng:
Addresses a good chunk of #858

Issue #2503, opened 2016-09-28
Segfault in tether


mhagen-vmware:
https://ci.vmware.run/vmware/vic/5227

Failed it looks like because the container panic'ed and thus was unable to power off.

```
2016-09-28T19:09:02.712Z| vcpu-0| I120: Guest: <6>[    3.366594] tether[207]: segfault at 38 ip 00000000004e8236 sp 000000c42003bab0 error 4 in tether[400000+707000]
2016-09-28T19:09:02.712Z| vcpu-0| I120: Guest: <0>[    3.427570] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000200
2016-09-28T19:09:02.712Z| vcpu-0| I120+ <0>[    3.427570] 
2016-09-28T19:09:02.712Z| vcpu-0| I120: Guest: <4>[    3.429707] CPU: 0 PID: 203 Comm: tether Not tainted 4.4.8-esx #1-photon
2016-09-28T19:09:02.712Z| vcpu-0| I120: Guest: <4>[    3.431193] Hardware name: VMware, Inc. VMware Virtual Platform/440BX Desktop Reference Platform, BIOS 6.00 09/21/2015
2016-09-28T19:09:02.712Z| vcpu-0| I120: Guest: <4>[    3.434055]  0000000000000000 ffffffff81215caf ffffffff816ebe50 ffff880071ebbd18
2016-09-28T19:09:02.712Z| vcpu-0| I120: Guest: <4>[    3.434710]  ffffffff810c7be6 0000000000000010 ffff880071ebbd28 ffff880071ebbcc0
2016-09-28T19:09:02.712Z| vcpu-0| I120: Guest: <4>[    3.435340]  ffff88006f879740 0000000000000200 ffff880079498010 ffff880079497b58
2016-09-28T19:09:02.712Z| vcpu-0| I120: Guest: <4>[    3.435980] Call Trace:
2016-09-28T19:09:02.713Z| vcpu-0| I120: Guest: <4>[    3.436619]  [<ffffffff81215caf>] ? dump_stack+0x5c/0x7d
2016-09-28T19:09:02.713Z| vcpu-0| I120: Guest: <4>[    3.437226]  [<ffffffff810c7be6>] ? panic+0xbf/0x1d7
2016-09-28T19:09:02.713Z| vcpu-0| I120: Guest: <4>[    3.438277]  [<ffffffff81045193>] ? do_exit+0xa63/0xa70
[...]

PR #2585, opened 2016-10-05
Redirect tether & init stderr to serial

gigawhitlocks:
Investigating #2503 proved fruitless because the `panic` was not included in the log bundle after the segfault caused the container to kernel panic. 

To help us investigate this issue should it arise in the future, and/or other issues stemming from `vic-init` and/or the tether, I've redirected stderr to the serial port (thanks @hickeng for helping me with this; these two identical code blocks are his) and with this change, should the tether segfault, the container log in the log bundle will contain a backtrace so we can see what actually caused the segfault to begin with.


cmd/tether/main_linux.go
```diff
@@ -38,6 +38,10 @@ func main() {
 		halt()
 	}()
 
+	logFile, _ := os.OpenFile("/dev/ttyS1", os.O_WRONLY|os.O_SYNC, 0644)
```
caglar10ur: do not ignore the error here

cmd/vic-init/main_linux.go
```diff
@@ -41,6 +41,10 @@ func main() {
 		reboot()
 	}()
 
+	logFile, _ := os.OpenFile("/dev/ttyS1", os.O_WRONLY|os.O_SYNC, 0644)
+	syscall.Dup2(int(logFile.Fd()), 2)
```
caglar10ur: ditto

cmd/tether/main_linux.go
```diff
@@ -38,6 +38,10 @@ func main() {
...
+	syscall.Dup2(int(logFile.Fd()), 2)
```
caglar10ur: Forgot to add also please don't hardcode 2, instead use os.Stderr, while at it you could switch to dup3 as some architectures don't have dup2 :)
```
syscall.Dup3(int(logFile.Fd()), int(os.StdErr.Fd()), 0)
```

gigawhitlocks: Roger, will do both of these thank you for reviewing @caglar10ur

gigawhitlocks: Actually @caglar10ur I only made this change to vic-init (which is Photon) and tether-linux, which also runs on Photon -- are you sure that there needs to be a concern here regarding architectures?

caglar10ur: I  think what you are saying photon is x86_64 and dup2 will be there but let me remind you we can cross compile it by just setting GOOS and if you set it to arm64 (for some reason) this will fail :)

It's not a must change but I thought we could just use dup3 while at it :)

gigawhitlocks: Yep I agree with your reasoning

caglar10ur:
LGTM :sheep: :it:

anchal-agrawal:
:shipit:

Issue #2530, opened 2016-09-30
IDs returned by docker images doesn't match VIC vs vanilla docker


fdawg4l:
```
# VIC
fahmed@hasselhoff: ~/src/vic/src/github.com/vmware/vic/bin$ docker -H 192.168.8.129:2376 --tls pull ubuntu
Using default tag: latest
Pulling from library/ubuntu
a3ed95caeb02: Pull complete 
cad964aed91d: Pull complete 
3a80a22fea63: Pull complete 
50de990d7957: Pull complete 
61e032b8f2cb: Pull complete 
9f03ce1741bf: Pull complete 
Digest: sha256:c8d92743cd9933fc7bb085dafd4df600cd09d835c95a98c1ea05b504369d9654
Status: Downloaded newer image for library/ubuntu:latest
fahmed@hasselhoff: ~/src/vic/src/github.com/vmware/vic/bin$ docker -H 192.168.8.129:2376 --tls images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
ubuntu              latest              827eb99e90cf        3 days ago          127 MB


# Vanilla
fahmed@hasselhoff: ~/src/vic/src/github.com/vmware/vic/infra/machines/vcsa$ docker pull ubuntu
Using default tag: latest
latest: Pulling from library/ubuntu
cad964aed91d: Pull complete 
3a80a22fea63: Pull complete 
50de990d7957: Pull complete 
61e032b8f2cb: Pull complete 
9f03ce1741bf: Pull complete 
Digest: sha256:28d4c5234db8d5a634d5e621c363d900f8f241240ee0a6a978784c978fe9c737
Status: Downloaded newer image for ubuntu:latest
fahmed@hasselhoff: ~/src/vic/src/github.com/vmware/vic/infra/machines/vcsa$ docker images
REPOSITORY                                              TAG                 IMAGE ID            CREATED             SIZE
ubuntu                                                  latest              c73a085dc378        3 days ago          127.1 MB

```

Looks like whatever ID we're using to represent images in the personality's image cache isn't the same as  whatever vanilla docker uses.

UPDATE from @hickeng:
We also seem to have two possible image IDs that we present, with the second appearing after a layer has been deleted and you re-pull:
```
vagrant@devbox:~/vic$ docker -H 192.168.78.132:2376 --tls pull ubuntu
Using default tag: latest
Pulling from library/ubuntu
a3ed95caeb02: Pull complete
cad964aed91d: Pull complete
[...]

Issue #1772, opened 2016-08-02
Avoid concurrent WriteImage of the same image ID from multiple clients in the PL


fdawg4l:
We need a way to prevent clients from downloading the same image ID at the same time.  Add a map of IDs in flight, check against it for every `WriteImage`, and throw an error for subsequent clients trying to write the same ID

fdawg4l:
We already have an image cache in the port layer.  This seems like the appropriate place for this map.

Issue #1409, opened 2016-07-07
Context deadline during Pull causes an already exists error from govmomi on a subsequent pull


matthewavery:
During a `pull busybox` it looks like a context deadline occurred somewhere on the portlayer side during a write image causing an inconsistent state. On repeat of the pull the layer is found to already exists after trying a new right. On a subsequent create the busybox image is still not found and an attempt to pull it is made. The appropriate response should have been that the layer already exists on the second pull. 

```
 % sudo ./vic-machine-linux create --target skinner.eng.vmware.com --user root --image-datastore Datastore --password ca\$hc0w --bridge-network vch-test --name mavery-vch-1 --compute-resource /Datacenter/host/Austin/Resources/mavery --volume-store=test-store:Datastore/testStore --generate-cert=false
INFO[2016-07-07T14:26:22-05:00] ### Installing VCH ####                      
WARN[2016-07-07T14:26:22-05:00] Configuring without TLS - to enable use -generate-cert or -key/-cert parameters 
INFO[2016-07-07T14:26:22-05:00] Validating supplied configuration            
INFO[2016-07-07T14:26:22-05:00] Firewall status: DISABLED on /Datacenter/host/Austin/patty.eng.vmware.com 
INFO[2016-07-07T14:26:22-05:00] Firewall configuration OK on hosts:          
INFO[2016-07-07T14:26:22-05:00]   /Datacenter/host/Austin/patty.eng.vmware.com 
INFO[2016-07-07T14:26:23-05:00] License check OK on hosts:                   
INFO[2016-07-07T14:26:23-05:00]   /Datacenter/host/Austin/patty.eng.vmware.com 
INFO[2016-07-07T14:26:23-05:00] DRS check OK on:                             
INFO[2016-07-07T14:26:23-05:00]   /Datacenter/host/Austin/Resources/mavery   
INFO[2016-07-07T14:26:27-05:00] Creating Resource Pool mavery-vch-1          
INFO[2016-07-07T14:26:27-05:00] Creating directory [Datastore] testStore     
INFO[2016-07-07T14:26:27-05:00] Datastore path is [Datastore] testStore      
INFO[2016-07-07T14:26:27-05:00] Creating appliance on target                 
INFO[2016-07-07T14:26:27-05:00] Network role client is sharing NIC with external 
INFO[2016-07-07T14:26:27-05:00] Network role management is sharing NIC with external 
INFO[2016-07-07T14:26:30-05:00] Uploading images for container               
[...]

Issue #1298, opened 2016-06-30
Pulling the same image concurrently with multiple clients


jzt:
Some functionality needs to be added to deal with multiple clients attempting to pull the same image at the same time.

- [ ] The storage portlayer needs to return an "already exists" error when multiple clients call `WriteImage` simultaneously for the same image. One call will inevitably win, making it impossible for the losing caller(s) to create the vmdk on the backend. This occurrence needs to be reported to the losing caller(s) so that appropriate action can be taken - this will likely occur before the layer is fully created so the loser will need to wait for it to complete before continuing with other layers.

- [ ] Once imagec is converted into a library, it must be able to track what image layers are currently being pulled. This will allow other clients currently attempting to pull the same layers to avoid duplicate downloads, also to handle and display progress output properly.

jzt:
https://github.com/docker/docker/tree/master/distribution/xfer

This package is what docker uses to deal with the 2nd issue here.

jzt:
@hmahmood and I sad down with this yesterday and came up with an approach. It's currently prioritized under p0 bugs, but it will be the next thing I move to.

shadjiiski:
@jzt, anything new on this issue? I am using JFrog Bintray build 4715 and have just run this command in 2 different terminals:
```ts
docker --tls run -itd alpine:3.1 sh
```
Both `docker run` calls did not find the alpine:3.1 image locally and both started pulling it from registry, however one of them succeeded (and provisioned an alpine container) and the other one failed with
```ts
C:\Program Files\Docker Toolbox\docker.exe: Failed to write to image store: [POST /storage/{store_name}][500] WriteImage default  &{Code:0xc4203e6370 Message:file already exists}.
See 'C:\Program Files\Docker Toolbox\docker.exe run --help'.
```
From your first message on this issue I couldn't get whether this is the expected behavior or this will change to be just like it is in plain Docker (two fully functional alpine containers in the end, single alpine image downloaded). Could you please elaborate on this?

jzt:
@shadjiiski The plan is to have the functionality match that of Docker's, just as you described. Unfortunately it is a non-trivial change that is currently prioritized below other changes to the same area of code that are making this somewhat of a moving target. At the point when things settle down and higher-priority items like p0 bugs are addressed, we can begin the rather large refactor that will be necessary to achieve this behavior. Right now, the caveat exists that concurrent pulls will exhibit this behavior.

mdubya66:
[...]

PR #2587, opened 2016-10-05
Imagec library enabling concurrent downloads

### LINKED ISSUE #1439: Make imagec work as a library

imagec should be updated to work as a standalone binary or as a library. As a library, imagec could provide information to the docker persona about the image it just pulled. The docker persona can then add this metadata to the cache without having to trigger another request to the portlayer that will have to iterate over the entire layer list to find anything it needs to cache.

It needs to be refactored to use functions defined on a struct that can be instantiated once for each client making a call. This should solve any concurrency issues previously seen when two calls are trying to share the same progress output. Related to #1298.

jzt:
This change converts imagec into working as a library, while preserving the standalone binary in the process. It also adds support for multiple clients downloading the same images concurrently. There is now a layer cache in the persona that is used primarily for tracking layer downloads. It is refreshed at startup by way of the image cache.

The majority of the changes you will see in the diff are resulting from moving files around. Most of the meat in these changes are located in `lib/imagec/download.go`, `lib/apiservers/engine/backends/cache/layer_cache.go`, and `lib/imagec/imagec.go`. 

This has a ways to go, and feedback is always welcome.  Tagging @hickeng, @caglar10ur, @sflxn, @dougm, @hmahmood, @vburenin.

Known issues:
- ~~The biggest issue is an intermittent `unexpected EOF` error will kill the client while an image is downloading. It looks really bad, and is probably a really bad user experience/CI killer, but from the persona on down through the portlayer, the image continues to pull and everything works out in the end. There are two ways that the user can know this: One is by simply running the same command a second time. The persona will attach that client to the currently in-progress download, and the client can watch it complete. The other way is by watching the persona logs and seeing the image pull complete. A call to `docker images` after either case will show that it succeeded. I will file an issue to follow up on that problem once this is merged.~~
- ~~Intermittent errors like `malformed chunked encoding`, `invalid byte in chunk length`, and others. I suspect these are being caused by a race condition between multiple goroutines writing to the client's `Outstream` simultaneously. Again, these do not affect the image pull on the backend, only what the client perceives.  These errors seem to show up when pulling with >=3 clients simultaneously.~~

EOFs and other weird funkiness have been solved. 👍 thanks @hmahmood! 

- Image IDs are not consistent with Docker's (expected as of 1.12), but whats more, if you do a `docker rmi` on an image, only the leaf layer gets removed. When re-pulling the image, only the leaf layer is pulled (since the remaining layers already exist), and only that layer is used to calculate the imageID. Thus, not only are the image IDs inconsistent with Docker, they are also inconsistent between pulls. I have a fix in the works, see #2530. 

These issues will be captured as bugs once (if) this PR gets merged.

Fixes #1439, #1154, #1298, #1772, #1250, #1409


caglar10ur:
Quick question before reviewing it, have you tried running tests or compiling persona with -race flag?

jzt:
@caglar10ur yeah, I have, but I saw nothing unusual (no output that isn't normally produced by a `make all`).

lib/imagec/docker.go
```diff
@@ -146,18 +149,22 @@ func LearnAuthURL(options ImageCOptions) (*url.URL, error) {
 }
 
 // FetchToken fetches the OAuth token from OAuth endpoint
-func FetchToken(url *url.URL) (*Token, error) {
+func FetchToken(ic *ImageC, url *url.URL, progressOutput progress.Output) (*Token, error) {
```
caglar10ur: Why is FetchImageBlob getting the ctx but this one is creating one?

lib/apiservers/engine/backends/cache/layer_cache.go
```diff
@@ -0,0 +1,97 @@
...
+
+package cache
+
+import (
+	"sync"
+)
+
+// LCache is an in-memory cache to account for existing image layers
+// It is used primarily by imagec when coordinating layer downloads
+// The cache is initially hydrated by way of the image cache at startup
```
caglar10ur: why do we have a cache? Can't we use fs semantics to see the status of the download? We are already streaming the file into the fs (which is tmpfs) so we could do what downloaders are doing (create the lock file, download the file, move etc.) No?

lib/imagec/docker.go
caglar10ur: Actually while at it, we could add ctx to the LearnX ones and use timeouts on them too

lib/imagec/docker.go
```diff
@@ -146,18 +149,22 @@ func LearnAuthURL(options ImageCOptions) (*url.URL, error) {
...
+		Timeout:            ic.Timeout,
+		Username:           ic.Username,
+		Password:           ic.Password,
+		InsecureSkipVerify: ic.InsecureSkipVerify,
+	},
+		progressOutput)
+
+	ctx, cancel := context.WithTimeout(context.Background(), ic.Timeout)
+	defer cancel()
+	tokenFileName, err := fetcher.Fetch(ctx, url, progressOutput)
```
caglar10ur: an empty line would be useful

lib/imagec/download.go
```diff
@@ -0,0 +1,330 @@
...
+	cache.LayerCache().Commit(id)
+
+	// stop tracking the download transfer
+	delete(ldm.downloadsByID, id)
+}
+
+// DownloadLayers ensures layers end up in the portlayer's image store
+// It handles existing and simultaneous layer download de-duplication
+// This code is utilizes Docker's xfer package: https://github.com/docker/docker/tree/v1.11.2/distribution/xfer
+func (ldm *LayerDownloader) DownloadLayers(ctx context.Context, ic *ImageC) error {
```
caglar10ur: just curios, why we are keeping our download code around if we are using xfer pkg? Can't we use this for everything?

lib/imagec/download.go
```diff
@@ -0,0 +1,330 @@
...
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
+
+	"github.com/docker/docker/distribution/xfer"
+	"github.com/docker/docker/pkg/progress"
+)
+
+var ()
```
caglar10ur: left over?

lib/imagec/download.go
```diff
@@ -0,0 +1,330 @@
...
+	return d.err
+}
+
+// LayerDownloader keeps track of what layers are currently being downloaded
+type LayerDownloader struct {
+	tm xfer.TransferManager
+	// the list of layers currently in flight
+	downloadsByID map[string]*downloadTransfer
+
+	m sync.Mutex
```
caglar10ur: https://dmitri.shuralyov.com/idiomatic-go#mutex-hat :)

lib/imagec/download.go
```diff
@@ -0,0 +1,330 @@
...
+	)
+
+	// lock here so that we get all layers in flight before another client comes along
+	ldm.m.Lock()
+
+	// Create the ImageWithMeta slice to hold Image structs
+	// This also informs downloaders when a layer already exists in the store
+	layers, err := ic.LayersToDownload(progressOutput)
+	if err != nil {
+		log.Errorf(err.Error())
```
caglar10ur: ldm is locked here
```
ldm.m.Unlock()
```

lib/imagec/download.go
```diff
@@ -0,0 +1,330 @@
...
+		d := &downloadTransfer{
+			Transfer: xfer.NewTransfer(),
+			layer:    layer,
+		}
+
+		go func() {
+
+			defer func() {
+				close(progressChan)
+			}()
```
caglar10ur: defer close(progressChan)?

lib/imagec/download.go
```diff
@@ -0,0 +1,330 @@
...
+
+			// Write blob to the storage layer
+			if err := ic.WriteImageBlob(layer, progressOutput, imageLayer); err != nil {
+				log.Infof("Got error writing image blob: %s", err)
+				d.err = err
+				return
+			}
+
+			// update tags
+			// TODO(jzt): this does not appear to work any longer
```
caglar10ur: sounds like an issue to me :)

lib/imagec/download.go
```diff
@@ -0,0 +1,330 @@
...
+	return func(progressChan chan<- progress.Progress, start <-chan struct{}, inactive chan<- struct{}) xfer.Transfer {
+
+		d := &downloadTransfer{
+			Transfer: xfer.NewTransfer(),
+			layer:    layer,
+		}
+
+		go func() {
+			defer func() {
+				close(progressChan)
```
caglar10ur: same as above, am I missing something here. why we are not calling it like this?

lib/imagec/download.go
```diff
@@ -0,0 +1,330 @@
...
+			default:
+				progress.Update(progressOutput, layer.String(), "Waiting")
+				<-start
+			}
+
+			if parentDownload != nil {
+				// bail if parent download failed or was cancelled
+				select {
+				case <-parentDownload.Done():
+					err := parentDownload.result()
```
caglar10ur: could be a single if ...; err!= nil { line

lib/imagec/download.go
```diff
@@ -0,0 +1,330 @@
...
+			if parentDownload != nil {
+				select {
+				case <-d.Transfer.Context().Done():
+					d.err = errors.New("layer download cancelled")
+					return
+				case <-parentDownload.Done(): // block until parent download completes
+				}
+
+				err := parentDownload.result()
+				if err != nil {
```
caglar10ur: could be a single if ...; err!= nil { line

lib/imagec/download.go
```diff
@@ -0,0 +1,330 @@
...
+					return
+				}
+			}
+
+			// is this the child-most leaf layer?
+			imageLayer := layer.ID == layers[0].ID
+
+			// if this is the child-most layer, we are done and can now create the image config
+			if imageLayer {
+				imageConfig, err := ic.CreateImageConfig(layers)
```
caglar10ur: could be a single if ...; err!= nil { line

lib/imagec/download.go
```diff
@@ -0,0 +1,330 @@
...
+
+			if parentDownload != nil {
+				select {
+				case <-d.Transfer.Context().Done():
+					d.err = errors.New("layer download cancelled")
+					return
+				case <-parentDownload.Done(): // wait for parent layer download to complete
+				}
+
+				err := parentDownload.result()
```
caglar10ur: could be a single if ...; err!= nil { line

lib/imagec/download.go
```diff
@@ -0,0 +1,330 @@
...
+			// parentDownload finished, but wait for it explicitly
+			// to be sure.
+			select {
+			case <-d.Transfer.Context().Done():
+				d.err = errors.New("layer download cancelled")
+				return
+			case <-sourceDownload.Done():
+			}
+
+			err := sourceDownload.result()
```
caglar10ur: could be a single if ...; err!= nil { line

lib/imagec/imagec.go
```diff
@@ -0,0 +1,623 @@
...
+				Parent: &parent,
+				Store:  ic.Storename,
+			},
+			meta:   history.V1Compatibility,
+			layer:  layer,
+			diffID: "",
+		}
+	}
+
+	/*
```
caglar10ur: I think you can nuke this part

lib/apiservers/engine/backends/cache/layer_cache.go
fdawg4l: @jzt, FYI the list of completed images are already cached in the storage layer, so asking the PL on-demand rather than caching that info here as well wouldn't be expensive.

lib/imagec/download.go
jzt: I had other stuff inside the `func()` in addition to the `close()` that I removed.. just forgot to cleanup after :)

lib/apiservers/engine/backends/cache/layer_cache.go
jzt: Good points, guys. I'm going to look into @caglar10ur's suggestion. @fdawg4l, that's good to know. I'm going to need to do that anyway, so that I can get the diffIDs for the existing layers in order to fix #2530.

More on this to follow.

lib/imagec/download.go
jzt: Thanks, removed

lib/imagec/download.go
jzt: Thanks for this, I had no idea but it makes so much sense. :)

lib/imagec/download.go
jzt: @hmahmood and I were asking ourselves the same question and it comes down to implementing some of Docker's interfaces. I believe we can port our download code over to these interface implementations, but there are a few things I'm not sure of, like where the line is between our layers and theirs (they call a lot of functions that are defined on docker layers that we don't have). I will look for a clear line of separation in order to make this possible. It would go a long way to making things a lot cleaner and easier to understand.

lib/imagec/download.go
jzt: Great catch, fixed!

lib/imagec/download.go
jzt: Fixed

lib/imagec/download.go
jzt: Fixed all of these.

lib/imagec/imagec.go
jzt: 💥  nuked! 💥

lib/imagec/docker.go
jzt: Good idea. I will clean this up and add ctx to the Learn* funcs.

lib/apiservers/engine/backends/cache/layer_cache.go
```diff
@@ -0,0 +1,97 @@
...
+
+// LayerNotFoundError is returned when a layer does not exist in the cache
+type LayerNotFoundError struct{}
+
+func (e LayerNotFoundError) Error() string {
+	return "Layer does not exist"
+}
+
+var (
+	layerCache *LCache
```
vburenin: if there is no complex logic in init, you really can use a one-liner here as:

var layerCache = &LCache{layers: make(map[string]bool)}

lib/apiservers/engine/backends/cache/layer_cache.go
```diff
@@ -0,0 +1,97 @@
...
+
+// LayerCache returns a reference to the layer cache
+func LayerCache() *LCache {
+	return layerCache
+}
+
+// AddExisting adds an existing layer to the cache
+func (lc *LCache) AddExisting(id string) {
+	lc.m.Lock()
+	defer lc.m.Unlock()
```
vburenin: FYI using defer in such trivial cases like this is just an overkill. It is relatively expensive operation.

lib/imagec/download.go
```diff
@@ -0,0 +1,322 @@
...
+
+		go func() {
+			defer close(progressChan)
+
+			<-start
+
+			close(inactive)
+
+			if parentDownload != nil {
+				select {
```
vburenin: This is repetitive logic. Could you extract it into a separate function?

lib/apiservers/engine/backends/cache/layer_cache.go
jzt: @caglar10ur I agree that getting rid of the cache would be a good thing, but if we use the fs semantics, I don't think that would allow another client to detect that a layer has already been downloaded. Example:

Client A pulls tomcat. He gets the lock and starts putting all the layers in flight. The lock releases and the layers are downloading.

Client B comes along and pulls the same image. He gets the lock, calls out to the PL and learns what images he needs to download (say all of them, since none have been written to the PL yet). At this point, the file locks will indicate which layers are in flight (good).

But what if Client A completes downloading some particular layer, writes to the PL and releases the file lock before Client B attempts to obtain the file lock?

Client B is not aware of this, obtains the file lock and repeats the download for that layer, eventually getting an "Already exists" error when attempting to write to the PL. Using the cache allows for Client B to know that a layer has already been downloaded. The existence of the layer in the cache is the indicating factor that Client B should not attempt a download, but the state (true/false) of the cached layer also indicates whether or not Client B should track that layer's progress during a download. Is there a way to accomplish this using the fs semantics?

caglar10ur: Client A holds the lock while downloading the layer, during that time Client B simply waits. When A finished it releases and B grabs it, asks PL, learns that layer is already there, drops the lock. 

No?

lib/imagec/download.go
jzt: @caglar10ur following up, this looks like it will be pretty time consuming. Would you be ok with punting on this for now?

lib/apiservers/engine/backends/cache/layer_cache.go
jzt: @caglar10ur then Client B would not be able to display/share the progress output for that particular layer. Instead it would be stuck at `Waiting` (I think) and then suddenly `Already exists`.

lib/imagec/docker.go
jzt: It looks like the Learn* functions are already using a context timeout by way of `fetcher.Head`.

FetchImageBlob is getting the context that the `Transfer` provides, so I wrap that in a context using our default timeout. For everything else, I've created a context using our default timeout in `PullImage` and I send that in to all the `Fetch*` functions from there. Sound ok?

lib/apiservers/engine/backends/cache/layer_cache.go
jzt: Good to know. I removed the defers and just unlock after map modification.

caglar10ur: IMHO let's not prematurely optimize the code unless we have a proof that defer's are hurting us. At this point micro-optimizations should be the last one on our to-do list

caglar10ur: And also let me remind you that, defer's are protecting us from programming errors like https://github.com/vmware/vic/pull/2587#discussion_r82254091 
<insert your favorite my 2 cents gif here> :)

lib/apiservers/engine/backends/cache/layer_cache.go
jzt: Fixed

lib/apiservers/engine/backends/cache/layer_cache.go
jzt: Fixed

jzt:
@caglar10ur @vburenin so I've gone ahead and skipped some of the "nice to haves" with regard to code formatting/cleanup. I'm more concerned with stability right now. At this point, it looks like it does just fine on VC (skinner) and I don't see anymore EOFs on ESX. I don't know that they're gone for good, but we can open issues for them and fix them as they come. This needs to get in today, and I can bring in some other fixes once that happens.

The weird behavior I see now is when I'm pulling with 3 or 4 clients simultaneously - sometimes one of those clients will get the `invalid byte in chunked data` message.

lib/apiservers/engine/backends/cache/layer_cache.go
caglar10ur: Yeah, but isn't what docker does? Simply waiting and showing a message like "X already being pulled by another client. Waiting." ?

caglar10ur: Hmm looks like that's the old behavior and it changed loooong time ago https://github.com/docker/docker/pull/18353

jzt: No it will actually attach that client to the progress output of the one who is downloading:

https://asciinema.org/a/7nxc3bvqf68xxn6t6ftji1gpr

cmd/imagec/imagec_test.go
```diff
@@ -401,125 +40,3 @@ func TestNewHook(t *testing.T) {
 		t.Errorf("Fatal test failed %s", string(b.Bytes()))
```
sgairo: More of a question than a comment: why use Errorf instead of Fatalf ?

caglar10ur:
LGTM

lib/apiservers/engine/backends/cache/layer_cache.go
```diff
@@ -0,0 +1,91 @@
...
+import (
+	"sync"
+)
+
+// LCache is an in-memory cache to account for existing image layers
+// It is used primarily by imagec when coordinating layer downloads
+// The cache is initially hydrated by way of the image cache at startup
+type LCache struct {
+	m sync.RWMutex
+
```
matthewavery: remove empty line in order to create a Mutex Hat

cmd/imagec/imagec.go
```diff
@@ -15,119 +15,32 @@
...
-
-	// DefaultTokenExpirationDuration specifies the default token expiration
-	DefaultTokenExpirationDuration = 60 * time.Second
-
-	// attribute update actions
-	Add = iota + 1
-	Remove
 )
 
 func init() {
```
hmahmood: Is this needed anymore?

lib/apiservers/engine/backends/image.go
```diff
@@ -145,63 +147,51 @@ func (i *Image) ExportImage(names []string, outStream io.Writer) error {
...
+			options.InsecureAllowHTTP = true
 			break
 		}
 	}
 
-	log.Debugf("PullImage: cmd = %s %+v\n", Imagec, cmdArgs)
-
-	cmd := exec.Command(Imagec, cmdArgs...)
-	cmd.Stdout = outStream
-	cmd.Stderr = outStream
```
hmahmood: I don't see an equivalent `options.StdErr`

lib/imagec/download.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+
+	// iterate backwards through layers to download
+	for i := len(layers) - 1; i >= 0; i-- {
+
+		imageLayer := layers[i]
+		id := imageLayer.ID
+
+		downloading, err := cache.LayerCache().IsDownloading(id)
+		if err != nil { // LayerNotFoundError
```
hmahmood: I would assert the error type here

cmd/imagec/imagec.go
jzt: Yeah, since we can still run imagec in standalone mode.

lib/imagec/download.go
jzt: Done

lib/imagec/download.go
fdawg4l: @caglar10ur  If the update was suppose to happen in the storage layer on the image store, then it was never a supported usecase.  

@jzt  you can probably nuke the comment.

hmahmood:
lgtm

jzt:
♻️

jzt:
lgtm?

PR #2608, opened 2016-10-07
Add more debugging messages to the tether startup

gigawhitlocks:
Add even more debugging info to keep tracking down #2503 -- we saw a failure to write to stderr and the error message was suppressed -- would've liked to have had that output 😦 
https://ci.vmware.run/vmware/vic/5494


anchal-agrawal:
:shipit:

sgairo:
LGTM

PR #2609, opened 2016-10-07
Collecting container logs for functional tests [skip ci]

rajanashok:
Gathering container logs from Nimbus.
Extracting the bintray files to bin folder instead of vic.
Disabling the private registry tests on Nimbus.


anchal-agrawal:
:shipit:

dougm:
lgtm

hickeng:
I noticed a lot of the manual tests are broken since: https://github.com/vmware/vic/commit/95d26214c232cf27765be954dd7fb90e032faceb

This is because the parameters to `Install VIC Appliance To Test Server` were changed significantly. They now have default values, and you can set only the specific argument you're after via `param=${value}`.
I mention this because a lot of them are passing `${false}` for `certs` - this can either be changed to `certs=${false}` or you can try leaving certs as true, which is what the rest of the build now does, barring compose.

rajanashok:
rebased.

PR #2604, opened 2016-10-07
Renames container logs and simplifies the fetch code in vicadmin

matthewavery:
Renames the log files that are output by the container and the tether. This results in a much simpler approach to fetching container logs in `vicadmin`. It also should lead to much less brittle log handling when it comes to container logs. 

Fixes #2562 
Fixes #2467

PR #2611, opened 2016-10-08
Restore collection of vmware.log

hickeng:
Restores the collection of vmware.log for the VMs and drops
collecting the containerVM output, which is purely customer application
data. Also renames misnamed variables.
Original changes were in #2604 

Fixes #2604

### DIFF
--- a/cmd/vicadmin/vicadm.go
+++ b/cmd/vicadmin/vicadm.go
@@ -63,7 +63,7 @@ var (
 
 	// VMFiles is the set of files to collect per VM associated with the VCH
 	vmFiles = []string{
-		"output.log",
+		"vmware.log",
 		"tether.debug",
 	}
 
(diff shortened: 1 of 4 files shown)


dougm:
lgtm

matthewavery:
Would this change also affect the target of the `docker logs` that was changed in #2604

matthewavery:
@hickeng  yes, looks like I misunderstood one of the logs. the `docker logs` target looks to now be collecting(or trying to) the `output.log` file that I renamed... instead of the <containerID>.log . Would it be worth it to rename the actual output log to `output.log`?

hickeng:
@matthewavery the actual output log _is still_ output.log which is what `docker logs` uses.
It's just not the log we wanted, which was vmware.log, which contains the vSphere level diagnostics.

matthewavery:
In that case LGTM

hickeng:
poke

matthewavery:
LGTM

Issue #2491, opened 2016-09-27
Tether not able to establish connection causes orphaned container VMs (resource leak)


sflxn:
I have found what I believe is a serious issue.  Sometime, when performing 'docker run -it ...', the tether fails to establish a connection, causing the attach to timeout.  VIC, subsequently "forgets" the existence of the container, yet the container is still running, utilizing CPU and memory space.  However, if we then do docker ps, it sees it, but we cannot inspect it.
1. docker run -it busybox /bin/sh
2. outputs: 'no such container'
3. docker ps
4. outputs: container is in the list
5. docker inspect containerID
6. outputs: 'no such image'

sflxn:
[container-logs (1).zip](https://github.com/vmware/vic/files/496330/container-logs.1.zip)
Container log bundle.

cgtexmex:
Isn't this a case of it being removed from the persona docker cache?  Thus the 'no such container' errors?

sflxn:
@cgtexmex I'm not sure.  Investigating the code and looking for that removal, but I am not seeing it.

hickeng:
A lossiness check is still present in the handshake code (artifact from tracking down a bug in the go term package). This violates the client as initiator rule in the handshake code.

This can be taken out - I'd prefer it if the code remains, but blocked out with either constant or dynamic check for debug (dynamic would be the debug setting for the VHC in the config). If dynamic it should be when debug > 1 as it's a change to behaviour:

const option:
```
const debug = false
if debug {
   ....
}
```

pkg/serial/handshake.go

chengwang86:
Is this similar to #1821  no such image error on docker run ?

mdubya66:
Marking as a bug and adding to the GA Milestone.

chengwang86:
I observed an interesting thing:

After I create a fresh new container, I do "docker start." 
After "**ContainerStart**" returns and before "**Established connection with container VM: container_ID**" is done, if I do "docker attach," the cmd would 

(i) hang for several sec and succeed finally, 
or 
(ii) return with "**Error: No such container container_ID**". 

For (ii), if I do "docker ps -a", the status of the container is "running". But if I do "inspect", it would not be able to find the container. However, **(ii) is not always reproducible**. 

(ii) is quite similar to your issue IMO. I guess it might be because that neither "docker start"(ContainerStart) nor "docker attach"(ContainerAttach) wait for the connection to the Tether process to be established. 

[...]

PR #2612, opened 2016-10-08
Lossiness check in attach is optional

hickeng:
The lossiness check in attach has a flag controlling it. It's also been
inverted to abide by the requirement that the client always initiates
interactions - this rule avoids the scenario of both sides waiting for
input.
The debug value of the appliance isn't yet plumbed through to this because
the attach connector is started in the interaction handler, not in the
main PL implementation at this time.

Towards #2491

@fdawg4l @caglar10ur for primary review


dougm:
lgtm

cmd/tether/attach_test.go
```diff
@@ -141,7 +141,7 @@ func mockBackChannel(ctx context.Context) (net.Conn, error) {
 			// FIXME: need to implement timeout of purging hangs with no content
 			// on the pipe
 			// serial.PurgeIncoming(ctx, conn)
-			err := serial.HandshakeClient(ctx, conn)
+			err := serial.HandshakeClient(ctx, conn, true)
```
caglar10ur: [cosmetic] how about adding a new one named HandshakeClientWithCheck and calling it when needed instead of adding another parameter to the function?

pkg/serial/handshake.go
```diff
@@ -100,45 +101,57 @@ func HandshakeClient(ctx context.Context, conn net.Conn) error {
...
+		// of data to the container and it will echo it back.  Verify the sent and
+		// received bufs are the same and we know the channel is lossless.
 
-	if n != len(rxbuf) {
-		err = fmt.Errorf("packet size mismatch (expected %d, received %d)", len(rxbuf), n)
-		log.Error(err)
-		return err
-	}
+		log.Debugf("Checking for lossiness")
+		txbuf := []byte("\x1b[32mhello world\x1b[39m!\n")
```
caglar10ur: we should send ```because containers``` or ```deal with it``` instead of this /s :)

caglar10ur:
LGTM :sheep: :it:

hickeng:
poke

PR #2613, opened 2016-10-08
distclean make target and correct extraconfig processessing

hickeng:
Misc set of convenience/correctness changes:
Adds a distclean make target to remove cached files (yum package cache, iso base and the iso staging files).
Modifies the clean target to delete everything except the cache files to speed clean builds.
Correctly uses extraconfig, replacing "<nil>" with "" in the DataSource


mdubya66:
lgtm

dougm:
lgtm

hickeng:
poke

PR #2618, opened 2016-10-09
Implements image volumes

### LINKED ISSUE #1819: Create volumes from VOLUME command in image metadata
Images can include a VOLUME which gets created when then container is started.  They're anonymous volumes and the image expects them to exist at the mount point specified AND include the data which  may already exist in that folder.

matthewavery:
This implements the creation of anonymous volumes from image metadata(a good example is the `mongo` image). It is important to note that a specified volume takes precedent over an anonymous volume. This includes the anonymous volumes from images. This also does not implement the copying of existing data from a mount path in the image, that will come from a later PR. 

Fixes #1819

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -1474,6 +1474,17 @@ func setCreateConfigOptions(config, imageConfig *containertypes.Config) {
 		config.Entrypoint = imageConfig.Entrypoint
 	}
 
+	if config.Volumes == nil {
+		config.Volumes = imageConfig.Volumes
+	} else {
+		for k, v := range imageConfig.Volumes {
+			//NOTE: the value of the map is an empty struct.
+			//      we also do not care about duplicates.
+			//      This Volumes map is really a Set.
+			config.Volumes[k] = v
+		}
+	}
+
 	// set up environment
 	setEnvFromImageConfig(config, imageConfig)
 }
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -745,7 +745,7 @@ func finalizeVolumeList(specifiedVolumes, anonymousVolumes []string) ([]volumeFi
 	}
 
 	finalizedVolumes := make([]volumeFields, 0, len(processedAnonVolumes))
-	for _, v := range processedVolumes {
+	for _, v := range processedAnonVolumes {
 		finalizedVolumes = append(finalizedVolumes, v)
 	}
 	return finalizedVolumes, nil
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -226,30 +226,34 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
 
 	// Collect all volume mappings. In a docker create/run, they
 	// can be anonymous (-v /dir) or specific (-v vol-name:/dir).
-	volumes := config.HostConfig.Binds
-	for anonVol := range config.Config.Volumes {
-		volumes = append(volumes, anonVol)
+	// anonymous volumes can also come from Image Metadata
+
+	rawAnonVolumes := make([]string, 0, len(config.Config.Volumes))
+	for k := range config.Config.Volumes {
+		rawAnonVolumes = append(rawAnonVolumes, k)
 	}
 
-	volList, err := processVolumeFields(volumes)
+	volList, err := finalizeVolumeList(config.HostConfig.Binds, rawAnonVolumes)
 	if err != nil {
 		return handle, BadRequestError(err.Error())
 	}
+
+	log.Infof("Finalized Volume list : %#v", volList)
 
 	// Create and join volumes.
 	for _, fields := range volList {
 
 		//we only set these here for volumes made on a docker create
-		driverArgs := make(map[string]string)
-		driverArgs[DriverArgFlagKey] = fields.Flags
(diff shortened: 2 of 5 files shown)


caglar10ur:
LGTM :sheep: :it:

tests/test-cases/Group1-Docker-Commands/1-19-Docker-Volume-Create.robot
```diff
@@ -8,16 +8,20 @@ Suite Teardown  Cleanup VIC Appliance On Test Server
 Simple docker volume create
     ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} volume create
     Should Be Equal As Integers  ${rc}  0
-    ${disk-size}=  Run  docker ${params} logs $(docker ${params} start $(docker ${params} create -v ${output}:/mydata busybox /bin/df -Ph) && sleep 10) | grep by-label | awk '{print $2}' 
+    ${disk-size}=  Run  docker ${params} logs $(docker ${params} start $(docker ${params} create -v ${output}:/mydata busybox /bin/df -Ph) && sleep 10) | grep by-label | awk '{print $2}'
```
hickeng: you could use `docker run -d` rather than chaining create & start like this

hickeng: there is also now `docker wait` to wait for it to stop.
This is your own VCH env, so you can also supply `--name=volumetest` for example and split these onto multiple lines. It will make it both easier to read in the test and easier to see what happened if one of the commands fails.

tests/test-cases/Group1-Docker-Commands/1-19-Docker-Volume-Create.robot
```diff
@@ -8,16 +8,20 @@ Suite Teardown  Cleanup VIC Appliance On Test Server
...
 Docker volume create named volume
     ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} volume create --name=test
     Should Be Equal As Integers  ${rc}  0
     Should Be Equal As Strings  ${output}  test
-    ${disk-size}=  Run  docker ${params} logs $(docker ${params} start $(docker ${params} create -v ${output}:/mydata busybox /bin/df -Ph) && sleep 10) | grep by-label | awk '{print $2}' 
+    ${disk-size}=  Run  docker ${params} logs $(docker ${params} start $(docker ${params} create -v ${output}:/mydata busybox /bin/df -Ph) && sleep 10) | grep by-label | awk '{print $2}'
     Should Be Equal As Strings  ${disk-size}  975.9M
 
+Docker volume create image volume
+       ${disk-size}=  Run  docker ${params} logs $(docker ${params} start $(docker ${params} create mongo /bin/df -Ph ) && sleep 10) | grep by-label | awk '{print $2}'
```
hickeng: Purely from an efficiency standpoint you can likely find a smaller container that has a VOLUME statement.

tests/test-cases/Group1-Docker-Commands/1-19-Docker-Volume-Create.robot
```diff
@@ -8,16 +8,20 @@ Suite Teardown  Cleanup VIC Appliance On Test Server
...
+       Should Contain  ${disk-size}  976M
+
```
hickeng: I do not see a test for a user specified anonymous volume:
`docker create -v /path/in/container busybox ...`

lib/apiservers/engine/backends/container.go
```diff
@@ -1436,6 +1436,17 @@ func setCreateConfigOptions(config, imageConfig *containertypes.Config) {
...
 	}
 
+	if config.Volumes == nil {
+		config.Volumes = imageConfig.Volumes
+	} else {
+		for k, v := range imageConfig.Volumes {
+			//NOTE: the value of the map is an empty struct.
+			//      we also do not care about duplicates.
+			//      This Volumes map is really a Set.
+			config.Volumes[k] = v
```
matthewavery: Yes that data is stored in the HostConfig.Binds, I believe this is because docker only approaches volumes from a mostly `bind mount` perspective. So the "specified volumes" are processed separately from the anonymous volumes. and then a final set is distilled from that taking the specified volumes over the anonymous ones. The image volumes actually come from the images `config.volumes` set. so those must be joined before the actual fields are processed.

tests/test-cases/Group1-Docker-Commands/1-19-Docker-Volume-Create.robot
matthewavery: Mike and I decided to put that case in the `create` tests. right here:

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -740,13 +740,11 @@ func finalizeVolumeList(specifiedVolumes, anonymousVolumes []string) ([]volumeFi
...
-		if _, ok := processedVolumes[k]; !ok {
-			processedVolumes[k] = v
-		}
+	for k, v := range processedVolumes {
+		processedAnonVolumes[k] = v
 	}
 
-	finalizedVolumes := make([]volumeFields, 0, len(processedVolumes))
+	finalizedVolumes := make([]volumeFields, 0, len(processedAnonVolumes))
 	for _, v := range processedVolumes {
```
hickeng: need to iterate over `processedAnonVolumes`

matthewavery: This has been fixed now, you will have to mark it resoled @hickeng

hickeng:
LGTM - please do add the command line anonymous volume however.

matthewavery:
♻️

Issue #2455, opened 2016-09-23
Tether might take up too much CPU in Ubuntu


chengwang86:
If I start a container from the vanilla **Ubuntu** image, sometimes the CPU utilization of the tether process could be quite high (see example below). However, this does not happen on a container started from the **busybox** image.

**Expected behavior:**(a snippet from the output of the linux "top" command on the busybox container)

  PID | PPID | USER | STAT | VSZ | %VSZ | CPU | %CPU | COMMAND
------------ | ------------- | ------------ | ------------- | ------------ | ------------- | ------------ | ------------- | ------------
    21 | 2 | root | SW | 0 | 0.0 | 0 | 1.9 | [kworker/0:1]
    1 | 0 | root | S | 340m | 16.9 | 0 | 1.3 | /.tether/tether
208 | 1 | root | S | 1188 | 0.0 | 0 | 0.0 | sh

  PID  PPID USER     STAT   VSZ %VSZ CPU %CPU COMMAND
   21     2 root     SW       0  0.0   0  1.9 [kworker/0:1]
    1     0 root     S     340m 16.9   0  1.3 /.tether/tether
  208     1 root     S     1188  0.0   0  0.0 sh

**Actual behavior**: (a snippet from the output of the linux "top" command on the Ubuntu container)

  PID | USER | PR | NI | VIRT | RES | SHR | S | %CPU | %MEM  | TIME+ | COMMAND
------------ | ------------- | ------------ | ------------- | ------------ | ------------- | ------------ | ------------- | ------------ | ------------- | ------------ | -------------
    4 | root | 20 | 0 | 0 | 0 | 0 | R | 71.7 | 0.0 | 1:09.17 | kworker/0:0
    1 | root | 20 | 0 | 349884 | 13688 | 4736 | S | _**53.3**_  | 0.7 | 1:12.54 | **_tether_**
244 | root | 20 | 0 | 0 | 0 | 0 | S | 0.3 | 0.0 | 0:00.08 | kworker/1:1

mdubya66:
Needs to be triaged for GA. May be Iceboxed depending on the result

chengwang86:
This is not needed for GA IMO.

hickeng:
This is alarmingly high! @chengwang86 do you know if the same happens when not attached? i.e. when the tether is not performing significant amounts of copying, etc to handle output.

@caglar10ur if you've got pprof running in the tether in a neat fashion with -debug then this could be a good use for it.

[...]

PR #2621, opened 2016-10-09
Propagate debug/pprof config into containers

hickeng:
We've previously had pprof servers running in all of the components on port 6060 however we cannot have those running and exposed by default when we ship.

This PR propagates the debug flag into the containerVMs and uses that to configure logging and pprof.

Towards #2455


caglar10ur:
LGTM :sheep: :it:

fdawg4l:
LGTM

emlin:
lgtm 👍 
This make the debug process complete.

hickeng:
poke

PR #2623, opened 2016-10-09
Change error msg on resize failure

sflxn:
Change error message on resize 404

Fixed an issue where the personality server respond with no such
container when the portlayer respond with 404 for a resize.  The
404 actually refers to the connection and not the container.

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -359,7 +359,7 @@ func (c *Container) ContainerResize(name string, height, width int) error {
 	_, err := client.Interaction.ContainerResize(plResizeParam)
 	if err != nil {
 		if _, isa := err.(*interaction.ContainerResizeNotFound); isa {
-			return NotFoundError(name)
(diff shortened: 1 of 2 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -359,8 +359,7 @@ func (c *Container) ContainerResize(name string, height, width int) error {
 	_, err := client.Interaction.ContainerResize(plResizeParam)
 	if err != nil {
 		if _, isa := err.(*interaction.ContainerResizeNotFound); isa {
-			cache.ContainerCache().DeleteContainer(id)
```
caglar10ur: This is already gone in the master branch

caglar10ur:
LGTM after you rebase :)

lib/apiservers/engine/backends/container.go
sflxn: See that you did.  Modifying the PR.  Still good to have a different error message than No such container, which is incorrect.

sflxn:
ping

Issue #2197, opened 2016-08-30
Add user authentication to the personality


hickeng:
The default docker mechanism is TLSVERIFY. This would entail:
a. modification to vic-machine to supply CA certificate to validate clients

There is also the extended docker authorization plugin mechanism which presumably requires that the client send user credentials (or session identifier) with each request. I do not know the details of this mechanism and a quick search of the remote API 1.24 docs didn't make it clear.

Tagged as vic-machine because any approach will require vic-machine changes for configuration. Estimate is for TLSVERIFY - see vicadmin for current mechanism.

https://bugzilla.eng.vmware.com/show_bug.cgi?id=1727625

mdubya66:
Moving to Icebox. GA is pretty packed.

hickeng:
Tooks significantly longer than expected primarily due to:
* requirement for DNS resolution of VCH or static IP
* updating tests to function correctly in a mix of environments where tlsverify is or isn't viable

hickeng:
Moving back to in-progress for the following refinement:
* do not overwrite certs unless --force
* generate pkf certificate if openssl is available
* if certs exist in vch-name/... then see if they permit --cname (if specified), and use them. If not fall back to generate path (with the behaviour above) and log warning about certificate reuse

The first avoids mistakenly destroying your CA and invalidating all client certs.
The second makes the initial portion of #2342 much more usable as there's a cert that can be directly imported into browsers.

mdubya66:
Moving to To Do. Not originally Dev Complete

andrewtchin:
Looks like when I install a VCH with existing certificates (generated previously by a VCH install) the installer doesn't have a client cert to poll the status of services so the install times out. The VCH though seems to work fine so this should just be a vic-machine issue.
```bin/vic-machine-linux create --target root:password@192.168.218.129 --name chin-vic-11 --tls-cname 192.168.218.200 --external-network-ip 192.168.218.200/24 --external-network-gateway 192.168.218.2/24 --thumbprint 50:13:2A:8F:4D:9F:2C:69:AC:F2:47:38:AF:64:93:F2:CE:91:56:5D --tls-ca="./chin-vic-11/ca.pem" --cert="./chin-vic-11/server-cert.pem" --key="./chin-vic-11/server-key.pem" --debug 1```
```
[...]

Issue #2624, opened 2016-10-09
Upgrade does not wait for components to start


hickeng:
**Expected behavior:**
After upgrading vic-machine blocks until the appliance has successfully launched, and rolls back if that doesn't happen in a reasonable time frame.

**Actual behavior:**
vic-machine doesn't block until the component have launched or validate function.

** Diagnosis**
This is a combination of two things:
* adding client authentication means that unless a client cert is supplied it's not always possible to access the DOCKER_API endpoint to confirm function. This resulted in the endpoint check moving out of the standard startAppliance path and into the callers that need it.
* session.started is persisted across reboots, so we don't block waiting for that either as they would be marked as successful from the previous initialization.

We have possible parts of solutions :
1. allow the host certificate to be used for client verification - that means we *always* have a cert available that can be used for the client auth if we have access to the config. This requires not masking the host private key from the API however.
2. make `session.started` non-persistent, and/or clear it on the shutdown path. Doing this would mean we need to split into `status` and `state` so that we can persist `status` which could contain errors such as `no such file` vs state which is contingent on the VM running.

PR #2629, opened 2016-10-09
Follow up fixes from TLSVERIFY work

hickeng:
Block vic-machine upgrade from completing until the components
have come up cleanly. This does **not** add rollback if the VCH fails
to initialize correctly.

Only serves the docker API on the client interface as designed and
removes some of the unnecessary flags.

Updates to examples and usage doc.

Towards #2624 (handles the --no-tls case)
Fixes: #2197

### DIFF
--- a/cmd/docker/main.go
+++ b/cmd/docker/main.go
@@ -19,6 +19,7 @@ import (
 	"crypto/x509"
 	"flag"
 	"fmt"
+	"net"
 	"os"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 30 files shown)


hickeng:
This is now rather overloaded and in need of me cherry picking into different PRs.

cmd/docker/main.go
```diff
@@ -222,14 +191,38 @@ func startServerWithOptions(cli *CliOptions) *apiserver.Server {
 		}
 	}
 
+	addr := "0.0.0.0"
+	// exposing this on all interfaces
+	if vchConfig.Diagnostics.DebugLevel <= 2 {
```
andrewtchin: Why does the lookup depend on the debug level?

cmd/vic-init/ops.go
```diff
@@ -69,6 +70,9 @@ func (t *operations) HandleSessionExit(config *tether.ExecutorConfig, session *t
 			return
 		}
 
+		// incredibly basic throttle
```
hmahmood: Please tag this as a TODO. This indicates that we have issues around synchronization; we should not be putting in random sleeps.

hmahmood: Actually why do we have this? I don't think this is a good idea.

hickeng:
Split this into multiple PRs:
#2703 
#2706 
#2707 

and there'll be another one for limiting the servers to listening on appropriate interfaces & #2632

PR #2631, opened 2016-10-09
Optimise make target dependency generation

hickeng:
I've been frustrated with `clean` and `distclean` taking a while to run, plus the rest taking longer than it should.

This adds Gnu Make Standard Library (https://sourceforge.net/projects/gmsl/) to the infra directory and uses it to cache outputs. @mdubya66 this is a BSD license, but the project doesn't package an actual license file as best I can determine.

Also skips dependency generation for the cleaning targets.


hmahmood:
lgtm

caglar10ur:
LGTM :sheep: :it:

hickeng:
poke

PR #2642, opened 2016-10-10
Walkthrough the golint warnings. make check, make codelint - run the …

vburenin:
Fixes #2583

…golint.


pkg/serial/handshake.go
```diff
@@ -32,10 +32,10 @@ import (
 )
 
 const (
-	SYN       = 0x16
-	ACK       = 0x06
-	DEBUG_ACK = 0x07
-	NAK       = 0x15
+	CodeSync     = 0x16
```
fdawg4l: What was the lint warning?  I'd rather not have `Code` in const name.  And `Syn` is probably fine.

Makefile
```diff
@@ -127,7 +127,7 @@ gas: $(GAS)
 # convenience targets
 all: components tethers isos vic-machine vic-ui
 tools: $(GOIMPORTS) $(GVT) $(GOLINT) $(SWAGGER) $(GAS) goversion
-check: goversion goimports govet golint copyright whitespace gas
+check: goversion goimports govet lintcode copyright whitespace gas
```
fdawg4l: Is the `lintcode` binary in the standard package?  If not, it'll need to be added to vendor, and appropriate `install` command added to the makefile.

fdawg4l: Why are we renaming the target?

vburenin: golint - is only go specific linter. lintcode may be applied for anything else, like python.

pkg/serial/handshake.go
vburenin: Capital letter for constants and underscores. It actually was only about DEBUG_ACK. If you can think of a better name for DEBUG_ACK with no underscore I can roll changes back.

dougm: Doesn't look like these constants are referenced outside the package, so let's not export them either (syn, ack, debugAck, nak).  Or if we want to prefix, maybe use 'flag' or 'seq'?  seqSyn, seqAck, etc

Makefile
fdawg4l: #neverpython

fdawg4l: More seriously, we will likely never use this repo for more than `go` (and possibly some `javascript`).  And in the case where we need to call more specific linters, we'll add targets specifically for those linters.

pkg/serial/handshake.go
vburenin: @dougm Sounds reasonable.

Makefile
vburenin: btw, target golint doesn't work for some reason. I am not an expert in Makefile(s), I would be really happy if somebody gives me a valuable advice why it is not working.

dougm: You mean if you name 'lintcode' to 'golint'?  Looks like you'd need to remove line 123 `golint: $(GOLINT)`.  Not sure why we have those set of targets is the first place (lines 122-125).  The `gas` target is overridden later too.  For example, 'make check' properly runs 'golint' again with the change below.  I also don't understand what `$(go-lint)` as a target was supposed to do as it isn't defined anywhere.

```diff
iff --git a/Makefile b/Makefile
index fb0a1fd..bcd1524 100644
--- a/Makefile
+++ b/Makefile
@@ -120,7 +120,6 @@ vic-ui: $(vic-ui-linux) $(vic-ui-windows) $(vic-ui-darwin)
 vic-dns: $(vic-dns-linux) $(vic-dns-windows) $(vic-dns-darwin)

 swagger: $(SWAGGER)
-golint: $(GOLINT)
 goimports: $(GOIMPORTS)
 gas: $(GAS)

@@ -175,7 +174,7 @@ whitespace:
 # exit 1 if golint complains about anything other than comments
 golintf = $(GOLINT) $(1) | sh -c "! grep -v 'should have comment'" | sh -c "! grep -v 'comment on exported'"

-$(go-lint): $(GOLINT)
+golint: $(GOLINT)
        @echo checking go lint...
        @$(call golintf,github.com/vmware/vic/cmd/...)
        @$(call golintf,github.com/vmware/vic/pkg/...)
```

vburenin: thanks. original golint seemed too confusing.
We might need to get some focus on Makefile and clean it up. It takes so long to run even trivial things. Especially if it is a mounted network disk. As well as weird overrides.

dougm:
lgtm

caglar10ur:
LGTM :sheep: :it:

PR #2652, opened 2016-10-11
Delete vmdk first before delete directory for vc 6.5 or newer

### LINKED ISSUE #2014: [6.5] vic-machine fails to remove
```
INFO[2016-08-16T21:15:51Z] ### Removing VCH ####
adsf;lkINFO[2016-08-16T21:15:54Z]
INFO[2016-08-16T21:15:54Z] VCH ID: VirtualMachine:4
jINFO[2016-08-16T21:15:54Z] Removing VMs
INFO[2016-08-16T21:15:59Z] Removing volume stores...
INFO[2016-08-16T21:15:59Z] Deleting volume store "default" on Datastore "datastore1" at path "."
ERRO[2016-08-16T21:16:00Z] Operation failed: Error caused by file /vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7 - (&{{} %!!(MISSING)s(*types.FileFault=&{{{<nil> []}} /vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7}) Error caused by file /vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7})
INFO[2016-08-16T21:16:00Z] Removing appliance VM network devices
INFO[2016-08-16T21:16:01Z] No network device attached
INFO[2016-08-16T21:16:01Z] Removing Portgroup "ZzZ"
INFO[2016-08-16T21:16:02Z] Removing VirtualSwitch "ZzZ"
ERRO[2016-08-16T21:16:02Z] Operation failed: Failed to delete the virtual machine, '/vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx'. - (&{{} %!!(MISSING)s(*types.GenericVmConfigFault=&{{{{<nil> [{{} msg.vigor.offline.deleteVM [{{} 1 /vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx}] Failed to delete the virtual machine, '/vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx'.} {{} msg.vigor.method.invalidConfig [] The virtual machine configuration is invalid.} {{} msg.vigor.vmx.load.error [{{} 1 /vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx}] Unable to load configuration file '/vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx'.} {{} msg.dictionary.load.openFailed [{{} 1 /vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx} {{} 2 2 (No such file or directory)}] Cannot open file "/vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx": 2 (No such file or directory). }]}}} Failed to delete the virtual machine, '/vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx'.}) Failed to delete the virtual machine, '/vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx'.})
INFO[2016-08-16T21:16:02Z] Collecting ha-host hostd.log
ERRO[2016-08-16T21:16:03Z] Failed to destroy vm "VirtualMachine:4": Operation failed: Failed to delete the virtual machine, '/vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx'. - (&{{} %!s(*types.GenericVmConfigFault=&{{{{<nil> [{{} msg.vigor.offline.deleteVM [{{} 1 /vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx}] Failed to delete the virtual machine, '/vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx'.} {{} msg.vigor.method.invalidConfig [] The virtual machine configuration is invalid.} {{} msg.vigor.vmx.load.error [{{} 1 /vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx}] Unable to load configuration file '/vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx'.} {{} msg.dictionary.load.openFailed [{{} 1 /vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx} {{} 2 2 (No such file or directory)}] Cannot open file "/vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx": 2 (No such file or directory). }]}}} Failed to delete the virtual machine, '/vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx'.}) Failed to delete the virtual machine, '/vmfs/volumes/57ae54e6-4e829e2c-8ece-000c29da2bd7/ZzZ/ZzZ.vmx'.})
ERRO[2016-08-16T21:16:03Z] --------------------
ERRO[2016-08-16T21:16:03Z] vic-machine-linux failed: delete failed
```

emlin:
Fixes #2014 
In VC 6.5, if we delete datastore file directories, with linked vmdk files inside, delete will fail, even we don't have any delta vmdks rely on them.
Delete from UI does not work as well.

The only workaround for this issue, is to delete in vSAN model, delete vmdk files first, and then file directory.


lib/install/management/store_files.go
```diff
@@ -126,7 +126,7 @@ func (d *Dispatcher) deleteDatastoreFiles(ds *object.Datastore, path string, for
 	}
 
 	m := object.NewFileManager(ds.Client())
-	if d.isVSAN(ds) {
+	if d.geVC65() || d.isVSAN(ds) {
```
fdawg4l: Can we do this all the time?  Do we need to _only_ do it for 65 or vsan?

hickeng: 1. Why is the caller responsible for knowing the exact model by which you have to delete files?
2. Why even bother with separate paths (this is purely an optimization)

Can we have d.deleteFiles(m, ds, dsPath) - if we really have to have the optimization, the check should be inside that function.

emlin: Actually, this will make the delete slower, especially while there are many images downloaded.

emlin: Here, the top level method to delete should be method 
```func (d *Dispatcher) deleteDatastoreFiles(ds *object.Datastore, path string, force bool) (bool, error)```
So caller doesn't need to care about the internal impl, I think.

The last comment is same to @fdawg4l mentioned above. But I think delete interactively is not an optimization. It will slow down the delete.
But if vsphere is going this way, and we don't care too much about 6.0, it's also fine to remove this branch.

File: lib/install/management/store_files.go [hunk 2]
```diff
@@ -139,6 +139,12 @@ func (d *Dispatcher) deleteDatastoreFiles(ds *object.Datastore, path string, for
 	return true, nil
 }
 
+func (d *Dispatcher) geVC65() bool {
+	vcVersion := d.session.Client.ServiceContent.About.Version
+	// vsphere only has versions 5.5.x, 6.0.x, 6.5.x, not likely to have 6.10.x, so compare string directly here
+	return strings.Compare(vcVersion, "6.5.0") >= 0
```
emlin: so you mean anything without "6.0." prefix? In case, vsphere does not provide any patch like 6.1.x. 
anyway, not likely...

lib/install/management/store_files.go
hickeng: I don't care too much about moving to a consolidated branch, but having this logic here instead of in the pkg/vsphere/datastore code means that every single caller who wants to delete a tree has to do this same version detection or know the specifics of how to do a recursive deletion.

lib/install/management/store_files.go
hickeng: The only reason for this method is to select a difference in a specific behaviour. If this is being broken out into a separate method then it should be something like `deleteRequiresEmptyDirectory()` that actually describes the use. Otherwise we'll end up with `geVC70` as well once 6.5 is released and we start testing against vmkernel main.

lib/install/management/store_files.go
emlin: the pkg/vsphere/datastore is meant to be a common pkg could be used by both vic-machine and storage portlayer, but there is some assumption not exists in vic-machine.
I could refactor code to merge them two, but probably not at this time.

dougm:
What is the error with 6.5?  And vSAN?

"Folder deletes are always recursive" http://pubs.vmware.com/vsphere-60/index.jsp#com.vmware.wssdk.apiref.doc/vim.FileManager.html#deleteFile

emlin:
@dougm Well the problem in vSAN and 6.5 is slightly different.

vSAN: to delete a folder, the folder must be empty. 
But it is a semi-random issue. Delete folder usually works well, but if there is any error to delete a folder. Clear all files, and then delete folder will solve the problem.

6.5: delete folder works well, if there is no linked disk chain.
In our case, image pull will create vmdk files, as well as metadata file. If clear the vmdk files, delete folder works well.
So in this case, clear all files first, and then delete folder can be the workaround.

For the recursive delete, thanks for the reference. I didn't know that. 
But even that's the case, if we have thousands of images, and even more layers, it will still be much better to call vsphere once and recursively delete folder, than to call vsphere API thousands of times.

hickeng:
LGTM

andrewtchin:
lgtm

mdubya66:
lgtm

emlin:
♻️

emlin:
https://ci.vmware.run/vmware/vic/6006 passed

PR #2663, opened 2016-10-11
Add check for concurrent poweroff during container stop

### LINKED ISSUE #2533: CI integration test failure: Docker stop with sigkill
Seen in CI fairly often recently.

```
Test-Cases.Group1-Docker-Commands.1-7-Docker-Stop :: Test 1-7 - Docker Stop   
192.168.31.9

Installing VCH to test server...
Installer completed successfully: VCH-5285-1056…

Stop a container with SIGKILL using specific grace period             | FAIL |
1 != 0
```

https://ci.vmware.run/vmware/vic/5285

anchal-agrawal:
* In this [1-7-Docker-Stop](https://github.com/vmware/vic/blob/master/tests/test-cases/Group1-Docker-Commands/1-7-Docker-Stop.robot#L88) test, we sometimes get a conflicting error from vSphere due to a race between the power off and guest shutdown paths. This PR adds a check for the specific error because the container is successfully shut down.
* Also resolves govet issue on [lib/apiservers/engine/backends/image.go#L84](https://github.com/vmware/vic/blob/master/lib/apiservers/engine/backends/image.go#L84)

Fixes #2533

### DIFF
--- a/lib/apiservers/engine/backends/image.go
+++ b/lib/apiservers/engine/backends/image.go
@@ -81,7 +81,7 @@ func (i *Image) ImageDelete(imageRef string, force, prune bool) ([]types.ImageDe
 	if err != nil {
 		switch err := err.(type) {
 		case *storage.DeleteImageLocked:
-			return nil, fmt.Errorf("Failed to remove image (%s): ", imageRef, err.Payload.Message)
+			return nil, fmt.Errorf("Failed to remove image (%s): %s", imageRef, err.Payload.Message)
 		default:
 			return nil, err
 		}
--- a/lib/apiservers/engine/backends/image.go
+++ b/lib/apiservers/engine/backends/image.go
@@ -81,7 +81,7 @@ func (i *Image) ImageDelete(imageRef string, force, prune bool) ([]types.ImageDe
 	if err != nil {
 		switch err := err.(type) {
 		case *storage.DeleteImageLocked:
-			return nil, fmt.Errorf("Failed to remove image (%s): %s", imageRef, err.Payload.Message)
+			return nil, fmt.Errorf("Failed to remove image %q: %s", imageRef, err.Payload.Message)
 		default:
 			return nil, err
 		}
--- a/lib/portlayer/exec/container.go
+++ b/lib/portlayer/exec/container.go
@@ -62,7 +62,7 @@ const (
 	propertyCollectorTimeout = 3 * time.Minute
 	containerLogName         = "output.log"
 
-	notSuspendedMsg = "The virtual machine is not suspended."
+	vmNotSuspendedKey = "msg.suspend.powerOff.notsuspended"
 )
 
 // NotFoundError is returned when a types.ManagedObjectNotFound is returned from a vmomi call
@@ -477,7 +477,9 @@ func (c *Container) stop(ctx context.Context, waitTime *int32) error {
 				log.Warnf("invalid power state during power off: %s", terr.ExistingState)
 
 			case *types.GenericVmConfigFault:
-				if terr.Reason == notSuspendedMsg {
+
+				// Check if the poweroff task was canceled due to a concurrent guest shutdown
+				if len(terr.FaultMessage) > 0 && terr.FaultMessage[0].Key == vmNotSuspendedKey {
 					log.Infof("power off %s task skipped due to guest shutdown", c.ExecConfig.ID)
 					return nil
 				}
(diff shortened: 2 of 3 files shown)


lib/portlayer/exec/container.go
```diff
@@ -475,12 +476,20 @@ func (c *Container) stop(ctx context.Context, waitTime *int32) error {
 				}
 				log.Warnf("invalid power state during power off: %s", terr.ExistingState)
 
+			case *types.GenericVmConfigFault:
+				if terr.Reason == notSuspendedMsg {
```
hmahmood: Is this the only way to check for this specific error?

lib/apiservers/engine/backends/image.go
```diff
@@ -81,7 +81,7 @@ func (i *Image) ImageDelete(imageRef string, force, prune bool) ([]types.ImageDe
 	if err != nil {
 		switch err := err.(type) {
 		case *storage.DeleteImageLocked:
-			return nil, fmt.Errorf("Failed to remove image (%s): ", imageRef, err.Payload.Message)
+			return nil, fmt.Errorf("Failed to remove image (%s): %s", imageRef, err.Payload.Message)
```
emlin: maybe using %q?
=> "Failed to remove image %q: %s"

lib/portlayer/exec/container.go
hickeng: Given @hmahmood 's comment, the Key would actually be a better field to check for  `msg.suspend.powerOff.notsuspended` - I don't think these messages are internationalized, but better safe than sorry

lib/portlayer/exec/container.go
```diff
@@ -475,12 +476,20 @@ func (c *Container) stop(ctx context.Context, waitTime *int32) error {
...
+					log.Infof("power off %s task skipped due to guest shutdown", c.ExecConfig.ID)
+					return nil
+				}
+				log.Warnf("hard power off failed due to: %#v", terr)
+
 			default:
 				log.Warnf("hard power off failed due to: %#v", terr)
 			}
 		}
 		c.State = existingState
```
hickeng: @cgtexmex I think this needs to change - perhaaps move this to a `event.Set(state State)`, `cache.Set(state State)`, so that there's a single place where this gets managed.

In that model, this could would call `event.Set(STOPPING)` to report that that's what's being attempted, avoiding the issues with other VM events from vSphere and reported state changes from other operations in the PL overwriting state.

My concern with this is that we could be overwriting a value that is not c.StateStopping (most likely c.Stopped set via events).

Not for this PR, but for reference.

lib/portlayer/exec/container.go
anchal-agrawal: Fixed, thank you!

lib/apiservers/engine/backends/image.go
anchal-agrawal: Fixed.

hmahmood:
lgtm

lib/portlayer/exec/container.go
```diff
@@ -61,6 +61,8 @@ const (
 
 	propertyCollectorTimeout = 3 * time.Minute
 	containerLogName         = "output.log"
+
+	vmNotSuspendedKey = "msg.suspend.powerOff.notsuspended"
```
vburenin: Where did you get this string? Are there any similar strings? Just curious.

lib/portlayer/exec/container.go
```diff
@@ -475,12 +476,22 @@ func (c *Container) stop(ctx context.Context, waitTime *int32) error {
...
 				log.Warnf("invalid power state during power off: %s", terr.ExistingState)
 
+			case *types.GenericVmConfigFault:
+
+				// Check if the poweroff task was canceled due to a concurrent guest shutdown
+				if len(terr.FaultMessage) > 0 && terr.FaultMessage[0].Key == vmNotSuspendedKey {
+					log.Infof("power off %s task skipped due to guest shutdown", c.ExecConfig.ID)
+					return nil
+				}
+				log.Warnf("hard power off failed due to: %#v", terr)
```
vburenin: Is there any way to avoid this warning duplicate?

anchal-agrawal: Fixed by moving it outside the `switch`. Thanks!

anchal-agrawal:
🏁

lib/portlayer/exec/container.go
anchal-agrawal: Follow-up: this doesn't log the specific typed error, which is useful for debugging intermittent failures such as the one in https://ci.vmware.run/vmware/vic/5831. I've posted a change in #2732.

PR #2665, opened 2016-10-11
Lower the number of volumes created in integration test

fdawg4l:
This test was taking entirely too long.  Lowered from 100 to 10 and
decreased size from 512MB to 16MB.


fdawg4l:
cc @matthewavery and @dougm

matthewavery:
LGTM - yeah Mike originally wrote that test in an attempt to stress the mounting process possibly and cause a `TIP` error. it never really worked. Still a valuable test nonetheless.

caglar10ur:
LGTM :sheep: :it:

PR #2687, opened 2016-10-12
Check container status before waiting it to shutdown. Safer work with…

### LINKED ISSUE #2661: Container wait race
I ran into this while running integration tests locally. exec.WaitForContainerStop will time out if the container stops while that code is executing (specifically before we subscribe to the powered off event). We do a check higher up in the stack in the ContainerWait handler, but that fails since the container is still running at that point.

vburenin:
I started working on #2661, I was significantly worried about a potential issues related to race conditions which could happen due to not protected container state. Thus, I started dig deeper pursuing two goals: get better understanding of codebase and eliminate potential race issues in the future.

I am not really confident that my fix is not breaking for anything else, so I would ask you folks to take a look if I didn't break anything.

Fixes #2661


lib/portlayer/exec/container.go
```diff
@@ -151,6 +151,20 @@ func (s State) String() string {
 	return ""
 }
 
+// CurrentState returns current state.
+func (c *Container) CurrentState() State {
```
fdawg4l: Why is this preferred vs accessing `State` publicly?  Are you trying to avoid torn writes or...?

fdawg4l: Chatted with @hickeng, and the idiom is preferred because of torn write/reads.  Makes sense.  But please do post if there's another reason.

lib/portlayer/exec/container.go
```diff
@@ -151,6 +151,20 @@ func (s State) String() string {
...
+	c.m.Lock()
+	defer c.m.Unlock()
```
fdawg4l: Might want to invert these two lines.  

```
defer c.m.Unlock()
c.m.Lock()
```

lib/portlayer/exec/container.go
```diff
@@ -151,6 +151,20 @@ func (s State) String() string {
...
+	return c.state
+}
+
+// SetState changes container state.
+func (c *Container) SetState(s State) {
```
hickeng: Also `Test & Set` or we still have a race - see the `state = existingState` path.

hmahmood: Why not use `defer c.m.Unlock()` here?

lib/portlayer/exec/handle.go
```diff
@@ -211,9 +213,8 @@ func (h *Handle) Close() {
 	removeHandle(h.key)
 }
 
-func (h *Handle) SetState(s State) {
-	h.State = new(State)
-	*h.State = s
+func (h *Handle) SetContainerState(s State) {
```
hmahmood: So this method is there to indicate a state change. This was not meant to change the cached state of the container, but to indicate during Commit if we want to change the state of the container.

We need two states here -- existing and new. This method should be to change the state. We can add an accessor for getting the current state -- that one could call h.Container.CurrentState().

lib/portlayer/exec/container.go
hmahmood: That path is protected by the same lock I think, so we should be OK.

hmahmood: Come to think of it, this should not be an exported method on the Container object. The only way you can change the state of the container is through the handle.SetState, followed by Commit.

lib/portlayer/exec/exec.go
```diff
@@ -224,6 +223,11 @@ func WaitForContainerStop(ctx context.Context, id string) error {
 	Config.EventManager.Subscribe(topic, sub, watch)
 	defer Config.EventManager.Unsubscribe(topic, sub)
 
+	// Container has changed it's status, existing.
+	if c.CurrentState() == StateStopped {
```
vburenin: I don't think we can miss anything here. We are already subscribed to events, thus it is only needed to avoid the case you submitted bug about.

vburenin:
Thank you guys for your feedback, this is really valuable since I am not sure about the impact it makes.

lib/portlayer/exec/container.go
hickeng: unless we've untangled the state the incoming events can change the container state, but that should not reflect in the handle.

hmahmood: Changing the state through events should be the only other way.

hickeng:
@vburenin the lifetime of a handle should only be while marshalling a container mutation and should be treated like the target of an `append` or a `realloc` - it might be the same thing after the call but you must not assume that.

```
// get new handle
handle := Get()

// do some things with it, with it being updated each time
handle2 := network.Join(handle, network)
// handle is now invalid
handle3 := volumes.Join(handle2, volume)
// handle2 is now invalid
handle4 := interaction.Join(handle3)
// handle3 is now invalid

// the the mutation is committed - whatever
err := exec.Commit(handle4)
// handle4 is now invalid
```

lib/portlayer/exec/handle.go
```diff
@@ -87,11 +86,9 @@ func newHandle(con *Container) *Handle {
 		committed:  false,
 		Container:  con,
 		ExecConfig: *con.ExecConfig,
-		State:      new(State),
+		state:      StateUnknown,
```
hmahmood: Shouldn't this be initialized to con.CurrentState()?

lib/portlayer/exec/container.go
vburenin: @hickeng I am not sure I get what you mean. "start" method is called by Commit, which acquires that lock.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -358,20 +358,16 @@ func (handler *ContainersHandlersImpl) ContainerWaitHandler(params containers.Co
 		})
 	}
 
-	// if the container is already stopped return
-	if c.State != exec.StateRunning {
+	select {
+	case <-c.WaitForState(exec.StateStopped):
+		c = exec.Containers.Container(uid.Parse(params.ID).String())
```
hmahmood: What do you we need this? You already have a container pointer.

lib/portlayer/exec/container.go
```diff
@@ -355,15 +408,15 @@ func (c *Container) start(ctx context.Context) error {
 	}
 	// get existing state and set to starting
 	// if there's a failure we'll revert to existing
-	existingState := c.State
-	c.State = StateStarting
+	finalState := c.updateState(StateStarting)
+
+	defer func() { c.updateState(finalState) }()
```
hmahmood: We are unconditionally setting the state back to what it was?

lib/portlayer/exec/container.go
```diff
@@ -563,7 +616,9 @@ func (c *Container) LogReader(ctx context.Context, tail int, follow bool) (io.Re
 		}
 	}
 
-	if follow && c.State == StateRunning {
+	c.m.Lock()
```
hmahmood: The lock should really be at the beginning of this function since we are accessing `c.vm`.

hmahmood:
Except for that one comment above, LGTM

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
vburenin: copy/paste

lib/portlayer/exec/container.go
vburenin: Makes sense. However, I am paying attention to the managing a state. Nevertheless, I addressed it.

lib/portlayer/exec/container.go
caglar10ur: @fdawg4l I'm confused, why you want to invert/

lib/portlayer/exec/handle.go
```diff
@@ -24,10 +24,9 @@ import (
 	"sync"
 	"time"
 
-	"golang.org/x/net/context"
-
 	log "github.com/Sirupsen/logrus"
 	"github.com/golang/groupcache/lru"
+	"golang.org/x/net/context"
```
caglar10ur: doesn't have to be in this PR but we should start switching to "context"

caglar10ur:
LGTM as I only have cosmetic comments

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -358,20 +358,15 @@ func (handler *ContainersHandlersImpl) ContainerWaitHandler(params containers.Co
 		})
 	}
 
-	// if the container is already stopped return
-	if c.State != exec.StateRunning {
+	select {
+	case <-c.WaitForState(exec.StateStopped):
```
hickeng: The rest handlers are purely about exposing the RPC - we shouldn't be packing additional logic into them.

Is there a reason not to put the select clause into exec.Wait?

lib/portlayer/exec/container.go
```diff
@@ -109,15 +109,18 @@ type Container struct {
...
 	con := &Container{
-		ExecConfig: &executor.ExecutorConfig{},
-		State:      StateCreating,
+		ExecConfig:     &executor.ExecutorConfig{},
+		state:          StateCreating,
+		newStateEvents: make(map[State]chan struct{}),
 	}
 	con.ExecConfig.ID = id.String()
-	return newHandle(con)
+	return newHandle(con, con.state)
```
hickeng: This should not be set to the current container state - the handle.state is a request to transition to that power state, not a record of the current state.

lib/portlayer/exec/container.go
```diff
@@ -146,11 +149,65 @@ func (s State) String() string {
...
 
+// CurrentState returns current state.
+func (c *Container) CurrentState() State {
+	c.m.Lock()
+	defer c.m.Unlock()
+	return c.state
+}
+
+// SetState changes container state.
+func (c *Container) SetState(s State) State {
```
hickeng: Please make this test&set  as per initial comments - otherwise we're still racing on restoring existingState on error paths.

lib/portlayer/exec/container.go
```diff
@@ -377,15 +433,15 @@ func (c *Container) start(ctx context.Context) error {
 
 	detail, err = c.vm.WaitForKeyInExtraConfig(ctx, key)
```
hickeng: once we've powered the VM on, this should be StateUnknown on general error paths - this allows subsequent calls to attempt an update.

lib/portlayer/exec/container.go
```diff
@@ -377,15 +433,15 @@ func (c *Container) start(ctx context.Context) error {
...
 	if err != nil {
-		c.State = existingState
 		return fmt.Errorf("unable to wait for process launch status: %s", err.Error())
 	}
 
 	if detail != "true" {
```
hickeng: This error path should set state to Stopping - that's the expected behaviour if we fail to launch a process in the containerVM

lib/portlayer/exec/handle.go
hickeng: We should be switching to trace.Operation ;)

lib/portlayer/exec/handle.go
```diff
@@ -81,17 +80,15 @@ func newHandleKey() string {
 	return hex.EncodeToString(b)
 }
 
-func newHandle(con *Container) *Handle {
+func newHandle(con *Container, state State) *Handle {
```
hickeng: should not be taking a state parameter here - possible this shouldn't even be the same type as the State's we've got elsewhere.
The only valid states for this currently are:
StateStopped
StateRunning

We may want to add suspend/resume in the future, but this field is an imperative instruction, not a description and as such intermediate states will always be invalid here.

lib/portlayer/exec/handle.go
```diff
@@ -181,17 +186,15 @@ func (h *Handle) Commit(ctx context.Context, sess *session.Session, waitTime *in
...
-		case StateRunning:
-			se := h.ExecConfig.Sessions[h.ExecConfig.ID]
-			se.StartTime = time.Now().UTC().Unix()
-			h.ExecConfig.Sessions[h.ExecConfig.ID] = se
-		case StateStopped:
-			se := h.ExecConfig.Sessions[h.ExecConfig.ID]
-			se.StopTime = time.Now().UTC().Unix()
-			h.ExecConfig.Sessions[h.ExecConfig.ID] = se
-		}
+	switch h.CurrentState() {
```
hickeng: Seeing this here I really think we should not be overloading the imperative state field in the handle, with the descriptive state field in the container. They have different valid values and are used for different things. This is going to lead to even more confusion in the future.

Issue #2669, opened 2016-10-12
Failure in build 5688 - cross container ping


hickeng:
Cross container ping test is failing. It's expecting `2 packets transmitted, 2 packets received` but the test isn't seeing it.

The tether logs shows us writing it (to both the console and the log):

```
time="2016-10-12T00:27:01Z" level=debug msg="[0xc4201a7e00] write \"64 bytes from 172.18.0.2: seq=1 ttl=64 time=0.170 ms\\n\\n--- cross1-container ping statistics ---\\n2 packets transmitted, 2 packets received, 0% packet loss\\nround-trip min/avg/max = 0.170/0.210/0.251 ms\\n\" to 2 writers (err: <nil>)" 
```

output.xml

```
<msg timestamp="20161012 00:29:18.982" level="INFO">Running command 'docker -H 192.168.31.210:2375 network connect cross1-network2 084b85f3cbc9457ec277c16603c4441cad79e505bf8f43631f70509f7e8a6b47 2&gt;&amp;1'.</msg>
<msg timestamp="20161012 00:29:19.129" level="INFO">${rc} = 0</msg>
<msg timestamp="20161012 00:29:19.129" level="INFO">${output} = </msg>
<status status="PASS" endtime="20161012 00:29:19.129" starttime="20161012 00:29:18.979"></status>
</kw>
<kw name="Should Be Equal As Integers" library="BuiltIn">
<doc>Fails if objects are unequal after converting them to integers.</doc>
<arguments>
<arg>${rc}</arg>
<arg>0</arg>
</arguments>
<msg timestamp="20161012 00:29:19.130" level="INFO">Argument types are:
&lt;type 'int'&gt;
&lt;type 'unicode'&gt;</msg>
<status status="PASS" endtime="20161012 00:29:19.130" starttime="20161012 00:29:19.130"></status>
</kw>
<kw name="Run And Return Rc And Output" library="OperatingSystem">
<doc>Runs the given command in the system and returns the RC and output.</doc>
<arguments>
<arg>docker ${params} start ${containerID}</arg>
</arguments>
<assign>
<var>${rc}</var>
<var>${output}</var>
</assign>
<msg timestamp="20161012 00:29:19.133" level="INFO">Running command 'docker -H 192.168.31.210:2375 start 084b85f3cbc9457ec277c16603c4441cad79e505bf8f43631f70509f7e8a6b47 2&gt;&amp;1'.</msg>
[...]

PR #2694, opened 2016-10-12
Revert "Adding output.log to container log bundle"

hmahmood:
This need to be guarded by DebugLevel > 2


mhagen-vmware:
lgtm

hickeng:
@hmahmood we need the output logs to figure out #2669 - this needs to be addressed before GA but it doesn't need to not be there right now ;)
It was only last week that I took this back out (it snuck in as part of the containerVM log renaming). There's also the fact that if we have debug then the application output is logged in tether.debug anyway.

PR #2697, opened 2016-10-12
Set tracing when DebugLevel > 0

caglar10ur:
Otherwise serial package creates so much noise.


pkg/serial/handshake.go
```diff
@@ -40,6 +40,9 @@ const (
 
 // PurgeIncoming is used to clear a channel of bytes prior to handshaking
 func PurgeIncoming(conn net.Conn) {
+	if tracing {
```
fdawg4l: BTW, you could modify `trace.newTrace`, make it public, and wrap this `if() { trace... }` in function with an appropriate skip (3 looks right).

fdawg4l:
LGTM

pkg/serial/handshake.go
caglar10ur: But that would either require passing a new param to Begin, or doing more calls to trace hence breaking that defer End(Begin(msg)) chain

anchal-agrawal:
:shipit:

caglar10ur:
♻️

PR #2700, opened 2016-10-12
Modify Operation to print the stack in stack order

fdawg4l:
Also only print the Begin header when an operation is created, and not
when it's inherited or mutated.


fdawg4l:
@hickeng these are the changes you asked for to make debugging slightly better.  Ping if I missed anything.

fdawg4l:
@caglar10ur , I had to modify the `Logger` in the trace package to specify a `TextFormatter` that handles control codes to do line wrapping.  Yes, really.  Please take a look and if you have a less crappy alternative, I'm all :ear:.

hickeng:
LGTM

pkg/trace/trace.go
```diff
@@ -16,13 +16,21 @@ package trace
...
 	"time"
 
 	log "github.com/Sirupsen/logrus"
 )
 
-var Logger = log.New()
+var Logger = &log.Logger{
+	Out: os.Stderr,
+	// Apparently this forces terminal handling even when Out isn't a TTY (like
+	// on a VCH, where out is currently a serial dev)
```
fdawg4l: It's actually a file.  Will fix this.

caglar10ur:
LGTM :sheep: :it:

fdawg4l:
.

Issue #2646, opened 2016-10-11
Test race: TestAbsPath is racing tether shutdown path


hickeng:
https://ci.vmware.run/vmware/vic/5591

```
time="2016-10-10T17:30:33Z" level=info msg="Started test setup for github.com/vmware/vic/lib/tether.TestPathLookup" 
time="2016-10-10T17:30:33Z" level=debug msg="Test configuration: (func(string, string) error)(0x496820)" 
time="2016-10-10T17:30:33Z" level=info msg="Registering tether extension Mocker" 
time="2016-10-10T17:30:33Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/tether.(*tether).Start:189] main tether loop" 
time="2016-10-10T17:30:33Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/tether.(*tether).setup:105] main tether setup" 
time="2016-10-10T17:30:33Z" level=info msg="Started reaping child processes" 
time="2016-10-10T17:30:33Z" level=info msg="Starting extension Mocker" 
time="2016-10-10T17:30:33Z" level=error msg="Unable to open PID file for /tmp/go-build980344785/github.com/vmware/vic/lib/tether/_test/tether.test : open /tmp/var/run/tether.test.pid: no such file or directory" 
time="2016-10-10T17:30:33Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/tether.(*tether).setup:105] [173.925µs] main tether setup" 
time="2016-10-10T17:30:33Z" level=info msg="Loading main configuration" 
time="2016-10-10T17:30:33Z" level=info msg="Loaded executor config" 
time="2016-10-10T17:30:33Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/tether.(*Mocker).SetHostname:137] mocking hostname to pathlookup" 
time="2016-10-10T17:30:33Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/tether.(*Mocker).SetHostname:137] [14.164µs] mocking hostname to pathlookup" 
time="2016-10-10T17:30:33Z" level=debug msg="Processing config for session pathlookup" 
time="2016-10-10T17:30:33Z" level=info msg="Launching process for session pathlookup" 
time="2016-10-10T17:30:33Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/tether.(*tether).launch:368] launching session pathlookup" 
[...]

Issue #2330, opened 2016-09-15
Ordering issue with child reaping in tether


hickeng:
```
time="2016-09-14T20:13:52Z" level=info msg="Started test setup for github.com/vmware/vic/lib/tether.TestPathLookup" 
time="2016-09-14T20:13:52Z" level=debug msg="Test configuration: (func(string, string) error)(0x496a10)" 
time="2016-09-14T20:13:52Z" level=info msg="Registering tether extension Mocker" 
time="2016-09-14T20:13:52Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/tether.(*tether).Start] main tether loop" 
time="2016-09-14T20:13:52Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/tether.(*tether).setup] main tether setup" 
time="2016-09-14T20:13:52Z" level=info msg="Started reaping child processes" 
time="2016-09-14T20:13:52Z" level=info msg="Starting extension Mocker" 
time="2016-09-14T20:13:52Z" level=info msg="listen tcp 0.0.0.0:6060: bind: address already in use" 
time="2016-09-14T20:13:52Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/tether.(*tether).setup] [598.298µs] main tether setup" 
time="2016-09-14T20:13:52Z" level=info msg="Loading main configuration" 
time="2016-09-14T20:13:52Z" level=info msg="Loaded executor config" 
time="2016-09-14T20:13:52Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/tether.(*Mocker).SetHostname] mocking hostname to pathlookup" 
time="2016-09-14T20:13:52Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/tether.(*Mocker).SetHostname] [440ns] mocking hostname to pathlookup" 
time="2016-09-14T20:13:52Z" level=debug msg="Processing config for session pathlookup" 
time="2016-09-14T20:13:52Z" level=info msg="Launching process for session pathlookup" 
time="2016-09-14T20:13:52Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/tether.(*tether).launch] launching session pathlookup" 
time="2016-09-14T20:13:52Z" level=debug msg="[0xc42000d100] created multireader" 
time="2016-09-14T20:13:52Z" level=debug msg="Resolved date to /bin/date" 
[...]

PR #2702, opened 2016-10-12
[full ci] Fix multiple data races reported by race detector

### LINKED ISSUE #2647: [tether] serverConn needs to protected by a lock
```
goroutine 12 [running]:
panic(0x8a3640, 0xc42000c050)
        /usr/local/go/src/runtime/panic.go:500 +0x1a1
main.(*attachServerSSH).run.func1(0xc420136088)
        /opt/go/src/github.com/vmware/vic/cmd/tether/attach.go:216 +0xa8
main.(*attachServerSSH).run(0xc420059ec0, 0x0, 0x0)
        /opt/go/src/github.com/vmware/vic/cmd/tether/attach.go:313 +0x175b
created by main.(*attachServerSSH).start
        /opt/go/src/github.com/vmware/vic/cmd/tether/attach.go:144 +0x334
```

caglar10ur:
- attachServerSSH uses sync/atomic to get/set enabled value as there was
a race between stop and run

- underlying *net.Conn is now protected by its own mutex as there was a
race between stop and run

- *ssh.ServerConn is now protected by its own mutex as we are sending
it to channel mux (as a part of the cleanup function)

- incoming signals channel in tether is now accompanied by another
channel called done. Its whole purpose is to stop the reaper loop after we
close the incoming channel. Without this we keep reading from it.

Fixes #2647, #2646 and #2330

### DIFF
--- a/cmd/tether/attach.go
+++ b/cmd/tether/attach.go
@@ -44,7 +44,7 @@ type AttachServer interface {
 	tether.Extension
 
 	start() error
-	stop()
+	stop() error
 }
 
 type attachServerSSH struct {
@@ -125,19 +125,22 @@ func (t *attachServerSSH) Stop() error {
 	defer t.m.Unlock()
 
 	// calling server.start not t.start so that test impl gets invoked
-	server.stop()
-	return nil
+	return server.stop()
 }
 
 func (t *attachServerSSH) start() error {
 	defer trace.End(trace.Begin("start attach server"))
 
 	if t == nil {
-		return errors.New("attach server is not configured")
+		err := fmt.Errorf("attach server is not configured")
+		log.Error(err)
+		return err
 	}
 
 	if t.IsEnabled() {
-		return nil
+		err := fmt.Errorf("attach server is already enabled")
+		log.Error(err)
+		return err
 	}
 
 	// don't assume that the key hasn't changed
@@ -175,11 +178,19 @@ func (t *attachServerSSH) start() error {
 }
 
 // stop is not thread safe with start
-func (t *attachServerSSH) stop() {
+func (t *attachServerSSH) stop() error {
 	defer trace.End(trace.Begin("stop attach server"))
 
-	if t == nil || !t.IsEnabled() {
-		return
+	if t == nil {
+		err := fmt.Errorf("attach server is not configured")
+		log.Error(err)
+		return err
+	}
+
+	if !t.IsEnabled() {
+		err := fmt.Errorf("attach server is not enabled")
+		log.Error(err)
+		return err
 	}
 
 	log.Debugf("Setting enabled to false")
@@ -197,6 +208,8 @@ func (t *attachServerSSH) stop() {
 	}
 	t.conn.Unlock()
 	log.Debugf("Released the connection lock")
+
+	return nil
 }
 
 // run should not be called directly, but via start
@@ -222,14 +235,13 @@ func (t *attachServerSSH) run() error {
 		// keep waiting for the connection to establish
 		for !established && t.IsEnabled() {
 			log.Debugf("Trying to establish a connection")
-			// tests are passing their own connections so do not create connections when testing is set
 
 			establishFn := func() error {
 				t.conn.Lock()
 				defer t.conn.Unlock()
 
+				// tests are passing their own connections so do not create connections when testing is set
 				if !t.testing {
-
 					// close the connection if required
 					if t.conn.conn != nil {
 						(*t.conn.conn).Close()
@@ -399,7 +411,6 @@ func (t *attachServerSSH) globalMux(reqchan <-chan *ssh.Request) {
 			}
 			msg := msgs.ContainersMsg{IDs: keys}
 			payload = msg.Marshal()
-
 		default:
 			ok = false
 			payload = []byte("unknown global request type: " + req.Type)
--- a/cmd/tether/attach.go
+++ b/cmd/tether/attach.go
@@ -204,15 +204,17 @@ func (t *attachServerSSH) stop() error {
 	log.Debugf("Setting enabled to false")
 	t.Disabled()
 
-	// This context is used by backchannel and we need to cancel it before
-	// trying to obtain the following lock so that backchannel releases it.
+	// This context is used by backchannel only. We need to cancel it before
+	// trying to obtain the following lock so that backchannel interrupts the
+	// underlying Read call by calling Close on it.
+	// The lock is  held by backchannel's caller and not released until it returns
 	log.Debugf("Canceling the context")
 	t.cancel()
 
 	log.Debugf("Acquiring the connection lock")
 	t.conn.Lock()
 	if t.conn.conn != nil {
-		log.Debugf("Closing the connection")
+		log.Debugf("Close called again on rawconn - squashing")
 		(*t.conn.conn).Close()
 		t.conn.conn = nil
 	}
@@ -235,25 +237,28 @@ func backchannel(ctx context.Context, conn *net.Conn) error {
 	ticker := time.NewTicker(10 * time.Millisecond)
 	defer ticker.Stop()
 
-	ch := make(chan struct{}, 1)
-
+	// We run this in a seperate goroutine because HandshakeServer
+	// calls a Read on rawconn which is a blocking call which causes
+	// the caller to block as well so this is the only way to cancel.
+	// Calling Close() will unblock us and on the next tick we will
+	// return ctx.Err()
 	go func() {
 		select {
 		case <-ctx.Done():
 			(*conn).Close()
-			close(ch)
 		}
 	}()
 
 	for {
 		select {
 		case <-ticker.C:
+			if ctx.Err() != nil {
+				return ctx.Err()
+			}
 			err := serial.HandshakeServer(ctx, *conn)
 			if err == nil {
 				return nil
 			}
-		case <-ch:
-			return ctx.Err()
 		}
 	}
 }
--- a/cmd/tether/attach.go
+++ b/cmd/tether/attach.go
@@ -15,15 +15,16 @@
 package main
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net"
 	"sync"
+	"sync/atomic"
 	"syscall"
 
 	log "github.com/Sirupsen/logrus"
 	"golang.org/x/crypto/ssh"
-	"golang.org/x/net/context"
 
 	"github.com/vmware/vic/cmd/tether/msgs"
 	"github.com/vmware/vic/lib/tether"
@@ -51,11 +52,18 @@ type attachServerSSH struct {
 	m sync.Mutex
 
 	// conn is held directly as it's how we stop the attach server
-	conn      *net.Conn
+	conn struct {
+		sync.Mutex
+		conn *net.Conn
+	}
+
 	config    *tether.ExecutorConfig
 	sshConfig *ssh.ServerConfig
 
-	enabled bool
+	enabled int32
+
+	ctx    context.Context
+	cancel context.CancelFunc
 
 	// INTERNAL: must set by testAttachServer only
 	testing bool
@@ -64,7 +72,12 @@ type attachServerSSH struct {
 // NewAttachServerSSH either creates a new instance or returns the initialized one
 func NewAttachServerSSH() AttachServer {
 	once.Do(func() {
-		server = &attachServerSSH{}
+		// create a cancelable context and assign it to the CancelFunc
+		ctx, cancel := context.WithCancel(context.Background())
+		server = &attachServerSSH{
+			ctx:    ctx,
+			cancel: cancel,
+		}
 	})
 	return server
 }
@@ -85,6 +98,21 @@ func (t *attachServerSSH) Reload(config *tether.ExecutorConfig) error {
 		return errors.New(detail)
 	}
 	return nil
+}
+
+// Enabled sets the enabled to true
+func (t *attachServerSSH) Enabled() {
+	atomic.StoreInt32(&t.enabled, 1)
+}
+
+// Disabled sets the enabled to false
+func (t *attachServerSSH) Disabled() {
+	atomic.StoreInt32(&t.enabled, 0)
+}
+
+// IsEnabled returns whether the enabled is true
+func (t *attachServerSSH) IsEnabled() bool {
+	return atomic.LoadInt32(&t.enabled) == 1
 }
 
 // Start is implemented at _ARCH.go files
@@ -108,7 +136,7 @@ func (t *attachServerSSH) start() error {
 		return errors.New("attach server is not configured")
 	}
 
-	if t.enabled {
+	if t.IsEnabled() {
 		return nil
 	}
 
@@ -140,7 +168,7 @@ func (t *attachServerSSH) start() error {
 	}
 	t.sshConfig.AddHostKey(pkey)
 
-	t.enabled = true
+	t.Enabled()
 	go t.run()
 
 	return nil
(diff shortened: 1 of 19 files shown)


cmd/tether/attach.go
```diff
@@ -87,6 +92,21 @@ func (t *attachServerSSH) Reload(config *tether.ExecutorConfig) error {
 	return nil
 }
 
+// Enabled sets the enabled to true
+func (t *attachServerSSH) Enabled() {
+	atomic.StoreInt32(&t.enabled, 1)
```
hmahmood: Do we need this given that you grab the attachServerSSH lock when you change enabled in start() and stop() below?

hmahmood: You need this for the IsReload() call in run() ... nvm.

cmd/tether/attach.go
```diff
@@ -200,21 +239,28 @@ func (t *attachServerSSH) run() error {
...
 		defer func() {
 			log.Debugf("cleanup on connection")
 
-			if serverConn != nil {
+			serverConn.Lock()
+			if serverConn.ServerConn != nil {
 				log.Debugf("closing underlying connection")
 				serverConn.Close()
 			}
+			serverConn.Lock()
```
hmahmood: `Unlock()`? I would use `defer` here.

caglar10ur: Done

cmd/tether/attach.go
```diff
@@ -150,46 +170,65 @@ func (t *attachServerSSH) start() error {
...
+		serverConn.Lock()
+		established = serverConn.ServerConn != nil
+		serverConn.Unlock()
 		// keep waiting for the connection to establish
-		for serverConn == nil {
+		for !established {
+			log.Debugf("Trying to establish a connection")
 			// tests are passing their own connections so do not create connections when testing is set
 			if !t.testing {
+				t.conn.Lock()
```
caglar10ur: Done

lib/tether/tether.go
```diff
@@ -65,6 +65,7 @@ type tether struct {
 	sink extraconfig.DataSink
 
 	incoming chan os.Signal
+	done     chan bool
```
hmahmood: If you just want to signal stop, use `chan struct{}`, and then close the channel when you want to signal.

caglar10ur:
@hmahmood refactoring the code showed another race between the Close and the Backchannel, would be glad if you could review again

lib/tether/tether_linux.go
```diff
@@ -126,10 +136,17 @@ func (t *tether) childReaper() error {
...
+	// Ordering is important otherwise we may one goroutine closing, and the other goroutine is trying to write afterwards
+	log.Debugf("Removing the signal notifier")
 	signal.Reset(syscall.SIGCHLD)
+
+	log.Debugf("Closing the channel done to reaper")
 	close(t.incoming)
+
+	log.Debugf("Sending true to the done channel")
+	// just closing the channel is not going to stop the iteration over the channel so signal it as well
+	t.done <- true
```
caglar10ur: We weren't now we are :)

cmd/tether/attach.go
caglar10ur: Done

caglar10ur: Found another issue and was fixing that, pushed couple of min. ago

pkg/serial/handshake.go
```diff
@@ -162,9 +162,26 @@ func HandshakeClient(ctx context.Context, conn net.Conn, debug bool) error {
...
+			n, err := conn.Read(b)
+			c <- ioret{n, err}
+			close(c)
+		}()
+
+		select {
+		case ret := <-c:
+			return ret.n, ret.err
+		case <-ctx.Done():
+			return 0, ctx.Err()
```
caglar10ur: SetReadDeadline doesn't work on serial (rawconn doesn't implement them) and I don't know any other way to cancel a read

cmd/tether/attach.go
```diff
@@ -150,68 +178,102 @@ func (t *attachServerSSH) start() error {
 func (t *attachServerSSH) stop() {
 	defer trace.End(trace.Begin("stop attach server"))
 
-	if t == nil || !t.enabled {
+	if t == nil || !t.IsEnabled() {
```
fdawg4l: Think this is worth a `Debugf`?

cmd/tether/attach.go
```diff
@@ -150,68 +178,102 @@ func (t *attachServerSSH) start() error {
...
+			log.Debugf("Trying to establish a connection")
 			// tests are passing their own connections so do not create connections when testing is set
-			if !t.testing {
-				// close the connection if required
-				if t.conn != nil {
-					(*t.conn).Close()
+
+			establishFn := func() error {
+				t.conn.Lock()
+				defer t.conn.Unlock()
```
fdawg4l: `for` loop with a `defer Unlock`.  You'll pile these `defer`s up if you don't exit the loop, right?

cmd/tether/attach.go
```diff
@@ -51,11 +52,18 @@ type attachServerSSH struct {
 	m sync.Mutex
 
 	// conn is held directly as it's how we stop the attach server
-	conn      *net.Conn
+	conn struct {
+		sync.Mutex
```
fdawg4l: Reading the code, I think i understand the lock contract: Enabled, conn lock, and Disabled, con unlock. I think :-). Can you add a (big honking) comment about the lock order?

lib/tether/tether_linux.go
```diff
@@ -68,9 +68,11 @@ func (t *tether) childReaper() error {
 
 	go func() {
 		var status syscall.WaitStatus
-
 		flag := syscall.WNOHANG | syscall.WUNTRACED | syscall.WCONTINUED
+
 		for range t.incoming {
+			log.Debugf("Woke up because of a received SIGCHLD")
```
hickeng: This message and "Inspecting children with status change" will always proc together - redundant

cmd/tether/attach.go
caglar10ur: Done (pushed as a new commit)

lib/tether/tether_linux.go
```diff
@@ -126,10 +137,16 @@ func (t *tether) childReaper() error {
 func (t *tether) stopReaper() {
```
hickeng: can you confirm that we don't call this until all directly launched children have exited?

cmd/tether/attach.go
```diff
@@ -140,78 +171,121 @@ func (t *attachServerSSH) start() error {
...
+	t.Disabled()
 
-	if conn != nil {
-		(*conn).Close()
+	log.Debugf("Canceling the context")
+	t.cancel()
+
+	log.Debugf("Acquiring the connection lock")
+	t.conn.Lock()
+	if t.conn.conn != nil {
```
fdawg4l: Looks like in `Start()`, Enabled is set, then `t.conn.Lock()` is acquired, and `t.conn.conn` is assigned in `run()`.  But here, we set disabled, acquire the `conn` lock , then do this check.  Is it possible Start() will set enabled, then in another routine we call `stop`, set disabled, acquire the `conn.Lock`, find this nil before `run()` has a chance to assign, fall out of `stop`, and run now has a dangling `conn`?

lib/tether/tether_linux.go
```diff
@@ -126,10 +137,16 @@ func (t *tether) childReaper() error {
...
 	defer trace.End(trace.Begin("Shutting down child reaping"))
 
-	// stop child reaping
-	log.Info("Shutting down reaper")
+	// Ordering is important otherwise we may one goroutine closing, and the other goroutine is trying to write afterwards
+	log.Debugf("Removing the signal notifier")
 	signal.Reset(syscall.SIGCHLD)
+
+	log.Debugf("Closing the channel done to reaper")
```
hickeng: Closing the reapers signal channel

lib/tether/tether_linux.go
```diff
@@ -126,10 +137,16 @@ func (t *tether) childReaper() error {
...
 	close(t.incoming)
+
+	log.Debugf("Sending true to the done channel")
+	// just closing the channel is not going to stop the iteration over the channel so signal it as well
```
hickeng: shouldn't `done` be closed first if incoming potentially has queued elements? It's the most immediate way of stopping reaping from continuing.
Of course calling code shouldn't be assuming reaping has stopped until this call returns, but...

cmd/tether/attach.go
caglar10ur: Yep, will do. It's a litte bit mess right now as we communicate by sharing memory :/

cmd/tether/attach.go
hickeng: @fdawg4l there is an error path starting at line 250 that I think would result in this.

cmd/tether/attach.go
```diff
@@ -140,78 +171,121 @@ func (t *attachServerSSH) start() error {
...
-			// wait for backchannel to establish
-			err = backchannel(context.Background(), conn)
-			if err != nil {
-				detail := fmt.Errorf("failed to establish backchannel: %s", err)
-				log.Error(detail)
-				continue
-			}
+				// create the SSH server
+				serverConn.Lock()
+				defer serverConn.Unlock()
```
fdawg4l: Ok, this _isn't_ in an anon func :-)

Here, can we Lock, defer the unlock, and come back around again and possibly deadlock ourselves?  Likewise the `defer` may never run, right?

cmd/tether/attach.go
caglar10ur: Correct me if I'm wrong but what you are saying is only valid (I think) if you are calling stop and start concurrently, right? Those are not exported, instead we have Start and Stop which syncronized via mutex so this shouldn't be a problem outside of this package. And if you are calling start/stop within the package then I think correct answer could be "deal with it" :)

cmd/tether/attach.go
hmahmood: Closing brace for anon func on line 278 :-)

hmahmood:
lgtm

cmd/tether/attach.go
caglar10ur: The diff is making it hard to see, they are all in the same function. Looking at it directly could help maybe https://github.com/caglar10ur/vic/blob/tether/cmd/tether/attach.go#L239

cmd/tether/attach.go
fdawg4l: Sounds good :-)

cmd/tether/attach.go
fdawg4l: AHHh...  Dammit :-) #imissgerrit.  Thanks!

fdawg4l:
lgtm

pkg/serial/handshake.go
hickeng: Implementing rawconn.SetReadDeadline might be easier to parse and is consistent with expected net.Conn behaviour.
I'd actually prefer conn.SetReadDeadline(ctx) but it doesn't match the interface

cmd/tether/attach.go
```diff
@@ -150,68 +178,102 @@ func (t *attachServerSSH) start() error {
...
-		for serverConn == nil {
+		for !established && t.IsEnabled() {
+			log.Debugf("Trying to establish a connection")
 			// tests are passing their own connections so do not create connections when testing is set
-			if !t.testing {
-				// close the connection if required
-				if t.conn != nil {
-					(*t.conn).Close()
+
+			establishFn := func() error {
```
caglar10ur: I think err is more idiomatic here as in reality that function ends up with an error so I prefer this way unless you really want bool there :)

cmd/tether/attach_linux.go
```diff
@@ -106,13 +106,15 @@ func (t *attachServerSSH) Start() error {
 		return detail
 	}
 
-	t.conn, err = rawConnectionFromSerial()
+	t.conn.Lock()
+	defer t.conn.Unlock()
+
+	t.conn.conn, err = rawConnectionFromSerial()
```
caglar10ur: Nope, just a leftover removing

lib/tether/tether_linux.go
caglar10ur: Done

lib/tether/tether_linux.go
caglar10ur: Eheeh not sure what I was trying to say :) Done

lib/tether/tether_linux.go
caglar10ur: Done

cmd/tether/attach_linux.go
caglar10ur: Done

lib/tether/tether_linux.go
caglar10ur: We do, Start calls the cleanup on its way out and cleanup is the only place where we call this function.

caglar10ur:
rawconn wasn't closing the channel, fixed that too with https://github.com/vmware/vic/pull/2702/commits/0f02d5d4f8bb92d3750750a88870d1a43af06009

pkg/serial/handshake.go
caglar10ur: unless I miss something it's more involved though as only way to cancel the read in rawconn is sending an error over error channel. So we can create a context with deadline in SetReadLine and write timeout err to that channel but that also means that a successful read needs to cancel the context.

To be clear, do you prefer SetReadDeadline to be implemented for this PR or are you OK postponing it to some other day?

caglar10ur:
@hmahmood @hickeng see https://github.com/vmware/vic/pull/2702/commits/03190baf04de59eea95e47db6ae95cc9837eb461 we are not interrupting the read by closing the underlying conn

caglar10ur:
♻️

caglar10ur:
♻️

caglar10ur:
♻️

cmd/tether/attach.go
```diff
@@ -140,26 +179,82 @@ func (t *attachServerSSH) start() error {
...
+		err := fmt.Errorf("attach server is not enabled")
+		log.Error(err)
+		return err
+	}
+
+	log.Debugf("Setting enabled to false")
+	t.Disabled()
+
+	// This context is used by backchannel and we need to cancel it before
+	// trying to obtain the following lock so that backchannel releases it.
```
hickeng: It's not that backchannel holds this lock, but that this lock is held by backchannel's caller and not released until backchannel returns.

cmd/tether/attach.go
```diff
@@ -140,26 +179,82 @@ func (t *attachServerSSH) start() error {
...
+	log.Debugf("Canceling the context")
+	t.cancel()
 
-	if conn != nil {
-		(*conn).Close()
+	log.Debugf("Acquiring the connection lock")
+	t.conn.Lock()
+	if t.conn.conn != nil {
+		log.Debugf("Closing the connection")
+		(*t.conn.conn).Close()
```
hickeng: This _should_ be returning "Close called again on rawconn - squashing" or similar. Close should already have been called as a result of t.cancel().

cmd/tether/attach.go
```diff
@@ -140,26 +179,82 @@ func (t *attachServerSSH) start() error {
...
+
+	// This needs to tick *faster* than the ticker in connection.go on the
+	// portlayer side.  The PL sends the first syn and if this isn't waiting,
+	// alignment will take a few rounds (or it may never happen).
+	ticker := time.NewTicker(10 * time.Millisecond)
+	defer ticker.Stop()
+
+	ch := make(chan struct{}, 1)
+
+	go func() {
```
hickeng: this deserves a comment as to why it cannot go in the same select as HandshakeServer

cmd/tether/attach.go
```diff
@@ -140,26 +179,82 @@ func (t *attachServerSSH) start() error {
...
+		select {
+		case <-ctx.Done():
+			(*conn).Close()
+			close(ch)
```
hickeng: `close(ch)` before `(*conn).Close()`
Otherwise HandshakeServer may return and re-enter before `ch` is closed.

Also - **I'm not sure what happens if you call read on a closed rawconn**

cmd/tether/attach.go
caglar10ur: Done

cmd/tether/attach.go
caglar10ur: Done

hickeng:
LGTM now we've got the backchannel interrupt sorted out.

caglar10ur:
♻️

PR #2703, opened 2016-10-12
TLS aware liveliness check for API

hickeng:
1. Updates documentation to include the new/renamed tls options.
2. Checks for liveliness of the API endpoint in a manner that should
work even when missing a client certificate. This should help in #2624 
3. Attempts to warn in the case of clock skew causing certificate issues.

Towards: #2624 #2197 

**MUST NOT MERGE BEFORE #2707 - --cname changed to --tls-cname**

### DIFF
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -21,9 +21,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
+	"net/url"
+	"os"
 	"path"
 	"strconv"
+	"syscall"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -50,7 +54,12 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/vm"
 )
 
-const portLayerPort = constants.SerialOverLANPort
+const (
+	portLayerPort = constants.SerialOverLANPort
+
+	// this is generated in the crypto/tls.alert code
+	badTLSCertificate = "tls: bad certificate"
+)
 
 var (
 	lastSeenProgressMessage string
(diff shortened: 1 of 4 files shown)


dougm:
lgtm

andrewtchin:
lgtm

lib/install/management/appliance.go
```diff
@@ -716,23 +720,76 @@ func (d *Dispatcher) CheckDockerAPI(conf *config.VirtualContainerHostConfigSpec,
...
 
-		// TODO: add check for rejection without cert - indicates a degree of initialization
-		// HTTP status 403.7. Change this to check for structured error and status codes if possible
-		if tlsErrExpected && err != nil {
-			log.Debugf("Expected TLS error from endpoint due to IP based addressing, skipping component check: %+v", err)
-			// TODO: check for 403.17 - we see that regardless until the appliance has updated
-			// it's clock in some cases
-			break
+		if err != nil {
+			// DEBU[2016-10-11T22:22:38Z] Error recieved from endpoint: Get https://192.168.78.127:2376/info: dial tcp 192.168.78.127:2376: getsockopt: connection refused &{%!t(string=Get) %!t(string=https://192.168.78.127:2376/info) %!t(*net.OpError=&{dial tcp <nil> 0xc4204505a0 0xc4203a5e00})}
```
hmahmood: Do we need this commented out code?

lib/install/management/appliance.go
```diff
@@ -716,23 +720,76 @@ func (d *Dispatcher) CheckDockerAPI(conf *config.VirtualContainerHostConfigSpec,
...
+					case *os.SyscallError:
+						if root.Err == syscall.Errno(syscall.ECONNREFUSED) {
+							// waiting for API server to start
+							log.Debug("connection refused")
+						}
+					default:
+						// the TLS package doesn't expose the raw reason codes
+						// but we're actually looking for alertBadCertificate (42)
+						errmsg := root.Error()
+						if errmsg == "tls: bad certificate" {
```
hmahmood: Should make the string a constant.

lib/install/management/appliance.go
hickeng: It's there as reference for why the code below is doing what it's doing.
Seeing as it took reflect/pretty printing to figure out how to get at the error details it'll be rather opaque without

hickeng:
poke

PR #2706, opened 2016-10-13
Purge prior component launch status during upgrade

### LINKED ISSUE #2624: Upgrade does not wait for components to start

hickeng:
If the Started status is not purged then we don't block or correctly report component
status in the case of failure.
This PR re-enables the blocking on component launch.
We should not purge IP address so we can maintain a stable address if DHCP so we
will not block on that.

In combination with #2703 which blocks waiting for API response:
Fixes #2624

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Project Bonneville was research aimed at determining best approaches to enabling
 Once built, pick up the correct binary based on your OS, and then the result can be installed with the following command.
 
 ```
-bin/vic-machine-linux create --target target-host[/datacenter] --image-store <datastore name> --name <vch-name> --user <username> --password <password> --compute-resource <cluster/a/resource/pool/path>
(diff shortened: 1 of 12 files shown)


dougm:
lgtm

caglar10ur:
LGTM :sheep: :it:

hmahmood:
lgtm

hickeng:
poke

PR #2723, opened 2016-10-13
Stop logging docker registry auth token during pull

jzt:
This change will prevent logging of the docker registry auth tokens if the debug level is not explicitly set to be > 2.

### DIFF
--- a/pkg/fetcher/fetcher.go
+++ b/pkg/fetcher/fetcher.go
@@ -391,7 +391,6 @@ func (u *URLFetcher) setBasicAuth(req *http.Request) {
 
 func (u *URLFetcher) setAuthToken(req *http.Request) {
 	if u.options.Token != nil {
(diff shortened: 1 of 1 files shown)


dougm:
lgtm - tho do we need to log it at all?

caglar10ur:
LGTM

jzt:
@hickeng fixed, although, like @dougm, the more I think about it, the more I wonder if we really need to log it at all.

caglar10ur:
I believe I added that part to the original imagec and it was purely debugging the imagec itself(if you have the token then you can use curl to see what's happening). I have no objections dropping it

hickeng:
@jzt I'm fine with dropping it entirely.

jzt:
Removed, rebased

jzt:
♻️

PR #2733, opened 2016-10-14
Using netlink to get client IP address in toolbox

### LINKED ISSUE #2688: Vch toolbox returns the wrong client interface address
Seen here: https://ci.vmware.run/vmware/vic/5678 (test 8-1-GuestTools)

The problem seems to be that we have `client` as an alias to the `external` interface; `ip link` output on the VCH:
```
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: bridge: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 00:0c:29:a4:82:70 brd ff:ff:ff:ff:ff:ff
3: external: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 00:0c:29:a4:82:7a brd ff:ff:ff:ff:ff:ff
    alias client
```
Usually it is the opposite: external is aliased to client.

A potential fix would be to not use the Go apis to get the interface address, and use the [netlink](https://github.com/vishvananda/netlink) library.

hmahmood:
Fixes #2688

### DIFF
--- a/lib/tether/toolbox.go
+++ b/lib/tether/toolbox.go
@@ -19,14 +19,14 @@ package tether
 import (
 	"errors"
 	"fmt"
-	"net"
+	"os"
 	"sync"
 	"syscall"
 	"time"
 
-	"golang.org/x/crypto/ssh"
-
 	log "github.com/Sirupsen/logrus"
+	"github.com/vishvananda/netlink"
+	"golang.org/x/crypto/ssh"
 
 	"github.com/vmware/vic/cmd/tether/msgs"
 	"github.com/vmware/vic/pkg/vsphere/toolbox"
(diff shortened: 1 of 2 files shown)


anchal-agrawal:
lgtm

andrewtchin:
lgtm

dougm:
lgtm

lib/tether/toolbox.go
```diff
@@ -219,25 +219,30 @@ func (t *Toolbox) halt() error {
 
 // externalIP attempts to find an external IP to be reported as the guest IP
 func (t *Toolbox) externalIP() string {
-	netif, err := net.InterfaceByName("client")
+	l, err := netlink.LinkByName("client")
```
dougm: We should probably consider moving "client" to a constant, as its now referenced 6 times within this routine and likely many places elsewhere in vic.

caglar10ur:
Lgtm

lib/tether/toolbox.go
hickeng: This should *not* be put in lib/portlayer if it's going to be referenced in lib/tether. That will again cause transitive inclusion of a vast amount of dead code.

We should consider moving this `vic-init` specific piece of logic into cmd/vic-init

hickeng:
Opened this: https://github.com/golang/go/issues/17445

lib/tether/toolbox.go
hmahmood: @dougm @hickeng moved code out of toolbox. Will have a follow up PR to make `"client"` a constant.

hmahmood:
:recycle:

PR #2735, opened 2016-10-14
Log thumbprint when target cert cannot be verified

dougm:
If the target cert is not signed by a known CA or command was run
without 'thumbprint' or 'force' flag, log the target and thumbprint.

See #2554


jzt:
LGTM

anchal-agrawal:
lgtm

dougm:
:recycle:

PR #2736, opened 2016-10-14
Not rebooting if debug level > 0 [full ci]

### LINKED ISSUE #1999: VCH should not reboot endlessly on panic
Currently if the vic-init process (which is the default process on the VCH) panics, we reboot the VCH. This is not useful at all because we need to look at the logs for the panic call stack, etc.; these are obliterated on reboot.

We need to:
1. stop rebooting on panic
2. launch vic-admin as a separate (systemd) service so that it is always running on the VCH, even if vic-init crashes

hmahmood:
Fixes #1999

### DIFF
--- a/cmd/vic-init/main_linux.go
+++ b/cmd/vic-init/main_linux.go
@@ -17,7 +17,6 @@ package main
 import (
 	"os"
 	"runtime/debug"
-	"strings"
 	"syscall"
 
 	log "github.com/Sirupsen/logrus"
@@ -29,7 +28,11 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/toolbox"
 )
 
-var tthr tether.Tether
+var (
+	tthr       tether.Tether
+	config     ExecutorConfig
+	debugLevel int
+)
 
 func init() {
 	trace.Logger.Level = log.DebugLevel
@@ -44,6 +47,17 @@ func main() {
 		reboot()
 	}()
 
+	src, err := extraconfig.GuestInfoSourceWithPrefix("init")
+	if err != nil {
+		log.Error(err)
+		return
+	}
+
+	extraconfig.Decode(src, &config)
+
+	debugLevel = config.Diagnostics.DebugLevel
+	setLogLevels()
+
 	logFile, err := os.OpenFile("/dev/ttyS1", os.O_WRONLY|os.O_SYNC, 0644)
 	if err != nil {
 		log.Errorf("Could not pipe stderr to serial for debugging info. Some debug info may be lost! Error reported was %s", err)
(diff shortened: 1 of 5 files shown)


sgairo:
LGTM

pkg/trace/operation_test.go
```diff
@@ -103,8 +107,8 @@ func TestNestedLogging(t *testing.T) {
 
 	// We arrive at 2 because we have the err line (line 0), then the root
 	// (line 11) of where we created the ctx.
-	if !assert.True(t, lines >= levels+2) {
-		t.Logf("exepected %d and got %d", levels+2, lines)
+	if !assert.True(t, lines >= levels) {
```
anchal-agrawal: Can we use `assert.False` instead of `!assert.True`?

anchal-agrawal:
LGTM

pkg/trace/operation_test.go
```diff
@@ -103,8 +107,8 @@ func TestNestedLogging(t *testing.T) {
...
+		t.Logf("exepected %d and got %d", levels, lines)
```
anchal-agrawal: `expected`

cmd/vic-init/main_linux.go
```diff
@@ -44,6 +47,17 @@ func main() {
 		reboot()
 	}()
 
+	src, err := extraconfig.GuestInfoSourceWithPrefix("init")
+	if err != nil {
+		log.Error(err)
```
vburenin: Is it an error that we don't care? Should not it be a warning?

vburenin: I see, it belongs to main, not to init. nvm.

anchal-agrawal:
lgtm

pkg/trace/trace.go
```diff
@@ -16,21 +16,13 @@ package trace
...
 
-var Logger = &log.Logger{
-	Out: os.Stderr,
-	// Apparently this forces terminal handling even when Out isn't a TTY (like
-	// on a VCH, where STDOUT/STDERR are currently redirected to a file).
-	Formatter: &log.TextFormatter{ForceColors: true},
-	Hooks:     make(log.LevelHooks),
-	Level:     log.InfoLevel,
-}
+var Logger = logrus.New()
```
caglar10ur: I think we need @fdawg4l eyes on this as he added this couple of days ago

fdawg4l: Yeah, this breaks our logging.  `logrus` does a syscall on the fd for STDOUT to check if it's a tty and if it is not, it blindly passes escape sequences like line-wrapping.    Yes, really.  If you want to verify, startup a VCH, and tail the port layer log and you'll see all of the `Operation` log messages will include their "\n"s and "\t"s when there is a stack trace.  Ergo the comment in the original version of this file.

fdawg4l: @hmahmood did the unit test pass on drone?  It would fail in very much the same way on the VCH.

fdawg4l: Also why do you need this?

hmahmood:
♻️

Issue #2726, opened 2016-10-14
CI: curl not providing certificates in vic-machine Create Basic tests


hickeng:
We are running raw curl in the Create Basic tests, instead of the `Curl` keyword that @gigawhitlocks added to handle certificates.
```
<msg timestamp="20161013 23:37:57.718" level="INFO">Running command 'curl -sk https://192.168.31.157:2378/container-logs.tar.gz | tar tvzf - 2&gt;&amp;1'.</msg>
<msg timestamp="20161013 23:37:57.768" level="INFO">${rc} = 2</msg>
<msg timestamp="20161013 23:37:57.768" level="INFO">${output} = 
gzip: stdin: unexpected end of file
tar: Child returned status 1
tar: Error is not recoverable: exiting now</msg>
```

The fix requires:
1. move `Curl` to Util.robot from VICAdmin CertAuth
2. update `Curl` to check whether certificates exist (if curl doesn't handle missing certs cleanly)
3. Replace all raw `curl` calls with keyword `Curl` calls except for negative tests in CertAuth

hickeng:
@gigawhitlocks Is this addressed?

hickeng:
@mdubya66 moving this out of 0.7 as @gigawhitlocks has adjusted the specific instances. Still needs to be open to move this function into Utils.

gigawhitlocks:
This is mostly addressed. The `curl` and `wget` calls were adjusted in-place to not break, but `Curl` is not in `Util.robot`. It would make sense to use that instead of custom `curl` and `wget` calls in individual tests, but that's just debt & this issue is not a blocker at this time. As far as functionality is concerned, I have already addressed this.

PR #2744, opened 2016-10-14
Use TLS friendly addresses in inspect output [full ci]

hickeng:
Attempts to determine the appropriate address to use based on the configured certificates.
This is necessary because with certificate validation happening you have to use the appropriate name/IP to connect to the system.

The primary effect here is that the vicadmin URL will be either IP or FQDN as needed

Full CI because anything that changes vic-machine output is fraught with danger.

Towards #2726


hickeng:
@dougm @gigawhitlocks seeing as you've both been messing with cert stuff you're good people to review.

dougm:
lgtm

lib/install/management/dispatcher.go
```diff
@@ -211,41 +211,87 @@ func (d *Dispatcher) CollectDiagnosticLogs() {
...
-		log.Debugf("Checked %s, no match for host cert", n)
+		candidateIPs = append(candidateIPs, ips...)
 	}
 
-	// no viable address
-	return ip, errors.New("unable to determine viable address")
+	// always add all the altname IPs - we're not checking for connectivity
+	candidateIPs = append(candidateIPs, cert.IPAddresses...)
+
+	return addrToUse(candidateIPs, cert, cas)
```
hmahmood: So this resolves the ips back to dns names, while we did the reverse in this function. Why?

cmd/vic-machine/create/create.go
```diff
@@ -861,6 +865,17 @@ func (c *Create) generateCertificates(ca bool) ([]byte, *certificate.KeyPair, er
 		log.Warnf("Failed to stash client certificate for later application level validation: %s", err)
 	}
 
+	// if openssl is present, try to generate a browser friendly pfx file
```
caglar10ur: What is browser friendly cert? Can we add a comment explaining what's going on here?

lib/install/management/appliance.go
```diff
@@ -678,31 +676,51 @@ func (d *Dispatcher) CheckDockerAPI(conf *config.VirtualContainerHostConfigSpec,
...
-			pool := x509.NewCertPool()
-			if !pool.AppendCertsFromPEM(conf.CertificateAuthorities) {
-				log.Debug("Unable to load CAs in config, if any")
-			}
+			// if tlsverify was configured at all then we must verify the remote
+			tr.TLSClientConfig.InsecureSkipVerify = false
 
-			// tr.TLSClientConfig.ClientCAs = pool
-			tr.TLSClientConfig.RootCAs = pool
+			func() {
```
caglar10ur: why do we have this? I don't see any defer's here so I guess we could remove the func(){ / {() parts, no?

cmd/vic-machine/create/create.go
hickeng: Added some additional comments.

caglar10ur:
LGTM

hickeng:
poke

lib/install/management/dispatcher.go
hickeng: We resolve the DNS names to IPs as a filter for which are valid at the location vic-machine is being run. We then use that set of possible IPs along with the certificates to see which addresses pass TLS validation.

PR #2747, opened 2016-10-14
Add block related dirs, dmesg, and system journal to logs bundle

### LINKED ISSUE #2675: vicadmin: collect additional logs
I'm attempting to diagnose a problem with the volume tests in https://ci.vmware.run/vmware/vic/5693

I need the following additional logs to be able to make a proper determination of where the problem is:
1. dmesg/all the journald output
~~2. ls -l /dev/disk/*/*-~~
3. if not in (1) then the systemd-udevd unit output

It's of note that the appliance vmware.log was instrumental in determining that the hotadd had happened in a timely fashion from a hardware perspective - we really should try to keep that in the bundle - worth noting in #2164 

In the course of this I discovered the following in the log bundle should be added or fixed:
1. all file handles, not just /proc/self - no one should be leaking file handles, but...
~~2. contents of /dev/disk (not inside the entries, but all the links)~~
3. process listing (cannot actually tell if there are other unexpected processes running in there currently)
4. systemd log output (mentioned in the initial list)
5. top style snapshot for both Mem and CPU

Update: We've already got the disk-by-label and disk-by-path

fdawg4l:
We need more info when a disk attach times out.  This will help us see
what the kernel sees.

Fixes #2675


fdawg4l:
cc @hickeng since you asked for these.

fdawg4l:
Also @sgairo if you could take a look, I'd appreciate it.  Thanks!

caglar10ur:
LGTM :sheep: :it:

sgairo:
Lorn 👍

fdawg4l:
@hickeng last chance to comment.

fdawg4l:
.

PR #2751, opened 2016-10-15
Add timestamp support back

caglar10ur:


anchal-agrawal:
lgtm

pkg/trace/trace.go
```diff
@@ -27,9 +27,14 @@ var Logger = &log.Logger{
...
-	Hooks:     make(log.LevelHooks),
-	Level:     log.InfoLevel,
+	Formatter: &log.TextFormatter{
+		ForceColors:   true,
+		FullTimestamp: true,
+		// from https://godoc.org/github.com/docker/docker/pkg/jsonlog#pkg-constants
+		TimestampFormat: "2006-01-02T15:04:05.000000000Z07:00",
+	},
+	Hooks: make(log.LevelHooks),
+	Level: log.InfoLevel,
```
hmahmood: In my PR I have changed this to use the standard logger, so this change may not be necessary.

fdawg4l: @caglar10ur is right; it was always INFO.

fdawg4l:
LGTM

caglar10ur:
OK, I think @fdawg4l and @hmahmood  are in the same office today and I'm not sure what will be the effect of going back to regular logger. 

Could you go over the other PR and decide whether we need  this or not? If you think this is no longer needed please feel free to close.

caglar10ur:
Closing this as I believe this is implemented differently

PR #2766, opened 2016-10-17
[skip ci] Replace hardcoded -linux string with OS

caglar10ur:


fdawg4l:
LGTM

hickeng:
LGTM - the quoting is odd, but I'll get over it.

caglar10ur:
♻️

PR #2770, opened 2016-10-17
Add log level change keyword[skip ci]

mhagen-vmware:


dougm:
lgtm

hickeng:
LGTM - should we start splitting the different Util functions into separate files. util/vsphere.robot for example? This file is getting large.

mhagen-vmware:
Yup, I have already started with nimbus-util.robot but I def plan a re-arch once I get some breathing room.

PR #2771, opened 2016-10-17
Network info persistence across appliance restarts [full ci]

### LINKED ISSUE #2448: Networks with no connected containers do not persist across VCH restarts
Currently there is no way to persist a network's info across VCH restarts. The network info is persisted as a map to guestinfo (see the `Configuration.ContainerNetworks` in lib/portlayer/network/config.go). Since serializing a map creates new keys for each of the map's keys, and new keys are not persisted to the vmx, information about a user created network can only be gleaned from containers that are connected to that network. Any networks that do not have containers connected cannot be reconstructed.

This requires some way to persist network info other than guestinfo. A k/v store is in the works #2080 , and is required for this.

hmahmood:
Fixes #2448


tests/test-cases/Group10-VCH-Restart/Group10-VCH-Restart.robot
```diff
@@ -0,0 +1,82 @@
+*** Settings ***
```
emlin: Are we changing the test file name to GroupXXX?

lib/portlayer/network/context.go
```diff
@@ -1058,9 +1113,27 @@ func (c *Context) ContainerByAddr(addr net.IP) *Endpoint {
...
+	var err error
+	if c.kv != nil {
+		op := trace.NewOperation(context.Background(), fmt.Sprintf("deleteing scope %s from kv", name))
+		if err = c.kv.Delete(op, fmt.Sprintf("context.scopes.%s", name)); err != nil && err != kvstore.ErrKeyNotFound {
+			return err
+		}
+	}
+
+	// FIXME: restore scope in kv store if deleteScope() fails
+	if err = c.deleteScope(name); err != nil {
```
emlin: Shouldn't be delete scope first, and then delete from kv store?

pkg/kvstore/kvstore.go
```diff
@@ -32,54 +34,59 @@ var (
...
 }
 
 // Create a new KeyValueStore instance using the given Backend with the given
 // file.  If the file exists on the Backend, it is restored.
-func NewKeyValueStore(op trace.Operation, store Backend, fileName string) (*KeyValueStore, error) {
-	p := &KeyValueStore{
-		b:        store,
-		kv:       make(map[string][]byte),
-		fileName: fileName,
+func NewKeyValueStore(ctx context.Context, store Backend, name string) (KeyValueStore, error) {
```
cgtexmex: Why are you removing the trace.Operation?

lib/portlayer/network/context.go
hmahmood: It could be either way; its just easier to restore the kv store data than it is to restore the scope.

pkg/kvstore/kvstore.go
hmahmood: I think its better to not use it here, and use it in higher level calls instead. It makes this API both compatible with context.Context and trace.Operation (since a trace.Operation can be passed in).

tests/test-cases/Group10-VCH-Restart/Group10-VCH-Restart.robot
hmahmood: I removed this file

pkg/kvstore/kvstore.go
cgtexmex: I think the intent though was to encourage / enforce the use of operation for the value it provides.  If we change the signature then some *might* not use it...just like this change passes a regular context in via the PUT kv_handler.  

That's my take, but possibly @fdawg4l didn't intend that when he created it.

lib/portlayer/store/store.go
```diff
@@ -60,16 +66,26 @@ var (
 func Init(ctx context.Context, session *session.Session, imgStoreURL url.URL) error {
 	defer trace.End(trace.Begin(imgStoreURL.String()))
 
-	mgr = &StoreManager{
-		dsStores:     make(map[string]*kvstore.KeyValueStore),
-		datastoreURL: imgStoreURL,
-	}
-	//create or restore the api accessible datastore backed k/v store
-	return NewDatastoreKeyValue(ctx, session, APIKV)
+	func() error {
```
fdawg4l: Not following why this idiom is preferred.  I feel like I've asked this before so apologies if I have.  A comment at the least would be appreciated.

pkg/kvstore/kvstore.go
fdawg4l: It's really up to you guys if you'd rather use the `Context` interface vs the `Operation` struct.  The interface is fine if all you're interested in is following the mutations of a context across API boundaries.  However, you'd need to be diligent in using the helper functions in the trace package and not the `context` package otherwise the chain is broken.   Also the `Operation` struct has some useful logging features which help you trace an operation by ID when using `Infof`, `Debugf`, and the like.

But for this use case, it's really up to you if you want to use them.  At some point @hickeng and I may go through and replace all instances of `context` with `Operation`.  It's already proven useful in debugging CDEs.

pkg/kvstore/kvstore.go
```diff
@@ -32,54 +34,59 @@ var (
...
-	fileName string
-
-	l sync.RWMutex
+type KeyValueStore interface {
+	Set(ctx context.Context, key string, value []byte) error
+	Get(key string) ([]byte, error)
+	List(re string) (map[string][]byte, error)
+	Delete(ctx context.Context, key string) error
+	Save(ctx context.Context) error
+	Name() string
```
fdawg4l: What is `Name` for?  Comment would be good since the rest, I think, are self explanatory.

pkg/kvstore/kvstore.go
```diff
@@ -128,8 +150,31 @@ func (p *KeyValueStore) Get(op trace.Operation, key string) ([]byte, error) {
...
+
+	regex, err := regexp.Compile(re)
+	if err != nil {
+		return nil, err
+	}
+
+	kv := make(map[string][]byte)
+	for k, v := range p.kv {
+		if regex.MatchString(k) {
+			kv[k] = v
```
fdawg4l: Are the vals being returned a reference or a copy of the val in the store?

lib/portlayer/network/context.go
```diff
@@ -126,18 +131,56 @@ func NewContext(config *Configuration) (*Context, error) {
...
 	}
 
+	// load saved scopes in the kv store
+	if kv != nil {
+		values, err := kv.List(`context\.scopes\..+`)
+		if err != nil && err != kvstore.ErrKeyNotFound {
+			log.Warnf("error listing scopes from key value store: %s", err)
+		} else {
+			for k, v := range values {
+				s := &Scope{
```
vburenin: Would you want to have a constructor for scope?

lib/portlayer/network/context.go
```diff
@@ -176,22 +219,30 @@ func reserveGateway(gateway net.IP, subnet *net.IPNet, ipam *IPAM) (net.IP, erro
...
+	if _, ok := c.scopes[s.name]; ok {
+		return DuplicateResourceError{}
+	}
+
 	var err error
-	var space *AddressSpace
 	var defaultPool bool
 	var allzeros, allones net.IP
+	var space *AddressSpace
+	spaces := s.spaces
```
vburenin: I am not sure I understand what is going on here. Is this assignment for convenience purpose only?

pkg/kvstore/kvstore.go
hmahmood: The values are byte slices, so I assume another slice.

lib/portlayer/network/context.go
hmahmood: Yes I moved this code over from another function, just didn't want to mess with it too much.

lib/portlayer/network/context.go
```diff
@@ -126,18 +131,54 @@ func NewContext(config *Configuration) (*Context, error) {
...
 		}
 
 		s.builtin = true
 	}
 
+	// load saved scopes in the kv store
+	if kv != nil {
+		values, err := kv.List(`context\.scopes\..+`)
+		if err != nil && err != kvstore.ErrKeyNotFound {
+			log.Warnf("error listing scopes from key value store: %s", err)
```
emlin: return here?

emlin:
minor comments. lgtm

pkg/kvstore/kvstore.go
```diff
@@ -32,54 +34,74 @@ var (
 // This package implements a very basic key/value store.  It is up to the
 // caller to provision the namespace.
 
-type KeyValueStore struct {
-	b Backend
+type KeyValueStore interface {
+	// Set adds a new key or modifies an existing key in the key-value store
+	Set(ctx context.Context, key string, value []byte) error
```
hmahmood: Yeah I think `Put` would be better.

andrewtchin:
lgtm

hmahmood:
♻️

PR #2772, opened 2016-10-17
Add reconfigure lock for disk attach detach on the VCH [full ci]

### LINKED ISSUE #2549: Failed to write to image store: [POST /storage/{store_name}][500] WriteImage default  &{Code:0xc42017c910 Message:Another task is already in progress.}
Docker integration tests currently fail with the above error

[4-1-Docker-Integration-VCH-0-1352-container-logs.zip](https://github.com/vmware/vic/files/507051/4-1-Docker-Integration-VCH-0-1352-container-logs.zip)

fdawg4l:
Concurrent `image pull`s from different clients *can* result in a Task Already In Progress error.  Luckily the task package handles this by issuing a retry, but it only does so a small number of times.  This change puts a lock around the reconfigure to avoid the disk package attaching/detaching disks simultaneously to a VCH.

Fixes #2549


emlin:
Just feel weird for this lock. If add this lock for disk manager, do we need to lock other VM reconfig operation?
Even you lock the disk operation, I assume if there is other vm reconfig operation running, you still might have this same issue.

emlin:
Okey, confirmed from @fdawg4l in person.
LGTM

fdawg4l:
Just following up with what we talked about in person-  this is (so far) the only Reconfigure operation on the vch.  It doesn't solve all of the potential Task In Progress errors in the rest of the system if more Reconfigures are required elsewhere, but it at least solves the ones we know about.

matthewavery:
LGTM

fdawg4l:
CI passed.

fdawg4l:
.

PR #2778, opened 2016-10-18
Move bridge code to ensure more tests use unique bridge

mhagen-vmware:
Move the vlan bridge stuff so that the tests not using the Install VIC Machine command will still use a unique bridge as long as they use Set Test Environment Variables at least...


anchal-agrawal:
LGTM

rajanashok:
lgtm

mhagen-vmware:
.

PR #2780, opened 2016-10-18
Correct primary IP logic [full ci]

### LINKED ISSUE #2777: vic-init cannot locate the client interface
There are some issues with #2733:
1. There are two error messages logged that are identical - no way to separate which you hit
2. The error check for LinkByName is incorrect - LinkByName does not return an error that IsNotExist will recognise:

Flawed error check:
https://github.com/vmware/vic/commit/3c2aec045f817063893408369fb25267550a0c75#diff-4294f1834b8a06be95501b213d755411R128
Error source:
https://github.com/vmware/vic/blob/3c2aec045f817063893408369fb25267550a0c75/vendor/github.com/vishvananda/netlink/link_linux.go#L841

hickeng:
Logic for obtaining the primary IP to report to vsphere (that assoicated
with the client interface) was incorrectly handling errors related to
LinkByName when client is an alias

Fixes #2777


gigawhitlocks:
LGTM

mdubya66:
lgtm

hickeng:
poke

PR #2781, opened 2016-10-18
Remove vmware.log from applianceVM (Vic Admin)

### LINKED ISSUE #2164: vic-admin: strip vmware.log

sgairo:
Fixes #2164


sgairo:
@jzt @hickeng

dougm:
Why not just remove vmware.log from the list? https://github.com/vmware/vic/blob/e08c84dbe6ecb7a925f7bb41be6ee37822d6a216/cmd/vicadmin/vicadm.go#L67

Maybe we can still include it if debug is enabled?  @hickeng ?  If not, can we change the robot helper to grab the log via govc datastore.download and add it to the log bundle.zip ?

jzt:
@dougm That will also remove vmware.log from the containerVM bundles.

dougm:
@jzt gotcha, disregard.

hickeng:
I was stripping it in this WIP: https://github.com/vmware/vic/pull/2784 and there's a big TODO about not using name comparison because it sucks. The WIP includes the changes necessary for GetSelf so we could use that for much more reliable comparison.
@dougm +1 for logging it with higher debug levels enabled (debug > 2 IMO as it's a pretty serious behavioural change)

jzt:
@hickeng It looks like your changes take care of this (minus the debug level check to allow for inclusion of the log). Should @sgairo close this PR and move on, or were you thinking something else?

PR #2787, opened 2016-10-18
Fix formatting for Operation CDE stack traces

fdawg4l:
The text formatter in logrus escapes CR and TAB characters when Out
isn't a TTY.  The only bool for disabling that check is forcing colors,
which adds extra garbage (color codes) to log lines.

Fix is to include our own minimal text formatter that doesn't escape
characters.


fdawg4l:
@hmahmood and @caglar10ur, if you guys could take 👀 , I'd appreciate it.

hmahmood:
Does this mean we don't get the nice color output when we are in the terminal? It would be nice if we could keep that for when we run in a terminal.

Also, should we use the same formatter in other places in the code? We use the global logger instance from logrus in other places, and that may cause some inconsistency in our log outputs.

pkg/trace/trace.go
```diff
@@ -16,13 +16,21 @@ package trace
...
 )
 
-var Logger = logrus.New()
+var Logger = &logrus.Logger{
+	Out: os.Stderr,
+	// We're using our own text formatter to skip the \n and \t escaping logrus
+	// was doing on non TTY Out (we redirect to a file) descriptors.
+	Formatter: &TextFormatter{},
+	Hooks:     make(logrus.LevelHooks),
+	Level:     logrus.InfoLevel,
```
caglar10ur: Do we need https://github.com/vmware/vic/pull/2751/files#diff-6b636c14743b03d96931844d09206b91R32?

caglar10ur:
LGTM

pkg/trace/trace.go
fdawg4l: Good call.  Thanks!

fdawg4l:
@hmahmood 

> Does this mean we don't get the nice color output when we are in the terminal? It would be nice if > we could keep that for when we run in a terminal.

This is just the logger used in the `trace` pkg.  The global logger is unaffected.  But yes, no colors on the console.  And when do we have a mode where we use the console to watch logs?  The only way we can read logs is via vicadmin which itself doesnt have a console to interpret colors.  Do you mean unit test output?  Even then, that's of marginal usefullness. 

However.  I have a change im going to open a PR for to `logrus` directly which will make this change redundant.  I just want this in before GA in case the change upstream takes too long.

> Also, should we use the same formatter in other places in the code? We use the global logger > instance from logrus in other places, and that may cause some inconsistency in our log  outputs.

I don't think we're using `\t` or `\n` anywhere else in the code so formatting shouldn't be an issue.

fdawg4l:
I opened (this)[https://github.com/Sirupsen/logrus/pull/432] against `logrus` but looking at the open PRs and how many there are, I'm not very confident these changes will make it in any time soon.  So if the `TextFormatter` needs to be used by more than just `trace`, then I'll create a `pkg` for it and we can reference the formatter from there.

fdawg4l:
Re: integrating this text formatter elsewhere, I'd rather do that in a different PR.

hmahmood:
lgtm

PR #2789, opened 2016-10-19
[full ci] Rearranged how lock is acquired to change container state + use atomics to change/read container state.

### LINKED ISSUE #2773: personality becomes unresponsive if an ongoing call blocks
Personality stops responding queries if an operation blocks (let's say ContainerStart)

vburenin:
Rearranged how lock is acquired to change container state + use atomics to change/read container state.

Fixes #2773


lib/portlayer/exec/exec.go
```diff
@@ -84,9 +84,6 @@ func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSou
 		// create the event manager &  register the existing collector
 		Config.EventManager = event.NewEventManager(ec)
 
-		// subscribe the exec layer to the event stream for Vm events
-		Config.EventManager.Subscribe(events.NewEventType(vsphere.VMEvent{}).Topic(), "exec", eventCallback)
```
hmahmood: Without this we are not going to get events from vsphere about out of band power offs, etc.

lib/portlayer/exec/container.go
```diff
@@ -157,23 +159,15 @@ func (s State) String() string {
...
+	return State(v)
 }
 
-func (c *Container) updateState(s State) State {
+func (c *Container) swapState(s State) State {
 	log.Debugf("Setting container %s state: %s", c.ExecConfig.ID, s)
 	prevState := c.state
 	if s != c.state {
-		c.state = s
+		atomic.StoreInt32((*int32)(&c.state), int32(s))
```
hmahmood: This seems out of place here. The other elements of Container are not protected at all, so one would assume there are no locks acquired when `swapState` is called. That may well be the case, and if it is, please see if you can just replace this with assignment.

lib/portlayer/exec/exec.go
```diff
@@ -113,11 +113,14 @@ func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSou
 func eventCallback(ie events.Event) {
 	// grab the container from the cache
 	container := Containers.Container(ie.Reference())
+
 	if container != nil {
+		container.m.Lock()
```
hmahmood: Do we need this?

vburenin: yes. eventCallback may be called asynchronously wile container may be running another logic that could change the state. So, to insure there is no concurrent change of the state I need this lock here. This function doesn't run for a long time, so it will not block forever, just for several hundred nanoseconds.

hmahmood: `swapState` already has a lock in there for the map, and you are using the atomic functions to change the state, so given that, is the lock necessary?

vburenin: just take a look at eventCallback

```
newState := eventedState(ie.String(), container.state)
if newState != container.state {
....
    container.swapState(newState)
}
```

I would like to keep those three operations protected, so when a new state is set it is guaranteed that previous state hasn't changed.

vburenin: atomic functions guarantee an atomic read/write as well as that compiler will not cache a value in a register. but it doesn't guarantee a consistency between three different reads/writes.

hmahmood: Does this re-introduce the deadlock? During `Container.Commit` we hold on to the lock during vsphere operations. If this callback happens during one of those operations, and the operation stretches long), won't we deadlock like before?

Also we have not addressed @caglar10ur original source of the problem: caching the state of the container in the handle.

vburenin: all events are executed in a separate goroutine, so no deadlock should happen unless commit calls that eventedState directly.

hmahmood: Yes this will be blocked momentarily until the Commit exits.

However, other calls like NewHandle will also be blocked by the lock in Commit; which is what @caglar10ur was seeing originally.

vburenin: I guess we are not talking about deadlock, we are talking about long lasting lock.

lib/portlayer/exec/container.go
```diff
@@ -157,28 +160,22 @@ func (s State) String() string {
...
-	return c.updateState(s)
+	v := atomic.LoadInt32((*int32)(&c.state))
+	return State(v)
 }
 
-func (c *Container) updateState(s State) State {
+func (c *Container) swapState(s State) State {
 	log.Debugf("Setting container %s state: %s", c.ExecConfig.ID, s)
+	c.stateEventMu.Lock()
 	prevState := c.state
```
caglar10ur: You could use atomic.CompareAndSwap here :)

lib/portlayer/exec/exec.go
```diff
@@ -113,11 +113,15 @@ func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSou
 func eventCallback(ie events.Event) {
 	// grab the container from the cache
 	container := Containers.Container(ie.Reference())
+
 	if container != nil {
+		// container should be locked to avoid unexpected state change.
```
caglar10ur: Shouldn't you grabbing the lock before checking it against nil?

caglar10ur:
LGTM

lib/portlayer/exec/exec.go
caglar10ur: Ah, morning coffee :)

caglar10ur:
Should we drop this PR because of #2820?

caglar10ur:
closing this as #2820 already merged in

PR #2795, opened 2016-10-19
Unmap previously mapped ports left behind after exiting attached container

### LINKED ISSUE #2519: Port not available after removing a container with the same port that had been attached and exited
Here is my action sequence:

1. docker -H 192.168.60.130:2375 create -it -p 1234:1234 busybox
2. docker -H 192.168.60.130:2375 start container_id
3. docker -H 192.168.60.130:2375 attach container_id
4. exit from container
5. docker -H 192.168.60.130:2375 stop container_id
6. docker -H 192.168.60.130:2375 rm container_id
7. docker -H 192.168.60.130:2375 create -it -p 1234:1234 busybox
8. docker -H 192.168.60.130:2375 start new_container_id

**Expected behavior:**
A new container with new_container_id should be started.

**Actual behavior:** (this is reproducible)

> > Error response from daemon: Server error from portlayer: error mapping ports: port 1234 is not available
Error: failed to start containers: new_container_id

[container-logs (1).zip](https://github.com/vmware/vic/files/501238/container-logs.1.zip)

**Note:**
This issue does not occur if I do not attach to and then exit from the first container.

andrewtchin:
Fixes #2519 

After attaching and exiting a container with mapped port(s) we would previously not unmap the ports.
Now when starting a container that maps ports, if the ports are not use (container that maps them is not running), unmap them and allow the starting container to map them.


lib/apiservers/engine/backends/container.go
```diff
@@ -415,6 +417,41 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
...
+			cc := cache.ContainerCache().GetContainer(mappedCtr)
+			running, err := c.containerProxy.IsRunning(cc)
+			if err != nil {
+				return fmt.Errorf("Failed to get container %q power state: %s", mappedCtr, err)
+			}
+			if !running {
+				// unmap ports if not powered on
+				log.Debugf("Unmapping port %s for powered off container %q", thePort, mappedCtr)
+				if len(endpoints) < 1 {
+					return fmt.Errorf("Failed to unmap port %s for container %q: no endpoints", thePort, mappedCtr)
```
hmahmood: Is this necessary to unmap? If we are keeping track of the iptables commands, we don't really need endpoint info about the container.

lib/apiservers/engine/backends/container.go
```diff
@@ -415,6 +417,41 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 	return nil
 }
 
+// checkActivePortBindings gets port bindings for the container and
+// unmaps ports if the cVM that previously bound them isn't powered on
+func (c *Container) checkActivePortBindings(vc *viccontainer.VicContainer, endpoints []*models.EndpointConfig) error {
+	for p, _ := range vc.HostConfig.PortBindings {
+		thePort := p.Port()
+		if _, mapped := containerByPort[thePort]; mapped {
+			mappedCtr := containerByPort[thePort]
```
vburenin: Why do you read this value twice? you could assign it in if statent where you actually ignoring it.

lib/apiservers/engine/backends/container.go
```diff
@@ -415,6 +417,41 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 	return nil
 }
 
+// checkActivePortBindings gets port bindings for the container and
+// unmaps ports if the cVM that previously bound them isn't powered on
+func (c *Container) checkActivePortBindings(vc *viccontainer.VicContainer, endpoints []*models.EndpointConfig) error {
+	for p, _ := range vc.HostConfig.PortBindings {
+		thePort := p.Port()
+		if _, mapped := containerByPort[thePort]; mapped {
```
hmahmood: This code block has too many indents :-) Structuring like this can help:
```go
if _, mapped := containerByPort[thePort]; !mapped {
  continue
}
```

lib/apiservers/engine/backends/container.go
```diff
@@ -415,6 +417,41 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
...
+			mappedCtr := containerByPort[thePort]
+			log.Debugf("Container %q maps port %s", mappedCtr, thePort)
+			// check state of the previously bound container with PL
+			cc := cache.ContainerCache().GetContainer(mappedCtr)
+			running, err := c.containerProxy.IsRunning(cc)
+			if err != nil {
+				return fmt.Errorf("Failed to get container %q power state: %s", mappedCtr, err)
+			}
+			if !running {
```
hmahmood:
```go
if running {
  log.Debugf(...)
  continue
}
```

lib/apiservers/engine/backends/container.go
```diff
@@ -479,6 +516,11 @@ func (c *Container) containerStart(name string, hostConfig *containertypes.HostC
 			}
 		}()
 
+		// unmap previously mapped ports if they're not powered on
+		err = c.checkActivePortBindings(vc, endpoints)
```
hmahmood: This is the wrong set of `endpoints`. The current container does not have any port bindings yet.

lib/apiservers/engine/backends/container.go
```diff
@@ -584,12 +630,36 @@ func (c *Container) mapPorts(op portmap.Operation, hostconfig *containertypes.Ho
 				}
 			}
 
+			// check if we should actually unmap based on current mappings
```
hmahmood: This whole code block that accesses `containerByPort` needs to be protected with a lock. Maybe even acquire the lock for the duration of the for loop.

lib/apiservers/engine/backends/portmap/portmap.go
```diff
@@ -108,6 +108,32 @@ func (p *portMapper) MapPort(op Operation, ip net.IP, port int, proto string, de
...
+	}
+	return nil
+}
+
+// iptablesDelete takes the saved args from the Append operation
+// and uses them to delete the previously added rules
+func iptablesDelete(args [][]string) error {
+	var errs []error
+	for _, cmd := range args {
+		cmd[2] = string(iptables.Delete) // change saved action from Append to Delete
```
vburenin: How confident are you there are at least three arguments in cmd? It seems very fragile.

lib/apiservers/engine/backends/portmap/portmap.go
```diff
@@ -117,52 +143,68 @@ func (p *portMapper) forward(action iptables.Action, ip net.IP, port int, proto,
...
-	} else if len(output) != 0 {
-		return iptables.ChainError{Chain: "FORWARD", Output: output}
-	}
-
 	ipStr := ""
 	if ip != nil && !ip.IsUnspecified() {
 		ipStr = ip.String()
 	}
 
+	key := bindKey{ipStr, port}
```
vburenin: Could you add key names for each parameter?

lib/apiservers/engine/backends/portmap/portmap.go
```diff
@@ -117,52 +143,68 @@ func (p *portMapper) forward(action iptables.Action, ip net.IP, port int, proto,
...
+			"-p", proto,
+			"-d", daddr,
+			"--dport", strconv.Itoa(port),
+			"-j", "DNAT",
+			"--to-destination", net.JoinHostPort(destAddr, strconv.Itoa(destPort))}
+		if err := iptablesRunAndCheck(args); err != nil {
+			return err
+		}
+		savedArgs = append(savedArgs, args)
+		p.bindings[key] = savedArgs
```
hmahmood: You can leave off the action arg (move it to the beginning of the slice so you can hack it off), and append it back on later when running delete.

lib/apiservers/engine/backends/container.go
andrewtchin: fixed

lib/apiservers/engine/backends/container.go
```diff
@@ -415,6 +417,41 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
...
+	for p, _ := range vc.HostConfig.PortBindings {
+		thePort := p.Port()
+		if _, mapped := containerByPort[thePort]; mapped {
+			mappedCtr := containerByPort[thePort]
+			log.Debugf("Container %q maps port %s", mappedCtr, thePort)
+			// check state of the previously bound container with PL
+			cc := cache.ContainerCache().GetContainer(mappedCtr)
+			running, err := c.containerProxy.IsRunning(cc)
+			if err != nil {
+				return fmt.Errorf("Failed to get container %q power state: %s", mappedCtr, err)
```
andrewtchin: fixed

lib/apiservers/engine/backends/container.go
andrewtchin: fixed

lib/apiservers/engine/backends/container.go
andrewtchin: fixed

lib/apiservers/engine/backends/container.go
andrewtchin: fixed

lib/apiservers/engine/backends/container.go
andrewtchin: fixed

lib/apiservers/engine/backends/portmap/portmap.go
andrewtchin: fixed

lib/apiservers/engine/backends/portmap/portmap.go
andrewtchin: fixed

lib/apiservers/engine/backends/container.go
```diff
@@ -68,13 +68,16 @@ var (
 		scope string
 	}
 
-	portMapper portmap.PortMapper
+	portMapper      portmap.PortMapper
+	containerByPort map[string]string // port:containerID
+	cbpLock         sync.RWMutex
```
anchal-agrawal: https://dmitri.shuralyov.com/idiomatic-go#mutex-hat :)

hmahmood: how about 
```go
cpb struct {
  sync.Mutex
  ports map[string]string
}
```

lib/apiservers/engine/backends/container.go
andrewtchin: fixed

lib/apiservers/engine/backends/container.go
andrewtchin: fixed

lib/apiservers/engine/backends/container.go
```diff
@@ -70,11 +70,15 @@ var (
...
 
+	cbpLock         sync.Mutex
+	containerByPort map[string]string // port:containerID
+
 	ctx = context.TODO()
 )
 
 func init() {
 	portMapper = portmap.NewPortMapper()
+	containerByPort = make(map[string]string)
```
vburenin: Question: Is it really an information that we would like to keep "global"? I would prefer to see data that belongs to a container inside container.

lib/apiservers/engine/backends/container.go
```diff
@@ -415,6 +419,43 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
...
+				mappedCtr, err)
+		}
+		if running {
+			log.Debugf("Running container %q still holds port %s", mappedCtr, thePort)
+			continue
+		}
+
+		log.Debugf("Unmapping port %s for powered off container %q",
+			thePort, mappedCtr)
+		err = c.mapPorts(portmap.Unmap, cc.HostConfig, &models.EndpointConfig{}, mappedCtr)
```
vburenin: mapPorts is a relatively confusing name in this context. MapPorts supposed to map ports not unmap them, right? It looks like a small restructuring needed.

andrewtchin: mapPorts takes an operation that can be map or unmap

lib/apiservers/engine/backends/container.go
```diff
@@ -415,6 +419,44 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 	return nil
 }
 
+// checkActivePortBindings gets port bindings for the container and
+// unmaps ports if the cVM that previously bound them isn't powered on
+func (c *Container) checkActivePortBindings(vc *viccontainer.VicContainer) error {
```
hmahmood: `cleanupPortBindings` might be more appropriate. This name suggests no changes are made to the system.

lib/apiservers/engine/backends/container.go
```diff
@@ -415,6 +419,44 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
...
+	for ctrPort, hostPorts := range vc.HostConfig.PortBindings {
+		for _, hostPort := range hostPorts {
+			hPort := hostPort.HostPort
+
+			cbpLock.Lock()
```
hmahmood: You need to grab the lock for the duration of the outer for loop. This map is changed elsewhere when a port binding is created or removed, so the unmap code below could potentially clash with that.

hmahmood: Actually nvm

lib/apiservers/engine/backends/container.go
```diff
@@ -537,59 +585,121 @@ func requestHostPort(proto string) (int, error) {
...
+
+		// update mapped ports
+		containerByPort[p.strHostPort] = containerID
+		log.Debugf("mapped port %s for container %s", p.strHostPort, containerID)
 	}
+	return nil
+}
 
+// unmapPorts unmaps ports defined in hostconfig for containerID
+func (c *Container) unmapPorts(hostconfig *containertypes.HostConfig, containerID string) error {
```
hmahmood: `conatinerID` is not used by the function at all other than for debugging statements. Can we take it out?

lib/apiservers/engine/backends/portmap/portmap.go
```diff
@@ -101,11 +93,49 @@ func (p *portMapper) MapPort(op Operation, ip net.IP, port int, proto string, de
...
+		return err
+	} else if len(output) != 0 {
+		return iptables.ChainError{Chain: "FORWARD", Output: output}
+	}
+	return nil
+}
+
+// iptablesDelete takes the saved args from the Append operation
+// and uses them to delete the previously added rules
+func iptablesDelete(args [][]string) error {
```
hmahmood: Ah ok, nvm.

lib/apiservers/engine/backends/portmap/portmap.go
```diff
@@ -117,52 +147,67 @@ func (p *portMapper) forward(action iptables.Action, ip net.IP, port int, proto,
...
+	key := bindKey{ip: ipStr, port: port}
 	switch action {
-	case iptables.Append:
-		p.bindings[bindKey{ipStr, port}] = nil
-
 	case iptables.Delete:
-		delete(p.bindings, bindKey{ipStr, port})
-	}
+		val := p.bindings[key] // lookup commands to reverse
+		if args, ok := val.([][]string); ok {
```
hmahmood: You could just change the `bindings` to `map[bindKey][][]string`

lib/apiservers/engine/backends/container.go
andrewtchin: fixed

lib/apiservers/engine/backends/container.go
andrewtchin: fixed

lib/apiservers/engine/backends/portmap/portmap.go
andrewtchin: fixed

vburenin:
LGTM

andrewtchin:
♻️

Issue #2798, opened 2016-10-19
Documentation needs to be updated now that imagec is a library


jzt:
Since imagec is no longer a binary, some of the documentation describing its function and purpose needs to be updated to reflect this. This may include modifying the imagec design doc, or creating one if it doesn't exist.

PR #2800, opened 2016-10-19
Remove remaining references to imagec

jzt:
This removes imagec targets and references from the Makefile, as well as the iso scripts and log bundles. Documentation still needs updating, captured in #2798.


andrewtchin:
lgtm

anchal-agrawal:
:shipit:

PR #2808, opened 2016-10-19
Replace some Info level calls with Debug

caglar10ur:


dougm:
lgtm

Why the 'Debug' -> 'Debugf' changes though?

fdawg4l:
If only we had component logging.  :-)

LGTM

caglar10ur:
@dougm that could be my search/replace operation, could change them back if you think it matters?

dougm:
@caglar10ur ok doesn't matter much, just avoids scanning for the format.. but likely little difference here

caglar10ur:
♻️

caglar10ur:
Removed the debug -> debugf conversion

PR #2809, opened 2016-10-19
Check for and return existing image blob [full ci]

### LINKED ISSUE #2782: docker concurrent pull same image failed
This issue happens in merge build the first time. Here is the build 
https://ci.vmware.run/vmware/vic/5991

jzt:
This change will allow `CreateImageConfig` to return an image blob if one associated with the given layerID already exists. This will prevent unnecessary calculation of an imageID for the same image with a different tag. It will also allow the repoCache to update with new tags for an existing image.

Fixes #2782

### DIFF
--- a/lib/imagec/imagec.go
+++ b/lib/imagec/imagec.go
@@ -325,12 +325,16 @@ func (ic *ImageC) WriteImageBlob(image *ImageWithMeta, progressOutput progress.O
 // CreateImageConfig constructs the image metadata from layers that compose the image
 func (ic *ImageC) CreateImageConfig(images []*ImageWithMeta) (metadata.ImageConfig, error) {
 
-	if len(images) == 0 {
-		return metadata.ImageConfig{}, nil
+	imageLayer := images[0] // the layer that represents the actual image
+
+	// if we already have an imageID associated with this layerID, we don't need
+	// to calculate imageID and can just grab the image config from the cache
+	id := cache.RepositoryCache().GetImageID(imageLayer.ID)
+	if image, err := cache.ImageCache().GetImage(id); err == nil {
+		return *image, nil
 	}
 
 	manifest := ic.ImageManifest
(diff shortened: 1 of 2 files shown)


dougm:
lgtm

lib/imagec/imagec.go
```diff
@@ -325,12 +325,16 @@ func (ic *ImageC) WriteImageBlob(image *ImageWithMeta, progressOutput progress.O
...
 func (ic *ImageC) CreateImageConfig(images []*ImageWithMeta) (metadata.ImageConfig, error) {
 
-	if len(images) == 0 {
-		return metadata.ImageConfig{}, nil
+	imageLayer := images[0] // the layer that represents the actual image
+
+	// if we already have an imageID associated with this layerID, we don't need
+	// to calculate imageID and can just grab the image config from the cache
+	id := cache.RepositoryCache().GetImageID(imageLayer.ID)
+	if image, err := cache.ImageCache().GetImage(id); err == nil {
```
fdawg4l: Is it possible to have an ID in the cache but no backing image in the PL?  If it is, then this check will prevent the user from rectifying the issue by issuing the `pull` again.   I'm hoping it's impossible.

fdawg4l:
LGTM

lib/imagec/imagec.go
jzt: I don't believe this can ever happen in any particular VIC codepath. However, I do not believe there currently to be any way of knowing if a user has deleted layer/image data from the datastore out-of-band. I don't think we are currently publishing any events when this happens, but I'm pinging @cgtexmex in case he knows for sure.

jzt:
@dougm, @fdawg4l Just a heads up: I removed the uncommented integration test for concurrent image pull as this doesn't actually fix the problem like I thought. These changes are needed though, so I'm merging them.

PR #2810, opened 2016-10-19
Fix wait for container so that it does something again

mhagen-vmware:
Running is not a state anymore. This was causing problems in longevity.


cgtexmex:
lgtm

dougm:
lgtm

mhagen-vmware:
.

PR #2826, opened 2016-10-20
Improve portlayer's API session handling [full ci]

### LINKED ISSUE #2398: portlayer is not closing the api session it creates
hence restarting it over and over renders ESXi useless as hostd starts to throw "Out of HTTP sessions: reached the limit of 500 sessions" errors

@dougm any idea whether we can do this in govmomi automagically?

anchal-agrawal:
This PR

* configures the shutdown func for the PL to account for the session created during startup. If the PL panics during initialization or is restarted by vic-init, the session will now be logged out.
* re-uses the session that is passed in to the storage handler instead of creating a new one.

Fixes #2398

### DIFF
--- a/lib/apiservers/portlayer/restapi/configure_port_layer.go
+++ b/lib/apiservers/portlayer/restapi/configure_port_layer.go
@@ -89,11 +89,9 @@ func configureAPI(api *operations.PortLayerAPI) http.Handler {
 	api.ServerShutdown = func() {
 		log.Infof("Shutting down port-layer-server")
 
-		if sess != nil {
-			// Logout the session
-			if err := sess.Logout(ctx); err != nil {
-				log.Warnf("unable to log out of session: %s", err)
-			}
+		// Logout the session
+		if err := sess.Logout(ctx); err != nil {
+			log.Warnf("unable to log out of session: %s", err)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


emlin:
lgtm 👍

dougm:
lgtm :sheep: :it:

lib/apiservers/portlayer/restapi/configure_port_layer.go
```diff
@@ -85,6 +85,18 @@ func configureAPI(api *operations.PortLayerAPI) http.Handler {
 		log.Fatalf("configure_port_layer ERROR: %s", err)
 	}
 
+	// Configure the func invoked if the PL panics or is restarted by vic-init
+	api.ServerShutdown = func() {
+		log.Infof("Shutting down port-layer-server")
+
+		if sess != nil {
```
anchal-agrawal: Good point. I added the nil check to be cautious during a panic.

vburenin: I think it is redundant. Session can not be nil if Create doesn't return an error.

anchal-agrawal: Fixed!

caglar10ur:
LGTM :sheep: :it:

PR #2829, opened 2016-10-20
Do less work on info level for tracing.

vburenin:
What really worries me that even though we can run in info mode, trace logic is still active. It does string formatting, it takes stack snapshots that is pretty expensive especially if we are talking about tether which logs a lot on a hot path if application has a lot to write to stderr and stdout.

A added ability to disable tracing if needed even if debug level is active, however, it is not used so far.


pkg/trace/trace.go
```diff
@@ -24,6 +23,18 @@ import (
 	"github.com/vmware/vic/pkg/log"
 )
 
+var tracingEnabled = true
```
caglar10ur: I think we need to protect this guy from concurrent acces

vburenin:
@caglar10ur I will take a look into Makefile how to do that. However, maybe in other PR?

jzt:
I agree with @caglar10ur re: debug & production versions of trace.go.

Otherwise, LGTM 👍

vburenin:
LGTM

vburenin:
LGTM

PR #2834, opened 2016-10-20
Layer/image data cleanup, image config refactoring [full ci]

### LINKED ISSUE #2530: IDs returned by docker images doesn't match VIC vs vanilla docker

jzt:
This change will stop image metadata from overwriting the metadata belonging to the associated leaf layer. Instead, the image metadata is now persisted in the K/V store on the portlayer. The layer cache has been brought up to speed, and will now persist in the portlayer's k/v store, bringing us finally to the point where layer metadata is actually immutable (nothing is changing it at least). Both caches will be rebuilt after a persona restart.

We could go one step further and get rid of all the portlayer storage functionality that stores metadata and keep it in the portlayer k/v store as well, leaving only vmdks in the layer directories, making them truly immutable. I'll let @fdawg4l marinate on that one for now.

This also enforces the path through `CreateImageConfig`, which will always produce an image ID, allowing for the `updateRepoCache` path to be moved into the leaf layer download rather than outside in imagec. This fixes the race condition that causes #2782.

Fixes #2530, #2782.


lib/apiservers/engine/backends/cache/layer_cache.go
```diff
@@ -15,18 +15,27 @@
...
 // It is used primarily by imagec when coordinating layer downloads
 // The cache is initially hydrated by way of the image cache at startup
 type LCache struct {
 	m sync.RWMutex
-	// layers maps from layer ID to boolean
-	// true indicates that the layer is currently downloading
-	// false indicates that the layer has been downloaded and written to the portlayer
-	layers map[string]bool
+	// Layers maps from layer id to layer metadata
+	Layers map[string]*metadata.LayerConfig
```
jzt: I'm actually going to ditch the metadata.LayerConfig and replace it with the ImageWithMeta struct we use in imagec.

jzt:
Found an issue, marking WIP for now.

jzt:
@cgtexmex, @vburenin I added a `WaitGroup` around the cache updates. @hickeng please sound off if you think this is a bad idea.

@mhagen-vmware I added a test around re-pulling a deleted image.

lib/apiservers/engine/backends/backends.go
```diff
@@ -93,44 +94,86 @@ func Init(portLayerAddr, product string, config *config.VirtualContainerHostConf
...
+	}
+	log.Debugf("New registry service with options %#v", serviceOptions)
+	RegistryService = registry.NewService(serviceOptions)
+
+	return nil
+}
+
+func hydrateCaches() error {
+	wg := sync.WaitGroup{}
+	wg.Add(4)
```
matthewavery: 4 seems like a magic number. Possibly make this a private constant?

matthewavery: seems like a good place to ask this question as well. Is there any chance that any of the `wg.Done()` might not get called? Is there a way to attach timeouts to some of those goroutines just in case they get stuck?

lib/imagec/download.go
```diff
@@ -154,15 +154,16 @@ func (ldm *LayerDownloader) DownloadLayers(ctx context.Context, ic *ImageC) erro
 				defer topDownload.Transfer.Release(watcher)
 
 				ldm.registerDownload(topDownload)
-				cache.LayerCache().AddNew(id)
+				layer.Downloading = true
+				LayerCache().Add(layer)
```
jzt: Done, also cleaned up some logging consistency issues and ordered the goroutines from layers -> image -> repo -> container, for aesthetic purposes.

lib/apiservers/engine/backends/backends.go
jzt: Done, also cleaned up some logging consistency issues and ordered the goroutines from layers -> image -> repo -> container, for aesthetic purposes.

lib/imagec/download.go
jzt: Good catch! Fixed

matthewavery:
lgtm - great PR :)

lib/apiservers/engine/backends/backends.go
```diff
@@ -92,44 +93,89 @@ func Init(portLayerAddr, product string, config *config.VirtualContainerHostConf
...
+
 	log.Info("Creating image store")
 	if err := createImageStore(); err != nil {
 		log.Errorf("Failed to create image store")
 		return err
 	}
 
+	serviceOptions := registry.ServiceOptions{}
+	for _, registry := range insecureRegs {
+		insecureRegistries = append(insecureRegistries, registry.Path)
```
fdawg4l: Ignore if we're logging this somewhere else, but a log of which registries are added would be useful.  Really anywhere where we set up hosts, paths, etc during startup is useful to know during runtime.

lib/apiservers/engine/backends/backends.go
```diff
@@ -92,44 +93,89 @@ func Init(portLayerAddr, product string, config *config.VirtualContainerHostConf
 	// the vic-machine installer timeout will intervene if this blocks for too long
 	pingPortLayer()
 
+	if err := hydrateCaches(); err != nil {
+		return err
```
jzt: All conditions should be fatal, and the first cache refresh that fails will be the one that is returned here (see line 174)

lib/imagec/imagec.go
```diff
@@ -233,9 +235,15 @@ func (ic *ImageC) LayersToDownload() ([]*ImageWithMeta, error) {
...
 				Store:  ic.Storename,
 			},
-			meta:   history.V1Compatibility,
-			layer:  layer,
-			diffID: "",
+			Meta:  history.V1Compatibility,
+			Layer: layer,
+		}
+
+		// Add the diff ID if this layer already exists
```
jzt: This comment no longer makes sense, will update.

lib/apiservers/engine/backends/backends.go
```diff
@@ -92,44 +93,89 @@ func Init(portLayerAddr, product string, config *config.VirtualContainerHostConf
...
 
-	serviceOptions := registry.ServiceOptions{}
-	for _, registry := range insecureRegs {
-		insecureRegistries = append(insecureRegistries, registry.Path)
-	}
-	if len(insecureRegistries) > 0 {
-		serviceOptions.InsecureRegistries = insecureRegistries
+	for err := range errors {
+		if err != nil {
+			return err
```
emlin: might be better to return all errors

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -196,8 +188,13 @@ func prefixImageID(imageID string) string {
 	return "sha256:" + imageID
 }
 
-// AddImage adds an image to the image cache
-func (ic *ICache) AddImage(imageConfig *metadata.ImageConfig) {
+// Add adds an image to the image cache
+func (ic *ICache) Add(imageConfig *metadata.ImageConfig, save bool) {
+	defer trace.End(trace.Begin(""))
```
emlin: looks like you have add with save or not, but do not have delete with save option. Is this option really necessary?

lib/apiservers/engine/backends/backends.go
jzt: That will happen in `vic/cmd/docker/main.go` here:

```go
	if err := vicbackends.Init(cli.portLayerAddr, productName, &vchConfig, vchConfig.InsecureRegistries); err != nil {
		log.Fatalf("failed to initialize backend: %s", err)
	}
```

lib/apiservers/engine/backends/backends.go
jzt: How? :)

lib/apiservers/engine/backends/cache/image_cache.go
jzt: Actually no, it's not necessary. I'll clean that up.

lib/apiservers/engine/backends/backends.go
emlin: if you don't care too much about the error type, you can just append all error messages. otherwise, there might have only one error message log out

lib/apiservers/engine/backends/backends.go
fdawg4l: Awesome!!  Thanks!

jzt:
@emlin To follow up, anyone who calls `Add()` or `RemoveImageByConfig()` must also explicitly call `Save()` if the changes are intended to be persisted to disk.

emlin:
LGTM 👍

lib/apiservers/engine/backends/backends.go
jzt: Fixed!

lib/imagec/layer_cache.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// LayerCache returns a reference to the layer cache
+func LayerCache() *LCache {
+	return layerCache
+}
+
+// NewLayerCache will create a new layer cache or rehydrate an
+// existing layer cache from the portlayer k/v store
+func NewLayerCache(client *client.PortLayer) error {
```
hickeng: Calling this NewLayerCache is odd given it's initializing a package wide global - InitializeLayerCache would be more indicative of what's actually occuring.

Most of this code is written as methods on a layerCache object - if it's viable you could have this return an *LCache that is stored into a global by the caller if that's the required usage model. That keeps the globals closer to the code that uses them

lib/imagec/layer_cache.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/vmware/vic/pkg/trace"
+)
+
+// LCache is an in-memory cache to account for existing image layers
+// It is used primarily by imagec when coordinating layer downloads
+// The cache is initially hydrated by way of the image cache at startup
+type LCache struct {
+	m sync.RWMutex
+	// Layers maps from layer id to layer metadata
+	Layers map[string]*ImageWithMeta
```
jzt: I thought about it, but it has docker's `types.V1Image` embedded inside which is why it is named like this to begin with (correct me if I'm wrong @caglar10ur).

Truth be told, I am not opposed to renaming image->layer everywhere it's appropriate. No small feat, though, since those terms are munged all over the place, originating mainly from the fact that the Portlayer refers to layers as images. :)

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -31,34 +29,41 @@ import (
...
 	"github.com/vmware/vic/lib/metadata"
-	"github.com/vmware/vic/pkg/vsphere/sys"
+	"github.com/vmware/vic/pkg/trace"
 )
 
 // ICache is an in-memory cache of image metadata. It is refreshed at startup
 // by a call to the portlayer. It is updated when new images are pulled or
 // images are deleted.
 type ICache struct {
 	m sync.RWMutex
```
jzt: @hickeng @vburenin sounds good. I wasn't aware of that pattern, though I prefer it to exporting the struct fields. Will push a fix.

lib/apiservers/engine/backends/backends.go
```diff
@@ -93,44 +94,86 @@ func Init(portLayerAddr, product string, config *config.VirtualContainerHostConf
...
+
 	log.Info("Creating image store")
 	if err := createImageStore(); err != nil {
 		log.Errorf("Failed to create image store")
 		return err
 	}
 
+	serviceOptions := registry.ServiceOptions{}
+	for _, registry := range insecureRegs {
+		insecureRegistries = append(insecureRegistries, registry.Path)
```
jzt: @hickeng would a simple prefix check do the trick?

lib/imagec/download.go
```diff
@@ -154,15 +151,16 @@ func (ldm *LayerDownloader) DownloadLayers(ctx context.Context, ic *ImageC) erro
...
 
 				continue
 			default:
 				return err
 			}
 		}
 
-		if downloading {
+		if layerConfig.Downloading {
```
jzt: Right now, code that want's to change `layerConfig.Downloading` calls `LayerCache().Commit()` which uses a lock, but there's nothing to stop code from changing this value manually. I think it's better not to export it, but instead expose a getter/setter in the form of `IsDownloading()` and `Commit()`. I will push a change.

lib/apiservers/engine/backends/cache/image_cache.go
jzt: @hickeng @vburenin I implemented this pattern for both the image cache and the layer cache. Both structs are now private and use anonymous structs for (de)serialization.

jzt:
@hickeng Right now, if the last client ctrl+C's out of a pull, the persona will continue to pull the image in the background. Could we use the context to do some cleanup in that case?

lib/imagec/layer_cache.go
```diff
@@ -0,0 +1,151 @@
...
+	log.Debugf("Updating layer cache")
+
+	val, err := kv.Get(client, layerCacheKey)
+	if err != nil && err != kv.ErrKeyNotFound {
+		return err
+	}
+
+	if val != "" {
+		if err = json.Unmarshal([]byte(val), layerCache); err != nil {
+			return fmt.Errorf("Failed to unmarshal layer cache: %s", err)
```
jzt: @hickeng There is currently no solution in place to recover from datastore corruption or version mismatch. It is, however, possible to do a consistency check for the layer and image caches at startup. I think that could be captured in a different issue.

There's also been talk between myself, @hmahmood, @vburenin, @matthewavery about using a distributed k/v store like etcd for this. Perhaps we should continue that discussion?

jzt:
♻️

PR #2842, opened 2016-10-21
VCSA cluster script updates

dougm:
- Require govc 0.10.0 or higher

- Configure ESX NTP

- Set ESX hostname

- Move firewall config to ESX script

- Add ssh authorized_key to ESX

- Include ESX hostname in all log messages

- Don't require vSAN in create-cluster


caglar10ur:
- Configure ESX NTP

THANK YOU! (I was fighting with it yesterday!)

LGTM

dougm:
.

PR #2852, opened 2016-10-21
Adding client ip to the list of host ips allowed for port mapping

### LINKED ISSUE #2648: Server error from portlayer: host IP is not supported for port bindings.
Currently using build 0.6.0. 

I have a use case for which I'd need to be able to configure a port mapping that includes the host IP address. 

Something like this: 

`docker -H 192.168.1.179:2375 run -p 192.168.1.179:6379:6379 -d redis:latest`

With the current VIC build this syntax isn't supported. You get:  

`docker: Error response from daemon: Server error from portlayer: host IP is not supported for port bindings.`

Right now the use case I am dealing with only requires being able to specify the host IP address in the port mapping albeit there <may be> niche use cases where being able to specify an IP that is different than the VCH proxy VM IP would be useful (?).

hmahmood:
Fixes #2648


lib/apiservers/engine/backends/container.go
```diff
@@ -1248,10 +1248,30 @@ func validateCreateConfig(config *types.ContainerCreateConfig) error {
...
+		if addrs, err := clientIPv4Addrs(); err != nil {
+			log.Warnf("could not get address for client interface: %s", err)
+		} else {
+			ips = make([]string, len(addrs))
+			for i := range addrs {
+				ips[i] = addrs[i].IP.String()
+			}
+		}
+
 		for _, pbs := range config.HostConfig.PortBindings {
```
caglar10ur: Do we need a map for the lookups here instead of this nested for loops?

hmahmood: Are you concerned about the new for loop I added? There should really be 1-2 addresses in that array.

The other nested loops are just an artifact of how things are presented to us by docker.

caglar10ur:
LGTM

lib/apiservers/engine/backends/container.go
```diff
@@ -1356,10 +1356,30 @@ func validateCreateConfig(config *types.ContainerCreateConfig) error {
 
 	// validate port bindings
 	if config.HostConfig != nil {
+		var ips []string
+		if addrs, err := clientIPv4Addrs(); err != nil {
```
hickeng: We should also allow the external address to be specified explicitly given that's the one we'd be using by default anyway.

lib/apiservers/engine/backends/container.go
```diff
@@ -1356,10 +1356,30 @@ func validateCreateConfig(config *types.ContainerCreateConfig) error {
...
+					// check if specified host ip equals any of the addresses on the "client" interface
+					found := false
+					for _, i := range ips {
+						if i == pb.HostIP {
+							found = true
+							break
+						}
+					}
+					if !found {
+						return InternalServerError("host IP is not supported for port bindings")
```
hickeng: Let's change this error message - it's incorrect once we' allow for *some* IP addresses to be specified in this manner.

andrewtchin:
lgtm

hmahmood:
♻️

hmahmood:
♻️

PR #2874, opened 2016-10-24
[full ci] Improve docker run for quick ending containers

### LINKED ISSUE #1578: Improve docker run for quick ending containers by delaying attach

caglar10ur:
by launching the process after a successful attach operation.
- Adds a cancelable context to tether to cancel long running goroutines,
- Adds RunBlock field to Sessions struct,
- Bind@InteractionHandler sets that value and tether uses it. If attach
  happens before the start this property will be persist in the vms,
  otherwise it will be thrown away by the Commit code.

also;
- Adds some more logging,
- Shuffles some code around,
- Fixes the Errwriter remove issue,
- Removes not used/needed Signal related code,

Fixes #1578

### DIFF
--- a/cmd/tether/attach.go
+++ b/cmd/tether/attach.go
@@ -369,7 +369,7 @@ func (t *attachServerSSH) run() error {
 			// The only channel type we'll support is attach
 			if attachchan.ChannelType() != attachChannelType {
 				detail := fmt.Sprintf("unknown channel type %s", attachchan.ChannelType())
-				attachchan.Reject(ssh.Prohibited, detail)
+				attachchan.Reject(ssh.UnknownChannelType, detail)
 				log.Error(detail)
 				continue
 			}
--- a/cmd/tether/attach.go
+++ b/cmd/tether/attach.go
@@ -21,7 +21,6 @@ import (
 	"net"
 	"sync"
 	"sync/atomic"
-	"syscall"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -70,6 +69,9 @@ type attachServerSSH struct {
 	ctx    context.Context
 	cancel context.CancelFunc
 
+	// Used for ordering events between global mux and channel mux
+	askedAndAnswered chan struct{}
+
 	// INTERNAL: must set by testAttachServer only
 	testing bool
 }
@@ -108,18 +110,18 @@ func (t *attachServerSSH) Reload(config *tether.ExecutorConfig) error {
 	return nil
 }
 
-// Enabled sets the enabled to true
-func (t *attachServerSSH) Enabled() {
+// Enable sets the enabled to true
+func (t *attachServerSSH) Enable() {
 	atomic.StoreInt32(&t.enabled, 1)
 }
 
-// Disabled sets the enabled to false
-func (t *attachServerSSH) Disabled() {
+// Disable sets the enabled to false
+func (t *attachServerSSH) Disable() {
 	atomic.StoreInt32(&t.enabled, 0)
 }
 
-// IsEnabled returns whether the enabled is true
-func (t *attachServerSSH) IsEnabled() bool {
+// Enabled returns whether the enabled is true
+func (t *attachServerSSH) Enabled() bool {
 	return atomic.LoadInt32(&t.enabled) == 1
 }
 
@@ -145,7 +147,7 @@ func (t *attachServerSSH) start() error {
 		return err
 	}
 
-	if t.IsEnabled() {
+	if t.Enabled() {
 		err := fmt.Errorf("attach server is already enabled")
 		log.Error(err)
 		return err
@@ -179,7 +181,8 @@ func (t *attachServerSSH) start() error {
 	}
 	t.sshConfig.AddHostKey(pkey)
 
-	t.Enabled()
+	// enable the server and start it
+	t.Enable()
 	go t.run()
 
 	return nil
@@ -195,20 +198,20 @@ func (t *attachServerSSH) stop() error {
 		return err
 	}
 
-	if !t.IsEnabled() {
+	if !t.Enabled() {
 		err := fmt.Errorf("attach server is not enabled")
 		log.Error(err)
 		return err
 	}
 
-	log.Debugf("Setting enabled to false")
-	t.Disabled()
+	// disable the server
+	t.Disable()
 
 	// This context is used by backchannel only. We need to cancel it before
 	// trying to obtain the following lock so that backchannel interrupts the
 	// underlying Read call by calling Close on it.
 	// The lock is  held by backchannel's caller and not released until it returns
-	log.Debugf("Canceling the context")
+	log.Debugf("Canceling AttachServer's context")
 	t.cancel()
 
 	log.Debugf("Acquiring the connection lock")
@@ -292,14 +295,14 @@ func (t *attachServerSSH) run() error {
 	var reqs <-chan *ssh.Request
 	var err error
 
-	for t.IsEnabled() {
+	for t.Enabled() {
 		serverConn.Lock()
 		established = serverConn.ServerConn != nil
 		serverConn.Unlock()
 
 		// keep waiting for the connection to establish
-		for !established && t.IsEnabled() {
-			log.Debugf("Trying to establish a connection")
+		for !established && t.Enabled() {
+			log.Infof("Trying to establish a connection")
 
 			establishFn := func() error {
 				// we hold the t.conn.Lock during the scope of this function
@@ -360,14 +363,14 @@ func (t *attachServerSSH) run() error {
 		// Global requests
 		go t.globalMux(reqs)
 
-		log.Println("ready to service attach requests")
+		log.Infof("Ready to service attach requests")
 		// Service the incoming channels
 		for attachchan := range chans {
 			// The only channel type we'll support is attach
 			if attachchan.ChannelType() != attachChannelType {
 				detail := fmt.Sprintf("unknown channel type %s", attachchan.ChannelType())
+				attachchan.Reject(ssh.Prohibited, detail)
 				log.Error(detail)
-				attachchan.Reject(ssh.UnknownChannelType, "unknown channel type")
 				continue
 			}
 
@@ -375,33 +378,21 @@ func (t *attachServerSSH) run() error {
 			bytes := attachchan.ExtraData()
 			if bytes == nil {
 				detail := "attach channel requires ID in ExtraData"
-				log.Error(detail)
 				attachchan.Reject(ssh.Prohibited, detail)
+				log.Error(detail)
 				continue
 			}
 
 			sessionid := string(bytes)
 			session, ok := t.config.Sessions[sessionid]
-
-			reason := ""
 			if !ok {
-				reason = "is unknown"
-			} else if session.Cmd.Process == nil {
-				reason = "process has not been launched"
-			} else if session.Cmd.Process.Signal(syscall.Signal(0)) != nil {
-				reason = "process has exited"
-			}
-
-			if reason != "" {
-				detail := fmt.Sprintf("attach request: session %s %s", sessionid, reason)
-				log.Error(detail)
+				detail := fmt.Sprintf("session %s is invalid", sessionid)
 				attachchan.Reject(ssh.Prohibited, detail)
+				log.Error(detail)
 				continue
 			}
 
-			log.Infof("accepting incoming channel for %s", sessionid)
 			channel, requests, err := attachchan.Accept()
-			log.Debugf("accepted incoming channel for %s", sessionid)
 			if err != nil {
 				detail := fmt.Sprintf("could not accept channel: %s", err)
 				log.Errorf(detail)
@@ -410,14 +401,20 @@ func (t *attachServerSSH) run() error {
 
 			// bind the channel to the Session
 			log.Debugf("binding reader/writers for channel for %s", sessionid)
+
+			log.Debugf("Adding [%p] to Outwriter", channel)
 			session.Outwriter.Add(channel)
+			log.Debugf("Adding [%p] to Reader", channel)
 			session.Reader.Add(channel)
 
 			// cleanup on detach from the session
-			detach := func() {
-				log.Debugf("cleanup on detach from the session")
+			cleanup := func() {
+				log.Debugf("Cleanup on detach from the session")
 
+				log.Debugf("Removing [%p] from Outwriter", channel)
 				session.Outwriter.Remove(channel)
+
+				log.Debugf("Removing [%p] from Reader", channel)
 				session.Reader.Remove(channel)
 
 				channel.Close()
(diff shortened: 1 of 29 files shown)


anchal-agrawal:
An integration test in [1-06-Docker-Run](https://github.com/vmware/vic/blob/master/tests/test-cases/Group1-Docker-Commands/1-06-Docker-Run.robot#L53) depends on this PR.

lib/tether/tether.go
```diff
@@ -173,128 +184,235 @@ func (t *tether) cleanup() {
...
+	switch t.config.DebugLevel {
+	case 0:
+		log.SetLevel(log.InfoLevel)
+		// TODO: do not echo application output to console without debug enabled
+		serial.DisableTracing()
+	case 1:
+		log.SetLevel(log.DebugLevel)
+		serial.EnableTracing()
+	case 2:
+		log.Info("Launching pprof server on port 6060")
```
vburenin: does it mean we don't want to enable/disable tracing on this level?

lib/tether/tether.go
```diff
@@ -173,128 +184,235 @@ func (t *tether) cleanup() {
...
-		extraconfig.Encode(t.sink, t.config)
 
-		//process the filesystem mounts - this is performed after networks to allow for network mounts
-		for k, v := range t.config.Mounts {
-			if v.Source.Scheme != "label" {
-				detail := fmt.Sprintf("unsupported volume mount type for %s: %s", k, v.Source.Scheme)
+			logwriter, err := t.ops.SessionLog(session)
+			if err != nil {
+				detail := fmt.Sprintf("failed to get log writer for session: %s", err)
 				log.Error(detail)
```
vburenin: This error is logger here, isn't it logged somewhere before?

lib/tether/tether.go
```diff
@@ -173,128 +184,235 @@ func (t *tether) cleanup() {
...
+}
 
-		// process the sessions and launch if needed
-		for id, session := range t.config.Sessions {
-			log.Debugf("Processing config for session %s", session.ID)
-			var proc = session.Cmd.Process
+func (t *tether) reloadExtensions() error {
+	// reload the extensions
+	for name, ext := range t.extensions {
+		log.Info("Passing config to " + name)
```
vburenin: Debug?
log.Debugf("Passing config to: %s ", name)

lib/tether/tether.go
```diff
@@ -173,128 +184,235 @@ func (t *tether) cleanup() {
...
+				// FIXME: we cannot have this embedded knowledge of the extraconfig encoding pattern, but not
+				// currently sure how to expose it neatly via a utility function
+				extraconfig.EncodeWithPrefix(t.sink, session, fmt.Sprintf("guestinfo.vice..sessions|%s", id))
+				log.Warnf("Re-launching process for session %s (count: %d)", id, session.Diagnostics.ResurrectionCount)
+				session.Cmd = *restartableCmd(&session.Cmd)
 			}
+			session.Unlock()
+
+			wg.Add(1)
+			go func(session *SessionConfig) {
```
vburenin: How many session could be seen here? if it is hundreds, wouldn't it make sense to impose a limit on a number of concurrent launches?

lib/tether/tether.go
```diff
@@ -173,128 +184,235 @@ func (t *tether) cleanup() {
...
+
+	wg.Wait()
+	// close the channel
+	close(resultsCh)
+
+	// iterate over the results
+	for result := range resultsCh {
+		if result.err != nil {
+			detail := fmt.Errorf("failed to launch %s for %s: %s", result.path, result.id, result.err)
+			log.Error(detail)
```
vburenin: same thing, does it make sense to log error here and then return it?

lib/tether/tether_linux.go
```diff
@@ -75,18 +76,17 @@ func (t *tether) childReaper() error {
 				// general resiliency
 				defer func() {
 					if r := recover(); r != nil {
-						fmt.Printf("Recovered in childReaper %v", r)
+						fmt.Printf("Recovered in childReaper %s", debug.Stack())
```
vburenin: is it something we really want to send into stdout?

lib/tether/tether.go
caglar10ur: hmmm good catch, fixing it

lib/tether/tether.go
caglar10ur: Done

lib/tether/tether.go
caglar10ur: I don't think we will have hundreds of them (maybe 2-3 at most), we can limit it if it turns out otherwise

lib/tether/tether.go
caglar10ur: We are doing that (logging first and then returning), or did I misunderstood what you are asking?

lib/tether/tether_linux.go
caglar10ur: This should be really rare, in fact it shouldn't be happening at all so when it happens I rather want to see everything

lib/tether/tether_linux.go
```diff
@@ -140,9 +140,8 @@ func (t *tether) stopReaper() {
 	signal.Reset(syscall.SIGCHLD)
 
 	// just closing the incoming channel is not going to stop the iteration
-	// so we use done channel to signal it
-	log.Debugf("Signalling the child reaper loop")
-	close(t.done)
+	// so we use the context cancellation to signal it
+	t.cancel()
```
hmahmood: `childReaper` should have its own channel (called `done` or something similar) that should stop the reaper loop only. You could shut down these side loops one by one with their own done channels (did I just make up a "task" abstraction?) when tether shuts down.

cmd/tether/attach.go
```diff
@@ -437,8 +446,15 @@ func (t *attachServerSSH) run() error {
 			log.Debugf("reader/writers bound for channel for %s", sessionid)
 
 			go t.channelMux(requests, session, detach)
-		}
 
+			// FIXME: This needs to be runblock instead of Attach
+			if session.Attach && session.Started != "true" {
+				// make sure that portlayer received the container id back
+				log.Debugf("Unblocking the launch of %s", sessionid)
+				session.ClearToLaunch <- <-t.askedAndAnswered
```
hmahmood: Does this work? Would be able to just do this with one channel that you close below and other places we just wait for that?

cmd/tether/attach.go
```diff
@@ -464,6 +483,13 @@ func (t *attachServerSSH) globalMux(reqchan <-chan *ssh.Request) {
 			}
 			msg := msgs.ContainersMsg{IDs: keys}
 			payload = msg.Marshal()
+
+			// unblock ^ (above)
+			pendingFn = func() {
+				once.Do(func() {
+					close(t.askedAndAnswered)
```
hmahmood: Why do we wait for this request to come in before launching sessions?

cmd/tether/attach.go
caglar10ur: I believe it does. t.askedAndAnswered will block until we close it and then we will send the zero value of the channel down to CleartToLaunc (reading from a closed channel sends its zero value)

cmd/tether/attach.go
caglar10ur: We need container-ids in the portlayer before unblocking the process which is what this request responds (per attach operation)

lib/tether/tether_linux.go
caglar10ur: Not sure I understood. Can you elaborate on why do you suggest having different signaling mechanisms?

hmahmood: `t.cancel()` seems to indicate that the tether is done. I just thought it would be more flexible if each of these operations had their own channel to stop them executing. Its conceptually also clearer; may make this code easier to understand.

lib/tether/tether.go
vburenin: I wonder if the caller of this function logs this error.

cmd/tether/attach.go
```diff
@@ -486,34 +500,42 @@ func (t *attachServerSSH) globalMux(reqchan <-chan *ssh.Request) {
...
 	}
 }
 
-func (t *attachServerSSH) channelMux(in <-chan *ssh.Request, session *tether.SessionConfig, detach func()) {
+func (t *attachServerSSH) channelMux(in <-chan *ssh.Request, session *tether.SessionConfig, cleanup func()) {
 	defer trace.End(trace.Begin("attach server channel request handler"))
 
-	// detach
-	defer detach()
+	var ok bool
```
vburenin: Why do you need declaration of these variables here? You could declare them right in place where you need them.

lib/tether/tether.go
```diff
@@ -173,128 +184,237 @@ func (t *tether) cleanup() {
...
+	// TODO: move all of this into an extension.Pre() block when we move to that model
+	// adjust the logging level appropriately
+	switch t.config.DebugLevel {
+	case 0:
+		log.SetLevel(log.InfoLevel)
+		// TODO: do not echo application output to console without debug enabled
+		serial.DisableTracing()
+	case 1:
+		log.SetLevel(log.DebugLevel)
+		serial.EnableTracing()
```
vburenin: Maybe no tracing on 1?
Also would be good to call:
trace.DisableTracing() or trace.EnableTracing() accordingly.

lib/tether/tether.go
caglar10ur: Done

lib/tether/tether.go
caglar10ur: Done

lib/tether/tether_linux.go
caglar10ur: Agreed but right now we are stopping them only when tether is done and ctx seems to be doing that well. We can add more granular control to tether but to be honest I prefer not doing it here but a separate PR (maybe with a big tether refactoring)

hmahmood:
lgtm
@hickeng should take a look as well

lib/tether/tether_linux.go
hickeng: Write this to os.Stderr - that's redirected to the debug log, but stdout is not.

lib/tether/tether_linux.go
```diff
@@ -228,14 +227,25 @@ func establishPty(session *SessionConfig) error {
 		// it frees up all resources - does that mean it frees the output buffers?
 		go func() {
 			_, gerr := io.Copy(session.Outwriter, session.Pty)
-			log.Debugf("PTY stdout copy: %s", gerr)
+			if e, ok := gerr.(*os.PathError); ok && e.Err == syscall.EIO {
+				// We can safely ignore this error, because
+				// it's just the PTY telling us that it closed all good.
+				log.Debugf("Ignoring EIO error returned by Outwriter io.Copy")
```
hickeng: If we're explicitly logging a different message because this is expected we may as well it actually distinct:
"PTY copy finished"

and make the other case a Warn at least given it's not the expected path.

lib/tether/toolbox.go
```diff
@@ -127,8 +127,8 @@ func (t *Toolbox) kill(name string) error {
 		return fmt.Errorf("failed to kill container: process not found")
 	}
 
-	session.m.Lock()
-	defer session.m.Unlock()
+	session.Lock()
```
hickeng: every use of session.Cmd.Process in toolbox now needs a nil check - the extension is now being initialized before process launch.

tests/test-cases/Group1-Docker-Commands/1-26-Docker-Hello-World.robot
```diff
@@ -1,14 +1,11 @@
...
-    ${status}=  Get State Of Github Issue  1578
-    Run Keyword If  '${status}' == 'closed'  Fail  Test 1-26-Docker-Hello-World.robot needs to be updated now that Issue #1578 has been resolved
-    Log  Issue \#1578 is blocking implementation  WARN
-    #${rc}  ${output}=  Run And Return Rc And Output  docker ${params} run hello-world
-    #Log  ${output}
-    #Should Be Equal As Integers  ${rc}  0
\ No newline at end of file
+    ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} run hello-world
+    Log  ${output}
+    Should Be Equal As Integers  ${rc}  0
```
hickeng: Should we be confirming what was in output as well?

lib/tether/tether_linux.go
caglar10ur: removed the redundant logging

lib/tether/toolbox.go
caglar10ur: done

tests/test-cases/Group1-Docker-Commands/1-26-Docker-Hello-World.robot
caglar10ur: we can't use -it here and -i is not guaranteed to give an output :/ I can check logs output

cmd/tether/attach.go
hickeng: @caglar10ur this will work given our current logic, but is there an inherent reason why? If a channel request is made with a legitimate ID, even without asking, is that not the trigger for launch?

lib/tether/interfaces.go
```diff
@@ -36,7 +36,7 @@ type Operations interface {
 	MountLabel(ctx context.Context, label, target string) error
 	Fork() error
 
-	SessionLog(session *SessionConfig) (dio.DynamicMultiWriter, error)
+	SessionLog(session *SessionConfig) (dio.DynamicMultiWriter, dio.DynamicMultiWriter, error)
```
hickeng: Comment about what the two writers are for

lib/tether/tether.go
```diff
@@ -173,128 +184,237 @@ func (t *tether) cleanup() {
...
 }
 
+func (t *tether) setLogLevel() {
+	// TODO: move all of this into an extension.Pre() block when we move to that model
+	// adjust the logging level appropriately
+	switch t.config.DebugLevel {
+	case 0:
+		log.SetLevel(log.InfoLevel)
+		// TODO: do not echo application output to console without debug enabled
+		serial.DisableTracing()
```
vburenin: please also call trace.DisableTracing()

lib/tether/tether.go
```diff
@@ -173,128 +184,237 @@ func (t *tether) cleanup() {
...
+	case 1:
+		log.SetLevel(log.DebugLevel)
+		serial.DisableTracing()
```
vburenin: please also call trace.DisableTracing()

lib/tether/tether.go
```diff
@@ -173,128 +184,237 @@ func (t *tether) cleanup() {
...
+	case 2:
+		log.SetLevel(log.DebugLevel)
+		serial.EnableTracing()
```
vburenin: please also call trace.EnableTracing()

lib/tether/tether.go
```diff
@@ -173,128 +184,237 @@ func (t *tether) cleanup() {
...
+
+		log.Info("Launching pprof server on port 6060")
+		fn := func() {
+			go http.ListenAndServe("0.0.0.0:6060", nil)
+		}
+
+		once.Do(fn)
+	default:
+		log.SetLevel(log.DebugLevel)
```
vburenin: would you want to enable/disable tracing here?

hickeng:
LGTM - there's stuff to be added and tidied, but this is functional for 0.7

caglar10ur:
♻️

caglar10ur:
♻️

PR #2883, opened 2016-10-24
[full ci] VIC Admin Username/Password auth

gigawhitlocks:
What works: user can log in w/ vsphere creds; invalid logins get booted out, and everything considered protected gets wrapped through `Authenticated` to enable authentication. Sessions last 24 hours before the token is considered expired. Token storing timestamp used for authentication is encrypted & stored as a cookie in the user's browser.

What still needs work: replacing sessions throughout with the one generated at login time. This is the next thing now that I have my sessions expiring properly and the rest seems to be working.

I also need to add a logout route to delete the user's session cookie. There also may be bugs, and I know there are some half-finished methods and changes in here that might not make a lot of sense; this is a WIP.


cmd/vicadmin/server.go
```diff
@@ -46,10 +55,78 @@ const (
...
+type UserSessionStore struct {
+	*sync.Mutex
+	sessions map[string]*UserSession
+	ticker   *time.Ticker
+}
+
+func (u *UserSessionStore) Add(username string, config *session.Config) *UserSession {
+	u.Lock()
+	defer u.Unlock()
+	u.sessions[username] = &UserSession{
```
vburenin: Have you considered to store a UserSession into variable and then return it? Double map look ups are not good.

cmd/vicadmin/server.go
```diff
@@ -46,10 +55,78 @@ const (
...
+}
+
+func (u *UserSessionStore) Delete(username string) {
+	u.Lock()
+	defer u.Unlock()
+	delete(u.sessions, username)
+}
+
+func (u *UserSessionStore) Get(username string) *UserSession {
+	return u.sessions[username]
```
vburenin: not protected.

vburenin: What is the difference between Get and GetSession?
I would definitely name this method as:
 UserSession(username string)
next one: VSphereSession(username string)

cmd/vicadmin/server.go
```diff
@@ -46,10 +55,78 @@ const (
...
+	}
+}
+
+// NewUserSessionStore creates & initializes a UserSessionStore and starts a session reaper in the background
+func NewUserSessionStore() *UserSessionStore {
+	u := &UserSessionStore{
+		sessions: make(map[string]*UserSession),
+		ticker:   time.NewTicker(time.Minute),
+	}
+	go u.reaper()
```
vburenin: Do you care if this never executed? Why reaper is given only a minute to wait? Why not 5 or 60?

cmd/vicadmin/server.go
```diff
@@ -114,39 +191,151 @@ func (s *server) listenPort() int {
...
+	defer trace.End(trace.Begin(""))
+
+	authHandler := func(w http.ResponseWriter, r *http.Request) {
+		websession, _ := store.Get(r, sessionCookieKey)
+
+		if len(r.TLS.PeerCertificates) > 0 { // the user is authenticated by certificate at connection time
+			usersess := s.uss.Add("root", &rootConfig.Config)
+			websession, _ := store.Get(r, sessionCookieKey)
+			timeNow, _ := usersess.created.MarshalText()
+			websession.Values["created"] = string(timeNow)
```
vburenin: You are using these keys in many places, please use constants for them.

cmd/vicadmin/server.go
```diff
@@ -46,10 +55,78 @@ const (
...
+// UserSession holds a user's session metadata
+type UserSession struct {
+	username string
+	created  time.Time
+	config   *session.Config
+}
+
+// UserSessionStore holds and manages user sessions
+type UserSessionStore struct {
+	*sync.Mutex
```
hmahmood: Why a pointer? Also if this object cannot be locked by callers, then you should make the lock an unexported field.

hmahmood: Also an `RWMutex` would be better suited here.

cmd/vicadmin/server.go
```diff
@@ -46,10 +55,78 @@ const (
...
+
+// GetSession logs into vSphere and returns a session object. Caller responsible for error handling/logout
+func (u *UserSessionStore) GetSession(username string) (vSphereSession *session.Session, err error) {
+	return vSphereSessionGet(u.Get(username).config)
+}
+
+// reaper takes abandoned sessions to a farm upstate so they don't build up forever
+func (u *UserSessionStore) reaper() {
+	select {
+	case <-u.ticker.C:
```
hmahmood: Don't understand why we need this. Just the for loop should suffice? Or should this be in a loop? See also https://golang.org/pkg/time/#Tick

cmd/vicadmin/server.go
```diff
@@ -114,39 +191,151 @@ func (s *server) listenPort() int {
...
+// Enforces authentication on route `link` and runs `handler` on successful auth
+func (s *server) Authenticated(link string, handler func(http.ResponseWriter, *http.Request)) {
+	defer trace.End(trace.Begin(""))
+
+	authHandler := func(w http.ResponseWriter, r *http.Request) {
+		websession, _ := store.Get(r, sessionCookieKey)
+
+		if len(r.TLS.PeerCertificates) > 0 { // the user is authenticated by certificate at connection time
+			usersess := s.uss.Add("root", &rootConfig.Config)
+			websession, _ := store.Get(r, sessionCookieKey)
```
hmahmood: Line 208 above?

hmahmood: Also should you check if websession is nil?

hmahmood: Is it safe to manipulate websession here without some sort of synchronization?

cmd/vicadmin/server.go
```diff
@@ -114,39 +191,151 @@ func (s *server) listenPort() int {
...
+			log.Errorf("vSphere service URL was not a valid format; parsing returned error: %s", err)
+		}
+		serviceURL.User = user
+		userconfig.Service = serviceURL.String()
+
+		// check login
+		usersession, err := vSphereSessionGet(&userconfig)
+		if err != nil {
+			// something went wrong or we could not authenticate
+			http.Redirect(res, req, "/authentication?unauthorized", 302)
```
hmahmood: Should you not return from the function after this?

cmd/vicadmin/server.go
```diff
@@ -114,39 +191,151 @@ func (s *server) listenPort() int {
...
+		http.Redirect(res, req, "/", 302)
+	}
+
+	// Render login page (shows up on non-POST requests):
+	sess, err := client(&rootConfig)
+	v := vicadmin.NewValidator(ctx, &vchConfig, sess)
+	tmpl, err := template.ParseFiles("auth.html")
+	err = tmpl.ExecuteTemplate(res, "auth.html", v)
+	if err != nil {
+		log.Errorf("Error parsing template: %s", err)
```
hmahmood: No error returned to client? Maybe a 500?

cmd/vicadmin/server.go
```diff
@@ -168,36 +357,35 @@ func (s *server) stop() error {
...
+func getSessionCookie(req *http.Request) *sessions.Session {
+	cookieSessionData, _ := store.Get(req, sessionCookieKey)
+	return cookieSessionData
+}
+
 func (s *server) bundleContainerLogs(res http.ResponseWriter, req *http.Request, f format) {
 	defer trace.End(trace.Begin(""))
 
 	readers := defaultReaders
+	c, err := s.getSessionFromRequest(req)
```
hmahmood: This always returns non-nil `c`?

gigawhitlocks:
@vburenin and @hmahmood this is a WIP and most everything you've pointed out is either broken or an issue in progress that I'm aware of.

gigawhitlocks:
This is a WIP and it's only up because I was asked to produce it; this code is not ready for thorough review over things like error checking, names of variables, etc, as I'm likely to change everything and anything without warning. Please comment on the approach, the structure, or any **blaring security flaws**, and save the "no error checking here" type comments for after I think I'm done.

cmd/vicadmin/server.go
gigawhitlocks: `store` is only modified by this thread. `websession` is a value in `store`

gigawhitlocks: Well I should definitely check that error, and will in the final version of this code. There are numerous places where I've skipped error checking while I work on the feature

cmd/vicadmin/server.go
hmahmood: UserSessionStore

cmd/vicadmin/server.go
hmahmood: The way you have it right now, its only going to run once; the ticker will fire, you will run the for loop and then exit the function. It appears you want this to be running as long as the store is still valid, so you would at the very least enclose this in a for loop.

See the example in the link I also included above, might simplify this code.

cmd/vicadmin/server.go
hmahmood: If another thread calls `store.Get` with the same session cookie key, will it get the same object?

gigawhitlocks:
OK this is starting to look more complete; constant expiration times were chosen out of thin air and I'd love comments regarding those now that things are stabilizing. @hmahmood in reply to one of your now-outdated comments above I'm also going to look at replacing that `select` statement with a `for _ := range(ticker)` sort of deal tomorrow.

I'll also have to update the non-client-certificate VicAdmin tests because this will cause them all to fail.

Also missing is a logout route, will add that next, and will be doing general cleanup. **small comments on nits are now welcome** as the overall structure is pretty much in-place here

cmd/vicadmin/server.go
gigawhitlocks: clearly; I wouldn't leave them like this at merge time.

cmd/vicadmin/server.go
gigawhitlocks: I will be adding error handling to check and make sure `err` is non-nil here. This is still prototype version.

cmd/vicadmin/server.go
```diff
@@ -114,39 +130,158 @@ func (s *server) listenPort() int {
...
+
+	authHandler := func(w http.ResponseWriter, r *http.Request) {
+		websession, err := store.Get(r, sessionCookieKey)
+		if err != nil {
+			log.Errorf("Got \"%s\" while looking up cookie store", err.Error())
+		}
+
+		if len(r.TLS.PeerCertificates) > 0 { // the user is authenticated by certificate at connection time
+			usersess := s.uss.Add("root", &rootConfig.Config)
+			websession, _ := store.Get(r, sessionCookieKey)
```
hmahmood: You ignore the error here but not on line 147?

cmd/vicadmin/server.go
```diff
@@ -114,39 +130,158 @@ func (s *server) listenPort() int {
...
+			ClusterPath:    rootConfig.ClusterPath,
+			DatacenterPath: rootConfig.DatacenterPath,
+			DatastorePath:  rootConfig.DatastorePath,
+			HostPath:       rootConfig.Config.HostPath,
+			PoolPath:       rootConfig.PoolPath,
+		}
+		user := url.UserPassword(req.FormValue("username"), req.FormValue("password"))
+		serviceURL, err := soap.ParseURL(rootConfig.Service)
+		if err != nil {
+			log.Errorf("vSphere service URL was not a valid format; parsing returned error: %s", err)
```
hmahmood: serviceURL is still in good shape after this error?

cmd/vicadmin/server.go
```diff
@@ -114,39 +130,158 @@ func (s *server) listenPort() int {
...
+		}
+		serviceURL.User = user
+		userconfig.Service = serviceURL.String()
+
+		// check login
+		usersession, err := vSphereSessionGet(&userconfig)
+		if err != nil || usersession == nil {
+			// something went wrong or we could not authenticate
+			log.Warnf("User %s failed to authenticated at %s", user, time.Now())
+			http.Redirect(res, req, "/authentication?unauthorized", 302)
```
hmahmood: You need to return from the function after this. See https://golang.org/doc/articles/wiki/#tmp_7

cmd/vicadmin/server.go
gigawhitlocks: nah if this error occurs here, something is very wrong. I'll make sure there's a redirect and/or return statement following.

cmd/vicadmin/server.go
gigawhitlocks: added a comment to explain why

cmd/vicadmin/server.go
gigawhitlocks: Yeah you're right, I'll make sure I have `return` statements after all of my `Redirect`s

cmd/vicadmin/server.go
gigawhitlocks: This returns a `500` error as well now.

cmd/vicadmin/server.go
```diff
@@ -114,39 +132,184 @@ func (s *server) listenPort() int {
...
+func (s *server) Handle(link string, h http.Handler) {
+	s.mux.Handle(link, gorillacontext.ClearHandler(h))
+}
+
+// Enforces authentication on route `link` and runs `handler` on successful auth
+func (s *server) Authenticated(link string, handler func(http.ResponseWriter, *http.Request)) {
+	defer trace.End(trace.Begin(""))
+
+	authHandler := func(w http.ResponseWriter, r *http.Request) {
+		websession, _ := store.Get(r, sessionCookieKey) // ignore error because it is okay if it doesn't exist
```
hmahmood: websession == nil if there is an error.

cmd/vicadmin/server.go
```diff
@@ -114,39 +132,184 @@ func (s *server) listenPort() int {
...
+
+		if len(r.TLS.PeerCertificates) > 0 { // the user is authenticated by certificate at connection time
+			usersess := s.uss.Add("root", &rootConfig.Config)
+
+			timeNow, err := usersess.created.MarshalText()
+			if err != nil {
+				// the only way (that I can think of!) this could happen is if the cookie is tampered with
+				// so we'll just treat this error as an auth failure
+				log.Errorf("Failed to unmarshal time object %+v into text due to error: %s", usersess.created, err)
+				http.Redirect(w, r, authFailure, 302)
```
hmahmood: Invalidate the user session in this case?

cmd/vicadmin/server.go
hmahmood: Get will actually create a new session if it does not exist. In some cases below, we abandon websession; do we leak websession then?

cmd/vicadmin/usersession.go
```diff
@@ -0,0 +1,94 @@
...
+	Add(username string, config *session.Config) *UserSession
+	Delete(username string)
+	GetRealSession(username string) (vSphereSession *session.Session, err error)
+	GetUserSession(username string) *UserSession
+}
+
+func (u *UserSessionStore) Add(username string, config *session.Config) *UserSession {
+	u.mutex.Lock()
+	defer u.mutex.Unlock()
+	u.sessions[username] = &UserSession{
```
hmahmood: @vburenin pointed this out before; you can store the new `UserSession` in a local variable to avoid the second lookup on line 52.

cmd/vicadmin/server.go
gigawhitlocks: We do want to create a new session if it does not exist. If `Save` isn't called, it's just garbage collected. If you call `Save` it goes into `store` and will be removed by Gorilla after `MaxAge` set in `Options` when creating the store.

gigawhitlocks: I should probably change the name of `store` to something more descriptive..

cmd/vicadmin/server.go
```diff
@@ -114,39 +132,184 @@ func (s *server) listenPort() int {
...
+
+		}
+		websession.Values["created"] = string(timeNow)
+		websession.Values["username"] = req.FormValue("username")
+		if err := websession.Save(req, res); err != nil {
+			log.Errorf("\"%s\" occurred while trying to save session to browser", err.Error())
+		}
+
+		// redirect to dashboard
+		http.Redirect(res, req, "/", 302)
```
gigawhitlocks: yeah definitely

cmd/vicadmin/server.go
```diff
@@ -114,39 +137,203 @@ func (s *server) listenPort() int {
...
+		// if the date on the cookie was valid, then the user is authenticated
+		handler(w, r)
+	}
+	s.mux.Handle(link, gorillacontext.ClearHandler(http.HandlerFunc(authHandler)))
+}
+
+// renders the page for login and handles authorization requests
+func (s *server) loginPage(res http.ResponseWriter, req *http.Request) {
+	defer trace.End(trace.Begin(""))
+	ctx := context.Background()
```
vburenin: req.Context()? https://golang.org/pkg/net/http/#Request.Context

cmd/vicadmin/usersession.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func (u *UserSessionStore) Delete(username string) {
+	u.mutex.Lock()
+	defer u.mutex.Unlock()
+	delete(u.sessions, username)
+}
+
+// Grabs the UserSession metadta object and doesn't establish a connection to vSphere
+func (u *UserSessionStore) Get(username string) *UserSession {
```
vburenin: Please name it as UserSession

cmd/vicadmin/usersession.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+}
+
+// NewUserSessionStore creates & initializes a UserSessionStore and starts a session reaper in the background
+func NewUserSessionStore() *UserSessionStore {
+	u := &UserSessionStore{
+		sessions: make(map[string]*UserSession),
+		ticker:   time.NewTicker(time.Minute),
+		mutex:    sync.RWMutex{},
+		cookies:  sessions.NewCookieStore([]byte(securecookie.GenerateRandomKey(64)), []byte(securecookie.GenerateRandomKey(32))),
```
vburenin: Please, split it into multiple lines.

cmd/vicadmin/usersession.go
gigawhitlocks: oh hey that makes a lot of sense :+1:

cmd/vicadmin/usersession.go
gigawhitlocks: you know, I would've because I agree that it's too long but I always catch crap when I wrap my lines :|

gigawhitlocks: Can't win I guess.

cmd/vicadmin/server.go
gigawhitlocks: no, this is a different type of context object
https://golang.org/x/net/context

hmahmood: It appears to be `context.Context` in both cases. You should probably use the request's context object.

cmd/vicadmin/server.go
hmahmood: redirects are normally 3xx; why don't we just return a response here, and not redirect?

cmd/vicadmin/server.go
```diff
@@ -112,39 +134,203 @@ func (s *server) listenPort() int {
...
+		// successful login above; user is authenticated
+		// log out, disregard errors
+		usersession.Client.Logout(context.Background())
+
+		// save user config locally
+		usersess := s.uss.Add(req.FormValue("username"), &userconfig)
+
+		// create a token to save as an encrypted & signed cookie
+		// ignore error because there will be one if the session store does not exist, however, it will be created
+		websession, _ := s.uss.cookies.Get(req, sessionCookieKey)
```
hmahmood: There is a case were Get() will return (nil, err). You should check for error.

cmd/vicadmin/server.go
```diff
@@ -166,36 +352,51 @@ func (s *server) stop() error {
 	return nil
 }
 
+// logout handler expires the user's session cookie by setting its creation time to the beginning of time
+func (s *server) logoutHandler(res http.ResponseWriter, req *http.Request) {
+	websession, err := s.uss.cookies.Get(req, sessionCookieKey)
```
hmahmood: Won't Get() create a cookie? The only case where it returns an error is if the cookie name is invalid.

cmd/vicadmin/server.go
```diff
@@ -166,36 +352,51 @@ func (s *server) stop() error {
...
+	if err != nil {
+		// if there's no session store, probably calling logout doesn't make much sense
+		http.Redirect(res, req, authFailure, 302)
+		return
+	}
+
+	// ignore parsing/marshalling errors because we're parsing a hardcoded beginning-of-time string
+	beginningOfTime, _ := time.Parse(time.RFC3339, "1970-01-01T00:00:00Z")
+	timeText, _ := beginningOfTime.MarshalText()
+	websession.Values[sessionCreationTimeKey] = string(timeText)
```
hmahmood:
```go
time.Unix(0, 0).String()
```

cmd/vicadmin/server.go
```diff
@@ -47,10 +53,26 @@ const (
...
+)
+
 func (s *server) listen() error {
 	defer trace.End(trace.Begin(""))
 
 	var err error
+	s.uss = NewUserSessionStore()
+	s.uss.cookies.Options = &sessions.Options{
+		Path:     "/",
+		MaxAge:   3600 * 24,
```
anchal-agrawal: This could be a constant.

cmd/vicadmin/server.go
```diff
@@ -112,39 +134,203 @@ func (s *server) listenPort() int {
...
+
+		if len(r.TLS.PeerCertificates) > 0 { // the user is authenticated by certificate at connection time
+			usersess := s.uss.Add("root", &rootConfig.Config)
+
+			timeNow, err := usersess.created.MarshalText()
+			if err != nil {
+				// it's probably safe to ignore this error since we just created usersess.created when we called Add() above
+				// but just in case..
+				log.Errorf("Failed to unmarshal time object %+v into text due to error: %s", usersess.created, err)
+				http.Redirect(w, r, authFailure, 302)
```
anchal-agrawal: Can we replace 302 with`http.StatusFound`?

cmd/vicadmin/usersession.go
```diff
@@ -0,0 +1,101 @@
...
+
+// Grabs the UserSession metadta object and doesn't establish a connection to vSphere
+func (u *UserSessionStore) Get(username string) *UserSession {
+	u.mutex.RLock()
+	defer u.mutex.RUnlock()
+	return u.sessions[username]
+}
+
+// Get logs into vSphere and returns a vSphere session object. Caller responsible for error handling/logout
+func (u *UserSessionStore) VSphere(username string) (vSphereSession *session.Session, err error) {
```
gigawhitlocks: no we cannot because it would become unexported.
I also omitted 'Session' because of semantic saturation of the term. The whole object is sessions.

cmd/vicadmin/server.go
```diff
@@ -112,39 +134,203 @@ func (s *server) listenPort() int {
...
+		// redirect to dashboard
+		http.Redirect(res, req, "/", 302)
+		return
+	}
+
+	// Render login page (shows up on non-POST requests):
+	sess, err := client(&rootConfig)
+	if err != nil {
+		log.Errorf("Could not render login page due to vSphere connection error: %s", err.Error())
+		http.Redirect(res, req, "/", 500)
```
gigawhitlocks: Yeah this should be an error, as we discussed IRL

cmd/vicadmin/server.go
gigawhitlocks: Probably yeah, I didn't see the constants for these.

cmd/vicadmin/server.go
gigawhitlocks: Yeah I gotta go through and replace this timeout in a bunch of places w/ the same constant.

cmd/vicadmin/server.go
gigawhitlocks: yeap you're right, we can ignore this error.

cmd/vicadmin/server.go
gigawhitlocks: (`http.StatusTemporaryRedirect` actually :+1: but thanks for helping me find the status code constants)

cmd/vicadmin/server.go
```diff
@@ -112,39 +127,208 @@ func (s *server) listenPort() int {
...
+		if c == nil { // no cookie, so redirect to login
+			log.Errorf("No authentication token: %+v", websession.Values)
+			http.Redirect(w, r, loginPagePath, http.StatusTemporaryRedirect)
+			return
+		}
+
+		// here we have a cookie, but we need to make sure it's not expired:
+		// parse the cookie creation time
+		created, err := time.Parse(time.RFC3339, c.(string))
+		if err != nil {
```
andrewtchin: Can this case just be treated as expired? Would the user have to manually clear cookies to login?

cmd/vicadmin/server.go
```diff
@@ -166,36 +350,49 @@ func (s *server) stop() error {
...
+func (s *server) logoutHandler(res http.ResponseWriter, req *http.Request) {
+	websession, _ := s.uss.cookies.Get(req, sessionCookieKey)
+	// ignore parsing/marshalling errors because we're parsing a hardcoded beginning-of-time string
+	beginningOfTime := time.Unix(0, 0).Format(time.RFC3339)
+	websession.Values[sessionCreationTimeKey] = beginningOfTime
+	if err := websession.Save(req, res); err != nil {
+		http.Error(res, "Failed to expire user session", http.StatusInternalServerError)
+		return
+	}
+	s.uss.Delete(websession.Values[sessionUsernameKey].(string))
```
andrewtchin: Should the Delete happen even if the Save has an error?

isos/vicadmin/auth.html
```diff
@@ -0,0 +1,59 @@
+<!DOCTYPE html>
+<!-- saved from url=(0030)http://10.20.232.35/community/ -->
```
andrewtchin: is this junk?

andrewtchin:
lgtm ⭐️ for commented code

cmd/vicadmin/server.go
```diff
@@ -47,11 +52,21 @@ const (
...
 )
 
+const (
+	sessionExpiration      = time.Hour * 24
+	sessionCookieKey       = "sessiondata"
+	sessionCreationTimeKey = "created"
+	sessionUsernameKey     = "username"
+	loginPagePath          = "/authentication"
+	authFailure            = loginPagePath + "?unauthorized"
+	genericErrorMessage    = "Internal Server Error; see /var/log/vic/vicadmin.log for details" // for http errors that shouldn't be displayed in the browser to the user
```
hmahmood: Reference to /var/log/vic/vicadmin.log does not really mean anything to customers or api callers. Instead a reference to the vicadmin portal might be better.

isos/vicadmin/auth.html
hmahmood: This should be removed

hmahmood:
lgtm

cmd/vicadmin/server.go
```diff
@@ -112,39 +127,208 @@ func (s *server) listenPort() int {
 	return s.l.Addr().(*net.TCPAddr).Port
 }
 
+// Enforces authentication on route `link` and runs `handler` on successful auth
+func (s *server) AuthenticatedHandle(link string, h http.Handler) {
```
hickeng: This and `Authenticated` seem to have exactly the same signature - presumably there's a reason why we have both?

isos/vicadmin/auth.html
gigawhitlocks: whoa! yeah I must've accidentally pasted into a comment in the HTML LOL

cmd/vicadmin/server.go
gigawhitlocks: If this error is displayed vicadmin is broken -- how else would they get logs?

cmd/vicadmin/server.go
```diff
@@ -112,39 +127,208 @@ func (s *server) listenPort() int {
...
+	}
+
+	// Render login page (shows up on non-POST requests):
+	sess, err := client(&rootConfig)
+	if err != nil {
+		log.Errorf("Could not render login page due to vSphere connection error: %s", err.Error())
+		http.Error(res, genericErrorMessage, http.StatusInternalServerError)
+		return
+	}
+	v := vicadmin.NewValidator(ctx, &vchConfig, sess)
```
hickeng: Why are we using the validator during the login path?

cmd/vicadmin/server.go
```diff
@@ -112,39 +127,208 @@ func (s *server) listenPort() int {
...
-	s.mux.HandleFunc("/container-logs.tar.gz", s.tarContainerLogs)
-	s.mux.HandleFunc("/container-logs.zip", s.zipContainerLogs)
+	s.Authenticated("/container-logs.tar.gz", s.tarContainerLogs)
+	s.Authenticated("/container-logs.zip", s.zipContainerLogs)
 
-	s.mux.Handle("/css/", http.StripPrefix("/css/", http.FileServer(http.Dir("css/"))))
-	s.mux.Handle("/images/", http.StripPrefix("/images/", http.FileServer(http.Dir("images/"))))
-	s.mux.Handle("/fonts/", http.StripPrefix("/fonts/", http.FileServer(http.Dir("fonts/"))))
+	// these assets bypass authentication & are world-readable
+	s.Handle("/css/", http.StripPrefix("/css/", http.FileServer(http.Dir("css/"))))
```
hickeng: http.FileServer docs says that the path should be sanitized - it's not clear whether this is occurring anywhere on this path.
Also - why are we not requiring auth for these assets?

tests/resources/Util.robot
```diff
@@ -134,7 +134,7 @@ Run VIC Machine Command
...
     Return From Keyword If  ${certs}  ${output}
 
-    ${output}=  Run Keyword Unless  ${certs}  Run  ${vic-machine} create --debug 1 --name=${vch-name} --target=%{TEST_URL}%{TEST_DATACENTER} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --image-store=%{TEST_DATASTORE} --appliance-iso=${appliance-iso} --bootstrap-iso=${bootstrap-iso} --password=%{TEST_PASSWORD} --force=true --bridge-network=%{BRIDGE_NETWORK} --external-network=%{EXTERNAL_NETWORK} --compute-resource=%{TEST_RESOURCE} --timeout %{TEST_TIMEOUT} --volume-store=%{TEST_DATASTORE}/test:${vol} --no-tls
+    ${output}=  Run Keyword Unless  ${certs}  Run  ${vic-machine} create --debug 1 --name=${vch-name} --target=%{TEST_URL}%{TEST_DATACENTER} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --image-store=%{TEST_DATASTORE} --appliance-iso=${appliance-iso} --bootstrap-iso=${bootstrap-iso} --password=%{TEST_PASSWORD} --force=true --bridge-network=%{BRIDGE_NETWORK} --external-network=%{EXTERNAL_NETWORK} --compute-resource=%{TEST_RESOURCE} --timeout %{TEST_TIMEOUT} --volume-store=%{TEST_DATASTORE}/test:${vol} --no-tlsverify
```
gigawhitlocks: `--no-tlsverify` turns off client certificate verification (authentication) but leaves tls/https active (just encryption) and -- more importantly for my use case -- then the vicadmin url is generated correctly (no-tls gives http:// which is incorrect; vicadmin is always https://, and no-tlsverify gives https://) 

In any case my understanding is that no-certs mostly exists for running tests locally, and it's really just tlsverify that we don't want because creating client certs is hard (DNS issues everywhere) in local testing environments.

cmd/vicadmin/server.go
```diff
@@ -166,36 +350,49 @@ func (s *server) stop() error {
...
+	websession, _ := s.uss.cookies.Get(req, sessionCookieKey)
+	// ignore parsing/marshalling errors because we're parsing a hardcoded beginning-of-time string
+	beginningOfTime := time.Unix(0, 0).Format(time.RFC3339)
+	websession.Values[sessionCreationTimeKey] = beginningOfTime
+	if err := websession.Save(req, res); err != nil {
+		http.Error(res, "Failed to expire user session", http.StatusInternalServerError)
+		return
+	}
+	s.uss.Delete(websession.Values[sessionUsernameKey].(string))
+	http.Redirect(res, req, "/", http.StatusTemporaryRedirect)
```
hickeng: This redirect will in turn redirect to the login page - should we just send the user there directly?

cmd/vicadmin/server.go
```diff
@@ -112,39 +127,208 @@ func (s *server) listenPort() int {
...
+	}
+}
+
 func (s *server) serve() error {
 	defer trace.End(trace.Begin(""))
 
 	s.mux = http.NewServeMux()
 
+	// s.mux.HandleFunc bypasses authentication
+	s.mux.HandleFunc(loginPagePath, s.loginPage)
```
hickeng: If a logged in user hits the login page, what's our expectation?

cmd/vicadmin/server.go
```diff
@@ -112,39 +127,208 @@ func (s *server) listenPort() int {
...
+func (s *server) Handle(link string, h http.Handler) {
+	s.mux.Handle(link, gorillacontext.ClearHandler(h))
+}
+
+// Enforces authentication on route `link` and runs `handler` on successful auth
+func (s *server) Authenticated(link string, handler func(http.ResponseWriter, *http.Request)) {
+	defer trace.End(trace.Begin(""))
+
+	authHandler := func(w http.ResponseWriter, r *http.Request) {
+		websession, _ := s.uss.cookies.Get(r, sessionCookieKey) // ignore error because it is okay if it doesn't exist
```
hickeng: can we confirm that the originating IP address is the expected one? Is that an inherent aspect of the cookie encryption you mention?

cmd/vicadmin/server.go
gigawhitlocks: The path should be sanitized? the hardcoded `"/css"` bits? Sorry, bear with me as I just changed these to match my auth/noauth pattern and did not investigate into whether the internal call was done correctly

cmd/vicadmin/server.go
gigawhitlocks: Yeah good catch

cmd/vicadmin/server.go
```diff
@@ -166,36 +350,49 @@ func (s *server) stop() error {
 	return nil
 }
 
+// logout handler expires the user's session cookie by setting its creation time to the beginning of time
+func (s *server) logoutHandler(res http.ResponseWriter, req *http.Request) {
+	websession, _ := s.uss.cookies.Get(req, sessionCookieKey)
+	// ignore parsing/marshalling errors because we're parsing a hardcoded beginning-of-time string
+	beginningOfTime := time.Unix(0, 0).Format(time.RFC3339)
```
hickeng: consider caching this as a constant

cmd/vicadmin/server.go
gigawhitlocks: How would we know if it was expected?

cmd/vicadmin/server.go
gigawhitlocks: This is because the routine that renders HTML templates uses the Validator and I didn't take the time to refactor out the requirement.

cmd/vicadmin/server.go
```diff
@@ -112,39 +127,208 @@ func (s *server) listenPort() int {
...
+			http.Error(res, "Authentication failed due to incorrect credential(s)", 400)
+			return
+		}
+
+		// successful login above; user is authenticated
+		// log out, disregard errors
+		usersession.Client.Logout(context.Background())
+
+		// save user config locally
+		usersess := s.uss.Add(req.FormValue("username"), &userconfig)
```
gigawhitlocks: Yeah I'm actually facepalming over here at this one -- you're absolutely right and in fact what you're describing was the design I intended and I didn't consider the case where the username might not be unique due to multiple users sharing credentials. I'll tie this key to the cookie somehow.. I think that should do it..

cmd/vicadmin/server.go
gigawhitlocks: I think so yeah

cmd/vicadmin/server.go
gigawhitlocks: Yep just checked and `websession` has a `string` `ID` so I'll use that as the unique identifier instead of the username -- I feel really silly for using the username actually hahaha

cmd/vicadmin/server.go
gigawhitlocks: yeah makes sense, no reason to do this each time

tests/resources/Util.robot
gigawhitlocks: I did this as a workaround though and it might've broken the Compose suite, still testing that. Hopefully I don't need to do anything more complex.

cmd/vicadmin/server.go
hickeng: We can embed a pointer to a help page once we have stable docs ;)

cmd/vicadmin/server.go
hickeng: @gigawhitlocks that's what I meant. Means that if someone can steal a cookie from an authenticated machine there's just more hoops to get through to use it.

cmd/vicadmin/server.go
hickeng: If that's the solution, could we have a "Click here to log in again" or similar?

cmd/vicadmin/server.go
hickeng: We should get this done for 0.8 without doubt. If it doesn't go in this PR, please open an issue to do so as it could easily slip through cracks otherwise.

cmd/vicadmin/server.go
hickeng: ok - fine with that.

cmd/vicadmin/server.go
hickeng: There is some code that disallows `..` in paths for example (in the FileServer), but only as a basic precaution - the comments on that package say you should sanitize the paths as the path is essentially untrusted user input

cmd/vicadmin/server.go
gigawhitlocks: I'll open an issue for it. I'm more than happy to finish up the cleanup work in this package (the unit tests need a lot of love too) once this PR is in -- I'd just keep going on this one but it's so big as it is

cmd/vicadmin/server.go
gigawhitlocks: According to the docs it's only an issue if the path is user-provided, which in this case it is not:

> If the provided name is constructed from user input, it should be sanitized before calling ServeFile. As a precaution, ServeFile will reject requests where r.URL.Path contains a ".." path element. 

So `..` is automatically excluded and we hard code the directory to serve -- maybe I'm being daft but I don't see an issue with this usage?

tests/resources/Util.robot
mhagen-vmware: There are several places in functional and docker integration tests that requires --no-tls for it to work. You can go ahead with this issue, but it sounds like there is a bug we need to fix around http vs https? Do we have an issue for it already?  I will fix any fall out from this on the functional and docker integration side.

tests/test-cases/Group9-VIC-Admin/9-01-VICAdmin-ShowHTML.robot
```diff
@@ -6,20 +6,65 @@ Suite Teardown  Cleanup VIC Appliance On Test Server
...
 
 *** Test Cases ***
-Display HTML
+Get Login Page
+    ${rc}  ${output}=  Run And Return Rc And Output  curl -sk ${vic-admin}/authentication
+    Should contain  ${output}  <title>VCH Admin</title>
+
+While Logged Out Fail To Display HTML
     ${rc}  ${output}=  Run And Return Rc And Output  curl -sk ${vic-admin}
+    Should not contain  ${output}  <title>VIC: ${vch-name}</title>
```
mhagen-vmware: This is a bit weak and dangerous, can you change this to a positive? Should contain HTTP 400 or something?

tests/test-cases/Group9-VIC-Admin/9-01-VICAdmin-ShowHTML.robot
```diff
@@ -6,20 +6,65 @@ Suite Teardown  Cleanup VIC Appliance On Test Server
...
+
+While Logged Out Fail To Get Portlayer Log
+    ${rc}  ${output}=  Run And Return Rc And Output  curl -sk ${vic-admin}/logs/port-layer.log
+    Should Not Contain  ${output}  Launching portlayer server
```
mhagen-vmware: This one and the ones below it are more specific and less dangerous, but it would still be good to add what you actually expect the user to see.

tests/test-cases/Group9-VIC-Admin/9-01-VICAdmin-ShowHTML.robot
```diff
@@ -6,20 +6,65 @@ Suite Teardown  Cleanup VIC Appliance On Test Server
...
+While Logged Out Fail To Get Container Logs
+    ${rc}  ${output}=  Run And Return Rc and Output  docker ${params} pull busybox
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${output}  Error
+    ${rc}  ${container}=  Run And Return Rc and Output  docker ${params} create busybox /bin/top
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${container}  Error
+    ${rc}  ${output}=  Run And Return Rc and Output  docker ${params} start ${container}
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${output}  Error
```
mhagen-vmware: just a simple docker run would be clearer and easier here.

docker run -itd busybox /bin/top or so

tests/test-cases/Group9-VIC-Admin/9-01-VICAdmin-ShowHTML.robot
```diff
@@ -6,20 +6,65 @@ Suite Teardown  Cleanup VIC Appliance On Test Server
...
+    ${rc}  ${output}=  Run And Return Rc and Output  curl -sk ${vic-admin}/container-logs.tar.gz | tar tvzf -
+    Should not Be Equal As Integers  ${rc}  0
+    Log  ${output}
+    Should not Contain  ${output}  ${container}/vmware.log
+    Should not Contain  ${output}  ${container}/tether.debug
```
mhagen-vmware: similar comment as above. positive case would be more safe here

tests/test-cases/Group9-VIC-Admin/9-01-VICAdmin-ShowHTML.robot
```diff
@@ -6,20 +6,65 @@ Suite Teardown  Cleanup VIC Appliance On Test Server
...
+
+While Logged Out Fail To Get VICAdmin Log
+    ${rc}  ${output}=  Run And Return Rc And Output  curl -sk ${vic-admin}/logs/vicadmin.log
+    Log  ${output}
+    Should not contain  ${output}  Launching vicadmin pprof server
+
+Login
+    ${rc}  ${output}=  Run And Return Rc And Output  curl -sk ${vic-admin}/authentication -XPOST -F username=%{GOVC_USERNAME} -F password=%{GOVC_PASSWORD} -D /tmp/cookies-${vch-name}
+    Should Be Equal As Integers  ${rc}  0
```
mhagen-vmware: Leave this the way it is for now, but please create a follow on issue and assign it to me for me to clean this up and not have tests rely on previous tests.

gigawhitlocks: Oh I thought that was OK in a single suite but sure, sorry.

tests/test-cases/Group9-VIC-Admin/9-01-VICAdmin-ShowHTML.robot
gigawhitlocks: It's the same as the other docker logs test, didn't want to change the meat of the test, just wanted to demonstrate that we could do the same thing and not be able to see it. If this one's unclear we should refactor the other one too. This PR is out of control so please add a comment to #2955 and we'll take care of cleaning up this test while addressing that issue.

tests/test-cases/Group9-VIC-Admin/9-01-VICAdmin-ShowHTML.robot
gigawhitlocks: Added a check for the `302 Redirect` that sends the user to `/authentication` to each of these tests as a positive case.

gigawhitlocks:
♻️

tests/resources/Util.robot
gigawhitlocks: The only place I found was the Compose Voting App test and I updated that to work with `--no-tlsverify` -- there really shouldn't be anything that requires `--no-tls` at this point, just `--no-tlsverify`.

PR #2884, opened 2016-10-25
Tether better handshake

vburenin:
With sync logic I had no failures attaching to container. With previous logic attachment could fail sometimes.

Fixes: #2889


vburenin:
Closing for now until failing tests are fixed.

pkg/serial/handshake.go
```diff
@@ -162,35 +158,11 @@ func HandshakeClient(ctx context.Context, conn net.Conn, debug bool) error {
 	return nil
 }
 
-func readMultiple(conn net.Conn, b []byte) (int, error) {
-	if runtime.GOOS != "windows" {
-		return conn.Read(b)
-	}
-
-	// we want a blocking read, but the behaviour described in the remarks section
```
hickeng: please don't just remove the comments relating to windows - they'll be useful to have around for when we pick that work up again.

pkg/serial/handshake.go
```diff
@@ -60,58 +60,57 @@ func HandshakeClient(ctx context.Context, conn net.Conn, debug bool) error {
...
-	log.Debugf("HandshakeClient: received synack: %#x == %#x\n", synack, buf[:2])
+	ack := []byte{flagAck, synAck[2] + 1}
 	log.Debug("client: writing ack")
-	ack[1] = buf[2] + 1
 	if debug {
 		ack[0] = flagDebugAck
 	}
 	conn.Write(ack)
 
+	// Ensure both sides are in sync. Both sides will send the same message
```
hickeng: Why do this at all? I cannot see that it adds any value.
If we've somehow stumbled into a correct handshake sequence that's still misaligned then the SSH protocol handshake will fail (which is essentially sending the same known string on both side in exactly the same way this is) and we'll retry from the top, the same way we do here.


The lossiness check below was added to assert that we weren't losing data during debugging but this is essentially a second handshake

pkg/serial/handshake.go
```diff
@@ -200,48 +172,56 @@ func HandshakeServer(ctx context.Context, conn net.Conn) error {
...
-		}
-
-		// to aid in debug we always dump the full handhsake
-		log.Debug(msg)
-		log.Debugf("server: read %v bytes: ", n)
-		for i := 0; i < n; i++ {
-			log.Debugf("%#x ", syn[i])
-		}
-		log.Debug("")
+	// syn is 2 bytes as that will eventually syn us again if we're offset
```
hickeng: This comment is no longer correct - if you're sending and reading 2 bytes then there is nothing in this piece of code that will result in eventual window realignment.

pkg/serial/handshake.go
```diff
@@ -200,48 +172,56 @@ func HandshakeServer(ctx context.Context, conn net.Conn) error {
...
-	ack[1] = synack[2] + 1
+	ack := []byte{flagAck, synackBuf[2] + 1}
 	log.Debug("server: reading ack")
-	readMultiple(conn, buf)
-	if (buf[0] != flagAck && buf[0] != flagDebugAck) || bytes.Compare(ack[1:], buf[1:]) != 0 {
-		msg := fmt.Sprintf("server: did not receive ack: %#x != %#x", ack, buf)
-		log.Debug(msg)
+
+	buf := make([]byte, 2)
+	io.ReadFull(conn, buf)
```
hickeng: This will not rapidly drain spurious data on the line as you're only reading 2 bytes max. You'll have to loop repeatedly to do so.

pkg/serial/handshake.go
vburenin: This comment is not longer relevant since new logic reads one byte at the time.

lib/portlayer/attach/connector.go
```diff
@@ -167,7 +167,7 @@ func (c *Connector) processIncoming(conn net.Conn) {
 		// the tether side (in tether_linux.go) or alignment may not happen.
 		// The PL sends the first SYN in the handshake and if the tether is not
 		// waiting, the handshake may never succeed.
-		ctx, cancel := context.WithTimeout(context.TODO(), 50*time.Millisecond)
+		ctx, cancel := context.WithTimeout(context.TODO(), time.Second)
```
fdawg4l: Can you update the comment? Looks like the ticker has been moved to `cmd/tether/attach.go`.  Also why is the timeout being made larger?

pkg/serial/handshake.go
```diff
@@ -56,19 +63,16 @@ func PurgeIncoming(conn net.Conn) {
...
-	syn := make([]byte, 2)
-	synack := make([]byte, 2)
-	ack := make([]byte, 2)
 
-	buf := make([]byte, 255)
-
-	syn[0] = flagSyn
-	synack[0] = flagAck
-	ack[0] = flagAck
+	buf1byte := make([]byte, 1)
```
fdawg4l: The variable name `buf1byte` now includes the cap and the type.  Is that really necessary?  `buf`, I think, was fine.

pkg/serial/handshake.go
```diff
@@ -77,212 +81,201 @@ func HandshakeClient(ctx context.Context, conn net.Conn, debug bool) error {
...
+			// read syn position
+			if _, err := conn.Read(buf1byte); err != nil {
+				return err
+			}
+			log.Debug("HandshakeClient: Sending ack")
+			if debug {
+				conn.Write([]byte{flagDebugAck, nextPos(buf1byte[0])})
+			} else {
+				conn.Write([]byte{flagAck, nextPos(buf1byte[0])})
+				break loop
```
fdawg4l: `break` just to go to `loop:`.  Why not just use `continue`?

pkg/serial/handshake.go
```diff
@@ -77,212 +81,201 @@ func HandshakeClient(ctx context.Context, conn net.Conn, debug bool) error {
...
-	conn.SetReadDeadline(time.Time{})
-
-	if buf[0] == flagDebugAck {
-		// Check for lossiness
-		rxbuf := make([]byte, 23)
-		log.Debugf("Checking for lossiness")
-		n, err := io.ReadFull(conn, rxbuf)
-		if err != nil {
-			log.Error(err)
+loop:
```
fdawg4l: Same comment as above.  Why not use `continue`?

vburenin:
Last change I added logic that will try to drain buffer as soon as possible. As well as added a comment to increased timeout. It also happened so, that 50 milliseconds timeout I increased to 1 seconds was critically small that could cause a flood of syn attempts to the server without actually receiving a response from server leading to one more sync to be sent. 1 seconds resolves this problem, however, it is not an accurate solution.

pkg/serial/handshake.go
```diff
@@ -56,19 +54,16 @@ func PurgeIncoming(conn net.Conn) {
...
-	synack := make([]byte, 2)
-	ack := make([]byte, 2)
-
-	buf := make([]byte, 255)
 
-	syn[0] = flagSyn
-	synack[0] = flagAck
-	ack[0] = flagAck
+	buf1byte := make([]byte, 1)
+	rand.Read(buf1byte)
```
caglar10ur:
```
if _, err := io.ReadFull(rand.Reader, buf1byte); err != nil {
    panic(err) // This shouldn't happen
}
```
maybe?

vburenin: it easily can happen if reader is closed or times out. I would not panic if it is not critical.

pkg/serial/handshake.go
```diff
@@ -77,212 +72,188 @@ func HandshakeClient(ctx context.Context, conn net.Conn, debug bool) error {
 		conn.SetReadDeadline(deadline)
 	}
 
-	rand.Read(syn[1:])
-
-	conn.Write(syn)
+	log.Debugf("HandshakeClient: Establishing connection...")
+	for {
```
hickeng: The client should not be running in a for loop at this point - there needs to be a purge incoming call before each iteration to deal with the case of queued incoming bytes - otherwise each byte will trigger sending a new syn packet, which can trigger more NACKs from the server side.

This loop needs to run slower than the consumption on server side to prevent a self-sustaining bouncing of syn/Nack and there's no throttle.

pkg/serial/handshake.go
```diff
@@ -77,212 +72,188 @@ func HandshakeClient(ctx context.Context, conn net.Conn, debug bool) error {
...
-	if n, err := readMultiple(conn, syn); n != 2 || err != nil || syn[0] != flagSyn {
-		var msg string
+	buf1byte := make([]byte, 1)
+	syncBuf := make([]byte, 4096)
+
+	rand.Read(buf1byte)
+	pos := nextPos(buf1byte[0])
+
+	log.Debugf("HandshakeServer: Establishing connection...")
+	for {
```
hickeng: Check `ctx.Error() != nil` so that we can cancel - this won't allow us to cancel an inflight read, and there's no timeout support in rawconn so there's no guarantee we'll hit the condition in a timely fashion, but it does mean we have an exit path other than EOF.

pkg/serial/handshake.go
```diff
@@ -77,212 +72,188 @@ func HandshakeClient(ctx context.Context, conn net.Conn, debug bool) error {
...
-			msg = fmt.Sprintf("server: received SYN (read %d) bytes", n)
+			return err
 		}
 
-		// to aid in debug we always dump the full handhsake
-		log.Debug(msg)
-		log.Debugf("server: read %v bytes: ", n)
-		for i := 0; i < n; i++ {
-			log.Debugf("%#x ", syn[i])
+		if n > 2 {
```
hickeng: Try assuming that `syncBuf[n-2:]` is the most _recent_ syn packet and continue with that. It works if there are only two bytes read as well as if there are 20 queued packets.

vburenin: Yes, it is even better.

pkg/serial/handshake.go
vburenin: because in this case break works within select clause.

caglar10ur:
LGTM

caglar10ur:
Please make sure @hickeng gives 👍  before merging this :)

cmd/tether/attach.go
```diff
@@ -255,12 +255,23 @@ func backchannel(ctx context.Context, conn *net.Conn) error {
 			if ctx.Err() != nil {
 				return ctx.Err()
 			}
-			err := serial.HandshakeServer(ctx, *conn)
-			if err != nil {
-				log.Error(err)
-			} else {
+			deadline, ok := ctx.Deadline()
+			if ok {
+				conn.SetReadDeadline(deadline)
```
hickeng: this has zero effect when backed by a rawconn - the underlying file Read has no timeout.
The only way to interrupt the call is by closing the fd.

Should likely change rawconn:SetReadDeadline, et al, to panic so there's no confusion now we have more people in this part of the code.

pkg/serial/handshake.go
```diff
@@ -54,206 +59,208 @@ func PurgeIncoming(conn net.Conn) {
...
-	if ok {
-		conn.SetReadDeadline(deadline)
-	}
 
-	log.Debugf("HandshakeClient: Establishing connection...")
-	for {
-		log.Debug("HandshakeClient: Sending syn.")
-		conn.Write([]byte{flagSyn, pos})
-		pos = nextPos(pos)
+	log.Debug("HandshakeClient: Sending syn.")
```
hickeng: The original reason for looking at this was to determine what was taking all the CPU time. On the tether side it turned out to be I/O, of which a LOT was logging.

In other packages we've ended up exposing a package level variable `verbose` and/or `trace` and guarding some of the more chatty debug messages with those.
Consider a similar approach, hooked into the VCH config DebugLevel value (probably >=2) if this turns out to generate a lot of noise in the logs.

pkg/serial/handshake_test.go
```diff
@@ -0,0 +1,387 @@
...
+	}
+}
+
+func TestHandshakeServerComportSync(t *testing.T) {
+	log.SetLevel(log.InfoLevel)
+	clientConn, serverConn := NewFakeConnection(time.Second * 3)
+
+	go func() {
+		buf := make([]byte, 10)
+		clientConn.Write([]byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22})
```
hickeng: Can you add a test that seeds the line (so the client will read them) with these bytes, and has the client and server hooked together as in `TestHandshakeServerVsClient` - that's the normal case that we see between containerVM and applianceVM. It's the containerVM boot sequence that generates this byte sequence.

lib/portlayer/attach/connector.go
```diff
@@ -168,16 +168,24 @@ func (c *Connector) processIncoming(conn net.Conn) {
 		// 2 or more bytes it will just wait, so client should timeout.
 		// However, if timeout is too short, client will flood server with Syn requests.
 		ctx, cancel := context.WithTimeout(context.TODO(), time.Second)
-		if err = serial.HandshakeClient(ctx, conn, c.debug); err == nil {
+		deadline, ok := ctx.Deadline()
+		if ok {
+			conn.SetReadDeadline(deadline)
+		}
+		if err = serial.HandshakeClient(conn, c.debug); err == nil {
```
hickeng: FYI, removing ctx from the signature is going to make it harder to add trace.Operation based logging when it comes.

cmd/tether/attach.go
```diff
@@ -255,10 +255,23 @@ func backchannel(ctx context.Context, conn *net.Conn) error {
 			if ctx.Err() != nil {
 				return ctx.Err()
 			}
-			err := serial.HandshakeServer(ctx, *conn)
+			deadline, ok := ctx.Deadline()
+			if ok {
+				conn.SetReadDeadline(deadline)
```
vburenin: yah, I just keep that logic as it is.

pkg/serial/handshake.go
vburenin: This is a negligible amount of logging comparable to what is logged before, it is even hard to find a trace of it.

hickeng:
LGTM

PR #2887, opened 2016-10-25
fix container vm invalid state issue

### LINKED ISSUE #2698: faild to start docker container after vm enter into invalid state
Upgrade tests hit this while trying to power a VM back on.

https://ci.vmware.run/vmware/vic/5693

As this is an issue different with #1823, open a new issue here.
@hickeng @chengwang86

emlin:
Fixes #2698 

If any vm operation returned invalid state error, will try to unregister and register vm, and then retry task.
As unregister vm will remove container from portlayer container cache, add event listening for vm registered event.

### DIFF
--- a/lib/portlayer/exec/exec.go
+++ b/lib/portlayer/exec/exec.go
@@ -24,6 +24,8 @@ import (
 
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/lib/portlayer/event"
 	"github.com/vmware/vic/lib/portlayer/event/collector/vsphere"
 	"github.com/vmware/vic/lib/portlayer/event/events"
@@ -31,11 +33,15 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/session"
 )
 
-var initializer sync.Once
+var (
+	initializer  sync.Once
+	eventSession *session.Session
+)
 
 func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSource, _ extraconfig.DataSink) error {
 	var err error
 	initializer.Do(func() {
+		eventSession = sess
 		f := find.NewFinder(sess.Vim25(), false)
 
 		extraconfig.Decode(source, &Config)
@@ -152,8 +158,37 @@ func eventCallback(ie events.Event) {
 
 			}
 		}
+		return
 	}
+	// check container registered event if this container is not found in container cache
+	switch ie.String() {
+	case events.ContainerRegistered:
+		moref := new(types.ManagedObjectReference)
+		if ok := moref.FromString(ie.Reference()); !ok {
+			log.Errorf("Failed to get event VM mob reference")
+			return
+		}
+		var vm mo.VirtualMachine
 
+		// current attributes we care about
+		attrib := []string{"config", "runtime.powerState", "summary", "resourcePool"}
+
+		// populate the vm properties
+		log.Infof("moref: %s", *moref)
+		ctx := context.Background()
+		if err := eventSession.RetrieveOne(ctx, *moref, attrib, &vm); err != nil {
+			log.Errorf("Failed to query registered vm object %s: %s", ie.Reference(), err)
+			return
+		}
+		if *vm.ResourcePool != Config.ResourcePool.Reference() {
+			log.Debugf("vm %q is not containers of this VCH, ignore it", vm.Config.Name)
+		}
+		registeredContainers := convertInfraContainers(ctx, eventSession, []mo.VirtualMachine{vm})
+		for i := range registeredContainers {
+			Containers.put(registeredContainers[i])
+			log.Debugf("Container %q is registered back", vm.Config.Name)
+		}
+	}
 	return
 }
 
--- a/lib/portlayer/exec/exec.go
+++ b/lib/portlayer/exec/exec.go
@@ -168,27 +168,42 @@ func eventCallback(ie events.Event) {
 			log.Errorf("Failed to get event VM mobref: %s", ie.Reference())
 			return
 		}
-		var vm mo.VirtualMachine
-
-		// current attributes we care about
-		attrib := []string{"config", "runtime.powerState", "summary", "resourcePool"}
-
-		// populate the vm properties
-		ctx := context.Background()
-		if err := eventSession.RetrieveOne(ctx, *moref, attrib, &vm); err != nil {
-			log.Errorf("Failed to query registered vm object %s: %s", ie.Reference(), err)
+		if !isManagedbyVCH(*moref) {
 			return
 		}
-		if *vm.ResourcePool != Config.ResourcePool.Reference() {
-			log.Debugf("container vm %q does not belong to this VCH, ignoring", vm.Config.Name)
+		log.Debugf("Register container VM %s", moref)
+		ctx := context.Background()
+		vms, err := populateVMAttributes(ctx, eventSession, []types.ManagedObjectReference{*moref})
+		if err != nil {
+			log.Error(err)
+			return
 		}
-		registeredContainers := convertInfraContainers(ctx, eventSession, []mo.VirtualMachine{vm})
+		registeredContainers := convertInfraContainers(ctx, eventSession, vms)
 		for i := range registeredContainers {
 			Containers.put(registeredContainers[i])
-			log.Debugf("Registered container %q", vm.Config.Name)
+			log.Debugf("Registered container %q", registeredContainers[i].Config.Name)
 		}
 	}
 	return
+}
+
+func isManagedbyVCH(moref types.ManagedObjectReference) bool {
+	var vm mo.VirtualMachine
+
+	// current attributes we care about
+	attrib := []string{"config"}
+
+	// populate the vm properties
+	ctx := context.Background()
+	if err := eventSession.RetrieveOne(ctx, moref, attrib, &vm); err != nil {
+		log.Errorf("Failed to query registered vm object %s: %s", moref.String(), err)
+		return false
+	}
+	if *vm.ResourcePool != Config.ResourcePool.Reference() {
+		log.Debugf("container vm %q does not belong to this VCH, ignoring", vm.Config.Name)
+		return false
+	}
+	return true
 }
 
 // eventedState will determine the target container
--- a/lib/portlayer/exec/exec.go
+++ b/lib/portlayer/exec/exec.go
@@ -34,13 +34,20 @@ import (
 )
 
 var (
-	initializer sync.Once
-	initError   error
+	initializer struct {
+		err  error
+		once sync.Once
+	}
 )
 
 func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSource, _ extraconfig.DataSink) error {
-	var err error
-	initializer.Do(func() {
+	initializer.once.Do(func() {
+		var err error
+		defer func() {
+			if err != nil {
+				initializer.err = err
+			}
+		}()
 		f := find.NewFinder(sess.Vim25(), false)
 
 		extraconfig.Decode(source, &Config)
@@ -93,7 +100,7 @@ func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSou
 		Config.EventManager.Subscribe(events.NewEventType(vsphere.VMEvent{}).Topic(), "exec", eventCallback)
 		// subscribe callback to handle vm registered event
 		Config.EventManager.Subscribe(events.NewEventType(vsphere.VMEvent{}).Topic(), "registeredVMEvent", func(ie events.Event) {
-			registeredVMCallback(ie, sess)
+			registeredVMCallback(sess, ie)
 		})
 
 		// instantiate the container cache now
@@ -114,10 +121,7 @@ func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSou
 			return
 		}
 	})
-	if err != nil {
-		initError = err
-	}
-	return initError
+	return initializer.err
 }
 
 // eventCallback will process events
@@ -168,7 +172,7 @@ func eventCallback(ie events.Event) {
 }
 
 // registeredVMCallback will process registeredVMEvent
-func registeredVMCallback(ie events.Event, sess *session.Session) {
+func registeredVMCallback(sess *session.Session, ie events.Event) {
 	// check container registered event if this container is not found in container cache
 	// grab the container from the cache
 	container := Containers.Container(ie.Reference())
@@ -183,7 +187,7 @@ func registeredVMCallback(ie events.Event, sess *session.Session) {
 			log.Errorf("Failed to get event VM mobref: %s", ie.Reference())
 			return
 		}
-		if !isManagedbyVCH(*moref, sess) {
+		if !isManagedbyVCH(sess, *moref) {
 			return
 		}
 		log.Debugf("Register container VM %s", moref)
@@ -202,11 +206,11 @@ func registeredVMCallback(ie events.Event, sess *session.Session) {
 	return
 }
 
-func isManagedbyVCH(moref types.ManagedObjectReference, sess *session.Session) bool {
+func isManagedbyVCH(sess *session.Session, moref types.ManagedObjectReference) bool {
 	var vm mo.VirtualMachine
 
 	// current attributes we care about
-	attrib := []string{"config"}
+	attrib := []string{"resourcePool"}
 
 	// populate the vm properties
 	ctx := context.Background()
--- a/lib/portlayer/exec/exec.go
+++ b/lib/portlayer/exec/exec.go
@@ -34,14 +34,13 @@ import (
 )
 
 var (
-	initializer  sync.Once
-	eventSession *session.Session
+	initializer sync.Once
+	initError   error
 )
 
 func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSource, _ extraconfig.DataSink) error {
 	var err error
 	initializer.Do(func() {
-		eventSession = sess
 		f := find.NewFinder(sess.Vim25(), false)
 
 		extraconfig.Decode(source, &Config)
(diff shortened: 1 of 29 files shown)


emlin:
@jzt

lib/portlayer/exec/exec.go
```diff
@@ -152,8 +158,36 @@ func eventCallback(ie events.Event) {
...
+		attrib := []string{"config", "runtime.powerState", "summary", "resourcePool"}
 
+		// populate the vm properties
+		ctx := context.Background()
+		if err := eventSession.RetrieveOne(ctx, *moref, attrib, &vm); err != nil {
+			log.Errorf("Failed to query registered vm object %s: %s", ie.Reference(), err)
+			return
+		}
+		if *vm.ResourcePool != Config.ResourcePool.Reference() {
+			log.Debugf("vm %q is not containers of this VCH, ignore it", vm.Config.Name)
```
jzt: Minor: I would reword this error to say "`container vm %q does not belong to this VCH, ignoring`" or something.

jzt:
LGTM

lib/portlayer/exec/exec.go
emlin: nice, thanks

lib/portlayer/exec/exec.go
```diff
@@ -152,8 +158,36 @@ func eventCallback(ie events.Event) {
...
+		}
+		registeredContainers := convertInfraContainers(ctx, eventSession, []mo.VirtualMachine{vm})
+		for i := range registeredContainers {
+			Containers.put(registeredContainers[i])
+			log.Debugf("Container %q is registered back", vm.Config.Name)
```
emlin: yes, sure

lib/portlayer/exec/exec.go
```diff
@@ -152,8 +158,36 @@ func eventCallback(ie events.Event) {
...
 			}
 		}
+		return
 	}
+	// check container registered event if this container is not found in container cache
+	switch ie.String() {
+	case events.ContainerRegistered:
+		moref := new(types.ManagedObjectReference)
+		if ok := moref.FromString(ie.Reference()); !ok {
+			log.Errorf("Failed to get event VM mob reference")
```
vburenin: I would log reference here.

pkg/vsphere/vm/vm.go
```diff
@@ -310,17 +310,91 @@ func (vm *VirtualMachine) DeleteExceptDisks(ctx context.Context) (*object.Task,
...
+
+	if err = vm.Properties(ctx, vm.Reference(), []string{"config.files"}, &mvm); err != nil {
+		log.Errorf("Unable to get vm config.files property: %s", err)
+		return "", err
+	}
+
+	return mvm.Config.Files.VmPathName, nil
+}
+
+func (vm *VirtualMachine) registerVM(ctx context.Context, path string, name string, vapp *types.ManagedObjectReference,
```
vburenin: So many parameters for one function. Is it really necessary? You could also make definition much shorter combining types together.
```
registerVM(ctx context.Context, path, name string, vapp, pool, host *typesManagedObjectReference, folder *object.Folder)
```

btw, Why folder? Why not path?

pkg/vsphere/vm/vm.go
```diff
@@ -310,17 +310,91 @@ func (vm *VirtualMachine) DeleteExceptDisks(ctx context.Context) (*object.Task,
...
+// FixInvalidState fix vm invalid state issue through unregister & register
+func (vm *VirtualMachine) FixInvalidState(ctx context.Context) error {
+	log.Debugf("Fix invalid state VM: %s", vm.Reference())
+	folders, err := vm.Session.Datacenter.Folders(ctx)
+	if err != nil {
+		log.Errorf("Unable to get vm folder: %s", err)
+		return err
+	}
+
+	log.Debugf("Get vm properties")
```
vburenin: I would log which properties I am trying to retrieve. Taking into account you already have an array of the already, it should be trivial.

pkg/vsphere/vm/vm.go
emlin: yeah, it's necessary. with combined types, it will be better

pkg/vsphere/vm/vm.go
emlin: folder is actually vmfolder, will rename it

pkg/vsphere/vm/vm.go
emlin: folder is actually vmfolder, will rename it

pkg/vsphere/vm/vm.go
emlin: sure

lib/portlayer/exec/exec.go
emlin: good point

pkg/vsphere/tasks/waiter.go
```diff
@@ -89,6 +102,56 @@ func WaitForResult(ctx context.Context, f func(context.Context) (Task, error)) (
...
+	if !ok {
+		return false
+	}
+	switch f.Fault().(type) {
+	case *types.InvalidState:
+	default:
+		log.Debugf("Do not fix non invalid state error")
+		return false
+	}
+	if target == nil {
```
hmahmood: You could check this first thing in this function.

pkg/vsphere/tasks/waiter.go
```diff
@@ -89,6 +102,56 @@ func WaitForResult(ctx context.Context, f func(context.Context) (Task, error)) (
 	}
 }
 
+func fixTask(ctx context.Context, sess *session.Session, target object.Reference) error {
+	switch vmm := target.(type) {
+	case *object.VirtualMachine:
+		vm := vm.NewVirtualMachine(ctx, sess, target.Reference())
```
hmahmood: You should get rid of the `sess` parameter to `NewVirtualMachine`. Its leaking badly into this API, and the `vm.VirtualMachine` class really uses it in only two places -- both of which can be replaced with a call to [`Client()`](https://github.com/vmware/vic/blob/master/vendor/github.com/vmware/govmomi/object/common.go#L62).

pkg/vsphere/tasks/waiter.go
```diff
@@ -41,21 +46,21 @@ type Task interface {
...
 // WaitForResult wraps govmomi operations and wait the operation to complete.
 // Return the operation result
 // Sample usage:
-//    info, err := WaitForResult(ctx, func(ctx) (*TaskInfo, error) {
-//       return vm.Reconfigure(ctx, config)
+//    info, err := WaitForResult(ctx, sess *session.Session, *object.Reference, func(ctx) (*TaskInfo, error) {
+//       return vm, vm.Reconfigure(ctx, config)
 //    })
-func WaitForResult(ctx context.Context, f func(context.Context) (Task, error)) (*types.TaskInfo, error) {
+func WaitForResult(ctx context.Context, sess *session.Session, target object.Reference, f func(context.Context) (Task, error)) (*types.TaskInfo, error) {
```
hmahmood: `TaskInfo.Entity` property holds the "target" of the task, so you can get rid of that parameter as well. See https://www.vmware.com/support/developer/vc-sdk/visdk41pubs/ApiReference/vim.TaskInfo.html and https://sourcegraph.com/github.com/vmware/vic@master/-/blob/vendor/github.com/vmware/govmomi/vim25/types/types.go#L39888

pkg/vsphere/tasks/waiter.go
```diff
@@ -70,10 +75,18 @@ func WaitForResult(ctx context.Context, f func(context.Context) (Task, error)) (
 		}
 
 		log.Errorf("task failed: %s", err)
-		if !isTaskInProgress(err) {
+		if !needsRetry(target, err) {
```
hmahmood: `needsRetry` calls `needsFix`. Maybe not have `needsRetry` call `needsFix` and instead call them separately here, caching the results for later reuse?

pkg/vsphere/vm/vm.go
```diff
@@ -310,17 +310,91 @@ func (vm *VirtualMachine) DeleteExceptDisks(ctx context.Context) (*object.Task,
...
+	}
+
+	return mvm.Config.Files.VmPathName, nil
+}
+
+func (vm *VirtualMachine) registerVM(ctx context.Context, path, name string,
+	vapp, pool, host *types.ManagedObjectReference, vmfolder *object.Folder) (*object.Task, error) {
+	log.Debugf("Register VM %s", name)
+
+	c := vm.Client.Client
```
hmahmood: `vm.Client()`?

pkg/vsphere/vm/vm.go
```diff
@@ -310,17 +310,91 @@ func (vm *VirtualMachine) DeleteExceptDisks(ctx context.Context) (*object.Task,
...
 	if err != nil {
+		return nil, err
+	}
+
+	return object.NewTask(vm.Client.Client, res.Returnval), nil
+}
+
+// FixInvalidState fix vm invalid state issue through unregister & register
+func (vm *VirtualMachine) FixInvalidState(ctx context.Context) error {
+	log.Debugf("Fix invalid state VM: %s", vm.Reference())
```
hmahmood: A sanity check to see if the VM is indeed in an invalid state before attempting to unregister would be nice.

pkg/vsphere/tasks/waiter.go
```diff
@@ -70,10 +75,18 @@ func WaitForResult(ctx context.Context, f func(context.Context) (Task, error)) (
...
 			return info, err
 		}
 
+		if needsFix(target, err) {
+			if nerr := fixTask(ctx, sess, target); nerr != nil {
+				log.Errorf("Failed to fix task failure: %s", nerr)
```
hmahmood: So no retry if we get a "task in progress" error?

pkg/vsphere/vm/vm.go
```diff
@@ -310,17 +310,91 @@ func (vm *VirtualMachine) DeleteExceptDisks(ctx context.Context) (*object.Task,
...
+		return err
+	}
+	// re-register vm will change vm reference, so reset the object reference here
+	if info.Error != nil {
+		return errors.New(info.Error.LocalizedMessage)
+	}
+
+	newRef := info.Result.(types.ManagedObjectReference)
+
+	vm.Common = object.NewCommon(vm.Client.Client, newRef)
```
hmahmood: Not `object.NewVirtualMachine`? There is an assumption here that `object.NewVirtualMachine` just creates `Common`, and if that changes in the future, we break here.

lib/portlayer/exec/exec.go
```diff
@@ -152,8 +158,36 @@ func eventCallback(ie events.Event) {
 
 			}
 		}
+		return
 	}
+	// check container registered event if this container is not found in container cache
+	switch ie.String() {
```
hmahmood: Can you explain why this is required to fix the invalid state issue?

pkg/vsphere/tasks/waiter.go
emlin: this method is not retry, it's only fix the object. task will retry in existing way

pkg/vsphere/tasks/waiter.go
emlin: I feel it's more clear to have one single retry detection method. And the needsFix detection is easy to do.

pkg/vsphere/tasks/waiter.go
emlin: I agree with it, but kind of, we're using session widely in this way, for all packages talk to govmomi.
This is a reasonable claim drive us to refactor session I think, but I would prefer to do it separately.

pkg/vsphere/tasks/waiter.go
emlin: Initially I'm doing that way, but I think invalid state will be a much more rare condition than any others. In fact, we'll never have a nil target, unless in test case.

pkg/vsphere/vm/vm.go
emlin: vm.Client() is not a valid method. but I can call vm.Vim25()

pkg/vsphere/vm/vm.go
emlin: Here, I'm trying not to impact the virtual machine object itself, and reset the vm reference.
From caller point of view, it might cache the object, and continue to use that after this waitforresult method.
But inside of waitforresult method, we actually changed the VM itself from VC, so the idea to change object as less as possible. Ideally I want to set object reference only back to the object, but no such method.
NewVirtualMachine is to create a common object only, but it might cache other values after some operation, we don't want to loss those caches just inside of one method call blindly.

pkg/vsphere/tasks/waiter.go
hmahmood: You are calling `needsFix` again after this; I thought we could get rid of the extra call this way.

pkg/vsphere/tasks/waiter.go
hmahmood: I think its better if we do it now, since changes to WaitForResult result in a lot of churn.

lib/portlayer/exec/exec.go
```diff
@@ -152,8 +158,36 @@ func eventCallback(ie events.Event) {
...
+	case events.ContainerRegistered:
+		moref := new(types.ManagedObjectReference)
+		if ok := moref.FromString(ie.Reference()); !ok {
+			log.Errorf("Failed to get event VM mobref: %s", ie.Reference())
+			return
+		}
+		var vm mo.VirtualMachine
+
+		// current attributes we care about
+		attrib := []string{"config", "runtime.powerState", "summary", "resourcePool"}
```
hmahmood: Why are we retrieving config, runtime.powerState and summary? Seems like only resourcePool is used below.

lib/portlayer/exec/exec.go
```diff
@@ -152,8 +158,36 @@ func eventCallback(ie events.Event) {
...
 
+		// populate the vm properties
+		ctx := context.Background()
+		if err := eventSession.RetrieveOne(ctx, *moref, attrib, &vm); err != nil {
+			log.Errorf("Failed to query registered vm object %s: %s", ie.Reference(), err)
+			return
+		}
+		if *vm.ResourcePool != Config.ResourcePool.Reference() {
```
hmahmood: This seems to be very specific to VM objects. Maybe we should move this "fix up" to the VirtualMachine class (via a WaitForResult or similar method), and not do it here? Adding these two parameters really makes this function very confusing, and ends up with WaitForResult doing to much imho.

pkg/vsphere/vm/vm.go
hmahmood: How about we just return a new VirtualMachine object? The old one would be defunct after this call.

hmahmood: What do you think about just returning a new VirtualMachine object? Coupled with a WaitForResult method/func in this file, that could work. The existing VirtualMachine object would be defunct after this call.

pkg/vsphere/tasks/waiter.go
emlin: The change itself turns to be much more confusing than everybody thought. But the result is that we have to do it in tasks.WaitForResult, cause every VM operation might end up with this failure, and the solution is not only to fix the error, but also retry task.
If we move the fix out of WaitForResult, we'll need to modify every VM operation call, patch with this workaround, that's not what we should do.

pkg/vsphere/tasks/waiter.go
emlin: Either way works. If you prefer, I can change it.

pkg/vsphere/tasks/waiter.go
hmahmood: I meant to say add it to the VirtualMachine class or a helper function in vm.go, that way we use that for VM objects only.

pkg/vsphere/tasks/waiter.go
emlin: We still rely on session to get VMFolder, if it's ESXi, or we do not use resource pool. Remove Session cannot simply make it easy.

lib/portlayer/exec/exec.go
emlin: I knew that. But I need one more attribute for the resource pool. If call populateVMAttributes, I'll need to call property collector twice.

lib/portlayer/exec/exec.go
emlin: which kind of helper? I assume that will include the VM property query, right?

andrewtchin:
lgtm

lib/portlayer/exec/exec.go
```diff
@@ -24,18 +24,24 @@ import (
...
 )
 
-var initializer sync.Once
+var (
+	initializer  sync.Once
+	eventSession *session.Session
+)
 
 func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSource, _ extraconfig.DataSink) error {
 	var err error
```
hmahmood: Unrelated... but the error we are returning is going to be always nil after the first called to Init().

pkg/vsphere/vm/vm.go
```diff
@@ -427,3 +425,110 @@ func (vm *VirtualMachine) UpgradeInProgress(ctx context.Context, upgradePrefix s
...
+		log.Errorf("Unable to register VM %q back: %s", name, err)
+		return err
+	}
+	info, err := task.WaitForResult(ctx, nil)
+	if err != nil {
+		return err
+	}
+	// re-register vm will change vm reference, so reset the object reference here
+	if info.Error != nil {
+		return errors.New(info.Error.LocalizedMessage)
```
hmahmood: We probably should just return the `error` as is from the `WaitForResult` call.

hmahmood:
Minor comments ... lgtm

pkg/vsphere/vm/vm.go
emlin: In this case, err is nil, but info.Error is not. I doubt this will happen, but if it does, this error might be better.

lib/portlayer/exec/exec.go
```diff
@@ -111,8 +114,10 @@ func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSou
 			return
 		}
 	})
-
-	return err
+	if err != nil {
```
hmahmood: This needs to be inside the `func()` above.

emlin: okey

lib/portlayer/exec/exec.go
```diff
@@ -187,15 +202,15 @@ func eventCallback(ie events.Event) {
 	return
 }
 
-func isManagedbyVCH(moref types.ManagedObjectReference) bool {
+func isManagedbyVCH(moref types.ManagedObjectReference, sess *session.Session) bool {
```
emlin: fine

emlin:
♻️

PR #2907, opened 2016-10-26
Limit VCH nic numbers

### LINKED ISSUE #1674: Failed to find slot for NIC when more than 3 interfaces are added to VCH
I ran into this when I created a VCH that had 4 NICs. We use slot numbers to lookup up interfaces in the tether code. The slot numbers are assigned by vic-machine. However, there is no guarantee that vsphere will honor these slot numbers. This is what happened here; vsphere assigned a different slot number to the one we set in guestinfo.

We need to stop using slot numbers to identify interfaces, and move to using device labels instead. Device labels are named Ethernet0, Ethernet1, and so on, in the order NICs are added to the VM's config.

emlin:
Fixes #1674 
We cannot support four nics right now, so limit the nic number in vic-machine create


andrewtchin:
lgtm

emlin:
♻️

PR #2931, opened 2016-10-27
Add support for encrypted secrets in extraconfig

### LINKED ISSUE #2630: Hide secrets from read-only API
We need to hide appliance secrets from the read-only API, while still having them be persistent through vic-machine reconfigure operations.

The problem here is that if the API cannot see them then it cannot re-hydrate those fields in the config before doing a VM reconfigure operation. We're looking at using guestinfo.ovfEnf for hiding from the API, however we must confirm what happens if you update the Extraconfig of a VM that has a guestinfo.ovfEnv specified via API reconfigure, but without again specifying that data - if it's overwritten hen we need a more involved approach.

The alternative to embedding all secrets in one key is that we use the same model drone does and embed a private key in guestinfo.ovfEnv and use that to encrypt any secrets that are stored in a normal fashion.

dougm:
Update vic-machine to encrypt config fields tagged "secret".
Still some fields to cleanup, but other components should not require
additional changes to store and fetch VCH config secrets.

Fixes #2630


dougm:
Example output from `govc -json VCH`:

```json
{
  "Key": "guestinfo.vice./connect/userpw@secret",
  "Value": "lTgywoDQr1Pq2bB+KhNt+7bkmPrYhtYRaxvkbnkEIrJK/5JDz1emaSHtr0hUlNgGXIkeL7mmylUZjc5wfSM/"
}
```

caglar10ur:
L(very)GTM

hmahmood:
lgtm

emlin:
lgtm.
Just wondering if it's worth to keep the secretKey from caller. If just doing like the MapSink, it should be enough.

mdubya66:
ping

mdubya66:
lgtm

Issue #2929, opened 2016-10-27
Postgres 9.4 does not run on VIC


sflxn:
I am pretty sure it used to run.  It is part of the compose voting app.  It refuses to run now.

docker run -d postgres:9.4

Docker logs don't show postgres outputting any errors.  It just quits half way through starting up.

VIC pre 0.7.0.

sflxn:
[container-logs (1).zip](https://github.com/vmware/vic/files/557141/container-logs.1.zip)

caglar10ur:
check https://github.com/vmware/vic/commit/7633e162a7242f19b86c69731fe85ea7b8e60611

sflxn:
This turned out to be a postgres specific issue.  @hickeng tracked this down to their init scripts. Their dockerfile creates an anonymous volume and maps it to /var/lib/postgres/data.  Since the anonymous volume is a disk, it has a lost+found folder.  The init script in the postgres container won't run if the data folder it's started with has an existing folder (e.g. lost+found) in it.

There is a work around.  Postgres looks for the data folder in an environment variable PGDATA.  As long as PGDATA points to a folder that does not container any other folder, it will startup fine.  By default PGDATA is set to /var/lib/postgresql/data.  Override that via

docker run -d -e PGDATA=/var/lib/postgresql/data/data postgres:9.4

The folder override specifified does not need to exist.

caglar10ur:
or we could delete it?

sflxn:
We could delete it, but this really is a postgres specific issue.  I suspect a volume driver that mounts a disk onto regular docker will also have the same issue with postgres.  Since regular docker will unlikely try to delete that folder, we probably should just keep it as is.  I know EMC has a volume plugin.  It would be curious to know if they've run into this same issue with regular docker.

kacole2:
@sflxn taken from one of the engineers. "that's why we create a `/data` directory on the volume so the database software see a completely clean directory and not an artifact of ext4"

You're going to run into this issue with any Dockerfile that has `VOLUME` listed in it. Give Redis a try and see what happens. https://github.com/dockerfile/redis/blob/master/Dockerfile#L28

PR #2933, opened 2016-10-28
Fixed anonymous vol issue for postgres in voting app

sflxn:
Our anonymous volume support exposed an issue with postgres and
broke our voting app.  When the modified voting app was created,
it was tested without anonymous volume support.  Our robot
scripts did not have sufficient checks to find this issue.

There is now a workaround in the compose file for postgres and
a check to make sure every container is up and running and not
just the webservers.

Related to #2929.

### DIFF
--- a/demos/compose/voting-app/docker-compose.yml
+++ b/demos/compose/voting-app/docker-compose.yml
@@ -17,9 +17,16 @@ services:
     container_name: worker
     image: victest/vote-worker
 
+# Postgres container will create an anonymous volume.  VIC will use a
+# vmdk for the volume.  Since it is a new volume, it will have a lost+found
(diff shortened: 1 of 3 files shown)


caglar10ur:
LGTM

emlin:
lgtm

demos/compose/voting-app/docker-compose.yml
```diff
@@ -20,6 +20,8 @@ services:
   db:
     container_name: db
     image: postgres:9.4
+    environment:
```
hickeng: can we have a comment as to why this workaround is needed - otherwise it'll become one of those legacy things that no-one is sure whether it's still needed or not.

sflxn:
ping

PR #2942, opened 2016-10-28
Disable two more tests that rely on closure of #2669 [skip ci]

jzt:
This issue is causing CI builds to fail erroneously, so disabling these two additional tests as well.


andrewtchin:
lgtm

matthewavery:
lgtm

mdubya66:
In the future let's make sure we get @mhagen-vmware's approval before disabling any tests

PR #2944, opened 2016-10-28
Improve the safety of datastore cleanup for concurrent builds

mhagen-vmware:
The previous method was outdated and unsafe.  This should be more safe, but also will leak a bit more.


emlin:
lgtm 👍

anchal-agrawal:
lgtm

mhagen-vmware:
.

PR #2946, opened 2016-10-28
First pass at cpu/mem configuration support [full ci]

### LINKED ISSUE #1517: Support -c and -m parameters for containerVMs
Right now they are hardcoded as 

```
        NumCPUs:  2,
        MemoryMB: 2048,
```

https://github.com/vmware/vic/blob/master/lib/portlayer/exec/handle.go#L209

jzt:
This change allows users to supply custom CPU/Memory configuration when creating or running containers.

Usage:  `docker $DOCKER_OPTS create -m <memory size> --cpuset-cpus=<num CPUS> image`

Example:

`docker $DOCKER_OPTS create -m 8g --cpuset-cpus 3 busybox`

`-m`:  Memory limit (format: number[unit]). Number is a positive integer. Unit can be one of b, k, m, or g. Minimum is 64M (the minimum for a VM to boot is 32M).
`--cpuset-cpus`: Number of virtual CPUs to allocate to the container VM.

Note: This is not what Docker intended `--cpuset-cpus` to control. We use it in place of the preferable `-c` because Docker now warns users that `-c` is deprecated, which can lead to confusion. We may change this option in the future to allow for input format as `"x,y"`, where x = number of cores per CPU, and y = number of CPUs. I decided to keep it simple at first, so currently a single number can be passed to this option. Changing this behavior is trivial.

I need some feedback on some of the constraints we use. Right now, the memory min/max is 64M/inf., with a default value of 4GB being used if the user does not supply a value. The CPU min is 1, with a default of 2. ESX will return an error through the portlayer if a user tries to create a containerVM with more vCPUs than logical CPUs in the ESX.

This is a first pass, so I'm tagging @hickeng, @caglar10ur for feedback on this implementation approach.

Fixes #1517


lib/apiservers/engine/backends/container.go
```diff
@@ -58,6 +58,19 @@ import (
 
 const (
 	bridgeIfaceName = "bridge"
+
```
caglar10ur: 64 MB could be too small. Have you tried running something with it?

lib/apiservers/engine/backends/container.go
```diff
@@ -58,6 +58,19 @@ import (
...
+	// MemoryMinMiB - the minimum allowable container memory size
+	MemoryMinMiB = 64
+	// MemoryMaxMiB - the maximum allowable container memory size
+	MemoryMaxMiB = 16384
```
caglar10ur: Can't we calculate Max values (for cpu/mem) on runtime - based on the host that we are on?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -709,6 +709,9 @@ func (c *ContainerProxy) AttachStreams(ctx context.Context, vc *viccontainer.Vic
 func dockerContainerCreateParamsToPortlayer(cc types.ContainerCreateConfig, layerID string, imageStore string) *containers.CreateParams {
 	config := &models.ContainerCreateConfig{}
 
+	config.NumCpus = &cc.HostConfig.CPUCount
+	config.Memory = &cc.HostConfig.Memory
```
caglar10ur: Suggest using MemoryMB, as that's what spec uses (https://www.vmware.com/support/developer/vc-sdk/visdk25pubs/ReferenceGuide/vim.vm.ConfigSpec.html)

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -118,6 +118,10 @@ func (handler *ContainersHandlersImpl) CreateHandler(params containers.CreatePar
 		Key:      pem.EncodeToMemory(&privateKeyBlock),
 		LayerID:  *params.CreateConfig.Image,
 		RepoName: *params.CreateConfig.RepoName,
+		Resources: executor.Resources{
+			NumCPUs: *params.CreateConfig.NumCpus,
+			Memory:  *params.CreateConfig.Memory,
```
caglar10ur: Memory -> MemoryMB

lib/apiservers/portlayer/swagger.json
```diff
@@ -2383,6 +2383,14 @@
 				"repoName": {
 					"type": "string"
 				},
+                                "numCPUs": {
+                                  "type": "integer",
+                                  "format": "int64"
+                                },
+                                "memory": {
```
caglar10ur: Memory -> MemoryMB

lib/config/executor/container_vm.go
```diff
@@ -211,3 +214,9 @@ type Detail struct {
 	StartTime  int64 `vic:"0.1" scope:"read-write" key:"starttime"`
 	StopTime   int64 `vic:"0.1" scope:"read-write" key:"stoptime"`
 }
+
+// Resources specifies CPUcount and Memory of a container VM
+type Resources struct {
+	NumCPUs int64
+	Memory  int64
```
caglar10ur: Memory -> MemoryMB

lib/apiservers/engine/backends/container.go
jzt: No, not really, it was just meant to be a starting point. However, when I tried with 64MB to run a busybox `/bin/top` container, it wasn't enough. I tried 256MB and it kernel panicked. I'm messing around now trying to figure out a minimal value that makes sense. I'll report back soon. In the meantime, given how ESX handles memory overcommitment, it seems we have a good amount of wiggle room, but I'm not really sure what should be considered "too much"...

lib/apiservers/engine/backends/container_proxy.go
jzt: Docker does everything in KiB/MiB/GiB but I guess it doesn't matter since (apparently) VC only cares if the memory you allocate to a VM is a multiple of 4. 👍

jzt: Update: now that I've changed it, I've noticed that VC doesn't really recognize things in MB/GB etc. If I create a VM with 1000MB, VC shows it as 1000MB of memory ... however, if I create one with 1024MB, VC shows it as 1GB of memory. Additionally, if I create a VM with `1GB` of memory, the VMX shows `memSize` to be `1024`.

lib/apiservers/engine/backends/container.go
jzt: That's a great idea, though I think we wanted to keep this initial stab pretty simple. Could we attack that question at a future date? On that subject, I wonder if we can make the memory max simple by just overcommitting on everything and letting ESX handle it. There's some interesting literature on the subject [here](https://labs.vmware.com/vmtj/memory-overcommitment-in-the-esx-server).

lib/apiservers/engine/backends/container.go
jzt: I set the min to 1GiB for now.. although I think that still may be too little. @hickeng @caglar10ur what did you guys use for sensible minimums/defaults in Bonneville?

lib/apiservers/engine/backends/container.go
```diff
@@ -1349,6 +1362,38 @@ func setPathFromImageConfig(config, imageConfig *containertypes.Config) {
 func validateCreateConfig(config *types.ContainerCreateConfig) error {
 	defer trace.End(trace.Begin("Container.validateCreateConfig"))
 
+	// process cpucount here
+	var cpuCount int64 = DefaultCPUs
```
vburenin: Just saying: another way of defining variable of a specific type is cpuCount := int64(DefaultCPUs)

lib/apiservers/engine/backends/container.go
```diff
@@ -58,6 +58,19 @@ import (
 
 const (
 	bridgeIfaceName = "bridge"
+
+	// MemoryMinMiB - the minimum allowable container memory size
+	MemoryMinMiB = 1024
+	// MemoryMaxMiB - the maximum allowable container memory size
+	MemoryMaxMiB = 16384
+	// MemoryDefaultMiB - the default container VM memory size
+	MemoryDefaultMiB = 4096
```
vburenin: I would consider giving a user a chance to define these defaults via vic-machine during VCH deployment.

lib/portlayer/exec/handle.go
```diff
@@ -245,8 +245,8 @@ func Create(ctx context.Context, sess *session.Session, config *ContainerCreateC
 
 	specconfig := &spec.VirtualMachineConfigSpecConfig{
 		// FIXME: hardcoded values
```
vburenin: I suspect this comment is no longer relevant.

lib/apiservers/engine/backends/container.go
jzt: This makes a lot of sense. I'm not sure if it fits within the current scope, though. @mdubya66?

lib/apiservers/engine/backends/container.go
hickeng: In Bville there was an alignment to 128MB - I think this was specifically for hotadd given the vmfork usage, but it's a reasonable constraint for us atm anyway given the bank size. If it's not functional with 512MB as a min then we really, really need to look into why, we should be able to go smaller.

Checks from bville:
```
	// validity checks
	if remainder := module.memoryMB % 128; remainder != 0 {
		log.Warnf("Default container VM memory must be 128 aligned for hotadd, rounding up.")
		module.memoryMB += 128 - remainder
	}

	// if we fail here it causes crashing in docker - it'd be nice to push this back to the installer
	// but we currently have no mechanism to do so
	if module.memoryMB < 384 {
		log.Warnf("Minimum memory for container template is 384MB, adjusting up.")
		module.memoryMB = 384
	}
```

lib/apiservers/engine/backends/container_proxy.go
hickeng: @jzt It was only standardized in 1998 and there's a lot of legacy usage (plus ingrained habits). MB used to mean 1024 but now means 1000 and MiB means 1024 (https://en.wikipedia.org/wiki/Mebibyte).

It appears that vSphere is using the original interpretation of MB/GB.

lib/apiservers/engine/backends/container.go
hickeng: the Container.ContainerVMSize field in lib/config has been around since inception for this, but yes, it's a scope question. That said, it may be faster to provide a mechanism to set defaults than to determine properly sane defaults that cannot easily be overridden :)

lib/apiservers/engine/backends/container.go
```diff
@@ -58,6 +58,19 @@ import (
 
 const (
 	bridgeIfaceName = "bridge"
+
+	// MemoryMinMiB - the minimum allowable container memory size
+	MemoryMinMiB = 1024
```
hickeng: The minimum should be associated with the bootstrap image that's in use - just a note for the future - but regardless is not something I'd considered user configurable.

lib/apiservers/engine/backends/container.go
```diff
@@ -1349,6 +1362,38 @@ func setPathFromImageConfig(config, imageConfig *containertypes.Config) {
...
 	defer trace.End(trace.Begin("Container.validateCreateConfig"))
 
+	// process cpucount here
+	var cpuCount int64 = DefaultCPUs
+	if config.HostConfig.CpusetCpus != "" {
+		cpus := strings.Split(config.HostConfig.CpusetCpus, ",")
+		if c, err := strconv.Atoi(cpus[0]); err == nil {
+			cpuCount = int64(c)
+		} else {
+			log.Warnf("Error parsing CPU count, using default of %d: %s", DefaultCPUs, err)
```
hickeng: This should prevent container creation and return a remediable error to the user.
Generally a container configuration is pass/fail; while we supply defaults we should not be attempting to provide a sort-of-like-what-you-asked-for if there were specifics that couldn't be fulfilled.

The only places where that isn't true that I can think of are floor/ceiling constraints where we'd be requiring the user to parrot system internal limitations. There a warning should be provided to the user, but we can adjust and continue.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
hickeng: @jzt I think @caglar10ur is requesting units be part of the field name rather than we use MB vs MiB.

hickeng: @jzt these go into the ConfigSpecConfig, not the executor config as they are an inherent part of the VM configuration: https://github.com/vmware/vic/blob/HEAD/lib/spec/spec.go#L57
Adding this into the executor would duplicate data and would not reflect out-of-band changes.

lib/config/executor/container_vm.go
```diff
@@ -141,6 +141,9 @@ type ExecutorConfig struct {
 	// TODO: a bit docker specific
 	RepoName string `vic:"0.1" scope:"read-only" key:"repo"`
 
+	// Resources specifies the CPU count and memory in a container VM
+	Resources Resources
```
hickeng: This should not be here. https://github.com/vmware/vic/blob/HEAD/lib/spec/spec.go#L57 instead.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
jzt: @hickeng thanks, will update both.

lib/portlayer/exec/handle.go
```diff
@@ -244,9 +251,8 @@ func Create(ctx context.Context, sess *session.Session, config *ContainerCreateC
 	}
 
 	specconfig := &spec.VirtualMachineConfigSpecConfig{
-		// FIXME: hardcoded values
-		NumCPUs:  2,
-		MemoryMB: 2048,
+		NumCPUs:  int32(config.Resources.NumCPUs),
```
caglar10ur: why we are defining this as int64 in the struct and then casting it to int32, can't we start with int32? Don't think we will overflow but just in case :)

caglar10ur:
LGTM

lib/portlayer/exec/handle.go
```diff
@@ -39,12 +39,19 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/session"
 )
 
+// Resources describes the resource allocation for the containerVM
+type Resources struct {
+	NumCPUs  int64
```
vburenin: as @caglar10ur it could be just int32.

lib/apiservers/engine/backends/container.go
```diff
@@ -1277,6 +1288,48 @@ func setPathFromImageConfig(config, imageConfig *containertypes.Config) {
...
+
+	// fix-up cpu/memory settings here
+	if cpuCount < MinCPUs {
+		config.HostConfig.CPUCount = MinCPUs
+	}
+	log.Infof("Container CPU count: %d", config.HostConfig.CPUCount)
+
+	// convert from bytes to MiB for vsphere
+	memoryMB := config.HostConfig.Memory / units.MiB
+	if memoryMB == 0 {
```
vburenin: does it mean if asked memory size is below 1MB we will set default? I would just set a minimum if not 0.

lib/portlayer/exec/handle.go
jzt: It comes in as int64 from docker. We have to cast it somewhere, I figured that was the right spot.

lib/apiservers/engine/backends/container.go
jzt: It comes in as 0 if the user didn't specify how much memory to limit the container to. Docker won't let you enter a memory amount lower than 4MB, but in our case that's too low. Anything below the minimum will become the minimum (currently 512MB), and anything unspecified will become the default (2GB).

lib/portlayer/exec/handle.go
jzt: It's a choice about when/where you want to do the cast. It comes in from Docker CLI as an int64 but govmomi requires int32, so it's basically just an arbitrary decision on where you cast the int.

anchal-agrawal:
LGTM 👍

Issue #2767, opened 2016-10-17
Context deadline exceeded on CI build


mhagen-vmware:
https://ci.vmware.run/vmware/vic/5951

```
ContainerWaitHandler(3a9c65336c8d107d94b96a78e589147ca5791a8de1e1db4fc474d642b2245077) Error: context deadline exceeded' cannot be converted to an integer: ValueError: invalid literal for int() with base 10: 'errorresponsefromdaemon:servererrorfromportlayer:containerwaithandler(3a9c65336c8d107d94b96a78e589147ca5791a8de1e1db4fc474d642b2245077)error:contextdeadlineexceeded'
```

mhagen-vmware:
[integration_logs_5951_da17da842e71a09f12c97980e158f8ec77152db3.zip](https://github.com/vmware/vic/files/534713/integration_logs_5951_da17da842e71a09f12c97980e158f8ec77152db3.zip)

matthewavery:
This appears to be some sort of issue with `ContainerWait`. From the logs the volume was added properly and the actualy container seems to have executed properly. Additionally, some of the container logs the target container appear to be missing making it hard to tell exactly what happened directly on the container.

matthewavery:
Marking as no repro, after several attempts to do so. I will be adding more logging and referencing this ticket in the PR.

matthewavery:
Closing this. pr #2949 added additional logging to help us determine what causes these kinds of errors in the future. Additionally We have not seen one like this for a little while and it is possible that it has been fixed. We should definitely reopen this ticket if we seen a context deadline exceeded from the build again, along with the updated logs that go along with that failure.

PR #2949, opened 2016-10-28
Adds Improved fault object logging on failed task retry

matthewavery:
Adding unexpected fault dumping so that we can see the entire fault with pointers being followed when a task retry operation fails. This is in response to #2767 as well as an attempt to glean more information from odd failures when making calls to vsphere.

### DIFF
--- a/pkg/vsphere/tasks/waiter.go
+++ b/pkg/vsphere/tasks/waiter.go
@@ -112,6 +112,6 @@ func isTaskInProgress(err error) bool {
 	}
 
 	//if none are found this is an unknown fault so we are dumping the structure.
-	log.Debugf("unexpected fault om task retry : %s", spew.Sdump(err))
+	log.Debugf("unexpected fault on task retry : %s", spew.Sdump(err))
 	return false
 }
--- a/pkg/vsphere/tasks/waiter.go
+++ b/pkg/vsphere/tasks/waiter.go
@@ -92,15 +92,19 @@ func WaitForResult(ctx context.Context, f func(context.Context) (Task, error)) (
 
 func isTaskInProgress(err error) bool {
 	if soap.IsSoapFault(err) {
-		if _, ok := soap.ToSoapFault(err).VimFault().(types.TaskInProgress); ok {
+		f := soap.ToSoapFault(err)
+		if _, ok := f.VimFault().(types.TaskInProgress); ok {
 			return true
 		}
+		log.Debugf("unexpected soap fault on task retry : %s", spew.Sdump(f))
 	}
 
 	if soap.IsVimFault(err) {
-		switch soap.ToVimFault(err).(type) {
+		switch f := soap.ToVimFault(err).(type) {
 		case *types.TaskInProgress:
 			return true
+		default:
+			log.Debugf("unexpected vim fault on task retry : %s", spew.Sdump(f))
 		}
 	}
 
@@ -109,9 +113,14 @@ func isTaskInProgress(err error) bool {
 		if _, ok := err.Fault().(*types.TaskInProgress); ok {
 			return true
 		}
+		log.Debugf("unexpected error on task retry : %s", spew.Sdump(err))
+	default:
+		if f, ok := err.(types.HasFault); ok {
+			log.Debugf("unexpected fault on task retry : %s", spew.Sdump(f))
+		} else {
+			log.Debugf("unexpected error on task retry : %s", spew.Sdump(err))
+		}
 	}
 
-	//if none are found this is an unknown fault so we are dumping the structure.
-	log.Debugf("unexpected fault on task retry : %s", spew.Sdump(err))
 	return false
 }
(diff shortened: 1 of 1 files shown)


emlin:
lgtm

pkg/vsphere/tasks/waiter.go
```diff
@@ -110,5 +111,7 @@ func isTaskInProgress(err error) bool {
 		}
 	}
 
+	//if none are found this is an unknown fault so we are dumping the structure.
+	log.Debugf("unexpected fault on task retry : %s", spew.Sdump(err))
```
hickeng: Suggest you omit ChangeVersion Conflict errors as well - those are expected. See lib/portlayer/exec/commit.go

mdubya66:
I'd love to see some example spew output

matthewavery:
here is an example of the original logging output(both are from the snapshot fault): 
```
time="2016-10-31T18:15:21Z" level=error msg="task failed: Invalid configuration for device '0'." 
```

and here is a "simple" example of the spew.SDump

```
time="2016-10-31T18:15:21Z" level=debug msg="unexpected error on task retry : (task.Error) Invalid configuration for device '0'.\n" 
```

I tried to find some actually unexpected fault, but in the end went with the snapshot fault that is reliable to reproduce. 

It gives us the same message before, but also with a type. Additionally, if we run into any other types besides the `Task.Error` type, `SDump` should also follow pointers and print the value of what they are pointing to as well. @mdubya66 

@dougm could you also take a quick look at this and make sure this looks sane?

mdubya66:
that's not too spewy, i'm fine with the output

matthewavery:
For larger objects it could be significant. I have also noticed that it post's `\n` in the string instead of an actual `newline` not sure if this is because it is being passed to logrus logger or not. I suspect it is.

pkg/vsphere/tasks/waiter.go
```diff
@@ -91,15 +91,19 @@ func WaitForResult(ctx context.Context, f func(context.Context) (Task, error)) (
 
 func isTaskInProgress(err error) bool {
 	if soap.IsSoapFault(err) {
-		if _, ok := soap.ToSoapFault(err).VimFault().(types.TaskInProgress); ok {
+		f := soap.ToSoapFault(err)
+		if _, ok := f.VimFault().(types.TaskInProgress); ok {
 			return true
 		}
+		log.Debugf("unexpected soap fault on task retry : %s", spew.Sdump(f))
```
dougm: What does this output look like?  I think at this point, where the fault has been unwrapped from Soap/Vim, just using '%#v' with 'f' should provide enough details here.  That's what we're doing elsewhere.  Not sure its worth introducing go-spew here, but if it is we ought to use it across the board - and maybe hidden behind a helper so it'll be simple to adjust, swap in a different pretty printer if needed, etc.

matthewavery: I introduced spew in this area since we have had problems in the past figuring out exactly what some unexpected faults are. This is more specifically for faults that are hard to reproduce, in those cases I think it would be useful to have as much information about that object as possible. I can change that section to using `%#v`, I plan to leave the others as is if you have no objections. @dougm

dougm: btw you should log `f.VimFault()` here, rather than `f` and `%#v` should be just fine.  If go-spew is helpful elsewhere, fine, just not seeing how it is with the provided example output.

emlin:
Would you mind to add unit test to cover some specific errors returned? I assume you can write your own task function, which can return any error you want.
Specifically, I'd like to see if nil pointer error can be printed correctly. As I just found one nil pointer issue from govmomi.

pkg/vsphere/tasks/waiter.go
dougm: Just to clarify - we do want the improved logging that comes from unwrapping the error from Soap/Vim as you've done here.  The type returned by `f.VimFault()` here, `soap.ToVimFault()` and `err.Fault()` below are at the level we want.

Versus the line you removed `log.Errorf("task failed: %s", err)` where we weren't getting any detail on the error type.

pkg/vsphere/tasks/waiter.go
```diff
@@ -108,6 +112,13 @@ func isTaskInProgress(err error) bool {
 		if _, ok := err.Fault().(*types.TaskInProgress); ok {
 			return true
 		}
+		log.Debugf("unexpected error on task retry : %s", spew.Sdump(err))
```
dougm: Should log `err.Fault()` here, so we'll get the right level of type detail.

matthewavery:
All that is left is the new test for @emlin

dougm:
lgtm

Regarding the tests, #2957 should have you covered - don't think we should be trying to catch/recover any NPE in this change @emlin , that is a separate/unrelated issue.

matthewavery:
Thanks, I will make a ticket for adding vcsim tests.

matthewavery:
♻️

PR #2950, opened 2016-10-29
Force stop if kill doesn't stop container

### LINKED ISSUE #2617: Docker kill not actually killing containers
**VIC Version:**

0.6.0 / build 5551 from BinTray

**Expected behavior:**
Running `docker kill` should stop the container passed

**Actual behavior:**
`docker ps` shows container still running and subsequent `docker rm` command fails as such.  The only workaround I find is to use `docker rm -f`

[logs.zip](https://github.com/vmware/vic/files/517933/logs.zip)

sflxn:
Use Stop() to ultimately stop a container if signaling the
process in the container does not kill it.

Also refactor containerStop() to container_proxy.go.

Resolves #2617

### DIFF
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -40,7 +40,6 @@ import (
 	"net/http"
 	"strings"
 	"sync"
-	"syscall"
 	"time"
 
 	"golang.org/x/net/context"
@@ -85,15 +84,17 @@ type VicContainerProxy interface {
 	AddVolumesToContainer(handle string, config types.ContainerCreateConfig) (string, error)
 	AddLoggingToContainer(handle string, config types.ContainerCreateConfig) (string, error)
 	AddInteractionToContainer(handle string, config types.ContainerCreateConfig) (string, error)
-	CommitContainerHandle(handle, imageID string) error
+	CommitContainerHandle(handle, containerID string, waitTime int32) error
 	StreamContainerLogs(name string, out io.Writer, started chan struct{}, showTimestamps bool, followLogs bool, since int64, tailLines int64) error
 
+	Stop(vc *viccontainer.VicContainer, name string, seconds int, unbound bool) error
 	IsRunning(vc *viccontainer.VicContainer) (bool, error)
 	Wait(vc *viccontainer.VicContainer, timeout time.Duration) (exitCode int32, processStatus string, containerState string, reterr error)
 	Signal(vc *viccontainer.VicContainer, sig uint64) error
 	Resize(vc *viccontainer.VicContainer, height, width int32) error
 	AttachStreams(ctx context.Context, vc *viccontainer.VicContainer, clStdin io.ReadCloser, clStdout, clStderr io.Writer, ca *backend.ContainerAttachConfig) error
 
+	Handle(id, name string) (string, error)
 	Client() *client.PortLayer
 }
 
@@ -121,6 +122,7 @@ const (
 	forceLogType                         = "json-file" //Use in inspect to allow docker logs to work
 	annotationKeyLabels                  = "docker.labels"
 	killWaitForExit        time.Duration = 2 * time.Second
+	killWaitBeforeForce    time.Duration = 10 * time.Second //Time to wait for signal to take effect before attempting force using Stop()
 	ShortIDLen                           = 12
 
 	DriverArgFlagKey      = "flags"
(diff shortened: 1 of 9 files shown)


lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -422,12 +463,92 @@ func (c *ContainerProxy) StreamContainerLogs(name string, out io.Writer, started
...
+		default:
+			return InternalServerError(err.Error())
+		}
+	}
+
+	handle = stateChangeResponse.Payload
+	wait := int32(seconds)
+	err = c.CommitContainerHandle(handle, vc.ContainerID, wait)
+	if err != nil {
+		if strings.Contains(err.Error(), "No such") {
```
emlin: No typed error here?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -532,10 +653,16 @@ func (c *ContainerProxy) Signal(vc *viccontainer.VicContainer, sig uint64) error
 			}
 		}
 
-		c.Wait(vc, -1*time.Second)
+		// Wait awhile to see if signal killed the container
+		if _, _, _, err := c.Wait(vc, killWaitBeforeForce); err != nil {
```
hmahmood: We are waiting above before we check the container state again; why don't we just call stop in that code? Also, is calling stop a good idea? I am not sure if sending a KILL signal to the container and stop are equivalent from the client's perspective. 

In the type switch above we are also missing a default case.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -422,12 +463,92 @@ func (c *ContainerProxy) StreamContainerLogs(name string, out io.Writer, started
...
+	if err != nil {
+		return err
+	}
+
+	// we have a container on the PL side lets check the state before proceeding
+	// ignore the error  since others will be checking below..this is an attempt to short circuit the op
+	// TODO: can be replaced with simple cache check once power events are propagated to persona
+	running, err := c.IsRunning(vc)
+	if err != nil {
+		cache.ContainerCache().DeleteContainer(vc.ContainerID)
```
sflxn: I didn't add that.  I moved that code from container.go to the proxy code and didn't change the logic.  I also questioned that code.  After your questioning, I looked into it and went ahead and change it to only do this if we get a not found from IsRunning.

lib/apiservers/engine/backends/container_proxy.go
sflxn: The distinction between stop and kill in VIC is very thin.  Stop in VIC causes a VM poweroff.  This is actually what we want if signaling fails.  Regular docker will kill the process if signaling the container fails.  Kill is kind of a definitive shutdown and not a nice shutdown.  This is why I don't think there is a need to retry.  It just so happens in VIC, Stop does the same thing.

Added a default case.

hmahmood: OK that makes sense. Do we need a wait then? Why is an error from the PL in response to the signal not enough to just go straight to the stop() call?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -422,12 +463,92 @@ func (c *ContainerProxy) StreamContainerLogs(name string, out io.Writer, started
...
+	// we have a container on the PL side lets check the state before proceeding
+	// ignore the error  since others will be checking below..this is an attempt to short circuit the op
+	// TODO: can be replaced with simple cache check once power events are propagated to persona
+	running, err := c.IsRunning(vc)
+	if err != nil && IsNotFoundError(err) {
+		cache.ContainerCache().DeleteContainer(vc.ContainerID)
+		return err
+	}
+	if !running {
+		return nil
```
emlin: looks like, if there is an error not IsNotFoundError, no err will return here.

sflxn: We don't actually want to return that error.  We only wanted to check for two pre-conditions: 1) not found, 2) not running.  For any other case, we just drop down to the rest of the code.

Btw, this code, logically is the same as before.  I moved this function from container.go to container_proxy.go to segregate the swagger code for stop out of container.go.

emlin: ok, that's fine.

sflxn:
I made another change.  I pushed the kill logic to the portlayer to optimize the operations.  It does run better.

lib/portlayer/exec/base.go
```diff
@@ -198,40 +198,39 @@ func (c *containerBase) stop(ctx context.Context, waitTime *int32) error {
...
+	sig := string(ssh.SIGKILL)
+	log.Infof("sending kill -%s %s", sig, c.ExecConfig.ID)
 
-				// Check if the poweroff task was canceled due to a concurrent guest shutdown
-				if len(terr.FaultMessage) > 0 && terr.FaultMessage[0].Key == vmNotSuspendedKey {
-					log.Infof("power off %s task skipped due to guest shutdown", c.ExecConfig.ID)
-					return nil
-				}
-				log.Warnf("generic vm config fault during power off: %#v", terr)
+	err := c.startGuestProgram(ctx, "kill", sig)
```
hmahmood: This is not correct. You are sending `kill SIGKILL`; should be sending `kill -SIGKILL`.

Why not just put the code in the for loop in shutdown() into its own function and have both places call that?

lib/portlayer/exec/base.go
```diff
@@ -276,6 +275,43 @@ func (c *containerBase) shutdown(ctx context.Context, waitTime *int32) error {
 	return fmt.Errorf("failed to shutdown %s via kill signals %s", c.ExecConfig.ID, stop)
 }
 
+func (c *containerBase) poweroff(ctx context.Context) error {
+	_, err := c.vm.WaitForResult(ctx, func(ctx context.Context) (tasks.Task, error) {
```
hmahmood: Should check `c.vm != nil`

sflxn: good catch.

sflxn:
Also unmapped ports if container has stopped running.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -121,6 +122,7 @@ const (
 	forceLogType                         = "json-file" //Use in inspect to allow docker logs to work
 	annotationKeyLabels                  = "docker.labels"
 	killWaitForExit        time.Duration = 2 * time.Second
+	killWaitBeforeForce    time.Duration = 10 * time.Second //Time to wait for signal to take effect before attempting force using Stop()
```
hmahmood: Do we need this anymore?

hmahmood:
lgtm

emlin:
lgtm

sflxn:
ping

lib/apiservers/engine/backends/container_proxy.go
sflxn: removed

Issue #2525, opened 2016-09-29
docker-compose complains if bringing up app on a VCH with a running container


sflxn:
As a VIC and docker compose user, I should be able to bring up a compose app on a VCH that already have running containers.

Acceptance Criteria:  Robot scripts that attempts to bring up a compose app on a VCH with a running container.

---

On a VCH, I have a private registry running:

```
loc@ubuntu:~/docker-tests/compose/voting-app$ docker ps -a
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                            NAMES
9b08d0ce5b18        registry:2.0        "registry cmd/registr"   6 days ago          Up 23 minutes       192.168.218.132:5000->5000/tcp   registry
```

I then try to bring up a compose app, it panics and dumps the following call stack:

```
loc@ubuntu:~/docker-tests/compose/voting-app$ docker-compose up -d
WARNING: Found orphan containers (egistry) for this project. If you removed or renamed this service in your compose file, you can run this command with the --remove-orphans flag to clean it up.
Traceback (most recent call last):
  File "<string>", line 3, in <module>
  File "compose/cli/main.py", line 61, in main
  File "compose/cli/main.py", line 113, in perform_command
  File "compose/cli/main.py", line 835, in up
  File "compose/project.py", line 379, in up
  File "compose/project.py", line 179, in get_services_without_duplicate
  File "compose/service.py", line 481, in remove_duplicate_containers
  File "compose/service.py", line 495, in duplicate_containers
  File "compose/container.py", line 87, in number
ValueError: Container 9b08d0ce5b18 does not have a com.docker.compose.container-number label
docker-compose returned -1
```

If I attempt this with regular docker, it runs fine.  I suspect it's looking for something in docker inspect and we're not returning the correct data.

sflxn:
#431

sflxn:
Upon investigating this in order to unblock myself, this may not be an issue.  It may just be a very bad limitation with docker-compose.  If I execute

```
docker-compose up --help
```

It displays the following snippet:

```
    --remove-orphans           Remove containers for services not
                               defined in the Compose file
```

It seems to indicate it does not like containers for services not defined in the compose file.  It still doesn't explain why with regular docker, it does not complain.

mdubya66:
Backlog for post dev complete

anchal-agrawal:
https://ci.vmware.run/vmware/vic/6370 and https://ci.vmware.run/vmware/vic/6390

[...]

PR #2951, opened 2016-10-29
Disable basic docker compose test [skip ci]

anchal-agrawal:
Disables the `3-03-Docker-Compose-Basic` integration test and tags it with #2525.


mdubya66:
LGTM

PR #2953, opened 2016-10-29
Add user agent header to http requests

### LINKED ISSUE #2934: "User Agent" of imagec should be recognisable.
This is an issue we saw when using **vic-product**

**VIC Version:**
0.6.0

**Description:**
Harbor filters the user-agent of the request to pull image, only request by docker will be recorded as access log.
The vanilla docker has user-agent like "docker/1.12.3 ......"

But when user runs "docker pull" via VIC engine the user agent is "Go-http-client/1.1", and this results that Harbor will not know if it's a real pull or just an API GET call.

I wish to suggest imagec should either use "docker..." or some other user-agent to identify itself.

jzt:
The user-agent for outgoing http requests will now be set to `vic/<version>`, i.e.:  `vic/0.6.0`.

Fixes #2934

### DIFF
--- a/pkg/fetcher/fetcher.go
+++ b/pkg/fetcher/fetcher.go
@@ -37,6 +37,7 @@ import (
 	"golang.org/x/net/context/ctxhttp"
 
 	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/version"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


matthewavery:
lgtm

jzt:
♻️

jzt:
lgtm :(

PR #2959, opened 2016-11-01
Do not return stale data to the caller

caglar10ur:


lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -380,7 +380,9 @@ func (handler *ContainersHandlersImpl) ContainerWaitHandler(params containers.Co
 
 	select {
 	case <-c.WaitForState(exec.StateStopped):
+		c.Refresh(context.Background())
```
hickeng: We already have a Refresh on this path, however it's got ordering issues:
see exec.go:145 and exec.go:156.

perhaps we should be moving the SetState call after the refresh (for success or failure)? The only one it's crucial for is stopped (exit code) but without knowing what the perf impact is I'm leaning towards state consistency.

hickeng:
lgtm - I'll let you decide on exec or this. We'll likely have to revisit it anyway.

mdubya66:
Lgtm

caglar10ur:
I'll keep this as is for now and merge. We can address the ordering later

PR #2964, opened 2016-11-01
Increase timeout up to 10 seconds and decrease amount of data.

### LINKED ISSUE #2963: TestHandshakeServerLotsOfTrashOnTheLine fails due to timeout.
Loaded system may require much longer time to transfer the data. 
Reducing amount of data in order of magnitude as well as increasing timeout to 10 seconds should help.

vburenin:
Increase timeout up to 10 seconds and decreased amount of data many times to avoid timeouts on a loaded system.
Fixes #2963

### DIFF
--- a/pkg/serial/handshake_test.go
+++ b/pkg/serial/handshake_test.go
@@ -129,14 +129,14 @@ func TestHandshakeServerNormalCaseScenario(t *testing.T) {
 
 func TestHandshakeServerLotsOfTrashOnTheLine(t *testing.T) {
 	log.SetLevel(log.InfoLevel)
-	clientConn, serverConn := NewFakeConnection(time.Second * 3)
+	clientConn, serverConn := NewFakeConnection(time.Second * 10)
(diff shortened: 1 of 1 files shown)


jzt:
LGTM

vburenin:
LGTM

pkg/serial/handshake_test.go
```diff
@@ -129,14 +129,14 @@ func TestHandshakeServerNormalCaseScenario(t *testing.T) {
 
 func TestHandshakeServerLotsOfTrashOnTheLine(t *testing.T) {
 	log.SetLevel(log.InfoLevel)
-	clientConn, serverConn := NewFakeConnection(time.Second * 3)
+	clientConn, serverConn := NewFakeConnection(time.Second * 10)
```
hickeng: If we're playing with test timeouts like this, I think we should be using infinite timeout and letting the Go test timeout kill it.

Issue #2537, opened 2016-10-01
Add proxy support to the appliance


hickeng:
There is currently no persistent mechanism for configuring a proxy server. While /etc/sysconfig/proxy can be modified it will not persist across reboot.

I'm unsure how much code, if any, in the appliance is necessary to support this or whether the Go http stack handles this transparently.

If we have to provide different proxies for the different networks it will get involved, if only from a command line args perspective. The raw number of options to vic-machine is steadily bumping the importance of the manifest support.

This is at the request of a customer via @cormachogan

mdubya66:
We will need to address this post GA. @karthik-narayan

hickeng:
To make changes to the applianceVM and a recent build of VIC then you can run the following to enable console login until reboot with the password you specify:
```vic-machine-<os> debug --target=... --rootpw=somepassword``` 

If you also add `--enable-ssh` argument then it will start sshd until next reboot. If you'd prefer to automate, you can combine it with `--authorized-key=<file>` to set the provided file as the authorized keys for root in the applianceVM.

This can allow for editing of system configuration files, non-persistently, that are not yet exposed for configuration via vic-machine.

mldmld68:
I have the  VCH version: v0.6.0-4890-4f98611, AFAIK it's the last one
# ./vic-machine-linux --version
vic-machine-linux version v0.6.0-4890-4f98611
and when I enter 
#./vic-machine-linux debug -t  "https://vc-eval.pocx86.tstwinx.net/Cloud Foundry" --user myuser --password mypass --name virtual-container-host   
It answers : No help topic for 'debug'
Thanks

hickeng:
@mldmld68 if you're on tagged builds rather than from HEAD then 0.7 will introduce the debug command but it's not tagged yet. 0.6 you can use `password` if it's unchanged.

Part of the 0.7 work is shifting to a less dev friendly, more deploy friendly set of defaults (including adding tlsverify support #2197 and certificate based auth to the logs #2691)

karthik-narayan:
@stuclem Could we also document the fix for adding proxy support. It looks like we've documented the SSH access, but not the details to set the proxy.

hickeng:
@karthik-narayan there's likely code changes needed to pick up the proxy configuration, even if you do get onto the appliance and configure various settings. If proxy support is needed for 0.8 it needs to be pulled in immediately.

karthik-narayan:
[...]

PR #2970, opened 2016-11-01
Set http Proxy function in fetcher

dougm:
With this field set, env vars can be used to configure the proxy for
imagec:  https://golang.org/pkg/net/http/#ProxyFromEnvironment

Issue #2537


mdubya66:
per doug, not a full fix but might be enough of a workaround. low risk.
lgtm

caglar10ur:
LGTM :sheep: :it:

PR #2976, opened 2016-11-02
Only adding needed dhcp options [full ci]

### LINKED ISSUE #1656: DHCP client request parameters should be based config, not hard coded
Currently these are hardcoded in lib/dhcp/client. They should instead be based on what is in the config, e.g. if a nameserver list is present, then DNS servers should not be requested from the DHCP server.

hmahmood:
Fixes #1656

### DIFF
--- a/cmd/vic-init/main_linux.go
+++ b/cmd/vic-init/main_linux.go
@@ -35,7 +35,8 @@ var (
 )
 
 func init() {
-	trace.Logger.Level = log.DebugLevel
+	// use the same logger as the log files
+	trace.Logger = log.StandardLogger()
 }
 
 func main() {
@@ -56,6 +57,9 @@ func main() {
 	extraconfig.Decode(src, &config)
 
 	debugLevel = config.Diagnostics.DebugLevel
+	if debugLevel > 1 {
+		enableShell()
+	}
 	setLogLevels()
 
 	logFile, err := os.OpenFile("/dev/ttyS1", os.O_WRONLY|os.O_SYNC, 0644)
--- a/cmd/vic-init/main_linux.go
+++ b/cmd/vic-init/main_linux.go
@@ -57,7 +57,7 @@ func main() {
 	extraconfig.Decode(src, &config)
 
 	debugLevel = config.Diagnostics.DebugLevel
-	if debugLevel > 1 {
+	if debugLevel > 2 {
 		enableShell()
 	}
 	setLogLevels()
--- a/lib/dhcp/client/client.go
+++ b/lib/dhcp/client/client.go
@@ -15,9 +15,9 @@
 package client
 
 import (
+	"bytes"
 	"fmt"
 	"net"
-	"sync"
 	"syscall"
 	"time"
 
@@ -27,6 +27,7 @@ import (
 
 	"github.com/vmware/vic/lib/dhcp"
 	"github.com/vmware/vic/pkg/ip"
+	"github.com/vmware/vic/pkg/trace"
 )
 
 // Client represents a DHCP client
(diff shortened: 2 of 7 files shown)


lib/dhcp/client/client.go
```diff
@@ -244,46 +266,128 @@ func (c *client) renew(id ID, ack dhcp4.Packet, cl *dhcp4client.Client) (dhcp4.P
...
+	opts := (*p).ParseOptions()
+
+	// the current parameter request list
+	rl := opts[dhcp4.OptionParameterRequestList]
+	params := make([]byte, len(c.params))
+	copy(params, c.params)
+	if len(rl) > 0 {
+		// figure out if there are any new parameters
+		for i := 0; i < len(params); {
+			if bytes.IndexByte(rl, params[i]) != -1 {
```
vburenin: I think it can be done a little better:
```
output_params := make([]byte, len(c.params), 0)
for _, p := range params {
    if bytes.IndexByte(rl, p) < 0 {
        output_params = append(output_params, p)
    }
}
```

vburenin: oh, there is a continuation. Even better. Taking into account that number of DHCP params is very limited, timings are negligible here,
```
rl := opts[dhcp4.OptionParameterRequestList]
for _, p := range params {
    if bytes.IndexByte(rl, p) < 0 {
        dirty = true
        rl = append(rl, p)
    }
}
opts[dhcp4.OptionParameterRequestList] = rl
```

lib/dhcp/client/client.go
```diff
@@ -244,46 +266,128 @@ func (c *client) renew(id ID, ack dhcp4.Packet, cl *dhcp4client.Client) (dhcp4.P
...
+func (c *client) SetParameterRequestList(params ...byte) {
+	defer trace.End(trace.Begin(""))
+
+	c.params = make([]byte, len(params))
+	copy(c.params, params)
+	log.Debugf("c.params=%#v", c.params)
+}
+
+// setOptions sets dhcp options on a dhcp packet
+func (c *client) setOptions(p *dhcp4.Packet) (*dhcp4.Packet, error) {
```
vburenin: why p is a pointer if it is a slice?

lib/tether/ops_linux.go
```diff
@@ -267,22 +267,33 @@ func renameLink(t Netlink, link netlink.Link, slot int32, endpoint *NetworkEndpo
...
 	// use dhcp to acquire address
-	id, err := client.NewID(link.Attrs().Index, link.Attrs().HardwareAddr)
+	dc, err := client.NewClient(link.Attrs().Index, link.Attrs().HardwareAddr)
 	if err != nil {
 		return nil, err
 	}
 
-	ack, err = dc.Request(id)
+	params := []byte{
+		byte(dhcp4.OptionSubnetMask),
```
vburenin: these are untyped constants, casting them to a specific type doesn't make much sense.

vburenin: ah, they are typed. nm.

vburenin:
LGTM with some minor comments.

vburenin:
thanks for addressing my comments. LGTM one more time.

cmd/vic-init/main_linux.go
```diff
@@ -56,6 +57,9 @@ func main() {
 	extraconfig.Decode(src, &config)
 
 	debugLevel = config.Diagnostics.DebugLevel
+	if debugLevel > 1 {
```
hmahmood: Actually I only did this for my debugging, and it should not have ended up in the PR :-) But if you think it is still a good idea for debug > 2, I'll keep it with that change.

andrewtchin:
lgtm

PR #2979, opened 2016-11-02
Update coverage server url

andrewtchin:
Remove DNS and Cloudflare as variables in failing coverage uploads by using local IP
Also moved coverage upload to the last step so we would still upload to bintray


casualjim:
lgtm

mdubya66:
lgtm

andrewtchin:
♻️  lgtm

PR #2980, opened 2016-11-02
Add proxy support to vic-machine for image fetching

### LINKED ISSUE #2537: Add proxy support to the appliance

hickeng:
The fetcher component (imagec) requires support for a proxy in many
customer deployments. This adds the vic-machine glue for correctly
populating the environment variables of the personality so that fetcher
has them available.

This does NOT provide or configure proxy support for containers run by
this VCH.

Fixes #2537


vendor/github.com/urfave/cli/app.go
```diff
@@ -467,6 +467,7 @@ func HandleAction(action interface{}, context *Context) (err error) {
 			switch r.(type) {
 			case error:
 				err = r.(error)
+				panic(err)
```
andrewtchin: Did you mean to leave this?

cmd/vic-machine/create/create.go
```diff
@@ -384,6 +387,23 @@ func (c *Create) Flags() []cli.Flag {
...
+			Value:       "",
+			Usage:       "An HTTPS proxy for use when fetching images, in the form https://fqdn_or_ip:port",
+			Destination: &c.httpsProxy,
+			Hidden:      true,
+		},
+
+		cli.StringFlag{
+			Name:        "http-proxy, hproxy",
+			Value:       "",
+			Usage:       "An HTTP proxy for use when fetching images, in the for http://fqdn_or_ip:port",
```
sgairo: "in the form"

vendor/github.com/urfave/cli/app.go
hickeng: no, no I did not. Wish there was a way to toggle this with --debug... it's a pain not having any idea where panic's come from.

emlin:
lgtm 👍

cmd/vic-machine/create/create.go
hickeng: done

andrewtchin:
lgtm

hickeng:
poke

Issue #2989, opened 2016-11-02
logs --follow does not display all output


hickeng:
Many builds exhibited this:
https://ci.vmware.run/vmware/vic/6453
https://ci.vmware.run/vmware/vic/6504
https://ci.vmware.run/vmware/vic/6506
https://ci.vmware.run/vmware/vic/6502

```
Create linked containers that can ping                                | FAIL |
...
does not contain '2 packets transmitted, 2 packets received'
```

https://github.com/vmware/vic/pull/2956 addressed the truncation of logs on the datastore, but the test is still failing.
This suggests that there is an issue with logs --follow in some fashion.

Suggested first pass:
1. record bytes read from the portlayer
2. record bytes written to the personality
--> log those byte counts when shutting down the log followers.
This allows us to determine where the disconnect is:
a. missing on datastore at time of power off event
b. dropped in-flight in portlayer
c. dropped in the personality

Original robot test was:
```
Create linked containers that can ping
    ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} pull debian
    Should Be Equal As Integers  ${rc}  0

    ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} create --link busy1:busy1 --name busy2 debian ping -c2 busy1
    Should Be Equal As Integers  ${rc}  0
    Should Not Contain  ${output}  Error
    ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} start busy2
    Should Be Equal As Integers  ${rc}  0
    Should Not Contain  ${output}  Error
    ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} logs --follow busy2
    Should Be Equal As Integers  ${rc}  0
    Should Contain  ${output}  2 packets transmitted, 2 packets received
```

This should be re-introduced in some manner when this issue is addressed as it's been valuable in finding problems.

dougm:
I'll take a look, could be that #2445 didn't cover all the bases.

dougm:
Talked to @hickeng , he found an issue in the interaction layer, it uses an extra super tiny 64 byte buffer.  Explains why #2445 didn't work in this case - we'll need to allow it to drain to EOF, not just do 1 more read.

dougm:
vmware/govmomi#624

I'll include a new robot test when updating vendor.

PR #2990, opened 2016-11-02
Work around logs --follow truncation in test

hickeng:
Use start/wait/logs instead of start/logs --follow

This adds none of the data needed for diagnosis, simply a bandaid to suppress
test failure without completely disabling it.

Bandaid #2989


jzt:
LGTM

dougm:
lgtm

PR #2991, opened 2016-11-02
Not applying network config during session restart

### LINKED ISSUE #1886: Network config should not be applied when a session process exits on the VCH
Currently a "reload" is triggered when a process on the VCH crashes, which in turn reconfigures the network. This should not happen if it succeeded the first time around. Only the process launch should be attempted again.

hmahmood:
Fixes #1886

### DIFF
--- a/lib/tether/ops_linux.go
+++ b/lib/tether/ops_linux.go
@@ -425,23 +425,34 @@ func (t *BaseOperations) Apply(endpoint *NetworkEndpoint) error {
 }
 
 func apply(nl Netlink, t *BaseOperations, endpoint *NetworkEndpoint) error {
+	if endpoint.applied {
+		log.Infof("skipping applying config for network %s as it has been applied already", endpoint.Network.Name)
+		return nil // already applied
+	}
+
 	// Locate interface
 	slot, err := strconv.Atoi(endpoint.ID)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


vburenin:
LGTM

jzt:
LGTM

lib/tether/ops_linux.go
```diff
@@ -425,23 +425,34 @@ func (t *BaseOperations) Apply(endpoint *NetworkEndpoint) error {
...
 	// Locate interface
 	slot, err := strconv.Atoi(endpoint.ID)
 	if err != nil {
-		detail := fmt.Sprintf("endpoint ID must be a base10 numeric pci slot identifier: %s", err)
-		return errors.New(detail)
+		return fmt.Errorf("endpoint ID must be a base10 numeric pci slot identifier: %s", err)
 	}
-	link, err := nl.LinkBySlot(int32(slot))
+
+	defer func() {
```
vburenin: I really dislike this pattern when deferred function checks for errors and set some flags, it should be done by a higher level caller. I see benefits of this approach, I understand the reason, it just makes this code so hard to follow. I needed to unroll so much code to ensure I am not missing anything. This function has to be already split into multiple smaller chunks to make it simpler.

hmahmood: Hm... actually I think we return early in at least one case below. For now I am going to leave this alone.

PR #2977, opened 2016-11-02
Update vendor/govmomi to release 0.11.2

dougm:
Fixes #2968 (NSX-T)

Avoid possible VirtualMachine.Device NPE seen in #2818

Fix for #2878 (docker image update still required)

Issue #2958, opened 2016-10-31
Check VM State before applying invalid fix.


vburenin:
ESXi: Latest. running on Baremetal with SSD.

```
➜  vic git:(master) ✗ docker run -ti ubuntu
Unable to find image 'ubuntu:latest' locally
Pulling from library/ubuntu
6bbedd9b76a4: Pull complete
fc19d60a83f1: Pull complete
de413bb911fd: Pull complete
2879a7ad3144: Extracting [==================================================>]    678 B/678 B
668604fde02e: Download complete
a3ed95caeb02: Download complete
docker: Failed to write to image store: [POST /storage/{store_name}][500] WriteImage default  &{Code:0xc420dae528 Message:The operation is not allowed in the current state.}.
See 'docker run --help'.
```
 Next time I run it operation succeded

Portlayer log
```
time="2016-10-31T22:22:53Z" level=debug msg="ImageCache: Image 6571ba684f54b0200982f324b5b84346e4aa0a97fdd006b36d8404eab3e8f6e8 not in cache, retreiving from datastore"
time="2016-10-31T22:22:53Z" level=info msg="Saving parent map (564d7111-c455-d869-5143-318f35ddc0e4/parentMap)"
time="2016-10-31T22:22:53Z" level=info msg="Moving [datastore1] vic-docker/VIC/564d7111-c455-d869-5143-318f35ddc0e4/parentMap.tmp to [datastore1] vic-docker/VIC/564d7111-c455-d869-5143-318f35ddc0e4/parentMap"
time="2016-10-31T22:22:53Z" level=info msg="Creating directory [datastore1] vic-docker/VIC/564d7111-c455-d869-5143-318f35ddc0e4/images/6571ba684f54b0200982f324b5b84346e4aa0a97fdd006b36d8404eab3e8f6e8"
time="2016-10-31T22:22:53Z" level=info msg="Writing metadata 564d7111-c455-d869-5143-318f35ddc0e4/images/6571ba684f54b0200982f324b5b84346e4aa0a97fdd006b36d8404eab3e8f6e8/imageMetadata/metaData"
[...]

PR #2992, opened 2016-11-03
fix NPE in vic-machine caused by vm invalid state error

### LINKED ISSUE #2818: RPC error in docker stop, and eventually, vic-machine delete constantly fail with invalid memory address
Seen in https://ci.vmware.run/vmware/vic/6045 and as a follow on from #2817 

Build https://ci.vmware.run/vmware/vic/6013 left `VCH-6013-4771` and `drunk_englebart-9f1ce9...` behind.

For some reason, all of the `vic-machine delete` operations that occur against that host attempt to delete `drunk_englebart`.
1. They **must not** try to do so as that VM does **not** belong to them - that they are suggests strongly that the SearchIndex is not being used correctly or has a bug.
2. Even if they were to do so, they should not be getting a nil pointer.

```
[34mINFO[0m[2016-10-19T18:09:37Z] Connect to docker:                           
[34mINFO[0m[2016-10-19T18:09:37Z] docker -H dhcp-192-168-31-221.ci.drone.local:2376 --tlsverify --tlscacert="./VCH-6053-1693/ca.pem" --tlscert="./VCH-6053-1693/cert.pem" --tlskey="./VCH-6053-1693/key.pem" info 
[34mINFO[0m[2016-10-19T18:09:37Z] Installer completed successfully             
[34mINFO[0m[2016-10-19T18:09:48Z] ### Removing VCH ####                        
[34mINFO[0m[2016-10-19T18:09:48Z]                                              
[34mINFO[0m[2016-10-19T18:09:48Z] VCH ID: VirtualMachine:2279                  
[34mINFO[0m[2016-10-19T18:09:48Z] Removing VMs                                 
[34mINFO[0m[2016-10-19T18:09:48Z] Removing image stores                        
[34mINFO[0m[2016-10-19T18:09:54Z] Removing volume stores                       
[34mINFO[0m[2016-10-19T18:09:54Z] Deleting volume store "default" on Datastore "datastore1" at path "test" 
[34mINFO[0m[2016-10-19T18:09:54Z] Removing appliance VM network devices        
[34mINFO[0m[2016-10-19T18:09:55Z] Bridge network was not created during VCH deployment, leaving it there 
[34mINFO[0m[2016-10-19T18:09:55Z] Removing Resource Pool "VCH-6053-1693"       
[34mINFO[0m[2016-10-19T18:09:55Z] Completed successfully                       
[34mINFO[0m[2016-10-19T18:09:56Z] ### Removing VCH ####                        
[34mINFO[0m[2016-10-19T18:09:57Z]                                              
[34mINFO[0m[2016-10-19T18:09:57Z] VCH ID: VirtualMachine:1906                  
[34mINFO[0m[2016-10-19T18:09:57Z] Removing VMs                                 
[34mINFO[0m[2016-10-19T18:09:57Z] Delete will attempt to remove datastore files for VM "drunk_engelbart-9f1ce9ea3dd4f8402d5049999af91256c143b0e55cca9f319efa3171a0b4d86e" 
[31mERRO[0m[2016-10-19T18:09:57Z] --------------------                         
[31mERRO[0m[2016-10-19T18:09:57Z] vic-machine-linux failed: runtime error: invalid memory address or nil pointer dereference
```

emlin:
Fixes #2818 
For another kind of invalid state vm, vsphere might return empty vm.Config property, so govmomi will panic for NPE. This is fixed in #2977 
This PR did following changes to fix this kind of failure:
* Override govmomi vm.Properties method in vm.go, and fix the vm if that's in invalid state, cause this function is not enclosed by waitForResult method.
* Fix waitForResult method, to handle non invalid state error, if vm is already in invalid state. This might include RPC send error.
* Update urfave/cli vendor code, cause it does not hide all error stack in latest version.

### DIFF
--- a/pkg/vsphere/vm/vm.go
+++ b/pkg/vsphere/vm/vm.go
@@ -35,6 +35,19 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/tasks"
 )
 
+const (
+	summary          = "summary"
+	connectionStatus = "summary.runtime.connectionState"
+)
+
+type InvalidState struct {
+	r types.ManagedObjectReference
+}
+
+func (i *InvalidState) Error() string {
+	return fmt.Sprintf("vm %s is invalid", i.r.String())
+}
+
 // VirtualMachine struct defines the VirtualMachine which provides additional
 // VIC specific methods over object.VirtualMachine as well as keeps some state
 type VirtualMachine struct {
@@ -458,7 +471,7 @@ func (vm *VirtualMachine) registerVM(ctx context.Context, path, name string,
 }
 
 // FixInvalidState fix vm invalid state issue through unregister & register
-func (vm *VirtualMachine) FixInvalidState(ctx context.Context) error {
+func (vm *VirtualMachine) fixVM(ctx context.Context) error {
 	log.Debugf("Fix invalid state VM: %s", vm.Reference())
 	folders, err := vm.Session.Datacenter.Folders(ctx)
 	if err != nil {
@@ -466,10 +479,10 @@ func (vm *VirtualMachine) FixInvalidState(ctx context.Context) error {
 		return err
 	}
 
-	properties := []string{"config.files", "summary.config", "summary.runtime", "resourcePool", "parentVApp"}
+	properties := []string{"summary.config", "summary.runtime", "resourcePool", "parentVApp"}
 	log.Debugf("Get vm properties %s", properties)
 	var mvm mo.VirtualMachine
-	if err = vm.Properties(ctx, vm.Reference(), properties, &mvm); err != nil {
+	if err = vm.VirtualMachine.Properties(ctx, vm.Reference(), properties, &mvm); err != nil {
 		log.Errorf("Unable to get vm properties: %s", err)
 		return err
 	}
@@ -481,7 +494,7 @@ func (vm *VirtualMachine) FixInvalidState(ctx context.Context) error {
 		return err
 	}
 
-	task, err := vm.registerVM(ctx, mvm.Config.Files.VmPathName, name, mvm.ParentVApp, mvm.ResourcePool, mvm.Summary.Runtime.Host, folders.VmFolder)
+	task, err := vm.registerVM(ctx, mvm.Summary.Config.VmPathName, name, mvm.ParentVApp, mvm.ResourcePool, mvm.Summary.Runtime.Host, folders.VmFolder)
 	if err != nil {
 		log.Errorf("Unable to register VM %q back: %s", name, err)
 		return err
@@ -513,7 +526,7 @@ func (vm *VirtualMachine) needsFix(ctx context.Context, err error) bool {
 		return true
 	default:
 		if vm.IsInvalidState(ctx) {
-			log.Debugf("vm is invalid")
+			log.Debugf("vm %s is invalid", vm.Reference())
 			return true
 		}
 		log.Debugf("Do not fix non invalid state error")
@@ -523,11 +536,11 @@ func (vm *VirtualMachine) needsFix(ctx context.Context, err error) bool {
 
 func (vm *VirtualMachine) IsInvalidState(ctx context.Context) bool {
 	var o mo.VirtualMachine
-	if err := vm.Properties(ctx, vm.Reference(), []string{"runtime"}, &o); err != nil {
+	if err := vm.VirtualMachine.Properties(ctx, vm.Reference(), []string{connectionStatus}, &o); err != nil {
 		log.Debugf("Failed to get vm properties: %s", err)
 		return false
 	}
-	if o.Runtime.ConnectionState == types.VirtualMachineConnectionStateInvalid {
+	if o.Summary.Runtime.ConnectionState == types.VirtualMachineConnectionStateInvalid {
 		return true
 	}
 	return false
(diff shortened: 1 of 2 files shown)


pkg/vsphere/vm/vm.go
```diff
@@ -512,23 +525,70 @@ func (vm *VirtualMachine) needsFix(err error) bool {
...
+func (vm *VirtualMachine) Properties(ctx context.Context, r types.ManagedObjectReference, ps []string, o *mo.VirtualMachine) error {
+	log.Debugf("get vm properties %s of vm %s", ps, r)
+	contains := false
+	for i := range ps {
+		if ps[i] == summary {
+			contains = true
+			break
+		}
+	}
+	var newps []string
```
hmahmood: You can just do (no need for newps)
```go
if !contains {
  ps = append(ps, connectionStatus)
}
```

pkg/vsphere/vm/vm.go
```diff
@@ -512,23 +525,70 @@ func (vm *VirtualMachine) needsFix(err error) bool {
...
+	if !contains {
+		newps = append(ps, connectionStatus)
```
hmahmood: You are checking for `summary` above, but then appending on `connectionStatus` here? Also, the constant names are less intuitive than the actual strings; makes the code a little less readable imo.

pkg/vsphere/vm/vm.go
emlin: this is because I need to double check properties, and the next time, I don't need to add new properties. So I kept the old variable.

vendor/manifest
```diff
@@ -888,6 +888,14 @@
 			"branch": "v1"
 		},
 		{
+			"importpath": "gopkg.in/urfave/cli.v1",
+			"repository": "https://gopkg.in/urfave/cli.v1",
```
hmahmood: Where are we using this?

pkg/vsphere/vm/vm.go
emlin: any suggestion on the name?

vendor/manifest
emlin: I searched the urfave/cli code, itself has this dependency, so I assume I cannot remove this directory simply.

dougm: Looks like we should be using "gopkg.in/urfave/cli.v1" as the import path in our own code to avoid the duplicate vendoring of this package.  If the NPE fix doesn't depend on updating this dependency, maybe the vendor update + import path change could be done in a separate PR?

pkg/vsphere/vm/vm.go
```diff
@@ -458,18 +471,18 @@ func (vm *VirtualMachine) registerVM(ctx context.Context, path, name string,
...
+func (vm *VirtualMachine) fixVM(ctx context.Context) error {
 	log.Debugf("Fix invalid state VM: %s", vm.Reference())
 	folders, err := vm.Session.Datacenter.Folders(ctx)
 	if err != nil {
 		log.Errorf("Unable to get vm folder: %s", err)
 		return err
 	}
 
-	properties := []string{"config.files", "summary.config", "summary.runtime", "resourcePool", "parentVApp"}
+	properties := []string{"summary.config", "summary.runtime", "resourcePool", "parentVApp"}
```
dougm: While you're in there, looks like we only use Runtime.Host here, maybe limit that to "summary.runtime.host".

vendor/manifest
emlin: Not directly related, just to improve the capability to print out NPE. I can separate it.

pkg/vsphere/vm/vm.go
hmahmood: I would not use the constants.

pkg/vsphere/vm/vm.go
hickeng: Do we actually need to check? What happens if we append "summary.runtime.connectionState" every time, irrespective of whether it's present already?

pkg/vsphere/vm/vm_test.go
```diff
@@ -190,6 +191,13 @@ func TestVM(t *testing.T) {
 		t.Errorf("Failed to find fixed vm: %s", err)
 	}
 	assert.Equal(t, vm.Reference(), newVM.Reference())
+
+	// VM properties
+	var ovm mo.VirtualMachine
```
hickeng: This is exactly the sort of test where we should be using vcsim to return InvalidState - @dougm is there an easy way to do this fault injection?

emlin: @dougm the registervm methods are folder method or vApp method, how to inject the folder or vApp in simulator?

pkg/vsphere/vm/vm.go
emlin: It doesn't matter to query duplicated properties, but as we're trying to make minimum properties query each time, it will help to reduce the array length.
And I think this array length is really short based on our current coding practice, I assume it doesn't hurt to loop that array.

pkg/vsphere/vm/vm_test.go
dougm: Added Register/Unregister: https://github.com/vmware/vic/pull/3068

emlin:
Changed fixed vm condition in WaitForResult for #2958

pkg/vsphere/vm/vm_test.go
```diff
@@ -436,3 +446,145 @@ func TestBfsSnapshotTree(t *testing.T) {
...
+	// Test the success path
+	var o mo.VirtualMachine
+	err = vmm.Properties(ctx, vmo.Reference(), []string{"config", "summary", "resourcePool", "parentVApp"}, &o)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	//	// Inject invalid connection state to vm
+	ref := simulator.Map.Get(vmo.Reference()).(*simulator.VirtualMachine)
+	ref.Summary.Config.VmPathName = "[LocalDS_0] ha-host_VM0/ha-host_VM0.vmx"
```
dougm: Looks like we need to set that field in vcsim.  You can also use this for now, rather than the hardcoded string:
`ref.Config.Files.VmPathName`

dougm:
lgtm

pkg/vsphere/vm/vm_test.go
emlin: good

emlin:
lgtm

Issue #1689, opened 2016-07-26
Add RBAC constraints to the ID that's used by the VCH.


hickeng:
The extension that is registered per VCH has unrestricted admin access currently. We need to apply RBAC constraints to this extension ID if that is possible.
- [X] determine if it is possible to apply RBAC to an extension ID
  - NOT POSSIBLE, RBAC REQUIRES USER
- [x] determine what RBAC privileges need to be specified for a VCH
  - [x] confirm VCH function with non-admin user
  - [x] confirm VCH function with RBAC constrained user
  - [x] document required privileges
- [ ] add tests to assert correctness of privilege docs #3951 

Optionally:
- [ ] programmatically apply RBAC privileges when creating VCH.

https://bugzilla.eng.vmware.com/show_bug.cgi?id=1729956

mdubya66:
who should get this one?

hickeng:
Added #2875 for supplying the creds of a separate operations user (precreated service account)

dougm:
govmomi PR: vmware/govmomi#628

dougm:
With the PR mentioned above, next steps would be in vic-machine:

* We can create a new role using govmomi object.AuthorizationManager.AddRole (see govc/role/create.go for reference)

* Role privilege set still TBD, but can be limited to specific entities in the step below (vs all managed entities as the Admin role is).  Note that the full privilege set can be viewed using `govc role.ls Admin`

* Use govmomi object.AuthorizationManager.SetEntityPermissions to grant principal (-op-username) role permission on the VCH vapp/resource-pool entity (see govc/permission/set.go for reference)

hickeng:
Started running through the necessary permissions.

For ESX:
* assign role to entire host - needed for finder operation or cannot discover cluster, pool, datastore, etc.
* low level file control - needed for KV store creation
* host->system management - needed for session ticket/file upload/download
* vm->configuration->add new disk

Going to drop this for now and just add the PR for a separate operations user

hickeng:
Dumping the RBAC config from a working VCH deployed on vCenter 6.5:

```
vagrant@devbox:~$ for i in "VCH - vcenter" "VCH - datacenter" "VCH - datastore" "VCH - network" "VCH - endpoint"; do  echo ======= Role $i ============;echo "******" Configuration; govc role.ls "$i"; echo;echo "******" Usage; govc role.usage "$i"; echo;echo;done  | grep -v '^System\.'
======= Role VCH - vcenter ============
****** Configuration
[...]

Issue #635, opened 2016-05-04
Milestone: service account for VCH user


hickeng:
Create a service account configured with appropriate RBAC limits for use by VCH.
This may well get promoted to an epic.

This can start with the following for ESX, but we need to find the equivilent for vCenter.
http://pubs.vmware.com/vsphere-60/index.jsp?topic=%2Fcom.vmware.wssdk.apiref.doc%2Fvim.host.LocalAccountManager.html

gigawhitlocks:
Going to capture some notes from a Slack conversation with @hickeng here mostly for my own reference, but might be useful for others too later:

> The 6.0 docs center (http://pubs.vmware.com/vsphere-60/index.jsp#com.vmware.ssosdk.pg.doc/sso_pg_usage.3.2.html) has some useful information about SSO, however we may be best off asking for guidence from the PSC (platform services controller) team as that's the SSO implementation in 6.0+ iirc.
> 
> For the API details, this is a good place to start: http://pubs.vmware.com/vsphere-60/index.jsp#com.vmware.wssdk.pg.doc/PG_Authenticate_Authorize.8.1.html
> 
> The user models for vCenter and ESX are somewhat different. From an implementation perspective, while we can fall back to asking the viadmin to create a service account for a VCH we shouldn't ask them to supply the SSO token,  so that's our implementation priority.

hickeng:
Moving this to onsite as it's not feasible for TP3.

mdubya66:
Moving to backlog.

corrieb:
@hickeng Everything this relates to seems to be either in icebox or untargeted. I know we want to prioritize this kind of stuff, but seems like there's no way this is making onsite.

lcastellano:
Replaced by: https://github.com/vmware/vic/issues/6488

PR #2993, opened 2016-11-03
Remove use of VC extension [full ci]

dougm:
Follow-on to #2931 and towards #635 , #1689

The LoginExtensionByCertificate API is deprecated and has showstoppers for VIC:

- Not possible to use RBAC with an extension

- Proxy connection to sdk-tunnel vhost does not play well with Go's
  built-in certificate verification, nor is it possible to use the
  thumbprint verfication as we do with direct TLS connections.

VCH config changes:

- Connection.Target is a string, avoids encoding all url.URL fields

- Connection.Target is pruned by vic-machine to only include URL.{Scheme,Host,Path}

- Add Connection.Username and Connection.Token fields, where Token is secret

- Separate set of vSphere "operation" credentials can be provided to vic-machine for use
  by the VCH

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -15,7 +15,6 @@
 package create
 
 import (
-	"bytes"
 	"crypto/tls"
 	"encoding"
 	"fmt"
@@ -989,15 +988,6 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
 	vConfig.HTTPSProxy = c.HTTPSProxy
 
 	vchConfig.InsecureRegistries = c.Data.InsecureRegistries
-
-	if validator.Session.IsVC() { // create certificates for VCH extension
-		var certbuffer, keybuffer bytes.Buffer
-		if certbuffer, keybuffer, err = certificate.CreateSelfSigned("", []string{"VMware Inc."}, 2048); err != nil {
-			return errors.Errorf("Failed to create certificate for VIC vSphere extension: %s", err)
-		}
-		vchConfig.ExtensionCert = certbuffer.String()
-		vchConfig.ExtensionKey = keybuffer.String()
-	}
 
 	// separate initial validation from dispatch of creation task
 	log.Info("")
--- a/cmd/vicadmin/vicadm.go
+++ b/cmd/vicadmin/vicadm.go
@@ -85,8 +85,6 @@ var (
 	defaultReaders map[string]entryReader
 
 	datastore types.ManagedObjectReference
-
-	datastoreInventoryPath string
 )
 
 type logfile struct {
@@ -394,6 +392,7 @@ func (r datastoreReader) open() (entry, error) {
 
 func vSphereSessionGet(sessconfig *session.Config) (*session.Session, error) {
 	session := session.NewSession(sessconfig)
+	session.UserAgent = version.UserAgent("vic-admin")
 	ctx := context.Background()
 	_, err := session.Connect(ctx)
 	if err != nil {
@@ -431,21 +430,6 @@ func client(config *vicAdminConfig) (*session.Session, error) {
 	return sess, nil
 }
 
-func findDatastore() error {
-	defer trace.End(trace.Begin(""))
-
-	session, err := client(&rootConfig)
-	if err != nil {
-		return err
-	}
-	defer session.Client.Logout(context.Background())
-
-	datastore = session.Datastore.Reference()
-	datastoreInventoryPath = session.Datastore.InventoryPath
-
-	return nil
-}
-
 type flushWriter struct {
 	f http.Flusher
 	w io.Writer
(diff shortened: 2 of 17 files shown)


dougm:
@hickeng add the operation user as requested.  Stuck with using `--force` to default to the credentials used by vic-machine for now, avoids having to make robot test changes just yet - but did a manual test for now:
```
govc host.account.create -id $USER -password "$(govc env GOVC_PASSWORD)"
govc permissions.set -principal $USER -role Admin
```

Where `$USER` == "dougm" in my case.

```
./bin/vic-machine-linux create --user root --password ... --op-user $USER --op-password ... ...
```

```
% govc vm.info -json $USER-vch | jq '.VirtualMachines[].Config.ExtraConfig | map(select(.Key == "guestinfo.vice./connect/username"))'
[
  {
    "Key": "guestinfo.vice./connect/username",
    "Value": "dougm"
  }
]
```

caglar10ur:
LGTM

cmd/vic-machine/create/create.go
```diff
@@ -454,7 +453,7 @@ func (c *Create) Flags() []cli.Flag {
 func (c *Create) processParams() error {
 	defer trace.End(trace.Begin(""))
 
-	if err := c.HasCredentials(); err != nil {
+	if err := c.HasOpCredentials(c.Force); err != nil {
```
emlin: This method will check operation credentials, and admin credentials, why not reuse the name HasCredentials?

cmd/vicadmin/vicadm.go
```diff
@@ -445,23 +436,6 @@ func main() {
...
-			vchConfig.Target.Path))
-		vchConfig.Target = *newurl
-	}
-
-	// FIXME: these should just be consumed directly inside Session
-	config.Service = vchConfig.Target.String()
-	config.ExtensionCert = vchConfig.ExtensionCert
-	config.ExtensionKey = vchConfig.ExtensionKey
-	config.ExtensionName = vchConfig.ExtensionName
-	config.Thumbprint = vchConfig.TargetThumbprint
```
emlin: so the thumbprint is not required as well?

emlin:
I'm wondering if we need to check operation credential's RBAC in vic-machine validator. That user should have enough privilege to mange vmdk and vms under specified resource pool, otherwise, we'll have to check portlayer logs to figure out what's wrong.
And that might not be intuitive for customer, cause vic-machine using admin user already did everything correctly.

dougm:
@emlin yes we do need the RBAC checks, just not part of this PR.

cmd/vic-machine/create/create.go
emlin: yeah, right

emlin:
lgtm 👍

vburenin:
LGTM

pkg/version/version.go
```diff
@@ -54,6 +55,15 @@ func String() string {
 	return GetBuild().String()
 }
 
+// UserAgent returns component/version in HTTP User-Agent header value format
+func UserAgent(component string) string {
```
hickeng: nice

pkg/vsphere/session/session.go
```diff
@@ -160,41 +152,23 @@ func (s *Session) Connect(ctx context.Context) (*Session, error) {
...
-	// "sdkTunnel" does not pass Go's tls.VerifyHostname check.
-	// We are moving away from LoginExtensionByCertificate anyhow, so disable thumbprint checks for now.
-	if s.HasCertificate() {
-		s.Insecure = true
-	}
-
 	// Update the service URL with expanded defaults
 	s.Service = soapURL.String()
 
+	// VCH components do no include credentials within the target URL
```
hickeng: `do not include`

cmd/vicadmin/vicadm.go
```diff
@@ -465,22 +449,9 @@ func main() {
...
-			vchConfig.Target.Path))
-		vchConfig.Target = *newurl
-	}
-
 	// FIXME: these should just be consumed directly inside Session
-	rootConfig.Service = vchConfig.Target.String()
-	rootConfig.ExtensionCert = vchConfig.ExtensionCert
-	rootConfig.ExtensionKey = vchConfig.ExtensionKey
-	rootConfig.ExtensionName = vchConfig.ExtensionName
+	rootConfig.Service = vchConfig.Target
```
hickeng: Do we need this here if session.Create is using vchConfig.Target directly?

dougm: session.Create() isn't used by vicadmin currently.

pkg/vsphere/session/session.go
dougm: fixed

hickeng:
LGTM

dougm:
.

PR #3026, opened 2016-11-04
fix typo in create help

### LINKED ISSUE #3010: Typo in vic-machine help?
Using build 6507 

In the vic-machine extended help there is: 

`--container-network-gateway value, --cng value   Gateway for the container network's subnet in CONTAINER-NETWORK:SUBNET format, e.g. vsphere-net:172.16.0.0/16.`

Shouldn't the example 172.16.0.1? Also, why do we need to specify the subnet for the GW as well? We already specify it with the `--container-network-ip-range` parameter. 

All in all I would have expected the e.g. to be something like: 

`e.g. vsphere-net:172.16.0.1.`

emlin:
Fixes #3010


caglar10ur:
Lgtm

andrewtchin:
lgtm

emlin:
♻️

PR #3038, opened 2016-11-07
[full ci] Add retry support for upstream functions

caglar10ur:
Commit can fail with ConcurrentAccess error and when that happens
upstream has to replay the operation on up2date handle and re-submit.

**This change assumes that every operation on the retry path is idempotent**


emlin:
Is there any reason to retry in docker persona, instead of in portlayer vm.WaitForResult?

caglar10ur:
@emlin yes we need persona to make the same changes on the handle when we get a conflict error. This is not just retrying the task, it's re-creating the same task based on the updated world.

emlin:
@caglar10ur even that's the case, why should we expose the conflict error back to docker persona? Shouldn't that be responsibility of portlayer?

caglar10ur:
@emlin Don't think so. Personality is the caller, it knows what operations are needed. Think this like a HTTP conflict error, when that happens client needs to send another query (or drop it etc.) and we don't expect http server to try it again by itself.

emlin:
hm... but we're still retrying in docker persona. This will make the operation sequence random. 
Anyway, the code LGTM.

lib/apiservers/engine/backends/container.go
```diff
@@ -143,6 +152,39 @@ func NewContainerBackend() *Container {
...
 }
 
+// Retry retries the given function DefaultRetryCount times while sleeping some time between unsuccesfull attempts
+func Retry(op func() error) error {
+	defer trace.End(trace.Begin(""))
+
+	var err error
+
+	// DefaultRetryInterval -> [DefaultRetryInterval - delta, DefaultRetryInterval + delta]
+	delta := DefaultRetryInterval.Seconds() * DefaultRandomizationFactor
```
hmahmood: You may want to make `delta`, `min`, and `max` constants as well.

lib/apiservers/engine/backends/container.go
```diff
@@ -143,6 +152,39 @@ func NewContainerBackend() *Container {
...
+	min := int(DefaultRetryInterval.Seconds() - delta)
+	max := int(DefaultRetryInterval.Seconds() + delta)
+
+	for retry := 1; retry <= DefaultRetryCount; retry++ {
+		err = op()
+		if err != nil && IsConflictError(err) {
+			// Pick a random integer from [min, max] range and multiply it with the index
+			next := time.Duration(retry * (rand.Intn(max-min+1) + min))
```
hmahmood: May want to look at how we do backoff here: https://github.com/vmware/vic/blob/master/pkg/vsphere/tasks/waiter.go#L76. Maybe make the backoff code modular so both places can use it?

lib/apiservers/engine/backends/container.go
```diff
@@ -143,6 +152,39 @@ func NewContainerBackend() *Container {
...
+			log.Errorf("Operation failed with Concurrent modification detected error, will try again after sleeping %s (attempt #%d)", next, retry)
+			time.Sleep(next)
+
+			continue
+		}
+		if err != nil {
+			log.Errorf("Operation failed with %#+v", err)
+			return err
+		}
+		if err == nil {
```
hmahmood: Well `err == nil` goes without saying here.

hmahmood:
lgtm

It would be nice if both the tasks library and this code used the same backoff mechanism.

lib/apiservers/engine/backends/container.go
```diff
@@ -69,6 +70,11 @@ const (
 	MinCPUs = 1
 	// DefaultCPUs - the default number of container VM CPUs
 	DefaultCPUs = 2
+
+	// Random numbers generated by fair dice roll :)
+	DefaultRetryCount          = 5
+	DefaultRandomizationFactor = 0.5
+	DefaultRetryInterval       = 2 * time.Second
```
vburenin: Have you considered randomized exponential retries?

lib/portlayer/exec/commit.go
```diff
@@ -138,45 +135,33 @@ func Commit(ctx context.Context, sess *session.Session, h *Handle, waitTime *int
...
-				// FIXME!!! this is a temporary hack until the concurrent modification retry logic is in place
-				if refresh {
-					base, err := h.updates(ctx)
-					if err == nil {
-						h.Runtime = base.Runtime
-						h.Config = base.Config
-					}
-				}
+			// poor man's test and set
+			s.ChangeVersion = h.Config.ChangeVersion
```
emlin: With this settings, the tasks retry will not work, right?
If there is task in progress error, in next retry, the configuration version is already changed by other finished task.
Right now, the vm reconfigure for VCH is locked, the container reconfigure cannot be retried for this setting, I think keep the task retry logic seems misleading.

caglar10ur: You are right, task in progress will fail but I'm not sure whether we should treat it like a conflict error. Right now we get a TaskInProgress error first and then retry in the task package and get ConccurrentAccess error and return that and retry the whole operation in persona.

emlin:
lgtm. But you might want to separate vendor code. It gonna be something new for 0.8

lib/portlayer/exec/commit.go
```diff
@@ -138,45 +135,33 @@ func Commit(ctx context.Context, sess *session.Session, h *Handle, waitTime *int
...
 
-				s.ChangeVersion = h.Config.ChangeVersion
+			// nilify ExtraConfig if vm is running
+			if h.Runtime.PowerState == types.VirtualMachinePowerStatePoweredOn {
```
hickeng: can we change this to != poweredOff?

lib/apiservers/engine/backends/container.go
```diff
@@ -69,6 +70,13 @@ const (
 	MinCPUs = 1
 	// DefaultCPUs - the default number of container VM CPUs
 	DefaultCPUs = 2
+
+	// Random numbers generated by fair dice roll :)
+	DefaultInitialInterval     = 2 * time.Second
```
hickeng: what is the minimum that this results in? Ideally we'd retry immediately on the first failure as our most common scenario is that there are only two operations competing (attach and start) and having received ConflictError once means that the competing operation has completed.

caglar10ur: They are randomly picked by me, you can experiment with them via

```
package main

import (
        "fmt"
        "time"

        "github.com/cenkalti/backoff"
)

const (
        DefaultInitialInterval     = 2 * time.Second
        DefaultRandomizationFactor = 0.5
        DefaultMultiplier          = 2
        DefaultMaxInterval         = 30 * time.Second
        DefaultMaxElapsedTime      = 1 * time.Minute
)

type Clock struct {
        now time.Time
}

func (c *Clock) Now() time.Time {
        return c.now
}

func (c *Clock) Sleep(i time.Duration) {
        c.now = c.now.Add(i)
}
func main() {
        clock := &Clock{now: time.Now()}
        b := &backoff.ExponentialBackOff{
                InitialInterval:     DefaultInitialInterval,
                RandomizationFactor: DefaultRandomizationFactor,
                Multiplier:          DefaultMultiplier,
                MaxInterval:         DefaultMaxInterval,
                MaxElapsedTime:      DefaultMaxElapsedTime,
                Clock:               clock,
        }
        // Reset the interval back to the initial retry interval and restart the timer
        b.Reset()
        for {
                next := b.NextBackOff()
                if next == backoff.Stop {
                        break
                }
                fmt.Printf("[%s]\n", next)
                clock.Sleep(next)
        }
}
```

running them 3 times shows following distribution

```
[conur@vir:/opt/go/src/github.com/vmware/vic(retry)] go run retry.go
[2.209320576s]
[5.762036353s]
[9.316480426s]
[15.003426995s]
[27.739124912s]
[conur@vir:/opt/go/src/github.com/vmware/vic(retry)] go run retry.go
[2.209320576s]
[5.762036353s]
[9.316480426s]
[15.003426995s]
[27.739124912s]
[conur@vir:/opt/go/src/github.com/vmware/vic(retry)] go run retry.go
[2.209320576s]
[5.762036353s]
[9.316480426s]
[15.003426995s]
[27.739124912s]
```

caglar10ur: 2 second to 500 ms

```
[conur@vir:/opt/go/src/github.com/vmware/vic(retry)] go run retry.go
[744.885075ms]
[1.295280183s]
[1.807981249s]
[5.942740531s]
[8.350248657s]
[18.68257744s]
[19.290742954s]
[39.812197537s]
[conur@vir:/opt/go/src/github.com/vmware/vic(retry)] go run retry.go
[296.411109ms]
[1.228648661s]
[1.986901726s]
[2.143151871s]
[11.24980646s]
[13.672829257s]
[29.271093666s]
[19.898402297s]
[conur@vir:/opt/go/src/github.com/vmware/vic(retry)] go run retry.go
[293.261689ms]
[1.331998864s]
[1.895614979s]
[5.448026869s]
[4.491979661s]
[21.151462222s]
[42.776509135s]
```

lib/portlayer/exec/commit.go
caglar10ur: done

lib/apiservers/engine/backends/container.go
caglar10ur: using a very small initial number and a big multiplier to start fast and then slow down

caglar10ur:
* waiting legal to approve 3rd party packages

mdubya66:
don't wait for legal approval. lgtm.

caglar10ur:
♻️

PR #3050, opened 2016-11-07
Not logging command arguments and environment when logging config [full ci]

### LINKED ISSUE #1659: Not enough information in build logs to debug container issues
Currently only the log from the container and vmware.log are present in the build log for a failed build. This currently does not give enough information, for instance, to debug networking issues. We need a way get more information about the container in the build logs. Maybe a dump of guestinfo, and other commands like we do for the VCH.

hmahmood:
Fixes #1659


lib/tether/tether.go
```diff
@@ -584,14 +582,37 @@ func logConfig(config *ExecutorConfig) {
...
+			if _, ok := keys[k]; !ok {
+				sink[k] = v
+			}
+
+			return nil
+		},
+		config,
+	)
+
+	for k, v := range sink {
```
hickeng: Why not log directly in the func passed to Encode?

hmahmood: The logging is in there is affected by the log level in the `extraconfig` package, which could be different from what the tether is using.

pkg/vsphere/extraconfig/keys.go
```diff
@@ -262,3 +262,114 @@ func addPrefixToKey(header, prefix, key string) string {
...
+	log.Debugf("v=%#v, field=%#v, prefix=%#v", v, field, prefix)
+	if v.Kind() == reflect.Ptr {
+		return calculateKeys(v.Elem(), field, prefix)
+	}
+
+	if field == "" {
+		return []string{prefix}
+	}
+
+	s := strings.SplitN(field, ".", 2)
```
fdawg4l: I'm new to this guest info stuff.  Is this a field separator inherent to extraconfig?  If so does this conflict or break with this (https://github.com/vmware/vic/pull/3062/files)?

fdawg4l:
LGTM

caglar10ur:
LGTM (would love to convert | to a const so that we can change it easily but that's for another PR)

hmahmood:
.

hmahmood:
♻️

PR #3062, opened 2016-11-08
Replace hardcoded separator with ·

caglar10ur:
Unicode middle dot is safe to use with guestinfo and solves the
ambiguity between ```1|2|3``` and ```dmesg | grep vmxnet```


dougm:
lgtm

anchal-agrawal:
We might need to change `ui/vic-ui/swf/src/main/flex/com/vmware/vicui/constants/AppConstants.as` too. @jooskim?

anchal-agrawal:
lgtm

caglar10ur:
♻️

PR #3064, opened 2016-11-08
Change name to VCH Admin Portal

### LINKED ISSUE #2804: Inconsistent naming of VIC Admin between vic-machine and UI plug-in
We currently use two completely different names for VIC Admin:
- The output of `vic-machine create` refers to the `vic-admin portal`.
- The Virtual Container Host portlet in the UI plug-in refers to `Docker log`.

The names used in both places should be consistent. Furthermore, trademarking rules dictate that we should avoid using the VIC acronym in customer-visible text wherever possible. 

So, could we use "VCH Admin Portal" throughout? I think that this reflects the purpose of the portal more accurately than "VIC Admin" and avoids abbreviating the product name.

mhagen-vmware:
Fixes #2804

### DIFF
--- a/lib/install/management/inspect.go
+++ b/lib/install/management/inspect.go
@@ -103,7 +103,7 @@ func (d *Dispatcher) ShowVCH(conf *config.VirtualContainerHostConfigSpec, key st
 	}
 
 	log.Infof("")
-	log.Infof("vic-admin portal:")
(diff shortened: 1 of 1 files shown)


anchal-agrawal:
lgtm

emlin:
lgtm

Issue #2329, opened 2016-09-15
Unit test CI build failure in pkg test


emlin:
see https://ci.vmware.run/vmware/vic/4784
# github.com/vmware/vic/vendor/github.com/vmware/govmomi/object

vendor/github.com/vmware/govmomi/object/host_service_system.go:25: can't find import: "github.com/vmware/vic/vendor/github.com/vmware/govmomi/vim25/types"
Makefile:227: recipe for target 'test-job-github.com/vmware/vic/pkg' failed
make: **\* [test-job-github.com/vmware/vic/pkg] Error 2
make: **\* Waiting for unfinished jobs....

emlin:
And didn't see any file imported github.com/vmware/vic/vendor/github.com/vmware/govmomi/vim25/types

hickeng:
You're looking for an import of `github.com/vmware/govmomi/vim25/types` - Go should be honoring the vendor directory as of 1.5 experimental iirc.

mhagen-vmware:
https://ci.vmware.run/vmware/vic/4906

gigawhitlocks:
Swiping this one from you @emlin

gigawhitlocks:
This is really strange; I was able to reproduce this exactly once (though actually it failed on a different package)...

gigawhitlocks:
Quite a heisenbug. Doesn't reproduce in a clean golang container on the troubled commit or `master`, or in a golang container w/ a volume mount to source & possibly artifacts from previous builds; doesn't reproduce before or after a clean build. 

We should probably just close this and reopen it if the same exact error is seen again and/or the issue becomes reproducible.

emlin:
> This is really strange; I was able to reproduce this exactly once (though actually it failed on a different package)...

I could understand such kind of issue probably happen anywhere, that just not error inside of our code base. It's but in either docker or go test

mhagen-vmware:
Eh, it has already happened three times. (Once for an Ian build that you restarted)  It is a challenging race condition but it keeps happening.  I definitely don't want to do nothing here and just close it. Again, is there more logging at least that we can add?

gigawhitlocks:
Reopened at @mdubya66 's request;

Also this
>  It is a challenging race condition but it keeps happening
[...]

PR #3077, opened 2016-11-09
List the path that keeps failing

mhagen-vmware:
For #2329 again


vburenin:
LGTM

rajanashok:
LGTM

mhagen-vmware:
.

hickeng:
@mhagen-vmware can we have slightly more context in the commit - this is confusing in the commit log. A prefix or similar that denotes it as an infra element rather than product code would more than suffice.

PR #3103, opened 2016-11-10
Add the full round robin for network interfaces

mhagen-vmware:
Not clear exactly what is going on here, but filing an issue to clarify.  For now, just allow that we appear to be randomly round robin'ing on which interfaces we report as sharing the same NIC.


mhagen-vmware:
#3104

matthewavery:
What happens if none of them share a NIC?

matthewavery:
lgtm

andrewtchin:
lgtm

mhagen-vmware:
.

PR #3105, opened 2016-11-10
Added log rotation.

### LINKED ISSUE #2848: Appliance log rotation
Currently the appliance logs grow unbounded and are not rotated. The logs need to have size caps, probably configured based on DebugLevel and configured appliance memory.

There is a related issue for persistence on shutdown #2079.

With the current setup, the appliance will fill up it's memory and then, hopefully, reboot. Marked as security as this is significantly potential DoS vector.

https://bugzilla.eng.vmware.com/show_bug.cgi?id=1727660

vburenin:
I added log rotation for VCH. However, this is not the final version. A lot more needs to be added during next iterations, see TODOs.

Several things to note:
1. Logs are not uploaded to vSphere storage.
2. cron is not used to run logrotate, since it pulls a lot of extra dependencies into the system. it is happening inside vic-init in a separate goroutine.
3. logrotate config is dynamic/generated, so when we are ready we can change it depending on available RAM and user requirements.
4. logrotate runs every 5 minutes, not too frequent, but not rare to avoid situation when logs are growing too large.
5. No tests yet unless, I get at least one LGTM.

Fixes #2848


caglar10ur:
Why we are generating a logrotate config? We could just ship a static one with the ISO.

vburenin:
@caglar10ur Depending on memory limits, log file size may be different.

cmd/vic-init/main_linux.go
```diff
@@ -90,6 +92,21 @@ func main() {
 	toolbox.PrimaryIP = externalIP
 	tthr.Register("Toolbox", toolbox)
 
+	// Check logs every 5 minutes and rotate them if their size exceeds 20MB.
+	// The history size we keep is 2 previous files in a compressed form.
+	// TODO: Check available memory to tune log size and history length for log files.
+	logrotate, err := logmgr.NewLogManager(time.Second * 300)
+	const maxLogSize = 20 * 1024 * 1024
```
jzt: minor: `maxLogSizeBytes`

pkg/logmgr/logmgr.go
```diff
@@ -0,0 +1,237 @@
...
+	rotateInterval RotateInterval
+	logFilePath    string
+	logFileName    string
+	maxLogSize     int64
+	maxLogFiles    int64
+	compress       bool
+}
+
+// ConfigFileContent formats log configuration according to logrotate requirements.
+func (lrc *logRotateConfig) ConfigFileContent() string {
```
hmahmood: You could just use json for the config file format and let Go do most of the marshaling/unmarshaling.

hmahmood:
lgtm

caglar10ur:
Not sure I'm on board with this approach.

Why log file size is a function of a memory size? I know the vch fs is a tmpfs but we can just pick a reasonable file size (say 128MB) and be done with it instead of keeping some code that generates 5 line of config file.

Also I think logrotate will require cooperation from the daemons (like sighup support) otherwise they will keep writing the same inode so I'm guessing you are going to adress that?

vburenin:
@caglar10ur Reasonable may mean very different things for different people, so I would rather have flexibility here when needed. You are right that log rotate requires cooperation with daemon to do things in the right way, but it also can handle situation when application doesn't want to cooperate, so yes, it is addressed.

fdawg4l:
@vburenin It's reasonable to rotate logs to save space (by way of compressing the previous log files).  But I don't think it's a good idea to merge this until the log bundle includes all archived and rotated log files.   The worst case scenario is having logfile A include an error which logfile B has more details in, but logfile B has been rotated so we can't triage the issue.

vburenin:
@fdawg4l Does it mean you want me to update vic-admin accordingly?

fdawg4l:
I would check with @hickeng, but I think we can't have one without the other.

vburenin:
@fdawg4l I updated vicadmin to include archived logs.
However, I there is a very rare chance of a race condition between logrotate and logs export.
But it only can happen once per 5 minutes and when one of VCH log files exceeded 20MB.

I would not deal with cross process synchronization right now. there are more important things to work on at the moment.

hmahmood:
What is the consequence of the race condition? Do we crash, simply not get the full logs, or something else?

vburenin:
@hmahmood Only one consequence that can happen is that one or two files may not be included.

cmd/vicadmin/logs.go
```diff
@@ -86,12 +90,33 @@ func (r dlogReader) open() (entry, error) {
 	return newBytesEntry(name+".log", buf.Bytes()), nil
 }
 
+// logFiles has a potential race condition since logrotation can rotate files this moment of time.
+// however, the likely hood of this race is so low that it doesn't worth investing the time to do
```
hmahmood: Famous last words? :-)

vburenin: haha, in this case it is just a collection of log data, the most recent logs will be collected in any case.

vburenin:
Ok. I added file based lock. It is way out for scope, since it is a practically a new feature. However, it eliminates race condition with logrotate runner and log exports.

pkg/filelock/flock_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	delta := (et.UnixNano() - st.UnixNano()) / 1000000
+	if delta < 50 {
+		t.Errorf("Wait time is less than 50: %d", delta)
+	}
+}
+
+func TestManyLocks(t *testing.T) {
+	lockName := "test_lock" + strconv.FormatInt(time.Now().UnixNano(), 10)
+	baseLock := NewFileLock(lockName)
```
hmahmood: Can you test without `baseLock`? i.e. just run the loop below?

fdawg4l:
lgtm

Issue #3083, opened 2016-11-09
vicadmin leaking vSphere sessions


dougm:
With #2883 hitting the vicadmin page in chrome, even without logging in, creates 2 vSphere sessions.  Hitting the page with curl creates 1 session.  Example:

```
% govc session.ls | wc -l
6
```

Open vicadmin in Chrome w/o login, e.g. https://172.16.124.186:2378/authentication

```
% govc session.ls | wc -l
8
```

Reload the page:

```
% govc session.ls | wc -l
10
```

Hit it with curl:

```
% curl -sk https://172.16.124.186:2378/authentication >/dev/null
% govc session.ls | wc -l
11
```

Note that PR #2993 sets the User-Agent header, so you can see which vic component is creating the session(s).

dougm:
@hickeng mentioned @anchal-agrawal was looking at removing a call to the validator, which could be related.

anchal-agrawal:
@dougm Yes, I'm removing the validator requirement from the login page, so we should see fewer sessions being created/leaked.

hickeng:
Adding this to 0.8 as it both a potential DoS path against vSphere and impacts diagnostics

gigawhitlocks:
I will look into this while I address #3088

gigawhitlocks:
This should be resolved, at least for the most part. Need to verify that we don't still spawn a session each time that a user w/ cert auth hits a protected endpoint, but w/r/t browser-based users now, only one session will be created per login and this session will be destroyed automatically after 24 hours or if the user logs out.

hmahmood:
@gigawhitlocks Should we close this issue?

mdubya66:
Closing based on the above.

PR #3106, opened 2016-11-10
Diagnostic improvements to vicadmin [full ci]

### LINKED ISSUE #2811: vicadmin fails to return data if connection to vsphere fails
If the [session.Connect call](https://github.com/vmware/vic/blob/master/cmd/vicadmin/vicadm.go#L391) fails then the Validator [created here](https://github.com/vmware/vic/blob/master/cmd/vicadmin/server.go#L272) is invalid (the session is nil) and there are multiple panics.

If altering [the creation of the validator](https://github.com/vmware/vic/blob/master/lib/vicadmin/validate.go#L80) to check for validity of the session and return `nil, fmt.Errorf("invalid session")` then nothing is rendered.

This failure to connect could be:
- incorrect thumbprint
- network outage
- firewall config issue
- vsphere credentials have changed
- ... (multitude of options)

Adding this to the backlog for GA as it significantly impacts our ability to diagnose some of the most common connectivity issues.

### LINKED ISSUE #3082: vic-admin should work even it can't speak to vsphere.
When the customer hit #3081 the port layer would just quit when it couldn't connect to the target. tether/vic-init would restart it and it would just quit again. During this time you could connect to vic-admin but it never returned any data. When the rest of the system is not working we need vic-admin the most. See also #3079.

### LINKED ISSUE #3086: vic-admin: add endpointVM system time to vic-admin pages
We should add the endpoint VM system time to the vic-admin pages, including the login page.
This is specifically useful in the case of clock-skew causing certificate validation errors as per #2670 

@stuclem when added this should be mentioned in the troubleshooting related to clock-skew

anchal-agrawal:
The vicadmin page doesn't function correctly if we don't have a vSphere session connection. This PR makes several changes to aid debugging.

Fixes #2811 
Fixes #3086 
Fixes #3082
Towards #3083

### DIFF
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -19,10 +19,12 @@ import (
 	"compress/gzip"
 	"crypto/tls"
 	"crypto/x509"
+	"fmt"
 	"html/template"
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"path/filepath"
 	"strings"
 	"time"
@@ -46,11 +48,18 @@ type server struct {
 	uss  *UserSessionStore
 }
 
+// LoginPageData contains items needed to render the login page template
+type LoginPageData struct {
+	Hostname   string
+	SystemTime string
+}
+
 type format int
 
 const (
 	formatTGZ format = iota
 	formatZip
+	VCHUnreachableMsg = " is not functioning: unable to connect to vSphere."
 )
 
 var beginningOfTime = time.Unix(0, 0).Format(time.RFC3339)
@@ -246,7 +255,7 @@ func (s *server) Authenticated(link string, handler func(http.ResponseWriter, *h
 // renders the page for login and handles authorization requests
 func (s *server) loginPage(res http.ResponseWriter, req *http.Request) {
 	defer trace.End(trace.Begin(""))
-	ctx := context.Background()
+
 	if req.Method == "POST" {
 		// take the form data and use it to try to authenticate with vsphere
 
@@ -278,7 +287,7 @@ func (s *server) loginPage(res http.ResponseWriter, req *http.Request) {
 		if err != nil || usersession == nil {
 			// something went wrong or we could not authenticate
 			log.Warnf("User %s from %s failed to authenticated at %s", user, req.RemoteAddr, time.Now())
-			http.Error(res, "Authentication failed due to incorrect credential(s)", 400)
+			http.Error(res, "Authentication failed due to incorrect credential(s)", http.StatusBadRequest)
 			return
 		}
 
@@ -326,16 +335,15 @@ func (s *server) loginPage(res http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	// Render login page (shows up on non-POST requests):
-	sess, err := client(&rootConfig)
-	if err != nil {
-		log.Errorf("Could not render login page due to vSphere connection error: %s", err.Error())
-		http.Error(res, genericErrorMessage, http.StatusInternalServerError)
-		return
+	// Render login page (shows up on non-POST requests)
+	hostName, _ := os.Hostname()
+	loginPageData := &LoginPageData{
+		Hostname:   hostName,
+		SystemTime: time.Now().Format(time.UnixDate),
 	}
-	v := vicadmin.NewValidator(ctx, &vchConfig, sess)
+
 	tmpl, err := template.ParseFiles("auth.html")
-	err = tmpl.ExecuteTemplate(res, "auth.html", v)
+	err = tmpl.ExecuteTemplate(res, "auth.html", loginPageData)
 	if err != nil {
 		log.Errorf("Error parsing template: %s", err)
 		http.Error(res, genericErrorMessage, http.StatusInternalServerError)
@@ -511,11 +519,40 @@ func (s *server) tailFiles(res http.ResponseWriter, req *http.Request, names []s
 	}
 }
 
+// deriveDisplayError takes a vSphere session error and returns
+// an error string that is safe to display on the vicadmin page.
+func deriveDisplayError(err error) string {
+	var displayErrStr string
+
+	if err != nil {
+		switch err := err.(type) {
+		case session.SDKURLError:
+			displayErrStr = fmt.Sprintf("SDK URL could not be parsed: %s", err.Err)
+		case session.SoapClientError:
+			displayErrStr = "unable to obtain a vim client"
+		case session.UserPassLoginError:
+			displayErrStr = "unable to log in with username and password"
+		default:
+			displayErrStr = err.Error()
+		}
+	}
+
+	return displayErrStr
+}
+
 func (s *server) index(res http.ResponseWriter, req *http.Request) {
 	defer trace.End(trace.Begin(""))
 	ctx := context.Background()
 	sess, err := s.getSessionFromRequest(req)
 	v := vicadmin.NewValidator(ctx, &vchConfig, sess)
+
+	if sess == nil {
+		// We're unable to connect to vSphere, so display an error message
+		displayErrStr := deriveDisplayError(err)
+		v.VCHIssues = template.HTML(fmt.Sprintf(
+			"<span class=\"error-message\">%s Error: %s</span>\n",
+			v.Hostname+VCHUnreachableMsg, displayErrStr))
+	}
 
 	tmpl, err := template.ParseFiles("dashboard.html")
 	err = tmpl.ExecuteTemplate(res, "dashboard.html", v)
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -286,8 +286,8 @@ func (s *server) loginPage(res http.ResponseWriter, req *http.Request) {
 		usersession, err := vSphereSessionGet(&userconfig)
 		if err != nil || usersession == nil {
 			// something went wrong or we could not authenticate
-			log.Warnf("User %s from %s failed to authenticated at %s", user, req.RemoteAddr, time.Now())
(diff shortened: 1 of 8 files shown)


anchal-agrawal:
Here are some mockups.

Login page:

login page

Dashboard (healthy):

normal homepage

cmd/vicadmin/server.go
```diff
@@ -266,7 +274,7 @@ func (s *server) loginPage(res http.ResponseWriter, req *http.Request) {
 		if err != nil || usersession == nil {
 			// something went wrong or we could not authenticate
 			log.Warnf("User %s from %s failed to authenticated at %s", user, req.RemoteAddr, time.Now())
-			http.Error(res, "Authentication failed due to incorrect credential(s)", 400)
+			http.Error(res, "Authentication failed due to incorrect credential(s)", http.StatusBadRequest)
```
anchal-agrawal: I changed this to `http.StatusBadRequest` solely to remove the hardcoded value. Would using `http.StatusUnauthorized` be more appropriate?

lib/vicadmin/validate.go
```diff
@@ -55,6 +54,9 @@ type Validator struct {
 	DockerPort       string
 	VCHStatus        template.HTML
 	VCHIssues        template.HTML
+	VCHReachable     bool
+	VCHUnreachable   bool
```
anchal-agrawal: `VCHUnreachable` is redundant. 

I tried using an `else` block [here](https://github.com/vmware/vic/pull/3106/files#diff-7372f4b2bbd435e1a9f6b78eeb5f04f0R105) to show the error string under the log bundle link, but the template couldn't be parsed. I'll modify the template logic to get rid of `VCHUnreachable` if we can.

anchal-agrawal:
@stuclem Once this is merged, please update `doc/user_doc/vic_admin/vicadmin_status_ref.md` to reflect the change from `Network Connectivity` to `Registry and Internet Connectivity`. We might need more changes to that doc to explain the error scenario in the picture above.

cmd/vicadmin/server.go
```diff
@@ -266,7 +275,7 @@ func (s *server) loginPage(res http.ResponseWriter, req *http.Request) {
 		if err != nil || usersession == nil {
 			// something went wrong or we could not authenticate
 			log.Warnf("User %s from %s failed to authenticated at %s", user, req.RemoteAddr, time.Now())
```
gigawhitlocks: "failed to authenticated" if you could fix that typo pls

lib/vicadmin/validate.go
gigawhitlocks: @sgairo can you help w/ this?

lib/vicadmin/validate.go
```diff
@@ -71,46 +73,64 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
+
+		v.LicenseStatus = BadStatus
+		v.LicenseIssues = template.HTML("")
 	} else {
-		v.FirewallStatus = BadStatus
-		for _, err := range firewallIssues {
-			v.FirewallIssues = template.HTML(fmt.Sprintf("%s<span class=\"error-message\">%s</span>\n", v.FirewallIssues, err))
+
+		v.VCHReachable = true
+		v.VCHUnreachable = false
```
gigawhitlocks: yeah this hurts, see if you can figure out how to use an `else` in the template as you mentioned before.. or even `if` and then `if not` somehow :\

anchal-agrawal: Agreed, it's painful. Working on it.

gigawhitlocks:
Minor nits, otherwise LGTM

lib/vicadmin/validate.go
anchal-agrawal: Fixed. Turns out the template doesn't like html comments between `{{if}}` and `{{else}}`.

cmd/vicadmin/server.go
```diff
@@ -314,16 +323,15 @@ func (s *server) loginPage(res http.ResponseWriter, req *http.Request) {
...
 	}
 
-	// Render login page (shows up on non-POST requests):
-	sess, err := client(&rootConfig)
-	if err != nil {
-		log.Errorf("Could not render login page due to vSphere connection error: %s", err.Error())
-		http.Error(res, genericErrorMessage, http.StatusInternalServerError)
-		return
+	// Render login page (shows up on non-POST requests)
+	hostName, _ := os.Hostname()
```
caglar10ur: do not ignore err here

cmd/vicadmin/server.go
caglar10ur: StatusUnauthorized

cmd/vicadmin/server.go
```diff
@@ -499,12 +507,41 @@ func (s *server) tailFiles(res http.ResponseWriter, req *http.Request, names []s
...
+			displayErrStr = "unable to obtain a vim client"
+		case session.UserPassLoginError:
+			displayErrStr = "unable to log in with username and password"
+		default:
+			displayErrStr = err.Error()
+		}
+	}
+
+	return displayErrStr
+}
```
caglar10ur: why do you need displayErrStr? Return directly within the switch statement eg; ```return fmt.Sprintf("SDK URL could not be parsed: %s", err.Err)```

cmd/vicadmin/server.go
```diff
@@ -499,12 +507,41 @@ func (s *server) tailFiles(res http.ResponseWriter, req *http.Request, names []s
...
 	ctx := context.Background()
 	sess, err := s.getSessionFromRequest(req)
 	v := vicadmin.NewValidator(ctx, &vchConfig, sess)
 
+	if sess == nil {
+		// We're unable to connect to vSphere, so display an error message
+		displayErrStr := deriveDisplayError(err)
+		v.VCHIssues = template.HTML(fmt.Sprintf(
+			"<span class=\"error-message\">%s Error: %s</span>\n",
+			v.Hostname+VCHUnreachableMsg, displayErrStr))
```
caglar10ur: year is 2016, I think it's safe to say that we can have longer lines now :) I suggest making this a single line

pkg/vsphere/session/errors.go
```diff
@@ -0,0 +1,47 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package session
+
+import "fmt"
```
caglar10ur: if they are session specific then you don't need to export them otherwise I suggest putting them in errors pkg

cmd/vicadmin/server.go
anchal-agrawal: I thought it would be cleaner to assign `displayErrStr` and return once at the end of the func, rather than putting several `return`s.

cmd/vicadmin/server.go
anchal-agrawal: Fixed.

cmd/vicadmin/server.go
anchal-agrawal: If the error is non-nil, `hostName` would be an empty string, which is what we'd want to display if we can't find the hostname. Please let me know if you have suggestions.

cmd/vicadmin/server.go
caglar10ur: IMHO - it's more confusing this way. You are hiding the fact that you are returning an empty string when err == nil

```
func deriveDisplayError(err error) string {
	if err != nil {
		switch err := err.(type) {
		case session.SDKURLError:
			return fmt.Sprintf("SDK URL could not be parsed: %s", err.Err)
 		case session.SoapClientError:
 			return "unable to obtain a vim client"
 		case session.UserPassLoginError:
 			return "unable to log in with username and password"
 		default:
 			return err.Error()
 		}
 	}
 
 	return ""
 }
```

anchal-agrawal: Sounds good, thanks! I'll make the change.

pkg/vsphere/session/errors.go
anchal-agrawal: We need exported errors to access them outside the `session` pkg in `cmd/vicadmin/server.go`. Since these errors are session-specific, I put them in the `session` pkg. Should we move them to `pkg/errors` anyway?

cmd/vicadmin/server.go
anchal-agrawal: I feel this improves readability by separating the html arg from the other args.

pkg/vsphere/session/errors.go
caglar10ur: Hmm missed the server.go part. Now sure, I guess they can stay and we can move it if another user shows up in the future

caglar10ur:
LGTM

cmd/vicadmin/server.go
anchal-agrawal: Thought about this some more. I'll change this so `hostName = "VCH"` when we get an error. It would be better to display "VCH Login" just "Login". Thanks @caglar10ur!

pkg/vsphere/session/errors.go
```diff
@@ -0,0 +1,47 @@
...
+
+// SDKURLError is returned when the soap SDK URL cannot be parsed
+type SDKURLError struct {
+	Service string
+	Err     error
```
anchal-agrawal: I've been running into issues with the way `fmt.Sprintf` handles anonymous `error` fields as arguments. I'll defer this change for now and look for the exact problem.

anchal-agrawal:
🏁

PR #3107, opened 2016-11-10
Long overdue adding rmi to regression

mhagen-vmware:
Long overdue turn on for this, still need to improve regression in general by quite a bit but we at least need to be testing rmi in our nightly.


matthewavery:
lgtm

anchal-agrawal:
LGTM

PR #3112, opened 2016-11-10
Use vmw_scsi target directly for disk attach [full ci]

### LINKED ISSUE #2340: "Failed to write to image store: [POST /storage/{store_name}][500] WriteImage default  &{Code:0xc4201fe0d8 Message:timeout waiting for layer to present as /dev/disk/by-path/pci-0000:03:00.0-scsi-0:0:0:0}"
Seen on build:
https://ci.vmware.run/vmware/vic/4804

[integration_logs_4804_9374cde28de68dffc53baf87abe87c2174f5ad05_1473959100.zip](https://github.com/vmware/vic/files/475199/integration_logs_4804_9374cde28de68dffc53baf87abe87c2174f5ad05_1473959100.zip)

fdawg4l:
Fixes #2340 

We were relying on systemd-udevd to create the appropriate device paths
in /dev/disk/by-path to mount attached devices to the VCH.  This change
uses the vmw_scsi device directly (by looking in
/sys/devices/pci/<address>) and calculating the target directory.

* Also added lsblk to log bundle
* Added operation logging where we were previously just using logrus

### DIFF
--- a/pkg/vsphere/disk/disk_manager.go
+++ b/pkg/vsphere/disk/disk_manager.go
@@ -20,8 +20,6 @@ import (
 	"os"
 	"sync"
 
-	log "github.com/Sirupsen/logrus"
-
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/pkg/errors"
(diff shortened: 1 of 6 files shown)


caglar10ur:
LGTM

pkg/vsphere/disk/disk_manager.go
```diff
@@ -108,10 +107,9 @@ func (m *Manager) CreateAndAttach(op trace.Operation, newDiskURI,
 		return nil, errors.Trace(err)
 	}
 
-	d.setAttached(devicePath)
-
-	if err := waitForPath(op, devicePath); err != nil {
-		op.Infof("waitForPath failed for %s with %s", newDiskURI, errors.ErrorStack(err))
+	blockDev, err := waitForDevice(op, devicePath)
```
matthewavery: It is still possible that this will be preempted by the switch-root that breaks the tether. I do not think we have removed the infinite loop there which spins waiting for devices to present. This could also help us determine if the vm is actually stuck in that state right?

matthewavery:
LGTM

fdawg4l:
.

PR #3113, opened 2016-11-10
Remove busybox container before rmi in integration test [skip ci]

anchal-agrawal:
Removes a busybox container in `6-04-Create-Basic` before running the regression tests, where we `rmi` busybox.


mhagen-vmware:
Lgtm

mhagen-vmware:
You don't really need to create an explicit var for the container name if you are just using it in two places right next to each other though 😁

chengwang86:
Lgtm

dougm:
lgtm

While you're in there, looks like we should be able to remove the 'sleep 10', or worst case change to `logs --follow`

mhagen-vmware:
I don't think we can remove it.  Still have outstanding truncation issue with short running containers AFAIK.

anchal-agrawal:
🏁

PR #3115, opened 2016-11-11
Always supply CN in cert, and add IP SANs for FQDN

hickeng:
Updates the server certificate generation:
1. always set a Common Name - docker and openssl are fine without, but chrome complains about
   corrupted certificate and refuses to display site when deployed with a static IP, not FQDN.
1. add IP SANs if FQDN is specified (allows for the harbor behaviour of redirecting to IP from FQDN)

Towards: #2197


dougm:
lgtm

caglar10ur:
LGTM

hickeng:
poke

PR #3128, opened 2016-11-11
Allow vicadmin to use TLS when personality does not

### LINKED ISSUE #3063: vic-admin --no-tls case
right now it crashes and requires following

```
diff --git a/cmd/vicadmin/server.go b/cmd/vicadmin/server.go
index 3176279..638d90c 100644
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -147,7 +147,7 @@ func (s *server) Authenticated(link string, handler func(http.ResponseWriter, *h
        authHandler := func(w http.ResponseWriter, r *http.Request) {
                websession, _ := s.uss.cookies.Get(r, sessionCookieKey) // ignore error because it is okay if it doesn't exist

-               if len(r.TLS.PeerCertificates) > 0 { // the user is authenticated by certificate at connection time
+               if r.TLS != nil && len(r.TLS.PeerCertificates) > 0 { // the user is authenticated by certificate at connection time
                        usersess := s.uss.Add(websession.ID, &rootConfig.Config)

                        timeNow, err := usersess.created.MarshalText()
```

applying that fix solves the crash but we show login page over http which is not secure, instead we should be showing a message saying that they can't use vic-admin cause --no-tls

gigawhitlocks:
When installing vicadmin with `--no-tls`, vicadmin was previously simply served w/o authentication or encryption on `http`. Now this is disabled and vicadmin always presents on `https`, using its own self-signed certificates if the VCH is installed with `--no-tls`.

Fixes #3063

### DIFF
--- a/cmd/vicadmin/vicadm.go
+++ b/cmd/vicadmin/vicadm.go
@@ -132,7 +132,7 @@ func init() {
 
 	extraconfig.Decode(src, &vchConfig)
 	if vchConfig.HostCertificate == nil {
-		log.Infoln("--no-tls is enabled")
+		log.Infoln("--no-tls is enabled on the personality")
 		rootConfig.serverCert = &ServerCertificate{}
 		rootConfig.serverCert.Cert, rootConfig.serverCert.Key, err = certificate.CreateSelfSigned(rootConfig.addr, []string{"VMware, Inc."}, 2048)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


andrewtchin:
lgtm

cmd/vicadmin/vicadm.go
```diff
@@ -123,6 +131,15 @@ func init() {
 	}
 
 	extraconfig.Decode(src, &vchConfig)
+	if vchConfig.HostCertificate == nil {
+		log.Infoln("--no-tls is enabled")
```
anchal-agrawal: If this line appears in the vicadmin logs, we can change this to `log.Infoln("--no-tls is enabled on the personality")` to be clearer.

gigawhitlocks: Definitely

anchal-agrawal:
LGTM

gigawhitlocks:
:recycle:

PR #3129, opened 2016-11-11
Adding gateway spec for vch networks that lets users specify routing destinations [full ci]

### LINKED ISSUE #3081: If management interface of the VCH is not Layer 2 adjacent to vSphere management endpoints VCH will not work.
Configuration of the management gateway for the VCH needs to include networks that should use that gateway.

Currently we have the options: 

```
--management-network 'network 2'
--management-network-gateway 192.168.2.1/24
--management-network-ip 192.168.2.10/24

```
for setting up a VCH. 

The problem is that the external (soon to be renamed public) interface should have the default route. Therefore, we need to allow the customer to specify the network(s) that need to route over the management network/interface of the VCH and through the gateway provided in --management-network-gateway. The way it works in .7.0 is that randomly one of the interfaces and gateways becomes the default route on the VCH. This means that if the external interface gateway is picked as the default route, then all vSphere management traffic that is not L2 adjacent to the VCH management interface will route out of the external interface and for all reasonable security configurations for customers will not reach the target.

**Proposal:**

Change --management-network-gateway 192.168.2.1/24 to --management-network-gateway 192.168.3.0/24,192.168.128.0/22:192.168.2.1/24. What this would do is cause the VCH to know that all of the network addresses in 192.168.3.0-255 and 192.168.128.0-192.168.131.255 can be reached by sending packets to 192.168.2.1. Which would result in vic-init adding the following routes into the VCH.

```
Destination     Gateway         Genmask
192.168.3.0    192.168.2.1     255.255.255.0
192.168.128.0  192.168.2.1    255.255.252.0
```

The default route would remain on the external interface. The specifier would allow 1 to many network spaces separated by commas with a colon and then the gateway with mask.

**Workaround for 0.7.0:**
Use vic-machine debug to turn on ssh access to the VCH.
Login to the VCH via ssh and ensure the default route is on the external interface.
Add routes as needed to the management network.

### LINKED ISSUE #3125: When client interface not shared with external, clients may not be able to connect.
When we fix #2802 we need to also add the same fix for the client interface and gateway that we did for the management interface in #3081.

hmahmood:
Fixes #3081 
Fixes #3125

### DIFF
--- a/lib/install/validate/network.go
+++ b/lib/install/validate/network.go
@@ -38,11 +38,13 @@ import (
 func (v *Validator) getEndpoint(ctx context.Context, conf *config.VirtualContainerHostConfigSpec, network data.NetworkConfig, epName, contNetName string, def bool, ns []net.IP) (*executor.NetworkEndpoint, error) {
 	defer trace.End(trace.Begin(""))
 	var gw net.IPNet
+	var dest []net.IPNet
 	var staticIP *net.IPNet
 
 	if !network.Empty() {
 		log.Debugf("Setting static IP for %q on port group %q", contNetName, network.Name)
 		gw = network.Gateway
+		dest = network.Destinations
 		staticIP = &network.IP
 	}
 
@@ -60,9 +62,10 @@ func (v *Validator) getEndpoint(ctx context.Context, conf *config.VirtualContain
 				Name: contNetName,
 				ID:   moid,
 			},
-			Default:     def,
-			Gateway:     gw,
-			Nameservers: ns,
+			Default:      def,
+			Destinations: dest,
+			Gateway:      gw,
+			Nameservers:  ns,
 		},
 		IP: staticIP,
 	}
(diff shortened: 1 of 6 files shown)


mlh78750:
Added comments inline.

Three things that I think need to be addressed:

 * The default route should only be set on the external (soon to be renamed public) interface.
 * We should require 1 or more destinations for the management and client network gateway configuration. 0 (== totally optional) routes makes no sense since it makes the gateway setting a no-op.
 * We should not accept routes for the external (soon to be public) interface since it will already have the default route and any other routes added will be redundant.

lib/tether/ops_linux.go
```diff
@@ -358,6 +358,27 @@ func linkAddrUpdate(old, new *net.IPNet, t Netlink, link netlink.Link) error {
...
+			Dst:       &d,
+			Gw:        gw.IP,
+		}
+
+		if err := t.RouteAdd(r); err != nil && !os.IsNotExist(err) {
+			log.Errorf("failed to add route for destination %s via gateway %s", d, gw.IP)
+		}
+	}
+
+	return updateDefaultRoute(t, link, endpoint)
```
hmahmood: Its called for all networks, but the first thing that function checks if its the "default" network. Effectively it will only do something for the external network since its the only one marked as default in the config.

lib/install/validate/network.go
```diff
@@ -169,6 +172,23 @@ func (v *Validator) network(ctx context.Context, input *data.Data, conf *config.
...
+	} {
+		if !ip.IsUnspecifiedIP(n.Gateway.IP) && len(n.Destinations) == 0 {
+			v.NoteIssue(fmt.Errorf("%s network gateway specified without at least one routing destination", nn))
+		}
+	}
+
+	// external network should not have any routing destinations specified
+	// if a gateway was specified
+	if !ip.IsUnspecifiedIP(input.ExternalNetwork.Gateway.IP) && len(input.ExternalNetwork.Destinations) > 0 {
+		v.NoteIssue(errors.New("external network must not have any routing destinations specified for gateway"))
```
mlh78750: Maybe change the error here to "external network has the default route and must not have ...

mlh78750:
Couple of text nits.

LGTM and good work getting this in so quickly.

lib/install/validate/network.go
```diff
@@ -169,6 +172,23 @@ func (v *Validator) network(ctx context.Context, input *data.Data, conf *config.
 	var e *executor.NetworkEndpoint
 	var err error
 
+	// client and management networks need to have at least one
+	// routing destination if gateway was specified
+	for nn, n := range map[string]*data.NetworkConfig{
```
vburenin: Please extract this map into a separate variable, it is so hard do read.

vburenin:
LGTM

hmahmood:
♻️

PR #3132, opened 2016-11-11
Add supplied parameter output to debug logging

### LINKED ISSUE #3037: Log the vic-machine command line used to install a VCH in the logs.
We commonly would like to see the command line options used by vic-machine to install the VCH. We should log this so it is in one of the logs that download from vic-machine. We should also log command line changes from things like vic-machine debug etc.

This is going to be very useful for quickly resolving customer found issues.

mhagen-vmware:
Fixes #3037 

When --debug is >1 you should see something like this:
```
DEBU[2016-11-11T13:28:14-06:00] Supplied install parameters: &{%!s(*common.Target=&{0xc42009d680 root 0xc4201b1c00 }) {%!s(int=1)} { virtual-container-host} {}    {  }  map[]  virtual-container-host { {<nil> <nil>} {<nil> <nil>}} {VM Network {<nil> <nil>} {<nil> <nil>}} { {<nil> <nil>} {<nil> <nil>}} [] map[] map[] map[] map[] %!s(int=0) %!s(int=0) %!s(*types.SharesInfo=<nil>) %!s(int=0) %!s(int=0) %!s(*types.SharesInfo=<nil>) 172.16.0.0/12 [] <nil> <nil> %!s(int=1) %!s(int=2048) 3m0s %!s(bool=true) %!s(bool=false) 8GB} 
```
In this case almost all parameters are nil, because I didn't supply them.  Of note:
1. target url and password are there but obfuscated with however we are doing that
2. the username is there in plain text


mlh78750:
I'm okay with the raw output if we're going to be able to easily understand what was passed in and reproduce.
I'd prefer this be done each time and be an INFO level log not hidden behind debug.
I think plaintext username is likely okay, but @hickeng has been talking with the security folks more and may have another opinion.

mlh78750:
LGTM

cmd/vic-machine/create/create.go
```diff
@@ -946,6 +946,8 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
 		return err
 	}
 
+        log.Infof("Supplied install parameters: %s", c.Data);
```
anchal-agrawal: The semicolon is not needed.

anchal-agrawal:
lgtm

mhagen-vmware:
.

mhagen-vmware:
Merging for now, we can revert later if @hickeng thinks there is any security issues.

hickeng:
@mhagen-vmware So long as we're not logging a supplied password I think we're okay from a security perspective.

I'm not sure this will give us what we need - you can already see in the output that most of the target parameter is the pointer values, not the argument value. If this is for serviceability it's worth spending a little extra effort on the code so that whomever is first line service can easily parse it.

mlh78750:
Longer term I'd like to just print the command line as it was entered with *** for usernames, passwords, and certs. I think it would require vic-machine passing argv over to the appliance to get logged and adding the ***'s.

hickeng:
@mlh78750 For the command line as entered I think pushing a PR to urfave/cli to add a log call is likely the best way to go.
The parameter parsing and default injection that happens in the validation means that the appliance doesn't have an easy way to recreate the actual command line. If we need it logged with the appliance we could upload a record to the applianceVM directory at the same time as the ISOs.

PR #3134, opened 2016-11-11
Fix license check in vicadmin on vC

### LINKED ISSUE #2305: Inconsistent license check reporting between vic-machine and vic-admin
When deploying to a nimbus testbed - from the installer:

```
INFO[2016-09-12T22:22:08Z] License check OK on hosts:
INFO[2016-09-12T22:22:08Z]   "/vcqaDC/host/cls/10.160.163.156"
INFO[2016-09-12T22:22:08Z]   "/vcqaDC/host/cls/10.160.185.119"
```

from vic-admin:

```
License
License does not meet minimum requirements to use VIC
```

I've not yet tracked down which is correct, but I am finding that while I can create containers I cannot get an active connection to them.

andrewtchin:
Fixes #2305 
Previously we were running the ESX check always instead of the vCenter check when appropriate.

### DIFF
--- a/lib/install/validate/validator.go
+++ b/lib/install/validate/validator.go
@@ -63,6 +63,7 @@ func CreateFromVCHConfig(ctx context.Context, vch *config.VirtualContainerHostCo
 	v := &Validator{}
 	v.Session = sess
 	v.Context = ctx
+	v.isVC = v.Session.IsVC()
 
 	return v, nil
 }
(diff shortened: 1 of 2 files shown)


sgairo:
LGTM

mhagen-vmware:
Lgtm

andrewtchin:
♻️

PR #3141, opened 2016-11-12
Improve UX for reuse of certificates [full ci]

### LINKED ISSUE #2197: Add user authentication to the personality

hickeng:
Adds a --cert-path option, used both for sourcing existing certs and
saving generated certs.

If existing certs are found, those will be used - this avoids the scenario
of redeploying a VCH and overwriting the generated (and potentially
distributed) certificates.
Various checks are performed to ensure certificate validity and attempts
to provide better diagnostics with cert errors.

Now uses the system root pool to validate server cert as well as the
generated CA when checking API liveliness.

Fixes #2197

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -40,6 +40,8 @@ import (
 	"github.com/vmware/vic/pkg/ip"
 	"github.com/vmware/vic/pkg/trace"
 
+	"crypto/x509"
+
 	"golang.org/x/net/context"
 )
 
@@ -48,6 +50,13 @@ const (
 	MaxVirtualMachineNameLen = 80
 	// Max permitted length of Virtual Switch name
 	MaxDisplayNameLen = 31
+
+	clientCert = "cert.pem"
+	clientKey  = "key.pem"
+	serverCert = "server-cert.pem"
+	serverKey  = "server-key.pem"
+	caCert     = "ca.pem"
+	caKey      = "ca-key.pem"
 )
 
 var EntireOptionHelpTemplate = `NAME:
@@ -71,8 +80,11 @@ OPTIONS:
 type Create struct {
 	*data.Data
 
-	cert       string
-	key        string
+	certPath   string
+	scert      string
+	skey       string
+	ccert      string
+	ckey       string
 	cacert     string
 	cakey      string
 	clientCert *tls.Certificate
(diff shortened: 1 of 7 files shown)


cmd/vic-machine/create/create.go
```diff
@@ -886,46 +1013,51 @@ func (c *Create) generateCertificates(ca bool) ([]byte, *certificate.KeyPair, er
...
-	if err = ckp.SaveCertificate(); err != nil {
-		log.Errorf("Failed to save client certificates: %s", err)
-		return nil, nil, err
-	}
+	if client {
+		log.Infof("Generating client certificate/key pair - private key in %s", c.ckey)
+		ckp := certificate.NewKeyPair(c.ccert, c.ckey, nil, nil)
+		err = ckp.CreateClientCertificate(c.cname, c.org, c.keySize, cakp)
+		if err != nil {
+			log.Errorf("Failed to generate server certificates: %s", err)
```
andrewtchin: should this be client certificates?

andrewtchin:
lgtm

cmd/vic-machine/create/create.go
```diff
@@ -777,81 +815,170 @@ func (c *Create) processProxies() error {
...
 	}
 
 	var keypair *certificate.KeyPair
-	if c.cert != "" && c.key != "" {
-		log.Infof("Loading certificate/key pair - private key in %s", c.key)
-		keypair = certificate.NewKeyPair(c.cert, c.key, nil, nil)
-
-		if err := keypair.LoadCertificate(); err != nil {
+	// default names
+	skey := fmt.Sprintf("./%s/%s", c.certPath, serverKey)
```
fdawg4l: Will this bale on $WINDOWS?  Might want to use `path.Join()` if it will.

fdawg4l:
LGTM

hickeng:
poke

hickeng:
poke

Issue #2357, opened 2016-09-16
Fix the LEMP test


sflxn:
The LEMP test have problems with VIC.  It need to be updated to actually work by removing the mapped volume.  Also, it cannot run until link/alias are working again.

sflxn:
depends on #2294

sflxn:
Unfortunately, attempting to fix this test has exposed another feature we need for docker-compose, volume inspect.

sflxn:
#2522

sflxn:
I think we need to backlog this issue.  There is no way to share files between the nginx and the php container with VIC right now.  We can tell users they need to roll their own image that contains both nginx and php together.

mdubya66:
Backlogged

chengwang86:
This is fixed by #3145 and #3153

PR #3145, opened 2016-11-13
fix #2954 and #2357 and bring Docker-Compose tests back [full ci]

### LINKED ISSUE #2954: Containers labels now broken, causing Compose to fail
Recent builds have complained about basic compose test failure.

Upon examining the issue, containers labels were broken.  I've identified the line causing the break but not yet sure why.

In a recent commit, I made a call to container.Refresh() in the portlayer rest server to update the network guest info data.  That somehow screwed up the labels.  The labels are actually there in the container's extraconfig when viewed with govc.

chengwang86:
Fixes #2954 #2357 

-  Changing the key of "Annotations" from "annotation" to "annotations" in lib/config/executor/container_vm.go
-  Bring Docker-Compose-Basic and Docker-Compose-LEMP test cases back

From "lib/apiserver/portlayer/models/container_create_config.go", I find this
 ```
Annotations map[string]string `json:"annotations,omitempty"`
```

It is possible that #2954 was incurred because we use "annotations" when creating the container, but then we use "annotation" when decoding the extraconfig and extracting the container labels.

### DIFF
--- a/lib/config/executor/container_vm.go
+++ b/lib/config/executor/container_vm.go
@@ -135,7 +135,7 @@ type ExecutorConfig struct {
 	LayerID string `vic:"0.1" scope:"read-only" key:"layerid"`
 
 	// Blob metadata for the caller
-	Annotations map[string]string `vic:"0.1" scope:"hidden" key:"annotation"`
+	Annotations map[string]string `vic:"0.1" scope:"hidden" key:"annotations"`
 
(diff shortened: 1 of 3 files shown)


caglar10ur:
LGTM

vburenin:
LGTM

chengwang86:
ping

PR #3162, opened 2016-11-14
[skip ci] Nightly global test status bug fix

### LINKED ISSUE #3161: Nightly global test status is reporting incorrect status
The nightly status is always getting reported as Passed even if a test fails. This is caused due to a bug in how we iterate the nightly status loop to fetch the global status.

rajanashok:
Fixing the bug that reports an incorrect nightly global status.
Fixes #3161

### DIFF
--- a/tests/nightly/nightly-kickoff.sh
+++ b/tests/nightly/nightly-kickoff.sh
@@ -75,10 +75,13 @@ for i in $nightly_list_var; do
     echo $count
 done
 
-for i in $nightlystatus; do
+for i in "${nightlystatus[@]}" 
+do
+    echo $i
     if [ $i = "Passed" ]
     then
     buildStatus=0
+    echo "Test Passed!"
     else
     buildStatus=1
     echo "Test failed, setting global test status to Failed!"


tests/nightly/nightly-kickoff.sh
```diff
@@ -75,10 +75,13 @@ for i in $nightly_list_var; do
     echo $count
 done
 
-for i in $nightlystatus; do
+for i in "${nightlystatus[@]}" 
+do
+    echo $i
     if [ $i = "Passed" ]
```
mhagen-vmware: I don't think this is still right, you need to check here if the global is ever false and do not set it to pass if it ever becomes false.

rajanashok: I've verified that this works with this following test code.

#!/bin/bash
nightlystatus=( Passed Failed Passed )
count=0

for i in "${nightlystatus[@]}" 
do
    echo $i
    if [ $i = "Passed" ]
    then
    buildStatus=0
    echo "Test Passed!"
    else
    buildStatus=1
    echo "Test failed, setting global test status to Failed!"
    break
    fi	
done

mhagen-vmware: oh, sorry I missed the break statement

mhagen-vmware:
lgtm

anchal-agrawal:
lgtm

rajanashok:
.

PR #3164, opened 2016-11-14
Changing external network name to public [full ci]

### LINKED ISSUE #3008: Some network nomenclature (around the notion of "external") perhaps needs revisiting
Currently using build 6507 (0.7.0 candidate). 

As I am trying to get a bit deeper into the VIC networking model I think I found some of the fundamental issues and misleading aspects that made some of the discussions with @mlh78750 so hard in the past few weeks. In retrospect I think there was a language mismatch confusion that isn't only due to documentation to be perfected (e.g.  https://github.com/vmware/vic/issues/3006 , https://github.com/vmware/vic/issues/3007) but also due to some nomenclature we are using inside the product (baked into the code) that I think we should consider changing. 

In particular we have a notion of (as far as I understand) an `external network` that is the network connected to the VCH end-point VM which is where services/containers get exposed via the `-p` option of a `docker run`. This is configured via the `vic-machine` parameter `--external-network`. 

In addition to the above, we have the notion of `container networks` (configured via the parameter `--container-network`). These are networks that containers can be attached directly to via the `--net` option of a `docker run`. 

Now the problem is that if `container networks` are configured as part of the `vic-machine create` command they show up inside the Docker context as `external` networks. 

Consider this example:
```
user1@ubuntu:~/vic$ ./vic-machine-linux create --name VCH1 ........ --bridge-network BridgePG1 --container-network PG2 --container-network PG3 --external-network "VM Network" ......
```
There is one external network (`VM Network`), one bridge network (`BridgePG1` and two container networks (`PG2` and `PG3`). 

However if I query the networks using the docker client I get: 

```
user1@ubuntu:~/vic$ docker -H 192.168.8.64:2375 network ls
NETWORK ID          NAME                DRIVER              SCOPE
96bbd64fe0fe        PG2                 external                                
83aac42374d0        PG3                 external                                
c3120e9089d3        bridge              bridge    
```

This is (IMO) creating a huge amount of confusion as people tend to give different meanings to what an `external network` actually is.  

My suggestion would be to consider changing the name of either the `--external-network` parameter or the `driver` we show in the `docker network ls`.

/cc @hmahmood @stuclem @hickeng

hmahmood:
Fixes #3008

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -99,9 +99,9 @@ type Create struct {
 	clientNetworkName         string
 	clientNetworkGateway      string
 	clientNetworkIP           string
-	externalNetworkName       string
-	externalNetworkGateway    string
-	externalNetworkIP         string
+	publicNetworkName         string
+	publicNetworkGateway      string
+	publicNetworkIP           string
 	managementNetworkName     string
 	managementNetworkGateway  string
 	managementNetworkIP       string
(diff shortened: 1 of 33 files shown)


andrewtchin:
lgtm

mlh78750:
Is there anything we need to do to make sure upgrades work? If I have .7 and I try to upgrade to .8 will stuff break due to this change?

LGTM if the answer is no. :)

hmahmood:
@mlh78750 good point. I don't think this will work for upgrades. More changes incoming.

hmahmood:
Closing for now

PR #3165, opened 2016-11-14
Switch to a whitelist logging to avoid possible security concerns

mhagen-vmware:
This is a follow up to the previous issue:
#3037 

Output now looks like:
```
[34mINFO[0m[2016-11-14T13:07:08-06:00] Supplied install parameters: --compute-resource {/ha-datacenter/host/localhost.localdomain/Resources VCH-0-8421} --image-store datastore1 --volumes map[default:datastore1/test] --bridge-network VCH-0-8421-bridge --client-network { [] {<nil> <nil>} {<nil> <nil>}} --external-network {VM Network [] {<nil> <nil>} {<nil> <nil>}}
```

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -980,7 +980,7 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


andrewtchin:
lgtm

mlh78750:
LGTM 

I'd like to add the networking options for proxy as well. But we can followup with that later. 

I'd actually just prefer we pass in the unparsed string from the original install command and log that on the VCH with perhaps some processing to hide usernames, passwords, and other credentials. But this is a better than not having it.

mhagen-vmware:
.

PR #3166, opened 2016-11-14
[skip ci] Avoid ssh prompt when provisioning esx

dougm:
Missed this when switching vib install to use ssh, as these options had been set in ~/.ssh/config


anchal-agrawal:
lgtm

dougm:
Actually, @caglar10ur pointed out the esx-ui vib is included with 6.0u2 and 6.5, I'll do another PR that just removes that.

PR #3168, opened 2016-11-14
Implement SSL Thmbprint Verifier for VIC UI Service

jooskim:
This PR replaces the all-accepting TrustManager and HostnameVerifier implemented in the original UI service code.

### DIFF
--- a/ui/vic-ui-service/src/main/java/com/vmware/utils/ssl/ThumbprintTrustManager.java
+++ b/ui/vic-ui-service/src/main/java/com/vmware/utils/ssl/ThumbprintTrustManager.java
@@ -1,0 +1,74 @@
+/* Copyright 2016 VMware, Inc. All rights reserved. -- VMware Confidential */
+
+package com.vmware.utils.ssl;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.cert.CertificateException;
+import java.util.HashSet;
+import java.util.Set;
+
+public class ThumbprintTrustManager implements javax.net.ssl.TrustManager, javax.net.ssl.X509TrustManager {
+
+   private static final Log _logger = LogFactory.getLog(ThumbprintTrustManager.class);
+   private static Set<String> _thumbprints = new HashSet<String>();
+
+   public static void setThumbprints(Set<String> thumbprints) {
+	   synchronized(ThumbprintTrustManager.class) {
+		   _thumbprints = thumbprints;
+	   }
+   }
+
+   @Override
+   public java.security.cert.X509Certificate[] getAcceptedIssuers() {
+      return null;
+   }
+
+   @Override
+   public void checkServerTrusted(java.security.cert.X509Certificate[] certs,
+                                  String authType) throws CertificateException {
+      for (java.security.cert.X509Certificate cert : certs) {
+         String thumbprint = getThumbprint(cert);
+         checkThumbprint(thumbprint);
+      }
+   }
+
+   @Override
+   public void checkClientTrusted(java.security.cert.X509Certificate[] certs, String authType) throws CertificateException {
+      return;
+   }
+
+   public static String getThumbprint(java.security.cert.X509Certificate cert) throws java.security.cert.CertificateException {
+      try {
+         MessageDigest md = MessageDigest.getInstance("SHA-1");
+         byte[] certBytes = cert.getEncoded();
+         byte[] bytes = md.digest(certBytes);
+
+         StringBuilder builder = new StringBuilder();
+         for (byte b : bytes) {
+            String hex = Integer.toHexString(0xFF & b);
+            if (hex.length() == 1) {
+               builder.append("0");
+            }
+            builder.append(hex);
+         }
+         return builder.toString().toLowerCase();
+      } catch (NoSuchAlgorithmException ex) {
+         return null;
+      }
+   }
+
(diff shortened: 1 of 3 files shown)


jooskim:
@mdubya66, I think we could close the Bugzilla issues once this PR gets reviewed and merged.

mdubya66:
agreed

mdubya66:
@hmahmood @hickeng can you give this a review

ui/vic-ui-service/src/main/java/com/vmware/vicui/VicUIServiceImpl.java
```diff
@@ -57,35 +61,30 @@ private static VimPortType initializeVimPort() {
...
+
+		javax.net.ssl.TrustManager[] tms = new javax.net.ssl.TrustManager[1];
+		javax.net.ssl.TrustManager tm = new ThumbprintTrustManager();
+		tms[0] = tm;
+		javax.net.ssl.SSLContext sc = null;
+
+		try {
+			sc = javax.net.ssl.SSLContext.getInstance("SSL");
+		} catch (NoSuchAlgorithmException e) {
+			_logger.info(e);
```
hmahmood: Is it OK to continue after this?

jooskim: @hmahmood I think it's ok to continue here unless you see something else that concerns you. It worked pretty solid for me.

hmahmood: Well for one thing is `sc` going to be non-null below if `NoSuchAlgorithmException` happens? Granted it is unlikely.

ui/vic-ui-service/src/main/java/com/vmware/vicui/VicUIServiceImpl.java
```diff
@@ -137,8 +136,17 @@ private ServerInfo getServerInfoObject(String serverGuid) {
 		return null;
 	}
 
+    private void setThumbprint(ServerInfo sinfo) {
+        String thumbprint = sinfo.thumbprint;
+        if (thumbprint != null) {
+            _thumbprints.add(thumbprint.replaceAll(":", "").toLowerCase());
+        }
+        ThumbprintTrustManager.setThumbprints(_thumbprints);
```
hmahmood: Instead of setting the whole collection, can we not have a function that just adds a new thumbprint to `ThumbprintTrustManager`?

ui/vic-ui-service/src/main/java/com/vmware/utils/ssl/ThumbprintTrustManager.java
```diff
@@ -0,0 +1,71 @@
...
+            }
+            builder.append(hex);
+         }
+         return builder.toString().toLowerCase();
+      } catch (NoSuchAlgorithmException ex) {
+         return null;
+      }
+   }
+
+   public static void checkThumbprint(String thumbprint) throws CertificateException {
```
hmahmood: Does this call need to be synchronized?

ui/vic-ui-service/src/main/java/com/vmware/vicui/VicUIServiceImpl.java
hmahmood: Also, how many thumbprints will this manage? Will there be more than one?

jooskim: I just checked java docs on `SSLContext.getInstance` and the code and it seems `sc` will remain null if `getInstance` fails. Here are some of the docs I referenced in case you need to verify it.

http://alvinalexander.com/java/jwarehouse/openjdk-8/jdk/src/share/classes/javax/net/ssl/SSLContext.java.shtml
http://grepcode.com/file/repository.grepcode.com/java/root/jdk/openjdk/6-b14/sun/security/jca/GetInstance.java#GetInstance.getInstance%28java.lang.String%2Cjava.lang.Class%2Cjava.lang.String%29

jooskim: For our case there will be only one thumbprint (one for VC), but `ThumbprintTrustManager` is made for general use where it could handle more than one certificate, as far as my understanding goes.

ui/vic-ui-service/src/main/java/com/vmware/utils/ssl/ThumbprintTrustManager.java
jooskim: Yes you're right. That makes sense to me

ui/vic-ui-service/src/main/java/com/vmware/utils/ssl/ThumbprintTrustManager.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+public class ThumbprintTrustManager implements javax.net.ssl.TrustManager, javax.net.ssl.X509TrustManager {
+
+   private static final Log _logger = LogFactory.getLog(ThumbprintTrustManager.class);
+   private static Set<String> _thumbprints = new HashSet<String>();
+
+   public static void setThumbprints(Set<String> thumbprints) {
+      _thumbprints = thumbprints;
```
hmahmood: This also needs to be synchronized.

hmahmood:
lgtm

jooskim:
thanks

mdubya66:
lgtm

jooskim:
♻️

PR #3169, opened 2016-11-14
fix panic in container_handler conversion func

### LINKED ISSUE #2985: Panic in portlayer: http: panic serving 192.168.31.143:42152: runtime error: invalid memory address or nil pointer dereference
Test failed in https://ci.vmware.run/vmware/vic/6483
We've seen this randomly fail before, but have been unable to determine root cause so far.
Test is disabled temporarily

cgtexmex:
Checks for existence of session id information in
execConfig prior to interrogating map

fixes #2985

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -384,7 +384,6 @@ func (handler *ContainersHandlersImpl) ContainerWaitHandler(params containers.Co
 
 	select {
 	case <-c.WaitForState(exec.StateStopped):
-		c.Refresh(context.Background())
 		containerInfo := convertContainerToContainerInfo(c.Info())
 
 		return containers.NewContainerWaitOK().WithPayload(containerInfo)
@@ -418,14 +417,6 @@ func convertContainerToContainerInfo(container *exec.ContainerInfo) *models.Cont
 	restart := int32(container.ExecConfig.Diagnostics.ResurrectionCount)
 	info.ContainerConfig.RestartCount = &restart
 
-	tty := container.ExecConfig.Sessions[ccid].Tty
-	info.ContainerConfig.Tty = &tty
-
-	attach := container.ExecConfig.Sessions[ccid].Attach
-	info.ContainerConfig.AttachStdin = &attach
-	info.ContainerConfig.AttachStdout = &attach
-	info.ContainerConfig.AttachStderr = &attach
-
 	info.ContainerConfig.StorageSize = &container.VMUnsharedDisk
 
 	if container.ExecConfig.Annotations != nil && len(container.ExecConfig.Annotations) > 0 {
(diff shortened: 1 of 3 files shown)


lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -436,28 +428,41 @@ func convertContainerToContainerInfo(container *exec.ContainerInfo) *models.Cont
 		}
 	}
 
-	path := container.ExecConfig.Sessions[ccid].Cmd.Path
-	info.ProcessConfig.ExecPath = &path
+	// in heavily loaded environments we were seeing a panic due a missing
+	// session id in execConfig -- this has only manifest itself in short lived containers
+	// that were initilized via run
+	if _, exists := container.ExecConfig.Sessions[ccid]; exists {
+		tty := container.ExecConfig.Sessions[ccid].Tty
```
hmahmood: Don't need to do `container.ExecConfig.Sessions[ccid]` again here; it has already been done on the line above.

cgtexmex: done

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -436,28 +428,41 @@ func convertContainerToContainerInfo(container *exec.ContainerInfo) *models.Cont
 		}
 	}
 
-	path := container.ExecConfig.Sessions[ccid].Cmd.Path
-	info.ProcessConfig.ExecPath = &path
+	// in heavily loaded environments we were seeing a panic due a missing
```
matthewavery: `...due to...` sorry :) just saw it and thought now was a good time.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -436,28 +428,41 @@ func convertContainerToContainerInfo(container *exec.ContainerInfo) *models.Cont
...
+	// session id in execConfig -- this has only manifest itself in short lived containers
```
matthewavery: `manifested`*

matthewavery:
lgtm

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -436,28 +428,41 @@ func convertContainerToContainerInfo(container *exec.ContainerInfo) *models.Cont
 		}
 	}
 
-	path := container.ExecConfig.Sessions[ccid].Cmd.Path
-	info.ProcessConfig.ExecPath = &path
+	// in heavily loaded environments we were seeing a panic due to a missing
+	// session id in execConfig -- this has only manifested itself in short lived containers
+	// that were initilized via run
```
caglar10ur: I think I didn't understand this comment. Do we know why session id is missing? It sounds like a corruption or race somewhere and if that's the case this is just a workaround

cgtexmex: No - I don't know the exact cause of the issue -- to your point I have only avoided the panic.  This happens in the CI intermittently and when the environment is apparently loaded / slow -- the test failed due to the panic, but would have failed because of network issues.

We can leave this change out and leave the issue open -- but reproduction is a challenge.  I was banking on the avoidance of a panic being the preferred approach.

hickeng: We can compromise by protecting against the panic if DebugLevel == 0.

Irrespective, we should log the actual contents - it's possible the entire extraconfig wasn't populated/unmarshalled correctly.

cgtexmex:
Updated with logging of the ExecConfig is sessionID is missing.  Additionally, introduced a minor refactoring to the eventCallback to avoid duplicating container refresh during the container wait path.

vburenin:
LGTM

hickeng:
LGTM - please open an issue to track down why we're getting nil session content.

mdubya66:
lgtm recycle

PR #3170, opened 2016-11-14
Add container op logging [full ci]

### LINKED ISSUE #1865: log which container is being operated on in trace
Currently the trace output doesn't include which container is being operated on. This makes it almost impossible to trace concurrent operations, and awkward to map even serial operations across components:

Portlayer:

```
time="2016-08-05T23:12:16Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/apiservers/portlayer/restapi/handlers.(*ContainersHandlersImpl).CommitHandler] Containers.CommitHandler" 
time="2016-08-05T23:12:16Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/portlayer/exec.(*Container).Commit] Committing handle" 
time="2016-08-05T23:12:16Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/portlayer/exec.(*Container).stop] Container.stop" 
time="2016-08-05T23:12:16Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/portlayer/exec.(*Container).stop] [916.358002ms] Container.stop" 
time="2016-08-05T23:12:17Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/portlayer/exec.(*Container).Commit] [1.275672261s] Committing handle" 
time="2016-08-05T23:12:17Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/apiservers/portlayer/restapi/handlers.(*ContainersHandlersImpl).CommitHandler] [1.276749265s] Containers.CommitHandler" 
```

Personality:

```
time="2016-08-05T23:09:37Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).containerCreate] [509.585876ms] Container.containerCreate" 
time="2016-08-05T23:09:37Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).ContainerCreate] [510.356025ms] ContainerCreate" 
time="2016-08-05T23:09:37Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).ContainerStart] ContainerStart" 
time="2016-08-05T23:09:47Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).ContainerStart] [10.576098821s] ContainerStart" 
time="2016-08-05T23:09:47Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).ContainerStop] ContainerStop" 
time="2016-08-05T23:10:34Z" level=debug msg="[ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).ContainerStop] [46.635692084s] ContainerStop" 
```

matthewavery:
This is an overhaul of our logging to allow us to reliably track concurrent calls on containers. It shifts us from using the normal logrus logger to using the `trace.Operation` object to assign operationIDs to call paths. Right now operationIDs are not reliably passed from the personality to the portlayer due to our usage of `context.Background()` which gives us a hollow context which does not actually store values across api boundaries. Both the personality server and the portlayer are affected by this change, and most function call paths have been modified to accept this operation object rather than the context object. Additionally any modifications made in the future on these functions should do their logging through the operation object rather than importing logrus for logging. 

This should also allow us to distinguish between two call paths on the same container with the same operation. 

Fixes #1865

### DIFF
--- a/lib/apiservers/engine/backends/backends.go
+++ b/lib/apiservers/engine/backends/backends.go
@@ -39,6 +39,7 @@ import (
 	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/lib/imagec"
 	"github.com/vmware/vic/pkg/errors"
+	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/sys"
 )
 
@@ -256,7 +257,8 @@ func InsecureRegistries() []string {
 
 // syncContainerCache runs once at startup to populate the container cache
 func syncContainerCache() error {
-	log.Debugf("Updating container cache")
+	op := trace.NewOperation(context.Background(), "")
+	defer trace.End(trace.Begin(op.SPrintf("Updating container cache")))
 
 	backend := NewContainerBackend()
 	client := backend.containerProxy.Client()
@@ -267,13 +269,13 @@ func syncContainerCache() error {
 		return errors.Errorf("Failed to retrieve container list from portlayer: %s", err)
 	}
 
-	log.Debugf("Found %d containers", len(containme.Payload))
+	op.Debugf("Found %d containers", len(containme.Payload))
 	cc := cache.ContainerCache()
 	var errs []string
 	for _, info := range containme.Payload {
 		container := ContainerInfoToVicContainer(*info)
 		cc.AddContainer(container)
-		if err = setPortMapping(info, backend, container); err != nil {
+		if err = setPortMapping(op, info, backend, container); err != nil {
 			errs = append(errs, err.Error())
 		}
 	}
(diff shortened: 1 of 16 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -1288,8 +1307,8 @@ func setPathFromImageConfig(config, imageConfig *containertypes.Config) {
 
 // validateCreateConfig() checks the parameters for ContainerCreate().
 // It may "fix up" the config param passed into ConntainerCreate() if needed.
```
sgairo: *ContainerCreate()

lib/portlayer/exec/base.go
```diff
@@ -292,22 +299,22 @@ func (c *containerBase) poweroff(ctx context.Context) error {
...
-					log.Warnf("power off %s task skipped (state was already %s)", c.ExecConfig.ID, terr.ExistingState)
+					op.Warnf("power off %s task skipped (state was already %s)", c.ExecConfig.ID, terr.ExistingState)
 					return nil
 				}
-				log.Warnf("invalid power state during power off: %s", terr.ExistingState)
+				op.Warnf("invalid power state during power off: %s", terr.ExistingState)
 
 			case *types.GenericVmConfigFault:
 
 				// Check if the poweroff task was canceled due to a concurrent guest shutdown
```
sgairo: "cancelled"

pkg/trace/operation.go
```diff
@@ -169,7 +177,7 @@ func FromContext(ctx context.Context) (Operation, error) {
 	case operation:
 		o.operation = val
 	default:
-		return Operation{}, fmt.Errorf("not an Operation")
+		return Operation{Context: context.Background()}, fmt.Errorf("not an Operation")
```
fdawg4l: I would rather this be empty since it's an error condition.   I also don't see an instance where it's necessary.

lib/apiservers/engine/backends/backends.go
```diff
@@ -256,7 +257,8 @@ func InsecureRegistries() []string {
 
 // syncContainerCache runs once at startup to populate the container cache
 func syncContainerCache() error {
-	log.Debugf("Updating container cache")
+	op := trace.NewOperation(context.Background(), "")
```
fdawg4l: You may want to embed the message you're adding on line 261 to this `New...` call.

Second, using `trace` `Begin`/`End` like this with an embedded operation call sounds interesting but is it redundant?  If it is useful, we may want to create a helper.  Popping up a level, what is your intention for this line (261) that you'd like to capture?

pkg/trace/operation.go
```diff
@@ -122,6 +122,14 @@ func (o *Operation) Errorf(format string, args ...interface{}) {
 	Logger.Errorf("%s: %s", o.header(), fmt.Sprintf(format, args...))
 }
 
+func (o *Operation) Warnf(format string, args ...interface{}) {
+	Logger.Warnf("%s: %s", o.header(), fmt.Sprintf(format, args...))
+}
+
+func (o *Operation) SPrintf(format string, args ...interface{}) string {
```
fdawg4l: Not sure how I feel about this one.  Smells useful, but I think we can hide whatever we're after in a helper since ultimately the string we're building is suppose to end up in the log.

matthewavery: Yeah, I found a use for it though when passing it to things like `trace.Begin`. If we ever want to make a string with the operation ID in it, this would be the simplest approach.

pkg/trace/operation.go
matthewavery: Now that I am rereading the manual though it also says that you should not store a context in a struct :/ hmm

mdubya66:
Is this PR still relevant?

matthewavery:
I believe it still is though @fdawg4l @hickeng @hmahmood may feel otherwise. I am open to closing it if we feel this is the wrong direction.

pkg/trace/operation.go
vburenin: @matthewavery Context is not a part of data, that is why it is complaining. Context is a thing that data belongs to. Attaching Context to the data leads to complicated structure re-usage later. It doesn't actually mean, it is the only truth here, there are different possible cases for storing context which are very rare.

lib/apiservers/engine/backends/backends.go
vburenin: Taking into account that this is a relatively common case when empty parameters are passed to the NewOperation, I would consider having a new constructor such as NewEmptyOperation that will do exactly same thing.

PR #3172, opened 2016-11-15
fix out of band docker start/stop time issue

### LINKED ISSUE #2451: OOB powerOn/Off result in inaccurate container start/stop time
If a container is powered on/off OOB then the container start or stop time is unavailable.  This impacts operations like docker ps and inspect.

poweredOn OOB ps output:

```
vic ps
CONTAINER ID        IMAGE               COMMAND             CREATED              STATUS              PORTS               NAMES
c43640cbf85f        busybox             "/bin/top"          About a minute ago   Running                                 furious_wing
```

Correct ps output:

```
vic ps
CONTAINER ID        IMAGE               COMMAND             CREATED              STATUS              PORTS               NAMES
f8ff06a26e17        busybox             "/bin/top"          About a minute ago   Up About a minute                       suspicious_mirzakhani
```

emlin:
Fixes #2451
Set start/stop time in tether, instead of portlayer, to make it more accurate, and works for out of band operations.
Add event handler to set stoptime, in case container is powered off directly.
New vc 6.5 vm powered on event.

Now out of band container start still does not work, cause we only have event for vm powered on. But at that time, session is still waiting for tether connection.


emlin:
.

emlin:
@cgtexmex @hickeng @caglar10ur 
This PR took a long time, for the sync up between tether and portlayer is problematic. Now all start/stop (includes OOB) worked well. Please help to have a look.
And hopefully I will have a few stable integration tests to cover this, but the point is that docker inspect can not be used, cause that API will sync with vsphere.

emlin:
@hickeng @caglar10ur Reverted the change in commit, so now it works for all other scenarios. For the start/stop, if run docker command, the start/stop time will be set in portlayer. 
Tether will set those value as well, but those will work only for OOB operations. If user power off VM directly, event handler will take care of that, and set stoptime.
The reason for this is because we're setting all guestinfo value set in handle.ExecConfig, no matter that's set for reconfiguration or just default value. And then we accumulate all operations in one vm.reconfigure. So in theory, we can only reconfigure vm after vm stop, and before vm start.
In this background, if I don't set start/stop time in portlayer, after vm is stopped, the default value of starttime/stoptime, that means the value before docker start/stop, will override the value set by tether, no matter we refresh or not.
At this time, set starttime/stoptime in portlayer is the workaround for this issue, but in long term, we'd better to have solution to encode only changed value in guestinfo.

lib/portlayer/exec/exec.go
```diff
@@ -102,7 +113,8 @@ func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSou
 		Config.EventManager.Subscribe(events.NewEventType(vsphere.VMEvent{}).Topic(), "registeredVMEvent", func(ie events.Event) {
 			registeredVMCallback(sess, ie)
 		})
-
+		// subscribe callbak to add container stopTime if not set by tether
```
caglar10ur: [typo][minor] callbak -> callback

caglar10ur:
LGTM

cgtexmex:
@emlin lets talk about this change -- you are adding another event call back that's duplicating much of the original event callback...which will result in additional refreshes (reads from vsphere for properties) that IMO aren't needed.

Also I question the value of this change for .8 -- the benefit is minimal (only OOB activity) and from my perspective the risk isn't justified.  I'm not slighting your effort...just felt it necessary to voice my opinion.  :-) 

cc: @hickeng @mlh78750 @mdubya66

emlin:
@cgtexmex for some of duplication of refresh, that might be. But I'm not that clear for those refresh in the vm events, so I didn't try to remove them. I assume the change should be removing the refresh in existing vm event handler for vm poweredOn and poweredOff events.

The new event handler is not only for OOB of operations. For the limitation of current commit mechanism, the result of poweredOff refresh will be override by portlayer commit, but it definitely add value for container start. It will sync back configuration set by tether after tether initialized.

lib/portlayer/exec/exec.go
```diff
@@ -171,6 +182,147 @@ func eventCallback(ie events.Event) {
...
+
+	if container.vm == nil {
+		log.Errorf("Container(%s) wait failed for vm is not found", container.ExecConfig.ID)
+		return false
+	}
+	log.Debugf("Waiting for container(%s) started", container.ExecConfig.ID)
+	for _, k := range extraconfig.CalculateKeys(container.ExecConfig, "Sessions.*.Detail.StartTime", "") {
+		var poweredoff bool
+		var detail string
+		waitFunc := func(pc []types.PropertyChange) bool {
```
hmahmood: Don't like this being copied here from vm.go.

If you could make this function block a var, that can be used by both vm.go and here, you can wrap it in another lambda that will just do the check for `detail != "0"` after the wrapped function (the new var) returns true.

lib/portlayer/exec/commit.go
```diff
@@ -170,7 +170,7 @@ func Commit(ctx context.Context, sess *session.Session, h *Handle, waitTime *int
 	// best effort update of container cache using committed state - this will not reflect the power on below, however
 	// this is primarily for updating ExtraConfig state.
 	if !creation {
-		defer c.RefreshFromHandle(ctx, h)
+		c.RefreshFromHandle(ctx, h)
```
hmahmood: This makes me nervous. Now we are refreshing before the end of the function. Why is this needed? And is refreshing at the end of the function not required anymore?

lib/portlayer/exec/exec.go
```diff
@@ -171,6 +182,147 @@ func eventCallback(ie events.Event) {
...
+	return true
+}
+
+func containerTimeCallback(ie events.Event) {
+	container := Containers.Container(ie.Reference())
+	if container == nil {
+		log.Debugf("Container(%s) is not found", ie.Reference())
+	}
+	switch ie.String() {
+	case events.ContainerStarted,
```
hmahmood: Is there an existing callback that handles these events? If yes, why not put this code there?

lib/portlayer/exec/commit.go
emlin: the refreshing is from handle to container base. we'll do refresh after container is started

lib/portlayer/exec/exec.go
emlin: no, not for container started

pkg/backoff/retry.go
```diff
@@ -0,0 +1,76 @@
...
+	DefaultInitialInterval     = 10 * time.Millisecond
+	DefaultRandomizationFactor = 0.5
+	DefaultMultiplier          = 10
+	DefaultMaxInterval         = 30 * time.Second
+	DefaultMaxElapsedTime      = 1 * time.Minute
+)
+
+// Retry retries the given function DefaultRetryCount times while sleeping some time between unsuccesfull attempts
+// if retryOnError return true, continue retry, otherwise, return error
+func Retry(operation func() error, retryOnError func(err error) bool) error {
```
emlin: This is a pretty straightforward change. And is used by other functions. I don't think simply copy/paste is good idea.

lib/portlayer/exec/exec.go
emlin: This method should merge with vm.WaitForKeyInExtraConfig, with customized condition check as input method. But this will impact the existing function.
As right now, this method is not used by anyone else, would you mind I push this refactor out of 0.8? Then I could focus on make this PR correct today.

lib/portlayer/exec/container.go
```diff
@@ -328,6 +335,8 @@ func (c *Container) start(ctx context.Context) error {
 	}
 
 	finalState = StateRunning
+	// get back configuration set by tether
+	c.Refresh(ctx)
```
cgtexmex: Why is there a refresh here and in the defer func()?

lib/portlayer/exec/exec.go
```diff
@@ -134,13 +146,26 @@ func eventCallback(ie events.Event) {
...
 				StateRunning,
-				StateStopped,
 				StateSuspended:
 
-				log.Debugf("Container(%s) state set to %s via event activity",
-					container.ExecConfig.ID, newState.String())
+				log.Debugf("Container(%s) state set to %s via event activity", container.ExecConfig.ID, newState.String())
 
 				container.SetState(newState)
 				if newState == StateStopped {
```
cgtexmex: not needed in case of StateStopping -- will be covered in StateStopped

lib/portlayer/exec/exec.go
```diff
@@ -171,6 +195,144 @@ func eventCallback(ie events.Event) {
...
+	}
+	go func() {
+		switch ie.String() {
+		case events.ContainerPoweredOn:
+			// container powered on event received, still need to wait container startTime set by tether, to sync between portlayer and vmx
+			// we'll do this in a go routine to avoid blocking
+			if !waitForContainerStarted(container) {
+				return
+			}
+			fallthrough
```
cgtexmex: Using `fallthrough` will always execute the next case, so during an OOB powerOn we'll refresh in the eventCallback and here..so duplicate calls for the same containerVM increasing burden on VC/ESXi

lib/portlayer/exec/container.go
emlin: oops, forgot to remove it

lib/portlayer/exec/exec.go
emlin: good catch.

lib/portlayer/exec/exec.go
emlin: This is the intentionally fall through. We need to refresh after the vm is powered on OOB, and then wait for at most 3 minutes. If we got container starttime set, that means the container is started.
Which is similar to docker start, which will wait for Started instead of starttime there, because docker start reset container started attribute before poweron VM. I need to use starttime for OOB poweron, here.
After that, should sync up with vsphere, for tether already configuration.

And we cannot skip the refresh immediately after powered on event received, because at that time, there are some attributes set, and if in any chance we failed to wait container starttime, we'll not refresh again.

lib/portlayer/exec/exec.go
```diff
@@ -171,6 +192,144 @@ func eventCallback(ie events.Event) {
...
+func waitForContainerStarted(container *Container) bool {
+	ctx, cancel := context.WithTimeout(context.Background(), propertyCollectorTimeout)
+	defer cancel()
+
+	if container.vm == nil {
+		log.Errorf("Container(%s) wait failed for vm is not found", container.ExecConfig.ID)
+		return false
+	}
+	log.Debugf("Waiting for container(%s) started", container.ExecConfig.ID)
+	for _, k := range extraconfig.CalculateKeys(container.ExecConfig, "Sessions.*.Detail.StartTime", "") {
```
hickeng: Having this loop on the outside means that we're calling vSphere once per session - not significant now given numbers of sessions, but will become more significant in the future.

mdubya66:
we pulled this from v0.8.0.

caglar10ur:
what is our plan with this PR? If we are not going to merge it as it is then would it be possible to split the retry pkg from this pr and submit it as another so that I can reuse it?

mdubya66:
We will not be merging this.

emlin:
@caglar10ur I'll do it. But when do you need this function? Still in .8?

Issue #3017, opened 2016-11-04
'echo test | docker run -i busybox cat' - doesn't succeed


vburenin:
Actual behavior

```
➜  vic git:(master) ✗ echo 'test' | docker run -i busybox
Unable to find image 'busybox:latest' locally
Pulling from library/busybox
56bec22e3559: Pull complete
a3ed95caeb02: Pull complete
Digest: sha256:98a0bd48d22ff96ca23bfda2fe1cf72034ea803bd79e64a5a5f274aca0f9c51c
Status: Downloaded newer image for library/busybox:latest
Error response from daemon: Server error from portlayer: ContainerWaitHandler(187537f9e81de0478c796eb81ecb7e70f0c80d762212854d5150dde5dcfba50a) Error: context deadline exceeded
```

Expected behavior

```
~ » echo test | docker run -i busybox cat
Unable to find image 'busybox:latest' locally
latest: Pulling from library/busybox
56bec22e3559: Pull complete
Digest: sha256:29f5d56d12684887bdfa50dcd29fc31eea4aaf4ad3bec43daf19026a7ce69912
Status: Downloaded newer image for busybox:latest
test
```

vburenin:
[container-logs.zip](https://github.com/vmware/vic/files/572125/container-logs.zip)

mlh78750:
cc @caglar10ur related to #2660

caglar10ur:
dup of #2660

caglar10ur:
re-opening this fixing #2660 won't fix this

mdubya66:
why do we think this is critical for v0.8?

caglar10ur:
This lets users to pipe things into the container, could be helpful for scripting interactions

```echo "caglar" | docker run busybox wc -c``` for example will return 7. Not sure whether it is critical for 0.8.

It heavily depends on #2660 and while testing, I realized fixing that won't fix this (This requires additional synchronization between reading from stdin and launching the process)

My personal suggestion would be marking this "good to have"

mlh78750:
If it doesn't make it to 0.8 we need to put this in known issues. This is something that will likely bite folks that are "kicking the tires" but not much else.

mdubya66:
it's not going to make it. tagging it for release notes.

corrieb:
This shouldn't be low priority IMHO. This is a compatibility issue that should be addressed. Verified that we still don't handle input being piped to a container. Changing to medium.

mdubya66:
[...]

PR #3185, opened 2016-11-15
Fix docker run/run -i and run -t truncation

### LINKED ISSUE #2660: docker run -itd busybox dmesg hangs and truncates
End's mid log message:
[    0.114184] TCP bind hash table entries: 16384 (order: 6, 262144 bytes)
[    0.114198] TCP: Hash tables configured (established 16384 bind 16384)
[    0.114209] UDP hash table entries: 1024 (order: 3, 32768 bytes)
[    0.114212] UDP-L

Docker, prints all and returns.

Attaching to the container just hangs, and doing a log on the container just prints the same output.  This also happens when directly attaching with run -it.

This is running on our new hardware VC cluster.

caglar10ur:
We now control Std{in|out|err}Pipe's directly and copy between them and dio

This PR also;

- Adds OpenStdin bool to the SessionConfig to indicate whether we should be opening stdin or not
- Sets the serial lines speed to 115200.
- Various cleanups

Fixes #2660


cmd/tether/attach_linux.go
```diff
@@ -33,6 +33,48 @@ import (
...
+
+	// copy it as the future
+	future := current.termios
+
+	// unset 9600 bps
+	future.Cflag &^= syscall.B9600
+	// set them to 115200 bps
+	future.Cflag |= syscall.B115200
+	future.Ispeed = syscall.B115200
+	future.Ospeed = syscall.B115200
```
vburenin: Have you tested this speed? I tried to change speed of serial ports and didn't notice any changes. Direct write/read into serial port device handles around 60KB/s

lib/tether/tether.go
```diff
@@ -442,24 +445,26 @@ func (t *tether) Register(name string, extension Extension) {
...
 	if session.wait != nil {
+		log.Debugf("Waiting on session.wait")
 		session.wait.Wait()
+		log.Debugf("Wait on session.wait completed")
 	}
 
-	// reader must be closed before calling wait, and this must interrupt the underlying Read
-	// or Wait will not return.
+	log.Debugf("Calling wait on cmd")
+	log.Errorf("Wait returned %s", session.Cmd.Wait())
```
vburenin: is it an error?

lib/tether/tether_linux.go
```diff
@@ -35,6 +34,10 @@ const (
 	//https://github.com/golang/go/blob/master/src/syscall/zerrors_linux_arm64.go#L919
 	SetChildSubreaper = 0x24
 	pidFilePath       = "var/run"
+
+	// in sync with lib/apiservers/portlayer/handlers/interaction_handler.go
+	// 115200 bps is 14.4 KB/s so use that
+	ioCopyBufferSize = 14 * 1024
```
vburenin: Real speed is faster, it is around 60KB.

cmd/tether/attach_linux.go
caglar10ur: This is just setting the correct value on the linux side, I don't think there is a speed limit on the ESXi side. By the way how did you test it?

lib/tether/tether.go
caglar10ur: Ahh, leftover. Taking care of

lib/tether/tether_linux.go
caglar10ur: see ^ above

cmd/tether/attach_linux.go
vburenin: there were two ways.
1. I measured transfer speed between two points by implementing my own read/write wrapper - it is not precise, however I was getting speeds like 30-50KB/s for terminal sessions while running midnight commander or vmware/powercli - these are doing massive data updates on screen.
2. dd if=/dev/zere of=/.tether/<serial_port>

lib/tether/tether.go
```diff
@@ -339,6 +340,8 @@ func (t *tether) processSessions() error {
 					err:  t.launch(session),
 				}
 			}(session)
+
+			session.Unlock()
```
hmahmood: You have this inside the `if {...}` clause only. If this clause is skipped, you will deadlock at the start of the loop when you try to lock again.

lib/tether/tether.go
```diff
@@ -442,24 +445,33 @@ func (t *tether) Register(name string, extension Extension) {
...
+	}
 
 	// close down the outputs
-	session.Outwriter.Close()
+	log.Debugf("Calling close on writers")
+	if err := session.Outwriter.Close(); err != nil {
+		log.Warnf("Close for Outwriter returned %s", err)
+	}
 	session.Errwriter.Close()
+	if err := session.Errwriter.Close(); err != nil {
```
hmahmood: Calling `session.Errwriter.Close()` twice?

lib/tether/tether_linux.go
```diff
@@ -213,29 +216,104 @@ func lookPath(file string, env []string, dir string) (string, error) {
...
+
+	if session.OpenStdin {
 		go func() {
-			_, gerr := io.Copy(session.Pty, session.Reader)
-			log.Debugf("PTY stdin copy: %s", gerr)
+			_, gerr := io.CopyBuffer(session.StdinPipe, session.Reader, make([]byte, ioCopyBufferSize))
+			log.Debugf("Reader stdin returned: %s", gerr)
+
+			if gerr == nil {
+				if cerr := session.StdinPipe.Close(); cerr != nil {
```
hmahmood: Do you need to close this? Documentation for Cmd.StdinPipe() stays it will closed when Cmd.Wait() returns.

lib/tether/tether.go
caglar10ur: ah good catch! Fixed

lib/tether/tether.go
caglar10ur: Fixed

lib/tether/tether_linux.go
caglar10ur: that's for the other case (if the command being run will not exit until standard input is closed, the caller must close the pipe.)

hmahmood: Is that the case here? i.e. we have to close this here for the command to exit? I thought we called Cmd.Wait() somewhere in the tether.

lib/portlayer/network/network.go
```diff
@@ -20,6 +20,8 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 
+	"golang.org/x/net/context"
```
hmahmood: Are we moving back to this from just `"context"`

caglar10ur: hmm nope, this was the result of goimports but while at it I'll move them to context

lib/tether/tether_linux.go
caglar10ur: We do call it in the childReaper but for that the command need to exit, tee for example won't exit unless you close the stdin

hmahmood:
lgtm

vburenin:
LGTM

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1242,7 +1245,9 @@ func containerConfigFromContainerInfo(vc *viccontainer.VicContainer, info *model
 		container.Tty = *info.ContainerConfig.Tty // Attach standard streams to a tty, including stdin if it is not closed.
 	}
 	// They are not coming from PL so set them to true unconditionally
```
hickeng: comment is now incorrect

lib/tether/tether.go
```diff
@@ -442,24 +445,33 @@ func (t *tether) Register(name string, extension Extension) {
...
-	session.Lock()
-	defer session.Unlock()
+	log.Debugf("Waiting on session.wait")
+	session.wait.Wait()
+	log.Debugf("Wait on session.wait completed")
 
-	if session.wait != nil {
-		session.wait.Wait()
+	log.Debugf("Calling wait on cmd")
+	if err := session.Cmd.Wait(); err != nil {
```
hickeng: I assume that we no longer need to close the Reader first?

lib/tether/tether.go
```diff
@@ -442,24 +445,33 @@ func (t *tether) Register(name string, extension Extension) {
 func (t *tether) handleSessionExit(session *SessionConfig) {
```
hickeng: comment that the caller needs to hold the session lock

lib/tether/tether_linux.go
```diff
@@ -213,29 +216,104 @@ func lookPath(file string, env []string, dir string) (string, error) {
...
+	}
 
-		// TODO: do we need to ensure all reads have completed before calling Wait on the process?
-		// it frees up all resources - does that mean it frees the output buffers?
-		go func() {
-			_, gerr := io.Copy(session.Outwriter, session.Pty)
-			log.Debugf("PTY stdout copy: %s", gerr)
+	session.wait.Add(1)
+	go func() {
+		_, gerr := io.CopyBuffer(session.Outwriter, session.Pty, make([]byte, ioCopyBufferSize))
```
hickeng: Most performant option would be to implement ReadFrom in MutliWriter.

lib/tether/tether_linux.go
```diff
@@ -213,29 +216,104 @@ func lookPath(file string, env []string, dir string) (string, error) {
...
+	}
+
+	if session.OpenStdin {
 		go func() {
-			_, gerr := io.Copy(session.Pty, session.Reader)
-			log.Debugf("PTY stdin copy: %s", gerr)
+			_, gerr := io.CopyBuffer(session.StdinPipe, session.Reader, make([]byte, ioCopyBufferSize))
+			log.Debugf("Reader stdin returned: %s", gerr)
+
+			if gerr == nil {
```
hickeng: Do we need to close this if the err is non-nil as well?

lib/tether/tether_linux.go
```diff
@@ -213,29 +216,104 @@ func lookPath(file string, env []string, dir string) (string, error) {
...
+	// Add 2 for Std{out|err}
+	session.wait.Add(2)
+	go func() {
+		_, gerr := io.CopyBuffer(session.Outwriter, session.StdoutPipe, make([]byte, ioCopyBufferSize))
+		log.Debugf("Writer goroutine for stdout returned: %s", gerr)
+
+		if session.StdinPipe != nil {
+			log.Debugf("(stdout): Writing zero byte to stdin pipe")
+			n, werr := session.StdinPipe.Write([]byte{})
+			if n == 0 && werr != nil && werr.Error() == "write |1: bad file descriptor" {
```
hickeng: What is the intent here? I would have thought that if Write is returning bad file descriptor, then the stdin Copy routine would have exited and we could close there.
If we need to close here, why are we not closing in all conditions, and only on bad file descriptor error?

tests/manual-test-cases/Group17-ManualTTY/17-1-TTY-Tests.robot
```diff
@@ -32,4 +32,28 @@ Docker run -it command that doesn't stop
     ${out}=  Run  docker ${params} logs ${containerID}
     Should Contain  ${out}  Mem:
     Should Contain  ${out}  CPU:
-    Should Contain  ${out}  Load average:
\ No newline at end of file
+    Should Contain  ${out}  Load average:
+
+Docker run with -i
+    ${rc}  ${output}=  Run And Return Rc And Output  docker ${params} run -i busybox dmesg
```
hickeng: can we use the following or similar - relying on dmesg output stability make break us in the future:
`run -i busybox /bin/ash -c "dmesg;echo complete"`

cmd/vic-machine/create/create.go
```diff
@@ -42,7 +42,7 @@ import (
 
 	"crypto/x509"
 
-	"golang.org/x/net/context"
+	"context"
```
anchal-agrawal: Should we move `context` to the first group of imports?

cmd/vic-machine/debug/debug.go
```diff
@@ -29,7 +30,7 @@ import (
 	"github.com/vmware/vic/pkg/version"
 	"github.com/vmware/vic/pkg/vsphere/vm"
 
-	"golang.org/x/net/context"
+	"context"
```
anchal-agrawal: ditto

lib/apiservers/engine/backends/container_proxy.go
caglar10ur: still true for the next line so moved there

lib/tether/tether.go
caglar10ur: nope

lib/tether/tether.go
caglar10ur: done

lib/tether/tether_linux.go
caglar10ur: err means StdinPipe is closed so we should be fine there

tests/manual-test-cases/Group17-ManualTTY/17-1-TTY-Tests.robot
caglar10ur: done

cmd/vic-machine/create/create.go
caglar10ur: done

lib/tether/tether_linux.go
caglar10ur: This is a left over from pipe work (#3017) shouldn't affect this.

caglar10ur:
♻️

caglar10ur:
[squashed the commits as I'm still not sure whether github squash & merge works or not]

caglar10ur:
♻️

PR #3191, opened 2016-11-16
Implement DeleteBranch in storage image cache

fdawg4l:
This allows deletion of a branch of images given a leaf node.  It
prevents deletion of nodes with order > 1 and deletion of nodes which
are otherwise attached or referenced by a tag (via the keep list).

This isn't called yet but will in be the default deletion mechanism from the docker personality.  That change will come in shortly.

* Also use an RWMutex for the index pkg


pkg/index/index.go
```diff
@@ -52,7 +52,7 @@ func (n *node) addChild(child *node) {
 type Index struct {
 	root        *node
 	lookupTable map[string]*node
-	m           sync.Mutex
+	m           sync.RWMutex
```
matthewavery: Mutex hat? This mutex should go at the top of the struct.

matthewavery:
lgtm - other than minor comments looks good

pkg/index/index.go
fdawg4l: Really?  Why?

lib/portlayer/storage/image_cache.go
```diff
@@ -366,6 +370,72 @@ func (c *NameLookupCache) DeleteImage(op trace.Operation, image *Image) error {
...
+	)
+
+	// map of images to keep
+	keep := make(map[url.URL]int)
+	for _, elem := range keepNodes {
+		keep[*elem] = 0
+	}
+
+	// Check if the error is actually an error.  If we deleted something,
+	checkErr := func(err error, deleted []*Image) ([]*Image, error) {
```
fdawg4l: I'd rather leave it where it is in case someone is tempted to use it for unrelated things.  I refuse to use a `goto` in `go` (and that's what I'd do if this was `C`), so this is close enough.

pkg/index/index.go
matthewavery: https://dmitri.shuralyov.com/idiomatic-go#mutex-hat

caglar10ur: I personally prefer mutex hat too :)

caglar10ur:
lgtm

fdawg4l:
.

pkg/index/index.go
fdawg4l: Just saw this.  Interesting.  For readability.  I like it!   I'll fix this in my next PR.

PR #3198, opened 2016-11-16
Fix empty session ID bug

gigawhitlocks:
While working on #3083 I noticed that the websession.ID field that was documented to be unique & provided for each cookie-based web session, turned out to be _a big stinkin' lie_ and isn't used by the type of cookie store being used. This results in the key being empty, and since we assume that the key would be unique, and in fact it's always the same, we likely cannot have more than one session at a time (in the current version on `master`) and this also may have been causing the session leakage in #3083 (still investigating/testing that, but this PR fixes a bug (or bugs) regardless, it just may be an unreported bug).

This change uses the key used to an actual UUID since @hmahmood  pointed out that we have google/uuid vendored (and I did not see this despite looking for it :facepalm:)

With a little added logging code (not included in this change) I was able to manually test two concurrent sessions from two separate browsers & did not witness the problems caused by the lies in the documentation of the `websession` object.

Egg on my face.

[Also, this bit adds some rudimentary caching to the vSphere lookup call.](https://github.com/vmware/vic/compare/master...gigawhitlocks:vicadmin-sessions?expand=1#diff-abef5e03992dd89faa301fb9709e9b78R79) This should help address some of the leaky session problem, as we won't create a new one every single time a user does a vSphere-dependent action.

### DIFF
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -554,9 +554,6 @@ func (s *server) index(res http.ResponseWriter, req *http.Request) {
 	defer trace.End(trace.Begin(""))
 	ctx := context.Background()
 	sess, err := s.getSessionFromRequest(ctx, req)
-	if err != nil {
-		panic(err.Error())
-	}
 	v := vicadmin.NewValidator(ctx, &vchConfig, sess)
 
 	if sess == nil {
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -113,7 +113,7 @@ func (s *server) listen() error {
 			}
 			c.ClientAuth = tls.VerifyClientCertIfGiven
 		} else {
-			log.Warnf("No certificate authorities found for certificate-based authentication. This may be intentional, however, authentication is disabled")
+			log.Warnf("No certificate authorities found for certificate-based authentication. This may be intentional, however, certificate-based authentication is disabled")
 		}
 
 		return &tls.Config{
@@ -171,6 +171,7 @@ func (s *server) Authenticated(link string, handler func(http.ResponseWriter, *h
 
 		if len(r.TLS.PeerCertificates) > 0 { // the user is authenticated by certificate at connection time
 			log.Infof("Authenticated connection via client certificate with serial %s from %s", r.TLS.PeerCertificates[0].SerialNumber, r.RemoteAddr)
+			key := uuid.New().String()
 			usersess := s.uss.Add(key, &rootConfig.Config)
 
 			timeNow, err := usersess.created.MarshalText()
@@ -183,7 +184,7 @@ func (s *server) Authenticated(link string, handler func(http.ResponseWriter, *h
 			}
 
 			websession.Values[sessionCreationTimeKey] = string(timeNow)
-			websession.Values[sessionKey], err = uuid.New().String()
+			websession.Values[sessionKey] = key
 
 			remoteAddr := strings.SplitN(r.RemoteAddr, ":", 2)
 			if len(remoteAddr) != 2 { // TODO: ctrl+f RemoteAddr and move this routine to helper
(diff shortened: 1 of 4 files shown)


cmd/vicadmin/server.go
```diff
@@ -170,7 +170,9 @@ func (s *server) Authenticated(link string, handler func(http.ResponseWriter, *h
 
 		if len(r.TLS.PeerCertificates) > 0 { // the user is authenticated by certificate at connection time
 			log.Infof("Authenticated connection via client certificate with serial %s from %s", r.TLS.PeerCertificates[0].SerialNumber, r.RemoteAddr)
-			usersess := s.uss.Add(websession.ID, &rootConfig.Config)
+			// this key just needs to be unique, so the time + connecting IP & port will be unique unless multiverse theory is accurate and suddenly apparent
```
matthewavery: This gave me a good laugh, but the multiverse bit is likely not relevant :P

anchal-agrawal:
lgtm

cmd/vicadmin/server.go
gigawhitlocks: shhh ok

cmd/vicadmin/server.go
```diff
@@ -550,7 +553,10 @@ func deriveErrorMessage(err error) string {
 func (s *server) index(res http.ResponseWriter, req *http.Request) {
 	defer trace.End(trace.Begin(""))
 	ctx := context.Background()
-	sess, err := s.getSessionFromRequest(req)
+	sess, err := s.getSessionFromRequest(ctx, req)
+	if err != nil {
+		panic(err.Error())
```
hickeng: Is this handled and turned into a somewhat friendly error? If not is there a recovery option other than panic we can use?

gigawhitlocks: hahaha omg I didn't mean to commit that

cmd/vicadmin/vicadm.go
```diff
@@ -420,16 +420,23 @@ func vSphereSessionGet(sessconfig *session.Config) (*session.Session, error) {
...
 
-func (s *server) getSessionFromRequest(r *http.Request) (*session.Session, error) {
+func (s *server) getSessionFromRequest(ctx context.Context, r *http.Request) (*session.Session, error) {
 	sessionData, err := s.uss.cookies.Get(r, sessionCookieKey)
 	if err != nil {
 		return nil, err
 	}
 
-	c, err := s.uss.VSphere(sessionData.ID)
+	c, err := s.uss.VSphere(ctx, sessionData.Values[sessionKey].(string))
```
gigawhitlocks: yeah.. I don't know what that cast will do if it's empty.. it really shouldn't ever be empty but, well, that's how I got myself into this situation

cmd/vicadmin/server.go
gigawhitlocks: And actually -- @anchal-agrawal 's change handles the `nil` case implicitly so we don't actually even need error checking here because the rest of the function will Do the Right Thing regardless.

cmd/vicadmin/vicadm.go
```diff
@@ -420,16 +420,25 @@ func vSphereSessionGet(sessconfig *session.Config) (*session.Session, error) {
...
 
-func (s *server) getSessionFromRequest(r *http.Request) (*session.Session, error) {
+func (s *server) getSessionFromRequest(ctx context.Context, r *http.Request) (*session.Session, error) {
 	sessionData, err := s.uss.cookies.Get(r, sessionCookieKey)
 	if err != nil {
 		return nil, err
 	}
-
-	c, err := s.uss.VSphere(sessionData.ID)
+	if sessionData.Values[sessionKey] == nil {
```
hmahmood: you can collapse this with the line below:
```go
if d, ok := sessionData.Values[sessionKey]; !ok {...} else { // use d }
```

gigawhitlocks: Cool, doing this. Also omitting `else` because `if !ok` will `return`

hmahmood:
lgtm

gigawhitlocks:
:recycle:

PR #3202, opened 2016-11-16
Make sure vic-machine is logged even in case of failure

mhagen-vmware:
We should not check for success until after logging the output.


chengwang86:
lgtm

rajanashok:
lgtm

mhagen-vmware:
.

PR #3210, opened 2016-11-17
Connectivity diagnostic with VC/ESXi from appliance 

### LINKED ISSUE #3066: vic-machine should test connectivity and name resolution on install of VCH and provide clear feedback and fail install if it doesn't work.
See: https://github.com/vmware/vic/issues/3031#issuecomment-259255064

If customer uses FQDN for install but it is only resolvable by the client (and possibly VC) but the VCH doesn't get configured with the same resolver then we will fail in very non-obvious ways. We should have vic-machine test the ability to resolve any passed in FQDN as well as get a list of hosts from the VC cluster/resource pool and if they are FQDN and not IP's make sure we can resolve them as well. 

Example, in #3031 was a cluster of one host with local storage. When installed it looks like the VC names the host with a FQDN but the VCH cannot resolve it.

vic-machine should test these conditions and fail with a clear error if name resolution on the VCH doesn't work.

vburenin:
```
INFO[2016-11-17T09:21:25-08:00] Waiting for IP information
INFO[2016-11-17T09:21:31-08:00] Waiting for major appliance components to launch

INFO[2016-11-17T09:21:31-08:00] Checking connectivity with the target vCenter/ESXi
INFO[2016-11-17T09:21:31-08:00] VC/ESXi API Ping Test: https://10.0.1.152 is reacheable and responds to ICMP ping request
INFO[2016-11-17T09:21:31-08:00] VC/ESXi API Test: https://10.0.1.152 API responds as expected

INFO[2016-11-17T09:21:34-08:00] Initialization of appliance successful
INFO[2016-11-17T09:21:34-08:00]
INFO[2016-11-17T09:21:34-08:00] VCH Admin Portal:
INFO[2016-11-17T09:21:34-08:00] https://10.0.1.156:2378
INFO[2016-11-17T09:21:34-08:00]
INFO[2016-11-17T09:21:34-08:00] Published ports can be reached at:
INFO[2016-11-17T09:21:34-08:00] 10.0.1.156
INFO[2016-11-17T09:21:34-08:00]
INFO[2016-11-17T09:21:34-08:00] Docker environment variables:
INFO[2016-11-17T09:21:34-08:00] DOCKER_HOST=10.0.1.156:2375
INFO[2016-11-17T09:21:34-08:00]
INFO[2016-11-17T09:21:34-08:00] Environment saved in vic-docker/vic-docker.env
INFO[2016-11-17T09:21:34-08:00]
INFO[2016-11-17T09:21:34-08:00] Connect to docker:
INFO[2016-11-17T09:21:34-08:00] docker -H 10.0.1.156:2375 info
INFO[2016-11-17T09:21:34-08:00] Installer completed successfully
```

```
INFO[2016-11-17T09:19:08-08:00] Waiting for IP information
INFO[2016-11-17T09:19:13-08:00] Waiting for major appliance components to launch
INFO[2016-11-17T09:19:13-08:00] Checking connectivity with the target vCenter/ESXi
ERRO[2016-11-17T09:19:14-08:00] VC/ESXi API Ping Test: https://brokenesx is unknown host name
INFO[2016-11-17T09:19:14-08:00] Collecting ha-host hostd.log
ERRO[2016-11-17T09:19:14-08:00] --------------------
ERRO[2016-11-17T09:19:14-08:00] vic-machine-linux failed: Access to VC/ESXi failed a ping test
```

Fixes #3066


cgtexmex:
It would be nice if the error message `vic-machine-linux failed: Access to VC/ESXi failed a ping test` indicated that the issue is between the appliance and vc/ESXi and not some where else...potentially even listing the source and target.

vburenin:
@cgtexmex Good point! I would really appreciate if people could take a look at error texts to rephrase them in a more human friendly way.

vburenin:
@caglar10ur golang.org/x/net/icmp is a new vendored dependency. I can create an issue to ICMP when vendor is unlocked.

vburenin:
@caglar10ur https://github.com/vmware/vic/issues/3219

vendor/github.com/vburenin/nsync/cwaitgroup.go
```diff
@@ -0,0 +1,82 @@
...
+import "sync"
+
+// ControlWaitGroup is a combination of the WaitGroup and Semaphore primitives to run goroutines.
+// WaitGroup is used to insure all tasks are complete, Semaphore server a lock job to limit
+// a number of concurrently running goroutines. Go is well known for its capabilities to run
+// tens of thousands of goroutines, however, it is also very easy to exceed available system resources
+// such as connections, opened files, etc. Thus, ControlWaitGroup fits well to limit the number
+// of running goroutines.
+type ControlWaitGroup struct {
+	sem     *Semaphore
```
fdawg4l: Where is this defined?

vendor/github.com/vburenin/nsync/cwaitgroup.go
```diff
@@ -0,0 +1,82 @@
...
+// dynamic so it only make sense to use it for logging purposes.
+func (cwg *ControlWaitGroup) Working() int {
+	return cwg.sem.Value()
+}
+
+// Waiting is a number of the currently scheduled, but not running goroutines. This value is highly
+// dynamic so it only make sense to use it for logging purposes.
+func (cwg *ControlWaitGroup) Waiting() int {
+	cwg.mu.Lock()
+	w := cwg.waiting
```
fdawg4l: Also using a buffered channel means you don't need `mu`.

vburenin: oh, it's accidentally sneaked in. I will remove this file.

caglar10ur:
using icmp pkg produces way less code than this diag pkg (and much easier to wortk with IMHO) so I'm still not sure why we prefer that. From that link above

```
// PrivilegedPing sends a ipv4.ICMPTypeEcho message to the given IP and returns true if it responds with a ipv4.ICMPTypeEchoReply
// Terminates if an error occurs
func PrivilegedPing(ip net.IP, seq int) bool {
	c, err := icmp.ListenPacket("ip4:icmp", "0.0.0.0")
	if err != nil {
		GuestRPCErrorf("Unable to listen: %s", err)
		os.Exit(ERR)
	}
	defer c.Close()

	// craft the icmp message
	wm := icmp.Message{
		Type: ipv4.ICMPTypeEcho,
		Code: 0,
		Body: &icmp.Echo{
			ID:   os.Getpid() & 0xffff,
			Seq:  1 << uint(seq),
			Data: []byte("KNOCK-IF-YOU-CAN-HEAR-ME"),
		},
	}

	// marshal it to write buffer
	wb, err := wm.Marshal(nil)
	if err != nil {
		GuestRPCErrorf("Unable to marshall the icmp message: %s", err)
		os.Exit(ERR)
	}

	dst := &net.IPAddr{IP: ip}
	// write it to wire
	if n, err := c.WriteTo(wb, dst); err != nil {
		GuestRPCErrorf("Unable to write: %s", err)
		os.Exit(ERR)
	} else if n != len(wb) {
		GuestRPCErrorf("Unable to write: got %v; want %v", n, len(wb))
		os.Exit(ERR)
	}

	// create read buffer
	rb := make([]byte, 1500)

	// Set read deadling to 3 sec.
	if err := c.SetReadDeadline(time.Now().Add(3 * time.Second)); err != nil {
		GuestRPCErrorf("Unable to set deadline: %s", err)
		os.Exit(ERR)
	}

	// read it from wire to read buffer
	n, _, err := c.ReadFrom(rb)
	if err != nil {
		GuestRPCErrorf("Unable to read: %s", err)
		os.Exit(ERR)
	}

	// parse the icmp message
	rm, err := icmp.ParseMessage(iana.ProtocolICMP, rb[:n])
	if err != nil {
		GuestRPCErrorf("Unable to parse: %s", err)
		os.Exit(ERR)
	}

	// type assert to see whether we have a response
	switch rm.Type {
	case ipv4.ICMPTypeEchoReply:
		return true
	default:
		return false
	}
}
```

but I'm fine with this if others are OK with it so I'll let them to decide

vburenin:
@caglar10ur it is just PING for IPv4. I will need to take care about name resolution, IPv6, etc. I agreed that this needs to be done, but not today. Issue has been filed, so I will take care about this after this PR for the next release.

caglar10ur:
we don't support anything other than ipv4 at this time and we have net.LookupAddr for name resolution, just saying...

pkg/diag/diag.go
```diff
@@ -0,0 +1,229 @@
...
+
+	"github.com/vmware/vic/pkg/trace"
+)
+
+// StatusCodeFatalThreshold defines a threshold after which all codes can be treated as fatal.
+const StatusCodeFatalThreshold = 64
+
+const (
+	// PingStatusOk Host name was resolved and all pings went through.
+	PingStatusOk = 0
```
hmahmood: I am not sure we need such detailed information about the ping output. How about just the output and exit code from the command, and the client can figure out what went wrong? In our case, we just need to know if the exit code != 0 to report bad connectivity.

cmd/vic-machine/create/create.go
```diff
@@ -1160,9 +1157,66 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
...
+	vch, err := executor.NewVCHFromComputePath(c.Data.ComputeResourcePath, c.Data.DisplayName, validator)
+	if err != nil {
+		executor.CollectDiagnosticLogs()
+		log.Errorf("Failed to get Virtual Container Host %s", c.Data.DisplayName)
+		log.Error(err)
+		return errors.New("Running diagnostics failed.")
+	}
+
+	// Checking if VC/ESXi can respond to ping request.
+	const pingTestTxt = "VC/ESXi API Ping Test:"
```
vburenin: well, I am on the other side about it ;)

pkg/diag/diag.go
vburenin: Giving more detailed diagnostics is better I think.

vburenin:
@caglar10ur This code example is not the good one. It doesn't take into account so many things: It doesn't check reponse ID, doesn't check sequence, doesn't track sequences that are being sent and received, it doesn't take into account possible intermitent delays and that packets may be delivered in different order, etc. I am working on my own code for this - but it is already a large one. Definitely not for this PR.

net.LookupAddr -> net.LookupIP

cmd/vic-machine/create/create.go
```diff
@@ -1160,9 +1157,66 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
 		return err
 	}
 
+	// vic-init will try to reach out to the vCenter/ESXi host.
+	log.Info("Checking connectivity with the target vCenter/ESXi")
```
hickeng: This message is unclear:
`Checking VCH connectivity with vSphere target`

cmd/vic-machine/create/create.go
```diff
@@ -1160,9 +1157,66 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
...
+	if err != nil {
+		executor.CollectDiagnosticLogs()
+		log.Errorf("Failed to get Virtual Container Host %s", c.Data.DisplayName)
+		log.Error(err)
+		return errors.New("Running diagnostics failed.")
+	}
+
+	// Checking if VC/ESXi can respond to ping request.
+	const pingTestTxt = "VC/ESXi API Ping Test:"
+	cd, err := executor.CheckVCPingFromAppliance(ctx, vch, vchConfig.Target)
```
hickeng: Don't put a ping test in here - there are plenty of people who will be blocking ICMP packets.
The only thing we care about is whether the endpointVM can talk to vSphere API - just the HEAD request will suffice.

Ping is solely for diagnostics in the failure case.

lib/install/management/dispatcher.go
```diff
@@ -218,6 +220,65 @@ func (d *Dispatcher) CollectDiagnosticLogs() {
 	}
 }
 
+// opManager creates a new operation manager.
```
hickeng: This comment is useless - please describe what an opManager is/does/why if you're going to have a comment here.

hickeng:
@vburenin 
This issue is about providing a means to determine:
a. can the target be resolved?
b. can we get an HTTP response from the target?

Additional diagnostics to aid in troubleshooting don't belong in vic-machine but in vic-admin.
If we wish to be able to report health status via vic-machine in general, that's a completely different PR.

I'd much prefer we add a structured mechanism to allow the components to report status, rather than add narrow checks like this into vic-machine. Done properly this can be leveraged to report health via vSphere alerts and in vic-admine as well as in all the other vic-machine operations.
Until we have a structured mechanism my request are:
1. drop/icebox this PR in it's current form
2. modify port-layer to update the vchconfig "status" field for the session with an appropriate (concise) error message if connection to vsphere fails (and potentially other fatal paths), and another message e.g. "initialization successful", when the portlayer is fully up
3. modify ensureApplianceInitializes to recognize the success message.
This allows for reporting a MUCH broader range of errors, with the errors determined by inline product code rather than sideline diagnostics.

pkg/diag/diag.go
```diff
@@ -0,0 +1,229 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
```
hickeng: This should be under `pkg/vsphere` - it's specifically checking the vsphere API

pkg/diag/diag.go
```diff
@@ -0,0 +1,229 @@
...
+		return "API test returned unknown code"
+	}
+}
+
+var statsRE = *regexp.MustCompile("\\d+ \\w{6,8} transmitted, (\\d+) ")
+
+// CheckPing runs ping to check if vSphere target is resolvable and pingable.
+func CheckPing(hn string) int {
+	cmd := exec.Command("ping", "-c", "4", "-W", "3", "-i", "0.1", hn)
+	return runPing(cmd.CombinedOutput())
```
hickeng: This is likely to fail intermittently as it's in a race with the childReaper.

vburenin: hm. Any suggestion on that?

vburenin: @hickeng Does vic-machine run childReaper?

vburenin: nm, it is inside vic-init

vburenin:
@hickeng I reduced scope of vSphere API test to focus purely on responses from client.Get
The rest will can do next time.

pkg/vsphere/diag/diag.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	return errorCode
+}
+
+func queryAPI(op trace.Operation, getter func(string) (*http.Response, error), apiURL string) int {
+	resp, err := getter(apiURL)
+	if err != nil {
+		errTxt := err.Error()
+		op.Errorf("Query error: %s", err)
+		if strings.Contains(errTxt, "no such host") {
```
hmahmood: Is it possible to actually check for errno's here instead of comparing error strings?

vburenin: Nope. I wish it was the case.

hmahmood:
lgtm

andrewtchin:
lgtm

PR #3214, opened 2016-11-17
Renaming "external" network to "public" [full ci]

### LINKED ISSUE #3008: Some network nomenclature (around the notion of "external") perhaps needs revisiting

### LINKED ISSUE #3180: Upgrade is broken
When running `vic-machine-linux upgrade` on a 0.7.0 install, I get this:
```
hmahmood@ubuntu:~/go/src/github.com/vmware/vic$ bin/vic-machine-linux upgrade --target 172.16.127.133 --user root --password password --thumbprint 4E:CA:0A:7C:E6:28:C1:B7:7E:B1:62:2C:C4:BA:D3:FA:C1:2F:F6:80
INFO[2016-11-15T11:56:29-06:00] ### Upgrading VCH ####
INFO[2016-11-15T11:56:29-06:00]
INFO[2016-11-15T11:56:29-06:00] VCH ID: VirtualMachine:48
ERRO[2016-11-15T11:56:29-06:00] --------------------
ERRO[2016-11-15T11:56:29-06:00] target is not set
ERRO[2016-11-15T11:56:29-06:00] target username is not set
ERRO[2016-11-15T11:56:29-06:00] target token is not set
ERRO[2016-11-15T11:56:29-06:00] Failed to migrate Virtual Container Host configuration virtual-container-host
ERRO[2016-11-15T11:56:29-06:00] validation of configuration failed
ERRO[2016-11-15T11:56:29-06:00] --------------------
ERRO[2016-11-15T11:56:29-06:00] vic-machine-linux failed: upgrade failed
```
The error messages here were introduced as part of https://github.com/vmware/vic/commit/e66e2220fb72b7606f213f3846b35965bd16fb2d

### LINKED ISSUE #3207: vic-machine upgrade hangs at deleting upgrade snapshot and locks the ESX server
Test installs a recent VCH build, creates a container, then attempts to upgrade to the latest VCH release.

INFO[2016-11-17T10:26:17-06:00] ### Upgrading VCH ####                       
INFO[2016-11-17T10:26:17-06:00]                                              
INFO[2016-11-17T10:26:17-06:00] VCH ID: VirtualMachine:125                   
INFO[2016-11-17T10:26:17-06:00] Uploading images for container               
INFO[2016-11-17T10:26:17-06:00] 	"/home/anon/vic/bin/bootstrap.iso"          
INFO[2016-11-17T10:26:17-06:00] 	"/home/anon/vic/bin/appliance.iso"          
INFO[2016-11-17T10:26:22-06:00] Creating snapshot upgrade for 6964           
INFO[2016-11-17T10:26:32-06:00] created snapshot VirtualMachineSnapshot:125-snapshot-1 
INFO[2016-11-17T10:26:32-06:00] Switching appliance iso to [datastore1] virtual-container-host/V0.7.0-6964-7B2D9AC-appliance.iso 
INFO[2016-11-17T10:26:32-06:00] Setting VM configuration                     
INFO[2016-11-17T10:26:33-06:00] Waiting for IP information                   
INFO[2016-11-17T10:26:33-06:00] Waiting for major appliance components to launch 
INFO[2016-11-17T10:26:46-06:00] Deleting upgrade snapshot "upgrade for 6964" 

While hung in this state, I can no longer login to my ESX either.  Says password incorrect, rebooting the ESX recovers, restarting the management agents from within ESX admin console fixes it as well - temporarily.  After a few minutes, I could not login again.

hmahmood:
Fixes #3008

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -208,8 +208,8 @@ func (c *Create) Flags() []cli.Flag {
 
 		// public
 		cli.StringFlag{
-			Name:        "public-network, en",
-			Value:       "",
+			Name:        "public-network, pn",
+			Value:       "VM Network",
 			Usage:       "The public network port group name (port forwarding and default route). Defaults to 'VM Network' and DHCP -- see advanced help (-x)",
 			Destination: &c.publicNetworkName,
 		},
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -110,9 +110,9 @@ type Create struct {
 	clientNetworkName         string
 	clientNetworkGateway      string
 	clientNetworkIP           string
-	externalNetworkName       string
-	externalNetworkGateway    string
-	externalNetworkIP         string
+	publicNetworkName         string
+	publicNetworkGateway      string
+	publicNetworkIP           string
 	managementNetworkName     string
 	managementNetworkGateway  string
 	managementNetworkIP       string
@@ -206,25 +206,25 @@ func (c *Create) Flags() []cli.Flag {
 			Hidden:      true,
 		},
 
-		// external
+		// public
 		cli.StringFlag{
-			Name:        "external-network, en",
-			Value:       "VM Network",
-			Usage:       "The external network port group name (port forwarding and default route). Defaults to 'VM Network' and DHCP -- see advanced help (-x)",
-			Destination: &c.externalNetworkName,
+			Name:        "public-network, en",
+			Value:       "",
+			Usage:       "The public network port group name (port forwarding and default route). Defaults to 'VM Network' and DHCP -- see advanced help (-x)",
+			Destination: &c.publicNetworkName,
 		},
 		cli.StringFlag{
-			Name:        "external-network-gateway",
+			Name:        "public-network-gateway",
 			Value:       "",
-			Usage:       "Gateway for the VCH on the external network, e.g. 10.0.0.1/24",
-			Destination: &c.externalNetworkGateway,
+			Usage:       "Gateway for the VCH on the public network, e.g. 10.0.0.1/24",
+			Destination: &c.publicNetworkGateway,
 			Hidden:      true,
 		},
 		cli.StringFlag{
-			Name:        "external-network-ip",
+			Name:        "public-network-ip",
 			Value:       "",
-			Usage:       "IP address for the VCH on the external network, e.g. 10.0.1.2/24",
-			Destination: &c.externalNetworkIP,
+			Usage:       "IP address for the VCH on the public network, e.g. 10.0.1.2/24",
+			Destination: &c.publicNetworkIP,
 			Hidden:      true,
 		},
 
@@ -254,7 +254,7 @@ func (c *Create) Flags() []cli.Flag {
 		cli.StringSliceFlag{
 			Name:   "dns-server",
 			Value:  &c.dns,
-			Usage:  "DNS server for the client, external, and management networks. Defaults to 8.8.8.8 and 8.8.4.4 when VCH uses static IP",
+			Usage:  "DNS server for the client, public, and management networks. Defaults to 8.8.8.8 and 8.8.4.4 when VCH uses static IP",
 			Hidden: true,
 		},
 
@@ -502,8 +502,8 @@ func (c *Create) processParams() error {
 		return err
 	}
 
-	if err := c.processNetwork(&c.Data.ExternalNetwork, "external", c.externalNetworkName,
-		c.externalNetworkIP, c.externalNetworkGateway); err != nil {
+	if err := c.processNetwork(&c.Data.PublicNetwork, "public", c.publicNetworkName,
+		c.publicNetworkIP, c.publicNetworkGateway); err != nil {
 		return err
 	}
 
@@ -784,7 +784,7 @@ func (c *Create) processNetwork(network *data.NetworkConfig, netName, pgName, st
 	return fmt.Errorf("Invalid %s network address: %s does not resolve to a gateway compatible IP", netName, staticIP)
 }
 
-// processDNSServers parses DNS servers used for client, external, mgmt networks
+// processDNSServers parses DNS servers used for client, public, mgmt networks
 func (c *Create) processDNSServers() error {
 	for _, d := range c.dns {
 		s := net.ParseIP(d)
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -81,7 +81,7 @@ const (
 )
 
 var (
-	externalIfaceName = "external"
+	publicIfaceName = "public"
 
 	defaultScope struct {
 		sync.Mutex
@@ -100,17 +100,17 @@ func init() {
 	portMapper = portmap.NewPortMapper()
 	containerByPort = make(map[string]string)
 
-	l, err := netlink.LinkByName(externalIfaceName)
+	l, err := netlink.LinkByName(publicIfaceName)
 	if l == nil {
-		l, err = netlink.LinkByAlias(externalIfaceName)
+		l, err = netlink.LinkByAlias(publicIfaceName)
 		if err != nil {
-			log.Errorf("interface %s not found", externalIfaceName)
+			log.Errorf("interface %s not found", publicIfaceName)
 			return
 		}
 	}
 
 	// don't use interface alias for iptables rules
-	externalIfaceName = l.Attrs().Name
+	publicIfaceName = l.Attrs().Name
 
 	// seed the random number generator
 	rand.Seed(time.Now().UTC().UnixNano())
@@ -747,7 +747,7 @@ func MapPorts(hostconfig *containertypes.HostConfig, endpoint *models.EndpointCo
 	cbpLock.Lock()
 	defer cbpLock.Unlock()
 	for _, p := range portMap {
-		if err = portMapper.MapPort(nil, p.intHostPort, p.portProto.Proto(), containerIP.String(), p.portProto.Int(), externalIfaceName, bridgeIfaceName); err != nil {
+		if err = portMapper.MapPort(nil, p.intHostPort, p.portProto.Proto(), containerIP.String(), p.portProto.Int(), publicIfaceName, bridgeIfaceName); err != nil {
 			return err
 		}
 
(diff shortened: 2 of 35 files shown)


emlin:
I understand data migration of our current metadata is a messy, and no clean way to do that in this short time. But I still want to raise my concern for the upgrade. 
We fixed the security issue in recent build, but that breaks upgrade. This upgrade fix works for others, but still not for the security. That means after upgrade to 0.8.0, user will still has 0.7.0 or 0.6.0 security issue. Was that accepted from production point of view? If that's not good, is it possible to postpone upgrade support after GA?

hmahmood:
@emlin which security issue are your referring to?

caglar10ur:
LGTM

emlin:
I didn't go through all security issues @hickeng and @dougm did, but one thing is the change in guestinfo. Did you remove the old keys has sensitive information in guestinfo? I remember there are some changes recently, but I cannot recall if there are after 0.7.0

lib/apiservers/engine/backends/container.go
```diff
@@ -1500,11 +1500,11 @@ func copyConfigOverrides(vc *viccontainer.VicContainer, config types.ContainerCr
 	vc.HostConfig = config.HostConfig
 }
 
-func externalIPv4Addrs() ([]netlink.Addr, error) {
-	l, err := netlink.LinkByName(externalIfaceName)
+func publicIPv4Addrs() ([]netlink.Addr, error) {
+	l, err := netlink.LinkByName(publicIfaceName)
 	if err != nil {
 		return nil, fmt.Errorf("Could not look up link from client interface name %s due to error %s",
```
hickeng: This message talks about client interface, not external - was this an error in the original code?

hmahmood: Yes it is left over from old code. I'll change it.

lib/config/executor/network_interface.go
```diff
@@ -60,17 +60,26 @@ type ContainerNetwork struct {
 
 	// The network scope the IP belongs to.
 	// The IP address is the default gateway
-	Gateway net.IPNet `vic:"0.1" scope:"read-write" key:"gateway"`
+	Gateway net.IPNet `vic:"0.1" scope:"read-only" key:"gateway2"`
```
emlin: This is a replacement of old gateway attribute. It might worth a version 0.2, or equal to 0.8 release

lib/config/executor/network_interface.go
```diff
@@ -60,17 +60,26 @@ type ContainerNetwork struct {
...
-	Nameservers []net.IP `vic:"0.1" scope:"read-write" key:"dns"`
+	Nameservers []net.IP `vic:"0.1" scope:"read-only" key:"dns2"`
 
 	// The IP ranges for this network
 	Pools []ip.Range `vic:"0.1" scope:"read-only" key:"pools"`
 
 	// set of network wide links and aliases for this container on this network
 	Aliases []string `vic:"0.1" scope:"hidden" key:"aliases"`
+
+	Assigned struct {
```
emlin: same thing. Versioned attribute will help on future's upgrade

hmahmood:
@emlin @dougm did split out the user name and password into their own fields, but we were not saving the credentials with the target URL anyway. The combined username and password were encrypted.

lib/install/validate/update.go
```diff
@@ -39,12 +47,58 @@ func (v *Validator) MigrateConfig(ctx context.Context, conf *config.VirtualConta
...
 func (v *Validator) migrateData(ctx context.Context, conf *config.VirtualContainerHostConfigSpec) (*config.VirtualContainerHostConfigSpec, error) {
+	bn, err := buildNumber(conf.Version)
+	if err != nil {
+		v.NoteIssue(err)
+		return nil, err
+	}
+
+	// migrate network
+	// change "external" to "public"
+	if n := conf.Networks["external"]; n != nil {
```
emlin: this check is part of .7 to .8, should have this inside of that scope

lib/install/validate/update.go
```diff
@@ -39,12 +47,58 @@ func (v *Validator) MigrateConfig(ctx context.Context, conf *config.VirtualConta
...
+		// copy over gateway and nameservers if the endpoint
+		// was not dynamic
+		for _, n := range conf.Networks {
+			if !n.Static && (n.IP == nil || ip.IsUnspecifiedIP(n.IP.IP)) {
+				continue
+			}
+			n.Network.Gateway = n.Network.Gateway_
+			n.Network.Nameservers = n.Network.Nameservers_
+		}
+	}
```
emlin: so these two blocks are all migrate data from .7 to .8. We should have a method explicitly named with that, and called in this migratedata method, that will help us while we drop support for .7

lib/config/executor/network_interface.go
emlin: then we'll need a definition for the structure. otherwise, we have no idea for which attributes should be considered in upgrade

emlin:
ok, never mind. As long as they are encrypted, I'm fine with it.

cmd/vic-machine/create/create.go
```diff
@@ -207,25 +207,25 @@ func (c *Create) Flags() []cli.Flag {
...
 		},
 
-		// external
+		// public
 		cli.StringFlag{
-			Name:        "external-network, en",
-			Value:       "VM Network",
-			Usage:       "The external network port group name (port forwarding and default route). Defaults to 'VM Network' and DHCP -- see advanced help (-x)",
-			Destination: &c.externalNetworkName,
+			Name:        "public-network, en",
```
hickeng: we should change the short name as well - en no longer makes sense.

hickeng:
@hmahmood Can we split this PR into it's component parts:
1. fix for dhcp on restart case
2. fix for network inspect - not sure that I understand why these values need to be persisted in the configuration - can we not pull them from the system?
3. rename vic-machine --external-network to --public-network (this is the piece of renaming to public that _needs_ to happen for 0.8)
4. rename external network throughout codebase (it'd be nice if this happened for 0.8 but little concern if it doesn't)

I've persuaded @mdubya66 that we're better off doing upgrade right than hacking it in like this and adding upgrade tech debt right from the beginning:
* I'm not happy with what we're having to do here for upgrade, or already having deprecated fields before we're even released.
* we need to do _significant_ work on the upgrade path in vic-machine as it doesn't currently extract the VCH configuration in a useful fashion for validation (meaning new required cli arguments will break us).
* we should be preserving credentials for the operations user

hmahmood:
@hickeng PR is updated with just external -> public changes.

> we should be preserving credentials for the operations user

This change was preserving them; not sure where you saw it wasn't.

hmahmood:
@hickeng https://github.com/vmware/vic/pull/3224/files for dhcp restart fix

hickeng:
@hmahmood comment about preserving the credentials relates to your comment about migrating userpw to new installs and the fact the we re-encrypt.
The changes doug made are positioning for the RBAC support and operations user - those will need to stay.

In general we need to:
1. add a stable Version for all stored data that we can access without knowing datastructure version details
2. add structured version by version migration mechanism (@caglar10ur and I have design notes for this but didn't realize we'd committed to it for 0.8, thought it was 0.9 so they're not written up)
3. add `vic-machine configure` and rework `create` to use that same path
4. modify `update` to actually extract the VCH configuration back into the InstallerData object, then follow `configure` path

Given the extent of that work I don't think it's viable for 0.8 unless one of us gets very bored over the holiday!

hmahmood:
So for userpw, I took out the re-encrypt in the upgrade path -- otherwise there is no way to migrate userpw. I was not sure why we were re-encrypting the config during upgrade anyway. But of course upgrade is broken in other ways than just that.

lib/apiservers/engine/backends/container.go
```diff
@@ -1429,7 +1429,7 @@ func validateCreateConfig(config *types.ContainerCreateConfig) error {
 	// validate port bindings
 	if config.HostConfig != nil {
 		var ips []string
-		if addrs, err := externalIPv4Addrs(); err != nil {
+		if addrs, err := publicIPv4Addrs(); err != nil {
 			log.Warnf("could not get address for external interface: %s", err)
```
mlh78750: log needs to refer to public as well

lib/install/validate/network.go
```diff
@@ -209,22 +209,22 @@ func (v *Validator) network(ctx context.Context, input *data.Data, conf *config.
...
-		v.suggestNetwork("--external-network", true)
+		v.NoteIssue(fmt.Errorf("Error checking network for --public-network: %s", err))
+		v.suggestNetwork("--public-network", true)
 	}
 	// Bridge network should be different than all other networks
-	v.checkNetworkConflict(input.BridgeNetworkName, input.ExternalNetwork.Name, "external")
+	v.checkNetworkConflict(input.BridgeNetworkName, input.PublicNetwork.Name, "public")
 	conf.AddNetwork(e)
 
 	// Client net - defaults to connect to same portgroup as external
```
mlh78750: public instead of external in the comments.

mlh78750:
couple of small fixes and LGTM

mlh78750:
I restarted your build since we had issues with aboot earlier.

hmahmood:
:recycle:

PR #3218, opened 2016-11-17
Unbind serial port for running containers during PL restart

### LINKED ISSUE #2528: [restart] portlayer should iterate over vms and unbind their network backed serial ports
otherwise esxi will keep trying to connect the serial to the old IP

anchal-agrawal:
This PR unbinds the `VirtualSerialPort` of running containers during portlayer init. 

If the portlayer crashes/restarts and the VCH is configured with DHCP, this change prevents us from binding the serial port to the old VCH IP.

Fixes #2528

### DIFF
--- a/lib/portlayer/portlayer.go
+++ b/lib/portlayer/portlayer.go
@@ -23,6 +23,7 @@ import (
 	"github.com/vmware/vic/lib/portlayer/network"
 	"github.com/vmware/vic/lib/portlayer/storage"
 	"github.com/vmware/vic/lib/portlayer/store"
+	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/datastore"
 	"github.com/vmware/vic/pkg/vsphere/extraconfig"
 	"github.com/vmware/vic/pkg/vsphere/session"
@@ -95,37 +96,45 @@ func Init(ctx context.Context, sess *session.Session) error {
 // for running containers. This is useful when the appliance or the portlayer restarts
 // and the VCH has a new IP. Any errors are logged and portlayer init proceeds as usual.
 func unbindSerialPorts(sess *session.Session) {
-	// Get all running containers from the cache
+	defer trace.End(trace.Begin(""))
+
+	ctx := context.Background()
+
+	// Get all running containers from the portlayer cache
 	runningState := new(exec.State)
 	*runningState = exec.StateRunning
 	containers := exec.Containers.Containers(runningState)
 
 	for i := range containers {
+		var containerID string
 		if containers[i].ExecConfig != nil {
-			log.Infof("unbinding serial port for running container: %s", containers[i].ExecConfig.ID)
+			containerID = containers[i].ExecConfig.ID
 		}
+		log.Infof("unbinding serial port for running container %s", containerID)
 
 		// Obtain a container handle
-		handle := containers[i].NewHandle(context.Background())
+		handle := containers[i].NewHandle(ctx)
 		if handle == nil {
-			log.Error("unable to obtain a handle for container")
+			log.Errorf("unable to obtain a handle for container %s", containerID)
 			continue
 		}
 
 		// Unbind the VirtualSerialPort
 		newHandle, err := attach.Unbind(handle)
 		if err != nil {
-			log.Errorf("unable to unbind serial port for container: %s", err)
+			log.Errorf("unable to unbind serial port for container %s: %s", containerID, err)
+			continue
+		}
+
+		execHandle, ok := newHandle.(*exec.Handle)
+		if !ok {
+			log.Errorf("handle type assertion failed for container %s", containerID)
 			continue
 		}
 
 		// Commit the handle
-		if execHandle, ok := newHandle.(*exec.Handle); !ok {
-			log.Error("handle type assertion failed for container")
-		} else {
-			if err := execHandle.Commit(context.Background(), sess, nil); err != nil {
-				log.Errorf("unable to commit handle for container: %s", err)
-			}
+		if err := execHandle.Commit(ctx, sess, nil); err != nil {
+			log.Errorf("unable to commit handle for container %s: %s", containerID, err)
 		}
 	}
 }
--- a/lib/portlayer/portlayer.go
+++ b/lib/portlayer/portlayer.go
@@ -18,6 +18,7 @@ import (
 	"path"
 
 	"github.com/vmware/vic/lib/guest"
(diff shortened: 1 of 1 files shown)


lib/portlayer/portlayer.go
```diff
@@ -83,5 +84,46 @@ func Init(ctx context.Context, sess *session.Session) error {
 		return err
 	}
 
+	unbindSerialPorts(sess)
```
matthewavery: You may want to add a comment hear as to why we are calling this here. I know there is a comment on the function definition. But the init code should also be as unambiguous as possible imo :)

matthewavery:
lgtm

lib/portlayer/portlayer.go
```diff
@@ -83,5 +84,46 @@ func Init(ctx context.Context, sess *session.Session) error {
...
+
+	for i := range containers {
+		if containers[i].ExecConfig != nil {
+			log.Infof("unbinding serial port for running container: %s", containers[i].ExecConfig.ID)
+		}
+
+		// Obtain a container handle
+		handle := containers[i].NewHandle(context.Background())
+		if handle == nil {
+			log.Error("unable to obtain a handle for container")
```
caglar10ur: Add the container id here so that we can see which container are we talking about

lib/portlayer/portlayer.go
```diff
@@ -83,5 +84,46 @@ func Init(ctx context.Context, sess *session.Session) error {
...
 }
+
+// unbindSerialPorts disconnects serial ports backed by network on the VCH's old IP
+// for running containers. This is useful when the appliance or the portlayer restarts
+// and the VCH has a new IP. Any errors are logged and portlayer init proceeds as usual.
+func unbindSerialPorts(sess *session.Session) {
+	// Get all running containers from the cache
+	runningState := new(exec.State)
+	*runningState = exec.StateRunning
+	containers := exec.Containers.Containers(runningState)
```
caglar10ur: containers := exec.Containers.Containers(&exec.StateRunning) should work here instead those, no?

lib/portlayer/portlayer.go
```diff
@@ -83,5 +84,46 @@ func Init(ctx context.Context, sess *session.Session) error {
...
+		}
+
+		// Unbind the VirtualSerialPort
+		newHandle, err := attach.Unbind(handle)
+		if err != nil {
+			log.Errorf("unable to unbind serial port for container: %s", err)
+			continue
+		}
+
+		// Commit the handle
```
caglar10ur: Too much going on here, you can separate the first if from the commit one
```
if ... !ok {
}

if err := ... {
....
}
```

lib/portlayer/portlayer.go
anchal-agrawal: I currently have

```
log.Infof("unbinding serial port for running container: %s", containers[i].ExecConfig.ID)
```

at the start of the every loop iteration. This prints the container ID for every container before operating on it, while avoiding a nil check for `container.ExecConfig` every time we want to log a message. `container.ExecConfig` will be non-nil in most circumstances, but I'm not sure if we can always guarantee that - specially since this code kicks in when the appliance or PL crashes. 

If we're sure that `container.ExecConfig` will always be non-nil, we can simply log the ID every time. Otherwise, I can add a nil-check before every log or do something like:
```
var containerID string
if container.ExecConfig != nil {
  containerID = container.ExecConfig.ID
}
...
log.Errorf("... %s", containerID)
```
What do you think?

caglar10ur: Not sure I followed. Yes you can assign containers[i].ExecConfig.ID to a variable and use it to shorten the code but this log.Error line comes after the nil check. What I was asking was simply changing this line 
```
log.Error("unable to obtain a handle for container")
```
to
```
log.Error("unable to obtain a handle for container %s", containers[i].ExecConfig.ID)
```

anchal-agrawal: Sorry about that. I meant to ask if we can guarantee that `container.ExecConfig` will be non-nil every time, because we deference it to get the ID.

hmahmood:
lgtm

How did you test this?

anchal-agrawal:
* Ran a long-running and a quick-ending container
* Killed the PL process or powered off/on the VCH from the UI
* The config spec was changed only for the long-running container and all functions invoked by this change returned w/o errors.

Please let me know if we can test this better.

caglar10ur:
LGTM

lib/portlayer/portlayer.go
caglar10ur: Yes you can as you are checking that in the beginning of the loop and this is init so it's guaranteed to run once before main

anchal-agrawal: Fixed.

lib/portlayer/portlayer.go
anchal-agrawal: Fixed.

lib/portlayer/portlayer.go
```diff
@@ -83,5 +84,54 @@ func Init(ctx context.Context, sess *session.Session) error {
...
+	unbindSerialPorts(sess)
+
 	return nil
 }
+
+// unbindSerialPorts disconnects serial ports backed by network on the VCH's old IP
+// for running containers. This is useful when the appliance or the portlayer restarts
+// and the VCH has a new IP. Any errors are logged and portlayer init proceeds as usual.
+func unbindSerialPorts(sess *session.Session) {
+	ctx := context.Background()
```
hickeng: can we have entry/exit trace as in other PL functions.

anchal-agrawal: Done!

hickeng:
@anchal-agrawal You will need to attach to the long running container, and still be attached when power cycling the applianceVM for the test to be legitimate.

anchal-agrawal:
@hickeng I observed the following:
```
$ sudo docker -H 192.168.77.163:2376 --tls run -it busybox
/ # 
 ... power off+on the VCH ...
/ # read tcp 192.168.77.165:46350->192.168.77.163:2376: read: connection reset by peer

$ sudo docker -H 192.168.77.163:2376 --tls run -it busybox
/ # 
 ... kill the PL ...
/ # Error response from daemon: Server error from portlayer: Get http://127.0.0.1:2377/containers/info/474550bf1b836d6f4a7245c431d919172fd835a4b56ee0052607f33dd4a76739: dial tcp 127.0.0.1:2377: getsockopt: connection refused

$ sudo docker -H 192.168.77.163:2376 --tls attach 4
/ #
```

We can attach again to the running container. Per my Slack conversation with @caglar10ur, this is expected behavior.

PR #3222, opened 2016-11-17
Log task MoID when vsphere task is created

fdawg4l:

### DIFF
--- a/pkg/vsphere/tasks/waiter.go
+++ b/pkg/vsphere/tasks/waiter.go
@@ -35,6 +35,7 @@ const (
 //FIXME: remove this type and refactor to use object.Task from govmomi
 //       this will require a lot of code being touched in a lot of places.
 type Task interface {
+	Reference() types.ManagedObjectReference
 	Wait(ctx context.Context) error
 	WaitForResult(ctx context.Context, s progress.Sinker) (*types.TaskInfo, error)
 }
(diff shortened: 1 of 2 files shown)


jzt:
LGTM

caglar10ur:
LGTM

fdawg4l:
.

pkg/vsphere/tasks/waiter.go
```diff
@@ -56,20 +57,21 @@ func Wait(ctx context.Context, f func(context.Context) (Task, error)) error {
...
 	var backoffFactor int64 = 1
 
 	for {
-		var t Task
-		if t, err = f(ctx); err == nil {
+		t, err := f(ctx)
+		log.Debugf("New task: %s", taskMoid(t))
+
+		if err == nil {
 			info, err = t.WaitForResult(ctx, nil)
```
mlh78750: We didn't check to see if t was nil also. This panic's vic-machine when we reach this code.

fdawg4l: This is very strange.  My original PR (which was squashed and merged) had a check in `taskMoid` for nil.  I sqaushed and merged *with this check!*

But, the revert doesn't include this code.  So 1-  PRs are not immuteable once they are commited since my revert of the revert now includes this check and 2- squash and merge is totally broken!

PR #3223, opened 2016-11-17
Add support for self generated CAs for registries

### LINKED ISSUE #2103: Support custom CA roots for registry validation
We need to be able to upload a self generated ca certificate to the appliance's in order to handle private registries using TLS with self-signed certificates.  This is for docker login and docker pull.

The docker personality server must also be able to use this certificate.

sflxn:
Add support for uploading certificate authority certs via vic-machine's
"--registry-ca" for docker registry login and pull usage.  The VCH's
default CA bundle will not be affected.  Any other application running
in the VCH will not be affected.

The docker personality will use these CA certs for docker login and
docker pull.

Resolves #2103

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -99,7 +99,8 @@ type Create struct {
 	noTLSverify     bool
 	advancedOptions bool
 
-	clientCAs cli.StringSlice
+	clientCAs   cli.StringSlice
+	registryCAs cli.StringSlice
 
 	containerNetworks         cli.StringSlice
 	containerNetworksGateway  cli.StringSlice
@@ -401,6 +402,12 @@ func (c *Create) Flags() []cli.Flag {
 
 		// registries
 		cli.StringSliceFlag{
+			Name:   "registry-ca, rc",
+			Usage:  "Specify a list of additional certificate authority files to use to verify secure registry servers",
+			Value:  &c.registryCAs,
+			Hidden: true,
+		},
+		cli.StringSliceFlag{
 			Name:  "insecure-registry, dir",
 			Value: &c.insecureRegistries,
 			Usage: "Specify a list of permitted insecure registry server URLs",
@@ -520,7 +527,7 @@ func (c *Create) processParams() error {
 		return errors.Errorf("Error occurred while processing volume stores: %s", err)
 	}
 
-	if err := c.processInsecureRegistries(); err != nil {
+	if err := c.processRegistries(); err != nil {
 		return err
 	}
 
@@ -811,7 +818,18 @@ func (c *Create) processVolumeStores() error {
 	return nil
 }
 
-func (c *Create) processInsecureRegistries() error {
+func (c *Create) processRegistries() error {
+	// load addtional certificate authorities for use with registries
+	if len(c.registryCAs) > 0 {
+		registryCAs, err := c.loadRegistryCAs()
+		if err != nil {
+			return errors.Errorf("Unable to load CA certificates for registry logins: %s", err)
+		}
+
+		c.RegistryCAs = registryCAs
+	}
+
+	// load a list of insecure registries
 	for _, registry := range c.insecureRegistries {
 		url, err := url.Parse(registry)
 		if err != nil {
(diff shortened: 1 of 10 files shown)


lib/imagec/docker.go
```diff
@@ -323,6 +327,7 @@ func FetchImageManifest(ctx context.Context, options Options, progressOutput pro
 		Password:           options.Password,
 		Token:              options.Token,
 		InsecureSkipVerify: options.InsecureSkipVerify,
+		RootCAs:			options.RegistryCAs,
```
caglar10ur: Looks like this file needs gofmt

cmd/vic-machine/create/create.go
```diff
@@ -964,6 +982,25 @@ func (c *Create) loadCertificates() ([]byte, *certificate.KeyPair, error) {
...
+	var registryCerts []byte
+	for _, f := range c.registryCAs {
+		b, err := ioutil.ReadFile(f)
+		if err != nil {
+			err = errors.Errorf("Failed to load authority from file %s: %s", f, err)
+			return nil, err
+		}
+
+		registryCerts = append(registryCerts, b...)
+		log.Infof("Loading registry CA from %s", f)
```
caglar10ur: I think this needs to be the first line inside the loop

lib/apiservers/engine/backends/backends.go
```diff
@@ -81,6 +82,10 @@ func Init(portLayerAddr, product string, config *config.VirtualContainerHostConf
 		} else {
 			portLayerName = product + " " + productVersion + " Backend Engine"
 		}
+
+		if len(config.Certificate.RegistryCertificateAuthorities) > 0 {
```
caglar10ur: you can drop this check as  loadRegistryCACerts already does it

lib/install/data/data.go
```diff
@@ -32,9 +32,10 @@ type Data struct {
 	common.Compute
 	common.VCHID
 
-	CertPEM   []byte
-	KeyPEM    []byte
-	ClientCAs []byte
+	CertPEM   	[]byte
+	KeyPEM    	[]byte
+	ClientCAs 	[]byte
+	RegistryCAs []byte
```
caglar10ur: gofmt

lib/install/validate/validator.go
```diff
@@ -403,6 +404,26 @@ func (v *Validator) certificateAuthorities(ctx context.Context, input *data.Data
...
 
+func (v *Validator) registries(ctx context.Context, input *data.Data, conf *config.VirtualContainerHostConfigSpec) {
+	defer trace.End(trace.Begin(""))
+
+	// copy the list of insecure registries
+	conf.InsecureRegistries = input.InsecureRegistries
+
+	if len(input.RegistryCAs) == 0 {
+		return
+	}
```
caglar10ur: check first copy later?

lib/apiservers/engine/backends/backends.go
```diff
@@ -310,3 +313,23 @@ func setPortMapping(info *models.ContainerInfo, backend *Container, container *c
...
+
+func loadRegistryCACerts() {
+	if len(vchConfig.RegistryCertificateAuthorities) == 0 {
+		return
+	}
+
+	rootCertPool, err := x509.SystemCertPool()
+	log.Debugf("Loaded %d CAs for registries from system CA bundle", len(rootCertPool.Subjects()))
+	if err != nil {
+		log.Fatalf("Unable to load system CAs")
```
hickeng: This will cause the personality to exit, then be restarted immediately by vic-init, repeated ad infinitum.

Use x509.NewPool() to initialize an empty pool and append to that - it's better than nothing.

lib/apiservers/engine/backends/backends.go
```diff
@@ -310,3 +313,23 @@ func setPortMapping(info *models.ContainerInfo, backend *Container, container *c
...
+	}
+
+	if !rootCertPool.AppendCertsFromPEM(vchConfig.RegistryCertificateAuthorities) {
+		log.Fatalf("Unable to load CAs for registry access in config")
```
hickeng: Log an error - potentially write it to a file of major errors that vic-admin can display, but we shouldn't become completely non-functional because of this.

hickeng:
LGTM once identified issues are addressed

lib/imagec/docker.go
sflxn: fixed

cmd/vic-machine/create/create.go
```diff
@@ -811,7 +818,18 @@ func (c *Create) processVolumeStores() error {
...
 
-func (c *Create) processInsecureRegistries() error {
+func (c *Create) processRegistries() error {
+	// load addtional certificate authorities for use with registries
+	if len(c.registryCAs) > 0 {
+		registryCAs, err := c.loadRegistryCAs()
+		if err != nil {
+			log.Errorf("Unable to load CA certificates for registry logins: %s", err)
+		} else {
+			c.RegistryCAs = registryCAs
```
sflxn: Returning an error now.

lib/apiservers/engine/backends/backends.go
sflxn: yeah

lib/install/validate/validator.go
sflxn: Don't understand the comment.

lib/apiservers/engine/backends/backends.go
sflxn: No longer fatal, just error now.

lib/apiservers/engine/backends/backends.go
sflxn: No longer fatal, just error now.

caglar10ur:
LGTM :sheep: :it:

PR #3232, opened 2016-11-18
Log task MoID when vsphere task is created

fdawg4l:


emlin:
lgtm

pkg/vsphere/tasks/waiter.go
```diff
@@ -56,20 +57,21 @@ func Wait(ctx context.Context, f func(context.Context) (Task, error)) error {
...
 func WaitForResult(ctx context.Context, f func(context.Context) (Task, error)) (*types.TaskInfo, error) {
-	var err error
 	var info *types.TaskInfo
 	var backoffFactor int64 = 1
 
 	for {
-		var t Task
-		if t, err = f(ctx); err == nil {
+		t, err := f(ctx)
+		log.Debugf("New task: %s", taskMoid(t))
```
mlh78750: If there is an err will t always be non-nil?

mlh78750: actually this is protected in the function you call with t, but below...

mlh78750: nevermind I see what you did in the function.. :)

mlh78750:
lgtm

pkg/vsphere/tasks/waiter.go
fdawg4l: and i had that in my *original* pr!  i squashed and merged and it went away!  It was it's own commit, was in the PR (i made sure the 3 lines were there) and poof, gone.  AND!  The original PR (since i reused the same branch) now includes these changes.  PRs apparently aren't immutable!  

Why are we using this broken system again?  If PRs aren't immutable, we can't rely on review history.

vburenin:
LGTM

fdawg4l:
.

PR #3233, opened 2016-11-18
Do not persist user credentials in memory in admin

### LINKED ISSUE #3088: vic-admin: use vim session cookie and depend on vsphere session management
Update vicadmin to store the govmomi session cookie as the cached credentials, instead of the user supplied credentials.
This allow us to match the session expiration behaviour configured for vSphere and to honor explicit cancellation of an individual session by a viadmin.

It may be viable to address #3083 at the same time.

This is follow on work to #2883

gigawhitlocks:
Fixes #3088 -- after initial login, we replace the session.Client held w/ one that does not have credentials, but instead has a SOAP cookie to talk to vSphere.

Still doing a little cleanup / ad-hoc testing but please review away.
If the vSphere session is expired for any reason, the user will be forced to log in again.

### DIFF
--- a/cmd/vicadmin/usersession.go
+++ b/cmd/vicadmin/usersession.go
@@ -44,20 +44,22 @@ type UserSessionStore struct {
 }
 
 type UserSessionStorer interface {
-	Add(id string, config *session.Config) *UserSession
+	Add(id string, config *session.Config, vs *session.Session) *UserSession
 	Delete(id string)
 	VSphere(id string) (vSphereSession *session.Session, err error)
 	UserSession(id string) *UserSession
 }
 
-// Add creates a config and initializes the UserSession and adds it to the UserSessionStore & returns the created UserSession
-func (u *UserSessionStore) Add(id string, config *session.Config) *UserSession {
+// Add a session. VS may be nil if host is plain ESX
+func (u *UserSessionStore) Add(id string, config *session.Config, vs *session.Session) *UserSession {
 	u.mutex.Lock()
 	defer u.mutex.Unlock()
 	sess := &UserSession{
 		id:      id,
 		created: time.Now(),
+		// TODO strip out config cause it's not needed anymore, but shows up in a number of places
 		config:  config,
+		vsphere: vs,
 	}
 	u.sessions[id] = sess
 	return sess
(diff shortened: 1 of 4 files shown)


pkg/vsphere/session/session.go
```diff
@@ -155,14 +155,10 @@ func (s *Session) Connect(ctx context.Context) (*Session, error) {
 		}
 	}
 
+	soapURL.User = s.User
 	// Update the service URL with expanded defaults
 	s.Service = soapURL.String()
```
hmahmood: Won't this  put the credentials in s.Service? I know you are nil-ing Service elsewhere, but this code is called from a  whole bunch of other places.

cmd/vicadmin/usersession.go
```diff
@@ -69,31 +71,31 @@ func (u *UserSessionStore) Delete(id string) {
...
+		return nil, fmt.Errorf("No vSphere session found. User with ID %s must authenticate again.", id)
 	}
 
-	log.Infof("Creating vSphere session for vicadmin usersession %s", id)
-	s, err := vSphereSessionGet(us.config)
-	if err != nil {
-		return nil, err
-	}
-	us.vsphere = s
+	// log.Infof("Creating vSphere session for vicadmin usersession %s", id)
```
hmahmood: Not needed anymore?

cmd/vicadmin/vicadm.go
```diff
@@ -405,29 +407,48 @@ func (r datastoreReader) open() (entry, error) {
...
-	usersession, err := session.SessionManager.UserSession(ctx)
+	usersession, err := s.SessionManager.UserSession(ctx)
 	if err != nil {
 		log.Errorf("Got %s while creating user session", err)
 		return nil, err
 	}
 
 	log.Infof("Got session from vSphere with key: %s username: %s", usersession.Key, usersession.UserName)
-	return session, nil
+	return s, nil
```
hmahmood: You could strip credentials before returning this.

cmd/vicadmin/usersession.go
hmahmood: This method is not creating a session anymore; what are the implications of that?

pkg/vsphere/session/session.go
gigawhitlocks: It's really the same behavior as before; the only change is that we skip out on checking to make sure soapURL.User is non-nil because if it's nil we still want it added to the soapURL. @hickeng can weigh in on whether this might cause an issue elsewhere, hopefully.

cmd/vicadmin/usersession.go
gigawhitlocks: This is where, last night, we were discussing what happens if vSphere expires the session or something like that -- the user just has to log in again. Perfectly OK behavior IMHO

pkg/vsphere/session/session.go
gigawhitlocks: I am worried that removing this entirely will cause issues in other parts of the codebase w/o much time for deep testing.

hmahmood:
lgtm

cmd/vicadmin/vicadm.go
```diff
@@ -405,29 +407,53 @@ func (r datastoreReader) open() (entry, error) {
 	return httpEntry(r.path, res)
 }
 
+// removes user credentials from "in"
```
anchal-agrawal: Nit - we can say `// stripCredentials removes user credentials from session`

gigawhitlocks: yep

anchal-agrawal: Ah, I meant to say that we should replace `"in"` with `session`. 😃

Just a tiny detail, we can leave it as-is.

anchal-agrawal:
LGTM

PR #3236, opened 2016-11-18
New vm power on event in vc 6.5

emlin:
otherwise, docker ps will still show container stopped, after OOB vm power on.


cgtexmex:
LGTM

hickeng:
LGTM

PR #3237, opened 2016-11-19
Hide upgrade command

emlin:
We still do not support upgrade from earlier version to .8, so make this command hidden from customer.

@stuclem @npakrasi please help to update user document as well


mdubya66:
lgtm

jzt:
LGTM

PR #3244, opened 2016-11-19
Recursive rmi implementation [full ci]

### LINKED ISSUE #2973: Implement `rmi` up to last tag in SPL
We're currently only deleting leaf images when an `rmi` is called on a given `repository:tag`.  Until recently we couldn't have more than 1 tag on a given image ID, but the personality's image cache now supports multiple tags and untagging on `rmi`.  We need to integrate this into the PL delete mechanism to delete upto the next image, thereby actually deleting more than just the last image.

Without this change, pull of an image (tag:repo) which you just deleted will show there's nothing to do because the majority of the data is still in the datastore.  (We picked this to be conservative until the personality's image cache had this tagging/persistance-of-tags which it has now.)

fdawg4l:
Fixes #2973 

Deletes a given image all the way up to the next tag, layer with
degree > 1, or used image.

### DIFF
--- a/lib/apiservers/engine/backends/image.go
+++ b/lib/apiservers/engine/backends/image.go
@@ -19,6 +19,7 @@ import (
 	"io"
 	"os"
 	"sort"
+	"strings"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -35,6 +36,7 @@ import (
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/storage"
 	"github.com/vmware/vic/lib/imagec"
 	"github.com/vmware/vic/lib/metadata"
+	"github.com/vmware/vic/lib/portlayer/util"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/uid"
 	"github.com/vmware/vic/pkg/vsphere/sys"
@@ -63,9 +65,10 @@ func (i *Image) Exists(containerName string) bool {
 func (i *Image) ImageDelete(imageRef string, force, prune bool) ([]types.ImageDelete, error) {
 	defer trace.End(trace.Begin(imageRef))
 
-	var deleted []types.ImageDelete
-	var userRefIsID bool
-	var imageRemoved bool
+	var (
+		deletedRes  []types.ImageDelete
+		userRefIsID bool
+	)
 
 	// Use the image cache to go from the reference to the ID we use in the image store
 	img, err := cache.ImageCache().Get(imageRef)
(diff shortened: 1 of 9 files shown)


lib/apiservers/engine/backends/image.go
```diff
@@ -90,17 +93,54 @@ func (i *Image) ImageDelete(imageRef string, force, prune bool) ([]types.ImageDe
...
+				i, err := imagec.LayerCache().Get(di.ID)
+				if err == nil {
+					id = i.Layer.BlobSum
+				}
+
+				// remove the layer from the layer cache (used by imagec)
+				imagec.LayerCache().Remove(di.ID)
+
+				// form the response
+				imageDeleted := types.ImageDelete{Deleted: strings.TrimPrefix(id, "sha:")}
```
jzt: should this be `sha256:` ?

fdawg4l: I *just* realized this.  force pushed.

jzt:
LGTM

lib/apiservers/engine/backends/image.go
```diff
@@ -90,17 +93,54 @@ func (i *Image) ImageDelete(imageRef string, force, prune bool) ([]types.ImageDe
...
+		params := storage.NewDeleteImageParamsWithContext(ctx).WithStoreName(storeName).WithID(img.ID).WithKeepNodes(keepNodes)
 		// TODO: This will fail if any containerVMs are referencing the vmdk - vanilla docker
 		// allows the removal of an image (via force flag) even if a container is referencing it
 		// should vic?
-		_, err = PortLayerClient().Storage.DeleteImage(params)
+		res, err := PortLayerClient().Storage.DeleteImage(params)
+
+		// We may have deleted images despite error.  Account for that in the cache.
+		if res != nil {
+			for _, di := range res.Payload {
```
emlin: is the di confusing? :)

emlin:
lgtm 👍

lib/apiservers/engine/backends/image.go
fdawg4l: Yeah, good point.  It meant "deletedImage".  I'll rename it.

lib/apiservers/engine/backends/image.go
```diff
@@ -131,7 +169,7 @@ func (i *Image) ImageDelete(imageRef string, force, prune bool) ([]types.ImageDe
 		// updates
 		refNamed, _ := cache.RepositoryCache().Remove(tags[i], false)
 		dd := types.ImageDelete{Untagged: refNamed}
-		deleted = append(deleted, dd)
+		deletedRes = append(deletedRes, dd)
```
fdawg4l: I didn't make this variable but I'll rename it.

fdawg4l:
.

Issue #2872, opened 2016-10-24
Longevity failure: missing output.log in vic-admin bundle


mhagen-vmware:
After almost 4 hours of runtime, hit the following issue:

```
    '-rw-r----- 0/0           65542 2016-10-24 05:13 docker/verbose
-rw-r----- 0/0           99683 2016-10-24 05:13 vpxd:vpxd.log
-rw-r----- 0/0         1608331 2016-10-24 05:13 vicadm/heap
-rw-r----- 0/0            1325 2016-10-24 05:13 lsmod
-rw-r----- 0/0          159458 2016-10-24 05:13 hostd/HostSystem:host-10 @ /stress-datacenter/host/cls/homer.eng.vmware.com
-rw-r----- 0/0          194449 2016-10-24 05:13 VCH-0-2745/vmware.log
-rw-r----- 0/0             996 2016-10-24 05:13 ip-addr
-rw-r----- 0/0              41 2016-10-24 05:13 vicadm/block
-rw-r----- 0/0             204 2016-10-24 05:13 free
-rw-r----- 0/0             129 2016-10-24 05:13 sys-block
-rw-r----- 0/0             265 2016-10-24 05:13 proc-self-fd
-rw-r----- 0/0          448810 2016-10-24 05:13 docker/heap
-rw-r----- 0/0              68 2016-10-24 05:13 disk-by-label
-rw-r----- 0/0          179702 2016-10-24 05:13 vic-init/heap
-rw-r----- 0/0         4398735 2016-10-24 05:13 portlayer/heap
-rw-r----- 0/0           77659 2016-10-24 05:13 vicadm/verbose
-rw-r----- 0/0            3624 2016-10-24 05:13 ps
-rw-r----- 0/0           21785 2016-10-24 05:13 mad_yalow-2f6aac4425b2734f7a6f5e520d4ff53e4d6eec2d523163321cdb4ea9835f4d5e/tether.debug
    [ Message content over the limit has been removed. ]
...--- 0/0         2463173 2016-10-24 05:13 var/log/vic/docker-personality.log
-rw-r----- 0/0        13957528 2016-10-24 05:13 var/log/vic/init.log
-rw-r----- 0/0            1537 2016-10-24 05:13 proc-mounts
-rw-r----- 0/0             519 2016-10-24 05:13 ip-link
-rw-r----- 0/0              41 2016-10-24 05:13 portlayer/block
[...]

PR #3247, opened 2016-11-20
Allow for logs to not show up for a little bit

mhagen-vmware:
Fix for longevity issue, we have tech debt that needs to be addressed at some point but the logs won't necessarily be available instantly after working on a container.

### DIFF
--- a/tests/resources/Util.robot
+++ b/tests/resources/Util.robot
@@ -392,7 +392,7 @@ Run Regression Tests
     Should Be Equal As Integers  ${rc}  0
     Should Contain  ${output}  Exited
 
-    Wait Until Keyword Succeeds  5x  10s  Check For The Proper Log Files
+    Wait Until Keyword Succeeds  5x  10s  Check For The Proper Log Files  ${container}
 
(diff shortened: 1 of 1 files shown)


fdawg4l:
LGTM

mdubya66:
lgtm

hmahmood:
I think we should leave this test alone for now. I want to see if @caglar10ur recent fixes also fix this. See my comment in https://github.com/vmware/vic/issues/2872

hickeng:
@hmahmood the log files are created by vsphere as soon as the VM powers on - nothing @caglar10ur does will influence their presence, only their content.

@mhagen-vmware don't close #2872 as we don't know what's actually happening- can we add in the usual github issue open check - in this case for the repeated check of the logs. That means we won't close #2872 forgetting why we're not seeing it occur any more.

mdubya66:
Do it

mhagen-vmware:
.

hmahmood:
@hickeng are you sure they are created on power on, and not on first write to the serial port?

PR #3248, opened 2016-11-20
Persist appliance init log

### LINKED ISSUE #2079: Persist appliance log bundle on shutdown
Currently when the appliance shuts down, all of the logs are dropped due to stateless filesystem. Generally this isn't an issue as they can be pulled via vicadmin however it leaves a hole between VM poweron and vicadmin initialization during which we're unable to tell what has occurred.

The log bundle is the contents of /var/log/vic/ at a minimum - ideally it would include the rest of the local contents from the full vic-admin log bundle.

The most simplistic approach is exactly the one used for the containers - leverage a file backed serial port and write the log bundle to that. In this scenario we would configure the port with Replace instead of Append.

On completion:
- [ ] appliance log bundle is persisted on shutdown

hickeng:
Add a mechanism to persist/expose the initialization log for the
appliance. This uses the same serial mechanism as the containerVMs.
If debug > 0 then the component logs will be duplicated in here as well.

This also includes some tidying of the installer output - namely logging
of the install parameters in a less ugly and adhoc fashion.

Fixes #2079

### DIFF
--- a/cmd/vicadmin/vicadm.go
+++ b/cmd/vicadmin/vicadm.go
@@ -346,6 +346,38 @@ func listVMPaths(ctx context.Context, s *session.Session) ([]logfile, error) {
 	return logfiles, nil
 }
 
+// addApplianceLogs whitelists the logs to include for the appliance.
+// TODO: once we've started encrypting all potentially senstive data and filtering out guestinfo.ovfEnv
+// we can resume collection of vmware.log and drop the appliance specific handling
+func addApplianceLogs(ctx context.Context, s *session.Session, readers map[string]entryReader) error {
+	self, err := guest.GetSelf(ctx, s)
+	if err != nil || self == nil {
+		return fmt.Errorf("Unable to collect appliance logs due to unknown self-reference: %s", err)
+	}
+
+	self2 := vm.NewVirtualMachineFromVM(ctx, s, self)
+	path, err := self2.DSPath(ctx)
+	if err != nil {
+		return err
+	}
+
+	ds, err := s.Finder.Datastore(ctx, path.Host)
+	if err != nil {
+		return err
+	}
+
+	wpath := fmt.Sprintf("appliance/tether.debug")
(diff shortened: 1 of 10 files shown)


cmd/vicadmin/vicadm.go
```diff
@@ -361,6 +390,11 @@ func findDatastoreLogs(c *session.Session) (map[string]entryReader, error) {
 		return nil, errors.New(detail)
 	}
 
+	err = addApplianceLogs(ctx, c, readers)
+	if err != nil {
+		log.Errorf("Issue collecting appliance logs: %s", err)
+	}
```
caglar10ur: Guessing we are swallowing this error by purpose

lib/install/management/appliance.go
```diff
@@ -284,6 +284,32 @@ func (d *Dispatcher) addParaVirtualSCSIController(devices object.VirtualDeviceLi
 	return devices, nil
 }
 
+// addSerialPort is separate from the configLogging because I could not successfully add a serial port after initial VM creation
+// The controller slot kept colliding with the Keyboard no matter whether it was explicitly specified or left as -1 for vSphere to determine
```
caglar10ur: hmm  govc vm.create -on=false caglar followed by govc device.serial.add -vm caglar seems to be working here

caglar10ur:
LGTM

hmahmood:
lgtm

hickeng:
Unsure why this works on my local deployments, but we're seeing the following in CI so I've added the path to the permissions-setup:
```
time="2016-11-20T04:00:52Z" level=error msg="Issue collecting appliance logs: Unable to collect appliance logs due to unknown self-reference: error retrieving vm uuid: open /sys/class/dmi/id/product_serial: permission denied"
```

cmd/vicadmin/vicadm.go
hickeng: yup - log and continue with best effort is the modus operandi for vic-admin

hickeng:
poke

PR #3250, opened 2016-11-20
Move lock file from /tmp to /var/run/lock. Do not run logrotate immediately on boot.

### LINKED ISSUE #3249: race condition with systemd-tmpfiles-clean.timer
On VCH boot systemd runs systemd-tmpfiles-clean.timer that cleans up /tmp directory. It results in race condition between logrotate runner that stores its file lock and config in /tmp directory.

vburenin:
Fixed #3249

### DIFF
--- a/pkg/filelock/flock.go
+++ b/pkg/filelock/flock.go
@@ -36,7 +36,7 @@ type FileLock struct {
 func NewFileLock(lockName string) *FileLock {
 	return &FileLock{
 		LockName: lockName,
-		LockFile: filepath.Join(os.TempDir(), lockName),
+		LockFile: filepath.Join("/var/run/lock", lockName),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


pkg/filelock/flock.go
```diff
@@ -36,7 +36,7 @@ type FileLock struct {
 func NewFileLock(lockName string) *FileLock {
 	return &FileLock{
 		LockName: lockName,
-		LockFile: filepath.Join(os.TempDir(), lockName),
+		LockFile: filepath.Join("/var/run/lock", lockName),
```
hmahmood: Does /var/run/lock exist always?

vburenin: yes, it is created during boot.

caglar10ur:
LGTM

hickeng:
LGTM

Issue #3256, opened 2016-11-20
[vic-machine] Cannot deploy to a vcenter


caglar10ur:
After #3248

```
time=2016-11-20T19:15:37.273500844Z level=debug msg=[BEGIN] [github.com/vmware/vic/lib/install/management.(*Dispatcher).addSerialPort:290]
time=2016-11-20T19:15:37.273689758Z level=debug msg=[ END ] [github.com/vmware/vic/lib/install/management.(*Dispatcher).addSerialPort:290] [186.07µs]
time=2016-11-20T19:15:37.274053516Z level=debug msg=[ END ] [github.com/vmware/vic/lib/install/management.(*Dispatcher).createApplianceSpec:314] [729.138117ms]
DEBU[2016-11-20T19:15:37Z] unexpected fault on task retry : &types.InvalidDeviceSpec{InvalidVmConfig:types.InvalidVmConfig{VmConfigFault:types.VmConfigFault{VimFault:types.VimFault{MethodFault:types.MethodFault{FaultCause:(*types.LocalizedMethodFault)(nil), FaultMessage:[]types.LocalizableMessage(nil)}}}, Property:"deviceChange[4].device.backing.fileName"}, DeviceIndex:4}
ERRO[2016-11-20T19:15:37Z] Unable to create appliance VM: Invalid configuration for device '4'.
time=2016-11-20T19:15:37.884169476Z level=debug msg=[ END ] [github.com/vmware/vic/lib/install/management.(*Dispatcher).createAppliance:453] [1.33996303s]
time=2016-11-20T19:15:37.884488470Z level=debug msg=[ END ] [github.com/vmware/vic/lib/install/management.(*Dispatcher).CreateVCH:33] [11.362940731s] ZzZ
time=2016-11-20T19:15:37.884807989Z level=debug msg=[BEGIN] [github.com/vmware/vic/lib/install/management.(*Dispatcher).CollectDiagnosticLogs:171]
INFO[2016-11-20T19:15:37Z] Collecting 538a808c-d007-4b61-9cec-b641d3d654cd vpxd.log
time=2016-11-20T19:15:38.077069342Z level=debug msg=[ END ] [github.com/vmware/vic/lib/install/management.(*Dispatcher).CollectDiagnosticLogs:171] [192.254789ms]
ERRO[2016-11-20T19:15:38Z] Creating the appliance failed with Invalid configuration for device '4'.. Exiting...
ERRO[2016-11-20T19:15:38Z] --------------------
```

[...]

PR #3260, opened 2016-11-21
Fix log persistence on vCenter

### LINKED ISSUE #2079: Persist appliance log bundle on shutdown

hickeng:
This fixes the log persistence code in a slightly hacky fashion.
We do not get to collect the datastore version of the log in the log
in some scenarios (type unknown) because the file is locked when the
appliance is powered on. Straight vmfs on local datastore this isn't an
issue but it shows up on fibrechannel shared storage on VC. Don't know for
vSAN. As such removing the check from the CI tests.

Also revising the format of the flag logging per Emma's request and
addressing minor issue in bash-helpers.

Fixes #2079


emlin:
lgtm

caglar10ur:
Will try vsan in 10 min

caglar10ur:
LGTM (for the https://github.com/vmware/vic/issues/3256 part) having vsan/vcenter issues so still not sure whether we can grab the log

caglar10ur:
Looks like it is the same for vsan too, the bundle cannot grab the datastore version of the file.

Issue #3259, opened 2016-11-20
vic-admin unable to get logs from VC after initial download of log bundle


jzt:
If you authenticate with vic admin and then download the log bundle + container logs, it will contain vpxd.log and container logs. If you immediately download the log bundle again, vpxd.log and container logs are missing from the archive. This happens for each subsequent log bundle download until you re-authenticate.

Logs show the following:

```
time=2016-11-19T00:44:10.629604061Z level=debug msg=[BEGIN] [main.findDatastoreLogs:383] 
time=2016-11-19T00:44:10.629644459Z level=debug msg=[BEGIN] [main.listVMPaths:283] 
time="2016-11-19T00:44:10Z" level=error msg="Unable to get children vm of resource pool : NotAuthenticated" 
time=2016-11-19T00:44:10.639456084Z level=debug msg=[ END ] [main.listVMPaths:283] [9.843807ms]  
time="2016-11-19T00:44:10Z" level=error msg="unable to perform datastore log collection due to failure looking up paths: NotAuthenticated" 
time=2016-11-19T00:44:10.639496807Z level=debug msg=[ END ] [main.findDatastoreLogs:383] [9.891966ms]  
time="2016-11-19T00:44:10Z" level=warning msg="error searching datastore: unable to perform datastore log collection due to failure looking up paths: NotAuthenticated" 
```

anchal-agrawal:
We are currently logging out of the session after the first bundle collection: https://github.com/vmware/vic/blob/master/cmd/vicadmin/server.go#L447

Removing that fixes this issue, and I no longer see #3258 as well.

emlin:
vicadmin override session client https://github.com/vmware/vic/blob/master/cmd/vicadmin/vicadm.go#L455
which will not include session keepalive, and then will cause user session timeout after longtime no operation. 
If we need to override client anyway, keepalive should be set similar to what we did in session
https://github.com/vmware/vic/blob/master/pkg/vsphere/session/session.go#L189

mdubya66:
Closing as #3270 was merged

Issue #3258, opened 2016-11-20
vic-admin times out and prevents re-login


mdubya66:
Using  vic-machine-darwin version v0.7.0-7115-d58eb98

<img width="1092" alt="screen shot 2016-11-20 at 3 00 21 pm" src="https://cloud.githubusercontent.com/assets/15821624/20466125/87f205de-af32-11e6-9eb2-8cb5835f67e5.png">

Panic seen in vicadmin.log

ime="2016-11-20T21:00:48Z" level=info msg="Setting hostname to vch7115"
time=2016-11-20T21:00:48.976850527Z level=debug msg=[BEGIN] [github.com/vmware/vic/lib/install/validate.CreateFromVCHConfig:62]
time=2016-11-20T21:00:48.976856811Z level=debug msg=[ END ] [github.com/vmware/vic/lib/install/validate.CreateFromVCHConfig:62] [4.53µs]
time="2016-11-20T21:00:48Z" level=info msg="Using management IP {172.16.127.199 ffffff00} for firewall check"
time="2016-11-20T21:00:48Z" level=error msg="Unable to get the list of hosts attached to given storage: NotAuthenticated"
time="2016-11-20T21:00:49Z" level=info msg="FirewallStatus set to: <i class=\"icon-attention\"></i>"
time="2016-11-20T21:00:49Z" level=info msg="FirewallIssues set to: <span class=\"error-message\">NotAuthenticated</span>\n"
time="2016-11-20T21:00:49Z" level=info msg="LicenseStatus set to: <i class=\"icon-attention\"></i>"
time="2016-11-20T21:00:49Z" level=info msg="LicenseIssues set to: <span class=\"error-message\">NotAuthenticated</span>\n"
time="2016-11-20T21:00:49Z" level=info msg="NetworkStatus set to: <i class=\"icon-ok\"></i>"
time="2016-11-20T21:00:49Z" level=info msg="NetworkIssues set to: "
time="2016-11-20T21:00:49Z" level=error msg="Unable to collect information for datastore datastore1: NotAuthenticated"
time=2016-11-20T21:00:49.181227079Z level=debug msg=[ END ] [github.com/vmware/vic/lib/vicadmin.NewValidator:106] [204.405796ms]
time=2016-11-20T21:00:49.181247539Z level=debug msg=[ END ] [main.(*server).index:551] [204.446849ms]
[...]

PR #3270, opened 2016-11-21
Fix panic in QueryDatastore when session becomes nil

gigawhitlocks:
When sessions are expired (or nuked via govc), we were seeing an inconsistent state in vicadmin caused by a panic in QueryDatastore. This fixes that by terminating QueryDatastore if the datastore is inaccessible. The method also returns an err now, though it's not necessary that we do anything with the error at call time currently.

### DIFF
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -439,6 +439,10 @@ func (s *server) bundleContainerLogs(res http.ResponseWriter, req *http.Request,
 	c, err := s.getSessionFromRequest(context.Background(), req)
 	if err != nil {
 		log.Errorf("Failed to get vSphere session while bundling container logs due to error: %s", err.Error())
+		if strings.Contains(err.Error(), "Could not validate session") {
+			http.Redirect(res, req, "/logout", http.StatusTemporaryRedirect)
+			return
+		}
 		http.Error(res, genericErrorMessage, http.StatusInternalServerError)
 		return
 	}
@@ -549,6 +553,10 @@ func (s *server) index(res http.ResponseWriter, req *http.Request) {
 	defer trace.End(trace.Begin(""))
 	ctx := context.Background()
 	sess, err := s.getSessionFromRequest(ctx, req)
+	if err != nil && strings.Contains(err.Error(), "Could not validate session") {
+		http.Redirect(res, req, "/logout", http.StatusTemporaryRedirect)
+		return
+	}
 	v := vicadmin.NewValidator(ctx, &vchConfig, sess)
 
 	if sess == nil {
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -19,7 +19,6 @@ import (
 	"compress/gzip"
 	"crypto/tls"
 	"crypto/x509"
-	"fmt"
 	"html/template"
 	"net"
 	"net/http"
@@ -526,45 +525,19 @@ func (s *server) tailFiles(res http.ResponseWriter, req *http.Request, names []s
 	}
 }
 
-// deriveErrorMessage takes a vSphere session error and returns
-// an error string that is safe to display on the vicadmin page.
-func deriveErrorMessage(err error) string {
-	if err != nil {
-		switch err := err.(type) {
-		case session.SDKURLError:
-			return fmt.Sprintf("SDK URL could not be parsed: %s", err.Err)
-		case session.SoapClientError:
-			return "unable to obtain a vim client"
-		case session.UserPassLoginError:
-			return "unable to log in with username and password"
-		default:
-			return genericErrorMessage
-		}
-	}
-
-	return ""
-}
-
 func (s *server) index(res http.ResponseWriter, req *http.Request) {
 	defer trace.End(trace.Begin(""))
 	ctx := context.Background()
 	sess, err := s.getSessionFromRequest(ctx, req)
 	if err != nil {
+		log.Errorf("While loading index page got %s looking up a vSphere session", err.Error())
 		http.Redirect(res, req, "/logout", http.StatusTemporaryRedirect)
 		return
 	}
 	v := vicadmin.NewValidator(ctx, &vchConfig, sess)
-
 	if sess == nil {
 		// We're unable to connect to vSphere, so display an error message
-		errMessage := ""
-		if err != nil {
-			errMessage = fmt.Sprintf(" Error: %s", deriveErrorMessage(err))
-		}
-		vchIssues := fmt.Sprintf("<span class=\"error-message\">%s%s</span>\n",
-			v.Hostname+" is not functioning: unable to connect to vSphere. You may wish to <a href=\"/logout\">try logging in again</a>.", errMessage)
-
-		v.VCHIssues = template.HTML(vchIssues)
+		v.VCHIssues = template.HTML("<span class=\"error-message\">We're having some trouble communicating with vSphere. <a href=\"/logout\">Logging in again</a> may resolve the issue.</span>\n")
 	}
 
 	tmpl, err := template.ParseFiles("dashboard.html")
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -557,9 +557,12 @@ func (s *server) index(res http.ResponseWriter, req *http.Request) {
 
 	if sess == nil {
 		// We're unable to connect to vSphere, so display an error message
-		errMessage := deriveErrorMessage(err)
-		vchIssues := fmt.Sprintf("<span class=\"error-message\">%s Error: %s</span>\n",
-			v.Hostname+" is not functioning: unable to connect to vSphere.", errMessage)
+		errMessage := ""
+		if err != nil {
+			errMessage = fmt.Sprintf(" Error: %s", deriveErrorMessage(err))
+		}
+		vchIssues := fmt.Sprintf("<span class=\"error-message\">%s%s</span>\n",
+			v.Hostname+" is not functioning: unable to connect to vSphere. You may wish to <a href=\"/logout\">try logging in again</a>.", errMessage)
 
 		v.VCHIssues = template.HTML(vchIssues)
 	}
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -439,11 +439,7 @@ func (s *server) bundleContainerLogs(res http.ResponseWriter, req *http.Request,
 	c, err := s.getSessionFromRequest(context.Background(), req)
 	if err != nil {
 		log.Errorf("Failed to get vSphere session while bundling container logs due to error: %s", err.Error())
-		if strings.Contains(err.Error(), "Could not validate session") {
-			http.Redirect(res, req, "/logout", http.StatusTemporaryRedirect)
-			return
-		}
-		http.Error(res, genericErrorMessage, http.StatusInternalServerError)
+		http.Redirect(res, req, "/logout", http.StatusTemporaryRedirect)
 		return
 	}
 
@@ -553,7 +549,7 @@ func (s *server) index(res http.ResponseWriter, req *http.Request) {
 	defer trace.End(trace.Begin(""))
 	ctx := context.Background()
 	sess, err := s.getSessionFromRequest(ctx, req)
-	if err != nil && strings.Contains(err.Error(), "Could not validate session") {
+	if err != nil {
 		http.Redirect(res, req, "/logout", http.StatusTemporaryRedirect)
 		return
 	}
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -538,7 +538,7 @@ func deriveErrorMessage(err error) string {
 		case session.UserPassLoginError:
 			return "unable to log in with username and password"
 		default:
-			return err.Error()
+			return genericErrorMessage
 		}
 	}
 
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -443,9 +443,6 @@ func (s *server) bundleContainerLogs(res http.ResponseWriter, req *http.Request,
 		return
 	}
 
-	// Note: we don't want to Logout() until tarEntries() completes below
-	defer c.Client.Logout(context.Background())
-
 	logs, err := findDatastoreLogs(c)
 	if err != nil {
 		log.Warningf("error searching datastore: %s", err)
--- a/cmd/vicadmin/usersession.go
+++ b/cmd/vicadmin/usersession.go
@@ -86,7 +86,7 @@ func (u *UserSessionStore) VSphere(ctx context.Context, id string) (*session.Ses
 		return nil, fmt.Errorf("User session with unique ID %s does not exist", id)
 	}
 	if us.vsphere == nil {
-		return nil, fmt.Errorf("No vSphere session found for user with ID %s", id)
+		return nil, fmt.Errorf("No vSphere session found for user: %s", id)
 	}
 
 	vsphus, err := us.vsphere.SessionManager.UserSession(ctx)
--- a/cmd/vicadmin/usersession.go
+++ b/cmd/vicadmin/usersession.go
@@ -91,11 +91,12 @@ func (u *UserSessionStore) VSphere(ctx context.Context, id string) (*session.Ses
 
 	vsphus, err := us.vsphere.SessionManager.UserSession(ctx)
 	if err != nil || vsphus == nil {
-		u.Delete(id)
 		if err != nil {
-			return nil, fmt.Errorf("Failed to validate user %s session: %v", id, err)
+			log.Warnf("Failed to validate user %s session: %v", id, err)
+			return nil, nil
 		}
-		return nil, fmt.Errorf("User %s session has expired", id)
+		log.Warnf("User %s session has expired", id)
+		return nil, nil
 	}
 	log.Infof("Found vSphere session for vicadmin usersession %s", id)
 	return us.vsphere, nil
--- a/cmd/vicadmin/usersession.go
+++ b/cmd/vicadmin/usersession.go
@@ -89,6 +89,15 @@ func (u *UserSessionStore) VSphere(ctx context.Context, id string) (*session.Ses
 		return nil, fmt.Errorf("No vSphere session found. User with ID %s must authenticate again.", id)
 	}
 
+	vsphus, err := us.vsphere.SessionManager.UserSession(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("vSphere UserSession was invalid; got error: %s", err.Error())
+	}
+
+	if vsphus == nil {
+		return nil, fmt.Errorf("vSphere UserSession came back nil but no error was reported")
+	}
+
 	log.Infof("Found vSphere session for vicadmin usersession %s", id)
 	return us.vsphere, nil
 }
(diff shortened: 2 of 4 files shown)


lib/vicadmin/validate.go
```diff
@@ -130,41 +130,41 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
 		v.FirewallIssues = template.HTML("")
 		v.LicenseStatus = BadStatus
 		v.LicenseIssues = template.HTML("")
-	} else {
+		return nil
```
gigawhitlocks: This hunk looks like a bunch of changes but really all I did was add `return nil` to the end of the `if` statement and outdent the `else` block & remove the `else` keyword.

emlin: what about change to
```
if sess != nil {
   return nil
}
```

anchal-agrawal: Returning `nil` here causes a panic on https://github.com/vmware/vic/blob/master/cmd/vicadmin/server.go#L560 when the session is nil.

lib/vicadmin/validate.go
```diff
@@ -258,7 +259,7 @@ func (v *Validator) QueryDatastore(ctx context.Context, vch *config.VirtualConta
 	sort.Sort(dataStores)
 	if err != nil {
 		log.Errorf("Error while accessing datastore: %s", err)
-		return
+		return err
```
emlin: curious for what are you gonna do with this returned error?

emlin:
I think your return can only fix the panic caused by not authenticated issue, but not actually fix the problem. And potentially will miss other errors if that's not not authenticated issue happens during datastore query.

I have one suggestion to fix the issue. Cause you verified if the session creation time is longer than 24h while user session is retrieved from session cache. 
I think you can do a simply vsphere call using that session like what keepalive did in the session.go file. https://github.com/vmware/vic/blob/master/pkg/vsphere/session/session.go#L189
If you got not authenticated error, you might just return error similar to the session is already longer than 24h.
Then suppose the user will be forward to the login page.

lib/vicadmin/validate.go
gigawhitlocks: Nothing, it's thrown out. It's just used to exit the function before we start trying to access `nil` members.

lib/vicadmin/validate.go
gigawhitlocks: I'll just undo this refactor. It's not relevant to this PR.

lib/vicadmin/validate.go
```diff
@@ -247,6 +247,7 @@ func (v *Validator) QueryDatastore(ctx context.Context, vch *config.VirtualConta
 		ds, err := sess.Finder.DatastoreOrDefault(ctx, dsName)
 		if err != nil {
 			log.Errorf("Unable to collect information for datastore %s: %s", dsName, err)
+			return err
```
gigawhitlocks: true

cmd/vicadmin/usersession.go
```diff
@@ -89,6 +89,15 @@ func (u *UserSessionStore) VSphere(ctx context.Context, id string) (*session.Ses
 		return nil, fmt.Errorf("No vSphere session found. User with ID %s must authenticate again.", id)
 	}
 
+	vsphus, err := us.vsphere.SessionManager.UserSession(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("vSphere UserSession was invalid; got error: %s", err.Error())
+	}
+
+	if vsphus == nil {
+		return nil, fmt.Errorf("vSphere UserSession came back nil but no error was reported")
```
hmahmood: Is this error user visible?

gigawhitlocks: Shouldn't be but I'll make sure. I've been surprised in a few places when an error has bubbled up to the UI :\

mdubya66:
what issue is this fixing?

cmd/vicadmin/vicadm.go
```diff
@@ -483,6 +483,9 @@ func vSphereSessionGet(sessconfig *session.Config) (*session.Session, error) {
 		log.Errorf("Got %s while creating user session", err)
 		return nil, err
 	}
+	if usersession == nil {
+		return nil, fmt.Errorf("UserSession from vSphere came back nil but no error was reported")
```
dougm: Not sure why you chose this error message.. as we discussed and as is documented, UserSession returns nil, nil if the session is no longer valid.  At which point you need to `Login()` to get a new session.

https://github.com/vmware/govmomi/blob/master/session/manager.go#L133-L134

lib/vicadmin/validate.go
```diff
@@ -252,13 +255,25 @@ func (v *Validator) QueryDatastore(ctx context.Context, vch *config.VirtualConta
...
 
+	if sess.Client == nil {
+		return fmt.Errorf("Non-nil session govmomi client is nil while looking for datastore info")
+	}
+
+	if sess.Client.Client == nil {
+		return fmt.Errorf("Non-nil session vim client is nil while looking for datastore info")
+	}
+
 	pc := property.DefaultCollector(sess.Client.Client)
```
dougm: property.DefaultCollector never returns nil.

lib/vicadmin/validate.go
```diff
@@ -252,13 +255,25 @@ func (v *Validator) QueryDatastore(ctx context.Context, vch *config.VirtualConta
 		}
 	}
 
+	if sess.Client == nil {
```
dougm: Where would this get set to nil?

lib/vicadmin/validate.go
```diff
@@ -252,13 +255,25 @@ func (v *Validator) QueryDatastore(ctx context.Context, vch *config.VirtualConta
...
+		return fmt.Errorf("Non-nil session govmomi client is nil while looking for datastore info")
+	}
+
+	if sess.Client.Client == nil {
```
dougm: Where would this get set to nil?

lib/vicadmin/validate.go
```diff
@@ -252,13 +255,25 @@ func (v *Validator) QueryDatastore(ctx context.Context, vch *config.VirtualConta
...
+		return fmt.Errorf("Non-nil session vim client is nil while looking for datastore info")
+	}
+
 	pc := property.DefaultCollector(sess.Client.Client)
+	if pc == nil {
+		return fmt.Errorf("Could not get default propery collector; prop-collector came back nil")
+	}
+
 	err := pc.Retrieve(ctx, refs, nil, &dataStores)
```
dougm: As we discussed in slack, the problem happens above, when looking up datastores with an invalid session.  The error is logged above
`log.Errorf("Unable to collect information for datastore %s: %s", dsName, err)`

But continues, and `refs` passed to `pc.Retrieve` has len == 0, causing the panic.  So the log.Errorf above should either be followed by a return error or you should check for `len(refs) == 0` here and avoid calling pc.Retrieve in that case.

lib/vicadmin/validate.go
```diff
@@ -252,13 +255,25 @@ func (v *Validator) QueryDatastore(ctx context.Context, vch *config.VirtualConta
...
+	if sess.Client == nil {
+		return fmt.Errorf("Non-nil session govmomi client is nil while looking for datastore info")
+	}
+
+	if sess.Client.Client == nil {
+		return fmt.Errorf("Non-nil session vim client is nil while looking for datastore info")
+	}
+
 	pc := property.DefaultCollector(sess.Client.Client)
+	if pc == nil {
```
dougm: property.DefaultCollector never returns nil.

gigawhitlocks:
@mdubya66 trying to fix the panics seen over the weekend -- idk if an issue was reported

lib/vicadmin/validate.go
gigawhitlocks: Well I had a `return err` there originally but @emlin told me not to return err there, because there may be other datastores that are working. I misremembered and thought that the properycollector could come back nil, not `refs`

gigawhitlocks: I'll change the above nil checks to check refs for len 0 instead & return error otherwise.

mdubya66:
#3258 and #3259 @gigawhitlocks

dougm:
We haven't been using persistent connections in vicadmin.  If you switch to using a keepalive function, you'll be making the connection persistent too.  hence the suggestion to use the UserSession helper to determine if a session cookie is still valid (similar to how govc does).  When using keepalive func, there's no need to use the UserSession helper (port layer case for example)

Sent from my iPhone

On Nov 21, 2016, at 3:40 PM, Ian Whitlock <notifications@github.com<mailto:notifications@github.com>> wrote:


@gigawhitlocks commented on this pull request.

________________________________

In cmd/vicadmin/vicadm.go<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_vmware_vic_pull_3270&d=DgMCaQ&c=uilaK90D4TOVoH58JNXRgQ&r=GroxMYS-chy4r563y5jUmA&m=7EoSKTD5zMZ6n4JlE7U6v12FbEruMoDs9rxH-BUKk9g&s=Sj_r8arsZRsuphKknZs-EL_-xMr5bMgnVq2xMhuEmnI&e=>:

> @@ -483,6 +483,9 @@ func vSphereSessionGet(sessconfig *session.Config) (*session.Session, error) {
                log.Errorf("Got %s while creating user session", err)
                return nil, err
        }
+       if usersession == nil {
+               return nil, fmt.Errorf("UserSession from vSphere came back nil but no error was reported")


My understanding is that we should not need to call Login but I might need to register my own KeepAlive as the one that gets registered when I call Connect() doesn't look like it's actually getting registered because the KeepAlive value == 0, so I suspect that's what's causing the connection to go stale. And then we don't have a way, nor do we want a way, to establish a new session if the session goes stale.

-

gigawhitlocks:
Got it, thought we were using persistent connections, thanks for clarifying @dougm .
I added a check to `usersession.VSphere` to try to use the `UserSession` helper to check to make sure that is valid and I'm testing that right now.

gigawhitlocks:
Starting to get real tired tho :(

gigawhitlocks:
OK I have the functionality working the way it should, but I'm doing it by comparing error messages, which is surely slow (though only happens when there is an error at all) and brittle, so I'm going to give it a little thought and come up with something better, but the PR does resolve the issue (tested by `vic-machine create` then log into vicadmin in the browser then use `govc` OOB to delete the sessions, then load the tab again (not Refresh, which will `POST` again and actually log in :P ) and we get directed to the login page. `Redirect` sends the browser to `logout` so that we can do cleanup.

gigawhitlocks:
@mdubya66 #3259 is addressed by a single line in this change and the rest is for #3258

cmd/vicadmin/server.go
```diff
@@ -439,13 +439,14 @@ func (s *server) bundleContainerLogs(res http.ResponseWriter, req *http.Request,
 	c, err := s.getSessionFromRequest(context.Background(), req)
 	if err != nil {
 		log.Errorf("Failed to get vSphere session while bundling container logs due to error: %s", err.Error())
+		if strings.Contains(err.Error(), "Could not validate session") {
```
vburenin: is there any chance message can be localized?

cmd/vicadmin/usersession.go
```diff
@@ -86,9 +86,18 @@ func (u *UserSessionStore) VSphere(ctx context.Context, id string) (*session.Ses
 		return nil, fmt.Errorf("User session with unique ID %s does not exist", id)
 	}
 	if us.vsphere == nil {
-		return nil, fmt.Errorf("No vSphere session found. User with ID %s must authenticate again.", id)
+		return nil, fmt.Errorf("No vSphere session found for user with ID %s", id)
```
vburenin: for user: %s

cmd/vicadmin/usersession.go
```diff
@@ -86,9 +86,18 @@ func (u *UserSessionStore) VSphere(ctx context.Context, id string) (*session.Ses
...
 	}
 
+	vsphus, err := us.vsphere.SessionManager.UserSession(ctx)
+	if err != nil || vsphus == nil {
+		u.Delete(id)
+		errMsg := fmt.Sprintf("Could not validate session for user %s; must log in again", id)
```
vburenin: Why do you even try to format this message not checking err first?

```
if err != nil {
    return nil, fmt.Errorf("%s. Got error %s", errMsg, err.Error())
}
return nil,  fmt.Errorf("Could not validate session for user %s; must log in again", id)
```

gigawhitlocks: I format the message w/o the error in the case where `err` == `nil` but `vsphus` is `nil`. If err is non-nil the err is appended. That is why I format first.

lib/vicadmin/validate.go
```diff
@@ -253,12 +256,19 @@ func (v *Validator) QueryDatastore(ctx context.Context, vch *config.VirtualConta
 	}
 
 	pc := property.DefaultCollector(sess.Client.Client)
+	if pc == nil {
+		return fmt.Errorf("Could not get default propery collector; prop-collector came back nil")
+	}
+
+	if len(refs) == 0 {
```
cgtexmex: To improve readability I'd move this right after the call for references

lib/vicadmin/validate.go
```diff
@@ -253,12 +256,19 @@ func (v *Validator) QueryDatastore(ctx context.Context, vch *config.VirtualConta
...
+		return fmt.Errorf("No datastore references found")
+	}
+
 	err := pc.Retrieve(ctx, refs, nil, &dataStores)
 
 	sort.Sort(dataStores)
 	if err != nil {
```
cgtexmex: Same as previous comment -- I'd move this up beneath the call to pc.Retrieve...

lib/vicadmin/validate.go
```diff
@@ -209,7 +209,10 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
 		v.DockerPort = fmt.Sprintf("%d", opts.DefaultTLSHTTPPort)
 	}
 
-	v.QueryDatastore(ctx, vch, sess)
+	err = v.QueryDatastore(ctx, vch, sess)
+	if err != nil {
+		log.Errorf("Had a problem querying the datastores: %s", err.Error())
```
gigawhitlocks: Yeah we just want to report because we want to continue to QueryVCHStatus and we still want to return `v` down at the bottom even if this step fails.

lib/vicadmin/validate.go
```diff
@@ -253,12 +256,19 @@ func (v *Validator) QueryDatastore(ctx context.Context, vch *config.VirtualConta
...
+	}
+
 	err := pc.Retrieve(ctx, refs, nil, &dataStores)
 
 	sort.Sort(dataStores)
 	if err != nil {
 		log.Errorf("Error while accessing datastore: %s", err)
-		return
 	}
 	for _, ds := range dataStores {
```
gigawhitlocks: Yeah, we show that it was inaccessible in the UI

cmd/vicadmin/usersession.go
vburenin: My point is that overall code organization could be better:
okay, let's reorganize it better:
```
if err != nil {
    return nil, fmt.Errorf("Failed to validate user %s session: %v", id, err)
}
return nil, fmt.Errorf("User %s session has expired", id)
```

lib/vicadmin/validate.go
```diff
@@ -252,13 +255,19 @@ func (v *Validator) QueryDatastore(ctx context.Context, vch *config.VirtualConta
...
+		return fmt.Errorf("No datastore references found")
+	}
+
 	pc := property.DefaultCollector(sess.Client.Client)
-	err := pc.Retrieve(ctx, refs, nil, &dataStores)
+	if pc == nil {
+		return fmt.Errorf("Could not get default propery collector; prop-collector came back nil")
+	}
 
 	sort.Sort(dataStores)
```
cgtexmex: this needs to be after the pc.retrieve or the sort won't work...

emlin:
The session valid checking is in getSessionFromRequest, and error is handled in index and bundleContainerLogs. So if the vsphere session is timeout, and if user access the <ip:2378>/ or the bundle log ip address, they will be redirected to login page.
But if user type portlayer log, persona log address directly, they don't need to relogin. That page will be available for 24hours.

You might test it with removed vsphere active session.

I think we're using vsphere account to login, and manage user session, we should make all URLs access same for session management. So idea is still to change the isAuthenticated method, to make sure access for any URLs rely on this session check.
@dougm @hickeng 

Another benefit is that will centralize all session related management and login/logout redirect logic. (my 2 cents)

gigawhitlocks:
> But if user type portlayer log, persona log address directly, they don't need to relogin. That page will be available for 24hours.

This was the desired behavior requested by @hickeng and is not related to this PR as it is not changed by this PR

>I think we're using vsphere account to login, and manage user session, we should make all URLs access same for session management. So idea is still to change the isAuthenticated method, to make sure access for any URLs rely on this session check.

This was the behavior I implemented and then  @anchal-agrawal was told to change it to the current behavior.

lib/vicadmin/validate.go
gigawhitlocks: oh, yeah, you're right, moving stuff around I definitely confused myself.

cmd/vicadmin/usersession.go
gigawhitlocks: oh I see

cgtexmex:
lgtm

cmd/vicadmin/server.go
```diff
@@ -551,6 +548,10 @@ func (s *server) index(res http.ResponseWriter, req *http.Request) {
 	defer trace.End(trace.Begin(""))
 	ctx := context.Background()
 	sess, err := s.getSessionFromRequest(ctx, req)
+	if err != nil {
+		http.Redirect(res, req, "/logout", http.StatusTemporaryRedirect)
+		return
```
anchal-agrawal: My understanding is that if `getSessionFromRequest` returns an error here, the `if sess == nil` block below will no longer get executed and the degraded dashboard will not kick in. Just wanted to check if that's the intended behavior or not.

Also, while doing a quick test with your branch,  if I set `sess = nil` explicitly right below the `sess, err := s.getSessionFromRequest(ctx, req)` line, I get too many redirects on the browser. I'll dig deeper to see what's up.

anchal-agrawal: Sidenote - while I was implementing the "degraded" dashboard behavior, I ran into this exact redirect issue upon redirecting the user to the logout handler if the session was nil.

cmd/vicadmin/server.go
hickeng: @vburenin if that's coming from vSphere, then yes. If it's from govmomi I don't think so.
@gigawhitlocks What's the real impact of redirecting to logout (and handling the case where you're not actually logged in) for ALL errors? I'm not seeing the alternative path added much value at this time.

cmd/vicadmin/vicadm.go
hickeng: @gigawhitlocks the only case where we're aiming for a degraded dashboard is if vic-admin cannot connect to vSphere to validate supplied credentials and the only thing we need to report at that time is that we cannot connect to vsphere.

The only reason to do this via dashboard was to provide some additional diagnostics as to _why_ we cannot connect but if it's causing issues then a message similar to that for internal server error suffices to let the user know why they cannot authenticate.

lib/vicadmin/validate.go
hickeng: @gigawhitlocks Should this check go inside the block where we've confirmed that we can connect to vSphere along with the firewall checks?

anchal-agrawal:
LGTM :+1:

gigawhitlocks:
♻️

PR #3271, opened 2016-11-21
Connect file backed serial ports after a vMotion

### LINKED ISSUE #3243: vMotion disconnects serial ports after a motion
vMotion disconnects file backed serial ports after a migration. This means that while applications will continue to function correctly without interruption, container logs will lose output after a vMotion.

caglar10ur:
This is unfortunately not the ideal solution but a workaround.

Logging subsystem now listens VmMigratedEvent and toggles the file
backed serial ports after getting triggered via event.

The underlying ESXi issue is now fixed but it is not clear whether it is
going to be backported so in the meantime this is the best we can do.

Changes since v1:

Portlayer now connects the file backed serial ports after a restart

Fixes #3243

### DIFF
--- a/lib/portlayer/logging/logging.go
+++ b/lib/portlayer/logging/logging.go
@@ -15,12 +15,115 @@
 package logging
 
 import (
+	"context"
 	"fmt"
+	"sync"
 
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/lib/portlayer/event/collector/vsphere"
+	"github.com/vmware/vic/lib/portlayer/event/events"
 	"github.com/vmware/vic/lib/portlayer/exec"
+	"github.com/vmware/vic/pkg/retry"
 	"github.com/vmware/vic/pkg/trace"
 )
+
+var once sync.Once
+
+func Init(ctx context.Context) error {
+	once.Do(func() {
+		// Subscribe to vm events
+		exec.Config.EventManager.Subscribe(
+			events.NewEventType(vsphere.VMEvent{}).Topic(),
+			"logging",
+			func(ie events.Event) {
+				eventCallback(ctx, ie)
+			})
+	})
+	return nil
+}
+
+// listens migrated events and connects the file backed serial ports
+func eventCallback(ctx context.Context, ie events.Event) {
+	defer trace.End(trace.Begin(""))
+
+	var err error
+
+	switch ie.String() {
+	case events.ContainerMigrated,
+		events.ContainerMigratedByDrs:
+		log.Debugf("Received %s event", ie)
+
+		// grab the container from the cache
+		container := exec.Containers.Container(ie.Reference())
+		if container == nil {
+			log.Errorf("Container %s not found", ie.Reference())
+			return
+		}
+
+		operation := func() error {
+			handle := container.NewHandle(ctx)
+			if handle == nil {
+				log.Errorf("Handle for %s cannot be created", ie.Reference())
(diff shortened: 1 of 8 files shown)


mdubya66:
please wait until after v0.8.0 GA to merge this

lib/portlayer/logging/logging.go
```diff
@@ -15,13 +15,123 @@
...
+	"github.com/vmware/vic/pkg/backoff"
 	"github.com/vmware/vic/pkg/trace"
 )
 
+var once sync.Once
+
+func Init(ctx context.Context) error {
+	once.Do(func() {
+		// Subscribe to vm events
+		exec.Config.EventManager.Subscribe(
```
cgtexmex: Currently vic has a very simplistic event handler, so we are unable to subscribe by specific event type -- say *VmGuestShutdownEvent* or *VmPoweredOnEvent* -- so we subscribe for all VM events.  We do this for the exec layer here: https://github.com/vmware/vic/blob/master/lib/portlayer/exec/exec.go#L101 which results in a call back to: https://github.com/vmware/vic/blob/master/lib/portlayer/exec/exec.go#L129 --> that func could be refactored to handle these events as well and just callout to this logging eventCallBack.  This would reduce the number of times that event is passed around.

You can keep this separate subscription if you like but then we'll have 3 different callbacks for VM Events that will fire for all of the events identified in the vsphere/collector.go (https://github.com/vmware/vic/blob/master/lib/portlayer/event/collector/vsphere/collector.go#L164) -- IMO causing undo noise / logging /etc..

I prefer the separation of concerns w/this approach but given the simplistic event handler my approach would be to refactor the existing exec.eventCallback

my $.02

caglar10ur:
Confirmed that locking problem still exists with 6.5 using vsan

mdubya66:
Approved for Merge

mdubya66:
LGTM

caglar10ur:
The underlying issue is now fixed in ESXi - serial/parallel ports are no longer require special treatment and works after vmotion. We still need to deal with it for the older releases so will refresh this while addressing Clint's comments

lib/portlayer/logging/logging.go
caglar10ur: Tried it but failed because of the import cycles. Don't want to put the serial related code into the exec and exec can't use other portlayer subsystems because of the exec.Handle usage.

lib/portlayer/logging/logging.go
```diff
@@ -15,13 +15,116 @@
...
+		}
+
+		if err := retry.Do(operation, exec.IsConcurrentAccessError); err != nil {
+			log.Errorf("Multiple attempts failed to commit handle after getting %s event for container %s: %s", ie, ie.Reference(), err)
+		}
+	}
+
+}
+
+func toggle(handle *exec.Handle, connected bool) (*exec.Handle, error) {
```
hmahmood: `connected` doesn't seem to be used anywhere.

cgtexmex:
lgtm

hmahmood:
lgtm

caglar10ur:
♻️

lib/portlayer/portlayer.go
```diff
@@ -86,17 +89,21 @@ func Init(ctx context.Context, sess *session.Session) error {
 		return err
 	}
 
+	if err = logging.Init(ctx); err != nil {
+		return err
+	}
+
 	// Unbind containerVM serial ports configured with the old VCH IP.
 	// Useful when the appliance restarts and the VCH has a different IP.
```
anchal-agrawal: We should update this comment as well, now that we've changed `unbindSerialPorts` to `TakeCareOfSerialPorts`.

PR #3275, opened 2016-11-22
Make multiple passes to log browser

### LINKED ISSUE #3274: vpxd.log truncated at 500 lines
This prevents us from being able to diagnose issues when vpxd.log is required to understand what is going on. This is caused by vpxd limiting the amount of log data it will return per call to `BrowseDiagnosticLog` to 500 lines. We need to update vicadmin to make multiple calls to the API in order to retrieve an adequate amount of lines.

jzt:
This ups the number of log lines gathered to 5000 (up from 1000, 500 for vpxd.log).

Fixes #3274

### DIFF
--- a/cmd/vicadmin/logs.go
+++ b/cmd/vicadmin/logs.go
@@ -90,7 +90,7 @@ func (r dlogReader) open() (entry, error) {
 
 		for _, line := range h.LineText {
 			buf.WriteString(line)
-			buf.WriteString("\n")
+			buf.WriteByte('\n')
 		}
 
 		start += lines
--- a/cmd/vicadmin/logs.go
+++ b/cmd/vicadmin/logs.go
@@ -45,6 +45,11 @@ const (
 	nBytes    = 1024
 	tailLines = 8
 	uint32max = (1 << 32) - 1
+
+	// how many lines of log data to collect
+	logLines = 5000
+	// how many lines to request per call
+	lines = 500
 )
 
 type dlogReader struct {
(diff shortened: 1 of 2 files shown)


cmd/vicadmin/logs.go
```diff
@@ -73,19 +78,22 @@ func (r dlogReader) open() (entry, error) {
...
 	var buf bytes.Buffer
+	for start < end {
+		h, err = m.BrowseLog(ctx, r.host, r.name, start, lines)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, line := range h.LineText {
+			buf.WriteString(line)
+			buf.WriteString("\n")
```
vburenin: WriteByte?

cmd/vicadmin/logs.go
```diff
@@ -73,19 +78,22 @@ func (r dlogReader) open() (entry, error) {
...
 	var buf bytes.Buffer
+	for start < end {
+		h, err = m.BrowseLog(ctx, r.host, r.name, start, lines)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, line := range h.LineText {
+			buf.WriteString(line)
+			buf.WriteByte(0x0a)
```
vburenin: I would you a different form for byte encoding. If you use single quotes it means byte. So, '\n' is enough. Otherwise it might be a little confusing for people who do not remember ASCII table.

vburenin:
Just a minor readability comment needs to be addressed, other than that code looks good.
LGTM

cgtexmex:
lgtm

caglar10ur:
LGTM

Issue #2163, opened 2016-08-28
vic-machine: remove timeout as a default behaviour


hickeng:
#1557 is "closed" in that we now provide a more helpful error message in the case where we time out, however that's not exactly addressing the root behaviour.

The primary reason for a timeout is to preserve liveliness in the overall system - in the case of `vic-machine` this basically means when it's being called as part of an automated workflow.

The original reason for the timeout was to 'detect' a failure during appliance initialization as we have no channel by which to determine progress. That's no longer the case for other than for a very small portion of the appliance initialization (in between power on and OS init).

The time for appliance boot and initialization is heavily dependent on the underlying hardware and load levels. The time for ISO upload is highly dependent on bandwidth between client machine and vSphere endpoint.

We should remove the default timeout and add liveliness detection to the stages:
a. ISO upload
b. appliance initialization

Even if we do timeout during (b) it's quite possible that the appliance will initialize if left alone in a busy system, so we should add a marker that `vic-machine ls` and `inspect` can use to determine whether the appliance has ever initialized successfully since it was last configured, and whether it's currently initializing.

mdubya66:
Moving to Icebox per 9/20 Triage

mlh78750:
Related https://github.com/vmware/vic/issues/2794#issuecomment-254868808

mdubya66:
we are seeing this more and more. prioritizing for v0.9.0.

mdubya66:
Almost all customers have run into this issue. Adding customer found.

emlin:
Need to be done for this issue
- [ ] Remove default timeout in vic-machine create
- [ ] Add uploading progress bar for image upload and appliance initialization
- [ ] Mention that the appliance information could be found through vic-machine inspect, and with that to go ahead with additional debug steps

The maker of appliance initialization, based on the experience of vSphere API check from vic-machine, I assume this function should be done together with vSphere API check in portlayer. #3479

PR #3276, opened 2016-11-22
Partial fix vic-machine create timeout issue

emlin:
Partial fix for #2163 
This fix removes timeout for all vm operations and image upload. Timeout is only used to check if service is available or not. Default timeout is still 3 minutes.

There is no cancel available as well. User can ctrl+c to stop the installation at any time. And use vic-machine delete to delete partial created vch.
No progress bar to show image upload percentage.

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -446,7 +446,7 @@ func (c *Create) Flags() []cli.Flag {
 		cli.DurationFlag{
 			Name:        "timeout",
 			Value:       3 * time.Minute,
-			Usage:       "Time to wait for create",
+			Usage:       "Time to wait for service ready",
 			Destination: &c.Timeout,
 		},
 	}
@@ -1259,8 +1259,9 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
 
 	// after create appliance, use timeout context to wait ip address
 	timeoutctx, cancel := context.WithTimeout(context.Background(), c.Timeout)
+	defer cancel()
 	defer func() {
-		if timeoutctx != nil && timeoutctx.Err() != nil && timeoutctx.Err() == context.DeadlineExceeded {
+		if timeoutctx.Err() != nil && timeoutctx.Err() == context.DeadlineExceeded {
 			//context deadline exceeded, replace returned error message
 			err = errors.Errorf("Create timed out: if slow connection, increase timeout with --timeout")
 		}
@@ -1272,8 +1273,6 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
 		cancel()
 		return err
 	}
-	timeoutctx = nil
-	cancel()
 
 	// vic-init will try to reach out to the vSphere target.
 	log.Info("Checking VCH connectivity with vSphere target")
@@ -1284,10 +1283,6 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
 		log.Error(err)
 		return errors.New("Running diagnostics failed.")
 	}
-
-	// after create appliance, use new timeout context to check service availability
-	timeoutctx, cancel = context.WithTimeout(context.Background(), c.Timeout)
-	defer cancel()
 
 	// Checking access to vSphere API
 	cd, err := executor.CheckAccessToVCAPI(timeoutctx, vch, vchConfig.Target)
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -1263,11 +1263,11 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
 	defer func() {
 		if timeoutctx.Err() == context.DeadlineExceeded {
 			//context deadline exceeded, replace returned error message
-			err = errors.Errorf("Creating VCH exceeded time limit of %s. Please, increase timeout using --time in case of slow network bandwidth or busy vSphere target.", c.Timeout.String())
+			err = errors.Errorf("Creating VCH exceeded time limit of %s. Please increase the timeout using --timeout to accommodate for a slow network connection or a busy vSphere target", c.Timeout.String())
 		}
 	}()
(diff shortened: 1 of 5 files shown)


cmd/vic-machine/create/create.go
```diff
@@ -1257,12 +1250,30 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
...
 		executor.CollectDiagnosticLogs()
 		log.Error(err)
 		return err
 	}
 
+	// after create appliance, use timeout context to wait ip address
+	timeoutctx, cancel := context.WithTimeout(context.Background(), c.Timeout)
+	defer cancel()
+	defer func() {
+		if timeoutctx.Err() != nil && timeoutctx.Err() == context.DeadlineExceeded {
```
vburenin: Does it even make sense to compare it with nil?

cmd/vic-machine/create/create.go
```diff
@@ -1257,12 +1250,30 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
...
+			//context deadline exceeded, replace returned error message
+			err = errors.Errorf("Create timed out: if slow connection, increase timeout with --timeout")
```
vburenin: Displaying the number of seconds for timeout would be very useful, since user may not have any idea what the default timeout is.
Also, If I was a user, I would be happier to see message like this. 

Creating VCH exceeded time limit of %s seconds. Please, increase timeout using --time in case of slow network bandwidth or busy vSphere target.

lib/install/management/appliance.go
```diff
@@ -911,21 +911,21 @@ func (d *Dispatcher) CheckDockerAPI(conf *config.VirtualContainerHostConfigSpec,
 }
 
 // ensureApplianceInitializes checks if the appliance component processes are launched correctly
-func (d *Dispatcher) ensureApplianceInitializes(conf *config.VirtualContainerHostConfigSpec) error {
+func (d *Dispatcher) EnsureApplianceInitializes(ctx context.Context, conf *config.VirtualContainerHostConfigSpec) error {
```
vburenin: I would name it as: InsureApplicaneInit

cmd/vic-machine/create/create.go
```diff
@@ -1257,12 +1250,30 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
...
+		}
+	}()
+
+	if err = executor.EnsureApplianceInitializes(timeoutctx, vchConfig); err != nil {
+		executor.CollectDiagnosticLogs()
+		log.Error(err)
+		cancel()
```
emlin: left over. thanks for find it

lib/install/management/appliance.go
```diff
@@ -911,21 +911,21 @@ func (d *Dispatcher) CheckDockerAPI(conf *config.VirtualContainerHostConfigSpec,
 }
 
 // ensureApplianceInitializes checks if the appliance component processes are launched correctly
-func (d *Dispatcher) ensureApplianceInitializes(conf *config.VirtualContainerHostConfigSpec) error {
+func (d *Dispatcher) InsureApplicaneInit(ctx context.Context, conf *config.VirtualContainerHostConfigSpec) error {
```
anchal-agrawal: Apologies Emma, but there are typos in the name `InsureApplicaneInit`. We should either revert to `EnsureApplianceInitializes` or change it to `EnsureApplianceInit` per @vburenin's suggestion.

cmd/vic-machine/create/create.go
```diff
@@ -1257,12 +1250,29 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
...
 		return err
 	}
 
+	// after create appliance, use timeout context to wait ip address
+	timeoutctx, cancel := context.WithTimeout(context.Background(), c.Timeout)
+	defer cancel()
+	defer func() {
+		if timeoutctx.Err() == context.DeadlineExceeded {
+			//context deadline exceeded, replace returned error message
+			err = errors.Errorf("Creating VCH exceeded time limit of %s. Please, increase timeout using --time in case of slow network bandwidth or busy vSphere target.", c.Timeout.String())
```
mlh78750: no comma after Please
agree with @andrewtchin it should be --timeout

Maybe this for the text:
Please increase the timeout using --timeout to accommodate for a slow network connection or a busy vSphere target.

cmd/vic-machine/create/create.go
```diff
@@ -1257,12 +1250,29 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
...
+		}
+	}()
+
+	if err = executor.InsureApplicaneInit(timeoutctx, vchConfig); err != nil {
```
mlh78750: Ensure in Insure

lib/install/management/appliance.go
emlin: ah.. my english...

cmd/vic-machine/create/create.go
emlin: sure

andrewtchin:
lgtm

cmd/vic-machine/create/create.go
```diff
@@ -1257,12 +1250,29 @@ func (c *Create) Run(cliContext *cli.Context) (err error) {
...
 		return err
 	}
 
+	// after create appliance, use timeout context to wait ip address
+	timeoutctx, cancel := context.WithTimeout(context.Background(), c.Timeout)
+	defer cancel()
+	defer func() {
+		if timeoutctx.Err() == context.DeadlineExceeded {
+			//context deadline exceeded, replace returned error message
+			err = errors.Errorf("Creating VCH exceeded time limit of %s. Please increase the timeout using --timeout to accommodate for a slow network connection or a busy vSphere target", c.Timeout.String())
```
hickeng: Slow network should be irrelevant here as it's after the ISO upload.

PR #3290, opened 2016-11-22
Remove keys and command args/env from logging

### LINKED ISSUE #3287: Sanitize logs: remove PPK elements from logs
Was included in #2849 but that's deferred. This is only for preventing the key from being logged.

hickeng:
Removes potentially sensitive values from logging unless debug >= 3
This includes the generated keys and arguments and environment variables
for the commands run in containers.

The full set of data is still in vmware.log but is at least contained.

Fixes #3287


andrewtchin:
lgtm

vburenin:
LGTM

hickeng:
poke

PR #3294, opened 2016-11-23
[full ci] WIP Clean up the certificates after the VCH is deleted

chengwang86:
This resolves the issue in build 7195 where a VCH is reusing the name of a prior VCH.

To reproduce and verify this, 

1. bin/vic-machine-linux create --target root:pwd@192.168.60.162 --name **cheng-30** --public-network-ip 192.168.60.**130** --public-network-gateway 192.168.60.2/24 --dns-server 192.168.60.2 --tls-cname 192.168.60.**130** --thumbprint=...

Or:

bin/vic-machine-linux create --target root:pwd@192.168.60.162 --name **cheng-30**  --thumbprint=...

Then the VCH cheng-30 would be created.

2. bin/vic-machine-linux delete --force --target root:1986911A@192.168.60.162 --name cheng-30  --thumbprint=...

3. bin/vic-machine-linux create --target root:pwd@192.168.60.162 --name **cheng-30** --public-network-ip 192.168.60.**140** --public-network-gateway 192.168.60.2/24 --dns-server 192.168.60.2 --tls-cname 192.168.60.**140** --thumbprint=...
```
INFO[2016-11-22T20:23:13-08:00] ### Installing VCH ####                      
ERRO[2016-11-22T20:23:13-08:00] Provided cname does not match that in existing server certificate: 192.168.60.130 
ERRO[2016-11-22T20:23:13-08:00] Unable to load certificates: cname option doesn't match existing server certificate in certificate path cheng-30 
ERRO[2016-11-22T20:23:13-08:00] --------------------                         
ERRO[2016-11-22T20:23:13-08:00] vic-machine-linux failed: cname option doesn't match existing server certificate in certificate path cheng-30
```

4. rm -rf cheng-30

5. repeat step 3.

```
Installer completed successfully
```

### DIFF
--- a/tests/resources/Util.robot
+++ b/tests/resources/Util.robot
@@ -178,10 +178,8 @@ Run VIC Machine Delete Command
     Wait Until Keyword Succeeds  6x  5s  Check Delete Success  %{VCH-NAME}
     Should Be Equal As Integers  ${rc}  0
     Should Contain  ${output}  Completed successfully
-    ${output}=  Run  rm -f %{VCH-NAME}-*.pem
+    ${output}=  Run  rm -rf %{VCH-NAME}
     [Return]  ${output}
-    Run  rm -f ${vch-name}/*
-    Run  rmdir ${vch-name}
 
 Cleanup Datastore On Test Server
     ${out}=  Run  govc datastore.ls
--- a/tests/resources/Util.robot
+++ b/tests/resources/Util.robot
@@ -180,8 +180,6 @@ Run VIC Machine Delete Command
     Should Contain  ${output}  Completed successfully
     ${output}=  Run  rm -rf %{VCH-NAME}
     [Return]  ${output}
-    Run  rm -f ${vch-name}/*
-    Run  rmdir ${vch-name}
 
 Cleanup Datastore On Test Server
     ${out}=  Run  govc datastore.ls
--- a/tests/resources/Util.robot
+++ b/tests/resources/Util.robot
@@ -180,6 +180,8 @@ Run VIC Machine Delete Command
     Should Contain  ${output}  Completed successfully
     ${output}=  Run  rm -rf %{VCH-NAME}
     [Return]  ${output}
+    Run  rm -f ${vch-name}/*
+    Run  rmdir ${vch-name}
 
 Cleanup Datastore On Test Server
     ${out}=  Run  govc datastore.ls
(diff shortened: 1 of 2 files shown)


tests/resources/Util.robot
```diff
@@ -177,6 +177,8 @@ Run VIC Machine Delete Command
     Should Be Equal As Integers  ${rc}  0
     Should Contain  ${output}  Completed successfully
     [Return]  ${output}
+    Run  rm -f ${vch-name}/*
```
anchal-agrawal: These should probably go before the `[Return]` statement.

chengwang86: I think the order does not matter. It does remove the particular folder.

anchal-agrawal: I was talking about putting the `rm` commands before the `[Return]`. Anyway, I checked; turns out `[Return]` doesn't necessarily have to be the last command in the keyword.

mhagen-vmware: Prefer just a:
Run  rm -rf ${vch-name}

don't need both and whether it works or not, it should be above the return statement, actually above the should be equals, so as to limit the possibility of it not happening due to vic-machine problem.

chengwang86:
@hickeng @mhagen-vmware 

This fix is in contradiction with " Test 9-03 - VICAdmin Log Failed Attempts" [https://github.com/vmware/vic/blob/master/tests/test-cases/Group9-VIC-Admin/9-03-VICAdmin-Log-Failed-Attempts.robot#L18](url)

In test case 9-03, we want to check if the old-certs still exists after removing the old VCH, whereas this PR deletes the old-certs right after removing the VCH. So test case 9-03 fails in build 7206.

We may have to block this PR if we don't want to re-design 9-03 at this time.

mdubya66:
I'm not approving this change yet. It needs further thought and we need to understand why the collision is happening.

mdubya66:
Is this PR still relevant?

chengwang86:
@mdubya66 Based on my discussion with @hickeng , since this PR only changes the test but not the vic code, it would not affect the functionality of our product and we should have it. 

For the conflict with 9-03, I will try either of the following and see if it passes the ci:

- generate a new incorrect certificate to use
- copy the other cert before deletion

mhagen-vmware:
lgtm

hickeng:
LGTM

chengwang86:
ping

PR #3306, opened 2016-11-23
Fix cert-path handling, map cname from client/public

hickeng:
Fix cert-path use so that absolute paths are viable.
* Fix help usage for the x-network-ip options
  - the netmask is ignored in processing and breaks the use as a cname
* Follow up to forcing that shared interface config be done via public-network
  - static IP used for cname can come from public-network-ip as well as client
* Debug message improvement

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -34,6 +34,8 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/urfave/cli"
 
+	"path/filepath"
+
 	"github.com/vmware/vic/lib/install/data"
 	"github.com/vmware/vic/lib/install/management"
 	"github.com/vmware/vic/lib/install/validate"
@@ -218,7 +220,7 @@ func (c *Create) Flags() []cli.Flag {
 		cli.StringFlag{
 			Name:        "client-network-ip",
 			Value:       "",
-			Usage:       "IP address for the VCH on the client network, e.g. 10.0.0.2/24",
+			Usage:       "IP address for the VCH on the client network, e.g. 10.0.0.2",
 			Destination: &c.clientNetworkIP,
 			Hidden:      true,
 		},
@@ -240,7 +242,7 @@ func (c *Create) Flags() []cli.Flag {
 		cli.StringFlag{
 			Name:        "public-network-ip",
 			Value:       "",
-			Usage:       "IP address for the VCH on the public network, e.g. 10.0.1.2/24",
+			Usage:       "IP address for the VCH on the public network, e.g. 10.0.1.2",
 			Destination: &c.publicNetworkIP,
 			Hidden:      true,
 		},
@@ -262,7 +264,7 @@ func (c *Create) Flags() []cli.Flag {
 		cli.StringFlag{
 			Name:        "management-network-ip",
 			Value:       "",
-			Usage:       "IP address for the VCH on the management network, e.g. 10.0.2.2/24",
+			Usage:       "IP address for the VCH on the management network, e.g. 10.0.2.2",
 			Destination: &c.managementNetworkIP,
 			Hidden:      true,
 		},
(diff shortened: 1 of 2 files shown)


mdubya66:
required to GA. approved after code reviews

cmd/vic-machine/create/create.go
```diff
@@ -931,9 +936,11 @@ func (c *Create) loadCertificates() ([]byte, *certificate.KeyPair, error) {
 
 	var keypair *certificate.KeyPair
 	// default names
-	skey := fmt.Sprintf("./%s/%s", c.certPath, serverKey)
-	scert := fmt.Sprintf("./%s/%s", c.certPath, serverCert)
-	ca := fmt.Sprintf("./%s/%s", c.certPath, caCert)
+	skey := fmt.Sprintf("%s/%s", c.certPath, serverKey)
```
vburenin: filepath.Join

cmd/vic-machine/create/create.go
```diff
@@ -610,9 +610,14 @@ func (c *Create) processCertificates() error {
...
 
 	// if we've not got a specific CommonName but do have a static IP then go with that.
-	if c.cname == "" && c.clientNetworkIP != "" {
-		c.cname = c.clientNetworkIP
-		log.Infof("Using client-network-ip as cname where needed - use --tls-cname to override: %s", c.cname)
+	if c.cname == "" {
+		if c.clientNetworkIP != "" {
+			c.cname = c.clientNetworkIP
+			log.Infof("Using client-network-ip as cname where needed - use --tls-cname to override: %s", c.cname)
+		} else if c.publicNetworkIP != "" && (c.publicNetworkName == c.clientNetworkName || c.clientNetworkName == "") {
```
hmahmood: What if the management ip is set only? The configuration are shared between networks if the port group is shared, so that should be reflected here as well.

cmd/vic-machine/create/create.go
vburenin: is fmt.Sprintf even needed here?

cmd/vic-machine/create/create.go
```diff
@@ -1062,7 +1069,7 @@ func (c *Create) generateCertificates(server bool, client bool) ([]byte, *certif
 
 	var certs []byte
 	// generate the certs and keys with names conforming the default the docker client expects
-	err := os.MkdirAll(fmt.Sprintf("./%s", c.certPath), 0700)
+	err := os.MkdirAll(fmt.Sprintf("%s", c.certPath), 0700)
```
hmahmood: So is the certPath absolute now? Where this directory gets created may not be predictable if its not.

hickeng: nope - good spot.

hmahmood:
lgtm

fdawg4l:
lgtm

pkg/certificate/certificate.go
```diff
@@ -115,12 +115,18 @@ func templateAsClientOnly(template *x509.Certificate) *x509.Certificate {
...
 	template.ExtKeyUsage = append(template.ExtKeyUsage, x509.ExtKeyUsageServerAuth)
 
-	// the specified domain is always the primary
 	template.Subject.CommonName = domain
 
 	// abide by the spec - if CN is an IP, put it in the subjectAltName as well
 	ip := net.ParseIP(domain)
+	if ip == nil {
+		// see if CIDR works
+		ip, _, _ = net.ParseCIDR(domain)
```
fdawg4l: domain is never empty, right?

hickeng: It can be "" in which case ip will be nil and you end up with a cert that has an empty CommonName field - same as if you didn't call this path at all other than the additional usage ExtKeyUsageServerAuth

hickeng:
poke

cmd/vic-machine/create/create.go
```diff
@@ -34,6 +34,8 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/urfave/cli"
 
+	"path/filepath"
```
vburenin: oh... misformatted due to goimports issue.

Issue #3397, opened 2016-11-29
Datastore access to VM logs limited to VM host when using shared storage


mhagen-vmware:
VM files open for writing such as vmware.log, output.log and tether.debug can only accessed by the ESX host where the VM is currently running.  In the case of shared storage, there can be multiple hosts with access to the Datastore - we must use a service ticket for the host where the VM is running to download these files.

sflxn:
Apparently, docker logs isn't returning the output.

mhagen-vmware:
Killing the command, causes the output to show up.   This appears to be related to VC, as on the CI this test passes.

sflxn:
After more investigation, here's what I found on my local setup:

1) I can reproduce this on my nested VC 6 + vSan setup
2) the logs is there on VC if I ssh into it and look for it
3) I cannot reproduce this on ESXi 6.

mdubya66:
I could not reproduce this on 6.5 VC without VSAN.
```
(mwilliamson) dubs:~/Downloads/vic_0.8.0 $ docker -H 10.160.177.175:2375 run -d busybox /bin/top
Unable to find image 'busybox:latest' locally
Pulling from library/busybox
56bec22e3559: Pull complete
a3ed95caeb02: Pull complete
Digest: sha256:98a0bd48d22ff96ca23bfda2fe1cf72034ea803bd79e64a5a5f274aca0f9c51c
Status: Downloaded newer image for library/busybox:latest
a09a51e4eba4c9420f587d97cf965b36296786716ad695787d3a6e08d3f85d62
(mwilliamson) dubs:~/Downloads/vic_0.8.0 $ docker -H 10.160.177.175:2375 ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
a09a51e4eba4        busybox             "/bin/top"          42 seconds ago      Up 40 seconds                           cocky_cray
(mwilliamson) dubs:~/Downloads/vic_0.8.0 $ docker -H 10.160.177.175:2375 logs a
Mem: 62696K used, 1990288K free, 11892K shrd, 576K buff, 26348K cached
CPU:  0.0% usr  0.0% sys  0.0% nic  100% idle  0.0% io  0.0% irq  0.0% sirq
Load average: 0.00 0.00 0.00 1/71 220
  PID  PPID USER     STAT   VSZ %VSZ CPU %CPU COMMAND
    1     0 root     S     332m 16.5   1  0.0 /.tether/tether
[...]

PR #3406, opened 2016-11-30
Fix docker run tests to work with VC better

mhagen-vmware:
Doesn't work currently due to #3397 but these changes are needed nonetheless as VC+VSAN is significantly slower than ESXi


andrewtchin:
lgtm

rajanashok:
lgtm

mhagen-vmware:
.

PR #3408, opened 2016-11-30
Encrypt pivate key for host certificate [full ci]

### LINKED ISSUE #2630: Hide secrets from read-only API

hickeng:
This encrypts the private key for the endpoint VM server certificate.
Also corrects missing --tlscacert value when loading certificates in
docker client example output during create.

Fixes #2630


fdawg4l:
lgtm

caglar10ur:
LGTM :sheep: :it:

dougm:
lgtm

mdubya66:
This is approved as it is a security requirement from our review and will be stop ship.

hickeng:
poke

hickeng:
poke

PR #3411, opened 2016-11-30
Strip network mask from IP if it is used as cname.

### LINKED ISSUE #3410: Strip network mask from IP address for cert check.
At the moment if --tls-cname is not specified vic-machine automatically takes clientNetworkIP/publicNetworkIP/managementNetworkIP as a value for cname. Unfortunately, it also means that it will automatically include a network mask that is not a part of a host name.
If user run this command second time with pre-generated certificates, it is going to fail since ip/mask is not equal IP.

vburenin:
Fixes #3410

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -612,26 +612,20 @@ func (c *Create) processCertificates() error {
 	// if we've not got a specific CommonName but do have a static IP then go with that.
 	if c.cname == "" {
 		if c.clientNetworkIP != "" {
-			if cnameIP, _, _ := net.ParseCIDR(c.clientNetworkIP); cnameIP != nil {
-				c.cname = cnameIP.String()
-			} else {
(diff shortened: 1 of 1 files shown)


hickeng:
There's no point in adding this code - at the moment the mask portion is ignored on the `--xxx-network-ip` options so simply not providing the /xx component of the value is an equivalent fix.

This should be delivered with an update that stops requiring the mask on the gateway argument and instead pulls it from the address argument.

vburenin:
@hickeng If mask is not supplied with IP address, how do we determine a subnet?

mdubya66:
Given the comments is this PR still relevant?

mlh78750:
agree with https://github.com/vmware/vic/pull/3411#issuecomment-263973505 We should not use /XX on the gateway and instead have it on the ip option.

vburenin:
Fixed in #3519

PR #3423, opened 2016-12-01
Update vendor/govmomi to release 0.12 [full ci]

### LINKED ISSUE #3362: Use vCenter for datastore HTTP access
Prior to #2993, with the use of LoginExtensionByCertificate vic required direct ESX host access for datastore HTTP access.  With this requirement removed, we can switch to vCenter for datastore HTTP access.

dougm:
Fixes #3362
Towards #3397

### DIFF
--- a/cmd/vicadmin/vicadm.go
+++ b/cmd/vicadmin/vicadm.go
@@ -434,7 +434,9 @@ func (r datastoreReader) open() (entry, error) {
 	}
 
 	req, err := http.NewRequest("GET", u.String(), nil)
-	req.AddCookie(ticket)
+	if ticket != nil {
+		req.AddCookie(ticket)
+	}
 
 	res, err := r.ds.Client().Do(req)
 	if err != nil {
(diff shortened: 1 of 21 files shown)


dougm:
Commits since last vendor update: https://github.com/vmware/govmomi/compare/v0.11.4...master

anchal-agrawal:
LGTM

caglar10ur:
LGTM :sheep: :it:

vburenin:
LGTM

cmd/vicadmin/vicadm.go
```diff
@@ -434,7 +434,9 @@ func (r datastoreReader) open() (entry, error) {
 	}
 
 	req, err := http.NewRequest("GET", u.String(), nil)
-	req.AddCookie(ticket)
+	if ticket != nil {
```
dougm: I noticed that too, but left it alone for now too keep the diff focused.

I think we should just s/err/_/ here, since the url has already been parsed and we know "GET" is a valid method - the only 2 cases where it would return in an error.

vburenin: makes sense.

dougm:
.

PR #3424, opened 2016-12-01
Use VM runtime host for datastore HTTP access [full ci]

### LINKED ISSUE #3397: Datastore access to VM logs limited to VM host when using shared storage

dougm:
Required when accessing VM files on shared storage as the VM's runtime host owns the file locks.

Fixes #3397

### DIFF
--- a/cmd/vicadmin/vicadm.go
+++ b/cmd/vicadmin/vicadm.go
@@ -98,6 +98,7 @@ var (
 type logfile struct {
 	URL    url.URL
 	VMName string
+	Host   *object.HostSystem
 }
 
 func init() {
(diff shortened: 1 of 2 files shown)


dougm:
credit to @hickeng for the vicadmin changes.

hickeng:
LGTM - validated on vsphere having vmotioned running container to new host.

cgtexmex:
lgtm

caglar10ur:
LGTM :sheep: :it:

PR #3430, opened 2016-12-02
[skip ci]Design: data migration in upgrade

emlin:
Thanks @caglar10ur and @hmahmood 's help to discuss the data migration in upgrade. Here is the draft document.

We already had upgrade design document for everything else except data migration, and we've implementing based on that. This time, I updated a little bit in the requirement section, to include data migration, and move the old design section to phase 1, and add most of the data migration design to phase 2, to make sure all upgrade related issues are covered in one document.

If you're not familiar with the first phase, no need to check the difference. Read the file in my repo  directly will be easier to review. 

One thing to mention is that @caglar10ur, @hmahmood, we reached agreement on the container upgrade part, but after think through the details, I feel like vic-machine upgrade does not add much value to it, so I added our discussion as one option, but mentioned that my preference is not to do it. 

@hickeng please help to have a look, and let me know your concern

### DIFF
--- a/doc/design/upgrade.md
+++ b/doc/design/upgrade.md
@@ -8,28 +8,25 @@ The VIC engine product will release a new, complete, software download bundle fo
 
 ### Scope
 
-vic-machine can upgrade VCH appliance no matter it's running or not, so vic-machine upgrade cannot rely on services running in VCH appliance.
+vic-machine can upgrade VCH endpointVM no matter it's running or not, so vic-machine upgrade cannot rely on services running in VCH endpointVM.
 
-For the VCH appliance, vic-machine upgrade will detect the version difference between the old VCH and itself, this includes detecting the guestinfo changes between the two versions. Then vic-machine upgrade will migrate the guestinfo metadata, and replace the iso files. If anything incorrect happens during this process, vic-machine upgrade will revert back to old version and status. The VCH will have downtime during upgrade, but should not be broken after this process.
+For the VCH endpointVM, vic-machine upgrade will detect the version difference between the old VCH and itself, this includes detecting the guestinfo changes between the two versions. Then vic-machine upgrade will migrate the guestinfo metadata, and upload updated iso files. If anything incorrect happens during this process, vic-machine upgrade will revert back to old version and status. The endpoint API will have downtime during upgrade, but the running containers will not. 
 
-The containers managed by this VCH will be left running the existing version. If the container is running, it will work well if it's not coupled with appliance through port-forwarding. But after VCH is upgraded, the container management might not be fully functional if there is configuration or communication level changes.
+The containers managed by this VCH will be left running existing version. If the container is running, it will work well if it's not coupled with endpointVM through port-forwarding. After VCH is upgraded, the container management will be resumed even there is container configuration change. Portlayer will be backward compatible to make sure old version's container still work well.
 
-To update the containers, especially while there is security patch available, two options are available. 
+But if there is security patch for container VM kernel or tether, user still need to recreate new container and then replace the old containers at once or gradually based on business requirement. vic-machine upgrade does not take care of container update, which is the same behavior with vanilla docker.
 
-- First is to upgrade the containers from vic-machine. We can migrate container configuration, replace iso file, and then restart the container process. This update requires container downtime as we cannot update tether without stop container process at this time.
+The exception is that in near future we might replace serial port with VMCI, to solve the VM vMotion issue. If after that change, portlayer lose backward compatibility, which means portlayer cannot talk with old container through tether, after endpointVM upgrade, part of container VM management function will loss, which means container list, inspect, network, volume, start/stop/remove still work, but container interaction, including attach, exec, log will not work.
 
- We could upgrade the containers sequentially or in parallel based on user preference.
-
-- Second, leave it to customer like what docker did. User could easily destroy and recreate same containers through docker commands. The new container will be in new version automatically as container iso file is already updated during the appliance upgrade.
-
-In theory, the first option does not add much value, as both options need downtime for the container. If user could accept container stop, they could recreate that container through their own script easily and flexibily, which could be related to their own business requirement.
-Ideally, we should improve tether to separate tether and container lifecycle, so we can replace iso file and restart tether process transparently. Then it will be helpful to have vic-machine upgrade to update containers.
-
-My preference: second option.
+Summary of upgrade requirement:
+- Only upgrade VCH endpointVM
+- Portlayer support backward compatibility
+- If container functionality is impact, user is responsible to replace old container with new container after VCH endpointVM upgrade
+- If there is vSphere credential changes, upgrade will replace old credentials with new one
 
 ### Version Difference
 
-VIC version includes three parts, "release tag"-"build id"-"github commit hash". vic-machine upgrade will rely on build id to detect which version is newer or older than another one.
+VIC version includes three parts, "release tag"-"build id"-"github commit hash". vic-machine upgrade will rely on build id to detect one version is newer or older than another one.
 
 User need to provide newer binary with bigger build id, to upgrade existing VCH. This also requires our build system to keep increase build id number no matter any kind of system change. 
 
@@ -37,9 +34,9 @@ Note: after introduced data migration between builds, we introduced one more int
 
 ### Impact
 
-The VCH appliance (control plane) will have downtime during upgrade. Container lifecyles are not coupled with that of the appliance and, when using mapped vsphere networks (vic-machine --container-network argument) instead of port forwarding, the network data paths are not dependent on the appliance either. 
+The VCH endpointVM (control plane) will have downtime during upgrade. Container lifecyles are not coupled with that of the endpointVM and, when using mapped vsphere networks (vic-machine --container-network argument) instead of port forwarding, the network data paths are not dependent on the endpointVM either. 
 
-There will be impact on container interaction while the appliance is down:
+There will be impact on container interaction while the endpointVM is down:
 
 - no direct access to container logs
 - no attach ability
@@ -47,26 +44,40 @@ There will be impact on container interaction while the appliance is down:
 
 All of those facets should resume normal operation after upgrade is complete. 
 
-Note: Exception might happen if there is container configuration or communication changes. Addtional operations are required to fix the problem.
+Note: Exception might happen if there is container communication changes. Addtional operations are required to fix the problem.
 
 ### VCH status
 
 vic-machine ls and inspect will show the VCH versions and upgrade status.
 
-### Roll Back
+### One Step Roll Back
 
-If anything wrong happens during upgrade, vic-machine will rollback VCH to original version and status. 
+If anything wrong happens during upgrade, vic-machine will rollback VCH to original version and status.
+
+And a little bit further, even vic-machine upgrade succeeds, user is still able to run vic-machine upgrade --rollback to do one step roll back, which means the endpointVM will be rollback to wherever version it was before this time's upgrade.
+
+The benefit is that if the docker endpoint API does not actually work after upgrade (due to vic-machine upgrade mis judgement) or user appliacation still prefer old version's API for whatever reason, they still could get back to old working version easily.
+
+Limitation:
+- If there is new version container created after upgrade, rollback will fail. In that case, user should delete new container through docker CLI or portlayer API.
+- Only one step roll-back is supported. As long as new upgrade is executed, no matter it's succeeds or not, the old version's rollback persisence will be removed.
+
+### Resume Interrupted Upgrade
+
+If upgrade failed, it will rollback to old version dynamically to make sure user application is not stopped due to upgrade failure. But there is one scenario still might break everything, that is user input Ctrl+C during upgrade, cause vic-machine upgrade is interrupte, it is not able to roll back everything it did, and then it is useful to resume this kind of operation from vic-machine. 
+
+Open Issue: if this is high priority user requirement?
 
 ### Downgrade
 
-Verified docker-engine upgrade/downgrade, which works well between 1.11.2 and 1.12. The running container will be stopped after upgrade/downgrade, but is good to start again in new version. And all images and containers information is not lost. But think of the complexity to support version downgrade, we'll not go with this at this moment, unless there is explicity user requirement.
+Verified docker-engine upgrade/downgrade, which works well between 1.11.2 and 1.12. The running container will be stopped after upgrade/downgrade, but is good to start again in new version. And all images and containers information is not lost. But think of the complexity to support version downgrade, we'll not go with this at this moment, only with a limited one step roll back.
 
 ### Internet connection
 Internet connection is not required to upgrade/downgrade VCH, but newer version's binary should be available for vic-machine.
 
 ## Design/Implementation - Phase1
 
-This section described the first simple implementation of upgrade. After this phase, user could upgrade VCH appliance from build to build, as long as there is no metadata changes, which means no changes in guestinfo, key value store and image metadata.
+This section described the first simple implementation of upgrade. After this phase, user could upgrade VCH endpointVM from build to build, as long as there is no metadata changes, which means no changes in guestinfo, key value store and image metadata.
 
 This code is already merged, and works for security patch update.
 
@@ -95,10 +106,6 @@ Image file name should not be changed in datastore, cause the file path is used 
 
 Note: No feasible golang library found for this function, so will write our own library to read iso metadata.
 
-### Configuration Data Migration
-
-To support VCH upgrade, any changes in VCH configuration and container VM configuration will introduce additional effort to migrate data from old version to newer version. And if we support a long upgrade path, for example from 2000->2500->3000->3500->4000,  there will have a list of utilies recognize old configuration structure to help data migration.
-
 #### Refactor current VCH configuration and container VM configuration structure
 To make data migration easier, we'll need to refactor current VCH and container VM configuration structure.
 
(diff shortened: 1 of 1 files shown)


doc/design/upgrade.md
```diff
@@ -1,20 +1,37 @@
 # Upgrade
 
-After installed VCH, VIC admin would want to upgrade existing instances to latest version. This document describes the upgrade requirement, and design options.
+After installed VCH, VIC admin would want to upgrade existing instances to latest version, to get new features or to fix exposed security issues. This document describes the upgrade requirement, and design options.
```
mdubya66: After installing a VCH, a VIC Engine admin will need to manage the lifecycle of the VCH. One aspect of this lifecycle management is upgrading and patching. For our purposes patching and upgrading will be treated the same. 

The VIC Engine product will release a new, complete, software download bundle for both patches and upgrades, versus, having a sparse patch or a separate full upgrade model.

doc/design/upgrade.md
```diff
@@ -1,20 +1,37 @@
...
 
 ## Requirements
 
 ### Scope
 
-While a new version is ready, new version's vic-machine upgrade can upgrade one running or powered off VCH appliance. The containers managed by this VCH will be left in existing version. To have a new version container, user need to create new container from upgraded VCH instance. If existing container is stopped and restarted, the old version should be kept.
+While a new version is ready, new version's vic-machine can upgrade one running or powered off VCH appliance.
```
mdubya66: This is not very clear to me.

doc/design/upgrade.md
```diff
@@ -1,20 +1,37 @@
...
 
-VIC version includes three parts, "release tag"-"build id"-"github commit hash".
+For VCH appliance, vic-machine upgrade will detect the version difference between old VCH and itself, find out the metadata change in those two versions, and then migrate metadata, replace iso files. If anything wrong happens in this process, revert back to old version and status. VCH will have downtime during upgrade, but should not be broken after this process.
```
mdubya66: Suggestion: 
For the VCH appliance, vic-machine upgrade will detect the version difference between the old VCH and itself. This includes detecting the guestinfo changes between the two versions. Then vic-machine upgrade will migrate the guestinfo metadata and replace the iso files. If anything incorrect happens during this process, vic-machine upgrade will revert back to old version and status. The VCH will have downtime during upgrade, but should not be broken after this process.

doc/design/upgrade.md
```diff
@@ -1,20 +1,37 @@
...
 
-vic-machine upgrade will rely on build id change only. User need to provide newer binary with bigger build id, to upgrade existing VCH.
+The containers managed by this VCH will be left in existing version. If container is running, it will work well if it's not coupled with appliance through port-forwarding. But after VCH is upgraded, the container management might not be fully functional if there is configuration or communication level changes.
```
mdubya66: The containers managed by this VCH will be left running the existing version. If the container is running, it will work well if it's not coupled with appliance through port-forwarding. But after VCH is upgraded, the container management might not be fully functional if there is configuration or communication level changes.

doc/design/upgrade.md
emlin: hmm... It's actually saying even appliance is powered off, vic-machine still can upgrade it. Will update

caglar10ur:
overall LGTM, adding @hickeng to the reviewers list just in case he ends up getting bored in his flight :)

doc/design/upgrade.md
```diff
@@ -125,15 +135,114 @@ Here is the upgrade workflow
...
+
+Appliance log files configuration will not be touched at this time. So upgraded VCH will use old log files if there is any such kind of change. We can revisit this while there is customer requiement on this.
+
+### vSphere Object Management Logic
+
+There will have new vSphere API come up, so the logic to manage vsphere objects will be enhanced, for example, as Caglar said the vmdk can be managed directly in vSphere 6.5, instead of through vm operations in vSphere 6.0. We'll switch to these new interfaces for image and volume management sometime later.
+
+vic-machine is not supposed to migrate old image data or volume data to new vmdk files, so portlayer will need to be backward compatible. (Suppose vSphere will be backward compatibile, there should be nothing to do in vic, but need to do some research on it)
+
+### How to Support Upgrade Between Builds
```
emlin: > 1. Why do we call a migration patch as a plugin? Will it be stored as a separate file on disk?

The plugin name is corresponding to upgrade framework. Each plugin is one migration function called by framework based on their version ID.

> 2. ...

We're doing that based on new data migration version. Which is not related to release version. The PR #3478 has one prototype for this idea. You might check that to see if there is any inconsistency with your suggestion.

> 3. 

We did backup guestinfo in current upgrade code, and in fatal upgrade, everything could be reverted.
But the only exception is keyvalue store, which is described in previous section.

emlin:
@hickeng although we roughly agree with the data migration idea, still prefer to get your review on this document.

doc/design/upgrade.md
```diff
@@ -125,15 +135,114 @@ Here is the upgrade workflow
...
+
+In the future, if there is any change on assumption no matter for what reason, performance, easy to management, data migration will be hard to do.
+
+So the idea is that, data migration framework does not assume where to load keyvalue store. If there is keyvalue store changes, plugin is the owner to read old version, update to new version and then write back to datastore files, or to anywhere defined in new version, suppose the configuration can be retrieved from guestinfo. So the input of keyvalue store plugin will be same to other appliance configuration update plugin.
+
+The problem of this solution is that each single plugin will persist its own change, not like the guestinfo update, which is migrated in memory and persisted by migration framework. So the rollback for guestinfo is easy, but not possible for keyvalue store plugin, unless we have data roll back plugin mechanism.
+
+The workaround for this idea is to restrict keyvalue store migration functions, e.g. only new key is supported. In the plugin, do not remove old keyvalues, so in the above failure case, old binary can still work with old keyvalues.
```
hickeng: I am ok with only allowing the addition of new keys into an existing kvstore, but from a rollback perspective it's potentially awkward. Example:
you have one key acting as an index for networks and another holding detailed network information
you migrate to new version that has new key for different detailed information but still using the old index
if changes are made to the key shared across versions then rollback will not result in consistent data.

I suggest we go with a very simple copy model for now - simply create a new copy of the datafile with a suffix that matches the build number (or version number). That way rollback will simply be to start using the appropriately named file again and data should be internally consistent (although on guarantees on whether containers have been added or deleted).

doc/design/upgrade.md
```diff
@@ -125,15 +135,114 @@ Here is the upgrade workflow
...
+This document does not include approach for tether communication change, that means from serial port to VMCI or something else. Need to think about how to support old version's container while that change is made.
+
 ## Restrictions
 
 User cannot run two upgrades for same VCH at the same time. 
 
-vic-machine will check if there is already another upgrade snapshot is created before it start to create snapshot. But as create vsphere snapshot will take some time, e.g. one minute, if at this time, another upgrade process is started, it will start upgrade again cause the snapshot of previous task is not finished yet.
- 
\ No newline at end of file
+vic-machine will check if there is already another upgrade snapshot is created before it starts to create snapshot. But as create vsphere snapshot will take some time, e.g. one minute, if at this time, another upgrade process is started, it will start upgrade again cause the snapshot of previous task is not finished yet.
```
hickeng: Ideally we can name the snapshots using a predictable version based scheme so that we get a name collision earlier if an upgrade is already in progress.

We _should_ look into how to resume a partial upgrade if vic-machine is killed part way through.

doc/design/upgrade.md
```diff
@@ -1,20 +1,39 @@
...
+
+- First is to upgrade the containers from vic-machine. We can migrate container configuration, replace iso file, and then restart the container process. This update requires container downtime as we cannot update tether without stop container process at this time.
+
+ We could upgrade the containers sequentially or in parallel based on user preference.
+
+- Second, leave it to customer like what docker did. User could easily destroy and recreate same containers through docker commands. The new container will be in new version automatically as container iso file is already updated during the appliance upgrade.
+
+In theory, the first option does not add much value, as both options need downtime for the container. If user could accept container stop, they could recreate that container through their own script easily and flexibily, which could be related to their own business requirement.
+Ideally, we should improve tether to separate tether and container lifecycle, so we can replace iso file and restart tether process transparently. Then it will be helpful to have vic-machine upgrade to update containers.
```
emlin: > This would only address updating tether, not updating kernel, kernel modules or security issues in the container image.

Yeah, you're right. Removed that section. 

For the security issue in container VM kernel, the suggestion is to recreate container, instead of update, cause they will all introduce container downtime.

The container will be updated only after re-creation, in another word, is immutable.

doc/design/upgrade.md
```diff
@@ -26,7 +45,9 @@ There will be impact on container interaction while the appliance is down:
 - no attach ability
 - network outage for NAT based port forwarding
 
-All of those facets should resume normal operation after upgrade is complete
+All of those facets should resume normal operation after upgrade is complete.
```
emlin: It should be and should be resume after VCH restart. But yeah, @hmahmood please confirm. thanks

doc/design/upgrade.md
```diff
@@ -38,14 +59,16 @@ If anything wrong happens during upgrade, vic-machine will rollback VCH to origi
 
 ### Downgrade
 
-For any running VCH, VIC admin can downgrade existing VCH to an old version. (I assume this function is only used for upgrade roll back, we might not want to recommend downgrade for anything working well, then the downgrade scope will be same to upgrade versions)
-
-Verified docker-engine upgrade/downgrade, which works well between 1.11.2 and 1.12. The running container will be stopped after upgrade/downgrade, but is good to start again in new version. And all images and containers information is not lost. So from technical point of view, we should support same story.
+Verified docker-engine upgrade/downgrade, which works well between 1.11.2 and 1.12. The running container will be stopped after upgrade/downgrade, but is good to start again in new version. And all images and containers information is not lost. But think of the complexity to support version downgrade, we'll not go with this at this moment, unless there is explicity user requirement.
```
emlin: So, what do you mean for roll-back of one update? I assume we don't have roll-back at all. What we did is to roll back only when upgrade failed.

emlin: >If containers are created with the upgraded VCH, what happens to them after rollback? 

I guess here you mean if container configuration upgrade failed, and then we rollback to old version, right? Cause if VCH upgrade is rolled back, then nothing will happen for container.
Currently, we only support guestinfo upgrade for container VM, not related to key/value store update. 
So actually two things related
- there is container related data persisted in key/value store, which is not backward compatible, so we might have function loss for old version's container. I'll figure out if we have that right now.
- container guestinfo data migration failed, and rolled back. Which is bad for our case, cause we don't want to roll back VCH upgrade in this case, that means we lost all support for that container. Portlayer need to handle that failure, but eventually, from docker client, user might not be able to see that container anymore.

doc/design/upgrade.md
hmahmood: The port mapping info is not stored in the kv store; instead at personality bring up the port mapping information is retrieved from the port layer as part of each container's configuration (during cache "sync up"). The port mapping info is stored as part of the containers' guestinfo.

doc/design/upgrade.md
```diff
@@ -125,15 +135,114 @@ Here is the upgrade workflow
 - Cleanup env after upgrade (remove upgrade snapshot, remove uploaded iso files if upgrade failed)
```
emlin: Okay, so this is the one version roll-back you mean in the previous comments.
As my understanding, this will be a separate command in vic-machine. Which is to roll-back the upgrade to whatever version it was before upgrade.

And this means most likely user will have left over every time after upgrade, snapshot, image files, etc, to make sure they have opportunity to run one step roll back. And based on this idea, why should we provide --cleanup option for them? Simply lose the opportunity of roll back?

To make the rollback works more cleanly, I will need a cleanup before upgrade, to remove anything left over from last time's upgrade, to avoid confusing in the rollback. And this in turn will not leak more resources just for upgrade.

We could add one cleanup command in the future if this becomes something higher priority.

doc/design/upgrade.md
emlin: I see your point. Please skip my reply above.
For this, I think we should limit the roll back function. If there is new version container created already, we should fail the rollback, and let user delete new container before rollback. As after rollback, most likely the new container will not function well, user should make decision on that.

doc/design/upgrade.md
```diff
@@ -125,15 +135,114 @@ Here is the upgrade workflow
...
+
+- Appliance guestinfo, which is used to persist VCH appliance configuration
+- Container guestinfo, which is used to persist container configuration
+- KeyValue Store, which is used to persist image metadata and network portlayer data
+- Appliance log files
+- vSphere object management logic change
+
+### Guestinfo Migration
+
+- One VirtualContainerHost structure per one version, this will be used by extraconfig package to read back old vch's configuration
```
emlin: This is a good intention towards to leverage portlayer for VCH appliance provisioning. Let me do it now. It's good to have some roughly idea before go ahead with the upgrade approach.

doc/design/upgrade.md
```diff
@@ -125,15 +135,114 @@ Here is the upgrade workflow
...
+- vic-machine upgrade framework will calculate the plugin chain based on old VCH's version and vic-machine version, and then call those plugins sequentially. This will look like r1->r2->r3->f4->latest or r3->r4->r5
+- Run guestinfo validation to figure out if there is any field missing (one possible option here is to translate configuration back to vic-machine input data, and then rerun validator to make sure all vsphere related objects are still exist and correct)
+- Prompt missing parameters if there is any (for new parameters added and cannot be empty)
+- Write new configuration back to appliance guestinfo
+
+With this design, developer is able to add/delete/repurpose existing configuration items, except VCH version. And we'll need one guidline for what to do while one configuration item is changed.
+
+Besides of this solution, we also dicussed another option, that is to revert configuration back to vic-machine create option, and try to recreate a new VCH appliance, and then migrate containers from the old appliance to new appliance. This solution can reduce VCH downtime, and even migrate alive.
```
emlin: For here, I have some different idea as I mentioned in the VCH reconfigure issue. There have pros and cons for different options. Let's talk about that in person first.

doc/design/upgrade.md
emlin: I agree with you, and thanks for find it. We've been talked about that while you're in Austin.
I actually updated the proto type and developer guide based on this idea. But unfortunately did not find here has this description as well. Need more linkage for this kind of change.

doc/design/upgrade.md
emlin: Will have version in snapshot. 
Just actually we don't care if there is name collision. vSphere will simply add another snapshot based on current one with same name. 
And I remember it doesn't report failure if there is another upgrade task is running. So the detection should be improved, probably through task checking.

> We should look into how to resume a partial upgrade if vic-machine is killed part way through.

For the resume. Actually we didn't have create resume, though we leave the partial created VCH there. 
But upgrade gonna rollback for anything wrong, so the upgrade resume will benefit the user who control-C during vic-machine upgrade is running.
Should not be a big deal if snapshot is already created. But if snapshot is not there, resume might not have much value. I'd rather not to rely on existing iso files to detect if upgrade is executed or not.

doc/design/upgrade.md
```diff
@@ -125,15 +135,114 @@ Here is the upgrade workflow
...
+
+Second concern is that we'll need to version more objects and logics, includes appliance guestinfo, vic-machine create options, and the logic to revert guestinfo back to vic-machine create options, those are all version sensitive.
+
+We didn't try to figure out how to do it at this time, cause it looks overdesigned.
+
+Note: extraconfig package should always be backward compatible. If it breaks this assumption, upgrade does not work for both options.
+
+### Secret Configuration
+
+In appliance configuration, there are few items are encrypted for security reason, including user password and certificate so far. The encryption key is readable from in guest only. vic-machine is reading configuration through vsphere API, so not be able to decrypt secret information from old appliance configuration.
```
emlin: Talked to @dougm for the possible vmodl API. 
I tried the API of VirtualMachine.ExtractOvfEnvironment, but it does not return expected key, possibly because we didn't use the XML schema where ovf inject relies on.
So we still do not have solution to read back encryption key even in VC.

dougm: Note that the value of guestinfo.ovfEnv is stored in the .vmx file, the API doesn't prevent an admin user from using Datastore.Download to get this file.  Same format for ESX and VC.

emlin: ah.. I see. this make sense. Then I cannot say it's unreadable at all. Thanks @dougm for the clarification.

emlin:
Updated the document with pretty much everything addressed, except the refactor of VirtualContainerHost, which is to help vic-machine to use portlayer. 
Still need more discussion on that part, but it is not directly related to upgrade, so please have another review.
thanks

doc/design/upgrade.md
hickeng: It's also of use if the vSphere connection drops for some reason. Agree that detecting ISO files isn't a good option - I brought this up because, depending on the mechanism, there's a possibility to mis-identify a partial upgrade as in-progress which would stall the ability to retry.
This is the same scenario as `resume` if resume is implemented as revert/retry.

doc/design/upgrade.md
hickeng: @hmahmood doesn't impact @emlin but which field in the contianer config? I can see exposed ports, but nothing which port they're forwarded to.

hmahmood: @hickeng the full port mapping spec is stored in the config (parsable by: https://github.com/vmware/vic/blob/master/vendor/github.com/docker/go-connections/nat/nat.go#L128)

doc/design/upgrade.md
emlin: Agree

emlin:
@hickeng still waiting your feedback. thanks

hickeng:
@emlin I cannot tell what changed due to the force push - saying yes at this point so we can move onto other things.

PR #3454, opened 2016-12-06
[skip ci] Using Start Process async wrapper util for ESX & VC deploy

rajanashok:
Adding the Start Process async command to deploy ESX and VC using the sshpass, and also added a wrapper function for the same.

Updated the Deploy Multiple ESX servers function to use the new async wrapper utils.

### DIFF
--- a/tests/resources/Nimbus-Util.robot
+++ b/tests/resources/Nimbus-Util.robot
@@ -41,14 +41,17 @@ Deploy Multiple Nimbus ESXi Servers in Parallel
     ${name1}=  Evaluate  'ESX-' + str(random.randint(1000,9999))  modules=random
     ${name2}=  Evaluate  'ESX-' + str(random.randint(1000,9999))  modules=random
     ${name3}=  Evaluate  'ESX-' + str(random.randint(1000,9999))  modules=random
-    Log To Console  \nDeploying Nimbus ESXi server: ${name1}
-    Log To Console  \nDeploying Nimbus ESXi server: ${name2}
-    Log To Console  \nDeploying Nimbus ESXi server: ${name3}
 
     Open Connection  %{NIMBUS_GW}
     Login  ${user}  ${password}
 
-    ${out}=  Execute Command  nimbus-esxdeploy ${name1} --disk=48000000 --ssd=24000000 --memory=8192 --nics 2 ${version} & nimbus-esxdeploy ${name2} --disk=48000000 --ssd=24000000 --memory=8192 --nics 2 ${version} & nimbus-esxdeploy ${name3} --disk=48000000 --ssd=24000000 --memory=8192 --nics 2 ${version}
+    ${out1}=  Deploy Nimbus ESXi Server Async  ${name1}
+    ${out2}=  Deploy Nimbus ESXi Server Async  ${name2}
+    ${out3}=  Deploy Nimbus ESXi Server Async  ${name3}
+
+    Wait For Process  ${out1}
+    Wait For Process  ${out2}
+    Wait For Process  ${out3}
 
     ${out}=  Execute Command  nimbus-ctl ip ${user}-${name1}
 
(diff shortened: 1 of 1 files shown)


tests/resources/Nimbus-Util.robot
```diff
@@ -135,6 +138,23 @@ Deploy Nimbus vCenter Server
...
+    Start Process  sshpass -p ${password} ssh -o StrictHostKeyChecking\=no ${user}@%{NIMBUS_GW} nimbus-esxdeploy ${name} --disk\=48000000 --ssd\=24000000 --memory\=8192 --nics 2 ${version}  shell=True  alias=${handle}
+    [Return]  ${handle}
+
+Deploy Nimbus vCenter Server Async
+    [Arguments]  ${user}  ${password}  ${version}=${VC_VERSION}
+    ${name}=  Evaluate  'VC-' + str(random.randint(1000,9999))  modules=random
+    ${handle}=  Evaluate  'vc-handle-' + str(random.randint(1000,9999))  modules=random
+    Log To Console  \nDeploying Nimbus vCenter server: ${name}
+
+    Start Process  sshpass -p ${password} ssh -o StrictHostKeyChecking\=no ${user}@%{NIMBUS_GW} nimbus-vcvadeploy --vcvaBuild ${version} ${name}  shell=True  alias=${handle}
```
mhagen-vmware: Same as above

tests/resources/Nimbus-Util.robot
```diff
@@ -135,6 +138,23 @@ Deploy Nimbus vCenter Server
     Close connection
     [Return]  ${user}-${name}  ${ip}
 
+Deploy Nimbus ESXi Server Async
+    [Arguments]  ${user}  ${password}  ${name}  ${version}=${ESX_VERSION} 
+    ${handle}=  Evaluate  'esx-handle-' + str(random.randint(1000,9999))  modules=random
+    Log To Console  \nDeploying Nimbus ESXi server: ${name}
+
+    Start Process  sshpass -p ${password} ssh -o StrictHostKeyChecking\=no ${user}@%{NIMBUS_GW} nimbus-esxdeploy ${name} --disk\=48000000 --ssd\=24000000 --memory\=8192 --nics 2 ${version}  shell=True  alias=${handle}
```
rajanashok: Didn't know it was possible with PID, reverted it to use the PID now.

mhagen-vmware:
In general this looks good, but a couple of things:
1. you should add Tag [secrets] to the keyword with sshpass
2. I would like to see a more generic wrapper here, that just allows you to run any command with sshpass (and is tagged secret), that way we can use the keyword for deploying ESX/VC/tesbeds/cleanup/get IP/etc. Then we can start sprinkling it throughout the manual suite to speed up all the tests.
3. We can then wrap that in other keywords including an ESX and VC wrapper like you have, if it makes sense still.

rajanashok:
Agreed. Overall, it makes sense still I need to understand the secrets tag in robot, will look into it.

rajanashok:
@mhagen-vmware Think I have captured the comments with this commit. If this is good then I can create a separate PR to expand this to other keywords.

tests/resources/Nimbus-Util.robot
```diff
@@ -138,6 +141,29 @@ Deploy Nimbus vCenter Server
     Close connection
     [Return]  ${user}-${name}  ${ip}
 
+Deploy Nimbus ESXi Server Async
+    [Tags]  secret
+    [Arguments]  ${user}  ${password}  ${name}  ${version}=${ESX_VERSION}
```
mhagen-vmware: I don't think we want the user and password here as an argument, as it could leak the user/pass in the calling keyword.  I think simplify and remove the args here and just pass %{NIMBUS_USER}/PASSWORD directly into secret sshpass command now.

mhagen-vmware:
one more minor change, but other than that

LGTM

anchal-agrawal:
LGTM

PR #3461, opened 2016-12-07
Support running a specific test suite in CI on PRs [specific ci=1-19-Docker-Volume-Create]

### LINKED ISSUE #3305: Allow running particular tests on CI in pull requests
We should add the ability to specify a particular integration test or a range of tests to run on CI in pull requests. This could involve working with Drone plugins or parsing the PR title/description.

anchal-agrawal:
Adds basic support for running a particular integration test suite on CI. Including `[specific ci=$suitename]` in the PR's title or commit messages will trigger this behavior.

Also adds a small note to the PR template.

Fixes #3305

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -12,3 +12,11 @@ Refer to CONTRIBUTING.MD for more details.
 -->
 
 Fixes #
+
+<!--
+To trigger a custom build with this PR, include one of these in the PR's title or commit messages:
+* [skip ci]
+* [full ci]
+* [specific ci=$suitename] e.g. [specific ci=1-01-Docker-Info]
+    With this option, running only one suite is supported.
+-->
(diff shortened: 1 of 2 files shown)


matthewavery:
Looking good thus far! Can there be a comma separated list to do multiple suites? like `[specific ci=<suite1>,<suite2>,...]. That would be awesome!

mhagen-vmware:
lgtm

mhagen-vmware:
I would prefer this approach, if you want multiple suites I think you should run full ci. This is cleaner and gives us the targeted test approach that we should be doing when we touch a specific area.

matthewavery:
my one concern is that when we do things like modify `volume create` that affects both the `container create` tests and the `volume create` tests. I would like the ability to run just 2 or 3 tests for that reason. @mhagen-vmware

mhagen-vmware:
In that example it should affect quite a few tests more than that - including regression, docker run, compose, etc. So run [full ci] :)

matthewavery:
LGTM

anchal-agrawal:
🏁

PR #3471, opened 2016-12-08
[skip ci] Longevity should use the static IP for austin site as well

mhagen-vmware:
So we are not hammering the austin site dhcp pool still.


anchal-agrawal:
LGTM

sgairo:
LGTM

PR #3478, opened 2016-12-09
[full ci]Do not remove vm from container cache if it's fixing

### LINKED ISSUE #3196: Error response from daemon: No such container: 1069fb0efd2f
From command in the CI build (https://ci.vmware.run/vmware/vic/6919):
```
docker -H dhcp-192-168-31-152.ci.drone.local:2376 --tls ps -aq | xargs -n1 docker -H dhcp-192-168-31-152.ci.drone.local:2376 --tls rm -f
```
```
	${output} = 5cd2c9a3166b
Error response from daemon: No such container: 1069fb0efd2f
4f51e5765af0
97bbcadf7124
1afecec824de
e1e05cb19b59
a514e546f78c
c5c215244d36
b4c792b7ddd9
```

[integration_logs_6919_2f30898a21a0caf6d8db00ddce758990b48856d0.zip](https://github.com/vmware/vic/files/594898/integration_logs_6919_2f30898a21a0caf6d8db00ddce758990b48856d0.zip)

Somehow, a container reported in ps -aq doesn't exist for docker rm -f?

emlin:
Fixes #3196

### DIFF
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -149,6 +149,20 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
 	defer trace.End(trace.Begin(fmt.Sprintf("vm %q, force %t", vm.String(), force)))
 
 	var err error
+	// get the actual folder name before we delete it
+	folder, err := vm.FolderName(d.ctx)
+	if err != nil {
+		// failed to get folder name, might not be able to remove files for this VM
+		name := d.getName(vm)
+		if name == "" {
+			log.Errorf("Unable to automatically remove all files in datastore for VM %q", vm.Reference())
+		} else {
+			// try to use the vm name in place of folder
+			log.Infof("Delete will attempt to remove datastore files for VM %q", name)
+			folder = name
+		}
+	}
+
 	power, err := vm.PowerState(d.ctx)
 	if err != nil || power != types.VirtualMachinePowerStatePoweredOff {
 		if err != nil {
@@ -170,19 +184,6 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
 			return vm.PowerOff(ctx)
 		}); err != nil {
 			log.Debugf("Failed to power off existing appliance for %s, try to remove anyway", err)
-		}
-	}
-	// get the actual folder name before we delete it
-	folder, err := vm.FolderName(d.ctx)
-	if err != nil {
-		// failed to get folder name, might not be able to remove files for this VM
-		name := d.getName(vm)
-		if name == "" {
-			log.Errorf("Unable to automatically remove all files in datastore for VM %q", vm.Reference())
-		} else {
-			// try to use the vm name in place of folder
-			log.Infof("Delete will attempt to remove datastore files for VM %q", name)
-			folder = name
 		}
 	}
 
--- a/lib/install/validate/fix_container.go
+++ b/lib/install/validate/fix_container.go
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package exec
+package validate
 
 import (
 	"context"
@@ -25,13 +25,12 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/vm"
 )
 
-// InitErrorHandler register ContainerFixer to tasks error handler. So if any container vm operations fail for invalid state issue, this handler will be executed to fix the error.
-// The benefit to register handler in portlayer is that it can set container state during vm fixing.
-func InitTasksErrorHandler() {
-	tasks.RegisterErrorHandler(fixContainerHandler)
+// init register fixVMInvalidStateHandler to tasks error handler. So if any container vm operations fail for invalid state issue, this handler will be executed to fix the error.
+func init() {
+	tasks.RegisterErrorHandler(fixVMInvalidStateHandler)
 }
 
-func fixContainerHandler(ctx context.Context, err error) (bool, error) {
+func fixVMInvalidStateHandler(ctx context.Context, err error) (bool, error) {
 	defer trace.End(trace.Begin(fmt.Sprintf("error: %s", err)))
 	o := ctx.Value(tasks.VMObjectKey)
 	if o == nil {
@@ -49,11 +48,6 @@ func fixContainerHandler(ctx context.Context, err error) (bool, error) {
 		return false, nil
 	}
 	log.Debugf("Try to fix failure %s", err)
-	container := Containers.Container(vm.Reference().String())
-	if container != nil {
-		oldState := container.SetState(StateFixing)
-		defer container.SetState(oldState)
-	}
 	if nerr := vm.FixVM(ctx); nerr != nil {
 		log.Errorf("Failed to fix task failure: %s", nerr)
 		return true, nerr
--- a/lib/portlayer/event/collector/vsphere/vm_event.go
+++ b/lib/portlayer/event/collector/vsphere/vm_event.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/lib/portlayer/event/collector/vsphere/vm_event.go
+++ b/lib/portlayer/event/collector/vsphere/vm_event.go
@@ -39,8 +39,6 @@ func NewVMEvent(be types.BaseEvent) *VMEvent {
 		ee = events.ContainerRemoved
 	case *types.VmGuestShutdownEvent:
 		ee = events.ContainerShutdown
-	case *types.VmRegisteredEvent:
-		ee = events.ContainerRegistered
 	case *types.VmMigratedEvent:
 		ee = events.ContainerMigrated
 	case *types.DrsVmMigratedEvent:
--- a/lib/portlayer/exec/exec.go
+++ b/lib/portlayer/exec/exec.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/lib/portlayer/exec/exec.go
+++ b/lib/portlayer/exec/exec.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/lib/portlayer/event"
 	"github.com/vmware/vic/lib/portlayer/event/collector/vsphere"
 	"github.com/vmware/vic/lib/portlayer/event/events"
--- a/lib/portlayer/exec/exec.go
+++ b/lib/portlayer/exec/exec.go
@@ -161,12 +161,6 @@ func eventCallback(ie events.Event) {
 				}()
 			case StateRemoved:
 				log.Debugf("Container(%s) %s via event activity", container, newState)
-				if container.vm != nil && container.vm.Fixing {
-					// is fixing vm, do not remove from containers
-					log.Debugf("Container(%s) %s is being fixed", container.ExecConfig.ID)
-					container.vm.Fixing = false
-					break
-				}
 				Containers.Remove(container.ExecConfig.ID)
 				publishContainerEvent(container.ExecConfig.ID, ie.Created(), ie.String())
 			}
@@ -300,7 +294,7 @@ func eventedState(e string, current State) State {
 			return StateSuspended
 		}
 	case events.ContainerRemoved:
-		if current != StateRemoving {
+		if current != StateRemoving && current != StateFixing {
 			return StateRemoved
 		}
 	}
--- a/lib/portlayer/exec/exec.go
+++ b/lib/portlayer/exec/exec.go
@@ -25,8 +25,6 @@ import (
 
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
-	"github.com/vmware/govmomi/vim25/mo"
-	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/lib/portlayer/event"
 	"github.com/vmware/vic/lib/portlayer/event/collector/vsphere"
 	"github.com/vmware/vic/lib/portlayer/event/events"
@@ -107,12 +105,6 @@ func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSou
 		)
 		// subscribe the exec layer to the event stream for Vm events
 		Config.EventManager.Subscribe(events.NewEventType(vsphere.VMEvent{}).Topic(), "exec", eventCallback)
-		// subscribe callback to handle vm registered event
-		Config.EventManager.Subscribe(events.NewEventType(vsphere.VMEvent{}).Topic(), "registeredVMEvent",
-			func(ie events.Event) {
-				registeredVMCallback(sess, ie)
-			},
-		)
 
 		// instantiate the container cache now
 		NewContainerCache()
@@ -169,6 +161,12 @@ func eventCallback(ie events.Event) {
 				}()
 			case StateRemoved:
 				log.Debugf("Container(%s) %s via event activity", container, newState)
+				if container.vm != nil && container.vm.Fixing {
+					// is fixing vm, do not remove from containers
+					log.Debugf("Container(%s) %s is being fixed", container.ExecConfig.ID)
+					container.vm.Fixing = false
+					break
+				}
 				Containers.Remove(container.ExecConfig.ID)
 				publishContainerEvent(container.ExecConfig.ID, ie.Created(), ie.String())
 			}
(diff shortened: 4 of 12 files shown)


pkg/vsphere/vm/vm.go
```diff
@@ -52,6 +52,8 @@ type VirtualMachine struct {
 	*object.VirtualMachine
 
 	*session.Session
+
+	Fixing bool
```
vburenin: This struct member needs a comment what this is and how it is used.

vburenin: Is there a possibility of a concurrent access by chance? It feels like it is.

lib/portlayer/exec/exec.go
```diff
@@ -161,69 +155,20 @@ func eventCallback(ie events.Event) {
 				}()
 			case StateRemoved:
 				log.Debugf("Container(%s) %s via event activity", container.ExecConfig.ID, newState.String())
+				if container.vm != nil && container.vm.Fixing {
+					// is fixing vm, do not remove from containers
+					log.Debugf("Container(%s) %s is being fixed", container.ExecConfig.ID)
+					container.vm.Fixing = false
```
hmahmood: You should use the golang [atomic](https://golang.org/pkg/sync/atomic/) package here instead of setting this field directly.

pkg/vsphere/vm/vm.go
```diff
@@ -484,6 +486,7 @@ func (vm *VirtualMachine) fixVM(ctx context.Context) error {
 
 	name := mvm.Summary.Config.Name
 	log.Debugf("Unregister VM %s", name)
+	vm.Fixing = true
```
hmahmood: Ditto

lib/portlayer/exec/exec.go
```diff
@@ -161,69 +155,20 @@ func eventCallback(ie events.Event) {
...
 				publishContainerEvent(container.ExecConfig.ID, ie.Created(), ie.String())
-
 			}
 		}
 	}
 	return
 }
 
-// registeredVMCallback will process registeredVMEvent
-func registeredVMCallback(sess *session.Session, ie events.Event) {
```
hmahmood: Why is this code being removed?

hickeng:
We already have code to suppress triggering behaviour on VM events if the recorded VM state warrants it.

I suggest we take this opportunity to add semi-structured error handling to the task package with the ability to register an result handler. This would allow us to do the following: 
1. add a log handler - this can log the opid so we can easily correlate between VIC operation and vpxd/hostd operation, even for successful operations
2. add an InvalidState handler - this can be implemented by the port layer and can:
  a. set the container state to REPAIRING
  b. use the container state to suppress the behaviours of Remove and Register events using the existing mechanisms
3. TaskInProgress handler for retry logic
4. Task queue errors - we're not seeing these yet but we will eventually once we add concurrent scale testing
5. other handlers for logging unexpected errors - this can inlcude vsphere error to VIC error message mapping

pkg/vsphere/vm/vm.go
emlin: It did feel like that, but this should not have concurrent access. Only before unregister, we'll set this value, and only while the event listener received vm unregistered event, it will reset this value.
So it will always be sequential, and if they are not by any chance, the problem is not fixable by lock.

lib/portlayer/exec/exec.go
emlin: Thought about that, but as the above comment, for this variable, it's always sequential accessed.

lib/portlayer/exec/exec.go
emlin: We don't remove vm for while it's unregistered, so no need to listen for registered vm event

pkg/vsphere/vm/vm.go
emlin: ^^^

emlin:
hmm... I could understand the benefit for task handlers, just is wondering for the first and last handlers.
We should use operation instead of logger to correlate VIC operations and vsphere operation, but why that should be a customized handler, instead of a general function in task packages and also for the logging unexpected errors? I assume even the function is called from vic-machine, these kind of logging are also helpful.

emlin:
Okay, gonna add some built in handlers in tasks, which are common for all functions, e.g. log handler, retry handler, etc. 
And offer the ability to add more handlers from caller. So portlayer could add invalid state vm handler, and task queue error handler in the future.

emlin:
Unfortunately, after more code reading, figure out more problems.
- tasks package is not where the VM is fixed, cause it does not have enough information to unregister/register. 
  Tasks might fail during task creation, and might fail during task execution. But tasks package only have one func to create task, which is not enough for vm operations. 
  That's why we add a separate WaitForResult in VirtualMachine object for vm's task handling only
  And then it's easy to know that even we add error handlers in tasks package, it's still not the solution for vm fixing.
- VM invalid state will not only fail VM operations, it will also fail vm.Properties call. In that case, we'll throw out NPE from anywhere we didn't check if the return value is empty.
  Now most of the CI failures are because we're fixing vm in vm.Properties method call, cause after docker stop we'll refresh container immediately.

Based on your idea, I could add one vm error handler, which could be registered from portlayer, to handle invalid state vm issue. Then this can avoid event handler issue.
And from vic-machine, the vm fixing is not really necessary after a small fix, so we'll not have that handler while vic-machine is executing.

But with this solution, tasks package is not touched, cause that's unrelated to this change at this moment.

@hickeng please let me know if you have any concerns on it.

hickeng:
@emlin There's a variety of ways in which we can pass additional data through to the handler.

Context.WithValue is one and could be used to pass through a govmomi object and the handler can check to see if the additional contextual data it needs to operate is available and if the type is appropriate.
An interface{} is another - this is basically the same as the above option, but not packing it into the context object.
Either of the above options can be done via the existing VM wrapper for the WaitForResult call - the idea is that we benefit from other generic handlers as well such logging without having to implement it in many different locations.

At the moment the form and placement of the tasks code makes it awkward to do any complex cross-cutting error handling. We can either fix that, we can implement those error handling aspects in multiple locations, or we can try to provide additional error handling code as a secondary call to another set of utility functions.

emlin:
@hickeng Make sense. 
Btw, are you looking forward to change context to operation in tasks package? That will involve lots of caller changes for the signature change. I saw some of the methods in portlayer changed to do it.

pkg/vsphere/tasks/waiter.go
```diff
@@ -27,12 +28,47 @@ import (
...
+		return true, nil
+	}
+	return false, nil
+}
+
+func init() {
+	RegisterErrorHandler(&TaskInProgressHandler{})
+}
+
+func RegisterErrorHandler(handler ErrorHandler) {
```
hmahmood: With a global list of error handlers, callers to WaitForResult may get unexpected results, since multiple handlers may be called, and handling depends on the order they were added. Why don't we just have another argument to WaitForResult for the handler?

pkg/vsphere/tasks/waiter.go
```diff
@@ -27,12 +28,47 @@ import (
...
+	VMContextObjectKey = "VMContextObject"
+)
+
+var (
+	m             sync.RWMutex
+	errorHandlers []ErrorHandler
 )
 
+type ErrorHandler interface {
+	// Handle returns true if the error is expected. Otherwise, return false. If any error happens during error handling, return (true, error).
```
anchal-agrawal: > ... Otherwise, returns false. If any error happens during error handling, returns (true, error).

lib/portlayer/exec/fix_container.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/vmware/vic/pkg/vsphere/vm"
+)
+
+type ContainerFixer struct {
+}
+
+func InitErrorHandler() {
+	tasks.RegisterErrorHandler(&ContainerFixer{})
+}
+
```
anchal-agrawal: A function description would be nice to have here.

pkg/vsphere/tasks/waiter.go
emlin: hmm... We need different logic for different errors. For this invalid state errors, one additional parameter definitely works. But for a common function, signature change means lots of changes unrelated.
And in the future we'll find more errors, and more error handling logic from anywhere might be required.
This separate error handling registration will be a more flexible solution compared to the signature change, I think.

Taking about the multiple handlers and execution sequence, I think this is defined for error handling only. Which means each time, there is one definitive error happens, and suppose only one corresponding error handler should be executed. Every other handlers will figure out that's not the one they can help, and return false without error. Which means next error handler will be tried.
So the execution sequence should not impact the result.

hmahmood: Yes, but you can have conflicts between handlers, where two handlers could behave differently for the same error.

Maybe we need to add another function that does take in a handler, and have the current function call this new one with some default error handlers. This would maintain the current behavior, allowing us to add more "default" handlers, while adding an option to add additional handlers per call.

emlin:
@hickeng I prefer to delay the change to switch ctx to operation in tasks.WaitForResult method, cause that will involve tons of changes.
After most of portlayer methods switch to operation, it will be more straightforward to do it in tasks package.
So now, I didn't add that change. Let me know for any concerns.

lib/portlayer/exec/fix_container.go
```diff
@@ -0,0 +1,65 @@
...
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/tasks"
+	"github.com/vmware/vic/pkg/vsphere/vm"
+)
+
+// InitErrorHandler register ContainerFixer to tasks error handler. So if any container vm operations fail for invalid state issue, this handler will be executed to fix the error.
+// The benefit to register handler in portlayer is that it can set container state during vm fixing.
+func InitErrorHandler() {
```
hmahmood: `InitTaskErrorHandlers()` maybe, since we could register more error handlers in the future, and the handlers are specifically for handling errors from tasks.

lib/portlayer/exec/fix_container.go
```diff
@@ -0,0 +1,65 @@
...
+	tasks.RegisterErrorHandler(FixContainerHandler)
+}
+
+func FixContainerHandler(ctx context.Context, err error) (bool, error) {
```
hmahmood: Does this need to be exported from the package?

pkg/vsphere/tasks/waiter.go
```diff
@@ -27,12 +28,40 @@ import (
...
+
+	VMContextObjectKey = "VMContextObject"
+)
+
+var (
+	m             sync.RWMutex
+	errorHandlers []func(ctx context.Context, err error) (bool, error)
 )
 
+func TaskInProgressHandler(ctx context.Context, err error) (bool, error) {
```
hmahmood: Ditto... does this need to be exported?

pkg/vsphere/tasks/waiter.go
```diff
@@ -27,12 +28,40 @@ import (
...
 
 const (
 	maxBackoffFactor = int64(16)
+
+	VMContextObjectKey = "VMContextObject"
+)
+
+var (
+	m             sync.RWMutex
+	errorHandlers []func(ctx context.Context, err error) (bool, error)
```
hmahmood: Define a type for the function type, e.g. `type ErrorHandler func(ctx context.Context, err error) (bool, error)`

hmahmood: Also, does the handler need to have additional context (not the go package), i.e. additional arguments? Like maybe task info, or the object that was involved in the task?

emlin: > Also, does the handler need to have additional context (not the go package), i.e. additional arguments? Like maybe task info, or the object that was involved in the task?

Now we're leveraging context to keep key/value for handler required objects, that's used in fix_container handler. I would rather to keep that behavior.
In the future, if the error handler relies on task or taskinfo, we can add them into context. I didn't do that right now, because no handler need that at this time yet.

lib/portlayer/exec/fix_container.go
emlin: good point

matthewavery:
@emlin I actually had a rejected PR that contained most of the `operation` logging changes to faciltate that change. the PR was rejected at the time as it was too risky to include in 0.8.0. If we do touch the tasks package I second the desire to see queuing added since I have been thinking about it since I was there last. But that is possibly another ticket since the concurrent testing should be added with it if possible to show we are more resilient to failures there. @hickeng Do you think the queuing thing should be separated out as it's own thing? it effects quite a bit and this PR could get much more sizeable with that change.

lib/portlayer/exec/fix_container.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/vmware/vic/pkg/vsphere/vm"
+)
+
+// InitErrorHandler register ContainerFixer to tasks error handler. So if any container vm operations fail for invalid state issue, this handler will be executed to fix the error.
+// The benefit to register handler in portlayer is that it can set container state during vm fixing.
+func InitTasksErrorHandler() {
+	tasks.RegisterErrorHandler(fixContainerHandler)
+}
+
+func fixContainerHandler(ctx context.Context, err error) (bool, error) {
```
matthewavery: Should we lock the container object here? I notice we also store the old state but it is still possible for the state to change after that right? I might have just missed the locking or something.

lib/portlayer/exec/fix_container.go
```diff
@@ -0,0 +1,65 @@
...
+	if !vm.IsInvalidState(ctx) {
+		log.Debugf("VM is not in invalid state, do not fix failure")
+		return false, nil
+	}
+	log.Debugf("Try to fix failure %s", err)
+	container := Containers.Container(vm.Reference().String())
+	if container != nil {
+		oldState := container.CurrentState()
+		defer container.SetState(oldState)
+		container.SetState(StateFixing)
```
matthewavery: refactor SetState to return the old state, and reject state changes altogether if the current state is `fixing`. Then add a separate function for switching back from `fixing` state. This will prevent the race condition.

lib/portlayer/event/collector/vsphere/vm_event.go
```diff
@@ -39,16 +39,12 @@ func NewVMEvent(be types.BaseEvent) *VMEvent {
...
-	case *types.VmRegisteredEvent:
-		ee = events.ContainerRegistered
 	case *types.VmMigratedEvent:
 		ee = events.ContainerMigrated
 	case *types.DrsVmMigratedEvent:
 		ee = events.ContainerMigratedByDrs
 	case *types.VmRelocatedEvent:
 		ee = events.ContainerRelocated
-	default:
-		panic("Unknown event")
```
vburenin: Do not remove default clause. We either need to log a missed event or explicitly assign empty string here.

cgtexmex: we shouldn't log for the default case -- we are listening for all events and filtering currently in code.  If we logged for all events we don't care about then we'll be logging ALOT.

cgtexmex: Actually ignore my comment -- we could log there....I'm curious how we hit that anywhere...

emlin: got this panic for removed registered event here, but do not in collector.
done

emlin:
@hickeng could you please help to review this PR, there still have lots of failure in our CI failed for this issue.

lib/install/management/appliance.go
```diff
@@ -149,6 +149,20 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
...
+	// get the actual folder name before we delete it
+	folder, err := vm.FolderName(d.ctx)
+	if err != nil {
+		// failed to get folder name, might not be able to remove files for this VM
+		name := d.getName(vm)
+		if name == "" {
+			log.Errorf("Unable to automatically remove all files in datastore for VM %q", vm.Reference())
+		} else {
+			// try to use the vm name in place of folder
+			log.Infof("Delete will attempt to remove datastore files for VM %q", name)
```
vburenin: I think this log message needs a better phrasing. However, I am not the best person to suggest a better text here. @matthewavery ?

lib/install/validate/fix_container.go
```diff
@@ -0,0 +1,58 @@
...
+
+func fixVMInvalidStateHandler(ctx context.Context, err error) (bool, error) {
+	defer trace.End(trace.Begin(fmt.Sprintf("error: %s", err)))
+	o := ctx.Value(tasks.VMObjectKey)
+	if o == nil {
+		log.Debugf("No vm object set, not vm operations.")
+		return false, nil
+	}
+
+	vm, ok := o.(*vm.VirtualMachine)
```
vburenin: How likely is that expected type is different here?

lib/install/validate/fix_container.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
```
vburenin: I see no unit tests for this file. Please, talk to me in person before you write one. I can show you some good practices.

lib/portlayer/exec/fix_container.go
```diff
@@ -0,0 +1,64 @@
...
+
+// InitErrorHandler register ContainerFixer to tasks error handler. So if any container vm operations fail for invalid state issue, this handler will be executed to fix the error.
+// The benefit to register handler in portlayer is that it can set container state during vm fixing.
+func InitTasksErrorHandler() {
+	tasks.RegisterErrorHandler(fixContainerHandler)
+}
+
+func fixContainerHandler(ctx context.Context, err error) (bool, error) {
+	defer trace.End(trace.Begin(fmt.Sprintf("error: %s", err)))
+	o := ctx.Value(tasks.VMObjectKey)
```
vburenin: I see this code block second time, it is time to build a helper tool to get specific value types from context.

pkg/vsphere/tasks/waiter.go
```diff
@@ -27,12 +28,42 @@ import (
 	"github.com/vmware/govmomi/vim25/progress"
 	"github.com/vmware/govmomi/vim25/soap"
 	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/pkg/trace"
 )
 
 const (
 	maxBackoffFactor = int64(16)
```
vburenin: maxBackoffFactor int64 = 16

emlin:
close this PR for most of the comments and changes are based on error handler solution. Reopen a new PR #3722 with the initial solution.

PR #3481, opened 2016-12-09
A huge migration to a newer swagger.

vburenin:
What can I say? It is a HUGE change. Full CI passed, which is exciting and scary at the same time. Not sure there is a good way to review it all carefully, pick a part you are familiar with and see what happens.

Fixes: #929


mdubya66:
fixes #3481 ?

lib/apiservers/engine/backends/container_test.go
```diff
@@ -615,7 +616,7 @@ func TestPortInformation(t *testing.T) {
 	mockContainerInfo := &plmodels.ContainerInfo{}
 	mockContainerConfig := &plmodels.ContainerConfig{}
 	containerID := "foo"
-	mockContainerConfig.ContainerID = &containerID
+	mockContainerConfig.ContainerID = containerID
```
hmahmood: There are a lot of these changes where we are taking away the pointer. Can we not do these in a separate PR; preferably in an iterative manner? These kinds of changes always have the danger of introducing subtle bugs.

vburenin: This is the least problematic change you can think of - at least for these values I was able to check the code surrounding it. The most scariest part is when I set "x-nullable":false - this is where zero values are no longer accepted. I tried to be careful with them, but things that are changed are not as scary as those that are unchanged.

hmahmood: Well, yes, this one in particular is not problematic. But that doesn't mean the multitude of others is not; its just not easy to go through everyone and understand the ramifications of a particular change. Is the "x-nullable" really necessary right now?

mdubya66:
My take on this is let's get it tested as much as possible. Run against the CI, the Nightlies, Stress and Longevity. We should aim to get it in sooner so that we can continue to iterate on it.

lib/apiservers/engine/backends/container_test.go
vburenin: @hmahmood without it the change would an order of a magnitude larger.

vburenin: @hmahmood the good part of this change is that it will expose problematic parts of our API and data checks earlier. What really worries me is that we do not really care about data validation relaying purely on a sane usage. At the moment it is fine, but AFAIK PortLayer API will be exposed in the future, so absence of data validation will bite us.

lib/apiservers/engine/backends/container.go
```diff
@@ -1521,17 +1531,19 @@ func portInformation(t *models.ContainerInfo, ips []netlink.Addr) []types.Port {
 	// (works with both IPv4 and IPv6 addresses)
 	var ports []types.Port
 
-	container := cache.ContainerCache().GetContainer(*t.ContainerConfig.ContainerID)
-	if container == nil {
-		log.Errorf("Could not find container with ID %s", *t.ContainerConfig.ContainerID)
+	cid := t.ContainerConfig.ContainerID
+	c := cache.ContainerCache().GetContainer(cid)
```
sflxn: +1.  I really hate go programmer's habit of overloading one line with a bazillion operations.  They need to remember the compiler doesn't care, and readability is more important.

lib/apiservers/engine/backends/backends.go
```diff
@@ -26,12 +26,12 @@ import (
...
 	"github.com/docker/docker/registry"
-	httptransport "github.com/go-swagger/go-swagger/httpkit/client"
-	"github.com/go-swagger/go-swagger/swag"
+	rc "github.com/go-openapi/runtime/client"
+	"github.com/go-openapi/swag"
 
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/container"
-	"github.com/vmware/vic/lib/apiservers/portlayer/client"
+	apiclient "github.com/vmware/vic/lib/apiservers/portlayer/client"
```
fdawg4l: `apiclient` isn't really necessary to migrate from 1 swagger version to another.  We can debate the naming of this later.  Lets keep this change restricted only only the _required_ changes to include the new swagger version.  The unnecessary churn makes reviewing harder.

lib/apiservers/engine/backends/container.go
```diff
@@ -985,22 +984,25 @@ func (c *Container) ContainerInspect(name string, size bool, version version.Ver
 	}
 	var started time.Time
 	var stopped time.Time
-	if results.Payload.ProcessConfig.StartTime != nil && *results.Payload.ProcessConfig.StartTime > 0 {
```
vburenin: This is not a cosmetic change. it a result of a variable becoming a value not a pointer during migration.

lib/apiservers/engine/backends/backends.go
```diff
@@ -123,52 +124,52 @@ func hydrateCaches() error {
 
 	wg := sync.WaitGroup{}
 	wg.Add(waiters)
-	errors := make(chan error, waiters)
+	errChan := make(chan error, waiters)
```
vburenin: This is the cosmetic change and the only one that you can say "big".

lib/apiservers/engine/backends/backends.go
vburenin: I used tools to deal with imports, all automatic changes are not easy to manage manually.

lib/apiservers/engine/backends/system.go
```diff
@@ -159,31 +159,31 @@ func (s *System) SystemInfo() (*types.Info, error) {
...
+		if vchInfo.CPUMhz > 0 {
+			info.NCPU = int(vchInfo.CPUMhz)
 
 			customInfo := [2]string{systemStatusMhz, fmt.Sprintf("%d Mhz", info.NCPU)}
 			info.SystemStatus = append(info.SystemStatus, customInfo)
 		}
-		if vchInfo.Memory != nil {
-			info.MemTotal = *vchInfo.Memory << 20 //Multiply by 1024*1024 to get Mebibytes
+		if vchInfo.Memory > 0 {
+			info.MemTotal = vchInfo.Memory * 1024 * 1024 // Get Mebibytes
```
caglar10ur: come one - 2^20 FTW :)

lib/apiservers/templates/server/server.gotmpl
```diff
@@ -1,225 +0,0 @@
-package {{ .APIPackage }}
```
caglar10ur: we needed this for 030764afb6f7475524844bd94c9cdca2edf5510d . How do we handle it with new swagger?

lib/apiservers/engine/backends/container.go
sflxn: @fdawg4l There are a few cosmetic changes, but I would gloss over them and focus on the pointer issue.  Vast majority of the changes I've seen so far are due to Swagger and not cosmetic.

lib/apiservers/engine/backends/system.go
sflxn: I'd prefer to keep it as is also.

lib/apiservers/portlayer/restapi/handlers/misc_handlers.go
```diff
@@ -29,28 +28,28 @@ import (
...
-	api.MiscPingHandler = misc.PingHandlerFunc(handler.Ping)
-	api.MiscGetVCHInfoHandler = misc.GetVCHInfoHandlerFunc(handler.GetVCHInfo)
+func (h *MiscHandlersImpl) Configure(api *operations.PortLayerAPI, handlerCtx *HandlerContext) {
+	api.MiscPingHandler = misc.PingHandlerFunc(h.Ping)
+	api.MiscGetVCHInfoHandler = misc.GetVCHInfoHandlerFunc(h.GetVCHInfo)
 }
 
 // Ping sends an OK response to let the client know the server is up
-func (handler *MiscHandlersImpl) Ping() middleware.Responder {
+func (h *MiscHandlersImpl) Ping(param misc.PingParams) middleware.Responder {
```
sflxn: Parameter-less operations in the swagger spec now causes parameters to be generated?

vburenin: yes, they contain some useful objects like context, request, etc.

lib/apiservers/portlayer/swagger.json
```diff
@@ -2544,11 +2574,12 @@
...
 				},
 				"Store": {
+					"x-nullable": false,
 					"type": "string"
 				},
 				"Capacity": {
-					"type": "integer",
-					"format": "int64"
+					"x-nullable": false,
+					"type": "integer"
```
sflxn: Why is this a type integer instead of int64 now?

vburenin: format != type.

lib/apiservers/portlayer/swagger.json
```diff
@@ -2660,11 +2697,11 @@
 					"type": "string"
 				},
 				"repoName": {
+					"x-nullable": true,
 					"type": "string"
 				},
 				"createTime": {
-					"type": "integer",
-					"format": "int64"
+					"type": "integer"
```
sflxn: Is Swagger defaulting to in64 now?  I believe unix time is int64.

lib/apiservers/portlayer/swagger.json
```diff
@@ -2721,8 +2763,7 @@
 					"$ref": "#/definitions/ReservationConfig"
 				},
 				"storageSize": {
-					"type": "integer",
-					"format": "int64"
+					"type": "integer"
```
sflxn: Same as above.  Why integer instead of int64?

lib/apiservers/engine/backends/system.go
vburenin: come one - compiler will do the same thing.

lib/apiservers/portlayer/swagger.json
vburenin: it always was :-|

lib/apiservers/portlayer/swagger.json
sflxn: The version of Swagger we used caused an int64 to get generated if we tag the parameter with format.  Latest version of Swagger no longer does this?

lib/apiservers/portlayer/swagger.json
vburenin: if it wasn't it would not compile.

lib/apiservers/portlayer/swagger.json
sflxn: After our slack conversation, I see that integer now defaults to int64 in the latest Swagger generator.

sflxn:
Nothing seemed out of the ordinary.  We'll just have to be aware of x-nullable and integer generation changes.  Is this the extent of all the changes necessary?

lgtm

lib/apiservers/engine/backends/system.go
sflxn: ok, if go compiler is smart to do this.  Still hurts the eyes of former C/C++ programmers. :)

lib/apiservers/templates/server/server.gotmpl
vburenin: @caglar10ur I integrated needed changes into swagger as well as updated portlayer accordingly.
https://github.com/go-swagger/go-swagger/pull/784

caglar10ur:
LGTM

mdubya66:
LGTM

PR #3519, opened 2016-12-15
[full ci] Take network mask from IP address instead of gateway.

### LINKED ISSUE #3517: Improve static IP and Gateway configuration
Currently network mask is set as a part of gateway parameter which is wrong. Network size has to be defined by static IP address mask.

vburenin:
Fixes #3517 #3410


hmahmood:
lgtm

cmd/vic-machine/create/create.go
```diff
@@ -212,14 +212,14 @@ func (c *Create) Flags() []cli.Flag {
 		cli.StringFlag{
 			Name:        "client-network-gateway",
 			Value:       "",
-			Usage:       "Gateway for the VCH on the client network, including one or more routing destinations in a comma separated list, e.g. 10.1.0.0/16,10.2.0.0/16:10.0.0.1/24",
+			Usage:       "Gateway for the VCH on the client network, including one or more routing destinations in a comma separated list, e.g. 10.1.0.0/16,10.2.0.0/16:10.0.0.1",
```
vburenin: @emlin The whole point of routing is that you define a list of networks you would like to reach out and to reach them out you need a gateway, so you provide an IP address of that gateway.
There is also a default gateway that can be defined as 0.0.0.0/0 - this one is picked if there is no other match.

emlin: thanks

emlin:
lgtm

PR #3521, opened 2016-12-15
[skip ci] Remove only the specific test vms on nightly teardown

rajanashok:
The teardown only deletes the VMs specific to the test. This is needed to support running multiple nightly runs on the same test account so we can run both 6.0 and 6.5 ESX versions.


anchal-agrawal:
LGTM

mhagen-vmware:
lgtm

mhagen-vmware:
nice job!

rajanashok:
.

PR #3523, opened 2016-12-16
vcsim: network related enhancements

dougm:
- Add DVS support

- Add host Vswitch/Portgroup support

- Add MoveIntoFolder_Task support

- Add CreateTask helper

### DIFF
--- a/pkg/vsphere/simulator/folder.go
+++ b/pkg/vsphere/simulator/folder.go
@@ -305,3 +305,54 @@ func (f *Folder) RegisterVMTask(c *types.RegisterVM_Task) soap.HasFault {
 
 	return r
 }
+
+func (f *Folder) MoveIntoFolderTask(c *types.MoveIntoFolder_Task) soap.HasFault {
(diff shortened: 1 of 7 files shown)


caglar10ur:
LGTM

pkg/vsphere/simulator/folder.go
```diff
@@ -305,3 +305,54 @@ func (f *Folder) RegisterVMTask(c *types.RegisterVM_Task) soap.HasFault {
 
 	return r
 }
+
+func (f *Folder) MoveIntoFolderTask(c *types.MoveIntoFolder_Task) soap.HasFault {
+	task := CreateTask(f, "moveIntoFolder", func(t *Task) (types.AnyType, types.BaseMethodFault) {
+		for _, ref := range c.List {
+			obj := Map.Get(ref).(mo.Entity)
```
vburenin: Is it safe to do a type assertion here without checking if it actually happened?

dougm: Yes, we do this quite a bit in package simulator.  If the type is not what we expect, its a bug in the simulator or one's test code, in which case we want the panic.

PR #3539, opened 2016-12-20
[skip ci] Make the email generation more dynamic

mhagen-vmware:
This should make the nightly execution file more readable and make it easier to add tests to the email.  I think now all you should need to do is add a new test to the nightly_list_var at the top of the file.


dougm:
lgtm

matthewavery:
LGTM

rajanashok:
Thanks for making this even more simpler. I assume the nightlystatus counter in the template is now reset between 6.5 and 6.0. It would end up setting the same status for both 6.5 and 6.0. Please correct me if I am wrong.

When I set the nightly status I keep incrementing the count, so it goes from 0 to 29.

rajanashok:
LGTM

PR #3540, opened 2016-12-20
[specific ci=1-06-Docker-Run] Add new test that checks quick exit run…

### LINKED ISSUE #1687: incorrect behaviour: `docker run busybox`
Running `docker run busybox`

Regular docker: exits immediately - presumably because stdin is closed and `ash` will exit in that case

VIC: doesn't exit, but also doesn't allow interaction - presumably we're not closing stdin

mhagen-vmware:
fixes #1687


anchal-agrawal:
Not sure if we should close #1687 with this PR, unless the issue described in that ticket has been fixed already.

mhagen-vmware:
It has been fixed, as this test passes now.

matthewavery:
LGTM

PR #3556, opened 2017-01-03
[skip ci] Setting user tags to identify vms during cleanup

rajanashok:
Missed the change to set the user variable in my previous commit. Setting the ${user} to %{NIMBUS_USER} to tag the vm name so it can come in handy to identify the vms during the cleanup flow.


rajanashok:
.

PR #3557, opened 2017-01-03
[specific ci=Group6-VIC-Machine]fix dc suggestion issue in vic-machine ls

### LINKED ISSUE #2859: vic-machine: doesn't handle datacenter `/` and doesn't suggest options with `ls`
Running against a local ESX I see the following when target has a trailing `/`, e.g. `https://user:password@esx/`

```
INFO[2016-10-23T16:16:44Z] ### Installing VCH ####
time=2016-10-23T16:16:44.152291463Z level=debug msg=[BEGIN] [github.com/vmware/vic/lib/install/validate.CreateNoDCCheck:80]
DEBU[2016-10-23T16:16:44Z] Using to login by username/password
DEBU[2016-10-23T16:16:44Z] Session Environment Info:                     API Type=HostAgent API Version=6.0 Build=3380124 Name=VMware ESXi OS Type=vmnix-x86 Product ID=embeddedEsx UUID= Vendor=VMware, Inc. Version=6.0.0
DEBU[2016-10-23T16:16:44Z] vSphere resource cache populating...
DEBU[2016-10-23T16:16:44Z] Error count populating vSphere cache: (5)
time=2016-10-23T16:16:44.227323863Z level=debug msg=[ END ] [github.com/vmware/vic/lib/install/validate.CreateNoDCCheck:80] [75.013344ms]
ERRO[2016-10-23T16:16:44Z] Datacenter must be specified in --target (e.g. https://addr/datacenter)
time=2016-10-23T16:16:44.227859362Z level=debug msg=[BEGIN] [github.com/vmware/vic/lib/install/validate.(*Validator).suggestDatacenter:170]
INFO[2016-10-23T16:16:44Z] Suggesting valid values for datacenter in --target
INFO[2016-10-23T16:16:44Z] Suggested datacenters:
INFO[2016-10-23T16:16:44Z]   "ha-datacenter"
time=2016-10-23T16:16:44.234730178Z level=debug msg=[ END ] [github.com/vmware/vic/lib/install/validate.(*Validator).suggestDatacenter:170] [6.746422ms]
ERRO[2016-10-23T16:16:44Z] Create cannot continue: failed to create validator
ERRO[2016-10-23T16:16:44Z] --------------------
ERRO[2016-10-23T16:16:44Z] vic-machine-linux failed: Datacenter must be specified in --target (e.g. https://addr/datacenter)
```

Also for `ls` with that same target there's no suggestion at all:

```
INFO[2016-10-23T16:18:41Z] ### Listing VCHs ####
INFO[2016-10-23T16:18:41Z] Validating target
ERRO[2016-10-23T16:18:41Z] --------------------
ERRO[2016-10-23T16:18:41Z] Failed to get host system: please specify a datacenter
ERRO[2016-10-23T16:18:41Z] List cannot continue - target validation failed: validation of configuration failed
ERRO[2016-10-23T16:18:41Z] --------------------
ERRO[2016-10-23T16:18:41Z] vic-machine-linux failed: validation of configuration failed
```

Datacenter selection and validation should be a common path for all vic-machine commands as it's an inherent part of the target, which is required for all commands.

emlin:
fixes #2859 

This is to fix the datacenter suggestion issue in vic-machine ls

### DIFF
--- a/cmd/vic-machine/delete/delete.go
+++ b/cmd/vic-machine/delete/delete.go
@@ -119,12 +119,12 @@ func (d *Uninstall) Run(clic *cli.Context) (err error) {
 	validator, err := validate.NewValidator(ctx, d.Data)
 	if err != nil {
 		log.Errorf("Delete cannot continue - failed to create validator: %s", err)
-		return errors.New("Delete failed")
+		return errors.New("delete failed")
 	}
 	_, err = validator.ValidateTarget(ctx, d.Data)
 	if err != nil {
 		log.Errorf("Delete cannot continue - target validation failed: %s", err)
-		return errors.New("Delete failed")
+		return errors.New("delete failed")
 	}
 
 	executor := management.NewDispatcher(validator.Context, validator.Session, nil, d.Force)
@@ -138,7 +138,7 @@ func (d *Uninstall) Run(clic *cli.Context) (err error) {
 	if err != nil {
 		log.Errorf("Failed to get Virtual Container Host %s", d.DisplayName)
 		log.Error(err)
-		return errors.New("Delete failed")
+		return errors.New("delete failed")
 	}
 
 	log.Infof("")
@@ -148,7 +148,7 @@ func (d *Uninstall) Run(clic *cli.Context) (err error) {
 	if err != nil {
 		log.Error("Failed to get Virtual Container Host configuration")
 		log.Error(err)
-		return errors.New("Delete failed")
+		return errors.New("delete failed")
 	}
 
 	// compare vch version and vic-machine version
@@ -156,7 +156,7 @@ func (d *Uninstall) Run(clic *cli.Context) (err error) {
 	if vchConfig.Version == nil || !installerBuild.Equal(vchConfig.Version) {
 		if !d.Data.Force {
 			log.Errorf("VCH version %q is different than installer version %s. Specify --force to force delete", vchConfig.Version.ShortVersion(), installerBuild.ShortVersion())
-			return errors.New("Delete failed")
+			return errors.New("delete failed")
 		}
 
 		log.Warnf("VCH version %q is different than installer version %s. Force delete will attempt to remove everything related to the installed VCH", vchConfig.Version.ShortVersion(), installerBuild.ShortVersion())
@@ -166,7 +166,7 @@ func (d *Uninstall) Run(clic *cli.Context) (err error) {
 	if err = executor.DeleteVCH(vchConfig); err != nil {
 		executor.CollectDiagnosticLogs()
 		log.Errorf("%s", err)
-		return errors.New("Delete failed")
+		return errors.New("delete failed")
 	}
 
 	log.Infof("Completed successfully")
--- a/cmd/vic-machine/delete/delete.go
+++ b/cmd/vic-machine/delete/delete.go
@@ -119,8 +119,14 @@ func (d *Uninstall) Run(clic *cli.Context) (err error) {
 	validator, err := validate.NewValidator(ctx, d.Data)
 	if err != nil {
 		log.Errorf("Delete cannot continue - failed to create validator: %s", err)
-		return err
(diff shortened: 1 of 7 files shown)


lib/install/validate/validator.go
```diff
@@ -161,18 +156,27 @@ func CreateNoDCCheck(ctx context.Context, input *data.Data) (*Validator, error)
...
+	var detail string
+	if v.DatacenterPath != "" {
+		detail = fmt.Sprintf("Datacenter %q in --target is not found", strings.TrimPrefix(v.DatacenterPath, "/"))
+	} else {
+		// this means multiple datacenter exists, but user did not specify it in --target
+		detail = "Datacenter must be specified in --target (e.g. https://addr/datacenter)"
 	}
-	v.DatacenterPath = v.Session.Datacenter.InventoryPath
-	return nil
+	log.Error(detail)
```
vburenin: As a good SOLID practice, It is not validator's job to log errors. Only caller should be logging an error when needed.

cmd/vic-machine/list/list.go
```diff
@@ -161,7 +161,7 @@ func (l *List) Run(clic *cli.Context) (err error) {
 
 	var validator *validate.Validator
 	if l.Data.ComputeResourcePath == "" {
-		validator, err = validate.CreateNoDCCheck(ctx, l.Data)
+		validator, err = validate.NewValidatorAllowEmptyDC(ctx, l.Data, true)
```
vburenin: Using different constructors to handle different cases is not the best practice.
To make it looking better you should add a separate method that would allow EmptyDC.
Something like.
validator.AllowEmptyDC().AllowEmptyWaterTank().AllowNoHumans()

This, you may extend functionality in a less harmless way, otherwise it gets really hacky.

lib/install/validate/validator.go
emlin: I understand your point, but current suggestion methods logging out message. We should have a separate PR for that refactor.

cmd/vic-machine/list/list.go
emlin: good point

cmd/vic-machine/list/list.go
```diff
@@ -161,24 +161,26 @@ func (l *List) Run(clic *cli.Context) (err error) {
...
 		return errors.New("list failed")
 	}
+	// If dc is not set, and multiple datacenter is available, vic-machine ls will list VCHs under all datacenters.
+	validator.AllowEmptyDC()
 
 	_, err = validator.ValidateTarget(ctx, l.Data)
 	if err != nil {
 		log.Errorf("List cannot continue - target validation failed: %s", err)
-		return err
+		return errors.New("list failed")
```
vburenin: Uppercase for List?

cmd/vic-machine/delete/delete.go
```diff
@@ -142,15 +148,15 @@ func (d *Uninstall) Run(clic *cli.Context) (err error) {
...
+		return errors.New("Delete failed")
 	}
 
 	// compare vch version and vic-machine version
 	installerBuild := version.GetBuild()
 	if vchConfig.Version == nil || !installerBuild.Equal(vchConfig.Version) {
 		if !d.Data.Force {
 			log.Errorf("VCH version %q is different than installer version %s. Specify --force to force delete", vchConfig.Version.ShortVersion(), installerBuild.ShortVersion())
-			return errors.New("delete failed")
+			return errors.New("Delete failed")
```
vburenin: I suspect it should be lowercase "delete" since you use lowercased error in other places.

cmd/vic-machine/list/list.go
vburenin: As it happened, it should be as is, but lowercase should be used for "delete" error.

emlin: As you mentioned, delete and upgrade are changed to lower case

andrewtchin:
lgtm

PR #3558, opened 2017-01-04
handle maintanence mode

### LINKED ISSUE #3492: Open session with container in DRS cluster blocks host maintenance mode
So I know we're aware of this limitation in the team, but I wonder whether we've doing enough to document or deal with the potential fallout from this limitation.

It's one thing to say that an open session inhibits intentional vMotion. In that scenario, it fails immediately, you get a nice message from vSphere saying you have an open serial port and you know that it's *this* VM you can't vMotion at this time.

However, in the case where a vSphere admin wants to put a host into maintenance mode, a single containerVM with an open session can block it indefinitely with no message and there's no way to tell which containerVM is responsible. If the user exits the session, the live migration completes and the host enters maintenance mode.

This is pretty ugly behavior, since a user is effectively able to block the day-to-day operations of the admin. This is not something you could argue a user could do in a non-VIC scenario, since they typically won't have access to configure VMs.

In addition, there's the bug Caglar found about file-backed serial ports not being re-connected after vMotion #3243 which means that logs will simply stop functioning.

caglar10ur:
- Start listening VmRelocatedEvent and call refresh when needed

Otherwise the cache will have incorrect data due to DRS. DRS can create
the vm on one host and then power it on on another (hence relocation).

- Add new host event type and use it in exec subsystem

We receive HostEnteringMaintenanceMode when the host goes into
maintanence mode. When this happens all virtual machine operations are
blocked, except the following;

* MigrateVM
* PowerOffVM
* SuspendVM
* ShutdownGuest
* StandbyGuest

All the vms that does not have an active serial connections will end up
getting migrated by DRS. The rest needs to be powered off manually.

We can't call Commit since we are not allowerd to call reconfigure so
we now have a EmergencyCommit variant which sets the Spec to nil.

Fixes #3492

### DIFF
--- a/lib/portlayer/event/collector/vsphere/host_event.go
+++ b/lib/portlayer/event/collector/vsphere/host_event.go
@@ -1,0 +1,56 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package vsphere
+
+import (
+	"github.com/vmware/vic/lib/portlayer/event/events"
+
+	"github.com/vmware/govmomi/vim25/types"
+)
+
+type HostEvent struct {
+	*events.BaseEvent
+}
+
+func NewHostEvent(be types.BaseEvent) *HostEvent {
+	var ee string
+	// host events that we care about
+	switch be.(type) {
+	case *types.EnteringMaintenanceModeEvent:
+		ee = events.HostEnteringMaintenanceMode
+	case *types.EnteredMaintenanceModeEvent:
+		ee = events.HostEnteredMaintenanceMode
+	case *types.ExitMaintenanceModeEvent:
+		ee = events.HostExitMaintenanceMode
+	}
+	e := be.GetEvent()
+	return &HostEvent{
+		&events.BaseEvent{
+			Event:       ee,
+			ID:          int(e.Key),
+			Detail:      e.FullFormattedMessage,
+			Ref:         e.Host.Host.String(),
+			CreatedTime: e.CreatedTime,
+		},
+	}
+
+}
+
+func (he *HostEvent) Topic() string {
+	if he.Type == "" {
+		he.Type = events.NewEventType(he)
+	}
+	return he.Type.Topic()
+}
--- a/lib/portlayer/event/collector/vsphere/host_event.go
+++ b/lib/portlayer/event/collector/vsphere/host_event.go
@@ -34,6 +34,8 @@ func NewHostEvent(be types.BaseEvent) *HostEvent {
 		ee = events.HostEnteredMaintenanceMode
 	case *types.ExitMaintenanceModeEvent:
 		ee = events.HostExitMaintenanceMode
+	default:
+		panic("Unknown event")
 	}
 	e := be.GetEvent()
 	return &HostEvent{
--- a/lib/portlayer/event/collector/vsphere/vm_event.go
+++ b/lib/portlayer/event/collector/vsphere/vm_event.go
@@ -45,6 +45,9 @@ func NewVMEvent(be types.BaseEvent) *VMEvent {
 		ee = events.ContainerMigrated
 	case *types.DrsVmMigratedEvent:
 		ee = events.ContainerMigratedByDrs
+	case *types.VmRelocatedEvent:
+		ee = events.ContainerRelocated
+
 	}
 	e := be.GetEvent()
 	return &VMEvent{
--- a/lib/portlayer/event/collector/vsphere/vm_event.go
+++ b/lib/portlayer/event/collector/vsphere/vm_event.go
@@ -47,7 +47,8 @@ func NewVMEvent(be types.BaseEvent) *VMEvent {
 		ee = events.ContainerMigratedByDrs
 	case *types.VmRelocatedEvent:
 		ee = events.ContainerRelocated
-
+	default:
+		panic("Unknown event")
 	}
 	e := be.GetEvent()
 	return &VMEvent{
--- a/lib/portlayer/exec/exec.go
+++ b/lib/portlayer/exec/exec.go
@@ -99,15 +99,20 @@ func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSou
 		// create the event manager &  register the existing collector
 		Config.EventManager = event.NewEventManager(ec)
 
-		Config.EventManager.Subscribe(events.NewEventType(vsphere.HostEvent{}).Topic(), "host", func(ie events.Event) {
-			hostEventCallback(ctx, ie)
-		})
+		// subscribe to host events
+		Config.EventManager.Subscribe(events.NewEventType(vsphere.HostEvent{}).Topic(), "host",
+			func(ie events.Event) {
+				hostEventCallback(ctx, ie)
+			},
+		)
 		// subscribe the exec layer to the event stream for Vm events
 		Config.EventManager.Subscribe(events.NewEventType(vsphere.VMEvent{}).Topic(), "exec", eventCallback)
 		// subscribe callback to handle vm registered event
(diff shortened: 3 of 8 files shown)


lib/portlayer/event/collector/vsphere/host_event.go
```diff
@@ -0,0 +1,56 @@
...
+func NewHostEvent(be types.BaseEvent) *HostEvent {
+	var ee string
+	// host events that we care about
+	switch be.(type) {
+	case *types.EnteringMaintenanceModeEvent:
+		ee = events.HostEnteringMaintenanceMode
+	case *types.EnteredMaintenanceModeEvent:
+		ee = events.HostEnteredMaintenanceMode
+	case *types.ExitMaintenanceModeEvent:
+		ee = events.HostExitMaintenanceMode
```
vburenin: default: panic("Unknown event") ?
It is not going to happen, but from readability standpoint it is nice to have.

lib/portlayer/event/collector/vsphere/host_event.go
```diff
@@ -0,0 +1,56 @@
...
+			Detail:      e.FullFormattedMessage,
+			Ref:         e.Host.Host.String(),
+			CreatedTime: e.CreatedTime,
+		},
+	}
+
+}
+
+func (he *HostEvent) Topic() string {
+	if he.Type == "" {
```
vburenin: Is this potentially lazy Type init needed here? Can we move it into constructor?

lib/portlayer/event/collector/vsphere/vm_event.go
```diff
@@ -45,6 +45,9 @@ func NewVMEvent(be types.BaseEvent) *VMEvent {
 		ee = events.ContainerMigrated
 	case *types.DrsVmMigratedEvent:
 		ee = events.ContainerMigratedByDrs
+	case *types.VmRelocatedEvent:
+		ee = events.ContainerRelocated
+
```
vburenin: same here, it would be great to see panic here for default use case.

lib/portlayer/exec/exec.go
```diff
@@ -97,6 +99,9 @@ func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSou
 		// create the event manager &  register the existing collector
 		Config.EventManager = event.NewEventManager(ec)
 
+		Config.EventManager.Subscribe(events.NewEventType(vsphere.HostEvent{}).Topic(), "host", func(ie events.Event) {
```
vburenin: Could you start closure declaration from the next line please?

lib/portlayer/exec/exec.go
```diff
@@ -155,18 +159,123 @@ func eventCallback(ie events.Event) {
 						container.onStop()
 					}
 					log.Debugf("Container(%s) state set to %s via event activity",
-						container.ExecConfig.ID, newState.String())
+						container, newState.String())
```
vburenin: +1

lib/portlayer/exec/exec.go
```diff
@@ -155,18 +159,123 @@ func eventCallback(ie events.Event) {
...
 					// regardless of update success failure publish the container event
 					publishContainerEvent(container.ExecConfig.ID, ie.Created(), ie.String())
 				}()
 			case StateRemoved:
-				log.Debugf("Container(%s) %s via event activity", container.ExecConfig.ID, newState.String())
+				log.Debugf("Container(%s) %s via event activity", container, newState.String())
```
vburenin: I think there is no need to call String method for newState.

lib/portlayer/exec/exec.go
```diff
@@ -155,18 +159,123 @@ func eventCallback(ie events.Event) {
...
 				Containers.Remove(container.ExecConfig.ID)
 				publishContainerEvent(container.ExecConfig.ID, ie.Created(), ie.String())
 			}
+		} else {
+			switch ie.String() {
+			case events.ContainerRelocated:
+				log.Debugf("Container %s received ContainerRelocated", container)
```
vburenin: I think you should take event name from event, do not embed its name as a part of the string.

lib/portlayer/exec/exec.go
```diff
@@ -155,18 +159,123 @@ func eventCallback(ie events.Event) {
...
+	// ShutdownGuest
+	// StandbyGuest
+	//
+	// Because of that limitation the only thing that we can do is shutting down the guest without calling reconfigure
+	case events.HostEnteringMaintenanceMode:
+		log.Debugf("Received %s event", ie)
+		ref := ie.Reference()
+
+		// we are interested with running vms
+		state := new(State)
```
vburenin: State is an int type. Do we really need a pointer for it? It makes code looking ugly. Of course, it is not a scope of this PR.

vburenin: I see it is actually a relatively trivial change to fix the state pointer, but it is a big enough to do that in a different PR.

lib/portlayer/exec/exec.go
```diff
@@ -155,18 +159,123 @@ func eventCallback(ie events.Event) {
...
+			log.Debugf("%q is on %q", v, ref)
+
+			// grab the container from the cache
+			container := Containers.Container(v.String())
+			if container == nil {
+				log.Errorf("Container %s not found", v)
+				continue
+			}
+
+			operation := func() error {
```
vburenin: Do you really need a closure here? It makes it hard to read the code.

lib/portlayer/exec/exec.go
```diff
@@ -155,18 +159,123 @@ func eventCallback(ie events.Event) {
...
+
+				// this check needs to be after we get a new handle otherwise we could receive stalled data
+				needsToBePoweredOff := false
+
+				// get the virtual device list
+				devices := object.VirtualDeviceList(v.Config.Hardware.Device)
+
+				// select the virtual serial ports
+				serials := devices.SelectByBackingInfo((*types.VirtualSerialPortURIBackingInfo)(nil))
+				log.Debugf("Found %d devices withe the desired backing", len(serials))
```
vburenin: "with the"

lib/portlayer/exec/exec.go
```diff
@@ -155,18 +159,123 @@ func eventCallback(ie events.Event) {
...
+
+				// iterate over them and set needsToBePoweredOff if neccesary
+				for _, serial := range serials {
+					log.Debugf("Connected: %t", serial.GetVirtualDevice().Connectable.Connected)
```
vburenin: needsToBePoweredOff = serial.GetVirtualDevice().Connectable.Connected
log.Debug("Connected: %t", needsToBePoweredOff)

vburenin: if neesToBePoweredOff {
   break
}

lib/portlayer/exec/exec.go
```diff
@@ -155,18 +159,123 @@ func eventCallback(ie events.Event) {
...
 				}()
 			case StateRemoved:
-				log.Debugf("Container(%s) %s via event activity", container.ExecConfig.ID, newState.String())
+				log.Debugf("Container(%s) %s via event activity", container, newState.String())
 				Containers.Remove(container.ExecConfig.ID)
 				publishContainerEvent(container.ExecConfig.ID, ie.Created(), ie.String())
 			}
+		} else {
+			switch ie.String() {
+			case events.ContainerRelocated:
```
emlin: why this event is handled in else statement? I assume portlayer will not put container into relocated status, and then it's not possible to be same to container current state

lib/portlayer/exec/exec.go
caglar10ur: done

lib/portlayer/event/collector/vsphere/host_event.go
cgtexmex: I _believe_ we can change it (I thought I attempted to include in constructor during init dev, but can't recall why it fell out this way.) -- with that said it's pretty minor and I'd vote to change it in a later commit as part of clean up / refactor of event handling

lib/portlayer/exec/handle.go
```diff
@@ -204,6 +204,17 @@ func (h *Handle) Commit(ctx context.Context, sess *session.Session, waitTime *in
 	return nil
 }
 
+// EmergencyCommit sets the handle's spec to nil so that Commit operation only does a state change and won't touch the extraconfig
+func (h *Handle) EmergencyCommit(ctx context.Context, sess *session.Session, waitTime *int32) error {
```
hmahmood: Is there a need for this? Since there is no locking involved, you can just set `h.Spec = nil` before calling `Commit`. The name `EmergencyCommit` is kind of unusual as well.

caglar10ur: We could, but then that line could be gone missing during a refactor etc.. I thought having it as a function will make more sense. 

Renamed it to CommitWithoutSpec, let me know if you have suggestions.

hmahmood: That's good enough 👍

PR #3560, opened 2017-01-04
start stripping binaries

caglar10ur:
```
old:
    46M  docker-engine-server
    41M  port-layer-server
   2.2M  rpctool
    12M  tether-linux
    29M  vic-init
    34M  vic-machine-linux
    35M  vicadmin

   121M  appliance.iso
    62M  bootstrap.iso
```

```
new:
    24M  docker-engine-server
    21M  port-layer-server
   1.3M  rpctool
   7.4M  tether-linux
    15M  vic-init
    17M  vic-machine-linux
    18M  vicadmin

   110M  appliance.iso
    61M  bootstrap.iso
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -251,37 +251,37 @@ endif
 
 $(vic-init): $$(call godeps,cmd/vic-init/*.go)
 	@echo building vic-init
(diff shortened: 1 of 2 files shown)


vburenin:
We can also use compression for the binaries which we distribute outside of ISO images, size can go down even more. However, I guess it should be a separate PR.

cgtexmex:
lgtm

PR #3562, opened 2017-01-04
Removing restriction to specify gateway for client and management network [full ci]

### LINKED ISSUE #3453: vic-machine gateway option should not be required for client and management networks when using static ips
Currently, to configure static IPs for client and management networks, a corresponding gateway is required (via the `--<network>-gateway` options). This also forces the user to specify at least one destination for routing through the gateway, even though no gateway should be required if the network in question is L2 adjacent. 

The fix should be to make the gateway options optional for client and management networks only. If gateway is specified, the requirement to specify at least one routing destination should remain.

hmahmood:
Fixes #3453

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -804,12 +804,8 @@ func (c *Create) processNetwork(network *data.NetworkConfig, netName, pgName, st
 
 	network.Name = pgName
 
-	if staticIP == "" && gateway == "" {
-		return nil
-	}
-
-	if staticIP == "" || gateway == "" {
-		return fmt.Errorf("%s network IP and gateway must both be specified", netName)
+	if gateway != "" && staticIP == "" {
+		return fmt.Errorf("Gateway provided without static IP for %s network", netName)
 	}
 
 	defer func(net *data.NetworkConfig) {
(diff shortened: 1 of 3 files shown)


cmd/vic-machine/create/create.go
```diff
@@ -818,28 +814,35 @@ func (c *Create) processNetwork(network *data.NetworkConfig, netName, pgName, st
...
+	if gateway != "" {
+		network.Destinations, network.Gateway, err = parseGatewaySpec(gateway)
+		if err != nil {
+			return fmt.Errorf("Invalid %s network gateway: %s", netName, err)
+		}
 	}
 
-	network.IP.IP = ipAddr
-	network.IP.Mask = ipNet.Mask
+	if staticIP != "" {
```
emlin: I think you can parse staticIP first, and then parse gateway if it's not empty, and then you can merge two `if gateway != ""` statements into one section.

hmahmood: You would be checking for `staticIP != ""` then, so back to square one :-)

emlin: https://github.com/vmware/vic/pull/3562/commits/606ab2080689216f0e1bc83dbb87a3a3ff364e2b#diff-5a99f2371e79e7b3ef78eefdc3419581R807
if gateway is not empty, staticIP is always not empty. :)

hmahmood: Done

emlin:
lgtm

hmahmood:
@vburenin @emlin @andrewtchin could you guys take a look at this commit: https://github.com/vmware/vic/pull/3562/commits/e2bc8f90e80ec3e7bffabfeb61ebc647c1892a52

This is to fix the build break in build 7513.

emlin:
hmm... The change lgtm for me. Just not clear which case is fixed by this change. I think the container not found issue is caused by vm fixing, which cannot be fixed by this change.

hmahmood:
@emlin yes this change will not fix the problem with vm fixing, but for the the `docker rm` case it removes the calls that do lead to it. I will open an issue for the vm fixing problem.

PR #3565, opened 2017-01-04
Revert to cbednarski/ubuntu-1604 as Emma reported nimbus failures

caglar10ur:


dougm:
What failures?  How does the devbox Vagrantfile impact nimbus tests?

emlin:
@dougm it's not nimbus test failure. If I run deploy-esx.sh to nimbus esxi, the devbox provision.sh stuck there for grub-pc package upgrade.
```
The grub-pc package is being upgraded. This menu allows you to select which 
devices you'd like grub-install to be automatically run for, if any.

Running grub-install automatically is recommended in most situations, to prevent
the installed GRUB core image from getting out of sync with GRUB modules or 
grub.cfg.

If you're unsure which drive is designated as boot drive by your BIOS, it is 
often a good idea to install GRUB to all of them.

Note: it is possible to install GRUB to partition boot records as well, and some
appropriate partitions are offered here. However, this forces GRUB to use the 
blocklist mechanism, which makes it less reliable, and therefore is not 
recommended.

  1. /dev/sda5 (42435 MB; Virtual_disk)
  2. /dev/sda (42949 MB; Virtual_disk)
  3. /dev/dm-0 (41360 MB; vagrant--vg-root)
  4. /dev/sda1 (510 MB; Virtual_disk)

(Enter the items you want to select, separated by spaces.)

GRUB install devices: 
```
Even I manually input the device as required.

If use the old ubuntu box, there is no such issue, probably because old box does not have grub-pc package at all.

caglar10ur:
There are some other solutions @ https://github.com/mitchellh/vagrant/issues/289

caglar10ur:
merging this, will send another pr after some testing with bento box

Issue #3425, opened 2016-12-01
vic-admin can not grab file lock in /var/run/lock/logrotate_run.lock


vburenin:
vic-admin runs under account of a not a privileged user (vicadmin), so this user can't grab the lock on the file that is owned by root.

Users can see this error in log files when they try to export the log bundle in vicadmin:

time="2016-12-01T22:33:19Z" level=error msg="Failed to acquire logrotate lock: open /var/run/lock/logrotate_run.lock: permission denied"

sgairo:
related to #3251 , will close and keep original issue

anchal-agrawal:
Closing as dup of #3251

vburenin:
It is not a dup of #3251 
adding vicadmin into sudoers will not solve the problem.
We need a special group to run all out VIC related services so we can manage permissions for that group only.

PR #3566, opened 2017-01-04
Add vicadmin to vic group to be able to share same file between diffe…

vburenin:
- Added new group "vic"
- User vicadmin is added into vic group
- Set "vic" group permissions for  /var/run/lock/logrotate_run.lock to be able to share it between different processes.

Fixes: #3425

### DIFF
--- a/isos/appliance.sh
+++ b/isos/appliance.sh
@@ -98,12 +98,23 @@ rm -f $(rootfs_dir $PKGDIR)/etc/systemd/network/*
 
 chroot $(rootfs_dir $PKGDIR) groupadd -g 1000 vicadmin
 chroot $(rootfs_dir $PKGDIR) useradd -u 1000 -g 1000 -G systemd-journal -m -d /home/vicadmin -s /bin/false vicadmin
+
+# Group vic should be used to run all VIC related services.
+chroot $(rootfs_dir $PKGDIR) groupadd -g 1001 vic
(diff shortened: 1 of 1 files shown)


isos/appliance.sh
```diff
@@ -98,12 +98,23 @@ rm -f $(rootfs_dir $PKGDIR)/etc/systemd/network/*
...
+
 cp -R ${DIR}/vicadmin/* $(rootfs_dir $PKGDIR)/home/vicadmin
 chown -R 1000:1000 $(rootfs_dir $PKGDIR)/home/vicadmin
+
 # so vicadmin can read the system journal via journalctl
 install -m 755 -d $(rootfs_dir $PKGDIR)/etc/tmpfiles.d
 echo "m  /var/log/journal/%m/system.journal 2755 root systemd-journal - -" > $(rootfs_dir $PKGDIR)/etc/tmpfiles.d/systemd.conf
 
+chroot $(rootfs_dir $PKGDIR) mkdir -p /var/run/lock
+chroot $(rootfs_dir $PKGDIR) chmod 1777 /var/run/lock
```
fdawg4l: Minor and feel free to ignore, but you could combine the mkdir and the chmod with `install`.

`chroot $(rootfs_dir $PKGDIR) install -d -m 1777 var/run/lock`

PR #3575, opened 2017-01-05
[cosmetic] gofmt + misspell + vet

caglar10ur:
Also vendor misspell and start using them in check target


caglar10ur:
@dougm done

PR #3576, opened 2017-01-06
Switch to bento box - again

caglar10ur:
@emlin would be glad if you could test this change with your setup. It is working here for my local ESXi setup.


emlin:
testing

emlin:
Sorry for didn't get test result, though I ran it few times, but for the VPN is too slow, usually script failed after image upload.
I'll try it next Monday in the office, and let you know the result.

emlin:
The deploy scrip works well this time, and is much faster than old box. Nice job 👍

PR #3577, opened 2017-01-06
Add test to scrape logs for password leaks

### LINKED ISSUE #3563: Write a regression/integration test to scrape vic log files for sensitive info
We should write a test to check for passwords or sensitive credentials in VIC's log files. This would be a good addition at the end of the regression and vicadmin integration tests. We could use a custom password and look for it explicitly in the log files.

/cc @gigawhitlocks @sgairo

anchal-agrawal:
Adds a test to check log files for password leaks in the regression and vic-admin integration test suites.

Fixes #3563

### DIFF
--- a/tests/resources/VCH-Util.robot
+++ b/tests/resources/VCH-Util.robot
@@ -159,6 +159,22 @@ Check For The Proper Log Files
     Should Contain  ${output}  ${container}/vmware.log
     Should Contain  ${output}  ${container}/tether.debug
 
+Scrape Logs For the Password
+    [Tags]  secret
+    ${rc}=  Run And Return Rc  curl -sk %{VIC-ADMIN}/authentication -XPOST -F username=%{TEST_USERNAME} -F password=%{TEST_PASSWORD} -D /tmp/cookies-%{VCH-NAME}
+    Should Be Equal As Integers  ${rc}  0
+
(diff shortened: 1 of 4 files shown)


tests/resources/VCH-Util.robot
```diff
@@ -159,6 +159,22 @@ Check For The Proper Log Files
     Should Contain  ${output}  ${container}/vmware.log
     Should Contain  ${output}  ${container}/tether.debug
 
+Scrape Logs For the Password
```
gigawhitlocks: Can we add a Docker operation that actually does a thing against vSphere after login but before we do the scrape for the password so that we can make sure that the password doesn't leak out in other logs? There really shouldn't be a way for it to, because it shouldn't ever be passed to those components in the first place as they have a token instead, but this would allow us ongoing assurance that it does not.

anchal-agrawal: I've added this test at the end of the regression suite, so that'll take care of checking logs after a bunch of docker commands have been run. Should we still add a docker command inside `Scrape Logs For the Password`?

sgairo: I feel like that last file is sufficient

gigawhitlocks:
LGTM

PR #3579, opened 2017-01-06
vic-machine delete removes custom bridge network

### LINKED ISSUE #3193: delete VCH does not remove the bridge network if it was created with a custom name
VIC: v0.7.0-6507-6c4627d
vSphere 6.0 GA: ESX ob-2494585

Steps
1. Create VCH with --bridge-network 'vswitch' flag
> ./vic-machine-linux create --name VCHESX --target w1-vim-cat-24.eng.vmware.com --user root --password 'ca$hc0w' --no-tls --image-store w1-vimcat24-local --volume-store=w1-vimcat24-local/registry:default --insecure-registry 10.160.162.65:80 --thumbprint C9:6F:ED:44:54:48:D9:F6:B6:6A:3B:13:46:55:F0:45:15:F8:51:3B --bridge-network 'vswitch2'

2. Delete the VCH --name VCHESX  flag & with --force or without force 
Delete always searches the bridge-network name as VCH name and fails to delete.

This is a simple issue, but it leaves the dangling network in vSphere :-)

andrewtchin:
Use the correct name to delete the bridge network now, previously used only the VCH name which failed if a custom bridge network was specified on ESX.

Fixes #3193 

```
INFO[2017-01-06T09:46:14-08:00] ### Removing VCH ####                        
INFO[2017-01-06T09:46:15-08:00] Validating target                            
INFO[2017-01-06T09:46:17-08:00]                                              
INFO[2017-01-06T09:46:17-08:00] VCH ID: VirtualMachine:5                     
INFO[2017-01-06T09:46:17-08:00] Removing VMs                                 
INFO[2017-01-06T09:46:18-08:00] Removing image stores                        
INFO[2017-01-06T09:46:26-08:00] Removing volume stores                       
INFO[2017-01-06T09:46:26-08:00] Removing appliance VM network devices        
INFO[2017-01-06T09:46:28-08:00] Removing Portgroup "asdf1"                   
INFO[2017-01-06T09:46:28-08:00] Removing VirtualSwitch "asdf1"               
INFO[2017-01-06T09:46:33-08:00] Removing Resource Pool "oceanlab"            
INFO[2017-01-06T09:46:34-08:00] Completed successfully       
```

### DIFF
--- a/lib/install/management/network.go
+++ b/lib/install/management/network.go
@@ -120,6 +120,7 @@ func (d *Dispatcher) removeNetwork(conf *config.VirtualContainerHostConfigSpec) 
 
 	netw, ok := net.(*object.Network)
 	if !ok {
+		log.Errorf("Expected Network Type, got %#v", net)
 		return fmt.Errorf("Failed to get network for %q", moref)
 	}
 	pgName := netw.Name()
--- a/lib/install/management/network.go
+++ b/lib/install/management/network.go
@@ -119,26 +119,18 @@ func (d *Dispatcher) removeNetwork(conf *config.VirtualContainerHostConfigSpec) 
 	log.Debugf("Delete bridge network: %s", net)
 
 	netw, ok := net.(*object.Network)
-	log.Infof("Delete bridge network: %s", net)
-	log.Infof("Delete bridge network: %s", netw)
 	if !ok {
 		log.Errorf("Failed to find network %q: %s", moref, err)
 		return err
 	}
-
-	if network, err := d.session.Finder.Network(d.ctx, netw.InventoryPath); err != nil || network == nil {
-		log.Infof("Unable to delete network - failed to find %q", netw.InventoryPath)
-		log.Debugf("Failed to find network to delete: %s", err)
-		return nil
-	}
-
 	pgName := netw.Name()
-	log.Infof("Removing Portgroup %q", pgName)
+
 	hostNetSystem, err := d.session.Host.ConfigManager().NetworkSystem(d.ctx)
 	if err != nil {
 		return err
 	}
 
+	log.Infof("Removing Portgroup %q", pgName)
 	err = hostNetSystem.RemovePortGroup(d.ctx, pgName)
 	if err != nil {
 		return err
--- a/lib/install/management/network.go
+++ b/lib/install/management/network.go
@@ -101,18 +101,18 @@ func (d *Dispatcher) removeNetwork(conf *config.VirtualContainerHostConfigSpec) 
 
 	br := conf.ExecutorConfig.Networks["bridge"]
 	if br == nil {
-		return fmt.Errorf("Bridge Network ID unknown")
+		return fmt.Errorf("Bridge Network ID is unknown")
 	}
 	name := br.Network.ID
 	log.Debugf("Remove bridge network based on %s", name)
 
-	moref := new(types.ManagedObjectReference)
+	moref := types.ManagedObjectReference{}
 	ok := moref.FromString(name)
 	if !ok {
 		return fmt.Errorf("Unable to delete port group - failed to get moref from: %q", name)
 	}
 
-	net, err := d.session.Finder.ObjectReference(d.ctx, *moref)
+	net, err := d.session.Finder.ObjectReference(d.ctx, moref)
 	if err != nil {
 		return fmt.Errorf("Unable to delete port group - failed to find network from: %q", name)
 	}
--- a/lib/install/management/network.go
+++ b/lib/install/management/network.go
@@ -15,9 +15,12 @@
 package management
 
 import (
+	"fmt"
+
 	log "github.com/Sirupsen/logrus"
 
 	"github.com/vmware/govmomi/find"
+	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/pkg/errors"
@@ -80,6 +83,7 @@ func (d *Dispatcher) createBridgeNetwork(conf *config.VirtualContainerHostConfig
 	bnet.ID = net.Reference().String()
 	bnet.Network.ID = net.Reference().String()
 	conf.CreateBridgeNetwork = true
+	log.Debugf("Created portgroup %q: %s", name, net)
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


lib/install/management/network.go
```diff
@@ -95,26 +99,47 @@ func (d *Dispatcher) removeNetwork(conf *config.VirtualContainerHostConfigSpec)
...
+	}
+	log.Debugf("Delete bridge network: %s", net)
+
+	netw, ok := net.(*object.Network)
+	if !ok {
+		log.Errorf("Failed to find network %q: %s", moref, err)
+		return err
+	}
+
+	if network, err := d.session.Finder.Network(d.ctx, netw.InventoryPath); err != nil || network == nil {
```
emlin: Is this necessary? I think the previous step already searched the object from host.

emlin:
overall lgtm.

lib/install/management/network.go
```diff
@@ -95,26 +99,47 @@ func (d *Dispatcher) removeNetwork(conf *config.VirtualContainerHostConfigSpec)
 		return nil
 	}
 
-	name := conf.Name
-	if network, err := d.session.Finder.Network(d.ctx, name); err != nil || network == nil {
-		log.Infof("Didn't find network %q", name)
-		log.Debugf("Didn't find network for %s", err)
+	log.Debugf("Remove bridge network based on %s", conf.ExecutorConfig.Networks["bridge"].Network.ID)
```
vburenin: is there any chance that there is no "bridge" key in the map?
You also extracted "name" in the next block, you could use that name in your log.debug call.

lib/install/management/network.go
```diff
@@ -95,26 +99,47 @@ func (d *Dispatcher) removeNetwork(conf *config.VirtualContainerHostConfigSpec)
...
+
+	name := conf.ExecutorConfig.Networks["bridge"].Network.ID
+	moref := new(types.ManagedObjectReference)
```
vburenin: I would move it below if, since there is a chance of error and creating an additional object in this case doesn't make sense. It doesn't really affect anything, but I would like all code to do only what's necessary at the specific state.

vburenin: Why did you create a pointer to an object instead of just object? Looking at usage you just dereference it creating a copy.

lib/install/management/network.go
vburenin: can network even become a nil here if there is no error?

lib/install/management/network.go
```diff
@@ -95,26 +99,47 @@ func (d *Dispatcher) removeNetwork(conf *config.VirtualContainerHostConfigSpec)
...
+	log.Debugf("Delete bridge network: %s", net)
+
+	netw, ok := net.(*object.Network)
+	if !ok {
+		log.Errorf("Failed to find network %q: %s", moref, err)
+		return err
+	}
+
+	if network, err := d.session.Finder.Network(d.ctx, netw.InventoryPath); err != nil || network == nil {
+		log.Infof("Unable to delete network - failed to find %q", netw.InventoryPath)
```
vburenin: I guess it is a warning or error.

lib/install/management/network.go
```diff
@@ -95,26 +99,47 @@ func (d *Dispatcher) removeNetwork(conf *config.VirtualContainerHostConfigSpec)
...
+		log.Debugf("Failed to find network to delete: %s", err)
```
vburenin: Is it really a debug? I am just asking to have more context.

lib/install/management/network.go
```diff
@@ -95,26 +99,47 @@ func (d *Dispatcher) removeNetwork(conf *config.VirtualContainerHostConfigSpec)
...
 		return nil
 	}
 
-	log.Infof("Removing Portgroup %q", name)
+	pgName := netw.Name()
+	log.Infof("Removing Portgroup %q", pgName)
```
vburenin: Why %q if pgName is a string?

lib/install/management/network.go
```diff
@@ -95,26 +99,47 @@ func (d *Dispatcher) removeNetwork(conf *config.VirtualContainerHostConfigSpec)
...
 
-	err = hostNetSystem.RemovePortGroup(d.ctx, name)
+	err = hostNetSystem.RemovePortGroup(d.ctx, pgName)
 	if err != nil {
 		return err
 	}
 
-	log.Infof("Removing VirtualSwitch %q", name)
-	err = hostNetSystem.RemoveVirtualSwitch(d.ctx, name)
+	log.Infof("Removing VirtualSwitch %q", pgName)
```
vburenin: Is there any chance that VirtualSwitch exists, but portgroup does not?

lib/install/management/network.go
andrewtchin: bridge is always expected to be present

vburenin: Is always expected, doesn't meant it always will be there, that is why I am asking.
Trying to access Network object like this looks a little spooky to me, since you could get a Nil object.

lib/install/management/network.go
andrewtchin: this appears to be the pattern to use finder https://github.com/vmware/vic/blob/master/lib/install/management/appliance.go#L222

lib/install/management/network.go
andrewtchin: we've used %q in a lot of places in vic-machine where we refer to a name so that the name is quoted

lib/install/management/network.go
andrewtchin: good call, removed

lib/install/management/network.go
andrewtchin: fixed

lib/install/management/network.go
andrewtchin: if the user manually deleted the portgroup and left the vswitch

lib/install/management/network.go
andrewtchin: removed

lib/install/management/network.go
andrewtchin: removed

lib/install/management/network.go
```diff
@@ -95,26 +99,44 @@ func (d *Dispatcher) removeNetwork(conf *config.VirtualContainerHostConfigSpec)
...
+	}
+
+	net, err := d.session.Finder.ObjectReference(d.ctx, *moref)
+	if err != nil {
+		return fmt.Errorf("Unable to delete port group - failed to find network from: %q", name)
+	}
+	log.Debugf("Delete bridge network: %s", net)
+
+	netw, ok := net.(*object.Network)
+	if !ok {
```
vburenin: if the asserted type is not matching, isn't that really bad? I suspect that error message should be a little more alerting

lib/install/management/network.go
vburenin: So, you don't want to exit here. You should just log that you could not delete something and proceed to the next item.

andrewtchin: we wouldn't have the pgName there to remove the vswitch because Finder wouldn't have found the network

lib/install/management/network.go
vburenin: that is not always a good pattern :)

PR #3587, opened 2017-01-10
Update vmw-guestinfo [full ci]

fdawg4l:
Update `vmw-guestinfo` vendor dependency.

This updates vmw-guestinfo to the pure go implementation of the backdoor protocol.


fdawg4l:
Alarmingly, the integration tests fail with this change.  I either broke something in `vmw-guestinfo` or we need to modify something in our code.  I'm betting the former.  Will take a look ASAP and get a handle on where things are broken.  cc @sigma.

fdawg4l:
Uh oh...  

```
 cat tether.debug
all stderr redirected to debug logall stdout redirected to debug logtime="2017-01-11T15:46:06Z" level=info msg="Registering tether extension Toolbox"
time="2017-01-11T15:46:06Z" level=info msg="Registering tether extension logrotate"
time="2017-01-11T15:46:06Z" level=info msg="opening /var/log/vic/init.log for debug log"
time="2017-01-11T15:46:06Z" level=info msg="Started reaping child processes"
time="2017-01-11T15:46:06Z" level=info msg="Launching vch-init pprof server on 127.0.0.1:6060"
time="2017-01-11T15:46:06Z" level=info msg="Starting extension Toolbox"
time="2017-01-11T15:46:06Z" level=info msg="toolbox: waiting for initialization"
panic: runtime error: index out of range

goroutine 10 [running]:
panic(0xb1f480, 0xc42000c090)
        /usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/vmware/vic/vendor/github.com/vmware/vmw-guestinfo/message.(*Channel).Send(0xc420242600, 0x0, 0x0, 0x0, 0x0, 0x0)
        /go/src/github.com/vmware/vic/vendor/github.com/vmware/vmw-guestinfo/message/message.go:140 +0x3c9
github.com/vmware/vic/pkg/vsphere/toolbox.(*Service).Start.func1(0xc42004ea20)
        /go/src/github.com/vmware/vic/pkg/vsphere/toolbox/service.go:129 +0x14c
created by github.com/vmware/vic/pkg/vsphere/toolbox.(*Service).Start
[root@office1-sfo2-dhcp326:/vmfs/volumes/573cd87e-131cbc3c-b93d-90b11c76903f/virtual-container-host]
```

Smells like an off by one error.  I'll take care of this upstream ASAP.

fdawg4l:
I think I have a handle on this.  Passing an empty buffer to `Send` is apparently supported.  I'll add support for this.

fdawg4l:
I left out half of the protocol to handle errors.  This broke the toolbox pretty hard.  I have a fix going in upstream and I'll rebase on top of that and try this again.  Marking this as WIP until then.

fdawg4l:
Fixes #3823

fdawg4l:
I'm still trying to get a clean CI run with this change.  It fails somewhat randomly in unrelated paths, but MASTER didn't exhibit this behavior.  Will drill down and pull people into this change so we can have confidence when we merge this.  Thanks for your patience.

fdawg4l:
@dougm this is ready to go!

pkg/vsphere/toolbox/backdoor_test.go
```diff
@@ -30,13 +30,8 @@ func TestBackdoorChannel(t *testing.T) {
 	}
 
 	for _, f := range funcs {
-		err := f()
-
-		if err != nil {
-			if err == ErrNotVirtualWorld {
```
dougm: Did this check not work anymore?  Only wanted to skip the test on this error.  If we can't do that check, adding err.Error() as a param to SkipNow would be good.

fdawg4l: The check will fail if the process isn't run as root so the error returned can be that or the not a virtual world error. I tried getting the tests to run as root in drone but there's no way that I could find to give our container CAP_SYS_RAWIO so the call would actually hit the back door without falling over outside of a vm. 

I'll add a parameter to pass err.

Issue #2248, opened 2016-09-06
Implement images filtering


matthewavery:
Implement the filter functionality for the `docker images` call. This will provide filtering functionality which is especially helpful when managing a large image store.

cgtexmex:
I'm hoping that this and filtering of containers #1676 can share some common filtering code, so while not related from a functional perspective they will be related via implementation.

mhagen-vmware:
lgtm

Issue #1035, opened 2016-06-09
docker images alpine:3.1 does not return as expected


mhagen-vmware:
Both filtering by repo and tag seem to be ignored currently:

```
docker -H 192.168.254.215:2375 images alpine:3.2
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
alpine              latest              f70c828098f5        6 days ago          4.795 MB
alpine              3.2                 9aae2f0938df        6 days ago          5.252 MB
alpine              3.1                 6ee9065c0936        6 days ago          5.038 MB
ubuntu              latest              2fa927b5cdd3        13 days ago         122 MB
busybox             latest              47bcc53f74dc        11 weeks ago        1.113 MB
```

```
docker -H 192.168.254.215:2375 images alpine
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
alpine              latest              f70c828098f5        6 days ago          4.795 MB
alpine              3.2                 9aae2f0938df        6 days ago          5.252 MB
alpine              3.1                 6ee9065c0936        6 days ago          5.038 MB
ubuntu              latest              2fa927b5cdd3        13 days ago         122 MB
busybox             latest              47bcc53f74dc        11 weeks ago        1.113 MB
```

hickeng:
@cgtexmex @sflxn @jzt is this something that can easily be addressed? I know some filtering occurs in the client, but no idea which pieces if any for images.
Leaving in new images until better understanding of implementation requirements.

corrieb:
NOTE: CLOSING THIS ISSUE MAY CAUSE AN INTEGRATION TEST FAILURE. Please re-enable the test in 1-3-Docker-Images.robot as part of a PR to fix this issue. Please also ensure that the test passes cleanly. If the test needs improving or updating, please address it. Thx

jzt:
Closing this, captured in #2248.

PR #3661, opened 2017-01-18
Fix issue number in docker images tests

jzt:
An issue was added (#2248) to capture image filtering that encompasses #1035, so #1035 was closed. There is an integration test that depends on this issue, so I changed it to reference #2248 instead of #1035.


mhagen-vmware:
lgtm

PR #3664, opened 2017-01-18
Fixing container lookup when there is partial network overlap between requester and requested containers  [full ci]

### LINKED ISSUE #2721: Two networks, with a bridge container does not work
1. Create a container on network 1
2. Create a container on network 2
3. Create a container on both networks
4. From the container on step 3, we should be able to ping 1 and 2.  With docker we can, with VIC we can ping the first container but the second container returns:
   ping: bad address 'busy2'

[container-logs.zip](https://github.com/vmware/vic/files/527850/container-logs.zip)

hmahmood:
Fixes #2721

### DIFF
--- a/lib/dns/dns.go
+++ b/lib/dns/dns.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -320,14 +320,58 @@ Redo:
 	return true, nil
 }
 
+// lookupByAlias looks up a container by alias, given the requesting container
+// and the network of the incoming request. It first does a container-scoped
+// alias search, followed by a network-scoped aliased search for alias. The
+// result is a collection of endpoints for the matching container that are
+// only on the same network as the requesting container.
+func lookupByAlias(netCtx *network.Context, scope *network.Scope, reqc *network.Container, alias string) []*network.Endpoint {
+	// container specific alias search
+	cons := netCtx.ContainersByAlias(network.ScopedAliasName(scope.Name(), reqc.Name(), alias))
+	if len(cons) == 0 {
+		// scope-wide alias search
+		cons = netCtx.ContainersByAlias(network.ScopedAliasName(scope.Name(), "", alias))
+	}
+
+	if len(cons) == 0 {
+		return nil
+	}
+
+	var eps []*network.Endpoint
+	for _, c := range cons {
+		if e := c.Endpoint(scope); e != nil {
+			eps = append(eps, e)
+		}
+	}
+
+	return eps
+}
+
+// lookupByName looks up a container by name given the requesting container. It returns a collection
+// of endpoints on networks that both the requesting and matching container share.
+func lookupByName(netCtx *network.Context, reqc *network.Container, name string) []*network.Endpoint {
+	var eps []*network.Endpoint
+	if m := netCtx.Container(name); m != nil {
+		// look if networks overlap for the requesting container and the
+		// matched container
+		for _, ec := range reqc.Endpoints() {
+			if em := m.Endpoint(ec.Scope()); em != nil {
+				eps = append(eps, em)
+			}
+		}
+	}
+
+	return eps
+}
+
 // HandleVIC returns a response to a container name/id request
 func (s *Server) HandleVIC(w mdns.ResponseWriter, r *mdns.Msg) (bool, error) {
 	defer trace.End(trace.Begin(r.String()))
 
 	question := r.Question[0]
 
-	ctx := network.DefaultContext
-	if ctx == nil {
+	netCtx := network.DefaultContext
+	if netCtx == nil {
 		log.Errorf("DefaultContext is not initialized")
 		return false, fmt.Errorf("DefaultContext is not initialized")
 	}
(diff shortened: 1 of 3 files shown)


lib/dns/dns.go
```diff
@@ -320,6 +320,48 @@ Redo:
...
+// and the network of the incoming request. It first does a container-scoped
+// alias search, followed by a network-scoped aliased search for alias. The
+// result is a collection of endpoints for the matching container that are
+// only on the same network as the requesting container.
+func lookupByAlias(ctx *network.Context, scope *network.Scope, reqc *network.Container, alias string) []*network.Endpoint {
+	// container specific alias search
+	cons := ctx.ContainersByAlias(fmt.Sprintf("%s:%s:%s", scope.Name(), reqc.Name(), alias))
+	if len(cons) == 0 {
+		// scope-wide alias search
+		cons = ctx.ContainersByAlias(fmt.Sprintf("%s:%s", scope.Name(), alias))
```
vburenin: No action needed: frequently str + ":" + str can be faster, even including GC work due to complexity of Sprintf.

lib/dns/dns.go
```diff
@@ -356,34 +398,31 @@ func (s *Server) HandleVIC(w mdns.ResponseWriter, r *mdns.Msg) (bool, error) {
 	if e == nil {
 		return false, fmt.Errorf("Could not find requesting container with ip %s", ip)
 	}
-	scope := e.Scope()
 
-	if domain != "" && scope.Name() != domain {
-		return false, fmt.Errorf("Inter-scope request for container %s in %s from %s", name, domain, scope.Name())
+	if domain != "" && e.Scope().Name() != domain {
```
vburenin: can e.Scope().Name() be an empty string?

hmahmood: Nope; we check that when we create the scope.

lib/dns/dns.go
```diff
@@ -320,6 +320,48 @@ Redo:
...
 }
 
+// lookupByAlias looks up a container by alias, given the requesting container
+// and the network of the incoming request. It first does a container-scoped
+// alias search, followed by a network-scoped aliased search for alias. The
+// result is a collection of endpoints for the matching container that are
+// only on the same network as the requesting container.
+func lookupByAlias(ctx *network.Context, scope *network.Scope, reqc *network.Container, alias string) []*network.Endpoint {
+	// container specific alias search
+	cons := ctx.ContainersByAlias(fmt.Sprintf("%s:%s:%s", scope.Name(), reqc.Name(), alias))
```
fdawg4l: This string looks to follow a well defined format (`<scopeName>:<containerName>:<string alias>`).  Can this format be hidden behind a helper that returns the result you want but doesnt require `Sprintf`?

lib/dns/dns.go
```diff
@@ -320,6 +320,48 @@ Redo:
 	return true, nil
 }
 
+// lookupByAlias looks up a container by alias, given the requesting container
+// and the network of the incoming request. It first does a container-scoped
+// alias search, followed by a network-scoped aliased search for alias. The
+// result is a collection of endpoints for the matching container that are
+// only on the same network as the requesting container.
+func lookupByAlias(ctx *network.Context, scope *network.Scope, reqc *network.Container, alias string) []*network.Endpoint {
```
fdawg4l: `ctx` name collides with `context.Context`.  How about `networkCtx` or really *anything* else?

lib/dns/dns.go
```diff
@@ -356,34 +398,31 @@ func (s *Server) HandleVIC(w mdns.ResponseWriter, r *mdns.Msg) (bool, error) {
...
 	// FIXME: Add AAAA when we support it
-	answer := make([]mdns.RR, len(cons))
-	// shuffle cons
-	for i := 0; i < len(cons); i++ {
-		j := random.Intn(len(cons))
-		cons[i], cons[j] = cons[j], cons[i]
+	answer := make([]mdns.RR, len(eps))
+	// shuffle eps
+	for i := 0; i < len(eps); i++ {
+		j := random.Intn(len(eps))
```
vburenin: this shuffle needs to be fixed. I and Hasan talked in chat. Here is a link Hasan found that explains why https://blog.codinghorror.com/the-danger-of-naivete/

lib/dns/dns.go
```diff
@@ -320,6 +320,48 @@ Redo:
...
+		cons = ctx.ContainersByAlias(fmt.Sprintf("%s:%s", scope.Name(), alias))
+	}
+
+	if len(cons) == 0 {
+		return nil
+	}
+
+	eps := make([]*network.Endpoint, len(cons))
+	for i, c := range cons {
+		eps[i] = c.Endpoint(scope)
```
fdawg4l: Will this ever return `nil`?  I'm wondering if a lock is held somewhere so the Endpoint for the scope isn't mutated somewhere else while in this iteration.

lib/dns/dns.go
vburenin: in this case checking for "" doesn't make sense.

lib/dns/dns.go
```diff
@@ -320,6 +320,48 @@ Redo:
...
+// lookupByName looks up a container by name given the requesting container. It returns a collection
+// of endpoints on networks that both the requesting and matching container share.
+func lookupByName(ctx *network.Context, reqc *network.Container, name string) []*network.Endpoint {
+	var eps []*network.Endpoint
+	if m := ctx.Container(name); m != nil {
+		// look if networks overlap for the requesting container and the
+		// matched container
+		for _, ec := range reqc.Endpoints() {
+			if em := m.Endpoint(ec.Scope()); em != nil {
+				eps = append(eps, em)
```
fdawg4l: On line `340` you `make()` the slice up front.  Here you `append`.  Why the different idioms?

lib/dns/dns.go
hmahmood: Good point; I'll fix that.

lib/dns/dns.go
hmahmood: There we know the number of endpoints up front, here we don't. But that code has that race condition, so it will end up looking the same as here with regards to `eps` size.

lib/dns/dns.go
hmahmood: `domain` can be optional, so we need to skip this check if it is not provided in the query.

lib/dns/dns.go
fdawg4l: Ah, right!  Somehow completely missed the `if`.  My bad.  Thanks!

lib/dns/dns.go
vburenin: netCtx

lib/dns/dns.go
hmahmood: I will export a helper function from the `network` package.

PR #3667, opened 2017-01-19
Exclude empty layer diffIDs from rootFS

### LINKED ISSUE #3602: VIC image IDs not consistent with docker image IDs
Currently, image IDs calculated by VIC do not match those calculated by docker. This is caused by VIC appending the empty layer diffID to the rootFS section of the image config JSON instead of (like docker) excluding it.

Since the image ID is effectively a sha256 of the image content data, the IDs should be consistent so that a user will know whether or not an image contains the correct bits.

jzt:
This changes how we construct image metadata to bring it in line with docker, resulting in image IDs that are consistent between VIC and docker. Fixes #3602


mhagen-vmware:
lgtm

PR #3684, opened 2017-01-20
[specific ci=1-22-Docker-Volume-RM] Move volume in use check to volume cache

fdawg4l:
Move the check into the cache so it can be used by other volume store
backends.  The same check will be needed despite the backend so moving
it up a layer is the answer.

[specific ci=1-22-Docker-Volume-RM]


matthewavery:
lgtm

PR #3691, opened 2017-01-21
[specific ci=1-19-Docker-Volume-Create] Move volumestore list into volume cache

fdawg4l:
We were keeping all of the volumestores in the implementation.  That
meant every new volumestore would go down into the vsphere
implementation and provision itself there and keep state private in the
impl.  This moves the volume store up into the volume cache so each new
volume store is an implementation itself.  This will allow us to support
volume stores beyond just vsphere (like nfs).

[specific ci=1-19-Docker-Volume-Create]

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
@@ -29,7 +29,7 @@ import (
 	"github.com/vmware/vic/lib/apiservers/portlayer/restapi/operations/storage"
 	epl "github.com/vmware/vic/lib/portlayer/exec"
 	spl "github.com/vmware/vic/lib/portlayer/storage"
-	vsphereSpl "github.com/vmware/vic/lib/portlayer/storage/vsphere"
+	"github.com/vmware/vic/lib/portlayer/storage/vsphere"
 	"github.com/vmware/vic/lib/portlayer/util"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/datastore"
@@ -58,7 +58,7 @@ func (h *StorageHandlersImpl) Configure(api *operations.PortLayerAPI, handlerCtx
 		log.Warningf("Multiple image stores found. Multiple image stores are not yet supported. Using [%s] %s", imageStoreURL.Host, imageStoreURL.Path)
 	}
 
-	ds, err := vsphereSpl.NewImageStore(op, handlerCtx.Session, &imageStoreURL)
+	ds, err := vsphere.NewImageStore(op, handlerCtx.Session, &imageStoreURL)
 	if err != nil {
 		log.Panicf("Cannot instantiate storage layer: %s", err)
 	}
@@ -68,13 +68,6 @@ func (h *StorageHandlersImpl) Configure(api *operations.PortLayerAPI, handlerCtx
 	// expensive metadata lookups.
 	h.imageCache = spl.NewLookupCache(ds)
 
-	// The same is done for volumes.  It's implemented via a cache which writes
-	// to an implementation that takes a datastore to write to.
-	vsVolumeStore, err := vsphereSpl.NewVolumeStore(op, handlerCtx.Session)
-	if err != nil {
-		log.Panicf("Cannot instantiate the volume store: %s", err)
-	}
-
 	// Get the datastores for volumes.
 	// Each volume store name maps to a datastore + path, which can be referred to by the name.
 	dstores, err := datastore.GetDatastores(op, handlerCtx.Session, spl.Config.VolumeLocations)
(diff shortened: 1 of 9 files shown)


lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -302,13 +297,7 @@ func (h *StorageHandlersImpl) VolumeStoresList(params storage.VolumeStoresListPa
...
 	}
 
-	resp := &models.VolumeStoresListResponse{
-		Stores: make(map[string]string),
-	}
-
-	for name, ds := range stores {
-		resp.Stores[name] = ds.String()
-	}
+	resp := &models.VolumeStoresListResponse{Stores: stores}
```
matthewavery: Is the default behavior here that the respond will call the .String() function on the values here? since stores is of type map[string]Url

fdawg4l: I changed returned model to be an array of strings.

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -68,32 +68,27 @@ func (h *StorageHandlersImpl) Configure(api *operations.PortLayerAPI, handlerCtx
...
+
 	// Add datastores to the vsphere volume store impl
 	for volStoreName, volDatastore := range dstores {
 		log.Infof("Adding volume store %s (%s)", volStoreName, volDatastore.RootURL)
-		_, err := vsVolumeStore.AddStore(op, volDatastore, volStoreName)
+
+		vs, err := vsphere.NewVolumeStore(op, volStoreName, handlerCtx.Session, volDatastore)
 		if err != nil {
-			log.Errorf("volume addition error %s", err)
+			log.Panicf("Cannot instantiate the volume store: %s", err)
```
vburenin: Are you sure you want to panic here right away? No retries, etc?

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
matthewavery: Nevermind, I see you changed the cache to returning an array. Ignore this, commented before I made it all the way through the PR.

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
fdawg4l: If we were to retry, it would be in the called function (`NewVolumeStore`).  

That said, I agree, `Panic`ing isn't the right thing to do here.  Will lower the severity to `Errorf`.

lib/portlayer/storage/volume_cache.go
```diff
@@ -35,22 +36,68 @@ type VolumeLookupCache struct {
 	vlc     map[string]Volume
 	vlcLock sync.RWMutex
 
-	// The underlying data storage implementation
-	volumeStore VolumeStorer
+	// Maps the service url of the volume store to the underlying data storage implementation
+	volumeStores map[url.URL]VolumeStorer
```
vburenin: url.URL has a pointer, are you sure it is a reasonable map key? Have you considered using string representation?

lib/portlayer/storage/volume_cache.go
```diff
@@ -35,22 +36,68 @@ type VolumeLookupCache struct {
...
 }
 
-func NewVolumeLookupCache(op trace.Operation, vs VolumeStorer) (*VolumeLookupCache, error) {
+func NewVolumeLookupCache(op trace.Operation) *VolumeLookupCache {
```
vburenin: Do you still need "op" here for constructor?

lib/portlayer/storage/volume_cache.go
```diff
@@ -35,22 +36,68 @@ type VolumeLookupCache struct {
...
 
-	return v, v.rebuildCache(op)
+	return v
 }
 
-// List the configured volume stores
-func (v *VolumeLookupCache) VolumeStoresList(op trace.Operation) (map[string]url.URL, error) {
-	return v.volumeStore.VolumeStoresList(op)
+// AddStore adds a volumestore by name.  The url returned is the service url to the volume store.
+func (v *VolumeLookupCache) AddStore(op trace.Operation, storeName string, vs VolumeStorer) (*url.URL, error) {
```
vburenin: We should rename trace.Operation to something like trace.Op or t.Op, it just consumes valuable space being as frequently used type.

lib/portlayer/storage/volume_cache.go
```diff
@@ -35,22 +36,68 @@ type VolumeLookupCache struct {
...
+	v.volumeStores[*u] = vs
+	return u, v.rebuildCache(op)
+}
+
+func (v *VolumeLookupCache) volumeStore(store *url.URL) (VolumeStorer, error) {
+
+	// find the datastore
+	vs, ok := v.volumeStores[*store]
+	if !ok {
+		return nil, VolumeStoreNotFoundError{Msg: fmt.Sprintf("volume store (%s) not found", store.String())}
```
vburenin: please make it multiline.

lib/portlayer/storage/volume_cache.go
```diff
@@ -35,22 +36,68 @@ type VolumeLookupCache struct {
...
+	return vs, nil
+}
+
+// VolumeStoresList returns a list of volume store names
+func (v *VolumeLookupCache) VolumeStoresList(op trace.Operation) ([]string, error) {
+
+	v.vlcLock.RLock()
+	defer v.vlcLock.RUnlock()
+
+	stores := make([]string, len(v.volumeStores))
```
vburenin: ouch! you create an array of empty strings and then successfully just append new elements. Is it intended?
This is the make api ```make([]T, len, cap)```
I think you wanted to do something like this: ```make([]string, 0, len(v.volumeStores))```

fdawg4l: DOH!  Yeah, good call.  WIll fix.  Thanks!

lib/portlayer/storage/volume_cache.go
fdawg4l: https://golang.org/pkg/net/url/#URL

I guess you mean UserInfo.  Good call.  I'll change this to a string.  I'm using that idiom all over the place so I'll push a new PR for the other places too after this one.

lib/apiservers/portlayer/swagger.json
```diff
@@ -2654,8 +2654,8 @@
 			],
 			"properties": {
 				"Stores": {
-					"type": "object",
-					"additionalProperties": {
+					"type": "array",
```
matthewavery: Some changes on the request side will also be needed for this change. I took a look and the code should still work fine, but the `VolumeStoreMap` var needs to be renamed. right here is where it might need a quick look: https://github.com/vmware/vic/blob/54921dd5b4ca95587db54792cdac8c046287bce7/lib/apiservers/engine/backends/system.go#L327-L341

lib/portlayer/storage/volume_cache.go
matthewavery: @vburenin I think this should be made a separate ticket if we really want to do this. We have a lot of tech debt discussion surrounding the changes to logging in general. I think that it is better to service that request along those lines rather than in this PR.

lib/apiservers/portlayer/swagger.json
fdawg4l: Var rename sounds good.  Will do.  The implicit type'ing going on in that function is bugging me.  This *should* have broken during build but because of the use of `:=`, it didn't.

lib/portlayer/storage/volume_cache.go
vburenin: @matthewavery Yah, definitely not here.

PR #3701, opened 2017-01-23
Add new log formatter [full ci]

### LINKED ISSUE #3615: Implement custom log formatter
Acceptance:
* log formatter should conform to the logurs [Formatter](https://github.com/sirupsen/logrus/blob/master/formatter.go) interface
* all uses of logrus in VIC engine use the new formatter

hmahmood:
Fixes #3615 

Example output:
```
Jan 23 2017 21:43:18.799Z INFO Started reaping child processes
Jan 23 2017 21:43:18.802Z DEBU writing "::1 ipv6-localhost ip6-loopback ipv6-loopback ip6-localhost\n" to /etc/hosts
Jan 23 2017 21:43:18.806Z DEBU writing "127.0.0.1 localhost.localdomain localhost\n" to /etc/hosts
Jan 23 2017 21:43:18.810Z DEBU writing "ff00:: ip6-mcastprefix\n" to /etc/hosts
Jan 23 2017 21:43:18.813Z DEBU writing "ff02::1 ip6-allnodes\n" to /etc/hosts
Jan 23 2017 21:43:18.816Z DEBU writing "fe00:: ip6-localnet\n" to /etc/hosts
Jan 23 2017 21:43:18.819Z DEBU writing "ff02::2 ip6-allrouters\n" to /etc/hosts
Jan 23 2017 21:43:18.822Z WARN Unable to resolve 'client.localhost': %!(EXTRA *net.DNSError=lookup client.localhost on [2001:4860:4860::8888]:53: dial udp [2001:4860:4860::8888]:53: connect: network is unreachable)
Jan 23 2017 21:43:18.830Z INFO Launching vch-init pprof server on 127.0.0.1:6060
Jan 23 2017 21:43:18.833Z INFO Starting extension logrotate
Jan 23 2017 21:43:18.835Z DEBU op=179.1 (delta:61.792074ms): logrotate config: /var/log/vic/port-layer.log {
Jan 23 2017 21:43:18.835Z DEBU     compress
Jan 23 2017 21:43:18.835Z DEBU     daily
Jan 23 2017 21:43:18.835Z DEBU     rotate 2
Jan 23 2017 21:43:18.835Z DEBU     size 20971520
Jan 23 2017 21:43:18.835Z DEBU     minsize 20971519
Jan 23 2017 21:43:18.835Z DEBU     copytruncate
Jan 23 2017 21:43:18.835Z DEBU     dateext
Jan 23 2017 21:43:18.835Z DEBU     dateformat -%Y%m%d-%s
Jan 23 2017 21:43:18.835Z DEBU }
```

### DIFF
--- a/cmd/docker/main.go
+++ b/cmd/docker/main.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cmd/docker/main.go
+++ b/cmd/docker/main.go
@@ -36,6 +36,7 @@ import (
 	vicbackends "github.com/vmware/vic/lib/apiservers/engine/backends"
 	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/lib/pprof"
+	viclog "github.com/vmware/vic/pkg/log"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/extraconfig"
 )
@@ -54,6 +55,7 @@ const (
 var vchConfig config.VirtualContainerHostConfigSpec
 
 func init() {
+	log.SetFormatter(viclog.NewTextFormatter())
 	trace.Logger.Level = log.DebugLevel
 	pprof.StartPprof("docker personality", pprof.DockerPort)
 }
--- a/cmd/port-layer-server/main.go
+++ b/cmd/port-layer-server/main.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cmd/port-layer-server/main.go
+++ b/cmd/port-layer-server/main.go
@@ -15,11 +15,11 @@
 package main
 
 import (
-	"log"
 	"os"
 	"os/signal"
 	"syscall"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/go-openapi/loads"
 	"github.com/jessevdk/go-flags"
 
@@ -27,6 +27,7 @@ import (
 	"github.com/vmware/vic/lib/apiservers/portlayer/restapi/operations"
 	"github.com/vmware/vic/lib/dns"
 	"github.com/vmware/vic/lib/pprof"
+	viclog "github.com/vmware/vic/pkg/log"
 )
 
 var (
@@ -34,6 +35,8 @@ var (
 )
 
 func init() {
+	log.SetFormatter(viclog.NewTextFormatter())
+
 	pprof.StartPprof("portlayer server", pprof.PortlayerPort)
 }
 
--- a/cmd/tether/main_darwin.go
+++ b/cmd/tether/main_darwin.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cmd/tether/main_darwin.go
+++ b/cmd/tether/main_darwin.go
@@ -24,6 +24,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 
 	"github.com/vmware/vic/lib/tether"
+	viclog "github.com/vmware/vic/pkg/log"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/extraconfig"
 )
@@ -50,7 +51,7 @@ func main() {
 	log.SetLevel(log.DebugLevel)
 
 	// Initiliaze logger with default TextFormatter
-	log.SetFormatter(&log.TextFormatter{DisableColors: true, FullTimestamp: true})
+	log.SetFormatter(viclog.NewTextFormatter())
 
 	// TODO: hard code executor initialization status reporting via guestinfo here
 	err := createDevices()
--- a/cmd/tether/main_linux.go
+++ b/cmd/tether/main_linux.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cmd/tether/main_linux.go
+++ b/cmd/tether/main_linux.go
@@ -24,13 +24,28 @@ import (
 	log "github.com/Sirupsen/logrus"
 
 	"github.com/vmware/vic/lib/tether"
+	viclog "github.com/vmware/vic/pkg/log"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/extraconfig"
 )
 
 var tthr tether.Tether
 
+func init() {
+	// Initiliaze logger with default TextFormatter
+	log.SetFormatter(viclog.NewTextFormatter())
+
+	// use the same logger for trace and other logging
+	trace.Logger = log.StandardLogger()
+	log.SetLevel(log.DebugLevel)
+}
+
 func main() {
+	if strings.HasSuffix(os.Args[0], "-debug") {
+		extraconfig.DecodeLogLevel = log.DebugLevel
+		extraconfig.EncodeLogLevel = log.DebugLevel
+	}
+
 	defer func() {
 		if r := recover(); r != nil {
 			log.Errorf("run time panic: %s : %s", r, debug.Stack())
@@ -53,17 +68,6 @@ func main() {
 
 	// where to look for the various devices and files related to tether
 	pathPrefix = "/.tether"
-
-	if strings.HasSuffix(os.Args[0], "-debug") {
-		extraconfig.DecodeLogLevel = log.DebugLevel
-		extraconfig.EncodeLogLevel = log.DebugLevel
-	}
-	// use the same logger for trace and other logging
-	trace.Logger = log.StandardLogger()
-	log.SetLevel(log.DebugLevel)
-
-	// Initiliaze logger with default TextFormatter
-	log.SetFormatter(&log.TextFormatter{DisableColors: true, FullTimestamp: true})
 
 	// TODO: hard code executor initialization status reporting via guestinfo here
 	err = createDevices()
--- a/cmd/tether/main_windows.go
+++ b/cmd/tether/main_windows.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cmd/tether/main_windows.go
+++ b/cmd/tether/main_windows.go
@@ -22,6 +22,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 
 	"github.com/vmware/vic/lib/tether"
+	viclog "github.com/vmware/vic/pkg/log"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/extraconfig"
 )
@@ -48,7 +49,7 @@ func main() {
 	log.SetLevel(log.DebugLevel)
 
 	// Initiliaze logger with default TextFormatter
-	log.SetFormatter(&log.TextFormatter{DisableColors: true, FullTimestamp: true})
+	log.SetFormatter(viclog.NewTextFormatter())
 
 	// get the windows service logic running so that we can play well in that mode
 	runService("VMware Tether", false)
--- a/cmd/toolbox/main.go
+++ b/cmd/toolbox/main.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cmd/toolbox/main.go
+++ b/cmd/toolbox/main.go
@@ -17,16 +17,20 @@ package main
 import (
 	"flag"
 	"fmt"
-	"log"
 	"os"
 	"os/signal"
 	"syscall"
 
+	log "github.com/Sirupsen/logrus"
+
+	viclog "github.com/vmware/vic/pkg/log"
 	"github.com/vmware/vic/pkg/vsphere/toolbox"
 )
 
 // This example can be run on a VM hosted by ESX, Fusion or Workstation
 func main() {
+	log.SetFormatter(viclog.NewTextFormatter())
+
 	flag.Parse()
 
 	in := toolbox.NewBackdoorChannelIn()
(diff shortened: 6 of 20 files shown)


pkg/log/text_formatter.go
```diff
@@ -15,191 +15,57 @@
...
+}
 
-	b = &bytes.Buffer{}
+func (f *TextFormatter) Format(entry *logrus.Entry) ([]byte, error) {
+	var b bytes.Buffer
 
-	// get the time string
-	ts := f.timeStamp(entry)
+	t := f.timeStamp(entry)
+	l := strings.ToUpper(trimOrPadToSize(entry.Level.String(), 4))
```
caglar10ur: I really prefer having 5 here so that we get DEBUG, ERROR etc.

pkg/log/text_formatter.go
```diff
@@ -15,191 +15,57 @@
...
 
-	if f.Colors {
-		f.printColored(b, entry, keys, ts)
+	if entry.Message == "" {
+		b.WriteString(fmt.Sprintf("%s %s\n", t, l))
```
caglar10ur: [performance] [minor] you could directly write to b avoid another string allocation made by Sprintf

pkg/log/text_formatter.go
```diff
@@ -15,191 +15,57 @@
...
-		if f.Timestamp {
-			f.appendKeyValue(b, "time", ts)
-		}
-
-		f.appendKeyValue(b, "level", entry.Level.String())
-		if entry.Message != "" {
-			f.appendKeyValue(b, "msg", entry.Message)
+		s := bufio.NewScanner(strings.NewReader(entry.Message))
+		for s.Scan() {
+			b.WriteString(fmt.Sprintf("%s %s %s\n", t, l, s.Text()))
```
caglar10ur: same as above

hmahmood: Benchmarks with and without using `fmt.Sprintf` (in that order):

With empty logging message:
```
200000	      8009 ns/op	     532 B/op	      15 allocs/op
300000	      6592 ns/op	     404 B/op	      11 allocs/op
```
With non-empty logging message:
```
100000	     12159 ns/op	    4869 B/op	      20 allocs/op
200000	     10186 ns/op	    4773 B/op	      17 allocs/op
```

Code:
```go
	if entry.Message == "" {
		// b.Grow(len(t) + len(" ") + len(l) + len("\n"))
		// b.WriteString(t)
		// b.WriteString(" ")
		// b.WriteString(l)
		// b.WriteString("\n")
		// return b.Bytes(), nil
		return []byte(fmt.Sprintf("%s %s\n", t, l)), nil
	}

	s := bufio.NewScanner(strings.NewReader(entry.Message))
	for s.Scan() {
		b.WriteString(fmt.Sprintf("%s %s %s\n", t, l, s.Text()))
		// b.Grow(len(t) + len(" ") + len(l) + len(" ") + len(s.Text()) + len("\n"))
		// b.WriteString(t)
		// b.WriteString(" ")
		// b.WriteString(l)
		// b.WriteString(" ")
		// b.WriteString(s.Text())
		// b.WriteString("\n")
	}
```

pkg/log/text_formatter.go
```diff
@@ -15,191 +15,67 @@
...
+	return s + strings.Repeat(" ", size-len(s))
 }
 
 func (f *TextFormatter) Format(entry *logrus.Entry) ([]byte, error) {
-	var b *bytes.Buffer
-	keys := make([]string, 0, len(entry.Data))
-	for k := range entry.Data {
-		keys = append(keys, k)
+	t := f.timeStamp(entry)
+	l := strings.ToUpper(trimOrPadToSize(entry.Level.String(), 5))
```
jzt: Minor, make 5 a constant

pkg/log/text_formatter.go
```diff
@@ -15,191 +15,67 @@
...
+
+	var b bytes.Buffer
+
+	if entry.Message == "" {
+		b.Grow(len(t) + len(" ") + len(l) + len("\n"))
```
jzt: You could probably make this simpler by just calling `b.WriteString(...)` here as `WriteString` calls `grow(len(s))` before appending:

```go
func (b *Buffer) WriteString(s string) (n int, err error) {
   	b.lastRead = opInvalid
   	m := b.grow(len(s))
   	return copy(b.buf[m:], s), nil
   }
   
```

jzt: Edit: Ignore if you're trying to be explicit. :)

pkg/log/text_formatter.go
vburenin: @caglar10ur @hmahmood fmt.SPrintf is using sync.Pool - which causes less allocations.
This is the only reason SPrintf is faster. I even remember fix in Go code related to that.

vburenin:
I would prefer logger API abstraction rather then using logrus. There is a movement to integrate a better logger into Go so all people can use it instead of diverse zoo of different implementations.

pkg/log/text_formatter.go
```diff
@@ -15,191 +15,67 @@
...
+		b.WriteString(l)
+		b.WriteString("\n")
+		return b.Bytes(), nil
 	}
 
-	if !f.DisableSorting {
-		sort.Strings(keys)
+	s := bufio.NewScanner(strings.NewReader(entry.Message))
+	for s.Scan() {
+		b.Grow(len(t) + len(" ") + len(l) + len(" ") + len(s.Text()) + len("\n"))
```
jzt: Same here, maybe something like:

```go
entry := t + l + " " + s.Text() + "\n"
b.WriteString(entry)
```

Or perhaps abstract this out into `f.AddEntry()` or something.

pkg/log/text_formatter.go
```diff
@@ -15,191 +15,67 @@
...
+		b.WriteString(" ")
+		b.WriteString(l)
+		b.WriteString("\n")
+		return b.Bytes(), nil
 	}
 
-	if !f.DisableSorting {
-		sort.Strings(keys)
+	s := bufio.NewScanner(strings.NewReader(entry.Message))
+	for s.Scan() {
```
vburenin: What is going on here? I do not understand :-(

pkg/log/text_formatter.go
vburenin: I would not trim anything here. Trimming is not really necessary for log level. Padding is.
```fmt.Printf("%-5s", entry.Level)```

File: cmd/docker/main.go [hunk 8]
```diff
@@ -36,6 +36,7 @@ import (
 	vicbackends "github.com/vmware/vic/lib/apiservers/engine/backends"
 	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/lib/pprof"
+	viclog "github.com/vmware/vic/pkg/log"
```
jzt: This file needs updated copyright.

cmd/port-layer-server/main.go
```diff
@@ -15,25 +15,28 @@
 package main
 
 import (
-	"log"
 	"os"
 	"os/signal"
 	"syscall"
 
+	log "github.com/Sirupsen/logrus"
```
jzt: Update copyright

cmd/tether/main_darwin.go
```diff
@@ -24,6 +24,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 
 	"github.com/vmware/vic/lib/tether"
+	viclog "github.com/vmware/vic/pkg/log"
```
jzt: Update copyright

cmd/tether/main_linux.go
```diff
@@ -24,13 +24,28 @@ import (
 	log "github.com/Sirupsen/logrus"
 
 	"github.com/vmware/vic/lib/tether"
+	viclog "github.com/vmware/vic/pkg/log"
```
jzt: Update copyright

cmd/tether/main_windows.go
```diff
@@ -22,6 +22,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 
 	"github.com/vmware/vic/lib/tether"
+	viclog "github.com/vmware/vic/pkg/log"
```
jzt: Copyright

cmd/toolbox/main.go
```diff
@@ -17,16 +17,20 @@ package main
 import (
 	"flag"
 	"fmt"
-	"log"
 	"os"
 	"os/signal"
 	"syscall"
 
+	log "github.com/Sirupsen/logrus"
```
jzt: Copyright

cmd/vic-dns/main.go
```diff
@@ -25,6 +25,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 
 	"github.com/vmware/vic/lib/dns"
+	vlog "github.com/vmware/vic/pkg/log"
```
jzt: ©️

cmd/vicadmin/vicadm.go
```diff
@@ -43,6 +43,7 @@ import (
 	"github.com/vmware/vic/lib/guest"
 	"github.com/vmware/vic/lib/pprof"
 	"github.com/vmware/vic/pkg/certificate"
+	viclog "github.com/vmware/vic/pkg/log"
```
jzt: ©️

lib/portlayer/util/util.go
```diff
@@ -15,9 +15,10 @@
 package util
 
 import (
-	"log"
 	"net/url"
 	"os"
+
```
jzt: ©️

pkg/log/text_formatter.go
vburenin: Add some comment that tells that you read line by line and log line by line. It is a little more clear.

lib/tether/scp/operation.go
```diff
@@ -18,11 +18,12 @@ import (
 	"bufio"
 	"fmt"
 	"io"
-	"log"
 	"os"
 	"path/filepath"
 	"regexp"
 	"strconv"
+
```
jzt: ©️

lib/tether/scp/scp.go
```diff
@@ -16,9 +16,9 @@ package scp
 
 import (
 	"io"
-	"log"
 	"os"
 
+	log "github.com/Sirupsen/logrus"
```
jzt: ©️

pkg/log/text_formatter.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
```
jzt: 👍

pkg/trace/trace.go
```diff
@@ -40,12 +40,9 @@ var Logger = &logrus.Logger{
 	Out: os.Stderr,
 	// We're using our own text formatter to skip the \n and \t escaping logrus
 	// was doing on non TTY Out (we redirect to a file) descriptors.
-	Formatter: &log.TextFormatter{
-		Timestamp:       true,
-		TimestampFormat: "2006-01-02T15:04:05.000000000Z07:00",
-	},
-	Hooks: make(logrus.LevelHooks),
-	Level: logrus.InfoLevel,
+	Formatter: log.NewTextFormatter(),
```
jzt: ©️opyright

pkg/vsphere/toolbox/service.go
```diff
@@ -17,11 +17,12 @@ package toolbox
 import (
 	"bytes"
 	"fmt"
-	"log"
 	"net"
 	"os"
 	"sync"
 	"time"
+
```
jzt: ©️opyright

cmd/port-layer-server/main.go
vburenin: :-(

jzt:
The rest of the changed files need a copyright update (sorry for previous spam).

pkg/log/text_formatter.go
hmahmood: Actually `fmt.Sprintf` has more allocations as I indicated above.

pkg/log/text_formatter.go
hmahmood: Trying to avoid reallocating buffer multiple times. I tried several combinations; this ones seems to result in the least amount of allocations.

pkg/log/text_formatter.go
hmahmood: Again, the call I used seems to result in the least amount of allocations.

pkg/log/text_formatter.go
hmahmood: Trimming is necessary to align everything properly. Some of the log levels (like verbose) are too long.

pkg/log/text_formatter.go
jzt: 👍

pkg/log/text_formatter.go
hmahmood: Actually using just concatenation as above results in the same number of allocations as using the `WriteString` calls. So I am going to revert to using concatenation as that is a little more readable.

pkg/log/text_formatter.go
```diff
@@ -15,191 +15,67 @@
...
-)
-
-var (
-	baseTimestamp time.Time
-	isTerminal    bool
-)
-
-func init() {
-	baseTimestamp = time.Now()
-	isTerminal = logrus.IsTerminal()
```
fdawg4l: Have you tested this with and without a `tty` attached?  IIRC if you redirect stdout to a file vs let stdout print to console, you got different formatting behavior.  Basically `\n` and control characters would end up in your log files.

hmahmood: I have been looking at our various log files and the output from vic-machine. All appear to be consistent, and I don't see the '\n' characters that we used to see before.

pkg/log/text_formatter.go
```diff
@@ -15,191 +15,86 @@
...
-	case logrus.ErrorLevel, logrus.FatalLevel, logrus.PanicLevel:
-		levelColor = red
-	default:
-		levelColor = blue
+	var b bytes.Buffer
+
+	// prefix each line of the message with timestamp
+	// level information
+	s := bufio.NewScanner(strings.NewReader(entry.Message))
+	// Define a split function that separates on newlines
```
fdawg4l: I thought `Scan()` splits on line by default.

See https://play.golang.org/p/Q8tycOFMM0

pkg/log/text_formatter.go
```diff
@@ -15,191 +15,86 @@
...
-	switch entry.Level {
-	case logrus.DebugLevel:
-		levelColor = gray
-	case logrus.WarnLevel:
-		levelColor = yellow
-	case logrus.ErrorLevel, logrus.FatalLevel, logrus.PanicLevel:
-		levelColor = red
-	default:
-		levelColor = blue
+	var b bytes.Buffer
```
fdawg4l: Feel free to ignore:  can we move this closer to where it's used?

fdawg4l: Also, and it's probably out of scope for this PR, but should we consider making a bytes.Buffer once and reusing it with `Reset()`?

pkg/log/text_formatter_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/Sirupsen/logrus"
+	"github.com/stretchr/testify/assert"
+)
+
+func BenchmarkFormatNonEmpty(b *testing.B) {
```
fdawg4l: Just of curiosity, how much faster / slower is this formatter than the logrus one?

pkg/log/text_formatter.go
hmahmood: It does, but it ignores the last token if it is empty. I actually adapted this from their docs: https://golang.org/pkg/bufio/#Scanner

pkg/log/text_formatter.go
hmahmood: Yes, I can move into `TextFormatter`, would save an allocation per call.

hmahmood: Actually that will not work since the Format call is not synchronized by logrus (entries in the log files get mixed up).

pkg/log/text_formatter_test.go
hmahmood: This change:
```
BenchmarkFormatNonEmpty-4   	  500000	      2056 ns/op	    4512 B/op	       7 allocs/op
BenchmarkFormatEmpty-4   	 3000000	       494 ns/op	     128 B/op	       3 allocs/op
```
With logrus's TextFormatter:
```
BenchmarkFormatNonEmpty-4   	 1000000	      1650 ns/op	     528 B/op	      11 allocs/op
BenchmarkFormatEmpty-4   	 1000000	      1688 ns/op	     528 B/op	      11 allocs/op
```

PR #3702, opened 2017-01-23
Case insensitivity and error reporting on bogus options in volume create

### LINKED ISSUE #3535: Bogus --opt input values on volume create are ignored
This led to the weird situation where I created a volume using the --opt capacity=4gb flag and it created a 1GB volume. Why? Because I didn't capitalize "capacity". Equality, entering --opt bogus=bogus successfully creates a volume with no warning.

This may be a limitation of the Docker client, but I'd like to at least return a warning of some kind that the option wasn't understood

gigawhitlocks:
Fixes #3535 by adding case insensitivity to valid options and a whitelist to allow error reporting if an unsupported option is provided during `docker volume create` or `docker create -v ..`

### DIFF
--- a/lib/apiservers/engine/backends/volume.go
+++ b/lib/apiservers/engine/backends/volume.go
@@ -43,12 +43,12 @@ const (
 )
 
 // define a set (whitelist) of valid driver opts keys for command line argument validation
-var validDriverOptsKeys = map[string]bool{
-	OptsVolumeStoreKey:    true,
-	OptsCapacityKey:       true,
-	DriverArgFlagKey:      true,
-	DriverArgContainerKey: true,
-	DriverArgImageKey:     true,
+var validDriverOptsKeys = map[string]struct{}{
+	OptsVolumeStoreKey:    {},
+	OptsCapacityKey:       {},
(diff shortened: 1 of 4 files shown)


lib/apiservers/engine/backends/volume.go
```diff
@@ -36,11 +37,20 @@ import (
...
 const (
-	OptsVolumeStoreKey     string = "VolumeStore"
-	OptsCapacityKey        string = "Capacity"
+	OptsVolumeStoreKey     string = "volumestore"
+	OptsCapacityKey        string = "capacity"
 	dockerMetadataModelKey string = "DockerMetaData"
 )
 
+// define a set (whitelist) of valid driver opts keys for command line argument validation
+var validDriverOptsKeys = map[string]bool{
```
matthewavery: you might consider just using struct{} instead of true. This is a common go idiom for making a set :)

matthewavery:
Talked out of band, other than the other two issues and am ready to approve :)

matthewavery:
lgtm

sflxn:
lgtm

PR #3705, opened 2017-01-24
Remove SetLevel calls from extraconfig pkg

caglar10ur:
Setting the log level should be caller's responsibility.


caglar10ur:
Dropping this PR as looks like that individual log level sorcery is there to hide the password from the logs.

Issue #1859, opened 2016-08-08
Missing functions in vcSim currently found in vic-machine unit test


emlin:
Here list the missing functions required by vic-machine unit test:
- [x] VirtualDeviceList.FindIDEController
- [x] Host.ConfigManager.NetworkSystem, and create distributed virtual switch and distributed port group
- [x] Read and write VM guestinfo
- [x] HostDatastoreBrowser:hostdatastorebrowser-34 does not implement: SearchDatastoreSubFolders_Task
- [x] SearchIndex. Error message: "no such object: SearchIndex:ha-searchindex"

emlin:
@dougm I opened one bug for you, to list all issues I met for vic-machine unit test. Just for your reference.

mdubya66:
Moving to Icebox per 9/20 Triage

dougm:
"Finder.ObjectReference does not return find.NotFoundError" - also the case against real ESX/VC, the method returns types.ManagedObjectNotFound in both cases.  I'm fixing the code/test to check for that.

"SearchDatastore does not return FileNotFound error" - it does, but the test was using an InvalidDatastorePath - upcoming PR fixes the test.

dougm:
I removed CreateVApp from the list above, since vic-machine handles that case gracefully and we may move away from VApps anyhow.  We can have another story to implement VApp methods if needed.

mhagen-vmware:
lgtm

PR #3706, opened 2017-01-24
vcsim: Add Portgroup creation to VPX Model

dougm:
- Add ReconfigureDvs_Task support

- Add ReconfigVM_Task support

- Return moref in Folder.CreateDVSTask

- Add a nic to VMs by default in Model.Create

- Add default VM device description

- Filter 'Self' field in property collector

Issue #1859

### DIFF
--- a/pkg/vsphere/simulator/dvs_test.go
+++ b/pkg/vsphere/simulator/dvs_test.go
@@ -16,6 +16,7 @@ package simulator
 
 import (
 	"context"
+	"reflect"
 	"testing"
 
 	"github.com/vmware/govmomi/find"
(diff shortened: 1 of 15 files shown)


pkg/vsphere/simulator/dvs_test.go
```diff
@@ -0,0 +1,179 @@
...
+		t.Fatalf("err=%v", err)
+	}
+
+	// Edit == fail (NotSupported)
+	for i := range config.Host {
+		config.Host[i].Operation = string(types.ConfigSpecOperationEdit)
+	}
+
+	dtask, err = dvs.Reconfigure(ctx, config)
+	if err != nil {
```
vburenin: Have you considered using assert or similar testing libraries? It could save you 2 lines per check.

pkg/vsphere/simulator/model.go
```diff
@@ -125,11 +130,25 @@ func (m *Model) Create() error {
...
 
+		if dvs != nil {
+			config := &types.DVSConfigSpec{
+				Host: []types.DistributedVirtualSwitchHostMemberConfigSpec{{
+					Operation: string(types.ConfigSpecOperationAdd),
+					Host:      host.Reference(),
+				}},
+			}
+
+			_, _ = dvs.Reconfigure(ctx, config)
```
vburenin: dvs.Reconfigure(ctx, config) - is enough.

pkg/vsphere/simulator/dvs_test.go
dougm: I have in the past.. but this test would still be annoyingly verbose, I'm converting it to a table driven test.

pkg/vsphere/simulator/model.go
dougm: That's for silencing go-errcheck when intentionally ignoring an error.

flycheck

PR #3713, opened 2017-01-24
Finalizes changes to the volume store flags in vic machine [specific ci=Group6-VIC-Machine]

### LINKED ISSUE #2899: Create VCH should create a 'default' volume store (esp for the images which has built with dockerfile)
I faced this issue when playing with VIC-Harbor integration testing.
Steps:
1. Create VCH - with insecure-registry - harbor registry (with no default volume store)
root@photon-yhNySW7sg [ ~/vic ]# ./vic-machine-linux create --name VCH2 --target 10.139.23.227 --compute-resource cls --user administrator@vsphere.local --password 'Admin!23' --bridge-network vicDVPG --image-store sharedVmfs-0 --no-tls --insecure-registry 10.139.23.226:80
2. Do docker login
3. Pull an image to local registry
4. Tag and Push that image to Harbor registry
5. Now, Pull it back from Harbor
6. Run the application
docker -H 10.139.23.221:2375 run --name nginxharbor 10.139.23.226:80/heimdall/nginx:1.9.0

**VIC Version:**
root@photon-yhNySW7sg [ ~/vic ]# ./vic-machine-linux version
vic-machine-linux version v0.6.0-4890-4f98611
Server:
 Version:      0.6.0
 API version:  1.23

**Expected behavior:**
VCH should either create a default volume store or fail in validation to ask user to create a default volume store

**Actual behavior:**
Running the application fails with "docker: Error response from daemon: No volume store named (default) exists."

matthewavery:
Fixes #2899 

This will add the --no-default-volume-store flag allowing users to explicitly create a VCH without a default VolumeStore. Additionally, overlapping paths to the same label are now detected and reported with a human readable error. default with the same path may be set with the -vsd and -vs with out error. if -vsd and -vs conflict for default a custom error message is displayed. This PR also adds integration tests which cover all the volume store declaration functionality. This will protect us from breaking backwards compatibility and flag behavior in the future(unless intentionally desired).


Makefile
```diff
@@ -141,7 +141,7 @@ misspell: $(MISSPELL)
 # convenience targets
 all: components tethers isos vic-machine vic-ui
 tools: $(GOIMPORTS) $(GVT) $(GOLINT) $(SWAGGER) $(GAS) $(MISSPELL) goversion
-check: goversion goimports gofmt misspell govet golint copyright whitespace gas
+check: goversion goimports gofmt govet golint copyright whitespace gas
```
anchal-agrawal: 👍

cmd/vic-machine/create/create.go
```diff
@@ -543,6 +550,10 @@ func (c *Create) processParams() error {
 		return err
 	}
 
+	if err := c.processDefaultVolumeStore(); err != nil {
```
vburenin: processDefaultVolumeStore will never return error. Just saying...

matthewavery: Currently it will not, neither will several other functions called here. I decided to put it in this context because the surrounding process functions also have this pattern. I will look at changing it :)

cmd/vic-machine/create/create.go
```diff
@@ -864,17 +875,47 @@ func (c *Create) processDNSServers() error {
...
+
+		switch label {
+		case defaultVolumeLabel:
+			otherPath, ok := c.VolumeLocations[label]
+			if !ok {
+				c.VolumeLocations[label] = path
+				continue
+			}
+			if path != otherPath {
+				return errors.Errorf("Multiple paths were tagged for the same label(%s), volumestore labels can only have one distinct path. The preferred method of setting the default store is the --default-volume-store flag.", label)
```
anchal-agrawal: Cosmetic: Error strings should start with lowercase.

cmd/vic-machine/create/create.go
```diff
@@ -864,17 +875,47 @@ func (c *Create) processDNSServers() error {
...
+			}
+		default:
+			if _, ok := c.VolumeLocations[label]; ok {
+				return errors.Errorf("Multiple paths were tagged for the same label(%s), volumestore labels can only have one distinct path.", label)
```
anchal-agrawal: Ditto.

cmd/vic-machine/create/create.go
```diff
@@ -864,17 +875,47 @@ func (c *Create) processDNSServers() error {
...
+	return nil
+}
+
+func (c *Create) processVolumeStores() error {
+	defer trace.End(trace.Begin(""))
 	for _, arg := range c.volumeStores {
+
 		splitMeta := strings.SplitN(arg, ":", 2)
 		if len(splitMeta) != 2 {
 			return errors.New("Volume store input must be in format datastore/path:label")
```
emlin: volume store format could be datastore:lable as well. Might update this error message, please?

cmd/vic-machine/create/create.go
emlin: nice description 👍

cmd/vic-machine/create/create.go
```diff
@@ -864,17 +875,47 @@ func (c *Create) processDNSServers() error {
...
+				continue
+			}
+			if path != otherPath {
+				return errors.Errorf("Multiple paths were tagged for the same label(%s), volumestore labels can only have one distinct path. The preferred method of setting the default store is the --default-volume-store flag.", label)
+			}
+		default:
+			if _, ok := c.VolumeLocations[label]; ok {
+				return errors.Errorf("Multiple paths were tagged for the same label(%s), volumestore labels can only have one distinct path.", label)
+			}
+			c.VolumeLocations[label] = path
```
emlin: thought your two switches actually did same thing. But slightly different. 
If you specified multiple paths for same label, you report error no matter if they are same.
But if default is specified, you'll compare if they are same.
Just a minor issue, but might worth some consolidation?

cmd/vic-machine/create/create.go
matthewavery: Sure, do have a suggestion for what it should be?

cmd/vic-machine/create/create.go
matthewavery: Yeah I actually wanted to make a function that I just passed a string too. I think I will do that.

lib/install/validate/storage.go
```diff
@@ -53,16 +53,27 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
...
-		dsURL, _, err := v.DatastoreHelper(ctx, volDSpath, label, "--volume-store")
-		v.NoteIssue(err)
-		if dsURL != nil {
-			conf.VolumeLocations[label] = dsURL
+
+		switch label {
+		case "default":
+			dsURL, _, err := v.DatastoreHelper(ctx, volDSpath, label, "--default-volume-store")
+			if err != nil {
+				v.NoteIssue(fmt.Errorf("%s : %s", err, "validation failed for --default-volume-store, please check that the supplied datastore path is properly formatted."))
```
vburenin: Not sure this message is good: I would phrase it as :
"supplied default volume store path is not properly formatted."

cmd/vic-machine/create/create.go
```diff
@@ -181,6 +182,12 @@ func (c *Create) Flags() []cli.Flag {
 		},
 
 		// volume
+		cli.StringFlag{
+			Name:        "default-volume-store, vsd",
+			Value:       "",
+			Usage:       "Specify the default volume store location, This location is used to support anonymous volumes such as image volumes. e.g.  \"datastore/path\" or \"datastore\".",
```
anchal-agrawal: `location, this location` or `location. This location`, or perhaps `location, which is`?

cmd/vic-machine/create/create.go
```diff
@@ -543,6 +550,10 @@ func (c *Create) processParams() error {
 		return err
 	}
 
+	if err := c.processDefaultVolumeStore(); err != nil {
+		return errors.Errorf("Error occurred while processing the default volume stores: %s", err)
```
anchal-agrawal: Cosmetic: error strings should start with lowercase.

cmd/vic-machine/create/create.go
```diff
@@ -864,17 +875,55 @@ func (c *Create) processDNSServers() error {
...
+}
+
+func (c *Create) processVolumeStores() error {
+	defer trace.End(trace.Begin(""))
 	for _, arg := range c.volumeStores {
+
 		splitMeta := strings.SplitN(arg, ":", 2)
 		if len(splitMeta) != 2 {
-			return errors.New("Volume store input must be in format datastore/path:label")
+			return errors.New("Volume store input must be in format datastore:label or datastore/path/to/volumestore:label")
```
anchal-agrawal: Lowercase here as well.

cmd/vic-machine/create/create.go
gigawhitlocks: Whichever wording you wind up with here, `", This"` is a comma splice so either make it two sentences or use a semicolon or avoid this wording entirely ;)

matthewavery: Not you TOO!

cmd/vic-machine/create/create.go
```diff
@@ -181,6 +182,12 @@ func (c *Create) Flags() []cli.Flag {
 		},
 
 		// volume
+		cli.StringFlag{
+			Name:        "default-volume-store, vsd",
```
anchal-agrawal: Should the short flag be `dvs`?

lib/install/validate/storage.go
matthewavery: Sure thing! It has been done :)

cmd/vic-machine/create/create.go
fdawg4l: Do we need a flag for `default`?  I though we were going to complain if a `volume-store` with the name `default` wasn't specified unless a flag was added to explicitly not created on.  Like `--no-default-volumestore` or some such.  This overloads an already existing flag of `volume -store` so the question is what if someone uses that flag and specifies a name of "default" using it.  Isn't that the same thing?

cmd/vic-machine/create/create.go
fdawg4l: I'm not sure what this error is meant to convey.  What does "process" mean?  And looking at the function itself, it returns nil for error exclusively.  So this error will never be thrown.

cmd/vic-machine/create/create.go
```diff
@@ -864,17 +875,55 @@ func (c *Create) processDNSServers() error {
...
-			return errors.New("Volume store input must be in format datastore/path:label")
+			return errors.New("Volume store input must be in format datastore:label or datastore/path/to/volumestore:label")
+		}
+		label := splitMeta[1]
+		path := splitMeta[0]
+		var err error
+
+		switch label {
+		case defaultVolumeLabel:
+			customErr := errors.Errorf("multiple paths were tagged for the same label(%s), volumestore labels can only have one distinct path. The preferred method of setting the default store is the --default-volume-store flag.", label)
```
fdawg4l: "label"?  When did we adopt that verbiage?  Please check the documentation and make sure we're not inventing a new word here.

lib/config/virtual_container_host.go
```diff
@@ -122,6 +122,7 @@ type Storage struct {
 	// Datastore URLs for image stores - the top layer is [0], the bottom layer is [len-1]
 	ImageStores []url.URL `vic:"0.1" scope:"read-only" key:"image_stores"`
 	// Permitted datastore URL roots for volumes
+
```
fdawg4l: Random new line.  Id imagine if you were to add a line for readability, it would go above this one (on line 124).

lib/install/validate/storage.go
```diff
@@ -53,16 +53,27 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
...
-
-	// TODO: add volume locations
+	//volume location population
 	for label, volDSpath := range input.VolumeLocations {
-		dsURL, _, err := v.DatastoreHelper(ctx, volDSpath, label, "--volume-store")
-		v.NoteIssue(err)
-		if dsURL != nil {
-			conf.VolumeLocations[label] = dsURL
+
+		switch label {
```
fdawg4l: The 2 cases are almost completely identical.  Can we nuke the `switch`, and use a simple `if {} else {} ` for the ` field that's different?

fdawg4l:
Other than the small issues in the code, I'm ok with the change.  But please address the comments I made.

However, before doing that, lets get clarification from @hickeng about the flags we want to introduce and make sure what you've implemented matches.  I noticed you asked in the issue (https://github.com/vmware/vic/issues/2899#issuecomment-270720679) but there didn't seem to be a response noted there.  Maybe you guys talked out of band, but lets close the loop on that so we're all in agreement.

Thanks!

lib/install/validate/storage.go
matthewavery: @fdawg4l We could, but if we wanted to add more in the future this makes it easier. I don't think we will.

lib/config/virtual_container_host.go
matthewavery: fixed, that is odd... this was here before and I fixed it before. not sure why it came out.

cmd/vic-machine/create/create.go
matthewavery: I did not realize that is what we wanted from the end of that ticket. We can also do that. I think having this flag is more user friendly, but we can get rid of it as well.

lib/install/validate/storage.go
fdawg4l: Yeah, if there were 3 cases, then yeah, maybe.  But as written, it's almost completely identical.  So lets unify where we can.

cmd/vic-machine/create/create.go
```diff
@@ -864,15 +875,44 @@ func (c *Create) processDNSServers() error {
...
+		var err error
+
+		splitDriverArgs := strings.SplitN(arg, ":", 2)
+		if len(splitDriverArgs) != 2 {
+			return errors.New("volume store input must be in format datastore:label or datastore/path/to/volumestore:label")
+		}
+		label := splitDriverArgs[1]
+		path := splitDriverArgs[0]
+
+		switch label {
```
hmahmood: I think if you move this switch under the `if path != otherPath` the code would be more readable.

cmd/vic-machine/create/create.go
```diff
@@ -864,15 +875,44 @@ func (c *Create) processDNSServers() error {
...
+		case defaultVolumeLabel:
+			err = errors.Errorf("multiple paths were tagged for the same label(%s), volumestore labels can only have one distinct path. The preferred method of setting the default store is the --default-volume-store flag.", label)
+		default:
+			err = errors.Errorf("multiple paths were tagged for the same label(%s), volumestore labels can only have one distinct path.", label)
+		}
+
+		otherPath, ok := c.VolumeLocations[label]
```
hmahmood: What if `label == ""` here?

matthewavery:
Per talking with @hickeng this is what we will do:

cmd/vic-machine/create/create.go
matthewavery: Good catch added a check for this as well. It should not be allowed since it is not helpful or readable it will be filtered out.

lib/install/validate/validator.go
```diff
@@ -548,6 +548,9 @@ func (v *Validator) checkDatastoresAreWriteable(ctx context.Context, conf *confi
 	validStores := make(map[types.ManagedObjectReference]*object.Datastore)
 	// remove any found datastores from requested datastores
 	for _, cds := range clusterDatastores {
+		var ds mo.Datastore
```
matthewavery: This shouldn't be here, accidently got included when I was talking to Mike Hagen.

matthewavery:
We are waiting for input from customers + Karthik as to how to proceed with this issue.

sgairo:
copyright range on validate.go -copyright bot :)

hickeng:
@karthik-narayan can we just go ahead with the updates to volume option behaviour. It sounds like we don't have customer feedback available.

lib/install/validate/validator.go
pdaigle: I think the requirements above are OK.
The limited customer feedback I have gathered on Volumes indicates that _"vic-machine error if no "default" volume store is provided"_ will be especially useful as many do not realize this will break anonymous volumes and images using them.

karthik-narayan:
@hickeng I'd asked @pdaigle to chime in and he did above. Here is his reply

"I think the requirements above are OK.
The limited customer feedback I have gathered on Volumes indicates that "vic-machine error if no "default" volume store is provided" will be especially useful as many do not realize this will break anonymous volumes and images using them."

PR #3717, opened 2017-01-24
[specific ci=Group1-Docker-Commands]Quarantine the time tests until we can fix the sync problem with our servers

mhagen-vmware:


mhagen-vmware:
Was able to actually fix the time on our ESX servers, by changing to public time servers.

PR #3722, opened 2017-01-25
[full ci]Do not remove container if it's fixing

### LINKED ISSUE #3196: Error response from daemon: No such container: 1069fb0efd2f

emlin:
@hickeng agree to drop the error handler solution in PR #3478 for the benefit of that solution for vmomit gateway is not clear yet, but will introduce much more error to manage handlers from everywhere.
And the issue of invalid state VM is not easy to be avoided even we didn't have non-persistent guestinfo values.

This PR is the easiest solution so far to workaround the problem.
fix #3196

### DIFF
--- a/lib/portlayer/exec/exec.go
+++ b/lib/portlayer/exec/exec.go
@@ -162,13 +162,10 @@ func eventCallback(ie events.Event) {
 				}()
 			case StateRemoved:
 				log.Debugf("Container(%s) %s via event activity", container, newState)
-				if container.vm != nil {
-					isFixing := container.vm.Fixing.Load()
-					if isFixing != nil && isFixing.(bool) {
-						// is fixing vm, which will be registered back soon, so do not remove from containers cache
-						log.Debugf("Container(%s) %s is being fixed", container.ExecConfig.ID)
-						break
-					}
+				if container.vm != nil && container.vm.IsFixing() {
+					// is fixing vm, which will be registered back soon, so do not remove from containers cache
+					log.Debugf("Container(%s) %s is being fixed", container.ExecConfig.ID)
+					break
 				}
 				Containers.Remove(container.ExecConfig.ID)
 				publishContainerEvent(container.ExecConfig.ID, ie.Created(), ie.String())
--- a/lib/portlayer/exec/exec.go
+++ b/lib/portlayer/exec/exec.go
@@ -167,7 +167,6 @@ func eventCallback(ie events.Event) {
 					if isFixing != nil && isFixing.(bool) {
 						// is fixing vm, which will be registered back soon, so do not remove from containers cache
 						log.Debugf("Container(%s) %s is being fixed", container.ExecConfig.ID)
-						container.vm.Fixing.Store(false)
 						break
 					}
 				}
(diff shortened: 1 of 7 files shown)


emlin:
@hickeng @hmahmood @vburenin This PR is the solution without error handler. Please help to have another round of review.
thanks

lib/portlayer/exec/exec.go
```diff
@@ -169,6 +162,15 @@ func eventCallback(ie events.Event) {
 				}()
 			case StateRemoved:
 				log.Debugf("Container(%s) %s via event activity", container, newState)
+				if container.vm != nil {
+					isFixing := container.vm.Fixing.Load()
+					if isFixing != nil && isFixing.(bool) {
+						// is fixing vm, which will be registered back soon, so do not remove from containers cache
+						log.Debugf("Container(%s) %s is being fixed", container.ExecConfig.ID)
+						container.vm.Fixing.Store(false)
```
hmahmood: Why are we setting it to false here? Shouldn't it be set to false in `fixVM` after the fixing is done?

emlin: make sense, should reset in vm fixing method

pkg/vsphere/vm/vm.go
```diff
@@ -52,6 +53,9 @@ type VirtualMachine struct {
 	*object.VirtualMachine
 
 	*session.Session
+
+	// Fixing is true means the VM will be unregistered and registered back.
+	Fixing *atomic.Value
```
vburenin: for clarity if I was doing that I would use an integer value for this flag.
atomic.Value is a lot of overhead for a simple flag.

```
// supposed to be atomic.
	var fixFlag int32
	
	v := atomic.LoadInt32(&fixFlag)
	atomic.StoreInt32(&fixFlag, 1)
```

lib/portlayer/exec/exec.go
```diff
@@ -169,6 +162,14 @@ func eventCallback(ie events.Event) {
 				}()
 			case StateRemoved:
 				log.Debugf("Container(%s) %s via event activity", container, newState)
+				if container.vm != nil {
+					isFixing := container.vm.Fixing.Load()
```
vburenin: you would not need this type assertion at all if you use atomic integer.

vburenin: vm object wants to have a method isFixing.

PR #3723, opened 2017-01-25
Skip logging header for Entry lines with embedded CRs

fdawg4l:
We don't need a header for each line in a log entry lines with embedded carriage returns.  Print the header once for each entry.

Before
```
BenchmarkFormatNonEmpty-4   	  500000	      2056 ns/op	    4512 B/op	       7 allocs/op
BenchmarkFormatEmpty-4   	 3000000	       494 ns/op	     128 B/op	       3 allocs/op
```

After 
```
BenchmarkFormatNonEmpty-4   	 3000000	       597 ns/op	     224 B/op	       3 allocs/op
BenchmarkFormatEmpty-4      	 3000000	       515 ns/op	     128 B/op	       3 allocs/op
```

### DIFF
--- a/pkg/log/text_formatter.go
+++ b/pkg/log/text_formatter.go
@@ -55,22 +55,7 @@ func (f *TextFormatter) Format(entry *logrus.Entry) ([]byte, error) {
 	t := f.timeStamp(entry)
 	l := levelToString(entry.Level)
 
-	var buf []byte
-	switch {
-	case entry.Message == "":
-		buf = []byte(t + " " + l + "\n")
-
-	case entry.Message[0] == '\n':
-		buf = []byte(t + " " + l + " " + entry.Message[1:])
-
-	case entry.Message[len(entry.Message)-1] == '\n':
-		buf = []byte(t + " " + l + " " + entry.Message)
-
-	default:
-		buf = []byte(t + " " + l + " " + entry.Message + "\n")
-	}
-
-	return buf, nil
+	return []byte(t + " " + l + " " + entry.Message + "\n"), nil
 }
 
 func (f *TextFormatter) timeStamp(entry *logrus.Entry) string {
--- a/pkg/log/text_formatter.go
+++ b/pkg/log/text_formatter.go
@@ -14,13 +14,7 @@
 
 package log
 
-import (
-	"bufio"
-	"bytes"
-	"strings"
-
-	"github.com/Sirupsen/logrus"
-)
+import "github.com/Sirupsen/logrus"
 
 // level strings padded to match the length of the longest level,
 // which is "UNKNOWN" currently. Indexed according to levels in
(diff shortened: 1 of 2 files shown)


fdawg4l:
> We are modifying the original message from the caller by dropping the trailing new line. If it was the caller's intent to insert a blank line after their message in one call we should let them.

The only place we do this is in `vic-machine`.  Frankly, I don't think we should be using the logging facility to report status to the interactive user.  But, I don't feel like changing it so I can nuke the trailing slash case.

> If we don't want to prefix the header, the code is much simpler; we can just output the message followed by a new line. Personally I think adding the header is much cleaner and easier to read. 

Assuming you're referring to log lines with embedded CRs.  Evaluating every byte in your logfile just to add a header on line breaks is kind of overkill.  I'd rather do without it.  And honestly, if I add a CR to a log message, the header in the preceding line is all I need to debug what happened chronologically.

hmahmood:
@fdawg4l its pretty simple then: just return `[]byte(t + " " + l + " " + entry.Message + "\n")`.

hickeng:
@hmahmood @fdawg4l you two are very opinionated about logging formatting, so before any logging changes get made, address the requirements. Please identify:

1. what our primary requirements are for logging, including
  a. when we need to allow multi-line log messages via the logging mechanism (it's a pretty print variety and should therefore be rare or debug only in general)
  b. does logrus support these requirements, if not what are the gaps and how can we address them.
  c. how should trace.Operation work and what does this mean for logging
2. what are our primary requirements for vic-machine user targeted output - using the logging mechanism for this was expedient but a serious abuse as it prevents us from providing status bars for iso upload and any other sort of ncurses style operation

If logrus is the elephant in the room then any changes to logging are pointless until this is addressed

hmahmood:
@hickeng my initial PR was just to address the logging format, not requirements for logging in general. I intend to take the different issues with logging on one at a time. I generally agree with @fdawg4l on these changes; I just think they can be much simpler for the goals outlined.

Logging format is going to be the same regardless of whether we use logrus or not, so that is why I opted to tackle that first. The only reason in my opinion to use one library over the other would be performance; they all pretty much offer the same features, and we only need a few to be present:

1. support for custom formats
2. support for structured logging, which may help us with tracing

logrus satisfies these right now, so that is why I haven't made any changes to strip it out, and that is why I would like to wait till the other changes are in before worrying about performance.

hickeng:
@hmahmood where does the requirement for multi-line processing in log messages come from? Is it a pervasive requirement or a rare case where we want pretty print? If it's the latter then using a different logger that's configured for pretty print might make sense instead of scanning all messages.
Without defining a usage requirement for the pretty print case it's hard to identify what the preferred solution is.

hmahmood:
The multiline requirement mainly comes from debug messages which may include things like a stack trace. The Operation object in particular does that. Also I noticed some variables when printed verbosely have multiline output. I don't think multiline output is the norm for other than debug messages, so you're right it may only be an edge concern when it comes to performance. 

I am not too hung up on how we output the multiline messages, other than we should not modify them. If we take out the `Scanner` in this code, and just output the message without walking it, it would be fine with me.

PR #3735, opened 2017-01-26
Cleanup port bindings for containers that poweroff OOB and are rm'd

### LINKED ISSUE #3465: Race between container remove and start when cleaning up port mappings
There is a race between [cleanupPortBindings()](https://github.com/vmware/vic/blob/master/lib/apiservers/engine/backends/container.go#L501) and another go routine running a container remove operation. Specifically, the `GetContainer()` call may not find the container that has already been removed, but was still in the cache when the `containerByPort` map was checked.

Result: container start fails for a new container that may have conflicting port mappings with the container that was removed. This condition persists for any new container that is trying to create the same port mapping, since the `containerByPort` still has the port binding in there.

Fix: `UnmapPorts` and `cleanupPortBindings` should be idempotent, i.e. they should ignore conditions where they find the container is missing in the cache, and assume that it has been removed by another go routine, and that port mappings have already been cleaned up.

anchal-agrawal:
Unmaps ports for containers that power off out-of-band and are then removed. This resolves the failure where a new container with the same mappings would report that the port is unavailable and fail to start.

Also fixes a race between container start and remove, and adds an integration test.

Fixes #3465

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -473,7 +473,7 @@ func (c *Container) cleanupPortBindings(vc *viccontainer.VicContainer) error {
 			if cc == nil {
 				// The container was removed from the cache and
 				// port bindings were cleaned up by another operation.
-				return nil
+				continue
 			}
 			running, err := c.containerProxy.IsRunning(cc)
 			if err != nil {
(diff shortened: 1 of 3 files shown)


mhagen-vmware:
lgtm

andrewtchin:
🐶 much lgtm wow 👍

sflxn:
lgtm

PR #3743, opened 2017-01-27
vcsim: Model improvements in support of vic-machine

dougm:
- Fixup cross resource morefs that vic-machine depends on

- Return MethodNotFound fault for unimplemented methods

- Set default cluster config

- Implement SearchIndex.FindChild

- Enable/fixup vic-machine related unit tests that run against vcsim

Issue #1859

### DIFF
--- a/lib/install/management/delete_test.go
+++ b/lib/install/management/delete_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/lib/install/data"
 	"github.com/vmware/vic/lib/install/validate"
+	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/session"
 	"github.com/vmware/vic/pkg/vsphere/simulator"
@@ -77,7 +78,6 @@ func TestDelete(t *testing.T) {
 		}
 
 		validator.DisableFirewallCheck = true
-		validator.DisableDRSCheck = true
 
 		conf, err := validator.Validate(ctx, input)
 		if err != nil {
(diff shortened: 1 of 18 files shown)


pkg/vsphere/simulator/virtual_machine.go
```diff
@@ -299,6 +309,10 @@ func (vm *VirtualMachine) configureDevice(devices object.VirtualDeviceList, devi
 		devices.AssignController(device, controller)
 	}
 
+	if d.Key == -1 {
```
caglar10ur: I think API accepts any negative integer as a temporary key (https://www.vmware.com/support/developer/vc-sdk/visdk25pubs/ReferenceGuide/vim.vm.device.VirtualDevice.html) so maybe use d.Key < 0?

dougm: Yes, I am planning to fix this within object.VirtualDeviceList.  In some cases it uses NewKey(), which returns a negative number unique to the list, in others it just uses -1.  So we're just making the Key unique on the vcsim side in that particular case.  vcsim also leaves the Key as a negative number currently, would also like to take care of that within VirtualDeviceList.

lib/install/management/delete_test.go
```diff
@@ -103,7 +104,7 @@ func createAppliance(ctx context.Context, sess *session.Session, conf *config.Vi
 		isVC:    sess.IsVC(),
 		force:   false,
 	}
-	delete(conf.Networks, "bridge") // FIXME: cannot create bridge network in simulator
+
 	if d.isVC {
 		if d.vchVapp, err = d.createVApp(conf, vConf); err != nil {
 			// FIXME: Got error: ServerFaultCode: ResourcePool:resourcepool-14 does not implement: CreateVApp. Simulator need to implement CreateVApp
```
emlin: Looks like this is fixed as well. Might remove it as well? 😄

emlin:
LGTM

lib/install/management/delete_test.go
dougm: I left it for now since CreateVApp still isn't implement by the simulator.  Mentioned in the issue: https://github.com/vmware/vic/issues/1859#issuecomment-275544136

PR #3758, opened 2017-01-30
Detach any attached disks on startup of the VCH [specific ci=10-01-VCH-Restart]

### LINKED ISSUE #309: Verify no image disks are attached to appliance vm on portlayer daemon startup
As a VI admin and VIC user, restarts of the VCH should not lead to orphaned disks attached to the appliance.

It is assumed no disks (other than required disks like root and whatever cache disks we end up using) are attached to the portlayer appliance vm.  

The expectation is while doing a pull, and the VCH has a disk attached, if the VCH (or host) is power cycled or restarted, the VCH should detach and clean up any half written or inconsistent disks.  Removal is already happening, but the detach on startup needs to be added.

TODO
 - [ ] come up with a simple design of how to address this
 - [ ] implement the design
 - [ ] write the integration test to exercise the scenario we're guarding against.

fdawg4l:
Detach any attached disks on startup.

Fixes #309

### DIFF
--- a/pkg/vsphere/disk/disk_manager.go
+++ b/pkg/vsphere/disk/disk_manager.go
@@ -291,12 +291,10 @@ func (m *Manager) Detach(op trace.Operation, d *VirtualDisk) error {
 		return errors.Trace(err)
 	}
 
-	func() {
-		select {
-		case <-m.maxAttached:
-		default:
-		}
-	}()
+	select {
+	case <-m.maxAttached:
+	default:
+	}
 
 	return d.setDetached()
 }
@@ -336,13 +334,14 @@ func (m *Manager) detachAll(op trace.Operation) error {
 	}
 
 	for _, disk := range disks {
-		if err = m.detach(op, disk); err != nil {
-			op.Errorf("error detaching disk: %s", err.Error())
-			return err
+		if er := m.detach(op, disk); err != nil {
+			// late exit on error
+			op.Errorf("error detaching disk: %s", er.Error())
+			err = er
 		}
 	}
 
-	return nil
+	return err
 }
 
 func (m *Manager) devicePathByURI(op trace.Operation, datastoreURI string) (string, error) {
--- a/pkg/vsphere/disk/disk_manager.go
+++ b/pkg/vsphere/disk/disk_manager.go
@@ -72,6 +72,12 @@ func NewDiskManager(op trace.Operation, session *session.Session) (*Manager, err
 		controller:   controller,
 		byPathFormat: byPathFormat,
 	}
+
+	// Remove any attached disks
+	if err = d.detachAll(op); err != nil {
+		return nil, err
+	}
+
 	return d, nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/vsphere/disk/disk_manager.go
```diff
@@ -275,24 +281,40 @@ func (m *Manager) Detach(op trace.Operation, d *VirtualDisk) error {
...
 	if err != nil {
 		return errors.Trace(err)
 	}
 
+	if err = m.detach(op, disk); err != nil {
+		op.Errorf("detach for %s failed with %s", d.DevicePath, errors.ErrorStack(err))
+		return errors.Trace(err)
+	}
+
+	func() {
```
vburenin: I am a little puzzled, what is the purpose if this anonymous function here?

pkg/vsphere/disk/disk_manager.go
```diff
@@ -303,23 +325,28 @@ func (m *Manager) Detach(op trace.Operation, d *VirtualDisk) error {
...
+		return err
 	}
 
-	func() {
-		select {
-		case <-m.maxAttached:
-		default:
+	for _, disk := range disks {
+		if err = m.detach(op, disk); err != nil {
+			op.Errorf("error detaching disk: %s", err.Error())
```
vburenin: does it make sense to do earlier exit if one of the possible detaches failed? Wouldn't it make sense to detach all and return last/first error?

pkg/vsphere/disk/util.go
```diff
@@ -242,15 +239,49 @@ func findDisk(op trace.Operation, vm *vm.VirtualMachine, name string) (*types.Vi
...
+
+	if len(candidates) == 0 {
+		return nil, nil
+	}
+
+	disks := make([]*types.VirtualDisk, len(candidates))
+	for idx, disk := range candidates {
+		d, ok := disk.(*types.VirtualDisk)
+		if !ok {
+			// will never happen
```
vburenin: if it never happens, this logic should not be here. You can just say the pre-filtering already happened right before this.

pkg/vsphere/disk/util.go
```diff
@@ -260,5 +291,22 @@ func findDisk(op trace.Operation, vm *vm.VirtualMachine, name string) (*types.Vi
...
-	return candidates[0].(*types.VirtualDisk), nil
+	return candidates[0], nil
+}
+
+func findAllDisks(op trace.Operation, vm *vm.VirtualMachine) ([]*types.VirtualDisk, error) {
+	defer trace.End(trace.Begin(vm.String()))
+
+	op.Debugf("Looking for all attached disks")
+
+	candidates, err := findDisk(op, vm, func(diskName string) bool {
```
vburenin: why candidates? I think naming it as "disks"  would be better. These are no longer candidates since you pre-filtered them already.

pkg/vsphere/disk/disk_manager.go
fdawg4l: Ya know... that really is idiotic.  Thanks for pointing it out.  I have no idea why it exists.  Will fix.

pkg/vsphere/disk/disk_manager.go
fdawg4l: I'm not sure what the right path is;  exit early like the code does or exit later after trying all options.  

But, I think you may be right.  Detatching late is probably what the caller wants.  I'll fix this.  Thanks!

pkg/vsphere/disk/util.go
fdawg4l: will do

pkg/vsphere/disk/disk_manager.go
```diff
@@ -275,24 +281,38 @@ func (m *Manager) Detach(op trace.Operation, d *VirtualDisk) error {
...
 	if err != nil {
 		return errors.Trace(err)
 	}
 
+	if err = m.detach(op, disk); err != nil {
+		op.Errorf("detach for %s failed with %s", d.DevicePath, errors.ErrorStack(err))
+		return errors.Trace(err)
+	}
+
+	select {
```
vburenin: I wander why do you even need this select. are you trying to drain the channel?
if len(m.maxAttached) > 0 {
 <- m.maxAttached
}

It is MUCH MORE lightweight comparable to a select. Go Select implementation is around 600 lines. The only thing that does't work in my example is that it is not concurrent safe. you potentially may block forever if two calls happen simultaneously. I leave it up to you to decide.

fdawg4l: Blocking forever isn't acceptable. I'd rather code defensively in that instance. To be fair, there should never be an instance where the channel can't be drained because we have a lock around the push.

tests/test-cases/Group10-VCH-Restart/10-01-VCH-Restart.robot
```diff
@@ -113,3 +113,14 @@ Created Network And Images Persists As Well As Containers Are Discovered With Co
...
+Create VCH attach disk and reboot
+    ${rc}=  Run And Return Rc  govc vm.disk.create -vm=%{VCH-NAME} -name=%{VCH-NAME}/deleteme -size "16M"
+    Should Be Equal As Integers  ${rc}  0
+
+    Reboot VCH
+
+    # wait for docker info to succeed
+    Wait Until Keyword Succeeds  20x  5 seconds  Run Docker Info  %{VCH-PARAMS}
+    ${rc}=  Run And Return Rc  govc device.ls -vm=%{VCH-NAME} | grep disk
+    Should Be Equal As Integers  ${rc}  1
```
fdawg4l: When `grep` finds no matches, it returns 1.

```
$ echo foo | grep bar &> /dev/null
$ echo $?
1

```

PR #3774, opened 2017-01-31
[specific ci=13-01-VAC-GuestFullName] Design and implementation of VAC integration

### LINKED ISSUE #3645: Phone home: determine permissible approaches
We have some constraints around using the current phonehome libraries in OSS. We need to follow up with the PhoneHome team and determine exactly what approaches are available to us at this time and in the next few months.

Output:
- [ ] Initial design doc for phone home integration detailing approaches and areas of concern if any

sflxn:
This is the design proposal for initial integration with Vmware
Analytics Cloud.  There is also an initial design for future
telemetry data collection.

Resolves #3645

### DIFF
--- a/lib/guest/linux.go
+++ b/lib/guest/linux.go
@@ -21,14 +21,13 @@ import (
 
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/lib/portlayer/constants"
 	"github.com/vmware/vic/lib/spec"
 	"github.com/vmware/vic/pkg/vsphere/session"
 	"github.com/vmware/vic/pkg/vsphere/sys"
 )
 
 const (
-	linuxGuestID = "other3xLinux64Guest"
-
 	scsiBusNumber = 0
 	scsiKey       = 100
 	ideKey        = 200
@@ -69,7 +68,8 @@ func NewLinuxGuest(ctx context.Context, session *session.Session, config *spec.V
 	s.AddVirtualCdrom(cdrom)
 
 	// Set the guest id
-	s.GuestId = linuxGuestID
+	s.GuestId = string(types.VirtualMachineGuestOsIdentifierOtherGuest64)
+	s.AlternateGuestName = constants.DefaultAltContainerGuestName()
 
 	return &LinuxGuestType{
 		VirtualMachineConfigSpec: s,
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -298,7 +298,8 @@ func (d *Dispatcher) createApplianceSpec(conf *config.VirtualContainerHostConfig
 	spec := &spec.VirtualMachineConfigSpec{
 		VirtualMachineConfigSpec: &types.VirtualMachineConfigSpec{
 			Name:     conf.Name,
-			GuestId:  "other3xLinux64Guest",
+			GuestId:  string(types.VirtualMachineGuestOsIdentifierOtherGuest64),
+			AlternateGuestName: constants.DefaultAltVCHGuestName(),
 			Files:    &types.VirtualMachineFileInfo{VmPathName: fmt.Sprintf("[%s]", conf.ImageStores[0].Host)},
 			NumCPUs:  int32(vConf.ApplianceSize.CPU.Limit),
 			MemoryMB: vConf.ApplianceSize.Memory.Limit,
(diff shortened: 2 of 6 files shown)


vburenin:
Is it in scope of this document to provide a list of initial metrics we would like to collect?

1. Executed docker commands and parameters. Execution time.
2. Error, panic, warnings messages.
3. Memory usage, CPU usage, disk usage.
4. List of used containers and repositories. Private repository maybe obscured by using word "private"
5. VCH Uptime.

For metrics purposes I would consider developing a separate daemon that will be only responsible for metric collection and sending them out. Thus if user disagrees on sending usage data we may just simply disable metrics daemon. However, this route has own disadvantages as well as advantages.

doc/design/proposals/telemetry.md
```diff
@@ -0,0 +1,20 @@
...
+
+In future updates, VIC will be updated to collect data on which features are most commonly used to guide Vmware development plans.
+
+## Design Proposal
+
+vSphere collects a VM's OS full name by querying this data from the guest tools running inside the VM.  VIC installs a minimal toolbox inside the VCH's appliance VM and all container VMs created by VIC.  The toolbox will be updated to provide the following OS name for both VMs:
+
+| VM Type | OS Full Name |
+| --- | --- |
+| VCH | Photon OS - VCH v.v.v |
```
hickeng: Don't think there's any need to have PhotonOS in here given we are deploying an appliance. We could/?should? name it inline with the docs:
`VIC endpoint v.v.v`
Same for the containerVM OS:
`VIC containerVM v.v.v`

doc/design/proposals/telemetry.md
```diff
@@ -0,0 +1,20 @@
...
+| container VM | Photon OS - Pico v.v.v |
+
+In the above, v.v.v represents the actual version.  These strings can either added to the toolbox as part of the build process, or they can be added to the VM's guestinfo, which toolbox can read.  These strings will help differentiate the VCH from the container VM.  It will also provide Vmware with a distribution of VIC versions install base.
+
+For future feature-usage based data collection, VIC will need to provide another mechanism to collect data.  The initial proposal for this near future work is to divide this problem into a collection problem and a data upload problem.  The collection can be solved by updating the personality server and the portlayer servers to track feature usage metrics and saving it in the VM's guestinfo.  The data upload can be performed by vicAdmin.  The Vmware Analytics Cloud collector within vSphere will need to have it's data schema updated for these new data fields.
```
hickeng: It's likely that we'll be able to provide an alternative mechanism for persisting logs and metrics in the medium term - just FYI.

hickeng:
Please add a section on testing, and verification/acceptance criteria.

doc/design/proposals/telemetry.md
sflxn: This was the initial name proposed by Karthik.  We can certainly change the name.

sflxn:
@mdubya66 I thought this is what we agreed on in Austin.  Design docs from design sprints go into github.  Where do we want to put these?

mdubya66:
what's wrong with https://github.com/vmware/vic/tree/master/doc/design

sflxn:
Moved the file to a new location and added info on the robot script needed to test.

mlh78750:
I'd like to see all version related strings defined in one place in the tree and then consumed everywhere else. If such a place already exists, please move the ID strings to that location and consume them here. If it does not yet exist, you don't need to change it resolve this issue. However, if you don't handle it here, let's open an issue to consolidate all our versions (or just let me know and I'll do it). 

LGTM

lib/guest/linux.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	linuxGuestID = "other3xLinux64Guest"
+	linuxAlternateGuestName = "vSphere Integrated Containers - Container 0.9.0"
```
emlin: we have a version definition, which is dynamically bound with release, build, commit, etc. https://github.com/vmware/vic/blob/master/pkg/version/version.go

lib/install/management/appliance.go
```diff
@@ -298,7 +298,8 @@ func (d *Dispatcher) createApplianceSpec(conf *config.VirtualContainerHostConfig
 	spec := &spec.VirtualMachineConfigSpec{
 		VirtualMachineConfigSpec: &types.VirtualMachineConfigSpec{
 			Name:     conf.Name,
-			GuestId:  "other3xLinux64Guest",
+			GuestId:  string(types.VirtualMachineGuestOsIdentifierOtherGuest64),
+			AlternateGuestName: "vSphere Integrated Containers - VCH 0.9.0",
```
emlin: ^^^ I think we should not hard code the version number.
In addition, besides of the release version, should we add more details, like build number? Which can make the telemetry more accurate.

lib/guest/linux.go
sflxn: Whoops, I did not mean to include this test code.  Though, it's a really simple implementation.  I might as well just include the code and tests with this PR.

PR #3786, opened 2017-01-31
NFS based volume design spec [skip ci]

### LINKED ISSUE #3622: Research Spike: NFS Shared Volumes
Faiyaz and I need to go over what the requirements for NFS Shared Volumes are. We need to do the following.

- [ ] Look at how customers would like to use shared volumes and verify that their expectations can be met. E.g.  explore docker's implementation and usage model and how we can make it fit ours.
- [ ] Decide on how it will be implemented by coming up with an interface and straw-man implementation.
  - [ ] Identify failure modes and how they will be handled
  - [ ] Specify how the NFS volume stores are described via vic-machine and the necessary validation steps
- [ ] Design integration tests and, if possible, unit tests to prove that the implementation will work.

Open questions
  Do we want to address #3482?  That is, do we want to address mounting a volume in a container's directory when that directory has data on it.  In docker, the contents of the dir gets copied to the volume.  But any change to support this on NFS will need to be supported with vmdk backed volumes as well, so the issue and implementation can be kept out of this spike.

fdawg4l:
Fixes #3622


doc/design/volumes/nfsvolumestores.md
```diff
@@ -0,0 +1,106 @@
...
+ target *url.URL
+}
+```
+
+_The implementation is still being worked on.  The open question is whether the VCH appliance will mount the target to manipulate the NFS target, or use an NFS client implementation in userspace instead.  I'd much (*MUCH*) rather do the latter.  The `linux` VFS implementation throws `sync` errors when mounts are unavailable.  And we don't want to bring down the appliance because of a network hiccup.  NFS is a simple protocol and there is a public pkg which implements most of it.  Adding the few primitives we need shouldn't be that difficult, but more evaluation of the work required is needed_
+
+#### VolumeCreate
+In the vsphere model, a volume is a `.vmdk` backed block device.  Creation of a volume entails attaching a new disk to the VCH, preparing it with a filesystem, and detaching it.  The resulting `.vmdk` lives in its own folder in the volume store directory (specified during install w/ `vic-machine`).  We're going to follow the same model except there is nothing to prepare.  Each volume will be a directory (which the container client will mount directly) and live at the top of the volume store directory (which we will prepare during install).
```
hickeng: could note that VolumeCreate requires creation of directory via NFS client, and how any volume metadata (if any) gets stored.

doc/design/volumes/nfsvolumestores.md
```diff
@@ -0,0 +1,106 @@
...
+// mkdir volPath
+// return volPath
+}
+```
+#### VolumeDestroy
+Likewise destroying the volume is simply removing the volume's top level directory.
+```
+func VolumeDestroy() {
+// volPath := vicVolumePath(nameOfVolume)
+// rm -rf volPath
```
hickeng: Is recursive delete implemented on the server side or does the client need to iterate and delete each file in turn?

doc/design/volumes/nfsvolumestores.md
```diff
@@ -0,0 +1,106 @@
...
+    * Answer:  I don't see a reason why we can't support this.
+ 1. Is there any mechanism by which we can indicate available space in the volumestore? Is this necessary data for functional usage.
+    * Answer: See Non-requirement 3
+ 1. Should we allow for read-only volume store? - e.g. publishing datasets for consumption
+    * Answer: Needs investigation.  What is RO here (the target or the directory) and what would the container user want to see or expect when such a target was used?
+ 1. Failure handling;  what do we do if a mount is unavailable, does the container go down?
+    * Answer:  Needs investigation.  We're relying on the kernel nfs client in the container to handle failures to the target.  There is little we can do during run-time, but we can check availability during container create at a minimum.
+ 1. NFS target mount options-  How do you pass a `uid`/`gid` or `nosquash` mapping?  How do you map `uid`/`gid` to the container at all?
+    * Answer:  Requirement and usage needs to be thought through at a higher level.  Mapping of users into containers and mapping of credentials into containers need to be solved in the system as a whole.
```
hickeng: iirc it's possible to pass raw mount options through at the moment when you specify `-v source:dest,opts`, but this is something that should really be hidden from the container user.

doc/design/volumes/nfsvolumestores.md
fdawg4l: It's psuedo code, but the client descends the dirents normally.  We'll supply a call to do that in our client.

doc/design/volumes/nfsvolumestores.md
fdawg4l: Agreed.  Will add this.

Issue #1676, opened 2016-07-26
enable filter for docker ps


cgtexmex:
The docker client allows a users to provide filter options for the ps command.  Details [here](https://docs.docker.com/engine/reference/commandline/ps/)

vic will need to provide parity with this functionality.

mdubya66:
adding to consideration in 0.9.0

mhagen-vmware:
lgtm

cgtexmex:
#3788

PR #3788, opened 2017-01-31
implement docker ps filtering [specific ci=1-10-Docker-PS]

### LINKED ISSUE #1545: Docker ps -l and ps -n=2 does not show containers that have been created but not started
docker create busybox ls
docker ps -l

Actual output:

```
anon@anon-Latitude-E7450:~/vic$ docker -H 192.168.254.172:2376 --tls create busybox ls
a3ed95caeb02: Pull complete 
8ddc19f16526: Pull complete 
9d928c7aad5547269313da1e17d92a4ccd06696566f6d22eb24020933b3d996c
anon@anon-Latitude-E7450:~/vic$ docker -H 192.168.254.172:2376 --tls ps -l
```

Expected output:

```
anon@anon-Latitude-E7450:~/vic$ docker create busybox ls
115184175ef93607b77dda774445bb5cb57b0bd31ff2f33bcf97f4374c97a24d
anon@anon-Latitude-E7450:~/vic$ docker ps -l
115184175ef9        busybox             "ls"                5 seconds ago       Created                                 admiring_minsky
```

cgtexmex:
The following docker ps filter options have been
implemented:
	* before, since, id, exited, label, name, status
        * network

If a filter option isn't supported by vic an error message
will be returned to the user.  Examples:
	* ancestor, isolation, health, is-task

The docker ps flags of --latest and --last have also been
implemented.

Known issue with network filtering: it only works with the
network name (bridge, etc).  network id will be added later.

Fixes #1545, #1676

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -20,7 +20,6 @@ import (
 	"io"
 	"math/rand"
 	"net"
-	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -46,6 +45,7 @@ import (
 
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	viccontainer "github.com/vmware/vic/lib/apiservers/engine/backends/container"
+	"github.com/vmware/vic/lib/apiservers/engine/backends/filter"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/portmap"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/containers"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/interaction"
@@ -56,6 +56,32 @@ import (
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/uid"
 )
+
+// valid filters as of docker commit 49bf474
+var acceptedPsFilterTags = map[string]bool{
+	"ancestor":  true,
+	"before":    true,
+	"exited":    true,
+	"id":        true,
+	"isolation": true,
+	"label":     true,
+	"name":      true,
+	"status":    true,
+	"health":    true,
+	"since":     true,
+	"volume":    true,
+	"network":   true,
+	"is-task":   true,
+}
+
+// currently not supported by vic
+var unSupportedPsFilters = map[string]bool{
+	"ancestor":  false,
+	"health":    false,
+	"isolation": false,
+	"is-task":   false,
+	"volume":    false,
+}
 
 const (
 	bridgeIfaceName = "bridge"
@@ -971,11 +997,6 @@ func (c *Container) ContainerInspect(name string, size bool, version version.Ver
 	}
 
 	log.Debugf("ContainerInspect json config = %+v\n", inspectJSON.Config)
-	if inspectJSON.NetworkSettings != nil {
-		log.Debugf("Docker inspect - network settings = %#v", inspectJSON.NetworkSettings)
-	} else {
-		log.Debugf("Docker inspect - network settings = null")
-	}
 
 	return inspectJSON, nil
 }
@@ -1032,11 +1053,18 @@ func (c *Container) ContainerTop(name string, psArgs string) (*types.ContainerPr
 
 // Containers returns the list of containers to show given the user's filtering.
 func (c *Container) Containers(config *types.ContainerListOptions) ([]*types.Container, error) {
+	defer trace.End(trace.Begin(fmt.Sprintf("ListOptions %#v", config)))
+
+	// validate filters for support and validity
+	listContext, err := filter.ValidateContainerFilters(config, acceptedPsFilterTags, unSupportedPsFilters)
+	if err != nil {
+		return nil, err
+	}
 
 	// Get an API client to the portlayer
 	client := c.containerProxy.Client()
 
-	containme, err := client.Containers.GetContainerList(containers.NewGetContainerListParamsWithContext(ctx).WithAll(&config.All))
+	containme, err := client.Containers.GetContainerList(containers.NewGetContainerListParamsWithContext(ctx).WithAll(&listContext.All))
 	if err != nil {
 		switch err := err.(type) {
 
@@ -1050,14 +1078,8 @@ func (c *Container) Containers(config *types.ContainerListOptions) ([]*types.Con
 	// TODO: move to conversion function
 	containers := make([]*types.Container, 0, len(containme.Payload))
 
+payloadLoop:
 	for _, t := range containme.Payload {
-		cmd := strings.Join(t.ProcessConfig.ExecArgs, " ")
-		// the docker client expects the friendly name to be prefixed
-		// with a forward slash -- create a new slice and add here
-		names := make([]string, 0, len(t.ContainerConfig.Names))
-		for i := range t.ContainerConfig.Names {
-			names = append(names, clientFriendlyContainerName(t.ContainerConfig.Names[i]))
-		}
 		var started time.Time
 		var stopped time.Time
 
@@ -1070,12 +1092,46 @@ func (c *Container) Containers(config *types.ContainerListOptions) ([]*types.Con
 		}
 
 		// get the docker friendly status
-		_, status := dockerStatus(
+		exitCode, status := dockerStatus(
 			int(t.ProcessConfig.ExitCode),
 			t.ProcessConfig.Status,
 			t.ContainerConfig.State,
 			started,
 			stopped)
+
+		// labels func requires a config be passed
+		// TODO: refactor labelsFromAnnotations func for broader use
+		tempConfig := &containertypes.Config{}
+		err = labelsFromAnnotations(tempConfig, t.ContainerConfig.Annotations)
+		if err != nil && config.Filter.Include("label") {
+			return nil, fmt.Errorf("unable to convert vic annotations to docker labels (%s)", t.ContainerConfig.ContainerID)
+		}
+
+		listContext.Labels = tempConfig.Labels
+		listContext.ExitCode = exitCode
+		listContext.ID = t.ContainerConfig.ContainerID
+
+		// prior to further conversion lets determine if this container
+		// is needed or if the list is complete -- if the container is
+		// needed conversion will continue and the container will be added to the
+		// return array
+		action := filter.IncludeContainer(listContext, t)
+		switch action {
+		case filter.ExcludeAction:
+			// skip to next container
+			continue payloadLoop
+		case filter.StopAction:
+			// we're done
+			break payloadLoop
+		}
+
+		cmd := strings.Join(t.ProcessConfig.ExecArgs, " ")
+		// the docker client expects the friendly name to be prefixed
+		// with a forward slash -- create a new slice and add here
+		names := make([]string, 0, len(t.ContainerConfig.Names))
+		for i := range t.ContainerConfig.Names {
+			names = append(names, clientFriendlyContainerName(t.ContainerConfig.Names[i]))
+		}
 
 		ips, err := publicIPv4Addrs()
 		var ports []types.Port
@@ -1110,11 +1166,15 @@ func (c *Container) Containers(config *types.ContainerListOptions) ([]*types.Con
 			Command: cmd,
 			SizeRw:  t.ContainerConfig.StorageSize,
 			Ports:   ports,
+			State:   filter.DockerState(t.ContainerConfig.State),
 		}
+
+		// The container should be included in the list
 		containers = append(containers, c)
+		listContext.Counter++
+
 	}
-	// sort on creation time
-	sort.Sort(sort.Reverse(containerByCreated(containers)))
+
 	return containers, nil
 }
 
--- a/lib/apiservers/engine/backends/filter/container.go
+++ b/lib/apiservers/engine/backends/filter/container.go
@@ -1,0 +1,227 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package filter
+
+import (
+	"fmt"
+	"strconv"
+
+	"github.com/docker/engine-api/types"
+
+	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
+	"github.com/vmware/vic/lib/apiservers/portlayer/models"
+)
+
+// reused from docker/docker/daemon/list.go
+type ContainerListContext struct {
+	FilterContext
+
+	// Counter is the container iteration index for this context
+	Counter int
+	// ExitCode for the passed container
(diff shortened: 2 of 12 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -57,6 +57,32 @@ import (
 	"github.com/vmware/vic/pkg/uid"
 )
 
+// valid filters as of docker commit 49bf474
+var acceptedPsFilterTags = map[string]bool{
```
vburenin: For maps like this, I would use map[string]struct{}
struct{} doesn't take any space in memory.
For lookup purposes:
```
if _, ok := map[val]; ok {
    ...
}
```

lib/apiservers/engine/backends/container.go
```diff
@@ -1070,13 +1092,47 @@ func (c *Container) Containers(config *types.ContainerListOptions) ([]*types.Con
...
+
+		// prior to further conversion lets determine if this container
+		// is needed or if the list is complete -- if the container is
+		// needed conversion will continue and the container will be added to the
+		// return array
+		action := filter.IncludeContainer(listContext, t)
+		switch action {
+		case filter.Exclude:
+			// skip to next container
+			continue payloadLoop
```
vburenin: You do not need label here. but it's okay for readability.

lib/apiservers/engine/backends/filter/container.go
```diff
@@ -0,0 +1,230 @@
...
+*		* Include
+*		* Exclude
+*		* Stop
+*
+ */
+func IncludeContainer(listContext *ContainerListContext, container *models.ContainerInfo) FilterAction {
+
+	// filter common requirements
+	act := filterCommon(&listContext.FilterContext)
+	if act != Include {
```
vburenin: "Include" and relatives need a different naming. IncludeFilter or something like that. At this point I spent some time to see what that Include is.

lib/apiservers/engine/backends/container.go
cgtexmex: The bool is a feature and not just an empty space...if false then it will be rejected as an invalid filter.

lib/apiservers/engine/backends/filter/container.go
```diff
@@ -0,0 +1,230 @@
...
+			filters: containerFilters},
+		exitAllowed:          filtExited,
+		ContainerListOptions: options,
+	}
+
+	err = containerFilters.WalkValues("before", func(value string) error {
+		var err error
+		before := cache.ContainerCache().GetContainer(value)
+		if before == nil {
+			err = fmt.Errorf("No such container: %s", value)
```
vburenin: I would return error right from here and nil at the end. but it is up to you

lib/apiservers/engine/backends/filter/container.go
```diff
@@ -0,0 +1,230 @@
...
+	case "Stopped":
+		state = "exited"
+	case "Running":
+		state = "running"
+	case "Created":
+		state = "created"
+	default:
+		// not sure what to do, so just return
+		// what was given
+		state = containerState
```
vburenin: in this case you may want to lowercase it to be consistent.

cgtexmex: I'd rather not -- we know it doesn't match a docker state, so I'd rather leave it as it.

lib/apiservers/engine/backends/filter/container.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	return state
+}
+
+// IsValidDockerState will verify the provided state is
+// a valid docker container state
+func IsValidDockerState(s string) bool {
```
vburenin: return s == "paused" || s == "restarting" || s == "removing" || s == "running" || s == "dead" || s == "created" || s == "exited"

lib/apiservers/engine/backends/container.go
```diff
@@ -1032,11 +1053,18 @@ func (c *Container) ContainerTop(name string, psArgs string) (*types.ContainerPr
 
 // Containers returns the list of containers to show given the user's filtering.
 func (c *Container) Containers(config *types.ContainerListOptions) ([]*types.Container, error) {
+	defer trace.End(trace.Begin(fmt.Sprintf("ListOptions %#v", config)))
```
sflxn: @cgtexmex This function is pretty big and complex.  Can we break it up into some util functions?

cgtexmex: we can -- I'd propose that we have a large discussion about it.  Specifically identifying a common approach for the utility functions.  For example we are converting to docker specific formats in many cases...that might be a struct, string, etc..I'd think we can centralize those in a package etc...

So how about we create an issue for cleaning up the persona -- we could do a research spike and then nail down tasks / deliverables -- thoughts?

lib/apiservers/engine/backends/filter/container.go
```diff
@@ -0,0 +1,230 @@
...
+	if listContext.Limit > 0 && listContext.Counter == listContext.Limit {
+		return Stop
+	}
+
+	// Do not include container if its exit code is not in the filter
+	if len(listContext.exitAllowed) > 0 {
+		shouldSkip := true
+		for _, code := range listContext.exitAllowed {
+			// note "Running" state is congruent with PortLayer and not docker
+			if code == listContext.ExitCode && container.ContainerConfig.State != "Running" && container.ProcessConfig.StartTime > 0 {
```
sflxn: I think we should normalize the state string to lower case and do a the check.  I know we haven't been doing this, we should really change.

cgtexmex: I agree -- but think it should be done with a persona clean up (my other comment).  We can change it here, but I certainly wouldn't want to do a broad change without really investigating it and I don't think we should hold up this change for that investigation.

lib/apiservers/engine/backends/container.go
sflxn: Yeah, I was thinking the same thing.  We may not need an entire research spike.  Maybe we can keep a list of items.  One of the items we can do is to split the code up into functional files like Docker.  That should help with multiple hands in the cookie jar, but that brings up the question of benefits since we don't know when the containerd work will be done.

sflxn:
lgtm

lib/apiservers/engine/backends/container.go
vburenin: if so, why all the values are true?

lib/apiservers/engine/backends/container.go
```diff
@@ -1070,13 +1092,47 @@ func (c *Container) Containers(config *types.ContainerListOptions) ([]*types.Con
...
+		case filter.Stop:
+			// we're done
+			break payloadLoop
+		}
+
+		cmd := strings.Join(t.ProcessConfig.ExecArgs, " ")
+		// the docker client expects the friendly name to be prefixed
+		// with a forward slash -- create a new slice and add here
+		names := make([]string, 0, len(t.ContainerConfig.Names))
+		for i := range t.ContainerConfig.Names {
```
vburenin:
```
for _, v := range t.ContainerConfig.Names {
    names = append(names, clientFriendlyContainerName(v))
} 
```

lib/apiservers/engine/backends/filter/container.go
vburenin: IncludeAction?

cgtexmex: had it like that in an earlier iteration, but since the type is FilterAction I see the Action suffix as too verbose.....I'm fine with it and will change it to include the Action suffix

lib/apiservers/engine/backends/container.go
cgtexmex: I used the index to avoid an implicit copy of the value

lib/apiservers/engine/backends/container.go
cgtexmex: There are two maps -- `acceptedPsFilterTags` that is what docker supports per the commit shown.  So they are all true because they are all supported by docker.  I have another that's `unSupportedPsFilters` those are not supported by vic and are all false.  As part of the filter validation we use docker code, so these maps follow their design.

lib/apiservers/engine/backends/filter/container.go
```diff
@@ -0,0 +1,230 @@
...
+
+	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
+	"github.com/vmware/vic/lib/apiservers/portlayer/models"
+)
+
+// reused from docker/docker/daemon/list.go
+type ContainerListContext struct {
+	FilterContext
+
+	// idx is the container iteration index for this context
```
jzt: minor: s/idx/Counter

cgtexmex: done

lib/apiservers/engine/backends/container.go
```diff
@@ -1070,13 +1092,47 @@ func (c *Container) Containers(config *types.ContainerListOptions) ([]*types.Con
...
+		// needed conversion will continue and the container will be added to the
+		// return array
+		action := filter.IncludeContainer(listContext, t)
+		switch action {
+		case filter.ExcludeAction:
+			// skip to next container
+			continue payloadLoop
+		case filter.StopAction:
+			// we're done
+			break payloadLoop
```
hickeng: Is the `payloadLoop` label for clarity or a language requirement? We're not in a nested loop at this point

lib/apiservers/engine/backends/filter/container.go
```diff
@@ -0,0 +1,230 @@
...
+	Counter int
+	// ExitCode for the passed container
+	ExitCode int
+	// exitAllowed is a list of exit codes allowed to filter with
+	exitAllowed []int
```
hickeng: how is this configured? I see that we're iterating over the array each time during the processing of this filter to determine if it's permitted - is `map[int]struct` a better option?

lib/apiservers/engine/backends/filter/container.go
```diff
@@ -0,0 +1,230 @@
...
+		// TODO: this is the incorrect ID
+		id := container.Endpoints[n].ID
+		if listContext.filters.Include("network") {
+			err := listContext.filters.WalkValues("network", func(value string) error {
+				if name == value || id == value {
+					return networkExist
+				}
+				return nil
+			})
+			if err != networkExist {
```
hickeng: This isn't testing error message equality - think it should be:
`err.Error() != "container part of network"`

https://play.golang.org/p/SpMdRpk1X7

lib/apiservers/engine/backends/filter/container.go
```diff
@@ -0,0 +1,230 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	err = containerFilters.WalkValues("since", func(value string) error {
+		var err error
+		since := cache.ContainerCache().GetContainer(value)
+		if since == nil {
+			err = fmt.Errorf("No such container: %s", value)
```
hickeng: should be kept consistent with the `before` handling if that's modified.

lib/apiservers/engine/backends/container.go
cgtexmex: The break w/o label will only break out of the switch and not the loop, thus the label

lib/apiservers/engine/backends/filter/container.go
cgtexmex: yeah...we could do that...

lib/apiservers/engine/backends/filter/container.go
cgtexmex: I don't follow...please expound

lib/apiservers/engine/backends/filter/container.go
cgtexmex: Interesting...this was (as a number of parts here are) leveraged docker code...it "worked" so didn't give it any real thought...I'll change

lib/apiservers/engine/backends/filter/container.go
cgtexmex: ahhh..._if_ it's modified...I'm wasn't planning on modifying...

lib/apiservers/engine/backends/container.go
vburenin: not sure what you mean. that is a function call, copy always happens during that call, you are also not using goroutines, thus v is always a value you expect.

lib/apiservers/engine/backends/filter/container.go
vburenin: @hickeng if exitAllowed is a small array, it is always preferable to use iteration instead of a map. the max size need to be benchmarked to find that edge.

lib/apiservers/engine/backends/filter/container.go
cgtexmex: Dug into this more -- the code is misleading -- it's not really concerned about error equality it's concerned about nil vs. !nil -- I'm updating the code to make that obvious.  Thanks for pointing it out.

lib/apiservers/engine/backends/filter/container.go
cgtexmex: @vburenin was having that thought last night as well...exitAllowed will be small...most likely less than 5 items

lib/apiservers/engine/backends/container.go
cgtexmex: Names is not a func, it's a string array attribute -- so we are iterating the array.

lib/apiservers/engine/backends/container.go
hickeng: why is there no _facepalm_ emoji ....

lib/apiservers/engine/backends/filter/container.go
```diff
@@ -0,0 +1,228 @@
...
+		return act
+	}
+
+	// Stop iteration when the index is over the limit
+	if listContext.Limit > 0 && listContext.Counter == listContext.Limit {
+		return StopAction
+	}
+
+	// Do not include container if its exit code is not in the filter
+	if len(listContext.exitAllowed) > 0 {
```
hickeng: I don't think the map adds clarity as it stands; I had expected the entire clause to be simplified if switching to a map:
```go
_, ok := listContext.exitAllowed[listContext.ExitCode]
if !ok || container.ProcessConfig.ExitCode != listContext.ExitCode || container.container.ContainerConfig.State == "Running" || container.ProcessConfig.StartTime == 0 {
	return ExcludeAction
}
```

If keeping it as an array, reversing the order of the tests avoids the iteration if not necessary:
```go
if len(listContext.exitAllowed) > 0 {
   // is container in valid state for exitcode check
    if container.ContainerConfig.State == "Running" || container.ProcessConfig.StartTime == 0 {
	return ExcludeAction
    }

    ok := false
    for _, code := range exitAllowed {
        if code == listContext.ExitCode {
            ok = true
            break
        }
    }
    if !ok {
        return ExcludeAction
    }
}
```

lib/apiservers/engine/backends/container.go
vburenin: I probably phrased it in a not clear way. clientFriendlyContainerName is a function, you create a copy here when you call this function. I also do not see any reason to take value directly from loop instead of access it each time by index.

lib/apiservers/engine/backends/filter/container.go
cgtexmex: I've left the map but updated the flow more in line w/your example...let me know if that satisfies your clarity concern.

lib/apiservers/engine/backends/container.go
cgtexmex: ok..understand now the question better...

I would think that if I don't use the index then a copy will be made on each iteration of the loop...then since the func takes a value and not a pointer a copy would be made for the func call.  So two copies instead of 1...possibly I'm wrong...

Again my natural tendency is to use the index when iterating an array to avoid the implicit copy....for now I'm leaving it..

PR #3804, opened 2017-02-01
Nfsspecnetwork

fdawg4l:
There are still some open issues regarding how to support VolumeStores on specific networks but they have network dependencies.

### DIFF
--- a/doc/design/volumes/nfsvolumestores.md
+++ b/doc/design/volumes/nfsvolumestores.md
@@ -6,8 +6,11 @@ Container users want to be able to access shared storage between their container
 
 Allow the VI admin to
  1. add an NFSv3 based `VolumeStore`
+ 2. specify a container network the `VolumeStore` can be found on
 
 Allow the container user to
+ 1. specify one or many NFS based `VolumeStore`s to create the container volumes on
+ 1. have network connectivity from created containers to the NFS targets
  1. create volumes on the NFS based `VolumeStore`
  1. create 1 or greater containers with NFS based volumes at the given location in the container filesystem namespace
  1. validate the volume is no longer in use and delete it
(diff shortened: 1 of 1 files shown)


doc/design/volumes/nfsvolumestores.md
```diff
@@ -18,10 +21,36 @@ Allow the container user to
...
 ### Implementation
 
 Adding shared storage to our model fits with the `VolumeStore` interface.  At install, a VI admin can specify an NFS target as a `VolumeStore` (potentially) using a `nfs://host/<path>` URI with a volume store name.  And provided the VCH has access to the network the target is on, the container user only needs to pass the volume store name as one of the `volume create` driver opts to create a volume which will be backed by this shared storage target.  Then many containers can be created with the specified volume attached.
 
+#### Runtime network connectivity validation
+We need to inform the user when a container is being created without the appropriate network required to get connectivity to the NFS target.  We're planning to make changes to the personality such that when the container is created, it check if the specified container network matches the container network specified when the `VolumeStore` is created.
+
+This is just an example, but we could do the following:
+```
+$ vic-machine-linux create --volume-store volumeStoreName/nfs://host/path?<nfs container network name>
```
hmahmood: Do we really need to explicitly associate the volume store with a container network? The vic user should make sure that the container is on the right network for the NFS store to successfully connect. Can we find that out during container init (in tether) and communicate this back during docker start/run?

doc/design/volumes/nfsvolumestores.md
```diff
@@ -18,10 +21,36 @@ Allow the container user to
  2. Exposing shared storage configuration via VIC (e.g. IOPS per client, storage policy, etc.)
  3. Management of shared storage via VIC (e.g. container quiesce for storage maintenance, quota manipulation of the target, etc.)
 
+### Networking
+
+Containers will need to be on an appropriate network for the NFS volume store to be accessible.
+
+There are ways this could be done:
+ - allow association of volume-store with container-network to allow direct coupling
```
hmahmood: I would prefer if the vic user is responsible for tracking this. All that should be required is that the container using the nfs store is connected to the correct network(s). We can validate this during container/tether bring up and communicate it to the caller.

hmahmood: I would prefer if the vic user is responsible for tracking this. All that should be required is that the container using the nfs store is connected to the correct network(s). We can validate this during container/tether bring up and communicate it to the caller.

fdawg4l: Well, chatting with @hickeng on this, the reason we allow the user to pass volume information by name and not by _host:mountpoint_ is to hide the details of the underlying infrastructure.  The user shouldn't need to know what the VI admin configured the VIC deployment with and at what addresses/hosts the NFS targets are available.  Rather, the `docker` user need only specify the `VolumeStore` name.  And in that respect, the user _won't_ know the network topology and can't be expected to know what network the volume store is listening on.  

For the short term, we can do what you're suggesting.  But the longer term is to do this configuration automagically with microsegmentation so the container is configured to talk to the NFS target by adding it to the appropriate network and defining the IP rules insulating the container from the NFS network and vice versa.

fdawg4l:
Uh...  I don't know how I got in this state but I apparently made this branch in the VIC repo and not my fork.  I'm closing this PR, creating a branch in my fork, nuking this branch in VIC, and re creating this PR.   Apologies, guys.

PR #3805, opened 2017-02-01
Fix string parsing logic in integration test script

anchal-agrawal:
Fixes the failure in https://ci.vmware.run/vmware/vic/7809 where environment variables aren't set correctly:

screen shot 2017-01-31 at 21 16 24

Robot's `Split String` command doesn't trim consecutive delimiters when a delimiter is specified. The above issue happens because `Feb` and `1` have two spaces in-between.


mhagen-vmware:
lgtm

PR #3819, opened 2017-02-01
Add networking section to the NFS spec

fdawg4l:
Continuation of https://github.com/vmware/vic/pull/3804


fdawg4l:
```

hmahmood an hour ago Member
I would prefer if the vic user is responsible for tracking this. All that should be required is that the container using the nfs store is connected to the correct network(s). We can validate this during container/tether bring up and communicate it to the caller.
￼  ￼  ￼￼
fdawg4l 42 minutes ago Member
Well, chatting with @hickeng on this, the reason we allow the user to pass volume information by name and not by host:mountpoint is to hide the details of the underlying infrastructure. The user shouldn't need to know what the VI admin configured the VIC deployment with and at what addresses/hosts the NFS targets are available. Rather, the docker user need only specify the VolumeStore name. And in that respect, the user won't know the network topology and can't be expected to know what network the volume store is listening on.

For the short term, we can do what you're suggesting. But the longer term is to do this configuration automagically with microsegmentation so the container is configured to talk to the NFS target by adding it to the appropriate network and defining the IP rules insulating the container from the NFS network and vice versa.
```

hmahmood:
@fdawg4l so I agree, yes, we need to hide the network/nfs details. But the docker user still has to specify the right `--net` argument, and for that to happen correctly they need to know on which network the nfs share is available on. So some level of knowledge of what the network provides is needed. I think we can validate that and provide feedback to the docker user without associating the volume store with the network configuration-wise.

I mostly disagree about the semantics of how this is done with vic-machine command line arguments. Do we actually need to know which network the nfs store is on to implement an nfs volume store? I am also afraid this would result in an overly complex command line syntax, and we already have enough of that with the `--container-network-*` options.

fdawg4l:
Just clarifying,

>   So some level of knowledge of what the network provides is needed. I think we can validate that and provide feedback to the docker user without associating the volume store with the network configuration-wise.

For now.  Ultimately we want to add the container to the network for the NFS target automatically *without* them having to specify it.  And that's because...

> Do we actually need to know which network the nfs store is on to implement an nfs volume store?

Yes.  NFS (well network attached storage in general) needs an endpoint (IP or name), auth creds (NOOP for nfsv3), and a target (the mount path on the server to get access to).  The `docker` user knows none of these things because the VI admin sets these up.

>  I am also afraid this would result in an overly complex command line syntax, and we already have enough of that with the --container-network-* options.

I agree, my example was complex.  It was just for illustration.  We can discuss/argue about the required flags and their usability later.  But, we require the 3 things i defined above (endpoint, creds, and target) to be specified in order for the container to attach the storage target.  One way or another, this needs to be passed via the `vic-machine` cli.

hmahmood:
@fdawg4l is endpoint == network though? I thought endpiont == ip/host name would suffice.

fdawg4l:
@hmahmood endpoint means ip/host in my example.  I guess I'm using the word wrong.  OK, so we need `{ip/host, creds, target}`.  But doesn't the need to get to a specific IP require a specific network capable of getting to that IP?  If I have an NFS target living on a network only reachable by 192.168.0.0/16, we would need to add the appropriate network to the system and add the container to it, right?

doc/design/volumes/nfsvolumestores.md
```diff
@@ -18,10 +21,36 @@ Allow the container user to
...
+```
+$ vic-machine-linux create --volume-store volumeStoreName/nfs://host/path?<nfs container network name>
+
+...
+%  docker %{VCH-PARAMS} run --name ${ContainerName} -d -v volumeStoreName:/mydata -net <the wrong container network>
+ ERROR:  volumeStoreName is on the <fs container network name> network.
+```
+
 #### VolumeStore
 The `VolumeStore` interface is used by the storage layer to implement the volume storage layer on different backend implementations.  The currently (and only) implementation used by VIC is to manimpulate vsphere `.vmdk` backed block devices on the Datastore.  We intend to create a similar implementation for NFSv3.
```
sgairo: "manimpulate" -> manipulate

hmahmood:
@fdawg4l the appropriate network would be added with the --conatiner-network-* options, yes. The docker user and VI admin need to communicate about which network (under docker network ls) hosts the nfs server; perhaps give it an obvious name, e.g. "nfsnet". But other than that I don't see how we can make it easier for the docker user to connect the container to the correct network for the nfs volume store to work. The actual connection to the nfs server can only happen on container start time; I think we should fail the run/start if we can't connect and report it back to the docker user. We cannot connect to the nfs server during vic-machine create time since we would have to connect the appliance VM to the right port group (involving adding yet another NIC), something that violates the networking model we have right now.

fdawg4l:
@hmahmood you make a point.  I'll ping @hickeng so we're all in agreement.  His requirement to me was the user need *not* know about the network topology or the target's actual network location and that the VCH should be able to resolve the dependency underneath the user without their direct involvement.

hickeng:
@fdawg4l requiring the user to add the container to a specific network to use a specific volume is viable as a first pass.

However I don't think we can consider the work complete until we can address this because of the knowledge requirement this imposes on the user:
* volumestore "x" depends on network "y" - this information would have to be communicated between vicadmin and user somehow.
* volume "a" is on volumestore "x" - this may require inspecting the volume if precursor knowledge of it's store isn't available
* container "m" is using volume "a" which has special constraints - not obvious if --volumes-from is used 

I'd rank fixing this knowledge issue below fixing the firewall requirement given there is a simple work around, and it's fail-fast, not fail-oh-s**t-why

fdawg4l:
@hickeng @hmahmood please rescan.  I'd like to close on this by COB Friday so please try to be as thorough as can be expected so if something needs to be addressed, we can do it with 1 round trip.  Thanks!

fdawg4l:
Poke @hickeng.

PR #3834, opened 2017-02-02
[specific ci=1-21-Docker-Volume-LS] Add support for volume ls filters

### LINKED ISSUE #1718: Implement volume ls filtering

anchal-agrawal:
Adds support for `volume ls -f` by `name`, `driver`, `dangling` and `label`.

Fixes #1718

### DIFF
--- a/lib/apiservers/engine/backends/volume.go
+++ b/lib/apiservers/engine/backends/volume.go
@@ -69,11 +69,12 @@ func NewVolumeModel(volume *models.VolumeResponse, labels map[string]string) *ty
 type Volume struct {
 }
 
-// acceptedVolumeFilters are filters that are supported by VIC
+// acceptedVolumeFilters are volume filters that are supported by VIC
 var acceptedVolumeFilters = map[string]bool{
 	"dangling": true,
 	"name":     true,
 	"driver":   true,
+	"label":    true,
 }
 
 var errPortlayerClient = fmt.Errorf("failed to get a portlayer client")
@@ -139,18 +140,20 @@ func (v *Volume) Volumes(filter string) ([]*types.Volume, []string, error) {
 	for _, vol := range volumeResponses {
 		log.Infof("%s", vol.Name)
 
+		volumeMetadata, err := extractDockerMetadata(vol.Metadata)
+		if err != nil {
+			return nil, nil, VolumeInternalServerError(fmt.Errorf("error unmarshalling docker metadata: %s", err))
+		}
+
 		// Set fields needed for filtering the output
 		volFilterContext.Name = vol.Name
 		volFilterContext.Driver = vol.Driver
 		_, volFilterContext.Joined = joinedVolumes[vol.Name]
+		volFilterContext.Labels = volumeMetadata.Labels
 
 		// Include the volume in the output if it meets the filtering criteria
 		filterAction := vicfilter.IncludeVolume(volumeFilters, volFilterContext)
 		if filterAction == vicfilter.IncludeAction {
-			volumeMetadata, err := extractDockerMetadata(vol.Metadata)
-			if err != nil {
-				return nil, nil, VolumeInternalServerError(fmt.Errorf("error unmarshalling docker metadata: %s", err))
-			}
 			volume := NewVolumeModel(vol, volumeMetadata.Labels)
 			volumes = append(volumes, volume)
 		}
--- a/lib/apiservers/engine/backends/volume.go
+++ b/lib/apiservers/engine/backends/volume.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -17,7 +17,6 @@ package backends
 import (
 	"encoding/json"
 	"fmt"
-	"net/http"
 
 	log "github.com/Sirupsen/logrus"
 	derr "github.com/docker/docker/errors"
@@ -27,9 +26,12 @@ import (
 	"strings"
 
 	"github.com/docker/engine-api/types"
+	"github.com/docker/engine-api/types/filters"
 	"github.com/docker/go-units"
 	"github.com/google/uuid"
 
+	vicfilter "github.com/vmware/vic/lib/apiservers/engine/backends/filter"
+	"github.com/vmware/vic/lib/apiservers/portlayer/client/containers"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/storage"
 	"github.com/vmware/vic/lib/apiservers/portlayer/models"
 	"github.com/vmware/vic/pkg/trace"
@@ -67,6 +69,15 @@ func NewVolumeModel(volume *models.VolumeResponse, labels map[string]string) *ty
 type Volume struct {
 }
 
+// acceptedVolumeFilters are filters that are supported by VIC
+var acceptedVolumeFilters = map[string]bool{
+	"dangling": true,
+	"name":     true,
+	"driver":   true,
+}
+
+var errPortlayerClient = fmt.Errorf("failed to get a portlayer client")
+
 // Volumes docker personality implementation for VIC
 func (v *Volume) Volumes(filter string) ([]*types.Volume, []string, error) {
 	defer trace.End(trace.Begin("Volume.Volumes"))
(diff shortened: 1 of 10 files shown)


cgtexmex:
I have a bit of rework concern w/this PR -- specifically my PR #3788 provides a common filtering mechanism for the persona.  There are common filters (name in the case of volume) that would be provided and no new code is required.  Additionally I'm introducing a filter package that will allow us to better segregate the filtering code from other code.  

If this PR is approved and merged then we'll need to refactor this filtering code into the new filter package.  We can do it that way, but it's churn on the code base that I don't think is really needed, but I'll leave that to others to decide.

anchal-agrawal:
Per my offline conversation with Clint, we'd like to wait for #3788 to be merged before this goes in. In the meanwhile, I'd appreciate reviews on my code's logic.

lib/apiservers/engine/backends/volume.go
```diff
@@ -96,12 +105,103 @@ func (v *Volume) Volumes(filter string) ([]*types.Volume, []string, error) {
...
+	if client == nil {
+		return nil, fmt.Errorf("failed to get a portlayer client")
+	}
+
+	all := true
+	cons, err := client.Containers.GetContainerList(containers.NewGetContainerListParamsWithContext(ctx).WithAll(&all))
+	if err != nil {
+		switch err := err.(type) {
+		case *containers.GetContainerListInternalServerError:
+			return nil, fmt.Errorf("error invoking GetContainerList: %s", err.Payload.Message)
```
fdawg4l: Not sure why we're doing this.  GetCOntainerListInt... is already an error.  Do we need to make it a generic error?

lib/apiservers/engine/backends/volume.go
```diff
@@ -96,12 +105,103 @@ func (v *Volume) Volumes(filter string) ([]*types.Volume, []string, error) {
...
+		default:
+			return nil, fmt.Errorf("error invoking GetContainerList: %s", err.Error())
```
fdawg4l: Same comment.  Why do we need this case?

lib/apiservers/engine/backends/volume.go
```diff
@@ -67,14 +69,21 @@ func NewVolumeModel(volume *models.VolumeResponse, labels map[string]string) *ty
...
+
 // Volumes docker personality implementation for VIC
 func (v *Volume) Volumes(filter string) ([]*types.Volume, []string, error) {
 	defer trace.End(trace.Begin("Volume.Volumes"))
 	var volumes []*types.Volume
 
 	client := PortLayerClient()
 	if client == nil {
-		return nil, nil, derr.NewErrorWithStatusCode(fmt.Errorf("Failed to get a portlayer client"), http.StatusInternalServerError)
+		return nil, nil, derr.NewErrorWithStatusCode(fmt.Errorf("failed to get a portlayer client"), http.StatusInternalServerError)
```
fdawg4l: This err string is mentioned 4 times in this file.  How about we make this a const or a var and return that instead?

lib/apiservers/engine/backends/volume.go
```diff
@@ -96,12 +105,103 @@ func (v *Volume) Volumes(filter string) ([]*types.Volume, []string, error) {
...
 		}
 		volume := NewVolumeModel(vol, volumeMetadata.Labels)
 		volumes = append(volumes, volume)
 	}
-	return volumes, nil, nil
+
+	// Parse and validate filter args
+	volumeFilters, err := filters.FromParam(filter)
+	if err != nil {
+		return nil, nil, derr.NewErrorWithStatusCode(err, http.StatusInternalServerError)
```
fdawg4l: This error instance is created more than once.  Lets make it a const or var and return that instead.

lib/apiservers/engine/backends/volume.go
```diff
@@ -96,12 +105,103 @@ func (v *Volume) Volumes(filter string) ([]*types.Volume, []string, error) {
...
+	// If the dangling filter is specified, gather required items beforehand
+	dangling := false
+	joinedVolumes := make(map[string]struct{})
+	if volumeFilters.Include("dangling") {
+
+		// Validate dangling filter's usage (per Docker code)
+		if volumeFilters.ExactMatch("dangling", "true") || volumeFilters.ExactMatch("dangling", "1") {
+			dangling = true
+		} else if !volumeFilters.ExactMatch("dangling", "false") && !volumeFilters.ExactMatch("dangling", "0") {
+			return nil, nil, fmt.Errorf("invalid filter 'dangling=%s'", volumeFilters.Get("dangling"))
```
fdawg4l: Curious, why is this invalid?  We don't support dangling == false?

lib/apiservers/engine/backends/volume.go
```diff
@@ -96,12 +105,103 @@ func (v *Volume) Volumes(filter string) ([]*types.Volume, []string, error) {
...
+		}
+
+		// Get all containers and use containers' metadata to obtain non-dangling (joined) volumes
+		conts, err := getAllContainers()
+		if err != nil {
+			return nil, nil, derr.NewErrorWithStatusCode(err, http.StatusInternalServerError)
```
fdawg4l: Same internal server error const comment.

lib/apiservers/engine/backends/volume.go
```diff
@@ -96,12 +105,103 @@ func (v *Volume) Volumes(filter string) ([]*types.Volume, []string, error) {
...
+		return nil, nil, derr.NewErrorWithStatusCode(err, http.StatusInternalServerError)
+	}
+
+	if volumeFilters.Len() == 0 {
+		return volumes, nil, nil
+	}
+
+	// If the dangling filter is specified, gather required items beforehand
+	dangling := false
+	joinedVolumes := make(map[string]struct{})
```
fdawg4l: Minor:  I think i understand what `joinedVolumes` are for, but it's possible (and likely) that I don't or will forget.  Can you please add a comment about what this is and what you're intending to do with it?

anchal-agrawal: Done.

lib/apiservers/engine/backends/volume.go
anchal-agrawal: Fixed.

lib/apiservers/engine/backends/volume.go
anchal-agrawal: `dangling=false` is valid. This is checking for supported `dangling` filter formats: `true`, `false`, `1` and `0` (per Docker).

lib/apiservers/engine/backends/volume.go
anchal-agrawal: Removed.

lib/apiservers/engine/backends/volume.go
```diff
@@ -67,50 +69,119 @@ func NewVolumeModel(volume *models.VolumeResponse, labels map[string]string) *ty
 type Volume struct {
 }
 
+// acceptedVolumeFilters are filters that are supported by VIC
+var acceptedVolumeFilters = map[string]bool{
+	"dangling": true,
+	"name":     true,
+	"driver":   true,
+}
+
```
cgtexmex: docker 1.13 adds support for label filtering -- you should either add that now or at a minimum have it as an unacceptedVolumeFilter

lib/apiservers/engine/backends/volume.go
```diff
@@ -67,50 +69,119 @@ func NewVolumeModel(volume *models.VolumeResponse, labels map[string]string) *ty
...
+		var s struct{}
+		for i := range conts {
+			for _, vol := range conts[i].VolumeConfig {
+				joinedVolumes[vol.Name] = s
+			}
+		}
+	}
+
+	log.Infoln("volumes found:")
```
cgtexmex: Are the `log.Infoln` on 138 & 140 still needed?

anchal-agrawal: Those have been there since 98378fc23eae0fa9f103b71ea34aa3be944e35b1 and seem like they'd be helpful to tell which volumes got filtered out.

cgtexmex: ok - it's not excessive so coolio

lib/apiservers/engine/backends/volume.go
anchal-agrawal: Added filtering by `label` as well!

matthewavery:
lgtm, very well done man! It really did balloon into a much larger change!

PR #3847, opened 2017-02-03
implement image filtering [specific ci=1-03-Docker-Images]

### LINKED ISSUE #2248: Implement images filtering

cgtexmex:
Provides support for filtering of images via the following criteria:
  * before, since, reference, label

Also provides support for filtering by command argument:
  * docker images busy*
This would return any images beginning with the string busy

Fixes #2248

### DIFF
--- a/lib/apiservers/engine/backends/image.go
+++ b/lib/apiservers/engine/backends/image.go
@@ -18,7 +18,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"sort"
 	"strings"
 	"time"
 
@@ -55,20 +54,8 @@ var acceptedImageFilterTags = map[string]bool{
 
 // currently not supported by vic
 var unSupportedImageFilters = map[string]bool{
-	"dangling":  false,
-	"label":     false,
-	"before":    false,
-	"since":     false,
-	"reference": false,
+	"dangling": false,
 }
-
-// byCreated is a temporary type used to sort a list of images by creation
-// time.
-type byCreated []*types.Image
-
-func (r byCreated) Len() int           { return len(r) }
-func (r byCreated) Swap(i, j int)      { r[i], r[j] = r[j], r[i] }
-func (r byCreated) Less(i, j int) bool { return r[i].Created < r[j].Created }
 
 type Image struct {
 }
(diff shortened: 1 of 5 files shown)


lib/apiservers/engine/backends/image.go
```diff
@@ -216,21 +203,45 @@ func (i *Image) Images(filterArgs string, filter string, all bool) ([]*types.Ima
...
+	for i := range images {
 
-	// sort on creation time
-	sort.Sort(sort.Reverse(byCreated(result)))
+		// provide filter with current ImageID
+		filterContext.ID = images[i].ImageID
+
+		// provide image labels
+		if images[i].Config != nil {
+			filterContext.Labels = images[i].Config.Labels
```
sflxn: Is this what provides labels filtering?

tests/test-cases/Group1-Docker-Commands/1-03-Docker-Images.robot
```diff
@@ -45,13 +45,27 @@ No-trunc images
...
+    Should Contain  ${output}  3.1
+
+Filter images since
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} images -f since=alpine:3.1
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${output}  Error
+    @{lines}=  Split To Lines  ${output}
+    Length Should Be  ${lines}  3
+    Should Contain  ${output}  latest
+
```
sflxn: I know it's tough, but you weren't able to find an image that has a label?  I think I tried to find one in the past and couldn't find one.  I assume you left out label filtering test because of this?

cgtexmex: Yeah..I couldn't find an image with a label..or google couldn't, so I created an image with a label and pushed to a private registry to test...

Yes..I didn't include an integration test for label, because I don't know of a public image with labels.  That said...label filtering is tested in `docker ps` and image filtering uses the same code.

lib/apiservers/engine/backends/image.go
cgtexmex: No - The [filterCommon func](https://github.com/cgtexmex/vic/blob/vogue/lib/apiservers/engine/backends/filter/filter.go#L63) provides the common code that enables label filtering.

That same logic will be used by `docker ps`, `docker images`, `docker volume ls` and `docker network ls` -- we just have to ensure we are presenting the labels correctly.

tests/test-cases/Group1-Docker-Commands/1-03-Docker-Images.robot
cgtexmex: FTR:  If you know of a public image w/a label(s) I'll be happy to create an image specific integration test...

lib/apiservers/engine/backends/image.go
```diff
@@ -216,21 +203,45 @@ func (i *Image) Images(filterArgs string, filter string, all bool) ([]*types.Ima
...
+		}
+
+		// determine if image should be part of list
+		action := vicfilter.IncludeImage(imageFilters, filterContext)
+
+		switch action {
+		case vicfilter.ExcludeAction:
```
fdawg4l: What about `IncludeAction`?

PR #3856, opened 2017-02-06
Rename VCH endpoint resource flags

### LINKED ISSUE #2760: appliance-memory and appliance-cpu option names are completely misleading
We have agreed upon the following nomenclature:
- The vApp (on VC) or resource pool (on ESX) that contains the VCH and container VMs = the **VCH appliance**
- The VM inside the vApp/RP = the **VCH endpoint VM**

The `vic-machine create --memory`, `vic-machine create --cpu` and associated reservation/shares settings adjust memory/CPU settings on the vApp/RP, i.e. on the whole appliance.

The old `vic-machine create --appliance-memory` and `vic-machine create --appliance-cpu` settings adjust memory/cpu settings on the VM inside the appliance, i.e. the VCH endpoint VM.

Now that the word `pool`has been removed from the `cpu`and `memory`option names, the naming of the old options `--appliance-memory` and `--appliance-cpu` is even more misleading than it was previously.

There was talk at one point of completely removing these older options. Should we do so?

andrewtchin:
Fixes #2760 

```
⇒  bin/vic-machine-linux create -x
NAME:
   vic-machine-linux create - Deploy VCH

USAGE:
   vic-machine-linux create [command options] [arguments...]

OPTIONS:
...
   --endpoint-memory value                          Memory for the VCH endpoint VM, in MB. Does not impact resources allocated per container. (default: 2048)
...
   --endpoint-cpu value                             vCPUs for the VCH endpoint VM. Does not impact resources allocated per container. (default: 1)
...

```

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -322,9 +322,9 @@ func (c *Create) Flags() []cli.Flag {
 			Hidden: true,
 		},
 		cli.IntFlag{
-			Name:        "appliance-memory",
+			Name:        "endpoint-memory",
 			Value:       2048,
-			Usage:       "Memory for the appliance VM, in MB. Does not impact resources allocated per container.",
+			Usage:       "Memory for the VCH endpoint VM, in MB. Does not impact resources allocated per container.",
 			Hidden:      true,
 			Destination: &c.MemoryMB,
 		},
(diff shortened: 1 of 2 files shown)


mhagen-vmware:
lgtm

PR #3871, opened 2017-02-07
[skip ci] Allow overriding guest id via an env. var

caglar10ur:


caglar10ur:
@mhagen-vmware  looks like this pr requires your approve

PR #3872, opened 2017-02-07
Add timestamps to container output log entries

### LINKED ISSUE #1738: Container logs need to have timestamp or a mechanism to index by timestamp
Container logs currently do not have any timestamps.  For full docker logs functionality, the logs must have some mechanism for associating log lines with a timestamp.

As a developer on admiral, I need VIC to implement timestamps on docker logs in order for admiral to work with VIC.

jzt:
This change adds a LogWriter that wraps the container VM process output in JSON along with a timestamp before sending it on its way down the serial connection to be written to `output.log`. It also adds a LogReader that does the reverse, pulling the log message out of the JSON struct before streaming it back down through the docker persona to the docker CLI after `docker logs` is called.

Fixes #1738

### DIFF
--- a/cmd/tether/ops_linux.go
+++ b/cmd/tether/ops_linux.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


hmahmood:
@caglar10ur that post makes no mention of timestamps for the log entries; how are those done in docker? The binary format is a good idea nonetheless.

jzt:
@caglar10ur Thanks for the heads up. I assumed from reading related issues around this that docker was using JSON as well. I will re-approach.

caglar10ur:
@hmahmood I think https://github.com/docker/docker/tree/master/pkg/jsonlog is what they use, at least that's what a quick search in their codebase showed

caglar10ur:
@jzt I think we can keep this PR open as I was just trying to learn why we ended up with json and whether thought about other possibilities. Grass may not be greener (aka binary log) on the other side depending on what we want. I think we first need to answer what we want :)

PR #3873, opened 2017-02-07
[skip ci] Appending the timestamp to nightly log file so we don't overwrite

rajanashok:
Appending the current timestamp to the log file name to make it unique during upload.
Setting the NSX test status to Not Implemented until the test is ready.


matthewavery:
lgtm

rajanashok:
Appending the current timestamp to the log file name to make it unique during upload.
Setting the NSX test status to Not Implemented until the test is ready.

PR #3874, opened 2017-02-07
[specific ci=1-01-Docker-Info] Reporting VCH stats in docker info

### LINKED ISSUE #3656: Reporting VCH stats
As a user of Admiral, I should be able to see CPU and memory information for a VIC VCH.

Outstanding questions to be answered before work starts:
1. where should this data be presented and in what form (docker info is one option)
3. what resources should be included:
  a. cpu & memory - is it an issue if we report MHz for cpu or should it be translated into vCPU counts?
  c. disk? per datastore? per volume store?
  d. network IO metrics?
4. limits on cpu & memory are easy, but how should be present priority? i.e. over-commit behaviours

chengwang86:
Fixes #3656 

Admiral requires that the vic engine reports the realtime cpu/mem usage of the resource pool in docker info.

### DIFF
--- a/lib/apiservers/engine/backends/system.go
+++ b/lib/apiservers/engine/backends/system.go
@@ -57,6 +57,8 @@ type System struct {
 const (
 	systemStatusMhz    = " VCH mhz limit"
 	systemStatusMemory = " VCH memory limit"
+	systemStatusCPUUsageMhz    = " VCH CPU usage"
+	systemStatusMemUsage = " VCH memory usage"
 	systemOS           = " VMware OS"
 	systemOSVersion    = " VMware OS version"
 	systemProductName  = " VMware Product"
@@ -158,6 +160,7 @@ func (s *System) SystemInfo() (*types.Info, error) {
 	if err != nil || vchInfo == nil {
 		log.Infof("System.SystemInfo unable to get vch info from port layer: %s", err.Error())
 	} else {
+		log.Debugf("In SystemInfo, vchInfo is: %+v", vchInfo)
 		if vchInfo.CPUMhz > 0 {
 			info.NCPU = int(vchInfo.CPUMhz)
 
@@ -168,6 +171,14 @@ func (s *System) SystemInfo() (*types.Info, error) {
 			info.MemTotal = vchInfo.Memory * 1024 * 1024 // Get Mebibytes
 
 			customInfo := [2]string{systemStatusMemory, units.BytesSize(float64(info.MemTotal))}
+			info.SystemStatus = append(info.SystemStatus, customInfo)
+		}
+		if vchInfo.CPUUsage > 0 {
+			customInfo := [2]string{systemStatusCPUUsageMhz, fmt.Sprintf("%d Mhz", int(vchInfo.CPUUsage))}
+			info.SystemStatus = append(info.SystemStatus, customInfo)
+		}
+		if vchInfo.MemUsage > 0 {
+			customInfo := [2]string{systemStatusMemUsage, units.BytesSize(float64(vchInfo.MemUsage))}
 			info.SystemStatus = append(info.SystemStatus, customInfo)
 		}
 		if vchInfo.HostProductName != "" {
--- a/lib/apiservers/engine/backends/system.go
+++ b/lib/apiservers/engine/backends/system.go
@@ -55,15 +55,15 @@ type System struct {
 }
 
 const (
-	systemStatusMhz    = " VCH CPU limit"
-	systemStatusMemory = " VCH memory limit"
-	systemStatusCPUUsageMhz    = " VCH CPU usage"
-	systemStatusMemUsage = " VCH memory usage"
-	systemOS           = " VMware OS"
-	systemOSVersion    = " VMware OS version"
-	systemProductName  = " VMware Product"
-	volumeStoresID     = "VolumeStores"
-	loginTimeout       = 20 * time.Second
+	systemStatusMhz         = " VCH CPU limit"
+	systemStatusMemory      = " VCH memory limit"
+	systemStatusCPUUsageMhz = " VCH CPU usage"
+	systemStatusMemUsage    = " VCH memory usage"
+	systemOS                = " VMware OS"
+	systemOSVersion         = " VMware OS version"
+	systemProductName       = " VMware Product"
+	volumeStoresID          = "VolumeStores"
+	loginTimeout            = 20 * time.Second
 )
 
 func NewSystemBackend() *System {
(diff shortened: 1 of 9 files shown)


chengwang86:
The test cases passed although the ci build failed.

lib/portlayer/exec/vchinfo.go
```diff
@@ -34,38 +37,29 @@ func NCPU(ctx context.Context, moref ...types.ManagedObjectReference) int64 {
...
-		return 0
-	}
-
-	limit := p.Config.CpuAllocation.GetResourceAllocationInfo().Limit
-	if limit == -1 {
-		return NCPU(ctx, *p.Parent)
-	}
-	return limit
-}
+	ps := []string{"config.cpuAllocation", "config.memoryAllocation", "runtime.cpu", "runtime.memory", "parent"}
```
cgtexmex: Nice!  one call -- kudos!

tests/test-cases/Group1-Docker-Commands/1-01-Docker-Info.robot
```diff
@@ -69,21 +86,39 @@ Correct container count
...
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -d busybox /bin/top
+    Should Be Equal As Integers  ${rc}  0
+
+    Sleep  10s  wait for vsphere stats update
+
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} info
+    Should Be Equal As Integers  ${rc}  0
+
+    ${cpuval}  ${memval}=  Get resource pool CPU and mem usages  ${output}
+    Should Not Be Equal As Integers  ${oldcpuval}  ${cpuval}
```
anchal-agrawal: Should we enforce a stricter comparison for these using less than / greater than instead of not equal?

chengwang86:
Make sense. Old values should be less than new values.

lib/apiservers/engine/backends/system.go
```diff
@@ -170,6 +172,14 @@ func (s *System) SystemInfo() (*types.Info, error) {
 			customInfo := [2]string{systemStatusMemory, units.BytesSize(float64(info.MemTotal))}
 			info.SystemStatus = append(info.SystemStatus, customInfo)
 		}
+		if vchInfo.CPUUsage >= 0 {
+			customInfo := [2]string{systemStatusCPUUsageMhz, fmt.Sprintf("%d Mhz", int(vchInfo.CPUUsage))}
```
vburenin: You should not specify a size here, otherwise you create an array that will be copied almost each time you pass it around:
```
customInfo := []string{systemStatusCPUUsageMhz, fmt.Sprintf("%d Mhz", int(vchInfo.CPUUsage))}
```

lib/apiservers/engine/backends/system.go
```diff
@@ -55,8 +55,10 @@ type System struct {
 }
 
 const (
-	systemStatusMhz    = " VCH mhz limit"
+	systemStatusMhz    = " VCH CPU limit"
 	systemStatusMemory = " VCH memory limit"
+	systemStatusCPUUsageMhz    = " VCH CPU usage"
```
vburenin: have you run go fmt? Please do.
in gogland you can change settings to run goimports on save.

lib/portlayer/exec/exec.go
```diff
@@ -112,8 +112,14 @@ func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSou
 
 		// Grab the AboutInfo about our host environment
 		about := sess.Vim25().ServiceContent.About
-		Config.VCHMhz = NCPU(ctx)
-		Config.VCHMemoryLimit = MemTotal(ctx)
+
+		p, err := GetVCHstats(ctx)
+		if err != nil {
+			log.Errorf("Portlayer GetVCHstats error: %s", err)
```
vburenin: "Failed to get VCH stats: %s" is a better message.

lib/portlayer/exec/vchinfo.go
```diff
@@ -19,12 +19,14 @@ import (
 
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
+	log "github.com/Sirupsen/logrus"
+	"fmt"
```
vburenin: fix import order please.

lib/portlayer/exec/vchinfo.go
```diff
@@ -34,38 +36,29 @@ func NCPU(ctx context.Context, moref ...types.ManagedObjectReference) int64 {
...
-	}
+	log.Debugf("The VCH stats are: %+v", stats)
 
-	if err := Config.ResourcePool.Properties(ctx, r, []string{"parent", "config.memoryAllocation"}, &p); err != nil {
-		return 0
+	// If any of the stats is -1 (s is true), we need to get the vch stats from the parent resource pool
+	s := false
+	for _, v := range stats {
+		if v == -1 {
+			s = true
```
vburenin: just call ```return GetVCHstats(ctx, *p.Parent)``` from here. no need for that weird "s" flag.

lib/portlayer/exec/vchinfo.go
caglar10ur: goimports -w -local github.com/vmware $(find . -mindepth 1 -maxdepth 1 -type d -not -name vendor)

lib/portlayer/exec/vchinfo.go
```diff
@@ -19,12 +19,14 @@ import (
...
+	"fmt"
 )
 
-// NCPU returns the CPU limit (MHZ)
-func NCPU(ctx context.Context, moref ...types.ManagedObjectReference) int64 {
+func GetVCHstats(ctx context.Context, moref ...types.ManagedObjectReference) (*mo.ResourcePool, error) {
+
 	if Config.ResourcePool == nil {
-		return 0
+		return &mo.ResourcePool{}, fmt.Errorf("Config.ResourcePool is nil")
```
caglar10ur: why we are returning an empty moref instead of nil?

lib/portlayer/exec/vchinfo.go
caglar10ur: Agreed, not clear why do you need s?

lib/apiservers/engine/backends/system.go
chengwang86: @vburenin Loc defined info.SystemStatus as "SystemStatus       [][2]string", so I couldn't simply use customInfo := []string..., unless Loc changes his code.

lib/portlayer/exec/exec.go
chengwang86: Done. Thanks

lib/portlayer/exec/vchinfo.go
chengwang86: Done. Thanks

PR #3875, opened 2017-02-08
Adding new iptables rules for bridge isolation [full ci]

### LINKED ISSUE #2498: Inter-container network isolation
2 tests:
1. two containers on multiple networks overlapping should be able to ping each other
~~2. two containers on different networks non-overlapping should not be able to ping each other~~ (this is due to the containers still being attached to the bridge network)

Currently, when I run this test it fails both tests.  1. the containers cannot ping each other, 2. the containers can ping each other.

[1-17-Docker-Network-Connect-VCH-0-1300-container-logs.zip](https://github.com/vmware/vic/files/498598/1-17-Docker-Network-Connect-VCH-0-1300-container-logs.zip)

[log.html.zip](https://github.com/vmware/vic/files/498602/log.html.zip)

hmahmood:
Fixes #2498

### DIFF
--- a/cmd/vic-init/main_linux.go
+++ b/cmd/vic-init/main_linux.go
@@ -169,6 +169,7 @@ func configureToolbox(t *tether.Toolbox) *tether.Toolbox {
 func externalIP() string {
 	l, err := netlink.LinkByName("client")
 	if err != nil {
+		log.Debugf("error looking up client interface by name: %s", err)
 		l, err = netlink.LinkByAlias("client")
 		if err != nil {
 			log.Errorf("error looking up client interface by alias: %s", err)
--- a/cmd/vic-init/main_linux.go
+++ b/cmd/vic-init/main_linux.go
@@ -169,8 +169,6 @@ func configureToolbox(t *tether.Toolbox) *tether.Toolbox {
 func externalIP() string {
 	l, err := netlink.LinkByName("client")
 	if err != nil {
-		log.Warnf("error looking up client interface by name: %s", err)
-
 		l, err = netlink.LinkByAlias("client")
 		if err != nil {
 			log.Errorf("error looking up client interface by alias: %s", err)
--- a/cmd/vic-init/ops.go
+++ b/cmd/vic-init/ops.go
@@ -92,7 +92,7 @@ func (t *operations) SetupFirewall() error {
 	if l == nil {
 		l, err = netlink.LinkByAlias(publicIfaceName)
 		if l == nil {
-			return fmt.Errorf("could not find public interface")
+			return fmt.Errorf("could not find interface: %s", publicIfaceName)
 		}
 	}
 
--- a/cmd/vic-init/ops.go
+++ b/cmd/vic-init/ops.go
@@ -24,6 +24,8 @@ import (
 	"time"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/docker/libnetwork/iptables"
+	"github.com/vishvananda/netlink"
 
 	"github.com/vmware/vic/lib/pprof"
 	"github.com/vmware/vic/lib/tether"
@@ -36,8 +38,9 @@ import (
 var pathPrefix string
 
 const (
-	logDir  = "var/log/vic"
-	initLog = "init.log"
+	logDir          = "var/log/vic"
+	initLog         = "init.log"
+	publicIfaceName = "public"
 )
 
 type operations struct {
@@ -81,6 +84,20 @@ func (t *operations) HandleSessionExit(config *tether.ExecutorConfig, session *t
 func (t *operations) SetHostname(name string, aliases ...string) error {
 	// switch the names around so we get the pretty name and not the ID
 	return t.BaseOperations.SetHostname(aliases[0])
+}
+
+func (t *operations) SetupFirewall() error {
+	// get the public interface name
+	l, err := netlink.LinkByName(publicIfaceName)
+	if l == nil {
+		l, err = netlink.LinkByAlias(publicIfaceName)
+		if l == nil {
+			return fmt.Errorf("could not find public interface")
+		}
+	}
+
+	_, err = iptables.Raw(string(iptables.Append), "FORWARD", "-i", "bridge", "-o", l.Attrs().Name, "-j", "ACCEPT")
+	return err
 }
 
 func (t *operations) Apply(endpoint *tether.NetworkEndpoint) error {
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"math/rand"
 	"net"
+	"os"
 	"strconv"
 	"strings"
 	"sync"
@@ -40,6 +41,7 @@ import (
 	timetypes "github.com/docker/engine-api/types/time"
 	"github.com/docker/go-connections/nat"
 	"github.com/docker/go-units"
+	"github.com/docker/libnetwork/iptables"
 	"github.com/docker/libnetwork/portallocator"
 	"github.com/vishvananda/netlink"
 
@@ -108,6 +110,11 @@ var (
 
 	portMapper portmap.PortMapper
 
+	// bridge-to-bridge rules, indexed by mapped port;
+	// this map is used to delete the rule once
+	// the container stops or is removed
+	btbRules map[string][]string
+
 	cbpLock         sync.Mutex
 	containerByPort map[string]string // port:containerID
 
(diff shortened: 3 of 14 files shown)


lib/tether/tether.go
```diff
@@ -390,6 +390,11 @@ func (t *tether) Start() error {
 		}
 		extraconfig.Encode(t.sink, t.config)
 
+		// setupt the firewall
```
anchal-agrawal: s/setupt/setup

cmd/vic-init/ops.go
```diff
@@ -83,6 +85,20 @@ func (t *operations) SetHostname(name string, aliases ...string) error {
 	return t.BaseOperations.SetHostname(aliases[0])
 }
 
+func (t *operations) SetupFirewall() error {
+	// get the public interface name
+	l, err := netlink.LinkByName("public")
```
vburenin: Can we have constants for these "public" and others? I see it is getting used too frequent, thus it becomes error prone.

lib/apiservers/engine/backends/portmap/portmap.go
```diff
@@ -165,7 +177,7 @@ func (p *portMapper) forward(action iptables.Action, ip net.IP, port int, proto,
 		}
 		return fmt.Errorf("Failed to find unmap data for %s:%d", ipStr, port)
 
-	case iptables.Append:
+	case Map:
```
vburenin: I am not happy to see such large chunk of logic within Switch/case clause, can it be extracted out?

tests/resources/Docker-Util.robot
```diff
@@ -31,6 +31,12 @@ Hit Nginx Endpoint
     ${rc}  ${output}=  Run And Return Rc And Output  wget ${vch-ip}:${port}
     Should Be Equal As Integers  ${rc}  0
 
+Get Container IP
+    [Arguments]  ${docker-params}  ${id}  ${network}=default
+    ${rc}  ${ip}=  Run And Return Rc And Output  docker ${docker-params} network inspect ${network} | jq '.[0].Containers."${id}".IPv4Address' | cut -d \\" -f 2 | cut -d \\/ -f 1
```
mhagen-vmware: Why pass in ${docker-params} instead of using %{VCH-PARAMS}?

hmahmood: Trying to keep the function as generic as possible.

cmd/vic-init/main_linux.go
```diff
@@ -169,8 +169,6 @@ func configureToolbox(t *tether.Toolbox) *tether.Toolbox {
 func externalIP() string {
 	l, err := netlink.LinkByName("client")
 	if err != nil {
-		log.Warnf("error looking up client interface by name: %s", err)
```
vburenin: I would log it on debug level, it might be useful.

cmd/vic-init/ops.go
```diff
@@ -36,8 +38,9 @@ import (
 var pathPrefix string
 
 const (
-	logDir  = "var/log/vic"
-	initLog = "init.log"
+	logDir          = "var/log/vic"
```
vburenin: Should not it start from root "/"?

cmd/vic-init/ops.go
```diff
@@ -83,6 +86,20 @@ func (t *operations) SetHostname(name string, aliases ...string) error {
 	return t.BaseOperations.SetHostname(aliases[0])
 }
 
+func (t *operations) SetupFirewall() error {
+	// get the public interface name
+	l, err := netlink.LinkByName(publicIfaceName)
```
vburenin: This logic seems familiar. Would we want to have a dedicated private function that will perform double lookup?

cmd/vic-init/ops.go
```diff
@@ -83,6 +86,20 @@ func (t *operations) SetHostname(name string, aliases ...string) error {
...
+	if l == nil {
+		l, err = netlink.LinkByAlias(publicIfaceName)
+		if l == nil {
+			return fmt.Errorf("could not find public interface")
```
vburenin: I would definitely include interface name into this error message.

cmd/vic-init/ops.go
```diff
@@ -83,6 +86,20 @@ func (t *operations) SetHostname(name string, aliases ...string) error {
...
+		}
+	}
+
+	_, err = iptables.Raw(string(iptables.Append), "FORWARD", "-i", "bridge", "-o", l.Attrs().Name, "-j", "ACCEPT")
```
vburenin: At the end I would prefer to see "dedicated" methods which are adding specific IP tables rules in a little more human way.

iptables.Append(chain, inInterface, outInterface, action string)

cmd/vic-init/main_linux.go
caglar10ur: Agreed, we could convert this to debug

lib/apiservers/engine/backends/container.go
```diff
@@ -108,6 +110,11 @@ var (
 
 	portMapper portmap.PortMapper
 
+	// bridge-to-bridge rules, indexed by mapped port;
+	// this map is used to delete the rule once
+	// the container stops or is removed
+	btbRules map[string][]string
```
caglar10ur: Don't we need a mutex to protect this map from concurrent access?

lib/apiservers/engine/backends/container.go
```diff
@@ -770,13 +783,60 @@ func UnmapPorts(hostconfig *containertypes.HostConfig) error {
...
+		args := []string{"-t", string(iptables.Filter),
+			"-i", bridgeIfaceName,
+			"-o", bridgeIfaceName,
+			"-p", proto,
+			"-d", containerAddr,
+			"--dport", containerPort,
+			"-j", "ACCEPT",
+		}
+
+		if _, err := iptables.Raw(append([]string{string(iptables.Insert), "VIC", "1"}, args...)...); err != nil && !os.IsExist(err) {
```
caglar10ur: Do you mind breaking this line? ```args...)...); ``` looks like a magic on a single line :)

lib/apiservers/engine/backends/container.go
```diff
@@ -770,13 +783,60 @@ func UnmapPorts(hostconfig *containertypes.HostConfig) error {
...
+
+		btbRules[hostPort] = args
+	case portmap.Unmap:
+		if args, ok := btbRules[hostPort]; ok {
+			args = append([]string{string(iptables.Delete), "VIC"}, args...)
+			if _, err := iptables.Raw(args...); err != nil && !os.IsNotExist(err) {
+				return err
+			}
+
+			delete(btbRules, hostPort)
```
caglar10ur: Wondering whether we need locks here in this function for btbRules

lib/apiservers/engine/backends/container.go
```diff
@@ -770,13 +783,60 @@ func UnmapPorts(hostconfig *containertypes.HostConfig) error {
...
+		}
+
 		// update mapped ports
 		delete(containerByPort, p.strHostPort)
 		log.Debugf("unmapped port %s", p.strHostPort)
 	}
 	return nil
 }
 
+// interBridgeTraffic enables traffic for exposed port from one bridge network to another
```
sflxn: Does this code really need to be in container.go?  Container.go and all the backend router files should really just be REST routing.

hmahmood: Currently all the port mapping code is in this file so I put this in there too. The `portmap` folder is a separate package, and this code is very docker-personality specific to put in there.

lib/apiservers/engine/backends/container.go
hmahmood: `interBridgeTraffic` is only called with `cpbLock` being held in the calling function.

cmd/vic-init/ops.go
hmahmood: Yes it is. The question is where to put the function?

cmd/vic-init/ops.go
hmahmood: Don't know; I didn't change the value, just the formatting changed.

lib/apiservers/engine/backends/portmap/portmap.go
sflxn: Maybe a comment before the function saying we don't lock here because we lock in the calling function?

sflxn:
outside from nitpicks, functionally lgtm.  We'll take care of the file overload soon.

cmd/vic-init/ops.go
vburenin: Anywhere close enough to be discovered by other people. In this case it might be even public I think.

I am not asking to do this for this review, but if you add TODO comment it would be great.

lib/apiservers/engine/backends/container.go
vburenin: I see that we've got bunch of dictionaries which require synchronized access, which points that we need to use "generics" to generate typed code that will encapsulate and protect those dictionaries.

PR #3883, opened 2017-02-08
[skip ci] Passing in the timestamp during logfile upload

rajanashok:
Missed the change to pass the updated log file name with the timestamp when calling log upload script.


rajanashok:
.

PR #3895, opened 2017-02-10
Bump to latest github.com/mailru/easyjson

jzt:
This vendor update includes parts of the github.com/mailru/easyjson repo that we were previously not including, as well as bringing the version up to date.


mdubya66:
LGTM

Issue #3754, opened 2017-01-27
Implement service components


jooskim:
As a UI developer I need to implement what was detailed in the design doc for the vSphere data service extension.

Details:
I would follow some existing patterns in the sample plugins, as they are recommended by the HTML SDK team. However, they are some points made in comment sections where some code is not intended for production usage. Thoughtful assessments are required here.

Acceptance Criteria:
1. Write required classes defined in the #3748 and unit test them
2. The extended VIC UI service is successfully loaded and works as expected

PR #3919, opened 2017-02-13
Add new features to the VIC UI service

jooskim:
This PR addresses #3754 by extending the vSphere Client Data Service. It adds a few custom data types that are to be consumed by the VIC UI Angular app.

Addresses #3754


.drone.sec
```diff
@@ -1 +1 @@
...
...V1pPYZgmwYAlWcRfqwqLTpvFpkwnAx4OOvoP5XZhSlyqhEIsr61WN9KySAGg8riv9ReiDKnJliRVKSIsVrDiZm619old5z5PDFXU1T1IAifTCAz33MY6sn1mwWjq2t6pWjzb6RJACJ6ozCz9PiTkYVrGHKtfO_pHQBYcTa4Agh6jK3UNP03qX5f0YKdS4f1r9cT_5Pa7fN4XT7IrZZZH0feT9sP0ElJePrjqCFQXAsnMm9Dj_aRHZa37Y4PFIhR2j8pGE50mkqzWB4sX5VUsSwMrOf9AZkFB1kd9WZvlAQitXJnNFchjNff4cUMoH2dqGs9IaT6m3dhcZmpUdJBi2x4Lm3jASsEPQUSKefOGep0u77Um9F_EeGchKMloLIugRR492n8O3rIaoBRXZo_BI05mwion1l29xa8vJBf09VmvftmSjx7x4WZg2xiWn5NLRMZYMMO48rTBUwM79Auj8e0OV6Z5sTNpo4KgCzophE6aYUlrIlN6g4gzfz4_pks4izJ1qTrd81CBiamb9WvKutTF5xOLoRcjIyzTQAnOZjqOXhECUHZ6gV7h45Aj7Z39r-bBcSDJVrLy3wpU-wDL2tHI6uXJO0xrmGJbXOunzyP2ZWWFuAVQj-ax54wjAS3m_LVNimpMITSrN7-aucpZGiwv9IgKwXGYltB1tmBlCEnlatZvbva87CU92Nla12mp8I8EXrFueyVcT-ZgppOkKF1xk55eekFvmkRuhd6QAjCZZF5FCXMqP41tkm6LsaquffdpR9tDjDA3R0tgZTBjsIUfOjlwXh_Zbzy2wRUeVTNVXBWEPfP2_QA44QjD7oN6F7zRjbd_lrXhIl6uw3UWnYy0uH0zZxSWLoAX-3v50-pzorQdsVTmgTo63frAVvbGTX8d9P_-_uVBBYOsi3-atnqNcTgPPHc9rs4HkTB5EnIoDRjVl69c4t2M7ChiOPHM0eOnNkfDxn1__it9YofZ0w0Kwkan_iBXy4eYbO1261JJAPw7wNilQHnHf8KiE_JgFAvRwJmwT0FKYRbeh-3OnxZ5UFdOeDwdz8SFEdIh232Ee20bbCuAwNknbR9adXIQPtAl0NuoNU47IAe7IxBZ1mGXvJt0M60gNzlP4aIoeXp827P4iYViofN3O7seujbtOUInu0YHBZozZ1kx5iNtN3r8w.Sj8H7qB-VK6LY0qmx1R6Yg
```
mhagen-vmware: Is this updated as of latest secrets? I just made a change at the end of last week, make sure you are diff'ing this off of the latest google doc version - should be 2-8-17

tests/nightly/nightly-autorun.sh
```diff
@@ -16,9 +16,13 @@
...
 # Get the latest code from vmware/vic repo
 cd ~/go/src/github.com/vmware/vic
-git checkout master
+git checkout -f master
 git remote update
 git rebase upstream/master
+git checkout -f ui-nightlytest/ui-tests \
+    tests/resources/VCH-Util.robot \
+    ui/vic-uia \
+    ui/vic-ui-h5c/uia
```
mhagen-vmware: Are you using this to throw away local changes that happen with the build process?  I think git reset might be a better idea?

ui/.gitignore
```diff
@@ -5,6 +5,26 @@
...
+vic-ui-linux
+vic-ui-darwin
+vic-ui-windows.exe
+vic-machine-linux
+vic-ui-automation-test/
+vic-ui-h5c/dev-build.sh
+vic-ui-h5c/prod-build.sh
+vic-ui-h5c/uia/h5-plugin-tests/ui-automation/vic-uia/work/results/
+vic-ui-h5c/uia/h5-plugin-tests/ui-automation/vic-uia/work/runlists/
+vic-ui-h5c/uia/h5-plugin-tests/ui-automation/vic-uia/work/testbeds/
```
mhagen-vmware: I think you just need the root folder to be excluded for git ignore to ignore all underneath right?  This could be potentially simplified unless there is a bunch of stuff in vic-ui-h5c/ that needs to be excluded?

.drone.sec
jooskim: I updated the secret file to the latest version and yes that's 2/8/17. I ran a build on my personal branch of CI and had to update it after seeing it fail with regression tests.

tests/nightly/nightly-autorun.sh
jooskim: This looks weird. Git merge must have mixed two versions in a wrong way. I'll correct this shortly. FYI, my intent is to only pull in files from the UI test branch and NOT overwrite any existing files.

ui/.gitignore
jooskim: That might work better. I'll add some more .gitignore files where I think they fit

jooskim:
@mhagen-vmware Thanks for your input. One reason for this PR being so big is that any WIP will break tests and thus the build pipeline. One way I could circumvent this is to defer pushing files that would affect the build process (such as .drone.yml), which I try from now on.

andrewtchin:
lgtm

rajanashok:
lgtm

PR #3924, opened 2017-02-13
Adding/modifying code to run component binaries in test mode [full-ci]

### LINKED ISSUE #3793: Instrument binaries for coverage
See #3631 for how to do this. Also attached is a patch that resulted from the research spike for the modifications that were done to `port-layer-server`.

[port-layer-server.diff.txt](https://github.com/vmware/vic/files/743262/port-layer-server.diff.txt)

Acceptance:
* makefile targets to build coverage enabled test binaries for `vic-init`, `port-layer-server`, and `docker-personality`

hmahmood:
Fixes #3793

### DIFF
--- a/cmd/docker/main.go
+++ b/cmd/docker/main.go
@@ -49,9 +49,12 @@ import (
 )
 
 type CliOptions struct {
-	serverPort    uint
-	portLayerAddr string
-	proto         string
+	serverPort    *uint
+	portLayerAddr *string
+	portLayerPort *uint
+	debug         *bool
+
+	proto string
 }
 
 const (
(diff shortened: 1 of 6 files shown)


cmd/docker/main.go
```diff
@@ -42,22 +42,37 @@ import (
...
+	vchConfig config.VirtualContainerHostConfigSpec
+	cli       CliOptions
+)
 
 func init() {
 	log.SetFormatter(viclog.NewTextFormatter())
 	trace.Logger.Level = log.DebugLevel
 	pprof.StartPprof("docker personality", pprof.DockerPort)
+
+	flag.Usage = Usage
```
vburenin: Is it really needs to be in inti() function? all parameter parsing work needs to be called from main() or any other deterministic place.

hmahmood: This is just setting up the options; the actual parsing happens in `main()`.

cmd/docker/main.go
```diff
@@ -68,18 +83,18 @@ func Usage() {
...
 	}
 
-	if err := vicbackends.Init(cli.portLayerAddr, productName, &vchConfig, vchConfig.InsecureRegistries); err != nil {
+	if err := vicbackends.Init(*cli.portLayerAddr, productName, &vchConfig, vchConfig.InsecureRegistries); err != nil {
 		log.Fatalf("failed to initialize backend: %s", err)
 	}
 
 	// Start API server wit options from command line args
-	api := startServerWithOptions(cli)
+	api := startServer()
```
vburenin: previous pattern was better. Using globals for config is evil. It will complicate unit testing.

cmd/docker/main.go
vburenin: still, these are globals. Using globals tends to be evil, especially for configuration, due to that it complicates testing especially parallel tests run.

cmd/docker/main_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"flag"
+	"testing"
+)
+
+var systemTest *bool
+
+func init() {
+	systemTest = flag.Bool("systemTest", false, "Run system test")
```
vburenin: what do you test here?

hmahmood: This is just a flag to run `main()` inside a unit test (see below) for code coverage. See the approach outlined here: https://www.elastic.co/blog/code-coverage-for-your-golang-system-tests

sflxn: Ok, I read the blog.  The question is how does this test the execution of the binary without personality or vicinit flags?

hmahmood: The goal is to run the binary with code coverage turned on, and the only way to do that is to use the go testing framework. Next sprint I will work on getting these actually running in the CI instead of the regular binaries.

Issue #3925, opened 2017-02-13
NFS client + NFS volume PL integration


matthewavery:
User statement:
As an engineer, I need to allow for the creation of nfs shared volume stores during the init cycle of the portlayer server. Once this is done the targets provided by vic-machine can be consumed.

Details: This ticket should most likely be worked on in conjunction with #3810 because there must be a way to communicate nfs targets to the portlayer `configure` api handler at boot time of the portlayer server. This is where the vsphere volume stores are made and this is where the nfs volume stores should also be created. This means that we must explore where we want to put the nfs targets to be read in at init time of the portlayer. Vic-machine will need to facilitate the persistence of these targets to the portlayer. This work has been marked separately from the `VolumeStore` implementation piece since it relies on both the VolumeStore being implemented and vic-machine passing in the appropriate targets to consume. This is also the main area where we will distinguish between a nfs and vsphere volume store. 

Acceptance criteria: Successful creation of volumestores on targeted nfs servers after the portlayer has finished it's init process and is waiting to serve requests.

fdawg4l:
@matthewavery I'm not following what changes you're describing.  What handlers specifically?  And what changes do they need?

The way we handle volumes for vsphere and nfs should be identical from the point of view of the PL REST API.  They should not need to change. 

If you specifically mean the `Configure` operation in the portlayer server which bootstraps the storage layer, then we should reword the summary of this issue.

fdawg4l:
Fixed by https://github.com/vmware/vic/pull/4154 and https://github.com/vmware/vic/pull/4173.

Issue #3803, opened 2017-01-31
NFS client implementation for NFS VolumeStore


fdawg4l:
The NFS backed volume implementation described [here](https://github.com/vmware/vic/blob/master/doc/design/volumes/nfsvolumestores.md) calls out the need for the VCH to manipulate the NFS target.  We _could_ use the VCH's kernel nfs-client to mount the target, and use VFS to prepare and iterate the target namespace for our volumes.  But, we want to limit the number of external dependencies to the `port-layer-daemon`, including the kernel nfs-client.  For our use-case, the target will be mounted as needed and unmounted when done.  From experience, if the target becomes unavailable while manipulating the namespace or having files open, this often leads to the kernel throwing `VFS sync` errors, causing the whole system to become unstable.  The `sync` syscall fails to work in this mode thereby requiring the system to be rebooted.  

Our usage model is much smaller and more well defined than that of a general purpose filesystem.  We're identifying the subset of calls we require (`mkdir`, `unlink`, `setattr`, `readdir`, and `lookup` so far) and need to implement a pure `go` implementation of the nfs client.  I have `mount`/`unmount`, `mkdir`, and `setattr` working so far.  The failure mode for any of these calls so far is local to the `go` routine calling them, thereby not effecting the availability of the whole appliance.

fdawg4l:
This is currently with legal and the OSS team.  I'm working with them to bring this into the vmware org on GH while maintaining the original license (public domain).  The expectation is we'll vendor it once we've got it in the vmware org but I'm working on getting this in sooner via my fork so we can at least make progress.

fdawg4l:
The licensing issue was taking too long relative to the amount of time it would take me to unencumber my changes from the original fork.  I'd rather rewrite the bits that are original than let this linger.  

I'm mostly through.  There is one package left and then other than some white space, the diff between my fork and the original should be complete.

Issue #2303, opened 2016-09-12
Shared data volumes


anchal-agrawal:
## Promoted to epic
There is a general requirement for shared volumes with concurrent access across multiple containers.
Some basic usages are in the conversation below and we should confirm those are addressed by however this feature is implemented.

Delivered feature design should cover at least:
1. Where the shared data is hosted (in appliance, container, or out-of-band)
2. How volume store is specified during VCH creation and operation, including credentials if necessary
3. What failure modes exist and how failures are addressed
4. Lifecycle of data vs vmdk based volume stores

####  Original detail by @anchal-agrawal:
Follow-up of #2281 and #2291

We should add support for mounting directories as a volume on containers with the `-v` option.

**Expected behavior:**

```
$ docker -H 192.168.77.131:2375 run -it -v ~/myapp:/myapp busybox
/ # ls
bin    dev    etc    home   myapp  proc   root   sys    tmp    usr    var
```

**Actual behavior:**

```
$ docker -H 192.168.77.131:2375 run -it -v ~/myapp:/myapp busybox
docker: Error response from daemon: path /storage/volumes/home/anchala/myapp was not found.
See 'docker run --help'.
```

hickeng:
@anchal-agrawal This just isn't going to work with vic as there's no mechanism available to us to share that local directory with the remote container.
Even with regular docker I don't think it has value beyond either:
* volume create (if a name is required), or
* anonymous volume
unless docker have added a mechanism to upload the contents of the client's directory to the server.
In your example, what is in the myapp directory when you're in the container?

anchal-agrawal:
Gotcha, thanks George. In my example, the `myapp` dir contains an empty file. Running `docker run -it -v ~/myapp:/myapp busybox` with vanilla docker, any changes inside `myapp` from the host are synced to the container.

hickeng:
Changes from the _host_ are reflected in the contianer, not change from the _client_ - I'm assuming they're separate systems?

anchal-agrawal:
Sorry, in my example, the host system and the client are the same system. I see that adding support for NFS volumes would achieve the intended goal.

mdubya66:
Moving to Icebox

mreferre:
@hickeng this is coming up more and more as of late. I think we should prioritize this work (NFS driver support) if possible. 

/cc @corrieb

npakrasi:
Added to 0.7.0 release notes known issues.

hickeng:
[...]

PR #3926, opened 2017-02-13
Nfs volume Store implementation

### LINKED ISSUE #3806: Create NFS Volume Store Implementation
per the design document generated from the NFS shared volume research spike([here](https://github.com/vmware/vic/blob/master/doc/design/volumes/nfsvolumestores.md), this ticket is responsible for implementing the interface for the Volume Store. The definition of this interface already exists and contains the operations `VolumeCreate, VolumeRemove, and VolumeList` Which are the three primary operations that we require for volume support in VIC thus far. 

The NFSv3VolumeStore will be implementing these calls by leveraging the interface defined in #3803 the `NFSTarget`. These calls should be fairly simple and only concern themselves with the interaction of one volume that has already been validated up to the filesystem(e.g. it exists in the cache, it is or isn't in use. etc.). 

For this piece it is important to note that the `Target` piece will be sticking as close to the `os` package signatures and errors as possible. This should make mocking and testing the NFSv3 volume store more simple and possible. 

This piece will be worked on in conjuction with #3803 in order to identify all of the needed endpoints that the Target interface will need to expose.

matthewavery:
fixes #3806 

This is the NFS Volume Store implementation which relies on the implementation from ticket #3803. The `Target` interface defines the api calls used by the nfs volume store implementation and the unit tests have mocked these out with go's base `os` package. The code for populating the cache with nfs volume stores has not been added yet. That will likely be part of the personality and portlayer handler work in ticket #3925.

There is no current path for creating nfs volume stores, this is just one part of the piece for moving us toward being able to write integration tests showing nfs volumestore support(#2303)

This may still need some additional alignment with @fdawg4l 's nfs client additions.

### DIFF
--- a/lib/portlayer/storage/nfs/disk.go
+++ b/lib/portlayer/storage/nfs/disk.go
@@ -18,30 +18,30 @@ import (
 	"net/url"
 )
 
-//  Device information for interaction with the tether and portlayer
-type NFSVolume struct {
+//  Volume identifies an NFS based volume
+type Volume struct {
 
 	// This is the nfs host the the volume belongs to
 	Host *url.URL
 
 	// Path on the Host where the volume is located
-	NFSPath string
+	Path string
 }
 
-func NewNFSVolumeDevice(host *url.URL, NFSPath string) NFSVolume {
-	v := NFSVolume{
-		Host:    host,
-		NFSPath: NFSPath,
+func NewVolume(host *url.URL, NFSPath string) Volume {
+	v := Volume{
+		Host: host,
+		Path: NFSPath,
 	}
 	return v
 }
 
-func (v NFSVolume) MountPath() (string, error) {
-	return "", nil
+func (v Volume) MountPath() (string, error) {
+	return v.Path, nil
 }
 
-// Includes url to nfs directory for container to mount,
-func (v NFSVolume) DiskPath() url.URL {
+// DiskPath includes the url to the nfs directory for the container to mount,
+func (v Volume) DiskPath() url.URL {
 	if v.Host == nil {
 		return url.URL{}
 	}
--- a/lib/portlayer/storage/nfs/disk.go
+++ b/lib/portlayer/storage/nfs/disk.go
@@ -1,0 +1,44 @@
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nfs
+
+import (
+	"github.com/vmware/vic/lib/portlayer/storage"
+)
+
+type NFSVolumeBacking struct {
+
+	//This is the path where the volume directory exists on the nfs server
+	NFSPath string
+
+	//This is the target filesystem path for mounting into a container
+	MountPath string
+}
+
+func NewNFSVolumeBacking(NFSPath string) NFSVolumeBacking {
+	v := NFSVolumeBacking{
+		NFSPath:   NFSPath,
+		MountPath: MountTargetPath,
+	}
+	return v
+}
+
+func (v *NFSVolumeBacking) MountPath() (string, error) {
+	return nil
+}
+
+func (v *NFSVolumeBacking) DiskPath() string {
+	return v.NFSPath
+}
--- a/lib/portlayer/storage/nfs/disk.go
+++ b/lib/portlayer/storage/nfs/disk.go
@@ -1,4 +1,4 @@
-// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
+// Copyright 2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -18,13 +18,13 @@ import (
 	"net/url"
 )
 
-//device information for interaction with the tether and portlayer
+//  Device information for interaction with the tether and portlayer
 type NFSVolume struct {
 
-	//This is the nfs host the the volume belongs to
+	// This is the nfs host the the volume belongs to
 	Host *url.URL
 
-	//Path on the Host where the volume is located
+	// Path on the Host where the volume is located
 	NFSPath string
 }
 
@@ -40,7 +40,7 @@ func (v NFSVolume) MountPath() (string, error) {
 	return "", nil
 }
 
-// includes url to nfs directory for container to mount,
+// Includes url to nfs directory for container to mount,
 func (v NFSVolume) DiskPath() url.URL {
 	if v.Host == nil {
 		return url.URL{}
--- a/lib/portlayer/storage/nfs/disk.go
+++ b/lib/portlayer/storage/nfs/disk.go
@@ -15,30 +15,35 @@
 package nfs
 
 import (
-	"github.com/vmware/vic/lib/portlayer/storage"
+	"net/url"
 )
 
-type NFSVolumeBacking struct {
+//device information for interaction with the tether and portlayer
+type NFSVolume struct {
 
-	//This is the path where the volume directory exists on the nfs server
+	//This is the nfs host the the volume belongs to
+	Host *url.URL
+
+	//Path on the Host where the volume is located
 	NFSPath string
-
-	//This is the target filesystem path for mounting into a container
-	MountPath string
 }
 
-func NewNFSVolumeBacking(NFSPath string) NFSVolumeBacking {
-	v := NFSVolumeBacking{
-		NFSPath:   NFSPath,
-		MountPath: MountTargetPath,
+func NewNFSVolumeDevice(host *url.URL, NFSPath string) NFSVolume {
+	v := NFSVolume{
+		Host:    host,
+		NFSPath: NFSPath,
 	}
 	return v
 }
 
-func (v *NFSVolumeBacking) MountPath() (string, error) {
-	return nil
+func (v NFSVolume) MountPath() (string, error) {
+	return "", nil
 }
 
-func (v *NFSVolumeBacking) DiskPath() string {
-	return v.NFSPath
+// includes url to nfs directory for container to mount,
+func (v NFSVolume) DiskPath() url.URL {
+	if v.Host == nil {
+		return url.URL{}
+	}
+	return *v.Host
 }
--- a/lib/portlayer/storage/nfs/target.go
+++ b/lib/portlayer/storage/nfs/target.go
@@ -1,4 +1,4 @@
-// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
+// Copyright 2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -17,34 +17,39 @@ package nfs
 import (
 	"io"
 	"net/url"
-
 	"os"
 )
 
-type MountHandler interface {
-	Mount(target *url.URL) (NFSTarget, error)
-	Unmount(target NFSTarget) error
+// The MountServer is an interface used to communicate with network attached storage.
+type MountServer interface {
+	// Mount initiates the NAS Target and returns a Target interface.
+	Mount(target *url.URL) (target, error)
+
+	// Unmount terminates the Mount on the Target.
+	Unmount(target target) error
 }
 
-type NFSTarget interface {
-	//opens a target path in a READONLY context
+// Target is the filesystem interface for performing actions against attached storage.:w
+
+type target interface {
+	// Opens a target path in a READONLY context
 	Open(path string) (io.ReadCloser, error)
 
-	//Opens targeted file with the supplied attr.
+	// Opens targeted file with the supplied attr.
 	OpenFile(path string, perm os.FileMode) (io.ReadWriteCloser, error)
 
-	//Creates a file, errors out if file already exists. assumes write permissions.
+	// Creates a file, errors out if file already exists. assumes write permissions.
 	Create(path string, perm os.FileMode) (io.ReadWriteCloser, error)
 
 	// Create directory path
-	MkDir(path string, perm os.FileMode) ([]byte, error)
+	Mkdir(path string, perm os.FileMode) ([]byte, error)
 
 	// Delete Directory Path, and children
 	RemoveAll(Path string) error
 
-	//Reads the contents of the targeted directory
+	// Reads the contents of the targeted directory
 	ReadDir(path string) ([]os.FileInfo, error)
 
-	//Looks up the file information for a target entry
+	// Looks up the file information for a target entry
 	Lookup(path string) (os.FileInfo, error)
 }
--- a/lib/portlayer/storage/nfs/target.go
+++ b/lib/portlayer/storage/nfs/target.go
@@ -1,0 +1,67 @@
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nfs
+
+import (
+	"errors"
+	"io"
+	"net/url"
+
+	"github.com/vmware/vic/pkg/trace"
+)
+
+type NFSTarget interface {
+	// write data to target nfs filesystem
+	Write(op trace.Operation, path string, r io.Writer) error
+
+	// read data from target nfs filesystem
+	Read(path string) io.Reader
+
+	// Create directory path
+	MkDir(path string, makeParents bool) (string, error)
+
+	// Delete Directory Path, and children
+	RemoveAll(Path string) error
+
+	//Reads the contents of the targeted directory
+	ReadDir(path string)
+}
+
+type NFSv3Target struct{}
+
+func (t *NFSv3Target) Write(op trace.Operation, path string, r io.Writer) error {
+	return nil
+}
+
+func (t *NFSv3Target) Read(path string) io.Reader {
+	return nil
+}
+
+func (t *NFSv3Target) MkDir(path string, makeParents bool) (string, error) {
+	return nil, nil
+}
+
+// TODO: client implementation here
+func NewNFSv3Target(op trace.Operation, fqdn *url.URL) (NFSTarget, error) {
+	return NFSv3Target{}, errors.New("FUNCTION NOT IMPLEMENTED")
+}
+
+func OpenNFSv3Target(targetURL *url.URL) (*NFSv3Target, error) {
+	return nil
+}
+
+func CloseNFSv3Target(target *NFSv3Target) error {
+	return nil
+}
--- a/lib/portlayer/storage/nfs/target.go
+++ b/lib/portlayer/storage/nfs/target.go
@@ -20,14 +20,24 @@ import (
 	"net/url"
 
 	"github.com/vmware/vic/pkg/trace"
+	"os"
 )
 
 type NFSTarget interface {
+	//returns the NFSTarget Endpoint
+	EndPoint() *url.URL
+
+	//Binds the target NFS server to the VCH(Possible concurrency/performance issue?)
+	Open() (NFSTarget, error)
+
+	//Unbinds NFS server from the VCH(ditto to open)
+	Close() error
+
 	// write data to target nfs filesystem
-	Write(op trace.Operation, path string, r io.Writer) error
+	Write(op trace.Operation, path string, r io.Reader) error
 
 	// read data from target nfs filesystem
-	Read(path string) io.Reader
+	Read(path string) (io.Reader, error)
 
 	// Create directory path
 	MkDir(path string, makeParents bool) (string, error)
@@ -36,32 +46,53 @@ type NFSTarget interface {
 	RemoveAll(Path string) error
 
 	//Reads the contents of the targeted directory
-	ReadDir(path string)
+	ReadDir(path string) ([]os.FileInfo, error)
 }
 
-type NFSv3Target struct{}
+type NFSv3Target struct {
+	//nfs enpoint
+	host *url.URL
 
-func (t *NFSv3Target) Write(op trace.Operation, path string, r io.Writer) error {
+	//Path to the the volume store
+	directoryPath string
+}
+
+func (t NFSv3Target) EndPoint() *url.URL {
 	return nil
 }
 
-func (t *NFSv3Target) Read(path string) io.Reader {
+func (t NFSv3Target) Write(op trace.Operation, path string, r io.Reader) error {
 	return nil
 }
 
-func (t *NFSv3Target) MkDir(path string, makeParents bool) (string, error) {
+func (t NFSv3Target) Read(path string) (io.Reader, error) {
 	return nil, nil
 }
 
+func (t NFSv3Target) MkDir(path string, makeParents bool) (string, error) {
+	return "", nil
+}
+
+func (t NFSv3Target) RemoveAll(path string) error {
+	return nil
+}
+
+func (t NFSv3Target) ReadDir(path string) ([]os.FileInfo, error) {
+	return nil, nil
+}
+
+func (t NFSv3Target) Open() (NFSTarget, error) {
+	return nil, nil
+}
+
+func (t NFSv3Target) Close() error {
+	return nil
+}
+
 // TODO: client implementation here
-func NewNFSv3Target(op trace.Operation, fqdn *url.URL) (NFSTarget, error) {
+func NewNFSv3Target(op trace.Operation, fqdn *url.URL, volumeDirectory string) (NFSTarget, error) {
 	return NFSv3Target{}, errors.New("FUNCTION NOT IMPLEMENTED")
 }
 
-func OpenNFSv3Target(targetURL *url.URL) (*NFSv3Target, error) {
-	return nil
-}
-
-func CloseNFSv3Target(target *NFSv3Target) error {
-	return nil
-}
+//MAKE MOCK TARGET
+//USER TEMP DIR
--- a/lib/portlayer/storage/nfs/target.go
+++ b/lib/portlayer/storage/nfs/target.go
@@ -15,11 +15,9 @@
 package nfs
 
 import (
-	"errors"
 	"io"
 	"net/url"
 
-	"github.com/vmware/vic/pkg/trace"
 	"os"
 )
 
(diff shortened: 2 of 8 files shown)


lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+			//XXX: Should we attempt to clean up what we wrote in this case?
+			return err
+		}
+	}
+	op.Infof("Successfully wrote metadata to (%s)", metadataPath)
+	return nil
+}
+
+func getMetadata(op trace.Operation, metadataPath string, target NFSTarget) (map[string][]byte, error) {
+	op.Infof("Attempting to volume metadata at (%s)", metadataPath)
```
anchal-agrawal: `"Attempting to volume metadata at"` - you're probably a word.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+			return nil, err // FIXME: refine errors later...
+		}
+
+		fileBlob, err := target.Open(targetPath)
+		if err != nil {
+			return nil, err
+		}
+
+		dataBlob, err := ioutil.ReadAll(fileBlob)
+
```
anchal-agrawal: Unneeded whitespace?

lib/portlayer/storage/nfs/disk.go
```diff
@@ -0,0 +1,49 @@
...
+		NFSPath: NFSPath,
+	}
+	return v
+}
+
+func (v NFSVolume) MountPath() (string, error) {
+	return "", nil
+}
+
+// includes url to nfs directory for container to mount,
```
anchal-agrawal: It's nice to have function comments that begin with the function's name for godoc puposes: https://blog.golang.org/godoc-documenting-go-code

lib/portlayer/storage/nfs/disk.go
```diff
@@ -0,0 +1,49 @@
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
```
anchal-agrawal: `Copyright 2017`

lib/portlayer/storage/nfs/target.go
```diff
@@ -0,0 +1,50 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package nfs
+
+import (
+	"io"
+	"net/url"
+
+	"os"
```
anchal-agrawal: gofmt: pkg `os` should be in the same group as the other two.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+import (
+	"net/url"
+	"os"
+	"path"
+
+	"github.com/vmware/vic/lib/portlayer/storage"
+	"github.com/vmware/vic/lib/portlayer/util"
+	"github.com/vmware/vic/pkg/trace"
+	"io/ioutil"
```
anchal-agrawal: gofmt: pkg `io/ioutil` should be in the top group.

lib/portlayer/storage/nfs/target.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+type MountHandler interface {
+	Mount(target *url.URL) (NFSTarget, error)
+	Unmount(target NFSTarget) error
+}
+
+type NFSTarget interface {
```
hmahmood: This seems generic enough that it could just be `FSTarget`, or even just `FileSystem`. You could even argue some of the functionality is part of https://github.com/vmware/vic/blob/master/lib/system/syscall.go

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+
+	if err := writeMetadata(op, v.volMetadataDirPath(ID), info, target); err != nil {
+		return nil, err
+	}
+
+	op.Infof("nfs volume (%s) successfully created on volume store (%s)", ID, v.Name)
+	return vol, nil
+}
+
+// Removes a volume and all of it's contents from the nfs store. We already know via the cache if it is in use.
```
anchal-agrawal: s/it's/its

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+func (v *NFSv3VolumeStore) VolumeDestroy(op trace.Operation, vol *storage.Volume) error {
+	target, err := v.MH.Mount(v.Target)
+	if err != nil {
+		return err
+	}
+
+	defer v.MH.Unmount(target)
+
+	op.Infof("Attempting to remove volume (%s) and it's metadata from volume store (%s)", vol.ID, v.Name)
```
anchal-agrawal: s/it's/its

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+
+	//remove volume directory and children
+	if err := target.RemoveAll(v.volDirPath(vol.ID)); err != nil {
+		op.Errorf("Failed to remove volume (%s) on volume store (%s) due to error (%s)", vol.ID, v.Name, err)
```
anchal-agrawal: Error strings should start with lowercase.

lib/portlayer/storage/nfs/target.go
```diff
@@ -0,0 +1,50 @@
...
+package nfs
+
+import (
+	"io"
+	"net/url"
+
+	"os"
+)
+
+type MountHandler interface {
```
hmahmood: This seems to be more of a system function than just nfs. How about using https://github.com/vmware/vic/blob/master/lib/system/syscall.go?

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+	VolumesDir = "volumes"
+
+	// path that namespaces the metadata for a specific volume
+	metadataDir = "volumedata"
+
+	//Stock file permissions that are set, In the future we may pass these in.
+	filePermissions = 0755
+)
+
+type NFSv3VolumeStore struct {
```
hmahmood: Does this type need to be exported? I assume it implements an interface?

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+// Creates a volume directory and volume object for NFS based volumes
+func (v *NFSv3VolumeStore) VolumeCreate(op trace.Operation, ID string, store *url.URL, capacityKB uint64, info map[string][]byte) (*storage.Volume, error) {
+	target, err := v.MH.Mount(v.Target)
+	if err != nil {
+		return nil, err
+	}
+
+	defer v.MH.Unmount(target)
+
+	if _, err := target.MkDir(v.volDirPath(ID), filePermissions); err != nil {
```
hmahmood: Do we clean this up if there is an error further down? What if the dir already exists; is that still an error?

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+	if err != nil {
+		return err
+	}
+
+	defer v.MH.Unmount(target)
+
+	op.Infof("Attempting to remove volume (%s) and it's metadata from volume store (%s)", vol.ID, v.Name)
+
+	//remove volume directory and children
+	if err := target.RemoveAll(v.volDirPath(vol.ID)); err != nil {
```
hmahmood: Still an error if the dir does not exist?

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+		op.Errorf("Failed to remove metadata for volume (%s) at path (%q) on volume store (%s)", vol.ID, v.volDirPath(vol.ID), v.Name)
+		//FIXME: Should we bail here? the volume is gone at this point...
+	}
+	op.Infof("Successfully removed volume (%s) from volumestore (%s)", vol.ID, v.Name)
+
+	return nil
+}
+
+func (v *NFSv3VolumeStore) VolumesList(op trace.Operation) ([]*storage.Volume, error) {
+	volumes := make([]*storage.Volume, 0)
```
hmahmood:
```go
var volumes []*storage.Volume
``` 
is the canonical way.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+		return nil, err
+	}
+
+	op.Infof("nfs volume (%s) successfully created on volume store (%s)", ID, v.Name)
+	return vol, nil
+}
+
+// Removes a volume and all of it's contents from the nfs store. We already know via the cache if it is in use.
+func (v *NFSv3VolumeStore) VolumeDestroy(op trace.Operation, vol *storage.Volume) error {
+	target, err := v.MH.Mount(v.Target)
```
hmahmood: Can this be done once, or do we have to mount/unmount for every call?

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+	dataKeys, err := target.ReadDir(metadataPath)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, metadataFile := range dataKeys {
+		targetPath := path.Join(metadataPath, metadataFile.Name())
+
+		//we need to make sure it is there since we are using open.
+		_, err := target.Lookup(targetPath)
```
hmahmood: Wouldn't open return an error if `targetPath` was not there?

lib/portlayer/storage/nfs/target.go
fdawg4l: @hmahmood we aren't using the `syscall` interface or VFS for this.

lib/portlayer/storage/nfs/target.go
fdawg4l: Same comment.  We aren't using VFS.

lib/portlayer/storage/nfs/volume.go
fdawg4l: @hmahmood lingering mounts for our model is wasteful.  Each mount allocates a page (or set of pages) on the server.  And for our usage model, that's really unnecessary.  And depending on the server implementation, mounting can be trivial.

lib/portlayer/storage/volume.go
```diff
@@ -27,9 +27,13 @@ import (
...
 
 type Disk interface {
+	//Path to this disk on the VCH
 	MountPath() (string, error)
-	DiskPath() string
-	//FIXME: Add a capacity and populate it.
+
+	//Path to the disk on the datastore
+	DiskPath() url.URL
+	//STUFF THE STRING INTO THE PATH FIELD FOR VSPHERE IMPLEMENTATION
```
fdawg4l: Please nuke this comment.

lib/portlayer/storage/volume.go
```diff
@@ -27,9 +27,13 @@ import (
 )
 
 type Disk interface {
+	//Path to this disk on the VCH
```
fdawg4l: Format this string with a space between the //.

lib/portlayer/storage/nfs/target.go
hmahmood: There is considerable overlap between the syscall interface and here; should you guys consider using that interface instead with your own implementation? Just trying to cut down on duplication here.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+func writeMetadata(op trace.Operation, metadataPath string, info map[string][]byte, target NFSTarget) error {
+	//NOTE: right now we do not support updating metadata, thus we make the ID directory here
+	_, err := target.MkDir(metadataPath, filePermissions)
+	if err != nil {
+		return err
+	}
+
+	op.Infof("Attempting to write metadata to (%s)", metadataPath)
```
fdawg4l: `s/Attempting to write/Writing/g`

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+		targetPath := path.Join(metadataPath, fileName)
+		blobFile, err := target.Create(targetPath, filePermissions)
+		// XXX: we might want to make sure the file we want to write does not exist...
+		if err != nil {
+			return err
+		}
+
+		_, err = blobFile.Write(data)
+		if err != nil {
+			//XXX: Should we attempt to clean up what we wrote in this case?
```
fdawg4l: Same consistency comment;  we'll revisit what to do in this case.  Likely for this specific case, we'll clean up in the cache.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+	//NOTE: right now we do not support updating metadata, thus we make the ID directory here
+	_, err := target.MkDir(metadataPath, filePermissions)
+	if err != nil {
+		return err
+	}
+
+	op.Infof("Attempting to write metadata to (%s)", metadataPath)
+	for fileName, data := range info {
+		targetPath := path.Join(metadataPath, fileName)
+		blobFile, err := target.Create(targetPath, filePermissions)
```
fdawg4l: You need to `defer blobFile.Close()`.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+func getMetadata(op trace.Operation, metadataPath string, target NFSTarget) (map[string][]byte, error) {
+	op.Infof("Attempting to volume metadata at (%s)", metadataPath)
+	metadataInfo := make(map[string][]byte)
+	dataKeys, err := target.ReadDir(metadataPath)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, metadataFile := range dataKeys {
+		targetPath := path.Join(metadataPath, metadataFile.Name())
```
fdawg4l: Name collision.  Lets reserve `target\c` to refer to the NFS Target itself.  How about renaming this to `pth`.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+
+		//we need to make sure it is there since we are using open.
+		_, err := target.Lookup(targetPath)
+		if err != nil {
```
fdawg4l: You can use the `os` helpers to check if the file exists or doesn't here.  https://golang.org/pkg/os/#IsNotExist

lib/portlayer/storage/nfs/volume.go
fdawg4l: @matthewavery , agree with @hmahmood here.  Just call `Open` directly.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+			return nil, err // FIXME: refine errors later...
+		}
+
+		fileBlob, err := target.Open(targetPath)
```
fdawg4l: Remember to `defer <file>.Close()`.

lib/portlayer/storage/nfs/volume_test.go
```diff
@@ -0,0 +1,516 @@
...
+	op := trace.NewOperation(context.TODO(), "TestOp")
+
+	//Create a Volume Store
+	vs, err := NewVolumeStore(op, "testStore", &targetURL, mockMount)
+	if !assert.Nil(t, err, "Failed during call to NewVolumeStore with err (%s)", err) {
+		return
+	}
+
+	_, err = os.Stat(dirpath)
+	if !assert.Nil(t, err, "Could not find the initial volume store directory after creation of volume store. err (%s)", err) {
```
fdawg4l: `assert.NoError()`

lib/portlayer/storage/nfs/volume_test.go
```diff
@@ -0,0 +1,516 @@
...
+	//Check Metadata Pathing
+	metadataPath := path.Join(dirpath, metadataDir)
+	volumePath := path.Join(dirpath, VolumesDir)
+
+	metaFilesDir, err := os.Open(metadataPath)
+	if !assert.Nil(t, err, "opening the metadata directory failed with err (%s)", err) {
+		return
+	}
+
+	volumeDir, err := os.Open(volumePath)
```
fdawg4l: defer close

lib/portlayer/storage/nfs/volume.go
matthewavery: Ah yes, when I wrote this I was thinking in a step by step manner. will do.

lib/portlayer/storage/volume.go
matthewavery: will do.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+
+	//nfs target for filesystem interaction
+	Target *url.URL
+
+	//Service selflink to volume store.
+	SelfLink *url.URL
+}
+
+func NewVolumeStore(op trace.Operation, storeName string, nfsTargetURL *url.URL, mount MountHandler) (*NFSv3VolumeStore, error) {
+	// XXX: Potential credential leak, maybe log less...
```
matthewavery: will do

lib/portlayer/storage/nfs/volume.go
matthewavery: the close() was called below in both cases. I will just add the defer instead. since that is canonical `go`.

lib/portlayer/storage/nfs/target.go
```diff
@@ -0,0 +1,50 @@
...
+
+type NFSTarget interface {
+	//opens a target path in a READONLY context
+	Open(path string) (io.ReadCloser, error)
+
+	//Opens targeted file with the supplied attr.
+	OpenFile(path string, perm os.FileMode) (io.ReadWriteCloser, error)
+
+	//Creates a file, errors out if file already exists. assumes write permissions.
+	Create(path string, perm os.FileMode) (io.ReadWriteCloser, error)
```
matthewavery: So for this one it was a nuance of the UnitTests. :-/ Looks like os.OpenFile will return an error if the actual file does not exist. So I moved to having a `Create` here after writing my tests. Let me know if we still want to avoid this. I can change the Target Mock to call os.Create for OpenFile if the file does not exist.

lib/portlayer/storage/nfs/disk.go
matthewavery: I will go back and add this when I get the chance. I want to take care of the more pressing stuff first.

lib/portlayer/storage/nfs/target.go
matthewavery: MountServer is good :-) I do have a bad habbit of calling things "Handlers" been trying to kill that habit for awhile.

lib/portlayer/storage/nfs/target.go
```diff
@@ -0,0 +1,50 @@
...
+	"os"
+)
+
+type MountHandler interface {
+	Mount(target *url.URL) (NFSTarget, error)
+	Unmount(target NFSTarget) error
+}
+
+type NFSTarget interface {
+	//opens a target path in a READONLY context
```
matthewavery: I capitalized the "o" here as well no worries :-)

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+
+const (
+	// This is the same as our vmdk approach namespacing the data further on the filesystem
+	VolumesDir = "volumes"
+
+	// path that namespaces the metadata for a specific volume
+	metadataDir = "volumedata"
+
+	//Stock file permissions that are set, In the future we may pass these in.
+	filePermissions = 0755
```
matthewavery: Done, and I updated the comment above to remove the word "file"

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+
+	defer mount.Unmount(target)
+
+	//we assume that nfsTargetURL.path already exists.
+	//make volumes directory
+	if _, err := target.MkDir(path.Join(nfsTargetURL.Path, VolumesDir), filePermissions); err != nil && !os.IsExist(err) {
+		return nil, err
+	}
+
+	//make metadata directory
```
matthewavery: It is made here during the creation of a VolumeStore. Then it is used on line 104 in function `volMetadataDirPath(ID string)` which will construct the path to a particular volumes metadata. Then in that directory in the `writeMetadata` function we write out the individual keys in the info map as flat binary files.

lib/portlayer/storage/nfs/volume.go
matthewavery: This is an interesting question, It will likely error with the dir does not exist, or if the path does not exist in particular since this can return a path error in the standard `os` package in golang as well(maybe not ours). However, If the dir is not there and the intention was to remove do we want to return an error or not if this scenario occurs? @fdawg4l my thoughts are that we are attempting to remove something that the cache thought was there, in which case we may want to error since something bad has occurred(most likely the directory was removed OOB via vsphere UI, govc, etc...).

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+
+	op.Infof("Attempting to remove volume (%s) and it's metadata from volume store (%s)", vol.ID, v.Name)
+
+	//remove volume directory and children
+	if err := target.RemoveAll(v.volDirPath(vol.ID)); err != nil {
+		op.Errorf("Failed to remove volume (%s) on volume store (%s) due to error (%s)", vol.ID, v.Name, err)
+		return err
+	}
+
+	//XXX: what should we do if we lose the nfs server between volume and metadata deletion?
```
matthewavery: Yes I will repost these comments to a new tickets when we get the chance. My feeling is that we may possibly need to recover and mark the volume as deleted if we sanely can. The issue to me is that if we do not mark the volume as deleted the first time around we no longer actually have the volume. We just have some metadata left, so if the user gets confused or another(or the same) user try to attach that volume to container they are going to get a "does not exist" style error from us.

lib/portlayer/storage/nfs/volume.go
matthewavery: filtered out `.` and `..` and moved the statement down. I thought I had already moved the statement lower, but didn't even realize the `.` and `..` cases. My unit tests did not catch this, so i might need to find a way to make sure that filter is tested as well.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,260 @@
...
+	volFileInfo, err := target.ReadDir(v.volumesDir())
+	if err != nil {
+		return nil, err
+	}
+
+	for _, volFileInfo := range volFileInfo {
+
+		volMetadata, err := getMetadata(op, v.volMetadataDirPath(volFileInfo.Name()), target)
+		if err != nil {
+			return nil, err //perhaps wait until the end of the loop to report this?
```
matthewavery: For now I continue and collecting the error in a slice. then at the end I append them all to one error and return them as a whole to the caller.

lib/portlayer/storage/nfs/volume.go
matthewavery: funny how one missing word makes things looks so bad :-P

lib/portlayer/storage/nfs/volume.go
matthewavery: removed useless call to LookUp.

lib/portlayer/storage/nfs/volume.go
matthewavery: yeah I used them above too, idk what I was thinking.

lib/portlayer/storage/nfs/volume.go
matthewavery: Yes I previously had them below. I think I deleted the calls to close but forgot to add the defer at the time. Thanks for catching that.

lib/portlayer/storage/volume.go
```diff
@@ -27,9 +27,13 @@ import (
...
 	MountPath() (string, error)
-	DiskPath() string
-	//FIXME: Add a capacity and populate it.
+
+	//Path to the disk on the datastore
+	DiskPath() url.URL
+	//STUFF THE STRING INTO THE PATH FIELD FOR VSPHERE IMPLEMENTATION
+	//NFS PUT THE STUFF IN THE HOST.  2 3
```
matthewavery: yeah this snuck itself in. it was supposed to only be a note to myself. also the ` 2 3` are clearly me trying to switch windows in spacemacs but I was in insert mode ;-P wow...

lib/portlayer/storage/nfs/volume_test.go
matthewavery: I went back and did this in a lot of places. Previously I was just closing everything I opened at the bottom of the function call.

lib/portlayer/storage/nfs/volume_test.go
```diff
@@ -0,0 +1,516 @@
...
+	volumeDir.Close()
+
+	//Remove the same Volume
+	err = vs.VolumeDestroy(op, vol)
+	if !assert.Nil(t, err, "Failed during a call to VolumeDestroy with err (%s)", err) {
+		return
+	}
+
+	metaFilesDir, err = os.Open(metadataPath)
+	if !assert.Nil(t, err, "opening the metadata directory failed with err (%s)", err) {
```
matthewavery: We are checking to make sure that the metadataPath was created properly after creating a new volumeStore. This path should exist at this point.

matthewavery: I do need to check to make sure that the path to the metadata file for the specific volume is gone though. I will add that. the path we need to check for should be(in this case) `/tmp/testStore/metaData/volOne` all of the contents of the map from volOne will be stored in the `volOne` directory.

lib/portlayer/storage/nfs/target.go
anchal-agrawal: Whoops! 😄

matthewavery:
@fdawg4l @hmahmood so I looked into why golint became angry once I unexported the `target` interface and it turns out that godoc will not document unexported types. This is what leads to the `this can be annoying to use` lint method for the nfs.target type. I also found out that lint will complain if you return a pointer from a "New" function with the same `annoying use` message if the "New" function returns a pointer.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,271 @@
...
+	var fetchErrors []error
+
+	for _, fileInfo := range volFileInfo {
+
+		if fileInfo.Name() == "." || fileInfo.Name() == ".." {
+			continue
+		}
+
+		volMetadata, err := getMetadata(op, v.volMetadataDirPath(fileInfo.Name()), target)
+		if err != nil {
```
fdawg4l: Can we log the error here by `Name()`?

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,271 @@
...
+
+		vol, err := storage.NewVolume(v.SelfLink, fileInfo.Name(), volMetadata, volDeviceBacking)
+		if err != nil {
+			return nil, err
+		}
+
+		volumes = append(volumes, vol)
+	}
+
+	if len(fetchErrors) > 0 {
```
fdawg4l: Really, just returning the last error is probably enough in this case.  We just want the caller to fail.  And if it's a systemic issue, you'll have to look at the logs anyway and provided you add the log line above, that should give the user enough chronology to figure out what happened in what order.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,271 @@
...
+	_, err := target.Mkdir(metadataPath, defaultPermissions)
+	if err != nil {
+		return err
+	}
+
+	op.Infof("Writing metadata to (%s)", metadataPath)
+	for fileName, data := range info {
+		targetPath := path.Join(metadataPath, fileName)
+		blobFile, err := target.Create(targetPath, defaultPermissions)
+		defer blobFile.Close()
```
fdawg4l: You should move this after the `if err != nil` block.  `blobFile` will `nil` in case of error and this will `segfault`.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,271 @@
...
+		// XXX: we might want to make sure the file we want to write does not exist...
+		if err != nil {
+			return err
+		}
+
+		_, err = blobFile.Write(data)
```
fdawg4l: `defer blobFile.Close()` after collecting the error.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,271 @@
...
+	if err != nil {
+		return err
+	}
+
+	op.Infof("Writing metadata to (%s)", metadataPath)
+	for fileName, data := range info {
+		targetPath := path.Join(metadataPath, fileName)
+		blobFile, err := target.Create(targetPath, defaultPermissions)
+		defer blobFile.Close()
+		// XXX: we might want to make sure the file we want to write does not exist...
```
fdawg4l: Please remove the comment.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,271 @@
...
+	dataKeys, err := target.ReadDir(metadataPath)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, metadataFile := range dataKeys {
+		pth := path.Join(metadataPath, metadataFile.Name())
+
+		fileBlob, err := target.Open(pth)
+		defer fileBlob.Close()
```
fdawg4l: Move the defer to after the error check.

lib/portlayer/storage/nfs/disk.go
```diff
@@ -0,0 +1,49 @@
...
+type NFSVolume struct {
+
+	// This is the nfs host the the volume belongs to
+	Host *url.URL
+
+	// Path on the Host where the volume is located
+	NFSPath string
+}
+
+func NewNFSVolumeDevice(host *url.URL, NFSPath string) NFSVolume {
```
matthewavery: In our case it looks like this will actually be `NewVolume`

lib/portlayer/storage/nfs/volume.go
```diff
@@ -0,0 +1,271 @@
...
+
+	op.Infof("Attempting to remove volume (%s) and its metadata from volume store (%s)", vol.ID, v.Name)
+
+	//remove volume directory and children
+	if err := target.RemoveAll(v.volDirPath(vol.ID)); err != nil {
+		op.Errorf("failed to remove volume (%s) on volume store (%s) due to error (%s)", vol.ID, v.Name, err)
+		return err
+	}
+
+	//XXX: what should we do if we lose the nfs server between volume and metadata deletion?
```
matthewavery: indeed it is now ticket #4003

lib/portlayer/storage/nfs/volume.go
matthewavery: Alright, I simply preserve the final error obtained from getting the metadata. Now we return an error and nil for the volumes. So the overall call is failed.

Issue #3787, opened 2017-01-31
We are leaking DOM_OBJECTS on VSAN


mhagen-vmware:
After running through a round of longevity testing we leaked the following objects on VSAN:
```
anon@anon-Latitude-E7450:~$ sshpass -p xxxx ssh root@xxxx cmmds-tool find | wc -l
1676
anon@anon-Latitude-E7450:~$ sshpass -p xxxx ssh root@xxxx cmmds-tool find -t DOM_OBJECT | wc -l
242
```

You can use cmmds-tool and /usr/var/lib/vmware/bin/objtool to observe and delete objects:
```
/usr/lib/vmware/osfs/bin/objtool delete -f -u 5ea33658-975d-86ff-287b-1866da1245c8
```
Here I delete all the dom objects that have been leaked:
```
sshpass -p xxxx ssh root@xxxx cmmds-tool find -t DOM_OBJECT -f json | jq -r ".entries[].uuid" | xargs -n1 sshpass -p xxxx ssh root@xxxx /usr/lib/vmware/osfs/bin/objtool delete -f -u
```

Eventually this results in the vsan component count exceeding the limit and you cannot create any VMs.

emlin:
The DOM is vsan Distributed Object Manager, which is used to manage vsan object. Not sure about their details, but through the command @mhagen-vmware listed, found following scenarios.

For create:
- Every time one VM or even just one namespace is created, there is one DOM created
- Add one VMDK file to one VM will create one DOM as well
- Copy VMDK file or upload datastore files do not create DOM
- Detach VMDK does not reduce DOM
- If VMDK is chained, the DOM number is less than disk numbers, I guess DOM is only created based on attached VMDK

For delete:
- DOM object will be deleted if VMDK is attached and VM is deleted
- If VMDK is detached, VM delete will leave vm folder there, and then DOM is kept as well
- If delete VMDK file directly, DOM is not deleted
- If delete VM folder through filemanager directly, with VMDK files inside, DOM is cleaned
- If delete folder, but with VMDK deleted early, DOM is left there
- Delete folder through namespace manager or filemanager, does not matter. Both of them can clean the DOM for root folder namespace, but not for the deleted VMDK files. (No need to remember the namespace UUID)

So overall, this issue is caused by our recursive delete operation, but that is the workaround for vSAN datastore cleanup. Although I do see sometimes, delete folder directly can succeed, but most of time, that does not work because our chained vmdk files.

emlin:
[...]

PR #3929, opened 2017-02-13
Add vsan dom cache, and concurrent access interface

emlin:
This is the vsan dom sync up code. Based on this, vic-machine delete, docker rmi, and docker volume rm should call this to delete
  vsan dom object

  The code in vic-machine and storage portlayer will be in other PRs
Related to #3787


pkg/vsphere/vsan/dom_cache.go
```diff
@@ -0,0 +1,226 @@
...
+		deletedVMDKs = append(deletedVMDKs, vmdk)
+	}
+
+	for _, vmdk := range deletedVMDKs {
+		delete(v.paths, vmdk)
+	}
+}
+
+// Refresh searches dom objects from vsan datastore, and build reverse index for vmdk files
+// Tthe vmdk file format is removed vsan datastore header, e.g. /vmfs/volumes/vsan:52932941b44e2147-f1490d38c9730c6d/, to make it searchable through vmfs file path
```
anchal-agrawal: s/Tthe/The

pkg/vsphere/vsan/dom_cache.go
```diff
@@ -0,0 +1,226 @@
...
+	defer trace.End(trace.Begin(path))
+
+	if !strings.HasSuffix(path, vmdkFileSuffix) {
+		log.Debugf("non vmdk file %s, no need to cache", path)
+		return "", nil
+	}
+
+	loc := vsanNameRegex.FindStringIndex(path)
+	if len(loc) == 0 {
+		err := errors.Errorf("patthern %s is not found in path %s", vsanNamePattern, path)
```
anchal-agrawal: s/patthern/pattern

pkg/vsphere/vsan/dom_cache.go
```diff
@@ -0,0 +1,226 @@
...
+			uuids = append(uuids, uuid)
+		} else {
+			ret = append(ret, path)
+		}
+	}
+	v.m.Unlock()
+
+	force := true
+	res, err := v.hvis.DeleteVsanObjects(ctx, uuids, &force)
+	var deletedDoms []string
```
anchal-agrawal: We should declare `deletedDoms` closer to where it's used, possibly on line 84.

pkg/vsphere/vsan/dom_cache.go
```diff
@@ -0,0 +1,226 @@
...
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/vic/pkg/errors"
+	"github.com/vmware/vic/pkg/trace"
+)
+
+const (
+	vsanNamePattern = "vsan:[a-zA-Z0-9-]*/"
```
andrewtchin: is `vsan:/` a valid name? if no, and at least one char should be required between `:` and `/`, the regex should be `vsan:[a-zA-Z0-9-]+/`

pkg/vsphere/vsan/host_vsan_internal_system.go
```diff
@@ -0,0 +1,159 @@
...
+
+	res, err := methods.GetVsanObjExtAttrs(ctx, m.cl, &req)
+	if err != nil {
+		return ret, err
+	}
+
+	if res == nil {
+		return ret, nil
+	}
+	log.Debugf("Returned attributes: %s", res.Returnval)
```
anchal-agrawal: Since this func doesn't have a trace, it'd be helpful to say something like `"Returned attributes from GetVsanObjExtAttrs:"` in this log statement.

pkg/vsphere/vsan/host_vsan_internal_system.go
```diff
@@ -0,0 +1,159 @@
...
+	ret = make(map[string]VSANExtAttrs)
+	json.Unmarshal([]byte(res.Returnval), &ret)
```
anchal-agrawal: Should we check for an error from `Unmarshal`, just in case?

pkg/vsphere/vsan/host_vsan_internal_system_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+		DatastorePath: env.DS(t),
+
+		Insecure:  true,
+		Keepalive: time.Duration(5) * time.Minute,
+	}
+
+	s, err := session.NewSession(config).Connect(context.Background())
+	s.Populate(context.Background())
+	if err != nil {
```
anchal-agrawal: Should this error check go above `s.Populate(context.Background())`?

pkg/vsphere/vsan/synced_cache.go
```diff
@@ -0,0 +1,439 @@
...
+}
+
+func (c *syncedDomCache) closeChannels() {
+	defer trace.End(trace.Begin(""))
+	close(c.refresh)
+	close(c.stop)
+	close(c.doneOnce)
+	close(c.waitCurrentRefresh)
+	c.m.Lock()
+	log.Debugf("Finish all watigroups")
```
anchal-agrawal: s/watigroups/waitgroups

pkg/vsphere/vsan/dom_cache.go
vburenin: Is there any length limit also? like 256 characters or something. I would impose the limit requirements into regex as {1, 256} or something.

pkg/vsphere/vsan/dom_cache.go
```diff
@@ -0,0 +1,226 @@
...
+	var deletedDoms []string
+	if err != nil {
+		return ret, DomDeleteError{
+			Err:         err,
+			FailedUuids: uuids,
+		}
+	}
+	var failedIds []string
+	for _, r := range res {
+		if !r.Success {
```
vburenin: for clarity purposes, I would just swap if else.if r.Success {
... do y
} else {
... do z.
}

pkg/vsphere/vsan/dom_cache.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+	}
+	return ret, nil
+}
+
+func (v *vsanDSDomCache) cleanDeletedUUIDs(deletedDoms []string) {
+	defer trace.End(trace.Begin(fmt.Sprintf("%s", deletedDoms)))
+
+	var deletedVMDKs []string
+	v.m.Lock()
```
vburenin: you are trying to lock same mutex in the caller, it is a potential risk to a deadlock.
In this case I would just grab it in the initial caller and defer unlock.

pkg/vsphere/vsan/dom_cache.go
```diff
@@ -0,0 +1,226 @@
...
+	defer v.m.Unlock()
+	for _, uuid := range deletedDoms {
+		vmdk := v.uuids[uuid]
+		delete(v.uuids, uuid)
+		deletedVMDKs = append(deletedVMDKs, vmdk)
+	}
+
+	for _, vmdk := range deletedVMDKs {
```
vburenin: no need two loops. everything can be done in one shot.

pkg/vsphere/vsan/dom_cache.go
```diff
@@ -0,0 +1,226 @@
...
+	defer v.m.Unlock()
+
+	uuids, err := v.hvis.QueryVsanObjectUuidsByFilter(ctx, nil, 0, 0)
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+	// do not query existing uuids, as we don't rename vmdk file
+	setIds := make(map[string]struct{}, len(uuids))
+	for i := range uuids {
```
vburenin: for _, uuid := range {
   setIds[uuid] = struct{}{}
}

pkg/vsphere/vsan/dom_cache.go
```diff
@@ -0,0 +1,226 @@
...
+		setIds[uuids[i]] = struct{}{}
+	}
+	for exist := range v.uuids {
+		if _, ok := setIds[exist]; ok {
```
vburenin: this if redundant. https://play.golang.org/p/3bX2yHFsT0

pkg/vsphere/vsan/dom_cache.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	leftIds := make([]string, len(setIds))
+	i := 0
+	for left := range setIds {
+		leftIds[i] = left
+		i++
+	}
+	mAttrs, err := v.hvis.GetVsanObjExtAttrs(ctx, leftIds)
+	if err != nil {
+		log.Error(err)
```
vburenin: is caller logs the error?

pkg/vsphere/vsan/dom_cache.go
emlin: I'm not try to validate the vmdk path here, all the path in cache will be queried from vSphere API. If you specify one wrong VMDK path to delete, it won't be deleted.
Too many restriction here just had too many relationship with vSAN impl I think.

pkg/vsphere/vsan/host_vsan_internal_system.go
emlin: hmm.. actually it's just one debug information during my development.
This should be in govmomi, and I didn't see any such kind of debug information in govmomi code.

pkg/vsphere/vsan/dom_cache.go
emlin: I don't want to lock vsphere operations, but to make it clear, removed the internal method with lock/unlock

pkg/vsphere/vsan/dom_cache.go
emlin: good catch

pkg/vsphere/vsan/host_vsan_internal_system.go
```diff
@@ -0,0 +1,159 @@
...
+import (
+	"context"
+	"encoding/json"
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/methods"
+	//	"github.com/vmware/govmomi/vim25/mo"
```
vburenin: Do you keep these imports on purpose?

pkg/vsphere/vsan/host_vsan_internal_system.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// TODO: HostVsanInternalSystem should be moved to govmomi
+type HostVsanInternalSystem struct {
+	object.Common
+	cl *vim25.Client
+}
+
+func NewHostVsanInternalSystem(c *vim25.Client, ref types.ManagedObjectReference) *HostVsanInternalSystem {
+	m := HostVsanInternalSystem{
```
vburenin: in this particular case you could just:
```
return &HostVsanInternalSystem{
    Common: object.NewCommon(c, ref),
    cl:     c,
}
```

However, this is a personal preference.

pkg/vsphere/vsan/host_vsan_internal_system.go
```diff
@@ -0,0 +1,159 @@
...
+		Common: object.NewCommon(c, ref),
+		cl:     c,
+	}
+
+	return &m
+}
+
+// QueryVsanObjectUuidsByFilter returns vsan DOM object uuids by filter
+func (m HostVsanInternalSystem) QueryVsanObjectUuidsByFilter(ctx context.Context, uuids []string, limit int32, version int32) ([]string, error) {
```
vburenin: Why do you use value receiver instead of a pointer?

pkg/vsphere/vsan/synced_cache.go
```diff
@@ -0,0 +1,439 @@
...
+
+func init() {
+	go SyncedDomCache.run()
+}
+
+type DomDeleteError struct {
+	Err         error
+	FailedUuids []string
+	// if delete dom objects returns no error, but with some of uuid delete failure, this result will return detail error information for each uuid
+	Result []types.HostVsanInternalSystemDeleteVsanObjectsResult
```
vburenin: Why do you use slice of values instead of array of pointers?

pkg/vsphere/vsan/synced_cache.go
```diff
@@ -0,0 +1,439 @@
...
+	}
+}
+
+func (c *syncedDomCache) closeChannels() {
+	defer trace.End(trace.Begin(""))
+	close(c.refresh)
+	close(c.stop)
+	close(c.doneOnce)
+	close(c.waitCurrentRefresh)
+	c.m.Lock()
```
vburenin: where is unlock?

pkg/vsphere/vsan/synced_cache.go
```diff
@@ -0,0 +1,439 @@
...
+	hvis := NewHostVsanInternalSystem(ds.Client(), *h.ConfigManager.VsanInternalSystem)
+	dsc := &vsanDSDomCache{
+		ds:    ds,
+		hvis:  hvis,
+		uuids: make(map[string]string),
+		paths: make(map[string]string),
+		m:     &sync.Mutex{},
+	}
+	c.m.Lock()
+	defer c.m.Unlock()
```
vburenin: you may want avoid using defer here. since log and Refresh are potentially blocking operations which may block execution or even lead to a deadlock.

pkg/vsphere/vsan/synced_cache.go
```diff
@@ -0,0 +1,439 @@
...
+	return nil
+}
+
+func (c *syncedDomCache) getNonRefreshingDS(requests []*dsWaitGroup) []*dsWaitGroup {
+	defer trace.End(trace.Begin(fmt.Sprintf("%#v", requests)))
+
+	c.m.Lock()
+	defer c.m.Unlock()
+	var ret []*dsWaitGroup
+	for i := range requests {
```
vburenin: request is an array, so a range can provide you a value that you would not access by index.

pkg/vsphere/vsan/synced_cache.go
```diff
@@ -0,0 +1,439 @@
...
+	var errs []string
+	for i, dsw := range dsws {
+		dsc, ok := c.dsMap[dsw.dsRef]
+		if !ok {
+			err := errors.Errorf("requested datastore %s does not exist", c.dsMap[dsw.dsRef])
+			log.Error(err)
+			errs = append(errs, err.Error())
+			continue
+		}
+		if dsws[i].wg != nil {
```
vburenin: why do you access element by index here, meanwhile you access element as a range provides?

pkg/vsphere/vsan/dom_cache.go
```diff
@@ -0,0 +1,209 @@
...
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/vic/pkg/errors"
+	"github.com/vmware/vic/pkg/trace"
+)
+
+const (
+	vsanNamePattern = "vsan:[a-zA-Z0-9-]+/"
```
hmahmood: Should this be more lax? i.e. allow all characters?

caglar10ur:
2 comments;

First one is for @gigawhitlocks, looks like pullapprove rules are lacking some patterns as this PR requires no approvals.

Second one is about our approach. Are we sure this GC-like approach is what we want? Is that the only solution?

pkg/vsphere/vsan/synced_cache.go
```diff
@@ -0,0 +1,439 @@
...
+	refreshings map[string][]*sync.WaitGroup
+}
+
+func newDomCache() *syncedDomCache {
+	return &syncedDomCache{
+		dsMap:              make(map[string]DSDomCache),
+		m:                  &sync.RWMutex{},
+		refresh:            make(chan *dsWaitGroup),
+		stop:               make(chan bool, 1),
+		doneOnce:           make(chan string),
```
hmahmood: This probably needs to be buffered since we do write to this channel multiple values at a time.

pkg/vsphere/vsan/synced_cache.go
```diff
@@ -0,0 +1,439 @@
...
+
+type dsWaitGroup struct {
+	dsRef string
+	wg    *sync.WaitGroup
+}
+
+type syncedDomCache struct {
+	// datastore map - this cache can support multiple vsan datastores
+	dsMap map[string]DSDomCache
+	m     *sync.RWMutex
```
hmahmood: Don't think this needs to be a pointer.

pkg/vsphere/vsan/synced_cache.go
```diff
@@ -0,0 +1,439 @@
...
+						c.refreshings[w.dsRef] = append(c.refreshings[w.dsRef], w.wg)
+					} else {
+						log.Errorf("waiting request for datastore %s should have non-nil waitgroup.", w.dsRef)
+					}
+				} else {
+					log.Debugf("No refresh running for datastore %s, done waitgroup immediately", w.dsRef)
+					w.wg.Done()
+				}
+			}()
+		case <-tick:
```
hmahmood: So the refresh happens for the whole cache, instead of each datastore individually?

pkg/vsphere/vsan/synced_cache.go
```diff
@@ -0,0 +1,439 @@
...
+	dsc := &vsanDSDomCache{
+		ds:    ds,
+		hvis:  hvis,
+		uuids: make(map[string]string),
+		paths: make(map[string]string),
+		m:     &sync.Mutex{},
+	}
+	c.m.Lock()
+	defer c.m.Unlock()
+	c.dsMap[ds.Reference().String()] = dsc
```
hmahmood: Is there an upper limit to how many items can go into this map? Maybe having an LRU style cache would be beneficial here.

emlin:
> Second one is about our approach. Are we sure this GC-like approach is what we want? Is that the only solution?

For this part, I really did lots of research in this Sprint, and talked with vsan team, but with less help on that still. For me, the knowledge I got is that we cannot use vm delete together with all image vmdks, cause that will delete whole image disk chain, we cannot use VirtualDiskManager to delete disk chain, because it does not release parent node's non-deletable attribute, and the only option for us FileManager, does not delete DOM object while we delete one single VMDK file.
That's why I come up with this solution.

emlin:
>add vSAN object helpers - these should go into govmomi

The file vsan_host_internal_system.go should be in govmomi, but I'll leave Doug for that work.

>delete DOM entries along with VMDK - this should go into pkg/vsphere/datastore as we're already doing vSAN specific behaviour in that code.

the code to delete DOM entries is not there yet. I'm separate that part to another issue #3931 which will call this helper to delete.

>delete already leaked DOM objects - this could be provided as a subcommand to vic-machine debug or similar.

This is one GC work, is supposed to be done in issue #3932 

But right now I have govmomi code and vsan cache in same package, which caused confusion. But I also tend to avoid that in datastore package, because it's just a code to build cache, not related to datastore operations. So I'll have a separate package for this cache and govmomi code.

emlin:
> do not know the consequence of deleting DOM objects that do not have a disk associated with them. Is there any possibility that we're deleting an DOM object that was only just created and will shortly have an associated VMDK? What happens if we remove a DOM object where another tool has removed the VMDK and is about to remove the DOM object (e.g. vsphere performing a VM deletion) - will that tool handle it cleanly or will it result in a concurrency error?)

This is a valid concern, which might really happen. Initially I just try to figure out how to clean existing leaking completely.
But what we can safely do is to delete Dom objects whose vmdk is under specific directory, which can be used by Vic-machine delete.
I can add a new method to delete doms under specific file directory, to remove the concurrency issue you mentioned, if we decided to do it in this way.

For the cleanup of existing leaking requirement, that might have to be this way. Cause old VCH already leaked Dom, without any clue for us to figure out if that's leaking or not. 
I knew it's problematic before I started to work on that. itself is only draft interface, could works unexpectedly. I personally ever see it cannot delete some Dom objects, but after a day, it works again.
The idea is to have a separate command which clean Dom only. So user can choose when to do it.

pkg/vsphere/vsan/host_vsan_internal_system.go
emlin: this is returned from vsphere api call, it's already been value receiver

pkg/vsphere/vsan/synced_cache.go
emlin: same ^^^

pkg/vsphere/vsan/synced_cache.go
emlin: ah.. this is evil

pkg/vsphere/vsan/synced_cache.go
emlin: why refresh may block? it just writes to channel.

pkg/vsphere/vsan/synced_cache.go
emlin: that will copy. not a big difference, but anyway

pkg/vsphere/vsan/dom_cache.go
emlin: hmm... this is a good idea. I basically don't what can be here

pkg/vsphere/vsan/synced_cache.go
emlin: if timer timeout, it will refresh all datastores. But Refresh() method will trigger refresh for one particular datastore

pkg/vsphere/vsan/synced_cache.go
emlin: tbh, I don't think there will have too many vSan datastores in one VC env. Each cluster will only one vSan. So I didn't try to optimize this now

PR #3953, opened 2017-02-14
correct the display unit of cpu usage/limit in docker info [specific ci=1-01-Docker-Info]

### LINKED ISSUE #3950: Docker info reports Mhz, should be MHz, GHz.
https://en.wikipedia.org/wiki/Hertz

chengwang86:
Change it from Mhz to MHz.

Fix #3950


sflxn:
lgtm

PR #3963, opened 2017-02-15
[skip ci] Delegates in pullapprove

gigawhitlocks:
Still adding delegates..


mdubya66:
Is there a way to have a super delegate? if so that can be myself and if needed mike hagen and george

PR #3984, opened 2017-02-17
Pure go NFSv3 client

### LINKED ISSUE #3803: NFS client implementation for NFS VolumeStore

fdawg4l:
This is a pure `go` implementation of a subset of the NFS protocol.  This change implements the `AUTH_UNIX`, `portmapper` service requests, and part of the `NFS` `MOUNT` program.

Some of this code came from https://github.com/davecheney/nfs, but the implementation was non functional and incomplete.  I tried pushing these changes upstream, but DC mentioned the code is dead and said to maintain our own fork instead.

Fixes #3803


fdawg4l:
Getting guidance about the VMW copyright header since this is derived work.  We may drop that and find an alternative.  Stand by.

Issue #3996, opened 2017-02-21
docker stop reports wrong exitcode when process returns non-zero after SIGTERM received


gigawhitlocks:
**User Statement:**

I expect VIC to have `docker stop` functional parity. If I call `docker stop` on a process that traps the signal and returns a custom exit code, the container's exit code should be the same as the exit code of the child process.

**Details:**

The `ExecConfig` that's inspected by `docker ps` appears to lose this information, or maybe it's never assigned, in the case that the process receives a `SIGTERM`. I have been able to see the exit code get reported properly, but it either does not get written to ExtraConfig or gets overwritten with an invalid value (`0`) before the `stop` process is complete 

**Acceptance Criteria:**
These procedures should produce the same outputs on VIC and on Docker:
`docker run -it busybox`
`$ trap "exit 22" SIGTERM; while true; do sleep 1; done`

Then in a new shell, `docker stop $container_from_above`
Then check `docker ps -a` and see `Exited (22)` in the output.

hickeng:
The core issue here is fixed by #5532, however this needs a test to confirm the function.
I've edited the original description to use an error code that packs cleanly into the available bits - original was `1111` which overflows.

corrieb:
I believe we can close this as a dup of #5610 at this time

PR #3997, opened 2017-02-21
[specific ci=Group1-Docker-Commands] Partial fix for `docker stop` exit code reporting

### LINKED ISSUE #3800: docker stop not resulting in correct exit code

gigawhitlocks:
Fixes #3800  -- partial.

For SIGTERM'd processes where the child process returns exit code 0, we should be reporting container exit code 143 for parity with Docker. This was broken as described in the above issue, and is fixed by the single line change and protected via a new integration test, found below.

However, to get real parity with Docker, if the child process captures the SIGTERM and returns a custom exit code, that code should show up in `docker ps -a` -- currently this does not work, and this change doesn't make it work. I am not sure if we ever had parity w/ Docker on this. The exit code is reported properly if the process exits of its own accord.

Regarding real Docker parity, I've broken this final edge case out into its own issue ( https://github.com/vmware/vic/issues/3996 ) since it is roadblocking me and I think it's time to say that it goes outside of the sized scope of issue #3800 considering that it appears to arise from a single value not getting set _somewhere_ where it should get set, and I've spent too much time digging around for it already for a single size 3 issue.


cgtexmex:
FTR - if a container returned a custom exit code we previously showed that exit code in `docker ps`, so we did have parity at one time.

gigawhitlocks:
We still do if the process exits on its own. Do you know if we ever did in the case that the process received a SIGTERM @cgtexmex ? Subtly different scenario

cgtexmex:
@gigawhitlocks ahhh...I don't know..I thought if a _SIGTERM_ was received it exited with the standard code...I guess I was wrong..

PR #4004, opened 2017-02-22
Design doc for container rename [skip ci]

### LINKED ISSUE #380: Design Container Rename
As a VIC user, I should be able to rename a running container
As a docker compose user, I should be able to rename a running service.

---
https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#rename-a-container
https://docs.docker.com/engine/reference/commandline/rename/

chengwang86:
Fixes #380

### DIFF
--- a/doc/design/docker-rename.md
+++ b/doc/design/docker-rename.md
@@ -8,7 +8,7 @@ Rename involves the containerVM’s display name and the name of the datastore f
 
 ### Proposed solution:
 
-- We use containerName-containerShortID to assemble the VM name. We do not use containerName-containerID as the VM name in order to avoid the scenario wherein the containerName gets truncated to satisfy the maximum length of a VM name in vSphere. In addition, we use the containerID to set the name of the datastore folder, thus there is no need to worry about the VM display name and datastore folder name being inconsistent. 
+- We use containerName-containerShortID to assemble the VM display name. We do not use containerName-containerID in order to avoid the scenario wherein the containerName gets truncated to satisfy the maximum length of a VM display name in vSphere. In addition, we use the containerID to set the name of the datastore folder, thus there is no need to worry about the VM display name and datastore folder name being inconsistent. 
 
 - VM Reconfiguration: Since vSAN requires the VM display name to be the same as the datastore folder name during VM creation, we set both the VM display name and the datastore folder name to containerID during VM creation, and then start a VM reconfiguration task after the VM is created to change its display name to containerName-containerShortID. 
 
--- a/doc/design/docker-rename.md
+++ b/doc/design/docker-rename.md
@@ -13,7 +13,7 @@ Risks might also arise in the underlying vSAN if the storage folder is renamed.
 
 |     |  S1  |  S2  |  S3  |
 |  ---  |  ---  |  ---  |  ---  |
-|  VM name on vSphere UI  |  Update using govmomi  |
+|  VM name on vSphere UI  |  Update using govmomi  |  Update using govmomi  |  Update using govmomi  |
 |  VM folder name  |  Do not update. The vi admin can easily find the VM folder path/name from the UI of the containerVM.  |  (1) Add the new name to the container metadata while keeping the old name. (2) docker inspect should show both the original name and the new name so that the vi admin/developers can easily for the VM folder using the original name. (3) When the container exits,  shuts down or restart, the portlayer commits the new name so that the VM folder name can be updated.  |  When creating the containerVM, only use containerID as the folder name. Therefore, there is no need to update the VM folder name during docker rename.  |
 
 ### Pros and Cons:
--- a/doc/design/docker-rename.md
+++ b/doc/design/docker-rename.md
@@ -10,14 +10,14 @@ Rename involves the containerVM’s display name and the name of the datastore f
 
 - We use `containerName-containerShortID` to assemble the VM display name. We do not use containerName-containerID in order to avoid the scenario wherein the containerName gets truncated to satisfy the maximum length of a VM display name in vSphere. In addition, we use the `containerID` to set the name of the datastore folder, thus there is no need to worry about the VM display name and datastore folder name being inconsistent. 
 
-- VM creation: 
+- VM configuration during creation: 
 
   - vSAN: Since vSAN requires the VM display name to be the same as the datastore folder name during VM creation, we set both the VM display name and the datastore folder name to `containerName-containerShortID` during VM creation.
   - Non-vSAN: We set the VM display name to `containerName-containerShortID` and set the datastore folder name to `containeriD`.
+  - VM guestinfo: We replace `guestinfo.vice./common/name` with `common/name` in the `ExtraConfig` of a containerVM and set the scope of `common/name` to `hidden`. Then when `docker rename` is called after the VM is created, we can update the `common/name` with the new name. By doing this, we can persist the new name for a running containerVM.  
 
 - Docker support for rename: When a customer calls `docker rename`, we update the VM display name to the new name in both the docker persona and the portlayer. 
-
-  - VM guestinfo: We replace `guestinfo.vice./common/name` with `common/name` in the `ExtraConfig` of a containerVM and set the scope of `common/name` to `hidden`. When `docker rename` is called, we update the `common/name` with the new name. By doing this, we can persist the new name for a running containerVM.  
+  
   - Network: 
 
     - Network alias should be updated.
--- a/doc/design/docker-rename.md
+++ b/doc/design/docker-rename.md
@@ -4,28 +4,33 @@ This is the design proposal for implementing “docker rename” on vSphere Inte
 
 ### The problem:
 
-Rename involves both the containerVM’s display name and the name of the storage folder (and the files in the folder) of the containerVM on the vsphere UI, as well as in the container cache of the VIC Engine. Currently both the containerVM's display name and the folder name are created during VM creation using containerName-containerID, which is done for better user experience (the vi admin would be able to find the containerVM and the storage folder easily from the vsphere UI). Renaming the VM display name on the UI can be easily achieved by using govc, which However does not update the storage folder name. In this case, the vi admin would observe inconsistent VM display name and storage folder name, and it becomes difficult the admin to reference the storage folder based on the new VM display name.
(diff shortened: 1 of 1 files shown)


doc/design/docker-rename.md
```diff
@@ -0,0 +1,36 @@
...
+
+- VM Reconfiguration: Since vSAN requires the VM display name to be the same as the storage folder name during VM creation, we set both the VM display name and the storage folder name to containerID during VM creation, and then start a VM reconfiguration task after the VM is created to change its display name to containerName-containerID. 
+
+- Docker support for rename: When a customer calls `docker rename`, we update the VM display name to the new name in both the docker persona and the portlayer. 
+
+  - Network: 
+
+    - Network alias should be updated.
+    - If `--link` is used when creating the container, HostConfig of relevant containers should be updated.
+    - The containerName shown in `/etc/host` within the container should be updated (Proposal: Remove containerName from `/etc/host`; docker does not put the containerName in this file).
```
hmahmood: +1 on removing the name from /etc/host

hmahmood: What about existing containers that have the name in /etc/hosts? How are we going to update them?

chengwang86: @hmahmood If we decide to remove all the container names from /etc/hosts, this issue does not exist, am I right?

emlin: this should be a backward compatibility issue. VCH can be upgraded from older version, then the older version's container already has the name in /etc/hosts.

doc/design/docker-rename.md
```diff
@@ -0,0 +1,36 @@
...
+Robot scripts will be written to test the following:
+
+1. VM reconfiguration:
+  - After a containerVM is created, use govc to check the display name (containerName-containerID) and storage folder name (containerID).
+
+2. Docker support for rename:
+  - The basic functionality of `docker rename`
+  - Check validity of network alias and HostConfig 
+  - `docker-compose up` when there are existing containers for the same service but the configuration or image has been changed
+  - `docker-compose up –force-recreate` when there are existing containers for the same service even if the configuration or image has not been changed
```
emlin: Concur the comments above, you will need one backward compatibility test to see if this works for older version's container.

doc/design/docker-rename.md
```diff
@@ -0,0 +1,36 @@
+This is the design proposal for implementing “docker rename” on vSphere Integrated Containers Engine (VIC Engine).
+
+## Design Proposal
+
+### The problem:
+
+Rename involves the containerVM’s display name and the name of the storage folder (and the files in the folder) of the containerVM on the vsphere UI, the container cache of the VIC Engine and the container network configuration (e.g., network alias). Currently both the containerVM's display name and the folder name are created during VM creation using containerName-containerID, which is done for matching the container information obtained from `docker ps` to the VM displayed on the vSphere UI. Renaming the VM display name on the UI can be easily achieved by using govc, which However does not update the storage folder name. In this case, the vi admin would observe inconsistent VM display name and storage folder name, and it becomes difficult the admin to reference the storage folder based on the new VM display name.
```
fdawg4l: `Renaming the VM display name... using govc, which However...`

doc/design/docker-rename.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Proposed solution:
+
+- We still use containerName-containerID to assemble the VM name. However, we only use the containerID to set the name of the VM storage folder; therefore, there is no need to worry about the VM display name and storage folder name being inconsistent. 
+
+- VM Reconfiguration: Since vSAN requires the VM display name to be the same as the storage folder name during VM creation, we set both the VM display name and the storage folder name to containerID during VM creation, and then start a VM reconfiguration task after the VM is created to change its display name to containerName-containerID.
```
fdawg4l: This sounds strange.  Do you see the same behavior for NFS datastores with VC?  The reason I ask is I want to avoid any `if IsVsan()` conditions where possible so we can have a unified behavior across datastore types.

doc/design/docker-rename.md
fdawg4l: @hmahmood can the hostname be updated via DHCP host field pushed down by the server? Then we can modify the dhclient (or whatever we use) to set the hostname on lease renewal.

doc/design/docker-rename.md
```diff
@@ -0,0 +1,36 @@
...
+
+- Docker support for rename: When a customer calls `docker rename`, we update the VM display name to the new name in both the docker persona and the portlayer. 
+
+  - Network: 
+
+    - Network alias should be updated.
+    - If `--link` is used when creating the container, HostConfig of relevant containers should be updated.
+    - The containerName shown in `/etc/host` within the container should be updated (Proposal: Remove containerName from `/etc/host`; docker does not put the containerName in this file).
+  
+  - Storage: Nothing needs to be updated if we set the storage folder name to containerID.
```
fdawg4l: True.

doc/design/docker-rename.md
chengwang86: @fdawg4l The vSAN related information is from George. But I don't think we need to check `if IsVsan()` in the vm reconfiguration step. My proposal is: we do VM reconfiguration regardless of the underlying storage type. Does this answer your concern?

doc/design/docker-rename.md
hmahmood: Currently it cannot be pushed down via DHCP. Even if it was, it is orthogonal to the problem here; we insert the hostname on tether startup in /etc/hosts.

doc/design/docker-rename.md
hickeng: @fdawg4l In prior testing (last year) the `files.vmPathName` is completely ignored and the displayName is used instead. I don't know why.

Regarding the reconfigure - we are aiming for an idempotent model, both at a high level and for low level operations - I'd suggest we structure this so that we're calling vm.Rename or similar so there's a place that can determine whether a vsphere operation is needed or not @cgtexmex this is the kind of thing I'm hoping to have vmomi gateway do to add clarity to the calling code.

doc/design/docker-rename.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Design Proposal
+
+### The problem:
+
+Rename involves the containerVM’s display name and the name of the storage folder (and the files in the folder) of the containerVM on the vsphere UI, the container cache of the VIC Engine and the container network configuration (e.g., network alias). Currently both the containerVM's display name and the folder name are created during VM creation using containerName-containerID, which is done for matching the container information obtained from `docker ps` to the VM displayed on the vSphere UI. Renaming the VM display name on the UI can be easily achieved by using govc, which however does not update the storage folder name. In this case, the vi admin would observe inconsistent VM display name and storage folder name, and it becomes difficult the admin to reference the storage folder based on the new VM display name.
+
+### Proposed solution:
+
+- We still use containerName-containerID to assemble the VM name. However, we only use the containerID to set the name of the VM storage folder; therefore, there is no need to worry about the VM display name and storage folder name being inconsistent.
```
hickeng: I prefer this - there was a reason that it wasn't being done but I don't recall what.
@dougm I *think* this was related only to some of the out-of-band govc testing - do you recall?

If we are doing this I'd prefer that we take the opportunity to move to using `name-shortid` so that we're no longer truncating the human readable name. Ideally we implement this by using a template string as per the constants defined in `lib/config/virtual_container_host.go:35` (add a shortid token). This provides for future requirements of allowing viadmin to apply a naming convention to containerVMs.

doc/design/docker-rename.md
```diff
@@ -0,0 +1,39 @@
...
+- We still use containerName-containerID to assemble the VM name. However, we only use the containerID to set the name of the VM storage folder; therefore, there is no need to worry about the VM display name and storage folder name being inconsistent. 
+
+- VM Reconfiguration: Since vSAN requires the VM display name to be the same as the storage folder name during VM creation, we set both the VM display name and the storage folder name to containerID during VM creation, and then start a VM reconfiguration task after the VM is created to change its display name to containerName-containerID. 
+
+- Docker support for rename: When a customer calls `docker rename`, we update the VM display name to the new name in both the docker persona and the portlayer. 
+
+  - Network: 
+
+    - Network alias should be updated.
+    - If `--link` is used when creating the container, HostConfig of relevant containers should be updated.
```
hickeng: HostConfig should be synthetic anyway and derived from the backend data - this should not need actual code, but is a good test.

doc/design/docker-rename.md
```diff
@@ -0,0 +1,39 @@
...
+    - The containerName shown in `/etc/host` within the container should be updated. Proposal: Remove containerName from `/etc/host`; docker does not put the containerName in this file. For backward compatibility, if we rename a container which is created with containerName in its `/etc/host`, we use the VMware RPC API's guestinfo facility to update `/etc/host`.
```
hickeng: `/etc/host` is generated - rename is a tether config update and should trigger a Reload - that should cause the network config to be inspected and changes to be applied.

We should *not* be bypassing the tether for this sort of change if it doesn't currently work as described, but fixing tether.
We *can* use the RPC mechanism to trigger a Reload in tether however
@caglar10ur please comment if needed given tether plans.

doc/design/docker-rename.md
```diff
@@ -0,0 +1,39 @@
...
+  
+  - Storage: Nothing needs to be updated if we set the storage folder name to containerID.
```
hickeng: please be precise with the nomenclature here - `datastore folder name to container ID`.

emlin:
lgtm for the upgrade part

doc/design/docker-rename.md
```diff
@@ -0,0 +1,42 @@
...
+
+- Docker support for rename: When a customer calls `docker rename`, we update the VM display name to the new name in both the docker persona and the portlayer. 
+
+  - Network: 
+
+    - Network alias should be updated.
+    - If `--link` is used when creating the container, HostConfig of relevant containers should be automatically updated based on the backend data.
+    - For the the containerName shown in `/etc/hosts` within the container: 
+      - Remove containerName from `/etc/hosts` for new containers
+      - Backward compatibility for containers created by a VCH of an older version: No matter the container is powered off or still powered on, change the network setting if needed and also use govmomi to update the name in the containerVM's `guestinfo.vice./common/name`. `/etc/hosts` will be automatically regenerated with the new containerName when the containerVM is rebooted.
```
hickeng: Updating `guestinfo.vice./common/name` with the container powered on will cause that key to become non-persistent. This is not an issue if the name is being set in `common/name` (the updated key) as well, but is problematic if not.

Please update this statement to describe precisely which keys will be set, and where the name will be pulled from by the portlayer in the case of the old key becoming non-persistent.

PR #4008, opened 2017-02-22
[specific ci=Group11-Upgrade]Support data migration in vic-machine update

### LINKED ISSUE #3609: vic-machine upgrade: include data migration
As a vic admin, after create VCH through vic-machine create, and new vic engine binary is available, I want to update existing VCH to latest version, without impact my current workload, which means my containers.
I don't care if underline vic engine configuration is changed or not, it should succeed if all env is correctly prepared, as long as the old version of existing VCH is still supported.
If there is anything wrong, the upgrade should be reverted and old workload should still run well. And I should get well defined error message, and good guideline to fix the problem.

At this moment, short time service unavailable during upgrade is acceptable.

emlin:
Fixes #3609

Also did a bit cleanup for timeout. In this way, timeout for vic-machine create and vic-machine upgrade will work for service waiting only. 
If iso file uploading or snapshot take long time, it would not timeout.

If upgrade timeout to wait service, it will rollback to old version cause it cannot make sure if there is anything real wrong or just a timing issue. And if that's really a timing issue, the roll back will fail to wait service ready, but it will try to finish all cleanup with a context without timeout.

@stuclem

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -1369,7 +1369,7 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 	defer func() {
 		if ctx.Err() != nil && ctx.Err() == context.DeadlineExceeded {
 			//context deadline exceeded, replace returned error message
-			err = errors.Errorf("Create timed out: if slow connection, increase timeout with --timeout")
+			err = errors.Errorf("Creating VCH exceeded time limit of %s. Please increase the timeout using --timeout to accommodate for a busy vSphere target")
 		}
 	}()
 
@@ -1377,7 +1377,11 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 		executor.CollectDiagnosticLogs()
 		cmd, _ := executor.GetDockerAPICommand(vchConfig, c.ckey, c.ccert, c.cacert)
 		log.Info("\tAPI may be slow to start - try to connect to API after a few minutes:")
-		log.Infof("\t\tRun command: %s", cmd)
+		if cmd != "" {
+			log.Infof("\t\tRun command: %s", cmd)
+		} else {
+			log.Infof("\t\tRun %s inspect to find API connection command and run the command if ip address is ready", clic.App.Name)
+		}
 		log.Info("\t\tIf command succeeds, VCH is started. If command fails, VCH failed to install - see documentation for troubleshooting.")
 		return err
 	}
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -43,7 +43,6 @@ import (
 	"github.com/vmware/vic/pkg/flags"
 	"github.com/vmware/vic/pkg/ip"
 	"github.com/vmware/vic/pkg/trace"
-	"github.com/vmware/vic/pkg/vsphere/diag"
 )
 
 const (
@@ -1330,15 +1329,8 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 		return errors.New("invalid CLI arguments")
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), c.Timeout)
-	defer cancel()
-	defer func() {
-		if ctx.Err() != nil && ctx.Err() == context.DeadlineExceeded {
-			//context deadline exceeded, replace returned error message
-			err = errors.Errorf("Create timed out: if slow connection, increase timeout with --timeout")
-		}
-	}()
-
+	// all these operations will be executed without timeout
+	ctx := context.Background()
 	validator, err := validate.NewValidator(ctx, c.Data)
 	if err != nil {
 		log.Error("Create cannot continue: failed to create validator")
@@ -1359,6 +1351,8 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 	vConfig.HTTPProxy = c.HTTPProxy
 	vConfig.HTTPSProxy = c.HTTPSProxy
 
+	vConfig.Timeout = c.Data.Timeout
+
 	// separate initial validation from dispatch of creation task
 	log.Info("")
 
@@ -1369,36 +1363,23 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 		return err
 	}
 
-	// vic-init will try to reach out to the vSphere target.
-	log.Info("Checking VCH connectivity with vSphere target")
-	vch, err := executor.NewVCHFromComputePath(c.Data.ComputeResourcePath, c.Data.DisplayName, validator)
-	if err != nil {
-		log.Warningf("Failed to get Virtual Container Host %s. Error: %v", c.Data.DisplayName, err)
-	} else {
-		// Checking access to vSphere API
-		if cd, err := executor.CheckAccessToVCAPI(ctx, vch, vchConfig.Target); err == nil {
-			code := int(cd)
-			if code > 0 {
-				log.Warningf("vSphere API Test: %s %s", vchConfig.Target, diag.UserReadableVCAPITestDescription(code))
-			} else {
-				log.Infof("vSphere API Test: %s %s", vchConfig.Target, diag.UserReadableVCAPITestDescription(code))
-			}
-		} else {
-			log.Warningf("Could not run VCH vSphere API target check due to %v", err)
+	// timeoout start to work from here, to make sure user does not wait forever
+	ctx, cancel := context.WithTimeout(context.Background(), c.Timeout)
+	defer cancel()
+	defer func() {
+		if ctx.Err() != nil && ctx.Err() == context.DeadlineExceeded {
+			//context deadline exceeded, replace returned error message
+			err = errors.Errorf("Create timed out: if slow connection, increase timeout with --timeout")
 		}
-	}
+	}()
 
-	// check the docker endpoint is responsive
-	log.Info("Checking Docker API endpoint")
-	if err = executor.CheckDockerAPI(vchConfig, c.clientCert); err != nil {
+	if err = executor.CheckServiceReady(ctx, vchConfig, c.clientCert); err != nil {
 		executor.CollectDiagnosticLogs()
 		cmd, _ := executor.GetDockerAPICommand(vchConfig, c.ckey, c.ccert, c.cacert)
-		msg := fmt.Sprintf("Docker API endpoint check failed: %s", err)
-		log.Info(msg)
 		log.Info("\tAPI may be slow to start - try to connect to API after a few minutes:")
 		log.Infof("\t\tRun command: %s", cmd)
 		log.Info("\t\tIf command succeeds, VCH is started. If command fails, VCH failed to install - see documentation for troubleshooting.")
-		return fmt.Errorf(msg)
+		return err
 	}
 
 	log.Infof("Initialization of appliance successful")
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -1369,7 +1369,7 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 	defer func() {
 		if ctx.Err() != nil && ctx.Err() == context.DeadlineExceeded {
 			//context deadline exceeded, replace returned error message
-			err = errors.Errorf("Creating VCH exceeded time limit of %s. Please increase the timeout using --timeout to accommodate for a busy vSphere target")
+			err = errors.Errorf("Creating VCH exceeded time limit of %s. Please increase the timeout using --timeout to accommodate for a busy vSphere target", c.Timeout)
 		}
 	}()
 
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -1367,7 +1367,7 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 	ctx, cancel := context.WithTimeout(context.Background(), c.Timeout)
 	defer cancel()
 	defer func() {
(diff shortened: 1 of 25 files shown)


emlin:
@gigawhitlocks @matthewavery suppose this will impact your work more or less.

cmd/vic-machine/delete/delete.go
```diff
@@ -155,13 +155,12 @@ func (d *Uninstall) Run(clic *cli.Context) (err error) {
 	installerBuild := version.GetBuild()
 	if vchConfig.Version == nil || !installerBuild.Equal(vchConfig.Version) {
 		if !d.Data.Force {
-			log.Errorf("VCH version %q is different than installer version %s. Specify --force to force delete", vchConfig.Version.ShortVersion(), installerBuild.ShortVersion())
+			log.Errorf("VCH version %q is different than installer version %s. Upgrade VCH before delete or specify --force to force delete or ", vchConfig.Version.ShortVersion(), installerBuild.ShortVersion())
```
andrewtchin: is the ending `or` a typo?

emlin: a typo

lib/install/management/delete.go
```diff
@@ -120,14 +120,14 @@ func (d *Dispatcher) getComputeResource(vmm *vm.VirtualMachine, conf *config.Vir
...
 	var err error
-	if len(conf.ImageStores) == 0 {
-		if !d.force {
+	if conf == nil || len(conf.ImageStores) == 0 {
+		if !force {
 			err = errors.Errorf("Cannot find image stores from configuration")
 			return nil, err
 		}
-		log.Warnf("Cannot find image stores from configuration, attempting to delete from vm datastore")
+		log.Debugf("Cannot find image stores from configuration, attempting to find from vm datastore")
```
gigawhitlocks: This is really minor, and I'm pretty sure I wrote this log message originally so this is actually my error, but can we change that comma to a semicolon please? It's grammatically incorrect as it is and it's bugging me ;D

lib/install/management/upgrade.go
```diff
@@ -77,47 +77,38 @@ func (d *Dispatcher) Upgrade(vch *vm.VirtualMachine, conf *config.VirtualContain
...
+	d.deleteSnapshot(snapshotName, conf.Name)
+	// return the error message for upgrade
 	return err
 }
 
-func (d *Dispatcher) deleteSnapshot(id types.ManagedObjectReference, snapshotName string, applianceName string) error {
+func (d *Dispatcher) deleteSnapshot(snapshotName string, applianceName string) error {
 	defer trace.End(trace.Begin(snapshotName))
 	log.Infof("Deleting upgrade snapshot %q", snapshotName)
 	// do clean up aggressively, even the previous operation failed with context deadline excceeded.
```
gigawhitlocks: everywhere this comment appears, there's a typo in `exceeded`

gigawhitlocks:
Also: is there a reason that we have `lib/install/management/upgrade.go` containing the `Dispatcher` code and the validator code is in `lib/install/validate/update.go` ? Can we get consistent naming on these? Maybe rename `validate/update.go` to `validate/upgrade.go`? Or are these named upgrade/update intentionally?

cmd/vic-machine/create/create.go
```diff
@@ -1369,36 +1363,27 @@ func (c *Create) Run(clic *cli.Context) (err error) {
...
-			} else {
-				log.Infof("vSphere API Test: %s %s", vchConfig.Target, diag.UserReadableVCAPITestDescription(code))
-			}
-		} else {
-			log.Warningf("Could not run VCH vSphere API target check due to %v", err)
+	// timeoout start to work from here, to make sure user does not wait forever
+	ctx, cancel := context.WithTimeout(context.Background(), c.Timeout)
+	defer cancel()
+	defer func() {
+		if ctx.Err() != nil && ctx.Err() == context.DeadlineExceeded {
```
anchal-agrawal: Isn't `ctx.Err() != nil` redundant?

lib/config/virtual_container_host.go
```diff
@@ -214,6 +216,16 @@ func (t *VirtualContainerHostConfigSpec) SetMoref(moref *types.ManagedObjectRefe
 	}
 }
 
+// SetIsCreating sets the ID of the VCH to a constant, to identify the creating VCH VM before the VM moref can be set into this property
+func (t *VirtualContainerHostConfigSpec) SetIsCreating() {
```
sflxn: You have a SetIsCreating() but no offsetting function.  Once you set it to creating, it's set like that for good?

cmd/vic-machine/delete/delete.go
```diff
@@ -155,13 +155,12 @@ func (d *Uninstall) Run(clic *cli.Context) (err error) {
 	installerBuild := version.GetBuild()
 	if vchConfig.Version == nil || !installerBuild.Equal(vchConfig.Version) {
 		if !d.Data.Force {
-			log.Errorf("VCH version %q is different than installer version %s. Specify --force to force delete", vchConfig.Version.ShortVersion(), installerBuild.ShortVersion())
+			log.Errorf("VCH version %q is different than installer version %s. Upgrade VCH before delete or specify --force to force delete", vchConfig.Version.ShortVersion(), installerBuild.ShortVersion())
```
anchal-agrawal: s/delete/deleting

lib/config/virtual_container_host.go
gigawhitlocks: No, she sets `t.ExecutorConfig.ID` to this constant stand-in value and it eventually gets overwritten with an actual ID, I believe. So it's set like that until that field receives a real value, at which point the ContainerVM is created.

I think it might be better to use a dedicated bool rather than to overload the `ExecutorConfig.ID` field's use like this, though, to prevent the confusion that @sflxn is having.

lib/install/management/upgrade.go
```diff
@@ -77,47 +77,38 @@ func (d *Dispatcher) Upgrade(vch *vm.VirtualMachine, conf *config.VirtualContain
...
-	d.ctx, cancel = context.WithTimeout(context.Background(), settings.RollbackTimeout)
-	defer cancel()
-
-	if rerr := d.rollback(conf, snapshotName); rerr != nil {
+	if rerr := d.rollback(conf, snapshotName, settings); rerr != nil {
 		log.Errorf("Failed to revert appliance to snapshot: %s", rerr)
-		// return the error message for upgrade, instead of rollback
 		return err
 	}
+	log.Infof("Appliance is rollbacked to old version")
```
anchal-agrawal: s/rollbacked/rolled back

lib/install/management/upgrade.go
```diff
@@ -204,10 +180,25 @@ func (d *Dispatcher) update(conf *config.VirtualContainerHostConfigSpec, setting
...
+
+	ctx, cancel := context.WithTimeout(d.ctx, settings.Timeout)
+	defer cancel()
+	if err = d.CheckServiceReady(ctx, conf, nil); err != nil {
+		if ctx.Err() != nil && ctx.Err() == context.DeadlineExceeded {
+			//context deadline exceeded, replace returned error message
+			err = errors.Errorf("Upgrading VCH exceeded time limit of %s. Please increase the timeout using --timeout to accommodate for a busy vSphere target", settings.Timeout)
+		}
+
+		log.Info("\tAPI may be slow to start - might retry with increased timeout using --timeout: %s", err)
```
anchal-agrawal: s/might/please

lib/install/management/delete_test.go
```diff
@@ -89,12 +89,37 @@ func TestDelete(t *testing.T) {
 		createAppliance(ctx, validator.Session, conf, installSettings, false, t)
 
 		testNewVCHFromCompute(input.ComputeResourcePath, input.DisplayName, validator, t)
+		//		testUpgrade(input.ComputeResourcePath, input.DisplayName, validator, installSettings, t) TODO: does not implement: CreateSnapshot_Task
```
sflxn: This is commented out, and I don't see anything else calling testUpgrade().

lib/install/management/dispatcher.go
```diff
@@ -154,6 +154,57 @@ func (d *Dispatcher) InitDiagnosticLogs(conf *config.VirtualContainerHostConfigS
 	m := diagnostic.NewDiagnosticManager(d.session)
 
 	for k, l := range diagnosticLogs {
+		if l == nil {
+			continue
+		}
+		// get LineEnd without any LineText
+		h, err := m.BrowseLog(d.ctx, l.host, l.key, math.MaxInt32, 0)
+
```
sflxn: remove blank line

lib/install/management/delete_test.go
emlin: good point, but there is one TODO item at the end of the line.
I cannot un-comment it at this time for vcsim still does not have the function support for snapshot, and I want to leave some hints there for what we should test using vcsim.

lib/install/management/delete.go
emlin: sure

emlin:
> Also: is there a reason that we have lib/install/management/upgrade.go containing the Dispatcher code and the validator code is in lib/install/validate/update.go ? Can we get consistent naming on these? Maybe rename validate/update.go to validate/upgrade.go? Or are these named upgrade/update intentionally?

For this file name, I tends to have it to be used by vic-machine update command as well. We should have similar workflow for upgrade and update, cause technically upgrade just changed the iso files, and update can do more.
But right now, we still do not have update designed yet.

lib/install/management/upgrade.go
```diff
@@ -233,7 +224,17 @@ func (d *Dispatcher) ensureRollbackReady(conf *config.VirtualContainerHostConfig
...
-	return d.startAppliance(conf)
+	if err = d.startAppliance(conf); err != nil {
+		return err
+	}
+
+	ctx, cancel := context.WithTimeout(d.ctx, settings.Timeout)
+	defer cancel()
+	if err = d.CheckServiceReady(ctx, conf, nil); err != nil {
+		// do not return error in this case, to make sure clean up continues
+		log.Info("\tAPI may be slow to start - try to connect to API after a few minutes")
```
sflxn: You don't need to check if err == deadline exceeded here?  you do so in other places.

emlin: the first place to check context deadline is to let user know the upgrade failure caused by timeout.
here, even we got timeout, user should just try to connect service few minutes later, cause this is for rollback only.
if rollback failed, we don't actually can do anything. And if the service works fine, rollback should as well.  I don't need to remind them retry with increased timeout.

lib/install/management/finder.go
```diff
@@ -152,6 +155,44 @@ func (d *Dispatcher) GetVCHConfig(vm *vm.VirtualMachine) (*config.VirtualContain
...
+}
+
+func (d *Dispatcher) FetchAndMigrateVCHConfig(vm *vm.VirtualMachine) (*config.VirtualContainerHostConfigSpec, error) {
+	defer trace.End(trace.Begin(""))
+
+	//this is the appliance vm
+	mapConfig, err := vm.FetchExtraConfigBaseOptions(d.ctx)
+	if err != nil {
+		err = errors.Errorf("Failed to get VM extra config of %q: %s", vm.Reference(), err)
+		log.Error(err)
```
anchal-agrawal: Since `FetchAndMigrateVCHConfig`'s caller logs the error too, we can skip logging it in this func.

lib/install/management/finder.go
```diff
@@ -152,6 +155,44 @@ func (d *Dispatcher) GetVCHConfig(vm *vm.VirtualMachine) (*config.VirtualContain
...
 	}
 
+	if vchConfig.IsCreating() {
+		vmRef := vm.Reference()
+		vchConfig.SetMoref(&vmRef)
+	}
+	return vchConfig, nil
+}
+
+func (d *Dispatcher) FetchAndMigrateVCHConfig(vm *vm.VirtualMachine) (*config.VirtualContainerHostConfigSpec, error) {
```
anchal-agrawal: We should add a function description comment here for documentation.

lib/install/management/finder.go
```diff
@@ -152,6 +155,44 @@ func (d *Dispatcher) GetVCHConfig(vm *vm.VirtualMachine) (*config.VirtualContain
...
+	data := extraconfig.MapSource(newMap)
+	vchConfig := &config.VirtualContainerHostConfigSpec{}
+	result := extraconfig.Decode(data, vchConfig)
+	if result == nil {
+		err = errors.Errorf("Failed to decode migrated VM configuration %q: %s", vm.Reference(), err)
+		log.Error(err)
+		return nil, err
+	}
+
 	//	vchConfig.ID
```
anchal-agrawal: Unneeded comment? I realize this wasn't added as part of your changes but we should remove it anyway.

lib/install/management/upgrade.go
```diff
@@ -77,97 +79,112 @@ func (d *Dispatcher) Upgrade(vch *vm.VirtualMachine, conf *config.VirtualContain
...
+		log.Infof("Is soap vim fault")
+		switch soap.ToVimFault(err).(type) {
+		case *types.SystemError:
+			return true
+		}
 	}
-	log.Error(err)
-	return nil, err
+
+	if terr, ok := err.(task.Error); ok {
```
anchal-agrawal: Nit-pick: since we're only looking for one `case` in all type switch assertions in this func, we can replace them with `if` assertions like on line 136.

anchal-agrawal:
A few minor comments. LGTM.

PR #4013, opened 2017-02-23
[skip ci] Initial draft design document for containerd integration work

### LINKED ISSUE #3659: Produce high-level design doc for containerd / Port Layer integration
As an architect, I need to understand and document how we can use containerd as a personality/shim to our Port Layer. 

The scope of work is to assume "ctr" is the client and treat the containerd as proposed today as a Port Layer personality. The output of the work is as follows:

- Propose integration points for containerd with the Port Layer. Our working assumption is that there will be enough extension points in the containerd codebase for us to work with. This is the desire of the containerd developers too. New or modified extension points if needed will be proposed.

- By the end of the sprint, we should have a draft design doc that we can present that goes through some simple workflows: Eg. pull, create and start and highlights the extension points.

- One important question that should be decided by the end of the sprint is whether it makes sense for us to try to emulate runc or not. We should have a justification for the decision.

There will inevitably be some overlap here with #3634 and #3630

corrieb:
Fixes #3659

This is an early draft proposal for containerd integration and contains as many questions as answers. This should be updated fairly regularly as we make progress.


hickeng:
@corrieb Is this ready for review or here primarily for sharing?

corrieb:
@hickeng Neither really. It's a draft that I want to be able to make public and make changes to in public. The committed output of my original sprint was a draft doc. There's still stuff to figure out.

doc/design/containerd/port-layer-integration.md
```diff
@@ -0,0 +1,276 @@
...
+
+*(Note that the VIC appliance local filesystem is intentionally ephemeral, so no data persisted on the local filesystem is permanent. We must assume that operations like upgrade or other control plane outages should have no impact on the running containers or image cache and as such, there must be a means for containerd to learn about existing state when it starts up).
+
+*Executor*
+
+The executor interface is responsible both for all aspects of container lifecycle, including interaction with the container - (stdout, stderr, stdin and tty).
+
+As it stands, the interface expect a single out-of-process instance of an executor per container and that the executor will keep interaction channels open if containerd is restarted. This is something of a mismatch with our existing Port Layer which is a single service offering async API calls for running and interacting with containers.
+
+This presents two options: Fake up shim processes that represent each container or add an ID field to the Executor interface. The latter approach is our preference for the main reason that shim lifecycle management becomes a pain in the event that the VICe endpoint VM is restarted.
```
vburenin: Faking up a shim process is the best approach, since it will result in much smaller memory footprint as well as it will scale better in case of many running containers. The only thing I am a little worried about is a number of file sockets we will need to keep opened. In case of containerd it is a little simpler since one host is not going to run hundreds of containers, in case of VIC engine it may run hundreds of VMs so it will result in 4 times more file sockets up and running.

doc/design/containerd/port-layer-integration.md
```diff
@@ -0,0 +1,276 @@
...
+
+*Metadata store*
+
+Metadata is what ties everything together. It will need to link image URIs to digests and links to blobs and snapshots. Given that VIC does not have a persistent filesystem layer, metadata must be persisted into vSphere in some way. 
+
+The question of how this should be done remains open. The KV store on the Port Layer is likely not rich enough to represent the data relationships, so an alternative needs to be sought.
+
+*Events*
+
+At present, the events interface is one of the least clearly defined. However, it’s not hard to imagine how events from vSphere could propagate through containerd via the Port Layer.
```
vburenin: it is still in a design phase, I will have some focus on it next sprint.

doc/design/containerd/port-layer-integration.md
```diff
@@ -0,0 +1,276 @@
...
+  </tr>
+  <tr>
+    <td>Exec’d process state</td>
+    <td>/proc PID ?</td>
+    <td>Tether via Port Layer</td>
+  </tr>
+</table>
+
+
+* There’s a question of whether VIC should support the notion of creating a flat bundle without snapshots. OCI supports this idea by extracting layers on top of each other on the local filesystem. There are potential performance advantages, but these would be significantly offset by the lack of layer reuse.
```
vburenin: For production workloads this is not the concern at all. Since in majority of the cases companies are running specific product for a relatively long time, so the time to merge bunch of layers together to come up with a snapshot is going to be negligible in comparison to everything else. Nevertheless, if there is a way to fake layers on our side, we should consider this approach of layer reuse, since potentially VIC engine may become a tool of developers or QA.

doc/design/containerd/port-layer-integration.md
```diff
@@ -0,0 +1,276 @@
...
+
+The question is how we handle TTY/attach and streaming of stdout and stderr. In the current version of containerd, the containerd-shim maintains a Fifo on the filesystem that can be reconnected to if containerd shuts down and restarts.
+
+<TBD>
+
+**Questions**
+
+There are a number of questions that still need to be worked out. Here’s are the most obvious:
+
+*Do we need to create a process per container in the VIC appliance to fit with the executor model?*
```
vburenin: No. We do not need it. I managed to avoid it.

doc/design/containerd/port-layer-integration.md
```diff
@@ -0,0 +1,276 @@
...
+
+Both runc and the containerd shim are a process-per-container model. Our Port Layer is a long-running service. We will need some kind of shim layer on the Port Layer to mate with the executor interface anyway. Does having entries in /proc for each container help in any way?
```
vburenin: there is a way to implement runtime plugin that can recover the state and fake up all running shims.

doc/design/containerd/port-layer-integration.md
```diff
@@ -0,0 +1,276 @@
...
+
+*What do we do with the config.json in a bundle?*
+
+When a bundle is extracted to a local filesystem, config.json is used as input to runc in terms of how to run the container. Config.json in the normal case will be a composite of image config and runtime config generated by containerd. For us to integrate seamlessly, we should expect the output of containerd to still be a config.json, but this will then need to be parsed by the Port Layer in creation of the VMConfigSpec. The config.json does not then need to end up in the bundle filesystem mounted to the container since runtime configuration goes into VM guestInfo.
```
vburenin: shim module I am faking has an access to this config, so we could transform it accordingly before talking to portlayer.

doc/design/containerd/port-layer-integration.md
```diff
@@ -0,0 +1,276 @@
...
+
+Other hypervisor implementations list a bunch of areas that they either [don’t support or ignore](https://github.com/01org/cc-oci-runtime/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20label%3Adomain%3AOCI). The spec itself lists a bunch of parts that are "optional". We should put a proposal together in another document.
+
+*Role of the Supervisor component*
+
+The Supervisor is a multiplexer for shims, so it's very similar in scope and function to our Port Layer.
+
+*Are snapshots created during image pull or container create?*
+
+In the current model, snapshots are created on pull. Containerd seems to be designed to have snapshots created as part of container create. This needs to be clarified as it will have significant performance implications on create as well as caching implications for the image data.
```
vburenin: at the moment I just point to vSphere storage/vmdk that should be used for container.

doc/design/containerd/port-layer-integration.md
```diff
@@ -0,0 +1,276 @@
...
+
+*What to do with pause and resume on the executor interface?*
+
+This could map to VM suspend and resume, but to date this is not something we’ve tried or even seen a need for. According to the containerd team, this is for the purpose of committing images.
```
vburenin: We could use it to pause VM, it might be necessary if people decide to continue further investigation of incorrectly running container on some other VM deployment. Just saying, not sure if it is even possible.

doc/design/containerd/port-layer-integration.md
```diff
@@ -0,0 +1,276 @@
...
+
+In the current model, snapshots are created on pull. Containerd seems to be designed to have snapshots created as part of container create. This needs to be clarified as it will have significant performance implications on create as well as caching implications for the image data.
+
+*What to do with pause and resume on the executor interface?*
+
+This could map to VM suspend and resume, but to date this is not something we’ve tried or even seen a need for. According to the containerd team, this is for the purpose of committing images. 
+
+*How to handle TTY/attach and streaming of stdout and stderr* 
+
+The containerd-shim uses Fifo on the filesystem to keep these streams open even if containerd shuts down. There isn’t a clear integration point to handle this, so we need to work out how to map this to our current implementation and whether an extension point needs to be created.
```
vburenin: I am planning to work on it next sprint.

doc/design/containerd/port-layer-integration.md
```diff
@@ -0,0 +1,276 @@
...
+
+*How to handle container **logs*
+
+In VICe, the portlayer handles retrieving logs, and the logs are streamed to the personality server using the attach mechanism.  The personality server performs some validation but is mostly hands off.
```
vburenin: we may need to fake it up and pull logs directly from vSphere, otherwise performance is going to suffer.

corrieb:
ping

Issue #3954, opened 2017-02-14
Store container log entries should be tagged with the stream they came from


jzt:
As an Admiral developer/user, I would like to be able to differentiate and filter between stdout and stderr containerVM log entries so that Admiral can work with VIC.

Currently (after #3882 is merged), logs are stored in the following format:

```json
{"msg":"log entry #1","ts":"2016-12-22T07:26:36.440445546Z"}
{"msg":"log entry #2","ts":"2016-12-22T07:26:36.440445546Z"}
{"msg":"log entry #3","ts":"2016-12-22T07:26:36.440445546Z"}
{"msg":"log entry #4","ts":"2016-12-22T07:26:36.440445546Z"}
```

Requirements for this story are that the log entries also contain the stream that produced them, like so:
```json
{"msg":"log entry #1","stream":"stdout","ts":"2016-12-22T07:26:36.440445546Z"}
{"msg":"log entry #2","stream":"stdout","ts":"2016-12-22T07:26:36.440445546Z"}
{"msg":"log entry #3","stream":"stderr","ts":"2016-12-22T07:26:36.440445546Z"}
{"msg":"log entry #4","stream":"stderr","ts":"2016-12-22T07:26:36.440445546Z"}
```

jzt:
Admiral does not actually need this feature, so I'm lowering the priority to medium so that I can have more time with #2539.

Issue #3935, opened 2017-02-13
Support for --timestamps option to docker logs


jzt:
As a developer on Admiral, I would like to be able to use the `--timestamps` option to `docker logs` so that I can see the timestamps associated with each log entry, in order for Admiral to work correctly with VIC.

Acceptance requirements:
 - [ ] adding the `--timestamps` option to `docker logs` should produce log output that is prefixed with the timestamp of each entry
- [ ] each timestamp needs to be padded on the right with zeros if necessary, in order to preserve alignment between the timestamps and the log entries
- [ ] containers created before these changes go in must report `"not yet implemented"` when `docker logs --timestamps` is attempted

jzt:
Requires completion of #1738.

Issue #2539, opened 2016-10-02
implement docker logs --since


mdubya66:
See https://docs.docker.com/engine/reference/commandline/logs/

mdubya66:
Moving into 1.0 to improve admiral support. Target is .10 for now

mdubya66:
Moving into 1.0 for basic admiral support. targeting .10 for now.

chengwang86:
I think this depends on #1738.  We need log timestamps to implement docker logs --since. If true, #1738 should have higher priority than this one.

PR #4014, opened 2017-02-23
Add timestamps to persisted containerVM output [specific ci=1-08-Docker-Logs]

### LINKED ISSUE #1738: Container logs need to have timestamp or a mechanism to index by timestamp

jzt:
This change replaces #3882. Notable differences:

- The container output entries are persisted as bytes (no longer wrapped in JSON), each entry with a header containing a description about the entry
- `docker logs` can now faithfully reproduce binary data
- entries are broken into smaller pieces if they exceed a certain size
- renamed `jsonlog` package to `iolog` (feedback expected)
- design doc updated
- added some tests around correctly reproducing binary output
- added some tests around breaking up large entries
- brought unit tests up to date

TODO:
- figure out how to tag entries with stream (stdout/stderr) - #3954 

Follow-on issues: #3935, #2539, #3954

Fixes #1738

### DIFF
--- a/cmd/tether/ops_linux.go
+++ b/cmd/tether/ops_linux.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -22,7 +22,9 @@ import (
 	"syscall"
 
 	log "github.com/Sirupsen/logrus"
+	"golang.org/x/crypto/ssh/terminal"
 
+	"github.com/vmware/vic/lib/iolog"
 	"github.com/vmware/vic/lib/tether"
 	"github.com/vmware/vic/pkg/dio"
 	"github.com/vmware/vic/pkg/trace"
@@ -50,6 +52,14 @@ func (t *operations) Log() (io.Writer, error) {
 
 	if err := setTerminalSpeed(f.Fd()); err != nil {
 		log.Errorf("Setting terminal speed failed with %s", err)
+	}
+
+	// enable raw mode
+	_, err = terminal.MakeRaw(int(f.Fd()))
+	if err != nil {
+		detail := fmt.Sprintf("Making ttyS1 raw failed with %s", err)
+		log.Error(detail)
+		return nil, errors.New(detail)
 	}
 
 	return io.MultiWriter(f, os.Stdout), nil
@@ -80,8 +90,20 @@ func (t *operations) SessionLog(session *tether.SessionConfig) (dio.DynamicMulti
 		log.Errorf("Setting terminal speed failed with %s", err)
 	}
 
+	// enable raw mode
+	_, err = terminal.MakeRaw(int(f.Fd()))
+	if err != nil {
+		detail := fmt.Sprintf("Making ttyS2 raw failed with %s", err)
+		log.Error(detail)
+		return nil, nil, errors.New(detail)
+	}
+
+	// wrap output in a LogWriter to serialize it into our persisted
+	// containerVM output format, using iolog.LogClock for timestamps
+	lw := iolog.NewLogWriter(f, iolog.LogClock{})
+
 	// use multi-writer so it goes to both screen and session log
-	return dio.MultiWriter(f, os.Stdout), dio.MultiWriter(f, os.Stderr), nil
+	return dio.MultiWriter(lw, os.Stdout), dio.MultiWriter(lw, os.Stderr), nil
 }
 
 func (t *operations) Setup(sink tether.Config) error {
--- a/doc/design/container_logging.md
+++ b/doc/design/container_logging.md
@@ -1,0 +1,26 @@
+# Container VM output logging in VIC
+
+When a container VM process produces output, that output is sent down through a serial connection into the vSphere backend where it is written to a logfile. When a user wants to see this output, they use the `docker logs` command, which reads from this logfile and displays the log entries on the command line in order from oldest to newest.
+
+However, in order for users to be able to use the `--since` option with `docker logs`, we must also couple timestamps to these log entries so that the user can selectively filter log messages to be displayed based on when those entries occurred.
+
+### Requirements
+
+The container VM logging mechanism must:  
+  1. Create a header for each log entry containing both the size of that entry, the time at which that entry occurred, and the stream from which that entry originated (stdout, stderr).  
+  2. Allow these entries to be read and at a later time, starting with the first entry occurring at or beyond the timestamp supplied by the user with the `--since` option to `docker logs`, or starting with the first entry in the logfile if `--since` was not used. 
+
+### Implementation
+
+A package for containerVM logging in VIC is added as `github.com/vmware/vic/lib/iolog`. In this package are two files, `log_writer.go` and `log_reader.go`. These files contain an implementation of Go's `io.Writer` and `io.ReadCloser` interfaces, respectively.
+
+The responsibilities of the `LogWriter` are:  
+  1. To create a header for the entry, containing the timestamp at which the entry occurred, the size of the entry, and the stream that produced the entry.
+  2. To write these bytes to the serial port associated with the containerVM logfile on the backend.
+  3. To flush the remaining bytes in the supplied buffer upon `Close()`
+
+The responsibilities of the `LogReader` are:  
+  1. To read in a header and decode it into the timestamp, size and stream of the entry.
+  2. To then read the following `size` bytes that contain the actual message.
+  3. To copy the message bytes into the underlying `Read` stream's `[]byte` slice. 
+  4. To preserve unwritten bytes in a call to `Read` in memory so that they may be written during the next call, in the case where the supplied `[]byte` slice was smaller than the log message we are trying to write. 
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -33,6 +33,7 @@ import (
 	"github.com/vmware/vic/lib/apiservers/portlayer/restapi/operations"
 	"github.com/vmware/vic/lib/apiservers/portlayer/restapi/operations/containers"
 	"github.com/vmware/vic/lib/config/executor"
+	"github.com/vmware/vic/lib/iolog"
 	"github.com/vmware/vic/lib/portlayer/exec"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/uid"
@@ -358,6 +359,13 @@ func (handler *ContainersHandlersImpl) GetContainerLogsHandler(params containers
 	if err != nil {
 		// Do not return an error here.  It's a workaround for a panic similar to #2594
 		return containers.NewGetContainerLogsInternalServerError()
+	}
+
+	// containers with PluginVersion > 0 will use updated output logging on the backend
+	// containers with ExecConfig.Version = nil are assumed to be PluginVersion 0
+	if container.ExecConfig.Version != nil && container.ExecConfig.Version.PluginVersion > 0 {
+		// wrap the reader in a LogReader to deserialize persisted containerVM output
+		reader = iolog.NewLogReader(reader)
 	}
 
 	detachableOut := NewFlushingReader(reader)
--- a/lib/iolog/constants.go
+++ b/lib/iolog/constants.go
@@ -1,0 +1,29 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package iolog
+
+const (
+	// TODO(jzt): re-arrange header layout so that size is on the LSB side, and TS is on MSB,
+	// with space for flags in the middle
+	futureFlag1 = 1 << iota
+	futureFlag2
+	futureFlag4
+	streamFlag
+
+	maxEntrySizeKB           = 4
+	maxEntrySizeBytes        = maxEntrySizeKB*1024 - 1
+	headerLengthBytes        = 10
+	encodedHeaderLengthBytes = 16 // base64 encoded header
+)
--- a/lib/iolog/log_reader.go
+++ b/lib/iolog/log_reader.go
@@ -1,0 +1,96 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package iolog
+
+import (
+	"encoding/base64"
+	"encoding/binary"
+	"io"
+)
+
+// LogReader reads containerVM entries from a stream and decodes them into
+// their original form
+type LogReader struct {
+	io.ReadCloser
+	prev []byte
+}
+
+// NewLogReader wraps an io.ReadCloser in a LogReader.
+func NewLogReader(r io.ReadCloser) *LogReader {
+	return &LogReader{ReadCloser: r}
+}
+
+// Read reads a 10 byte header and decodes it into the timestamp, stream and
+// size of an entry. It uses the size to read the next set of bytes as the
+// message, and then copies the message into the supplied buffer, saving
+// what will not fit in the buffer for the next call to Read.
+func (lr *LogReader) Read(p []byte) (int, error) {
+	var (
+		err  error
+		n, w int
+	)
+
+	ehdr := make([]byte, encodedHeaderLengthBytes)
+	msg := lr.prev
+
+	if msg == nil {
+		// read a header
+		n = 0
+		for n < encodedHeaderLengthBytes {
+			w, err = lr.ReadCloser.Read(ehdr[n:])
+			n += w
+			if err != nil {
+				return 0, err
+			}
+		}
+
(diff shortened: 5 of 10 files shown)


lib/iolog/log_writer.go
```diff
@@ -0,0 +1,109 @@
...
+		// The first 8 bytes are the timestamp in int64 unix epoch format
+		// The first 14 bits of the final two bytes represent the size of the entry
+		// 0x2 represents the stream - 0 for stdout, 1 for stderr
+		// 0x1 represents whether or not the entry is a partial entry - 0 for complete, 1 for a partial entry.
+		header := make([]byte, 10)
+		// prepare the header
+		size := len(entry)
+		s := uint16(size << 2) // make some room for stream and partial flags
+		stream := 0            // TODO(jzt): defaults to 0 (stdout) until we figure out how to add stream tag
+		s = s + uint16(stream<<1)
```
hmahmood: `s |= stream << 1`

lib/iolog/log_writer.go
```diff
@@ -0,0 +1,109 @@
...
+		} else {
+			// add a trailing newline to a partial entry
+			entry = append(entry, '\n')
+		}
+		// Each entry has a 10-byte header that describes the entry as follows:
+		// The first 8 bytes are the timestamp in int64 unix epoch format
+		// The first 14 bits of the final two bytes represent the size of the entry
+		// 0x2 represents the stream - 0 for stdout, 1 for stderr
+		// 0x1 represents whether or not the entry is a partial entry - 0 for complete, 1 for a partial entry.
+		header := make([]byte, 10)
```
hmahmood: I think you should have an encoder and decoder for your LogEntry struct that has this code. Should make it cleaner and easier to test.

lib/iolog/log_writer.go
```diff
@@ -0,0 +1,109 @@
...
+		stream := 0            // TODO(jzt): defaults to 0 (stdout) until we figure out how to add stream tag
+		s = s + uint16(stream<<1)
+		s = s + uint16(partial)
+		binary.LittleEndian.PutUint64(header[:8], uint64(time.Now().Unix()))
+		binary.LittleEndian.PutUint16(header[8:], s)
+		entry = append(header, entry...)
+		w, err := lw.w.Write(entry)
+		n += w
+		if err != nil {
+			break
```
hmahmood: Should we write out some partial entries or none if there is an error?

lib/iolog/log_reader.go
```diff
@@ -0,0 +1,96 @@
...
+			if err != nil {
+				break
+			}
+		}
+		if err != nil && err != io.EOF {
+			// only return if not EOF as we may actually have some bytes to copy
+			return 0, err
+		}
+		if partial {
+			msg = msg[:n-1] // do not preserve the trailing '\n' on partial entries
```
jzt: So that they don't stretch across the log file on the datastore

cmd/tether/ops_linux.go
```diff
@@ -80,8 +82,18 @@ func (t *operations) SessionLog(session *tether.SessionConfig) (dio.DynamicMulti
 		log.Errorf("Setting terminal speed failed with %s", err)
 	}
 
+	// enable raw mode
+	_, err = terminal.MakeRaw(int(f.Fd()))
+	if err != nil {
+		log.Errorf("Making ttyS2 raw failed with %s", err)
```
caglar10ur: Agreed with @hmahmood also I think we make ttyS1 raw too

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -360,6 +361,13 @@ func (handler *ContainersHandlersImpl) GetContainerLogsHandler(params containers
 		return containers.NewGetContainerLogsInternalServerError()
 	}
 
+	// containers with PluginVersion > 0 will use updated output logging on the backend
+	// containers with ExecConfig.Version = nil are assumed to be PluginVersion 0
+	if container.ExecConfig.Version != nil && container.ExecConfig.Version.PluginVersion > 0 {
```
caglar10ur: I suggest checking the version explicitly as we can introduce another change later and that chance could be incompatible with this. > 0 test will cause problems in that case.

I also argue having a version numbers means nothing to the reader so I strongly suggest defining some constants and using them as conditions, eg [obviously not tested, not sure even the syntax is right :)];

```
const (
        TheVersionThatILike = 1 << iota
        TheVersionNobodyLikes
        TheBestVersionEver
        TheBrokenOne
        TheBest
        YuggggeeeVersion
)

CurrentPluginVersion = TheBest
```

and then you can create a function called Satisfies which takes an int argument and call it like

```
if err :=version.Satisfies(TheVersionThatILike | YuggggeeeVersion); err !=nil {
...
} else if err := version.Satisfies(TheBrokenOne); err !=nil  {
...
}
```

which could be implemented like

```
func (v *Version) Satisfies(flags int) error {
	if flags&TheVersionThatILike != 0 && CurrentPluginVersion < TheVersionThatILike {
		return ErrNotSupported
	}
....
	return nil
}
```

lib/iolog/log_reader.go
```diff
@@ -0,0 +1,96 @@
...
+// limitations under the License.
+
+package iolog
+
+import (
+	"encoding/binary"
+	"io"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
```
caglar10ur: logging in logger /s ♻️

lib/iolog/log_reader.go
```diff
@@ -0,0 +1,96 @@
...
+	msg := lr.prev
+	if msg == nil {
+		// read a header
+		h := make([]byte, 10)
+		n = 0
+		for n < 10 {
+			w, err = lr.ReadCloser.Read(h[n:])
+			n += w
+			if err != nil {
+				break
```
caglar10ur: looks like you can return from here, no need to break from the loop, check one more time and then return

lib/iolog/log_reader.go
```diff
@@ -0,0 +1,96 @@
...
+		msg = make([]byte, size)
+		n = 0
+		for n < size {
+			w, err = lr.ReadCloser.Read(msg[n:])
+			n += w
+			if err != nil {
+				break
+			}
+		}
+		if err != nil && err != io.EOF {
```
caglar10ur: same here, you can check the condition in the loop and either return from there or break out and continue with the next statement

lib/iolog/log_writer.go
```diff
@@ -0,0 +1,109 @@
...
+	"bytes"
+	"encoding/binary"
+	"io"
+	"time"
+)
+
+const (
+	// MaxEntrySizeBytes is the maximum allowable size of a log entry
+	// before it will be broken into multiple entries
+	MaxEntrySizeBytes = 16 * 1024
```
caglar10ur: Is this 16K for a reason or is it just a random number we picked with a fair coin flip :) ?

lib/iolog/log_writer.go
```diff
@@ -0,0 +1,109 @@
...
+func NewLogWriter(w io.Writer) *LogWriter {
+	return &LogWriter{w: w}
+}
+
+// Write scans the supplied buffer, breaking off indvidual entries
+// and writing them to the underlying Writer, flusing any leftover bytes
+func (lw *LogWriter) Write(p []byte) (n int, err error) {
+	start := 0
+	for {
+		i := 0
```
caglar10ur: not that important but move this out of the loop not to allocate new var. in every iteration;

```
i := 0
for {
    i = 0
....
```

Same goes for end, entry and n assuming Write will be busy

lib/iolog/log_writer.go
```diff
@@ -0,0 +1,109 @@
...
+// adds a header to each entry before writing to the underlying Writer
+func (lw *LogWriter) write(b []byte) (n int, err error) {
+	entries := [][]byte{}
+	// break the entry up into multiple entries if necessary
+	for len(b) > MaxEntrySizeBytes {
+		entries = append(entries, b[:MaxEntrySizeBytes])
+		b = b[MaxEntrySizeBytes:]
+	}
+	entries = append(entries, b)
+	for i := 0; i < len(entries); i++ {
```
caglar10ur: Actually I prefer this if we are after performance as len loop eliminates bound checks made by range :)

lib/iolog/log_writer.go
jzt: Good catch, thanks

lib/iolog/log_writer.go
jzt: It's actually pretty huge, but it was determined by the number of leftover bits in the final two bytes of the header. We can make it smaller if we like, but reasonable (ie: not 64B) smaller values will leave us with some unused bits in the header (probably not a bad thing if you consider those optional future flags). I could make it 8K or 4K and save 1 or 2 bits respectively for future use. I'm down to tweak it to optimize for latency, or take suggestions as I have no strong opinion one way or the other since almost all entries right now will be well below 16K!

lib/iolog/log_reader.go
jzt: Haha :)

lib/iolog/log_reader.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+		if err != nil {
+			return 0, err
+		}
+		// parse header
+		ts := time.Unix(int64(binary.LittleEndian.Uint64(h[:8])), 0)
+		s := binary.LittleEndian.Uint16(h[8:10])
+		partial := (0x2&s)>>1 == 1
+		stream := int(0x1 & s)
+		size := int(s >> 2)
```
jzt: No, since `s` is a `uint`, it will be a logical right shift.

lib/iolog/log_reader.go
```diff
@@ -0,0 +1,96 @@
...
+				break
+			}
+		}
+		if err != nil {
+			return 0, err
+		}
+		// parse header
+		ts := time.Unix(int64(binary.LittleEndian.Uint64(h[:8])), 0)
+		s := binary.LittleEndian.Uint16(h[8:10])
+		partial := (0x2&s)>>1 == 1
```
jzt: It's on the "very near future" roadmap :)

lib/iolog/log_writer.go
jzt: I'm going to lower it to 8K for now.

lib/iolog/log_writer.go
jzt: Ignoring as per our conversation in meatspace.

caglar10ur: I imagine something like this http://mariogalluccistudio.com/wp-content/uploads/2014/07/MEATSPACE-PANO-669x1024.jpg

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
jzt: @caglar10ur would you mind if this was captured in an issue and approached at a later date? Time constraints and all..

caglar10ur: Sure, as long as later date is not 3 years :)

cmd/tether/ops_linux.go
```diff
@@ -80,8 +82,20 @@ func (t *operations) SessionLog(session *tether.SessionConfig) (dio.DynamicMulti
 		log.Errorf("Setting terminal speed failed with %s", err)
 	}
```
caglar10ur: My previous comment got lost for some reason,  I think you should also set ttys1 to raw mode

lib/iolog/constants.go
```diff
@@ -0,0 +1,25 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package iolog
+
+const (
+	maxEntrySizeKB    = 4
+	maxEntrySizeBytes = maxEntrySizeKB*1024 - 1
+	headerLength      = 10
+	streamFlag        = 0x8
```
caglar10ur:
```go
const (
        futureFlag1 = 1 << iota
        futureFlag2
        futureFlag4
        streamFlag
)
```

lib/iolog/log_writer.go
```diff
@@ -0,0 +1,126 @@
...
+	Now() time.Time
+}
+
+// LogClock is an implementation of the Clock interface that
+// returns time.Now() for use in the logio package
+type LogClock struct{}
+
+// Now returns the local time time.Now()
+func (LogClock) Now() time.Time {
+	return time.Now()
```
caglar10ur: I think you should be using UTC as Now is location dependent (includes timezone) so what happens if the client is in east cost and server is on west.

lib/iolog/constants.go
jzt: Ah, cool. Learned something new. 👍

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
jzt: Captured in #4059, feel free to edit (it's late & my brain is mushy).

doc/design/container_logging.md
```diff
@@ -0,0 +1,25 @@
...
+
+When a container VM process produces output, that output is sent down through a serial connection into the vSphere backend where it is written to a logfile. When a user wants to see this output, they use the `docker logs` command, which reads from this logfile and displays the log entries on the command line in order from oldest to newest.
+
+However, in order for users to be able to use the `--since` option with `docker logs`, we must also couple timestamps to these log entries so that the user can selectively filter log messages to be displayed based on when those entries occurred.
+
+### Requirements
+
+The container VM logging mechanism must:  
+  1. Create a header for each log entry containing both the size of that entry, the time at which that entry occurred, and the stream from which that entry originated (stdout, stderr).  
+  2. Allow these entries to be read and at a later time, starting with the first entry occurring at or beyond the timestamp suppled by the user with the `--since` option to `docker logs`, or starting with the first entry in the logfile if `--since` was not used.
```
sgairo: "suppled" to "supplied"

lib/iolog/log_reader.go
```diff
@@ -0,0 +1,81 @@
...
+// limitations under the License.
+
+package iolog
+
+import (
+	"encoding/binary"
+	"io"
+)
+
+// LogReader unwraps entires into LogEntry structs.
```
sgairo: "entires" to "entries"

lib/iolog/log_writer.go
```diff
@@ -0,0 +1,126 @@
...
+// NewLogWriter wraps an io.WriteCloser in a LogWriter
+func NewLogWriter(w io.Writer, clock Clock) *LogWriter {
+	return &LogWriter{
+		w:     w,
+		Clock: clock,
+	}
+}
+
+// Write scans the supplied buffer, breaking off indvidual entries
+// and writing them to the underlying Writer, flusing any leftover bytes
```
sgairo: "flusing" to "flushing"

lib/iolog/constants.go
```diff
@@ -0,0 +1,26 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package iolog
+
+const (
+	futureFlag1 = 1 << iota
```
hmahmood: Why not the other way round? i.e. just
```go
streamFlag = 1 << iota
```

lib/iolog/log_reader.go
```diff
@@ -0,0 +1,84 @@
...
+		n = 0
+		for n < headerLengthBytes {
+			w, err = lr.ReadCloser.Read(hdr[n:])
+			n += w
+			if err != nil {
+				return 0, err
+			}
+		}
+		// parse header
+		// ts := time.Unix(int64(binary.LittleEndian.Uint64(h[:8])), 0)
```
hmahmood: We are not reading the timestamp?

lib/iolog/log_writer.go
```diff
@@ -0,0 +1,126 @@
...
+// Write scans the supplied buffer, breaking off indvidual entries
+// and writing them to the underlying Writer, flushing any leftover bytes
+func (lw *LogWriter) Write(p []byte) (n int, err error) {
+	var (
+		start, end, i int
+		entry         []byte
+	)
+	for {
+		i = 0
+		if i = bytes.IndexByte(p[start:], '\n') + 1; i == 0 {
```
hmahmood: I think this is a little inefficient. If there are no newlines in `p` then we scan it fully here, and then break, only to scan `p` again in `write`.

How about you just go through `p` once, and write out an entry every time you hit a newline or the max bytes for an entry?

lib/iolog/constants.go
jzt: I was packing bits that are being used to the left, it just made more sense to me to have to the remaining stuff on the less significant side.

lib/iolog/log_reader.go
jzt: This will be utilized in #3935.

lib/iolog/log_writer.go
jzt: This is more of a readability optimization, but I added a TODO for when I'm working on it next sprint.

lib/iolog/log_writer.go
```diff
@@ -0,0 +1,128 @@
...
+		n, err = lw.write(entry)
+		if err != nil {
+			return n, err
+		}
+		// advance starting index
+		start = end
+	}
+	// flush the buffer if needed
+	if start < len(p) {
+		lw.write(p[start:])
```
hickeng: What sort of inflation are we getting in size? I'm guessing that most writes are small numbers of bytes (unless io.Copy has buffered into a large lump) where the header might be significant.

lib/iolog/constants.go
caglar10ur: endian problems

lib/iolog/log_writer.go
```diff
@@ -0,0 +1,128 @@
...
+// Clock defines an interface that wraps time.Now()
+type Clock interface {
+	Now() time.Time
+}
+
+// LogClock is an implementation of the Clock interface that
+// returns time.Now() for use in the logio package
+type LogClock struct{}
+
+// Now returns the local time time.Now()
```
caglar10ur: the comment is wrong it doesn't return local time :)

jzt: Good catch. How about if I leave it returning `time.Now()` and then when I use it, I use it as `logClock.Now().UTC().UnixNano()`?

lib/iolog/log_writer.go
jzt: The header adds 10 bytes, but otherwise the entry stays exactly the same. Most writes are likely fairly small (human readable text in reasonably-sized entries, if I had to guess).

lib/iolog/log_writer.go
sflxn: Are we sure, UTC is what we want?  What if my infrastructure is in California?

jzt: Isn't UTC used to address that specific issue?  See https://github.com/vmware/vic/pull/4014#discussion_r103370786

PR #4025, opened 2017-02-24
Sync drone secrets [specific ci=Group3-Docker-Compose]

jzt:
This (hopefully) fixes an out-of-sync drone secrets file that was committing buildocide.


andrewtchin:
lgtm if the build passes :-)

PR #4026, opened 2017-02-24
Create version file in log bundles

### LINKED ISSUE #3870: Log bundle needs version file
As a support engineer on VIC, I would like to see a version file in the log bundle so that I can know what version of VIC a customer is using when trying to triage an issue.

sgairo:
Fixes #3870


cmd/vicadmin/logs.go
```diff
@@ -174,6 +174,8 @@ func configureReaders() map[string]entryReader {
 		"journalctl":   commandReader("/bin/journalctl --no-pager"),
 		"dmesg":        commandReader("/bin/journalctl --dmesg --no-pager"),
 		"sys-block":    commandReader("ls -l /sys/block/"),
+		// To check version
+		"version":		versionReader("version"),
```
jzt: Can you convert whichever `"version"` is the actual filename to all caps?

cmd/vicadmin/vicadm.go
```diff
@@ -183,6 +183,13 @@ func newBytesEntry(name string, b []byte) entry {
 	}
 }
 
+type versionReader string
+
+func (path versionReader) open() (entry,error) {
+	defer trace.End(trace.Begin(string(path)))
```
jzt: Minor, but I think the `trace` is unnecessary here, since this is not doing anything particularly complex (in contrast to `commandReader`, for example).

jzt: Actually, ignore that. It would be nice to see the path in the vicadmin logs.

PR #4024, opened 2017-02-23
Update govmomi to latest

caglar10ur:
Required https://github.com/vmware/govmomi/commit/93064c0677521a409cfc701a9b6739ca2ee0a1f4

Fixes #3868

PR #4043, opened 2017-02-26
Use DatastoreFileManager to delete files

### LINKED ISSUE #3931: Delete vsan dom objects in vic-machine delete and image/volume delete
As a customer, I don't want to leak storage resources after use VCH, no matter it's docker rmi, docker volume rm or vic-machine delete. I need the storage used by VCH is deleted cleanly.

Related to this story, we need to add following code:
- [ ] in vic-machine delete, call vsan dom cache to clean all orphan dom objects, which mean vmdk is deleted, but dom still exists.
- [ ] in storage portlayer, call vsan dom cache to delete dom object related to specific vmdk file
- [ ] move image storage parent namespace creation from portlayer to vic-machine, like volume did, and record the namespace path, instead of the dir alias, which is not retrievable in vsan dom cache.

depends on PR #3929, additional work for #3787

dougm:
Use DatastoreFileManager to delete files

DatastoreFileManager is a new govmomi helper that combines
FileManager, VirtualDiskManager and Datastore types for managing
files on a datastore.

For now the focus is on deleting files, and in particular, not leaking
DOM objects on vSAN.

Fixes #3931

### DIFF
--- a/lib/install/management/store_files.go
+++ b/lib/install/management/store_files.go
@@ -16,7 +16,6 @@ package management
 
 import (
 	"bytes"
-	"context"
 	"fmt"
 	"path"
 	"sort"
@@ -32,7 +31,6 @@ import (
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/datastore"
-	"github.com/vmware/vic/pkg/vsphere/tasks"
 )
 
 const (
@@ -138,7 +136,7 @@ func (d *Dispatcher) deleteDatastoreFiles(ds *object.Datastore, path string, for
 		return empty, nil
 	}
 
-	m := object.NewFileManager(ds.Client())
+	m := ds.NewFileManager(d.session.Datacenter, true)
 	if err = d.deleteFilesIteratively(m, ds, dsPath); err != nil {
 		return empty, err
 	}
@@ -151,7 +149,7 @@ func (d *Dispatcher) isVSAN(ds *object.Datastore) bool {
 	return dsType == types.HostFileSystemVolumeFileSystemTypeVsan
 }
 
-func (d *Dispatcher) deleteFilesIteratively(m *object.FileManager, ds *object.Datastore, dsPath string) error {
+func (d *Dispatcher) deleteFilesIteratively(m *object.DatastoreFileManager, ds *object.Datastore, dsPath string) error {
 	defer trace.End(trace.Begin(dsPath))
 
 	// Get sorted result to make sure children files listed ahead of folder. Then we can empty folder before delete it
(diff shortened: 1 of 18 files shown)


dougm:
@mdubya66 we need your review to merge #4024

I'll also be updating vendor/govmomi again with this PR once vmware/govmomi#672 is merged.

pkg/vsphere/vm/vm.go
```diff
@@ -86,7 +86,7 @@ func (vm *VirtualMachine) FolderName(ctx context.Context) (string, error) {
 func (vm *VirtualMachine) DSPath(ctx context.Context) (url.URL, error) {
 	var mvm mo.VirtualMachine
 
-	if err := vm.Properties(ctx, vm.Reference(), []string{"runtime.host", "config"}, &mvm); err != nil {
+	if err := vm.Properties(ctx, vm.Reference(), []string{"config.files.vmPathName"}, &mvm); err != nil {
```
dougm: Note that this is not a required change for this PR, but in the process notice we can reduce the data collected quite a bit here.

lib/install/management/store_files.go
```diff
@@ -174,12 +172,10 @@ func (d *Dispatcher) deleteFilesIteratively(m *object.FileManager, ds *object.Da
...
 }
 
-func (d *Dispatcher) deleteVMFSFiles(m *object.FileManager, ds *object.Datastore, dsPath string) error {
+func (d *Dispatcher) deleteVMFSFiles(m *object.DatastoreFileManager, ds *object.Datastore, dsPath string) error {
 	defer trace.End(trace.Begin(dsPath))
 
-	if _, err := tasks.WaitForResult(d.ctx, func(ctx context.Context) (tasks.Task, error) {
-		return m.DeleteDatastoreFile(ctx, dsPath, d.session.Datacenter)
-	}); err != nil {
+	if err := m.Delete(d.ctx, dsPath); err != nil {
```
hmahmood: Is the retry on certain failures on needed here?

pkg/vsphere/datastore/datastore.go
```diff
@@ -244,9 +244,7 @@ func (d *Helper) Mv(ctx context.Context, fromPath, toPath string) error {
 func (d *Helper) Rm(ctx context.Context, pth string) error {
 	f := path.Join(d.RootURL, pth)
 	log.Infof("Removing %s", pth)
-	return tasks.Wait(context.TODO(), func(ctx context.Context) (tasks.Task, error) {
-		return d.fm.DeleteDatastoreFile(ctx, f, d.s.Datacenter)
-	})
+	return d.ds.NewFileManager(d.s.Datacenter, true).Delete(ctx, f) // TODO: NewHelper should create the DatastoreFileManager
```
emlin: Is this TODO supposed to be done in other PR?

lib/install/management/store_files.go
dougm: We retry TaskInProgress faults elsewhere, but that isn't thrown by DeleteVirtualDisk_Task or DeleteDatastoreFile_Task: http://pubs.vmware.com/vsphere-65/index.jsp#com.vmware.wssdk.apiref.doc/vim.fault.TaskInProgress.html

I don't see any faults we should be retrying for those methods and the caller(s) of this code already handle the FileNotFound case.

pkg/vsphere/datastore/datastore.go
dougm: Another PR, since this would be a cleanup/refactor that would simplify unrelated existing code, but not required to solve this issue.

lib/install/management/store_files.go
hickeng: How are we handling NotAuthenticated issues from session timeout or similar? If there's any aspect style remediation performed in the tasks package beyond handling InProgress retrys then we should be using it across the board. Otherwise not.

@cgtexmex something else for vmomi gateway

PR #4045, opened 2017-02-27
[full ci]Implement the in memory migration of container configs in the container cache

### LINKED ISSUE #3610: port-layer: support in memory data migration for container configuration
As a user, I want to upgrade VCH to a newer version, and leave existing container in old version. If I need newer version's container, I will recreate new container, which should be in newer version. So that I get similar workflow with vanilla docker.

matthewavery:
fixes #3610

The first thing to note is that some documentation changes will be needed to inform users of the possibility that a migrated container will be restricted in viable commands.

This PR is intended to check the container cache for needed ExecConfig migrations after an upgrade. It should mark containers that are migrated and cannot have their configs committed back to the vm's config in vsphere. To do this it should check during the commit phase of operations and reject any commit that involves a config marked as migrated in the cache.

### DIFF
--- a/lib/config/executor/container_vm.go
+++ b/lib/config/executor/container_vm.go
@@ -143,6 +143,9 @@ type ExecutorConfig struct {
 
 	// version
 	Version *version.Build `vic:"0.1" scope:"read-only" key:"version"`
+
+	// migrated - this is used during in memory migration to assign whether an execConfig is viable for a commit phase
+	Migrated bool
 }
 
 // Cmd is here because the encoding packages seem to have issues with the full exec.Cmd struct
--- a/lib/portlayer/exec/base.go
+++ b/lib/portlayer/exec/base.go
@@ -27,6 +27,7 @@ import (
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/lib/config/executor"
+	"github.com/vmware/vic/lib/migration"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/extraconfig"
 	"github.com/vmware/vic/pkg/vsphere/extraconfig/vmomi"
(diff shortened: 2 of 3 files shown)


lib/config/executor/container_vm.go
```diff
@@ -143,6 +143,9 @@ type ExecutorConfig struct {
 
 	// version
 	Version *version.Build `vic:"0.1" scope:"read-only" key:"version"`
+
+	// migrated - this is used during in memory migration to assign whether an execConfig is viable for a commit phase
+	Migrated bool
```
emlin: This value is actually used by portlayer in memory only, so suppose that attribute should be added into container cache, instead of here.
This definition is only for the persistent configuration.

lib/portlayer/exec/base.go
```diff
@@ -72,9 +73,22 @@ func newBase(vm *vm.VirtualMachine, c *types.VirtualMachineConfigInfo, r *types.
...
+		//TODO: We need to decide how we want to handle "partial migrations" since it is very possible the container could be broken at the end of it.
+		if isMigration && err == nil {
+			// this is the case that a migration has occurred.
+			migratedConfValues := vmomi.OptionValueFromMap(migratedConf)
+			migratedConfSource := vmomi.OptionValueSource(migratedConfValues)
+
+			extraconfig.Decode(migratedConfSource, base.ExecConfig)
+			base.ExecConfig.Migrated = true
+		} else {
+			extraconfig.Decode(src, base.ExecConfig)
```
emlin: if err != nil, it will go to this section.

lib/portlayer/exec/base.go
```diff
@@ -72,9 +73,22 @@ func newBase(vm *vm.VirtualMachine, c *types.VirtualMachineConfigInfo, r *types.
 	// construct a working copy of the exec config
 	if c != nil && c.ExtraConfig != nil {
 		src := vmomi.OptionValueSource(c.ExtraConfig)
-		extraconfig.Decode(src, base.ExecConfig)
+		containerExecKeyValues := vmomi.OptionValueMap(c.ExtraConfig)
+
+		migratedConf, isMigration, err := migration.MigrateContainerConfig(containerExecKeyValues)
+		//TODO: We need to decide how we want to handle "partial migrations" since it is very possible the container could be broken at the end of it.
+		if isMigration && err == nil {
```
emlin: before it, check if err != nil first. In that case, we should return err with user friendly error message, cause that might go back as docker command error message.

lib/portlayer/exec/commit.go
```diff
@@ -35,6 +35,13 @@ func Commit(ctx context.Context, sess *session.Session, h *Handle, waitTime *int
 	defer trace.End(trace.Begin(h.ExecConfig.ID))
 
 	c := Containers.Container(h.ExecConfig.ID)
+
+	// check for an in memory container migration and reject the commit if it is the case.
+	if c.ExecConfig.Migrated {
```
emlin: cannot bail out at this point, cause the vm operation is still allowed in this case.

emlin: check the migration status at here:
https://github.com/vmware/vic/pull/4045/commits/d5c7717969793557f357f1ba969d17bf14669755#diff-8207405211ed1d9756b8e63d8396f253R106

lib/config/executor/container_vm.go
emlin: Here should be the one you want to change: https://github.com/vmware/vic/blob/master/lib/portlayer/exec/container.go#L136

matthewavery: @emlin why not here in the container base instead? https://github.com/matthewavery/vic/blob/d5c7717969793557f357f1ba969d17bf14669755/lib/portlayer/exec/base.go#L53-L63 

ince we want to perform the migration close to the `decode` call.

emlin: ah.. okay, containerBase is good. That is cached as well.

PR #4054, opened 2017-02-27
[full ci] Move to new ESXI machines

### LINKED ISSUE #3673: Switch the test infrastructure over to the new servers
Need to make one PR that switches the IP list in the secrets file over to the new server IPs.

jakedsouza:
Fixes #3673

* Moves the ESXI test urls from the nested esxi's to 8 physical esxi machines.
* 4 on 6.0u2 and 4 on 6.5
* New machines have internal network with DHCP on vSwitchLAN instead of vSwitch0


jakedsouza:
CC @mdubya66 @rajanashok @casualjim

PR #4055, opened 2017-02-27
[skip ci] Serial Communication Research Spike Doc

kreamyx:
This document investigates some options to solve the vMotion issue during active serial connections to containerVMs, and proposes a solution along with action items for next steps.


doc/design/serial_communication_comparing_solutions.md
```diff
@@ -0,0 +1,54 @@
...
+
+<sup>3</sup>The Rabbitmq node (or cluster) has to be connected to the management network
+
+# Third: Proposal
+Based on the priority criteria put forward in issue #3937, vSPC seems to be the right way to go because of its ease of implememntation/integration with VIC's code base. The only significant difference between using CAF and vSPC is the high-bandwidth communication we are projected to get should we use CAF. Until performance becomes an issue, vSPC seems to be a very good candidate.
+Adopting the vSPC solution will require the following:
+
+1. Implementing an extensible telnet server
+2. Implementing a vSPC
+3. integrating the vSPC with VIC
```
caglar10ur: What does this mean? Are we planning to implement vSPC functionality inside the VCH or vic-machine (or vch) is going to create a container that runs the concentrator?

kreamyx: The integration part is not very clear to me yet, but as far as I understand the vSPC will be a binary that runs in the VCH appliance

corrieb: @caglar10ur The goal of this sprint was to come up with a recommendation of which solution we should go for. Karim has done a POC with the vSPC and concluded that it will not only meet our needs but could potentially be completed in time for 1.1. Next sprint, the goal will be to figure out how to integrate his current work

doc/design/serial_communication_comparing_solutions.md
```diff
@@ -0,0 +1,54 @@
...
+
+<sup>2</sup>Even though, AFAIK, there is no ESX host agent requirement, there is an extra overhead associated with refactoring the code to utilize the CAF API. Furthermore, the `rabbitmqproxy` in the ESX host should be configured to be able to communicate with the broker
+
+<sup>3</sup>The Rabbitmq node (or cluster) has to be connected to the management network
+
+# Third: Proposal
+Based on the priority criteria put forward in issue #3937, vSPC seems to be the right way to go because of its ease of implememntation/integration with VIC's code base. The only significant difference between using CAF and vSPC is the high-bandwidth communication we are projected to get should we use CAF. Until performance becomes an issue, vSPC seems to be a very good candidate.
+Adopting the vSPC solution will require the following:
+
+1. Implementing an extensible telnet server
```
hickeng: Just a note that we may not need the entire spec, just those pieces needed for vSPC comms.
I do **not** know how BINARY mode is negotiated with ESX, or if it can be - is that made explicit in the docs anywhere?

If we cannot negotiate BINARY with ESX then I do not know if or how the data encoding is done - it's possible we'd need to do a base64 stream encoding or similar.

doc/design/serial_communication_comparing_solutions.md
hickeng: @kreamyx If we implement the vSPC as a separate binary in the appliance then we'll need to relay the data again to the portlayer.
We may want to start with this approach as it isolates the changes completely, but it also adds several copys to the data path - we would likely look at in-lining the capability eventually.

doc/design/serial_communication_comparing_solutions.md
```diff
@@ -0,0 +1,54 @@
...
+------------------|------|------
+Ease of implementation/integration|relatively easy|more challenging<sup>1</sup>
+Communication Bandwidth|relatively small|relatively large
+Connection Persistence during vMotion|Supported|Supported
+No ESX host agent| Not required | Not required<sup>2</sup>
+No Network stack required for containerVMs|None required|None required<sup>3</sup>
+
+<sup>1</sup>CAF will require a lot of across-team communication and a major code refactoring
+
+<sup>2</sup>Even though, AFAIK, there is no ESX host agent requirement, there is an extra overhead associated with refactoring the code to utilize the CAF API. Furthermore, the `rabbitmqproxy` in the ESX host should be configured to be able to communicate with the broker
```
hickeng: Do you have a reference for the guest side CAF API? This is for personal curiosity as CAF didn't have this comms channel when I last looked at the proposal back in 2012.

kreamyx: This is the most recent information I found about CAF https://confluence.eng.vmware.com/display/INF/Common+Agent+-+Getting+Started#CommonAgent-GettingStarted-Programmer'sGuide

Martin Ghazaryan is a good contact person

doc/design/serial_communication_comparing_solutions.md
kreamyx: Do you mean Binary mode negotiation between a VM's serial port and the VM running the proxy? if yes, then I can confirm BINARY can be negotiated

PR #4057, opened 2017-02-28
toolbox: Add guest NicInfo RPC

### LINKED ISSUE #3824: Implement guest NicInfo RPC in vsphere/toolbox
As a customer, I need the VCH and container VM toolbox to populate the VirtualMachine guest.net property (GuestNicInfo) so I can use this data outside of the VM.

The toolbox currently populates the vm.ip property, but does not populate guest.net.  Certain tools depend on the latter.

Acceptance criteria:

1) Robot test that uses govc to validate guest.net data is populated by both the VCH and container VMs.

dougm:
Fixes #3824


vendor/manifest
```diff
@@ -248,10 +248,10 @@
 		},
 		{
 			"importpath": "github.com/davecgh/go-xdr/xdr2",
-			"repository": "https://github.com/davecgh/go-xdr",
+			"repository": "https://github.com/rasky/go-xdr",
```
caglar10ur: this uses rasky but we vendor and use davecgh one. Is it intentional?

dougm: yes, meant to mention that - this pulls in davecgh/go-xdr#6 until it lands, we need it for xdr optional pointer field support.

Issue #3720, opened 2017-01-25
[v1.13] docker client hangs


caglar10ur:
running ``` docker run hello-world``` using 1.13 client ends up with a hung client.  The underlying VM starts and then powers off as expected. Tether, portlayer and personality logs shows no errors. Looking at the pporf output suggests that the client waits an event. Using v1.12 client works as expected.

caglar10ur:
[logs.zip](https://github.com/vmware/vic/files/728360/logs.zip)

caglar10ur:
This could be related with the API version piece as client seems to waiting docker events to arrive. Raising another one for documenting the fact that we don't support new client versions.

mdubya66:
did hello world ever work? iirc we had problems all along with the default one.

caglar10ur:
Yes, hello-world works fine with older clients

stuclem:
Added proposed release note entry in https://github.com/vmware/vic/issues/3734.

sflxn:
I've identified the new code that is causing the hang.  The 1.13 CLI changed the run code to now subscribe to events, which we have not yet implemented.  It's waiting on an event from the daemon that says the container died, detach, or destroyed.

The only way to fix this is to implement events subscription.

stuclem:
Removing kind/note from this one as the RN update is being tracked in https://github.com/vmware/vic/issues/3734.

sflxn:
#2252 for event subscription manager on portlayer side
#367 for event subscription on personality side

Acceptance criteria for events will be new docker command robot scripts for events.  Specifically, for docker run, we need to check for "die", "detach", and "destroy".

mhagen-vmware:
lgtm

PR #4056, opened 2017-02-28
initial portlayer event stream

cgtexmex:
An events endpoint has been added to the portLayer REST
api.  That endpoint will allow clients to subscribe to a
stream of portLayer events.  The event stream is currently
limited to container events.

Fixes #4029

PR #4058, opened 2017-02-28
Add Docker container event support

### LINKED ISSUE #367: Docker API: Docker Container Event Subscription
Implement docker even subscription in the engine-api server and integrate it with issue #81.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#monitor-docker-s-events

sflxn:
Added passive docker events for containers and images.  Volumes
and networks events not yet added.  Also added monitoring of the
portlayer's event stream for container stop event.  Added support
for Docker's pubsub for events.

Resolves #367 and #3720

### DIFF
--- a/lib/apiservers/engine/backends/eventmonitor.go
+++ b/lib/apiservers/engine/backends/eventmonitor.go
@@ -1,0 +1,217 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


sflxn:
This will build once #4056 is merged.

lib/apiservers/engine/backends/eventmonitor.go
```diff
@@ -0,0 +1,217 @@
...
+	done := make(chan struct{})
+	go func() {
+		wg.Wait()
+		close(errors)
+		close(done)
+	}()
+
+	select{
+		case <- done:
+			for err := range errors {
```
hmahmood: We don't call `cancel()` here? You should probably just do `defer cancel()` right after you create the context further up.

sflxn: There really isn't a need to cancel here because we know that if we get here, both goroutines have already exited.

hmahmood: I think you have to call cancel on a context regardless (https://golang.org/pkg/context/):

> Failing to call the CancelFunc leaks the child and its children until the parent is canceled or the timer fires. The go vet tool checks that CancelFuncs are used on all control-flow paths.

PR #4071, opened 2017-02-28
[skip ci] Investigation on NSX integration

### LINKED ISSUE #3936: Investigate NSX micro-segmentation via Docker network create
As a VIC user, I want to be be able to create isolated networks using Docker network create that define secure scopes for deployed containers.

**High-level problem scope:**
Currently in VIC, we are only able to _segment_ networks using Docker network create, but we cannot provide full isolation. Customers who are already using NSX should be able to make use of its capabilities to create isolated network scopes for containers.
For example, I may have a Web Server that has an identity on the public network, but that also has an identity on a private network. I may want the database traffic from that Web Server to a Database VM to be isolated from any other containers deployed into that VCH on the same private network.

**Goals for this research spike:**
- [ ] Get a working installation of NSX-v
- [ ] Clarify how to communicate with NSX control plane
  - [ ] How to authenticate with the control plane (assume SSO with vSphere)
  - [ ] Does vSphere RBAC apply to NSX? Can a vSphere admin limit what a vSphere user can do?
  - [ ] What is the API endpoint URI of this rest interface? Same as vSphere APIs?
- [ ] Determine which API calls need to be made and which parameters to be passed
  - [ ] Assumption that Docker network create involves a call to create a policy group
  - [ ] Assumption that Docker create involves a call to associate VM network interface with policy group
  - [ ] Clean up call on VCH deletion
- [ ] How does tenancy / project scope in NSX for a policy group map to a single VCH?
  - [ ] Assume "local" scope vs "global"
- [ ] Document and determine how to create an NSX distributed PortGroup with logical switch
  - [ ] Can you specify one of these NSX created networks using vic-machine today?

**Acceptance criteria / output:**
Output should be a proposal / draft design doc for how we should do this integration. The proposal should cover the following:
- Authentication / authorization to NSX control plane
- Port Layer Integration. How should we expose this via the Port Layer API?
- Where do we store state? How to map Docker network name to NSX policy group?
  - Can we store that in NSX control plane? Needs to persist beyond the lifespan of the VCH appliance.
- Defined control flows - basic psuedocode or diagram that shows what calls what.
  - Create network
  - Create container
  - List networks
  - Delete network (how to fail if a container is attached to it? What does Docker do?)
  - Delete VCH

amandafeng001:
Fixes #3936


doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,109 @@
...
+https://{nsxmanager's hostname or IP}/api 
+
+### API calls
+(details in http://pubs.vmware.com/nsx-63/topic/com.vmware.ICbase/PDF/nsx_63_api.pdf , note that many NSX API methods reference vCenter object IDs in URI parameters, query parameters, request bodies, and response bodies, also it takes XML as request body (sad) )
+* docker network create
+  * create IP Sets
+     * POST /2.0/services/ipset/{scopeMoref}
+  * or create Security Tag
+     * POST /2.0/services/securitytags/tag
+  * create security group withe IP Sets or Security Tag as members
```
hickeng: `withe` --> `with`

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,109 @@
...
+  * delete the network (security group, policy group, IP Sets)
+     * DELETE /2.0/services/securitytags/tag/{tagId}
+     * DELETE /2.0/services/securitygroup/{objectId} 
+     * DELETE /2.0/services/ipset/{ipsetId}
+
+* docker network list
+  * list all the security groups
+    * GET /2.0/services/securitygroup/internal/scope/{scopeId}
+  * filter security group names by `<vch name>-SecurityGroup` prefix
+  * or if we keep KV pairs of each created network, then we just need to use that to list all the networks
```
hickeng: The infrastructure is supposed to be the source of truth as this allows a degree of dual-mode management. So, pulling from the infrastructure would be preferred.

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,109 @@
...
+
+* docker run <image> --net=<network-name>
+  * create a containerVM
+  * if using security tag as the membership condition, need to assign security tag to the containerVM
+    * POST /2.0/services/securitytags/vm/{vmId}
+
+## State Storage
+  * we can map a docker network name to a security group name as `<vch name>-SecurityGroup-<network name>`
+  * we can store KV pairs for each created network and then everything else will be in the NSX control plane
```
hickeng: what state needs to be stored outside of NSX? One obvious possibility is the IPAM configuration, but worth being explicit.

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Unclear / need discussion:
+### Security Tag v.s IP Sets
+  * Security Tag we only need to add the tag to a VM then it is added to a security group, which makes adding VMs to multiple security groups convinient. We do not need to update security groups everytime we create a new containerVM in a network.
+  * With IP Sets, we need to add the VM's IP to an IP Set. The downside is that 1. it requires the VM's IP to be static. 2. if we have more than one vch in the environment, containerVMs' IPs overlap, we cannot allow one VM in multiple security groups. 3. it may limit the number of VMs added to an IP Set (how many IPs can be added one IP set?). 4. everytime adding a new containerVM in a network, we need to add its IP to the IP set.
```
hickeng: Please reflow the numbered points so they render as a list.

Consideration of security tag:
1. Seems to apply to a VM instead of an interface on a VM - does this allow for different policies per vNIC?

Other possible considerations of IPs:
5. currently we only assign an IP to a VM when it is powered on so VMs wouldn't have a security group membership when powered off (meaning possible issues with out-of-band power on)
6. no way to prevent a container changing it's IP address

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,109 @@
...
+### Details of integrating with current Port Layer API 
+  * does the current Port Layer API map perfectly to our goals in using NSX ?
+  * if not, what changes do we need to make?
+
+### Bridged containers with exposed port or directly connected to public network
+  * I have not looked into how/if this can work with NSX yet
+  * all the user cases mentioned in the networking design document should still be valid or improved
+
+### To support both cases: with or without NSX?
+  * need to find out how to verify if a bridge network specified is a logic switch?
```
hickeng: bridge network is currently specified as a portgroup

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,109 @@
...
+
+### User management
+  * NSX allows RBAC, we can add Vcenter users and assign roles for them
+  * to my understanding, in terms of using NSX for 'docker network create', we will not need RBAC controls? Unless if we want to restrict users from creating networks, otherwise we may not need RBAC at this point?
```
hickeng: It depends what else can be done via the NSX management connection.
We started using admin users for VC and as a result have some awkward usages and have had to experimentally determine what privileges we need for what operations.

It's much easier to start with zero privileges while developing and document the addition as they are needed.

The other related question is can NSX use vCenter users for authZ or do we need to add a capability to add/specify an NSX user.
If the latter then we may need two users, one that can create the bridge network for vic-machine, the other to create security groups on it for endpointVM operations.

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,109 @@
...
+### Prerequisite
+VSphere admin already has NSX installed and configured on VCenter. That means there is an NSX logical switch created by the admin, which automatically creates a port group on distributed switch.
+So this port group can be used as a bridge network for `vic-machine create`
+
+![Distributed Switch with logic switch added] (pics/dswitch.png)
+
+### Create VCH 
+specifying 'bridge-network' as the port group 'vxw-dvs-55-virtualwire-1-sid-5000-logical-switch-1':
+  * I use this VCH to create containers, they are automatically added to this port group (i.e. logical switch)
+  * containerVMs get their IPs in the internal network
```
amandafeng001: This would depends on the NSX configuration and topology. NSX edge can act like a DHCP server, but can only serve interfaces directly connected to it. or NSX Distrubted Logical Router(DLR) can have external DHCP servers via DHCP relay configuration. However, we cannnot configure a DHCP on an ESG and then use the DHCP Relay function from the DLR to get an IP address from the ESG.

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,109 @@
...
+  * containerVMs can ping each other
+
+### docker network create network-a
+Our goal:
+  * VMs can communicate with VMs in the same security group.
+  * VMs cannot talk to VMs in a differrent security group.
```
amandafeng001: I would say they are completely independent. 
we can tag a VM to make it belong to one security group. but NSX uses this VM's IP to identify and apply security policies.

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,109 @@
...
+  * one VM to be added to more than one security group.
```
amandafeng001: Yes. We can associate a NIC with a security group by configuring dynamic membership as `entity belong-to <the vNIC>`.
I just experimented with VMs with multiple vNICs. The policy on one vNIC does not affect another vNIC.
If I understand correctly, you mean a containerVM may be attached to some external container-network and also have an identity in VIC created internal network?  I will need to verify if it works in this case.

doc/design/networking/Investigation-on-NSX-integration.md
amandafeng001: thanks. will update!

doc/design/networking/Investigation-on-NSX-integration.md
amandafeng001: Thanks. Get it. I will update the document.

doc/design/networking/Investigation-on-NSX-integration.md
amandafeng001: sorry I was not clear here. I meant storing some key/value pair like: 
  [key]: docker network name, [value]: information of the related objects created on NSX side. 
can help fetching network info faster. 
but as your previous comment mentioned, pulling from infrastructure would be preferred. Then no states need to be stored outside of NSX.

doc/design/networking/Investigation-on-NSX-integration.md
amandafeng001: Thanks Geroge. This is very important input. 
Actually at the point when I sent this PR, security tag did not work on VIC. that's why I was considering IP Sets at that time.
After Doug's change https://github.com/vmware/vic/pull/4057, security tag works on VIC.
But associating security group with security tag/VM may not work for the case you mentioned, where a container VM may have an external and internal identity. 
I will investigate more on associating security group with vNIC and update the document.

doc/design/networking/Investigation-on-NSX-integration.md
amandafeng001: yes. I meant, if we want to support both cases:
1. user has NSX setup.
2. user does not have NSX setup.
then we need a way to figure out if the port group here is an NSX created logical switch or just a regular port group. 
I think we can search for the `sid-` inside the name of the port group (for example, a logical switch name may be like this `vxw-dvs-55-virtualwire-12-sid-5010-logical-switch-23`)
but this approach may give false positives if someone creates a port group name as `vxw-dvs-55-virtualwire-12-sid-1`

amandafeng001:
@hickeng , Thanks for the comments. 
I'll investigate on associating vNIC with security group for VIC, and update the document soon.

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,180 @@
+# Investigation on NSX micro-segmentation via Docker network create
+This document is for this issue: https://github.com/vmware/vic/issues/3936
+
+
+## A Working Installation of NSX-v
+For VMWare engineers using Nimbus, here is a document about how to install and configure NSX with Vsphere: How to set up VM-to-VM communication on NSX logical switch with Nimbus
```
anchal-agrawal: s/VMWare/VMware

s/Vsphere/vSphere

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,180 @@
...
+  * containerVM3 , containerVM4 and containerVM5 are in bridge network `network-a` 
+  * containerVM5 and containerVM6 are in bridge network `network-b`
+    (note that adding containerVM5 to both network `network-a` and network `network-b` are not supported on current VIC yet.)
+
+![An Example of How We Apply Security Group on Container Networks] (pics/example-SG-vs-network.png)
+
+Let us use this example to explain the workflow.
+
+### Prerequisite
+We assume VSphere admin already has NSX installed and configured on VCenter. That means there is an NSX logical switch created by the admin, which automatically creates a port group on distributed switch.
```
anchal-agrawal: s/VSphere/vSphere

s/VCenter/vCenter

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,180 @@
...
+
+## About NSX Security Groups and Security Policies
+This section is a brief introduction about NSX security groups and security policies (some words are taken from this document http://www.vmware.com/content/dam/digitalmarketing/vmware/en/pdf/products/nsx/vmw-nsx-network-virtualization-design-guide.pdf).  
+  * Security Groups
+  
+    NSX provides various grouping mechanisms :
+      * vCenter Objects: VMs, Distributed Switches, Clusters, etc.
+      * VM Properties: vNICs, VM names, VM operating Systems, etc.
+      * NSX Objects: Logical Switches, Security Tags, Logical Routers, etc. 
+    In this document, we propose to use vNICs as the grouping creteria. We will dicuss more on this later.
```
anchal-agrawal: s/creteria/criteria

s/dicuss/discuss

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,180 @@
...
+    * PUT /2.0/services/securitygroup/bulk/{objectId}
+
+
+## Discussion:
+### Multi-tenants
+NSX-v supports multi-tenancy, and here is an example of a NSX network topology:
+
+![NSX-v multi-tenancy] (pics/nsx-multi-tenants.png)
+
+It worths to mention that we would expect only one VCH is created for one tenant. Otherwise there IP overlapping would issue.
```
anchal-agrawal: We can replace this with something like:

"It is worth to mention that we would expect only one VCH to be created for one tenant. Otherwise IP overlapping would occur."

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,180 @@
...
+  * if not, what changes do we need to make?
+
+### With or Without NSX?
+Do we surpport both cases :
+  1. User does not have NSX setup, then we continue providing with current VIC
+  2. User has NSX setup, then we provide VIC with NSX integration
+  * need to find out how to verify if a bridge network specified is a logic switch?
+
+### User Management
+  * NSX allows RBAC, we can add Vcenter users and assign roles for them
```
anchal-agrawal: s/Vcenter/vCenter

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,180 @@
...
+      * We only need to tag a VM then it is added to a security group. 
+      * we do not need to update security groups everytime a new containerVM is created. 
+    * Cons:
+      * It does not allow a containerVM to be added in multiple networks.  
+
+  * vNIC:
+    * Pros:
+      * It serves our goals.
+    * Cons:
+      * Everytime a containerVM is created, we need to update related security groups' by adding one more membership criterias. Not sure the maximum number of membership criterias NSX-v supports.
```
anchal-agrawal: s/criterias/criteria

doc/design/networking/Investigation-on-NSX-integration.md
amandafeng001: Thanks Anchal. I will update!

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,254 @@
...
+  * containerVM3 , containerVM4 and containerVM5 are in bridge network `network-a` 
+  * containerVM5 and containerVM6 are in bridge network `network-b`
+    (note that adding containerVM5 to both network `network-a` and network `network-b` are not supported on current VIC yet.)
+
+![An Example of How We Apply Security Group on Container Networks](pics/example-SG-vs-network.png)
+
+Let us use this example to explain the workflow.
+
+### Prerequisite
+We assume vSphere admin already has NSX installed and configured on vCenter. That means there is an NSX logical switch created by the admin, which automatically creates a port group on distributed switch.
```
fdawg4l: I'm just curious;  can we create the port group for the user in the installer?

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,254 @@
...
+  * VIC creates VCH
+  * NSX manager creates a security group `<vch name>-SG-VCH` with dynamic membership: NIC (of VCH) 
+  * NSX manager creates these security policies shown in the following picture
+  * NSX manager applies security policy `<vch name>-SP-0` on security group `<vch name>-SG-VCH`
+
+  ![Security policies](pics/security-policies.png)
+
+### docker run --name=containerVM1 <image name>
+  * VIC create containerVM1
+  * NSX manager creates security group `<vch name>-SG-bridge` with dynamic membership: NIC-1 (of containerVM1)
```
fdawg4l: Is the intent that we create a `bridge` security group for every container?

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,254 @@
...
+  * containerVMs in the network can connect to their gateway at VCH
+
+## NSX Manager API:
+![Code flow to use NSX Manager API](pics/with-nsx.png)
+
+### API endpoint
+https://{nsxmanager's hostname or IP}/api 
+
+### API calls
+(details in http://pubs.vmware.com/nsx-63/topic/com.vmware.ICbase/PDF/nsx_63_api.pdf , note that many NSX API methods reference vCenter object IDs in URI parameters, query parameters, request bodies, and response bodies, also it takes XML as request body (sad) )
```
fdawg4l: LOL @ "sad".  Is it SOAP by chance?  If so, you could probably reuse some of the client code in `govmomi` to do the wsdl stuff.

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,254 @@
...
+## Discussion:
+
+### Bridged Containers with Exposed Port or Directly Connected to Public network
+  * all the user cases mentioned in the networking design document should still be valid 
+
+## Unclear
+### Adding a ContainerVM to Multiple Networks
+This is not supported in current VIC implementation. 
+  * Is adding another NIC to a containerVM the right way to do it?
+  * Does having multiple interfaces matter ? How would user use multiple NICs?
```
fdawg4l: I feel like this overlaps with the "scopes" work @hmahmood did.  @hmahmood do we add a new scope for a new set of containers that doesn't exactly fit the smaller set or does the scope get extended?

Basically my question is whether or not the semantic is different than we currently have with scopes.

doc/design/networking/Investigation-on-NSX-integration.md
```diff
@@ -0,0 +1,254 @@
...
+  * Only security Admin or enterprise admin can operate on NSX security groups and policies. 
+  * Is there a way to assign users to a logical switch?
+
+### Approaches Comparison: Security Tag versus vNIC versus Logical Switch
+  * Security Group Membership with Security Tag: 
+    * Pros:
+      * We only need to tag a VM then it is added to a security group. 
+      * we do not need to update security groups everytime a new containerVM is created. 
+    * Cons:
+      * It seems at the backend of NSX, IP address is used to identify the membership of a security group. So it does not work in this scenario, for example, when there are two port groups on the same VDS and each of them have a VIC (say VIC-a and VIC-b) created. Then the security group and policy on VIC-a aslo affects VIC-b. (One NSX-v tenant is not allowed to have overlapping IPs. But VIC users might be the same tenant (if our understanding of a NSX tenant is correct).)
```
fdawg4l: RE:  IP mapping to SG.  I'm really surprised by this.  You can't tag by something more unique like mac or MoRef?

doc/design/networking/Investigation-on-NSX-integration.md
amandafeng001: Yes. We can do that. As long as NSX has been installed, in the installer we can make API calls to NSX to create logical switch, which will create corresponding port group :)

doc/design/networking/Investigation-on-NSX-integration.md
amandafeng001: Sorry for the confusion here. I mean we create a security group for each container network. And then everytime we add a container (say with NIC-1) to that network, we update the security group by adding a criteria of 'entitiy equals to NIC-1'. This is how the container's NIC-1 to the security group.

doc/design/networking/Investigation-on-NSX-integration.md
amandafeng001: it is a REST API... I think the plan is to add code in govmomi. Then make calls to govmomi from port layer. I'm not familiar with wsdl..what is it? and do you think at least I can use it to handle the xml part?

doc/design/networking/Investigation-on-NSX-integration.md
amandafeng001: If we go with the Security Group with vNIC association approach, we can build it on top of what we already have. I think it is loosely like this:  `scope == docker network` versus `scope + security group == docker network`.
The scopes will still be there. We add security groups on top.  The difference is when we add a container to another network, we need to create another vNIC to give it an identify in the new network. Would love to hear @hmahmood 's input on this.

doc/design/networking/Investigation-on-NSX-integration.md
amandafeng001: I actually tried to use mac to associate security group. Turns out it eventually still uses IP to identify membership. only with vNIC, security group membership can be correctly identified even with overlapping IP. 
I asked a NSX-v expert, he said "As dataplane is an IP classification engine, all the members of the security group will be converted to the IP addresses before sending them down to the ESXi host."

amandafeng001:
I am now working on another issue. So this work is paused for now. I will come back to work on it when I finish the other one.

doc/design/networking/Investigation-on-NSX-integration.md
hmahmood: Currently it is possible to add a container to multiple networks. A new vNIC is added if there is no NIC that is currently connected to the port group in question. It sounds like `scope == security group == docker network` to me.

amandafeng001:
This work is paused.

PR #4072, opened 2017-02-28
Fix upgrade version[specific ci=11-01-Upgrade]

emlin:
upgrade version is not updated. This is one regression.


gigawhitlocks:
LGTM THANK YOU @emlin

PR #4077, opened 2017-02-28
[specific ci=1-13-Docker-Version] Fix log collection in specific ci logic

anchal-agrawal:
This change fixes an issue with `specific ci` where the pybot log files (most importantly, `log.html`) of the chosen test suite would get overwritten by those of the regression suite.

The issue was seen in https://ci.vmware.run/vmware/vic/8229 and future `specific ci` builds.


anchal-agrawal:
Verified that this PR's build fixes the issue.

PR #4122, opened 2017-03-02
[specific ci=Group11-Upgrade] vic-machine upgrade --rollback

### LINKED ISSUE #3765: vic-machine upgrade --rollback
As vic admin, I want to upgrade my existing VCH endpoint VM to latest version. But after upgrade, if there is anything incompatible with my application, or script, I would like to roll back the previous upgrade, to make sure my workload is not impacted, and then I get some more time to update my script/application first.

And if there is any unexpected interrupt happens during upgrade, which make the upgrade stop failed and not roll back, this command should help me to roll back the VCH endpoint services. 

Limitation in this story: only one step roll back is supported.

gigawhitlocks:
Resolves #3765 
 
Still needs tests and a little cleanup but I wanted to go ahead and post this.

This is `vic-machine upgrade --rollback` which allows you to manually go back to a saved snapshot after an upgrade if the upgrade succeeds (from the point of view of vic-machine or the product as a whole) but the user's application doesn't function properly, or for any other reason. Also this contains a couple of commits from @emlin 's open PR and a number of small typo/grammar fixes.

### DIFF
--- a/cmd/vic-machine/upgrade/upgrade.go
+++ b/cmd/vic-machine/upgrade/upgrade.go
@@ -59,10 +59,9 @@ func (u *Upgrade) Flags() []cli.Flag {
 			Usage:       "Time to wait for upgrade",
 			Destination: &u.Timeout,
 		},
-		cli.IntFlag{
+		cli.BoolFlag{
 			Name:        "rollback",
-			Value:       0,
-			Usage:       "Rollback n versions, e.g. --rollback=1 to go to the version before the last upgrade",
+			Usage:       "Roll back VCH version to before the previous upgrade",
 			Destination: &u.Rollback,
 		},
 	}
@@ -163,7 +162,7 @@ func (u *Upgrade) Run(clic *cli.Context) (err error) {
 	vConfig.Timeout = u.Timeout
 
 	// only care about versions if we're not doing a manual rollback
-	if u.Data.Rollback <= 0 {
+	if !u.Data.Rollback {
 		if err := validator.AssertVersion(vchConfig); err != nil {
 			log.Error(err)
 			return errors.New("upgrade failed")
@@ -176,7 +175,7 @@ func (u *Upgrade) Run(clic *cli.Context) (err error) {
 		return errors.New("upgrade failed")
 	}
 
-	if u.Data.Rollback <= 0 {
+	if !u.Data.Rollback {
 		err = executor.Upgrade(vch, vchConfig, vConfig)
 	} else {
 		err = executor.Rollback(vch, vchConfig, vConfig)
--- a/cmd/vic-machine/upgrade/upgrade.go
+++ b/cmd/vic-machine/upgrade/upgrade.go
@@ -59,6 +59,12 @@ func (u *Upgrade) Flags() []cli.Flag {
 			Usage:       "Time to wait for upgrade",
 			Destination: &u.Timeout,
 		},
+		cli.IntFlag{
+			Name:        "rollback",
+			Value:       0,
+			Usage:       "Rollback n versions, e.g. --rollback=1 to go to the version before the last upgrade",
+			Destination: &u.Rollback,
+		},
 	}
 
 	target := u.TargetFlags()
@@ -156,9 +162,12 @@ func (u *Upgrade) Run(clic *cli.Context) (err error) {
 	vConfig.BootstrapISO = path.Base(u.BootstrapISO)
 	vConfig.Timeout = u.Timeout
 
-	if err := validator.AssertVersion(vchConfig); err != nil {
-		log.Error(err)
-		return errors.New("upgrade failed")
+	// only care about versions if we're not doing a manual rollback
+	if u.Data.Rollback <= 0 {
+		if err := validator.AssertVersion(vchConfig); err != nil {
+			log.Error(err)
+			return errors.New("upgrade failed")
+		}
 	}
 
 	if vchConfig, err = validator.ValidateMigratedConfig(ctx, vchConfig); err != nil {
@@ -167,7 +176,13 @@ func (u *Upgrade) Run(clic *cli.Context) (err error) {
 		return errors.New("upgrade failed")
 	}
 
-	if err = executor.Upgrade(vch, vchConfig, vConfig); err != nil {
+	if u.Data.Rollback <= 0 {
+		err = executor.Upgrade(vch, vchConfig, vConfig)
+	} else {
+		err = executor.Rollback(vch, vchConfig, vConfig)
+	}
+
+	if err != nil {
 		// upgrade failed
 		executor.CollectDiagnosticLogs()
 		if err == nil {
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -449,6 +449,14 @@ func (d *Dispatcher) configLogging(conf *config.VirtualContainerHostConfigSpec, 
 	return []types.BaseVirtualDevice{serial}, nil
 }
 
+func (d *Dispatcher) setDockerPort(conf *config.VirtualContainerHostConfigSpec, settings *data.InstallerData) {
+	if conf.HostCertificate != nil {
+		d.DockerPort = fmt.Sprintf("%d", opts.DefaultTLSHTTPPort)
+	} else {
+		d.DockerPort = fmt.Sprintf("%d", opts.DefaultHTTPPort)
+	}
+}
+
 func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec, settings *data.InstallerData) error {
 	defer trace.End(trace.Begin(""))
 
@@ -534,11 +542,7 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 	},
 	)
 
-	if conf.HostCertificate != nil {
-		d.DockerPort = fmt.Sprintf("%d", opts.DefaultTLSHTTPPort)
-	} else {
-		d.DockerPort = fmt.Sprintf("%d", opts.DefaultHTTPPort)
-	}
+	d.setDockerPort(conf, settings)
 
 	personality := executor.Cmd{
 		Path: "/sbin/docker-engine-server",
@@ -1057,7 +1061,7 @@ func (d *Dispatcher) CheckServiceReady(ctx context.Context, conf *config.Virtual
 			log.Infof("vSphere API Test: %s %s", conf.Target, diag.UserReadableVCAPITestDescription(code))
 		}
 	} else {
-		log.Warningf("Could not run VCH vSphere API target check due to %v", err)
+		log.Warningf("Could not run VCH vSphere API target check: %v", err)
 	}
 
 	if err := d.CheckDockerAPI(conf, clientCert); err != nil {
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -45,7 +45,6 @@ import (
 	"github.com/vmware/vic/lib/spec"
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/ip"
-	"github.com/vmware/vic/pkg/retry"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/compute"
 	"github.com/vmware/vic/pkg/vsphere/diag"
@@ -1054,30 +1053,15 @@ func (d *Dispatcher) CheckServiceReady(ctx context.Context, conf *config.Virtual
 	// vic-init will try to reach out to the vSphere target.
 	log.Info("Checking VCH connectivity with vSphere target")
 	// Checking access to vSphere API
-
-	err := retry.Do(
-
-		// check access
-		func() error {
-			cd, err := d.CheckAccessToVCAPI(d.ctx, d.appliance, conf.Target)
-			if err == nil {
-				code := int(cd)
-				if code > 0 {
-					log.Warningf("vSphere API Test: %s %s", conf.Target, diag.UserReadableVCAPITestDescription(code))
-				} else {
-					log.Infof("vSphere API Test: %s %s", conf.Target, diag.UserReadableVCAPITestDescription(code))
-				}
-			}
-			return err
-		},
-
-		// retry on error
-		func(err error) bool {
-			return true
-		})
-
-	if err != nil {
-		log.Warningf("Could not run VCH vSphere API target check: %v", err)
+	if cd, err := d.CheckAccessToVCAPI(d.ctx, d.appliance, conf.Target); err == nil {
+		code := int(cd)
+		if code > 0 {
+			log.Warningf("vSphere API Test: %s %s", conf.Target, diag.UserReadableVCAPITestDescription(code))
+		} else {
+			log.Infof("vSphere API Test: %s %s", conf.Target, diag.UserReadableVCAPITestDescription(code))
+		}
+	} else {
+		log.Warningf("Could not run VCH vSphere API target check due to %v but the VCH may still function normally", err)
 	}
 
 	if err := d.CheckDockerAPI(conf, clientCert); err != nil {
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -45,6 +45,7 @@ import (
 	"github.com/vmware/vic/lib/spec"
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/ip"
+	"github.com/vmware/vic/pkg/retry"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/compute"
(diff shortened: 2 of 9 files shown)


cmd/vic-machine/upgrade/upgrade.go
```diff
@@ -59,6 +59,12 @@ func (u *Upgrade) Flags() []cli.Flag {
 			Usage:       "Time to wait for upgrade",
 			Destination: &u.Timeout,
 		},
+		cli.IntFlag{
+			Name:        "rollback",
+			Value:       0,
+			Usage:       "Rollback n versions, e.g. --rollback=1 to go to the version before the last upgrade",
```
emlin: hmm.. I think we don't support multiple steps rollback, do we need this int option here?

lib/install/management/upgrade.go
```diff
@@ -85,7 +85,9 @@ func (d *Dispatcher) Upgrade(vch *vm.VirtualMachine, conf *config.VirtualContain
 	}
 
 	if err = d.update(conf, settings); err == nil {
-		d.retryDeleteSnapshot(snapshotName, conf.Name)
+		// TODO don't save too many old snapshots.. uncomment this and delete old ones
```
emlin: old snapshot does not help the upgrade, might delete them before or after new snapshot is created

lib/install/management/appliance.go
```diff
@@ -1018,15 +1023,30 @@ func (d *Dispatcher) CheckServiceReady(ctx context.Context, conf *config.Virtual
...
-		code := int(cd)
-		if code > 0 {
-			log.Warningf("vSphere API Test: %s %s", conf.Target, diag.UserReadableVCAPITestDescription(code))
-		} else {
-			log.Infof("vSphere API Test: %s %s", conf.Target, diag.UserReadableVCAPITestDescription(code))
-		}
-	} else {
-		log.Warningf("Could not run VCH vSphere API target check due to %v", err)
+
+	err := retry.Do(
```
emlin: I think we don't want to repeat this check, cause this operation actually might block portlayer initialization.
If it failed, just leave it there should be fine. Otherwise, might make vch service restart more times, I guess.

emlin: And this is just one workaround for future's service detection running in VCH

lib/install/management/upgrade.go
```diff
@@ -99,10 +101,37 @@ func (d *Dispatcher) Upgrade(vch *vm.VirtualMachine, conf *config.VirtualContain
...
 	return err
 }
 
+func (d *Dispatcher) Rollback(vch *vm.VirtualMachine, conf *config.VirtualContainerHostConfigSpec, settings *data.InstallerData) error {
+
+	// some setup that is only necessary because we didn't just create a VCH in this case
+	d.appliance = vch
+	d.setDockerPort(conf, settings)
+	if !strings.HasSuffix(conf.Target, "/sdk") {
+		conf.Target = fmt.Sprintf("%s/sdk", conf.Target)
```
emlin: why this is needed?

lib/install/management/upgrade.go
```diff
@@ -99,10 +101,37 @@ func (d *Dispatcher) Upgrade(vch *vm.VirtualMachine, conf *config.VirtualContain
...
+	}
+
+	snapshot, err := d.appliance.GetCurrentSnapshotTree(d.ctx)
+	if err != nil {
+		return errors.Errorf("could not find a snapshot to roll back to: %s", err)
```
emlin: the relationship between the old upgrade version and snapshot might not be clear enough for customer. here might add some more meaningful information, sth like no upgrade version is found

lib/install/management/upgrade.go
```diff
@@ -99,10 +101,37 @@ func (d *Dispatcher) Upgrade(vch *vm.VirtualMachine, conf *config.VirtualContain
...
+	}
+
+	if snapshot == nil {
+		return errors.Errorf("search for snapshot completed but no snapshot found")
```
emlin: same here

lib/install/management/upgrade.go
```diff
@@ -237,14 +266,13 @@ func (d *Dispatcher) update(conf *config.VirtualContainerHostConfigSpec, setting
 func (d *Dispatcher) rollback(conf *config.VirtualContainerHostConfigSpec, snapshot string, settings *data.InstallerData) error {
 	defer trace.End(trace.Begin(fmt.Sprintf("old appliance iso: %q, snapshot: %q", d.oldApplianceISO, snapshot)))
 
-	// do not power on appliance in this snapsthot revert
+	// do not power on appliance in this snapshot revert
```
emlin: 👍

lib/install/management/upgrade.go
```diff
@@ -158,15 +187,15 @@ func (d *Dispatcher) deleteSnapshot(snapshotName string, applianceName string) e
...
 	defer trace.End(trace.Begin(name))
 
-	upgrading, snapshot, err := d.appliance.UpgradeInProgress(d.ctx, UpgradePrefix)
-	if err != nil {
-		return err
-	}
-	if upgrading {
-		return errors.Errorf("Detected another upgrade process in progress. If this is incorrect, manually remove appliance snapshot %q and restart upgrade", snapshot)
-	}
+	// upgrading, snapshot, err := d.appliance.UpgradeInProgress(d.ctx, UpgradePrefix)
```
emlin: might remove them with some comments

emlin:
overall is good

lib/install/management/appliance.go
gigawhitlocks: Well if we don't do this, the rollback always beats the endpoint coming up, and we get a long timeout plus a nasty error message :\

lib/install/management/upgrade.go
gigawhitlocks: I don't know. You would think that `conf.Target` would have the `/sdk` suffix, but it does not, and it is needed.

lib/install/management/upgrade.go
gigawhitlocks: kk

lib/install/management/upgrade.go
gigawhitlocks: yeah I won't leave it like this. This is primarily what I was referring to in the description regarding cleanup

lib/install/management/appliance.go
gigawhitlocks: Plus I'm not sure why this would block the PortLayer from coming up? We're just resuming a snapshot..

lib/install/management/upgrade.go
gigawhitlocks: yes that was my original intent / design and I forgot to actually add it in :) I will go ahead and add in that functionality.

cmd/vic-machine/upgrade/upgrade.go
gigawhitlocks: This was in original design with an eye towards the future; actually already had it swapped to a `bool` locally :)

lib/install/management/appliance.go
emlin: the checkAccessToVCAPI will call vsphere API to run script inside of vm, that operation will be conflict with image portlayer operation to create image metadata. Either side might fail, and portlayer will be restarted for this conflict.
If this check failed, we simply print out warning message and continue. So no need to retry.

gigawhitlocks: Well, it *always* fails. Maybe I should just skip the check then?

gigawhitlocks: This is also a bummer to me too because frankly I think these little anonymous functions embedded in the call to `retry.Do` are cool :(

emlin: let's figure out why that failed, instead of skip it.

lib/install/management/upgrade.go
```diff
@@ -99,10 +101,41 @@ func (d *Dispatcher) Upgrade(vch *vm.VirtualMachine, conf *config.VirtualContain
...
+	}
+
+	notfound := "A VCH version available from before the last upgrade could not be found."
+	snapshot, err := d.appliance.GetCurrentSnapshotTree(d.ctx)
+	if err != nil {
+		return errors.Errorf("%s An error was reported while trying to discover it: %s", notfound, err)
+	}
+
+	if snapshot == nil {
+		return errors.Errorf("%s No error was reported, so it's possible that this VCH has never been upgraded or the saved previous version was removed out-of-band.", notfound)
```
emlin: 👍

tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
```diff
@@ -63,6 +63,39 @@ Upgrade VCH with containers
     Should Not Contain  ${output}  Rolling back upgrade
     Should Be Equal As Integers  ${rc}  0
 
+    # check version
```
emlin: this is duplicated with line 99

tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
```diff
@@ -63,6 +63,39 @@ Upgrade VCH with containers
...
+    Log  ${output}
+    Get Docker Params  ${output}  ${true}
+
+    Log To Console  \nTesting rollback...
+    ${rc}  ${output}=  Run And Return Rc And Output  bin/vic-machine-linux upgrade --debug 1 --name=%{VCH-NAME} --target=%{TEST_URL} --user=%{TEST_USERNAME} --password=%{TEST_PASSWORD} --force=true --compute-resource=%{TEST_RESOURCE} --timeout %{TEST_TIMEOUT} --rollback
+    Should Contain  ${output}  Completed successfully
+    Should Be Equal As Integers  ${rc}  0
+
+    # check version again, but old version this time
+    ${rc}  ${output}=  Run And Return Rc And Output  bin/vic-machine-linux version
```
emlin: old version should use this binary: ./vic/vic-machine-linux

tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
```diff
@@ -63,6 +63,39 @@ Upgrade VCH with containers
...
+    @{vers}=  Split String  ${output}
+    ${rc}  ${output}=  Run And Return Rc And Output  bin/vic-machine-linux inspect --name=%{VCH-NAME} --target=%{TEST_URL} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --password=%{TEST_PASSWORD} --compute-resource=%{TEST_RESOURCE}
+    Should Contain  ${output}  Completed successfully
+    Should Contain  ${output}  @{vers}[1]
```
emlin: is it @{vers}[1] or @{vers}[2]?

gigawhitlocks: vers[1] because it's the old version. (after rollback!) Also why above isn't a duplicate (I think -- I'm about to go double-check)

tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
gigawhitlocks: it's okay to use the new version of the vic-machine-linux binary here -- same as I've been doing with my ad-hoc testing.

tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
gigawhitlocks: Yeah, because what we do is: upgrade, check version, roll back, check version, upgrade, check version. It's intentional. It'd be clearer if I broke the steps out into a callable but Robot is such a pain LOL

tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
```diff
@@ -63,6 +63,39 @@ Upgrade VCH with containers
...
+    Should Be Equal As Integers  ${rc}  0
+    Log  ${output}
+    Get Docker Params  ${output}  ${true}
+
```
emlin: I think you should reorder this test. The ideal sequence should be
upgrade 
docker commands
regression
rollback
docker commands
regression again

but because it's hard to keep old version's regression test case, we could ignore the last step.

tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
gigawhitlocks: So we test that the initial upgrade can be rolled back, and then that we can upgrade back to the new version from the rolled-back version. When I test ad-hoc I usually do this back and forth three or four times.

gigawhitlocks:
Gonna rope in @rajanashok since you've moved on to reviewing tests @emlin

tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
gigawhitlocks: OK, to do that I'll have to actually organize this robot file so that there are keywords that we can call. Will do, but probably won't have it done today.

gigawhitlocks: Writing reusable code in Robot is a pain *sigh*

tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
gigawhitlocks: should probably be neither of these, actually. sigh.

pkg/vsphere/vm/vm.go
```diff
@@ -393,6 +393,7 @@ func (vm *VirtualMachine) GetSnapshotTreeByName(ctx context.Context, name string
 	return vm.bfsSnapshotTree(q, compareName), nil
 }
 
+// Finds a snapshot tree based on comparator function 'compare' via a breadth first search of the snapshot tree attached to the VM
```
anchal-agrawal: 👍  for adding a comment for a func otherwise not part of your changes.

gigawhitlocks: lol thanks @anchal-agrawal  I try

lib/install/management/upgrade.go
```diff
@@ -99,10 +101,41 @@ func (d *Dispatcher) Upgrade(vch *vm.VirtualMachine, conf *config.VirtualContain
...
+
+	err = d.rollback(conf, snapshot.Name, settings)
+	if err != nil {
+		return errors.Errorf("could not complete manual rollback: %s", err)
+	}
+
+	return d.retryDeleteSnapshot(snapshot.Name, conf.Name)
+}
+
 // retryDeleteSnapshot will retry to delete snpashot if there is GenericVmConfigFault returned. This is a workaround for vSAN delete snapshot
```
sgairo: "snapshot" instead of "snpashot"

pkg/vsphere/vm/vm.go
```diff
@@ -409,14 +410,19 @@ func (vm *VirtualMachine) bfsSnapshotTree(q *list.List, compare func(node types.
...
 func (vm *VirtualMachine) UpgradeInProgress(ctx context.Context, upgradePrefix string) (bool, string, error) {
 	node, err := vm.GetCurrentSnapshotTree(ctx)
 	if err != nil {
 		return false, "", fmt.Errorf("Failed to check upgrade snapshot status: %s", err)
 	}
 
-	if node != nil && strings.HasPrefix(node.Name, upgradePrefix) {
+	if IsUpgradeSnapshot(node, upgradePrefix) {
 		return true, node.Name, nil
 	}
```
sgairo: copyright range to 2016-2017 in header, I think.

gigawhitlocks: looks like it got updated on master since you left this comment

sgairo: perf

tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
```diff
@@ -128,5 +119,56 @@ Upgrade VCH with containers
...
+Upgrade VCH
+
+    Create Docker Containers
+    Log To Console  \nUpgrading VCH...
+
+    Upgrade
+    Check Upgraded Version
+
+    Rollback
+    Check Original Version
```
mhagen-vmware: fundamentally I am confused why we have a "check upgrade version" and a "check original version" here?  I would think it would just be Upgrade -> Check version -> roll back -> check version -> upgrade -> check version...

gigawhitlocks: I mean, I would've done that in anything but Robot, but the routines for checking the old and updated versions are subtly different beyond just checking for a specific version (so it's not as simple as extracting the version and passing it as an argument) and there are no control structures or anything like that in Robot script.

gigawhitlocks:
@mhagen-vmware the tests are documented.

tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
gigawhitlocks: Also it's not like we're going to need a generalized version-checking function in the future of this test suite

PR #4123, opened 2017-03-02
Uncompress kernel modules

### LINKED ISSUE #3858: Photon compressed kernel modules prevent Docker 1.12 daemon running nested in various distros: "Exec format error"
The photon kernel we use compresses its kernel modules using xzip:

Eg: 

```
> modinfo iptable_nat

filename:       /lib/modules/4.4.41-1.ph1-esx/kernel/net/ipv4/netfilter/iptable_nat.ko.xz
license:        GPL
depends:        nf_nat_ipv4
intree:         Y
vermagic:       4.4.41-1.ph1-esx SMP mod_unload
```
Many Linux distributions we can run using "docker run" include a version of kmod that either doesn't have support for compressed modules or is compiled without the support included.

Eg. Ubuntu:xenial
```
> modinfo --version

kmod version 22
-XZ -ZLIB -EXPERIMENTAL
```
Anything you run in a containerVM in one of these distros that needs to load a kernel module fails with "Exec format error".

caglar10ur:
Fixes #3858


isos/base.sh
```diff
@@ -69,6 +69,13 @@ cp $DIR/base/yum.conf $(rootfs_dir $PKGDIR)/etc/yum/
 
 # install the core packages
 yum_cached -c $cache -u -p $PKGDIR install filesystem coreutils linux-esx --nogpgcheck -y
+
+
+# Issue 3858: find all kernel modules and unpack them and run depmod against that directory
+find $(rootfs_dir $PKGDIR)/lib/modules -name "*.ko.xz" | xargs xz -d
+KERNEL_VERSION=$(basename $(rootfs_dir $PKGDIR)/lib/modules/*)
+chroot $(rootfs_dir $PKGDIR) depmod $KERNEL_VERSION
```
fdawg4l: Did you try `depmod -b`?
```
  -b basedir, --basedir basedir
           If your modules are not currently in the (normal) directory > /lib/modules/version, but in a staging area, you can specify a basedir which is prepended to the directory name. This basedir is
           stripped from the resulting modules.dep file, so it is ready to be moved into the normal location. Use this option if you are a distribution vendor who needs to pre-generate the meta-data files
           rather than running depmod again later.
```

caglar10ur: I wanted to use photon's depmod instead of the build machines :)

PR #4124, opened 2017-03-02
[specific ci=3-03-Docker-Compose-Basic] Fix docker-compose kill

### LINKED ISSUE #3707: Docker-compose kill fails but docker kill succeeds
As a user of VIC, I should be able to call docker-compose kill <service> and have the service container get killed.

Acceptance Criteria:  Robot script exercising docker-compose kill and verifying it worked.

---
Docker-compose kill calls ContainerKill with sig=0.  Docker kill calls ContainerKill with sig=9.  We're not handling ContainerKill properly for sig=0.

sflxn:
Handle docker kill with sig == 0.  This is the value that is used
when docker-compose kill is called.

Resolves #3707

### DIFF
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -41,6 +41,7 @@ import (
 	"net/http"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


sflxn:
After some thoughts, I decided to move all commands into Docker-Compose-Basic.  This is different from how we do docker command tests but it should improve test time since we won't have to deploy a new VCH for every command set.

Issue #4075, opened 2017-02-28
tether should update (instead of appending) containerName in /etc/hosts after changing the VM name in guestinfo 


chengwang86:
**VIC version:**

0.8

**Deployment details:**

What was the vic-machine create command used to deploy the VCH?
```bin/vic-machine-linux create --name=vch-1  --target=root:pwd@192.168.60.162 --thumbprint=...  --debug 1 --no-tlsverify --no-tls```

**Steps to reproduce:**
```
1.	docker run –it –name=t1 busybox
2.	govc vm.change -vm t1* -e guestinfo.vice./common/name=t2
3.	docker stop t1
4.	docker start containerID_of_t1
5.	docker attach containerID_of_t1
```

**Actual behavior:**
```
vi /etc/hosts
127.0.1.1 t1 cb4acd2807e0 t2
```

**Expected behavior:**
```
vi /etc/hosts
127.0.1.1 cb4acd2807e0 t2
```

chengwang86:
By this PR https://github.com/vmware/vic/pull/4134, we have removed containerName from `/etc/hosts`. So this issue does not exist anymore.

PR #4134, opened 2017-03-02
[specific ci=Group11-Upgrade] Reconfigure VM display name

### LINKED ISSUE #3972: Reconfigure the VM to change the display name and datastore folder name after VM creation
As a vic user, on non-vsan setup, I want to have the VM display name as `containerName-containerShortID`, and the datastore folder name as `containerID`; on vsan setup, I want to have both the VM display name and the datastore folder name to be `containerName-containerShortID`. By doing this, I can match the output from `docker ps` to the VM display name on vsphere web UI. 

This is first part of the docker rename support.

In the future, we may improve this feature and allow the customer to set up a customized naming template for the VM display name and datastore folder name.

**Acceptance Criteria:**

After a containerVM is created

On non-vsan:
- The VM display name on the vsphere web UI should be containerName-containerShortID.
- The datastore folder name should be containerID. 

On vsan:
- Both the VM display name on the vsphere web UI and its datastore folder name should be containerName-containerShortID.

chengwang86:
Fixes #3972, #4075 

This PR is the first step for `docker rename` support.

- added `ExecutorConfigCommon` in `container_vm.go` with the property of "name" setting to "hidden", so that the containerVM's `common/name` could be persisted to the vmx file while the user performs `docker rename` on a running containerVM

- `/etc/hosts` does not contain containerName anymore since `common/name` is set to "hidden". Therefore, we don't need to update `/etc/hosts` while renaming a container. This will solve #4075 

- when creating the containerVM, set the VM display name to `containerName-containerShortID` and datastore folder name to `containerID` for non-vsan setup, and set both names to `containerName-containerShortID` for vsan setup

- Updated all the relevant test cases where the containerVM's display name does not match with the new naming convention

- added a test case in 1-04-Docker-Create.robot: After creating a container, check that the VM displayname=containerName-shortID and datatstore folder name=containerID

- added a plugin for VCH upgrade since we changed the scope of `common.name` of the VCH

- added a plugin for containerVM data migration since we changed the scope of `common.name` of containerVMs (relies on #4146 )

- in vicadmin, obtain the VCH name from vsphere instead of from within the VCH (since the VCH hostname is hidden)


lib/config/executor/container_vm.go
```diff
@@ -44,6 +44,21 @@ type Common struct {
 	Notes string `vic:"0.1" scope:"hidden" key:"notes"`
 }
 
+// Common data (specifically for a containerVM) between managed entities, across execution environments.
+type Common2 struct {
```
sgairo: What is the thought process behind naming this Common2, I just want to be sure that there isn't a better or more appropriate name for this.

chengwang86: @sgairo We want to persist the `name` in Common while the containeVM is still running, so we need to set the property of `name` to `hidden`. In order not to mess up with the original `Common` which might require `read-only` name by other objects, we create `common2` instead.

lib/install/management/delete.go
```diff
@@ -161,23 +161,25 @@ func (d *Dispatcher) DeleteVCHInstances(vmm *vm.VirtualMachine, conf *config.Vir
...
-			continue
-		}
 		//Leave VCH appliance there until everything else is removed, cause it has VCH configuration. Then user could retry delete in case of any failure.
-		if name == conf.Name {
+		ok, err := d.isVCH(child)
+		if ok {
 			continue
 		}
+		if err != nil {
+			errs = append(errs, err.Error())
```
emlin: if error happens, we should continue to avoid delete vch by mistake

lib/migration/migrator_test.go
```diff
@@ -34,10 +36,10 @@ func setUp() {
 	// register sample plugin into test
 	log.SetLevel(log.DebugLevel)
 	trace.Logger.Level = log.DebugLevel
-	version.MaxPluginVersion = 1
+	version.MaxPluginVersion = 4
```
emlin: version.MaxPluginVersion = version.MaxPluginVersion + 1

lib/migration/samples/plugins/plugin1/stop_signal_rename_sample.go
```diff
@@ -31,7 +31,7 @@ import (
 // If only a couple of items changed in the configuration, you don't have to copy all VirtualContainerHost. Only define the few items used by
 // this upgrade plugin will simplify the extraconfig encoding/decoding process
 const (
-	version = 1
+	version = 4
```
emlin: why do you need to change this?

lib/migration/plugins/init.go
```diff
@@ -15,3 +15,7 @@
 package plugins
 
 // import all plugin packages here to register plugins
+import (
+	_ "github.com/vmware/vic/pkg/version/plugin2"
+	_ "github.com/vmware/vic/pkg/version/plugin3"
+)
```
emlin: add new line

lib/migration/samples/plugins/plugin1/stop_signal_rename_sample.go
chengwang86: @emlin No need to do this. I will fix it.

lib/portlayer/exec/container.go
```diff
@@ -555,6 +558,37 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
...
+
+	shortID := stringid.TruncateID(c.ExecConfig.ID)
+	nameMaxLen := maxVMNameLength - len(shortID)
+	prettyName := c.ExecConfig.Name
+	if len(prettyName) > nameMaxLen-1 {
+		prettyName = prettyName[:nameMaxLen-1]
+	}
+
+	fullName := fmt.Sprintf("%s-%s", prettyName, shortID)
+	task, err := c.vm.VirtualMachine.Rename(ctx, fullName)
```
emlin: using vm.WaitForResult method, that will retry for TaskInProgress error

lib/portlayer/exec/container.go
```diff
@@ -35,6 +35,7 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/vm"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/docker/docker/pkg/stringid"
```
emlin: Should we have portlayer rely on docker package?

lib/vicadmin/validate.go
```diff
@@ -209,14 +204,33 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
+	self, err := guest.GetSelf(ctx, sess)
+	if err != nil || self == nil {
+		log.Errorf("Unable to get VCH name due to unknown self-reference: %s", err)
+		log.Infof("Setting the VCH name to VCH")
+		return vchName
+	}
+	self2 := vm.NewVirtualMachineFromVM(ctx, sess, self)
+	vchName, err = self2.Name(ctx)
+	if err != nil {
+		log.Infof("Unable to get VCH name: %s", err)
```
emlin: if err != nil, vchname will be ""

chengwang86: @emlin No, it will be `VCH` instead of "".

pkg/version/plugin2/commonSpecForVCH.go
```diff
@@ -0,0 +1,123 @@
...
+	ID string `vic:"0.1" scope:"read-only" key:"id"`
+
+	// Convenience field to record a human readable name
+	Name string `vic:"0.1" scope:"read-only" key:"name"`
+
+	// Freeform notes related to the entity
+	Notes string `vic:"0.1" scope:"hidden" key:"notes"`
+}
+
+type NewVirtualContainerHostConfigSpec struct {
```
emlin: this looks like the method name to create struct VirtualContainerHostConfigSpec

pkg/version/plugin2/commonSpecForVCH.go
```diff
@@ -0,0 +1,123 @@
...
+	// Freeform notes related to the entity
+	Notes string `vic:"0.1" scope:"hidden" key:"notes"`
+}
+
+func (p *AddCommonSpecForVCH) Migrate(ctx context.Context, s *session.Session, data interface{}) error {
+	defer trace.End(trace.Begin(fmt.Sprintf("AddCommonSpecForVCH version: %d", version)))
+	if data == nil {
+		return nil
+	}
+	mapData := data.(map[string]string)
```
emlin: should assert data type here. If not a map object, might log error and return nil, to let other plugins continue

lib/portlayer/exec/container.go
chengwang86: You are right. I will remove this from portlayer.

pkg/version/plugin2/commonSpecForVCH.go
```diff
@@ -0,0 +1,123 @@
...
+	oldStruct := &VirtualContainerHostConfigSpec{}
+	result := extraconfig.Decode(extraconfig.MapSource(mapData), oldStruct)
+	log.Debugf("The oldStruct is %+v", oldStruct)
+	if result == nil {
+		return &errors.DecodeError{}
```
emlin: might create a error with message like decode which structure failed, and add into DecodeError

pkg/version/plugin3/commonSpecForContainer.go
```diff
@@ -0,0 +1,112 @@
...
+	ID string `vic:"0.1" scope:"read-only" key:"id"`
+
+	// Convenience field to record a human readable name
+	Name string `vic:"0.1" scope:"read-only" key:"name"`
+
+	// Freeform notes related to the entity
+	Notes string `vic:"0.1" scope:"hidden" key:"notes"`
+}
+
+type NewExecutorConfig struct {
```
emlin: ditto

pkg/version/plugin3/commonSpecForContainer.go
```diff
@@ -0,0 +1,112 @@
...
+	defer trace.End(trace.Begin(fmt.Sprintf("AddCommonSpecForContainer version %d", version)))
+	if data == nil {
+		return nil
+	}
+	mapData := data.(map[string]string)
+	oldStruct := &ExecutorConfig{}
+	result := extraconfig.Decode(extraconfig.MapSource(mapData), oldStruct)
+	log.Debugf("The oldStruct is %+v", oldStruct)
+	if result == nil {
+		return &errors.DecodeError{}
```
emlin: ditto

lib/migration/migrator_test.go
```diff
@@ -21,6 +21,8 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 
+	"strconv"
```
anchal-agrawal: gofmt: pkg `strconv` should be in the top-level pkg group:

```go
import (
        "strconv"
        "strings"
 	"testing"

        ...
)
```

lib/portlayer/exec/container.go
```diff
@@ -555,6 +558,37 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
...
+	}
+
+	fullName := fmt.Sprintf("%s-%s", prettyName, shortID)
+	task, err := c.vm.VirtualMachine.Rename(ctx, fullName)
+	if err != nil {
+		return err
+	}
+
+	_, err = task.WaitForResult(ctx, nil)
+	if err != nil {
```
anchal-agrawal:
```go
if err != nil {
    return err
}

return nil
```

can be replaced by `return err`

lib/portlayer/exec/commit.go
```diff
@@ -92,6 +92,12 @@ func Commit(ctx context.Context, sess *session.Session, h *Handle, waitTime *int
 		// clear the spec as we've acted on it - this prevents a reconfigure from occurring in follow-on
 		// processing
 		h.Spec = nil
+
+		// reconfigure vm display name to containerName-containerID
+		if err = c.UpdateDisplayName(ctx); err != nil {
+			return err
```
emlin: It's not a right position to call rename. The container is already put into container cache, and container created event is published. Other work rely on this event will use wrong information.
Another question is about what happens if rename failed. Do you need to remove the vm previous created? If leave the vm here, and return error to persona, persona cache will not have this container, but portlayer cache will cache a container with wrong display name and half created configuration.

lib/vicadmin/validate.go
```diff
@@ -209,14 +204,33 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
-	err = v.QueryDatastore(ctx, vch, sess)
+	err := v.QueryDatastore(ctx, vch, sess)
 	if err != nil {
 		log.Errorf("Had a problem querying the datastores: %s", err.Error())
 	}
 	v.QueryVCHStatus(vch, sess)
 	return v
 }
 
+func GetVCHName(ctx context.Context, sess *session.Session) string {
```
anchal-agrawal: Nit-pick: Go's canonical way of naming "getters" recommends excluding "Get" from the function name - so this one would be `VCHName`.

https://golang.org/doc/effective_go.html#Getters

This isn't a getter per se, but anyway.

tests/test-cases/Group1-Docker-Commands/1-04-Docker-Create.robot
```diff
@@ -120,8 +120,7 @@ Create a container with no command specified
 Create a container with custom CPU count
     ${rc}  ${id}=  Run And Return Rc And Output  docker %{VCH-PARAMS} create -it --cpuset-cpus 3 busybox
     Should Be Equal As Integers  ${rc}  0
-    ${rc}  ${id}=  Run And Return Rc And Output  govc datastore.ls |grep ${id}
-    Should Be Equal As Integers  ${rc}  0
+    ${id}=  Get VM display name  ${id}
```
emlin: is this supposed to be id or name?

emlin:
Still need to add test case in upgrade group, to check old container's display name is shown correctly

lib/vicadmin/validate.go
emlin: How? You have reset vchName at line 226

lib/config/executor/container_vm.go
```diff
@@ -44,6 +44,21 @@ type Common struct {
 	Notes string `vic:"0.1" scope:"hidden" key:"notes"`
 }
 
+// Common data (specifically for a containerVM) between managed entities, across execution environments.
+type CommonSpecForVM struct {
```
hmahmood: This is not really "common" any more since its only used by one struct, `ExecutorConfig`. You can just as well embed this like: 
```go
type ExecutorConfig struct {
  Common struct {
  // A reference to the components hosting execution environment, if any
   	ExecutionEnvironment string
  
 	// Unambiguous ID with meaning in the context of its hosting execution environment
 	ID string `vic:"0.1" scope:"read-only" key:"id"`
 
 	// Convenience field to record a human readable name
 	Name string `vic:"0.1" scope:"hidden" key:"name"`
 
 	// Freeform notes related to the entity
 	Notes string `vic:"0.1" scope:"hidden" key:"notes"`
   } `vic:"0.1" scope:"read-only" key:"common"`
```
At the very least, I think you should rename it to something like `ExecutorConfigCommon`.

tests/test-cases/Group1-Docker-Commands/1-04-Docker-Create.robot
chengwang86: @emlin `Get VM display name  ${id}` returns `containerName-containerShortID`. This test case should be fine. I have tested it.

lib/portlayer/exec/commit.go
hmahmood: Can we do the rename by setting it in the config spec (`h.Spec`)? This really should not be a separate step here, but should be done as part of the reconfigure.

lib/portlayer/exec/container.go
```diff
@@ -555,6 +558,37 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
 	return nil
 }
 
+// Update the VM display name on vSphere UI
+func (c *Container) UpdateDisplayName(ctx context.Context) error {
```
hmahmood: This may not be needed if we can just do the rename as part of a reconfigure (which is done by `Commit`).

lib/vicadmin/validate.go
hmahmood: I think it will be `""`. In any case you should return an error from this function if `err != nil`.

lib/config/executor/container_vm.go
```diff
@@ -108,7 +123,7 @@ type ContainerVM struct {
 // in that there is no process inherently associated - this is closer to a ThreadPool than a Thread and
 // is the owner of the shared filesystem environment. This is the guest visible complement to ContainerVM.
 type ExecutorConfig struct {
-	Common `vic:"0.1" scope:"read-only" key:"common"`
```
gigawhitlocks: So.. what happened to `Common` here? Or was `Common` renamed to `CommonSpecForVM` everywhere? Because if it is renamed everywhere, why was it renamed? I am skeptical that this longer name is any more semantically useful given the context of the object than the original, shorter name.

lib/config/executor/container_vm.go
gigawhitlocks: It's also exactly the same as `Common`. It's copy and pasted. @chengwang86 just use the original object.

lib/config/executor/container_vm.go
chengwang86: @gigawhitlocks The original common is still there, used by other objects. We just need to define a separate `common` for `executorConfig`. The reason is explained in my PR ^^^

pkg/version/plugin3/commonSpecForContainer.go
gigawhitlocks: @emlin with this code as written, would we wind up with objects actually named `NewExecutorConfig` elsewhere in the code?

lib/config/executor/container_vm.go
chengwang86: @gigawhitlocks See the scope of name `Name string `vic:"0.1" scope:"hidden" key:"name"``

lib/vicadmin/validate.go
```diff
@@ -209,14 +204,33 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
+	defer trace.End(trace.Begin(""))
+
+	var vchName = fmt.Sprintf("VCH")
```
sflxn: you can code this as defaultVchName := "VCH".  no need for fmt.Sprintf.

lib/vicadmin/validate.go
sflxn: if err != nil {
    vchName = defaultVchName
...
}

I think it's fine not to return err since the original cod just attempts to get a name and if not present, then just set to "VCH"

sflxn:
I don't think commonSpecForVCH.go and commonSpecForContainer.go should go into pkg/version.

Also, I think you can refactor those two into one file.  The main difference between them seems to be target manager from what I can see.  You can have the init() create one for VCH and one for container.

chengwang86:
@sflxn The reason that commonSpecForVCH.go and commonSpecForContainer.go are written in this way is here https://github.com/vmware/vic/blob/master/lib/migration/plugins/plugins.md, which is required by emma's data migration framework.

I think each plugin should be a separate file. @emlin comments?

lib/config/executor/container_vm.go
gigawhitlocks: It's still.. Idk, if there's no other way to do it, so be it, but it just kills me to have such similar structs with such similar names and even the same comments right next to each other. I mean I know we don't have classic inheritance but this just jumps out at me as begging for some sort of polymorphism or encapsulation to make the naming more succinct/clearer and/or to reuse code and not have to copy/paste..

But maybe it's not possible here, I don't know, it just seems like there should be a cleaner way. If not, I'll live.

lib/portlayer/exec/handle.go
```diff
@@ -283,6 +284,13 @@ func Create(ctx context.Context, sess *session.Session, config *ContainerCreateC
 		Metadata: config.Metadata,
 	}
 
+	// if not vsan, set the datastore folder name to containerID
+	if dsType, _ := sess.Datastore.Type(ctx); dsType != types.HostFileSystemVolumeFileSystemTypeVsan {
```
chengwang86: Thanks for pointing out this. I will handle this error.

lib/migration/plugins/init.go
emlin: hmm.. migrator is good, but I have used that in another place. 😢

cmd/vicadmin/vicadm.go
```diff
@@ -324,11 +326,25 @@ func listVMPaths(ctx context.Context, s *session.Session) ([]logfile, error) {
...
+
+			logname, err = child.Name(ctx)
+			if err != nil {
+				log.Errorf("Unable to get the vm name for %s: %s", child.Reference(), err)
+				continue
+			}
+		} else {
+			// for the appliance VM, id is in guestinfo.vice./init/common/id;
+			// however, this is not problematic since we don't collect these logs for the appliance VM
+			logname = ec[containerIDInExtraConfig]
```
emlin: even the container id key is not supposed to be changed, using this way to get configuration is still not recommended.
vm.DSPath can return the vm datastore folder name, no need to query extraconfig here.

lib/install/management/delete.go
```diff
@@ -161,23 +161,26 @@ func (d *Dispatcher) DeleteVCHInstances(vmm *vm.VirtualMachine, conf *config.Vir
...
 		return err
 	}
 
 	for _, child := range children {
-		name, err := child.Name(d.ctx)
-		if err != nil {
-			errs = append(errs, err.Error())
+		//Leave VCH appliance there until everything else is removed, cause it has VCH configuration. Then user could retry delete in case of any failure.
+		ok, err := d.isVCH(child)
+		if ok {
```
emlin: we usually check error first. switch this with the if err != nil

lib/portlayer/exec/handle.go
```diff
@@ -283,6 +284,19 @@ func Create(ctx context.Context, sess *session.Session, config *ContainerCreateC
 		Metadata: config.Metadata,
 	}
 
+	dsType, err := sess.Datastore.Type(ctx)
```
emlin: this is the place to do this: https://github.com/vmware/vic/blob/master/lib/spec/disk.go#L64
and session.go already provide method IsVSAN

cmd/vicadmin/vicadm.go
emlin: talked with Cheng offline. Here is to create log bundle folder for container, so vm name is better than container id. Do not need this change

lib/portlayer/exec/handle.go
chengwang86: @hickeng What do you think? You suggested we added the vsan logic here in handle.go.

lib/portlayer/logging/logging.go
```diff
@@ -135,11 +135,26 @@ func Join(h interface{}) (interface{}, error) {
...
+	}
+
+	logFilePath = fmt.Sprintf("[%s] %s", dsName, VMID)
+	if dsType == types.HostFileSystemVolumeFileSystemTypeVsan {
+		logFilePath = fmt.Sprintf("%s/%s", VMPathName, VMName)
+	}
 
 	for _, logFile := range []string{"tether.debug", "output.log"} {
-		filename := fmt.Sprintf("%s/%s/%s", VMPathName, VMName, logFile)
+		filename := fmt.Sprintf("%s/%s", logFilePath, logFile)
```
emlin: isn't the spec.VMPathName() already the one you want?

lib/portlayer/util/util.go
```diff
@@ -53,3 +57,15 @@ func ServiceURL(serviceName string) *url.URL {
 
 	return s
 }
+
+// Update the VM display name on vSphere UI
+func DisplayName(id, name string) string {
+	shortID := id[:shortIDLen]
+	nameMaxLen := maxVMNameLength - len(shortID)
+	prettyName := name
```
emlin: do we need to make it smarter? like if pretty name is shorter than nameMaxLen, can we make the id longer than a shortID

lib/portlayer/logging/logging.go
chengwang86: @emlin If it is on a vsan setup, the `VMPathName` would become `[datastore_name]`, whereas what we want here is `[datastore_name]/vmname`

pkg/version/plugin2/commonSpecForVCH.go
```diff
@@ -0,0 +1,128 @@
...
+				ID:                   oldStruct.ID,
+				ExecutionEnvironment: oldStruct.ExecutionEnvironment,
+				Notes:                oldStruct.Notes,
+			},
+		},
+	}
+
+	cfg := make(map[string]string)
+	extraconfig.Encode(extraconfig.MapSink(cfg), newStruct)
+	log.Debugf("The newStruct is %+v", newStruct)
```
emlin: feels like the log is duplicated with https://github.com/vmware/vic/pull/4134/files#diff-b6b8049fb355ddb0278cc831df513dcdR124

pkg/version/plugin3/commonSpecForContainer.go
```diff
@@ -0,0 +1,117 @@
...
+		UpdatedCommon: UpdatedCommon{
+			Name:                 oldStruct.Name,
+			ID:                   oldStruct.ID,
+			Notes:                oldStruct.Notes,
+			ExecutionEnvironment: oldStruct.ExecutionEnvironment,
+		}}
+
+	cfg := make(map[string]string)
+	extraconfig.Encode(extraconfig.MapSink(cfg), newStruct)
+	log.Debugf("The newStruct is %+v", newStruct)
```
emlin: ditto^^^

lib/portlayer/util/util.go
chengwang86: I think we can stick to the current approach now. In the future, maybe we could allow the customer himself to decide his container naming convention, as long as the name is within maxVMnameLength.

pkg/version/plugin2/commonSpecForVCH.go
chengwang86: Thanks Emma. This was for my own debug info. I will remove it.

tests/test-cases/Group11-Upgrade/11-01-Upgrade.md
```diff
@@ -17,9 +17,11 @@ This test requires that a vSphere server is running and available
 7. Roll back to the previous version
 8. Upgrade again to the upgraded version
 9. Check the previous created container and image are still there
+10. Check the previous created container's display name and datastore folder name
```
emlin: suggest to add case to check new container after upgrade will have different display name

lib/portlayer/logging/logging.go
emlin: vsan behavior is same, why before your change, it works well?

chengwang86: @emlin Sorry I might have misunderstood your question. Yes, I still use the VMPathName to assemble the log path for vsan setup, which is the same as before. The only difference is on non-vsan setup:

```
if dsType != types.HostFileSystemVolumeFileSystemTypeVsan {
		specconfig.VMPathName = fmt.Sprintf("[%s] %s/%s.vmx", sess.Datastore.Name(), specconfig.ID, specconfig.ID)
	}
```
So on non-vsan setup, the VMPathName has `...vmx` and we cannot use it to assemble the log path.

tests/resources/VCH-Util.robot
```diff
@@ -51,6 +51,11 @@ Set Test Environment Variables
     Run Keyword If  ${status}  Set Environment Variable  HOST_TYPE  ESXi
     Run Keyword Unless  ${status}  Set Environment Variable  HOST_TYPE  VC
 
+    ${about}=  Run  govc datastore.info %{TEST_DATASTORE} |grep 'Type'
```
gigawhitlocks: please put a space after that pipe

lib/vicadmin/validate.go
```diff
@@ -209,14 +204,36 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
+	}
+
+	self2 := vm.NewVirtualMachineFromVM(ctx, sess, self)
+	vchName, err := self2.Name(ctx)
+	if err != nil {
+		log.Errorf("Unable to get VCH name: %s", err)
+		log.Infof("Setting the VCH name to VCH")
+		return DefaultVCHName
+	}
+
```
gigawhitlocks: Please add a comment to document its functionality as well.

lib/portlayer/logging/logging.go
```diff
@@ -135,11 +135,29 @@ func Join(h interface{}) (interface{}, error) {
...
 	}
 
+	var logFilePath string
+
 	VMPathName := handle.Spec.VMPathName()
 	VMName := handle.Spec.Spec().Name
+	VMID := handle.Spec.ID()
+	dsName := handle.Spec.Datastore.Name()
+
+	//dsType, err := handle.Spec.Datastore.Type(context.Background())
```
gigawhitlocks: actually delete this code; do not leave it commented out. if it's checked in we can always get it out of the git history if necessary.

lib/vicadmin/validate.go
```diff
@@ -209,14 +204,36 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
 	return v
 }
 
+func VCHName(ctx context.Context, sess *session.Session) string {
+	defer trace.End(trace.Begin(""))
+
+	self, err := guest.GetSelf(ctx, sess)
+	if err != nil {
+		log.Errorf("Unable to get VCH name due to unknown self-reference: %s", err)
+		log.Infof("Setting the VCH name to VCH")
```
gigawhitlocks: Please replace the hardcoded "VCH" at the end with a reference to the constant so that if it is changed the log message is updated as well.

gigawhitlocks: `"Setting the VCH name to %s", DefaultVCHName`

lib/vicadmin/validate.go
```diff
@@ -209,14 +204,36 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
-	err = v.QueryDatastore(ctx, vch, sess)
+	err := v.QueryDatastore(ctx, vch, sess)
 	if err != nil {
 		log.Errorf("Had a problem querying the datastores: %s", err.Error())
 	}
 	v.QueryVCHStatus(vch, sess)
 	return v
 }
 
+func VCHName(ctx context.Context, sess *session.Session) string {
```
gigawhitlocks: OK so I have a few concerns about this function. For one, there's no comment string and it's named like a field so it's not clear what it does. Further, the log message at the end which states that it is "setting the VCH hostname" only makes sense in the context of its caller on line 116, where we _actually_ set the hostname. As written the function is a getter, but the log message claims it's a setter, and it's a function not a method so its name gives no hint as to which it is at all.

So! I would suggest transforming this function into a method which receives a `Validator` object, perform the assignment that is currently on line `116` inside this method instead, change the return type to `error`, and rename the method to `SetVCHName` and then on line 116 do: `if err = v.SetVCHName(); err != nil { /*handle error*/ } `

lib/vicadmin/validate.go
```diff
@@ -209,14 +204,37 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
+
+	newVM := vm.NewVirtualMachineFromVM(ctx, sess, self)
+	vchName, err := newVM.Name(ctx)
+	if err != nil {
+		log.Errorf("Unable to get VCH name: %s", err)
+		log.Infof("Setting the VCH name to %s", DefaultVCHName)
+		return DefaultVCHName
+	}
+
+	log.Infof("Setting the VCH name to %s", vchName)
```
gigawhitlocks: Please either move this log message out of this function and rename the function to GetVCHName since it is a function and not a method (the Get is implied if it's a method because it would be called like v.VCHName() but as a bare function its functionality is unclear with the name given) or make this function into a _method_ `func (v *Validator) SetVCHName(ctx, sess) err {` and do `v.Hostname = vchName` here so that the log message is correct. Currently someone could call `VCHName()` to get the name of the VCH and it will log "Setting the VCH name" when nothing of the sort is actually happening. You have a few ways you can go with this, but it can't remain as it is.

anchal-agrawal: I'll take the blame for the current func name (https://github.com/vmware/vic/pull/4134#discussion_r104963502) - it was initially named `GetVCHName`. I agree with @gigawhitlocks.

tests/resources/Docker-Util.robot
```diff
@@ -51,6 +51,20 @@ Get Container IP
...
+    [Return]  ${shortID}
+
+Get VM display name
+    [Arguments]  ${id}
+    ${rc}  ${name}=  Run And Return Rc And Output  docker %{VCH-PARAMS} inspect --format='{{.Name}}' ${id}
+    Should Be Equal As Integers  ${rc}  0
+    ${name}=  Get Substring  ${name}  1
+    ${shortID}=  Get container shortID  ${id}
+    ${vmName}=  Catenate  SEPARATOR=-  ${name}  ${shortID}
+    [Return]  ${vmName}
```
mhagen-vmware: The catenate is unnecessary, just Return ${vmName}-${shortID}

tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
```diff
@@ -104,6 +104,34 @@ Run Docker Checks
...
     Should Not Contain  ${output}  Exited (0)
 
+    # check the display name and datastore folder name of an existing container
+    ${rc}  ${output}=  Run Keyword If  '%{HOST_TYPE}' == 'VC'  Run And Return Rc And Output  govc vm.info %{VCH-NAME}/*-%{ID1}
+    Run Keyword If  '%{HOST_TYPE}' == 'VC'  Should Be Equal As Integers  ${rc}  0
+    Run Keyword If  '%{HOST_TYPE}' == 'VC'  Should contain  ${output}  vch-restart-tes-%{ID1}
+    ${rc}  ${output}=  Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Run And Return Rc And Output  govc vm.info *-%{ID1}
+    Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Should Be Equal As Integers  ${rc}  0
+    Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Should Contain  ${output}  vch-restart-tes-%{ID1}
+    ${rc}  ${output}=  Run And Return Rc And Output  govc datastore.ls | grep vch-restart-tes-%{ID1}
```
mhagen-vmware: does this need the VSAN check? and where did the vch-restart-tes come from, should it be vch-restart-test?

chengwang86: @mhagen-vmware For an existing container before VCH upgrade, the pretty name in the display name would be truncated if it is longer than 15 chars. So `vch-restart-test1` becomes `vch-restart-tes` and the display name becomes `vch-restart-tes-id`

lib/migration/plugins/init.go
caglar10ur: then lets rename all in a separate PR :)

lib/portlayer/util/util.go
```diff
@@ -24,6 +25,9 @@ import (
 const (
 	// XXX leaving this as http for now.  We probably want to make this unix://
 	scheme = "http://"
+
+	maxVMNameLength = 80
+	shortIDLen      = 12
```
caglar10ur: I think they are generic enough to deserve getting moved into lib/portlayer/constants/constants.go

lib/portlayer/logging/logging.go
```diff
@@ -135,11 +137,21 @@ func Join(h interface{}) (interface{}, error) {
...
 	}
 
+	var logFilePath string
+
 	VMPathName := handle.Spec.VMPathName()
 	VMName := handle.Spec.Spec().Name
 
+	logFilePath = fmt.Sprintf("%s/%s", VMPathName, VMName)
+	// on non-vsan setup, VMPathName is set to "[datastore_name] containerID/containerID.vmx"
+	if strings.Contains(VMPathName, "vmx") {
```
emlin: strings.HasSuffix(VMPathName, ".vmx")?

lib/migration/plugins/init.go
emlin: gonna need a new name for the original migrator as well

lib/vicadmin/validate.go
```diff
@@ -223,6 +219,35 @@ func (d dsList) Len() int           { return len(d) }
 func (d dsList) Swap(i, j int)      { d[i], d[j] = d[j], d[i] }
 func (d dsList) Less(i, j int) bool { return d[i].Name < d[j].Name }
 
+// Obtain the VCH name from vsphere
+func (v *Validator) SetVCHName(ctx context.Context, sess *session.Session) error {
```
hickeng: This is more GetVCHName than SetVCHName

lib/vicadmin/validate.go
```diff
@@ -62,8 +62,9 @@ type Validator struct {
 }
 
 const (
-	GoodStatus = template.HTML(`<i class="icon-ok"></i>`)
-	BadStatus  = template.HTML(`<i class="icon-attention"></i>`)
+	GoodStatus     = template.HTML(`<i class="icon-ok"></i>`)
+	BadStatus      = template.HTML(`<i class="icon-attention"></i>`)
+	DefaultVCHName = `VCH`
```
hickeng: Can we use something that makes it clear that it's a failure to resolve the name, rather than something that could be an actual name - otherwise we risk confusion if there's a conflict.

pkg/version/plugin2/commonSpecForVCH.go
```diff
@@ -0,0 +1,127 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package plugin2
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/vmware/vic/lib/migration/manager"
```
hickeng: In general, packages in `pkg` shouldn't be referencing packages in `lib` - pkg is for supporting code that isn't necessarily VIC specific.

I'm also unsure about the naming `plugin2`, and `plugin3` - this implies an ordering of some kind and doesn't provide any information about what/why they exist.

pkg/version/version.go
```diff
@@ -31,7 +31,7 @@ var (
 	State       string
 
 	// MaxPluginVersion must be increased to add new plugin and make sure the new plugin version is same to this value
-	MaxPluginVersion = 1
+	MaxPluginVersion = 3
```
hickeng: I'm going to have to look through the Migration code in more detail, but if this is actually a version and not something else, then this should be 2 at the most given we've only got one major version released.

lib/vicadmin/validate.go
chengwang86: @hickeng We already logged the failure if we cannot get the VCH name from vsphere. It will look weird if we set the defaultVCHName to something like `VCH_defaultName` or something that indicates the failure.

pkg/version/plugin2/commonSpecForVCH.go
chengwang86: @emlin What do you think? Can we move the plugins to the lib folder so that `pkg` does not reference packages in `lib`?

lib/vicadmin/validate.go
hickeng: Logging the failure requires looking at the logs - changing the name of the VCH to something potentially misleading when someone looks at the admin console is a different matter.
Looking weird is fine - we could also set it to something obviously not a name, such as "<unable to retrieve vch name from vsphere>"

PR #4139, opened 2017-03-03
Integrate nfs volume store impl with client

fdawg4l:
This is the beginning of the integration effort.  This change just unifies the interfaces and addresses some issues I found while integrating the client.

The store url wasn't being used in volume create.  Instead, the nfs
target's url was.  That's a bug.  Fixed.

The Target interface doesn't need a pointer to the target it's
implementing.

Paths in the volumesdir and metadatadir were relative to to the nfs
target's url path.  Mounts are already relative paths, since the
namespace doesn't extend to the nfs server.  That is /foo/bar/mnt is
only known to the nfs client, the server only cares about paths after
the mnt/ directory.  This was a bug and is now fixed.

Added some logging and refactored tests for the above changes.

### DIFF
--- a/lib/portlayer/storage/nfs/target.go
+++ b/lib/portlayer/storage/nfs/target.go
@@ -18,15 +18,19 @@ import (
 	"io"
 	"net/url"
 	"os"
+
+	"github.com/vmware/vic/pkg/trace"
 )
 
 // MountServer is an interface used to communicate with network attached storage.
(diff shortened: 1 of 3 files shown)


matthewavery:
lgtm, originally I had been avoiding the use of the `SelfLink` field. But it just makes sense to use it here.

fdawg4l:
lgtm

PR #4146, opened 2017-03-06
Migrate container configuration in portlayer in memory

### LINKED ISSUE #3610: port-layer: support in memory data migration for container configuration

emlin:
Fixes #3610 

- Migrate container configuration in memory if there is migration plugin registered
- In commit method, drop extraconfig change if data is migrated
- If data migration failed, keep partial migrated value, and keep migration error in container cache
- In docker inspect and docker ps show container status as "broken" if data migration failed
- If container status is broken
  - allow container stop
  - force delete even for "docker rm"

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -485,12 +485,12 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 	if config.ForceRemove {
 		c.containerProxy.Stop(vc, name, &secs, true)
 	} else {
-		status, err := c.containerProxy.Status(vc)
+		state, err := c.containerProxy.State(vc)
 		if err != nil {
 			return err
 		}
 		// force stop if container is broken to make sure container is deletable later
-		if status == ContainerError {
+		if state.Status == ContainerError {
 			c.containerProxy.Stop(vc, name, &secs, true)
 		}
 	}
@@ -546,12 +546,12 @@ func (c *Container) cleanupPortBindings(vc *viccontainer.VicContainer) error {
 				// port bindings were cleaned up by another operation.
 				continue
 			}
-			status, err := c.containerProxy.Status(vc)
+			state, err := c.containerProxy.State(cc)
 			if err != nil {
 				return fmt.Errorf("Failed to get container %q power state: %s",
 					mappedCtr, err)
 			}
-			if status == ContainerRunning {
+			if state.Running {
 				log.Debugf("Running container %q still holds port %s", mappedCtr, hPort)
 				continue
 			}
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -489,7 +489,7 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 		if err != nil {
 			return err
 		}
-		// force stop if container is broken to make sure container is deletable later
+		// force stop if container state is error to make sure container is deletable later
 		if state.Status == ContainerError {
 			c.containerProxy.Stop(vc, name, &secs, true)
 		}
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -485,12 +485,12 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 	if config.ForceRemove {
 		c.containerProxy.Stop(vc, name, &secs, true)
 	} else {
-		broken, err := c.containerProxy.IsBroken(vc)
+		status, err := c.containerProxy.Status(vc)
 		if err != nil {
 			return err
 		}
 		// force stop if container is broken to make sure container is deletable later
-		if broken {
+		if status == ContainerError {
 			c.containerProxy.Stop(vc, name, &secs, true)
 		}
 	}
@@ -546,12 +546,12 @@ func (c *Container) cleanupPortBindings(vc *viccontainer.VicContainer) error {
 				// port bindings were cleaned up by another operation.
 				continue
 			}
-			running, err := c.containerProxy.IsRunning(cc)
+			status, err := c.containerProxy.Status(vc)
 			if err != nil {
 				return fmt.Errorf("Failed to get container %q power state: %s",
 					mappedCtr, err)
 			}
-			if running {
+			if status == ContainerRunning {
 				log.Debugf("Running container %q still holds port %s", mappedCtr, hPort)
 				continue
 			}
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -481,8 +481,18 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 
 	// Use the force and stop the container first
 	secs := 0
+
 	if config.ForceRemove {
 		c.containerProxy.Stop(vc, name, &secs, true)
+	} else {
+		broken, err := c.containerProxy.IsBroken(vc)
+		if err != nil {
+			return err
+		}
+		// force stop if container is broken to make sure container is deletable later
+		if broken {
+			c.containerProxy.Stop(vc, name, &secs, true)
+		}
 	}
 
 	//call the remove directly on the name. No need for using a handle.
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -88,6 +88,7 @@ type VicContainerProxy interface {
 
 	Stop(vc *viccontainer.VicContainer, name string, seconds *int, unbound bool) error
 	IsRunning(vc *viccontainer.VicContainer) (bool, error)
+	IsBroken(vc *viccontainer.VicContainer) (bool, error)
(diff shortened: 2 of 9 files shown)


emlin:
@matthewavery added you as reviewer cause you ever worked on this issue.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -88,6 +88,7 @@ type VicContainerProxy interface {
 
 	Stop(vc *viccontainer.VicContainer, name string, seconds *int, unbound bool) error
 	IsRunning(vc *viccontainer.VicContainer) (bool, error)
+	IsBroken(vc *viccontainer.VicContainer) (bool, error)
```
matthewavery: Are we sure this will only be needed in the container endpoints? I am wondering if we will need this information for networks or volumes at some point.

lib/config/executor/container_vm.go
```diff
@@ -34,7 +34,7 @@ type Common struct {
 	// A reference to the components hosting execution environment, if any
 	ExecutionEnvironment string
 
-	// Unambiguous ID with meaning in the context of its hosting execution environment
+	// Unambiguous ID with meaning in the context of its hosting execution environment. Change this definition will cause container backward compatibility issue
```
matthewavery: please add "Please don't change this." I am not sure, you might add a "without talking to emma lin" to that. That way if someone does want to change it for  whatever reason they can talk to you about why they likely shouldn't.

matthewavery:
lgtm 😄 , other than the comment on the container ID field. Though that is simply a suggestion.

lib/apiservers/engine/backends/container_proxy.go
emlin: Right now, guess this is fine. 
This information is needed by container stop/start/rm, cause we still allow user to delete the broken container. But for all the other functions, if the configuration is available after half migration, those piece of function will looks good, it otherwise, that will not. The function is consistent.
We don't that care about the function of broken containers atm, as long as that's deletable.

emlin:
For easy to understand, here I posted the output of docker commands if there is one container broken for data migration error(the error if faked, I cannot have a real failure for that now):
```docker -H 10.160.58.173:2376 --tls ps -a
fc67c420adb3        busybox             "sh"                10 minutes ago      broken                                  wonderful_mestorf```
```docker -H 10.160.58.173:2376 --tls inspect fc
[
    {
        "Id": "fc67c420adb38c166c3402fe925c9f28b47c036ae43ff07d43338aa6f3933e1e",
        "Created": "2017-03-06T17:22:04Z",
        "Path": "sh",
        "Args": [],
        "State": {
            "Status": "broken",
            "Running": false,
            "Paused": false,
            "Restarting": false,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 0,
            "ExitCode": 0,
            "Error": "Migration failed: Faked migration error",
            "StartedAt": "2017-03-06T17:22:28Z",
            "FinishedAt": "2017-03-06T17:22:52Z"
        },
...
```

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -541,29 +543,44 @@ func (c *ContainerProxy) Stop(vc *viccontainer.VicContainer, name string, second
...
+	defer trace.End(trace.Begin(""))
+
+	inspectJSON, err := c.dockerInfo(vc)
 	if err != nil {
-		log.Errorf("containerInfoToDockerContainerInspect failed with %s", err)
 		return false, err
 	}
-
-	return inspectJSON.State.Running, nil
+	return inspectJSON.State.Status == "broken", nil
```
vburenin: Are there any statuses defined as constants? Using strings is not the best solution especially if it used more than 2 times across all project.
I also don't like IsBroken method. It's just awkward. There should be a method that returns status only it would be way more useful for other possible cases too.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -477,12 +478,13 @@ func (c *ContainerProxy) Stop(vc *viccontainer.VicContainer, name string, second
...
 	// TODO: can be replaced with simple cache check once power events are propagated to persona
-	running, err := c.IsRunning(vc)
+	inspectJSON, err := c.dockerInfo(vc)
 	if err != nil && IsNotFoundError(err) {
 		cache.ContainerCache().DeleteContainer(vc.ContainerID)
 		return err
 	}
-	if !running {
+	// attempt to stop container if status is running or broken
+	if !inspectJSON.State.Running && inspectJSON.State.Status != "broken" {
```
caglar10ur: can we make this string a const?

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -422,8 +422,15 @@ func convertContainerToContainerInfo(container *exec.ContainerInfo) *models.Cont
 	ccid := container.ExecConfig.ID
 	info.ContainerConfig.ContainerID = ccid
 
-	s := container.State().String()
-	info.ContainerConfig.State = s
+	var state string
+	if container.MigrationError != nil {
+		state = "broken"
```
caglar10ur: Ditto

lib/config/executor/container_vm.go
```diff
@@ -34,7 +34,7 @@ type Common struct {
 	// A reference to the components hosting execution environment, if any
 	ExecutionEnvironment string
 
-	// Unambiguous ID with meaning in the context of its hosting execution environment
+	// Unambiguous ID with meaning in the context of its hosting execution environment. Change this definition will cause container backward compatibility issue. Please don't change this.
```
caglar10ur: Change/Changing

Also the comment might be little bit more than this so that reader can understand why he/she shouldn't be changing this

lib/portlayer/exec/commit.go
```diff
@@ -146,6 +146,12 @@ func Commit(ctx context.Context, sess *session.Session, h *Handle, waitTime *int
 				log.Debugf("Nilifying ExtraConfig as we are running")
 				s.ExtraConfig = nil
 			}
+			// nilify ExtraConfig if container configuration is migrated
+			// in this case, VCH and container are in different version. Migrated configuration cannot be written back to old container, to avoid data loss in old version's container
+			if h.Migrated {
```
caglar10ur: We are setting ExtraConfig to nil if we are powered on so why do we need this?

emlin: This is because configuration data is migrated. If write back to container vmx file, the old binary cannot read back new data. We ever talked about that part in one meeting, remember?

lib/apiservers/engine/backends/container.go
```diff
@@ -481,8 +481,18 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
...
+
 	if config.ForceRemove {
 		c.containerProxy.Stop(vc, name, &secs, true)
+	} else {
+		state, err := c.containerProxy.State(vc)
+		if err != nil {
+			return err
+		}
+		// force stop if container is broken to make sure container is deletable later
+		if state.Status == ContainerError {
```
sflxn: This is checking for status "error".  Is this what you're using for "broken"?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -537,33 +543,31 @@ func (c *ContainerProxy) Stop(vc *viccontainer.VicContainer, name string, second
...
 
-// IsRunning returns true if the given container is running
-func (c *ContainerProxy) IsRunning(vc *viccontainer.VicContainer) (bool, error) {
+// State returns container state
+func (c *ContainerProxy) State(vc *viccontainer.VicContainer) (*types.ContainerState, error) {
 	defer trace.End(trace.Begin(""))
 
 	if c.client == nil {
-		return false, InternalServerError("ContainerProxy.IsRunning failed to get a portlayer client")
+		return nil, InternalServerError("ContainerProxy.dockerInfo failed to get a portlayer client")
```
sflxn: .State and not .dockerInfo

lib/apiservers/engine/backends/container.go
sflxn: So, I see where you're setting it on the portlayer side.  Since this is our state string, why not use a more explicit string, such as "migrating"?

emlin: yes

emlin: This is one error state caused by migration, not a working state as migrating. I think we might have other incorrect scenarios other than migration error, so I didn't use migration error as the state.

sflxn: Ok, I just got broken was an actual status.  It's actually a description, meaning the status is error.

Can you change the comment to say "if container is in an error state"?  Or is broken the best description?

sflxn:
small nitpicks, but LGTM

lib/apiservers/engine/backends/container.go
emlin: I'll update the comment.

PR #4148, opened 2017-03-06
[specific ci=1-01-Docker-Info] Show volume drivers in the plugins section of docker info

### LINKED ISSUE #3994: docker info does not show volume plugins information
**VIC version:**

0.8.0 / master / c8ac999

**Deployment details:**

./vic-machine-darwin create --target 10.161.5.44 --thumbprint ... --user root --no-tls --image-store datastore1 --volume-store datastore1:default --name vch-7540-no-tls

**Steps to reproduce:**
docker -H 10.161.19.231:2375 info

**Actual behavior:**
Plugins:
 Volume: 
 Network: bridge

**Expected behavior:**
I expect the list of supported plugins in the Volume line
Plugins:
 Volume: vsphere
 Network: bridge

I'm verifying volumes integration between Admiral and VIC, and I have noticed that supported volume plugins are not listed (although supported network plugins are). We rely on the list of plugins supported by the host in order to place the containers in a suitable host.

anchal-agrawal:
This change populates a list of supported volume drivers in the `Plugins: Volume` section of `docker info`:
```
$ sudo docker -H 192.168.77.225:2376 --tls info
Containers: 0
...
Plugins: 
 Volume: vsphere
...
```

Fixes #3994

### DIFF
--- a/lib/apiservers/engine/backends/volume.go
+++ b/lib/apiservers/engine/backends/volume.go
@@ -53,9 +53,11 @@ var validDriverOptsKeys = map[string]struct{}{
 	DriverArgImageKey:     {},
 }
 
+// Volume drivers currently supported. "local" is the default driver supplied by the client
+// and is equivalent to "vsphere" for our implementation.
 var supportedVolDrivers = map[string]struct{}{
-	"local":   struct{}{},
 	"vsphere": struct{}{},
+	"local":   struct{}{},
 }
 
 //Validation pattern for Volume Names
--- a/lib/apiservers/engine/backends/volume.go
+++ b/lib/apiservers/engine/backends/volume.go
@@ -53,6 +53,11 @@ var validDriverOptsKeys = map[string]struct{}{
 	DriverArgImageKey:     {},
 }
 
+var supportedVolDrivers = map[string]struct{}{
+	"local":   struct{}{},
+	"vsphere": struct{}{},
+}
+
 //Validation pattern for Volume Names
 var volumeNameRegex = regexp.MustCompile("^[a-zA-Z0-9][a-zA-Z0-9_.-]*$")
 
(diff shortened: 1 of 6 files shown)


sgairo:
I think you'll want to add the copyright to this .robot file because I think you might've kicked off on CI before my merge for copyright checker in .robot files went in etc.

lib/apiservers/engine/backends/volume.go
```diff
@@ -53,6 +53,11 @@ var validDriverOptsKeys = map[string]struct{}{
 	DriverArgImageKey:     {},
 }
 
+var supportedVolDrivers = map[string]struct{}{
+	"local":   struct{}{},
```
hickeng: We only support one driver currently. Check with @matthewavery why we have "local" - it may be that it's the default supplied by the client. We could do with a comment in this code to explain what the difference is between local & vsphere.

It's quite possible that we want to map local -> vsphere if specified (assuming my guess about it being the client's default value is correct), but I don't think we should return it in the list. Returning two drivers implies that there's a choice and that they have different behaviours.

matthewavery: @hickeng this is indeed the case. The docker CLI sends the "local" driver to the personality if none is specified. I included the "vsphere" portion when I wrote the code to demonstrate that we can take whatever argument we want from the `driver` that is supplied from the docker CLI. This allows us to have different behavior if we so chose, but would not line up at all with docker's use of drivers. +1 on the only returning one driver, and if we do honestly it should be local. If we do return vsphere then we should intercept local and always interpret it silently as vsphere as well since that is our native "local"ly supported method of storage.

anchal-agrawal: @hickeng @matthewavery I've changed the code to only display `vsphere` in `docker info`. If we want to display either `vsphere` or `local`, I think that displaying the former will help us differentiate between a VCH and other hosts in Admiral that support the `local` driver.

Issue #4100, opened 2017-03-01
Stand-alone implementation of Virtual Serial Port Concentrator


corrieb:
**User Statement:**

As a VIC developer, I need to be able to run and test Virtual Serial Port Concentrator in the form in which it will be integrated into the Port Layer, but as a stand-alone binary.

**Details:**

Proof of concept of the vSPC implementation to enable vMotion of connected serial ports was completed under #3937, written in Python. That was sufficient to demonstrate that this is the solution we want to pursue to satisfy #472.

This issue is to cover the actual implementation work required to create a telnet server with vSPC extensions which could then either be run as a stand-alone service or be integrated into the Port Layer.

**Acceptance Criteria:**

Demonstrate vMotion with a connected serial port using the stand-alone binary implemented here.

PR #4149, opened 2017-03-06
added an extensible telnet library that would support vSPC integration

kreamyx:
Added a telnet server that supports extending the telnet options and commands. This is related to issue #4100.

The library allows the user to negotiate extra options, sub-negotiate these options, and provide custom handlers to these extra commands.

An example of a stand-alone virtual serial port concentrator that uses this library can be found in https://github.com/kreamyx/go-vspc

This library along with the vSPC binary should conclude #4100

### DIFF
--- a/pkg/telnetlib/connection.go
+++ b/pkg/telnetlib/connection.go
@@ -1,3 +1,17 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package telnetlib
 
 import (
--- a/pkg/telnetlib/connection.go
+++ b/pkg/telnetlib/connection.go
@@ -1,0 +1,302 @@
+package telnetlib
+
+import (
+	"bytes"
+	"io"
+	"io/ioutil"
+	"net"
+	"sync"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+type connOpts struct {
+	conn        net.Conn
+	fsm         *telnetFSM
+	cmdHandler  CmdHandlerFunc
+	dataHandler DataHandlerFunc
+	serverOpts  map[byte]bool
+	clientOpts  map[byte]bool
+	optCallback func(byte, byte)
+}
+
+type TelnetConn struct {
+	conn              net.Conn
+	readCh            chan []byte
+	writeCh           chan []byte
+	unackedServerOpts map[byte]bool
+	unackedClientOpts map[byte]bool
+	//server            *TelnetServer
+	serverOpts         map[byte]bool
+	clientOpts         map[byte]bool
+	dataRW             io.ReadWriter
+	cmdBuffer          bytes.Buffer
+	fsm                *telnetFSM
+	fsmInputCh         chan byte
+	handlerWriter      io.Writer
+	cmdHandler         CmdHandlerFunc
+	dataHandler        DataHandlerFunc
+	dataHandlerCloseCh chan chan struct{}
+	dataWrittenCh      chan bool
+	optionCallback     func(byte, byte)
+	connReadDoneCh     chan chan struct{}
+	connWriteDoneCh    chan chan struct{}
+	negotiationDone    chan struct{}
+	closedMutex        sync.Mutex
+	closed             bool
+}
+
+// Safely read/write concurrently to the data Buffer
+// databuffer is written to by the FSM and it is read from by the dataHandler
+type dataReadWriter struct {
+	dataBuffer bytes.Buffer
+	dataMux    *sync.Mutex
+}
+
+func (drw *dataReadWriter) Read(p []byte) (int, error) {
+	drw.dataMux.Lock()
+	defer drw.dataMux.Unlock()
+	return drw.dataBuffer.Read(p)
+}
+
+func (drw *dataReadWriter) Write(p []byte) (int, error) {
+	drw.dataMux.Lock()
+	defer drw.dataMux.Unlock()
+	return drw.dataBuffer.Write(p)
+}
+
+type connectionWriter struct {
+	ch chan []byte
+}
+
+func (cw *connectionWriter) Write(b []byte) (int, error) {
+	cw.ch <- b
+	return len(b), nil
+}
+
+func newTelnetConn(opts connOpts) *TelnetConn {
+	tc := &TelnetConn{
+		conn:               opts.conn,
+		readCh:             make(chan []byte),
+		writeCh:            make(chan []byte),
+		unackedServerOpts:  make(map[byte]bool),
+		unackedClientOpts:  make(map[byte]bool),
+		cmdHandler:         opts.cmdHandler,
+		dataHandler:        opts.dataHandler,
+		dataHandlerCloseCh: make(chan chan struct{}),
+		dataWrittenCh:      make(chan bool),
+		serverOpts:         opts.serverOpts,
+		clientOpts:         opts.clientOpts,
+		optionCallback:     opts.optCallback,
+		fsmInputCh:         make(chan byte),
+		connReadDoneCh:     make(chan chan struct{}),
+		connWriteDoneCh:    make(chan chan struct{}),
+		negotiationDone:    make(chan struct{}),
+		closed:             false,
+	}
+	if tc.optionCallback == nil {
+		tc.optionCallback = tc.handleOptionCommand
+	}
+	tc.handlerWriter = &connectionWriter{
+		ch: tc.writeCh,
+	}
+	tc.dataRW = &dataReadWriter{
+		dataMux: new(sync.Mutex),
+	}
+	fsm := opts.fsm
+	fsm.tc = tc
+	tc.fsm = fsm
+	for k := range tc.serverOpts {
+		tc.unackedServerOpts[k] = true
+	}
+	for k := range tc.clientOpts {
+		tc.unackedClientOpts[k] = true
+	}
+	return tc
+}
+
+func (c *TelnetConn) connectionLoop() {
+	log.Debugf("Entered connectionLoop")
+	// this is the reading thread
+	go func() {
+		for {
+			select {
+			case readBytes := <-c.readCh:
+				for _, ch := range readBytes {
+					c.fsmInputCh <- ch
+				}
+
+			case ch := <-c.connReadDoneCh:
+				ch <- struct{}{}
+				return
+			}
+		}
+	}()
+	// this is the writing thread
+	go func() {
+		for {
+			select {
+			case writeBytes := <-c.writeCh:
+				c.conn.Write(writeBytes)
+			case ch := <-c.connWriteDoneCh:
+				ch <- struct{}{}
+				return
+			}
+		}
+	}()
+}
+
+// reads from the connection and dumps into the connection read channel
+func (c *TelnetConn) readLoop() {
+	defer func() {
+		log.Debugf("read loop closed")
+	}()
+	for {
+		buf := make([]byte, 4096)
+		n, err := c.conn.Read(buf)
+		if n > 0 {
+			log.Debug("read %d bytes from the TCP Connection %v", n, buf[:n])
+			c.readCh <- buf[:n]
+		}
+		if err != nil {
+			log.Debugf("connection read: %v", err)
+			c.Close()
+			break
+		}
+	}
+}
+
+func (c *TelnetConn) startNegotiation() {
+	for k := range c.serverOpts {
+		log.Infof("sending WILL %d", k)
+		c.unackedServerOpts[k] = true
+		c.sendCmd(WILL, k)
+	}
+	for k := range c.clientOpts {
+		log.Infof("sending DO %d", k)
+		c.unackedClientOpts[k] = true
+		c.sendCmd(DO, k)
+	}
+	select {
+	case <-c.negotiationDone:
+		log.Infof("Negotiation finished")
+		return
+	case <-time.After(10 * time.Second):
+		log.Infof("Negotiation failed. Exiting")
+		c.Close()
+		c.closed = true
+		return
+	}
+}
+
+// Close closes the telnet connection
+func (c *TelnetConn) Close() {
+	log.Infof("Closing the connection")
+	c.conn.Close()
+	c.closeConnLoopRead()
+	c.closeConnLoopWrite()
+	c.closeFSM()
+	c.closeDatahandler()
+	log.Infof("telnet connection closed")
+	c.closedMutex.Lock()
+	defer c.closedMutex.Unlock()
+	c.closed = true
+}
+
+func (c *TelnetConn) closeConnLoopRead() {
+	connLoopReadCh := make(chan struct{})
+	c.connReadDoneCh <- connLoopReadCh
+	<-connLoopReadCh
+	log.Infof("connection loop read-side closed")
+}
+
+func (c *TelnetConn) closeConnLoopWrite() {
+	connLoopWriteCh := make(chan struct{})
+	c.connWriteDoneCh <- connLoopWriteCh
+	<-connLoopWriteCh
+	log.Infof("connection loop write-side closed")
+}
+
+func (c *TelnetConn) closeFSM() {
+	fsmCh := make(chan struct{})
+	c.fsm.doneCh <- fsmCh
+	<-fsmCh
+}
+
+func (c *TelnetConn) closeDatahandler() {
+	dataCh := make(chan struct{})
+	c.dataHandlerCloseCh <- dataCh
+	<-dataCh
+}
+
+func (c *TelnetConn) sendCmd(cmd byte, opt byte) {
+	b := []byte{IAC, cmd, opt}
+	log.Infof("Sending command: %v %v", cmd, opt)
+	c.writeCh <- b
+}
+
+func (c *TelnetConn) handleOptionCommand(cmd byte, opt byte) {
+	if cmd == WILL || cmd == WONT {
+		if _, ok := c.clientOpts[opt]; !ok {
+			c.sendCmd(DONT, opt)
+			return
+		}
+
+		if _, ok := c.unackedClientOpts[opt]; ok {
+			delete(c.unackedClientOpts, opt)
+			if len(c.unackedClientOpts) == 0 && len(c.unackedServerOpts) == 0 {
+				close(c.negotiationDone)
+			}
+		} else {
+			c.sendCmd(DO, opt)
+		}
+	}
+
+	if cmd == DO || cmd == DONT {
+		if _, ok := c.serverOpts[opt]; !ok {
+			c.sendCmd(WONT, opt)
+			return
+		}
+		if _, ok := c.unackedServerOpts[opt]; ok {
+			log.Infof("removing from the unack list")
+			delete(c.unackedServerOpts, opt)
+			if len(c.unackedClientOpts) == 0 && len(c.unackedServerOpts) == 0 {
+				close(c.negotiationDone)
+			}
+		} else {
+			log.Infof("Sending WILL command")
+			c.sendCmd(WILL, opt)
+		}
+	}
+}
+
+func (c *TelnetConn) dataHandlerWrapper(w io.Writer, r io.Reader) {
+	defer func() {
+		log.Infof("data handler closed")
+	}()
+	for {
+		select {
+		case ch := <-c.dataHandlerCloseCh:
+			ch <- struct{}{}
+			return
+		case <-c.dataWrittenCh:
+			if b, err := ioutil.ReadAll(r); err == nil {
+				c.dataHandler(w, b, c)
+			}
+		}
+	}
+}
+
+func (c *TelnetConn) cmdHandlerWrapper(w io.Writer, r io.Reader) {
+	if cmd, err := ioutil.ReadAll(r); err == nil {
+		c.cmdHandler(w, cmd, c)
+	}
+}
+
+// IsClosed returns true if the connection is already closed
+func (c *TelnetConn) IsClosed() bool {
+	c.closedMutex.Lock()
+	defer c.closedMutex.Unlock()
+	return c.closed
+}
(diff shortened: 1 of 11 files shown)


mdubya66:
Should this go in pkg/vsphere? 

Do we need unit tests?

caglar10ur:
Bunch of comments, will review later;

1. Every file needs copyright header,
2. We need to refer the original repo if this is coming from somewhere,
3. If we wrote this library then we need tests,
4. I think telnetlib name needs to change but I really don't have a good suggestion
5. I also think this needs to go under vsphere as - judging from its description - this is not a full fledged standard telnet implementation

kreamyx:
I added the copyright headers. The tests are in fsm_test.go.
For vsphere, this library has no associations with vsphere, it is a custom telnet server that can be extended to do any functionality. The vSPC mentioned above uses this library to add the telnet extensions, but the vSPC is not part of this code.

pkg/telnetlib/fsm.go
```diff
@@ -0,0 +1,123 @@
...
+			nextState = dataState
+		} else if ch == IAC { // escaping IAC
+			nextState = subnegState
+			fsm.tc.cmdBuffer.WriteByte(ch)
+		} else {
+			nextState = errorState
+		}
+	case errorState:
+		nextState = dataState
+		log.Infof("Finite state machine is in an error state. This should not happen for correct telnel protocol syntax")
```
sgairo: Should this be telnet instead of telnel?

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,316 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
```
sgairo: Should these copyrights be 2017?

pkg/telnetlib/protocol.go
```diff
@@ -0,0 +1,94 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package telnetlib
+
+const (
+	/* Telnet Protocol Characters */
+	IAC  byte = 255 // "Interpret As Command"
```
fdawg4l: ALLCAPS constants are apparently not a `go` thing.  Can you CamelCase these?

pkg/telnetlib/server.go
```diff
@@ -0,0 +1,94 @@
...
+	opts := connOpts{
+		conn:        conn,
+		cmdHandler:  ts.CmdHandler,
+		dataHandler: ts.DataHandler,
+		serverOpts:  ts.ServerOptions,
+		clientOpts:  ts.ClientOptions,
+		fsm:         newTelnetFSM(),
+	}
+	tc := newTelnetConn(opts)
+	go tc.connectionLoop()
```
fdawg4l: Where do we drain these `go` routines?  We start them here in `Accept` and it doesnt look like they are bounded by a `sync.WaitGroup`.  Is it possible these will leak?

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,316 @@
...
+	defer drw.dataMux.Unlock()
+	return drw.dataBuffer.Write(p)
+}
+
+type connectionWriter struct {
+	ch chan []byte
+}
+
+func (cw *connectionWriter) Write(b []byte) (int, error) {
+	cw.ch <- b
```
fdawg4l: What if `ch` is closed?  Can that happen?

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,316 @@
...
+}
+
+// reads from the connection and dumps into the connection read channel
+func (c *TelnetConn) readLoop() {
+	defer func() {
+		log.Debugf("read loop closed")
+	}()
+	for {
+		buf := make([]byte, 4096)
+		n, err := c.conn.Read(buf)
```
fdawg4l: You can turn `conn` into a buffered `io.Reader` using `bufio.NewReader`.  Up to you to use it, but it's there. :-)

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,316 @@
...
+	connWriteDoneCh    chan chan struct{}
+	negotiationDone    chan struct{}
+	closedMutex        sync.Mutex
+	closed             bool
+}
+
+// Safely read/write concurrently to the data Buffer
+// databuffer is written to by the FSM and it is read from by the dataHandler
+type dataReadWriter struct {
+	dataBuffer bytes.Buffer
```
fdawg4l: Where is this assigned?  I don't see this used.  If it's part of the api to be used by fsm, should there be a `New...` func where this gets passed in?

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,316 @@
...
+	dataMux    *sync.Mutex
+}
+
+func (drw *dataReadWriter) Read(p []byte) (int, error) {
+	drw.dataMux.Lock()
```
fdawg4l: I'm guessing we can't read from the dataReadWriter AND write to it at the same time?  Can we use a RW lock and update a cursor so we can call Read more than once?

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,316 @@
...
+	conn        net.Conn
+	fsm         *telnetFSM
+	cmdHandler  CmdHandlerFunc
+	dataHandler DataHandlerFunc
+	serverOpts  map[byte]bool
+	clientOpts  map[byte]bool
+	optCallback func(byte, byte)
+}
+
+type TelnetConn struct {
```
fdawg4l: It's probably obvious what all of these are for just by their names, but (particularly the *Handler functions) not all of them are.  Can you add some comments as to what each does and where it's used.

fdawg4l:
I just noticed the `[skip ci]` tag.  Since this does include unit tests, it's worth going through ci.

pkg/telnetlib/fsm.go
kreamyx: good catch! thanks

pkg/telnetlib/connection.go
kreamyx: I will change these to 2017

pkg/telnetlib/protocol.go
kreamyx: yes you are right! I will change them.

pkg/telnetlib/server.go
kreamyx: these go routines are drained in the TelnetConn.Close()

pkg/telnetlib/connection.go
kreamyx: Good point! I will add more comments here.

pkg/telnetlib/connection.go
kreamyx: It is assigned in the newTelnetConn(), particularly tc.dataRW

pkg/telnetlib/connection.go
kreamyx: Yes you are right! Do you mean reading from more than one goroutine? The dataReadWriter has one reader, the data handler, and one writer, the fsm (both the dataHandler and the fsm are singletons per telnet connection). I also think that even if we have multiple readers, Read() will change the underlying bytes.Buffer which is not safe for multiple readers

pkg/telnetlib/connection.go
kreamyx: it should get closed when the telnet connection closes. I changed the TelnetConnection.handlerWriter to be an io.WriteCloser instead of io.Writer. When the TelnetConn.Close() is called, it closes this writer which in turns closes the channel and set it to nil. In the code above, I will always check that cw.ch is not nil before writing to it.

pkg/telnetlib/connection.go
kreamyx: Thanks :). Good to know!

pkg/telnetlib/connection.go
fdawg4l: I'm just curious- if there is a reader (data handler) and a writer (fsm), then why do we need a buffer here?  Why cant this just be for protecting (via the mutex) the io.ReadWriter?

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,345 @@
...
+	clientOpts  map[byte]bool
+	optCallback func(byte, byte)
+}
+
+// TelnetConn is the struct representing the telnet connection
+type TelnetConn struct {
+	// conn is the underlying connection
+	conn              net.Conn
+	readCh            chan []byte
+	writeCh           chan []byte
```
fdawg4l: The issue I guess I'm having (and it's really just curiosity) is why channels?  If there is a singleton  for reads and  another singleton for writes, why can't we just use io.Reader and io.Writer to/from the handlers?  Then you won't need the go routine below (connectionLoop) which basically implements io.Copy but for channels.

kreamyx: Yes! I re-thought the design of the components. There were definitely some components that does nothing more than relay data from component A to component B. With the latest commit, the FSM reads from the connection directly which eliminates the readLoop go routine and the readCh field in the TelnetConn struct. Same thing for the the read-side part of the connectionLoop. The writeCh (and the writeLoop goroutine) still exists though because there can be multiple writers to the connection. (to clarify, there is a single writer to the dataReadWriter.dataBuffer but there are multiple writers to the underlying TCP connection)

pkg/telnetlib/connection.go
kreamyx: I am not very clear what you mean here. To clarify, the dataBuffer is the shared io.ReadWriter. It is protected via the Mutex and it is shared between the FSM and the dataHandler. The FSM writes to it, and the dataHandler reads from it.

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,297 @@
...
+		log.Infof("sending DO %d", k)
+		c.unackedClientOpts[k] = true
+		c.sendCmd(Do, k)
+	}
+	c.clientOptsLock.Unlock()
+	select {
+	case <-c.negotiationDone:
+		log.Infof("Negotiation finished")
+		return
+	case <-time.After(10 * time.Second):
```
caglar10ur: Make this a const

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,297 @@
...
+
+func (c *TelnetConn) startNegotiation() {
+	c.srvrOptsLock.Lock()
+	for k := range c.serverOpts {
+		log.Infof("sending WILL %d", k)
+		c.unackedServerOpts[k] = true
+		c.sendCmd(Will, k)
+	}
+	c.srvrOptsLock.Unlock()
+	c.clientOptsLock.Lock()
```
caglar10ur: Do we need to hold server lock while client negotiates?

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,297 @@
...
+func (c *TelnetConn) Close() {
+	log.Infof("Closing the connection")
+	c.conn.Close()
+	c.closeConnLoopWrite()
+	c.closeDatahandler()
+	c.handlerWriter.Close()
+	log.Infof("telnet connection closed")
+	c.closedMutex.Lock()
+	defer c.closedMutex.Unlock()
+	c.closed = true
```
caglar10ur: just curious, why is it is OK to call c.conn.Close/closeConnLoopWrite/closeConnLoopWrite and closeConnLoopWrite without taking the lock but setting c.closed is not?

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,297 @@
...
+func (c *TelnetConn) closeDatahandler() {
+	dataCh := make(chan struct{})
+	c.dataHandlerCloseCh <- dataCh
+	<-dataCh
+}
+
+func (c *TelnetConn) sendCmd(cmd byte, opt byte) {
+	b := []byte{Iac, cmd, opt}
+	log.Infof("Sending command: %v %v", cmd, opt)
+	c.writeCh <- b
```
caglar10ur: c.writeCh <- []byte{Iac, cmd, opt}  maybe?

hmahmood:
Any reason we didn't use a third-party telnet library?

kreamyx:
@hmahmood yes there is! Available ones do not support telnet command extensions.

hickeng:
@kreamyx We've already spoken about the dataReadWriter vs `<-chan string` but I was playing around with the code to help me understand it. It looks like embedded the options struct is completely viable and reduced confusion for me:
```diff
diff --git a/pkg/telnetlib/connection.go b/pkg/telnetlib/connection.go
index 8a69859..e0eca33 100644
--- a/pkg/telnetlib/connection.go
+++ b/pkg/telnetlib/connection.go
@@ -26,52 +26,52 @@ import (
 )

 type connOpts struct {
-       conn        net.Conn
-       fsm         *telnetFSM
-       cmdHandler  CmdHandlerFunc
+       // conn is the underlying connection
+       conn net.Conn
+       // core telnet finite state machine
+       fsm *telnetFSM
+
+       // cmdHandler is the command handler for the telnet server
+       // it is a callback function when receiving commands issued by the telnet client
+       // it is wrapped by the cmdHandlerWrapper
+       cmdHandler CmdHandlerFunc
+
+       // dataHandler is the data handler of the telnet server
+       // it is a call back function when receiving data from the telnet client
+       // it is wrapped by the dataHandlerWrapper
        dataHandler DataHandlerFunc
-       serverOpts  map[byte]bool
-       clientOpts  map[byte]bool
+
+       // callBack function for receiving an option during negotiation
        optCallback func(byte, byte)
+
+       serverOpts map[byte]bool
+       clientOpts map[byte]bool
 }

 // TelnetConn is the struct representing the telnet connection
 type TelnetConn struct {
-       // conn is the underlying connection
-       conn net.Conn
+       connOpts
+
        // the connection write channel. Everything required to be written to the connection goes to this channel
-       writeCh           chan []byte
+       writeCh chan []byte
+
        srvrOptsLock      sync.Mutex
        unackedServerOpts map[byte]bool
+
        clientOptsLock    sync.Mutex
        unackedClientOpts map[byte]bool
        //server            *TelnetServer
-       serverOpts map[byte]bool
-       clientOpts map[byte]bool

        // dataRW is the data buffer. It is written to by the FSM and read from by the data handler
        dataRW        io.ReadWriter
        cmdBuffer     bytes.Buffer
-       fsm           *telnetFSM
        handlerWriter io.WriteCloser
-       // cmdHandler is the command handler for the telnet server
-       // it is a callback function when receiving commands issued by the telnet client
-       // it is wrapped by the cmdHandlerWrapper
-       cmdHandler CmdHandlerFunc
-
-       // dataHandler is the data handler of the telnet server
-       // it is a call back function when receiving data from the telnet client
-       // it is wrapped by the dataHandlerWrapper
-       dataHandler DataHandlerFunc

        // used in the dataHandlerWrapper to notify that the telnet connection is closed
        dataHandlerCloseCh chan chan struct{}
        // used in the dataHandlerWrapper to notify that data has been writeen to the dataRW buffer
        dataWrittenCh chan bool

-       // callBack function for receiving an option during negotiation
-       optionCallback func(byte, byte)
-
        // connWriteDoneCh closes the write loop when the telnet connection is closed
        connWriteDoneCh chan chan struct{}
        // negotiationDone notifies that telnet negotiation has been done
@@ -84,20 +84,20 @@ type TelnetConn struct {
 // Safely read/write concurrently to the data Buffer
 // databuffer is written to by the FSM and it is read from by the dataHandler
 type dataReadWriter struct {
-       dataBuffer bytes.Buffer
-       dataMux    *sync.Mutex
+       sync.Mutex
+       bytes.Buffer
 }

 func (drw *dataReadWriter) Read(p []byte) (int, error) {
-       drw.dataMux.Lock()
-       defer drw.dataMux.Unlock()
-       return drw.dataBuffer.Read(p)
+       drw.Lock()
+       defer drw.Unlock()
+       return drw.Read(p)
 }

 func (drw *dataReadWriter) Write(p []byte) (int, error) {
-       drw.dataMux.Lock()
-       defer drw.dataMux.Unlock()
-       return drw.dataBuffer.Write(p)
+       drw.Lock()
+       defer drw.Unlock()
+       return drw.Write(p)
 }

 type connectionWriter struct {
@@ -117,35 +117,26 @@ func (cw *connectionWriter) Close() error {
        return nil
 }

-func newTelnetConn(opts connOpts) *TelnetConn {
+func newTelnetConn(opts *connOpts) *TelnetConn {
        tc := &TelnetConn{
-               conn:               opts.conn,
+               connOpts:           *opts,
                writeCh:            make(chan []byte),
                unackedServerOpts:  make(map[byte]bool),
                unackedClientOpts:  make(map[byte]bool),
-               cmdHandler:         opts.cmdHandler,
-               dataHandler:        opts.dataHandler,
                dataHandlerCloseCh: make(chan chan struct{}),
                dataWrittenCh:      make(chan bool),
-               serverOpts:         opts.serverOpts,
-               clientOpts:         opts.clientOpts,
-               optionCallback:     opts.optCallback,
                connWriteDoneCh:    make(chan chan struct{}),
                negotiationDone:    make(chan struct{}),
                closed:             false,
        }
-       if tc.optionCallback == nil {
-               tc.optionCallback = tc.handleOptionCommand
+       if tc.optCallback == nil {
+               tc.optCallback = tc.handleOptionCommand
        }
        tc.handlerWriter = &connectionWriter{
                ch: tc.writeCh,
        }
-       tc.dataRW = &dataReadWriter{
-               dataMux: new(sync.Mutex),
-       }
-       fsm := opts.fsm
-       fsm.tc = tc
-       tc.fsm = fsm
+       tc.dataRW = &dataReadWriter{}
+       tc.fsm.tc = tc
        for k := range tc.serverOpts {
                tc.unackedServerOpts[k] = true
        }
@@ -276,7 +267,7 @@ func (c *TelnetConn) dataHandlerWrapper(w io.Writer, r io.Reader) {
                        ch <- struct{}{}
                        return
                case <-c.dataWrittenCh:
-                       if b, err := ioutil.ReadAll(r); err == nil {
+                       if b, _ := ioutil.ReadAll(r); len(b) > 0 {
                                c.dataHandler(w, b, c)
                        }
                }
@@ -284,7 +275,7 @@ func (c *TelnetConn) dataHandlerWrapper(w io.Writer, r io.Reader) {
 }

 func (c *TelnetConn) cmdHandlerWrapper(w io.Writer, r io.Reader) {
-       if cmd, err := ioutil.ReadAll(r); err == nil {
+       if cmd, _ := ioutil.ReadAll(r); len(cmd) > 0 {
                c.cmdHandler(w, cmd, c)
        }
 }
diff --git a/pkg/telnetlib/fsm.go b/pkg/telnetlib/fsm.go
index 861ffa1..e1cd447 100644
--- a/pkg/telnetlib/fsm.go
+++ b/pkg/telnetlib/fsm.go
@@ -101,7 +101,7 @@ func (fsm *telnetFSM) nextState(ch byte) state {
                fsm.tc.cmdBuffer.WriteByte(ch)
                opt := ch
                cmd := fsm.tc.cmdBuffer.Bytes()[0]
-               fsm.tc.optionCallback(cmd, opt)
+               fsm.tc.optCallback(cmd, opt)
                fsm.tc.cmdBuffer.Reset()
                nextState = dataState
        case subnegState:
diff --git a/pkg/telnetlib/server.go b/pkg/telnetlib/server.go
index bc89b6e..8784e84 100644
--- a/pkg/telnetlib/server.go
+++ b/pkg/telnetlib/server.go
@@ -90,7 +90,7 @@ func (ts *TelnetServer) Accept() (*TelnetConn, error) {
                clientOpts:  ts.ClientOptions,
                fsm:         newTelnetFSM(),
        }
-       tc := newTelnetConn(opts)
+       tc := newTelnetConn(&opts)
        go tc.writeLoop()
        go tc.dataHandlerWrapper(tc.handlerWriter, tc.dataRW)
        go tc.fsm.start()
```

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,297 @@
...
+	optCallback func(byte, byte)
+}
+
+// TelnetConn is the struct representing the telnet connection
+type TelnetConn struct {
+	// conn is the underlying connection
+	conn net.Conn
+	// the connection write channel. Everything required to be written to the connection goes to this channel
+	writeCh           chan []byte
+	srvrOptsLock      sync.Mutex
```
kreamyx: good to know! thanks for the link

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,297 @@
...
+	unackedServerOpts map[byte]bool
+	clientOptsLock    sync.Mutex
+	unackedClientOpts map[byte]bool
+	//server            *TelnetServer
```
kreamyx: good catch! yup

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+func (cw *connectionWriter) Write(b []byte) (int, error) {
+	if cw.ch != nil {
+		cw.ch <- b
+	}
+	return len(b), nil
+}
+
+func (cw *connectionWriter) Close() error {
```
kreamyx: it is defined as a WriteCloser. Are you suggesting to change the function name?

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,297 @@
...
+	close(cw.ch)
+	cw.ch = nil
+	return nil
+}
+
+func newTelnetConn(opts connOpts) *TelnetConn {
```
kreamyx: @caglar10ur like TCPConn, the only way to get an instance of TelnetConn is through the server's Accept() function. The user of the library should not be able to create a telnet connection in any different way.

For the ConnOpts, it is a small struct and it signals to the user of the library that the library is not going to change the values of the options that are passed in

pkg/telnetlib/connection.go
```diff
@@ -0,0 +1,297 @@
...
+		tc.optionCallback = tc.handleOptionCommand
+	}
+	tc.handlerWriter = &connectionWriter{
+		ch: tc.writeCh,
+	}
+	tc.dataRW = &dataReadWriter{
+		dataMux: new(sync.Mutex),
+	}
+	fsm := opts.fsm
+	fsm.tc = tc
```
kreamyx: Since a telnet connection is only instantiated through Accept(), then we are guaranteed that fsm is not nil

pkg/telnetlib/connection.go
kreamyx: Not necessary. They are protecting different data.

pkg/telnetlib/connection.go
kreamyx: @caglar10ur All of the Close methods are synchronized via channels. The closed bool flag in the telnet connection struct can be read via an exported method IsClosed. That's why it needs protection.

kreamyx:
@caglar10ur and @hickeng : all the changes based on your review comments will appear in https://github.com/vmware/vic/pull/4300

caglar10ur:
@kreamyx so are you going to drop this? If not please address those comments here so that we can merge this and then you can build on top of this in #4300

kreamyx:
@caglar10ur Done!

caglar10ur:
I'm confused, are you going to drop this PR? If so please close it so that I don't end up coming back to this in every 3 day :) Otherwise please drop the telnet lib changes from the other PR and rebased it on top of this

kreamyx:
@caglar10ur yes I will. It is much easier this way. Please look at #4300 :). I will close this one

PR #4154, opened 2017-03-06
NFS client integration with volumestore

fdawg4l:
This change implements the wrapper to the NFS client to use it as a volume store `MountServer` and `Target`.  Most of this change is integrating the unit test.

To run against an nfs server, run the following
`$ NFS_TEST_URL="nfs://127.0.0.1/home/fahmed/f" sudo -E go test -v .`


matthewavery:
lgmt, You might want to add the unit test invocation line to the current volume contribution documentations.

Issue #4110, opened 2017-03-01
Add session to live container config


hickeng:
**Story**
As a VIC developer I want to be able to add sessions to a containers configuration while it's live

**Detail**
We have had issues previously with being able to modify guestinfo config resulting in non-persistence of existing keys. We need to determine a means to:
* add a session without destroying the existing configuration
* add that session to the session index (maybe we need a secondary index that we check so we don't touch the base configuration)
* have the new session be non-persistent (ideally)

**Acceptance**
Session is added to containerVM
After power off/power on, the containerVM is still functional and launches only the original session

hickeng:
**Story**
As a user of the pkg/vsphere/extraconfig package I want to be able to encode only keys that can be changed with the VM powered on

**Detail**
guestinfo keys changed by the VMODL API when the VM is powered on become non-persistent - this means that they are deleted when the VM is powered off. This constraint means that we need to have a clear separation of data for portions of the config structure that will be updated while the VM is powered on, including indexes (such as for arrays or maps).
We are going to start with a simple approach:
1. add a `non-persistent` scope to the extraconfig package
2. add a scope filter to the Encode and EncodeWithPrefix calls so only keys matching the filter are encoded
3. update handle.Commit to apply an appropriate filter based on VM power state (power state changes are reflected in ChangeVersion)

**Acceptance**
Tests added to pkg/vsphere/extraconfig for filter behaviour
Add codepath to portlayer to allow for live config change

PR #4161, opened 2017-03-07
Allow extraconfig updates for running VM

hickeng:
This allows extraconfig updates for a running VM, filtering the keys to
those that are explicitly noted as non-persistent or are not guestinfo.

Also adds some make targets to the set of those that don't have dependency
memoization run so as to improve build time in CI.

Towards #4110


sgairo:
Do these copyright headers need to have the 2016-2017 range?

cgtexmex:
agree with @sgairo that copyright should be updated, but other than that LGTM

fdawg4l:
I don't understand the makefile change. Are the clean targets no longer cleaning up everything?

fdawg4l:
Is the change to extra config simply an optimization or did something functional change?

hickeng:
@fdawg4l The Makefile change is purely an optimization. In CI we call `make mark` and `make sincemark` at every build step. Each of those invocations would build the dynamic dependency graphs for the Go targets - this causes the dependency generation to be skipped for those targets.

PR #4173, opened 2017-03-08
Configure volumestores by URI [specific ci=1-19-Docker-Volume-Create]

fdawg4l:
Allows configuration of a volumestore by uri.  We check the URI and provision the volumestore accordingly.  

This will instantiate the nfs volume store if the appropriate scheme is passed.  `vic-machine` needs to be modified to pass this scheme, but that will come later.


sgairo:
copyright :)

matthewavery:
lgtm, looks great!

PR #4189, opened 2017-03-08
Limit the coverage plugin so that it doesn't inadvertently try to parse an inappropriate file

mhagen-vmware:
Fixes an issue we have seen in recent builds where the coverage plugin tried to parse ui files.


mdubya66:
lgtm

PR #4193, opened 2017-03-09
Reload tether when it receives HUP signal

caglar10ur:


fdawg4l:
I think my change munges this change a bit.  But it's fine.  We can talk about it when i push.  But LGTM.

caglar10ur:
I can drop this @fdawg4l if we are conflicting, I send this thinking that you are not going to touch tether and it was sitting in my branch :)

Issue #3761, opened 2017-01-30
Document that vic-machine validates bridge network range subnet mask


stuclem:
Per https://github.com/vmware/vic/issues/3212, which is implemented in PR https://github.com/vmware/vic/pull/3746. Affects the entry on [bridge-network-range](https://vmware.github.io/vic-product/assets/files/html/0.8/vic_installation/vch_installer_options.html#bridge-range) in VCH Deployment Options and might be mentioned elsewhere.

This is for Sprint 1.

stuclem:
PR https://github.com/vmware/vic/pull/3746 is merged.

stuclem:
@andrewtchin if this bridge network range is now validated during deployment, is this troubleshooting topic now redundant?
https://vmware.github.io/vic-product/assets/files/html/0.8/vic_app_dev/ts_bridge_pool_mask_error.html

Thanks!

stuclem:
Change is in https://github.com/vmware/vic/pull/4175/commits/a42b93791224c59481daabf5cd9f795b1f341e82

andrewtchin:
I would say that the troubleshooting page is now redundant

stuclem:
Removed in https://github.com/vmware/vic/pull/4196/commits/cb849d30f6707075b55286019857ed342fc4640f

Issue #3212, opened 2016-11-17
network validation does not block wrong bridge ip range


emlin:
This is one beta customer reported issue. They set bridge ip range, but portlayer network initialization failed with following error message
```
could not initialize port layer: bridge mask is not compatible with bridge pool mask
```

For this issue, at least vic-machine create should fail the deployment with reasonable error message. 
And right now bridge ip range is exposed, but bridge ip width is not, which can be exposed as well, to leave user more flexibility to configure bridge network.

andrewtchin:
As a customer I want VCH install to validate my network settings so that I get a correctly configured deployment or receive an error message if my configuration is incorrect.

mhagen-vmware:
lgtm

andrewtchin:
follow up issue is https://github.com/vmware/vic/issues/3771

PR #4196, opened 2017-03-09
Removed redundant troubleshooting topic from doc

stuclem:
The fix made in https://github.com/vmware/vic/issues/3212 makes this TS topic redundant.

Fixes https://github.com/vmware/vic/issues/3761


stuclem:
Thanks @andrewtchin

PR #4204, opened 2017-03-09
[specific ci=1-19-Docker-Volume-Create]Nfs tether mount work

### LINKED ISSUE #3812: Tether NFS Mount Implementation
User statement: 
As an engineer,  I need to modify the tether to support mounting nfs targets to specified locations in the container filesystem. This will allow usage of nfs volumes on containers in VIC. 

Details: The Current NFS mount code in the tether needs to be examined and modified to suit the needs created for epic #2303. The known amount of work for this piece is not complete and will be part exploratory in the beginning.

perceived needs/problems: 

* the communication of the nfs targets to the tether during init time.
* the container/tether actually being able to reach the nfs target(This likely adds a level of complexity to the work). 
* We must make sure that the switch root does not break anything we might need. 
* Some discussion must occur around failure scenarios such as the nfs target being reachable but non servicing at tether init time. 
* Modification to the mount code, the state is not fully known at this moment. 

Acceptance Criteria:
Successful mounting and usage of a nfs mount at the specified target location on containers.

matthewavery:
Fixes #3812 

This PR adds a number of things. These are listed below:

- **New operation in the tether.Ops interface found in the tether `interface.go` file. This operation is `MountFileSystem`. It was chosen to signify that we are not mounting a block device, but more specifically a filesystem.**

- **Implementations of said new operation now exist for each flavor of the tether. Though naturally two of them are `NOT IMPLEMENTED`**

- **Some slight changes have been made to the nfs device backing struct to make sure that the url return from `DiskPath()` is indeed marked with a scheme of `nfs`**

- **A new version of `VolumeJoin` has been written for the nfs portion of the storage layer within the portlayer libraries. The meat of this function is that we do not need to add a new device to the device config of the vm since we are not mounting a block device. It involves the creation and addition of the mountspec to the exec config of the container.**

- **A refactor occurred for the previous `vsphere` version of VolumeJoin to make the code more readable and understandable as it is.**

- **Changes to the VolumeJoin Handler in the portlayer storage handler(the communication endpoint between docker personality and portlaer) were also needed to distinguish between which library based VolumeJoin to call. This is also done on Scheme based on the changes that @fdawg4l made for the  portlayer handler work in pr #4173**

The remaining work at this point should just be the vic-machine related work for validating proper nfs uri inputs as well as making sure the nfs endpoints are communicated effectively to the configure portion of the portlayer. I have tested this change by hand by modifying the `vmx` file of the containers and restarting them. This took significant time and was quite error prone due to the need of an available file target on an otherwise unknown filesystem(since it is a container). Further hand testing will be done against this piece during the `vic-mcahine` related work as well.

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
@@ -505,9 +505,9 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
 	}
 
 	switch volume.Device.DiskPath().Scheme {
-	case "nfs":
+	case nfsScheme:
 		actualHandle, err = nfs.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
-	case "ds":
+	case dsScheme:
 		actualHandle, err = vsphere.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
 	default:
 		err = fmt.Errorf("unknown scheme (%s) for Volume (%s)", volume.Device.DiskPath().Scheme, *volume)
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
@@ -16,6 +16,7 @@ package handlers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -491,7 +492,7 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
 	//Note: Name should already be populated by now.
 	volume, err := h.volumeCache.VolumeGet(op, params.Name)
 	if err != nil {
-		log.Errorf("Volumes: StorageHandler : %#v", err)
+		op.Errorf("Volumes: StorageHandler : %#v", err)
 
 		return storage.NewVolumeJoinInternalServerError().WithPayload(&models.Error{
 			Code:    http.StatusInternalServerError,
@@ -499,9 +500,18 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
 		})
 	}
 
-	actualHandle, err = vsphere.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
+	switch volume.Device.DiskPath().Scheme {
+	case "nfs":
+		actualHandle, err = nfs.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
+	case "ds":
+		actualHandle, err = vsphere.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
+	default:
+		errMsg := fmt.Sprintf("Unkown scheme (%s) for Volume (%s)", volume.Device.DiskPath().Scheme, *volume)
+		err = errors.New(errMsg)
+	}
+
 	if err != nil {
-		log.Errorf("Volumes: StorageHandler : %#v", err)
+		op.Errorf("Volumes: StorageHandler : %#v", err)
 
 		return storage.NewVolumeJoinInternalServerError().WithPayload(&models.Error{
 			Code:    http.StatusInternalServerError,
@@ -509,7 +519,7 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
 		})
 	}
 
-	log.Infof("volume %s has been joined to a container", volume.ID)
+	op.Infof("volume %s has been joined to a container", volume.ID)
 	return storage.NewVolumeJoinOK().WithPayload(actualHandle.String())
 }
 
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
@@ -16,7 +16,6 @@ package handlers
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
@@ -43,6 +43,11 @@ type StorageHandlersImpl struct {
 	volumeCache *spl.VolumeLookupCache
 }
 
+const (
+	nfsScheme = "nfs"
+	dsScheme  = "ds"
+)
+
 // Configure assigns functions to all the storage api handlers
 func (h *StorageHandlersImpl) Configure(api *operations.PortLayerAPI, handlerCtx *HandlerContext) {
 	var err error
@@ -102,7 +107,7 @@ func (h *StorageHandlersImpl) configureVolumeStores(op trace.Operation, handlerC
 	// Each volume store name maps to a datastore + path, which can be referred to by the name.
 	for name, dsurl := range spl.Config.VolumeLocations {
 		switch dsurl.Scheme {
-		case "nfs":
+		case nfsScheme:
 			uid := nfs.DefaultUID
 
 			if dsurl.User != nil && dsurl.User.Username() != "" {
@@ -119,7 +124,7 @@ func (h *StorageHandlersImpl) configureVolumeStores(op trace.Operation, handlerC
 				return err
 			}
 
-		case "ds":
+		case dsScheme:
 			ds, err := datastore.NewHelperFromURL(op, handlerCtx.Session, dsurl)
 			if err != nil {
 				return fmt.Errorf("cannot find datastores: %s", err)
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
@@ -506,7 +506,7 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
 	case "ds":
 		actualHandle, err = vsphere.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
 	default:
-		errMsg := fmt.Sprintf("Unknown scheme (%s) for Volume (%s)", volume.Device.DiskPath().Scheme, *volume)
+		errMsg := fmt.Sprintf("unknown scheme (%s) for Volume (%s)", volume.Device.DiskPath().Scheme, *volume)
 		err = errors.New(errMsg)
 	}
 
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
@@ -1,4 +1,4 @@
-// Copyright 2017 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
@@ -506,7 +506,7 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
 	case "ds":
 		actualHandle, err = vsphere.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
 	default:
-		errMsg := fmt.Sprintf("Unkown scheme (%s) for Volume (%s)", volume.Device.DiskPath().Scheme, *volume)
+		errMsg := fmt.Sprintf("Unknown scheme (%s) for Volume (%s)", volume.Device.DiskPath().Scheme, *volume)
 		err = errors.New(errMsg)
 	}
 
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
@@ -506,8 +506,7 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
 	case "ds":
 		actualHandle, err = vsphere.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
 	default:
-		errMsg := fmt.Sprintf("unknown scheme (%s) for Volume (%s)", volume.Device.DiskPath().Scheme, *volume)
-		err = errors.New(errMsg)
+		err = fmt.Errorf("unknown scheme (%s) for Volume (%s)", volume.Device.DiskPath().Scheme, *volume)
 	}
 
 	if err != nil {
--- a/lib/portlayer/storage/nfs/vm.go
+++ b/lib/portlayer/storage/nfs/vm.go
@@ -25,11 +25,24 @@ import (
 )
 
 const (
+	//Below is a list of options included in the mount options and a brief reason why.
+	// "rw" : this indicates that the mount is for reading and writing to/from the mounted filesystem. NOTE: this could be configurable in the future.
+	// "noatime" : this option prevents read's from triggering a write to update the accesstime of an inode. Helps with efficient reads
+	// "vers=3" : we want to use NFSv3 it is simpler to implement and we do not need all the features of NFSv4 at this time
+	// "rsize=131072" : indicates the maximum read size for data on the NFS server. If the rsize is too big for either the client or server a negotion will occur to determine a supportable size. the value chosen is a default for /bin/mount for ubuntu 16.04
+	// "wsize=131072" : indicates the maximum write size for data on the NFS server. Like rsize it is negotiated if the value is too large. the value chosen is a default for /bin/mount for ubuntu 16.04
+	// "hard" : implies that we will retry indefinitely upon a failed transmission of data. this was agreed upon to indicate that problems have occurred with the mount if a hang on a write occurs.
+	// "proto=tcp" : tcp is a realiable protocol. The client used by the VCH also uses TCP as it's protocol. meaning we gain consistency in the communication between the tether->nfs and portlayer->nfs
+	// "timeo=600" : 600 deciseconds. This means a 60 second timout. With "hard" enabled this option likely does not matter.
+	// "sec=sys" : this means the NFS client uses the AUTH_SYS security flavor for all NFS requests on this mount point. This requires UID and GID of the user for permissions. also allows squashing permissions
+	// "mountvers=3" : this is listed as the RPC bind version. However, it is listed as a default by /bin/mount even when RPC is not the protocol used.
+	// "mountProto=TCP" : since the VCH uses TCP we should be using it as well here on the tether. Additionally, the mountProto does effect the initial protocol used for interacting with an nfs server. Keeping everything as the same protocol makes protocol issues easier to detect.
+	// "local_lock=none" : Specifies whether locks are local to the nfs client. We have set this to none since the same volume could be shared among several containers.
 	nfsMountOptions = "rw,noatime,vers=3,rsize=131072,wsize=131072,hard,proto=tcp,timeo=600,sec=sys,mountvers=3,mountproto=tcp,local_lock=none"
 )
 
 func VolumeJoin(op trace.Operation, handle *exec.Handle, volume *storage.Volume, mountPath string, diskOpts map[string]string) (*exec.Handle, error) {
-	defer trace.End(trace.Begin("nfs.VolumeJoin"))
+	defer trace.End(trace.Begin(fmt.Sprintf("handle.ID(%s), volume(%s), mountPath(%s)", handle.ExecConfig.ID, volume.ID, mountPath)))
 
 	if _, ok := handle.ExecConfig.Mounts[volume.ID]; ok {
 		return nil, fmt.Errorf("Volume with ID %s is already in container %s's mountspec config", volume.ID, handle.ExecConfig.ID)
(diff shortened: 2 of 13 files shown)


lib/tether/ops_darwin.go
```diff
@@ -53,6 +54,14 @@ func (t *BaseOperations) MountLabel(ctx context.Context, label, target string) e
 	return errors.New("not implemented on OSX")
 }
 
+// MountFileSystem performs a mount based on teh target path from the source url
```
sgairo: "the target"

sgairo:
copyrights :)

lib/portlayer/storage/nfs/vm.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func VolumeJoin(op trace.Operation, handle *exec.Handle, volume *storage.Volume, mountPath string, diskOpts map[string]string) (*exec.Handle, error) {
+	defer trace.End(trace.Begin("nfs.VolumeJoin"))
+
+	if _, ok := handle.ExecConfig.Mounts[volume.ID]; ok {
+		return nil, fmt.Errorf("Volume with ID %s is already in container %s's mountspec config", volume.ID, handle.ExecConfig.ID)
+	}
+
+	//constuct MountSpec for the tether
```
sgairo: "construct"

lib/portlayer/storage/nfs/vm.go
```diff
@@ -0,0 +1,58 @@
...
+	if handle.ExecConfig.Mounts == nil {
+		handle.ExecConfig.Mounts = make(map[string]executor.MountSpec)
+	}
+	handle.ExecConfig.Mounts[volume.ID] = mountSpec
+
+	return handle, nil
+}
+
+func createMountSpec(host url.URL, mountPath string, diskOpts map[string]string) executor.MountSpec {
+	deviceMode := nfsMountOptions + ",addr=" + host.Host
```
anchal-agrawal: We can replace this with
```go
deviceMode := fmt.Sprintf("%s,addr=%s", nfsMountOptions, host.Host)
```

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -454,25 +456,34 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
...
 	}
 
-	actualHandle, err = vsphere.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
+	switch volume.Device.DiskPath().Scheme {
+	case "nfs":
+		actualHandle, err = nfs.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
+	case "ds":
+		actualHandle, err = vsphere.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
+	default:
+		errMsg := fmt.Sprintf("Unknown scheme (%s) for Volume (%s)", volume.Device.DiskPath().Scheme, *volume)
```
anchal-agrawal: +1 for separating this into 2 lines for readability. Error strings should start with lowercase though 😉

lib/tether/interfaces.go
```diff
@@ -35,6 +36,7 @@ type Operations interface {
 	SetupFirewall() error
 	Apply(endpoint *NetworkEndpoint) error
 	MountLabel(ctx context.Context, label, target string) error
+	MountFileSystem(ctx context.Context, source url.URL, target string, mountOptions string) error
```
fdawg4l: The name is redundant.  You can only mount filesystems.  How about `MountTarget`?  Open to suggestions.  But MountFilesystem sounds redundant.

lib/tether/ops_linux.go
```diff
@@ -661,13 +663,32 @@ func (t *BaseOperations) MountLabel(ctx context.Context, label, target string) e
...
 
 	return nil
 }
 
+// MountFileSystem performs a mount based on the target path from the source url
+// This assumes that the source url is valid and available.
+func (t *BaseOperations) MountFileSystem(ctx context.Context, source url.URL, target string, mountOptions string) error {
+	defer trace.End(trace.Begin(fmt.Sprintf("Mounting %s on %s", source.String(), target)))
+
+	if err := os.MkdirAll(target, 0600); err != nil {
```
fdawg4l: `0600`?  A non-root user won't be able to `cd` into this directory.  I think `0644` is generally more preferred.

lib/tether/ops_linux.go
```diff
@@ -661,13 +663,32 @@ func (t *BaseOperations) MountLabel(ctx context.Context, label, target string) e
...
+		// same as MountLabel error for consistency
+		return fmt.Errorf("unable to create mount point %s: %s", target, err)
+	}
+
+	if err := Sys.Syscall.Mount(source.String(), target, "nfs", 0, mountOptions); err != nil {
```
fdawg4l: If I pass a url to protocol://somehost/foo.bar, the `.String()` method, i thought, includes the protocol.  Does the mount syscall take URLs??

lib/tether/tether.go
```diff
@@ -236,12 +236,17 @@ func (t *tether) setNetworks() error {
...
 func (t *tether) setMounts() error {
 	for k, v := range t.config.Mounts {
-		if v.Source.Scheme != "label" {
+		switch v.Source.Scheme {
+		case "label":
+			// this could block indefinitely while waiting for a volume to present
+			t.ops.MountLabel(context.Background(), v.Source.Path, v.Path)
+
+		case "nfs":
+			t.ops.MountFileSystem(context.Background(), v.Source, v.Path, v.Mode)
```
fdawg4l: `MountFileSystem(ctx context.Context, source url.URL, target string, mountOptions string)`

That means you're passing the `Mode` as the mountOptions?  I thought it needed to be encoded or something.  Maybe adding a comment in the MountFileSystem call for what form the mountOptions take is what's missing.

lib/tether/ops_linux.go
```diff
@@ -661,13 +663,32 @@ func (t *BaseOperations) MountLabel(ctx context.Context, label, target string) e
...
+		detail := fmt.Sprintf("mounting %s on %s failed: %s", source.String(), target, err)
+		return errors.New(detail)
```
fdawg4l: I get why you'd want to make a string out of the error, but the error returned can be type asserted to an errno and doing this hides that.  Can you just add a log message when this happens but return the error as is here?

lib/portlayer/storage/nfs/vm.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/vmware/vic/lib/config/executor"
+	"github.com/vmware/vic/lib/portlayer/exec"
+	"github.com/vmware/vic/lib/portlayer/storage"
+	"github.com/vmware/vic/pkg/trace"
+)
+
+const (
+	// TODO: pick potentially more sane defaults for these values
+	nfsMountOptions = "rw,noatime,vers=3,rsize=131072,wsize=131072,namlen=255,hard,proto=tcp,timeo=600,retrans=2,sec=sys,mountvers=3,mountproto=tcp,local_lock=none"
```
fdawg4l: Where did these options come from?  Are they preferred?

`namelen` seems bad.  Can you drop that?  Also retrans 2 seems low but I'm happy to be wrong if this is recommended somewhere.  Also, what is local_lock?  Does that mean `lockd`?

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -454,25 +456,34 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
...
 	}
 
-	actualHandle, err = vsphere.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
+	switch volume.Device.DiskPath().Scheme {
+	case "nfs":
+		actualHandle, err = nfs.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
+	case "ds":
+		actualHandle, err = vsphere.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
+	default:
+		errMsg := fmt.Sprintf("unknown scheme (%s) for Volume (%s)", volume.Device.DiskPath().Scheme, *volume)
```
fdawg4l: `fmt.Errorf`

lib/portlayer/storage/nfs/vm.go
fdawg4l: @anchal-agrawal i think i like what's there better.  Why use `printf` when you don't have to. :-)

lib/portlayer/storage/nfs/volume.go
```diff
@@ -107,6 +107,7 @@ func (v *VolumeStore) VolumeCreate(op trace.Operation, ID string, store *url.URL
 	}
 
 	u, _ := v.Service.URL()
+	u.Scheme = "nfs"
```
fdawg4l: This is a bug in `Service.URL()` then.  Can you make the change there instead of here?  This seems error prone and out of place.

lib/tether/ops_linux.go
matthewavery: @fdawg4l sounds good to me, we should probably change the for `MountLabel` as well.

lib/tether/ops_linux.go
matthewavery: That is a good question. I will check this. Might be safer to just put `Host` though if the port is important then we will need to find a way to put that in as well.

lib/tether/ops_linux.go
matthewavery: sure, We can keep the string as the error. and just return the error here.

lib/tether/tether.go
matthewavery: Possible yes, in my small test program I passed a string in directly and it seemed to be happy with that.

lib/portlayer/storage/nfs/vm.go
matthewavery: Ah 😄 I was waiting for this comment. So yes `namelen` can go. 2 was the default for the `/bin/mount` if you do not specify we can try more. What would you suggest? `local_lock=none` is also a default for `/bin/mount` so this I will read up on and come back with an answer.

lib/portlayer/storage/nfs/volume.go
matthewavery: yeah I felt that way as well. I will look at it in just a moment. I wanted to hand test and forget to inspect this more.

lib/tether/ops_linux.go
fdawg4l: Assuming this reply was for the review comment above.  If so, then Mountlabel sounds fine since you're mounting a filesystem by a filesystem label.

lib/tether/ops_linux.go
fdawg4l: :-).  You can't define the port in NFS.  You ask the server (via the `portmapper`) what port to use to talk to the nfs server. The client can't assume.

lib/portlayer/storage/nfs/volume.go
fdawg4l: IIRC this is assigned [here](https://github.com/vmware/vic/blob/master/lib/portlayer/storage/nfs/target.go#L73).  Just check the `Scheme` there and throw an error if it's not `nfs`.  Then you can drop this line here.

lib/tether/interfaces.go
matthewavery: sorry got pulled away for a little bit. `MountTarget` is good with me.

lib/tether/ops_linux.go
matthewavery: I meant that mount label is using `0600` and so that should likely change as well.

lib/tether/ops_linux.go
matthewavery: gotcha, so MountPort is not an option that we set. I see. I was confused on how that was decided when I started to think about it more.

lib/tether/tether.go
matthewavery: I mean it does need some sort of encoding I am sure, But I am not sure if the raw syscall will take care of that or not. There isn't a provided encoding that I know of for nfs.

matthewavery: Confirmed that the input for that part of `syscall.Mount` is `... data string)`

matthewavery: Looks like they always call this from inside the `Syscall.Mount`:

```datap, err := BytePtrFromString(data)```

before using the data.

File: lib/apiservers/portlayer/restapi/handlers/storage_handlers.go [hunk 13]
```diff
@@ -454,25 +456,33 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
...
+		op.Errorf("Volumes: StorageHandler : %#v", err)
 
 		return storage.NewVolumeJoinInternalServerError().WithPayload(&models.Error{
 			Code:    http.StatusInternalServerError,
 			Message: err.Error(),
 		})
 	}
 
-	actualHandle, err = vsphere.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
+	switch volume.Device.DiskPath().Scheme {
```
fdawg4l: Can we move this into the volume cache?  Above we do a `VolumeGet`.  We should be able to do an `h.volumeCache.VolumeJoin` on the ID used to do the get.   The cache can then do a lookup on the ID, find the volumestorer impementation, and call volume join on the right implementation.

I feel like that's much cleaner than we we have here.  I don't want to leak details about nfs or vsphere in the swagger handler like we're doing here.  If you make a new issue, we can get to it next sprint.

matthewavery: I agree, though I am going to make a ticket for the next sprint for this. Since I would like to get this work in and take care of the vic-machine work. that will give me time to test end to end and possibly get some initial integration tests out of the way.

fdawg4l:
Nice work.  Make it so.

mdubya66:
yes, nice work. lgtm.

matthewavery:
Thanks, now I just need to get CI to actually pull the correct changes... 😞

matthewavery:
@caglar10ur only your approval remains to be needed 😄

lib/portlayer/storage/nfs/vm.go
```diff
@@ -0,0 +1,57 @@
...
+	"net/url"
+
+	"github.com/vmware/vic/lib/config/executor"
+	"github.com/vmware/vic/lib/portlayer/exec"
+	"github.com/vmware/vic/lib/portlayer/storage"
+	"github.com/vmware/vic/pkg/trace"
+)
+
+const (
+	nfsMountOptions = "rw,noatime,vers=3,rsize=131072,wsize=131072,hard,proto=tcp,timeo=600,sec=sys,mountvers=3,mountproto=tcp,local_lock=none"
```
caglar10ur: How do we choose those parameters? r/wsize might be problematic for some setups. Also hard will cause everything to block on a failure with no way to interrupt it. Is that the behaviour we want?

Also I really don't like free form strings so please convert them to individual constants and craft the end result using fmt.Sprint (or make them a map and then iterate over key and values)

lib/portlayer/storage/nfs/vm.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+func VolumeJoin(op trace.Operation, handle *exec.Handle, volume *storage.Volume, mountPath string, diskOpts map[string]string) (*exec.Handle, error) {
+	defer trace.End(trace.Begin("nfs.VolumeJoin"))
```
caglar10ur: trace already gives you the name of the function, we should log other stuff here instead of stutter.

lib/portlayer/storage/nfs/vm.go
caglar10ur: I agree with @anchal-agrawal and prefer Sprint but I think this needs to be replaced anyway because of my lack of free form string parameter list :)

lib/portlayer/storage/nfs/vm.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func createMountSpec(host url.URL, mountPath string, diskOpts map[string]string) executor.MountSpec {
+	deviceMode := nfsMountOptions + ",addr=" + host.Host
+	newMountSpec := executor.MountSpec{
+		Source: host,
+		Path:   mountPath,
+		Mode:   deviceMode,
+	}
+	return newMountSpec
```
caglar10ur: Just curios, why we are not returning a pointer to executer.MountSpec but instead returning the whole struct?

lib/portlayer/storage/nfs/volume.go
```diff
@@ -109,6 +110,11 @@ func (v *VolumeStore) VolumeCreate(op trace.Operation, ID string, store *url.URL
 	}
 
 	u, _ := v.Service.URL()
+	if u.Scheme != "nfs" {
```
caglar10ur: Make nfs string a const and use it here

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -491,25 +491,33 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
...
+	case "nfs":
```
caglar10ur: make nfs string a const and use it here

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -491,25 +491,33 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
...
+		actualHandle, err = nfs.VolumeJoin(op, actualHandle, volume, params.JoinArgs.MountPath, params.JoinArgs.Flags)
+	case "ds":
```
caglar10ur: make ds string a const and use it here

lib/portlayer/storage/vsphere/vm.go
```diff
@@ -29,18 +29,26 @@ func VolumeJoin(op trace.Operation, handle *exec.Handle, volume *storage.Volume,
...
-			Path:   volume.Label,
-		},
-		Path: mountPath,
-		Mode: diskOpts["Mode"],
+	//constuct MountSpec for the tether
+	mountSpec := createMountSpec(volume, mountPath, diskOpts)
+
+	//append a device addition spec change to the container config
+	diskDevice := createVolumeVirtualDisk(volume)
+	config := createDeviceConfigSpec(diskDevice)
```
caglar10ur: Just curios, why are these createX functions are first class functions but not methods on those struct they use?

lib/tether/ops_linux.go
```diff
@@ -661,13 +663,33 @@ func (t *BaseOperations) MountLabel(ctx context.Context, label, target string) e
...
+// This assumes that the source url is valid and available.
+func (t *BaseOperations) MountTarget(ctx context.Context, source url.URL, target string, mountOptions string) error {
+	defer trace.End(trace.Begin(fmt.Sprintf("Mounting %s on %s", source.String(), target)))
+
+	if err := os.MkdirAll(target, 0644); err != nil {
+		// same as MountLabel error for consistency
+		return fmt.Errorf("unable to create mount point %s: %s", target, err)
+	}
+
+	rawSource := source.Hostname() + ":/" + source.Path
```
caglar10ur: Prefer fmt.Sprint instead

lib/tether/ops_linux.go
```diff
@@ -661,13 +663,33 @@ func (t *BaseOperations) MountLabel(ctx context.Context, label, target string) e
...
+	if err := Sys.Syscall.Mount(rawSource, target, "nfs", 0, mountOptions); err != nil {
```
caglar10ur: Nfs string to const and while at it could you please make ext4 one above a const too :)

lib/tether/interfaces.go
```diff
@@ -35,6 +36,7 @@ type Operations interface {
 	SetupFirewall() error
 	Apply(endpoint *NetworkEndpoint) error
 	MountLabel(ctx context.Context, label, target string) error
+	MountTarget(ctx context.Context, source url.URL, target string, mountOptions string) error
```
caglar10ur: Maybe we pass a struct here instead of this, which contains values we wanna pass to this function and that struct happens to have some methods on it to convert those values to the format that consumer consumes? (eg; MountOptions struct and String(), Source(), Target() methods))

lib/tether/ops_windows.go
```diff
@@ -50,6 +51,14 @@ func (t *BaseOperations) MountLabel(ctx context.Context, label, target string) e
 	return errors.New("not implemented on windows")
 }
 
+// MountTarget performs a mount based on the target path from the source url
+// This assumes that the source url is valid and available.
+func (t *BaseOperations) MountTarget(ctx context.Context, source url.URL, target string, mountOptions string) error {
+	defer trace.End(trace.Begin(fmt.Sprintf("Mounting %s on %s", source.String(), target)))
+
+	return errors.New("not implemented on OSX")
```
caglar10ur: This is not OSX

lib/portlayer/storage/nfs/vm.go
matthewavery: r/wsize will actually size itself properly if an error is experienced according the `man 5 nfs`. Will stringify this.

matthewavery: I have talked to @fdawg4l about this, and we landed on using `hard` over a retransmission number. We previously blocked indefinitely when a label would not present so I do not see this as any different.

lib/portlayer/storage/vsphere/vm.go
matthewavery: Because these are specific unexported helpers for creating the exact configfuration we need in this instance. They are just meant to tidy up this function and divide out the logical creation steps in a readable manner.

lib/tether/ops_linux.go
matthewavery: I usually avoid using `SprintF` since it is an order of magnitude slower thank concatentation. Though it might be an order of magnitude greater in readability. I may not make this change this time around. The Tether is going to be rewritten at some point right?

lib/portlayer/storage/nfs/vm.go
matthewavery: punting

lib/portlayer/storage/nfs/vm.go
matthewavery: I will comment on the chosen options above in the code so that it is more apparent why they have been chosen as default behavior for now.

fdawg4l: @caglar10ur the mount options are constant for now.  They will all be overrideable going further.

We chose hard as the default because soft has IN ALL CAPS the possibility of dataloss.  And I'm allergic to all things involving dataloss on any level :-).

lib/portlayer/storage/nfs/volume.go
fdawg4l: @matthewavery This really shouldn't be here at all.  See previous review comment about moving that to the Service so this is not needed.

lib/portlayer/storage/nfs/vm.go
matthewavery: changed it to a pointer, a needless copy is no good. Thanks!

lib/tether/interfaces.go
matthewavery: This has been punted to the tether rewrite

lib/tether/ops_windows.go
matthewavery: I will change it to windows hahaha sorry 😄

Issue #4194, opened 2017-03-09
Modify persona container cache to include container.DataVersion


jzt:
To ensure backward compatibility with new features and old containers, we sometimes need to check the version of the container in the persona and return an error to the user when that feature is unsupported.

An example of this currently exists in the `--timestamps` option for `docker logs`. There is currently no way of knowing whether or not a container supports `--timestamps` without making an initial call to the portlayer in order to learn the container's `DataVersion`.

If we were able to include this `DataVersion` in our persona container cache, we could populate it at VCH startup during the initial cache refresh, saving ourselves the need to make that extra portlayer call in the course of normal operation.

This change should probably be considered during the implementation of  https://github.com/vmware/vic/issues/4059.

mhagen-vmware:
Please prioritize and estimate work and move to backlog.

Issue #4059, opened 2017-02-28
Use named constants instead of container version numbers for better backward compatibility


jzt:
Currently our container versioning relies on just a number (we are currently at `PluginVersion == 0`, and that is about to change to `1`.

Defining some constants and using them as conditions would make more sense in the long run, and make more sense to the reader, eg:
```
const (
        TheVersionThatILike = 1 << iota
        TheVersionNobodyLikes
        TheBestVersionEver
        TheBrokenOne
        TheBest
        YuggggeeeVersion
)

CurrentPluginVersion = TheBest
```

and then we can create a function called `Satisfies` which takes an int argument, and call it thusly:
```
if err :=version.Satisfies(TheVersionThatILike | YuggggeeeVersion); err !=nil {
...
} else if err := version.Satisfies(TheBrokenOne); err !=nil  {
...
}
```
which could be implemented like
```
func (v *Version) Satisfies(flags int) error {
	if flags&TheVersionThatILike != 0 && CurrentPluginVersion < TheVersionThatILike {
		return ErrNotSupported
	}
....
	return nil
}
```

caglar10ur:
@emlin what do you think?

emlin:
@caglar10ur 👍 
Just one comment. We might want to change the method name with plugin version mentioned? Cause this is just one item in the version definition.

PR #4209, opened 2017-03-09
Support --timestamps option for docker logs

### LINKED ISSUE #3935: Support for --timestamps option to docker logs

jzt:
This change adds support for the `--timestamps` options to `docker logs`.

There is a backwards-compatibility check that will return `container <container ID> does not support '--timestamps'` when `docker logs --timestamps` is attempted against a container that was created before the VCH was updated to a version on or after this commit. Unfortunately, this check requires an extra call to the portlayer to get the container version. This is required as the `docker logs` endpoint cannot return an error without the docker client misinterpreting it. That prevents us from being able to make a single portlayer call and return a specific error to the client when we realize the container version is too low. This is caused by a limitation in swagger, but will not be necessary once we are including the container version in the persona container cache that is hydrated at VCH startup (see issue #4194), as we can simply check locally in the persona.

Fixes #3935

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -1747,7 +1747,7 @@ func (c *Container) validateContainerLogsConfig(vc *viccontainer.VicContainer, c
 	}
 
 	unsupported := func(opt string) (int64, int64, error) {
-		return 0, 0, fmt.Errorf("%s does not yet support '--%s'", ProductName(), opt)
+		return 0, 0, fmt.Errorf("container %s does not support '--%s'", vc.ContainerID, opt)
 	}
 
 	tailLines := int64(-1)
@@ -1768,8 +1768,20 @@ func (c *Container) validateContainerLogsConfig(vc *viccontainer.VicContainer, c
 		since = time.Unix(s, n)
 	}
 
+	// TODO(jzt): this should not require an extra call to the portlayer. We should
+	// update container.DataVersion when we hydrate the container cache at VCH startup
+	// see https://github.com/vmware/vic/issues/4194
 	if config.Timestamps {
-		return unsupported("timestamps")
+		// check container DataVersion to make sure it's supported
+		params := containers.NewGetContainerInfoParams()
+		params.SetID(vc.ContainerID)
+		info, err := PortLayerClient().Containers.GetContainerInfo(params)
+		if err != nil {
+			return 0, 0, err
+		}
+		if info.Payload.DataVersion == 0 {
+			return unsupported("timestamps")
+		}
 	}
 
 	if config.Since != "" {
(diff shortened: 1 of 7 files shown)


tests/test-cases/Group1-Docker-Commands/1-08-Docker-Logs.robot
```diff
@@ -27,10 +27,63 @@ Grep Logs And Count Lines
...
 
+Install VIC with version to Test Server
+  [Arguments]  ${version}=7315
+  Log To Console  \nDownloading vic ${version} from bintray...
+  ${rc}  ${output}=  Run And Return Rc And Output  wget https://bintray.com/vmware/vic-repo/download_file?file_path=vic_${version}.tar.gz -O vic.tar.gz
+  ${rc}  ${output}=  Run And Return Rc And Output  tar zxvf vic.tar.gz
+  Set Environment Variable  TEST_TIMEOUT  20m0s
+  Install VIC Appliance To Test Server  vic-machine=./vic/vic-machine-linux  appliance-iso=./vic/appliance.iso  bootstrap-iso=./vic/bootstrap.iso  certs=${false}
+  Set Environment Variable  VIC-ADMIN  %{VCH-IP}:2378
+  Set Environment Variable  INITIAL-VERSION  ${version}
```
mhagen-vmware: Both of these are set in Install VIC Appliance To Test Server already.

tests/test-cases/Group1-Docker-Commands/1-08-Docker-Logs.robot
```diff
@@ -27,10 +27,63 @@ Grep Logs And Count Lines
...
+
+Clean up VIC Appliance And Local Binary
+  Cleanup VIC Appliance On Test Server
+  Run  rm -rf vic.tar.gz vic
```
mhagen-vmware: To simplify this whole thing, we can just move this rm -rf to the end of the install and eliminate the need for this wrapper keyword.

jzt: Ok but I stole this from another test, not saying which one. :)

lib/apiservers/engine/backends/container.go
```diff
@@ -1768,8 +1768,20 @@ func (c *Container) validateContainerLogsConfig(vc *viccontainer.VicContainer, c
...
 	if config.Timestamps {
-		return unsupported("timestamps")
+		// check container DataVersion to make sure it's supported
+		params := containers.NewGetContainerInfoParams()
+		params.SetID(vc.ContainerID)
+		info, err := PortLayerClient().Containers.GetContainerInfo(params)
+		if err != nil {
+			return 0, 0, err
+		}
+		if info.Payload.DataVersion == 0 {
```
hickeng: suggest we create `lib/apiserver/engine/backends/feature-support.go` or similar and add some constant mappings:
```go
// Feature to min/max data version. -1 means unlimited
const (
LOG_SINCE_MIN = 1
LOG_SINCE_MAX = -1
)
```

or using functions 
```go
func logSinceSupported(data interface{}) bool {
...
}
```

This will prevent us having the checks scattered over the code and would provide a single reference location for the conditionals that apply to support.

jzt: @hickeng I captured something similar in https://github.com/vmware/vic/issues/4059, can you take a look and provide input/feedback?

Issue #4109, opened 2017-03-01
Add trigger for tether reload


hickeng:
**Story**
As a VIC developer I want to be able to cause tether to reload its configuration

**Detail**
Add a toolbox message to cause a reload of the configuration and add a portlayer method for sending toolbox reload message.
This is very similar to the Signal path.

This may require completion of the tether reorg work to provide a tether reference to toolbox for calling reload.

caglar10ur:
Estimating as 2 - without bringing in tether reorg, otherwise this is multi sprint effort

mhagen-vmware:
i think the fixes statement in you PR needs to be on 2 separate lines to auto close

PR #4210, opened 2017-03-10
Add tether and port layer support for exec

hickeng:
This updates tether to check for whether a given SessionConfig should be active,
and to allow for two tiers of session - persistent & non-persistent which maps to
those created with the container and those exec'd while the container was live.
Additionally merges in @fdawg4l reload trigger as that was the easiest way to do
integration.

It also provides portlayer side package (task) that provides the Join, Bind, Unbind,
and Remove mechanics. If a task is Joined to a container while it's powered on the
resulting session is non-persistent.

This pulls in personality side changes for basic `docker exec` create/start support
from @caglar10ur but does not support the following exec mechanisms:
1. inspect
2. attach
3. start does not block until we've successful report of process launch
4. expanded tests

Fixes: #4110, #4109


caglar10ur:
@hickeng any reason for not to merge this? Looks like tests are passing now as well

cmd/tether/main_linux.go
```diff
@@ -33,22 +33,17 @@ import (
...
 
-	go func() {
-		for _ = range sigs {
-			log.Infof("Got A HUP signal! Reloading tether...")
-			tthr.Reload()
-		}
-	}()
+	// init and start the signal handler
+	if err := startSignalHandler(); err != nil {
+		log.Error(err)
```
fdawg4l: If you can't install the signal handle, should we `Panic()` here?

fdawg4l: Ah.  This func always returns nil.   Nevermind :-).

lib/apiservers/engine/backends/cache/container_cache.go
```diff
@@ -40,6 +42,8 @@ func init() {
 		idIndex:          truncindex.NewTruncIndex([]string{}),
 		containersByID:   make(map[string]*container.VicContainer),
 		containersByName: make(map[string]*container.VicContainer),
+
```
jzt: ✂️  whitespace ✂️

cmd/tether/main_linux.go
```diff
@@ -33,22 +33,17 @@ import (
...
 	log.SetLevel(log.DebugLevel)
 
-	go func() {
-		for _ = range sigs {
-			log.Infof("Got A HUP signal! Reloading tether...")
-			tthr.Reload()
-		}
-	}()
+	// init and start the signal handler
+	if err := startSignalHandler(); err != nil {
```
vburenin: why would you even want to check for error if function always return nil? I would just update that function to return nothing, so less lines of code.

lib/portlayer/task/common.go
```diff
@@ -0,0 +1,64 @@
...
+	task := taskS
+	if handle.Runtime != nil && handle.Runtime.PowerState != types.VirtualMachinePowerStatePoweredOff {
+		op.Debugf("Task bind configuration applies to ephemeral set")
+		task = taskE
+
+		// TODO: add check for container version - if the tether doesn't support reload/exec then
+		// this should fail
+	}
+
+	// if no task has been bound to the
```
jzt: This comment seems incomplete

lib/portlayer/task/remove.go
```diff
@@ -0,0 +1,62 @@
...
+	tasks := stasks
+	if handle.Runtime != nil && handle.Runtime.PowerState != types.VirtualMachinePowerStatePoweredOff {
+		op.Debugf("Task configuration applies to ephemeral set")
+		tasks = etasks
+
+		// TODO: add check for container version - if the tether doesn't support reload/exec then
+		// this should fail
+	}
+
+	// if no task has been bound to the
```
jzt: Incomplete comment

caglar10ur:
Where is the @sgairo - the copyright bot :) ?

sgairo:
@caglar10ur copyright bot glitched, but is back online now

emlin:
Lgtm

emlin:
lgtm

PR #4217, opened 2017-03-10
Fixes to support concurrent VCH and container VM state changes [full ci]

### LINKED ISSUE #3137: Shutdown and (re)Start of VCH has an improper behaviour
Version 0.7.0

Freshly deployed VCH.

```
user1@ubuntu:~/vic$ docker run -d -p 81:80 --name nginx1 nginx 
bca2d70d6d85a656d81ac1a5c4bf66697c536bee613b4651422237a3d9dfe85a

user1@ubuntu:~/vic$ docker run -d -p 82:80 --name nginx2 nginx 
ef75aa8277eb110f3bec52407345d2154eb3dac35b129079a17164fea279ee0b

user1@ubuntu:~/vic$ docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED              STATUS              PORTS                     NAMES
ef75aa8277eb        nginx               "nginx -g daemon off;"   About a minute ago   Up 36 seconds       192.168.8.57:82->80/tcp   nginx2
bca2d70d6d85        nginx               "nginx -g daemon off;"   2 minutes ago        Up About a minute   192.168.8.57:81->80/tcp   nginx1
```

At this point we *SHUTDOWN THE VCH* 

We then *START THE VCH* back up 

On the vSphere side both containerVMs have re-started (along with the VCH EndPoint Docker VM).  

But the docker client doesn’t seem to be able to cope with what’s going on:
```
user1@ubuntu:~/vic$ docker ps -a
CONTAINER ID        IMAGE               COMMAND                  CREATED             STATUS              PORTS                     NAMES
ef75aa8277eb        nginx               "nginx -g daemon off;"   8 minutes ago       Stopped             192.168.8.57:82->80/tcp   nginx2
bca2d70d6d85        nginx               "nginx -g daemon off;"   8 minutes ago       Stopped             192.168.8.57:81->80/tcp   nginx1
```

I believe there are two things here:

https://github.com/vmware/vic/issues/701 should address this inconsistency between the Docker view and the vSphere view.

https://github.com/vmware/vic/issues/744 should implement the proper logic to be applied when the VCH is shut-down and then restarted. Ultimately (IMO) the VCH should behave pretty much like a Docker host that is shutdown and then restarted. 

/cc @emlin

hmahmood:
- Added subscriber implementation that can be "paused", i.e. not
  process events until "unpaused"
- Event manager no longer launches a go routine to process each
  event that is received; instead thee event is not queued to
  preserve the order the events are recieved
- Modified high availability test to include stopped and running
  containers at time HA is triggered
 
Fixes #3137

### DIFF
--- a/lib/portlayer/event/subscriber.go
+++ b/lib/portlayer/event/subscriber.go
@@ -1,0 +1,208 @@
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package event
+
+import (
+	"sync/atomic"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/vmware/vic/lib/portlayer/event/events"
+)
+
+const (
+	suspendDisabled int32 = iota
+	suspendDiscard
+	suspendQueue
+)
+
+type Subscriber interface {
+	// Topic returns the topic this subscriber is subscribed to
+	Topic() string
+	// Name returns the name of the subscriber
+	Name() string
+
+	// Suspend suspends processing events by the subscriber. If
+	// queueEvents is true, the events are queued until Resume()
+	// is called. If queueEvents is false, events passed into
+	// onEvent() after this call are discarded.
+	Suspend(queueEvents bool)
+	// Resume resumes processing of events by the subscriber.
+	// If Suspend() was called with queueEvents as true, any events
+	// that were passed to onEvent() after Suspend() returned are
+	// processed first.
+	Resume()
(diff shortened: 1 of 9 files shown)


lib/portlayer/exec/container.go
```diff
@@ -555,6 +551,96 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
...
+	// do we have a state change
+	if newState != c.state {
+		switch newState {
+		case StateStopping,
+			StateRunning,
+			StateStopped,
+			StateSuspended:
+
+			// container state has changed so we need to update the container attributes
+			// we'll do this in a go routine to avoid blocking
```
cgtexmex: should remove reference to use of a go routine

cgtexmex:
small comment change requested, but LGTM

lib/portlayer/event/subscriber.go
```diff
@@ -0,0 +1,208 @@
...
+			}
+
+			// not suspended
+			select {
+			case e := <-s.eventQ:
+				s.callback(e)
+			case c := <-s.suspend:
+				suspended = c.suspend
+				done = c.done
+			}
```
caglar10ur: I think this non suspended case needs to go into the select statement above. Otherwise - if I'm reading it correctly - it can be paused just after the suspended check and before the 2nd select statement and blocks there forever. No?

hmahmood: If suspended is false, that indicates Resume() was called before. If we end up on line 110, and Suspend() is called, line 113 should be hit, and we should set suspended to true, and loop back around.

PR #4238, opened 2017-03-13
Implement tether reconfigure toolbox to tether cmd

### LINKED ISSUE #4109: Add trigger for tether reload

fdawg4l:
Implements the `reload` guest program handler to reload the tether config (via signal at the moment).

Fixes #4109

### DIFF
--- a/cmd/tether/main_linux.go
+++ b/cmd/tether/main_linux.go
@@ -33,9 +33,6 @@ import (
 var tthr tether.Tether
 
 func init() {
-	sigs := make(chan os.Signal, 1)
-	signal.Notify(sigs, syscall.SIGHUP)
-
 	// Initiliaze logger with default TextFormatter
 	log.SetFormatter(viclog.NewTextFormatter())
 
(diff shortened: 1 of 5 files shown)


sgairo:
Copyrights
-Copyright bot

fdawg4l:
@sgairo youre the best bot of all the bots.

lib/portlayer/exec/container.go
```diff
@@ -405,6 +405,16 @@ func (c *Container) Signal(ctx context.Context, num int64) error {
 	return c.startGuestProgram(ctx, "kill", fmt.Sprintf("%d", num))
 }
 
+func (c *Container) ReloadConfig(ctx context.Context) error {
```
hickeng: should be in `base.go` as we may need to invoke it during commit

cmd/tether/main_linux.go
```diff
@@ -156,3 +151,22 @@ func halt() {
 	syscall.Sync()
 	syscall.Reboot(syscall.LINUX_REBOOT_CMD_POWER_OFF)
 }
+
+func startSignalHandler() error {
```
hickeng: This this should be in `lib/tether/tether_linx.go` - given it's the `lib/tether` code that uses the config it makes sense to have the reload trigger there.

fdawg4l:
handing this off to @hickeng

Issue #3913, opened 2017-02-13
Investigate OVF properties required to support OVA installer use case


frapposelli:
**User Statement:** As developer of the OVA, I need full access to the OVF properties provided by the underlying vSphere infra.

**Acceptance Criteria:** 
- Must be able to read pertinent `guestinfo.*` properties from inside the GOS used in the OVA.
- Must support Phone Home (VAC) with granular properties about what is currently installed in the appliance

mdubya66:
If this doesn't capture setting the guest_full_name for phone home (VAC) purposes we should add one that does. See https://github.com/vmware/vic/issues/3645

frapposelli:
VAC support will require a research spike in the next sprint

frapposelli:
Tracking VAC requirement in #4250

frapposelli:
Rest of the story completed with #4240

PR #4240, opened 2017-03-14
Add DCUI and automated disk/network/password setup to OVA (#3913)

frapposelli:


installer/ovatools/ovfenv/main.go
```diff
@@ -44,6 +44,10 @@ func main() {
 			Value: "",
 			Usage: "Get single OVF property",
 		},
+		cli.BoolFlag{
+			Name:  "dump",
+			Usage: "Dump the OVF Environment XML",
```
vburenin: dump and exit?

frapposelli: yep, dumps the XML and quits, useful for debugging when the go parser fails to handle proper xml 😄

installer/ovatools/vic-ova-ui/main.go
```diff
@@ -0,0 +1,222 @@
...
+)
+
+const (
+	VT_ACTIVATE   = 0x5606
+	VT_WAITACTIVE = 0x5607
+)
+
+func main() {
+
+	if runtime.GOOS == "linux" {
```
dougm: There's other linux specific stuff in here, maybe just skip the check?  Or if we want it to run on more than linux, could move the platform specifics to main_linux.go, main_darwin.go, etc.

installer/ovatools/vic-ova-ui/main.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	defer ui.Close()
+
+	gray := ui.ColorRGB(1, 1, 1)
+	blue := ui.ColorCyan
+
+	// Check if we're running inside a VM
+	isVM, err := vmcheck.IsVirtualWorld()
+	if err != nil {
+		fmt.Printf("error: %s\n", err.Error())
```
dougm: errors to os.Stderr instead of Stdout?

installer/ovatools/vic-ova-ui/main.go
```diff
@@ -0,0 +1,222 @@
...
+		Value string `xml:"value,attr"`
+	}
+
+	type propertySection struct {
+		Property []property `xml:"Property"`
+	}
+
+	var environment struct {
+		XMLName         xml.Name        `xml:"Environment"`
+		PropertySection propertySection `xml:"PropertySection"`
```
dougm: There is a an ovf/env package in govmomi, maybe we could use/extend that? https://github.com/vmware/govmomi/blob/master/ovf/env.go#L51

installer/ovatools/vic-ova-ui/main.go
frapposelli: I thought about that, reality is that the only use case for this is to run on a Linux os, the check is just to help when doing interactive tests on a non-linux platform.

installer/ovatools/vic-ova-ui/main.go
frapposelli: Good call 😄

installer/ovatools/vic-ova-ui/main.go
frapposelli: Interesting, I'll definitely add a story to extend that and use it here.

installer/ovatools/vic-ova-ui/main.go
frapposelli: Actually, I'll make it linux, it's going to be easier to maintain going forward anyway

PR #4246, opened 2017-03-14
[specific ci=Group6-VIC-Machine] Nfs vic machine work

### LINKED ISSUE #3810: Update Vic-Machine Volume Store flag to handle NFS volume stores
Vic-Machine will now need to distinguish between vsphere datastores and NFS servers as volume store targets. The new nfs targets will need a validation strategy as well as commented [here](https://github.com/vmware/vic/blob/master/doc/design/volumes/nfsvolumestores.md) to make sure that at VCH install time we can reach the intended target. 

This should not involve any major changes to the --volume-store flag. It should however require a change to how we parse the flag as well as a new distinction field for volume stores indicating the type that we need to be creating. Most of this work should occur in the `Storage Validator` and the `processVolumeStores` function located [here](https://github.com/vmware/vic/blob/84bf428e80da5e6bb61c82834fdc235240ac4d17/cmd/vic-machine/create/create.go#L867-L879). There will also need to be changes to the addVolumeStore functions which occur after validation. Some of this work will likely be contained in parts of the managment portion(dispatcher section) of vic-machine([here](https://github.com/vmware/vic/blob/84bf428e80da5e6bb61c82834fdc235240ac4d17/lib/install/management/create.go#L71-L73)) as well as a new pkg/nfs may need to be written for adding volume store directories to the nfs target before the VCH is installed similar to [this](https://github.com/vmware/vic/blob/84bf428e80da5e6bb61c82834fdc235240ac4d17/pkg/vsphere/datastore/datastore.go) vsphere pkg library.

matthewavery:
Fixes #3810 

This adds a number of checks and filters to vic-machine in order to support `nfs volumestores` and preserve normal functionality. these are listed below:

- **Added checks during the initial volume store paramter parsing to check if a url is present. The main check for datastores is present as well. Now we verify that atleast one of these parameter checks must pass to accept the input.**

- **Modified volume store validation steps to do additional checks against the URL if found in the `VolumeLocations` list that is populated after string parameter checking. These checks in the validator make sure that the scheme of the url is `nfs` and that Host and Path are present in the URL. it is possible that a customer may want the root of their nfs-server to be a share point. at the moment this is not accounted for and we might want to.(I welcome comments on this)**

- **When attempting to check if all datastores are writable we filter out the nfs volumestores from the aggregation function. This is important to avoid treating nfs stores as datastores. The check is of the type `u.Scheme != dsScheme` meaning that if more types of schemes are accepted this code should not need to be touched again**

- **During a delete the same `u.Scheme != dsScheme` is used to avoid attempting deletes on anything not managed by vsphere. This is primarily because we cannot be sure that vic-machine can reach anything that cannot reach.**

- **During the `createVolumeStores` function which we use to create volume stores on vsphere datastores. We now filter out anything that is `u.Scheme != dsScheme` as well to avoid attempting to create non vsphere datastores as volumestores on a datastore**

- **Updated several error messages to include instructions on nfs based volume stores. Also added additional logging to volumestore deletion logic to indicate which volume's are not getting deleted during `--force` when they are not vsphere based volume stores.**

Please double check if I have modified all the appropriate error messages. It is possible I have missed one or two.


cmd/vic-machine/common/utils.go
```diff
@@ -57,8 +59,30 @@ func CheckUnsupportedCharsDatastore(s string) error {
...
+	if len(parts) < 2 {
+		return nil
+	}
+
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return nil
+	}
+
+	// at the very lease, these fields must exist for a good input to exist for an nfs share
```
sgairo: "least" ?

cmd/vic-machine/create/create.go
```diff
@@ -876,14 +876,25 @@ func (c *Create) processVolumeStores() error {
...
+			}
 		}
-		c.VolumeLocations[splitMeta[1]] = splitMeta[0]
+
+		if err := common.CheckUnsupportedChars(arg); err != nil && u == nil {
+			return fmt.Errorf("--volume-store contains unsupported characters for datastore target: %s Allowed characters are alphanumeric, space and symbols - _ ( ) / :", err)
+		}
+
+		lastIndex := len(splitMeta)
+		reconstructedTarget := strings.Join(splitMeta[0:lastIndex-1], ":")
```
hmahmood: Isn't `reconstructedTarget == u.String()`?

lib/install/validate/storage.go
```diff
@@ -58,18 +62,56 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
 	}
 
 	// TODO: add volume locations
```
hmahmood: remove TODO?

lib/install/validate/storage.go
```diff
@@ -58,18 +62,56 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
...
+
 		if dsURL != nil {
 			conf.VolumeLocations[label] = dsURL
 		}
 	}
 }
 
+// NFSHelper is used to construct a potential nfs target. If a valid NFS target cannot be contsructed a nil pointer will be returned.
+func NFSHelper(rawPath string) (*url.URL, error) {
+	// NOTE: issues from this function will need to be noted at the end of the datastore validation attempt. since we have multiple acceptable input formats now.
```
hmahmood: NOTE: is superfluous; the purpose of the comment is a note :-)

lib/install/validate/storage.go
```diff
@@ -58,18 +62,56 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
...
+	// NOTE: USER is checked at portlayer configuration time, since we cannot validate whether one is needed or not during vic-machine create time.
+
+	u, err := url.Parse(rawPath)
```
hmahmood: Isn't this code doing the same thing as `CheckURLValidation`?

cmd/vic-machine/create/create.go
```diff
@@ -876,14 +876,25 @@ func (c *Create) processVolumeStores() error {
 	defer trace.End(trace.Begin(""))
 	c.VolumeLocations = make(map[string]string)
 	for _, arg := range c.volumeStores {
-		if err := common.CheckUnsupportedCharsDatastore(arg); err != nil {
-			return fmt.Errorf("--volume-store contains unsupported characters: %s Allowed characters are alphanumeric, space and symbols - _ ( ) / :", err)
+		splitMeta := strings.Split(arg, ":")
+		if len(splitMeta) < 1 {
```
matthewavery: There are two checks. One is in the validator and one is at parameter checking time. They are subtly different. The issue is that there is the possibility to reject some strings that we might want during this time. So we need a way to not reject those now since they are valid.

cmd/vic-machine/create/create.go
matthewavery: yes I will fix that.

lib/install/validate/storage.go
matthewavery: will do

lib/install/validate/storage.go
matthewavery: But the `NOTE` portion shows up as a different color in most editors which means people are more likely to read them ;-)

cmd/vic-machine/common/utils.go
matthewavery: yeah, apparently I was thinking about my lease signing at gables when I wrote this :-P

cmd/vic-machine/common/utils.go
```diff
@@ -57,8 +59,30 @@ func CheckUnsupportedCharsDatastore(s string) error {
...
 	}
 	return fmt.Errorf("unsupported character %q in %q", s[v[0]:v[1]], s)
 }
+
+// CheckNFSURlValidation is used to validate propernfs arguments to the --volume-store flag and avoid checking inputs for datastore restricted characters if a url is provided
+func CheckURLValidation(input string) *url.URL {
+	parts := strings.Split(input, ":")
+	rawURL := strings.Join(parts[0:len(parts)-1], ":")
+
+	if len(parts) < 2 {
```
matthewavery: yes it should. I will move it.

cmd/vic-machine/common/utils.go
```diff
@@ -57,8 +59,30 @@ func CheckUnsupportedCharsDatastore(s string) error {
...
+	//this is validation step for characters in a datastore URI
 	if v = re.FindIndex(st); v == nil {
 		return nil
 	}
 	return fmt.Errorf("unsupported character %q in %q", s[v[0]:v[1]], s)
 }
+
+// CheckNFSURlValidation is used to validate propernfs arguments to the --volume-store flag and avoid checking inputs for datastore restricted characters if a url is provided
+func CheckURLValidation(input string) *url.URL {
+	parts := strings.Split(input, ":")
```
matthewavery: Yes, it should since this is the initial check on the input. The input for --volume-store is <volume store target>`:`<label>. Because of this parts must always be atleast 2 or more. In which case it should likely be greater than 1, whoops.... :/ but we can assume a lower bound of the parts for sure.

cmd/vic-machine/common/utils.go
```diff
@@ -57,8 +59,30 @@ func CheckUnsupportedCharsDatastore(s string) error {
...
 	var v []int
+	//this is validation step for characters in a datastore URI
 	if v = re.FindIndex(st); v == nil {
 		return nil
 	}
 	return fmt.Errorf("unsupported character %q in %q", s[v[0]:v[1]], s)
 }
+
+// CheckNFSURlValidation is used to validate propernfs arguments to the --volume-store flag and avoid checking inputs for datastore restricted characters if a url is provided
+func CheckURLValidation(input string) *url.URL {
```
matthewavery: Sounds a lot better :-)

cmd/vic-machine/common/utils_test.go
```diff
@@ -120,3 +119,66 @@ func TestCheckUnsupportedCharsDatastore(t *testing.T) {
 		t.Errorf("got %q, expected error for %q", err, test.S)
 	}
 }
+
+func TestCheckURLValidationOutputParity(t *testing.T) {
+	tests := []struct {
+		S     string
+		valid bool
+	}{
+		{"nfs://127.0.0.1:/my/data/here:labels", true},
```
hmahmood: What about the case where there are no labels? i.e. something like `"nfs://127.0.0.1:/my/data/here"`? Would your code handle that case correctly?

cmd/vic-machine/common/utils.go
gigawhitlocks: Don't forget to update that comment too :)

lib/install/validate/storage.go
```diff
@@ -58,18 +62,56 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
...
+	if err != nil {
+		return nil, err
+	}
+
+	// XXX: the below cases indicate a successful url parsing path, and should be accounted for in the error output from the validator
+	// at a minimum we expect a proper scheme, a host, and a path. without these a nfs store is not identifiable and able to be targeted
+	if u.Scheme != nfsScheme {
+		return u, fmt.Errorf("volume store target (%s) needs to have scheme 'nfs' if it is intended as an nfs volume-store. format: <nfs://<user>:<password>@<host>/<share point path>:label", rawPath)
```
fdawg4l: @hickeng and I were talking about this and we have an altenative.

First, `NFS` doesn't have passwords.  The user here is the UID and the password was intended to be the GID ('uid:gid`, ala the args passed to `chown`).  This UID/GID mapping is used to authenticate with the server.  The expectation is the local system has authenticated this user already and passing the uid/gid for him/her is all that the rpc server needs to allow the target to call the mount program.  Normally this is `0:0` when mounting from a host to an nfs server.  We're allowing you (the user) to change this to conform to whatever your environment expects.  We then use this uid/gid on the VCH only (not the container) to create files as this uid/gid (see anonuid in `man 5 nfs`).

So the alternative we have is allowing the user to use `nfs://<host>/path/to/target,comma,separated,mount,options,like,proto=tcp,or,vers=3`.  We can then scan for the `uid=` and `gid=` fields and use them instead of abusing the `User()` and `Password()` sections of the URL struct.

hmahmood: Or use a query string for the fields? That would let you parse the URL with the url lib.

matthewavery:
This has been shelved for `Sprint 5`. I will return to this next sprint potentially.

cmd/vic-machine/common/utils_test.go
matthewavery: If there is no label the check before this function is called will fail the validation. I will add a case for it here, since an error should occur during the url parsing. I will also add the empty string which should get caught during the url field checking.

matthewavery:
For now one proposed solution is to report errors in the guestinfo of the VCH(no need to persist it). Then check for these after the portlayer is up. This might be useable for multiple error paths in the VCH. What do you think? @hickeng

matthewavery:
@emlin has suggested that I change the type of the `VolumeLocation` list to url to avoid duplicated checking and help make the validator process more direct. I will include that work in this PR since it will directly impact how we control the storage validation process for multiple types of stores.

lib/install/management/store_files.go
```diff
@@ -298,12 +305,19 @@ func (d *Dispatcher) deleteVolumeStoreIfForced(conf *config.VirtualContainerHost
 		for label, url := range conf.VolumeLocations {
 			volumeStores.WriteString(fmt.Sprintf("\t%s: %s\n", label, url.Path))
 		}
-		log.Warnf("Since --force was not specified, the following volume stores will not be removed. Use the vSphere UI to delete content you do not wish to keep.\n%s", volumeStores.String())
+		log.Warnf("Since --force was not specified, the following volume stores will not be removed. Use the vSphere UI or supplied nfs targets to delete content you do not wish to keep.\n%s", volumeStores.String())
```
emlin: Is it possible to clearly separate the nfs volume store and others, and log correct message for different type?

lib/install/management/store_files.go
```diff
@@ -265,6 +266,12 @@ func (d *Dispatcher) lsFolder(ds *object.Datastore, dsPath string) (*types.HostD
 func (d *Dispatcher) createVolumeStores(conf *config.VirtualContainerHostConfigSpec) error {
 	defer trace.End(trace.Begin(""))
 	for _, url := range conf.VolumeLocations {
+
+		// NFS volumestores need only make it into the config of the vch
+		if url.Scheme != dsScheme {
+			continue
```
emlin: We might need log message here, to let user cleanup manually in nfs server

lib/install/management/store_files.go
matthewavery: At this point we most likely do not know which format the user might have been attempting since both checks failed. When I make the change to using the `url` object I will look at this more closely.

lib/install/management/store_files.go
matthewavery: Alright I will look at this as well when I come back to the PR.

cmd/vic-machine/common/utils.go
```diff
@@ -57,8 +59,30 @@ func CheckUnsupportedCharsDatastore(s string) error {
 func checkUnsupportedChars(s string, re *regexp.Regexp) error {
 	st := []byte(s)
 	var v []int
+	//this is validation step for characters in a datastore URI
```
vburenin: add space after "//" :-)

matthewavery: haha thanks :) I have been having randomissues with my `gofmt` not doing this on save for me.

cmd/vic-machine/common/utils.go
```diff
@@ -46,7 +48,7 @@ func LogErrorIfAny(clic *cli.Context, err error) error {
 
 // CheckUnsupportedChars returns an error if string contains special characters
 func CheckUnsupportedChars(s string) error {
-	return checkUnsupportedChars(s, reUnsupp)
+	return checkUnsupportedChars(s, reUnsuppDatastore)
```
emlin: I think these two methods CheckUnsupportedChars and CheckUnsupportedCharsDatastore are checking two different character set, why you need to change this method same to CheckUnsupportedCharsDatastore?

cmd/vic-machine/common/utils.go
emlin: I would suggest to make this more meaningful like CheckVolumeURL, cause this looks more than a general URL validation.

cmd/vic-machine/common/utils.go
```diff
@@ -57,8 +59,30 @@ func CheckUnsupportedCharsDatastore(s string) error {
...
 	return fmt.Errorf("unsupported character %q in %q", s[v[0]:v[1]], s)
 }
+
+// CheckNFSURlValidation is used to validate propernfs arguments to the --volume-store flag and avoid checking inputs for datastore restricted characters if a url is provided
+func CheckURLValidation(input string) *url.URL {
+	parts := strings.Split(input, ":")
+	rawURL := strings.Join(parts[0:len(parts)-1], ":")
+
+	if len(parts) < 2 {
+		return nil
```
emlin: Looks like you'll have error for different scenario, instead of return a nil URL object, is it better to return a detailed error?
That can help user to figure out what's wrong in their input string

cmd/vic-machine/common/utils.go
```diff
@@ -57,8 +59,30 @@ func CheckUnsupportedCharsDatastore(s string) error {
...
+	}
+
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return nil
+	}
+
+	// at the very lease, these fields must exist for a good input to exist for an nfs share
+	if u.Host == "" || u.Scheme == "" || u.Path == "" {
+		return nil
```
emlin: ditto ^^^

cmd/vic-machine/create/create.go
emlin: agree with @hmahmood . suggest to move the logic of strings.Split here into the URL checking method, and return detailed error from that method.
I assume that method is used in parameter checking only.

cmd/vic-machine/create/create.go
```diff
@@ -876,14 +876,25 @@ func (c *Create) processVolumeStores() error {
...
+
+		u := common.CheckURLValidation(arg)
+		if u != nil {
+			if u.Scheme != "nfs" {
+				return fmt.Errorf("parsed url for option --volume-store does not have scheme 'nfs',  valid inputs are datastore/path:label or <nfs://<user>:<password>@<host>/<share point path>:label")
+			}
 		}
-		c.VolumeLocations[splitMeta[1]] = splitMeta[0]
+
+		if err := common.CheckUnsupportedChars(arg); err != nil && u == nil {
```
emlin: I guess this is why you changed the character set in this method. Originally the code is using CheckUnsupportedCharsDatastore 😄

lib/install/management/store_files.go
emlin: you can create two byte buffer, and print nfs volume store to one buffer, and others into another buffer, then print them separately.
feels like accurate error message is better

cmd/vic-machine/create/create.go
```diff
@@ -876,14 +876,25 @@ func (c *Create) processVolumeStores() error {
...
+			return fmt.Errorf("--volume-store contains unsupported characters for datastore target: %s Allowed characters are alphanumeric, space and symbols - _ ( ) / :", err)
+		}
+
+		lastIndex := len(splitMeta)
+		reconstructedTarget := strings.Join(splitMeta[0:lastIndex-1], ":")
+		c.VolumeLocations[splitMeta[lastIndex-1]] = reconstructedTarget
```
emlin: I would suggest to change the VolumeLocations to URL array, instead of string array, cause you already parsed the string here

lib/install/validate/storage.go
```diff
@@ -57,18 +61,56 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
 	}
 
 	// TODO: add volume locations
-	for label, volDSpath := range input.VolumeLocations {
-		dsURL, _, err := v.DatastoreHelper(ctx, volDSpath, label, "--volume-store")
-		v.NoteIssue(err)
+	for label, volPath := range input.VolumeLocations {
+		nfsURL, nfsErr := NFSHelper(volPath)
```
emlin: pass in a url here can help you to determine if this is one nfs volume store easily

lib/install/validate/storage.go
```diff
@@ -57,18 +61,56 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
...
+	// NOTE: USER is checked at portlayer configuration time, since we cannot validate whether one is needed or not during vic-machine create time.
+
+	u, err := url.Parse(rawPath)
+	if err != nil {
+		return nil, err
+	}
+
+	// XXX: the below cases indicate a successful url parsing path, and should be accounted for in the error output from the validator
+	// at a minimum we expect a proper scheme, a host, and a path. without these a nfs store is not identifiable and able to be targeted
+	if u.Scheme != nfsScheme {
```
emlin: I think this test is wrong. A normal volume store configuration will hit here.
The logic should be outside of this method, and only while the scheme is nfs, they should be validated by this method additionally, for the uid/gid scan, etc.

cmd/vic-machine/create/create.go
matthewavery: I believe this has been suggested previously as well. I am perfectly happy to make this change. My one concern is that it will cause significant code churn. This Map is used in around 15 places between vic-machine and the portlayer. So all of those places will need changes as well.

cmd/vic-machine/common/utils.go
matthewavery: the input should be `nfs://host/path/,some,options,here` so maybe that check could be a little more robust I am looking at it now.

matthewavery: the input will be `<some url>:<label>` that we are pulling apart. the validation for this part is actually outside of this function so I likely do not need this. I probably should just pass in the raw url to this function. and drop this check altogether.

cmd/vic-machine/common/utils.go
matthewavery: this was likely an accident. I will fix it after taking another look.

matthewavery:
As a note, Things that are still needed: **update all integration tests that do error string checking to look for the newly formatted error messages**

cmd/vic-machine/create/create.go
```diff
@@ -874,16 +877,41 @@ func (c *Create) processDNSServers() error {
...
 func (c *Create) processVolumeStores() error {
 	defer trace.End(trace.Begin(""))
-	c.VolumeLocations = make(map[string]string)
+	c.VolumeLocations = make(map[string]*url.URL)
 	for _, arg := range c.volumeStores {
-		if err := common.CheckUnsupportedCharsDatastore(arg); err != nil {
-			return fmt.Errorf("--volume-store contains unsupported characters: %s Allowed characters are alphanumeric, space and symbols - _ ( ) / :", err)
+		splitMeta := strings.Split(arg, ":")
+		if len(splitMeta) < 1 {
+			return errors.New("volume store input must be in format datastore/path:label or <nfs://<host>/<url-path>?<mount option as query parameters>:,<comma separated mount options>label")
```
emlin: I assume there still have one colon before label, right?
`...<comma separated mount options>label")`

cmd/vic-machine/create/create.go
```diff
@@ -874,16 +877,41 @@ func (c *Create) processDNSServers() error {
 
 func (c *Create) processVolumeStores() error {
 	defer trace.End(trace.Begin(""))
-	c.VolumeLocations = make(map[string]string)
+	c.VolumeLocations = make(map[string]*url.URL)
 	for _, arg := range c.volumeStores {
-		if err := common.CheckUnsupportedCharsDatastore(arg); err != nil {
-			return fmt.Errorf("--volume-store contains unsupported characters: %s Allowed characters are alphanumeric, space and symbols - _ ( ) / :", err)
+		splitMeta := strings.Split(arg, ":")
+		if len(splitMeta) < 1 {
```
emlin: <2 ?
Feels like there should be at least 2 items

cmd/vic-machine/create/create.go
```diff
@@ -874,16 +877,41 @@ func (c *Create) processDNSServers() error {
...
+
+		// divide out the label with the target
+		lastIndex := len(splitMeta)
+		label := splitMeta[lastIndex-1]
+		rawTarget := strings.Join(splitMeta[0:lastIndex-1], ":")
+
+		// raw target input should be in the form of a url
+		urlTarget, err := url.Parse(rawTarget)
+		if err != nil {
+			return fmt.Errorf("parsed url for option --volume-store does not have scheme 'nfs',  valid inputs are datastore/path:label or <nfs://<host>/<share point path>?<mount options as query params>:label")
```
emlin: looks like the valid format error message is used multi places, and you had one with a bit problem (no colon before label^^^), so suggest to define a variable for the common error message

cmd/vic-machine/create/create.go
```diff
@@ -874,16 +877,41 @@ func (c *Create) processDNSServers() error {
...
-		splitMeta := strings.SplitN(arg, ":", 2)
-		if len(splitMeta) != 2 {
-			return errors.New("Volume store input must be in format datastore/path:label")
+
+		switch urlTarget.Scheme {
+		case nfsScheme:
+			// nothing needs to be done here. parsing the url is enough for pre-validation checking.
+		default:
+			// a datastore target is our default assumption
+			urlTarget.Scheme = dsScheme
```
emlin: if I specified sth like http://host:port:label, which is absolute not supported atm, but looks like it will be override by set scheme here

cmd/vic-machine/create/create.go
matthewavery: Yes I missed this one from the old comments during the restructure.

cmd/vic-machine/create/create.go
matthewavery: Yeah I was just thinking about that 😄 I am on it!

cmd/vic-machine/create/create.go
matthewavery: switch case for `"", "ds"` and default should return an error message stating the usage case. error message can be "Please specify a datastore or nfs target. See -vs usage for examples."

cmd/vic-machine/create/create.go
```diff
@@ -1389,6 +1417,8 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 
 	log.Infof("Initialization of appliance successful")
 
+	// We must check for the volume stores that are present after the portlayer presents.
```
emlin: wondering if it will be better to move the comments into CheckDockerAPI? I think you extended that method to check docker volume

lib/install/validate/storage.go
```diff
@@ -31,6 +31,11 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/datastore"
 )
 
+const (
+	nfsScheme = "nfs"
```
emlin: this constant is defined several times in vic-machine, you might move that into cmd/vic-machine/common, or rely on portlayer constant directly.
maybe the latter is better, cause vic-machine will rely on portlayer anyway

lib/install/validate/storage.go
```diff
@@ -56,30 +61,60 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
...
+func validateNFSTarget(nfsURL *url.URL) error {
+	if nfsURL.Host == "" || nfsURL.Path == "" {
+		return fmt.Errorf("nfs volume target (%s) requires both a host and a share point path", nfsURL.String())
+	}
+
+	if nfsURL.Host == "" {
+		return fmt.Errorf("volume store target (%s) is missing the host field. format: <nfs://<user>:<password>@<host>/<share point path>:label", nfsURL.String())
+	}
+
+	if nfsURL.Path == "" {
```
emlin: is the line 90 duplicated with 94 and 98?

lib/install/data/data.go
```diff
@@ -42,7 +42,7 @@ type Data struct {
 	common.Images
 
 	ImageDatastorePath     string
-	VolumeLocations        map[string]string
+	VolumeLocations        map[string]*url.URL
```
matthewavery: yeah 😄 this is what constituted a fair amount of change, but it was desperately needed!

cmd/vic-machine/create/create.go
matthewavery: ah yes that was supposed to have a TODO in it so that I could search for it... thanks :D

lib/install/validate/storage.go
matthewavery: yes I originally did this to avoid pulling in a lot of packages. I will look at the common package

lib/install/validate/storage.go
matthewavery: ah 90 was meant to be replaced by 94 and 98.

cmd/vic-machine/create/create.go
```diff
@@ -186,7 +189,7 @@ func (c *Create) Flags() []cli.Flag {
 		cli.StringSliceFlag{
 			Name:  "volume-store, vs",
 			Value: &c.volumeStores,
-			Usage: "Specify a list of location and label for volume store, e.g. \"datastore/path:label\" or \"datastore:label\".",
+			Usage: "Specify a list of location and label for volume store, nfs stores can have mount options specified as query parameters in the url target. \n\t examples for a vsphere backed volume store are:  \"datastore/path:label\" or \"datastore:label\" or \"ds://my-datastore-name:store-label\"\n\t Examples for nfs back volume stores are: \"nfs://127.0.0.1/path/to/share/point?uid=1234&gid=5678&proto=tcp:my-volume-store-label\" or \"nfs://my-store/path/to/share/point:my-label\"",
```
emlin: hmm.. not quite sure for the impact of \n\t in the usage message. Could you please post the output of vic-machine create -x with two different screen width? I just want to make sure the output is not weird. 😄

matthewavery: yea sure :)

cmd/vic-machine/create/create.go
```diff
@@ -874,16 +876,43 @@ func (c *Create) processDNSServers() error {
...
 func (c *Create) processVolumeStores() error {
 	defer trace.End(trace.Begin(""))
-	c.VolumeLocations = make(map[string]string)
+	c.VolumeLocations = make(map[string]*url.URL)
 	for _, arg := range c.volumeStores {
-		if err := common.CheckUnsupportedCharsDatastore(arg); err != nil {
-			return fmt.Errorf("--volume-store contains unsupported characters: %s Allowed characters are alphanumeric, space and symbols - _ ( ) / :", err)
+		splitMeta := strings.Split(arg, ":")
+		if len(splitMeta) < 2 {
+			return errors.New("volume store input must be in format datastore/path:label or ")
```
hmahmood: `or "`?

cmd/vic-machine/create/create.go
```diff
@@ -874,16 +876,43 @@ func (c *Create) processDNSServers() error {
...
+
+		// divide out the label with the target
+		lastIndex := len(splitMeta)
+		label := splitMeta[lastIndex-1]
+		rawTarget := strings.Join(splitMeta[0:lastIndex-1], ":")
+
+		// raw target input should be in the form of a url
+		urlTarget, err := url.Parse(rawTarget)
+		if err != nil {
+			return fmt.Errorf("parsed url for option --volume-store does not have scheme 'nfs',  valid inputs are datastore/path:label or %s", nfsInputformat)
```
hmahmood: Two spaces instead of one space after comma.

emlin:
nice job. lgtm

cmd/vic-machine/create/create.go
hmahmood: Also seems like error is nfs specific, when in fact you may be parsing a `ds://` url here as well.

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -75,10 +75,8 @@ func (h *StorageHandlersImpl) Configure(api *operations.PortLayerAPI, handlerCtx
 	// expensive metadata lookups.
 	h.imageCache = spl.NewLookupCache(ds)
 
-	// add the volume stores
-	if err = h.configureVolumeStores(op, handlerCtx); err != nil {
-		log.Panicf("Cannot instantiate volume stores: %s", err.Error())
-	}
+	// add the volume stores, errors are logged within this function.
+	h.configureVolumeStores(op, handlerCtx)
```
hmahmood: We don't quit if there is an error? Before we used to panic.

cmd/vic-machine/create/create.go
matthewavery: will fix asap

cmd/vic-machine/create/create.go
matthewavery: yeah that needs an upgrade. I should likely say something more along the lines of "invalid url input, ..." stating the proper input.

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
matthewavery: No this was originally the problem, we want to make sure that all volume stores were established at vic-machine create time. in order to do this we must allow the portlayer to establish all possible volume-stores. Vic-machine will then compare the list of returned volume-stores during a `docker info` call and compare that to the desired volume store calls. The portlayer should not be aborted just because an nfs store was not reachable. The datastores are checked at vic-machine time, but as previously discussed we cannot do that for nfs stores.

cmd/vic-machine/create/create.go
matthewavery: Since this could be a ds url or a non scheme url as well. I have updated this comment and I will push it a long with other smalls changes shortly.

cmd/vic-machine/create/create.go
```diff
@@ -183,7 +185,7 @@ func (c *Create) Flags() []cli.Flag {
 		cli.StringSliceFlag{
 			Name:  "volume-store, vs",
 			Value: &c.volumeStores,
-			Usage: "Specify a list of location and label for volume store, e.g. \"datastore/path:label\" or \"datastore:label\".",
+			Usage: "Specify a list of location and label for volume store, nfs stores can have mount options specified as query parameters in the url target. \n\t Examples for a vsphere backed volume store are:  \"datastore/path:label\" or \"datastore:label\" or \"ds://my-datastore-name:store-label\"\n\t Examples for nfs back volume stores are: \"nfs://127.0.0.1/path/to/share/point?uid=1234&gid=5678&proto=tcp:my-volume-store-label\" or \"nfs://my-store/path/to/share/point:my-label\"",
```
fdawg4l: This comment puts the `uid` and `gid` in the ODATA query parameters of this URL, but https://github.com/vmware/vic/pull/4246/files#diff-d17427b3b302b85a6c75b5a5a264bdc3R581 expects it in the `username` and `password` fields.

fdawg4l: Minor, if you're going to use ODATA query parameters AND THEN use `:` to tokenize another parameter, why not just use the query parameter for that too?  I'm not a useability expert, but it's something to consider.  Check with @hickeng to get a sense of which direction to take here.  I'm not a fan of tokenizing URLs.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -59,6 +59,7 @@ func NewVolumeStore(op trace.Operation, storeName string, mount MountServer) (*V
 
 	target, err := mount.Mount(op)
 	if err != nil {
+		op.Infof("error occurred while attempting to mount volumestore (%s). err: (%s)", storeName, err.Error())
```
fdawg4l: I think we have an `op.Errorf`.

cmd/vic-machine/create/create.go
matthewavery: Yeah, I think that would great if we had started that way. Unfortunately, we need to be backwards compatible with previous versions of vic-machine. That is my feeling on it, I will talk to @hickeng and @emlin about this possible change.

lib/portlayer/storage/nfs/volume.go
matthewavery: we do I will modify that.

cmd/vic-machine/create/create.go
matthewavery: I have a ticket out to change where we look for the uid and gid. I can include it in this PR if you would like @fdawg4l . I am trying to make sure we are all happy with the vic-machine error messages first so that i can modify the integration tests to pass.

matthewavery:
The following commits will support including the `nfsv3` kmod, slightly adjusting the nfs mount options, and adding the `noatime` flag onto the mount syscall rather than in the mount options pointer. Much credit goes to @fdawg4l for figuring out some pretty mysterious errors I was seeing.

lib/install/validate/validator_test.go
```diff
@@ -392,12 +457,18 @@ func testStorage(v *Validator, input *data.Data, conf *config.VirtualContainerHo
...
 			conf.ImageStores = conf.ImageStores[1:]
 			for key, volume := range conf.VolumeLocations {
-				assert.Equal(t, test.expectVolumes[key], volume.String())
+				if _, ok := test.expectVolumes[key]; !ok {
+					assert.Fail(t, "Could not find volume store that was expected to present", "volume : %s", volume.String())
+				} else {
+					assert.Equal(t, test.expectVolumes[key].String(), volume.String())
+				}
 			}
+			log.Print("COMPLETED TEST CASE")
```
matthewavery: did not mean to leave this behind. I will remove this after the results of the current CI run.

cmd/vic-machine/create/create_test.go
```diff
@@ -260,3 +260,41 @@ func TestParseGatewaySpec(t *testing.T) {
 		}
 	}
 }
+
+func TestProcessVolumeStoreParam(t *testing.T) {
+	posiitveTestCases := []string{
```
mhagen-vmware: positiveTestCases?  looks like you somehow managed to mispell it the exact same way below as well :)

PR #4248, opened 2017-03-14
Update vendor directory

caglar10ur:
- Removes not-needed packages
- Updates golang.org packages
- Updates gvt and gas


caglar10ur:
(missed doing a git add for a directory which is not fixed)

PR #4270, opened 2017-03-15
Use govmomi DatastorePath for datastore pathing in helper [full ci]

### LINKED ISSUE #4243: Failed to write to image store parentMap.tmp was not found
docker: Failed to write to image store: [POST /storage/{store_name}][500] WriteImage default  &{Code:0xc42095ef08 Message:File /vmfs/volumes/53061925-77305773-73ad-74867ae3d1ae/VCH02/VIC/564d9379-ab62-acac-5931-06df153c5794/parentMap.tmp was not found}.

**VIC version:**

v0.8.0-7315-c8ac999

**Deployment details:**

./vic-machine-linux create --user root -t '10.112.200.227' --name VCH02 --no-tlsverify --no-tls --force

**Steps to reproduce:**
docker -H 10.112.201.244:2375 run -it bash

**Actual behavior:**
bash should run, container should be created

**Expected behavior:**
Error. Container is not getting created

**Logs:**

[vicadmin.txt](https://github.com/vmware/vic/files/841195/vicadmin.txt)
[docker-personality.txt](https://github.com/vmware/vic/files/841198/docker-personality.txt)
[init.txt](https://github.com/vmware/vic/files/841196/init.txt)
[port-layer.txt](https://github.com/vmware/vic/files/841197/port-layer.txt)

fdawg4l:
We were incorrectly parsing the datastore string to host and path by
splitting on spaces.  This is horrifically incorrect.  Use the govmomi
DatastorePath and its parser instead.

Fixes #4243


pkg/vsphere/datastore/datastore.go
```diff
@@ -89,12 +88,13 @@ func NewHelperFromURL(ctx context.Context, s *session.Session, u *url.URL) (*Hel
...
-		RootURL: u.Path,
+		ds: vsDs,
+		s:  s,
+		fm: fm,
 	}
 
+	d.RootURL.FromString(u.Path)
+
 	return d, nil
 }
```
hickeng: add a `New HelperFromDatastorePath(context.Context, *session.Session, object.DatastorePath)`  so that we've got an easy transition to using object.DatastorePath directly in the config instead of URL.
The only reason we were using URL at all was for the structured path data and url.Parse - given we now have a govmomi structured path type and the url parsing is broken for spaces and unicode in the "hostname" portion it adds no value.

pkg/vsphere/datastore/datastore.go
```diff
@@ -250,7 +250,7 @@ func (d *Helper) Mv(ctx context.Context, fromPath, toPath string) error {
...
 
 func (d *Helper) Rm(ctx context.Context, pth string) error {
-	f := path.Join(d.RootURL, pth)
+	f := path.Join(d.RootURL.String(), pth)
```
hickeng: Why not have a DatastorePath.Join method.

pkg/vsphere/datastore/datastore.go
```diff
@@ -319,32 +319,28 @@ func (d *Helper) mkRootDir(ctx context.Context, rootdir string) error {
...
-func (d *Helper) rootDir() string {
-	return strings.SplitN(d.RootURL, " ", 2)[1]
-}
-
-// Parse the datastore format ([datastore1] /path/to/thing) to groups.
-var datastoreFormat = regexp.MustCompile(`^\[([\w\d\(\)-_\.\s]+)\]`)
-var pathFormat = regexp.MustCompile(`\s([\/\w-_\.]*$)`)
-
 // Converts `[datastore] /path` to ds:// URL
 func ToURL(ds string) (*url.URL, error) {
```
hickeng: If we're going to do this we should allow symmetry - so that we can parse back to DatastorePath from the url format.
Two options:
1. add a FromURL field that parses the url, treating the Host part as the datastore - this cannot be with url.Parse because that doesn't handle many characters in the host portion.
2. update FromString so that it can accept an "ipv6" encoded form: `ds://[datastore name with whatever]/path/to/other`

fdawg4l:
Copying here for posterity-  @hickeng and I decided I'm going to revert the regex bits and save unification for another PR.  This change will at least get the user out of the issue with spaces in the datastore name that s/he is in.

pkg/vsphere/datastore/datastore.go
fdawg4l: Good call.  Will update.

fdawg4l: I'm renegging on this.  Changing the constructor from URL to DatastorePath is the right call, but the trickle is too large for this specific bug fix.  I could just make a new constructor from datastore path and leave the from url constructor, but that doesn't do very much if nothing is calling it.  So I'm going to punt on adding this for now and get back to it after this PR.  I just want to fix the issue mentioned in the fix.  The rest is just fixing debt and i don't want to conflate the two and blow out my point calculation for the story.

PR #4286, opened 2017-03-16
Add timeout to vicadmin WAN check

### LINKED ISSUE #4283: VIC Admin takes extremely long time to load when Internet connection is not available
**Details:**
When a connection to the WAN is severed, VIC Admin tries to do an Internet connectivity test by reaching out to a known IP on the Web. This test does not fail until the attempted connection times out, causing the status page to take about 8 minutes to load, as it waits for the test to complete before loading.

**Acceptance Criteria:**
Put a much more aggressive timeout on the Internet connectivity test displayed on the VIC Admin dashboard so that when a WAN connection is not present, the page loads in a reasonable amount of time. I do not personally think this timeout should be greater than 30 seconds. 

**VIC version:**
v0.8

**Steps to reproduce:**
Deploy VCH on a network with ports 80 and 443 to the outside blocked

**Actual behavior:**
Takes 8 minutes to load the dashboard

**Expected behavior:**
Dashboard loads quickly.

gigawhitlocks:
Fixes #4283 by replacing `net.Dial` with `net.DialWithTimeout` with a much shorter timeout.


lib/vicadmin/validate.go
```diff
@@ -180,7 +180,7 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
 	nwErrors := []error{}
 
 	for _, host := range hosts {
-		conn, err := net.Dial("tcp", host)
+		conn, err := net.DialTimeout("tcp", host, time.Minute/2)
```
jzt: Suggestion: make the time a constant?

caglar10ur: and maybe 30  * time.Seconds instead of Minute /2 :)

gigawhitlocks:
Verified via ad-hoc testing -- used iptables to block outgoing traffic on 443/80 -- with patch vicadmin loads after 30 seconds.

lib/vicadmin/validate.go
gigawhitlocks: We discussed this on slack @caglar10ur and figured that Minute / 2 will be optimized to a bit shift so it doesn't matter

gigawhitlocks: Heck, Go might even be smart enough to realize that operation is a constant and do it at compile-time, idk, I don't think it's an important optimization lmao

caglar10ur: Who cares what compiler does for this case? I wasn't trying to micro optimize the generated code but instead trying to make the source more readable - at least for myself 30 * time.Seconds is much more easy to parse than time.Minute /2 

Anyhow, it's merged so no big deal.

PR #4287, opened 2017-03-16
fixed incorrect image digest persistance

### LINKED ISSUE #4282: docker persona not saving image repo digest correctly
The docker persona is not saving the image metadata repo digest correctly.  This results in the `docker images` command providing incorrect output.

Current - 1.11.2 client:
```
docker images
REPOSITORY          TAG                                                                IMAGE ID            CREATED             SIZE
busybox             latest                                                             00f017a8c2a6        7 days ago          1.11 MB
sha256              32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f   00f017a8c2a6        7 days ago          1.11 MB
```
Expected - 1.11.2 client:
```
docker images
busybox             latest              00f017a8c2a6        7 days ago          1.11 MB
busybox             <none>              00f017a8c2a6        7 days ago          1.11 MB
```

cgtexmex:
The image digest was being persisted as part of image metadata
and not in the repository cache.  That has been corrected.

Additionally, the docker 1.11.2 client presents the docker images command
differently than 1.13, so tests were updated to satisfy our current
testing client of 1.11.2.

Fixes #4282

### DIFF
--- a/lib/apiservers/engine/backends/filter/image.go
+++ b/lib/apiservers/engine/backends/filter/image.go
@@ -101,6 +101,11 @@ func IncludeImage(imgFilters filters.Args, listContext *ImageListContext) Filter
 		refs := cache.RepositoryCache().References(listContext.ID)
 		// reference filters
 		refFilters := imgFilters.Get("reference")
+
+		// reset the tags / digests
+		listContext.Tags = nil
+		listContext.Digests = nil
+
 		// iterate of reporsitory references and filters
 		for _, ref := range refs {
 			for _, rf := range refFilters {
@@ -108,7 +113,6 @@ func IncludeImage(imgFilters filters.Args, listContext *ImageListContext) Filter
 				matchRef, _ := path.Match(rf, ref.String())
 				// match on repo only ie. busybox
 				matchName, _ := path.Match(rf, ref.Name())
-
 				// if either matched then add to tag / digest
 				if matchRef || matchName {
 					if _, ok := ref.(reference.Canonical); ok {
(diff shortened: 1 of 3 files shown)


lib/apiservers/engine/backends/filter/image.go
```diff
@@ -101,14 +101,18 @@ func IncludeImage(imgFilters filters.Args, listContext *ImageListContext) Filter
 		refs := cache.RepositoryCache().References(listContext.ID)
 		// reference filters
 		refFilters := imgFilters.Get("reference")
+
+		// current filter matches
+		listContext.Tags = make([]string, 0)
```
vburenin: not sure if it makes sense to even bother, but if there is a chance that these arrays will hold at least 1 element, it is better to add some capacity into these array. Let say for 4 elements. However, if there are no elements expected, its better to just keep nil, since nil is treated as a perfect empty array.

lib/imagec/imagec.go
```diff
@@ -289,13 +289,23 @@ func updateRepositoryCache(ic *ImageC) error {
...
-	// to the portlayer k/v
+	// AddReference will add the repo:tag to the repositoryCache and save to the portLayer
 	err = repoCache.AddReference(ref, ic.ImageID, true, imageLayerID, true)
 	if err != nil {
 		return fmt.Errorf("Unable to Add Image Reference(%s): %s", ref.String(), err.Error())
 	}
 
+	dig, err := reference.ParseNamed(fmt.Sprintf("%s@%s", ref.Name(), ic.ManifestDigest))
+	if err != nil {
+		return fmt.Errorf("Unable to parse digest: %s", err.Error())
```
vburenin: Why do you add your own error here? Would not it be better just return original one?

PR #4288, opened 2017-03-16
Inspect support for exec'ed session

### LINKED ISSUE #4264: Add inspect support to exec'ed session
Story
As a VIC developer I want to be able to inspect a container with exec'ed sessions 

Detail
We now have exec cmd's running in detached mode, we need to add support of inspecting them

Acceptance
docker inspect ID needs to show ExecIDs of exec'ed session, personalities exec backend needs to return identical data to https://docs.docker.com/engine/api/v1.25/#operation/ExecInspect

caglar10ur:
Fixes #4264 

Includes commits from #4210 for now, will rebase once that gets in (hence WIP).


pkg/vsphere/extraconfig/basic_test.go
```diff
@@ -357,13 +357,16 @@ func TestInheritenceOfNonPersistence(t *testing.T) {
...
+
+		Notes string `vic:"0.1" scope:"hidden" key:"notes"`
 	}
 
 	Struct := Type{
 		Common: CommonPersistence{
 			ID:   "0xDEADBEEF",
 			Name: "Struct",
 		},
+		Notes: "notes",
```
hickeng: Had deliberately left this out to ensure that empty fields didn't get serialized.

hickeng:
We also need to correct the configuration of `runblock` so it's not possible if attach is disabled in `lib/portlayer/attach/attach.go:toggle`

pkg/vsphere/extraconfig/basic_test.go
caglar10ur: Will drop this bit

caglar10ur:
I'm planning to fix the runblock in attach PR along with other bits and pieces

sgairo:
copyright ranges :)  -copyright bot

caglar10ur:
ping @cgtexmex @jzt @sflxn

lib/apiservers/engine/backends/container/viccontainer.go
```diff
@@ -30,32 +29,24 @@ type VicContainer struct {
 	HostConfig  *containertypes.HostConfig
 
 	m     sync.RWMutex
-	execs map[string]*types.ExecConfig
+	execs map[string]struct{}
```
jzt: Probably a dumb question, but where do we keep the `ExecConfig` objects cached now that we aren't storing them here?

jzt: Assuming we need to store them, that is. :)

caglar10ur: :) We don't keep ExecConfig because we don't need them. When we create we persist that data on the portlayer side and later when we need them we simply call inspect

cgtexmex:
lgtm

Issue #3759, opened 2017-01-30
Document location and contents of VIC Summary tab in vSphere H5 client


stuclem:
Per https://github.com/vmware/vic/issues/3749.

stuclem:
Summary tab is implemented in PR https://github.com/vmware/vic/pull/4006.

PR #4290, opened 2017-03-17
Added info for H5 client upgrade and usage.

stuclem:
Fixes https://github.com/vmware/vic/issues/3759.

Also removed all uses of "engine" when referring to the vSphere Integrated Containers plug-in and fixed some link text to reflect a new topic title.

@jooskim can you please take a look? Thanks!


doc/user_doc/vic_admin/upgrade_ui_plugin.md
```diff
@@ -1,19 +1,19 @@
-# Upgrade the vSphere Integrated Containers Engine Plug-In #
+# Upgrade the Flex-Based vSphere Web Client Plug-In #
```
jooskim: This is just a note, but we don't bring additional features to the Flex plugin, so we probably want to either have a message regarding that, or drop this section. What's your thought?

stuclem: Good point. I vote to drop the section and say that no new work will be done.

Issue #4101, opened 2017-03-01
Virtual Serial Port Concentrator integration with Port Layer


corrieb:
**User Statement:**

As a VIC admin, I want to be able to put hosts into maintenance mode and have container workloads vMotioned off that host, regardless of the attach status of the container.

**Details:**

This issue captures integration of the stand-alone implementation #4100 of the vSPC into the Port Layer.

The integration points are as follows:

- vSPC will run as a separate binary in the VCH endpoint VM
- The portlayer AttachServer will listen to localhost:2379 instead of <vch_ip>:2377
- vSPC will listen to <vch_ip>:2377 and relay data from/to the serial port to the AttachServer
- The code will be modified such that vSPC starts running after the appliance start up

**Acceptance Criteria:**

Demonstrate vMotion with a connected serial port without using a standalone vSPC.

PR #4300, opened 2017-03-18
vMotion support for attached containerVMs via vSPC

### LINKED ISSUE #4100: Stand-alone implementation of Virtual Serial Port Concentrator

kreamyx:
closes #4100 and #4101 

These are the major changes in this PR:
- The port-layer attach server is listening to `localhost:2379` instead of `<vch-ip>: 2377`
- The vSPC is taking on the-attach server responsibility of listening to incoming connections at `<vch-ip>:2377`
- The vSPC acts as a relay between the attach server and the containerVM serial port
- VMs are no longer powered off in the event of the host entering maintenance mode

To Do:
- write tests


cmd/vspc/handlers.go
```diff
@@ -0,0 +1,191 @@
...
+import (
+	"bytes"
+	"crypto/rand"
+	"fmt"
+	"io"
+	"net"
+
+	log "github.com/Sirupsen/logrus"
+
+	"strings"
```
anchal-agrawal: gofmt: pkg `strings` should go in the top-level group.

cmd/vspc/vspc.go
```diff
@@ -0,0 +1,200 @@
...
+	"io/ioutil"
+	"net"
+	"sync"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/vmware/vic/pkg/telnet"
+
+	"time"
+
+	"io"
```
anchal-agrawal: ditto for `time` and `io`.

cmd/vspc/handlers.go
kreamyx: Done!

cmd/vspc/vspc.go
kreamyx: Done!

lib/portlayer/vspc/handlers.go
```diff
@@ -0,0 +1,190 @@
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
```
sgairo: Are these entirely new files? If so, they should all have just the 2017 copyright.

lib/portlayer/vspc/handlers.go
```diff
@@ -0,0 +1,190 @@
...
+	for _, v := range seq {
+		if v == telnet.Iac {
+			escapedSeq = append(escapedSeq, telnet.Iac)
+		}
+		escapedSeq = append(escapedSeq, v)
+	}
+	secret := make([]byte, 4)
+	var escapedSecret []byte
+	rand.Read(secret)
+	// escapiaing Iac
```
sgairo: "escaping"

lib/portlayer/vspc/vspc.go
```diff
@@ -0,0 +1,227 @@
...
+	logFileName                         = "/var/log/vic/vspc.log"
+)
+const remoteConnReadDeadline = 1 * time.Second
+
+// VM is a struct that represents a VM
+type VM struct {
+	sync.Mutex
+	// this is the current connection between the vspc and the containerVM serial port
+	containerConn *telnet.Conn
+	// in case of vmotion, at some point we will have two telnet connections to th containerVM
```
sgairo: "the"

lib/portlayer/vspc/handlers.go
kreamyx: Done!

lib/portlayer/vspc/handlers.go
kreamyx: good catch. thanks!

lib/portlayer/vspc/vspc.go
kreamyx: Done!

lib/portlayer/vspc/helpers.go
```diff
@@ -0,0 +1,99 @@
...
+		return false
+	}
+	return cmd[0] == telnet.Sb && cmd[1] == VmwareExt && cmd[2] == VMVCUUID
+}
+
+func getVMUUID() []byte {
+	return []byte{telnet.Iac, telnet.Sb, VmwareExt, GetVMVCUUID, telnet.Iac, telnet.Se}
+}
+
+func lookupVCHIP() (net.IP, error) {
```
caglar10ur: look we have another here 🎉

lib/portlayer/vspc/vspc.go
```diff
@@ -0,0 +1,230 @@
...
+	GetVMBiosUUID         byte          = 85
+	VMLocationUUID        byte          = 86
+	GetsVMLocationUUID    byte          = 87
+	VMHealthCheckPeriod   time.Duration = 10 * time.Second
+	vspcPort              int           = 2377
+)
+const remoteConnReadDeadline = 1 * time.Second
+
+// VM is a struct that represents a VM
+type VM struct {
```
caglar10ur: The name could be something else (I don't have a good suggestion)

lib/portlayer/vspc/vspc.go
```diff
@@ -0,0 +1,230 @@
...
+	sync.Mutex
```
caglar10ur: don't embed the mutex as this exports Lock/Unlock to callers, make it ```m sync.Mutex```

lib/portlayer/vspc/vspc.go
```diff
@@ -0,0 +1,230 @@
...
+		containerConn:    tc,
+		inVmotion:        false,
+		vmotionStarted:   make(chan chan struct{}),
+		vmotionCompleted: make(chan chan struct{}),
+	}
+}
+
+// Vspc is all the vspc singletons
+type Vspc struct {
+	vmManagerMu *sync.Mutex
```
caglar10ur: why is this a pointer?

lib/portlayer/vspc/vspc.go
```diff
@@ -0,0 +1,230 @@
...
+	hdlr := Handler{vspc}
+	opts := telnet.ServerOpts{
+		Addr:        fmt.Sprintf("%s:%d", address, port),
+		ServerOpts:  []byte{telnet.Binary, telnet.Sga, telnet.Echo},
+		ClientOpts:  []byte{telnet.Binary, telnet.Sga, VmwareExt},
+		DataHandler: hdlr.DataHdlr,
+		CmdHandler:  hdlr.CmdHdlr,
+	}
+	vspc.Server = telnet.NewServer(opts)
+	go vspc.monitorVMConnections()
```
caglar10ur: IMHO NewX shouldn't take any action but just return the said struct. This could be a Start/Monitor etc. method on Vspc

lib/portlayer/vspc/vspc.go
```diff
@@ -0,0 +1,230 @@
...
+	return vspc
+}
+
+// getVM returns the VM struct from its uuid
+func (vspc *Vspc) getVM(uuid string) (*VM, bool) {
```
caglar10ur: drop the get - https://golang.org/doc/effective_go.html#Getters :)

lib/portlayer/vspc/vspc.go
```diff
@@ -0,0 +1,230 @@
...
+
+// addVM adds a VM to the map
+func (vspc *Vspc) addVM(uuid string, vm *VM) {
+	vspc.vmManagerMu.Lock()
+	defer vspc.vmManagerMu.Unlock()
+	vspc.vmManager[uuid] = vm
+}
+
+// relayReads reads from the AttachServer connection and relays the data to the telnet connection
+func (vspc *Vspc) relayReads(containervm *VM, conn net.Conn) {
```
caglar10ur: see my comment on the line where you are calling this

lib/portlayer/vspc/vspc.go
```diff
@@ -0,0 +1,230 @@
...
+		case ch := <-containervm.vmotionStarted:
+			vmotion = true
+			ch <- struct{}{}
+			log.Infof("vspc started to buffer data coming from the remote system")
+		case ch := <-containervm.vmotionCompleted:
+			vmotion = false
+			ch <- struct{}{}
+			log.Infof("vspc stopped buffering data coming from the remote system")
+		default:
+			b := make([]byte, 4096)
```
caglar10ur: we are copying back and forth here so why we are not replacing this while block with io.Copy?

cmd/port-layer-server/main.go
```diff
@@ -92,6 +96,24 @@ func main() {
...
+		vspc := vspc.NewVspc(vchIP.String(), constants.SerialOverLANPort, "127.0.0.1", constants.AttachServerPort, doneCh)
+		for {
+			_, err := vspc.Accept()
+			if err != nil {
+				log.Errorf("vSPC cannot accept connection: %v", err)
+				doneCh <- true
+				log.Errorf("vSPC exiting...")
+				return
+			}
+		}
```
kreamyx: Done!

lib/portlayer/vspc/handlers.go
```diff
@@ -0,0 +1,195 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
```
kreamyx: done!

lib/portlayer/vspc/handlers.go
```diff
@@ -0,0 +1,195 @@
...
+	vm, exists := h.vspc.vmFromTelnetConn(tc)
+	if !exists {
+		log.Errorf("vspc datahandler cannot find the vm with the provided data")
+	}
+	vm.remoteConn.Write(b)
+}
+
+// CmdHdlr is the telnet command handler
+func (h *Handler) CmdHdlr(w io.Writer, b []byte, tc *telnet.Conn) {
+	if isKnownSuboptions(b) {
```
kreamyx: hmm that requires switching over the contents of the byte slice:)

lib/portlayer/vspc/handlers.go
```diff
@@ -0,0 +1,195 @@
...
+		h.HandleVmotionComplete(w, tc, b)
+	} else if isVmotionAbort(b) {
+		log.Infof("vspc received VMOTION-ABORT command")
+		h.HandleVmotionAbort(w, b)
+	}
+}
+
+// HandleVMUUID handles the telnet vm-uuid response
+func (h *Handler) HandleVMUUID(w io.Writer, tc *telnet.Conn, b []byte) {
+	vmuuid := strings.Replace(string(b[3:len(b)-1]), " ", "", -1)
```
kreamyx: Actually all the handlers should be unexported. In fact, the only things that this pkg should export are `New`, `Start`, and `Stop`. Everything else is internal. Regarding your question, it is because the uuid we get back from the telnet command response includes extra spaces between the characters. We have no control over that response. It comes from the ESX host of the containerVM

lib/portlayer/vspc/handlers.go
```diff
@@ -0,0 +1,195 @@
...
+		vm.Lock()
+		defer vm.Unlock()
+		vm.prevContainerConn = vm.containerConn
+		vm.containerConn = tc
+	}
+}
+
+// HandleKnownSuboptions handles the known suboptions telnet command
+func (h *Handler) HandleKnownSuboptions(w io.Writer, b []byte) {
+	log.Infof("handling KNOWN-SUBOPTIONS")
```
kreamyx: done!

lib/portlayer/vspc/handlers.go
```diff
@@ -0,0 +1,195 @@
...
+	if vm, exists := h.vspc.vmFromTelnetConn(tc); exists {
+		vm.Lock()
+		vm.inVmotion = true
+		vm.Unlock()
+		ch := make(chan struct{})
+		vm.vmotionStarted <- ch
+		<-ch
+	}
+	log.Infof("handling VMOTION-BEGIN")
+	seq := b[3 : len(b)-1]
```
kreamyx: Each slicing, even though all of them started at 3, represents different type of information base on the telnet command in question. Coming up with a proper function name would be challenging:)

lib/portlayer/vspc/handlers.go
```diff
@@ -0,0 +1,195 @@
...
+	var escapedSeq []byte
+	for _, v := range seq {
+		if v == telnet.Iac {
+			escapedSeq = append(escapedSeq, telnet.Iac)
+		}
+		escapedSeq = append(escapedSeq, v)
+	}
+	secret := make([]byte, 4)
+	var escapedSecret []byte
+	rand.Read(secret)
```
kreamyx: Thanks! I will add it

lib/portlayer/vspc/handlers.go
caglar10ur: can't you call those functions in a switch

```
switch {
case isKnownSuboptions(b):
...
case XYZ(b):
   ...
default:
	panic()
}
```

lib/portlayer/vspc/handlers.go
caglar10ur: three2len :)

lib/portlayer/vspc/vspc.go
kreamyx: that's right. The struct should be unexported though!

lib/portlayer/vspc/handlers.go
kreamyx: yeah that works!

lib/portlayer/vspc/vspc.go
```diff
@@ -0,0 +1,230 @@
...
+	prevContainerConn *telnet.Conn
+	// inVmotion is a boolean denoting whether this VM is in a state of vmotion or not
+	inVmotion bool
+
+	// remoteConn is the remote-system connection
+	// this is the connection between the vspc and the attach-server
+	remoteConn net.Conn
+
+	vmUUID           string
+	vmotionStarted   chan chan struct{}
```
kreamyx: done!

lib/portlayer/vspc/vspc.go
kreamyx: In this particular scenario, it can be a pointer or not. But I usually prefer pointers to protect against the scenario when the struct is passed by value but since this is not the case here then a pointer is not necessary

lib/portlayer/vspc/vspc.go
kreamyx: yes I agree. fixed. now it is in Start()

lib/portlayer/vspc/vspc.go
kreamyx: done!

lib/portlayer/vspc/handlers.go
```diff
@@ -0,0 +1,195 @@
...
+		log.Infof("attempting to connect to the attach server")
+		remoteConn, err := net.Dial("tcp", fmt.Sprintf("%s:%d", h.vspc.attachSrvrAddr, h.vspc.attachSrvrPort))
+		if err != nil {
+			log.Errorf("cannot connect to the attach-server: %v", err)
+		}
+		vm.remoteConn = remoteConn
+		vm.vmUUID = vmuuid
+		h.vspc.addVM(vmuuid, vm)
+		// relay Reads from the remote system connection to the telnet connection associated with this vm
+		go h.vspc.relayReads(vm, remoteConn)
```
kreamyx: I don't agree with the writer that ioutil.ReadAll is to be avoided at all costs :). I guess it is only bad if the user does not really know what concrete type the underlying io.Reader which is not the case here. That said, I looked into the bytes package and there is a more efficient way to read the buffer content than ioutil.ReadAll by using buffer.Bytes() and the buffer.Reset()

lib/portlayer/vspc/vspc.go
caglar10ur: go vet complains about that so I personally trust that :P

lib/vspc/handlers.go
```diff
@@ -0,0 +1,204 @@
...
+		h.handleVmotionPeer(w, b)
+	case isVMUUID(b):
+		log.Infof("vspc received VMUUID command")
+		h.handleVMUUID(w, tc, b)
+	case isVmotionComplete(b):
+		log.Infof("vspc received VMOTION-COMPLETE command")
+		h.handleVmotionComplete(w, tc, b)
+	case isVmotionAbort(b):
+		log.Infof("vspc received VMOTION-ABORT command")
+		h.handleVmotionAbort(w, tc, b)
```
caglar10ur: just wondering, should we log or panic in the default case?

Also maybe we could drop handle suffix from those functions as it stutters (h is already handler so h.DoProxy, h.VMUUID etc.)

lib/vspc/handlers.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+}
+
+// handleVMUUID handles the telnet vm-uuid response
+func (h *handler) handleVMUUID(w io.Writer, tc *telnet.Conn, b []byte) {
+	defer trace.End(trace.Begin("handling VMUUID"))
+	vmuuid := strings.Replace(string(b[3:len(b)-1]), " ", "", -1)
```
caglar10ur: [cosmetic] an empty line after defer would be perfect

lib/vspc/handlers.go
```diff
@@ -0,0 +1,204 @@
...
+		cvm.Lock()
+		defer cvm.Unlock()
+		cvm.prevContainerConn = cvm.containerConn
+		cvm.containerConn = tc
+	}
+}
+
+// handleKnownSuboptions handles the known suboptions telnet command
+func (h *handler) handleKnownSuboptions(w io.Writer, b []byte) {
+	defer trace.End(trace.Begin("handling KNOWN-SUBOPTIONS"))
```
caglar10ur: [cosmetic] an empty line after defer would be perfect

lib/vspc/handlers.go
```diff
@@ -0,0 +1,204 @@
...
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/vmware/vic/pkg/telnet"
+	"github.com/vmware/vic/pkg/trace"
+)
+
+// handler is the handler struct for the vspc
+type handler struct {
+	vspc *Vspc
```
caglar10ur: [cosmetic] you could embed *Vspc directly which simplifies h.vspc.XYZ type calls

lib/vspc/handlers.go
```diff
@@ -0,0 +1,204 @@
...
+
+	if bytes.IndexByte(suboptions, GetVMVCUUID) != -1 && bytes.IndexByte(suboptions, VMVCUUID) != -1 {
+		resp = append(resp, getVMUUID()...)
+	}
+	w.Write(resp)
+}
+
+// handleDoProxy handles the DO-PROXY telnet command
+func (h *handler) handleDoProxy(w io.Writer, b []byte) {
+	defer trace.End(trace.Begin("handling DO-PROXY"))
```
caglar10ur: [cosmetic] an empty line after defer would be perfect

lib/vspc/handlers.go
```diff
@@ -0,0 +1,204 @@
...
+	var resp []byte
+	resp = append(resp, []byte{telnet.Iac, telnet.Sb, VmwareExt, WillProxy, telnet.Iac, telnet.Se}...)
+	log.Debugf("response to DO-PROXY: %v", resp)
+	w.Write(resp)
+}
+
+// handleVmotionBegin handles the VMOTION-BEGIN telnet command
+func (h *handler) handleVmotionBegin(w io.Writer, tc *telnet.Conn, b []byte) {
+	defer trace.End(trace.Begin("handling VMOTION-BEGIN"))
```
caglar10ur: [cosmetic] an empty line after defer would be perfect

lib/vspc/handlers.go
```diff
@@ -0,0 +1,204 @@
...
+	resp = append(resp, escapedSeq...)
+	resp = append(resp, escapedSecret...)
+	resp = append(resp, telnet.Iac, telnet.Se)
+	log.Debugf("response to VMOTION-BEGIN: %v", resp)
+	w.Write(resp)
+}
+
+// handleVmotionPeer handles the VMOTION-PEER telnet command
+func (h *handler) handleVmotionPeer(w io.Writer, b []byte) {
+	defer trace.End(trace.Begin("handling VMOTION-PEER"))
```
caglar10ur: [cosmetic] an empty line after defer would be perfect

lib/vspc/handlers.go
```diff
@@ -0,0 +1,204 @@
...
+	resp = append(resp, []byte{telnet.Iac, telnet.Sb, VmwareExt, VmotionPeerOK}...)
+	resp = append(resp, cookie...)
+	resp = append(resp, telnet.Iac, telnet.Se)
+	log.Debugf("response to VMOTION-PEER: %v", resp)
+	w.Write(resp)
+}
+
+// handleVmotionComplete handles the VMOTION-Complete telnet command
+func (h *handler) handleVmotionComplete(w io.Writer, tc *telnet.Conn, b []byte) {
+	defer trace.End(trace.Begin("handling VMOTION-COMPLETE"))
```
caglar10ur: [cosmetic] an empty line after defer would be perfect

lib/vspc/handlers.go
```diff
@@ -0,0 +1,204 @@
...
+	var escapedSeq []byte
+	for _, v := range seq {
+		if v == telnet.Iac {
+			escapedSeq = append(escapedSeq, telnet.Iac)
+		}
+		escapedSeq = append(escapedSeq, v)
+	}
+	secret := make([]byte, 4)
+	var escapedSecret []byte
+	rand.Read(secret)
```
caglar10ur: Seed is still missing

lib/vspc/handlers.go
```diff
@@ -0,0 +1,204 @@
...
+		log.Info("vMotion completed successfully")
+	} else {
+		log.Errorf("couldnt find previous information of vm after vmotion (vmuuid: %s)", cvm.vmUUID)
+	}
+
+}
+
+// handleVmotionAbort handles the VMOTION-abort telnet command
+func (h *handler) handleVmotionAbort(w io.Writer, tc *telnet.Conn, b []byte) {
+	defer trace.End(trace.Begin("handling VMOTION-ABORT"))
```
caglar10ur: [cosmetic] an empty line after defer would be perfect

caglar10ur:
Overall LGTM with minor nitpicks. We should have a review from @hickeng before merging it.

lib/vspc/handlers.go
kreamyx: crypto/rand does not have/need a seed https://golang.org/pkg/crypto/rand/
math/rand does.

lib/vspc/handlers.go
kreamyx: hmmm I think we can just log an Error. According to the vsphere documentation, there should be no more vmware telnet commands. We should be made aware should any unexpected command arrives. Panicking on the other hand would terminate the port-layer server which begs the question, does the port-layer server restart if it terminates unexpectedly?

kreamyx:
@mhagen-vmware the unit tests are in now!

lib/portlayer/attach/attach.go
```diff
@@ -70,6 +70,12 @@ func toggle(handle *exec.Handle, connected bool) (*exec.Handle, error) {
...
 
 	serviceURI := fmt.Sprintf("tcp://%s:%d", ip, constants.SerialOverLANPort)
+	proxyURI := fmt.Sprintf("telnet://%s:%d", ip, constants.SerialOverLANPort)
+
+	if b.ProxyURI == proxyURI && c.Connected == connected {
+		log.Debugf("Already in the desired state (connected: %t, proxyURI: %s)", connected, proxyURI)
+		return handle, nil
+	}
 
 	if b.ServiceURI == serviceURI && c.Connected == connected {
```
hickeng: Only one of proxyURI vs serviceURI is used if I recall correctly.
Are we keeping this because proxyURI isn't needed on ESX only? If that's the case are we suppressing launch of vspc for esx only?

kreamyx: The serviceURI is not used at all but ESX complains if it is not there

kreamyx: In fact, the serviceURI is forwarded to the proxy to tell what the address of the remote system should be. In our case, it is known and fixed (localhost:2379) which is the address of the telnet server

PR #4303, opened 2017-03-20
Added packer scripts to pull dev image from Admiral. Issue: 4256.

lcastellano:
Also add scripts to start and stop docker.


lcastellano:
I split it because I thought we needed to start/stop Docker for Harbor as well. With this approach we can start Docker, pull Admiral, pull Harbor, then stop Docker.

frapposelli:
I'd like each script to be as idempotent as possible so we only need a begin and an end script, everything else can be moved around.

installer/packer/scripts/admiral_setup.sh
```diff
@@ -0,0 +1,31 @@
...
+echo "Docker started"
+
+# pull admiral image
+echo "Pulling Admiral Docker image.."
+docker pull docker.io/vmware/admiral:dev
+echo "Pulled Admiral image"
+
+# stop docker
+echo "stopping Docker .."
+systemctl daemon-reload
```
frapposelli: no need to `daemon-reload` here as we're not changing systemd units

PR #4308, opened 2017-03-20
[full ci] Block maintenance mode. Don't stop running containers.

gigawhitlocks:
Current behavior is such that pressing the maintenance mode button on the VC UI causes containers to stop (but not the applianceVM) and blocks maintenance mode. This backs out most of https://github.com/vmware/vic/commit/63fd97ed5864bfa78a2dad1f7c203a2c65566ad1 to block maintenance mode altogether and prevent it from powering down containers. All running containers block maintenance mode, but it was blocked anyway on `master` because the applianceVM doesn't get shut down. I think this is acceptable behavior (either we can shut down everything or we should shut down nothing) but of course if different behavior is desired (I notice that the original issue only specifies that _attached_ containers should stop maintenance mode, but since the appliance VM stops it anyway..) then I am more than happy to go ahead and implement whatever other behavior we desire.

Full CI to be sure there aren't any unintended side effects and to be sure I did not break tests, which I very well might've ;D


caglar10ur:
I'm not sure whether we want to remove the event code - thinking that we might be using it someday, @cgtexmex what do you think? If that's the case we could just drop the subscription to the host event and callback bits @lib/portlayer/exec/exec.go otherwise LGTM

gigawhitlocks:
Roger; PullApprove has him on the review so his approval is necessary before merging anyway.

cgtexmex:
I have no concerns technically, so from that perspective LGTM.  I do have a concern from a usability / business impact perspective.

The default behavior will now be to block maintenance mode for any host that hosts a containerVM.  IMO this nullifies a large value prop of vSphere and undermines vic -- specifically the ability for the VI Admin to maintain the cluster w/o impacting the end user.    I think the default behavior should be to only block maintenance mode when hosted containerVMs have active attach sessions -- otherwise just migrate the containerVM.  I think @karthik-narayan @corrieb and / or others should weigh in on the usability / business impact.

gigawhitlocks:
Just for the record, the current behavior of the product on `master` requires the VI Admin to power off the appliance VM manually, so maintenance mode will already affect the end-user in the current state of the code. 

Changing maintenance mode to power off all unattached containers and then the appliance VM is new functionality. That's fine, but if we want to do that we should reflect it through story points by resizing this story or writing a new story that this one is dependent upon.

cgtexmex:
@gigawhitlocks in a vSphere cluster a poweredOn VM will be auto-migrated when putting a host into maintenance mode.  So I'd expect that the appliance VM would just migrate to another available host in the cluster -- I don't think there's any serial port tomfoolery that would keep it pinned to a specific host...but I could be wrong...

In stand-alone ESXi maintenance mode will be blocked until all VMs are poweredOff. 

So I thought the previous change was to powerOff the tethered VMs to allow maintenance mode w/o any VI Admin intervention.  The non-tethered VMs would just migrate based on the behavior of vSphere...that would include the appliance.  Again..possibly I'm wrong...

gigawhitlocks:
Nope, you're probably right and my testing method is wrong. I'm going to close this PR for now and go do this.

PR #4309, opened 2017-03-20
[skip ci] Add George to config component

gigawhitlocks:
Adds George to the config component in PullApprove. Does what it says on the tin.


gigawhitlocks:
dummy commits cause I forgot [skip ci] sigh

PR #4313, opened 2017-03-20
Get vmkernel log in tests

### LINKED ISSUE #4167: Changes to the ssh settings on the ESX servers is causing vmkernel.log not to be collected anymore
We have a regression in the logs that we collect, in our current CI we disabled SSH by password and switched to public key auth only, so now the collection of vmkernel.log which used scp is not working.  We need to switch to using the public key auth to grab that file so we can have it for debugging purposes again.

jakedsouza:
Changes made
1. Added ssh private key during test run to download vmkernel logs from esxi
2. The key is stored as a secret and written to disk during test run
3. Also changed image from golang 1.8 to the integration test image.
   The integration test image is based of go1.8 and has already run
   apt get-update

Fixes #4167


tests/resources/VCH-Util.robot
```diff
@@ -187,7 +187,8 @@ Gather Logs From Test Server
     Remove File  vic-admin-cookies
     ${out}=  Run  govc datastore.download %{VCH-NAME}/vmware.log %{VCH-NAME}-vmware.log
     Should Contain  ${out}  OK
-    ${out}=  Run  sshpass -p %{TEST_PASSWORD} scp %{TEST_USERNAME}@%{TEST_URL}:/var/log/vmkernel.log ./vmkernel.log
+    ${out}=  Run  sshpass -p %{TEST_PASSWORD} scp -o StrictHostKeyChecking\=no %{TEST_USERNAME}@%{TEST_URL}:/var/log/vmkernel.log ./vmkernel.log
```
mhagen-vmware: if you are using the private key, do you need to use sshpass and the password anymore?

.drone.yml
```diff
@@ -28,7 +28,7 @@ build:
       - tests/wait_until_previous_builds_complete.sh
 
   vic:
-    image: golang:1.8
+    image: $${TEST_BUILD_IMAGE=gcr.io/eminent-nation-87317/vic-integration-test:1.22}
```
mhagen-vmware: @dougm you ok with this change?

tests/resources/VCH-Util.robot
jakedsouza: I left the sshpass, which will continue to work for esxi/vcenters where we still have ssh enabled with username/password e.g for nimbus and I assume nightly builds. Also for people running tests against their own environments.

jakedsouza:
> Note that you can use govc logs command to get vmkernel.log w/o the ssh hassles.

Yes, I did look at using govc logs. The support bundle generation takes quite a long time, This is called often enough that it would increase test run time.

dougm:
No, see the examples in 'govc logs -h'. 
It doesn't generate a bundle, that's 'logs.download'

jakedsouza:
> No, see the examples in 'govc logs -h'.
> It doesn't generate a bundle, that's 'logs.download'

I see. Will try it and update the PR.

.drone.yml
caglar10ur: just curios, what is eminent-nation-87317? If that's the name of the repo, can't we rename it to something else?

tests/integration-test.sh
```diff
@@ -22,6 +22,7 @@ dpkg -l > package.list
 
 buildinfo=$(drone build info vmware/vic $DRONE_BUILD_NUMBER)
 
+
```
dougm: Why add this blank line and the line below?

tests/resources/VCH-Util.robot
```diff
@@ -187,7 +187,7 @@ Gather Logs From Test Server
     Remove File  vic-admin-cookies
     ${out}=  Run  govc datastore.download %{VCH-NAME}/vmware.log %{VCH-NAME}-vmware.log
     Should Contain  ${out}  OK
-    ${out}=  Run  sshpass -p %{TEST_PASSWORD} scp %{TEST_USERNAME}@%{TEST_URL}:/var/log/vmkernel.log ./vmkernel.log
+    Run  govc logs -log=vmkernel -n=10000 > vmkernel.log
```
dougm: :+1:

tests/integration-test.sh
jakedsouza: Removed the file.

.drone.yml
jakedsouza: eminent-nation-87317 is the google cloud project id. AFAIK we can not change it.

Issue #4253, opened 2017-03-15
Create systemd unit for Admiral start/stop


frapposelli:
User Statement: Want to have Admiral started up and shut down on startup/shutdown of the OVA

Acceptance Criteria: Must support startup and clean shutdown of the Admiral application

PR #4327, opened 2017-03-21
Add Harbor to the OVA Build

frapposelli:
This is work in progress, still requires fine tuning and systemd unit for start/stop as of #4253


installer/packer/scripts/harbor/configure_harbor.sh
```diff
@@ -0,0 +1,189 @@
...
+    return
+  fi
+
+  echo "Use the existing CA, certificate and key file"
+  echo "Copy CA certificate to $ca_download_dir"
+  cp $ca_cert $ca_download_dir/
+}
+
+function detectHostname {
+  hostname=$(hostnamectl status --static) || true
```
hickeng: `hostnamectl` might do this - the hostname you want for the cert is the one that dns provides, not the one configured for the system (yes, they could be the same). I've ended up using `nslookup` or equivalent once the appropriate interface is determined.
If there's multiple interfaces then this could be multi-valued and the alternates need to go into subjectAltName fields for IP address and DNS name respectively.

installer/packer/scripts/harbor/configure_harbor.sh
```diff
@@ -0,0 +1,189 @@
...
+  appliance.email_password 
+  appliance.email_from 
+  appliance.email_ssl
+  harbor.admin_password
+  harbor.db_password
+  harbor.gc_enabled
+)
+
+hostname=""
+ip_address=$(ip route get 8.8.8.8 | gawk 'NR==1 {print $NF}')
```
hickeng: not sure what happens here when there's no route to google nameservers.

installer/packer/scripts/harbor/configure_harbor.sh
```diff
@@ -0,0 +1,189 @@
...
+  echo "self-signed" > $flag
+  echo "Copy CA certificate to $ca_download_dir"
+  cp $ca_cert $ca_download_dir/
+}
+
+function secure {
+  ssl_cert=$(ovfenv -k harbor.ssl_cert)
+  ssl_cert_key=$(ovfenv -k harbor.ssl_cert_key)
+  if [ -n "$ssl_cert" ] && [ -n "$ssl_cert_key" ]; then
+    echo "ssl_cert and ssl_cert_key are both set, using customized certificate"
```
hickeng: This is trusting that they got the cert inputs correct - generally viable, could add an `openssl` check of the cert/key pair to validate.

installer/packer/scripts/harbor/harbor.cfg
```diff
@@ -0,0 +1,113 @@
+## Configuration file of Harbor
+
+#The IP address or hostname to access admin UI and registry service.
+#DO NOT use localhost or 127.0.0.1, because Harbor needs to be accessed by external clients.
+hostname = reg.mydomain.com
+
+#The protocol for accessing the UI and token/notification service, by default it is http.
```
hickeng: This will come up in the product security review - should be secure by default and with appropriate minimum TLS versions/approved cipher suites.

installer/packer/scripts/harbor/harbor.cfg
```diff
@@ -0,0 +1,113 @@
...
+#The expiration time (in minute) of token created by token service, default is 30 minutes
+token_expiration = 30
+
+#The flag to control what users have permission to create projects
+#Be default everyone can create a project, set to "adminonly" such that only admin can create project.
+project_creation_restriction = everyone
+
+#Determine whether the job service should verify the ssl cert when it connects to a remote registry.
+#Set this flag to off when the remote registry uses a self-signed or untrusted certificate.
+verify_remote_cert = on
```
hickeng: There may need to be a way to add custom CAs to the system root pool to be able to perform this verification.

installer/packer/scripts/systemd/network/network-config.sh
```diff
@@ -13,16 +13,19 @@
...
   set -- $(( ($2 - ${#3})*2 )) ${1%%${3%%.*}*} 
   echo $(( $1 + (${#2}/4) )) 
 }
 
+fqdn=$(ovfenv --key network.fqdn)
 network_address=$(ovfenv --key network.ip0)
 network_conf_file=/etc/systemd/network/09-vic.network
 
+# Set hostname
+[[ x$fqdn != "x" ]] && ( hostnamectl set-hostname $fqdn )
```
hickeng: +1 for proper variable set check

installer/packer/scripts/harbor/configure_harbor.sh
hickeng: I see in network-config.sh that hostname is being set to an ovf supplied value - still, consider a check that the FQDN resolves to the supplied IPs in the deployed environment (we had bonneville issues with asymmetric DNS lookups across an organisation).

installer/packer/scripts/harbor/configure_harbor.sh
frapposelli: I found this to be the easiest way to gather the source ip for the interface that is associated with the default route, if no default route is available it will error out.

installer/packer/scripts/harbor/harbor.cfg
frapposelli: This is the default configuration file, it is managed by the `configure_harbor.sh` script and is forced to be https only, the cipher suite is determined by the nginx config and provided by the harbor team https://github.com/vmware/harbor/blob/master/make/common/templates/nginx/nginx.https.conf#L34-L37

installer/packer/scripts/harbor/harbor.cfg
frapposelli: we can deal with this requirement by creating a procedure to add a custom CA manually after the deployment.

installer/packer/scripts/harbor/configure_harbor.sh
frapposelli: This is a complex use case, I'm trying to narrow down the permutations leveraging what the Harbor team has done already.

If the user provides a certificate, we simply use that, if the user doesn't provide a cert, we take the FQDN as source of truth, if FQDN is default (`localhost.localdomain`) we add a SAN with the IP address.

This configuration process happens at every boot of the appliance so the user can recover easily from botched deployments

installer/packer/scripts/harbor/configure_harbor.sh
frapposelli: That's a great idea, I'll revise the script in the next iteration to check for that.

installer/packer/scripts/harbor/configure_harbor.sh
frapposelli: Actually this might not be the best idea if the deployment is done on a non-routed network, will amend this right now

PR #4332, opened 2017-03-22
Move Harbor download URL to a public one

frapposelli:


frapposelli:
self-LGTM on this one

PR #4340, opened 2017-03-22
add systemd unit files for admiral

maplain:
Add systemd unit files for admiral: https://github.com/vmware/vic/issues/4253


installer/packer/scripts/systemd/admiral/admiral.service
```diff
@@ -0,0 +1,15 @@
...
+Description=Admiral is a highly scalable and very lightweight Container Management platform for deploying and managing container based applications.
+Documentation=https://github.com/vmware/admiral
+After=docker.service systemd-networkd.service systemd-resolved.service
+Requires=docker.service
+
+[Service]
+Type=simple
+Restart=on-failure
+RestartSec=5
+ExecStart=/usr/local/bin/docker run -d -p 8282:8282 --name vic-admiral docker.io/vmware/admiral:dev
```
frapposelli: Can you add an `ExecStartPre` here to ensure the container is no longer there? that way we're sure every start is a clean slate environment.

also, you can omit `docker.io/` from the container name

installer/packer/scripts/systemd/vic-service.target
```diff
@@ -0,0 +1,8 @@
+[Unit]
```
frapposelli: what's the purpose of this target?

installer/packer/scripts/systemd/admiral/admiral.service
lcastellano: It is better to run Admiral with the log parameters suggested in the config doc:

--log-driver=json-file --log-opt max-size=1g --log-opt max-file=10

may be even less than 1g for max-size

frapposelli: I've noticed I've commented on an old commit, still I'd rather prefer to remove the container on `ExecPreStart` rather than as part of the stop.

maplain: yeah, sure 👌

installer/packer/scripts/systemd/vic-service.target
maplain: to group services together thus we have a synchronization point to wait on if we want to do tests on interactions between services. If this is unnecessary, I'll remove it.

installer/packer/scripts/systemd/admiral/admiral.service
maplain: got it. thx

installer/packer/scripts/systemd/admiral/admiral.service
```diff
@@ -0,0 +1,16 @@
...
+After=docker.service systemd-networkd.service systemd-resolved.service
+Requires=docker.service
+
+[Service]
+Type=simple
+Restart=on-failure
+RestartSec=5
+ExecStartPre=-/usr/local/bin/docker stop vic-admiral && -/usr/local/bin/docker rm vic-admiral
+ExecStart=/usr/local/bin/docker run -d -p 8282:8282 --name vic-admiral --log-driver=json-file --log-opt max-size=1g --log-opt max-file=10 vmware/admiral:dev
+ExecStop=/usr/local/bin/docker stop vic-admiral && /usr/local/bin/docker rm vic-admiral
```
frapposelli: Ideally, this should be idempotent, `ExecStop` should only run stop, while `ExecStartPre` should only run rm

PR #4343, opened 2017-03-22
Fix insecure-registry URL parsing in GO1.8, #4141

### LINKED ISSUE #4141: vic machine fails install when insecure-registry specified
if installing a VCH and you specify an insecure registry the vic-machine parsing of the flag fails.

vic built from source using `go1.8`

vic-machine create command:
`vic-machine-linux create -t 172.16.73.170 -u root -p vagrant -n vickie -vs datastore1/vickie-vols:default -vs datastore1/vickie-vols:jojo -k -f --insecure-registry 172.16.73.147:5000`

vic-machine result:
```
Mar  3 2017 22:59:57.228Z INFO  ### Installing VCH ####
Mar  3 2017 22:59:57.229Z WARN  Using administrative user for VCH operation - use --ops-user to improve security (see -x for advanced help)
Mar  3 2017 22:59:57.229Z WARN  Configuring without TLS - all communications will be insecure
Mar  3 2017 22:59:57.229Z ERROR --------------------
Mar  3 2017 22:59:57.230Z ERROR vic-machine-linux create failed: 172.16.73.147:5000 is an invalid format for registry url
```

This worked with `go version 1.7.x` but fails with `go version 1.8` -- specifically the call to `url.Parse(registry_string)` throws an error.  An example is at:  https://play.golang.org/p/SwZAq5NszD -- this is executing in go1.8, but if you run locally with go1.7x no error will be thrown.

Additionally, if you provide a url that will pass the validation then the pull from the private registry doesn't work.

From my perspective this effectively breaks insecure-registry support.

lcastellano:
This change includes:
- New URL parsing function (originally from GOVMOMI)
- Changed vic-machine to use new parser
- Changed backend to read host:port from the Host component of the URL

Fixes #4141

### DIFF
--- a/lib/apiservers/engine/backends/backends.go
+++ b/lib/apiservers/engine/backends/backends.go
@@ -112,7 +112,11 @@ func Init(portLayerAddr, product string, config *config.VirtualContainerHostConf
 
 	serviceOptions := registry.ServiceOptions{}
 	for _, r := range insecureRegs {
-		insecureRegistries = append(insecureRegistries, r.Path)
+		// host:port was previously stored in the Path
+		// component of the URL. Go 1.8 parser puts it
+		// the Host component. VCH upgrade must move it
+		// from Path to Host.
+		insecureRegistries = append(insecureRegistries, r.Host)
 	}
 	if len(insecureRegistries) > 0 {
 		serviceOptions.InsecureRegistries = insecureRegistries
@@ -253,7 +257,7 @@ func createImageStore() error {
 }
 
 func InsecureRegistries() []string {
-	registries := make([]string, len(insecureRegistries))
+	registries := []string{}
 	for _, reg := range insecureRegistries {
 		registries = append(registries, reg)
 	}
--- a/lib/install/validate/validator.go
+++ b/lib/install/validate/validator.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"context"
+	"regexp"
 
 	log "github.com/Sirupsen/logrus"
 	units "github.com/docker/go-units"
@@ -154,6 +155,30 @@ func NewValidator(ctx context.Context, input *data.Data) (*Validator, error) {
 	}
 
 	return v, nil
+}
+
+var schemeMatch = regexp.MustCompile(`^\w+://`)
+
+// Starting from Go 1.8 the URL parser does not
+// work properly with URLs with no Scheme,
+// this function adds "https" as Scheme if necessary
+func ParseURL(s string) (*url.URL, error) {
+	var err error
+	var u *url.URL
+
+	if s != "" {
+		// Default the scheme to https
+		if !schemeMatch.MatchString(s) {
+			s = "https://" + s
+		}
+
+		u, err = url.Parse(s)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return u, nil
 }
 
 func (v *Validator) AllowEmptyDC() {
--- a/lib/install/validate/validator_test.go
+++ b/lib/install/validate/validator_test.go
@@ -32,10 +32,83 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/simulator"
 
 	"context"
+	"fmt"
 )
 
 type TestValidator struct {
 	Validator
+}
+
+func TestParseURL(t *testing.T) {
+	var hosts = []string{
+		"host.domain.com",
+		"host.domain.com:123",
+		"1.2.3.4",
+		"1.2.3.4:10",
+		"[2001:4860:0:2001::68]",
+		"[2001:db8:1f70::999:de8:7648:6e8]:123",
+	}
+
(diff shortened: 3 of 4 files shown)


lib/install/validate/validator.go
```diff
@@ -156,6 +157,30 @@ func NewValidator(ctx context.Context, input *data.Data) (*Validator, error) {
 	return v, nil
 }
 
+var schemeMatch = regexp.MustCompile(`^\w+://`)
```
caglar10ur: I really prefer using strings package (HasPrefix) instead of regex

dougm: How to in this case, since it could be "http", "https", maybe even another scheme.

lib/install/validate/validator.go
```diff
@@ -156,6 +157,30 @@ func NewValidator(ctx context.Context, input *data.Data) (*Validator, error) {
...
+
+// Starting from GO 1.8 the URL parser does not
```
dougm: Doc should start with the function name (to avoid go-lint warnings).
s/GO/Go/
s/Schema/Scheme/g

lib/install/validate/validator.go
caglar10ur: https://play.golang.org/p/mRrrzoOS1_ could work but I'm OK with the change if I'm the only one stuck on regex :)

lib/apiservers/engine/backends/backends.go
```diff
@@ -112,7 +112,13 @@ func Init(portLayerAddr, product string, config *config.VirtualContainerHostConf
 
 	serviceOptions := registry.ServiceOptions{}
 	for _, r := range insecureRegs {
-		insecureRegistries = append(insecureRegistries, r.Path)
+		/*
```
anchal-agrawal: Nitpick - we should use Go's idiomatic way of (non-package level) multi-line comments. We should stick to `//` for this.

https://golang.org/doc/effective_go.html#commentary
> Comments do not need extra formatting such as banners of stars.

gigawhitlocks:
Could we put the URL parsing function in a more generic location? I think it's going to be needed in places outside of the installer and it feels weird importing `lib/install/validator.go` to get the URL parsing function that you wrote.

jzt:
LGTM

mdubya66:
Approved for merge to unblock others

mdubya66:
Successful build was https://ci.vmware.run/vmware/vic/9035

lib/install/validate/validator.go
```diff
@@ -156,6 +157,30 @@ func NewValidator(ctx context.Context, input *data.Data) (*Validator, error) {
 	return v, nil
 }
 
+var schemeMatch = regexp.MustCompile(`^\w+://`)
+
+// Starting from Go 1.8 the URL parser does not
```
vburenin: This might be misleading comment. Is a Go bug? If so, then a reference to that bug needs to be added.
If it is a corrected behavior that was not good before then change a comment to be something like:
"ParseURL parses URL adding default https scheme if no scheme is provided."

lib/install/validate/validator.go
```diff
@@ -156,6 +157,30 @@ func NewValidator(ctx context.Context, input *data.Data) (*Validator, error) {
...
+// work properly with URLs with no Scheme,
+// this function adds "https" as Scheme if necessary
+func ParseURL(s string) (*url.URL, error) {
```
vburenin: It seems like this function can be drastically reduced to:

```
func ParseURL(s string) (*url.URL, error) {
	if s != "" && !schemeMatch.MatchString(s) {

		s = "https://" + s
	}
	return url.Parse(s)
}
```

lib/apiservers/engine/backends/backends.go
```diff
@@ -253,7 +257,7 @@ func createImageStore() error {
 }
 
 func InsecureRegistries() []string {
-	registries := make([]string, len(insecureRegistries))
+	registries := []string{}
```
vburenin: This is incorrect fix. Initial line had a bug not providing size only.
Correct fix is this:
```registries := make([]string, 0, len(insecureRegistries))```

File: lib/install/validate/validator_test.go [hunk 7]
```diff
@@ -32,12 +32,85 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/simulator"
 
 	"context"
+	"fmt"
```
vburenin: imports order needs to be fixed.

lib/install/validate/validator.go
```diff
@@ -22,6 +22,7 @@ import (
 	"strings"
 
 	"context"
+	"regexp"
```
vburenin: built-in import blocks should be merged.

dougm:
Just learned we already had similar code here: https://github.com/vmware/vic/blob/master/pkg/flags/url_flag.go#L29-L42

PR #4355, opened 2017-03-23
[specific ci=Group6-VIC-Machine]These two tests were innacurately set to true when they were doing --…

mhagen-vmware:


dougm:
lgtm

hickeng:
@mhagen-vmware This PR description is useless to me - I can see what it was meant to say from the code, but that's not really helpful.
There's also a high likelyhood that this PR would be merged with the PR title as the git commit message which is a crime against source control.

mhagen-vmware:
@hickeng this was a temporary branch PR. I kept it so that Loc could use it for the basis of his real PR to fix the actual problem.

hickeng:
@mhagen-vmware Ok - can those be created in the http://github.com/mhagen-vmware/vic repo if needed - in general all PRs should be coming from personal repos, then there's no need for a PR at all.

PR #4359, opened 2017-03-23
added docker create and start events

cgtexmex:
The docker persona event monitor will now publish create & start
events.  This is needed in support of docker stats.

### DIFF
--- a/lib/apiservers/engine/backends/eventmonitor.go
+++ b/lib/apiservers/engine/backends/eventmonitor.go
@@ -40,6 +40,18 @@ import (
 	eventtypes "github.com/docker/docker/api/types/events"
 )
 
+const (
+	containerDieEvent     = "die"
+	containerStopEvent    = "stop"
+	containerStartEvent   = "start"
+	containerCreateEvent  = "create"
+	containerRestartEvent = "restart"
+	containerAttachEvent  = "attach"
+	containerDetachEvent  = "detach"
+	containerKillEvent    = "kill"
+	containerResizeEvent  = "resize"
+)
+
 // for unit testing purposes
 type eventproxy interface {
 	StreamEvents(ctx context.Context, out io.Writer) error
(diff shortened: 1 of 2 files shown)


lib/apiservers/engine/backends/eventmonitor.go
```diff
@@ -198,19 +198,28 @@ func (m *PortlayerEventMonitor) monitor() error {
...
 
 	vc := cache.ContainerCache().GetContainer(event.Ref)
 	if vc == nil {
 		log.Errorf("Portlayer event for container %s but not found in cache", event.Ref)
 	}
+	var dockerEvent string
+
+	switch event.Event {
+	case plevents.ContainerStopped:
+		dockerEvent = "die"
```
vburenin: Are these events are used somewhere else? Should not they be defined as constants?

lib/apiservers/engine/backends/eventmonitor.go
```diff
@@ -198,19 +198,28 @@ func (m *PortlayerEventMonitor) monitor() error {
...
+	case plevents.ContainerStarted:
+		dockerEvent = "start"
+	case plevents.ContainerCreated:
+		dockerEvent = "create"
```
sflxn: Do we need this in addition to the "create" event we generate in container.go during ContainerCreate?

cgtexmex: nope -- thanks for pointing that out...

lib/apiservers/engine/backends/eventmonitor.go
cgtexmex: I created constants for most of the events and updated where appropriate...

lib/apiservers/engine/backends/eventmonitor.go
```diff
@@ -198,19 +210,26 @@ func (m *PortlayerEventMonitor) monitor() error {
...
 	if vc == nil {
 		log.Errorf("Portlayer event for container %s but not found in cache", event.Ref)
 	}
+	var dockerEvent string
+
+	switch event.Event {
+	case plevents.ContainerStopped:
+		dockerEvent = containerDieEvent
+	case plevents.ContainerStarted:
+		dockerEvent = containerStartEvent
```
sflxn: So, I should have brought this up before.  If we were to follow Docker's pattern, their events are generally the result of a request fullfilled and not the operation completing.  In other words, with the exception of 1 or 2 events (e.g. die), all their events are published in the routers at the end of their related function.  So for start, that would get published at the end of a successful ContainerStart.  Also, not all events will have a corresponding event from the portlayer.  I think we should move this event to ContainerStart.

cgtexmex: I don't really care what docker's pattern is... :-)  When I updated this I had that thought...meaning move to start..anyhoo...I'll do that..

PR #4366, opened 2017-03-23
Fix make check target

caglar10ur:
and also correct the issues pointed by it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,30 +196,26 @@ whitespace:
 	@infra/scripts/whitespace-check.sh
 
 # exit 1 if golint complains about anything other than comments
-golintf = $(GOLINT) $(1) | sh -c "! grep -v 'should have comment'" | sh -c "! grep -v 'comment on exported'" | sh -c "! grep -v 'by other packages, and that stutters'" | sh -c "! grep -v 'error strings should not be capitalized'"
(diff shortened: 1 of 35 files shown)


Makefile
```diff
@@ -196,30 +196,26 @@ whitespace:
...
-	@$(call golintf,github.com/vmware/vic/lib/apiservers/engine/backends/...)
+	@$(call golintf,github.com/vmware/vic/lib/...)
 
 # For use by external tools such as emacs or for example:
 # GOPATH=$(make gopath) go get ...
 gopath:
 	@echo -n $(GOPATH)
 
-$(go-imports): $(GOIMPORTS) $(find . -type f -name '*.go' -not -path "./vendor/*") $(PORTLAYER_DEPS)
+goimports: $(GOIMPORTS) $(find . -type f -name '*.go' -not -path "./vendor/*") $(PORTLAYER_DEPS)
```
hickeng: If you remove the touch then there's little point in having the dependency list - may as well move it to the body of the target so that it doesn't impact the first-pass parse time.

cgtexmex:
LGTM

Makefile
caglar10ur: dropped the dep.

jzt:
LGTM

emlin:
lgtm

PR #4370, opened 2017-03-23
Get image ID using layer ID

jzt:
`docker compose up -d --force-recreate` causes a call to the `GetImage` endpoint using a layer ID as a parameter. This adds a safety check into the image cache to get the image ID from the repo cache just in case we have a layer ID and not an image ID/prefix/name.

This is a temporary workaround for #4378.

### DIFF
--- a/lib/apiservers/engine/backends/cache/image_cache.go
+++ b/lib/apiservers/engine/backends/cache/image_cache.go
@@ -155,6 +155,12 @@ func (ic *ICache) Get(idOrRef string) (*metadata.ImageConfig, error) {
 		return copyImageConfig(config), nil
 	}
 
+	// let's try to get the image ID from the repocache just in case we have a layer ID
+	// see https://github.com/vmware/vic/issues/4378#issuecomment-289070799
+	if id := RepositoryCache().GetImageID(strings.Split(idOrRef, ":")[0]); id != "" {
+		idOrRef = id
+	}
+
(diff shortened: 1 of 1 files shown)


jzt:
ping @cgtexmex for sanity check

chengwang86:
Since we find it in compose test examples, if later all the compose tests work, we don't need to add extra tests

jzt:
Closing as this was being caused by inspect and should be fixed there.

cgtexmex:
Is including LayerID in the VicContainer struct to big of a change?  So pass LayerID to the PL for commit, etc...but in the persona utilize ImageID for look up, etc...

The change you have looks fine...but it continues the abuse of the terms..

jzt:
@cgtexmex that's actually the alternative fix I'm working on. If we want to get something working in the interest of time, this fix might be it (if it won't cause any collision problems because of images sharing layers). I'm currently testing a fix that persists both layer and image IDs in the create path, and extracts them in the inspect path (this is the more correct fix, like you suggested), but if I run into problems, we may want to just use this for now.

cgtexmex:
LGTM

jzt:
correct fix here: https://github.com/vmware/vic/pull/4387  (waiting until later to do a full CI run)

jzt:
♻️

chengwang86:
lgtm

PR #4372, opened 2017-03-23
[specific ci=11-02-Upgrade-Exec] Return error when attempting to run tasks on a container pre 0.9.0

### LINKED ISSUE #4310: Check to ensure that container version supports exec/reload
Broken out from #4110 

We should be returning an error if a task operation is performed on a powered on container where the tether does not support reload.

**Acceptance**
Upgrade test from 0.8 with a live container - container should reject `exec` with a useful error message explaining that the container version does not support exec.

hmahmood:
Fixes #4310

### DIFF
--- a/tests/test-cases/Group11-Upgrade/11-02-Upgrade-Exec.md
+++ b/tests/test-cases/Group11-Upgrade/11-02-Upgrade-Exec.md
@@ -1,0 +1,21 @@
+Test 11-02 - Upgrade Exec
+=======
+
+#Purpose:
+To verify that exec does not work in VIC version 0.9.0 and lower
+
+#Environment:
+This test requires that a vSphere server is running and available
+
+#Test Steps:
+1. Download vic_8351.tar.gz from bintray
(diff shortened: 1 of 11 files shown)


tests/test-cases/Group11-Upgrade/11-02-Upgrade-Exec.md
```diff
@@ -0,0 +1,19 @@
...
+
+#Environment:
+This test requires that a vSphere server is running and available
+
+#Test Steps:
+1. Download vic_8351.tar.gz from bintray
+2. Deploy VIC 8351 to vsphere server
+3. Create a busybox container running the background
+4. Upgrade VCH to latest version
+5. Run docker exec on container created in step (3.)
```
emlin: Should test docker exec works on new containers after upgrade

hmahmood: Added

PR #4376, opened 2017-03-24
Use QueryVirtualDiskInfo in storage [specific ci=Group1-Docker-Commands]

### LINKED ISSUE #4260: Integrate disk method into vsphere storage layer
Once we have the internal method implemented in govmomi (from #4259), we can remove the parentMap.

This issue will encompass all of the changes necessary to integrate the call into the vsphere storage implementation for images.

fdawg4l:
This change removes the parentMap and uses `vdm.QueryVirtualDiskInfo` in its place.  Now vsphere will give us the image parent pointers with which we can derive the tree of images and their relation ships.

Also integrate `object.DatastorePath` where we were currently munging datastore path strings manually.

Fixes #4260


lib/portlayer/storage/vsphere/image.go
```diff
@@ -464,26 +441,34 @@ func (v *ImageStore) GetImage(op trace.Operation, store *url.URL, ID string) (*p
...
-	if parentID != "" {
-		parentURL, _ = util.ImageURL(storeName, parentID)
+	dsk, err := v.dm.Get(op, diskDsURI)
+	if err != nil {
+		return nil, err
+	}
+
+	var parentURL *url.URL
+	if dsk.ParentDatastoreURI != nil {
+		_, vmdk := filepath.Split(dsk.ParentDatastoreURI.Path)
```
dougm: `path.Base` instead of `filepath.Split`.  And in general, always `path` instead of `filepath`, we don't need or want any special Windows treatment that filepath provides.

lib/portlayer/storage/vsphere/image.go
```diff
@@ -464,26 +441,34 @@ func (v *ImageStore) GetImage(op trace.Operation, store *url.URL, ID string) (*p
...
+		parentURL, err = util.ImageURL(storeName, strings.Split(vmdk, ".")[0])
```
dougm: `strings.TrimSuffix(path.Ext(vmdk))` instead of `strings.Split(vmdk, ".")[0]`

fdawg4l: COOL!  Done.  I didn't like this at all.  Thanks!

fdawg4l:
@mhagen-vmware the existing tests should do just fine.  I have to deal with an issue with the unit test compiling now.  Stand by for that before trying to pull this change and attempt to test it.  Thanks!

fdawg4l:
@mhagen-vmware Fixed the unit test issues.  We're good again.  Thanks!

Issue #4375, opened 2017-03-23
Support name resolution for renamed running containers


anchal-agrawal:
**User Statement:**

As a user of VIC, I should be able to rename a running container (`c1`) and use the new name to create a link to or ping from another container (`c2`) without restarting container `c1`.

**Details:**

We need to update the alias and any name resolution metadata for the container during its rename operation. This is a follow-up of #4082, whose scope includes support for name resolution for containers that are restarted after renaming them, and containers that are created, renamed and then started. Renaming a running container *will work* after #4082 is closed, but name resolution for it won't.

**Acceptance Criteria:**

Passing tests for the following use cases:

```
# Container link
$ docker run -dit --name c1 busybox
$ docker rename c1 c11
$ docker run --link c11:c11alias busybox ping -c2 c11alias
...
2 packets transmitted, 2 packets received...

# Ping by name for containers on the same network
$ docker run -dit --name c1 busybox
$ docker rename c1 c11
$ docker run busybox ping -c2 c11
...
2 packets transmitted, 2 packets received...
```

anchal-agrawal:
Need an estimate from @hmahmood - I suspect this will be no less than a 3.

anchal-agrawal:
There's an integration test in `Group1-Docker-Commands/1-36-Docker-Rename` tagged against this issue.

hickeng:
This has been found by @smaliakkal - the necessity for rename in this specific scenario would be eliminated, and come with a performance improvement and complexity reduction, by adding naming convention support for containerVMs. The epic #773 references the naming conventions, both for inventory folders and cVMs, but I didn't find a dedicated issue for the cVM specifically.

renmaosheng:
closing per previous comment, please reopen if you think otherwise.

luwang-vmware:
Update the CI scripts and get PASS https://ci-vic.vmware.com/vmware/vic/20826. Close it now.

PR #4377, opened 2017-03-24
[specific ci=1-36-Docker-Rename] Docker rename

### LINKED ISSUE #4082: Docker API: Container rename
As a vic user, I want to be able to rename my container after it is created.

This requires implementing the `docker rename` operation in the engine-api server and integrate with port layer.

This functionality is also used by `docker-compose up` when there are existing containers for the same service but the configuration or image has been changed, and `docker-compose up –force-recreate` when there are existing containers for the same service even if the configuration or image has not been changed.

Related issues:
#380 #3972  

Acceptance criteria:

- successful operation of cli `docker rename` on a container that is created, running or stopped.
- after we restart a renamed container, network links and aliases should work
- for containers created by a VCH with older version, rename should report warning msg that `rename is not supported`
- successful operation of `docker-compose up` when there are existing containers for the same service but the configuration or image has been changed
- successful operation of `docker-compose up –force-recreate` when there are existing containers for the same service even if the configuration or image has not been changed

Next step (add name resolution for network after renaming a running container):
#4375

chengwang86:
Fixes #4082 

- This PR enables `docker rename`

- Upon calling rename, the vsphere display name will be updated to `newName-containerShortID`; the name in `ExtraConfig.name` and the portlayer containerCache will be updated with the new name (by the CommitHandler); finally, the containerCache in the persona would also be updated with the new name.

- container log bundles from vicadmin will use the latest containerName-containerShortID as the log folder name of the container

- `docker-compose` test cases mentioned in 4082 should work under this PR

- `docker rename` on a running container still has name resolution issue for network connection (see #4375)


lib/apiservers/engine/backends/container.go
```diff
@@ -1607,6 +1601,60 @@ func (c *Container) containerAttach(name string, ca *backend.ContainerAttachConf
...
+
+	if oldName == "" || newName == "" {
+		return fmt.Errorf("neither old nor new names may be empty")
+	}
+
+	if !utils.RestrictedNamePattern.MatchString(newName) {
+		return fmt.Errorf("Invalid container name (%s), only %s are allowed", newName, utils.RestrictedNameChars)
+	}
+
+	var err error
```
vburenin: Why do you need it even here? You can just define it in place where needed.

lib/apiservers/engine/backends/container.go
```diff
@@ -1607,6 +1601,60 @@ func (c *Container) containerAttach(name string, ca *backend.ContainerAttachConf
...
+
+	// Look up the container name in the metadata cache to get long ID
+	vc := cache.ContainerCache().GetContainer(oldName)
+	if vc == nil {
+		log.Errorf("Container %s not found", oldName)
```
vburenin: Since you return error all in almost all cases, sometimes you log it, sometimes don't. Can't caller take care about logging the error?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -870,6 +871,57 @@ func (c *ContainerProxy) AttachStreams(ctx context.Context, vc *viccontainer.Vic
 	return nil
 }
 
+// Rename() calls the portlayer's containerRenamehandler to update the container name in the handle,
+// and then calls CommitHandler to commit the new name to vSphere
+func (c *ContainerProxy) Rename(vc *viccontainer.VicContainer, newName string) error {
+	defer trace.End(trace.Begin(vc.ContainerID))
+
+	var err error
```
vburenin: no need for this.

lib/apiservers/engine/backends/container.go
chengwang86: @vburenin I think the caller is the docker client. It won't log.

lib/apiservers/portlayer/swagger.json
```diff
@@ -1748,6 +1748,63 @@
 				}
 			}
 		},
+		"/containers/{handle}/name": {
+			"put": {
```
vburenin: that has to be post or update, not put.

pkg/version/version.go
```diff
@@ -31,7 +31,7 @@ var (
 	State       string
 
 	// MaxPluginVersion must be increased to add new plugin and make sure the new plugin version is same to this value
-	MaxPluginVersion = 2
+	MaxPluginVersion = 3
```
vburenin: did you change format of any stored data? If not, why did you increase it?

lib/apiservers/portlayer/swagger.json
chengwang86: @vburenin I discussed with Hasan and decided to use put instead of post. What do you think @hmahmood

pkg/version/version.go
chengwang86: @vburenin This is requested by @emlin . currently without `rename` getting in, the dataversion of a container is `2`; so any container with dataveresion < 3 should not suppport `rename`.

lib/apiservers/engine/backends/cache/container_cache.go
```diff
@@ -129,3 +130,25 @@ func (cc *CCache) GetContainerFromExec(eid string) *container.VicContainer {
...
+	cc.m.Lock()
+	defer cc.m.Unlock()
+
+	container := cc.getContainer(oldName)
+	if container == nil {
+		return fmt.Errorf("no such container: %s", oldName)
+	}
+
+	delete(cc.containersByName, container.Name)
+	if exists := cc.getContainer(newName); exists != nil {
```
sflxn: If you check if the new name is available AFTER deleting the container reference from the name list, will this affect being able to lookup by name?

lib/apiservers/engine/backends/container.go
```diff
@@ -1607,6 +1601,62 @@ func (c *Container) containerAttach(name string, ca *backend.ContainerAttachConf
...
+	defer trace.End(trace.Begin(newName))
+
+	if oldName == "" || newName == "" {
+		err := fmt.Errorf("neither old nor new names may be empty")
+		log.Errorf("%s", err.Error())
+		return err
+	}
+
+	if !utils.RestrictedNamePattern.MatchString(newName) {
+		err := fmt.Errorf("Invalid container name (%s), only %s are allowed", newName, utils.RestrictedNameChars)
```
anchal-agrawal: gofmt: error strings should start with lowercase.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -870,6 +871,55 @@ func (c *ContainerProxy) AttachStreams(ctx context.Context, vc *viccontainer.Vic
...
+
+	// Get an API client to the portlayer
+	client := PortLayerClient()
+	if client == nil {
+		err = InternalServerError("wait failed to create a portlayer client")
+		return err
+	}
+
+	// Call the rename functionality in the portlayer.
+	result, err := client.Containers.ContainerRename(containers.NewContainerRenameParamsWithContext(ctx).WithName(newName).WithHandle(handle))
```
anchal-agrawal: Please break this into two lines for readability, if possible.

```go
renameParams := containers.NewContainerRenameParamsWithContext(ctx).WithName(newName).WithHandle(handle)
result, err := client.Containers.ContainerRename(renameParams)
```

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -390,6 +394,35 @@ func (handler *ContainersHandlersImpl) ContainerWaitHandler(params containers.Co
...
+		return containers.NewContainerRenameNotFound()
+	}
+
+	if container.ExecConfig.Name == params.Name {
+		return containers.NewContainerRenameConflict()
+	}
+
+	// rename on container version < supportVersionForRename is not supported
+	if h.DataVersion < supportVersionForRename {
+		return containers.NewContainerRenameInternalServerError().WithPayload(&models.Error{Message: fmt.Sprintf("container %s does not support rename", container.ExecConfig.Name)})
```
anchal-agrawal: Please break this into multiple lines.

```go
err := &models.Error{
    Message: fmt.Sprintf("container %s does not support rename", container.ExecConfig.Name),
}
return containers.NewContainerRenameInternalServerError().WithPayload(err)
```

lib/apiservers/portlayer/swagger.json
vburenin: Oh, I am sorry, I forgot that PUT can also mean update in terms of action. And UPDATE is actually PATCH.
PATCH seems a better fit here because of:
```
GET	Read
POST Create
PUT	Update/Replace
PATCH	Update/Modify
DELETE	Delete
```
We are really renaming, so it is really an update or modification not a replacement that PUT can mean.

pkg/version/version.go
vburenin: I am not sure that I understand. Why can't dataversion 2 can't support rename? Did you change anything in such a way that it is not possible?

chengwang86: @vburenin In the PR I merged yesterday, I already used version=2, and at that time `rename` feature was not added yet. So containers with dataversion=2 does not support rename. Actually rename on such containers does work; but it is Emma's strong suggestion from her upgrade framework's POV.

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -155,6 +155,12 @@ func (ic *ICache) Get(idOrRef string) (*metadata.ImageConfig, error) {
 		return copyImageConfig(config), nil
 	}
 
+	// let's try to get the image ID from the repocache just in case we have a layer ID
+	// see https://github.com/vmware/vic/issues/4378#issuecomment-289070799
+	if id := RepositoryCache().GetImageID(strings.Split(idOrRef, ":")[0]); id != "" {
```
chengwang86: This is from Zac's newly merged PR. I might have done something wrong when I rebase my code base.

lib/apiservers/engine/backends/cache/container_cache.go
```diff
@@ -129,3 +130,27 @@ func (cc *CCache) GetContainerFromExec(eid string) *container.VicContainer {
...
+	cc.m.Lock()
+	defer cc.m.Unlock()
+
+	container := cc.getContainer(oldName)
+	if container == nil {
+		return fmt.Errorf("no such container: %s", oldName)
+	}
+
+	if exists := cc.getContainer(newName); exists != nil {
+		err := fmt.Errorf("conflict in container cache. The name %q is already in use by container %s.", newName, exists.ContainerID)
```
vburenin: Let's make it shorter: The name is already in use.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -870,6 +871,56 @@ func (c *ContainerProxy) AttachStreams(ctx context.Context, vc *viccontainer.Vic
...
+
+	// Call the rename functionality in the portlayer.
+	renameParams := containers.NewContainerRenameParamsWithContext(ctx).WithName(newName).WithHandle(handle)
+	result, err := client.Containers.ContainerRename(renameParams)
+	if err != nil {
+		switch err := err.(type) {
+		case *containers.ContainerRenameNotFound:
+			return NotFoundError(vc.Name)
+		case *containers.ContainerRenameConflict:
+			return derr.NewRequestConflictError(fmt.Errorf("container name collision in portlayer"))
```
vburenin: Just state "container name already exists" and no need to mention port layer. , the word "collision" is not a really good fit here.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -870,6 +871,56 @@ func (c *ContainerProxy) AttachStreams(ctx context.Context, vc *viccontainer.Vic
...
+		default:
+			return InternalServerError(err.Error())
+		}
+	}
+
+	h := result.Payload
+
+	// commit handle
+	_, err = client.Containers.Commit(containers.NewCommitParamsWithContext(ctx).WithHandle(h))
```
vburenin: Did you copy paste this piece of code? It is repeated so frequently that I would simply extract it into a separate methods such as:
```
    CommitHandle(pc *PortLayerClient, handle string) err {
}
```

vburenin: I am not sure this a good fit for this PR however, some work is involved. However, it's a trivial one.

lib/apiservers/portlayer/swagger.json
vburenin: One more thing, docker doesn't implement full REST api, we simply should not follow their idiom here, it would become a mess.

hmahmood: Agree with @vburenin on docker not really using REST. I would however go with PUT; PATCH is not really very widely used.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -870,6 +871,56 @@ func (c *ContainerProxy) AttachStreams(ctx context.Context, vc *viccontainer.Vic
 	return nil
 }
 
+// Rename() calls the portlayer's containerRenamehandler to update the container name in the handle,
+// and then calls CommitHandler to commit the new name to vSphere
+func (c *ContainerProxy) Rename(vc *viccontainer.VicContainer, newName string) error {
```
hmahmood: Can you make sure we return the same http error codes the docker specifies in their documentation? Same goes for `Container.Rename` above.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -40,7 +40,10 @@ import (
 	"github.com/vmware/vic/pkg/version"
 )
 
-const containerWaitTimeout = 3 * time.Minute
+const (
+	containerWaitTimeout    = 3 * time.Minute
+	supportVersionForRename = 3
```
hmahmood: You're going to have to update this to 4 since I stole 3 already (its merged) :-)

lib/apiservers/engine/backends/container_proxy.go
chengwang86: Good point. I will do that.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
chengwang86: Thanks for letting me know.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -390,6 +394,38 @@ func (handler *ContainersHandlersImpl) ContainerWaitHandler(params containers.Co
 	}
 }
 
+func (handler *ContainersHandlersImpl) RenameContainerHandler(params containers.ContainerRenameParams) middleware.Responder {
+	defer trace.End(trace.Begin(fmt.Sprintf("Rename container to %s", params.Name)))
```
emlin: also log out the container old name?

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -390,6 +394,38 @@ func (handler *ContainersHandlersImpl) ContainerWaitHandler(params containers.Co
...
+	}
+
+	// get the indicated container for rename
+	container := exec.Containers.Container(h.ExecConfig.ID)
+	if container == nil {
+		return containers.NewContainerRenameNotFound()
+	}
+
+	if container.ExecConfig.Name == params.Name {
+		return containers.NewContainerRenameConflict()
```
emlin: I think you need another error type for this failure. name conflict usually means the new name is conflict with other containers, but if it's same to itself, usually returns different message.
You can compare it with docker command

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
chengwang86: At this point the container's old name is not passed into the renamecontainerhandler as a parameter

emlin: okay

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
chengwang86: This is a good point. I will use containers.InternalServerError

tests/resources/Docker-Util.robot
```diff
@@ -64,6 +64,23 @@ Get VM display name
...
+    [Arguments]  ${oldname}  ${newname}  ${contID}
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} ps -a
+    Should Be Equal As Integers  ${rc}  0
+    Should Contain  ${output}  ${newname}
+    Should Not Contain  ${output}  ${oldname}
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} inspect -f '{{.Name}}' ${newname}
+    Should Be Equal As Integers  ${rc}  0
+    Should Contain  ${output}  ${newname}
+    ${vmName}=  Get VM display name  ${contID}
+    ${rc}  ${output}=  Run Keyword If  '%{HOST_TYPE}' == 'VC'  Run And Return Rc And Output  govc vm.info %{VCH-NAME}/${vmName}
```
mhagen-vmware: don't think we need this anymore? I think the latest govc handles pathing for vm.info now?

anchal-agrawal:
@mhagen-vmware I'll push the `.md`-related changes for the integration tests as part of a separate PR.

lib/apiservers/engine/backends/container.go
```diff
@@ -1605,6 +1599,62 @@ func (c *Container) containerAttach(name string, ca *backend.ContainerAttachConf
...
+	// Look up the container name in the metadata cache to get long ID
+	vc := cache.ContainerCache().GetContainer(oldName)
+	if vc == nil {
+		log.Errorf("Container %s not found", oldName)
+		return NotFoundError(oldName)
+	}
+
+	oldName = vc.Name
+	if oldName == newName {
+		err := fmt.Errorf("renaming a container with the same name as its current name")
```
sflxn: Is this the same message that you get when using regular docker?

lib/apiservers/engine/backends/container.go
```diff
@@ -1605,6 +1599,62 @@ func (c *Container) containerAttach(name string, ca *backend.ContainerAttachConf
...
+		return derr.NewRequestConflictError(err)
+	}
+
+	if err := c.containerProxy.Rename(vc, newName); err != nil {
+		log.Errorf("Rename error: %s", err)
+		return err
+	}
+
+	// update containerCache
+	if err := cache.ContainerCache().UpdateContainerName(oldName, newName); err != nil {
```
sflxn: You're going to get a race condition here.  If you check above for availability, do the the reconfigure, then update the cache, another operation can come in during that time and take your name "slot".  It's better to "reserve" your name slot.  Perform the reconfigure.  Then update the cache.  To reserve the name slot, the easiest way to do that is to add the container with the new name while keeping the old name in the cache.  Then delete the old name after the reconfigure.

lib/apiservers/engine/backends/container.go
chengwang86: @sflxn Yes, this error msg is from the code of regular docker daemon here https://github.com/docker/docker/blob/master/daemon/rename.go#L38

lib/apiservers/engine/backends/container.go
```diff
@@ -1605,6 +1599,70 @@ func (c *Container) containerAttach(name string, ca *backend.ContainerAttachConf
...
+		log.Errorf("Failed to update container cache: %s", err)
+		cache.ContainerCache().ReleaseName(newName)
+		return err
+	}
+
```
anchal-agrawal: If `rename` succeeds, I suspect we'll have that container by both names at this point - we should `ReleaseName(oldName)` before returning `nil`.

lib/apiservers/engine/backends/container.go
```diff
@@ -1605,6 +1599,70 @@ func (c *Container) containerAttach(name string, ca *backend.ContainerAttachConf
...
+// ContainerRename changes the name of a container, using the oldName
+// to find the container. An error is returned if newName is already
+// reserved.
+func (c *Container) ContainerRename(oldName, newName string) error {
+	defer trace.End(trace.Begin(newName))
+
+	if oldName == "" || newName == "" {
+		err := fmt.Errorf("neither old nor new names may be empty")
+		log.Errorf("%s", err.Error())
+		return err
```
anchal-agrawal: To echo @hmahmood's comment - we should return docker errors with appropriate response codes. See https://docs.docker.com/engine/api/v1.25/#operation/ContainerRename.

Also, I don't think this error is worth logging.

lib/apiservers/engine/backends/container.go
```diff
@@ -1605,6 +1599,70 @@ func (c *Container) containerAttach(name string, ca *backend.ContainerAttachConf
...
+	}
+
+	if !utils.RestrictedNamePattern.MatchString(newName) {
+		err := fmt.Errorf("invalid container name (%s), only %s are allowed", newName, utils.RestrictedNameChars)
+		log.Errorf("%s", err.Error())
```
anchal-agrawal: Ditto - this error isn't worth logging IMO.

lib/apiservers/engine/backends/container.go
```diff
@@ -1605,6 +1599,70 @@ func (c *Container) containerAttach(name string, ca *backend.ContainerAttachConf
...
+	if exists := cache.ContainerCache().GetContainer(newName); exists != nil {
+		err := fmt.Errorf("conflict. The name %q is already in use by container %s. You have to remove (or rename) that container to be able to re use that name.", newName, exists.ContainerID)
+		log.Errorf("%s", err.Error())
+		return derr.NewRequestConflictError(err)
+	}
+
+	// reserve the new name in containerCache
+	if err := cache.ContainerCache().ReserveName(vc, newName); err != nil {
+		log.Errorf("%s", err.Error())
+		return fmt.Errorf("Error when allocating new name: %v", err)
```
anchal-agrawal: This will only happen when a container named `newName` already exists - I think we should return a `derr.NewRequestConflictError` similar to the one above.

lib/apiservers/engine/backends/cache/container_cache.go
```diff
@@ -129,3 +131,51 @@ func (cc *CCache) GetContainerFromExec(eid string) *container.VicContainer {
...
+	delete(cc.containersByName, container.Name)
+
+	container.Name = newName
+	cc.containersByName[newName] = container
+	cc.containersByID[container.ContainerID] = container
+
+	return nil
+}
+
+// reserve the name for container
```
anchal-agrawal: Please expand upon the function comment - we can say
```
// ReserveName is used during a container rename operation to prevent concurrent
// container create operations from grabbing the new name.
```

lib/apiservers/engine/backends/cache/container_cache.go
```diff
@@ -129,3 +131,51 @@ func (cc *CCache) GetContainerFromExec(eid string) *container.VicContainer {
...
+	if _, exist := cc.containersByName[name]; exist {
+		return fmt.Errorf("name is reserved")
+	}
+
+	cc.containersByName[name] = container
+
+	return nil
+}
+
+// release the "reserved" name
```
anchal-agrawal: Ditto re: function comment
```
// ReleaseName is used during a container rename operation to allow concurrent
// container create operations to use the name.
```

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -874,6 +875,56 @@ func (c *ContainerProxy) AttachStreams(ctx context.Context, vc *viccontainer.Vic
 	return nil
 }
 
+// Rename() calls the portlayer's containerRenamehandler to update the container name in the handle,
```
anchal-agrawal: Nit-pick: s/containerRenamehandler/containerRenameHandler

lib/portlayer/exec/handle.go
```diff
@@ -136,6 +136,21 @@ func (h *Handle) Reload() {
 	h.reload = true
 }
 
+// Rename updates the container name in ExecConfig as well as the vSphere display name
```
anchal-agrawal: 👍

lib/apiservers/engine/backends/container.go
chengwang86: This error msg is the same with that from regular docker. Check out their code for reservename

lib/apiservers/engine/backends/container.go
chengwang86: Yes, these are the same with regular docker. Check out their code

lib/apiservers/engine/backends/container.go
chengwang86: I added these logs based on others comments. Let's keep the logs

lib/apiservers/engine/backends/container.go
anchal-agrawal: This is the case that's applicable to us - https://github.com/docker/docker/blob/master/daemon/names.go#L76.

lib/apiservers/engine/backends/container.go
anchal-agrawal: The error message is good - but we should return a docker API error:

```go
err := fmt.Errorf("neither old nor new names may be empty")
return derr.NewErrorWithStatusCode(err, http.StatusInternalServerError)
```

Since Docker only returns 404, 409 or 500 errors for `rename`, I suppose 500 is suitable here.

lib/apiservers/engine/backends/container.go
anchal-agrawal: Sure, that's fine.

lib/apiservers/engine/backends/cache/container_cache.go
chengwang86: thanks. I will apply these comments.

lib/apiservers/engine/backends/container.go
anchal-agrawal: Clarified via Slack that `UpdateContainerName` takes care of this already - my bad.

lib/apiservers/portlayer/swagger.json
fdawg4l: The google API uses PATCH quite liberally (https://cloud.google.com/compute/docs/reference/latest/). I think PATCH is the right thing to do here.

lib/apiservers/engine/backends/container.go
```diff
@@ -1605,6 +1600,70 @@ func (c *Container) containerAttach(name string, ca *backend.ContainerAttachConf
...
+// reserved.
+func (c *Container) ContainerRename(oldName, newName string) error {
+	defer trace.End(trace.Begin(newName))
+
+	if oldName == "" || newName == "" {
+		err := fmt.Errorf("neither old nor new names may be empty")
+		log.Errorf("%s", err.Error())
+		return derr.NewErrorWithStatusCode(err, http.StatusInternalServerError)
+	}
+
```
fdawg4l: Can you add a log.Info line which includes the old name being renamed to the new name?

`log.Infof("Container %s renamed to %s", oldname, newname)`

chengwang86: sure. I will

lib/apiservers/engine/backends/container.go
```diff
@@ -1605,6 +1600,70 @@ func (c *Container) containerAttach(name string, ca *backend.ContainerAttachConf
...
+	oldName = vc.Name
+	if oldName == newName {
+		err := fmt.Errorf("renaming a container with the same name as its current name")
+		log.Errorf("%s", err.Error())
+		return derr.NewErrorWithStatusCode(err, http.StatusInternalServerError)
+	}
+
+	if exists := cache.ContainerCache().GetContainer(newName); exists != nil {
```
fdawg4l: ReserveName looks like it already does this check.  Do we really need to do this here?

chengwang86: @fdawg4l ReserveName only checks the map `containersByName` in containerCache and then reserve that name in the cache. However, `GetContainer` will check all the maps in the cache, although it is really rare that the `newName` could be the same with the `id` of an existing container.

chengwang86: But on a second thought, what you mentioned makes more sense. Even if `newName` is already used as another container's id, it does not matter. So actually this check may not be necessary. Let me see what others think about it.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -874,6 +875,56 @@ func (c *ContainerProxy) AttachStreams(ctx context.Context, vc *viccontainer.Vic
...
+	//retrieve client to portlayer
+	handle, err := c.Handle(vc.ContainerID, vc.Name)
+	if err != nil {
+		return InternalServerError(err.Error())
+	}
+
+	// Get an API client to the portlayer
+	client := PortLayerClient()
+	if client == nil {
+		return InternalServerError("wait failed to create a portlayer client")
```
jzt: `wait failed`? Should this be `rename failed`?

chengwang86: lol. I probably just copied it from containerWaitHandler.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -391,6 +395,41 @@ func (handler *ContainersHandlersImpl) ContainerWaitHandler(params containers.Co
...
+
+	if container.ExecConfig.Name == params.Name {
+		err := &models.Error{
+			Message: fmt.Sprintf("renaming a container with the same name as its current name"),
+		}
+		return containers.NewContainerRenameInternalServerError().WithPayload(err)
+	}
+
+	// rename on container version < supportVersionForRename is not supported
+	log.Infof("The container DataVersion is: %d", h.DataVersion)
```
jzt: Minor: this seems more appropriate for `Debug` log level.

lib/portlayer/exec/handle.go
```diff
@@ -136,6 +136,21 @@ func (h *Handle) Reload() {
...
+func (h *Handle) Rename(newName string) *Handle {
+	defer trace.End(trace.Begin(""))
```
jzt: Would it be useful to trace on `newName`?

pkg/version/version.go
```diff
@@ -31,7 +31,7 @@ var (
 	State       string
 
 	// MaxPluginVersion must be increased to add new plugin and make sure the new plugin version is same to this value
-	MaxPluginVersion = 3
+	MaxPluginVersion = 4
```
jzt: Not sure how we previously missed it, but this file needs 2017 added to copyright header.

lib/apiservers/engine/backends/container.go
vburenin: This is not an error. It's just warning. I also would change the error message to:

"Container name (%s) contains not allowed characters. Only %s characters are allowed."

Since this message sent to the client, phrasing does matter.

lib/apiservers/engine/backends/container.go
vburenin: btw, it's warning, not error.

lib/apiservers/engine/backends/container.go
chengwang86: @vburenin I've got comments re: error msg from many ppl. Sounds like it may be better to leave it the same with the msg on regular docker. Since regular docker also throws an error instead of warning, let's do the same.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -391,6 +395,41 @@ func (handler *ContainersHandlersImpl) ContainerWaitHandler(params containers.Co
...
+
+	// get the indicated container for rename
+	container := exec.Containers.Container(h.ExecConfig.ID)
+	if container == nil {
+		return containers.NewContainerRenameNotFound()
+	}
+
+	if container.ExecConfig.Name == params.Name {
+		err := &models.Error{
+			Message: fmt.Sprintf("renaming a container with the same name as its current name"),
```
vburenin: are you sure you want this error without any parameters?  in this case no need to call Sprintf at all, it is simply a string.

chengwang86: good catch! I should have added the `name` as a parameter

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -874,6 +875,56 @@ func (c *ContainerProxy) AttachStreams(ctx context.Context, vc *viccontainer.Vic
...
+	defer trace.End(trace.Begin(vc.ContainerID))
+
+	//retrieve client to portlayer
+	handle, err := c.Handle(vc.ContainerID, vc.Name)
+	if err != nil {
+		return InternalServerError(err.Error())
+	}
+
+	// Get an API client to the portlayer
+	client := PortLayerClient()
```
caglar10ur: proxy already have a client, you should use it
```
	if c.client == nil {
		return "", InternalServerError("ContainerProxy.CreateContainerTask failed to create a portlayer client")
	}
....
     ...c.client.XYZ...
```

lib/apiservers/engine/backends/cache/container_cache.go
```diff
@@ -129,3 +131,52 @@ func (cc *CCache) GetContainerFromExec(eid string) *container.VicContainer {
...
+	container := cc.getContainer(oldName)
+	if container == nil {
+		return derr.NewRequestNotFoundError(fmt.Errorf("no such container: %s", oldName))
+	}
+
+	delete(cc.containersByName, container.Name)
+
+	container.Name = newName
+	cc.containersByName[newName] = container
+	cc.containersByID[container.ContainerID] = container
```
caglar10ur: you are not updating the exec cache here (containersByExecID), you should

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -40,7 +40,10 @@ import (
 	"github.com/vmware/vic/pkg/version"
 )
 
-const containerWaitTimeout = 3 * time.Minute
+const (
+	containerWaitTimeout    = 3 * time.Minute
+	supportVersionForRename = 4
```
caglar10ur: minVersionForRename

Issue #4380, opened 2017-03-24
auto create user for container


emlin:
As a docker admin, I want to have application running with customized uid:gid. If that user does not exist in container image, the user should be auto created, and group is set as specification. To make sure application access data and remote file system with correct permission.

Acceptance Criteria:
docker run with -u can start container with specified uid:gid, and the user is created if not exist in image.

Open questions: 
1. docker does not support this function. Do we need to do it?
2. do we need to think about remote user integration?

related to #4172

corrieb:
To be very clear, this is nothing to do with #4172. In the case of --user and USER in a Dockerfile, VIC should not be creating users. A pre-existing user is either baked into the image or it's not.

There is a need however to support consistent user profiles across containers in the case where multiple containers are sharing resource, such as an NFS volume mount. There may be other cases where we want to be able to make externally defined user directories available to the container environment. Not sure if that's covered by this.

hickeng:
@matthewavery do we support defining uid/gid options for an nfs volume or volume store?

PR #4385, opened 2017-03-24
[specific ci=Group6-VIC-Machine]support user in docker command and Dockerfile

### LINKED ISSUE #4172: Security: VIC does not respect USER in Dockerfile
Simple Dockerfile:

```
FROM debian:jessie

RUN useradd -d /home/vmware -m -s /bin/bash vmware
USER vmware
```

Run this on regular Docker:

```
vagrant@ubuntu-1604-vmware:~/vic/src/github.com/vmware/dev-release$ docker run -it 10.118.69.100/test/noob-base
vmware@848f6fdf0e5b:/$ 
```

Run this on VIC:

```
vagrant@ubuntu-1604-vmware:~/test/noob$ docker run -it 10.118.69.100/test/noob-base
root@840603cf7882:/# 
```

It is recommendation 4.1 in https://benchmarks.cisecurity.org/tools2/docker/CIS_Docker_1.12.0_Benchmark_v1.0.0.pdf that containers should run as a non-root user. 

There are at least three scenarios where this is potentially problematic:

1) A base image is used containing data that should either not be read or executed by child images. VIC would run this, but with compromised security. 

2) VIC container creates output as root, rather than a specified user, which is then not accessible by other processes/loggers.

3) Container process in VIC overwrites data it shouldn't be able to

What's worse is that the customer may not be aware that we're ignoring this important directive.

emlin:
fixes #4172 

This is to respect "USER" in docker run command line and Dockerfile. The behavior will look like as following:
1. If username or username:groupname is specified, and found in image, container is started through that user.
2. If username or username:groupname is specified, will search username and groupname. Anyone does not exist, docker start will fail with following error message:
```
docker: Error response from daemon: Server error from portlayer: unable to find group notexist: group: unknown group notexist.
```
or 
```
docker: Error response from daemon: Server error from portlayer: unable to find user root: user: lookup username root: no such file or directory.
```
Depends on which image is used.

3. If userid, or userid:groupid is specified, user is set to the userid:groupid, no matter if that user exist or not.
4. Docker inspect will show user settings in Config section.
```
        "Config": {
            "Hostname": "bb946b4dbe52",
            "Domainname": "",
            "User": "2000:2000",
            "AttachStdin": true,
...
```

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -460,7 +460,7 @@ func convertContainerToContainerInfo(container *exec.ContainerInfo) *models.Cont
 		info.ProcessConfig.StartTime = session.StartTime
 		info.ProcessConfig.StopTime = session.StopTime
 
-		if session.User != "" && session.Group != "" {
+		if session.Group != "" {
 			info.ProcessConfig.User = fmt.Sprintf("%s:%s", session.User, session.Group)
 		} else {
 			info.ProcessConfig.User = session.User
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -460,10 +460,9 @@ func convertContainerToContainerInfo(container *exec.ContainerInfo) *models.Cont
 		info.ProcessConfig.StartTime = session.StartTime
 		info.ProcessConfig.StopTime = session.StopTime
 
+		info.ProcessConfig.User = session.User
 		if session.Group != "" {
 			info.ProcessConfig.User = fmt.Sprintf("%s:%s", session.User, session.Group)
-		} else {
-			info.ProcessConfig.User = session.User
 		}
 	} else {
 		// log that sessionID is missing and print the ExecConfig
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -460,6 +460,11 @@ func convertContainerToContainerInfo(container *exec.ContainerInfo) *models.Cont
 		info.ProcessConfig.StartTime = session.StartTime
 		info.ProcessConfig.StopTime = session.StopTime
 
+		if session.User != "" && session.Group != "" {
+			info.ProcessConfig.User = fmt.Sprintf("%s:%s", session.User, session.Group)
+		} else {
+			info.ProcessConfig.User = session.User
+		}
 	} else {
 		// log that sessionID is missing and print the ExecConfig
 		log.Errorf("Session ID is missing from execConfig: %#v", container.ExecConfig)
--- a/lib/apiservers/portlayer/restapi/handlers/task_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/task_handlers.go
@@ -86,15 +86,12 @@ func (handler *TaskHandlersImpl) JoinHandler(params tasks.JoinParams) middleware
 	// parsing user
 	if params.Config.User != "" {
 		parts := strings.Split(params.Config.User, ":")
-		var user, group string
 		if len(parts) > 0 {
-			user = parts[0]
+			sessionConfig.User = parts[0]
 		}
 		if len(parts) > 1 {
-			group = parts[1]
+			sessionConfig.Group = parts[1]
 		}
-		sessionConfig.User = user
-		sessionConfig.Group = group
 	}
 
 	handleprime, err := task.Join(&op, handle, sessionConfig)
--- a/lib/apiservers/portlayer/restapi/handlers/task_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/task_handlers.go
@@ -16,6 +16,7 @@ package handlers
 
 import (
 	"context"
+	"strings"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/go-openapi/runtime/middleware"
@@ -64,6 +65,7 @@ func (handler *TaskHandlersImpl) JoinHandler(params tasks.JoinParams) middleware
 	op.Debugf("Path: %#v", params.Config.Path)
 	op.Debugf("WorkingDir: %#v", params.Config.WorkingDir)
 	op.Debugf("OpenStdin: %#v", params.Config.OpenStdin)
+	op.Debugf("User: %s", params.Config.User)
 
 	sessionConfig := &executor.SessionConfig{
 		Common: executor.Common{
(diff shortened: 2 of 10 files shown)


emlin:
Although the full ci passed for first commit, do find one issue in busybox image, that is the user.Lookup could not find user, even that's root. The error is ```vagrant@vagrant:~/docker$ docker -H 10.160.61.73:2376 --tls run -u root:root -it busybox /bin/sh
docker: Error response from daemon: Server error from portlayer: unable to find user root: user: lookup username root: no such file or directory.```
So changed solution to using docker user package for user specification verification, which will search /etc/passwd and /etc/group file directly.
Test works well and tether binary size is not increased.

lib/tether/tether.go
```diff
@@ -580,10 +584,14 @@ func (t *tether) launch(session *SessionConfig) error {
...
 	}()
 
-	// Special case here because UID/GID lookup need to be done
-	// on the appliance...
-	if len(session.User) > 0 {
-		session.Cmd.SysProcAttr = getUserSysProcAttr(session.User)
+	if len(session.User) > 0 || len(session.Group) > 0 {
+		user, err := getUserSysProcAttr(session.User, session.Group)
+		if err != nil {
+			log.Errorf("user lookup failed %s:%s, %s", session.User, session.Group, err)
```
dougm: Maybe session.Started should be set to this same error message?

lib/apiservers/portlayer/restapi/handlers/task_handlers.go
```diff
@@ -81,6 +83,20 @@ func (handler *TaskHandlersImpl) JoinHandler(params tasks.JoinParams) middleware
...
+	if params.Config.User != "" {
+		parts := strings.Split(params.Config.User, ":")
+		var user, group string
+		if len(parts) > 0 {
+			user = parts[0]
+		}
+		if len(parts) > 1 {
+			group = parts[1]
+		}
+		sessionConfig.User = user
```
vburenin: Are sessionConfig User and Group are empty before this moment? If so, then you don't need those temporary variables.

lib/tether/ops_darwin.go
```diff
@@ -96,24 +95,58 @@ func (t *BaseOperations) Cleanup() error {
 	return nil
 }
 
-// Need to put this here because Windows does not
-// support SysProcAttr.Credential
-func getUserSysProcAttr(uname string) *syscall.SysProcAttr {
```
dougm: Do we need the duplicate code in ops_darwin? If no, let's just return nil like the windows version.  If yes, we could move this function to ops_unix.go and it'll be shared by linux+darwin.

lib/tether/ops_darwin.go
```diff
@@ -96,24 +95,58 @@ func (t *BaseOperations) Cleanup() error {
...
+// Examples of valid user specifications are:
+//     * ""
+//     * "user"
+//     * "uid"
+//     * "user:group"
+//     * "uid:gid
+//     * "user:gid"
+//     * "uid:group"
+func getUserSysProcAttr(uid, gid string) (*syscall.SysProcAttr, error) {
+	if len(uid) == 0 && len(gid) == 0 {
```
vburenin: no need to take length: if uid == "" && gid == ""

lib/tether/ops_darwin.go
```diff
@@ -20,12 +20,11 @@ import (
 	"fmt"
```
vburenin: are changes very different from ops_linux? I would consider keeping platform independent logic in one file

lib/tether/ops_linux.go
```diff
@@ -784,23 +784,58 @@ func (t *BaseOperations) Cleanup() error {
...
 		},
 		Setsid: true,
 	}
+	for _, sgid := range execUser.Sgids {
+		sysProc.Credential.Groups = append(sysProc.Credential.Groups, uint32(sgid))
+	}
+	return sysProc, nil
+}
+
+func searchUser(user string) (*dockerUser.ExecUser, error) {
```
hickeng: This doesn't seem significant enough to warrant splitting out into a separate function if we have to pack/unpack data to return it. This could be inlined or return the Credentials struct directly.

lib/tether/ops_linux.go
```diff
@@ -784,23 +784,58 @@ func (t *BaseOperations) Cleanup() error {
...
+	defaultExecUser := dockerUser.ExecUser{
```
hickeng: This is a completely static construction - consider making it a const

lib/tether/tether.go
emlin: Now the error message of docker command looks like this
```docker: Error response from daemon: Server error from portlayer: Unable to find user nonexist.```
do I need to add more information to it?
For the log itself, feels like a bit more things should be okay

File: lib/tether/ops_darwin.go (hunk 3 above)
Comment (emlin) [inline lib/tether/ops_darwin.go]: I don't know, the original code duplicate the logic in ops_darwin, so I thought they are necessary.
But all other functions in ops_darwin is not implemented, so I guess we still do not tether darwin fully functional, so I'll remove the impl. @hickeng let me know, if you have any concern.

File: lib/tether/tether.go (hunk 1 above)
Comment (dougm) [inline lib/tether/tether.go]: lgtm

File: lib/apiservers/portlayer/restapi/handlers/containers_handlers.go [hunk 8]
```diff
@@ -460,6 +460,11 @@ func convertContainerToContainerInfo(container *exec.ContainerInfo) *models.Cont
 		info.ProcessConfig.StartTime = session.StartTime
 		info.ProcessConfig.StopTime = session.StopTime
 
+		if session.Group != "" {
+			info.ProcessConfig.User = fmt.Sprintf("%s:%s", session.User, session.Group)
+		} else {
+			info.ProcessConfig.User = session.User
```
caglar10ur: [suggestion] this could be
```
info.ProcessConfig.User = session.User
if session.Group != "" {
    info.ProcessConfig.User = fmt.Sprintf("%s:%s", session.User, session.Group)
 } 
			
```

lib/tether/ops_linux.go
```diff
@@ -32,6 +31,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/d2g/dhcp4"
+	dockerUser "github.com/opencontainers/runc/libcontainer/user"
```
caglar10ur: Not the fan of idea having libcontainer deps in the tether. I know you had issues with https://golang.org/pkg/os/user/ so please document them here.

Also why we refer this package as dockerUser? We could drop that and use user.XYZ

sflxn:
Can someone explain to me why we are supporting User in docker file?  Do we currently have a mechanism to create a user and group in the VCH?

sflxn:
LGTM based on technicality

emlin:
@sflxn I would prefer to close this issue, and track the user creation requirement in #4380, as even docker does not support user creation in docker run. The only thing easier for them is that they can easily map /etc/passwd into container with ro permission, to make sure host user present inside of container.
But for enterprise env, no local system file mapping, it will be same with us.

emlin:
> Will there be a robot test w/ user other than root?

filed issue #4394 for integration test. We'll need a custom image, to create a user, and push that image to private repo, then we can test user other than root

PR #4387, opened 2017-03-24
Store image ID in container create path

### LINKED ISSUE #4378: docker inspect container shows incorrect image info
**VIC version:**

0.9

**Deployment details:**

```
bin/vic-machine-linux create --name=vch-1 --target=root:pwd@192.168.60.162 --thumbprint= --no-tlsverify --no-tls
```
**Steps to reproduce:**
```
docker run -d --naem t1 busybox /bin/top
docker inspect t1
```

**Actual behavior:**
```
"Image": "d7f9d170aa714bf287eb994aada10a57f4fb1bf8fd748cbfa36a7ab9549190c1"
```

**Expected behavior:**
This is the output from regular docker
```
"Image": "sha256:00f017a8c2a6e1fe2ffd05c281f27d069d2a99323a8cd514dd35f228ba26d2ff"
```

This needs to be fixed for `docker-compose up`. Currently compose was trying to check the existence of image `d7f9...` from vic but `no such image`.

jzt:
This change is the correct fix for #4378. `ContainerCreate` will now persist the image ID on the backend so that `ContainerInspect` data can be properly constructed, allowing functionality that relies on inspect to function properly.

Fixes #4378

### DIFF
--- a/lib/apiservers/engine/backends/container/viccontainer.go
+++ b/lib/apiservers/engine/backends/container/viccontainer.go
@@ -23,7 +23,8 @@ import (
 // VicContainer is VIC's abridged version of Docker's container object.
 type VicContainer struct {
 	Name        string
-	ImageID     string
+	ImageID     string // maps to the image used by this container
+	LayerID     string // child-most layer ID used to find vmdk for this container
 	ContainerID string
 	Config      *containertypes.Config //Working copy of config (with overrides from container create)
 	HostConfig  *containertypes.HostConfig
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -79,7 +79,7 @@ import (
 
 // VicContainerProxy interface
 type VicContainerProxy interface {
-	CreateContainerHandle(imageID string, config types.ContainerCreateConfig) (string, string, error)
+	CreateContainerHandle(vc *viccontainer.VicContainer, config types.ContainerCreateConfig) (string, string, error)
 	CreateContainerTask(handle string, id string, config types.ContainerCreateConfig) (string, error)
 	AddContainerToScope(handle string, config types.ContainerCreateConfig) (string, error)
 	AddVolumesToContainer(handle string, config types.ContainerCreateConfig) (string, error)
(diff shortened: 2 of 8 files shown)


lib/apiservers/engine/backends/container/viccontainer.go
```diff
@@ -24,6 +24,7 @@ import (
 type VicContainer struct {
 	Name        string
 	ImageID     string
+	LayerID     string // used to find the vmdk to use with this container
```
cgtexmex: Might consider adding a comment for the ImageID..just to be clear...and then for LayerID modify the comment to note that it is the child most ID for vmdk identification

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -169,28 +169,32 @@ func (c *ContainerProxy) Client() *client.PortLayer {
...
-	defer trace.End(trace.Begin(imageID))
+func (c *ContainerProxy) CreateContainerHandle(vc *viccontainer.VicContainer, config types.ContainerCreateConfig) (string, string, error) {
+	defer trace.End(trace.Begin(vc.ImageID))
 
 	if c.client == nil {
 		return "", "", InternalServerError("ContainerProxy.CreateContainerHandle failed to create a portlayer client")
 	}
 
-	if imageID == "" {
+	if vc.ImageID == "" {
```
matthewavery: Please put a comment here stating that this is check has backwards compatibility implications for future VCH upgrades, Since this is a new field in the ExecConfig.

jzt: I actually don't think a backward compatibility issue exists here, since this is the create path (by definition, all containers created with this new code are new containers).

But I do think one exists in the inspect path, so thanks for pointing that out. I'll see what I can come up with (maybe by bugging @emlin).

jzt: Just tested with a v0.8.0 container - inspect works fine, although the problem addressed by this PR still exists. There's nothing we can do about that, though, but it isn't an issue since we didn't support `compose` in v0.8 anyway.

matthewavery: sgtm

vburenin:
LGTM

PR #4391, opened 2017-03-25
WIP for --insecure-registry fix

gigawhitlocks:
WIP because I'm handing this off to @matthewavery for the weekend (thanks dude)

What works:
Pulling from an insecure registry specified with --insecure-registry works
Pulling from an insecure registry w/o --insecure-registry fails correctly
Pulling from a secure registry works

What doesn't quite work:
Pulling an empty or nonexistent image from an insecure registry doesn't give a certificate error and instead reports that the image wasn't found (or a successful pull) -- this shouldn't happen and there should be an SSL error instead. @lcastellano and I believe that this originates from the `LearnRegistryURL` method below doing something it should not.

Also there are no tests.

### DIFF
--- a/lib/imagec/docker.go
+++ b/lib/imagec/docker.go
@@ -1,4 +1,3 @@
-// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -47,7 +46,7 @@ const (
 	DigestSHA256EmptyTar = string(dlayer.DigestSHA256EmptyTar)
 )
 
-// FSLayer is a container struct for BlobSums defined in an image manifest
+// FSLayer is a contanseciner struct for BlobSums defined in an image manifest
 type FSLayer struct {
 	// BlobSum is the tarsum of the referenced filesystem image layer
 	BlobSum string `json:"blobSum"`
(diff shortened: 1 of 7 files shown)


gigawhitlocks:
Looks like there's also a conflict that needs to be resolved :(

lib/imagec/docker.go
```diff
@@ -47,7 +46,7 @@ const (
 	DigestSHA256EmptyTar = string(dlayer.DigestSHA256EmptyTar)
 )
 
-// FSLayer is a container struct for BlobSums defined in an image manifest
```
gigawhitlocks: and this typo snuck in somehow

lib/imagec/imagec_test.go
```diff
@@ -1,4 +1,3 @@
-// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
```
gigawhitlocks: also didnt mean to delete this

lib/imagec/docker.go
```diff
@@ -1,4 +1,3 @@
-// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
```
gigawhitlocks: also didnt mean to delete this

matthewavery:
I want to clarify the issue briefly. If the registry is truly ` insecure` then imagec and the portlayer behave perfectly. It is only the case of self signed certs that should report an SSL error from the insecure registry option.

matthewavery:
After further inspection of this behavior I believe it is a non issue. I will be opening a separate PR that does a little spring cleaning and rectifies the typos that @gigawhitlocks has mentioned above. Then I will rebase. There are some merge conflicts, but it is a formating conflict that I will also resolve/prune. All changes in the new PR will be the same changes here the @lcastellano and @gigawhitlocks have made.

matthewavery:
closing this PR. it is the parent PR to this PR(#4396). #4396 will the be the merged changes.

gigawhitlocks:
Code was merged in #4396

PR #4395, opened 2017-03-25
Add script to tail logfiles from the VCH [skip ci]

fdawg4l:
`./tail.sh <hostname or ip> <docker.log | port-layer.log>` to tail the log file from your dev env.

### DIFF
--- a/infra/scripts/tail.sh
+++ b/infra/scripts/tail.sh
@@ -1,0 +1,53 @@
+#!/bin/bash -e
+# Copyright 2016 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


infra/scripts/tail.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash -e
+# Copyright 2016 VMware, Inc. All Rights Reserved.
```
mdubya66: 2017

fdawg4l: Thanks sophie-bot!

infra/scripts/tail.sh
```diff
@@ -0,0 +1,53 @@
...
+    # extract password from $GOVC_URL
+    password=$(govc env | grep GOVC_PASSWORD= | cut -d= -f 2-)
+fi
+
+if [ -z "$username" ] ; then
+    username=$(govc env | grep GOVC_USERNAME= | cut -d= -f 2-)
+fi
+
+jar=$(mktemp -t cookie-XXXX)
+
```
dougm: could use `trap 'rm $jar' EXIT`, so the temp dir is removed even if commands below fail.

dougm:
Also, bonus points if you make the script shellcheck clean.

hickeng:
You could also add a vic-tail function to the bash-helpers - if there's a client certificate then you can go via vicadmin without doing the ssh dance by providing that to curl.

infra/scripts/vic-logs.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash -e
+# Copyright 2016 VMware, Inc. All Rights Reserved.
```
mhagen-vmware: 2017

PR #4399, opened 2017-03-25
Fix docker login registry url parsing

### LINKED ISSUE #4344: Docker login is broken when the host is specified as <IP-ADDRESS>:<PORT-NUM>
**VIC version:**

0.80.0 / master / e0c5365

** Environment **
Running with Harbor as secure repository

**Steps to reproduce:**

docker login IP-ADDRESS:PORT-NUMBER
or
docker login HOSTNAME:PORT-NUMBER

fails with, (after specifying username/password): 

Error response from daemon: parse 10.118.67.211:80/v2/: first path segment in URL cannot contain colon

**Expected behavior:**

It should succeed

**Logs:**
[logs(1).zip](https://github.com/vmware/vic/files/863114/logs.1.zip)

jzt:
Note: `registry-1.docker.io` will return a 503 if you attempt to tack on `:443` to the end of the address, `https://registry-1.docker.io` works just fine (maybe `url.Parse()` is broken for them, too, and they haven't fixed it yet?). For this reason, we would need to use a separate registry in an integration test for this change.

PS: I know the install package's validator has this small fix too, but I didn't want to import install code into the backend. If anyone thinks this one-liner is worth having an architectural discussion over in order to prevent a 1-line DRY violation, this would be the place to have that discussion. :)

Fixes #4344

### DIFF
--- a/lib/apiservers/engine/backends/system.go
+++ b/lib/apiservers/engine/backends/system.go
@@ -288,6 +288,12 @@ func (s *System) AuthenticateToRegistry(ctx context.Context, authConfig *types.A
 		registryAddress = registryAddress + "/v2/"
 	}
 
+	// TODO(jzt) Ensuring the scheme exists in the url happens at several places in our code.
+	// We should consolidate these into a common method to deal with this.
+	if !strings.HasPrefix(registryAddress, "https://") && !strings.HasPrefix(registryAddress, "http://") {
+		registryAddress = "https://" + registryAddress
+	}
+
 	registryURL, err := url.Parse(registryAddress)
 	if err != nil {
 		msg := fmt.Sprintf("Bad login address: %s", registryAddress)


lib/apiservers/engine/backends/system.go
```diff
@@ -288,6 +288,10 @@ func (s *System) AuthenticateToRegistry(ctx context.Context, authConfig *types.A
 		registryAddress = registryAddress + "/v2/"
 	}
 
+	if !strings.HasPrefix(registryAddress, "https://") && !strings.HasPrefix(registryAddress, "http://") {
+		registryAddress = "https://" + registryAddress
```
sflxn: So what happens if the address is tcp://<addr>?  We'll tack on https:// to it?

sflxn:
After talking to @mhagen-vmware on the slack channel, the Harbor integration into our CI won't be ready to test this fix against in this sprint.  That shouldn't prevent us from merging.

lib/apiservers/engine/backends/system.go
matthewavery: Yeah without writing our own custom parser this is always going to bite us though. I got around this in my nfs-volume-vic-machine work by doing the url.Parse and then checking all the fields I cared about for sane inputs then rebuilding the string again. it looked awful, but the other option is to break out the regex.

jzt: @sflxn is that something you are concerned about? Is that a normal use case?

vburenin: I think that you should look for [letters]:// prefix rather than specific protocols, it should be more transparent for the future.

jzt: If we really want to support any and all protocols, we should capture it in an issue around writing our own custom url parser, which is definitely not within the scope of what we are trying to do right now.

dougm: Can't imagine that we need to support any protocol other than http or https in this case.  However, this is at least the 4th place where tack on a default scheme.  The other 3 that I know of are:
 
* govmomi's soap.ParseURL 

* vic/lib/install/validate/validator.go:ParseURL

* vic/pkg/flags/url_flag.go:Set

For now maybe just add a "TODO: ..." comment, that we could use common method to deal with this.

Issue #4256, opened 2017-03-15
Add Admiral to the Unified OVA


frapposelli:
User Statement: As a user, I want to be able to deploy the management ui with a single unified deployment

Acceptance Criteria: Admiral added in the build of the unified OVA

PR #4401, opened 2017-03-26
Fixed a few initialization problems in Admiral for OVA. #4256

lcastellano:
Also added openjre in preparation of using 'keytool' to add the server cert to the Xenon truststore.

### DIFF
--- a/installer/packer/scripts/admiral/configure_admiral.sh
+++ b/installer/packer/scripts/admiral/configure_admiral.sh
@@ -25,6 +25,7 @@ fi
 
 data_dir=/data/admiral
 conf_dir=/etc/vmware/admiral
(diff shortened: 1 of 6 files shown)


installer/packer/scripts/admiral/configure_admiral.sh
```diff
@@ -25,6 +25,7 @@ fi
 
 data_dir=/data/admiral
 conf_dir=/etc/vmware/admiral
+javadir=/var/opt/OpenJDK-1.8.0.112-bin/bin/
```
dougm: How about: `javadir=/var/opt/OpenJDK-*-bin/bin/`
Then you don't have to change this line when the dep is updated.

PR #4402, opened 2017-03-26
[skip ci] Add markdown files for docker rename tests

anchal-agrawal:
Follow-up of #4377

### DIFF
--- a/tests/test-cases/Group1-Docker-Commands/1-36-Docker-Rename.md
+++ b/tests/test-cases/Group1-Docker-Commands/1-36-Docker-Rename.md
@@ -1,0 +1,79 @@
+Test 1-36 - Docker Rename
+=======
+
+# Purpose:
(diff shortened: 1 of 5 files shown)


chengwang86:
I'm still not quite clear why we need to list all the docker cmds we use in the markdown file, e.g., `issue docker run -dit ...`. Although I totally understand that these are `steps`, it does not help explain why we are taking these steps. 

Actually I prefer the styles used in 11-01 where we only list the high level goals and methods for the test while leaving the details to the robot file.

These are just my comments. I don't require any changes.

chengwang86:
lgtm

anchal-agrawal:
@chengwang86 I concur with your point. We've been using this pattern extensively so far - see `1-04-Docker-Create.md` for example. I'll let @mhagen-vmware and @rajanashok make the call on whether we should move to your suggested approach.

rajanashok:
lgtm

Issue #4171, opened 2017-03-08
container output handler doesn't release output stream


cgtexmex:
The portLayer makes use of a custom handler to facilitate streaming to the client.  This [ContainerOutputHandler's](https://github.com/vmware/vic/blob/master/lib/apiservers/portlayer/restapi/handlers/interaction_handlers.go#L488) `WriteResponse` func will never release the output stream.  Even if the client closes the transport. This will eventually cause issues for the portLayer API.

The [EventHandler](https://github.com/vmware/vic/blob/master/lib/apiservers/portlayer/restapi/handlers/event_handlers.go#L114) makes use of a similar custom handler that will release the output stream when the client closes the transport.  

I suggest we refactor this bit of code to create a reusable streaming handler that that will be available to any portLayer handler that needs to stream data.  At a minimum the ContainerOutputHandler should be updated to release the stream.

mhagen-vmware:
Please prioritize and move to backlog when ready.

cgtexmex:
I will most likely fix this as part of my work on stats #3964 in sprint 4

cgtexmex:
fixed by #4413

PR #4413, opened 2017-03-27
docker cpu and memory stats implemented

### LINKED ISSUE #3965: PortLayer CPU/Mem container statistics
Provide the ability to access containerVM CPU & Memory statistics via the PortLayer API.

cgtexmex:
Initial implementation of docker stats.  This commit will
provide CPU & Memory stats -- disk / network will be addressed
in the future.

Additional refactored the custom container output handler to
be a generic streams handler.

Fixes #3965, #4171

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -1336,7 +1336,38 @@ func (c *Container) ContainerLogs(ctx context.Context, name string, config *back
 // ContainerStats writes information about the container to the stream
 // given in the config object.
 func (c *Container) ContainerStats(ctx context.Context, name string, config *backend.ContainerStatsConfig) error {
-	return fmt.Errorf("%s does not yet implement ContainerStats", ProductName())
+	defer trace.End(trace.Begin(name))
+
+	// Look up the container name in the metadata cache to get long ID
+	vc := cache.ContainerCache().GetContainer(name)
+	if vc == nil {
+		return NotFoundError(name)
+	}
+
+	// get the configured CPUMhz for this VCH so that we can calculate docker CPU stats
+	cpuMhz, err := systemBackend.SystemCPUMhzLimit()
+	if err != nil {
+		// wrap error to provide a bit more detail
+		sysErr := fmt.Errorf("unable to gather system CPUMhz for container(%s): %s", vc.ContainerID, err)
+		log.Error(sysErr)
+		return InternalServerError(sysErr.Error())
+	}
+
+	out := config.OutStream
+	if config.Stream {
+		// Outstream modification (from Docker's code) so the stream is streamed with the
+		// necessary headers that the CLI expects.  This is Docker's scheme.
+		wf := ioutils.NewWriteFlusher(config.OutStream)
+		defer wf.Close()
+		wf.Flush()
+		out = io.Writer(wf)
+	}
+
+	err = c.containerProxy.StreamContainerStats(ctx, vc.ContainerID, out, config.Stream, cpuMhz, vc.HostConfig.Memory)
+	if err != nil {
+		log.Errorf("error while streaming container (%s) stats: %s", vc.ContainerID, err)
+	}
+	return nil
 }
 
 // ContainerTop lists the processes running inside of the given
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -63,6 +63,7 @@ import (
 
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	viccontainer "github.com/vmware/vic/lib/apiservers/engine/backends/container"
+	"github.com/vmware/vic/lib/apiservers/engine/backends/convert"
 	epoint "github.com/vmware/vic/lib/apiservers/engine/backends/endpoint"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/containers"
@@ -87,6 +88,7 @@ type VicContainerProxy interface {
 	AddInteractionToContainer(handle string, config types.ContainerCreateConfig) (string, error)
 	CommitContainerHandle(handle, containerID string, waitTime int32) error
 	StreamContainerLogs(name string, out io.Writer, started chan struct{}, showTimestamps bool, followLogs bool, since int64, tailLines int64) error
+	StreamContainerStats(ctx context.Context, id string, out io.Writer, stream bool, CPUMhz int64, memory int64) error
 
 	Stop(vc *viccontainer.VicContainer, name string, seconds *int, unbound bool) error
 	State(vc *viccontainer.VicContainer) (*types.ContainerState, error)
@@ -506,6 +508,65 @@ func (c *ContainerProxy) StreamContainerLogs(name string, out io.Writer, started
 		}
 	}
 
+	return nil
+}
+
+// StreamContainerStats will provide a stream of container stats written to the provided
+// io.Writer.  Prior to writing to the provided io.Writer there will be a transformation
+// from the portLayer representation of stats to the docker format
+func (c *ContainerProxy) StreamContainerStats(ctx context.Context, id string, out io.Writer, stream bool, CPUMhz int64, mem int64) error {
+	defer trace.End(trace.Begin(id))
+
+	plClient, transport := c.createNewAttachClientWithTimeouts(attachConnectTimeout, 0, attachAttemptTimeout)
+	defer transport.Close()
+
+	// create a child context that we control
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	params := containers.NewGetContainerStatsParamsWithContext(ctx)
+	params.ID = id
+	params.Stream = stream
+
+	// converter config
+	config := convert.ContainerStatsConfig{
+		Ctx:         ctx,
+		Cancel:      cancel,
+		VchMhz:      CPUMhz,
+		Stream:      stream,
+		ContainerID: id,
+		Out:         out,
+		Memory:      mem,
+	}
+	// create our converter
+	containerConverter := convert.NewContainerStats(config)
+	// provide the writer for the portLayer and start listening for metrics
+	writer := containerConverter.Listen()
+	if writer == nil {
+		// problem with the listener
+		return InternalServerError(fmt.Sprintf("unable to gather container(%s) statistics", id))
+	}
+
+	_, err := plClient.Containers.GetContainerStats(params, writer)
+	if err != nil {
+		switch err := err.(type) {
+		case *containers.GetContainerStatsNotFound:
+			return NotFoundError(id)
+		case *containers.GetContainerStatsInternalServerError:
+			return InternalServerError("Server error from the interaction port layer")
+		default:
+			if ctx.Err() == context.Canceled {
+				return nil
+			}
+			//Check for EOF.  Since the connection, transport, and data handling are
+			//encapsulated inside of Swagger, we can only detect EOF by checking the
+			//error string
+			if strings.Contains(err.Error(), swaggerSubstringEOF) {
+				return nil
+			}
+			return InternalServerError(fmt.Sprintf("Unknown error from the interaction port layer: %s", err))
+		}
+	}
 	return nil
 }
 
--- a/lib/apiservers/engine/backends/convert/stats.go
+++ b/lib/apiservers/engine/backends/convert/stats.go
@@ -1,0 +1,287 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package convert
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/docker/docker/api/types"
+
+	"github.com/vmware/vic/lib/portlayer/metrics"
+)
+
+// ContainerStats encapsulates the conversion of VMMetrics to
+// docker specific metrics
+type ContainerStats struct {
+	config         ContainerStatsConfig
+	dockerStats    *types.StatsJSON
+	currentMetrics *metrics.VMMetrics
+	totalVCHMhz    uint64
+	dblVCHMhz      uint64
+	preTotalMhz    uint64
+
+	// reader/writer for stream
+	reader *io.PipeReader
+	writer *io.PipeWriter
+}
+
+type ContainerStatsConfig struct {
+	Ctx         context.Context
+	Cancel      context.CancelFunc
+	Out         io.Writer
+	ContainerID string
+	Memory      int64
+	Stream      bool
+	VchMhz      int64
+}
+
+type InvalidOrderError struct {
+	current  time.Time
+	previous time.Time
+}
+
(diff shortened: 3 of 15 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -434,6 +434,7 @@ func (c *Container) ContainerExtractToDir(name, path string, noOverwriteDirNonDi
 // ContainerStatPath stats the filesystem resource at the specified path in the
 // container identified by the given name.
 func (c *Container) ContainerStatPath(name string, path string) (stat *types.ContainerPathStat, err error) {
+
```
fdawg4l: Random new line.

cgtexmex: oops..meant to add comment: left blank intentionally

:-)

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -0,0 +1,265 @@
...
+	finished := make(chan struct{})
+
+	var vmm metrics.VMMetrics
+	var previousStat *types.StatsJSON
+
+	go func() {
+		for {
+			select {
+			case <-cs.config.Ctx.Done():
+				close(finished)
```
fdawg4l: You could add `fallthrough` here, I think.

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -0,0 +1,265 @@
...
+type ContainerStatsConfig struct {
+	Ctx         context.Context
+	Cancel      context.CancelFunc
+	Out         io.Writer
+	ContainerID string
+	Stream      bool
+	VchMhz      int64
+}
+
+type InvalidOrder struct {
```
anchal-agrawal: We should rename `InvalidOrder` to `InvalidOrderError` since it's an error type.

lib/portlayer/metrics/container.go
```diff
@@ -0,0 +1,332 @@
...
+		defer currentSubs[mos[0]].publisher.Close()
+	}
+
+	// get the sample..
+	sample, err := cc.perfMgr.SampleByName(ctx, spec, names, mos)
+	if err != nil {
+		log.Errorf("unable to get metric sample: %s", err)
+		return
+	}
+	// convert to metrtics
```
anchal-agrawal: s/metrtics/metrics

lib/portlayer/metrics/container.go
```diff
@@ -0,0 +1,332 @@
...
+		perfMgr: performance.NewManager(session.Vim25()),
+	}
+
+	// kick off the subscription sampler
+	go cc.sampler()
+
+	return cc
+}
+
+// sample will check for subscriptions and sample when needed
```
anchal-agrawal: s/sample/sampler?

lib/apiservers/engine/backends/container.go
```diff
@@ -1336,7 +1336,36 @@ func (c *Container) ContainerLogs(ctx context.Context, name string, config *back
...
+		return InternalServerError("unable to gather container statistics")
+	}
+
+	out := config.OutStream
+	if config.Stream {
+		// Outstream modification (from Docker's code) so the stream is streamed with the
+		// necessary headers that the CLI expects.  This is Docker's scheme.
+		wf := ioutils.NewWriteFlusher(config.OutStream)
+		defer wf.Close()
+		wf.Flush()
```
dougm: Is there something that needs to be flushed at this point?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -509,6 +511,64 @@ func (c *ContainerProxy) StreamContainerLogs(name string, out io.Writer, started
...
+		case *containers.GetContainerStatsInternalServerError:
+			return InternalServerError("Server error from the interaction port layer")
+		default:
+			if ctx.Err() == context.Canceled {
+				return nil
+			}
+			//Check for EOF.  Since the connection, transport, and data handling are
+			//encapsulated inside of Swagger, we can only detect EOF by checking the
+			//error string
+			if strings.Contains(err.Error(), swaggerSubstringEOF) {
```
dougm: We can't use `err == io.EOF` in this case?

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -0,0 +1,266 @@
...
+			case <-finished:
+				close(metric)
+				cs.Stop()
+				return
+			default:
+				for dec.More() {
+					err := dec.Decode(&vmm)
+					if err != nil {
+						log.Errorf("container metric decoding error for container(%s): %s", cs.config.ContainerID, err)
+						cs.config.Cancel()
```
dougm: don't we want to avoid sending `metric <- vmm` below in the case or error?

lib/portlayer/metrics/container.go
```diff
@@ -0,0 +1,327 @@
...
+func (cc *collectorVM) Sample(contain interface{}) (chan interface{}, error) {
+	defer trace.End(trace.Begin(""))
+
+	containerSub, err := containerSubscriber(contain)
+	if err != nil {
+		return nil, err
+	}
+
+	// create a manager for this request
+	mgr := performance.NewManager(cc.session.Vim25())
```
dougm: It would be ideal if we could create just 1 `performance.Manager` instance, to use the same PerfCounterInfo cache.

lib/apiservers/engine/backends/container.go
cgtexmex: Not sure..was just following the same thing done in ContainerLogs -- seems to work fine.

lib/apiservers/engine/backends/container_proxy.go
cgtexmex: Not sure -- again just following other patterns of interacting with a stream and swagger...

lib/portlayer/metrics/container.go
cgtexmex: This specific case is for a non-streaming request -- instead of waiting for the next instance of the subscription perf Manager to handle the request I create a new one so that the user doesn't have to wait up to 20 seconds to get a reply.  I originally had one performance manager that handled all of the requests -- but because of the 20s interval I opted against it.

Additionally, I was going to create an issue stating that we may eventually need a pool of managers as the documentation states that for best performance you should limit the manager to 50 objects.  So from my perspective I don't think we can get to one.

I think for now this works and we can open appropriate issues for a different direction if needed.

lib/apiservers/engine/backends/convert/stats.go
cgtexmex: It shouldn't matter -- it would go to the conversion stage and wouldn't be able to be converted to a docker stat and nothing would happen.  Plus the context cancel would probably catch it before it went to far.  

If you feel strongly that it should be a return after that I can do that...

lib/portlayer/metrics/container.go
dougm: The `performance.NewManager` call returns reference to the singleton on the VC/ESX side, there is never more than one instance on the ESX/VC side.  On the Go side, other than that reference, the performance.Manager has the cache of PerfCounterInfo we use to make metric name to id, etc.  I can see how we might want to limit the `QueryPerf` method call to N objects, but there's no reason we can't reuse the `performance.Manager` singleton to get that cache win.  But yes, functionally is fine to call `performance.NewManager` more than once.

lib/portlayer/metrics/container.go
```diff
@@ -0,0 +1,327 @@
...
+				Interval:   sampleInterval,
+			}
+			// the series will have values for each sample
+			for _, v := range met.Value {
+				switch v.Name {
+				case "cpu.usagemhz.average":
+					// convert to int so we skip the aggregate measurement
+					iid, err := strconv.Atoi(v.Instance)
+					if err != nil {
+						break
```
dougm: If you want to skip the aggregate measurement, should check `if v.Instance == "" { continue }`.  In any case, seems you'd want to continue here rather than break.

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -0,0 +1,266 @@
...
+		dockerCPU.CPUUsage.PercpuUsage[ci] = uint64(current.MhzUsage)
+		dockerCPU.CPUUsage.TotalUsage += uint64(current.MhzUsage)
+	}
+	// The first stat available for a VM will be missing detail
+	if cpuCount > 0 {
+		dockerCPU.CPUUsage.TotalUsage /= uint64(cpuCount)
+	}
+
+	// Current SystemUsage will always be double the VCH
+	dockerCPU.SystemUsage = cs.dblVCHMhz
```
hickeng: Why is this the case?

lib/portlayer/metrics/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+	Unsubscribe(interface{}, chan interface{})
+	// SubscriberCount returns the number of subscribers for the collector
+	SubscriberCount() int
+	// Sample will return metrics without a subscription
+	Sample(interface{}) (chan interface{}, error)
+}
+
+func Init(ctx context.Context, session *session.Session) error {
+	defer trace.End(trace.Begin(""))
+	initializer.once.Do(func() {
```
dougm: Looks like this is another existing pattern.. just curious why we'd be using sync.Once in these Init functions, especially if there's an error.. these Init bodies can never be called again.

This function would be a good place to call `performance.NewManager`

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -0,0 +1,266 @@
...
+	// PreRead will be zero if we don't have two samples
+	if cs.dockerStats.PreRead.IsZero() {
+		return nil, nil
+	}
+	return cs.dockerStats, nil
+}
+
+func (cs *ContainerStats) memory() {
+	// given KB (i.e. 2048) convert to GB
+	cs.dockerStats.MemoryStats.Limit = uint64(cs.config.Memory * 1000 * 1000)
```
hickeng: given KB, convert to GB -> multiplication
given KB, convert to Bytes -> also multiplication?

Seems like one of these should be a division, or one comment is incorrect.

Also, given this is memory not disk are we sure that it's 1000 and not 1024? Just a confirmation - it's always ambiguous for me these days.

lib/apiservers/engine/backends/container.go
```diff
@@ -1336,7 +1336,36 @@ func (c *Container) ContainerLogs(ctx context.Context, name string, config *back
...
+	defer trace.End(trace.Begin(name))
+
+	// Look up the container name in the metadata cache to get long ID
+	vc := cache.ContainerCache().GetContainer(name)
+	if vc == nil {
+		return NotFoundError(name)
+	}
+
+	// get the configured CPUMhz for this VCH so that we can calculate docker CPU stats
+	cpuMhz, err := systemBackend.SystemCPUMhzLimit()
```
hickeng: I don't recall what we were wanting to report here - was it the percentage consumed of the entire VCH (resource pool) limit or the percentage of the limit imposed on that containerVM (nominal vCPU MHz * number vCPU / ...)

lib/portlayer/metrics/container.go
```diff
@@ -0,0 +1,327 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
```
hickeng: I think this is likely to break down as the different portlayer components providing metrics related to their area of concern, rather than a separate metrics package handling all of it.
For example, consider the case where NSX manager has the network usage data, not VC.

Don't change this for now, but please be aware we'll likely have to shuffle in the short/medium future.

lib/apiservers/engine/backends/container.go
```diff
@@ -1336,7 +1336,36 @@ func (c *Container) ContainerLogs(ctx context.Context, name string, config *back
...
+	if err != nil {
+		log.Errorf("unable to gather systemCPUMhz for container(%s): %s", vc.ContainerID, err)
+		return InternalServerError("unable to gather container statistics")
```
chengwang86: should we also return the actual `err` here?

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -0,0 +1,266 @@
...
+
+// Stop will clean up remaining conversion resources
+func (cs *ContainerStats) Stop() {
+	if cs.reader != nil && cs.writer != nil {
+		cs.reader.Close()
+		cs.writer.Close()
+	}
+}
+
+// Listen will listen for new metrics from the portLayer, covert to docker format
```
sgairo: "convert"?

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -0,0 +1,266 @@
...
+// and encode to the configured Writer.  The returned PipeWriter is the source of
+// the vic metrics that will be transformed to docker stats
+func (cs *ContainerStats) Listen() *io.PipeWriter {
+	// TODO: could split decode / encode into separte funcs -- would provide for easier
```
sgairo: "separate"

lib/apiservers/engine/backends/container.go
cgtexmex: We are producing the % consumed of the entire VCH

lib/portlayer/metrics/container.go
cgtexmex: Agree to an extent..also from my perspective the vmomiGW would be ideal place for this type of infra interaction.

lib/apiservers/engine/backends/container.go
chengwang86: I think Admiral just needs everything to be in the same format as what regular docker reports. Regular docker only reports CPU percentage.

lib/apiservers/engine/backends/convert/stats.go
cgtexmex: The docker client has a formula that it uses to calculate the CPU usage for a given interval.  So we need to return the data so that the output of the formula makes sense for vSphere. This is something that needs to be documented -- I've started to do that, but don't have it complete.  

If you'd like more detail now let me know.

cgtexmex: FWIW: I'm adding some additional comments around these items in currentCPU & previousCPU

lib/portlayer/metrics/container.go
cgtexmex: The ID for the processor in the CPUUsage struct is an int, so I need an int there, thus the conversion as a check.

lib/apiservers/engine/backends/convert/stats.go
cgtexmex: For memory limit we are given MB not KB -- comment changed

For active memory we are given KB, so multiplier changed to 1024

lib/portlayer/metrics/container.go
cgtexmex: I've changed the code to utilize only one perfManager

lib/portlayer/metrics/metrics.go
cgtexmex: `performance.NewManager` is called in this section as part of the vmCollector creation

lib/portlayer/metrics/container.go
cgtexmex: I've updated this to check specifically for the aggregate instance and if not then convert.  If the conversion fails (highly doubtful) it will log the error and not publish the broken metric.

Additional comments added to provide background / clarity as well.

lib/apiservers/engine/backends/container.go
cgtexmex: @chengwang86 this will be in the same format that can be consumed by admiral, docker client, etc..

lib/portlayer/metrics/metrics.go
dougm: Still see 2 calls to performance.NewManager in the code..

cgtexmex: Should only be one now...I might have commented as I was still testing and hadn't pushed into branch...forgive me

lib/apiservers/engine/backends/convert/stats.go
cgtexmex: So this had to be changed back -- the correct multiplier for active memory is 1000.  If we multiply by 1024 the active memory is incorrect as it's inflated.

cgtexmex: 3rd and final update on this...1000 didn't make any sense to me as it should be 1024, but the docker client was over reporting.  It's apparently and issue with the docker 1.11.x client and works as expected in 1.13 -- so both values are now multiplied by 1024

PR #4466, opened 2017-03-29
Make UI installer work with vic-ui-plugins in Makefile

### LINKED ISSUE #4465: H5 Client UI installer script needs to be refactored
**User Statement:**

As the UI build steps are moved into Makefile, and `installer/scripts/build-bundle.sh` is ready, I want to make sure the UI build scripts are in sync with some of the changes #4438 made. For example, the vSphere Client plugin manifest file needs to be refactored to accommodate what `build-bundle.sh` is doing.

**Acceptance Criteria:**

UI installer script for H5 Client (ui/installer/HTML5Client) should work the same as before.

jooskim:
This PR doesn't change the existing functionality, but simply refactors it for later improvement: Currently we have `VCSA`, `HTML5Client` and `vCenterForWindows` under `ui/installer` but eventually there would only be VCSA and vCenterForWindows, as HTML5Client itself is not a platform and is a bit misleading.

Fixes #4465

### DIFF
--- a/installer/scripts/build-bundle.sh
+++ b/installer/scripts/build-bundle.sh
@@ -74,9 +74,8 @@ fi
 if [ -d "$bin_dir/ui/vCenterForWindows" ]; then
     cp -r $bin_dir/ui/vCenterForWindows $TEMP_DIR/ui
 fi
-if [ -d "$bin_dir/ui/plugin-packages" ]; then
-    find "$bin_dir/ui/plugin-packages" -name "vc_extension_flags" -exec cp {} $TEMP_DIR/ui \;
-    mv $TEMP_DIR/ui/vc_extension_flags $TEMP_DIR/ui/plugin-manifest
+if [ -f "$bin_dir/ui/plugin-manifest" ]; then
+    cp $bin_dir/ui/plugin-manifest $TEMP_DIR/ui/
 fi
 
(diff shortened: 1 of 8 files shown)


installer/scripts/build-bundle.sh
```diff
@@ -74,9 +74,8 @@ fi
 if [ -d "$bin_dir/ui/vCenterForWindows" ]; then
     cp -r $bin_dir/ui/vCenterForWindows $TEMP_DIR/ui
 fi
-if [ -d "$bin_dir/ui/plugin-packages" ]; then
-    find "$bin_dir/ui/plugin-packages" -name "vc_extension_flags" -exec cp {} $TEMP_DIR/ui \;
-    mv $TEMP_DIR/ui/vc_extension_flags $TEMP_DIR/ui/plugin-manifest
+if [ -f "$bin_dir/ui/plugin-manifest" ]; then
+    cp $bin_dir/ui/plugin-manifest $TEMP_DIR/ui/
```
andrewtchin: So is there only one plugin-manifest?

jooskim: Yes - When the Flex plugin is built it will generate the manifest and then the H5C plugin build process will append H5C plugin key to that file.

jooskim:
thanks @andrewtchin and @frapposelli

PR #4472, opened 2017-03-30
reduce the verbosity of vspc

kreamyx:
vspc becomes verbose only if the vch config debug level is greater than 2

### DIFF
--- a/lib/vspc/vspc.go
+++ b/lib/vspc/vspc.go
@@ -23,9 +23,11 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 
+	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/lib/portlayer/constants"
 	"github.com/vmware/vic/pkg/telnet"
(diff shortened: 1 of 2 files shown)


caglar10ur:
ping @kreamyx, I need this in my life :)

kreamyx:
@caglar10ur haha it is ready! just waiting for 'Verify'

mdubya66:
@mhagen-vmware I'd like to consider this for RC3.

mhagen-vmware:
@mdubya66   This is just a nice to have reduction in logging verbosity right?  I wouldn't think that would be a candidate for an RC build...

mdubya66:
how verbose is the question? is it causing the log to rotate where we lose the ability to debug?

kreamyx:
@mdubya66 it really depends on the serial data moving back and forth. It can get pretty ugly if there is a lot of data flowing. Moreover, the verbose logging will only be needed if debugging a problem caused by the vspc. Otherwise, it is not really relevant for normal debugging purposes.

Issue #4394, opened 2017-03-25
integration test for docker user


emlin:
For #4172, we have docker user and user in Dockerfile support, we need to have integration test covering this scenario, to make sure no regression happens in the future.

Test scenarios:
covering following user format:
- nonexist
- nonexist:nonexist
- root:nonexist
- root:root
- newuser:newgroup (created in image)
- 2000
- 2000:2000
- 0:0

Should test the above user format in both docker command and Dockerfile
The first three format should have correct error message, similar to docker errors

PR #4476, opened 2017-03-30
[specific-ci Group1-Docker-Commands/1-37-Docker-USER] Create integration test for USER directive in Dockerfiles

gigawhitlocks:
Addressing #4394 , this PR contains 3 dockerfiles (which I've published as images to DockerHub but the dockerfiles are checked into `resources/` for future reference) and 3 tests to make sure that `USER` works as expected

Emma laid out a few other test cases, but a few of them do not make sense or are redundant:

- nonexist
- nonexist:nonexist
- root:nonexist

None of these are possible, you cannot be a named user that doesn't exist on a Linux system as far as I am aware. Testing confirmed this hypothesis in vanilla Docker. 

- root:root

Redundant since this is the default if USER isn't specified; normal `Docker Run` tests basically test this

- 0:0

Root is uid 0, see above

- newuser:newgroup
- 2000
- 2000:2000

Tests for these are present in this change

### DIFF
--- a/tests/resources/dockerfiles/Dockerfile.1-37-Docker-USER-UID:GID-2000:2000
+++ b/tests/resources/dockerfiles/Dockerfile.1-37-Docker-USER-UID:GID-2000:2000
@@ -1,0 +1,3 @@
+from busybox:latest
+user 2000:2000
+cmd id
--- a/tests/test-cases/Group1-Docker-Commands/1-37-Docker-USER.robot
+++ b/tests/test-cases/Group1-Docker-Commands/1-37-Docker-USER.robot
@@ -1,0 +1,36 @@
+# Copyright 2016-2017 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#	http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License
+
+*** Settings ***
+Documentation   Test 1-37 - Docker Run As USER
+Resource        ../../resources/Util.robot
+Suite Setup     Install VIC Appliance To Test Server
+Suite Teardown  Cleanup VIC Appliance On Test Server
+
+*** Test Cases ***
+Run as NewUser in NewGroup
+     ${rc}    ${output}=    Run And Return Rc And Output    docker %{VCH-PARAMS} run gigawhitlocks/1-37-docker-user-newuser-newgroup:latest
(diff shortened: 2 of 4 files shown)


tests/resources/dockerfiles/Dockerfile.1-37-Docker-USER-UID:GID-2000:2000
```diff
@@ -0,0 +1,3 @@
+from busybox:latest
+user 2000:2000
```
emlin, line 2000: Please add another image with not exist user specified in the USER command, for example, notexist, notexist:notexist.
To test the not existed user will cause error during docker run.

tests/test-cases/Group1-Docker-Commands/1-37-Docker-USER.robot
```diff
@@ -0,0 +1,36 @@
...
+Run as UID 2000
+     ${rc}    ${output}=    Run And Return Rc And Output    docker %{VCH-PARAMS} run gigawhitlocks/1-37-docker-user-uid-2000:latest
+     Should Be Equal As Integers    ${rc}       0
+     Should Contain                 ${output}   uid=2000 gid=0(root)
+
+Run as UID:GID 2000:2000
+     ${rc}    ${output}=    Run And Return Rc And Output    docker %{VCH-PARAMS} run gigawhitlocks/1-37-docker-user-uid-gid-2000-2000:latest
+     Should Be Equal As Integers    ${rc}       0
+     Should Contain                 ${output}   uid=2000 gid=2000
+
```
emlin: this command also support docker run -u, and all different combination works in the -u option as well. It would be better to have that covered as well.

tests/test-cases/Group1-Docker-Commands/1-37-Docker-USER.robot
```diff
@@ -0,0 +1,36 @@
+# Copyright 2016-2017 VMware, Inc. All Rights Reserved.
```
mhagen-vmware: only 2017

emlin:
> nonexist
> nonexist:nonexist
> root:nonexist

Ian, I think this is to test the sad path, to make sure our code works as expected for these wrong scenarios

tests/test-cases/Group1-Docker-Commands/1-37-Docker-USER.robot
mhagen-vmware: .md file also please

emlin:
btw, Mikkel ever mentioned the nightly test env, which will have private registry setup, then we can have the docker images pushed to there during test.
Not sure if we still want that part. @mhagen-vmware

gigawhitlocks:
I see, didn't realize you wanted those to be negative tests @emlin thank you for the clarification. I will add those as well.

gigawhitlocks:
and @mhagen-vmware will write the .md file after tests since it does nothing and I don't want to make a bunch of changes as I go

gigawhitlocks:
also lol @ "the sad path" -- great term for it

gigawhitlocks:
oh my god i made this branch in the main repo; closing PR; deleting branch

Issue #4403, opened 2017-03-27
Improve integration test 11-03 to test with an actual registry


jakedsouza:
#For stories, please include the information below:

**User Statement:**

As a vic engineer, I want to ensure that upgrade of VIC to the latest version updates the insecure registries correctly 

**Details:**
For issue #4388 we have an integration test 11-03-Upgrade-InsecureRegistry which tests that an insecure registry added to an older version of vic is updated correctly during an upgrade 
Currently the test checks guestinfo information to validate that the upgrade was successful
We should test against an actual insecure registry, instead of relying on guestinfo 
 
**Acceptance Criteria:**
 Update test `11-03-Upgrade-InsecureRegistry` to use an actual insecure registry

1. Deploy harbor as an insecure registry both https with untrusted certs and on http
2. Deploy an older VCH with the registry created in step1
3. Do a docker pull from the registries  
4. Upgrade the VCH to the latest 
5. Do a docker pull again

mhagen-vmware:
Thanks Jake, I am going to keep this high as this needs to go into next sprint.

PR #4485, opened 2017-03-31
add dind support in test environment

maplain:
1. Add dind in test image;
2. add keywords in Docker-Util.robot to support start and stop customized docker daemon both through host daemon and local binary. Started daemon will run alongside test container in the former case, and locally inside the test container in the latter case.

Note:
1. local binary can not be used out-of-the-box since it requires test container is started with '--privileged' turned on, which is not the case right now.

Manually tested in gcr.io/eminent-nation-87317/vic-integration-test:1.22

Also, there is an example here: https://github.com/maplain/vic/blob/dockerd-example/tests/resources/dockerd-example.robot which is not included in this PR.


maplain:
This is part of the attempt to address this issue: https://github.com/vmware/vic/issues/4403
need to start/stop customized docker daemon(specifically, --insecure-registry)

infra/integration-image/Dockerfile.test
```diff
@@ -57,3 +57,31 @@ RUN curl -fsSLO https://releases.hashicorp.com/packer/0.12.2/packer_0.12.2_linux
 
 RUN curl -sL https://deb.nodesource.com/setup_7.x | bash -
 RUN apt-get install -y nodejs
+
+# Add docker in docker support
```
jakedsouza: You need to push this container to the repo `gcr.io/eminent-nation-87317/vic-integration-test:xx` and bump up the URL in all locations, including secrets for this to be picked up

maplain: isn't that the next step after this PR is merged(modifications on this Dockerfile are approved)?

tests/resources/Docker-Util.robot
```diff
@@ -51,6 +53,44 @@ Get Container IP
...
+    ${handle}=  Start Process  ${dockerd-path} ${dockerd-params} >${log} 2>&1  shell=True
+    Process Should Be Running  ${handle}
+    :FOR  ${IDX}  IN RANGE  5
+    \   ${pid}=  Run  ps aux | grep '[d]ockerd ' | awk '{print $2}' # Space after [d]ockerd in grep is essential to filter shell process
+    \   Run Keyword If  '${pid}' != '${EMPTY}'  Set Test Variable  ${dockerd-pid}  ${pid}
+    \   Exit For Loop If  '${pid}' != '${EMPTY}'
+    \   Sleep  1s
+    [Return]  ${handle}  ${dockerd-pid}
+    
+Cleanup Container On ID
```
mhagen-vmware: cleanup local container on id

tests/resources/Docker-Util.robot
```diff
@@ -51,6 +53,44 @@ Get Container IP
...
+    [Arguments]  ${id}
+    ${rc}  ${output}=  Run And Return Rc And Output  docker stop ${id}
+    Should Be Equal As Integers  ${rc}  0
+    Should Be Equal  ${output}  ${id}
+    ${rc}  ${output}=  Run And Return Rc And Output  docker rm ${id}
+    Should Be Equal As Integers  ${rc}  0
+    Should Be Equal  ${output}  ${id}
```
mhagen-vmware: simple is probably better here?  docker rm -f  ${id}

tests/resources/Docker-Util.robot
```diff
@@ -51,6 +53,44 @@ Get Container IP
...
+
+Kill Local Docker Daemon
+    [Arguments]  ${handle}  ${dockerd-pid}
+    Terminate Process  ${handle}
+    Process Should Be Stopped  ${handle}
+    ${rc}=  Run And Return Rc  kill -9 ${dockerd-pid}
+    Should Be Equal As Integers  ${rc}  0
+
```
mhagen-vmware: How are you using this? Why are you terminating process and running a kill -9?  Would it be cleaner and nicer to just do 'service docker stop'?

tests/resources/Docker-Util.robot
```diff
@@ -51,6 +53,44 @@ Get Container IP
     Should Be Equal As Integers  ${rc}  0
     [Return]  ${ip}
 
+# Use host docker daemon to start a dind container alongside this test container
+Start Docker Daemon On TCP
```
mhagen-vmware: not sure where TCP comes from here?

tests/resources/Docker-Util.robot
```diff
@@ -51,6 +53,44 @@ Get Container IP
...
+    [Arguments]  ${docker-params}  ${dind-version}=docker:1.13-dind
+    ${rc}  ${id}=  Run And Return Rc And Output  docker run -it --privileged ${docker-params} -d ${dind-version}
+    Should Be Equal As Integers  ${rc}  0
+    [Return]  ${id}
```
mhagen-vmware: The id is only use able here like this if you have already done a pull, instead of leaving the pull for other code you should probably have the pull here.

tests/resources/Docker-Util.robot
```diff
@@ -51,6 +53,44 @@ Get Container IP
...
+# The local dind version is embedded in Dockerfile
+# docker:1.13-dind
+# If you are running this keyword in a container, make sure it is run with --privileged turned on
+Start Docker Daemon Locally
+    [Arguments]  ${dockerd-params}  ${dockerd-path}=/usr/local/bin/dockerd-entrypoint.sh  ${log}=/var/log/daemon-local.log
+    File Should Exist  ${dockerd-path}
+    ${handle}=  Start Process  ${dockerd-path} ${dockerd-params} >${log} 2>&1  shell=True
+    Process Should Be Running  ${handle}
+    :FOR  ${IDX}  IN RANGE  5
+    \   ${pid}=  Run  ps aux | grep '[d]ockerd ' | awk '{print $2}' # Space after [d]ockerd in grep is essential to filter shell process
```
mhagen-vmware: for debuggability you should really break this up some other wise the pipes all mask out failures within the statement.

tests/resources/Docker-Util.robot
```diff
@@ -51,6 +53,44 @@ Get Container IP
...
+    [Arguments]  ${docker-params}  ${dind-version}=docker:1.13-dind
+    ${rc}  ${id}=  Run And Return Rc And Output  docker run -it --privileged ${docker-params} -d ${dind-version}
+    Should Be Equal As Integers  ${rc}  0
+    [Return]  ${id}
+
+# The local dind version is embedded in Dockerfile
+# docker:1.13-dind
+# If you are running this keyword in a container, make sure it is run with --privileged turned on
+Start Docker Daemon Locally
+    [Arguments]  ${dockerd-params}  ${dockerd-path}=/usr/local/bin/dockerd-entrypoint.sh  ${log}=/var/log/daemon-local.log
```
mhagen-vmware: for the CI at least, we don't grab /var/log logs, default might better be set to the local folder?

tests/resources/Docker-Util.robot
maplain: that container is running alongside with drone containers, not inside drone test container.

tests/resources/Docker-Util.robot
maplain: oh, it should be something like "Start Docker Daemon Over Host Socket"

tests/resources/Docker-Util.robot
maplain: because it's started by dockerd binary not 'service docker start'. 'service docker start' doesn't seem to work properly. besides, host docker socket is mapped into container, we can use it directly by -H unix:///var/run/docker.sock. This is what "Start Docker Daemon Over Host Socket" does.
added dockerd binary is trying to give us the ability to start a daemon with customized parameter(like --insecure-registry) running locally(inside drone test container) on unix:///var/run/docker-local.sock

tests/resources/Docker-Util.robot
maplain: I don't really understand. as far as I know, docker run will pull automatically if image is not available. and ${id} corresponds to stdout. lengthy information from docker pull will not fill in $id.

tests/resources/Docker-Util.robot
maplain: I didn't explain this clearly. when I used TCP, I mean this docker daemon is only accessible through tcp://xx.xxx.xx, not through local unix socket.

tests/resources/Docker-Util.robot
maplain: 👌

tests/resources/Docker-Util.robot
maplain: 👌

tests/resources/Docker-Util.robot
maplain: 👌

tests/resources/Docker-Util.robot
mhagen-vmware: ok, that is fine

tests/resources/Docker-Util.robot
mhagen-vmware: are you going to make this change?

maplain: yes, when this line failed, it will dump information in every step.

PR #4491, opened 2017-03-31
Add logic to support testing and publishing to release and tagged branches

### LINKED ISSUE #4470: Tagging release process needs to support non-master branch
When we tag a commit for a release, it comes into drone as something like:
mdubya66 authored 13 minutes ago to refs/tags/v1.1.0-rc2

We need to modify .drone.yml and integration-test.sh in order to support this running the full tests and publishing the artifact to bintray.

mhagen-vmware:
When we tag for a release and if we need to create a temporary branch for patching a release, we need to support testing and publishing the artifacts.

fixes #4470


rajanashok:
lgtm

PR #4497, opened 2017-03-31
Add vic-machine update doc

### LINKED ISSUE #4418: Research Spike: UX define for vic-machine reconfigure
As a developer, we need a user friendly vic-machine command line for reconfigure operation to make sure it really helps user to solve their problems.

In details, the requirement is to define vic-machine command keywords for operations: create, read(list), update and delete.
And possibly rollback.

Not every operation requires all these four keywords, but for any required operation, they should use the keywords defined in this story.

If there is requirement to update existing vic-machine command, for example inspect can be expanded for existing configuration display, the requirement should be investigated in this story.

Another think about user experience improvement is to support yml format manifest file for vic-machine create/reconfigure. But if this option is took, we'll need a change in vic-machine create as well, this should be investigated in this story.

Output:
Define doc for new commands definition
Stories under this reconfigure epic for all existing vic-machine command update, e.g. inspect

Not required:
This story is for the user experience design, not for each particular configuration change, so do not add stories for vic-machine reconfigure for each different operation, e.g. update vSphere thumbprint, bridge network, etc.

emlin:
@hickeng @corrieb @karthik-narayan @mdubya66 
Here is the draft of overall commands and impact on existing vic-machine commands. There are two options introduced for how to define vic-machine update options. Please leave your preference or suggestions on it. Thanks
Fixes #4418 

btw, if you prefer to make any feature happens marked in P1, please leave your comments as well. By default, P1 means not in the scope.


doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,207 @@
...
+
+  In vic-machine create, we already have 57 options to define VCH configurations, though not all of them can be updated, quite amount of them should be able to be changed, and this number will grow up over time, so we definitely do not want to introduce 3X that number of options to update (update means add/modify/remove)
+
+## Options
+First option is to introduce update primitives
+
+### Delta Options
+#### Update Primitives
+
+- add
```
hickeng: This approach is procedural, however the current vic-machine options are predominantly declarative.
This is going to result in a very large matrix of possible combinations which will be extremely awkward to vet and test.

I suggest using a different approach:
0. *DO NOT SUPPORT DELETE AT THIS TIME* - it's just fraught
1. extract the current VCH configuration into installData (or similar base struct - this is the in memory representation of the manifest concept).
2. run through the `reconfigure` options using the same parsing that we have for create (basically no new code) to generate a second installData
3. provide code to combine two installData's
  a. for elements with names (e.g. container-net, volume-store) if there's a value provided against that name in the second installData it's a modify op
  b. for elements without names (e.g. --cert) if there's a value provided it's a replacement
  c. _maybe_ have some `reconfigure` specific arguments for those options that generate lists: `--clear-registry-cas`, `--clear-tls-cas` to allow a means to remove existing entries (the default list argument behaviour being to append) that would be specified before new `--registry-ca` options
4. run the validator against the combined configuration
  a. this should include the validation for ip address range conflicts, etc, which are currently inlined in the argument processing.

doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,207 @@
...
+
+We do not allow command like following
+- vic-machine update --add --volume-store ... --container-network ... --rm --volume-store ...  --dns-server ...
+
+In this way, the update command can only run one kind of change at one time
+
+### Whole Thing Update
+Other than introducing delta primitives, we could also require user to specify all configurations during update. That means the old VCH configuration will be replaced completely by this new configuration.
+
+Anything specified in the new configuration, which does not exist in old configuration will be added, anything missing in new configuration will be removed. And for the configuration can only have one item, will be replaced.
```
hickeng: This is essentially what I described in my comment on the update primitives, however by dropping delete it means you only specify the deltas, not the entire thing.

doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,207 @@
...
+
+To have a better user experience, vic-machine should notify user for what's changed, for example, list of volumes will be added, and another list will be removed.
+
+Pros of this option is no new options introduced and all options used in vic-machine create could be reused in update command.
+
+Cons of this option is that every time to update a VCH, user need to know all the old configuration first, otherwise some configuration will be removed by mistake.
```
hickeng: see previous comment.

doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,207 @@
...
+
+vic-machine inspect command is available now, just it shows only the execution endpoint of VCH, instead of the configuration. We can extend that command, to support configuration inspection as well.
+
+- vic-machine inspect <--configuration, --conf>
+  show VCH configuration only
+- vic-machine inspect <--execution, -e>
+  show VCH execution endpoint (default behavior)
+
+### Configuration Inspection
+For configuration inspection, two output format should be supported
```
hickeng: This is basically generating the vic-machine manifest that we spoke about last year - if we do this we should take care to structure it in such a way that it can be used as such:
* input for creating a new VCH
* the format doesn't prohibit use for authority delegation.

doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,207 @@
...
+Endpoint:
+	Cpu: 1
+	Memory: 2048 MHZ
+	Debug: 3
+	SSHEnabled: true
+```
+
+- Command Option Format (corresponding to Delta Update option)
+
+  The yml format output is more readable for VIC admin, but if users want to recreate their vic-machine create or update command, it's hard to do based on that format, because there is no one on one mapping between the configuration and command options.
```
hickeng: The command line options were only intended to bootstrap a manifest - if we have one then we should allow that to be used as an input, with options overriding/modifying the manifest contents, in _exactly_ the same way I describe `reconfigure` in previous comment.

mdubya66:
@lweitzman can you please review regarding update vs. reconfigure vs something else. 
+1 for going the manifest approach
+1 for not worrying about delete at this time. 

Looking at the prioritized list from PM we should focus on these 3 for 1.2
Reconfig	[2.1] Update VC creds	P0	10		24																				
Reconfig	[2.2] Update the TLS of VCH	P0	10		25			(this needs further clarification around what this means vis-a-vis certificates)																	
Reconfig	[2.3] Enable / disable debug state	P0	20		52	

From engineering we feel that these are the next priorities
Reconfig	[2.7] Update container network	P0	70		77			(this should be Add only, not update)
Reconfig	[2.8] Add / remove Volume store	P0	80		78			(not remove)
Reconfig	[2.10] Add/remove insecure registries	P0	100		90		(note this is for secure registries too)

doc/design/vic-machine/update.md
emlin: sounds good. The implementation part is similar to what I'm thinking, just in this story, we're not focused on that part, so no more description about it.

For the CLI design, this is an update for anything specified.
And also, the idea for sth really need to be cleared, using new options --clear-registry-cas, --clear-tls-cas.
This might introduce many new options over time, but user can only specify required options in one reconfigure command. So should be better.

doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,207 @@
...
+
+- add
+- modify
+- remove, rm
+
+In one update command, only one delta primitive is acceptable, and only that kind of operation will be executed.
+
+#### Example Command
+
+- vic-machine update --add --volume-store ... --container-network ...
```
emlin: sure

doc/design/vic-machine/update.md
emlin: yeah, your suggestion combined those two options.

doc/design/vic-machine/update.md
emlin: I tends to have manifest file more like the yml file format, because that format data is more readable. And technically to create/update from a manifest file, we could accept any data format.
I mentioned this command line option is because without manifest file support, user can only create/update from command options.
This should be the subtle difference for manifest file solution as well.

doc/design/vic-machine/update.md
emlin: So far, I prefer to use yml format configuration file as the manifest file, which is a flexible structure. I think we can have a separate section for authority delegation while we need to support that.

emlin:
if we decide to use reconfigure command name, I'll move the update.md file to reconfigure.md file, but before finish review, I'll leave it unchanged, to track all review comments update

doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,175 @@
+# VCH Reconfigure (P0)
+This document defines the VCH reconfigure command line, and the related changes in vic-machine, to make VCH reconfigure process more reasonable.
+
+Here is the key points considerred in this Definition
```
mhagen-vmware: Here are the key points considered in this definition:

doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,175 @@
...
+- Batch Operation
+
+  We'd like to change VCH configurations more than one item at one time to make the command easy to use
+- Reuse all options used in vic-machine create
+
+  In vic-machine create, we already have 57 options to define VCH configurations, though not all of them can be updated, quite amount of them should be able to be changed, and this number will grow up over time, so we definitely do not want to introduce 3X that number of options to reconfigure (update means add/modify/remove)
```
mhagen-vmware: quite amount of them -> many of them

mhagen-vmware: this number will grow ~~up~~ over time

doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Options
+First option is to reconfigure all in one command
+
+### Implicit Add/Modify and Explicit Delete
+All existing vic-machine create options are supported in vic-machine reconfigure command. The meaning of those options are as following:
+- for elements with labels (e.g. container-network, volume-store), if there's a value provided against that label in the reconfigure, it's a modify operation
+- for elements without labels (e.g. --cert), if there's a value provided it's a replacement
+- if need to remove existing configuration (e.g. static ip configuration for client-network, or registry-ca), define new options as --clear-registry-cas, --clear-client-network. Then existing configuration for registry-ca and/or client-network will be cleared. If not new configuration e.g. for client-network specified, it will share with other networks with same logic in vic-machine create.
```
mhagen-vmware: I would think to keep things simple we would just want to use the same option and pass an empty argument, it seems excessive to have another option to clear the parameter.  That will look to double or more our already massive list of options?

doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,175 @@
...
+The idea of this option is to update a few configuration each time, which is one delta update option. For example, to update volume-store, we can use following command
+```
+vic-machine reconfigure volume-store --rm ds://datastore/volume:default
+vic-machine reconfigure volume-store --add nfs://host:port/container:nfs
+```
+
+The command format is like ```vic-machine reconfigure <reconfigure object> --<reconfigure keyword> --<reconfigure option> <configuration value>```
+
+Pros:
+- The whole configuration options are splitted to many sub commands. Then in each sub command, only a small number of options are available, and that does not increase over time. Only the sub command number will increase.
```
mhagen-vmware: splitted -> split

doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,175 @@
...
+
+Cons:
+- We'll need to reinvent option names
+
+  For example, to update client network, user need to specify all options if they want to use static ip. Then in one update command, we'll need all those options. Here is the command looks like ```vic-machine reconfigure bridge-network --port-group value --gateway value --ip value```
+  The existing option name in vic-machine create is ```--client-network value   --client-network-gateway value  --client-network-ip value```, we'll need to remove the prefix cause that is already shown as sub-command name.
+
+Preference:
+
+- Looks like the first option is better, just leave the sub command option here, in case someone could find out more benefit from that option
```
mhagen-vmware: It seems to me the clearer and more user friendly approach to this would be to just pass all parameters as you would to vic-machine create but to vic-machine reconfigure? Not adding any new options and using all the same names and formats that create uses. It doesn't look like either approach does that though?  The user should also be able to pass one or more parameters that they want changed?

emlin: hmm... I ever had one option to pass in all vic-machine command options, without any additional option introduced.
That means user need to update everything in one time. Anything new means add, or replace, but anything missing in the options means delete.

@hickeng left comments for how to define the delta update, which is currently the first option

Compare to current delta update option, the whole update options looks heavy, and we're not support reconfiguration for all configurations in one release, delta update looks more reasonable. So I removed the whole update option from the document

doc/design/vic-machine/update.md
emlin: cause sometimes we use several options to define one configuration, e.g. static ip for any network. In that case, one clear option is enough. So we don't need that massive number of clear options, but certainly, that will be a lot.

But this clear option is introduced for delete existing configuration, empty option cannot solve this issue for list options. User might want to add a few into existing list, or remove all existing values, and add another few values.
For the latter case, empty option does not help

doc/design/vic-machine/update.md
mhagen-vmware: @emlin from above: "The user should also be able to pass one or more parameters that they want changed?" I don't think you understood my comment, I was not suggesting a user had to pass in all parameters at once, I am simply saying the options should be the same as vic-machine create and they should be able to pass one or more in to change the value. I understand the delta update option, I just wanted it to be a bit simpler..

doc/design/vic-machine/update.md
hmahmood: For list options, since specifying a value is modification, that can be used to delete/add items. No value could be interpreted as clearing the list. I agree with @mhagen-vmware about using empty values.

doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Implicit Add/Modify and Explicit Delete
+All existing vic-machine create options are supported in vic-machine reconfigure command. The meaning of those options are as following:
+- for elements with labels (e.g. container-network, volume-store), if there's a value provided against that label in the reconfigure, it's a modify operation
+- for elements without labels (e.g. --cert), if there's a value provided it's a replacement
+- if need to remove existing configuration (e.g. static ip configuration for client-network, or registry-ca), define new options as --clear-registry-cas, --clear-client-network. Then existing configuration for registry-ca and/or client-network will be cleared. If not new configuration e.g. for client-network specified, it will share with other networks with same logic in vic-machine create.
+
+The benefit of this option is it can update all things need to be updated in one command, but will introduce a bit more new delete options.
+
+### Separate Reconfigure to Sub Commands
```
hmahmood: This seems less intuitive than the previous item.

doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,175 @@
...
+--bridge-network        bridge
+--bridge-network-range  172.16.0.0/12
+...
+```
+
+Based on this output, it's easy to modify existing configuration to reconfigure command options
+
+Note: Special character might be escaped based on current OS platform. But even with this, the command options cannot be copied to other platform.
+
+- Subcommand Inspect (corresponding to reconfigure subcommand option)
```
hmahmood: My 2c is that this would be overkill. Output of just the vic-machine options from a yml file or "live" VCH should be sufficient.

doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,175 @@
...
+  The reconfigure subcommand option might change command option based on each different sub command requirement, to get current configuration, we'll need to have command inspect for each subcommand. So the option is to extend vic-machine inspect command to include sub command as well.
+  For example, ```vic-machine inspect bridge-network``` will print out current configuration in the command format, to make sure user can easily generate their reconfigure command based on that.
+
+## Update In Progress Status
+Same to vic-machine upgrade, concurrent reconfigure requests for same VCH is not allowed. vic-machine should be able to detect if one VCH is running reconfigure/upgrade. If true, return useful error message. And also should not leave that reconfigure flag in there after reconfigure is stopped, or even interrupted.
+
+Issue https://github.com/vmware/vic/issues/4069 is tracking this problem.
+
+## Backward compatible
```
hmahmood: `compatible -> compatibility`

doc/design/vic-machine/update.md
```diff
@@ -0,0 +1,175 @@
...
+The Reconfigure Rollback is talking about after a successful reconfigure, vic-machine is still able to rollback VCH's old configuration. The command is like vic-machine reconfigure --rollback
+
+This feature is nice to have, but as vic-machine upgrade already have this capability, mark it as P0.5.
+
+# Manifest File
+vic-machine create already has more than 50 options, and the number will increase over time. With this reconfigure feature added, most of those options need to be updated as well. To improve user experience, in my opinion, we need Manifest file to simplify vic-machine command generation.
+
+[vic-machine design](vic-machine.md#vch-manifest) mentioned manifest idea, here we'll start from a simple implementation. The manifest file format will be same to the yml file format, mentioned in [configuration inspection](#configuration-inspection)
+
+- vic-machine inspect --conf --manifest filename will dump all configurations in yml file format to the manifest file.
```
hmahmood: Can we combine the two options into one? Is there more than one sub option for `--conf`?

doc/design/vic-machine/update.md
emlin: I didn't get your point. Can you give example for your suggestion? Cause for the delta update, to handle delete, we have to introduce new options

doc/design/vic-machine/update.md
emlin: agree, just leave it there for one possibility

doc/design/vic-machine/update.md
emlin: This is used to work with separated update command, cause that will introduce different command options. If that one is dropped, this will not be used as well

doc/design/vic-machine/update.md
emlin: hmm.. should be fine. then all options for `vic-machine inspect` will be as following:
- `vic-machine inspect` will show the VCH endpoint information, which is current behavior
- `vic-machine inspect --cmd` will dump command options format configuration
- `vic-machine inspect --conf` will dump yml file format configuration
- `vic-machine inspect --manifest` will dump yml format configuration into manifest file

emlin:
@hmahmood 
> For list options, since specifying a value is modification, that can be used to delete/add items. No value could be interpreted as clearing the list. I agree with @mhagen-vmware about using empty values.

There are two different usage for list options. One is with name, e.g. container network, volume store, etc. For that case, we're okay to use empty configuration e.g. `--container-network :cn1` to delete existing `cn1` container network, but if the list option is anonymous, e.g. registry-ca, how to delete all existing registry-ca, and add a new one at the same time?

emlin:
I tried to add an empty value for the list slice, and also add other items at the same time, the cli framework works, e.g. with options `--vs ''  --vs string1 --vs string2`, we'll get input like following:
["", "string1", "string2"]

We could define the empty string as delete all. But is that sth intuitive?

@hmahmood @mhagen-vmware 
If this is what you you want, we might need @ali5ter help to make decision on this

ali5ter:
@emlin explained the two questions that seem to be emerging from this thread. Forgive my iterating them here; it helps me frame them and allows you to tell me if I misunderstood anything :)

### Configuration command to update an object after creation
**Problem**
`vic-machine upgrade [object]`
may be confused with
`vic-machine update [object]`

**Suggested solution**
`vic-machine reconfigure [object]`

**My 2 cents**
If I understand this correctly, the intent is to configure the settings of an object that has already been deployed. Rather than re-configure, I suggest the verb 'configure', e.g.`vic-machine configure [object]` This seems easier to type and also more concise.

### Option format to add/delete values, specifically list values
**Problem**
Can add using
`--registry-ca ca1 --registry-ca ca2`
but how to remove?

**Suggested solutions**
`--clear-registry-ca --registry-ca ca4`
`--registry-ca "" --registry-ca ca4`

**My 2 cents**
CLIs are usually pretty explicit when adding/removing items and don't necessarily allow you to delete and add something in one command, e.g.
`cmd configure [object] add [key] [value]`
`cmd configure [object] add [key] [value] [value] [value]` (adding a list of values)
`cmd configure [object] remove [key]`
`cmd configure [object] remove [key] [value]` (removing a value from a list)`

I agree that adding delete or clear options for each setting is going to expand the number of options considerably. However, it does make the notion of removing the setting much more explicit.

Using the same option to clear and add values is very efficient but I don't think it particularly intuitive. The user would need to know this format - perhaps (re)discovered through trial and error and looking at the help command. It also suggests order, i,e, you need to clear before you add. So the user would have to understand that in certain circumstances, the option is a positional parameter.

So what to do? 

Well, another way to do this, which may have been suggested above, is to always define the values for the option, e.g.
`--registry-ca ca1 ca2 ca3` (registry-ca=['ca1', 'ca2', 'ca3'])
`--registry-ca ca1 ca3` (registry-ca=['ca1', 'ca3'], i.e. deletes ca2)
`--registry-ca` (registry-ca=[], clears all values - not sure this is a state you want to allow)

This seems to tick the following boxes:
- [x] explicitly defines the values for the setting
- [x] doesn't add any more options to clear or delete
- [x] avoids forcing the user to learn any new syntax

emlin:
@ali5ter this is awesome, just what we need, no any misunderstood.
I'll update the document following your suggestions.

doc/design/vic-machine/configure.md
```diff
@@ -0,0 +1,182 @@
...
+- anything not specified means no change
+- any option specified in configure command means replacement.
+  - if the old value is not specified during creation, this new value will be added
+  - if there is old value specified, no matter it's a list value or not, we'll replace the old value with the new one specified in configure command.
+
+    here are the possibilities to change list option ```--registry-ca```, if the create command option is ```--registry-ca ca1 --registry-ca ca2```
+	* ```--registry-ca ca2 --registry-ca ca3```, means new configuration will be [ca2, ca3], ca1 is deleted, and ca3 is added
+	* ```--registry-ca ca2```, means new configuration will be [ca2], ca1 is deleted
+	* ```--registry-ca ""```, means ca1 and ca2 are all deleted // Here the empty string value is required, and the representation in different os platform is different.
+	* nothing specified for registry-ca, then existing [ca1, ca2] is not changed
```
hickeng: I agree with this over my suggestion - it's completely consistent behaviour without trying to construct a special-case for addition. While this will make some common operations more awkward it's better than adding inconsistencies.

doc/design/vic-machine/configure.md
```diff
@@ -0,0 +1,182 @@
...
+	OpsUser: "Operator"
+	Name: "VCH-1"
+	Proxies: 
+		- "https://host:port"
+		- "http://host:port"
+Compute:
+	Path: " myCluster/Resources/myRP"
+	CpuLimit: 0 MHz
+	MemoryLimit: 0 MHZ
+Storages:
```
hickeng: `Storage` or `Stores`

doc/design/vic-machine/configure.md
```diff
@@ -0,0 +1,182 @@
...
+Registries:
+	InsecureRegistries:
+		- "..."
+	RegistryCA:
+		- "ca file"
+Endpoint:
+	Cpu: 1
+	Memory: 2048 MHZ
+	Debug: 3
+	SSHEnabled: true
```
hickeng: I do not see any positives to having SSHEnabled in the schema, or an ability to set it on by default. The only partial positive is to remove a single step in allowing a viadmin to immediately ssh to the endpoint to check it's config - which is mostly pointless as any checks they would perform normally we should be adding into vic-machine. Doing so would also come with significant security obligations around password management and persistence.

For development we already have scripts that ssh into the endpoint with a single command (see `vic-ssh` from `infra/scripts/bash-helpers.sh`) so no value added there.

I realize this is a sample, but I've observed a tendency for samples to get adopted in their entirety without being revisited with project time pressure.

doc/design/vic-machine/configure.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+- Command Option Format
+
+  The yml format output is more readable for VIC admin, but if users want to recreate their vic-machine create or configure command, it's hard to do based on that format, because there is no one on one mapping between the configuration and command options. 
+
+  Here we'd support another option as "vic-machine inspect --cmd --conf", this will print out the vic-machine create command options minus secret information.
```
hickeng: The only obvious use cases I can see for this mechanism are:
1. generating `vic-machine create` command lines for testing
2. viadmin copy/paste when moving between systems and `cat manifest/cat >manifest` is more steps.

As such I think this should be considered as a secondary mechanism to the manifest and could be done via multiple steps:
1. extract manifest
2. construct command line from manifest (`--dry-run` option to `create` or `configure` perhaps)

If we do keep this as a specific option in `inspect` then we should also allow it to operate on manifests as well as existing VCHs.

doc/design/vic-machine/configure.md
```diff
@@ -0,0 +1,182 @@
...
+It's possible that after vic engine released new version, user is trying to configure older version's VCH with newer vic-machine.
+
+There are few options:
+- Run configure using same version's vic-machine (this option should technically always work)
+
+  For any customer that does not want to change vic-machine version, they need to run configure using same version's vic-machine.
+
+- Combine upgrade process into configure
+
+  Upgrade is actually one special kind of configure, technically it's easy to have upgrade run together with configure. User might want to configure VCH, and also want to upgrade it after or before configure. In that case, a combined configure and upgrade command can provide a better user experience.
```
hickeng: Definitely. There is also value in being able to rollback an incorrect configure as noted below.

doc/design/vic-machine/configure.md
```diff
@@ -0,0 +1,182 @@
...
+This feature is nice to have, but as vic-machine upgrade already have this capability, mark it as P0.5.
+
+# Manifest File
+vic-machine create already has more than 50 options, and the number will increase over time. With this configure feature added, most of those options need to be updated as well. To improve user experience, in my opinion, we need Manifest file to simplify vic-machine command generation.
+
+[vic-machine design](vic-machine.md#vch-manifest) mentioned manifest idea, here we'll start from a simple implementation. The manifest file format will be same to the yml file format, mentioned in [configuration inspection](#configuration-inspection)
+
+- vic-machine inspect --manifest filename will dump all configurations in yml file format to the manifest file.
+- vic-machine create --manifest will load the manifest file and create VCH based on it.
+- vic-machine configure --manifest will load the manifest file and update VCH configuration based on it.
```
hickeng: I would also like us to be able to use a manifest as a base, with CLI options to refine or specialize. For example the following would use the same semantics as `configure` to merge the CLI options with the configuration specified by the manifest.:
```
vic-machine create --manifest=conf.yml --name=new-vch --memory=12GB
```

This allows both easy use of manifests as templates and positions us well for adding authority delegation mechanisms to the manifests.

hickeng:
@emlin We should also see if there's a Go yaml schema mechanism available - I don't know if it would add value without seeing how it operations, but any time you add a config file of any complexity you also introduce a need for very precise error messages about any errors it contains, beyond just basic format parsing.

doc/design/vic-machine/configure.md
emlin: I did add that item as a sample, but agree with you, sample also means functionality in the future. Will remove that, and be more careful about what should be added.

doc/design/vic-machine/configure.md
emlin: hmm.. didn't quite get your point. I assume this option is the way to get current configuration and the information can easily be updated a little bit for `vic-machine configure`.
As for the manifest file, it's doable to use inspect command to read and even modify manifest file, is that the point?

doc/design/vic-machine/configure.md
```diff
@@ -170,13 +169,33 @@ The Configure Rollback is talking about after a successful configure, vic-machin
 
 This feature is nice to have, but as vic-machine upgrade already have this capability, mark it as P0.5.
 
+# Inspect Command Enhancement
+
+- Dump certificates
```
hickeng: This is needed as part of `--conf` but I don't think there's any value in having it separate. The VCH doesn't store the certificates needed by the client.

PR #4462, opened 2017-03-29
Fix state change logic for fixing invalid state cVM

anchal-agrawal:
Fixes #4397

PR #4503, opened 2017-03-31
Fix state change logic for fixing invalid state cVM

anchal-agrawal:
From #4462


mdubya66:
Approved for inclusing in 1.1 RC3

Issue #4486, opened 2017-03-31
Deploy OVA with certificate chain will store certificate in wrong format


lubronzhan:
Same as https://github.com/vmware/harbor/issues/1259
Fixed in Harbor not in VIC OVA.

## Details 

When Deploy OVA With certificate chain, like
```
-----BEGIN CERTIFICATE----
-----END CERTIFICATE-----
-----BEGIN CERTIFICATE----
-----END CERTIFICATE-----
```
The format of cert file will be twisted, for example, the first line of cert file will become
`——BEGIN CERTIFICATE—————BEGIN CERTIFICATE——`

But if deploy the OVA with certificate with only one root certificate inside the cert file, like
```
-----BEGIN CERTIFICATE----
-----END CERTIFICATE-----
```
Harbor can come up, and the cert format is correct

## Reproduce
Deploy the `harbor_0.5.0-9e4c90e.ova` with certificate chain. Could use the certificate using DBC account under `/build/apps/machines/puppet/puppetmaster/modules/certificates/files`

lubronzhan:
@frapposelli

lubronzhan:
Still have problem when deploy OVA using VC UI. 
When deploy it, it requires input ssl Certificate, I just copy from http://w3-dbc302.eng.vmware.com/dmaceachern/vic/eng.vmware.com.crt
and pasted it into the column, then it becomes one line, The output of `ovfenv --key ssl_cert_key' is also oneline without anything in between, like 
```
[...]

PR #4508, opened 2017-04-01
Dummy fix for parsing certificate reading form VC UI

lubronzhan:
When input the certificate file into VC UI, it will convert it into single line. This is to parse it as certificate file.

Dummy fix


lubronzhan:
related to #4486

Issue #4421, opened 2017-03-28
docker stats integration tests


cgtexmex:
User statement:  As a user of VIC, I want to know that docker stats functionality has been fully tested so that I can have confidence in it's behavior.

Details: Create integration tests for the docker stats functionality represented in #4413 -- CPU and Memory.

Acceptance criteria:
Tests defined and running in either CI or nightly.

mhagen-vmware:
This was merged and fixed into the docker 1.13 PR.

cgtexmex:
@mhagen-vmware I don't see where this was merged -- do you have a commit that includes the stats integration tests?

mhagen-vmware:
sorry good catch, I thought you said it was in Jake's PR, but it is not.  Why is you PR closed then?  Should you reopen and merge now?

cgtexmex:
I mistakenly closed it..I'll reopen in a bit...but the stats work is now waiting on #4653 -- stats needs the 1.13 client AND  the most recent version of govc (0.14)

PR #4550, opened 2017-04-04
fixed race condition and memory leak in docker stats

### LINKED ISSUE #4549: docker stats race condition
When executing `docker stats` the VCH appliance _might_ encounter a race condition that would orphan a go routine.  This orphaned routine would never be garbage collected resulting in a memory leak.

cgtexmex:
In some cases a race could be encountered when executing docker
stats.  The race would result in an orphaned go routine that
would be a memory leak. This change fixes that issue as well
as increases the unit testing coverage.

Fixes #4549

### DIFF
--- a/lib/apiservers/engine/backends/convert/stats.go
+++ b/lib/apiservers/engine/backends/convert/stats.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"sync"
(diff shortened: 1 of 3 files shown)


caglar10ur:
LGTM

cgtexmex:
This doesn't fix the issue...will close and open a future PR containing fix for race and #4421

Issue #4368, opened 2017-03-23
switch CI to use docker 1.13 as default


mhagen-vmware:
As discussed in the last sprint demo meeting, now that we have support for it, we want to make docker 1.13 the default for all of our CI tests.

anchal-agrawal:
#4322 should be addressed before this.

jakedsouza:
This can happen with moving to a new drone env #4267

jakedsouza:
Blocked by 
#4561   #4560 #4559 #4564

Issue #4504, opened 2017-03-31
docker run --rm will not auto remove container with 1.13.1 client


cgtexmex:
If you execute `docker run --rm busybox date` with a docker 1.13.1 client vic will not auto remove the container when execution is complete.  The same command works as expected with the docker 1.11 client.

Example:
```
docker version
Client:
 Version:      1.13.1
 API version:  1.25 (downgraded from 1.26)
 Go version:   go1.7.5
 Git commit:   092cba3
 Built:        Wed Feb  8 08:47:51 2017
 OS/Arch:      linux/amd64

docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES

docker run --rm busybox date
Fri Mar 31 20:22:52 UTC 2017

docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS                     PORTS               NAMES
2ef1541d38c0        busybox             "date"              17 seconds ago      Exited (0) 6 seconds ago                       eloquent_thompson
```

The container should have been removed.  I've not investigated, but potentially due to the fact that we are not populating the docker die event with the container exit code.

cgtexmex:
possible dupe #4241

cgtexmex:
This has been moved back to `InProgress` as it's dependent upon #4696 --

PR #4551, opened 2017-04-04
 [full ci]  Use docker1.13 by default

jakedsouza:
Changes -
- Change docker file for integration image to use 1.13 as the
  default docker binary
- Updated all references to the integration image to use a new tag 1.24
- Updated secrets file with the same change

Fixes: #4368


jakedsouza:
@maplain I pushed tag 1.24. For the DIND work could you push a new tag which includes this change as 1.25

mdubya66:
For the #4504 issue should we release note #4504 and comment out that test until it's fixed? @mhagen-vmware?

mdubya66:
FYI, We support Docker API version 1.25, not 1.24. I am not sure that is what is being referenced. 

Docker client version 1.13.0
Docker API version 1.25
Docker Compose version 1.9.0 and v2 and v2.1 compose files only

mhagen-vmware:
Yep, known issue with our current testing requirements so we should comment out #4504 test.

mhagen-vmware:
For the disposition of the failures in 1-24, ~we need to comment out the end of the test and tag it with a github check for #4561 ~ actually we can just change the last rc check to an output check:
```
    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run --net jedi debian ping -c1 first

    # cannot reach first from another network
    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run debian ping -c1 first
    ~Should Not Be Equal As Integers  ${rc}  0~
    Should contain  ${output}  unknown host
```

jakedsouza:
> For the disposition of the failures in 1-24, we need to comment out the end of the test and tag it with a github check for 4561 actually we can just change the last rc check to an output check:
> 
> ${rc} ${output}= Run And Return Rc And Output docker %{VCH-PARAMS} run --net jedi debian ping -c1 first
> Should Be Equal As Integers ${rc} 0
> Should Not Contain ${output} Error
> 
> # cannot reach first from another network
> ${rc} ${output}= Run And Return Rc And Output docker %{VCH-PARAMS} run debian ping -c1 first
> Should Not Be Equal As Integers rc 0
> Should contain ${output} unknown host

@mhagen-vmware This would have to be done for some other tests that check for a non zero return code. I think the same "fix" would work for test failures in 1-17

mhagen-vmware:
Yep, based on the current bug, any test that checks for non-0 rc will need to be changed.

mhagen-vmware:
we need to have at least one of them that does a github check instead of modifying the test though:
```${status}=  Get State Of Github Issue  780
    Run Keyword If  '${status}' == 'closed'  Fail  Test 1-1-Docker-Info.robot needs to be updated now that Issue #780 has been resolved
```

jakedsouza:
> Blocked by #4504

@caglar10ur #4504 does not seem to break any existing tests.

Issue #4297, opened 2017-03-17
Possible collision of tests using the same volume store directory


hickeng:
Looking into the failure of https://ci.vmware.run/vmware/vic/8746 the hostd output appears to show that there is a vic-machine delete running at the same time as our test.
Working assumption is that there was an attempt to delete another VCH that was using the same [datastore1] tests/volumes/ directory

This is one of the reasons that #4295 is open - if we're seeing this in an artificially simple CI environment then it will happen in a customer setup.

Failing fragment from PL log:
```
Mar 17 2017 19:16:59.741Z DEBUG op=259.46 (delta:4.257µs): [NewOperation] op=259.46 (delta:1.697µs) [github.com/vmware/vic/lib/apiservers/portlayer/restapi/handlers.(*StorageHandlersImpl).CreateVolume:373]
Mar 17 2017 19:16:59.741Z INFO  Creating directory [datastore1] test/volumes/test5
Mar 17 2017 19:16:59.746Z DEBUG Creating [datastore1] test/volumes/test5 error: ServerFaultCode: File [datastore1] test/volumes/test5 was not found
Mar 17 2017 19:16:59.746Z ERROR storagehandler: VolumeCreate error: soap.soapFaultError{fault:(*soap.Fault)(0xc4206ccf00)}
```

Failing volume create fragment from hostd log:
```
2017-03-17T19:17:00.521Z info hostd[C340B70] [Originator@6876 sub=Vimsvc.TaskManager opID=1e8b1af0 user=droneci] Task Created : haTask--vim.FileManager.makeDirectory-111434978
2017-03-17T19:17:00.521Z verbose hostd[C340B70] [Originator@6876 sub=PropertyProvider opID=1e8b1af0 user=droneci] RecordOp ADD: recentTask["haTask--vim.FileManager.makeDirectory-111434978"], ha-taskmgr. Applied change to temp map.
2017-03-17T19:17:00.521Z verbose hostd[BDC5B70] [Originator@6876 sub=PropertyProvider opID=1e8b1af0 user=droneci] RecordOp ASSIGN: info, haTask--vim.FileManager.makeDirectory-111434978. Applied change to temp map.
2017-03-17T19:17:00.522Z info hostd[BDC5B70] [Originator@6876 sub=Nfcsvc opID=1e8b1af0 user=droneci] Create requested for /vmfs/volumes/589f95c2-90c1a797-4b8f-0cc47ac8174a/test/volumes/test5
[...]

PR #4567, opened 2017-04-04
need to make the best of build servers

maplain:
Fixes: https://github.com/vmware/vic/issues/4569


mhagen-vmware:
Server selection in VCH-Util.robot is still round robin, if you merge this and don't fix the below then you will still get multiple builds on the same ESX test server:

```
    @{URLs}=  Split String  %{TEST_URL_ARRAY}
    ${len}=  Get Length  ${URLs}
    ${IDX}=  Evaluate  %{DRONE_BUILD_NUMBER} \% ${len}

    Set Environment Variable  TEST_URL  @{URLs}[${IDX}]
    Set Environment Variable  GOVC_URL  %{TEST_USERNAME}:%{TEST_PASSWORD}@%{TEST_URL}
```

anchal-agrawal:
Sidenote: `[specific ci=7-01-Regression]` is unnecessary because a PR build runs regression tests by default.

maplain:
updated.  @mhagen-vmware @anchal-agrawal  @jakedsouza 
my approach is checking every build server whether there is already a VCH-xxxx-xxxx running by using govc, before starting integration test. This makes its best effort to make sure url in filtered TEST_URL_ARRAY maps to idle server, though without guarantee it still holds true when that url is consumed. 
feel like It's not straightforward to do the synchronization among drone jobs. any idea?

jakedsouza:
> my approach is checking every build server whether there is already a VCH-xxxx-xxxx running by using govc, before starting integration test. 

This might skip machines where a VCH was running from a test that failed to clean it up

mhagen-vmware:
I think this could actually work, but we need to improve this a bit more to eliminate the chance of dangerous races like Jake suggests.  As long as you are grabbing the last 50 drone builds, just need to check status like we were doing before and we can build a fuller, more accurate picture of build status, but I think we need to focus on the drone side not the govc/esx side.

maplain:
I think what Jake suggested is not races, but skipping usable resources.

I'm really confused though. My first version is fetching latest 50 builds and s
imply making sure extra builds are waiting on number of status 'running'. However, as you pointed out above, VCH-Util is still doing round-robin which still risks running multiple builds on one machine. What's more, this approach has a severe flaw that in face of a sudden burst of builds, all of them will be waiting and that number of 'running' will never decrease.

So I switched to another approach which goes to every build server to check whether there is already a VCH-xxxx-xxxx running and filter out these machines by returning a modified TEST_URL_ARRAY. Thus VCH-Util can still do its round-robin cuz every url is usable. Problem is we don't have a strict synchronization here, thus url in TEST_URL_ARRAY is not 100% guaranteed to be idle. think about 3 jobs are stuck on waiting for idle servers and 2 of those build servers finish their job simultaneously.  But I think practically speaking, this would happen only in a small probability and even though it happened, there is no severe damage, just running more than 1 VCH-xxxx-xxxx on one build server, which already is happening right now.

As for the fact jake pointed out, it's caused by leak resources. And just to make functionality clearer, I think it's fair to make the assumption that VCH-XXXX-XXXX running on build server is the real test. On the other hand, tests should be responsible for the lifecycle of whatever resources they created. In a word, I think it should not be the scheduler's job to figure out whether some resources should be garbage collected.

maplain:
well, it seems harder to make sure environment is cleaned up when tests finished than started. but still it should be an another issue to address the leakage of resources.

caglar10ur:
Just curious, don't we have a race in the wait_for_idle_server function? We install and remove VCH's in every test so it looks like things will break if that function gets triggered while we are removing VCH but not finished the whole test suit.

maplain:
@caglar10ur  I'm not sure I understand you correctly.
But I wouldn't say that's a race since every vch installed in test is parameterized by drone build number and a 4-digit random number. At most, it's risking installing multiple vch on the same machine, which is actually already happening, and waiting too long before started even though the infrastructure is clean. if there were any conflict, I think it should be the datastore, since every vch also creates and destroys a specific portgroup.

caglar10ur:
@maplain if we are OK with running multiple VCH's on the same machine then why are we checking?

What I was thinking follows for reference;

I submit commit A
You submit commit B
Drone task XYZ (triggered by your commit) runs on machine ABC, it starts and runs test docker-1 and after that it starts uninstalling the VCH
Drone tash QWE (triggered by my commit) keeps checking machines to see whether there is any ESXi with no VCH{regex} on them
Drone task QWE just happens to check the exact moment while task XZY uninstalling docker-1 test and sees machine A as available and starts
Now those two races for the resources on the machine A

maplain:
@caglar10ur you're right. I misunderstood the aim of that script. Jake and I discussed this a little bit, then I realized by subtracting 8, it guarantees that.

I think the collision is https://github.com/vmware/vic/blob/master/tests/resources/VCH-Util.robot#L159-L163, before creating, we clean up all dangling resources by matching up their names. Thus, new test will essentially disrupt old one. 

And even if they are able to run together, there are still some real resource conflicts. like this issue jake pointed to me https://github.com/vmware/vic/issues/4297. And check out how we create vch, https://github.com/vmware/vic/blob/master/tests/resources/VCH-Util.robot#L176, all of them share the same path of image-store and volume store. I opened this issue a few days ago based on a discussion with @jakedsouza and @kreamyx, https://github.com/vmware/vic/issues/4651. A easy step would be parameterize the volume-store and image-store path like vch name. 

I'm not clear on all the resources vch need, thus not sure whether there are other potential conflicts.

maplain:
actually, substracting 8 is not enough, we should substract any multiples of eight.

mhagen-vmware:
That would be stronger, but unless someone cancels a job, then only going back 8 works.

PR #4587, opened 2017-04-05
[full ci] Add static code analysis improvements

### LINKED ISSUE #4443: Static code analysis
As a developer, I want CI to run static code analysis on the VIC code base and report the results.

VIC currently has a 'make gas' target to run [gas](https://github.com/GoASTScanner/gas) against the code base.

We should look at the current results and determine:

- [x] Which warnings can/should be acted upon

- [x] Which rules should be disabled, if any

- [x] Decide if/when to make CI fail based on gas results

anchal-agrawal:
This PR:
* Addresses or manually-verifies (with the `#nosec` comment) issues reported by the `gas` tool used in our builds
* Enforces a strict `gas` check in builds - builds will fail if changes cause a new warning in the `gas` reports 

Fixes #4443


cmd/tether/ops_linux.go
```diff
@@ -43,7 +43,7 @@ func (t *operations) Log() (io.Writer, error) {
 
 	// redirect logging to the serial log
 	log.Infof("opening %s/ttyS1 for debug log", pathPrefix)
-	f, err := os.OpenFile(pathPrefix+"/ttyS1", os.O_RDWR|os.O_SYNC|syscall.O_NOCTTY, 0777)
+	f, err := os.OpenFile(pathPrefix+"/ttyS1", os.O_RDWR|os.O_SYNC|syscall.O_NOCTTY, 0)
```
vburenin: 0600? I am actually not sure if it affects an existing file.

anchal-agrawal: If `os.O_CREATE` isn't specified in the flag, the mode doesn't matter.

cmd/docker/main.go
```diff
@@ -164,6 +164,7 @@ func startServer() *apiserver.Server {
 	}
 
 	// FIXME: assignment copies lock value to tlsConfig: crypto/tls.Config contains sync.Once contains sync.Mutex
+	// #nosec
```
caglar10ur: What is that FIXME and why we are not fixing it?

anchal-agrawal: @caglar10ur I'm not familiar with the implications, and that line seems to be coming from https://github.com/vmware/vic/pull/1994/files#diff-f346ba150f298a88e29c9f07d68d8c7aR179. The `tls.Config` has two fields that the FIXME comment mentions:
```go
	serverInitOnce sync.Once // guards calling (*Config).serverInit

	// mutex protects sessionTicketKeys and originalConfig.
	mutex sync.RWMutex
```

However, we don't seem to be copying those fields - `tlsconfig.ServerDefault()` doesn't set those fields.

anchal-agrawal:
@caglar10ur regarding adding reasons for each `#nosec` comment - the warnings `gas` generated with our code were mostly of three types:

1) File/dir permissions not being 600 and 700 respectively
2) Errors returned from functions were ignored
3) Commands being exec'd with variable params

For 1), the reasons are fairly self-explanatory - they're either log directories or files (which should be readable/writable by the owner and readable by other users). For 2), I've manually verified each case and the code implicitly takes care of the scenario where the error is returned. For 3), I've checked that the variables always have sane values.

Overall, I feel that the reasons for `#nosec` comments should come from the original developers themselves - hence, we can make that a practice from now on. If you'd like, I can add reasons but they wouldn't do full justice to the train of thought of the original developer.

caglar10ur:
they are fairly self-explanatory to you as you have been working on that code but they won't be for anyone new :) 

I don't think we need to write essays about why we add #nosec there but having something will help. It could be just as simple as repeating the error gas shows

```
 //FIXME(#nosec): Subprocess launching with variable.
```

etc.

anchal-agrawal:
@caglar10ur I've added the gas rule alongside each `#nosec` comment - thanks!

PR #4590, opened 2017-04-05
bump the integration image

mhagen-vmware:


jakedsouza:
What changed in the dockerfile?

mhagen-vmware:
@jakedsouza nothing, that is the problem. dockerfile changes went in earlier.

Issue #4585, opened 2017-04-05
docker stats --no-stream might hang indefinitely


cgtexmex:
if a user calls `docker stats --no-stream` before any stats are available for a given VM (i.e. it just started and no stats have been generated) then the call will not return.  

The only remedy is for a user to cancel the request.

PR #4596, opened 2017-04-05
fixed race and added support for all option in docker stats

### LINKED ISSUE #4549: docker stats race condition

cgtexmex:
In some cases a race could be encountered when executing docker
stats.  The race would result in an orphaned go routine that
would be a memory leak.

In addition if stats --no-stream -all was requested the persona
would hang on any stopped containers.  Stats now will consider
the container state when processing these requests.

Finally, the unit testing coverage was increased to just over
90%.

Fixes #4549, #4585, #4421


cgtexmex:
Add the integration tests to this PR -- this is dependent on #4551

Issue #4384, opened 2017-03-24
Error response from daemon: Server error from portlayer: A general system error occurred: vDS host error: see faultCause


mhagen-vmware:
Saw this error on latest Longevity run

Logs:
[container-logs.zip](https://github.com/vmware/vic/files/868610/container-logs.zip)

docker-personality
```
Mar 23 2017 21:20:55.521Z DEBUG [ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.(*ContainerProxy).AddVolumesToContainer:256] [23.813µs] 06a5cce29421084e8d7a0d7efcb60df2
Mar 23 2017 21:20:55.521Z DEBUG [BEGIN] [github.com/vmware/vic/lib/apiservers/engine/backends.(*ContainerProxy).CommitContainerHandle:399] 06a5cce29421084e8d7a0d7efcb60df2
Mar 23 2017 21:27:27.312Z DEBUG [ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.(*ContainerProxy).CommitContainerHandle:399] [6m31.790765765s] 06a5cce29421084e8d7a0d7efcb60df2
Mar 23 2017 21:27:27.312Z DEBUG [ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).containerCreate:322] [6m31.877874584s] Container.containerCreate
Mar 23 2017 21:27:27.312Z DEBUG [ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).ContainerCreate:260] [6m31.878095023s] 
Mar 23 2017 21:27:27.312Z ERROR Handler for POST /v1.23/containers/create returned error: Server error from portlayer: A general system error occurred: vDS host error: see faultCause
```

portlayer:
```
Mar 23 2017 21:20:55.521Z DEBUG [BEGIN] [github.com/vmware/vic/lib/apiservers/portlayer/restapi/handlers.(*ContainersHandlersImpl).CommitHandler:226] handle(06a5cce29421084e8d7a0d7efcb60df2)
Mar 23 2017 21:20:55.521Z DEBUG [BEGIN] [github.com/vmware/vic/lib/portlayer/exec.Commit:35] 8df08f3a7ac5fb7e52dca517b8ce2f3c87ee39f46a40496a324d3194d6951f69
[...]

PR #4597, opened 2017-04-05
[specific ci=Group6-VIC-Machine.6-07-Create-Network] Retry on intermittent failure

### LINKED ISSUE #3022: Longevity Failure: Error response from daemon: Server error from portlayer: A specified parameter was not correct: path
[14-1-Longevity-VCH-0-4365-container-logs.zip](https://github.com/vmware/vic/files/572455/14-1-Longevity-VCH-0-4365-container-logs.zip)

emlin:
fixes #3022 #4384 

@jzt and I are changing same file for different bugs, so this PR includes his change for bug #4384 as well.

### DIFF
--- a/pkg/vsphere/tasks/waiter.go
+++ b/pkg/vsphere/tasks/waiter.go
@@ -97,7 +97,7 @@ const (
 
 // isRetryErrors will return true for vSphere errors, which can be fixed by retry.
 // Currently the error includes TaskInProgress, NetworkDisruptedAndConfigRolledBack and InvalidArgument
-// The reason to retry for these errors is because sometimes vSphere needs longer time to finish sync up inside
+// Retry on errors NetworkDisruptedAndConfigRolledBack and InvalidArgument is to workaround vSphere or vSAN intermittent failure
 func isRetryError(err error) bool {
 	if soap.IsSoapFault(err) {
 		switch f := soap.ToSoapFault(err).VimFault().(type) {
--- a/pkg/vsphere/tasks/waiter.go
+++ b/pkg/vsphere/tasks/waiter.go
@@ -97,7 +97,8 @@ const (
 
 // isRetryErrors will return true for vSphere errors, which can be fixed by retry.
 // Currently the error includes TaskInProgress, NetworkDisruptedAndConfigRolledBack and InvalidArgument
-// Retry on errors NetworkDisruptedAndConfigRolledBack and InvalidArgument is to workaround vSphere or vSAN intermittent failure
+// Retry on NetworkDisruptedAndConfigRolledBack is to workaround vSphere issue
+// Retry on InvalidArgument(invlid path) is to workaround vSAN bug: https://bugzilla.eng.vmware.com/show_bug.cgi?id=1770798. TODO: Should remove it after vSAN fixed the bug
 func isRetryError(err error) bool {
 	if soap.IsSoapFault(err) {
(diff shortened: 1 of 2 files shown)


pkg/vsphere/tasks/waiter.go
```diff
@@ -70,7 +70,7 @@ func WaitForResult(ctx context.Context, f func(context.Context) (Task, error)) (
 			}
 		}
 
-		if !isTaskInProgress(err) {
+		if !isRetryErrors(err) {
```
hmahmood: `isRetryError`

pkg/vsphere/tasks/waiter.go
```diff
@@ -89,11 +89,26 @@ func WaitForResult(ctx context.Context, f func(context.Context) (Task, error)) (
...
+// The reason to retry for these errors is because sometimes vSphere needs longer time to finish sync up inside
+func isRetryErrors(err error) bool {
 	if soap.IsSoapFault(err) {
 		switch f := soap.ToSoapFault(err).VimFault().(type) {
 		case types.TaskInProgress:
 			return true
+		case *types.NetworkDisruptedAndConfigRolledBack:
+			logExpectedFault(soapFault, f)
+			return true
+		case *types.InvalidArgument:
```
hmahmood: `InvalidArgument` really something we can retry?

mhagen-vmware: I would guess this is because before the vsan converges, we are passing an invalid path to it.  Unfortunate return code in my opinion, but what we have to deal with...

vburenin: If it is an error that needs to be retried, it definitely needs a comment clarifying why.

emlin: The error thrown in the failed case is this error, with a invalid path message.
There is one vSAN bug talking about partition, and if there is host partition, vSAN files will be only accessible to one group, and not for others. And the bug is that even vSAN suggests host to start VM, DRS or PBM might still pick up the host has no access to data to power on VM.
I don't know how the partition happens, but if that's because of network congestion, it's possible in our env.
And because the failed VM is able to be powered on some time later, it's possible to be fixed by retry.
Another reason to retry this error is because in general, all VM configuration is specified by VIC without any user input, there should not have any invalidArgument error happens unless there is system issue.

cgtexmex:
The commit message should be updated -- "workaround two intermittent errors:" not a very informative one liner -- I think you can summarize the change better in the subject

pkg/vsphere/tasks/waiter.go
hmahmood: This sounds like a specific case for VSAN, and I think I have seen in the past where we have gotten this error because of a bug in generating the VM config. Which is why I think it is useful to not retry this error (otherwise we mask the bug), and handle this elsewhere within the context of VSAN.

emlin: hmm.. I don't think we're masking bugs. Even with retry, the bug will still fail after a few times retry.
No matter how, our bug needs to be fixed, no matter retry or not. But if we move the retry logic to somewhere else, that means anyone powerOn VM needs to check if that's run on vSAN, and needs to take care of the retry.
I personally dislike to spread the workaround code for vSAN bug, or even not a bug (just test env issue) in everywhere.
So if it doesn't hurt, I'm fine to have some retry workaround in one central place, or we might just need to justify this should be fixed in vSAN. 😄

PR #4648, opened 2017-04-07
[specific ci=6-07-Create-Network] Change FAILURE error msg with correct expected ERROR msg.

### LINKED ISSUE #4547: Test 6-07 create network: bridge network - vcenter none
```
    ${output}=  Run  bin/vic-machine-linux create --name=%{VCH-NAME} --target="%{TEST_USERNAME}:%{TEST_PASSWORD}@%{TEST_URL}" --thumbprint=%{TEST_THUMBPRINT} --image-store=%{TEST_DATASTORE} ${vicmachinetls}
    Should Contain  ${output}  FAILURE
```
failure does not exists here, should be looking for:
Apr 3 2017 17:20:37.205-05:00 ERROR An existing distributed port group must be specified for bridge network on vCenter: network 'VCH-0-7850' not found

rogeliosanchez:
Changed the expected check from a FAILURE message to an ERROR message.  We now look for "ERROR An existing distributed port group must be specified for bridge network on vCenter" instead of "FAILURE" in the output.

Fixes #4547

### DIFF
--- a/tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.robot
+++ b/tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.robot
@@ -224,7 +224,8 @@ Bridge network - vCenter none
     Run Keyword And Ignore Error  Cleanup Datastore On Test Server
 
     ${output}=  Run  bin/vic-machine-linux create --name=%{VCH-NAME} --target="%{TEST_USERNAME}:%{TEST_PASSWORD}@%{TEST_URL}" --thumbprint=%{TEST_THUMBPRINT} --image-store=%{TEST_DATASTORE} ${vicmachinetls}
-    Should Contain  ${output}  FAILURE
+    Should Contain  ${output}  ERROR
+    Should Contain  ${output}  An existing distributed port group must be specified for bridge network on vCenter
 
     # Delete the portgroup added by env vars keyword
     Cleanup VCH Bridge Network  %{VCH-NAME}


tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.robot
```diff
@@ -224,7 +224,8 @@ Bridge network - vCenter none
     Run Keyword And Ignore Error  Cleanup Datastore On Test Server
 
     ${output}=  Run  bin/vic-machine-linux create --name=%{VCH-NAME} --target="%{TEST_USERNAME}:%{TEST_PASSWORD}@%{TEST_URL}" --thumbprint=%{TEST_THUMBPRINT} --image-store=%{TEST_DATASTORE} ${vicmachinetls}
-    Should Contain  ${output}  FAILURE
+    Should Contain  ${output}  ERROR
+    Should Contain  ${output}  An existing distributed port group must be specified for bridge network on vCenter
```
mhagen-vmware: the vcenter here still has me concerned, does this work on ESX?  does it still report this message?  if so, we need a bug to fix the output, but either way you should cut the check off right before that in case it gets fixed.

rogeliosanchez: When running this on just an ESX, this test is skipped.  No checks happen here in that scenario.  When you throw in vcenter on top of an ESX, this test runs and then hits that check.  Is this what you are referring to @mhagen-vmware ?  

I'll look into the syntax/formatting for checking to see if an error occurs, then running this check if there is an error present, instead of just assuming there will always be an ERROR.

rajanashok:
lgtm

PR #4649, opened 2017-04-07
don't publish detach event on an attach and provide exit code with die event

### LINKED ISSUE #4561: docker client 1.13 always returns 0 with VIC
Steps to reproduce:
1. `docker run busybox false` should return 1, but returns 0

cgtexmex:
ContainerAttach was publishing a detach event on func exit.  This
resulted in the 1.13 client exiting docker run before container exit.
It will now only publish a detach event when a detach is detected.

Also exit code has been added to the die event, which will allow
the docker 1.13 client to return exit codes.

Fixes #4561

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -1589,13 +1589,13 @@ func (c *Container) containerAttach(name string, ca *backend.ContainerAttachConf
 
 	actor := CreateContainerEventActorWithAttributes(vc, map[string]string{})
 	EventService().Log(containerAttachEvent, eventtypes.ContainerEventType, actor)
-	defer func() {
-		actor := CreateContainerEventActorWithAttributes(vc, map[string]string{})
-		EventService().Log(containerDetachEvent, eventtypes.ContainerEventType, actor)
-	}()
 	err = c.containerProxy.AttachStreams(context.Background(), vc, clStdin, clStdout, clStderr, ca)
 	if err != nil {
 		if _, ok := err.(DetachError); ok {
+			// fire detach event
+			actor := CreateContainerEventActorWithAttributes(vc, map[string]string{})
+			EventService().Log(containerDetachEvent, eventtypes.ContainerEventType, actor)
+
 			log.Infof("Detach detected, tearing down connection")
 			client = c.containerProxy.Client()
 			handle, err = c.Handle(id, name)
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -39,6 +39,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
@@ -100,6 +101,7 @@ type VicContainerProxy interface {
 
 	Handle(id, name string) (string, error)
 	Client() *client.PortLayer
+	exitCode(vc *viccontainer.VicContainer) (string, error)
 }
 
 // ContainerProxy struct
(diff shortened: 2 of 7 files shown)


lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -135,9 +137,13 @@ const (
...
 )
 
+// used by other engine components
+var containerProxy *ContainerProxy
+
 // NewContainerProxy creates a new ContainerProxy
 func NewContainerProxy(plClient *client.PortLayer, portlayerAddr string, portlayerName string) *ContainerProxy {
-	return &ContainerProxy{client: plClient, portlayerAddr: portlayerAddr, portlayerName: portlayerName}
+	containerProxy = &ContainerProxy{client: plClient, portlayerAddr: portlayerAddr, portlayerName: portlayerName}
+	return containerProxy
```
caglar10ur: if we are going to use it like this, then I prefer the singleton method we use in tether

https://github.com/vmware/vic/blob/957f703a2ff47d5975a5282709bbdca62f363bbd/cmd/tether/attach.go#L39
https://github.com/vmware/vic/blob/957f703a2ff47d5975a5282709bbdca62f363bbd/cmd/tether/attach.go#L80

otherwise LGTM

lib/apiservers/engine/backends/container.go
```diff
@@ -1589,13 +1589,13 @@ func (c *Container) containerAttach(name string, ca *backend.ContainerAttachConf
...
 	actor := CreateContainerEventActorWithAttributes(vc, map[string]string{})
 	EventService().Log(containerAttachEvent, eventtypes.ContainerEventType, actor)
-	defer func() {
-		actor := CreateContainerEventActorWithAttributes(vc, map[string]string{})
-		EventService().Log(containerDetachEvent, eventtypes.ContainerEventType, actor)
-	}()
 	err = c.containerProxy.AttachStreams(context.Background(), vc, clStdin, clStdout, clStderr, ca)
 	if err != nil {
 		if _, ok := err.(DetachError); ok {
+			defer func() {
```
hickeng: I think it's actually viable to dispatch the event directly rather than as a `defer` - the remaining work has to occur to tidy up, but the detach has already happened from the client's point of view, as the AttachStreams have been closed.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -678,6 +687,36 @@ func (c *ContainerProxy) State(vc *viccontainer.VicContainer) (*types.ContainerS
 	return inspectJSON.State, nil
 }
 
+// exitCode returns container exitCode
+func (c *ContainerProxy) exitCode(vc *viccontainer.VicContainer) (string, error) {
```
hickeng: @caglar10ur I think this code is essentially TaskInspect - please make sure your branch is correctly populating the exitCode in the eventMonitor.

lib/apiservers/engine/backends/container.go
cgtexmex: Agreed...I just lazily did a cut and paste...I'll change when I rebase.

anchal-agrawal:
There are several integration tests which were modified in #4551 to pass w/o exit codes - we should address them before merging this. https://github.com/vmware/vic/blob/master/tests/test-cases/Group1-Docker-Commands/1-06-Docker-Run.robot#L86 is directly dependent on this PR's issue.

cgtexmex:
@anchal-agrawal ah...crap...I missed that had been merged...I'll fix'em..

cgtexmex:
@anchal-agrawal tests updated....

cc/ @rajanashok @mhagen-vmware

PR #4675, opened 2017-04-10
Inspecting a non-existent object now returns No such object

### LINKED ISSUE #4573: VIC with Docker client 1.13 returns incorrect error message on non-existing object
**VIC version:**

Server:
 Version:      0.8.0
 API version:  1.25 (minimum version 1.19)
 Go version:   go1.8
 Git commit:   866fae4
 Built:        Mon Apr  3 22:25:08 2017
 OS/Arch:      linux/amd64
 Experimental: true

**Steps to reproduce:**

Run the following docker command with docker 1.13 client:

docker inspect fake

where **fake** is non existent object

the output is:

_Error response from daemon: vSphere Integrated Containers does not yet support Docker Swarm_

**Actual behavior:**

Both the old and the new client connect to the server to look up different object types that can match **fake**, the last lookup is for a _service_ object. In both cases the VCH returns

_Error response from daemon: vSphere Integrated Containers does not yet support Docker Swarm_

The 1.12 client overrides the server message with the following:

_Error: No such image, container or task: fake_

The 1.13 client does not, therefore the incorrect server error message is returned.

The docker-ce client (Version 17.03.1-ce) against the VCH has a different behavior. In this case the lookup for a "service" is not executed.

**Expected behavior:**

When running against the Docker server with the 1.13 client the output of the command is:

_Error: No such object: fake_

**Logs:**

**This issue is currently blocking: https://github.com/vmware/vic/issues/4368**
[log bundle test 1.23.zip](https://github.com/vmware/vic/files/894865/log.bundle.test.1.23.zip)

cgtexmex:
Fixes #4573

### DIFF
--- a/lib/apiservers/engine/backends/errors.go
+++ b/lib/apiservers/engine/backends/errors.go
@@ -86,6 +86,14 @@ func ConflictError(msg string) error {
 	return derr.NewRequestConflictError(fmt.Errorf("Conflict error from portlayer: %s", msg))
 }
 
+func PluginNotFoundError(name string) error {
+	return derr.NewErrorWithStatusCode(fmt.Errorf("plugin %s not found", name), http.StatusNotFound)
+}
+
+func SwarmNotSupportedError() error {
+	return derr.NewErrorWithStatusCode(fmt.Errorf("swarm is not supported by %s", ProductName()), http.StatusNotFound)
+}
+
 // Error type check
 
 func IsNotFoundError(err error) bool {
--- a/lib/apiservers/engine/backends/plugins.go
+++ b/lib/apiservers/engine/backends/plugins.go
@@ -44,7 +44,7 @@ func (p *Plugin) List() ([]enginetypes.Plugin, error) {
 }
 
 func (p *Plugin) Inspect(name string) (*enginetypes.Plugin, error) {
-	return nil, nil
+	return nil, PluginNotFoundError(name)
 }
 
 func (p *Plugin) Remove(name string, config *enginetypes.PluginRmConfig) error {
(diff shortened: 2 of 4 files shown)


cgtexmex:
crap..this works with 1.13.1 but not 1.13.0...thanks docker... ;-/

chengwang86:
@cgtexmex Anchal and I find that in Docker 1.13.0, when `docker inspect` is called, it will inspect container, image, volume, service (https://github.com/vmware/vic/blob/master/lib/apiservers/engine/backends/swarm.go#L46). However, for later versions, it does not inspect service (swarm) but inspect plugins instead. 

So if we just want to be compatible with Docker 1.13, we should apply what you do in this PR to service inspect in `swarm.go`. If we also want to be compatible with later versions of Docker, we need to apply your changes to both plugin and service inspect.

cgtexmex:
Yep..that's what I saw as well and am testing now...

cgtexmex:
the client sends a few requests (node, task & service) -- I've added errors for those and this works w/client 1.13.0 and 1.13.1

lib/apiservers/engine/backends/errors.go
```diff
@@ -86,6 +86,14 @@ func ConflictError(msg string) error {
 	return derr.NewRequestConflictError(fmt.Errorf("Conflict error from portlayer: %s", msg))
 }
 
+func PluginNotFoundError(name string) error {
+	return derr.NewErrorWithStatusCode(fmt.Errorf("plugin %s not found", name), http.StatusNotFound)
+}
+
+func SwarmNotSupportedError() error {
+	return derr.NewErrorWithStatusCode(fmt.Errorf("Swarm is not supported by %s", ProductName()), http.StatusNotFound)
```
anchal-agrawal: golint nitpick: error strings should start with lowercase.

hickeng:
@cgtexmex @mhagen-vmware Is this another issue that's needed for basic 1.13 support or a nice to have?

mhagen-vmware:
@hickeng I have kind/noted the underlying bug, so I don't think this is a show stopper for 1.1

PR #4694, opened 2017-04-11
[specific ci=11-03-Upgrade-InsecureRegistry] improve integration test 11-03 to test against real registry

maplain:
local manual test verified
Fixes: #4403


tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
```diff
@@ -14,25 +14,85 @@
...
-    ${host}=  Run  govc vm.info -e -json %{VCH-NAME} | jq -c '.VirtualMachines[0].Config.ExtraConfig[] | select(.Key | . and contains("guestinfo.vice./registry/insecure_registries|0/Host")) .Value'
-    ${path}=  Run  govc vm.info -e -json %{VCH-NAME} | jq -c '.VirtualMachines[0].Config.ExtraConfig[] | select(.Key | . and contains("guestinfo.vice./registry/insecure_registries|0/Path")) .Value'
-    [Return]  ${host}  ${path}
+Setup Test Environment
+    [arguments]  ${insecure_registry}
+    ${handle}  ${docker_daemon_pid}=  Start Docker Daemon Locally  --insecure-registry ${insecure_registry}
+    Setup VCH And Registry  ${insecure_registry}
+    [Return]  ${handle}  ${docker_daemon_pid}
+
+Add Project On Registry
```
anchal-agrawal: We should use `[Tags]  secret` here.

maplain: 👌

tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
```diff
@@ -14,25 +14,87 @@
...
+    Should Be Equal As Integers  ${rc}  0
+    Should Contain  ${output}  Login Succeeded
+    ${rc}=  Run And Return Rc  docker -H ${default_local_docker_endpoint} push ${registry_ip}/test/busybox
+    Should Be Equal As Integers  ${rc}  0
+
+Test VCH And Registry
+    [Arguments]  ${vch_endpoint}  ${registry_ip}
+    ${rc}  ${output}=  Run And Return Rc And Output  docker -H ${vch_endpoint} pull ${registry_ip}/test/busybox
+    Should Be Equal As Integers  ${rc}  0
+    Log To Console  ${output}
```
anchal-agrawal: Please remove this as it adds verbosity to the test output. Instead, we can check for the expected output. See https://github.com/vmware/vic/blob/master/tests/resources/Docker-Util.robot#L34

maplain: sure

tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
```diff
@@ -14,25 +14,100 @@
...
+    [Arguments]  ${registry_ip}  ${protocol}  ${user}=admin  ${password}=%{TEST_PASSWORD}
+    # Harbor API: https://github.com/vmware/harbor/blob/master/docs/swagger.yaml
+    Run Keyword If  '${protocol}' == 'https'  Set Test Variable  ${curl_option}  --insecure
+    Run Keyword If  '${protocol}' == 'http'  Set Test Variable  ${curl_option}  ${EMPTY}
+    :FOR  ${i}  IN RANGE  12
+    \   Sleep  10s
+    \   ${rc}=  Run And Return Rc  curl ${curl_option} -u ${user}:${password} -H "Content-Type: application/json" -X POST -d '{"project_name": "test","public": 1}' ${protocol}://${registry_ip}/api/projects
+    \   Continue For Loop If  '${rc}' != '0'
+    \   Run Keyword If  '${rc}' == '0'  Set Test Variable  ${pass}  1
+    \   Exit For Loop If  '${rc}' == '0'
```
mhagen-vmware: Nice use of the API, but very unclear here - 
1. why is this in a for loop?  does the API curl not work sometimes?
2. regardless of the answer to 1. all this logic is confusing and unnecessary. these 3 lines and the one below just need to be return from keyword if, then it will continue to loop otherwise.

tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
```diff
@@ -14,25 +14,100 @@
...
+    ${rc}  ${output}=  Run And Return Rc And Output  echo "From busybox" | ${docker} -H ${default_local_docker_endpoint} build -t ${registry_ip}/test/busybox -
+    Should Be Equal As Integers  ${rc}  0
+    Should Contain  ${output}  Successfully built
+    Log To Console  \nbusybox built successfully
+    :FOR  ${i}  IN RANGE  6
+    \   ${rc}  ${output}=  Run And Return Rc And Output  ${docker} -H ${default_local_docker_endpoint} login --username ${registry_user} --password ${registry_password} ${registry_ip}
+    \   ${status}  ${msg}=  Run Keyword And Ignore Error  Should Contain  ${output}  Login Succeeded
+    \   Run Keyword If  '${rc}' == '0' and '${status}' != 'FAIL'  Set Test Variable  ${pass}  2
+    \   Exit For Loop If  '${rc}' == '0' and '${status}' != 'FAIL'
+    \   Sleep  10s
```
mhagen-vmware: again, weird for loop here... docker login should not fail and need to be retried

tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
```diff
@@ -14,25 +14,100 @@
...
+
+Cleanup Test Environment
+    [Arguments]  ${handle}  ${docker_daemon_pid}  ${harbor_name}  ${harbor_ip}  ${docker}=DOCKER_API_VERSION=1.23 docker
+    ${rc}=  Run And Return Rc  ${docker} -H ${default_local_docker_endpoint} rmi ${harbor_ip}/test/busybox
+    Should Be Equal As Integers  ${rc}  0
+    Kill Local Docker Daemon  ${handle}  ${docker_daemon_pid}
+    Cleanup Harbor  ${harbor_name}
+    Clean up VIC Appliance And Local Binary
+
+*** Variables ***
```
mhagen-vmware: Please put variables above keywords.

tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
```diff
@@ -14,25 +14,100 @@
...
+    \   Sleep  10s
+    \   ${rc}=  Run And Return Rc  curl ${curl_option} -u ${user}:${password} -H "Content-Type: application/json" -X POST -d '{"project_name": "test","public": 1}' ${protocol}://${registry_ip}/api/projects
+    \   Continue For Loop If  '${rc}' != '0'
+    \   Run Keyword If  '${rc}' == '0'  Set Test Variable  ${pass}  1
+    \   Exit For Loop If  '${rc}' == '0'
+    Should Be Equal As Integers  ${pass}  1
+
+Setup VCH And Registry
+    [Arguments]  ${registry_ip}  ${registry_user}=admin  ${registry_password}=%{TEST_PASSWORD}  ${docker}=DOCKER_API_VERSION=1.23 docker
+    ${rc}  ${output}=  Run And Return Rc And Output  echo "From busybox" | ${docker} -H ${default_local_docker_endpoint} build -t ${registry_ip}/test/busybox -
```
mhagen-vmware: where did this dockerfile come from? I don't see it anywhere.

tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
```diff
@@ -14,25 +14,100 @@
...
-    ${oldHost}  ${oldPath}=  Get host and path from guestinfo
-    Log  ${oldHost} ${oldPath}
-    Should Be Equal As Strings  ${oldHost}  "<nil>"
-    Should Be Equal As Strings  ${oldPath}  "10.10.10.10:1234"
+Upgrade VCH with Harbor On HTTP
+    ${harbor_ip}=  Install Harbor To Test Server  ${http_harbor_name}
+    Add Project On Registry  ${harbor_ip}  http
+    ${handle}  ${docker_daemon_pid}=  Setup Test Environment  ${harbor_ip}
+
+    Install VIC with version to Test Server  ${test_vic_version}  --insecure-registry ${harbor_ip} --no-tls --no-tlsverify
```
mhagen-vmware: both no-tls and no-tlsverify are not valid. either one or the other needs to be specified.

tests/resources/Harbor-Util.robot
```diff
@@ -15,50 +15,53 @@
...
+    ${out}=  Run  ovftool --noSSLVerify --acceptAllEulas --datastore=${datastore} --name=${name} --net:"Network 1"='${network}' --diskMode=thin --powerOn --X:waitForIp --X:injectOvfEnv --X:enableHiddenProperties --prop:root_pwd=${password} --prop:harbor_admin_password=${password} --prop:db_password=${db_password} --prop:auth_mode=db_auth --prop:verify_remote_cert=${verify} --prop:protocol=${protocol} ${HARBOR_VERSION}.ova 'vi://${user}:${password}@${host}'
     Should Contain  ${out}  Received IP address:
+    Should Not Contain  ${out}  None
 
     ${out}=  Split To Lines  ${out}
     :FOR  ${line}  IN  @{out}
     \   ${status}=  Run Keyword And Return Status  Should Contain  ${line}  Received IP address:
     \   ${ip}=  Run Keyword If  ${status}  Fetch From Right  ${line}  ${SPACE}
-    \   Run Keyword If  ${status}  Set Environment Variable  HARBOR_IP  ${ip}
+    \   Run Keyword If  ${status}  Set Test Variable  ${harbor-ip}  ${ip}
```
maplain: ok

tests/resources/Harbor-Util.robot
```diff
@@ -15,50 +15,53 @@
...
     \   Exit For Loop If  ${status}
 
     Log To Console  Waiting for Harbor to Come Up...
-    :FOR  ${i}  IN RANGE  10
-    \  ${out}=  Run  curl -k https://%{HARBOR_IP}
-    \  Log    ${out}
+    :FOR  ${i}  IN RANGE  40
```
maplain: yes, sometimes, ci could be very slow.. I'll change it back.

tests/resources/Harbor-Util.robot
```diff
@@ -15,50 +15,53 @@
...
+    \  ${out}=  Run  curl -k ${protocol}://${harbor-ip}
+    \  Log  ${out}
     \  ${status}=  Run Keyword And Return Status  Should Not Contain  ${out}  502 Bad Gateway
-    \  Return From Keyword If  ${status}
-    \  Sleep  30
-    Fail  Harbor failed to come up properly!
+    \  ${status}=  Run Keyword If  ${status}  Run Keyword And Return Status  Should Not Contain  ${out}  Connection refused
+    \  ${status}=  Run Keyword If  ${status}  Run Keyword And Return Status  Should Contain  ${out}  <title>Harbor</title>
+    \  Run Keyword If  ${status}  Set Test Variable  ${status}  ${status}
```
maplain: they're overwriting ${status} based on a few situations I observed in practice since 'should not contain  502 bad gateway' alone doesn't guarantee harbor is up.

tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
maplain: 👌 I'll remove no-tlsverify

tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
maplain: sure

tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
maplain: yes, this is weird to me too. I've seen it failed a few times though everything looks fine. after adding this, it passed.

tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
maplain: oh, it's on the fly. the dockerfile is "From busybox" and passed to docker build by pipe and "-"

tests/resources/Harbor-Util.robot
```diff
@@ -627,3 +630,19 @@ Basic Docker Command With Harbor
     Log  ${output}
     Should Be Equal As Integers  ${rc}  0
     Should Contain  ${output}  Untagged
+
+Fetch Harbor Self Signed Cert
+    [Tags]  secret
+    [Arguments]  ${harbor-ip}  ${user}=admin  ${password}=%{TEST_PASSWORD}
+    ${rc}=  Run And Return Rc  wget -q --tries=10 --connect-timeout=10 --auth-no-challenge --no-check-certificate --user ${user} --password ${password} https://${harbor-ip}/api/systeminfo/${harbor_cert}
```
maplain: it's sort of static because its value is given in ***Variables***. I make it a suite variable instead of hardcoding it to leave room for harbor changing api.

tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
mhagen-vmware: we need to create a bug in the harbor project for that then. And you can reference that bug with a github check then.

tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
mhagen-vmware: ah, gotcha. :+1:

tests/resources/Harbor-Util.robot
```diff
@@ -627,3 +630,19 @@ Basic Docker Command With Harbor
...
+    Should Be Equal As Integers  ${rc}  0
+    [Return]  ${harbor_cert}
```
maplain: like I said above, I think any test which is using this keyword should not be aware of either harbor api or name of this file. so it's passed to others by returning.

tests/resources/Harbor-Util.robot
maplain: I make it a test variable and return it at the end of this keyword because I feel it makes contracts between keywords clear. depending on environmental variables could be very confusing when they're used and changed everywhere. i spent lots of time on it. imho, it doesn't make things simple, but sort of force test writer to be familiar with every keyword that is going to be used to make sure environmental variable is correctly set.

maplain: test can always use '${harbor_ip}=  Install Harbor to Test server' to get its ip.

tests/resources/Harbor-Util.robot
```diff
@@ -15,50 +15,53 @@
...
     \  ${status}=  Run Keyword And Return Status  Should Not Contain  ${out}  502 Bad Gateway
-    \  Return From Keyword If  ${status}
-    \  Sleep  30
-    Fail  Harbor failed to come up properly!
+    \  ${status}=  Run Keyword If  ${status}  Run Keyword And Return Status  Should Not Contain  ${out}  Connection refused
+    \  ${status}=  Run Keyword If  ${status}  Run Keyword And Return Status  Should Contain  ${out}  <title>Harbor</title>
+    \  Run Keyword If  ${status}  Set Test Variable  ${status}  ${status}
+    \  Exit For Loop If  ${status}
+    \  Sleep  30s
+    Run Keyword If  not ${status}  Fail  Harbor failed to come up properly!
```
maplain: in that case, if ${status} were not correctly set during the whole for loop, there is no check for failure. and it returns before making sure harbor is really up.

tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
maplain: 1. yes, if practice, I've seen that even though harbor ui is up, curl may still fail, getting a 'connection refused' or something. i feel like this is because that part of harbor is initialized after ui;
2. like above, we cannot use return from keyword if because that assumes we can always succeed in this for loop. but sometimes, for example, ci server is really slow, it could use out of all loop without succeeding, in that case, test should fail here instead of silently going through every loop and returning.

maplain: oh, my bad, we can use 'Return From Keyword If' in the loop if the return value is set, and a 'Fail' following for loop.

mdubya66:
what's going on with this PR?

maplain:
@mdubya66 updated yesterday, waiting for comments.

maplain:
@mhagen-vmware can you take a look again? thx...

mhagen-vmware:
lgtm

PR #4707, opened 2017-04-11
Event based removal of containers and portmaps from persona cache

### LINKED ISSUE #4696: persona cache confusion due to OOB removal
Versions Impacted: All

Example:
Run and verify container with mapped ports:
```
docker run -d --name jojo -p 80:8080 busybox /bin/top

docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS                          NAMES
e3f0c148907a        busybox             "/bin/top"          12 seconds ago      Up 12 seconds       192.168.251.146:80->8080/tcp   jojo
```
Remove the container Out of Band and verify:
```
govc vm.destroy "jojo*"
docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS                          NAMES
```
Create a new container mapped to same port:
```
docker run -d --name sally -p 80:8080 busybox /bin/top
0dafb361efa1101e5ee6854d8d98447423de9d4f8b9f426a1e3125d86c8c692f
docker: Error response from daemon: Failed to get container "e3f0c148907a3601930145e3351fd4076ef8e6b3c605e83b41ec1cb95ecf0f03" power state: No such container: jojo.
```
The container is created, but the start will fail and the error message given relates to the previously removed container.  The error message is generated from a func that is attempting to unbind ports from stopped containers, but instead returns an error for the container that has been orphaned in the persona cache.

If the user removes the second container (`sally` in this example) then the portmap cache is _cleaned_ up but the original container (`jojo` in this example) still resides in the container cache.  If another container is attempted to be started w/that name it will fail:

```
docker run -d --name jojo busybox /bin/top
docker: Error response from daemon: Conflict. The name "jojo" is already in use by container 5dedd9fc336c55683b02d4d9fe03e726eaa0a191409d4c7a881f1881f0828d77. You have to remove (or rename) that container to be able to re use that name..
See 'docker run --help'.

docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS                          NAMES
```
The only way to clean up that container is to restart the appliance or do a force remove by name -- vic will throw an error, but you can then start a container by the same name:
```
docker rm -f jojo
Error response from daemon: No such container: jojo

docker run -d --name jojo busybox /bin/top
df122c926cc0266839ea7897890036120d3657ed4a8a323b20bd459af17323a8

docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS                          NAMES
df122c926cc0        busybox             "/bin/top"          52 seconds ago      Up 52 seconds                                      jojo
```

We do have an integration test that checks for [OOB removal](https://github.com/vmware/vic/blob/master/tests/test-cases/Group1-Docker-Commands/1-11-Docker-RM.robot#L86).  That test passes because the code has a bug that returns the expected error (`Error response from daemon: No such container: blah`), but the container has not been removed from the persona cache.

cgtexmex:
Fixes #4696

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -672,7 +672,11 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 	} else {
 		state, err := c.containerProxy.State(vc)
 		if err != nil {
-			return err
+			if IsNotFoundError(err) {
+				cache.ContainerCache().DeleteContainer(id)
+				return NotFoundError(name)
+			}
+			return InternalServerError(err.Error())
 		}
 		// force stop if container state is error to make sure container is deletable later
 		if state.Status == ContainerError {
@@ -696,22 +700,13 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 		}
 	}
 
-	// Unmap ports if the container was stopped out-of-band and then removed,
-	// in which case ports have been left mapped because the container is dropped
-	// from the persona cache before the mappings could be removed.
-	// If there's an error during unmap, don't fail the remove op.
-	if err = UnmapPorts(vc.HostConfig); err != nil {
-		log.Warn(err)
-	}
-
-	// delete container from the cache
-	cache.ContainerCache().DeleteContainer(id)
 	return nil
 }
 
 // cleanupPortBindings gets port bindings for the container and
 // unmaps ports if the cVM that previously bound them isn't powered on
 func (c *Container) cleanupPortBindings(vc *viccontainer.VicContainer) error {
+	defer trace.End(trace.Begin(vc.ContainerID))
 	for ctrPort, hostPorts := range vc.HostConfig.PortBindings {
 		for _, hostPort := range hostPorts {
 			hPort := hostPort.HostPort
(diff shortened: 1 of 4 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -671,7 +671,11 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 	} else {
 		state, err := c.containerProxy.State(vc)
 		if err != nil {
-			return err
+			if IsNotFoundError(err) {
+				cache.ContainerCache().DeleteContainer(id)
```
hmahmood: Should we unmap ports here as well?

cgtexmex: We could (and I believe there are similar calls elsewhere), but the reality is we'll get them w/events... _famous last words_

cgtexmex: That's kinda what I was getting at in my TODO line in eventmonitor.go....and I can take the time and do that now...but this is much better than what we had...

cgtexmex: FWIW (yes I'm verbose tonight) if you feel strongly about it...I'll close this PR and wrap it into the `autoremove` PR -- that's where became aware of the issue...

PR #4722, opened 2017-04-12
Only attempt to pull ESX log when it is an ESX

mhagen-vmware:


tests/resources/VCH-Util.robot
```diff
@@ -214,7 +214,7 @@ Gather Logs From Test Server
     Remove File  vic-admin-cookies
     ${out}=  Run  govc datastore.download %{VCH-NAME}/vmware.log %{VCH-NAME}-vmware.log
     Should Contain  ${out}  OK
-    Run  govc logs -log=vmkernel -n=10000 > vmkernel.log
+    Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Run  govc logs -log=vmkernel -n=10000 > vmkernel.log
```
dougm: Note that you can use the logs '-host' flag against VC and `govc find / -type h` to get the full list of hosts paths.  Works fine that way against ESX too.

PR #4483, opened 2017-03-30
Use ComputeResource as search root in vic-machine ls

dougm:
Using the ResourcePool as search root here does not work on VC,
due to VirtualApp being the immediate child.

Fixes #4482

PR #4345, opened 2017-03-22
[specific ci=Group6-VIC-Machine] Use Finder to search child ResourcePools

dougm:
Avoids 'vic-machine ls' reporting the same VCH twice.

Updated vic-machine finder_test to validate we don't find/list a VCH more than once.

Fixes #3975

PR #4752, opened 2017-04-12
vcsim: Add ContainerView support

dougm:


dougm:
This work had been started in support of #4345 #4483, but we decided to hold off on using ContainerView there.  Finished the vcsim portion over the weekend for testing the new govmomi/govc release.  We'll be making use of ContainerView at some point in VIC.

pkg/vsphere/simulator/folder.go
```diff
@@ -145,6 +145,35 @@ func (f *Folder) CreateFolder(c *types.CreateFolder) soap.HasFault {
 	return r
 }
 
+type StoragePod struct {
```
emlin: Might be helpful to have a comment for StoragePod here? you mentioned it is storage cluster in a test method. 😄

emlin:
lgtm

PR #4759, opened 2017-04-12
[skip ci] Enable promiscuous mode for ESX host portgroup

dougm:
This has happened to all users of these scripts at least once..
The host ESX system does not have promiscuous mode enabled by default.
The scripts run fine, but VMs on the nested ESX do not get an IP address.

Bump ESX/VCSA versions to 6.5.0b

### DIFF
--- a/infra/machines/vcsa/create-esxi-vm.sh
+++ b/infra/machines/vcsa/create-esxi-vm.sh
@@ -29,8 +29,8 @@ EOF
 
 disk=48
 mem=16
-# 6.5.0a - http://pubs.vmware.com/Release_Notes/en/vsphere/65/vsphere-esxi-650a-release-notes.html
-iso=VMware-VMvisor-6.5.0-4887370.x86_64.iso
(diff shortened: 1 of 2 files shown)


infra/machines/vcsa/create-esxi-vm.sh
```diff
@@ -29,8 +29,8 @@ EOF
 
 disk=48
 mem=16
-# 6.5.0a - http://pubs.vmware.com/Release_Notes/en/vsphere/65/vsphere-esxi-650a-release-notes.html
-iso=VMware-VMvisor-6.5.0-4887370.x86_64.iso
+# 6.5.0b - http://pubs.vmware.com/Release_Notes/en/vsphere/65/vsphere-vcenter-server-650b-release-notes.html
+iso=VMware-VMvisor-6.5.0-5146846.x86_64.iso
```
caglar10ur: looks like we released d today http://pubs.vmware.com/Release_Notes/en/vsphere/65/vsphere-vcenter-server-650d-release-notes.html :)

dougm: Yes indeed.  updated the vcsa script, but looks like ESX is still the 6.5b build

Issue #4683, opened 2017-04-11
vic-machine output and generated env is wrong if --cert-path specified


corrieb:
If you run vic-machine with --cert-path and ask it to generate certificates into the path specified, it does that fine. However, the .env file it creates and the helpful output produced by vic-machine does not take the --cert-path into account and shows the default path.

Example:

```
vic-machine-linux create --tls-cname *.eng.vmware.com --cert-path=/certs --thumbprint=F2:97:19:09:64:0A:06:3C:67:57:F7:CB:B5:3A:21:50:30:46:A0:54 --name=dev-cert --target=10.118.69.238 --user=Administrator@vsphere.local --image-store=iSCSI-nas-2 --password='xxxxxxxx' --bridge-network=dev-bridge --public-network=ExternalNetwork --compute-resource=iSCSI --timeout 60s --volume-store=iSCSI-nas-2/dev:default --container-network=ExternalNetwork
```

The above cmd-line generates certs into the right place, but the instructions from VIC machine and the output of the env file are wrong:

```
Apr 11 2017 16:00:22.309Z INFO  Docker environment variables:
Apr 11 2017 16:00:22.309Z INFO  DOCKER_TLS_VERIFY=1 DOCKER_CERT_PATH=/home/vagrant/vic/src/github.com/vmware/dev-release/vic/dev-cert DOCKER_HOST=office2-sfo2-dhcp335.eng.vmware.com:2376
Apr 11 2017 16:00:22.310Z INFO  
Apr 11 2017 16:00:22.310Z INFO  Environment saved in /certs/dev-cert.env

> cat /certs/dev-cert.env 
DOCKER_TLS_VERIFY=1 DOCKER_CERT_PATH=/home/vagrant/vic/src/github.com/vmware/dev-release/vic/dev-cert DOCKER_HOST=office2-sfo2-dhcp335.eng.vmware.com:2376
```

corrieb:
Assessment: Using certificates will be the default mode in production for companies and this will be an annoying and highly visible problem for them.

hickeng:
Looks like the specific file paths used are correct in `lib/install/management/inspect.go:GetDockerAPICommand` however the cert path option was never integrated into that logic.
This may be a little more involved to fix than might be assumed as this code is used for `vic-machine inspect` and leveraged in `vic-machine create` - inspect does not have a `--cert-path` option and shouldn't need one.

[...]

Issue #4490, opened 2017-03-31
Update vic-machine inspect with DOCKER_CERT_PATH for tlsverify configurations


hickeng:
**Story**
As a user of vic-machine I expect `vic-machine inspect` to provide the information I require to access a VCH

**Details**
Inspect is not returning a certificate path - this is because it's potentially hard to do so without some foreknowledge - however it's also not warning about the fact that necessary data is missing.
With the actual output you cannot connect to the docker API as you do not have the client cert.

Actual:
```
vagrant@devbox:~/vic$ vic-inspect
Mar 31 2017 14:21:10.319Z INFO  ### Inspecting VCH ####
Mar 31 2017 14:21:10.517Z INFO  Validating target
Mar 31 2017 14:21:10.675Z WARN  Disabling ha-host hostd.log collection (ServerFaultCode: Cannot complete the operation due to an incorrect request to the server.)
Mar 31 2017 14:21:10.676Z INFO
Mar 31 2017 14:21:10.676Z INFO  VCH ID: VirtualMachine:874
Mar 31 2017 14:21:10.691Z INFO
Mar 31 2017 14:21:10.691Z INFO  Installer version: v0.9.0-0-63b1d99
Mar 31 2017 14:21:10.692Z INFO  VCH version: v0.9.0-0-63b1d99
Mar 31 2017 14:21:10.692Z INFO
Mar 31 2017 14:21:10.692Z INFO  VCH upgrade status:
Mar 31 2017 14:21:10.692Z INFO  Installer has same version as VCH
Mar 31 2017 14:21:10.692Z INFO  No upgrade available with this installer version
Mar 31 2017 14:21:10.812Z INFO
Mar 31 2017 14:21:10.814Z INFO  VCH Admin Portal:
Mar 31 2017 14:21:10.814Z INFO  https://192.168.78.127:2378
Mar 31 2017 14:21:10.814Z INFO
Mar 31 2017 14:21:10.814Z INFO  Published ports can be reached at:
Mar 31 2017 14:21:10.814Z INFO  192.168.78.127
Mar 31 2017 14:21:10.825Z INFO
Mar 31 2017 14:21:10.825Z INFO  Docker environment variables:
Mar 31 2017 14:21:10.826Z INFO  DOCKER_TLS_VERIFY=1 DOCKER_HOST=192.168.78.127:2376
Mar 31 2017 14:21:10.826Z INFO
Mar 31 2017 14:21:10.826Z INFO  Connect to docker:
[...]

PR #4783, opened 2017-04-13
[specific ci=Group11-Upgrade] Correct certPath output on vic-machine create

gigawhitlocks:
This fixes the displayed certificate path after a create. 

@emlin do I need to update `inspect` to show this information as well, in order to close issue #4683 ? Or is `inspect`'s output fine? Or do we have another issue for updating `inspect`?

### DIFF
--- a/lib/install/management/inspect.go
+++ b/lib/install/management/inspect.go
@@ -89,11 +89,11 @@ func (d *Dispatcher) InspectVCH(vch *vm.VirtualMachine, conf *config.VirtualCont
 		log.Debugf("No host certificates provided")
 	}
 
-	d.ShowVCH(conf, "", "", "", "")
+	d.ShowVCH(conf, "", "", "", "", "")
 	return nil
 }
 
-func (d *Dispatcher) ShowVCH(conf *config.VirtualContainerHostConfigSpec, key string, cert string, cacert string, envfile string) {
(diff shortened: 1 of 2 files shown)


emlin:
@gigawhitlocks #4490 is for the inspect error

lib/install/management/inspect.go
```diff
@@ -147,7 +147,7 @@ func (d *Dispatcher) GetDockerAPICommand(conf *config.VirtualContainerHostConfig
 			}
 
 			dEnv = append(dEnv, "DOCKER_TLS_VERIFY=1")
-			info, err := os.Stat(conf.ExecutorConfig.Name)
+			info, err := os.Stat(certpath)
```
emlin: what happens if user didn't specify certpath, but used custom cert and key files?

gigawhitlocks: That's a good question. What is the correct output if they're in different locations?

emlin: I feels like user does not have to specify cert-path if custom key/cert files are specified, so here I think the output should allow empty certpath case.

gigawhitlocks: Looks like --key and --cert are only for input. --cert-path specifies input (if certs exist) AND output, so this change should work fine with --key and --cert

PR #4792, opened 2017-04-13
Fix incorrect port number for the attach test

caglar10ur:
-1 is not a valid port and it should 0 so that operating system can assign one.

### DIFF
--- a/lib/portlayer/attach/server.go
+++ b/lib/portlayer/attach/server.go
@@ -22,7 +22,6 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 
-	"github.com/vmware/vic/lib/portlayer/constants"
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/trace"
 )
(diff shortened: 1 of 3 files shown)


lib/portlayer/attach/server.go
```diff
@@ -54,6 +50,9 @@ func (n *Server) Start(debug bool) error {
 	log.Infof("Attach server listening on %s:%d", n.ip, n.port)
 
 	addr, err := net.ResolveTCPAddr("tcp", fmt.Sprintf("%s:%d", n.ip, n.port))
+	if err != nil {
+		return fmt.Errorf("Attach server error %s: %s", addr, errors.ErrorStack(err))
```
jakedsouza: addr will be nil, print n.ip and n.port

jakedsouza:
lgtm except for a nit comment

lib/portlayer/attach/server.go
caglar10ur: Done

PR #4815, opened 2017-04-17
[specific ci=1-06-Docker-Run] Add time accuracy check for docker run command in docker run test

### LINKED ISSUE #3710: Test for relative accuracy in docker run tests
In 1-06-Docker-Run:
Docker run verify container start and stop time

We should try and check additionally that we are at least within the ballpark of accuracy. See https://github.com/vmware/vic/pull/3690

I like it, but can we do a bit better? Something like:
containerStart-cmdStart <= 100 ( or something reasonably close, I don't know what the units here are - I would say it should be within 20 or 30 seconds )

maybe:
containerStop - cmdStart <= 60 or 70 seconds

rogeliosanchez:
Added an extra check on the docker run command.
We first check to make sure that the start time stamp actually happened before the stop time.  After that, we make sure that the simple start/stop docker command runs within 60 seconds (60,000 milliseconds).

Fixes #3710

### DIFF
--- a/tests/test-cases/Group1-Docker-Commands/1-06-Docker-Run.robot
+++ b/tests/test-cases/Group1-Docker-Commands/1-06-Docker-Run.robot
@@ -125,6 +125,8 @@ Docker run verify container start and stop time
     Run Keyword Unless  ${startStatus}  Fail  container start time before command start
     ${stopStatus}=  Run Keyword And Return Status  Should Be True  ${cmdStart} < ${containerStop}
     Run Keyword Unless  ${stopStatus}  Fail  container stop time before command start
+    ${timeDiff}=  Evaluate  ${containerStop}-${cmdStart}
+    Should Be True  0 < ${timeDiff} < 60000
 
 Docker run verify name and id are not conflated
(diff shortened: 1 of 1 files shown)


rajanashok:
lgtm

tests/test-cases/Group1-Docker-Commands/1-06-Docker-Run.robot
```diff
@@ -125,6 +125,9 @@ Docker run verify container start and stop time
     Run Keyword Unless  ${startStatus}  Fail  container start time before command start
     ${stopStatus}=  Run Keyword And Return Status  Should Be True  ${cmdStart} < ${containerStop}
     Run Keyword Unless  ${stopStatus}  Fail  container stop time before command start
+    ${timeDiff}=  Evaluate  ${containerStop}-${cmdStart}
+    ${timeLimit}=  Run Keyword And Return Status  Should Be True  0 < ${timeDiff} < 60000
+    Run Keyword Unless  ${timeLimit}  Fail  start to stop time limit reached, over 60 seconds
```
mhagen-vmware: This looks good, just a bit redundant I think, which adds complexity in reading it. You can really just rely on the Should be True keyword and get rid of the entire 3rd line and half the second line.

Issue #2802, opened 2016-10-19
vic-init fails with 4 NICs configured


hickeng:
The output in #2794 is caused by vic-init exiting when 4 separate NICs are specified. The installer generates the wrong slot number for the 4th NIC and vic-init does not handle this well:

```
DEBU[2016-10-19T16:36:40Z] Setting "bridge" to slot 288
```

```
ethernet3.pciSlotNumber = "1184"
```

We should not be exiting vic-init if/when this occurs - while rebooting the VM can redress some issues, a hardware misconfiguration is not one of them. Without vic-init running we cannot enable SSH or console for manual debugging.

vic-init output for reference:

```
time="2016-10-19T16:36:47Z" level=info msg="DHCP response: IP=192.168.78.152, SubnetMask=ffffff00, Gateway=192.168.78.2, DNS=[192.168.78.2], Lease Time=30m0s"
time="2016-10-19T16:36:47Z" level=info msg="setting ip address 192.168.78.152/24 for link eth3"
time="2016-10-19T16:36:47Z" level=debug msg="added address 192.168.78.152/24 to link eth3"
time="2016-10-19T16:36:47Z" level=debug msg="updateEndpoint(192.168.78.152/24, &{Common:{ExecutionEnvironment: ID:256 Name: Notes:} Static:false IP:<nil> Assigned:{IP:<nil> Mask:<nil>} Network:{Common:{ExecutionEnvironment: ID:Network:HaNetwork-VM Network Name:management Notes:} Type: Gateway:{IP:<nil> Mask:<nil>} Default:false Nameservers:[] Pools:[] Aliases:[]} DHCP:0xc4203f4880})"
time="2016-10-19T16:36:47Z" level=debug msg="not setting route for network: default=false gateway=192.168.78.2"
time="2016-10-19T16:36:47Z" level=debug msg="&{Common:{ExecutionEnvironment: ID:256 Name: Notes:} Static:false IP:<nil> Assigned:{IP:192.168.78.152 Mask:ffffff00} Network:{Common:{ExecutionEnvironment: ID:Network:HaNetwork-VM Network Name:management Notes:} Type: Gateway:{IP:192.168.78.2 Mask:ffffff00} Default:false Nameservers:[192.168.78.2] Pools:[] Aliases:[]} DHCP:0xc4203f4880}"
[...]

Issue #4613, opened 2017-04-06
docker run -it doesn't work. container will stuck in starting mode.


malikkal:
docker run without -d doesn't work. Engine version: v0.8.0-7315-c8ac999

Steps to reproduce.

Deploy a VCH with separate management and public network and with or without additional container networks.
Run a container out of the registry with -it. In this case a Redhat 6.8 base image.
Generic error Error response from daemon: Server error from portlayer: unable to wait for process launch status
Container will be stuck at 'Starting' when issuing a ps -a

However, the container will run without issues when deployed via Admiral or with -itd.

In summary no interactive session with container possible.

Also refer #4212, #3315 and #4223. (Similar issues)

VCH routes for reference.

![image](https://cloud.githubusercontent.com/assets/16984768/24751782/7265f5ae-1afd-11e7-948c-43b7f7595a9a.png)

malikkal:
This is resolved. probably a bug?

Enable mode 2 for rp_filter in the VCH host and make it permanent.
echo 2 > /proc/sys/net/ipv4/conf/default/rp_filter
sysctl -w "net.ipv4.conf.all.rp_filter=2"

For details refer:

http://www.slashroot.in/linux-kernel-rpfilter-settings-reverse-path-filtering

hickeng:
@malikkal given the symptoms I assume this is related to setup of the network serial port connection between ESX host and the endpointVM (`protocol:tcp src:esxhost dest:endpoint.management.net:2377`).
Does your comment mean that:
* the `esxhost` is not routable from the endpointVM's management interface (I presume `eth0` as we do not rename the management interface), or
* the underlying routing from esx to endpointVM means the packets are coming in on a different (I would guess `client`)

Altering the reverse packet filter opens up the possibility of spoofing, with someone pretending to be an ESX host creating a backchannel. While this should be handled at multiple levels (#2849) it's a weakening of the default configuration. I'd like to understand the specific issue you're seeing as it's unexpected if there's a separate management network.

If it's the first option I'm hoping this can be addressed by adding a more permissive route to the management interface, with `rp_filter=2` being a fallback approach.

hickeng:
For implementing a global `rp_filter` change - update: https://github.com/vmware/vic/blob/master/isos/appliance/nat-setup#L30

To address more permissive routes we would need to:
a. require that the management network config have a netmask that's broad enough to encompass the ESX hosts as well as the vSphere endpoing, or
[...]

PR #4009, opened 2017-02-22
[specific ci=6-07-Create-Network] Add bridge to public network ping test

rajanashok:
Adding test to verify if ping works from a bridge to a public network.

PR #4816, opened 2017-04-18
[full ci] Marking and routing traffic from bridge to public interface only

### LINKED ISSUE #3970: Containers have access to Management assets
The [documentation](https://vmware.github.io/vic-product/assets/files/html/0.8/vic_installation/networks.html#management) for VCH network types asserts that the VCH Management network is a sensitive resource requiring some degree of protection:

> always use a secure network

> use separate networks for the management network and the container networks

It comes as some surprise then that containers attached to "bridge" are allowed to NAT _through_ the VCH and have full access to assets on the Management and/or Client networks, or reachable via the gateway on those networks.

Any container can speak to vSphere assets. When it does, it speaks with the IP address of the VCH, making the (unwanted?) traffic impossible to filter.

I understand this is a non-trivial problem: The VCH needs to be able to reach all of these subnets and since the VCH _also_ performs IP transit (routing) duty for containers, this is a non-trivial problem.

I'd submit that the best solution to the issue is separating the various VCH functions (docker API listener, vSphere client, container gateway) into independent network namespaces so that their routing tables aren't co-mingled.

So... Is the masquerading of containers onto these other networks by design? Is it reasonable to rely on (containerize some software which acts as a vSphere or docker-engine client) this feature?

hmahmood:
Fixes #3970 #4613

### DIFF
--- a/cmd/vic-init/ops_linux.go
+++ b/cmd/vic-init/ops_linux.go
@@ -1,0 +1,60 @@
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/docker/libnetwork/iptables"
+	"github.com/vishvananda/netlink"
+
+	"github.com/vmware/vic/lib/tether"
+)
+
+const (
+	publicIfaceName = "public"
+)
+
+func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
+	// get the public interface name
+	l, err := netlink.LinkByName(publicIfaceName)
+	if l == nil {
+		l, err = netlink.LinkByAlias(publicIfaceName)
+		if l == nil {
+			return fmt.Errorf("could not find interface: %s", publicIfaceName)
+		}
+	}
+
+	if _, err = iptables.Raw(string(iptables.Append), "FORWARD", "-i", "bridge", "-o", l.Attrs().Name, "-j", "ACCEPT"); err != nil {
+		return err
+	}
+
+	if config.AsymmetricRouting {
+		// set rp_filter to "loose" mode; see https://en.wikipedia.org/wiki/Reverse_path_forwarding#Loose_mode
+		//
+		// this is so the kernel will not drop packets sent to the public interface from an
+		// address that is reachable by other interfaces on the VCH. specifically, when
+		// packets from the bridge network are directed to another network (say, management),
+		// the incoming reply to the VCH can be dropped if rp_filter is set to the default 1
+		if err = ioutil.WriteFile(fmt.Sprintf("/proc/sys/net/ipv4/conf/%s/rp_filter", l.Attrs().Name), []byte("2"), 0644); err != nil {
+			// not a fatal error, so just log it here
+			log.Warnf("error while setting rp_filter for interface %s: %s", l.Attrs().Name, err)
+		}
+	}
+
+	return nil
+}
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -466,6 +466,12 @@ func (c *Create) Flags() []cli.Flag {
 			Usage:       "Time to wait for create",
 			Destination: &c.Timeout,
 		},
+		cli.BoolFlag{
+			Name:        "asymmetric-routes, ar",
+			Usage:       "Set up the Virtual Container Host for asymmetric routing",
+			Destination: &c.AsymmetricRouting,
+			Hidden:      true,
+		},
 	}
 
 	help := []cli.Flag{
--- a/isos/appliance/nat-setup
+++ b/isos/appliance/nat-setup
@@ -39,8 +39,21 @@ echo 2 > /proc/sys/net/ipv4/ip_dynaddr
 # too many routers are still ignorant
 echo 0 > /proc/sys/net/ipv4/tcp_ecn
 
+# add a routing table for traffic from bridge that is not destined
+# for a local address
+echo 201 bridge.out >> /etc/iproute2/rt_tables
+ip rule add fwmark 1 table bridge.out
+
 # blow away any existing rules with iptables-restore
 iptables-restore <<RULES
+*mangle
+:PREROUTING ACCEPT [0:0]
+:INPUT ACCEPT [0:0]
+:FORWARD ACCEPT [0:0]
+:OUTPUT ACCEPT [0:0]
+:POSTROUTING ACCEPT [0:0]
+-A PREROUTING -i bridge -m addrtype ! --dst-type LOCAL -j MARK --set-xmark 0x1/0xffffffff
+COMMIT
 *nat
 :PREROUTING ACCEPT [0:0]
 :INPUT ACCEPT [0:0]
--- a/lib/tether/ops_linux.go
+++ b/lib/tether/ops_linux.go
@@ -433,7 +433,7 @@ func updateDefaultRoute(t Netlink, link netlink.Link, endpoint *NetworkEndpoint)
 	// delete the default route for the bridge.out table, if it exists
 	bTablePresent := bridgeTableExists(t)
 	if bTablePresent {
-		if err := t.RouteDel(&netlink.Route{LinkIndex: link.Attrs().Index, Dst: defaultNet, Table: 201}); err != nil {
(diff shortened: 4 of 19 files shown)


lib/tether/ops_linux.go
```diff
@@ -57,6 +57,7 @@ const (
 	pciDevPath         = "/sys/bus/pci/devices"
 	nfsFileSystemType  = "nfs"
 	ext4FileSystemType = "ext4"
+	bridgeTableNumber  = 201
```
mlh78750: Is this truly a constant? or should is there a way to look up what the table number is? I'm just concerned that down the road linux renumbers these things by default and we should think about defensive code for that.

hmahmood:
```
$ more /etc/iproute2/rt_tables 
#
# reserved values
#
255	local
254	main
253	default
0	unspec
#
# local
#
#1	inr.ruhep
```
As long as we don't use the reserved values, and no one else inserts another table with the same id (low chance of that happening since we own the VCH), we should be fine.

cmd/vic-init/ops_linux.go
```diff
@@ -0,0 +1,57 @@
...
+		return err
+	}
+
+	// set rp_filter to "loose" mode; see https://en.wikipedia.org/wiki/Reverse_path_forwarding#Loose_mode
+	//
+	// this is so the kernel will not drop packets sent to the public interface from an
+	// address that is reachable by other interfaces on the VCH. specifically, when
+	// packets from the bridge network are directed to another network (say, management),
+	// the incoming reply to the VCH can be dropped if rp_filter is set to the default 1
+	if err = ioutil.WriteFile(fmt.Sprintf("/proc/sys/net/ipv4/conf/%s/rp_filter", l.Attrs().Name), []byte("2"), 0644); err != nil {
```
hickeng: Can we enable this with a asymmetric-routing flag - it's basically disabling spoofing protection if we set this by default.
This will also address #4613 - but in many cases I think setting appropriate additional gateway elements on the management network would address this.

hickeng: If we fix #2802 and routing so that we can have the 4 NICs irrespective of sharing then we don't need this in code at all - it can go into `isos/appliance/nat-setup`

tests/test-cases/Group1-Docker-Commands/1-17-Docker-Network-Connect.robot
```diff
@@ -99,7 +99,7 @@ Connect containers to multiple networks non-overlapping
 
     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run --net cross2-network2 --name cross2-container2 debian ping -c2 ${ip}
     Should Not Be Equal As Integers  ${rc}  0
-    Should contain  ${output}  Destination Port Unreachable
+    Should Contain  ${output}  2 packets transmitted, 0 packets received, 100% packet loss
```
hickeng: Very happy with this test of isolation - can we also have a test that ensures we get the same response when trying to connect to management-network assets - that was the original issue and I'm not seeing a test that checks that isolation.

IIRC @mreferre also had a request that containers be able to communicate with one another via the port fowarding on the public interface. While it's a nasty indirect path, that is the public network access so if someone passes it to a container it should be accessible. At least we should have a test that asserts current behaviour and doc it appropriately.

lib/tether/ops_linux.go
```diff
@@ -409,12 +430,29 @@ func updateDefaultRoute(t Netlink, link netlink.Link, endpoint *NetworkEndpoint)
 		}
 	}
 
+	// delete the default route for the bridge.out table, if it exists
+	bTablePresent := bridgeTableExists(t)
+	if bTablePresent {
+		if err := t.RouteDel(&netlink.Route{LinkIndex: link.Attrs().Index, Dst: defaultNet, Table: 201}); err != nil {
```
caglar10ur: s/201/bridgeTableNumber/g

lib/tether/ops_linux.go
```diff
@@ -409,12 +430,29 @@ func updateDefaultRoute(t Netlink, link netlink.Link, endpoint *NetworkEndpoint)
...
+	}
+
 	log.Infof("Setting default gateway to %s", gw.IP)
 	route := &netlink.Route{LinkIndex: link.Attrs().Index, Dst: defaultNet, Gw: gw.IP}
 	if err := t.RouteAdd(route); err != nil {
 		return fmt.Errorf("failed to add gateway route for endpoint %s: %s", endpoint.Network.Name, err)
 	}
 
+	if bTablePresent {
+		route = &netlink.Route{LinkIndex: link.Attrs().Index, Dst: defaultNet, Gw: gw.IP, Table: 201}
```
caglar10ur: s/201/bridgeTableNumber/g

lib/tether/ops_linux.go
```diff
@@ -393,6 +399,21 @@ func updateRoutes(t Netlink, link netlink.Link, endpoint *NetworkEndpoint) error
 	return nil
 }
 
+func bridgeTableExists(t Netlink) bool {
```
hickeng: Is the bridgeTable primarily being added to support changes to default routes?

caglar10ur:
LGTM

tests/test-cases/Group1-Docker-Commands/1-17-Docker-Network-Connect.robot
hmahmood: We need two tests as you have pointed out before:
1. verify that bridge to management traffic is blocked
2. verify that bridge to management traffic is routed to the public network

For both these tests we need two port groups at least. Currently the CI only uses one port group, and I am not sure how to use another one. We most likely would add this to the nightly test run where the environment is more flexible. @rajanashok @mhagen-vmware

cmd/vic-init/ops_linux.go
hmahmood: I did think about putting this in `nat-setup` but I really want to restrict this only on the interface where it is needed. 

Can you clarify what you mean by `asymmetric-routing flag`? There is still some level of spoofing protection with a `2` setting; a `0` would have completely turned this off.

It sounds like #4613 should be fixed by having appropriate routes and not by changing `rp_filter` for the management interface. I am a bit wary of using this to fix other problems; in this case, there was really no other way to fix the issue.

lib/tether/ops_linux.go
hmahmood: It is required here so that we force redirect traffic that would normally be routed to one interface (management in this case), to another (public). A custom routing table seems to be the standard way to do this.

tests/test-cases/Group1-Docker-Commands/1-17-Docker-Network-Connect.robot
hmahmood: Looking at #4009 again, I see the test is not correct. I will add one to this PR.

isos/appliance/nat-setup
```diff
@@ -39,8 +39,21 @@ echo 2 > /proc/sys/net/ipv4/ip_dynaddr
...
+
 # blow away any existing rules with iptables-restore
 iptables-restore <<RULES
+*mangle
+:PREROUTING ACCEPT [0:0]
+:INPUT ACCEPT [0:0]
+:FORWARD ACCEPT [0:0]
+:OUTPUT ACCEPT [0:0]
+:POSTROUTING ACCEPT [0:0]
+-A PREROUTING -i bridge -m addrtype ! --dst-type LOCAL -j MARK --set-xmark 0x1/0xffffffff
```
fdawg4l: What's the mark used for?  I don't see anything looking for this mark in this `restore` script.

cmd/vic-init/ops_linux.go
hickeng: I mean a configuration flag/option of some kind that allows a viadmin to enable rp_filter=2 - there are some configurations where it just _has_ to be done, but I'd prefer not to have it be the default if possible. It may be that the UX is just sufficiently problematic that `2` is a desirable default but I'd prefer to keep the secure-by-default stance.

isos/appliance/nat-setup
hmahmood: Mark is used above to make packets go through the bridge.out table. See line 45 above.

fdawg4l: Didn't know you could do that.  Very cool!

cmd/vic-machine/create/create.go
```diff
@@ -466,6 +466,12 @@ func (c *Create) Flags() []cli.Flag {
 			Usage:       "Time to wait for create",
 			Destination: &c.Timeout,
 		},
+		cli.BoolFlag{
+			Name:        "asymmetric-routes, ar",
```
andrewtchin: Do we need a doc to explain when to use this?

hmahmood: yes, yes, we do

cmd/vic-machine/create/create.go
```diff
@@ -466,6 +466,12 @@ func (c *Create) Flags() []cli.Flag {
...
+			Usage:       "Set up the Virtual Container Host for asymmetric routing",
```
emlin: well, is it possible to add more description here for what it means

hmahmood: I know its cryptic, but I think adding more here will just add more confusion. @hickeng has already requested @stuclem document this in https://github.com/vmware/vic/issues/4613#issuecomment-296783581, which is where most of the detail would go.

hmahmood:
@hickeng @mhagen-vmware https://github.com/vmware/vic/issues/4865 for tracking the testing work.

PR #4824, opened 2017-04-18
Retry tether DHCP client ops when syscall is interrupted

### LINKED ISSUE #4558: Container run failure with "unable to wait for process launch status: summary.runtime.powerState=poweredOff"
On my new ESXi host, I quite regularly get a container start failure with the above message. I suspect it's timing related because it seems to happen randomly and I haven't yet seen it on my much slower VC cluster. That's not to say it's definitively an ESXi-only problem.

Port layer log file shows the following, which seems to suggest the error occurs before the container is even started. Maybe we're polling for extra config from the host that isn't ready yet, so rather than retrying, we just immediately fail?

Verified with build 9394

```
Apr  4 2017 20:58:39.928Z DEBUG [BEGIN] [github.com/vmware/vic/lib/portlayer/exec.(*containerBase).refresh:102] af2ff34eb778ef413a0f91820fc5771516d4ac0fd7367c5cf090c7c5c7a68364
Apr  4 2017 20:58:39.928Z DEBUG [BEGIN] [github.com/vmware/vic/lib/portlayer/exec.(*containerBase).updates:117] af2ff34eb778ef413a0f91820fc5771516d4ac0fd7367c5cf090c7c5c7a68364
Apr  4 2017 20:58:39.928Z DEBUG get vm properties [config runtime] of vm VirtualMachine:224
Apr  4 2017 20:58:39.928Z DEBUG properties: [config runtime summary.runtime.connectionState]
Apr  4 2017 20:58:39.946Z ERROR Unable to wait for extra config property guestinfo.vice..sessions|af2ff34eb778ef413a0f91820fc5771516d4ac0fd7367c5cf090c7c5c7a68364.started: summary.runtime.powerState=poweredOff
Apr  4 2017 20:58:39.946Z DEBUG [ END ] [github.com/vmware/vic/lib/portlayer/exec.(*containerBase).start:181] [3.407822254s] af2ff34eb778ef413a0f91820fc5771516d4ac0fd7367c5cf090c7c5c7a68364
Apr  4 2017 20:58:39.946Z DEBUG [ END ] [github.com/vmware/vic/lib/portlayer/exec.(*Container).start:336] [3.407913552s] af2ff34eb778ef413a0f91820fc5771516d4ac0fd7367c5cf090c7c5c7a68364
```

dougm:
Retry logic was already in place for EAGAIN, do the same for EINTR.

- Add logging when we hit one of those cases

- Improve error message when VM powers off before started flag is set

- Remove dead code (WaitForMac function)

Fixes #4558

### DIFF
--- a/lib/dhcp/client/client.go
+++ b/lib/dhcp/client/client.go
@@ -85,17 +85,36 @@ func (c *client) SetTimeout(t time.Duration) {
 	c.timeout = t
 }
 
-func withRetry(op func() error) error {
+// Note that the Go runtime sets SA_RESTART for syscalls which retries them automatically if they interrupted.
+// However, from the signal man page:
+//
+//      The following interfaces are never restarted after being interrupted
+//      by a signal handler, regardless of the use of SA_RESTART; they always
+//      fail with the error EINTR when interrupted by a signal handler:
+//
+//      * "Input" socket interfaces, when a timeout (SO_RCVTIMEO) has been
+//      set on the socket using setsockopt(2): accept(2), recv(2),
+//      recvfrom(2), recvmmsg(2) (also with a non-NULL timeout argument),
+//      and recvmsg(2).
+//
+//      * "Output" socket interfaces, when a timeout (SO_RCVTIMEO) has been
+//      set on the socket using setsockopt(2): connect(2), send(2),
+//      sendto(2), and sendmsg(2).
+func withRetry(name string, op func() error) error {
 	defer trace.End(trace.Begin(""))
 
 	for {
 		if err := op(); err != nil {
-			if errno, ok := err.(syscall.Errno); !ok || errno != syscall.EAGAIN {
-				return err
+			if errno, ok := err.(syscall.Errno); ok {
+				if errno == syscall.EAGAIN || errno == syscall.EINTR {
+					log.Debugf("retrying %q: errno=%d, error=%s", name, errno, err)
+					continue
+				}
 			}
-		} else {
-			return nil
+			return err
 		}
+
+		return nil
 	}
 }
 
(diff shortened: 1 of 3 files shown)


lib/dhcp/client/client.go
```diff
@@ -85,17 +85,21 @@ func (c *client) SetTimeout(t time.Duration) {
...
 
-func withRetry(op func() error) error {
+func withRetry(name string, op func() error) error {
 	defer trace.End(trace.Begin(""))
 
 	for {
 		if err := op(); err != nil {
-			if errno, ok := err.(syscall.Errno); !ok || errno != syscall.EAGAIN {
-				return err
+			if errno, ok := err.(syscall.Errno); ok {
```
caglar10ur: not that important but a note would be nice here saying that go runtime sets SA_RESTART for syscalls and retires them automatically if they interrupted.

I believe the reason we need to do this here because we set a timeout on the socket and signal man page (http://man7.org/linux/man-pages/man7/signal.7.html) says following;

```
 The following interfaces are never restarted after being interrupted
       by a signal handler, regardless of the use of SA_RESTART; they always
       fail with the error EINTR when interrupted by a signal handler:

       * "Input" socket interfaces, when a timeout (SO_RCVTIMEO) has been
         set on the socket using setsockopt(2): accept(2), recv(2),
         recvfrom(2), recvmmsg(2) (also with a non-NULL timeout argument),
         and recvmsg(2).

       * "Output" socket interfaces, when a timeout (SO_RCVTIMEO) has been
         set on the socket using setsockopt(2): connect(2), send(2),
         sendto(2), and sendmsg(2).
...
```

Issue #3434, opened 2016-12-02
Docker ps displays "poweredoff" container on build 7299


chengwang86:
**VIC version:**

 v0.8.0-rc3-7299-5373289

**Deployment details:**

ci build 7299 running on local ESXi host. 

**Steps to reproduce:**

I cannot reproduce it on my local setup by running the test steps manually.

**Expected behavior:**

In "Test-Cases.Group1-Docker-Commands.1-10-Docker-PS", we started a container running "ls" and expected "docker ps" would not display this container, since it should have exited after "ls".

**Actual behavior:**

"docker ps" displayed the information of the container running "ls", even if we did "docker ps" after "Wait Until VM Powers Off". The power state information was obtained by using govc, the output of which indicated that the VM had been powered off before "docker ps".

**Logs:**
[Test-Cases.Group1-Docker-Commands.1-10-Docker-PS-VCH-7299-7088-container-logs.zip](https://github.com/vmware/vic/files/628096/Test-Cases.Group1-Docker-Commands.1-10-Docker-PS-VCH-7299-7088-container-logs.zip)

Error:
```
Full Name:	Test-Cases.Group1-Docker-Commands.1-10-Docker-PS.Docker ps only running containers
Start / End / Elapsed:	20161202 10:02:31.810 / 20161202 10:03:06.693 / 00:00:34.883
Status:	FAIL (critical)
Message:	Length of '[
u'CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES', 
u'ffad9b735eaa        busybox             "/bin/top"          3 minutes ago       Up 11 seconds                           cocky_franklin', 
u'd95aaea998e4        busybox             "ls"                4 minutes ago       Up 24 seconds                           serene_euclid
']' should be 2 but is 3.
```

From hostd.log
```
2016-12-02T15:59:12.244Z info hostd[5DCC2B70] [Originator@6876 sub=Vmsvc.vm:/vmfs/volumes/577fceda-9b1fac4c-110b-0050569ccda1/serene_euclid-d95aaea998e492179b3dd012e55f665cdaa7be72acb5a8165ec7122431621634/serene_euclid-d95aaea998e492179b3dd012e55f665cdaa7be72acb5a8165ec7122431621634.vmx] State Transition (VM_STATE_ON -> VM_STATE_OFF)
[...]

PR #4825, opened 2017-04-18
Fix error messages seen with inspect [full ci]

### LINKED ISSUE #4573: VIC with Docker client 1.13 returns incorrect error message on non-existing object

jzt:
Credit goes to @cgtexmex for these changes. This PR basically un-reverts #4675. Tests pass locally, so I'm running a full ci here to see how it goes. Fixing the `Inspect()` error message in `swarm.go` changes some of the `docker service` and `docker node` errors from 'swarm not supported` to `No such servce/node`, so I've updated the tests around those to expect the proper output. Since our lack of swarm support is widely documented, this seemed like a decent trade-off.

Fixes #4573


caglar10ur:
LGTM

tests/test-cases/Group1-Docker-Commands/1-33-Docker-Service.robot
```diff
@@ -42,12 +42,12 @@ Docker serivce rm
 Docker service scale
     ${rc}  ${output}=  Run And Return Rc And Output  docker1.13 %{VCH-PARAMS} service scale test-service=3
```
mhagen-vmware: =3? should that be there?

tests/test-cases/Group1-Docker-Commands/1-10-Docker-PS.robot
```diff
@@ -71,7 +71,7 @@ Docker ps only running containers
     ${output}=  Split To Lines  ${output}
     ${len}=  Get Length  ${output}
     Create several containers
-    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} ps
+    ${rc}  ${output}=  Wait Until Keyword Succeeds  5x  5 seconds  Run And Return Rc And Output  docker %{VCH-PARAMS} ps
```
mhagen-vmware: why would we need to run docker ps multiple times until it succeeds?

jzt: https://github.com/vmware/vic/issues/3434#issuecomment-267375796

tests/test-cases/Group1-Docker-Commands/1-33-Docker-Service.robot
jzt: Yeah: `docker service scale <id of service>=<number of tasks>`. 

https://docs.docker.com/engine/swarm/swarm-tutorial/scale-service/  for more info

cgtexmex:
LGTM

tests/test-cases/Group1-Docker-Commands/1-10-Docker-PS.robot
mhagen-vmware: but you aren't looking for anything? docker ps should always work, just the power state of some containers might be inconsistent for a while right?

PR #4831, opened 2017-04-20
vcsim: Use soapenv namespace as required by other clients

dougm:
With this set of changes, the pyvmomi/sample programs and rbvmomi+rvc
basics work against vcsim.

### DIFF
--- a/pkg/vsphere/simulator/property_collector.go
+++ b/pkg/vsphere/simulator/property_collector.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


dougm:
@lamw @jakerobinson tested these changes with pyvmomi and rbvmomi/rvc, haven't tried power cli yet.

pkg/vsphere/simulator/property_collector.go
```diff
@@ -112,7 +112,11 @@ func fieldValue(rval reflect.Value, p string) (interface{}, error) {
 	fields := strings.Split(p, ".")
```
sgairo: ^^copyright needs a range :)

pkg/vsphere/simulator/property_collector_test.go
```diff
@@ -339,5 +339,88 @@ func TestExtractEmbeddedField(t *testing.T) {
 	if ok {
```
sgairo: ^^ ditto

jakerobinson:
@jrgarcia you should be trying this out :) Here's a recap and quickstart (thanks William) http://www.virtuallyghetto.com/2017/04/govcsim-neat-incubation-project-vcenter-server-esxi-api-based-simulator.html

jrgarcia:
@jakerobinson You better believe I've already been looking at this. :smile:

PR #4834, opened 2017-04-20
[skip ci] Add a new bash helper for upgrade

caglar10ur:


lib/install/validate/update.go
```diff
@@ -85,9 +85,11 @@ func (v *Validator) AssertVersion(conf *config.VirtualContainerHostConfigSpec) (
...
 	}
-	if !older {
-		v.NoteIssue(errors.Errorf("%q has same or newer version %s than installer version %s. No upgrade is available.", conf.Name, conf.Version.ShortVersion(), installerBuild.ShortVersion()))
-		return err
+	if !force {
+		if !older {
+			v.NoteIssue(errors.Errorf("%q has same or newer version %s than installer version %s. No upgrade is available.", conf.Name, conf.Version.ShortVersion(), installerBuild.ShortVersion()))
+			return err
+		}
 	}
```
emlin: It would be nice to have a warning message if we continue upgrade but with older binary, that means data migration will never happen

Issue #4774, opened 2017-04-13
Correctness: vet code paths for correct idempotent behaviour


hickeng:
The entire basis of the project is a declarative idempotent design, from deployment to operation.

We need to vet and add tests (likely requires #3640 or #3639) to ensure that this approach is followed.

mhagen-vmware:
@hickeng   We need this defined so we can pull it into our release plan.  Can you work on that?

Issue #696, opened 2016-05-17
Milestone: basic vmfork


hickeng:
Support starting a container via vmfork. Some of the tasks below are aspirational and noted here simply because they will inform the approach to early starts
#### Target
- [ ] Determine which hosts in cluster support vmfork
- [ ] Ensure container fork parent on each target host (if existing ISO match not available)
  - [ ] Create fork parent if existing ISO match not available
  - [ ] Register parent per host with ISO checksum
  - [ ] Note interest in parent (add VCH MOID to parent extraConfig)
- [ ] Handle add/remove of hosts from cluster
  - [ ] Add fork parent to incoming hosts
  - [ ] Remove fork parent from outgoing hosts

mdubya66:
needs review, given the state of vmfork

PR #4835, opened 2017-04-20
Copy existing volume data to mounts [full ci]

### LINKED ISSUE #3482: Volumes clobber underlying files
When volumes are mounted in the container they clobber the underlying files instead of copying them like docker does:
https://docs.docker.com/engine/tutorials/dockervolumes/

> Volumes are initialized when a container is created. If the container’s base image contains data at the specified mount point, that existing data is copied into the new volume upon volume initialization. (Note that this does not apply when mounting a host directory.)

Steps to reproduce:
Build & run container from the following dockerfile, `/etc/example/nowyouseeme` should exist.
Works correctly on Docker 1.10.3, does not work on VIC.
```
FROM debian:jessie

RUN mkdir -p /etc/example/nowyouseeme

VOLUME ["/etc/example"]

ENTRYPOINT ["/bin/bash"]
```

jakedsouza:
**Issue:**
  A containers base image data is **not** copied when a volume is mounted, if the containers base image contains data at the mount point. 

**Fix:**
  After mounting the volumes, copy the underlying base image files to the mount point. 
 
Changes - 
1. After all volumes are mounted, call `populateVolumes` 
2. For each mountpoint, `bind mount` the underlying *parent directory* to a tmp directory `/.bind` 
3. Copy the base image data to the mounted volume 
4. Cleanup the temp directory 
5. Add a markfile `.mountscopied` in the root so that we don't perform the operation again 
6. Introduce `CopyMode` field in MountSpec. Which defines whether to copy the data or not. 
   Currently it always copies once.  

Fixes #3482

### DIFF
--- a/cmd/vic-init/tether_test.go
+++ b/cmd/vic-init/tether_test.go
@@ -165,6 +165,13 @@ func (t *Mocker) MountTarget(ctx context.Context, source url.URL, target string,
 	return nil
 }
 
+// CopyExistingContent copies the underlying files shadowed by a mount on a directory
+// to the volume mounted on the directory
+func (t *Mocker) CopyExistingContent(source string) error {
+	defer trace.End(trace.Begin(fmt.Sprintf("mocking copyExistingContent from %s", source)))
+	return nil
+}
+
 // Fork triggers vmfork and handles the necessary pre/post OS level operations
 func (t *Mocker) Fork() error {
 	defer trace.End(trace.Begin("mocking fork"))
--- a/lib/config/executor/container_vm.go
+++ b/lib/config/executor/container_vm.go
@@ -29,6 +29,17 @@ const (
 	KILLED
 )
 
+// CopyMode type to define whether to copy data from the base image on mount
+type CopyMode int
+
+const (
+	// CopyNever Dont copy data on mount
+	CopyNever CopyMode = iota + 1
+
+	// CopyNew Copy data to the volume when it is first mounted
+	CopyNew
+)
+
 // Common data between managed entities, across execution environments
 type Common struct {
 	// A reference to the components hosting execution environment, if any
@@ -78,17 +89,6 @@ type ExitLog struct {
 	ExitStatus int
 	Message    string
 }
-
-// CopyMode type to define whether to copy data from the base image on mount
-type CopyMode int
-
-const (
-	// CopyNever Dont copy data on mount
-	CopyNever CopyMode = iota
-
-	// CopyNew Copy data to the volume when it is first mounted
-	CopyNew
-)
 
 // MountSpec details a mount that must be executed within the executor
 // A mount is a URI -> path mapping with a credential of some kind
--- a/lib/config/executor/container_vm.go
+++ b/lib/config/executor/container_vm.go
@@ -79,6 +79,17 @@ type ExitLog struct {
 	Message    string
 }
 
+// CopyMode type to define whether to copy data from the base image on mount
+type CopyMode int
+
+const (
+	// CopyNever Dont copy data on mount
+	CopyNever CopyMode = iota
+
+	// CopyNew Copy data to the volume when it is first mounted
+	CopyNew
+)
+
 // MountSpec details a mount that must be executed within the executor
 // A mount is a URI -> path mapping with a credential of some kind
 // In the case of a labeled disk:
@@ -94,6 +105,9 @@ type MountSpec struct {
 	// Freeform mode string, which could translate directly to mount options
 	// We may want to turn this into a more structured form eventually
 	Mode string `vic:"0.1" scope:"read-only" key:"mode"`
+
+	// CopyMode specifies if data should be copied from the base image on first mount
+	CopyMode CopyMode `vic:"0.1" scope:"read-only" key:"copymode"`
 }
 
 // ContainerVM holds that data tightly associated with a containerVM, but that should not
--- a/lib/portlayer/storage/nfs/volume.go
+++ b/lib/portlayer/storage/nfs/volume.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path"
 
+	"github.com/vmware/vic/lib/config/executor"
 	"github.com/vmware/vic/lib/portlayer/storage"
 	"github.com/vmware/vic/lib/portlayer/util"
 	"github.com/vmware/vic/pkg/trace"
@@ -117,7 +118,7 @@ func (v *VolumeStore) VolumeCreate(op trace.Operation, ID string, store *url.URL
 		return nil, fmt.Errorf("Unexpected scheme (%s) for volume store (%s)", u.Scheme, v.Name)
 	}
 
-	vol, err := storage.NewVolume(v.SelfLink, ID, info, NewVolume(u, v.volDirPath(ID)))
+	vol, err := storage.NewVolume(v.SelfLink, ID, info, NewVolume(u, v.volDirPath(ID)), executor.CopyNew)
 	if err != nil {
 		return nil, err
 	}
@@ -130,7 +131,7 @@ func (v *VolumeStore) VolumeCreate(op trace.Operation, ID string, store *url.URL
 	return vol, nil
 }
 
-// Removes a volume and all of its contents from the nfs store. We already know via the cache if it is in use.
+// VolumeDestroy Removes a volume and all of its contents from the nfs store. We already know via the cache if it is in use.
 func (v *VolumeStore) VolumeDestroy(op trace.Operation, vol *storage.Volume) error {
 	target, err := v.Service.Mount(op)
 	if err != nil {
@@ -184,7 +185,7 @@ func (v *VolumeStore) VolumesList(op trace.Operation) ([]*storage.Volume, error)
 		}
 
 		u, _ := v.Service.URL()
-		vol, err := storage.NewVolume(v.SelfLink, fileInfo.Name(), volMetadata, NewVolume(u, v.volDirPath(fileInfo.Name())))
+		vol, err := storage.NewVolume(v.SelfLink, fileInfo.Name(), volMetadata, NewVolume(u, v.volDirPath(fileInfo.Name())), executor.CopyNew)
 		if err != nil {
 			op.Errorf("Failed to create volume struct from volume directory (%s)", fileInfo.Name())
 			return nil, err
--- a/lib/portlayer/storage/volume.go
+++ b/lib/portlayer/storage/volume.go
@@ -22,6 +22,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/vmware/vic/lib/config/executor"
 	"github.com/vmware/vic/lib/portlayer/util"
 	"github.com/vmware/vic/pkg/trace"
 )
@@ -70,10 +71,12 @@ type Volume struct {
 
 	// Backing device
 	Device Disk
+
+	CopyMode executor.CopyMode
 }
 
 // NewVolume creates a Volume
-func NewVolume(store *url.URL, ID string, info map[string][]byte, device Disk) (*Volume, error) {
+func NewVolume(store *url.URL, ID string, info map[string][]byte, device Disk, copyMode executor.CopyMode) (*Volume, error) {
 	storeName, err := util.VolumeStoreName(store)
 	if err != nil {
 		return nil, err
@@ -93,8 +96,8 @@ func NewVolume(store *url.URL, ID string, info map[string][]byte, device Disk) (
 		SelfLink: selflink,
 		Device:   device,
 		Info:     info,
+		CopyMode: copyMode,
 	}
-
 	return vol, nil
 }
 
--- a/lib/tether/interfaces.go
+++ b/lib/tether/interfaces.go
@@ -37,6 +37,7 @@ type Operations interface {
 	Apply(endpoint *NetworkEndpoint) error
 	MountLabel(ctx context.Context, label, target string) error
 	MountTarget(ctx context.Context, source url.URL, target string, mountOptions string) error
+	CopyExistingContent(source string) error
 	Fork() error
 	// Returns two DynamicMultiWriters for stdout and stderr
 	SessionLog(session *SessionConfig) (dio.DynamicMultiWriter, dio.DynamicMultiWriter, error)
--- a/lib/tether/ops_linux.go
+++ b/lib/tether/ops_linux.go
@@ -31,7 +31,6 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/d2g/dhcp4"
-	"github.com/docker/docker/pkg/archive"
 	// need to use libcontainer for user validation, for os/user package cannot find user here if container image is busybox
 	"github.com/opencontainers/runc/libcontainer/user"
 	"github.com/vishvananda/netlink"
@@ -41,11 +40,6 @@ import (
 	"github.com/vmware/vic/pkg/ip"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vmw-guestinfo/rpcout"
-)
-
-const (
-	bindDir      = "/.bind"
-	mountsCopied = "/.mountscopied"
 )
 
 var (
@@ -721,11 +715,6 @@ func (t *BaseOperations) MountLabel(ctx context.Context, label, target string) e
 		return errors.New(detail)
 	}
 
-	if err := copyExistingContent(target); err != nil {
-		detail := fmt.Sprintf("copying existing data to %s failed: %s", target, err)
-		return errors.New(detail)
-	}
-
 	return nil
 }
 
@@ -743,11 +732,6 @@ func (t *BaseOperations) MountTarget(ctx context.Context, source url.URL, target
 	if err := Sys.Syscall.Mount(rawSource, target, nfsFileSystemType, 0, mountOptions); err != nil {
 		log.Errorf("mounting %s on %s failed: %s", source.String(), target, err)
 		return err
-	}
-
-	if err := copyExistingContent(target); err != nil {
-		detail := fmt.Sprintf("copying existing data to %s failed: %s", target, err)
-		return errors.New(detail)
 	}
 
 	return nil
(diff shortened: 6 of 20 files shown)


lib/tether/tether.go
```diff
@@ -42,11 +44,16 @@ import (
...
+	// MaxDeathRecords The maximum number of records to keep for restarting processes
 	MaxDeathRecords = 5
 
 	// the length of a truncated ID for use as hostname
 	shortLen = 12
+
+	// temp directory to copy existing data to mounts
+	bindDir = "/.bind"
+	// markfile to avoid re-copying existing data to mounts
+	mountsCopied = "/.mountscopied"
```
fdawg4l: I'm not sure this is worth implementing since we're using the same path for shared volumes.  In which case container1 and container2 and both mount the same volume (`someVol`), and both will race on copying their contents to it.

Rather, lets remove this check and always copy the dirents in the underlying mount point (the bind mount) as long as they are newer or do not exist.

jakedsouza: I think we should not copy multiple times. 

As per docker's implementation, this operation is performed only once when the container is created. 

Also it looks like docker copies *ALL* the data and not the new and non-existent ones.   

https://github.com/vmware/vic/blob/master/vendor/github.com/docker/docker/container/container_unix.go#L379-L397

lib/tether/ops_linux.go
```diff
@@ -699,6 +700,56 @@ func (t *BaseOperations) MountTarget(ctx context.Context, source url.URL, target
...
+
+	source = filepath.Clean(source)
+
+	log.Debugf("creating directory %s", bindDir)
+	if err := os.MkdirAll(bindDir, 0644); err != nil {
+		log.Errorf("error creating directory %s: %+v", bindDir, err)
+		return err
+	}
+
+	parentDir := filepath.Dir(filepath.Clean(source))
```
matthewavery: `source` has already had `filepath.Clean()` called on it at this point. I believe that `filepath.Dir()` calls a clean on the supplied target. So maybe just eliminate the nested function call here.

lib/tether/tether.go
```diff
@@ -252,6 +257,41 @@ func (t *tether) setMounts() error {
 			return fmt.Errorf("unsupported volume mount type for %s: %s", k, v.Source.Scheme)
 		}
 	}
+	return t.populateVolumes()
```
matthewavery: So this is getting called after we already mount the nfs/device. I am not sure of the behavior of a bind mount after you have already mounted the target. I believe that you will actually end up with the mounted contents of the mounted device. I have not tried it yet though, so I could be wrong.

lib/tether/tether.go
```diff
@@ -42,11 +44,16 @@ import (
...
 const (
-	// The maximum number of records to keep for restarting processes
+	// MaxDeathRecords The maximum number of records to keep for restarting processes
 	MaxDeathRecords = 5
 
 	// the length of a truncated ID for use as hostname
 	shortLen = 12
+
+	// temp directory to copy existing data to mounts
+	bindDir = "/.bind"
```
jakedsouza: We don't have a label for nfs mounts. We could use a combination of hostname and path replacing `/` with `-`.

lib/tether/tether.go
jakedsouza: This works fine because we bind mount the parent directory and not the actual target. Bind mounting the parent directory results in mounting the original dir and not the mounted device 

https://superuser.com/a/389262/486095

lib/tether/ops_linux.go
jakedsouza: My bad, had it twice. 
We still need to call clean once even though filepath.Dir calls clean. We are trying to get the parent directory here. If the source has a trailing slash, filepath.Dir returns the same source instead of the parent 

https://play.golang.org/p/8evsPBsDH9

matthewavery: Sure thing 😄 yeah I was mainly just looking at the filepath.clean in the function call.

lib/tether/tether.go
jakedsouza: Same question as above. 
1. What do we do for NFS volumes with a URL? 
2. Do we create a different file per volume?

lib/tether/ops_linux.go
jakedsouza: Fixed

lib/tether/tether.go
matthewavery: Ah yes, I had tried this when I left this comment, but indeed from our conversation above if the `/` is left on the end you actually end up with the mount being bound to the target location. Sans the `/` works great! Sorry about that.

matthewavery: Ah I see, hence the Dir call we talked about above. makes sense, sorry about that.

cmd/vic-init/tether_test.go
```diff
@@ -165,6 +165,13 @@ func (t *Mocker) MountTarget(ctx context.Context, source url.URL, target string,
 	return nil
 }
 
+// CopyExistingContent copies the underlying files shadowed by a mount on a directory
+// to the volume mounted on the directory
+func (t *Mocker) CopyExistingContent(source string) error {
```
hickeng: @caglar10ur Please think about how we can change this test setup when continuing with the tether restructure. It was painful at the time to have to duplicate code but with time to do it correctly there should be a better option.

lib/tether/tether.go
```diff
@@ -252,6 +258,46 @@ func (t *tether) setMounts() error {
...
+}
+
+func (t *tether) populateVolumes() error {
+	defer trace.End(trace.Begin(fmt.Sprintf("populateVolumes")))
+	// skip if no mounts present
+	if len(t.config.Mounts) == 0 {
+		return nil
+	}
+	// skip if this was done before
+	if _, err := os.Stat(mountsCopied); err == nil {
```
hickeng: This is fine for now, but somewhat violates our goal of idempotent behaviour - if Reload is called and a new volume has been configured (or the CopyMode has changed) then we will not populate newly added volumes.

If we add more complex Copy behaviours (rsync for example) then this would be more simply phrased as a checking function per Copy type that takes a source and dest and returns whether or not to perform the copy, or just copies if needed.

Please open a follow up issue as a child of #4774 to revisit this in a tech debt, with a reference in #696 as that's where we will likely need the revised behaviour.

tests/test-cases/Group1-Docker-Commands/1-06-Docker-Run.robot
```diff
@@ -133,3 +133,13 @@ Docker run verify name and id are not conflated
...
+
+Docker run verify anonymous volume contains files
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run jakedsouza/group-1-06-docker-verify-volume-files:1.0 ls /etc/example
+	Should Be Equal As Integers  ${rc}  0
+	Should Contain  ${output}  thisshouldexist
+	Should Contain  ${output}  testfile.txt
+
+	${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run jakedsouza/group-1-06-docker-verify-volume-files:1.0 cat /etc/example/testfile.txt
+    Should Be Equal As Integers  ${rc}  0
+	Should Contain  ${output}  TestFile
```
hickeng: IIRC you said that if a volume is explicitly specified with `-v name:/path` then this copy behaviour doesn't occur.
Please add a test for this case as well.

lib/config/executor/container_vm.go
```diff
@@ -79,6 +79,17 @@ type ExitLog struct {
 	Message    string
 }
 
+// CopyMode type to define whether to copy data from the base image on mount
+type CopyMode int
```
hickeng: Please move this to the top of the file with the `State` enum.

lib/portlayer/storage/volume.go
```diff
@@ -70,10 +71,12 @@ type Volume struct {
 
 	// Backing device
 	Device Disk
+
+	CopyMode executor.CopyMode
 }
 
 // NewVolume creates a Volume
-func NewVolume(store *url.URL, ID string, info map[string][]byte, device Disk) (*Volume, error) {
+func NewVolume(store *url.URL, ID string, info map[string][]byte, device Disk, copyMode executor.CopyMode) (*Volume, error) {
```
hickeng: This probably doesn't need a backwards compatibility check given CopyNew is the default (by virtual of `0` being the nil value of `int`) but worth confirming that the backwards compatibility tests have a volume configuration that will exercise the copy paths in the future.

lib/tether/ops_linux.go
```diff
@@ -699,6 +700,52 @@ func (t *BaseOperations) MountTarget(ctx context.Context, source url.URL, target
...
+// to the volume mounted on the directory
+// see bug https://github.com/vmware/vic/issues/3482
+func (t *BaseOperations) CopyExistingContent(source string) error {
+	defer trace.End(trace.Begin(fmt.Sprintf("copyExistingContent from %s", source)))
+
+	source = filepath.Clean(source)
+
+	log.Debugf("creating directory %s", bindDir)
+	if err := os.MkdirAll(bindDir, 0644); err != nil {
+		log.Errorf("error creating directory %s: %+v", bindDir, err)
```
fdawg4l: Might want to check for `EEXISTS` and continue in that case.

lib/tether/ops_linux.go
```diff
@@ -699,6 +700,52 @@ func (t *BaseOperations) MountTarget(ctx context.Context, source url.URL, target
...
 }
 
+// CopyExistingContent copies the underlying files shadowed by a mount on a directory
+// to the volume mounted on the directory
+// see bug https://github.com/vmware/vic/issues/3482
+func (t *BaseOperations) CopyExistingContent(source string) error {
+	defer trace.End(trace.Begin(fmt.Sprintf("copyExistingContent from %s", source)))
+
+	source = filepath.Clean(source)
+
```
fdawg4l: Before creating directories and `bind` mounting the underlying mount point, it may be worth checking if there are even files there.  There's no sense descending down the mount/copy path if there's nothing to copy, right?

lib/tether/ops_linux.go
```diff
@@ -699,6 +700,52 @@ func (t *BaseOperations) MountTarget(ctx context.Context, source url.URL, target
...
+		return err
+	}
+
+	log.Debugf("unmounting %s", bindDir)
+	if err := Sys.Syscall.Unmount(bindDir, syscall.MNT_DETACH); err != nil {
+		log.Errorf("error unmounting %+v", err)
+		return err
+	}
+	log.Debugf("removing %s", bindDir)
+	if err := os.Remove(bindDir); err != nil {
```
fdawg4l: I would do this in a `defer` after 712 (in the err == nil case).

lib/tether/ops_linux.go
```diff
@@ -699,6 +700,52 @@ func (t *BaseOperations) MountTarget(ctx context.Context, source url.URL, target
...
+	// copy data from the bindDir to the source
+	// e.g if source is /foo/bar, copy ./bindDir/bar to /foo/bar
+	log.Debugf("copying contents from to %s to %s", mountedSource, source)
+	if err := archive.CopyWithTar(mountedSource, source); err != nil {
+		log.Errorf("err copying %s to %s: %+v", mountedSource, source, err)
+		return err
+	}
+
+	log.Debugf("unmounting %s", bindDir)
+	if err := Sys.Syscall.Unmount(bindDir, syscall.MNT_DETACH); err != nil {
```
fdawg4l: I would defer this after `Mount` is successful.

lib/tether/tether.go
fdawg4l: Echo'ing @hickeng here.  For NFS volumes where 2 containers mount the same volume, the 2nd container will attempt to copy the contents of the mount point *again*.  I thought we discussed this and were simply going to skip copying anything into the volume if the volume was non-empty.

lib/tether/ops_linux.go
```diff
@@ -699,6 +700,52 @@ func (t *BaseOperations) MountTarget(ctx context.Context, source url.URL, target
...
+	if err := Sys.Syscall.Mount(parentDir, bindDir, ext4FileSystemType, syscall.MS_BIND, ""); err != nil {
+		log.Errorf("error mounting to %s: %+v", bindDir, err)
+		return err
+	}
+
+	mountedSource := filepath.Join(bindDir, filepath.Base(source))
+	// copy data from the bindDir to the source
+	// e.g if source is /foo/bar, copy ./bindDir/bar to /foo/bar
+	log.Debugf("copying contents from to %s to %s", mountedSource, source)
+	if err := archive.CopyWithTar(mountedSource, source); err != nil {
```
fdawg4l: Isn't this an explicit copy?  It will copy over the dest even if the dest exists, no?  We really don't want that.  Infact, we shouldn't be copying at all if anything exists in the volume (except for the `lost+found` directory).

lib/config/executor/container_vm.go
```diff
@@ -79,6 +79,17 @@ type ExitLog struct {
...
+
+const (
+	// CopyNever Dont copy data on mount
+	CopyNever CopyMode = iota
```
caglar10ur: make this iota + 1 so that we can differentiate zero value

lib/portlayer/storage/nfs/volume.go
```diff
@@ -117,7 +118,7 @@ func (v *VolumeStore) VolumeCreate(op trace.Operation, ID string, store *url.URL
 		return nil, fmt.Errorf("Unexpected scheme (%s) for volume store (%s)", u.Scheme, v.Name)
 	}
 
-	vol, err := storage.NewVolume(v.SelfLink, ID, info, NewVolume(u, v.volDirPath(ID)))
+	vol, err := storage.NewVolume(v.SelfLink, ID, info, NewVolume(u, v.volDirPath(ID)), executor.CopyNew)
```
caglar10ur: Just ranting, no change necessary. I'm really not a big fan of keep adding parameters to the functions. It would be really great to introduce option types for these NewX calls and pass them (like https://github.com/vmware/vic/blob/5126aeb0d93a99e8d21d5ba04072a8d75923c1ad/lib/dns/dns.go#L53)

lib/tether/ops_linux.go
```diff
@@ -31,6 +31,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/d2g/dhcp4"
+	"github.com/docker/docker/pkg/archive"
```
caglar10ur: pfffff, here comes a docker dep into tether :/

fdawg4l: We didn't like this either.  But, this is a fairly concise package.  *coughs*

lib/tether/tether.go
jakedsouza: Done. Removed check for mountsCopied 

Added functionality, if the mounted volume is non empty (except for lost+found) skip the copy flow.

tests/test-cases/Group1-Docker-Commands/1-06-Docker-Run.robot
jakedsouza: I was wrong about the copy behaviour. As per the docker docs 

*Volumes are initialized when a container is created. If the container’s base image contains data at the specified mount point, that existing data is copied into the new volume upon volume initialization. **(Note that this does not apply when mounting a host directory.)***

So the only case when data is *not* copied is when a host directory is mounted. This is something vic does not support.   

I added a test case to verify that the data is copied to a named volume. This is the same as  dockers current behaviour

lib/config/executor/container_vm.go
jakedsouza: Done

lib/portlayer/storage/volume.go
jakedsouza: `CopyNever` is the default, `CopyNew` has a value 1 

Changed the enum to be iota+1 by default as requested by @caglar10ur below

lib/tether/ops_linux.go
jakedsouza: We would need to bind anyways mount to check if there are any files in the underlying directory 

To avoid doing a bind mount, we could change the code flow to the steps below 
1. In mountLabel and mountTarget, check if the original folder is empty 
2. Mount the volume 
3. If step 1 is false then go ahead with copy

lib/tether/ops_linux.go
jakedsouza: For MkdirAll `If path is already a directory, MkdirAll does nothing and returns nil.` so we dont need to check for EEXISTS

lib/tether/ops_linux.go
jakedsouza: Will do. Also added a test case for this.

lib/tether/ops_linux.go
jakedsouza: Done

lib/tether/ops_linux.go
jakedsouza: Done

lib/config/executor/container_vm.go
jakedsouza: Done

jakedsouza:
> Question before approving. What will happen if tether receives a reload in the middle of copy operation?

The `populateVolumes` operation occurs during tether.Start during which, the reload chan is already blocked. So calling reload should not affect this operation.  I might be wrong here though

lib/tether/interfaces.go
```diff
@@ -37,6 +37,7 @@ type Operations interface {
 	Apply(endpoint *NetworkEndpoint) error
 	MountLabel(ctx context.Context, label, target string) error
 	MountTarget(ctx context.Context, source url.URL, target string, mountOptions string) error
+	CopyExistingContent(source string) error
```
jakedsouza: > Is there a reason we do this after all of the volumes have been mounted? Another option would have been to inline this behaviour with the MountTarget logic.

I chose to follow this approach to avoid duplicating the code in MountLabel and MountTarget functions. 

If I understood your second question correctly,
Contents are not copied from VolumeA to VolumeB, they are always from the base image to a volume. One, cannot mount two volumes on the same path.

caglar10ur:
Can you try or write a test to make sure?

caglar10ur:
Otherwise LGTM

rajanashok:
lgtm

jakedsouza:
> Can you try or write a test to make sure?

@caglar10ur Tried to repro manually. I see a single error which I opened up as a separate issue. Im able to repro it on master https://github.com/vmware/vic/issues/4881

matthewavery:
@jakedsouza this looks great 😄

PR #4836, opened 2017-04-20
Add func to get guestinfo secret for vic-machine

### LINKED ISSUE #4516: Load guestinfo encryption key from ovfenv
As a vic-machine developer, I need to know vch configuration encryption key out of VCH guestOS, to decrypt/reencrypt VCH configuration during VCH update.

Acceptance:
API to retrieve VCH configuration encryption key

anchal-agrawal:
Fixes #4516

### DIFF
--- a/lib/install/management/config.go
+++ b/lib/install/management/config.go
@@ -1,0 +1,93 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package management
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/datastore"
+	"github.com/vmware/vic/pkg/vsphere/extraconfig"
+)
+
+var (
+	errSecretKeyNotFound = fmt.Errorf("unable to find guestinfo secret")
+	errNilDatastore      = fmt.Errorf("session's datastore is not set")
+)
+
+// extractSecretFromFile reads and extracts the GuestInfoSecretKey value from the input.
+func extractSecretFromFile(rc io.ReadCloser) (string, error) {
+
+	scanner := bufio.NewScanner(rc)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		// The line is of the format: key = "value"
+		if strings.HasPrefix(line, extraconfig.GuestInfoSecretKey) {
+
+			tokens := strings.SplitN(line, "=", 2)
+			if len(tokens) < 2 {
+				return "", fmt.Errorf("parse error: unexpected token count in line")
+			}
+
+			// Ensure that the key fully matches the secret key
+			if strings.Trim(tokens[0], ` `) != extraconfig.GuestInfoSecretKey {
+				continue
+			}
+
+			// Trim double quotes and spaces
+			return strings.Trim(tokens[1], `" `), nil
+		}
+	}
+
+	return "", errSecretKeyNotFound
+}
+
+// GuestInfoSecret downloads the VCH's .vmx file and returns the GuestInfoSecretKey
+// value. This function expects the datastore in the dispatcher's session to be set.
+func (d *Dispatcher) GuestInfoSecret(vchName string) (*extraconfig.SecretKey, error) {
+	defer trace.End(trace.Begin(""))
+
+	if d.session.Datastore == nil {
+		return nil, errNilDatastore
+	}
+
+	helper, err := datastore.NewHelper(d.ctx, d.session, d.session.Datastore, d.vmPathName)
+	if err != nil {
+		return nil, err
+	}
+
+	// Download the VCH's .vmx file
+	path := fmt.Sprintf("%s.vmx", vchName)
+	rc, err := helper.Download(d.ctx, path)
+	if err != nil {
+		return nil, err
+	}
+
+	secret, err := extractSecretFromFile(rc)
+	if err != nil {
+		return nil, err
+	}
+
+	secretKey := &extraconfig.SecretKey{}
+	if err = secretKey.FromString(secret); err != nil {
+		return nil, err
+	}
+
+	return secretKey, nil
+}
--- a/lib/install/management/config_test.go
+++ b/lib/install/management/config_test.go
@@ -1,0 +1,146 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package management
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+	"path"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/pkg/vsphere/extraconfig"
+	"github.com/vmware/vic/pkg/vsphere/session"
+	"github.com/vmware/vic/pkg/vsphere/simulator"
+)
+
+func TestGuestInfoSecret(t *testing.T) {
+	ctx := context.Background()
+
+	for i, m := range []*simulator.Model{simulator.ESX(), simulator.VPX()} {
+
+		defer m.Remove()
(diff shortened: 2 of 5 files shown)


lib/install/management/config.go
```diff
@@ -0,0 +1,65 @@
...
+	"strings"
+
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/datastore"
+)
+
+// GuestInfoSecret downloads the VCH's vmx file and extracts the guestinfo.ovfEnv
+// value from it. This function expects the datastore in the dispatcher's session
+// to be set.
+func (d *Dispatcher) GuestInfoSecret(vchName string) (string, error) {
```
emlin: Usually secret key is defined as a byte array

lib/install/management/config.go
```diff
@@ -0,0 +1,65 @@
...
+	defer trace.End(trace.Begin(""))
+
+	if d.session.Datastore == nil {
+		return "", fmt.Errorf("session's datastore is not set")
+	}
+
+	helper, err := datastore.NewHelper(d.ctx, d.session, d.session.Datastore, vchName)
```
emlin: the vm folder name might not be always consistent with vm name.

emlin: you should be able to use d.vmPathName directly here

lib/install/management/config.go
```diff
@@ -0,0 +1,65 @@
...
+		return "", err
+	}
+
+	// Extract the guestinfo.ovfEnv value
+	scanner := bufio.NewScanner(rc)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		// The line is of the format: key = "value"
+		if strings.HasPrefix(line, "guestinfo.ovfEnv") {
```
emlin: find one library to parse ini format properties file: https://github.com/toml-lang/toml
might have a look, so that you don't have to parse strings

lib/install/management/config_test.go
```diff
@@ -0,0 +1,111 @@
...
+		assert.Equal(t, secret, "")
+		assert.Equal(t, err, fmt.Errorf("unable to find guestinfo.ovfEnv in the vmx file"))
+
+		// Write key-value pairs to the .vmx file
+		dir := simulator.Map.Get(ds.Reference()).(*simulator.Datastore).Info.(*types.LocalDatastoreInfo).Path
+		err = ioutil.WriteFile(path.Join(dir, vmx), []byte("foo.bar = \"baz\"\n"), 0)
+		if err != nil {
+			t.Fatal(err)
+		}
+		err = ioutil.WriteFile(path.Join(dir, vmx), []byte("guestinfo.ovfEnv = \"secret\"\n"), 0)
```
emlin: just curious, is this because vcsimulator does not write extraconfig during vm creation?

anchal-agrawal: Yes, extraconfig is not written by the simulator. The .vmx file is empty.

lib/install/management/config.go
anchal-agrawal: Good call, thanks Emma!

lib/install/management/config.go
```diff
@@ -0,0 +1,66 @@
...
+		line := scanner.Bytes()
+
+		// The line is of the format: key = "value"
+		if bytes.HasPrefix(line, []byte(extraconfig.GuestInfoSecretKey)) {
+			fields := bytes.Fields(line)
+			if len(fields) != 3 {
+				return nil, fmt.Errorf("error parsing vmx file: unexpected token count in line")
+			}
+
+			return bytes.Trim(fields[2], "\""), nil
```
anchal-agrawal: @emlin @dougm I'm trimming the surrounding `"` here, would it be better to do something like `fields[2][1:len(fields[2]-1]` in case the secret value itself has leading/trailing `"`?

lib/install/management/config.go
```diff
@@ -0,0 +1,66 @@
...
+	path := fmt.Sprintf("%s.vmx", vchName)
+	rc, err := helper.Download(d.ctx, path)
+	if err != nil {
+		return nil, err
+	}
+
+	// Extract the GuestInfoSecretKey value
+	scanner := bufio.NewScanner(rc)
+	for scanner.Scan() {
+		line := scanner.Bytes()
```
dougm: Might be easier to use Text() here and work with string type.

lib/install/management/config.go
```diff
@@ -0,0 +1,66 @@
...
+
+		// The line is of the format: key = "value"
+		if bytes.HasPrefix(line, []byte(extraconfig.GuestInfoSecretKey)) {
+			fields := bytes.Fields(line)
```
dougm: Maybe instead:
```go
fields := strings.SplitN(line, "=", 2)
return strings.Trim(fields[1], `" `) // trim spaces and double quotes
```

lib/install/management/config_test.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		// Attempt to extract the secret with an empty .vmx file
+		secret, err := d.GuestInfoSecret(name)
+		assert.Equal(t, secret, []byte(nil))
+		assert.Equal(t, err, fmt.Errorf("unable to find %s in the vmx file", extraconfig.GuestInfoSecretKey))
+
+		// Write key-value pairs to the .vmx file
+		dir := simulator.Map.Get(ds.Reference()).(*simulator.Datastore).Info.(*types.LocalDatastoreInfo).Path
+		err = ioutil.WriteFile(path.Join(dir, vmx), []byte("foo.bar = \"baz\"\n"), 0)
```
dougm: Think you just want 1 call to ioutil.WriteFile, otherwise you just overwrite the existing data.

lib/install/management/config_test.go
```diff
@@ -0,0 +1,116 @@
...
+		d := &Dispatcher{
+			session:    s,
+			ctx:        ctx,
+			vmPathName: name,
+		}
+
+		// Attempt to extract the secret with an empty .vmx file
+		secret, err := d.GuestInfoSecret(name)
+		assert.Equal(t, secret, []byte(nil))
+		assert.Equal(t, err, fmt.Errorf("unable to find %s in the vmx file", extraconfig.GuestInfoSecretKey))
```
dougm: consider adding `var errSecretKeyNotFound = fmt.Errorf(...)` to config.go rather than 2 copies of the message.

lib/install/management/config_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+		// Write key-value pairs to the .vmx file
+		dir := simulator.Map.Get(ds.Reference()).(*simulator.Datastore).Info.(*types.LocalDatastoreInfo).Path
```
dougm: Add a TODO: simulator should write ExtraConfig to the .vmx file for us

lib/install/management/config.go
```diff
@@ -0,0 +1,66 @@
...
+		return nil, err
+	}
+
+	// Download the VCH's vmx file
+	path := fmt.Sprintf("%s.vmx", vchName)
+	rc, err := helper.Download(d.ctx, path)
+	if err != nil {
+		return nil, err
+	}
+
```
dougm: Could split this here into 2 functions, easier to unit test code below for error handling, etc.

anchal-agrawal: Done!

lib/install/management/config.go
```diff
@@ -0,0 +1,89 @@
...
+func extractSecretFromFile(rc io.ReadCloser) (string, error) {
+
+	scanner := bufio.NewScanner(rc)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		// The line is of the format: key = "value"
+		if strings.HasPrefix(line, extraconfig.GuestInfoSecretKey) {
+
+			fields := strings.Fields(line)
```
dougm: Why still using strings.Fields rather than strings.SplitN?  In general, "=" is the delimiter, not whitespace.  While the secret value likely won't have spaces, it is still correct to split on "=", then Trim with the cutset `" ` (trims both double quotes + spaces)

anchal-agrawal: I added the `SplitN` logic and then added a unit test case for a malformed .vmx file, where there's no `=` in-between the key and value:

```
guestinfo.ovfEnv "value"
```
or
```
guestinfo.ovfEnv "valu=e"
```

For these cases, we get a different number of tokens depending on whether the value itself has `=` in it. I want to add logic before returning the second token to check that the tokens are valid (something like `len(tokens) should be 2`) but that doesn't apply to both cases above.

anchal-agrawal: Hmm, this is a unit test issue, not a code logic issue. The code should still ensure that `len(tokens) = 2`, but the unit test can check for the error to be `invalid token count` or `invalid secret key`. I've modified the unit test to do so, and the code to use `SplitN`. Thanks!

lib/install/management/config.go
```diff
@@ -0,0 +1,88 @@
...
+	"io"
+	"strings"
+
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/datastore"
+	"github.com/vmware/vic/pkg/vsphere/extraconfig"
+)
+
+var (
+	errSecretKeyNotFound = fmt.Errorf("unable to find %s in the .vmx file", extraconfig.GuestInfoSecretKey)
```
emlin: might not mention .vmx file here? wondering we don't want to expose how secret information is persistent

lib/install/management/config.go
```diff
@@ -0,0 +1,88 @@
...
+	scanner := bufio.NewScanner(rc)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		// The line is of the format: key = "value"
+		if strings.HasPrefix(line, extraconfig.GuestInfoSecretKey) {
+
+			tokens := strings.SplitN(line, "=", 2)
+			if len(tokens) < 2 {
+				return "", fmt.Errorf("error parsing .vmx file: unexpected token count in line")
```
emlin: same ^^^

lib/install/management/config.go
anchal-agrawal: Fixed!

PR #4839, opened 2017-04-21
Introduce feature subpackage

caglar10ur:
so that we can all use it instead of defining random numbers in
different subsystems.


pkg/version/version.go
```diff
@@ -84,7 +83,7 @@ func GetBuild() *Build {
 		BuildDate:     BuildDate,
 		BuildNumber:   BuildNumber,
 		State:         State,
-		PluginVersion: MaxPluginVersion,
+		PluginVersion: feature.MaxPluginVersion,
```
emlin: (MaxPluginVersion - 1) should be used as current version, because the plugin is used to migrate data from (X-1) to X.
if we use MaxPluginVersion directly here, plugin should work as X => (X+1).

I think current logic is easy to understand, so please keep that, and make sure current version is always same to latest plugin version.

lib/migration/feature/feature.go
```diff
@@ -0,0 +1,27 @@
...
+package feature
+
+const (
+	AddCommonSpecForVCHVersion = iota + 1
+	AddCommonSpecForContainerVersion
+	TasksSupportedVersion
+	RenameSupportedVersion
+	MigrateRegistryVersion
+	ExecSupportedVersion
+
```
emlin: add a comment here as well? to make sure we can only add new items in the end

lib/migration/manager/manager.go
```diff
@@ -75,9 +75,9 @@ func (m *DataMigrator) Register(ver int, target string, plugin Plugin) error {
 			Message: "Empty Plugin object is not allowed",
 		}
 	}
-	if ver > version.MaxPluginVersion {
+	if ver > feature.MaxPluginVersion {
```
emlin: feels like we could have a common method in feature package for this check as well. 😄

emlin:
thanks @caglar10ur for this refactory 👍

lib/migration/manager/manager.go
caglar10ur: I left this like that for now as this is the only caller, we can add a fn if we find ourselves doing this check over and over again :)

PR #4840, opened 2017-04-21
is able to stop container not only in running status 

### LINKED ISSUE #4296: Ensure that runblocking is configured anew for each container start
**Story**
As a docker API user calling start I expect the container process to launch when the container starts
As a docker API user calling attach before start I expect to receive _all_ of the container process output

**Detail**
We have a mechanism called runblocking that inhibits the container process from starting until an attach is performed. This is triggered by the following API call ordering:
1. attach
2. start

If calling attach after start then no blocking occurs. This is implemented via a flag that suppresses launching of the container process until that attach has occurred (or a timeout triggers IIRC - need to confirm).
I'm raising this issue because I don't think there's a test the covers the following case:
1. attach
2. start
3. stop
4. start
-> check that process launched without blocking

Estimated at 5 for the case where behavioural changes are needed - if after investigation the behaviour is correct and only a test is needed this should be a 2.

**Acceptance**
- [ ] container process launches without blocking on the second `docker start <id>` having been launched previously with `docker run -i` (i.e. runblocking enabled)

emlin:
Fixes #4296 
Sometimes container might stuck in unexpected status, if we do not stop container not in running status, user has no way to recover the system from error. So this fix is to open the limitation in docker stop.

There is one case that container may stay in starting status because tether is waiting connection between container and VCH. Here is how to trigger that:
- docker run -it <image> date
- disconnect serial port from VC
- docker start

As @caglar10ur said, this behavior in tether is as expected, because we do not want to lose any output from container.


caglar10ur:
LGTM

PR #4845, opened 2017-04-21
Enable vsyscall emulation [specific ci=4817]

### LINKED ISSUE #4817: Server error from portlayer: ServerFaultCode: Permission to perform this operation was denied.
docker image created with RHEL base works fine on engine version v0.8.0-7315-c8ac999. However fails on engine version: v1.1.0-9852-e974a51.

cVM powers on and powers off immediately afterwards during docker run -itd. using -it throws "Server error from portlayer: ServerFaultCode: Permission to perform this operation was denied." to the console.

could it be something on the startup script that offends the new engine? I change the hostname, modify hosts file and resolver etc. there.

I didn't find this in known issues. Also, I assume this release is enterprise [supported](https://github.com/vmware/vic/releases); will log a call shortly.

hickeng:
Enables vsyscall emulation and adds test to confirm. The test is dependent
on centos:6 image continuing to have a glibc variant that relies on
vsyscall as a primary path, not a fallback.

Fixes #4817

### DIFF
--- a/tests/test-cases/Group0-Bugs/4817.robot
+++ b/tests/test-cases/Group0-Bugs/4817.robot
@@ -1,0 +1,24 @@
+# Copyright 2016-2017 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#	http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License
+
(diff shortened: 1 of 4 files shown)


caglar10ur:
LGTM

tests/test-cases/Group0-Bugs/4817.robot
```diff
@@ -0,0 +1,24 @@
...
+*** Settings ***
+Documentation  Test 4817
+Resource  ../../resources/Util.robot
+Suite Setup  Install VIC Appliance To Test Server  certs=${false}
+Suite Teardown  Cleanup VIC Appliance On Test Server
+
+*** Test Cases ***
+Check kernel vsyscall
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -t centos:6 bash -c date
+    Should Be Equal As Integers  ${rc}  0
```
mhagen-vmware: Can we check the date too to at least see that it mostly works? Maybe just the day/month/year, don't really care about time accuracy, but would like to see that the command returns something reasonable while we are here.

PR #4846, opened 2017-04-21
[full ci] Migrator fix: migrateConfig should use current version

caglar10ur:


lib/migration/migrator.go
```diff
@@ -97,7 +97,7 @@ func migrateConfig(ctx context.Context, s *session.Session, data map[string]stri
 	}
 
 	_, err = manager.Migrator.Migrate(ctx, s, target, currentID, dst)
-	dst[verKey] = strconv.Itoa(feature.MaxPluginVersion)
+	dst[verKey] = strconv.Itoa(feature.MaxPluginVersion - 1)
```
emlin: wondering why this has spaces around -, but all others are not. 😄

Issue #4478, opened 2017-03-30
Specifying volume-store == image-store leads to unintended volume loss (ESX)


corrieb:
Related to #4233, it turns out that VMFS on ESX has different behavior to the iSCSi datastore I was using in 4233. 

On ESX, if you specify ``--name=dev --image-store=datastore1 --volume-store=datastore1/dev:default`` in vic-machine create, then all your volumes will go into the *same* folder as your images and the VCH. On iSCSI, it creates a dev_1 folder (see #4233). 

There may be other datastore types that exhibit the ESX behavior.

If I then create a volume called foo and run vic-machine delete without --force, I get the message:

``Mar 30 2017 20:37:30.345Z WARN  Since --force was not specified, the following volume stores will not be removed. Use the vSphere UI to delete content you do not wish to keep.
	default: [datastore1] dev``

However, when I look at the datastore, my foo volume was deleted.

I would argue that any case where we delete persist data when we're not supposed to is serious. 

There are two courses of action here: 

(1) Prevent volume stores being specified as the same folder name as the VCH name. Although that is quite a natural thing to want to do if you want your volumes to have the same lifecycle as the VCH.

(2) Allow the current behavior, but don't delete the volumes unless --force is specified.

mhagen-vmware:
please estimate and move to backlog when ready

stuclem:
@mhagen-vmware does this need to go in the release notes for 1.1?

mhagen-vmware:
@stuclem yes, this should be release noted

stuclem:
Proposed release note:

------

- <b>Specifying the same datastore for volume store and images store leads to unintended volume loss on ESXi hosts. [#4557](https://github.com/vmware/vic/issues/4557)</b><br>When deploying VCHs directly to ESXi hosts, if you specify `vic-machine create --name dev --image-store datastore1 --volume-store datastore1/dev:default`, volumes will go into the same folder as images and the VCH. If you then run `vic-machine delete`, the volumes are deleted, even if you do no specify `--force`. This does not occur when deploying to vCenter Server.

------

@corrieb @mhagen-vmware is this OK?

mhagen-vmware:
lgtm

stuclem:
Thanks @mhagen-vmware

hickeng:
[...]

PR #4848, opened 2017-04-24
fix unexpected appending "_1" on applianceVM directory on vc

maplain:
one simple step towards the fix of https://github.com/vmware/vic/issues/4478

### DIFF
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -490,6 +490,10 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 		log.Errorf("Create appliance reported: %s", info.Error.LocalizedMessage)
 	}
 
+	if err = d.createVolumeStores(conf); err != nil {
+		return errors.Errorf("Exiting because we could not create volume stores due to error: %s", err)
+	}
+
 	// get VM reference and save it
 	moref := info.Result.(types.ManagedObjectReference)
 	conf.SetMoref(&moref)
--- a/lib/install/management/create.go
+++ b/lib/install/management/create.go
@@ -46,10 +46,6 @@ func (d *Dispatcher) CreateVCH(conf *config.VirtualContainerHostConfigSpec, sett
 		return err
 	}
 
-	if err = d.createVolumeStores(conf); err != nil {
-		return errors.Errorf("Exiting because we could not create volume stores due to error: %s", err)
-	}
-
 	if err = d.createAppliance(conf, settings); err != nil {
 		return errors.Errorf("Creating the appliance failed with %s. Exiting...", err)
 	}


lib/install/management/appliance.go
```diff
@@ -146,7 +149,7 @@ func (d *Dispatcher) getName(vm *vm.VirtualMachine) string {
 	return name
 }
 
-func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
+func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, conf *config.VirtualContainerHostConfigSpec, force bool) error {
```
emlin: this method is called for both appliance vm delete and container vm delete, I think we should not start from here to detect if the folder is same to volume store

lib/install/management/store_files.go
```diff
@@ -201,20 +214,24 @@ func (d *Dispatcher) getChildren(ds *object.Datastore, dsPath string) ([]string,
...
+func (d *Dispatcher) lsSubFolder(ds *object.Datastore, dsPath string, matchPattern ...string) (*types.ArrayOfHostDatastoreBrowserSearchResults, error) {
 	defer trace.End(trace.Begin(dsPath))
 
+	// match every subfolers by default
+	if len(matchPattern) == 0 {
+		matchPattern = append(matchPattern, "*")
+	}
 	spec := types.HostDatastoreBrowserSearchSpec{
-		MatchPattern: []string{"*"},
+		MatchPattern: matchPattern,
```
emlin: What's the result, if you list folder test1/, and exclude test1/abc/volumes?
will test1/abc returns?
I'm asking this question, because if that is returned, you're still delete test1/abc folder no matter how.

maplain:
@anchal-agrawal sure. thanks for pointing it out. I'll pay more attention in my future PRs.

anchal-agrawal:
Thanks @maplain - we try to adhere to the commit guidelines at https://github.com/vmware/vic/blob/master/CONTRIBUTING.md#format-of-the-commit-message. The subject of the commit message should be short (about 50 characters) and more details (if necessary) should be included in the message body, but what we have now is much better.

lib/install/management/store_files.go
maplain: actually I'm not sure. So I just tried to create a vch 'dev' with volumeStore being 'dev/abc', thus 'dev/abc/volumes' is excluded. after I delete it without --force, dev/abc still exists, VIC, kvStore, appliance is gone. what's weird is appliance.iso, bootstrap.iso and tether.debug are also not removed.

maplain: though according to the API, I feel like you're right, dev/abc would be returned and removed. I need to figure out why .iso and tether.debug are not removed. Is this the default behavior when --force is not specified?

emlin: No, I don't think so. Even no --force, delete will try to delete folder anyway, only volume store is exception.
The problem is not only this one, you'll end up to delete volume store if parent dir is returned in non vsan datastore in 6.0

maplain: @emlin 
1. deleteIteratively -> getSortedChildren -> getChildren which only collects files. so tests1/abc is never returned;
2. But you're right, if there were other nested directories, they will not be cleaned up;
3. Because in the original implementation, their removal is achieved indirectly by removing the root directory

lib/install/management/appliance.go
maplain: @emlin what's your advice on where to put this logic?

lib/install/management/store_files.go
emlin: the requirement for original recursively delete is different to what you're adding now, that's why I said this does not work as expected.

lib/install/management/appliance.go
emlin: Still depends on how would you implement this logic, but ideally it does not impact container vm deletion. You might add a new method for appliance vm deletion.

maplain:
@emlin can you elaborate on why does it matter? As long as they're created and destroyed separately based on --force without conflicts, I feel like it doesn't matter which directory applianceVM or volumeStore are put to.

maplain:
in the latest PR, I wrote a deleteFilesRecursively() to address the issue you mentioned earlier today. I build and test it, it works as expected. and you can use --debug 1 to check the log.

emlin:
@maplain please check the discussion in this PR. I think we need to reach agreement before actually deliver a big change. How do you think?

maplain:
@emlin I don't really understand your question... in that case, new vch is created in a new directory, volume store reuses the old one. I think as long as delete(with/without) works as expected, new vch being created in dev_1 instead of dev should not be a problem.. 

like I suggested before, before this patch is merged in, this is exactly our code's behavior on vc, applianceVM is always put in dev_1, because when createAppliance is called, dev is already created by createVolumeStore, and vApp.createChildVM will append _1 to dev.

but delete with/without works correctly.

emlin:
@maplain I know that before your fix, the code works in this way, but I think we want to make it better.
The thing is that if user specify the same folder with VCH name as the volume store, what does that mean.
If user assumes that makes sure the volume goes into VCH folder, then our implementation has problem anyhow.
If not, this ability just make our file management messy. I don't think it's a good implementation.

maplain:
@emlin yes, if user just simply wants volumestore to be in the same directory, it's not enforced. but if user wants to make them have the same lifecycle, then whether they're put together, they live or die together already. 
I agree with you on this, `The thing is that if user specify the same folder with VCH name as the volume store, what does that mean.` But let's assume different users may have different intents when they do the same thing, my point is: from a developer's perspective, as long as the management logic in code is guaranteed to have the correct index or handle on that directory whatever its name is, everything should be working fine. after all, it's just a variable to code and it has no idea what's the value of it.
in terms of file management, i think regardless of how this issue is finally handled, it stands out as a problem on its own already. creations and destructions of directories are spread over the code base, which is confusing and error-prone, IMHO. 
for example, if you try to create a vch on vc(name: dev, volumestore: datastore1/dev), you will find out that 'VIC' is created under dev, but kvStore under dev_1, and 'VIC' and 'kvStore' are both created during the initialization of portlayer if I recall it correctly. (they're deleted correctly)
and path for object is a simple string, but switches to '[datastore1] /xx' encoding sometimes and 'ds://xx' encoding at others. this becomes problematic when conf is shared, since some code changes its encoding and other code depends on the assumption it's already encoded in some way. this essentially puts a restraint on the order code is executed.

maplain:
before we reach an agreement on how this should be addressed, I'll keep the [WIP] in title, though I'm not sure whether this is the way it works.

PR #4856, opened 2017-04-24
[specific ci=Group6-VIC-Machine] Fix the outputted --cert-path

### LINKED ISSUE #4683: vic-machine output and generated env is wrong if --cert-path specified

gigawhitlocks:
Fixes #4683 

Also prevents destruction of provided path after --cert-path if files already exist there
Also causes the browser-friendly certificate to be generated in the right place when --cert-path is specified.

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -1204,7 +1204,8 @@ func (c *Create) generateCertificates(server bool, client bool) ([]byte, *certif
 
 		// If openssl is present, try to generate a browser friendly pfx file (a bundle of the public certificate AND the private key)
 		// The pfx file can be imported directly into keychains for client certificate authentication
-		args := strings.Split(fmt.Sprintf("pkcs12 -export -out ./%[1]s/cert.pfx -inkey ./%[1]s/key.pem -in ./%[1]s/cert.pem -certfile ./%[1]s/ca.pem -password pass:", c.DisplayName), " ")
+		certPath := filepath.Clean(c.certPath)
+		args := strings.Split(fmt.Sprintf("pkcs12 -export -out %[1]s/cert.pfx -inkey %[1]s/key.pem -in %[1]s/cert.pem -certfile %[1]s/ca.pem -password pass:", certPath), " ")
 		// #nosec: Subprocess launching with variable
 		pfx := exec.Command("openssl", args...)
 		out, err := pfx.CombinedOutput()
@@ -1212,7 +1213,7 @@ func (c *Create) generateCertificates(server bool, client bool) ([]byte, *certif
 			log.Debug(out)
 			log.Warnf("Failed to generate browser friendly PFX client certificate: %s", err)
 		} else {
-			log.Infof("Generated browser friendly PFX client certificate - certificate in ./%s/cert.pfx", c.DisplayName)
+			log.Infof("Generated browser friendly PFX client certificate - certificate in %s/cert.pfx", certPath)
 		}
 	}
 
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -1099,7 +1099,7 @@ func (c *Create) generateCertificates(server bool, client bool) ([]byte, *certif
 	// generate the certs and keys with names conforming the default the docker client expects
 	files, err := ioutil.ReadDir(c.certPath)
 	if len(files) > 0 {
-		return nil, nil, fmt.Errorf("Specified directory to store certificates, \"%s\", is not empty. Please specify a new path in which to store generated certificates using --cert-path or remove the contents of \"%s\" and run vic-machine again.", c.certPath, c.certPath)
+		return nil, nil, fmt.Errorf("Specified directory to store certificates is not empty. Specify a new path in which to store generated certificates using --cert-path or remove the contents of \"%s\" and run vic-machine again.", c.certPath)
 	}
 
 	err = os.MkdirAll(c.certPath, 0700)
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -1097,7 +1097,12 @@ func (c *Create) generateCertificates(server bool, client bool) ([]byte, *certif
 
 	var certs []byte
 	// generate the certs and keys with names conforming the default the docker client expects
-	err := os.MkdirAll(c.certPath, 0700)
+	files, err := ioutil.ReadDir(c.certPath)
+	if len(files) > 0 {
+		return nil, nil, fmt.Errorf("Specified directory to store certificates, \"%s\", is not empty. Please specify a new path in which to store generated certificates using --cert-path or remove the contents of \"%s\" and run vic-machine again.", c.certPath, c.certPath)
+	}
+
+	err = os.MkdirAll(c.certPath, 0700)
 	if err != nil {
 		log.Errorf("Unable to make directory to hold certificates (set via --cert-path)")
 		return nil, nil, err
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -1383,7 +1383,7 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 
 	if err = executor.CheckServiceReady(ctx, vchConfig, c.clientCert); err != nil {
 		executor.CollectDiagnosticLogs()
-		cmd, _ := executor.GetDockerAPICommand(vchConfig, c.ckey, c.ccert, c.cacert)
+		cmd, _ := executor.GetDockerAPICommand(vchConfig, c.ckey, c.ccert, c.cacert, c.certPath)
 		log.Info("\tAPI may be slow to start - try to connect to API after a few minutes:")
 		if cmd != "" {
 			log.Infof("\t\tRun command: %s", cmd)
@@ -1396,7 +1396,7 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 
 	log.Infof("Initialization of appliance successful")
 
-	executor.ShowVCH(vchConfig, c.ckey, c.ccert, c.cacert, c.envFile)
+	executor.ShowVCH(vchConfig, c.ckey, c.ccert, c.cacert, c.envFile, c.certPath)
 	log.Infof("Installer completed successfully")
 	return nil
 }
--- a/doc/user/usage.md
+++ b/doc/user/usage.md
@@ -241,7 +241,6 @@ ERRO[2016-11-07T19:53:44Z] vic-machine-linux failed: provide Common Name for ser
 
 The [`--cert-path`](#certificate-names-and---cert-path) option applies to all of the TLS configurations other than --no-tls.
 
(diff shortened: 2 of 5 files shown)


cmd/vic-machine/create/create.go
```diff
@@ -1091,7 +1091,12 @@ func (c *Create) generateCertificates(server bool, client bool) ([]byte, *certif
 
 	var certs []byte
 	// generate the certs and keys with names conforming the default the docker client expects
-	err := os.MkdirAll(c.certPath, 0700)
+	files, err := ioutil.ReadDir(c.certPath)
+	if len(files) > 0 {
```
emlin: hmm... why should we do this? Feels like if user wants to generate certificate, that could be put into whatever directories

gigawhitlocks: Makes sure it's empty before we go writing files there all willy-nilly. Do note that it tries to load certificates first, so we only bother to do this if the load failed. So this way, vic-machine will not overwrite certificates that were inadvertently used for the install that don't work for the VCH.

The old way was: try to load. If failure -> generate certs. If certs exist and have the same name as the outputs, they get overwritten. This is what was happening to me trying to provide a cert and produced an undue amount of confusion because I didn't realize my certs had been overwritten and couldn't figure out why vic-machine wasn't accepting them :(

I can write a test for this behavior if it makes it clearer and/or makes you feel more comfortable with this functionality

emlin: I looked at the certificate validation code, which looks not bad. 
My opinion is that we'd better to stop the installation after a failed validation, instead of here. But might need @andrewtchin comments.

andrewtchin: I'm good with the implementation in this PR

emlin: okay 👍

tests/test-cases/Group6-VIC-Machine/6-13-TLS.robot
```diff
@@ -32,7 +32,27 @@ Create VCH - defaults with --no-tls
     Run Regression Tests
     Cleanup VIC Appliance On Test Server
 
+Create VCH - defaults custom cert path
+    Set Test Environment Variables
+    Run Keyword And Ignore Error  Cleanup Dangling VMs On Test Server
+    Run Keyword And Ignore Error  Cleanup Datastore On Test Server
+
+    ${pwd}=     Run  pwd
```
mhagen-vmware: I think you just want ${CURDIR} or ${EXECDIR}, looks like the latter.

mhagen-vmware: curdir is the absolute path to the directory the test file is located in and execdir is the absolute path to the directory that execution of the script occurred in

tests/test-cases/Group6-VIC-Machine/6-13-TLS.robot
```diff
@@ -32,7 +32,27 @@ Create VCH - defaults with --no-tls
...
+    ${output}=  Run  bin/vic-machine-linux create ${vicmachinetls} --name=%{VCH-NAME} --target="%{TEST_USERNAME}:%{TEST_PASSWORD}@%{TEST_URL}" --thumbprint=%{TEST_THUMBPRINT} --image-store=%{TEST_DATASTORE} --bridge-network=%{BRIDGE_NETWORK} --public-network=%{PUBLIC_NETWORK} --cert-path=${pwd}/foo-bar-certs/
+    Should Contain  ${output}  --tlscacert="${pwd}/foo-bar-certs/ca.pem" --tlscert="${pwd}/foo-bar-certs/cert.pem" --tlskey="${pwd}/foo-bar-certs/key.pem"
```
mhagen-vmware: does this work?  I think you need to at least escape the = signs?

tests/test-cases/Group6-VIC-Machine/6-13-TLS.robot
```diff
@@ -32,7 +32,27 @@ Create VCH - defaults with --no-tls
...
+    ${save_env}=  Run  cat ${pwd}/foo-bar-certs/%{VCH-NAME}.env
+    Should Contain  ${save_env}  DOCKER_CERT_PATH=${pwd}/foo-bar-certs
```
mhagen-vmware: weird intermix here... I would prefer you exhaust all ${output} checks first before jumping to a different testable item.

tests/test-cases/Group6-VIC-Machine/6-13-TLS.robot
gigawhitlocks: yeah this works. I had accidentally put `$(pwd)` on my first iteration and that didn't pass, but this does, which indicates to me that that argument is interpreted as a string

tests/test-cases/Group6-VIC-Machine/6-13-TLS.robot
gigawhitlocks: ${EXECDIR}, thank you, didn't know it was a variable.

tests/test-cases/Group6-VIC-Machine/6-13-TLS.robot
gigawhitlocks: k

gigawhitlocks:
OK @mhagen-vmware 's comments are addressed and assuming that's good for him in the morning hopefully @stuclem can also approve the 1-line doc change and this will be ready to go

doc/user/usage.md
```diff
@@ -241,7 +241,6 @@ ERRO[2016-11-07T19:53:44Z] vic-machine-linux failed: provide Common Name for ser
 
 The [`--cert-path`](#certificate-names-and---cert-path) option applies to all of the TLS configurations other than --no-tls.
 
-Using `--force` will allow a `create` operation to move past some certificate checks performed for existing certificates, however will do so by generating __new__ certificates, __overwriting__ the old.
```
stuclem: @gigawhitlocks I am fine with removing this from usage.md. I checked the user docs and the only thing that we say about using` vic-machine create --force` in relation to certs is that it skips the VC thumbprint check. So, I think that we're fine.

gigawhitlocks: Yes, that part is still valid

cmd/vic-machine/create/create.go
```diff
@@ -1097,7 +1097,12 @@ func (c *Create) generateCertificates(server bool, client bool) ([]byte, *certif
...
+		return nil, nil, fmt.Errorf("Specified directory to store certificates, \"%s\", is not empty. Please specify a new path in which to store generated certificates using --cert-path or remove the contents of \"%s\" and run vic-machine again.", c.certPath, c.certPath)
```
hickeng: I think we can be a little more concise - if certPath is long this error message is going to become unwieldy. It's also imperative so drop the "Please".
`("\"%s\" is not empty. Remove the contents or specify a different path for generated certificates via -cert-path", c.certPath)`

I also agree with @emlin that if we're not overwriting certs then it would make sense to bail immediately after validation of existing certificates fails, with a similar, but specific error message:
`("cannot validate certificates in \"%s\" for use with this VCH. Specify a different path via -cert-path, or remove existing certificates to generate new ones", c.certPath)`

Agree completely that silently overwriting is an issue and that we should change that behaviour unless --force is specified.
Not sure whether `--force` should overwrite, or use unvalidated certs.... It's a very coarse mechanism.

lib/install/management/inspect.go
```diff
@@ -89,11 +89,11 @@ func (d *Dispatcher) InspectVCH(vch *vm.VirtualMachine, conf *config.VirtualCont
 		log.Debugf("No host certificates provided")
 	}
 
-	d.ShowVCH(conf, "", "", "", "")
+	d.ShowVCH(conf, "", "", "", "", "")
```
hickeng: Is there any way we can guess at a cert path and then in ShowVCH check the certs against the VCH for validity? As it stands `inspect` is never going to provide useful output for secure deployments.

We could check:
default cert-path location
DOCKER_CERT_PATH location

for existing certs.

emlin: pulled #4490 into 1.2 milestone, which should address this issue.

gigawhitlocks: We have another issue open to address this in `inspect`, so that will be addressed along with that issue

cmd/vic-machine/create/create.go
gigawhitlocks: @hickeng reviewing the code the only path I see, which is the one I was hitting, where we do not bail entirely upon failing to load certificates is when the `Leaf` of the certificate is `nil`. See the error on line 996 in `create.go`:

```log.Warnf("Failed to load x509 leaf: Unable to confirm server certificate cname matches provided cname %q. Continuing...", c.cname)```

I was hitting this path by using a wildcard certificate, which (apparently) has a `nil` `Leaf` in the certificate itself, presumably because it's not defined in a wildcard certificate due to the nature of that type of certificate. I don't think we want to bail in this situation, or we would block the usage of valid wildcard certificates, which is why this simply emits a warning.

Before my change, however, in that case the certificate would be overwritten on the filesystem if it existed in the same place as specified by `--cert-path`. Basically, this is a final-step cover-our-butts check for ridiculous corner cases like the one described (that I somehow managed to hit myself). The behavior you've requested is already present for the vast majority of cases.

Also I'm going to make that error statement more terse, as requested.

gigawhitlocks:
Went ahead and merged @hickeng -- let me know if you have any further issues with that behavior after seeing the reply I left you above (hidden under an "outdated" block because I updated the error message) and I'm more than happy to refine it post-hoc

PR #4870, opened 2017-04-24
Expose ports on the external network and block all others [specific ci=Group1-Docker-Commands/1-25-Docker-Port-Map]

### LINKED ISSUE #4452: Secure containers using NFS shared volumes from external traffic
When a user adds an NFS volume to a container, and adds the relevant network to the container where the external server is hosted, the container is now completely open to the external network.  To mitigate this, we should add a netfilter rule to DENY all incoming from that external network.

In order to add the rule or any fw configuration to the container need to identify the nfs target server (which we already push to the container), twiddle and the rule via the tether (either via iptc directly, iptables, or netlink to the kernel and make the ADD call directly).

TODO:
- [ ] Add iptc OR netlink to netfilter to the tether
- [ ] Add a rule when the tether is insrtucted to mount an NFS target to DENY incoming from the network the nfs server is on.

fdawg4l:
This change adds iptables to the bootstrap iso and passes the `Ports` bound to the endpoint down to the tether.  We then expose the relevant ports if and only if the interface is not a container network (e.g. an external network) and block all others.  Bridged networks are unaffected.

TBD-  make this work outside of a photon container.  Also adding port maps will be future work.

Fixes #4452

### DIFF
--- a/cmd/tether/ops_linux.go
+++ b/cmd/tether/ops_linux.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -27,6 +28,7 @@ import (
 	"github.com/vmware/vic/lib/iolog"
 	"github.com/vmware/vic/lib/tether"
 	"github.com/vmware/vic/pkg/dio"
+	"github.com/vmware/vic/pkg/netfilter"
 	"github.com/vmware/vic/pkg/trace"
 )
 
@@ -124,6 +126,47 @@ func (t *operations) Setup(sink tether.Config) error {
 	if err = tether.Sys.Syscall.Unmount(runMountPoint, syscall.MNT_DETACH); err != nil {
 		if errno, ok := err.(syscall.Errno); !ok || errno != syscall.EINVAL {
 			return err
+		}
+	}
+
+	return nil
+}
+
+// SetupFirewall sets up firewall rules on the external scope only.  Any
+// portmaps are honored as are port exposes.
+//func (t *operations) SetupFirewall(endpoints map[string]*tether.NetworkEndpoint) error {
+func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
+	if err := netfilter.Flush(context.Background(), ""); err != nil {
+		return err
+	}
+
+	// default rule set
+	established := &netfilter.Rule{
+		Chain:     netfilter.Input,
+		States:    []netfilter.State{netfilter.Established},
+		Interface: "external",
+		Target:    netfilter.Accept,
+	}
+
+	reject := &netfilter.Rule{
+		Chain:     netfilter.Input,
+		Interface: "external",
+		Target:    netfilter.Reject,
+	}
+
+	for _, endpoint := range config.Networks {
+		if endpoint.Name == "external" {
+
+			// XXX handle port maps
+			if err := established.Commit(context.TODO()); err != nil {
+				return err
+			}
+
+			if err := reject.Commit(context.TODO()); err != nil {
+				return err
+			}
+
+			break
 		}
 	}
 
--- a/cmd/tether/ops_linux.go
+++ b/cmd/tether/ops_linux.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strconv"
+	"strings"
 	"syscall"
 
 	log "github.com/Sirupsen/logrus"
@@ -157,18 +159,66 @@ func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
 	for _, endpoint := range config.Networks {
 		if endpoint.Name == "external" {
 
-			// XXX handle port maps
-			if err := established.Commit(context.TODO()); err != nil {
-				return err
+			// XXX It looks like we'd want to collect the errors here, but we
+			// can't.  Since this is running inside init (tether) and tether
+			// reaps all children, the os.exec package won't be able to collect
+			// the error code in time before the reaper does.  The exec package
+			// calls wait and attemps to collect its child, but the reaper will
+			// have raptured the pid before that.  So, best effor, just keep going.
+			established.Commit(context.TODO())
+
+			// handle the ports
+			for _, p := range endpoint.Ports {
+				// parse the port maps
+				r, err := portToRule(p)
+				if err != nil {
+					log.Errorf("can't apply port rule: %s", err.Error())
+					continue
+				}
+
+				log.Infof("Applying rule for port %s", p)
+				r.Commit(context.TODO())
 			}
 
-			if err := reject.Commit(context.TODO()); err != nil {
-				return err
-			}
+			reject.Commit(context.TODO())
 
 			break
 		}
 	}
 
 	return nil
+}
+
+func portToRule(p string) (*netfilter.Rule, error) {
+	if strings.Contains(p, ":") {
+		return nil, errors.New("port maps are TBD")
+	}
+
+	// 9999/tcp
+	s := strings.Split(p, "/")
+	if len(s) != 2 {
+		return nil, errors.New("can't parse port spec: " + p)
+	}
+
+	rule := &netfilter.Rule{
+		Chain:     netfilter.Input,
+		Interface: "external",
+		Target:    netfilter.Accept,
+	}
+
+	switch netfilter.Protocol(s[1]) {
+	case netfilter.UDP:
+		rule.Protocol = netfilter.UDP
+	case netfilter.TCP:
+		rule.Protocol = netfilter.TCP
+	}
+
+	port, err := strconv.Atoi(s[0])
+	if err != nil {
+		return nil, err
+	}
+
+	rule.FromPort = port
+
+	return rule, nil
 }
--- a/cmd/tether/ops_linux.go
+++ b/cmd/tether/ops_linux.go
@@ -28,6 +28,7 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 
 	"github.com/vmware/vic/lib/iolog"
+	"github.com/vmware/vic/lib/portlayer/constants"
 	"github.com/vmware/vic/lib/tether"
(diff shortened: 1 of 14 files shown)


cmd/tether/ops_linux.go
```diff
@@ -129,3 +133,89 @@ func (t *operations) Setup(sink tether.Config) error {
...
+
+// SetupFirewall sets up firewall rules on the external scope only.  Any
+// portmaps are honored as are port exposes.
+func (t *operations) SetupFirewall(endpoints map[string]*tether.NetworkEndpoint) error {
+	// XXX It looks like we'd want to collect the errors here, but we
+	// can't.  Since this is running inside init (tether) and tether
+	// reaps all children, the os.exec package won't be able to collect
+	// the error code in time before the reaper does.  The exec package
+	// calls wait and attemps to collect its child, but the reaper will
+	// have raptured the pid before that.  So, best effor, just keep going.
```
sgairo: "effort"  :)

pkg/netfilter/netfilter_linux.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (r *Rule) args() ([]string, error) {
+	var args []string
+
+	if r.Table != "" {
+		args = append(args, "-t", string(r.Table))
+	}
+
+	args = append(args, "-A", string(r.Chain))
```
hmahmood: Should check if `Chain` is empty.

pkg/netfilter/netfilter_linux.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	if len(r.States) > 0 {
+		for _, state := range r.States {
+			args = append(args, "-m", "state", "--state", string(state))
+		}
+	}
+
+	if r.Interface == "" {
+		return nil, errors.New("interface cannot be empty")
```
hmahmood: Not sure if this is true. It may be better to have the `iptables` command check what is required.

lib/portlayer/network/context.go
```diff
@@ -1073,10 +1086,20 @@ func (c *Context) AddContainer(h *exec.Handle, options *AddContainerOptions) err
 		h.ExecConfig.Networks = make(map[string]*executor.NetworkEndpoint)
 	}
 
+	// If the scope type is external, name the nic as such.  We can only have
```
hmahmood: So this is true currently, but I am not sure this will be or should be true in the future. I would just go with a randomly generated name if that is possible.

pkg/netfilter/netfilter_linux.go
hmahmood: Also would be nice to have an option to do `-I`.

cmd/tether/ops_linux.go
```diff
@@ -129,3 +133,89 @@ func (t *operations) Setup(sink tether.Config) error {
...
+	netfilter.Flush(context.Background(), "")
+
+	// default rule set
+	established := &netfilter.Rule{
+		Chain:     netfilter.Input,
+		States:    []netfilter.State{netfilter.Established},
```
hmahmood: I believe this should also have `CONNECTED` as well.

fdawg4l: Sure, but we arent using it :-)

lib/portlayer/network/context.go
fdawg4l: Unless something changed since rebasing, i didn't observe it to be true.  I'll try on master and get back to you.

The problem is the network is configured in a different routine than setup firewall.  We could pass the scope type down along with the Port in the `tether.NetworkEndpoint`.  Is that what you're suggesting?  I didn't want to add redundant fields and I'm not that familiar with the API here so suggests are welcome.  This just seemed the most straight forward.

pkg/netfilter/netfilter_linux.go
fdawg4l: re: empty, see `Flush`.  It's up to the caller to do the right thing when.

Re: `-I`, maybe?  I'd need to think about it.

pkg/netfilter/netfilter_linux.go
fdawg4l: Good point.  I'll omit this check.

lib/portlayer/network/context.go
hmahmood: The `Network.Type` field in the `NetworkEndpoint` already has the type, so you can use that.

fdawg4l: I'm not sure how I missed that.  I'll try it!  Thanks!

lib/apiservers/engine/backends/container.go
```diff
@@ -1097,31 +1099,36 @@ func (c *Container) defaultScope() string {
...
+		return nil, nil
 	}
 
-	if len(listRes.Payload) != 1 || listRes.Payload[0].ScopeType != "bridge" {
+	if l := len(listRes.Payload); l != 1 {
+		log.Warnf("found %d scopes", l)
+		return nil, nil
+	}
+
+	if listRes.Payload[0].ScopeType != "bridge" {
```
hmahmood: Should use the constant defined in the `github.com/vmware/vic/lib/portlayer/constants` package.

cmd/tether/ops_linux.go
```diff
@@ -129,3 +133,105 @@ func (t *operations) Setup(sink tether.Config) error {
...
+		Target: netfilter.Accept,
+	}
+
+	reject := &netfilter.Rule{
+		Chain:  netfilter.Input,
+		Target: netfilter.Reject,
+	}
+
+	for _, endpoint := range config.Networks {
+		if endpoint.Network.Type == "external" {
```
hmahmood: Should use the constant defined in the `github.com/vmware/vic/lib/portlayer/constants` package.

lib/tether/ops_linux.go
```diff
@@ -891,6 +891,10 @@ func (t *BaseOperations) Cleanup() error {
 	return nil
 }
 
+func (t *BaseOperations) SetupFirewall(endpoints map[string]*NetworkEndpoint) error {
```
hmahmood: Is this needed? If it is, isn't the argument supposed to be `*ExecutorConfig`?

fdawg4l: It's not needed.  I'll revert this file.  Thanks!

cmd/tether/ops_linux.go
```diff
@@ -129,3 +133,105 @@ func (t *operations) Setup(sink tether.Config) error {
...
+		Chain:     netfilter.Input,
+		Interface: "external",
+		Target:    netfilter.Accept,
+	}
+
+	switch netfilter.Protocol(s[1]) {
+	case netfilter.UDP:
+		rule.Protocol = netfilter.UDP
+	case netfilter.TCP:
+		rule.Protocol = netfilter.TCP
```
caglar10ur: should we panic or return an error here for the default case?

isos/bootstrap/bootstrap
```diff
@@ -34,6 +46,27 @@ if mount -t ext4 /dev/disk/by-label/containerfs ${MOUNTPOINT}; then
 
     cp /bin/tether ${MOUNTPOINT}/.tether/tether
 
+
+    # linux-vdso.so.1 (0x00007ffc94d0d000)
```
caglar10ur: can you add a short note here explaining why we are doing this? Possibly with a "No Animals Were Harmed" disclaimer :)

caglar10ur:
LGTM

cmd/tether/ops_linux.go
fdawg4l: oops.  will return an error here.

isos/bootstrap/bootstrap
fdawg4l: This is a good question.  We're doing it because `netfilter` doesnt have an appropriate userspace api that one can write against.  All disclaimers say we should use `iptables-restore`.  Kinda dumb, if you ask me.  So we're reduced to copying binaries to make it work in the `initrd`.  I'll add a comment.  Thanks!

PS.  I was harmed.  Does that count?

fdawg4l:
OK, this works on busybox so this should work on all images now.  The magic was calling `iptables` from `ld`.

cmd/tether/ops_windows.go
```diff
@@ -149,3 +149,7 @@ func (t *operations) Setup(sink tether.Config) error {
 
 	return nil
 }
+
+func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
+	return errors.New("Not implended on windows")
```
cgtexmex: misspelled implemented...

fdawg4l: thanks

lib/apiservers/engine/backends/container.go
```diff
@@ -963,16 +964,18 @@ func (c *Container) containerStart(name string, hostConfig *containertypes.HostC
...
-		e := c.findPortBoundNetworkEndpoint(hostConfig, endpoints)
-		if err = MapPorts(hostConfig, e, id); err != nil {
-			return InternalServerError(fmt.Sprintf("error mapping ports: %s", err))
-		}
-
-		defer func() {
-			if err != nil {
-				UnmapPorts(hostConfig)
+		scope, e := c.findPortBoundNetworkEndpoint(hostConfig, endpoints)
+		if scope != nil && scope.ScopeType == "bridged" {
```
cgtexmex: Why not use `constants.BridgeScopeType` ?

cmd/tether/ops_windows.go
```diff
@@ -149,3 +149,7 @@ func (t *operations) Setup(sink tether.Config) error {
 
 	return nil
 }
+
+func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
+	return errors.New("Not implented on windows")
```
cgtexmex: close...but not quite....implemented...

fdawg4l: Wow.  I'm amazing.

cgtexmex:
LGTM

PR #4874, opened 2017-04-25
Stats converter unit tests use retry pkg to avoid timing failures

### LINKED ISSUE #4873: stats unit test fails on loaded system
There is a `docker stats` unit [test](https://github.com/vmware/vic/blob/master/lib/apiservers/engine/backends/convert/stats_test.go#L187) that attempts to verify that resources are cleaned up when the func ends.  In a loaded environment this takes longer than the allocated 500ms.  This has resulted in a number of CI failures -- https://ci.vmware.run/vmware/vic/10032 -- being the latest.

cgtexmex:
The stats converter unit tests would fail in a loaded
environment.  Now using the retry package to mitigate
timing issues.

Fixes #4873


mdubya66:
lgtm

PR #4919, opened 2017-04-26
Correct handling of incorrect --id option [specific ci=6-11-Debug]

### LINKED ISSUE #4918: Panic in `vic-machine debug` if incorrect id is specified
Observed in a debugging session - if an incorrect ID is specified a panic results.
Looking at the code this would apply to:
* inspect
* debug
* delete

hickeng:
Previously this was resulting in a panic. This now reports that the id
provided could not be found.

Fixes #4918


lib/install/management/finder.go
```diff
@@ -55,7 +55,7 @@ func (d *Dispatcher) NewVCHFromID(id string) (*vm.VirtualMachine, error) {
 			return nil, err
 		}
 		log.Debugf("Appliance is not found")
-		return nil, nil
+		return nil, fmt.Errorf("id \"%s\" could not be found", id)
```
caglar10ur: use %q instead

PR #4920, opened 2017-04-26
Use UUID instead of symlink for endpoint VM

jzt:
This is a speculative change that uses the actual UUID directory name instead of the symlink to avoid the scenario where the symlink has not presented by the time a VMDK creation requires it. This issue pertains primarily to vSAN environments under load (like longevity/nightly tests).

Longevity tests with these changes are currently running on bart (VC 6.0).

Applies to #4666

### DIFF
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -600,8 +600,14 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 	},
 	)
 
+	// fix up those parts of the config that depend on the final applianceVM folder name
 	conf.BootstrapImagePath = fmt.Sprintf("[%s] %s/%s", conf.ImageStores[0].Host, d.vmPathName, settings.BootstrapISO)
 
+	if len(conf.ImageStores[0].Path) == 0 {
+		conf.ImageStores[0].Path = d.vmPathName
+	}
+
+	// apply the fixed-up configuration
 	spec, err = d.reconfigureApplianceSpec(vm2, conf, settings)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


lib/install/validate/storage.go
```diff
@@ -44,7 +44,7 @@ func (v *Validator) storage(ctx context.Context, input *data.Data, conf *config.
 
 	// provide a default path if only a DS name is provided
 	if imageDSpath.Path == "" {
-		imageDSpath.Path = input.DisplayName
+		log.Debug("No path specified on datastore for image store - will use the endpointVM folder")
```
hickeng: I know I wrote this but thinking that a more pedantically correct message would be:
`No path element specified for image store - will use default`

As the validator isn't responsible for populating the path element it shouldn't be making assertions about what the value will be.

mhagen-vmware:
lgtm

lib/install/management/appliance.go
```diff
@@ -596,8 +596,14 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 	},
 	)
 
+	// fix up those parts of the config that depend on the final applianceVM folder name
 	conf.BootstrapImagePath = fmt.Sprintf("[%s] %s/%s", conf.ImageStores[0].Host, d.vmPathName, settings.BootstrapISO)
 
+	if len(conf.ImageStores[0].Path) == 0 {
```
vburenin: I suspect there is no doubt that ImageStores array has at least one element, correct? Is there any chance it does'n?

jzt: Correct, it will error before this point if no default datastore is found (in the event the user doesn't supply one at create time).

PR #4951, opened 2017-04-27
Fix broken Flex Client UI plugin

### LINKED ISSUE #4948: UI plugin is broken for the Flex plugin
**User Statement:**
As a UI engineer I want to make sure the plugins (both the Flex and HTML5 ones) always work as expected.

**Expected behavior:**
Plugins should appear in their respective platform.

**Actual behavior:**
The Flex client plugin doesn't show up in the Flex client. Plugin is registered fine and didn't throw any error in the log message, but portlets are not visible. This is because of Cheng's change a couple weeks ago that changed some key names in extraConfig that the portlets were using. Although I made an according change for the HTML5 Client I forgot to do the same for the Flex client plugin.

**Acceptance Criteria:**
Flex client should be back up.

jooskim:
This commit updates keys in config.extraConfig that are used by
the Flex Client UI plugin.

Fixes #4948

### DIFF
--- a/ui/vic-ui/swf/src/main/flex/com/vmware/vicui/constants/AppConstants.as
+++ b/ui/vic-ui/swf/src/main/flex/com/vmware/vicui/constants/AppConstants.as
@@ -2,10 +2,10 @@ package com.vmware.vicui.constants {
 
 	public class AppConstants {
 
-		public static const VM_CONTAINER_NAME_PATH:String = "guestinfo.vice./common/name";
+		public static const VM_CONTAINER_NAME_PATH:String = "common/name";
 		public static const VM_CONTAINER_IMAGE_PATH:String = "guestinfo.vice./repo";
 		public static const VM_CONTAINER_PORTMAPPING:String = "guestinfo.vice./networks|bridge/ports~";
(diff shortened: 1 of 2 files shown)


ui/vic-ui/swf/src/main/flex/com/vmware/vicui/constants/AppConstants.as
```diff
@@ -2,10 +2,10 @@ package com.vmware.vicui.constants {
 
 	public class AppConstants {
 
-		public static const VM_CONTAINER_NAME_PATH:String = "guestinfo.vice./common/name";
+		public static const VM_CONTAINER_NAME_PATH:String = "common/name";
 		public static const VM_CONTAINER_IMAGE_PATH:String = "guestinfo.vice./repo";
 		public static const VM_CONTAINER_PORTMAPPING:String = "guestinfo.vice./networks|bridge/ports~";
```
mhagen-vmware: do we not need to get rid of these vice references too?

jooskim: those remaining vice keys are still there and used by the portlets.

Issue #2928, opened 2016-10-27
Deleting containers VMs in vSphere may cause containers that use the same base image to fail to create.


mlh78750:
See #2927. Users should not delete container vms from the vsphere side. However, this is extremely non-intuitive for those users. 

So we need to protect the VCH from having the images removed out from under them by VC. We should have a test that covers these conditions and that is tracked in #2925

mlh78750:
Possible fix from @hickeng: I'm only aware of one workaround, which is to create an additional child VMDK of all the entries in the disk chain so that VC will consider the disks still in use.

fdawg4l:
Weirdly, I'm having trouble reproducing this.  I know im doing something wrong but working on what that is.  :-)

```
$ docker -H 192.168.8.174:2376 --tls ps -a                                                                                                                
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
4f20f21b111c        busybox             "/bin/top"          41 seconds ago      Up 39 seconds                           tender_heisenberg

$ govc ls vm
/ha-datacenter/vm/virtual-container-host
/ha-datacenter/vm/tender_heisenbe-4f20f21b111c53134ce3302ccfa3627c548e9d5478c2af1a31334c237e049701

$ govc vm.destroy /ha-datacenter/vm/tender_heisenbe-4f20f21b111c53134ce3302ccfa3627c548e9d5478c2af1a31334c237e049701

$ govc ls vm
/ha-datacenter/vm/virtual-container-host

$ docker -H 192.168.8.174:2376 --tls ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES

$ docker -H 192.168.8.174:2376 --tls images
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
busybox             latest              e292aa76ad3b        2 weeks ago         1.093 MB

$ docker -H 192.168.8.174:2376 --tls run -d busybox /bin/top                                                                                              
1af6d08f64b4ec703156348ff3d7f7275041c4667a5b603e74ea0d2ae3cbbb29

$ docker -H 192.168.8.174:2376 --tls ps -a                                                                                                                
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
1af6d08f64b4        busybox             "/bin/top"          18 seconds ago      Up 15 seconds                           trusting_chandrasekhar
```

fdawg4l:
[...]

PR #4965, opened 2017-04-28
Disable VM delete action in VC UI

dougm:
Issue #2928

### DIFF
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -512,6 +512,8 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 	}
 	vm2 := vm.NewVirtualMachineFromVM(d.ctx, d.session, gvm)
 
+	vm2.DisableDestroy(d.ctx)
+
 	// update the displayname to the actual folder name used
 	if d.vmPathName, err = vm2.FolderName(d.ctx); err != nil {
 		log.Errorf("Failed to get canonical name for appliance: %s", err)
--- a/pkg/vsphere/vm/vm.go
+++ b/pkg/vsphere/vm/vm.go
@@ -588,3 +588,29 @@ func (vm *VirtualMachine) SetVCHUpdateStatus(ctx context.Context, status bool) e
 
 	return err
 }
+
+// DisableDestroy attempts to disable the VirtualMachine.Destroy_Task method.
+// When the method is disabled, the VC UI will disable/grey out the VM "delete" action.
+// Requires the "Global.Disable" VC privilege.
+// The VirtualMachine.Destroy_Task method can still be invoked via the API.
+func (vm *VirtualMachine) DisableDestroy(ctx context.Context) {
+	if !vm.IsVC() {
+		return
+	}
+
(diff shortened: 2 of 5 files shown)


pkg/vsphere/vm/vm.go
```diff
@@ -588,3 +588,29 @@ func (vm *VirtualMachine) SetVCHUpdateStatus(ctx context.Context, status bool) e
 
 	return err
 }
+
+// DisableDestroy attempts to disable the VirtualMachine.Destroy_Task method.
+// When the is disabled, the VC UI will disable/grey out the VM "delete" action.
```
chengwang86: when the "  " is disabled

pkg/vsphere/vm/vm.go
```diff
@@ -588,3 +588,29 @@ func (vm *VirtualMachine) SetVCHUpdateStatus(ctx context.Context, status bool) e
...
+	if !vm.IsVC() {
+		return
+	}
+
+	m := object.NewAuthorizationManager(vm.Vim25())
+
+	method := []object.DisabledMethodRequest{
+		{
+			Method: "Destroy_Task",
+			Reason: "Managed by VIC Engine",
```
chengwang86: Does this `reason` show up somewhere on the UI (e.g. via a msg box) or is it just for our own reference?

chengwang86: maybe we can have a general method here which takes as input a list of the names of the methods to be disabled. It doesn't matter at the moment though since we only want to disable OOB deletion.

lib/install/management/appliance.go
```diff
@@ -508,6 +508,8 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 	}
 	vm2 := vm.NewVirtualMachineFromVM(d.ctx, d.session, gvm)
 
+	vm2.DisableDestroy(d.ctx)
```
chengwang86: so are we also disabling OOB deletion of the applianceVM?

dougm: Yes

pkg/vsphere/vm/vm.go
dougm: I'll fix that

pkg/vsphere/vm/vm.go
dougm: Yeah, AuthorizationManager.DisableMethods already is that general method.  I'll leave it this way for now, easy to change if want to grey out more methods.

dougm: I haven't seen Reason show up in the UI, it can be accessed from another internal API.

chengwang86:
lgtm

pkg/vsphere/vm/vm.go
hickeng: Can we expand on the reason:
`Managed by VIC Engine - direct deletion will delete image hierarchy and attached volumes`

or a link to the corresponding KB when @dbarkelew has published it.

dougm: @hickeng done

dougm:
@emlin done

Issue #4858, opened 2017-04-24
Nightly 6.5 HA test fails during rmi busybox


rajanashok:
```KEYWORD ${rc}, ${output} = OperatingSystem . Run And Return Rc And Output docker %{VCH-PARAMS} rmi busybox
Documentation:	
Runs the given command in the system and returns the RC and output.
Start / End / Elapsed:	20170423 21:38:08.550 / 20170423 21:38:10.326 / 00:00:01.776
21:38:08.558	INFO	Running command 'docker -H 10.160.68.255:2376 --tls rmi busybox 2>&1'.	
21:38:10.326	INFO	${rc} = 1	
21:38:10.326	INFO	${output} = Error response from daemon: [DELETE /storage/{store_name}/info/{id}][500] DeleteImage default  &{Code:500 Message:Can't find parent http:///storage/images/4209e5d7-0c7c-6382-7742-5ce5e99ae0c9/d0fb090f...

**Comment by rajanashok** [comment]
[5-4-High-Availability.zip](https://github.com/vmware/vic/files/952917/5-4-High-Availability.zip)

**Comment by chengwang86** [comment]
Also seen in nightly test with build 9992

[5-4-High-Availability.zip](https://github.com/vmware/vic/files/959344/5-4-High-Availability.zip)

**Comment by chengwang86** [comment]
Here is a successful run of `docker rmi busybox` on vic  (with an nfsDatastore)
```
sudo docker -H 10.160.151.18:2375 rmi busybox
Deleted: a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4
Deleted: 7520415ce76232cdd62ecc345cea5ea44f5b6b144dc62351f2cd2b08382532a3
Untagged: busybox:latest
```
From portlayer.log
```
Apr 27 2017 17:02:30.247Z DEBUG op=252.12 (delta:9.112µs): [NewOperation] op=252.12 (delta:2.915µs) [github.com/vmware/vic/lib/apiservers/portlayer/restapi/handlers.(*StorageHandlersImpl).DeleteImage:236]
Apr 27 2017 17:02:30.247Z INFO  op=252.12 (delta:32.101µs): DeleteBranch: deleting branch starting at http:///storage/images/42237a29-63bc-4bdc-cfd1-ef387bc16974/d7f9d170aa714bf287eb994aada10a57f4fb1bf8fd748cbfa36a7ab9549190c1
[...]

PR #4968, opened 2017-04-28
[skip ci] add more log information to debug nightly test failure in 5-4 HA test

chengwang86:
Whenever there is an `image`-related operation in the test, we log the output of `govc datastore.ls` and see if all necessary image files are there. 

This relates t #4858

### DIFF
--- a/tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
+++ b/tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
@@ -17,6 +17,73 @@ Documentation  Test 5-4 - High Availability
 Resource  ../../resources/Util.robot
 Suite Teardown  Nimbus Cleanup  ${list}
 
+*** Keywords ***
+Check ImageStore
+    ${rc}  ${output}=  Run And Return Rc And Output  govc datastore.ls -R -ds=nfsDatastore %{VCH-NAME}/VIC
+    Should Be Equal  ${rc}  0
+    Log To Console  ${output}
+
+Run Regression Test With More Log Information
+    Check ImageStore
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull busybox
+    Should Be Equal As Integers  ${rc}  0
+    Check ImageStore
+    # Pull an image that has been pulled already
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull busybox
+    Should Be Equal As Integers  ${rc}  0
+    Check ImageStore
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} images
+    Should Be Equal As Integers  ${rc}  0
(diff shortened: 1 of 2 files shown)


tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
```diff
@@ -17,6 +17,78 @@ Documentation  Test 5-4 - High Availability
 Resource  ../../resources/Util.robot
 Suite Teardown  Nimbus Cleanup  ${list}
 
+*** Keywords ***
+Check VM Info
+    ${rc}  ${output}=  Run And Return Rc And Output  govc vm.info \*
+    Should Be Equal  ${rc}  0
+    Log To Console  ${output}
```
mhagen-vmware: change to Log not Log To Console

mhagen-vmware: and we have a get vm info in resources vsphere-util already, it works slightly differently but can that not work for you as defined?

tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
```diff
@@ -17,6 +17,78 @@ Documentation  Test 5-4 - High Availability
...
+
+Check ImageStore
+    ${rc}  ${output}=  Run And Return Rc And Output  govc datastore.ls -R -ds=nfsDatastore %{VCH-NAME}/VIC
+    Should Be Equal  ${rc}  0
+    Log To Console  ${output}
```
mhagen-vmware: log not log to console
-ds should be %{TEST_DATASTORE}
also, if it doesn't exist already this seems valuable to be placed in resources/vsphere-util

tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
```diff
@@ -17,6 +17,78 @@ Documentation  Test 5-4 - High Availability
...
+
+Run Regression Test With More Log Information
```
mhagen-vmware: This is fine here, but should have a github check statement that will let us know to remove this when we close that issue

tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
chengwang86: Here we want to list all the vms whereas the existing `Get VM Info` keyword takes `${vm}` as input. I think it is more convenient to use `govc vm.info \*` here. And we would remove these additional logs after we fix this issue.

mhagen-vmware: but it just passes ${vm} on wards so you can do Get VM Info    \*
The only problem is if the -r is a problem for you?

chengwang86: Oh you are right 👍

PR #5007, opened 2017-05-03
[specific ci=Group1-Docker-Commands] POC: refactor vic-machine create to use portlayer

### LINKED ISSUE #4096: Research Spike: Design vic-machine reconfigure
**Story**
As a VIC developer I want a design document that details how we will implement reconfigure in vic-machine. This document should address following issues:

- [x] split current `vic-machine create` into create and configure (most of the create options are relevant to configure)
- [x] add an installer mechanism to load the VCH config into a [Data structure](https://github.com/vmware/vic/blob/master/lib/install/data/data.go#L29)
- [x] configure and upgrade should combine provided CLI options with the loaded Data (do not support _removal_ of elements such as image stores, volume stores or container networks - that is significant extra complexity)
- [x] configuration is validated using the full path used by create currently, not the shortcut that upgrade added (this shortcut should be removed).
- [x] the configuration should be committed to the VCH in much the same manner on all paths, create, upgrade, and reconfigure (appliance powered off, snapshot taken, config applied, ....)
- [x] leveraging portlayer for vic-machine operations

**Acceptance**
This should deliver a design doc for reconfigure epic
Add stories for vic-machine refactories

### LINKED ISSUE #4619: vic-machine reconfigure command
As vic-machine developer, I'd like to combine upgrade and reconfigure workflow, as upgrade is actually one special reconfiguration of VCH, to improve command usability and code maintainability. 

#4497 tracked the document for vic-machine reconfigure commands

Acceptance:
vic-machine upgrade works as before
vic-machine reconfigure works from end2end without any options included

emlin:
Refactor vic-machine create to use portlayer libraries, instead of portlayer service running in VCH.
Portlayer assumes itself is managing container only, so there have quite a few changes in portlayer, and also need to expose more information from it to satisfy vic-machine requirements.

This code change only works for vic-machine create only.
I would expect this POC can clarify the requirement from vic-machine to portlayer
Fixes #4096

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -1440,7 +1440,9 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 		}
 	}()
 
-	logErr := func() {
+	// error will be checked in check service ready
+	executor.StartAppliance(ctx)
+	if err = executor.CheckServiceReady(ctx, vchConfig, c.clientCert); err != nil {
 		executor.CollectDiagnosticLogs()
 		cmd, _ := executor.GetDockerAPICommand(vchConfig, c.ckey, c.ccert, c.cacert, c.certPath)
 		log.Info("\tAPI may be slow to start - try to connect to API after a few minutes:")
@@ -1450,12 +1452,6 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 			log.Infof("\t\tRun %s inspect to find API connection command and run the command if ip address is ready", clic.App.Name)
 		}
 		log.Info("\t\tIf command succeeds, VCH is started. If command fails, VCH failed to install - see documentation for troubleshooting.")
-	}
-
-	// error will be checked in check service ready
-	executor.StartAppliance(ctx)
-	if err = executor.CheckServiceReady(ctx, vchConfig, c.clientCert); err != nil {
-		logErr()
 		return err
 	}
 
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -1440,7 +1440,7 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 		}
 	}()
 
-	if err = executor.CheckServiceReady(ctx, vchConfig, c.clientCert); err != nil {
+	logErr := func() {
 		executor.CollectDiagnosticLogs()
 		cmd, _ := executor.GetDockerAPICommand(vchConfig, c.ckey, c.ccert, c.cacert, c.certPath)
 		log.Info("\tAPI may be slow to start - try to connect to API after a few minutes:")
@@ -1450,6 +1450,12 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 			log.Infof("\t\tRun %s inspect to find API connection command and run the command if ip address is ready", clic.App.Name)
 		}
 		log.Info("\t\tIf command succeeds, VCH is started. If command fails, VCH failed to install - see documentation for troubleshooting.")
+	}
+
+	// error will be checked in check service ready
+	executor.StartAppliance(ctx)
+	if err = executor.CheckServiceReady(ctx, vchConfig, c.clientCert); err != nil {
+		logErr()
 		return err
 	}
 
--- a/lib/config/virtual_container_host.go
+++ b/lib/config/virtual_container_host.go
@@ -41,6 +41,8 @@ const (
 	Name = "{name}"
 	// ID represents the VCH in creating status, which helps to identify VCH VM which still does not have a valid VM moref set
 	CreatingVCH = "CreatingVCH"
+	// VCHPrefix represents VCH executor configuration encoding prefix
+	VCHPrefix = "init"
 )
 
 // Can we just treat the VCH appliance as a containerVM booting off a specific bootstrap image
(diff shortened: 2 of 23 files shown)


cmd/vic-machine/create/create.go
```diff
@@ -1440,7 +1440,7 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 		}
 	}()
 
-	if err = executor.CheckServiceReady(ctx, vchConfig, c.clientCert); err != nil {
+	logErr := func() {
```
vburenin: This is an arguable change. I would not make a closure here. keeping this logic where it was is ok.

lib/config/virtual_container_host.go
```diff
@@ -228,7 +233,7 @@ func (t *VirtualContainerHostConfigSpec) SetIsCreating(creating bool) {
 
 // IsCreating is checking if this configuration is for one creating VCH VM
 func (t *VirtualContainerHostConfigSpec) IsCreating() bool {
-	return t.ExecutorConfig.ID == CreatingVCH
+	return strings.HasPrefix(t.ExecutorConfig.ID, CreatingVCH)
```
vburenin: I better name CreatingVCH as CreatingVCHPrefix or something like that. I also would use "creating" instead of "CreatingVCH" to be consistent with "init"

lib/install/management/create.go
```diff
@@ -46,15 +46,23 @@ func (d *Dispatcher) CreateVCH(conf *config.VirtualContainerHostConfigSpec, sett
 		return err
 	}
 
-	if err = d.createAppliance(conf, settings); err != nil {
+	d.applianceID, err = d.createAppliance(conf, settings)
```
vburenin: d.applianceID will be affected here in error case as well. Is it a desirable behavior?
I would not assign any value to d.applianceID in case of an error.
It seems like that it doesn't really matter, but is from readability stand point it is a little confusing.

lib/install/management/create.go
```diff
@@ -91,14 +101,22 @@ func (d *Dispatcher) createPool(conf *config.VirtualContainerHostConfigSpec, set
...
 	defer trace.End(trace.Begin(""))
 
+	var h interface{}
 	var err error
-	_, err = d.appliance.WaitForResult(d.ctx, func(ctx context.Context) (tasks.Task, error) {
-		return d.appliance.PowerOn(ctx)
-	})
 
+	if h = d.pl.NewHandle(ctx, d.applianceID); h == nil {
+		err = errors.Errorf("Unable to get handle %s: %s", d.applianceID, err)
```
vburenin: err is nil at this point. it should not be used
```return errors.Errorf("Unable to get handle for %s:", d.applianceID)```
should be enough.

File: lib/install/pllib/pl_lib.go [hunk 5]
```diff
@@ -0,0 +1,316 @@
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
```
vburenin: why is it called pllib? it seems like it is PortLayer client, right?

lib/install/pllib/pl_lib.go
```diff
@@ -0,0 +1,316 @@
...
+	if c == nil {
+		return nil
+	}
+	return c.NewHandle(ctx)
+}
+
+func (pl *Client) AddTask(ctx context.Context, h interface{}, t *executor.SessionConfig) (interface{}, error) {
+	handle, ok := h.(*exec.Handle)
+	if !ok {
+		return nil, fmt.Errorf("Type assertion failed for %#+v", handle)
```
vburenin: Should not we panic here? that would be a clear bug.

lib/install/pllib/pl_lib.go
```diff
@@ -0,0 +1,316 @@
...
+
+	for name, endpoint := range networks {
+		if pnic, ok := nets[endpoint.Network.Common.ID]; ok {
+			// there's already a NIC on this network
+			endpoint.Common.ID = pnic.Common.ID
+			log.Infof("Network role %q is sharing NIC with %q", name, pnic.Network.Common.Name)
+			continue
+		}
+
+		moref := new(types.ManagedObjectReference)
```
vburenin: try to stick to one approach of creating object. We tend to use:
```
moref := &types.ManagedObjectReference{}
```

lib/config/virtual_container_host.go
emlin: the string "CreatingVCH" is used in old version's vch, it'd be better not to change that value. In addition, I'd like to use something special for VCH, to make sure it's less possible to be confused by any other product's VM attributes.

lib/install/management/create.go
emlin: good catch

lib/install/pllib/pl_lib.go
emlin: initially I named it as plclient, but later, I figured out portlayer function is not enough for vic-machine, so it's changed to library, cause a few server functions added

lib/install/pllib/pl_lib.go
emlin: these are supposed to be in portlayer, so should not panic anyway

emlin:
Am closing this PR for I'm not going to merge it. Just keep it as a requirement for portlayer

Issue #4618, opened 2017-04-06
vic-machine inspect: show existing VCH configuration


emlin:
As vic admin, I need to know existing VCH configurations easily to regenerate reconfigure or create command line, so that I can reconfigure VCH or use it recreate another VCH with similar configuration.

Refer to #4497 for the command design and output format

Update:  a scenario exists wherein resources can be renamed through the vSphere UI, rendering the output of this command inaccurate. Special checks need to be made to ensure truthful output in this situation on the following items:
- [ ] networks/portgroups (bridge, client, public, management, container) - possibly using the slot number?
- [ ] datastore resources (image store, volume store, container store)
- [ ] compute resources

jzt:
The first iteration implemented in #5046. Currently, we simply store the formatted command options originally used to create the VCH. The next step is to actually verify the resources on the backend in the event that any were renamed, as well as detecting the OS on which the `inspect` command is being run, so that we format the output correctly.

PR #5011, opened 2017-05-03
First pass at vic-machine inspect --configuration

jzt:
This is the first pass at `vic-machine inspect --conf`. The purpose of this change is to print the command (flags & their args) to be used in order to deploy a VCH with the same configuration as the target VCH that is being inspected. This is the first iteration, and will display the original command used to create the VCH. Further iterations will reflect values that have been changed through the VC backend, such as compute resources or port groups being renamed, etc.

Output:
```
May  3 2017 21:15:09.710Z INFO  Target VCH created with the following options:

        --target=https://192.168.218.149
        --thumbprint=59:72:60:DE:70:D4:C2:6E:B7:FB:82:9D:15:D9:64:BF:47:31:99:09
        --name=VCH-0-4927
        --compute-resource=/ha-datacenter/host/localhost.localdomain/Resources
        --image-store=datastore1
        --volume-store=datastore1/test:default
        --bridge-network=VCH-0-4927-bridge
        --public-network=VM Network
        --no-tlsverify=true
        --appliance-iso=bin/appliance.iso
        --bootstrap-iso=bin/bootstrap.iso
        --force=true
        --timeout=1m0s
        --debug=1
```

First pass at #4618

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -27,6 +27,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"time"
 
@@ -102,16 +103,16 @@ type Create struct {
 	noTLSverify     bool
 	advancedOptions bool
 
-	clientCAs   cli.StringSlice
-	registryCAs cli.StringSlice
+	clientCAs   cli.StringSlice `arg:"tls-ca"`
+	registryCAs cli.StringSlice `arg:"registry-ca"`
 
-	containerNetworks         cli.StringSlice
-	containerNetworksGateway  cli.StringSlice
-	containerNetworksIPRanges cli.StringSlice
-	containerNetworksDNS      cli.StringSlice
-	volumeStores              cli.StringSlice
-	insecureRegistries        cli.StringSlice
-	dns                       cli.StringSlice
+	containerNetworks         cli.StringSlice `arg:"container-network"`
+	containerNetworksGateway  cli.StringSlice `arg:"container-network-gateway"`
+	containerNetworksIPRanges cli.StringSlice `arg:"container-network-ip-range"`
+	containerNetworksDNS      cli.StringSlice `arg:"container-network-dns"`
+	volumeStores              cli.StringSlice `arg:"volume-store"`
+	insecureRegistries        cli.StringSlice `arg:"insecure-registry"`
+	dns                       cli.StringSlice `arg:"dns-server"`
 	clientNetworkName         string
 	clientNetworkGateway      string
 	clientNetworkIP           string
@@ -138,6 +139,28 @@ func NewCreate() *Create {
 	create.Data = data.NewData()
 
 	return create
+}
+
+// SetFields iterates through the fields in the Create struct, searching for fields
+// tagged with the `arg` key. If the value of that tag matches the supplied `flag`
+// string, a nil check is performed. If the field is not nil, then the user supplied
+// this flag on the command line and we need to persist it.
+// This is a workaround for cli.Context.IsSet() returning false when
+// the short option for a cli.StringSlice is supplied instead of the long option.
+// See https://github.com/urfave/cli/issues/314
+func (c *Create) SetFields() map[string]struct{} {
+	result := make(map[string]struct{})
+	create := reflect.ValueOf(c).Elem()
+	for i := 0; i < create.NumField(); i++ {
+		t := create.Type().Field(i)
+		if tag := t.Tag.Get("arg"); tag != "" {
+			ss := create.Field(i)
+			if !ss.IsNil() {
+				result[tag] = struct{}{}
+			}
+		}
+	}
+	return result
 }
 
 // Flags return all cli flags for create
@@ -1273,15 +1296,23 @@ func (c *Create) generateCertificates(server bool, client bool) ([]byte, *certif
 	return cakp.CertPEM, skp, nil
 }
 
-func logArguments(cliContext *cli.Context) {
+func (c *Create) logArguments(cliContext *cli.Context) []string {
+	args := []string{}
+	sf := c.SetFields() // StringSlice options set by the user
+
 	for _, f := range cliContext.FlagNames() {
-		if !cliContext.IsSet(f) {
+		_, ok := sf[f]
+		if !cliContext.IsSet(f) && !ok {
 			continue
 		}
 
 		// avoid logging sensitive data
-		if f == "user" || f == "password" || f == "ops-user" || f == "ops-password" {
+		if f == "user" || f == "password" || f == "ops-password" {
 			log.Debugf("--%s=<censored>", f)
+			continue
+		}
+
+		if f == "cert" || f == "cert-path" || f == "key" || f == "registry-ca" || f == "tls-ca" {
 			continue
 		}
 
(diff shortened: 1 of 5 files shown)


cmd/vic-machine/create/create.go
```diff
@@ -1285,12 +1289,31 @@ func logArguments(cliContext *cli.Context) {
...
+		bT := cliContext.BoolT(f)
+		if b && !bT {
+			flag := fmt.Sprintf("--%s=%t", f, true)
+			log.Debug(flag)
+			args = append(args, flag)
+			continue
+		}
+
 		match = func() (result bool) {
 			result = false
```
vburenin: I would restructure this function a little bit. I am little suspicion about presence of recover() and usage of result variable. Not sure recover() is really needed here.

```
match = func() boot {
    is := cliContext.IntSlice(f)
    if is != nil {
        ...
    }
    return is != nil
}
```

cmd/vic-machine/create/create.go
```diff
@@ -1302,10 +1325,14 @@ func logArguments(cliContext *cli.Context) {
 		// generic last because it matches everything
 		g := cliContext.Generic(f)
 		if g != nil {
-			log.Debugf("--%s=%#v", f, g)
+			flag := fmt.Sprintf("--%s=%#v", f, g)
+			log.Debug(flag)
+			args = append(args, flag)
 			continue
```
vburenin: presence of continue here is redundant.

cmd/vic-machine/create/create.go
```diff
@@ -1273,10 +1274,29 @@ func (c *Create) generateCertificates(server bool, client bool) ([]byte, *certif
...
+
 	for _, f := range cliContext.FlagNames() {
 		if !cliContext.IsSet(f) {
-			continue
+			skip := true
+			// This is a workaround for cli.Context.IsSet() returning false when
+			// the short option for a cli.StringSlice is supplied instead of the long option
+			// see https://github.com/urfave/cli/issues/314
+			create := reflect.ValueOf(c).Elem()
+			for i := 0; i < create.NumField(); i++ {
```
emlin: The reflection for Create structure can be done out of loop, to generate one map, and use map to check if that flag is set or not.

cmd/vic-machine/create/create.go
```diff
@@ -1273,10 +1274,29 @@ func (c *Create) generateCertificates(server bool, client bool) ([]byte, *certif
...
+						break
+					}
+				}
+			}
+			if skip {
+				continue
+			}
 		}
 
 		// avoid logging sensitive data
```
emlin: user password is not persisted here.
probably should think about encrypting the whole command in guestinfo, so keep sensitive data is not an issue.

cmd/vic-machine/create/create.go
emlin: recover is because of one cli bug as well

cmd/vic-machine/create/create.go
```diff
@@ -1355,10 +1394,13 @@ func logArguments(cliContext *cli.Context) {
 		// generic last because it matches everything
 		g := cliContext.Generic(f)
```
emlin: just wondering why not use Generic for everything?

cmd/vic-machine/create/create.go
```diff
@@ -1292,61 +1312,80 @@ func logArguments(cliContext *cli.Context) {
 				continue
 			}
 			url.User = nil
-			log.Debugf("--target=%s", url.String())
+			flag := fmt.Sprintf("--target=%s", url.String())
+			log.Debug(flag)
+			args = append(args, flag)
```
emlin: could we compare the default value here before actually log out parameter and append command?
I'm asking this is because with our command definition, cli will fit some default value into the input. If we could filter out those, this solution would be perfect.

cmd/vic-machine/inspect/inspect.go
```diff
@@ -130,29 +135,42 @@ func (i *Inspect) Run(clic *cli.Context) (err error) {
...
+		log.Info("")
+		log.Info("Target VCH created with the following options:")
 
-	log.Info("")
-	log.Infof("Installer version: %s", installerVer.ShortVersion())
-	log.Infof("VCH version: %s", vchConfig.Version.ShortVersion())
-	log.Info("")
-	log.Info("VCH upgrade status:")
-	i.upgradeStatusMessage(ctx, vch, installerVer, vchConfig.Version)
+		fmt.Println("")
```
emlin: this is good for console output, but does not write into log file

cmd/vic-machine/inspect/inspect.go
```diff
@@ -130,29 +135,42 @@ func (i *Inspect) Run(clic *cli.Context) (err error) {
...
 
-	if err = executor.InspectVCH(vch, vchConfig); err != nil {
-		executor.CollectDiagnosticLogs()
-		log.Errorf("%s", err)
-		return errors.New("inspect failed")
+		for _, arg := range vchConfig.VicMachineCreateOptions {
+			fmt.Printf("\t%s\n", arg)
+		}
+
+		fmt.Println("")
```
emlin: maybe return here, and then we don't need the else

lib/config/virtual_container_host.go
```diff
@@ -81,6 +81,9 @@ type VirtualContainerHostConfigSpec struct {
 
 	// configuration for vic-machine
 	CreateBridgeNetwork bool `vic:"0.1" scope:"read-only" key:"create_bridge_network"`
+
+	// vic-machine create options used to create or reconfigure the VCH
+	VicMachineCreateOptions []string `vic:"0.1" scope:"read-only" key:"vic_machine_create_options"`
```
emlin: As I mentioned above, you might want to use secret flag for this option, so you can put username/password into it

cmd/vic-machine/create/create.go
jzt: Do we want to persist user/pass? I assumed that was privileged information required ahead of time when doing a create or reconfigure.

cmd/vic-machine/create/create.go
jzt: I'm not sure what you mean as only flags that were explicitly specified by the user will make it into this path, so defaults should not be persisted. A user can supply a default explicitly, and that would be persisted, for example:  `--no-tls-verify=false`. Can you please elaborate in case I'm misunderstanding?

cmd/vic-machine/create/create.go
jzt: One reason that I can think of is that we need to explicitly know when we are dealing with a `StringSlice`, so that we can log/append an individual instance of that option for each element in the slice. For output such as this, for example:
```
--insecure-registry=https://registry1
--insecure-registry=https://registry2
--insecure-registry=https://registry3
```
vs.
```
---insecure-registry=[https://registry1 https://registry2 https://registry3]

File: cmd/vic-machine/inspect/inspect.go (hunk 8 above)
Comment (jzt) [inline cmd/vic-machine/inspect/inspect.go]: Good idea, done!

File: lib/config/virtual_container_host.go (hunk 9 above)
Comment (jzt) [inline lib/config/virtual_container_host.go]: I can do that but are we sure we want to persist the user/pass in the first place? I mean, what benefit does it have if nobody can see it anyway because we obscure it?

File: cmd/vic-machine/inspect/inspect.go (hunk 7 above)
Comment (jzt) [inline cmd/vic-machine/inspect/inspect.go]: Fixed

File: cmd/vic-machine/create/create.go (hunk 6 above)
Comment (jzt) [inline cmd/vic-machine/create/create.go]: Do you mean only in the context of the `target` flag?

Comment (emlin) [inline cmd/vic-machine/create/create.go]: no, it's for all options. For example, https://github.com/vmware/vic/blob/master/cmd/vic-machine/create/create.go#L179, cli will set a default value for this option, if we do not compare the default value, I'm afraid it will persist almost all options

Comment (jzt) [inline cmd/vic-machine/create/create.go]: We only persist options that the user explicitly supplied during `vic-machine create`, so this should not be a problem.

Comment (emlin) [inline cmd/vic-machine/create/create.go]: I see 👍

File: tests/test-cases/Group6-VIC-Machine/6-09-Inspect.md [hunk 10]
```diff
@@ -16,3 +16,6 @@ This test requires that a vSphere server is running and available
 3. Run inspect for VCH1
 4. Using inspect result to run docker command
 5. Verify docker VM is created under correct VCH resource pool or Virtual App through govc
+6. Install VCH
+7. Issue vic-machine inspect --conf command
+8. Verify output contains expected flags & values
```
mhagen-vmware: this step should be in:
#Expected Results

jzt: Fixed

jzt:
@mhagen-vmware we sync'd up, he's waiting on a build. I'll rebase and merge once his build/merge is complete.

Issue #599, opened 2016-04-29
Migrate configuration onto pkg/vsphere/extraconfig


hickeng:
This builds on #229 to move all of the component configuration currently done via CLI flags, environment vars, etc onto the extraconfig/guestinfo model.

The sub-issues are for the individual components.

hickeng:
Adding debt - anything that's not already migrated is definitely debt.

PR #5021, opened 2017-05-04
Syslog [specific ci=6-15-Syslog]

### LINKED ISSUE #4861: syslog support for VCH logs
Add support for sending logs in `/var/log/vic` folder on the appliance to a syslog endpoint.

Acceptance:
* vic-machine option to add syslog endpoint
* log files under `/var/log/vic` are sent to syslog endpoint not located on appliance

hmahmood:
Fixes #4861


cmd/port-layer-server/main.go
```diff
@@ -72,7 +73,31 @@ func main() {
...
 	}
 
-	pprof.StartPprof("portlayer server", pprof.PortlayerPort)
+	logcfg := viclog.NewLoggingConfig()
+	if ploptions.PortLayerOptions.Debug {
+		logcfg.Level = log.DebugLevel
+	}
+
+	if ploptions.PortLayerOptions.SyslogAddr != nil {
+		u, err := url.Parse(*ploptions.PortLayerOptions.SyslogAddr)
```
emlin: is it better to use URL format for SyslogAddr? then you don't need to parse it here

hmahmood: Yeah that might be better.

cmd/vic-machine/create/create.go
```diff
@@ -447,6 +447,13 @@ func (c *Create) Flags() []cli.Flag {
 			Destination: &c.httpProxy,
 			Hidden:      true,
 		},
+
+		cli.StringFlag{
+			Name:        "syslog-address",
+			Value:       "",
+			Usage:       "Address of the syslog server to send Virtual Container Host logs to. Must be in the format transport://host[:port], where transport is either tcp or udp. transport is either udp, tcp, udp6, or tcp6. port defaults to 514 if not specified",
```
emlin: is the transport duplicated mentioned?

mdubya66:
How are we going to test this?

lib/install/data/data.go
```diff
@@ -84,6 +84,12 @@ type Data struct {
 
 	ScratchSize string
 	Rollback    bool
+
+	SyslogConfig SyslogConfig
```
emlin: SyslogConfig SyslogConfig => SyslogConfig

lib/install/validate/validator.go
```diff
@@ -37,13 +37,16 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/cmd/vic-machine/common"
 	"github.com/vmware/vic/lib/config"
+	"github.com/vmware/vic/lib/config/executor"
```
emlin: this import is not used

lib/install/validate/validator.go
```diff
@@ -658,3 +663,48 @@ func (v *Validator) AddDeprecatedFields(ctx context.Context, conf *config.Virtua
 
 	return &dconfig
 }
+
+func (v *Validator) syslog(conf *config.VirtualContainerHostConfigSpec, input *data.Data) {
```
emlin: feels like this is just url parsing for input parameter, could you move it to cmd package? 
I think validator is used for vsphere related object validation.

emlin: at least, we could move the url parsing to cmd package, and leave the schema, port detection here.

pkg/log/syslog/syslog.go
```diff
@@ -0,0 +1,248 @@
...
+	// for testing purposes
+	d dialer
+}
+
+func (d *defaultDialer) dial(cfg *SyslogConfig) (Writer, error) {
+	w, err := d.d.dial(cfg)
+	if err != nil {
+		switch err.(type) {
+		case *net.AddrError, *net.ParseError:
+			return nil, err
```
emlin: what happens if err is not address or parse error?

emlin: Got that you'll redial if this failed here, but we might need message out if dail failed eventually

cmd/docker/main.go
```diff
@@ -130,16 +136,31 @@ func handleFlags() bool {
...
 
+func initLogging() error {
+	logcfg := viclog.NewLoggingConfig()
+	if *cli.debug || vchConfig.Diagnostics.DebugLevel > 0 {
+		logcfg.Level = log.DebugLevel
+	}
+
+	if vchConfig.Diagnostics.SysLogConfig != nil {
+		logcfg.Syslog = &syslog.SyslogConfig{
+			Network:   vchConfig.Diagnostics.SysLogConfig.Network,
```
emlin: why using guestinfo configuration to initialize logging in persona but not in portlayer?

lib/install/data/data.go
hmahmood: That would embed the struct; I wanted it to be separate.

lib/install/validate/validator.go
hmahmood: Will do

cmd/docker/main.go
hmahmood: I was just following the pattern already in the code. port layer seems to use env variables for configuration, docker persona is using guestinfo. Actually port layer seems to be the outlier here.

cmd/vic-machine/create/create.go
hmahmood: Yes, yes it is. I'll fix that.

hmahmood:
@mdubya66 there are unit tests that cover almost all of the syslog package code that I added. As for integration/functional testing, that depends. If we want to verify that we successfully connect to a syslog server, that is relatively easy. If we want to verify that log messages in the log files are the same as the log messages received by the syslog server, that would be more challenging, since we would need to get the VCH logs and compare. Regardless, this would not end up in the CI since running a syslog server is not straightforward as I found out.

cmd/port-layer-server/main.go
hmahmood: Actually, I did try that first. The go-flags package does not support `url.URL`, so had to revert back to `string`.

pkg/log/syslog/syslog.go
hmahmood: You mean if dial succeeds eventually?

mdubya66:
We need tests. Work with @mhagen-vmware on what solution works to satisfy the requirements

lib/install/validate/validator.go
```diff
@@ -658,3 +663,48 @@ func (v *Validator) AddDeprecatedFields(ctx context.Context, conf *config.Virtua
...
+	}
+
+	network := u.Scheme
+	if len(network) == 0 {
+		v.NoteIssue(errors.New("syslog address does not have network specified"))
+		return
+	}
+
+	switch network {
+	case "udp":
```
anchal-agrawal: Nitpick - these cases can be simplified to

```go
case "udp", "tcp", "tcp6", "udp6":
```

pkg/log/syslog/syslog.go
emlin: I mean failed after retry, there is no message out for what reason, e.g. the connection has problem

cmd/port-layer-server/main.go
emlin: gotcha

lib/install/data/data.go
emlin: sure

emlin:
btw, is wondering if syslog address is configured, do we need to have vic-machine integrate with that as well?

pkg/log/syslog/syslog.go
hmahmood: We can't put out a log message as that would cause an infinite loop since the log call is going to end up in this code, and this code would call log again ...

lib/install/data/data.go
jzt: Just FYI, you can still refer to it by the same name if you embed it, i.e.: `if conf.Diagnostics.SyslogConfig != nil { ... }`

cmd/docker/main.go
hickeng: We never completed #599 so there's still some misalignment. We also need to consider how read-only configuration such as this could and should be exposed for regular containers in order to be consistent with the vic-machine/pl work.

lib/config/executor/container_vm.go
```diff
@@ -81,6 +81,18 @@ type Diagnostics struct {
...
 	// restartable entities
 	ExitLogs []ExitLog `vic:"0.1" scope:"read-write" key:"exitlogs"`
+
+	// SyslogConfig holds configuration for connecting to a syslog
+	// server
+	SysLogConfig *SysLogConfig `vic:"0.1" scope:"read-only" key:"syslog"`
+}
+
+// SyslogConfig holds the configuration necessary to connect to a syslog server
+type SysLogConfig struct {
```
hickeng: Is it viable to have a different log level for rsyslog than for the local logs? Is it desirable? I ask because we pump out a lot of logging at debug level and I'm not sure people will want to have their syslog servers flooded.

It could be useful if there's a loginsight analytics package for the debug level output. Also of note is the fact that if the remote goes down we will retry dialing for every single log message.

pkg/log/syslog/syslog.go
```diff
@@ -0,0 +1,268 @@
...
+
+type writerWrapper struct {
+	writer Writer
+	cfg    SyslogConfig
+	dialer dialer
+}
+
+func (w *writerWrapper) withRetry(f func()) {
+	if w.writer == nil {
+		w.writer, _ = w.dialer.dial(&w.cfg)
```
hickeng: This might need to use the inflight package to block multiple attempts at dailing - we definitely call logging concurrently. See https://github.com/vmware/vic/blob/master/lib/portlayer/attach/communication/connector.go#L33 for example.

pkg/log/syslog/syslog.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	return w.writer.Close()
+}
+
+const maxTagLen = 32
+
+// MakeTag makes an RFC 3164 compliant tag (32 characters or less)
+// using the provided proc. If proc is empty, the name of the current
+// executable is used, trauncated to maxTagLen characters if
```
hickeng: truncated

pkg/log/syslog/syslog.go
```diff
@@ -0,0 +1,268 @@
...
+// necessary.
+func MakeTag(proc string) string {
+	if len(proc) == 0 {
+		proc = filepath.Base(os.Args[0])
+	}
+	proc = strings.TrimSpace(proc)
+	if len(proc) >= maxTagLen {
```
hickeng: Why `>=`? Unless the tag is nil terminated it would seem to want to slicing the last character. Conversely the slice below doesn't slice the last character so has no effect in the `==` case.

cmd/docker/main.go
```diff
@@ -130,16 +136,30 @@ func handleFlags() bool {
...
+			RAddr:     vchConfig.Diagnostics.SysLogConfig.RAddr,
+			Formatter: syslog.RFC3164,
+			Priority:  srslog.LOG_INFO | srslog.LOG_DAEMON,
+		}
+	}
+
+	return viclog.Init(logcfg)
```
hmahmood: We can put out log messages here at the beginning here, but once we are inside the syslog hook, we can't produce any more log messages since they end up again in the syslog loop, thus resulting in a positive feedback loop that will run away exponentially.

The syslog library code will keep retrying if the remote goes down or is down to begin with. We redial if we never have a `srslog.Writer` to begin with -- the case where the remote is down at the start. Once we have a `srslog.Writer` object, it will retry on its own, and we don't have to redial.

lib/config/executor/container_vm.go
hmahmood: There is certainly a way to configure logrus hooks so that they are only called for certain log levels. How we expose that to the user is the difficult part here.

pkg/log/log.go
```diff
@@ -0,0 +1,78 @@
...
+					logrus.SetFormatter(f)
+					logrus.SetLevel(l)
+				}
+			}()
+
+			logrus.SetFormatter(cfg.Formatter)
+			logrus.SetLevel(cfg.Level)
+
+			if cfg.Syslog != nil {
+				cfg.Syslog.Tag = syslog.MakeTag(cfg.Syslog.Tag)
```
hmahmood: I did play with that a little bit, but then we have to decide what to truncate -- the process name or the vch name, since tag is limited to 32 characters by syslog. I went with just the proc name since syslog messages do have hostnames in them, and that can be used to distinguish VCH's. However, if the ip address changes then of course this becomes a little bit untenable.

pkg/log/syslog/syslog.go
hmahmood: Yes that is a left over from old code; I'll remove it.

pkg/log/syslog/syslog.go
hmahmood: We do call logging concurrently, however logrus does serialize all calls by using a lock, so we're fine here.

However, as a standalone writer implementation, yes, we do need something to serialize the calls here.

hmahmood: Actually I should correct that. logrus *does not* grab a lock before calling the hooks. It does, however, grab a lock when calling `Write` on the `io.Writer` it uses. The hook implementation is ok since the `Fire` calls just write to a channel. `writerWrapper` is however not thread-safe. This is OK for now, since it is only called by the hook. For future uses we can document that this is not thread-safe, and have any caller protect calls made to it.

cmd/docker/main.go
hickeng: @hmahmood does the hook block the log call from returning or is it async. If the former, can we push the log messages to a queue for dispatch to the syslog - I'd rather not pin our performance to that of the network & syslogd system.
Also - default TCP timeout is 10min in the case of no response, which is way too long for this scenario - I  don't know whether Go applies a different timeout value.

hmahmood: We are pushing it onto a queue; the hook `Fire` call is not async, which is why I just put the entry into a buffered channel for precisely the reason you point out.

The timeout is OK, we just drop logs if that happens; no blocking happens of the actual log call. I don't want to make this very complicated -- for now no connection, we drop stuff is enough.

.drone.yml
```diff
@@ -82,6 +82,9 @@ pipeline:
       GS_CLIENT_EMAIL: ${GS_CLIENT_EMAIL}
       GS_PRIVATE_KEY: ${GS_PRIVATE_KEY}
       DOMAIN: ${CI_DOMAIN}
+      SYSLOG_SERVER: ${SYSLOG_SERVER}
```
jakedsouza: You can use `service containers`  with drone. http://readme.drone.io/0.5/usage/services/ 

In this case you could deploy a syslog service container for running tests.

hmahmood:
Need @hickeng and @mdubya66 to review please. Thanks.

pkg/log/syslog/mock_Writer_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+import mock "github.com/stretchr/testify/mock"
+
+// MockWriter is an autogenerated mock type for the Writer type
+type MockWriter struct {
+	mock.Mock
+}
+
+// Close provides a mock function with given fields:
+func (_m *MockWriter) Close() error {
```
fdawg4l: Why not `m`? Basically why the underscore?

pkg/log/syslog/mock_Writer_test.go
```diff
@@ -0,0 +1,129 @@
...
+	if rf, ok := ret.Get(1).(func([]byte) error); ok {
+		r1 = rf(p)
+	} else {
+		r1 = ret.Error(1)
+	}
+
+	return r0, r1
+}
+
+var _ Writer = (*MockWriter)(nil)
```
fdawg4l: What is this about? I don't undetectable why we're casting nil.

hmahmood: This is auto-generated code.

PR #5023, opened 2017-05-04
[specific ci=1-06-Docker-Run] Allow for up to 3 seconds worth of NTP drift

### LINKED ISSUE #4984: CI integration test failure: docker run verify start/stop time
**User Statement:**

As a contributor to VIC, I should have integration tests that don't fail due to clock skew between the host machine and the ESX/vCenter setup.

**Details:**

From build https://ci.vmware.run/vmware/vic/10211:

```
Test-Cases.Group1-Docker-Commands.1-06-Docker-Run :: Test 1-06 - Docker Run   
Docker run verify container start and stop time                       | FAIL |
container start time before command start
```

This looks like a follow-up of #3710. The test compares timestamps between the container running the tests and the VCH, which can be unreliable.

The best fix would be to issue a `docker info` call at the start of the test, extract the system time from the output and use it a reference timestamp instead of `date +%s` on the build system. With this approach, the changes made in #4815 should not be required.

**Acceptance Criteria:**

The above test passes reliably.

mhagen-vmware:
fixes #4984

### DIFF
--- a/tests/test-cases/Group1-Docker-Commands/1-06-Docker-Run.robot
+++ b/tests/test-cases/Group1-Docker-Commands/1-06-Docker-Run.robot
@@ -113,7 +113,7 @@ Docker run verify container start and stop time
     Should Be Equal As Integers  ${rc}  0
     Should Not Contain  ${output}  Error
     ${cmdStart}=  Run  date +%s
-    Sleep  1
+    Sleep  3
     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run --name startStop busybox
     Should Be Equal As Integers  ${rc}  0
     Should Be Empty  ${output}


cgtexmex:
Help me to understand how the sleep will help with NTP?  Also why not utilize `docker info` like @anchal-agrawal suggested?

Finally not a fan of having the ci specific stuff in the commit message...in the PR sure..but does it need to be in the commit?  IMO impacts the `git log --oneline` aesthetics.

mhagen-vmware:
not going to end up in the commit message, if I remember to remove it, which I usually do (I think).

I do not like the docker info approach as it doesn't tell us much if anything, the VCH could be reporting completely wrong times and we would not know at all, it also does not eliminate the need for #4815 as he states. Just because you use docker info, does not obviate the need to test for relative accuracy later. We could use docker info to get the VCH time, it is off by a crap ton, we pass the command doesn't start before it should portion of the test but the VCH is reporting a second time that is days in the future.

By adding a 3 second sleep, it allows for up to a 3 second drift backwards in NTP time between the two systems. Because we get the client time, sleep 3 seconds, then get the VCH time.

anchal-agrawal:
While this change alleviates https://github.com/vmware/vic/issues/4984, I don't think this PR should close the issue.

mhagen-vmware:
why?  you think the systems are going to drift more than 3 seconds? or is there some other failure you are concerned about?

anchal-agrawal:
(Posted the above response when @mhagen-vmware's comment was not visible)

My understanding was that the main goal of that test was to verify that the container's stop time is after the start time in the `docker inspect` output. Using the timestamp from`docker info` as reference would be enough in that case. If we also intend to test cross-system timestamps, then I concur with @mhagen-vmware's approach. Clock skew is my only concern.

mhagen-vmware:
And yes, apparently I suck cause I just did it earlier today :(

https://github.com/vmware/vic/commit/59e861489ade203778f88879a6dfc89918eaa470

I will try to be better...

cgtexmex:
LGTM

mhagen-vmware:
ok, apparently I really suck:
https://github.com/vmware/vic/commit/5f3172c6acf4103cffcbe53a6a33a6c2172a27a1
https://github.com/vmware/vic/commit/12a3bbedec209b31b752644ba5b8794b586a8b68
https://github.com/vmware/vic/commit/2e51c2959ec277317c3fc375fe3e756769c140cc
https://github.com/vmware/vic/commit/3637009885d6caf97cddd51d42996ced0513de95

in my defense, quarantining a crap ton of tests yesterday didn't put me in a great mood, but that is a horrific list of failures on my part :crying_cat_face: :sob: :cry:

Issue #4641, opened 2017-04-07
vic-machine: reconfigure VCH (1.2)


emlin:

mdubya66:
Should we add the manifest file to the list. At least the start of one. Thoughts?

emlin:
move #4617 to this epic as the start of manifest file support

karthik-narayan:
Emma and I just spoke. There are 4 items that I do not see in here
1. Updating the proxy settings 
2. Updating the DNS 
3. Updating VCH memory / CPU 
4. Updating the endpoint VM memory / CPU

The first two are important and we will need to pull them into the 1.2 release planning. 

The other two are also important, but we do have a UI workaround. Our general philosophy has been to ask customers to only use the vic-machine CLI to make changes and to not use the vSphere UI. We will need to be caveat that this time around. All said, if we have the bandwidth, I'd like us to pull these in as well.

mdubya66:
@karthik-narayan can we make these opportunistic if we work through this list?

PR #5046, opened 2017-05-08
[specific ci=Group6-VIC-Machine]Add vic-machine configure command

### LINKED ISSUE #4619: vic-machine reconfigure command

emlin:
Use same process with upgrade.

Currently this command is still hidden, for two reasons:
- No any reconfigure option added (issues listed in #4641 )
- Convert data from guestinfo back to command option or create data structure is not done yet ( #4618 ).

Fixes #4619


cmd/vic-machine/configure/configure.go
```diff
@@ -0,0 +1,248 @@
...
+
+	return configure
+}
+
+// Flags return all cli flags for upgrade
+func (c *Configure) Flags() []cli.Flag {
+	util := []cli.Flag{
+		cli.BoolFlag{
+			Name:        "force, f",
+			Usage:       "Force the configure",
```
anchal-agrawal: `Force the configure operation` sounds better.

cmd/vic-machine/configure/configure.go
```diff
@@ -0,0 +1,248 @@
...
+	if !c.Data.Rollback {
+		err = executor.Configure(vch, vchConfig, vConfig)
+	} else {
+		err = executor.Rollback(vch, vchConfig, vConfig)
+	}
+
+	if err != nil {
+		// upgrade failed
+		executor.CollectDiagnosticLogs()
+		if err == nil {
```
anchal-agrawal: Would `err == nil` be possible here?

lib/install/data/data_test.go
```diff
@@ -0,0 +1,36 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
```
anchal-agrawal: 2017

lib/install/data/data.go
```diff
@@ -134,12 +149,35 @@ type InstallerData struct {
...
+			MappedNetworksIPRanges: make(map[string][]ip.Range),
+			MappedNetworksDNS:      make(map[string][]net.IP),
+		},
+		Timeout: 3 * time.Minute,
 	}
 	return d
 }
+
+// CopyNonEmpty will shallow copy src value to override existing value if the value is set
+// This copy will take care of relationship between variables, that means if any variable in ContainerNetwork or NetworkConfig is not empty, the whole ContainerNetwork or NetworkConfig will be copied
```
anchal-agrawal: It'd be nice to break this line into two for readability.

cmd/vic-machine/configure/configure.go
```diff
@@ -0,0 +1,245 @@
...
+			Destination: &c.Timeout,
+		},
+		cli.BoolFlag{
+			Name:        "resetInProgressFlag",
+			Usage:       "Reset the UpdateInProgress flag. Warning: Do not reset this flag if another upgrade/configure process is running",
+			Destination: &c.ResetInProgressFlag,
+		},
+		cli.BoolFlag{
+			Name:        "rollback",
+			Usage:       "Roll back VCH configuration to before the previous upgrade/configure",
```
andrewtchin: Should this be "to before the current upgrade/configure"?

lib/install/management/configure.go
```diff
@@ -240,7 +254,11 @@ func (d *Dispatcher) update(conf *config.VirtualContainerHostConfigSpec, setting
 		}
 	}
 
-	if err = d.reconfigVCH(conf, fmt.Sprintf("[%s] %s/%s", conf.ImageStores[0].Host, d.vmPathName, settings.ApplianceISO)); err != nil {
+	isoFile := settings.ApplianceISO
```
anchal-agrawal: Here, if `settings.ApplianceISO == ""`, `isoFile = ""` as well.

```go
        isoFile := ""
 	if settings.ApplianceISO != "" {
 		isoFile = fmt.Sprintf("[%s] %s/%s", conf.ImageStores[0].Host, d.vmPathName, settings.ApplianceISO)
 	}
```

The above snippet is easier to understand and has the same behavior as the original code.

lib/install/data/data_test.go
```diff
@@ -0,0 +1,36 @@
...
+package data
+
+import (
+	"net"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCopy(t *testing.T) {
```
anchal-agrawal: We should rename this func to `TestCopyNonEmpty`. Also, can we add cases to check that empty fields are not copied?

cmd/vic-machine/configure/configure.go
```diff
@@ -0,0 +1,245 @@
...
+			Destination: &c.Force,
+		},
+		cli.DurationFlag{
+			Name:        "timeout",
+			Value:       3 * time.Minute,
+			Usage:       "Time to wait for configure",
+			Destination: &c.Timeout,
+		},
+		cli.BoolFlag{
+			Name:        "resetInProgressFlag",
```
andrewtchin: Can this be changed to `reset-progress` or something since none of the other flags are camel case?

emlin: hmm.. that is used in upgrade command. Anyway, let me change both of them

cmd/vic-machine/configure/configure.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	if err != nil {
+		log.Error("Failed to get Virtual Container Host configuration")
+		log.Error(err)
+		return errors.New("configure failed")
+	}
+
+	// TODO: Convert guestinfo *VirtualContainerHost back to *Data, decrypt secret data
+	oldData := &data.Data{}
+	// oldData, err := converter.GtoD(vchConfig)
```
anchal-agrawal: Unneeded comment?

emlin: This is the comment intentionally left here. Those code is not there yet, but I want to have them added here.

anchal-agrawal:
@emlin, can we add a `log.Infof("")` on line 209 in [this file](https://github.com/vmware/vic/blob/master/lib/install/management/inspect.go#L209) so we have a blank INFO line between the IP and the `Unable to find valid client certs` message?

picture

This request came from George during the sprint demo and it'd be nice to make this change here since it's related and has zero impact.

PR #5047, opened 2017-05-08
Increased event collector page size to avoid missing events [specific ci=1-10-Docker-PS]

### LINKED ISSUE #4997: docker ps does not pick up shut down container
from:
https://ci.vmware.run/vmware/vic/10236

After about 30 seconds the docker create ls container still had not exited. From my investigation though, it looks like it ran to completion within a reasonable amount of time:
/brave_goldwasser-af72af66d7f6/vmware.log
```
2017-05-03T04:07:44.893Z| vmx| I125: Log for VMware ESX pid=1982267 version=6.0.0 build=build-5224934 option=Release
```
snip
```
2017-05-03T04:07:50.087Z| vmx| W115: VMX has left the building: 0.
```
Less than 6 seconds, so seemingly our follow up docker ps somehow didn't pick that up for over 20 seconds after that.

[integration_logs_10236_59d19f32a0bfddd194852d3a292ed70d74f69d49.zip](https://github.com/vmware/vic/files/973432/integration_logs_10236_59d19f32a0bfddd194852d3a292ed70d74f69d49.zip)

cgtexmex:
Additional logging also added to help identify events that are
ignored at the collector or in the container event callback.

Fixes #4997

### DIFF
--- a/lib/portlayer/event/collector/vsphere/collector.go
+++ b/lib/portlayer/event/collector/vsphere/collector.go
@@ -117,8 +117,11 @@ func (ec *EventCollector) Start() error {
 	// we don't want the event listener to timeout
 	ctx := context.Background()
 
-	// events per page
-	pageSize := int32(1)
+	// pageSize is the number of events on the last page of the eventCollector
+	// as new events are added the oldest are removed.  Originally this value
+	// was 1 and we encountered missed events due to them being evicted
+	// before being processed.  A setting of 25 should provide ample buffer.
+	pageSize := int32(25)
 	// bool to follow the stream
 	followStream := true
 	// don't exceed the govmomi object limit
@@ -152,12 +155,11 @@ func (ec *EventCollector) Start() error {
 // may provide more flexibility
 func evented(ec *EventCollector, page []types.BaseEvent) {
 	if ec.callback == nil {
+		log.Warn("No callback defined for EventManager")
 		return
 	}
 
 	for i := range page {
-		var event events.Event
-
 		// what type of event do we have
 		switch page[i].(type) {
 		case *types.VmGuestShutdownEvent,
@@ -169,11 +171,16 @@ func evented(ec *EventCollector, page []types.BaseEvent) {
 			*types.VmMigratedEvent,
 			*types.DrsVmMigratedEvent,
 			*types.VmRelocatedEvent:
-			event = NewVMEvent(page[i])
-		}
 
-		if event != nil {
-			ec.callback(event)
+			// we have an event we need to process
+			ec.callback(NewVMEvent(page[i]))
+		case *types.VmReconfiguredEvent:
+			// reconfigures happen often, so completely ignore for now
+			return
+		default:
+			// log the skipped event
+			e := page[i].GetEvent()
+			log.Debugf("vSphere Event %s for eventID(%d) ignored by the event collector", e.FullFormattedMessage, int(e.Key))
 		}
 	}
 
--- a/lib/portlayer/exec/exec_test.go
+++ b/lib/portlayer/exec/exec_test.go
@@ -15,7 +15,6 @@
 package exec
 
 import (
-	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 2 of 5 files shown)


cgtexmex:
I've looked at about 5 failures of this test.  In all of the cases the first containerVM that was created / started never was transitioned to a stopped state based on the poweredOff event.  The reason it never transitioned was the portLayer never receives the event.  

I think the reason we are missing the event is because the vSphere `EventHistoryCollector` is created with a [page size of 1](https://github.com/vmware/vic/blob/master/lib/portlayer/event/collector/vsphere/collector.go#L121).  This means that the collector will keep the most recent event on the last page of the event collector, but also means that if a new event comes it will evict the oldest event -- which in this case is the only event.  This causes an issue when you have a spike of events and their is latency on the system -- specifically we miss events.    

To remedy this I've increased the page size to 25.  I've run roughly 10 test jobs with this setting and have not seen the issue occur again.  One concern I had with a page size greater than 1 would be the possibility of repeated events -- I have not seen that in the testing with increased page size.  I'll run this thru the CI some more (specifically when there is latency) for additional testing.

@dougm please take a look at this approach and see if it looks reasonable.

dougm:
@cgtexmex page size of 25 should be fine

lib/portlayer/event/collector/vsphere/collector.go
```diff
@@ -169,12 +169,18 @@ func evented(ec *EventCollector, page []types.BaseEvent) {
...
+			ec.callback(NewVMEvent(page[i]))
+		case *types.VmReconfiguredEvent:
+			// reconfigures happen often, so completely ignore for now
+			return
+		default:
+			// log the skipped event
+			e := page[i].GetEvent()
+			log.Debugf("vSphere Event %s for eventID(%d) ignored by the event collector", e.FullFormattedMessage, int(e.Key))
 		}
+
```
dougm: RNL

lib/portlayer/exec/exec_test.go
```diff
@@ -29,32 +29,32 @@ import (
...
-	assert.EqualValues(t, StateSuspended, eventedState(event, StateSuspended))
-	assert.EqualValues(t, StateSuspended, eventedState(event, StateRunning))
-
-	// removed event
-	event = events.ContainerRemoved
-	assert.EqualValues(t, StateRemoved, eventedState(event, StateRunning))
-	assert.EqualValues(t, StateRemoved, eventedState(event, StateStopped))
-	assert.EqualValues(t, StateRemoving, eventedState(event, StateRemoving))
-}
+// func TestEventedState(t *testing.T) {
```
dougm: Why is this commented out?  If it shouldn't just be removed, then use t.Skip("Reason") instead of commenting out.

cgtexmex: Thanks for catching..will update

lib/portlayer/event/collector/vsphere/collector.go
```diff
@@ -169,12 +169,18 @@ func evented(ec *EventCollector, page []types.BaseEvent) {
...
+			// we have an event we need to process
+			ec.callback(NewVMEvent(page[i]))
+		case *types.VmReconfiguredEvent:
+			// reconfigures happen often, so completely ignore for now
+			return
+		default:
+			// log the skipped event
+			e := page[i].GetEvent()
+			log.Debugf("vSphere Event %s for eventID(%d) ignored by the event collector", e.FullFormattedMessage, int(e.Key))
 		}
```
hickeng: if you're concerned about duplicate events, should we store the EventID (assuming that's unique) or other identifying data just for the last page processed and check against that as a filter?.

lib/portlayer/event/collector/vsphere/collector.go
```diff
@@ -118,7 +118,7 @@ func (ec *EventCollector) Start() error {
 	ctx := context.Background()
 
 	// events per page
-	pageSize := int32(1)
+	pageSize := int32(25)
```
hickeng: suggest more comment about why this page size - git blame could identify this issue, but there's ways that can get mixed up.

lib/portlayer/event/collector/vsphere/collector.go
cgtexmex: I did that in testing and saw no dupes or out of order events, so removed from this code

If we hit an issue due to that then the other logging should provide enough detail...meaning the ability to follow an event from entry to exit

PR #5072, opened 2017-05-09
Fix Drone Env Variable

maplain:
Use $ to reference env variable instead of $$

It's uploading something like: 
  vic_10BUILD_NUMBER.tar.gz
  vic_9BUILD_NUMBER.tar.gz


.drone.nightly.yml
```diff
@@ -4,37 +4,37 @@ clone:
 
 build:
   integration-test:
-    image: $${TEST_BUILD_IMAGE=gcr.io/eminent-nation-87317/vic-integration-test:1.30}
+    image: ${TEST_BUILD_IMAGE=gcr.io/eminent-nation-87317/vic-integration-test:1.30}
```
jakedsouza: I dont think we are updating the nightly builds to 05 yet. @mhagen-vmware can provide details

.drone.ova.yml
```diff
@@ -15,7 +15,7 @@ clone:
 
 build:
   unified-ova:
-    image: $${TEST_BUILD_IMAGE=gcr.io/eminent-nation-87317/vic-integration-test:1.30}
+    image: ${TEST_BUILD_IMAGE=gcr.io/eminent-nation-87317/vic-integration-test:1.30}
```
jakedsouza: same here, no change yet to drone 05

.drone.nightly.yml
mhagen-vmware: we were not intending too... does it not work anymore if we don't upgrade nightly?

.drone.ova.yml
mhagen-vmware: correct

.drone.yml
```diff
@@ -176,7 +176,6 @@ pipeline:
   report-coverage:
     image: plugins/coverage
     server: https://coverage.vmware.run
-    token: ${GITHUB_AUTOMATION_API_KEY}
```
mhagen-vmware: I think we need this to avoid hitting the non-token limit, which is very strict?

maplain: @jakedsouza

jakedsouza: The plugin accepts a secret env var `GITHUB_TOKEN` from drone, it is still being passed in, not via the yml but via drone .
As per their docs https://github.com/drone-plugins/drone-coverage

jakedsouza: I pushed the corresponding secret to drone, will push to our internal repo once this PR is in

PR #5079, opened 2017-05-10
Adding version number in more visible places

stuclem:
Following an internal customer using the 0.8 version of the docs with version 1.1 of the product, attempting to make the version number more apparent. In the HTML output, this puts "Version 0.8" at the top of the left-hand TOC and in the title text that appears  if you move your mouse to the top of the page. This is as much as the older version of Gitbook that we use for the older docs will allow us to do. Things are better in version 1.1, which uses a later version of Gitbook to generate output.


stuclem:
♻️

PR #5087, opened 2017-05-10
Add join of path for mount and vs endpoint for nfs volume target

### LINKED ISSUE #5065: nfs device backing returns the incorrect DiskPath in container for nfs mount
**VIC version:**

1.1.0 / master / commit ID

**Steps to reproduce:**

create an nfs volume and attach it to a container. touch a file. 
then make a new volume separate from that volume. and attach that volume to a container, note that the touched file from the other volume is visible and writable by the container mounting a different volume. 

**Actual behavior:**
can write to the volumestore target point rather than the volume mount point in the volume store.
**Expected behavior:**
mount the appropriate volume endpoint. 

This is definitely caused by the target returned from the `DiskPath` function for the nfs device backing struct. once that is fixed the proper volume mount should occur.

matthewavery:
Fixes #5065 

The diskpath for the nfs device path that was conveyed to the tether had an incorrect path in it. leading the nfs volume to mount at `<vs store path>` rather than `<vs store path>/<volume target path>`. This led to nfs volume data all being stored on the vs store target path which was completely incorrect. you could see data from other volumes and the volume data was not namespaced properly.

### DIFF
--- a/lib/portlayer/storage/nfs/disk.go
+++ b/lib/portlayer/storage/nfs/disk.go
@@ -30,14 +30,14 @@ type Volume struct {
 }
 
 func NewVolume(host *url.URL, NFSPath string) Volume {
-	diskLocation := &url.URL{
+	volumeLocation := &url.URL{
 		Scheme: host.Scheme,
 		Host:   host.Host,
 		Path:   path.Join(host.Path, NFSPath),
 	}
 
 	v := Volume{
-		Host: diskLocation,
+		Host: volumeLocation,
 		Path: NFSPath,
 	}
 	return v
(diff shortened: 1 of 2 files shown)


lib/portlayer/storage/nfs/disk.go
```diff
@@ -16,21 +16,28 @@ package nfs
...
+	// VS Host + Path to the actual volume
 	Host *url.URL
 
-	// Path on the Host where the volume is located
+	// Path of the volume from the volumestore target
 	Path string
 }
 
 func NewVolume(host *url.URL, NFSPath string) Volume {
+	diskLocation := &url.URL{
```
fdawg4l: Minor.  We're naming challenged.  However, lets not conflate disks with nfs targets.  This isn't a disk.

matthewavery:
@mhagen-vmware I will need a verify on this, accidently merged before verify. Please let me know if there are changes you would like and I will make them immediately and tag them in this PR and the adjoining.

PR #5095, opened 2017-05-11
[specific ci=1-04-Docker-Create] Fix test for docker create with no CMD

anchal-agrawal:
The public alpine image was recently updated to include a default CMD, so this change replaces it with centos:6.6 in the test.


rajanashok:
lgtm

Issue #5082, opened 2017-05-10
vic-machine: convert guestinfo back to create data structure


emlin:
As a vic-machine developer, I need a library to convert VCH configuration from guestinfo, back to vic-machine create data structure, so that I can make additional changes based on vic-machine configure command input, and then reuse same data validation process, to make sure all configurations are valid while vic-machine configure is executed.

Details:
The current design to use reflection, and flag tag in guestinfo data structure, and convert data.

Acceptance Criteria:
- library is implemented
- unit test

PR #5098, opened 2017-05-11
[specific ci=Group11-Upgrade]encrpyt/decrypt configuration in reconfigure

emlin:
#5082


lib/install/management/appliance.go
```diff
@@ -639,21 +639,51 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
...
+		if err != nil {
+			err = errors.Errorf("Failure finding image store from VCH VM %q: %s", name, err.Error())
+			return nil, err
+		}
+		path, err := vm.FolderName(d.ctx)
+		if err != nil {
+			err = errors.Errorf("Failed to get VM %q datastore path: %s", name, err.Error())
+			return nil, err
+		}
+		log.Debug("getting secret key")
```
anchal-agrawal: We may not need this log since `GuestInfoSecret` has a trace.

lib/install/management/inspect.go
```diff
@@ -206,6 +206,7 @@ func (d *Dispatcher) GetDockerAPICommand(conf *config.VirtualContainerHostConfig
 					dEnv = append(dEnv, fmt.Sprintf("DOCKER_CERT_PATH=%s", abs))
 				}
 			} else {
+				log.Infof("")
```
sgairo: Is this cosmetic?

anchal-agrawal: This is from https://github.com/vmware/vic/pull/5046#issuecomment-300278215.

sgairo: I love anything that makes reading logs easier

lib/install/management/appliance.go
```diff
@@ -639,21 +639,49 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
...
+	if d.secret == nil {
+		name, err := vm.Name(d.ctx)
+		if err != nil {
+			err = errors.Errorf("Failed to get vm name %q: %s", vm.Reference(), err)
+			return nil, err
+		}
+		// set session datastore to where the VM is running
+		ds, err := d.getImageDatastore(vm, nil, true)
+		if err != nil {
+			err = errors.Errorf("Failure finding image store from VCH VM %q: %s", name, err.Error())
```
vburenin: no need to call err.Error(), %s does that automatically. Also, why would not you use fmt.Errorf as we use it everywhere as of now.

lib/install/management/appliance.go
```diff
@@ -639,21 +639,49 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
...
 		}
-
 		d.secret = s
 	}
 
-	cfg := make(map[string]string)
-	extraconfig.Encode(d.secret.Sink(extraconfig.MapSink(cfg)), conf)
-
-	return cfg, nil
+	var conf config.VirtualContainerHostConfigSpec
```
vburenin: for clarity purposes, I would rewrite it as:
```
conf := &config.VirtualContainerHostConfigSpec{}
extraconfig.Decode(d.secret.Source(extraconfig.MapSource(cfg)), &conf)
return conf, nil
```

Technically there is not difference, but I would prefer to do operation & only once instead of twice.

lib/install/management/configure.go
```diff
@@ -431,16 +433,15 @@ func extractSecretFromFile(rc io.ReadCloser) (string, error) {
...
 
-// GuestInfoSecret downloads the VCH's .vmx file and returns the GuestInfoSecretKey
-// value. This function expects the datastore in the dispatcher's session to be set.
-func (d *Dispatcher) GuestInfoSecret(vchName string) (*extraconfig.SecretKey, error) {
+// GuestInfoSecret downloads the VCH's .vmx file and returns the GuestInfoSecretKey value.
+func (d *Dispatcher) GuestInfoSecret(vchName, vmPath string, ds *object.Datastore) (*extraconfig.SecretKey, error) {
 	defer trace.End(trace.Begin(""))
 
-	if d.session.Datastore == nil {
+	if ds == nil {
```
vburenin: How can it happen that ds is nill? Due to the normal condition or due to a bug?

lib/install/management/inspect.go
vburenin: I would add a comment that this is purely for cosmetic purposes to improve logs readability. Otherwise it can be removed by another developer since it really looks redundant.

lib/install/management/appliance.go
emlin: we planed to use errors sometime ago, but now both of them are used. I'm not going to touch that in this PR.

lib/install/management/appliance.go
emlin: what's the overhead of `&`?

lib/install/management/configure.go
emlin: not sure, most likely is bug

vburenin: In this case I would rather fail with the nil panic.

emlin: hmm.. I prefer to keep an error here, to make sure no matter how user will get an error message, instead of a panic

PR #5099, opened 2017-05-12
Allow environment variables instead of some command line options [specific ci=6-04-Create-Basic]

### LINKED ISSUE #3531: vic-machine should be able to take env vars for common params
Being able to set username, password, target and thumbprint as env vars would make vic-machine usage so much more pleasant. 

As an example, going from 
> vic-machine ls
> vic-machine info --name=<name>

... is so much cleaner if all the common options are implied.

markpeek:
Add environment variable overrides for:
- --target => VIC\_MACHINE\_URL
- --user => VIC\_MACHINE\_USER
- --password => VIC\_MACHINE\_PASSWORD
- --thumbprint => VIC\_MACHINE_THUMBPRINT

Fixes #3531


hickeng:
@markpeek does this work as expected when you've a VIC_MACHINE_TARGET set, but are using profiles via the https://github.com/vmware/vic/blob/master/infra/scripts/bash-helpers.sh script?
Question is essentially whether the argument or the envvar takes precedence and what you're expected behaviour is.

markpeek:
@hickeng hmmm...odd, your comment didn't show up in my email but the others comments did.

Not VIC\_MACHINE\_TARGET but VIC\_MACHINE\_URL. No, I'm not running bash-helpers.sh (didn't know it existed) however, command line arguments override environment variables which is the right thing to do.

mdubya66:
@hickeng is @markpeek clear to merge now?

Issue #4681, opened 2017-04-11
Not able to docker login into registry with self-signed certificates


shadjiiski:
**User Statement:**

As a VIC-engine user, I want to be able to login from the command line (`docker login`) into registries (including harbor) that are secured with self-signed certificates.

**Details:**
I have a `--no-tls` VCH, configured with the `--insecure-registry X.X.X.X:443` flag. (This is the harbor registry, deployed as part of the 1.1 OVA rc4). Doing a `docker login` fails:
```
$ docker login X.X.X.X:443 -u admin -p <admin-pass>
Error response from daemon: Unexpected http code: 400, URL: http://X.X.X.X:443/v2/
```

The error message suggests that a wrong schema is being used, however this is misleading - checking the docker personality logs reveals that the correct address is used first, but the connection fails, because the certificate is self-signed and untrusted. Then a http schema is applied for a second connection attempt that fails with the error that gets reported in the CLI.

**related docker-personality logs**
```
Apr 11 2017 13:44:34.444Z DEBUG [BEGIN] [github.com/vmware/vic/lib/apiservers/engine/backends.(*System).AuthenticateToRegistry:289]
Apr 11 2017 13:44:34.445Z DEBUG [BEGIN] [github.com/vmware/vic/pkg/fetcher.(*URLFetcher).Head:361] https://10.192.189.62:443/v2/
Apr 11 2017 13:44:34.483Z DEBUG [ END ] [github.com/vmware/vic/pkg/fetcher.(*URLFetcher).Head:361] [38.577614ms] https://10.192.189.62:443/v2/
Apr 11 2017 13:44:34.484Z ERROR Looking up OAuth URL failed: Head https://10.192.189.62:443/v2/: x509: certificate signed by unknown authority
Apr 11 2017 13:44:34.484Z DEBUG [BEGIN] [github.com/vmware/vic/pkg/fetcher.(*URLFetcher).Head:361] http://10.192.189.62:443/v2/
Apr 11 2017 13:44:34.485Z DEBUG [ END ] [github.com/vmware/vic/pkg/fetcher.(*URLFetcher).Head:361] [1.289215ms] http://10.192.189.62:443/v2/
Apr 11 2017 13:44:34.485Z ERROR Looking up OAuth URL failed: Unexpected http code: 400, URL: http://10.192.189.62:443/v2/
[...]

Issue #2911, opened 2016-10-26
vic-machine: reconfigure appliance


hickeng:
We **require** an ability to reconfigure the appliance in place - this is absolutely crucial to allow for credential update (#1879), and allows for very likely operational updates, such as adding volume stores, container networks, permitted registries, et al.

This overlaps directly with the upgrade path as upgrade is essentially a reconfigure with updated ISOs.

Addressing this issue is entirely in vic-machine and should:
- [ ] split current `vic-machine create` into create and configure (most of the create options are relevant to configure)
- [ ] add an installer mechanism to load the VCH config into a [Data structure](https://github.com/vmware/vic/blob/master/lib/install/data/data.go#L29)
- [ ] configure and upgrade should combine provided CLI options with the loaded Data (do not support _removal_ of elements such as image stores, volume stores or container networks - that is significant extra complexity)
- [ ] configuration is validated using the full path used by create currently, not the shortcut that upgrade added (this shortcut should be removed).
- [ ] the configuration should be committed to the VCH in much the same manner on all paths (appliance powered off, snapshot taken, config applied, ....)

Creating this as an epic in the expectation that the main tasks be broken out where they can be done in standalone fashion.

Related:
#1797 #1879

mobla:
It will definitely be useful if we have an option to reconfigure VCH options esp. to configure with private registries (harbor) and also need a capability to show that it is configured to store images to a private registry in "docker -H info" command. Currently it shows "Registry: registry-1.docker.io"

emlin:
Two things still to be considered in this story:
- We didn't define the scope of update from user perspective. Not everything configured in vic-machine create can be updated.
  For example, the bridge network ip range. If one range is already there, and some containers created in that scope, we cannot simply remove the previous ip range and update to a separated new range. But we probably could add new range scope if we support multiple discontinuous ip ranges.
  And same for the image store, if there is already images persisted in existing image datastore, we might insert another image store, as long as we could support multiple image stores from portlayer.
  Right now, we still do not have this kind of function support from portlayer, and we might not need them immediately, so from story point of view, we still need a update scope.
- For load VCH configure back to create data, I still have some concern:
  * Create data is designed for easy user experience. For example, all vSphere element is using name, instead of mob-id. But VCH configure is using vSphere mob-id to make sure any rename post vic-machine create, does not impact VCH functionality. 
[...]

PR #5107, opened 2017-05-13
Add support for whitelisting registries

### LINKED ISSUE #3586: Whitelist Container Registries
**Story**
As an administrator of a VCH I would like to be able to restrict where images can be obtained from during VCH deployment

**Details**
This will specifically deal with `pull` as the image acquisition mechanism, not import/load/et al. As such this issue is about restricting registries that may be used as image sources for a given VCH.
Registry whitelist will be specified by a vic-machine list option `--whitelist-registry` taking the following values:
* IP address
* FQDN

`vic-machine` should confirm that at least one registry is accessible with the supplied configuration - i.e. certificate validation succeeds using public or `--registry-cas`, or registry is specified as insecure.
If `vic-machine` cannot connect to the specified registries a warning should be given that the configuration could not be validated.

**Acceptance**
- [ ] `pull` from a whitelisted registry succeeds, whether FQDN, IP, or shortname is used (latter may require DNS search suffix: #2582) - assumes appropriate registry TLS IPSANs, etc.
- [ ] `pull` from a non-whitelisted registry fails, whether FQDN, IP, shortname, IPv6 addr, etc.
- [ ] `pull` from an "insecure" whitelisted registry succeeds only if registry is also listed via `--insecure-registry`
- [ ] vic-machine reports an error if no registry can be verified
- [ ] if no whitelist is specified, pulls from secure registries succeed and pulls from insecure registries fail without `--insecure-registry`
- [ ] scenario doc for restricting to:
  - only docker hub
  - only harbor
  - _other probable scenarios_

Related: #2582 

**Original request**
To ensure use only of approve registries, would it be possible to allow whitelisting of registries in VIC Engine? This would allow VIC Engine and Harbour to be setup in a private, corporate environment, and then restrict access in VIC Engine to images in the local Harbour registry only.

Something similar has has been implemented in OpenShift (https://docs.openshift.com/enterprise/3.2/install_config/install/docker_registry.html#whitelisting-docker-registries)

sflxn:
Add ability to put VCH into whitelist mode.  In whitelist
mode, only registries specified in --whitelist-registry and
--insecure-registry may be access.  In public mode, any
public registry or --insecure-registry may be access.

Removed caching of vch config from personality server.
Registry certs and registry lists are now created dynamically.
This was done to support dynamic config in the future.

Fixed docker login for insecure registries to support testing
whitelist + insecure registries installation.

Fixed docker info to display whitelist and insecure registries.

Add support for wildcard domains and CIDR registry format for
docker login and docker pull.

Added integration tests that also deploys Harbor to verify
a matrix of whitelist support in both vic-machine and docker
info, docker login, docker pull.

resolves #3586 and #4681

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -104,6 +104,7 @@ type Create struct {
 	containerNetworksDNS      cli.StringSlice `arg:"container-network-dns"`
 	volumeStores              cli.StringSlice `arg:"volume-store"`
 	insecureRegistries        cli.StringSlice `arg:"insecure-registry"`
+	whitelistRegistries       cli.StringSlice `arg:"whitelist-registry"`
 	dns                       cli.StringSlice `arg:"dns-server"`
 	clientNetworkName         string
 	clientNetworkGateway      string
@@ -445,7 +446,12 @@ func (c *Create) Flags() []cli.Flag {
 		cli.StringSliceFlag{
 			Name:  "insecure-registry, dir",
 			Value: &c.insecureRegistries,
-			Usage: "Specify a list of permitted insecure registry server URLs",
+			Usage: "Specify a list of permitted insecure registry server addresses",
+		},
+		cli.StringSliceFlag{
+			Name:  "whitelist-registry, wr",
+			Value: &c.whitelistRegistries,
+			Usage: "Specify a list of permitted whitelist registry server addresses (insecure addresses still require the --insecure-registry option in addition)",
 		},
 
 		// proxies
@@ -996,6 +1002,16 @@ func (c *Create) processRegistries() error {
 			return cli.NewExitError(fmt.Sprintf("%s is an invalid format for registry url", registry), 1)
 		}
 		c.InsecureRegistries = append(c.InsecureRegistries, *regurl)
+	}
+
+	// load a list of whitelisted registries
+	for _, registry := range c.whitelistRegistries {
+		regurl, err := validate.ParseURL(registry)
+
+		if err != nil {
+			return cli.NewExitError(fmt.Sprintf("%s is an invalid format for registry url", registry), 1)
+		}
+		c.WhitelistRegistries = append(c.WhitelistRegistries, *regurl)
 	}
 
 	return nil
--- a/doc/design/registries/whitelist-registry.md
+++ b/doc/design/registries/whitelist-registry.md
@@ -1,0 +1,159 @@
+# Whitelist Registries
+
+vSphere Integrated Containers 1.2 (VIC) added the ability to whitelist registry access in an installed VCH.  When one or more registries are whitelisted for the VCH at install time the VCH goes into 'whitelist mode'.  From this point on, the VCH will only allow access to registries in its list of whitelisted registries.  In this mode, users will not be able to access any non-whitelisted registries, public or private.
+
+## Specifying Whitelist Registries at Installation
+
+Whitelisted registries can be declared during a VCH installation with *vic-machine* parameter, `--whitelist-registry`.  Two other vic-machine parameters affect whitelist registries, `--registry-ca` and `--insecure-registry`.  When whitelisted registries are declared during installation, the latter two parameters acts as modifiers to the whitelisted registries.
+
+Registry-ca declares additional certificates to verify access to registry servers secured with TLS.  If a registry is declared as a whitelist registry and not an insecure registry (discussed below), the VCH must have access to the server certificate to verify access.  The Photon-based VM that VIC uses has a base set of well-known certificates from public CAs.  If a whitelist registry uses a certificate that is not in that set of well-known certificates, the certificate must be uploaded to the VCH via vic-machine's `--registry-ca` parameter.
+
+Insecure-registry declares a registry server that can be used without requiring TLS certificate verification.  This modifies the whitelist label and takes precedence.  For instance, if a registry is declared with `--whitelist-registry` and with `--insecure-registry`, the VCH will assume the registry is an insecure whitelisted registry.  If the registry is listed with only `--whitelist-registry`, then the VCH will attempt to verify access using certificates.
+
+If a registry is declared with `--insecure-registry` but not with `--whitelist-registry`, vic-machine will add the insecure registries to the list of whitelist registries *IF* at least one whitelist registry was declared.
+
+A note about certificates.  During installation, vic-machine will attempt to verify the registry server is actually a valid registry server.  It will also attempt to validate that the certificates declared in `--registry-ca` are valid for the secure whitelisted registries.  Vic-machine only performs best effort validation of registry servers.  It will not remove the server's access from the VCH if it cannot validate the server.
+
+Acceptable values for whitelist registry values are numbered IP, FQDN, CIDR formatted range, and wildcard domains.  If a CIDR format is used, e.g. 192.168.1.1/24, then the VCH will whitelist any IP address within that subnet.  Vic-machine will not try to validate CIDR defined ranges.  If a wildcard domain is provided, e.g. *.company.com, the VCH will whitelist any IP address or FQDN address that it can validate against the domain provided during installation.  A numeric IP address will cause the VCH to perform a reverse DNS lookup to validate against that wild card domain.  As with CIDR values, vic-machine will not attempt to validate wildcard domains during installation.  Examples are provided below.
+
+The parameter `--whitelist-registry` creates a list of registries.  If multiple whitelist registries need to be declared, repeat `--whitelist-registry` multiple times during installation for each registry.
+
+### Example: vch installation with vic-machine
+
+This example installs 2 whitelist registries and 1 insecure registry.
+
+```
+./vic-machine-linux create --target=10.2.2.5 --image-store=datastore1 --name=vic-docker --user=root --password=xxxxx --compute-resource="/ha-datacenter/host/office2-sfo2-dhcp121.mycompany.com/Resources" --bridge-network=vic-network --debug=0 --volume-store=datastore1/test:default --tls-cname=*.mycompany.com --whitelist-registry="10.2.40.40:443" --whitelist-registry=10.2.2.1/24 --whitelist-registry=*.mycompany.com --insecure-registry=192.168.100.207  --registry-ca=/home/admin/mycerts/ca.crt
+```
+
+### Example: vic-machine's output during installation
+
+Below is a snippet from the vic-machine output for the above command.
+
+```
+May 15 2017 16:36:12.453-07:00 WARN  Unable to confirm insecure registry 192.168.100.207 is a valid registry at this time.
+May 15 2017 16:36:12.505-07:00 INFO  Insecure registries = 192.168.100.207
+May 15 2017 16:36:12.505-07:00 INFO  Whitelist registries = 10.2.40.40:443, 10.2.2.1/24, *.mycompany.com, 192.168.100.207
+```
+
+Had the above command also included --debug=1 (or higher), the following would be the output
+
+```
+May 15 2017 16:36:12.453-07:00 WARN  Unable to confirm insecure registry 192.168.100.207 is a valid registry at this time.
+May 15 2017 16:36:12.505-07:00 DEBUG  Secure registry 10.2.40.40:443 confirmed.
+May 15 2017 16:36:12.505-07:00 DEBUG  Skipping registry validation for 10.2.2.1/24
+May 15 2017 16:36:12.505-07:00 DEBUG Skipping registry validation for *.eng.vmware.com
+May 15 2017 16:36:12.505-07:00 INFO  Insecure registries = 192.168.100.207
+May 15 2017 16:36:12.505-07:00 INFO  Whitelist registries = 10.2.40.40:443, 10.2.2.1/24, *.mycompany.com, 192.168.100.207
+```
+
+There are a few things to note from this snippet.
+
+1. The confirmation of the insecure registry was not attempted.
+2. The whitelist registry that is secured was confirmed.
+3. Both CIDR and wildcard domain declared as whitelist were skipped during validation.
+4. The final whitelist registry list contains all registries declared with both --whitelist-registry and --insecure-registry.
+5. While not stated yet, the IP address above that contained :443 will not prevent users to use just the server address during docker commands.  This will be discussed below.
+
+## Using Docker Commands Against a VCH in 'Whitelist mode'
+
+VIC currently supports Docker commands that are most applicable for production deployment of containers.  The commands that are affected by whitelist mode are docker info, docker login, and docker pull.  Below are examples of docker commands issued for the VCH installed with the command above.  Let's assume vic-machine properly installs a VCH with the above command and it reports the VCH has an FQDN of myvch.mycompany.com.
+
+### Example: docker -H myvch.mycompany.com info
+
+```
+devbox:~/$ docker -H myvch.mycompany.com:2376 --tlsverify --tlscacert="vic-docker/ca.pem" --tlscert="vic-docker/cert.pem" --tlskey="vic-docker/key.pem" info
+
+Containers: 0
+ Running: 0
+ Paused: 0
+ Stopped: 0
+Images: 0
+Server Version: v1.1.0-rc3-0-c913391
+Storage Driver: vSphere Integrated Containers v1.1.0-rc3-0-c913391 Backend Engine
+VolumeStores: default
+vSphere Integrated Containers v1.1.0-rc3-0-c913391 Backend Engine: RUNNING
+ VCH CPU limit: 10414 MHz
+ VCH memory limit: 58.61 GiB
+ VCH CPU usage: 3103 MHz
+ VCH memory usage: 56.03 GiB
+ VMware Product: VMware ESXi
+ VMware OS: vmnix-x86
+ VMware OS version: 6.0.0
+ Insecure Registries: 192.168.100.207
+ Registry Whitelist Mode: enabled
+ Whitelisted Registries: 10.2.40.40:443, 10.2.2.1/24, *.mycompany.com, 192.168.100.207
+Plugins: 
+ Volume: vsphere
+ Network: bridge
+Swarm: inactive
+Operating System: vmnix-x86
+OSType: vmnix-x86
+Architecture: x86_64
+CPUs: 10414
+Total Memory: 58.61 GiB
+ID: vSphere Integrated Containers
+Docker Root Dir: 
+Debug Mode (client): false
+Debug Mode (server): false
+Registry: registry-1.docker.io
+Experimental: false
+Live Restore Enabled: false
+```
+
+There are a few things to note in the output of this docker info call.
+
+1. Insecure Registry and whitelist registry lists are shown.
+2. There is a message, 'Registry Whitelist Mode: enabled'.  If no whitelist registries are declared during installation, this message will not be shown.
+3. 'Registry: registry-1.docker.io' is displayed even though that address was not whitelisted.  This is the address for docker hub.  It does not mean docker hub is accessible (shown in example below).  It is simply the default registry that is attempted when attempting to login or pull without a registry address.
+
+### Example: docker -H myvch.mycompany.com login 10.2.40.40
+
+```
+devbox:~/$ docker -H myvch.mycompany.com:2376 --tlsverify --tlscacert="vic-docker/ca.pem" --tlscert="vic-docker/cert.pem" --tlskey="vic-docker/key.pem" login 10.2.40.40
+
+Username: 
+Password: 
+Login Succeeded
+```
+
+In this example, a command was issued to log onto a registry that was declared during installation.  Note, :443 was included during installation but left off during docker login.  The VCH will accept either form of the address.
+
+### Example: docker -H myvch.mycompany.com login
+
+```
+devbox:~/$ docker -H myvch.mycompany.com:2376 --tlsverify --tlscacert="vic-docker/ca.pem" --tlscert="vic-docker/cert.pem" --tlskey="vic-docker/key.pem" login
+
+Login with your Docker ID to push and pull images from Docker Hub. If you don't have a Docker ID, head over to https://hub.docker.com to create one.
+Username: user
+Password: 
+Error response from daemon: Access denied to unauthorized registry (registry-1.docker.io) while VCH is in whitelist mode
+```
+
+Notice when the registry address is left off, it attempts to access docker hub (which was indicated in the docker info output above), but the VCH denies access.
+
+### Example: docker -H myvch.mycompany.com pull 10.2.40.40/test/busybox
+
+```
+devbox:~/$ docker -H myvch.mycompany.com:2376 --tlsverify --tlscacert="vic-docker/ca.pem" --tlscert="vic-docker/cert.pem" --tlskey="vic-docker/key.pem" pull 10.2.40.40/test/busybox
+
+Using default tag: latest
+Pulling from test/busybox
+c05511d7505a: Pull complete 
+a3ed95caeb02: Pull complete 
+Digest: sha256:85f3a6aadbb0f25e148d9cfbcf23fbb206f7e6159ea168c33ac51e76fdff4b8e
+Status: Downloaded newer image for test/busybox:latest
+```
+
+This succeeds as it should.
+
+### Example: docker pull busybox
+
+```
+devbox:~/$ docker -H myvch.mycompany.com:2376 --tlsverify --tlscacert="vic-docker/ca.pem" --tlscert="vic-docker/cert.pem" --tlskey="vic-docker/key.pem" pull busybox
+
+Using default tag: latest
+Access denied to unauthorized registry (docker.io) while VCH is in whitelist mode
+```
+
+An attempt to pull from docker hub fails with a message that access was denied while the VCH is in whitelist mode.
--- a/lib/apiservers/engine/backends/backends.go
+++ b/lib/apiservers/engine/backends/backends.go
@@ -26,7 +26,6 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/docker/daemon/events"
-	"github.com/docker/docker/registry"
 	rc "github.com/go-openapi/runtime/client"
 	"github.com/go-openapi/swag"
 
@@ -45,8 +44,7 @@ import (
 )
 
 const (
-	PortlayerName      = "Backend Engine"
-	IndexServerAddress = "registry-1.docker.io"
+	PortlayerName = "Backend Engine"
 
 	// RetryTimeSeconds defines how many seconds to wait between retries
 	RetryTimeSeconds = 2
@@ -58,16 +56,16 @@ var (
 	portLayerName       string
 	productName         string
 	productVersion      string
+	whitelistRegistries string //Friendly formatted list
+	insecureRegistries  string //Friendly formatted list
 
-	vchConfig *config.VirtualContainerHostConfigSpec
-
-	insecureRegistries []string
-	RegistryCertPool   *x509.CertPool
+	vchConfig        *config.VirtualContainerHostConfigSpec
+	RegistryCertPool *x509.CertPool
 
 	eventService *events.Events
 )
 
-func Init(portLayerAddr, product string, config *config.VirtualContainerHostConfigSpec, insecureRegs []url.URL) error {
+func Init(portLayerAddr, product string, config *config.VirtualContainerHostConfigSpec) error {
 	_, _, err := net.SplitHostPort(portLayerAddr)
 	if err != nil {
 		return err
@@ -86,6 +84,7 @@ func Init(portLayerAddr, product string, config *config.VirtualContainerHostConf
 			portLayerName = product + " " + productVersion + " Backend Engine"
 		}
 
+		parseFriendlyRegistryString(config) // For docker info
 		loadRegistryCACerts()
 	} else {
 		portLayerName = product + " Backend Engine"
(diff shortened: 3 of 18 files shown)


doc/design/registries/whitelist-registry.md
```diff
@@ -0,0 +1,151 @@
...
+
+Registry-ca declares additional certificates to verify access to registry servers secured with TLS.  If a registry is declared as a whitelist registry and not an insecure registry (discussed below), the VCH must have access to the server certificate to verify access.  The Photon-based VM that VIC uses has a base set of well-known certificates from public CAs.  If a whitelist registry uses a certificate that is not in that set of well-known certificates, the certificate must be uploaded to the VCH via vic-machine's --registry-ca parameter.
+
+Insecure-registry declares access to a registry server skip the TLS verification.  This modifies the whitelist label and takes precedence.  For instance, if a registry is declared with --whitelist-registry and with --insecure-registry, the VCH will assume the registry is an insecure whitelisted registry.  If the registry is listed with only --whitelist-registry, then the VCH will attempt to verify access using certificates.
+
+If a registry is declared with --insecure-registry but not with --whitelist-registry, vic-machine will add the insecure registries to the list of whitelist registries *IF* at least one whitelist registry was declared.
```
emlin: Please also add description for vic-machine configure.
In general, if multiple configurations are related to each other, we'd tell user to update those configurations at one time, to avoid confusion. For example, if there is one white list registry added, and another one as insecure registry. If user removed the first white list registry, should we think white list registry is disabled?

In addition, we have listed reconfiguration for all current options, please add new issue to track the requirement for new configuration items. We already had a cut line for 1.2 release, so you might put the new issue into overall reconfiguration epic: #2911

lib/apiservers/engine/backends/backends.go
```diff
@@ -212,8 +196,18 @@ func ProductVersion() string {
 	return productVersion
 }
 
+// VchConfig gets the VIC appliance VM's configspec
 func VchConfig() *config.VirtualContainerHostConfigSpec {
-	return vchConfig
+	var vchConfig config.VirtualContainerHostConfigSpec
```
emlin: why do you need to reload the appliance vm configuration? It has been set during Init.
the reason I ask this is because this step will assume portlayer is running in VCH VM, then it cannot be used by vic-machine

lib/apiservers/engine/backends/backends.go
```diff
@@ -256,13 +250,35 @@ func createImageStore() error {
 	return nil
 }
 
-func InsecureRegistries() []string {
-	registries := []string{}
-	for _, reg := range insecureRegistries {
-		registries = append(registries, reg)
+func parseFriendlyRegistryString(config *config.VirtualContainerHostConfigSpec) {
+	// Set index server to list whitelist servers
+	if len(config.RegistryWhitelist) > 0 {
```
emlin: I think the next for loop will skip if list length is 0

lib/apiservers/engine/backends/system.go
```diff
@@ -381,3 +413,46 @@ func FetchVolumeStores(client *client.PortLayer) (string, error) {
...
 	return strings.Join(res.Payload.Stores, " "), nil
 }
+
+// RegistrySetContains checks if a URL is within the set of insecure registries
+func RegistrySetContains(registrySet []url.URL, remoteURL *url.URL) bool {
+	containedOK := false
+	for _, rs := range registrySet {
+		// if CIDR was provided for insecure registry range, check if login URL within that ip net
+		if strings.Index(rs.Host, "/") != -1 {
+			remoteIP := net.ParseIP(remoteURL.Hostname())
```
emlin: this line can be moved out the registrySet for loop, to avoid duplicate parsing in loop.

lib/apiservers/engine/backends/system.go
```diff
@@ -381,3 +413,46 @@ func FetchVolumeStores(client *client.PortLayer) (string, error) {
...
+				containedOK = true
+				break
+			}
+		}
+
+		// if wildcard domain was provided for insecure registry range, check if login URL within
+		// that domain
+		if strings.HasPrefix(rs.Host, "*") {
+			rsDomains := strings.SplitAfter(rs.Host, "*.")
+			remoteDomains, err := net.LookupAddr(remoteURL.Host)
```
emlin: due to same reason, this can be moved out of for loop as well

lib/apiservers/engine/backends/backends.go
sflxn: VIC Product wants to be able to dynamically load configuration.  We're being forced to do this.

emlin: but who is going to modify the configuration?

sflxn: We don't know yet.  It was being hashed out the last time I heard.  It maybe Admiral.

lib/apiservers/engine/backends/backends.go
sflxn: Good point

lib/apiservers/engine/backends/backends.go
sflxn: After chatting with @hickeng, it appears we are not going to get config pushed to us.  Instead we're going to pull more restrictive config from somewhere else so I reverted the code back to caching the VCH config.

lib/apiservers/engine/backends/backends.go
```diff
@@ -45,8 +43,8 @@ import (
 )
 
 const (
-	PortlayerName      = "Backend Engine"
-	IndexServerAddress = "registry-1.docker.io"
+	PortlayerName         = "Backend Engine"
+	HubIndexServerAddress = "registry-1.docker.io"
```
anchal-agrawal: We can use `imagec.DefaultDockerURL` instead.

lib/apiservers/engine/backends/system.go
```diff
@@ -43,6 +44,8 @@ import (
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/version"
 
+	"net"
```
anchal-agrawal: gofmt: we should move this to the top import group.

doc/design/registries/whitelist-registry.md
```diff
@@ -0,0 +1,151 @@
+# Whitelist Registries
+
+vSphere Integrated Containers 1.2 (VIC) added the ability to whitelist registry access in an installed VCH.  When a registry is added to the VCH at installed time, the VCH goes into 'whitelist' mode.  From this point on, the VCH will only allow access to registries in its list of whitelisted registries.  In this mode, users will not be able to access public registries, such as docker hub.
```
jzt: `s/installed time/install time`

doc/design/registries/whitelist-registry.md
```diff
@@ -0,0 +1,151 @@
...
+
+## Specifying Whitelist Registries at Installation
+
+Whitelist registries can be declared during a VCH installation with *vic-machine* parameter, --whitelist-registry.  Two other vic-machine parameters affect whitelist registries, --registry-ca and --insecure-registry.  When whitelist registries are declared during installation, the latter two parameters acts as modifiers to the whitelist registries.
```
jzt: `s/Whitelist registries/Whitelisted registries`

jzt: Unless we want to call them "Whitelist registries".. I'm seeing that peppered through the doc, not sure if that's our intended nomenclature, ignore if so!

doc/design/registries/whitelist-registry.md
```diff
@@ -0,0 +1,151 @@
...
+
+Registry-ca declares additional certificates to verify access to registry servers secured with TLS.  If a registry is declared as a whitelist registry and not an insecure registry (discussed below), the VCH must have access to the server certificate to verify access.  The Photon-based VM that VIC uses has a base set of well-known certificates from public CAs.  If a whitelist registry uses a certificate that is not in that set of well-known certificates, the certificate must be uploaded to the VCH via vic-machine's --registry-ca parameter.
+
+Insecure-registry declares access to a registry server skip the TLS verification.  This modifies the whitelist label and takes precedence.  For instance, if a registry is declared with --whitelist-registry and with --insecure-registry, the VCH will assume the registry is an insecure whitelisted registry.  If the registry is listed with only --whitelist-registry, then the VCH will attempt to verify access using certificates.
```
jzt: `Insecure-registry declares access to a registry server skip the TLS verification.`  This reads a little off, specifically "...to a registry server skip the TLS verification"

doc/design/registries/whitelist-registry.md
```diff
@@ -0,0 +1,151 @@
...
+
+If a registry is declared with --insecure-registry but not with --whitelist-registry, vic-machine will add the insecure registries to the list of whitelist registries *IF* at least one whitelist registry was declared.
+
+A note about certificates.  During installation, vic-machine will attempt to verify the registry server is actually a valid registry server.  It will also attempt to validate the certificates declared in --registry-ca are valid for the secure whitelist registries.  Vic-machine only performs best effort validation of registry servers.  It will not remove the server's access from the VCH if it cannot validate the server.
```
jzt: I think `attempt to validate the certificates declared in --registry-ca are valid` should be `attempt to validate that the certificates declared in --registry are valid`

doc/design/registries/whitelist-registry.md
```diff
@@ -0,0 +1,151 @@
...
+
+Acceptable values for whitelist registry values are numbered IP, FQDN, CIDR formatted range, and wildcard domains.  If a CIDR format is used, e.g. 192.168.1.1/24, then the VCH will whitelist any IP address within that subnet.  Vic-machine will not try to validate CIDR defined ranges.  If a wildcard domain is provided, e.g. *.company.com, the VCH will whitelist any IP address or FQDN address that it can validate against the domain provided during installation.  A numeric IP address will cause the VCH to perform a reverse DNS lookup to validate against that wild card domain.  As with CIDR values, vic-machine will not attempt to validate wildcard domains during installation.  Examples are provided below.
```
jzt: `Acceptable values for whitelist registry are...`

cmd/vic-machine/create/create.go
```diff
@@ -104,6 +104,7 @@ type Create struct {
 	containerNetworksDNS      cli.StringSlice `arg:"container-network-dns"`
 	volumeStores              cli.StringSlice `arg:"volume-store"`
 	insecureRegistries        cli.StringSlice `arg:"insecure-registry"`
+	whitelistRegistries       cli.StringSlice `arg:"whitelist-registry"`
```
jzt: 👍

lib/install/validate/validator.go
```diff
@@ -501,21 +504,140 @@ func (v *Validator) certificateAuthorities(ctx context.Context, input *data.Data
...
+
 	// copy the list of insecure registries
-	conf.InsecureRegistries = input.InsecureRegistries
+	conf.InsecureRegistries = insecureRegistries
+
+	// copy the list of whitelist registries
+	conf.RegistryWhitelist = whitelistRegistries
+
+	// create vic-machine info message
+	if len(conf.InsecureRegistries) > 0 {
```
jzt: Minor: you could factor these 3 loops into a function and just call it three times.

doc/design/registries/whitelist-registry.md
lcastellano: The behavior where the exclusive insecure-registry is added to the whitelist, only if the whitelist is empty, seems like a justification for an internal implementation choice. Is there a rationale for this?

doc/design/registries/whitelist-registry.md
hmahmood: Can't I add docker hub to the whitelist? Why the restriction with docker hub?

doc/design/registries/whitelist-registry.md
lcastellano: Regarding this line:
_.. the VCH will attempt to verify access using certificates_
Does that mean that the VCH will attempt to verify access validating the registry server-certificate. If that is the case, it should be worded that way.

lib/apiservers/engine/backends/system.go
```diff
@@ -381,3 +408,45 @@ func FetchVolumeStores(client *client.PortLayer) (string, error) {
 
 	return strings.Join(res.Payload.Stores, " "), nil
 }
+
+// RegistrySetContains checks if a URL is within the set of insecure registries
+func RegistrySetContains(registrySet []url.URL, remoteURL *url.URL) bool {
+	containedOK := false
+	remoteIP := net.ParseIP(remoteURL.Hostname())
+	remoteDomains, _ := net.LookupAddr(remoteURL.Hostname())
```
hmahmood: You may not need to do this if none of the registry entries are domain names, wild-carded or otherwise.

lib/apiservers/engine/backends/system.go
```diff
@@ -381,3 +408,45 @@ func FetchVolumeStores(client *client.PortLayer) (string, error) {
...
+
+	for _, rs := range registrySet {
+		// if CIDR was provided for insecure registry range, check if login URL within that ip net
+		if strings.Index(rs.Path, "/") != -1 {
```
hmahmood: Instead of the three `if` clauses, can I suggest that you use an interface instead? Concrete implementations for the three cases can then have the logic in them, and you just make one call to the method here, e.g. `Contains`, etc.

lib/install/validate/validator.go
```diff
@@ -501,21 +504,140 @@ func (v *Validator) certificateAuthorities(ctx context.Context, input *data.Data
...
+
+	// Remove intersection between insecure registries and whitelist registries from whitelist set so
+	// we can ensure we test the exclusion set with certs
+	var exist bool
+	var idx int
+	var s url.URL
+	for _, r := range input.InsecureRegistries {
+		exist = false
+		for idx, s = range secureRegistries {
+			if r.Host == s.Host {
```
hmahmood: Will this always be true? What if one is an ip address, and the other the fqdn for the ip address?

lib/install/validate/validator.go
```diff
@@ -501,21 +504,140 @@ func (v *Validator) certificateAuthorities(ctx context.Context, input *data.Data
...
 
-	conf.RegistryCertificateAuthorities = input.RegistryCAs
+	// Test insecure registries' reachability
+	for _, r := range input.InsecureRegistries {
+		// Make sure address is not a wildcard domain or CIDR.  If it is, do not validate.
+		if strings.HasPrefix(r.Host, "*") {
+			log.Infof("Skipping registry validation for %s", r.Host)
+			continue
+		}
+		if strings.Index(r.Path, "/") != -1 {
```
hmahmood: Use `net.ParseCIDR` to check if address is a CIDR.

lib/install/validate/validator.go
```diff
@@ -501,21 +504,140 @@ func (v *Validator) certificateAuthorities(ctx context.Context, input *data.Data
...
+		} else {
+			log.Infof("Secure registry %s confirmed.", w.Host)
+		}
+	}
+
+	// Return output
+	insecureRegistries = input.InsecureRegistries
+	// If vic-machine had whitelist registry specified
+	if len(input.WhitelistRegistries) > 0 {
+		whitelistRegistries = append(secureRegistries, insecureRegistries...)
```
hmahmood: Do we want to empty out insecureRegistries here?

cmd/vic-machine/create/create.go
```diff
@@ -998,6 +1004,16 @@ func (c *Create) processRegistries() error {
 		c.InsecureRegistries = append(c.InsecureRegistries, *regurl)
 	}
 
+	// load a list of whitelisted registries
+	for _, registry := range c.whitelistRegistries {
+		regurl, err := validate.ParseURL(registry)
```
hmahmood: I am not sure storing all the registries as `url.URL` objects is a good idea. In particular, storing a CIDR this way seems to more trouble that its worth. I would store them just as strings, and then have a custom `parse` function to give me back an object that represents a registry entry. You can add methods to that object/interface to check if a given ip address or hostname matches against that entry.

lib/apiservers/engine/backends/system.go
sflxn: my stupid editor rearranged it

doc/design/registries/whitelist-registry.md
sflxn: You can.  Bad wording.  Let me change it.

lib/apiservers/engine/backends/system.go
sflxn: if registry entry is number IP, we need to do this.

lib/install/validate/validator.go
sflxn: No, we do not empty insecure registries.

cmd/vic-machine/create/create.go
sflxn: They're currently stored as url.  Insecure registries are also stored as url.  If I change it, then I am in vic-machine upgrade territory and that seems more trouble.

lib/install/validate/validator.go
sflxn: We're not checking for equivalence now but can be added.

tests/manual-test-cases/Group21-Registries/21-1-Whitelist.robot
```diff
@@ -0,0 +1,318 @@
...
+    
+    # Install a Harbor server with HTTPS a Harbor server with HTTP
+    Install Harbor To Test Server  protocol=https  name=harbor-https
+    Set Environment Variable  HTTPS_HARBOR_IP  %{HARBOR-IP}
+
+    Install Harbor To Test Server  protocol=http  name=harbor-http
+    Set Environment Variable  HTTP_HARBOR_IP  %{HARBOR-IP}
+
+#    Set Environment Variable  HTTPS_HARBOR_IP  10.118.69.161
+#    Set Environment Variable  HTTP_HARBOR_IP  10.118.69.160
```
mhagen-vmware: Can you remove this?

tests/manual-test-cases/Group21-Registries/21-1-Whitelist.robot
```diff
@@ -0,0 +1,318 @@
...
+    Log  ${output}
+    ${rc}  ${output}=  Run And Return Rc And Output  docker push ${HTTPS_HARBOR_IP}/library/busybox   
+    Should Be Equal As Integers  ${rc}  0
+    Log  ${output}
+#    ${rc}  ${output}=  Run And Return Rc And Output  docker tag busybox ${HTTP_HARBOR_IP}/library/busybox
+#    Should Be Equal As Integers  ${rc}  0
+#    Log  ${output}
+#    ${rc}  ${output}=  Run And Return Rc And Output  docker push ${HTTP_HARBOR_IP}/library/busybox   
+#    Should Be Equal As Integers  ${rc}  0
+#    Log  ${output}
```
mhagen-vmware: Not a big fan of random portions un-commented in the middle of an almost all commented out test.  It doesn't look like this is actually called anywhere either.  Should we just remove this whole keyword for now?

tests/manual-test-cases/Group21-Registries/21-1-Whitelist.robot
```diff
@@ -0,0 +1,318 @@
...
+
+    # Try to login and pull from docker hub (should fail)
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} login --username=victest --password=vmware!123
+    Should Be Equal As Integers  ${rc}  1
+    Should Contain  ${output}  Access denied to unauthorized registry
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull victest/busybox
+    Should Be Equal As Integers  ${rc}  1
+    Should Contain  ${output}  Access denied to unauthorized registry
+
+    Cleanup VIC Appliance On Test Server
```
mhagen-vmware: Any failure within the test case will leave this hanging. Better to include this in your overall cleanup and/or test cleanup keyword.  You can have both a suite teardown and a test teardown, so you could leave the harbor instances through the tests but cleanup each vic appliance.

tests/resources/Harbor-Util.robot
```diff
@@ -54,8 +55,7 @@ Install Harbor To Test Server
     :FOR  ${i}  IN RANGE  20
     \  ${out}=  Run  curl -k ${protocol}://%{HARBOR-IP}
     \  Log  ${out}
-    \  ${status}=  Run Keyword And Return Status  Should Not Contain  ${out}  502 Bad Gateway
-    \  ${status}=  Run Keyword If  ${status}  Run Keyword And Return Status  Should Not Contain  ${out}  Connection refused
+    \  ${status}=  Run Keyword And Return Status  Should Not Contain Any  ${out}  502 Bad Gateway  Connection refused  Connection timed out
```
mhagen-vmware: nice addition!

tests/manual-test-cases/Group21-Registries/21-1-Whitelist.robot
sflxn: Added a test tear down as well

lib/apiservers/engine/backends/system.go
```diff
@@ -211,6 +219,16 @@ func (s *System) SystemInfo() (*types.Info, error) {
 			customInfo := [2]string{systemOSVersion, vchInfo.HostOSVersion}
 			info.SystemStatus = append(info.SystemStatus, customInfo)
 		}
+		if len(vchConfig.InsecureRegistries) > 0 {
+			customInfo := [2]string{insecureRegistriesLabel, InsecureRegistries()}
+			info.SystemStatus = append(info.SystemStatus, customInfo)
+		}
+		if len(vchConfig.RegistryWhitelist) > 0 {
+			customInfo := [2]string{vchWhitelistMode, "enabled"}
```
hickeng: Should we have "disabled" as a value? Easier for parsing/doc'ing than saying that the line doesn't exist in one case.

lib/apiservers/engine/backends/system.go
```diff
@@ -381,3 +408,45 @@ func FetchVolumeStores(client *client.PortLayer) (string, error) {
 
 	return strings.Join(res.Payload.Stores, " "), nil
 }
+
+// RegistrySetContains checks if a URL is within the set of insecure registries
```
hickeng: ~~insecure~~

lib/apiservers/engine/backends/system.go
```diff
@@ -381,3 +408,45 @@ func FetchVolumeStores(client *client.PortLayer) (string, error) {
...
+				}
+				domainParts := strings.SplitAfterN(fqdn, ".", 2)
+				if domainParts[1] == rsDomains[1] {
+					containedOK = true
+					break
+				}
+			}
+		}
+
+		// if IP provided for insecure registry
```
hickeng: suggest a comment that we deliberately don't attempt to check FQND<->IP equivalence

lib/install/validate/validator.go
```diff
@@ -501,21 +504,147 @@ func (v *Validator) certificateAuthorities(ctx context.Context, input *data.Data
 func (v *Validator) registries(ctx context.Context, input *data.Data, conf *config.VirtualContainerHostConfigSpec) {
 	defer trace.End(trace.Begin(""))
 
+	// Check if CAs can be loaded
+	pool := x509.NewCertPool()
```
hickeng: We should add a means to build the endpoint public registry set into vic-machine so we can use a known set of public certs. That way we won't validate successfully on client machine, but fail on VCH.

lib/install/validate/validator.go
```diff
@@ -501,21 +504,147 @@ func (v *Validator) certificateAuthorities(ctx context.Context, input *data.Data
...
+		if exist {
+			secureRegistries = append(secureRegistries[:idx], secureRegistries[idx+1:]...)
+		}
+	}
+
+	// Test insecure registries' reachability
+	for _, r := range input.InsecureRegistries {
+		// Make sure address is not a wildcard domain or CIDR.  If it is, do not validate.
+		if strings.HasPrefix(r.Host, "*") {
+			log.Infof("Skipping registry validation for %s", r.Host)
```
hickeng: Make this debug - we should tell the user if a check fails, we don't tell them about success or skipping of other checks (or we shouldn't be).
We should also explain why we're skipping the check when it's in debug.

lib/install/validate/validator.go
hickeng: Can we also use CIDR in insecure-registry?

lib/install/validate/validator.go
```diff
@@ -501,21 +504,147 @@ func (v *Validator) certificateAuthorities(ctx context.Context, input *data.Data
...
+
+		_, err = registryutils.Reachable(r.Host, "https", "", "", nil, registryValidationTime, true)
+		if err != nil {
+			_, err = registryutils.Reachable(r.Host, "http", "", "", nil, registryValidationTime, true)
+		}
+
+		if err != nil {
+			log.Warnf("Unable to confirm insecure registry %s is a valid registry at this time.", r.Host)
+		} else {
+			log.Infof("Insecure registry %s confirmed.", r.Host)
```
hickeng: Debug

lib/install/validate/validator.go
```diff
@@ -501,21 +504,147 @@ func (v *Validator) certificateAuthorities(ctx context.Context, input *data.Data
...
+
+	// Test insecure registries' reachability
+	for _, r := range input.InsecureRegistries {
+		// Make sure address is not a wildcard domain or CIDR.  If it is, do not validate.
+		if strings.HasPrefix(r.Host, "*") {
+			log.Infof("Skipping registry validation for %s", r.Host)
+			continue
+		}
+		if strings.Index(r.Path, "/") != -1 {
+			log.Infof("Skipping registry validation for %s%s", r.Host, r.Path)
```
hickeng: Debug

lib/install/validate/validator.go
```diff
@@ -501,21 +504,147 @@ func (v *Validator) certificateAuthorities(ctx context.Context, input *data.Data
...
+			log.Infof("Skipping registry validation for %s%s", w.Host, w.Path)
+			continue
+		}
+
+		_, err = registryutils.Reachable(w.Host, "https", "", "", pool, registryValidationTime, false)
+
+		if err != nil {
+			log.Warnf("Unable to confirm secure registry %s is a valid registry at this time.", w.Host)
+		} else {
+			log.Infof("Secure registry %s confirmed.", w.Host)
```
hickeng: debug

lib/install/validate/validator.go
```diff
@@ -501,21 +504,147 @@ func (v *Validator) certificateAuthorities(ctx context.Context, input *data.Data
...
+
+	// Test secure registries' reachability
+	for _, w := range secureRegistries {
+		// Make sure address is not a wildcard domain or CIDR.  If it is, do not validate.
+		if strings.HasPrefix(w.Hostname(), "*") {
+			log.Infof("Skipping registry validation for %s", w.Host)
+			continue
+		}
+		if strings.Index(w.Path, "/") != -1 {
+			log.Infof("Skipping registry validation for %s%s", w.Host, w.Path)
```
hickeng: debug

lib/install/validate/validator.go
```diff
@@ -501,21 +504,147 @@ func (v *Validator) certificateAuthorities(ctx context.Context, input *data.Data
...
+		} else {
+			log.Infof("Insecure registry %s confirmed.", r.Host)
+		}
+	}
+
+	// Test secure registries' reachability
+	for _, w := range secureRegistries {
+		// Make sure address is not a wildcard domain or CIDR.  If it is, do not validate.
+		if strings.HasPrefix(w.Hostname(), "*") {
+			log.Infof("Skipping registry validation for %s", w.Host)
```
hickeng: debug

pkg/registry/utils.go
```diff
@@ -0,0 +1,41 @@
...
+		InsecureSkipVerify: skipVerify,
+		RootCAs:            registryCAs,
+	})
+
+	headers, err := fetcher.Head(url)
+	if err != nil {
+		return "", err
+	}
+	// v2 API requires this check
+	if headers.Get("Docker-Distribution-API-Version") != "registry/2.0" {
```
sflxn: Created a new issue #5259 for this.

doc/design/registries/whitelist-registry.md
```diff
@@ -0,0 +1,151 @@
...
+Registry: registry-1.docker.io
+Experimental: false
+Live Restore Enabled: false
+```
+
+There are a few things to note in the output of this docker info call.
+
+1. Insecure Registry and whitelist registry lists are shown.
+2. There is a message, 'Registry Whitelist Mode: enabled'.  If no whitelist registries are declared during installation, this message will not be shown.
+3. 'Registry: registry-1.docker.io' is displayed even though the VCH is in whitelist mode.  This is the address for docker hub.  It does not mean docker hub is accessible (shown in example below).  It is simply the default registry that is attempted when attempting to login or pull without a registry address.
```
sflxn: Created #5260 to add this feature.

lib/install/validate/validator.go
sflxn: @hickeng Yes, I added the symmetry for insecure-registry since it's goes through the same code path.

lib/apiservers/engine/backends/system.go
sflxn: Went ahead and added "disabled" also.

lib/apiservers/engine/backends/system.go
sflxn: was added

lib/install/validate/validator.go
sflxn: #5268

lib/apiservers/engine/backends/image.go
```diff
@@ -344,12 +345,39 @@ func (i *Image) PullImage(ctx context.Context, image, tag string, metaHeaders ma
...
+	if len(vchConfig.RegistryWhitelist) > 0 && !whitelistOk {
+		err = fmt.Errorf("Access denied to unauthorized registry (%s) while VCH is in whitelist mode", hostnameURL.Hostname())
+		log.Errorf(err.Error())
+		sf := streamformatter.NewJSONStreamFormatter()
+		outStream.Write(sf.FormatError(err))
+		return nil
+	}
+	options.InsecureAllowHTTP = insecureOk
+
+	if insecureOk {
```
sflxn: We don't remove the system root CA pool.  Nil means the user probably didn't pass in a cert for this server (although it's a possibility) so we check the system certs.  If that doesn't work, we bypass TLS.

pkg/registry/utils.go
```diff
@@ -0,0 +1,41 @@
...
+	"net/url"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+
+	urlfetcher "github.com/vmware/vic/pkg/fetcher"
+)
+
+// Reachable test if a registry is a valid registry for VIC use and returns a url with scheme prepended
+func Reachable(registry, schema, username, password string, registryCAs *x509.CertPool, timeout time.Duration, skipVerify bool) (string, error) {
```
sflxn: I wanted this function to be as basic as possible.  We can create another function that does what you're suggesting.  Currently only two functions in the entire code base calls Reachable so it's below my threshold for refactoring the higher level logic you're describing.

PR #5108, opened 2017-05-14
[specific ci=Group1-Docker-Commands]reset runblock in tether after process is launched

### LINKED ISSUE #4296: Ensure that runblocking is configured anew for each container start

emlin:
fixes #4296 

after this change, the container in older version will not block process running even user meant to attach the container because the guestinfo key is changed to writable. 
this issue cannot be fixed by upgrade plugin because we cannot write back container configuration if tether is in older version.

but if container is new, the issue #4296 can be fixed, here is the test result:
```
vagrant@vagrant:~/gopath/src/github.com/vmware/vic$ docker -H 10.192.162.98:2376 --tls run -it busybox sh
Pulling from library/busybox
7520415ce762: Pull complete 
a3ed95caeb02: Pull complete 
Digest: sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f
/ # exit
vagrant@vagrant:~/gopath/src/github.com/vmware/vic$ docker -H 10.192.162.98:2376 --tls ps -a
6c13e905e3ef        busybox             "sh"                4 minutes ago       Exited (0) 10 seconds ago                       modest_engelbart
vagrant@vagrant:~/gopath/src/github.com/vmware/vic$ docker -H 10.192.162.98:2376 --tls start 6c
6c
vagrant@vagrant:~/gopath/src/github.com/vmware/vic$ docker -H 10.192.162.98:2376 --tls ps -a
6c13e905e3ef        busybox             "sh"                5 minutes ago       Up 34 seconds                           modest_engelbart
```

### DIFF
--- a/lib/tether/tether.go
+++ b/lib/tether/tether.go
@@ -582,6 +582,8 @@ func (t *tether) cleanupSession(session *SessionConfig) {
 		log.Debugf("Calling close chan: %s", session.ID)
 		close(session.ClearToLaunch)
 		session.ClearToLaunch = nil
+		// rest Runblock to unblock process start next time
+		session.RunBlock = false
 	}
 }
 
@@ -679,6 +681,10 @@ func (t *tether) launch(session *SessionConfig) error {
 		case <-session.ClearToLaunch:
 			log.Infof("Received the clear signal to launch %s", session.ID)
 		}
+		// reset RunBlock to unblock process start next time
+		session.RunBlock = false
+		prefix := extraconfig.CalculateKeys(t.config, fmt.Sprintf("%s.%s", session.extraconfigKey, session.ID), "")[0]
+		extraconfig.EncodeWithPrefix(t.sink, session, prefix)
 	}
 
 	pid := 0
--- a/lib/tether/tether.go
+++ b/lib/tether/tether.go
@@ -683,8 +683,6 @@ func (t *tether) launch(session *SessionConfig) error {
 		}
 		// reset RunBlock to unblock process start next time
 		session.RunBlock = false
-		prefix := extraconfig.CalculateKeys(t.config, fmt.Sprintf("%s.%s", session.extraconfigKey, session.ID), "")[0]
-		extraconfig.EncodeWithPrefix(t.sink, session, prefix)
 	}
 
 	pid := 0
--- a/lib/tether/tether.go
+++ b/lib/tether/tether.go
@@ -582,7 +582,7 @@ func (t *tether) cleanupSession(session *SessionConfig) {
 		log.Debugf("Calling close chan: %s", session.ID)
 		close(session.ClearToLaunch)
 		session.ClearToLaunch = nil
-		// rest Runblock to unblock process start next time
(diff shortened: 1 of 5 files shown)


lib/tether/tether.go
```diff
@@ -582,6 +582,8 @@ func (t *tether) cleanupSession(session *SessionConfig) {
 		log.Debugf("Calling close chan: %s", session.ID)
 		close(session.ClearToLaunch)
 		session.ClearToLaunch = nil
+		// rest Runblock to unblock process start next time
```
matthewavery: `reset` instead of `rest`?

lib/tether/tether.go
```diff
@@ -679,6 +681,10 @@ func (t *tether) launch(session *SessionConfig) error {
 		case <-session.ClearToLaunch:
 			log.Infof("Received the clear signal to launch %s", session.ID)
 		}
+		// reset RunBlock to unblock process start next time
+		session.RunBlock = false
+		prefix := extraconfig.CalculateKeys(t.config, fmt.Sprintf("%s.%s", session.extraconfigKey, session.ID), "")[0]
+		extraconfig.EncodeWithPrefix(t.sink, session, prefix)
```
caglar10ur: do we need to call this? We are already calling EncodeWithPrefix in a defer statement above

emlin: no, should not

matthewavery:
@emlin I just want to clarify a little more, this means that once the portlayer has been upgraded the old containers are still functional? but will suffer the same issue. So the advised approach after an upgrade would be to remake your existing containers anew to avoid the issue? 😄

emlin:
> just want to clarify a little more, this means that once the portlayer has been upgraded the old containers are still functional? but will suffer the same issue. So the advised approach after an upgrade would be to remake your existing containers anew to avoid the issue? 😄

unfortunately, it's not. after upgrade VCH, the portlayer will be new, and container is still old. That means portlayer will write new runblock, try to block container to launch process, but container is still reading old runblock attribute, which will always be the old value.
so if it's set to true, then it will suffer the same issue, but if it's false, even you run `docker start -ai`, it will not actually wait there for attach.
so it will be worse technically.

PR #5115, opened 2017-05-15
[specific ci=Group6-VIC-Machine] retry functionality for ISO upload during vic-machine create

### LINKED ISSUE #4989: Nightlies 6.0 VSAN Simple and VSAN Complex fail on Should Contain ${line}  .vcva-${VC_VERSION}' is up.
Build 9992 - 05/01/17 vSphere 6.0

There are many instances of this error in the output.xml file.  This happens right after deploying a nimbus testbed.  Both VSAN Simple and VSAN Complex share this issue.

VSAN Simple Error:
`<msg timestamp="20170502 13:35:34.347" level="FAIL">'2017-05-02 06:12:27 -0700: Loading plugin /mts/git/nimbus/lib/testframeworks/test-vpx ...' does not contain '.vcva-3634791' is up. IP:'</msg>`

VSAN Complex Error:
`<msg timestamp="20170502 14:51:21.530" level="FAIL">'2017-05-02 07:12:24 -0700: Loading plugin /mts/git/nimbus/lib/testframeworks/test-vpx ...' does not contain '.vcva-3634791' is up. IP:'</msg>`

### LINKED ISSUE #5041: Nightlies 6.5 vMotion-VCH-Appliance fails to upload iso and continues
Build#9992
vSphere v6.5

```
May  5 2017 09:13:41.425Z INFO  Uploading images for container
May  5 2017 09:13:41.425Z INFO  	"bin/bootstrap.iso"
May  5 2017 09:13:41.427Z INFO  	"bin/appliance.iso"
May  5 2017 09:16:13.765Z ERROR 		Upload failed for "bin/appliance.iso": Put https://10.192.208.115/folder/f1400c59-4370-87f4-e7bf-0200387d3623/V1.1.0-RC3-9992-8F0526F-appliance.iso?dcPath=%2FvcqaDC&dsName=vsanDatastore: write tcp 172.17.0.3:39630->10.192.208.115:443: write: connection reset by peer
May  5 2017 09:16:13.765Z WARN  		Continuing despite failures (due to --force option)
May  5 2017 09:16:13.765Z WARN  		Note: The VCH will not function without "bin/appliance.iso"...
```
This may cause errors of this type:
`<msg timestamp="20170427 09:01:48.932" level="FAIL">'2017-04-27 01:34:22 -0700: Loading plugin <plugin name is different for each error> ...' does not contain '.vcva-4602587' is up. IP:'</msg>`
in the output.xml file.

Testbed deployment took around 30min to fire up on Nimbus.

matthewavery:
Fixes #5041 
fixes #4989 

This implements a very very simple retry with a retry count of 5 for uploading the isos. the retry is done individually for each iso. so currently there are a total of 10 retries. We automatically retry the upload without waiting for now(comment if you do not like that). This should also help with deployments that use `--force`, we attempt more iso uploads before continuing despite a failure.


lib/install/management/create.go
```diff
@@ -121,14 +125,24 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
...
-				log.Errorf("\t\tUpload failed for %q: %s", image, err)
+
+			var uploadErr error
+			for i := 1; i < uploadRetryCount; i++ {
+				uploadErr := d.session.Datastore.UploadFile(d.ctx, image, path.Join(d.vmPathName, key), nil)
+				if uploadErr == nil {
+					break
+				}
+				log.Infof("Upload failed - retrying %d times - upload failed with error: %s", ((uploadRetryCount - 1) - i), uploadErr.Error())
+				uploadErr = nil
```
emlin: should do it here. this set will make you move ahead at line 139

lib/install/management/create.go
```diff
@@ -121,14 +125,24 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
 			defer wg.Done()
 
 			log.Infof("\t%q", image)
-			err := d.session.Datastore.UploadFile(d.ctx, image, path.Join(d.vmPathName, key), nil)
-			if err != nil {
-				log.Errorf("\t\tUpload failed for %q: %s", image, err)
+
+			var uploadErr error
+			for i := 1; i < uploadRetryCount; i++ {
+				uploadErr = nil
```
emlin: feels like you don't need this, cause you'll set it in next line

lib/install/management/create.go
```diff
@@ -121,14 +125,24 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
...
+				uploadErr := d.session.Datastore.UploadFile(d.ctx, image, path.Join(d.vmPathName, key), nil)
```
emlin: do not define variable again

fdawg4l:
I'm puzzled as to why you need the retry at all.  Can we make *any* API calls when the test fails or is it just calls to _this_ datastore which fail?  Retries are fine but they need to be scoped.  It looks like this change will retry the upload regardless of the error even though we know the error we saw in the test.  Under what cases is the retry useful?

Can someone walk me through the debugging here which obviated the need for a retry in this specific method and not others?

fdawg4l:
This comment (https://github.com/vmware/vic/issues/4989#issuecomment-301527996) makes the issue seem like it's specific to our test deployment.  Do we need to make changes to our code for this or can we modify the test?

mhagen-vmware:
So, if I can summarize our conversation between me, @emlin and @matthewavery :
1. We could push this failure back to the user as we are doing now and let them deal with it, so I could certainly wrap some retries around in my test
2. But in large environments that can be a large penalty for the user - in very large VC scenarios just our inventory analysis and checks before we even start the ISO upload can take quite a while.
3. Also, the ISO upload can be a very long time that at any moment could have a network hiccup...
4. Eventually we would love a continue previous ISO upload functionality

But as for now this is an easy fix that gives customers a little less penalty they have to pay when a network hiccup would have normally make them start all over again.

fdawg4l:
The isos are measured in hundreds of megs. And the TCP timeout for Linux is 5m. If we can't upload a hundred meg ISO in under 5m, I think it's safe to say the user experience beyond uploading isos is compromised and we can't account for that. 

Second, inventory analysis may take a while. I don't see how that corresponds to ISO upload issues. The two are not related unless there is a context deadline at play and in that case the fix is simpler than a retry. 

If an ISO upload fails, continuing silently is not the answer. 

I'm NACK'ing this change.

lib/install/management/create.go
```diff
@@ -29,6 +29,10 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/tasks"
 )
 
+const (
+	uploadRetryCount = 5
```
caglar10ur: we have a package for this (pkg/retry) so please use it if retry is the way to go

mhagen-vmware:
@fdawg4l Why are you bringing up TCP timeouts in Linux?  Not understanding how that has anything to do with this...  and not really sure what you are saying at all.  The TCP timeout is an idle timeout, not a hard limit on TCP connections, otherwise no one would be able to transfer large files over a slow connection?

This is just a simple case of if any networking error happens, we should give it another shot instead of bailing immediately and making a customer start all over again..

lib/install/management/create.go
matthewavery: Yes I discussed this with @mhagen-vmware and @emlin . We decided that in the end for now a simple for loop retry will suffice. This is also not meant to be a complex fix. this is simply going to allow vic-machine to recover from *minor* network blips. Nothing more.

matthewavery: If it does not fix the overlying minor issues we can move to `retry.do` which was where I went first :smile:

matthewavery: Alright we are doing that, we at first decided not to. But now we will 😄

matthewavery:
@caglar10ur @emlin Alright I have modified the retry package to accept custom `MaxElapsedTime` arguments. I have also configured vic-machine to attempt retrys for a maximum time of 5 minutes. Please let me know if the changes look acceptable.

pkg/retry/retry.go
```diff
@@ -32,9 +32,15 @@ const (
...
-// Retry retries the given function DefaultRetryCount times while sleeping some time between unsuccessful attempts
+// Do retries the given function until DefaultMaxInterval time passes, while sleeping some time between unsuccessful attempts
 // if retryOnError returns true, continue retry, otherwise, return error
 func Do(operation func() error, retryOnError func(err error) bool) error {
+	return RetryWithConfiguredTime(operation, retryOnError, DefaultMaxElapsedTime)
+}
+
+// RetryWithConfiguredTime will retry with an exponential back off until "maxRetryTime" worth of sleeping has occurred.
+// this allows users of this function to customize how long they intend to retry a function before giving up on the operation.
+func RetryWithConfiguredTime(operation func() error, retryOnError func(err error) bool, maxRetryTime time.Duration) error {
```
caglar10ur: This is stuttering (retry.RetryWithConfiguredTime) also you don't need a function, just set DefaultMaxElapsedTime to the value you desired before calling Do, that's why they are exported

caglar10ur: Hmm didn't realized that they are constants, I think we need functions so that we can set them other than default values

matthewavery: This may be a problem, the uploads are done concurrently. The backoff struct states that it is not thread safe, but the max interval var is read only afaik.

caglar10ur: Alternatively we could introduce retry.Config with default values and change that when we want to use different values

```
type Config struct {
        Interval ...
        RandomizationFactor 
        etc....
}

DefaultConfig {
       Interval: DefaultInitialInterval  , 
       RandomizationFactor: 	DefaultRandomizationFactor
       etc..
}
```

We introduce DoWithConfig (or something like that) which accepts the config struct as a parameter and Do calls it with DefaultConfig

matthewavery: Just saw that the `backoff` is a pointer. How do you advise that we do the concurrent uploads with this? We will likely have to do them back to back unless we make two different exponentialBackoffs. Do you recommend this @caglar10ur ?

caglar10ur: We are already creating two backoff-er as they run in their own goroutines tracking their upload, isn't that what we want?

matthewavery: yes that is what we want, I misread the package on first glance. This is good, not reusing a exponential backoff. I will look at doing the config thing tomorrow morning.

matthewavery:
I want to make a note that we could have allowed users to pass in their own backoff object instead of a config. However, since the exponential back off struct is not thread safe I went ahead and to that out of the hands of the caller(as it was before anyways). Every time `DoWithConfig` is called a new backoff is made even if the config is the same.

pkg/retry/retry.go
```diff
@@ -32,20 +32,43 @@ const (
...
+	MaxInterval         time.Duration
 
-	var err error
-	var next time.Duration
+	// this field will indicate the maximum amount of "sleep" time that will occur.
+	MaxElapsedTime time.Duration
+}
 
-	b := &backoff.ExponentialBackOff{
+// Do retries the given function until DefaultMaxInterval time passes, while sleeping some time between unsuccessful attempts
```
emlin: Do will use all default variables, no only DefaultMaxInterval. Anyway a minor issue

emlin:
btw, please paste one output of retried image upload 😄

lib/install/management/create.go
```diff
@@ -109,35 +117,97 @@ func (d *Dispatcher) startAppliance(conf *config.VirtualContainerHostConfigSpec)
...
 
-	var wg sync.WaitGroup
-
 	// upload the images
 	log.Infof("Uploading images for container")
 
-	wg.Add(len(files))
-	results := make(chan error, len(files))
+	results := make(chan error)
+	logMsgs := make(chan string)
```
hmahmood: Why do we need this channel? Why not just output the log messages?

lib/install/management/create.go
```diff
@@ -109,35 +117,97 @@ func (d *Dispatcher) startAppliance(conf *config.VirtualContainerHostConfigSpec)
...
 	defer trace.End(trace.Begin(""))
 
-	var wg sync.WaitGroup
-
 	// upload the images
 	log.Infof("Uploading images for container")
 
-	wg.Add(len(files))
-	results := make(chan error, len(files))
+	results := make(chan error)
```
hmahmood: Don't understand why you need this either. All you're doing with the errors is printing them to the log. Why not just print them to the log? To know when you're done you could just use a `WaitGroup` as the old code was using.

matthewavery:
Posting a copy of the retry logging here per @emlin 's request. 

```
May 30 2017 09:24:35.203-07:00 INFO  Uploading images for container
May 30 2017 09:24:35.210-07:00 INFO  	"bin/bootstrap.iso"
May 30 2017 09:24:35.211-07:00 INFO  	"bin/appliance.iso"
May 30 2017 09:24:37.087-07:00 WARN  Will try again in 7.142638726s. Operation failed with detected error
May 30 2017 09:24:37.087-07:00 INFO  failed an attempt to upload isos with err (Put https://blinky.eng.vmware.com/folder/uploadRetryTest/V1.1.0-RC3-0-F9C881A-bootstrap.iso?dcPath=%2Fstress-datacenter&dsName=kodos-local: write tcp 192.168.218.151:45696->10.17.109.9:443: write: connection reset by peer), 4 retries remain
May 30 2017 09:24:37.087-07:00 INFO  failed an attempt to upload isos with err (Put https://blinky.eng.vmware.com/folder/uploadRetryTest/V1.1.0-RC3-0-F9C881A-appliance.iso?dcPath=%2Fstress-datacenter&dsName=kodos-local: write tcp 192.168.218.151:45702->10.17.109.9:443: write: connection reset by peer), 4 retries remain
May 30 2017 09:24:37.087-07:00 WARN  Will try again in 8.806571893s. Operation failed with detected error
May 30 2017 09:24:44.294-07:00 WARN  Will try again in 1m10.745080555s. Operation failed with detected error
May 30 2017 09:24:44.294-07:00 INFO  failed an attempt to upload isos with err (Put https://blinky.eng.vmware.com/folder/uploadRetryTest/V1.1.0-RC3-0-F9C881A-bootstrap.iso?dcPath=%2Fstress-datacenter&dsName=kodos-local: dial tcp: lookup blinky.eng.vmware.com on 127.0.1.1:53: no such host), 3 retries remain
May 30 2017 09:24:45.928-07:00 WARN  Will try again in 1m21.749486247s. Operation failed with detected error
May 30 2017 09:24:45.929-07:00 INFO  failed an attempt to upload isos with err (Put https://blinky.eng.vmware.com/folder/uploadRetryTest/V1.1.0-RC3-0-F9C881A-appliance.iso?dcPath=%2Fstress-datacenter&dsName=kodos-local: dial tcp: lookup blinky.eng.vmware.com on 127.0.1.1:53: no such host), 3 retries remain
May 30 2017 09:26:10.546-07:00 INFO  Waiting for IP information
```

hmahmood:
@matthewavery can we get rid of the sub-second part of that time output? Looks really odd.

matthewavery:
Note: this PR is blocked until the upload bug is fixed. We plan to test the logger external from the code base to make sure there is no issue with the logger and the `waitgroup` implementation. if so, we will be moving back to the single channel approach with the wait group. Either way we are likely going to remove the log message channel and move back to logging.

matthewavery:
@hmahmood I can look into it. Looks like that comes from some vendored code though. I will truncate it when I take another look at this.

pkg/retry/retry.go
```diff
@@ -32,20 +32,43 @@ const (
...
 		Multiplier:          DefaultMultiplier,
 		MaxInterval:         DefaultMaxInterval,
 		MaxElapsedTime:      DefaultMaxElapsedTime,
+	}
+
+	return DoWithConfig(operation, retryOnError, bConf)
+}
+
+// DoWithConfig will attempt an operation while retrying using an exponential back off based on the config supplied by the caller. The retry decider is the supplied function retryOnError
+func DoWithConfig(operation func() error, retryOnError func(err error) bool, config *BackoffConfig) error {
```
matthewavery: talked to @caglar10ur and we have decided to make a `NewConfig` function which will populate the config object with all the defaults and let the user override them at will.

matthewavery:
We will also be attempting a delete and retrying on anything except a file not found error this will get us around the broken ISO bug that we have encountered for now and prevent us from leaving around junked up isos as well.

pkg/retry/retry.go
```diff
@@ -74,3 +97,11 @@ func Do(operation func() error, retryOnError func(err error) bool) error {
 		continue
 	}
 }
+
+// OnError is the simplest of retry deciders. If an error occurs it will indicate a retry is needed.
+func OnError(err error) bool {
+	if err != nil {
+		return true
+	}
+	return false
```
caglar10ur: return err != nil

pkg/retry/retry.go
```diff
@@ -32,20 +32,43 @@ const (
...
+	InitialInterval     time.Duration
+	RandomizationFactor float64
+	Multiplier          float64
+	MaxInterval         time.Duration
 
-	var err error
-	var next time.Duration
+	// this field will indicate the maximum amount of "sleep" time that will occur.
+	MaxElapsedTime time.Duration
+}
```
caglar10ur: As we discussed over slack, I'm guessing you are going to introduce a new type called DefaultBackoffConfig with default values (or a NewBackoffConfig function that returns a config struct with defaults) so that caller can override the fields

matthewavery:
sample output

matthewavery:
.

hmahmood:
@matthewavery so we didn't fix the extra accurate time output?

PR #5122, opened 2017-05-15
Filter out scratch in vsphere impl of ListImages [specific ci=Group1-Docker-Commands]

fdawg4l:
Remove a redundant lookup on scratch when restarting the VCH.


lib/portlayer/storage/vsphere/image.go
```diff
@@ -495,6 +495,11 @@ func (v *ImageStore) ListImages(op trace.Operation, store *url.URL, IDs []string
 
 		ID := file.Path
 
+		// filter out scratch
```
chengwang86: I don't know if this helps. The portlayer was looking for one of the layers of the busybox image (according to the original error msg), which I don't think is the scratch layer.

PR #5127, opened 2017-05-16
Fix --force flag help message

### LINKED ISSUE #5119: Update the text for the --force option in vic-machine CLI
The accompanying text for vic-machine current says "Force the install, removing existing if present" and this is inaccurate. The right text should be "Ignore error messages and proceed".

morris-jason:
Fixes #5119


cmd/vic-machine/create/create.go
```diff
@@ -485,7 +485,7 @@ func (c *Create) Flags() []cli.Flag {
 
 		cli.BoolFlag{
 			Name:        "force, f",
-			Usage:       "Force the install, removing existing if present",
```
matthewavery: I would also keep the clarification that it will install over an existing deployment in the error message 😄

sgairo: ^^^perhaps comment that on Karthik's issue above :)

matthewavery: Yeah, I just read this issue sorry I did not look at it ahead of time. I will leave the comment on karthiks issue. @karthik-narayan please take a look at this as well please 😄

matthewavery: @morris-jason so turns out I had a misunderstanding 😄  `--force` does not install over existing installments. Looks great!

andrewtchin: Maybe `Ignore error messages and proceed with install`

karthik-narayan:
LGTM and if we keep up this pace, I'm going to have to start writing new PRDs daily.

PR #5129, opened 2017-05-16
Continuing deleting other dangling VCHs even if one fails

mhagen-vmware:


rajanashok:
lgtm

sgairo:
Even if only one fails, list would still fail correct? Is this just for keeping the buildup smaller?

chengwang86:
lgtm

mhagen-vmware:
correct @sgairo It will still cause other problems for other tests, but this at least prevents the list from piling up.

PR #5019, opened 2017-05-04
Go back to two approvals

mhagen-vmware:
While we moved to a new pull approval process, we did not change our policy on number of approvals AFAIK.

PR #5130, opened 2017-05-16
Move to three approvers, since it was counting the author

mhagen-vmware:
This is a follow on from #5019 , because pull approve counts author as 1 it still was not enforcing 2 LGTMs as we have agreed to as a team.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -11,7 +11,7 @@ group_defaults:
 
 groups:
   contributors:
-    required: 2
+    required: 3
     teams:
       - vic-maintainers
 


mhagen-vmware:
@caglar10ur  This is in direct conflict with your desire here:  #5135  I would like an LGTM from you on this before I merge, or a better solution or just different solution that still enforces the our team's requirements.

caglar10ur:
Just to set the record straight neither #5135 nor need for component owners, how we review code or using pullapprove is a result of my personal desires.

As far as I concern, pullapprove (and component ownership) is useless because everyone gets added to every PR as the reviewer. 

If we are going to keep using pullapprove only for counting ACKs and gating merges but not for populating the reviewers list then please remove all that jazz from the yml file and communicate this decision with the team. 

This PR LGTM as long as our official stance is "everyone reviews everything" instead of "component owners need to review".

mhagen-vmware:
I understand your point, but I am not sure you are understanding mine..
This PR enforces our already agreed to policy that we need at least 2 LGTMs (from any team member) before merging. In January we added component owners, which is also a requirement which the rest of the yml enforces. We essentially want to enforce the policy the any pertinent component owners need to approve, but we also need to make sure that if only one component is touched then at least one other person approves of the PR also, which has not been happening.  There have been several PRs go in that only had 1 approval.

I am not committed to any specific policy in this regard, but until we address with the team and change our standing policy of 2 LGTMs we should at least enforce it.

caglar10ur:
I think I understand what this PR doing. My question/objection was towards the initial PR (https://github.com/vmware/vic/commit/6968ae07490dce72a46b528f2f690649c8863edf) where we added the contributor group to the yml file.

As far as I see, the broken part is not just email or web ui, the pullapprove integration itself. Let me give you an example;

if you send a PR that touches the tether code, the rules say we need 2+ ACKs and either George or Caglar needs to approve. However the contributors group now overrides the component owner requirement. So any other 3 people can give heads up to that PR and that PR can be merged without George or Caglar have a chance to review it. 

Am I missing something?

caglar10ur:
Hmmmm no just checked every open PR and AFAIS this adds contributors as an _additional group_ so integration is OK, UI and emails are not.

mhagen-vmware:
Yep, basically two scenarios:
1. You touch a bunch of components - you should still need approval from each owner and that will cover the general check that this implements of 2+
2. You touch only one component, worse case scenario your own component - this will at least prevent you from merging without 2 other LGTMs. (which has happened)

PR #5142, opened 2017-05-17
[specific ci=Group6-VIC-Machine]fix vic-machine delete backward compatibility issue

emlin:
this is to fix master build failure

### DIFF
--- a/cmd/vic-machine/inspect/inspect.go
+++ b/cmd/vic-machine/inspect/inspect.go
@@ -144,7 +144,7 @@ func (i *Inspect) Run(clic *cli.Context) (err error) {
 		return errors.New("inspect failed")
 	}
 
-	vchConfig, err := executor.GetVCHConfig(vch)
+	vchConfig, err := executor.GetNoSecretVCHConfig(vch)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


cmd/vic-machine/inspect/inspect.go
```diff
@@ -144,7 +144,7 @@ func (i *Inspect) Run(clic *cli.Context) (err error) {
 		return errors.New("inspect failed")
 	}
 
-	vchConfig, err := executor.GetVCHConfig(vch)
+	vchConfig, err := executor.GetNoSecretVCHConfig(vch)
```
vburenin: Thinking about this, I would have one method to get VCH config where the second option would be a secret. If the secret is empty, I would treat it as no secret. Just saying.

emlin: the secret is query inside of method, instead of provided by the caller. for backward compatibility reason, if caller does not need the secret, we might leave it there.
so it's hard to pass in empty secret in this layer, but it's worth to make it better. will think about that in next PR, but not for this one. I prefer a quick fix for the problem.

PR #5170, opened 2017-05-18
Fix Check ImageStore util

jzt:
Fixes an incorrect check in the `Check ImageStore` test util that causes nightly failures.


mhagen-vmware:
I tested this, just need one more LGTM

rajanashok:
lgtm

PR #5175, opened 2017-05-18
[skip ci] Update NSX test to install into an NSX logical switch

mhagen-vmware:
I test this manually and it appears to actually work, so this should be fine.  Because of the current state of our NSX test, I had to manually log in and create the logical switch though.


rajanashok:
lgtm

PR #5183, opened 2017-05-18
Up parallel builds to match our larger build servers now

mhagen-vmware:
Tagging @caglar10ur and other pertinent people on this PR.  We want to make a strong push in the next week to try and get our build times down to under 2 hours.  This is hopefully a first step along that path.


.drone.yml
```diff
@@ -44,7 +44,7 @@ pipeline:
       COMMIT: ${DRONE_COMMIT}
     commands:
       - make mark
-      - make -j3 all
```
hmahmood: Also for the `make -j3 test` line below?

mhagen-vmware:
We are no longer doing this, turns out the latest golang has improved their parallel support so we don't need to do this anymore and actually it causes problems with build dependency concurrency.

Issue #5187, opened 2017-05-19
Support for schema 2 image manifests


anchal-agrawal:
**User Statement:**

As a user of VIC engine, I should be able to use images with [schema 2](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md) manifests.

**Details:**

We currently rely on [schema 1](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-1.md) fields to obtain image layers and fetch them. Most mainstream images still support schema 1, which is why this issue is not widely evident. For schema 2 manifests, the registry [offers backward compatibility](https://github.com/docker/distribution/blob/master/docs/spec/manifest-v2-2.md#backward-compatibility) with schema 1: 

> If the manifest being requested uses the new format, and the appropriate media type is not present in an Accept header, the registry will assume that the client cannot handle the manifest as-is, and rewrite it on the fly into the old format.

However, while working on supporting pull-by-digest (#1186), I found that some images when pulled by digest do not offer schema 2 manifests even with the manifest request headers set appropriately. I've seen this issue with two images so far:

* `nginx@sha256:ebd2b85803e78100a582385f7eac56cd367561f0f2bab005f784cda95818d41f`
* `busybox@sha256:c79345819a6882c31b41bc771d9a94fc52872fa651b36771fbe0c8461d7ee558` (digest of `busybox:latest`)

This issue will be masked till #1186 is closed because we pull by the `latest` tag when pull-by-digest is attempted, and I haven't yet found an image that, when pulled by name/tag, offers only a schema 2 manifest. **With the current scope of #1186, we'll be able to pull images that offer schema 1 when pulled by digest. If an image when pulled by name/tag/digest offers only schema 2, the pull operation will fail.**

 We currently fetch the schema 2 manifest, but only use it to calculate the digest shasum. Estimated as a 5 because the work involves parsing the schema 2 manifest and fetching/building the image. Also, in the schema 2 manifest, layers are ordered in the opposite order than schema 1's manifest.

**Acceptance criteria:**

* [ ] Imagec supports schema 2 manifests
* [ ] Integration tests for schema 2 images

stuclem:
A pitiful attempt at a release note:

-----

- <b>Schema 2 image manifests not supported. [#5187](https://github.com/vmware/vic/issues/5187)</b><br>vSphere Integrated Containers Engine does not support pulling images that only have schema 2 image manifests.

-----

[...]

PR #5194, opened 2017-05-19
[specific ci=3-04-Docker-Compose-Fixtures] Add pull-by-digest support for schema 1 images

### LINKED ISSUE #1186: Support pull by image digest
Currently imagec is only able to pull images by tag.  For example:

```
docker pull busybox:1.24.2
```

We need to provide support for pulling images via image digest:

```
docker pull busybox@sha256:4a731fb46adc5cefe3ae374a8b6020fc1b6ad667a279647766e9a3cd89f6fa92
```

While vic (via imagec) will attempt a pull when a user provides a digest, it's not pulling via the digest -- it's actually pulling via the latest tag.  Since the digest is ignored the wrong image might be pulled.

Depends on #1484

anchal-agrawal:
Fixes #1186

Follow-up issue for schema 2: #5187

### DIFF
--- a/lib/imagec/docker.go
+++ b/lib/imagec/docker.go
@@ -325,7 +325,8 @@ func FetchImageBlob(ctx context.Context, options Options, image *ImageWithMeta, 
 	return diffID, nil
 }
 
-// tagOrDigest returns an image's tag or digest based on the image reference.
+// tagOrDigest returns an image's digest if it's pulled by digest, or its tag
+// otherwise.
 func tagOrDigest(r reference.Named, tag string) string {
 	if digested, ok := r.(reference.Canonical); ok {
 		return digested.Digest().String()
@@ -406,16 +407,6 @@ func decodeManifestSchema1(filename string, options Options, registry string) (i
 		return nil, "", err
 	}
 
-	// Verify schema 1 manifest's fields per
-	// docker/docker/distribution/pull_v2.go:verifySchema1Manifest()
-	numFSLayers := len(manifest.FSLayers)
-	if numFSLayers == 0 {
-		return nil, "", fmt.Errorf("no FSlayers in manifest")
-	}
-	if numFSLayers != len(manifest.History) {
-		return nil, "", fmt.Errorf("length of history not equal to number of layers")
-	}
-
 	digest, err := getManifestDigest(content, options.Reference)
 	if err != nil {
 		return nil, "", err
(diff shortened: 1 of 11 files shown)


lib/imagec/docker.go
```diff
@@ -335,8 +349,12 @@ func FetchImageManifest(ctx context.Context, options Options, schemaVersion int,
 	if err != nil {
 		return nil, "", err
 	}
-	url.Path = path.Join(url.Path, options.Image, "manifests", options.Tag)
 
+	manifestRef := options.Tag
+	if digested, ok := options.Reference.(reference.Canonical); ok {
```
jzt: Seems a little D.R.Y... Any reason why you can't just call `manifestRef := tagOrDigest(options.Reference, options.Tag)`?

anchal-agrawal: Oh, I forgot to change that. Good catch!

cgtexmex:
LGTM

I'll echo @jzt -- Nice work!

lib/imagec/docker.go
```diff
@@ -455,6 +481,15 @@ func getManifestDigest(content []byte) (string, error) {
 	}
 
 	log.Debugf("Canonical Bytes: %d", len(bytes))
+
+	// Verify the manifest digest if the image is pulled by digest.
+	// https://docs.docker.com/registry/spec/api/#content-digests
+	if digested, ok := ref.(reference.Canonical); ok {
```
fdawg4l: Can you add a comment why line `493` is ok when this type assertion fails?

lib/imagec/docker.go
```diff
@@ -416,6 +426,22 @@ func decodeManifestSchema1(filename string, options Options, registry string) (i
...
+	if !verifier.Verified() {
+		return fmt.Errorf("image manifest verification failed for digest %s", digested.Digest())
+	}
+
+	return nil
+}
+
 // decodeManifestSchema2() reads a manifest schema 2 and creates a Docker
 // defined Manifest structure and returns the digest of the manifest as a string.
 func decodeManifestSchema2(filename string, options Options) (interface{}, string, error) {
```
fdawg4l: It's not strictly part of this change, but this feels like it should be part of an `interface` where schema1 and schema2 are implementations.

lib/imagec/docker.go
fdawg4l: Also is there a test that exercises the !ok path?

lib/imagec/imagec.go
```diff
@@ -158,29 +155,21 @@ func init() {
...
-	}
-
-	ic.Tag = reference.DefaultTag
-	if !reference.IsNameOnly(ref) {
-		if tagged, ok := ref.(reference.NamedTagged); ok {
+func (ic *ImageC) ParseReference() {
+	if reference.IsNameOnly(ic.Reference) {
+		ic.Tag = reference.DefaultTag
+	} else {
+		if tagged, isTagged := ic.Reference.(reference.NamedTagged); isTagged {
```
fdawg4l: Minor. I think the idiom is to use `ok` and test for `!ok`.

anchal-agrawal: I see. Here, we want to set `ic.Tag` if the type-assert is successful, so we may want to keep this as-is.

lib/imagec/docker.go
anchal-agrawal: Done and done! Please see d226c7c.

PR #5203, opened 2017-05-19
[specific ci=1-01-Docker-Info]  Start to pull from our new harbor instance

mhagen-vmware:
Dynamic variable library so that we can continue to support local as well as CI builds, switch regression and 1st test to using the new dynamic vars.


tests/resources/VCH-Util.robot
```diff
@@ -174,11 +174,11 @@ Install VIC Appliance To Test Server
...
     Return From Keyword If  ${certs}  ${output}
 
-    ${output}=  Run Keyword Unless  ${certs}  Run  ${vic-machine} create --debug 1 --name=%{VCH-NAME} --target=%{TEST_URL}%{TEST_DATACENTER} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --image-store=%{TEST_DATASTORE} --appliance-iso=${appliance-iso} --bootstrap-iso=${bootstrap-iso} --password=%{TEST_PASSWORD} --force=true --bridge-network=%{BRIDGE_NETWORK} --public-network=%{PUBLIC_NETWORK} --compute-resource=%{TEST_RESOURCE} --timeout %{TEST_TIMEOUT} --volume-store=%{TEST_DATASTORE}/test:${vol} --no-tlsverify
+    ${output}=  Run Keyword Unless  ${certs}  Run  ${vic-machine} create --debug 1 --name=%{VCH-NAME} --target=%{TEST_URL}%{TEST_DATACENTER} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --image-store=%{TEST_DATASTORE} --appliance-iso=${appliance-iso} --bootstrap-iso=${bootstrap-iso} --password=%{TEST_PASSWORD} --force=true --bridge-network=%{BRIDGE_NETWORK} --public-network=%{PUBLIC_NETWORK} --compute-resource=%{TEST_RESOURCE} --timeout %{TEST_TIMEOUT} --insecure-registry harbor.ci.drone.local --volume-store=%{TEST_DATASTORE}/test:${vol} --no-tlsverify
```
andrewtchin: Do these need an if IS_LOCAL?

mhagen-vmware: Nope, abstracted into the variables file.

mhagen-vmware: Should be late binding polymorphism so that as long as you have the env var set prior to using the car it will work.

mhagen-vmware: Ah, I get you... Still shouldn't need it because I don't believe vic-machine does any check until you actually use it so in the case of local testing it will just become an unused insecure registry.

mhagen-vmware: @andrewtchin you good with this?

PR #5213, opened 2017-05-22
Bump govmomi to latest

jzt:
Bumps govmomi to latest version.


cgtexmex:
Do we need to update vagrant & CI image?

PR #5215, opened 2017-05-23
Check last Nak in handshake

### LINKED ISSUE #5015: Exec echo test failure: Server error from portlayer: &{0 } (*models.Error) is not supported by the ByteStreamConsumer
```Server error from portlayer: &{0 } (*models.Error) is not supported by the ByteStreamConsumer, can be resolved by supporting Writer/BinaryUnmarshaler interface != Help me, Obi-Wan Kenobi. You're my only hope.```

[integration_logs_10275_3bf485a6d108954f125adbadd01911b85f4fda43.zip](https://github.com/vmware/vic/files/976222/integration_logs_10275_3bf485a6d108954f125adbadd01911b85f4fda43.zip)

from personality:
```
May  3 2017 21:04:49.974Z ERROR container attach: stdout (2278f5bc074baa258e28bb3d1d3a637a82920bc3428050727bf930b20bfd8d1b): Server error from portlayer: &{0 } (*models.Error) is not supported by the ByteStreamConsumer, can be resolved by supporting Writer/BinaryUnmarshaler interface
May  3 2017 21:04:49.974Z ERROR container attach: stderr (2278f5bc074baa258e28bb3d1d3a637a82920bc3428050727bf930b20bfd8d1b): Server error from portlayer: &{0 } (*models.Error) is not supported by the ByteStreamConsumer, can be resolved by supporting Writer/BinaryUnmarshaler interface
May  3 2017 21:04:49.974Z INFO  cleaned up connections to 2278f5bc074baa258e28bb3d1d3a637a82920bc3428050727bf930b20bfd8d1b. Checking errors
May  3 2017 21:04:49.974Z ERROR container attach error: Server error from portlayer: &{0 } (*models.Error) is not supported by the ByteStreamConsumer, can be resolved by supporting Writer/BinaryUnmarshaler interface
```

weirdly, the portlayer seems to report success...
```
May  3 2017 21:04:50.010Z DEBUG op=259.30 (delta:28.377µs): ID: "2278f5bc074baa258e28bb3d1d3a637a82920bc3428050727bf930b20bfd8d1b"
May  3 2017 21:04:50.010Z DEBUG op=259.30 (delta:33.935µs): Path: "/bin/echo"
May  3 2017 21:04:50.010Z DEBUG op=259.30 (delta:45.111µs): Args: []string{"/bin/echo", "Help me, Obi-Wan Kenobi. You're my only hope."}
May  3 2017 21:04:50.010Z DEBUG op=259.30 (delta:51.564µs): Running: 0
May  3 2017 21:04:50.010Z DEBUG op=259.30 (delta:55.274µs): ExitCode: 0
```

### PR DESCRIPTION
**Pull request by caglar10ur**
Client now checks the last Nak before declaring success. Otherwise server might
reject that connection attempt but client never gets it.

Also removes debugFlag, PurgeIncoming and related code.

Fixes #5015

Fixes #

### DIFF
--- a/pkg/serial/handshake.go
+++ b/pkg/serial/handshake.go
@@ -11,15 +11,34 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
+//
+//
+//	Client:
+//		generate a random uint8 (#)
+//		send 2 bytes Syn|#
+//
+//	Server:
+//		generate a random uint8 (&)
+//		read at least 2 bytes and make sure Syn|# is received
+//		send 3 bytes Ack|#+1|& (or Nak)
+//
+//  Client:
+//	    read at least 3 bytes and make sure Ack|#+1|& is received
+//		send 2 bytes Ack|&+1
+//
+//	Server:
+//		read at least 2 bytes and make sure Ack|&+1 is received
+//		send 1 byte Ack (or Nak)
+//	Client:
+//		read at leat 1 byte and make sure Ack is received
 
 package serial
 
 import (
-	"bytes"
-	"crypto/rand"
 	"fmt"
 	"io"
-	"net"
+	"math"
+	"math/rand"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -28,10 +47,9 @@ import (
 )
 
 const (
-	flagSyn      byte = 0x16
-	flagAck      byte = 0x06
-	flagDebugAck byte = 0x07
-	flagNak      byte = 0x15
+	flagSyn byte = 0x16
+	flagAck      = 0x06
+	flagNak      = 0x15
 )
 
 // HandshakeError should only occure if the protocol between HandshakeServer and HandshakeClient was violated.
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: pkg/serial/handshake.go [hunk 1]
```diff
@@ -28,134 +47,97 @@ import (
...
-	buf := make([]byte, 255)
 
-	// read until the incoming channel is empty
-	conn.SetReadDeadline(time.Now().Add(time.Duration(10 * time.Millisecond)))
-	for n, err := conn.Read(buf); n != 0 || err == nil; n, err = conn.Read(buf) {
-		log.Debugf("discarding following %d bytes from input channel\n", n)
-		log.Debugf("%+v\n", buf[0:n])
+	// however if we read more than l, it means buffer is not empty
+	if n != l {
+		buffer = buffer[n-l:]
```
hmahmood: Why is it OK here to throw away the first `n-l` bytes?

caglar10ur: We get a lots of garbage data when we get the initial connection this handshake supposed to clear them and leave the connection in a known state so that the real ssh handshake can start. Also the client and server is looping with different frequencies (their read timeouts are 10ms vs 100ms I believe) so client could send multiple requests because of different timeouts. By getting the last l bytes we are saying that we are not interested with garbage data and also eliminating duplicated flags by only using the last one

caglar10ur: and I think I should write this as a comment into that function :)

hmahmood: Hm... ok. So do we know if bytes discarded here are garbage for sure?

caglar10ur: I think it doesn't matter whether it is real garbage or not. Client and server will get in sync eventually even if we drop legitimate handshake packages as garbage or read in the middle of a write etc. This is only for our handshake so after that SSH will take over and will do its own thing so there is nothing valuable to lose in that stage.

PR #5218, opened 2017-05-23
[specific ci=Group6-VIC-Machine]convert vch configuration back to vic-machine data structure

### LINKED ISSUE #5082: vic-machine: convert guestinfo back to create data structure

emlin:
fixes #5082 

With this mechanism, vic-machine configure can combine user input and old configuration and then redo validation again.
This change does not add any input for vic-machine configure command, so still leave the command hidden in vic-machine. The command can be executed now, but no configuration will be changed.

vic-machine inspect is updated to get create command options from data translated from VirtualContainerHostConfigSpec


cmd/vic-machine/converter/converter.go
```diff
@@ -0,0 +1,505 @@
...
+		}
+		if err := convert(src.MapIndex(pkey), prefix, tags, tempMap); err != nil {
+			return err
+		}
+		if err := handler(tempMap, pkey.String()); err != nil {
+			return err
+		}
+		for k, v := range tempMap {
+			addValues(dest, k, v)
+			delete(tempMap, k)
```
vburenin: This may be dangerous to do during iteration in the future. Deleting a value from the map may shrink the map which may change the order of the map key. It is not the case yet, but there are proposals already to shrink the map when elements are deleted. Since you really need just an empty map I would just move its init inside the loop.

emlin: hmm... for this one, it is. But I do have another case, need to replace the key in the map at https://github.com/vmware/vic/pull/5218/files/dfbf0e9799247d373e0b39d8c9b12d0bff667d24#diff-f1c87f067c3afedcd7836400c915a2b5R467

cmd/vic-machine/converter/converter.go
```diff
@@ -0,0 +1,505 @@
...
+	defer trace.End(trace.Begin(fmt.Sprintf("prefix: %s, src: %s", prefix, src.String())))
+
+	if prefix == "" {
+		return nil
+	}
+	if tags.Get(cmdTag) == "" {
+		return nil
+	}
+
+	ipAddr, ok := src.Interface().(net.IP)
```
vburenin: Taking into account you are using map of types. I would not be so tolerant politely returning error, I would panic here, since this situations is simply not possible.

cmd/vic-machine/converter/converter.go
```diff
@@ -0,0 +1,506 @@
...
+			networks.MappedNetworksIPRanges[v] = ipRange
+			delete(networks.MappedNetworksIPRanges, k)
+		}
+		delete(networks.MappedNetworks, k)
+		networks.MappedNetworks[v] = k
+	}
+	return convertStruct(reflect.ValueOf(networks), prefix, tags, dest)
+}
+
+// addValue will apend value without duplicates
```
anchal-agrawal: s/apend/append

lib/install/validate/config_to_data.go
```diff
@@ -0,0 +1,314 @@
...
+		return "", err
+	}
+	hsize := units.HumanSize(float64(is))
+	hsize = strings.Replace(hsize, " ", "", -1)
+	return hsize, nil
+}
+
+func setImageStore(d *data.Data, conf *config.VirtualContainerHostConfigSpec) error {
+	if len(conf.ImageStores) != 1 {
+		return fmt.Errorf("no image store or multiple image store configured")
```
chengwang86: maybe we can split this into two errors as `no image store configured` and `multiple image stores configured` for better debugging?

lib/install/validate/config_to_data.go
```diff
@@ -0,0 +1,314 @@
...
+	return nil
+}
+
+func setVolumeLocations(d *data.Data, conf *config.VirtualContainerHostConfigSpec) {
+	d.VolumeLocations = make(map[string]*url.URL, len(conf.VolumeLocations))
+
+	var dsURL object.DatastorePath
+	for k, v := range conf.VolumeLocations {
+		if ok := dsURL.FromString(v.Path); !ok {
+			log.Debugf("%s is not datastore path", v.Path)
```
chengwang86: Why isn't it an error?

tests/test-cases/Group6-VIC-Machine/6-16-Config.robot
```diff
@@ -0,0 +1,35 @@
...
+Resource  ../../resources/Util.robot
+Test Teardown  Run Keyword If Test Failed  Cleanup VIC Appliance On Test Server
+
+*** Test Cases ***
+Configure VCH
+    Set Test Environment Variables
+    Run Keyword And Ignore Error  Cleanup Dangling VMs On Test Server
+    Run Keyword And Ignore Error  Cleanup Datastore On Test Server
+
+    ${output}=  Run  bin/vic-machine-linux create --name=%{VCH-NAME} --target=%{TEST_URL} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --image-store=%{TEST_DATASTORE} --appliance-iso=bin/appliance.iso --bootstrap-iso=bin/bootstrap.iso --password=%{TEST_PASSWORD} --force=true --bridge-network=%{BRIDGE_NETWORK} --public-network=%{PUBLIC_NETWORK} --compute-resource=%{TEST_RESOURCE} --timeout %{TEST_TIMEOUT} --volume-store=%{TEST_DATASTORE}/test:default ${vicmachinetls}
```
chengwang86: This looks the same with the `vic-machine create` cmd` in `Suite setup  Install VIC Appliance To Test Server`.  Why don't we simply use that keyword?

cmd/vic-machine/converter/converter.go
```diff
@@ -0,0 +1,506 @@
...
+		return nil
+	}
+
+	for k, v := range networks.MappedNetworks {
+		if k == v {
+			continue
+		}
+		if dns, ok := networks.MappedNetworksDNS[k]; ok {
+			networks.MappedNetworksDNS[v] = dns
+			delete(networks.MappedNetworksDNS, k)
```
hmahmood: Why are we modifying the input here? Generally for code like this, the input is not modified. If any modifications are needed, then operate on a copy.

lib/install/validate/config_to_data.go
emlin: nfs volume store is not datastore path, so in that case, we set the datastore path directly

tests/test-cases/Group6-VIC-Machine/6-16-Config.robot
emlin: really? let me check

emlin: yeah, will update it

cmd/vic-machine/converter/converter.go
emlin: found that golang iterator supports delete in the loop, so it can save the data copy logic between two maps.

emlin: actually, it just saved one string array to store the keys need to be deleted in the map

tests/test-cases/Group6-VIC-Machine/6-16-Config.robot
```diff
@@ -0,0 +1,29 @@
...
+
+*** Test Cases ***
+Configure VCH
+    Run Regression Tests
+
+    ${output}=  Run  bin/vic-machine-linux configure --name=%{VCH-NAME} --target=%{TEST_URL} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --password=%{TEST_PASSWORD} --timeout %{TEST_TIMEOUT}
+    Should Contain  ${output}  Completed successfully
+
+    Run Regression Tests
+    Cleanup VIC Appliance On Test Server
```
chengwang86: Is this dupe of line 19?

chengwang86:
lgtm  👍

cmd/vic-machine/configure/configure.go
```diff
@@ -104,6 +104,13 @@ func (c *Configure) processParams() error {
 	return nil
 }
 
+// copyChangedConf should copy all configuration change specified by user to old configuration
+// currently we cannot automatically override old configuration with any difference in the new configuration, because there are some configuration set in VCH
+// creation process, for example, image store path, volume store path, network slot id, etc. So we'll copy changes based on user input
+func (c *Configure) copyChangedConf(o *config.VirtualContainerHostConfigSpec, n *config.VirtualContainerHostConfigSpec) {
+	//TODO: copy changed data
```
anchal-agrawal: Is there an open issue for this TODO? If so, please mention the issue number here for reference.

cmd/vic-machine/configure/configure.go
```diff
@@ -185,20 +192,22 @@ func (c *Configure) Run(clic *cli.Context) (err error) {
 		return errors.New("configure failed")
 	}
 
-	// TODO: Convert guestinfo *VirtualContainerHost back to *Data, decrypt secret data
-	oldData := &data.Data{}
-	// oldData, err := converter.GtoD(vchConfig)
+	// Convert guestinfo *VirtualContainerHost back to *Data, decrypt secret data
+	oldData, err := validate.NewDataFromConfig(ctx, validator.Session.Finder, vchConfig)
```
anchal-agrawal: If we're ignoring `err` here, it would be helpful to add a comment why.

cmd/vic-machine/configure/configure.go
emlin: hmm.. this should be for all vic-machine reconfigure issues. If it's expected, it can be the epic issue #2911

cmd/vic-machine/configure/configure.go
emlin: should stop configure here. good catch

cmd/vic-machine/converter/converter.go
```diff
@@ -0,0 +1,508 @@
...
+
+	network, ok := src.Interface().(data.NetworkConfig)
+	if !ok {
+		panic(fmt.Sprintf(src.Type().String() + " is not NetworkConfig"))
+	}
+	if !network.IsSet() {
+		return nil
+	}
+
+	if len(network.Destinations) > 0 || !ip.Empty(network.Gateway) {
```
anchal-agrawal: It would be nice to add a comment explaining this `if` block since this code is a bit involved.

cmd/vic-machine/converter/converter_test.go
```diff
@@ -0,0 +1,611 @@
...
+		return "bridge", nil
+	case "DistributedVirtualPortgroup:dvportgroup-56":
+		return "external", nil
+	default:
+		return "unknown", nil
+	}
+}
+
+func TestMain(t *testing.T) {
+	//	ConverterLogLevel = log.DebugLevel
```
anchal-agrawal: Unneeded comments?

lib/install/validate/config_to_data.go
```diff
@@ -0,0 +1,317 @@
...
+
+	// TODO: set VCH resource limits and VCH endpoint VM resource limits
+	return nil
+}
+
+// NewDataFromConfig converts VirtualContainerHostConfigSpec back to data.Data object
+// This method does not touch any configuration for VCH VM or resource pool, which should be retrieved from VM or vApp attributes
+func NewDataFromConfig(ctx context.Context, finder Finder, conf *config.VirtualContainerHostConfigSpec) (d *data.Data, err error) {
+	if conf == nil {
+		return
```
anchal-agrawal: Should we set `err` before returning here?

lib/install/validate/config_to_data.go
```diff
@@ -0,0 +1,317 @@
...
+	hsize = strings.Replace(hsize, " ", "", -1)
+	return hsize, nil
+}
+
+func setImageStore(d *data.Data, conf *config.VirtualContainerHostConfigSpec) error {
+	if len(conf.ImageStores) == 0 {
+		return fmt.Errorf("no image store configured")
+	}
+	if len(conf.ImageStores) > 1 {
+		return fmt.Errorf(fmt.Sprintf("%d image store configured", len(conf.ImageStores)))
```
anchal-agrawal: s/store/stores

cmd/vic-machine/converter/converter.go
emlin: hmm.. thought to add something more here. But this method already has comments for what is did. 
This if block is simply convert the two elements back to one option. which kind of comment you think is good here?

cmd/vic-machine/converter/converter_test.go
emlin: Because the converter log is disabled by default, and not set through general log level, this comments is used to show samples. 😄

cmd/vic-machine/converter/converter.go
anchal-agrawal: I see - that makes sense. We can leave it as-is. 😃

cmd/vic-machine/converter/converter.go
```diff
@@ -0,0 +1,508 @@
...
+
+	labelHandlers[keyAfterValueLabel] = keyAfterValueLabelHandler
+	labelHandlers[valueAfterKeyLabel] = valueAfterKeyLabelHandler
+}
+
+// DataToOption convert data.Data structure to vic-machine create command options based on tags defined in data.Data structure
+// Note: need to make sure the tags are consistent with command line option name
+func DataToOption(data *data.Data) (map[string][]string, error) {
+	defer log.SetLevel(log.GetLevel())
+	log.SetLevel(ConverterLogLevel)
```
hmahmood: This will set the global log level, not just here. If you want to use different log levels for this package, then use package level logger, no the global one.

cmd/vic-machine/configure/configure.go
anchal-agrawal: I see. We can leave it as-is then.

tests/test-cases/Group6-VIC-Machine/6-16-Config.md
```diff
@@ -0,0 +1,20 @@
...
+
+# Purpose:
+Verify vic-machine configure
+
+# References:
+* vic-machine-linux create -h
+
+# Environment:
+This test requires that a vSphere server is running and available
+This test requires that a syslog server is running and available
```
mhagen-vmware: don't need syslog server

tests/test-cases/Group6-VIC-Machine/6-16-Config.md
```diff
@@ -0,0 +1,20 @@
...
+
+# Test Steps
+1. Deploy VCH
+2. Configure VCH
+3. VCH still function well
```
mhagen-vmware: Run a variety of docker commands

lib/install/validate/config_to_data.go
```diff
@@ -0,0 +1,318 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
```
mhagen-vmware: This is a large new file, but it doesn't look like you added any unit tests?  Can you please add at least some unit tests for this new file?

emlin: this file convert VirtualContainerHostConfigSpec to Data structure, and then another file converter.go will convert Data structure back to command option, these two functions are covered by test cases in converter_test.go, so I didn't write any test case for this file, because the verification of string map is a lot more straight forward than a structure.

PR #5219, opened 2017-05-23
[specific ci=1-11-Docker-RM] Enable Destroy_Task before calling vsphere API to remove the containerVM

### LINKED ISSUE #5198: containerRemoveInternalServerError in 5-4-High-Availability nightly test

**VIC version:**

Build 10681

**Deployment details:**

Run 5-4-High-Availability test manually. The error occurs at this step https://github.com/vmware/vic/blob/master/tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot#L201

**Actual behavior:**
```
docker -H 10.160.72.57:2375 ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS                     PORTS               NAMES
db8038f571f0        busybox             "ls"                5 minutes ago       Exited (0) 5 minutes ago                       tender_poitras
b8c0f179726e        busybox             "ls"                6 minutes ago       Exited (0) 6 minutes ago                       priceless_borg
2fd02be27c66        busybox             "ls"                7 minutes ago       Exited (0) 6 minutes ago                       gifted_meitner
e28263285d83        busybox             "sh"                8 minutes ago       Up 8 minutes                                   sad_visvesvaraya
f0270ca3b265        busybox             "sh"                8 minutes ago       Up 8 minutes                                   thirsty_shockley
7fce212a0a3f        busybox             "sh"                9 minutes ago       Up 9 minutes                                   jolly_kalam

docker -H 10.160.72.57:2375 rm -f 7f
Error response from daemon: Server error from portlayer: [DELETE /containers/{id}][500] containerRemoveInternalServerError
```

**Expected behavior:**
`docker rm -f` should finish successfully.

**Logs.**
The original logs for the operations when this error occurred cannot be found in our nightly test since we restarted the VCH. However, I was able to reproduce the error by running the test manually. See the attached log bundle.
[container-logs (1).zip](https://github.com/vmware/vic/files/1015396/container-logs.1.zip)

**Note:** 
This error does not appear in the same test under build 9992 and the builds prior to 9992. It always fails starting from build 10634. There is no nightly test using builds between 9992 and 10634. 

**Logs:**

chengwang86:
Fixes #5198 

In nightly test 5-4-High-Availability, the VCH throws an error of `containerRemoveInteranlServerError` when performing `docker rm -f container` after HA event.

We find that the error is reproducible ever since build 10290 which relates to PR #4965 (disable VM delete action from VC UI).

In this PR, I enable `Destroy_Task` right before calling vpshere API for container removal and it seems that the issue can be fixed by doing so.

However, we still do not understand why disabling `Destroy_Task` would affect `docker rm -f container` after HA, since disabling `Destroy_Task` is suppose to only disable/grey the VM deletion action from VC UI but not to block calling vsphere API.

### DIFF
--- a/pkg/vsphere/vm/vm.go
+++ b/pkg/vsphere/vm/vm.go
@@ -618,7 +618,7 @@ func (vm *VirtualMachine) DisableDestroy(ctx context.Context) {
 	}
 }
 
-// EnableDestroy attempts to enable the VitualMachine.Destroy_Task method
+// EnableDestroy attempts to enable the VirtualMachine.Destroy_Task method
 // so that the VC UI can enable the VM "delete" action.
 func (vm *VirtualMachine) EnableDestroy(ctx context.Context) {
 	if !vm.IsVC() {
--- a/pkg/vsphere/vm/vm.go
+++ b/pkg/vsphere/vm/vm.go
@@ -38,7 +38,10 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/tasks"
 )
 
-const UpdateStatus = "UpdateInProgress"
+const (
+	Destroy_Task = "Destroy_Task"
+	UpdateStatus = "UpdateInProgress"
+)
 
 type InvalidState struct {
 	r types.ManagedObjectReference
@@ -602,7 +605,7 @@ func (vm *VirtualMachine) DisableDestroy(ctx context.Context) {
 
 	method := []object.DisabledMethodRequest{
 		{
-			Method: "Destroy_Task",
+			Method: Destroy_Task,
 			Reason: "Managed by VIC Engine",
 		},
 	}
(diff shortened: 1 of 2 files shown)


dougm:
@chengwang86 I looked into it, not related to HA but restart of VCH.  When VCH is restarted, a new vSphere session is created.  Turns out, only the session that disabled the method can invoke it.

For example, disable with `-persist-session=false` so different sessions a different session is used:

```
govc object.method -enable=false -name Destroy_Task -persist-session=false /vcqaDC/vm/foo: finished.
govc vm.destroy /vcqaDC/vm/foo
govc: ServerFaultCode: The method is disabled by 'govc'
```

chengwang86:
@dougm Thanks for explanation. Do you happen to know why it is designed like this? Is it because the `AuthorizationManager` has `global` privilege? 

`When VCH is restarted, a new vSphere session is created. Turns out, only the session that disabled the method can invoke it.` But disabling `Destroy_Task` should only affect VC UI, right?

dougm:
@chengwang86 not related to the privilege.  I think it makes sense, I had just assumed it was a UI-only thing since my Disable+Destroy testing had been done using the same session.

pkg/vsphere/vm/vm.go
```diff
@@ -611,6 +614,23 @@ func (vm *VirtualMachine) DisableDestroy(ctx context.Context) {
 
 	err := m.DisableMethods(ctx, obj, method, "VIC")
 	if err != nil {
-		log.Warnf("failed to disable method %s for %s: %s", method[0].Method, obj[0], err)
+		log.Warnf("Failed to disable method %s for %s: %s", method[0].Method, obj[0], err)
+	}
+}
+
+// EnableDestroy attempts to enable the VitualMachine.Destroy_Task method
```
caglar10ur: VitualMachine -> VirtualMachine

pkg/vsphere/vm/vm.go
```diff
@@ -611,6 +614,23 @@ func (vm *VirtualMachine) DisableDestroy(ctx context.Context) {
...
 	err := m.DisableMethods(ctx, obj, method, "VIC")
 	if err != nil {
-		log.Warnf("failed to disable method %s for %s: %s", method[0].Method, obj[0], err)
+		log.Warnf("Failed to disable method %s for %s: %s", method[0].Method, obj[0], err)
+	}
+}
+
+// EnableDestroy attempts to enable the VirtualMachine.Destroy_Task method
+// so that the VC UI can enable the VM "delete" action.
+func (vm *VirtualMachine) EnableDestroy(ctx context.Context) {
```
anchal-agrawal: Minor suggestion: it would be helpful to have this func return an error so the caller can decide what to do in that case. This may not be super relevant for this change, but it would be for extensibility.

chengwang86: Let's keep it like this for now. I agree with your point. Here I just try to be consistent with the way the other similar method `DisableDestroy()` deals with its own error.

PR #5227, opened 2017-05-23
Trigger the downstream build

mhagen-vmware:


.drone.yml
```diff
@@ -176,6 +176,14 @@ pipeline:
       branch: [ refs/tags/* ]
       status: success
 
+  notify:
+    downstream:
+      repositories:
+        -vmware/vic-product
```
andrewtchin: does this need to be `- vmware/vic-product`? added space

.drone.yml
```diff
@@ -176,6 +176,14 @@ pipeline:
       branch: [ refs/tags/* ]
       status: success
 
+  notify:
```
jakedsouza: The push build failed with  `ERROR: Image not found` 

As per the docs http://plugins.drone.io/drone-plugins/drone-downstream/ 
I think you dont need the step `notify:` as per the docs here 

I think you should have it as 

```
pipeline:
... 
...
  trigger-downstream:
    image: plugins/downstream
    server: https://ci.vcna.io
    token: ${DRONE_TOKEN}
    fork: true
    repositories:
       - vmware/vic-product
```

mhagen-vmware:
Latest succeeded, see https://ci.vcna.io/vmware/vic/10933

mhagen-vmware:
however, I think we might not actually want fork: true, it seems to have not done quite what we wanted here:
https://ci.vcna.io/vmware/vic-product/44

It just forked the last build and we want a new build...

PR #5232, opened 2017-05-24
Add hgfs server support to toolbox

### LINKED ISSUE #4103: toolbox support for guest file manager operations
**Story**
As a user of toolbox I want to be able to transfer files to and from the guest

**Detail**
Guest Operations provides the following for guest file manipulation: http://pubs.vmware.com/vsphere-60/index.jsp#com.vmware.wssdk.apiref.doc/vim.vm.guest.FileManager.html
The govmomi side of FileManager should already be available.

The capabilities we're primarily interested in adding to toolbox are:
* InitiateFileTransferFromGuest
* InitiateFileTransferToGuest

**Acceptance**
Able to round trip a file to the guest via govc

dougm:
Add hgfs server support to toolbox

This makes it possible to copy files from inside the guest to the outside,
using the GuestFileManager.InitiateFileTransferFromGuest API.

Exercising the API, for example with:
  govc guest.download -vm $vm /etc/lsb-release -

Results in the VMX calling the following RPCs:

- vixCommandInitiateFileTransferFromGuest

- vmxiHgfsSendPacketCommand(s):
  * OpCreateSessionV4
  * OpGetattrV4
  * OpOpen
  * OpReadV3 [N times depending on file size]
  * OpClose
  * OpDestroySessionV4

For VMX reference, see: lib/hgFileCopy/hgFileCopy.c

Writing files to the guest is not included in this change, but will be a similar flow,
with Write/Setattr rather than Getattr/Read.

Fixes #4103

### DIFF
--- a/pkg/vsphere/toolbox/hgfs/encoding.go
+++ b/pkg/vsphere/toolbox/hgfs/encoding.go
@@ -1,0 +1,67 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hgfs
+
+import (
+	"bytes"
+	"encoding"
+	"encoding/binary"
+)
+
+// MarshalBinary is a wrapper around binary.Write
+func MarshalBinary(fields ...interface{}) ([]byte, error) {
+	buf := new(bytes.Buffer)
+
+	for _, p := range fields {
+		switch m := p.(type) {
+		case encoding.BinaryMarshaler:
+			data, err := m.MarshalBinary()
+			if err != nil {
+				return nil, ProtocolError(err)
+			}
+
+			_, _ = buf.Write(data)
+		case []byte:
+			_, _ = buf.Write(m)
+		case string:
+			_, _ = buf.WriteString(m)
+		default:
+			err := binary.Write(buf, binary.LittleEndian, p)
+			if err != nil {
+				return nil, ProtocolError(err)
+			}
+		}
+	}
+
+	return buf.Bytes(), nil
+}
+
+// UnmarshalBinary is a wrapper around binary.Read
+func UnmarshalBinary(data []byte, fields ...interface{}) error {
+	buf := bytes.NewBuffer(data)
+
+	for _, p := range fields {
+		if m, ok := p.(encoding.BinaryUnmarshaler); ok {
+			return m.UnmarshalBinary(buf.Bytes())
+		}
+
+		err := binary.Read(buf, binary.LittleEndian, p)
+		if err != nil {
+			return ProtocolError(err)
+		}
+	}
+
+	return nil
+}
--- a/pkg/vsphere/toolbox/hgfs/hgfs_linux.go
+++ b/pkg/vsphere/toolbox/hgfs/hgfs_linux.go
@@ -1,0 +1,52 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hgfs
+
+import (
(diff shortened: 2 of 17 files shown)


pkg/vsphere/toolbox/toolbox_linux.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	size := info.Size()
+	mtime := info.ModTime().Unix()
+	perm := info.Mode().Perm()
+
+	sys := info.Sys().(*syscall.Stat_t)
+
+	atime := time.Unix(sys.Atim.Unix()).Unix()
+	uid := sys.Uid
```
hickeng: We may need to do something different here that ties into uid/gid that exists in the containerVM. There's likely an error if trying to use a gid/uid other than superuser or a specific user that exists in the container explicitly.

It might be that this is checking/normalization needs to happen on the PL side instead of here.

pkg/vsphere/toolbox/vix_command.go
```diff
@@ -23,13 +23,22 @@ import (
...
+	"github.com/vmware/vic/pkg/vsphere/toolbox/hgfs"
 )
 
 const (
 	vixCommandMagicWord = 0xd00d0001
 
 	vixCommandGetToolsState = 62
-	vixCommandStartProgram  = 185
+
+	vixCommandKillProcess     = 85
```
hickeng: We going to migrate the "kill" command to this?

pkg/vsphere/toolbox/hgfs/hgfs_linux.go
```diff
@@ -0,0 +1,52 @@
...
+	nt := func(t syscall.Timespec) uint64 {
+		return uint64(t.Nano()) // TODO: this is supposed to be Windows NT system time, not needed atm
+	}
+
+	a.AccessTime = nt(sys.Atim)
+	a.WriteTime = nt(sys.Mtim)
+	a.CreationTime = a.WriteTime // see HgfsGetCreationTime
+	a.AttrChangeTime = nt(sys.Ctim)
+
+	a.SpecialPerms = uint8((sys.Mode & (syscall.S_ISUID | syscall.S_ISGID | syscall.S_ISVTX)) >> 9)
```
fdawg4l: Are these offsets defined somewhere?  Is this an artifact of the protocol or some datastructure?

pkg/vsphere/toolbox/vix_command_test.go
```diff
@@ -233,3 +291,87 @@ func TestVixCommandErrors(t *testing.T) {
 		t.Error("expected error")
 	}
 }
+
+func TestVixInitiateFileTransfer(t *testing.T) {
+	c := NewVixCommandClient()
+
+	request := new(VixMsgListFilesRequest)
+
+	names := []string{"enoent", "/etc/resolv.conf"}
```
hickeng: Linux specific test - replace with tmpfile?

pkg/vsphere/toolbox/hgfs/encoding.go
```diff
@@ -0,0 +1,67 @@
...
+	return buf.Bytes(), nil
+}
+
+// UnmarshalBinary is a wrapper around binary.Read
+func UnmarshalBinary(data []byte, fields ...interface{}) error {
+	buf := bytes.NewBuffer(data)
+
+	for _, p := range fields {
+		if m, ok := p.(encoding.BinaryUnmarshaler); ok {
+			return m.UnmarshalBinary(buf.Bytes())
```
dougm: There will have been >= 1 calls to binary.Read below, so buf.Bytes() here is the remaining unread portion of data.

pkg/vsphere/toolbox/hgfs/hgfs_linux.go
dougm: Just spreading the st_mode bits to these fields as vmtoolsd does: https://github.com/vmware/open-vm-tools/blob/master/open-vm-tools/lib/hgfsServer/hgfsServerLinux.c#L2548-L2551

pkg/vsphere/toolbox/hgfs/protocol.go
```diff
@@ -0,0 +1,681 @@
...
+// FileName as defined in hgfsProto.h:HgfsFileName
+type FileName struct {
+	Length uint32
+	Name   string
+}
+
+// MarshalBinary implements the encoding.BinaryMarshaler interface
+func (f *FileName) MarshalBinary() ([]byte, error) {
+	name := f.Name
+	f.Length = uint32(len(f.Name))
```
dougm: Assuming Go's string is ok there, as os.File.Name() is also a string type (https://golang.org/pkg/os/#FileInfo).

pkg/vsphere/toolbox/toolbox_linux.go
dougm: This is just reading from a stat() call inside the container VM.  We may need to do some user/group name to id mapping in the PL when hgfs Setattr support is added though.

pkg/vsphere/toolbox/vix_command.go
dougm: I grabbed this constant along with vixCommandListProcessesEx while I was in there, with doing a proper async vixCommandStartProgram impl.  But the pseudo "kill" command will probably stay as-is, since the TerminateProcessInGuest method (which maps to vixCommandKillProcess), does not take any arguments other than a PID.

pkg/vsphere/toolbox/vix_command_test.go
dougm: Yeah I'll fix that, its fine for linux + darwin, but will fail on windows.

PR #5233, opened 2017-05-24
4557/proxy vic admin

### LINKED ISSUE #4557: Make vicadmin respect proxy settings
Proxy settings are not applied to the applianceVM as a whole, so the settings must be read out of the environment (stored in an env var) and passed into the Golang HTTP client that does the health checks in vicadmin so that those health checks respect the proxy settings used elsewhere in the application.

morris-jason:
Fixes #4557 

- Removed HTTP[S]_PROXY env var for vicadmin. It was causing govmomi to attempt connections to vSphere over the proxy.
- WAN Checks now go over the proxy if it exists.


lib/vicadmin/validate.go
```diff
@@ -172,20 +174,42 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
+		proxy = sproxy
+	} else if hproxy := vchArgs["--http-proxy"]; hproxy != "" {
+		proxy = hproxy
+	}
+	if proxy != "" {
+		url, err := url.Parse(proxy)
 		if err != nil {
 			nwErrors = append(nwErrors, err)
 		} else {
-			conn.Close()
```
morris-jason: this is a diff bug

sgairo:
Could you add tests please?

lib/vicadmin/validate.go
```diff
@@ -172,20 +174,42 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
+			client.Transport = &http.Transport{Proxy: http.ProxyURL(url)}
 		}
 	}
 
+	// preform the wan check
```
sgairo: "perform"

morris-jason: Oooh

lib/vicadmin/validate.go
```diff
@@ -172,20 +174,42 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
 	}
+
 	nwErrors := []error{}
 
-	for _, host := range hosts {
-		conn, err := net.DialTimeout("tcp", host, time.Minute/2)
+	// create a http client with a custom transport. The transport uses the [s/h]proxy defined in vicmachine create if applicable.
+	client := &http.Client{Timeout: 30 * time.Second}
+	vchArgs := make(map[string]string)
+	for _, args := range vch.VicMachineCreateOptions {
```
emlin: vicmachine create options is added for debug purpose so far. Why not read from env settings?

morris-jason: http.Client automatically picks up the proxy env var, which would be good for vicadmin expect govmomi also uses http.Client. When the proxy was supplied, govmomi was attempting to connect to vSphere/ESX over the proxy.

lib/vicadmin/validate.go
```diff
@@ -172,20 +174,42 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
+		splitArgs := strings.Split(args, "=")
+		vchArgs[splitArgs[0]] = splitArgs[1]
+	}
+	//priority given to https proxies
+	proxy := ""
+	if sproxy := vchArgs["--https-proxy"]; sproxy != "" {
```
anchal-agrawal:
```go
proxy := vchArgs["--https-proxy"]
if proxy == "" {
 	proxy = vchArgs["--http-proxy"]
}
```

The above snippet has the same behavior and is easier to read.

matthewavery: Additionally, we will likely want to check if the element is actually present in the map. It is true that even when an element is not present in a `map[string]string`the return value witll be `""`. However, it is still of interest of us whether or not the value was set to that or just defaulted to that value. 

```
 proxy, ok := vchArgs["--https-proxy"]
if !ok {
 	proxy = vchArgs["--http-proxy"]
}
```
when retrieving values from a map if `golang` you always get two values back. the value and a boolean for if the key was present.

lib/vicadmin/validate.go
```diff
@@ -172,20 +174,42 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
-	for _, host := range hosts {
-		conn, err := net.DialTimeout("tcp", host, time.Minute/2)
+	// create a http client with a custom transport. The transport uses the [s/h]proxy defined in vicmachine create if applicable.
+	client := &http.Client{Timeout: 30 * time.Second}
+	vchArgs := make(map[string]string)
+	for _, args := range vch.VicMachineCreateOptions {
+		splitArgs := strings.Split(args, "=")
+		vchArgs[splitArgs[0]] = splitArgs[1]
+	}
+	//priority given to https proxies
```
anchal-agrawal: Minor gofmt suggestion: have a space b/w `//` and the comment text.

lib/vicadmin/validate.go
matthewavery: We also should likely have a constant for this raw string, just in case it gets reused in the future.

matthewavery: ditto for the `--http-proxy` string 😄

morris-jason: Using Anchal's code snippet, the [h/s]proxy will be available in the proxy var.

lib/vicadmin/validate.go
morris-jason: Done, I should really pay attention to golinter.

lib/vicadmin/validate.go
morris-jason: @matthewavery Seems like that just adds reading complexity in this case?

lib/vicadmin/validate.go
emlin: hmm.. got your point.
So here in vicadmin, you'll need two connections, one to vSphere without proxy and one to googles needs proxy.
But you might not want to change the http proxy settings in guestinfo, which will involves upgrade change to support old version's VCH.
One suggestion is to look at the session creation code, to make it ignore default http proxy if is specified.

morris-jason: I don't think what I changes should alter guestinfo, just access it. 
Even if I changed the Session Client to ignore the http variable there still may be other places that use http.Client in vicadmin or it's vendored dependencies. 
Basically, it seems like the proper approach to only include the proxy where needed instead of excluding it in all the place it's not needed.

tests/test-cases/Group9-VIC-Admin/9-01-VICAdmin-ShowHTML.md
```diff
@@ -22,3 +22,68 @@ This test requires that a vSphere environment be running and available
 * VICadmin responds with a log file indicating VCH init has begun reaping processes
 * VICadmin responds with log file indicating docker personality service has started
 * VICadmin responds with a ZIP file containing at a minimum the vmware.log file from the new container
+
+#Unauthenticated Test Cases
```
anchal-agrawal: Nit: this doesn't render the markdown heading correctly. `# Unauthenticated Test Cases` is the correct format. Bonus points for updating all headings in this file 😃

morris-jason: Weird..... Markdown Linter lied to me! Will update after the other build finished.

lib/vicadmin/validate.go
matthewavery: @morris-jason for the change to checking the `ok` portion it is subtle but it does add a difference. In one case we are saying "did the user of this map give us this key with the value `""`"  the case where we use `!ok` protects us from this distinction. We will now be able to tell if someone attempted to give us `""` or did not attempt to populate that field at all.

lib/vicadmin/validate.go
```diff
@@ -172,20 +174,40 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
-		conn, err := net.DialTimeout("tcp", host, time.Minute/2)
+	// create a http client with a custom transport. The transport uses the [s/h]proxy defined in vicmachine create if applicable.
+	client := &http.Client{Timeout: 30 * time.Second}
+	vchArgs := make(map[string]string)
+	for _, args := range vch.VicMachineCreateOptions {
+		splitArgs := strings.Split(args, "=")
+		vchArgs[splitArgs[0]] = splitArgs[1]
+	}
+	// priority given to https proxies
+	proxy := vchArgs["--https-proxy"]
```
matthewavery: the raw strings here and 2 lines below should still be made constants.

morris-jason: Done.

lib/vicadmin/validate.go
```diff
@@ -65,6 +67,8 @@ const (
 	GoodStatus     = template.HTML(`<i class="icon-ok"></i>`)
 	BadStatus      = template.HTML(`<i class="icon-attention"></i>`)
 	DefaultVCHName = ` `
+	HTTPSProxy     = "--https-proxy"
```
anchal-agrawal: These should be unexported, i.e. `httpsProxy` and `httpProxy` unless they're being used outside this package. 

Actually, it would be nice to leave these exported and use them in https://github.com/vmware/vic/blob/master/cmd/vic-machine/create/create.go (lines 453 and 461) so we can consolidate these variables. Check with @emlin on this approach.

lib/install/management/appliance.go
```diff
@@ -536,12 +536,6 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 		},
 		Dir: "/home/vicadmin",
 	}
-	if settings.HTTPProxy != nil {
```
hmahmood: You could just rename the env vars so that golang http client doesn't pick them up and then use them for the connectivity check.

morris-jason: Will change to this approach following @emlin's comment that VicMachineCreateOptions was added for debug purposes.

emlin:
this solution can workaround the issue for vsphere connection, but it still involves upgrade and reconfigure change.
you need a upgrade plugin following this instruction https://github.com/vmware/vic/blob/master/lib/migration/plugins/plugins.md
and you'll need to update the reconfigure a little bit. But if you can finish this PR quickly, I'll do it for you, otherwise, will let you know where to change it.
basically, any change in persisted configuration, will introduce these two more issues.

lib/vicadmin/validate.go
```diff
@@ -172,20 +175,35 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
...
+		"http://google.com",
+		"https://docker.io",
 	}
+
 	nwErrors := []error{}
 
-	for _, host := range hosts {
-		conn, err := net.DialTimeout("tcp", host, time.Minute/2)
+	// create a http client with a custom transport using the proxy from env vars
+	client := &http.Client{Timeout: 30 * time.Second}
```
caglar10ur: THANK YOU :) That time.Minute/2 was triggering my OCD

morris-jason: Haha :)

lib/migration/plugins/plugin6/vicadminProxyVarRename.go
```diff
@@ -0,0 +1,105 @@
...
+		return nil
+	}
+	mapData := data.(map[string]string)
+	oldStruct := &VCHConfig{}
+	result := extraconfig.Decode(extraconfig.MapSource(mapData), oldStruct)
+	if result == nil {
+		return &errors.DecodeError{}
+	}
+	//translate old proxy env var keys into to proxy env var keys
+	newEnvs := make([]string, 0)
```
hmahmood:
```go
var newEnvs []string
```

morris-jason:
CI Is failing lib/migration tests. Not sure why - my plugin has a unique number. Someone please take a peek for me. 😃

lib/migration/plugins/plugin6/vicadminProxyVarRename.go
```diff
@@ -0,0 +1,105 @@
...
+	oldStruct := &VCHConfig{}
+	result := extraconfig.Decode(extraconfig.MapSource(mapData), oldStruct)
+	if result == nil {
+		return &errors.DecodeError{}
+	}
+	//translate old proxy env var keys into to proxy env var keys
+	var newEnvs []string
+	for _, envVar := range oldStruct.Sessions["vicadmin"].Cmd.Env {
+		if strings.Contains(envVar, oldHProxy) {
+			proxyVarValue := strings.Split(envVar, "=")[1]
```
emlin: might be better to check the split string length, to avoid panic here.

lib/migration/plugins/plugin6/vicadminProxyVarRename.go
```diff
@@ -0,0 +1,105 @@
...
+			newEnvs = append(newEnvs, fmt.Sprintf("%s=%s", newHProxy, proxyVarValue))
+		} else if strings.Contains(envVar, oldSProxy) {
+			proxyVarValue := strings.Split(envVar, "=")[1]
```
emlin: ditto

lib/migration/plugins/plugin6/vicadminProxyVarRename.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	mapData := data.(map[string]string)
+	oldStruct := &VCHConfig{}
+	result := extraconfig.Decode(extraconfig.MapSource(mapData), oldStruct)
+	if result == nil {
+		return &errors.DecodeError{}
+	}
+	//translate old proxy env var keys into to proxy env var keys
+	var newEnvs []string
+	for _, envVar := range oldStruct.Sessions["vicadmin"].Cmd.Env {
```
emlin: a constant for "vicadmin"?

emlin:
You can try to update this line https://github.com/vmware/vic/blob/master/lib/migration/migrator_test.go#L35 to -1, because you have a plugin registered with latest version.

lib/migration/plugins/plugin6/vicadminProxyVarRename.go
morris-jason: Done.

lib/migration/plugins/plugin6/vicadminProxyVarRename.go
morris-jason: Done.

lib/migration/plugins/plugin6/vicadminProxyVarRename.go
morris-jason: Done.

lib/migration/plugins/init.go
```diff
@@ -20,5 +20,5 @@ import (
 	_ "github.com/vmware/vic/lib/migration/plugins/plugin1"
 	_ "github.com/vmware/vic/lib/migration/plugins/plugin2"
 	_ "github.com/vmware/vic/lib/migration/plugins/plugin5"
-	_ "github.com/vmware/vic/lib/migration/plugins/plugin7"
+	_ "github.com/vmware/vic/lib/migration/plugins/plugin6"
```
emlin: why change it back to 6?

morris-jason: 7 was causing the test failure because of the order of versions in feature.go. I swapped it's place with plugin 6 and now it works.

lib/migration/feature/feature.go
```diff
@@ -20,8 +20,8 @@ const (
 	TasksSupportedVersion
 	RenameSupportedVersion
 	MigrateRegistryVersion
-	ExecSupportedVersion
 	VicadminProxyVarRenameVersion
+	ExecSupportedVersion
```
emlin: this is wrong, you never insert your new plugin version before any others

lib/migration/plugins/init.go
emlin: Did you try this?
`You can try to update this line https://github.com/vmware/vic/blob/master/lib/migration/migrator_test.go#L35 to -1, because you have a plugin registered with latest version.`

morris-jason: Yes. Still getting an error. The problem is that the migrator_test needs 2 plugin versions free to update. If my plugin is last there's only one plugin version free (the MaxPluginVersion).

morris-jason: Done.

lib/migration/feature/feature.go
morris-jason: Done.

anchal-agrawal:
@morris-jason please remember to add a descriptive commit message while merging the PR - "4557/proxy vic admin" doesn't convey the intent clearly.

lib/migration/plugins/plugin7/vicadminProxyVarRename.go
```diff
@@ -0,0 +1,108 @@
...
+type Cmd struct {
+	Env []string `vic:"0.1" scope:"read-only" key:"Env"`
+}
+
+func (p *VicadminProxyVarRename) Migrate(ctx context.Context, s *session.Session, data interface{}) error {
+	defer trace.End(trace.Begin(fmt.Sprintf("%d", feature.VicadminProxyVarRenameVersion)))
+	if data == nil {
+		return nil
+	}
+	mapData := data.(map[string]string)
```
jzt: Any chance `data` would be something other than a `map[string]string`?

lib/migration/plugins/plugin7/vicadminProxyVarRename.go
```diff
@@ -0,0 +1,108 @@
...
+	oldStruct := &VCHConfig{}
+	result := extraconfig.Decode(extraconfig.MapSource(mapData), oldStruct)
+	if result == nil {
+		return &errors.DecodeError{}
+	}
+	//translate old proxy env var keys into to proxy env var keys
+
```
jzt: whitespace looks like it should be on line 80 instead :)

lib/migration/plugins/plugin7/vicadminProxyVarRename.go
```diff
@@ -0,0 +1,108 @@
...
+	if oldStruct.Sessions != nil && oldStruct.Sessions[sessionName] != nil && oldStruct.Sessions[sessionName].Cmd.Env != nil {
```
gigawhitlocks: presumably you need to bail and return an error here if one of these is not nil.
In Go it is idiomatic to have the happy path unindented, so you likely want to refactor this to say, roughly
``` if oldStruct.Sessions == nil || oldStruct.Sessions[... {
          return error 
} 
// do stuff that was previously inside the if statement, but now on the outer level
```

morris-jason: We can't error though because the plugin will be run regardless if it actually needs to. 
If I error, the migration process will stop whenever the proxies aren't defined and in all future versions. I'll refactor the inside code to be unindented and add a log statement though.

lib/migration/plugins/plugin7/vicadminProxyVarRename.go
```diff
@@ -77,9 +77,12 @@ func (p *VicadminProxyVarRename) Migrate(ctx context.Context, s *session.Session
 	if result == nil {
 		return &errors.DecodeError{}
 	}
-	//translate old proxy env var keys into to proxy env var keys
 
-	if oldStruct.Sessions != nil && oldStruct.Sessions[sessionName] != nil && oldStruct.Sessions[sessionName].Cmd.Env != nil {
+	// translate old proxy env var keys into to proxy env var keys
+	// skip upgrading if the proxy isn't defined or something is wrong with the the vicadmin executor
+	if oldStruct.Sessions == nil || oldStruct.Sessions[sessionName] == nil || oldStruct.Sessions[sessionName].Cmd.Env == nil {
+		log.Debugln("vicadmin Session found. skipping proxy rename")
```
emlin: not found

lib/migration/plugins/init.go
```diff
@@ -20,4 +20,5 @@ import (
 	_ "github.com/vmware/vic/lib/migration/plugins/plugin1"
 	_ "github.com/vmware/vic/lib/migration/plugins/plugin2"
 	_ "github.com/vmware/vic/lib/migration/plugins/plugin5"
+	_ "github.com/vmware/vic/lib/migration/plugins/plugin7"
```
caglar10ur: we should send a pr to rename those, pluginX is really confusing. It could be the name of the enum for example, _ "github.com/vmware/vic/lib/migration/plugins/VicadminProxyVarRenameVersion"

lib/migration/plugins/plugin7/vicadminProxyVarRename.go
emlin: not likely. this is in the data migration chain, from a specific version to another one.
Even if we changed the data format in the future, the format for this specific plugin should still be this map only.
the input is defined as interface, just to make sure we don't break the data migration chain if meta data is moved somewhere else with different data format

lib/migration/plugins/plugin7/vicadminProxyVarRename.go
gigawhitlocks: OK cool so it's not an error condition, just looks like one. For what it's worth I've yet to write one of these migration plugins so you probably know more than me about them by now 😆

PR #5237, opened 2017-05-24
[specific ci=1-10-Docker-PS]Fix bug in "Docker ps only running containers" in test 1-10-Docker-PS

chengwang86:


rogeliosanchez:
lgtm

PR #5255, opened 2017-05-25
Add research spike as a high priority

mhagen-vmware:
As we discussed in our sprint planning and retrospective, until we have a better process defined around future release planning features we need to include research spikes as high priority items because without researching and decomposing future feature work we will be unable to work on it.


mhagen-vmware:
@mdubya66 @hickeng @karthik-narayan @corrieb I would still like approval from some/all of you on this change.

PR #5256, opened 2017-05-25
Remove parallelization in make

mhagen-vmware:
As discussed in sprint planning and retro, we are concerned that our build process is not really make parallel safe, so this is an attempt to go back to no parallel make building.  We need to analyze what this does to our build times.

P.S. not sure how -j9 just disappeared.  @hmahmood and I can't see it anywhere in our git history anymore...


.drone.yml
```diff
@@ -44,14 +44,14 @@ pipeline:
...
-      - make -j3 all
+      - make all
       - make sincemark
       - make mark
       - echo `ls vendor/github.com/vmware/govmomi/vim25/methods`
       - echo `ls vendor/github.com/vmware/govmomi/vim25/types`
       - echo `ls vendor/github.com/docker/docker/vendor/github.com/opencontainers/runc/libcontainer/system`
       - export VIC_ESX_URL_ARRAY="`tests/get_test_url.sh`"
-      - make -j3 test
+      - make test
```
hmahmood: I think for the `go test` portion, I would like make -j3 since I believe go will only parallelize for the compile step. @hickeng @caglar10ur can clarify if this is true.

mhagen-vmware:
@hmahmood I think most of the failure we are trying to address with this are actually in test

mhagen-vmware:
Time of two jobs with this change:  1503s, 773s  (The 1503s appears to have come entirely from the make all not the make test. It appears a giant chunk of it was due to building vicadmin...)

Two jobs with -j3: 735s, 752s

mhagen-vmware:
Re-runs of this PR:  800s and 778s

So 3/4 are within the same ballpark of -j3, not sure where that 1500s came from, but I am going to merge this now as the -j3 doesn't appear to be providing much benefit at all.

PR #5271, opened 2017-05-26
Add functionality to push comment to github

mhagen-vmware:


cgtexmex:
Why not allow the method to return the error and let the caller decide if they want to ignore or not?  I assume in most cases it would just be a log entry (that may never be detected) but at least there's some historical evidence of an attempt.

PR #5281, opened 2017-05-28
Add toolbox support for guest file manager write operations

### LINKED ISSUE #5234: toolbox support for guest file manager write operations
**Story**
As a user of toolbox I want to be able to transfer files into the guest

**Detail**
Guest Operations provides the following for guest file manipulation: http://pubs.vmware.com/vsphere-60/index.jsp#com.vmware.wssdk.apiref.doc/vim.vm.guest.FileManager.html
The govmomi side of FileManager is already be available.

With general hgfs server suppoort (#4103) in place, we now need to add toolbox support for:

* InitiateFileTransferToGuest

* hgfs OpWrite

* hgfs OpSetattr

**Acceptance**
Able to tranfer a file into the guest via govc guest.upload

dougm:
Closes #5234

### DIFF
--- a/pkg/vsphere/toolbox/hgfs/server.go
+++ b/pkg/vsphere/toolbox/hgfs/server.go
@@ -42,6 +42,9 @@ type Server struct {
 	sessions map[uint64]*session
 	mu       sync.Mutex
 	handle   uint32
+
+	chmod func(string, os.FileMode) error
+	chown func(string, int, int) error
 }
 
 // NewServer creates a new Server instance with the default handlers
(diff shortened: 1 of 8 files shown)


pkg/vsphere/toolbox/hgfs/server.go
```diff
@@ -228,6 +236,57 @@ func (s *Server) GetattrV2(p *Packet) (interface{}, error) {
...
+	req := new(RequestSetattrV2)
+	err := UnmarshalBinary(p.Payload, req)
+	if err != nil {
+		return nil, err
+	}
+
+	name := req.FileName.Path()
+
+	uid := -1
+	if req.Attr.Mask&AttrValidUserID == AttrValidUserID {
```
hickeng: This will need normalizing in some manner - we shouldn't be writing these files with the client side uid/gid (which is what I'd expect to see in the tar).
If the docker client doesn't do translation when constructing the tar archive then we need to settle on what to use, which pretty much has to be 0 or the currently defined container USER.

pkg/vsphere/toolbox/hgfs/server.go
```diff
@@ -228,6 +236,57 @@ func (s *Server) GetattrV2(p *Packet) (interface{}, error) {
...
+	if req.Attr.Mask&AttrValidGroupPerms == AttrValidGroupPerms {
+		perm |= os.FileMode(req.Attr.GroupPerms) << 3
+	}
+
+	if req.Attr.Mask&AttrValidOtherPerms == AttrValidOtherPerms {
+		perm |= os.FileMode(req.Attr.OtherPerms)
+	}
+
+	if perm != 0 {
+		err = s.chmod(name, perm)
```
hickeng: Is it possible to extract this information before we start writing the file? Otherwise there's a period of time where the permissions and uid/gid do not align with the attrs.
If not we should likely make sure there's an exclusive lock until we have applied said attrs.

dougm: The fields are in the OpenV3 request protocol, but they are unused by the VMX - it waits until the file has been written to call SetattrV2.

Issue #5245, opened 2017-05-25
After container exits VIC is still reporting in as running.


lcastellano:
**VIC version:**
v0.8.0-100000-bcddf87
master

**Deployment details:**
VCH deployed on Nimbus

**Steps to reproduce:**
Create 10 short lived containers concurrently. 

docker create busybox /bin/ls
docker .. start

**Actual behavior:**
1 or 2 containers are still marked as running, however the correspondent VMs are powered off.
Those containers cannot be cleaned-up.

**Expected behavior:**
All containers should be powered off and ready for removal.

**Logs:**
[container-logs.zip](https://github.com/vmware/vic/files/1029700/container-logs.zip)

**Additional details as necessary:**

anchal-agrawal:
This issue is captured in https://github.com/vmware/vic/issues/5235 and https://github.com/vmware/vic/issues/4997. Closing.

PR #5284, opened 2017-05-30
Partial fix for issues: #5235, #5245

### LINKED ISSUE #5235: "Docker ps only running containers" fail in ci test 1-10-Docker-PS
Seen here https://ci.vcna.io/vmware/vic/10853.

lcastellano:
This fixes a real problem where events go missing. It was seen in multiple/concurrent container create/start. However, there could be other problems. In particular if an event is not received from VSphere, VIC engine could still be in an inconsistent state. Where the VM is powered off but the container is still listed as running.

Fixes #5235

### DIFF
--- a/lib/portlayer/event/collector/vsphere/collector.go
+++ b/lib/portlayer/event/collector/vsphere/collector.go
@@ -128,7 +128,7 @@ func (ec *EventCollector) Start() error {
 	force := false
 
 	//TODO: need a proper way to handle failures / status
-	go func(page int32, follow bool, ff bool, refs []types.ManagedObjectReference, ec *EventCollector) error {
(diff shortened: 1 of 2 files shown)


lib/portlayer/event/collector/vsphere/collector.go
```diff
@@ -176,7 +176,7 @@ func evented(ec *EventCollector, page []types.BaseEvent) {
 			ec.callback(NewVMEvent(page[i]))
 		case *types.VmReconfiguredEvent:
 			// reconfigures happen often, so completely ignore for now
-			return
+			continue
 		default:
```
cgtexmex: Ugh...nice catch I totally screwed that up...FTR this was added to help identify missed events but this would've made it worse.

PR #5288, opened 2017-05-30
Need to actually pull the latest

mhagen-vmware:


morris-jason:
LGTM

mhagen-vmware:
don't think we need this.

PR #5292, opened 2017-05-30
Clean and fetch only the branch we need for nightlies

mhagen-vmware:


tests/nightly/nightly-autorun.sh
```diff
@@ -16,7 +16,8 @@
 
 # Get the latest code from vmware/vic repo
 cd ~/go/src/github.com/vmware/vic
-git fetch
+git clean -f
```
rogeliosanchez: This command removes the `nightly_test_secrets.yml` which the executor needs.  Should we move this file to another directory?
I was planning on cloning the vic-internal repo to the home directory so we could have the most updated version of this nightly_test_secrets.yml file on the nightly machine.  Obviously, we would need to point the drone command to execute this file from the right place.

tests/nightly/nightly-autorun.sh
```diff
@@ -16,8 +16,10 @@
 
 # Get the latest code from vmware/vic repo
 cd ~/go/src/github.com/vmware/vic
-git fetch
+git clean -fd
```
anchal-agrawal: I think there's a nightly secrets file in the project's directory and doing this will remove it.

tests/nightly/nightly-autorun.sh
```diff
@@ -16,8 +16,10 @@
...
+git fetch https://github.com/vmware/vic master
```
anchal-agrawal: I think we can just say `git fetch origin` instead of this. `git pull` on line 22 shouldn't be necessary either.

tests/nightly/nightly-autorun.sh
rogeliosanchez: Yes @anchal-agrawal, this will remove that `nightly_test_secrets.yml` file.  @mhagen-vmware, is there a way we can move that nightly secrets file somewhere else and update the drone execution command to use the new location?
I want to clone in the vic-internal repo somewhere on the nightly executor machine and update this but I'm still having some issues getting repo 'read' privileges for a user that is not me for this machine.

tests/nightly/nightly-autorun.sh
mhagen-vmware: I was just playing with it, it didn't look like fetch merged in the changes, so it looked to me like we need the pull too.

tests/nightly/nightly-autorun.sh
mhagen-vmware: Yea, the secrets should definitely be in a different place.  On my system I have the secrets checked out in a separate folder that we should git fetch and pull as well so that we are running off the latest secrets always too.

tests/nightly/nightly-autorun.sh
anchal-agrawal: `fetch` won't, but we do a `reset --hard` after it and that should suffice.

mhagen-vmware:
@anchal-agrawal @rogeliosanchez Ok, some cleanup here. Please take another look when you can.  Thanks!

tests/nightly/nightly-autorun.sh
```diff
@@ -16,8 +16,9 @@
 
 # Get the latest code from vmware/vic repo
 cd ~/go/src/github.com/vmware/vic
-git fetch
-git reset --hard origin/master
+sudo git clean -fd
```
anchal-agrawal: How did we deal with the secrets file?

rogeliosanchez: @mhagen-vmware, also, do we need to change where all of the nighty_$DATE_.txt files are stored?  That way we can keep logs for troubleshooting.

mhagen-vmware:
@rogeliosanchez @anchal-agrawal need lgtm from you still

tests/nightly/nightly-autorun.sh
anchal-agrawal: We should address this point before merging this change, since those nightly text log files are useful for debugging failures.

tests/nightly/nightly-autorun.sh
```diff
@@ -27,4 +27,4 @@ git reset --hard origin/master
 
 # Kick off the nightly
 now=$(date +"%m_%d_%Y")
-sudo ./tests/nightly/nightly-kickoff.sh > nightly_$now.txt 2>&1
+sudo ./tests/nightly/nightly-kickoff.sh > /home/vicadmin/nightly-log/nightly_$now.txt 2>&1
```
anchal-agrawal: Does the `~/nightly-log/` directory exist already?

rogeliosanchez: Yes.  I told Mike about it.  :)

mhagen-vmware: yep

PR #5300, opened 2017-05-31
[full ci] Initialize connections lazily

### LINKED ISSUE #4950: [portlayer] Connector could lazily initialize the connections
Follow up work from https://github.com/vmware/vic/pull/4339/files#r108397273

We create a new connection when we receive ids back from tether (as a result of a reload). Instead we could change the connections map to hold closures that knows how to create real connections and call them to create connections as needed (hence lazy initialization). The original problem George reported resolved differently, we no longer activate all task but the correct one. This though still will cause the connection map to have more then required connections.

https://github.com/vmware/vic/pull/4339/files#diff-4a8e153ac25578d27311e828f2772579R327

### LINKED ISSUE #4990: Exec call doesn't exit as expected
CI: https://ci.vmware.run/vmware/vic/10225
Logs: unavailable due to drone timeout - specific container-log bundle added captured directly: [container-logs.zip](https://github.com/vmware/vic/files/972269/container-logs.28.zip)

Test Group: `Test 1-38`
VCH Name: `VCH-10225-1432`
Failing test: `Exec Sort -i`
Failing container: `3d075279d005b834cedb49754561fa76682bdfbcc8377762c880aa2902f10a59`
(exec session: `db773b83d9db1556ced2d2ebe658e671a6cf2bef5d07e3512b4be0bfed6c2d77`)

caglar10ur:
We create a new connection when we receive container ids back from tether.
This PR changes the connections map to hold closures that knows how to
create connections. We now call them on demand.

Fixes #4990
Fixes #4950

### DIFF
--- a/lib/portlayer/attach/communication/connector.go
+++ b/lib/portlayer/attach/communication/connector.go
@@ -37,7 +37,7 @@ import (
 type Connector struct {
 	mutex        sync.RWMutex
 	cond         *sync.Cond
-	interactions map[string]SessionInteractor
+	interactions map[string]*LazySessionInteractor
 
 	listener net.Listener
 	// Quit channel for serve
@@ -55,7 +55,7 @@ func NewConnector(listener net.Listener) *Connector {
 	defer trace.End(trace.Begin(""))
 
 	connector := &Connector{
-		interactions: make(map[string]SessionInteractor),
+		interactions: make(map[string]*LazySessionInteractor),
 		listener:     listener,
 		done:         make(chan struct{}),
 	}
@@ -64,29 +64,45 @@ func NewConnector(listener net.Listener) *Connector {
 	return connector
 }
 
-func (c *Connector) aliveAndKicking(ctx context.Context, id string) SessionInteractor {
-	c.mutex.Lock()
-	defer c.mutex.Unlock()
+// SessionIfAlive returns SessionInteractor or error
+func (c *Connector) SessionIfAlive(ctx context.Context, id string) (SessionInteractor, error) {
+	c.mutex.RLock()
+	v, ok := c.interactions[id]
+	c.mutex.RUnlock()
 
-	conn := c.interactions[id]
-	// we already established this connection, let's check it status
-	if conn != nil {
-		log.Infof("attach connector: Pinging for %s", id)
+	if !ok {
+		return nil, fmt.Errorf("attach connector: no such connection in the map")
+	}
+	// we have an entry in the map, let's check its status
+	var conn SessionInteractor
+	var err error
 
-		if err := conn.Ping(); err == nil {
-			log.Infof("attach connector: Pinged %s, returning", id)
-
-			if err := conn.Unblock(); err == nil {
-				log.Infof("attach connector: Unblocked %s, returning", id)
-			}
-			return conn
-		}
-		// ping failed so we need to remove it from the map
-		log.Infof("attach connector: Ping test failed, removing %s from connection map", id)
-		delete(c.interactions, id)
+	conn, err = v.Initialize()
+	if err != nil {
+		goto Error
 	}
 
-	return nil
+	log.Debugf("attach connector: Pinging for %s", id)
+	if err = conn.Ping(); err != nil {
+		goto Error
+	}
+
+	log.Debugf("attach connector: Unblocking for %s", id)
+	if err = conn.Unblock(); err != nil {
+		goto Error
+	}
+	log.Debugf("attach connector: Unblocked %s, returning", id)
+
+	return conn, nil
+
+Error:
+	log.Debugf("attach connector: liveness check failed, removing %s from connection map", id)
+
+	c.mutex.Lock()
+	delete(c.interactions, id)
+	c.mutex.Unlock()
+
+	return nil, err
 }
 
 // Interaction returns the interactor corresponding to the specified ID. If the connection doesn't exist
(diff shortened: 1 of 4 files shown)


caglar10ur:
FYI, this PR includes #5215 for now, I'll rebase this after I merge #5215

lib/portlayer/attach/communication/lazy.go
```diff
@@ -0,0 +1,48 @@
...
+	fn LazyInitializer
+}
+
+// Initialize either returns either already initialized connection or returns the connection after initializing it
+func (l *LazySessionInteractor) Initialize() (SessionInteractor, error) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
+	if l.si == nil && l.fn == nil {
+		return nil, fmt.Errorf("fn is nil")
```
dougm: seems like you'd want panic in this case vs. returning an error?

caglar10ur: done

lib/portlayer/attach/communication/connector.go
```diff
@@ -112,33 +128,46 @@ func (c *Connector) Interaction(ctx context.Context, id string) (SessionInteract
...
-			conn, ok = c.interactions[id]
+			v, ok = c.interactions[id]
 			if ok {
-				// no need to test this as we just created it
-				if err := conn.Unblock(); err == nil {
-					log.Infof("attach connector: Unblocked %s, returning", id)
+				conn, err := v.Initialize()
+				if err != nil {
+					log.Error(err)
+					ok = false
```
emlin: looks like it will break for loop for err happens, so result channel is not set, and context is not done yet, but the select https://github.com/vmware/vic/pull/5300/files#diff-4a8e153ac25578d27311e828f2772579R178 here does not handle that situation.

lib/portlayer/attach/communication/connector.go
```diff
@@ -64,29 +64,45 @@ func NewConnector(listener net.Listener) *Connector {
...
-			}
-			return conn
+		log.Debugf("attach connector: Pinging for %s", id)
+		if err = conn.Ping(); err != nil {
+			goto Error
 		}
-		// ping failed so we need to remove it from the map
-		log.Infof("attach connector: Ping test failed, removing %s from connection map", id)
-		delete(c.interactions, id)
+		log.Debugf("attach connector: Pinged %s, returning", id)
```
caglar10ur: done

lib/portlayer/attach/communication/connector.go
```diff
@@ -64,29 +64,45 @@ func NewConnector(listener net.Listener) *Connector {
...
+	v, ok := c.interactions[id]
+	c.mutex.RUnlock()
 
-	conn := c.interactions[id]
-	// we already established this connection, let's check it status
-	if conn != nil {
-		log.Infof("attach connector: Pinging for %s", id)
+	err := fmt.Errorf("attach connector: no such connection in the map")
+	// we have an entry in the map, let's check its status
+	if ok {
```
caglar10ur: done

lib/portlayer/attach/communication/connector.go
```diff
@@ -64,29 +64,45 @@ func NewConnector(listener net.Listener) *Connector {
...
+			goto Error
+		}
+		log.Debugf("attach connector: Unblocked %s, returning", id)
+
+		return conn, nil
 	}
 
-	return nil
+Error:
+	log.Debugf("attach connector: removing %s from connection map", id)
```
caglar10ur: done

lib/portlayer/attach/communication/connector.go
```diff
@@ -64,29 +64,45 @@ func NewConnector(listener net.Listener) *Connector {
...
+
+	c.mutex.Lock()
+	delete(c.interactions, id)
```
caglar10ur: I believe it does

lib/portlayer/attach/communication/connector.go
caglar10ur: done

caglar10ur:
@hickeng waiting the rest of your comments :)

lib/portlayer/attach/communication/connector.go
```diff
@@ -112,33 +128,44 @@ func (c *Connector) Interaction(ctx context.Context, id string) (SessionInteract
...
-
 	if ctx.Err() == context.DeadlineExceeded {
 		return nil, fmt.Errorf("attach connector: no such connection")
 	}
 
+	conn, err := c.SessionIfAlive(ctx, id)
+	if conn != nil && err == nil {
+		return conn, nil
+	}
+
```
hickeng: should we move this above the context deadline exceeded check? It may reduce the edge case behaviour where we've got one foot on either side of the deadline (e.g. stdout returned before deadline, stderr hitting the timeout and failing). This is the case where singleflight doesn't block the calls because they're not quite concurrent.

The change I suggest should help reduce the frequency of the issue, however we should also be checking in the personality that all of the required streams are successful or failing - iirc we were previously able to get into a state where one would succeeded and we would have no stdin, for a container run with `-it` for example, but stdout was successfuly so the client wouldn't exit.

lib/portlayer/attach/communication/connector.go
```diff
@@ -112,33 +128,44 @@ func (c *Connector) Interaction(ctx context.Context, id string) (SessionInteract
...
+			v, ok = c.interactions[id]
 			if ok {
-				// no need to test this as we just created it
-				if err := conn.Unblock(); err == nil {
-					log.Infof("attach connector: Unblocked %s, returning", id)
+				conn, err := v.Initialize()
+				if conn != nil && err == nil {
+					// no need to test this connection as we just created it, unblock if needed
+					log.Debugf("attach connector: Unblocking for %s", id)
+					err = conn.Unblock()
```
hickeng: There is a race in the failure/reset path of Unblock vs other Unblock calls - don't know if it's important.

lib/portlayer/attach/communication/lazy.go
```diff
@@ -0,0 +1,47 @@
...
+	mu sync.Mutex
+	si SessionInteractor
+	fn LazyInitializer
+}
+
+// Initialize either returns either already initialized connection or returns the connection after initializing it
+func (l *LazySessionInteractor) Initialize() (SessionInteractor, error) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
```
hickeng: Go idiom - fast path at the top then reduce indent of initialization portion:
```go
if l.si != nil {
  return l.si
}
```

lib/portlayer/attach/communication/connector.go
caglar10ur: done

lib/portlayer/attach/communication/connector.go
caglar10ur: done

lib/portlayer/attach/communication/lazy.go
caglar10ur: done

caglar10ur:
@hickeng pushed them as a new commit so that you can review

Issue #595, opened 2016-04-29
Migrate portlayer to pkg/vsphere/extraconfig


hickeng:
The portlayer should be getting it's configuration directly from guestinfo via extraconfig package instead of through the command line.

Consider moving to a single global `session` as part of this work, avoiding the function signature pollution from passing them around everywhere. This will position us better for moving to a vmomi gateway in the mid-term.

hmahmood:
We will use the struct here for serializing/deserializing the port layer config from guestinfo: https://github.com/vmware/vic/blob/master/metadata/virtual_container_host.go#L35 . The port-layer specific options will be split out into their own struct, and there will be global objects in the port layer package to access this config and a session object.

hickeng:
This is happening slowly as we have to make changes to where we source this information.This should be done for GA for cleanliness purposes.

PR #5304, opened 2017-05-31
[specific ci=Group6-VIC-Machine] enable vic-machine configure --debug

### LINKED ISSUE #4530: vic-machine reconfigure: Enable or disable the debug state
**User Statement:**
As a VIC operation user, I need the ability to enable or disable the debug state. Currently the VCH debug state is disabled by default. So I don't have enough log information when failures occur. It will be very helpful if I can switch to debug state (or a higher debug level with more information) to diagnose the failure/error.
This story does not include changing debug level for existing container. That can be down through govc (#4645)

**Example Use Case**
I deployed a VCH with the debug state disabled. A failure occurred in the portlayer and I would like to collect useful logs from the VCH by enabling the debug state.

**Details:**
- supported operations:
  - update the debug state of the VCH
- hot update:
  - no. Need to restart the VCH
- input/output data format:
  - input: e.g., ` --debug 1` (`debug=0` means disabling the debug state)
  - output: N/A
- impact on VCH/containerVMs:
  - VCH: The VCH will produce more log messages for debugging.
  - containerVMs: After we update the debug level of the VCH, new containers will automatically have the new debug level whereas the existing containers (created before vic-machine update) still have the previous debug level.
- is upgrade needed: 

**Acceptance Criteria:**
- add the ability to update the debug state.
- add relevant test cases

chengwang86:
fixes #4530 

This PR also fixes another bug in our existing code that the reconfigure snapshot is not deleted if the `configure` or `upgrade` operation succeeds.


cmd/port-layer-server/main.go
```diff
@@ -76,8 +79,17 @@ func main() {
...
+	if err != nil {
+		log.Errorf("Unable to load configuration from guestinfo")
+		return
+	}
+
+	extraconfig.Decode(src, &vchConfig)
+
 	logcfg := viclog.NewLoggingConfig()
-	if ploptions.PortLayerOptions.Debug {
+	if vchConfig.Diagnostics.DebugLevel {
```
emlin: why make this change?

tests/test-cases/Group6-VIC-Machine/6-16-Config.robot
```diff
@@ -18,9 +18,39 @@ Resource  ../../resources/Util.robot
 Suite Setup  Install VIC Appliance To Test Server
 Suite Teardown  Cleanup VIC Appliance On Test Server
 
+*** Keywords ***
+Verify VCH Debug State
+    [Arguments]  ${expected}
+    ${rc}  ${output}=  Run And Return Rc And Output  govc vm.info -e %{VCH-NAME} | grep guestinfo.vice./init/diagnostics/debug
+    Should Be Equal As Integers  ${rc}  0
+    Should Contain  ${output}  ${expected}
```
emlin: also need to check if vic-machine inspect can show the command option correctly

cmd/port-layer-server/main.go
chengwang86: The original `ploptions.PortlayerOptions.Debug` has default value of `debug` and is not initialized using `vchConfig.Diagnositics.DebugLevel`. Based on discussion with Hasan, I find that there is an existing issue #595 which requests obtaining all the portlayeroptions via extraconfig (what I did here).

tests/test-cases/Group6-VIC-Machine/6-16-Config.robot
chengwang86: Thanks. I have updated the test.

cmd/port-layer-server/main.go
emlin: I think issue #595 is too old, just occasionally not closed. Portlayer already query guestinfo configuration in Init method: https://github.com/vmware/vic/blob/master/lib/portlayer/portlayer.go#L60
Might consolidate these two queries

cmd/port-layer-server/main.go
```diff
@@ -76,8 +79,17 @@ func main() {
...
 	}
 
+	// load the vch config
+	src, err := extraconfig.GuestInfoSource()
+	if err != nil {
+		log.Errorf("Unable to load configuration from guestinfo")
+		return
+	}
+
+	extraconfig.Decode(src, &vchConfig)
```
emlin: I think issue #595 is too old, just occasionally not closed. Portlayer already query guestinfo configuration in Init method: https://github.com/vmware/vic/blob/master/lib/portlayer/portlayer.go#L60
Might consolidate these two queries

cmd/vic-machine/common/debug.go
```diff
@@ -14,19 +14,24 @@
...
-		cli.IntFlag{
-			Name:        "debug, v",
-			Destination: &d.Debug,
-			Usage:       "[0(default),1...n], 0 is disabled, 1 is enabled, >= 1 may alter behaviour",
-			Hidden:      true,
+		cli.GenericFlag{
+			Name:   "debug, v",
+			Value:  flags.NewOptionalInt(&d.Debug),
+			Usage:  "[0(default),1...n], 0 is disabled, 1 is enabled, >= 1 may alter behaviour",
+			Hidden: false,
```
emlin: vic-machine create should still keep this option hidden

lib/install/management/configure.go
```diff
@@ -99,7 +99,7 @@ func (d *Dispatcher) Configure(vch *vm.VirtualMachine, conf *config.VirtualConta
 	if err = d.update(conf, settings); err == nil {
 		// compatible with old version's upgrade snapshot name
 		if oldSnapshot != nil && (vm.IsConfigureSnapshot(oldSnapshot, ConfigurePrefix) || vm.IsConfigureSnapshot(oldSnapshot, UpgradePrefix)) {
-			d.retryDeleteSnapshot(oldSnapshot.Name, conf.Name)
+			d.retryDeleteSnapshotByRef(oldSnapshot, conf.Name)
```
emlin: please change line 150 as well, and merge two delete snapshot methods

lib/install/validate/validator.go
```diff
@@ -328,9 +328,12 @@ func (v *Validator) ValidateTarget(ctx context.Context, input *data.Data) (*conf
 func (v *Validator) basics(ctx context.Context, input *data.Data, conf *config.VirtualContainerHostConfigSpec) {
 	defer trace.End(trace.Begin(""))
 
+	if input.Debug.Debug != nil && *input.Debug.Debug > 0 {
```
emlin: no need to test value here, and why switch the order here?

pkg/vsphere/vm/vm.go
```diff
@@ -634,3 +634,19 @@ func (vm *VirtualMachine) EnableDestroy(ctx context.Context) {
 		log.Warnf("Failed to enable Destroy_Task for %s: %s", obj[0], err)
 	}
 }
+
+// RemoveSnapshot removes a snapshot by reference
+func (vm *VirtualMachine) RemoveSnapshotByRef(ctx context.Context, snapshot *types.VirtualMachineSnapshotTree, removeChildren bool, consolidate *bool) (*object.Task, error) {
```
emlin: you just need the snapshot reference here

tests/test-cases/Group6-VIC-Machine/6-16-Config.robot
```diff
@@ -18,9 +18,45 @@ Resource  ../../resources/Util.robot
 Suite Setup  Install VIC Appliance To Test Server
 Suite Teardown  Cleanup VIC Appliance On Test Server
 
+*** Keywords ***
+Verify VCH Debug State
+    [Arguments]  ${expected}
+    ${rc}  ${output}=  Run And Return Rc And Output  govc vm.info -e %{VCH-NAME} | grep guestinfo.vice./init/diagnostics/debug
```
emlin: this keyword can be more generic to accept grep parameter, to make it useful for other attribute

tests/test-cases/Group6-VIC-Machine/6-16-Config.robot
```diff
@@ -18,9 +18,45 @@ Resource  ../../resources/Util.robot
...
+    Should Be Equal As Integers  ${rc}  0
+    Should Contain  ${output}  ${expected}
+
+Verify Container Debug State
```
emlin: ditto, so you don't need two keywords

tests/test-cases/Group6-VIC-Machine/6-16-Config.robot
emlin: also return output for additional check

tests/test-cases/Group6-VIC-Machine/6-16-Config.robot
```diff
@@ -18,9 +18,45 @@ Resource  ../../resources/Util.robot
...
+    ${output}=  Run  bin/vic-machine-linux configure --debug 0 --name=%{VCH-NAME} --target=%{TEST_URL} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --password=%{TEST_PASSWORD} --timeout %{TEST_TIMEOUT}
+    Should Contain  ${output}  Completed successfully
+    Verify VCH Debug state  0
+    Verify Container Debug State  ${vm1}  1
+    ${rc}  ${id2}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -itd ${busybox}
+    Should Be Equal As Integers  ${rc}  0
+    ${vm2}=  Get VM display name  ${id2}
+    Verify Container Debug State  ${vm2}  0
+    ${output}=  Run  bin/vic-machine-linux configure --debug 1 --name=%{VCH-NAME} --target=%{TEST_URL} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --password=%{TEST_PASSWORD} --timeout %{TEST_TIMEOUT}
+    Should Contain  ${output}  Completed successfully
```
emlin: verify the old snapshot is deleted here

cmd/port-layer-server/main.go
chengwang86: I will try to merge these two queries.

cmd/port-layer-server/main.go
chengwang86: Thanks. I will merge these two queries.

lib/install/validate/validator.go
chengwang86: good catch 👍

tests/resources/Vsphere-Util.robot
```diff
@@ -221,3 +221,10 @@ Enable Host Firewall
 
 Disable Host Firewall
     Run  govc host.esxcli network firewall set --enabled false
+
+Check VM Guestinfo
+    [Arguments]  ${vm}  ${str}
+    ${rc}  ${output}=  Run And Return Rc And Output  govc vm.info -e ${vm} | grep ${str}
+    Should Be Equal As Integers  ${rc}  0
+    [Return]  ${output}
+
```
emlin: one more empty line?

lib/portlayer/portlayer.go
```diff
@@ -58,6 +61,30 @@ func Init(ctx context.Context, sess *session.Session) error {
 
 	// Grab the storage layer config blobs from extra config
 	extraconfig.Decode(source, &storage.Config)
+
+	logcfg := viclog.NewLoggingConfig()
```
hmahmood: This should not be here. Logging setup should be part of a binary's setup, not at the library level.

cmd/vic-machine/common/debug.go
```diff
@@ -14,19 +14,24 @@
...
+import (
+	"gopkg.in/urfave/cli.v1"
+
+	"github.com/vmware/vic/pkg/flags"
+)
 
 type Debug struct {
-	Debug int `cmd:"debug"`
+	Debug *int `cmd:"debug"`
+	IsSet bool
```
hmahmood: Why do we need this? Should just checking `Debug != nil` be enough?

chengwang86: Make sense. Thank 👍

cmd/vic-machine/common/debug.go
```diff
@@ -14,19 +14,23 @@
...
 
 func (d *Debug) DebugFlags() []cli.Flag {
 	return []cli.Flag{
-		cli.IntFlag{
-			Name:        "debug, v",
-			Destination: &d.Debug,
-			Usage:       "[0(default),1...n], 0 is disabled, 1 is enabled, >= 1 may alter behaviour",
-			Hidden:      true,
+		cli.GenericFlag{
+			Name:   "debug, v",
```
hmahmood: I thought `--debug` was already an option flag, so why do we need to change this?

chengwang86: Are you asking why I changed it from `IntFlag` to `GenericFlag`? I did this because `IntFlag` does not work with `*int`

PR #5323, opened 2017-06-01
[specific ci=Group6-VIC-Machine] update http proxy in vic-machine configure

### LINKED ISSUE #4531: vic-machine reconfigure: Update or reset the HTTP and HTTPS proxy values
**User Statement:**

As a VIC operation user, I need the ability to update or reset the HTTP and HTTPS proxy values. The proxy is used only when pulling images.

**Details:**
- supported operations:
  - update (or add) the HTTP and HTTPS proxy values
  - remove existing HTTP and HTTPS proxy values
- hot update:
- input/output data format:
  - input:  `--https-proxy http://proxy_server_address:port`. The `proxy_server_address` could be either an FQDN or an IP address.
  - output: N/A
- impact on VCH/containerVMs:
  - VCH: The VCH will use the updated proxy server to pull images.  
  - containerVMs: no. Note that configuring a VCH to use a proxy server does not configure proxy support on the containers that this VCH runs. Container developers must configure proxy servers on containers when they create them.
- is upgrade needed: 

**Acceptance Criteria:**
- add the ability to update the HTTP/HTTPS proxy values.
- add relevant test cases

emlin:
fixes #4531


cmd/vic-machine/configure/configure.go
```diff
@@ -109,6 +123,38 @@ func (c *Configure) processParams() error {
...
+			hProxy = c.HTTPProxy.String()
+		}
+		sProxy := ""
+		if c.HTTPSProxy != nil {
+			sProxy = c.HTTPSProxy.String()
+		}
+		personaSession := o.ExecutorConfig.Sessions[config.PersonaService]
+		updateSessionEnv(personaSession, config.GeneralHTTPProxy, hProxy)
+		updateSessionEnv(personaSession, config.GeneralHTTPSProxy, sProxy)
+		vicAdminsession := o.ExecutorConfig.Sessions[config.VicAdminService]
```
anchal-agrawal: s/vicAdminsession/vicAdminSession

lib/install/data/data.go
```diff
@@ -190,4 +190,8 @@ func (d *Data) CopyNonEmpty(src *Data) {
 		d.ContainerNetworks = src.ContainerNetworks
 	}
 	d.Timeout = src.Timeout
+	if src.HTTPProxy != nil || src.HTTPSProxy != nil {
+		d.HTTPProxy = src.HTTPProxy
```
anchal-agrawal: Just for my understanding: is it ok to update both `d.HTTPProxy` and `d.HTTPSProxy` if either of them are non-nil?

tests/test-cases/Group11-Upgrade/11-05-Configure.md
```diff
@@ -0,0 +1,16 @@
+Test 11-05 - Configure
+=======
+
+#Purpose:
```
anchal-agrawal: Minor: a space between `#` and the heading text renders the heading markdown correctly.

lib/install/data/data.go
emlin: hmm... it's a good question.
we're taking set of configuration to replace the old set of configuration if any one of them appear in the options, for example to switch network configuration from static ip back to dhcp.
It's hard to know how user feels about that.
For this specific issue, is there anyone using different proxy server for http and https? If that's not the common case, I guess this definition is fine.

cmd/vic-machine/configure/configure.go
```diff
@@ -109,6 +123,38 @@ func (c *Configure) processParams() error {
...
+	if c.proxies.IsSet {
+		hProxy := ""
+		if c.HTTPProxy != nil {
+			hProxy = c.HTTPProxy.String()
+		}
+		sProxy := ""
+		if c.HTTPSProxy != nil {
+			sProxy = c.HTTPSProxy.String()
+		}
+		personaSession := o.ExecutorConfig.Sessions[config.PersonaService]
```
anchal-agrawal: Would it be better to initialize `personaSession` and `vicAdminSession` outside the `if` block in case they're needed by code for other reconfigure items?

cmd/vic-machine/configure/configure.go
```diff
@@ -109,6 +123,38 @@ func (c *Configure) processParams() error {
...
+		updateSessionEnv(vicAdminSession, config.VICAdminHTTPProxy, hProxy)
+		updateSessionEnv(vicAdminSession, config.VICAdminHTTPSProxy, sProxy)
+	}
+}
+
+func updateSessionEnv(sess *executor.SessionConfig, envName, envValue string) {
+	envs := sess.Cmd.Env
+	var newEnvs []string
+	for _, env := range envs {
+		if strings.HasPrefix(env, envName) {
```
anchal-agrawal: Is checking for the prefix enough or should we check for equality?

cmd/vic-machine/common/proxy.go
```diff
@@ -0,0 +1,70 @@
...
+	HTTPSProxy *string
+	HTTPProxy  *string
+	IsSet      bool
+}
+
+func (p *Proxies) ProxyFlags(hidden bool) []cli.Flag {
+	return []cli.Flag{
+		// proxies
+		cli.GenericFlag{
+			Name:   "https-proxy, sproxy",
```
andrewtchin: Personal preference, but I say remove the short options for both of these

emlin: hmm.. I just realize these two short name in this work, and tbh, I don't like this name as well. let me find some help on this

PR #5327, opened 2017-06-01
Quarantine tether unit test until we can resolve the race

mhagen-vmware:


hmahmood:
This probably won't be enough to fix the failures since the other tether tests are very similar.

caglar10ur:
trying to reason with the failures, will let you know ASAP I find something

hmahmood:
@caglar10ur we're exiting the attach server before `backchannel()` is called, and `t.conn.conn` is `nil` inside `backchannel()`. It looks like the tests have there own mock backchannel, so using the once in the actual tether is redundant; we need to find a way to somehow not use it.

caglar10ur:
```
Jun  1 2017 20:37:28.974Z DEBUG [ END ] [github.com/vmware/vic/cmd/tether.(*attachServerSSH).channelMux:593] [3.001606ms] attach server channel request handler
Jun  1 2017 20:37:28.974Z DEBUG [BEGIN] [github.com/vmware/vic/cmd/tether.(*attachServerSSH).Stop:157] stop attach server
Jun  1 2017 20:37:28.974Z INFO  Incoming attach channel closed
Jun  1 2017 20:37:28.974Z DEBUG [BEGIN] [github.com/vmware/vic/cmd/tether.(*attachServerSSH).stop:234] stop attach server
Jun  1 2017 20:37:28.974Z INFO  Trying to establish a connection
Jun  1 2017 20:37:28.974Z DEBUG [ END ] [github.com/vmware/vic/cmd/tether.(*attachServerSSH).stop:234] [8.599µs] stop attach server
Jun  1 2017 20:37:28.974Z DEBUG [BEGIN] [github.com/vmware/vic/cmd/tether.backchannel:270] establish tether backchannel
Jun  1 2017 20:37:28.974Z INFO  Stopped test attach server
Jun  1 2017 20:37:28.974Z DEBUG [ END ] [github.com/vmware/vic/cmd/tether.(*attachServerSSH).Stop:157] [40.22µs] stop attach server
Jun  1 2017 20:37:28.974Z DEBUG [ END ] [github.com/vmware/vic/lib/tether.(*tether).cleanup:176] [141.992µs] main tether cleanup
```

if stop and run interleave like the above then we loop back (Incoming attach channel closed), try to establish a new one (Trying to establish a connection). At the same time attachServerSSH.stop returns which calls t.Disabled, cancels the context, closes the connection and sets conn to nil. Just after that the loop calls backchannel with nil conn and test panics. This is a test only problem cause we have a conditional in the loop changing this behavior based on whether we are testing or not 

Following should fix this problem

```
diff --git a/cmd/tether/attach.go b/cmd/tether/attach.go
index 60eed72b..a9e26700 100644
--- a/cmd/tether/attach.go
+++ b/cmd/tether/attach.go
@@ -269,6 +269,10 @@ func (t *attachServerSSH) stop() error {
 func backchannel(ctx context.Context, conn net.Conn) error {
        defer trace.End(trace.Begin("establish tether backchannel"))

+       if conn == nil {
+               return fmt.Errorf("nil connection passed")
+       }
+
        // used for shutting down the goroutine cleanly otherwise we leak a goroutine for every successful return from this function
        done := make(chan struct{})

```

sending a PR with other minor cleanups

caglar10ur:
please see https://github.com/vmware/vic/pull/5337

PR #5337, opened 2017-06-01
Fix the panic'ing tether test

caglar10ur:
Also remove the unneccesarry calls to runtime and fix printing test
names.

For more;

https://github.com/vmware/vic/pull/5327#issuecomment-305619860

### DIFF
--- a/cmd/tether/attach.go
+++ b/cmd/tether/attach.go
@@ -362,6 +362,13 @@ func (t *attachServerSSH) run() error {
 						log.Error(detail)
 						return detail
 					}
+				} else {
+					// A series of unfortunate events can lead calling backchannel with nil when we run unit tests.
+					// https://github.com/vmware/vic/pull/5327#issuecomment-305619860
+					// This check is here to handle that
+					if t.conn.conn == nil {
+						return fmt.Errorf("nil connection")
+					}
 				}
 
 				// wait for backchannel to establish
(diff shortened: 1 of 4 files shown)


caglar10ur:
(and yes that loop if not the whole file needs to be refactored)

cmd/tether/attach.go
```diff
@@ -269,6 +269,13 @@ func (t *attachServerSSH) stop() error {
 func backchannel(ctx context.Context, conn net.Conn) error {
 	defer trace.End(trace.Begin("establish tether backchannel"))
 
+	// A series of unfortunate events can lead calling backchannel with nil when we run unit tests.
+	// https://github.com/vmware/vic/pull/5327#issuecomment-305619860
+	// This check is here to handle that
+	if conn == nil {
```
hmahmood: Can we also check `t.testing` for this?

caglar10ur: I could that if you want but didn't want to spread that disease around :) Your call :)

caglar10ur: How about checking that before calling backchannel? This way we can check for t.testing too otherwise we need to find another way to pass that info to backchannel

caglar10ur: Changed the code to do this

caglar10ur:
@hmahmood you may want to take another look now

PR #5340, opened 2017-06-01
WIP Implement docker logs --since [skip ci]

jzt:
This change implements `docker logs --since`. It's a WIP right now while I write an integration test or two, and it also depends on a govmomi vendor update once https://github.com/vmware/govmomi/pull/728 is merged.


tests/test-cases/Group1-Docker-Commands/1-08-Docker-Logs.robot
```diff
@@ -173,9 +173,6 @@ Docker text logs
     Should Be Equal As Integers  ${rc}  0
 
 Docker logs with timestamps and since certain time
-    ${status}=  Get State Of Github Issue  2539
-    Run Keyword If  '${status}' == 'closed'  Fail  Test 1-8-Docker-Logs.robot needs to be updated now that Issue #2539 has been resolved
-    Log  Issue \#2539 is blocking implementation  WARN
     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull busybox
     Should Be Equal As Integers  ${rc}  0
     ${rc}  ${containerID}=  Run And Return Rc And Output  docker %{VCH-PARAMS} create busybox /bin/sh -c 'a=0; while [ $a -lt 5 ]; do echo "line $a"; a=`expr $a + 1`; sleep 1; done;'
```
mhagen-vmware: the bottom of this test needs to be updated to remove the does not support '--since', but apparently it won't let me comment on code beyond the section it wants me to :)

jzt: Yeah just noticed that, testing some changes now. I plan to add a few more checks to this test as well, to fully exercise `--since`.

tests/test-cases/Group1-Docker-Commands/1-08-Docker-Logs.robot
```diff
@@ -173,19 +173,20 @@ Docker text logs
...
+    Should Not Be Equal As Integers  ${lines}  10
+    ${rc}  ${lines}=  Run And Return Rc And Output  docker %{VCH-PARAMS} logs --since=1h ${containerID} |wc -l
+    Should Be Equal As Integers  ${rc}  0
+    Should Be Equal As Integers  ${lines}  10
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} logs ${containerID} |wc -l
+    Should Be Equal As Integers  ${rc}  0
+    Should Be Equal As Integers  ${output}  10
     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} logs --timestamps ${containerID}
     Should Be Equal As Integers  ${rc}  0
     Should Not Contain  ${output}  container ${containerID} does not support '--timestamps'
```
mhagen-vmware: Can you add tests for:
1. docker logs --since=notarealtime
2. something better than Should not contain - does not support --timestamps?  Can we at least verify that the timestamp looks generally alright?

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -396,7 +397,11 @@ func (handler *ContainersHandlersImpl) GetContainerLogsHandler(params containers
 		tail = int(*params.Taillines)
 	}
 
-	reader, err := container.LogReader(context.Background(), tail, follow)
+	if *params.Since > 0 {
+		since = *params.Since
+	}
+
```
caglar10ur: do we care about this check? if params.Since is 0 than that's OK to as since's default is 0. I think you can remove the if check altogether

lib/iolog/log_reader.go
```diff
@@ -35,68 +35,102 @@ type LogReader struct {
...
+			return nil, err
+		}
+	}
+
+	// decode base64 header
+	hdr, err := base64.StdEncoding.DecodeString(string(ehdr))
+	if err != nil {
+		return nil, err
+	}
+
```
caglar10ur: do we need to check the length of the hdr before using?

tests/test-cases/Group1-Docker-Commands/1-08-Docker-Logs.robot
mhagen-vmware: Ok, if the validation happens on the client side then ignore 1.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
jzt: If `since` isn't specified, it actually comes in from the docker client as `-62135596800` :)

lib/iolog/log_reader.go
jzt: I don't think its necessary, since they are always exactly 16 bytes. If we don't somehow get the whole thing in the previous `Read`, we 
 would (in theory) error during that `for n < encodedHeaderLengthBytes` loop, and should also get an error attempting to base64 decode.

caglar10ur:
LGTM

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
caglar10ur: wat? :)

PR #5341, opened 2017-06-02
Fix the redirect issue #5020 [specific ci=Group9-VIC-Admin]

### LINKED ISSUE #5020: Vic Admin flow forces a re-login after viewing logs
**User Statement:**

As a user of vic admin, I want to be able to view logs without needing to re-login each time, so as to improve my workflow. 

**Details:**
Currently, if a log or live log is clicked, the vic admin page is redirected to a log view. If the back button is clicked, there is a problem with the form re submission because of the authentication process. The user then has to go back to the login page and re-login to get back to the dashboard.

**Acceptance Criteria:**
Either adjust the authentication requirements to allow for the user to click back from the log view, or open the log view in a new window.

willsherwood:
Fixes #5020 

A 303 "see other" redirect is needed instead of 307 "temporary redirect" HTTP response. 303 will redirect only to GET, but 307 will continually try to POST the previous data, leading to the form resubmission page. See https://en.wikipedia.org/wiki/HTTP_303

### DIFF
--- a/cmd/vicadmin/server.go
+++ b/cmd/vicadmin/server.go
@@ -353,7 +353,7 @@ func (s *server) loginPage(res http.ResponseWriter, req *http.Request) {
 		}
 
 		// redirect to dashboard
-		http.Redirect(res, req, "/", http.StatusTemporaryRedirect)
+		http.Redirect(res, req, "/", http.StatusSeeOther)
 		return
 	}
 


willsherwood:
A 303 "see other" redirect is needed instead of 307 "temporary redirect" HTTP response. 303 will redirect only to GET, but 307 will continually try to POST the previous data, leading to the form resubmission page. See https://en.wikipedia.org/wiki/HTTP_303

matthewavery:
@willsherwood be sure and move this to verify since it is now approved by engineering. :)

PR #5350, opened 2017-06-02
Re-establish vsphere event stream [full ci]

### LINKED ISSUE #5343: Loss of vsphere events after re-establishment of dropped SDK connection
**Story**
As a user I expect my VCH to report consistent state
As a user I expect my VCH to report whether status is stale (i.e. no connection to vSphere)

**Details**
We rely on vSphere events to trigger state transitions in the VCH when those transitions are not caused by a VCH operation. This results in the following requirements:
* report in VCH info if the event stream is not currently live, resulting in potentially stale status
* re-establishment of event stream if dropped
* invalidation of recorded state if event stream is dropped

**Acceptance**
- [ ] event collector modified to re-establish event stream after re-authentication
- [ ] test added to confirm the three main points above - use `govc` to delete the active session for the VCH, causing NotAuthenticated errors

**Notes**
Opened as result of log bundle analysis for #4872

hickeng:
The vSphere event collector should be re-established after SDK connection
drops for any reason. This adds a test to check that we see an event for
an out-of-band power operation and that the VCH status for the container
in question is correctly updated.

Using full ci as I've made some neatness changes to a robot resource file. 

Fixes #5343

### DIFF
--- a/lib/portlayer/event/collector/vsphere/collector.go
+++ b/lib/portlayer/event/collector/vsphere/collector.go
@@ -36,6 +36,8 @@ type EventCollector struct {
 	vmwManager *vmwEvents.Manager
 	mos        monitoredCache
 	callback   func(events.Event)
+
+	lastProcessedID int32
 }
 
 type monitoredCache struct {
@@ -48,6 +50,8 @@ func NewCollector(client *vim25.Client, objects ...string) *EventCollector {
 	ec := &EventCollector{
 		vmwManager: vmwEvents.NewManager(client),
 		mos:        monitoredCache{mos: make(map[string]types.ManagedObjectReference)},
+		// initialize to an index that will not be present in a page
+		lastProcessedID: -1,
 	}
 
 	for i := range objects {
@@ -134,16 +138,16 @@ func (ec *EventCollector) Start() error {
 		//
 		// the manager will be closed with the session
 
-		err := ec.vmwManager.Events(ctx, refs, 1, followStream, force, func(_ types.ManagedObjectReference, page []types.BaseEvent) error {
-			evented(ec, page)
-			return nil
-		})
-		// TODO: this will disappear in the ether
-		if err != nil {
-			log.Debugf("Error configuring %s: %s", name, err.Error())
-			return err
+		for {
+			err := ec.vmwManager.Events(ctx, refs, pageSize, followStream, force, func(_ types.ManagedObjectReference, page []types.BaseEvent) error {
+				evented(ec, page)
+				return nil
+			})
+			// TODO: this will disappear in the ether
+			if err != nil {
+				log.Debugf("Error configuring %s: %s", name, err.Error())
+			}
 		}
-		return nil
 	}(pageSize, followStream, force, refs, ec)
 
 	return nil
@@ -159,7 +163,21 @@ func evented(ec *EventCollector, page []types.BaseEvent) {
 		return
 	}
 
+	if len(page) == 0 {
+		return
+	}
+
+	// skip events already seen
+	oldIndex := len(page)
 	for i := range page {
+		if page[i].GetEvent().Key == ec.lastProcessedID {
+			oldIndex = i
+		}
+	}
+
+	// events appear in page with most recent first - need to reverse for sane ordering
+	// we start from the first new event after the last one processed
+	for i := oldIndex - 1; i >= 0; i-- {
 		// what type of event do we have
 		switch page[i].(type) {
 		case *types.VmGuestShutdownEvent,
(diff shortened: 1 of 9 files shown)


lib/install/management/appliance.go
```diff
@@ -599,6 +599,7 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 				"CS_PATH=" + settings.ClusterPath,
 				"POOL_PATH=" + settings.ResourcePoolPath,
 				"DS_PATH=" + conf.ImageStores[0].Host,
+				"VC_KEEPALIVE=20s",
```
chengwang86: I find that we have different keep_alive values in our code here and there. Why do you set it to `20s` here?

lib/portlayer/event/collector/vsphere/collector.go
```diff
@@ -159,7 +163,21 @@ func evented(ec *EventCollector, page []types.BaseEvent) {
 		return
 	}
 
+	if len(page) == 0 {
+		return
+	}
+
+	// skip events already seen
```
dougm: Not sure if we can make use of it in this case, but CreateCollectorForEvents takes a filter param:
http://pubs.vmware.com/vsphere-60/topic/com.vmware.wssdk.apiref.doc/vim.event.EventFilterSpec.html

We already use it filter by entity.  It also has a time field: http://pubs.vmware.com/vsphere-60/index.jsp#com.vmware.wssdk.apiref.doc/vim.event.EventFilterSpec.ByTime.html

Guessing we could set the beginTime to that of the last event timestamp we had processed, and avoid fetching these "stale" events at all.  Would require a govmomi change though: https://github.com/vmware/govmomi/blob/master/event/processor.go#L53

cc @cgtexmex 

But functionally, this should be fine as-is.

lib/install/management/appliance.go
hickeng: This is the keep alive between the port layer and the vSphere environment. Unfortunately at the moment the only means of triggering re-authentication is tied to the keepalive interval.
Ideally we'd switch to something that would trigger after a single operation fails but keepalive was simplest approach when @dougm added this.

lib/portlayer/event/collector/vsphere/collector.go
hickeng: Yeah, the two main facets of this PR need to roll into govmomi at some point:
1. option to eliminate previously collected events - can/should we present a simple channel instead of the page callback? (vmware/govmomi#730)
2. re-auth mechanism not tied to keepalive (vmware/govmomi#729)

hickeng:
Updated to move timeout change into portlayer options default - reduces the number of reviewers by one and there's little reason for it to be in the extraconfig.

lib/portlayer/event/collector/vsphere/collector.go
cgtexmex: @dougm agree - I had a similar thought on Friday but hadn't had time to look into it more.

cgtexmex:
LGTM

tests/resources/VCH-Util.robot
```diff
@@ -219,6 +219,14 @@ Check UpdateInProgress
     Should Be Equal As Integers  ${rc}  0
     Should Contain  ${output}  ${expected}
 
+Portlayer Log Should Match Regexp
+    [Tags]  secret
+    [Arguments]  ${pattern}
+    ${out}=  Run  curl -k -D /tmp/cookies-%{VCH-NAME} -Fusername=%{TEST_USERNAME} -Fpassword=%{TEST_PASSWORD} %{VIC-ADMIN}/authentication
+    Log  ${out}
```
mhagen-vmware: Because of [tags] secret, this will get wiped either way.  For debuggability, we really need to get away from stacking too many calls into a secret keyword, as then we can't really debug anything if it goes wrong.   This would really be the 3rd or 4th place where we are curling the cookies and tagging the entire keyword as secret.  I am happy to clean this up and you can check this in, but please create a tech debt/component/test issue for someone to clean up all these?

tests/test-cases/Group0-Bugs/5343.robot
```diff
@@ -0,0 +1,66 @@
...
+Suite Setup  Install VIC Appliance To Test Server
+Suite Teardown  Cleanup VIC Appliance On Test Server
+
+
+
+*** Test Cases ***
+Check vsphere event stream
+    ${status}=  Get State Of Github Issue  5343
+    Run Keyword If  '${status}' == 'closed'  Fail  Test 5343.robot needs to be updated now that Issue #5343 has been resolved
+    Log  Issue \#5343 is blocking implementation  WARN
```
mhagen-vmware: This needs to be removed.

tests/resources/VCH-Util.robot
hickeng: #5358

PR #5351, opened 2017-06-03
Add vic-machine configure container-network support

### LINKED ISSUE #4527: vic-machine reconfigure: Add the container network settings
**User Statement:**

As a VIC operation user, I need the ability to add or remove the container network settings.

**Details:**
- supported operations:
  - add container network. If the container network `label` has been used by existing container networks, vic-machine throws an error msg and the `add` operation will fail.
- hot update:
- input/output data format:
  - input: `--container-network cont-net-1:net1` (`net1` is the network label)
  - output: N/A
- impact on VCH/containerVMs:
  - VCH: the VCH should be able to allocate new containers which can be connected to the newly added container network.
  - containerVMs: no. 
- is upgrade needed: 

**Acceptance Criteria:**
- add the ability to add the container network settings.
- add relevant test cases

anchal-agrawal:
Fixes #4527


cmd/vic-machine/common/cnetworks.go
```diff
@@ -0,0 +1,270 @@
...
+	ContainerNetworksGateway  cli.StringSlice `arg:"container-network-gateway"`
+	ContainerNetworksIPRanges cli.StringSlice `arg:"container-network-ip-range"`
+	ContainerNetworksDNS      cli.StringSlice `arg:"container-network-dns"`
+	IsSet                     bool
+}
+
+// CNetworkFields holds items returned by ProcessContainerNetworks for each
+// container network.
+type CNetworkFields struct {
+	VICNet   string
```
emlin: is this the net alias? VICNet is kind of confusing

cmd/vic-machine/configure/configure.go
```diff
@@ -92,10 +93,11 @@ func (c *Configure) Flags() []cli.Flag {
 	compute := c.ComputeFlags()
 	iso := c.ImageFlags(false)
 	debug := c.DebugFlags()
+	cNetwork := c.cNetworks.CNetworkFlags(false)
 
 	// flag arrays are declared, now combined
 	var flags []cli.Flag
-	for _, f := range [][]cli.Flag{target, id, compute, iso, proxies, util, debug} {
+	for _, f := range [][]cli.Flag{target, id, compute, iso, proxies, util, debug, cNetwork} {
```
emlin: the command help will show options in the slice order, so it will be better to keep it's original order, instead of add all network options in the end.

cmd/vic-machine/common/cnetworks.go
```diff
@@ -0,0 +1,270 @@
...
+		cli.StringSliceFlag{
+			Name:  "container-network, cn",
+			Value: &c.ContainerNetworks,
+			Usage: "vSphere network list that containers can use directly with labels, e.g. vsphere-net:backend. Defaults to DCHP - see advanced help (-x).",
+		},
+		cli.StringSliceFlag{
+			Name:   "container-network-gateway, cng",
+			Value:  &c.ContainerNetworksGateway,
+			Usage:  "Gateway for the container network's subnet in CONTAINER-NETWORK:SUBNET format, e.g. vsphere-net:172.16.0.1/16",
+			Hidden: hidden,
```
emlin: the advanced container network options are all hidden in vic-machine create, because we have '-x' option to show all advanced options, but in configure, we're not supporting all create options at this time, it might be better to keep all of them open atm.
You can allow the method to accept one option, to make it different in create and configure command

cmd/vic-machine/common/cnetworks.go
anchal-agrawal: Yes, this is the Docker alias (label) for the vSphere network. I chose `VICNet` because it was named like that previously, but I can change it.

cmd/vic-machine/common/cnetworks.go
anchal-agrawal: This func takes the `hidden` parameter, which is set to `false` for configure and `true` for create. The `--container-network` flag is always non-hidden. I think that's what we intend to do.

cmd/vic-machine/create/create.go
```diff
@@ -506,10 +477,11 @@ func (c *Create) Flags() []cli.Flag {
 	iso := c.ImageFlags(true)
 	debug := c.DebugFlags()
 	proxies := c.proxies.ProxyFlags(true)
+	cNetwork := c.containerNetworks.CNetworkFlags(true)
 
 	// flag arrays are declared, now combined
 	var flags []cli.Flag
-	for _, f := range [][]cli.Flag{target, compute, create, proxies, syslog, iso, util, debug, help} {
+	for _, f := range [][]cli.Flag{target, compute, create, proxies, syslog, iso, util, debug, help, cNetwork} {
```
emlin: sorry the order comment should be in here 😄

anchal-agrawal: Good catch!

cmd/vic-machine/common/cnetworks.go
anchal-agrawal: Changed to `NetAlias` and added a comment.

cmd/vic-machine/common/cnetworks.go
emlin: cool. 👍

cmd/vic-machine/common/cnetworks.go
```diff
@@ -0,0 +1,272 @@
...
+	ContainerNetworks         cli.StringSlice `arg:"container-network"`
+	ContainerNetworksGateway  cli.StringSlice `arg:"container-network-gateway"`
+	ContainerNetworksIPRanges cli.StringSlice `arg:"container-network-ip-range"`
+	ContainerNetworksDNS      cli.StringSlice `arg:"container-network-dns"`
+	IsSet                     bool
+}
+
+// CNetworkFields holds items returned by ProcessContainerNetworks for each
+// container network.
+type CNetworkFields struct {
```
emlin: Could you use data.ContainerNetworks directly here?

anchal-agrawal: `data.ContainerNetworks` holds fields that we use to store and validate networks, while the fields in `CNetworkFields` are used for storing the command's input.

emlin: this is the data after processing, and data.ContainerNetworks is for the same purpose. That's why I'm wondering if you can use data.ContainerNetworks directly

lib/install/data/data.go
```diff
@@ -169,10 +173,74 @@ func NewData() *Data {
...
 }
 
-// CopyNonEmpty will shallow copy src value to override existing value if the value is set
-// This copy will take care of relationship between variables, that means if any variable in ContainerNetwork or
-// NetworkConfig is not empty, the whole ContainerNetwork or NetworkConfig will be copied
-func (d *Data) CopyNonEmpty(src *Data) {
+// areEqualIPRanges checks if two ip.Range slices are equal by using
+// reflect.DeepEqual, with an extra check that returns true when the
+// lengths of both slices are zero.
+func areEqualIPRanges(a, b []ip.Range) bool {
```
emlin: areEqual... => equal... ?

anchal-agrawal: Fixed!

cmd/vic-machine/common/cnetworks.go
anchal-agrawal: Oh, I see. I'm using `CNetworkFields` here because `common.ProcessContainerNetworks()` has a `*CNetworks` receiver, so it can be used by configure and create.

anchal-agrawal: Fixed! Please see 4e0a93f.

cmd/vic-machine/common/cnetworks.go
chengwang86: @emlin This is useful. Do you think if we should apply this mechanism for all the options that have different visibility for `create` vs. `configure`?

PR #5360, opened 2017-06-05
Support xfs as a filesystem

casualjim:
fix a race condition in trace pkg test

These are changes I made in the kubernetes version of the vsphere package, want to keep these 2 in sync so that we can keep updating as changes are made in VIC

### DIFF
--- a/pkg/fs/xfs.go
+++ b/pkg/fs/xfs.go
@@ -1,0 +1,84 @@
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fs
+
+import (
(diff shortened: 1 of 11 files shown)


sgairo:
Copyright update needed in some of the files :)

pkg/fs/xfs.go
```diff
@@ -0,0 +1,84 @@
...
+		return err
+	}
+	log.Debugf("Filesystem created on device %s", devPath)
+
+	return nil
+}
+
+// Mount mounts an xfs formatted device at the given path.  From the Docker
+// mount pkg, args must in the form arg=val.
+func (e *XFS) Mount(devPath, targetPath string, options []string) error {
```
jzt: I'm curious to know if we couldn't just leave `Mount`/`Unmount`/`SetLabel` in `disk.go` and pull them out of `pkg/fs/ext4.go` and `pkg/fs/xfs.go` so there's a little less DRYness. None of them seem to require the receiver so I'm wondering if they can't just be consolidated.

hickeng: I agree in principle because of the duplicated code, however it's quite possible that different filesystems have very different options. While the only difference in the call here is the filesystem type "xfs", NTFS for example has a significantly different set of extended options.

jzt: Yeah I just noticed that `disk.go:Mount` calls `d.fs.Mount` which would be the (potentially) specialized version unique to the fs.

caglar10ur:
can we merge this @casualjim? I'm touching the same set of files and would love to fix the conflicts before I add more code :)

caglar10ur:
@jzt @casualjim ping, I really think we need to merge this first and then @jzt needs to rebase ref count changes on top of this

PR #5374, opened 2017-06-05
Refactor run loop a bit to make it shorter

caglar10ur:
While at it drop darwin and windows support from tether. Having them is
pointless at this time as we don't support those. We also didn't do
a good job of keeping them up2date so they are not in sync with the rest
of the code which makes them useless too.

Also start using /dev for device nodes and drop the createDevices support

Includes #5300 for now, will rebase on top of that once it get merged


Makefile
```diff
@@ -356,11 +342,11 @@ $(appliance): isos/appliance.sh isos/appliance/* isos/vicadmin/** $(rpctool) $(v
 	@$(TIME) $< -p $(appliance-staging) -b $(BIN)
 
 # main bootstrap target
-$(bootstrap): isos/bootstrap.sh $(tether-linux) $(rpctool) $(bootstrap-staging) isos/bootstrap/*
+$(bootstrap): isos/bootstrap.sh $(tether) $(rpctool) $(bootstrap-staging) isos/bootstrap/*
```
hickeng: don't need rpctool in the bootstrap any more.

cmd/tether/attach_linux.go
```diff
@@ -90,41 +90,17 @@ func rawConnectionFromSerial() (net.Conn, error) {
...
-
-func (t *attachServerSSH) Start() error {
-	defer trace.End(trace.Begin(""))
-
-	t.m.Lock()
-	defer t.m.Unlock()
-
-	var err error
-
-	rand.Reader, err = os.Open(pathPrefix + "/urandom")
```
hickeng: unsure about this - I vaguely recall having a scenario where /dev/urandom wasn't available. Think that was in the bville/tclinux days but ...

isos/bootstrap.sh
```diff
@@ -81,7 +81,7 @@ else
 fi
 
 # copy in our components
-cp ${BIN}/tether-linux $(rootfs_dir $PKGDIR)/bin/tether
```
hickeng: please continue to call this tether-linux. We still have an intent to add other OS support.

Makefile
caglar10ur: Done

cmd/tether/attach_linux.go
caglar10ur: https://golang.org/src/crypto/rand/rand.go says ```On Linux, Reader uses getrandom(2) if available, /dev/urandom otherwise.``` so I think we are safe :)

README.md
```diff
@@ -86,7 +86,7 @@ To build the component binaries, ensure `GOPATH` is set, then issue the followin
 ```
 $ make components
 ```
-This will install required tools and build the component binaries `tether-windows`, `tether-linux`, `rpctool` and server binaries `docker-engine-server`, `port-layer-server`.  The binaries will be created in the `$BIN` directory, ./bin by default.
+This will install required tools and build the component binaries `tether`, `rpctool` and server binaries `docker-engine-server`, `port-layer-server`.  The binaries will be created in the `$BIN` directory, ./bin by default.
```
hickeng: tether-linux

cmd/tether/attach_linux.go
hickeng: That code would have been written to support much earlier version of Go as well, so possible the runtime has updated to fix the issue it was addressing.

README.md
caglar10ur: done

cmd/tether/attach.go
```diff
@@ -324,6 +328,59 @@ func backchannel(ctx context.Context, conn net.Conn) error {
...
+	// tests are passing their own connections so do not create connections when testing is set
+	if !t.testing {
+		// close the connection if required
+		if t.conn.conn != nil {
+			t.conn.conn.Close()
+			t.conn.conn = nil
+		}
+		t.conn.conn, err = rawConnectionFromSerial()
+		if err != nil {
+			detail := fmt.Errorf("failed to create raw connection: %s", err)
```
anchal-agrawal: Nit: `errDetail`

cmd/tether/attach.go
```diff
@@ -324,6 +328,59 @@ func backchannel(ctx context.Context, conn net.Conn) error {
...
+		// This check is here to handle that
+		if t.conn.conn == nil {
+			return fmt.Errorf("nil connection")
+		}
+	}
+
+	// wait for backchannel to establish
+	err = backchannel(t.ctx, t.conn.conn)
+	if err != nil {
+		detail := fmt.Errorf("failed to establish backchannel: %s", err)
```
anchal-agrawal: `errDetail`

PR #5376, opened 2017-06-05
[specific ci=Group11-Upgrade]workaround the upgrade issue reported in 5343

emlin:
the only thing possible is session datacenter or cluster is nil, but without any vic-machine log, cannot know what caused that.
this is to workaround the possible nil reference, and add some more logging to see what happens


chengwang86:
Lgtm

emlin:
more log in vic-machine upgrade #5379

PR #5378, opened 2017-06-06
Restore the original uid/gid of the mount point

### LINKED ISSUE #5377: Volumes are mounted as root:root instead of their original uid:gid
And because of that some applications are failing to start, seen in jenkins due to this line https://github.com/jenkinsci/docker/blob/6a74b7da7ef84a36679fdb39a6f94e1d75b4054f/Dockerfile#L53

tether should do a chown after a successful mount as ext4 doesn't support uid/gid options

caglar10ur:
Fixes #5377


lib/tether/ops_linux.go
```diff
@@ -721,6 +724,19 @@ func (t *BaseOperations) MountLabel(ctx context.Context, label, target string) e
 		return errors.New(detail)
 	}
 
+	// change the ownership of the target directory to the original uid/gid
```
hmahmood: How will this affect nfs mounts? @matthewavery

caglar10ur: I believe this function is for local disks only

Issue #5321, opened 2017-06-01
Create an agreed upon api endpoint definition for docker cp portlayer work


matthewavery:
#For stories, please include the information below:

**User Statement:**
This is ticket is designed to allow parallel work on  the personality as well as the portlayer with minimal stepping on toes. It simply requests a planned out implementation of the swagger spec definition of `docker cp` 

**Details:**
There are two modes for `docker cp` 

1. Copying from the client to the container. (mode 1)

2. Copying from the container to the client.  (mode 2) 

These two modes are fundamentally opposites and should be different endpoints. 

I propose that we create a `/container/filesystem?ID=<container ID>&overwrite=<overwrite directory flag>`(this definition likely needs more fleshing out. hence the ticket) endpoint for the set of api calls. 

Further more I believe that `mode 1` should be a `PUT` request endpoint.  `Mode 2` should be a `GET` call on that endpoint. Each will have its own response and request modules. 

Things that need to be communicated between the personality and portlayer:

* container ID
* tar stream of file contents
* permissions(?)

note: for permissions we should likely relay that from the personality as well in case there are other personalities that actually let you set permissions on copy from the CLI. If we do not think this will be the case then we should do this at portlayer time instead.

**Acceptance Criteria:**
A well thought out and planned swagger spec portlayer api endpoint is designed. This endpoint is functional and complete for the people working on both the personality and the portlayer.

hickeng:
This should have been captured in the design doc that @matthewavery merged. I intensely dislike the approach of starting with swagger for API design as it generally produces poor code structure for the resulting implementation; as such this is done with Go interfaces that will be exposed and an example rest handler at the end:

The following are the core interfaces that allow hiding of the current usage state of a given storage element such as a volume:
```go
// DataSource defines the methods for exporting data from a specific storage element as a tar stream
type DataSource interface {
	// Close releases all resources associated with this source. Shared resources should be reference counted.
	io.Closer

	// Export performs an export of the specified files, returning the data as a tar stream. This is single use; once
	// the export has completed it should not be assumed that the source remains functional.
	//
	// spec: specifies which files will be included/excluded in the export and allows for path rebasing/stripping
	// data: if true the actual file data is included, if false only the file headers are present
	Export(op trace.Operation, spec *archive.FilterSpec, data bool) (io.ReadCloser, error)

	// Source returns the mechanism by which the data source is accessed
	// Examples:
	//     vmdk mounted locally: *os.File
[...]

Issue #4872, opened 2017-04-25
cVM powered off, but container appears as running during ps -a


malikkal:
engine version: v1.1.0-9852-e974a51

Steps to reproduce.

`docker -H ip:port --tls run -it --cpuset-cpus 2 --memory 4g --name <name> --net <network> registry/project/hello-world`

Ctrl-c to exit the shell, which will power off the cVM, followed by a docker ps will display the container as running. docker rm or rm -f will fail with:

`You cannot remove a running container. Stop the container before attempting removal or use -f`

delete the cVM from vSphere. I am aware, this is not recommended; followed by docker ps will display them as **still** running and a docker stop will result in:

`No such container: ID`

power off the vAPP and power on will result in correct output for docker ps. however, redeploying the same image will result in.

`Server error from portlayer: Invalid configuration for device '1'.`

malikkal:
...pardon me. I couldn't upload the logs here, as it was not tested in the dev environment.  I have created a directory github4872 in the ftpsite and uploaded the logs there.

hickeng:
@malikkal The latter part of the problem (invalid configuration for device '1') is #2928. The incorrect state reporting requires investigation.

@dbarkelew Is there a KB in-progress for the out-of-band delete? I know @chengwang86 wrote the docs from the engineering side, but don't know where it stands in the GSS/publication path.
Could you also move the ftpsite files into a PR so dev can access them.

hickeng:
Estimate is for initial investigation of the container state aspect - I expect the estimate to be updated once more knowledge is available.

sflxn:
@malikkal Can I get some more info?  This hello-world container is your own or docker's?  I ask because the hello-world container I am familiar with only dumps out a greeting and some info and takes you back to your original shell so -it will have no effect since it doesn't put you into another shell.  I must assume you're running another container.  Can you provide me some info on how to get the one you're using?

I also tried this with just busybox, which puts me into sh, but ctrl-c cannot break me out of the container's sh shell.

hickeng:
@malikkal Please can you open an SR with GSS to get us the logs. It turns out that there's automation hooked into that process that makes logs uploaded to ftpsite available internally, I assume to enforce information access controls around customer confidential logs. I've not found a way to access those files directly from the ftpsite.
[...]

Issue #4105, opened 2017-03-01
integrate toolbox filemanager support into tether


hickeng:
**Story**
As a VIC developer I want tether to support transfer of files/directories to and from the guest

**Detail**
This should build on #4103 to allow transfer of files to and from the guest. This needs to be able to handle directory copy as well as singular files, so is likely to involve a mime type and tar/untar.

**Acceptance**
Able to roundtrip binary file to/from containerVM
Able to copy directory from containerVM
Able to copy directory to containerVM and retrieve single file from containerVM

dougm:
Part 1 is in review (#5436), which completes the guest FileManager methods.  I'll implement the auto tar/untar in a separate PR.  cc @matthewavery @sflxn

PR #5397, opened 2017-06-07
[skip ci] WIP Official Docker CP Design Doc

matthewavery:
This PR is the place for us to come to a consensus on the approach, architecture, and concerns that need to be tackled for the implementation of `docker cp` to come to fruition. 

This will likely be quite fluid at first, but should end with a solid operationalized approach to getting the work done. 

fixes: #5321

### DIFF
--- a/doc/design/docker-cp.md
+++ b/doc/design/docker-cp.md
@@ -1,0 +1,303 @@
+## DOCKER CP DESIGN DOC
+
+### Initial Rundown of docker cp
+
+Docker cp has many possible scenarios that we will have to account for and they will be described in this section.
+
+We will be supporting docker cp in it's entirety. This means we will need to copy files to and from a container regardless of whether it is on or off. This will lead to 4 scenarios. 
+
+1. The container is *on*, and we want to copy data *from* it.
+
+2. The container is *on*, and we want to copy data *to* it.
+
+3. The container is *off*, and we want to copy data *from* it.(will constitute multiple calls from the personality)
+
+4. The container is *off*, an we want to copy data *to* it.(will constitute multiple calls from the personality)
+
+According to the docker remote api the target for a copy must be in `identity (no compression), gzip, bzip2, xz`. As the docker cli handles packaging a *to* request we can expect on of these formats. We must package our *from* response in one of these formats as well. 
+
+ _NOTE_: docker has discussed copying between containers in the past. This will be yet another set of 4 possibilities and endpoints. Something that we can add to this plan later if it is needed in the future. 
+
+Here is a more complex look at the call operation/state flow.
+
+|Operation| state| volumes | scenario |
+|---|---|---|---|
+|  Export | ContainerRunning | No | Single call from personality to portlayer. Guest Tools will be used on the target rather than `Read` |
+| Export | ContainerRunning | Yes |  Single call from personality to portlayer. Guest Tools will be used on the target rather than `Read` |
+| Export | ContainerStopped | No | Single call from personality to portlayer. `Read` is called based on the supplied filespec. |
+| Export | ContainerStopped | Yes | Multiple calls from personality to portlayer. One for the r/w layer, and n more calls where n is the number of volumes. `Read` is invoked n+1 times based on each call. Powerstatus likely won't matter here because we mount non-persistent disks|
+| Import |  ContainerRunning | No | Single call from personality to portlayer. Guest tools will be used on the target rather than a `Write` |
+|  Import | ContainerRunning | Yes |  Single call from personality to portlayer. Guest tools will be used on the target rather than a `Write` |
+| Import | ContainerStopped |  No | Single call from personality to portlayer. `Write` will be used to mount the r/w layer and then write the contents based on the supplied filespec  |
+| Import | ContainerStopped | Yes |  Multiple calls from personality to portlayer. One for the r/w layer, and n more calls where n is the number of volumes. `Write` is invoked n+1 times based on each call from the personality. If the container is started during this time we cannot mount the volumes or the r/w layer and we will report a failure requesting the user to try again. Alternatively, we block start events until operation completetion. |
+
+
+### Personality to Portlayer Communication
+
+Since all 4 of these must be supported we will require both distinctions for when the container is off and which endpoint to call to distinguish between *to* and *from* operations. 
+
+The proposed solution is to have two endpoints designed for the portlayer api swagger server. Both endpoints will exist along the same request path as two different verbs. The request path should be as such : 
+
+The portlayer functional target for `ExportArchive` and `ImportArchive` will need to be called multiple times from the personality for each device that constitutes the full filesystem of the container. A `PathSpec` struct will be used to mark inclusion and exclusion paths. We will also need to know when to strip the paths provided to the portlayer since the view of a volume would be `/` while in a running container the target path could be `/path/to/volume/` where the path before the final `/` exists on the container r/w layer(or another volume, but we are not worrying about that now.) Because there will be multiple calls for the potential Import of the archive(write calls) we will need to have a way to pass or reference the stream of information to the portlayer multiple times. This may constitute another input for the function header in order to identify the stream that we care about. 
+
+the portlayer functional targets will look as such: 
+
+__ExportArchive__
+
+```
+// ID : container/vm target ID mainly for power state checking in the portlayer to determine Exports logical behavior.
+// deviceID : target deviceID for copying
+// fileSpec : just a map[string]string of keyed paths to exclusion and strip operations determined by the desired behavior of the read
+// 
+func Read(ID, deviceID, filespec map[string]string) (io.reader, error)
+
+// no need for the data bool since we know that this operation involves data. we can set that as always true on the portlayer side of things.
+```
+
+
+__ImportArchive__
+```
+// StoreID : ID of the store where the target device resides be it image store, volume store, or something else in the future.
+// ID : target container/vm for the portlayer to check powerstate to determine the logical approach of the ImportArchive function
+// deviceID : device id found in the targeted store that is to receive the archive
+// fileSpec : just a map[string]string of keyed paths to exclusion and strip operations determined by the desired behavior of the write
+// tarStream : the actual tar stream to be imported to the target
+func Write(ID, deviceID string, fileSpec map[string]string, tarStream io.Writer) error
+```
+
+Note: this call will involve multiple docker endpoints, Stat is needed with we plan to support the `-L` functionality. We will also need the interaction piece for streaming tar archive's to the portlayer with the same tar stream from docker being streamed to multiple calls(hopefully we might have a way to not stream the same data several times over the network).
+
+## Personality design
+
+
+### Copy Operation
+
+The personality has Three endpoints that are called depending on the to/from situation above(5 if you count the two deprecated calls that were made previously.). The portlayer will divide the behavior of to/from as two separate endpoints. These endpoints will behave in an `Export/Import` behavior. Aside from the Import/Export behavior the docker cli also calls `ContainerStatPath` when determining whether to follow symlinks via the `-L` option. We will need this endpoint implemented all the way through to the portlayer if we want to support following symlinks.
+
+Error codes expected by docker : 
+
+```
+200 
+
+400
+
+403 - in the event of a readonly filesystem
+
+404
+
+409
+
+500
+
+```
+
+Docker does not do much in the way of string checking as far as the path is concerned. This is likely not a big concern for the personality as the actual copy command will fail if a bad path string is provided. 
+
+For the multiple calls to the portlayer we will need to assemble Pathspecs which detail inclusion, exclusion, rebase,and strip configurations for the tar stream on both reads and writes. Below will be some scenarios and what the path spec should look like for those scenarios.
+
+Exclude the tether path. Exclude any paths that docker does not allow for a copy. the filespec map with a serialized struct or keys for `exclusion`, `strip`.
+
+examples of a pathspec(please note that these are simple scenarios for the time being):
+```
+For a CopyTo to path "/volumes" :
+container has one volume mounted at /volumes/data
+
+this will invoke two write calls. 
+
+first pathspec: 
+
+// this pathspec would be empty since you would be writing against the r/w of the containerFS and the starting path would be '/' of the containerFS. No stripping of the write headers necessary.
+spec.rebase="volumes"
+spec.strip = ""
+spec.includes = map[string]string {"": struct{}{}}
+spec.excludes = map[string]string {"volumes/data": struct{}{}}
+
+Second Pathspec:
+spec.Strip = "volumes/data/"  // stripped since the volume will be mounted and the starting path will be "/"
+spec.Rebase = ""
+spec.includes = map[string]string {"": struct{}{}}
+}
+```
+
+```
+For a Copy from path "/volumes/" :
+container has one volume mounted at /volumes/data
+
+this will invoke two read calls
+
+// this pathspec would be empty since you would be writing against the r/w of the containerFS and the starting path would be '/' of the containerFS. No rebase necessary for the headers.
+spec.Rebase = "volumes"
+spec.Strip = "volumes"
+// in this case strip and rebase are the same since this is a 1st level directory e.g for a target of /volumes/data rebase would be "data" and strip would be "volumes/data" 
+spec.include = map[string]string {"volumes":struct{}{}}
+spec.Exclude = map[string]string {"": struct{}{}}
+
+// this pathspec will be for the volume filesystem
+spec.Rebase = "volumes"
+spec.Strip = ""
+spec.include = map[string]string {"":struct{}{}}
+
+```
+
+
+NOTE: use TarAppender in docker/pkg/archive to possibly merge the different tar streams during an ExportArchive. 
+
+### Stat Operation
+
+
+
+## Portlayer Design
+
+
+
+ There will be two situations to be concerned with in the portlayer. When the container is *on* and when the container is *off*.
+ 
+ in the event of the *off* scenario we will convey the requested operations to the `Read`/`Write` calls in the storage portion of the portlayer. The results of those calls should be returned to the user.
+ 
+ Regardless of whether the container is on or off. Some investigation has shown that it works in the scenario of both containers being on, off, and one on and one off. We will need to architect a way for this to work for us. If the target is a volume that is attached to a container we will need to understand what is needed in the case of a vmdk volume target. This behavior should not be an issue for the nfs volume store targets.
+ 
+ The portlayer should behave at a basic level as such: 
+ 
+ ```
+ 1. look up appropriate storage object
+2. check object usage
+3. mount object or initiate copy based on (2)
+3b. check for error in initialization of copy due to power state change, repeat (2) if found
+4. mounted disk prevents incompatible operations, or, tether copy in progress squashes in guest shutdown
+ ```
+ 
+ some additiona notes surrounding the portlayer and swagger endpoint design:
+ 
+
+use query string to pass FilterSpec so that it's not needed to be packaged in a custom format in body along with the tar
+this assumes that callers have some knowledge of mount topology and correctly populate the FilterSpec to avoid recursing into mounts that will be directly addressed by a later call.
+because it's possible for the various storage elements of a container to be in different states (e.g. container is not running so r/w and volumeA are not in use, but volumeB from the container is now used by container X) we require that separate calls be made to Import/Export for each of the distinct storage elements. This allows those calls to be routed appropriately based on current owners and usage of the element in question.
+ 
+ 
+### Stat Operation
+
+We will need a `stat` endpoint which will be used to target devices with a filesystem that can return fileinfo back to the caller. This will need to be tolerant of the power status of the target, if the target container/vm is on then we can use guest tools. Otherwise, mounting the appropriate device for the stat is necessary. Like the Read/Write calls we should expect a store ID and a device ID in addition to the target of the stat operation. if the compute is not active then we should mount the target device specified for the stat. 
+
+```
+underlying filesystem stat :
+// storename : this is the store that the target device resides on 
+// deviceID : this is the target device for the stat operation
+// target : is the filesys
+//
+// stat will mount the target and stat the filesystem target in order to obtain the requested filesystem info.
+func Stat(storename string, deviceID string, spec archive.FilterSpec) (FileInfo, error)
+
+// it is the responsibility of the caller to determine the status of this device. If it is already mounted or in use the caller must determine the action to be taken. 
+```
+
+ 
+### Portlayer Import/Export Behavior for writing and reading to storage devices.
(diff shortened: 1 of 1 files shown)


doc/design/docker-cp.md
```diff
@@ -0,0 +1,144 @@
...
+ _NOTE_: docker has discussed copying between containers in the past. This will be yet another set of 4 possibilities and endpoints. Something that we can add to this plan later if it is needed in the future. 
+
+### Personality to Portlayer Communication
+
+Since all 4 of these must be supported we will require both distinctions for when the container is off and which endpoint to call to distinguish between *to* and *from* operations. 
+
+The proposed solution is to have two endpoints designed for the portlayer api swagger server. Both endpoints will exist along the same request path as two different verbs. The request path should be as such : 
+
+```
+/container/filesystem?ID=
```
hickeng: The ID is packaged in a path element for all of the other container operations, not in the query string.

doc/design/docker-cp.md
```diff
@@ -0,0 +1,144 @@
...
+
+According to the docker remote api the target for a copy must be in `identity (no compression), gzip, bzip2, xz`. As the docker cli handles packaging a *to* request we can expect on of these formats. We must package our *from* response in one of these formats as well. 
+
+ _NOTE_: docker has discussed copying between containers in the past. This will be yet another set of 4 possibilities and endpoints. Something that we can add to this plan later if it is needed in the future. 
+
+### Personality to Portlayer Communication
+
+Since all 4 of these must be supported we will require both distinctions for when the container is off and which endpoint to call to distinguish between *to* and *from* operations. 
+
+The proposed solution is to have two endpoints designed for the portlayer api swagger server. Both endpoints will exist along the same request path as two different verbs. The request path should be as such :
```
hickeng: Generally we should be designing the Go API for operations first, otherwise we end up implementing REST style operations in Go which is an awkward fit. REST is just the way in which we're currently exposing the API, I think we'd all be happy to have a GRPC endpoint as well/instead.

doc/design/docker-cp.md
```diff
@@ -0,0 +1,144 @@
...
+
+* ID - this is the target container
+
+Request Model : 
+
+* path - container filesystem target for the copy operation
+
+Response Model :
+
+* stream - tar stream of the data to be moved to the container
```
hickeng: I'm unclear why the data is in the response for the copyTo case - is this a copy/paste error?

doc/design/docker-cp.md
```diff
@@ -0,0 +1,144 @@
...
+
+403 - in the event of a readonly filesystem
+
+404
+
+500
+
+```
+
+Because we need to prevent a poweroff(unless forced) of a container during a copy, we will pass the ID directly to the portlayer rather than fetching a handle.
```
hickeng: Handles are only supposed to be used for element composition - this isn't part of the composition API (Join, Bind, Unbind, Leave, Commit) so we wouldn't be using a Handle anyway.

doc/design/docker-cp.md
```diff
@@ -0,0 +1,144 @@
...
+
+OR 
+
+* Error - error response from the portlayer
+
+```
+
+### Personality design
+
+The personality has two endpoints that are called depending on the to/from situation above. Each of those endpoints will call their respective portlayer counterparts. There will be several checks during this process.
```
hickeng: I'd prefer not to require the caller to know what the current state of the container is - it's a fundamentally racy thing as the state can change between being checked and the copy call being made. Let's have this be a single endpoint and the switching logic done by the portlayer. If you're talking about direction then the phrasing is confusing.

From a practical perspective I suggest thinking about the port layer side as import/export or read/write rather than copyFrom/copyTo. This is because the same paths are needed for `load/import` and `save/export` and copy is actually a subcase of these operations.

doc/design/docker-cp.md
```diff
@@ -0,0 +1,144 @@
...
+The personality code should be fairly straight forward, the goal is to call the portlayer for the copy while passing the tar stream either to the portlayer or from it. The best way to do this is to hook the writiers together using a PipeReader and then shepherding the data in the direction it needs to go. 
+
+## Portlayer Design
+
+ There will be two situations to be concerned with in the portlayer. When the container is *on* and when the container is *off*.
+ 
+ 
+### Off Container Scenario
+ 
+ This is the case where guest tools is not needed since the guest in off. This means that all action will need to occur on the VCH. There are several concerns and many of them are listed in this ticket [here](https://github.com/vmware/vic/issues/4105)
```
hickeng: #4105 is inherently when we *are* using toolbox. Either this is a typo and a different ticket was intended or clarification is required.

doc/design/docker-cp.md
```diff
@@ -0,0 +1,144 @@
...
+ 
+ The most concern surrounding this piece of work will be surrounding VCH restart and confirming that the disk is still left in a good state if a copy was occuring mid restart. Additionally we need to block a container from starting when we have the r/w layer mounted. While that container is blocked from being started we might look at the possibility of returning a 500 error for a start invocation on a blocked container. We must also block other copy operations until the current one is successful, one major reason is because we cannot do any kind of sanity checking on available space with concurrent writes to the same disk.
```
hickeng: We can do just as much sanity checking with concurrent writes as we can with serialized writes. I see no reason to prevent concurrent copy to/from _apart from the fact that we cannot provide stable semantics as to what occurs_. The same issues apply to concurrent copy with normal `cp` command and a local filesystem.

We should be returning "409 Conflict" rather than 500. Longer term we should probably allow configuration of whether to block in these circumstances or return an error - blocking is preferable in many automation scenarios as it prevents the need for the caller to differentiate errors that will resolve themselves in time.

doc/design/docker-cp.md
```diff
@@ -0,0 +1,144 @@
...
+ 
+ Additionally, it must be ensured that regardless of error or success the disk must be detached from the VCH. We must ensure that this also occurs when we boot up the VCH in the event that the disk is attached to the VCH and is hard shutdown. If the VCH does not already have a disk detach functionality at boot time we should be adding this. This is an ideal time to invoke calls to `fsck` and `journal` in order to recover from mid copy shutdowns. Since the VCH is meant to be stateless(and the portlayer) we should aim to jettison and disk that is attached to the VCH at boot time if we do write this portion of the portlayer(We should even make it a separate ticket as testing and ensuring quality on this could take time.)
```
hickeng: There is already code (and I think a test) for ensuring that disks are detached on VCH boot if present.
Unsure that `journal` needs to be run explicitly for ext4 - the journal should be replayed on mount regardless (unless `noload`) is specified. Likewise fsck is of marginal value for a journaled filesystem.
There is this: https://github.com/vmware/vic/issues/4872#issuecomment-305608290 which suggests we may be exposed in some manner, but we'd need to determine what sort of writes could lead us to that circumstance (O_DIRECT maybe?)

doc/design/docker-cp.md
```diff
@@ -0,0 +1,144 @@
...
+ 
+ In An Ideal scenario without any failures the process will be as such : 
+ 
+ ```
+ 1) get request from personality for a file copy, it will have the stream to readfrom/write to as well as a path and a container ID.
+ 
+ 2) depending on the endpoint hit depends on the to/from scenario either way in an *off* scenario we would first get the container object and confirm that it is off. 
+ 
+ 3) Once confirmed that the container is off we will need to immediately block starts(and even wait if one is triggered between the status check and requesting the lockdown). In addition to this, we must block other `cp` operations until the current one is complete.
```
hickeng: We *must not* block starts after a call that checks state - we've no guarantee that the follow up copy will ever actually occur. If blocking of this type needs to occur it must occur within a single portlayer call.

doc/design/docker-cp.md
```diff
@@ -0,0 +1,144 @@
...
+ In An Ideal scenario without any failures the process will be as such : 
+ 
+ ```
+ 1) get request from personality for a file copy, it will have the stream to readfrom/write to as well as a path and a container ID.
+ 
+ 2) depending on the endpoint hit depends on the to/from scenario either way in an *off* scenario we would first get the container object and confirm that it is off. 
+ 
+ 3) Once confirmed that the container is off we will need to immediately block starts(and even wait if one is triggered between the status check and requesting the lockdown). In addition to this, we must block other `cp` operations until the current one is complete. 
+ 
+ 4) Then we will mount the r/w layer at a temp directory location. and assemble that path along with the target copy path.
```
hickeng: The r/w layer should be mounted non-persistent if we're doing a copyFrom to protect against accidental modification.

Should we also be mounting volume disks that are associated with the container? What's the docker behaviour when copying from volume paths?
We have issues here if the volume is a vmdk and is currently in use by a running container. Even if it's not in use, we'd need to block containers using it in the same fashion as for the r/w disk.

doc/design/docker-cp.md
```diff
@@ -0,0 +1,144 @@
...
+ 
+ 5) Begin streaming the tar file to the mounted disk. 
+ 
+ 6) We will need to untar this in place while checking for existing files and behaving according to the overwrite dir flag. We will need this to be in place since we could be operating with limited space, we want to do this in a fashion that we have as little storage overhead during unpacking as possible. Checking flags for `untar` may help significantly with this. Otherwise we may need to read the tar headers themself before attempting the write and determining if we have the size to do this.
```
hickeng: Do we need to check ahead of time? What is the docker behaviour here - my estimation is it's likely fail when the filesystem returns an out-of-space error.

doc/design/docker-cp.md
matthewavery: No reason not to, I have not worked on the swagger lately. Path would be even better, previously I had remembered seeing it in the query.

doc/design/docker-cp.md
matthewavery: Agreed, this was a simple proposal for the personality work @sflxn has taken up the work for ferrying the data to the portlayer.

doc/design/docker-cp.md
matthewavery: yes it is a typo.

doc/design/docker-cp.md
matthewavery: This was simply that I did not get a chance to reread the typed up document before posting the PR. I had plans to vet this for copy paste errors before pasting it. I wanted to get it up here so that we can align our efforts though. Sorry about this.

doc/design/docker-cp.md
matthewavery: It should be a single target in the portlayer. This is referring to the personality design in that a `docker cp` call will call two different endpoints depending on the *to* or *from* situation. After talking with @jzt I wholly agree, I was unaware of the conversation about aligning the archiving efforts for all of these commands and it is very ideal.

matthewavery: Also I this portion is talking about direction the difference between a `copy to` and a `copy from` operation. as far as the `on` and `off` behavior this is definitely planned to be contained in a single functional endpoint. Direction should be two different functional endpoints however. This is a logical separation of behavior, since the personality will have two different endpoints for this behavior and the behavior will either expect data or give data this lends to the `import/export` behavior. These functions do not have names yet though I was planning to name them something along the lines of `ExportArchiveToTarget`(did not use container here since the portlayer should not necessarily care as long as the ID is resolvable) and `ImportArchiveFromTarget`.

doc/design/docker-cp.md
matthewavery: I have updated the document to mention this in the design doc and eliminated the streams completely.

doc/design/docker-cp.md
matthewavery: Yes, I added this line in my first pass on the doc and did not remove it later on.

doc/design/docker-cp.md
matthewavery: The implication of that piece is that if there are concurrent writes we cannot ensure either succeed. If we only allow one write at a time then we may have one write succeed and one fail due to space constraints. In addition to that, we would also need to build in a mechanic for determining a disk attachment cohesion for multiple calls to `ReadImage` and `WriteImage`. This may already be the case as I have not read the implementation of `WriteImage` recently. If we are concerned with wanting concurrent writes then it is certainly within our ability easily. I think it is a matter of whether we would like to give each `import` a chance to succeed individually. As far as streaming all of the contents into the disk of the container, we could end up with two partially fulfilled copies, unless our implementation rollsback failed copies.

matthewavery: I agreed with the 409 I will add that as part of the response. The models are very fluid as they are simply the initial approach without input from you or @sflxn 😄

matthewavery: And blocking was always my intention. I will include that as well.

doc/design/docker-cp.md
matthewavery: I was 90% sure that this code already existed. I will remove the additional commenting surrounding that. I will also update the comments(likely remove) surrounding `journal` and `fsck`.

doc/design/docker-cp.md
matthewavery: right I was not advocating that we block on check state, this was poorly worded. The blocking operation should only be the portlayer call that has the r/w layer mounted. I will rectify this wording.

doc/design/docker-cp.md
matthewavery: This is also something that needs investigation. I have thought about this in the past. As far as copying an archive to a filesystem with volumes. For a first pass we may desire to not attempt the copies of files to volumes in an `offline` scenario. The disk behavior would require us to actually power down another container which is not acceptable. Any disk based volume approach would be subject to this. For the nfs volumes this is still an issue as well, since we could have another r/w actor writing to the same location at the same time. the behavior and error scenarios for this are uncertain. It could also very adversely affect a volume that is readonly for some containers.

matthewavery: some info on this

```
It is not possible to copy certain system files such as resources under /proc, /sys, /dev, tmpfs, and mounts created by the user in the container.
```
This does not cover volumes mounted through docker, I am reading through further implementation details now.

matthewavery: The above suggests that docker may remount vanilla volumes for the copy. I do not think this can be garunteed for their plugin based behavior though. I will take a look through their code.

doc/design/docker-cp.md
matthewavery: If we do concurrent writes then checking ahead of time may not be very useful. I do not see any mention of this in their documentation so I will take a look in their code. I also suspect it is fail hard on write error(OOS error). That would be more conducive to multiple concurrent writes.

doc/design/docker-cp.md
```diff
@@ -0,0 +1,146 @@
...
+
+409
+
+500
+
+```
+
+Docker does not do much in the way of string checking as far as the path is concerned. This is likely not a big concern for the personality as the actual copy commmand will fail if a bad path string is provided. 
+
+The personality code should be fairly straight forward, the goal is to call the portlayer for the copy while passing the tar stream either to the portlayer or from it. The best way to do this is to hook the writiers together using a PipeReader and then shepherding the data in the direction it needs to go.
```
sflxn: I think all you need to say here is that we can use the same streaming mechanism used for attach, exec, and logs.

matthewavery: 👍 will modify this part.

doc/design/docker-cp.md
```diff
@@ -0,0 +1,146 @@
...
+## Portlayer Design
+
+ There will be two situations to be concerned with in the portlayer. When the container is *on* and when the container is *off*.
+ 
+ 
+### Off Container Scenario
+ 
+ This is the case where guest tools is not needed since the guest in off. This means that all action will need to occur on the VCH. There are several concerns and many of them are listed in this ticket [here](https://github.com/vmware/vic/issues/5295)
+ 
+ The most concern surrounding this piece of work will be surrounding VCH restart and confirming that the disk is still left in a good state if a copy was occuring mid restart. Additionally we need to block a container from starting when we have the r/w layer mounted. While that container is blocked from being started we might look at the possibility of returning a 500 error for a start invocation on a blocked container. We must also block other copy operations until the current one is successful, one major reason is because we cannot do any kind of sanity checking on available space with concurrent writes to the same disk. Blocking operations until another is completed is ideal, since we work in a highly automated environment.
```
sflxn: The "important concern".  Most concern doesn't sound right.

doc/design/docker-cp.md
```diff
@@ -119,15 +119,17 @@ The personality code should be fairly straight forward, the goal is to call the
  
  2) depending on the endpoint hit depends on the to/from scenario either way in an *off* scenario we would first get the container object and confirm that it is off. 
  
- 3) Once confirmed that the container is off we will need to immediately block starts(and even wait if one is triggered between the status check and requesting the lockdown). In addition to this, we must block other `cp` operations until the current one is complete. 
+ 3) Once confirmed that the container is off. We will make the request to mount for the import/export operation.
```
hickeng: From a practical perspective it's likely that vSphere gives us everything we need to reconcile a race between copy and power-on operations. Only one of attach/powerOn will succeed as both will cause the container r/w layer to be locked.

In the case of powerOn winning, we should simply fall back to the online path for the copy. In the case of attach winning we would return the 409.

doc/design/docker-cp.md
```diff
@@ -119,15 +119,17 @@ The personality code should be fairly straight forward, the goal is to call the
...
  
- 5) Begin streaming the tar file to the mounted disk. 
  
- 6) We will need to untar this in place while checking for existing files and behaving according to the overwrite dir flag. We will need this to be in place since we could be operating with limited space, we want to do this in a fashion that we have as little storage overhead during unpacking as possible. Checking flags for `untar` may help significantly with this. Otherwise we may need to read the tar headers themself before attempting the write and determining if we have the size to do this. 
  
- 7) Upon successfully copying the data we detach the disk.
+ 7) Upon successfully copying the data we will adjust the permissions to be appropriately aligned with our decisions. Then detach the disk.
```
hickeng: We should be able to set the permissions during the extraction rather than doing it after the fact (which is racy in an error scenario). If need be we can update the archive/tar package to add this capability.

doc/design/docker-cp.md
```diff
@@ -0,0 +1,146 @@
...
+
+ There will be two situations to be concerned with in the portlayer. When the container is *on* and when the container is *off*.
+ 
+ 
+### Off Container Scenario
+ 
+ This is the case where guest tools is not needed since the guest in off. This means that all action will need to occur on the VCH. There are several concerns and many of them are listed in this ticket [here](https://github.com/vmware/vic/issues/5295)
+ 
+ The most concern surrounding this piece of work will be surrounding VCH restart and confirming that the disk is still left in a good state if a copy was occuring mid restart. Additionally we need to block a container from starting when we have the r/w layer mounted. While that container is blocked from being started we might look at the possibility of returning a 500 error for a start invocation on a blocked container. We must also block other copy operations until the current one is successful, one major reason is because we cannot do any kind of sanity checking on available space with concurrent writes to the same disk. Blocking operations until another is completed is ideal, since we work in a highly automated environment.
+
```
sflxn: I would remove "The merrits for calling `fsck` have not been determined yet.  Investigation will be needed to see what kind of writes and operations could make us susceptible to needing `fsck`" and simply say,

We may need to further investigate whether we will also need `fsck` to check filesystem consistency.

doc/design/docker-cp.md
matthewavery: thanks

doc/design/docker-cp.md
```diff
@@ -0,0 +1,146 @@
...
+ 
+ 2) depending on the endpoint hit depends on the to/from scenario either way in an *off* scenario we would first get the container object and confirm that it is off. 
+ 
+ 3) Once confirmed that the container is off. We will make the request to mount for the import/export operation. 
+ 
+ 4) Then we will mount the r/w layer at a temp directory location and block all start operations on this container as we have the r/w already mounted to different compute device. In the even of a `copyFrom` scenario the disk should be mounted as `non-persistent`
+ 
+ 5) Begin streaming the tar file to the mounted disk at the assembled path. 
+ 
+ 6) We will need to untar this in place while checking for existing files and behaving according to the overwrite dir flag. We will need this to be in place since we could be operating with limited space, we want to do this in a fashion that we have as little storage overhead during unpacking as possible. Checking flags for `untar` may help significantly with this. Otherwise we may need to read the tar headers themself before attempting the write and determining if we have the size to do this.
```
sflxn: We might not need to do this anymore.  Talk to @dougm for latest info.  I believe he's going to try to do some of this in the govmomi code.

doc/design/docker-cp.md
```diff
@@ -0,0 +1,146 @@
...
+ 
+ 4) Then we will mount the r/w layer at a temp directory location and block all start operations on this container as we have the r/w already mounted to different compute device. In the even of a `copyFrom` scenario the disk should be mounted as `non-persistent`
+ 
+ 5) Begin streaming the tar file to the mounted disk at the assembled path. 
+ 
+ 6) We will need to untar this in place while checking for existing files and behaving according to the overwrite dir flag. We will need this to be in place since we could be operating with limited space, we want to do this in a fashion that we have as little storage overhead during unpacking as possible. Checking flags for `untar` may help significantly with this. Otherwise we may need to read the tar headers themself before attempting the write and determining if we have the size to do this.  
+ 
+ 
+ 
+ 7) Upon successfully copying the data we will adjust the permissions to be appropriately aligned with our decisions. Then detach the disk.
```
sflxn: Same as above.  Talk to @dougm.

doc/design/docker-cp.md
hickeng: @sflxn Unsure, but I think you're talking about the online copy case whereas @matthewavery is talking about the offline copy case.

sflxn:
We should add some info about ContainerStatPath as docker cp won't work without this.  I'm looking at the cli code right now to determine their workflow.

matthewavery:
I have added ContainerStatPath to the doc. Pushing an update shortly. I will be talking to @dougm on Monday most likely.

doc/design/docker-cp.md
matthewavery: correct, this section is listed under the offline copy case. I have not listed anything on the `Online` case since this sprint only contains the offline ticket for now.

doc/design/docker-cp.md
matthewavery: I have looked at @dougm stuff, We likely do not need to worry about it. I would like to leave it there so that readers understand that we do have an obligation to change the permissions after the untar.

doc/design/docker-cp.md
```diff
@@ -0,0 +1,192 @@
...
+
+
+
+### Personality to Portlayer Communication
+
+Since all 4 of these must be supported we will require both distinctions for when the container is off and which endpoint to call to distinguish between *to* and *from* operations. 
+
+The proposed solution is to have two endpoints designed for the portlayer api swagger server. Both endpoints will exist along the same request path as two different verbs. The request path should be as such : 
+
+@sflxn has already implemented grpc streaming behavior for other components and we will aim for this implementation rather than passing streams through a restful approach between the personality and the portlayer.
```
sflxn: we have no grpc in our system yet.  It's just streaming.

Issue #4617, opened 2017-04-06
VCH reconfigure command


emlin:

emlin:
@mdubya66 @hickeng disagree with this solution and the manifest definition described in vic-machine design document is combined with authority delegation, which is not clear yet. I'd like to postpone this story atm.

emlin:
not clear for this feature yet, leave it in backlog atm

emlin:
@corrieb Saw your demo yesterday for the json manifest file support, as this feature will be the first step to support manifest file in vic-machine, could you please add the data format definition in this issue? To make sure we have a consistent result
thanks

corrieb:
@emlin The format is simply a proposal that made most sense to me. It's probably a good starting point for a YML-based format. I want my solution to be redundant the moment you have a good YML manifest solution - there's no sense in having both - so there's no need for "consistency" in that sense. My solution is what made most sense to me as a reasonable work-around. It's not a bad starting point for a data format.  

See https://github.com/corrieb/bensdoings/blob/master/vic-machine/example-complete.json

emlin:
@corrieb thanks for providing the reference.
@corrieb I understand we don't need to keep both of our code, cause the function is duplicated after vic-machine support create/configure from manifest file.
The consistency I talked about is the design idea. We all know that current vic-machine create options is a little bit overwhelming for customer, but to have a really good design to make it beautiful is also not easy, cause we have to have that many options.
Manifest file can solve some of the concerns, but obviously will have new issue introduced. For example, we'll need way to explain all keywords defined in manifest, to make sure if user want to add new attributes into their manifest file, they can get help on that.

So far, it's easy to see that manifest file looks much better than the command options, so it definitely will be a good start for our enhancement.
What I want you to leave here is to write the enhancement you think that should have in the manifest file format.

Also @karthik-narayan @mdubya66 @hickeng @ali5ter Please leave your expectation here, doesn't matter if we can do it or not immediately. It's appreciated if you can do it as early as possible, so we can address your requirements in this story.

corrieb:
To be clear, there is no requirement in 1.2 to support a manifest format. @karthik-narayan has confirmed. The functional requirement to be able to serialize the configuration of an existing VCH to be used as input to reconfigure makes sense. However, this can be achieved much more simply by outputting the configuration of a VCH as arguments that can be piped to a file or cut and pasted.

I recommend we change the name of this issue to better reflect the functional requirement.

[...]

PR #5418, opened 2017-06-08
[specific ci=Group6-VIC-Machine] implement vic-machine inspect config

willsherwood:
Implements #4617


tests/test-cases/Group6-VIC-Machine/6-09-Inspect.robot
```diff
@@ -38,6 +38,23 @@ Inspect VCH Configuration
...
+    Should Contain  ${output}  --volume-store=ds://%{TEST_DATASTORE}
+    Should Contain  ${output}  --bridge-network=%{BRIDGE_NETWORK}
+    Should Not Contain  ${output}  --cpu
+    Should Not Contain  ${output}  --cpu-shares
+    Should Not Contain  ${output}  --memory
+    Should Not Contain  ${output}  --memory-shares
+    Should Not Contain  ${output}  --base-image-size
+    Should Not Contain  ${output}  --bridge-network-range
+    Should Be Equal As Integers  0  ${rc}
+
```
emlin: Could check no log/info printed

tests/test-cases/Group6-VIC-Machine/6-09-Inspect.robot
```diff
@@ -62,6 +79,26 @@ Inspect VCH Configuration with Resource Limitation
...
+    Should Contain  ${output}  --memory-shares=6000
+    Should Contain  ${output}  --memory-reservation=512
+    Should Contain  ${output}  --memory=8000
+    Should Contain  ${output}  --cpu=10000
+    Should Contain  ${output}  --cpu-reservation=512
+    Should Contain  ${output}  --cpu-shares=high
+    Should Contain  ${output}  --endpoint-memory=4096
+    Should Contain  ${output}  --endpoint-cpu=2
+    Should Be Equal As Integers  0  ${rc}
+
```
emlin: one more empty line

emlin: ditto^^^

emlin:
@ali5ter Need your comments on the command options.
Here we already have command line like `vic-machine inspect --configuration(--conf)` to show VCH configuration, and we need a new option to output command line only, without any other debug/log messages, so user is able to script their command easily.
Now the impl is like `vic-machine inspect --rawconfig`. How do you think about that?

tests/test-cases/Group6-VIC-Machine/6-09-Inspect.robot
```diff
@@ -38,6 +38,25 @@ Inspect VCH Configuration
...
+    Should Contain  ${output}  --compute-resource=%{TEST_RESOURCE}
+    Should Contain  ${output}  --volume-store=ds://%{TEST_DATASTORE}
+    Should Contain  ${output}  --bridge-network=%{BRIDGE_NETWORK}
+    Should Not Contain  ${output}  --cpu
+    Should Not Contain  ${output}  --cpu-shares
+    Should Not Contain  ${output}  --memory
+    Should Not Contain  ${output}  --memory-shares
+    Should Not Contain  ${output}  --base-image-size
+    Should Not Contain  ${output}  --bridge-network-range
+	Should Not Contain  ${output}  INFO
```
emlin: in the test code, we're using 4 spaces instead of tab. 😞

tests/test-cases/Group6-VIC-Machine/6-09-Inspect.robot
```diff
@@ -62,6 +81,27 @@ Inspect VCH Configuration with Resource Limitation
...
+    Should Contain  ${output}  --bridge-network=%{BRIDGE_NETWORK}
+    Should Contain  ${output}  --memory-shares=6000
+    Should Contain  ${output}  --memory-reservation=512
+    Should Contain  ${output}  --memory=8000
+    Should Contain  ${output}  --cpu=10000
+    Should Contain  ${output}  --cpu-reservation=512
+    Should Contain  ${output}  --cpu-shares=high
+    Should Contain  ${output}  --endpoint-memory=4096
+    Should Contain  ${output}  --endpoint-cpu=2
+	Should Not Contain  ${output}  INFO
```
emlin: ditto

ali5ter:
@emlin Before making a recommendation, I wanted to understand the goal...

Need a way from the CLI to display the VCH configuration but output it in different ways. 

Your description seems to imply that --configuration outputs some human friendly format that includes debug and log information and you want --rawconfig to display something parsable for pipe workflows on the command line. Is this correct?

Assuming it is...

#### Questions:
1. Will there be any point in the future when `vic-machine` will be used to edit the VCH configuration?
2. Will the Cloud Admin typically use `vic-machine` to dump the config into a parsable format and not for human readable output?

### Approach 1
* The common command here is 'configuration' - it doesn't feel like an option
* A typical shortening of configuration is config, e.g. git config
* The way the output of a CLI command is formatted is an option
* 'configuration' may need to be extensible

So, with this in mind: 
`vic-machine inspect config` to echo a human readable output
`vic-machine inspect config --output raw` to modify the output format to whatever your raw format is... depending on what it is, the option could be more specific, e.g. `--output json`

### Approach 2
IF one was to use the `--` prefix then I'd suggest:
`vic-machine inspect --configuration`
`vic-machine inspect --config`
`vic-machine inspect --configuration-raw`
`vic-machine inspect --config-raw`

However, this still feels like a command and I prefer approach 1. However, I may not fully understand what the goal is.

emlin:
@ali5ter good comment 👍 , thank you.

For your question, yeah, generally vic-machine inspect will dump some log message to help user understand we're still executing, and then after that, a configuration output will be shown. Here is one example output:
```
Jun 12 2017 18:41:14.574Z INFO  ### Inspecting VCH ####
Jun 12 2017 18:41:14.714Z INFO  Validating target
Jun 12 2017 18:41:14.990Z INFO  
Jun 12 2017 18:41:14.991Z INFO  Target VCH created with the following options: 

	--target=https://10.162.47.34
	--thumbprint=2C:8A:C6:95:DF:C6:ED:E2:91:31:86:7C:B7:0E:28:D8:F6:13:82:E7
	--name=test1
	--compute-resource=/vcqaDC/host/cls/Resources
	--ops-user=administrator@vsphere.local
	--image-store=ds://vsanDatastore
	--bridge-network=bridge
	--public-network=external
	--http-proxy=http://proxy.vmware.com:2318
```

So it's not user friendly format or not, just generally vic-machine command is printing out whatever might help the user experience.


>    Will there be any point in the future when vic-machine will be used to edit the VCH configuration?

Yes, I think so. That should be the purpose of we add the manifest file. Even this option format might not be the final manifest file format, it will be a vic-machine function to help user managing the manifest file

>    Will the Cloud Admin typically use vic-machine to dump the config into a parsable format and not for human readable output?

I think it's useful, cause the reason of this output is to help user generate configure command to reconfigure existing VCH, or use this template to create a new VCH.

In the end, I also agree with the first option. If this answer doesn't change your mind, let's do it.

emlin:
@ali5ter one more question, will the --output option slightly confusing if we end up with another option to actually write into a file in vic-machine?
I do not have that requirement, but if vic-machine is used to create manifest file, edit manifest file and also use manifest file to create VCH, write into file might be natual

ali5ter:
@emlin You are correct. `--output` is commonly used for identifying the file path where output is written. Good catch. `--format` may be a better option name.

emlin:
@willsherwood please use `--format raw`, thanks

cmd/vic-machine/inspect/inspect.go
```diff
@@ -89,6 +97,17 @@ func (i *Inspect) Flags() []cli.Flag {
 	return flags
 }
 
+func (i *Inspect) ConfigFlags() []cli.Flag {
+	output := cli.StringFlag{
+		Name:        "output",
```
emlin: As discussion with @ali5ter , output might be confusing if we need to write manifest file. Please change output to format

cmd/vic-machine/inspect/inspect.go
```diff
@@ -89,6 +97,17 @@ func (i *Inspect) Flags() []cli.Flag {
...
+		Value:       "",
+		Usage:       "Determine the format of configuration output",
```
emlin: feels like user will need to know what is accepted value for this option. So far, it's only "raw".

cmd/vic-machine/inspect/inspect.go
willsherwood: Right, thanks.

cmd/vic-machine/inspect/inspect.go
```diff
@@ -99,21 +118,24 @@ func (i *Inspect) processParams() error {
...
 	defer func() {
 		err = common.LogErrorIfAny(clic, err)
 	}()
 
-	if err = i.processParams(); err != nil {
-		return err
-	}
-
-	if i.Debug.Debug != nil && *i.Debug.Debug > 0 {
+	if i.Output != "" {
```
emlin: here, might compare the value with "raw", if not "raw", take it as "", with warning message

cmd/vic-machine/inspect/inspect.go
```diff
@@ -156,55 +178,76 @@ func (i *Inspect) Run(clic *cli.Context) (err error) {
...
-	log.Infof("Target VCH created with the following options: \n\n\t%s\n", strOptions)
+	if i.Output == "raw" {
+		strOptions := strings.Join(options, " ")
+		fmt.Println(strOptions)
+	} else if i.Output == "" {
+		strOptions := strings.Join(options, "\n\t")
+		log.Info("")
+		log.Infof("Target VCH created with the following options: \n\n\t%s\n", strOptions)
+	} else {
+		return fmt.Errorf("Invalid configuration output format %s", i.Output)
```
emlin: this should be a warning shown earlier

emlin:
As the command line is change, be sure to run all related cases (the safe way is all vic-machine test cases) before merge

emlin:
sorry for closed by mistake 😄

PR #5373, opened 2017-06-05
OVA web service to create demo VCH

morris-jason:
Fixes #4731, #4732

PR #5424, opened 2017-06-08
[specific ci=6-16-Config] Add support for vic-machine configure dns 

### LINKED ISSUE #4517: vic-machine reconfigure: update DNS servers
**User Statement:**

As a VIC operation user, I need the ability to update the dns servers. 

**Example Use Case**
The default dns server is `8.8.8.8`. Now I want to switch to my own dns server.

**Details:**
- supported operations:
  - restore default dns servers (--reset-dns-server)
  - replace existing DNS servers (this corresponds to remove existing servers first and then add new servers. For example, if only one dns server is specified, all the existing dns servers will be replaced by this single one dns server ) (--dns-server)
- hot update:
- input/output data format:
  - input: e.g., `--clear-dns-server ` or `--dns-server 172.16.10.10 --dns-server 172.16.11.11`
  - output: N/A
- impact on VCH/containerVMs:
  - VCH: The VCH will reload the dns configuration upon VCH restart. 
  - containerVMs: no
- is upgrade needed: 
   -  no

**Acceptance Criteria:**
-  add the ability to update the DNS servers 
- add relevant test cases

jialin-li:
Fixes #4517

### DIFF
--- a/cmd/vic-machine/common/dns.go
+++ b/cmd/vic-machine/common/dns.go
@@ -1,0 +1,66 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"net"
+
+	log "github.com/Sirupsen/logrus"
+	"gopkg.in/urfave/cli.v1"
+
+	"github.com/vmware/vic/pkg/errors"
+)
+
+// general dns
+type DNS struct {
+	dns   cli.StringSlice `arg:"dns-server"`
+	IsSet bool
+}
+
+func (d *DNS) DNSFlags(hidden bool) []cli.Flag {
+	return []cli.Flag{
+		cli.StringSliceFlag{
+			Name:   "dns-server",
+			Value:  &d.dns,
+			Usage:  "DNS server for the client, public, and management networks. Defaults to 8.8.8.8 and 8.8.4.4 when VCH uses static IP",
+			Hidden: hidden,
+		},
+	}
+}
+
+// processDNSServers parses DNS servers used for client, public, mgmt networks
+func (d *DNS) ProcessDNSServers() ([]net.IP, error) {
+	var parsedDNS []net.IP
+	if len(d.dns) > 0 {
+		d.IsSet = true
+	}
+
+	for _, n := range d.dns {
+		if n != "" {
+			s := net.ParseIP(n)
+			if s == nil {
+				return nil, errors.Errorf("Invalid DNS server specified: %s", n)
+			}
+			parsedDNS = append(parsedDNS, s)
+		}
+	}
+
+	if len(parsedDNS) > 3 {
+		log.Warn("Maximum of 3 DNS servers allowed. Additional servers specified will be ignored.")
+	}
+
+	log.Debugf("VCH DNS servers: %s", parsedDNS)
+	return parsedDNS, nil
+}
--- a/cmd/vic-machine/configure/configure.go
+++ b/cmd/vic-machine/configure/configure.go
@@ -41,6 +41,7 @@ type Configure struct {
 
 	proxies   common.Proxies
 	cNetworks common.CNetworks
+	dns       common.DNS
 
 	upgrade  bool
 	executor *management.Dispatcher
@@ -87,7 +88,7 @@ func (c *Configure) Flags() []cli.Flag {
 	}
 
 	proxies := c.proxies.ProxyFlags(false)
-
+	dns := c.dns.DNSFlags(true)
 	target := c.TargetFlags()
 	ops := c.OpsCredentials.Flags(false)
 	id := c.IDFlags()
@@ -98,7 +99,7 @@ func (c *Configure) Flags() []cli.Flag {
 
 	// flag arrays are declared, now combined
 	var flags []cli.Flag
-	for _, f := range [][]cli.Flag{target, ops, id, compute, iso, cNetwork, proxies, util, debug} {
+	for _, f := range [][]cli.Flag{target, ops, id, compute, iso, dns, cNetwork, proxies, util, debug} {
 		flags = append(flags, f...)
 	}
 
@@ -109,6 +110,11 @@ func (c *Configure) processParams() error {
 	defer trace.End(trace.Begin(""))
 
 	if err := c.HasCredentials(); err != nil {
+		return err
+	}
+
+	var err error
+	if c.DNS, err = c.dns.ProcessDNSServers(); err != nil {
 		return err
 	}
 
@@ -171,6 +177,12 @@ func (c *Configure) copyChangedConf(o *config.VirtualContainerHostConfigSpec, n 
 
 	// Copy the thumbprint directly since it has already been validated.
 	o.TargetThumbprint = n.TargetThumbprint
+
+	if c.dns.IsSet {
+		for k, v := range o.ExecutorConfig.Networks {
+			v.Network.Nameservers = n.ExecutorConfig.Networks[k].Network.Nameservers
+		}
+	}
 }
 
 func updateSessionEnv(sess *executor.SessionConfig, envName, envValue string) {
@@ -302,7 +314,6 @@ func (c *Configure) Run(clic *cli.Context) (err error) {
 	// evaluate merged configuration
 	newConfig, err := validator.Validate(ctx, c.Data)
 	if err != nil {
-		log.Error("Configuring cannot continue: configuration validation failed")
 		return err
 	}
 
--- a/cmd/vic-machine/configure/configure.go
+++ b/cmd/vic-machine/configure/configure.go
@@ -88,6 +88,7 @@ func (c *Configure) Flags() []cli.Flag {
 	}
 
 	proxies := c.proxies.ProxyFlags(false)
+	// TODO (Jialin): after issue #5472 is fixed, change hidden back to false
 	dns := c.dns.DNSFlags(true)
 	target := c.TargetFlags()
 	ops := c.OpsCredentials.Flags(false)
@@ -314,6 +315,7 @@ func (c *Configure) Run(clic *cli.Context) (err error) {
 	// evaluate merged configuration
 	newConfig, err := validator.Validate(ctx, c.Data)
 	if err != nil {
+		log.Error("Configuring cannot continue: configuration validation failed")
 		return err
 	}
 
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -100,7 +100,7 @@ type Create struct {
 	volumeStores             cli.StringSlice `arg:"volume-store"`
 	insecureRegistries       cli.StringSlice `arg:"insecure-registry"`
 	whitelistRegistries      cli.StringSlice `arg:"whitelist-registry"`
-	dns                      cli.StringSlice `arg:"dns-server"`
+	dns                      common.DNS
 	clientNetworkName        string
 	clientNetworkGateway     string
 	clientNetworkIP          string
@@ -265,14 +265,6 @@ func (c *Create) Flags() []cli.Flag {
 			Usage:       "IP address with a network mask for the VCH on the management network, e.g. 10.0.2.2/24",
 			Destination: &c.managementNetworkIP,
 			Hidden:      true,
-		},
-
-		// general DNS
-		cli.StringSliceFlag{
-			Name:   "dns-server",
-			Value:  &c.dns,
-			Usage:  "DNS server for the client, public, and management networks. Defaults to 8.8.8.8 and 8.8.4.4 when VCH uses static IP",
-			Hidden: true,
 		},
 	}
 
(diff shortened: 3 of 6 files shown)


cmd/vic-machine/common/dns.go
```diff
@@ -0,0 +1,56 @@
+package common
```
anchal-agrawal: This file is missing the copyright header - see `cmd/vic-machine/configure/configure.go` for example.

cmd/vic-machine/common/dns.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"gopkg.in/urfave/cli.v1"
```
anchal-agrawal: go imports not formatted correctly - see [this thread](https://github.com/vmware/vic/pull/5373#discussion_r120218740).

cmd/vic-machine/common/dns.go
sgairo: @anchal-agrawal we should see if we should adjust the Makefile to pass check-headers.sh the fix argument so that the make all would add copyrights for us, though I wonder how it would work for new files, because I think it puts the range by default.

hmahmood: I have changed it so that it puts in 2017. However, we should not modify the make target since its runs on CI, and we don't want to modify files during builds.

cmd/vic-machine/common/dns.go
```diff
@@ -0,0 +1,71 @@
...
+			if s == nil {
+				return nil, errors.New("Invalid DNS server specified")
+			}
+			parsedDNS = append(parsedDNS, s)
+		}
+	}
+
+	// user input empty string, reset dns-server to default
+	if d.IsSet && parsedDNS == nil {
+		return []net.IP{net.ParseIP("8.8.8.8"), net.ParseIP("8.8.4.4")}, nil
```
hmahmood: This should only be the case if DHCP is not being used. See https://github.com/vmware/vic/blob/master/lib/install/validate/network.go#L237 for what happens during create.

jialin-li: Just fixed it!

anchal-agrawal:
@jialin-li the commit history seems wonky - possibly due to a forced git push? You should use `--force-with-lease` instead.

jialin-li:
Sorry I was doing some rebase stuff to keep it up to date with vmware/master.  Fixing it now!

cmd/vic-machine/configure/configure.go
```diff
@@ -111,6 +113,12 @@ func (c *Configure) processParams() error {
 		return err
 	}
 
+	var err error
+	usingDHCP := ip.IsUnspecifiedIP(c.Data.ClientNetwork.IP.IP) || ip.IsUnspecifiedIP(c.Data.PublicNetwork.IP.IP) || ip.IsUnspecifiedIP(c.Data.ManagementNetwork.IP.IP)
```
anchal-agrawal: This line could be extracted into a function since it's being used here and in `create.go`.

jialin-li: I wanted to have this line in common/dns.go, but it uses Data struct in lib/install/data/data.go which already includes the common package so including the Data struct in dns.go would have caused a circular import. Not too sure which function to move this line to. The possible place I see is into data.go itself, but not sure if that's appropriate.

anchal-agrawal: @hmahmood, what do you think?

cmd/vic-machine/create/create.go
```diff
@@ -482,10 +475,11 @@ func (c *Create) Flags() []cli.Flag {
 	debug := c.DebugFlags()
 	proxies := c.proxies.ProxyFlags(true)
 	cNetwork := c.containerNetworks.CNetworkFlags(true)
+	dns := c.dns.DNSFlags(true)
 
 	// flag arrays are declared, now combined
 	var flags []cli.Flag
-	for _, f := range [][]cli.Flag{target, compute, create, cNetwork, memory, cpu, tls, registries, proxies, syslog, iso, util, debug, help} {
+	for _, f := range [][]cli.Flag{target, compute, create, cNetwork, memory, cpu, tls, registries, proxies, dns, syslog, iso, util, debug, help} {
```
anchal-agrawal: Please place `dns` between `create` and `cNetwork` - the flags are printed in the slice's order and we should keep the same order.

cmd/vic-machine/common/dns.go
```diff
@@ -0,0 +1,71 @@
...
+	var parsedDNS []net.IP
+	if len(d.dns) > 0 {
+		d.IsSet = true
+	}
+
+	for _, n := range d.dns {
+		if n != "" {
+			s := net.ParseIP(n)
+			if s == nil {
+				return nil, errors.New("Invalid DNS server specified")
```
emlin: better to print out the wrong value here

cmd/vic-machine/configure/configure.go
emlin: this method is to test value set in data structure, so it should be fine to have the method in data.go

cmd/vic-machine/configure/configure.go
```diff
@@ -111,6 +113,12 @@ func (c *Configure) processParams() error {
...
+	if c.DNS, err = c.dns.ProcessDNSServers(usingDHCP); err != nil {
```
emlin: In reconfigure, this logic is not incorrect. User might not specify ip during reconfiguration, but that just means they're not trying to change ip settings this time.
If the dns value is set to empty, the default value should be reset after we query back old configuration, and based on that, to see if dhcp is used or not.

cmd/vic-machine/configure/configure.go
```diff
@@ -97,7 +99,7 @@ func (c *Configure) Flags() []cli.Flag {
 
 	// flag arrays are declared, now combined
 	var flags []cli.Flag
-	for _, f := range [][]cli.Flag{target, id, compute, iso, cNetwork, proxies, util, debug} {
+	for _, f := range [][]cli.Flag{target, id, compute, iso, cNetwork, proxies, dns, util, debug} {
```
emlin: as @anchal-agrawal mentioned for create command, please also use same option order with create here

cmd/vic-machine/configure/configure.go
emlin: well, this piece of code is actually part of validator logic, why do you copy that to here?

tests/test-cases/Group6-VIC-Machine/6-16-Config.md
```diff
@@ -33,6 +33,8 @@ This test requires that a vSphere server is running and available
 20. Configure the VCH by attempting to change an existing container network
 21. Configure VCH http proxy
 22. Verify http proxy is set correctly through govc
+23. Configure VCH dns server to default 8.8.8.8 and 8.8.4.4
+24. Verify dns server is set correctly through govc
```
emlin: need to verify the configuration through vic-machine inspect

tests/test-cases/Group6-VIC-Machine/6-16-Config.md
```diff
@@ -33,6 +33,8 @@ This test requires that a vSphere server is running and available
 20. Configure the VCH by attempting to change an existing container network
 21. Configure VCH http proxy
 22. Verify http proxy is set correctly through govc
+23. Configure VCH dns server to default 8.8.8.8 and 8.8.4.4
```
emlin: please also configure using empty dns input, to verify if the reset function works.
btw, in that case, you might need to use another dns for reconfiguration, to make sure the reset function actually works.

tests/test-cases/Group6-VIC-Machine/6-16-Config.md
jialin-li: Inspect doesn't show any information on dns, should I add an option for inspect then?

emlin: inspect should work for dns, but if it doesn't, think that's a bug should be fixed in this issue. :)
let me know how you need any help.

jialin-li: You are right! I wasn't looking at the right place, found dns information in inspect.

jialin-li:
@emlin If user starts with default DHCP, then configure with DNS servers, and later use the reset option "", should it reset to default DNS server values or should it clear the dns fields since DHCP is the default?

emlin:
> If user starts with default DHCP, then configure with DNS servers, and later use the reset option "", should it reset to default DNS server values or should it clear the dns fields since DHCP is the default?

The option "" means reset to default value. But default value itself is different based on combined old configuration and user's change.

So you should query back old configuration, merge that with new configuration, and then run the logic.

If the merged configuration is using dhcp, and the option is "", then reset to use dhcp, and if merged configuration has ip address configured, then reset to default dns 8.8.8.8 and 8.8.4.4.

the code to detect default value is copied from validator, but I think you don't need to do that. If you have the configuration merged correctly, validator will set dns for you.

cmd/vic-machine/common/dns.go
```diff
@@ -0,0 +1,71 @@
...
+			s := net.ParseIP(n)
+			if s == nil {
+				return nil, errors.Errorf("Invalid DNS server specified: %s", n)
+			}
+			parsedDNS = append(parsedDNS, s)
+		}
+	}
+
+	// user input empty string, modify the DNS to represent reset
+	if d.IsSet && parsedDNS == nil {
```
emlin: not necessary

lib/install/data/data.go
```diff
@@ -256,5 +256,12 @@ func (d *Data) CopyNonEmpty(src *Data) error {
 
 	d.Timeout = src.Timeout
 
+	if src.DNS != nil {
```
emlin: do you need this logic? you should just check if src.DNS.IsSet, and then copy the value to d is enough.

cmd/vic-machine/common/dns.go
jialin-li: This is used to return different results for no input and empty strings since empty string means reset. If user doesn't use the dns flag DNS is set to null, otherwise it's set to []net.IP{null, null}. This is used to distinguish reset when copying non empty data.

lib/install/data/data.go
jialin-li: The DNS in data is []net.IP so it doesn't have an IsSet field. I am trying to not change existing data structure within the data struct. So I used this logic to enable reset for dns.

cmd/vic-machine/common/dns.go
emlin: https://github.com/vmware/vic/blob/master/lib/install/validate/network.go#L236
this code will handle empty DNS logic, why should you differentiate from nil and empty net.IP?

cmd/vic-machine/configure/configure.go
```diff
@@ -87,7 +88,7 @@ func (c *Configure) Flags() []cli.Flag {
 	}
 
 	proxies := c.proxies.ProxyFlags(false)
-
+	dns := c.dns.DNSFlags(true)
```
emlin: might add a comment here to mention the new issue, and after that is fixed, it should be reverted back to false.

tests/test-cases/Group6-VIC-Machine/6-16-Config.robot
```diff
@@ -148,3 +148,12 @@ Configure VCH https-proxy through vch id
     ${rc}  ${output}=  Run And Return Rc And Output  govc vm.info -e %{VCH-NAME} | grep HTTPS_PROXY
     Should Be Equal As Integers  ${rc}  1
     Should Not Contain  ${output}  proxy.vmware.com:3128
+
+Configure VCH DNS server
+    ${output}=  Run  bin/vic-machine-linux inspect --name=%{VCH-NAME} --target=%{TEST_URL} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --password=%{TEST_PASSWORD} --timeout %{TEST_TIMEOUT} --conf
```
anchal-agrawal: If https://github.com/vmware/vic/pull/5418 gets merged before this, please change `--conf` to `config`.

tests/test-cases/Group6-VIC-Machine/6-16-Config.robot
```diff
@@ -148,3 +148,12 @@ Configure VCH https-proxy through vch id
...
+    Should Not Contain  ${output}  --dns-server  
+    ${output}=  Run  bin/vic-machine-linux configure --name=%{VCH-NAME} --target=%{TEST_URL} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --password=%{TEST_PASSWORD} --timeout %{TEST_TIMEOUT} --dns-server 10.118.81.1 --dns-server 10.118.81.2
+    Should Contain  ${output}  Completed successfully
+    ${output}=  Run  bin/vic-machine-linux inspect --name=%{VCH-NAME} --target=%{TEST_URL} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --password=%{TEST_PASSWORD} --timeout %{TEST_TIMEOUT} --conf
```
anchal-agrawal: Ditto. Please co-ordinate with @willsherwood while merging because the PR that goes in later will need to fix this test 😃

tests/test-cases/Group6-VIC-Machine/6-16-Config.robot
anchal-agrawal: https://github.com/vmware/vic/pull/5418 is merged.

cmd/vic-machine/configure/configure.go
```diff
@@ -87,7 +88,8 @@ func (c *Configure) Flags() []cli.Flag {
 	}
 
 	proxies := c.proxies.ProxyFlags(false)
-
+	// after issue #5472 is fixed, change hidden back to false
```
anchal-agrawal: It'd be nice to add a `TODO(<your name here>)` before the comment. Often, `git blame` is tough to navigate when code gets moved around.

lib/install/data/data.go
```diff
@@ -263,5 +263,7 @@ func (d *Data) CopyNonEmpty(src *Data) error {
 
 	d.Timeout = src.Timeout
 
+	d.DNS = src.DNS
```
anchal-agrawal: Question - should this be checking that `src.DNS.IsSet == true` before copying to `d.DNS`?

jialin-li: IsSet is checked in copychangedconfig, so it's safe to copy over. DNS here is []net.IP so I can't check IsSet here

PR #5427, opened 2017-06-08
Embed build number during ISO generation

hickeng:
Allows use of the staging caches in the build while still being able to update the build number.

Added during upgrade confirmation of #4872


caglar10ur:
LGTM

PR #5428, opened 2017-06-09
Introduce VirtualDiskConfig to DiskManager

caglar10ur:
This is a non-functional change to make DiskManager methods consistent.
 NewVirtualDisk now checks the diskmanager's cache before creating a new
 struct. Before that each call was creating a new instance and making
 locking and ref counting problematic.

 I still want to split CreateAndAttach into multiple pieces to make each
 function do a one thing instead of current behaviour. That will be a follow-up
 PR after this.

### DIFF
--- a/lib/portlayer/storage/vsphere/image.go
+++ b/lib/portlayer/storage/vsphere/image.go
@@ -62,11 +62,11 @@ var (
 )
 
 const (
-	StorageImageDir  = "images"
-	defaultDiskLabel = "containerfs"
-	defaultDiskSize  = 8 * 1024 * 1024
-	metaDataDir      = "imageMetadata"
-	manifest         = "manifest"
+	StorageImageDir     = "images"
+	defaultDiskLabel    = "containerfs"
+	defaultDiskSizeInKB = 8 * 1024 * 1024
+	metaDataDir         = "imageMetadata"
+	manifest            = "manifest"
 )
 
 type ImageStore struct {
@@ -409,7 +409,7 @@ func (v *ImageStore) scratch(op trace.Operation, storeName string) error {
 	imageDiskDsURI := v.imageDiskDSPath(storeName, portlayer.Scratch.ID)
 	op.Infof("Creating image %s (%s)", portlayer.Scratch.ID, imageDiskDsURI)
 
-	size = defaultDiskSize
+	size = defaultDiskSizeInKB
 	if portlayer.Config.ScratchSize != 0 {
 		size = portlayer.Config.ScratchSize
 	}
--- a/lib/portlayer/storage/vsphere/image.go
+++ b/lib/portlayer/storage/vsphere/image.go
@@ -64,7 +64,7 @@ var (
 const (
 	StorageImageDir  = "images"
 	defaultDiskLabel = "containerfs"
-	defaultDiskSize  = 8388608
+	defaultDiskSize  = 8 * 1024 * 1024
 	metaDataDir      = "imageMetadata"
 	manifest         = "manifest"
 )
@@ -79,9 +79,15 @@ type ImageStore struct {
 }
 
 func NewImageStore(op trace.Operation, s *session.Session, u *url.URL) (*ImageStore, error) {
-	dm, err := disk.NewDiskManager(op, s, DetachAll)
+	dm, err := disk.NewDiskManager(op, s)
 	if err != nil {
 		return nil, err
+	}
+
+	if DetachAll {
+		if err = dm.DetachAll(op); err != nil {
+			return nil, err
+		}
 	}
 
 	datastores, err := s.Finder.DatastoreList(op, u.Host)
@@ -277,7 +283,7 @@ func (v *ImageStore) cleanupDisk(op trace.Operation, ID, storeName string, vmdis
 
 		if vmdisk.Attached() {
 			op.Debugf("Detaching abandoned disk")
-			v.dm.Detach(op, vmdisk)
+			v.dm.Detach(op, vmdisk.VirtualDiskConfig)
 		}
 	}
 
@@ -322,8 +328,9 @@ func (v *ImageStore) writeImage(op trace.Operation, storeName, parentID, ID stri
 		v.cleanupDisk(op, ID, storeName, vmdisk)
 	}()
 
+	config := disk.NewPersistentDisk(diskDsURI).WithParent(parentDiskDsURI)
 	// Create the disk
-	vmdisk, err = v.dm.CreateAndAttach(op, diskDsURI, parentDiskDsURI, 0, os.O_RDWR, disk.Ext4)
+	vmdisk, err = v.dm.CreateAndAttach(op, config)
 	if err != nil {
 		return nil, err
 	}
@@ -359,7 +366,7 @@ func (v *ImageStore) writeImage(op trace.Operation, storeName, parentID, ID stri
 		return nil, err
 	}
 
-	if err = v.dm.Detach(op, vmdisk); err != nil {
+	if err = v.dm.Detach(op, vmdisk.VirtualDiskConfig); err != nil {
 		return nil, err
 	}
 
@@ -414,12 +421,14 @@ func (v *ImageStore) scratch(op trace.Operation, storeName string) error {
 		v.cleanupDisk(op, portlayer.Scratch.ID, storeName, vmdisk)
 	}()
 
+	config := disk.NewPersistentDisk(imageDiskDsURI).WithCapacity(size)
 	// Create the disk
-	vmdisk, err = v.dm.CreateAndAttach(op, imageDiskDsURI, nil, size, os.O_RDWR, disk.Ext4)
+	vmdisk, err = v.dm.CreateAndAttach(op, config)
 	if err != nil {
 		op.Errorf("CreateAndAttach(%s) error: %s", imageDiskDsURI, err)
 		return err
 	}
+
 	op.Debugf("Scratch disk created with size %d", portlayer.Config.ScratchSize)
 
 	// Make the filesystem and set its label to defaultDiskLabel
@@ -431,7 +440,7 @@ func (v *ImageStore) scratch(op trace.Operation, storeName string) error {
 		return err
 	}
 
-	if err = v.dm.Detach(op, vmdisk); err != nil {
+	if err = v.dm.Detach(op, vmdisk.VirtualDiskConfig); err != nil {
 		return err
 	}
 
@@ -469,7 +478,8 @@ func (v *ImageStore) GetImage(op trace.Operation, store *url.URL, ID string) (*p
 
 	var s = *store
 
-	dsk, err := v.dm.Get(op, diskDsURI, disk.Ext4)
+	config := disk.NewPersistentDisk(diskDsURI)
+	dsk, err := v.dm.Get(op, config)
 	if err != nil {
 		return nil, err
 	}
--- a/lib/portlayer/storage/vsphere/volume.go
+++ b/lib/portlayer/storage/vsphere/volume.go
@@ -51,9 +51,15 @@ func NewVolumeStore(op trace.Operation, storeName string, s *session.Session, ds
 		return nil, err
 	}
 
-	dm, err := disk.NewDiskManager(op, s, DetachAll)
+	dm, err := disk.NewDiskManager(op, s)
 	if err != nil {
 		return nil, err
+	}
+
+	if DetachAll {
+		if err = dm.DetachAll(op); err != nil {
+			return nil, err
+		}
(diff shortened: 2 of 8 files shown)


lib/portlayer/storage/vsphere/volume.go
```diff
@@ -51,11 +51,17 @@ func NewVolumeStore(op trace.Operation, storeName string, s *session.Session, ds
 		return nil, err
 	}
 
-	dm, err := disk.NewDiskManager(op, s, DetachAll)
+	dm, err := disk.NewDiskManager(op, s)
 	if err != nil {
 		return nil, err
 	}
 
+	if DetachAll {
```
hickeng: I think DetachAll should be during pl.storage initialization rather than in the NewX methods - this will cause havoc if we end up trying to create scratch at the same time as a NewVolumeStore for example - ditto if we ever have dynamic config.

lib/portlayer/storage/vsphere/volume.go
```diff
@@ -100,12 +106,13 @@ func (v *VolumeStore) VolumeCreate(op trace.Operation, ID string, store *url.URL
 	// Get the path to the disk in datastore uri format
 	volDiskDsURL := v.volDiskDsURL(ID)
 
+	config := disk.NewPersistentDisk(volDiskDsURL).WithCapacity(int64(capacityKB))
```
hickeng: expecting to see a `.WithCapacity()` call for scratch image creation as well.

pkg/vsphere/disk/config.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (d *VirtualDiskConfig) WithCapacity(capacity int64) *VirtualDiskConfig {
+	d.CapacityInKB = capacity
+
+	return d
+}
+
+func (d *VirtualDiskConfig) Hash() uint64 {
+	key := fmt.Sprintf("%s-%s-%d-%s", d.DatastoreURI, d.ParentDatastoreURI, d.CapacityInKB, d.DiskMode)
```
hickeng: guessing this is used for unique reference/lookup for a given vmdk - I don't think capacity should be in there... while I'm not sure how it works there is a live growth mechanism and it could easily be that the capacity has changed but it's still the same VMDK.

pkg/vsphere/disk/disk_manager.go
```diff
@@ -46,17 +46,28 @@ type Manager struct {
...
 
 	// The PCI + SCSI device /dev node string format the disks can be attached with
 	byPathFormat string
 
-	reconfig sync.Mutex
+	// serialize reconfigure operations
+	mu sync.Mutex
+
+	// map of URIs to VirtualDisk structs so that we can return the same instance to the caller, required for ref counting
+	Disks map[uint64]*VirtualDisk
```
hickeng: How are we populating this with existing disks - walk stores on start or lazily?

pkg/vsphere/disk/disk_manager.go
```diff
@@ -190,13 +196,12 @@ func (m *Manager) Create(op trace.Operation, newDiskURI *object.DatastorePath,
...
 		},
-
-		CapacityKb: capacityKB,
+		CapacityKb: config.CapacityInKB,
 	}
 
 	op.Infof("Creating vmdk for layer or volume %s", d.DatastoreURI)
 	err = tasks.Wait(op, func(ctx context.Context) (tasks.Task, error) {
-		return vdm.CreateVirtualDisk(ctx, d.DatastoreURI.String(), nil, spec)
+		return m.vdm.CreateVirtualDisk(ctx, d.DatastoreURI.String(), nil, spec)
```
hickeng: surprised to see us using CreateVirtualDisk - I'm unaware of a scenario when we're not going to then immediately attach it to a VM.
Can you clarify?

pkg/vsphere/disk/disk_manager.go
```diff
@@ -269,7 +272,11 @@ func (m *Manager) Get(op trace.Operation, diskURI *object.DatastorePath, fst Fil
 // }
 
 // Attach attempts to attach a virtual disk
-func (m *Manager) Attach(op trace.Operation, disk *types.VirtualDisk) error {
+func (m *Manager) Attach(op trace.Operation, config *VirtualDiskConfig) error {
+	defer trace.End(trace.Begin(""))
+
+	disk := m.toSpec(config)
```
hickeng: I don't see a check for whether the disk is already attached - given this method is exposed let's make it idempotent.

pkg/vsphere/disk/disk_manager.go
```diff
@@ -281,7 +288,9 @@ func (m *Manager) Attach(op trace.Operation, disk *types.VirtualDisk) error {
 	machineSpec := types.VirtualMachineConfigSpec{}
 	machineSpec.DeviceChange = append(machineSpec.DeviceChange, changeSpec...)
 
-	m.reconfig.Lock()
+	m.mu.Lock()
+	defer m.mu.Unlock()
```
hickeng: reference counting should be here?

pkg/vsphere/disk/disk_manager.go
```diff
@@ -302,8 +310,14 @@ func (m *Manager) Attach(op trace.Operation, disk *types.VirtualDisk) error {
...
-func (m *Manager) Detach(op trace.Operation, d *VirtualDisk) error {
-	defer trace.End(trace.Begin(d.DevicePath))
+func (m *Manager) Detach(op trace.Operation, config *VirtualDiskConfig) error {
+	defer trace.End(trace.Begin(""))
+
+	d, err := NewVirtualDisk(config, m.Disks)
+	if err != nil {
+		return errors.Trace(err)
+	}
+
```
hickeng: unclear why the `d.attachedRefs--` path is returning an error - please check that it's a soft error.

lib/portlayer/storage/vsphere/image.go
```diff
@@ -64,7 +64,7 @@ var (
 const (
 	StorageImageDir  = "images"
 	defaultDiskLabel = "containerfs"
-	defaultDiskSize  = 8388608
+	defaultDiskSize  = 8 * 1024 * 1024
```
caglar10ur: done

lib/portlayer/storage/vsphere/volume.go
caglar10ur: Agreed but this is so tangled right now so I rather not to touch it but keep the current logic. The bigger issue is the storage rest api is not simply acting like a proxy but doing a lot more than it. They call NewVolumeStore in the configure phase  which is called by portlayer Init during the initialization.

lib/portlayer/storage/vsphere/volume.go
caglar10ur: ImageStore/scratch has that already. Am I missing something?

pkg/vsphere/disk/config.go
caglar10ur: Done

pkg/vsphere/disk/disk_manager.go
caglar10ur: we don't, as far as I see we are detaching everything on the fresh start. Or maybe I misunderstand the ask :)

pkg/vsphere/disk/disk_manager.go
caglar10ur: Tests are the only user of this

pkg/vsphere/disk/disk_manager.go
caglar10ur: CreateAndAttach is the only caller, that is also why this is not taking any locks. I'm hoping to change that with a follow-up pr (planning to drop Create, provide only reconfigure based disk operations and a Wait function to make sure that disk becomes visible in the guest). 

I'll stop exporting this

pkg/vsphere/disk/disk_manager.go
caglar10ur: Ehehe welcome to the club :) CreateAndAttach is calling setAttached@disk and that is incrementing the count right now.

pkg/vsphere/disk/disk_manager.go
caglar10ur: I thought not being able to get VirtualDisk is a hard error, ref counting and reconfigure operation uses it

lib/portlayer/storage/vsphere/volume.go
hickeng: Means we cannot use these calls to construct the base disk spec for scratch, meaning even more of a special case.

pkg/vsphere/disk/disk_manager.go
hickeng: Let me rephrase - does this _only_ hold currently attached disks?

pkg/vsphere/disk/disk_manager.go
hickeng: That's racy - as best I can tell there's no lock that would prevent a parallel call detaching before we call setAttached.

pkg/vsphere/disk/disk_manager.go
hickeng: My comment is here because it's as close as I could get it to the line in question (#355 in the new code). I'm wondering why successfully decrementing the ref count is an error.

pkg/vsphere/disk/disk_manager.go
caglar10ur: Currently this holds the results of every NewVirtualDisk call for VCH, regardless of whether they are attached or detached. There is no eviction policy so this map keeps getting bigger and bigger with time. The manager doesn't know whether caller is done with the disk, that's why there is no clever way to evict items from this map. I'm thinking introducing a method (eg; MarkedAsDone) called by caller to signal DiskManager to drop that entry from the map.

pkg/vsphere/disk/disk_manager.go
caglar10ur: That is what the original code does, (https://github.com/vmware/vic/blob/master/pkg/vsphere/disk/disk_manager.go#L317) Maybe @jzt can shed a light :)

pkg/vsphere/disk/disk_manager.go
caglar10ur: Correct, the locking and consistency of the pkg is not something I really liked. There are two locks there, manager has a reconfigure lock which serializes reconfigure operations, and disks have a mutex to serialize attach/detach operations.

The way we use those locks are confusing. Disk.go uses locks for some of the exported functions but manager mostly calls non-exported without holding any lock. Will try to address this, stay tuned :)

caglar10ur: Pushed a new commit to introduce locking for every exported method on disk and diskmanager. DiskManager now also holds the locks during internal calls

pkg/vsphere/disk/disk_manager.go
hickeng: If we remove the map entry then GC will take care of deleting it only when the callers have dropped references. From the perspective of using it for locking, the manager would be going back to _this_ map each time (correct?), so is the caller holding a virtual disk that's not in sync a problem?
Can we evict on detach if this is purely for managing access to attached disks?
Not a fan of introducing what is essentially alloc/free.

caglar10ur: Done

pkg/vsphere/disk/disk_manager.go
caglar10ur: Grrr this is not going to work cause non-exported ones are calling exported functions...

hickeng: yeah - I'd mentioned that in an as yet unsubmitted commit. Given it's now against an outdated diff:
> A single idempotent entry point for having the disk attached/detached that reference counts and "does the right thing" seems less convoluted. Is there a reason you've found for calling setAttached separately from the Attach call? I've not run through the code explicitly looking for a reason so forgive me if it's something obvious.

pkg/vsphere/disk/disk_manager.go
```diff
@@ -161,20 +161,25 @@ func (m *Manager) CreateAndAttach(op trace.Operation, config *VirtualDiskConfig)
...
+		if detachErr := m.detach(op, disk); detachErr != nil {
+			op.Debugf("detach(%s) failed with %s", d.DatastoreURI, errors.ErrorStack(detachErr))
 		}
 
 		return nil, errors.Trace(err)
 	}
-
-	d.ParentDatastoreURI = config.ParentDatastoreURI
 	d.setAttached(blockDev)
```
caglar10ur: CreateAndAttach is WaitTillDiskAppearsInTheGuest and Attach is WaitTillReconfigure finishes, I'm guessing that's why it is separated that way (coming from the original code)

PR #5435, opened 2017-06-09
Add vic-machine configure ops and tprint support

### LINKED ISSUE #4495: vic-machine reconfigure: Update vCenter server credentials
**User Statement:**

As a VIC operation user, I need the ability to update the server credentials (username, password/SSO token and server thumbprint) for the appliance, in case the vCenter server credentials have been changed. See one example here #1879.

**Details:**
- supported operations:
  - update username
  - update password
  - update thumbprint
- hot update:
- input/output data format:
  - input: plain text, e.g., `--username=new-username -- password=new-pwd -- thumbprint=new-thumbprint`
  - output: N/A
- impact on VCH/containerVMs:
  - VCH: The changes will be persisted to the vmx file of the VCH. Then the portlayer will need to reload the VCH configure from the vmx file during VCH restart. 
  - containerVMs: no
- is upgrade needed: 

**Acceptance Criteria:**
-  add the ability to update the server credentials such as username, password/SSO token and server thumbprint though `vic-machine reconfigure`; and the changes will be applied after VCH restart
- add relevant test cases

anchal-agrawal:
Fixes #4495

### DIFF
--- a/lib/install/data/data.go
+++ b/lib/install/data/data.go
@@ -36,8 +36,7 @@ type Data struct {
 	common.Compute
 	common.VCHID
 
-	OpsUser     string `cmd:"ops-user"`
-	OpsPassword *string
+	OpsCredentials common.OpsCredentials
 
 	CertPEM     []byte
 	KeyPEM      []byte
@@ -252,6 +251,14 @@ func (d *Data) CopyNonEmpty(src *Data) error {
 		if err := d.copyContainerNetworks(src); err != nil {
 			return err
 		}
+	}
+
+	if src.OpsCredentials.IsSet {
+		d.OpsCredentials = src.OpsCredentials
+	}
+
+	if src.Target.Thumbprint != "" {
+		d.Target.Thumbprint = src.Target.Thumbprint
 	}
 
 	d.Timeout = src.Timeout
--- a/lib/install/validate/validator.go
+++ b/lib/install/validate/validator.go
@@ -417,16 +417,16 @@ func (v *Validator) managedbyVC(ctx context.Context) {
 
 func (v *Validator) credentials(ctx context.Context, input *data.Data, conf *config.VirtualContainerHostConfigSpec) {
 	// empty string for password is horrific, but a legitimate scenario especially in isolated labs
-	if input.OpsPassword == nil {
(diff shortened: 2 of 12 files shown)


lib/install/data/data.go
```diff
@@ -253,6 +252,14 @@ func (d *Data) CopyNonEmpty(src *Data) error {
 		}
 	}
 
+	if src.OpsCredentials.IsSet {
+		d.OpsCredentials = src.OpsCredentials
+	}
+
+	if src.Target.Thumbprint != "" {
```
emlin: feels like you can compare if they are same here

emlin: or just copy

emlin:
@anchal-agrawal did you try to test using --force? suppose user does not have to specify thumbprint with --force specified.

anchal-agrawal:
@emlin Good catch. If `--force` is specified and the input thumbprint is empty, it gets overwritten in the validator code. This change remains unaffected, so the following workflow works well:

* Create VCH with thumbprint `foo`
* Update thumbprint to `bar`
* `inspect --conf` - shows old thumbprint `foo`
* `configure --force ...` (no thumbprint specified)
* `inspect --conf` - shows new thumbprint `bar`

lib/install/validate/validator.go
```diff
@@ -417,16 +417,18 @@ func (v *Validator) managedbyVC(ctx context.Context) {
...
 
 	// check target with ops credentials
 	u := input.Target.URL
 
-	conf.Username = input.OpsUser
-	conf.Token = *input.OpsPassword
+	if input.OpsCredentials.OpsUser != nil {
+		conf.Username = *input.OpsCredentials.OpsUser
+	}
+	conf.Token = *input.OpsCredentials.OpsPassword
```
emlin: opspassword could be nil as well

anchal-agrawal: That's taken care of at the top of the function 😃

emlin: hmm.. no, actually you tested the password, but do not for the user, which is slightly confusing.
I think you might just do the same thing for the opsuser name, how do you think?

anchal-agrawal: Sounds good, fixed.

PR #5436, opened 2017-06-11
toolbox: implement remaining guest FileManager methods

dougm:
While we don't need the entire set of methods at the moment,
added all for completeness as kov will need some and also gets us closer
to govmomi/vcsim being able to use toolbox for testing outside of VIC.

Issue #4105


caglar10ur:
LGTM 👍

sflxn:
LGTM, but how do we have a way to get guest file attributes?  I only saw set attributes.

dougm:
@sflxn ListFilesInGuest to get guest file attributes

PR #5440, opened 2017-06-11
Add reference counting for mount and attach

### LINKED ISSUE #5396: Add reference counting to storage layer mounts
As a developer, I need to be sure that a disk is no longer in use before I detach it after an unmount.

Details:

Currently, the storage portlayer does not implement reference counting on mounted disks. This opens the door for a disk to be detached from the VCH while in use by other users in a concurrency situation.

In order to avoid premature detach after an unmount, we need to add a reference count to disk mounts (both the mount/umount operations keyed by target path, and the attach/detach operations keyed by PCI device path).

Basic flow:
1. `disk.go` in the storage portlayer needs to check before attach/mounting if the disk is already attached/mounted. If so, it needs to bump a reference count by one.
2. On unmount, the reference count should be decremented by one, and if the result is zero, the disk should be unmounted/detached.

jzt:
This change implements reference counting for keeping track of attached and mounted disks. Fixes #5396.

### DIFF
--- a/pkg/vsphere/disk/disk.go
+++ b/pkg/vsphere/disk/disk.go
@@ -20,18 +20,27 @@ import (
 	"strings"
 	"sync"
 
+	log "github.com/Sirupsen/logrus"
+
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/vic/pkg/fs"
 )
 
+// FilesystemType represents the filesystem in use by a virtual disk
 type FilesystemType uint8
 
 const (
+	// Ext4 represents the ext4 file system
 	Ext4 FilesystemType = iota + 1
+
+	// Xfs represents the XFS file system
 	Xfs
+
+	// Ntfs represents the NTFS file system
 	Ntfs
 )
 
+// Filesystem defines the interface for handling an attached virtual disk
 type Filesystem interface {
 	Mkfs(devPath, label string) error
 	SetLabel(devPath, labelName string) error
@@ -39,6 +48,8 @@ type Filesystem interface {
 	Unmount(path string) error
 }
 
+// FilesystemTypeToFilesystem returns a new Filesystem implementation
+// associated with the supplied FilesystemType
 func FilesystemTypeToFilesystem(fstype FilesystemType) Filesystem {
 	switch fstype {
 	case Xfs:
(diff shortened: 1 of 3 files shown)


pkg/vsphere/disk/disk.go
```diff
@@ -91,13 +106,22 @@ func (d *VirtualDisk) unlock() {
 
 func (d *VirtualDisk) setAttached(devicePath string) error {
 	if d.Attached() {
-		return fmt.Errorf("%s is already attached (%s)", d.DatastoreURI, devicePath)
```
caglar10ur: Not that your changes is wrong but maybe following could be little bit easier to read (aka bump to count when we don't return an error)

```
defer func() {
if err == nil {
    d.attachedRefs++
    log.Debugf("Increased attach references for %s to %d", d.DatastoreURI, d.attachedRefs)
}
```

pkg/vsphere/disk/disk_manager.go
```diff
@@ -310,6 +315,9 @@ func (m *Manager) Detach(op trace.Operation, d *VirtualDisk) error {
 	}
 
 	if err := d.canBeDetached(); err != nil {
+		// even though canBeDetached() is called here and nowhere else, it does not imply
+		// an attempt to detach, so decrease the ref count from here for that reason
+		d.attachedRefs--
```
emlin: Feels confused. If any error returned here, that means the disk cannot be detached, and no detach happens as well, why attachedRefs is reduced here?

jzt: `canBeDetached` will return an error if the disk is mounted, the disk is already detached, or the disk is in use by someone else (either mounted or attached). So basically, this client attempted to detach the disk because it no longer needs it, but the disk is in use elsewhere and thus can't be detached yet, but the reference count needs to be updated to show one less consumer using the disk.

emlin: gotcha, thanks 👍

pkg/vsphere/disk/disk.go
jzt: This is what I was originally going for but I couldn't think of a clean way to do it. Thanks!

PR #5451, opened 2017-06-13
[skip ci] WIP populate scratch with minimum OS structure

### LINKED ISSUE #489: Populate scratch with the minimum OS structure
There are some core files and directories that need to exist for the containers to function correctly.
For Linux these are:
/etc/hostname
/etc/hosts
/etc/resolv.conf
/proc
/sys
/lib/modules - needed for bindmount of appropriate kernel modules for insmod
/usr/lib/iptables - possibly needed, but needs confirmation

/.tether

These should be created on scratch so that their presence doesn't show up in image diffs.
The files are present so that they can be used as bindmount targets, not to be populated with specific data.

chengwang86:
fixes #489

### DIFF
--- a/isos/bootstrap/bootstrap
+++ b/isos/bootstrap/bootstrap
@@ -76,11 +76,30 @@ if mount -t ext4 /dev/disk/by-label/containerfs ${MOUNTPOINT}; then
     mkdir -p ${MOUNTPOINT}/usr/lib
     cp -r /usr/lib/iptables ${MOUNTPOINT}/usr/lib/
 
+    mkdir -m 755 -p ${MOUNTPOINT}/.tether/etc ${MOUNTPOINT}/.tether/lib/modules ${MOUNTPOINT}/.tether/usr/lib/iptables
+    mkdir -m 555 -p ${MOUNTPOINT}/.tether/proc ${MOUNTPOINT}/.tether/sys
+
+    touch ${MOUNTPOINT}/.tether/etc/hostname ${MOUNTPOINT}/.tether/etc/hosts ${MOUNTPOINT}/.tether/etc/resolv.conf
+    chmod 644 ${MOUNTPOINT}/.tether/etc/hostname ${MOUNTPOINT}/.tether/etc/hosts ${MOUNTPOINT}/.tether/etc/resolv.conf
+
     until [[ $(ls -1 /dev/disk/by-label | wc -l) -eq $(ls -1 /sys/block | wc -l) ]]; do sleep 0.1;done
 
     echo "switching to the new mount"
     systemctl switch-root ${MOUNTPOINT} /.tether/tether 2>&1
     echo "switched to the new mount"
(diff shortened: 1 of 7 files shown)


isos/bootstrap/bootstrap
```diff
@@ -76,11 +76,30 @@ if mount -t ext4 /dev/disk/by-label/containerfs ${MOUNTPOINT}; then
     mkdir -p ${MOUNTPOINT}/usr/lib
     cp -r /usr/lib/iptables ${MOUNTPOINT}/usr/lib/
 
+    mkdir -m 755 -p ${MOUNTPOINT}/.tether/etc ${MOUNTPOINT}/.tether/lib/modules ${MOUNTPOINT}/.tether/usr/lib/iptables
+    mkdir -m 555 -p ${MOUNTPOINT}/.tether/proc ${MOUNTPOINT}/.tether/sys
+
+    touch ${MOUNTPOINT}/.tether/etc/hostname ${MOUNTPOINT}/.tether/etc/hosts ${MOUNTPOINT}/.tether/etc/resolv.conf
```
jzt: Curious:  Is it necessary to touch these files in `${MOUNTPOINT}/.tether`?  I know they need to be created on scratch, but doesn't the tether create them here automatically?

isos/bootstrap/bootstrap
```diff
@@ -76,11 +76,30 @@ if mount -t ext4 /dev/disk/by-label/containerfs ${MOUNTPOINT}; then
...
     systemctl switch-root ${MOUNTPOINT} /.tether/tether 2>&1
     echo "switched to the new mount"
+
+    sleep 10s
+
+    mount --bind /.tether/etc/hostname /etc/hostname
+    mount --bind /.tether/etc/hosts /etc/hosts
+    mount --bind /.tether/etc/resolv.conf /etc/resolv.conf
+
+    sleep 1m
```
jzt: Sorry for commenting on your WIP :) Just curious if this 1m sleep is just for testing.

chengwang86: @jzt Yes both `sleep` are just used for my own testing. Caglar told me that the only way to check/debug the `bootstrap` was to watch the screen output of the containerVM from vsphere web client. Unless I add enough `sleep` time between cmds, the output from the photonOS would flush all the useful information I want to see. 

However, in my local test, it seemed like the extra sleep time didn't work.

isos/bootstrap/bootstrap
chengwang86: @jzt tether only creates these files under `/etc` but not `.tether/etc` which is why I need to create them manually in bootstrap.

jzt:
Should this PR be closed in favor of #5493?

chengwang86:
I will keep this PR for now for my reference and close it shortly.

chengwang86:
This is not needed anymore.

PR #5452, opened 2017-06-13
Generate mock layer.tar data

### LINKED ISSUE #5331: Mock up fake diff data in imagec
Acceptance criteria:

- [ ] Create mock data that can be used to work on image construction in imagec while we wait for the completion of docker diff.

---

Docker push cannot complete without docker diff.  While we wait for the completion of docker diff, we can use docker save to create us an image tar and hand create diff layer tars to mock up the data that docker diff would have returned.  This will allow docker push's image construction work to proceed.

sgairo:
Fixes #5331 

This data will be used as a stub until other components are finished, but will still be useful for unit testing afterward.


lib/imagec/imagec_test.go
```diff
@@ -96,8 +97,61 @@ const (
 	]
 }
 	`
+	//the following constants will be used for generating mock tar data
```
morris-jason: spaces needed after opening comment, ie `// ....`

sgairo: ty

morris-jason:
lgtm

sgairo:
As outlined in #5331, this approach will no longer work. Closing this PR.

PR #5474, opened 2017-06-14
Remove installer - moved to vic-product

andrewtchin:
Removed `installer` directory
Partial for 4424
Moved to https://github.com/vmware/vic-product/tree/master/installer


morris-jason:
yolo

PR #5476, opened 2017-06-14
Limit the lease on our servers, to prevent collisions

### LINKED ISSUE #5280: Nightly 6.5 5-11-Multiple-Cluster failure: incorrect login credentials in govc host.account.update
Seen while deploying ESX hosts:

```
govc host.account.update -id root -password ${NIMBUS_ESX_PASSWORD}
govc: ServerFaultCode: Cannot complete login due to an incorrect user name or password.' should be empty. 
```

[5-11-Multiple-Cluster.zip](https://github.com/vmware/vic/files/1033740/5-11-Multiple-Cluster.zip)

mhagen-vmware:
fixes #5280 

The problem with 5280 is that there were a bunch of servers left over from previous failed tests.  I manually cleaned up those servers and this change should prevent that from happening again as now they will only hang around for a day.

### DIFF
--- a/tests/manual-test-cases/Group5-Functional-Tests/5-3-Enhanced-Linked-Mode.robot
+++ b/tests/manual-test-cases/Group5-Functional-Tests/5-3-Enhanced-Linked-Mode.robot
@@ -33,7 +33,7 @@ Test
     Set Test Variable  ${user}  %{NIMBUS_USER}
     Log To Console  \nDeploying Nimbus Testbed: ${name}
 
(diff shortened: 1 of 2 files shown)


chengwang86:
SO now that we have set `--lease 1`, even if the test exits without cleanup the servers/testbeds, they will only stay in the system for 1 day, which may not affect the nightly run on the next day, right?

But it is still possible that test-5-3 exits abnormally, leaving its testbed not cleaned up whereas the next test-5-4 (started right after 5-3 exited) could have server name collision.

mhagen-vmware:
Correct, but we only have 1/1000 chance of that being a name collision.

PR #5478, opened 2017-06-14
[specific ci=Group11-Upgrade] reconfigure VCH cpu/memory

### LINKED ISSUE #4513: vic-machine reconfigure: Update the Virtual Container Host memory and CPU details via the vic-machine CLI
**User Statement:**

As a VIC operation user, I need the ability to update the VCH's CPU and memory details via the vic-machine CLI.

**Details:**
- supported operations (both vsphere UI and vic-machine cli):
  - update CPU details (cpu limit in MHz, cpu-reservation, cpu-shares)
  - update memory details (memory limit, memory-reservation, memory-shares)
- hot update:
  - yes.
- input/output data format:
  - input: e.g., `--cpu --cpu-reservation --cpu-shares --memory --memory-reservation --memory-shares`
  - output: N/A
- impact on VCH/containerVMs (setting CPU limit on vsphere UI as an example):
  - if we increase the limit to be more than the physical resource limit, vsphere will throw error msg `The available Memory resources in the parent resource pool are insufficient for the operation.`
  - if we increase the limit within the physical resource limit of the resource pool, the operation can succeed
  - if we decrease the limit to be less than its current usage, no error msgs. However, vch restart will fail due to lack of resource and we cannot run new containers.

- is upgrade needed: 
   -  no. We can directly call vSphere APIs to update the cpu and memory details of the resource pool. 

**Acceptance Criteria:**
- add the ability to update the VCH's cpu and memory details via vic-machone cli.
- add relevant test cases

emlin:
fixes #4513 
VCH CPU/Memory is set in VCH parent vApp/Resource pool, not part of meta data persisted in guestinfo, so in this PR, added a separate step to configure and rollback CPU/Memory settings in parent vApp/Resource pool.
If reconfiguration failed, the old CPU/Memory settings will be reverted.


cmd/vic-machine/common/limits.go
```diff
@@ -0,0 +1,78 @@
...
+
+import (
+	"gopkg.in/urfave/cli.v1"
+
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/pkg/flags"
+)
+
+type ResourceLimits struct {
+	VCHCPULimitsMHz       *int              `cmd:"cpu"`
```
chengwang86: I discussed with @gigawhitlocks about the usage of this `*int`. If the user does not specify `--VCHCPULimitsMHz`, the value of this variable will be `nil`, so that we know that the user does not want to change/reconfigure the current value for this option. 

Alternatively, we can define `-1` as the default value of this option. So if the user doesn't specify a value for this option, its value will be `-1` instead of `nil`. The benefit is that we don't need an additional datatype `OptionalInt`.

What do you think?

gigawhitlocks: This might be an appropriate use-case for an `iota` type. Same w/ the loglevel variable, honestly. Don't hold me to that though, just thinking aloud.

emlin: The problem is to differentiate default value and no input at all. With nil value of pointer int, we clearly know if user has that option or not, and no requirement for end user.
The downside of this option is how to show the default value. nil is not likely a good one.

But with general int value, we do not have a way to know if user has default value as input or no input at all. So the way is to define a special default value. If we find that value, we "think" the value is not from user input.
In this way, if no input for this parameter, cli will set the invalid value as option value. So our code has to accommodate the invalid value to make it work, e.g. default value for VCHCPUReservationMHz is -1. If we found that's -1, we need to change that value to 0, and move forward without any error message.
if user do have -1 as input, we'll move forward in the same way, even that's not a valid input.

So in this way, if customer used some special value, they'll fall into the black hole, without any useful message.

atm, I still prefer to keep optional int solution. But it's always good to have such discussion.

lib/install/management/configure.go
```diff
@@ -121,6 +133,39 @@ func (d *Dispatcher) Configure(vch *vm.VirtualMachine, conf *config.VirtualConta
 	return err
 }
 
+func (d *Dispatcher) rollbackResourceSettings(poolName string, settings *data.InstallerData) error {
+	if !settings.VCHSizeIsSet || d.oldVCHResources == nil {
+		log.Debugf("VCH resource settings is not changed")
```
anchal-agrawal: s/is/are

lib/install/management/configure.go
```diff
@@ -121,6 +133,39 @@ func (d *Dispatcher) Configure(vch *vm.VirtualMachine, conf *config.VirtualConta
...
+		return nil
+	}
+	return updateResourcePoolConfig(d.ctx, d.vchPool, poolName, d.oldVCHResources)
+}
+
+func (d *Dispatcher) updateResourceSettings(poolName string, settings *data.InstallerData) error {
+	if !settings.VCHSizeIsSet {
+		log.Debugf("VCH resource settings is not changed")
```
anchal-agrawal: ditto

lib/install/management/configure.go
```diff
@@ -121,6 +133,39 @@ func (d *Dispatcher) Configure(vch *vm.VirtualMachine, conf *config.VirtualConta
...
+		err = errors.Errorf("Failed to get parent resource pool %q: %s", poolName, err)
+		return err
+	}
+	oldSettings, err := d.getPoolResourceSettings(d.vchPool)
+	if err != nil {
+		err = errors.Errorf("Failed to get parent resource settings %q: %s", poolName, err)
+		return err
+	}
+	if reflect.DeepEqual(oldSettings, &settings.VCHSize) {
+		log.Debugf("VCH resource settings is same to old value")
```
anchal-agrawal: `VCH resource settings are same as old values`

lib/install/validate/config_to_data.go
```diff
@@ -108,25 +108,45 @@ func setVCHResources(ctx context.Context, vch *object.ResourcePool, d *data.Data
...
-func setResources(limit *int, reservation *int, shares **types.SharesInfo, allocation *types.ResourceAllocationInfo) {
-	if limit != nil && allocation.Limit > -1 {
+func HandleDefaultSettings(allocation *types.ResourceAllocationInfo) {
+	if allocation == nil {
+		return
+	}
+	if allocation.Limit == -1 {
+		allocation.Limit = 0
+	}
+	if allocation.Reservation == 1 {
```
anchal-agrawal: Question - why is this necessary?

emlin: During creation, to workaround govmomi omit empty value issue (which is generated from vsphere WSDL), we have to set reservation to 1 instead of 0.
But this 1 is not a custom setting, we don't want to show to that 1 in the user configuration, so reset that back to correct default.
Also add few comments in the code.

PR #5487, opened 2017-06-15
Give the system a break if it isn't ready for adding hosts

### LINKED ISSUE #5199: Nightly 6.0 HA test fails trying to add ESX host to VC cluster
No logs provided in the bundle, only the report.

```
Failed to add the host to the VC in 3 attempts
```

mhagen-vmware:
fixes #5199 

If the system is not succeeding to add a host, it doesn't help to keep hammering it, so let's sleep for a few minutes and try again after the sleep.

### DIFF
--- a/tests/resources/Vsphere-Util.robot
+++ b/tests/resources/Vsphere-Util.robot
@@ -203,6 +203,7 @@ Add Host To VCenter
     \   ${out}=  Run  govc cluster.add -hostname=${host} -username=${user} -dc=${dc} -password=${pw} -noverify=true
     \   ${status}=  Run Keyword And Return Status  Should Contain  ${out}  OK
     \   Return From Keyword If  ${status}
+    \   Sleep  3 minutes
     Fail  Failed to add the host to the VC in 3 attempts
 
 Get Host Firewall Enabled


tests/resources/Vsphere-Util.robot
```diff
@@ -203,6 +203,7 @@ Add Host To VCenter
     \   ${out}=  Run  govc cluster.add -hostname=${host} -username=${user} -dc=${dc} -password=${pw} -noverify=true
     \   ${status}=  Run Keyword And Return Status  Should Contain  ${out}  OK
     \   Return From Keyword If  ${status}
+    \   Sleep  3 minutes
```
anchal-agrawal: Does the sleep help with the `A general system error occurred: Timed out waiting for vpxa to start` error?

mhagen-vmware: no that is coming from govmomi, not much to do there AFAICT, but we should not hammer a system that is having problems.

anchal-agrawal: I'm curious because we see this particular error only with the `5-5-Heterogenous-ESXi` test. Could this be occurring due to the specific environment in this test? Also pinging @dougm for input.

mhagen-vmware: In this case you are right, we are adding a 6.5 host to a 6.0 vc which is not allowed, I have already fixed that also in this #5468 but I would still like to not hammer a sick VC system with repeated calls without break.

dougm: Retry seems fine since the resources were just created and, nimbus.  Don't think you need to wait so long in between attempts, but your call.

PR #5490, opened 2017-06-15
[skip ci] Add research spike output for #4854

### LINKED ISSUE #4854: DinV Dev: Investigate and propose developer-oriented deployment tool
**User Statement**

As a Docker developer using VIC, I want to be able to provision nested Docker hosts in VIC in using a tool that is familiar and simple. I want the tool to be able to manage the lifecycle of the nested Docker hosts, help me to configure it and help me secure it. I am already familiar with docker-machine.

**Details**

Our DinV images for developers need a consumption model. This means we need to make it super easy to configure, provision and secure the images being deployed. The most obvious way to do this would be to integrate docker-machine and point it at a VCH, but this comes with limitations due to the nature of the plugin model in docker-machine. As such, we need to make a decision about what tool we should use that satisfies the following requirements:

1) Configuration

It should be simple and intuitive to configure the following options:
- VM size (CPU and mem)
- Image cache volume (local or persistent, size)
- Docker engine options
- Easy to consume VIC engine certs for lifecycle management
- Show the config for a given nested host
- Show the state of a given nested host, in particular the free disk space in the image cache
- Configure a shell to connect to a given endpoint

2) Lifecycle

It should be simple and intuitive to manage the lifecycle of nested docker hosts
- List the hosts
- Update certificates
- Copy files into and out of the docker host (use ``docker cp`` in VIC)
- Run a shell in the nested docker host (use ``docker exec`` in VIC) - don't assume SSH
- Upgrade: Replace the nested docker host with an updated one. Keep the certs, keep the persistent image cache if it exists.

3) Roll your own

It should be simple to build your own nested docker host. This needs some thought, but there will almost certainly be companies which for compliance reasons, want to configure a DinV host in a particular way. This may be as simple as documenting how to do it and creating example Dockerfiles from VMware. However, I think there's definitely scope for making it slicker than that. 

At the very least, it should be: 
- obvious how to add simple daemon processes to the image
- obvious how to use a custom distro / patch set
- obvious how to add sshd if this is deemed necessary
- clear what the dependencies need to be and the way to "do it correctly"

The question of whether this is part of the tool, whether we use something like LinuxKit or whether this should be just a documentation exercise should be part of the output here.

4) Secure it

Spinning up a Docker host into vSphere with a remote API is a potentially big security risk if the right precautions are not taken. At the very least, we should make it:

- Simple to configure and consume TLS comms with the remote API
- Simple to configure secure communications with a local registry
- Well-integrated with whatever vulnerability scanning and image signing capabilities we introduce
- We should investigate whether there's integration we can do via Guest Tools or VIC logging to provide an audit log of the commands executed in the nested host.

**Acceptance Criteria**

Output of this Issue should be a recommendation for what tool or tools should be used as the consumption model that address the above in the most effective manner. This should include some control flows to illustrate user experience for each of the points raised.

Regardless of the outcome, there should be a specific analysis done of what we can or can't do with docker-machine so that the decision to integrate or not integrate is well-justified.

frapposelli:
This is what closed #4854

### DIFF
--- a/doc/design/dinv-dev-tool.md
+++ b/doc/design/dinv-dev-tool.md
@@ -1,0 +1,106 @@
+# Goal
+
+The goal for this investigation, is to find a suitable tool to enable "easy-button" provisioning of Docker hosts on top of VIC, also referred to as DinV, or Docker-in-VIC.
+
+The short term goal is to provide a tool for the user to provision a prebuilt docker image containing a full fledged docker host and make it available to the user, automating the provisioning and security part (deployment, creation of certificates, registry cache persistency) as much as possible, giving the developers a tool to self-provision docker hosts with an easy button.
+
+# Options
+
+We're currently evaluating two options: use Docker Machine, or build our own tool.
+
+# Investigation: Why docker machine isn't a good choice.
+
+Docker machine is an open source tool published by docker, helps users in getting started with creating docker hosts, it works on various platforms, including desktops (Virtualbox, HyperV, Fusion), mega clouds (aws, azure, google cloud) and private clouds (vsphere, openstack).
+
+Docker machine has a modular plugin architecture for the infrastructure part, but lacks a modular provisioning system, which means we cannot provision anything outside the OS/binaries that are built into its core.
+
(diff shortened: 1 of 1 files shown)


frapposelli:
ping @hickeng 

this has been waiting for some time

PR #5491, opened 2017-06-15
toolbox: Add support for directory transfer to/from guests

dougm:
Issue #4105

### DIFF
--- a/pkg/vsphere/toolbox/hgfs/archive.go
+++ b/pkg/vsphere/toolbox/hgfs/archive.go
@@ -1,0 +1,227 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package hgfs
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+)
+
+// archive implements the hgfs.File and os.FileInfo interfaces.
+type archive struct {
+	name string
+	size int64
+	done func() error
+
+	io.Reader
+	io.Writer
+}
+
+// Name implementation of the os.FileInfo interface method.
+func (a *archive) Name() string {
+	return a.name
+}
+
+// Size implementation of the os.FileInfo interface method.
+func (a *archive) Size() int64 {
+	return a.size
+}
+
+// Mode implementation of the os.FileInfo interface method.
+func (a *archive) Mode() os.FileMode {
+	return 0600
+}
+
+// ModTime implementation of the os.FileInfo interface method.
+func (a *archive) ModTime() time.Time {
+	return time.Now()
+}
+
+// IsDir implementation of the os.FileInfo interface method.
+func (a *archive) IsDir() bool {
+	return false
+}
+
+// Sys implementation of the os.FileInfo interface method.
+func (a *archive) Sys() interface{} {
+	return nil
+}
+
+// The trailer is required since TransferFromGuest requires a Content-Length,
+// which toolbox doesn't know ahead of time as the gzip'd tarball never touches the disk.
+// HTTP clients need to be aware of this and stop reading when they see the 2nd gzip header.
+var gzipHeader = []byte{0x1f, 0x8b, 0x08} // rfc1952 {ID1, ID2, CM}
+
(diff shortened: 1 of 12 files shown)


pkg/vsphere/toolbox/hgfs/archive_test.go
```diff
@@ -0,0 +1,263 @@
...
+func TestReadArchive(t *testing.T) {
+	Trace = testing.Verbose()
+
+	dir, err := ioutil.TempDir("", "toolbox-")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	nfiles := 5
+	subdir := path.Join(dir, "hangar-18")
```
frapposelli: nice homage 😉

pkg/vsphere/toolbox/hgfs/protocol.go
```diff
@@ -282,7 +282,11 @@ func (r *Packet) Reply(payload interface{}, err error) ([]byte, error) {
 	}
 
 	if Trace {
-		fmt.Fprintf(os.Stderr, "[hgfs] response %#v\n", p.Header)
+		rc := "OK"
+		if err != nil {
+			rc = err.Error()
+		}
+		fmt.Fprintf(os.Stderr, "[hgfs] response %#v [%s]\n", p.Header, rc)
```
hickeng: I think so, but does this get correctly redirected to tether.debug?

pkg/vsphere/toolbox/hgfs/archive.go
```diff
@@ -0,0 +1,223 @@
...
+
+		switch header.Typeflag {
+		case tar.TypeDir:
+			err = os.MkdirAll(name, mode)
+		case tar.TypeReg:
+			_ = os.MkdirAll(filepath.Dir(name), 0700)
+
+			var f *os.File
+
+			f, err = os.OpenFile(name, os.O_CREATE|os.O_RDWR, mode)
```
hickeng: why different flags vs regular file? `os.O_WRONLY | os.O_CREATE | os.O_TRUNC`

pkg/vsphere/toolbox/hgfs/server.go
```diff
@@ -106,8 +109,75 @@ func (s *Server) Dispatch(packet []byte) ([]byte, error) {
...
+		}
+	}
+
+	switch mode {
+	case OpenModeReadOnly:
+		file, err = os.Open(name)
+	case OpenModeWriteOnly:
+		flag := os.O_WRONLY | os.O_CREATE | os.O_TRUNC
+		file, err = os.OpenFile(name, flag, 0600)
+	default:
```
hickeng: presume we hit this if `!s.Archive && info.IsDir()`? Should we be returning the VIX_E_NOT_A_FILE error you mention in the comment below?
Ah - I see it's translated in vix_command.go

pkg/vsphere/toolbox/hgfs/server.go
```diff
@@ -440,9 +490,11 @@ func (s *Server) ReadV3(p *Packet) (interface{}, error) {
 
 	buf := make([]byte, req.RequiredSize)
 
-	n, err := file.ReadAt(buf, int64(req.Offset))
-	if err != nil {
-		if err != io.EOF || n <= 0 {
+	// Use ReadFull as Read() of an archive io.Pipe may return much smaller chunks,
+	// such as when we've read a tar header.
+	n, err := io.ReadFull(file, buf)
```
hickeng: Not sure how it would play out, but should we be implementing ReadFrom/WriteTo for efficient copy in any of these objects?

pkg/vsphere/toolbox/hgfs/archive.go
hickeng: Is there a uid/gid entry in the tar as well?

pkg/vsphere/toolbox/hgfs/protocol.go
dougm: No, its always false within the tether.  Its mainly useful to watch traffic when developing new RPCs and can be enabled in tests and toolbox main.

pkg/vsphere/toolbox/hgfs/server.go
dougm: Don't think we need ReadFrom/WriteTo in this case (in memory io.Pipe), but we could reuse the `buf` by attaching it to the archive struct (and hence the hgfs session).

pkg/vsphere/toolbox/hgfs/archive.go
dougm: Yes, we do need to chown(), may just add that as a TODO comment for now.

hickeng: @dougm can you add a follow up issue as TODOs are not usefully tracked.

pkg/vsphere/toolbox/hgfs/archive.go
```diff
@@ -0,0 +1,227 @@
...
+				err = f.Close()
+				if cerr != nil {
+					err = cerr
+				}
+			}
+		case tar.TypeSymlink:
+			err = os.Symlink(header.Linkname, name)
+		}
+
+		// TODO: Uid/Gid may not be meaningful here without some mapping.
```
dougm: Created issue #5550 as a start, we may want a separate issue specific to how id mapping would be implemented.

PR #5492, opened 2017-06-15
Bring imagec binary back for dev tests

### LINKED ISSUE #5330: Add imagec binary build back into VIC for dev purposes
Acceptance criteria:

- [ ] Update makefile so we can test the transfer component of docker push while we wait for docker diff to complete.

---

Docker push has a dependency on docker diff.  We cannot do the image construction part of push without diff.  If we reassembled the imagec binary, we can at least test docker push on a tagged image or at least work on that transfer code while we wait for docker diff to complete.

chengwang86:
fixes #5330 

This is only used for dev test purposes so there is no need to add integration test.
 
To run it in standalone mode without starting persona/portlayer, run `sudo bin/imagec --reference=busybox:latest --operation=push --standalone`; this cmd will pull `busybox` from dockerhub by default. 

In standalone mode, the image tar will be written to a local folder `images`

### DIFF
--- a/cmd/imagec/hook.go
+++ b/cmd/imagec/hook.go
@@ -1,0 +1,47 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+	"io"
+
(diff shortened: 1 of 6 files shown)


cmd/imagec/hook.go
```diff
@@ -0,0 +1,47 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
```
sgairo: 2017 :)

lib/imagec/imagec.go
```diff
@@ -107,6 +106,9 @@ type Options struct {
 
 	// RegistryCAs will not be modified by imagec
 	RegistryCAs *x509.CertPool
+
+	// If being set to true, do not bother portlayer or persona
```
jzt: `s/being set to//g`

chengwang86:
From ci test:
```
building /go/bin/gas...
checking security problems
make: *** [gas] Error 1
Makefile:231: recipe for target 'gas' failed
```
I need to figure out where the `gas` check was violated.

sflxn:
Can you remove it from most and other make targets?  This is just for development purposes so we should just build it if we call it out at the command line.

sflxn:
lgtm.  just rebase and squash.

PR #5493, opened 2017-06-15
populate scratch with min os structure

chengwang86:
This is about the first part of #489 . When creating the `scratch` during `image-store` creation, I also create certain files/folders required as the minimum OS structure so that these files/folders do not show up in `docker diff`.

Next step (in a separate PR): bind-mount these files/folders as discussed [here](https://github.com/vmware/vic/issues/489#issuecomment-308573659).

### DIFF
--- a/lib/portlayer/storage/vsphere/image.go
+++ b/lib/portlayer/storage/vsphere/image.go
@@ -55,6 +55,8 @@ var (
 		"/proc":             0555,
 		"/sys":              0555,
 		"/usr/lib/iptables": 0755,
+		// The permission of .tether should be drwxrwxrwt.
+		// The sticky bit 't' is added when mounting the tmpfs in bootstrap
 		"/.tether":          0777,
 	}
 )
--- a/lib/portlayer/storage/vsphere/image.go
+++ b/lib/portlayer/storage/vsphere/image.go
@@ -49,14 +49,14 @@ var (
 		"/etc/hosts":       0644,
 		"/etc/resolv.conf": 0644,
 	}
+	// Here the permission of .tether should be drwxrwxrwt.
+	// The sticky bit 't' is added when mounting the tmpfs in bootstrap
 	DirForMinOS = map[string]os.FileMode{
 		"/etc":              0755,
 		"/lib/modules":      0755,
 		"/proc":             0555,
 		"/sys":              0555,
 		"/usr/lib/iptables": 0755,
-		// The permission of .tether should be drwxrwxrwt.
-		// The sticky bit 't' is added when mounting the tmpfs in bootstrap
 		"/.tether":          0777,
 	}
 )
--- a/lib/portlayer/storage/vsphere/image.go
+++ b/lib/portlayer/storage/vsphere/image.go
@@ -662,7 +662,7 @@ func imagesInUse(op trace.Operation, ID string) error {
 	return nil
 }
 
-// populate the scratch with minimum OS structure refined in fileMinOS and dirMinOS
+// populate the scratch with minimum OS structure defined in FileForMinOS and DirForMinOS
 func populateMinOS(op trace.Operation, vmdisk *disk.VirtualDisk) error {
 	// tmp dir to mount the disk
 	dir, err := ioutil.TempDir("", "mnt-"+portlayer.Scratch.ID)
@@ -679,7 +679,7 @@ func populateMinOS(op trace.Operation, vmdisk *disk.VirtualDisk) error {
 		dirPath := fmt.Sprintf("%s%s", dir, dname)
 		m, err := strconv.ParseUint(dmode, 0, 0)
 		if err != nil {
-			op.Errorf("Failed to ParseUInt for %s: %e", dmode, err)
+			op.Errorf("Failed to ParseUint for %s: %e", dmode, err)
 		}
 		if err = os.MkdirAll(dirPath, os.FileMode(m)); err != nil {
 			op.Errorf("Failed to create directory %s: %s", dirPath, err)
@@ -697,7 +697,7 @@ func populateMinOS(op trace.Operation, vmdisk *disk.VirtualDisk) error {
 		}
 		n, err := strconv.ParseUint(fmode, 0, 0)
 		if err != nil {
-			op.Errorf("Failed to ParseUInt for %s: %e", fmode, err)
+			op.Errorf("Failed to ParseUint for %s: %e", fmode, err)
 		}
 		err = f.Chmod(os.FileMode(n))
 		if err != nil {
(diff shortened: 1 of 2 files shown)


lib/portlayer/storage/vsphere/image.go
```diff
@@ -641,3 +661,61 @@ func imagesInUse(op trace.Operation, ID string) error {
...
+	for fname, fmode := range FileForMinOS {
+		filePath := fmt.Sprintf("%s%s", dir, fname)
+		f, err := os.Create(filePath)
+		if err != nil {
+			op.Errorf("Failed to create file %s: %s", filePath, err)
+			return err
+		}
+		n, err := strconv.ParseUint(fmode, 0, 0)
+		if err != nil {
+			op.Errorf("Failed to ParseUInt for %s: %e", fmode, err)
```
jzt: `s/ParseUInt/ParseUint`

lib/portlayer/storage/vsphere/image.go
```diff
@@ -641,3 +661,61 @@ func imagesInUse(op trace.Operation, ID string) error {
 
 	return nil
 }
+
+// populate the scratch with minimum OS structure refined in fileMinOS and dirMinOS
```
jzt: `s/refined/defined`
`s/fileMinOS/FileForMinOS`
`s/dirMinOS/DirForMinOS`

lib/portlayer/storage/vsphere/image.go
```diff
@@ -641,3 +661,61 @@ func imagesInUse(op trace.Operation, ID string) error {
...
+
+	if err = vmdisk.Mount(dir, nil); err != nil {
+		op.Errorf("Failed to mount device %s to dir %s", vmdisk.DevicePath, dir)
+	}
+
+	for dname, dmode := range DirForMinOS {
+		dirPath := fmt.Sprintf("%s%s", dir, dname)
+		m, err := strconv.ParseUint(dmode, 0, 0)
+		if err != nil {
+			op.Errorf("Failed to ParseUInt for %s: %e", dmode, err)
```
jzt: `s/ParseUInt/ParseUint`

jzt:
LGTM with some minor edits

lib/portlayer/storage/vsphere/image.go
```diff
@@ -41,7 +42,22 @@ import (
...
+		"/etc/hostname":    "0644",
+		"/etc/hosts":       "0644",
+		"/etc/resolv.conf": "0644",
+	}
+	DirForMinOS = map[string]string{
+		"/etc":              "0755",
+		"/lib/modules":      "0755",
+		"/proc":             "0555",
+		"/sys":              "0555",
+		"/usr/lib/iptables": "0755",
```
hickeng: `/.tether`

lib/portlayer/storage/vsphere/image.go
```diff
@@ -641,3 +661,61 @@ func imagesInUse(op trace.Operation, ID string) error {
...
+func populateMinOS(op trace.Operation, vmdisk *disk.VirtualDisk) error {
+	// tmp dir to mount the disk
+	dir, err := ioutil.TempDir("", "mnt-"+portlayer.Scratch.ID)
+	if err != nil {
+		op.Errorf("Failed to create tempDir: %s", err)
+	}
+	defer os.RemoveAll(dir)
+
+	if err = vmdisk.Mount(dir, nil); err != nil {
+		op.Errorf("Failed to mount device %s to dir %s", vmdisk.DevicePath, dir)
```
hickeng: return here?

lib/portlayer/storage/vsphere/image.go
```diff
@@ -641,3 +661,61 @@ func imagesInUse(op trace.Operation, ID string) error {
...
+	}
+
```
hickeng: defer unmount here - means the umount will occur before the deferred recursive deletion of directories...

lib/portlayer/storage/vsphere/image.go
```diff
@@ -41,7 +42,22 @@ import (
 var StorageParentDir = "VIC"
 
 // Set to false for unit tests
-var DetachAll = true
+var (
+	DetachAll = true
+
+	FileForMinOS = map[string]string{
```
hickeng: `map[string]os.FileMode` - means we don't have to do the runtime conversion all the time and cleans up the loops.

lib/portlayer/storage/vsphere/image.go
```diff
@@ -641,3 +661,61 @@ func imagesInUse(op trace.Operation, ID string) error {
...
+		op.Errorf("Failed to create tempDir: %s", err)
+	}
+	defer os.RemoveAll(dir)
+
+	if err = vmdisk.Mount(dir, nil); err != nil {
+		op.Errorf("Failed to mount device %s to dir %s", vmdisk.DevicePath, dir)
+	}
+
+	for dname, dmode := range DirForMinOS {
+		dirPath := fmt.Sprintf("%s%s", dir, dname)
```
hickeng: `path.Join(dir, dname)` - adds system appropriate path separator and makes it obvious it's path manipulation rather than just string manipulation

lib/portlayer/storage/vsphere/image.go
```diff
@@ -641,3 +661,61 @@ func imagesInUse(op trace.Operation, ID string) error {
...
+		if err = os.MkdirAll(dirPath, os.FileMode(m)); err != nil {
+			op.Errorf("Failed to create directory %s: %s", dirPath, err)
+			return err
+		}
+	}
+
+	// The directory has to exist before creating the new file
+	for fname, fmode := range FileForMinOS {
+		filePath := fmt.Sprintf("%s%s", dir, fname)
+		f, err := os.Create(filePath)
```
hickeng: If you use the following, along with the mode suggestion for the directory map then this loop becomes much smaller:
`os.OpenFile(filepath, os.O_CREATE, mode)`

lib/portlayer/storage/vsphere/image.go
```diff
@@ -641,3 +663,55 @@ func imagesInUse(op trace.Operation, ID string) error {
 
 	return nil
 }
+
+// populate the scratch with minimum OS structure defined in FileForMinOS and DirForMinOS
+func populateMinOS(op trace.Operation, vmdisk *disk.VirtualDisk) error {
```
hickeng: if you name the error return then you can return an error from the deferred function blocks (see https://play.golang.org/p/KySTHY2gM4):
```go
func populateMinOS(op trace.Operation, vmdisk *disk.VirtualDisk) (err error)
```
generally naming returns tends to be confusing, but it's the only way I'm aware of to allow deferred functions to contribute to the returned values.

PR #5500, opened 2017-06-16
[specific ci=Group11-Upgrade]Remove old key from guestinfo

### LINKED ISSUE #5472: Remove extraconfig keys given empty strings
**User Statement:**

For vic-machine configure, before vm reconfiguration, remove extraconfig keys when empty string is given as a value.

**Details:**
- in the reconfigure code, before call to vm reconfiguration, query back the existing extraconfig
- compare the new extraconfig keys with the old one, check if the key has vic prefix (otherwise, it will override system extraconfig key)
- if there is some key not found in new key/value pairs, with vic prefix, set the value to ""

**Expected Outcome**
- Example: When a VCH is created using default DHCP, after it's configured to some dns servers, giving empty string arguments to --dns-server would clear the dns servers.

emlin:
After vic-machine reconfiguration, there might have some guestinfo key not needed, we need to have a way to remove old key.
This PR pull in the fix in govmomi, and set old key to empty value explicitly in vic-machine, to make sure old key/value is removed from guestinfo.
Reconfigure DNS rely on this feature, so reenabled `vic-machine configure --dns-server` option.
Fixes #5472


vendor/github.com/vmware/govmomi/govc/device/cdrom/insert.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright (c) 2014-2015 VMware, Inc. All Rights Reserved.
+Copyright (c) 2014-2016 VMware, Inc. All Rights Reserved.
```
rogeliosanchez: @emlin Why do some files have the copyright range from 2014-2017 and some have the copyright range from 2014-2016 like this one?  I'm just curious here.

emlin: This the vendor code, so it's different with our own copyright.
In general, the left number means which year the file is created, and the right number means which year the last change happens.

Issue #5485, opened 2017-06-15
syslogd hook firing when not configured


hickeng:
Found as part of #5445

When running stress loads to recreate #5445 we occasionally get into a state where we're firing the syslogd hook, despite it not being configured.
Need to check what the defaults are - the panic from below was injected into the error paths in `log/syslog/writer.go:queueWrite`.
```
Jun 15 2017 18:22:41.075Z DEBUG HandshakeClient: Sending syn with pos 182
Jun 15 2017 18:22:41.075Z ERROR syn: write failed
panic: unexpected

goroutine 775 [running]:
github.com/vmware/vic/pkg/log/syslog.(*writer).queueWrite(0xc420068990, 0x3, 0x7ffe54d06f18, 0x11, 0xc4208a3bc0, 0x53, 0xc420b74af8, 0x40220e)
	/home/vagrant/vicsmb/src/github.com/vmware/vic/pkg/log/syslog/writer.go:165 +0x152
github.com/vmware/vic/pkg/log/syslog.(*writer).Err(0xc420068990, 0xc4208a3bc0, 0x53, 0x40a3be, 0xc420b74b77)
	/home/vagrant/vicsmb/src/github.com/vmware/vic/pkg/log/syslog/writer.go:130 +0x5a
github.com/vmware/vic/pkg/log/syslog.(*Hook).writeEntry(0xc4202f5c70, 0xc42082ea50, 0xc4204b6540, 0x1d08560)
	/home/vagrant/vicsmb/src/github.com/vmware/vic/pkg/log/syslog/hook.go:61 +0x9c
github.com/vmware/vic/pkg/log/syslog.(*Hook).Fire(0xc4202f5c70, 0xc42082ea50, 0xc420b74b77, 0xc4204b6580)
	/home/vagrant/vicsmb/src/github.com/vmware/vic/pkg/log/syslog/hook.go:45 +0x35
github.com/vmware/vic/vendor/github.com/Sirupsen/logrus.LevelHooks.Fire(0xc420010210, 0xc4047b3e02, 0xc42082ea50, 0x53, 0xc420b74c18)
	/home/vagrant/vicsmb/src/github.com/vmware/vic/vendor/github.com/Sirupsen/logrus/hooks.go:28 +0x8b
[...]

Issue #3662, opened 2017-01-18
Component level logging


hmahmood:

vburenin:
not sure this is the right place, however, if we are going to use our own logger I would consider having a little more virtual log levels.

1. critical
2. error
3. warning
4. info
5. debug
6. trace - to log large chunks of info
7. sensitive - to log user credentials, cookies, etc.

fdawg4l:
Just echoing what @vburenin said here;  a way to toggle these levels per component at runtime would also be nice.

dbarkelew:
Would this tie into the logging set with --debug [1-3] flag options? 
https://vmware.github.io/vic-product/assets/files/html/1.3/vic_vsphere_admin/vch_general_settings.html#debug 

Or would these logging levels be set in a different way?

PR #5507, opened 2017-06-16
Addressing logging issues

### LINKED ISSUE #5388: Replace placeholder code for squashing verbose logging in config package
Found while working on #4872 

**Story**
As a VIC developer I expect package level log verbosity decisions to be isolated to that package

**Detail**
When first adding the extraconfig package it was extremely verbose and, lacking package level logging levels (#3662), there was placeholder code added to squash the debug output.
This placeholder code is still present and is not safe for concurrent calls - it's switching the global logging level to INFO and it's a ratchet style effect meaning once it's switched once it'll never switch back.

While the full fix here should be done as part of #3662 we need to fix this behaviour:
1. Switch to a dedicated package logger as per the trace package
2. Configure that logger to DEBUG only when the DebugLevel is >= 3 (as it's extremely verbose).

**Acceptance**
- [ ] extraconfig package has a dedicated logger with it's own log level

hickeng:
* Adds separate loggers for encode and decode of extraconfig packages which are
extremely verbose. #3662 would be the full fix for this, but this was impacting
serviceability in the field
* Fixes syslogd configuration when no syslogd specified.
* Reduces unnecessarily verbose build output

Fixes #5388 #5485

### DIFF
--- a/pkg/vsphere/extraconfig/encode.go
+++ b/pkg/vsphere/extraconfig/encode.go
@@ -19,35 +19,17 @@ import (
 	"errors"
 	"fmt"
 	"net"
-	"os"
 	"reflect"
 	"sort"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
-
-	"github.com/Sirupsen/logrus"
-
-	"github.com/vmware/vic/pkg/log"
 )
 
 var (
 	ErrKeyNotFound = errors.New("key not found")
 )
-
-var encodeLogger = &logrus.Logger{
-	Out: os.Stderr,
-	// We're using our own text formatter to skip the \n and \t escaping logrus
-	// was doing on non TTY Out (we redirect to a file) descriptors.
-	Formatter: log.NewTextFormatter(),
-	Hooks:     make(logrus.LevelHooks),
-	Level:     logrus.InfoLevel,
-}
-
-func SetEncodeLogLevel(level logrus.Level) {
-	encodeLogger.Level = level
-}
 
 type encoder func(sink DataSink, src reflect.Value, prefix string, depth recursion)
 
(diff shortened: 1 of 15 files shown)


pkg/vsphere/extraconfig/encode.go
```diff
@@ -19,22 +19,36 @@ import (
...
+	"github.com/vmware/vic/pkg/log"
 )
 
 var (
-	// EncodeLogLevel value
-	EncodeLogLevel = log.InfoLevel
 	ErrKeyNotFound = errors.New("key not found")
 )
 
+var encodeLogger = &logrus.Logger{
```
hmahmood: Why the separate loggers for encode and decode? Generally I thought one logger per package was a good idea. Additionally you are setting log levels for both every time you set it for one, so you may be able to just use one logger for the extraconfig package.

hmahmood: Why the separate loggers for encode and decode? Generally I thought one logger per package was a good idea. Additionally you are setting log levels for both every time you set it for one, so you may be able to just use one logger for the extraconfig package.

pkg/vsphere/extraconfig/keys.go
```diff
@@ -161,7 +161,7 @@ func calculateScopeFromKey(key string) []string {
 	return scopes
 }
 
-func calculateKeyFromField(field reflect.StructField, prefix string, depth recursion) (string, recursion) {
+func calculateKeyFromField(log logrus.FieldLogger, field reflect.StructField, prefix string, depth recursion) (string, recursion) {
```
hmahmood: This is very unusual. Is this a side effect of two loggers?

hickeng: We call this from locations outside of the encode/decode paths so either we get no debug from this when those calls are set to debug or we pass the logger through.
One of the perks of moving to trace everywhere - option to use operation level debug insteadof/combinedwith component level debug.

pkg/vsphere/extraconfig/encode.go
hickeng: Both paths are very verbose and I only cared about one while debugging - didn't see a reason not to provide for this flexibility for now given it's trivial extra work at this point.

pkg/vsphere/extraconfig/keys.go
hmahmood: You could just have one logger in the package and use that here. This looks very strange passing in a logger to a function. I would be ok using `trace.Operation` here. But I think using just one package level logger would be much simpler for now.

hickeng: This basically requires either a choice between encode/decode logger or collapsing those back into one. I'm inclined to leave it as is then swap it out for trace operation asap.

hmahmood: I think changes like moving to trace.Operation should be separate from this PR, and part of a bigger move towards using context's in our APIs. I would prefer if you would stick with one logger for now, especially since we're not leveraging the two loggers differently anyway. Thanks for putting this fix up so quickly.

hickeng:
@hmahmood made the change to a single package logger - please sanity check.

hmahmood:
@hickeng too late now, but it looks good for what it is worth.

PR #5514, opened 2017-06-17
Create .tether folder in bootstrap

chengwang86:
This should fix all the test failures seen in ci build 11638.

The reason is:
In PR #5493 , the `.tether` folder is already created in `scratch` when creating the image-store, therefore there is no need to `mkdir /.tether` in `bootstrap`. However, in the VCH upgrade related tests, the image-store was created by an older-version VCH which didn't create the `.tether` folder in the scratch vmdk. So VM creation from an upgraded VCH would fail when running the bootstrap.


matthewavery:
lgtm

PR #5521, opened 2017-06-18
bind-mount files required by the minimum os structure at tether startup

### LINKED ISSUE #489: Populate scratch with the minimum OS structure

chengwang86:
This is the second half required by #489 . 

Fixes #489 

- This PR bind-mounts the files required by the minimum os structure based on the discussion here https://github.com/vmware/vic/issues/489#issuecomment-308573659

- This PR creates the target files (`/etc/hosts, /etc/hostname, /etc/resolv.conf`) at tether startup since they may not present in the scratch disk if the image-store is created by an older-version VCH.

### DIFF
--- a/lib/etcconf/etcconf.go
+++ b/lib/etcconf/etcconf.go
@@ -106,5 +106,11 @@ func save(filePath string, walker EntryWalker) error {
 		return err
 	}
 
+	// make sure the file is readable
+	// #nosec: Expect file permissions to be 0600 or less
+	if err := os.Chmod(filePath, 0644); err != nil {
+		return err
+	}
+
 	return nil
 }
--- a/lib/etcconf/etcconf.go
+++ b/lib/etcconf/etcconf.go
@@ -36,7 +36,7 @@ type EntryWalker interface {
 
 type Conf interface {
 	Load() error
-	Save() error
+	Save(string) error
 }
 
 func load(filePath string, con EntryConsumer) error {
--- a/lib/etcconf/etcconf.go
+++ b/lib/etcconf/etcconf.go
@@ -106,11 +106,5 @@ func save(filePath string, walker EntryWalker) error {
 		return err
 	}
 
-	// make sure the file is readable
-	// #nosec: Expect file permissions to be 0600 or less
-	if err := os.Chmod(filePath, 0644); err != nil {
-		return err
-	}
-
 	return nil
 }
--- a/lib/etcconf/etcconf.go
+++ b/lib/etcconf/etcconf.go
@@ -36,7 +36,8 @@ type EntryWalker interface {
 
 type Conf interface {
 	Load() error
-	Save(string) error
+	Save() error
+	Path() string
 }
 
 func load(filePath string, con EntryConsumer) error {
--- a/lib/etcconf/hosts.go
+++ b/lib/etcconf/hosts.go
@@ -17,7 +17,6 @@ package etcconf
 import (
 	"fmt"
 	"net"
-	"os"
 	"strings"
 	"sync"
 
@@ -41,7 +40,6 @@ type Hosts interface {
 	RemoveAll()
 
 	HostIP(hostname string) net.IP
-	GetPath() string
 }
 
 type hosts struct {
@@ -71,7 +69,7 @@ func (w *hostsWalker) Next() string {
 
 func NewHosts(path string) Hosts {
 	if path == "" {
-		path = hostsPath
+		path = HostsPath
 	}
 
 	return &hosts{
@@ -117,7 +115,7 @@ func (h *hosts) Load() error {
 	return nil
 }
 
-func (h *hosts) Save(filePath string) error {
+func (h *hosts) Save() error {
 	h.Lock()
 	defer h.Unlock()
 
@@ -142,13 +140,7 @@ func (h *hosts) Save(filePath string) error {
 		}
 	}
 
-	if err := save(filePath, &hostsWalker{entries: entries}); err != nil {
-		return err
-	}
-
-	// make sure the file is readable
-	// #nosec: Expect file permissions to be 0600 or less
-	if err := os.Chmod(h.path, 0644); err != nil {
+	if err := save(h.path, &hostsWalker{entries: entries}); err != nil {
 		return err
 	}
 
@@ -191,6 +183,6 @@ func (h *hosts) HostIP(hostname string) net.IP {
 	return h.hosts[hostname]
 }
 
-func (h *hosts) GetPath() string {
+func (h *hosts) Path() string {
 	return h.path
 }
--- a/lib/etcconf/hosts.go
+++ b/lib/etcconf/hosts.go
@@ -20,9 +20,12 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"syscall"
 
 	log "github.com/Sirupsen/logrus"
 )
+
+const ext4FileSystemType = "ext4"
 
 type hostEntry struct {
 	IP        net.IP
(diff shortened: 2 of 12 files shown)


lib/etcconf/hosts.go
```diff
@@ -41,6 +41,7 @@ type Hosts interface {
 	RemoveAll()
 
 	HostIP(hostname string) net.IP
+	GetPath() string
```
hmahmood: Should be just 
```go 
Path()
```

lib/etcconf/etcconf.go
```diff
@@ -36,7 +36,7 @@ type EntryWalker interface {
 
 type Conf interface {
 	Load() error
-	Save() error
+	Save(string) error
```
hmahmood:
```go
Save(filepath string)
```

lib/etcconf/resolvconf.go
```diff
@@ -42,6 +42,7 @@ type ResolvConf interface {
 	Timeout() time.Duration
 	SetAttempts(uint)
 	SetTimeout(time.Duration)
+	GetPath() string
```
hmahmood: Don't need this here since it is already in `Conf`

lib/etcconf/resolvconf.go
```diff
@@ -162,7 +163,7 @@ func (r *resolvConf) Save() error {
 
 	walker := &resolvConfWalker{lines: r.lines()}
 	log.Debugf("%+v", walker)
-	if err := save(r.path, walker); err != nil {
+	if err := save(filePath, walker); err != nil {
```
hmahmood: You should use `r.path` if `filePath == ""`

lib/tether/ops_linux.go
```diff
@@ -183,7 +191,10 @@ func (t *BaseOperations) SetHostname(hostname string, aliases ...string) error {
 	for _, a := range append(aliases, hostname) {
 		Sys.Hosts.SetHost(a, lo4)
 	}
-	if err = Sys.Hosts.Save(); err != nil {
+	if err = Sys.Hosts.Save(hostsPathBindSrc); err != nil {
```
hmahmood: So instead of this, you could just set `Sys.Hosts` to `etcconf.NewHosts(hostsPathBindSrc)` in an `init()` function in `/lib/tether/tether.go`. Same goes for `Sys.ResolvConf`. That way you do not need to modify the `Conf` interface.

lib/tether/ops_linux.go
```diff
@@ -492,7 +506,10 @@ func (t *BaseOperations) updateNameservers(endpoint *NetworkEndpoint) error {
 		log.Infof("Added nameserver: %s", gw.IP)
 	}
 
-	if err := Sys.ResolvConf.Save(); err != nil {
+	if err := Sys.ResolvConf.Save(resolvConfPathBindSrc); err != nil {
```
hmahmood: You have these two calls everywhere; why don't you combine them into one helper function?

lib/tether/tether.go
```diff
@@ -54,8 +54,24 @@ const (
 	bindDir = "/.tether/.bind"
 )
 
-var Sys = system.New()
-var once sync.Once
+var (
+	Sys = system.New()
+
+	fileForMinOS = map[string]os.FileMode{
```
hmahmood: `file -> files`

lib/tether/tether.go
```diff
@@ -54,8 +54,24 @@ const (
...
+		"/etc/hostname":            0644,
+		"/etc/hosts":               0644,
+		"/etc/resolv.conf":         0644,
+		"/.tether/etc/hostname":    0644,
+		"/.tether/etc/hosts":       0644,
+		"/.tether/etc/resolv.conf": 0644,
+	}
+	dirForMinOS = map[string]os.FileMode{
```
hmahmood: `dir -> dirs`

lib/tether/tether.go
```diff
@@ -118,6 +134,10 @@ func (t *tether) lenChildPid() int {
 func (t *tether) setup() error {
 	defer trace.End(trace.Begin("main tether setup"))
 
+	if err := createBindSrcTgt(); err != nil {
```
hmahmood: `Tgt -> Target`

lib/tether/tether.go
```diff
@@ -893,3 +913,27 @@ func killHelper(session *SessionConfig) error {
...
+// Create necessary folders/files as the src/target for bind mount.
+// See https://github.com/vmware/vic/issues/489
+func createBindSrcTgt() error {
+	for dirPath, dmode := range dirForMinOS {
+		if err := os.MkdirAll(dirPath, dmode); err != nil {
+			return fmt.Errorf("failed to create directory %s: %s", dirPath, err)
+		}
+	}
+
+	// The directory has to exist before creating the new file
```
hmahmood: Why don't we ensure that is the case by creating the file's parent directory here, and not having a separate map for the dirs?

lib/tether/ops_linux.go
```diff
@@ -38,20 +38,33 @@ import (
...
+	"github.com/vmware/vic/lib/etcconf"
 	"github.com/vmware/vic/pkg/ip"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vmw-guestinfo/rpcout"
 )
 
 var (
-	hostnameFile = "/etc/hostname"
-	byLabelDir   = "/dev/disk/by-label"
+	hostnameFile          = "/etc/hostname"
```
hmahmood: These should be consts.

lib/etcconf/hosts.go
```diff
@@ -145,12 +144,6 @@ func (h *hosts) Save() error {
 		return err
 	}
 
-	// make sure the file is readable
```
hmahmood: Why did you take this out?

chengwang86: @hmahmood This is taken care of in bindMountAndSave()

hmahmood: I think you should leave this here, since this code should run independent of bind-mount. You can keep the chmod after bind-mount elsewhere.

lib/tether/ops_linux.go
```diff
@@ -872,6 +891,13 @@ func (t *BaseOperations) Fork() error {
 }
 
 func (t *BaseOperations) Setup(config Config) error {
+	if err := createBindSrcTarget(filesForMinOSLinux); err != nil {
+		return err
+	}
+
+	Sys.Hosts = etcconf.NewHosts(hostsPathBindSrc)
```
hmahmood: `Setup` is called per tether instance, while Sys is package global. You should put this in an init() function.

lib/tether/tether.go
```diff
@@ -898,3 +898,28 @@ func killHelper(session *SessionConfig) error {
 
 	return nil
 }
+
+// Create necessary folders/files as the src/target for bind mount.
+// See https://github.com/vmware/vic/issues/489
+func createBindSrcTarget(files map[string]os.FileMode) error {
```
hickeng: suggest this goes into `tether_linux.go:Setup` - tether.go is supposed to be the pure tether logic without OS capability assumptions. Setup is the function provided for ensuring a given OS type has basic setup performed.

lib/tether/ops_linux.go
```diff
@@ -986,3 +1014,35 @@ func readDir(dir string) ([]os.FileInfo, error) {
...
+func bindMount(src, target string) error {
+	// no need to return if unmount fails; it's possible that the target is not mounted previously
+	log.Infof("unmounting %s", target)
+	if err := Sys.Syscall.Unmount(target, syscall.MNT_DETACH); err != nil {
+		log.Errorf("failed to unmount %s: %s", target, err)
+	}
+
+	// bind mount src to target
+	log.Infof("bind-mounting %s on %s", src, target)
+	if err := Sys.Syscall.Mount(src, target, ext4FileSystemType, syscall.MS_BIND, ""); err != nil {
```
hickeng: I _think_ that MS_BIND ignores fstype and you coud be less presumptive here:
https://github.com/mindreframer/golang-devops-stuff/blob/master/src/github.com/dotcloud/docker/vendor/src/github.com/docker/libcontainer/mount/init.go#L40

mhagen-vmware:
actually Bindmount looks like it can potentially just be tested with local mounts as well, no need for vcsim.

Issue #2972, opened 2016-11-01
Sync time with host via toolbox


fdawg4l:
We need a way to sync time with the host and keep it in sync in the appliance with the ESX host.  We were using ntp (via the `timesync` daemon [which helpfully `systemd` decided we required]) until [this](https://github.com/vmware/vic/pull/2971) change.  The `timesync` daemon would slam the clock leading to sudden skew in the logs and weirdness with respect to deadline based timeouts (which use wallclock time).

Disabling `timesync` was the first fix, the next fix is to keep the vch in sync with the host over time.  Apparently the `open-vm-tools` toolbox uses a timesync message to keep this in sync via a background task.  We need to implement this piece in our toolbox to do the same.  If we don't the clock may continue to drift and eventually logs between the VCH and the ESX host won't line up (this is already a problem irrespective of the `timesync` service change).

fdawg4l:
See #2340 for details on clock skew affecting deadline based timeouts (in the `context` package).

fdawg4l:
Assigning to myself for now, but no reason this can't be rebalanced.

fdawg4l:
cc @dougm, @hickeng, and @caglar10ur.

hickeng:
@jzt This is the work I mentioned in the sprint demo needed for the logs to function as expected over time.
This work would almost certainly fall to Doug.

@dougm Is there someone you think would be good to train on the toolbox systems (not just the current code but how to determine implementation path for new function) so we don't have to always assign these to you?

dougm:
@hickeng yes, this would be a good one start with as we already have the protocol handling in place for the Set_Option RPC, iirc that's where the time sync request comes in.  I could pair up with @jzt or whoever is interested.

hickeng:
@dougm @jzt Sounds good - please consider this as high priority once the cp work is complete.

jzt:
@dougm @hickeng sounds good, thanks.

PR #5524, opened 2017-06-19
Log worker and server time in order to detect drift in tests

mhagen-vmware:


hickeng:
I agree with logging to allow detection of drift, however we should also look at preventing it from being an issue:

There's #2972 for the VCH side.
If you're seeing drift in the workers then I suggest looking into why host timesync isn't taking effect.
If this is for drift between hosts then we should ensure they're all configured with aggressive ntp settings.

mhagen-vmware:
Agreed, this is just the first step.  All workers and hosts should be on NTP at this point, so I need to see which is drifting and then take corrective action as to why it is drifting.

PR #5527, opened 2017-06-19
Improve build speed and pin req versions with pre-installed reqs

### LINKED ISSUE #5161: Deploy local yum repo
User statement:  As an engineer on VIC, I would like a build process that is fast and consistent. Local repos will help with both of those goals.

Details:
https://www.digitalocean.com/community/tutorials/how-to-set-up-and-use-yum-repositories-on-a-centos-6-vps#custom-yum-repository

Acceptance Criteria:
1. Builds are done using local repos.

mhagen-vmware:
This appears to save us about 1.5 minutes every build as well as should hopefully prevent random build breaks during release cycles when we get dependency updates.

fixes #5161 

I am calling this a fix for 5161, because if we agree on this approach, then we don't need to roll out any local repos.  This may break local builds depending on how you are doing local builds:  @caglar10ur @dougm @hickeng Please let me know if I can tweak this at all, or provide some conditionals so as to resolve any local build breakage for you.


isos/base.sh
```diff
@@ -56,23 +56,22 @@ fi
...
-PKGDIR=$(mktemp -d)
+PKGDIR="/build"
 
 # initialize the bundle
 initialize_bundle $PKGDIR
 
 # base filesystem setup
 mkdir -p $(rootfs_dir $PKGDIR)/{etc/yum,etc/yum.repos.d}
 ln -s /lib $(rootfs_dir $PKGDIR)/lib64
-cp $DIR/base/*.repo $(rootfs_dir $PKGDIR)/etc/yum.repos.d/
```
hickeng: I'd be surprised if this didn't break things for local build where it's not being run from the Dockerfile

infra/integration-image/Dockerfile
```diff
@@ -34,11 +34,23 @@ RUN curl -sL https://deb.nodesource.com/setup_7.x | bash - && \
...
+    xorriso && \
 	# Cleanup
     apt-get autoremove -y && \
     rm -rf /var/lib/apt/lists/*
 
+COPY *.repo /etc/yum/repos.d/
+RUN yum-config-manager --enable photon photon-updates
+
+RUN mkdir -p /build/rootfs
+RUN yum --installroot /build/rootfs install coreutils filesystem linux-esx e2fsprogs gzip haveged iproute2 iptables iputils logrotate net-tools openssh photon-release procps-ng sudo systemd tdnf vim systemd -y --nogpgcheck
```
hickeng: I think this will inflate the size of the bootstrap ISO significantly as we're installing many unnecessary packages into the base archive.
Why not look at using the caching mechanism that's already written into the scripts for local use?

We had enabled the drone cache option in 0.4 but over time it seemed to inflate and took a long time to unpack so it was disabled. However we can try it again in 0.5 - combined with the `make clean` call (not `make distclean` which would delete the staging builds) this would have a similar effect, maybe better as it wouldn't cause rebuild of the application-stanging.tgz and bootstrap-staging.tgz archives each time.

mhagen-vmware: Speed is one thing, but there are other important reasons such as pinning the version and of course this (which is now causing all builds to master to fail this morning):
```
https://vmware.bintray.com/photon_release_1.0_x86_64/x86_64/gmp-6.0.0a-3.ph1.x86_64.rpm: [Errno 14] PYCURL ERROR 52 - "Empty reply from server"
Trying other mirror.
Error Downloading Packages:
  gmp-6.0.0a-3.ph1.x86_64: failure: x86_64/gmp-6.0.0a-3.ph1.x86_64.rpm from photon: [Errno 256] No more mirrors to try.
```

mhagen-vmware: I can fix the size of bootstrap and make sure local builds work as well with just a couple of conditionals.  If I do that, are you generally ok with this approach?

caglar10ur: I'm not sure I understand why we are not starting to host our own photon mirror and trying to cache packages on build time? Is bintray not providing a way to mirror? If so it shouldn't be hard to parse and download the content of https://vmware.bintray.com/photon_release_1.0_x86_64/x86_64/ via some script.

Also I don't use docker to build so I believe this pr will end up creating more confusion as I'll test a different set of versions while CI uses another set of packages.

mhagen-vmware: ah, good point @caglar10ur I am going to close this then for now and re-work it so that we don't have that mixing of versions.

caglar10ur: Other alternative would be declaring an official way of building binaries, in this case asking everyone to use docker build for example, if we are going to do this pinning.

Issue #4922, opened 2017-04-26
[portlayer] reconfigure call scales linearly


caglar10ur:
Following is the graph that shows what happens if you create a container and then exec a short living session 100 times - sequentially 

![screenshot 2017-04-26 14 22 18](https://cloud.githubusercontent.com/assets/205498/25457658/e142ac4a-2a8b-11e7-81ba-213a085e6b96.png)

Every new exec call adds a new set of keys to the guestinfo. That increases the reconfigure time linearly. 

**Acceptance Criteria:**

reconfigure operation should take a constant time no matter how many exec'ed tasks we have

hickeng:
The working assumption is that reconfigure is taking longer because of the increasing size of the ConfigSpec - however it's possible that this is actually a vSphere behaviour with volatile keys that we're simply exacerbating with the repeated specification of existing exec session state.

hickeng:
Addressed in #5532 to a large degree. Scaling still occurs, but now much less problematic: https://github.com/vmware/vic/pull/5532#issuecomment-311476394

Issue #3728, opened 2017-01-25
Container VM process exits non-zero but is not caught by the portlayer


jzt:
Seen in a CI test:

A `/bin/top` container is created without `-t`, causing `top` to exit with status `1` because no `TERM` is set. This causes the container to shutdown, but the non-zero exit status is not bubbled up, and a CI test passes when it should have failed.

Excerpt from tether.debug:
```
time="2017-01-18T23:58:18Z" level=info msg="Launching process for session 933e252ce6fd063388d571849dc52f044a892c67becf175dfbcb63b6c1a5da52" 
time="2017-01-18T23:58:18Z" level=debug msg="[BEGIN] [github.com/vmware/vic/lib/tether.(*tether).launch:502] launching session 933e252ce6fd063388d571849dc52f044a892c67becf175dfbcb63b6c1a5da52" 
time="2017-01-18T23:58:18Z" level=debug msg="[BEGIN] [main.(*attachServerSSH).run:285] main attach server loop" 
time="2017-01-18T23:58:18Z" level=info msg="Trying to establish a connection" 
time="2017-01-18T23:58:18Z" level=info msg="opening ttyS0 for backchannel" 
time="2017-01-18T23:58:18Z" level=info msg="setting terminal to raw mode" 
time="2017-01-18T23:58:18Z" level=debug msg="s = &terminal.State{termios:syscall.Termios{Iflag:0x500, Oflag:0x5, Cflag:0xcbd, Lflag:0x8a3b, Line:0x0, Cc:[32]uint8{0x3, 0x1c, 0x7f, 0x15, 0x4, 0x0, 0x1, 0x0, 0x11, 0x13, 0x1a, 0x0, 0x12, 0xf, 0x17, 0x16, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0}, Pad_cgo_0:[3]uint8{0x0, 0x0, 0x0}, Ispeed:0x0, Ospeed:0x0}}" 
time="2017-01-18T23:58:18Z" level=info msg="creating raw connection from ttyS0 (fd=7)" 
time="2017-01-18T23:58:18Z" level=debug msg="[BEGIN] [main.backchannel:231] establish tether backchannel" 
time="2017-01-18T23:58:18Z" level=debug msg="HandshakeServer: Waiting for incoming syn request..." 
[...]

PR #5532, opened 2017-06-20
Reduce incidence of #5445 [full ci]

### LINKED ISSUE #5445: State race when updating container config
From: #4872

We have observed one incidence in the logs of a container having its state damaged.
From the logs we can see multiple concurrent operations:
1. an in-guest shutdown resulting from explicit invocation of `exit` in the shell
2. a stop operation
3. a restart operation

My hypothesis is that the powerOn from the restart (3) is occuring just after the powerOff operation from (2) has succeeded, but before the container configuration is updated with the revised StoppedTime.
This results in us using a full config as that's based on the target power state, not the non-persistent portion suited to a powered on cVM. This hypothesis was generated by code inspection with input from the logs.

@cgtexmex and I have been unable to recreate this failure remotely reliably with released codebase, so I have injected code to manipulate the timings. This has resulted in the expected `SetVolatileEntry` lines in the container `vmware.log`.

In specific, this adds a sleep of 5s before https://github.com/vmware/vic/blob/master/lib/portlayer/exec/commit.go#L119 and races:
1. docker stop -t 1 _id_
2. docker start _id_

hickeng update: this can be mostly addressed in VIC code, but not completely without bug1898149 - I will let this issue get closed by the work around, and keep the platform issue open.

### LINKED ISSUE #5501: Stop wait time is not applied to startGuestProgram path
The stop wait time (10s by default or supplied by `-t`) is not honoured by the startGuestProgram path for the soft shutdown. We need to add the wait time as a timeout to the startGuestProgram path in `base.go:shutdown()`.

This will be hard to test without vcsim as it's primary needed to handle failure paths in the vSphere or tether parts of the call. Estimate is for fix without test.

hickeng:
This adds the following:
* only set changed extraconfig keys (doesn't support deletion of keys at this time)
* handle net.Operror returns from interaction path
* honor context timeout in startGuestProgram call

The PR currently includes #5184 as those are needed to properly exercise the code paths at fault and obtain useful diagnostics.
Helps reduce the impact of #4922 but does not eliminate the problem.

This does not completely fix #5445 however it does everything possible within the VIC code, so closing it with this PR. The platform issue will remain open.

Fixes #5501 #5610 #5445 
Towards #4922

### DIFF
--- a/lib/apiservers/engine/backends/convert/state.go
+++ b/lib/apiservers/engine/backends/convert/state.go
@@ -78,7 +78,7 @@ func State(info *models.ContainerInfo) *types.ContainerState {
 				// most likely the process was killed via the cli
 				// or received a sigkill
 				exitCode = 137
-			} else if info.ProcessConfig.Status == "" && exitCode == 0 {
+			} else if info.ProcessConfig.Status == "" && exitCode == -1 {
 				// the process was stopped via the cli
 				// or received a sigterm
 				exitCode = 143
--- a/tests/resources/VCH-Util.robot
+++ b/tests/resources/VCH-Util.robot
@@ -451,7 +451,7 @@ Check Original Version
     @{vers}=  Split String  ${output}
     ${rc}  ${output}=  Run And Return Rc And Output  bin/vic-machine-linux inspect --name=%{VCH-NAME} --target=%{TEST_URL}%{TEST_DATACENTER} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --password=%{TEST_PASSWORD} --compute-resource=%{TEST_RESOURCE}
     Should Contain  ${output}  Completed successfully
-    Should Contain  ${output}  @{vers}[2]
+    Should Contain  ${output}  %{INITIAL-VERSION}
     Should Be Equal As Integers  ${rc}  0
     Log  ${output}
     Get Docker Params  ${output}  ${true}
--- a/tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
+++ b/tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
@@ -41,7 +41,15 @@ Run Docker Checks
     Should Be Equal As Integers  ${rc}  0
     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} ps -a
     Should Be Equal As Integers  ${rc}  0
-    Should Contain  ${output}  Exited (143)
+
+
+    ${status}=  Get State Of Github Issue  5653
+    Run Keyword If  '${status}' == 'closed'  Should Contain  ${output}  Exited (143)
+    Run Keyword If  '${status}' == 'closed'  Fail  Exit code check below needs to be updated now that Issue #5653 has been resolved
+    # Disabling the precise check for error code until https://github.com/vmware/vic/issues/5653 is fixed - we can get rid of the
+    # conditional around the exit code check once the issue is closed
+    #Should Contain  ${output}  Exited (143)
+
     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} start vch-restart-test1
     Should Be Equal As Integers  ${rc}  0
     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} ps -a
@@ -142,6 +150,9 @@ Upgrade VCH with unreasonably short timeout and automatic rollback after failure
     Run Keyword If  '%{HOST_TYPE}' == 'VC'  Should Not Contain  ${output}  upgrade
     ${rc}  ${output}=  Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Run And Return Rc And Output  govc snapshot.tree -vm=%{VCH-NAME}
     Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Should Not Contain  ${output}  upgrade
+
+    # confirm that the rollback took effect
+    Check Original Version
 
 Upgrade VCH
     Create Docker Containers
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -269,7 +269,7 @@ func (handler *ContainersHandlersImpl) GetContainerInfoHandler(params containers
 
 	// Refresh to get up to date network info
 	container.Refresh(context.Background())
-	containerInfo := convertContainerToContainerInfo(container.Info())
+	containerInfo := convertContainerToContainerInfo(container)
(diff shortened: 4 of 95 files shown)


hickeng:
The current problem with the docker stop test appears to be because `backends/convert/state.go` isn't performing the expected translation. I've observed uncertainty in this path in @caglar10ur 's local build without this change, switching between `137` and `143` as a return code.
I believe that #3728 is also an instance of this. Example of reported exit code changing over time:
```
$ docker ps -a
aa248f0cc50c        busybox             "/bin/top"          2 minutes ago       Exited (143) 4 seconds ago                       stoic_hugle
$ docker inspect aa | jq '.[]|.["State"]|.["ExitCode"]'
-1
$ docker inspect aa
[
    {
        "Id": "aa248f0cc50ca651f19fdc327383d66ee51f961bb90b220e7458081278d0030f",
        "Created": "2017-06-22T22:56:57Z",
        "Path": "/bin/top",
        "Args": [],
        "State": {
            "Status": "exited",
            "Running": false,
            "Paused": false,
            "Restarting": false,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 0,
            "ExitCode": -1,
            "Error": "",
            "StartedAt": "2017-06-22T22:56:58Z",
            "FinishedAt": "2017-06-22T22:58:25Z"
        },
<snip>
$ docker ps -a
aa248f0cc50c        busybox             "/bin/top"          17 minutes ago      Exited (-1) 15 minutes ago                       stoic_hugle
```

caglar10ur:
here are the timing of running echo on master, each line represents 10 exec sessions (support bundle http://10ur.org/esx-localhost.localdomain-2017-06-23--00.53-71885.tgz)

1m7.450s
2m5.626s
3m3.659s
4m14.441s
5m21.224s
7m25.063s

will build this branch and repeat

cgtexmex:
@hickeng the original intent of the code conversion was to replicate docker's status codes as follows:

- If a container finishes on it's own (i.e. `docker run busybox date`) then present the exit code returned from the container.  This would allow for programs to throw specific success / error codes.

- If a container is stopped (i.e. `docker stop container-name`) the exit code should be set to 143

- if a container is killed (i.e. `docker kill container-name`) the exit code should be set to 137

This conversion only happens in the persona based on the data available in extraConfig, etc...so the transformation would take place during `ps` and `inspect` operations.  This means that if `docker kill` was executed the code of 137 would NOT be persisted to the extraConfig.  There have been some changes to the code since it's inception so potentially bugs were introduced.

I can take a look at it later today if you'd like.

hickeng:
Need to add filtering of non-persistent keys when powered off - we're picking up the non-persistent exec keys and setting them in the powered off VM currently.

caglar10ur:
OK, will switch to 5565 and come back to this when you implement the filtering

hickeng:
@caglar10ur Done - think I've also fixed the issue with the stop state as well.

caglar10ur:
OK, timing of 60 runs on master
```
1.39
1.76
2.13
1.87
2.30
2.84
3.09
2.75
2.58
2.65
2.88
2.92
3.22
4.34
3.36
3.40
3.73
4.01
4.12
4.62
4.82
5.02
5.32
5.65
6.18
6.15
7.16
6.53
7.22
7.44
7.83
8.08
8.50
8.96
9.13
9.64
10.17
10.41
10.59
10.95
11.53
12.17
12.24
13.57
13.24
14.77
14.97
14.68
15.51
15.65
16.36
17.13
17.77
18.31
18.67
18.97
20.70
20.44
20.82
21.88
```

currently building your branch will update here

caglar10ur:
new graph, master vs branch on a real hw without debug option

screenshot 2017-06-27 13 30 29

caglar10ur:
[support bundle for the last run](10ur.orgesx-office2-sfo2-dhcp327.eng.vmware.com-2017-06-27--20.23-408156.tgz)

caglar10ur:
[container-logs.zip](https://github.com/vmware/vic/files/1106692/container-logs.zip)

hickeng:
@cgtexmex This is now ready for review. Updates to the exit code reflect us now _not_ overwriting exit code state in the 143 stop path.

lib/apiservers/engine/backends/convert/state.go
```diff
@@ -78,7 +78,7 @@ func State(info *models.ContainerInfo) *types.ContainerState {
 				// most likely the process was killed via the cli
 				// or received a sigkill
 				exitCode = 137
-			} else if info.ProcessConfig.Status == "" && exitCode == 0 {
+			} else if info.ProcessConfig.Status == "" && exitCode == -1 {
```
hmahmood: Is this change intentional?

hickeng:
@mhagen-vmware ready for CI review

tests/test-cases/Group11-Upgrade/11-01-Upgrade.robot
```diff
@@ -41,7 +41,15 @@ Run Docker Checks
     Should Be Equal As Integers  ${rc}  0
     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} ps -a
     Should Be Equal As Integers  ${rc}  0
-    Should Contain  ${output}  Exited (143)
+
+
+    ${status}=  Get State Of Github Issue  5653
+    Run Keyword If  '${status}' == 'closed'  Fail  Exit code check below needs to be updated now that Issue #5653 has been resolved
+    Run Keyword If  '${status}' == 'closed'  Should Contain  ${output}  Exited (143)
```
mhagen-vmware: If you want this to actually trigger you need to switch these two. Otherwise the test exits on first line.

tests/resources/VCH-Util.robot
```diff
@@ -451,7 +451,7 @@ Check Original Version
     @{vers}=  Split String  ${output}
     ${rc}  ${output}=  Run And Return Rc And Output  bin/vic-machine-linux inspect --name=%{VCH-NAME} --target=%{TEST_URL}%{TEST_DATACENTER} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --password=%{TEST_PASSWORD} --compute-resource=%{TEST_RESOURCE}
     Should Contain  ${output}  Completed successfully
-    Should Contain  ${output}  @{vers}[2]
+    Should Contain  ${output}  %{INITIAL-VERSION}
     Should Be Equal As Integers  ${rc}  0
     Log  ${output}
     Get Docker Params  ${output}  ${true}
```
mhagen-vmware: I very much dislike the structure of both of these version keywords, but I will take care of that - you don't need to:
https://github.com/vmware/vic/issues/5670

lib/apiservers/engine/backends/convert/state.go
hickeng: yup - the only reason we were seeing `0` in master is because we're overwriting the exit code from the container in this path.

PR #5544, opened 2017-06-20
add scratch version number to the metadata of scratch disk

chengwang86:
This is part of #489 

- If the scratch version number is less than a certain number, the scratch disk wouldn't have the minimum os structure, and then PL needs to use a file mask for `diff`;
- Otherwise, the scratch is already populated with min os structure, so there is no need for the file mask.

### DIFF
--- a/lib/portlayer/storage/vsphere/image.go
+++ b/lib/portlayer/storage/vsphere/image.go
@@ -64,11 +64,12 @@ var (
 )
 
 const (
-	StorageImageDir  = "images"
-	defaultDiskLabel = "containerfs"
-	defaultDiskSize  = 8388608
-	metaDataDir      = "imageMetadata"
-	manifest         = "manifest"
+	StorageImageDir     = "images"
+	defaultDiskLabel    = "containerfs"
+	defaultDiskSize     = 8388608
+	metaDataDir         = "imageMetadata"
+	manifest            = "manifest"
+	scratchMetaDataFile = "metaData"
 )
 
 type ImageStore struct {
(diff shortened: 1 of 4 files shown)


lib/portlayer/storage/vsphere/image.go
```diff
@@ -395,7 +397,12 @@ func (v *ImageStore) scratch(op trace.Operation, storeName string) error {
 
 	// Write the metadata to the datastore
 	metaDataDir := v.imageMetadataDirPath(storeName, portlayer.Scratch.ID)
-	if err := writeMetadata(op, v.ds, metaDataDir, nil); err != nil {
+	scratchMetaData := &metadata.ScratchMetaData{
+		ScratchVersion: feature.ScratchBaseStructureVersion,
+	}
+	meta := make(map[string][]byte)
+	meta["metaData"] = []byte(fmt.Sprintf("%+v", scratchMetaData))
```
hickeng: Suggest using a constant for this key, a meaningful key as everything stored in this fashion is metadata, and input closer to the namespaced style used with WriteImage, e.g
```
portlayer.storage.scratch
```

chengwang86: @hickeng The key is used as the name of the metaData file.  So by setting the key to `metaData`, I generate a file named `metaData` with the content of `{scratch_version: 9}` where `9` is the value of `feature.ScratchBaseStructureVersion`. Are you talking about setting the content of the metaData file to `{portlayer.storage.scratch: 9}`?

PR #5556, opened 2017-06-21
Host a local repo of our yum dependencies

### LINKED ISSUE #5161: Deploy local yum repo

mhagen-vmware:
I have skipped CI because I am still struggling to host this within our CI on a VCH, but I have tested it locally and it works great.  Build times are down from about 13 minutes to 6.5 minutes.  Not going to merge this until I can get the CI VCH up and running and can test this within CI, but this is ready for review.

fixes #5161


isos/base/photon-updates-local.repo
```diff
@@ -0,0 +1,5 @@
+[photon-updates]
+name=VMware Photon Linux 1.0(x86_64)
+baseurl=ftp://192.168.31.16/photon-updates
```
caglar10ur: where is that IP coming from? Could it be a different for different runs?

mhagen-vmware: This is the static IP address that I have reserved in the CI, it won't change.

infra/integration-image/LocalRepo
```diff
@@ -0,0 +1,26 @@
...
+RUN mkdir /photon && mkdir /photon-updates
+
+ENV EXCLUDE_LIST "index.html*,openjdk*,openjre*,perl*,postgresql*,python-*,\
+python3*,ruby*,subversion*,gnome*,NetworkManager*,cloud*,docker*,grub*,ktap*,\
+kubernetes*,linux-docs*,linux-sound*,linux-tools*,docbook*,httpd*,go-*,jna*,\
+linux-debuginfo*,linux-dev*,linux-docs*,linux-drivers*,linux-oprofile*,linux-sound*,\
+linux-tools*,linux-esx-debuginfo*,linux-esx-devel*,linux-esx-docs*,nginx*,sysdig*"
+
+RUN wget -e robots=off -r -nH -nd -np -R $EXCLUDE_LIST https://vmware.bintray.com/photon_release_1.0_x86_64/x86_64/ -P /photon
+RUN wget -e robots=off -r -nH -nd -np -R $EXCLUDE_LIST https://dl.bintray.com/vmware/photon_updates_1.0_x86_64/x86_64/ -P /photon-updates
```
mhagen-vmware: Yea, I had that and would prefer that as well, unfortunately that ends up being over 12GB image. Which is incredibly unwieldy (both upload/download and filling up our harbor instance).  It also takes almost 2 hrs to build the image if we don't exclude.

mhagen-vmware: And I am a bit concerned about hammering their repo with a crawler that pulls from them for 2hrs straight.

caglar10ur: That 12GB is a one time cost though right? After that we can use rsync or something like that to get only changed files

mhagen-vmware:
Me got angry with FTP and now it is HTTP :)

frapposelli:
@mhagen-vmware why not hosting the rpms in a host directory/docker volume and having a cron that updates it? so you can feed the rpm list independently of the container.

mhagen-vmware:
@frapposelli Yea, I was considering that. It would be a nice addition, but according to the original issue #5161 we actually want to pin the version per release.  And that I think would put even more stress on the original repo as well if multiple people start using this and hammering the repo with regularity.  I think the biggest goal though is to not update our dep at random times though, so that we don't get a build break because they change something major in photon in the middle or even worse on the last day of our release cycle.

PR #5559, opened 2017-06-22
Fix mount information in docker inspect

### LINKED ISSUE #5557: Mount data not present in container inspect data
As a user of VIC, I expect docker inspect to return mount data for the container.

Acceptance Criteria:

- [ ] updated docker inspect robot scripts that check the mount data is valid

---
We need to derive this data during docker inspect.  We also need to ensure this data is cached in the persona's container cache.

sflxn:
Fixed mount information in docker inspect after container create.
There is still a problem with VCH restart that will be addressed
in 5558.

Resolves #5557


matthewavery:
lgtm

PR #5572, opened 2017-06-22
Move toolbox to vmware/govmomi repo

### LINKED ISSUE #5548: Move vsphere/toolbox package to govmomi
As a developer, I want to use pkg/vsphere/toolbox in projects outside of VIC, without vendoring the vmware/vic repo.

@vmware/kov-maintainers are starting to use the toolbox instead of open-vm-tools, let's move to a common repo rather than diverge as we started to with the simulator.

While toolbox could standalone in its own repo, proposing vmware/govmomi as the new home for now:

- vcsim can implement GuestOperationsManager that drives the toolbox without requiring a real VC/ESX endpoint or use of the guest backdoor RPC channel.

- we use the govc guest.* commands to for end-to-end toolbox testing

- govmomi+govc tests currently depend on an old ttylinux + custom (outdated) build of open-vm-tools.  These tests can be changed to vcsim and/or any lightweight linux distro with a simple deploy of toolbox/main.go (similar to the existing toolbox-test.sh)
 
- projects consuming toolbox are also consuming govmomi, so its 1 less repo to manage, but we can still move to its own repo down the road if desired.

Acceptance Criteria:

vmware/vic/pkg/vsphere/toolbox moves to vmware/govmomi/toolbox

vmware/vic/cmd/toolbox moves to vmware/govmomi/toolbox/toolbox

VIC and kov are updated to use the simulator from vendored govmomi.

See also: #5392

dougm:
Add dougm vendor/ approvals

Closes #5548


dougm:
@gigawhitlocks I removed pullapprove sections that no longer apply.  And added myself to approve vendor/ updates @mdubya66

PR #5575, opened 2017-06-23
Move iptables plugins to tmpfs

hickeng:
To keep all of the injected and runtime state on the /.tether `tmpfs` this
moves the iptables plugins onto the tmpfs and alters the invocation to
use chroot so that dlopen path resolution works correctly.

This will require a follow-on to removing `/usr/lib/iptables` from the
scratch - a adjustment from the recent #5521 merge.

`netfilter` package is moved out of `pkg` to `lib/tether` because it has
hardcoded assumptions that are contingent on tether code. We can
and should move this back once we've addressed that tight coupling.

Towards #489


isos/bootstrap/bootstrap
```diff
@@ -63,19 +63,17 @@ if mount -t ext4 /dev/disk/by-label/containerfs ${MOUNTPOINT}; then
...
-    mkdir ${MOUNTPOINT}/.tether/lib
-    mkdir ${MOUNTPOINT}/.tether/lib64
-    cp -a /sbin/*tables* ${MOUNTPOINT}/.tether/
+    ln -s lib64 ${MOUNTPOINT}/.tether/lib
+    mkdir -p ${MOUNTPOINT}/.tether/{lib64,usr/lib/iptables}
 
+    cp -a /sbin/*tables* ${MOUNTPOINT}/.tether/
     cp -a /lib/libm.* /lib/libm-* /lib/libgcc_s* /lib/libip*tc* /lib/libxtables* /lib/libdl* /lib/libc.so* /lib/libc-* ${MOUNTPOINT}/.tether/lib
     cp -a /lib64/ld-* ${MOUNTPOINT}/.tether/lib64
+    cp -r /usr/lib/iptables ${MOUNTPOINT}/.tether/usr/lib/
```
caglar10ur: this whole thing hurts my eyes :) I know currently there is no other way but that never stopped me ranting :P

hickeng: Unfortunately, as @fdawg4l and I both found the offical stance for using netfilter for ipchains is "Don't. Use the iptables binary"

hickeng:
Updated to use prophylactic rtld as per @fdawg4l out-of-band comment. Was functioning in testing without this, but we cannot rely on images to be problem free in this regard.

PR #5589, opened 2017-06-26
[specific ci=Group6-VIC-Machine] Add no-tls back in

gigawhitlocks:
Specific CI didn't run on my last PR for some reason (let's see if it actually runs the tests I requested this time) and I missed this, which I stripped out, and added back in, and stripped again, and apparently forgot to add back, as we waffled around on whether or not `--no-tls` should be an option in `configure`

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -247,12 +247,12 @@ func (c *Create) Flags() []cli.Flag {
 		})
 
 	tls := c.certs.CertFlags()
+
 	tls = append(tls, cli.BoolFlag{
-		Name:        "no-tls",
-		Value:       false,
-		Usage:       "Disable TLS entirely",
+		Name:        "no-tls, k",
+		Usage:       "Disable TLS support completely",
+		Destination: &c.noTLS,
 		Hidden:      true,
-		Destination: c.certs.NoTLS,
 	})
 
 	registries := []cli.Flag{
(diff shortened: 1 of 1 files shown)


cmd/vic-machine/create/create.go
```diff
@@ -247,6 +247,13 @@ func (c *Create) Flags() []cli.Flag {
 		})
 
 	tls := c.certs.CertFlags()
+	tls = append(tls, cli.BoolFlag{
+		Name:        "no-tls",
+		Value:       false,
+		Usage:       "Disable TLs entirely",
```
anchal-agrawal: s/TLs/TLS

gigawhitlocks: ffs thanks @anchal-agrawal

cmd/vic-machine/create/create.go
```diff
@@ -247,6 +247,13 @@ func (c *Create) Flags() []cli.Flag {
 		})
 
 	tls := c.certs.CertFlags()
+	tls = append(tls, cli.BoolFlag{
+		Name:        "no-tls",
```
anchal-agrawal: Looks like we had a short-flag `-k` for this too, should we keep it? https://github.com/vmware/vic/blob/b3f5b28d65511e5175073558d486bf397d83eb78/cmd/vic-machine/create/create.go#L301

PR #5595, opened 2017-06-26
[skip ci] Add a test case for lazy disk detach operation

### LINKED ISSUE #5565: portlayer: lazy disk detach operations
We have two orthogonal disk interfaces, one for IO interaction and another for lifecycle operations:
```go
type Archive interface {
   // Import data into a specific storage object
   Import(...)
   // Export data from a specific storage object
   Export(...)
```
The lifecycle operations are basically pure CRUD:
```go
type Store interface {
   // Create a storage object in the implementing store
   Create(...)
   // Inspect details of a specific storage object
   Read(...)
   // Modify core details of a specific storage object, e.g. max size,
   Update(...)
   // Delete a specific storage object
   Delete(...)
}
```
To support an efficient implementation of the existing `WriteImage` call using this interface separation we need to move to a deferred detach of disks from the endpointVM - this means that a detach call doesn't necessarily take immediate effect. The detach should take effect when:
1. a reconfigure is invoked for another operation - the easy opportunistic path - or 
2. a VM that references the disk has an event or operation performed on it - avoid locking a disk that may now be needed by a different VM - or
3. a given delay has lapsed - avoid a disk being held indefinitely and to support dual mode management

To support (1) in conjunction with a `pull` operation we need to confirm whether a single reconfigure operation can:
* detach disk A
* create and attach disk B where B is a child of A

If we can use a composite reconfigure operation for (1) then it will halve the number of vSphere calls for pulling of a given image.

Estimate is for initial check of composite reconfigure behaviour.

caglar10ur:
Fixes #5565


pkg/vsphere/disk/lazy_detach_test.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	fm := object.NewFileManager(client.Vim25())
+
+	// create a directory in the datastore
+	// eat the error because we dont care if it exists
+	fm.MakeDirectory(context.TODO(), imagestore.String(), nil, true)
+	// Nuke the image store
+	defer func() {
+		task, err := fm.DeleteDatastoreFile(context.TODO(), imagestore.String(), nil)
```
hickeng: This is potentially damaging given we're not asserting that the directory is freshly created.
This is not likely an issue right now given the way these tests are run, however it will collide with other in-flight tests if run as part of a build in the future. Suggest using mktmp or similar and asserting that there's no error

pkg/vsphere/disk/lazy_detach_test.go
```diff
@@ -0,0 +1,185 @@
...
+		if err != nil && err.Error() == "can't find the hosting vm" {
+			t.Skip("Skipping: test must be run in a VM")
```
hickeng: There is a GetSelf() call that we can make right at the beginning of the test and t.FailNow if we're not in an appropriate VM - this isn't the sort of error that should show up in cleanup.

pkg/vsphere/disk/lazy_detach_test.go
```diff
@@ -0,0 +1,185 @@
...
+		return
+	}
+	t.Logf("scratch created and attached")
+
+	// ref to scratch (needed for detach as initial spec's Key and UnitNumber was unset)
+	disk, err := findDiskByFilename(op, vdm.vm, scratch.Path)
+	if !assert.NoError(t, err) {
+		return
+	}
+
```
hickeng: can we add a filesystem & touch a file here.

pkg/vsphere/disk/lazy_detach_test.go
```diff
@@ -0,0 +1,185 @@
...
+		return
+	}
+	t.Logf("scratch detached, child created and attached")
+
+	// ref to child (needed for detach as initial spec's Key and UnitNumber was unset)
+	disk, err = findDiskByFilename(op, vdm.vm, child.Path)
+	if !assert.NoError(t, err) {
+		return
+	}
+
```
hickeng: check for the file created on scratch here - this already validates that the combined remove/add operation is permitted, but doesn't confirm actual inheritance.

pkg/vsphere/disk/lazy_detach_test.go
caglar10ur: imagestore path is constructed via  datastore.TestName("lazyReconfigure") which creates a random string and concatenates the result with the passed parameter

pkg/vsphere/disk/lazy_detach_test.go
caglar10ur: Done

pkg/vsphere/disk/lazy_detach_test.go
caglar10ur: done

pkg/vsphere/disk/lazy_detach_test.go
caglar10ur: done

caglar10ur:
@hickeng this PR needs you

caglar10ur:
@hickeng another ping

PR #5601, opened 2017-06-27
Parallelize integration tests

mhagen-vmware:
[skip unit]
[full ci]


tests/resources/VCH-Util.robot
```diff
@@ -272,13 +274,21 @@ Scrape Logs For the Password
...
+    Run Keyword If  ${keepVCH}  Cleanup Containers And Images
+    Return From Keyword If  ${keepVCH}  Keeping the existing VCH - %{VCH-NAME}
     Log To Console  Deleting the VCH appliance %{VCH-NAME}
     ${output}=  Run VIC Machine Delete Command
     Run Keyword And Ignore Error  Cleanup VCH Bridge Network  %{VCH-NAME}
+    Remove Environment Variable  VCH-NAME
     [Return]  ${output}
 
+Cleanup Containers And Images
+    ${out}=  Run  docker %{VCH-PARAMS} ps -aq | xargs -n1 docker %{VCH-PARAMS} rm -f
```
hickeng: you shouldn't need the `-n1` but I don't think it'll hurt

hickeng:
Can we have each test suite make a "Get VCH" call or similar. Flow would be:
* is there an existing VCH?
  * does the existing VCH pass basic functional check (e.g. docker info)?
    * if yes, use this VCH
* deploy VCH
  * is there a VCH deploy already in progress?
    * if so, wait for deploy to complete and retry from top
  * start new VCH deploy
    * wait for deploy to complete and retry from top

mdubya66:
LGTM!

mhagen-vmware:
Still seeing TLS handshake timeouts with @caglar10ur latest changes.

caglar10ur:
I think we are getting those cause the ESXi is getting slower and eventually timing out. Can we collect the host statistics during the run to see whether 1) This speculation is correct 2) The slowness is CPU or IO bound 3) we can do something about it?

mdharamadas1:
@mhagen-vmware This seems to have issue with cleaning up container VMs as observed in [this](https://ci.vcna.io/vmware/vic/14177) build running on builds-16.vmware.run. There were about 87 VMs left over.

caglar10ur:
```
Several failures occurred:

1) 'govc: 503 Service Unavailable (Failed to connect to endpoint: [N7Vmacore4Http16LocalServiceSpecE:0x31eb2938] _serverNamespace = /sdk _isRedirect = false _port = 8307)' does not contain 'OK'

2) 'govc: 503 Service Unavailable (Failed to connect to endpoint: [N7Vmacore4Http16LocalServiceSpecE:0x31eb2938] _serverNamespace = /sdk _isRedirect = false _port = 8307)' does not contain 'OK'
```

looks like someone restarted the host ESXi server while the tests are running

mhagen-vmware:
I restarted it afterwards, because the system was in a state that we continue to see.  In which everything appears up, but the system reports wrong password when trying to login or do anything via govc. A system reboot through OVH is needed and fixes it.

.drone.local.yml
```diff
@@ -11,7 +11,7 @@ pipeline:
     recursive: false
 
   vic-integration-test-on-pr:
-    image: ${TEST_BUILD_IMAGE=gcr.io/eminent-nation-87317/vic-integration-test:1.39}
+    image: ${TEST_BUILD_IMAGE=gcr.io/eminent-nation-87317/vic-integration-test:1.37}
```
hickeng: should be we reverting to `1.37` from `1.39`?

tests/integration-test.sh
```diff
@@ -22,24 +22,25 @@ dpkg -l > package.list
 
 
 buildinfo=$(drone build info vmware/vic $DRONE_BUILD_NUMBER)
+export GOVC_TLS_HANDSHAKE_TIMEOUT=30s
 
 if [[ $DRONE_BRANCH == "master" || $DRONE_BRANCH == "releases/"* ]] && [[ $DRONE_REPO == "vmware/vic" ]] && [[ $DRONE_BUILD_EVENT == "push" ]]; then
     echo "Running full CI for $DRONE_BUILD_EVENT on $DRONE_BRANCH"
-    pybot --removekeywords TAG:secret --exclude skip tests/test-cases
+    pabot --processes 2 --removekeywords TAG:secret --exclude skip tests/test-cases
```
hickeng: parameterize the number of processes or do we want to be able to chose per-scenario values?

PR #5602, opened 2017-06-27
[full ci] Fix ci test failure in test 6-07

chengwang86:
Fix the ci test failure seen in builds 11898 and 11907


isos/bootstrap/bootstrap
```diff
@@ -38,6 +38,9 @@ if mount -t ext4 /dev/disk/by-label/containerfs ${MOUNTPOINT}; then
     mount -t tmpfs -o size=64m tmpfs ${MOUNTPOINT}/.tether/
 
     # enable full system functionality in the container
+    ln -s lib64 ${MOUNTPOINT}/.tether/lib
+    mkdir -p ${MOUNTPOINT}/.tether/{lib64,usr/lib/iptables}
```
mhagen-vmware: should be not ln to the dir after we mkdir it?

chengwang86: @mhagen-vmware It looks like `ln` will force `mkdir` if the dir does not exist. For example, the following succeeded on my local machine:
```
mkdir aaa
ln -s aaa bbb
ls -al | grep bbb
```
output: `bbb -> aaa`

chengwang86: @mhagen-vmware when you create symbolic links by `ln -s`, the target file does not have to exist beforehand.

mhagen-vmware: sure but it just increases readability maybe

PR #5608, opened 2017-06-28
[skip ci] Docker push logic flow

sgairo:


hickeng:
How involved is it to stream the layers up? We are going to have issues with many images if we are attempting to cache the entire image data before starting to push it to the registry, without even considering concurrent pulls/pushes.
Is it going to be simpler to implement the streaming rather than implement checks that determine whether we're going to be able to stage all of the layers?

sgairo:
@hickeng its definitely _more_ involved but we haven't been able to focus on exactly _how_ involved it is. It was our understanding that we would do it the suboptimal way as a first pass and come back around to improving it. If your recommendation, considering time as a factor, is to go ahead and stream it right off the bat, that is how I can proceed. Other than that piece, did you approve of the workflow?

sgairo:
@hickeng After talking with those on docker push, the logic here is still sound. I removed the text from this image. Also, we did decide to go ahead and stream the layers. Could I get you to look at this again so we can wrap it up. If there is still work that needs to be changed let me know. Thanks!

Issue #5552, opened 2017-06-21
Improve H5 plugin install.sh


hickeng:
**Story**
As a viadmin I want to be able install/upgrade the H5 plugin with as minimal interaction as possible

**Detail**
The current install and upgrade scripts for the H5 client requires editing of a config file before running the scripts. The bulk of the edits are for the connection data for the OVA file server.
This issue should generate the config file that's returned, injecting the URL, and thumbprint for the OVA file server.
User should still be prompted for:
* vCenter IP
* vCenter credentials

It's possible that the config file should be removed completely and these changes be injected into the install/upgrade scripts - this would allow a single file instead of needing a tar or multiple downloads.

jooskim:
Assigning myself to this issue along with Emma, as I wrote the scripts

emlin:
@hickeng Per discussion with @jooskim , we prefer to keep the current config file, as that will keep similar user experience.
And also cause the configuration is separated, user could easily know what server information is used. In the worst case, if there is anything wrong happens during ova data population, it is still easy to manually correct the separate config file instead of search in a long script file.

andrewtchin:
After the user enters the credentials and target, I would like this to display the fingerprint and prompt the user to accept it so they can verify it if they want

jooskim:
I think that's a good addition. @emlin what do you think? Assuming we ask the VC IP in `install.sh`, then it should prompt the user four times - VC IP, VC admin ID, VC admin PW and thumbprint verification.

I can modify the script such that it calls `vic-ui` twice for Flex and H5 Client plugins.

emlin:
@andrewtchin are you talking about VC thumbprint or file server's thumbprint?

andrewtchin:
That's for vcenter fingerprint since we don't want to give vcenter creds to man in the middle

jooskim:
btw, if we want to verify the VC fingerprint how do we get it from the machine where the script is run? we don't assume the user has `openssl` or something

emlin:
@jooskim that's what I'm typing. With this requirement, vic-ui command should be updated to get back this information.
In your script, you can printed out the thumbprint got from vic-ui, and then user could simply trust that information, or try to get back that information from web browser, and posted into the command

emlin:
@karthik-narayan @bradmeiseles @mdubya66 

To improve user experience of installing UI plugin in 1.2 release, we're planning to improve the script a little bit. Then the user experience will look like following:
- Go to ova file server URL
- Download file package
[...]

PR #5611, opened 2017-06-28
Improve UI installer scripts

jooskim:
This is a second PR for #5552. Based on the requirements of the issue and discussions with @emlin the updated script will no longer require the user to edit the `configs` file before running `install.sh` or `install.bat`. For this to happen, two pre-requisites need to be satisfied:

1. The OVA appliance should modify the `configs` file at its boot such that the plugin URL and thumbprint information get populated.
2. `vic-ui` binary should be able to detect if there is any existing installation of VIC Engine UI and what version, if any. This is already taken care of in #5609 

One of the major changes this PR brings is that the installer, upgrader or uninstaller will now work  on the BOTH plugins (for Flex Client and H5 Client) so that the user no longer has to specify which plugin s/he wants to install/uninstall.

Next, it now requires the user to validate and accept the VC's thumbprint. If the user decides not to trust it, then s/he will be asked to enter one manually. Once the thumbprint is accepted, the `install.sh|bat` script will check for an existing plugin and WILL fail if there is any. This can be remedied by running `install.sh|bat --force`, or `upgrade.sh|bat`. `upgrade.sh|bat` will prompt the user what version is going to be installed and will warn the user if the version is the same as or less than the already installed version. `install.sh|bat --force` on the other hand will proceed to install the plugins directly after verifying the VC thumbprint. 

It is also made clear in the `configs` file that it should no longer be edited by the user manually, but in case it needs to, the thumbprint could have either `:` or ` ` delimiters, as the script will replace spaces with colons.

### DIFF
--- a/tests/manual-test-cases/Group18-VIC-UI/18-2-VIC-UI-Uninstaller.robot
+++ b/tests/manual-test-cases/Group18-VIC-UI/18-2-VIC-UI-Uninstaller.robot
@@ -17,19 +17,10 @@ Documentation  Test 18-2 - VIC UI Uninstallation
 Resource  ../../resources/Util.robot
 Resource  ./vicui-common.robot
 Test Teardown  Cleanup Installer Environment
-Suite Setup  Check Config And Install VCH
-Suite Teardown  Uninstall VCH
+Suite Setup  Check Config And Install VCH  install
+Suite Teardown  Uninstall VCH  ${TRUE}
 
 *** Test Cases ***
-Ensure Vicui Is Installed Before Testing
-    Set Vcenter Ip
-    Install Vicui Without Webserver  ${TEST_VC_USERNAME}  ${TEST_VC_PASSWORD}  ${TEST_VC_ROOT_PASSWORD}  ${TRUE}
-    ${output}=  OperatingSystem.GetFile  install.log
-    ${passed}=  Run Keyword And Return Status  Should Contain  ${output}  was successful
-    Run Keyword Unless  ${passed}  Copy File  install.log  install-fail-ensure-vicui-is-installed-before-testing.log
-    Remove File  install.log
-    Should Be True  ${passed}
-
 Attempt To Uninstall With Configs File Missing
     # Rename the configs file and run the uninstaller script to see if it fails in an expected way
     Move File  ${UI_INSTALLER_PATH}/configs  ${UI_INSTALLER_PATH}/configs_renamed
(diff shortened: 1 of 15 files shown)


jooskim:
@stuclem I guess would be great if we could review this together, as the workflow has pretty much changed now.

jooskim:
Another thing to be noted is that the `NOURL` workaround was dropped, as it no longer makes sense when we can get `configs` pre-populated.

ui/installer/VCSA/install.sh
```diff
@@ -36,14 +39,10 @@ while IFS='' read -r line; do
     eval $line
 done < ./configs
 
-# check for the VC IP
-if [[ $VCENTER_IP == "" ]] ; then
-    echo "Error! vCenter IP cannot be empty. Please provide a valid IP in the configs file"
-    cleanup
-    exit 1
-fi
+# replace space delimiters with colon delimiters
```
emlin: The thumbprint set into configs is already in this format:
`"4E:A2:37:C4:DF:B3:1A:EE:DE:94:70:79:60:5E:DC:CF:34:8F:B9:83"`

ui/installer/VCSA/uninstall.sh
```diff
@@ -26,11 +33,8 @@ while IFS='' read -r line; do
     eval $line
 done < ./configs
 
-# check for the VC IP
-if [[ $VCENTER_IP == "" ]] ; then
-    echo "Error! vCenter IP cannot be empty. Please provide a valid IP in the configs file"
-    exit 1
-fi
+# replace space delimiters with colon delimiters
+VIC_UI_HOST_THUMBPRINT=$(echo $VIC_UI_HOST_THUMBPRINT | sed -e 's/[[:space:]]/\:/g')
```
emlin: ditto

ui/installer/VCSA/install.sh
jooskim: Yeah I am aware of that, but in case some user manually updates it which should never be the case now that we have a warning message in the configs file, this still could work as a safety net.

tests/manual-test-cases/Group18-VIC-UI/18-2-VIC-UI-Uninstaller.robot
```diff
@@ -22,10 +22,14 @@ Suite Teardown  Uninstall VCH
 
 *** Test Cases ***
 Ensure Vicui Is Installed Before Testing
```
mhagen-vmware: This should really be in the suite setup, and this whole structure is pretty error prone - if any of these attempted uninstalls suddenly work for some reason then all the rest of the tests fail.  Better to have a test setup that checks whether it is installed first, then installs if necessary on each test.

andrewtchin:
screen shot 2017-06-28 at 1 06 59 pm

jooskim:
`upgrade.sh\bat` were merely `install.sh|bat --force` but they have their own logics now. That's why there's a big jump

jooskim:
@andrewtchin @frapposelli agreed. Fortunately though we are considering to use a Web installer  for 1.3, which would do pretty much the same as what these scripts do.

tests/manual-test-cases/Group18-VIC-UI/18-3-VIC-UI-NGC-tests.robot
```diff
@@ -36,17 +39,21 @@ Check Prerequisites
...
-    Install Vicui Without Webserver  ${TEST_VC_USERNAME}  ${TEST_VC_PASSWORD}  ${TEST_VC_ROOT_PASSWORD}  ${TRUE}
+    #Install Vicui Without Webserver  ${TEST_VC_USERNAME}  ${TEST_VC_PASSWORD}  ${TEST_VC_ROOT_PASSWORD}  ${TRUE}
+    Update Config To Use Webserver For Plugin Testing
+    Install With Webserver  ${TEST_VC_USERNAME}  ${TEST_VC_PASSWORD}
     ${output}=  OperatingSystem.GetFile  install.log
-    Should Contain  ${output}  was successful
+    ${passed}=  Run Keyword And Return Status  Should Contain  ${output}  was successful
+    Run Keyword Unless  ${passed}  Copy File  install.log  ngc-tests-fail-install-plugin-using-webserver-for-plugin-testing.log
+    Remove File  install.log
+    Should Be True  ${passed}
```
mhagen-vmware: This whole section is kinda weird...
1. maybe just move instead of copy, then remove?
2. you are just trying to ensure that the install succeeded right? this seems like a lot of extra steps that are more than that
3. Install with webserver should really just ensure that the install was successful within it.

tests/manual-test-cases/Group18-VIC-UI/18-3-VIC-UI-NGC-tests.robot
```diff
@@ -72,7 +89,7 @@ Reboot vSphere Client
     Run Keyword If  %{TEST_VSPHERE_VER} == 60  Execute Command  service vsphere-client start  ELSE  Execute Command  service-control --start vsphere-ui
 
     # wait until vsphere-client/ui server is up and running
-    Wait Until Keyword Succeeds  10x  60 s  Is vSphere Client Ready
+    Wait Until Keyword Succeeds  20x  30 s  Is vSphere Client Ready
```
mhagen-vmware: 30 s -> 30s please

tests/manual-test-cases/Group18-VIC-UI/18-3-VIC-UI-NGC-tests.robot
```diff
@@ -56,6 +63,16 @@ Run Ngc Tests Project
 
 
 *** Keywords ***
+Update Config To Use Webserver For Plugin Testing
```
mhagen-vmware: I don't have a good view into the larger picture here, but this looks like it is something that needs to be pulled out into a UI-Util.robot and you have a lot of other repeated code that should probably be pulled into a unified library to make these tests more readable as well.

tests/manual-test-cases/Group18-VIC-UI/18-3-VIC-UI-NGC-tests.robot
```diff
@@ -151,23 +168,20 @@ Create And Run Test Container
...
     ${rc}  ${container_name}=  Run And Return Rc And Output  docker %{VCH-PARAMS} inspect ${container_id} | jq '.[0].Name' | sed 's/[\"\/]//g'
     Should Be Equal As Integers  ${rc}  0
-    Set Suite Variable  ${CONTAINER_VM_NAME}  ${container_name}-${container_id}
-    Set Suite Variable  ${CONTAINER_VM_TRUNCATED_NAME}  ${container_name}-${container_id}
-    ${p_name_length}=  Get Length  ${container_name}
-    ${container_name_truncated}=  Run Keyword If  ${p_name_length} > 15  Get Substring  ${container_name}  0  15
-    Run Keyword If  ${p_name_length} > 15  Log To Console  Container name is longer than 15 characters. Truncating it to the first 15 characters... (${container_name_truncated}-${container_id})
-    Run Keyword If  ${p_name_length} > 15  Set Suite Variable  ${CONTAINER_VM_TRUNCATED_NAME}   ${container_name_truncated}-${container_id}
+    ${short_container_id}=  Get Substring  ${container_id}  0  12
+    Set Suite Variable  ${CONTAINER_VM_TRUNCATED_NAME}  ${container_name}-${short_container_id}
```
mhagen-vmware: We have this already:
${shortID}=  Get container shortID  ${id}

tests/manual-test-cases/Group18-VIC-UI/setup-testbed.robot
```diff
@@ -207,7 +207,7 @@ Deploy Nimbus BrowserVm For NGC Testing
     [Return]  ${user}-${name}  ${ip}
 
 Deploy Nimbus ESXi Server For NGC Testing
```
mhagen-vmware: This is just the standard nimbus esxi deploy is it not?  Why not use our existing keyword with the version set to the right one?

tests/manual-test-cases/Group18-VIC-UI/setup-testbed.robot
```diff
@@ -236,7 +236,7 @@ Deploy Nimbus ESXi Server For NGC Testing
     [Return]  ${user}-${name}  ${ip}
 
 Deploy Nimbus vCenter Server For NGC Testing
```
mhagen-vmware: This also appears to be just the same as our deploy nimbus vcenter with the --useQaNgc flag, which can be pushed into our keyword by overloading the name field.

tests/manual-test-cases/Group18-VIC-UI/vicui-common.robot
```diff
@@ -105,29 +107,18 @@ Set Script Filename
...
     Should Exist  ${UI_INSTALLER_PATH}/configs
 
 Force Remove Vicui Plugin
-    Uninstall Vicui  ${TEST_VC_USERNAME}  ${TEST_VC_PASSWORD}
-    ${output}=  OperatingSystem.GetFile  uninstall.log
-    Should Match Regexp  ${output}  (unregistration was successful|failed to find target plugin)
-    Remove File  uninstall.log
+    # Uninstall Vicui  ${TEST_VC_IP}  ${TEST_VC_USERNAME}  ${TEST_VC_PASSWORD}  ${TRUE}
+    ${rc}  ${output}=  Run And Return Rc And Output  ../../../ui-nightly-run-bin/vic-ui-linux remove --thumbprint %{VC_FINGERPRINT} --target ${TEST_VC_IP} --user ${TEST_VC_USERNAME} --password ${TEST_VC_PASSWORD} --key com.vmware.vic.ui
+    ${rc}  ${output}=  Run And Return Rc And Output  ../../../ui-nightly-run-bin/vic-ui-linux remove --thumbprint %{VC_FINGERPRINT} --target ${TEST_VC_IP} --user ${TEST_VC_USERNAME} --password ${TEST_VC_PASSWORD} --key com.vmware.vic
```
mhagen-vmware: This file appears to be what I was commenting about above.  This should be pulled into the resources folder as UI-Util.robot (or some such)

tests/manual-test-cases/Group18-VIC-UI/18-3-VIC-UI-NGC-tests.robot
jooskim: After each run of UI nightly test I want any failed test case to have its log ready in the bundle (in this case it would be ngc-tests-fail-blabla.log) Note that I have a different kickoff script that runs only on the VIC-Nightly VM.

tests/manual-test-cases/Group18-VIC-UI/18-2-VIC-UI-Uninstaller.robot
jooskim: Will move `Ensure Vicui Is Installed Before Testing` and a similar one in 18-1 to suite setup. Since the actual UI plugin test happens in 18-3, all that matters for 18-1 and 18-2 is whether or not the extension is registered with VC.

tests/manual-test-cases/Group18-VIC-UI/18-3-VIC-UI-NGC-tests.robot
jooskim: `vicui-common.robot` is pretty much the place for those common stuff. I'll try refactoring some existing code.

tests/manual-test-cases/Group18-VIC-UI/setup-testbed.robot
jooskim: The main reason I decided to use my own keyword is that I need to use a different naming convention for the testbeds. For example, the name for a VCVA to be created should follow the following convention.

```
${name}=  Evaluate  'UITEST-VC%{TEST_VSPHERE_VER}-' + str(random.randint(1000,9999))  modules=random
```
On the other hand, the existing keyword from Nimbus-Util.robot doesn't capture my need.
```
${name}=  Evaluate  'VC-' + str(random.randint(1000,9999))  modules=random
```

tests/manual-test-cases/Group18-VIC-UI/setup-testbed.robot
jooskim: If the name field can be overloaded I could use the existing keyword. I don't currently see it capable of doing that.

tests/manual-test-cases/Group18-VIC-UI/vicui-common.robot
jooskim: `vicui-common.robot` is the place for commonly used keywords for my case

jooskim:
@stuclem I just tested upgrading from 1.1.0 to 1.1.1 in both clients and there is one very important note to be made here. While the upgrade process should work out of the box without having to restart the Flex & H5 Client servers, the UI plugin upgrade process from 1.1.0 to 1.1.1, or any later version should be followed by the restart of the both client services. This is because one of the critical issues (that was solved by the version 1.1.1) was the H5 Client plugin getting activated in the Flex Client while it's not supposed to, and once activated, it cannot be completely disabled unless the Flex Client service is restarted.

andrewtchin:
@jooskim how do you restart Flex Client service? 
Pinging @alextopuzov too

jooskim:
@andrewtchin Users can use VAMI (https://$VCSA_IP:5480) to reboot the entire VCSA. It was possible before to restart the Flex Client in itself but it no longer seems possible. Restarting the VCSA would reboot the both clients.

tests/manual-test-cases/Group18-VIC-UI/setup-testbed.robot
mhagen-vmware: sorry should have been more clear, we are deprecating the standard ESX and VC deploys in favor of async deploys, so overloading the name field for VC deploy would be:

Deploy Nimbus vCenter Server Async  UITEST-VC%{TEST_VSPHERE_VER}-rand --useQANgc

(note single space between name and useQANgc)

tests/manual-test-cases/Group18-VIC-UI/setup-testbed.robot
mhagen-vmware: Deploy Nimbus ESXi Server Async

and pass in whatever name you want

PR #5623, opened 2017-06-29
Don't rely on line numbers and only verify the VCH we care about

### LINKED ISSUE #5258: Improve "cleanup dangling VMs On Test Server"
Currently in the test script we do not remove a dangling VCH if it is in the `running` status. However, a VCH might be left in the `running` status if the test case that is performed on it fails. Therefore, certain later tests (e.g., 6-10-List in Group6-vic-machine) might fail due to the existence of such dangling VCH's. An example is here https://ci.vcna.io/vmware/vic/10897.

Another negative impact is that these dangling `running` VCH's might pile up on the test server and still consuming resources. 

Suggestions:
1. after we finish a build, remove all the dangling VCH's left in this build even if it is `running`.
2. force deleting any `running` VCH's belonging to the current build before we run the next test in the same build. It might be dangerous to do so for all the tests since I assume some of the tests might rely on the `running` VCH's created in previous tests. However, at least we need to do this for test 6-10-List due to reasons above.

Acceptance criteria:
- after a build finishes, there is no `running` VCH's left with that particular build number.
- test 6-10-List should not fail due to the above reason

mhagen-vmware:
fixes #5258

### DIFF
--- a/tests/test-cases/Group6-VIC-Machine/6-10-List.robot
+++ b/tests/test-cases/Group6-VIC-Machine/6-10-List.robot
@@ -20,7 +20,7 @@ Suite Teardown  Cleanup VIC Appliance On Test Server
 
 *** Keywords ***
 Verify Listed Machines
-    [Arguments]  ${list}  ${total}
+    [Arguments]  ${list}
     Should Contain  ${list}  ID
     Should Contain  ${list}  PATH
     Should Contain  ${list}  NAME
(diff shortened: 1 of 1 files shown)


chengwang86:
so the goal of this test for `vic-machine ls` is to check whether the cmd proceeds successfully instead of verifying the validity of the output?

For example, back to several months ago, we had the issue that `vic-machine ls` output each vch twice due to a bug in filtering the inventory folder. What if a similar bug occurred? Will it be masked by this PR?

mhagen-vmware:
@chengwang86 done

Issue #5551, opened 2017-06-21
Run admiral setup command at OVA startup


hmahmood:
Admiral will provide a command that will:
* Setup OIDC/SAML-WS-TRUST for Admiral against the PSC,
* Create a few (most likely 3) “solution-users” in the PSC (Admiral, Harbor, VCH-content-trust), their ID and (potentially )keys will be outputted by this command. The generated output files should be stored on a volume which is accessible by the OVA init script and by the containers running Admiral/Harbor. It is not clear yet how many files will be generated (I believe there might be a couple of files per user)

This command needs to be run on OVA first run.

Acceptance:
* Admiral provided command runs at OVA first run

andrewtchin:
Need to determine if this API will require creds

hmahmood:
@andrewtchin it definitely does ... I think @lcastellano mentioned that in his email.

anchal-agrawal:
Merged PR https://github.com/vmware/vic-product/pull/415 covers:
* registering Admiral with PSC to create PSC config and cert files
* not starting Admiral and Harbor until the above

The next steps are:
* starting Admiral in the OVA with the said PSC config file as a parameter (waiting on implementation details from Admiral)
* running the PSC tool's `get-token` command to create token files - this can be done with systemd unit(s)
* based on our discussions, Harbor startup should be blocked until the token file is generated.

PR #5624, opened 2017-06-29
Add query options helper in pkg/vsphere/session

anchal-agrawal:
Needed for obtaining the hostname of the vCenter host for the PSC command in the OVA.

Towards #5551

### DIFF
--- a/pkg/vsphere/session/session.go
+++ b/pkg/vsphere/session/session.go
@@ -319,6 +319,23 @@ func (s *Session) Folders(ctx context.Context) *object.DatacenterFolders {
 	return s.folders
 }
 
+// QueryOptionValue looks up the option and returns the value. If more/less values are
+// received, an error is returned.
+func (s *Session) QueryOptionValue(ctx context.Context, option string) (string, error) {
+	client := s.Vim25()
+	optMgr := object.NewOptionManager(client, *client.ServiceContent.Setting)
+	opts, err := optMgr.Query(ctx, option)
+	if err != nil {
+		return "", fmt.Errorf("error querying option %q: %s", option, err)
+	}
+
+	if len(opts) == 1 {
+		return fmt.Sprintf("%v", opts[0].GetOptionValue().Value), nil
+	}
+
+	return "", fmt.Errorf("%d values querying option %q", len(opts), option)
+}
+
 func isNotAuthenticated(err error) bool {
 	if soap.IsSoapFault(err) {
 		switch soap.ToSoapFault(err).VimFault().(type) {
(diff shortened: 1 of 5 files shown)


pkg/vsphere/session/session.go
```diff
@@ -319,6 +319,28 @@ func (s *Session) Folders(ctx context.Context) *object.DatacenterFolders {
...
+		return "", fmt.Errorf("session client is nil")
+	}
+
+	optMgr := object.NewOptionManager(client, *client.ServiceContent.Setting)
+	opts, err := optMgr.Query(ctx, option)
+	if err != nil {
+		return "", fmt.Errorf("error querying option: %s", err)
+	}
+
+	for _, opt := range opts {
```
dougm: Maybe instead of the for loop, if len(opt) == 1, return o.Value.  Else, fallthrough to return fmt.Errorf("%d values querying option %q", len(opt), option)

pkg/vsphere/session/session.go
```diff
@@ -319,6 +319,28 @@ func (s *Session) Folders(ctx context.Context) *object.DatacenterFolders {
 	return s.folders
 }
 
+// QueryOption looks up the option's values and returns the first one from the
+// ones received on success.
+func (s *Session) QueryOption(ctx context.Context, option string) (string, error) {
+	client := s.Vim25()
+	if client == nil {
```
dougm: Do we need this check here?

anchal-agrawal: Originally added this as a fail-safe, but I traced the validator/session calls and looks like the client will be non-nil here since errors from `session.Connect()` are not ignored. I'll remove this check.

anchal-agrawal:
Pinging @dougm for another gander - I've added a unit test and cleaned up some code.

pkg/vsphere/session/session.go
anchal-agrawal: Fixed!

pkg/vsphere/session/session.go
```diff
@@ -328,3 +345,49 @@ func isNotAuthenticated(err error) bool {
 	}
 	return false
 }
+
+// GetESXSession returns a general-purpose ESX session for tests.
+func GetESXSession(ctx context.Context, service string) (*Session, error) {
```
dougm: Not a fan of having test-only helpers in packages that are used outside of tests.  Sadly we have some of that already in vsphere/datastore/datastore_util.go.  The tough part would be coming up with a decent package name ;-/

emlin: Just had a conversation with @anchal-agrawal for this code change. Cause that is to make the test code easier to use vcsim, it might be better to see some helper function in vcsim, but now vcsim is moved out of vic repo.
So far, the suggestion is to move it into https://github.com/vmware/vic/blob/master/pkg/vsphere/test/test.go, where the package is only used by test code.
How do you think?

dougm: Yes, I think a `test.SessionWithESX` and `test.SessionWithVPX` would be a good place for these helpers.  Don't think these helpers could be in the simulator package anyhow, since they are specific to the Session type.  Also note that Session.Populate would take care of filling in most of the Path fields as it uses the finder OrDefault methods, so they shouldn't need to be hardcoded.

anchal-agrawal:
Closing this PR since we've decided to use a different approach with this helper function. I'll open another one.

PR #5625, opened 2017-06-29
Add 5-21-Datastore-Path test to nightly email summary

rogeliosanchez:


rajanashok:
I think you might have to update the following line, as this gets incremented.

# Setting the NSX test status to Not Implemented.
nightlystatus[24]="N/A"

PR #5627, opened 2017-06-30
Add InUse methods to DiskManager

### LINKED ISSUE #5564: Portlayer: Implement functionality to answer "who uses the disk" question
Story
As a VIC developer I need find the VM that uses the disk to support online copy operation.

Detail
vSphere API provides no direct way to get that info. Need to consult storage team to learn whether there is one.
If not, getting the powered on VMs and iterating over them could be a solution.

Acceptance

A method returns the VM that is using the disk and tests

caglar10ur:
Fixes #5564

### DIFF
--- a/pkg/vsphere/disk/disk_manager.go
+++ b/pkg/vsphere/disk/disk_manager.go
@@ -21,7 +21,11 @@ import (
 	"os"
 	"sync"
 
+	log "github.com/Sirupsen/logrus"
+
 	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/view"
+	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/trace"
@@ -51,7 +55,10 @@ type Manager struct {
 	// VirtualDiskManager that is used to create vmdks directly on datastore
 	// from https://pubs.vmware.com/vsphere-65/index.jsp?topic=%2Fcom.vmware.vspsdk.apiref.doc%2Fvim.VirtualDiskManager.html
 	// Most VirtualDiskManager APIs will be DEPRECATED as of vSphere 6.5. Please use VStorageObjectManager APIs to manage Virtual disks.
-	vdm *object.VirtualDiskManager
+	vdMngr *object.VirtualDiskManager
+
+	// ContainerView - https://pubs.vmware.com/vsphere-6-0/index.jsp#com.vmware.wssdk.apiref.doc/vim.view.ContainerView.html
+	view *view.ContainerView
 
 	// The controller on this vm.
 	controller *types.ParaVirtualSCSIController
@@ -67,7 +74,7 @@ type Manager struct {
 }
 
 // NewDiskManager creates a new Manager instance associated with the caller VM
-func NewDiskManager(op trace.Operation, session *session.Session) (*Manager, error) {
+func NewDiskManager(op trace.Operation, session *session.Session, v *view.ContainerView) (*Manager, error) {
 	defer trace.End(trace.Begin(""))
 
 	vm, err := guest.GetSelf(op, session)
@@ -85,7 +92,8 @@ func NewDiskManager(op trace.Operation, session *session.Session) (*Manager, err
 	return &Manager{
 		maxAttached:  make(chan bool, MaxAttachedDisks),
 		vm:           vm,
-		vdm:          object.NewVirtualDiskManager(vm.Vim25()),
+		vdMngr:       object.NewVirtualDiskManager(vm.Vim25()),
+		view:         v,
 		controller:   controller,
 		byPathFormat: byPathFormat,
 		Disks:        make(map[uint64]*VirtualDisk),
@@ -165,6 +173,13 @@ func (m *Manager) CreateAndAttach(op trace.Operation, config *VirtualDiskConfig)
 	// ensure we abide by max attached disks limits
 	m.maxAttached <- true
 
+	// make sure the op is still valid as the above line could block for a long time
+	select {
+	case <-op.Done():
+		return nil, op.Err()
+	default:
+	}
+
 	op.Infof("Create/attach vmdk %s from parent %s", config.DatastoreURI, config.ParentDatastoreURI)
 
 	if err := m.attach(op, config); err != nil {
@@ -228,7 +243,7 @@ func (m *Manager) Create(op trace.Operation, config *VirtualDiskConfig) (*Virtua
 
 	op.Infof("Creating vmdk for layer or volume %s", d.DatastoreURI)
 	err = tasks.Wait(op, func(ctx context.Context) (tasks.Task, error) {
-		return m.vdm.CreateVirtualDisk(ctx, d.DatastoreURI.String(), nil, spec)
+		return m.vdMngr.CreateVirtualDisk(ctx, d.DatastoreURI.String(), nil, spec)
 	})
 
 	if err != nil {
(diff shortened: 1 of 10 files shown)


pkg/vsphere/disk/disk_manager.go
```diff
@@ -421,3 +428,48 @@ func (m *Manager) devicePathByURI(op trace.Operation, datastoreURI *object.Datas
 
 	return sysPath, nil
 }
+
+func (m *Manager) InUse(op trace.Operation, config *VirtualDiskConfig) (*vm.VirtualMachine, error) {
```
matthewavery: Is it possible that the same disk can be found in several hardware config of several vms? I know when we join a volume to a cvm we perform a  `types.VirtualDeviceConfigSpecOperationAdd` if this populates the `vms[i].Config.Hardware.Device` portion then it is possible that we might find multiple matches for this search. I poked a round for a short bit in the `VirtualHardware` api page [here](https://www.vmware.com/support/developer/vc-sdk/visdk41pubs/ApiReference/vim.vm.VirtualHardware.html). If it is possible Could we possibly return a list instead of a single target? 

I also noticed in the VirtualDevice object( [here](https://www.vmware.com/support/developer/vcdk/visdk41pubs/ApiReference/vim.vm.device.VirtualDevice.html) ) that there is a `connectable` attribute which contains some useful information device connection. 
I was wondering if we could possibly check for this for the case of whether the disk is in use rather than looking through only our containers.

caglar10ur: It is not possible to attach the same vmdk to multiple vms with our current usage (means there is a way but we are not using shared disks). 

I'm not sure what you mean by the second part but I'm guessing you are suggesting checking connected field when we find a match before returning it. AFAIK it doesn't matter whether the disk is connected or not as we still cannot attach it to some other vm without detaching first.

pkg/vsphere/disk/disk_manager_test.go
```diff
@@ -369,3 +371,179 @@ func TestRefCounting(t *testing.T) {
 		return
 	}
 }
+
+func TestInUse(t *testing.T) {
```
matthewavery: Possibly have a test for a disk that is attached to two vms. One on and one off.

pkg/vsphere/disk/disk_manager.go
```diff
@@ -49,7 +53,11 @@ type Manager struct {
 	// VirtualDiskManager that is used to create vmdks directly on datastore
 	// from https://pubs.vmware.com/vsphere-65/index.jsp?topic=%2Fcom.vmware.vspsdk.apiref.doc%2Fvim.VirtualDiskManager.html
 	// Most VirtualDiskManager APIs will be DEPRECATED as of vSphere 6.5. Please use VStorageObjectManager APIs to manage Virtual disks.
-	vdm *object.VirtualDiskManager
+	vdMngr *object.VirtualDiskManager
+
+	// ViewManager that is used to create views directly
+	// from https://vdc-repo.vmware.com/vmwb-repository/dcr-public/6b586ed2-655c-49d9-9029-bc416323cb22/fa0b429a-a695-4c11-b7d2-2cbc284049dc/doc/vim.view.ContainerView.html
```
hickeng: Slightly more public/stable? url:
http://pubs.vmware.com/vsphere-6-0/index.jsp#com.vmware.wssdk.apiref.doc/vim.view.ContainerView.html

pkg/vsphere/disk/disk_manager.go
hickeng: There are two primary use cases here:
1. find the _singluar_ VM that is powered on and using the disk (vSphere enforces this unless `ddb.multiWriter` is set to true)
2. find all VMs with the disk attached, including powered off VMs 

For this method to be used as a replacement for the exec cache lookup we can either have this method take a filter argument for power state or we can have it return a list and have that filtering done at a higher level.

pkg/vsphere/disk/disk_manager.go
```diff
@@ -421,3 +430,53 @@ func (m *Manager) devicePathByURI(op trace.Operation, datastoreURI *object.Datas
 
 	return sysPath, nil
 }
+
+func (m *Manager) InUse(op trace.Operation, config *VirtualDiskConfig) (*vm.VirtualMachine, error) {
+	defer trace.End(trace.Begin(""))
+
+	var vms []mo.VirtualMachine
+
+	ctx, cancel := context.WithCancel(context.Background())
```
hickeng: use `op` as the parent context

pkg/vsphere/disk/disk_manager.go
```diff
@@ -421,3 +430,53 @@ func (m *Manager) devicePathByURI(op trace.Operation, datastoreURI *object.Datas
...
+	defer cancel()
+
+	parent, err := m.vm.ResourcePool(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Create view of VirtualMachine objects under the VCH's resource pool
+	v, err := m.vMngr.CreateContainerView(ctx, parent.Reference(), []string{"VirtualMachine"}, true)
```
hickeng: Can we hang the view off `storage.Config` or similar and create it on-demand if not present (or not functional)? That would avoid three API calls per invocation of this method.

pkg/vsphere/disk/disk_manager.go
```diff
@@ -421,3 +430,53 @@ func (m *Manager) devicePathByURI(op trace.Operation, datastoreURI *object.Datas
...
+
+	// Retrieve config.hardware property of all machines under this view
+	err = v.Retrieve(ctx, []string{"VirtualMachine"}, []string{"config.hardware", "runtime.powerState"}, &vms)
+	if err != nil {
+		return nil, err
+	}
+
+	// iterate over them to see whether they have the disk we want
+	for i := range vms {
+		if vms[i].Runtime.PowerState != types.VirtualMachinePowerStatePoweredOn {
```
hickeng: I don't know if a suspended VM locks the disk...

pkg/vsphere/disk/disk_manager.go
caglar10ur: I swear this was not working yesterday and redirecting to new doc.vmware.com :)

pkg/vsphere/disk/disk_manager.go
caglar10ur: Done

pkg/vsphere/disk/disk_manager.go
caglar10ur: It does as far as my silly test goes so added that

pkg/vsphere/disk/disk_manager.go
caglar10ur: Working on it

pkg/vsphere/disk/disk_manager.go
```diff
@@ -421,3 +436,40 @@ func (m *Manager) devicePathByURI(op trace.Operation, datastoreURI *object.Datas
...
+	err := m.view.Retrieve(op, []string{"VirtualMachine"}, []string{"config.hardware", "runtime.powerState"}, &vms)
+	if err != nil {
+		return nil, err
+	}
+
+	// iterate over them to see whether they have the disk we want
+	for i := range vms {
+		name := vms[i].Name
+		state := vms[i].Runtime.PowerState
+		if state != types.VirtualMachinePowerStatePoweredOn && state != types.VirtualMachinePowerStateSuspended {
```
emlin: I manually tested to attach one disk in one suspended vm, that works well, and if I try to start the suspended vm after attached it's vmdk, the vm start failed.
So I'm wondering why suspended vm is not filtered out?

caglar10ur: it is filtered out but I'm going to change this filtering logic like George suggested https://github.com/vmware/vic/pull/5627#discussion_r125130965

pkg/vsphere/disk/disk_manager.go
caglar10ur: Done

pkg/vsphere/disk/disk_manager.go
caglar10ur: Done

caglar10ur:
@hickeng FYI, I opted to creating a ContainerView on portlayer starts as we cannot get the list of views unless that view is associated with the session we are using. And views are getting removed when the session is destroyed (confirmed that behaviour with mob browser)

caglar10ur:
@hickeng @emlin ping

caglar10ur:
@jzt told me that @hickeng says yes :) Merging

PR #5630, opened 2017-06-30
Add query options helper in pkg/vsphere

anchal-agrawal:
Needed for obtaining the hostname of the vCenter host for the PSC command in the OVA. Towards #5551

(Follow-up to unmerged PR #5624)

### DIFF
--- a/lib/install/management/configure_test.go
+++ b/lib/install/management/configure_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/vmware/vic/lib/install/data"
 	"github.com/vmware/vic/pkg/vsphere/extraconfig"
 	"github.com/vmware/vic/pkg/vsphere/session"
+	"github.com/vmware/vic/pkg/vsphere/test"
 	"github.com/vmware/vic/pkg/vsphere/vm"
 )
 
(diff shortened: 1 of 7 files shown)


anchal-agrawal:
@dougm The reasons we moved `QueryOptionValue` to its own pkg are:

1) The only use of `session` in `QueryOptionValue` is to grab the client. Looking at the other functions in `pkg/vsphere/session`, @emlin and I felt that `QueryOptionValue` didn't fit in that pkg well.
2) We ran into an import cycle with the test helpers `SessionWithESX` and `SessionWithVPX` upon putting them in `pkg/vsphere/test` or `pkg/vsphere/session/test`. These helpers aren't directly related to this change at all - I needed to use `SessionWithVPX` and noticed that we had the helpers duplicated in two places, so I consolidated them.

I understand that it's not ideal to put test-only helpers in non-test packages. Therefore, we can:

1) put `QueryOptionValue` in a separate pkg, or
2) put it in `pkg/vsphere/session`, and not de-duplicate the test session helpers. I think duplicating test-only helpers is preferable to having them in the main pkg.

What do you think?

dougm:
"session" was meant to be a collection of small reusable shortcuts around govmomi, fine with me to include QueryOptionValue and/or the test helpers there.  The idea of a new package that will only ever have 1 method doesn't seem right, but is just cosmetic, its fine functionally of course.  Up to you.

PR #5633, opened 2017-07-02
Fix refcount tracking for CreateAndAttach

jzt:
Fixes a small bug in refcount tracking. Required for https://github.com/vmware/vic/pull/5581.

### DIFF
--- a/pkg/vsphere/disk/disk.go
+++ b/pkg/vsphere/disk/disk.go
@@ -19,6 +19,7 @@ import (
 	"net/url"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	log "github.com/Sirupsen/logrus"
 )
@@ -45,6 +46,43 @@ type Filesystem interface {
 	Unmount(path string) error
 }
 
+// Semaphore represents the number of references to a disk
+type Semaphore struct {
+	resource string
+	refname  string
+	count    uint64
+}
+
+// NewSemaphore creates and returns a Semaphore initialized to 0
+func NewSemaphore(r, n string) *Semaphore {
+	return &Semaphore{
+		resource: r,
+		refname:  n,
+		count:    0,
+	}
+}
+
+// Increment increases the reference count by one
+func (r *Semaphore) Increment() uint64 {
+	return atomic.AddUint64(&r.count, 1)
+}
+
+// Decrement decreases the reference count by one
+func (r *Semaphore) Decrement() uint64 {
+	return atomic.AddUint64(&r.count, ^uint64(0))
+}
+
+// Count returns the current reference count
+func (r *Semaphore) Count() uint64 {
+	return atomic.LoadUint64(&r.count)
+}
+
+// InUseError is returned when a detach is attempted on a disk that is
+// still in use
+type InUseError struct {
+	error
+}
+
 // VirtualDisk represents a VMDK in the datastore, the device node it may be
 // attached at (if it's attached), the mountpoint it is mounted at (if
 // mounted), and other configuration.
@@ -60,9 +98,9 @@ type VirtualDisk struct {
 	// To avoid attach/detach races, this lock serializes operations to the disk.
 	l sync.Mutex
 
-	mountedRefs int
+	mountedRefs *Semaphore
 
-	attachedRefs int
+	attachedRefs *Semaphore
 }
 
 // NewVirtualDisk creates and returns a new VirtualDisk object associated with the
@@ -78,8 +116,11 @@ func NewVirtualDisk(config *VirtualDiskConfig, disks map[uint64]*VirtualDisk) (*
 	}
 	log.Debugf("Didn't find the disk %s in the DiskManager cache, creating it", config.DatastoreURI)
 
+	uri := config.DatastoreURI.String()
 	d := &VirtualDisk{
 		VirtualDiskConfig: config,
+		mountedRefs:       NewSemaphore(uri, "mount"),
+		attachedRefs:      NewSemaphore(uri, "attach"),
 	}
 	disks[config.Hash()] = d
 
@@ -90,8 +131,7 @@ func (d *VirtualDisk) setAttached(devicePath string) (err error) {
 	defer func() {
 		if err == nil {
 			// bump the attached reference count
-			d.attachedRefs++
-			log.Debugf("Increased attach references for %s to %d", d.DatastoreURI, d.attachedRefs)
+			d.attachedRefs.Increment()
 		}
 	}()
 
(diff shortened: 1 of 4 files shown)


pkg/vsphere/disk/disk.go
```diff
@@ -45,6 +45,40 @@ type Filesystem interface {
 	Unmount(path string) error
 }
 
+type ReferenceCount struct {
+	owner      string
+	refname    string
+	references int
```
vburenin: references -> count

pkg/vsphere/disk/disk.go
```diff
@@ -45,6 +45,40 @@ type Filesystem interface {
...
+
+func NewReferenceCount(o, n string) *ReferenceCount {
+	return &ReferenceCount{
+		owner:      o,
+		refname:    n,
+		references: 0,
+	}
+}
+
+func (r *ReferenceCount) Increment() {
```
vburenin: I would consider using atomics in this case you would eliminate a whole class of a potential concurrency problems.

pkg/vsphere/disk/disk.go
```diff
@@ -45,6 +45,40 @@ type Filesystem interface {
...
+	r.references++
+	log.Infof("Increased %s references for %s to %d", r.refname, r.owner, r.references)
```
vburenin: This has to be a debug. However, I would even argue you need a log here at all. If you need to log a number of counts, I would add a String() methods for logging purposes somewhere else and I would log it where it is really necessary.

pkg/vsphere/disk/disk_manager.go
```diff
@@ -142,6 +142,21 @@ func (m *Manager) toSpec(config *VirtualDiskConfig) *types.VirtualDisk {
...
 	defer trace.End(trace.Begin(config.DatastoreURI.String()))
 
+	// reuse the disk if it already exists and is attached
+	if _, err := findDiskByFilename(op, m.vm, config.DatastoreURI.String()); err == nil {
+		// if the disk is attached, it should be in the disk cache
+		d, err := m.Get(op, config)
+		if err != nil {
+			return nil, err
+		}
+		d.l.Lock()
```
vburenin: Using atomics inside ref counters would eliminate a need for these locks.

vburenin: Could you also rename "l" to "mu"? It's just inconsistent.

jzt: The locks aren't used specifically to bump reference counters, only in this case (if I recall correctly). They are used elsewhere while creating, attaching, etc.

pkg/vsphere/disk/disk.go
jzt: See below.. atomics are used in places where the disk needs to be locked, and these counters are wrapped in those locks.

pkg/vsphere/disk/disk.go
jzt: Removed

pkg/vsphere/disk/disk.go
```diff
@@ -45,6 +45,43 @@ type Filesystem interface {
...
+	return &ReferenceCount{
+		owner:   o,
+		refname: n,
+		count:   0,
+	}
+}
+
+// Increment increases the reference count by one
+func (r *ReferenceCount) Increment() {
+	r.count++
```
chengwang86: Do you think if we need to make these operations atomic by adding locks?

pkg/vsphere/disk/disk.go
```diff
@@ -45,6 +45,43 @@ type Filesystem interface {
 	Unmount(path string) error
 }
 
+// ReferenceCount represents the number of references to a disk
+type ReferenceCount struct {
```
hmahmood: I think just `Reference` seems closer to what this is actually.

pkg/vsphere/disk/disk.go
hmahmood: Use the functions in the `sync/atomic` package.

pkg/vsphere/disk/disk_manager_test.go
```diff
@@ -158,6 +158,26 @@ func TestCreateAndDetach(t *testing.T) {
 		if !assert.NoError(t, cerr) {
 			return
 		}
+		refs := child.attachedRefs.Count()
```
hickeng: how about a separate unit test for the reference counting? Use goroutines to get some parallelism.

pkg/vsphere/disk/disk.go
```diff
@@ -45,6 +45,43 @@ type Filesystem interface {
...
+func NewReferenceCount(o, n string) *ReferenceCount {
+	return &ReferenceCount{
+		owner:   o,
+		refname: n,
+		count:   0,
+	}
+}
+
+// Increment increases the reference count by one
+func (r *ReferenceCount) Increment() {
```
hickeng: make this and decrement return the new value - it enables more types of usage.

pkg/vsphere/disk/disk.go
```diff
@@ -45,6 +45,43 @@ type Filesystem interface {
...
+type ReferenceCount struct {
+	owner   string
+	refname string
+	count   int
+}
+
+// NewReferenceCount creates and returns a ReferenceCount initialized to 0
+func NewReferenceCount(o, n string) *ReferenceCount {
+	return &ReferenceCount{
+		owner:   o,
```
hickeng: Owner is confusing when applied to locking structures - this is not the owner of a lock, it's the identifier of the resource.
Different name please.

pkg/vsphere/disk/disk.go
hickeng: This is an unbounded counting semaphore - can see why we're not implementing it using channels, but calling it a semaphore might be less confusing.

pkg/vsphere/disk/disk.go
jzt: There are already locks around this code albeit at a higher level. Are you sure you want to lock within a lock?

pkg/vsphere/disk/disk.go
jzt: Ehh... just `Semaphore`?  Feels weird, but I'll go with it for now unless someone has something more fitting. I agree, though, this is a count of references, not a reference.

pkg/vsphere/disk/disk.go
jzt: Fixed

pkg/vsphere/disk/disk.go
jzt: Fixed?

pkg/vsphere/disk/disk.go
jzt: Fixed

pkg/vsphere/disk/disk_manager.go
jzt: Fixed?

pkg/vsphere/disk/disk.go
jzt: Renamed to `resource`.

pkg/vsphere/disk/disk_manager_test.go
jzt: Added, let me know what you think.

Issue #5426, opened 2017-06-08
Configure level of trust for container-network


hickeng:
**Story**
As a viadmin I know whether a container-network should be trusted or not
As an application author I don't want unnecessary firewall complexity to deal with on trusted networks

**Detail**
From https://github.com/vmware/vic/issues/4872#issuecomment-306972999

The secure-by-default configuration conflicts with ease-of-use and network performance in environments where trust can be assumed. This issue is to provide a means of configuring firewall requirements on a per container-network level.
Initial proposal is to add a `--container-network-firewall=<name>|<level>` option to vic-machine that will apply at `docker create` or `docker network connect` time.
The firewall levels would be along the lines of:
* open - no firewall
* peers - open on all ports to source addresses configured via `--container-network-ip-range`
* exposed - ports defined via `Dockerfile:EXPOSE` or `--expose` will be open for inbound
* published - ports explicitly published via `-p`  or `-P` will be open for inbound
* outbound - outbound connections only (implied for open, exposed, and published)
* blocked - firewall blocks everything by default and the container process manages access via iptables

One alternative would be to allow combination configuration such as `--container-network-firewall=<name>:outbound+published`

**Acceptance**
- [ ] able to connect between containers on the same container network on arbitrary ports
- [ ] <refined acceptance criteria from @malikkal

hickeng:
This is flagged as need info to ensure the solution meets @malikkal's needs before further decomposition.
Added to 1.2 to ensure we deliver the change before the next supported drop.

malikkal:
Thanks @hickeng This would be a great help. 

#4872 4872#issuecomment-307307571

hickeng:
@malikkal The rules are constructed from the cVM configuration on containerVM boot - the primary inputs being the networks the container is connected to and the published ports.

Currently this checks whether the interface is for a container-network - this would switch to checking if the network is trusted, with container networks being trusted by default.

As such the configuration the rules are constructed from are set at `create` and `network connect` time.

hickeng:
Estimate is a coarse 5 - still decisions needed as to how specifically to present this at a `vic-machine` level

willsherwood:
No other flags in `vic-machine` have the format `--flag=value` but rather `--flag value`. Considering that, we might want to change the flag from `--container-network-firewall=<name>|<level>` to `--container-network-firewall <name>|<level>`. 

[...]

PR #5641, opened 2017-07-05
[skip ci] Configure level of trust for container-network (#5426)

willsherwood:
Implements #5426 .

Details:

- Adds a new flag `--container-network-firewall`/`-cnf` to vic-machine's create and configure.
    - open: All ports on containers connected to this network are open.
    - closed: All ports on containers connected to this network are closed.
    - outbound: All incoming traffic is dropped, but all outbound traffic is allowed.
    - peers: Same as outbound, except all ports are open to the ip address range configured by --container-network-ip-range.
    - published: Same as outbound, except ports published with the `-p` flag are allowed for 
inbound.
- vic-machine configure on container networks already set up fails with message "vic-machine configure failed: changes to existing container networks are not supported".
    - This is in line with all other container network operations.
    - Configuring new container networks is allowed (see Ex2).
- This does not handle container network isolation. (see Ex3).

Ex1 (Create)
Create a container network `'out'` that rejects incoming traffic, but can establish connections with other hosts and communicate bidirectionally between them:
```
vic-machine create -cn out -cnf out:outbound
docker -H $DOCKER_HOST run -it --network out busybox
    $ ping 8.8.8.8
    >>> PING 8.8.8.8 (8.8.8.8) 56(84) bytes of data.
        64 bytes from 8.8.8.8: icmp_seq=1 ttl=128 time=11.0ms
    $ nc -l -p 123 # open a server on port 123
    >>> No machine can connect to this endpoint.
``` 

Ex2 (Reconfigure)
Correctly reconfiguring a VCH to add a new container network with published firewall:
```
vic-machine create -cn mynetwork -cnf mynetwork:closed
vic-machine configure -cn anewnetwork -cnf anewnetwork:published
>>> a new container network is created and its firewall configured.
```
Incorrectly reconfiguring a VCH to modify an existing container network:
```
vic-machine create -cn mynetwork -cnf mynetwork:closed
vic-machine configure -cn mynetwork -cnf mynetwork:published
>>> vic-machine configure failed: changes to existing container networks are not supported
```

Ex3 (Network Isolation)
_Vic_: Two distinct external networks can communicate over open ports. (This is up to vSphere)
```vic-machine create --container-network a --container-network b
docker -H $DOCKER_HOST run -it --network a -p 777 busybox
    $ nc -l -p 777 // Open a server listening on port 777
docker -H $DOCKER_HOST run -it --network b busybox
    $ nc $Aipaddress
    >>> connects successfully and can relay messages from both
```
_Vanilla Docker_: Two distinct external networks cannot communicate over open ports.
```
docker run --network a -p 777 busybox
    >>> nc -l -p 777
docker run --network b busybox
    >>> nc $aipaddress
    No connection
```


sgairo:
Make sure to update the copyright on some of these files :) If they have copyright 2016 and you are editing them now, it needs to be adjusted to 2016-2017.

lib/config/executor/network_interface.go
```diff
@@ -74,6 +75,9 @@ type ContainerNetwork struct {
 	// set of network wide links and aliases for this container on this network
 	Aliases []string `vic:"0.1" scope:"hidden" key:"aliases"`
 
+	// Level of trust configured for this network
+	TrustLevel common.TrustLevel
```
hickeng: Please add the base vic version annotation - you shouldn't need any of the others:
`vic:"0.1"`

This will inherit the scope defined for the ContainerNetwork struct which I presume is "read-only" - this value _should_ be read only.

tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.md
```diff
@@ -181,7 +181,20 @@ This test requires that a vSphere server is running and available
 8. Docker container created with network attached with net1, got ip address inside of network range
 9. Docker create another container, and link to previous one, can talk to the the first container successfully
 
-
+## Container Firewalls
+1. Create an open container and verify another open container can connect to it on arbitrary ports.
+2. Create a closed container and verify an open container cannot connect to it on an arbitrary port.
+3. a. Create an outbound container. Verify first that a container on the bridge cannot connect to the
```
hickeng: Adding:
* verify that a container on the bridge network *can* connect to the outbound container via its bridge network address when both are connected to the same bridge network - i.e. bridge networks always have "Peer+Published" trust.
* Verify that an outbound container on the same external network cannot connect to the first outbound container.

cmd/tether/ops_linux.go
```diff
@@ -177,33 +178,85 @@ func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
...
+					sourceAddresses[i] = v.String()
+				}
+				(&netfilter.Rule{
+					Chain:           netfilter.Input,
+					Target:          netfilter.Accept,
+					SourceAddresses: sourceAddresses,
+					Interface:       ifaceName,
+				}).Commit(context.TODO())
+
+			case common.Published:
```
hickeng: With Published, and probably for Peers, we should ensure that ICMP is allowed - it's going to be very common practice to ping these containers.

cmd/tether/ops_linux.go
```diff
@@ -177,33 +178,85 @@ func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
...
-					log.Errorf("can't apply port rule (%s): %s", p, err.Error())
-					continue
+			switch endpoint.Network.TrustLevel {
+			case common.Open:
+				// Accept all incoming and outgoing traffic
+				generalPolicy(netfilter.Accept)
+
+			case common.Closed:
+				// Reject all incoming and outgoing traffic
+				generalPolicy(netfilter.Drop)
```
hickeng: Should only apply to the external interface, not to any bridge network interfaces. Ditto for Open, but that's less problematic.

tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.md
```diff
@@ -181,7 +181,20 @@ This test requires that a vSphere server is running and available
 8. Docker container created with network attached with net1, got ip address inside of network range
 9. Docker create another container, and link to previous one, can talk to the the first container successfully
 
-
+## Container Firewalls
+1. Create an open container and verify another open container can connect to it on arbitrary ports.
+2. Create a closed container and verify an open container cannot connect to it on an arbitrary port.
```
hickeng: Adding:
* Create a Closed container on an external network, and on bridge network - ensure that a container on the same bridge network _can_ connect to the Closed container.

cmd/vic-machine/common/cnetworks.go
```diff
@@ -26,21 +26,50 @@ import (
 	"github.com/vmware/vic/pkg/ip"
 )
 
+type TrustLevel int
+
+const (
+	Open TrustLevel = iota
```
hickeng: I think this will make "Open" the default - can we have Published be the default please.

cmd/tether/ops_linux.go
```diff
@@ -177,33 +178,85 @@ func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
...
-
 			break
 		}
 	}
 
 	return nil
 }
 
+func setupOutboundFirewall(ifaceName string) {
+	netfilter.Policy(context.TODO(), netfilter.Input, netfilter.Drop)
```
hickeng: I'm unsure whether these rules will be appended or inserted and couldn't quickly find whether this is using `iptables-restore` to atomically perform the changed.

There are various operations that cause the tether to reapply it's configuration - the philosophy is "idempotence" so that isn't a problem in general, but I cannot confirm that we don't slowly build up more and more rules. Please run 'exec' against a container and confirm that:
a. the tether.debug shows the rules being checked/applied a second time if needed
b. the iptables state is sane afterwards and we're not appending a new set of duplicate rules each time.

willsherwood: At the top of `SetupFirewall` in `cmd/tether/ops_linux.go`, we flush the tables and set it up again. Line 149: `_ = netfilter.Flush(context.Background(), "")`. Therefore, the idempotence property is upheld, and we won't be appending rules on top of each other continually. 

Could you please clarify under which conditions `exec` would necessarily _need_ to reapply the firewall conditions? Are we allowing some sort of network configuration switching on a running container? @hickeng

cmd/tether/ops_linux.go
willsherwood: Currently the default policy is to drop traffic, and explicit rules need to be created to accept traffic. Should it be that the default policy is to accept, and we should explicitly reject/drop certain traffic instead?

Also, when running a container on an external network, `ifconfig` shows the external interface `eth0` and the loopback, but no bridge interface. Is there a case that a container can be attached to both a bridge and an external network? `docker network connect` does not seem to affect the interfaces either.

cmd/tether/ops_linux.go
willsherwood: Under peers, should only peers be allowed to ping the container? Should ICMP be added to outbound as well?

cmd/tether/ops_linux.go
```diff
@@ -177,33 +178,85 @@ func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
...
+				// Reject all incoming and outgoing traffic
+				generalPolicy(netfilter.Drop)
+
+			case common.Outbound:
+				// Reject all incoming traffic, but allow outgoing
+				setupOutboundFirewall(ifaceName)
+
+			case common.Peers:
+				// Outbound + all ports open to source addresses in --container-network-ip-range
+				setupOutboundFirewall(ifaceName)
```
willsherwood: I was just thinking that it might be useful to make API calls to an external service or something similar. Actually, having published ports might be useful too, in certain circumstances.

cmd/vic-machine/converter/converter.go
```diff
@@ -328,6 +329,19 @@ func convertIP(src reflect.Value, prefix string, tags reflect.StructTag, dest ma
 	return nil
 }
 
+func convertTrustLevel(src reflect.Value, prefix string, tags reflect.StructTag, dest map[string][]string) error {
+	log.Debugf("convertTrustLevel: prefix: %s, src: %s", prefix, src.String())
+	trustLevel, ok := src.Interface().(common.TrustLevel)
+	if !ok {
+		panic(fmt.Sprintf(src.Type().String() + " is not TrustLevel"))
```
willsherwood: Realistically, `ConvertTrustLevel` is only called when the type of a struct member is of type `common.TrustLevel`, so the panic should never be reached in any scenario. If there was some issue, the panic will help to triage the code error similar to the surrounding convert functions.

cmd/tether/ops_linux.go
willsherwood: All network traffic on bridge interfaces are accepted -- implemented.

cmd/tether/ops_linux.go
```diff
@@ -177,33 +178,85 @@ func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
...
+			switch endpoint.Network.TrustLevel {
+			case common.Open:
+				// Accept all incoming and outgoing traffic
+				generalPolicy(netfilter.Accept)
+
+			case common.Closed:
+				// Reject all incoming and outgoing traffic
+				generalPolicy(netfilter.Drop)
+
+			case common.Outbound:
```
willsherwood: Published now also uses outbound as a base firewall, and opens ports on top of it.

cmd/vic-machine/common/cnetworks.go
willsherwood: Published is now the default container network firewall.

cmd/vic-machine/common/cnetworks.go
```diff
@@ -75,6 +104,12 @@ func (c *CNetworks) CNetworkFlags(hidden bool) []cli.Flag {
 			Usage:  "DNS servers for the container network in CONTAINER-NETWORK:DNS format, e.g. vsphere-net:8.8.8.8. Ignored if no static IP assigned.",
 			Hidden: hidden,
 		},
+		cli.StringSliceFlag{
+			Name:   "container-network-firewall, cnf",
+			Value:  &c.ContainerNetworksFirewall,
+			Usage:  "Container network trust level. Options: Open, Closed, Outbound, Peers, Published.",
```
willsherwood: Order changed from least permissive to most permissive.

cmd/vic-machine/common/cnetworks.go
```diff
@@ -26,21 +26,50 @@ import (
...
-	MappedNetworksIPRanges map[string][]ip.Range `cmd:"ip-range" label:"key-value"`
-	MappedNetworksDNS      map[string][]net.IP   `cmd:"dns" label:"key-value"`
+	MappedNetworks          map[string]string     `cmd:"parent" label:"key-value"`
+	MappedNetworksGateways  map[string]net.IPNet  `cmd:"gateway" label:"key-value"`
+	MappedNetworksIPRanges  map[string][]ip.Range `cmd:"ip-range" label:"key-value"`
+	MappedNetworksDNS       map[string][]net.IP   `cmd:"dns" label:"key-value"`
+	MappedNetworksFirewalls map[string]TrustLevel `cmd:"firewall" label:"key-value"`
 }
 
 func (c *ContainerNetworks) IsSet() bool {
```
emlin: this flag is to test if there is any container network options set. Should be changed to include container network firewall configuration as well

cmd/vic-machine/converter/converter.go
emlin: right, this panic is just for someone put this type's converter to convert other type's object.
Any expected error should be returned

lib/config/executor/network_interface.go
emlin: This common.TrustLevel is used in vic-machine, portlayer and tether. Feels like it should be moved to a common package, instead of one in cmd/vic-machine.
Assume these packages are used only in vic-machine  cli.

willsherwood: TrustLevel is now moved to lib/config/executor package.

cmd/vic-machine/common/cnetworks.go
willsherwood: Added a check in IsSet method for added Trust Levels.

cmd/vic-machine/common/cnetworks.go
```diff
@@ -131,6 +141,22 @@ func parseContainerNetworkDNS(cds []string) (map[string][]net.IP, error) {
...
 }
 
+func parseContainerNetworkFirewalls(cfs []string) (map[string]executor.TrustLevel, error) {
+	firewalls := make(map[string]executor.TrustLevel)
+	for _, cf := range cfs {
+		vnet, value, err := splitVnetParam(cf)
+		if err != nil {
+			return nil, fmt.Errorf("Error parsing container network parameter %s: %s", cf, err)
+		}
+		trust, err := executor.ConvertStringToTrustLevel(value)
```
hmahmood: `ParseTrustLevel` would follow other packages (like `net`) that have similar functions.

lib/portlayer/network/context.go
```diff
@@ -1045,6 +1047,13 @@ func (c *Context) AddContainer(h *exec.Handle, options *AddContainerOptions) err
 				return err
 			}
 		}
+		// Check that ports are only opened on published network firewall configuration.
+		if len(options.Ports) > 0 && s.TrustLevel() != executor.Published {
```
hmahmood: What about "opened" or "peers"?

cmd/tether/ops_linux.go
```diff
@@ -177,33 +168,146 @@ func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
...
-				// parse the port maps
-				r, err := portToRule(p)
-				if err != nil {
-					log.Errorf("can't apply port rule (%s): %s", p, err.Error())
-					continue
+			case executor.Peers:
+				// Outbound + all ports open to source addresses in --container-network-ip-range
+				setupOutboundFirewall(ifaceName)
+				sourceAddresses := make([]string, len(endpoint.Network.Pools))
+				for i, v := range endpoint.Network.Pools {
```
hmahmood: Does this work for DHCP?

tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.robot
```diff
@@ -390,6 +390,156 @@ Container network - space in network name valid
...
+    ...  --force=true --bridge-network=bridge --compute-resource=%{TEST_RESOURCE} --no-tlsverify
+    ...  --insecure-registry harbor.ci.drone.local
+    ...  --container-network open-net --container-network-firewall open-net:open
+    ...  --container-network closed-net --container-network-firewall closed-net:closed
+    ...  --container-network outbound-net --container-network-firewall outbound-net:outbound
+    ...  --container-network published-net --container-network-firewall published-net:published
+    ...  --container-network peers-net-1 --container-network-firewall peers-net-1:peers
+    ...  --container-network-ip-range peers-net-1:10.10.10.0/24 --container-network-gateway peers-net-1:10.10.10.1/24
+    ...  --container-network peers-net-2 --container-network-firewall peers-net-2:peers
+    ...  --container-network-ip-range peers-net-2:192.168.0.0/16 --container-network-gateway peers-net-2:192.168.0.1/16
```
hmahmood: What about when `--container-network-firewall` is not specified?

cmd/tether/ops_linux.go
```diff
@@ -146,22 +147,12 @@ func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
...
-		Chain:  netfilter.Input,
-		Target: netfilter.Reject,
-	}
+	generalPolicy(netfilter.Drop)
+	configuredExternal := false
 
 	for _, endpoint := range config.Networks {
-		if endpoint.Network.Type == constants.ExternalScopeType {
-
+		if endpoint.Network.Type == constants.ExternalScopeType && !configuredExternal {
```
willsherwood: Previously, there was a "break" statement at the end of any external scope type. I was just replicating that behavior, but guaranteeing that all of the bridges are set up as well. I will remove the "configuredExternal" flag.

tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.robot
willsherwood: When `--container-network` is specified and `--container-network-firewall` is not specified, it defaults to a published firewall. Upon `vic-machine inspect config`, the flag `--container-network-firewall=xxx:published` is shown, even if the argument was not explicitly given. I guess the line should not be shown at all? @hmahmood

lib/portlayer/network/context.go
willsherwood: Since "Opened" already has all ports open, it doesn't make sense to supply any additional ports to open. And currently, how George initially described `peers`, it does not have the ability to expose ports. 

The network topology would be like a bunch of containers on a peers network doing work and communicating with each other. Then, when they finish work, they send it to a published network over a bridge, which will be able to communicate to the world.

cmd/tether/ops_linux.go
```diff
@@ -177,33 +168,146 @@ func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
...
+
+			case executor.Published:
+				setupPublishedFirewall(endpoint, ifaceName)
+
+			default:
+				log.Warningf("Received invalid firewall configuration %v: defaulting to published.",
+					endpoint.Network.TrustLevel)
+				setupPublishedFirewall(endpoint, ifaceName)
+			}
+		} else if endpoint.Network.Type == constants.BridgeScopeType {
```
willsherwood: Done.

cmd/tether/ops_linux.go
```diff
@@ -177,33 +168,146 @@ func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
...
+		r.Interface = ifaceName
+		_ = r.Commit(context.TODO())
+	}
+}
+
+func generalPolicy(target netfilter.Target) {
+	for _, chain := range []netfilter.Chain{netfilter.Input, netfilter.Output, netfilter.Forward} {
+		err := netfilter.Policy(context.TODO(), chain, target)
+		if err != nil {
+			log.Errorf("Couldn't set policy rule: %v.", err)
```
willsherwood: Error is now returned to caller and correctly returned from SetupFirewall.

cmd/vic-machine/common/cnetworks.go
willsherwood: Done.

tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.robot
hmahmood: Yes you should likely not show that since the user did not explicitly specify that. You should have a test case for this though; it should behave _as if_ published was specified.

lib/portlayer/network/context.go
hmahmood: So if the firewall is "open" and I specify ports that is going to fail? That seems counter intuitive.

willsherwood:
Sometimes calling iptables fails. A retry of sorts will need to be introduced to make sure the firewall is initialized properly. Logs below show what happens when a retry is introduced (Eventually the calls make it through).

```Jul 14 2017 17:40:22.235Z INFO  Execing iptables ["-F"]
Jul 14 2017 17:40:22.249Z WARN  Reaped zombie process PID 277
Jul 14 2017 17:40:22.272Z ERROR iptables error: waitid: no child processes
Jul 14 2017 17:40:22.283Z INFO  Execing iptables ["-P" "INPUT" "DROP"]
Jul 14 2017 17:40:22.305Z INFO  Execing iptables ["-P" "OUTPUT" "DROP"]
Jul 14 2017 17:40:22.348Z INFO  Execing iptables ["-P" "FORWARD" "DROP"]
Jul 14 2017 17:40:22.392Z INFO  Execing iptables ["-A" "INPUT" "-m" "state" "--state" "ESTABLISHED,RELATED" "-i" "eth0" "-j" "ACCEPT"]
Jul 14 2017 17:40:22.405Z INFO  Execing iptables ["-A" "OUTPUT" "-o" "eth0" "-j" "ACCEPT"]
Jul 14 2017 17:40:22.414Z WARN  Reaped zombie process PID 284
Jul 14 2017 17:40:22.414Z ERROR iptables error: wait: no child processes
Jul 14 2017 17:40:22.417Z WARN  Failed. Retrying (Attempt 1).
Jul 14 2017 17:40:22.419Z INFO  Execing iptables ["-F"]
Jul 14 2017 17:40:22.435Z INFO  Execing iptables ["-P" "INPUT" "DROP"]
Jul 14 2017 17:40:22.442Z WARN  Reaped zombie process PID 286
Jul 14 2017 17:40:22.442Z ERROR iptables error: wait: no child processes
Jul 14 2017 17:40:22.450Z WARN  Failed. Retrying (Attempt 2).
Jul 14 2017 17:40:22.455Z INFO  Execing iptables ["-F"]
Jul 14 2017 17:40:22.460Z INFO  Execing iptables ["-P" "INPUT" "DROP"]
Jul 14 2017 17:40:22.464Z INFO  Execing iptables ["-P" "OUTPUT" "DROP"]
Jul 14 2017 17:40:22.470Z INFO  Execing iptables ["-P" "FORWARD" "DROP"]
Jul 14 2017 17:40:22.475Z INFO  Execing iptables ["-A" "INPUT" "-m" "state" "--state" "ESTABLISHED,RELATED" "-i" "eth0" "-j" "ACCEPT"]
Jul 14 2017 17:40:22.482Z INFO  Execing iptables ["-A" "OUTPUT" "-o" "eth0" "-j" "ACCEPT"]
Jul 14 2017 17:40:22.489Z INFO  Execing iptables ["-A" "INPUT" "-p" "icmp" "--icmp-type" "echo-request" "-i" "eth0" "-j" "ACCEPT"]
Jul 14 2017 17:40:22.497Z INFO  Execing iptables ["-A" "OUTPUT" "-p" "icmp" "--icmp-type" "echo-reply" "-o" "eth0" "-j" "ACCEPT"]
Jul 14 2017 17:40:22.506Z INFO  Applying rule for port 123/tcp
Jul 14 2017 17:40:22.508Z INFO  Execing iptables ["-A" "INPUT" "-p" "tcp" "--dport" "123" "-i" "eth0" "-j" "ACCEPT"]```

File: tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.robot (hunk 16 above)
Comment (willsherwood) [inline tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.robot]: 6-09-Inspect Step 12 verifies the correct behavior. If a firewall is not specified on a container network, it acts as if `--container-network-firewall network:published` was specified, but the flag is not shown in vic machine inspect config. This is the expected behavior.

File: cmd/tether/ops_linux.go (hunk 15 above)
Comment (willsherwood) [inline cmd/tether/ops_linux.go]: When `--container-network-firewall net:peers` is specified, but `--container-network-ip-range net:...` is not specified, peers will open all traffic on the network interface.  I believe this should be the expected behavior since we have no way of knowing what ip address containers will receive by DHCP and how to classify an ip address as a peer or not.

File: lib/portlayer/network/context.go (hunk 14 above)
Comment (willsherwood) [inline lib/portlayer/network/context.go]: Now, specified ports only fail if the container network firewall is closed.

File: tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.robot [hunk 20]
```diff
@@ -390,6 +390,156 @@ Container network - space in network name valid
...
+    # Connection should be established on the peer network. Let's try a non-peer now...
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -d --net=peers-net-1 --name p8 ${busybox} nc -l -p 1234
+    Should Be Equal As Integers  ${rc}  0
+
+    ${ip}=  Run  docker %{VCH-PARAMS} inspect --format '{{range .NetworkSettings.Networks}}{{.IPAddress }}{{end}}' p8
+
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run --net=peers-net-2 ${busybox} nc ${ip} 1234
+    Should Not Be Equal As Integers  ${rc}  0
+ 
+    Cleanup VIC Appliance On Test Server
```
hmahmood: This is not going to get called if the test fails, so you'll leak the VCH. You may need to move this to its own robot file so the cleanup happens regardless of test failure.

willsherwood: `Test Teardown  Run Keyword If Test Failed  Cleanup VIC Appliance On Test Server` cleans up the VCH if test fails.

tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.md
```diff
@@ -181,7 +181,23 @@ This test requires that a vSphere server is running and available
...
 9. Docker create another container, and link to previous one, can talk to the the first container successfully
 
-
+## Container Firewalls
+1. Create an open container and verify another open container can connect to it on arbitrary ports.
+2. a. Try to publish a port on a closed firewall and verify an error is received.
+   b. Create a closed container and verify an open container cannot connect to it on an arbitrary port.
+   c. Create a container connected to a bridge and a closed network. Verify that another container connected
+      to the same bridge can connect to the closed container.
+3. a. Create an outbound container. Verify that an outbound container on the same external network cannot
```
hickeng: 3c. verify that two outbound containers on the same external network and on the same bridge network _can_ talk to one another via the bridge (confirms that routing/name resolution are working appropriately). We may need to use the fully qualified `containerA.bridgenet` name if the external network is the default.

tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.robot
mhagen-vmware: but it doesn't cleanup all the portgroups you have created. those are going to leak either way.  I don't see the bridge cleanup in either the positive case or the negative case?

tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.md
willsherwood: This test is now added.

tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.robot
willsherwood: Ports are now cleaned up with a custom tear down.

tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.robot
```diff
@@ -390,6 +401,175 @@ Container network - space in network name valid
...
+    # Connection should be established on the peer network. Let's try a non-peer now...
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -d --net=peers-net-1 --name p8 ${busybox} nc -l -p 1234
+    Should Be Equal As Integers  ${rc}  0
+
+    ${ip}=  Run  docker %{VCH-PARAMS} inspect --format '{{range .NetworkSettings.Networks}}{{.IPAddress }}{{end}}' p8
+
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run --net=peers-net-2 ${busybox} nc ${ip} 1234
+    Should Not Be Equal As Integers  ${rc}  0
+
+    [Teardown]  Cleanup Container Firewall Portgroups
```
mhagen-vmware: I am not sure about layering multiple teardowns like this - have you tested the failure case here? Does it cleanup both the portgroups and the VCH?

Either way, I would rather see a combined cleanup keyword that cleans it all up in both failure and passing cases.  Having the cleanup in multiple places is confusing and unclear.

tests/test-cases/Group6-VIC-Machine/6-09-Inspect.robot
```diff
@@ -119,7 +145,7 @@ Verify inspect output for a full tls VCH
     Should Contain  ${output}  Unable to find valid client certs
     Should Contain  ${output}  DOCKER_CERT_PATH must be provided in environment or certificates specified individually via CLI arguments
 
-    Cleanup VIC Appliance On Test Server
+    [Teardown]  Cleanup Port Groups
```
mhagen-vmware: same comment as above

tests/test-cases/Group6-VIC-Machine/6-07-Create-Network.robot
willsherwood: I changed the name from "Cleanup Container Firewall Portgroups" to "Cleanup Container Firewall Test". I also added a comment to help make the statement clear that it does not execute multiple cleanups, but rather replaces the standard test cleanup unconditionally. The keyword cleans up the VCH and port groups so that under a failed test or passing test, the keyword is executed and resources are closed correctly.

willsherwood:
As seen in https://ci.vcna.io/vmware/vic/12482: 

```
1) docker create -it ...
2) docker network connect ...
3) docker start ...
4) docker exec -d ... ./something
5) docker run ...
```

Sometimes because of the detached exec, step 5 runs before the command specified to step 4 actually even starts running. Any time this pattern is used, the following should replace it:

```
1) docker create ... ./something
2) docker start ...
3) docker run ...
```

Issue #5550, opened 2017-06-21
toolbox directory archive tar and untar should be customizable


dougm:
**User Statement:**

As a consumer of the toolbox, I want to customize how a directory (gzip'd tar stream) is unpacked when transfter to the guest and how it is packed when transferred out of the guest.

**Details:**

The toolbox is primarily concerned with guest RPC protocol handling and provides generic default implementations, most of which can be overridden.  For example, `toolbox.Service.PrimaryIP` can be used to override how the VM IP address is reported.  The toolbox's `archive.unpack` method cannot currently be overridden easily.  VIC may need this to handle the mapping of archive uid/gid to container VM ids, filtering or other customizations. Likewise for the toolbox `archive.pack` method, where we may want to filter outgoing files and provide other options (such as follow symlinks).

**Acceptance Criteria:**
The toolbox `archive.unpack` and `archive.pack` methods can be overridden outside of the package.

Follow on to #4105

dougm:
The default implementation also needs some updates to handle:

* tar.Type(s) other than Dir, Reg, Symlink

* Default directory FileMode (currently set to 0700 by default)

dougm:
Some of the supporting work for this was done in vmware/govmomi#763 (hgfs.FileHandler interface)

dougm:
PR: vmware/govmomi#765

dougm:
Example of where & how to override the toolbox's tar/untar methods:

```diff
modified   lib/tether/toolbox.go
@@ -27,6 +27,7 @@ import (
 	"golang.org/x/crypto/ssh"
 
 	"github.com/vmware/govmomi/toolbox"
+	"github.com/vmware/govmomi/toolbox/hgfs"
 	"github.com/vmware/govmomi/toolbox/vix"
 	"github.com/vmware/vic/cmd/tether/msgs"
 )
@@ -113,6 +114,11 @@ func (t *Toolbox) InContainer() *Toolbox {
 	cmd.Authenticate = t.containerAuthenticate
 	cmd.ProcessStartCommand = t.containerStartCommand
 
+	cmd.FileServer.RegisterFileHandler(hgfs.ArchiveScheme, &hgfs.ArchiveHandler{
+		Read:  nil, // func to read an incoming tar here
+		Write: nil, // func to write an outgoing tar here
+	})
+
 	return t
 }
```

PR #5651, opened 2017-07-06
Implement Import for the Archive interface

### LINKED ISSUE #5649: Implement WriteArchive on the Archiver interface
We will need to implement WriteArchive for offline `docker cp` and potentially `docker commit`. This will involve targeting a storage device and unpacking a tar based on an informed `PathSpec` object. It will include interface implementations on both storage caches and all volume stores if possible. We will need to implement our own handler and pathspec information for what it means to untar and how to process the headers. Much of this work has already been done in a previous pr for ReadArchive by @jzt

matthewavery:
This PR will implement the WriteArchive funtion on the VolumeLookupCache, NameLookupCache, vsphere VolumeStore, and the Image Store. This will be needed for offline docker cp and possibly for docker commit. It will include unit tests for the untar function targetting all of the behaviors that we intend to have based on the design of the pathspec object by @jzt

Fixes #5649


lib/archive/archiver.go
```diff
@@ -63,11 +68,20 @@ type Archiver interface {
...
+	// id - the id for the device that is contained within the store
+	// spec - describes filters on paths found in the data (include, exclude, strip, target)
+	// tarstream - the tar stream that is to be written to the target on the device
+	Import(op trace.Operation, store *url.URL, id string, spec *FilterSpec, tarstream io.ReadCloser) error
 }
 
 // CreateFilterSpec creates a FilterSpec from a supplied map
 func CreateFilterSpec(op trace.Operation, spec map[string]FilterType) (*FilterSpec, error) {
-	var rebase, strip string
+	var rebase, strip, target string
```
jzt: Just realized we can get rid of these guys and just use `fs.RebasePath`, `fs.StripPath`, `fs.TargetPath` once we create the filterspec below. We can check in the switch against those instead of these vars.

lib/archive/archiver.go
```diff
@@ -47,6 +51,7 @@ type FilterSpec struct {
 	Exclusions map[string]struct{}
 	RebasePath string
 	StripPath  string
+	TargetPath string
```
jzt: Did we determine that this was a value-add over just using `Rebasepath`? I still think they're the same behavior.

lib/archive/archiver.go
```diff
@@ -63,11 +68,20 @@ type Archiver interface {
 	// spec - describes filters on paths found in the data (include, exclude, rebase, strip)
 	// data - include file data in the tar archive if true, headers only otherwise
 	Export(op trace.Operation, store *url.URL, id, ancestor string, spec *FilterSpec, data bool) (io.ReadCloser, error)
+
+	// Import will process the input tar stream based on the supplied path spec and write the stream to the
+	// target device.
+	//
+	// store - the device store containing the target device
+	// id - the id for the device that is contained within the store
+	// spec - describes filters on paths found in the data (include, exclude, strip, target)
```
jzt: also `rebase` (probably a copypaste from my previous mistake :))

lib/archive/untar.go
```diff
@@ -0,0 +1,135 @@
...
+// the pathSpec will include the following elements
+// - include : any tar entry that has a path below(after stripping) the include path will be written
+// - strip : The strip string will indicate the
+// - exlude : marks paths that are to be excluded from the write operation
+// - target : marks the the write path that will be tacked onto the "unpackPath". e.g /tmp/unpack + /my/target/path = /tmp/unpack/my/target/path
+func Untar(op trace.Operation, tarStream io.ReadCloser, filter *FilterSpec, unpackPath string) error {
+	tr := tar.NewReader(tarStream)
+
+	strip := filter.StripPath
+	target := filter.TargetPath
```
jzt: Yeah now that I see this in action, it's exactly the same behavior as `RebasePath`. Might as well just use that for consistency/symmetry.

lib/archive/untar.go
```diff
@@ -0,0 +1,135 @@
...
+
+	finalTargetPath := filepath.Join(unpackPath, target)
+	op.Debugf("finalized target path for Tar unpack operation at (%s)", finalTargetPath)
+
+	// process the tarball onto the filesystem
+	for {
+		header, err := tr.Next()
+		if err == io.EOF {
+			// This indicates the end of the archive
+			tarStream.Close()
```
jzt: Should this just happen in a `defer` or at the end of the goroutine outside this loop? Thinking of other error cases, like the one right below this, for example.

jzt: Ignore the goroutine part. :)

lib/portlayer/storage/image_cache.go
```diff
@@ -237,6 +237,19 @@ func (c *NameLookupCache) Export(op trace.Operation, store *url.URL, id, ancesto
 	return c.DataStore.Export(op, store, id, ancestor, spec, data)
 }
 
+func (c *NameLookupCache) Import(op trace.Operation, store *url.URL, diskID string, spec *archive.FilterSpec, tarStream io.ReadCloser) error {
+	_, err := util.ImageStoreName(store)
+	if err != nil {
+		return err
+	}
+
+	if archiver, ok := c.DataStore.(archive.Archiver); ok {
```
jzt: This pattern is no longer necessary, see https://github.com/vmware/vic/blob/master/lib/portlayer/storage/image_cache.go#L236

lib/archive/archiver.go
matthewavery: I think for us it might just be a s/target/rebase

lib/archive/untar.go
matthewavery: Yeah was thinking about this. Definitely would be safer to defer and the just break out of the loop here.

lib/portlayer/storage/image_cache.go
matthewavery: Ah right this is a pointless type conversion.

lib/archive/archiver.go
matthewavery: Removed, using rebase

lib/archive/untar.go
```diff
@@ -0,0 +1,136 @@
...
+)
+
+// Untar untars the given tarstream(if it is a tar stream) on the local filesystem based on the unpackPath in the path spec
+//
+// the pathSpec will include the following elements
+// - include : any tar entry that has a path below(after stripping) the include path will be written
+// - strip : The strip string will indicate the
+// - exlude : marks paths that are to be excluded from the write operation
+// - rebase : marks the the write path that will be tacked onto the "unpackPath". e.g /tmp/unpack + /my/target/path = /tmp/unpack/my/target/path
+func Untar(op trace.Operation, tarStream io.ReadCloser, filter *FilterSpec, unpackPath string) error {
```
jzt: Kinda iffy about the verb `Untar`.. since a tar is a thing, I kinda like `Unpack` or `Extract`? Just a thought :)

lib/archive/untar.go
```diff
@@ -0,0 +1,136 @@
...
+		op.Debugf("Strip path was set to \"\"")
+	}
+
+	if _, err := os.Stat(unpackPath); err != nil {
+		// the target unpack path does not exist. We should not get here.
+		op.Errorf("tar unpack target does not exist (%s)", unpackPath)
+		return err
+	}
+
+	finalTargetPath := filepath.Join(unpackPath, target)
```
jzt: `finalRebasePath`

jzt: Nevermind, I see how you're using target here.

lib/portlayer/storage/vsphere/image.go
```diff
@@ -280,6 +280,7 @@ func (v *ImageStore) WriteImage(op trace.Operation, parent *portlayer.Image, ID
 // spec - describes filters on paths found in the data (include, exclude, rebase, strip)
 // data - set to true to include file data in the tar archive, false to include headers only
 func (v *ImageStore) Export(op trace.Operation, store *url.URL, id, ancestor string, spec *archive.FilterSpec, data bool) (io.ReadCloser, error) {
+	isNotPersistentDisk := false
```
jzt: Double negative, can you just use `persistent := true`?

lib/portlayer/storage/vsphere/image.go
```diff
@@ -325,6 +326,29 @@ func (v *ImageStore) Export(op trace.Operation, store *url.URL, id, ancestor str
 	}, nil
 }
 
+func (v *ImageStore) Import(op trace.Operation, store *url.URL, ID string, spec *archive.FilterSpec, tarStream io.ReadCloser) error {
+	isPersistentDisk := true
```
jzt: same, `persistent := true`

lib/portlayer/storage/vsphere/volume.go
```diff
@@ -247,3 +248,26 @@ func (v *VolumeStore) Export(op trace.Operation, store *url.URL, id, ancestor st
 		clean:      cleanFunc,
 	}, nil
 }
+
+func (v *VolumeStore) Import(op trace.Operation, store *url.URL, id string, spec *archive.FilterSpec, tarstream io.ReadCloser) error {
+	isPersistentDisk := true
```
jzt: `persistent := true`

lib/archive/untar.go
matthewavery: Sure, `unpack` is the verb that doug's guest tools uses as well.

lib/archive/untar.go
matthewavery: Yeah I designed it such that when we changed to using rebase instead of the `TargetString` it was a one line change.

lib/portlayer/storage/vsphere/image.go
matthewavery: sure, I will use `persistent = true` as a constant and use that verb above. When I wrote that I realized how funny it looked after the fact :P

lib/archive/unpack.go
```diff
@@ -0,0 +1,136 @@
...
+// Untar untars the given tarstream(if it is a tar stream) on the local filesystem based on the unpackPath in the path spec
+//
+// the pathSpec will include the following elements
+// - include : any tar entry that has a path below(after stripping) the include path will be written
+// - strip : The strip string will indicate the
+// - exlude : marks paths that are to be excluded from the write operation
+// - rebase : marks the the write path that will be tacked onto the "unpackPath". e.g /tmp/unpack + /my/target/path = /tmp/unpack/my/target/path
+func Unpack(op trace.Operation, tarStream io.ReadCloser, filter *FilterSpec, unpackPath string) error {
+	// the tar stream should be wrapped up at the end of this call
+	defer tarStream.Close()
```
hmahmood: This is a little strange ... even if there is an error the function will close the stream passed in? Should `tarStream` be just a `io.Reader`? Whoever created the `io.ReadCloser` should close it (i.e. release the resource it created).

lib/archive/util.go
```diff
@@ -0,0 +1,86 @@
...
+// limitations under the License.
+
+package archive
+
+import (
+	"archive/tar"
+	"bytes"
+)
+
+// Below are some structs and utility functions for unit testing the untar functionality
```
hmahmood: If these are just for testing, move them to a file that has `_test` suffix.

pkg/vsphere/disk/disk_manager.go
```diff
@@ -443,8 +443,15 @@ func (m *Manager) devicePathByURI(op trace.Operation, datastoreURI *object.Datas
 }
 
 // AttachAndMount creates and attaches a vmdk as a non-persistent disk, mounts it, and returns the mount path.
-func (m *Manager) AttachAndMount(op trace.Operation, datastoreURI *object.DatastorePath) (string, error) {
-	config := NewNonPersistentDisk(datastoreURI)
+func (m *Manager) AttachAndMount(op trace.Operation, datastoreURI *object.DatastorePath, persistent bool) (string, error) {
+	var config *VirtualDiskConfig
+
+	if !persistent {
+		config = NewNonPersistentDisk(datastoreURI)
```
hmahmood: Why is there `NewNonPersistentDisk` and `NewPersistentDisk`, and not just a `persistent` flag that is passed into one function?

lib/archive/util.go
matthewavery: Originally they were in a file that has a `_test` suffix. I intend to remove the code that is there and reference this function in two different test files. I will look to move it to `Unpack_test.go` the other place they are referenced is in image_test.go for the vsphere image_store I believe.

pkg/vsphere/disk/disk_manager.go
matthewavery: This is due to new changes to the diskmanager that @caglar10ur put in a few days ago. Previously I believe that we only supported persistent disks and I am not sure if any other functions call the Persistent disk version. @caglar10ur might know more on this.

lib/archive/unpack.go
matthewavery: I see, yes I think this should be just an `io.reader` and then the stream should be closed at the handler level. Atleast, that makes sense to me since the handler owns the responsibility of returning the response.

lib/archive/unpack.go
```diff
@@ -0,0 +1,135 @@
...
+const (
+	// fileWriteFlags is a collection of flags configuring our writes for general tar behavior
+	//
+	// O_CREATE = Create file if it does not exist
+	// O_TRUNC = truncate file to 0 length if it does exist(overwrite the file)
+	// O_WRONLY = We use this since we do not intend to read, we only need to write.
+	fileWriteFlags = os.O_CREATE | os.O_TRUNC | os.O_WRONLY
+)
+
+// Untar untars the given tarstream(if it is a tar stream) on the local filesystem based on the unpackPath in the path spec
```
mhagen-vmware: Unpack untars...

lib/archive/unpack_test.go
```diff
@@ -0,0 +1,1035 @@
...
+	l2file4 = filepath.Join(path2dir1, "file2")
+
+	// l3 directories
+	path1dir4 = filepath.Join(path1dir3, "path1dir4")
+
+	// l3 files
+	l3file1 = filepath.Join(path1dir3, "file1")
+	l3file2 = filepath.Join(path2dir2, "file1")
+)
+
```
mhagen-vmware: I don't see a test for an invalid tarstream?

lib/archive/unpack.go
matthewavery: ah yes I missed this comment :(

lib/archive/unpack_test.go
matthewavery: like a non tar stream?

mhagen-vmware: What happens if the tarstream is 0 bytes?
What happens if a network error happens and the tarstream gets truncated and is not a valid archive format?

matthewavery: There is handling for that in the `unpack` function. I have added logging and a comment to that section. As well as including a simple invalid stream unit test. I could also devise a truncated tar stream test as well if you think it would add even more value.

matthewavery: I think in the case of the a 0 byte buffer the tar.Reader will actually return with an `EOF` immediately indicating the end of an archive. in which case we will not respond with an error we will simply say that the unpack was successful.

matthewavery: should be all good 👍

lib/archive/unpack.go
```diff
@@ -0,0 +1,137 @@
...
+)
+
+// unpack will unpack the given tarstream(if it is a tar stream) on the local filesystem based on the unpackPath in the path spec
+//
+// the pathSpec will include the following elements
+// - include : any tar entry that has a path below(after stripping) the include path will be written
+// - strip : The strip string will indicate the
+// - exlude : marks paths that are to be excluded from the write operation
+// - rebase : marks the the write path that will be tacked onto the "unpackPath". e.g /tmp/unpack + /my/target/path = /tmp/unpack/my/target/path
+func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, unpackPath string) error {
```
hickeng: We also have archive handling code here: `github.com/vmware/govmomi/toolbox/hgfs/archive.go`

Longer term we should add the FilterSpec to the govmomi toolbox and update `govc` to allow richer glob patterns for copy.
For now we can use this mechanism as suggested by @dougm: https://github.com/vmware/vic/issues/5550#issuecomment-313462597 to provide the tether side filter spec aware implementation.
I suspect there's some integration needed on that path with @morris-jason's PR.

lib/portlayer/storage/vsphere/image.go
```diff
@@ -288,14 +288,14 @@ func (v *ImageStore) Export(op trace.Operation, store *url.URL, id, ancestor str
 	mounts := []*object.DatastorePath{}
 	cleanFunc := func() {
 		for _, mount := range mounts {
-			if err := v.dm.UnmountAndDetach(op, mount); err != nil {
+			if err := v.dm.UnmountAndDetach(op, mount, !persistent); err != nil {
```
hickeng: @caglar10ur what would it take to pass a *VirtualDisk here instead? Doing it this way makes us dependent on knowing the persistence value that was used when mounting, however that might not be as simple as in this scenario.

pkg/vsphere/disk/disk_manager.go
caglar10ur: I - personally - don't like passing flags to functions to change their behavior and also believe this is way more readable to declare the intent hence NewNonPersistentDisk/NewPersistentDisk.

lib/portlayer/storage/vsphere/image.go
caglar10ur: Just change the UnmountAndDetach to accept VirtualDisk and use DiskMode internally in that function to decide what to do https://github.com/vmware/vic/blob/master/pkg/vsphere/disk/config.go#L39

matthewavery: @hickeng do you want that change in this PR? or is this future work?

lib/archive/unpack.go
matthewavery: 👍 Jason has read this commment and will address some of it in his PR I believe. This tar function is designed as a package to be used with our filterspec implementation and expectations.

PR #5652, opened 2017-07-06
Update vendor govmomi / toolbox

### LINKED ISSUE #5550: toolbox directory archive tar and untar should be customizable

dougm:
Fixes #5550

### DIFF
--- a/vendor/github.com/vmware/govmomi/object/network.go
+++ b/vendor/github.com/vmware/govmomi/object/network.go
@@ -20,6 +20,7 @@ import (
 	"context"
 
 	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/types"
 )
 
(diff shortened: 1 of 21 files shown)


vendor/github.com/vmware/govmomi/object/network.go
```diff
@@ -34,12 +35,20 @@ func NewNetwork(c *vim25.Client, ref types.ManagedObjectReference) *Network {
 }
 
 // EthernetCardBackingInfo returns the VirtualDeviceBackingInfo for this Network
-func (n Network) EthernetCardBackingInfo(_ context.Context) (types.BaseVirtualDeviceBackingInfo, error) {
-	name := n.Name()
+func (n Network) EthernetCardBackingInfo(ctx context.Context) (types.BaseVirtualDeviceBackingInfo, error) {
+	var e mo.Network
+
+	// Use Network.Name rather than Common.Name as the latter does not return the complete name if it contains a '/'
+	// We can't use Common.ObjectName here either as we need the ManagedEntity.Name field is not set since mo.Network
```
matthewavery: Thought this might be a chopped comment possibly. Should it possibly read "We can't use Common.ObjectName here either as we need the ManagedEntity.Name field *which* is not set since mo.Network"

vendor/github.com/vmware/govmomi/toolbox/command.go
```diff
@@ -472,19 +490,22 @@ func (c *CommandServer) ListFiles(header vix.CommandRequestHeader, data []byte)
...
 	if err != nil {
 		return nil, err
 	}
 
+	var dir string
 	var files []os.FileInfo
 
 	if info.IsDir() {
+		dir = r.GuestPathName
 		files, err = ioutil.ReadDir(r.GuestPathName)
```
matthewavery: use **dir** instead of r.GuestPathName since it is assigned above possibly.

dougm:
@matthewavery ok thanks, want to submit a govmomi PR with those fixes?  Not required for this vic vendor update, as they are cosmetic, so you could wait to do that.

dougm:
@andrewtchin @caglar10ur @hmahmood need a review for the port-layer-network component (test change).

PR #5664, opened 2017-07-06
Add build number in UI plugin version

### LINKED ISSUE #5663: Add build number in UI plugin version
**User Statement:**
As a developer I would like to display the precise version of the H5C and Flex Client plugins so that troubleshooting UI plugin related issues is made a bit easier. This will also help the UI plugin upgrader script distinguish existing and new plugins better.

**Details:**
The current format of the plugin version is: `$major.$minor.$patch`, which could be improved by appending the build number to it. This format (`$major.$minor.$patch.$build`) is already supported by the H5 Client and Flex Client (see the error messages in #5022), so it should be an easily achievable UX improvement.

**Acceptance Criteria:**
User should be able to see the plugin version in a format such as 1.2.0.12000.

jooskim:
Fixes #5663


jooskim:
Upgrader scripts will show the build number information as follows:
image
image

H5 Client will display the build number for the plugin as follows:
image

PR #5665, opened 2017-07-06
[skip ci] WIP implement offline conatinerStatPath 

jialin-li:
Offline statpath works for volumes(vmdk and nfs volume).
It does not work for r/w layer. Code in lib/portlayer/storage/image.go are likely to change a lot.

### DIFF
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -41,6 +41,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"os"
 	"strconv"
 	"strings"
 	"sync"
@@ -103,6 +104,7 @@ type VicContainerProxy interface {
 
 	ArchiveExportReader(ctx context.Context, store, ancestorStore, deviceID, ancestor string, data bool, filterSpec archive.FilterSpec) (io.ReadCloser, error)
 	ArchiveImportWriter(ctx context.Context, store, deviceID string, filterSpec archive.FilterSpec) (io.WriteCloser, error)
+	StatPath(ctx context.Context, sotre, deviceID, path string) (*types.ContainerPathStat, error)
 
 	Stop(vc *viccontainer.VicContainer, name string, seconds *int, unbound bool) error
 	State(vc *viccontainer.VicContainer) (*types.ContainerState, error)
@@ -153,6 +155,7 @@ const (
 	swaggerSubstringEOF                 = "EOF"
 	forceLogType                        = "json-file" //Use in inspect to allow docker logs to work
 	ShortIDLen                          = 12
+	archiveStreamBufSize                = 64 * 1024
 
 	DriverArgFlagKey      = "flags"
 	DriverArgContainerKey = "container"
@@ -840,6 +843,39 @@ func (c *ContainerProxy) ArchiveImportWriter(ctx context.Context, store, deviceI
 	}()
 
 	return pipeWriter, nil
+}
+
+// StatPath requests the portlayer to stat the filesystem resource at the
+// specified path in the container vc.
+func (c *ContainerProxy) StatPath(ctx context.Context, store, deviceID, path string) (*types.ContainerPathStat, error) {
+	defer trace.End(trace.Begin(deviceID))
+
+	statPathParams := storage.
+	NewStatPathParamsWithContext(ctx).
+		WithStore(store).
+		WithDeviceID(deviceID).
+		WithTargetPath(path)
+	statPathOk, err := c.client.Storage.StatPath(statPathParams)
+	if err != nil {
+		log.Errorf(err.Error())
+		return nil, InternalServerError(err.Error())
+	}
+
+	stat := &types.ContainerPathStat{
+		Name:       statPathOk.Name,
+		Mode:       os.FileMode(statPathOk.Mode),
+		Size:       statPathOk.Size,
+		LinkTarget: statPathOk.LinkTarget,
+	}
+
+	var modTime time.Time
+	if err := modTime.GobDecode([]byte(statPathOk.ModTime)); err != nil {
+		log.Debugf("error getting mod time from statpath: %s", err.Error())
+	} else {
+		stat.Mtime = modTime
+	}
+
+	return stat, nil
 }
 
 // Stop will stop (shutdown) a VIC container.
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -847,14 +847,14 @@ func (c *ContainerProxy) ArchiveImportWriter(ctx context.Context, store, deviceI
 
 // StatPath requests the portlayer to stat the filesystem resource at the
 // specified path in the container vc.
-func (c *ContainerProxy) StatPath(ctx context.Context, store, deviceID, path string) (*types.ContainerPathStat, error) {
+func (c *ContainerProxy) StatPath(ctx context.Context, store, deviceID, spec string) (*types.ContainerPathStat, error) {
 	defer trace.End(trace.Begin(deviceID))
 
 	statPathParams := storage.
 	NewStatPathParamsWithContext(ctx).
 		WithStore(store).
 		WithDeviceID(deviceID).
-		WithTargetPath(path)
+		WithFilterSpec(spec)
 	statPathOk, err := c.client.Storage.StatPath(statPathParams)
 	if err != nil {
 		log.Errorf(err.Error())
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
@@ -586,24 +586,30 @@ func (h *StorageHandlersImpl) StatPath(params storage.StatPathParams) middleware
 	// do offline container stat path, if fails do online
 	op := trace.NewOperation(context.Background(), fmt.Sprintf("StatPath(%s)", params.DeviceID))
 
-	fileStat, err := h.volumeCache.StatPath(op, "", params.DeviceID, params.TargetPath)
-	if err != nil {
-		//fileStat, err = h.imageCache.StatPath(op, params.DeviceID, params.TargetPath)
-		// for debugging only:
-		return storage.
-		NewStatPathOK().
-			WithMode(1).
-			WithLinkTarget("symlink").
-			WithName(err.Error()).
-			WithSize(60)
+	resp, err := h.performOfflineStatPath(op, params)
+	if err == nil {
+		return resp
 	}
 
-	// offline succeeded
+	if spl.IsErrDiskInUse(err) {
+		// do online statpath
+		// return ok for testing now
+		return storage.NewStatPathInternalServerError()
+	}
+
+	return storage.NewStatPathNotFound()
+}
+
+func (h *StorageHandlersImpl) performOfflineStatPath(op trace.Operation, params storage.StatPathParams) (middleware.Responder, error) {
+	spec := vicarchive.DecodeFilterSpec(&params.FilterSpec)
+
+	// try volume store first
+	fileStat, err := h.volumeCache.StatPath(op, params.Store, params.DeviceID, spec.RebasePath)
+	// offline volume succeeded
 	if err == nil {
 		modTimeBytes, err := fileStat.ModTime.GobEncode()
 		if err != nil {
 			op.Debugf("failed to encode modtime from statpath %s", err.Error())
-			return storage.NewStatPathNotFound()
 		}
 
 		return storage.
@@ -612,11 +618,17 @@ func (h *StorageHandlersImpl) StatPath(params storage.StatPathParams) middleware
 			WithLinkTarget(fileStat.LinkTarget).
 			WithName(fileStat.Name).
 			WithSize(fileStat.Size).
-			WithModTime(string(modTimeBytes))
+			WithModTime(string(modTimeBytes)), nil
 	}
 
-	//TODO: if throws disk in use error, proceed to online
-	return storage.NewStatPathNotFound()
+	//offline volume failed and it's not a disk in use error, try r/w layer
+	if !spl.IsErrDiskInUse(err) {
+		//fileStat, err = h.imageCache.StatPath(op, params.DeviceID, params.TargetPath)
+		// for debugging only
+		return storage.NewStatPathNotFound(), err
+	}
+
+	return nil, err
 }
 
 //utility functions
--- a/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
@@ -104,6 +104,7 @@ func (h *StorageHandlersImpl) Configure(api *operations.PortLayerAPI, handlerCtx
 
 	api.StorageExportArchiveHandler = storage.ExportArchiveHandlerFunc(h.ExportArchive)
 	api.StorageImportArchiveHandler = storage.ImportArchiveHandlerFunc(h.ImportArchive)
+	api.StorageStatPathHandler = storage.StatPathHandlerFunc(h.StatPath)
 }
 
 func (h *StorageHandlersImpl) configureVolumeStores(op trace.Operation, handlerCtx *HandlerContext) {
@@ -579,6 +580,43 @@ func (h *StorageHandlersImpl) ImportArchive(params storage.ImportArchiveParams) 
 	log.Infof(buf.String())
 
 	return storage.NewImportArchiveOK()
+}
+
+func (h *StorageHandlersImpl) StatPath(params storage.StatPathParams) middleware.Responder {
+	// do offline container stat path, if fails do online
+	op := trace.NewOperation(context.Background(), fmt.Sprintf("StatPath(%s)", params.DeviceID))
+
+	fileStat, err := h.volumeCache.StatPath(op, "", params.DeviceID, params.TargetPath)
+	if err != nil {
+		//fileStat, err = h.imageCache.StatPath(op, params.DeviceID, params.TargetPath)
+		// for debugging only:
+		return storage.
+		NewStatPathOK().
+			WithMode(1).
+			WithLinkTarget("symlink").
+			WithName(err.Error()).
+			WithSize(60)
+	}
+
+	// offline succeeded
+	if err == nil {
+		modTimeBytes, err := fileStat.ModTime.GobEncode()
+		if err != nil {
+			op.Debugf("failed to encode modtime from statpath %s", err.Error())
+			return storage.NewStatPathNotFound()
+		}
+
+		return storage.
+		NewStatPathOK().
+			WithMode(fileStat.Mode).
+			WithLinkTarget(fileStat.LinkTarget).
+			WithName(fileStat.Name).
+			WithSize(fileStat.Size).
+			WithModTime(string(modTimeBytes))
+	}
+
+	//TODO: if throws disk in use error, proceed to online
+	return storage.NewStatPathNotFound()
 }
 
 //utility functions
--- a/lib/apiservers/portlayer/swagger.json
+++ b/lib/apiservers/portlayer/swagger.json
@@ -360,13 +360,13 @@
 						"type": "string",
 						"x-nullable": false
 					},
-					{
-						"name": "targetPath",
-						"in": "query",
-						"required": true,
-						"type": "string",
-						"x-nullable": false
-					}
+                    {
+                        "name": "filterSpec",
+                        "in": "query",
+                        "required": true,
+                        "type": "string",
+                        "x-nullable": false
+                    }
 				],
 				"responses": {
 					"200": {
@@ -396,21 +396,12 @@
 					},
 					"404": {
 						"description": "supplied target not found",
-						"schema": {
-							"$ref": "#/definitions/Error"
-						}
 					},
 					"500": {
 						"description": "failed to fetch stat path from target",
-						"schema": {
-							"$ref": "#/definitions/Error"
-						}
 					},
 					"default": {
 						"description": "error",
-						"schema": {
-							"$ref": "#/definitions/Error"
-						}
 					}
 				}
 			}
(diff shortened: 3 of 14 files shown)


vendor/github.com/vmware/govmomi/toolbox/command.go
```diff
@@ -472,19 +473,22 @@ func (c *CommandServer) ListFiles(header vix.CommandRequestHeader, data []byte)
 		return nil, err
 	}
 
-	info, err := os.Stat(r.GuestPathName)
+	info, err := os.Lstat(r.GuestPathName)
```
matthewavery: For these kinds of changes we will either need to make a PR request against govmomi or request that doug possibly make them in the govmomi. @dougm thoughts?

pkg/vsphere/disk/disk_manager.go
```diff
@@ -421,3 +423,66 @@ func (m *Manager) devicePathByURI(op trace.Operation, datastoreURI *object.Datas
 
 	return sysPath, nil
 }
+
+// AttachAndMount creates and attaches a vmdk as a non-persistent disk, mounts it, and returns the mount path.
```
matthewavery: Just want to note here for the future. Once #5651 is merged you should be able to rebase on upstrream/master and remove this change :)

lib/portlayer/storage/vsphere/volume.go
```diff
@@ -191,3 +193,22 @@ func (v *VolumeStore) VolumesList(op trace.Operation) ([]*storage.Volume, error)
 
 	return volumes, nil
 }
+
+func (v *VolumeStore) StatPath(op trace.Operation, deviceId string, target string) (stat *compute.FileStat, err error) {
+	diskDsURI := v.volDiskDsURL(deviceId)
+	mountPath, err := v.dm.AttachAndMount(op, diskDsURI, true)
```
matthewavery: we will want a constant for the `true`. #5651 puts this in a common location for this package, so once that is merged we you will be able to use `persistent` and `!persistent` 😄

lib/apiservers/engine/backends/container.go
```diff
@@ -1156,12 +1119,12 @@ func interBridgeTraffic(op portmap.Operation, hostPort, proto, containerAddr, co
 		// chain rather than appended to supersede bridge-to-bridge
 		// traffic blocking
 		baseArgs := []string{"-t", string(iptables.Filter),
-			"-i", bridgeIfaceName,
-			"-o", bridgeIfaceName,
-			"-p", proto,
-			"-d", containerAddr,
-			"--dport", containerPort,
-			"-j", "ACCEPT",
+				     "-i", bridgeIfaceName,
```
matthewavery: I recommend just reverting this change since it is a blank space change set. will make the pr a little easier to read.

matthewavery: or a `checkout HEAD~2 -- lib/apiservers/engine/backends/container.go`

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -507,6 +515,142 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
...
+	}
+
+	//TODO: if throws disk in use error, proceed to online
+
+	// assume it's online container and obj id is container id first.
+	defer trace.End(trace.Begin(params.DeviceID))
+
+	vc := epl.Containers.Container(params.DeviceID)
+	if vc == nil {
+		log.Debugln("%#v", spew.Sdump(epl.Containers))
```
matthewavery: We should be using the `op trace.opertation` here. so `op.Debug()` spew may also be a pretty heavy log here. We might need to find some key information for us to log. maybe the container ID. and the mountspecs possibly.

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -507,6 +515,142 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
...
+			WithMode(fileStat.Mode).
+			WithLinkTarget(fileStat.LinkTarget).
+			WithName(fileStat.Name).
+			WithSize(fileStat.Size)
+	}
+
+	//TODO: if throws disk in use error, proceed to online
+
+	// assume it's online container and obj id is container id first.
+	defer trace.End(trace.Begin(params.DeviceID))
```
matthewavery: Move this to the top of the function.

vendor/github.com/vmware/govmomi/toolbox/command.go
dougm: In general, yes.  But this change (and the others I believe), are from govmomi upstream.  Just needs a rebase (or gvt update).

lib/portlayer/storage/compute/archive.go
```diff
@@ -0,0 +1,103 @@
...
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/vmware/govmomi/guest"
+	"github.com/vmware/govmomi/vim25/soap"
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/lib/portlayer/exec"
+	"github.com/vmware/vic/pkg/trace"
+)
+
+func FileTransferFromGuest(ctx context.Context, vc *exec.Container, path string) (io.Reader, error) {
```
matthewavery: should change from using `ctx context.Context` to `op trace.Operation` and use `op` for the logging.

lib/portlayer/storage/compute/archive.go
```diff
@@ -0,0 +1,103 @@
...
+	defer trace.End(trace.Begin(vc.Config.Name))
+
+	// set up file manager
+	client := vc.VIM25Reference()
+	filemgr, err := guest.NewOperationsManager(client, vc.VMReference()).FileManager(ctx)
+	if err != nil {
+		log.Errorf("OPS MGR ERR: %#v", err)
```
matthewavery: `%s` instead of `%#v` err is usually interpreted either by it's type or by err.Error

lib/portlayer/storage/compute/archive.go
```diff
@@ -0,0 +1,103 @@
...
+	// authenticate client and parse container host/port
+	guestInfo, err := filemgr.InitiateFileTransferFromGuest(ctx, &auth, path)
+	if err != nil {
+		log.Errorf("INIT ERR: %#v --- %#v --- %#v --- %#v", err, ctx, &auth, path)
+		return nil, err
+	}
+
+	url, err := client.ParseURL(guestInfo.Url)
+	if err != nil {
+		log.Errorf("PARSE ERR: %#v", err)
```
matthewavery: same here

lib/portlayer/storage/compute/archive.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil, err
+	}
+	auth := types.NamePasswordAuthentication{
+		Username: vc.ExecConfig.ID,
+	}
+
+	// authenticate client and parse container host/port
+	guestInfo, err := filemgr.InitiateFileTransferFromGuest(ctx, &auth, path)
+	if err != nil {
+		log.Errorf("INIT ERR: %#v --- %#v --- %#v --- %#v", err, ctx, &auth, path)
```
matthewavery: same here

matthewavery: and %s for the `path` var

lib/portlayer/storage/compute/archive.go
```diff
@@ -0,0 +1,103 @@
...
+	if err != nil {
+		log.Errorf("PARSE ERR: %#v", err)
+		return nil, err
+	}
+
+	rc, _, err := client.Download(url, &soap.DefaultDownload)
+	defer rc.Close()
+
+	if err != nil {
+		log.Errorf("DOWNLOAD ERR: %#v", err)
```
matthewavery: same here

lib/portlayer/storage/compute/archive.go
```diff
@@ -0,0 +1,103 @@
...
+
+	url, err := client.ParseURL(guestTransferURL)
+	if err != nil {
+		log.Errorf("PARSE ERR: %#v", err)
+		return err
+	}
+
+	err = client.Upload(reader, url, &soap.DefaultUpload)
+	if err != nil {
+		log.Errorf("UPLOAD ERR: %#v", err)
```
matthewavery: %#v to %s

lib/portlayer/storage/compute/archive.go
```diff
@@ -0,0 +1,103 @@
...
+
+	if err != nil {
+		log.Errorf("DOWNLOAD ERR: %#v", err)
+		return nil, err
+	}
+	log.Debugf("DOWNLOADED %#v", rc)
+	return rc, nil
+}
+
+func FileTransferToGuest(ctx context.Context, vc *exec.Container, path string, reader io.Reader) error {
```
matthewavery: ` ctx context.Context` to `op trace.operation`. then use op for all logging.

lib/portlayer/storage/compute/archive.go
```diff
@@ -0,0 +1,103 @@
...
+		return err
+	}
+	auth := types.NamePasswordAuthentication{
+		Username: vc.ExecConfig.ID,
+	}
+
+	// authenticate client and parse container host/port
+	guestTransferURL, err := filemgr.InitiateFileTransferToGuest(ctx, &auth, path, &types.GuestPosixFileAttributes{}, 0, true)
+	if err != nil {
+		log.Errorf("INIT ERR: %#v", err)
```
matthewavery: %s

lib/portlayer/storage/compute/archive.go
```diff
@@ -0,0 +1,103 @@
...
+		return err
+	}
+
+	url, err := client.ParseURL(guestTransferURL)
+	if err != nil {
+		log.Errorf("PARSE ERR: %#v", err)
```
matthewavery: %s

lib/portlayer/storage/compute/archive.go
```diff
@@ -0,0 +1,103 @@
...
+		return err
+	}
+
+	err = client.Upload(reader, url, &soap.DefaultUpload)
+	if err != nil {
+		log.Errorf("UPLOAD ERR: %#v", err)
+		return err
+	}
+	log.Debugf("UPLOADED %#v", reader)
```
matthewavery: Not sure how useful logging the reader object will be. This would be covered by doing a `trace.End(trace.Begin("%s", vc.ExecConfig.ID))`

lib/portlayer/storage/compute/stat.go
```diff
@@ -0,0 +1,107 @@
...
+	Size       int64
+}
+
+// interface for offline container statpath
+type ContainerStatPath interface {
+	// deviceId, filepath
+	StatPath(op trace.Operation, deviceId string, target string) (*FileStat, error)
+}
+
+func StatPath(ctx context.Context, vc *exec.Container, path string) (*types.GuestFileInfo, error) {
```
matthewavery: op trace.operation instead of the context object. use `op` for logging

lib/portlayer/storage/compute/stat.go
```diff
@@ -0,0 +1,107 @@
...
+	defer trace.End(trace.Begin(vc.Config.Name))
+
+	// below implementation of searching one directory up does not work at root, but we can assume root is a directory
+	if path == "/" {
+		return &types.GuestFileInfo{
+			Path: "/",
+			Type: string(types.GuestFileTypeDirectory),
+			Size: int64(4096),
```
matthewavery: Create a constant for the Magic value. Include a comment discussing what that constant is used for.

lib/portlayer/storage/compute/stat.go
```diff
@@ -0,0 +1,107 @@
...
+	filemgr, err := guest.NewOperationsManager(vc.VIM25Reference(), vc.VMReference()).FileManager(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	auth := types.NamePasswordAuthentication{
+		Username: vc.ExecConfig.ID,
+	}
+
+	// list files one directory up form the target path to include directories.
```
matthewavery: s/form/from

lib/portlayer/storage/compute/stat.go
```diff
@@ -0,0 +1,107 @@
...
+	pathBasename := filepath.Base(path)
+	for _, path := range []string{path, strings.Replace(path, pathBasename, "", -1)} {
```
matthewavery: rather than calling base and then replacing it with `""` you should just use `filepath.Split(path)` This will return the path to the parent of the target and then only the target string. e.g. filepath.Split("/my/example/target") -> "/my/example" and "target"

lib/portlayer/storage/compute/stat.go
```diff
@@ -0,0 +1,107 @@
...
+		var offset int32
+		files, err := filemgr.ListFiles(ctx, &auth, path, offset, 0, "")
```
matthewavery: We might want some constants for the magic values here

lib/portlayer/storage/compute/stat.go
```diff
@@ -0,0 +1,107 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("file listing for container %s failed\n: %s", vc.ExecConfig.ID, err.Error())
+		}
+
+		for _, file := range files.Files {
+			log.Debugf("Stats for file %s --- %s\n", path, file)
```
matthewavery: might want a %#v for the logging of the file struct

lib/portlayer/storage/nfs/volume.go
```diff
@@ -272,3 +274,37 @@ func (v *VolumeStore) getMetadata(op trace.Operation, ID string, target Target)
...
+		result = result + " " + file.Name()
+	}
+
+	fileInfo, _, err := target.Lookup(path.Join(v.volDirPath(deviceId), targetPath))
+	if err != nil {
+		return nil, err
+	}
+
+	// nfs server returns "" for file name when doing a look up, need to fill in the value from targetPath instead
+	// trimsuffix shouldn't be needed since the statpath personality should trim the suffix
```
matthewavery: We still have to handle that case here. The portlayer is it's own entity and other people can call it to. We are not guaranteed that they will also trim their paths.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -272,3 +274,37 @@ func (v *VolumeStore) getMetadata(op trace.Operation, ID string, target Target)
...
+	newPath := targetPath
+	for strings.HasSuffix(newPath, "/") {
```
matthewavery: this could also be `for trimmed := strings.TrimSuffix(newpath, "/"); trimmed != newpath {newpath = trimmed}` 

I am not sure what exactly is being achieved here( I may be missing something jlmk). but I think you could also just do a `_, filename := filepath.Split(targetPath)` if all you want is the name of the file here.

lib/portlayer/storage/nfs/volume.go
```diff
@@ -272,3 +274,37 @@ func (v *VolumeStore) getMetadata(op trace.Operation, ID string, target Target)
...
+		newPath = strings.TrimSuffix(newPath, "/")
+	}
+	// length of pathTokens is guaranteed to be >= 1
+	pathTokens := strings.Split(strings.TrimSuffix(newPath, "/"), "/")
+	fileName := pathTokens[len(pathTokens) - 1]
+
+	// FIXME: cannot read symlink.
```
matthewavery: Talked to George about this. I need to make a ticket for implementing `ReadLink` in the go-nfs-client for this. Most likely during the hardening time if possible.

lib/portlayer/storage/volume_cache.go
```diff
@@ -232,3 +234,22 @@ func volumeInUse(ID string) error {
 
 	return nil
 }
+
+func (v *VolumeLookupCache) StatPath(op trace.Operation, deviceId string, target string) (*compute.FileStat, error) {
+	v.vlcLock.Lock()
+	defer v.vlcLock.Unlock()
+	// FIXME: If we change the structure of volume cache, we can no longer get volume store from a deviceID.
```
matthewavery: Yes in the future this is likely to change to the caller needing to know the store that contains their volume. In that case we will be given the storeID along with the device ID .

lib/portlayer/storage/vsphere/image.go
```diff
@@ -732,3 +736,47 @@ func createBaseStructure(op trace.Operation, vmdisk *disk.VirtualDisk) (err erro
...
+		op.Debugf("Failed to determine host UUID")
+		return nil, err
+	}
+
+	if err = v.verifyImage(op, host, deviceId); err != nil {
+		op.Debugf("Device is not an image")
+		return nil, err
+	}
+
+	// TODO: not sure if these checks are needed
```
matthewavery: Others may chime in ofcourse, But I actually think these checks could be useful. It could really catch an edge case that is occurring that we may not think possible. Such as someone deleting the image store out from under us. Or somehow someone is able to get us to call the imagestore with a non existant image.  I think those could be useful checks @jzt @hickeng might also have an opinion on this as well.

lib/portlayer/storage/vsphere/image.go
```diff
@@ -732,3 +736,47 @@ func createBaseStructure(op trace.Operation, vmdisk *disk.VirtualDisk) (err erro
...
+	//}
+
+	diskDsURI := v.imageDiskDSPath(host, deviceId)
+	mountPath, err := v.dm.AttachAndMount(op, diskDsURI, true)
+	if err != nil {
+		return nil, err
+	}
+
+	defer func() {
+		e1 := v.dm.UnmountAndDetach(op, diskDsURI, true)
```
matthewavery: if you want to assign `err` here in this fashion you will still need to check that error after the defer. Might be better to log the Error here instead of reassigning it. It is entirely possible that this err will be related to a refcount err in which case logging the error would be very helpful to us.

lib/portlayer/storage/nfs/volume.go
jialin-li: I did this to handle filepath like "\vol\test.txt\\\\\" since using filepath.Split would return "" as the filename.

jialin-li: I tried the for loop approach `newpath := "";
	for trimmed := targetPath; trimmed != newpath; trimmed = strings.TrimSuffix(newpath, "/") {
		newpath = trimmed
	}`but thought the original code is more readable.

lib/apiservers/engine/backends/copy.go
```diff
@@ -0,0 +1,284 @@
...
+	vc := cache.ContainerCache().GetContainer(name)
+	if vc == nil {
+		return nil, NotFoundError(name)
+	}
+
+	stat, err := c.containerProxy.StatPath(context.Background(), name, path)
+	if err != nil {
+		return nil, err
+	}
+	log.Debugf("%#v", stat)
```
sflxn: Maybe add some other text so we know context in the logs.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -673,6 +683,207 @@ func (c *ContainerProxy) StreamContainerStats(ctx context.Context, config *conve
 	return nil
 }
 
+// StatPath requests the portlayer to stat the filesystem resource at the
+// specified path in the container vc.
+func (c *ContainerProxy) StatPath(ctx context.Context, name, path string) (*types.ContainerPathStat, error) {
+	defer trace.End(trace.Begin(name))
+
+	vc := cache.ContainerCache().GetContainer(name)
```
sflxn: you should just pass vc in from the backend.  We have a bunch of functions in the proxy that takes vc as an arg.  You will still need name to get passed in as name can be friendly name or alphanumeric ID.

lib/apiservers/portlayer/swagger.json
```diff
@@ -175,6 +176,222 @@
...
+						"in": "query",
+						"required": true,
+						"type": "string",
+						"x-nullable": false
+					}
+				],
+				"responses": {
+					"200": {
+						"description": "OK",
+						"headers": {
```
sflxn: As a rule, we've been defining complex json in the models below in the swagger spec and specifying the schema in the responses.  For example,
```
					"201": {
						"description": "Created",
						"schema": {
							"$ref": "#/definitions/StoreUrl"
						}
					},
```

lib/apiservers/portlayer/swagger.json
```diff
@@ -175,6 +176,222 @@
...
+								"type": "integer",
+								"format": "int64"
+							},
+							"linkTarget": {
+								"type": "string",
+								"x-nullable": false
+							}
+						}
+					},
+					"404": {
```
sflxn: Since this isn't a streaming handler, you can return our model, Error, and then add more context in the errors returned from the portlayer.  For instance,
```
"404" : {
	"description": "Supplied target not found",
	"schema": {
		"$ref": "#/definitions/Error"
	}
},
```

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -507,6 +515,142 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
 	return storage.NewVolumeJoinOK().WithPayload(actualHandle.String())
 }
 
+func (h *StorageHandlersImpl) StatPath(params storage.StatPathParams) middleware.Responder {
+	// do offline container stat path, if fails do online
+	op := trace.NewOperation(context.Background(), fmt.Sprintf("StatPath(%s)", params.DeviceID))
+
+	fileStat, err := h.volumeCache.StatPath(op, *params.Store, params.DeviceID, params.TargetPath)
```
sflxn: I would prefer that we don't overload the cache with an operation like statpath, but I'll defer this to @jzt who will now own storage.  To me, a cache is just a lookup structure.

lib/portlayer/storage/vsphere/image.go
jialin-li: Just updated the defer function to log error instead.

lib/apiservers/portlayer/swagger.json
jialin-li: I don't think we can use schema here because this is a head request and cannot have response body.

jialin-li:
Integrating with Geroge's changes #5700, closing this pr.

PR #5671, opened 2017-07-07
[specific ci=1-05-Docker-Start] Implement basic attach docker start test

### LINKED ISSUE #5659: Fix and move docker start -a test back into CI.
User story: As a customer of VIC, I would like to have confidence that docker start -a works and is tested regularly.

Details: This test if it is re-written to avoid non-halting containers, can be run in the CI.  Potentially just use busbox ls with start -a

Acceptance criteria:  Start with attach test is moved back to CI, not manual tests.

mhagen-vmware:
fixes #5659


tests/test-cases/Group1-Docker-Commands/1-05-Docker-Start.robot
```diff
@@ -120,3 +120,12 @@ Start a container with removed network
     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} start ${container}
     Should Be Equal As Integers  ${rc}  1
     Should Contain  ${output}  network test-network not found
+
+Simple start with attach
+    Run And Return Rc And Output  docker %{VCH-PARAMS} pull ${busybox}
+    ${rc}  ${container}=  Run And Return Rc And Output  docker %{VCH-PARAMS} create ${busybox} ls
```
rogeliosanchez: Since you are collecting a `rc` value here, should you check it before you proceed further?  Just an observation.

mhagen-vmware: naw, I intentionally didn't check it on either command because it is not the purpose of the test.  I only checked the last RC for that reason.

Issue #5654, opened 2017-07-06
Container Cache is not refreshed properly after a hard VCH restart leading to unexpected behavior


matthewavery:
**Deployment details:**

What was the vic-machine create command used to deploy the VCH?

bin/vic-machine-linux create --name='inspect-test' --target=10.192.60.158 --user=Administrator@vsphere.local --image-store=datastore1 --appliance-iso=bin/appliance.iso --bootstrap-iso=bin/bootstrap.iso --password=*<redacted>* --force=true --bridge-network=bridge --public-network=vm-network --compute-resource=cls --timeout 30m -vs *<redacted>*:nfs-store -vs datastore1/test:default --debug 1 --no-tls

**Steps to reproduce:**

1) docker volume create --name test-vol
2) docker create --name inspect-test --it -v test-vol:/mydata busybox
3) docker inspect inspect-test > before_power
4) restart the VCH(power off then on)
5) docker instpect inspect-test > after_power
6) diff before_power after_power
7) see this output:
(left = pre-restart, right = post-restart)
```
33,35c33
<             "Binds": [
<                 "test-vol:/mydata"
<             ],
---
>             "Binds": null,
39c37
<                 "Config": {}
---
>                 "Config": null
44c42
<                 "Name": "no",
---
>                 "Name": "",
53,54c51,52
<             "DnsOptions": [],
<             "DnsSearch": [],
---
>             "DnsOptions": null,
>             "DnsSearch": null,
108,119c106
<         "Mounts": [
<             {
<                 "Type": "volume",
<                 "Name": "test-vol",
<                 "Source": "test-vol",
<                 "Destination": "/mydata",
<                 "Driver": "vsphere",
<                 "Mode": "",
<                 "RW": true,
<                 "Propagation": ""
<             }
<         ],
---
>         "Mounts": null,
134,137c121
<             "Cmd": [
<                 "sh"
<             ],
<             "ArgsEscaped": true,
---
>             "Cmd": null,
139c123
<             "Volumes": {},
---
>             "Volumes": null,
143c127
<             "Labels": {}
---
>             "Labels": null
```

**Actual behavior:**

we see the above differences in an inspect after the restart. ]

**Expected behavior:**

The container inspect output should be the same even after any number of subsequent VCH restarts. 

**Logs:**

[Inspect-Test.zip](https://github.com/vmware/vic/files/1128557/Inspect-Test.zip)

matthewavery:
[...]

PR #5674, opened 2017-07-07
Add more information about the mounts when populating the persona cache

matthewavery:
Based on some logging this information makes it back into the personality cache. Previously we were only returning the name of the volume after the VCH restarts. Now when the container cache in the personality is rehydrated we will also get `rw` and `mountpath` information for containers. This did however expose a bug in our `docker inspect` after a VCH restart which is reference below. This does not fix that bug, but should provide the needed information for the active `docker cp` work after a VCH restart. A separate PR will need to address the persona container cache missing additional information after a restart, as well as the broken `docker inspect` after VCH restart bugs. 

This PR is in partial reference to #5654

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -506,10 +506,13 @@ func convertContainerToContainerInfo(container *exec.ContainerInfo) *models.Cont
 	}
 
 	// Populate volume information
-	for volName := range container.ExecConfig.Mounts {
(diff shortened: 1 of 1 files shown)


matthewavery:
Looked into this some more when talking to @mhagen-vmware and I may need to add additional code and logging to make sure that the hydration of the cache is actually copying the information into the viccontaine struct from a ContainerInfo struct on hydration. On first glance this is not actually happening.

matthewavery:
There will need to be additional changes to populating fields from the actual ContainerInfo struct during the mount processing portion of the cache hydration. I will be Working to add this soon.

PR #5675, opened 2017-07-09
Docker Archive API support in the Persona code

### LINKED ISSUE #385: Docker API: Copy files or folder
Implement the docker wait operation in the engine-api server and integrate with interaction port layer.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#copy-files-or-folders-from-a-container
https://docs.docker.com/engine/reference/commandline/cp/

sflxn:
Added Persona support for docker copy APIs, ContainerExtractToDir
and ContainerArchivePath.  Also added stream support for archive
import/export in the portlayer's REST server.

Resolves #385

### DIFF
--- a/lib/apiservers/engine/backends/archive.go
+++ b/lib/apiservers/engine/backends/archive.go
@@ -1,0 +1,561 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package backends
+
+import (
+	"archive/tar"
+	"bytes"
+	"context"
+	"encoding/gob"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path"
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/tchap/go-patricia/patricia"
+
+	viccontainer "github.com/vmware/vic/lib/apiservers/engine/backends/container"
+	vicarchive "github.com/vmware/vic/lib/archive"
+	"github.com/vmware/vic/pkg/trace"
+
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/pkg/archive"
+)
+
+const (
+	containerStoreName = "container"
+	volumeStoreName    = "volume"
+)
+
+// ContainerArchivePath creates an archive of the filesystem resource at the
+// specified path in the container identified by the given name. Returns a
+// tar archive of the resource and whether it was a directory or a single file.
+func (c *Container) ContainerArchivePath(name string, path string) (content io.ReadCloser, stat *types.ContainerPathStat, err error) {
+	err = fmt.Errorf("%s does not yet implement ContainerArchivePath", ProductName())
+
+	// defer trace.End(trace.Begin(name))
+
+	// vc := cache.ContainerCache().GetContainer(name)
+	// if vc == nil {
+	// 	return nil, nil, NotFoundError(name)
+	// }
+
+	// var reader io.ReadCloser
+
+	// reader, err = c.exportFromContainer(vc, path)
+	// if err != nil && IsResourceInUse(err) {
+	// 	log.Errorf("ContainerArchivePath failed, resource in use: %s", err.Error())
+	// 	err = fmt.Errorf("Resource in use")
+	// }
+	// if err == nil || reader != nil {
+	// 	content = reader
+	// }
+	// stat = nil
+
+	return
+}
+
+func (c *Container) exportFromContainer(vc *viccontainer.VicContainer, path string) (io.ReadCloser, error) {
+	mounts := mountsFromContainer(vc)
+	mounts = append(mounts, types.MountPoint{Destination: "/"})
+	readerMap := NewArchiveStreamReaderMap(mounts)
+
+	readers, err := readerMap.ReadersForSourcePath(c.containerProxy, vc.ContainerID, path)
+	if err != nil {
+		log.Errorf("Errors getting readers for export: %s", err.Error())
+		return nil, err
+	}
+
+	//FIXME: We need a multi reader that can be closed.  MultiReader returns a regular reader
+	log.Infof("Got %d archive readers", len(readers))
+	finalTarReader := io.MultiReader(readers...)
+
+	return ioutil.NopCloser(finalTarReader), nil
+}
+
+// ContainerCopy performs a deprecated operation of archiving the resource at
+// the specified path in the container identified by the given name.
+func (c *Container) ContainerCopy(name string, res string) (io.ReadCloser, error) {
+	return nil, fmt.Errorf("%s does not yet implement ContainerCopy", ProductName())
+}
+
+// ContainerExport writes the contents of the container to the given
+// writer. An error is returned if the container cannot be found.
+func (c *Container) ContainerExport(name string, out io.Writer) error {
+	return fmt.Errorf("%s does not yet implement ContainerExport", ProductName())
+}
+
+// ContainerExtractToDir extracts the given archive to the specified location
+// in the filesystem of the container identified by the given name. The given
+// path must be of a directory in the container. If it is not, the error will
+// be ErrExtractPointNotDirectory. If noOverwriteDirNonDir is true then it will
+// be an error if unpacking the given content would cause an existing directory
+// to be replaced with a non-directory and vice versa.
+func (c *Container) ContainerExtractToDir(name, path string, noOverwriteDirNonDir bool, content io.Reader) error {
+	err := fmt.Errorf("%s does not yet implement ContainerExtractToDir", ProductName())
+
+	// defer trace.End(trace.Begin(name))
+
+	// vc := cache.ContainerCache().GetContainer(name)
+	// if vc == nil {
+	// 	return NotFoundError(name)
+	// }
+
+	// err := c.importToContainer(vc, path, content)
+	// if err != nil && IsResourceInUse(err) {
+	// 	log.Errorf("ContainerExtractToDir failed, resource in use: %s", err.Error())
+
+	// 	err = fmt.Errorf("Resouce in use")
+	// }
+
+	return err
+}
+
+func (c *Container) importToContainer(vc *viccontainer.VicContainer, path string, content io.Reader) error {
+	rawReader, err := archive.DecompressStream(content)
+	if err != nil {
+		log.Errorf("Input tar stream to ContainerExtractToDir not recognized: %s", err.Error())
+		return StreamFormatNotRecognized()
+	}
+	tarReader := tar.NewReader(rawReader)
+
+	mounts := mountsFromContainer(vc)
+	mounts = append(mounts, types.MountPoint{Destination: "/"})
+	writerMap := NewArchiveStreamWriterMap(mounts, path)
+	defer writerMap.Close() // This should shutdown all the stream connections to the portlayer.
+
+	for {
+		header, err := tarReader.Next()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return err
+		}
+
+		// Lookup the writer for that mount prefix
+		writer, err := writerMap.WriterForAsset(c.containerProxy, vc.ContainerID, path, *header)
+		if err != nil {
+			return err
+		}
+
+		var buf bytes.Buffer
+		enc := gob.NewEncoder(&buf)
+		err = enc.Encode(header)
+		if err != nil {
+			log.Errorf("Unable to encode header")
+			return err
+		}
+
+		headerReader := bytes.NewReader(buf.Bytes())
+		_, err = io.Copy(writer, headerReader)
+		if err != nil {
+			log.Errorf("Error while copying tar header: %s", err.Error())
+			return err
+		}
+
+		_, err = io.Copy(writer, tarReader)
+		if err != nil {
+			log.Errorf("Error while copying tar data for %s: %s", header.Name, err.Error())
+			return err
+		}
+	}
+
+	return nil
+}
+
+// ContainerStatPath stats the filesystem resource at the specified path in the
+// container identified by the given name.
+func (c *Container) ContainerStatPath(name string, path string) (stat *types.ContainerPathStat, err error) {
+	defer trace.End(trace.Begin(fmt.Sprintf("** statpath, name=%s, path=%s", name, path)))
+
+	fakeStat := &types.ContainerPathStat{}
+	fakeStat.Mode = os.ModeDir
+
+	return fakeStat, nil
+}
+
+//----------------------------------
+// Docker cp utility
+//----------------------------------
+
+type ArchiveWriter struct {
+	mountPoint types.MountPoint
+	filterSpec vicarchive.FilterSpec
+	writer     io.WriteCloser
+}
+
+type ArchiveReader struct {
+	mountPoint types.MountPoint
+	filterSpec vicarchive.FilterSpec
+	reader     io.ReadCloser
+}
+
+// ArchiveStreamWriterMap maps mount prefix to io.WriteCloser
+type ArchiveStreamWriterMap struct {
+	prefixTrie *patricia.Trie
+}
+
+// ArchiveStreamReaderMap maps mount prefix to io.ReadCloser
+type ArchiveStreamReaderMap struct {
+	prefixTrie *patricia.Trie
+}
+
+// NewArchiveStreamWriterMap creates a new ArchiveStreamWriterMap.  The map contains all information
+// needed to create a writers for every volume mounts for the container.  This includes the root
+// volume of the container.
+//
+// mounts is the mount data from inspect
+// containerDestPath is the destination path in the container
+func NewArchiveStreamWriterMap(mounts []types.MountPoint, containerDestPath string) *ArchiveStreamWriterMap {
+	writerMap := &ArchiveStreamWriterMap{}
+	writerMap.prefixTrie = patricia.NewTrie()
+
+	for _, m := range mounts {
+		aw := ArchiveWriter{
+			mountPoint: m,
+			writer:     nil,
+		}
+
+		// If container destination path is part of this mount point's prefix, we must remove it and
+		// add to the filterspec.  If the container destination path is "/" we do no stripping.
+		//
+		//	e.g. mount A at /mnt/A
+		//
+		//		cp /mnt cid:/mnt
+		//
+		// file data.txt from local /mnt/A/data.txt will come to the persona as A/data.txt.  We must
+		// tell the storage portlayer to remove "A".
+		//
+		//	e.g. mount A at /mnt/A
+		//
+		//		cp / cid:/
+		//
+		// file data.txt from local /mnt/A/data.txt will come to the persona as mnt/A/data.txt.
+		// Here, we must tell the portlayer to remove "mnt/A".  The key to determining whether to
+		// strip "A" or "mnt/A" is based on the container destination path.
+		if containerDestPath != "/" && strings.HasPrefix(aw.mountPoint.Destination, containerDestPath) {
+			aw.filterSpec.StripPath = strings.TrimPrefix(aw.mountPoint.Destination, containerDestPath)
+		}
+
+		writerMap.prefixTrie.Insert(patricia.Prefix(m.Destination), &aw)
+	}
+
+	return writerMap
+}
+
+// NewArchiveStreamReaderMap creates a new ArchiveStreamReaderMap.  After the call, it contains
+// information to create readers for every volume mounts for the container
+//
+// mounts is the mount data from inspect
+func NewArchiveStreamReaderMap(mounts []types.MountPoint) *ArchiveStreamReaderMap {
+	readerMap := &ArchiveStreamReaderMap{}
+	readerMap.prefixTrie = patricia.NewTrie()
+
+	for _, m := range mounts {
+		ar := ArchiveReader{
+			mountPoint: m,
+			reader:     nil,
+		}
+
+		// If the mount point is not the root file system, we must tell the portlayer to rebase the
+		// files in the return tar stream with the mount point path since the volume does not know
+		// the path it is mounted to.  It only knows it's root file system.
+		//
+		//	e.g. mount A at /mnt/A with a file data.txt in A
+		//
+		//	/mnt/A/data.txt		<-- from container point of view
+		//	/data.txt			<-- from volume point of view
+		//
+		// Neither the volume nor the storage portlayer knows about /mnt/A.  The persona must tell
+		// the portlayer to rebase all files from this volume to the /mnt/A/ in the final tar stream.
+		if ar.mountPoint.Destination != "/" {
+			ar.filterSpec.RebasePath = ar.mountPoint.Destination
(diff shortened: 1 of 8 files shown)


lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -921,6 +1090,36 @@ func (c *ContainerProxy) createNewAttachClientWithTimeouts(connectTimeout, respo
...
+	}
+
+	r.Transport = transport
+
+	plClient := client.New(r, nil)
+	r.Consumers["application/octet-stream"] = runtime.ByteStreamConsumer()
+	r.Producers["application/octet-stream"] = runtime.ByteStreamProducer()
+
+	r.Consumers["application/x-tgz"] = runtime.ConsumerFunc(func(rdr io.Reader, data interface{}) error {
+		bsc := runtime.ByteStreamConsumer()
```
jzt: I think this guy is better off created up in `var()` according to @casualjim

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -921,6 +1090,36 @@ func (c *ContainerProxy) createNewAttachClientWithTimeouts(connectTimeout, respo
...
+		gzReader, err := gzip.NewReader(tar.NewReader(rdr))
+		if err != nil {
+			return err
+		}
+		return bsc.Consume(gzReader, data)
+	})
+	r.Consumers["application/x-tar"] = runtime.ConsumerFunc(func(rdr io.Reader, data interface{}) error {
+		bsc := runtime.ByteStreamConsumer()
```
jzt: Same here

lib/apiservers/engine/backends/copy.go
```diff
@@ -0,0 +1,514 @@
...
+}
+
+func (c *Container) offlineContainerExtractToDir(vc *viccontainer.VicContainer, path string, tarReader *tar.Reader) error {
+	mounts := mountsFromContainer(vc)
+	mounts = append(mounts, types.MountPoint{Destination: "/"})
+	writerMap := NewArchiveStreamWriterMap(mounts, path)
+
+	for {
+		header, err := tarReader.Next()
+		if err != nil || err == io.EOF {
```
jzt: Minor: are we not caring about non-EOF errors here? Looks like we break in either case and don't report.

pkg/archive/archive.go
```diff
@@ -0,0 +1,9 @@
+package archive
```
jzt: This can probably be nuked once you rebase since its already in master

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -921,6 +1090,36 @@ func (c *ContainerProxy) createNewAttachClientWithTimeouts(connectTimeout, respo
...
+
+	r.Transport = transport
+
+	plClient := client.New(r, nil)
+	r.Consumers["application/octet-stream"] = runtime.ByteStreamConsumer()
+	r.Producers["application/octet-stream"] = runtime.ByteStreamProducer()
+
+	r.Consumers["application/x-tgz"] = runtime.ConsumerFunc(func(rdr io.Reader, data interface{}) error {
+		bsc := runtime.ByteStreamConsumer()
+		gzReader, err := gzip.NewReader(tar.NewReader(rdr))
```
emlin: Can this gzip reader be optional? From docker commit, I need both gzip format and tar format. If return from here with gzip format, I'll need two more times decompression to get blobsum, tar file size, and diffID, like what imagec did.
Feels like if I can get tar format file, it will be easier

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -507,6 +517,70 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
 	return storage.NewVolumeJoinOK().WithPayload(actualHandle.String())
 }
 
+// ExportArchive creates a tar archive and returns to caller
+func (h *StorageHandlersImpl) ExportArchive(params storage.ExportArchiveParams) middleware.Responder {
+	defer trace.End(trace.Begin(""))
+
+	var filterSpec vicarchive.FilterSpec
```
morris-jason: Can we put this logic in a helper method? I believe there's a filter spec decoder in the archive package on upstream master. @matthewavery

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -507,6 +517,70 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
...
+	detachableOut := NewFlushingReader(pipeReader)
+
+	return NewStreamOutputHandler("ExportArchive").WithPayload(detachableOut, params.DeviceID, nil)
+}
+
+// ImportArchive takes an input tar archive and unpacks to destination
+func (h *StorageHandlersImpl) ImportArchive(params storage.ImportArchiveParams) middleware.Responder {
+	defer trace.End(trace.Begin(""))
+
+	var filterSpec vicarchive.FilterSpec
```
morris-jason: Same here.

lib/apiservers/portlayer/swagger.json
```diff
@@ -175,8 +176,166 @@
...
+              "description": "Exports a tar archive from the target device.",
+				      "summary": "Export tar archive from supplied target",
+				      "tags": [
+					        "storage"
+				      ],
+				      "consumes": [
+					        "application/json"
+				      ],
+				      "produces": [
+					        "application/x-tar"
```
morris-jason: Why not application/octet-stream?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -673,6 +681,167 @@ func (c *ContainerProxy) StreamContainerStats(ctx context.Context, config *conve
...
+		defer close(done)
+
+		params := storage.NewExportArchiveParamsWithContext(ctx).
+			WithStore(store).
+			WithAncestorStore(&ancestorStore).
+			WithDeviceID(deviceID).
+			WithAncestor(&ancestor).
+			WithData(data)
+
+		// Encode the filter spec
```
morris-jason: Helper func for doing this would be better, as it'll be needed in multiple places. Does upstream master have an encode function for filterspec? @matthewavery

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -673,6 +681,167 @@ func (c *ContainerProxy) StreamContainerStats(ctx context.Context, config *conve
...
+			switch err := err.(type) {
+			case *storage.ExportArchiveInternalServerError:
+				plErr := InternalServerError(fmt.Sprintf("Server error from archive reader for device %s", deviceID))
+				log.Errorf(plErr.Error())
+				pipeWriter.CloseWithError(plErr)
+			case *storage.ImportArchiveLocked:
+				plErr := ResourceLockedError(fmt.Sprintf("Resource locked for device %s", deviceID))
+				log.Errorf(plErr.Error())
+				pipeWriter.CloseWithError(plErr)
+			default:
```
morris-jason: There's also a 404 error case that's not accounted for here.

lib/apiservers/portlayer/swagger.json
```diff
@@ -175,8 +176,166 @@
...
+					}
+				],
+				"responses": {
+					"200": {
+						"description": "OK"
+					},
+					"404" : {
+						"description": "Supplied target not found",
+						"schema": {
+							"$ref": "#/definitions/Error"
```
morris-jason: The custom tar transport we're creating in container proxy uses byte stream consumers, but this Error Model doesn't have an interface for producing byte stream. We either need to remove the message body here and produce no payloads in the storage handler, OR we need to add a byte stream producer to the error model.

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
matthewavery: I was mistaken when I commented on that. I was working under the assumption that you were decoding a map[string]string. which is what the create function does.

lib/apiservers/engine/backends/container_proxy.go
matthewavery: It does not, putting it in the archive package would be for the best though 😄 then other people could use it in the future if they need to. This may mean we might want to move the `archive` package back into `/pkg/archive/` so that other people can use that package on custom clients to the portlayer in the future.

lib/apiservers/portlayer/swagger.json
matthewavery: Looks like you are referring to a response here. While we only expect to stream binary **to** the server side. We would want a producer here for this. The global producer set for our swagger spec is `application/json` which should be good for our error model. Which is why this should function well with a defined `produces` section.

morris-jason: Sorry, my comment was meant to be on ExportArchive above, which is set to produce x-tar.

matthewavery: ah gotcha :)

lib/apiservers/engine/backends/copy.go
sflxn: We do, that's why there is err != nil before the eof check.

lib/apiservers/engine/backends/container_proxy.go
sflxn: We can do that in a future round.  This won't be the last PR for this.  This will be one of the last PR to get merged in a chain of merges.

lib/apiservers/engine/backends/container_proxy.go
sflxn: This is based on the returned data mime-type.  There are three consumers defined above for three mime-types.  If you need more decompression, you can take the reader and feed it through another decompression filter.

lib/apiservers/engine/backends/container_proxy.go
sflxn: I looked at the code for ByteStreamConsumer(), it's just returning an interface with a go function.  This should be fine.  I also don't like global vars if I can help it.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -673,6 +681,167 @@ func (c *ContainerProxy) StreamContainerStats(ctx context.Context, config *conve
...
+// ArchiveImportWriter initializes a write stream for a path.  This is usually called
+// for gettine a writer during docker cp TO container.
+func (c *ContainerProxy) ArchiveImportWriter(ctx context.Context, store, deviceID string, filterSpec archive.FilterSpec) (io.WriteCloser, error) {
+	defer trace.End(trace.Begin(deviceID))
+
+	if store == "" || deviceID == "" {
+		return nil, fmt.Errorf("ArchiveImportWriter called with either empty store or deviceID.  This is not allowed!")
+	}
+
+	plClient, transport := c.createNewAttachClientWithTimeouts(attachConnectTimeout, 0, attachAttemptTimeout)
```
morris-jason: why not createGzipTarClient here as well?

lib/apiservers/portlayer/swagger.json
sflxn: For the time being, just return the http status without the modeled errors.  That's another project that I'll add for the future, which we'll use in all of the streaming code.  We currently do not return any modeled errors in any of our streaming code right now.

lib/apiservers/engine/backends/container_proxy.go
sflxn: That handler doesn't return x-tar or gzip.  It just returns json.

lib/apiservers/engine/backends/archive.go
```diff
@@ -0,0 +1,562 @@
...
+	writerMap := NewArchiveStreamWriterMap(mounts, path)
+
+	for {
+		header, err := tarReader.Next()
+		if err != nil || err == io.EOF {
+			break
+		}
+
+		// Lookup the writer for that mount prefix
+		writer, err := writerMap.WriterForAsset(c.containerProxy, vc.ContainerID, path, *header)
```
morris-jason: Can you wrap this in a TarWriter so writing the header is easier to follow and more consistent with our other streaming code?

lib/apiservers/engine/backends/archive.go
```diff
@@ -0,0 +1,562 @@
...
+func (c *Container) ContainerArchivePath(name string, path string) (content io.ReadCloser, stat *types.ContainerPathStat, err error) {
+	defer trace.End(trace.Begin(name))
+
+	vc := cache.ContainerCache().GetContainer(name)
+	if vc == nil {
+		return nil, nil, NotFoundError(name)
+	}
+
+	var reader io.ReadCloser
+	reader, err = c.offlineContainerArchivePath(vc, path)
```
matthewavery: The portlayer is the one that should be handling the offline and online behavior.  For the portlayer if the disk is in use, then we use guest tools. @hickeng amd @caglar10ur are working on an extension to guest tools which involves providing a disk label to guest tools for unpacking.

lib/apiservers/engine/backends/archive.go
```diff
@@ -0,0 +1,562 @@
...
+	}
+
+	rawReader, err := archive.DecompressStream(content)
+	if err != nil {
+		log.Errorf("Input tar stream to ContainerExtractToDir not recognized: %s", err.Error())
+		return StreamFormatNotRecognized()
+	}
+	tarReader := tar.NewReader(rawReader)
+
+	err = c.offlineContainerExtractToDir(vc, path, tarReader)
```
matthewavery: same here, portlayer should take care of this for you.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -921,6 +1090,35 @@ func (c *ContainerProxy) createNewAttachClientWithTimeouts(connectTimeout, respo
...
+	}
+
+	r.Transport = transport
+
+	plClient := client.New(r, nil)
+	bsc := runtime.ByteStreamConsumer()
+	r.Consumers["application/octet-stream"] = bsc
+	r.Producers["application/octet-stream"] = runtime.ByteStreamProducer()
+
+	r.Consumers["application/x-gzip"] = runtime.ConsumerFunc(func(rdr io.Reader, data interface{}) error {
```
morris-jason: Nothing in our swagger spec consumes x-gzip?

lib/apiservers/engine/backends/archive.go
sflxn: Do you not remember our meeting last week?

lib/apiservers/engine/backends/archive.go
sflxn: NO.

lib/apiservers/engine/backends/archive.go
matthewavery: Yes, we agreed that the portlayer would be handling the power side of things. The portlayer will attempt things offline first, then attempt the online. Since the portlayer is closer to the source of truth. There are additions to the guest tools that are still pending for providing a disk label to guest tools for unpacking.

hickeng: In the meeting we decided that the online/offline choice would be determined by a try-it-and-see for offline:
1. try the offline mount - vSphere will reject it if the disk is locked
2. locate the owner of the disk
3. try the online path
4. return sane error if online path fails for some reason.

My intention was that the switch between behavioural paths be done in the PL, below even the point of the storage_handlers, as the personality is further from current state and there's no way to prevent state changes.
Doing this in the port layer allows the various Stores to provide switch logic suited to the specific instance, e.g. NFS volume store might _not_ provide a separate online path given locking isn't a consideration.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -921,6 +1090,35 @@ func (c *ContainerProxy) createNewAttachClientWithTimeouts(connectTimeout, respo
...
+		gzReader, err := gzip.NewReader(tar.NewReader(rdr))
```
morris-jason: I think this should be
```go
gzReader, err := gzip.NewReader(rdr)
if err != nil {
    return err
}
tarReader := tar.NewReader(gzReader)
return bsc.Consume(tarReader, data)
```

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -673,6 +681,167 @@ func (c *ContainerProxy) StreamContainerStats(ctx context.Context, config *conve
...
+// ArchiveExportReader streams a tar archive from the portlayer.  Once the stream is complete,
+// an io.Reader is returned and the caller can use that reader to parse the data.
+func (c *ContainerProxy) ArchiveExportReader(ctx context.Context, store, ancestorStore, deviceID, ancestor string, data bool, filterSpec archive.FilterSpec) (io.ReadCloser, error) {
+	defer trace.End(trace.Begin(deviceID))
+
+	if store == "" || deviceID == "" {
+		return nil, fmt.Errorf("ArchiveExportReader called with either empty store or deviceID.  This is not allowed!")
+	}
+
+	plClient, transport := c.createGzipTarClient(attachConnectTimeout, 0, attachAttemptTimeout)
```
sflxn: Because we need a client that can consume x-tar.  ExportArachive produces x-tar.  Remember the proxy is a consumer.

lib/apiservers/engine/backends/container_proxy.go
sflxn: That might change.

lib/apiservers/engine/backends/container_proxy.go
sflxn: This is a topic that is unclear.  If you try to google mime types for .tgz, you get very differing opinions.  The only one that seems valid is x-gzip, which seems like an overloading of that mime type.  Some of the other mime types that are popular with .tgz do not appear to be standard.

lib/apiservers/engine/backends/archive.go
sflxn: Look, we talked about http status code to use to do this check.  I did not want this in the persona.

lib/apiservers/engine/backends/container_proxy.go
morris-jason: But the ImportArchive  handler consumes x-tar while the attach client has no producer for x-tar. Isn't that an issue?

lib/apiservers/engine/backends/archive.go
```diff
@@ -0,0 +1,563 @@
...
+		log.Errorf("Errors getting readers for export: %s", err.Error())
+		return nil, err
+	}
+
+	log.Infof("Got %d archive readers", len(readers))
+	finalTarReader := io.MultiReader(readers...)
+
+	// var buf bytes.Buffer
+
+	// log.Infof("Copying output from export archive readers")
```
emlin: is this still needed?

lib/apiservers/engine/backends/container_proxy.go
sflxn: The GzipTarClient only implements consumers for x-tar not producers.  You're looking at it in reverse.  The handler is consuming for this call and the proxy is producing.

lib/apiservers/engine/backends/archive.go
```diff
@@ -0,0 +1,553 @@
...
+		// lazy initialize.
+		log.Debugf("Lazily initializing import stream for %s", aw.mountPoint.Destination)
+		var deviceID string
+		if aw.mountPoint.Destination == "/" {
+			// Special case. / refers to container VMDK and not a volume vmdk.
+			deviceID = cid
+		} else {
+			deviceID = aw.mountPoint.Name
+		}
+		streamWriter, err = proxy.ArchiveImportWriter(context.Background(), "container", deviceID, aw.filterSpec)
```
emlin: This method is using "container" as the store. Are they using same store for volume mount and container disk?
btw, can we have constants for all supported store type?

lib/apiservers/engine/backends/archive.go
```diff
@@ -0,0 +1,553 @@
...
+	return streamWriter, nil
+}
+
+// Close visits all the archive writer in the trie and closes the actual io.WritCloser
+func (wm *ArchiveStreamWriterMap) Close() {
+	defer trace.End(trace.Begin(""))
+
+	closeStream := func(prefix patricia.Prefix, item patricia.Item) error {
+		if aw, ok := item.(*ArchiveWriter); ok && aw.writer != nil {
+			aw.writer.Close()
```
emlin: is it necessary to set aw.writer to nil here?
it might help if there is another caller use this map after it's closed

lib/apiservers/engine/backends/archive.go
```diff
@@ -0,0 +1,553 @@
...
+func (rm *ArchiveStreamReaderMap) FindArchiveReaders(containerSourcePath string) ([]ArchiveReader, error) {
+	defer trace.End(trace.Begin(containerSourcePath))
+
+	var nodes []ArchiveReader
+	var startingNode ArchiveReader
+	var err error
+
+	findStartingPrefix := func(prefix patricia.Prefix, item patricia.Item) error {
+		if _, ok := item.(ArchiveReader); !ok {
+			return fmt.Errorf("item not ArchiveWriter")
```
emlin: ArchiveWriter  => ArchiveReader?

lib/apiservers/engine/backends/archive.go
```diff
@@ -0,0 +1,553 @@
...
+		}
+
+		startingNode = item.(ArchiveReader)
+		return nil
+	}
+
+	// go function used later for searching
+	walkPrefixSubtree := func(prefix patricia.Prefix, item patricia.Item) error {
+		if _, ok := item.(ArchiveReader); !ok {
+			return fmt.Errorf("item not ArchiveWriter")
```
emlin: ditto ^^^

lib/apiservers/engine/backends/archive.go
```diff
@@ -0,0 +1,553 @@
...
+		}
+
+		startingNode = item.(ArchiveReader)
+		return nil
+	}
+
+	// go function used later for searching
+	walkPrefixSubtree := func(prefix patricia.Prefix, item patricia.Item) error {
```
emlin: this method looks similar to findStartingPrefix. One suggestion is to combine them, with a switch to differentiate if need to loop all available items.

lib/apiservers/engine/backends/archive.go
sflxn: Yes, it's a temporary solution for this release since we do not cache the volume stores in the persona right now.  sure.

matthewavery: This is the intended use of the `storeID` section when targeting the R/W layer for Import/Export.

lib/apiservers/engine/backends/archive.go
sflxn: good point

lib/apiservers/engine/backends/archive.go
```diff
@@ -0,0 +1,553 @@
...
+	return streamReaders, nil
+}
+
+// Close visits all the archive readers in the trie and closes the actual io.ReadCloser
+func (rm *ArchiveStreamReaderMap) Close() {
+	defer trace.End(trace.Begin(""))
+
+	closeStream := func(prefix patricia.Prefix, item patricia.Item) error {
+		if aw, ok := item.(*ArchiveReader); ok && aw.reader != nil {
+			aw.reader.Close()
```
emlin: similar with writer, is it better to set aw.reader to nil here?

lib/apiservers/engine/backends/archive.go
sflxn: They look the same, but they're not.  one assigns a value.  The other appends data.  These are go functions that operates on the results independently.  I can't know which operation is actually being performed, a VisitPrefixes or VisitSubtree so I have nothing to switch on.

emlin: that's fine

lib/apiservers/engine/backends/archive.go
emlin: but why this is different with line https://github.com/vmware/vic/pull/5675/files#diff-517d0b1e7a3f32a767fbbc140bba71afR520?
thought they should follow same rule to set store name here

PR #5679, opened 2017-07-10
Replace porcelain with ls-files to speed up build

caglar10ur:
Originally from @hickeng


caglar10ur:
@mhagen-vmware don't think it will change anything on CI but for local builds this removes the ~30 sec overhead per make command

PR #5683, opened 2017-07-10
Fix logging for restartable sessions

### LINKED ISSUE #5682: Restart-able sessions are not initializing their Session logs
vic-init relaunched the session when a service fails. Currently we stop logging when that happens. Instead we should keep writing to service's log file as expected.

caglar10ur:
```
... install VCH ...
...   vic-ssh   ...
root@ [ ~ ]# ps aux | grep port-layer
root       278  4.2  2.0 250696 42808 ?        Sl   22:41   0:00 /sbin/port-layer-server --host=localhost --port=2377
root@ [ ~ ]# kill -9 278
root@ [ ~ ]# ps aux | grep port-layer
root       291  1.0  1.8 238932 38344 ?        Sl   22:41   0:00 /sbin/docker-engine-server -port=2375 -port-layer-port=2377
root@ [ ~ ]# grep Serving /var/log/vic/port-layer.log
Jul 10 2017 22:41:17.337Z INFO  Serving port layer at http://127.0.0.1:2377
Jul 10 2017 22:41:46.726Z INFO  Serving port layer at http://127.0.0.1:2377
root@ [ ~ ]#
```

Fixes #5682

### DIFF
--- a/lib/tether/tether.go
+++ b/lib/tether/tether.go
@@ -294,67 +294,25 @@ func (t *tether) initializeSessions() error {
 		"Execs":    t.config.Execs,
 	}
 
-	var writer syslog.Writer
-
 	// we need to iterate over both sessions and execs
 	for name, m := range maps {
-
 		// Iterate over the Sessions and initialize them if needed
 		for id, session := range m {
-			// make it a func so that we can use defer
-			err := func() error {
-				session.Lock()
-				defer session.Unlock()
+			log.Debugf("Initializing session %s", id)
 
-				if session.wait != nil {
-					log.Warnf("Session %s already initialized", id)
-					return nil
-				}
-				log.Debugf("Initializing session %s", id)
-
+			session.Lock()
+			if session.wait != nil {
+				log.Warnf("Session %s already initialized", id)
+			} else {
 				if session.RunBlock {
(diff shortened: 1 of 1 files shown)


lib/tether/tether.go
```diff
@@ -649,6 +607,42 @@ func (t *tether) handleSessionExit(session *SessionConfig) {
...
+	}
+	session.Outwriter = stdout
+	session.Errwriter = stderr
+	session.Reader = dio.MultiReader()
+
+	if session.Diagnostics.SysLogConfig != nil {
+		cfg := session.Diagnostics.SysLogConfig
+		var w syslog.Writer
+		if writer == nil {
+			writer, err = syslog.Dial(cfg.Network, cfg.RAddr, syslog.Info|syslog.Daemon, fmt.Sprintf("%s", t.config.ID[:shortLen]))
```
hmahmood: Hm... this is going to create one syslog connection per session, while before the syslog connection was being shared between all the sessions. Any way to keep that behavior?

caglar10ur: Grrr I was planning to move  var writer syslog.Writer into tether struct to keep that behavior but looks like I forgot that, changing now

caglar10ur: Done, pushed a new commit to do that

Issue #5538, opened 2017-06-20
nightly test 21-1-whitelist unable to confirm secure registry


chengwang86:
Seen on vsphere 6.5 

[21-1-Whitelist.zip](https://github.com/vmware/vic/files/1089156/21-1-Whitelist.zip)

also 6.0
[21-1-Whitelist.zip](https://github.com/vmware/vic/files/1089167/21-1-Whitelist.zip)

From vic-machine log
```
Jun 20 2017 08:26:17.631Z DEBUG [ END ] [github.com/vmware/vic/pkg/fetcher.(*URLFetcher).Head:361] [606.93µs] http://harbor.ci.drone.local/v2/
Jun 20 2017 08:26:17.631Z WARN  Unable to confirm insecure registry harbor.ci.drone.local is a valid registry at this time.
Jun 20 2017 08:26:17.631Z DEBUG URL: 10.160.126.191/v2/
Jun 20 2017 08:26:17.631Z DEBUG [BEGIN] [github.com/vmware/vic/pkg/fetcher.(*URLFetcher).Head:361] 10.160.126.191/v2/
Jun 20 2017 08:26:17.631Z DEBUG [ END ] [github.com/vmware/vic/pkg/fetcher.(*URLFetcher).Head:361] [40.38µs] 10.160.126.191/v2/
Jun 20 2017 08:26:17.631Z WARN  Unable to confirm secure registry 10.160.126.191 is a valid registry at this time.
Jun 20 2017 08:26:17.631Z INFO  Insecure registries = harbor.ci.drone.local
Jun 20 2017 08:26:17.631Z INFO  Whitelist registries = 10.160.126.191, harbor.ci.drone.local
```

corrieb:
This is continuing to fail:

```Jun 22 2017 19:04:26.025Z WARN Unable to confirm insecure registry harbor.ci.drone.local is a valid registry at this time.```

corrieb:
Still failing on 6/26

hickeng:
I would not expect `harbor.ci.drone.local` to resolve in the nightly environment - i've opened #5598 to address the fact it's present in the nightlies at all.
Looking at the drone output the IP we're trying to connect to (`10.192.180.174` in the 21-1-Whitelist bundle for 6.0) appears to be correct and the page pulled down by curl during the harbor deployment looks correct.

[...]

PR #5684, opened 2017-07-10
Fix whitelist matching in persona & vic-machine & whitelist test

### LINKED ISSUE #5622: Cannot login to an insecure-registry with http://
**VIC version:**

vic 1.1 (actually based on binaries compiled from this branch https://github.com/sflxn/vic/tree/docker-push)

**Deployment details:**
```
vic-machine-linux create --force --name=vch-1 --insecure-registry http://kang.eng.vmware.com --target=root:pwd@192.168.213.129 --debug 1 --image-store=datastore1 --no-tlsverify --no-tls
```

**Steps to reproduce:**
after deploying the vch, run `sudo docker -H ... login http://kang.eng.vmware.com` or `sudo docker -H ... kang.eng.vmware.com`.  Here `kang.vmware.com` is a harbor instance deployed on a Ubuntu desktop VM.

**Actual behavior:**
Both return the same error msg: `rror response from daemon: Head https://kang.eng.vmware.com/v2/: dial tcp 10.17.109.201:443: getsockopt: connection refused`. However, both cmds suceeed on vanilla docker. 

**Expected behavior:**
`sudo docker login kang.eng.vmware.com` succeeds.

**Logs:**
[container-logs.zip](https://github.com/vmware/vic/files/1112605/container-logs.zip)

gigawhitlocks:
Fixes a whitelist matching issue that was causing the whitelist nightly to fail ( #5538 ). First, `vic-machine` required a certain format in order for verification to function, and then secondly when using `docker` the user had to match the URL exactly, regardless as to whether the e.g. port/scheme were implicitly defined via the context.

This test was failing for 2 reasons. First, a bare IP was provided as a secure whitelisted registry. This should be allowed as secure implies https and port 443, so the user should not need to specify. This, however, was a format got eaten by the URL parser during validation.

Simply assuming the protocol fixed this issue, but then the whitelisted URL would need to be specified on the command line when using docker or matching would fail.

I've implemented an approach to fill the whitelist with the handful of representations a URL might take on when requested for whitelisting on the vic-machine commandline or referred to on the docker commandline. This way, the user doesn't need to be more explicit than necessary & whitelisting will work the way (s)he expects, rather than in the brittle manner that was previously occurring. I also suspect this will resolve #5622 although I need to do some testing to ensure. In any case, I believe fixing this bug was at a minimum a dependency of #5622

### DIFF
--- a/lib/install/validate/validator_test.go
+++ b/lib/install/validate/validator_test.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"sort"
 	"strings"
 	"testing"
 
@@ -36,6 +37,7 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/lib/install/data"
+	"github.com/vmware/vic/pkg/registry"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/session"
 )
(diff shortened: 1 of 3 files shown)


lib/install/validate/validator_test.go
```diff
@@ -117,6 +119,72 @@ func TestParseURL(t *testing.T) {
...
+			"https://foo.local",
+			"https://foo.local:443",
+			"foo.local:443",
+		},
+
+		"foo.local:80": {
+			"http://foo.local",
+			"http://foo.local:80",
+			"foo.local:80",
+		},
```
mhagen-vmware: http://foo.local?
http://foo.local:80?

lib/install/validate/validator_test.go
```diff
@@ -117,6 +119,72 @@ func TestParseURL(t *testing.T) {
...
+		}
+		for _, entry := range entries {
+			outputs = append(outputs, entry.String())
+		}
+
+		return
+	}
+
+	testCases := map[string][]string{
+		"foo.local": {
```
mhagen-vmware: So this still has me concerned... this is essentially saying that we expect if someone supplies foo.local then we assume and only assume https?  If a user puts in foo.local, we can't really assume anything right?  We need to support http and https if a user supplies only a fqdn without a scheme or port right?  Should we not try all combinations?

PR #5693, opened 2017-07-11
[skip ci] NFS Shared Volumes Integration Tests

### LINKED ISSUE #3813: Integration Infrastructure and Testing for NFS Shared Volumes
User statement: As a customer of VIC, I want to be assured that the new NFS shared volume support is well tested so that when I use it then it will work as expected.

There will need to be some additions to current CI capabilities in order to test NFS shared volumes. We will need an available NFS server to mark as a nfs target. Whether this is an additional VCH that creates a nfs server container or we have a constant nfs server that must be maintained remains to be discussed. Additionally, if it is a containerized server we must contend with how we want to configure the server after it's creation. There are also networking concerns that go along with this piece since the nfs target must be reachable by the containers. 

We already have an NFS datastore test, so potentially we could just modify that test and install directly into the NFS server provided in that case.

Acceptance criteria:
1. New test is documented
2. New test is implemented and included in either CI or manual test area.

rogeliosanchez:
Simple integration tests done in the manual tests section.
Some tests were reused from the regular volumes tests Group 1.
These will run with the nightlies and the result of the tests will appear in the nightly email summary.

Fixes #3813

### DIFF
--- a/tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
+++ b/tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
@@ -1,0 +1,23 @@
+Test 5-22 - NFS Volume
+=======
+
+# Purpose:
+To verify that NFS shared volumes work with current supported docker commands
+
+# References:
+[1 - Best practices for running VMware vSphere on NFS](http://www.vmware.com/content/dam/digitalmarketing/vmware/en/pdf/techpaper/vmware-nfs-bestpractices-white-paper-en.pdf)
+
+# Environment:
+This test requires access to VMware Nimbus cluster for dynamic ESXi and vCenter creation
+
+# Test Steps:
+1. Deploy a new vCenter with a simple cluster
+2. Deploy an NFS server
+3. Create a new datastore out of a NFS share on the NFS server
+4. Install the VIC appliance in cluster
+
+
+# Expected Outcome:
+
+# Possible Problems:
+None
--- a/tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
+++ b/tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
@@ -2,22 +2,82 @@ Test 5-22 - NFS Volume
 =======
 
 # Purpose:
-To verify that NFS shared volumes work with current supported docker commands
+To verify that NFS shared volumes work with currently supported docker commands
 
 # References:
 [1 - Best practices for running VMware vSphere on NFS](http://www.vmware.com/content/dam/digitalmarketing/vmware/en/pdf/techpaper/vmware-nfs-bestpractices-white-paper-en.pdf)
 
+[2 - Docker Command Line Reference - Volume Create](https://docs.docker.com/engine/reference/commandline/volume_create/)
+
+[3 - Docker Command Line Reference - Exec](https://docs.docker.com/engine/reference/commandline/exec/)
+
+[4 - Docker Command Line Reference - Volume Inspect](https://docs.docker.com/engine/reference/commandline/volume_inspect/)
+
+[5 - Docker Command Line Reference](https://docs.docker.com/engine/reference/commandline/volume_ls/)
+
+
 # Environment:
-This test requires access to VMware Nimbus cluster for dynamic ESXi and vCenter creation
+This test requires access to VMware Nimbus cluster for dynamic ESXi, vCenter, and NFS server creation
 
 # Test Steps:
-1. Deploy a new vCenter with a simple cluster
-2. Deploy an NFS server
-3. Create a new datastore out of a NFS share on the NFS server
-4. Install the VIC appliance in cluster
+1. Deploy VIC appliance to vCenter with a simple cluster and one NFS server; nfs share point added here.
+2. Issue docker volume create using no name volume (unnamed) on VolumeStore=nfsVolumeStoreDefault (NFS volume)
+3. Run docker mount on container using unnamed volume, run docker wait, then docker rm on temp container
+4. Issue docker volume create --name=nfs_default_%{VCH-NAME} on VolumeStore=nfsVolumeStoreDefault (NFS volume)
+5. Run docker mount on container using named volume, run docker wait, then docker rm on temp container
+6. Issue docker volume create on unnamed volume
+7. Issue docker volume create on named volume
+8. Issue docker volume create --name=test???
+9. Create container ${createFileContainer} using named nfs volume
+10. Issue docker exec -i ${createFileContainer} echo # --> to write contents to a file (created by this echo command) on NFS volume
+11. Issue docker exec -i ${createFileContainer} ls   # to verify file is created in the correct directory on NFS volume
+12. Issue docker exec -i ${createFileContainer} cat  # to verify contents of the file
+13. Create container ${addToFileContainer} using named nfs volume
+14. Issue docker exec -i ${addToFileContainer} echo # to append to contents of the file used by earlier container
+15. Issue docker exec -i ${createFileContainer} echo # to append to the same file
+16. Issue docker exec -i ${addToFileContainer} cat # verify contents of the file
+17. Issue docker exec -i ${addToFileContainer} echo # write to the file some more
+18. Issue docker exec -i ${createFileContainer} cat # verify contents of the file
+19. Issue docker volume inspect ${nfsNamedVolume}
+20. Issue docker volume ls
+21. Issue docker volume rm ${nfsDefaultVolume}
+22. Issue docker volume rm ${nfsNamedVolume}
 
 
 # Expected Outcome:
+* Step 2 should complete successfully and return a long string name for the volume created
+* Step 3 should verify that the NFS volume is mounted on a temp container; we then rm this container for cleanup
+* Step 4 should complete successfully and return named volume
+* Step 5 same as step 3 but using the named volume instead
+* Step 6 should result in error with the following error message:
+```
+Error response from daemon: A volume named ${nfsDefaultVolume} already exists. Choose a different volume name.
+```
+* Step 7 should result in error with the following error message:
+ ```
+ Error response from daemon: A volume named ${nfsNamedVolume} already exists. Choose a different volume name.
+ ```
+* Step 8 should result in error with the following message:
+```
+Error response from daemon: create test???: "test???" includes invalid characters for a local volume name, only "\[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed
+```
+* Step 9 should complete successfully creating ${createFileContainer} using ${nfsNamedVolume} (the named nfs volume from before)
+* Step 10 should result in success and create/write contents to the /mydata/test_nfs_file.txt file using the ${createFileContainer}
+* Step 11 should result in success and verify that the 'test_nfs_file.txt' file is in the mydata directory using the ${createFileContainer}
+* Step 12 should result in success and verify the contents of the 'test_nfs_file.txt' file using the ${createFileContainer}
+* Step 13 should complete successfully creating ${addToFileContainer} using ${nfsNamedVolume} (the named nfs volume from before)
+* Step 14 should result in success and append a line to the end of the 'test_nfs_file.txt' using ${addToFileContainer}
+* Step 15 should result in success and append a line to the end of the 'test_nfs_file.txt' using ${createFileContainer}
+* Step 16 should result in success and verify the contents of the 'test_nfs_file.txt' file using the ${addToFileContainer}
+* Step 17 same as Step 14
+* Step 18 same as Step 12
+* Step 19 should result in a properly formatted JSON response
+* Step 20 should result in each nfs volume being listed with both driver and volume name
+* Step 21 should result in success and the volume should not be listed anymore
+* Step 22 should result in error with the following message:  
+```
+Error response from daemon: volume ${nfsNamedVolume} in use by
+```
 
 # Possible Problems:
-None
+Mount command may be affected by Nimbus's performance returning '' when the volume was successfully created/mounted
--- a/tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
+++ b/tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
@@ -17,67 +17,89 @@ To verify that NFS shared volumes work with currently supported docker commands
 
 
 # Environment:
-This test requires access to VMware Nimbus cluster for dynamic ESXi, vCenter, and NFS server creation
+This test requires access to VMware Nimbus for dynamic ESXi and NFS server creation
 
 # Test Steps:
-1. Deploy VIC appliance to vCenter with a simple cluster and one NFS server; nfs share point added here.
-2. Issue docker volume create using no name volume (unnamed) on VolumeStore=nfsVolumeStoreDefault (NFS volume)
-3. Run docker mount on container using unnamed volume, run docker wait, then docker rm on temp container
-4. Issue docker volume create --name=nfs_default_%{VCH-NAME} on VolumeStore=nfsVolumeStoreDefault (NFS volume)
-5. Run docker mount on container using named volume, run docker wait, then docker rm on temp container
-6. Issue docker volume create on unnamed volume
-7. Issue docker volume create on named volume
-8. Issue docker volume create --name=test???
-9. Create container ${createFileContainer} using named nfs volume
-10. Issue docker exec -i ${createFileContainer} echo # --> to write contents to a file (created by this echo command) on NFS volume
-11. Issue docker exec -i ${createFileContainer} ls   # to verify file is created in the correct directory on NFS volume
-12. Issue docker exec -i ${createFileContainer} cat  # to verify contents of the file
-13. Create container ${addToFileContainer} using named nfs volume
-14. Issue docker exec -i ${addToFileContainer} echo # to append to contents of the file used by earlier container
-15. Issue docker exec -i ${createFileContainer} echo # to append to the same file
-16. Issue docker exec -i ${addToFileContainer} cat # verify contents of the file
-17. Issue docker exec -i ${addToFileContainer} echo # write to the file some more
-18. Issue docker exec -i ${createFileContainer} cat # verify contents of the file
-19. Issue docker volume inspect ${nfsNamedVolume}
-20. Issue docker volume ls
-21. Issue docker volume rm ${nfsDefaultVolume}
-22. Issue docker volume rm ${nfsNamedVolume}
+1. Deploy VIC appliance to an ESX and use fake NFS mount point
+2. Deploy VIC appliance to an ESX and use valid NFS mount point
+3. Issue docker volume create using no name for the volume (unnamed) on VolumeStore=nfsVolumeStore (NFS volume)
+4. Run docker mount on container using unnamed volume, run docker wait, then docker rm on temp container
+5. Issue docker volume create --name=nfs_default_%{VCH-NAME} on VolumeStore=nfsVolumeStore (NFS volume)
+6. Run docker mount on container using named volume, run docker wait, then docker rm on temp container
+7. Issue docker volume create on unnamed volume
+8. Issue docker volume create on named volume
+9. Issue docker volume create --name="test!@\#$%^&*()"
+10. Create container ${createFileContainer} using named nfs volume
+11. Issue docker exec -i ${createFileContainer} echo # --> to write contents to a file (created by this echo command) on NFS volume
+12. Issue docker exec -i ${createFileContainer} ls   # to verify file is created in the correct directory on NFS volume
+13. Issue docker exec -i ${createFileContainer} cat  # to verify contents of the file
+14. Create a container using named nfs volume and echo # to append more contents to the file used by earlier container
+15. Create a container using named nfs volume and echo # to append to the same file
+16. Create a container using named nfs volume and echo # to append to the same file
+17. Create a container using named nfs volume and cat # verify contents of the file
+18. Create a detached container using named nfs volume using named nfs volume to cat the file from last test
+19. Issue docker logs to see the results of the cat command # verify contents of the file
+20. Create a container using named nfs volume and rm the file just used.
+21. Create a container using named nfs volume and cat the file that was just removed
+22. Issue docker start on detached container from earlier
+23. Issue docker logs on detached container
+24. Spin up on container per item in a list to write once a sec to a file the value passed in from the list and save the container ids
+25. Create container using named nfs volume and cat the contents of the file from the previous step
+26. Check output from each container that was writing to the file.
+27. Stop all the running write containers.
+28. Issue docker volume inspect ${nfsNamedVolume}
+29. Issue docker volume ls
+30. Issue docker volume rm ${nfsDefaultVolume}
+31. Issue docker volume rm ${nfsNamedVolume}
+32. Create a detached container using named nfs volume and write to file every second
+33. Create a container using named nfs volume and tail the file from previous step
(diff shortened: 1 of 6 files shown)


tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
```diff
@@ -0,0 +1,83 @@
+Test 5-22 - NFS Volume
```
mhagen-vmware: This really looks good:
~1. Please add the new test into Test-Cases.md as well so it is indexed~
2. Is there any variations on the install itself related to nfs shares?  All your variations in here are just after VCH install...

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
```diff
@@ -0,0 +1,83 @@
...
+[4 - Docker Command Line Reference - Volume Inspect](https://docs.docker.com/engine/reference/commandline/volume_inspect/)
+
+[5 - Docker Command Line Reference](https://docs.docker.com/engine/reference/commandline/volume_ls/)
+
+
+# Environment:
+This test requires access to VMware Nimbus cluster for dynamic ESXi, vCenter, and NFS server creation
+
+# Test Steps:
+1. Deploy VIC appliance to vCenter with a simple cluster and one NFS server; nfs share point added here.
```
mhagen-vmware: See comment above regarding install variations, but what happens when you try to use a read only share?  What happens when the share is nfsv4 only?
You don't need to implement all of these cases, but if they make sense as test cases then they should be documented at least and we can come back to add implementations (create a follow on story)

mhagen-vmware: what about an nfs share that is just unreachable?
1. during install
2. after install, if the share then becomes unreachable after initially being reachable

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
```diff
@@ -0,0 +1,166 @@
...
+
+*** Variables ***
+${createFileContainer}=  createFileContainer
+${addToFileContainer}=  addToFileContainer
+
+
+*** Keywords ***
+Setup NFS Server
+    Log To Console  \nStarting test...
+    ${esx3}  ${esx4}  ${esx5}  ${vc}  ${esx3-ip}  ${esx4-ip}  ${esx5-ip}  ${vc-ip}=  Create a Simple VC Cluster  datacenter1  cls1
```
mhagen-vmware: I am not sure that VC actually adds anything here? Maybe just need a single ESXi server?

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
```diff
@@ -0,0 +1,166 @@
...
+Suite Teardown  Run Keyword And Ignore Error  Nimbus Cleanup  ${list}
+
+
+*** Variables ***
+${createFileContainer}=  createFileContainer
+${addToFileContainer}=  addToFileContainer
+
+
+*** Keywords ***
+Setup NFS Server
```
mhagen-vmware: should be a more descriptive keyword - since you are deploying nimbus, nfs server, and vch all in one keyword

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
```diff
@@ -0,0 +1,166 @@
...
+    #Need to set nfs volume at vic creation time
+    Install VIC Appliance To Test Server  additional-args=--volume-store="nfs://${NFS_DEFAULT_IP}/store?uid=0&gid=0:nfsVolumeStoreDefault"
+
+Verify NFS Volume Basic Setup
+    [Arguments]  ${prevOutput}  ${ContainerName}  ${nfsIP}  ${rwORro}
+
+    ${rc}  ${outputTemp}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run --name ${ContainerName} -d -v ${prevOutput}:/mydata ${busybox} mount
+    Should Be Equal As Integers  ${rc}  0
+
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} start -a ${ContainerName}
```
mhagen-vmware: not sure why you did docker run followed by docker start?  You should just be able to pull the should container stuff out of the docker run output?

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
```diff
@@ -0,0 +1,166 @@
...
+
+    Set Suite Variable  ${namedNFSVolContainer}  namednfsVolContainer
+    Set Suite Variable  ${nfsNamedVolume}  ${outputDefault}
+
+    Verify NFS Volume Basic Setup  ${nfsNamedVolume}  ${namedNFSVolContainer}  ${NFS_DEFAULT_IP}  rw
+
+Docker volume create already named volume
+    Verify NFS Volume Already Created  ${nfsDefaultVolume}
+
+    Verify NFS Volume Already Created  ${nfsNamedVolume}
```
mhagen-vmware: This is relying on the previous test to succeed, we should not create that dependency

mhagen-vmware: AFAICT the keyword verify nfs volume already create is only used here? if that is true I don't think the keyword is needed.. just do the test here

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
```diff
@@ -0,0 +1,166 @@
...
+
+Docker volume create with possibly invalid name
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} volume create --name=test??? --opt VolumeStore=nfsVolumeStoreDefault
```
mhagen-vmware: make this a bit more interesting:
--name=test!@#$%^&*()

or something similar.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
```diff
@@ -0,0 +1,166 @@
...
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} exec -i ${createFileContainer} sh -c "echo 'The Texas and Chile flag look similar.\n' > /mydata/test_nfs_file.txt"
+    Should Be Equal As Integers  ${rc}  0
+
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} exec -i ${createFileContainer} sh -c "ls mydata/"
+    Should Be Equal As Integers  ${rc}  0
+    Should Contain  ${output}  test_nfs_file.txt
+
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} exec -i ${createFileContainer} sh -c "cat mydata/test_nfs_file.txt"
+    Should Be Equal As Integers  ${rc}  0
+    Should Contain  ${output}  The Texas and Chile flag look similar.
```
mhagen-vmware: All of this exec work is nice, but wholly more complicated then needed... in order to test this is could just be:
docker run echo 'asdf' > /mydata/file.txt; cat /mydata/file.txt

similarly the multi container scenario could just be:
docker run echo 'asdf' > /mydata/file.txt
docker run cat /mydata/file.txt

etc.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
```diff
@@ -0,0 +1,166 @@
...
+Docker write from one container and read from another in named volume
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} exec -i ${addToFileContainer} sh -c "echo 'Norway and Iceland have flags that are basically inverses of each other.\n' >> /mydata/test_nfs_file.txt"
+    Should Be Equal As Integers  ${rc}  0
+
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} exec -i ${createFileContainer} sh -c "cat mydata/test_nfs_file.txt"
+    Should Be Equal As Integers  ${rc}  0
+    Should Contain  ${output}  The Texas and Chile flag look similar.
+    Should Contain  ${output}  The Chad and Romania flag look the same.
+    Should Contain  ${output}  The Luxembourg and the Netherlands flag look exactly the same.
+    Should Contain  ${output}  Norway and Iceland have flags that are basically inverses of each other.
```
mhagen-vmware: Also, could use a case:
multi container, one rm file, other container attempts to read file (both cold read and hot read)

mhagen-vmware: Also, not really sure of the implications - but we at least need a multi-writer test as well and need to figure out what the expected results should be. At least no panics/crashes..
Get at least 5 containers simultaneously writing to the same file.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
rogeliosanchez: @matthewavery Can probably answer this a bit further, but my understanding is that "volume support for VIC starts at deployment time when invoking vic-machine create."  So this would be where we would add the NFS volume stores that we want to use.  And for variations, at the moment we can only do read/write volumes and not read-only.  I can setup different configurations of the NFS server like different disk size and whether we want to use "share" or "/exports/mount-point" instead but for these tests the NFS configuration was not important.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
rogeliosanchez: At the moment we can't do a "read only" share.  @matthewavery can explain further on that one.

As for nfsv4, we only support nfsv3 at the moment.

I can implement the case where an nfs share is not reachable during install.  I can see what that would do to the vic-machine create command.

I need to see how I can kill an NFS server share after install.  That would be a good scenario to investigate/create a test for.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
rogeliosanchez: I agree.  I'll change this to something more descriptive.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
rogeliosanchez: Using just an ESX would speed things up quite a bit.  I don't think the use case with a VC adds any value to this test.

I'll redo this using just an ESX setup.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
rogeliosanchez: For some reason, using the 'mount' command while using "run" was not producing any output.  The container stops immediately after the run command.  That's why I have these two commands in that order.  I'll double check this again just to make sure.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
rogeliosanchez: Good catch.  I'll get rid of the dependency.

I didn't want to repeat the same thing twice here, that's why I created the keyword but I can get rid of the keyword and just have the steps to verify that spelled out here instead.  I do this for the default and named nfs volume.  
@mhagen-vmware Want me to get rid of the keyword and just write out the commands here instead?

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
rogeliosanchez: I was having issues doing `docker run echo 'asdf' > /mydata/file.txt` since the echo command though I wanted to write to a local file instead of the nfs volume.  I experimented with wrapping the echo command in quotes and double quotes but that didn't work.  That's why I created a container that didn't stop after the run command and then used the exec commands afterwards.
The `docker run cat` command should work though.  I don't need to know/keep track of which container is reading that file.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
rogeliosanchez: This is cool.  I'll implement both of these use cases.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
rogeliosanchez: - [ ] use `--name=test!@#$%^&*()` instead

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
mhagen-vmware: Sure, but we should still have cases to test what happens even if our official support statement is we don't do read only and we don't do nfsv4.  We need to know that we fail correctly, because customers will def try it anyways even if we tell them no.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
mhagen-vmware: That would be a bug

rogeliosanchez: - [ ] make sure this is not a bug

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
matthewavery: @mhagen-vmware I can confirm this, right now we do not pass additional mount options or share options through. It is possible(and very likely) that we will want to implement mount options per `VolumeJoin` which would be at `docker create` time. For now we do not, and if you create a `ReadOnly` share point we will fail at install since the portlayer requires the ability to create some namespacing as well as write metadata.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
matthewavery: I will chime in here as well: 

1. Correct, no readonly shares, it might be possible in some scenarios where the volume store was created and configured the exact way a customer wanted. Then we would only join existing containers(probably a conversation for another place/time).

2. Correct, we only support NFSv3. On deployment we would not fail the `vic-machine create` command since we cannot know that it is not mountable until portlayer init time. What we would expect is a warning message from the vic-machine output stating that volume stores x,y, and z were not established successfully. It will also point the customer/user to the portlayer logs for more explanation. We should make sure this is the case. This will encompass the test for trying to establish a store that is NFSv4. Vic-machine should warn you that the store was not created, and you should find a line in the portlayer logs stating `Invalid Protocol` during the volume store init process. 

3. I think it is important to see how we behave when the nfs server falls over after containers are running, though we do not manage the runtime after the container starts. So for that scenario we would expect weird and bad things to start happening to the container process and potentially the r/w layer as well since it is possible logging/writing for that volume could fallback to the underlying FS depending on the running app. The other problem with this, and this is has a ticket open for it, is that the tether will fail silently on NFS mounts that fail at the moment, which can lead customers/users to think they are writing to the nfs server when they are not.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
matthewavery: Yeah that is odd, you should definitely get some output especially if you call `docker logs` on that container too. I am interested to see what you get back from such a call after the container has exited. `mount` writes to STDOUT afaik.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
matthewavery: that should be interesting :)

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
matthewavery: it should definitely be doable, I remember us talking about this some sunday afternoon(maybe saturday?). it will need to be wrapped in `'`s since that include bash style io redirects and you are using a bash interpreter to run the command.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
matthewavery: +1 to this

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
rogeliosanchez: So I was running `run` with `-d` which would cause STDOUT to not be written out.  @matthewavery told me what `-d` was doing in this case.  I tried the `run` command w/o `-d` and now I'm seeing the output I wanted.
I'm rewriting this to make it more simple now that I don't need two commands to get the output I needed.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
rogeliosanchez: @mhagen-vmware, I was able to do the `run` command w/o naming a container after reading your comment, but I remember why I wrote this part of the test that way.  I wanted to make sure that I was dealing with just one named running container in the background.  The exec calls in this test are just for that one container that has already been mounted with an NFS volume.  Also, I use this running container later on in this test for another scenario.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
```diff
@@ -0,0 +1,249 @@
...
+    ${rc}  ${tailOutput}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -v ${nfsNamedVolume}:/mydata ${busybox} sh -c "tail -5 /mydata/test_nfs_kill.txt"
+    Should Be Equal As Integers  ${rc}  125
+    Should Contain  ${tailOutput}  Server error from portlayer: unable to wait for process launch status:
+
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -v ${nfsNamedVolume}:/mydata ${busybox} sh -c "echo 'Where am I writing to?...\n' >> /mydata/test_nfs_kill.txt"
+    Should Be Equal As Integers  ${rc}  125
+
+    ${rc}  ${lsOutput}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -v ${nfsNamedVolume}:/mydata ${busybox} sh -c "ls mydata"
+    Should Be Equal As Integers  ${rc}  125
+    Should Contain  ${lsOutput}  Server error from portlayer: unable to wait for process launch status:
```
mhagen-vmware: Nice, we need bugs for this though.  That error is unclear to the user, if I got that there is no way I would think to check on my NFS server.

rogeliosanchez: @matthewavery, please correct me if I'm wrong here. 
@mhagen-vmware  I don't think I can mount a read-only NFS volume to the VCH at install time at the moment.
I can do something like, mount a read/write NFS volume and then maybe change the permissions of the volume folder inside to read-only afterwards?  I'm not sure if that is what you were looking for in this case but the volume will be created as read/write from the beginning.

As for the error output of that last test, this is what I get:
```
docker: Error response from daemon: Server error from portlayer: unable to wait for process launch status: Post https://10.192.219.205/sdk: context deadline exceeded.
[vic-integration-test-on-pr:L86:1521s] time="2017-07-17T09:38:44Z" level=error msg="error getting events from daemon: net/http: request canceled" 
```
@mhagen-vmware , do you want me to write a bug asking for better output on that?

rogeliosanchez: @mhagen-vmware I was able to deploy a read only NFS mount and attempt a VCH install.  I've create a test for that use case and captured the error messages.  Look at the very first test case.

rogeliosanchez: @mhagen-vmware I was able to deploy a read only NFS mount and attempt a VCH install.  I've create a test for that use case and captured the error messages.  Look at the very first test case.

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
```diff
@@ -0,0 +1,112 @@
...
+# Environment:
+This test requires access to VMware Nimbus for dynamic ESXi and NFS server creation
+
+# Test Steps:
+1. Deploy VIC appliance to an ESX and use a read only NFS mount point
+2. Issue docker volume create on read only volume
+3. Deploy VIC appliance to an ESX and use fake NFS mount point
+4. Deploy VIC appliance to an ESX and use valid NFS mount point
+5. Issue docker volume create using no name for the volume (unnamed) on VolumeStore=nfsVolumeStore (NFS volume)
+6. Run docker mount on container using unnamed volume, run docker wait, then docker rm on temp container
```
anchal-agrawal: This can be made clearer with:

```
6. Issue docker run -v using unnamed volume and the mount command, run docker wait, then docker rm on container
```

matthewavery:
lgtm

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
```diff
@@ -0,0 +1,112 @@
...
+7. Issue docker volume create --name=nfs_default_%{VCH-NAME} on VolumeStore=nfsVolumeStore (NFS volume)
+8. Run docker mount on container using named volume, run docker wait, then docker rm on temp container
```
anchal-agrawal: ditto

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
```diff
@@ -0,0 +1,112 @@
...
+35. Create a container using named nfs volume and tail the file from previous step
+36. Kill the NFS Server from Nimbus
+37. Create a container using named nfs volume from killed NFS server and tail the file from previous step
+38. Create a container using named nfs volume from killed NFS server and write to file from previous step
+39. Create a container using named nfs volume from killed NFS server and ls the mydata directory
+
+
+
+# Expected Outcome:
+* Step 1 will pass VCH creation but should fail in mounting the read only NFS mount poin
```
anchal-agrawal: s/poin/point

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.md
```diff
@@ -0,0 +1,112 @@
...
+* Step 33 should result in error with the following message:  
+```
+Error response from daemon: volume ${nfsNamedVolume} in use by
+```
+* Step 34 - 36 should result in success; step 36 should kill/drop the server
+* Step 37 should result in error with the following message:
+```
+Server error from portlayer: unable to wait for process launch status:
+```
+* Step 38 - 39 should result in error with the rc = 125.
```
anchal-agrawal: s/Step/Steps

PR #5695, opened 2017-07-12
Add inclusion, exclusion filter spec for archive from container

sflxn:
Build the inclusion and exclusion filter spec for archiving from container.
Without exclusion, reader for a volume will read all the data in the submounts.
Without inclusion, readers will not know which subdirectory specified in
the container source path to read from.

### DIFF
--- a/lib/apiservers/engine/backends/archive.go
+++ b/lib/apiservers/engine/backends/archive.go
@@ -253,6 +253,9 @@ func NewArchiveStreamWriterMap(mounts []types.MountPoint, containerDestPath stri
 			aw.filterSpec.StripPath = strings.TrimPrefix(aw.mountPoint.Destination, containerDestPath)
 		}
 
+		aw.filterSpec.Exclusions = make(map[string]struct{})
+		aw.filterSpec.Inclusions = make(map[string]struct{})
+
 		writerMap.prefixTrie.Insert(patricia.Prefix(m.Destination), &aw)
 	}
 
@@ -288,7 +291,10 @@ func NewArchiveStreamReaderMap(mounts []types.MountPoint) *ArchiveStreamReaderMa
 			ar.filterSpec.RebasePath = ar.mountPoint.Destination
 		}
 
-		readerMap.prefixTrie.Insert(patricia.Prefix(m.Destination), ar)
+		ar.filterSpec.Exclusions = make(map[string]struct{})
+		ar.filterSpec.Inclusions = make(map[string]struct{})
+
+		readerMap.prefixTrie.Insert(patricia.Prefix(m.Destination), &ar)
 	}
 
 	return readerMap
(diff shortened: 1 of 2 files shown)


lib/apiservers/engine/backends/archive.go
```diff
@@ -487,13 +490,92 @@ func (rm *ArchiveStreamReaderMap) FindArchiveReaders(containerSourcePath string)
 		}
 
 		if !found {
-			nodes = append(nodes, startingNode)
+			// prepend the starting node at the beginning
+			nodes = append([]*ArchiveReader{startingNode}, nodes...)
 		}
+	} else {
+		startingNode = nodes[0]
```
hickeng: len(node) sanity check? I'm not sure what case would allow it, if any..

lib/apiservers/engine/backends/archive.go
```diff
@@ -487,13 +490,92 @@ func (rm *ArchiveStreamReaderMap) FindArchiveReaders(containerSourcePath string)
...
+	buildExclusion := func(path string, node *ArchiveReader) error {
+		var walkResults []string
+
+		childWalker := func(prefix patricia.Prefix, item patricia.Item) error {
+			if _, ok := item.(*ArchiveReader); !ok {
+				return fmt.Errorf("item not ArchiveReader")
+			}
+
+			ar, _ := item.(*ArchiveReader)
+			walkResults = append(walkResults, ar.mountPoint.Destination)
```
hickeng: Why not the following instead of walkResults?
```
dest := ar.MountPoint.Destination
if path != dest {
  node.filterSpec.Exclusions[dest] = struct{}{}
}
```

We should also make sure that `dest` has a `/` suffix - that way the mountpoint itself can be `stat`'d, for example, but files under it will not be included.

lib/apiservers/engine/backends/archive.go
```diff
@@ -487,13 +490,92 @@ func (rm *ArchiveStreamReaderMap) FindArchiveReaders(containerSourcePath string)
...
+	//		ArchiveReader path -	/
+	//		Inclusion filter -		file.txt
+	//
+	//	Example 2:
+	//		containerSourcePath -	/mnt/A/a/file.txt
+	//
+	//		ArchiveReader path -	/mnt/A
+	//		Inclusion filter -		a/file.txt
+	inclusionPath := strings.TrimPrefix(containerSourcePath, startingNode.mountPoint.Destination)
+	inclusionPath = strings.TrimPrefix(inclusionPath, "/")
```
hickeng: is it viable to keep ensure a leading `/` instead?

lib/apiservers/engine/backends/archive.go
```diff
@@ -487,13 +490,92 @@ func (rm *ArchiveStreamReaderMap) FindArchiveReaders(containerSourcePath string)
...
+	if len(nodes) == 1 {
+		nodes[0].filterSpec.Inclusions[inclusionPath] = struct{}{}
+	}
+
```
hickeng: I think we need Rebase for the correct packing/unpacking of the archive:
```
node.filterSpec.Rebase = startingNode.mountPoint.Destination
```

unpack
```
for file := range archive {
  stripped := strings.TrimPrefix(file.Path, filter.Rebase)
  if included && !exlcuded {
     // write file to target filesystem at filesystem_mountpoint/stripped
  }
}
```

pack
```
for file := range FilesUnderMountpoint {
  if included && !exlcuded {
     stripped := strings.TrimPrefix(file.Path, filesystem-mountpoint)
     rebased := path.Join(filter.Rebase, stripped)
     // add file to archive with header path "rebased"
  }
}
```

lib/apiservers/engine/backends/archive_test.go
```diff
@@ -99,34 +105,137 @@ func TestFindArchiveReaders(t *testing.T) {
 		{
 			containerSourcePath: "/",
 			expectedPrefices:    []string{"/", "/mnt/A", "/mnt/B", "/mnt/A/AB"},
+			expectedFilterSpecs: map[string]ReaderFilters{
+				"/": {
+					rebase:  "",
+					exclude: []string{"/mnt/A", "/mnt/B", "/mnt/A/AB"},
```
hickeng: I think we could only walk one level of subtree for the excludes as "/mnt/A/AB" should be redundant - no need for a change now though, and this is probably more resilient anyway.

lib/apiservers/engine/backends/archive.go
hickeng: hmm. from the tests this is already being added - I've just missed where it's happening.

lib/apiservers/engine/backends/archive_test.go
```diff
@@ -99,34 +105,137 @@ func TestFindArchiveReaders(t *testing.T) {
...
+				"/mnt/B": {
+					rebase:  "/mnt/B",
+					exclude: []string{},
+					include: "",
+				},
+				"/mnt/A/AB": {
+					rebase:  "/mnt/A/AB",
+					exclude: []string{},
+					include: "",
+				},
```
mhagen-vmware: what about the case of an exclude/include collision?

hickeng: That's down to the interpretation - on the portlayer side currently, include wins. There's a need for unit tests to `lib/archive/` to assert this behaviour.

lib/apiservers/engine/backends/archive.go
sflxn: We can.  Just need to define the rules for it.

lib/apiservers/engine/backends/archive_test.go
sflxn: The logic behind include/exclude won't ever produce a collision, but I'll add a check for it anyway.  Can't hurt.

lib/apiservers/engine/backends/archive.go
sflxn: Yeah, that's better.  I made the change and unit test passes.

PR #5698, opened 2017-07-12
[skip ci] Re-implement whitelist test and move to CI

### LINKED ISSUE #5243: Shorten the whitelist CI tests and move them to automatic test set
As a VIC developer, I should see the whitelist testing on PR merges without extending the testing times too long.

Acceptance criteria:  These manual tests runs on every PR merge.

---

Whitelist capability was added in #3586.  The tests for this feature is currently under manual right now because it installs harbor and harbor takes a very long time to start up.  Since this feature is a 1.2 deliverable, we should move this test to the automatic test set, but we need to shorten the length of time it runs or else the builds will time out.

Two ways this can be achieve is to use a preinstalled instance of Harbor and also to use a standard docker registry running in a container instead of using Harbor.  The first proposal has an existing issue, https://github.com/vmware/vic/issues/4181.

mhagen-vmware:
fixes #5243 

Changed my mind... this won't work until docker push support is in, but I can still push this out for review as this is the only viable approach I can see in the near future until harbor improves/simplifies their install or VCH adds more compose support to allow for running harbor

I kept the dinv work in here, because I don't want to lose it, but this PR doesn't make use of it either as that approach also won't work right now until we can establish a better mechanism of linking local volumes and mapping config files into containers being run within dinv

this PR also won't work until #5684 is merged


gigawhitlocks:
FYI #5864 was closed in lieu of #5702 after a discussion with Hasan resulted in a different approach

sgairo:
this should fix #5822 as well

tests/test-cases/Group21-Registries/21-1-Whitelist.robot
```diff
@@ -0,0 +1,113 @@
...
+Resource  ../../resources/Harbor-Util.robot
+Suite Setup  Setup Registries
+Suite Teardown  Cleanup Registries
+Test Teardown  Cleanup VIC Appliance On Test Server
+
+*** Keywords ***
+Setup Registries
+    Install VIC Appliance To Test Server
+    Remove Environment Variable  DOCKER_HOST
+    Set Environment Variable  REGISRTRY-URL  %{VCH-IP}
```
sflxn: misspell

sflxn:
We don't check for CIDR and wildcard domains anymore?

mhagen-vmware:
No plan to support docker push

Issue #3962, opened 2017-02-15
VIC Statistic design


cgtexmex:

cgtexmex:
implemented via #5699

PR #5699, opened 2017-07-12
[specific ci=1-39-Docker-Stats] Add network and disk stats to docker stats

### LINKED ISSUE #4422: docker network and storage stats
Docker stats was partially implemented with #4413 by providing CPU & Memory stats.  Network and storage stats still need to be implemented.

cgtexmex:
Docker stats now reports network and disk stats. Additionally,
stats collection refactored to only run when there are active
subscriptions.

Fixes #4422, #3962, #5746


cgtexmex:
robot tests are in progress, but the code is complete

lib/portlayer/metrics/container.go
```diff
@@ -66,113 +62,103 @@ type MemoryMetrics struct {
...
-		if cc.SubscriberCount() == 0 {
-			continue
+// Start will begin the collection polling process
+func (cc *vmCollector) Start() {
+	// create timer with sampleInterval alignment
+	cc.timer = time.NewTicker(time.Duration(int64(sampleInterval)) * time.Second)
+	// loop on the channel
+	go func() {
+		for {
+			select {
```
dougm: Stop() doesn't close the ticker channel, don't you need another case in here to break the loop / return from the go func?

cgtexmex: Yes -- I thought that since Stop() "released the ticker resources" that it would close...but that's not the case.  I've added logic (locally) to break out of the loop -- I'll push up for review in a few.

cgtexmex:
@dougm now correctly exiting from the polling/collection loop

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -316,3 +419,40 @@ func (cs *ContainerStats) currentCPU() {
...
 }
+
+const abc = "abcdefghijklmnopqrstuvxyz"
+
+// toChar will take the provided int and slice the
+// alpha constant to return the string representation.  For example
+// if 1 is provided 'a' will be returned, 2 == 'b', 3 == 'c', etc..
+func toChar(i int) string {
+	return abc[i-1 : i]
+}
```
mhagen-vmware: I can't see where this is used in this PR? Simple grep at least doesn't show anything..

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -316,3 +419,40 @@ func (cs *ContainerStats) currentCPU() {
...
+		// unlikely this would happen, but if it does and there is more than one disk on the vm
+		// then it could go undetected
+		log.Errorf("stats error generating container(%s) disk(%s) minor: %s", containerID, name, err)
+	}
+	// minor identifiers are multiples of 16
+	minor *= 16
+	return uint64(minor)
+}
+
+func createBlkioStatsEntry(minor uint64, op string, value uint64) types.BlkioStatEntry {
```
mhagen-vmware: should this have some error handling?  minor and string are generally covered by your own code - i.e. minor is either going to be valid or 0 and string is hard coded to read/write/total in all places, but value does not appear to be guaranteed to be a uint64 necessarily in the case of an error?

lib/portlayer/metrics/container.go
```diff
@@ -224,117 +215,200 @@ func (cc *collectorVM) collect(currentSubs map[types.ManagedObjectReference]*sub
...
-	cc.mu.Lock()
-	defer cc.mu.Unlock()
-	return len(cc.subs)
+// instanceID coverts the ID or Key of a metric from a string to int.
+func instanceID(containerID string, instance string) (int, error) {
+	converted, err := strconv.Atoi(instance)
+	if err != nil {
+		// I don't expect this to ever happen, but if it does log and don't publish
+		log.Errorf("metrics failed to convert the container(%s) device id to an int - value(%#v): %s", containerID, instance, err)
+		return converted, err
```
mhagen-vmware: I am not familiar enough with golang Atoi, but since you move this here and out of the switch statement above - does Atoi at least populate converted with 0/-1 or something sensible so that we are guaranteed we are not returning some totally invalid value?

cgtexmex: If it fails the int will be zero -- but if it fails we'll return the err and we'll skip that device, so we won't report.

I moved it out of the switch as I thought I'd be using it in other areas (disk, net & not just CPU), but didn't need it for those...

lib/apiservers/engine/backends/convert/stats.go
cgtexmex: the value will always be uint64...IMO no error handling needed

lib/apiservers/engine/backends/convert/stats.go
cgtexmex: Yeah..it's not used...I thought originally I would need to use it...I'll remove.

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -42,6 +44,11 @@ type ContainerStats struct {
 	curDockerStat *types.StatsJSON
 	currentMetric *metrics.VMMetrics
 
+	// disk & net stats are accumulated during the life of the
+	// subscription. These maps will assist in that accululation
```
sgairo: "accumulation" :)

lib/portlayer/metrics/container.go
```diff
@@ -66,113 +62,117 @@ type MemoryMetrics struct {
...
-	// create a map with this single subscriber
-	single := make(map[types.ManagedObjectReference]*subscriber)
-	single[containerSub.ref] = containerSub
-
-	// collect the metrics
-	go cc.collect(single, true)
-
-	return ch, nil
+	cc.timer.Stop()
+	cc.stopper <- true
```
dougm: You can just close the channel here, won't be used after this anyhow.  The golang nerds would also tell you to use `chan struct{}` instead of `chan bool`, since it uses less memory.

cgtexmex: You can't close the channel as it's receive only.  (https://play.golang.org/p/NACq0RYGGd)

I originally had as `chan struct{}` for that reason, but then changed to bool...

lib/apiservers/engine/backends/convert/stats.go
cgtexmex: I like my spelling better... ;-)

lib/portlayer/metrics/container.go
dougm: I meant close the cc.stopper channel, not the ticker.C

cgtexmex: ah..yeah..I could...was just avoiding the `make` on each Start...regardless I'm switching to an empty struct and will be closing...so we can all be happy ;-)

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -42,6 +43,11 @@ type ContainerStats struct {
 	curDockerStat *types.StatsJSON
 	currentMetric *metrics.VMMetrics
 
+	// disk & net stats are accumulated during the life of the
+	// subscription. These maps will assist in that accumulation.
+	diskStats map[string]metrics.VirtualDisk
+	netStats  map[string]metrics.Network
```
caglar10ur: Do we drop the ContainerStats when the container gets removed? Also how about restarts, do we just keep adding or somehow start from zero values?

cgtexmex: We drop the stats when the stats stream (subscription) ends.  That could be via the client or the API.  If the interaction is without streaming (`docker stats --no-stream`) then there's no accumulation and only a point in time look at disk and network.

lib/portlayer/metrics/subscription.go
```diff
@@ -0,0 +1,180 @@
...
+	var name string
+
+	// did we previously find this device
+	if name, exists := sub.deviceInstanceToKey[instance]; exists {
+		return name
+	}
+
+	// convert instance to key - we are expecting regular failures, so no logging
+	// or returning of error
+	key, err := strconv.Atoi(instance)
```
hickeng: do you have an example of an `instance` that shouldn't be translated we can add to doc?
I'm not clear on what would be passed in that's not translated - I also cannot see where this method is called.

lib/portlayer/metrics/subscription.go
```diff
@@ -0,0 +1,180 @@
...
+
+	"github.com/vmware/govmomi/object"
+
+	"github.com/vmware/vic/lib/portlayer/exec"
+)
+
+// subscription is a 1:1 relationship to a container and a
+// 1:M relationship to subscribers
+type subscription struct {
+	c *exec.Container
```
hickeng: Is it viable for this to use `vm.VirtualMachine` instead of exec.Container? I'm wondering if this needs to be in the portlayer or if it will happily sit in `pkg/vsphere`/potential inclusion into govmomi.

lib/portlayer/metrics/subscription.go
```diff
@@ -0,0 +1,180 @@
...
+	return sub.pub.Subscribe()
+}
+
+// Evict will remove the channel from the publisher ending the subscription
+func (sub *subscription) Evict(ch chan interface{}) {
+	sub.pub.Evict(ch)
+}
+
+// newContainerSubscription is a helper func to convert the interface to a subscription
+func newContainerSubscription(contain interface{}) (*subscription, error) {
```
hickeng: Also couldn't find a call - in this PR or searching current master.

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -235,6 +247,95 @@ func (cs *ContainerStats) ToContainerStats(current *metrics.VMMetrics) (*types.S
...
+}
+
+// disk will calculate supported stats by disk device.  The stats presented will be the
+// disk stats accumulated during the stats subscription.  This differs from vanilla
+// docker as it provides the disk stats for the lifetime of the container.
+//
+// Supported stats are io_service_bytes_recursive and io_serviced_recursive, so bytes and iops
+// during the stats subscription
+//
+// TODO: Currently disk assumes a single scsi device.  Multiple scsi devices will need
```
cgtexmex: Controllers, not devices...updated comment forthcoming...

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -205,10 +213,8 @@ func (cs *ContainerStats) Listen() *io.PipeWriter {
 func (cs *ContainerStats) ToContainerStats(current *metrics.VMMetrics) (*types.StatsJSON, error) {
 	// if we have a current metric then validate and transform
 	if cs.currentMetric != nil {
-		// do we have the same metric as before?
-		if cs.currentMetric.SampleTime.Equal(current.SampleTime) {
-			// we've already got this as current, so skip and wait for the
-			// next sample
+		// do we have the same metric or has the metric not been initialized?
+		if cs.currentMetric.SampleTime.Equal(current.SampleTime) || current.SampleTime.IsZero() {
```
cgtexmex: yes - at the end of a stats call there was an intermittent issue where a final metric would be produced, but it wasn't completely initialized.   The easiest way to tell is the fact that the sampleTime wasn't set.

lib/portlayer/metrics/subscription.go
cgtexmex: Yes we could use `vm.VirtualMachine` -- want me to do that now or later?

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -316,3 +417,31 @@ func (cs *ContainerStats) currentCPU() {
...
 }
+
+// diskMinor will parse the vSphere provided disk name and return the minor id of
+// the disk device.  The func assumes that minor identifiers are multiples of
+// 16 (0,16,32,48,etc) and that the containerVM has 10 or less disks
+func diskMinor(containerID string, name string) uint64 {
+	// retrieve the last char
+	last := string(name[len(name)-1])
+	// the int suffix should always be last
+	minor, err := strconv.Atoi(last)
```
cgtexmex: when creating this I apparently switched controller maximum w/disk maximum...so this was coded to having fewer than 10 disks.  I'll need to parse the string differently.  FTR the names look like this: `scsi0:0`

lib/apiservers/engine/backends/convert/stats.go
```diff
@@ -75,6 +81,8 @@ func NewContainerStats(config *ContainerStatsConfig) *ContainerStats {
 		curDockerStat: &types.StatsJSON{},
 		totalVCHMhz:   uint64(config.VchMhz),
 		dblVCHMhz:     uint64(config.VchMhz * 2),
```
cgtexmex: It is not related to vCPU count.  I have a comment as to why this is the case when the variable is used:  https://github.com/cgtexmex/vic/blob/stats-4422/lib/apiservers/engine/backends/convert/stats.go#L402

lib/portlayer/exec/base.go
```diff
@@ -97,6 +98,19 @@ func (c *containerBase) VMReference() types.ManagedObjectReference {
 	return moref
 }
 
+// Devices will return a VirtualDeviceList
+func (c *containerBase) Devices() (object.VirtualDeviceList, error) {
```
cgtexmex: Yes...I'll get it added...

cgtexmex:
@hickeng I've moved the performance collector to `pkg/vsphere/performance` thus decoupling from exec and positioning for the gateway.  Additionally I've added tracing to the flow.  Let me know if you have other questions / concerns.

lib/portlayer/metrics/metrics.go
```diff
@@ -34,18 +38,19 @@ var (
...
+type UnsupportedTypeError struct {
+	subscriber interface{}
+}
+
+func (ute UnsupportedTypeError) Error() string {
+	t := reflect.TypeOf(ute.subscriber)
+	if t.Kind() == reflect.Ptr {
+		t = t.Elem()
+	}
+	return fmt.Sprintf("type %s is not supported by metrics", t.Name())
```
hickeng: "%T" didn't provide sufficient detail?

lib/portlayer/metrics/metrics.go
```diff
@@ -67,13 +72,36 @@ func Init(ctx context.Context, session *session.Session) error {
...
 	}
 }
 
-// VMCollector will return the vm metrics collector
-func (s *super) VMCollector() Collector {
-	return s.vms
+func (s *super) Subscribe(op trace.Operation, subscriber interface{}) (chan interface{}, error) {
+	switch subscriber.(type) {
+	case *exec.Container:
+		container := subscriber.(*exec.Container)
```
hickeng: type cast is unneeded in a type switch. It's automatic through Go magic.

lib/portlayer/metrics/metrics.go
```diff
@@ -67,13 +72,36 @@ func Init(ctx context.Context, session *session.Session) error {
...
+	}
+	op.Errorf("%s", err)
+
+	return nil, err
+}
+
+func (s *super) Unsubscribe(op trace.Operation, subscriber interface{}, ch chan interface{}) {
+	switch subscriber.(type) {
+	case *exec.Container:
+		s.vms.Unsubscribe(op, subscriber.(*exec.Container).VMReference(), ch)
```
hickeng: follow the code structure/patter in `Subscribe`

pkg/vsphere/performance/vm_collector.go
```diff
@@ -0,0 +1,407 @@
...
+				default:
+					name := sub.DeviceName(v.Instance)
+					// if we have a name gather the stats
+					if name != "" {
+						// get the network to update
+						net := findNetwork(metric, name)
+						switch v.Name {
+						case "net.bytesRx.average":
+							// perfManager returns kbps -- convert to sum of bytes
+							sum := summation(v.Value[s]) * 1024
```
hickeng: `kbps * 1024` translates to `bps` - is `Rx.Bytes` actually `Rx.BytesPerSecond` or is there an unstated assumption that the interval is 1s?

lib/portlayer/metrics/metrics.go
cgtexmex: Not sure that I'm following.  I'm going to call methods if it's a container, so I have to cast it.  What am I missing?

lib/portlayer/metrics/metrics.go
cgtexmex: I've changed it...to just the type...so `fmt.Sprintf("type %s is not supported by metrics", t)`

pkg/vsphere/performance/vm_collector.go
cgtexmex: This is total bytes for the interval.  The `summation` func will multiply the `kbps` by the `interval` and then we multiply that by `1024` to get total bytes.

lib/portlayer/metrics/metrics.go
cgtexmex: done `switch sub := subscriber.(type){...}`

Issue #5634, opened 2017-07-03
Implement container store in the storage portlayer


jzt:
As a developer, I need to be able to attach/detach VMDKs that represent container R/W layers and perform operations on them (copy to/from, diff, etc).

Details:
Currently, a container's R/W layer VMDK lives outside any of our caches or stores, so we don't currently have a way to attach and mount those disks so that we can perform filesystem operations on them.  A container store on the storage portlayer can be used as the component for handling this, similar to how the volume and image stores handle volume/image VMDKs.

Proposed solution:

1. Docker persona passes container ID, container store URL, image layer ID, and image store URL (along with any other required parameters) to the portlayer `storage.ExportArchive` endpoint (currently being implemented by @sflxn)
2. `storage.ExportArchive` uses the image cache to get the path to the image leaf layer VMDK on the datastore
3. `storage.ExportArchive` passes the container ID and the image layer disk path and into the container store's `Export` function
4. `ContainerStore.Export` derives the path of the R/W layer VMDK on the datastore
5. `ContainerStore.Export` attaches & mounts the R/W layer VMDK and the image layer VMDK
6. `ContainerStore.Export` calls `archive.Diff` on the two mount paths and returns the result

Requirements:
- [ ] a `ContainerStorer` interface (similar to `ImageStorer` and `VolumeStorer`) initially defining a single function `Export`, eventually to include `Import`
- [ ] Logic in the storage portlayer's `ExportArchive` function to determine that its dealing with a container filesystem and taking the correct path

Note: this will not work on a running container without additional logic to manage the container state during this operation.

Issue #5318, opened 2017-05-31
Docker persona entrypoint into diff implementation


jzt:
The entrypoint into `docker diff` needs to be implemented in the docker persona at the [`func (c *Container) ContainerChanges(name string) ([]archive.Change, error)`](https://github.com/vmware/vic/blob/master/lib/apiservers/engine/backends/container.go#L1313) endpoint.

This will involve a call to the portlayer's [`GetImageTar`](#5319) endpoint that will return information reflecting the changes. This information will take the form of a header-only tar archive of the filesystem diff.

The persona will create and return the `[]archive.Change` result by enumerating the contents of the result.

Acceptance criteria:

- [ ] `ContainerChanges()` is implemented to make the portlayer call to [`GetImageTar`](#5319)
- [ ] The result is analyzed and used to create and return the `[]archive.Change` slice (see [here](https://github.com/vmware/vic/blob/master/vendor/github.com/docker/docker/pkg/archive/changes.go#L50))
- [ ] unit and/or integration tests where appropriate

**Notes**
This is likely to break down into two pieces of work:
1. pkg/archive/changes.go - convert a header only `archive/tar` archive into a ChangeSet
2. use (1) to implement the glue from portlayer.ReadImage to the client response

Main item of dependency with #5320 is whether we use vendor headers to differentiate updated files from created files - the bulk of this work can be done without that knowledge using a stub in container_proxy that uses `archive/tar` to create a test diff archive.

Issue #5295, opened 2017-05-30
Portlayer: offline copy via disk attach


hickeng:
**Story**
As a VIC developer I want a copy API implementation for when the target container is powered *off*
As a Port Layer caller I want a copy API that doesn't require knowledge of target state

**Detail**
Copying files to/from a container disk while the container is off requires attaching the VMDK to the endpoint VM, mounting it, then extracting or injecting the files.
Container disk should be detached on both success and error paths.
Parallel copying of files from multiple docker clients should be blocked, otherwise we cannot ensure consistency.
Attempts to power on container while disk is attached to endpoint should be blocked or rejected.

**Questions**
- [X] what is the expected behaviour when it comes to volumes and files on volumes - are copyFrom/copyTo symmetric in their behaviour with respect to volumes?
  - recursion into volumes is the responsibility of the caller to the Import/Export endpoints defined in #5321  and behaviour should be symmetrical. For a full implementation the caller should ensure that recursion into volumes occurs.

**Acceptance**
- [x] file copy into container - subsequent start of container shows file structure in container
- [x] file copy out of container - file structure present on client system
- [ ] container disks are not attached to endpoint irrespective of success or error paths
- [x] parallel copy requests do not collide
- [x] container start should not attempt to power on the container while copy is in progress - fails with X in use.
- [ ] container should start successfully even if endpointVM goes down in the middle of copying - does fsck and journal replay get performed correctly?

matthewavery:
We must know in the portlayer whether the directory override flag is set or not. Since we are mounting the disk directly to the VCH.

hickeng:
Added the dependency with ReadImage - this isn't a new dependency but simply captures the fact that the backend portion of diff has to do essentially the same work in the copyFrom case.

For the copyTo scenario we have two options:
1. add a new portlayer endpoint specifically for copying files into a container
2. update WriteImage to be able to write directly to an existing image if it has no children.

My preference is (2) from a gut feel - I'd prefer to have time to think through the implications but doubt that'll be available. Primary concerns for (2):
* increased conceptual complexity for the call as it's no longer write-once then immutable.
* changing one of our core paths could introduce/expose bugs (while we shouldn't be making decisions based on this issue, but it has to be mentioned given time considerations).

PR #5700, opened 2017-07-12
Import/Export support [full ci]

hickeng:
Adds Import and Export support, including diff.

Current issues:
* does not integrate online copy or any stat paths
* rebase/strip paths probably need work

Towards: #5295 
Fixes: #5712 #5634 #5318 

Impacted by: #5729


hickeng:
@chengwang86 I've added you as a reviewer because of the commit to reduce the number of diff entries. I've ended up adding to the base image, but need to clarify why some of the files exist. Without those changes we were seeing the following diff:
```
vagrant@devbox:~/vicsmb/src/github.com/vmware/vic$ docker run -it -v darth:/mnt/darth busybox
/ # echo hello > /tmp/goodbye
/ # echo hello > /mnt/darth/world
/ # exit
vagrant@devbox:~/vicsmb/src/github.com/vmware/vic$ docker ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED              STATUS                      PORTS               NAMES
640acf5c535d        busybox             "sh"                About a minute ago   Exited (0) 12 seconds ago                       tender_ride
vagrant@devbox:~/vicsmb/src/github.com/vmware/vic$ docker diff 64
C /etc/
A /etc/mtab
A /mnt/
A /mnt/darth/
C /root/
A /root/.ash_history
A /run/
A /sbin/
C /tmp/
A /tmp/goodbye
C /var/
A /var/run/
A /var/run/tether.pid
```

I've still no idea where `sbin` is coming from but I've also not tried busybox with regular docker for reference.

chengwang86:
@hickeng I tried the above steps on vanilla docker and got the following output 
```
A /mnt
A /mnt/darth
C /root
A /root/.ash_history
C /tmp
A /tmp/goodbye
```

chengwang86:
@hickeng Is there a design doc for diff? Did you use `file/dir mask` in diff?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -681,6 +683,26 @@ func (c *ContainerProxy) StreamContainerStats(ctx context.Context, config *conve
 	return nil
 }
 
+func (c *ContainerProxy) GetContainerChanges(vc *viccontainer.VicContainer) (io.ReadCloser, error) {
```
emlin: Need to have a method to export container r/w layer. The only difference with this method is to set data parameter to true while calling the ArchiveExportReader.
So wondering is it better to add one more parameter in this method?

hickeng: I believe this entry point is constrained by the docker API logic- they could both call a common function however.

emlin: this is the interface of ContainerProxy, not Docker API, that's why I want to have one more parameter here, so that I can get data back. I did that change in my PR #5719

hickeng: ah yes, my lazy reading. Sounds good.

lib/apiservers/engine/backends/archive.go
```diff
@@ -249,9 +229,15 @@ func NewArchiveStreamWriterMap(mounts []types.MountPoint, containerDestPath stri
...
 		// Here, we must tell the portlayer to remove "mnt/A".  The key to determining whether to
 		// strip "A" or "mnt/A" is based on the container destination path.
-		if containerDestPath != "/" && strings.HasPrefix(aw.mountPoint.Destination, containerDestPath) {
-			aw.filterSpec.StripPath = strings.TrimPrefix(aw.mountPoint.Destination, containerDestPath)
-		}
+
+		// hickeng: the current logic is strip, then rebase for export - more efficient to collapse
+		// any reducdency here than apply to every tar entry when packing but skipping for now.
+		aw.filterSpec.StripPath = aw.mountPoint.Destination
+		aw.filterSpec.RebasePath = strings.TrimPrefix(dest, aw.mountPoint.Destination)
```
jzt: A `RebasePath` should be tacked onto the front of an existing path, but it looks like we are taking the end of an existing path to use here.

lib/apiservers/engine/backends/container.go
```diff
@@ -1274,8 +1276,53 @@ func (c *Container) ContainerWait(name string, timeout time.Duration) (int, erro
...
+	changes := []docker.Change{}
+
+	tarFile := tar.NewReader(r)
+
+	defer r.Close()
+
+	for {
+		hdr, err := tarFile.Next()
+		if err == io.EOF {
+			log.Infof("Got EOF!! BAILING")
```
jzt: Can remove this, it was for debugging :)

lib/apiservers/engine/backends/container.go
```diff
@@ -1274,8 +1276,53 @@ func (c *Container) ContainerWait(name string, timeout time.Duration) (int, erro
...
+		case "D":
+			change.Kind = docker.ChangeDelete
+		case "C":
+			change.Kind = docker.ChangeModify
+		default:
+			return []docker.Change{}, InternalServerError("Invalid change type")
+		}
+		changes = append(changes, change)
+	}
+	log.Infof("\t\t\tChanges: %#v", changes)
```
jzt: Same, remove

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -719,8 +741,10 @@ func (c *ContainerProxy) ArchiveExportReader(ctx context.Context, store, ancesto
 			log.Infof(" encodedFilter = %s", encodedFilter)
 		}
 
+		log.Debugf("Starting export from %s", deviceID)
 		_, err = PortLayerClient().Storage.ExportArchive(params, pipeWriter)
 		if err != nil {
+			log.Infof("\n\n\nGot error from ExportArchive: %s", err.Error())
```
jzt: Debug logging, can remove (or set to `Debugf` and maybe remove the `\n`'s

lib/apiservers/engine/backends/archive.go
matthewavery: This should get fixed with the spec generation PR that I have out(https://github.com/vmware/vic/pull/5720). Currently this will generate invalid spec for things like sub volume targets and sub volume CopyTo/CopyFroms that would target the root of the volume itself.

lib/apiservers/engine/backends/archive.go
```diff
@@ -107,26 +100,24 @@ func (c *Container) ContainerExport(name string, out io.Writer) error {
...
+	if vc == nil {
+		return NotFoundError(name)
+	}
 
-	// err := c.importToContainer(vc, path, content)
-	// if err != nil && IsResourceInUse(err) {
-	// 	log.Errorf("ContainerExtractToDir failed, resource in use: %s", err.Error())
+	err := c.importToContainer(vc, path, content)
+	if err != nil && IsResourceInUse(err) {
+		log.Errorf("ContainerExtractToDir failed, resource in use: %s", err.Error())
```
matthewavery: Since this is likely to be caused by another person/script issuing commands( primarily `docker stop/start`). We should also let the user know that an additional cp call should succeed unless the resource is being thrashed(multiple actors calling start/stop repeatedly). Maybe just add on, "Please reissue the command".  -emma,zach,matt

lib/apiservers/engine/backends/archive.go
```diff
@@ -29,51 +27,46 @@ import (
...
-	// if err != nil && IsResourceInUse(err) {
-	// 	log.Errorf("ContainerArchivePath failed, resource in use: %s", err.Error())
-	// 	err = fmt.Errorf("Resource in use")
-	// }
-	// if err == nil || reader != nil {
-	// 	content = reader
-	// }
-	// stat = nil
+	reader, err = c.exportFromContainer(vc, path)
+	if err != nil && IsResourceInUse(err) {
```
matthewavery: ditto to the reissue command comment.

lib/apiservers/engine/backends/archive.go
```diff
@@ -287,9 +273,11 @@ func NewArchiveStreamReaderMap(mounts []types.MountPoint) *ArchiveStreamReaderMa
 		//
 		// Neither the volume nor the storage portlayer knows about /mnt/A.  The persona must tell
 		// the portlayer to rebase all files from this volume to the /mnt/A/ in the final tar stream.
-		if ar.mountPoint.Destination != "/" {
-			ar.filterSpec.RebasePath = ar.mountPoint.Destination
-		}
+
+		// hickeng: the current logic is rebase, then strip for export - more efficient to collapse
+		// any reducdency here than apply to every tar entry when packing but skipping for now.
+		ar.filterSpec.RebasePath = ar.mountPoint.Destination
```
matthewavery: This will not work for every case, but should be addressed in the FilterSpec generation PR.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -672,15 +683,37 @@ func (c *ContainerProxy) StreamContainerStats(ctx context.Context, config *conve
...
+		return nil, InternalServerError("Failed to determine host UUID")
+	}
+
+	parent := vc.LayerID
+	spec := archive.FilterSpec{
+		Inclusions: map[string]struct{}{},
+		Exclusions: map[string]struct{}{},
+	}
+
+	r, err := c.ArchiveExportReader(context.Background(), "container", host, vc.ContainerID, parent, false, spec)
```
matthewavery: use the constant you created :) instead of "container"

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -719,8 +751,9 @@ func (c *ContainerProxy) ArchiveExportReader(ctx context.Context, store, ancesto
 			log.Infof(" encodedFilter = %s", encodedFilter)
 		}
 
-		_, err = PortLayerClient().Storage.ExportArchive(params, pipeWriter)
+		_, err = plClient.Storage.ExportArchive(params, pipeWriter)
 		if err != nil {
+			log.Infof("\n\n\nGot error from ExportArchive: %s", err.Error())
```
matthewavery: Please either remove this or make the format nice. >:) 
- zach

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1062,6 +1089,53 @@ func (c *ContainerProxy) Signal(vc *viccontainer.VicContainer, sig uint64) error
...
+	}
+
+	r.Transport = transport
+
+	plClient := client.New(r, nil)
+	bsc := runtime.ByteStreamConsumer()
+	r.Consumers["application/octet-stream"] = bsc
+	r.Producers["application/octet-stream"] = runtime.ByteStreamProducer()
+
+	r.Consumers["application/x-gzip"] = runtime.ConsumerFunc(func(rdr io.Reader, data interface{}) error {
```
matthewavery: This block is unused and incorrect so might as well be removed. 
-zach

lib/archive/diff_test.go
```diff
@@ -264,6 +264,8 @@ func TestDiffNoData(t *testing.T) {
 }
 
 func TestDiffFilterSpec(t *testing.T) {
+	t.Skip()
```
mhagen-vmware: Pretty sure I saw somewhere why we are skipping all of these now, but I can't remember now.  Can you put the reason into the body of the PR and update the body for that matter too, as I don't think that the original comment is still valid?

lib/archive/unpack_test.go
```diff
@@ -463,6 +463,8 @@ func TestSimpleExclusion(t *testing.T) {
 }
 
 func TestInclusionAfterExclusion(t *testing.T) {
+	t.Skip()
```
mhagen-vmware: please add explanation here too

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -517,68 +532,76 @@ func (h *StorageHandlersImpl) VolumeJoin(params storage.VolumeJoinParams) middle
...
 	}
 
-	params.Archive.Close()
+	store, ok := spl.GetExporter(params.Store)
+	if !ok {
+		op.Errorf("Failed to locate export capable store %s", params.Store)
+		op.Debugf("Available exporters are: %+q", spl.GetExporters())
 
-	log.Infof(buf.String())
+		// TODO: this should be a 404 but cannot seem to figure that out #shamed
```
matthewavery: You did figure it out remove this comment.
-zach

lib/archive/diff.go
```diff
@@ -103,16 +113,28 @@ func Tar(op trace.Operation, dir string, changes []docker.Change, spec *FilterSp
...
+						case <-op.Done():
+							f.Close()
+						case <-done:
+						}
+					}()
+
 					_, err = io.Copy(tw, f)
+					close(done)
 					if err != nil {
 						op.Errorf("Error writing archive data: %s", err.Error())
```
matthewavery: Should we bail out from this? or just return a partially copied file?

lib/archive/diff_test.go
matthewavery: Add a `TODO` so that we know to fix this later.
-emma

lib/archive/unpack.go
```diff
@@ -34,37 +34,31 @@ const (
...
+	if err != nil {
 		// the target unpack path does not exist. We should not get here.
-		op.Errorf("tar unpack target does not exist (%s)", unpackPath)
+		op.Errorf("tar unpack target does not exist: %s", root)
 		return err
 	}
 
-	finalTargetPath := filepath.Join(unpackPath, target)
-	op.Debugf("finalized target path for Tar unpack operation at (%s)", finalTargetPath)
+	if !fi.IsDir() {
```
matthewavery: Add a TODO here to handle file overwrite in case the target is a file instead of a directory. 
-matt

morris-jason: Not sure this is accurate. 
You can't unpack a tar archive of a regular file to a regular file, can you? 
You would unpack the tar archive to the destination's directory.
ie
```bash
docker cp /test/file.txt cid:/tmp/file.xt
```
You would unpack the tar archive with header name (file.txt) to /tmp.

lib/archive/unpack.go
```diff
@@ -73,64 +67,54 @@ func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, unpackP
...
 		// skip excluded elements unless explicitly included
-		if excluded(header.Name, filter) {
+		if filter.Excludes(op, header.Name) {
 			continue
 		}
 
 		// fix up path
-		strippedTargetPath := strings.TrimPrefix(header.Name, strip)
-		writePath := filepath.Join(finalTargetPath, strippedTargetPath)
+		stripped := strings.TrimPrefix(header.Name, filter.StripPath)
```
matthewavery: after the filterspec generation PR is in, all the inclusion and exclusions will be relative to the mount. So move this line above line 79. And change `filter.Excludes(op, header.Name)` to `filter.Excludes(op, stripped)`.

lib/archive/unpack.go
```diff
@@ -73,64 +67,54 @@ func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, unpackP
...
-			// a symlink. this behavior should be evaluated for a proper
-			// response, for now a regular file is made since that is the
-			// target of a sym link operation
-			err := os.Symlink(header.Linkname, writePath)
-			if err == nil {
-				continue
+			err := os.Symlink(header.Linkname, absPath)
+			if err != nil {
+				op.Errorf("Failed to create symlink %s->%s: %s", absPath, header.Linkname, err)
+				return err
```
matthewavery: Should we be bailing out of the entire operation due to the inability to make one symlink? Or should we just log this and then write it as a regular file empty file?

lib/archive/unpack_test.go
matthewavery: Please add a TODO so that I know to fix this in my Filterspec PR.

lib/portlayer/storage/data_source.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		op.Errorf("Unable to stat mount path %s for data source: %s", name, err)
+		return nil, err
+	}
+
+	if !fi.IsDir() {
+		return nil, errors.New("path must be a directory")
+	}
+
+	// NOTE: this isn't actually diffing - it's just creating a tar. @jzt to explain why
```
matthewavery: naming is hard 
-zach

lib/portlayer/storage/data_sink.go
```diff
@@ -0,0 +1,89 @@
...
+		return errors.New("Path must be a directory")
+	}
+
+	// This assumes that m.Path was opened with a useful path (i.e. absolute) as that argument is what's
+	// returned by Name.
+	op.Infof("Importing supplied data stream to %s", name)
+	return archive.Unpack(op, data, spec, name)
+}
+
+func (m *MountDataSink) Close() error {
```
matthewavery: op.Infof that we are closing the DataSink here, since the source also logs in it's close.

lib/portlayer/storage/vsphere/export.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	// online - Owners() should filter out the appliance VM
+	owners, _ := v.Owners(op, uri, disk.LockedVMDKFilter)
+	if len(owners) == 0 {
+		op.Infof("No online owners were found for %s", id)
+		return nil, errors.New("unable to create offline data source and no online owners found")
+	}
+
+	// TODO(jzt): tweak this when online export is available
```
matthewavery: Remove this comment, tweaking complete
-zach

lib/portlayer/storage/vsphere/export.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	op.Debugf("Created mount data source for access to %s at %s", url, mountPath)
+	return storage.NewMountDataSource(op, f, cleanFunc), nil
+}
+
+func (v *VolumeStore) newOnlineDataSource(op trace.Operation, owner *vm.VirtualMachine, id string) (storage.DataSource, error) {
+	return &ToolboxDataSource{
+		VM: owner,
+		// TODO: there's some mangling that happens from volume id to disk label so this isn't currently correct
```
matthewavery: is this referring to the 16 character limit on disk labels?

lib/portlayer/storage/vsphere/image.go
```diff
@@ -50,13 +53,19 @@ var (
 		"/etc/hosts":       0644,
 		"/etc/resolv.conf": 0644,
 	}
+
+	SymLinkForMinOS = map[string]string{
```
matthewavery: Do we need one of these for sbin?

lib/portlayer/storage/vsphere/toolbox_datasink.go
```diff
@@ -0,0 +1,68 @@
...
+	"io"
+
+	"github.com/vmware/govmomi/guest"
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/lib/archive"
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/vm"
+)
+
+// ToolboxDataSink implements the DataSource interface for mounted devices
```
matthewavery: s/datasource/datasink

lib/portlayer/storage/vsphere/toolbox_datasink.go
```diff
@@ -0,0 +1,68 @@
...
+type ToolboxDataSink struct {
+	ID    string
+	VM    *vm.VirtualMachine
+	Clean func()
+}
+
+// Source returns the data source associated with the DataSource
```
matthewavery: s/datasource/datasink

lib/portlayer/storage/vsphere/toolbox_datasink.go
```diff
@@ -0,0 +1,68 @@
...
+func (t *ToolboxDataSink) Sink() interface{} {
+	return t.VM
+}
+
+// Import writes `data` to the data source associated with this DataSource
```
matthewavery: ditto

pkg/vsphere/disk/vmdk.go
```diff
@@ -0,0 +1,64 @@
...
+
+	mountPath, err := v.AttachAndMount(op, dsPath, persistent)
+	return mountPath, cleanFunc, err
+}
+
+func LockedVMDKFilter(vm *mo.VirtualMachine) bool {
+	return vm.Runtime.PowerState == types.VirtualMachinePowerStatePoweredOn
+}
+
+func IsLockedError(err error) bool {
```
matthewavery: Does this get called anywhere currently?

vendor/github.com/vmware/govmomi/object/datastore_file.go
```diff
@@ -344,9 +345,10 @@ func (f *DatastoreFile) TailFunc(lines int, include func(line int, message strin
 }
 
 type followDatastoreFile struct {
```
matthewavery: Is this a change in upstream govmomi? or should this also be in a PR for govmomi? @dougm

matthewavery:
@jzt @emlin and I all reviewed this at once together and left our comments below. I do not think we saw any blockers for the merging of this PR. We left comments about potential house keeping and suggestions only.

cmd/tether/ops_linux.go
```diff
@@ -118,16 +118,8 @@ func (t *operations) Setup(sink tether.Config) error {
 		return err
 	}
 
-	// symlink /etc/mtab to /proc/mounts
-	var err error
-	if err = tether.Sys.Syscall.Symlink("/proc/mounts", "/etc/mtab"); err != nil {
```
caglar10ur: any reason why we are dropping this?

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -46,8 +48,11 @@ import (
 	"time"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/go-openapi/runtime"
+	rc "github.com/go-openapi/runtime/client"
 	"github.com/go-openapi/strfmt"
 	"github.com/google/uuid"
+	httpclient "github.com/mreiferson/go-httpclient"
```
caglar10ur: why are we using this?

lib/archive/unpack.go
matthewavery: in this case you are actually targeting a file source to a file destination. So you are requesting that the file in the container be overwritten. I am testing this now, docker may actually reject if the file exists, not 100% sure on that. But otherwise this is definitely a use case for cp.

lib/apiservers/engine/backends/container_proxy.go
caglar10ur: Can you pass ctx down to this function? That way you can use the same context for the below call (ArchiveExportReader)

lib/archive/diff_test.go
hickeng: This should all be fixed along with the final updates to the filterspec - updated Skip comment to note why it's skipped.

lib/archive/diff.go
hickeng: We do bail out - this will trip us into the next iteration of the loop which checks context.Err()
We do end up writing the archive terminator, which maybe makes it unclear that it's a truncated archive - updated to suppress tar closure.

lib/archive/unpack.go
hickeng: @morris-jason `docker cp a.txt cid:/tmp/b.txt` should end up with the file being called `b.txt` - this is a rename behaviour but can still need to overwrite.

lib/archive/unpack.go
hickeng: Don't assume there are todo's everywhere where filterspec updates are needed - that needs vetting and testing in the filterspec PR.

lib/archive/unpack.go
hickeng: if you cannot create a symlink then it's an operational error. Under no circumstances should it be transformed into a regular file.

lib/portlayer/storage/vsphere/export.go
hickeng: tweaking isn't complete - there's almost certainly extra error checking to be done, however I'm okay leaving that with @morris-jason's PR.

lib/portlayer/storage/vsphere/export.go
hickeng: yes - see @morris-jason PR

lib/portlayer/storage/vsphere/image.go
hickeng: maybe - we need to track down where/why sbin is being created first.

pkg/vsphere/disk/vmdk.go
hickeng: yes - in all of the NewDataSource/NewDataSink calls (guru is your friend).

vendor/github.com/vmware/govmomi/object/datastore_file.go
hickeng: yes it is.

cmd/tether/ops_linux.go
hickeng: It's moved into the base image instead of created for every container so it doesn't show up in diffs.

lib/apiservers/engine/backends/container_proxy.go
hickeng: I have no idea - the curse of this being an integration of various changes.

hickeng: ah - this was pull in by the revert of hasan's fix.

PR #5708, opened 2017-07-13
Removing specially created clients for docker logs and stats [full ci]

### LINKED ISSUE #5685: VIC Port layer reporting socket: too many open files
When using Admiral with VCH endpoint seeing messages in the logs similar to:

**Port-layer.log**
ServerFaultCode: A specified parameter was not correct: spec.url

accept4: too many open files; retrying in 5ms

**Docker-personality.log**
ERROR Handler for GET /v1.21/volumes returned error: error from portlayer server: Get http://127.0.0.1:2377/storage/volumes: dial tcp 127.0.0.1:2377: socket: too many open files

ERROR Handler for GET /v1.21/containers/json returned error: Error invoking GetContainerList: Get http://127.0.0.1:2377/containers/list?all=true: dial tcp 127.0.0.1:2377: socket: too many open files

**Version**: `1.1.1`

bug-ref 1912860

hmahmood:
Fixes #5685

### DIFF
--- a/lib/apiservers/engine/backends/backends.go
+++ b/lib/apiservers/engine/backends/backends.go
@@ -37,6 +37,7 @@ import (
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/scopes"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/storage"
 	"github.com/vmware/vic/lib/apiservers/portlayer/models"
+	"github.com/vmware/vic/lib/archive"
 	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/lib/config/dynamic"
 	"github.com/vmware/vic/lib/imagec"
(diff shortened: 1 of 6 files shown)


lib/apiservers/engine/backends/backends.go
```diff
@@ -109,6 +110,8 @@ func Init(portLayerAddr, product string, config *config.VirtualContainerHostConf
 	}
 
 	t := rc.New(portLayerAddr, "/", []string{"http"})
+	t.Consumers["application/x-tar"] = archive.TarConsumer()
```
hickeng: Looks like the original also included "application/x-gzip" mime type.

hmahmood: But the port layer does not have an endpoint that either consumes or produces that mime-type, so it looks its now being used.

corrieb:
@hmahmood Does this fix 5685 or does it just fix one element of that bug? It would be worth explaining how this fixes the "too many open files" problem, because it's not obvious from the commit.

PR #5719, opened 2017-07-14
docker commit

emlin:
This code is still wip, and do not compile without #5700. 
Open this PR for manual test and review

### DIFF
--- a/lib/apiservers/engine/backends/cache/image_cache.go
+++ b/lib/apiservers/engine/backends/cache/image_cache.go
@@ -217,14 +217,8 @@ func (ic *ICache) Add(imageConfig *metadata.ImageConfig) error {
 	ic.m.Lock()
 	defer ic.m.Unlock()
 
-	// Normalize the name stored in imageConfig using Docker's reference code
-	ref, err := reference.WithName(imageConfig.Name)
-	if err != nil {
-		return fmt.Errorf("error trying to create reference from %s: %s", imageConfig.Name, err)
-	}
-
 	imageID := prefixImageID(imageConfig.ImageID)
-	err = ic.iDIndex.Add(imageConfig.ImageID)
+	err := ic.iDIndex.Add(imageConfig.ImageID)
 	if err != nil && !os.IsExist(err) {
 		return fmt.Errorf("error adding image %s to index: %s", imageID, err)
 	}
@@ -233,6 +227,17 @@ func (ic *ICache) Add(imageConfig *metadata.ImageConfig) error {
 
 	ic.cacheByID[imageID] = imageConfig
 	ic.dirty = true
+
+	if imageConfig.Name == "" {
+		log.Debugf("Image %s has no name", imageID)
+		return nil
+	}
+	// construct reference after image is added into cacheByID, to make sure image without name can be added
+	// Normalize the name stored in imageConfig using Docker's reference code
+	ref, err := reference.WithName(imageConfig.Name)
+	if err != nil {
+		return fmt.Errorf("error trying to create reference from %s: %s", imageConfig.Name, err)
+	}
 
 	for _, tag := range imageConfig.Tags {
 		ref, err = reference.WithTag(ref, tag)
--- a/lib/apiservers/engine/backends/cache/image_cache.go
+++ b/lib/apiservers/engine/backends/cache/image_cache.go
@@ -232,7 +232,10 @@ func (ic *ICache) Add(imageConfig *metadata.ImageConfig) error {
 		log.Debugf("Image %s has no name", imageID)
 		return nil
 	}
-	// construct reference after image is added into cacheByID, to make sure image without name can be added
+
+	// Construct a reference after the image is added into cacheByID so that an image
+	// without a name can at least be added to cacheByID.
+
 	// Normalize the name stored in imageConfig using Docker's reference code
 	ref, err := reference.WithName(imageConfig.Name)
 	if err != nil {
--- a/lib/apiservers/engine/backends/commit.go
+++ b/lib/apiservers/engine/backends/commit.go
@@ -15,7 +15,6 @@
 package backends
 
 import (
-	"archive/tar"
 	"compress/gzip"
 	"context"
 	"crypto/sha256"
@@ -262,39 +261,10 @@ func downloadDiff(rc io.ReadCloser, containerID string, options imagec.Options) 
 	diffID := digest.NewDigestFromBytes(digest.SHA256, diffIDSum)
 	log.Debugf("container %s diff id: %s", containerID, diffID.String())
 
-	layerFile, err := os.Open(string(tmpLayerFileName))
-	if err != nil {
-		return nil, err
-	}
-	defer layerFile.Close()
-
-	decompressed, err := gzip.NewReader(layerFile)
-	if err != nil {
-		return nil, err
-	}
-	defer decompressed.Close()
-
-	// get a tar reader
-	tr := tar.NewReader(decompressed)
-
-	// iterate through tar headers to get file sizes
-	var layerSize int64
-	for {
-		tarHeader, terr := tr.Next()
-		if terr == io.EOF {
-			break
-		}
-		if terr != nil {
-			err = terr
-			return nil, err
-		}
-		layerSize += tarHeader.Size
-	}
-	log.Debugf("layer size: %d, copy size: %d", layerSize, size)
-	if layerSize == 0 {
+	if size == 0 {
 		diffID = digest.Digest(dockerLayer.DigestSHA256EmptyTar)
 	}
-	log.Debugf("container %s size: %d", containerID, layerSize)
+	log.Debugf("container %s diff size: %d", containerID, size)
 
 	// Ensure the parent directory exists
 	destination := path.Join(imagec.DestinationDirectory(options), layerID)
@@ -318,7 +288,7 @@ func downloadDiff(rc io.ReadCloser, containerID string, options imagec.Options) 
 		Layer: imagec.FSLayer{
 			BlobSum: blobSum.String(),
 		},
-		Size: layerSize,
+		Size: size,
 	}
 	return lm, nil
 }
@@ -353,12 +323,12 @@ func compressDiffToTmpFile(rc io.ReadCloser, containerID string) (string, []byte
 	}
 
 	// compress tar file using gzip and calculate blobsum and diffID all together using multi writer
-	blobSum := sha256.New()
-	diffID := sha256.New()
-	compressedMW := io.MultiWriter(out, blobSum)
+	gzSum := sha256.New()
+	tarSum := sha256.New()
+	compressedMW := io.MultiWriter(out, gzSum)
 
 	gzWriter = gzip.NewWriter(compressedMW)
-	tarMW := io.MultiWriter(gzWriter, diffID)
+	tarMW := io.MultiWriter(gzWriter, tarSum)
 	size, err = io.Copy(tarMW, rc)
 	if err != nil {
 		log.Errorf("failed to stream to file: %s", err)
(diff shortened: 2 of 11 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -1274,8 +1275,53 @@ func (c *Container) ContainerWait(name string, timeout time.Duration) (int, erro
...
-func (c *Container) ContainerChanges(name string) ([]archive.Change, error) {
-	return make([]archive.Change, 0, 0), fmt.Errorf("%s does not yet implement ontainerChanges", ProductName())
+func (c *Container) ContainerChanges(name string) ([]docker.Change, error) {
+	defer trace.End(trace.Begin(name))
+
+	vc := cache.ContainerCache().GetContainer(name)
+	if vc == nil {
+		return nil, NotFoundError(name)
+	}
+	log.Debugf("Found %q in cache as %q", name, vc.ContainerID)
```
sflxn: we can remove this, right?

lib/apiservers/engine/backends/commit.go
```diff
@@ -0,0 +1,472 @@
...
+	c, err := containerEngine.ContainerInspect(name, false, "")
+	if err != nil {
+		return "", InternalServerError(err.Error())
+	}
+	container, ok := c.(*types.ContainerJSON)
+	if !ok {
+		return "", InternalServerError(fmt.Sprintf("Container type assertion failed"))
+	}
+	if container.State.Running || container.State.Restarting {
+		return "", ConflictError(fmt.Sprintf("%s does not support commit of a running container", ProductName()))
```
sflxn: I think we should say "does not yet"

lib/apiservers/engine/backends/commit.go
```diff
@@ -0,0 +1,472 @@
...
+	if err != nil {
+		return "", err
+	}
+	imagec.LayerCache().Add(layer)
+
+	var layers []*imagec.ImageWithMeta
+
+	layers = append(layers, layer)
+	lm := layer
+	for pl := lm.Parent; pl != imagec.ScratchLayerID; pl = lm.Parent {
```
sflxn: This for loop is fine, but it was strange to look at initially.  I'd suggest the following.  It's up to you whether you use it.

```
	pl := lm.Parent
	for {
 		if pl == imageC.ScratchLayerID {
			break
 		}

 		// populate manifest layer with existing cached data
 		if lm, err = imagec.LayerCache().Get(pl); err != nil {
 			return "", InternalServerError(fmt.Sprintf("Failed to get parent image layer %s: %s", pl, err))
 		}
 		layers = append(layers, lm)

 		pl = lm.Parent
 	}
```

lib/apiservers/engine/backends/commit.go
```diff
@@ -0,0 +1,472 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer decompressed.Close()
+
+	// get a tar reader
+	tr := tar.NewReader(decompressed)
+
+	// iterate through tar headers to get file sizes
+	var layerSize int64
```
sflxn: I don't believe you need to do all this to get the size of the uncompressed tar file.  In your compressDiffToTmpFile() function, you're doing an io.Copy.  The first return from that is bytes written for the uncompressed tar file.

I'm doing something similar in push.  We can refactor them together later.

https://github.com/sflxn/vic/blob/docker-push/lib/imagec/imagec.go#L915

lib/apiservers/engine/backends/commit.go
```diff
@@ -0,0 +1,472 @@
...
+		Layer: imagec.FSLayer{
+			BlobSum: blobSum.String(),
+		},
+		Size: layerSize,
+	}
+	return lm, nil
+}
+
+// compressDiffToTmpFile will write stream to temp file, and return temp file name and tar file checksum, compressed file checksum
+func compressDiffToTmpFile(rc io.ReadCloser, containerID string) (string, []byte, []byte, error) {
```
sflxn: Why not just return Digest instead of []byte?

lib/apiservers/engine/backends/commit.go
```diff
@@ -0,0 +1,472 @@
...
+	}
+
+	// compress tar file using gzip and calculate blobsum and diffID all together using multi writer
+	blobSum := sha256.New()
+	diffID := sha256.New()
+	compressedMW := io.MultiWriter(out, blobSum)
+
+	gzWriter = gzip.NewWriter(compressedMW)
+	tarMW := io.MultiWriter(gzWriter, diffID)
+	_, err = io.Copy(tarMW, rc)
```
sflxn: This is where I was saying you can get the size of the uncompressed tar in my comments above.

lib/apiservers/engine/backends/commit.go
```diff
@@ -0,0 +1,472 @@
...
+	if err != nil {
+		log.Errorf("failed to stream to file: %s", err)
+		return "", nil, nil, err
+	}
+
+	// close writer before calculate checksum
+	fileName := out.Name()
+	gzWriter.Flush()
+	cleanup()
```
sflxn: Should you call cleanup() twice?  You have a defer on it above.  I would also move the Flush operation into the defer function.

lib/apiservers/engine/backends/container.go
```diff
@@ -1274,8 +1275,53 @@ func (c *Container) ContainerWait(name string, timeout time.Duration) (int, erro
...
+	changes := []docker.Change{}
+
+	tarFile := tar.NewReader(r)
+
+	defer r.Close()
+
+	for {
+		hdr, err := tarFile.Next()
+		if err == io.EOF {
+			log.Infof("Got EOF!! BAILING")
```
sflxn: debug msg?  remove?

lib/apiservers/engine/backends/container.go
```diff
@@ -1274,8 +1275,53 @@ func (c *Container) ContainerWait(name string, timeout time.Duration) (int, erro
...
+			break
+		}
+		if err != nil {
+			return []docker.Change{}, InternalServerError(err.Error())
+		}
+		log.Infof("Got header %s", hdr.Name)
```
sflxn: debug msg?  remove?

sflxn:
Overall, it's fine.  I would like to move a lot of the commit code from the backend into imagec, but we can do the refactor later.  You and I have some duplicate code also and we can refactor once docker push gets in.

lib/apiservers/engine/backends/container.go
emlin: right, I'll remove it. not part of my change though 😄

lib/apiservers/engine/backends/commit.go
emlin: hmm.. is this for loop better? feels like my code has less lines. And it's less likely to have wrong assignment for variables. 
Anyway, I gonna keep my code for now. 😄

lib/apiservers/engine/backends/commit.go
emlin: okay, will update it

lib/apiservers/engine/backends/commit.go
emlin: I need to flush and close writer, before call `diffID.Sum(nil), blobSum.Sum(nil)` to get correct sha256. So I set the writer to nil to avoid double close() in cleanup method

lib/apiservers/engine/backends/container.go
emlin: hmm.. @hickeng it's from your code, should we change that to `Debugf`?

lib/apiservers/engine/backends/commit.go
sflxn: I see what you're doing then.  That's fine.

lib/apiservers/engine/backends/commit.go
emlin: so far, I feel that this method is just compressing tar file to a gz file, and could get some knowledge about the checksum, and file size.
the caller should be responsible for how to use those values. So I just renamed the variables.

emlin:
> I would like to move a lot of the commit code from the backend into imagec, but we can do the refactor later. You and I have some duplicate code also and we can refactor once docker push gets in.

Yeah, right. Let's do more refactor after commit and push all get in

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -234,6 +228,17 @@ func (ic *ICache) Add(imageConfig *metadata.ImageConfig) error {
 	ic.cacheByID[imageID] = imageConfig
 	ic.dirty = true
 
+	if imageConfig.Name == "" {
+		log.Debugf("Image %s has no name", imageID)
+		return nil
+	}
+	// construct reference after image is added into cacheByID, to make sure image without name can be added
```
vburenin: ```construct a reference after an image is added into the id cache to insure the no name image can be added.``` please correct me too on the grammar side.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -674,7 +674,7 @@ func (c *ContainerProxy) StreamContainerStats(ctx context.Context, config *conve
 	return nil
 }
 
-func (c *ContainerProxy) GetContainerChanges(ctx context.Context, vc *viccontainer.VicContainer) (io.ReadCloser, error) {
+func (c *ContainerProxy) GetContainerChanges(ctx context.Context, vc *viccontainer.VicContainer, data bool) (io.ReadCloser, error) {
```
vburenin: Could you add a description what that data variable is for?

lib/imagec/download.go
```diff
@@ -196,7 +196,7 @@ func (ldm *LayerDownloader) DownloadLayers(ctx context.Context, ic *ImageC) erro
 			return err
 		}
 	} else {
-		if err := updateRepositoryCache(ic); err != nil {
+		if err := UpdateRepositoryCache(ic); err != nil {
```
vburenin: I would even rename it to UpdateRepoCache.

lib/imagec/imagec.go
```diff
@@ -310,7 +321,7 @@ func (ic *ImageC) WriteImageBlob(image *ImageWithMeta, progressOutput progress.O
 	destination := DestinationDirectory(ic.Options)
 
 	id := image.Image.ID
-	log.Infof("Path: %s", path.Join(destination, id, id+".targ"))
+	log.Infof("Path: %s", path.Join(destination, id, id+".tar"))
```
vburenin: Could you pre-calculate a value into a separate variable?

lib/apiservers/engine/backends/commit_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// Create a new tar archive.
+	tw := tar.NewWriter(buf)
+
+	// Add some files to the archive.
+	var files = []struct {
+		Name, Body string
+	}{
+		{"readme.txt", "This archive contains some text files."},
+		{"gopher.txt", "Gopher names:\nGeorge\nGeoffrey\nGonzo"},
```
vburenin: I love the file name ;)

lib/apiservers/engine/backends/commit_test.go
```diff
@@ -0,0 +1,110 @@
...
+	"path"
+	"testing"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/docker/docker/api/types/backend"
+
+	"github.com/vmware/vic/lib/imagec"
+)
+
+func getMockReader() (io.ReadCloser, error) {
```
vburenin: I would pass a testing t here and would use that logging instead. fatal call in tests is kinda rough.

lib/apiservers/engine/backends/cache/image_cache.go
emlin: @anchal-agrawal could you please help on this? 😄

anchal-agrawal: Per my understanding of our intention here, the comment below should suffice.

```
// Construct a reference after the image is added into cacheByID so that an image 
// without a name can at least be added to cacheByID.
```

sflxn:
lgtm

emlin:
@sflxn @jzt 
After your approval, revert to get tar file size by accumulate each file size in tar file header, because the io.Copy returned value contains the padding zero in tar file.
Let me know if you find any problem there

PR #5720, opened 2017-07-14
Add lib/archive GenerateFilterSpec implementation

matthewavery:
GenerateFilterSpec can be used to create a filter spec with the appropriate `Rebase` and `Strip` targets. To generate the appropriate spec 4 aspects are need. These are described below:

- archive target path: This is the path which you will either unpack to or pack from. 
- mount point: this is the device mount point for the target runtime filesystem. (for the read write layer this would be "/", for some volume an example would be "/mnt/vol/A" or "/mydata". etc...)
- PrimaryTarget: this is a boolean indicating whether the given device path is the primary device for the Copy operation. The introspective question for this is "Am I mounted before the archive target path and am I involved with this copy operation?"
- operation direction: This is to designate whether this is a Copy From or a Copy To. Constants are included in the portlayer archive package. These are `archive.CopyFrom` and `archive.CopyTo`

based on these 4 inputs we are able to determine the appropriate Rebase and Strip path.

### DIFF
--- a/lib/apiservers/engine/backends/archive.go
+++ b/lib/apiservers/engine/backends/archive.go
@@ -229,8 +229,8 @@ func NewArchiveStreamWriterMap(mounts []types.MountPoint, dest string) *ArchiveS
 		// file data.txt from local /mnt/A/data.txt will come to the persona as mnt/A/data.txt.
 		// Here, we must tell the portlayer to remove "mnt/A".  The key to determining whether to
 		// strip "A" or "mnt/A" is based on the container destination path.
-		isPrimary := !strings.Contains(aw.mountPoint.Destination, containerDestPath) || aw.mountPoint.Destination == containerDestPath
-		aw.filterSpec = vicarchive.GenerateFilterSpec(containerDestPath, aw.mountPoint.Destination, isPrimary, vicarchive.CopyTo)
+		isPrimary := !strings.Contains(aw.mountPoint.Destination, dest) || aw.mountPoint.Destination == dest
+		aw.filterSpec = vicarchive.GenerateFilterSpec(dest, aw.mountPoint.Destination, isPrimary, vicarchive.CopyTo)
 
 		writerMap.prefixTrie.Insert(patricia.Prefix(m.Destination), &aw)
 	}
--- a/lib/apiservers/engine/backends/archive.go
+++ b/lib/apiservers/engine/backends/archive.go
@@ -229,18 +229,9 @@ func NewArchiveStreamWriterMap(mounts []types.MountPoint, dest string) *ArchiveS
 		// file data.txt from local /mnt/A/data.txt will come to the persona as mnt/A/data.txt.
 		// Here, we must tell the portlayer to remove "mnt/A".  The key to determining whether to
 		// strip "A" or "mnt/A" is based on the container destination path.
-
-		// hickeng: the current logic is strip, then rebase for export - more efficient to collapse
-		// any reducdency here than apply to every tar entry when packing but skipping for now.
-		aw.filterSpec.StripPath = aw.mountPoint.Destination
-		aw.filterSpec.RebasePath = strings.TrimPrefix(dest, aw.mountPoint.Destination)
-
-		// if containerDestPath != "/" && strings.HasPrefix(aw.mountPoint.Destination, containerDestPath) {
-		// 	aw.filterSpec.StripPath = strings.TrimPrefix(aw.mountPoint.Destination, containerDestPath)
-		// }
-
-		aw.filterSpec.Exclusions = make(map[string]struct{})
-		aw.filterSpec.Inclusions = make(map[string]struct{})
+		if containerDestPath != "/" && strings.HasPrefix(aw.mountPoint.Destination, containerDestPath) {
+			aw.filterSpec = vicarchive.GenerateFilterSpec(containerDestPath, aw.mountPoint, !strings.Contains(aw.mountPoint, containerDestPath), vicarchive.CopyTo)
+		}
 
 		writerMap.prefixTrie.Insert(patricia.Prefix(m.Destination), &aw)
 	}
@@ -273,14 +264,9 @@ func NewArchiveStreamReaderMap(mounts []types.MountPoint, dest string) *ArchiveS
 		//
 		// Neither the volume nor the storage portlayer knows about /mnt/A.  The persona must tell
 		// the portlayer to rebase all files from this volume to the /mnt/A/ in the final tar stream.
-
-		// hickeng: the current logic is rebase, then strip for export - more efficient to collapse
-		// any reducdency here than apply to every tar entry when packing but skipping for now.
-		ar.filterSpec.RebasePath = ar.mountPoint.Destination
-		ar.filterSpec.StripPath = path.Join("/", path.Dir(dest))
-
-		ar.filterSpec.Exclusions = make(map[string]struct{})
-		ar.filterSpec.Inclusions = make(map[string]struct{})
+		if ar.mountPoint.Destination != "/" {
+			ar.filterSpec = vicarchive.GenerateFilterSpec(containerDestPath, ar.mountPoint.Destination, !strings.Contains(ar.mountPoint.Destination, containerDestPath), vicarchive.CopyTo)
+		}
 
 		readerMap.prefixTrie.Insert(patricia.Prefix(m.Destination), &ar)
 	}
--- a/lib/apiservers/engine/backends/archive.go
+++ b/lib/apiservers/engine/backends/archive.go
@@ -265,7 +265,7 @@ func NewArchiveStreamReaderMap(mounts []types.MountPoint, dest string) *ArchiveS
 		// Neither the volume nor the storage portlayer knows about /mnt/A.  The persona must tell
 		// the portlayer to rebase all files from this volume to the /mnt/A/ in the final tar stream.
 		if ar.mountPoint.Destination != "/" {
-			ar.filterSpec = vicarchive.GenerateFilterSpec(containerDestPath, ar.mountPoint.Destination, !strings.Contains(ar.mountPoint.Destination, containerDestPath), vicarchive.CopyTo)
+			ar.filterSpec = vicarchive.GenerateFilterSpec(containerDestPath, ar.mountPoint.Destination, !strings.Contains(ar.mountPoint.Destination, containerDestPath), vicarchive.CopyFrom)
 		}
 
 		readerMap.prefixTrie.Insert(patricia.Prefix(m.Destination), &ar)
--- a/lib/apiservers/engine/backends/archive.go
+++ b/lib/apiservers/engine/backends/archive.go
@@ -229,9 +229,8 @@ func NewArchiveStreamWriterMap(mounts []types.MountPoint, dest string) *ArchiveS
 		// file data.txt from local /mnt/A/data.txt will come to the persona as mnt/A/data.txt.
 		// Here, we must tell the portlayer to remove "mnt/A".  The key to determining whether to
 		// strip "A" or "mnt/A" is based on the container destination path.
-		if containerDestPath != "/" && strings.HasPrefix(aw.mountPoint.Destination, containerDestPath) {
-			aw.filterSpec = vicarchive.GenerateFilterSpec(containerDestPath, aw.mountPoint, !strings.Contains(aw.mountPoint, containerDestPath), vicarchive.CopyTo)
-		}
+		isPrimary := !strings.Contains(aw.mountPoint.Destination, containerDestPath) || aw.mountPoint.Destination == containerDestPath
+		aw.filterSpec = vicarchive.GenerateFilterSpec(containerDestPath, aw.mountPoint.Destination, isPrimary, vicarchive.CopyTo)
 
 		writerMap.prefixTrie.Insert(patricia.Prefix(m.Destination), &aw)
 	}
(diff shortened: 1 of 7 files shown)


lib/apiservers/engine/backends/archive.go
```diff
@@ -288,12 +285,9 @@ func NewArchiveStreamReaderMap(mounts []types.MountPoint) *ArchiveStreamReaderMa
 		// Neither the volume nor the storage portlayer knows about /mnt/A.  The persona must tell
 		// the portlayer to rebase all files from this volume to the /mnt/A/ in the final tar stream.
 		if ar.mountPoint.Destination != "/" {
-			ar.filterSpec.RebasePath = ar.mountPoint.Destination
+			ar.filterSpec = vicarchive.GenerateFilterSpec(containerDestPath, ar.mountPoint.Destination, !strings.Contains(ar.mountPoint.Destination, containerDestPath), vicarchive.CopyTo)
```
morris-jason: vicarchive.copyFrom?

lib/apiservers/engine/backends/archive.go
```diff
@@ -250,12 +250,9 @@ func NewArchiveStreamWriterMap(mounts []types.MountPoint, containerDestPath stri
 		// Here, we must tell the portlayer to remove "mnt/A".  The key to determining whether to
 		// strip "A" or "mnt/A" is based on the container destination path.
 		if containerDestPath != "/" && strings.HasPrefix(aw.mountPoint.Destination, containerDestPath) {
-			aw.filterSpec.StripPath = strings.TrimPrefix(aw.mountPoint.Destination, containerDestPath)
+			aw.filterSpec = vicarchive.GenerateFilterSpec(containerDestPath, aw.mountPoint, !strings.Contains(aw.mountPoint, containerDestPath), vicarchive.CopyTo)
```
morris-jason: primaryTarget = !strings.Contains(aw.mountPoint, containerDestPath) is incorrect. It's more complicated than this.
Ie given mountpoints: 
```
/mnt/volA/
/mnt/volA/volC/
/mnt/VolB/
```
container destination: /mnt/volA/volC/fileABCD.txt.
The primary mount is /mnt/volA/volC. 
But your code will give
```
/mnt/volA/ - primary target = !strings.Contains(aw.mountPoint, containerDestPath) = false
/mnt/volA/volC/ - primary target = !strings.Contains(aw.mountPoint, containerDestPath) = false
/mnt/VolB/ - primary target = !strings.Contains(aw.mountPoint, containerDestPath) = true
```

lib/archive/util.go
```diff
@@ -0,0 +1,90 @@
...
+// limitations under the License.
+
+package archive
+
+import (
+	"path/filepath"
+	"strings"
+)
+
+const (
```
morris-jason: This and the GenerateFilterSpec function seem unnecessarily complex. I think the consumer should just call GenerateCopy<direction>FilterSpec functions since that indicates direction.

lib/archive/util.go
```diff
@@ -0,0 +1,90 @@
...
+	filter.Exclusions = make(map[string]struct{})
+	filter.Inclusions = make(map[string]struct{})
+	return filter
+}
+
+func generateCopyFromFilterSpec(targetPath string, mountPoint string, primaryTarget bool) FilterSpec {
+	var filter FilterSpec
+
+	//we only need the right most element.
+	_, first := filepath.Split(targetPath)
```
morris-jason: basename?

lib/apiservers/engine/backends/archive.go
matthewavery: woops yeah actually. it needs to be more than that for sure since aw.mountpoint is not correct. Sorry about that for some reason my IDE did not catch that. Additionally, it must only be called against mounts that are involved with the operation. if you look any one of those targets that you supplied would be a primary target, but it you are doing a `write` then ---> /mnt/volA and /mnt/VolB are not involved in the operation at all and thus a writer/reader should never be initialized for them. if you are doing a read you get the same thing. I might have made a false assumption about where the filter spec should occur.

lib/apiservers/engine/backends/archive.go
morris-jason: resolve copy paste error.

matthewavery: haha thanks *doh*

lib/archive/util.go
matthewavery: For this I left the specific functions enexported so that callers understand that the direction makes a big difference. If others feel that this makes it too complicated then we can change it.

lib/archive/util.go
matthewavery: I will change it to `copyPath`

lib/archive/util.go
```diff
@@ -0,0 +1,118 @@
...
+const (
+	// CopyTo is used to indicate that the desired filter spec is for a CopyTo direction
+	CopyTo = true
+
+	// CopyFrom is used to indicate that the desired filter spec is for the CopyFrom direction
+	CopyFrom = false
+)
+
+// GenerateFilterSpec will populate the appropriate relative Rebase and Strip paths based on the supplied scenarios. Inclusion/Exclusion should be constructed separately. Please also note that any mount that exists before the copy target that is not primary and comes before the primary target will have a bogus filterspec, since it would not be written or read to.
+func GenerateFilterSpec(copyPath string, mountPoint string, primaryTarget bool, direction bool) FilterSpec {
```
emlin: For my understanding, this filter spec is requited and becomes more and more complicated, is because we're not mounting everything one time, and do one diff for the whole file system like what docker did. In that way, we could just have one rebase path, and everything works.
But we have to mount container r/w layer, and volumes one after another. If we're copying to multiple devices, with same parent mount point, the strip and rebase becomes different for different device.
And for the same reason, the primaryTarget is invented, just is used to generate strip and rebase string. This variable is not used by the filterSpec consumer, and could be derived from the mount points, copyTo/From target and file path.

So here my suggestion is that, change the signature to accept copySource, copyTarget, direction, and all the mount points, instead of only one, and return one filterspec per mount point.

In this method, the inclusion/exclusion should be filled based on the input. If there could have some like the strip slash, etc.

If user have their own inclusion/exclusion specification, this method could have that information as input as well, and fill the Inclusion/Exclusion in FilterSpec.

We should not have the caller aware of too many internal details, e.g. PrimaryTarget, to generate one filter spec.

lib/archive/util.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	// non primary target, this implies that the asset header has part of the mount point path in it. We must strip out that part since the non primary target will be mounted and be looking at the world from it's own root "/"
+	filter.RebasePath = ""
+	filter.StripPath = removeLeadingSlash(strings.TrimPrefix(mountPoint, copyPath))
+
+	return filter
+}
+
+func AddMountInclusionsExclusions(currentMount string, filter *FilterSpec, mounts []string, copyTarget string) error {
```
emlin: As the comment I mentioned above, we could this kind of interface to add more Inclusion/Exclusion into the FilterSpec, but that should not the default one could be derived from mount points.
That should be specified by the exclusion/inclusion parameters.

lib/archive/util.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func generateCopyFromFilterSpec(copyPath string, mountPoint string, primaryTarget bool) FilterSpec {
+	var filter FilterSpec
+
+	//we only need the right most element.
+	_, first := filepath.Split(copyPath)
+
+	// primary target was provided so we wil need to split the target and take the right most element for the rebase.
+	// then strip set the strip as the target path.
```
emlin: strip is set the target path - mount point, because in PL the mount point would not show up in the tart file path

lib/archive/util.go
emlin: still for the primaryTarget. Can you tell that based on who is longer between the mountPoint and copyPath?

lib/archive/util.go
```diff
@@ -0,0 +1,118 @@
...
+	copyPath = strings.TrimSuffix(copyPath, "/")
+
+	// Note I know they are just booleans, if that changes then this statement will not need to.
+	if direction == CopyTo {
+		filter = generateCopyToFilterSpec(copyPath, mountPoint, primaryTarget)
+	} else {
+		filter = generateCopyFromFilterSpec(copyPath, mountPoint, primaryTarget)
+	}
+
+	filter.Exclusions = make(map[string]struct{})
```
emlin: could have this in the NewFilterSpec method

lib/archive/util.go
```diff
@@ -0,0 +1,118 @@
...
+
+	return filter
+}
+
+func AddMountInclusionsExclusions(currentMount string, filter *FilterSpec, mounts []string, copyTarget string) error {
+	if filter == nil {
+		return fmt.Errorf("filterSpec for (%s) was nil, cannot add exclusions or inclusions", currentMount)
+	}
+
+	if filter.Exclusions == nil || filter.Inclusions == nil {
```
emlin: with this set by default, you don't have to check if they are nil

lib/archive/util.go
```diff
@@ -0,0 +1,118 @@
...
+	if primaryTarget {
+		filter.RebasePath = removeLeadingSlash(first)
+		filter.StripPath = removeLeadingSlash(strings.TrimPrefix(copyPath, mountPoint))
+		return filter
+	}
+
+	// non primary target was provided. in this case we will need rebase to include the right most member of the target(or "/") joined to the front of the mountPath - the target path. 3
+	filter.RebasePath = removeLeadingSlash(filepath.Join(first, strings.TrimPrefix(mountPoint, copyPath)))
+	filter.StripPath = ""
+	return filter
```
emlin: what happens if the mount point and copy target does not have a common prefix? I think this method should test that scenario

lib/archive/util.go
```diff
@@ -0,0 +1,118 @@
...
+	if filter == nil {
+		return fmt.Errorf("filterSpec for (%s) was nil, cannot add exclusions or inclusions", currentMount)
+	}
+
+	if filter.Exclusions == nil || filter.Inclusions == nil {
+		return fmt.Errorf("either the inclusions or exclusions map was nil for (%s)", currentMount)
+	}
+
+	if strings.HasPrefix(copyTarget, currentMount) {
+		filter.Inclusions[removeLeadingSlash(strings.TrimPrefix(copyTarget, currentMount))] = struct{}{}
```
emlin: how the inclusion is defined? If no inclusion and exclusion, what is the default value? Include? If that's the case, feels like you don't have to add this inclusion

lib/archive/util.go
emlin: actually I think I better way is to new map if they are nil, instead of return error

lib/archive/util.go
emlin: this method should be one method in FilterSpec, and the `mounts []string` should be just `exclusion []string`, which is the term understand by the FilterSpec.
Mount should be excluded, but that's the persona's knowledge

lib/archive/util.go
emlin: btw, even you have to have this logic included, this should be part of definition of FilterSpec, instead of here while more exclusion is added.

lib/archive/util.go
```diff
@@ -0,0 +1,118 @@
...
+		filter.RebasePath = removeLeadingSlash(strings.TrimPrefix(copyPath, mountPoint))
+		filter.StripPath = ""
+		return filter
+	}
+
+	// non primary target, this implies that the asset header has part of the mount point path in it. We must strip out that part since the non primary target will be mounted and be looking at the world from it's own root "/"
+	filter.RebasePath = ""
+	filter.StripPath = removeLeadingSlash(strings.TrimPrefix(mountPoint, copyPath))
+
+	return filter
```
emlin: ditto the mismatched mountPoint and copyPath

lib/archive/util.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// primary target was provided so we will need to rebase header assets for this mount to have the target in front for the write.
+	if primaryTarget {
+		filter.RebasePath = removeLeadingSlash(strings.TrimPrefix(copyPath, mountPoint))
+		filter.StripPath = ""
+		return filter
+	}
+
+	// non primary target, this implies that the asset header has part of the mount point path in it. We must strip out that part since the non primary target will be mounted and be looking at the world from it's own root "/"
+	filter.RebasePath = ""
```
emlin: I'm actually wondering if this case is handled properly?
This means the copy path is less than the mount point. You'll need to figure out if the file should go into one volume, after the file is untared.
This function is called in persona, but tar stream is processed in PL, how does this work?

emlin: btw, this case is not tested in the UT

lib/archive/util.go
matthewavery: It will not get trimmed off in this case which means that the mount point was not along the path and should not be asked for by the caller. the mount path either is or is not along the copy path. If it is not along the copy path in any form then it would not be a target of the copyTo or CopyFrom.

lib/archive/util.go
matthewavery: This case is tested in the UT. This is a CopyTo operation so we are writing to a disk. If you are not the primary target of the write then no rebase exists for that tar asset. You will need to remove any of the pathing in the asset leading up to the mount point(including the mount point itself). for instance, you have target `/mnt/vols` and a mount at `/mnt/vols/A`. In the provided stream there is an asset with path `A/file.txt`. In this case you will need to strip `A` from the header only. which is mountpoint minus copypath. but no rebase necessary.

lib/archive/util.go
matthewavery: you will need to trim copyPath from Mountpoint. Here we know that MountPoint is longer than copyPath. and we need only strip from the header the remaining bits leading up to the disk filesystem.

lib/archive/util.go
matthewavery: yes this is determined in the persona currently.

matthewavery:
Please note, This is being merged into the docker-cp feature branch. there are many engineers on this, and that means that some fixes may exist for broken tests in this PR for another change above this PR. Comments for this PR will be adressed.

PR #5726, opened 2017-07-15
Implement Online Datasource

morris-jason:
Fixes 5606, 5715, and 5570.

Supports Online Export/Import using DataSource and DataSink, overriding the default behavior of the toolbox's ArchiveFileHandler. Unpack and Diff are leveraged using DiskLabels in the tether.

### DIFF
--- a/lib/apiservers/engine/backends/archive.go
+++ b/lib/apiservers/engine/backends/archive.go
@@ -20,13 +20,12 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"os"
 	"path"
 	"strings"
 
-	log "github.com/Sirupsen/logrus"
 	"github.com/tchap/go-patricia/patricia"
 
+	"github.com/docker/docker/pkg/archive"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	viccontainer "github.com/vmware/vic/lib/apiservers/engine/backends/container"
 	vicarchive "github.com/vmware/vic/lib/archive"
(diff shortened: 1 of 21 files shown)


lib/apiservers/engine/backends/archive.go
```diff
@@ -248,46 +284,6 @@ func NewArchiveStreamWriterMap(mounts []types.MountPoint, dest string) *ArchiveS
 	return writerMap
 }
 
-// NewArchiveStreamReaderMap creates a new ArchiveStreamReaderMap.  After the call, it contains
```
morris-jason: diff bug.

lib/apiservers/engine/backends/archive.go
```diff
@@ -500,72 +546,6 @@ func (rm *ArchiveStreamReaderMap) FindArchiveReaders(containerSourcePath string)
 	return nodes, nil
 }
 
-func (rm *ArchiveStreamReaderMap) buildFilterSpec(containerSourcePath string, nodes []*ArchiveReader, startingNode *ArchiveReader) error {
```
morris-jason: diff bug

lib/portlayer/storage/data_source.go
```diff
@@ -53,6 +54,46 @@ func (m *MountDataSource) Source() interface{} {
 }
 
 // Export reads data from the associated data source and returns it as a tar archive
+func (m *MountDataSource) Stat(op trace.Operation, spec *archive.FilterSpec) (*FileStat, error) {
```
hickeng: If we're passing in a FilterSpec as the input, we should return a `map[string]*FileStat` for the result - while it's fine for an implementation to return an error stating that it cannot deal with multiple inputs the signature should be consistent between intput/output to allow for other implementation decisions.

morris-jason: Ok. Should I return 422 then if there's more than one path in filterspec Inclusions?

PR #5747, opened 2017-07-18
Remove whiteout prefix in diff output

jzt:
This is a quick fix to the output for deleted files that removes the whiteout prefix:

From 
```
A /NEW_FILE
A /a
A /b
C /etc/
C /etc/bash.bashrc
D /etc/.wh.bindresvport.blacklist
```
To:
```
A /NEW_FILE
A /a
A /b
C /etc/
C /etc/bash.bashrc
D /etc/bindresvport.blacklist
```

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -22,6 +22,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"sync"
@@ -1315,6 +1316,9 @@ func (c *Container) ContainerChanges(name string) ([]docker.Change, error) {
 			change.Kind = docker.ChangeAdd
 		case "D":
 			change.Kind = docker.ChangeDelete
+			path := strings.TrimSuffix(change.Path, "/")
+			p := strings.TrimPrefix(filepath.Base(path), docker.WhiteoutPrefix)
+			change.Path = filepath.Join(filepath.Dir(path), p)
 		case "C":
 			change.Kind = docker.ChangeModify
 		default:


lib/apiservers/engine/backends/container.go
```diff
@@ -1315,6 +1316,8 @@ func (c *Container) ContainerChanges(name string) ([]docker.Change, error) {
 			change.Kind = docker.ChangeAdd
 		case "D":
 			change.Kind = docker.ChangeDelete
+			p := strings.TrimPrefix(filepath.Base(change.Path), docker.WhiteoutPrefix)
+			change.Path = filepath.Join(filepath.Dir(change.Path), p)
```
caglar10ur: should we check first before doing this?

jzt: sorry, check what?

matthewavery: will change.Path ever end with a `/`? I have found that Dir("/path/dir/") ---> "/path/dir" and "". where as Dir("/path/dir/") ---> "/path" and "dir".

jzt: If you mean for the whiteout prefix, we know all changes of type delete will have it in the path as we are the ones who put it there when we build the archive :)

jzt: @matthewavery I don't think so, but I can attempt to strip a trailing `/` to be safe.

caglar10ur: oh wait, this is Delete case already so it is guaranteed to have WhiteoutPrefix in it. 

LGTM

caglar10ur:
LGTM

lib/apiservers/engine/backends/container.go
matthewavery: cool, we could always attempt the strip off of "/". I think it should be fine though. I have seen this show up in weird ways for other parts of VIC. So I just wanted to point this out.

matthewavery:
LGTM

PR #5751, opened 2017-07-18
Allow UI scripts to take VC params via command arguments

### LINKED ISSUE #5749: Make UI scripts take params for VC IP and admin credentials
**User Statement:**
For the UI plugin install/uninstall/upgrade scripts, a suggestion came up during a Sprint 12 demo where we can provide the user with an option to provide VC IP and admin credentials in the form of command arguments.

**Details:**
Any details of what you want that might clarify for the developer how to approach the implementation.

**Acceptance Criteria:**
The user should be able to pass the aforementioned parameters via CLI arguments as another option, while keeping the interactive mode.

jooskim:
This PR allows for passing VC related parameters to install/uninstall/upgrade scripts via command arguments. Users can still use the interactive mode, but alternatively they can run commands as follows:

./install.(sh|bat) [-i vc_ip] [-u vc admin_username] [-p vc_admin_password]
./uninstall.(sh|bat) [-i vc_ip] [-u vc admin_username] [-p vc_admin_password]
./upgrade.(sh|bat) [-i vc_ip] [-u vc admin_username] [-p vc_admin_password]

Fixes #5749

### DIFF
--- a/ui/installer/VCSA/install.sh
+++ b/ui/installer/VCSA/install.sh
@@ -14,23 +14,55 @@
 # limitations under the License.
 #
 
-cleanup () {
-    unset VCENTER_ADMIN_USERNAME
-    unset VCENTER_ADMIN_PASSWORD
-}
+read_vc_information () {
+    while getopts ":fi:u:p:" o; do
+        case "${o}" in
+            i)
+                VCENTER_IP=$OPTARG
+                ;;
+            u)
+                VCENTER_ADMIN_USERNAME=$OPTARG
+                ;;
+            p)
+                VCENTER_ADMIN_PASSWORD=$OPTARG
+                ;;
+            f)
+                FORCE_INSTALL=1
+                ;;
+            *)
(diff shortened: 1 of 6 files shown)


jooskim:
The difference of install -f from upgrade is that it will install the plugins without checking the existing version. I use that for test cases, and thought some users might use it. Considering that the same could be done by the upgrade script though, it might be better to keep the force flag supported, but drop it from documentation or usage, to keep the experience consistent across the scripts.

jooskim:
@alextopuzov could you also review these for documentation? Maybe we can have a tip saying the user can provide VC information via command arguments alternatively.

andrewtchin:
Is there any reason to not always use force?

jooskim:
@emlin had a concern over defaulting to any sort of force mode when I had discussions with her while rewriting the scripts. Since it is possible that multiple admins can try installing the VIC UI plugins against a VC I think it makes sense that the default behavior of the installer script is to throw an error if the plugins exist.

andrewtchin:
OK that's cool

PR #5753, opened 2017-07-19
Concurrent copy

### LINKED ISSUE #5742: Docker cp fails on concurrent copy for offline container
When two cp requests to an offline container are targeted at the same volume or r/w layer, one request would successfully attached the vmdk, and return from portlayer. When the second request comes in, the vmdk is already attached and mounted to the VCH, but the current logic failover to online and assumes the owner is a container VM. This is a problem because if the VCH starts a guest tool connection to itself, ref counting for mount is not updated. The vmdk will be unmounted whenever the first request finishes, and can cause the second copy operation to fail in the middle.

Acceptance:
- [ ] many (>20) concurrent copyTo operations to the same offline container succeed
- [ ] detection for images/volumes/containers for the case where the online Owner identified _is_ the applianceVM and appropriate retry logic (unsure how to test this).

hickeng:
Updates to disk reference counting to handle concurrent attach case.
Currently exploits personality bug to get back end concurrency via: 
```
for i in {1..50};do docker cp bin/foo.txt offline:/; docker cp bin/bar offline:/;done
```

Does not yet handle accidental detection of the endpointVM as the online owner.

Towards: #5742

### DIFF
--- a/lib/portlayer/storage/vsphere/container.go
+++ b/lib/portlayer/storage/vsphere/container.go
@@ -236,7 +236,7 @@ func (c *ContainerStore) Export(op trace.Operation, id, ancestor string, spec *a
 	}
 
 	if ancestor == "" {
-		op.Infof("No ancester specified so following basic export path")
+		op.Infof("No ancestor specified so following basic export path")
 		return l.Export(op, spec, data)
 	}
 
@@ -249,11 +249,11 @@ func (c *ContainerStore) Export(op trace.Operation, id, ancestor string, spec *a
 		l.Close()
 		return nil, err
 	}
-	op.Debugf("Mapped ancester %s to %s", ancestor, img.String())
+	op.Debugf("Mapped ancestor %s to %s", ancestor, img.String())
 
 	r, err := c.newDataSource(op, img)
 	if err != nil {
-		op.Debugf("Unable to get datasource for ancester: %s", err)
+		op.Debugf("Unable to get datasource for ancestor: %s", err)
 
 		l.Close()
 		return nil, err
--- a/lib/portlayer/storage/vsphere/container.go
+++ b/lib/portlayer/storage/vsphere/container.go
@@ -68,6 +68,9 @@ func NewContainerStore(op trace.Operation, s *session.Session, imageResolver sto
 
 // URL converts the id of a resource to a URL
 func (c *ContainerStore) URL(op trace.Operation, id string) (*url.URL, error) {
+	// using diskfinder with a basic suffix match is an inefficient and potentially error prone way of doing this
+	// mapping, but until the container store has a structured means of knowing this information it's at least
+	// not going to be incorrect without an ID collision.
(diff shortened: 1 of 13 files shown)


lib/portlayer/storage/vsphere/export.go
```diff
@@ -130,13 +131,19 @@ func (i *ImageStore) Export(op trace.Operation, id, ancestor string, spec *archi
 	// this allows us to assume it's an image
 	r, err := i.NewDataSource(op, ancestor)
 	if err != nil {
+		op.Debugf("Unable to get datasource for ancester: %s", err)
```
caglar10ur: Based on the TV advertisement that I keep seeing, I'm sure http://www.ancestor.com/ will help you there :P

pkg/vsphere/disk/disk_manager.go
```diff
@@ -384,15 +413,14 @@ func (m *Manager) Detach(op trace.Operation, config *VirtualDiskConfig) error {
...
+
 	select {
 	case <-m.maxAttached:
 	default:
 	}
 
-	if err = d.setDetached(op, m.Disks); err != nil {
-		op.Errorf(err.Error())
-	}
-
```
caglar10ur: good catch.

matthewavery:
@hickeng there is now a feature branch for docker-cp for the final merges of funtionality. I will be sure to rebase above this to show that we are safe from this.

lib/portlayer/storage/vsphere/container.go
```diff
@@ -231,7 +236,7 @@ func (c *ContainerStore) Export(op trace.Operation, id, ancestor string, spec *a
 	}
 
 	if ancestor == "" {
-		op.Infof("No ancester specified so following basic export path")
+		op.Infof("No ancestorr specified so following basic export path")
```
jzt: out of the frying pan an into the fire :)

`s/ancestorr/ancestor`

lib/portlayer/storage/vsphere/export.go
```diff
@@ -156,7 +163,10 @@ func (i *ImageStore) Export(op trace.Operation, id, ancestor string, spec *archi
 		return nil, err
 	}
 
-	return tar, nil
+	return &storage.ProxyReadCloser{
```
jzt: Side question: are we still using the CleanupReader? If not, we could probably remove the file if it hasn't been already.

lib/portlayer/storage/vsphere/container.go
hickeng: just how did I manage that?

lib/portlayer/storage/vsphere/export.go
hickeng: It's gone. I'm not actually sure why CleanupReader wasn't getting typed as an io.Closer but the very slightly different `ProxyCloser` I tried worked, so I switched.

PR #5763, opened 2017-07-19
Make target for focused unit tests

hickeng:
Allows easy unit testing of all packages that differ from vmware/master.
This is specifically because I keep failing to run _all_ of the necessary
tests after making changes.


infra/scripts/focused-test.sh
```diff
@@ -0,0 +1,35 @@
...
+# limitations under the License.
+#
+# Runs go test on any packages that differ from vmware/master that have
+# Go test files.
+#
+# Usage: infra/scripts/focused-test.sh
+#
+
+echo "Finding modified packages with test files"
+PKG=$(git diff --stat vmware/master --name-only | xargs dirname | uniq | while read i; do
```
zjs: This assumes everyone is using the same remote name, `vmware`, for GitHub. That doesn't seem like a universally safe assumption. (I suspect some people stick with `origin`.)

Are there cases where diffing against `master` would incorrect?

infra/scripts/focused-test.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+# Copyright 2016 VMware, Inc. All Rights Reserved.
```
zjs: nit: it's now 2017

infra/scripts/focused-test.sh
hickeng: Should figure out a generic remote. There are definitely cases where diffing against master will be incorrect - if for example we're working with a feature branch.
I'll likely make the remote an environment variable with a default. I had considered just upstream branch, but that's not always set and not necessarily what I wanted for this.

I wanted this to test those changes that would be present in a PR vs `vmware/vic` which is where the tests tend to fail.

Makefile
```diff
@@ -257,6 +257,11 @@ install-govmomi:
 
 test: install-govmomi portlayerapi $(TEST_JOBS)
 
+
+focused-test:
+# test only those packages that have changes
+	infra/scripts/focused-test.sh $(REMOTE)
```
mhagen-vmware: Not seeing where $(REMOTE) is getting set prior to calling this... ?

infra/scripts/focused-test.sh
```diff
@@ -0,0 +1,43 @@
...
+#
+# Usage: infra/scripts/focused-test.sh
+#
+
+if [ -z "$1" ]; then
+    export REMOTE=$(git remote -v | grep "github.com/vmware/vic.git (fetch)" | awk '{print$1;exit}')/master
+    echo "Using ${REMOTE} as default remote"
+else
+    echo "Using ${REMOTE} as specified remote"
+    export REMOTE="$1"
```
zjs: The order of these lines seems backwards; we `echo` using `$REMOTE` before we've set it equal to `$1`.

Makefile
hickeng: It's not - if you call it in the following style it'll get passed through, otherwise it tries to determine a default in the script:
```
REMOTE=vmware/master make focused-test
```

hickeng:
Now I'd like to add the `make focused-check` for linting, et al as that's been another common failure in the `feature/docker-cp` branch.

Issue #4754, opened 2017-04-12
port layer: implement push image/layer to a repository


hmahmood:

hickeng:
portlayer is not involved in pushing images to repo. At most there is use of inspect and diff APIs

sflxn:
This is done and on the docker-push feature branch.  We are postponing docker push until VIC 1.3 and until we can validate the layer archives are correct.

PR #5728, opened 2017-07-16
Revert "Removing specially created clients for docker logs and stats [full ci]"

hmahmood:
Reverts vmware/vic#5708

PR #5774, opened 2017-07-20
Implement docker push image/manifest building and upload manager

### LINKED ISSUE #5312: Docker Persona:  Image construction within imagec

sflxn:
Added ability to build image/manifest and pull layer tars to support docker push.

- Implemented PushImage which mirrors PullImage.
- Refactored some imagec.go to make code easier to read/understand and reuse
- Added manifest building using Sophies's initial work as base.
- Added an upload manager based on Docker's upload manager.
- Added ability to pull layer diff tars from the new archive streamers.
- Added Cheng's initial backend work.
- Refactored the archive streamers out of container proxy in order for other
  code to get access to them.

Resolves #5312 and #4754

### DIFF
--- a/lib/apiservers/engine/backends/archive.go
+++ b/lib/apiservers/engine/backends/archive.go
@@ -16,19 +16,30 @@ package backends
 
 import (
 	"archive/tar"
-	"context"
+	"bytes"
+	"encoding/base64"
+	"encoding/gob"
+	"encoding/json"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"path"
 	"strings"
+	"time"
+
+	"golang.org/x/net/context"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/go-openapi/runtime"
+	rc "github.com/go-openapi/runtime/client"
+	httpclient "github.com/mreiferson/go-httpclient"
 	"github.com/tchap/go-patricia/patricia"
 
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	viccontainer "github.com/vmware/vic/lib/apiservers/engine/backends/container"
+	"github.com/vmware/vic/lib/apiservers/portlayer/client"
+	"github.com/vmware/vic/lib/apiservers/portlayer/client/storage"
 	vicarchive "github.com/vmware/vic/lib/archive"
 	"github.com/vmware/vic/lib/portlayer/constants"
 	"github.com/vmware/vic/pkg/trace"
@@ -36,6 +47,11 @@ import (
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/pkg/archive"
 )
+
+type VicArchiveProxy interface {
+	ArchiveExportReader(ctx context.Context, store, ancestorStore, deviceID, ancestor string, data bool, filterSpec vicarchive.FilterSpec) (io.ReadCloser, error)
+	ArchiveImportWriter(ctx context.Context, store, deviceID string, filterSpec vicarchive.FilterSpec) (io.WriteCloser, error)
+}
 
 // ContainerArchivePath creates an archive of the filesystem resource at the
 // specified path in the container identified by the given name. Returns a
@@ -68,7 +84,7 @@ func (c *Container) exportFromContainer(vc *viccontainer.VicContainer, path stri
 	mounts = append(mounts, types.MountPoint{Destination: "/"})
 	readerMap := NewArchiveStreamReaderMap(mounts, path)
 
-	readers, err := readerMap.ReadersForSourcePath(c.containerProxy, vc.ContainerID, path)
+	readers, err := readerMap.ReadersForSourcePath(archiveProxy, vc.ContainerID, path)
 	if err != nil {
 		log.Errorf("Errors getting readers for export: %s", err.Error())
 		return nil, err
@@ -140,7 +156,14 @@ func (c *Container) importToContainer(vc *viccontainer.VicContainer, target stri
 		}
 
 		// Lookup the writer for that mount prefix
-		writer, err := writerMap.WriterForAsset(c.containerProxy, vc.ContainerID, target, *header)
+		writer, err := writerMap.WriterForAsset(archiveProxy, vc.ContainerID, target, *header)
+		if err != nil {
+			return err
+		}
+
+		var buf bytes.Buffer
+		enc := gob.NewEncoder(&buf)
+		err = enc.Encode(header)
 		if err != nil {
 			return err
 		}
@@ -360,7 +383,7 @@ func (wm *ArchiveStreamWriterMap) FindArchiveWriter(containerDestPath, assetName
 //
 // As demonstrated above, the mount prefix and writer cannot be determined with just the
 // container destination path.  It must be combined with the actual asset's name.
-func (wm *ArchiveStreamWriterMap) WriterForAsset(proxy VicContainerProxy, cid, containerDestPath string, assetHeader tar.Header) (io.WriteCloser, error) {
+func (wm *ArchiveStreamWriterMap) WriterForAsset(proxy VicArchiveProxy, cid, containerDestPath string, assetHeader tar.Header) (io.WriteCloser, error) {
 	defer trace.End(trace.Begin(assetHeader.Name))
 
 	var err error
@@ -573,7 +596,7 @@ func (rm *ArchiveStreamReaderMap) buildFilterSpec(containerSourcePath string, no
 //
 //			containaerSroucePath -	/mnt/A
 // In the above, both readers are within the the container source path.
-func (rm *ArchiveStreamReaderMap) ReadersForSourcePath(proxy VicContainerProxy, cid, containerSourcePath string) ([]io.Reader, error) {
+func (rm *ArchiveStreamReaderMap) ReadersForSourcePath(proxy VicArchiveProxy, cid, containerSourcePath string) ([]io.Reader, error) {
 	defer trace.End(trace.Begin(containerSourcePath))
 
 	var streamReaders []io.Reader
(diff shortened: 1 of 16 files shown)


lib/apiservers/engine/backends/push.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/docker/docker/pkg/streamformatter"
+	"github.com/docker/docker/reference"
+)
+
+// PushImage initiates a push operation on the repository named localName.
+func (i *Image) PushImage(ctx context.Context, image, tag string, metaHeaders map[string][]string, authConfig *types.AuthConfig, outStream io.Writer) error {
+	// return fmt.Errorf("%s does not yet implement image.PushImage", ProductName())
+	defer trace.End(trace.Begin(fmt.Sprintf("%s:%s", image, tag)))
+
+	_, err := cache.ImageCache().Get(image)
```
chengwang86: `image` should be replaced by `imageID` or `reference`. This line would throw an imageNotFoundErr if the image is not tagged as `latest`. I will fix this bug in my PR.

lib/apiservers/engine/backends/push.go
```diff
@@ -0,0 +1,135 @@
...
+
+	ic := imagec.NewImageC(options, streamformatter.NewJSONStreamFormatter(), SimpleArchiveReader)
+	err = ic.PushImage()
+	if err != nil {
+		return err
+	}
+
+	//TODO:  Need repo name as second parameter.  Leave blank for now
+	actor := CreateImageEventActorWithAttributes(image, "", map[string]string{})
+	EventService().Log("pull", eventtypes.ImageEventType, actor)
```
chengwang86: `pull` --> `push`

lib/imagec/upload.go
```diff
@@ -0,0 +1,149 @@
...
+			case <-start:
+			default:
+				progress.Update(progressOutput, layerID, "Waiting")
+				<-start
+			}
+
+			// PutImageBlob will handle retries and backoff
+			err := PushImageBlob(u.Transfer.Context(), ic.Options, aStream, reader, progressOutput)
+			if err != nil {
+				// log.Errorf("Error pushing image blob for %s/%s: %s", ic.Image, layer.ID, err)
```
chengwang86: We need to either log this `err` or return it back to the daemon. Or we can do something similar to this https://github.com/moby/moby/blob/master/distribution/xfer/upload.go#L137

lib/apiservers/engine/backends/push.go
chengwang86: I think we can also remove this check since you also did it here https://github.com/vmware/vic/pull/5774/files#diff-981781345721c7111992d43025cc69d2R695

lib/apiservers/engine/backends/archive.go
```diff
@@ -642,3 +665,190 @@ func (rm *ArchiveStreamReaderMap) Close() {
...
+	r := rc.New(PortLayerServer(), "/", []string{"http"})
+	transport := &httpclient.Transport{
+		ConnectTimeout:        connectTimeout,
+		ResponseHeaderTimeout: responseHeaderTimeout,
+		RequestTimeout:        responseTimeout,
+	}
+
+	r.Transport = transport
+
+	plClient := client.New(r, nil)
```
hmahmood: This reintroduces the leak I just fixed with #5728 and #5730 . I think you should just use the client that the container proxy uses. You can have the container proxy object implement the ArchiveProxy interface.

lib/imagec/imagec.go
```diff
@@ -569,16 +682,325 @@ func (ic *ImageC) PullImage() error {
...
+
+		log.Infof("Image data for layer %s = %#v", layerID, *layer.Image)
+
+		// Check for scratch ID
+		if layer.Image.Parent == storage.Scratch.ID {
+			break
+		}
+
+		// set the layer to the parent layer
+		layerID = layer.Image.Parent
```
sgairo: Does layer.Image.Parent result in the most immediate parent or the base parent?

lib/apiservers/engine/backends/archive.go
```diff
@@ -37,6 +48,11 @@ import (
 	"github.com/docker/docker/pkg/archive"
 )
 
+type VicArchiveProxy interface {
+	ArchiveExportReader(ctx context.Context, store, ancestorStore, deviceID, ancestor string, data bool, filterSpec vicarchive.FilterSpec) (io.ReadCloser, error)
```
emlin: might rename to ExportReader only? feels a bit duplicate with the interface name

lib/imagec/upload.go
```diff
@@ -0,0 +1,149 @@
...
+
+			select {
+			case <-start:
+			default:
+				progress.Update(progressOutput, layerID, "Waiting")
+				<-start
+			}
+
+			// PutImageBlob will handle retries and backoff
+			err := PushImageBlob(u.Transfer.Context(), ic.Options, aStream, reader, progressOutput)
```
sgairo: Is this PutImageBlob or PushImageBlob?

lib/apiservers/engine/backends/archive.go
```diff
@@ -642,3 +665,190 @@ func (rm *ArchiveStreamReaderMap) Close() {
...
+		return nil, fmt.Errorf("ArchiveExportReader called with either empty store or deviceID")
+	}
+
+	plClient, transport := createGzipTarClient(attachConnectTimeout, 0, attachAttemptTimeout)
+
+	var err error
+
+	pipeReader, pipeWriter := io.Pipe()
+
+	// done := make(chan struct{})
```
emlin: might remove this unused code?

lib/apiservers/engine/backends/archive.go
```diff
@@ -642,3 +665,190 @@ func (rm *ArchiveStreamReaderMap) Close() {
...
+
+	if store == "" || deviceID == "" {
+		return nil, fmt.Errorf("ArchiveImportWriter called with either empty store or deviceID")
+	}
+
+	var err error
+
+	pipeReader, pipeWriter := io.Pipe()
+
+	// done := make(chan struct{})
```
emlin: ditto

lib/apiservers/engine/backends/push.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/docker/distribution/digest"
+	"github.com/docker/docker/api/types"
+	eventtypes "github.com/docker/docker/api/types/events"
+	"github.com/docker/docker/pkg/streamformatter"
+	"github.com/docker/docker/reference"
+)
+
+// PushImage initiates a push operation on the repository named localName.
+func (i *Image) PushImage(ctx context.Context, image, tag string, metaHeaders map[string][]string, authConfig *types.AuthConfig, outStream io.Writer) error {
+	// return fmt.Errorf("%s does not yet implement image.PushImage", ProductName())
```
emlin: remove comment

lib/apiservers/engine/backends/push.go
```diff
@@ -0,0 +1,135 @@
...
+	defer trace.End(trace.Begin(fmt.Sprintf("%s:%s", image, tag)))
+
+	_, err := cache.ImageCache().Get(image)
+	if err != nil {
+		return ImageNotFoundError(image, tag)
+	}
+
+	log.Debugf("PushImage: image = %s, tag = %s, metaheaders = %+v\n, authConfig = %+v\n", image, tag, metaHeaders, authConfig)
```
emlin: might merge this message with trace log? as Loc mentioned remove the above check

lib/apiservers/engine/backends/push.go
```diff
@@ -0,0 +1,135 @@
...
+	if portLayerServer != "" {
+		options.Host = portLayerServer
+	}
+
+	log.Infof("PushImage: reference: %s, %s, portlayer: %#v",
+		options.Reference,
+		options.Host,
+		portLayerServer)
+
+	log.Infof("-------------The imagecOption:-----------\n %+v", options)
```
emlin: are they all message in Info level?

lib/imagec/docker.go
```diff
@@ -296,6 +297,8 @@ func FetchImageBlob(ctx context.Context, options Options, image *ImageWithMeta,
 
 	progress.Update(progressOutput, image.String(), "Download complete")
 
+	log.Infof("Layer %s blobsum, diff id = (%s, %s)", id, layer, diffID)
```
emlin: debug message?

lib/imagec/imagec.go
```diff
@@ -569,16 +682,325 @@ func (ic *ImageC) PullImage() error {
...
 
+	pusher.PushManifest.Layers = layers
+
+	log.Infof("Final manifest = %#v", pusher.PushManifest)
+
 	return nil
 }
+
+func (ic *ImageC) pushManifest(ctx context.Context) error {
+	if ic.ImageManifestSchema2 != nil {
```
chengwang86: This might be a problem. Where is this `ImageManifestSchema2` assigned? In my local testing, it is always nil. Since we only construct image manifest following schema2, shall we remove this check?

lib/imagec/imagec.go
```diff
@@ -123,12 +135,35 @@ type ImageWithMeta struct {
...
+	layerFile     *os.File
+	size          int64
+	digest        string
+	layerID       string
+	parentLayerID string
+}
+
+// Pusher contains all "prepared" data needed to push an image
+type Pusher struct {
+	// pushWg          sync.WaitGroup
```
jzt: Remove

lib/imagec/imagec.go
```diff
@@ -463,11 +499,79 @@ func (ic *ImageC) CreateImageConfig(images []*ImageWithMeta) (metadata.ImageConf
...
+	return nil
+}
 
-	// ctx
+// PushImage pushes an image to a registry
+func (ic *ImageC) PushImage() error {
 	ctx, cancel := context.WithTimeout(ctx, ic.Options.Timeout)
 	defer cancel()
 
+	// Authenticate, get URL, get token
```
jzt: Remove commented code

lib/imagec/imagec.go
```diff
@@ -463,11 +499,79 @@ func (ic *ImageC) CreateImageConfig(images []*ImageWithMeta) (metadata.ImageConf
...
+	if err := lum.UploadLayers(ctx, ic); err != nil {
+		return err
+	}
+
+	if err := ic.FinalizeManifest(); err != nil {
+		return err
+	}
+
+	// Push up the image manifest
+	// if err := ic.pushManifest(ctx); err != nil {
```
jzt: Remove

lib/imagec/imagec.go
```diff
@@ -569,16 +682,325 @@ func (ic *ImageC) PullImage() error {
...
+		return fmt.Errorf("Cannot push image manifest schema 1")
+	} else {
+		return fmt.Errorf("attempt to push manifest when non exist")
+	}
+
+	return nil
+}
+
+// dockerImageFromVicImage takes a slice of VIC image with meta and returns a docker.Image,
+// the calculated digest of that struct, the size of
```
jzt: This comment looks to be incomplete

lib/imagec/imagec.go
```diff
@@ -569,16 +682,325 @@ func (ic *ImageC) PullImage() error {
...
+		return nil, "", 0, fmt.Errorf("Failed to marshall image metadata: %s", err)
+	}
+
+	digest := fmt.Sprintf("%x", sha256.Sum256(imageConfigBytes))
+	configSize := int64(len(imageConfigBytes))
+
+	return result, digest, configSize, nil
+}
+
+// GetReaderForLayer returns a io.ReadCloser for the archive stream connection to the
```
jzt: Remove commented code

lib/imagec/imagec_test.go
```diff
@@ -600,3 +603,143 @@ func TestFetchScenarios(t *testing.T) {
...
+func TestPrepareManifestAndLayers(t *testing.T) {
+	type PushMock struct {
+		image string
+	}
+	// schema1Manifests := map[string]schema1.Manifest{
+	// 	"ubuntu": {
+	// 		Name: "library/busybox",
+	// 		Tag:  "latest",
+	// 		// Digest: "sha256:74155e76e3396bc23d5b8cedf606bc04f7f79660ee5bce14014b84651fc4f672",
+	// 		FSLayers: []schema1.FSLayer{
```
jzt: Commented code

lib/apiservers/engine/backends/archive.go
sflxn: I put this in to get it to build.  Since the client is only accessible to the container backend.  Let me rework this and figure out how to make that client available across the backends.

lib/imagec/imagec.go
sflxn: Immediate parent, I believe.  From my testing, it appears to be correct.

lib/apiservers/engine/backends/push.go
sflxn: No, we need the check here.  It's our standard practice.  Also, even though imagec is part of the persona, it's a separate library.  We should not NOT do a check because we know a library code will  also do the check.  The responsibility of this check should be in the backend.

lib/imagec/docker.go
sflxn: yes.  will remove once everything works.

lib/imagec/imagec.go
```diff
@@ -569,16 +671,259 @@ func (ic *ImageC) PullImage() error {
...
+	if err := PutImageManifest(ctx, ic.Pusher, ic.Options, ic.progressOutput); err != nil {
 		return err
 	}
 
 	return nil
 }
+
+// dockerImageFromVicImage takes a slice of VIC image with meta and returns a docker.Image,
+// the calculated digest of that struct, the size of the config
+func (ic *ImageC) dockerImageFromVicImage(images []*ImageWithMeta) (*docker.Image, string, int64, error) {
```
emlin: This method looks similar to imagec CreateImageConfig, is there any difference with that?

sflxn: i will refactor before we merge to master

lib/imagec/upload.go
```diff
@@ -0,0 +1,149 @@
...
+			case <-u.Transfer.Done():
+				reader.Close()
+			}
+		}()
+
+		go func() {
+			log.Info("About to upload...")
+			defer func() {
+				close(progressChan)
+				aStream.Close()
```
emlin: aStream.Close will close reader, isn't this duplicate with line 119?
the reader is already set into aStream.layerfile, why do you still need a separate returned reader, and close those two separately.

lib/imagec/upload.go
```diff
@@ -0,0 +1,149 @@
...
+
+			select {
+			case <-start:
+			default:
+				progress.Update(progressOutput, layerID, "Waiting")
+				<-start
+			}
+
+			// PushImageBlob will handle retries and backoff
+			err := PushImageBlob(u.Transfer.Context(), ic.Options, aStream, reader, progressOutput)
```
emlin: Where is the PushImageBlob?

sgairo: I think its called PutImageBlob now, but im not sure? I asked about it in an earlier comment.

chengwang86: @emlin @sgairo https://github.com/vmware/vic/blob/feature/docker-push/lib/imagec/docker.go#L478

PR #5780, opened 2017-07-20
VMDK owner and status checking

### LINKED ISSUE #5742: Docker cp fails on concurrent copy for offline container

hickeng:
If online owner of VMDK is the appliance it will try the offline path again, once.
Exposes the vmdk locked check in the `pkg/vsphere/disk` package.

Both of these are untested as yet, but wanted them available early for use.
Testing of both will be easiest when online cp is available.

Fixes #5742

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ endif
 # Caches dependencies to speed repeated calls
 define godeps
 	$(call assert,$(call gmsl_compatible,1 1 7), Wrong GMSL version) \
-	$(if $(filter-out clean distclean mrrobot mark sincemark .DEFAULT,$(MAKECMDGOALS)), \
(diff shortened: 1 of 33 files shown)


hickeng:
This PR has got confused - closing and reopening a new one.

PR #5781, opened 2017-07-20
fix IsLockedError check

### LINKED ISSUE #5713: Add proper check for locked VMDK in NewDataSource/Sink paths
Applies to #5700 

Currently the check for whether the VMDK is locked simply returns `true` and fails over to the online path irrespective of actual failure.
Even if we want to keep the assured failover, we should correctly check.

The error we're looking for appears in reason chain:
```
types.LocalizableMessage{DynamicData:types.DynamicData{}, Key:"msg.fileio.lock", Arg:[]types.KeyAnyValue(nil), Message:"Failed to lock the file"}
```

Implementation is in `vmdk.go:IsLockedError`

jialin-li:
fixes #5713

### DIFF
--- a/pkg/vsphere/disk/vmdk.go
+++ b/pkg/vsphere/disk/vmdk.go
@@ -68,7 +68,7 @@ func LockedVMDKFilter(vm *mo.VirtualMachine) bool {
 func IsLockedError(err error) bool {
 	disks := LockedDisks(err)
 	//if device is locked, disks will not be nil
-	return disks != nil
+	return len(disks) > 0
 }
 
 // LockedDisks returns locked devices path in the error if it's device lock error
--- a/pkg/vsphere/disk/vmdk.go
+++ b/pkg/vsphere/disk/vmdk.go
@@ -79,14 +79,13 @@ func IsLockedError(op trace.Operation, err error) bool {
 			message := &fault.FaultMessage[i]
 			switch message.Key {
 			case DiskBackendKey:
-				op.Debugf("diskbackend is true")
 				diskBackend = true
 			case LockedFileKey:
-				op.Debugf("lockedfilekey is true")
 				fileLock = true
 			}
 
 			if diskBackend && fileLock {
+				op.Debugf("Error %s is a disk locked error", err.Error())
 				return true
 			}
 		}
@@ -94,6 +93,7 @@ func IsLockedError(op trace.Operation, err error) bool {
 		op.Debugf("IsLockedError: this is not a task error")
 	}
 
+	op.Debugf("Error %s is a NOT disk locked error", err.Error())
 	return false
 }
 
--- a/pkg/vsphere/disk/vmdk.go
+++ b/pkg/vsphere/disk/vmdk.go
@@ -65,36 +65,10 @@ func LockedVMDKFilter(vm *mo.VirtualMachine) bool {
 	return vm.Runtime.PowerState == types.VirtualMachinePowerStatePoweredOn
 }
 
-func IsLockedError(op trace.Operation, err error) bool {
-	switch err := err.(type) {
-	case task.Error:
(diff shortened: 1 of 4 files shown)


pkg/vsphere/disk/vmdk.go
```diff
@@ -58,7 +64,34 @@ func LockedVMDKFilter(vm *mo.VirtualMachine) bool {
...
+		// indicates it's specific to a disk
+		diskBackend := false
+		// indicates it's specific to locking
+		fileLock := false
+
+		for i := range fault.FaultMessage {
+			message := &fault.FaultMessage[i]
+			switch message.Key {
+			case DiskBackendKey:
+				op.Debugf("diskbackend is true")
```
emlin: might log out the value of this key, so that we get to know which disk is locked

pkg/vsphere/disk/vmdk.go
```diff
@@ -58,7 +64,34 @@ func LockedVMDKFilter(vm *mo.VirtualMachine) bool {
 	return vm.Runtime.PowerState == types.VirtualMachinePowerStatePoweredOn
 }
 
-func IsLockedError(err error) bool {
-	// TODO: add check here for actual error
-	return true
+func IsLockedError(op trace.Operation, err error) bool {
```
emlin: some client would want to know which disk is locked, and this method already go through all the messages, it would be nice to return the locked disk name as well.

emlin:
please add unit test as well

pkg/vsphere/disk/vmdk.go
emlin: The in use device might be multiple, so returned devices should be []string type.

hickeng: In the import/export specific case we don't care at all as we _know_ it's a single resource we're trying to add.
From a general use perspective, identifying which disks are locked is a definite improvement.

To keep this particular call simple I'd suggest adding a second call (doesn't have to be part of this PR, but likely needed by @emlin):
`LockedDisks(trace.Operation, error) ([]string)`, or
`LockedDisks(trace.Operation, error) ([]object.DatastorePath)`

pkg/vsphere/disk/vmdk.go
hickeng: If we're logging the original error then we don't need these debug messages - I'm wishing again we had more debug levels as this is fine grained trace.

Can you consolidate the debugging messages so we get a single log on exit that tells us the details.

pkg/vsphere/disk/vmdk.go
emlin: This is another option, for the logic in these two methods will be similar, and not that easy to consolidate.
But I'm Okay with it, as it will release the dependency of my PR and this one.

emlin: okay, anyway, didin't figure out many difference between these two methods. But the method LockedDisks will do more than isLockedError, so we'll have isLockedError call LockedDisks, just do not return the locked disks.

So we'll have two methods for different users with different interest, but the logic will be in LockedDisks.

jialin-li:
@hickeng I have tested this function with copy from a running container and in use volumes, it properly checks the error.

pkg/vsphere/disk/vmdk.go
```diff
@@ -66,8 +66,9 @@ func LockedVMDKFilter(vm *mo.VirtualMachine) bool {
 }
 
 func IsLockedError(err error) bool {
-	// TODO: add check here for actual error
-	return true
+	disks := LockedDisks(err)
+	//if device is locked, disks will not be nil
+	return disks != nil
```
hickeng: this is correct, but a more implementation tolerant check is below - this will handle both `nil` and the case of an empty array:
```go
len(disks) > 0
```

PR #5783, opened 2017-07-21
Inspect - Restart VCH - Inspect test case complete for standard and NFS volumes.

### LINKED ISSUE #5758: Volume Test Case for Inspect-Restart VCH-Inspect
As an engineer, I want to make sure that re-starting a VCH will produce the same inspect information on a volume (and NFS volume as well).

**Acceptance Criteria:**
Add test case for standard and NFS volumes where we compare the output from a volume using inspect, restart the VCH, and run inspect again.
Info should be the same.

rogeliosanchez:
Fixes #5758


matthewavery:
I think this is good, we will also want to do something like 

```
docker inspect cid > out1
reboot VCH
docker inspect cid > out2
diff out1 out2
make sure the diff is acceptable(ideally there should be no difference)
```

rogeliosanchez:
Closing since there is a better place for these tests to be implemented.  Also, adding the test to more areas.

Issue #5157, opened 2017-05-18
Concurrent deletion of about 70+ containers failed(from Management UI)


lgayatri:
Concurrent deletion of about 70+ containers failed with below error:

request failed, no additional details provided; Reason: {"message":"request failed, no additional details provided","stackTrace":["com.vmware.xenon.common.Operation.fail(Operation.java:1227)","com.vmware.xenon.common.http.netty.NettyHttpServiceClient.failExpiredRequests(NettyHttpServiceClient.java:834)","com.vmware.xenon.common.http.netty.NettyHttpServiceClient.handleMaintenance(NettyHttpServiceClient.java:781)","com.vmware.admiral.adapter.docker.service.RemoteApiDockerAdapterCommandExecutorImpl.handleMaintenance(RemoteApiDockerAdapterCommandExecutorImpl.java:515)","com.vmware.admiral.adapter.docker.service.AbstractDockerAdapterService.handleMaintenance(AbstractDockerAdapterService.java:102)","com.vmware.xenon.common.ServiceMaintenanceTracker.lambda$performServiceMaintenance$1(ServiceMaintenanceTracker.java:171)","com.vmware.xenon.common.ServiceHost.executeRunnableSafe(ServiceHost.java:5019)","com.vmware.xenon.common.ServiceHost.lambda$schedule$55(ServiceHost.java:5013)","java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)","java.util.concurrent.FutureTask.run(FutureTask.java:266)","java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)","java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)","java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)","java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)","java.lang.Thread.run(Thread.java:745)"],"statusCode":408,"documentKind":"com:vmware:xenon:common:ServiceErrorResponse","errorCode":0}

Attached log bundle
[vm-2017-05-18.6905.tar.gz](https://github.com/vmware/vic/files/1010804/vm-2017-05-18.6905.tar.gz)

mdubya66:
Should this be opened in admiral?

jzt:
[...]

Issue #1723, opened 2016-07-28
Creating 75 containers fails due to connection reset


hmahmood:
This is from Erin on the performance team. One of the tests she has it to start 75 containers simultaneously, and she is reporting over 50 of the starts failing. Attached are the logs from the tests. It seems like we are dropping the sdk connection at some point, and subsequent starts fail.

[esxtop.csv.gz](https://github.com/vmware/vic/files/388881/esxtop.csv.gz)
[results.txt](https://github.com/vmware/vic/files/388879/results.txt)
[vic_esx_docker-personality.txt](https://github.com/vmware/vic/files/388880/vic_esx_docker-personality.txt)
[vic_esx_install.txt](https://github.com/vmware/vic/files/388878/vic_esx_install.txt)
[vic_esx_logs.tar.gz](https://github.com/vmware/vic/files/388882/vic_esx_logs.tar.gz)
[vic_esx_run.txt](https://github.com/vmware/vic/files/388883/vic_esx_run.txt)

corrieb:
I feel like we should learn to walk before we can run here. How do we do with 5 or 10 or 20? Our existing 30 second timeout is going to probably be insufficient at a far smaller number than 75

carlsone:
The same test consistently succeeds with 10 parallel containers.

corrieb:
I feel we should focus on #1724 first. Solving that - or at least improving it - would help significantly here. As a second order concern, we should adapt vcsim so that we can simulate creating containers very fast. This will help us figure out where the bottlenecks are in our control plane. Created #1731

mdubya66:
Moving to Icebox per 9/20 Triage

gigawhitlocks:
Volunteered for this but was given something higher priority before I could get started; sorry for the assignee / pipeline spam above.

karthik-narayan:
@mdubya66 Can we bump the priority on this one for the next sprint? Both the performance and scale teams will be working close to these limits and will be blocked if we are not able to fix this.

mdubya66:
I think we should consider it for inclusion into 1.2 but for 1.1 this would be too risky for the final sprint, IMO. @karthik-narayan. Setting to medium

karthik-narayan:
@mdubya66 Apologies. I meant to say for 1.2, not 1.1.

anishmehta1:
[...]

Issue #5647, opened 2017-07-06
hostd error on container that we created: Invalid transition requested (VM_STATE_RECONFIGURING -> VM_STATE_RECONFIGURING)


mhagen-vmware:
Generated from:
docker %{VCH-PARAMS} create ${busybox} /bin/top

It actually causes an exception that the server ends up catching from the looks of it, but I don't think we should be causing exceptions in the server in our standard day to day ops.
```
2017-07-06T13:02:58.250Z info hostd[33240B70] [Originator@6876 sub=Vimsvc.TaskManager opID=c3cfeabc user=droneci] Task Created : haTask-50-vim.VirtualMachine.reconfigure-26096
2017-07-06T13:02:58.252Z error hostd[335C2B70] [Originator@6876 sub=Vmsvc.vm:/vmfs/volumes/58af7eb5-d9fc5174-6127-0cc47a9be492/bbf8e32d8ac3b32b972757053620b355d7215e30cad124fbf4294a68a1c10576/bbf8e32d8ac3b32b972757053620b355d7215e30cad124fbf4294a68a1c10576.vmx opID=c3cfeabc user=droneci] Invalid transition requested (VM_STATE_RECONFIGURING -> VM_STATE_RECONFIGURING): Tasks running
2017-07-06T13:02:58.252Z warning hostd[335C2B70] [Originator@6876 sub=Vmsvc.vm:/vmfs/volumes/58af7eb5-d9fc5174-6127-0cc47a9be492/bbf8e32d8ac3b32b972757053620b355d7215e30cad124fbf4294a68a1c10576/bbf8e32d8ac3b32b972757053620b355d7215e30cad124fbf4294a68a1c10576.vmx opID=c3cfeabc user=droneci] Failed to start Reconfiguring: vim.fault.TaskInProgress
2017-07-06T13:02:58.252Z info hostd[335C2B70] [Originator@6876 sub=Default opID=c3cfeabc user=droneci] AdapterServer caught exception: vim.fault.TaskInProgress
2017-07-06T13:02:58.252Z info hostd[335C2B70] [Originator@6876 sub=Vimsvc.TaskManager opID=c3cfeabc user=droneci] Task Completed : haTask-50-vim.VirtualMachine.reconfigure-26096 Status error
```

mhagen-vmware:
[vmsupport-bundle.zip](https://github.com/vmware/vic/files/1128364/vmsupport-bundle.zip)

corrieb:
[...]

Issue #5621, opened 2017-06-29
vSphere SDK API Endpoint becomes unresponsive - govc: Post https://builds-14.vmware.run/sdk: net/http: TLS handshake timeout


mhagen-vmware:
Seeing this error:
```
09:33:31.081	INFO	Running command 'curl -sk https://192.168.31.114:2378/container-logs.tar.gz -b /tmp/cookies-VCH-12000-3110 | (cd /tmp; tar xvzf - inspiring_mccarthy-bb906f539ebd/tether.debug inspiring_mccarthy-bb906f539ebd/vmware.log) 2>&1'.	
09:33:31.116	INFO	${rc} = 2	
09:33:31.116	INFO	${output} = 
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

And in vic-admin.log:
```
Jun 29 2017 14:33:38.004Z INFO  User with a valid auth cookie at 192.168.31.248 is authenticated.
Jun 29 2017 14:33:38.007Z DEBUG [BEGIN] [main.(*server).bundleContainerLogs:451]
Jun 29 2017 14:33:38.007Z ERROR Failed to acquire logrotate lock: open /var/run/lock/logrotate_run.lock: permission denied
Jun 29 2017 14:33:38.007Z DEBUG [BEGIN] [main.configureReaders:135]
```

mhagen-vmware:
[Test-Cases.Group9-VIC-Admin.9-01-VICAdmin-ShowHTML-VCH-12000-3110-container-logs.zip](https://github.com/vmware/vic/files/1112565/Test-Cases.Group9-VIC-Admin.9-01-VICAdmin-ShowHTML-VCH-12000-3110-container-logs.zip)

sgairo:
`Jun 29 2017 14:33:23.123Z ERROR Error creating new govmomi client without credentials but with auth cookie: Post https://192.168.31.28/sdk: net/http: TLS handshake timeout
Jun 29 2017 14:33:23.123Z WARN  192.168.31.248:55808 failed to authenticate 
`

mhagen-vmware:
Yep, I think my original description is actually a red herring as I see that on successful runs as well. It looks like perhaps the ESX server just stop responding.  I think we need to add a retry around this call. I have added one in my test now, but we should not bail immediately when the ESX is unresponsive (which can happen easily)

mhagen-vmware:
From @sgairo :
[...]

PR #5784, opened 2017-07-21
[full ci] Limit the number of soap requests in flight

caglar10ur:
Otherwise ESXi starts to kick us out and most of the portlayer code
cannot handle the error and fails to reconnect as of this moment.

This also protects ESXi from us.

Problem found while testing with following;

```bash
#!/bin/bash
COUNT=50

docker pull busybox

for i in $(seq $COUNT);
do
    docker create -t --name $i busybox /bin/sleep $i > /dev/null &
done

read -p "Press any key to continue"

for i in $(seq $COUNT);
do
    docker rm $i > /dev/null &
done
```

Helps #1723
Helps #5157

### DIFF
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -195,23 +195,24 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
 	_, err = vm.WaitForResult(d.ctx, func(ctx context.Context) (tasks.Task, error) {
 		return vm.DeleteExceptDisks(ctx)
 	})
+	// We are getting ConcurrentAccess errors from DeleteExceptDisks - even though we don't set ChangeVersion in that path
+	// We are ignoring the error because in reality the operation finishes successfully.
+	ignore := false
 	if err != nil {
-		f, ok := err.(types.HasFault)
-		if !ok {
-			return err
-		}
-		switch f.Fault().(type) {
-		case *types.InvalidState:
-			log.Warnf("Container VM is in invalid state, unregistering")
-			if err := vm.Unregister(d.ctx); err != nil {
-				log.Errorf("Error while attempting to unregister container VM: %s", err)
-				return err
+		if f, ok := err.(types.HasFault); ok {
+			switch f.Fault().(type) {
+			case *types.ConcurrentAccess:
+				log.Warnf("DeleteExceptDisks failed with ConcurrentAccess error. Ignoring it.")
+				ignore = true
 			}
-		case *types.ConcurrentAccess:
-			log.Warnf("DeleteExceptDisks failed with ConcurrentAccess error. Ignoring it.")
-		default:
-			log.Debugf("Fault while attempting to destroy vm: %#v", f.Fault())
-			return err
+		}
+		if !ignore {
+			err = errors.Errorf("Failed to destroy VM %q: %s", vm.Reference(), err)
+			err2 := vm.Unregister(d.ctx)
+			if err2 != nil {
+				return errors.Errorf("%s then failed to unregister VM: %s", err, err2)
+			}
+			log.Infof("Unregistered VM to cleanup after failed destroy: %q", vm.Reference())
 		}
 	}
 
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -191,17 +191,30 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
 		}
 	}
 
+	//removes the vm from vsphere, but detaches the disks first
 	_, err = vm.WaitForResult(d.ctx, func(ctx context.Context) (tasks.Task, error) {
 		return vm.DeleteExceptDisks(ctx)
 	})
 	if err != nil {
-		err = errors.Errorf("Failed to destroy VM %q: %s", vm.Reference(), err)
-		err2 := vm.Unregister(d.ctx)
-		if err2 != nil {
-			return errors.Errorf("%s then failed to unregister VM: %s", err, err2)
+		f, ok := err.(types.HasFault)
+		if !ok {
+			return err
 		}
-		log.Infof("Unregistered VM to cleanup after failed destroy: %q", vm.Reference())
+		switch f.Fault().(type) {
+		case *types.InvalidState:
+			log.Warnf("Container VM is in invalid state, unregistering")
+			if err := vm.Unregister(d.ctx); err != nil {
+				log.Errorf("Error while attempting to unregister container VM: %s", err)
+				return err
+			}
+		case *types.ConcurrentAccess:
+			log.Warnf("DeleteExceptDisks failed with ConcurrentAccess error. Ignoring it.")
+		default:
+			log.Debugf("Fault while attempting to destroy vm: %#v", f.Fault())
+			return err
+		}
 	}
+
 	if _, err = d.deleteDatastoreFiles(d.session.Datastore, folder, true); err != nil {
 		log.Warnf("Failed to remove datastore files for VM path %q: %s", folder, err)
 	}
--- a/lib/portlayer/exec/container.go
+++ b/lib/portlayer/exec/container.go
@@ -596,6 +596,8 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
 				log.Errorf("Error while attempting to unregister container VM: %s", err)
 				return err
 			}
+		case *types.ConcurrentAccess:
+			log.Warnf("DeleteExceptDisks failed with ConcurrentAccess error. Ignoring it.")
 		default:
 			log.Debugf("Fault while attempting to destroy vm: %#v", f.Fault())
 			c.updateState(existingState)
--- a/lib/portlayer/exec/container.go
+++ b/lib/portlayer/exec/container.go
@@ -597,6 +597,8 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
 				return err
 			}
 		case *types.ConcurrentAccess:
+			// We are getting ConcurrentAccess errors from DeleteExceptDisks - even though we don't set ChangeVersion in that path
+			// We are ignoring the error because in reality the operation finishes successfully.
 			log.Warnf("DeleteExceptDisks failed with ConcurrentAccess error. Ignoring it.")
 		default:
 			log.Debugf("Fault while attempting to destroy vm: %#v", f.Fault())
--- a/pkg/vsphere/session/session.go
+++ b/pkg/vsphere/session/session.go
@@ -26,7 +26,10 @@ package session
 
 import (
 	"fmt"
+	"net/http"
 	"net/url"
+	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -45,6 +48,10 @@ import (
 	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/vsphere/extraconfig"
+)
+
+const (
+	defaultMaxInFlight = 16
 )
 
 // Config contains the configuration used to create a Session.
@@ -84,6 +91,32 @@ type Session struct {
 	VMFolder *object.Folder
 
 	Finder *find.Finder
+}
+
+// RoundTripFunc alias
+type RoundTripFunc func(*http.Request) (*http.Response, error)
+
+// RoundTrip method
+func (rt RoundTripFunc) RoundTrip(r *http.Request) (*http.Response, error) {
+	return rt(r)
+}
+
+// LimitConcurrency limits how many requests can be processed at once
+func LimitConcurrency(rt http.RoundTripper, limit int) http.RoundTripper {
+	limiter := make(chan struct{}, limit)
+
+	return RoundTripFunc(func(r *http.Request) (*http.Response, error) {
+		// reserve a slot
+		limiter <- struct{}{}
+
+		// free the slot
+		defer func() {
+			<-limiter
+		}()
+
+		// use the given round tripper
+		return rt.RoundTrip(r)
+	})
 }
 
 // NewSession creates a new Session struct. If config is nil,
@@ -180,8 +213,19 @@ func (s *Session) Connect(ctx context.Context) (*Session, error) {
 
 	soapClient.SetThumbprint(soapURL.Host, s.Thumbprint)
 
-	// TODO: option to set http.Client.Transport.TLSClientConfig.RootCAs
+	maxInFlight := defaultMaxInFlight
+	if e := os.Getenv("VIC_MAX_IN_FLIGHT"); e != "" {
+		if i, err := strconv.Atoi(e); err == nil {
+			maxInFlight = i
+		}
+	}
+	// Limit the concurrenty of SOAP requests
+	if t, ok := soapClient.Transport.(*http.Transport); ok {
+		t.MaxIdleConnsPerHost = maxInFlight
+	}
+	soapClient.Transport = LimitConcurrency(soapClient.Transport, maxInFlight)
 
+	// TODO: option to set http.Client.Transport.TLSClientConfig.RootCAs
 	vimClient, err := vim25.NewClient(ctx, soapClient)
 	if err != nil {
 		return nil, SoapClientError{
--- a/pkg/vsphere/tasks/waiter.go
+++ b/pkg/vsphere/tasks/waiter.go
@@ -121,7 +121,6 @@ func isRetryError(op trace.Operation, err error) bool {
 		default:
 			logSoapFault(op, f)
 			return false
-
 		}
 	}
 
@@ -154,20 +153,18 @@ func isRetryError(op trace.Operation, err error) bool {
 			return true
 		default:
 			logFault(op, err.Fault())
+			return false
 		}
 	default:
-		if f, ok := err.(types.HasFault); ok {
-			logFault(op, f.Fault())
-		} else {
-			// retry the temporary errors
-			t, ok := err.(temporary)
-			if ok && t.Temporary() {
-				return true
-			}
-			logError(op, err)
+		// retry the temporary errors
+		t, ok := err.(temporary)
+		if ok && t.Temporary() {
+			logExpectedError(op, err)
+			return true
 		}
+		logError(op, err)
+		return false
 	}
-	return false
 }
 
 // Helper Functions
@@ -185,4 +182,8 @@ func logError(op trace.Operation, err error) {
 
 func logExpectedFault(op trace.Operation, kind string, fault interface{}) {
 	op.Debugf("task retry on expected %s fault: %#v", kind, fault)
+}
+
+func logExpectedError(op trace.Operation, err error) {
+	op.Debugf("task retry on expected error %s", err)
 }
--- a/pkg/vsphere/tasks/waiter.go
+++ b/pkg/vsphere/tasks/waiter.go
@@ -61,7 +61,6 @@ func Wait(ctx context.Context, f func(context.Context) (Task, error)) error {
 //    })
 func WaitForResult(ctx context.Context, f func(context.Context) (Task, error)) (*types.TaskInfo, error) {
 	var err error
-	var info *types.TaskInfo
 	var backoffFactor int64 = 1
 
 	op, err := trace.FromContext(ctx)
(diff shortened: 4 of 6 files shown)


pkg/vsphere/tasks/waiter.go
```diff
@@ -164,11 +201,11 @@ func logFault(op trace.Operation, fault types.BaseMethodFault) {
 }
 
 func logSoapFault(op trace.Operation, fault types.AnyType) {
-	op.Debugf("unexpected soap fault on task retry : %#v", fault)
+	op.Debugf("unexpected soap fault on task retry : %s", fault)
```
emlin: somehow, print out all fields can help us understand what's wrong. Why change it back to %s, and curious why not for the above logFault?

pkg/vsphere/tasks/waiter.go
```diff
@@ -39,6 +47,10 @@ type Task interface {
 	WaitForResult(ctx context.Context, s progress.Sinker) (*types.TaskInfo, error)
 }
 
+type temporary interface {
+	Temporary() bool
```
emlin: is this for test purpose? didn't see test using it as well

pkg/vsphere/tasks/waiter.go
```diff
@@ -60,6 +72,23 @@ func WaitForResult(ctx context.Context, f func(context.Context) (Task, error)) (
 	var info *types.TaskInfo
 	var backoffFactor int64 = 1
 
+	// initialize the limiter channel once
+	once.Do(func() {
+		inFlight = make(chan struct{}, maxInFlight)
+	})
+
+	// reserve the spot
+	inFlight <- struct{}{}
```
emlin: This way will block the task directly without any message out.
Is thinking if do not want to have one more lock for available spots, might just get op before this, and log out message before and after get the spot.
So we could figure out the time gap to query one available task spot, to help debug.

dougm:
I think this PR is fine to limit the number of `task.Wait` calls (and hence connections), as it does not block task from completing on the ESX side.

But the issue isn't so much the tasks in flight (as they run on the ESX side), but the number of property collectors we have waiting within the VCH for the ESX/VC tasks to complete.  Each call to `task.Wait` creates its own property collector and connection to wait for the result.

We could instead wait for N tasks on a single connection, rather than the current 1-to-1.  Need to think about / research the best way to do this.  How about we discuss tomorrow?

Related to hostd connection limitations: https://github.com/vmware/vic/issues/5621#issuecomment-316849333

And somewhat related, #5647, we could could avoid starting a new task (such as ReconfigVM) until a current task has completed.

pkg/vsphere/tasks/waiter.go
```diff
@@ -60,6 +72,23 @@ func WaitForResult(ctx context.Context, f func(context.Context) (Task, error)) (
 	var info *types.TaskInfo
 	var backoffFactor int64 = 1
 
+	// initialize the limiter channel once
+	once.Do(func() {
```
hmahmood: You could do this in `init()`.

pkg/vsphere/tasks/waiter.go
```diff
@@ -60,6 +72,23 @@ func WaitForResult(ctx context.Context, f func(context.Context) (Task, error)) (
...
+	inFlight <- struct{}{}
+
+	// check whether we timed out/canceled
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+	}
+	// give the spot back
+	defer func() { <-inFlight }()
```
hmahmood: This should be on line 82.

pkg/vsphere/tasks/waiter.go
caglar10ur: It was dumping the struct but not showing the error (instead it is showing the address of the err field)

pkg/vsphere/tasks/waiter.go
caglar10ur: it is used for the type assertion below

caglar10ur:
@dougm agreed that we need to think more about this. I implemented this as a ugly/bad workaround for the bug in question but you are right this is not going to solve anything other than giving us some time

pkg/vsphere/tasks/waiter.go
caglar10ur: 👍  doing that but I think I won't be merging this until we discuss all the possible solutions first

mhagen-vmware:
So, I am still a bit confused by this.. My understanding is that 6.0 has a task limit of 1024 and 6.5 upped that to 4096. On an empty system with us creating 75 containers do we really exceed 1024 task limit?  In my parallel PR #5621 when I actively monitored the system the task queue only got up to about 40 and the skd endpoint stopped responding. That tells me that the underlying problem is not necessarily the amount of calls we are making but what calls we are making?

dougm:
@mhagen-vmware see my comment above "But the issue isn't so much the tasks in flight (as they run on the ESX side), but the number of property collectors we have waiting within the VCH for the ESX/VC tasks to complete. Each call to task.Wait creates its own property collector and connection to wait for the result."

In other words, we need to reduce/limit the number of *connections* we have open to ESX.   The number of *tasks* running on the ESX side isn't the issue here.

mhagen-vmware:
Right, understood.. I saw that, but this PR will then only coincidentally solve this problem without directly solving the problem then right?  Or even worse this PR will just give us a potentially major performance hit without actually solving the underlying problem?

caglar10ur:
I think the task limit and being able create that many tasks concurrently are different things. What I'm observing is happening only if you are doing things in parallel. As far my test goes we can't send more than 50-60 tasks in parallel against a single desktop PC that runs ESXi. This number does not include direct calls so I'm guessing we are hitting number of connections per session limit (which is 100 I believe)

caglar10ur:
@mhagen-vmware @dougm @hickeng I'll schedule a short meeting for Monday to discuss this in detail and in the meantime I'll experiment a bit to have at least an idea what might be going on :)

hmahmood:
@caglar10ur can you add me to the meeting too please?

caglar10ur:
Change of plans, I started to limit number of  SOAP requests in-flight

mhagen-vmware:
Parallel test build with this patch:
https://ci.vcna.io/vmware/vic/12602

pkg/vsphere/session/session.go
```diff
@@ -86,6 +91,32 @@ type Session struct {
 	folders *object.DatacenterFolders
 }
 
+// RoundTripFunc alias
+type RoundTripFunc func(*http.Request) (*http.Response, error)
+
+// RoundTrip method
+func (rt RoundTripFunc) RoundTrip(r *http.Request) (*http.Response, error) {
```
hmahmood: Is this used anywhere?

caglar10ur: Yep, without it

```
building Portlayer API server...
# github.com/vmware/vic/pkg/vsphere/session
pkg/vsphere/session/session.go:119: cannot use RoundTripFunc(func literal) (type RoundTripFunc) as type http.RoundTripper in return argument:
        RoundTripFunc does not implement http.RoundTripper (missing RoundTrip method)
```

caglar10ur:
@mhagen-vmware what is the setup in that test? Do we use a single esxi/single vch for multiple tests or we have multiple vch or esxi or both?

mhagen-vmware:
@caglar10ur 1 ESX with up to 4 VCH during the run.  The ESX is a 4 CPU/8Core 32GB mem, SSD drives.

caglar10ur:
bumped the default MaxIdleConnsPerHost from2 to maxInFlight this way we don't keep opening and closing connections instead reuse them (ESXi seems to be counting TIME_WAIT as a connection too so I believe this way we don't exceed the connection limit)

This is what I observe while creating 100 containers concurrently
```
Every 1.0s: govc host.esxcli network ip connection list | grep ":443"                                                                                                                                                        Tue Jul 25 22:50:26 2017

tcp      0        0        10.118.69.221:443                       10.118.67.228:55716  ESTABLISHED  66764     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.225:59916  ESTABLISHED  66764     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.225:59914  ESTABLISHED  66764     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.225:59912  ESTABLISHED  66764     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.225:59906  ESTABLISHED  66764     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.225:59904  ESTABLISHED  66763     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.225:59896  ESTABLISHED  66763     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.225:59892  ESTABLISHED  66764     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.225:59890  ESTABLISHED  66764     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.225:59888  ESTABLISHED  66764     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.225:59872  ESTABLISHED  66764     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.225:59862  ESTABLISHED  66764     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.225:59842  ESTABLISHED  66764     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.225:59834  ESTABLISHED  66764     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.225:59390  ESTABLISHED  66763     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.69.149:46886  ESTABLISHED  66763     newreno  rhttpproxy-IO
tcp      0        0        10.118.69.221:443                       10.118.67.228:50734  ESTABLISHED  66764     newreno  rhttpproxy-IO
tcp      0        0        0.0.0.0:443                             0.0.0.0:0            LISTEN       66758     newreno  rhttpproxy
tcp      0        0        :::443                                  :::0                 LISTEN       66758     newreno  rhttpproxy
```

caglar10ur:
@mhagen-vmware do you mind trying with this version again?

mhagen-vmware:
@caglar10ur https://ci.vcna.io/vmware/vic/12637

caglar10ur:
just wanted to follow up, we are still working on this issue. Rate limiting the creation of vms uncovered another issue and currently debugging it with esxi/vcenter team.

caglar10ur:
Just finished creating/starting/stopping/removing 128 VMs concurrently and rate limiter limited the operations 16 at a time as expected. I started to believe what we are observing in parallel tests are just slowness due to doing lots of things in parallel.

pkg/vsphere/session/session.go
```diff
@@ -178,8 +211,19 @@ func (s *Session) Connect(ctx context.Context) (*Session, error) {
...
+			maxInFlight = i
+		}
+	}
+	// Limit the concurrenty of SOAP requests
+	if t, ok := soapClient.Transport.(*http.Transport); ok {
+		t.MaxIdleConnsPerHost = maxInFlight
+	}
+	soapClient.Transport = LimitConcurrency(soapClient.Transport, maxInFlight)
 
+	// TODO: option to set http.Client.Transport.TLSClientConfig.RootCAs
```
hickeng: Missing RootCAs here seems like a significant omission as it could prevent us from accepting a legitimate vSphere.
Not that we have a means to provide a CA for vSphere specifically or to be added to the system pool.

pkg/vsphere/session/session.go
```diff
@@ -282,6 +326,16 @@ func (s *Session) Populate(ctx context.Context) (*Session, error) {
...
 	}
 
+	if s.Datacenter != nil {
+		folders, err := s.Datacenter.Folders(ctx)
+		if err != nil {
+			errs = append(errs, fmt.Sprintf("Failure finding folders (%s): %s", s.DatacenterPath, err.Error()))
+		} else {
+			log.Debugf("Cached folders: %s", s.DatacenterPath)
+		}
+		s.VMFolder = folders.VmFolder
```
hickeng: This is being used without being checked elsewhere - should be a hard error if it cannot be set.

pkg/vsphere/tasks/waiter.go
```diff
@@ -152,6 +159,11 @@ func isRetryError(op trace.Operation, err error) bool {
 		if f, ok := err.(types.HasFault); ok {
 			logFault(op, f.Fault())
 		} else {
+			// retry the temporary errors
```
hickeng: There is a DoNotRetry error used in `pkg/fetcher` and I'm sure similar logic in `pkg/retry`. I have a todo item to get retry and tasks package working together or unified. This is just informational.

pkg/vsphere/tasks/waiter.go
hickeng: The purposes of the tasks package were:
1. retry in the case of temporary errors, such as InProgress
2. operate in a queue aware fashion (such as throttling)
3. provide easy access to error types and messages

We can start adding code to extract the messages from the embedded faults and print that. Dumping the structs is massively verbose and hard to parse.

Also of note is that we get this even for _expected_ errors because we have no way of indicating whether an error is expected or not.

internal/govmomi/host_internal_config_manager.go
```diff
@@ -0,0 +1,48 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
```
caglar10ur: Go documentation says it is possible ```any package in or under a directory named internal may be imported by packages rooted in the same subtree.``` but I not sure that is what we want as that naming dictates that only pkv/vsphere can import that directory.

infra/scripts/bash-helpers.sh
```diff
@@ -90,15 +89,35 @@ vic-admin () {
...
+vic-tail-portlayer() {
+    unset keyarg
+    if [ -e $HOME/.ssh/authorized_keys ]; then
+        keyarg="--authorized-key=$HOME/.ssh/authorized_keys"
+    fi
+
+    out=$($(vic-path)/bin/vic-machine-"$OS" debug --target="$GOVC_URL" --compute-resource="$COMPUTE" --name=${VIC_NAME:-${USER}test} --enable-ssh "$keyarg" --rootpw=password --thumbprint=$THUMBPRINT $*)
+    host=$(echo $out | grep DOCKER_HOST | awk -F"DOCKER_HOST=" '{print $2}' | cut -d ":" -f1 | cut -d "=" -f2)
+
+    sshpass -ppassword ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${host} tail -f /var/log/vic/port-layer.log
```
emlin: `-ppassword`
is that `-p password?`

lib/install/management/appliance.go
```diff
@@ -202,6 +205,21 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
...
+		if err := vm.Unregister(d.ctx); err != nil {
+			ok := false
+			if soap.IsSoapFault(err) {
+				// because internal
+				ok = soap.ToSoapFault(err).String == "vim.fault.ConcurrentAccess"
+			}
+			if !ok {
+				return fmt.Errorf("Failed to unregister VM %q: %s", vm.Reference(), err)
+			}
+			log.Debugf("Ignoring ConcurrentAccess error for %s", vm.Reference())
```
emlin: what gonna happen if concurrent access error is returned? is that vm unregistered or still there

caglar10ur:
This PR includes the other PRs for now. I'll keep rebasing to make this smaller once the dependencies go in to the repo.

caglar10ur:
Removed all the other changes from this PR so this only does limiting but nothing else

caglar10ur:
@emlin can you take a look at this PR again? I believe this now works without requiring the internal llpm calls

emlin:
>@emlin can you take a look at this PR again? I believe this now works without requiring the internal llpm calls

Still didn't look yet. But you mean you'll not change to LLPM?

lib/install/management/appliance.go
```diff
@@ -191,17 +191,30 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
...
+	//removes the vm from vsphere, but detaches the disks first
 	_, err = vm.WaitForResult(d.ctx, func(ctx context.Context) (tasks.Task, error) {
 		return vm.DeleteExceptDisks(ctx)
 	})
 	if err != nil {
-		err = errors.Errorf("Failed to destroy VM %q: %s", vm.Reference(), err)
-		err2 := vm.Unregister(d.ctx)
-		if err2 != nil {
-			return errors.Errorf("%s then failed to unregister VM: %s", err, err2)
+		f, ok := err.(types.HasFault)
```
emlin: hmm.. for vic-machine, if we reach here, we already powered off the VM, and plan to delete the vm even there is error happens, I guess.
If user wants to delete the VCH together with container, if we differentiate the error type here, user might have to rerun the delete, and does not care about the error anyway.

caglar10ur: Good point, I'll fix the logic

lib/portlayer/exec/container.go
```diff
@@ -596,6 +596,8 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
 				log.Errorf("Error while attempting to unregister container VM: %s", err)
 				return err
 			}
+		case *types.ConcurrentAccess:
```
emlin: is this error happens during vm delete or unregister? I think that's in unregister if you're using llpm, but not sure for here.
btw, if this error happens, do you need a return here?

pkg/vsphere/session/session.go
```diff
@@ -180,8 +213,19 @@ func (s *Session) Connect(ctx context.Context) (*Session, error) {
 
 	soapClient.SetThumbprint(soapURL.Host, s.Thumbprint)
 
-	// TODO: option to set http.Client.Transport.TLSClientConfig.RootCAs
+	maxInFlight := defaultMaxInFlight
+	if e := os.Getenv("VIC_MAX_IN_FLIGHT"); e != "" {
+		if i, err := strconv.Atoi(e); err == nil {
+			maxInFlight = i
+		}
```
emlin: might an else here, to log out the env setting is wrong.

pkg/vsphere/vm/vm.go
```diff
@@ -262,10 +263,21 @@ func (vm *VirtualMachine) DeleteExceptDisks(ctx context.Context) (*object.Task,
...
 	disks := devices.SelectByType(&types.VirtualDisk{})
-
 	err = vm.RemoveDevice(ctx, true, disks...)
 	if err != nil {
-		return nil, err
+		switch f := err.(type) {
+		case task.Error:
+			switch f.Fault().(type) {
+			case *types.ConcurrentAccess:
+				log.Errorf("RemoveDevice failed for %s with ConcurrentAccess error. Ignoring it", vm)
```
emlin: Just curious, does this mean someone else is removing device? Cause the device removing is critical for us, I have to confirm for this. 😄

lib/portlayer/exec/container.go
caglar10ur: I need to document those quirks, without llpm we are getting occasional ConcurrentAccess errors during DeleteExceptDisk, that's why we are ignoring this

pkg/vsphere/session/session.go
caglar10ur: VIC_MAX_IN_FLIGHT is useful only during development so I ignored this but let me know if you really want t that log line :)

lib/install/management/appliance.go
```diff
@@ -191,17 +191,31 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
...
-		if err2 != nil {
-			return errors.Errorf("%s then failed to unregister VM: %s", err, err2)
+		if f, ok := err.(types.HasFault); ok {
+			switch f.Fault().(type) {
+			case *types.ConcurrentAccess:
+				log.Warnf("DeleteExceptDisks failed with ConcurrentAccess error. Ignoring it.")
+				ignore = true
+			}
+		}
+		if !ignore {
```
emlin: 👍

PR #5787, opened 2017-07-21
[specific ci=1-42-Docker-Push] Add cross repository blob mount to docker push

chengwang86:
- add `[]V2MetaData` (from vanilla docker) to `ImageWithMeta` to record the source repositories of each layer. Then we can obtain a list of source repositories to mount the layer from during docker push. If the layer to be uploaded can be mount from one of these source repos, there is no need to actually upload the layer to the original target repo. This is called `cross repository blob mount`.

- update the V2MetaData of each layer after an image has been successfully pulled or pushed.

- currently only the `SourceRepository` attribute in `V2MetaData` is used for `cross repo blob mount`. However, I leave `Digest` and `HMAC` there to be consistent with vanilla docker and for future optimization, since vanilla docker uses such information to choose the mount candidates from the list of source repos more intelligently and avoid unnecessary mount trials.

### DIFF
--- a/lib/imagec/imagec.go
+++ b/lib/imagec/imagec.go
@@ -673,8 +673,27 @@ func (ic *ImageC) prepareTransfer(ctx context.Context, imageOperation string) er
 
 // pullManifest attempts to pull manifest for an image.  Attempts to get schema 2 but will fall back to schema 1.
 func (ic *ImageC) pullManifest(ctx context.Context) error {
+
+	// Attempt to get schema2 manifest
+	manifest, digest, err := FetchImageManifest(ctx, ic.Options, 2, ic.progressOutput)
+	if err == nil {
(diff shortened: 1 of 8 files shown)


lib/imagec/imagec.go
```diff
@@ -602,7 +636,15 @@ func (ic *ImageC) prepareTransfer(ctx context.Context) error {
...
 
 	// Get the URL of the OAuth endpoint
-	url, err := LearnAuthURL(ic.Options)
+	var url *url.URL
+	if imageOperation == PushImage {
+		url, err = LearnAuthURLForPush(ic.Options, ic.progressOutput)
+	} else if imageOperation == PullImage {
+		url, err = LearnAuthURL(ic.Options)
+	} else {
+		err = fmt.Errorf("invlid image operation: %s", imageOperation)
```
sgairo: "invalid"

PR #5790, opened 2017-07-21
Check if vmdk owner is appliance

hickeng:
Adds check for vmdk owner being the appliance and basic retry of offline path.
Corrects some errors with disk handling.

### DIFF
--- a/lib/portlayer/storage/vsphere/container.go
+++ b/lib/portlayer/storage/vsphere/container.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/vic/lib/archive"
+	"github.com/vmware/vic/lib/guest"
 	"github.com/vmware/vic/lib/portlayer/storage"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/datastore"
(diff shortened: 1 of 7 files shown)


lib/portlayer/storage/vsphere/container.go
```diff
@@ -103,6 +104,10 @@ func (c *ContainerStore) NewDataSource(op trace.Operation, id string) (storage.D
 		return nil, err
 	}
 
+	offlineAttempt := 0
+offline:
```
morris-jason: weird spacing?

hickeng: `gofmt` - it surprises me every time I see it, but this idiomatic.

lib/portlayer/storage/vsphere/export.go
```diff
@@ -75,6 +81,14 @@ func (v *VolumeStore) NewDataSource(op trace.Operation, id string) (storage.Data
 	}
 
 	for _, o := range owners {
+		// sanity check to see if we are the owner - this should catch transitions
+		// from container running to diff or commit for example between the offline attempt and here
+		self, _ := guest.IsSelf(op, o)
+		if self && offlineAttempt < 2 {
+			op.Infof("Appliance is owner of online vmdk - retrying offline source path")
+			goto offline
```
matthewavery: First time I have seen a goto used in golang, cool :)

Issue #5810, opened 2017-07-24
vic-ui build task fails for no obvious reason


jooskim:
@mhagen-vmware reported that the 'vic-ui' task failed breaking build 12560. Based on the log messages, it seems that it is the Angular build that failed, but for some reason the logs were not captured.

![image](https://user-images.githubusercontent.com/3834071/28545384-6cdb2c00-708c-11e7-828b-09b298820cea.png)

**Acceptance Criteria**
vic-ui build should capture log & error messages, store them into a file and upload it to Google Cloud.

jooskim:
build 12577 is also impacted. marking it priority/high

jooskim:
Figured out a detailed error message:

![image](https://user-images.githubusercontent.com/3834071/28548554-21d6c4ae-709a-11e7-8eec-45e5fcda6216.png)

This confirms that the build failure has to do with some (Angular-related) dependency. Doing more investigation to figure out which module is the root cause.

jooskim:
https://github.com/angular/angular-cli/issues/7113

PR #5811, opened 2017-07-24
Fix broken VIC UI build

jooskim:
This PR helps figure out causes of issues like #5810 by ensuring logs and errors during plugin builds are captured and uploaded to Google Cloud.


jooskim:
Closing this PR, as I figured out what the exact cause is, and the current branch name is misleading.

PR #5812, opened 2017-07-25
Fix yarn version and address angular bug

### LINKED ISSUE #5810: vic-ui build task fails for no obvious reason

jooskim:
A new critical Angular issue was filed on 7/23/17 for angular-cli which
breaks the production VIC UI plugin build. Production build uses
Ahead-of-Time compiler which is impacted by the bug, and can be
addressed by pinpointing the `enhanced-resolve` package version to 3.3.0.
Also, the latest version of the Yarn package manager has its own bug in
dependency resolving strategy which won't allow us to use any other
version for the `enhanced-resolve` package but v3.4.1 which is declared
as a dependency of `webpack`. As a result, the following packages have
been pinned to specific versions that are known to work fine.

`yarn@0.24.6`
`enhanced-resolve@3.3.0`

Fixes #5810


Makefile
```diff
@@ -407,7 +407,7 @@ ENV_FLEX_SDK_HOME = "/tmp/sdk/flex_sdk_min"
 ENV_HTML_SDK_HOME = "/tmp/sdk/html-client-sdk"
 
 vic-ui-plugins:
-	@npm install -g yarn > /dev/null
+	@npm install -g yarn@0.24.6 > /dev/null
```
mhagen-vmware: This should be pinned in the integration image

ui/build-ui-plugins.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
```
mhagen-vmware: I really wish you would break up you PRs more :)

jooskim: yeah i did sneak in some extra stuff here because i thought it would be nice to address lack of ui build logs getting in google cloud. will do better next time

PR #5830, opened 2017-07-25
[specific ci=1-23-Docker-Inspect.robot]show container volume information correctly after VCH restart

### LINKED ISSUE #5558: Add mount data from portlayer to replenish persona container cache on VCH restart
As a user of VIC, I expect docker inspect to return mount information for a container after a VCH restarts.

Acceptance Criteria
- [ ] Updated robot scripts to verify that the docker inspect returns valid mount data after a VCH restarts

---
This is the second part of the docker inspect/mount data issue.  This task is to add either a new portlayer call to get the mount data for a container or add the data to our current ContainerInfo call.  It also requires replenishing the container cache on start.  This missing data is needed for docker cp.

### LINKED ISSUE #5845: docker cp fails to target correct device for anonymous volumes
When trying to copy to or copy from a container with anonymous volume, we fail to get the correct device when the target is the anonymous volume.

Steps to reproduce:
`docker create -i -v /vol1 busybox`
`docker cp test.txt cid:/vol1`

Right now if we repeat  that cp command multiple times, we can see that the personality gives a different device ID each time. And the deviceID given by the personality is not consistent with what we see in `docker volume ls` or `docker inspect cid`

Cause:
mountsFromContainer uses processVolumeParam which is not the right way to look up anonymous volumes because it creates a new UUID for anonymous volumes instead of looking it up.

emlin:
fixes #5558 
also fixes #5845

### DIFF
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -389,9 +389,6 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
 		return handle, BadRequestError(err.Error())
 	}
 
-	// Reset hostconfig binds because we don't have bind mount supported. In this way we'll get consistent result for container mount after VCH restart
-	config.HostConfig.Binds = []string{}
-
 	log.Infof("Finalized Volume list : %#v", volList)
 
 	// Create and join volumes.
@@ -1653,8 +1650,14 @@ func mountsFromContainerInfo(vc *viccontainer.VicContainer, info *models.Contain
 		return nil
 	}
 
-	// Iterate through info.VolumeConfig and build the config.volumes
-	// We don't have bind mount supported, so do not set hostconfig.Binds
+	if vc.HostConfig == nil {
+		vc.HostConfig = &container.HostConfig{}
+	}
+	if vc.Config == nil {
+		vc.Config = &container.Config{}
+	}
+	// Iterate through info.VolumeConfig and build the config.volumes and hostconfig.binds
+	// We don't know where the mounts come from, so just fill all of them to both config.volumes and hostconfig.binds
 	vc.Config.Volumes = make(map[string]struct{}, len(info.VolumeConfig))
 	for _, volume := range info.VolumeConfig {
 		mounts := make([]string, 2)
@@ -1665,6 +1668,7 @@ func mountsFromContainerInfo(vc *viccontainer.VicContainer, info *models.Contain
 		}
 		mount := strings.Join(mounts, ":")
 		vc.Config.Volumes[mount] = struct{}{}
+		vc.HostConfig.Binds = append(vc.HostConfig.Binds, mount)
 		log.Debugf("added volume mount: %s", mount)
 	}
 	// Derive the mount data
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -1491,14 +1491,12 @@ func processVolumeParam(volString string) (volumeFields, error) {
 func processVolumeFields(volumes []string) (map[string]volumeFields, error) {
 	volumeFields := make(map[string]volumeFields)
 
-	for i, v := range volumes {
+	for _, v := range volumes {
 		fields, err := processVolumeParam(v)
 		log.Infof("Processed volume arguments: %#v", fields)
 		if err != nil {
 			return nil, err
 		}
-		// replace container volumes string with correct volume id, to avoid regeneration during inspect
-		volumes[i] = getMountString(fields.ID, fields.Dest, fields.Flags)
 		volumeFields[fields.Dest] = fields
 	}
 	return volumeFields, nil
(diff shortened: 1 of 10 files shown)


sflxn:
@emlin , Archive.go is still using mountsFromContainer().  Can you look at that and determine if we should still call that function or just use the cached mount data?

emlin:
@sflxn the mountsFromContainer() is getting mount data from cached hostconfig.binds and config.volumes, which is filled up in persona init or cached by container create.
To make sure that works well, I removed the method mountsFromContainerInfo, so inspect and others all call the same method to get out same mount value.
So now you're good to call that safely with correct mount.
I didn't cache mounts directly in persona, because that will need the same logic to get mounts from those values after container is created. I feels like no need to cache duplicated data and that might potentially cause inconsistency in the future.

chengwang86:
Since you have added new test files in 10-1-VCH-restart, I think you should also test your PR with specific ci=10-1-VCH-Restart.

tests/test-cases/Group10-VCH-Restart/10-01-VCH-Restart.robot
```diff
@@ -131,3 +131,22 @@ Create VCH attach disk and reboot
     Wait Until Keyword Succeeds  20x  5 seconds  Run Docker Info  %{VCH-PARAMS}
     ${rc}=  Run And Return Rc  govc device.ls -vm=%{VCH-NAME} | grep disk
     Should Be Equal As Integers  ${rc}  1
+
+Docker inspect mount data after reboot
+    ${rc}  ${container}=  Run And Return Rc And Output  docker %{VCH-PARAMS} volume create --name=named-volume
+    Should Be Equal As Integers  ${rc}  0
+    ${rc}  ${container}=  Run And Return Rc And Output  docker %{VCH-PARAMS} create --name=mount-data-test -v /mnt/test -v named-volume:/mnt/named busybox
+    Should Be Equal As Integers  ${rc}  0
+    ${rc}  ${out}=  Run And Return Rc And Output  docker %{VCH-PARAMS} inspect -f '{{.Mounts}}' mount-data-test
```
matthewavery: currently there is another issue with `inspect` where the `mountsFromContainer` function returns a different UUID for an anonymous volume each time(this is a bug). Because of this the output for anonymous volumes may be incorrect everytime inspect is called. You might check for this after your current PR. The behavior is likely the same and will need a separate fix to persist the appropriate information.

matthewavery: This is not an issue for you, it is a separate matter. I just wanted to note that this test could potentially be affected by it at the moment.

emlin: Talked with @matthewavery directly. The new issue is not fixed by this PR, but that problem is not there after VCH restart, because that time all volumes already have an ID.
Cause the new issue is already found, and it's related to this fix, I'd like to continue the work, to fix both issues.

emlin:
> Since you have added new test files in 10-1-VCH-restart, I think you should also test your PR with specific ci=10-1-VCH-Restart.

I just want to make sure it does not break docker test. This restart test already passed locally. I'll run that specific ci before I merge 😄

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1478,12 +1491,14 @@ func processVolumeParam(volString string) (volumeFields, error) {
...
 
-	for _, v := range volumes {
+	for i, v := range volumes {
 		fields, err := processVolumeParam(v)
 		log.Infof("Processed volume arguments: %#v", fields)
 		if err != nil {
 			return nil, err
 		}
+		// replace container volumes string with correct volume id, to avoid regeneration during inspect
+		volumes[i] = getMountString(fields.ID, fields.Dest, fields.Flags)
```
emlin: this should be deleted

lib/config/executor/container_vm.go
```diff
@@ -53,6 +53,7 @@ const (
 	Outbound
 	Peers
 )
+const Mode = "Mode"
```
hickeng: This should be in the `lib/portlayer/constants` package.

emlin: hmm.. can we have persona code use portlayer constant directly?

matthewavery:
@emlin looks great! I was wondering if it would be possible for you to do a check that the uuid reported by `inspect` is not in line with the an ID that shows up in the `volume ls` command. This would ensure that the uuid that we are getting is now a valid volume for the deployment. This is really not a major request and if it would take much time at all I would not worry about it for now.

matthewavery:
LGTM

sflxn:
lgtm

emlin:
> I was wondering if it would be possible for you to do a check that the uuid reported by inspect is now in line with the an ID that shows up in the volume ls command. This would ensure that the uuid that we are getting is now a valid volume for the deployment. This is really not a major request and if it would take much time at all I would not worry about it for now.

I thought about that during this PR, but that will require to query the volume id from inspect in the robot script. Might take longer time to do so.

PR #5832, opened 2017-07-25
Improve disk in use error for conflicting starts

morris-jason:
Fixes 5562


matthewavery:
lgtm - sorry I thought I approved this earlier

Issue #5717, opened 2017-07-14
Ensure Useful errors are propagated to the client on copy failure


hickeng:
Applies to #5700 

Currently a copy failure may result in silent return (presumably a missing error return somewhere in the DataSink Import path), or a singularly useless error about Server internal error calling "archive reader" or "archive writer" returning an error.

If the error is anything user remediable then the user should get a useful message, for example:
* no such directory
* permission denied
* no such file

This may well include returning usefully typed errors from the portlayer swagger calls, e.g. 422 in the case of a failure to decode the filter spec.

jialin-li:
Fix is merged in feature/docker-cp

PR #5835, opened 2017-07-26
Resolve chown errors with missing tar headers [specific ci=Group1-Docker-Commands] 

### LINKED ISSUE #5715: Add tether support for label relative paths
Applies to #5700 

To support online copy to volumes, we locate the current owner of a volume to perform the copy. There is no assurance that this is the container by which the command was initiated (that container could be offline and the volume in use by another), or that the path on which the volume is mounted is the same.
As such we will address the volume via `label://disk-label/sub/path/on/volume` and the tether should understand this construction.

The toolbox already has URL support for the file transfer paths and can switch implementations based on scheme.

### LINKED ISSUE #5716: Investigate and address client side errors for chown and xattr during copy
Applies to #5700 

Currently there are client side errors when copying files from a container:
* operation not permitted - I believe this is related to an attempt to chown the file to `root` while running as a non-root user. This should be aligned with regular docker behaviour - we may need to pass configuration as to whether uid/gid should be included in the resulting archive.
* operation not supported - I believe this is related to the extended attributes added to denote addition/change/deletion. These should be framed as optional so that the client doesn't object during unpack.

It is possible that this issue go away if moving adopting the `pkg/archive` code mentioned in #5714 so that should be addressed first.

morris-jason:
Fixes #5715 and fixes #5716 

#### Background
During a CopyFrom, docker will throw chown errors if the destination directory doesn't exist, and the source directory header is not included in the tar archive.

e.g. 
```bash
$ al-docker cp nginx:/usr/share/nginx/html test1
$ al-docker cp nginx:/usr/share/nginx/html/ test2
chown /Users/morrisjason/go/src/github.com/vmware/vic/test2: operation not permitted
$
```

This pr fixes the second case, where directory headers were missing from the tar archive when a trailing slash was included in the source path. 
It also fixes the the auth bug that didn't allow online volume copies.

Fixes: #5716

### DIFF
--- a/lib/apiservers/engine/backends/archive_test.go
+++ b/lib/apiservers/engine/backends/archive_test.go
@@ -235,7 +235,7 @@ func TestFindArchiveReaders(t *testing.T) {
 					rebase:  "etc",
 					strip:   "etc",
 					exclude: []string{""},
-					include: "etc/",
+					include: "etc",
 				},
 			},
 		},
--- a/lib/archive/util_test.go
+++ b/lib/archive/util_test.go
@@ -528,6 +528,44 @@ func TestReadLevelOneDirpec(t *testing.T) {
 
 }
 
+func TestReadLevelOneNoTrailingSlash(t *testing.T) {
+	copyTarget := "/mnt"
+	direction := CopyFrom
+
+	mounts := []testMount{
+		{
+			mount:              "/",
+			CopyTarget:         copyTarget,
+			primaryMountTarget: true,
+			direction:          direction,
+		},
+	}
+
+	expectedFilterSpecs := map[string]FilterSpec{
+		"/": {
+			RebasePath: "mnt",
+			StripPath:  "mnt",
+			Exclusions: make(map[string]struct{}),
+			Inclusions: make(map[string]struct{}),
+		},
+	}
+
+	for _, v := range mounts {
+		actualFilterSpec := GenerateFilterSpec(v.CopyTarget, v.mount, v.primaryMountTarget, v.direction)
+		expectedFilterSpec := expectedFilterSpecs[v.mount]
+
+		if !assert.Equal(t, expectedFilterSpec.RebasePath, actualFilterSpec.RebasePath, "rebase path check failed (%s)", v.mount) {
+			return
+		}
+
+		if !assert.Equal(t, expectedFilterSpec.StripPath, actualFilterSpec.StripPath, "strip path check failed (%s)", v.mount) {
+			return
+		}
+
+	}
+
+}
+
 func TestReadFileSpec(t *testing.T) {
 	copyTarget := "/mnt/vols/A/a/path/file.txt"
 	direction := CopyFrom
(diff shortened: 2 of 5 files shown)


lib/apiservers/engine/backends/archive_test.go
```diff
@@ -235,7 +235,7 @@ func TestFindArchiveReaders(t *testing.T) {
 					rebase:  "etc",
 					strip:   "etc",
 					exclude: []string{""},
-					include: "etc/",
+					include: "etc",
```
hickeng: can we have a test that explicitly has the trailing `/` to assert behaviour in that case.

lib/tether/toolbox.go
```diff
@@ -139,7 +139,9 @@ func (t *Toolbox) Reload(config *ExecutorConfig) error {
 	t.authIDs[config.ID] = struct{}{}
 	// we also allow any device IDs that are attached
 	for _, mspec := range config.Mounts {
-		t.authIDs[mspec.Source.Host] = struct{}{}
+		// mounstpect.source.path is the disk label for vmdks
+		// XXX: this is not the case for other volumes, eg nfs vols.
```
hickeng: the `mspec.Source.Scheme` should be `label` if I recall correctly for vmdk mounts and can be used as a check.

lib/tether/toolbox.go
```diff
@@ -389,7 +391,7 @@ func mountDiskLabel(label string) (string, error) {
 
 	// label exists, bind mount it to a tmp directory
 	tmpDir, err := ioutil.TempDir("", fmt.Sprintf("toolbox-%s", label))
-	if err := Sys.Syscall.Mount(path, tmpDir, "bind", syscall.MS_BIND, ""); err != nil {
+	if err := Sys.Syscall.Mount(path, tmpDir, "ext4", syscall.MS_NOATIME, ""); err != nil {
```
hickeng: There is a const for this `ext4FileSystemType`

does remind me - we should be check that we're mounting with NOATIME in the offline path.

lib/tether/toolbox.go
morris-jason: done.

lib/tether/toolbox.go
morris-jason: done.

lib/apiservers/engine/backends/archive_test.go
morris-jason: that's what this case is.

morris-jason:
Merging. Only remaining online test failure will be fixed by #5717

lib/apiservers/engine/backends/archive_test.go
matthewavery: The actual target has the trailing slash in this case @hickeng but we will expect that a target in this case will generate an include without the trailing `/` since we still want the directory for `docker cp` behavior. we will still exclude things ending with a `/` but for docker cp the filter spec include should not have a `/` since the behavior is the same `/` or no.

lib/archive/util_test.go
```diff
@@ -528,6 +528,44 @@ func TestReadLevelOneDirpec(t *testing.T) {
 
 }
 
+func TestReadLevelOneNoTrailingSlash(t *testing.T) {
```
matthewavery: I want to note here for ease of review that this is the same test as the above test except that the target has no `/` so we cover both cases here now.

lib/archive/util_test.go
```diff
@@ -727,6 +765,62 @@ func TestAddExclusionsNonNested(t *testing.T) {
 
 }
 
+func TestAddExclusionsTrailingSlash(t *testing.T) {
```
matthewavery: Additional ease of review. This test is the same as the below test except for the trailing `/` so we now test for proper behavior for both in the inclusion/exclusion list as well.

matthewavery:
LGTM - this looks good and this appears to pass the tests for volumes in the online case. I think pending @hickeng 's approval that we should get this merged.

matthewavery:
Please note that the build for this fails right now due to the import cycle. @jialin-li 's current PR will fix this.

PR #5842, opened 2017-07-26
[skip ci] Update vsan test bed to include DVS and DVPG

### LINKED ISSUE #5540: nightly test 5-6-2 exited silently when creating a distributed switch
Seen on vsphere 6.0

[5-6-2-VSAN-Complex.zip](https://github.com/vmware/vic/files/1089214/5-6-2-VSAN-Complex.zip)

This is from output.xml
start
```
<robot generated="20170620 11:17:56.585" generator="Robot 3.0.2 (Python 2.7.9 on linux2)">
```
end
```
<arg>\nCreate a distributed switch</arg>
</arguments>
<status status="PASS" endtime="20170620 12:04:45.044" starttime="20170620 12:04:45.043"></status>
</kw>
<kw name="Run" library="OperatingSystem">
<doc>Runs the given command in the system and returns the output.</doc>
<arguments>
<arg>govc dvs.create -product-version 5.5.0 -dc=${datacenter} ${dvs}</arg>
</arguments>
<assign>
<var>${out}</var>
</assign>
<msg timestamp="20170620 12:04:45.052" level="INFO">Running command 'govc dvs.create -product-version 5.5.0 -dc=vcqaDC test-ds 2&gt;&amp;1'.</msg>
<msg timestamp="20170620 12:06:15.072" level="INFO">${out} = govc: Post https://10.160.125.122/sdk: dial tcp 10.160.125.122:443: i/o timeout
```
I don't think this is due to nimbus timeout. Even if `govc dvs.create ...` failed, the test should still proceed.

mhagen-vmware:
fixes #5540

We have constant problems with nimbus instability in govc, this pushes the network creation at least into the test bed so that we can need to make less govc calls.

### DIFF
--- a/tests/manual-test-cases/Group5-Functional-Tests/5-6-1-VSAN-Simple.robot
+++ b/tests/manual-test-cases/Group5-Functional-Tests/5-6-1-VSAN-Simple.robot
@@ -21,7 +21,7 @@ Test Teardown  Run Keyword And Ignore Error  Nimbus Cleanup  ${list}
 Simple VSAN
     ${name}=  Evaluate  'vic-vsan-' + str(random.randint(1000,9999))  modules=random
     Set Test Variable  ${user}  %{NIMBUS_USER}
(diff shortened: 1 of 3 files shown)


mhagen-vmware:
@rajanashok we might need to make some updates at some point, but they really should not make any fundamental changes like that, otherwise a ton of vsphere tests would then need to be updated.

PR #5846, opened 2017-07-26
[specific ci=1-43-Docker-CP-Offline]fix import cycle and update tests

jialin-li:


matthewavery:
LGTM - pending a review from @hickeng This looks really good! fixes many things that needed fixing as well as cleans up our error messages significantly (in that in some cases we even returned an error :) )

matthewavery:
re lgtming :)

PR #5854, opened 2017-07-27
Add tar stripper for combining tar streams [specific ci=1-43-Docker-CP-Offline]

hickeng:
Adds a tar aware stripper that drops the end of archive records. Adds a
multi reader that can expliot it.


lib/archive/stripper_test.go
```diff
@@ -0,0 +1,311 @@
...
+		}
+		if !assert.Equal(t, n, size, "Expected file data size to match target generated size") {
+			t.FailNow()
+		}
+	}
+
+	wg.Wait()
+}
+
+func TestCombinedStrippersWithFinale(t *testing.T) {
```
jzt: 👍  on these test names.

PR #5855, opened 2017-07-27
Fix broken docker diff

jzt:
Xattrs were being excluded, which are required for diff to work. This fix adds the xattrs in all cases where the data is not requested (for now, just diff).

### DIFF
--- a/lib/portlayer/storage/vsphere/container.go
+++ b/lib/portlayer/storage/vsphere/container.go
@@ -314,7 +314,10 @@ func (c *ContainerStore) Export(op trace.Operation, id, ancestor string, spec *a
 		return nil, errors.New("mismatched datasource types")
 	}
 
-	tar, err := archive.Diff(op, fl.Name(), fr.Name(), spec, data, false)
+	// if we want data, include the xattrs, otherwise assume diff
+	xattrs := !data
+
+	tar, err := archive.Diff(op, fl.Name(), fr.Name(), spec, data, xattrs)
 	if err != nil {
 		go closers()
 		return nil, err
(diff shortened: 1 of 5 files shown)


matthewavery:
LGTM

lib/portlayer/storage/vsphere/container.go
```diff
@@ -314,7 +314,10 @@ func (c *ContainerStore) Export(op trace.Operation, id, ancestor string, spec *a
 		return nil, errors.New("mismatched datasource types")
 	}
 
-	tar, err := archive.Diff(op, fl.Name(), fr.Name(), spec, data, false)
+	// if we want data, include the xattrs, otherwise assume diff
```
hickeng: `if we want data exclude the xattrs...`

lib/portlayer/storage/vsphere/export.go
```diff
@@ -175,7 +175,10 @@ func (i *ImageStore) Export(op trace.Operation, id, ancestor string, spec *archi
 		return nil, fmt.Errorf("mismatched datasource types: %T, %T", ls, rs)
 	}
 
-	tar, err := archive.Diff(op, fl.Name(), fr.Name(), spec, data, false)
+	// if we want data, include the xattrs, otherwise assume diff
```
hickeng: ditto

pkg/vsphere/disk/disk_manager.go
```diff
@@ -508,7 +508,14 @@ func (m *Manager) AttachAndMount(op trace.Operation, datastoreURI *object.Datast
 		return "", err
 	}
 
-	return d.Mount(op, nil)
+	// don't update access time - that would cause the diff operation to mutate the filesystem
+	opts := []string{"noatime", ""}
```
hickeng: I've no idea why we have the extra "" in there - I don't think it's needed.

I'm also not sure what happens here when combined with a copyTo operation that needs a writable mount. The writable mount should have `persistent == true` but don't know whether that's actually possible at the same time as a non-persistent mount. Hopefully we're not using `!persistent` atm ;)

PR #5857, opened 2017-07-27
[specific ci=Group6-VIC-Machine] Vic machine leaks sessions

### LINKED ISSUE #5355: vic-machine should logout after operation is complete
**User Statement:**
As a viadmin I expect vSphere clients not to leak sessions

**Details:**
Currently vic-machine does not clean up it's sessions after exit, meaning they are left lying around in `old` state until session timeout occurs and they are collected. `session.Logout` should be called on the exit paths.

**Acceptance Criteria:**
`govc session.ls` does not show sessions associated with vic-machine after vic-machine has exited

gigawhitlocks:
Logs out of vSphere after each of the various vic-machine commands, and an added step to at least one of the tests for each of the vic-machine commands to confirm the change. In which I do the same thing in a bunch of places. Also, I tried to choose a 'basic' test for each of the tests to update, idk, I can move that stuff around if folks don't like it how it is.

Fixes #5355

### DIFF
--- a/cmd/vic-machine/configure/configure.go
+++ b/cmd/vic-machine/configure/configure.go
@@ -321,12 +321,12 @@ func (c *Configure) Run(clic *cli.Context) (err error) {
 	defer cancel()
 
 	validator, err := validate.NewValidator(ctx, c.Data)
-	defer validator.Session.Logout(ctx)
-
 	if err != nil {
 		log.Errorf("Configuring cannot continue - failed to create validator: %s", err)
 		return errors.New("configure failed")
 	}
+	defer validator.Session.Logout(ctx)
+
 	_, err = validator.ValidateTarget(ctx, c.Data)
 	if err != nil {
 		log.Errorf("Configuring cannot continue - target validation failed: %s", err)
--- a/cmd/vic-machine/configure/configure.go
+++ b/cmd/vic-machine/configure/configure.go
@@ -321,6 +321,8 @@ func (c *Configure) Run(clic *cli.Context) (err error) {
 	defer cancel()
 
 	validator, err := validate.NewValidator(ctx, c.Data)
+	defer validator.Session.Logout(ctx)
+
 	if err != nil {
 		log.Errorf("Configuring cannot continue - failed to create validator: %s", err)
 		return errors.New("configure failed")
--- a/tests/resources/VCH-Util.robot
+++ b/tests/resources/VCH-Util.robot
@@ -177,7 +177,6 @@ Install VIC Appliance To Test Server
     Get Docker Params  ${output}  ${certs}
     Log To Console  Installer completed successfully: %{VCH-NAME}...
 
-    No Leaked Sessions  vic-machine
     [Return]  ${output}
 
 Run VIC Machine Command
@@ -202,7 +201,6 @@ Run Secret VIC Machine Inspect Command
     [Arguments]  ${name}
     ${rc}  ${output}=  Run And Return Rc And Output  bin/vic-machine-linux inspect --name=${name} --target=%{TEST_URL}%{TEST_DATACENTER} --user=%{TEST_USERNAME} --password=%{TEST_PASSWORD} --thumbprint=%{TEST_THUMBPRINT}
 
-    No Leaked Sessions  vic-machine
     [Return]  ${rc}  ${output}
 
 Run VIC Machine Delete Command
@@ -283,7 +281,6 @@ Cleanup VIC Appliance On Test Server
     Log To Console  Deleting the VCH appliance %{VCH-NAME}
     ${output}=  Run VIC Machine Delete Command
     Run Keyword And Ignore Error  Cleanup VCH Bridge Network  %{VCH-NAME}
-    No Leaked Sessions  vic-machine
     [Return]  ${output}
 
 Cleanup VCH Bridge Network
@@ -398,7 +395,6 @@ Get VCH ID
     [Arguments]  ${vch-name}
     ${ret}=  Run  bin/vic-machine-linux ls --target=%{TEST_URL}%{TEST_DATACENTER} --thumbprint=%{TEST_THUMBPRINT} --user %{TEST_USERNAME} --password=%{TEST_PASSWORD}
     Should Not Contain  ${ret}  Error
-    No Leaked Sessions  vic-machine
     @{lines}=  Split To Lines  ${ret}
     :FOR  ${line}  IN  @{lines}
     \   # Get line with name ${vch-name}
@@ -435,7 +431,6 @@ Upgrade
     Should Contain  ${output}  Completed successfully
     Should Not Contain  ${output}  Rolling back upgrade
     Should Be Equal As Integers  ${rc}  0
-    No Leaked Sessions  vic-machine
 
 Upgrade with ID
     Log To Console  \nUpgrading VCH using vch ID...
--- a/tests/resources/VCH-Util.robot
+++ b/tests/resources/VCH-Util.robot
@@ -173,8 +173,11 @@ Install VIC Appliance To Test Server
     ${output}=  Run VIC Machine Command  ${vic-machine}  ${appliance-iso}  ${bootstrap-iso}  ${certs}  ${vol}  ${additional-args}
     Log  ${output}
     Should Contain  ${output}  Installer completed successfully
+
     Get Docker Params  ${output}  ${certs}
     Log To Console  Installer completed successfully: %{VCH-NAME}...
+
+    No Leaked Sessions  vic-machine
     [Return]  ${output}
 
 Run VIC Machine Command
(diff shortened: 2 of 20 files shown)


tests/test-cases/Group6-VIC-Machine/6-04-Create-Basic.robot
```diff
@@ -128,6 +128,9 @@ Create VCH - defaults
     Log To Console  Installer completed successfully: %{VCH-NAME}
 
     Run Regression Tests
+
+    ${sessions}=  Check vSphere Sessions  "vic-machine"
+    Should Be Equal As Integers  ${sessions}  0
```
mhagen-vmware: can you actually just add this to the regression tests keyword? Then you don't need it here and it will be run in a lot of other scenarios.

cmd/vic-machine/configure/configure.go
```diff
@@ -321,6 +321,8 @@ func (c *Configure) Run(clic *cli.Context) (err error) {
 	defer cancel()
 
 	validator, err := validate.NewValidator(ctx, c.Data)
+	defer validator.Session.Logout(ctx)
```
dougm: Can you add a test that uses an invalid target and/or credentials?  afiact, every `defer Logout` in this PR would panic in that case.. looks like the `defer Logout`'s should be after the err != nil checks.

gigawhitlocks: oh _good call_ @dougm I would _not_ have caught that

tests/test-cases/Group6-VIC-Machine/6-04-Create-Basic.robot
gigawhitlocks: I started refactoring this before I saw your comment because I was actually unhappy with the way I had this designed. I moved this check into a couple of keywords so that this check runs each time we do `vic-machine create` (I'm still running tests and making sure stuff works how I want) but I could, instead, do a single test that runs all of the vic-machine subcommands in Regression Tests, since maybe adding it to the keywords in the current version is overkill?.. I've been torn all day about which approach is better.

cmd/vic-machine/configure/configure.go
```diff
@@ -325,6 +325,8 @@ func (c *Configure) Run(clic *cli.Context) (err error) {
 		log.Errorf("Configuring cannot continue - failed to create validator: %s", err)
 		return errors.New("configure failed")
 	}
+	defer validator.Session.Logout(ctx)
```
emlin: there is one case that session is created successfully, but still returns error in the NewValidator method. Might also add one defer in that method

tests/test-cases/Group7-Regression/7-01-Regression.robot
```diff
@@ -29,3 +29,4 @@ Regression test
     ${ret}=  Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Run  govc vm.info %{VCH-NAME}
     Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Set Test Variable  ${out}  ${ret}
     Should Contain  ${out}  Photon - VCH
+    No Leaked Sessions  vic-machine
```
mhagen-vmware: Sorry, I meant - Docker-Util.robot/Run Regression Tests, which will ensure that it gets run in all of our tests that call that keyword, not just this test which only gets run on PRs.

tests/test-cases/Group6-VIC-Machine/6-02-Default.robot
```diff
@@ -23,3 +23,23 @@ Delete with defaults
...
+  Should Contain  ${ret}  vic-machine-linux inspect  failed
+
+  ${ret}=  Run  bin/vic-machine-linux ls --target %{TEST_URL} --thumbprint=%{TEST_THUMBPRINT} --user %{TEST_USERNAME} --password=INCORRECT
+  Should Contain  ${ret}  vic-machine-linux ls failed
+
+  ${ret}=  Run  bin/vic-machine-linux upgrade --target %{TEST_URL} --thumbprint=%{TEST_THUMBPRINT} --user %{TEST_USERNAME} --password=INCORRECT
+  Should Contain  ${ret}  vic-machine-linux upgrade failed
+
+  ${ret}=  Run  bin/vic-machine-linux configure --target %{TEST_URL} --thumbprint=%{TEST_THUMBPRINT} --user %{TEST_USERNAME} --password=INCORRECT
+  Should Contain  ${ret}  vic-machine-linux configure failed
```
mhagen-vmware: Did you intend to check for panic as well as per the name?

tests/resources/VCH-Util.robot
```diff
@@ -412,6 +415,8 @@ Install VIC with version to Test Server
     ${rc}  ${output}=  Run And Return Rc And Output  tar zxvf vic.tar.gz
     Set Environment Variable  TEST_TIMEOUT  20m0s
     Install VIC Appliance To Test Server  vic-machine=./vic/vic-machine-linux  appliance-iso=./vic/appliance.iso  bootstrap-iso=./vic/bootstrap.iso  certs=${false}  vol=default ${insecureregistry}
+
+    Clean VMOMI Sessions  vic-machine  # old versions of vic-machine may leak sessions, confusing other tests
```
mhagen-vmware: I would think we might actually need this in the install keyword instead/in-addition of delete keyword, if we want to make sure to cleanup previous builds.

tests/resources/Vsphere-Util.robot
```diff
@@ -234,3 +234,14 @@ Check VM Guestinfo
     ${rc}  ${output}=  Run And Return Rc And Output  govc vm.info -e ${vm} | grep ${str}
     Should Be Equal As Integers  ${rc}  0
     [Return]  ${output}
+
+Clean VMOMI Sessions
+    [Arguments]  ${user-agent}
+    Run  for x in $(govc session.ls); grep "${user-agent}" | xargs govc session.rm
```
mhagen-vmware: From a quick glance, I am thinking this is way too open.  I just took a quick look at govc session.ls and session.rm and I am not very familiar with it, but I am imagining this is going to wipe out any session related to vic-machine on the host, which would include other tests that might be running in parallel?

tests/resources/Vsphere-Util.robot
```diff
@@ -234,3 +234,14 @@ Check VM Guestinfo
...
+
+No Leaked Sessions
```
mhagen-vmware: Should Not Have Any VMOMI Sessions

tests/test-cases/Group6-VIC-Machine/6-02-Default.robot
gigawhitlocks: No, we just make sure it actually says "failed". If it says failed, it didn't panic, because the process exited normally.

gigawhitlocks: I'll add a step that checks for the word "panic" even though it's technically redundant it will make the test look more complete

tests/resources/Vsphere-Util.robot
gigawhitlocks: If that's the case my other check won't work either and random tests will fail if other tests are running at the same time as this check occurs. I didn't realize we were running tests in parallel against the same hosts. This test may not be possible to implement in the infrastructure as I now understand it.

tests/resources/VCH-Util.robot
gigawhitlocks: This isn't cleaning out an old build, it's cleaning up a VMOMI session from a version of vic-machine that contains the bug this PR fixes. If we added this to the install / delete keywords, then the actual test would be pointless because we would clean out all of the leaked vic-machine sessions and then check to see if there are any leaked sessions.

gigawhitlocks:
Currently rewriting tests since both my method of checking for existing sessions and removing them are incompatible with tests run in parallel against the same host

cmd/vic-machine/configure/configure.go
gigawhitlocks: Is this what you mean @emlin https://github.com/vmware/vic/pull/5857/commits/2440332e8421ce51726b54e6957db40444a26c55

gigawhitlocks:
sorry about all the extraneous whitespace in this PR 😬

cmd/vic-machine/configure/configure.go
emlin: yes, that's it

tests/test-cases/Group6-VIC-Machine/6-02-Default.robot
```diff
@@ -16,10 +16,82 @@
...
+    Log  ${output}
+    ${thumbprint}=  Get Thumbprint From Log  ${output}
+    Should Not Have VMOMI Session  ${thumbprint}
+
+    ${output}=  Run  bin/vic-machine-linux configure --target %{TEST_URL} --debug=1 --thumbprint=%{TEST_THUMBPRINT} --user %{TEST_USERNAME} --password=%{TEST_PASSWORD} --name=%{VCH-NAME}
+    Log  ${output}
+    ${thumbprint}=  Get Thumbprint From Log  ${output}
+    Should Not Have VMOMI Session  ${thumbprint}
+
+    ${output}=  Run  bin/vic-machine-linux delete --target %{TEST_URL} --debug=1 --thumbprint=%{TEST_THUMBPRINT} --user %{TEST_USERNAME} --password=%{TEST_PASSWORD} --name=%{VCH-NAME}
```
anchal-agrawal: This appears to be leaking network artifacts (`VCH-12949-9989-bridge` in logs from build https://ci.vcna.io/vmware/vic/12949). I'd suggest using `Cleanup VIC Appliance On Test Server` https://github.com/vmware/vic/blob/master/tests/resources/VCH-Util.robot#L275 instead since it cleans up the bridge network as well.

Issue #5847, opened 2017-07-26
Pull in code from push feature branch to help validate docker commit


sflxn:
As a user of VIC, I should be confident that images I create via docker commit works and all data within the newly created layers are valid.

Acceptance Criteria:
- [ ] imagec library has enough code to allow getting archive streams for layers when building a standalone imagec binary.
---

We have the functionality to pull layer archives in the current docker push feature branch.  We need to pull it out of there and merge it into VIC 1.2 master.  We also need to do some additions to expose the portlayer to imagec in standalone mode.

PR #5859, opened 2017-07-27
Refactor archive proxy out of persona's backend

### LINKED ISSUE #5741: Diff for layers failed
As a user of VIC, I should be able to perform docker push.

Acceptance Criteria:

dif between layers returns a reader that returns data

----

- image_cache doesn't properly implement importer/exporter
- storage_handler doesn't register new image stores when the persona calls CreateImageStore on startup
- archive/diff.go's Tar() returns a lot of errors

sflxn:
Refactor the archive proxy out of the persona's backend so that imagec
binary can use it to access the diff stream from the portlayer.  This
will be used to help validate docker commit.  It will also be used in
docker push in the future.

Added fix for storage portlayer that prevented streaming of diff
archives between two layers.

Added listlayers command to imagec binary to list all layers for an
image.

Small refactor of imagec.go.

Resolves #5741 and #5847

### DIFF
--- a/cmd/imagec/main.go
+++ b/cmd/imagec/main.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"io"
@@ -26,10 +27,16 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/docker/pkg/streamformatter"
 	"github.com/docker/docker/reference"
+	"github.com/go-openapi/runtime"
+	rc "github.com/go-openapi/runtime/client"
 	"github.com/pkg/profile"
 
+	"github.com/vmware/vic/lib/apiservers/engine/proxy"
+	apiclient "github.com/vmware/vic/lib/apiservers/portlayer/client"
+	vicarchive "github.com/vmware/vic/lib/archive"
 	"github.com/vmware/vic/lib/imagec"
 	"github.com/vmware/vic/pkg/version"
+	"github.com/vmware/vic/pkg/vsphere/sys"
 )
 
 var (
@@ -40,8 +47,9 @@ var (
 )
 
 const (
-	PullImage = "pull"
-	PushImage = "push"
+	PullImage  = "pull"
+	PushImage  = "push"
+	ListLayers = "listlayers"
 )
 
 // ImageCOptions wraps the cli arguments
@@ -85,7 +93,7 @@ func init() {
 	flag.StringVar(&imageCOptions.profiling, "profile.mode", "", "Enable profiling mode, one of [cpu, mem, block]")
 	flag.BoolVar(&imageCOptions.tracing, "tracing", false, "Enable runtime tracing")
 
-	flag.StringVar(&imageCOptions.operation, "operation", "pull", "Pull/push an image")
+	flag.StringVar(&imageCOptions.operation, "operation", "pull", "Pull image/push image/listlayers for image")
 
 	flag.StringVar(&imageCOptions.options.Registry, "registry", imagec.DefaultDockerURL, "Registry to pull/push images (default: registry-1.docker.io)")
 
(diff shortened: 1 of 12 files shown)


cmd/imagec/main.go
```diff
@@ -100,12 +108,18 @@ func init() {
 func main() {
 	defer func() {
 		if r := recover(); r != nil {
-			fmt.Fprintf(os.Stderr, string(sf.FormatError(fmt.Errorf("%s : %s", r, debug.Stack()))))
+			_, err := fmt.Fprintf(os.Stderr, string(sf.FormatError(fmt.Errorf("%s : %s", r, debug.Stack()))))
+			if err != nil {
+				//do this to pass security check
```
jzt: If this is for gas and you want to ignore the error here, you can just call `fmt.Fprintf(...)` above (and don't assign the output to anything), with a `#nosec` on the line above.

cmd/imagec/main.go
```diff
@@ -169,9 +184,78 @@ func main() {
 		if err := ic.PullImage(); err != nil {
 			log.Fatalf("Pulling image failed due to %s\n", err)
 		}
-	} else if imageCOptions.operation == PushImage {
+	case PushImage:
 		log.Errorf("The operation '%s' is not implemented\n", PushImage)
-	} else {
+	case ListLayers:
+		options := imageCOptions.options
```
jzt: Minor: should we rename `imageCOptions` to `imagecOptions`? I don't think the `c` was ever meant to be capitalized. @caglar10ur do you recall?

cmd/imagec/results.json
```diff
@@ -0,0 +1,9 @@
+{
```
jzt: Should these `results.json` files be checked in?

lib/apiservers/engine/backends/results.json
```diff
@@ -0,0 +1,9 @@
+{
```
jzt: Maybe checked in with a `git add .` ?

cmd/imagec/results.json
sflxn: doh

cmd/imagec/main.go
```diff
@@ -169,9 +184,78 @@ func main() {
...
+func writeArchiveFile(archiveProxy proxy.VicArchiveProxy, layerID, parentID, archivePath string) error {
+	var filterSpec vicarchive.FilterSpec
+
+	host, err := sys.UUID()
+	if err != nil {
+		return err
+	}
+
+	//Initialize an archive stream from the portlayer for the layer
+	ar, err := archiveProxy.ArchiveExportReader(context.Background(), host, host, layerID, parentID, true, filterSpec)
```
chengwang86: can `filterSpec` be nil?

chengwang86:
I found ```--- FAIL: TestCreateImageStore``` from your build failure. And I observed the same error in the ci build for feature/docker-push. So I guess something in this code might have broken that unit test.

lib/apiservers/engine/backends/archive.go
```diff
@@ -16,19 +16,21 @@ package backends
...
 	"archive/tar"
-	"context"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
 	"path"
 	"strings"
 
+	"golang.org/x/net/context"
```
chengwang86: why don't we use the first `context`?

lib/imagec/imagec.go
```diff
@@ -463,11 +464,78 @@ func (ic *ImageC) CreateImageConfig(images []*ImageWithMeta) (metadata.ImageConf
...
 	defer cancel()
 
+	// Authenticate, get URL, get token
+	if err := ic.prepareTransfer(ctx); err != nil {
+		return err
+	}
+
+	// Output message
+	tagOrDigest := tagOrDigest(ic.Reference, ic.Tag)
+	progress.Message(ic.progressOutput, "", tagOrDigest+": Fetching layers from "+ic.Image)
```
chengwang86: This is not needed since we are not really fetching layers.

chengwang86:
Overall lgtm; but the unit test has to be fixed

cmd/imagec/main.go
```diff
@@ -169,9 +184,78 @@ func main() {
...
+}
+
+func archiveProxy(portLayerAddr string) proxy.VicArchiveProxy {
+	plClient := portlayerClient(portLayerAddr)
+	archiveProxy := proxy.NewArchiveProxy(plClient)
+
+	return archiveProxy
+}
+
+func writeArchiveFile(archiveProxy proxy.VicArchiveProxy, layerID, parentID, archivePath string) error {
```
chengwang86: Is this function used anywhere? Or is it left here for Emma to call in her PR?

lib/apiservers/engine/backends/archive.go
sflxn: The version of docker we are pinned to (1.13) still uses this context.  Newer version may use latest go context.

lib/imagec/imagec.go
sflxn: It's a status message.  It's not needed, but it's useful.

sflxn:
ping

jzt:
🏓

jzt:
lgtm

PR #5865, opened 2017-07-28
[specific ci=1-*Docker-CP-*] Docker CP Feature Implementation

### LINKED ISSUE #769: Milestone: docker cp
This has two very distinct parts:
1. stopped container - via portlayer storage
2. live container - via tether hgfs toolbox support

### LINKED ISSUE #4106: Portlayer: copy API
**Story**
As a Port Layer caller I want a copy API that doesn't require knowledge of target state
As a Port Layer caller I want to be able to copy files to a target container
As a Port Layer caller I want to be able to copy files from a target container

### LINKED ISSUE #5295: Portlayer: offline copy via disk attach

### LINKED ISSUE #5296: Portlayer: online copy via toolbox
**Story**
As a VIC developer I want a copy API implementation for when the target container is powered *on*
As a Port Layer caller I want a copy API that doesn't require knowledge of target state

**Detail**
Copying files to/from a container disk while the container is *on* requires using the DnD (drag and drop) mechanisms from toolbox guest operations.
Parallel copying of files should be viable, but for consistency with offline copy we should block this.
Attempts to power off container while copy is in progress should block until copy is complete, or until force is specified.

**Acceptance**
- [ ] file copy into container - container shows file structure in container
- [ ] file copy out of container - file structure present on client system
- [ ] parallel copy requests do not collide
- [ ] container stop should not attempt to power off the container while copy is in progress unless --force is specified
- [ ] copying file out of container should not block writing of that file in the container (e.g. copying a live log).

matthewavery:
This pr closes #769, closes #4106, closes #5295, and closes #5296. 
fixes #769 
fixes #4106 
fixes #5295 
fixes #5296


lib/archive/diff.go
```diff
@@ -51,14 +54,17 @@ func Diff(op trace.Operation, newDir, oldDir string, spec *FilterSpec, data bool
 		}
 	}
 
+	// this is in the case that we are part of a read that is crossing a mount point.
+	// It should be addressed by include/excludes.
```
jzt: Remove this comment

lib/archive/diff.go
```diff
@@ -51,14 +54,17 @@ func Diff(op trace.Operation, newDir, oldDir string, spec *FilterSpec, data bool
...
 	changes, err := docker.ChangesDirs(newDir, oldDir)
 	if err != nil {
 		return nil, err
 	}
-
+	// this is in the case that we are part of a read that is crossing a mount point.
```
jzt: Remove this comment

lib/archive/diff.go
```diff
@@ -84,28 +109,33 @@ func Tar(op trace.Operation, dir string, changes []docker.Change, spec *FilterSp
 				op.Errorf("Error closing tar writer: %s", cerr.Error())
 			}
 			if err == nil {
-				op.Debugf("Closing down tar writer with clean exit: %s", cerr)
+				if cerr != nil {
+					op.Errorf("Closing down tar writer with clean exit: %s", cerr)
+				} else {
+					op.Infof("Closing down tar writer with pristine exit")
```
jzt: Should this be `Debugf`?

lib/archive/diff.go
```diff
@@ -162,35 +195,40 @@ func createHeader(op trace.Operation, dir string, change docker.Change, spec *Fi
...
-		hdr.Name = filepath.Join(spec.RebasePath, change.Path)
-
-		if hdr.Typeflag == tar.TypeDir {
-			hdr.Name += "/"
-		}
+		change.Path = strings.TrimPrefix(change.Path, "/")
+		strippedName := strings.TrimPrefix(change.Path, spec.StripPath)
+		change.Path = strings.TrimPrefix(change.Path, "/")
+		hdr.Name = filepath.Join(spec.RebasePath, strippedName)
 	}
```
jzt: Before we return we need to make sure directory names have a trailing `/`:

```go
if fi.IsDir() && !strings.HasSuffix(change.Path, "/") {
        name += "/"
}
```

lib/archive/util.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+// GenerateFilterSpec will populate the appropriate relative Rebase and Strip paths based on the supplied scenarios. Inclusion/Exclusion should be constructed separately. Please also note that any mount that exists before the copy target that is not primary and comes before the primary target will have a bogus filterspec, since it would not be written or read to.
+func GenerateFilterSpec(copyPath string, mountPoint string, primaryTarget bool, direction bool) FilterSpec {
+	var filter FilterSpec
+
+	// NOTE: this solidifies that this function is very heavily designed for docker cp behavior.
+	// which is why this should mainly only be used by the docker personality
+	// scrub trailing '/' before passing copyPath along and create a clean fs path
+	copyPath = strings.TrimSuffix(copyPath, "/")
```
jzt: could go all the way and use [`filepath.Clean()`](https://golang.org/pkg/path/filepath/#Clean)

lib/archive/unpack.go
```diff
@@ -45,7 +45,14 @@ const (
 func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root string) error {
 	op.Debugf("unpacking archive to root: %s, filter: %+v", root, filter)
 
-	tr := tar.NewReader(tarStream)
+	// Online datasource is sending a tar reader instead of an io reader.
+	// Type check here to see if we actually need to create a tar reader.
+	var tr *tar.Reader
+	if trCheck, ok := tarStream.(*tar.Reader); ok {
```
jzt: Is this necessary?

morris-jason: Yes, Online Datasource uses this.

lib/portlayer/storage/data_source.go
```diff
@@ -70,7 +77,7 @@ func (m *MountDataSource) Export(op trace.Operation, spec *archive.FilterSpec, d
 
 	// NOTE: this isn't actually diffing - it's just creating a tar. @jzt to explain why
```
jzt: It relies on docker's `ChangesDirs` func in the archive package. It takes two directories, new and old, and produces a list of the differences. If you supply `""` as the old directory, it will basically diff against nothing, so you end up with everything.

lib/portlayer/storage/data_source.go
```diff
@@ -79,8 +86,47 @@ func (m *MountDataSource) Export(op trace.Operation, spec *archive.FilterSpec, d
 	}, err
 }
 
+// Export reads data from the associated data source and returns it as a tar archive
```
jzt: This comment needs revision

lib/portlayer/storage/vsphere/toolbox_common.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/vm"
+)
+
+const (
+	DiskLabelQueryName  = "disk-label"
+	FilterSpecQueryName = "filter-spec"
+)
+
+// Parse Archive does something.
```
jzt: Comment needs a fixer-upper

lib/portlayer/storage/vsphere/toolbox_datasource.go
```diff
@@ -39,30 +39,81 @@ func (t *ToolboxDataSource) Source() interface{} {
...
+			return nil, err
+		}
+		op.Debugf("Downloaded from %s with size %d", target, contentLength)
+		readers = append(readers, tar)
+
+	}
+	return ioutil.NopCloser(io.MultiReader(readers...)), nil
+}
+
+// Export reads data from the associated data source and returns it as a tar archive
```
jzt: Copypasta needs update

jzt:
I'm ok with the stuff I haven't seen already, but it feels like I've reviewed most of this PR before it was consolidated. LGTM, overall. I do agree that the use of `# nosec` is just a tad out of control and needs to be addressed. 👍

lib/portlayer/storage/vsphere/toolbox_datasource.go
morris-jason: Done.

lib/portlayer/storage/vsphere/toolbox_common.go
morris-jason: Done.

lib/archive/util.go
morris-jason: I think that would break the copy case with '/.' at the end, where we want to copy the contents of a directory not the directory itself.

matthewavery:
all of the non necessary `#nosecs` should be removed at this point.

lib/tether/tether.go
```diff
@@ -251,20 +252,32 @@ func (t *tether) setNetworks() error {
 }
 
 func (t *tether) setMounts() error {
+	// provides a lookup from path to volume reference.
+	NameLookupMap := make(map[string]string, 0)
```
caglar10ur: [cosmetic] NameLookupMap -> l maybe? NameLookupMap creates an illusion that map itself is exported

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
...
+
+	// special behavior when using disk-labels and filterspec
+	diskLabel := u.Query().Get(vsphere.DiskLabelQueryName)
+	filterSpec := u.Query().Get(vsphere.FilterSpecQueryName)
+	if diskLabel != "" && filterSpec != "" {
+		op := trace.NewOperation(context.Background(), "ToolboxOnlineDataSink: %s", u.String())
+		op.Debugf("Reading from tar archive to path %s: %s", u.Path, u.String())
+		spec, err := archive.DecodeFilterSpec(op, &filterSpec)
+		if err != nil {
+			op.Debugf(err.Error())
```
caglar10ur: Errorf?

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
...
+			return err
+		}
+		diskPath, err := mountDiskLabel(diskLabel)
+		if err != nil {
+			op.Debugf(err.Error())
```
caglar10ur: ^ ditto

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
...
+			return err
+		}
+		defer unmount(op, diskPath)
+
+		// no need to join on u.Path here. u.Path == spec.Rebase, but
+		// Unpack will rebase tar headers for us. :thumbsup:
+		err = archive.Unpack(op, tr, spec, diskPath)
+		if err != nil {
+			op.Debugf(err.Error())
```
caglar10ur: ^ ditto

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
...
+	skiprecurse, _ := strconv.ParseBool(u.Query().Get(vsphere.SkipRecurseQueryName))
+	skipdata, _ := strconv.ParseBool(u.Query().Get(vsphere.SkipDataQueryName))
+
+	if diskLabel != "" && filterSpec != "" {
+		op := trace.NewOperation(context.Background(), "ToolboxOnlineDataSource: %s", u.String())
+		op.Debugf("Writing to archive from %s: %s", u.Path, u.String())
+
+		spec, err := archive.DecodeFilterSpec(op, &filterSpec)
+		if err != nil {
+			op.Debugf(err.Error())
```
caglar10ur: ^ ditto

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
...
+			return err
+		}
+
+		// get the container fs mount
+		diskPath, err := mountDiskLabel(diskLabel)
+		if err != nil {
+			op.Debugf(err.Error())
```
caglar10ur: ^ ditto

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
...
+				},
+			}
+
+			rc, err = archive.Tar(op, diskPath, changes, spec, !skipdata, false)
+		} else {
+			rc, err = archive.Diff(op, diskPath, "", spec, !skipdata, false)
+		}
+
+		if err != nil {
+			op.Debugf(err.Error())
```
caglar10ur: ^ ditto

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
...
+			tr := tar.NewReader(rc)
+			for {
+				hdr, err := tr.Next()
+				if err == io.EOF {
+					err = nil
+					op.Debugf("Finished writing to archive from %s: %s with error %#v", u.Path, u.String(), err)
+					break
+				}
+				if err != nil {
+					op.Debugf("error writing tar: %s", err.Error())
```
caglar10ur: ^ ditto

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
...
+					break
+				}
+				op.Debugf("Writing header: %#s", *hdr)
+				err = tw.WriteHeader(hdr)
+				if err != nil {
+					op.Debugf("error writing tar header: %s", err.Error())
```
caglar10ur: ^ ditto

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
...
+					break
+				}
+				_, err = io.Copy(tw, tr)
+				if err != nil {
+					op.Debugf("error writing tar contents: %s", err.Error())
```
caglar10ur: ^ ditto

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
...
+func unmount(op trace.Operation, unmountPath string) {
+	// don't unmount the root vmdk
+	if unmountPath == "/" {
+		return
+	}
+
+	// unmount the disk from the temporary directory
+	if err := Sys.Syscall.Unmount(unmountPath, syscall.MNT_DETACH); err != nil {
+		op.Debugf("failed to unmount %s: %s", unmountPath, err.Error())
+	}
```
caglar10ur: I think this is wrong. If umount fails only log something and then we will remove everything under that tree.

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
...
+
+func unmount(op trace.Operation, unmountPath string) {
+	// don't unmount the root vmdk
+	if unmountPath == "/" {
+		return
+	}
+
+	// unmount the disk from the temporary directory
+	if err := Sys.Syscall.Unmount(unmountPath, syscall.MNT_DETACH); err != nil {
+		op.Debugf("failed to unmount %s: %s", unmountPath, err.Error())
```
caglar10ur: ^ ditto

pkg/vsphere/disk/disk_manager.go
```diff
@@ -538,13 +552,19 @@ func (m *Manager) UnmountAndDetach(op trace.Operation, datastoreURI *object.Data
 func (m *Manager) InUse(op trace.Operation, config *VirtualDiskConfig, filter func(vm *mo.VirtualMachine) bool) ([]*vm.VirtualMachine, error) {
 	defer trace.End(trace.Begin(""))
 
-	if m.view == nil {
-		return nil, fmt.Errorf("ContainerView is nil")
+	mngr := view.NewManager(m.vm.Vim25())
```
caglar10ur: Can we create an issue and refer it from here to explain why we are not using m.view?

pkg/vsphere/disk/vmdk.go
```diff
@@ -62,9 +62,18 @@ func (v *Vmdk) Mount(op trace.Operation, uri *url.URL, persistent bool) (string,
 }
 
 func LockedVMDKFilter(vm *mo.VirtualMachine) bool {
+	if vm == nil {
```
caglar10ur: Why? I would argue calling this function with nil is a programming error and we should panic :)

lib/tether/toolbox.go
hickeng: This was a problem in the appliance also - I replaced the recursive delete with a directory delete. I see no circumstance under which we actually _want_ recursive deletion here.

pkg/vsphere/disk/disk_manager.go
hickeng: I opened https://github.com/vmware/govmomi/issues/781 for this but doug could not recreate. We may need to confirm again in place with VIC.

lib/tether/tether.go
matthewavery: yeah and it stuters the type :/ I have been thinking about what to call this one.

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.md
```diff
@@ -0,0 +1,78 @@
...
+* Step 15-16 should all succeed
+* Step 17 should fail
+* Step 18-30 should all succeed
+* Step 31-43 should all succeed
+* Step 44-51 should all succeed
+
+# Possible Problems:
+12-13 are skipped until we properly propagate errors back to user
+17 should fail with not enough disk space, but currently returns a broken pipe
+31-43 are skipped until we properly handle concurrent requests for offline cp
```
mhagen-vmware: add *'s

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.robot
```diff
@@ -0,0 +1,274 @@
...
+    Create File  ${CURDIR}/foo.txt   hello world
+    Create File  ${CURDIR}/content   fake file content for testing only
+    Create Directory  ${CURDIR}/bar
+    Create Directory  ${CURDIR}/mnt
+    Create Directory  ${CURDIR}/mnt/vol1
+    Create Directory  ${CURDIR}/mnt/vol2
+    Create File  ${CURDIR}/mnt/root.txt   rw layer file
+    Create File  ${CURDIR}/mnt/vol1/v1.txt   vol1 file
+    Create File  ${CURDIR}/mnt/vol2/v2.txt   vol2 file
+    ${rc}  ${output}=  Run And Return Rc And Output  dd if=/dev/zero of=${CURDIR}/largefile.txt count=4096 bs=4096
```
mhagen-vmware: /dev/random would be more interesting.

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.robot
```diff
@@ -0,0 +1,274 @@
...
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} volume create --name smallVol --opt Capacity=1
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${output}  Error
+
+Clean up test files and VIC appliance to test server
+    Remove File  ${CURDIR}/foo.txt
+    Remove File  ${CURDIR}/content
+    Remove File  ${CURDIR}/largefile.txt
+    Remove Directory  ${CURDIR}/bar  recursive=True
+    Remove Directory  ${CURDIR}/mnt  recursive=True
```
mhagen-vmware: All of these should be prefaced with Run Keyword and Continue on Failure

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.robot
```diff
@@ -0,0 +1,274 @@
...
+    ${pids}=  Create List
+    Log To Console  \nIssue 10 docker cp commands for large file
+    :FOR  ${idx}  IN RANGE  0  10
+    \   ${pid}=  Start Process  docker %{VCH-PARAMS} cp offline:/vol1/lg-${idx} ${CURDIR}  shell=True
+    \   Append To List  ${pids}  ${pid}
+    Log To Console  \nWait for them to finish and check their RC
+    :FOR  ${pid}  IN  @{pids}
+    \   Log To Console  \nWaiting for ${pid}
+    \   ${res}=  Wait For Process  ${pid}
+    \   Log To Console  ${res.stdout}
```
mhagen-vmware: don't log this to console, just Log this otherwise it is not easy to debug after the fact

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.robot
```diff
@@ -0,0 +1,274 @@
...
+    ${pids}=  Create List
+    Log To Console  \nIssue 10 docker cp commands for large file
+    :FOR  ${idx}  IN RANGE  0  10
+    \   ${pid}=  Start Process  docker %{VCH-PARAMS} cp ${CURDIR}/largefile.txt offline:/vol1/lg-${idx}  shell=True
+    \   Append To List  ${pids}  ${pid}
+    Log To Console  \nWait for them to finish and check their RC
+    :FOR  ${pid}  IN  @{pids}
+    \   Log To Console  \nWaiting for ${pid}
+    \   ${res}=  Wait For Process  ${pid}
+    \   Log To Console  ${res.stdout}
```
mhagen-vmware: Log

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.robot
```diff
@@ -0,0 +1,274 @@
...
+    ${pids}=  Create List
+    Log To Console  \nIssue 10 docker cp commands for small file
+    :FOR  ${idx}  IN RANGE  0  10
+    \   ${pid}=  Start Process  docker %{VCH-PARAMS} cp ${CURDIR}/foo.txt offline:/foo-${idx}  shell=True
+    \   Append To List  ${pids}  ${pid}
+    Log To Console  \nWait for them to finish and check their RC
+    :FOR  ${pid}  IN  @{pids}
+    \   Log To Console  \nWaiting for ${pid}
+    \   ${res}=  Wait For Process  ${pid}
+    \   Log To Console  ${res.stdout}
```
mhagen-vmware: Log

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
```diff
@@ -0,0 +1,304 @@
...
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} volume create --name v2
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${output}  Error
+
+Clean up test files and VIC appliance to test server
+    Remove File  ${CURDIR}/foo.txt
+    Remove File  ${CURDIR}/content
+    Remove File  ${CURDIR}/largefile.txt
+    Remove Directory  ${CURDIR}/bar  recursive=True
+    Remove Directory  ${CURDIR}/mnt  recursive=True
```
mhagen-vmware: these should be prefaced with Run Keyword and Continue on Failure

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
```diff
@@ -0,0 +1,304 @@
...
+
+Start container and run command
+    [Arguments]  ${containerName}  ${cmd}
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} start ${containerName}
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${output}  Error
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} exec ${containerName} ${cmd}
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${output}  Error
+    [Return]  ${output}
```
mhagen-vmware: Pull this out of both of these files and put it in Docker-Util.robot please

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
```diff
@@ -0,0 +1,304 @@
...
+
+Clean up test files and VIC appliance to test server
+    Remove File  ${CURDIR}/foo.txt
+    Remove File  ${CURDIR}/content
+    Remove File  ${CURDIR}/largefile.txt
+    Remove Directory  ${CURDIR}/bar  recursive=True
+    Remove Directory  ${CURDIR}/mnt  recursive=True
+    Cleanup VIC Appliance On Test Server
+
+Start container and run command
```
mhagen-vmware: Start container and exec command

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
```diff
@@ -0,0 +1,304 @@
...
+    ${pids}=  Create List
+    Log To Console  \nIssue 10 docker cp commands for small file
+    :FOR  ${idx}  IN RANGE  0  10
+    \   ${pid}=  Start Process  docker %{VCH-PARAMS} cp ${CURDIR}/foo.txt concurrent:/foo-${idx}  shell=True
+    \   Append To List  ${pids}  ${pid}
+    Log To Console  \nWait for them to finish and check their RC
+    :FOR  ${pid}  IN  @{pids}
+    \   Log To Console  \nWaiting for ${pid}
+    \   ${res}=  Wait For Process  ${pid}
+    \   Log To Console  ${res.stdout}
```
mhagen-vmware: Log

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
```diff
@@ -0,0 +1,304 @@
...
+    ${pids}=  Create List
+    Log To Console  \nIssue 10 docker cp commands for large file
+    :FOR  ${idx}  IN RANGE  0  10
+    \   ${pid}=  Start Process  docker %{VCH-PARAMS} cp ${CURDIR}/largefile.txt concurrent:/vol1/lg-${idx}  shell=True
+    \   Append To List  ${pids}  ${pid}
+    Log To Console  \nWait for them to finish and check their RC
+    :FOR  ${pid}  IN  @{pids}
+    \   Log To Console  \nWaiting for ${pid}
+    \   ${res}=  Wait For Process  ${pid}
+    \   Log To Console  ${res.stdout}
```
mhagen-vmware: Log

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
```diff
@@ -0,0 +1,304 @@
...
+    ${pids}=  Create List
+    Log To Console  \nIssue 10 docker cp commands for large file
+    :FOR  ${idx}  IN RANGE  0  10
+    \   ${pid}=  Start Process  docker %{VCH-PARAMS} cp concurrent:/vol1/lg-${idx} ${CURDIR}  shell=True
+    \   Append To List  ${pids}  ${pid}
+    Log To Console  \nWait for them to finish and check their RC
+    :FOR  ${pid}  IN  @{pids}
+    \   Log To Console  \nWaiting for ${pid}
+    \   ${res}=  Wait For Process  ${pid}
+    \   Log To Console  ${res.stdout}
```
mhagen-vmware: Log

lib/tether/toolbox.go
morris-jason: Done.

lib/tether/toolbox.go
morris-jason: Done.

lib/tether/toolbox.go
morris-jason: Done.

lib/tether/toolbox.go
morris-jason: Done.

lib/tether/toolbox.go
morris-jason: Done.

lib/tether/toolbox.go
morris-jason: Done.

lib/tether/toolbox.go
morris-jason: Done.

lib/tether/toolbox.go
morris-jason: Done.

lib/tether/toolbox.go
morris-jason: Done.

lib/tether/toolbox.go
morris-jason: Done.

lib/tether/toolbox.go
morris-jason: Done.

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.md
morris-jason: Done.

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.robot
morris-jason: Done.

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.robot
morris-jason: Done.

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.robot
morris-jason: Done.

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.robot
morris-jason: Done.

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.robot
morris-jason: Done.

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
morris-jason: Done.

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
morris-jason: Done.

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
morris-jason: Done.

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
morris-jason: Done.

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
morris-jason: Done.

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
morris-jason: Done.

lib/tether/tether.go
matthewavery: @caglar10ur it looks like I should be in this area again soon to address another potential bug. I may make this change at that time in order to avoid additional small commits onto the feature branch.

lib/apiservers/engine/backends/archive.go
```diff
@@ -164,38 +189,89 @@ func (c *Container) importToContainer(vc *viccontainer.VicContainer, target stri
...
+		store = constants.VolumeStoreName
+	}
+
+	stat, err = c.containerProxy.StatPath(op, store, deviceID, fs)
+	if err != nil {
+		op.Errorf("error getting statpath: %s", err.Error())
+		switch err := err.(type) {
+		case *storage.StatPathNotFound:
+			return nil, ResourceNotFoundError(vc.Name, "file or directory")
+		default:
```
matthewavery: We return a `NewStatPathUnprocessableEntity` and we might need to explore exactly what kind of error message that we return in the event that it gets propagated back to the docker cli here. Ideally that error will only occur if we(the personality) malform the filterspec or the decode of the filterspec goes awry somehow. But in the interest of returning useful error messages back to the docker cli this might be an issue.

matthewavery: https://github.com/vmware/vic/pull/5865/files#diff-d17427b3b302b85a6c75b5a5a264bdc3R625

lib/apiservers/engine/backends/archive.go
```diff
@@ -231,18 +310,8 @@ func NewArchiveStreamWriterMap(mounts []types.MountPoint, dest string) *ArchiveS
...
-		aw.filterSpec.StripPath = aw.mountPoint.Destination
-		aw.filterSpec.RebasePath = strings.TrimPrefix(dest, aw.mountPoint.Destination)
-
-		// if containerDestPath != "/" && strings.HasPrefix(aw.mountPoint.Destination, containerDestPath) {
-		// 	aw.filterSpec.StripPath = strings.TrimPrefix(aw.mountPoint.Destination, containerDestPath)
-		// }
-
-		aw.filterSpec.Exclusions = make(map[string]struct{})
-		aw.filterSpec.Inclusions = make(map[string]struct{})
+		isPrimary := !strings.Contains(aw.mountPoint.Destination, dest) || aw.mountPoint.Destination == dest
```
matthewavery: I wanted to drop a brief comment here, This can be refactored further in the future to make things even more simple for spec generation by internalizing the `isPrimary` logic. I do not it should be done in this PR as it would be a refactor at this point.

lib/apiservers/engine/backends/archive.go
jialin-li: Not sure what we can respond to the user since user has no knowledge of filterspec. We have the same problem for ImportArchive and ExportArchive. Do you suggest a retry of generating and encoding the filter spec or something like `error processing target`?

lib/apiservers/engine/backends/archive.go
morris-jason: Yeah, agreed, this should not have to be done before the call to GenerateFilterSpec.

matthewavery: yeah it can just be internalized since it is not clear to the caller what `PrimaryTarget` means.

lib/apiservers/engine/backends/archive.go
matthewavery: solved out of band in slack. 
from @jialin-li in slack:

```
I’ve just updated the error msg to `failed to process given path` and added it for import and export as well
```

lib/portlayer/storage/vsphere/toolbox_common.go
```diff
@@ -0,0 +1,77 @@
...
+	target = path.Join("/archive:/", target)
+
+	// if diskLabel is longer than 16 characters, then the function was passed a containerID
+	// use containerfs as the diskLabel
+	if len(disklabel) > 16 {
+		disklabel = "containerfs"
+	}
+
+	// note that the query parameters a SkipX for recurse and data so values are inverted
+	target += fmt.Sprintf("?%s=%s&%s=%s&%s=%t&%s=%t", DiskLabelQueryName, disklabel, FilterSpecQueryName, *encodedSpec, SkipRecurseQueryName, !recurse, SkipDataQueryName, !data)
```
dougm: This ought to be:
```go
	target += (url.Values{
		DiskLabelQueryName: []string{diskLabel},
                FilterSpecQueryName: []string{encodedSpec},
                ...
	}).Encode()
```

This way any special chars within the values will be properly url encoded.

lib/archive/archiver.go
```diff
@@ -164,42 +164,35 @@ func EncodeFilterSpec(op trace.Operation, spec *FilterSpec) (*string, error) {
...
+				inclusionLength = temp
 			}
 		}
 	}
 
 	for path := range spec.Exclusions {
 		if strings.HasPrefix(filePath, path) {
-			if len(path) > len(exclusion) {
+			temp := len(path)
+			if temp > exclusionLength {
```
matthewavery: refactor(likely not needed in this PR): this can be condensed to one line.

lib/tether/tether.go
hickeng: @matthewavery Agree with caglar - shouldn't be capitalized.
You can also drop an extra map lookup and skip the `targetRef` indirection unless you must have it for logging:
```go
	mounts := make([]string, len(t.config.Mounts))
	nameLookupMap := make(map[string]*url.URL, len(t.config.Mounts))
	for k, v := range t.config.Mounts {
		mounts = append(mounts, v.Path)
		nameLookupMap[v.Path] = &v.Source
	}
```

lib/tether/toolbox.go
```diff
@@ -46,6 +59,17 @@ type Toolbox struct {
 	stop chan struct{}
 }
 
+var (
+	defaultArchiveHandler hgfs.ArchiveHandler
```
morris-jason: Ah, that's what I was looking for. Thanks.

lib/tether/toolbox.go
```diff
@@ -46,6 +59,17 @@ type Toolbox struct {
...
+)
+
+func init() {
```
morris-jason: Thanks.

lib/portlayer/storage/vsphere/toolbox_common.go
morris-jason: Awesome, thanks!

lib/portlayer/storage/vsphere/toolbox_datasource.go
```diff
@@ -39,30 +40,86 @@ func (t *ToolboxDataSource) Source() interface{} {
...
 	if err != nil {
+		op.Errorf("Cannot get toolbox client: %s", err.Error())
 		return nil, err
 	}
 
-	// authenticate client and parse container host/port
-	auth := types.NamePasswordAuthentication{
-		Username: t.ID,
+	var readers []io.Reader
+	for inclusion := range spec.Inclusions {
```
hickeng: @morris-jason I'm not sure this should be a loop - looks like the tether side will handle a spec with multiple inclusions and we're not stripping the spec before passing it through.
Will this result in multiple copies of the same file?

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
...
+				}
+			}
+		}()
+		wg.Wait()
+		return err
+	}
+	return defaultArchiveHandler.Write(u, tw)
+}
+
+func mountDiskLabel(label string) (string, error) {
```
hickeng: Should really be part of the Operations interface in ops_linux.go as this will not function for non-linux containers. Might involve up-leveling the function to CopyDontCrossMounts for example.
Doesn't block PR.

lib/archive/unpack.go
jzt: I meant is there a situation wherein we wouldn't be passed a `tar.Reader`? Or, alternatively, should we stick to the convention of expecting an `io.ReadCloser`? Just curious, not a show-stopper.

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
...
+		}
+
+		if err != nil {
+			op.Errorf(err.Error())
+			return err
+		}
+
+		wg := sync.WaitGroup{}
+		wg.Add(1)
+		go func() {
```
morris-jason: Leftover from when we were writing to a gzip writer. Removing.

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
...
+	if unmountPath == "/" {
+		return
+	}
+
+	// unmount the disk from the temporary directory
+	if err := Sys.Syscall.Unmount(unmountPath, syscall.MNT_DETACH); err != nil {
+		op.Debugf("failed to unmount %s: %s", unmountPath, err.Error())
+	}
+	// finally, remove the temporary directory
+	os.RemoveAll(unmountPath)
```
morris-jason: Already done.

lib/portlayer/storage/vsphere/toolbox_datasource.go
morris-jason: We have to have a single inclusion path, as client.Download/Upload uses a single path for FileTransfer{To/From} Guest.

lib/tether/toolbox.go
morris-jason: Will do this in S14.

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
 
 	return session.Cmd.Process.Kill()
 }
+
+// ----------
+// Online DataSource and DataSink Override Handlers
+// ----------
+
+func toolboxOverrideArchiveRead(u *url.URL, tr *tar.Reader) error {
```
morris-jason: Will do this in S14.

lib/portlayer/storage/vsphere/export.go
```diff
@@ -157,7 +175,10 @@ func (i *ImageStore) Export(op trace.Operation, id, ancestor string, spec *archi
 		return nil, fmt.Errorf("mismatched datasource types: %T, %T", ls, rs)
 	}
 
-	tar, err := archive.Diff(op, fl.Name(), fr.Name(), spec, data)
+	// if we want data, exclude the xattrs, otherwise assume diff
+	xattrs := !data
```
hickeng: inline it in the call.

lib/archive/diff.go
matthewavery: I believe this still needs to be addressed, out of band with @jzt it is determined that this is a must.

lib/portlayer/storage/storage.go
```diff
@@ -150,6 +159,9 @@ type DataSource interface {
 	//     nfs volume:  		 XDR-client
 	//     via guesttools:  	 toolbox client
 	Source() interface{}
+
+	// Given a filter spec
+	Stat(op trace.Operation, spec *archive.FilterSpec) (*FileStat, error)
```
hickeng: Comment does not conform to Godoc standard. Should be:
`Stat returns a minimal set of file stats for files matched by the spec`

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -695,14 +697,54 @@ func (c *ContainerProxy) GetContainerChanges(ctx context.Context, vc *viccontain
...
+	spec, err := archive.EncodeFilterSpec(op, &filterSpec)
+	if err != nil {
+		op.Errorf(err.Error())
+		return nil, InternalServerError(err.Error())
+	}
+	statPathParams = statPathParams.WithFilterSpec(spec)
+
+	statPathOk, err := c.client.Storage.StatPath(statPathParams)
+	if err != nil {
+		op.Errorf(err.Error())
```
hickeng: Should this be checking for 404/NotFound? Given that's an expected result for a stat operation does it require different handling?

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -566,6 +566,9 @@ func (h *StorageHandlersImpl) ImportArchive(params storage.ImportArchiveParams)
 	if err != nil {
 		// hickeng: see if we can return usefully typed errors here
```
hickeng: This note can be removed now.

lib/archive/diff.go
matthewavery: Yeah I think it should be, if we do not get an error then we likely should not log all of the time imo. The above situation is pretty much binary.

lib/archive/util.go
```diff
@@ -0,0 +1,144 @@
...
+			exclusion := Clean(strings.TrimPrefix(mount, currentMount), true) + "/"
+			filter.Exclusions[exclusion] = struct{}{}
+		}
+	}
+	return nil
+}
+
+// Clean run filepath.Clean on the target and will remove leading
+// and trailing slashes from the target path corresponding to supplied booleans
+func Clean(path string, leading bool) string {
```
morris-jason: It is being used in archive.go.

lib/portlayer/storage/data_source.go
matthewavery: altering comment to clarify in brief detail the use of diff.

lib/portlayer/storage/vsphere/toolbox_datasource.go
```diff
@@ -39,30 +40,86 @@ func (t *ToolboxDataSource) Source() interface{} {
...
+	if err != nil {
+		op.Errorf("Cannot get toolbox client: %s", err.Error())
+		return nil, err
 	}
 
-	_ = filemgr
-	_ = auth
+	// should only find a single path to stat, but make sure here.
+	var statPath string
+	for inclusion := range spec.Inclusions {
```
morris-jason: done

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
morris-jason: done.

lib/apiservers/engine/backends/container_proxy.go
jialin-li: The error type check is done in its caller `ContainerStatPath` in archive.go where it has access to the container name used for `ResourceNotFound` message.

morris-jason: No, this should be fine in both situations. NotFound/InternalServer errors are treated the same way by the cli.
 - In CopyFrom, the operation is halted in both situations inside ContainerArchivePath
 - In CopyTo, the cli is the caller and will disregard any errors, treating the destination as a directory.

lib/tether/tether.go
matthewavery: Changed this to `pathIndex` I kept the targetRef for logging. I think it is useful to us to have that information in the logs. It is better to have this mapping for the lookup rather than attempt to work backwards to find the name again in this instance.

lib/archive/util.go
```diff
@@ -0,0 +1,144 @@
...
+
+const (
+	// CopyTo is used to indicate that the desired filter spec is for a CopyTo direction
+	CopyTo = true
+
+	// CopyFrom is used to indicate that the desired filter spec is for the CopyFrom direction
+	CopyFrom = false
+)
+
+// GenerateFilterSpec will populate the appropriate relative Rebase and Strip paths based on the supplied scenarios. Inclusion/Exclusion should be constructed separately. Please also note that any mount that exists before the copy target that is not primary and comes before the primary target will have a bogus filterspec, since it would not be written or read to.
```
matthewavery: yes very long, just realized that as well when I changed my resolution.

lib/archive/util.go
```diff
@@ -0,0 +1,144 @@
...
+func GenerateFilterSpec(copyPath string, mountPoint string, primaryTarget bool, direction bool) FilterSpec {
+	var filter FilterSpec
+
+	// NOTE: this solidifies that this function is very heavily designed for docker cp behavior.
+	// which is why this should mainly only be used by the docker personality
+	// scrub trailing '/' before passing copyPath along and create a Clean fs path
+	copyPath = strings.TrimSuffix(copyPath, "/")
+
+	// Note I know they are just booleans, if that changes then this statement will not need to.
+	if direction == CopyTo {
```
matthewavery: the only other case here is a container to container copy I believe. In docker they actually have a switch case reporting an error in that situation. For us, we likely would not even recieve that call since the docker CLI will detect the situation and return from it before making an actual rest call. Since this function is primarily used for `docker cp` it is unlikely to change from only situations. This statement looks really odd because `CopyTo` is a boolean so doing an == on `direction` is really redundant and was only done at the time to make it clear what we are implying for that logical branch. I iwll look at what it would take to shift those to an `iota/enumeration` to make it more clear.

lib/archive/unpack.go
```diff
@@ -74,50 +80,50 @@ func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root st
...
-			continue
 		case tar.TypeSymlink:
-
 			err := os.Symlink(header.Linkname, absPath)
 			if err != nil {
 				op.Errorf("Failed to create symlink %s->%s: %s", absPath, header.Linkname, err)
 				return err
 			}
-
 		case tar.TypeReg:
```
matthewavery: @hickeng do you think we should also include tar.TypeRegA in this case of the switch? just in case?

lib/archive/unpack.go
```diff
@@ -74,50 +80,50 @@ func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root st
 			return err
 		}
 
-		op.Debugf("processing tar header: %s", header.Name)
-
+		op.Debugf("processing tar header: %#v", *header)
+		op.Debugf("using FilterSpec : (%#v)", *filter)
```
matthewavery: It likely is not necessary. I included it here since anyone could be calling this function and the filterspec is integral to what is exected from this function. I have moved it out of the loop :)

Makefile
```diff
@@ -238,7 +238,7 @@ govet:
 
 gas: $(GAS)
 	@echo checking security problems
-	@$(GAS) -skip=*_responses.go -quiet lib/... cmd/... pkg/... 2> /dev/null
+	@for i in cmd lib pkg; do pushd $$i > /dev/null; $(GAS) -skip=*_responses.go ./... > ../$$i.gas 2> /dev/null || exit 1; popd > /dev/null; done
```
anchal-agrawal: This change should be undone.

jialin-li: Done, reverted in my pr.

lib/apiservers/engine/backends/archive.go
```diff
@@ -644,3 +687,90 @@ func (rm *ArchiveStreamReaderMap) Close() {
 
 	rm.prefixTrie.Visit(closeStream)
 }
+
+// try to fake the statpath for path that targets the mountpoint or along the mountpoint
```
anchal-agrawal: `// tryFakeStatPath tries to fake the statpath for path that targets the mountpoint or along the mountpoint`

lib/apiservers/engine/backends/archive.go
```diff
@@ -644,3 +687,90 @@ func (rm *ArchiveStreamReaderMap) Close() {
...
+
+	for _, mount := range mounts {
+		if strings.HasPrefix(mount.Destination, target) {
+			isMountPathTarget = true
+		}
+	}
+
+	// check to see if the path is a mount point, if so, return fake path
+	if isMountPathTarget {
+		return &types.ContainerPathStat{
```
anchal-agrawal: Formatting - this should look like

```go
return &types.ContainerPathStat{
			Name:       filepath.Base(target),
			Size:       int64(4096),
			Mode:       os.ModeDir,
			Mtime:      time.Now(),
			LinkTarget: "",
        }, true
```

lib/apiservers/engine/backends/archive.go
```diff
@@ -644,3 +687,90 @@ func (rm *ArchiveStreamReaderMap) Close() {
...
+			Size:       int64(4096),
+			Mode:       os.ModeDir,
+			Mtime:      time.Now(),
+			LinkTarget: ""}, true
+	}
+
+	return nil, false
+}
+
+// use mountpoints to strip the target to a relative path
```
anchal-agrawal: Function comments should be formatted to start with the function name.

lib/apiservers/engine/backends/archive.go
```diff
@@ -644,3 +687,90 @@ func (rm *ArchiveStreamReaderMap) Close() {
...
+		return err
+	}
+
+	for _, node := range foundNodes {
+		vicarchive.AddMountInclusionsExclusions(node.mountPoint.Destination, &node.filterSpec, mounts, containerSourcePath)
+	}
+	return nil
+}
+
+// This function will return the node pointers from the prefix tree as well as all mounts involved in the operation
```
anchal-agrawal: Re: function comment.

lib/apiservers/engine/backends/archive.go
```diff
@@ -644,3 +687,90 @@ func (rm *ArchiveStreamReaderMap) Close() {
...
+func (rm *ArchiveStreamReaderMap) buildMountsAndNodes(path string, node *ArchiveReader) ([]string, []*ArchiveReader, error) {
+
+	// We can modify this to make proper exclusions in the future. For now,
```
anchal-agrawal: It would be nice to prefix this comment block with a `NOTE(author):` so it's easier to address in the future.

lib/apiservers/engine/backends/archive.go
jialin-li: done

lib/portlayer/storage/vsphere/export.go
matthewavery: comment was preserved for clarity

lib/apiservers/engine/backends/archive_test.go
```diff
@@ -15,12 +15,15 @@
 package backends
 
 import (
+	"context"
 	"testing"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/stretchr/testify/assert"
 
 	"github.com/docker/docker/api/types"
```
anchal-agrawal: goimports/gofmt: there should be three import groups for packages: standard, third-party, and own.

There are four here.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -681,9 +685,7 @@ func (c *ContainerProxy) StreamContainerStats(ctx context.Context, config *conve
 	return nil
 }
 
-// GetContainerChanges returns container changes from portlayer.
-// Set data to true will return file data, otherwise, only return file headers with change type.
-func (c *ContainerProxy) GetContainerChanges(ctx context.Context, vc *viccontainer.VicContainer, data bool) (io.ReadCloser, error) {
+func (c *ContainerProxy) GetContainerChanges(op trace.Operation, vc *viccontainer.VicContainer, data bool) (io.ReadCloser, error) {
```
anchal-agrawal: Exported functions should have function comments.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -695,14 +697,54 @@ func (c *ContainerProxy) GetContainerChanges(ctx context.Context, vc *viccontain
...
-	r, err := archiveProxy.ArchiveExportReader(ctx, constants.ContainerStoreName, host, vc.ContainerID, parent, data, spec)
+	r, err := archiveProxy.ArchiveExportReader(op, constants.ContainerStoreName, host, vc.ContainerID, parent, data, spec)
 	if err != nil {
 		return nil, InternalServerError(err.Error())
 	}
 
 	return r, nil
 }
 
+// StatPath requests the portlayer to stat the filesystem resource at the
```
anchal-agrawal: 👍

lib/portlayer/storage/storage.go
matthewavery: changed to this : `// Stat stats the filesystem target indicated by the last entry in the given Filterspecs inclusion map`

lib/apiservers/engine/backends/container_test.go
```diff
@@ -42,7 +42,10 @@ import (
 	plclient "github.com/vmware/vic/lib/apiservers/portlayer/client"
 	plscopes "github.com/vmware/vic/lib/apiservers/portlayer/client/scopes"
 	plmodels "github.com/vmware/vic/lib/apiservers/portlayer/models"
+	"github.com/vmware/vic/lib/archive"
 	"github.com/vmware/vic/lib/metadata"
+
+	"github.com/vmware/vic/pkg/trace"
```
anchal-agrawal: Re: pkg import group comment: https://github.com/vmware/vic/pull/5865/files#r130988908

lib/apiservers/engine/proxy/errors.go
```diff
@@ -30,3 +30,8 @@ func InternalServerError(msg string) error {
 func ResourceLockedError(msg string) error {
 	return derr.NewErrorWithStatusCode(fmt.Errorf("Resource locked: %s", msg), http.StatusLocked)
 }
+
+// ResourceLockedError returns a 423 http status
```
anchal-agrawal: This seems to return a 404 instead of a 423 as stated here.

lib/archive/diff.go
```diff
@@ -67,12 +73,31 @@ func Tar(op trace.Operation, dir string, changes []docker.Change, spec *FilterSp
 		hdr *tar.Header
 	)
 
+	// I don't like this as we've not assurance that it's appropriately buffered, but cannot think of a go-idomatic way to
```
anchal-agrawal: Would be nice to prefix this with `NOTE(author):`

lib/archive/stripper.go
```diff
@@ -0,0 +1,174 @@
...
+
+		// io.Copy never returns EOF so treat nil as EOF but keeping
+		// EOF for clarity
+		if err == io.EOF || err == nil {
+			continue
+		}
+
+		// err was non-nil/EOF and we read data - legitimate error scenario
+		if n > 0 {
+
```
anchal-agrawal: Unneeded newline.

lib/portlayer/storage/data_source.go
```diff
@@ -79,8 +86,47 @@ func (m *MountDataSource) Export(op trace.Operation, spec *archive.FilterSpec, d
...
+func (m *MountDataSource) Stat(op trace.Operation, spec *archive.FilterSpec) (*FileStat, error) {
+	// retrieve relative path
+
```
anchal-agrawal: Unneeded newline? Not sure what code block this comment is tied to.

lib/portlayer/storage/vsphere/container.go
```diff
@@ -275,7 +310,10 @@ func (c *ContainerStore) Export(op trace.Operation, id, ancestor string, spec *a
 		return nil, errors.New("mismatched datasource types")
 	}
 
-	tar, err := archive.Diff(op, fl.Name(), fr.Name(), spec, data)
+	// if we want data, exclude the xattrs, otherwise assume diff
+	xattrs := !data
```
jzt: Yeah, I did that so people coming along later reading this wouldn't be puzzled as to why.

lib/portlayer/storage/vsphere/export.go
jzt: Yeah I'm cool with inlining, all we really need is the comment.

jzt: Even though I bet the compiler does it automatically (hopefully) :)

lib/tether/toolbox.go
```diff
@@ -112,7 +136,11 @@ func (t *Toolbox) Reload(config *ExecutorConfig) error {
 	t.authIDs[config.ID] = struct{}{}
 	// we also allow any device IDs that are attached
 	for _, mspec := range config.Mounts {
-		t.authIDs[mspec.Source.Host] = struct{}{}
+		// mounstpect.source.path is the disk label for vmdks
+		// XXX: this is not the case for other volumes, eg nfs vols.
```
anchal-agrawal: Please replace the `XXX` with a `NOTE` or `TODO`.

lib/tether/toolbox.go
```diff
@@ -237,3 +270,161 @@ func (t *Toolbox) halt() error {
 
 	return session.Cmd.Process.Kill()
 }
+
+// ----------
```
anchal-agrawal: Function comments don't need to have extra formatting - please remove lines 274, 76 and 77.

pkg/trace/operation.go
```diff
@@ -175,6 +175,22 @@ func WithDeadline(parent *Operation, expiration time.Time, format string, args .
 	return op, cancelFunc
 }
 
+// WithCancel
```
anchal-agrawal: This comment should be expanded upon or removed entirely.

pkg/trace/operation.go
```diff
@@ -175,6 +175,22 @@ func WithDeadline(parent *Operation, expiration time.Time, format string, args .
...
+func WithCancel(parent *Operation, format string, args ...interface{}) (Operation, context.CancelFunc) {
+	ctx, cancelFunc := context.WithCancel(parent.Context)
+	op := parent.newChild(ctx, fmt.Sprintf(format, args...))
+
+	return op, cancelFunc
+}
+
+// WithValue
```
anchal-agrawal: Ditto: function comment.

lib/archive/diff.go
matthewavery: update: @jzt and I have decided to keep this change out of this merge as it has potential implications for tar correctness and as of this moment all of the diff integration and unit tests are passing in the state they are in.

lib/portlayer/storage/vsphere/toolbox_datasource.go
hickeng: good point - I'm sure I knew that ;)

lib/archive/util.go
hickeng: did a search and didn't find it #gurufail

lib/archive/unpack.go
hickeng: yes

matthewavery:
It appears that all approvals are present. One last full ci will

PR #5872, opened 2017-07-28
Fix maxattach release path

hickeng:


pkg/vsphere/disk/disk_manager.go
```diff
@@ -469,7 +469,7 @@ func (m *Manager) detach(op trace.Operation, disk *types.VirtualDisk) error {
 		return t, er
 	})
 
-	if err != nil {
+	if err == nil {
```
chengwang86: Can we add some log here to see whether the maxAttached is true or not before `if`? We don't even know whether this would be an error or not

hickeng: @chengwang86 It's a channel - there's no way to tell how many items are in it. It's the main curse of using one as a counting semaphore, but they are very simple and well tested.

chengwang86:
lgtm

hickeng:
merged in https://github.com/vmware/vic/pull/5861

PR #5874, opened 2017-07-29
Fix for subvolume cp and some house cleaning

matthewavery:
Still needs a little cleanup before merging.


lib/apiservers/engine/backends/archive.go
```diff
@@ -195,20 +195,26 @@ func (c *Container) ContainerStatPath(name string, path string) (stat *types.Con
...
-			stat = &types.ContainerPathStat{
-				Name:       filepath.Base(fs.RebasePath),
-				Size:       int64(4096),
-				Mode:       os.ModeDir,
-				Mtime:      time.Now(),
-				LinkTarget: ""}
-			op.Debugf("faking container stat path %#v", stat)
-			return stat, nil
+
+	isMountPathTarget := false
```
matthewavery: we should only do this if `stat` of the path returns not exists.

PR #5880, opened 2017-07-31
[specific ci=1-43-Docker-CP-Offline] Prevent personality from returning early on copy to case, fix statpath 

### LINKED ISSUE #5717: Ensure Useful errors are propagated to the client on copy failure

jialin-li:
Fixes #5717

### DIFF
--- a/lib/apiservers/engine/backends/archive.go
+++ b/lib/apiservers/engine/backends/archive.go
@@ -22,6 +22,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	"golang.org/x/net/context"
@@ -148,7 +149,7 @@ func (c *Container) importToContainer(op trace.Operation, vc *viccontainer.VicCo
 	writerMap := NewArchiveStreamWriterMap(op, mounts, target)
 	defer func() {
 		// This should shutdown all the stream connections to the portlayer.
-		e1 := writerMap.Close()
+		e1 := writerMap.Close(op)
 		if err == nil {
 			err = e1
 			op.Debugf("import to container: assigned err as %v", err)
@@ -268,6 +269,7 @@ type ArchiveWriter struct {
 type ArchiveStreamWriterMap struct {
 	prefixTrie *patricia.Trie
 	op         trace.Operation
+	wg         *sync.WaitGroup
 	errchan    chan error
 }
 
@@ -281,7 +283,8 @@ func NewArchiveStreamWriterMap(op trace.Operation, mounts []types.MountPoint, de
 	writerMap := &ArchiveStreamWriterMap{}
 	writerMap.prefixTrie = patricia.NewTrie()
 	writerMap.op = op
-	writerMap.errchan = make(chan error, len(mounts))
+	writerMap.errchan = make(chan error, 1)
+	writerMap.wg = &sync.WaitGroup{}
 
 	for _, m := range mounts {
 		aw := ArchiveWriter{
@@ -447,7 +450,7 @@ func (wm *ArchiveStreamWriterMap) WriterForAsset(proxy proxy.VicArchiveProxy, ci
 			deviceID = aw.mountPoint.Name
 			store = constants.VolumeStoreName
 		}
-		rawWriter, err := proxy.ArchiveImportWriter(wm.op, store, deviceID, aw.filterSpec, wm.errchan)
+		rawWriter, err := proxy.ArchiveImportWriter(wm.op, store, deviceID, aw.filterSpec, wm.wg, wm.errchan)
 		if err != nil {
 			err = fmt.Errorf("Unable to initialize import stream writer for mount prefix %s", aw.mountPoint.Destination)
 			wm.op.Errorf(err.Error())
@@ -461,31 +464,36 @@ func (wm *ArchiveStreamWriterMap) WriterForAsset(proxy proxy.VicArchiveProxy, ci
 }
 
 // Close visits all the archive writer in the trie and closes the actual io.WritCloser
-func (wm *ArchiveStreamWriterMap) Close() error {
+func (wm *ArchiveStreamWriterMap) Close(op trace.Operation) error {
 	defer trace.End(trace.Begin(""))
 
-	numWriter := 0
 	closeStream := func(prefix patricia.Prefix, item patricia.Item) error {
 		if aw, ok := item.(*ArchiveWriter); ok && aw.writer != nil {
 			aw.tarWriter.Close()
 			aw.writer.Close()
 			aw.tarWriter = nil
 			aw.writer = nil
-			numWriter++
 		}
 		return nil
 	}
 
 	wm.prefixTrie.Visit(closeStream)
 
+	// wait for all pl calls to return and close the channel
+	go func() {
+		wm.wg.Wait()
+		close(wm.errchan)
+	}()
+
 	var err error
 	// wait for all the streams to finish
-	for i := 0; i < numWriter; i++ {
-		result := <-wm.errchan
+	for result := range wm.errchan {
 		if result != nil {
 			err = result
+			op.Errorf("Error received from portlayer for import streams: %s", result.Error())
 		}
 	}
+
 	return err
 }
 
--- a/lib/apiservers/engine/backends/archive.go
+++ b/lib/apiservers/engine/backends/archive.go
@@ -155,7 +155,6 @@ func (c *Container) importToContainer(op trace.Operation, vc *viccontainer.VicCo
 		}
 	}()
 
-
 	for {
 		header, err := tarReader.Next()
 		if err == io.EOF {
@@ -462,16 +461,16 @@ func (wm *ArchiveStreamWriterMap) WriterForAsset(proxy proxy.VicArchiveProxy, ci
 }
 
 // Close visits all the archive writer in the trie and closes the actual io.WritCloser
-func (wm *ArchiveStreamWriterMap) Close() (err error) {
+func (wm *ArchiveStreamWriterMap) Close() error {
 	defer trace.End(trace.Begin(""))
 
 	numWriter := 0
 	closeStream := func(prefix patricia.Prefix, item patricia.Item) error {
 		if aw, ok := item.(*ArchiveWriter); ok && aw.writer != nil {
-			aw.writer.Close()
 			aw.tarWriter.Close()
-			aw.writer = nil
+			aw.writer.Close()
 			aw.tarWriter = nil
+			aw.writer = nil
 			numWriter++
 		}
 		return nil
@@ -479,14 +478,15 @@ func (wm *ArchiveStreamWriterMap) Close() (err error) {
 
 	wm.prefixTrie.Visit(closeStream)
 
+	var err error
 	// wait for all the streams to finish
 	for i := 0; i < numWriter; i++ {
-		result := <- wm.errchan
+		result := <-wm.errchan
 		if result != nil {
 			err = result
 		}
 	}
-	return
+	return err
 }
 
 // FindArchiveReaders finds all archive readers that are within the container source path.  For example,
@@ -768,4 +768,3 @@ func (rm *ArchiveStreamReaderMap) buildMountsAndNodes(path string, node *Archive
 
 	return mounts, nodes, nil
 }
-
--- a/lib/apiservers/engine/backends/archive.go
+++ b/lib/apiservers/engine/backends/archive.go
@@ -55,7 +55,7 @@ func (c *Container) ContainerArchivePath(name string, path string) (io.ReadClose
 
 	stat, err := c.ContainerStatPath(name, path)
 	if err != nil {
-		return nil, nil, InternalServerError(err.Error())
+		return nil, nil, err
 	}
 
 	reader, err := c.exportFromContainer(op, vc, path)
@@ -135,7 +135,7 @@ func (c *Container) ContainerExtractToDir(name, path string, noOverwriteDirNonDi
 	return err
 }
 
-func (c *Container) importToContainer(op trace.Operation, vc *viccontainer.VicContainer, target string, content io.Reader) error {
+func (c *Container) importToContainer(op trace.Operation, vc *viccontainer.VicContainer, target string, content io.Reader) (err error) {
 	rawReader, err := archive.DecompressStream(content)
 	if err != nil {
 		op.Errorf("Input tar stream to ContainerExtractToDir not recognized: %s", err.Error())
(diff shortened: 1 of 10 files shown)


lib/apiservers/engine/backends/archive.go
```diff
@@ -461,20 +457,31 @@ func (wm *ArchiveStreamWriterMap) WriterForAsset(proxy proxy.VicArchiveProxy, ci
...
 // Close visits all the archive writer in the trie and closes the actual io.WritCloser
-func (wm *ArchiveStreamWriterMap) Close() {
+func (wm *ArchiveStreamWriterMap) Close() (err error) {
 	defer trace.End(trace.Begin(""))
 
+	numWriter := 0
 	closeStream := func(prefix patricia.Prefix, item patricia.Item) error {
 		if aw, ok := item.(*ArchiveWriter); ok && aw.writer != nil {
 			aw.writer.Close()
 			aw.tarWriter.Close()
```
morris-jason: tarWriter.Close() needs to be before writer.Close() in case it needs to flush.

lib/apiservers/engine/proxy/archive.go
```diff
@@ -96,18 +95,18 @@ func (a *ArchiveProxy) ArchiveExportReader(ctx context.Context, store, ancestorS
 			switch err := err.(type) {
```
morris-jason: Can you change L94 to op.Errorf?

lib/apiservers/engine/backends/archive.go
```diff
@@ -190,9 +198,18 @@ func (c *Container) ContainerStatPath(name string, path string) (stat *types.Con
...
 	}
 
+	path = filepath.Clean(path)
 	mounts := mountsFromContainer(vc)
 	mounts = append(mounts, types.MountPoint{Destination: "/"})
 
+	// handle the special case of targeting a volume mount point before it exists.
+	// this will be important for non started container cp, will also be important
+	// to certain behaviors for diff on a non started container.
+	if stat, succeed := tryFakeStatPath(mounts, path); succeed {
```
hickeng: I'm not clear why we're doing this before we try in the real filesystem. Shouldn't the logic be:
a. if target found via full stat path, return actual stat results
b. if target not found and target is known to be a mount point, return fake mount dir stat.

lib/apiservers/engine/proxy/archive.go
```diff
@@ -157,23 +156,29 @@ func (a *ArchiveProxy) ArchiveImportWriter(ctx context.Context, store, deviceID
...
 		}
 
+		var plErr error
 		_, err = a.client.Storage.ImportArchive(params)
 		if err != nil {
 			switch err := err.(type) {
 			case *storage.ImportArchiveInternalServerError:
-				plErr := InternalServerError(fmt.Sprintf("Server error from archive writer for device %s", deviceID))
-				log.Errorf(plErr.Error())
+				plErr = InternalServerError(fmt.Sprintf("Server error from archive writer for device %s", deviceID))
```
hickeng: This may be a translation of the current error, but I don't think it's useful to the user. Can we change to something such as:
`error writing files to %s`

lib/apiservers/engine/proxy/archive.go
```diff
@@ -157,23 +156,29 @@ func (a *ArchiveProxy) ArchiveImportWriter(ctx context.Context, store, deviceID
...
+				op.Errorf(plErr.Error())
 				pipeReader.CloseWithError(plErr)
 			case *storage.ImportArchiveLocked:
-				plErr := ResourceLockedError(fmt.Sprintf("Resource locked for device %s", deviceID))
-				log.Errorf(plErr.Error())
+				plErr = ResourceLockedError(fmt.Sprintf("Resource locked for device %s", deviceID))
```
hickeng: for this and others - general style is to have lower case start for strings in errors as they are frequently concatenated.

lib/apiservers/engine/backends/archive.go
```diff
@@ -461,20 +457,31 @@ func (wm *ArchiveStreamWriterMap) WriterForAsset(proxy proxy.VicArchiveProxy, ci
...
 	wm.prefixTrie.Visit(closeStream)
+
+	// wait for all the streams to finish
+	for i := 0; i < numWriter; i++ {
+		result := <- wm.errchan
+		if result != nil {
+			err = result
+		}
+	}
+	return
```
matthewavery: Noting here that we just do a `return` but we do not log the error or return it. Is this intentional?

matthewavery: ah never mind named return :)

matthewavery: maybe remove the named return and just do a `return err` here for readability.

lib/apiservers/engine/proxy/archive.go
jialin-li: done!

lib/apiservers/engine/proxy/archive.go
```diff
@@ -182,6 +187,8 @@ func (a *ArchiveProxy) ArchiveImportWriter(ctx context.Context, store, deviceID
 		} else {
 			pipeReader.Close()
 		}
+		op.Debugf("Stream for device %s has returned from PL. Err received is %v ", deviceID, plErr)
+		errchan <- plErr
```
hickeng: given this _must_ happen for the `Close` call to exit it's good practice to put it in a defer for safety - ensures someone doesn't accidentally add in an exit path that skips it.

lib/apiservers/engine/proxy/archive.go
```diff
@@ -123,7 +122,7 @@ func (a *ArchiveProxy) ArchiveExportReader(ctx context.Context, store, ancestorS
 
 // ArchiveImportWriter initializes a write stream for a path.  This is usually called
 // for getting a writer during docker cp TO container.
-func (a *ArchiveProxy) ArchiveImportWriter(ctx context.Context, store, deviceID string, filterSpec vicarchive.FilterSpec) (io.WriteCloser, error) {
+func (a *ArchiveProxy) ArchiveImportWriter(op trace.Operation, store, deviceID string, filterSpec vicarchive.FilterSpec, errchan chan error) (io.WriteCloser, error) {
 	defer trace.End(trace.Begin(deviceID))
 
 	if store == "" || deviceID == "" {
```
matthewavery: If this ends up being the case we would return before populating the `errchan` which could cause us to wait forever later on.

jialin-li: If the stream fails to initialize, then a pipe writer wouldn't be returned, so numWriter wouldn't include this and wouldn't wait for it either.

lib/apiservers/engine/backends/archive.go
jialin-li: The benefit of faking statpath is that we can avoid making calls to portlayer and mount vmdk when the target is a mountpoint or along the mountpoint.

matthewavery: I think an additional detail for this case is as such: This is the personality so it will be a docker specific ignore. The other cases are that this path is a file or such type situation. Since this is an actual path of a disk mount on the container and the personality knows it, the personality can reasonably assume that the container mount path is valid since a `docker create` must successfully execute before this(meaning the mount point is valid). That being said we should note that currently VIC has a bug where we do not confirm if the mount point is valid on a `create` operation. Leading us to create broken containers that can never start due to the mount points actually being invalid targets. Since we have a useful `StatPath` from the portlayer we can now fix that during a `docker create` and this code will make more sense. We might still make a case for attempting the actual stat results as it could help us understand if we attempted to stat the wrong thing or some other such scenario. That is my $0.02.

lib/apiservers/engine/proxy/archive.go
matthewavery: +1 on this.

matthewavery:
LGTM - would like to come to a conclusion on the discussion above with @hickeng as well though

lib/apiservers/engine/backends/archive.go
```diff
@@ -636,12 +643,34 @@ func (rm *ArchiveStreamReaderMap) Close() {
 	rm.prefixTrie.Visit(closeStream)
 }
 
+// try to fake the statpath for path that targets the mountpoint or along the mountpoint
+func tryFakeStatPath(mounts []types.MountPoint, target string) (*types.ContainerPathStat, bool) {
+	isMountPathTarget := false
+
+	for _, mount := range mounts {
+		if strings.HasPrefix(mount.Destination, target) {
```
sflxn: Are you sure this is correct?  What if mount has '/mnt/A' and target is '/mnt'?  If I understand this logic right, I think you're looking for an exact match.

jialin-li: We want to fake the statpath for `/mnt` because any directory along the mount path will be created. The problem is that before the container is started, directory in the mount path are not created yet, so stating it on the container filesystem will fail, even though it's a valid target. It's hard to see with `/mnt` since that already exists. But if the mount path is `/vol/A` then 'docker cp test.txt cid:/vol` is a valid cp command, but stating it on local filesystem will fail. So we need to include directories along the mount path as well.

lib/apiservers/engine/backends/archive.go
```diff
@@ -447,7 +442,7 @@ func (wm *ArchiveStreamWriterMap) WriterForAsset(proxy proxy.VicArchiveProxy, ci
 			deviceID = aw.mountPoint.Name
 			store = constants.VolumeStoreName
 		}
-		rawWriter, err := proxy.ArchiveImportWriter(wm.op, store, deviceID, aw.filterSpec)
+		rawWriter, err := proxy.ArchiveImportWriter(wm.op, store, deviceID, aw.filterSpec, wm.errchan)
```
sflxn: I've mainly seen channels that return data returned from functions, but passing one in works too.  I'm not sure if there is any idiomatic rule about this.  Just a comment.

lib/apiservers/engine/proxy/archive.go
```diff
@@ -32,7 +32,7 @@ import (
 
 type VicArchiveProxy interface {
 	ArchiveExportReader(ctx context.Context, store, ancestorStore, deviceID, ancestor string, data bool, filterSpec vicarchive.FilterSpec) (io.ReadCloser, error)
-	ArchiveImportWriter(ctx context.Context, store, deviceID string, filterSpec vicarchive.FilterSpec) (io.WriteCloser, error)
+	ArchiveImportWriter(ctx context.Context, store, deviceID string, filterSpec vicarchive.FilterSpec, errchan chan error) (io.WriteCloser, error)
```
sflxn: Already commented above.  I usually see something like this,

ArchiveImportWriter(ctx context.Context, store, deviceID string, filterSpec vicarchive.FilterSpec) (io.WriteCloser, <-
chan error, error)

The <- indicates it's for reading only.  Not sure if returning it rather than passing in via arg is idiomatic.  Not a big deal.

lib/apiservers/engine/backends/archive.go
```diff
@@ -146,7 +146,14 @@ func (c *Container) importToContainer(op trace.Operation, vc *viccontainer.VicCo
 	mounts := mountsFromContainer(vc)
 	mounts = append(mounts, types.MountPoint{Destination: "/"})
 	writerMap := NewArchiveStreamWriterMap(op, mounts, target)
-	defer writerMap.Close() // This should shutdown all the stream connections to the portlayer.
+	defer func() {
+		// This should shutdown all the stream connections to the portlayer.
+		e1 := writerMap.Close()
+		if err == nil {
```
sflxn: Do you need to get the error here?  Working through it in my head, only one io.Copy() occurs at any one time.  If there is an error, the error is returned from io.Copy and then the defer function gets called.  There shouldn't be any other errors from any other writers.

jialin-li: This is to get errors from portlayer calls. I've updated Close() to wait for all portlayer calls to return and return errors from portlayer if any.

lib/apiservers/engine/backends/archive.go
jialin-li: The error gets populated from PL response as well. So if copy fails at unpack stage, then this channel will get populated with error.

lib/apiservers/engine/backends/archive.go
sflxn: I see.

lib/apiservers/engine/proxy/archive.go
jialin-li: The reason I passed this in is that I didn't want to wait on a variable number of channels to unblock. I could use a wait group but I will still have to allocate buffer and give index to each stream to store their returned error. So I am using one buffered error channel with enough error entries for maximum number of streams.

lib/apiservers/engine/backends/archive.go
sflxn: The error is also returned during any writes on the writer.  I see why you want to do this, but we serialize all writes to the writers so if there is an error, it gets returned in io.Copy() and we catch it there, log it, and return.  Have you seen a case where that isn't so?

lib/apiservers/engine/backends/archive.go
sflxn: We should be getting error from portlayer via the error returned from io.Copy().  In the proxy, the call to pipeReader.CloseWithErrors() propogates the error back to the backend.  Is this not happening?

lib/apiservers/engine/backends/archive.go
jialin-li: We do catch the `io.Copy` error in `unpack()`. All the errors are returned but we weren't waiting for them before. The error channel is used to block personality from returning until all writes are finished.

jialin-li: The CloseWithErrors() works if one PL call returns before all writes are finished. But the error can happen after all pipewriter finishes, and in that case, we won't catch them with closeWithErrors because there won't be anymore writes.

sflxn:
lgtm

lib/apiservers/engine/backends/archive.go
```diff
@@ -461,20 +461,32 @@ func (wm *ArchiveStreamWriterMap) WriterForAsset(proxy proxy.VicArchiveProxy, ci
...
+			numWriter++
 		}
 		return nil
 	}
 
 	wm.prefixTrie.Visit(closeStream)
+
+	var err error
+	// wait for all the streams to finish
+	for i := 0; i < numWriter; i++ {
```
vburenin: I am not really sure I really understand what's going on here not seeing all pieces of the code, however, I would consider doing this:
1. Why would you even need to push a nil into errchan? Just push errors only.
2. To wait all writers to complete their job, I would consider using sync.WaitGroup, it will be more clear than depend on the channel content. Also, I think errchan has to be closed as soon all jobs are done, thus you can just iterate over channel using ```for e := range errchan``` loop until you get the last value if that one is needed. However, would definitely consider logging all errors if there are many.

jialin-li: I am waiting for an arbitrary number of go routines to finish, so it's more for a state checking. Also I won't know how many of them returns errors.  I've looked into sync.waitGroup and using an additional array to store errors but it looks messier than just using the chan error. 

I can't use for range because the buffer allocated for channel can be larger than the number of writers I am waiting for. I have to allocate the buffer size before I know how many active writers are there. But yes, I will add a close after waiting for the writer.

PR #5881, opened 2017-07-31
Include Mountpoint Headers [specific ci=1-44-Docker-CP-Online]

morris-jason:
Alters filterspec to add a mountpoint directory header to the diff archive.


matthewavery:
LGTM - discussed and reviewed this change out of band with @morris-jason . Looks good for finding a parent disk and requesting the additional read. The archive filterspec generation should generate the correct filterspec for this situation without change. leaving an include for only the directory desired and an exclude for `""` while stripping the mountpoint and rebasing the base of the target on as the header.

PR #5884, opened 2017-08-01
Add ordered manner to tether mount handling

matthewavery:
`docker create -v vol1:/some/path -v vol2:/some/path/further/down busybox` should have it's volumes mounted in the right order at boot time. This Means that the tether will no longer non-deterministically appply mounts during container start time.


matthewavery:
I have run the offline tests several times locally and they are all passing with this PR.

matthewavery:
not sure why it is not picking up my changes...

emlin:
> not sure why it is not picking up my changes...

you just need to restart the CI test, it will pick up latest

PR #5894, opened 2017-08-01
Obtain cached portlayer container state without a lock

### LINKED ISSUE #5736: Concurrent run / ps test deadlocks VIC engine
**User Statement:**

As a container user, I want to be able to control VIC engine from multiple clients concurrently. This may be because I want to look at logs while something is coming up; one client may be a CI pipeline and I need to interact with an existing container... there are numerous cases where multiple clients are important.

**Details:**

This issue was found by a customer.

It's relatively easy to deadlock VIC engine with multiple clients. At the point of deadlock, the control plane will blocks for every client indefinitely. I can reproduce with both VIC engine and ESX and with a variety of commands. 

Rebooting the endpoint VM corrects the problem.

The easiest way to reproduce is to run the following commands in two shells:

```
while true; do date; docker run --rm ubuntu sleep 10; done
while true; do date; docker ps; sleep 1; done
```

It's so simple to reproduce, there's not much point attaching logs. I can reproduce with a recent master build. The Port Layer seems to get stuck here:

```
[BEGIN] [github.com/vmware/vic/lib/apiservers/portlayer/restapi/handlers.(*ContainersHandlersImpl).GetContainerListHandler:284]
```

**Acceptance Criteria:**

While it's impossible to test every combination of command concurrently, we should be testing common scenarios such as getting logs and stats concurrent with container lifecycle operations.

I am making this very high priority to fix

cgtexmex:
A lock was used when obtaining the current state of a
container which would cause a deadlock in certain parallel
portLayer operations.  Now the cache will obtain without
a container lock.

Fixes #5736

### DIFF
--- a/lib/portlayer/exec/container_cache.go
+++ b/lib/portlayer/exec/container_cache.go
@@ -65,8 +65,10 @@ func (conCache *containerCache) Containers(state *State) []*Container {
 		if !isContainerID(id) {
 			continue
 		}
-
-		if state == nil || *state == con.CurrentState() {
+		// filter by container state
+		// DO NOT use container.CurrentState as that can
+		// cause cache deadlocks
+		if state == nil || *state == con.State() {
 			containers = append(containers, con)
 		}
 	}


lib/portlayer/exec/container_cache.go
```diff
@@ -65,8 +65,10 @@ func (conCache *containerCache) Containers(state *State) []*Container {
 		if !isContainerID(id) {
 			continue
 		}
-
-		if state == nil || *state == con.CurrentState() {
+		// filter by container state
+		// DO NOT use container.CurrentState as that can
+		// cause cache deadlocks
+		if state == nil || *state == con.State() {
```
caglar10ur: Should we name those functions appropriately? Eg; StateWithoutLock()

PR #5911, opened 2017-08-03
[specific ci=1-*Docker-CP-*] Add retry to all toolbox cp and stat operations

matthewavery:
This change should make us more resilient to the `InvalidStat` fault. Allowing more stability for cp operations against an online container.

### DIFF
--- a/lib/portlayer/storage/vsphere/toolbox_datasink.go
+++ b/lib/portlayer/storage/vsphere/toolbox_datasink.go
@@ -16,7 +16,6 @@ package vsphere
 
 import (
 	"io"
-	"time"
 
 	"github.com/vmware/govmomi/vim25/soap"
 	"github.com/vmware/govmomi/vim25/types"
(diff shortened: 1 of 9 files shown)


pkg/vsphere/tasks/waiter.go
```diff
@@ -108,10 +108,10 @@ func isRetryError(op trace.Operation, err error) bool {
 		switch f := soap.ToSoapFault(err).VimFault().(type) {
 		case types.TaskInProgress:
 			return true
-		case *types.NetworkDisruptedAndConfigRolledBack:
+		case types.NetworkDisruptedAndConfigRolledBack:
```
matthewavery: This is a slight fix for an issue that has not been filed against checks here not working previously. We may not have seen it yet. But we will fix it here and move on.

lib/portlayer/storage/vsphere/toolbox_datasink.go
```diff
@@ -54,10 +56,22 @@ func (t *ToolboxDataSink) Import(op trace.Operation, spec *archive.FilterSpec, d
 
 	// upload the gzip archive.
 	p := soap.DefaultUpload
-	err = client.Upload(op, data, target, p, &types.GuestPosixFileAttributes{}, true)
+	// These numbers are somewhat arbitrary best guesses
+	retryConf := retry.NewBackoffConfig()
```
morris-jason: Looks like this is in every file. Can you please make it a constant in toolbox_common?

PR #5912, opened 2017-08-03
Disable nonpersistent disk use [ci specific=1-43-*Offline]

hickeng:
For now we need to disable non-persistent disk use as Stat and Import
clash against each other. Stat requests non-persistent, Import requires
persistent. We concurrent copyTo operations this results in a transient
failure due to the disk being attached non-persistent and held by multiple
Stat calls, but an Import call needed it in persistent mode.
Until/unless we have an ability to wait for the disk to become available
for a different mode we cannot be this granular.
This also disables r/o mount for the read only case as we do not yet
support per mount ref counts - it's all tied to the base disk so mounting
the same disk in two locations, one read/only one not isn't feasible.


matthewavery:
lgtm

Issue #5848, opened 2017-07-26
Modify imagec binary to write tar files for layers to validate docker commit's data


sflxn:
As a user of VIC, I should be confident that images I create via docker commit works and all data within the newly created layers are valid.

Acceptance Criteria:

- [ ] An imagec binary that accepts a command to write a layer's tar from a VCH's image store.

---

This depends on #5847

emlin:
bump to 3, more work thank I thought

PR #5914, opened 2017-08-03
[specific ci=1-17-Docker-Network-Connect]add imagec save command

emlin:
- imagec save command is to download image archive from portlayer. it will download all layers of the specified image
- expose portlayer API at `<clientIP>:2380` if VCH is deployed with debug > 2
- fix image store register to make the importer/exporter works after VCH is restarted

  Here is one sample command to run imagec save:
  `imagec -insecure-skip-verify  -reference busybox -destination /tmp/save -standalone -debug -operation save -host 10.162.45.156:2380 -image-store 421791f0-21f7-73ea-c86a-fab0da0dd0ce`

User need to specify image store ID in the command line as the sample. ImageC does not assume running inside of VCH. The ID could be retrieved from vSphere datastore file path through govc

fixes: #5848


cmd/docker/main.go
```diff
@@ -134,6 +134,16 @@ func handleFlags() bool {
 	}
 	extraconfig.Decode(src, &vchConfig)
 
+	if vchConfig.Diagnostics.DebugLevel > 2 {
+		// use debug port layer mode
+		*cli.portLayerPort = constants.DebugPortLayerPort
+		clientIP, err := util.ClientIP()
+		if err != nil {
+			log.Fatalf("Unable to look up %s for portlayer API: %s", constants.ClientHostName, err)
+			log.Fatal(err)
```
sflxn: Do you need this second Fatal log if you have the one above?

emlin: good catch, will delete it

cmd/imagec/main.go
```diff
@@ -195,11 +200,79 @@ func main() {
...
+		ic := imagec.NewImageC(options, streamformatter.NewJSONStreamFormatter())
+		if err := saveImage(ap, ic, options.Reference); err != nil {
+			log.Fatalf("Saving image %s failed due to %s\n", options.Reference.String(), err)
+		}
 	default:
 		log.Errorf("The operation '%s' is not valid\n", imageCOptions.operation)
 	}
 }
 
+func saveImage(ap proxy.VicArchiveProxy, ic *imagec.ImageC, ref reference.Named) error {
```
sflxn: It doesn't appear you really need to ref here.  You mainly use it for logging and  you can get that from ic.

cmd/imagec/main.go
```diff
@@ -195,11 +200,79 @@ func main() {
...
+			defer wg.Done()
+			var err error
+			log.Debugf("parent id: %s, layer id: %s", pid, cid)
+
+			defer func() {
+				errChan <- err
+			}()
+
+			fileDir := path.Join(dest, cid)
+			err = os.MkdirAll(fileDir, 0755) /* #nosec */
```
sflxn: I think you should just create one folder and pass the folder name into this go routine.  I think one folder for each layer would get annoying at some point.

cmd/port-layer-server/main.go
```diff
@@ -104,6 +106,17 @@ func main() {
 		}
 	}
 
+	if vchConfig.Diagnostics.DebugLevel > 2 {
+		// expose portlayer service on client interface
+		server.Port = constants.DebugPortLayerPort
+		clientIP, err := util.ClientIP()
+		if err != nil {
+			log.Fatalf("Unable to look up %s to serve portlayer API: %s", constants.ClientHostName, err)
+			log.Fatal(err)
```
sflxn: Do we need to duplicate error?  We already output err in the previous fatal.

sflxn:
I think you should also modify the operation flag to add the save option in the help text.

cmd/imagec/main.go
emlin: imagec pull will create folder for each layer file anyway.
I think it might be okay as long as we have same folder name for both imagec pull and image save.

cmd/port-layer-server/main.go
emlin: sorry for the left over

emlin:
> I think you should also modify the operation flag to add the save option in the help text.

changed to `Pull image/push image/listlayers/save image`

sflxn:
lgtm

PR #5928, opened 2017-08-04
[full ci] Fix tether iptables race

### LINKED ISSUE #5921: Container networks still configuring after the container comes up
User statement: As a user of VIC, I expect container networks to be properly configured by the time a container starts in any condition. 

Description: Due to the nature of tether-init, starting a process (i.e. iptables) does not block until completion. This means that there is a race condition where too many calls to the iptables binary are required, and the session is started before the firewall is properly set up. This will fix stability failures related to 6-07, 1-24, and 1-36.

Logs as seen in https://ci.vcna.io/vmware/vic/12932:
```
DEBUG [BEGIN] [main.(*attachServerSSH).Reload:110] attach reload
DEBUG [ END ] [main.(*attachServerSSH).Reload:110] [413.187µs] attach reload
INFO  Launching process for session 165111f4ad5166cb28b57edb353c4281cf02fd63c597c48a79e310fcac9976a
DEBUG [BEGIN] [main.(*attachServerSSH).run:387] main attach server loop
DEBUG [BEGIN] [main.backchannel:274] establish tether backchannel
>> DEBUG Reaped process 285, return code: 0 <<
```

Acceptance criteria: All iptables calls should complete before the return of the `SetupFirewalls` function.

gigawhitlocks:
Sometimes there's a race condition in the tether due to not blocking after making calls to `iptables`. This change makes those calls to iptables block so that the next call in the sequence (a ping) can properly connect to the outside world. Fixes #5921

### DIFF
--- a/cmd/tether/main_test.go
+++ b/cmd/tether/main_test.go
@@ -138,6 +138,10 @@ func (m *mockery) Stop() error {
 func (m *mockery) Register(name string, config tether.Extension) {
 }
 
+func (m *mockery) LaunchUtility(fn func() (*os.Process, error)) (<-chan int, error) {
+	return nil, nil
+}
+
 // Test reloading via signal helper
 func TestReload(t *testing.T) {
 	m := &mockery{make(chan bool)}
--- a/cmd/tether/ops_linux.go
+++ b/cmd/tether/ops_linux.go
@@ -131,14 +131,14 @@ func (t *operations) Setup(sink tether.Config) error {
 
 // SetupFirewall sets up firewall rules on the external scope only.  Any
 // portmaps are honored as are port exposes.
-func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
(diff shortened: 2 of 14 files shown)


gigawhitlocks:
ok yes i think this is failing literally every integration test so something is wrong

hickeng:
Added initialization for utilityPid map but still failed.
Added explicit panic check to tether.Start and was able to log panics.

Seems that what we're doing with passing the tether via context.WithValue is resulting in a zero value tether when we retrieve it. I'm unclear why.

Opened PR against @gigawhitlocks branch with working changes.

hickeng:
@gigawhitlocks I ended up opening https://github.com/vmware/vic/pull/5931 as I want to run full CI against the changes. I've still no idea why your changes didn't work once the map was initialized. I gave up on that - it's probably really easy to fix if you have a clue what's going on, but....

PR #5931, opened 2017-08-07
Support waiting on iptables processes

hickeng:
* Restores the ability to run tether unit tests as non-root
* Adds methods to launch and wait on processes avoiding racing with the
  reaper
* Adds unit test glue to support exercising the firewall logic
* Re-enables squashed ping tests.

Does NOT currently provide firewall unit tests in the absence of mocking
for the actual invocation.

This PR is heavily based off @gigawhitlocks PR, but needing to push it
through CI urgently so raised again here.

### DIFF
--- a/cmd/tether/net_linux_test.go
+++ b/cmd/tether/net_linux_test.go
@@ -1,0 +1,126 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// +build linux
+
+package main
+
+import (
+	"fmt"
+	"strconv"
+	"testing"
+
+	"github.com/vishvananda/netlink"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/vmware/vic/lib/config/executor"
+	"github.com/vmware/vic/lib/tether"
+	"github.com/vmware/vic/pkg/vsphere/extraconfig"
+)
+
+// Utility method to add an interface to Mocked
+// This assigns the interface name and returns the "slot" as a string
+func AddInterface(name string, mocker *Mocker) string {
+	mocker.maxSlot++
+
+	mocker.Interfaces[name] = &Interface{
+		LinkAttrs: netlink.LinkAttrs{
+			Name:  name,
+			Index: mocker.maxSlot,
+		},
+		Up: true,
+	}
+
+	return strconv.Itoa(mocker.maxSlot)
+}
+
+func TestOutboundRuleAndCmd(t *testing.T) {
+	mocker := testSetup(t)
+	defer testTeardown(t, mocker)
+
+	bridge := AddInterface("eth1", mocker)
+
+	ip, _ := netlink.ParseIPNet("172.16.0.2/24")
+	gwIP, _ := netlink.ParseIPNet("172.16.0.1/24")
+
+	cfg := executor.ExecutorConfig{
+		ExecutorConfigCommon: executor.ExecutorConfigCommon{
+			ID:   "outboundrule",
+			Name: "tether_test_executor",
+		},
+		Diagnostics: executor.Diagnostics{
+			DebugLevel: 3,
+		},
+		Networks: map[string]*executor.NetworkEndpoint{
+			"bridge": {
+				Common: executor.Common{
+					ID: bridge,
+					// interface rename
+					Name: "bridge",
+				},
+				Network: executor.ContainerNetwork{
+					Common: executor.Common{
+						Name: "bridge",
+					},
+					Default: true,
+					Gateway: *gwIP,
+				},
+				Static: true,
+				IP:     ip,
+			},
+		},
+
+		Sessions: map[string]*executor.SessionConfig{
+			"outboundrule": {
+				Common: executor.Common{
+					ID:   "outboundrule",
+					Name: "tether_test_session",
+				},
+				Tty:    false,
+				Active: true,
+
+				Cmd: executor.Cmd{
+					// test relative path
+					Path: "./date",
+					Args: []string{"./date", "--reference=/"},
+					Env:  []string{"PATH="},
+					Dir:  "/bin",
+				},
+			},
+		},
+	}
+
+	_, src, _ := StartTether(t, &cfg, mocker)
+
+	fmt.Println("Waiting for tether start")
+	<-mocker.Started
+
+	// wait for tether to exit
+	fmt.Println("Waiting for tether exit")
+	<-mocker.Cleaned
+
+	result := tether.ExecutorConfig{}
+	extraconfig.Decode(src, &result)
+
+	// confirm outbound rules configured
+	// this should modify state depending on prior rule state
+
+	// confirm command ran - necessary to detect early exit due to net config error
+	// TODO: this should be modifed to fail if the last rule to be configured hasn't completed with expected output
+	// when this is run. Pending mocked iptables interface
+	assert.Equal(t, "true", result.Sessions["outboundrule"].Started, "Expected command to have been started successfully")
+	assert.Equal(t, 0, result.Sessions["outboundrule"].ExitStatus, "Expected command to have exited cleanly")
+
+}
--- a/cmd/tether/net_linux_test.go
+++ b/cmd/tether/net_linux_test.go
@@ -1,4 +1,4 @@
-// Copyright 2016 VMware, Inc. All Rights Reserved.
+// Copyright 2017 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -21,18 +21,17 @@ import (
 	"strconv"
 	"testing"
 
-	"github.com/vishvananda/netlink"
-
 	"github.com/stretchr/testify/assert"
+	"github.com/vishvananda/netlink"
 
 	"github.com/vmware/vic/lib/config/executor"
 	"github.com/vmware/vic/lib/tether"
 	"github.com/vmware/vic/pkg/vsphere/extraconfig"
 )
 
-// Utility method to add an interface to Mocked
+// addInterface utility method to add an interface to Mocked
 // This assigns the interface name and returns the "slot" as a string
-func AddInterface(name string, mocker *Mocker) string {
+func addInterface(name string, mocker *Mocker) string {
 	mocker.maxSlot++
 
 	mocker.Interfaces[name] = &Interface{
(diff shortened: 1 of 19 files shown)


cmd/tether/net_linux_test.go
```diff
@@ -0,0 +1,126 @@
...
+// +build linux
+
+package main
+
+import (
+	"fmt"
+	"strconv"
+	"testing"
+
+	"github.com/vishvananda/netlink"
```
vburenin: I don't want to sound too "annoying" here, but could you please remove this space line?

cmd/tether/net_linux_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/vmware/vic/lib/config/executor"
+	"github.com/vmware/vic/lib/tether"
+	"github.com/vmware/vic/pkg/vsphere/extraconfig"
+)
+
+// Utility method to add an interface to Mocked
```
vburenin: Just in case, all method/function descriptions should start with its name:
```AddInterface is a utility method...```

File: cmd/tether/net_linux_test.go [hunk 3]
```diff
@@ -0,0 +1,126 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
```
chengwang86: 2017

cmd/tether/ops_linux_test.go
```diff
@@ -0,0 +1,158 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
```
willsherwood: 2017

lib/tether/netfilter/netfilter_linux.go
```diff
@@ -158,7 +160,7 @@ func (r *Rule) args() ([]string, error) {
 	return args, nil
 }
 
-func iptables(ctx context.Context, args []string) error {
+func iptables(ctx context.Context, args []string) (tether.UtilityFn, error) {
 	args = append(args, "-w")
```
willsherwood: Now that the iptables calls are being executed sequentially, the flag to wait for the xtables lock is no longer necessary.

cmd/tether/ops_linux.go
```diff
@@ -129,19 +129,43 @@ func (t *operations) Setup(sink tether.Config) error {
...
 // portmaps are honored as are port exposes.
-func (t *operations) SetupFirewall(config *tether.ExecutorConfig) error {
-	// XXX It looks like we'd want to collect the errors here, but we
-	// can't.  Since this is running inside init (tether) and tether
-	// reaps all children, the os.exec package won't be able to collect
-	// the error code in time before the reaper does.  The exec package
-	// calls wait and attempts to collect its child, but the reaper will
-	// have raptured the pid before that.  So, best effort, just keep going.
-	if err := netfilter.Flush(context.Background(), ""); err != nil {
+func (t *operations) SetupFirewall(ctx context.Context, config *tether.ExecutorConfig) error {
```
vburenin: using "t" for operations is very confusing, it always triggers a thinking it is a testing.T
Is there any way to mass rename it? Gogland can actually do that in one shot.

lib/tether/net_linux_test.go
```diff
@@ -155,3 +157,89 @@ func TestSetIpAddress(t *testing.T) {
...
+	<-mocker.Cleaned
+
+	result := ExecutorConfig{}
+	extraconfig.Decode(src, &result)
+
+	// confirm command ran - necessary to detect early exit due to net config error
+	assert.Equal(t, "true", result.Sessions["outboundrule"].Started, "Expected command to have been started successfully")
+	assert.Equal(t, 0, result.Sessions["outboundrule"].ExitStatus, "Expected command to have exited cleanly")
+
+	// confirm outbound rules configured
```
chengwang86: Is this supposed to be removed or implemented?

cmd/tether/ops_linux.go
```diff
@@ -129,19 +129,43 @@ func (t *operations) Setup(sink tether.Config) error {
...
+	return setupFirewall(ctx, &t.BaseOperations, config)
+}
+
+// setupFirewall is broken out from SetupFirewall so that it can be referenced from the test code
```
vburenin: I think setupFirewall can always be referenced from the test code, it is just a matter of how you initialize BaseOperation structure. having it out of structure makes things a little confusing.

lib/tether/ops_linux.go
```diff
@@ -99,10 +105,10 @@ type Netlink interface {
 	LinkBySlot(slot int32) (netlink.Link, error)
 }
 
-func init() {
-	Sys.Hosts = etcconf.NewHosts(hostsPathBindSrc)
-	Sys.ResolvConf = etcconf.NewResolvConf(resolvConfPathBindSrc)
-}
+// func init() {
```
willsherwood: Commented out function left in. Was this intended?

cmd/tether/ops_linux_test.go
vburenin: I wonder if implementing all this logic makes sense. We could use go-mock to generate code that could pre-precord expected behavior, so you can easily test error cases, etc.

chengwang86:
Overall lgtm

lib/tether/interfaces.go
```diff
@@ -44,6 +48,12 @@ type Operations interface {
 	// Returns a function to invoke after the session state has been persisted
 	HandleSessionExit(config *ExecutorConfig, session *SessionConfig) func()
 	ProcessEnv(env []string) []string
+	// LaunchUtility starts a process and provides a way to block on completion and retrieve
+	// it's exit code. This is needed to co-exist with a childreaper.
+	LaunchUtility(UtilityFn) (<-chan int, error)
```
vburenin: I think it would make sense to start using utilities which generate interfaces from the existing data structure: https://github.com/vburenin/ifacemaker - this is a good one :-)

lib/tether/ops_linux.go
vburenin: Do you want to keep it commented out?

lib/tether/ops_linux.go
```diff
@@ -1072,3 +1079,54 @@ func createBindSrcTarget(files map[string]os.FileMode) error {
 
 	return nil
 }
+
+// LaunchUtility allows for starting, blocking on, and receiving the exit code of a
+// process while in the presence of an embedded child reaper.
+// The function passed in will be launched under lock and MUST NOT wait for the process to
+// exit. It's expected the function be a closure wrapped around StartProcess or similar.
+func (t *BaseOperations) LaunchUtility(fn UtilityFn) (<-chan int, error) {
```
vburenin: this call redirect is a little confusing. can we keep the code inside this method?

lib/tether/tether_linux.go
```diff
@@ -130,20 +130,27 @@ func (t *tether) childReaper() error {
...
 
 						t.handleSessionExit(session)
 						session.Unlock()
-					} else {
-						// This is an adopted zombie. The Wait4 call already clean it up from the kernel
-						log.Warnf("Reaped zombie process PID %d", pid)
+						continue
+					}
+
+					ok = t.ops.HandleUtilityExit(pid, exitCode)
```
gigawhitlocks: I think fundamentally this is what my PR was missing, this along w/ the `HandleUtilityExit` method. Also all those beautiful tests.

lib/tether/tether.go
```diff
@@ -464,6 +474,15 @@ func (t *tether) Start() error {
 	}
 	defer t.cleanup()
 
+	defer func() {
```
matthewavery: Should this be removed now? Or is this here to help with the CI run?

lib/tether/ops_linux.go
```diff
@@ -1072,3 +1079,54 @@ func createBindSrcTarget(files map[string]os.FileMode) error {
...
+	return launchUtility(t, fn)
+}
+
+func launchUtility(t *BaseOperations, fn UtilityFn) (<-chan int, error) {
```
vburenin: This function is definitely has to be a member of a structure, since it is extremely invasive from side effect stand point.

cmd/tether/net_linux_test.go
hickeng: Yup - not even sure why that's exported. I think it's carried over from tether package when I was exploring how to inherit mock structures (conclusion is that Go makes it hard).
The tether mocking in general needs to be completely revised - my preference is to leave it until vcsim is fleshed out with the ability to "run" workloads and see where we're lacking vs the OS interface that will require.

I've not looked at go-mock recently but it didn't seem sufficient when assembling the tether last year.

lib/tether/netfilter/netfilter_linux.go
hickeng: @willsherwood I think it's a useful addition anyway - it will avoid issues if the user is attempting to update a rule when a Reload happens for example. Absent any compelling reason I will leave it in.

cmd/tether/ops_linux.go
hickeng: A simple search/replace could be done pretty quickly. I generally used t as shorthand for `this` but more importantly for the moment it's consistent across the usages of operations - if it's not then the linter complains. I'm not inclined to make this change at this time.

lib/tether/net_linux_test.go
hickeng: implemented but I'm out of time.

cmd/tether/ops_linux.go
hickeng: I tried for days when hitting the same problem with `Apply` in lib/tether for the network config and did not find a better solution. I agree, and in general the entire test code supporting tether is convoluted and there's massive copy/paste between the `cmd/x_test` and `lib/x_test` files. A lot of it comes down to function being in _test files to exclude it from production build, but Go preventing the use of exported function from those across packages. This was at a time before the compiler did dead code elimination.

Given this is impacting master I'm not going to dive into that rat hole again now. It does need to be addressed however, and in a more pervasive/coherent manner.

lib/tether/ops_linux.go
hickeng: If you do that it becomes much more awkward to reference the implementation from the test Mock.

lib/tether/tether_linux.go
hickeng: note that none of those tests actually confirm the added function as written - they do exercise it though.

lib/tether/ops_linux.go
hickeng: Agree that it's not nice. Also not restructuring all the test code at this time. See prior comments :)

lib/tether/netfilter/netfilter_linux.go
hickeng: Removed given your updated on timing. We have a retry so should be functional viable in my proposed scenario.

lib/tether/tether.go
```diff
@@ -467,22 +467,23 @@ type TetherKey struct{}
 func (t *tether) Start() error {
 	defer trace.End(trace.Begin("main tether loop"))
 
+	defer func() {
+		e := recover()
+		if e != nil {
+			log.Errorf("Panic seem in main tether loop: %s: %s", e, debug.Stack())
```
willsherwood: s/seem/seen

hickeng:
@vburenin concur - it's part of the tether restructure work that @caglar10ur started at the beginning of the year. Unfortunately was put on hold for feature work.

cmd/tether/ops_linux.go
```diff
@@ -304,41 +334,46 @@ func setupPublishedFirewall(endpoint *tether.NetworkEndpoint, ifaceName string)
...
+	fn, err = (&netfilter.Rule{
 		Chain:           netfilter.Output,
 		Target:          netfilter.Accept,
 		Interface:       ifaceName,
 		Protocol:        netfilter.ICMP,
 		ICMPType:        netfilter.EchoReply,
 		SourceAddresses: sourceAddresses,
-	}).Commit(context.TODO()); err != nil {
+	}).Commit(ctx)
+	if err := invoke(t, fn, err, "allow poing outbound"); err != nil {
```
caglar10ur: poing -> ping

cmd/tether/ops_linux.go
caglar10ur: I prefer something other than t as well (like o in this case) but it's not a big of a deal

anchal-agrawal:
Requesting review from @rajanashok and removing mine since I'm not in the build-ci-infra group.

PR #5935, opened 2017-08-07
[skip ci] retry logic to add the VC license

rajanashok:
Seeing intermittent issues with the Enhanced-Linked-Mode nightly test while adding/assigning the VC license. Adding the retry logic to make it a bit robust.

```
Running command 'govc license.add <key>
${out} = govc: ServerFaultCode: Access to perform the operation was denied.
```


chengwang86:
@rajanashok There are two places in this test where we update the vcenter license. Shall we add retry logic to both?

rajanashok:
@chengwang86 good catch..will update

chengwang86:
To be honest, the reason that I added `govc license.add` is just due to a `one-time` nightly test failure (only occurred once). However, it seems that my original fix has caused `intermittent` failures now. This makes me wonder whether we should even care about that `one-time` failure at all and whether we should add more `govc` operations into the tests.

PR #5937, opened 2017-08-07
Populate /dev/random entropy using haveged

### LINKED ISSUE #5680: SECURITY: Populating /dev/random entropy should be built-in
**User Statement:**

As a container developer, I need a dynamically populated entropy pool when the container starts so that calls to /dev/random don't block.

**Details:**

When a container VM starts, it has zero entropy. As a result, it can take Tomcat server minutes to start while it waits for sufficient entropy. This is not the only problem inherent in low entropy. The quality of the entropy pool has a direct impact on the quality of SSL/TLS and other block ciphers.

User impact of this is that they have to either wait for minutes until their application is usable, they can switch to /dev/urandom (less secure) or they can install haveged and try to figure out how to make it run as a daemon process. 

This latter solution is the most "correct" approach, but containers are not designed to run multiple processes, so this forces the user to either start haveged with docker exec or to create a script that correctly deals with the signal handling. Both these options are pug ugly.

We already copy haveged into bootstrap.iso, but we never actually initialize and run it. I would strongly suggest that we investigate what it would take to do this, because I can guarantee that a customer is going to run into this issue and it will suck for them.

Long-run, this should be handled by para-virtualization and @hickeng has make this request. We cannot afford to wait for this capability before coming up with a reasonable solution.

**Acceptance Criteria:**

A container VM should have a dynamically populated entropy pool when it starts up.

caglar10ur:
Introducing the rngd (RDRAND) means having a very strict dependency on
the Broadwell architecture, which is roughly two years old.

Instead start carrying haveged with bootstrap and start it via tether.

Fixes #5680

### DIFF
--- a/cmd/tether/haveged_linux.go
+++ b/cmd/tether/haveged_linux.go
@@ -12,38 +12,35 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package tether
+package main
 
 import (
 	"fmt"
 	"os"
 	"os/exec"
 
-	"github.com/Sirupsen/logrus"
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/vmware/vic/lib/tether"
 )
 
-// Executor is a tether extension that wraps command
-type Executor struct {
+// Haveged is a tether extension that wraps command
+type Haveged struct {
 	p    *os.Process
 	exec func() (*os.Process, error)
 }
 
-// NewExecutor returns a tether.Extension that wraps the executor service
-func NewExecutor() *Executor {
-	return &Executor{}
-}
-
 // NewHaveged returns a tether.Extension that wraps haveged
-func NewHaveged() *Executor {
-	return &Executor{
+func NewHaveged() *Haveged {
+	return &Haveged{
 		exec: func() (*os.Process, error) {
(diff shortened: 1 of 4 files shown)


lib/tether/executer.go
```diff
@@ -0,0 +1,77 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
```
hickeng: Should be in `cmd/tether` - use the systemd-haveged in the appliance as is currently the case.

cmd/tether/haveged_linux.go
```diff
@@ -0,0 +1,74 @@
...
+	p    *os.Process
+	exec func() (*os.Process, error)
+}
+
+// NewHaveged returns a tether.Extension that wraps haveged
+func NewHaveged() *Haveged {
+	return &Haveged{
+		exec: func() (*os.Process, error) {
+			args := []string{"/.tether/lib/ld-linux-x86-64.so.2", "--library-path", "/.tether/lib", "/.tether/haveged", "-w", "1024", "-v", "1", "-F"}
+			// #nosec: Subprocess launching with variable
```
anchal-agrawal: 👍

PR #5955, opened 2017-08-08
[specific ci=1-02-Docker-Pull] Pull manifestSchemaV1 first to match the image manifest digest

### LINKED ISSUE #5958: docker pull integration test: match digest shasum output with vanilla docker's
**Details:**

We currently pull image manifest schema 2 to show a manifest digest shasum that matches vanilla docker's output: https://github.com/vmware/vic/blob/master/lib/imagec/imagec.go#L613

There seems to be a regression where we now pull the schema 2 manifest _before_ schema 1, causing the digest shasum to be overwritten by the schema 1 digest shasum. This results in the manifest digest output from VIC engine not matching docker's. This is being fixed in https://github.com/vmware/vic/pull/5955.

We should write an integration test for this scenario to ensure that we don't regress.

**Acceptance Criteria:**

Integration test (preferably in the regression suite) that pulls an image by name/tag using VIC and vanilla docker, and verifies that the digest shasums match.

chengwang86:
Currently we pull image manifestSchemaV2 first in docker pull. Although it does not cause any malfunction as of now, the image manifest digest would eventually be overwritten by the digest of manifestSchemaV1 (pulled after we pull manifestSchemaV2). This results in inconsistent screen output with vanilla docker w.r.t. image manifest digest. 

To fix it, this PR switches the order of pulling image manifests. 

Fixes #5958


mdubya66:
should this not run the specific ci for the pull tests, at a minimum?

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
```diff
@@ -152,3 +152,9 @@ Pull images from gcr.io
     Wait Until Keyword Succeeds  5x  15 seconds  Pull image  gcr.io/google_samples/gb-redisslave:v1
     Wait Until Keyword Succeeds  5x  15 seconds  Pull image  gcr.io/google_samples/cassandra:v11
     Wait Until Keyword Succeeds  5x  15 seconds  Pull image  gcr.io/google_samples/cassandra:v12
+
+Verify image manifest digest against vanilla docker
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull busybox:1.26
```
anchal-agrawal: We can use the `Pull image` robot keyword here.

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
```diff
@@ -152,3 +152,9 @@ Pull images from gcr.io
     Wait Until Keyword Succeeds  5x  15 seconds  Pull image  gcr.io/google_samples/gb-redisslave:v1
     Wait Until Keyword Succeeds  5x  15 seconds  Pull image  gcr.io/google_samples/cassandra:v11
     Wait Until Keyword Succeeds  5x  15 seconds  Pull image  gcr.io/google_samples/cassandra:v12
+
+Verify image manifest digest against vanilla docker
+    ${rc}  ${output}=  Run And Return Rc And Output  Pull image  busybox:1.26
```
anchal-agrawal: My bad - I don't think `Pull image` (in `Docker-Util.robot`) returns anything. We should probably just keep the original test logic.

PR #5957, opened 2017-08-08
Static analysis: annotate unhandled error cases

### LINKED ISSUE #4665: 1.2 static code analysis
Run static code analysis tools against the VIC product code.

For Go: https://github.com/GoASTScanner/gas

anchal-agrawal:
Fixes #4665

### DIFF
--- a/lib/install/validate/validator.go
+++ b/lib/install/validate/validator.go
@@ -660,6 +660,7 @@ func (v *Validator) reachableRegistries(ctx context.Context, input *data.Data, p
 	if len(input.WhitelistRegistries) > 0 {
 		// ignoring error since default merge policy is union, so should never return
 		// an error
+		// #nosec: Errors unhandled.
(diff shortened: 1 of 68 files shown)


lib/install/validate/validator.go
```diff
@@ -660,6 +660,7 @@ func (v *Validator) reachableRegistries(ctx context.Context, input *data.Data, p
 	if len(input.WhitelistRegistries) > 0 {
 		// ignoring error since default merge policy is union, so should never return
 		// an error
+		// #nosec: Errors unhandled.
 		m, _ := secureRegistriesSet.Merge(insecureRegistriesSet, nil)
```
emlin: in theory, this error should be checked, although I checked the code for Merge, there is no any error returned right now, just has error in the interface.
but if in the future, if there is new struct introduced and start to return error, here will have problem.

anchal-agrawal: Good point - it would be best to check with @hmahmood, but I think this is good for now. What do you think?

PR #5959, opened 2017-08-08
Remove not exercised stdin code

caglar10ur:
The session ends way before we have a chance to call ClostStdin().
Since session cleanup tears down the connection, we end up with a leaking
goroutine.

Seen in https://ci.vcna.io/vmware/vic/13043

### DIFF
--- a/cmd/tether/tether_test.go
+++ b/cmd/tether/tether_test.go
@@ -24,6 +24,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"sync"
 	"testing"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


mdubya66:
Approved

cmd/tether/tether_test.go
```diff
@@ -88,8 +91,10 @@ func (t *Mocker) Stop() error {
 
 // Reload implements the extension method
 func (t *Mocker) Reload(config *tether.ExecutorConfig) error {
-	// the tether has definitely finished it's startup by the time we hit this
-	close(t.Started)
+	t.once.Do(func() {
+		// the tether has definitely finished it's startup by the time we hit this
```
anchal-agrawal: s/it's/its

PR #5971, opened 2017-08-09
add test to compare vic image archive with docker image archive

### LINKED ISSUE #5850: Add data validation for docker commit
Story:
Have integration test to verify the vic archive file contains same content with docker archive, so that we can make sure docker commit command, based on VIC archive in PL, works correctly.

The test process is to get tar from docker save, and also get VIC generate tar file through imagec #5848. And then compare those two tar files.

For this test story, the criteria is as following:
- same files, no more no less
- same single file digest

We're not verifying tar file digest at this time, that is a goal for docker push

emlin:
fixes #5850 
This new test is to pull docker image archive and get VIC image archive through imagec command, and compare the layer files one by one.
The target of 1.2 is to get tar file content consistent and each file's digest parity, not the whole tar file digest parity.
This test can make sure we're returning exact same content during docker diff and commit.

This PR cannot be merged before @jzt fixed the production code


tests/test-cases/Group2-Docker-Archive/2-01-Docker-Archive.md
```diff
@@ -0,0 +1,29 @@
...
+
+# Environment:
+This test requires that a debug = 3 VIC appliance is available to execute the docker integration tests against
+
+# Test Steps:
+1. Deploy VIC appliance with debug = 3 to a test server
+2. Issue docker pull command to pull image ubuntu to VCH
+3. Issue imagec pull command to download image ubuntu layers tar from docker hub, which will store docker image layers to local file system
+4. Issue imagec save command to download image ubuntu tar files from PL, which will store VIC image layers to local file system
+5. Compare tar file content through tar -tvf
```
sflxn: I'm not sure you need to compare the individual asset in the tar files.  The tar digest should be enough.

sflxn: I see why you're doing this after our chat.

sflxn:
You should do this nightly only.  You can target the test for any changes to the commit path via ci specific.

tests/test-cases/Group2-Docker-Archive/2-01-Docker-Archive.robot
```diff
@@ -0,0 +1,114 @@
...
+    Run  mkdir -p /tmp/compare/${image}/fileA
+    Run  mkdir -p /tmp/compare/${image}/fileB
+    ${rc}  ${output}=  Run And Return Rc And Output  tar -xvf ${tarA} -C /tmp/compare/${image}/fileA
+    Log  ${output}
+    Should Be Equal As Integers  ${rc}  0
+    ${rc}  ${output}=  Run And Return Rc And Output  tar -xvf ${tarB} -C /tmp/compare/${image}/fileB
+    Log  ${output}
+    Should Be Equal As Integers  ${rc}  0
+
+    ${out}=  Run  find /tmp/compare/${image}/fileA -type f
```
sflxn: What about directories, UID/GID, and permissions?

emlin: all of those should be covered by the keyword: Compare Tar File Content
we'll using tar -tvf to dump out all files, and compare the output with docker tar file. if that passed, we could get exactly same file directories, UID/GID, permissions, and even file orders.

emlin:
> You should do this nightly only. You can target the test for any changes to the commit path via ci specific.

hmm.. we might keep it in CI for now?

sflxn:
lgtm

tests/test-cases/Group2-Docker-Archive/2-01-Docker-Archive.robot
```diff
@@ -0,0 +1,114 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License
+
+*** Settings ***
+Documentation  Test 2-01 - Docker Archive
+Resource  ../../resources/Util.robot
+Suite Setup  Install VIC Appliance To Test Server  additional-args=--debug=3
```
mhagen-vmware: We are already passing --debug 1 as the default level during all tests, not sure how this works (passing debug level 2x).  We should probably re-work Install VIC Appliance To Test Server to allow different debug levels.

emlin: vic-machine command will take the last option value if that's not a list type option. So this test works, but it might not be that ideal

mhagen-vmware:
Closing this in favor of #6056 @sflxn and @emlin Please review that PR now.

PR #5988, opened 2017-08-10
[specific ci=Group1-Docker-Commands] Have fetcher fail quickly on unretryable 4XX errors

### LINKED ISSUE #5966: [Harbor] VCH should not retry pulling when registry returns 4xx
This is another finding along #5950 

When the image doesn't match the project level policy, Harbor will return 412 for the "GET manifest" request.  After receiving 412 VCH keeps retrying 5 times, so it will take around a minute for the "pull" action to really fail.

Generally, I don't think VCH should retry upon receiving a 4xx error code from server.  It probably should fail right away except for receiving a 401.

willsherwood:
Fixes #5966.

Out of the 4XX codes, 'request timeout', 'locked', and 'too many requests' _could_ be resolved with a retry, so I excluded them from the retry block.


pkg/fetcher/fetcher.go
```diff
@@ -401,6 +406,19 @@ func (u *URLFetcher) IsStatusNotFound() bool {
 	return u.StatusCode == http.StatusNotFound
 }
 
+// IsClientError returns true if status code is an unretryable 4XX error. This includes all
```
reasonerjt: The comment doesn't match the function name?

sflxn:
lgtm.  Will, can you change from Unretryable to Nonretable?

pkg/fetcher/fetcher.go
```diff
@@ -401,6 +406,19 @@ func (u *URLFetcher) IsStatusNotFound() bool {
...
 
+// IsNonretryableClientError returns true if status code is a nonretryable 4XX error. This includes
+// all 4XX errors except 'request timeout', 'locked', and 'too many requests'. Additionally, it
+// does not return 'not found' or 'unauthorized' errors since those are handled uniquely.
+func (u *URLFetcher) IsNonretryableClientError() bool {
+	s := u.StatusCode
+	return 400 <= s && s < 500 &&
+		s != http.StatusNotFound &&
+		s != http.StatusUnauthorized &&
+		s != http.StatusRequestTimeout &&
```
caglar10ur: Why we are not retrying in StatusRequestTimeout case?

pkg/fetcher/fetcher.go
```diff
@@ -401,6 +406,19 @@ func (u *URLFetcher) IsStatusNotFound() bool {
 	return u.StatusCode == http.StatusNotFound
 }
 
+// IsNonretryableClientError returns true if status code is a nonretryable 4XX error. This includes
+// all 4XX errors except 'request timeout', 'locked', and 'too many requests'. Additionally, it
+// does not return 'not found' or 'unauthorized' errors since those are handled uniquely.
+func (u *URLFetcher) IsNonretryableClientError() bool {
```
caglar10ur: Would it be more readable/less error prone/and easy to modify if we just create a map of errors to bools and do a lookup here?

chengwang86: One more comment:

Maybe we can do something like this:
```
if u.IsNonretryableClientError() {
    if u.IsStatusUnauthorized() {
     }
     if u.IsStatusNotFound() {
    }
    ....
    err = fmt.Errorf("Nonretryable error '%s (%d)': URL %s.", http.StatusText(u.StatusCode), u.StatusCode, url)
    return nil, nil, DoNotRetry{Err: err}
}
```
By doing this, you don't need to exclude `NotFound` and `unauthoized` from `IsNonretryableClientError`.

pkg/fetcher/fetcher.go
willsherwood: Ah, I misread the description of status 408. StatusRequestTimeout should retry upon failure.

Issue #5829, opened 2017-07-25
Block polite stop if copy is in progress


hickeng:

jialin-li:
Investigation:

Online Import 
- docker stop in the middle of a cp operation results in an internal server error in guest operation
`Download error: ServerFaultCode: The guest operations agent could not be contacted.`
- govc vm.power -off and -suspend has similar result, toolbox datasink receives`500 Internal Server Error`

Will add error checking for internal server error from toolbox, propagate a retry message to user and notify possible truncated copy result.

mdubya66:
The 1.2 requirement is completed, removing from the release.

mdubya66:
not a requirment for 1.3, removing

PR #5992, opened 2017-08-10
[specific ci=1-*Docker-CP-*] add error checking for toolbox state change

jialin-li:
Fixes the first part of #5829

### DIFF
--- a/lib/portlayer/storage/vsphere/toolbox_common.go
+++ b/lib/portlayer/storage/vsphere/toolbox_common.go
@@ -15,9 +15,11 @@
 package vsphere
 
 import (
+	"net"
 	"net/url"
 	"path"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/vmware/govmomi/guest"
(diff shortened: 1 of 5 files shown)


lib/portlayer/storage/vsphere/toolbox_common.go
```diff
@@ -125,3 +128,34 @@ func isInvalidStateError(err error) bool {
...
+	if soap.IsSoapFault(err) {
+		switch soap.ToSoapFault(err).VimFault().(type) {
+		case types.InvalidState:
+			return true
+		case types.InvalidPowerState:
+			return true
+		case types.GuestOperationsUnavailable:
+			return true
+		case types.SystemError:
+			return strings.Contains(err.Error(), VIX_E_TOOLS_NOT_RUNNING)
```
jzt: Just curiou8s, but is there a more granular way to check this error? Do we know the format of this error type so that we could parse the code out? I ask because 3016 is bound to show up in a timestamp or something at some point .. although I don't know how large a variety of `types.SystemError`s we get back.

jzt: Also disregard this if timestamps aren't included in the error string (I think they probably aren't).

jialin-li: The error I've seen with 3016 looks like ` A general system error occurred: vix error codes = (3016, 0)`  and it's returned from toolbox calls. Don't think there's a timestamp.

matthewavery:
lgmt - looks great! good job!

PR #5991, opened 2017-08-10
[specific ci=Group1-Docker-Commands] Allow container traffic over loopback interface

willsherwood:
Fixes #5989.

PR #5994, opened 2017-08-10
[specific ci=1-17-Docker-Network-Connect] Add integration test for a container pinging itself when connected to multiple containers

### LINKED ISSUE #5660: First non-default bridge network created has non-functional nameserver in multi-network case
**User Statement:**

As a container developer, I need containers to be able to refer to each other using their name if they're deployed to the same custom network. This is documented here: https://vmware.github.io/vic-product/assets/files/html/1.1/vic_app_dev/network_use_cases.html#newbridge

Inconsistent name resolution will cause application failures on deployment.

**Details:**

If I run the following series of commands with a fresh VCH and try to ping the name of the container, it always fails.

```
docker network create bridge1
docker network create bridge2
docker create -it --name test1 --net bridge1 bensdoings/ifconfig
docker network connect bridge2 test1
docker start -ai test1
     $ ping test1
        ping: unknown host test1
```
However, if I install a fresh VCH and create and delete a bridge network before running the above commands, it works:

```
docker network create foo
docker network rm foo
docker network create bridge1
docker network create bridge2
docker create -it --name test1 --net bridge1 bensdoings/ifconfig
docker network connect bridge2 test1
docker start -ai test1
     $ ping test1
        PING test1 (172.18.0.2) 56(84) bytes of data.
        64 bytes from bridge1.localhost (172.18.0.2): icmp_seq=1 ttl=64 time=0.012 ms
        64 bytes from bridge1.localhost (172.18.0.2): icmp_seq=2 ttl=64 time=0.023 ms
```
There is something about the initialization of that first network that causes problems. I've tried this multiple times with the same results.

**Acceptance Criteria:**

This is a significant problem. It will severely impact the ability to deploy multi-container applications. I would suggest this should be high priority to fix.

We must have a regression test for this.

chengwang86:
Fixes #5660 

#5660 requires that a fresh VCH creates two bridge networks. So I move `Connect containers to multiple networks overlapping` to be the first sub-test in this test case.

The firewall rules in bootstrap are from PR #5991 . So it can go in after 5991 is merged.

### DIFF
--- a/tests/test-cases/Group1-Docker-Commands/1-17-Docker-Network-Connect.robot
+++ b/tests/test-cases/Group1-Docker-Commands/1-17-Docker-Network-Connect.robot
@@ -19,58 +19,6 @@ Suite Setup  Install VIC Appliance To Test Server  certs=${false}
 Suite Teardown  Cleanup VIC Appliance On Test Server
 
 *** Test Cases ***
-Container ping itself when created on multiple bridge networks
-    ${status}=  Get State Of Github Issue  5660
-    Run Keyword If  '${status}' == 'closed'  Fail  Test 1-07-Docker-Network-Connect.robot needs to be updated now that Issue #5660 has been resolved
-    
-    #${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} network create bridge1
-    #Should Be Equal As Integers  ${rc}  0
-    #${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} network create bridge2
-    #Should Be Equal As Integers  ${rc}  0
-    #${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull ${busybox}
-    #Should Be Equal As Integers  ${rc}  0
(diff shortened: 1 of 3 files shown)


tests/test-cases/Group1-Docker-Commands/1-17-Docker-Network-Connect.robot
```diff
@@ -19,12 +19,47 @@ Suite Setup  Install VIC Appliance To Test Server  certs=${false}
 Suite Teardown  Cleanup VIC Appliance On Test Server
 
 *** Test Cases ***
-Connect container to a new network
-    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} network create test-network
+Connect containers to multiple networks overlapping
```
willsherwood: 'Connect containers to multiple bridge networks' is a slightly more descriptive name.

Note that when a container is connected to an external vSphere network and a bridge network, a unique interface for each network is set up.

chengwang86: Thanks for pointing this out. I agree with you since all we do here is to create two bridge networks.

PR #6001, opened 2017-08-11
Add VmFolder as a field

caglar10ur:
That is the only use of Folders call, no need to call the same
function over and over again.


caglar10ur:
@andrewtchin @emlin I need one of you :)

Issue #5997, opened 2017-08-11
Whiteout files are created as hard links in docker images, but not in VICs


jzt:
As a user, I need immutable image data that I've acquired from docker to remain unchanged as it moves through its different forms (from tar archive -> vmdk -> tar archive).

Description:
tar archives produced by VIC's export backend are currently not 100% identical to the archives created by docker's backend.

A SHA256 of exported layers shows that we have a discrepancy between a single layer exported from the ubuntu image (marked with `->`), with the rest of the layers matching:

docker:
```
   zach@arch:~/tmp/theirs$ sha256sum *|sort
   25e0901a71b8c6a9df21590604a70517eb7b074071ef6af1033d50037baf3dd5  1300883d87d5ae0582a4e0f3f86ad93985b18676e8d829f6f3b4990403fdf145.tar
   5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef  a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4.tar
-> 625c7a2a783b4736cf488efd1fafc41736b9998ec087e20da946c30522ec9ad7  c220aa3cfc1bff0c1df733eac869a66476ea35136c69d881da1e297ed8d0af2f.tar
   8aa4fcad5eeb286fe9696898d988dc85503c6392d1a2bd9023911fb0d6d27081  d5c6f90da05dc7e77d2e5fef63c341ab05ba2a03396ab5ae8f18814a7bbf5265.tar
   9c42c2077cdea659ac116f148b63ded203d0e83f017accb6d7987377d1363673  2e9398f099dcc5a3e5521205e1f55ee5d7665e203193cf48b32553eabcca323c.tar
   a09947e71dc0d591901870f2fd3c18565339b960f6bb32793d14604a85d67393  dc27a084064fee108f4d486469939bbfa6a5b16f6d6d3c6b5b2475e938f4c71e.tar
```

VIC:
```
   zach@arch:~/tmp/ours$ sha256sum * |sort
   25e0901a71b8c6a9df21590604a70517eb7b074071ef6af1033d50037baf3dd5  9cb11a61de959ab1f5766e4f188e89c73e2b9016ee0acee7ad7f3ec0cd8f552d.tar
[...]

Issue #5871, opened 2017-07-28
Unpredictable nested volume behavior in tether


jialin-li:
As a developer, I hope to use nested volumes with predictable behavior. 

Right now volumes are stored in a map, and iterating through the volumes can give a different order each time. When starting a container with nested volumes, it may or may not succeed.

Tether debug log for this behavior is [here](https://gist.github.com/jialin-li/71bba93b1514ea4741ab6b0eaff7a454)
Step to reproduce (may need to run multiple times)
``` 
docker -H $HOST:2376 --tls create -i -v v1:/vol1 -v v2:/vol1/vol2 -v v3:/vol1/vol2/vol3 --name test busybox
docker -H $HOST:2376 --tls cp text.txt test:/vol1/vol2/vol3                                             
docker -H $HOST:2376 --tls start test

jialin-li:
closed by #5865

matthewavery:
This ticket was fixed by the PR which merged in docker-cp functionality(#5865)

PR #6002, opened 2017-08-11
Improve tar archive correctness [specific ci=Group1-Docker-Commands]

### LINKED ISSUE #5714: Add support in pack/unpack for missing file types
Applies to #5700 

Currently only the basic file types are supported, but this could easily lead to infinite reads on `/dev/random` and definitely results in a failure to correctly export layers.
I suggest switching to `vendor\github.com\docker\docker\pkg\archive\archive.go` instead of the recently added `lib/archive` and modifying it as needed, or using as a reference for file type behaviours that are missing.
Changes likely needed if switching to it:
* addition of xattr headers to denote Added/Changed/Deleted - can this be done as a WhitelistFormat?
* translation of Import/Export/Rebase into `TarOptions`
* modify to take *os.File inputs instead of string paths - this allows for subsequent use of a non-mounted FileWalker style interface, e.g. nfs client.

Whichever approach we use, tether will need to have the same level of capability, or we will have to drop online `copyFrom`.

**Acceptance**
* encompass single file support & rename for import e.g. `docker cp file.txt cid:/foo.txt`
* copy entire container, including FIFOs, device nodes, etc
* must not duplicate data for hardlinks
* should include unit tests for filterspec generation and use by archive package

jzt:
This change brings our tar archives closer in-line with dockers. File contents are now 100% correct, but there is one issue around whiteout files that doesn't affect us with regard to filesystem layout: https://github.com/vmware/vic/issues/5997.

This change meets all acceptance criteria in #5714.

The integration test for this will be modified from @emlin's test found here:  https://github.com/vmware/vic/pull/5971. 

Fixes #5714

### DIFF
--- a/lib/archive/diff.go
+++ b/lib/archive/diff.go
@@ -37,12 +37,40 @@ const (
 // CancelNotifyKey allows for a notification when cancelation is complete
 type CancelNotifyKey struct{}
 
+var (
+	seenFiles map[uint64]string
+)
+
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
 // for sort.Sort
 type changesByPath []docker.Change
 
-func (c changesByPath) Less(i, j int) bool { return c[i].Path < c[j].Path }
-func (c changesByPath) Len() int           { return len(c) }
-func (c changesByPath) Swap(i, j int)      { c[j], c[i] = c[i], c[j] }
+// This will walk the two path components up to min(i, j) length.
+// If at any time those components do not match, a comparison of the non-matching
+// components is returned.
+// If the shorter path is a prefix of the longer path, the shorter path wins.
+func (c changesByPath) Less(i, j int) bool {
+	a := strings.Split(c[i].Path[1:], "/")
+	b := strings.Split(c[j].Path[1:], "/")
+
+	m := min(len(a), len(b))
+	for x := 0; x < m; x++ {
+		if a[x] != b[x] {
+			return a[x] < b[x]
+		}
+	}
+
+	return len(a) < len(b)
+}
+
+func (c changesByPath) Len() int      { return len(c) }
+func (c changesByPath) Swap(i, j int) { c[j], c[i] = c[i], c[j] }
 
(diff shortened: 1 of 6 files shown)


mdubya66:
let's file an issue for some tests, @mhagen-vmware

emlin:
> can we get some tests for the new archive files?

The test will be in my PR #5871, but because of issue #5997, the test still cannot pass for archive file content is inconsistent.
Will update test case to make it pass after this PR is merged

lib/archive/diff.go
```diff
@@ -210,26 +240,54 @@ func createHeader(op trace.Operation, dir string, change docker.Change, spec *Fi
...
 		if err != nil {
 			op.Errorf("Error getting file info: %s", err.Error())
 			return nil, err
 		}
 
-		link, err := os.Readlink(filepath.Join(dir, change.Path))
-		if err != nil {
-			link = change.Path
+		link := ""
+		if fi.Mode()&os.ModeSymlink != 0 {
```
emlin: need spaces around of `&`

lib/archive/diff.go
```diff
@@ -210,26 +240,54 @@ func createHeader(op trace.Operation, dir string, change docker.Change, spec *Fi
...
+		name = strings.TrimPrefix(name, spec.StripPath)
+		name = filepath.Join(spec.RebasePath, name)
+
+		if fi.IsDir() && !strings.HasSuffix(change.Path, "/") {
+			name += "/"
+		}
+
+		hdr.Name = strings.TrimPrefix(name, "/")
+
+		inode, err := setHeaderForSpecialDevice(hdr, fi.Sys())
```
matthewavery: minor: it appears that we are using this function for a dual purpose here. Both to get the inode of the target as well as fixing the provided info. It might be more clear to provide the inode outside of this function. My thought on this is mainly because this function servers to correct the missing parts of the header for special devices and does not actually need the Inode at all to do this.

matthewavery:
lgtm - other than the minor comment posted above this changeset looks really well composed!

Issue #5819, opened 2017-07-25
Nightly Test 5-22- NFS Volume portlayer server error


sgairo:
Seen in 6.5:
[5-22-NFS-Volume.zip](https://github.com/vmware/vic/files/1173873/5-22-NFS-Volume.zip)

Seen in Personality:
```Jul 24 2017 07:02:41.304Z ERROR container attach: stderr (50b6f4a03b7a6d57dab7f00642d8eb580dd915095e6f9345b84af6b7bb22650c): Server error from portlayer: &{0 } (*models.Error) is not supported by the ByteStreamConsumer, can be resolved by supporting Writer/BinaryUnmarshaler interface
Jul 24 2017 07:02:41.332Z DEBUG * stdout attach end Mon Jul 24 07:02:41 UTC 2017
Jul 24 2017 07:02:41.332Z ERROR container attach: stdout (50b6f4a03b7a6d57dab7f00642d8eb580dd915095e6f9345b84af6b7bb22650c): Server error from portlayer: &{0 } (*models.Error) is not supported by the ByteStreamConsumer, can be resolved by supporting Writer/BinaryUnmarshaler interface
Jul 24 2017 07:02:41.332Z INFO  cleaned up connections to 50b6f4a03b7a6d57dab7f00642d8eb580dd915095e6f9345b84af6b7bb22650c. Checking errors
Jul 24 2017 07:02:41.332Z ERROR container attach error: Server error from portlayer: &{0 } (*models.Error) is not supported by the ByteStreamConsumer, can be resolved by supporting Writer/BinaryUnmarshaler interface
Jul 24 2017 07:02:41.332Z ERROR Operation failed with errors.apiError{error:(*errors.errorString)(0xc420168430), statusCode:500}
Jul 24 2017 07:02:41.332Z DEBUG [ END ] [github.com/vmware/vic/pkg/retry.DoWithConfig:73] [30.776609454s] 
Jul 24 2017 07:02:41.332Z DEBUG [ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).ContainerAttach:1596] [30.776642372s] 50b6f4a03b7a6d57dab7f00642d8eb580dd915095e6f9345b84af6b7bb22650c
[...]

PR #6012, opened 2017-08-14
init session logging earlier

emlin:
This PR fixes #6004 and second issue in #5819

PR #6044, opened 2017-08-15
fix byte consumer/producer issue

emlin:
to make it parse error correctly

Together with #6012, fixes: #5819

### DIFF
--- a/lib/apiservers/portlayer/restapi/configure_port_layer.go
+++ b/lib/apiservers/portlayer/restapi/configure_port_layer.go
@@ -17,6 +17,9 @@ package restapi
 import (
 	"context"
 	"crypto/tls"
+	"encoding/json"
+	errs "errors"
+	"io"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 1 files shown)


lib/apiservers/portlayer/restapi/configure_port_layer.go
```diff
@@ -17,6 +17,9 @@ package restapi
 import (
 	"context"
 	"crypto/tls"
+	"encoding/json"
+	errs "errors"
```
caglar10ur: fmt.Errorf and you no longer need to import this :)

morris-jason:
Should we make the same change for TarConsumer/Producer?

lib/apiservers/portlayer/restapi/configure_port_layer.go
willsherwood: Inconsistent with how we have imported "errors" without the alias in all other files.

lib/apiservers/portlayer/restapi/configure_port_layer.go
```diff
@@ -163,3 +167,38 @@ func setupMiddlewares(handler http.Handler) http.Handler {
 func setupGlobalMiddleware(handler http.Handler) http.Handler {
 	return handler
 }
+
```
willsherwood: exported function might need a brief comment to explain what it does, like ByteStreamProducer.

emlin:
> Should we make the same change for TarConsumer/Producer?

This is actually a good question. Did you ever find this kind of error? I could not reproduce it even I broken the cp process.

lib/apiservers/portlayer/restapi/configure_port_layer.go
emlin: this is from @casualjim 😄

PR #6050, opened 2017-08-16
[specific ci=Group1-Docker-Commands] Mask lost+found folder in volumes and untangle transitive imports

### LINKED ISSUE #5777: Docker volume /lost+found folder is incompatible with popular images that are potentially hard to reconfigure
**User Statement:**

As a container developer, I expect that popular images such as postgres, mysql, redis and many others will work with VIC volumes without modification. 

**Details:**

We've had a bug open for this issue before. It was #2929. The assertion in the bug was that this is likely only a problem for postgres and that there was a reasonable workaround. As such, I think at the time it seemed expedient to continue the incompatibility. 

What's been clear to me though working with scenario-based testing recently is that this problem is far more widespread than just postgres. Take mysql as an example. It took me an hour trawling through documentation to figure out how to configure mysql to write its data to a subdirectory of the mounted volume. And this was just to work around a VIC incompatibility. 

Here's an example. If you want a named volume with mysql with docker, you use:

`docker run -v myvol:/var/lib/mysql mysql:5.7`

This doesn't work with VIC. Instead, you have to specify:

`docker run -v myvol:/var/lib/mysql mysql:5.7 --datadir=/var/lib/mysql/data`

There is no environment variable to achieve this.

It's one thing to say to a customer, "you just have to specify a subdirectory", but if they then have to spend time figuring out how to reconfigure the container they're running... that's a very bad UX.

To be fair, this will be the case with any ext4 disk-based volume implementation. I'm quite sure it's true of the vSphere Docker Volume Service. I'm going to make the case though that we need to do better.

**Acceptance Criteria:**

The alternative is that we should take the top N most popular docker images and document how to work around this problem for each of them. I don't want that job :)

I think we can pull of this subdirectory trick ourselves by automatically creating a /data dir in every mounted volume and creating a symbolic link to it. Deleting /lost+found is not the right answer.

jialin-li:
Fixes #5777 
Includes 2 commits from @hickeng 

Mask logic mainly in `lib/portlayer/storage/vsphere/volume.go` and `lib/tether/ops_linux.go`, updated `lib/tether/toolbox.go` to make online cp compatible with the new mask and mount logic.

* Added change to `disk.go` to reflect new logic.


lib/tether/ops_linux.go
```diff
@@ -767,6 +798,34 @@ func (t *BaseOperations) MountLabel(ctx context.Context, label, target string) e
...
+	// do..while ! timedout
+	var timeout bool
+	for timeout = false; !timeout; {
+		_, err := os.Stat(label)
+		if err == nil || !os.IsNotExist(err) {
+			break
+		}
+
+		deadline, ok := ctx.Deadline()
+		timeout = ok && time.Now().After(deadline)
```
caglar10ur: This pattern looks suspicious, why we are not simply waiting os.Stat and checking ctx.Done for timeout?

lib/tether/ops_linux.go
```diff
@@ -728,24 +739,44 @@ func (t *BaseOperations) MountLabel(ctx context.Context, label, target string) e
...
 
-	if timeout {
-		detail := fmt.Sprintf("timed out waiting for %s to appear", label)
-		return errors.New(detail)
+	// at this point bindTarget should be mounted successfully
+
+	mntsrc := path.Join(bindTarget, volumeDataDir)
+	mnttype := "bind"
+	mntflags := uintptr(syscall.MS_BIND)
+	// if the volume contains a volumeDataDir directory then mount that instead of the root of the filesystem
```
matthewavery: Other than for backwards compatibility with previous versions of VIC I think this approach is interesting. Since we are in charge of making the data dir (and this should likely be done at `volume create` time, haven't looked yet sorry...) we would definitely expect one to be present. Ideally if we were looking to be backwards compatible with previous native volumes of VIC we would care mostly about volumes that had something other that `lost+found` in their root directory. Otherwise it would be an empty volume anyways and we might make special logic to simply add a data dir to it. lmkwyt @jialin-li @caglar10ur @hickeng

pkg/vsphere/disk/disk.go
```diff
@@ -260,19 +268,30 @@ func (d *VirtualDisk) Mount(op trace.Operation, options []string) (string, error
...
+		mntsrc, err := d.getMountSource(op, options)
+		if err != nil {
+			op.Error(err)
+			return "", err
+		}
+
+		// then bind mount it at the correct source
+		// append bind mount options because mntsrc is already mounted
+		options = append(options, "bind")
+		// device path is not right
```
matthewavery: should this comment be moved lower? or possibly removed?

tests/test-cases/Group1-Docker-Commands/1-04-Docker-Create.robot
```diff
@@ -46,7 +46,7 @@ Create with anonymous volume
 
 Create with named volume
     ${disk-size}=  Run  docker %{VCH-PARAMS} logs $(docker %{VCH-PARAMS} start $(docker %{VCH-PARAMS} create -v test-named-vol:/testdir ${busybox} /bin/df -Ph) && sleep 10) | grep by-label | awk '{print $2}'
-    Should Be Equal As Strings  ${disk-size}  975.9M
+    Should Contain  ${disk-size}  975.9M
```
mhagen-vmware: this would seemingly be unrelated, why are we making all these disk size changes? And why would be we make it less narrow?

jialin-li: Because we mount the vmdk rootfs and then bind mount its datadir, so when running the above command, two disks show up and the `975.9M` shows up twice so it no longer equals to the string with just one `975.9M`. That's why I had to change it to contain rather than equal to an exact string.

sflxn:
lgtm.

I'll let the storage guys and Caglar to look closely at all the mounting code.  I did the best I could.
Disks showing up twice when performing df may confuse users who don't know why it shows up, but this is something we can document and should not effect container execution.  It's still better than not running containers out of the box without working around the lost+found folder.  @stuclem We should document the fact that df will show disks twice and why (jian's explanation above).

gigawhitlocks:
the one line change in vicadmin lgtm lol

matthewavery:
Can confirm that older volumes still do not work, but newly created volumes in the old volume store will. This is really good work! :) We would ideally have some tests for backwards compaitibility and to show that we can take an older version of VIC that will not work on those images and then install a newer one with the same volume store and they will work. Users can even migrate their data to the new volume if need be if there was anything important. However, most of these images expect a completely empty volume either way, so this is not a major issue with not being backwards compatible at all. The biggest thing with including this pr is that @pdaigle @karthik-narayan will have to let customers know that they absolutely must upgrade to this version if they want these kinds of images to work. @mdubya66 I am signing off on this for 1.2 though I think we will need more tests I will put together a ticket for that when I get the chance.

matthewavery:
lgtm - nice work @jialin-li and @hickeng this one has been needed for quite some time.

PR #6056, opened 2017-08-17
[specific ci=Group2-Docker-Archive] Initial implementation of docker archive tests

### LINKED ISSUE #5850: Add data validation for docker commit

mhagen-vmware:
fixes #5850


tests/test-cases/Group2-Docker-Archive/2-01-Docker-Archive.robot
```diff
@@ -0,0 +1,114 @@
...
+    ${rc}  ${output}=  Run And Return Rc And Output  tar -tvf ${tarB} > /tmp/compare/${image}/b
+    Should Be Equal As Integers  ${rc}  0
+    Should Be Empty  ${output}
+    ${out}=  Run  cat /tmp/compare/${image}/b
+    Log  ${out}
+
+    ${rc}  ${output}=  Run And Return Rc And Output  diff /tmp/compare/${image}/a /tmp/compare/${image}/b
+    Log  ${output}
+    Should Be Equal As Integers  ${rc}  0
+    Should Be Empty  ${output}
```
emlin: It's here to do more work for the diff output.
We still could not get actually file content consistent, because we generated different .wh... file with docker.

PR #6058, opened 2017-08-17
Correct log routes in vic-admin

### LINKED ISSUE #5729: Problem curl'ing port layer log
I've observed two issues in this build (#5700): https://ci.vcna.io/vmware/vic/12327

1. Group0-Bugs.5343
2. Group9-VIC-Admin.9-02-VICAdmin-CertAuth

Both of these relate to not finding expected content in the port layer log. The second provides a really good hint at why the first failed:

```
Get Portlayer Log                                                     | FAIL |
'<!DOCTYPE html>
<html itemscope="" itemtype="http://schema.org/Organization" lang="en-US">
  <head>
    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
    <title>VIC: VCH-12327-8972</title>
    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1">
    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0">
    <meta content="yes" name="apple-mobile-web-app-capable">
    <meta name="apple-mobile-web-app-status-bar-style" content="black-translucent">
    <link rel="stylesheet" href="css/clarity-ui.min.css">
    <link rel="stylesheet" href="css/fontello.css">
    <link rel="stylesheet" href="css/style.css">
  </head>
    [ Message content over the limit has been removed. ]
                  <div class="sixty"><a href="/logs/init.log">Initialization & watchdog</a></div>
                  <div class="forty"><a href="/logs/tail/init.log">Live Log</a></div>
                </div>
                <div class="row">
                  <div class="sixty"><a href="/logs/vicadmin.log">Admin Server</a></div>
                  <div class="forty"><a href="/logs/tail/vicadmin.log">Live Log</a></div>
                </div>
              </div>
            </div>
          </div>
        </section>
      </main>
    </div>
  </body>
</html>' does not contain 'Launching portlayer server'
```

It looks like we've got the index page instead of the portlayer log. Took a look at vicadmin log, but didn't see an obvious cause, however I'm unfamiliar with normal for that log. Did wonder why there's so many Authenticated blocks in a row when we start.

Most likely unrelated is why we're not gathering profile data on any of the pprof endpoints:
```
Jul 16 2017 00:24:18.899Z DEBUG [BEGIN] [main.urlReader.open:292] http://127.0.0.1:6060/debug/pprof/profile
Jul 16 2017 00:24:20.899Z DEBUG [ END ] [main.urlReader.open:292] [2.000216475s] http://127.0.0.1:6060/debug/pprof/profile
Jul 16 2017 00:24:20.899Z WARN  error reading vic-init/profile(http://127.0.0.1:6060/debug/pprof/profile): Get http://127.0.0.1:6060/debug/pprof/profile: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
```

morris-jason:
Fixes #5729 and fixes broken log rotation functionality.

### DIFF
--- a/cmd/vicadmin/logs.go
+++ b/cmd/vicadmin/logs.go
@@ -20,11 +20,9 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"math"
 	"os"
 	"path"
-	"strings"
 	"sync"
 	"time"
 
@@ -104,28 +102,12 @@ func (r dlogReader) open() (entry, error) {
 // cross process synchronization.
 func logFiles() []string {
 	defer trace.End(trace.Begin(""))
-	files, err := ioutil.ReadDir(logFileDir)
-	if err != nil {
-		log.Errorf("Failed to get a list of log files: %s", err)
-		return nil
-	}
 
 	names := []string{}
-	for _, fileInfo := range files {
-		if fileInfo.IsDir() {
-			continue
-		}
-		fname := fileInfo.Name()
-		log.Debugf("Found potential file for export: %s", fname)
-
-		for _, f := range logFileListPrefixes {
-			if strings.HasPrefix(fname, f) {
-				fp := filepath.Join(logFileDir, fname)
-				log.Debugf("Adding file for export: %s", fp)
-				names = append(names, fp)
-				break
-			}
-		}
+	for _, f := range logFileList {
+		fp := filepath.Join(logFileDir, f)
+		log.Debugf("Adding file for export: %s", fp)
+		names = append(names, fp)
 	}
 
 	return names
--- a/cmd/vicadmin/logs.go
+++ b/cmd/vicadmin/logs.go
@@ -18,11 +18,16 @@ import (
 	"archive/tar"
 	"archive/zip"
 	"bytes"
+	"compress/gzip"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"math"
+	"net/http"
 	"os"
 	"path"
+	"sort"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


gigawhitlocks:
Looks good, but can you run the VicAdmin CI suite before you merge?

vburenin:
as we discussed offline, this PR is not solving the root cause of the problem, in fact it makes it impossible to export rotated logs. There is a different place to look at that should completely eliminate the issue.

morris-jason:
@vburenin I've reworked to include rotated logs.

cmd/vicadmin/logs.go
```diff
@@ -104,31 +105,58 @@ func (r dlogReader) open() (entry, error) {
 // cross process synchronization.
 func logFiles() []string {
 	defer trace.End(trace.Begin(""))
+
```
morris-jason: Look at the raw file, this is a bad diff.

cmd/vicadmin/server.go
```diff
@@ -262,7 +263,7 @@ func (s *server) Authenticated(link string, handler func(http.ResponseWriter, *h
 		log.Infof("User with a valid auth cookie at %s is authenticated.", connectingAddr[0])
 		handler(w, r)
 	}
-	s.mux.Handle(link, gorillacontext.ClearHandler(http.HandlerFunc(authHandler)))
+	s.Handle(link, http.HandlerFunc(authHandler))
```
morris-jason: @gigawhitlocks Is this ok, using the custom Handle? Seemed fine in tests.

cmd/vicadmin/logs.go
```diff
@@ -131,6 +134,72 @@ func logFiles() []string {
...
+	}
+
+	// write file contents to w
+	for _, fileName := range names {
+		file, err := os.Open(fileName)
+		if err != nil {
+			log.Errorf("error opening file %s: %s", fileName, err.Error())
+			continue
+		}
+		var r io.ReadCloser = file
```
vburenin: not sure assigning it to interface makes sense, just use file as it is.

cmd/vicadmin/logs.go
```diff
@@ -131,6 +134,72 @@ func logFiles() []string {
...
+		if strings.HasSuffix(fileName, "gz") {
+			r, err = gzip.NewReader(file)
+			if err != nil {
+				log.Errorf("error opening gzipped file %s: %s", fileName, err.Error())
+				continue
+			}
+		}
+		defer r.Close()
```
vburenin: close this explicitly at the end no need defer in the loop, it may lead to resource exhaustion due to too many files are open.

cmd/vicadmin/logs.go
morris-jason: Added comment. Reassigning r to a gzip reader if needed.

cmd/vicadmin/logs.go
morris-jason: Done.

cmd/vicadmin/server.go
gigawhitlocks: Uhm, I think you still need to nest the `ClearHandler` because it clears out context related to the session at the end of the request, as I recall.. docs are here http://www.gorillatoolkit.org/pkg/context

Unfortunately I don't remember exactly what the side effect of not using it is, but I believe it amounts to leaking something we don't want to leak.

morris-jason: https://github.com/vmware/vic/blob/master/cmd/vicadmin/server.go#L161
s.Handle will make a call to ClearHandler with the httlHandler.

gigawhitlocks: good stuff

PR #6061, opened 2017-08-18
[specific ci=Group10-VCH-Restart] restart issues with the personality cache

matthewavery:
fixes the cmd portion of #5654 There are some portions that show as `null` still when they are empty. But this is not a major concern. It has been found that things look healthy on the portlayer side so this is really just an issue of us not propagating the needed information back. a regression test has been written for this that is tied into the VCH-Restart tests. We add a field to the container config returned in the container info struct from the portlayer. Now the `CMD` data is returned anytime a container's info object is obtained. and more specifically it is returned when we get a container list on first cache hydration of  the persona at boot(at restart of the VCH).

### DIFF
--- a/lib/apiservers/engine/backends/backends.go
+++ b/lib/apiservers/engine/backends/backends.go
@@ -30,8 +30,6 @@ import (
 	rc "github.com/go-openapi/runtime/client"
 	"github.com/go-openapi/swag"
 
-	"github.com/davecgh/go-spew/spew"
-
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/container"
 	vicproxy "github.com/vmware/vic/lib/apiservers/engine/proxy"
@@ -291,9 +289,6 @@ func syncContainerCache() error {
 		if err = setPortMapping(info, backend, container); err != nil {
 			errs = append(errs, err.Error())
 		}
-		log.Debugf("%s ------> %s", info.ContainerConfig.Cmd, container.Config.Cmd)
-
-		log.Debugf("\n\n %s", spew.Sdump(container))
 	}
 
 	if len(errs) > 0 {
--- a/lib/apiservers/engine/backends/backends.go
+++ b/lib/apiservers/engine/backends/backends.go
@@ -30,6 +30,8 @@ import (
 	rc "github.com/go-openapi/runtime/client"
 	"github.com/go-openapi/swag"
 
+	"github.com/davecgh/go-spew/spew"
+
 	"github.com/vmware/vic/lib/apiservers/engine/backends/cache"
 	"github.com/vmware/vic/lib/apiservers/engine/backends/container"
 	vicproxy "github.com/vmware/vic/lib/apiservers/engine/proxy"
@@ -289,7 +291,11 @@ func syncContainerCache() error {
 		if err = setPortMapping(info, backend, container); err != nil {
 			errs = append(errs, err.Error())
 		}
+		log.Debugf("%s ------> %s", info.ContainerConfig.Cmd, container.Config.Cmd)
+
+		log.Debugf("\n\n %s", spew.Sdump(container))
 	}
+
 	if len(errs) > 0 {
 		return errors.Errorf("Failed to set port mapping: %s", strings.Join(errs, "\n"))
 	}
--- a/lib/apiservers/engine/backends/backends.go
+++ b/lib/apiservers/engine/backends/backends.go
@@ -204,6 +204,11 @@ func hydrateCaches() error {
 	if len(errs) > 0 {
 		e = fmt.Errorf(strings.Join(errs, ", "))
 	}
+
+	if e != nil {
+		log.Errorf("Errors occurred during cache hydration at VCH start: %s", e)
+	}
+
 	return e
 }
 
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -1819,7 +1819,7 @@ func ContainerInfoToVicContainer(info models.ContainerInfo) *viccontainer.VicCon
 		log.Debugf("add volume mount %s to config.volumes and hostconfig.binds", mount)
 	}
 
-	vc.Config.Cmd = info.ContainerConfig.Cmd
+	vc.Config.Cmd = info.ProcessConfig.ExecArgs
 
 	return vc
 }
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -1794,6 +1794,7 @@ func ContainerInfoToVicContainer(info models.ContainerInfo) *viccontainer.VicCon
 	if info.ContainerConfig.LayerID != "" {
 		vc.LayerID = info.ContainerConfig.LayerID
 	}
+
 	if info.ContainerConfig.ImageID != "" {
 		vc.ImageID = info.ContainerConfig.ImageID
 	}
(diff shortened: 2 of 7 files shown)


lib/apiservers/portlayer/swagger.json
```diff
@@ -3295,6 +3295,12 @@
 						"type": "string"
 					}
 				},
+        "cmd": {
```
morris-jason: grrrrrrr, spacing.

tests/test-cases/Group10-VCH-Restart/10-01-VCH-Restart.robot
```diff
@@ -132,16 +132,22 @@ Create VCH attach disk and reboot
...
+    ${rc}  ${container}=  Run And Return Rc And Output  docker %{VCH-PARAMS} create --name=mount-data-test -v /mnt/test -v named-volume:/mnt/named busybox /bin/ls -la /
     Should Be Equal As Integers  ${rc}  0
     ${rc}  ${out}=  Run And Return Rc And Output  docker %{VCH-PARAMS} inspect -f '{{.Mounts}}' mount-data-test
     Should Be Equal As Integers  ${rc}  0
     Should Contain  ${out}  /mnt/test
     Should Contain  ${out}  /mnt/named
 
+    ${rc}  ${out}=  Run And Return Rc And Output  docker %{VCH-PARAMS} inspect -f '{{.Config.Cmd}}' mount-data-test
+    Should Be Equal As Integers  ${rc}  0
+    Should Contain  ${out}  /bin/ls
```
hmahmood: Can you check if this does not appear more than once in `$out`?

matthewavery: I will look into how to do that.

sflxn: Why would it appear more than once?

matthewavery: It should not, but if someone changes it in the future and it starts to we want the test to fail. I found that the `sessions.Path` and the sessions.Args have a shared element which originally had a duplicate. Easy enough to fix, but is nice if the test would protect us from regressing back to accidentally having a duplicate item.

lib/apiservers/engine/backends/backends.go
```diff
@@ -290,6 +290,7 @@ func syncContainerCache() error {
 			errs = append(errs, err.Error())
 		}
 	}
+
```
mhagen-vmware: here

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1794,6 +1794,7 @@ func ContainerInfoToVicContainer(info models.ContainerInfo) *viccontainer.VicCon
 	if info.ContainerConfig.LayerID != "" {
 		vc.LayerID = info.ContainerConfig.LayerID
 	}
+
```
mhagen-vmware: here

lib/apiservers/portlayer/swagger.json
```diff
@@ -3875,4 +3881,4 @@
 			}
 		}
 	}
-}
\ No newline at end of file
+}
```
mhagen-vmware: here

lib/portlayer/exec/exec.go
```diff
@@ -124,9 +125,12 @@ func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSou
 		// sync container cache
 		vmSub.Suspend(true)
 		defer vmSub.Resume()
+		log.Info("Syncing container cache")
 		if err = Containers.sync(ctx, sess); err != nil {
+			log.Infof("Error encountered during container cache sync during init process: %s", err)
 			return
 		}
+
```
mhagen-vmware: here

lib/apiservers/engine/backends/container_proxy.go
matthewavery: This is separating two if statements. in the same manner as the code below it.

lib/apiservers/portlayer/swagger.json
matthewavery: :/ stupid spacemacs

lib/apiservers/engine/backends/backends.go
matthewavery: This seemed really cluttered with the if statement pushed right against the for loop.

lib/apiservers/portlayer/swagger.json
```diff
@@ -3298,6 +3298,12 @@
 						"type": "string"
 					}
 				},
+        			"cmd": {
+					"type": "array",
+					"items": {
+						"type": "string"
+					}
+				},
```
caglar10ur: There shouldn't be a cmd in the ContainerConfig struct. Sessions have their own world (see ProcessConfig)

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -562,6 +562,8 @@ func convertContainerToContainerInfo(c *exec.Container) *models.ContainerInfo {
 		info.ContainerConfig.AttachStderr = &session.Attach
 		info.ContainerConfig.OpenStdin = &session.OpenStdin
 
+		info.ContainerConfig.Cmd = append(info.ContainerConfig.Cmd, session.Cmd.Args...)
```
caglar10ur: I think you can use continer.ProcessConfig  here

lib/apiservers/portlayer/swagger.json
matthewavery: I gave up on this... Not sure why it keeps getting added in but shouldn't hurt anyone and I wanted to look at the lost+found issue

lib/apiservers/engine/backends/backends.go
```diff
@@ -204,6 +204,11 @@ func hydrateCaches() error {
 	if len(errs) > 0 {
 		e = fmt.Errorf(strings.Join(errs, ", "))
 	}
+
+	if e != nil {
+		log.Errorf("Errors occurred during cache hydration at VCH start: %s", e)
```
emlin: the error is logged out in main()

matthewavery: Oh I see, this logging is much closer to the issue though. Had a hard time finding these errors in the logs.

lib/portlayer/exec/exec.go
```diff
@@ -40,6 +40,7 @@ var (
 )
 
 func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSource, _ extraconfig.DataSink) error {
+	log.Info("Beginning initialization of portlayer exec component")
```
cgtexmex: Why Info and not Debug?

cgtexmex:
a couple of nits around logging level, but overall LGTM

lib/portlayer/exec/exec.go
```diff
@@ -124,7 +125,9 @@ func Init(ctx context.Context, sess *session.Session, source extraconfig.DataSou
 		// sync container cache
 		vmSub.Suspend(true)
 		defer vmSub.Resume()
+		log.Info("Syncing container cache")
 		if err = Containers.sync(ctx, sess); err != nil {
+			log.Infof("Error encountered during container cache sync during init process: %s", err)
```
matthewavery: good point!

lib/portlayer/exec/exec.go
matthewavery: I left this as info since it will only occur once on every start of the persona. It can be handy to see this log in customer log bundles since customers are likely not going to be running a debug version of VIC.

Issue #5940, opened 2017-08-07
Nightly 5-22: container creation failed because context deadline exceeded


sflxn:
8/6, build 12979, vSphere 6.5
[5-22-NFS-Volume.zip](https://github.com/vmware/vic/files/1206468/5-22-NFS-Volume.zip)

CDE error during container handle commit.  From portlayer logs:

```
Aug  7 2017 07:03:40.008Z DEBUG [ END ] [github.com/vmware/vic/lib/portlayer/exec.Commit:36] [3m13.826296994s] 4b312973d580e04b972a4f4fdda677672157c090f329d442d77ced9001824475
Aug  7 2017 07:03:40.008Z ERROR CommitHandler error on handle(cc33a23b46aa0f9439355fa4d958cfab) for 4b312973d580e04b972a4f4fdda677672157c090f329d442d77ced9001824475: &errors.errorString{s:"unable to wait for process launch status: Post https://10.160.131.72/sdk: context deadline exceeded"}
Aug  7 2017 07:03:40.008Z DEBUG [ END ] [github.com/vmware/vic/lib/apiservers/portlayer/restapi/handlers.(*ContainersHandlersImpl).CommitHandler:216] [3m13.826798326s] handle(cc33a23b46aa0f9439355fa4d958cfab)
```

emlin:
@sflxn @hickeng @caglar10ur 
This error is because our nimbus nightly test environment is super slow.

Our current docker attach timeout is 30s, if tether is slow, like in our nightly test environment, attach timeout before tether start to wait for clear to launch. So even after the connection is established after 30s, attach already quit, so no one will send the clear to launch for tether, and then tether stuck in the waiting clear to launch status.

But in the other hand, docker commit does not timeout after 30s. The default property collector timeout is 3m.
In one docker run command, docker attach will be called first, and then docker start, so docker attach will timeout first even with same timeout settings.
But now we have much shorter attach timeout, 30s, so literally, PL will do nothing after first 30s until property collector timeout after 3m.

This is the reason for the nightly test failure, so the suggestion is to bump docker attach timeout to 3m, same to PL internal timeout. At least it will accommodate the slow environment a little bit.

rajanashok:
Seen this again on 6.0 8/22 nightly run. The zip file is too big to attach. Log URL below.

https://storage.cloud.google.com/vic-ci-logs/functional_logs_08_22_18_30_vic_13338.tar.gz.zip?authuser=1

[...]

Issue #6062, opened 2017-08-18
docker attach timeout enhancement


emlin:
So far the default attach timeout is hard coded a short value, and mismatch with other timeout in portlayer, e.g. the property collector, which is 3m.
The reason is because docker attach will hold docker cli, until the attach connection is established or timed out. Before that, whatever input does not have any response, including ctrl+C/D.

To avoid this awkward situation, we set a short timeout for attach, but that does not work for slow environment. In portlayer log, user will see this kind of error message:

```
Aug  7 2017 07:06:52.028Z DEBUG HandshakeClient: Sending syn with pos 183
Aug  7 2017 07:06:53.028Z DEBUG HandshakeClient: Sending syn with pos 244
Aug  7 2017 07:06:54.028Z DEBUG HandshakeClient: Sending syn with pos 29
Aug  7 2017 07:06:55.029Z DEBUG HandshakeClient: Sending syn with pos 10
Aug  7 2017 07:06:55.251Z ERROR attach connector: Connection not found error for id:61d553db2f3fe04dff420ce9de8fab0b138dd9fd3795f831f22f62402f935820: context deadline exceeded
Aug  7 2017 07:06:55.251Z DEBUG [ END ] [github.com/vmware/vic/lib/portlayer/attach/communication.(*Connector).interaction:128] [29.99949287s] 61d553db2f3fe04dff420ce9de8fab0b138dd9fd3795f831f22f62402f935820
Aug  7 2017 07:06:55.251Z DEBUG [ END ] [github.com/vmware/vic/lib/portlayer/attach/communication.(*Connector).Interaction:111] [29.999547547s] 61d553db2f3fe04dff420ce9de8fab0b138dd9fd3795f831f22f62402f935820
Aug  7 2017 07:06:55.251Z DEBUG [ END ] [github.com/vmware/vic/lib/portlayer/attach/communication.(*Server).Interaction:104] [29.999566015s] 61d553db2f3fe04dff420ce9de8fab0b138dd9fd3795f831f22f62402f935820
Aug  7 2017 07:06:55.251Z ERROR attach connector: Connection not found error for id:61d553db2f3fe04dff420ce9de8fab0b138dd9fd3795f831f22f62402f935820: context deadline exceeded
[...]

PR #6063, opened 2017-08-18
bump attach timeout to 50s

emlin:
So far the attach timeout (30s) is too short for slow environment. But we don't want to increase the timeout to super long time, because before attach connection is established, docker cli hung there without any response to user input, including ctrl+C/D
If docker attach timeout, docker cli will accept ctrl+C to stop the command.

Now to avoid the awkward time period mentioned above, we set a short timeout for docker attach, but this timeout does not match with other timeout in portlayer, e.g. the property collector, which is 3m. So even after docker attach timeout, docker run will not return until 3m later. The only way to stop that is to ctrl+C.

This PR is just one workaround for usually in a slightly slower environment, the attach connection takes 40 seconds. Bump to 50 seconds, could accommodate slow environment a little bit better, and not that long time to wait.

The long term solution for this issue is to remove the attach timeout, and parsing the stdin stream data. If there is ctrl+C/D, stop all the work, and return an error to say that user cancelled the operation. But this is not likely should be done right before release, so filed another issue for this enhancement: #6062

fixes: #5940


emlin:
> NAK, we need the property collector to make sure the process really starts.
> 
> As a workaround we should either make the timeout more than 3 min or remove them altogether.

@caglar10ur Did I misunderstand anything? I didn't change the property collector timeout. Just I didn't change the attach timeout to 3m as well, for attach will hold the docker cli, without any response.

caglar10ur:
You removed the check from lib/portlayer/exec/base.go https://github.com/vmware/vic/pull/6063/files#diff-67f57bbb6bec9888ba685e947a8a00cfL219

emlin:
> You removed the check from lib/portlayer/exec/base.go https://github.com/vmware/vic/pull/6063/files#diff-67f57bbb6bec9888ba685e947a8a00cfL219

this is the second change, to avoid setting container back to stopped, if only session is not initialized.
the wait for session method is called in https://github.com/emlin/vic/blob/a3420f7c0a10465695ced9cc14308429a297d446/lib/portlayer/exec/container.go#L385.
If I don't remove this line, the container will be set back to stopped, if session hung there to wait for clear to launch channel, because of this line: https://github.com/emlin/vic/blob/a3420f7c0a10465695ced9cc14308429a297d446/lib/portlayer/exec/container.go#L366

caglar10ur:
Sooooo looked at the code a bit more and I think the behavior that you are describing is what we want and correct. When you start a container with -i, we set runblock flag of a session (this is exec session not ssh) to true and won't start the process unless the attach operation succeeds. We do that in order to not lost any output from the command that will run inside the container. 

I understand the timeout in attach path ends up timing out waitForSession. The issue is we can't tell whether it is really an attach timeout due to slowness of the system, a real ssh problem or something happened on the container side (tether crashed, vm got stuck etc.)

Also the same start function is used for detached operations so that -d can wait the process to start and then return to the caller, removing that will cause -d to return immediately without waiting command to start.

I believe - unless we do overall change in design - the only thing we can do here is making both ssh and property collector timeouts the same value and increasing them a bit to accommodate slow systems (2 min or something). This way fast systems will keep running as of today and slow systems will waiting more.

caglar10ur:
@emlin pointed out that we are waiting the session twice in the start path (once in base.Start and then container.Start) so I'm OK for this workaround until we fix it for real :)

emlin:
To fix it for real, issue #6072 is filed

PR #5483, opened 2017-06-15
[specific ci=1-27-Docker-Login] Fail quickly on docker login against harbor registry with incorrect credentials

willsherwood:
Fixes #5195

PR #6067, opened 2017-08-18
[specific ci=1-02-Docker-Pull] Grab out error message body for nonretryable client errors and all other unexpected http code

### LINKED ISSUE #5951: VIC engine should render the error message in a similar way as docker does
This is a suggestion.

Harbor will return 412 when the image violates the "project level content trust" and "vulnerable" policy.

The response body is a json:
```
{"code":412,"message":"The image is not signed in Notary.","details":"The image is not signed in Notary."}
```
The response json is following the convention of docker, so when user issues `docker pull` a more meaningful error message will be returned:
```
Error response from daemon: unknown: The image is not signed in Notary.
```

It would be better if VIC engine can extract error message from the response to show more information to user.

AngieCris:
Fixes #5951

### DIFF
--- a/pkg/fetcher/fetcher.go
+++ b/pkg/fetcher/fetcher.go
@@ -444,6 +444,8 @@ func (u *URLFetcher) buildRegistryErrMsg(url *url.URL, respBody io.ReadCloser) s
 
 // extractErrResponseMessage extracts `message` field from error response body stream.
 func extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
+	defer rdr.Close()
+
 	out := bytes.NewBuffer(nil)
 	_, err := io.Copy(out, rdr)
 	if err != nil {
--- a/pkg/fetcher/fetcher.go
+++ b/pkg/fetcher/fetcher.go
@@ -249,6 +249,7 @@ func (u *URLFetcher) fetch(ctx context.Context, url *url.URL, reqHdrs *http.Head
 	u.StatusCode = res.StatusCode
 
 	if u.IsNonretryableClientError() {
+
 		if u.options.Token == nil && u.IsStatusUnauthorized() {
 			hdr := res.Header.Get("www-authenticate")
 			if hdr == "" {
@@ -280,13 +281,30 @@ func (u *URLFetcher) fetch(ctx context.Context, url *url.URL, reqHdrs *http.Head
 			}
 		}
 
-		err = fmt.Errorf("Nonretryable error '%s (%d)': URL %s.", http.StatusText(u.StatusCode), u.StatusCode, url)
+		// for all other non-retryable client error, grab the error message if there is one (#5951)
+		err, msg := u.extractMsgBody(res.Body)
+		if err != nil {
+			err = fmt.Errorf("Nonretryable error '%s (%d)': URL %s", http.StatusText(u.StatusCode), u.StatusCode, url)
+		} else {
+			err = fmt.Errorf("Nonretryable error '%s (%d)': URL %s, Message: %s", http.StatusText(u.StatusCode),
+				u.StatusCode, url, msg)
+		}
+
 		return nil, nil, DoNotRetry{Err: err}
 	}
 
 	// FIXME: handle StatusTemporaryRedirect and StatusFound
+	// for all other unexpected http code, grab the message out if there is one (#5951)
 	if !u.IsStatusOK() {
-		return nil, nil, fmt.Errorf("Unexpected http code: %d, URL: %s", u.StatusCode, url)
+
+		err, msg := u.extractMsgBody(res.Body)
+		if err != nil {
+			err = fmt.Errorf("Unexpected http code: %d, URL %s", u.StatusCode, url)
+		} else {
+			err = fmt.Errorf("Unexpected http code: %d, URL %s, Message: %s", u.StatusCode, url, msg)
+		}
+
+		return nil, nil, err
 	}
 
 	log.Debugf("URLFetcher.fetch() - %#v, %#v", res.Body, res.Header)
@@ -412,6 +430,16 @@ func (u *URLFetcher) IsNonretryableClientError() bool {
 	s := u.StatusCode
 	return 400 <= s && s < 500 &&
 		s != http.StatusLocked && s != http.StatusTooManyRequests
+}
+
+func (u *URLFetcher) extractMsgBody(rdr io.ReadCloser) (error, string) {
+	out := bytes.NewBuffer(nil)
+	_, err := io.Copy(out, rdr)
+	if err != nil {
+		return err, ""
+	}
+	return nil, string(out.Bytes())
+
 }
 
 func (u *URLFetcher) setUserAgent(req *http.Request) {
--- a/pkg/fetcher/fetcher.go
+++ b/pkg/fetcher/fetcher.go
@@ -442,8 +442,13 @@ func (u *URLFetcher) buildRegistryErrMsg(url *url.URL, respBody io.ReadCloser) s
 	return errMsg
 }
 
+// malformedJsonErrFormat is the error format for malformed json response body
+// used in function extractErrResponseMessage
+var malformedJsonErrFormat = fmt.Errorf("error response json has unconventional format")
+
 // extractErrResponseMessage extracts `message` field from error response body stream.
 func extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
+	// close the stream after done
 	defer rdr.Close()
 
 	out := bytes.NewBuffer(nil)
@@ -465,7 +470,7 @@ func extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
 
 	if len(errResponse.Errors) == 0 {
 		log.Debugf("Error response wrong format. Response body: %s", string(res))
-		return "", fmt.Errorf("error response json has unconventional format")
+		return "", malformedJsonErrFormat
 	}
 
 	// grab out every error message
@@ -483,7 +488,7 @@ func extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
 
 	// if no message available, treat it as a malformed json error
 	if len(errString) == 0 {
-		return "", fmt.Errorf("error response json has unconventional format")
+		return "", malformedJsonErrFormat
 	}
 
 	return errString, nil
--- a/pkg/fetcher/fetcher.go
+++ b/pkg/fetcher/fetcher.go
@@ -444,7 +444,7 @@ func (u *URLFetcher) buildRegistryErrMsg(url *url.URL, respBody io.ReadCloser) s
 
 // malformedJsonErrFormat is the error format for malformed json response body
 // used in function extractErrResponseMessage
-var malformedJsonErrFormat = fmt.Errorf("error response json has unconventional format")
+var errJsonFormat = fmt.Errorf("error response json has unconventional format")
 
 // extractErrResponseMessage extracts `message` field from error response body stream.
 func extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
@@ -470,7 +470,7 @@ func extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
 
 	if len(errResponse.Errors) == 0 {
 		log.Debugf("Error response wrong format. Response body: %s", string(res))
-		return "", malformedJsonErrFormat
+		return "", errJsonFormat
 	}
 
 	// grab out every error message
@@ -488,7 +488,7 @@ func extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
 
 	// if no message available, treat it as a malformed json error
 	if len(errString) == 0 {
-		return "", malformedJsonErrFormat
+		return "", errJsonFormat
 	}
 
 	return errString, nil
--- a/pkg/fetcher/fetcher.go
+++ b/pkg/fetcher/fetcher.go
@@ -444,7 +444,7 @@ func (u *URLFetcher) buildRegistryErrMsg(url *url.URL, respBody io.ReadCloser) s
 
 // malformedJsonErrFormat is the error format for malformed json response body
 // used in function extractErrResponseMessage
-var errJsonFormat = fmt.Errorf("error response json has unconventional format")
+var errJSONFormat = fmt.Errorf("error response json has unconventional format")
 
 // extractErrResponseMessage extracts `message` field from error response body stream.
 func extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
@@ -470,7 +470,7 @@ func extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
 
 	if len(errResponse.Errors) == 0 {
 		log.Debugf("Error response wrong format. Response body: %s", string(res))
-		return "", errJsonFormat
+		return "", errJSONFormat
 	}
 
 	// grab out every error message
@@ -488,7 +488,7 @@ func extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
 
 	// if no message available, treat it as a malformed json error
 	if len(errString) == 0 {
-		return "", errJsonFormat
+		return "", errJSONFormat
 	}
 
 	return errString, nil
--- a/pkg/fetcher/fetcher.go
+++ b/pkg/fetcher/fetcher.go
@@ -100,10 +100,9 @@ type URLFetcher struct {
 	options Options
 }
 
-// RegistryErrorRespBody struct
-// struct for unmarshalling json error response body by containers registry
-// error response json is assumed to follow Docker API convention (field `details` is dropped)
-// see: https://docs.docker.com/registry/spec/api/#errors
+// RegistryErrorRespBody is used for unmarshaling json error response body from image registries.
+// Error response json is assumed to follow Docker API convention (field `details` is dropped).
+// See: https://docs.docker.com/registry/spec/api/#errors
 type RegistryErrorRespBody struct {
 	Errors []struct {
 		Code    string
@@ -298,7 +297,7 @@ func (u *URLFetcher) fetch(ctx context.Context, url *url.URL, reqHdrs *http.Head
 	}
 
 	// FIXME: handle StatusTemporaryRedirect and StatusFound
-	// for all other unexpected http code, grab the message out if there is one (#5951)
+	// for all other unexpected http codes, grab the message out if there is one (#5951)
 	if !u.IsStatusOK() {
 		err := fmt.Errorf(u.buildRegistryErrMsg(u.StatusCode, url, res.Body))
 		return nil, nil, err
@@ -429,8 +428,8 @@ func (u *URLFetcher) IsNonretryableClientError() bool {
 		s != http.StatusLocked && s != http.StatusTooManyRequests
 }
 
-// Build error message for unexpected http code (nonretryable client errors and all other errors)
-// extract message details from response body stream if there is one
+// buildRegistryErrMsg builds error message for unexpected http code (nonretryable client errors and all other errors)
+// and extracts message details from response body stream if there is one (#5951).
 func (u *URLFetcher) buildRegistryErrMsg(httpStatusCode int, url *url.URL, respBody io.ReadCloser) string {
 	errMsg := fmt.Sprintf("Unexpected http code: %d (%s), URL: %s", httpStatusCode, http.StatusText(httpStatusCode), url)
 
@@ -440,12 +439,10 @@ func (u *URLFetcher) buildRegistryErrMsg(httpStatusCode int, url *url.URL, respB
 	}
 
 	errMsg += fmt.Sprintf(", Message: %s", errDetail)
-
 	return errMsg
 }
 
-// Extract message from error response body (#5951)
-// The error response body is assumed to follow the convention of Docker
+// extractErrResponseMessage extracts `message` field from error response body json (#5951).
 func (u *URLFetcher) extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
 	out := bytes.NewBuffer(nil)
 	_, err := io.Copy(out, rdr)
--- a/pkg/fetcher/fetcher.go
+++ b/pkg/fetcher/fetcher.go
@@ -100,18 +100,14 @@ type URLFetcher struct {
 	options Options
 }
 
-// ErrorResponseBody struct
-type ErrorResponseBody struct {
+// RegistryErrorRespBody struct
+// struct for unmarshalling json error response body by containers registry
+// error response json is assumed to follow Docker API convention (field `details` is dropped)
+// see: https://docs.docker.com/registry/spec/api/#errors
+type RegistryErrorRespBody struct {
 	Errors []struct {
-		Code string
+		Code    string
 		Message string
-
-		Detail []struct {
-			Type string
-			Class string
-			Name string
-			Action string
-		}
 	}
 }
 
@@ -264,7 +260,6 @@ func (u *URLFetcher) fetch(ctx context.Context, url *url.URL, reqHdrs *http.Head
 	u.StatusCode = res.StatusCode
 
 	if u.IsNonretryableClientError() {
-
 		if u.options.Token == nil && u.IsStatusUnauthorized() {
 			hdr := res.Header.Get("www-authenticate")
 			if hdr == "" {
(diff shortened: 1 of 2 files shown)


AngieCris:
Now the error message for all other nonretryable client error and unexpected http code, should contain the error response body.
The response body should be a json, but since it could come from either docker, harbor or gcp, the structure of it is undetermined.
Currently in the code, I just converted the response body bytes to string directly, instead of unmarshalling it to json, due to the reason described above.

AngieCris:
Pending an integration test. 
Need to check if there is a non-generic error message upon unexpected http code (sth other than NotFound and Unauthorized, like 412)
It takes some effort to reproduce errors like 412 - need to configure Harbor and Admiral, and I'm not sure if Admiral provides API for accessing Harbor instances (as mentioned in https://github.com/vmware/vic/pull/5483)

pkg/fetcher/fetcher.go
```diff
@@ -280,13 +281,30 @@ func (u *URLFetcher) fetch(ctx context.Context, url *url.URL, reqHdrs *http.Head
 			}
 		}
 
-		err = fmt.Errorf("Nonretryable error '%s (%d)': URL %s.", http.StatusText(u.StatusCode), u.StatusCode, url)
+		// for all other non-retryable client error, grab the error message if there is one (#5951)
+		msg, err := u.extractMsgBody(res.Body)
+		if err != nil {
```
chengwang86: I suggest logging the err first before overwriting it:
```
if err != nil {
    log.Errorf("....", err)
    err = fmt.Errorf("Nonretryabe error ...")
} else ...
```

AngieCris:
@chengwang86 I've confirmed that as for harbor and docker, they follow the same structure. Not sure about gcp. 
Right now, the code just outputs the whole response body - converts the whole json object into string. Just for the purpose of demonstration, for example an Unauthorized will look like this on the client side:
`Image not found. Message:  {“errors”:[{“code”:“UNAUTHORIZED”,“message”:“authentication required”,“detail”:[{“Type”:“repository”,“Class”:“”,“Name”:“library/jiowengew”,“Action”:“pull”}]}]}`
which does not look very good...

I'm thinking about unmarshalling the json according to the docker convention, and extract the error message body. 
If the marshal failed (because the error message does not follow the convention), we simply just go with the generic error message ("unexpected http code" or "nonretryable error").

Better suggestions, anyone?

chengwang86:
For gcr, we have a test for docker pull here https://github.com/vmware/vic/blob/master/tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot#L150

You can run this test manually to see whether it follows the same pattern.

AngieCris:
now it should grab the message field out from the error response body. 
The json convention is assumed to be:
`{
    "errors:" [{
            "code": <error identifier>,
            "message": <message describing condition>,
            "detail": <unstructured>
        },
        ...
    ]
}`

Tested on status Unauthorized (just for the testing purpose. I haven't figured out how to reproduce a 412 yet), it works
Harbor should follow the same convention (verified from @reasonerjt)
I'm verifying that it works for GCP

pkg/fetcher/fetcher.go
```diff
@@ -100,6 +100,21 @@ type URLFetcher struct {
 	options Options
 }
 
+// ErrorResponseBody struct
```
anchal-agrawal: Please expand upon this comment. Also, `PullErrorRespBody` would be a more suitable name for this struct.

pkg/fetcher/fetcher.go
```diff
@@ -100,6 +100,21 @@ type URLFetcher struct {
...
+type ErrorResponseBody struct {
+	Errors []struct {
+		Code string
+		Message string
+
+		Detail []struct {
```
chengwang86: Is `Detail` always a struct like this? Is it possible that it is just a string? From the documentation, I don't see such definition.

pkg/fetcher/fetcher.go
```diff
@@ -100,6 +100,21 @@ type URLFetcher struct {
 	options Options
 }
 
+// ErrorResponseBody struct
+type ErrorResponseBody struct {
+	Errors []struct {
+		Code string
```
anchal-agrawal: Formatting: this might need some `gofmt`.

pkg/fetcher/fetcher.go
```diff
@@ -249,6 +264,7 @@ func (u *URLFetcher) fetch(ctx context.Context, url *url.URL, reqHdrs *http.Head
 	u.StatusCode = res.StatusCode
 
 	if u.IsNonretryableClientError() {
+
```
anchal-agrawal: Unneeded whitespace.

pkg/fetcher/fetcher.go
```diff
@@ -280,13 +296,32 @@ func (u *URLFetcher) fetch(ctx context.Context, url *url.URL, reqHdrs *http.Head
 			}
 		}
 
-		err = fmt.Errorf("Nonretryable error '%s (%d)': URL %s.", http.StatusText(u.StatusCode), u.StatusCode, url)
+		// for all other non-retryable client error, grab the error message if there is one (#5951)
```
anchal-agrawal: client error -> client errors

pkg/fetcher/fetcher.go
```diff
@@ -280,13 +296,32 @@ func (u *URLFetcher) fetch(ctx context.Context, url *url.URL, reqHdrs *http.Head
...
 	}
 
 	// FIXME: handle StatusTemporaryRedirect and StatusFound
+	// for all other unexpected http code, grab the message out if there is one (#5951)
 	if !u.IsStatusOK() {
-		return nil, nil, fmt.Errorf("Unexpected http code: %d, URL: %s", u.StatusCode, url)
+
+		msg, err := u.extractErrResponseMessage(res.Body)
+		if err != nil {
+			log.Debugf("Unexpected http code error in unconventional format: %s", err)
```
chengwang86: this is not needed given that you have already added the debug log in `extractErrResponseMessage`

pkg/fetcher/fetcher.go
```diff
@@ -414,6 +449,44 @@ func (u *URLFetcher) IsNonretryableClientError() bool {
...
+	log.Debugf("Error message json string: %s", string(res))
+
+	var errResponse ErrorResponseBody
+	err = json.Unmarshal(res, &errResponse)
+	if err != nil {
+		log.Debugf("Error when unmarshalling error response body: %s", err)
+		return "", err
+	}
+
+	if errResponse.Errors == nil || len(errResponse.Errors) == 0 {
```
anchal-agrawal: We can simply say `if len(errResponse.Errors) == 0 {` here - the other check is redundant.

pkg/fetcher/fetcher.go
```diff
@@ -414,6 +449,44 @@ func (u *URLFetcher) IsNonretryableClientError() bool {
...
+	var errString string
+	for i := 0; i < len(errResponse.Errors); i++ {
+		errString += errResponse.Errors[i].Message
+		if i != len(errResponse.Errors) - 1 {
+			errString += ", "
+		}
+	}
+
+	return errString, nil
+
```
anchal-agrawal: Unneeded whitespace.

pkg/fetcher/fetcher.go
AngieCris: My bad - rechecked the documentation, `Detail` is actually unstructured. I could drop the `Detail` field for now (since only `Message` field is used, and pattern of `Detail` is undetermined)

pkg/fetcher/fetcher.go
AngieCris: Does `PullErrorRespBody` mean error response for pull request? It is not necessarily for pull request only (convention: https://docs.docker.com/registry/spec/api/#errors) tho it could be that it's only used when pulling from registry tho

anchal-agrawal: Good point - then we can call it `RegistryErrorRespBody` or leave it as-is. Your call!

pkg/fetcher/fetcher.go
AngieCris: !! forgot to configure `gofmt` in Gogland

pkg/fetcher/fetcher.go
```diff
@@ -414,6 +449,44 @@ func (u *URLFetcher) IsNonretryableClientError() bool {
...
+		return "", err
+	}
+
+	if errResponse.Errors == nil || len(errResponse.Errors) == 0 {
+		log.Debugf("Error response wrong format. Response body: %s", string(res))
+		return "", fmt.Errorf("Error response wrong format")
+	}
+
+	// grab out every error message
+	var errString string
```
sflxn: This can be made simpler and cleaner with

```
for i, error := range errResponse.Errors {
  if i > 0 {
    errString += ", "
  }
  errString += error.Message
}
```

They're equivalent, but I find for range cleaner than traditional C style for loops.

AngieCris: ohhh that looks a lot nicer. thanx!

pkg/fetcher/fetcher.go
```diff
@@ -280,13 +296,32 @@ func (u *URLFetcher) fetch(ctx context.Context, url *url.URL, reqHdrs *http.Head
 			}
 		}
 
-		err = fmt.Errorf("Nonretryable error '%s (%d)': URL %s.", http.StatusText(u.StatusCode), u.StatusCode, url)
+		// for all other non-retryable client error, grab the error message if there is one (#5951)
+		msg, err := u.extractErrResponseMessage(res.Body)
+		if err != nil {
+			log.Debugf("Nonretryable client error in unconventional format: %s", err)
+			err = fmt.Errorf("Nonretryable error '%s (%d)': URL %s", http.StatusText(u.StatusCode), u.StatusCode, url)
```
sflxn: I think this is too verbose.  Users don't like to see to information overload.  If we're returning error, we don't really need to say 'nonretryable error'.  It's redundant, and they have no idea what that means.  It's fine for our logs.

pkg/fetcher/fetcher.go
```diff
@@ -280,13 +296,32 @@ func (u *URLFetcher) fetch(ctx context.Context, url *url.URL, reqHdrs *http.Head
...
+		} else {
+			err = fmt.Errorf("Nonretryable error '%s (%d)': URL %s, Message: %s", http.StatusText(u.StatusCode),
```
sflxn: same comment about verbosity here.

pkg/fetcher/fetcher.go
```diff
@@ -280,13 +296,32 @@ func (u *URLFetcher) fetch(ctx context.Context, url *url.URL, reqHdrs *http.Head
...
+
 		return nil, nil, DoNotRetry{Err: err}
 	}
 
 	// FIXME: handle StatusTemporaryRedirect and StatusFound
+	// for all other unexpected http code, grab the message out if there is one (#5951)
 	if !u.IsStatusOK() {
-		return nil, nil, fmt.Errorf("Unexpected http code: %d, URL: %s", u.StatusCode, url)
+
+		msg, err := u.extractErrResponseMessage(res.Body)
```
sflxn: Since you're repeating the error messaging above, we should pull it out into another function.

pkg/fetcher/fetcher.go
AngieCris: yes, you're right. one thing i'm not sure about: the `Nonretryable client error` is in the 4xx http code category, and we are treating it differently from all other unexpected http error code. Do we need to give user this info in some way or we can just say `Unexpected http code` for all error codes?

pkg/fetcher/fetcher.go
AngieCris: yes but before i worked on this issue, we're differentiating nonretryable error (4xx http code) from all other errors (5xx or sth else), and we're outputing messages differently, so that's why i did the same thing...
if we don't want the user to get the verbose `nonretryable error` message, i can just squash this condition and put it together with everything else

AngieCris:
addressed the comments (thanx everyone), now should be good for review
@chengwang86 @anchal-agrawal @sflxn

pkg/fetcher/fetcher.go
```diff
@@ -414,6 +429,58 @@ func (u *URLFetcher) IsNonretryableClientError() bool {
...
+func (u *URLFetcher) buildRegistryErrMsg(httpStatusCode int, url *url.URL, respBody io.ReadCloser) string {
+	errMsg := fmt.Sprintf("Unexpected http code: %d (%s), URL: %s", httpStatusCode, http.StatusText(httpStatusCode), url)
+
+	errDetail, err := u.extractErrResponseMessage(respBody)
+	if err != nil {
+		return errMsg
+	}
+
+	errMsg += fmt.Sprintf(", Message: %s", errDetail)
+
```
anchal-agrawal: Unneeded whitespace.

pkg/fetcher/fetcher.go
```diff
@@ -100,6 +100,17 @@ type URLFetcher struct {
 	options Options
 }
 
+// RegistryErrorRespBody struct
```
anchal-agrawal: Minor edits - this comment block should say

```
// RegistryErrorRespBody is used for unmarshaling json error response body from image registries.
// Error response json is assumed to follow Docker API convention (field `details` is dropped).
// See: https://docs.docker.com/registry/spec/api/#errors
```

pkg/fetcher/fetcher.go
```diff
@@ -414,6 +429,58 @@ func (u *URLFetcher) IsNonretryableClientError() bool {
...
+		return errMsg
+	}
+
+	errMsg += fmt.Sprintf(", Message: %s", errDetail)
+
+	return errMsg
+}
+
+// Extract message from error response body (#5951)
+// The error response body is assumed to follow the convention of Docker
```
anchal-agrawal: `The error response body is assumed to follow the convention of Docker` is redundant since the comment for `RegistryErrorRespBody` includes this.

pkg/fetcher/fetcher.go
```diff
@@ -414,6 +429,58 @@ func (u *URLFetcher) IsNonretryableClientError() bool {
 		s != http.StatusLocked && s != http.StatusTooManyRequests
 }
 
+// Build error message for unexpected http code (nonretryable client errors and all other errors)
```
anchal-agrawal: Godoc comments for functions should start with the function name. See https://golang.org/doc/effective_go.html#commentary.

pkg/fetcher/fetcher.go
```diff
@@ -414,6 +429,58 @@ func (u *URLFetcher) IsNonretryableClientError() bool {
...
+	if err != nil {
+		return errMsg
+	}
+
+	errMsg += fmt.Sprintf(", Message: %s", errDetail)
+
+	return errMsg
+}
+
+// Extract message from error response body (#5951)
```
anchal-agrawal: Ditto re: godoc comment for functions.

pkg/fetcher/fetcher.go
```diff
@@ -414,6 +429,58 @@ func (u *URLFetcher) IsNonretryableClientError() bool {
...
+	var errResponse RegistryErrorRespBody
+	err = json.Unmarshal(res, &errResponse)
+	if err != nil {
+		log.Debugf("Error when unmarshalling error response body: %s", err)
+		return "", err
+	}
+
+	if len(errResponse.Errors) == 0 {
+		log.Debugf("Error response wrong format. Response body: %s", string(res))
+		return "", fmt.Errorf("Error response wrong format")
```
anchal-agrawal: Error strings should start with lowercase - see https://github.com/golang/go/wiki/CodeReviewComments#error-strings.

AngieCris: error strings lowercase, debug strings uppercase?

pkg/fetcher/fetcher.go
```diff
@@ -414,6 +428,56 @@ func (u *URLFetcher) IsNonretryableClientError() bool {
...
+	}
+
+	if len(errResponse.Errors) == 0 {
+		log.Debugf("Error response wrong format. Response body: %s", string(res))
+		return "", fmt.Errorf("error response json has unconventional format")
+	}
+
+	// grab out every error message
+	var errString string
+	for i, error := range errResponse.Errors {
```
anchal-agrawal: This loop can be made more efficient by avoiding a copy of each element of `errResponse.Errors` at every iteration, like so:

```go
	for i := range errResponse.Errors {
		if i > 0 {
			errString += ", "
		}
		errString += errResponse.Errors[i].Message
	}
```

For more details, read the Golang spec for `for` statements: https://golang.org/ref/spec#For_statements

AngieCris: ahhh thanx good point `errResponse.Errors` could be pretty large due to the arbitrary `details` field and we don't need that

pkg/fetcher/fetcher.go
```diff
@@ -414,6 +428,56 @@ func (u *URLFetcher) IsNonretryableClientError() bool {
...
+	if err != nil {
+		return errMsg
+	}
+
+	errMsg += fmt.Sprintf(", Message: %s", errDetail)
+	return errMsg
+}
+
+// extractErrResponseMessage extracts `message` field from error response body json (#5951).
+func (u *URLFetcher) extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
```
anchal-agrawal: This function isn't using any fields of the `URLFetcher` receiver - should we remove the function receiver? Pinging @sflxn for input.

AngieCris: it's essentially just a helper for `buildRegistryErrMsg` for clearer logic

anchal-agrawal: Yeah, we should remove the `URLFetcher` receiver.

anchal-agrawal: Bonus: doing this would make it easier to unit test this function, which we should. You can write a unit test in a new file, `pkg/fetcher/fetcher_test.go`.

anchal-agrawal:
Please also add a `[specific ci=1-02-Docker-Pull]` to your PR title so we can test the `docker pull` suite against your changes. https://github.com/vmware/vic/blob/master/tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot

AngieCris:
@anchal-agrawal can you take a look at the unit test i just added? thanx

pkg/fetcher/fetcher_test.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	"bytes"
+	"io/ioutil"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	ErrJsonStr401 = `{
```
anchal-agrawal: These constants don't need to be exported, i.e. they should start with lowercase.

pkg/fetcher/fetcher_test.go
```diff
@@ -0,0 +1,110 @@
...
+		  "message":"authentication required",
+		  "detail":[{"Type":"repository","Class":"","Name":"library/jiowengew","Action":"pull"}]
+		},
+		{"code":"NOTFOUND",
+		 "message": "image not found",
+		 "detail": "image not found"
+		}]
+	}
+	`
+	RandomStr                    = `random`
```
anchal-agrawal: Nit: this isn't `Random` per-se, `UnexpectedStr` is more suitable.

pkg/fetcher/fetcher_test.go
```diff
@@ -0,0 +1,110 @@
...
+	RandomJsonStr                = `{"nope":"nope"}`
```
anchal-agrawal: `UnexpectedJsonStr ` would be better.

pkg/fetcher/fetcher.go
```diff
@@ -414,6 +428,67 @@ func (u *URLFetcher) IsNonretryableClientError() bool {
...
+	var errResponse RegistryErrorRespBody
+	err = json.Unmarshal(res, &errResponse)
+	if err != nil {
+		log.Debugf("Error when unmarshaling error response body: %s", err)
+		return "", err
+	}
+
+	if len(errResponse.Errors) == 0 {
+		log.Debugf("Error response wrong format. Response body: %s", string(res))
+		return "", fmt.Errorf("error response json has unconventional format")
```
anchal-agrawal: After reviewing the unit test: since `fmt.Errorf("error response json has unconventional format")` is used in multiple places, you can define a variable in this file like so: `var errJsonFormat = fmt.Errorf("error response json has unconventional format")` and use it here and in the unit test.

pkg/fetcher/fetcher_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// Test 3: random string in the stream that is not a json
+	msg, err = extractErrResponseMessage(randomStrTestStream)
+	assert.Equal(t, "", msg, "test: (non-json string) no message should be extracted")
+	assert.NotNil(t, err, "test: (non-json string) extraction should fail")
+
+	// Test 4: malformed json string
+	msg, err = extractErrResponseMessage(malformedJSONTestStream)
+	assert.Equal(t, "", msg, "test: (malformed json string) no message should be extracted")
+	assert.NotNil(t, err, "test: (malformed json string) extraction should fail")
```
anchal-agrawal: After making this change (https://github.com/vmware/vic/pull/6067/files#r134591535), we can just `assert.Equal(t, err, errJsonFormat)` instead of the two asserts here.

anchal-agrawal: Same for the other test cases below.

AngieCris: will `err.Errors()` fail if `err == nil`? so i was checking if `err` is nil first

anchal-agrawal: The `err.Error()` assert won't be necessary - we can directly check the error for equality once the above changes are made.

AngieCris: got it, thanx

pkg/fetcher/fetcher.go
AngieCris: good point. global variable (const) in this file?

anchal-agrawal: It'd be a `var` in this file and can be added just above this function.

AngieCris: got it

pkg/fetcher/fetcher.go
```diff
@@ -414,6 +428,72 @@ func (u *URLFetcher) IsNonretryableClientError() bool {
...
+		return errMsg
+	}
+
+	errMsg += fmt.Sprintf(", Message: %s", errDetail)
+	return errMsg
+}
+
+// malformedJsonErrFormat is the error format for malformed json response body
+// used in function extractErrResponseMessage
+var malformedJsonErrFormat = fmt.Errorf("error response json has unconventional format")
```
anchal-agrawal: The idiom for declaring error variables is to prefix them with `err`. So `errJsonFormat` should be better here.

pkg/fetcher/fetcher_test.go
```diff
@@ -0,0 +1,103 @@
...
+	// Test 2: multiple error message extraction
+	msg, err = extractErrResponseMessage(multipleErrTestStream)
+	assert.Nil(t, err, "test: (multiple error messages) extraction should success for well-formatted error json")
+	assert.Equal(t, "authentication required, image not found", msg,
+		"test: (multiple error messages) extracted message: %s; expected: authentication required, image not found", msg)
+
+	// Test 3: random string in the stream that is not a json
+	msg, err = extractErrResponseMessage(unexpectedStrTestStream)
+	assert.Equal(t, "", msg, "test: (non-json string) no message should be extracted")
+	assert.NotNil(t, err, "test: (non-json string) extraction should fail")
```
anchal-agrawal: Instead of `assert.NotNil(t, err, "test: (non-json string) extraction should fail")`, we should have something like `assert.Equal(t, err, errJsonFormat, "test: (non-json string) extraction should fail")`.

AngieCris: this error message comes from a `json.unmarshal` fail, and i'm not sure about the message...let me try on go playground

pkg/fetcher/fetcher_test.go
```diff
@@ -0,0 +1,103 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fetcher
+
+import (
+	"bytes"
+	"github.com/stretchr/testify/assert"
```
anchal-agrawal: goimports: `"github.com/stretchr/testify/assert"` is a third-party pkg and should be in a separate group below the standard pkgs.

pkg/fetcher/fetcher_test.go
anchal-agrawal: My bad - this is fine as-is.

pkg/fetcher/fetcher_test.go
AngieCris: ohhhhh got it i was confused why goimports failed me

pkg/fetcher/fetcher_test.go
AngieCris: :)

AngieCris:
Final comment before closing this pr:
there could be potential problem upon error response when pulling from `gcr.io`
just as this problem scenario, when pulling the manifest, upon unexpected http code there is no meaningful message
but pulling from `gcr.io` does not go thru the error rendering routine in the `fetcher()` function in `pkg/fetcher`, it goes thru some other routine in `lib/imagec/docker.go` (i didn't get enough time to figure out exactly how does it work) where message detail is dropped again, giving client only generic error message (sth like `unexpected http response`)
this could come out as an issue later...maybe or maybe not...just put it here for (maybe) future reference

PR #6068, opened 2017-08-18
[specific ci=1-17-Docker-Network-Connect] Prioritize internal networks in /etc/resolv.conf

### LINKED ISSUE #5738: Ordering in /etc/resolve.conf for container+bridge networks is inconsistent across container starts
**User Statement:**

As a container developer, I need to know that if I attach a container to multiple networks, the containers will get consistent network configuration each time.

**Details:**

I've been trying to debug an issue in a compose script where I see intermittent success and failure. I've narrowed it down to the fact that the ordering in /etc/resolve.conf doesn't seem to be consistent across container starts. If it's configured in one way, it works. Another way and it doesn't. There is no workaround that I can find.

Here's what I did (note that --internal makes no difference on the network created):

```
docker network create --internal mynet
docker run -d --name foo --net mynet bensdoings/ifconfig sleep 1000
docker create -it --name bar --net ExternalNetwork -p 80 bensdoings/ifconfig /bin/bash
docker network connect mynet bar
docker start -ai bar
$ ping foo
$ exit
docker start -ai bar
$ ping foo
$ exit
...
```
I would expect to be able to ping foo by name each time. However, /etc/resolve.conf seems to get configured in two different ways, seemingly at random.

It works in the case where resolve.conf looks like this:
```
nameserver 172.21.0.1
nameserver 10.118.65.1
nameserver 10.118.65.2
options timeout:15
options attempts:5
```
It does not work if resolve.conf looks like this:
```
nameserver 10.118.65.1
nameserver 10.118.65.2
nameserver 172.21.0.1
options timeout:15
options attempts:5
```

**Acceptance Criteria:**

This particular kind of network setup is one we need to document as the classic container + internal network scenario. It needs to work consistently as this should form the best practice basis of any deployment that needs to be externally accessible.

As such, this should be high priority to fix.

Note also there is no test for the container network + internal network scenario in https://github.com/vmware/vic/blob/master/tests/test-cases/Group1-Docker-Commands/1-17-Docker-Network-Connect.robot#L55. We really need to add one

gigawhitlocks:
Adds a boolean to let us know if a networkendpoint is for an internal network & uses this additional metadata to prioritize the ordering of /etc/resolv.conf in order to resolve #5738 

I've been testing like so:

```bash
 /home/ian/go/src/github.com/vmware/vic/bin/vic-machine-linux delete --target $(testbed-ip) --user Administrator@vsphere.local  --password xxx --name ian-vch --thumbprint=$(testbed-thumbprint) --force \
      && /home/ian/go/src/github.com/vmware/vic/bin/vic-machine-linux create --target $(testbed-ip)/vcqaDC --user Administrator@vsphere.local  --password xxx --name ian-vch --image-store=sharedVmfs-0 --tls-cname="*.eng.vmware.com" --bridge-network=bridge  --thumbprint=$(testbed-thumbprint) --debug=2 --timeout=15m \
      && vic-env \
      && docker network create --internal mynet \
      && docker run -d --name foo --net mynet bensdoings/ifconfig sleep 1000 \
      && docker create -it --name bar --net bridge -p 80 bensdoings/ifconfig ping -c3 foo \
      && docker network connect mynet bar \
      && docker start -ai bar
```

(for reference, `vic-env` is just the following to set docker env vars, nothing fancy)

```bash
vic-env () {
	vch="$@" 
	[[ $vch = "" ]] && vch=$vch_default_name 
	for x in $(cat $GOPATH/src/github.com/vmware/vic/"$vch"/"$vch".env)
	do
		export $x
	done
	docker info
}
```

Ran this successfully 5 times in a row; considering /etc/resolv.conf has 2 nameservers in this scenario that gives us a 1/2 chance of getting the ordering right randomly (as was happening before) on each run, so the chance of getting it right by chance 5 times in a row is 1/32 or 3.125% chance... I'd say it's fixed. Still need to write a test, which I will add to the PR over the weekend (tomorrow probably).

### DIFF
--- a/lib/tether/tether.go
+++ b/lib/tether/tether.go
@@ -251,8 +251,18 @@ func (t *tether) setHostname() error {
 }
 
 func (t *tether) setNetworks() error {
+	var networks []*NetworkEndpoint
+	// prioritize internal networks
 	for _, v := range t.config.Networks {
-		if err := t.ops.Apply(v); err != nil {
+		if !v.Internal {
+			networks = append(networks, v)
+		} else {
+			networks = append([]*NetworkEndpoint{v}, networks...)
+		}
+	}
+
+	for _, network := range networks {
+		if err := t.ops.Apply(network); err != nil {
 			return fmt.Errorf("failed to apply network endpoint config: %s", err)
 		}
 	}
--- a/lib/tether/tether.go
+++ b/lib/tether/tether.go
@@ -254,10 +254,10 @@ func (t *tether) setNetworks() error {
 	var networks []*NetworkEndpoint
 	// prioritize internal networks
 	for _, v := range t.config.Networks {
-		if !v.Internal {
-			networks = append(networks, v)
+		if v.Internal {
+			t.ops.Apply(v)
 		} else {
-			networks = append([]*NetworkEndpoint{v}, networks...)
+			networks = append(networks, v)
 		}
 	}
 
--- a/lib/tether/tether.go
+++ b/lib/tether/tether.go
@@ -255,7 +255,9 @@ func (t *tether) setNetworks() error {
 	// prioritize internal networks
 	for _, v := range t.config.Networks {
 		if v.Internal {
-			t.ops.Apply(v)
+			if err := t.ops.Apply(v); err != nil {
+				return fmt.Errorf("failed to apply network endpoint config: %s", err)
+			}
 		} else {
 			networks = append(networks, v)
 		}
--- a/lib/tether/tether.go
+++ b/lib/tether/tether.go
@@ -252,6 +252,7 @@ func (t *tether) setHostname() error {
 
 func (t *tether) setNetworks() error {
 
+	// internal networks must be applied first
 	for _, network := range t.config.Networks {
 		if !network.Internal {
 			continue
--- a/lib/tether/tether.go
+++ b/lib/tether/tether.go
@@ -251,23 +251,24 @@ func (t *tether) setHostname() error {
 }
 
 func (t *tether) setNetworks() error {
-	var networks []*NetworkEndpoint
-	// prioritize internal networks
-	for _, v := range t.config.Networks {
-		if v.Internal {
-			if err := t.ops.Apply(v); err != nil {
-				return fmt.Errorf("failed to apply network endpoint config: %s", err)
-			}
-		} else {
-			networks = append(networks, v)
-		}
-	}
 
-	for _, network := range networks {
+	for _, network := range t.config.Networks {
+		if !network.Internal {
+			continue
+		}
 		if err := t.ops.Apply(network); err != nil {
 			return fmt.Errorf("failed to apply network endpoint config: %s", err)
 		}
 	}
+	for _, network := range t.config.Networks {
+		if network.Internal {
+			continue
+		}
+		if err := t.ops.Apply(network); err != nil {
+			return fmt.Errorf("failed to apply network endpoint config: %s", err)
+		}
+	}
+
 	return nil
 }
 
--- a/tests/test-cases/Group1-Docker-Commands/1-17-Docker-Network-Connect.robot
+++ b/tests/test-cases/Group1-Docker-Commands/1-17-Docker-Network-Connect.robot
@@ -146,20 +146,21 @@ Connect containers to an internal network
     Should Contain  ${output}  2 packets transmitted, 2 packets received
 
 Check Name Resolution Between Containers On Internal Network
-    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -d --name foo --net internal-net alpine:latest sleep 2000
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -d --name foo --net internal-net alpine:latest sleep 5000
     Log  ${output}
     Should Be Equal As Integers  ${rc}  0
 
-    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} create -it --name bar --net public-net -p 80 alpine:latest /bin/ping -c3 foo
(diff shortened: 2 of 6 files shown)


chengwang86:
Does it have any impact on upgrade or reconfigure?

gigawhitlocks:
Potentially, I'll speak with Emma

lib/tether/tether.go
```diff
@@ -251,8 +251,18 @@ func (t *tether) setHostname() error {
 }
 
 func (t *tether) setNetworks() error {
+	var networks []*NetworkEndpoint
+	// prioritize internal networks
 	for _, v := range t.config.Networks {
-		if err := t.ops.Apply(v); err != nil {
+		if !v.Internal {
```
vburenin: Very cosmetic comment: It doesn't really change anything, especially for the binary output, however I would remove "!" and swap if/else clauses.

lib/tether/tether.go
```diff
@@ -251,8 +251,18 @@ func (t *tether) setHostname() error {
 }
 
 func (t *tether) setNetworks() error {
+	var networks []*NetworkEndpoint
+	// prioritize internal networks
 	for _, v := range t.config.Networks {
```
vburenin: from the future proof stand point if user end-up using tens of thousands of networks it's better to go over local networks first and put them to the front and in the second loop add the rest, so you would avoid making bunch of large copies. The is a little more efficient way of doing this in one pass, but it will make this code a little less readable. However, you may accept the challenge.

lib/tether/tether.go
```diff
@@ -251,8 +251,18 @@ func (t *tether) setHostname() error {
...
-		if err := t.ops.Apply(v); err != nil {
+		if !v.Internal {
+			networks = append(networks, v)
+		} else {
+			networks = append([]*NetworkEndpoint{v}, networks...)
+		}
+	}
+
+	for _, network := range networks {
```
vburenin: Looking at this logic, it makes me think that you don't even need to build an extra array. Just go over local networks first and than over non local.

lib/tether/tether.go
```diff
@@ -251,8 +251,18 @@ func (t *tether) setHostname() error {
 }
 
 func (t *tether) setNetworks() error {
+	var networks []*NetworkEndpoint
+	// prioritize internal networks
 	for _, v := range t.config.Networks {
-		if err := t.ops.Apply(v); err != nil {
+		if v.Internal {
+			t.ops.Apply(v)
```
vburenin: where is an error check?

emlin:
> Does it have any impact on upgrade or reconfigur

this should be fine, because for old version's container, we don't change tether logic, so it will never work for this function. but for the new container, with updated PL and tether, this will be same for upgraded or new installed VCH.
So no need to add a migration plugin for this attribute.

gigawhitlocks:
My test is passing reliably when run against a Nimbus VC:
```
Connect containers to multiple bridge networks overlapping            | PASS |
Connect container to a new network                                    | PASS |
Connect to non-existent container                                     | PASS |
Connect to non-existent network                                       | PASS |
Connect containers to multiple networks non-overlapping               | PASS |
Connect containers to an internal network                             | PASS |
Check Name Resolution Between Containers On Internal Network          ....PING foo (172.24.0.2): 56 data bytes
64 bytes from 172.24.0.2: seq=0 ttl=64 time=15.246 ms
64 bytes from 172.24.0.2: seq=1 ttl=64 time=0.621 ms
64 bytes from 172.24.0.2: seq=2 ttl=64 time=0.737 ms

--- foo ping statistics ---
3 packets transmitted, 3 packets received, 0% packet loss
round-trip min/avg/max = 0.621/5.534/15.246 ms
Check Name Resolution Between Containers On Internal Network          | PASS |
``

.... still trying to figure out why it's not passing in Drone

File: lib/tether/tether.go [hunk 5]
```diff
@@ -251,11 +251,25 @@ func (t *tether) setHostname() error {
...
 func (t *tether) setNetworks() error {
-	for _, v := range t.config.Networks {
-		if err := t.ops.Apply(v); err != nil {
+
+	// internal networks must be applied first
+	for _, network := range t.config.Networks {
+		if !network.Internal {
+			continue
+		}
+		if err := t.ops.Apply(network); err != nil {
```
anchal-agrawal: Being pedantic here - this block can be made terse with

```go
if network.Internal && t.ops.Apply(network) != nil {
    ...
}
```

This is less readable than the current form though, so we can leave it as-is.

tests/test-cases/Group1-Docker-Commands/1-17-Docker-Network-Connect.robot
```diff
@@ -144,3 +144,23 @@ Connect containers to an internal network
...
     Should Contain  ${output}  2 packets transmitted, 2 packets received
+
+Check Name Resolution Between Containers On Internal Network
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} network create --internal mynet
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} network create pubnet
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -d --name foo --net mynet alpine:latest sleep 10000
+    Log  ${output}
+    Should Be Equal As Integers  ${rc}  0
+
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} create -i --name baz --net pubnet -p 80 alpine:latest ping -c3 foo
```
mhagen-vmware: the -i and the -p 80 don't really contribute much to this test do they?

gigawhitlocks: `-i` makes it feckin' pass, because drone I guess
`-p` was just lifted from Ben's example, don't think it does anything but I wanted to keep parity

lib/tether/tether.go
gigawhitlocks: boy

emlin:
lgtm

gigawhitlocks:
Last build just failed because Drone got stuck `waiting 5 minutes until previous build is complete` forever

Issue #4850, opened 2017-04-24
Nfs volume mounts fail silently in the Tether when an incorrect network is attached for pathing to the the NFS server


matthewavery:
#For bug reports, please include the information below:

**VIC version:**

post 1.1 - once pr #4246 is merged

**Deployment details:**

`docker run -it -v nfs-volume:/myData busybox //note the lack of a network option` 

**Steps to reproduce:**

With a valid nfs volume store, Run the above in two separate terminals and touch some files in each container. These files will not show up in the other container. An easier way to check as well is to run the mount command to display active mounts. From there you can note that the desired nfs mount is not present. 

**Actual behavior:**

Mount fails silently and container continues operation without displaying any error to the user. 

**Expected behavior:**

We gracefully handle a failed mount, currently our option is to panic the tether or fail silently. We may want to work out an acceptable solution to reporting back to the user. 

**Logs:**
[nfs-tether-mount-failure.zip](https://github.com/vmware/vic/files/952429/nfs-tether-mount-failure.zip)

mhagen-vmware:
the example given is an easy repro step, but this problem will be seen anytime the nfs share become unreachable due to any number of networking or nfs protocol related issues.

hickeng:
To have the tether fail in a structured manner due to mount failure we need to have @caglar10ur 's restructure work. A core part of that was structured error reporting, which is supposed to go hand-in-hand with pairing errors against Handle elements when we can detect the failure earlier - unfortunately that's part of #3636  work.

The complete hack approach would be to have the mount set the `started` field for the primary session to an error message, then stop the tether.

I suspect we have a middle ground by bringing forwards at least part of the restructure work and making the mount code an extension. That way it can return an error which will trigger tether shutdown. We will still need to add a path for the error reporting however.

Issue #3636, opened 2017-01-18
Portlayer: Interface and structure for port layer components


hickeng:
The port layer components should be presenting common usage patterns across the components.

Should result in:
- [ ] code layout
- [ ] component interfaces - excluding full type definitions
- [ ] interface version mechanisms
- [ ] definition of common port layer supporting services (vmomi gateway)
- [ ] common error returns
- [ ] usage patterns and documentation
- [ ] service & diagnostics mechanism (high level approach, not with per component details)

Notes:
* should allow for components to be co-hosted or separated
* does not encompass REST portion
* should we address other components/functions here such as logging?
* should consider but not be gated by how Pods are created and managed

hickeng:
This ended up being a lot more indepth and inter-related than anticipated. Some pieces could be done now (e.g. the file structure and layout) but would end up changing again as a consequence of the output of this issue.
It may be worth doing regardless so that there's consistency and clarity in the meantime.

PR #6075, opened 2017-08-21
[specific ci=1-44-Docker-CP-Online]return error if nfs volume mount failed

### LINKED ISSUE #5942: Nightly 5-22: NFS simultaneous write to file failed
8/6, build 12979, vSphere 6.5
[5-22-NFS-Volume.zip](https://github.com/vmware/vic/files/1206539/5-22-NFS-Volume.zip)

NFS simultaneous write failed somewhere.  No indication of failures in the logs.  Test 'Simultaneous Container Write to File' during check

emlin:
fixes #5942 

Based on discussion this afternoon with @matthewavery @anchal-agrawal @mhagen-vmware @mdubya66 , we don't want to have nfs volume mount failed silently as described in issue #4850, cause in the case, if user didn't realize the volume is not mounted, and continue to write to the volume store, the data will eventually go to container r/w layer, and after container is gone, the data is lost permanently.

So the decision for 1.2 is that we fail the container if nfs volume mount failed. With this fix, if volume store mount failed in tether, container will quit immediately, and vm is powered off with following error:

```
docker: Error response from daemon: Server error from portlayer: unable to wait for process launch s
Aug 21 2017 22:20:20.581Z INFO  Environment saved in testnfs2/testnfs2.env                           │tatus: container VM has unexpectedly powered off.
```
Do not get back the real error message is because we lack of channel to get tether error from PL, that should be addressed in #4850 in future release.

### DIFF
--- a/lib/tether/tether.go
+++ b/lib/tether/tether.go
@@ -303,11 +303,15 @@ func (t *tether) setMounts() error {
 		switch mountTarget.Source.Scheme {
 		case "label":
 			// this could block indefinitely while waiting for a volume to present
-			t.ops.MountLabel(context.Background(), mountTarget.Source.Path, mountTarget.Path)
-
+			// return error if mount volume failed, to fail container start
+			if err := t.ops.MountLabel(context.Background(), mountTarget.Source.Path, mountTarget.Path); err != nil {
+				return err
+			}
 		case "nfs":
-			t.ops.MountTarget(context.Background(), mountTarget.Source, mountTarget.Path, mountTarget.Mode)
-
+			// return error if mount nfs volume failed, to fail container start, so user knows there is something wrong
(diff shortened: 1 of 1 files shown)


matthewavery:
lgtm - glad this ended up being a simple change :) This is a good fix for us right now. Hopefully we can revisit #4850 and #3636 in the future so that we will be able to propagate an error all the way back to the user.

caglar10ur:
Should we also check the err for the case above?

emlin:
> Should we also check the err for the case above?

I ever checked with @matthewavery for that line, and he said that even if this mount error is not checked, the udev will fail eventually. But I think fail here should be better.

emlin:
no any other change except rebase with master, and remove caglar's code, cause that's already merged to master

caglar10ur:
👍

PR #6077, opened 2017-08-21
Check for power off after kill regardless of startGuestProgram() return

### LINKED ISSUE #5803: Nightly Test 5-6-1- VSAN Simple - container does not exit (InvalidPowerState)
Seen in 6.5: 
[5-6-1-VSAN-Simple.zip](https://github.com/vmware/vic/files/1170573/5-6-1-VSAN-Simple.zip)

screen shot 2017-07-24 at 12 06 43 pm

Initial Thoughts: Is this an issue confirming insecure registries?

sflxn:
If a container VM starts up and we stop it before VC detects the toolbox, startGuestProgram
returns an error 3016 even if the toolbox executed and the VM was powered off.  When this
happens, we try to power off the VM, throwing vSphere into an invalid state transition, which
it needs to recover from.  During this time, we have seen refreshing properties return invalid
powerstate value and prevents us from reading the exit code in portlayer's commit code.

To prevent this, we now wait to see if the container VM powered off after startGuestProgram,
regardless if it returns an error.

Resolves #5803

Fixes #

### DIFF
--- a/lib/portlayer/exec/base.go
+++ b/lib/portlayer/exec/base.go
@@ -248,20 +248,23 @@ func (c *containerBase) kill(ctx context.Context) error {
 	log.Infof("sending kill -%s %s", sig, c.ExecConfig.ID)
 
 	err := c.startGuestProgram(timeout, "kill", sig)
-	if err == nil {
-		log.Infof("waiting %s for %s to power off", wait, c.ExecConfig.ID)
-		err := c.vm.WaitForPowerState(timeout, types.VirtualMachinePowerStatePoweredOff)
-		if err == nil {
-			return nil // VM has powered off
-		}
-
-		if timeout.Err() != nil {
-			log.Warnf("timeout (%s) waiting for %s to power off via SIG%s", wait, c.ExecConfig.ID, sig)
-		}
+	if err == nil && timeout.Err() != nil {
+		log.Warnf("timeout (%s) waiting for %s to power off via SIG%s", wait, c.ExecConfig.ID, sig)
 	}
-
 	if err != nil {
 		log.Warnf("killing %s attempt resulted in: %s", c.ExecConfig.ID, err)
+	}
+
+	// Even if startGuestProgram failed above, it may actually have executed.  If the container came up and then
+	// we kill it before VC gets a chance to detect the toolbox, vSphere can execute the kill but report an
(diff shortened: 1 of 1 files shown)


lib/portlayer/exec/base.go
```diff
@@ -248,22 +248,25 @@ func (c *containerBase) kill(ctx context.Context) error {
...
-		}
+	if err == nil && timeout.Err() != nil {
+		log.Warnf("timeout (%s) waiting for %s to power off via SIG%s", wait, c.ExecConfig.ID, sig)
 	}
-
 	if err != nil {
 		log.Warnf("killing %s attempt resulted in: %s", c.ExecConfig.ID, err)
 	}
 
+	// Even if strtGuestProgram failed above, it may actually have executed.  If the container came up and then
```
dougm: typo "strt" -> "start"

lib/portlayer/exec/base.go
```diff
@@ -248,22 +248,25 @@ func (c *containerBase) kill(ctx context.Context) error {
...
+	// we kill it before VC gets a chance to detect the toolbox, vSphere can execute the kill but report an
+	// error indicating the guest toolbox wasn't found.  If we then try to poweroff, it may throw vSphere into
+	// an invalid transition and will need to recover.  We work around this by waiting on the power state,
+	// regardless of error from startGuestProgram.
+	// https://github.com/vmware/vic/issues/5803
+	log.Infof("waiting %s for %s to power off", wait, c.ExecConfig.ID)
+	err = c.vm.WaitForPowerState(timeout, types.VirtualMachinePowerStatePoweredOff)
```
cgtexmex: I guess I'm not seeing it...but how is this different than before?  If there's no `err` we will `WaitForPowerState` - isn't that what we did before?

cgtexmex: ok..been a long day...I gotcha... :-)

cgtexmex:
LGTM

sflxn:
ping

cgtexmex:
@sflxn you're changing the Kill method, but the particular issue was discovered while calling Stop -- help me to understand how this change would help the current issue?

sflxn:
@cgtexmex @mhagen-vmware Good catch.  I updated the PR by changing line 302 to issue a warning in the logs instead of returning an error from the function shutdown.

PR #6081, opened 2017-08-22
Fix broken H5 Client plugin on vSphere 6.5u1

jooskim:
This PR fixes a customer-found issue where the H5 Client plugin for VIC Engine was not showing up in the H5 Client that ships with the latest release of vSphere (6.5u1). This fix ensures that the plugin is deployed successfully and behaves the same as before on previous builds of vSphere 6.5. It was also tested on a 6.5.0d VCSA instance to ensure backwards compatibility.

It should be noted, however, that there is an internal bug filed against H5 Client that affects the normal operation of the plugin after successful deployment of it. The workaround suggested is to restart the H5 Client service.


jooskim:
@stuclem, feel free to ping me about details on this issue.

PR #6082, opened 2017-08-22
[full ci] Run initilization code once while reloading

### LINKED ISSUE #5833: docker exec causing init customizations to revert
Steps to reproduce; using Build: v1.1.0-rc3-**12176**-7738cb5

1. Customize /etc/hosts file, hostname in entrypoint script.
2. /etc/init.d/iptables stop in entrypoint script.
3. docker -H IP:port --tls exec -d <container> sh /path/to/**any**script
4. /etc/hosts and hostname revert to defaults and iptables service back online.

caglar10ur:
This is a temporary workaround for now until we make those functions
idempotent.

Fixes #5833

### DIFF
--- a/lib/tether/tether.go
+++ b/lib/tether/tether.go
@@ -62,6 +62,9 @@ var (
 	// that are mounted over the system ones.
 	BindSys = system.NewWithRoot("/.tether")
 	once    sync.Once
+
+	// used for running vm initialization logic once in the reload loop
+	initialize sync.Once
 )
 
 type tether struct {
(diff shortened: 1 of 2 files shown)


lib/tether/tether.go
```diff
@@ -521,44 +526,50 @@ func (t *tether) Start() error {
...
+			}
 
-		// process the networks then publish any dynamic data
-		if err := t.setNetworks(); err != nil {
-			log.Error(err)
-			return err
-		}
-		extraconfig.Encode(t.sink, t.config)
+			// process the networks then publish any dynamic data
+			if err = t.setNetworks(); err != nil {
```
hmahmood: Including this here may cause problems later when we (if ever) enable hot add for NICs (to support `docker network connect` on running containers), but for now I don't see any other way to solve the problem at hand.

PR #6083, opened 2017-08-22
Sync container VM power state when refreshing [specific ci=Group1-Docker-Commands]

dougm:
Issue #4872

### DIFF
--- a/lib/portlayer/exec/container.go
+++ b/lib/portlayer/exec/container.go
@@ -319,7 +319,19 @@ func (c *Container) Refresh(ctx context.Context) error {
 	c.m.Lock()
 	defer c.m.Unlock()
 
-	return c.refresh(ctx)
+	if err := c.refresh(ctx); err != nil {
+		return err
+	}
+
+	// sync power state (see issue 4872)
+	switch c.containerBase.Runtime.PowerState {
+	case types.VirtualMachinePowerStatePoweredOn:
+		c.state = StateRunning
+	case types.VirtualMachinePowerStatePoweredOff:
+		c.state = StateStopped
+	}
+
+	return nil
 }
 
 func (c *Container) refresh(ctx context.Context) error {


mdubya66:
Nice

cgtexmex:
When you issue a remove the persona will call and check state -- that call results in an update to the container on the PL.  The Refresh() method is called which initiates a property collector call and ultimately an update to the `base` container -- are we not propagating the updated status to the `container` object?  Just curious if we need this new call or do we just need to fix what we have?

lib/portlayer/exec/container.go
```diff
@@ -546,7 +546,11 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
 
 	// check state first
 	if c.state == StateRunning {
-		return RemovePowerError{fmt.Errorf("Container is powered on")}
+		// double check to workaround issue 4872
+		state, _ := c.vm.PowerState(ctx)
```
anchal-agrawal: We should add a `// #nosec: Errors unhandled.` above this for the gas tool.

dougm:
@cgtexmex I'll look at changing Refresh() to update the power state, it does not do so currently.

hmahmood:
@cgtexmex @dougm `Refresh()` does quite a lot; maybe just add another helper function to just refresh the state?

sflxn:
@cgtexmex I saw something similar in the issue I merged in today where the property refresh was not getting the correct power state.  My suspicion was vSphere correcting it's internal FSM while we were refreshing properties.  In my issue, VIC's internal state tracking was right and VC knew the VM was powered off but property refresh got incorrect state.

dougm:
@hmahmood can you re-review?

@sflxn that's not the problem here, see recent comments in the linked issue.

cgtexmex:
@dougm you should change your commit message...we are no longer double checking but just propagating the state during a refresh...

sflxn:
We should add [specific ci=Group1-Docker-Commands] to test this against through the docker commands to see if everything is fine first.

sflxn:
lgtm.  but you need Caglar or George to lgtm.

PR #6098, opened 2017-08-23
[skip ci] Replace awk with cut in UI scripts

### LINKED ISSUE #6094: Dependency on awk causes plugin install/upgrade to fail
**VIC version:**

vic dev-build 13338

**Steps to reproduce:**

Run the vCenter Plugin install/upgrade (vic/ui/VCSA/[install.sh,upgrade.sh] on a system with `awk` missing, the install.sh and upgrade.sh scripts fails.

image

**Actual behavior:**

The install.sh and upgrade.sh scripts fail.

While the output states `./install.sh: line 129: awk: command not found`, the actual error message being printed is unclear (caused by the thumbprint field being blank).

**Additional details:**

`awk` is not part of the default Photon OS install and needs to be added with `tdnf install gawk`.

jooskim:
Fixes #6094

### DIFF
--- a/ui/installer/VCSA/install.sh
+++ b/ui/installer/VCSA/install.sh
@@ -126,7 +126,7 @@ check_prerequisite () {
     fi
 
     # retrieve VC thumbprint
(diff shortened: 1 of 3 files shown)


jak-atx:
This PR is out of date. We have this change on jooskim fork and will push the changes into our new vic-ui repo once ready.

PR #6102, opened 2017-08-23
[specific ci=6-04-Create-Basic] bump default debug level to 1

### LINKED ISSUE #6096: set debug to 1 by default
As a VIC user, if any error happens, I want to download log bundle with enough information immediately, so that I don't have to enable debug and reproduce the issue once again.

emlin:
fixes #6096

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -40,6 +40,8 @@ const (
 	MaxVirtualMachineNameLen = 80
 	// Max permitted length of Virtual Switch name
 	MaxDisplayNameLen = 31
+	// Default debug level in create
+	DefaultDebug = 1
 )
 
 var EntireOptionHelpTemplate = `NAME:
@@ -424,17 +426,9 @@ func (c *Create) processParams() error {
 
 func (c *Create) processCertificates() error {
 
-	// debuglevel is a pointer now so we have to do this song and dance
-	var debug int
-	if c.Debug.Debug == nil {
-		debug = 0
-	} else {
-		debug = *c.Debug.Debug
-	}
-
 	c.certs.Networks = c.Networks
 
-	if err := c.certs.ProcessCertificates(c.DisplayName, c.Force, debug); err != nil {
+	if err := c.certs.ProcessCertificates(c.DisplayName, c.Force, *c.Debug.Debug); err != nil {
 		return err
 	}
 
@@ -682,7 +676,12 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 
 	log.Infof("### Installing VCH ####")
 
-	if c.Debug.Debug != nil && *c.Debug.Debug > 0 {
+	if c.Debug.Debug == nil {
+		// set default value to 1, to avoid additional step to collect logs
+		debug := DefaultDebug
+		c.Debug.Debug = &debug
+	}
+	if *c.Debug.Debug > 0 {
 		log.SetLevel(log.DebugLevel)
 		trace.Logger.Level = log.DebugLevel
 	}
--- a/tests/test-cases/Group6-VIC-Machine/6-04-Create-Basic.md
+++ b/tests/test-cases/Group6-VIC-Machine/6-04-Create-Basic.md
@@ -101,6 +101,7 @@ vic-machine create --name=<VCH_NAME> --target=<TEST_URL> \
 
 ### Expected Outcome
 * Deployment succeed
+* VCH is deployed with debug enabled
 * Regression test pass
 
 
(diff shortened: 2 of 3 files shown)


mdubya66:
we should plan on picking this for RC2

mdubya66:
also may require a small doc change to indicate the level is 1 and that it can be changed. @stuclem

mdubya66:
what is the behavior after an upgrade?

mdubya66:
from @hickeng on slack
1) verify that log rotation is working
2) confim that tether is not logging load/save of extraconfig and process output to debug log as part of debug level 1 and if it is move it to level 2. @caglar10ur 
3) get perf results of container vm's as this could impact them the most.

emlin:
> what is the behavior after an upgrade?

This change does not impact upgrade, also upgrade will not change debug configuration. So it will be same to before.

emlin:
>     verify that log rotation is working
@caglar10ur Do we have log rotation? I'm running test to verify if we have, but didn't see it in the code.
>     confirm that tether is not logging load/save of extraconfig and process output to debug log as part of debug level 1 and if it is move it to level 2. @caglar10ur
Right now, tether is logging configuration at level 1, I can move it to level 2.
>     get perf results of container vm's as this could impact them the most.
@hickeng now the pprof service is started at level 2 debug. Do you want it at level 1? Wouldn't that impact container performance?

mdubya66:
should upgrade change the default debug? I think it should so we can support those customers.

emlin:
In the current logic, upgrade does not change configuration, and configure does not upgrade old version.
We could mix them with additional options, like to add --configure into upgrade command, or --upgrade into configure command, but it is one feature requirement, not a bug fixing.
It will introduce other questions, for examples, if that will change other properties in upgrade, for example, ops-user token. If we don't answer those questions clearly, this function will add more confusion for customer.

btw, upgrade will involve power cycle anyway, even if we don't combine the configure and upgrade features, if users don't mind to upgrade their VCH, run another configure command does not introduce too much work.
So I'll prefer not to change debug configuration in upgrade atm.

emlin:
Checked the log rotation worked well in VCH, but there is no log rotation for tether.debug. @caglar10ur for confirmation.
And I think other log level settings and debug service settings are also not enabled by default.

cmd/vic-machine/create/create.go
```diff
@@ -424,17 +426,9 @@ func (c *Create) processParams() error {
...
-	if c.Debug.Debug == nil {
-		debug = 0
-	} else {
-		debug = *c.Debug.Debug
-	}
-
 	c.certs.Networks = c.Networks
 
-	if err := c.certs.ProcessCertificates(c.DisplayName, c.Force, debug); err != nil {
+	if err := c.certs.ProcessCertificates(c.DisplayName, c.Force, *c.Debug.Debug); err != nil {
```
caglar10ur: I think we still need to do the check before this as c.Debug.Debug can be nil, right?

emlin: I made change below to set it to 1, if it's nil, so removed this check

caglar10ur: 👍 didn't realize that runs before this fn

emlin:
After talked with @caglar10ur , need to correct the pprof information.
In VCH, pprof is started for all services, and will be collected disregard of the debug level.
In container VM, pprof is still not started unless the debug level is 2 or more.

mdubya66:
what's the expected perf hit for pprof in the containervm?

caglar10ur:
There is no hard overhead number that we can simply use, we need to measure. Why we are interested with pprof in containervm? That shouldn't be really needed unless we debug tether live

mdubya66:
The purpose of this is to be able to support production customers without them having to take their applications offline because they are not always able to do so.

caglar10ur:
@mdubya66 I meant we shouldn't enable pprof for container vms by default, that's mainly for debugging tether which I think not out main goal here. For the services, it is already enabled in VCH so we are already paying that overhead.

mdubya66:
I defer to you. The issue we are trying to resolve is reducing the amount of back and forth we need to have with a customer when they have a VCH or C-VM issue, especially after they experience a VC outage.

caglar10ur:
I still think we don't need pprof on the cvm side but I defer to @hickeng and you to decide :)

mdubya66:
Since i deferred to you first then we don't enable it. ;)

mhagen-vmware:
~I would vote to reduce the back and forth with a customer, that instead of this change, we actually add the system log bundle (or a select subset of it) to our vicadmin log collection...~
Actually, we already do. Part of the problem was that we need the VC logs as well, but our container log bundle actually include hostd logs for all hosts it looks like.

stuclem:
Created https://github.com/vmware/vic-product/issues/654 to update the doc if we do decide to do this.

PR #6104, opened 2017-08-24
Fixing panic in case object received is invalid

hmahmood:


lib/config/dynamic/admiral/source.go
```diff
@@ -422,6 +422,11 @@ func (o *productDiscovery) Discover(ctx context.Context, sess *session.Session)
 
 	var vms []*vm.VirtualMachine
 	for _, o := range objs {
+		if o.Type == nil || o.ID == nil {
+			log.Warnf("skipping invalid object reference %+v", o)
+			continue
+		}
+
 		if o.Type != nil && *o.Type != "VirtualMachine" {
```
emlin: likely this o.Type != nil is not needed

hmahmood: Fixed

Issue #5628, opened 2017-06-30
Delete containers failed "Server error from portlayer: [DELETE /containers/{id}][500] containerRemoveInternalServerError "


lgayatri:
./admiral --version
Version: 1.2.0-SNAPSHOT

root@Ubuntu-VIc:/home/vmlib/VIC/VIC-11728/vic/admiral-cli-1.2.0-SNAPSHOT/bin# ./admiral rm --insecure --query RUNNING -f
2017.06.30 01:51:39 Task ID: a15d1ce39d1934755532704aee8f8
|=============>------------------------------------------|    25%

|=============>------------------------------------------|    25%

|=======================================================>|    100%
2017.06.30 01:52:34 The task has ERROR.
Service https://10.197.37.173:2376/v1.24/containers/70e032674cebc56e62e7f0873f4fbd3adfb2b1103bd17a7d57c4355b4e24d31e?v=1&force=1 returned error 500 for DELETE. id 27788395 message Server error from portlayer: [DELETE /containers/{id}][500] containerRemoveInternalServerError ; Reason: {"message":"Server error from portlayer: [DELETE /containers/{id}][500] containerRemoveInternalServerError "}

[viclogs.zip](https://github.com/vmware/vic/files/1114259/viclogs.zip)

Logs attached:

corrieb:
@lgayatri The logs included aren't useful in diagnosing the issue here. Please can you include logs from the vic-admin server when you raise a bug like this. It has the port layer logs that can tell us what might have been wrong. Also if you can run your VCHs with --debug 1, this will include more information in the logs.

Can you reproduce this?

lgayatri:
@corrieb , I moved to newer VC/ESX/NSX builds as my test bed  crashed and now having issues with Logical switch creation on NSX. Once I get it resolved will provide the needinfo

corrieb:
Note there are numerous issues open reporting a containerRemoveInternalServerError. 
#5198
#5157 
#4858
#1823 

We need to get to the bottom of this.

lgayatri:
Due to issue on the storage switch, my test bed is unavailable. Tkt: RITM0187046
Will be able to proceed once the ticket gets resolved.

lgayatri:
My Test bed came up partially, and here is the output on build 12202
![image](https://user-images.githubusercontent.com/25006749/28522325-2df7c0da-7095-11e7-89f4-427295b972f0.png)

[...]

PR #6105, opened 2017-08-24
[specific ci=Group1-Docker-Commands] Wait for awhile for stable properties after a poweroff

sflxn:
Attempt to wait for awhile for power state property after we perform
a poweroff operation.  If the VM is already powered off and we attempt
to perform a poweroff operation and then we refresh properties, we
often see the properties refresh return invalid power state (reports
powered on when vSphere sees stable powered off).  It is believe this
causes vSphere to go into a transition to return to a proper state for
the VM.  We wait for awhile to test the power state after a poweroff
in an attempt to detect this occurance.

Might help with #5628

### DIFF
--- a/lib/portlayer/exec/base.go
+++ b/lib/portlayer/exec/base.go
@@ -331,14 +331,14 @@ func (c *containerBase) poweroff(ctx context.Context) error {
 	})
 
 	if err != nil {
-
 		// It is possible the VM has finally shutdown in between, ignore the error in that case
 		if terr, ok := err.(task.Error); ok {
+			waitForStableOff := false
 			switch terr := terr.Fault().(type) {
 			case *types.InvalidPowerState:
 				if terr.ExistingState == types.VirtualMachinePowerStatePoweredOff {
 					log.Warnf("power off %s task skipped (state was already %s)", c.ExecConfig.ID, terr.ExistingState)
-					return nil
+					waitForStableOff = true
 				}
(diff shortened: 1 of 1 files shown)


sflxn:
@dougm We may not need this change.  The heavy tests on VC was done with an old build that doesn't include our stop fix for 1.2.  We're going to wait to see if we can reproduce it with the latest build.

caglar10ur:
I believe that 10 sec. is coming from vcenter as that's their scan interval. If we are going to do this I suggest using a bit longer than 2 x interval plus some small interval (like 21 sec.) to make sure that we cover all window

lib/portlayer/exec/base.go
```diff
@@ -331,14 +331,14 @@ func (c *containerBase) poweroff(ctx context.Context) error {
...
-
 		// It is possible the VM has finally shutdown in between, ignore the error in that case
 		if terr, ok := err.(task.Error); ok {
+			waitForStableOff := false
 			switch terr := terr.Fault().(type) {
 			case *types.InvalidPowerState:
 				if terr.ExistingState == types.VirtualMachinePowerStatePoweredOff {
 					log.Warnf("power off %s task skipped (state was already %s)", c.ExecConfig.ID, terr.ExistingState)
-					return nil
+					waitForStableOff = true
```
emlin: Am wondering if it's fine to return true here directly, instead of wait another few seconds?
Is there any possible side effect for that?

Issue #6084, opened 2017-08-22
Nightly 5-14-Remove-Container-OOB


cgtexmex:
This test will remove a containerVM via govc, but when installing to a vSphere cluster we disable the vSphere VM remove functionality so that we control when containerVMs can be removed.

We will need to re-enable the vSphere method prior to attempting the removal OOB.  This is strictly a test issue and not a code issue.

[5-14-Remove-Container-OOB-08202017.zip](https://github.com/vmware/vic/files/1243337/5-14-Remove-Container-OOB-08202017.zip)

cgtexmex:
set priority to high as this will fail every night...

chengwang86:
Does it just start to fail recently? I remember it didn't fail before.

cgtexmex:
I couldn't find any historical data on the test, so not sure...it's been failing this past week.

chengwang86:
I checked out our previous emails for nightly tests. It started to fail on Aug 2nd. Let's see what nightly test related PR were merged on/before that day.

cgtexmex:
Based on the git history looks like the test had been disabled for other issues and was just recently enabled.

chengwang86:
You are right. It was disabled by PR https://github.com/vmware/vic/pull/5089 since May 10th.

chengwang86:
This is reproducible in my local manual test.

The fix is simple as you suggested:
```
govc object.method -name Destroy_Task -enable %{TEST_DATACENTER}/vm/%{VCH-NAME}/${vm}
govc vm.destroy %{VCH-NAME}/${vm}
```

chengwang86:
The fix is blocked until we bump the integration-test image to use govc 0.15, since `govc object.method` is introduced since 0.15.

chengwang86:
@cgtexmex @mhagen-vmware On a second thought, since we disable OOB VM deletion intentionally in our code, this failure is expected. So instead of enabling `Destroy_Task` before OOB removal, we might do `Should Contain  ${out}   ... is disabled by VIC` after the govc cmd.

PR #6109, opened 2017-08-25
[full ci] Update integration test image to govc 0.15.0

mhagen-vmware:
partial fix for #6084


chengwang86:
Irrelevant question:  why is the image tagged as 1.35 not 1.34?  Lol

mhagen-vmware:
@chengwang86 because there is an existing 1.34 image that a PR is using

PR #6097, opened 2017-08-23
[skip ci] Re-enable destroy_task on containeVM before OOB removal

chengwang86:
Fixes #6084

PR #6111, opened 2017-08-25
[skip ci] Enable OOB delete before attempting it

### LINKED ISSUE #6084: Nightly 5-14-Remove-Container-OOB

mhagen-vmware:
fixes #6084

### DIFF
--- a/tests/resources/Vsphere-Util.robot
+++ b/tests/resources/Vsphere-Util.robot
@@ -36,7 +36,7 @@ Power Off VM OOB
 
 Destroy VM OOB
     [Arguments]  ${vm}
-    ${rc}  ${output}=  Run Keyword If  '%{HOST_TYPE}' == 'VC'  Run And Return Rc And Output  govc object.method -name Destroy_Task -enable %{VCH-NAME}/${vm}
+    ${rc}  ${output}=  Run Keyword If  '%{HOST_TYPE}' == 'VC'  Run And Return Rc And Output  govc object.method -name Destroy_Task -enable %{TEST_DATACENTER}/vm/%{VCH-NAME}/${vm}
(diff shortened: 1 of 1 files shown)


tests/resources/Vsphere-Util.robot
```diff
@@ -36,6 +36,8 @@ Power Off VM OOB
 
 Destroy VM OOB
     [Arguments]  ${vm}
+    ${rc}  ${output}=  Run Keyword If  '%{HOST_TYPE}' == 'VC'  Run And Return Rc And Output  govc object.method -name Destroy_Task -enable %{VCH-NAME}/${vm}
```
chengwang86: when I tried this manually, it didn't work on vc. You need to have `%{TEST_DATACENTER}/vm/%{VCH-NAME}/${vm}` . See https://github.com/vmware/vic/pull/6097/files#diff-fa372ce03d7f9d1bc404459cab0adc75R39

Another question:  why don't we re-enable destroy_task if the host_type is esx?

dougm: You need to use the full inventory path with the object.* commands: `/dc-name/vm/vapp-name/vm-name` 
Of if there is a single datacenter, it can be relative to that: `vm/vapp-name/vm-name`

dougm: @chengwang86 ESX doesn't support {Enable,Disable}Methods

PR #6128, opened 2017-08-28
[skip ci] Switch this test to check for disabled method

mhagen-vmware:
Our fix for the underlying issue was to disable the ability to remove a container OOB, so the underlying issue is still present just masked. It doesn't make sense to bypass our "fix" and allow the OOB delete to then just get the same failure as we originally worked around.

### DIFF
--- a/tests/manual-test-cases/Group5-Functional-Tests/5-14-Remove-Container-OOB.robot
+++ b/tests/manual-test-cases/Group5-Functional-Tests/5-14-Remove-Container-OOB.robot
@@ -24,10 +24,11 @@ Docker run an image from a container that was removed OOB
 
     Install VIC Appliance To Test Server
 
-    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull busybox
+    ${rc}  ${out}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull busybox
     Should Be Equal As Integers  ${rc}  0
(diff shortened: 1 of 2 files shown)


tests/manual-test-cases/Group5-Functional-Tests/5-14-Remove-Container-OOB.robot
```diff
@@ -24,10 +24,11 @@ Docker run an image from a container that was removed OOB
...
+    ${rc}  ${out}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull busybox
     Should Be Equal As Integers  ${rc}  0
     ${rc}  ${container}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -itd --name removeOOB busybox /bin/top
     Should Be Equal As Integers  ${rc}  0
-    Destroy VM OOB  removeOOB*
-    ${rc}  ${container}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -itd busybox /bin/top
-    Should Be Equal As Integers  ${rc}  0
+
+    ${rc}  ${out}=  Run And Return Rc And Output  govc vm.destroy %{VCH-NAME}/removeOOB*
+    Should Not Be Equal As Integers  ${rc}  0
```
chengwang86: Are we sure that the ${rc} is 0 in this case?

mhagen-vmware: should not be equal

chengwang86: Sorry my bad. I was probably reading the deleted lines :)

PR #6124, opened 2017-08-25
[skip ci] Document the design for the vic-machine API

zjs:
This document proposes a design for a comprehensive `vic-machine` API, the implementation of which will be tracked by #6116.

Subsets of this API (tracked by #5721, #6123, and eventually others) will be implemented incrementally, and the design will be revised as those efforts progress to reflect changes to the long-term vision.

Implementation status will be captured in GitHub, not this document.

Resolves #6014

PR #6132, opened 2017-08-28
[skip ci] Define an OpenAPI document for vic-machine

### LINKED ISSUE #6015: Define the VCH management API as an OpenAPI/Swagger document
Define the VCH management API as an OpenAPI/Swagger document and use it generate server code during the build process.

zjs:
Use an OpenAPI (Swagger) document to define a preliminary API for
vic-machine based on the vic-machine API design document (#6116).

This document is expected to evolve as implementation proceeds.

Resolves #6015.

Note: this change builds on #6124  and, once approved, I would like this to land as the second commit on `feature/vic-machine-service`.


lib/apiservers/service/swagger.json
```diff
@@ -0,0 +1,879 @@
...
+        "produces": [
+          "text/plain"
+        ],
+        "responses": {
+          "200": { "$ref": "#/responses/version" }
+        },
+        "security": []
+      }
+    },
+    "/{target}": {
```
hmahmood: Should this be `/targets/{target}`? This would reduce the chances of target name colliding with other resource names under `/`.

lib/apiservers/service/swagger.json
```diff
@@ -0,0 +1,879 @@
...
+        "responses": {
+          "202": { "$ref": "#/responses/vsphere-task" },
+          "default": { "$ref": "#/responses/error" }
+        }
+      }
+    },
+    "/{target}/datacenter/{datacenter}": {
+      "get": {
+        "summary": "Show information about the specified vSphere resources",
+        "description": "Making a `GET` request on a datacenter will return information about the state of the host firewall on those resources.",
```
hmahmood: The path of the resource `/{target}/datacenter/` does not indicate it has anything to do with firewalls. Can we change that reflect what the call is actually about?

zjs: I expect that the resource will include additional information; firewall status is just the first item.

We could expose separate resources for each piece of information (`/{target}/datacenter/{datacenter}/firewall`, `/{target}/datacenter/{datacenter}/foo`, `/{target}/datacenter/{datacenter}/bar`), but instead I think it makes sense to have a single `/{target}/datacenter/{datacenter}` endpoint with a response body like below, and add sub-resources on an as-needed basis if necessary:
```json
{
    "firewall": { 
    },
    "foo": {
    },
    "bar": {
    }
}
```

lib/apiservers/service/swagger.json
zjs: Because `{target}` must be a valid IP address or hostname, I think the risk of collision is low.

Given the current `vic-machine` operations, and general model that all VIC operations involve a vSphere resource of some sort, `/version` may be the _only_ resource directly under `/`.

lib/apiservers/service/swagger.json
hmahmood: Yes, this is fine if more than just firewall info is going to be retrieved.

lib/apiservers/service/swagger.json
hmahmood: Name collisions aside, it just seems odd to address a target under `/`. We're not managing the target after all. As a consumer of the API, I would expect endpoints under `/` to be directly relevant to the vic-machine service itself, like `/version`.

cmd/vic-machine-server/main.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/vmware/vic/lib/apiservers/service/restapi"
+	"github.com/vmware/vic/lib/apiservers/service/restapi/operations"
+)
+
+// This file was generated by the swagger tool.
+// Make sure not to overwrite this file after you generated it because all your edits would be lost!
+
+func main() {
+
```
caglar10ur: extra line

zjs: > This file was generated by the swagger tool.

Given that the file was generated, and `go fmt` didn't remove it, I'm inclined to leave it as-is.

If I start touching the file, I'm going to wonder if I should also remove line 55, wonder why there's no line between 32 and 33, but is a line between 50 and 52, and... none of that seems like a good use of time.

PR #6135, opened 2017-08-28
[specific ci=1-02-Docker-Pull] Show Harbor error response detail to client side

AngieCris:
Extend on pr #6067 
For error response from Harbor registry (like 412), the error message now shows up on the client side.

Upon a 412 (precondition failed) from Harbor due to unsigned image:
before the change, the client sees:
```
Unexpected http code 412
```
Now, the client sees:
```
Unexpected http code: 412 (Precondition Failed), URL: ..., Message: The image is not signed in Notary.
```
also the complete error response body json string is included in the debug log.

(related issues:
original issue https://github.com/vmware/vic/issues/5951
Opened this pr because Harbor's err json does not match Docker's https://github.com/vmware/harbor/issues/3124)

### DIFF
--- a/pkg/fetcher/fetcher.go
+++ b/pkg/fetcher/fetcher.go
@@ -461,15 +461,20 @@ func extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
 	res := []byte(out.Bytes())
 	log.Debugf("Error message json string: %s", string(res))
 
-	var errResponse RegistryErrorRespBody
-	err = json.Unmarshal(res, &errResponse)
-	if err != nil {
-		log.Debugf("Error when unmarshaling error response body: %s", err)
-		return "", err
+	// extract directly if the response has a `message` field
+	var errMsg struct {
+		Message string `json:"message"`
+	}
+	err = json.Unmarshal(res, &errMsg)
+	if err == nil && errMsg.Message != "" {
+		return errMsg.Message, nil
 	}
 
-	if len(errResponse.Errors) == 0 {
-		log.Debugf("Error response wrong format. Response body: %s", string(res))
+	// otherwise, treat the response according to Docker format
+	var errResponse RegistryErrorRespBody
+	err = json.Unmarshal(res, &errResponse)
+	if err != nil || len(errResponse.Errors) == 0 {
+		log.Debugf("Error response has irregular format. Response body: %s", string(res))
 		return "", errJSONFormat
 	}
 
--- a/pkg/fetcher/fetcher.go
+++ b/pkg/fetcher/fetcher.go
@@ -474,7 +474,7 @@ func extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
 	var errResponse RegistryErrorRespBody
 	err = json.Unmarshal(res, &errResponse)
 	if err != nil || len(errResponse.Errors) == 0 {
-		log.Debugf("Error response has irregular format. Response body: %s", string(res))
+		log.Infof("Error response has irregular format. Response body: %s", string(res))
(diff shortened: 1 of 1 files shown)


pkg/fetcher/fetcher.go
```diff
@@ -461,15 +461,20 @@ func extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
...
+		return errMsg.Message, nil
 	}
 
-	if len(errResponse.Errors) == 0 {
-		log.Debugf("Error response wrong format. Response body: %s", string(res))
+	// otherwise, treat the response according to Docker format
+	var errResponse RegistryErrorRespBody
+	err = json.Unmarshal(res, &errResponse)
+	if err != nil || len(errResponse.Errors) == 0 {
+		log.Debugf("Error response has irregular format. Response body: %s", string(res))
```
matthewavery: I am curious if we should actually log this at the `Info` level. I am wondering how often we would expect to see this? and it might be useful to have this kind of error logged on a standard deployment rather than encountering it and then having to reproduce it in a debug environment.

AngieCris:
not necessarily needed for release 1.2...
Harbor's error response json conforms to an old format that may get phased out by Docker at any moment
sometime in the future Harbor will likely to update the error response json format
so this is just an expedient change

pkg/fetcher/fetcher.go
```diff
@@ -461,15 +461,20 @@ func extractErrResponseMessage(rdr io.ReadCloser) (string, error) {
 	res := []byte(out.Bytes())
 	log.Debugf("Error message json string: %s", string(res))
```
AngieCris: I logged the error response json string in two places:
1. every time when there's an error response body (here)
2. when the err json is malformed
basically if the err json has expected format, the complete string will be logged once 
if the err json is irregular, the err string will be logged twice
I just got comments from @matthewavery that logging at debug level is expensive
is it better if I just log it only when the json is malformed, and log it at info level instead? 
(ping @sflxn for advice

matthewavery: loggin at debug level is not expensive, we want to avoid logging anything that might be large above the `debug` level. The debug level is for when we are wanting the maximum amount of logging from VIC. This means that it will slow us down the most, we want to avoid logging anything above debug if we do not have to since that can cause performance or space issues. We want to log this above the debug level(in my opinion) because it is valuable for us to get the information on the first failure. if we would be logging the malformed json body twice(above debug) then that is likely not necessary.

AngieCris: i see. yes you're definitely right that it's valuable to log the first failure (malformed json response) above debug level. fixing this right now.
about logging the complete response json for every error (not only the malformed one), i think we can do it at debug level - it's additional information to the client (the error json could contain other details not extracted to the command line output)

AngieCris:
UPDATE:
Harbor updated their error format to match with Docker's in pr https://github.com/vmware/harbor/pull/3125

But since there could be other registries with different error format, I'll keep this pr just in case of irregular err json convention

matthewavery:
@anchal-agrawal @sflxn @chengwang86 pinging other potential reviewers. Since we can probably merge this soon.

chengwang86:
Angie made this PR based on the observation that the error response returned by Harbor is something like 
```
var errMsg struct {
		Message string `json:"message"`
}
```
which doesn't comply with docker's documentation. Since Harbor has merged their PR to fix this issue, I don't think we need to merge it at the moment, unless we find another registry that suffers from the same issue.

mdubya66:
let's close this @AngieCris

AngieCris:
Close this issue @mdubya66

Issue #6127, opened 2017-08-28
Longevity failure: Error response from daemon: Conflict error from portlayer: [PUT /containers/{handle}][409] commitConflict &{Code:0 Message:Reconfigure: collision of concurrent operations


mhagen-vmware:
From the personality log:
```
Aug 26 2017 18:29:37.086Z DEBUG [ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.(*ContainerProxy).Stop:755] [27.756302201s] f0ad2cd7d5bdfa1ab07aff974dce4d5315b2bbb1ccdd500b49241ba8f47c7a2a
Aug 26 2017 18:29:37.086Z ERROR Will stop trying again. Operation failed with errors.apiError{error:(*errors.errorString)(0xc421837070), statusCode:409}
Aug 26 2017 18:29:37.086Z DEBUG [ END ] [github.com/vmware/vic/pkg/retry.DoWithConfig:73] [1m9.102444141s] 
Aug 26 2017 18:29:37.086Z DEBUG [ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.(*Container).ContainerStop:1216] [1m9.102455544s] f0ad2cd7d5bdfa1ab07aff974dce4d5315b2bbb1ccdd500b49241ba8f47c7a2a
Aug 26 2017 18:29:37.086Z ERROR Handler for POST /v1.23/containers/f0ad2cd7d5bdfa1ab07aff974dce4d5315b2bbb1ccdd500b49241ba8f47c7a2a/stop returned error: Conflict error from portlayer: [PUT /containers/{handle}][409] commitConflict  &{Code:0 Message:Reconfigure: collision of concurrent operations - expected power state poweredOff, found poweredOn}
Aug 26 2017 18:29:37.095Z DEBUG [BEGIN] [github.com/vmware/vic/lib/apiservers/engine/backends.DockerEventPublisher.PublishEvent:217]
Aug 26 2017 18:29:37.095Z DEBUG [ END ] [github.com/vmware/vic/lib/apiservers/engine/backends.DockerEventPublisher.PublishEvent:217] [11.869µs] 
Aug 26 2017 18:29:37.095Z DEBUG [BEGIN] [github.com/vmware/vic/lib/apiservers/engine/backends.(*ContainerProxy).exitCode:867]
Aug 26 2017 18:29:37.108Z DEBUG [BEGIN] [github.com/vmware/vic/lib/apiservers/engine/backends/convert.State:32]
[...]

PR #6136, opened 2017-08-28
 log VM power state if there is state mismatch

emlin:
help #6127 to log out more message

### DIFF
--- a/lib/portlayer/exec/commit.go
+++ b/lib/portlayer/exec/commit.go
@@ -138,6 +138,11 @@ func Commit(ctx context.Context, sess *session.Session, h *Handle, waitTime *int
 			if s.ExtraConfig != nil && h.TargetState() == StateStopped && h.Runtime.PowerState != types.VirtualMachinePowerStatePoweredOff {
 				detail := fmt.Sprintf("Reconfigure: collision of concurrent operations - expected power state poweredOff, found %s", h.Runtime.PowerState)
 				log.Warn(detail)
+
+				// log out current vm power state and runtime power state got from refresh, to see if there is anything mismatch,
+				// cause in issue #6127, we see the runtime power state is not updated even after 1 minute
+				ps, _ := h.vm.PowerState(ctx)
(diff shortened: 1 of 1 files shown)


lib/portlayer/exec/commit.go
```diff
@@ -138,6 +138,11 @@ func Commit(ctx context.Context, sess *session.Session, h *Handle, waitTime *int
 			if s.ExtraConfig != nil && h.TargetState() == StateStopped && h.Runtime.PowerState != types.VirtualMachinePowerStatePoweredOff {
 				detail := fmt.Sprintf("Reconfigure: collision of concurrent operations - expected power state poweredOff, found %s", h.Runtime.PowerState)
 				log.Warn(detail)
+
+				// log out current vm power state and runtime power state got from refresh, to see if there is anything mismatch,
+				// cause in issue #6127, we see the runtime power state is not updated even after 1 minute
+				ps, _ := h.vm.PowerState(ctx)
+				log.Debugf("Container %s power state: %s, runtime power state", h.ExecConfig.ID, ps, h.Runtime.PowerState)
```
caglar10ur: you have 3 arguments but 2 formatting verbs here

lib/portlayer/exec/commit.go
```diff
@@ -138,6 +138,11 @@ func Commit(ctx context.Context, sess *session.Session, h *Handle, waitTime *int
 			if s.ExtraConfig != nil && h.TargetState() == StateStopped && h.Runtime.PowerState != types.VirtualMachinePowerStatePoweredOff {
 				detail := fmt.Sprintf("Reconfigure: collision of concurrent operations - expected power state poweredOff, found %s", h.Runtime.PowerState)
```
caglar10ur: Ad @dougm suggested we should add %T here

lib/portlayer/exec/commit.go
emlin: oh.. no

lib/portlayer/exec/commit.go
caglar10ur: Nevermind this, I mixed up files :)

Issue #6108, opened 2017-08-25
Update ESX assignment approach to drone workers


mdharamadas1:
Currently, ESXi is assigned to drone workers randomly with the following approach: 
Env variable TEST_URL_ARRAY is set with all the available ESXs. Given there are 8 ESXs, current build waits for the previous 8th build to complete before executing the next steps in the CI pipeline. The one issue with this approach is if the latter builds are completed before the 8th build, the current build will still be waiting even though the ESX is available. Thats why we see builds waiting even though there are less than 8 running builds.
Also, recently we added 6 more workers / ESX , so the approach is outdated and needs an update.

Suggestion:
I think test code shouldn't worry about the environment at all and it needs to be provisioned to it instead by CI. I would like to tie an ESX server with the drone worker assuming we are running one build per worker at a time. 
One way to do this is to have env variable TEST_URL (set to single ESX ip) set permanently in the worker VM (may be in /etc/environment ? ) so that the build running on that VM always runs tests on that specific ESX. I think this also makes it easier in debugging / maintenance of the worker / ESX.

mdharamadas1:
@mhagen-vmware would like to know your thoughts on this suggestion?

mhagen-vmware:
That could work, and I think it would be good, potential problems:
1. I am still seeing multiple containers running on a worker at the same time.
2. you might need to make changes to drone in order to push that env var into the container as I believe the only env vars that are pushed in currently are the well known DRONE_ ones and the ones in the secret file.

But doing this will solve the problem of trying to find out which worker a job is running on, since we already print the ESX IP, we will be able to identify the worker with this approach by matching that to a table.

mdharamadas1:
OK cool!
I will investigate multiple container issue before working on this and yes we need to pass that env var to drone containers.
Thanks for the feedback!

anchal-agrawal:
@mdharamadas1 please assign a priority so we can pull this into the backlog. Assignees should be set only if an item is in ToDo or In Progress. Thanks.

mdharamadas1:
Done @anchal-agrawal , sorry missed that.

mhagen-vmware:
https://github.com/drone/drone/issues/2216

mdharamadas1:
sure @mhagen-vmware , I will do #2216 as part of this.

mhagen-vmware:
If you are up for it, might start looking here:
[...]

PR #6148, opened 2017-08-29
[full ci] Update ESX assignment approach to drone workers

mdharamadas1:
fix for #6108.. worker VMs have a file named `test_env` under home dir which is volume mounted inside the docker container to set two env variables `TEST_ESX_IP` and `DRONE_MACHINE`. This maps each worker VM to its ESX server.


mdharamadas1:
@mhagen-vmware @rogeliosanchez @singhshwetaQE This is ready for your review. Pls note I won't be merging this until #6169 is complete.
`DRONE_MACHINE` is not supported with current drone version we are using but will be available in `v0.8`. Also, I didn't create a mapping table yet and instead have `TEST_ESX_IP` as an env variable in each worker VM. When we are ready to upgrade to `v0.8`, we can get rid of this volume mounted file having env variables and instead use mapping table to retrieve ESX IP using `DRONE_MACHINE` value.
Pls note recent CI build is passed with `full-ci`.

@rajanashok Pls let me know if any of these changes impact nightlies. I didn't run that build.

.drone.local.yml
```diff
@@ -35,8 +36,8 @@ pipeline:
...
       GS_PRIVATE_KEY: ${GS_PRIVATE_KEY}
       DOMAIN: ${DOMAIN}
-    commands:
-      - tests/integration-test.sh
-      #- pybot tests/test-cases/Group1-Docker-Commands/1-13-Docker-Version.robot
     volumes:
       - /tmp
+    commands:
+      # - tests/integration-test.sh
+      - pybot tests/test-cases/Group1-Docker-Commands/1-13-Docker-Version.robot
```
mhagen-vmware: change these comments around?

.drone.local.yml
```diff
@@ -20,9 +20,10 @@ pipeline:
       VIC_ESX_TEST_URL: ${VIC_ESX_TEST_URL}
       LOG_TEMP_DIR: install-logs
       DRONE_SERVER:  ${DRONE_SERVER}
+      DRONE_MACHINE: ${DRONE_MACHINE}
```
mhagen-vmware: this might not be set currently right?

.drone.local.yml
mdharamadas1: not intentional, will revert it.

.drone.local.yml
mdharamadas1: No. For local builds, we don't volume mount env var file, so instead set them as env var. It is needed.

tests/resources/VCH-Util.robot
```diff
@@ -27,11 +27,7 @@ Set Test Environment Variables
     ${status}  ${message}=  Run Keyword And Ignore Error  Environment Variable Should Be Set  TEST_DATACENTER
     Run Keyword If  '${status}' == 'FAIL'  Set Environment Variable  TEST_DATACENTER  ${SPACE}
 
-    @{URLs}=  Split String  %{TEST_URL_ARRAY}
-    ${len}=  Get Length  ${URLs}
-    ${IDX}=  Evaluate  %{DRONE_BUILD_NUMBER} \% ${len}
-
-    Set Environment Variable  TEST_URL  @{URLs}[${IDX}]
+    Set Environment Variable  TEST_URL  %{TEST_ESX_IP}
```
mhagen-vmware: This will actually break all the nightlies as the nimbus setups only set TEST_URL_ARRAY and don't currently set TEST_URL or TEST_ESX_IP.  I would actually not like this to be called TEST_ESX_IP as we are actively moving towards VC installs and it doesn't make sense for nightlies either.  Just TEST_URL seems appropriate, and we can change the nightlies to set the TEST_URL var.

mdharamadas1: I thought about renaming it to something else but wasn't sure. I agree with you to rename it to `TEST_URL` instead. I will update that and also change all `TEST_URL_ARRAY` for nightlies to `TEST_URL`.

mdharamadas1:
@mhagen-vmware Updated this as per review comments and also drone build passed with full-ci. Pls check.

anchal-agrawal:
@mdharamadas1 Please resolve the conflicts, thanks.

mdharamadas1:
@anchal-agrawal done.

tests/get_test_url.sh
```diff
@@ -13,8 +13,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-DRONE_BUILD_NUMBER=${DRONE_BUILD_NUMBER:=0}
-unit_test_array=($TEST_URL_ARRAY)
-arrLen=${#unit_test_array[@]}
-
-echo $TEST_USERNAME:$TEST_PASSWORD@${unit_test_array[$(($DRONE_BUILD_NUMBER % $arrLen))]}
+echo $TEST_USERNAME:$TEST_PASSWORD@${TEST_URL}
```
anchal-agrawal: Is this file needed? The only reference to it has been removed.

mdharamadas1:
@mhagen-vmware @anchal-agrawal I wanted to remove dependency on #6169, so I did some changes after your approval. As #6169 seems not possible with current version of drone and I don't want to just throw away changes as part of this, I had to add back waiting step :(, but its still better than current approach. Also, did some renaming to files to avoid confusions. Please take a look and let me know.

mdharamadas1:
Looking at the failed drone build.

mdharamadas1:
build is good now, running with full ci next.

mdharamadas1:
full ci build passed on this one, @mhagen-vmware let me know if I can go ahead and merge this?

.drone.yml
```diff
@@ -17,46 +17,52 @@ pipeline:
...
+  wait-for-build:
+    image: harbor.ci.drone.local/library/vic-integration-test:1.36
+    pull: true
+    environment:
+      BIN: bin
+      GOPATH: /go
+      SHELL: /bin/bash
+      DRONE_SERVER: ${DRONE_SERVER}
+      DRONE_TOKEN: ${DRONE_TOKEN}
+      ACTIVE_BUILD_NUMBER: ${ACTIVE_BUILD_NUMBER}
```
mhagen-vmware: where is this coming from and what is it supposed to represent?

mdharamadas1: `ACTIVE_BUILD_NUMBER` represents the latest build run on that worker and is the additional env var set in the volume mounted file `test_env`, its initial value is `0` but when some job is run on that worker the value is set to that build number so that any other concurrent job running after that can wait until its done.

`test_env` looks like below, which will have 3 variables set.

```
TEST_URL=192.168.31.57
DRONE_MACHINE=vic-drone-05-worker-13
ACTIVE_BUILD_NUMBER=0
```

mhagen-vmware: I think this ultimately is going to be actually worse than the existing solution.. right now, every 14th builds will wait until the previous build completes.  With this solution, every other build will wait until it completes, because drone server appears to load up two jobs on a worker at once or in near vicinity of each other as opposed to a true round robin.  In order to get the benefits of this solution, I believe you will need to re-work it into pinning two ESX per worker in test_env and then somehow decide which one is open...

mdharamadas1: Based on my experience so far with Drone, I think it will not run more than one job per worker concurrently unless all the workers are busy with jobs. That means, waiting is triggered only after 15th running job. Also, we haven't set `DRONE_MAX_PROCS` value, so there is no guarantee that only two jobs per worker at once are run, so pinning 2 ESXs won't help.

mhagen-vmware: Everytime I have looked in the past couple of weeks, there were 2 job running on 3 or 4 workers and the rest were empty.. If you are sure that the server does in fact do a round robin, then LGTM.

mdharamadas1: Cool, I will verify it before I merge this. Thanks!

mdharamadas1:
Closing this PR for now.

PR #6154, opened 2017-08-29
[skip ci] Implement REST API handler to return the version 

### LINKED ISSUE #6016: Implement an API handler for getting version information
Implement an API handler for getting version information

zjs:
Resolves #6016.

Note: this change builds on #6132 and, once approved, I would like this to land as the third commit on `feature/vic-machine-service`.

### DIFF
--- a/lib/apiservers/service/restapi/configure_vic_machine.go
+++ b/lib/apiservers/service/restapi/configure_vic_machine.go
@@ -23,6 +23,7 @@ import (
 	"github.com/go-openapi/runtime/middleware"
 	"github.com/tylerb/graceful"
 
+	"github.com/vmware/vic/lib/apiservers/service/restapi/handlers"
 	"github.com/vmware/vic/lib/apiservers/service/restapi/operations"
 )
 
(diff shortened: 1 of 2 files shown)


lib/apiservers/service/restapi/configure_vic_machine.go
```diff
@@ -47,9 +48,7 @@ func configureAPI(api *operations.VicMachineAPI) http.Handler {
 	})
 
 	// GET /container/version
-	api.GetVersionHandler = operations.GetVersionHandlerFunc(func(params operations.GetVersionParams) middleware.Responder {
-		return middleware.NotImplemented("operation .GetVersion has not yet been implemented")
-	})
```
caglar10ur: how about returning the version directly here without version_get.go (assuming it works)? 

```
api.GetVersionHandler = operations.GetVersionHandlerFunc(func(params operations.GetVersionParams) middleware.Responder {
        return operations.NewGetVersionOK().WithPayload(version.GetBuild().ShortVersion()) 
 })
```

zjs: I believe that would work fine (and is similar to how I implemented it initially), but I think there's a lot of value in having the APIs handled consistently.

If `configure_vic_machine.go` is simply a dispatcher to logic in other files, and never the place where handler logic is implemented, I think it's easier to reason about. (It also allows for consistent patterns around, for example, testing.)

lib/portlayer/attach/communication/lazy.go
```diff
@@ -50,3 +50,11 @@ func (l *LazySessionInteractor) Initialize() (SessionInteractor, error) {
 	}
 	return l.si, nil
 }
+
+// Cleanup returns either an initialized connection, or nil if it was never initialized
+func (l *LazySessionInteractor) Cleanup() (SessionInteractor, error) {
```
jzt: Might be worth rebasing - I seem to remember these changes in a different PR (@hickeng's) that I suspect has already been merged.

zjs: It looks like this picked up a bunch of random CLs from other people. Doing a force push seems to have cleared that up.

PR #6155, opened 2017-08-29
[skip ci] Define REST API handlers for listing VCHs

### LINKED ISSUE #6026: Implement an API handler for listing VCHs
Implement an API handler for listing VCHs

zjs:
Introduce a pair of handlers for listing information about VCHs, optionally scoped to a compute resource or datacenter.

Include basic authentication and error handling, which should be improved on as development continues.

Resolves #6026.

Note: this change builds on #6154 and, once approved, I would like this to land as the fourth commit on `feature/vic-machine-service`.

### DIFF
--- a/lib/apiservers/service/restapi/handlers/vch_list_get.go
+++ b/lib/apiservers/service/restapi/handlers/vch_list_get.go
@@ -1,0 +1,213 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package handlers
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+	"path"
+
+	"github.com/docker/docker/opts"
+	"github.com/go-openapi/runtime/middleware"
+
+	"github.com/vmware/vic/cmd/vic-machine/common"
+	"github.com/vmware/vic/lib/apiservers/service/models"
+	"github.com/vmware/vic/lib/apiservers/service/restapi/handlers/util"
+	"github.com/vmware/vic/lib/apiservers/service/restapi/operations"
+	"github.com/vmware/vic/lib/install/data"
+	"github.com/vmware/vic/lib/install/management"
+	"github.com/vmware/vic/lib/install/validate"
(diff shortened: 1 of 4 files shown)


lib/apiservers/service/restapi/handlers/vch_list_get.go
```diff
@@ -0,0 +1,209 @@
...
+	return "Invalid upgrade status"
+}
+
+func vchsToModels(ctx context.Context, vchs []*vm.VirtualMachine, executor *management.Dispatcher) []*models.VCHListItem {
+	installerVer := version.GetBuild()
+	payload := make([]*models.VCHListItem, 0)
+	for _, vch := range vchs {
+		var version *version.Build
+		var dockerHost string
+		var adminPortal string
```
caglar10ur: you may want to declare them outside of the loop and reusing them if vchs potentially a big slice of vms

lib/apiservers/service/restapi/handlers/vch_list_get.go
```diff
@@ -0,0 +1,209 @@
...
+		if vchConfig, err := executor.GetNoSecretVCHConfig(vch); err == nil {
+			version = vchConfig.Version
+
+			if public := vchConfig.ExecutorConfig.Networks["public"]; public != nil {
+				if public_ip := public.Assigned.IP; public_ip != nil {
+					var docker_port string
```
caglar10ur:
```
docker_port := fmt.Sprintf("%d", opts.DefaultHTTPPort)
if !vchConfig.HostCertificate.IsNil() {
	docker_port = fmt.Sprintf("%d", opts.DefaultTLSHTTPPort)
}
```

saves a declaration and branch :)

lib/apiservers/service/restapi/handlers/vch_list_get.go
```diff
@@ -0,0 +1,209 @@
...
+		name := path.Base(vch.InventoryPath)
+
+		model := &models.VCHListItem{ID: vch.Reference().Value, Name: name}
+
+		if version != nil {
+			model.Version = version.ShortVersion()
+			model.UpgradeStatus = upgradeStatusMessage(ctx, vch, installerVer, version)
+		}
+
+		if adminPortal != "" {
```
caglar10ur: the default values of AdminPortal and DockerHost is "" anyway so I think you could just assign them without checking

lib/apiservers/service/restapi/handlers/vch_list_get.go
zjs: If I declare them outside the loop, I need to re-zero them on each iteration.

vburenin: Just for curiosity I would take a look at the generated code, I suspect all these variable are allocated on stack.

zjs: I'm not sure why it matters whether or not these three variables are allocated on the stack. Could you elaborate?

jzt:
Are the tests you added here related to the VCH listing api handlers?

PR #6158, opened 2017-08-30
[specific ci=3-03-Docker-Compose-Basic] Wait until container removed from network after docker-compose kill

### LINKED ISSUE #5948: docker-compose integration tests not cleaning up on failures

```
...
Removing network vic_default
```

**Proposed solutions:**

chengwang86:
Fixes #5948 . 

Based on the discussion here https://github.com/vmware/vic/issues/5948#issuecomment-326182552, `docker-compose kill redis` may leave the container `vic_redis_1` on the network `vic_default` for some time until the vic engine receives a vsphere event of container being poweredoff, then the vic-engine would unbind the container from the network. 

Therefore, in this PR, we wait until the container is actually removed the network before `docker-compose down` to avoid the failure that `network has active endpoints`. In addition, I've also added more logs in this test to help debug in case the failure occurs again.

### DIFF
--- a/tests/test-cases/Group3-Docker-Compose/3-03-Docker-Compose-Basic.robot
+++ b/tests/test-cases/Group3-Docker-Compose/3-03-Docker-Compose-Basic.robot
@@ -17,6 +17,7 @@ Documentation  Test 3-03 - Docker Compose Basic
 Resource  ../../resources/Util.robot
 Suite Setup  Install VIC Appliance To Test Server  certs=${true}
 Suite Teardown  Cleanup VIC Appliance On Test Server
+Test Teardown  Cleanup Orphan Containers
 
 *** Variables ***
 ${yml}  version: "2"\nservices:\n${SPACE}web:\n${SPACE}${SPACE}image: python:2.7\n${SPACE}${SPACE}ports:\n${SPACE}${SPACE}- "5000:5000"\n${SPACE}${SPACE}depends_on:\n${SPACE}${SPACE}- redis\n${SPACE}redis:\n${SPACE}${SPACE}image: redis\n${SPACE}${SPACE}ports:\n${SPACE}${SPACE}- "5001:5001"
@@ -31,6 +32,17 @@ Check Compose Logs
     Should Be Equal As Integers  ${rc}  0
     Should Contain  ${output}  PING aaa
     Should Not Contain  ${output}  bad address 'aaa'
+
+Cleanup Orphan Containers
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} ps -a
+    Should Be Equal As Integers  ${rc}  0
+    Log  ${output}
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} network inspect vic_default
+    Should Be Equal As Integers  ${rc}  0
+    Log  ${output}
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} rm -f $(docker %{VCH-PARAMS} -a -q)
+    Should Be Equal As Integers  ${rc}  0
+    Log  ${output}
 
 *** Test Cases ***
 Compose basic
@@ -49,6 +61,9 @@ Compose basic
     Log  ${output}
     Should Be Equal As Integers  ${rc}  0
     ${rc}  ${output}=  Run And Return Rc And Output  docker-compose %{COMPOSE-PARAMS} --file basic-compose.yml stop
+    Log  ${output}
+    Should Be Equal As Integers  ${rc}  0
+    ${rc}  ${output}=  Run And Return Rc And Output  docker-compose %{COMPOSE-PARAMS} --file basic-compose.yml ps
     Log  ${output}
     Should Be Equal As Integers  ${rc}  0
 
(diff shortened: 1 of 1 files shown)


tests/test-cases/Group3-Docker-Compose/3-03-Docker-Compose-Basic.robot
```diff
@@ -61,6 +71,16 @@ Compose kill
...
 
+    ${rc}  ${out}=  Run And Return Rc And Output  docker %{VCH-PARAMS} ps -a
+    Log  ${out}
+    Should Be Equal As Integers  ${rc}  0
+
+    ${rc}  ${out}=  Run And Return Rc And Output  docker %{VCH-PARAMS} network inspect vic_default
+    Log  ${out}
+    Should Be Equal As Integers  ${rc}  0
+
+    Wait Until Keyword Succeeds  1 min  5 sec  Check Container Removed From Network  vic_redis_1  vic_default
```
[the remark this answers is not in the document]
chengwang86: Ideally it should be done within seconds since we run it on ESXi instead of VC. I could bump it up to 5 min. If it doesn't succeed within 5 min, definitely we have an issue here. What do you think?

Even if we run this test on VC, the host sync for power state update between the vcenter server and the esxi host is around 20 sec. So if the network is ok, we should be able to receive this event within 1 min.

tests/test-cases/Group3-Docker-Compose/3-03-Docker-Compose-Basic.robot
```diff
@@ -61,6 +71,16 @@ Compose kill
     Log  ${out}
     Should Be Equal As Integers  ${rc}  0
 
+    ${rc}  ${out}=  Run And Return Rc And Output  docker %{VCH-PARAMS} ps -a
+    Log  ${out}
+    Should Be Equal As Integers  ${rc}  0
+
+    ${rc}  ${out}=  Run And Return Rc And Output  docker %{VCH-PARAMS} network inspect vic_default
```
hmahmood: Do we need this since we `Check Container Removed From Network` does this anyway?

chengwang86: You are right. I'll remove it.

tests/test-cases/Group3-Docker-Compose/3-03-Docker-Compose-Basic.robot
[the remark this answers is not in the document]
anchal-agrawal: Sounds good. A "correct" timeout is difficult to determine here, since I'm not aware of an upper bound to the event time delay. I was thinking 2 mins, but if you think 1 min is sufficient, that's fine too.

tests/test-cases/Group3-Docker-Compose/3-03-Docker-Compose-Basic.robot
```diff
@@ -32,6 +32,13 @@ Check Compose Logs
     Should Contain  ${output}  PING aaa
     Should Not Contain  ${output}  bad address 'aaa'
 
+Check Container Removed From Network
+    [Arguments]  ${name}  ${network}
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} network inspect ${network} | jq '.[] | .Containers | .[] | .Name'
+    Should Be Equal As Integers  ${rc}  0
+    Log  ${output}
```
mhagen-vmware: move this above the rc check, otherwise you don't get the log if it fails

tests/test-cases/Group3-Docker-Compose/3-03-Docker-Compose-Basic.robot
```diff
@@ -61,6 +71,16 @@ Compose kill
     Log  ${out}
     Should Be Equal As Integers  ${rc}  0
 
+    ${rc}  ${out}=  Run And Return Rc And Output  docker %{VCH-PARAMS} ps -a
+    Log  ${out}
+    Should Be Equal As Integers  ${rc}  0
```
mhagen-vmware: Are we checking for something here?  Should redis no longer show up in ps?

tests/test-cases/Group3-Docker-Compose/3-03-Docker-Compose-Basic.robot
```diff
@@ -61,6 +71,16 @@ Compose kill
...
+
+    ${rc}  ${out}=  Run And Return Rc And Output  docker %{VCH-PARAMS} network inspect vic_default
+    Log  ${out}
+    Should Be Equal As Integers  ${rc}  0
+
+    Wait Until Keyword Succeeds  1 min  5 sec  Check Container Removed From Network  vic_redis_1  vic_default
+
```
mhagen-vmware: I don't necessarily like this... what are we asking the user to do here? I would think that in our response to users is to run the compose down again? So why don't we just run compose down until it succeeds for a reasonable number of times?

tests/test-cases/Group3-Docker-Compose/3-03-Docker-Compose-Basic.robot
chengwang86: Both redis and web should show up, and I want to see their running status. This is added as extra debug information.

tests/test-cases/Group3-Docker-Compose/3-03-Docker-Compose-Basic.robot
chengwang86: No, I don't think we can run compose down here. 

If you run `docker-compose down`, it will first remove the container and then remove the network. If the vsphere event for the container being powered off is received after `compose down` removes the container, we may fail to unbind it from the network (container does not exist). And the network will always have this dangling container (although it is already removed from cache).

mhagen-vmware: So at that point, after doing a perfectly normal set of steps, the customer is screwed and cannot do anything?

PR #6159, opened 2017-08-30
Run only upgrade and regression tests on tag event

anchal-agrawal:
This change borrows the `specific ci` logic to run only `Group11-Upgrade` and `7-01-Regression` tests on a tag build (`tag` event on a `*refs/tags*` branch).


anchal-agrawal:
Canceled the build since CI is overwhelmed. I've had a green build on this PR already: https://ci.vcna.io/vmware/vic/13481

PR #6190, opened 2017-09-01
[skip ci] Remove dangling container during network rm

### LINKED ISSUE #5948: docker-compose integration tests not cleaning up on failures

chengwang86:
Fixes #5948 . 

The reason we observe `network has active endpoints` failure during `docker-compose down` is because that a container is already removed from the PL container cached but still exists in the network layer container cache. We refer to such a container as a `dangling` container on the network.

In this PR, when `network rm` is called, we remove these dangling containers (if any).

### DIFF
--- a/lib/portlayer/network/context.go
+++ b/lib/portlayer/network/context.go
@@ -1270,16 +1270,19 @@ func (c *Context) DeleteScope(ctx context.Context, name string) error {
 	}
 
 	// remove dangling endpoints
-	for _, eps := range s.Endpoints() {
-		if exec.Containers.Container(eps.ID().String()) == nil {
-			log.Debugf("Remove dangling endpoint (%s)", eps.ID().String())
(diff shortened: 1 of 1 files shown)


lib/portlayer/network/context.go
```diff
@@ -1269,6 +1269,18 @@ func (c *Context) DeleteScope(ctx context.Context, name string) error {
 		return fmt.Errorf("cannot remove builtin scope")
 	}
 
+	// remove dangling endpoints
+	if exec.Containers != nil {
+		for _, eps := range s.Endpoints() {
+			if exec.Containers.Container(eps.ID().String()) == nil {
```
matthewavery: small nit: instead of calling `eps.ID().String()` 3 times here you could assign it a variable once and then just use that variable instead :) will make the code a little more friendly to read.

PR #6205, opened 2017-09-04
[full ci] toolbox: reset session when invalidated by the vmx

### LINKED ISSUE #6152: toolbox needs to reset session when invalidated by ESX
Seen cases where ESX is under heavy load, where the toolbox session
cookie is invalidated by the vmx.

On the guest side, this results in a flood of:
```
   Message: Unable to send a message over the communication channel 0
   Message: Unable to poll for messages over the communication channel 0
```

On the vmx site (vmware.log), this results in a flood of:
```
   GuestMsg: Channel 0, Protocol error, state: 0
   GuestMsg: channel 0: wrong cookie, discarding message.
```

Related to #5621

dougm:
Cherry pick of vmware/govmomi#829

master branch has this change via the full govmomi update in #6144

Fixes #6152

### DIFF
--- a/vendor/github.com/vmware/govmomi/toolbox/service.go
+++ b/vendor/github.com/vmware/govmomi/toolbox/service.go
@@ -48,6 +48,10 @@ var (
 	}
 
 	netInterfaceAddrs = net.InterfaceAddrs
+
+	// If we have an RPCI send error, the channels will be reset.
+	// open-vm-tools/lib/rpcChannel/rpcChannel.c:RpcChannelCheckReset also backs off in this case
+	resetDelay = time.Duration(500) // 500 * 10ms == 5s
 )
 
 // Service receives and dispatches incoming RPC requests from the vmx
@@ -59,6 +63,7 @@ type Service struct {
 	stop     chan struct{}
 	wg       *sync.WaitGroup
 	delay    time.Duration
+	rpcError bool
 
 	Command *CommandServer
 	Power   *PowerCommandHandler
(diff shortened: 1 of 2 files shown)


cgtexmex:
LGTM

vendor/github.com/vmware/govmomi/toolbox/service.go
```diff
@@ -110,14 +115,37 @@ func (s *Service) backoff() {
...
+}
+
+func (s *Service) startChannel() error {
 	err := s.in.Start()
 	if err != nil {
 		return err
 	}
 
-	err = s.out.Start()
+	return s.out.Start()
```
emlin: is it possible to have s.in.Start() succeeds but s.out.Start failed? And if that happens, do you need to stop the s.in first before return error?

vburenin:
LGTM

vendor/github.com/vmware/govmomi/toolbox/service.go
dougm: I don't know of a case where `out.Start` would error, other than if not running inside a VM.  But in that case, `in.Start` would have error'd first.

The `in` channel is for **vmx to guest** RPC, there can only be 1.  So `in.Start` can fail if another instance of toolbox (or vmtoolsd) is already registered.

The `out` channel is for the **guest to vmx** RPC, there is no limit that I'm aware of.

emlin: I see the point. thanks

PR #6209, opened 2017-09-05
[specific ci=3-03-Docker-Compose-Basic] Unbind container in containerRm

### LINKED ISSUE #5948: docker-compose integration tests not cleaning up on failures

chengwang86:
Fixes #5948 

This PR tries to unbind the container in the before removing the container. 

This may potentially degrade the performance of containerRm. However, since the PL will not do anything if the container is already unbound, its impact on performance may be neligible.


hmahmood:
Will this actually solve the issue with `kill` not actually unbinding the container from the network?

chengwang86:
@hmahmood I think the original issue can be fixed by doing this. The reason why the failure occurs is that the container is removed before it is unbound, and that `compose down` calls `container stop` followed by `container rm ` instead of `container rm -f`. So this PR unbinds the container in the code path where the containerRM() is called without `-f`.

chengwang86:
Deleting `dangling containerVMs` from network layer when calling `deleteScope` is also an option, but it requires that the network layer knows about the status of the exec cache, which violates the isolation of layers.

hmahmood:
So if call `docker kill` for a container, it can potentially still be connected to its network afterwards if I don't follow it with an `rm`?

hmahmood:
Why do we not have a special case for `SIGKILL` in https://github.com/vmware/vic/blob/master/lib/apiservers/engine/backends/container.go#L634 as we do on the portlayer side (https://github.com/vmware/vic/blob/master/lib/portlayer/exec/container.go#L454), and unbind the container there?

chengwang86:
@hmahmood If `docker kill` is not followed by `docker rm`, the container will not be removed from cache and the containerPoweredOff event will finally be processed properly. So the container will be unbound from the network in this case.

The failure occurs because `docker kill` is followed by `docker rm` before the event is processed.

About why I didn't have a special case checking `SIGKILL` in `containerKill()`, George's opinion was that we should target at a more general set of failures that might occur due to this race condition, instead of only focusing on `docker kill`. @hickeng Thoughts?

hmahmood:
What about port mappings? Do they also get unmapped when with a `kill`?

chengwang86:
@hmahmood ContainerKill() calls container_proxy.Signal(), which would **unmap** the ports if the state is not `running`. So I think the ports should've been unmapped after `docker kill`.

In this case, I think I can move `unmapPorts()` outside of `UnbindContainerfFromNetwork()`.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -835,6 +826,28 @@ func (c *ContainerProxy) Stop(vc *viccontainer.VicContainer, name string, second
...
+		switch err := err.(type) {
+		case *scopes.UnbindContainerNotFound:
+			// ignore error
+			log.Warnf("Container %s not found by network unbind", vc.ContainerID)
+		case *scopes.UnbindContainerInternalServerError:
+			return "", InternalServerError(err.Payload.Message)
+		default:
+			return "", InternalServerError(err.Error())
+		}
+	} else {
```
anchal-agrawal: Don't think the `else` block is needed - we can just say

```go
    if {
        ...
    }

    return ub.Payload.Handle, nil
```

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -835,6 +826,28 @@ func (c *ContainerProxy) Stop(vc *viccontainer.VicContainer, name string, second
 	return nil
 }
 
+func (c *ContainerProxy) UnbindContainerFromNetwork(vc *viccontainer.VicContainer, handle string) (string, error) {
```
anchal-agrawal: We should have function comments for exported functions.

lib/apiservers/engine/backends/container.go
```diff
@@ -761,6 +761,16 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 		if state.Status == "Starting" {
 			return derr.NewRequestConflictError(fmt.Errorf("The container is starting.  To remove use -f"))
 		}
+
+		handle, err := c.Handle(id, name)
+		if err != nil {
+			return err
```
anchal-agrawal: Is this a docker-typed error?

chengwang86: I think so. It returns either `NotFoundError` or `InternalServerError`.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -835,6 +826,27 @@ func (c *ContainerProxy) Stop(vc *viccontainer.VicContainer, name string, second
 	return nil
 }
 
+// UnbindContainerFromNetwork unbinds a container from the networks that it connects to
+func (c *ContainerProxy) UnbindContainerFromNetwork(vc *viccontainer.VicContainer, handle string) (string, error) {
+	defer trace.End(trace.Begin(""))
```
jzt: Would it be worthwhile to trace on the container ID here?

PR #6212, opened 2017-09-05
[full ci] Add a test timeout within the drone timeout limit

mhagen-vmware:
We are seeing more and more drone inactivity timeouts within our CI tests, this forces a kill before hitting that timeout so that we can collect logs.


mdubya66:
This is to help resolve the issues seen where drone has been timing out more often.

PR #6216, opened 2017-09-05
[specific ci=Group10-VCH-Restart] toolbox: avoid race when closing channels on stop

dougm:
It was possible for the Stop() method close the RPC channels when
the main loop was in the middle of Send/Receive.


mdubya66:
Since this is now a panic we have to fix it. Approved for 1.2. LGTM.

PR #6222, opened 2017-09-06
[skip ci] Implement new vsphere restart tests

mhagen-vmware:
currently blocked by:

https://bugzilla.eng.vmware.com/show_bug.cgi?id=1953928


hickeng:
Looks like we may have a configuration solution in the form of an advanced VC config option: `vpxd.SerialPort.PersistNetworkBacking=True`

If we can confirm, this should be added as a documented requirement and checked in vic-machine

PR #6247, opened 2017-09-07
[specifc ci=1-07-Docker-Stop] Retry WaitForPowerState in case of InvalidPowerState

### LINKED ISSUE #6236: Failure on docker stop ${container} "Error response from daemon: Conflict error from portlayer... collision of concurrent operations"
**VIC version:**
vic_13602

VCH-0-5249 is still up on "blinky" and responding to docker commands.

**Logs:**
[longevity_6-5_VCH-0-5249.zip](https://github.com/vmware/vic/files/1284020/longevity_6-5_VCH-0-5249.zip)

screen shot 2017-09-07 at 3 38 09 am

chengwang86:
Fixes #6236

### DIFF
--- a/lib/portlayer/exec/base.go
+++ b/lib/portlayer/exec/base.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"time"
 
+	log "github.com/Sirupsen/logrus"
 	"golang.org/x/crypto/ssh"
 
 	"github.com/vmware/govmomi/guest"
@@ -27,14 +28,12 @@ import (
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/lib/config/executor"
 	"github.com/vmware/vic/lib/migration"
+	"github.com/vmware/vic/pkg/retry"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/extraconfig"
 	"github.com/vmware/vic/pkg/vsphere/extraconfig/vmomi"
 	"github.com/vmware/vic/pkg/vsphere/tasks"
 	"github.com/vmware/vic/pkg/vsphere/vm"
-
-	log "github.com/Sirupsen/logrus"
-	"github.com/vmware/vic/pkg/retry"
 )
 
 const defaultWaitTime = 10 * time.Second
@@ -246,7 +245,7 @@ func (c *containerBase) stop(ctx context.Context, waitTime *int32) error {
 		isTimeoutErr := func(err error) bool {
 			return timeout.Err() != nil
 		}
-				
+
 		config := retry.NewBackoffConfig()
 		config.MaxElapsedTime = 2 * time.Minute
 
--- a/lib/portlayer/exec/base.go
+++ b/lib/portlayer/exec/base.go
@@ -36,7 +36,7 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/vm"
 )
 
-const defaultWaitTime = 10 * time.Second
+const defaultPowerStateWaitTime = 10 * time.Second
 
 // NotYetExistError is returned when a call that requires a VM exist is made
 type NotYetExistError struct {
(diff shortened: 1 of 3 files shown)


lib/portlayer/exec/base.go
```diff
@@ -231,7 +233,37 @@ func (c *containerBase) stop(ctx context.Context, waitTime *int32) error {
...
+		}
+
+		isTimeoutErr := func(err error) bool {
+			return timeout.Err() != nil
+		}
+
+		config := retry.NewBackoffConfig()
+		config.MaxElapsedTime = 2 * time.Minute
+
+		if err = retry.DoWithConfig(operation, isTimeoutErr, config); err != nil {
```
hmahmood: Is the intent here to call `WaitForPowerState` with a timeout of `10s` retrying up to 2 minutes? I don't think the `timeout` context is viable after the first for `WaitForPowerState` times out; you will need to create a new context every time.

lib/portlayer/exec/base.go
```diff
@@ -231,7 +233,37 @@ func (c *containerBase) stop(ctx context.Context, waitTime *int32) error {
...
+	if err == nil {
+		timeout, cancel := context.WithTimeout(ctx, defaultPowerStateWaitTime)
+		defer cancel()
+
+		operation := func() error {
+			return c.vm.WaitForPowerState(timeout, types.VirtualMachinePowerStatePoweredOff)
+		}
+
+		isTimeoutErr := func(err error) bool {
+			return timeout.Err() != nil
```
hmahmood: You should not retry here if `timeout` was canceled. Only retry if `Err() == DeadlineExceeded`.

lib/portlayer/exec/base.go
```diff
@@ -231,7 +233,37 @@ func (c *containerBase) stop(ctx context.Context, waitTime *int32) error {
 
 	log.Warnf("stopping %s via hard power off due to: %s", c.ExecConfig.ID, err)
 
-	return c.poweroff(ctx)
+	err = c.poweroff(ctx)
+
+	// In case of InvalidPowerState and GenericVmConfigFault during hard poweroff, nil would be returned.
+	// At the time the VM must've been already powered off. But we may need to wait for extra time
+	// such that we can obtain the latest powerstate from the VC.
+	if err == nil {
```
hmahmood: Should we retry in the InvalidPowerState case here? Also in the `GenericVmConfigFault` case we should probably return the error from `poweroff()`, so that we can do this extra wait here. Or do the actual wait in `poweroff()`?

lib/portlayer/exec/base.go
chengwang86: Thanks. I will correct this one.

lib/portlayer/exec/base.go
chengwang86: For the case of `GenericVmConfigFault`, in `poweroff()`, if the error msg indicates `vmNotSuspendedKey` or `vmPoweringOffKey`, `poweroff()` returns `nil` and the retry logic will be executed. Otherwise, the error will be returned and the retry logic will not be called. So I think it should be fine to either have the retry logic here instead or inside `poweroff()`.

hmahmood: `poweroff()` returns nil in this case: https://sourcegraph.com/github.com/chengwang86/vic@ab324c9076cbf71e94fac1532e6d5e7a5b945ea9/-/blob/lib/portlayer/exec/base.go#L373-374 . We should not retry in this case.

lib/portlayer/exec/base.go
```diff
@@ -231,7 +233,37 @@ func (c *containerBase) stop(ctx context.Context, waitTime *int32) error {
...
+		timeout, cancel := context.WithTimeout(ctx, defaultPowerStateWaitTime)
+		defer cancel()
+
+		operation := func() error {
+			return c.vm.WaitForPowerState(timeout, types.VirtualMachinePowerStatePoweredOff)
```
emlin: wait for timeout is one method waiting and retry inside already, why do you still need the backoff retry here?
if the expected longest time is 2 min, you could set timeout to 2 mins directly

emlin:
kill have same behavior with stop: https://github.com/vmware/vic/pull/6247/files#diff-67f57bbb6bec9888ba685e947a8a00cfR305
you'd better to fix that at the same time

emlin:
https://github.com/vmware/vic/pull/6247/files#diff-67f57bbb6bec9888ba685e947a8a00cfR336
https://github.com/vmware/vic/pull/6247/files#diff-67f57bbb6bec9888ba685e947a8a00cfR333
As @hickeng said, it might fail for power state has been powered off, so need to check the error in the above two places as well.

PR #6251, opened 2017-09-08
fix InvalidArgument error during resource pool creation (#6249)

emlin:
cherry pick #6249 to hotpatch branch

### DIFF
--- a/lib/install/management/resource_pool.go
+++ b/lib/install/management/resource_pool.go
@@ -54,16 +54,11 @@ func (d *Dispatcher) createResourcePool(conf *config.VirtualContainerHostConfigS
 	log.Infof("Creating Resource Pool %q", conf.Name)
 	// TODO: expose the limits and reservation here via options
 	resSpec := types.ResourceConfigSpec{
-		CpuAllocation: &types.ResourceAllocationInfo{
-			ExpandableReservation: types.NewBool(true),
-		},
-		MemoryAllocation: &types.ResourceAllocationInfo{
-			ExpandableReservation: types.NewBool(true),
-		},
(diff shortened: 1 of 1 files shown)


lib/install/management/resource_pool.go
```diff
@@ -88,6 +83,9 @@ func setResources(spec *types.ResourceAllocationInfo, resource types.ResourceAll
 			Level: types.SharesLevelNormal,
 		}
 	}
+	if spec.ExpandableReservation == nil {
+		spec.ExpandableReservation = types.NewBool(true)
+	}
```
matthewavery: Hey emma :) are we good to not set the overhead limit here? that was my only concern. Looks like it is fine since it defaults to `-1` but I was wondering if we should go ahead and explicitly set it to -1 just in case for now. :)

matthewavery:
lgtm - thanks taking this up for me @emlin 😄

lib/install/management/resource_pool.go
emlin: @matthewavery that is the one we cannot set. I thought we should set all of them as you said, but it took me a couple of hour to figure out that this value is wrong, not accepted by VC, although it does not say it clearly.

matthewavery: Oh! interesting! Yeah the api docs say `-1` is used to set things as unbounded for that var... Is there a channel for us to get corrections to the public vsphere API documentation? Because it seems like we should get this fixed in the docs just to add more clarity in general.

emlin: tbh, I'm not sure if that -1 is wrong, but in our combination, if set that to -1, there will have InvalidArgument error throw out.
It might work with other parameters, but I didn't try to figure out all working values.

Issue #6252, opened 2017-09-08
Docker rm -f has a race condition surrounding timing with the Soft Power Off


matthewavery:
**Details:**
`docker rm -f` fails with `cannot remove a running container`. On the second run the container is removed successfully without issue. This appears to be a state race. We are initiating a hard power off when we perceive the soft power off has failed. However, the soft power off does succeed, because of this we actually get an error for the hard power off which causes us to error out. However, this error is that the container is already off(which is what we wanted anyways). so instead of failing we should continue with the removal. 

note: a simple workaround is to just run the `docker rm -f` command again(no modification needed). 

**VIC version:**

VIC ENGINE : 1.2_RC5 --> VIC-OVA-RC3   

**Deployment details:**

Any deployed VCH which can create and destroy containers is vulnerable to this bug. It is much easier to reproduce if you perform concurrent creates and rms on 10+ containers. 

**Steps to reproduce:**

customer ran into this by doing 

docker create <image> * 7 in parallel
docker rm -f <cid> * 7 in parallel

**Actual behavior:**

docker rm -f <cid> --> "cannot remove a running container"

**Expected behavior:**

docker rm -f <cid> --> "<cid>" 

**Logs:**

we cannot provide logs for this from the repro with the customer since they reveal some aspects of the customers environment. Reproduction of this issue should not be too terribly difficult, but the state race is somewhat straight forward in analysis.

matthewavery:
For Posterity these are the notes taken onsite during the advent of the bug.

```
For the failure. This was a standard state race. Powering off a container on the hard poweroff. The hard poweroff can fail if... on the soft power off we are successful but it takes long enough that we attempt the hard poweroff. Then during the attempted hard poweroff we get an error stating "the container is already off". But... the error case for this in the code reports "cannot turn off a running container"  in the event of an error. This is a bug in VIC. What we should do is note that the container did reach the target state and then proceed with the removal of the container. Right now the implication is that the rm command will need to be run a second time.
```

matthewavery:
I will be dropping the analysis in two places. Here and in #6236 since both of these issues are one and the same. I will be using logging from #6236 to break down what is happening.

matthewavery:
There are two primary issues encompassed in this scenario. One we are not fully seeing the effect of from the docker cli, but it is present in the logs. it is this log message

matthewavery:
[...]

PR #6254, opened 2017-09-08
[specific ci=1-07-Docker-Stop] Detect VM poweredoff state to avoid hard poweroff during docker stop

### LINKED ISSUE #6236: Failure on docker stop ${container} "Error response from daemon: Conflict error from portlayer... collision of concurrent operations"

chengwang86:
Fixes #6236 

- this PR avoids the subsequent c.poweroff() if the previous `SIGKILL` or `SIGTERM` already bring the containerVM down.

- this PR increases the maximum elapsed time to 2min for the retry logic during `containerStop()` in backend/container.go

- this PR may also help with #6252

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -108,6 +108,8 @@ const (
 	DefaultCPUs = 2
 	// Default timeout to stop a container if not specified in container config
 	DefaultStopTimeout = 10
+
+	maxElapsedTime = 2 * time.Minute
 )
 
 var (
@@ -1246,8 +1248,11 @@ func (c *Container) ContainerStop(name string, seconds *int) error {
 	operation := func() error {
 		return c.containerProxy.Stop(vc, name, seconds, true)
 	}
-	if err := retry.Do(operation, IsConflictError); err != nil {
-		return err
+
+	config := retry.NewBackoffConfig()
+	config.MaxElapsedTime = maxElapsedTime
+	if err := retry.DoWithConfig(operation, IsConflictError, config); err != nil {
+		return ConflictError(err.Error()+" --- Retrying the operation may resolve the issue.")
 	}
 
 	actor := CreateContainerEventActorWithAttributes(vc, map[string]string{})
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -109,6 +109,7 @@ const (
 	// Default timeout to stop a container if not specified in container config
 	DefaultStopTimeout = 10
 
+	// maximum elapsed time for retry
 	maxElapsedTime = 2 * time.Minute
 )
 
@@ -1252,7 +1253,7 @@ func (c *Container) ContainerStop(name string, seconds *int) error {
 	config := retry.NewBackoffConfig()
 	config.MaxElapsedTime = maxElapsedTime
 	if err := retry.DoWithConfig(operation, IsConflictError, config); err != nil {
-		return ConflictError(err.Error()+" --- Retrying the operation may resolve the issue.")
+		return ConflictError(err.Error() + " --- Retrying the operation may resolve the issue.")
 	}
 
 	actor := CreateContainerEventActorWithAttributes(vc, map[string]string{})
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -1253,7 +1253,7 @@ func (c *Container) ContainerStop(name string, seconds *int) error {
 	config := retry.NewBackoffConfig()
 	config.MaxElapsedTime = maxElapsedTime
 	if err := retry.DoWithConfig(operation, IsConflictError, config); err != nil {
-		return ConflictError(err.Error() + " --- Retrying the operation may resolve the issue.")
+		return err
 	}
 
 	actor := CreateContainerEventActorWithAttributes(vc, map[string]string{})
--- a/lib/portlayer/exec/base.go
+++ b/lib/portlayer/exec/base.go
@@ -34,7 +34,11 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/vm"
 
 	log "github.com/Sirupsen/logrus"
+	"strings"
 )
+
+// The full string should be "The attempted operation cannot be performed in the current state (Powered off)"
+const opNotPerformedPoweredOff = "current state (Powered off)"
 
 // NotYetExistError is returned when a call that requires a VM exist is made
 type NotYetExistError struct {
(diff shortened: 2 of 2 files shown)


chengwang86:
Interesting behavior after running this script on a vc cluster 
```
for i in {1..10}
do
    docker -H ${vch}:2375 run -d --name a$i busybox /bin/top
done

for i in {1..10}
do
    docker -H ${vch}:2375 stop a$i &
done
```
For one of these containers, I find 
PL.log
```
Sep  8 2017 19:51:52.822Z WARN  sending kill -KILL 5ef7f21215473902266d3a9de784676f6dd942249303551225acf0448a5d6d38: ServerFaultCode: The attempted operation cannot be performed in the current state (Powered off).
Sep  8 2017 19:51:52.896Z DEBUG Set container 5ef7f21215473902266d3a9de784676f6dd942249303551225acf0448a5d6d38 state: Stopped
Sep  8 2017 19:51:52.896Z DEBUG Container(5ef7f21215473902266d3a9de784676f6dd942249303551225acf0448a5d6d38) closing 0 log followers
Sep  8 2017 19:51:52.896Z DEBUG [ END ] [github.com/vmware/vic/lib/portlayer/exec.(*Container).stop:419] [14.568767095s] 5ef7f21215473902266d3a9de784676f6dd942249303551225acf0448a5d6d38
Sep  8 2017 19:51:52.896Z DEBUG [BEGIN] [github.com/vmware/vic/lib/portlayer/exec.(*containerBase).updates:119] 5ef7f21215473902266d3a9de784676f6dd942249303551225acf0448a5d6d38
...
Sep  8 2017 19:51:56.597Z DEBUG [ END ] [github.com/vmware/vic/lib/portlayer/exec.(*containerBase).updates:119] [3.701285691s] 5ef7f21215473902266d3a9de784676f6dd942249303551225acf0448a5d6d38
Sep  8 2017 19:51:56.597Z INFO  Reconfigure: attempting update to 5ef7f21215473902266d3a9de784676f6dd942249303551225acf0448a5d6d38 with change version "2017-09-08T19:47:42.747786Z" (poweredOff)
...
Sep  8 2017 19:51:57.269Z INFO  Reconfigure: committed update to 5ef7f21215473902266d3a9de784676f6dd942249303551225acf0448a5d6d38 with change version: 2017-09-08T19:47:42.747786Z
```
At the persona side, no retry was done. Seems like avoiding the extra hard poweroff solved the problem!

lib/apiservers/engine/backends/container.go
```diff
@@ -1246,8 +1249,11 @@ func (c *Container) ContainerStop(name string, seconds *int) error {
 	operation := func() error {
 		return c.containerProxy.Stop(vc, name, seconds, true)
 	}
-	if err := retry.Do(operation, IsConflictError); err != nil {
-		return err
+
+	config := retry.NewBackoffConfig()
+	config.MaxElapsedTime = maxElapsedTime
+	if err := retry.DoWithConfig(operation, IsConflictError, config); err != nil {
+		return ConflictError(err.Error() + " --- Retrying the operation may resolve the issue.")
```
hmahmood: This may not be a `ConflictError`, so better just to leave it as it it was before.

lib/portlayer/exec/base.go
```diff
@@ -36,6 +37,9 @@ import (
 	log "github.com/Sirupsen/logrus"
 )
 
+// The full string should be "The attempted operation cannot be performed in the current state (Powered off)"
+const opNotPerformedPoweredOff = "current state (Powered off)"
```
hmahmood: You should use the full string. Is there no way to check the actual type and content of the error?

lib/apiservers/engine/backends/container.go
chengwang86: Do you mean `return err`?

hmahmood: Yes.

lib/portlayer/exec/base.go
```diff
@@ -36,6 +37,9 @@ import (
 	log "github.com/Sirupsen/logrus"
 )
 
+// vSphere error msg when sending the kill signal to a containerVM that is already powered off
+const opNotPerformedPoweredOff = "The attempted operation cannot be performed in the current state (Powered off)"
```
matthewavery: I believe this error message is constructed from lower down, and potentially contains a fault type. Looking.

chengwang86: I think the error msg is constructed in a way similar to this https://github.com/vmware/vic/blob/master/vendor/github.com/vmware/govmomi/simulator/simulator.go#L94. And the msg itself comes from the http response body of `startGuestProgram()`. I suspect if we can find the error definition.

chengwang86: According to @dougm , this is the `invalidPowerState` error.

lib/portlayer/exec/base.go
```diff
@@ -302,6 +309,13 @@ func (c *containerBase) shutdown(ctx context.Context, waitTime *int32) error {
 			// Just warn and proceed to waiting for power state per issue https://github.com/vmware/vic/issues/5803
 			// Description above in function kill()
 			log.Warnf("%s: %s", msg, err)
+
+			// If the error tells us "The attempted operation cannot be performed in the current state (Powered off)" (InvalidPowerState),
+			// we can avoid hard poweroff (issues #6236 and #6252). Here we wait for the power state changes instead of return
+			// immediately to avoid excess vSphere queries
+			if isInvalidPowerStateError(err) {
+				killed = true
```
hmahmood: Shouldn't we just return here? Why should we move on to call `WaitForPowerState`?

chengwang86: When I tried `docker stop` on 10 containers in parallel, I found that the container with the `InvalidPowerState` error retried `containerProxy.stop()` several times within 8 sec. Actually more than half of the containers (out of 10) suffered from the same invalidPowerstate error. So we had too many vsphere queries. And this would be normal in production if the customer starts/stops many containers in parallel. Therefore, instead of returning immediately, I want to waitForPowerState() here to avoid the excess vsphere queries.  The performance will not be worse than the case without my code change.

lib/portlayer/exec/base.go
matthewavery: Yeah I saw that as well, looks like doug beat me to it :) thanks @chengwang86

lib/portlayer/exec/base.go
sflxn: Doesn't waitForPowerState() also issue a vsphere query?

chengwang86: @sflxn Yes `waitforPowerState()` is a vsphere query. In my local test, I find that `returning nil` without `WaitForPowerState()` results in way too many vsphere requests (due to the retry operations at the persona side) within several seconds, whereas we could've avoid these requests by calling `WaitForPowerState()`.

PR #6265, opened 2017-09-10
Adds basic name convention support [full ci]

hickeng:
Adds redo logic for concurrent modifications to the rename path
Adds very basic naming convention mechanism

Does not add tests or reconfigure support for the new option.

Updates needed in priority order:
- [ ] only apply to the vsphere displayName, not the container name
- [ ] provide `{id}` (shortID) as a token as well as `{name}` - name could be omitted in initial support
- [ ] support reconfigure of convention
- [ ] convention should persist over docker renames
- [ ] reserve name on the create path

End goal is:
vSphere display name like `cVM_prod-short_id`, from a convention template of `cVM_prod-{id}`
container name as normal, like `random_name`
`docker rename` does not impact vsphere display name unless `{name}` is used in convention

### DIFF
--- a/tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
+++ b/tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
@@ -15,7 +15,7 @@
 *** Settings ***
 Documentation  Test 11-03 - Upgrade-InsecureRegistry
 Resource  ../../resources/Util.robot
-Test Teardown  Cleanup Test Environment
+#Test Teardown  Cleanup Test Environment
 
 *** Variables ***
 ${test_vic_version}  7315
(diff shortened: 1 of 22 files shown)


tests/test-cases/Group11-Upgrade/11-03-Upgrade-InsecureRegistry.robot
```diff
@@ -113,6 +114,7 @@ Upgrade VCH with Harbor On HTTP
     Test VCH And Registry  %{VCH-IP}:%{VCH-PORT}  ${harbor_ip}
 
 Upgrade VCH with Harbor On HTTPS
+    Pass Execution  Need to use a different insecure registry, because Harbor does not support VIC as insecure and 0.5.0 is too old
     ${out}=  Cleanup Harbor  ${http_harbor_name}
     Log  ${out}
     ${out}=  Cleanup Harbor  ${https_harbor_name}
```
mhagen-vmware: bed merge?  This is in master already..

mhagen-vmware:
I think this can just be closed in favor of #6284

chengwang86:
@hickeng @cgtexmex    Is this still needed? A big chunk of duplicates with #6284

mhagen-vmware:
closing in favor of #6284 as per email from @hickeng

Issue #6281, opened 2017-09-12
Attach sessions hang (both attach and exec for new and existing containers) 


hickeng:
RemoveInteraction is being called and is triggering the lazy initialization:
```
1 @ 0x42e58a 0x42e66e 0x406b51 0x406785 0x8970c2 0x896e24 0xd0141d 0xd03a44 0xd02579 0xcfed20 0xd00962 0x45b641
#	0x8970c1	github.com/vmware/vic/vendor/golang.org/x/crypto/ssh.(*mux).openChannel+0x1f1				/home/vagrant/vicsmb/src/github.com/vmware/vic/vendor/golang.org/x/crypto/ssh/mux.go:322
#	0x896e23	github.com/vmware/vic/vendor/golang.org/x/crypto/ssh.(*mux).OpenChannel+0x63				/home/vagrant/vicsmb/src/github.com/vmware/vic/vendor/golang.org/x/crypto/ssh/mux.go:298
#	0xd0141c	github.com/vmware/vic/lib/portlayer/attach/communication.NewSSHInteraction+0x10c			/home/vagrant/vicsmb/src/github.com/vmware/vic/lib/portlayer/attach/communication/interactor.go:113
#	0xd03a43	github.com/vmware/vic/lib/portlayer/attach/communication.(*Connector).ids.func1+0xe3			/home/vagrant/vicsmb/src/github.com/vmware/vic/lib/portlayer/attach/communication/connector.go:385
#	0xd02578	github.com/vmware/vic/lib/portlayer/attach/communication.(*LazySessionInteractor).Initialize+0xa8	/home/vagrant/vicsmb/src/github.com/vmware/vic/lib/portlayer/attach/communication/lazy.go:47
#	0xcfed1f	github.com/vmware/vic/lib/portlayer/attach/communication.(*Connector).RemoveInteraction+0x1cf		/home/vagrant/vicsmb/src/github.com/vmware/vic/lib/portlayer/attach/communication/connector.go:201
#	0xd00961	github.com/vmware/vic/lib/portlayer/attach/communication.(*Connector).reqs+0x761			/home/vagrant/vicsmb/src/github.com/vmware/vic/lib/portlayer/attach/communication/connector.go:457
```

Will need confirming, but I think the removal was triggered by the VM stopping or being deleted/dropping vSPC connection.
We should not:
a. be calling initialization on removal - if it's not initialized then we don't need to close the connection
b. calling any potentiall blocking or remote calls while holding a lock

[...]

PR #6283, opened 2017-09-12
Fix for attach hang during RemoveInteraction

hickeng:
This changes the logic on the RemoveInteraction path to:
1. avoid initializing an uninitialized connection solely so it can be Closed
2. moves the potentially complex operation outside of the mutex code
that is only intended to protect access to the map.

This is missing:
1. knowledge of why the mux blocked indefinitely.
2. tests that would exercise this path reliably

Towards: #6281

### DIFF
--- a/lib/portlayer/attach/communication/lazy.go
+++ b/lib/portlayer/attach/communication/lazy.go
@@ -50,3 +50,11 @@ func (l *LazySessionInteractor) Initialize() (SessionInteractor, error) {
 	}
 	return l.si, nil
 }
+
+// Cleanup returns either an initialized connection, or nil if it was never initialized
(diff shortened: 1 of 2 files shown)


lib/portlayer/attach/communication/lazy.go
```diff
@@ -50,3 +50,11 @@ func (l *LazySessionInteractor) Initialize() (SessionInteractor, error) {
 	}
 	return l.si, nil
 }
+
+// Cleanup returns either an initialized connection, or nil if it was never initialized
+func (l *LazySessionInteractor) Cleanup() (SessionInteractor, error) {
```
dougm: Maybe want to name this 'conn' or 'session'?  Doesn't look like it needs to be exported either.

cgtexmex:
LGTM

Issue #6267, opened 2017-09-11
Support templated naming convention for containerVMs


hickeng:
**Story**
As a viadmin I want to be able to enforce a naming convention for cVMs to support chargeback and other organization requirements

**Detail**
Container naming convention applies to the vSphere VM displayName, not to the container name itself.

hickeng:
#6265 is a very early and not formally tested PR with basic naming convention support. The only replaceable token is `{name}` and that will be replace with either:
a. the name provided with `docker create`
b. a generated name as when `--name` is omitted from `docker create`

hickeng:
@stuclem 
A very basic form of this has been merged for engine 1.2.1 to address an urgent need. #6332 is open to polish and complete the work and has more detail than this issue, so closing in favour.
Current mechanism adds the `--cnc` or `--container-name-convention` option. This takes a name template of the form `some_fixed_pattern_{replace-with-value}-static-suffix` - basically just a static string conforming to existing VM naming restrictions with some replaceable tokens. Current tokens are:
* `{name}` - the docker pretty name for the container (either generated or supplied by `docker run --name=some-name`)
* `{id}` - the shortID of the container

This convention applies _only_ to the vSphere display name of the VM and does not impact the names known and shown by docker.

hickeng:
As noted but not done in previous comment, closing in favour of #6332

PR #6284, opened 2017-09-12
Enforce container naming convention specified in appliance config

hickeng:
This is #6265 but applied to master. @cgtexmex contribued a chunk of this
code and will be picking it up from here.

Provides basic container naming convention via {id} and {name} tokens.
The naming is enforced only on vSphere and will not be visible via
the docker client.

Adds retry logic in rename call for ConflictError case

Missing:
1. tests
2. checks on the vSAN impact of changing the displayName like this
(might have to be non-vSAN only).
3. documentation

Towards: #6267

### DIFF
--- a/cmd/vic-machine/common/container.go
+++ b/cmd/vic-machine/common/container.go
@@ -1,0 +1,41 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package common
+
+import (
+	"gopkg.in/urfave/cli.v1"
+)
+
+type ContainerConfig struct {
+	// NameConvention
+	ContainerNameConvention string `cmd:"container-name-convention"`
+}
+
+func (c *ContainerConfig) ContainerFlags() []cli.Flag {
+	return []cli.Flag{
+		// container naming convention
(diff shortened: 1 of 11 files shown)


chengwang86:
Overall LGTM. Just a couple of places where a little bit elaboration is needed:

1. why do we decide to only allow {name} or {id} but now allow for {name}-{id}? Is it because of the additional complexity or DBS restrictions/requirements?

2. why do we only use the template when the nameConvention is {name}? The case where the template has `{id}` is not quite straight-forward. If I do `docker rename newName` while the `--container-name-convention` is `MyOrg-{id}`. Then the `newName` will not be used if this PR is merged. Is this the reason why we obsolete `{id}`? https://github.com/vmware/vic/pull/6284/files#diff-68bcd1ace5fa1fc82a91d9d6852ae463R153

cmd/vic-machine/common/container.go
```diff
@@ -0,0 +1,41 @@
...
+	ContainerNameConvention string `cmd:"container-name-convention"`
+}
+
+func (c *ContainerConfig) ContainerFlags() []cli.Flag {
+	return []cli.Flag{
+		// container naming convention
+		cli.StringFlag{
+			Name:        "container-name-convention, cnc",
+			Value:       "",
+			Usage:       "Provide a naming convention. Allows a token of '{name}' or '{id}', that will be replaced.",
```
chengwang86: Can we make it clear to the user that this option only changes the `display name` of a containerVM on vsphere. So if `docker ps` will still show the pretty name of a containerVM as usual.

cgtexmex: Yes..I think that's needed, but let's do it in a future iteration.

cgtexmex:
@chengwang86 here are my thoughts on your questions - @hickeng can add his POV:

1.  First to avoid complexity -- start simple and grow as needed.  Secondly, {name}-{id} is essentially what we have as the default naming convention. 
2.  In a rename of course the {name} changes, so we apply that to vSphere as well.  In the case of rename with {id} as the convention...nothing about the vSphere display name changes.  The user provided template is static and the ID is immutable, so there's nothing to change on vSphere.

HTH..

cgtexmex:
LGTM

PR #6290, opened 2017-09-12
Reduce WAN check timeout

gigawhitlocks:
Reduce the WAN check in vicadmin timeout to 5 seconds so that the page will load quickly regardless of whether or not the VCH has extranet access.

### DIFF
--- a/lib/vicadmin/validate.go
+++ b/lib/vicadmin/validate.go
@@ -26,6 +26,7 @@ import (
 	"path"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


lib/vicadmin/validate.go
```diff
@@ -186,7 +187,7 @@ func NewValidator(ctx context.Context, vch *config.VirtualContainerHostConfigSpe
 	nwErrors := []error{}
 
 	// create a http client with a custom transport using the proxy from env vars
-	client := &http.Client{Timeout: 30 * time.Second}
+	client := &http.Client{Timeout: 10 * time.Second}
```
chengwang86: So `10 sec` or `5 sec`? I thought you wanted to reduce the timeout to 5 sec

gigawhitlocks: Talked with @hmahmood and settled on 10, @chengwang86

gigawhitlocks: Sorry, didn't update the original description after the first commit ;P

PR #6291, opened 2017-09-12
return error if resource pool creation failed

### LINKED ISSUE #6272: support use-rp option in vic-machine create/inspect/configure/upgrade
@hickeng as your request, we'll need to support --use-rp in create/configure/delete. We already support it in create/delete with one recent bug fix #6250 

But the problem is that how to support it in configure command.
I still didn't find one way to convert vApp to Resource Pool or vise versa. So if we want to update VCH configuration from using vApp to Resourse Pool, we'll need to create new resource pool, and move VCH together with containers to the new resource pool.

Feels like it's possible, but what if anything error happens in the middle, I didn't see good protection way for that level's operation.
So I'm wondering what's the use case to solve through this reconfiguration.

Other issues needs to be solved related to --use-rp
- [ ] if create with --use-rp and --force, vic-machine will ignore resource pool creation failure, and continue to deploy VCH in parent resource pool.
The problem is that this VCH is not recognized by vic-machine anymore, cause the resource pool convention is broken. To fix this problem, we should fail deployment even --force is specified.
- [ ] add CI test for --use-rp in create/upgrade/delete, to avoid regression

emlin:
fixes #6272

### DIFF
--- a/lib/install/management/create.go
+++ b/lib/install/management/create.go
@@ -87,14 +87,7 @@ func (d *Dispatcher) createPool(conf *config.VirtualContainerHostConfigSpec, set
 	} else {
 		if d.vchPool, err = d.createResourcePool(conf, settings); err != nil {
 			detail := fmt.Sprintf("Creating resource pool failed: %s", err)
-			if !d.force {
-				return errors.New(detail)
-			}
-
-			log.Error(detail)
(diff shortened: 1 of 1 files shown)


matthewavery:
lgtm - thanks @emlin

Issue #6033, opened 2017-08-14
Update core vic-machine code to allow authentication using a session key, session cookie, or SAML token


zjs:
Update core `vic-machine` code to allow authentication using a SAML token (preferably) or some other session credential available to the vSphere H5 client (such as a session key or session cookie) in place of a username and password.

Eventually, this will be exposed by the API as well (#6034).

zjs:
To capture some additional information:

 * Based on information provided by @jooskim, it looks like we have access to a SAML Holder-of-Key Token, but _not_ the corresponding private key. This may rule out SAML authentication.
 * We do have access to a username, session key, and session cookie. Based on information provided by @dougm, it sounds like this is similar to what we get back when we log in using the username and password. Therefore, we should be able to use these in place of a username and password when interacting with vSphere on the user's behalf.

**However**, we don't have anything that we can persist when creating the VCH so that _it_ can interact with vSphere (the session key/cookie will expire, at which point the VCH would stop working, which would be... bad). One option is to require that the user always supply operations user credentials when creating a VCH via the API (even if those credentials are for the user's own account). This is a decent _short term_ option, so that the rest of this Epic is not blocked. Much better would be to create a minimally-privileged service account for the VCH as a part of its creation (which is tracked by #635, which @hickeng notes may be very out-of-date at this point).

I'd propose re-purposing this issue to track gathering additional information, proposing updates to the design document, and then reflecting those changes in GitHub (in this Epic and/or #635, as appropriate).

jonathanibers:
So investigating here and #635, I had some questions about this story.

According to #2875, vic-machine currently has support for a seperate user, specified by the --ops-user and --ops-password options.  Looking at the code, it's used in create, and if not specified, vic-machine uses the user/pass from --username and --password.

There is documentation about using --ops-user [https://github.com/vmware/vic/blob/08-docs/doc/user_doc/vic_installation/vch_installer_examples.md#ops-user](link), and vic-machine-linux already accepts an ops-user.   By the documentation, there is a process to create this ops user with fewer priviledges.

But it looks like the ops-user account has to be set before creation, and it seems like the only way to create it is through the UI in vsphere.  Do we want this story to be about creating this ops account in vic-machine, and / or have this created automatically when creating a VCH?  So when the admin calls vic-machine create, it prompts for a user/password to be used for that VCH that becomes our ops user?

Thanks for any help!

@zjs @hickeng

zjs:
[...]

Issue #6032, opened 2017-08-14
Refactor API handlers and vic-machine code to avoid duplication between CLI and API


zjs:
Refactor vic-machine and API handlers code to avoid duplication between CLI and API

zjs:
@hickeng has requested to be included on the design review for this work: https://github.com/vmware/vic/pull/7323#discussion_r168449634

zjs:
- [ ] If this work includes making a manifest the primary means of defining a VCH configuration, consider https://github.com/vmware/vic/pull/7323#discussion_r168451137. (If not, ensure a separate issue is opened to track that eventual work, referencing this comment.)

zjs:
Because this involves de-duplicating code, it should effectively increase test coverage; we'll be able to rely on _both_ API and CLI tests to verify the behavior of shared code. As such, it dovetails nicely with our Quality efforts.

zjs:
Related: #7840

renmaosheng:
moving to 'not ready' for further triage in VIC 1.5

zjs:
> moving to 'not ready' for further triage in VIC 1.5
Is there specific information missing from this Epic which should be added before moving it to the backlog?

renmaosheng:
> > moving to 'not ready' for further triage in VIC 1.5
> 
> Is there specific information missing from this Epic which should be added before moving it to the backlog?
No. As several Epics are not in the list from James, so I would like to have further triage for those Epics for VIC 1.5.

PR #4847, opened 2017-04-21
[specific ci=Group11-Upgrade] Add UpgradeInProgress flag for VCH upgrade

chengwang86:
Fixes #4069 

This PR adds a flag `UpgradeInProgress` to the VCH's ExtraConfig. 

- during VCH upgrade, if `UpgradeInProgress==true`, the current upgrade operation will fail since there might be another upgrade operation running.

- I also add `vic-machine upgrade --resetInProgressFlag` so that the user could reset the flag to `false` in case the flag is polluted by previous operations that do not finish properly.

- test cases are added

PR #6318, opened 2017-09-13
[skip ci] Define REST API handlers for inspecting VCHs

### LINKED ISSUE #6030: Implement unit tests for the API handler for inspecting a VCH
Implement unit tests for the API handler for inspecting a VCH (#6029)

zjs:
Introduce a pair of handlers for accessing information about the VCH with a given identifier within a vSphere target or datacenter.
    
There are ways in which this work is incomplete, marked by `TODO`:
 - Malformed managed object references are ignored
 - Malformed or unexpected PKI material is ignored

I think these are acceptable for now. We can address proper handling of invalid data returned by lower layers later, in a holistic way.

Resolves #6030.

### DIFF
--- a/lib/apiservers/service/restapi/configure_vic_machine.go
+++ b/lib/apiservers/service/restapi/configure_vic_machine.go
@@ -76,9 +76,7 @@ func configureAPI(api *operations.VicMachineAPI) http.Handler {
 	})
 
 	// GET /container/target/{target}/vch/{vch-id}
-	api.GetTargetTargetVchVchIDHandler = operations.GetTargetTargetVchVchIDHandlerFunc(func(params operations.GetTargetTargetVchVchIDParams, principal interface{}) middleware.Responder {
-		return middleware.NotImplemented("operation .GetTargetTargetVchVchID has not yet been implemented")
-	})
+	api.GetTargetTargetVchVchIDHandler = &handlers.VCHGet{}
 
 	// PUT /container/target/{target}/vch/{vch-id}
 	api.PutTargetTargetVchVchIDHandler = operations.PutTargetTargetVchVchIDHandlerFunc(func(params operations.PutTargetTargetVchVchIDParams, principal interface{}) middleware.Responder {
@@ -114,9 +112,7 @@ func configureAPI(api *operations.VicMachineAPI) http.Handler {
 	})
 
 	// GET /container/target/{target}/datacenter/{datacenter}/vch/{vch-id}
-	api.GetTargetTargetDatacenterDatacenterVchVchIDHandler = operations.GetTargetTargetDatacenterDatacenterVchVchIDHandlerFunc(func(params operations.GetTargetTargetDatacenterDatacenterVchVchIDParams, principal interface{}) middleware.Responder {
-		return middleware.NotImplemented("operation .GetTargetTargetDatacenterDatacenterVchVchID has not yet been implemented")
-	})
+	api.GetTargetTargetDatacenterDatacenterVchVchIDHandler = &handlers.VCHDatacenterGet{}
 
 	// PUT /container/target/{target}/datacenter/{datacenter}/vch/{vch-id}
 	api.PutTargetTargetDatacenterDatacenterVchVchIDHandler = operations.PutTargetTargetDatacenterDatacenterVchVchIDHandlerFunc(func(params operations.PutTargetTargetDatacenterDatacenterVchVchIDParams, principal interface{}) middleware.Responder {
--- a/lib/apiservers/service/restapi/handlers/common.go
+++ b/lib/apiservers/service/restapi/handlers/common.go
@@ -1,0 +1,124 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package handlers
+
+import (
+	"context"
+	"fmt"
+	"net/url"
+
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/cmd/vic-machine/common"
+	"github.com/vmware/vic/lib/apiservers/service/restapi/handlers/util"
+	"github.com/vmware/vic/lib/install/data"
+	"github.com/vmware/vic/lib/install/validate"
+	"github.com/vmware/vic/pkg/version"
+	"github.com/vmware/vic/pkg/vsphere/vm"
+)
+
+func buildData(ctx context.Context, url url.URL, user string, pass string, thumbprint *string, datacenter *string, computeResource *string) (*data.Data, error) {
+	d := data.Data{
+		Target: &common.Target{
+			URL:      &url,
+			User:     user,
+			Password: &pass,
+		},
+	}
+
+	if thumbprint != nil {
+		d.Thumbprint = *thumbprint
+	}
+
+	if datacenter != nil {
+		validator, err := validateTarget(ctx, &d)
+		if err != nil {
+			return nil, util.WrapError(500, err)
+		}
+
+		datacenterManagedObjectReference := types.ManagedObjectReference{Type: "Datacenter", Value: *datacenter}
+
+		datacenterObject, err := validator.Session.Finder.ObjectReference(ctx, datacenterManagedObjectReference)
+		if err != nil {
+			return nil, util.WrapError(500, err)
+		}
+
+		d.Target.URL.Path = datacenterObject.(*object.Datacenter).InventoryPath
+	}
+
+	if computeResource != nil {
+		d.ComputeResourcePath = *computeResource
+	}
+
+	return &d, nil
+}
+
+func validateTarget(ctx context.Context, d *data.Data) (*validate.Validator, error) {
+	if err := d.HasCredentials(); err != nil {
+		return nil, fmt.Errorf("Invalid Credentials: %s", err)
+	}
+
+	validator, err := validate.NewValidator(ctx, d)
+	if err != nil {
+		return nil, fmt.Errorf("Validation Error: %s", err)
+	}
+	// If dc is not set, and multiple datacenter is available, vic-machine ls will list VCHs under all datacenters.
+	validator.AllowEmptyDC()
+
+	_, err = validator.ValidateTarget(ctx, d)
+	if err != nil {
+		return nil, fmt.Errorf("Target validation failed: %s", err)
+	}
+	_, err = validator.ValidateCompute(ctx, d, false)
+	if err != nil {
+		return nil, fmt.Errorf("Compute resource validation failed: %s", err)
+	}
+
+	return validator, nil
+}
+
+// Copied from list.go, and appears to be present other places. TODO: deduplicate
+func upgradeStatusMessage(ctx context.Context, vch *vm.VirtualMachine, installerVer *version.Build, vchVer *version.Build) string {
+	if sameVer := installerVer.Equal(vchVer); sameVer {
+		return "Up to date"
+	}
+
+	upgrading, err := vch.VCHUpdateStatus(ctx)
+	if err != nil {
+		return fmt.Sprintf("Unknown: %s", err)
+	}
+	if upgrading {
+		return "Upgrade in progress"
+	}
+
+	canUpgrade, err := installerVer.IsNewer(vchVer)
+	if err != nil {
+		return fmt.Sprintf("Unknown: %s", err)
+	}
+	if canUpgrade {
+		return fmt.Sprintf("Upgradeable to %s", installerVer.ShortVersion())
+	}
+
+	oldInstaller, err := installerVer.IsOlder(vchVer)
+	if err != nil {
+		return fmt.Sprintf("Unknown: %s", err)
+	}
+	if oldInstaller {
+		return fmt.Sprintf("VCH has newer version")
+	}
+
+	// can't get here
+	return "Invalid upgrade status"
+}
--- a/lib/apiservers/service/restapi/handlers/vch_get.go
+++ b/lib/apiservers/service/restapi/handlers/vch_get.go
@@ -1,0 +1,428 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package handlers
+
+import (
+	"bytes"
+	"context"
+	"encoding/pem"
+	"fmt"
+	"net"
+	"net/url"
+	"strings"
+
+	"github.com/docker/docker/opts"
+	"github.com/go-openapi/runtime/middleware"
+	"github.com/go-openapi/strfmt"
+
+	"github.com/vmware/govmomi/vim25/types"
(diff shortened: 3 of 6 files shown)


jzt:
Lots of these files are in need of copyright headers.

jzt:
@zjs I was thinking back on the portlayer when looking at this and wondering if we couldn't also benefit here from a refactor in order to organize the handlers into discrete chunks. Its a little different here since the portlayer is broken into each of its different component handlers (containers, images, storage, etc), but we could perhaps set `configure_vic_machine.go` up in a similar scheme. We could have a list of handlers [like this](https://github.com/vmware/vic/blob/master/lib/apiservers/portlayer/restapi/configure_port_layer.go#L48) and basically call [configure](https://github.com/vmware/vic/blob/master/lib/apiservers/portlayer/restapi/configure_port_layer.go#L133) for each one. That way maybe we could have a much cleaner `configure_vic_machine.go` with a collection of files by operation, like:  `vch_get.go`, `vch_list.go`, `vch_create.go`, etc.

Probably not necessary now but maybe if we start thinking about it we can avoid a big refactor later on down the line.

zjs:
> We could have a list of handlers like this and basically call configure for each one.

When I started writing `configure_vic_machine.go`, I noticed that there seemed to be multiple approaches used by different projects. I followed the pattern Ivan recommended.

The impression I got was that the portlayer approach ended up being unnecessarily complex; it makes the `configure_*.go` file a bit shorter, but it's much less clear which handler will be invoked for a given API call since that information is now spread out across all of the `Configure` methods.

I'm not attached to any specific pattern, and there may be advantages to the portlayer approach I'm not aware of.

lib/apiservers/service/restapi/configure_vic_machine.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// POST /container/target/{target}/datacenter/{datacenter}
+	api.PostTargetTargetDatacenterDatacenterHandler = operations.PostTargetTargetDatacenterDatacenterHandlerFunc(func(params operations.PostTargetTargetDatacenterDatacenterParams, principal interface{}) middleware.Responder {
+		return middleware.NotImplemented("operation .PostTargetTargetDatacenterDatacenter has not yet been implemented")
+	})
+
+	// GET /container/target/{target}/datacenter/{datacenter}/vch
+	api.GetTargetTargetDatacenterDatacenterVchHandler = &handlers.VCHDatacenterListGet{}
+
+	// POST /container/target/{target}/datacenter/{datacenter}/vch
```
jzt: Minor, but maybe we could edit the spec such that `/container/target/{target}/datacenter/{datacenter}/vch` becomes something like `/container/target/{address}/datacenter/{name}/vch` so we end up with method names like `api.GetTargetAddressDatacenterNameVchHandler` instead of `api.GetTargetTargetDatacenterDatacenterVchHandler`.

Again, pretty minor - not worth the time spent on it if you don't see it as a long term improvement.

zjs: I really don't like the method names, and am open to changing the spec.

I worry that including vague things like `{name}` or `{id}` in the path might limit us in the future. If we start with a path like `/foo/{name}` and then want to add a sub-resource `bar`, we can't have `/foo/{name}/bar/{name}`, so we're stuck being inconsistent (with something like `/foo/{name}/bar/{barName}`) or refactoring  (to something like `/foo/{fooName}/bar/{barName}`) and likely breaking clients using generated bindings.

To avoid painting ourselves into a corner like that, I think we'd want to qualify any generally-applicable parameters up front. (So `{address}` is probably okay as-is, but `{name}` or `{id}` should probably be qualified.)

I also think readability of the generated documentation is at least as important as readability of the generated go code, so we should keep that in mind as we consider changes.

Given that, maybe we should compare both the generated documentation and generate method names for the current style (`/foo/{foo}`) with some potential new styles like `/foo/{fooId}`?

zjs:
> Lots of these files are in need of copyright headers.

This should be fixed across all of my pending PRs.

lib/apiservers/service/restapi/configure_vic_machine.go
jzt: I agree with your concerns around `{name}` and `{id}`, especially in cases where you could supply a name or an id (datacenters/datastores for example?). I like the `/foo/{fooID}` suggestion as `fooID` can imply whatever it is you are using to identify a particular `foo`, and is not restricted to being a name or a UUID or whatever, but can be anything.

jzt:
@zjs once this is merged I will use it to build from when I take one of the tasks from the epic.

lib/apiservers/service/restapi/handlers/vch_get.go
```diff
@@ -0,0 +1,412 @@
...
+		url.URL{Host: params.Target},
+		principal.(Credentials).user,
+		principal.(Credentials).pass,
+		params.Thumbprint,
+		nil,
+		nil)
+
+	d.ID = params.VchID
+
+	vch, err := getVCH(params.HTTPRequest.Context(), d)
```
emlin: Now portlayer is moving to trace.NewOperation to track different operation, to help log analysis. So suggest to add that into vic-machine service as well

lib/apiservers/service/restapi/handlers/vch_get.go
```diff
@@ -0,0 +1,412 @@
...
+}
+
+// VCHGet is the handler for inspecting a VCH within a Datacenter
+type VCHDatacenterGet struct {
+}
+
+func (h *VCHGet) Handle(params operations.GetTargetTargetVchVchIDParams, principal interface{}) middleware.Responder {
+	d := buildData(
+		url.URL{Host: params.Target},
+		principal.(Credentials).user,
```
emlin: is wondering how the principal is set. is that the credential to login vic-machine service?

lib/apiservers/service/restapi/handlers/vch_get.go
```diff
@@ -0,0 +1,412 @@
...
+			Reservation: asMHz(d.ResourceLimits.VCHCPUReservationsMHz),
+			Shares:	     asShares(d.ResourceLimits.VCHCPUShares),
+		},
+		Memory: &models.VCHComputeMemory{
+			Limit:       asMiB(d.ResourceLimits.VCHMemoryLimitsMB),
+			Reservation: asMiB(d.ResourceLimits.VCHMemoryReservationsMB),
+			Shares:	     asShares(d.ResourceLimits.VCHMemoryShares),
+		},
+		Resource: &models.ManagedObject{
+			ID: vchConfig.Container.ComputeResources[0].String(), // TODO: This doesn't seem to be an ID?
```
emlin: it's resource pool reference object, *types.ManagedObjectReference

lib/apiservers/service/restapi/handlers/vch_get.go
```diff
@@ -0,0 +1,412 @@
...
+		},
+	}
+
+	// network
+	model.Network = &models.VCHNetwork{
+		Bridge: &models.VCHNetworkBridge{
+			PortGroup: &models.ManagedObject{
+				ID: vchConfig.Network.BridgeNetwork,
```
emlin: did you test it? feels it's not right
you might reference here: https://github.com/vmware/vic/blob/master/lib/install/validate/config_to_data.go#L366
You could also reference the code in the above file for how to read back VCH configuration

lib/apiservers/service/restapi/handlers/vch_get.go
zjs: Currently, we've only implemented basic authentication (where a username and password are passed via an `Authorization` header using the `Basic` scheme: https://tools.ietf.org/html/rfc7617)

Relevant code:
* https://github.com/vmware/vic/blob/feature/vic-machine-service/lib/apiservers/service/restapi/configure_vic_machine.go#L55
* https://github.com/vmware/vic/blob/feature/vic-machine-service/lib/apiservers/service/restapi/handlers/authentication.go

Implementing an additional type of authentication is tracked by #6033

lib/apiservers/service/restapi/handlers/vch_get.go
zjs: Good catch! It looks like it's always returning "bridge", which happened to be the name of my bridge network, so I didn't notice.

lib/apiservers/service/restapi/handlers/vch_get.go
zjs: This seems like a good pattern to follow. I filed a separate issue for it so that we can make sure to fix this across the board: https://github.com/vmware/vic/issues/6374 (several handlers have already been committed).

lib/apiservers/service/restapi/handlers/vch_get.go
```diff
@@ -0,0 +1,428 @@
...
+	"encoding/pem"
+	"fmt"
+	"net"
+	"net/url"
+	"strings"
+
+	"github.com/docker/docker/opts"
+	"github.com/go-openapi/runtime/middleware"
+	"github.com/go-openapi/strfmt"
+	"github.com/vmware/govmomi/vim25/types"
```
jzt: `gofmt` will complain that there should be a space above this line separating vmware imports from external imports.

lib/apiservers/service/restapi/handlers/vch_get.go
```diff
@@ -0,0 +1,428 @@
...
+		params.Thumbprint,
+		nil,
+		nil)
+	if err != nil {
+		return operations.NewGetTargetTargetVchVchIDDefault(util.StatusCode(err)).WithPayload(&models.Error{Message: err.Error()})
+	}
+
+	d.ID = params.VchID
+
+	vch, err := getVCH(params.HTTPRequest.Context(), d)
```
jzt: If we wrap `parmas.HTTPRequest.Context()` in a `trace.NewOperation()` we can chase these calls through the logs easier when debugging:
```go
op := trace.NewOperation(params.HTTPRequest.Context(), "vch: %s", params.VchID)
vch, err := getVCH(op, d)
```

You can then use it to log with `op.Infof()`, `op.Debugf()`, etc.

https://github.com/vmware/vic/blob/a65d66a404b93d7dde1df886e3dbfb79956ebec1/pkg/trace/operation.go#L154

lib/apiservers/service/restapi/handlers/vch_get.go
```diff
@@ -0,0 +1,428 @@
...
+		params.Thumbprint,
+		&params.Datacenter,
+		nil)
+	if err != nil {
+		return operations.NewGetTargetTargetDatacenterDatacenterVchVchIDDefault(util.StatusCode(err)).WithPayload(&models.Error{Message: err.Error()})
+	}
+
+	d.ID = params.VchID
+
+	vch, err := getVCH(params.HTTPRequest.Context(), d)
```
jzt: Same here, re: `trace.Operation`

lib/apiservers/service/restapi/handlers/vch_get.go
```diff
@@ -0,0 +1,428 @@
...
+
+	// endpoint
+	model.Endpoint = &models.VCHEndpoint{
+		UseResourcePool: d.UseRP,
+		Memory:          asMiB(&d.MemoryMB),
+		CPU: &models.VCHEndpointCPU{
+			Sockets: int64(d.NumCPUs),
+		},
+		OperationsCredentials: &models.VCHEndpointOperationsCredentials{
+			User: vchConfig.Connection.Username,
```
jzt: Worth excluding username as well?

lib/apiservers/service/restapi/handlers/vch_get.go
```diff
@@ -0,0 +1,428 @@
...
+	model.Runtime.UpgradeStatus = upgradeStatus
+
+	powerState, err := vch.PowerState(context.Background())
+	if err != nil {
+		powerState = "error"
+	}
+	model.Runtime.PowerState = string(powerState)
+
+	if public := vchConfig.ExecutorConfig.Networks["public"]; public != nil {
+		if public_ip := public.Assigned.IP; public_ip != nil {
```
jzt: s/public_ip/publicIP

lib/apiservers/service/restapi/handlers/vch_get.go
```diff
@@ -0,0 +1,428 @@
...
+			var docker_port string
+			if !vchConfig.HostCertificate.IsNil() {
+				docker_port = fmt.Sprintf("%d", opts.DefaultTLSHTTPPort)
+			} else {
+				docker_port = fmt.Sprintf("%d", opts.DefaultHTTPPort)
```
jzt: s/docker_port/dockerPort

(golint will complain about using underscores in var names)

lib/apiservers/service/restapi/handlers/vch_list_get.go
```diff
@@ -43,15 +41,18 @@ type VCHDatacenterListGet struct {
 }
 
 func (h *VCHListGet) Handle(params operations.GetTargetTargetVchParams, principal interface{}) middleware.Responder {
-	d := buildData(
+	d, err := buildData(params.HTTPRequest.Context(),
```
jzt: Same here, re: `trace.Operation`

jzt:
@zjs Sorry, disregard my comments about `trace.NewOperation`.. just saw [this](https://github.com/vmware/vic/pull/6318#discussion_r139260039).

lib/apiservers/service/restapi/handlers/vch_get.go
zjs: I think including the operations usename is useful, and potentially important.

If the VCH suddenly stops working, being able to look at it and say "Oh, it's using my account, and I just changed the password!" or "Oh, it was using Joe's account, but... we just fired Joe and deactivated his account." might be helpful. (Or, alternatively, to verify that it's using a service account, as expected, and _not_ Joe's account.)

lib/apiservers/service/restapi/handlers/common.go
```diff
@@ -0,0 +1,124 @@
...
+			User:     user,
+			Password: &pass,
+		},
+	}
+
+	if thumbprint != nil {
+		d.Thumbprint = *thumbprint
+	}
+
+	if datacenter != nil {
```
emlin: hmm... is this necessary? saw this line: https://github.com/vmware/vic/pull/6318/files#diff-adbd7d5a586ec7aeba3724c39fc6a462R78

lib/apiservers/service/restapi/handlers/common.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	_, err = validator.ValidateCompute(ctx, d, false)
+	if err != nil {
+		return nil, fmt.Errorf("Compute resource validation failed: %s", err)
+	}
+
+	return validator, nil
+}
+
+// Copied from list.go, and appears to be present other places. TODO: deduplicate
```
emlin: also appears at inspect.go, should be unified

lib/apiservers/service/restapi/handlers/common.go
```diff
@@ -0,0 +1,124 @@
...
+	if sameVer := installerVer.Equal(vchVer); sameVer {
+		return "Up to date"
+	}
+
+	upgrading, err := vch.VCHUpdateStatus(ctx)
+	if err != nil {
+		return fmt.Sprintf("Unknown: %s", err)
+	}
+	if upgrading {
+		return "Upgrade in progress"
```
emlin: it could also be configure in progress, see the method in inspect.go

lib/apiservers/service/restapi/handlers/common.go
zjs: This block translates a datacenter id (what we want the API to use) into a datacenter name (what the rest of the product uses). The `validateTarget` method is called here with an incomplete data object and the `validator` which is returned is only used within this block. (We can't look up the datacenter name if we have an invalid target or credentials.)

In all cases, including when `datacenter` is `nil`, `validateTarget` is called later to validate the full data object returned by this method.

lib/apiservers/service/restapi/handlers/common.go
zjs: I agree this should be unified, but don't think this PR is the right place to do that, which is why I've left this TODO.

lib/apiservers/service/restapi/handlers/common.go
zjs: It sounds like we have a bug then; this method is copied as-is from `list.go`, which wasn't updated in #4847 when the reference to configure was added to `inspect.go`'s version of this method.

We can sort that our when we address the above TODO and de-duplicate the code between `inspect.go`, `list.go`, and `common.go`. At the _latest_, this would happen as a part of #6032 (which will be resolved before this feature branch is merged to `master`).

lib/apiservers/service/restapi/handlers/common.go
emlin: saw the todo, just add one comment for another duplication.

lib/apiservers/service/restapi/handlers/common.go
emlin: sure

PR #6320, opened 2017-09-14
[specific ci=1-00-Docker-Commands] Improve docker commands CI run time [WIP]

sflxn:
Moved docker command tests to subtests folder and make tests
robot keywords.  This should remove the need to create a VCH
for each test.


mhagen-vmware:
@sflxn and @hickeng I am a bit concerned with the path that this PR looks like it is heading down.  In order to accomplish our goal, I think we really should be just focusing on parameterizing install and cleanup of a VCH in VCH-Util.robot to allow for gathering logs without without deleting the VCH and if %{VCH-NAME} env variable exists, then use the existing VCH.  This would minimize the changes to our tests ( you would be able to just change the tear down and setup of each test ) and would allow us to still dynamically run with multiple VCH or a single VCH as desired.

PR #6376, opened 2017-09-16
Fixing failure cases and timeouts for Admiral config update [full ci]

### LINKED ISSUE #6346: docker info should not be unresponsive if admiral is slow responding
**Story**
As a docker user I use `docker info` to check the status of my container host and expect instant response if the host is healthy

**Detail**
We should have a much lower time out on the Admiral integration update during `docker info` path - we may want to dispatch the update in a background thread and report age of the config data instead of fetch it live every time.

We also hit an issue here due to overly broad locking on the `source.discover` path that caused `docker info` to hang indefinitely on acquiring the config lock at the beginning of `source.Get`
https://github.com/vmware/vic/blob/master/lib/config/dynamic/admiral/source.go#L115

**Acceptance**
- [ ] `docker info` should return almost immediately - if linked with an admiral that cannot be contacted then that should be noted in the status along with the last time the configuration was validated.
- [ ] If not linked with an admiral then no link should be established with an unresponsive candidate.
- [ ] The locking for dynamic config should be narrowed so that a stalled update does not hang docker API operations (returning errors about the delay could be okay).

Related: #6348

### LINKED ISSUE #6347: Incorrect construction of target URL string for tag service
The targetURL is being incorrectly constructed with a populated UserInfo structure before the assignment to BasicAuth for the actual request.
Either the credentials should be stored differently or the URL should be sanitized before being passed to NewRequest:
```go
  target := *c.endpoint
  target.User = nil
```

https://github.com/vmware/vic/blob/master/pkg/vsphere/tags/rest_client.go#L160

### LINKED ISSUE #6348: RegistryCheck should have a much lower timeout and failure modes should be different
**Story**
As a user I expect commands such as `docker login`, `docker pull`, and `docker info` (see #6346 ) to respond in a timely fashion

**Details**
We had a scenario where the tag service was completely unresponsive (vcsupport bundle via other means) and caused complete unresponsiveness in the following commands until we hit a timeout:
- login
- pull
- info

The timeout in question appears to be passed in via the docker router code and used without adjustment by us. It's unclear what that value actually is from the logs, but it's on the order of 35-40min looking at the logs. https://github.com/vmware/vic/blob/master/lib/apiservers/engine/backends/backends.go#L362

The timeout for the RegistryCheck needs to be much, much lower, and to return a useful error, if any, to the user - see the behaviours described in Acceptance.
Currently the timeout triggers #6347 (high severity)

Estimate is for reducing timeout only - follow up issues should be opened for changing behaviours.
Priority is high for timeout adjustment.

**Acceptance**
* a VCH that is not already linked with an admiral should not block or delay `login` or `pull` if vCenter tag service cannot be accessed or if an admiral is slow to respond.
* a VCH that is already linked should continue with it's saved configuration and note in `docker info` that the configuration is stale.
* content trust checkbox should inspect the `docker info` from VCHs in the project to confirm acceptance before confirming that content trust is enabled.

Related: #6346

hmahmood:
Fixes #6347 
Fixes #6346 
Fixes #6348

### DIFF
--- a/lib/apiservers/engine/backends/backends.go
+++ b/lib/apiservers/engine/backends/backends.go
@@ -84,6 +84,10 @@ type dynConfig struct {
 
 	Whitelist, Blacklist, Insecure registry.Set
 	remoteWl                       bool
+
+	updateCh      chan bool
+	updateWaiters chan chan *dynConfig
+	lastCfg       *dynConfig
 }
 
 func Init(portLayerAddr, product string, port uint, config *config.VirtualContainerHostConfigSpec) error {
@@ -359,7 +363,48 @@ func EventService() *events.Events {
 // RegistryCheck checkes the given url against the registry whitelist, blacklist, and insecure
 // registries lists. It returns true for each list where u matches that list.
 func (d *dynConfig) RegistryCheck(ctx context.Context, u *url.URL) (wl bool, bl bool, insecure bool) {
-	m := d.update(ctx)
+	var m *dynConfig
+	ch := make(chan *dynConfig, 1)
+
+	select {
+	case d.updateWaiters <- ch:
+		select {
+		case d.updateCh <- true:
+			go func() {
+				ctx, cancel := context.WithTimeout(ctx, 2*time.Minute)
+				defer cancel()
+				m := d.update(ctx)
+				done := false
+				for !done {
+					select {
+					case ch := <-d.updateWaiters:
+						ch <- m
+					case <-time.After(5 * time.Second):
+						done = true
+					}
+				}
+
+				<-d.updateCh
+			}()
+		default:
+		}
+
+		select {
+		case m = <-ch:
+		case <-time.After(5 * time.Second):
+		}
+	default:
+	}
+
+	if m == nil {
+		d.Lock()
+		m = d.lastCfg
+		if m == nil {
+			m = d
+		}
+		d.Unlock()
+	}
+
 	us := u.String()
 	wl = len(m.Whitelist) == 0 || m.Whitelist.Match(us)
 	bl = len(m.Blacklist) == 0 || !m.Blacklist.Match(us)
@@ -401,6 +446,7 @@ func (d *dynConfig) update(ctx context.Context) *dynConfig {
 		}
 	}
 
+	d.lastCfg = m
 	return m
 }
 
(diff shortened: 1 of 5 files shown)


lib/apiservers/engine/backends/backends.go
```diff
@@ -359,7 +363,48 @@ func EventService() *events.Events {
 // RegistryCheck checkes the given url against the registry whitelist, blacklist, and insecure
 // registries lists. It returns true for each list where u matches that list.
 func (d *dynConfig) RegistryCheck(ctx context.Context, u *url.URL) (wl bool, bl bool, insecure bool) {
-	m := d.update(ctx)
+	var m *dynConfig
```
hickeng: Please replace this with singleflight as used here:
https://github.com/vmware/vic/blob/master/lib/portlayer/attach/communication/connector.go#L114

I think the current code is racy (particularly the draining of `d.updateCh` and the associated `case` statement - possible fall through with no one actually doing the work)

lib/config/dynamic/admiral/source.go
```diff
@@ -387,25 +387,14 @@ func (o *productDiscovery) Discover(ctx context.Context, sess *session.Session)
...
 
-		return t, nil
-	}()
-
-	if err != nil {
-		return nil, err
+	o.mu.Lock()
+	t := o.clients[service.String()]
+	if t == nil {
+		t = tags.NewClient(service, sess.Insecure, sess.Thumbprint)
```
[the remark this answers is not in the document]
hmahmood: Yes, `Login` is only called after the first rest api call fails with a 401.

pkg/vsphere/tags/rest_client.go
```diff
@@ -51,8 +54,12 @@ func NewClient(u *url.URL, insecure bool, thumbprint string) *RestClient {
 		sc.SetThumbprint(u.Host, thumbprint)
 	}
 
+	user := u.User
+	u.User = nil
```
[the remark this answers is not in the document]
hmahmood: Changed so that a copy is used now.

lib/apiservers/engine/backends/backends.go
hmahmood: Replaced with singleflight. Also moved this call to `update` since `SystemInfo` was also calling `update`.

pkg/vsphere/tags/rest_client.go
```diff
@@ -185,3 +192,17 @@ func (c *RestClient) Login(ctx context.Context) error {
...
+
+func (c *RestClient) newRequest(method, urlStr string, body io.Reader) (*http.Request, error) {
+	req, err := http.NewRequest(method, c.endpoint.String()+urlStr, body)
+	if err != nil {
+		return nil, err
+	}
+
+	if c.user != nil {
+		password, _ := c.user.Password()
+		req.SetBasicAuth(c.user.Username(), password)
```
dougm: At some point we should look into how we can re-use the sdk session cookie, rather than sending user+pass credentials again.  For example, the sdk cookie name is `vmware_soap_session`, the pbm and vsan API extensions accept the same cookie value for authn, but with a different cookie name: `vcSessionCookie`

cgtexmex:
LGTM

PR #6381, opened 2017-09-17
pkg/vsphere/tags: Add UpdateCategory

vancluever:
Just adding the update (`PATCH`) method for a category. Pretty basic,
sourced from the create method, with new types added to specifically
denote update parameters.

### DIFF
--- a/pkg/vsphere/tags/categories.go
+++ b/pkg/vsphere/tags/categories.go
@@ -33,7 +33,18 @@ type CategoryCreateSpec struct {
 	CreateSpec CategoryCreate `json:"create_spec"`
 }
 
+type CategoryUpdateSpec struct {
+	UpdateSpec CategoryUpdate `json:"update_spec"`
+}
+
 type CategoryCreate struct {
+	AssociableTypes []string `json:"associable_types"`
+	Cardinality     string   `json:"cardinality"`
+	Description     string   `json:"description"`
+	Name            string   `json:"name"`
+}
+
+type CategoryUpdate struct {
 	AssociableTypes []string `json:"associable_types"`
 	Cardinality     string   `json:"cardinality"`
 	Description     string   `json:"description"`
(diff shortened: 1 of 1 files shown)


pkg/vsphere/tags/categories.go
```diff
@@ -33,13 +33,24 @@ type CategoryCreateSpec struct {
...
+}
+
 type CategoryCreate struct {
 	AssociableTypes []string `json:"associable_types"`
 	Cardinality     string   `json:"cardinality"`
 	Description     string   `json:"description"`
 	Name            string   `json:"name"`
 }
 
+type CategoryUpdate struct {
```
hmahmood: Would this work?
```go
type CategoryUpdate CategoryCreate
```

vancluever: Hey @hmahmood - I think you all might know better that me in this respect, but even though the fields are the same, personally I would leave it properly typed and keep the name and the redundancy. They are, technically, separate types all the way down (see `update_spec` doc [here](https://vdc-repo.vmware.com/vmwb-repository/dcr-public/1cd28284-3b72-4885-9e31-d1c6d9e26686/71ef7304-a6c9-43b3-a3cd-868b2c236c81/doc/structures/com/vmware/cis/tagging/category/svc.update_spec-structure.html)), so I think the client should reflect that.

Personally I think things should have been structured like the following, which kind of illustrates my point a bit better:

```go
type CategoryCreateInput struct {
	CreateSpec CategoryCreateSpec `json:"create_spec"`
}

type CategoryUpdateInput struct {
	UpdateSpec CategoryUpdateSpec `json:"update_spec"`
}

type CategoryCreateSpec struct {
	AssociableTypes []string `json:"associable_types"`
	Cardinality     string   `json:"cardinality"`
	Description     string   `json:"description"`
	Name            string   `json:"name"`
}

type CategoryUpdateSpec struct {
	AssociableTypes []string `json:"associable_types"`
	Cardinality     string   `json:"cardinality"`
	Description     string   `json:"description"`
	Name            string   `json:"name"`
}
```

PR #6387, opened 2017-09-18
[skip ci] WIP: live streaming VCH creation log to VCH datastore folder

### LINKED ISSUE #6036: Enhance vic-machine to stream a creation log to the VCH's datastore folder
Enhance `vic-machine` to stream a creation log to the VCH's datastore folder.

As a VCH is created, logs should be written to a file in the VCH's datastore folder.

AngieCris:
work in progress, fixes #6036 

**Goal**
Enable live VCH creation log streaming to VCH datastore folder

**Design details**
A new package called `vchlog` dealing with log streaming is added under `lib/install/`
`vchlog` has a `vchlog.Run()` function that blocks until the VCH datastore folder is available, then it uploads a buffered pipe readwriter (for log messages) to datastore folder via govmomi.
`vchlog` is initialized and the `vchlog.Run()` is called in `main.go` under `cmd/vic-machine` in a go routine.

govmomi upload function spins in a loop reading from the stream until EOF. By the implementation of the buffered pipe within `vchlog`, the read is blocked until a waiting writer in queue is done writing. Thus log messages are concurrently read and uploaded to the datastore while they're being written ("real-time" stream).

**Concerns and problems**

1. Multiple vic-machine processes can act on a single VCH. Need to have a smart way to deal with concurrent reads/writes. As far as I know, the vic-machine currently does not have unique identifier. Need to figure out an identifier for each vic-machine process somehow (like operation ID? timestamp?)
2. We want to enable live log messages streaming to other vic-machine commands as well, not only for VCH creation. Currently the only way to talk to datastore folder is using the `datastore.Upload` function by govmomi's API, and since `datastore.Upload` function does not append, each vic-machine command has to write to a different log file.
3. Not sure if using govmomi would be a performance blocker under extreme use cases. It seems like the only way to talk to VCH datastore folder directly.
4. This design and implementation of live log streaming will affect the impl of `vic-machine-service` (vic-machine API). Later possibly we would like `vic-machine-service` to do the log streaming, and levitate this design to the API endpoint impl. 
This design is not finalized, not sure how long it can go.


AngieCris:
The workflow of VCH creation goes from cmd layer to management layer. Specifically about log streaming, cmd layer sets the log output to streams, and management layer uploads the log stream to VCH datastore folder.
Because of this architecture, any way of live streaming VCH creation log has to involve passing down log stream from cmd to management layer (the create object under lib/install), which is not pretty.

In this implementation:
I implemented an asynchronous buffered readwrite pipe that puts all writes in a buffer, and blocks read until there's a write (basically golang io.Pipe with a buffer)
In cmd layer, set log output to the pipe, pass the pipe down to management layer
Then every log.Infof or the like will be a write to the pipe that gets buffered
In management layer (lib/install/management/create.go), upload the log stream to VCH datastore folder via govmomi (in a go routine, concurrently with the log writes). According to the source of govmomi upload function, it spins in a loop reading from the stream and stops upon EOF. Since the pipe blocks read if no write is available, the streaming is real-time.
(This implementation is in experimental stage - sorry about the trailing whitespaces and commented out dead code)

Current problems:

It's kind of ugly passing down a log stream from cmd to management, and this design will very likely break if we later want the cmd layer to call the vic-machine API instead. But currently, because of the different responsibility assignment of cmd and management layer, it seems like passing down a log stream is the only way. Any better suggestion is appreciated.
Naturally we might want to also enable live log streaming for other vic-machine commands as well. But the design is going to look uglier because we would have to pass the log stream to every one of the command implementation in management layer. To change this, we could abstract all the commands to implement a single interface (or sth along the line to avoid duplicate work) but it's gonna be a lot of refactoring of code.
Continuing on the previous point of enabling live log streaming of other commands, since govmomi upload function overwrites (essentially a http POST request), every vic-machine command has to make a new log file instead of appending to the VCH creation log file.
More than one vic-machine processes could run on a single VCH. To work around this, every log file can be named with an identifier of the vic-machine (like a randomly generated string??), but as far as I know, there's no existing unique identifier for vic-machine yet.
It seems like the only option available to upload to datastore folder is via govmomi. I'm not sure if it's gonna be a performance blocker in extreme situations.
I'm working on figuring out a neater design to enable live log streaming (one that does not involve passing a stream down from CLI to management). The future API design would be an influencing factor too - can this current logging strategy be applied to other log streaming? Is syslog an acceptable workaround?

Any comment/input/suggestion/concern/question is appreciated, thanks!

AngieCris:
**UPDATE**:

Refactored the code
right now it doesn't have to pass down a log stream writer to management layer
a new logger package is added in the management layer. It spins in a loop listening for signal that VCH datastore folder is ready, and start uploading once signal is received.

Still these concerns are not addressed:
1. multiple vic-machine binaries working on the same VCH
2. other vic-machine command log streaming

AngieCris:
**Design details**
a new package dealing with log streaming is added under `lib/install/`
this package has a run function that blocks until the VCH datastore folder is available, then upload a buffered pipe readwriter (for log messages) to datastore folder via govmomi
this package is initialized and the run function is called in `main.go` under `cmd/vic-machine` in a go routine

**Concerns and problems**
1. Multiple vic-machine processes can act on a single VCH. Need to have a smart way to deal with concurrent reads/writes. As far as I know, the vic-machine currently does not have unique identifier. Need to figure out an identifier for each vic-machine process somehow (like a hash value?)
2. We want to enable live log messages streaming to other vic-machine commands as well, not only for VCH creation. Currently the only way to talk to datastore folder is using the `datastore.Upload` function in govmomi's API, and since the `datastore.Upload` function does not append, each vic-machine command has to write to different log file.

lib/install/vchlog/vchlogger.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/vmware/govmomi/object"
+	"context"
+	"path"
+)
+
+
+type VCHCreatedSignal struct {
+	Datastore *object.Datastore
+	LogFileName string
+	Contex context.Context
```
jzt: s/Contex/Context

AngieCris: oops thanks

AngieCris: sry for the uncommented files: the design is not finalized yet

AngieCris:
Closing this issue: base changed to branch `feature/vic-machine-service`

PR #6399, opened 2017-09-20
pkg/vsphere/tags: Add UpdateTag

vancluever:
Similar to `UpdateCategory`, looks like `UpdateTag` was missing too. This is
a simple update to expose this functionality.


dougm:
Due to pullapprove configuration we need 1 more approval since @vancluever isn't in the @vic-maintainers group

vancluever:
Thanks all! :+1:

PR #6403, opened 2017-09-21
[skip ci] live streaming VCH creation log to VCH datastore folder 

### LINKED ISSUE #6036: Enhance vic-machine to stream a creation log to the VCH's datastore folder

AngieCris:
Fixes #6036 

**Goal**
Enable live VCH creation log streaming to VCH datastore folder

**Design details**
A new package called `vchlog` dealing with log streaming is added under `lib/install/`
vchlog has a `vchlog.Run()` function that blocks until the VCH datastore folder is available, then it uploads a buffered pipe readwriter (for log messages) to datastore folder via govmomi.
vchlog is initialized and the `vchlog.Run()` is called in main.go by `cmd/vic-machine` in a go routine.

govmomi upload function spins in a loop reading from the stream until EOF. By the implementation of the buffered pipe within `vchlog`, the read is blocked until a write has done writing (when there's new data available). Thus log messages are concurrently read and uploaded to the datastore while they're being written ("real-time" stream).

**Remaining work**
1. Enable live log streaming for other commands 
Currently only `vic-machine create` can live stream logs to VCH datastore folder. For other commands, this pr is a no-op.
2. Investigate operation ID for each vic-machine operation

Can people please review this PR? Thanks! @zjs @andrewtchin @jzt

### DIFF
--- a/cmd/vic-machine/main.go
+++ b/cmd/vic-machine/main.go
@@ -32,6 +32,7 @@ import (
 	"github.com/vmware/vic/cmd/vic-machine/list"
 	"github.com/vmware/vic/cmd/vic-machine/update"
 	"github.com/vmware/vic/cmd/vic-machine/upgrade"
+	"github.com/vmware/vic/lib/install/vchlog"
 	viclog "github.com/vmware/vic/pkg/log"
 	"github.com/vmware/vic/pkg/version"
 )
@@ -138,6 +139,12 @@ func main() {
 		defer f.Close()
 		logs = append(logs, f)
 	}
+
+	// create the logger for streaming VCH log messages
+	vchlog.Init()
+	logs = append(logs, vchlog.GetPipe())
+	go vchlog.Run()
+	defer vchlog.Close() // close the logger pipe when done
 
 	// Initiliaze logger with default TextFormatter
 	log.SetFormatter(viclog.NewTextFormatter())
--- a/lib/install/management/create.go
+++ b/lib/install/management/create.go
@@ -26,6 +26,7 @@ import (
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/lib/install/data"
+	"github.com/vmware/vic/lib/install/vchlog"
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/retry"
 	"github.com/vmware/vic/pkg/trace"
@@ -37,6 +38,7 @@ const (
 	uploadMaxElapsedTime  = 30 * time.Minute
 	uploadMaxInterval     = 1 * time.Minute
 	uploadInitialInterval = 10 * time.Second
+	timeFormat            = "2006-01-02T15:04:05-0700"
 )
 
 func (d *Dispatcher) CreateVCH(conf *config.VirtualContainerHostConfigSpec, settings *data.InstallerData) error {
@@ -59,6 +61,16 @@ func (d *Dispatcher) CreateVCH(conf *config.VirtualContainerHostConfigSpec, sett
 	if err = d.createAppliance(conf, settings); err != nil {
 		return errors.Errorf("Creating the appliance failed with %s. Exiting...", err)
 	}
+
+	// send the signal to VCH logger to indicate VCH datastore path is ready
+	datastoreReadySignal := vchlog.DatastoreReadySignal{
+		Datastore:   d.session.Datastore,
+		LogFileName: "vic-machine-create",
+		Operation:   trace.NewOperation(d.ctx, "vic-machine create"),
+		VMPathName:  d.vmPathName,
+		Timestamp:   time.Now().UTC().Format(timeFormat),
+	}
+	vchlog.Signal(datastoreReadySignal)
 
 	if err = d.uploadImages(settings.ImageFiles); err != nil {
 		return errors.Errorf("Uploading images failed with %s. Exiting...", err)
@@ -139,7 +151,7 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
 					switch err.(type) {
 					// if not found, do nothing
 					case object.DatastoreNoSuchFileError:
-					// otherwise force delete
+						// otherwise force delete
 					default:
 						log.Debugf("target delete path = %s", isoTargetPath)
 						err := fm.Delete(d.ctx, isoTargetPath)
--- a/lib/install/vchlog/vchlogger.go
+++ b/lib/install/vchlog/vchlogger.go
@@ -1,0 +1,70 @@
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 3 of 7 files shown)


cmd/vic-machine/main.go
```diff
@@ -139,6 +140,12 @@ func main() {
 		logs = append(logs, f)
 	}
 
+	// create VCH logger for streaming creation logs to datastore folder
```
zjs: It seems like the logging framework we're using, logrus, [uses hooks as an extensibility model](https://github.com/sirupsen/logrus#hooks). Could our new logging functionality be simplified by following that pattern?

AngieCris: hook is essentially callback function that gets called every time a log message is written. `Datastore.Upload` provided by govmomi overwrites the original content, so we can't do upload in hook (every `log.infof`, `log.debugf` and the like will be followed with a upload, and the previous log message got overwritten by the next one, etc.)
Also, we know exactly what point of time the datastore folder is ready, so there's no need to do signaling work in callback hook.

AngieCris:
@zjs any read from `io.PipeReader` is blocked until a write is done, even after wrapping it with `bufio.Reader`. 
I tried wrapping the `io.PipeWriter` with `bufio.Writer`, but after inspecting the source code, the writer won't write to the underlying buffer without flushing. Since every `log.infof` and `log.debugof` and the like is a write to the writer, I could do flushing in a callback logrus hook but I think it's just neater to implement our own logging package.

AngieCris:
Update:

now the log file name contains a unique operation ID for every `vic-machine` process.
still, it only works for VCH creation, not other vic-machine commands. VCH creation is partially asynchronous. So this update is in prep for future development of live log streaming for other VCH commands

zjs:
> any read from `io.PipeReader` is blocked until a write is done, even after wrapping it with `bufio.Reader`

Isn't this the behavior we want? That is, don't we expect the upload process reading the stream to block between writes?

AngieCris:
@zjs I just edited the comments - we want the writes to be buffered, not the reads, so even we wrap the `io.PipeReader` with `bufio.Reader`, still the second write would be blocked.
As I mentioned in the comment, I tried wrapping `io.PipeWriter` with `bufio.Writer`, but the write won't be flushed automatically to the underlying writer stream (so that the reader can receive it) unless we explicitly do `writer.Flush()`

lib/install/vchlog/vchlogger.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// Run waits until the signal arrives and uploads the streaming pipe to datastore
+func Run() {
+	sig := <-signalChan
+
+	// suffix the log file name with caller operation ID as vic-machine process identifier
+	callerOp := trace.NewOperation(sig.Context, "vic-machine")
+	opID := callerOp.ID()
+	logFileNameWithOpID := sig.LogFileName + "_" + opID
```
zjs: Can we include a timestamp here for easy sorting of the log files? (Maybe a `startTime` could be stored as a part of the operation created by `trace.NewOperation`?)

lib/install/vchlog/vchlogger.go
```diff
@@ -0,0 +1,69 @@
...
+	pipe = NewBufferedPipe()
+	signalChan = make(chan VCHCreatedSignal)
+}
+
+// Run waits until the signal arrives and uploads the streaming pipe to datastore
+func Run() {
+	sig := <-signalChan
+
+	// suffix the log file name with caller operation ID as vic-machine process identifier
+	callerOp := trace.NewOperation(sig.Context, "vic-machine")
```
zjs: This seems like the wrong place to be calling `trace.NewOperation`. Shouldn't this code be passed a `context` where the operation has already been created? (That way the operation id in the log messages matches the operation id in the filename.)

lib/install/vchlog/vchlogger.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"context"
+	"path"
+
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/vic/pkg/trace"
+)
+
+// VCHCreatedSignal serves as a signal struct indicating datastore folder path is available
```
zjs: Can we come up with a more accurate name for this? (We're sending the datastore is created, not the VCH; we're only part of the way through the VCH creation process.)

AngieCris: thanks, good catch

lib/install/vchlog/vchlogger.go
AngieCris: yes you're right. This is def not a good place to do `trace.NewOperation`...
fixing it

lib/install/vchlog/vchlogger.go
AngieCris: good idea. `Operation` package currently does not have a timestamp field, but I can pass that as a signal to the logger along with the operation. Do we want the timestamp to be unix nano in msec, or year-month-day?

AngieCris:
Clarifications on the design of a self-implemented log package (`bufferedPipe`):

We have the following goals that a logger that streams VCH creation logs to VCH datastore folder needs to achieve:
1. buffer an unfixed (possibly unbounded) number of writes 
2. block the read when there's no data to read (waiting for new data to come)
The reasons being that:
1. VCH datastore folder path is available quite late in the process, after the params parsing, network configuring, pool checking, etc. Before it becomes available, there is an uncertain number of log messages being generated and we should buffer them instead of dropping them. After the datastore folder path is ready, we begin reading. 
2. The logger does not write continuously. Since the VCH creation process is across packages/layers, there's necessary context switch that results in lag time between some writes. Since `Datastore.Upload` reads in a tight loop, data will be exhausted during the context switch before the new write comes after context switch. Hence the read must be blocked when data is exhausted. 
If the read is not blocked, during context switch time in VCH creation process, the log messages are exhausted by read, and further reads give EOF which early terminates the streaming which results in incomplete log files.

According to 1. above, we must have a `bytes.Buffer` as underlying structure since it grows and buffers data.
According to 2., the read is not blocked unless the logger pipe has been explicitly closed. And the usage of `sync.Cond` in the `bufferedPipe` package enables this with `Wait()` and `Signals()`.

Why standard libraries by golang does not work:
1. Use `io.Pipe` and wrap `io.PipeReader` with `bufio.Reader`
`bufio.Reader` does not buffer writes. 
2. Use `io.Pipe` and wrap `io.PipeWriter` with `bufio.Writer`
This does buffer writes, but it does not flush to the underlying pipe for the reader to grab unless we explicitly do `writer.Flush()`. This means at the moment datastore path becomes available, we flush the buffered writes. For all log messages after that point, we flush for every write (because we want it to be "real-time"). 
Since every `log.infof`, `log.debugf` and the like is a write, the only way we can do this is via a callback log hook. But this hook is applied automatically to all writes alike, we can't twerk it to act differently for log messages in different situations.
3. Use an intermediate `bytes.Buffer` to be consumed and produced at the same time
The main issue is that read must be blocked. During context switch time, if read is not blocked, incomplete log files would be generated due to early EOF.

lib/install/management/create.go
```diff
@@ -60,6 +62,16 @@ func (d *Dispatcher) CreateVCH(conf *config.VirtualContainerHostConfigSpec, sett
 		return errors.Errorf("Creating the appliance failed with %s. Exiting...", err)
 	}
 
+	// send the signal to VCH logger to indicate VCH datastore path is ready
+	datastoreReadySignal := vchlog.DatastoreReadySignal{
+		Datastore: d.session.Datastore,
+		LogFileName: "vic-machine-create",
+		Operation: trace.NewOperation(d.ctx, "vic-machine create"),
```
emlin: didn't see much value for the operation ID here, because you're writing logs to different files for different operation in different time.

AngieCris: there could be more than one vic machine processes at the same time, to prevent overwriting, we also included the operation ID together with the timestamp

PR #6423, opened 2017-09-25
Increase maxInFlight and TLSHandshakeTimeout

caglar10ur:
Helps #5621


cgtexmex:
LGTM

caglar10ur:
I believe @hickeng asked me to keep 16 for < 6.5 and increase it to 32 only for 6.5. Is that still the case George? If so I'll refresh the PR before merging

hickeng:
@caglar10ur Don't remember why - might have been related to 1.2.1. Feel free to bump it up, but we should confirm with Lakshmi in a 6.0u3 env with multiple VCHs active concurrently, not just present.

caglar10ur:
OK, merging this as it is. We can revisit it later if we found out this is aggressive for multiple vch case.

PR #6457, opened 2017-09-26
[specific ci=1-23-Docker-Inspect] Conditionally sync container state during refresh

### LINKED ISSUE #6372: Status of newly created container is "exited"
This has proved to be an issue in my experimentations with SwarmKit - SwarmKit creates a container and then looks at the status to decide how to proceed. When it sees "exited" and not "created", it goes down a wrong code path and loops trying to recreate the container.

Steps to reproduce:

```
docker create --name=foo busybox
docker inspect foo
```

You'll see the State.Status as "exited", when it should be "created".

I debugged the reason this is happening.
- The container is created and its state is set to "created".
- The inspect call causes Refresh() to be called in lib/portlayer/exec/container.go#L331 and this looks at the power state of the VM. It slams the state from "created" to "stopped" because the VM is powered off.
- The personality then turns the "stopped" status into an "exited" status here: lib/apiservers/engine/backends/container_proxy.go#L1491.

The simplest fix would be to add a clause to Refresh() so that if the state is "created", it doesn't make the modification. Arguably the Port Layer should only care about stopped or started, but the "created" state needs to be preserved and passed back to the persona for the inspect result.

cgtexmex:
The refresh method was syncing the state of the container
soley based on the power state of the VM.  Now the refresh
will include process status and current state as criteria to
syncing.

Fixes #6372

### DIFF
--- a/lib/portlayer/exec/container.go
+++ b/lib/portlayer/exec/container.go
@@ -337,12 +337,25 @@ func (c *Container) Refresh(ctx context.Context) error {
 		return err
 	}
 
-	// sync power state (see issue 4872)
+	var started bool
+	// determine if the containerVM has started
+	if session, exists := c.ExecConfig.Sessions[c.ExecConfig.ID]; exists {
+		if session.Started == "true" {
+			started = true
+		}
+	}
+
+	// conditionally sync state (see issue 4872, 6372)
 	switch c.containerBase.Runtime.PowerState {
 	case types.VirtualMachinePowerStatePoweredOn:
-		c.state = StateRunning
+		// only set to running if the container process has started
+		if started {
+			c.state = StateRunning
+		}
 	case types.VirtualMachinePowerStatePoweredOff:
-		c.state = StateStopped
+		if c.state != StateCreated {
+			c.state = StateStopped
+		}
 	}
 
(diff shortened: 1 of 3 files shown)


cgtexmex:
ping @hickeng @sflxn

lib/portlayer/exec/container.go
```diff
@@ -323,12 +323,25 @@ func (c *Container) Refresh(ctx context.Context) error {
...
+			started = true
+		}
+	}
+
+	// conditionally sync state (see issue 4872, 6372)
 	switch c.containerBase.Runtime.PowerState {
 	case types.VirtualMachinePowerStatePoweredOn:
-		c.state = StateRunning
+		// only set to running if the container process has started
+		if started {
```
hickeng: I think this will mean that we report "Running" for a containerVM that's been out-of-band restarted but before it's actually up.
We _could_ see if volatile guestinfo keys are dropped during a soft power cycle (IIRC they are not, but worth checking) and if so use one as an indicator of freshness.

hickeng: I've approve it as-is because it's needed for #6557 and it's a significant improvement over current state.

Issue #6341, opened 2017-09-14
Concurrent container auto remove doesn't remove all containers


cgtexmex:
I created 16 containers concurrently with --rm specified and named 2 - 17 via:

`docker run -d --rm --name {#} ubuntu /bin/sh -c 'a=0; while [ $a -lt 75 ]; do echo "line $a"; a=`expr $a + 1`; sleep 2; done;'`

This resulted in 16 containers:
```
docker ps
CONTAINER ID        IMAGE               COMMAND                  CREATED              STATUS              PORTS               NAMES
b07f1ab1d37a        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       5
bb7c09c14676        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       4
111fb3a28833        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       12
2c733c90a322        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       13
5062fadd9df3        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       15
15ba4f59d9ba        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       11
fee7c5d822b8        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       2
c03a83fb6a85        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       7
f81b635bceca        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       3
4c672443ec4d        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       6
6a1cb2bb5636        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       16
b0e0de57d268        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       10
6d929e8800ef        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       17
bf9edaf8f983        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       14
29ff1eb418ea        ubuntu              "/bin/sh -c a=0; w..."   About a minute ago   Up About a minute                       9
[...]

PR #6470, opened 2017-09-27
Minor log message changes (to help #6341)

caglar10ur:


cgtexmex:
LGTM

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -223,7 +223,7 @@ func (handler *ContainersHandlersImpl) CommitHandler(params containers.CommitPar
 	}
 
 	if err := h.Commit(context.Background(), handler.handlerCtx.Session, params.WaitTime); err != nil {
-		log.Errorf("CommitHandler error on handle(%s) for %s: %#v", h.String(), h.ExecConfig.ID, err)
+		log.Errorf("CommitHandler error on handle(%s) for %s: %s", h.String(), h.ExecConfig.ID, err)
```
vburenin: No need to call String for stringers. %s should do the job.

pkg/vsphere/vm/vm.go
```diff
@@ -174,7 +174,6 @@ func (vm *VirtualMachine) WaitForExtraConfig(ctx context.Context, waitFunc func(
 	// https://www.vmware.com/support/developer/vc-sdk/visdk2xpubs/ReferenceGuide/vim.vm.ConfigInfo.html
 	err := property.Wait(ctx, p, vm.Reference(), []string{"config.extraConfig", object.PropRuntimePowerState}, waitFunc)
 	if err != nil {
```
vburenin: I guess you would want just to return err... no need to check it for nil

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
caglar10ur: It is already %s?

caglar10ur: Ahh the argument is still calling String(), will fix in a sec.

caglar10ur:
@vburenin done

PR #6475, opened 2017-09-28
Define end-to-end test for version endpoint [specific ci=Group23-Vic-Machine-Service]

zjs:
Introduce a single-test robot suite for the version endpoint as well as a utility resource for use by this suite and others within the group.

### DIFF
--- a/tests/resources/Group23-VIC-Machine-Service-Util.robot
+++ b/tests/resources/Group23-VIC-Machine-Service-Util.robot
@@ -1,0 +1,59 @@
+# Copyright 2017 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#	http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License
+
+*** Settings ***
+Documentation	This resource contains keywords which are helpful for using curl to test the vic-machine API.
+
+
+*** Variables ***
+${HTTP_PORT}	1337
+${HTTPS_PORT}	31337
+
+
+*** Keywords ***
+
+Start VIC Machine Server
+    Start Process    ./bin/vic-machine-server --port ${HTTP_PORT} --scheme http    shell=True    cwd=/go/src/github.com/vmware/vic
+
+
+Get Path
+    [Arguments]    ${path}
+    ${RC}  ${OUTPUT}=    Run And Return Rc And Output    curl -s -w "\n\%{http_code}\n" -X GET "http://127.0.0.1:${HTTP_PORT}/container/${PATH}"
+    ${OUTPUT}    ${STATUS}=    Split String From Right    ${OUTPUT}    \n    1
+    Set Test Variable    ${RC}
+    Set Test Variable    ${OUTPUT}
+    Set Test Variable    ${STATUS}
+
+
+Get Path Under Target
+    [Arguments]    ${path}
+    ${auth}=    Evaluate    base64.b64encode("%{TEST_USERNAME}:%{TEST_PASSWORD}")    modules=base64
+    ${RC}  ${OUTPUT}=    Run And Return Rc And Output    curl -s -w "\n\%{http_code}\n" -X GET "http://127.0.0.1:${HTTP_PORT}/container/target/%{TEST_URL}/${PATH}?thumbprint=%{TEST_THUMBPRINT}" -H "Accept: application/json" -H "Authorization: Basic ${auth}"
+    ${OUTPUT}    ${STATUS}=    Split String From Right    ${OUTPUT}    \n    1
+    Set Test Variable    ${RC}
+    Set Test Variable    ${OUTPUT}
+    Set Test Variable    ${STATUS}
+
+
+Verify Return Code
+    Should Be Equal As Integers    ${RC}    0
+
+
+Verify Status
+    [Arguments]    ${expected}
+    Should Be Equal As Integers    ${expected}    ${STATUS}
+
+
+Verify Status Ok
+    Verify Status    200
--- a/tests/test-cases/Group23-VIC-Machine-Service/23-01-Version.md
+++ b/tests/test-cases/Group23-VIC-Machine-Service/23-01-Version.md
@@ -1,0 +1,23 @@
+Test 23-01 - Version
+=======
+
+# Purpose:
+To verify vic-machine-server returns a valid version number
+
+# References:
+1. [The design document](../../../doc/design/vic-machine/service.md)
+
+# Environment:
+This test has no environmental requirements
+
+# Test Steps:
+1. Start the vic-machine-server
+2. Wait 1 second
+3. Use curl to issue a GET request for the version
+
+# Expected Outcome:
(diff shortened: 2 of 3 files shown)


tests/test-cases/Group23-VIC-Machine-Service/23-00-Util.robot
```diff
@@ -0,0 +1,63 @@
...
+    Set Test Variable    ${OUTPUT}
+    Set Test Variable    ${STATUS}
+
+
+Get Path Under Target
+    [Arguments]    ${path}
+    ${auth}=    Evaluate    base64.b64encode("%{TEST_USERNAME}:%{TEST_PASSWORD}")    modules=base64
+    ${RC}  ${OUTPUT}=    Run And Return Rc And Output    curl -s -w "\n\%{http_code}\n" -X GET "http://127.0.0.1:${HTTP_PORT}/container/target/%{TEST_URL}/${PATH}?thumbprint=%{TEST_THUMBPRINT}" -H "Accept: application/json" -H "Authorization: Basic ${auth}"
+    ${STATUS}=    Get Line    ${OUTPUT}    -1
+    Set Test Variable    ${RC}
```
jzt: I haven't seen this pattern before, but I think I like it. It makes the tests very clean.

mdharamadas1: I wouldn't suggest having these keywords in a separate file as I think it will be run as test suite as its under test suite dir but @mhagen-vmware can confirm that. Also, I don't like the corresponding `.md` empty file.

zjs: > I wouldn't suggest having these keywords in a separate file as I think it will be run as test suite as its under test suite dir but @mhagen-vmware can confirm that.

If it gets run as a suite, it seems like it will get run as a suite with zero tests. That isn't a problem locally. Is it a problem for our CI infrastructure?

(Replying to @mhagen-vmware here, to keep related conversation in a single thread.)
>  this should be a resource file within our resources folder and needs to be added into resources/Util.robot please

These keywords aren't intended to be used by other test groups. Defining them in the resources folder and adding that file to `resources/Util.robot` essentially adds them to a global namespace.

Do we have a better pattern for sharing keywords between suites within a single test group?

Is there a functional problem with this pattern, or is the complaint only that it's stylistically different than what we've done in the past?

> I don't like the corresponding `.md` file.

I don't like it either, but was attempting to following the formatting conventions. I'd rather use that file as documentation for these keywords, but I think I'd need to format it differently. Are we okay with that?

mdharamadas1: > Is it a problem for our CI infrastructure?

No, but that pattern (having util files under tests) doesn't seem right to me.

> Are we okay with that?

I would suggest removing it for now as we don't have any documentation for existing keywords.

zjs: > No, but that pattern (having util files under tests) doesn't seem right to me.

I moved the resource file so that it's under resources, and named it in a way that makes it clear it's intended for use within this test group.

> I would suggest removing it for now as we don't have any documentation for existing keywords.

I removed the corresponding md file.

tests/resources/Group23-VIC-Machine-Service-Util.robot
```diff
@@ -0,0 +1,59 @@
+# Copyright 2017 VMware, Inc. All Rights Reserved.
```
mdharamadas1: Can we rename this file to something more general, may be like `VIC-Machine-Service-Util.robot`? We may use it in other test suites as well.

tests/resources/Group23-VIC-Machine-Service-Util.robot
```diff
@@ -0,0 +1,59 @@
...
+    [Arguments]    ${path}
+    ${auth}=    Evaluate    base64.b64encode("%{TEST_USERNAME}:%{TEST_PASSWORD}")    modules=base64
+    ${RC}  ${OUTPUT}=    Run And Return Rc And Output    curl -s -w "\n\%{http_code}\n" -X GET "http://127.0.0.1:${HTTP_PORT}/container/target/%{TEST_URL}/${PATH}?thumbprint=%{TEST_THUMBPRINT}" -H "Accept: application/json" -H "Authorization: Basic ${auth}"
+    ${OUTPUT}    ${STATUS}=    Split String From Right    ${OUTPUT}    \n    1
+    Set Test Variable    ${RC}
+    Set Test Variable    ${OUTPUT}
+    Set Test Variable    ${STATUS}
+
+
+Verify Return Code
```
mdharamadas1: I would put all these verify keywords in a separate util file, may be, called `Verify-Util.robot`.

tests/resources/Group23-VIC-Machine-Service-Util.robot
```diff
@@ -0,0 +1,59 @@
...
+${HTTPS_PORT}	31337
+
+
+*** Keywords ***
+
+Start VIC Machine Server
+    Start Process    ./bin/vic-machine-server --port ${HTTP_PORT} --scheme http    shell=True    cwd=/go/src/github.com/vmware/vic
+
+
+Get Path
```
mdharamadas1: As this is in global space now, can we rename this keyword to little more specific?

tests/resources/Group23-VIC-Machine-Service-Util.robot
```diff
@@ -0,0 +1,59 @@
...
+    [Arguments]    ${path}
+    ${RC}  ${OUTPUT}=    Run And Return Rc And Output    curl -s -w "\n\%{http_code}\n" -X GET "http://127.0.0.1:${HTTP_PORT}/container/${PATH}"
+    ${OUTPUT}    ${STATUS}=    Split String From Right    ${OUTPUT}    \n    1
+    Set Test Variable    ${RC}
+    Set Test Variable    ${OUTPUT}
+    Set Test Variable    ${STATUS}
+
+
+Get Path Under Target
```
mdharamadas1: As this is in global space now, can we rename this keyword to little more specific?

tests/resources/Group23-VIC-Machine-Service-Util.robot
zjs: I don't want to give the file a more general name _because_ I don't want these keywords used in other test groups.

Our current process allows for only two types of keywords: suite-scoped and global. This file follows a different pattern: group-scoped.

Defining all shared keywords in the global scope carries many of the same drawbacks as defining variables in the global scope: identifying relevant keywords becomes harder, determining where keywords are used becomes harder, safely changing the behavior of keywords becomes harder, concisely and accurately naming keywords becomes harder, etc.

We'll iterate on this design as we learn more about this pattern, but I think this is a reasonable starting point.

tests/resources/Group23-VIC-Machine-Service-Util.robot
zjs: Placing these keywords in the global scope would, as you suggest, require providing more specific names for the keywords. However, within the scope of this test group I think the keyword names are clear.

tests/resources/Group23-VIC-Machine-Service-Util.robot
zjs: As above.

tests/resources/Group23-VIC-Machine-Service-Util.robot
zjs: Given the close dependency between the way the `Get` (and eventually `Post`/`Put`/`Delete`) methods store results and the way the `Verify` methods make assertions, I think it's important to keep the keywords in the same file; using the `Verify` keywords without the `Get` keywords wouldn't make sense.

zjs:
> I'm fine with iterating this further later, would suggest adding an issue for that.

Good idea! Filed #6508.

PR #6487, opened 2017-10-02
Alphabetize imports in bufferedPipe.go [specific ci=Group23-Vic-Machine-Service]

zjs:
Organize the the imports in bufferedPipe.go in alphabetical order.

Without this change, `make goimports` (run as a part of CI) fails.

Example failure: https://ci.vcna.io/vmware/vic/13891

Note: This could be squashed into "live streaming VCH creation log to VCH datastore folder" before merging to master.


anchal-agrawal:
While we're at it, can we rename this file to `buffered_pipe.go`? While I couldn't find the spec for filenames with compound words, we have some files that use snake case.

https://github.com/vmware/vic/blob/master/lib/apiservers/engine/backends/container_proxy.go
https://github.com/vmware/vic/blob/master/lib/apiservers/engine/backends/system_portlayer.go

PR #6494, opened 2017-10-03
[specific ci=6-07-Create-Network] Fix 6-07-Create-Network tests for vCenter

### LINKED ISSUE #6441: 6-07-Create-Network fails against vCenter
- Reset VCH doesn't cause unintentionally exposed ports from container network
   - Public network - default
   - Management network - valid
   - Container Firewalls
   - Bridge network - valid
   - Container network - space in network name valid

anchal-agrawal:
This commit modifies several tests in the `6-07-Create-Network` suite
to run on vCenter. It adds new Robot keywords to create and remove
distributed portgroups on a VC and uses these keywords if the test
is running on VC.

Fixes #6441

### DIFF
--- a/tests/resources/VCH-Util.robot
+++ b/tests/resources/VCH-Util.robot
@@ -297,6 +297,14 @@ Cleanup VCH Bridge Network
     ${out}=  Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Run  govc host.portgroup.info
     Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Should Not Contain  ${out}  ${name}-bridge
 
+Add VC Distributed Portgroup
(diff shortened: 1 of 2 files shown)


anchal-agrawal:
@mdharamadas1 Good points.

> we could improve this further to follow DRY

Based on the docs I found, there's no efficient mechanism to run multiple keywords inside a `Run Keyword If` - we could either create a nested keyword to wrap all individual statements, or use `Run Keywords`. Both don't fit in well here since the statements differ based on the test.

> Also, we need to come up with some global pattern to run these tests on either ESX or VC (i don't like those if stmts everywhere in tests).

An alternative for now would be to create separate files/suites for ESX and VC, but that would cause churn every time we add or change tests.

PR #6499, opened 2017-10-03
Add/clarify some logging in portlayer

caglar10ur:
Also added a new event "ContainerFailed" where we send from portlayer and listen in persona when start fails.

To help #6341

### DIFF
--- a/lib/apiservers/engine/backends/eventmonitor.go
+++ b/lib/apiservers/engine/backends/eventmonitor.go
@@ -218,34 +218,32 @@ func (p DockerEventPublisher) PublishEvent(event plevents.BaseEvent) {
 
 	vc := cache.ContainerCache().GetContainer(event.Ref)
 	if vc == nil {
-		log.Warnf("Portlayer event for container %s but not found in cache", event.Ref)
+		log.Warnf("Received portlayer event %s for container %s but not found in cache", event.Event, event.Ref)
 		return
 	}
 
 	var attrs map[string]string
-	// TODO: move to a container.OnEvent() so that container drives the necessary changes
-	// based on event activity
+	// TODO: move to a container.OnEvent() so that container drives the necessary changes based on event activity
 	switch event.Event {
 	case plevents.ContainerStopped,
-		plevents.ContainerPoweredOff:
-		// since we are going to make a call to the portLayer lets execute this in a
-		// go routine
+		plevents.ContainerPoweredOff,
+		plevents.ContainerFailed:
+		// since we are going to make a call to the portLayer lets execute this in a go routine
 		go func() {
 			attrs := make(map[string]string)
 			// get the containerEngine
 			code, err := NewContainerBackend().containerProxy.exitCode(vc)
 			if err != nil {
 				// log the error, but continue
-				log.Errorf("unable to get exitCode for die event: %s", err)
+				log.Warnf("Unable to get exit code for %s: %s", vc.ContainerID, err)
 			}
-			// if the docker client is unable to convert the code to an int then
-			// then the client will return 125
-			attrs["exitCode"] = code
 			log.Infof("Sending die event for container %s - code: %s", vc.ContainerID, code)
+			// if the docker client is unable to convert the code to an int the client will return 125
+			attrs["exitCode"] = code
 			actor := CreateContainerEventActorWithAttributes(vc, attrs)
 			EventService().Log(containerDieEvent, eventtypes.ContainerEventType, actor)
 			if err := UnmapPorts(vc.ContainerID, vc.HostConfig); err != nil {
-				log.Warn(err)
+				log.Warnf("Failed to unmap ports for container %s: %s", vc.ContainerID, err)
 			}
 
 			// auto-remove if required
@@ -257,17 +255,17 @@ func (p DockerEventPublisher) PublishEvent(event plevents.BaseEvent) {
 
 				err := NewContainerBackend().ContainerRm(vc.Name, config)
 				if err != nil {
-					log.Warnf("failed to auto remove container: %s", err.Error())
+					log.Errorf("Failed to remove container %s: %s", vc.ContainerID, err)
 				}
 			}
-
 		}()
 	case plevents.ContainerRemoved:
 		attrs = make(map[string]string)
-		//pop the destroy event...
+		// pop the destroy event...
 		actor := CreateContainerEventActorWithAttributes(vc, attrs)
 		EventService().Log(containerDestroyEvent, eventtypes.ContainerEventType, actor)
 		if err := UnmapPorts(vc.ContainerID, vc.HostConfig); err != nil {
+			log.Errorf("Failed to unmap ports for container %s: %s", vc.ContainerID, err)
 			log.Warn(err)
 		}
 		// remove from the container cache...
--- a/lib/apiservers/engine/backends/eventmonitor.go
+++ b/lib/apiservers/engine/backends/eventmonitor.go
@@ -232,18 +232,15 @@ func (p DockerEventPublisher) PublishEvent(event plevents.BaseEvent) {
 		go func() {
 			attrs := make(map[string]string)
 			// get the containerEngine
-			code, err := NewContainerBackend().containerProxy.exitCode(vc)
-			if err != nil {
-				// log the error, but continue
-				log.Warnf("Unable to get exit code for %s: %s", vc.ContainerID, err)
-			}
+			code, _ := NewContainerBackend().containerProxy.exitCode(vc)
+
 			log.Infof("Sending die event for container %s - code: %s", vc.ContainerID, code)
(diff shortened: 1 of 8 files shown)


lib/portlayer/event/manager.go
```diff
@@ -72,7 +72,7 @@ func NewEventManager(collectors ...collector.Collector) *Manager {
 			log.Debugf("Found %d subscribers to id: %d - %s: %s", len(subs), e.EventID(), e.Topic(), e.Message())
 
 			for sub, s := range subs {
-				log.Debugf("Event manager calling back to %s", sub)
+				log.Debugf("Event manager calling back to %s for %d - %s: %s", sub, e.EventID(), e.Topic(), e.Message())
```
cgtexmex: Do we need the message & topic on this line since we have it on the line 72?  Just add eventID and then we are able to associate?

caglar10ur: Yes, otherwise we see lots of lines saying "Event manager calling back to exec" around the same area of the log file when we do things concurrently and no way to understand which one we missed etc.

cgtexmex: I was getting ready to say the event id allows us to tie it back...which I thought was the case, but since we are generating events w/o an id I understand why we'd need the dupe info...

cgtexmex:
LGTM

caglar10ur:
I changed more log lines and started to use c instead c.ExecConfig.ID as String method already does that. I also added a new event "ContainerFailed" where we send from portlayer and listen in persona when start fails.

lib/apiservers/engine/backends/eventmonitor.go
```diff
@@ -257,17 +255,17 @@ func (p DockerEventPublisher) PublishEvent(event plevents.BaseEvent) {
...
 		}()
 	case plevents.ContainerRemoved:
 		attrs = make(map[string]string)
-		//pop the destroy event...
+		// pop the destroy event...
 		actor := CreateContainerEventActorWithAttributes(vc, attrs)
 		EventService().Log(containerDestroyEvent, eventtypes.ContainerEventType, actor)
 		if err := UnmapPorts(vc.ContainerID, vc.HostConfig); err != nil {
+			log.Errorf("Failed to unmap ports for container %s: %s", vc.ContainerID, err)
 			log.Warn(err)
```
patiner: Do we need to use both `Errorf` and `Warn` here?

chengwang86: Remove this since `err` is logged twice.

lib/portlayer/exec/container.go
```diff
@@ -220,6 +230,10 @@ func (c *ContainerInfo) State() State {
 	return c.state
 }
 
+func (c *Container) String() string {
```
chengwang86: Since we already have `containerBase.String()` and `ContainerInfo.String()` (in container.go), do we really need this method?

lib/portlayer/network/context.go
```diff
@@ -858,6 +858,85 @@ func (c *Context) container(h *exec.Handle) (*Container, error) {
 	return nil, ResourceNotFoundError{error: fmt.Errorf("container %s not found", id.String())}
 }
 
+func (c *Context) UnbindContainerByID(id string) ([]*Endpoint, error) {
```
chengwang86: This method has some duplications with `UnbindContainer`. Is it easy to consolidate these two methods?

lib/portlayer/exec/commit.go
```diff
@@ -196,6 +196,7 @@ func Commit(ctx context.Context, sess *session.Session, h *Handle, waitTime *int
 
 		// start the container
 		if err := c.start(ctx); err != nil {
+			publishContainerEvent(h.ExecConfig.ID, time.Now().UTC(), events.ContainerFailed)
```
hmahmood: What is this for?

caglar10ur:
Dropping the UnbindContainer related bits as they are now a different PR (https://github.com/vmware/vic/pull/6525). Sorry for the confusion :/

lib/apiservers/engine/backends/eventmonitor.go
caglar10ur: Done

lib/portlayer/exec/container.go
caglar10ur: Not really, but one day someone could change one of those to output something else and I didn't want to get affected by it.

lib/portlayer/exec/commit.go
caglar10ur: To workaround a VC/ESXi problem. We observed that PowerOn_Task gets stuck on VC time to time even though the VM was starting fine on the host ESXi. Eventually the task was getting timed out (After 20 min.) and that was setting the container state back to Stopped. During that time VC is not generating any other event so the persona listener was getting nothing. This new event is for signaling the eventmonitor so that it can autoremove the container after this failure if it is marked as auto-removable.

lib/apiservers/engine/backends/eventmonitor.go
```diff
@@ -218,34 +218,32 @@ func (p DockerEventPublisher) PublishEvent(event plevents.BaseEvent) {
...
+		plevents.ContainerFailed:
+		// since we are going to make a call to the portLayer lets execute this in a go routine
 		go func() {
 			attrs := make(map[string]string)
 			// get the containerEngine
 			code, err := NewContainerBackend().containerProxy.exitCode(vc)
 			if err != nil {
 				// log the error, but continue
-				log.Errorf("unable to get exitCode for die event: %s", err)
+				log.Warnf("Unable to get exit code for %s: %s", vc.ContainerID, err)
```
hickeng: This may be expected in the container failed case - if so can we avoid logging it as it will confuse matters in the logs.

lib/apiservers/engine/backends/eventmonitor.go
```diff
@@ -218,34 +218,32 @@ func (p DockerEventPublisher) PublishEvent(event plevents.BaseEvent) {
...
-			// then the client will return 125
-			attrs["exitCode"] = code
 			log.Infof("Sending die event for container %s - code: %s", vc.ContainerID, code)
+			// if the docker client is unable to convert the code to an int the client will return 125
+			attrs["exitCode"] = code
 			actor := CreateContainerEventActorWithAttributes(vc, attrs)
 			EventService().Log(containerDieEvent, eventtypes.ContainerEventType, actor)
 			if err := UnmapPorts(vc.ContainerID, vc.HostConfig); err != nil {
-				log.Warn(err)
+				log.Warnf("Failed to unmap ports for container %s: %s", vc.ContainerID, err)
```
hickeng: Think this is an error - it potentially leaves a port mapping in place that directs to a non-existent cVM IP.
Worse is if that IP is then reused by a container that's not supposed to have a mapping...

We may want to not release the IP lease if we cannot undo the port forwarding - while this would leak IPs it's more of a fail safe approach.
If we do do that then we should have a path that retries the unmap/release paths for dangling IPs - presumably on `network inspect` or ip pool exhaustion and `network.Bind`

lib/portlayer/exec/commit.go
hickeng: Might be worth at least a short comment about why the event is being generated - not necessarily the VC part, but noting that this allows async logic hung off the event stream to process the failure.

lib/portlayer/exec/container.go
```diff
@@ -591,40 +602,57 @@ func (c *Container) Remove(ctx context.Context, sess *session.Session) error {
...
 		default:
-			log.Debugf("Fault while attempting to destroy vm: %#v", f.Fault())
+			log.Debugf("Unhandled fault while attempting to destroy vm %s: %#v", c, f.Fault())
+
 			c.updateState(existingState)
 			return err
 		}
 	}
 
+	if concurrent && c.vm.IsVC() {
```
hickeng: move this into vm.DeleteExceptDisks instead of inlined here

caglar10ur: Hmm why? The above 15 lines or so already doing similar thing (aka handling the err returned by DeleteExcept), moving this there will just hide this but won't change anything

caglar10ur: I slept on it and decided to move the whole corner case handling into DeleteExceptDisks :) Will refresh the PR

caglar10ur: And nope, not going to happen unless we change all the callers. Everyone wraped the DeleteExceptDisk with task package (so that fn. returns the task not the result). Since ConcurrentAccessError could be thrown by either Reconfigure or Destroy we also need to keep that info somewhere so that next retry could do the right thing. So unless I'm missing something (monday/haven't finished the first cup of coffee) I'll keep this PR intact

caglar10ur:
@hickeng pushed a new commit to address your comments.

Issue #6240, opened 2017-09-07
Nightly 6.5 - Container remove error containerRemoveInternalServerError in 5-4-High-Availability


rajanashok:
This issue is seen on the VIC 1.2.0 RC5 candidate. Previously logged ticket for the same issue. #5198 

```KEYWORD ${rc}, ${output} = OperatingSystem . Run And Return Rc And Output docker %{VCH-PARAMS} rm ${containerMountDataTestID}
Documentation:    
Runs the given command in the system and returns the RC and output.
Start / End / Elapsed:    20170907 01:32:20.593 / 20170907 01:32:21.546 / 00:00:00.953
01:32:20.602    INFO    Running command 'docker -H 10.192.211.83:2376 --tls rm 82faff32a1dccb0bc0f97ffa598a380e6b6bdfb87e1f31154f5a9f595254eb04 2>&1'.    
01:32:21.545    INFO    ${rc} = 1    
01:32:21.546    INFO    ${output} = Error response from daemon: Server error from portlayer: [DELETE /containers/{id}][500] containerRemoveInternalServerError    
00:00:00.001KEYWORD BuiltIn . Should Be Equal As Integers ${rc}, 0
Documentation:    
Fails if objects are unequal after converting them to integers.
Start / End / Elapsed:    20170907 01:32:21.547 / 20170907 01:32:21.548 / 00:00:00.001
01:32:21.548    INFO    Argument types are:
<type 'int'>
<type 'unicode'>    
01:32:21.548    FAIL    1 != 0```

**Comment by rajanashok** [comment]
[5-4-High-Availability.tar.gz](https://github.com/vmware/vic/files/1285328/5-4-High-Availability.tar.gz)

**Comment by mdubya66** [comment]
I also see

```
01:41:30.270 INFO Running command 'docker -H 10.192.211.83:2376 --tls rmi busybox 2>&1'.
01:41:31.864 INFO ${rc} = 1
01:41:31.865 INFO ${output} = Error response from daemon: [DELETE /storage/{store_name}/info/{id}][500] DeleteImage default  &{Code:500 Message:Can't find parent http:///storage/images/4226175f-0e27-cc26-e302-90e7a0a1d258/4560cca0...
```

anchal-agrawal:
@mdubya66 The `rmi busybox` issue is https://github.com/vmware/vic/issues/4858 and is currently in progress (ping @cgtexmex).

anchal-agrawal:
[...]

PR #6511, opened 2017-10-04
Report a better message for container remove if host ESX disconnected

gigawhitlocks:
…host that has gone away

ok my commit message is too long but that's the gist of it.

Before the issue could be seen by running a container, stopping it, powering off the host it was deployed to, and then trying to do docker rm, and we would see an error like this on the command line:
```
Server error from portlayer: [DELETE /containers/{id}][500] containerRemoveInternalServerError 
```

Now, instead, we get this output after trying to `rm` a stopped container from a powered off host:

```
~ 43s
❯  docker -H sc-rdops-vm05-dhcp-151-213.eng.vmware.com:2376 --tlsverify --tlscacert="ian-vch/ca.pem" --tlscert="ian-vch/cert.pem" --tlskey="ian-vch/key.pem" rm brave_golick
Error response from daemon: Server error from portlayer: Couldn't remove container because the host the container is placed on is not connected
```

No test for this (yet) because we need VC in order to reproduce this, but I can either create an issue to create a test when CI v2 is ready or I can write a manual test

### DIFF
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -260,7 +260,7 @@ func (handler *ContainersHandlersImpl) RemoveContainerHandler(params containers.
 				switch f.Fault().(type) {
 				case *types.HostNotConnected:
 					p := &models.Error{Message: "Couldn't remove container. The ESX host is temporarily disconnected. Please try again later."}
-					return containers.NewContainerRemoveDefault(http.StatusInternalServerError).WithPayload(p)
+					return containers.NewContainerRemoveInternalServerError().WithPayload(p)
 				}
 			}
 			return containers.NewContainerRemoveInternalServerError()
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -32,7 +32,6 @@ import (
 	"github.com/go-openapi/runtime/middleware"
 
 	"github.com/vmware/govmomi/vim25/types"
-
 	"github.com/vmware/vic/lib/apiservers/portlayer/models"
 	"github.com/vmware/vic/lib/apiservers/portlayer/restapi/operations"
 	"github.com/vmware/vic/lib/apiservers/portlayer/restapi/operations/containers"
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -31,6 +31,8 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/go-openapi/runtime/middleware"
 
+	"github.com/vmware/govmomi/vim25/types"
+
 	"github.com/vmware/vic/lib/apiservers/portlayer/models"
 	"github.com/vmware/vic/lib/apiservers/portlayer/restapi/operations"
 	"github.com/vmware/vic/lib/apiservers/portlayer/restapi/operations/containers"
@@ -43,8 +45,6 @@ import (
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/uid"
 	"github.com/vmware/vic/pkg/version"
-
-	"github.com/vmware/govmomi/vim25/types"
 )
 
 const (
@@ -260,7 +260,8 @@ func (handler *ContainersHandlersImpl) RemoveContainerHandler(params containers.
 			if f, ok := err.(types.HasFault); ok {
 				switch f.Fault().(type) {
 				case *types.HostNotConnected:
-					return containers.NewContainerRemoveDefault(500).WithPayload(&models.Error{Message: "Couldn't remove container. The ESX host is temporarily disconnected. Please try again later."})
+					p := &models.Error{Message: "Couldn't remove container. The ESX host is temporarily disconnected. Please try again later."}
+					return containers.NewContainerRemoveDefault(http.StatusInternalServerError).WithPayload(p)
 				}
 			}
 			return containers.NewContainerRemoveInternalServerError()
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -257,11 +257,10 @@ func (handler *ContainersHandlersImpl) RemoveContainerHandler(params containers.
 		case exec.RemovePowerError:
 			return containers.NewContainerRemoveConflict().WithPayload(&models.Error{Message: err.Error()})
 		default:
-			f, ok := err.(types.HasFault)
-			if ok {
+			if f, ok := err.(types.HasFault); ok {
 				switch f.Fault().(type) {
 				case *types.HostNotConnected:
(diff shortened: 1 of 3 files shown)


gigawhitlocks:
Also this resolves #6240

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -255,6 +257,13 @@ func (handler *ContainersHandlersImpl) RemoveContainerHandler(params containers.
 		case exec.RemovePowerError:
 			return containers.NewContainerRemoveConflict().WithPayload(&models.Error{Message: err.Error()})
 		default:
+			f, ok := err.(types.HasFault)
+			if ok {
+				switch f.Fault().(type) {
+				case *types.HostNotConnected:
+					return containers.NewContainerRemoveDefault(501).WithPayload(&models.Error{Message: "Couldn't remove container because the host the container is placed on is not connected"})
```
sflxn: Let's shorten this message and remind the user to try again later.

```
Couldn't remove container.  The ESX host is temporarily disconnected.  Please try again later.
```

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -255,6 +257,13 @@ func (handler *ContainersHandlersImpl) RemoveContainerHandler(params containers.
 		case exec.RemovePowerError:
 			return containers.NewContainerRemoveConflict().WithPayload(&models.Error{Message: err.Error()})
 		default:
+			f, ok := err.(types.HasFault)
+			if ok {
```
sflxn: You can combine these two lines into one.  It's more go-like.

```
if f, ok := err.(types.HasFault); ok {
```

gigawhitlocks: yeah that's just an oversight on my part; will do

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
gigawhitlocks: Roger

gigawhitlocks:
Will add 
```
If a user issues a docker rm and the host ESX disconnects from VC during that operation, we currently
return a nonsensical message.  Now we report the host ESX was disconnected and asks the user to retry later.

Resolves #6240
``` to extended commit description during squash and merge

File: lib/apiservers/portlayer/restapi/handlers/containers_handlers.go [hunk 3]
```diff
@@ -255,6 +257,12 @@ func (handler *ContainersHandlersImpl) RemoveContainerHandler(params containers.
 		case exec.RemovePowerError:
 			return containers.NewContainerRemoveConflict().WithPayload(&models.Error{Message: err.Error()})
 		default:
+			if f, ok := err.(types.HasFault); ok {
+				switch f.Fault().(type) {
+				case *types.HostNotConnected:
+					return containers.NewContainerRemoveDefault(501).WithPayload(&models.Error{Message: "Couldn't remove container.  The ESX host is temporarily disconnected.  Please try again later."})
```
anchal-agrawal: Why are we returning a 501 here?

sflxn: That's a good point.  I was going to comment on this but forgot to do so.  Thanks for reminding me.  I think this needs to be code 500.  The swagger definition should also change from

```
					"500": {
						"description": "server error"
					},
```

to

```
					"500": {
						"description": "server error"
						"schema": {
							"$ref": "#/definitions/Error"
						}
					},
```

container.go will also need to change to accomodate the error.

anchal-agrawal: We should also use a constant from the `http` pkg here.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -255,6 +257,12 @@ func (handler *ContainersHandlersImpl) RemoveContainerHandler(params containers.
 		case exec.RemovePowerError:
 			return containers.NewContainerRemoveConflict().WithPayload(&models.Error{Message: err.Error()})
 		default:
+			if f, ok := err.(types.HasFault); ok {
+				switch f.Fault().(type) {
+				case *types.HostNotConnected:
+					return containers.NewContainerRemoveDefault(500).WithPayload(&models.Error{Message: "Couldn't remove container. The ESX host is temporarily disconnected. Please try again later."})
```
anchal-agrawal: Suggest replacing 500 with the constant from the `http` pkg. With that, I imagine this line will be quite lengthy, so we can declare the error message string in a separate line.

gigawhitlocks: oh right yeah I saw your other comment and forgot :+1:

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -43,6 +43,8 @@ import (
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/uid"
 	"github.com/vmware/vic/pkg/version"
+
+	"github.com/vmware/govmomi/vim25/types"
```
anchal-agrawal: This import should belong in the group right above this line.

lib/apiservers/portlayer/swagger.json
```diff
@@ -1570,7 +1570,10 @@
 						}
 					},
 					"500": {
-						"description": "server error"
+						  "description": "server error",
+              "schema": {
```
anchal-agrawal: Indentation seems off, possibly a Spacemacs issue.

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -31,6 +31,8 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/go-openapi/runtime/middleware"
 
+	"github.com/vmware/govmomi/vim25/types"
```
anchal-agrawal: Go imports nit - `github.com/vmware/govmomi/vim25/types` should be in the same group as the other `vmware` imports.

gigawhitlocks: I thought we made a distinction between vmware imports from this project and external vmware projects (e.g. govmomi) but alright i'll delete the newline beneath it so it's in the same group

lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -255,6 +256,13 @@ func (handler *ContainersHandlersImpl) RemoveContainerHandler(params containers.
 		case exec.RemovePowerError:
 			return containers.NewContainerRemoveConflict().WithPayload(&models.Error{Message: err.Error()})
 		default:
+			if f, ok := err.(types.HasFault); ok {
+				switch f.Fault().(type) {
+				case *types.HostNotConnected:
+					p := &models.Error{Message: "Couldn't remove container. The ESX host is temporarily disconnected. Please try again later."}
+					return containers.NewContainerRemoveDefault(http.StatusInternalServerError).WithPayload(p)
```
sflxn: You should use NewContainerRemoveInternalServerError() hear instead of RemoveDefault.

gigawhitlocks: right right

sflxn:
lgtm.  Remember to squash.

gigawhitlocks:
Absolutely

PR #6514, opened 2017-10-05
NFS fixes for volume store and volume creation

### LINKED ISSUE #6500: Incorrect marshalling of XDR data for NFS mkdir and create file operations
Impact:
Prevents volume store creation from succeeding on some NFS servers (Linux seems to tolerate the trailing data)

Details:
XDR serialization is incorrectly including optional data from discriminated union as described here: https://tools.ietf.org/html/rfc1813#page-27

hickeng:
Updates go-nfs-client to pick up protocol fixes for most paths when dealing
with sparse server replies (discriminated union support).

Adds error logging in nfs volume creation paths.

Tests would require NFS servers that are not Linux kernel based as that impl
seems to be very tolerant and doesn't exploit the sparsity elements of NFS
protocol aggressively.
Testing was performed against NetApp simulator ONTAP9.1

Dropping the requirement for fine detail error reporting via vic-machine.
It's problematic without completing the structured status reporting that's
been deferred.

Notes:
- Observed an NFS3ERR_ACCES failure in local nested deployment if omitting
`uid`, `gid`, and `proto` args. This turns out to be because the default uid/gid
is 1000/1000. The problem here is that this needs to be inferred from the logs
because the ACCES error never propagates to the user directly.

Fixes #6500

### DIFF
--- a/vendor/github.com/vmware/go-nfs-client/nfs/example/test/main.go
+++ b/vendor/github.com/vmware/go-nfs-client/nfs/example/test/main.go
@@ -139,7 +139,7 @@ func main() {
 	}
 
 	if len(outDirs) != baseDirCount {
-		log.Fatalf("directory shoudl be empty of our created files!")
+		log.Fatalf("directory should be empty of our created files!")
 	}
 
 	if err = mount.Unmount(); err != nil {
--- a/vendor/github.com/vmware/go-nfs-client/nfs/example/test/main.go
+++ b/vendor/github.com/vmware/go-nfs-client/nfs/example/test/main.go
@@ -47,6 +47,13 @@ func main() {
 	}
 	defer v.Close()
 
+	// discover any system files such as lost+found or .snapshot
+	dirs, err := ls(v, ".")
+	if err != nil {
+		log.Fatalf("ls: %s", err.Error())
+	}
+	baseDirCount := len(dirs)
+
 	if _, err = v.Mkdir(dir, 0775); err != nil {
 		log.Fatalf("mkdir error: %v", err)
 	}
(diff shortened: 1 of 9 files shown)


vendor/github.com/vmware/go-nfs-client/nfs/example/test/main.go
```diff
@@ -131,15 +138,16 @@ func main() {
 		log.Fatalf("ls: %s", err.Error())
 	}
 
-	if len(outDirs) != 2 {
-		log.Fatalf("directory shoudl be empty!")
+	if len(outDirs) != baseDirCount {
+		log.Fatalf("directory shoudl be empty of our created files!")
```
matthewavery: minor : s/shoudl/should

vendor/github.com/vmware/go-nfs-client/nfs/nfs.go
```diff
@@ -222,7 +264,10 @@ func dialService(addr string, port int) (*rpc.Client, error) {
 				Port: p,
 			}
 
-			client, err = rpc.DialTCP("tcp", ldr, fmt.Sprintf("%s:%d", addr, port))
+			raddr := fmt.Sprintf("%s:%d", addr, port)
+			util.Debugf("Connecting to %s", raddr)
+
+			client, err = rpc.DialTCP("tcp", ldr, raddr)
```
matthewavery: We should probably make a constant at some point for `"tcp"` since it is used a lot. I am surprised that the `rpc` does not already have some thing like this.

vendor/github.com/vmware/go-nfs-client/nfs/rpc/client.go
```diff
@@ -25,6 +26,9 @@ const (
 	Success = iota
```
matthewavery: perhaps somewhere around these const definitions we might make a const for `"tcp"`

vendor/github.com/vmware/go-nfs-client/nfs/rpc/client.go
```diff
@@ -135,11 +142,24 @@ func (c *Client) Call(call interface{}) (io.ReadSeeker, error) {
 		case Success:
 			return res, nil
 		case ProgUnavail:
-			return nil, fmt.Errorf("PROG_UNAVAIL")
+			return nil, fmt.Errorf("rpc: PROG_UNAVAIL - server does not recognize the program number")
 		case ProgMismatch:
-			return nil, fmt.Errorf("rpc: PROG_MISMATCH")
+			return nil, fmt.Errorf("rpc: PROG_MISMATCH - program version does not exist on the")
```
matthewavery: seems like there might need a bit more on the end here. e.g. `program version does not exist on the server`?

matthewavery:
Other than a few very very minor comments it lgtm

matthewavery:
lgtm to the additional error message changes

PR #6524, opened 2017-10-05
Implement REST API handler for getting VCH creation logs [specific ci=Group23-Vic-Machine-Service]

### LINKED ISSUE #6023: Implement an API handler for getting the creation logs for a VCH
Implement an API handler for getting the creation logs for a VCH

AngieCris:
Fixes #6023 

**Http Method**
 `GET`

**API endpoints**
`/container/target/{target}/vch/{vch-id}/log`
`/container/target/{target}/datacenter/{datacenter}/vch/{vch-id}/log`

**Output**
Success: 200, `plain/text`: content of all log files concatenated
(log files for a VCH in datastore folder are all prefixed with `vic-machine-`, for example: `vic-machine_2017-10-04T15:18:01 0000_create_45798.3`, and they are sorted by timestamp)
When datastore is not ready / no log file available: 404

**Remaining issues**
The handler can either be used during VCH creation (for example, when the create process has started but datastore folder is not ready) or after VCH creation completes. 
The integration test only deals with the case when VCH creation completes.


AngieCris:
There are some duplicates in all the handlers (`vch_get.go`, `vch_cert_get.go`, `vch_log_get.go`, etc.) and some of the boilerplates can be abstracted into `common.go`, like validation, getting VM, getting config, getting datastore, etc.

lib/apiservers/service/swagger.json
```diff
@@ -900,6 +937,15 @@
       "description": "A VCH host certificate",
       "schema": { "$ref": "#/definitions/PEM" }
     },
+    "vch-log": {
+      "description": "Log messages for a VCH",
+      "schema": {
+        "type": "string"
+      },
+      "examples": {
+        "text/plain": "Sep  8 2017 09:41:56.772-05:00 INFO  ### Installing VCH ####"
```
zjs: A real example would be really long, so I think I'd just leave the `examples` section off of `vch-log`.

lib/install/vchlog/vchlogger.go
```diff
@@ -50,7 +50,7 @@ func Init() {
 func Run() {
 	sig := <-signalChan
 	// suffix the log file name with caller operation ID and timestamp
-	logFileName := sig.LogFileName + "_time_" + sig.Timestamp + "_op_" + sig.Operation.ID()
+	logFileName := "vic-machine" + "_" + sig.Timestamp + "_" + sig.Name + "_" + sig.Operation.ID()
```
zjs: While we're adjusting this, maybe add a ".log" at the end? (And then add that to the search logic in the handler, so that if we ever have non-log files in the datastore directory we don't accidentally include them in the response.)

lib/apiservers/service/restapi/handlers/vch_log_get.go
```diff
@@ -0,0 +1,200 @@
...
+		principal.(Credentials).pass,
+		params.Thumbprint,
+		nil,
+		nil)
+	if err != nil {
+		log.Printf("error building data")
+		return operations.NewGetTargetTargetVchVchIDLogDefault(util.StatusCode(err)).WithPayload(&models.Error{Message: err.Error()})
+	}
+
+	log.Printf("done building data")
```
zjs: I think we should remove this.

lib/apiservers/service/restapi/handlers/vch_log_get.go
```diff
@@ -0,0 +1,200 @@
...
+func (h *VCHLogGet) Handle(params operations.GetTargetTargetVchVchIDLogParams, principal interface{}) middleware.Responder {
+	d, err := buildData(params.HTTPRequest.Context(),
+		url.URL{Host: params.Target},
+		principal.(Credentials).user,
+		principal.(Credentials).pass,
+		params.Thumbprint,
+		nil,
+		nil)
+	if err != nil {
+		log.Printf("error building data")
```
zjs: I think we should remove this.

lib/apiservers/service/restapi/handlers/vch_log_get.go
```diff
@@ -0,0 +1,200 @@
...
+	// TODO (angiew): abstract some of the boilerplate into helpers in common.go
+	validator, err := validateTarget(ctx, d)
+	if err != nil {
+		return nil, util.WrapError(400, err)
+	}
+
+	executor := management.NewDispatcher(validator.Context, validator.Session, nil, false)
+	vch, err := executor.NewVCHFromID(d.ID)
+	if err != nil {
+		return nil, util.NewError(500, fmt.Sprintf("failed to create VCH %s: %s", d.ID, err))
```
zjs: "failed to create VCH" doesn't seem like quite the right message

lib/apiservers/service/restapi/handlers/vch_log_get.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	return paths, nil
+}
+
+// getContentFromLogFile downloads all log files in the list, concatenates the content of each log file and outputs a string of contents
+func getContentFromLogFile(ctx context.Context, helper *datastore.Helper, paths []string) (string, error) {
+	var buffer bytes.Buffer
+
+	for _, p := range paths {
```
zjs: Can we sort these paths before iterating through them so that we always get the operations in the order that they happened?

zjs:
> [skip ci]

Can we run the vic-machine group for this PR (which will include your new test)?

lib/install/vchlog/vchlogger.go
AngieCris: good idea, thanks

lib/apiservers/service/restapi/handlers/vch_log_get.go
AngieCris: This piece is copied from `getVCHConfig` function in `common.go`
I'll update both places.
What should be a more suitable message? like `VCH not found`?

lib/apiservers/service/restapi/handlers/vch_log_get.go
AngieCris: The log files are automatically sorted by timestamp in datastore, and ls() preserves the order. 
But I should probably double check anyways... I'll add a sort function

AngieCris:
@zjs can we tho? The new test is not in the branch yet. Will a vic-machine CI group trigger the new test in this pr?

lib/apiservers/service/restapi/handlers/common.go
```diff
@@ -136,7 +136,7 @@ func getVCHConfig(op trace.Operation, d *data.Data) (*config.VirtualContainerHos
 	executor := management.NewDispatcher(validator.Context, validator.Session, nil, false)
 	vch, err := executor.NewVCHFromID(d.ID)
 	if err != nil {
-		return nil, util.NewError(500, fmt.Sprintf("failed to create VCH %s: %s", d.ID, err))
+		return nil, util.NewError(500, fmt.Sprintf("Unable to find VCH %s: %s", d.ID, err))
```
zjs: Sorry I missed this on my first pass: should this be a 404?

AngieCris: Yea, 500 isn't the best choice. Since the ID provided by the API client should be an existing one, is 400 bad request more suitable? 
This is copied from `common.go`, if this applies for all API handlers, I'll update `common.go` as well

AngieCris: Oh actually 404 is better. it's "VCH not found", not "invalid VCH ID syntax"

tests/test-cases/Group23-VIC-Machine-Service/23-05-VCH-Logs.robot
```diff
@@ -0,0 +1,56 @@
...
+    Run Keyword If  '%{TEST_DATACENTER}' == '${SPACE}'  Set Environment Variable  TEST_DATACENTER  ${dc}
+    ${dcID}=  Get Datacenter ID
+    ${rc}  ${output}=  Run And Return Rc And Output  curl -s -w "\%{http_code}\n" -X GET "http://127.0.0.1:31337/container/target/%{TEST_URL}/datacenter/${dcID}/vch/${vch-id}/log?thumbprint=%{TEST_THUMBPRINT}" -H "authorization: Basic ${auth}"
+    Set Environment Variable  TEST_DATACENTER  ${orig}
+    [Return]  ${rc}  ${output}
+
+
+*** Test Cases ***
+Get VCH Creation Log succeeds after installation completes
+    Install VIC Appliance To Test Server
```
mhagen-vmware: Please move this to Test Setup

tests/test-cases/Group23-VIC-Machine-Service/23-05-VCH-Logs.md
```diff
@@ -0,0 +1,22 @@
...
+=======
+
+# Purpose:
+To verify vic-machine-server can provide logs for a VCH host when available
+
+# References:
+[1 - VIC Machine Service API Design Doc - VCH Certificate](../../../doc/design/vic-machine/service.md)
+
+# Environment:
+This test runs an external service binary that exposes the vic-machine API.
```
mhagen-vmware: The environment is what you need before you can run this test,
i.e. This test requires that a vSphere server is running and available

tests/test-cases/Group23-VIC-Machine-Service/23-05-VCH-Logs.robot
```diff
@@ -0,0 +1,56 @@
...
+    [Arguments]  ${vch-id}  ${auth}
+    ${rc}  ${output}=  Run And Return Rc And Output  curl -s -w "\%{http_code}\n" -X GET "http://127.0.0.1:31337/container/target/%{TEST_URL}/vch/${vch-id}/log?thumbprint=%{TEST_THUMBPRINT}" -H "authorization: Basic ${auth}"
+    [Return]  ${rc}  ${output}
+
+Curl Datacenter
+    [Arguments]  ${vch-id}  ${auth}
+    ${orig}=  Get Environment Variable  TEST_DATACENTER
+    ${dc}=  Run Keyword If  '%{TEST_DATACENTER}' == '${SPACE}'  Get Datacenter Name
+    Run Keyword If  '%{TEST_DATACENTER}' == '${SPACE}'  Set Environment Variable  TEST_DATACENTER  ${dc}
+    ${dcID}=  Get Datacenter ID
```
mhagen-vmware: It doesn't look like you need any of this above this line.  And then you don't need the set environ variable below either.

mhagen-vmware: Whole keyword should just be:
Get Datacenter ID
Curl command

lib/apiservers/service/restapi/handlers/vch_log_get.go
```diff
@@ -0,0 +1,202 @@
...
+
+	// sort log files based on timestamp
+	sort.Strings(paths)
+
+	for _, p := range paths {
+		reader, err := helper.Download(ctx, p)
+		if err != nil {
+			return "", fmt.Errorf("Unable to download log file %s: %s", p, err)
+		}
+		_, err = buffer.ReadFrom(reader)
```
anchal-agrawal: Minor - this can be expressed concisely as:

```go
if _, err = buffer.ReadFrom(reader); err != nil {
    return "", fmt.Errorf("Error reading from log file: %s", err)
}
```

anchal-agrawal: Minor - this can be expressed concisely as:

```go
if _, err = buffer.ReadFrom(reader); err != nil {
    return "", fmt.Errorf("Error reading from log file: %s", err)
}
```

lib/apiservers/service/restapi/handlers/common.go
```diff
@@ -136,7 +136,7 @@ func getVCHConfig(op trace.Operation, d *data.Data) (*config.VirtualContainerHos
 	executor := management.NewDispatcher(validator.Context, validator.Session, nil, false)
 	vch, err := executor.NewVCHFromID(d.ID)
 	if err != nil {
-		return nil, util.NewError(500, fmt.Sprintf("failed to create VCH %s: %s", d.ID, err))
+		return nil, util.NewError(404, fmt.Sprintf("Unable to find VCH %s: %s", d.ID, err))
```
anchal-agrawal: We should be using constants from the `http` pkg instead of hard numbers. There are several such instances in this PR.

lib/apiservers/service/restapi/handlers/vch_log_get.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/vmware/vic/lib/install/management"
+	"github.com/vmware/vic/lib/install/validate"
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/datastore"
+)
+
+// logFilePrefix is the prefix for file names of all vic-machine log files
+// logFileSuffix is the suffix for file names of all vic-machine log files
+// example: "vic-machine_2017-10-04T15:18:01 0000_create_45798.log"
+//			"vic-machine_2017-10-04T15:20:12 0000_inspect_452847.log"
```
anchal-agrawal: Indentation seems off.

lib/apiservers/service/restapi/handlers/vch_log_get.go
```diff
@@ -0,0 +1,202 @@
...
+const LogFilePrefix = "vic-machine"
```
anchal-agrawal: These `const` declarations can be grouped.

lib/apiservers/service/restapi/handlers/vch_log_get.go
```diff
@@ -0,0 +1,202 @@
...
+		return nil, util.WrapError(400, err)
+	}
+
+	executor := management.NewDispatcher(validator.Context, validator.Session, nil, false)
+	vch, err := executor.NewVCHFromID(d.ID)
+	if err != nil {
+		return nil, util.NewError(404, fmt.Sprintf("Unable to find VCH %s: %s", d.ID, err))
+	}
+
+	err = validate.SetDataFromVM(validator.Context, validator.Session.Finder, vch, d)
```
anchal-agrawal: Ditto, this `err` can be checked in the same line.

lib/apiservers/service/restapi/handlers/vch_log_get.go
```diff
@@ -0,0 +1,202 @@
...
+
+	if len(paths) == 0 {
+		return nil, util.NewError(404, "No log file available in datastore folder")
+	}
+
+	return paths, nil
+}
+
+// getContentFromLogFile downloads all log files in the list, concatenates the content of each log file and outputs a string of contents
+func getContentFromLogFile(ctx context.Context, helper *datastore.Helper, paths []string) (string, error) {
```
anchal-agrawal: `getContentFromLogFiles` would be a better name for this function.

lib/apiservers/service/restapi/handlers/vch_log_get.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/vmware/vic/lib/apiservers/service/restapi/handlers/util"
+	"github.com/vmware/vic/lib/apiservers/service/restapi/operations"
+	"github.com/vmware/vic/lib/install/data"
+	"github.com/vmware/vic/lib/install/management"
+	"github.com/vmware/vic/lib/install/validate"
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/datastore"
+)
+
+// logFilePrefix is the prefix for file names of all vic-machine log files
```
anchal-agrawal: The description comment for a field should be directly above its declaration and not grouped with other such comments.

lib/apiservers/service/restapi/handlers/common.go
AngieCris: thanks! good catch

PR #6527, opened 2017-10-05
Implement tests for the VCH Creation API [specific ci=Group23-Vic-Machine-Service]

zjs:
Implement a suite of positive and negative tests for API-based VCH creation. These tests are not intended to be exhaustive; we can use unit tests for that.

Note that these tests only verify the resulting VCHs using `inspect` and do not verify their functionality using docker. That would be desirable validation, but the existing `Run Regression Tests` keyword (and related code) isn't currently written in a way that allows it to be used by these tests. Exposing the `Run Regression Tests` functionality in a way that more clearly captures the expected set of inputs would be a good first step towards being able to re-use it in this manner.

### DIFF
--- a/tests/test-cases/Group23-VIC-Machine-Service/23-03-VCH-Create.robot
+++ b/tests/test-cases/Group23-VIC-Machine-Service/23-03-VCH-Create.robot
@@ -1,0 +1,265 @@
+# Copyright 2017 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#	http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


tests/test-cases/Group23-VIC-Machine-Service/23-03-VCH-Create.robot
```diff
@@ -0,0 +1,143 @@
...
+Count VCHs
+    Get Path Under Target    vch
+    ${count}=    Run    echo '${OUTPUT}' | jq -r '.vchs | length'
+    ${count}=    Convert To Integer    ${count}
+    [Return]    ${count}
+
+
+*** Test Cases ***
+Create minimal VCH
+    ${before}=    Count VCHs
```
mdharamadas1: I'm not sure if counting number of VCHs is required here but it may create issue when we have these tests running concurrently. Just FYI.

zjs: Good point. I'll remove the counting; I think being able to inspect it is sufficient proof that the VCH was properly created.

zjs:
> Do we have a plan of adding more tests for other create params as well? e.g. syslogs, container-name-convention, etc.

I added verification for the information I could find in `inspect config` output. To verify other settings, I'll probably need to use the inspect API. I'll see how easy that would be to do as a part of this PR.

PR #6528, opened 2017-10-05
Create process group and use as target for signals [specific ci=1-14-Docker-Kill]

### LINKED ISSUE #6199: Regression: VIC power-off failure on stop / kill for unresponsive containers
**User Statement:**

In the situation where a container process is not responding to signals, a stop or a kill should cause the container to be powered off.

**Details:**

Note: This worked in 1.1.1, but does not work in 1.2 rc3.
Note: This appears to only be a problem on ESXi - at least as far as I can reproduce it.

Steps to reproduce: Build a very simple container that will have an unresponsive process: 
```
$ cat Dockerfile
FROM debian
COPY run.sh /
CMD /run.sh
$ cat run.sh
#!/bin/bash
sleep infinity
```
The above will be unresponsive because the script will not pass signals to the process. There are plenty of other reasons though why a process may become unresponsive. 

Then run the container as a daemon and try to stop it:
```
$ docker run --name test -d bensdoings/test
$ docker stop test
Error response from daemon: Server error from portlayer: The attempted operation cannot be performed in the current state (Powered on).
$ docker kill test
Error response from daemon: Cannot kill container ca1: Server error from portlayer: The attempted operation cannot be performed in the current state (Powered on).
$ docker ps
CONTAINER ID        IMAGE               COMMAND                CREATED             STATUS              PORTS               NAMES
ca1f0b049979        bensdoings/test     "/bin/sh -c /run.sh"   2 minutes ago       Up 34 seconds                           silly_swanson
```
The ESXi client reports the issue, but powering off via the UI works just fine. The problem with this is that it becomes impossible to clean up the container from the docker CLI.

hickeng:
This creates a process group for non-tty processes launched by tether and
targets signals as the process group and not just the parent.

It also provides some basic checking in vic-machine for whether the
components launched successfully even in the case of a static IP.

Outstanding:
- [x] regression test
- [X] change toolbox locking, or make killhelper a goroutine to avoid toolbox hang in original error case
  - call is now made by the async toolbox Start logic.

Fixes #6199


lib/tether/tether_linux.go
```diff
@@ -280,6 +281,13 @@ func establishNonPty(session *SessionConfig) error {
 	defer trace.End(trace.Begin("initializing nonpty handling for session " + session.ID))
 	var err error
 
+	// configure a process group so we can kill any decendents
+	if session.Cmd.SysProcAttr == nil {
+		session.Cmd.SysProcAttr = &syscall.SysProcAttr{}
+	}
+	// session.Cmd.SysProcAttr.Setpgid = true
```
chengwang86: remove this?

chengwang86:
Maybe we could add an integration test to avoid regression. It could come in another PR if needed.

lib/tether/tether_linux.go
```diff
@@ -245,6 +245,7 @@ func lookPath(file string, env []string, dir string) (string, error) {
 func establishPty(session *SessionConfig) error {
 	defer trace.End(trace.Begin("initializing pty handling for session " + session.ID))
 
+	// pty.Start creates a process group anyway so no change needed to kill all decendents
```
chengwang86: decendents -> descendents

lib/tether/tether_linux.go
```diff
@@ -280,6 +281,13 @@ func establishNonPty(session *SessionConfig) error {
 	defer trace.End(trace.Begin("initializing nonpty handling for session " + session.ID))
 	var err error
 
+	// configure a process group so we can kill any decendents
```
chengwang86: decendents -> descendents

hickeng:
@chengwang86 yup - meant to note that this is incomplete. Needs the test and to change the toolbox locking so even without the setsid fix we don't get the hang.

lib/install/management/appliance.go
```diff
@@ -1076,6 +1076,20 @@ func (d *Dispatcher) ensureApplianceInitializes(conf *config.VirtualContainerHos
 	// as we're only using it for IP in the success case
 	updateErr := d.applianceConfiguration(conf)
 
+	// confirm components launched correctly
```
hickeng: This was added because I managed to break process launch which prevented vic-init from starting endpoint components, but that went completely unreported by vic-machine

Issue #6482, opened 2017-09-29
vic-machine upgrade only looks at build numbers


sflxn:
As a VIC developer, I should be able to run CI tests locally that performs upgrade and not have vic-machine fail because of build number confusion.

Acceptance Criteria:

- [ ] Local Docker Logs CI test passes when running an RC release

---

VIC's version comparison test only considers build number.  During RC phase, the build number is set to RC1, RC2, etc, which confuses vic-machine upgrade process.  For instance,

```
VCH-0-6683" has same or newer version v0.8.0-7315-c8ac999 than installer version v1.2.0-rc1-0-1efa1ba. No upgrade is available.
```

This problem might only appear when running CI tests (that have an upgrade process) locally.  It also only occur during periods where the build number is not numeric.

mdubya66:
there is an environment variable you can set for local builds to force the build number to some number greater than the last release. this needs to be documented in the Contributing document. @anchal-agrawal can you help?

anchal-agrawal:
Sure, we can document this in this section: https://github.com/vmware/vic/blob/master/README.md#common-build-problems

anchal-agrawal:
@sflxn The steps are:

* Set `BUILD_NUMBER` to a high value at the top of the makefile: `BUILD_NUMBER ?= 9999999999`
* `sudo make distclean && sudo make clean && sudo make all`

anchal-agrawal:
Test doc changes are in #6532 - @sflxn please verify.

zjs:
Evaluating whether an upgrade can be performed using only the build number is a problem not only for developers, but can impact customers as well.

As an example of when it would be a problem: If we ship 1.3.0 in December and then need to ship a 1.2.2 security fix in January, the 1.2.2 build number will be higher than the 1.3.0 build number. As a result, customers would be able to "upgrade" from 1.3.0 to 1.2.2, but not the other way around.

zjs:
Re-closing this; using #6857 to track the new concern.

PR #6532, opened 2017-10-09
[skip ci] Add note about build number in upgrade tests

anchal-agrawal:
This commit adds a troubleshooting note about setting `BUILD_NUMBER`
for the `vic-machine` binary so that upgrade tests that rely on a
larger upgrade-to `vic-machine` build number can pass when the test
is run locally.

Towards #6482


sflxn:
Can you add a comment in our docs' md file?

anchal-agrawal:
@sflxn I added the note to the project's `README.md` as well - https://github.com/vmware/vic/pull/6532/files#diff-04c6e90faac2675aa89e2176d2eec7d8.

sflxn:
lgtm

PR #6556, opened 2017-10-11
Add specific tag to busybox in regression tests

### LINKED ISSUE #6131: Longevity 6.5 - Failed to pull busybox
**VIC version:**
build: 13393

Longevity 6.5 Testing

**Logs:**
[6-5_longevity_vch-0-4372.zip](https://github.com/vmware/vic/files/1257535/6-5_longevity_vch-0-4372.zip)

screen shot 2017-08-28 at 1 36 27 am

gigawhitlocks:
Resolves #6131. We believe that occasionally the image
may be updated upstream, which would cause a layer to change during our pull
when a specific image tag is not specified. Since longevity runs the regression
tests in a loop, it is more likely that this could happen during longevity,
so this change adds a specific tag to the regression tests so that if the
latest tag's images are updated, our longevity tests will not be affected.


gigawhitlocks:
WELP

sflxn:
You need to move the "Resolves ..." to the last line of the commit message with one blank line above it.

Issue #5776, opened 2017-07-20
Docker Compose + Container network + internal bridge network cannot express an open port on the container network


corrieb:
**User Statement:**

As a container developer, network isolation and firewalls are important to me. I want to create a simple classic app in which containers talk on an internal bridge network and one of the containers exposes a port on an external container network. I want to use Compose to achieve this. I do not want to use port mapping for reasons of performance and SPOF.

**Details:**

This is not a complex scenario. However, it is a more realistic real-world scenario than the ones we've been testing and documenting. One of our core value propositions is the ability to use container networks and we've recently introduced a change where ports on a container network need to be explicitly opened using -p. The problem is, Docker Compose doesn't seem to be able to express that. Compose seems to assume you always want port mapping and regardless of the ordering in the Compose file, it will not pass -p to the external network.

This is partly a limitation in the way that the multiple network support in compose was designed - there is no obvious way to assign port mappings to a particular network.

I've spent hours trying to figure out a workaround here. I've raised a issue with the compose team here: https://github.com/docker/compose/issues/5042 to at least see if there's a way to influence the ordering.

Bottom line is that as it stands, docker-compose does not work for a very important scenario. See example compose file here:

```
version: '2'

services:
   db:
     image: mysql:5.7
     command: --datadir=/var/lib/mysql/data
     volumes:
       - db-data:/var/lib/mysql
     networks:
       - db-net
     environment:
       MYSQL_ROOT_PASSWORD: somewordpress
       MYSQL_DATABASE: wordpress
       MYSQL_USER: wordpress
       MYSQL_PASSWORD: wordpress

   wordpress:
     depends_on:
       - db
     image: wordpress:latest
     ports:
       - "80"
     volumes:
       - html-data:/var/www/html
     networks:
       - db-net
       - outside
     environment:
       WORDPRESS_DB_HOST: db:3306
       WORDPRESS_DB_USER: wordpress
       WORDPRESS_DB_PASSWORD: wordpress

volumes:
    db-data:
       driver: "vsphere"
       driver_opts:
          Capacity: "4G"
          VolumeStore: "default"
    html-data:
       driver: "vsphere"
       driver_opts:
          Capacity: "2G"
          VolumeStore: "default"

networks:
    outside:
       external:
          name: ExternalNetwork
    db-net:
       internal: true

```

**Acceptance Criteria:**

[...]

PR #6557, opened 2017-10-12
Blend dynamic and static hostfile entries and fix DHCP config

### LINKED ISSUE #6515: Assigning container network IP via DHCP causes /etc/hosts customisation to revert
VIC Engine - vic_1.2.1-rc2, vic_1.2.1

User statement: As a customer of VIC, I would like to set custom entries in /etc/hosts due to application expectations.

Details: Current versions do not implement run --add-host. However, this is tweaked using a workaround; refer https://github.com/vmware/vic/issues/4195#issuecomment-285645969. You may also refer #5833 .

Issue Description: This worked fine so long as the VCH was built with --cnr <range>. However, switching to DHCP fails this immediately. /etc/hosts are reverted to original within few seconds of editing.

Easily reproducible. Refer screenshot.

image

@hickeng

hickeng:
* Allow a user to construct a base /etc/hosts for a container and have those entries blended
with entries generated by tether.
* Update the base set of addresses with some common entries and supports multiple occurrences of a hostname if split across IPv4 and IPv6 addresses.
* Retry DHCP lease renewal on error
* Add iptables rule for DHCP when needed
* Populate `docker ps` output with correct IP address to reflect ports published via container network

TODO:
- [ ] public/private IP transition which breaks from alias set - this is not actually needed at this time so skipping
- [x] complete unit tests
- [x] add robot test
- [x] consider when we should display IP addresses for container-network other than with -p - for the moment I've decided not to provide it in other scenarios; the only one that's really applicable is Open firewall and there we'd have to somehow omit the ports.

Additional work for the port forwarding:
- associate forwarded ports with an external scope network irrespective of network connect ordering (see #5776) so long as network is configured with Published or Open trust levels. Not going to be in this PR.

Fixes #6515


lib/etcconf/hosts.go
```diff
@@ -163,31 +208,110 @@ func (h *hosts) SetHost(hostname string, ip net.IP) {
...
+
+	if hentry != nil {
+		// existing entry with no changes
+		if bytes.Equal(hentry.IP, ip) {
+			h.dirty = false
+			return
+		}
+
+		// existing hostname with a new address - change address for all assocated hostnames
+		hentry.setAddress(ip)
```
hmahmood: Why not just remove the host from the entry instead since it is associated with a different IP?

hickeng: because someone may have entered an extra host alias for this entry.
For example if you have the following entry:
```
10.118.20.236 container-network.localnet my-custom-alias-for-app
```
and the DHCP lease changes, then you can either:
a. edit the IP address (what this change does)
b. create a new entry and move _some_ of the hosts

So our choices are:
```
# edit the IP
10.118.22.33 container-network.localnet my-custom-alias-for-app
```
or
```
# split the entry
10.118.20.236 my-custom-alias-for-app
10.118.22.33 container-network.localnet
```

In the latter case we've likely broken some assumptions and quite possibly are still referencing an address that no longer belongs to that container.

tests/resources/VCH-Util.robot
```diff
@@ -190,11 +190,11 @@ Install VIC Appliance To Test Server
...
+    ${output}=  Run Keyword Unless  ${certs}  Run  ${vic-machine} create --debug ${debug} --name=%{VCH-NAME} --target=%{TEST_URL}%{TEST_DATACENTER} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --image-store=%{TEST_DATASTORE} --appliance-iso=${appliance-iso} --bootstrap-iso=${bootstrap-iso} --password=%{TEST_PASSWORD} --force=true --bridge-network=%{BRIDGE_NETWORK} --public-network=%{PUBLIC_NETWORK} --compute-resource=%{TEST_RESOURCE} --timeout %{TEST_TIMEOUT} --insecure-registry harbor.ci.drone.local --volume-store=%{TEST_DATASTORE}/%{VCH-NAME}-VOL:${vol} --container-network=%{PUBLIC_NETWORK}:public --no-tlsverify ${additional-args}
```
mhagen-vmware: IDK if we want to specify container network as the default for all of our tests?  Is this really the recommended operation for all of our users now?

tests/test-cases/Group0-Bugs/6515.robot
```diff
@@ -0,0 +1,44 @@
...
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} cp ${CURDIR}/6515.resource ${id}:/etc/hosts
+    Should Be Equal As Integers  ${rc}  0
+
+    ${rc}=  Run And Return Rc  docker %{VCH-PARAMS} start ${name}
+    Should Be Equal As Integers  ${rc}  0
+
+    ${rc}  ${hosts}=  Run And Return Rc And Output  docker %{VCH-PARAMS} exec ${id} cat /etc/hosts
+    Should Be Equal As Integers  ${rc}  0
+    Log  ${hosts}
+    Should Contain  ${hosts}  robot-test-alias
```
mhagen-vmware: Can we throw in a docker restart and cat /etc/hosts again as well?

tests/resources/VCH-Util.robot
hickeng: Using a container network has _always_ been the recommended mechanism for exposing services. Port forwarding via the endpointVM is purely for docker compatibility.
I figure that adding a container network to our default install at least makes it easier to write tests that reflect that usage.

PR #6561, opened 2017-10-13
[specific ci=Group1-Docker-Commands] Ensure unique container names during create

### LINKED ISSUE #6261: non-unique container names
We do not guarantee unique names for parallel create operations, only once create has completed:

```
vagrant@devbox:~/vicsmb/src/github.com/vmware/vic$ docker create -it --name=test alpine & docker create -it --name=test alpine
[1] 14943
04d490689cb4c44d04384f723d2f5005dfad4c493a00136330d92610db442014
7832e805e4afe92474a312a8799b960a59861333c29c8574a9d6469aff0e006d
[1]+  Done                    docker create -it --name=test alpine
vagrant@devbox:~/vicsmb/src/github.com/vmware/vic$ docker ps -a
04d490689cb4        alpine              "/bin/sh"           5 seconds ago       Created                                 test
7832e805e4af        alpine              "/bin/sh"           5 seconds ago       Created                                 test
```

and then, after deleting one, the other cannot be removed by name:
```
vagrant@devbox:~/vicsmb/src/github.com/vmware/vic$ docker rm -v test
Error response from daemon: No such container: test
vagrant@devbox:~/vicsmb/src/github.com/vmware/vic$ docker ps -a
04d490689cb4        alpine              "/bin/sh"           2 minutes ago       Created                                 test
```

I suggest using the `ReserveName` mechanism from the container.Rename path, and adding a similar mechanism for the generated name path as well.

anchal-agrawal:
This commit modifies the container create logic in the docker persona
to ensure that containers have unique names, specially during a
parallel container create operation with the same name.

Earlier, the duplicate name check took place at the start of the
create operation, but the name wasn't tracked in the persona cache
until the create operation had succeeded in the portlayer. This made
it possible for multiple containers to have the same name if created
in parallel. Now, the container create function uses the ReserveName
function in the persona's container cache once the container object
representation has been created in the persona. If the container
fails to be created in the portlayer, the name is released.

Fixes #6261

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -528,13 +528,6 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (contain
 
 	var err error
 
-	// bail early if container name already exists
-	if exists := cache.ContainerCache().GetContainerByName(config.Name); exists != nil {
-		err := fmt.Errorf("Conflict. The name %q is already in use by container %s. You have to remove (or rename) that container to be able to re use that name.", config.Name, exists.ContainerID)
-		log.Errorf("%s", err.Error())
-		return containertypes.ContainerCreateCreatedBody{}, derr.NewRequestConflictError(err)
-	}
-
 	// get the image from the cache
 	image, err := cache.ImageCache().Get(config.Config.Image)
 	if err != nil {
@@ -558,9 +551,15 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (contain
 		return containertypes.ContainerCreateCreatedBody{}, err
 	}
 
+	// Reserve the container name to prevent duplicates during a parallel operation.
+	if err := cache.ContainerCache().ReserveName(container, config.Name); err != nil {
+		return containertypes.ContainerCreateCreatedBody{}, derr.NewRequestConflictError(err)
+	}
+
 	// Create an actualized container in the VIC port layer
 	id, err := c.containerCreate(container, config)
 	if err != nil {
+		cache.ContainerCache().ReleaseName(config.Name)
 		return containertypes.ContainerCreateCreatedBody{}, err
 	}
 
--- a/tests/test-cases/Group1-Docker-Commands/1-04-Docker-Create.robot
+++ b/tests/test-cases/Group1-Docker-Commands/1-04-Docker-Create.robot
@@ -208,3 +208,34 @@ Create disables VC destroy
     Should Be Equal As Integers  ${rc}  0
     Run Keyword If  '%{HOST_TYPE}' == 'VC'  Should Contain  ${output}  Destroy_Task
     Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Should Not Contain  ${output}  Destroy_Task
+
+Parallel creates with same container name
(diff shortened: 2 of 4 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -528,13 +528,6 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (contain
 
 	var err error
```
sflxn: If you remove the check for name conflict, what do we report when someone does docker create --name "..." that conflicts?

anchal-agrawal: We return the same error - it is returned from the `ReserveName` function: https://github.com/vmware/vic/blob/master/lib/apiservers/engine/backends/cache/container_cache.go#L178

The only difference between this error message string and the one in the check I'm removing is that `Conflict` is not capitalized in the former.

sflxn:
lgtm

tests/test-cases/Group1-Docker-Commands/1-04-Docker-Create.robot
```diff
@@ -208,3 +208,35 @@ Create disables VC destroy
     Should Be Equal As Integers  ${rc}  0
     Run Keyword If  '%{HOST_TYPE}' == 'VC'  Should Contain  ${output}  Destroy_Task
     Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Should Not Contain  ${output}  Destroy_Task
+
+Parallel creates with same container name
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull ${busybox}
+    Should Be Equal As Integers  ${rc}  0
+
+    ${pid1}=  Start Process  docker %{VCH-PARAMS} create --name testDuplicates ${busybox}  shell=True
```
mdharamadas1: I would suggest to create a unique random suite variable instead of this fixed and repetitive `testDuplicates` to avoid typo issues and other issues involving concurrent test executions later.

anchal-agrawal: Good point - fixed.

tests/test-cases/Group1-Docker-Commands/1-04-Docker-Create.robot
```diff
@@ -208,3 +208,33 @@ Create disables VC destroy
     Should Be Equal As Integers  ${rc}  0
     Run Keyword If  '%{HOST_TYPE}' == 'VC'  Should Contain  ${output}  Destroy_Task
     Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Should Not Contain  ${output}  Destroy_Task
+
+Parallel creates with same container name
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull ${busybox}
+    Should Be Equal As Integers  ${rc}  0
+
+    ${suffix}=  Evaluate  '%{DRONE_BUILD_NUMBER}-' + str(random.randint(1000,9999))  modules=random
```
mdharamadas1: This still doesn't avoid typo issues. I think the following would be more maintenance friendly:
```
${suffix}=  Generate Random String  8
Set Test Variable  $container_name  testDuplicates-${suffix}

```

anchal-agrawal: Fixed!

PR #6562, opened 2017-10-13
[specific ci=1-23-Docker-Inspect] Fix 1-23-Docker-Inspect test for local test runs

### LINKED ISSUE #6432: 1-23-Docker-Inspect fails against vCenter
- Docker inspect image specifying incorrect type

anchal-agrawal:
This commit modifies the 'Docker inspect image specifying incorrect
type' integration test case to expect the correct error output from
a docker inspect operation depending on whether the test is being run
locally or in CI. The test will now expect the Harbor image name in the
error output only when the test is run in CI.

Fixes #6432


sflxn:
lgtm

PR #6572, opened 2017-10-17
Switch to using VCH resource pool by default [skip ci]

### LINKED ISSUE #6471: vic-machine: switch default behaviour to use a resource pool instead of a vApp by default
Follow on to bug1952991 (concurrent power on of mutliple VMs inside a vApp on versions prior to vSphere6.5). The only viable workaround for VIC on pre6.5 was not to use a vApp.

The primary reason for using a vApp was to provide the following:
1. a visual differentiation from a regular resource pool
2. a reminder that the VMs within the VCH should be managed via the VCH, not via the vSphere UI directly

Now that we have addressed the most dangerous issue relating to direct manipulation of cVMs (#2928) the significance of (2) is somewhat reduced.
My inclination is to switch to using a resource pool on all vSphere versions (make `--use-rp` the default):
* consistency
* reduces the scope of the change
* easier to document

Estimate is (2) because of changes needed to test assets rather than complexity of product code change.

jzt:
This change makes vic-machine use a resource pool instead of a vApp.
This is equivalent to using the `--use-rp` flag in previous versions, only
that flag has been removed along with vApp support altogether.

Fixes #6471


jzt:
@mhagen-vmware CI looks good. I'm setting up a scenario to test with a full CI against VC, should be kicked off tomorrow.

mdubya66:
can we get this in ASAP

jzt:
Setting `[skip ci]` while I update a bunch of tests.

jzt:
All tests updated (I think).. running a full CI against VC in nimbus right now. Once that passes, I will re-run a full CI against ESX in OVH.

jzt:
The tests croaked midway through on the client side. I'm restarted them now. They will be run from an ubuntu executor on a nimbus VC against another nimbus VC. I'll provide results as soon as I get them.

PR #6585, opened 2017-10-18
Automate longevity test deployment and execution

### LINKED ISSUE #6131: Longevity 6.5 - Failed to pull busybox

gigawhitlocks:
Small change in this PR, big change in how we run Longevity tests. The following scripts are available on the Longevity box, and maybe they should be checked in also (**edit:** I've gone ahead and checked these in under `tests/longevity-tests`), but I'm going to just post them as such for now so that can be decided. I'll add them to this PR if requested, with the secrets censored out like they are here, but of course on the actual infra we have the secrets in place.

In the PR itself, I've done a small refactor on dynamic_vars.py to allow 2 things:

1) new images can be added by just adding them to the list and calling it a day. they now will get assigned to the appropriate variable names dynamically.

2) since we also have a list of images after this change, I can use the list to populate Harbor with the necessary images in Longevity, dynamically. If we add or remove images, nothing will need to be done to update Longevity.

Also I have updated the howto in Confluence to reference the new way of doing this. 

Resolves #6131 (final commit message will reflect this) by moving our Longevity procedure from manual to automated and adds deployment of Harbor to that procedure, which should prevent us from having layers changed upstream during pull events, which is what we believe caused the hanging behavior witnessed in that issue.

(**edit** scripts were in the description but since I stripped the secrets and checked them in, they're in the diff now)

### DIFF
--- a/tests/longevity-tests/get-and-start-harbor.bash
+++ b/tests/longevity-tests/get-and-start-harbor.bash
@@ -1,0 +1,40 @@
+#!/bin/bash
+set -e
+
+if [ $# -ne 1 ]; then
+    echo "Usage: $0 harbor-version"
+    exit 1
+fi
+version=$1
+pushd /home/$USER
+[ -e harbor ] \
+    && echo "/home/$USER/harbor exists. Delete it if you want to install a newer version and re-run $0" \
+    && pushd harbor && docker-compose start && popd && exit 0
+
+echo "Pulling down version ${version} of Harbor..."
+wget https://github.com/vmware/harbor/releases/download/v${version}/harbor-online-installer-v${version}.tgz -qO - | tar xz
+pushd harbor
+echo "Configuring Harbor"
+sed -i.bak 's/hostname = reg.mydomain.com/hostname = harbor.longevity/g' harbor.cfg
+if [[ ! $(grep harbor.longevity /etc/hosts) ]]; then
+    echo "Adding harbor.longevity to /etc/hosts"
+    sudo sh -c 'echo "127.0.0.1  harbor.longevity" >> /etc/hosts'
+fi
+
+echo "Installing & starting Harbor"
+sudo ./install.sh
+popd
+popd
+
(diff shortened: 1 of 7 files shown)


tests/longevity-tests/get-and-start-harbor.bash
```diff
@@ -0,0 +1,40 @@
...
+fi
+
+echo "Installing & starting Harbor"
+sudo ./install.sh
+popd
+popd
+
+echo "Preparing Harbor..."
+echo "Logging in..."
+docker login harbor.longevity --username=admin --password="Harbor12345"
```
gigawhitlocks: this password ain't no secret, so i dont think it matters if it's here

gigawhitlocks:
OK @mhagen-vmware that was all great feedback. I've incorporated your feedback and also I have committed the scripts to the codebase (except secrets, obviously)

sflxn:
lgtm

tests/resources/dynamic-vars.py
```diff
@@ -1,10 +1,28 @@
...
+    else:
+        return TestEnvironment.LOCAL
+
+def getName(image):
+    return {TestEnvironment.DRONE: 'harbor.ci.drone.local/library/{}'.format(image),
+            TestEnvironment.LONGEVITY: 'harbor.longevity/library/{}'.format(image),
+            TestEnvironment.LOCAL: image}[getEnvironment()]
+
+# this global variable (images) is used by the Longevity scripts. If you change this, change those!
+# and don't inline it!
```
gigawhitlocks: note for self, add to comment "but do feel free to add images to the list"

PR #6587, opened 2017-10-18
[specific ci=Group23-VIC-Machine-Service] Accept session clone ticket via request header

### LINKED ISSUE #6033: Update core vic-machine code to allow authentication using a session key, session cookie, or SAML token

jzt:
This change allows the vic-machine-server to accept  a vSphere session clone ticket in the `X-VMWARE-TICKET` header of incoming requests. This ticket is used by govmomi to clone the user's session to allow the vic-machine-server govmomi client to make vSphere requests on the user's behalf.

The govmomi changes in this PR will be moved to a PR in the govmomi repo once I've verified that this is the proper approach.

Fixes #6033

### DIFF
--- a/lib/apiservers/service/restapi/handlers/common.go
+++ b/lib/apiservers/service/restapi/handlers/common.go
@@ -33,26 +33,38 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/vm"
 )
 
-func buildData(ctx context.Context, url url.URL, user string, pass string, thumbprint *string, datacenter *string, computeResource *string) (*data.Data, error) {
+type buildDataParams struct {
+	target          string
+	thumbprint      *string
+	datacenter      *string
+	computeResource *string
+}
+
+func buildData(ctx context.Context, params buildDataParams, principal interface{}) (*data.Data, error) {
 	d := data.Data{
 		Target: &common.Target{
-			URL:      &url,
-			User:     user,
-			Password: &pass,
+			URL: &url.URL{Host: params.target},
 		},
 	}
 
-	if thumbprint != nil {
-		d.Thumbprint = *thumbprint
+	if c, ok := principal.(Credentials); ok {
+		d.Target.User = c.user
+		d.Target.Password = &c.pass
+	} else {
+		d.Target.CloneTicket = principal.(Session).ticket
 	}
 
-	if datacenter != nil {
+	if params.thumbprint != nil {
+		d.Thumbprint = *params.thumbprint
+	}
+
+	if params.datacenter != nil {
 		validator, err := validateTarget(ctx, &d)
 		if err != nil {
 			return nil, util.WrapError(http.StatusInternalServerError, err)
 		}
 
-		datacenterManagedObjectReference := types.ManagedObjectReference{Type: "Datacenter", Value: *datacenter}
+		datacenterManagedObjectReference := types.ManagedObjectReference{Type: "Datacenter", Value: *params.datacenter}
 
 		datacenterObject, err := validator.Session.Finder.ObjectReference(ctx, datacenterManagedObjectReference)
 		if err != nil {
@@ -62,8 +74,8 @@ func buildData(ctx context.Context, url url.URL, user string, pass string, thumb
 		d.Target.URL.Path = datacenterObject.(*object.Datacenter).InventoryPath
 	}
 
-	if computeResource != nil {
-		d.ComputeResourcePath = *computeResource
+	if params.computeResource != nil {
+		d.ComputeResourcePath = *params.computeResource
 	}
 
 	return &d, nil
--- a/lib/apiservers/service/restapi/handlers/vch_cert_get.go
+++ b/lib/apiservers/service/restapi/handlers/vch_cert_get.go
@@ -17,7 +17,6 @@ package handlers
 import (
 	"fmt"
 	"net/http"
-	"net/url"
 
 	"github.com/go-openapi/runtime"
 	"github.com/go-openapi/runtime/middleware"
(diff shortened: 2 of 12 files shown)


lib/apiservers/service/restapi/handlers/common.go
```diff
@@ -152,3 +163,12 @@ func getVCHConfig(op trace.Operation, d *data.Data) (*config.VirtualContainerHos
 
 	return vchConfig, nil
 }
+
+func setAuth(b *BuildDataParams, p interface{}) {
+	if c, ok := p.(Credentials); ok {
+		b.user = c.user
+		b.password = c.pass
+	} else {
+		b.ticket = p.(Session).ticket
```
zjs: Can we be defensive here and make this `} else if s, ok := p.(Session); ok {`?

That way we can have an `else` block so we don't panic if `p` isn't either.

lib/apiservers/service/restapi/handlers/vch_cert_get.go
```diff
@@ -35,21 +35,23 @@ type VCHCertGet struct{}
...
-		url.URL{Host: params.Target},
-		principal.(Credentials).user,
-		principal.(Credentials).pass,
-		params.Thumbprint,
-		nil,
-		nil)
+	op := trace.NewOperation(params.HTTPRequest.Context(), "VCHCertGet: %s", params.VchID)
+
+	b := BuildDataParams{
+		url:        url.URL{Host: params.Target},
```
zjs: Since you're adjusting this pattern anyway, maybe `BuildDataParams` should take a raw target so we can have the `url.URL{Host: params.Target}` happen in a single place (i.e., `buildData`).

lib/apiservers/service/restapi/handlers/common.go
```diff
@@ -33,26 +33,37 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/vm"
 )
 
-func buildData(ctx context.Context, url url.URL, user string, pass string, thumbprint *string, datacenter *string, computeResource *string) (*data.Data, error) {
+type BuildDataParams struct {
```
zjs: I'd only expect this to be used within `handlers`. Does it need to be exported?

lib/apiservers/service/restapi/handlers/vch_cert_get.go
```diff
@@ -35,21 +35,23 @@ type VCHCertGet struct{}
...
 
 func (h *VCHCertGet) Handle(params operations.GetTargetTargetVchVchIDCertificateParams, principal interface{}) middleware.Responder {
-	d, err := buildData(params.HTTPRequest.Context(),
-		url.URL{Host: params.Target},
-		principal.(Credentials).user,
-		principal.(Credentials).pass,
-		params.Thumbprint,
-		nil,
-		nil)
+	op := trace.NewOperation(params.HTTPRequest.Context(), "VCHCertGet: %s", params.VchID)
```
zjs: Can you ensure that anywhere that you're introducing an operation, it replaces all occurrences of `params.HTTPRequest.Context()` (not just the first)?

lib/apiservers/service/swagger.json
```diff
@@ -830,6 +830,12 @@
     }
   },
   "parameters": {
+    "session-ticket": {
```
zjs: This doesn't seem to be used.

lib/apiservers/service/restapi/handlers/common.go
dougm: I think that's a case where you want to panic, as such a case would be a programmer error rather than error due to user input.

jzt: Also, we shouldn't really get this far without one or the other as swagger will throw an auth error (though we probably shouldn't rely on that behavior).

lib/apiservers/service/restapi/handlers/common.go
jzt: Good catch, fixed

lib/apiservers/service/restapi/handlers/vch_cert_get.go
jzt: Good idea, fixed!

lib/apiservers/service/swagger.json
jzt: Hmm yeah, it looks like the ticket need only be used in the session security definition. I'll remove and re-test.

lib/apiservers/service/restapi/handlers/vch_cert_get.go
jzt: Fixed, thanks

lib/apiservers/service/restapi/handlers/common.go
```diff
@@ -33,26 +33,38 @@ import (
...
-	if thumbprint != nil {
-		d.Thumbprint = *thumbprint
+	if c, ok := principal.(Credentials); ok {
+		d.Target.User = c.user
+		d.Target.Password = &c.pass
+	} else {
+		d.Target.CloneTicket = principal.(Session).ticket
+	}
+
+	if params.thumbprint != nil {
```
andrewtchin: what do we do if there is no thumbprint?

zjs: Being able to call the API without passing a thumbprint just requires that the vSphere target has a well-signed, or otherwise trusted, certificate (like calling the CLI without a thumbprint).

PR #6590, opened 2017-10-19
Fix docker login with artifactory

### LINKED ISSUE #5338: Missing scope in bearer auth challenge when fetching from private repository
I'm using both the remote docker client and the management portal to try to pull images from my private artifactory repository but I'm getting

```
sudo docker -H 192.168.X.X:2376 --tls pull company-docker.jfrog.io/image:latest
Error response from daemon: Failed to obtain OAuth endpoint: missing scope in bearer auth challenge
```

Which I believe is the same issue reported in a different project here:
https://github.com/rkt/rkt/issues/2248

The fix they merged in is here:
https://github.com/appc/docker2aci/pull/126

There is very similar code in this project here: https://github.com/vmware/vic/blob/5a932c8d88d46db7c497762ebb545f98213dc39f/pkg/fetcher/fetcher.go#L452

If you can look at the change and possibly merge it into the project that would be of great help to me and others using private repositories like quay.io or artifactory

sflxn:
Artifactory's docker registry does not return the expected HTTP
header after sending it an HTTP HEAD request.  This header contains
the address to go retrieve the auth token.  Artifactory's docker
registry does return the expected header if we use HTTP GET instead.

Resolves #5338

### DIFF
--- a/lib/apiservers/engine/backends/system.go
+++ b/lib/apiservers/engine/backends/system.go
@@ -370,10 +370,17 @@ func (s *System) AuthenticateToRegistry(ctx context.Context, authConfig *types.A
 			InsecureSkipVerify: skipVerify,
 		})
 
-		// Attempt to get the Auth URL from HEAD operation to the registry
-		hdr, err := fetcher.Head(loginURL)
-		if err == nil && fetcher.IsStatusUnauthorized() {
-			authURL, err = fetcher.ExtractOAuthURL(hdr.Get("www-authenticate"), nil)
+		// Attempt to get the Auth URL from a simple ping operation (GET) to the registry
+		hdr, err := fetcher.Ping(loginURL)
+		if err == nil {
+			if fetcher.IsStatusUnauthorized() {
+				log.Debugf("Looking up OAuth URL from server %s", loginURL)
+				authURL, err = fetcher.ExtractOAuthURL(hdr.Get("www-authenticate"), nil)
+			} else {
+				// We're not suppose to be here, but if we do end up here, use the login
+				//	URL for the auth URL.
+				authURL = loginURL
+			}
 		}
 		if err != nil {
 			log.Errorf("Looking up OAuth URL failed: %s", err)
@@ -385,10 +392,12 @@ func (s *System) AuthenticateToRegistry(ctx context.Context, authConfig *types.A
 		// Just check if we get a token back.
 		token, err := fetcher.FetchAuthToken(authURL)
 		if err != nil || token.Token == "" {
-			log.Errorf("Fetch auth token failed: %s", err)
 			// At this point, if a request cannot be solved by a retry, it is an authentication error.
+			log.Errorf("Fetch auth token failed: %s", err)
 			if _, ok := err.(urlfetcher.DoNotRetry); ok {
 				err = fmt.Errorf("Get %s: unauthorized: incorrect username or password", loginURL)
+			} else {
+				err = urlfetcher.AuthTokenError{TokenServer: *authURL}
 			}
 			return "", err
 		}
--- a/pkg/fetcher/errors.go
+++ b/pkg/fetcher/errors.go
@@ -14,7 +14,10 @@
 
 package fetcher
 
-import "fmt"
+import (
+	"fmt"
+	"net/url"
+)
 
 // DoNotRetry is an error wrapper indicating that the error cannot be resolved with a retry.
 type DoNotRetry struct {
@@ -42,4 +45,14 @@ type TagNotFoundError struct {
 
 func (e TagNotFoundError) Error() string {
 	return fmt.Sprintf("image tag not found: %s", e.Err.Error())
+}
+
+// AuthTokenError is returned when authentication with a registry fails
+type AuthTokenError struct {
+	TokenServer url.URL
+	Err         error
+}
+
+func (e AuthTokenError) Error() string {
+	return fmt.Sprintf("Failed to fetch auth token from %s", e.TokenServer.Host)
 }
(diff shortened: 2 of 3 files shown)


lib/apiservers/engine/backends/system.go
```diff
@@ -370,25 +370,31 @@ func (s *System) AuthenticateToRegistry(ctx context.Context, authConfig *types.A
...
-		if err == nil && fetcher.IsStatusUnauthorized() {
-			authURL, err = fetcher.ExtractOAuthURL(hdr.Get("www-authenticate"), nil)
+		// Attempt to get the Auth URL from a simple ping operation (GET) to the registry
+		hdr, err := fetcher.Ping(loginURL)
+		if err == nil {
+			if fetcher.IsStatusUnauthorized() {
+				log.Debugf("Looking up OAuth URL")
+				authURL, err = fetcher.ExtractOAuthURL(hdr.Get("www-authenticate"), nil)
+			} else {
+				// We're not suppose to be here, but if we do end up here, use the login
```
anchal-agrawal: s/suppose/supposed

lib/apiservers/engine/backends/system.go
```diff
@@ -370,25 +370,31 @@ func (s *System) AuthenticateToRegistry(ctx context.Context, authConfig *types.A
...
+				//	URL for the auth URL.
```
anchal-agrawal: Comment indentation seems off.

lib/apiservers/engine/backends/system.go
```diff
@@ -370,25 +370,31 @@ func (s *System) AuthenticateToRegistry(ctx context.Context, authConfig *types.A
...
+				authURL = loginURL
+			}
 		}
 		if err != nil {
 			log.Errorf("Looking up OAuth URL failed: %s", err)
 			return "", err
 		}
 
-		log.Debugf("logging onto %s", authURL.String())
```
anchal-agrawal: Why are we removing this log?

lib/apiservers/engine/backends/system.go
```diff
@@ -370,25 +370,31 @@ func (s *System) AuthenticateToRegistry(ctx context.Context, authConfig *types.A
...
-
 		// Just check if we get a token back.
 		token, err := fetcher.FetchAuthToken(authURL)
 		if err != nil || token.Token == "" {
-			log.Errorf("Fetch auth token failed: %s", err)
```
anchal-agrawal: Why are we removing this log?

pkg/fetcher/errors.go
```diff
@@ -43,3 +46,13 @@ type TagNotFoundError struct {
 func (e TagNotFoundError) Error() string {
 	return fmt.Sprintf("image tag not found: %s", e.Err.Error())
 }
+
+// AuthTokenError is returned when authentication with a registry occurs
```
anchal-agrawal: s/occurs/fails would be more suitable.

pkg/fetcher/errors.go
```diff
@@ -43,3 +46,13 @@ type TagNotFoundError struct {
...
+type AuthTokenError struct {
+	TokenServer url.URL
+	Err         error
+}
+
+func (e AuthTokenError) Error() string {
+	return fmt.Sprintf("Failed to fetch auth token from %s", e.TokenServer.Host)
```
anchal-agrawal: If we're just using the `TokenServer.Host` field, can we store just the `Host` in the `AuthTokenError` struct instead of the entire URL?

pkg/fetcher/fetcher.go
```diff
@@ -375,10 +376,29 @@ func (u *URLFetcher) fetchToString(ctx context.Context, url *url.URL, reqHdrs *h
...
+	defer cancel()
+
+	res, err := ctxhttp.Get(ctx, u.client, url.String())
+	if err != nil {
+		return nil, err
+	}
+	defer res.Body.Close()
+
+	u.StatusCode = res.StatusCode
+
```
anchal-agrawal: Unneeded whitespace.

pkg/fetcher/fetcher.go
```diff
@@ -375,10 +376,29 @@ func (u *URLFetcher) fetchToString(ctx context.Context, url *url.URL, reqHdrs *h
...
+		log.Debugf("header = %#v", res.Header)
+		return res.Header, nil
+	}
+
+	return nil, fmt.Errorf("Unexpected http code: %d, URL: %s", u.StatusCode, url)
+}
+
 // Head sends a HEAD request to url
 func (u *URLFetcher) Head(url *url.URL) (http.Header, error) {
-	defer trace.End(trace.Begin(url.String()))
```
anchal-agrawal: Why are we removing this trace?

lib/apiservers/engine/backends/system.go
```diff
@@ -370,25 +370,31 @@ func (s *System) AuthenticateToRegistry(ctx context.Context, authConfig *types.A
...
 
-		// Attempt to get the Auth URL from HEAD operation to the registry
-		hdr, err := fetcher.Head(loginURL)
-		if err == nil && fetcher.IsStatusUnauthorized() {
-			authURL, err = fetcher.ExtractOAuthURL(hdr.Get("www-authenticate"), nil)
+		// Attempt to get the Auth URL from a simple ping operation (GET) to the registry
+		hdr, err := fetcher.Ping(loginURL)
+		if err == nil {
+			if fetcher.IsStatusUnauthorized() {
+				log.Debugf("Looking up OAuth URL")
```
anchal-agrawal: Should we log the URL too? Since we're removing a log below that does that.

sflxn:
ping

gigawhitlocks:
LGTM but do we have a test for this scenario?

gigawhitlocks:
I guess the better question is: "is it feasible to write a test for this?"

sflxn:
I'm going to write tests once I can get an account set up for the VIC team with VMware's internal artifactory server.

PR #6591, opened 2017-10-19
[skip ci] Keep the nightly secrets oh so secret

### LINKED ISSUE #6583: In Robot files, tag `secret` not secret
Seen here, the `Keyword` `Run Secret SSHPass Command` has tag `secret` but the password is included in test logs:

screen shot 2017-10-18 at 1 52 46 pm

mhagen-vmware:
fixes #6583


zjs:
If there's a chance we previously exposed a password which should be secret, we should change that password.

gigawhitlocks:
:+1:

Issue #6574, opened 2017-10-18
Print drone worker name in CI builds


mdharamadas1:
**User Statement:**

As a developer, I should be able to see drone worker name in the logs on which the CI build executed.

**Details:**
This is already done in #6108, separating out so we can merge this, which is very useful in debugging CI build issues.

**Acceptance Criteria:**
Drone worker / agent name on which build ran is printed in the logs.

mdharamadas1:
@andrewtchin FYI, I'm adding this in current sprint instead of #6108, which is blocked until we upgrade drone itself.

mdharamadas1:
@mhagen-vmware could you pls verify this?

mhagen-vmware:
lgtm, you could have just done fixes #6574 in your PR.

mdharamadas1:
@mhagen-vmware noted, I'm still getting used to GitHub.

PR #6592, opened 2017-10-19
DRONE_MACHINE not needed for local tests

mdharamadas1:
follow up fix for #6574


mdharamadas1:
@anchal-agrawal Thanks, fixed! Also in PR title.

Issue #2739, opened 2016-10-14
Add opid to log messages from our calls to vsphere


mhagen-vmware:
Don't know if govmomi returns the OP ID but the api calls that it is using does according to @derekbeard

It would help future debugging efforts if we added this.

As an engineer on VIC, it would be useful for debugging purposes to include the OP ID on vsphere operations so that we can track operations through vsphere related logs while debugging.

hickeng:
@carlsone Has mentioned the utility of this opid for full stack performance analysis in the past.
The pkg/vsphere/task package was supposed to provide logging for the bridging between VHC operation and vSphere operations.
This should be glued together with the recent pkg/trace operation code.

matthewavery:
I could not find the exact opID listed in the `hostd` operation calls. I have not yet found a way to obtain that opID from the `TaskInfo` object that is returned to us. @derekbeard @hickeng Any advice on where to find it? I went

derekbeard:
I tracked this down a bit more.  I spoke with Kira on our team about how VCQE is able to log the opid.  It turns out, that VC supports client-generated OpIDs, so VCQE generates it and passes it as an “operationID” SOAP element in the SOAP header.  My guess is that we could update the govmomi SOAP client (https://github.com/vmware/govmomi/tree/master/vim25/soap) to pass an OPID generated upstream.

hickeng:
Blocked by vmware/govmomi#749 - zenhub doesn't seem to support cross repo linking for blockers.

PR #6593, opened 2017-10-19
[full ci] Improve exec error message for a CVM shutdown mid operation

### LINKED ISSUE #6370: Docker exec following a docker restart of a node container results in unknown task ID error
```
anon@anon-Latitude-E7450:~/vic$ docker -H 192.168.254.178:2376 --tls restart node1
node1
anon@anon-Latitude-E7450:~/vic$ docker -H 192.168.254.178:2376 --tls exec -it node1 /bin/bash
Server error from portlayer: Unknown task ID: e76d137a9d9d136cffd1353f6c902d571a14b2abdaeb8220868fe0408d607739
Error response from daemon: Server error from portlayer: [GET /tasks][500] inspectInternalServerError  &{Code:0 Message:unknown task ID: e76d137a9d9d136cffd1353f6c902d571a14b2abdaeb8220868fe0408d607739}
```

[container-logs.zip](https://github.com/vmware/vic/files/1307639/container-logs.zip)

matthewavery:
This adds some error types and propagates them back from the portlayer to the personality. TaskInspect now returns a `ConflictError` if the supplied ID does not exist and we see that the state of the cvm is powered off. This error case may be too broad, but in TaskInspect we are not sure if the ID supplied was meant to be an exec ID or a session task ID. This means the solution is best guess, if the id is not found and we are off it is assumed that the operation has been interupted. I thought about just doing the state check again on the personality side of things, however the unknown task ID issue is already triggered after making it past the original state check in CreateExecTask, so I deemed that the task inspect check might need to be closer to the SoT(feel free to correct me on  this as well). 

Additionally, the guest reload config function has been wrapped in a retry that retries on transient errors. We may want to narrow down the potential retry case as I used the waiter.go intermittent error decider function in first pass since it was designed to retry tasks against vsphere that are supposed to be inherently transient. Based on discussion in this PR we may change that. 

I have also added some logging, with some `trace.Operation` this is pending potential removal per @cgtexmex 's review. I added it originally since it made it easier to track what all was getting called when I was learning the Exec path of operations. It was also needed since I needed to track concurrent calls against the same container. 

note: please keep in mind this was my first exposure to the Exec portion of the code base, if I have done something silly please let me know. I spent a decent while poking around and trying understand end to end exactly what was going on. 

Fixes #6370

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -235,11 +235,13 @@ misspell: $(MISSPELL)
 
 govet:
 	@echo checking go vet...
+	@$(GO) tool vet -all -lostcancel -tests $$(find . -mindepth 1 -maxdepth 1 -type d -not -name vendor)
 # 	one day we will enable shadow check
 # 	@$(GO) tool vet -all -shadow -lostcancel -tests $$(find . -mindepth 1 -maxdepth 1 -type d -not -name vendor)
 
 gas: $(GAS)
 	@echo checking security problems
+	@$(GAS) -skip=*_responses.go -quiet lib/... cmd/... pkg/... 2> /dev/null
 
 vendor: $(GVT)
 	@echo restoring vendor
--- a/Makefile
+++ b/Makefile
@@ -235,13 +235,11 @@ misspell: $(MISSPELL)
 
 govet:
 	@echo checking go vet...
-	@$(GO) tool vet -all -lostcancel -tests $$(find . -mindepth 1 -maxdepth 1 -type d -not -name vendor)
 # 	one day we will enable shadow check
 # 	@$(GO) tool vet -all -shadow -lostcancel -tests $$(find . -mindepth 1 -maxdepth 1 -type d -not -name vendor)
 
 gas: $(GAS)
 	@echo checking security problems
-	@$(GAS) -skip=*_responses.go -quiet lib/... cmd/... pkg/... 2> /dev/null
 
 vendor: $(GVT)
 	@echo restoring vendor
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -222,7 +222,6 @@ func (c *Container) TaskInspect(cid, cname, eid string) (*models.TaskInspectResp
 	if err != nil {
 		return nil, err
 	}
-
 	return resp.Payload, nil
 
 }
@@ -259,7 +258,7 @@ func (c *Container) TaskWaitToStart(cid, cname, eid string) error {
 // ContainerExecCreate sets up an exec in a running container.
 func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (string, error) {
 	op := trace.NewOperation(context.TODO(), "")
-	defer trace.End(trace.Begin(op.Sprintf("name=(%s)", name)))
+	defer trace.End(trace.Begin(fmt.Sprintf("%s: name=(%s)", op, name)))
 
 	// Look up the container name in the metadata cache to get long ID
 	vc := cache.ContainerCache().GetContainer(name)
@@ -332,7 +331,7 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
 // command. An error is returned if the exec cannot be found.
 func (c *Container) ContainerExecInspect(eid string) (*backend.ExecInspect, error) {
 	op := trace.NewOperation(context.TODO(), "")
-	defer trace.End(trace.Begin(fmt.Sprintf("opID=(%s) eid=(%s)", op.ID(), eid)))
+	defer trace.End(trace.Begin(fmt.Sprintf("opID=(%s) eid=(%s)", op, eid)))
 
 	// Look up the container name in the metadata cache to get long ID
 	vc := cache.ContainerCache().GetContainerFromExec(eid)
@@ -408,7 +407,7 @@ func (c *Container) ContainerExecResize(eid string, height, width int) error {
 // std streams are set up.
 func (c *Container) ContainerExecStart(ctx context.Context, eid string, stdin io.ReadCloser, stdout io.Writer, stderr io.Writer) error {
 	op := trace.NewOperation(ctx, "")
-	defer trace.End(trace.Begin(fmt.Sprintf("opID=(%s) eid=(%s)", op.ID(), eid)))
+	defer trace.End(trace.Begin(fmt.Sprintf("opID=(%s) eid=(%s)", op, eid)))
 
 	// Look up the container name in the metadata cache to get long ID
 	vc := cache.ContainerCache().GetContainerFromExec(eid)
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -309,10 +309,10 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
 	if err != nil {
 		switch err := err.(type) {
 		case *tasks.InspectInternalServerError:
-			log.Debugf("received an internal server error during task inspect: %s", err.Payload.Message)
+			op.Debugf("received an internal server error during task inspect: %s", err.Payload.Message)
 			return "", InternalServerError(err.Payload.Message)
 		case *tasks.InspectConflict:
-			log.Debugf("received a conflict error during task inspect: %s", err.Payload.Message)
+			op.Debugf("received a conflict error during task inspect: %s", err.Payload.Message)
 			return "", ConflictError(fmt.Sprintf("Cannot complete the operation, container %s has been powered off during execution", id))
 		default:
 			return "", InternalServerError(err.Error())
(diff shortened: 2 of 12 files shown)


matthewavery:
TIL that the docker client calls multiple personality endpoints. So there will still be a concurrent modification error when we report the shutdown on the race condition edge. so now the error message looks like this

lib/portlayer/exec/commit.go
```diff
@@ -210,3 +209,50 @@ func Commit(ctx context.Context, sess *session.Session, h *Handle, waitTime *int
...
+			}
+			return nil
+		}
+
+		// nothing to be done.
+		return nil
+	}
+
+	err := retry.Do(retryFunc, isIntermittentFailure)
+
```
caglar10ur: extra space

lib/apiservers/engine/backends/container.go
```diff
@@ -271,15 +273,19 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
 	if err != nil {
 		return "", InternalServerError(err.Error())
 	}
+
+	// This does not appear to be working... we may need to do this check further down the stack. it is going to be very race filled.
```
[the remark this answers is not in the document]
matthewavery: ah that was just an investigative comment :) I thought it was removed sorry :P I was thinking that it was an early check on state that was far from the source of truth. But I figured it is more of a soft check anyways in case the user tried to exec an already off container.

Makefile
```diff
@@ -235,13 +235,11 @@ misspell: $(MISSPELL)
 
 govet:
 	@echo checking go vet...
-	@$(GO) tool vet -all -lostcancel -tests $$(find . -mindepth 1 -maxdepth 1 -type d -not -name vendor)
```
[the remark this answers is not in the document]
matthewavery: We are not :( I had that removed since mine misbehaves and it snuck in... this happens to me occasionally as I get `add` happy with git.

lib/apiservers/engine/backends/container.go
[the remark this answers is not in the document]
matthewavery: I am going to remove this completely. It is a good check to have and we can always handle things down stream as well.

lib/apiservers/engine/backends/container.go
```diff
@@ -257,7 +258,8 @@ func (c *Container) TaskWaitToStart(cid, cname, eid string) error {
 
 // ContainerExecCreate sets up an exec in a running container.
 func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (string, error) {
-	defer trace.End(trace.Begin(name))
+	op := trace.NewOperation(context.TODO(), "")
+	defer trace.End(trace.Begin(fmt.Sprintf("opid=(%s) name=(%s)", op.ID(), name)))
```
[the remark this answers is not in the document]
matthewavery: like a `op.sprintf`? :) yes I would love to do that! I think I had a PR that never got merged awhile back that had one.

Makefile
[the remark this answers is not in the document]
caglar10ur: OK, then keep it in your local directory. Why did you did a git add and pushed it as a part of this PR?

lib/apiservers/engine/backends/container.go
```diff
@@ -222,6 +222,7 @@ func (c *Container) TaskInspect(cid, cname, eid string) (*models.TaskInspectResp
 	if err != nil {
 		return nil, err
 	}
+
 	return resp.Payload, nil
```
caglar10ur: [minor] extra line :P

lib/apiservers/engine/backends/container.go
```diff
@@ -257,7 +258,8 @@ func (c *Container) TaskWaitToStart(cid, cname, eid string) error {
 
 // ContainerExecCreate sets up an exec in a running container.
 func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (string, error) {
-	defer trace.End(trace.Begin(name))
+	op := trace.NewOperation(context.TODO(), "")
+	defer trace.End(trace.Begin(op.Sprintf("name=(%s)", name)))
```
caglar10ur: op.Sprintf -> fmt.Sprintf

lib/apiservers/engine/backends/container.go
```diff
@@ -381,7 +407,8 @@ func (c *Container) ContainerExecResize(eid string, height, width int) error {
 // ContainerExecStart starts a previously set up exec instance. The
 // std streams are set up.
 func (c *Container) ContainerExecStart(ctx context.Context, eid string, stdin io.ReadCloser, stdout io.Writer, stderr io.Writer) error {
-	defer trace.End(trace.Begin(eid))
+	op := trace.NewOperation(ctx, "")
+	defer trace.End(trace.Begin(fmt.Sprintf("opID=(%s) eid=(%s)", op.ID(), eid)))
```
caglar10ur: you can add a Stringer to op and use op here instead of op.ID()

```
func (o *Operation) String() string {
	return o.id
}
```

lib/apiservers/engine/backends/container.go
matthewavery: wait why? I added the op.Sprintf since i thought that was the `stringer` that you wanted :)

matthewavery: ah I just saw the below

lib/apiservers/engine/backends/container.go
matthewavery: Ah I see. So that it has the same behavior as `error`

lib/apiservers/engine/backends/container.go
matthewavery: hmmm 👍 I can make that change, I see what you meant by stringer now, I still like the Sprintf though unless there is an egregious reason not to have it.

lib/apiservers/engine/backends/container.go
matthewavery: I used `return o.header()` instead of `o.id` :)

lib/apiservers/engine/backends/container.go
caglar10ur: Personally  I don't think having a Sprintf on op makes sense. fmt already provides that and that's the standart way of doing that.

Others can chime in if they think this way or that way :)

lib/apiservers/engine/backends/container.go
matthewavery: I am also changing this to using the op.Sprintf since it will tack the header on by default. I wanted to promote the use of op.Sprintf for providing strings to the `Trace.Begin` and `Trace.End` calls since I still love those and it offers a much cleaner way of adding the Operation header to that logging.

cgtexmex:
From my perspective I wouldn't add `pkg/trace/Operation` functionality in this PR.  #2739 will result in changes to the package and I'll be adding operations in "strategic" places....so to minimize rework / wasted effort I'd vote for you to use `fmt` for now...

matthewavery:
You got it @cgtexmex

cgtexmex:
You have added the creation of an Operation at the beginning of most (if not all) the exec funcs -- please use the `op.Debugf(log message)` consistently.  Specifically thinking about the `type` switches as part of the error handling -- use `op.Debugf` and not `log.Debugf`

cgtexmex:
LGTM

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.robot
```diff
@@ -16,7 +16,6 @@
 Documentation  Test 1-38 - Docker Exec
 Resource  ../../resources/Util.robot
 Suite Setup  Install VIC Appliance To Test Server
-Suite Teardown  Cleanup VIC Appliance On Test Server
```
mhagen-vmware: put that back in

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.robot
```diff
@@ -58,7 +57,7 @@ Exec Echo -t
     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull ${busybox}
     Should Be Equal As Integers  ${rc}  0
     Should Not Contain  ${output}  Error
-    ${rc}  ${id}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -d ${busybox} /bin/top -d 600
+    ${rc}  ${id}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -td ${busybox} /bin/top -d 600
```
mhagen-vmware: don't add -t as this isn't run with a TTY

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.robot
```diff
@@ -123,3 +122,24 @@ Exec NonExisting
     #\   ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} exec ${id} /NonExisting
     #\   Should Be Equal As Integers  ${rc}  0
     #\   Should Contain  ${output}  no such file or directory
+
+Exec During PowerOff
+     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull ${busybox}
+     Should Be Equal As Integers  ${rc}  0
+     Should Not Contain  ${output}  Error
+     ${rc}  ${id}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -d ${busybox} /bin/top -d 600
```
mhagen-vmware: don't think you need -d 600 here.. no reason to create a race here, just let it run

Issue #6019, opened 2017-08-14
Implement end-to-end tests for the API handler for creating a VCH


zjs:
Implement end-to-end tests for the API handler for creating a VCH (#6017)

PR #6607, opened 2017-10-20
Always process ops credentials in VCH creation API [specific ci=Group23-Vic-Machine-Service]

zjs:
This change moves the logic which processes operations credentials up a level so that they will always be processed, even if no endpoint settings are provided.

Note that this was discovered while working on #6019 and will be tested as a part of #6527.

### DIFF
--- a/lib/apiservers/service/restapi/handlers/vch_create.go
+++ b/lib/apiservers/service/restapi/handlers/vch_create.go
@@ -399,9 +399,9 @@ func buildCreate(op trace.Operation, d *data.Data, finder *find.Finder, vch *mod
 					OpsPassword: &opsPassword,
 				}
 			}
-			if err := c.OpsCredentials.ProcessOpsCredentials(true, c.Target.User, c.Target.Password); err != nil {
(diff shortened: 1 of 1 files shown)


lib/apiservers/service/restapi/handlers/vch_create.go
```diff
@@ -398,9 +398,9 @@ func buildCreate(op trace.Operation, d *data.Data, finder *find.Finder, vch *mod
 					OpsPassword: &opsPassword,
 				}
 			}
-			if err := c.OpsCredentials.ProcessOpsCredentials(true, c.Target.User, c.Target.Password); err != nil {
-				return nil, util.WrapError(400, err)
-			}
+		}
+		if err := c.OpsCredentials.ProcessOpsCredentials(true, c.Target.User, c.Target.Password); err != nil {
+			return nil, util.WrapError(400, err)
```
anchal-agrawal: We should use constants from the `net/http` pkg instead of hard numbers.

jzt: @anchal-agrawal these are being fixed in a separate PR.

PR #6608, opened 2017-10-21
Propagate validation in VCH creation API [specific ci=Group23-Vic-Machine-Service]

### LINKED ISSUE #6547: Clearly relay validator errors during API-based VCH creation
Currently, validator errors during API-based VCH creation are essentially swallowed. This makes it very hard for a user of the API to determine what went wrong during a VCH creation request.

Simply returning all issues raised by the validator as a part of an HTTP error doesn't seem to be the right solution as many issues seem to be more of a warning than an error.

Reporting errors to the user without failing unnecessarily may require some changes to the validator logic itself to distinguish these cases. Alternatively, maybe the API-based creation workflow can be adjusted so that it does not produce extraneous warnings.

cc @cristianfalcone

zjs:
As a part of handling a VCH creation API request, we pass the desired VCH configuration to the validator code.

When the validator returns an error, we should report it to the user.

This will be tested as a part of #6019.

Fixes #6547

### DIFF
--- a/lib/apiservers/service/restapi/handlers/vch_create.go
+++ b/lib/apiservers/service/restapi/handlers/vch_create.go
@@ -426,6 +426,16 @@ func buildCreate(op trace.Operation, d *data.Data, finder *find.Finder, vch *mod
 
 func handleCreate(op trace.Operation, c *create.Create, validator *validate.Validator) (*strfmt.URI, error) {
 	vchConfig, err := validator.Validate(validator.Context, c.Data)
+	if err != nil {
+		issues := validator.GetIssues()
+		messages := make([]string, 0, len(issues))
+		for _, issue := range issues {
+			messages = append(messages, issue.Error())
+		}
+
+		return nil, util.NewError(400, fmt.Sprintf("Failed to validate VCH: %s", strings.Join(messages, ", ")))
+	}
+
 	vConfig := validator.AddDeprecatedFields(validator.Context, vchConfig, c.Data)
 
 	// TODO: make this configurable


lib/apiservers/service/restapi/handlers/vch_create.go
```diff
@@ -426,6 +426,16 @@ func buildCreate(op trace.Operation, d *data.Data, finder *find.Finder, vch *mod
...
 func handleCreate(op trace.Operation, c *create.Create, validator *validate.Validator) (*strfmt.URI, error) {
 	vchConfig, err := validator.Validate(validator.Context, c.Data)
+	if err != nil {
+		issues := validator.GetIssues()
+		messages := make([]string, 0, len(issues))
+		for _, issue := range issues {
+			messages = append(messages, issue.Error())
+		}
+
+		return nil, util.NewError(400, fmt.Sprintf("Failed to validate VCH: %s", strings.Join(messages, ", ")))
```
anchal-agrawal: We should use constants from the `net/http` pkg instead of hard numbers.

lib/apiservers/service/restapi/handlers/vch_create.go
```diff
@@ -426,6 +426,16 @@ func buildCreate(op trace.Operation, d *data.Data, finder *find.Finder, vch *mod
 
 func handleCreate(op trace.Operation, c *create.Create, validator *validate.Validator) (*strfmt.URI, error) {
 	vchConfig, err := validator.Validate(validator.Context, c.Data)
+	if err != nil {
+		issues := validator.GetIssues()
+		messages := make([]string, 0, len(issues))
+		for _, issue := range issues {
```
AngieCris: just a suggestion: will it be neater if we move the logic of appending multiple errors to `util`? issues is just a list of `error`

zjs: I think my preference would be to leave the functionality where it is for now, and refactor when we actually have the second use case.

I suspect we'll want to deduplicate a lot of logic between VCH creation and VCH re-configuration, but I don't want to presuppose any specific architecture.

Issue #6505, opened 2017-10-04
VCH-Update: implementation of create/validate/repair of vSphere Permissions for ops-user and VCH resources


lcastellano:
**User Statement:**
When a VCH is updated the ops-user should have Permissions on the various vSphere resources used by the VCH.
**Details:**
During VCH update the set of vSphere Permissions should be added for the appropriate Roles of the ops-user should be updated as well. The Permissions are applied to different vSphere managed resources that the VCH need to access and modify. The issue should track:

- Implementation of the code that updates the right Permission given the set of resources and the ops-user username,
- Unit tests,
- Manual basic integration testing with vCenter

**Acceptance Criteria:**
Code and tests should be implemented.

anchal-agrawal:
@lcastellano Please add a priority and team label(s), thanks.

PR #6617, opened 2017-10-24
Add function that returns the Datacenter from a Resource Pool

lcastellano:
Implementation of a function that retrieves the Datacenter object from a Resource Pool object. The function walks the tree "upwards" until it finds the first Datacenter. This is a utility function to implement the feature described in #6505.


pkg/vsphere/compute/rp.go
```diff
@@ -88,3 +88,51 @@ func (rp *ResourcePool) GetCluster(ctx context.Context) (*object.ComputeResource
...
+	client := rp.Session.Vim25()
+
+	ancestors, err := mo.Ancestors(ctx, client, client.ServiceContent.PropertyCollector, rp.Reference())
+
+	if err != nil {
+		log.Errorf("Unable to get ancestors of rp %s: %s", rp.Name(), err)
+		return nil, err
+	}
+
+	outAncestors := make([]types.ManagedObjectReference, 0)
```
hickeng: you may as well make the last parameter to make be the maximum length - just polite as Go has (had?) issues with memory fragmentation.
```
make([]types.ManagedObjectReference, 0, len(ancestors)) 
```

pkg/vsphere/compute/rp.go
```diff
@@ -88,3 +88,51 @@ func (rp *ResourcePool) GetCluster(ctx context.Context) (*object.ComputeResource
...
+		if ancestor.Self.Type == inType {
+			a := ancestor.Self
+			outAncestors = append(outAncestors, a)
+		}
+	}
+
+	return outAncestors, nil
+}
+
+func (rp *ResourcePool) getLowestAncestor(ctx context.Context, inType string) (*types.ManagedObjectReference, error) {
```
hickeng: This could do with a comment or possible a different name. I interpret lowest to mean closest to the leaf (`rp`), whereas this is returning the furthest from the leaf if I'm reading the `mo.Ancestors` code correctly.

pkg/vsphere/compute/rp.go
```diff
@@ -88,3 +88,51 @@ func (rp *ResourcePool) GetCluster(ctx context.Context) (*object.ComputeResource
...
+	}
+
+	return object.NewDatacenter(rp.Client.Client, *dcRef), nil
+}
+
+func (rp *ResourcePool) getAncestors(ctx context.Context, inType string) ([]types.ManagedObjectReference, error) {
+	client := rp.Session.Vim25()
+
+	ancestors, err := mo.Ancestors(ctx, client, client.ServiceContent.PropertyCollector, rp.Reference())
+
```
anchal-agrawal: Unneeded whitespace.

pkg/vsphere/compute/rp.go
```diff
@@ -88,3 +88,51 @@ func (rp *ResourcePool) GetCluster(ctx context.Context) (*object.ComputeResource
...
+			outAncestors = append(outAncestors, a)
+		}
+	}
+
+	return outAncestors, nil
+}
+
+func (rp *ResourcePool) getLowestAncestor(ctx context.Context, inType string) (*types.ManagedObjectReference, error) {
+	ancestors, err := rp.getAncestors(ctx, inType)
+
```
anchal-agrawal: Unneeded whitespace.

pkg/vsphere/compute/rp.go
```diff
@@ -88,3 +88,51 @@ func (rp *ResourcePool) GetCluster(ctx context.Context) (*object.ComputeResource
 
 	return object.NewComputeResource(rp.Client.Client, mrp.Owner), nil
 }
+
+func (rp *ResourcePool) GetDatacenter(ctx context.Context) (*object.Datacenter, error) {
+	dcRef, err := rp.getLowestAncestor(ctx, "Datacenter")
+
+	if err != nil || dcRef == nil {
+		log.Errorf("Unable to get datacenter ancestor of rp %s: %s", rp.Name(), err)
```
anchal-agrawal: Should this be `..., err.Error())`?

pkg/vsphere/compute/rp.go
```diff
@@ -88,3 +88,51 @@ func (rp *ResourcePool) GetCluster(ctx context.Context) (*object.ComputeResource
...
+
+	if err != nil {
+		log.Errorf("Unable to get ancestors of rp %s: %s", rp.Name(), err)
+		return nil, err
+	}
+
+	outAncestors := make([]types.ManagedObjectReference, 0)
+	for _, ancestor := range ancestors {
+		if ancestor.Self.Type == inType {
+			a := ancestor.Self
```
anchal-agrawal: A few notes about this `for` loop:

* `a` is an ambiguous name - doesn't matter a lot currently since it's only used right below, but it will be if used elsewhere in the future.
* `ancestor.Self` is looked up twice (lines 115 and 116). I think it'd be better to declare a var for `ancestor.Self` at the top of the `for` loop, and use that inside.

pkg/vsphere/compute/rp.go
lcastellano: Actually it builds a list from the top of the tree. The top of the tree is element 0. The last element is closest to the leaf.

pkg/vsphere/compute/rp.go
lcastellano: Not sure on which line.

pkg/vsphere/compute/rp.go
lcastellano: Is the unneeded whitespace between the two return values? Go fmt works on this set of changes. I would be wary of removing whitespaces if go fmt doesn't complain.

pkg/vsphere/compute/rp.go
lcastellano: Done

pkg/vsphere/compute/rp.go
lcastellano: Done

pkg/vsphere/compute/rp.go
anchal-agrawal: Line 107. I think it's a good practice to do this
```
err := funcErr()
if err != nil {
```

instead of
```
err := funcErr()

if err != nil {
```

pkg/vsphere/compute/rp.go
anchal-agrawal: There's unneeded whitespace between calling a function that returns an error and the `if` condition that checks for the error.

lcastellano: Ah, ok you mean there is an extra/unnecessary newline in between the function call and the testing of the condition. Ok, I will fix that. Thanks for the review!!

PR #6631, opened 2017-10-25
Implement Role creation/validation for ops-user

### LINKED ISSUE #6504: Implementation of create/validate/repair of vSphere roles for ops-user
**User Statement:**
When the VCH is created the VIC should check that the appropriate roles for the ops-user are in place in vCenter. If not it should either create or repair the roles.
**Details:**
This issue should cover:
- APIs to create/validate/repair vSphere roles for ops-user
- Unit test
- Manual integration test against an instance of vCenter
**Acceptance Criteria:**
Code and tests implemented.

lcastellano:
Fixes #6504

Implement Role creation/validation/repair for ops-user.

### DIFF
--- a/lib/install/management/rbac.go
+++ b/lib/install/management/rbac.go
@@ -1,0 +1,290 @@
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package management
+
+import (
+	"context"
+
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/lib/config"
+)
+
+const (
+	VCenter = iota
+	Datacenter
+	Cluster
+	DatastoreFolder
+	Datastore
+	VSANDatastore
+	Network
+	Endpoint
+)
+
+type AuthzManager struct {
+	authzManager *object.AuthorizationManager
+	client       *vim25.Client
+	vchConfig    *config.VirtualContainerHostConfigSpec
+	principal    string
+}
+
+type RBACResource struct {
+	Type      int8
+	Propagate bool
+	Role      types.AuthorizationRole
+}
+
+type RBACConfig struct {
+	Resources []RBACResource
+}
+
+type PermissionList []types.Permission
+
+var rolePrefix = "vic-vch-"
+
+var RoleVCenter = types.AuthorizationRole{
+	Name: "vcenter",
+	Privilege: []string{
+		"Datastore.Config",
+	},
+}
+
+var RoleDataCenter = types.AuthorizationRole{
+	Name: "datacenter",
+	Privilege: []string{
+		"Datastore.Config",
+		"Datastore.FileManagement",
+	},
+}
+
+var RoleDataStore = types.AuthorizationRole{
+	Name: "datastore",
+	Privilege: []string{
+		"Datastore.AllocateSpace",
+		"Datastore.Browse",
+		"Datastore.Config",
+		"Datastore.DeleteFile",
+		"Datastore.FileManagement",
+		"Host.Config.SystemManagement",
+	},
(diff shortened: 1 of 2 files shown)


cmd/vic-machine/ops-user/rbac.go
```diff
@@ -0,0 +1,286 @@
...
+
+func (am *AuthzManager) DeleteRoles(ctx context.Context) (int, error) {
+	return am.deleteRoles(ctx, getTargetRoles())
+}
+
+func (am *AuthzManager) RoleList(ctx context.Context) (object.AuthorizationRoleList, error) {
+	return am.getRoleList(ctx)
+}
+
+func (am *AuthzManager) createOrRepairRoles(ctx context.Context, targetRoles []types.AuthorizationRole) (int, error) {
```
hickeng: We may need to be able to report the missing privileges/roles/permissions in the case where the user has _not_ checked the `Grant Permissions` box. You should consider splitting assembly/identification of missing permissions from the modification - maybe that's a follow up PR after function's complete but will likely reduce duplication.

cmd/vic-machine/ops-user/rbac.go
```diff
@@ -0,0 +1,286 @@
...
+	Role      types.AuthorizationRole
+}
+
+type RBACConfig struct {
+	Resources []Resource
+}
+
+type PermissionList []types.Permission
+
+var rolePrefix = "vch-"
```
hickeng: Can we make this `vic-vch-` - while VCH is a very common acronym to us VIC is the name of the product that most people seem to use.

cmd/vic-machine/ops-user/rbac.go
```diff
@@ -0,0 +1,286 @@
...
+	err := mgr.UpdateRole(ctx, fRole.RoleId, fRole.Name, fRole.Privilege)
+
+	return true, err
+}
+
+func getTargetRoles() []types.AuthorizationRole {
+	count := len(RBACConf.Resources)
+	roles := make([]types.AuthorizationRole, 0)
+	dSet := make(map[string]bool)
+	for i := 0; i < count; i++ {
```
hickeng: why not use `for i := range RBACConf.Resources`?

cmd/vic-machine/ops-user/rbac.go
```diff
@@ -0,0 +1,286 @@
...
+		name := rolePrefix + RBACConf.Resources[i].Role.Name
+		// Discard duplicates
+		if _, found := dSet[name]; !found {
```
hickeng: Not sure why you're discarding duplicates given it's an inline struct - should we `panic` instead as it seems like it's programmer error in this circumstance?

cmd/vic-machine/ops-user/rbac_test.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		t.SkipNow()
+	}
+
+	mgr := NewAuthzManager(ctx, session, "ops-user")
+
+	doTestRoles(ctx, t, mgr)
+}
+
+func initRoles(ctx context.Context, t *testing.T, am *AuthzManager,
```
hickeng: ditto line wrapping comment

cmd/vic-machine/ops-user/rbac_test.go
```diff
@@ -0,0 +1,177 @@
...
+		}
+		if role.Name == targetRoleName2 {
+			removePrivilege(&role, "VirtualMachine.Config.AddNewDisk")
+			am.authzManager.UpdateRole(ctx, role.RoleId, role.Name, role.Privilege)
+		}
+	}
+
+	// Test
+	roleCount = 2
+	count, err = am.createOrRepairRoles(ctx, targetRoles)
```
hickeng: There are `testify/assert` (continue but fail test) and `testify/require` (fail immediately and stop) packages that simplify this conditional logic and allow for embedding textual commentary in the test output. It makes the tests much more declarative.

cmd/vic-machine/ops-user/rbac.go
lcastellano: I am thinking of using a different approach for verification of pre-existing permissions. I would log in as the ops-user and use the vmomi API that checks if the user has the permission to do an operation on a resource. I still have to investigate the details of that would work but I doubt there will be much code sharing at the Role level.

cmd/vic-machine/ops-user/rbac.go
lcastellano: In the inline structure there are roles that legitimately appear in multiple entries. For instance, the Datastore Role applies to Cluster, Datastore and SANDatastore. This method returns the set of roles (no duplication).

cmd/vic-machine/ops-user/rbac_test.go
lcastellano: I left one explicit check. It is:

if checkCount && count != roleCount {
   // fail
}

I tried require.True() but the code becomes less readable in my opinion.

PR #6636, opened 2017-10-26
Gather tether.debug

### LINKED ISSUE #6378: CI: collect appliance tether.debug log as well as vmware.log
Builds:
https://ci.vcna.io/vmware/vic/13746
https://ci.vcna.io/vmware/vic/13758

#6375 failed but insufficient logs were collected to diagnose the issue. Some details here: https://github.com/vmware/vic/pull/6375#issuecomment-330000058

We should be collecting the applianceVM `tether.debug` log in the case of being unable to download via vicadmin in addition to vmware.log.

Setting to high as it impacts ability to diagnose CI failures.

gigawhitlocks:
Resolves #6378 .. also I fixed a misaligned indent that I found in the iso build script


tests/resources/VCH-Util.robot
```diff
@@ -201,6 +201,7 @@ Run VIC Machine Command
 Run Secret VIC Machine Delete Command
     [Tags]  secret
     [Arguments]  ${vch-name}
+    Run  govc vm.unregister ${vch-name}
```
sflxn: Why do you need to unregister the VCH here?

gigawhitlocks: was trying to fix another issue but mhagen informed me there's more to it (I misunderstood him in the first place, in thinking this line was all that was needed) so I just removed this

PR #6643, opened 2017-10-27
[skip ci]Fix regression in the NSX test

mhagen-vmware:


matthewavery:
Could you detail the regression  in a brief paragraph? I am not sure what went wrong here.

mhagen-vmware:
This PR https://github.com/vmware/vic/pull/6557 added --container network as default to all VIC machine installs, unfortunately this single test is a bit of a mess with a permanent install that doesn't support that.  So I have just copied our existing install process and removed the container networks again.

PR #6646, opened 2017-10-27
[skip ci] Fix NFS datastore leak in nightly test

mhagen-vmware:


anchal-agrawal:
Please add a bit more detail to the commit message. Something like `Fix NFS datastore leak in nightly test` would suffice.

PR #6656, opened 2017-10-30
WIP Add tcp idle timeout to portlayer nfs client usage

### LINKED ISSUE #6620: Add idle connection deadline to the portlayer nfs client usage
A PR has been merged to our go-nfs-client which allows us to set an idle deadline for stale connections. We need to utilize this so that we can protect our selves from dead connections or even refresh connections that might have become stale, but are still accessible. 

see this pr: https://github.com/vmware/go-nfs-client/pull/4

This should be set right when we declare the use of an NFS client. This should be during the storage portlayer initialization (here)[https://github.com/vmware/vic/blob/5331a29466efc45a4f78970bb3bc5018b5e78d66/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go#L57]. It might not be directly here, but that is the first place that the nfs client is used so it is a good starting place to figure out where to put the configuration for the nfs clients. 

This will make the nfs support in the portlayer more resilient in general.

matthewavery:
Currently this bumps our go-nfs-client to having the latest changes. This includes the changes needed to set the timeout on the underlying connection for the nfs client. This will allow us to reject an nfs volume store if the connection is idle for over 2 minutes saving us error out time against down nfs servers or very flaky connections to an nfs server. This will also affect operations like creating NFS volumes and reading metadata. If the server goes down at any time we should atleast now cancel the operation after the timeout and report a TCP connection timeout. 

fixes #6620


lib/portlayer/storage/nfs/target.go
```diff
@@ -18,13 +18,19 @@ import (
...
 
 	nfsClient "github.com/vmware/go-nfs-client/nfs"
 	"github.com/vmware/go-nfs-client/nfs/rpc"
 
 	"github.com/vmware/vic/pkg/trace"
 )
 
+const (
+	// nfs client volume store mount idle timeout duration, this should protect us from waiting too long on idle NFS servers
+	nfsClientIdleTimeout = 2 * time.Minute
```
hickeng: I suspect that this is too high and we will have other timeouts that kick in before this lapses.
Really this needs to be configurable, so that it can be extended in very busy environments but is low for good responsiveness in most environments.

PR #6657, opened 2017-10-30
Fix small wording issue in logging statement

matthewavery:
Just performs s/changed/change. Pretty small correction that @anchal-agrawal noticed.


matthewavery:
and alas @cgtexmex has stolen my easy one liner :P

PR #6671, opened 2017-11-01
(WIP) Fix tar unpacking with symlinks point out of unpack root

gigawhitlocks:
Small change to unpack & unit test to reproduce throwing an error when a tar is unpacked with a symlink pointing outside of the unpack directory

Marked WIP because I'm sort of expecting @matthewavery or @hickeng to tell me I've overlooked an edge case.


sgairo:
+1 for the unit test

PR #3116, opened 2016-11-11
Delete old VCH

emlin:
Fixes #2719
If vic-machine identified the vm is one VCH instance, will delete the container VMs and VCH instance itself. Even the guestinfo configuration is changed, and some of configuration cannot be identified.
The problem is that if there is custom image store, or volume store configured, but vic-machine cannot read those configuration back, those resources will not be cleaned up.

PR #6679, opened 2017-11-02
[full ci] Prevent vic-machine from deleting non cVMs

### LINKED ISSUE #6603: vic-machine delete can result in non-containerVMs being deleted
The vic-machine delete command is responsible for removing a VCH appliance and it's containerVMs.  The following issues have been identified:

- a VM is determined to be a containerVM if it resides in the appliance's resource pool and does not have extraConfig that would identify it as a VCH appliance.  PoweredOff VMs are deleted by default and if `--force` is supplied to the `vic-machine delete` then both poweredOn and poweredOff VMs are deleted.  This results in the deletion of any non-containerVMs that happen to reside in the appliance resource pool.
- If another VCH appliance is in the target VCH appliance resource pool then any disks attached to it will be detached.  This would cause problems for the non-target VCH appliance when there is a docker pull, commit, cp or diff operation in process.  

To address these issues the following changes need to be made:
- [ ] vic-machine delete will fail if the target resource pool is the default resource pool of the cluster -- an appliance should never be able to be deployed to the default rp of the cluster, but this will be an added measure of safety.
- [x] A VM will only be deleted if it's determined to be a containerVM.  This determination will be made by resource pool residence AND interrogation of the VM extraConfig for containerVM entries.  This will prevent non-containerVMs from being deleted.
- [x] A VCH appliance will only be acted on (detaching disks) if it's the target VCH.  If another VCH resides in the target resource pool it will be untouched.  That determination will be made by the vSphere moRef and the VM extraConfig.

andrewtchin:
Fixes #6603

### DIFF
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -97,7 +97,7 @@ func (d *Dispatcher) isContainerVM(vm *vm.VirtualMachine) (bool, error) {
 	}
 	defer trace.End(trace.Begin(vm.InventoryPath))
 	var cspec executor.ExecutorConfig
-	info, err := vm.FetchExtraConfig(d.ctx)
+	info, err := vm.FetchExtraConfig(d.op)
 	if err != nil {
 		err = errors.Errorf("Failed to fetch guest info of appliance vm: %s", err)
 		return false, err
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -91,6 +91,24 @@ func (d *Dispatcher) isVCH(vm *vm.VirtualMachine) (bool, error) {
 	return false, nil
 }
 
+func (d *Dispatcher) isContainerVM(vm *vm.VirtualMachine) (bool, error) {
+	if vm == nil {
+		return false, errors.New("nil parameter")
+	}
+	defer trace.End(trace.Begin(vm.InventoryPath))
+	var cspec executor.ExecutorConfig
+	info, err := vm.FetchExtraConfig(d.ctx)
+	if err != nil {
+		err = errors.Errorf("Failed to fetch guest info of appliance vm: %s", err)
+		return false, err
+	}
+	extraconfig.Decode(extraconfig.MapSource(info), &cspec)
+	if cspec.Version == nil {
+		return false, nil
+	}
+	return true, nil
+}
+
 func (d *Dispatcher) checkExistence(conf *config.VirtualContainerHostConfigSpec, settings *data.InstallerData) error {
 	defer trace.End(trace.Begin(""))
 
--- a/lib/install/management/delete.go
+++ b/lib/install/management/delete.go
@@ -58,10 +58,28 @@ func (d *Dispatcher) DeleteVCH(conf *config.VirtualContainerHostConfigSpec, cont
 		return nil
 	}
 
-	if err = d.DeleteVCHInstances(vmm, conf, containers); err != nil {
-		// if container delete failed, do not remove anything else
-		d.op.Info("Specify --force to force delete")
-		return err
+	d.parentResourcepool, err = d.getComputeResource(vmm, conf)
+	if err != nil {
+		log.Error(err)
+		if !d.force {
+			d.op.Infof("Specify --force to force delete")
+			return err
+		}
+		// Can't find the RP VCH was created in to delete cVMs, continue anyway
+		d.op.Warnf("No container VMs found, but proceeding with delete of VCH due to --force")
+		err = nil
+	}
+	if d.parentResourcepool != nil {
+		if err = d.DeleteVCHInstances(vmm, conf, containers); err != nil {
+			log.Error(err)
+			if !d.force {
+				// if container delete failed, do not remove anything else
+				d.op.Infof("Specify --force to force delete")
+				return err
+			}
+			d.op.Warnf("Proceeding with delete of VCH due to --force")
+			err = nil
+		}
 	}
 
 	if err = d.deleteImages(conf); err != nil {
@@ -86,6 +104,17 @@ func (d *Dispatcher) DeleteVCH(conf *config.VirtualContainerHostConfigSpec, cont
 		d.op.Debugf("Error deleting appliance VM %s", err)
 		return err
 	}
+
+	defaultrp, err := d.session.Cluster.ResourcePool(d.ctx)
+	if err != nil {
+		return err
+	}
+
+	if d.parentResourcepool != nil && d.parentResourcepool.Reference() == defaultrp.Reference() {
+		log.Warnf("VCH resource pool is cluster default pool - skipping delete")
+		return nil
+	}
+
 	if err = d.destroyResourcePoolIfEmpty(conf); err != nil {
 		d.op.Warnf("VCH resource pool is not removed: %s", err)
 	}
@@ -99,23 +128,10 @@ func (d *Dispatcher) getComputeResource(vmm *vm.VirtualMachine, conf *config.Vir
 	ignoreFailureToFindComputeResources := d.force
 
 	if len(conf.ComputeResources) == 0 {
-		if !ignoreFailureToFindComputeResources {
-			err = errors.Errorf("Cannot find compute resources from configuration")
-			return nil, err
-		}
-		d.op.Warn("Cannot find compute resources from configuration, attempting to delete under parent resource pool")
-		parent, err := vmm.Parent(d.op)
-		if err != nil {
-			return nil, err
-		}
-		if parent == nil {
-			err = errors.Errorf("Cannot find VCH parent resource pool")
-			return nil, err
-		}
-		rpRef = *parent
-	} else {
-		rpRef = conf.ComputeResources[len(conf.ComputeResources)-1]
+		err = errors.Errorf("Cannot find compute resource from configuration")
+		return nil, err
 	}
+	rpRef = conf.ComputeResources[len(conf.ComputeResources)-1]
 
 	ref, err := d.session.Finder.ObjectReference(d.op, rpRef)
 	if err != nil {
@@ -216,11 +232,6 @@ func (d *Dispatcher) DeleteVCHInstances(vmm *vm.VirtualMachine, conf *config.Vir
 
 	var err error
 	var children []*vm.VirtualMachine
-	d.parentResourcepool, err = d.getComputeResource(vmm, conf)
-	if err != nil {
-		return err
-	}
-
 	if children, err = d.parentResourcepool.GetChildrenVMs(d.op, d.session); err != nil {
 		return err
 	}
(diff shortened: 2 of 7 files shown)


lib/install/management/delete.go
```diff
@@ -46,11 +47,36 @@ func (d *Dispatcher) DeleteVCH(conf *config.VirtualContainerHostConfigSpec) erro
...
+		log.Warnf("Proceeding with delete of VCH due to --force")
+		err = nil
+	}
+	defaultrp, err := d.session.Cluster.ResourcePool(d.ctx)
+	if err != nil {
 		return err
 	}
+	if d.parentResourcepool != nil {
+		if d.parentResourcepool.Reference() == defaultrp.Reference() {
+			return fmt.Errorf("Targed VCH is in cluster's default resource pool. Refusing to delete it.")
```
cgtexmex: Why would we not remove if it's in the cluster default?  We won't install there, so it's been moved there...why not remove it?

lib/install/management/delete.go
```diff
@@ -46,11 +47,36 @@ func (d *Dispatcher) DeleteVCH(conf *config.VirtualContainerHostConfigSpec) erro
...
-		log.Infof("Specify --force to force delete")
+	d.parentResourcepool, err = d.getComputeResource(vmm, conf)
+	if err != nil {
+		log.Error(err)
+		if !d.force {
+			log.Infof("Specify --force to force delete")
+			return err
+		}
+		// Assume the cVMs and RP have already been deleted
+		log.Warnf("Proceeding with delete of VCH due to --force")
```
cgtexmex: Will there be later feedback / logging that no containerVMs have been removed?  If not should that be stated here?  i.e. "No container VMs found, but proceeding with delete of VCH due to use of --force"

IMO there's a possibility that the containerVMs exist else where, so they would be "orphaned". Just want to make sure the user understands that no other VMs will be removed...

lib/install/management/delete.go
andrewtchin: I thought this was a requirement, but that makes sense to me. I removed this

lib/install/management/delete.go
andrewtchin: Good point. Added this

lib/install/management/delete.go
cgtexmex: Well..hell...looking at the issue I see why you'd think that was a requirement..I think the first item should have said "vic-machine _create_ will fail if the target rp is the default rp"...was really thinking of adding a new validation rule.  I'm not sure that's really needed...

I think we prevent the install into a cluster default RP by creating an RP on each deploy.  Previously there was an issue that would (in some cases) result in a VCH deployed to the default RP, but I think since 1.2.1 that's fixed...

andrewtchin: yea you're right 👍

lib/install/management/delete.go
```diff
@@ -222,13 +235,29 @@ func (d *Dispatcher) DeleteVCHInstances(vmm *vm.VirtualMachine, conf *config.Vir
 		}
 
 		if ok {
-			// child is vch; detach all attached disks so later removal of images is successful
+			// Do not delete a VCH in the target RP if it is not the target VCH
+			if child.Reference() != vmm.Reference() {
```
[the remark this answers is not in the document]
andrewtchin: This was to address if the user OOB moved a VCH into another VCH's RP. I will open an issue to disable moving a VCH

lib/install/management/delete.go
```diff
@@ -46,10 +46,28 @@ func (d *Dispatcher) DeleteVCH(conf *config.VirtualContainerHostConfigSpec) erro
 		return nil
 	}
 
-	if err = d.DeleteVCHInstances(vmm, conf); err != nil {
-		// if container delete failed, do not remove anything else
-		log.Infof("Specify --force to force delete")
-		return err
+	d.parentResourcepool, err = d.getComputeResource(vmm, conf)
```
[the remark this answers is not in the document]
andrewtchin: This and the next change below was to allow a VCH that is moved to another RP to be deleted if identified by moid (`--id`). In this case parentResourcepool would previously have returned an error causing this to bail and not delete the VCH even though it is found by its moid

lib/install/management/appliance.go
```diff
@@ -93,6 +94,25 @@ func (d *Dispatcher) isVCH(vm *vm.VirtualMachine) (bool, error) {
...
+		return false, errors.New("nil parameter")
+	}
+	defer trace.End(trace.Begin(vm.InventoryPath))
+	var cspec executor.ExecutorConfig
+	info, err := vm.FetchExtraConfig(d.ctx)
+	if err != nil {
+		err = errors.Errorf("Failed to fetch guest info of appliance vm: %s", err)
+		return false, err
+	}
+	extraconfig.Decode(extraconfig.MapSource(info), &cspec)
```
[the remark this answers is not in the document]
andrewtchin: fixed

lib/install/management/delete.go
```diff
@@ -46,10 +46,28 @@ func (d *Dispatcher) DeleteVCH(conf *config.VirtualContainerHostConfigSpec) erro
...
+		log.Error(err)
+		if !d.force {
+			log.Infof("Specify --force to force delete")
+			return err
+		}
+		// Assume the cVMs and RP have already been deleted
+		log.Warnf("No container VMs found, but proceeding with delete of VCH due to --force")
+		err = nil
+	}
+	if d.parentResourcepool != nil {
```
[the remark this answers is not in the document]
andrewtchin: this is required since I use d.parentResourcepool above and that could be nil if the parent RP was deleted OOB

lib/install/management/delete.go
```diff
@@ -46,10 +46,28 @@ func (d *Dispatcher) DeleteVCH(conf *config.VirtualContainerHostConfigSpec) erro
...
-		return err
+	d.parentResourcepool, err = d.getComputeResource(vmm, conf)
+	if err != nil {
+		log.Error(err)
+		if !d.force {
+			log.Infof("Specify --force to force delete")
+			return err
+		}
+		// Assume the cVMs and RP have already been deleted
+		log.Warnf("No container VMs found, but proceeding with delete of VCH due to --force")
```
hickeng: I do not see a path by which both `err` and `d.parentResourcepool` are non-nil so this logic will simply skip the attempt to delete the VCH when we hit the `d.parentResourcePool != nil` condition.

The text about containerVMs is also misleading given we've not looked for containerVMs at this point.

lib/install/management/delete.go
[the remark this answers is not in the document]
hickeng: `d.parentResourcePool` can have the following values:
a. moid of originally targeted resource pool (endpointVM may not still be in that pool)
b. current parent pool of the endpointVM

In either case we now delete children of that pool that can be identified as cVMs, then attempt to destroy that pool if it's empty.

I'm unsure what happens if the endpointVM is the last VM in the cluster and the pool returned is the root pool for the cluster. I do not know what happens if you call destroy on the root resource pool of a cluster, but to be safe I'd rather we never do so.

tests/test-cases/Group6-VIC-Machine/6-03-Delete.md
```diff
@@ -36,3 +36,13 @@ This test requires that a vSphere server is running and available
...
+
+## Delete VCH with non-cVM in same RP
+1. Create VCH though vic-machine create
+2. Create VM in the VCH's resource pool
+3. Delete VCH
+4. Check vSphere through govc to make sure the VM was not deleted
+5. Delete the VM and RP to cleanup
+
+### Expected Outcome:
+1. All steps should succeed
```
hickeng: We need another two tests:
1. endpointVM is moved out of it's pool, pool is left alone
2. endpointVM is moved into the root of the cluster and the original pool is deleted.

andrewtchin: Just to be clear, for 1) the endpointVM should be successfully deleted and the RP and its children should not be deleted
and for 2) the original pool should be deleted OOB after endpointVM is moved to the root pool then vic-machine delete should successfully delete the endpointVM?
Thanks

hickeng: 1) correct
2) correct - and other VMs in the root pool should be left alone (which is what the current test asserts but worth being explicit)

andrewtchin:
Added the integration tests, working through them now.
@hickeng 
- For "I think we should remove the failover path for the actual parent pool - this code was added in #3116 but I'm unsure what specific scenario was being addressed with this path." what failover path is that that i need to remove?
- I have confirmed that on create we always create a new RP so we can't have a VCH in the root RP. Given that do we still need this? "a sanity check to return nil if the ComputeResource[0] is the root pool."
- Does this "I do not know what happens if you call destroy on the root resource pool of a cluster, but to be safe I'd rather we never do so." mean that after we have deleted a VCH that was moved into the root RP we do not want to call the pool destroy?

hickeng:
* This is the failover path I'm talking about - https://github.com/vmware/vic/pull/3116/files#diff-786c272aaac433e0eaae7299e08b52b7R91
* If we remove the failover path that switches to the actual parent and not the recorded parent then no, we do not need the sanity check.
* Correct - we do not want to even try to destroy the root pool

There should be a follow up issue to look at disabling _Move_ for the individual VMs.

lib/install/management/delete.go
```diff
@@ -46,10 +46,28 @@ func (d *Dispatcher) DeleteVCH(conf *config.VirtualContainerHostConfigSpec) erro
...
+	d.parentResourcepool, err = d.getComputeResource(vmm, conf)
+	if err != nil {
+		log.Error(err)
+		if !d.force {
+			log.Infof("Specify --force to force delete")
+			return err
+		}
+		// Can't find the RP VCH was created in to delete cVMs, continue anyway
+		log.Warnf("No container VMs found, but proceeding with delete of VCH due to --force")
+		err = nil
```
hickeng: getComputeResource never returns non-nil for both rp and err. If you set `err = nil` here then we will still skip over the delete of the containerVMs (not sure why the method is called `DeleteVCHInstances`).

I don't know what that will mean for deleting images if they are still attached to cVMs - I suspect it will fail with errors about locked vmdks if any of the cVMs are powered on. I don't see that that will cause any other "interesting" behaviours like deleting non-VCH related things, but please sanity check me if you haven't looked at this path.

tests/test-cases/Group6-VIC-Machine/6-03-Delete.robot
```diff
@@ -95,3 +104,128 @@ Attach Disks and Delete VCH
...
+    Log  ${output}
+    Should Contain  ${output}  ${dummyvm}
+
+    # Move VCH to root RP
+    ${rc}  ${output}=  Run And Return Rc And Output  govc vm.migrate -pool %{TEST_RESOURCE} %{VCH-NAME}
+    Should Be Equal As Integers  ${rc}  0
+
+    # Delete with force
+    ${moid}=  Get VM Moid  %{VCH-NAME}
+    ${ret}  ${output}=  Run And Return Rc And Output  bin/vic-machine-linux delete --target %{TEST_URL} --user %{TEST_USERNAME} --password=%{TEST_PASSWORD} --compute-resource=%{TEST_RESOURCE} --id ${moid} --force
```
hickeng: As a note, if you're specifying `--id` then I don't think you need `--compute-resource`. I frequently use `--id` to override which VCH I'm deleting when working with nimbus or vmc.

tests/test-cases/Group6-VIC-Machine/6-03-Delete.robot
```diff
@@ -95,3 +104,128 @@ Attach Disks and Delete VCH
...
+
+    ${rc}  ${output}=  Run And Return Rc And Output  govc pool.destroy "%{TEST_RESOURCE}/%{VCH-NAME}"
+    Log  ${output}
+    Should Be Equal As Integers  ${rc}  0
+
+    Run Keyword And Ignore Error  Cleanup VCH Bridge Network  %{VCH-NAME}
+
+
+Delete VCH moved from its RP
+    Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Pass Execution  Test skipped on ESX due to unable to move into RP
```
hickeng: ESX does have resource pools and we should be able to move VMs between them. It's possible that govc is missing exposure of the appropriate command.
(I'm not sure how to view resource pools in the H5 client however).

andrewtchin:
Replacement PR in #6816 to get the build to run

PR #6680, opened 2017-11-02
[specific ci=1-11-Docker-RM] Add support for docker rm -v

### LINKED ISSUE #6262: docker rm -v does not delete volumes
```
vagrant@devbox:~/vicsmb/src/github.com/vmware/vic$ docker volume ls;docker create -v /mnt/test --name=test alpine; docker rm -v test;docker volume ls
DRIVER              VOLUME NAME
2446c3b950ff681300ef963b54518d5dacbd6aaa51e67743387a60d9e0f95667
test
DRIVER              VOLUME NAME
vsphere             23c8987e-9612-11e7-ac2d-000c2954bf09
vagrant@devbox:~/vicsmb/src/github.com/vmware/vic$
```

anchal-agrawal:
This commit adds support in the Docker personality for the 'docker rm
-v' command, which removes the container and also any anonymous
volumes joined to it. If an anonymous volume is in use by another
container, it is not removed. Named volumes (specified by name in the
create/run command) are not deleted.

Fixes #6262

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -774,15 +774,9 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 		return NotFoundError(name)
 	}
 	id := vc.ContainerID
-	// Get the portlayer Client API
-	client := c.containerProxy.Client()
-
-	// TODO: Pass this RemoveVolume flag to somewhere
-	_ = &config.RemoveVolume
 
 	// Use the force and stop the container first
 	secs := 0
-
 	if config.ForceRemove {
 		c.containerProxy.Stop(vc, name, &secs, true)
 	} else {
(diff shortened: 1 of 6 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -775,26 +769,8 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
...
-			return derr.NewRequestConflictError(fmt.Errorf("You cannot remove a running container. Stop the container before attempting removal or use -f"))
-		case *containers.ContainerRemoveInternalServerError:
-			if err.Payload == nil || err.Payload.Message == "" {
-				return InternalServerError(err.Error())
-			}
-			return InternalServerError(err.Payload.Message)
-		default:
-			return InternalServerError(err.Error())
-		}
+	if err := c.containerProxy.Remove(vc, config.RemoveVolume); err != nil {
```
sflxn: thanks for doing this.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -104,6 +104,7 @@ type VicContainerProxy interface {
 	Signal(vc *viccontainer.VicContainer, sig uint64) error
 	Resize(id string, height, width int32) error
 	Rename(vc *viccontainer.VicContainer, newName string) error
+	Remove(vc *viccontainer.VicContainer, removeVols bool) error
```
matthewavery: Since we are using the docker types here in the container proxy we should use `types.ContainerRmConfig` instead of creating `removeVols`. Since it contains all of the options from a docker personality perspective for a `docker rm` situation.

anchal-agrawal: Fixed - thanks 👍

matthewavery:
Looks great!

sflxn:
lgtm.  Had to work through your tests.

Issue #6492, opened 2017-10-03
Setup correct RBAC roles/permissions for ops-user in VCH (--ops-user)


lcastellano:
**User Statement:**
The VCH must perform operations in VCenter (e.g. create/delete users). In VIC 1.2 the creator of the VCH has two options:

1. Supply the VCenter administrator credentials,
2. Create and manually configure another VCenter user with minimum roles and permission.

While the first alternative wins on usability it does not provide the adequate barriers among VCHs. In addition, it suffers from security weaknesses since each VCH can act as the full VCenter administration. The second approach requires a lot of tedious and error prone manual configuration in VSphere, but it provides the needed security and isolation.
The goal of this epic is to automate the creation/configuration of the VCenter user (ops-user) described by (2).

**Details:**
Automation of creation and configuration of the VIC user should be performed by vic-machine (or vic-machine-service) and implemented as a set of APIs. The are three main tasks:

1. Creation of ops-user
2. Creation of the roles that ops-user will have, this has to be done only once in VCenter,
3. Assignment of permissions for the different VCenter resources managed by the VCH. Some of the resources are created by vic-machine and therefore permission assignment must be done at the end of VCH creation.
4. Validation of the roles and permission given and existing VCH and an exiting ops-user.

Since creation of the user is not currently supported in go-vmomi, we believe that manual creation of the user + automated Role/Permission assignment and Validation of the ops-user should be a valuable intermediate goal.

**Acceptance Criteria:**
Design and implementation of the feature described above. The first implementation will skip the creation of the ops-User.

lcastellano:
This is from issue: https://github.com/vmware/vic/issues/1689

Dumping the RBAC config from a working VCH deployed on vCenter 6.5:

```
vagrant@devbox:~$ for i in "VCH - vcenter" "VCH - datacenter" "VCH - datastore" "VCH - network" "VCH - endpoint"; do  echo ======= Role $i ============;echo "******" Configuration; govc role.ls "$i"; echo;echo "******" Usage; govc role.usage "$i"; echo;echo;done  | grep -v '^System\.'
======= Role VCH - vcenter ============
****** Configuration
Datastore.Config

****** Usage
Role           Entity  Principal          Propagate
VCH - vcenter  /       VSPHERE.LOCAL\ops  No


======= Role VCH - datacenter ============
****** Configuration
Datastore.Config
Datastore.FileManagement

****** Usage
Role              Entity  Principal          Propagate
[...]

PR #6682, opened 2017-11-03
Implementation of permission settings for opt-user roles.

lcastellano:
Implements the code for issue #6492. Also adds test environment for govmomi simulator.


lib/install/management/rbac_test.go
```diff
@@ -62,36 +72,164 @@ func TestRolesVCenter(t *testing.T) {
...
+	resourcePermissions, err := am.SetupRolesAndPermissions(ctx)
+	require.NoError(t, err)
+	defer cleanup(ctx, t, am, true)
+	require.True(t, len(am.rbacConfig.Resources) >= len(resourcePermissions))
+
+	verifyResourcePermissions(ctx, t, am, resourcePermissions, am.rbacConfig.Resources)
+}
+
+func verifyResourcePermissions(ctx context.Context, t *testing.T, am *AuthzManager, retPerms []RBACResourcePermission, configResources []RBACResource) {
+	// Verify resource permissions
```
anchal-agrawal: Redundant comment - this is just the function name. We should either expand on the comment here, or include more details in a function comment above the function.

lib/install/management/rbac.go
```diff
@@ -158,29 +178,243 @@ var RBACConf = RBACConfig{
...
+		return nil, errors.Errorf("Ops-User: AuthzManager, Unable to build Bridged Network MoRef: %s", bridgeNet.ID)
+	}
+	resourceDescs = append(resourceDescs, ResourceDesc{Network, *bridgeNetRef})
+
+	// Get VM MoRef
+	vmRef := &types.ManagedObjectReference{}
+	vmRef.FromString(am.configSpec.ID)
+	resourceDescs = append(resourceDescs, ResourceDesc{Endpoint, *vmRef})
+
+	resourcePermissions := make([]RBACResourcePermission, 0, 16)
```
anchal-agrawal: Should 16 be a constant? Not sure why this is 16.

lib/install/management/rbac.go
```diff
@@ -158,29 +178,243 @@ var RBACConf = RBACConfig{
...
+	return am.deleteRoles(ctx)
 }
 
 func (am *AuthzManager) RoleList(ctx context.Context) (object.AuthorizationRoleList, error) {
 	return am.getRoleList(ctx)
 }
 
-func (am *AuthzManager) createOrRepairRoles(ctx context.Context, targetRoles []types.AuthorizationRole) (int, error) {
+func (am *AuthzManager) SetupRolesAndPermissions(ctx context.Context) ([]RBACResourcePermission, error) {
+	_, err := am.CreateRoles(ctx)
```
anchal-agrawal: This can be collapsed to:

```go
if _, err := am.CreateRoles(ctx); err != nil {
    return nil, err
}
```

lib/install/management/rbac.go
```diff
@@ -158,29 +178,243 @@ var RBACConf = RBACConfig{
...
+	if err != nil {
+		return nil, err
+	}
+	perms, err := am.SetupPermissions(ctx)
```
anchal-agrawal: Lines 223-224 can be collapsed to `return am.SetupPermissions(ctx)`

lib/install/management/rbac.go
```diff
@@ -17,10 +17,19 @@ package management
 import (
 	"context"
 
+	"fmt"
```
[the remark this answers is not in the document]
lcastellano: Sorry Anchal, but I can't figure out what you want me to do there. I got lost into that thread. If there are guidelines they should be in a document not in a thread. Can you be more specific about what needs to be done? Thanks.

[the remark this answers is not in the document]
anchal-agrawal: Sure. https://github.com/golang/go/wiki/CodeReviewComments#imports

lcastellano:
This code is not called from anywhere yet. In my next check-in I am planning to add an option to vic-machine create and call it from there. That PR should will contain the integration test.

lib/install/management/rbac.go
```diff
@@ -16,11 +16,20 @@ package management
 
 import (
 	"context"
+	"fmt"
 
+	"net/url"
```
anchal-agrawal: `"net/url"` belongs in the standard pkg group above - therefore, there should be no newline between `"fmt"` and this import.

lib/install/management/rbac.go
```diff
@@ -158,29 +178,243 @@ var RBACConf = RBACConfig{
...
+	resourceDescs := make([]ResourceDesc, 0, 32)
+
+	// Get a reference to the top object
+	finder := find.NewFinder(am.client, false)
+
+	root, err := finder.Folder(ctx, "/")
+	if err != nil {
+		return nil, errors.Errorf("Ops-User: AuthzManager, Unable to find top object: %s", err.Error())
+	}
+
```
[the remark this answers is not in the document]
anchal-agrawal: @lcastellano please address this - thanks!

lib/install/management/rbac.go
lcastellano: Thanks. Done.

PR #6690, opened 2017-11-03
Fix version file in VCH log bundle

### LINKED ISSUE #6688: VERSION file in log bundle doesn't report the full version
**VIC version:**

master / cd24fcd

**Steps to reproduce:**

* Download the log bundle from VCH admin
* Open the `VERSION` file

**Actual behavior:**

The `VERSION` file contains only the git tag. For a master build off cd24fcd, the file contains just `v1.2.0-rc1`.

**Expected behavior:**

The `VERSION` file should also include the git commit and the build number if possible.

**Logs:**

[VERSION.txt](https://github.com/vmware/vic/files/1442214/VERSION.txt)

Marked as medium priority as this hurts debugging.

anchal-agrawal:
This commit fixes the contents of the VERSION file in the log bundle
downloaded from VCH admin. Previously, it contained only the git tag.
With this change, it contains the short version which is of the format
"gittag-buildnumber-commit".

Fixes #6688

### DIFF
--- a/cmd/vicadmin/vicadm.go
+++ b/cmd/vicadmin/vicadm.go
@@ -205,7 +205,7 @@ type versionReader string
 
 func (path versionReader) open() (entry, error) {
 	defer trace.End(trace.Begin(string(path)))
-	return newBytesEntry(string(path), []byte(version.Version)), nil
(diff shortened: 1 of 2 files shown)


sgairo:
+1  for the extra test

PR #6692, opened 2017-11-03
Add operation id to vSphere calls

### LINKED ISSUE #2739: Add opid to log messages from our calls to vsphere

cgtexmex:
Increases use of `trace.Operation` and passes the operation id to vSphere resulting in vic
operation observability within vSphere logging.

Fixes #2739

### DIFF
--- a/lib/portlayer/event/manager.go
+++ b/lib/portlayer/event/manager.go
@@ -18,11 +18,11 @@ import (
 	"fmt"
 	"sync"
 
-	log "github.com/Sirupsen/logrus"
-
 	"github.com/vmware/vic/lib/portlayer/event/collector"
 	"github.com/vmware/vic/lib/portlayer/event/events"
 	"github.com/vmware/vic/pkg/trace"
+
+	log "github.com/Sirupsen/logrus"
 )
 
 type Manager struct {
@@ -69,10 +69,10 @@ func NewEventManager(collectors ...collector.Collector) *Manager {
 			subs := mgr.subs.subscribers[e.Topic()]
 			mgr.subs.mu.RUnlock()
 
-			log.Debugf("Found %d subscribers to id: %d - %s: %s", len(subs), e.EventID(), e.Topic(), e.Message())
+			log.Debugf("Found %d subscribers to %s: %s - %s", len(subs), e.EventID(), e.Topic(), e.Message())
 
 			for sub, s := range subs {
-				log.Debugf("Event manager calling back to %s for %d - %s: %s", sub, e.EventID(), e.Topic(), e.Message())
+				log.Debugf("Event manager calling back to %s for Event(%s): %s", sub, e.EventID(), e.Topic())
 				s.onEvent(e)
 			}
 		}
--- a/lib/portlayer/exec/base.go
+++ b/lib/portlayer/exec/base.go
@@ -33,8 +33,6 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/extraconfig/vmomi"
 	"github.com/vmware/vic/pkg/vsphere/tasks"
 	"github.com/vmware/vic/pkg/vsphere/vm"
-
-	log "github.com/Sirupsen/logrus"
 )
 
 // NotYetExistError is returned when a call that requires a VM exist is made
@@ -104,10 +102,10 @@ func (c *containerBase) VMReference() types.ManagedObjectReference {
 }
 
 // unlocked refresh of container state
-func (c *containerBase) refresh(ctx context.Context) error {
-	base, err := c.updates(ctx)
+func (c *containerBase) refresh(op trace.Operation) error {
+	base, err := c.updates(op)
 	if err != nil {
-		log.Errorf("Update: unable to update container %s: %s", c, err)
+		op.Errorf("Update: unable to update container %s: %s", c, err)
 		return err
 	}
 
@@ -117,8 +115,8 @@ func (c *containerBase) refresh(ctx context.Context) error {
 }
 
 // updates acquires updates from the infrastructure without holding a lock
-func (c *containerBase) updates(ctx context.Context) (*containerBase, error) {
-	defer trace.End(trace.Begin(c.ExecConfig.ID))
+func (c *containerBase) updates(op trace.Operation) (*containerBase, error) {
+	defer trace.End(trace.Begin(c.ExecConfig.ID, op))
 
 	var o mo.VirtualMachine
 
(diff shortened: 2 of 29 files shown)


lib/portlayer/logging/logging.go
```diff
@@ -54,19 +54,20 @@ func eventCallback(ctx context.Context, ie events.Event) {
 	switch ie.String() {
 	case events.ContainerMigrated,
 		events.ContainerMigratedByDrs:
-		log.Debugf("Received %s event", ie)
+		op := trace.FromContext(ctx, "LoggingEvent")
```
hickeng: Change signature instead of `FromContext`?

pkg/trace/operation.go
```diff
@@ -201,20 +203,28 @@ func FromOperation(parent Operation, format string, args ...interface{}) Operati
...
+// FromContext will return an Operation
+//
+// The Operation returned will be one of the following:
+//   The operation in the context value
+//   The operation passed as the context param
+//   A new operation
+func FromContext(ctx context.Context, message string) Operation {
+
+	// do we have a context w/the op added as a value
+	if op, ok := ctx.Value(OpTraceKey).(Operation); ok {
```
hickeng: can we have an extra sanity check in the condition `op != nil`. Stupid things can happen and while this really should result in a panic we also need to continue functioning.

pkg/trace/trace.go
```diff
@@ -62,30 +67,50 @@ func (t *Message) delta() time.Duration {
...
 	pc, _, line, ok := runtime.Caller(skip)
 	if !ok {
 		return nil
 	}
 
-	name := runtime.FuncForPC(pc).Name()
+	// lets only return the func name from the repo (vic)
+	// down - i.e. vic/lib/etc vs. github.com/vmware/vic/lib/etc
+	// if github.com/vmware doesn't match then the original is returned
+	name := strings.Replace(runtime.FuncForPC(pc).Name(), "github.com/vmware/", "", -1)
```
hickeng: There is a `strings.TrimPrefix` function which I think is a cleaner fit.

lib/portlayer/exec/base.go
```diff
@@ -304,16 +302,16 @@ func (c *containerBase) shutdown(ctx context.Context, waitTime *int32) error {
...
-		timeout, cancel := context.WithTimeout(ctx, wait)
+		timeout, cancel := trace.WithTimeout(&op, wait, "shutdown")
 		defer cancel()
 
 		err := c.startGuestProgram(timeout, "kill", sig)
 		if err != nil {
 			// Just warn and proceed to waiting for power state per issue https://github.com/vmware/vic/issues/5803
 			// Description above in function kill()
-			log.Warnf("%s: %s", msg, err)
+			timeout.Warnf("%s: %s", msg, err)
```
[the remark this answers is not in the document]
cgtexmex: doesn't really matter...it will retain the opID since it was created via `trace.WithTimeout` -- but possibly more clear to use op...so will change to op

lib/portlayer/exec/container.go
```diff
@@ -315,9 +316,10 @@ func (c *Container) WaitForState(s State) <-chan struct{} {
 func (c *Container) NewHandle(ctx context.Context) *Handle {
```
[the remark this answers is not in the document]
cgtexmex: You have the freedom to pass an operation and if you do it will use that operation...otherwise create a new one.  I didn't change the signature in an effort to limit the changes..

lib/portlayer/exec/handle.go
```diff
@@ -202,6 +202,9 @@ func (h *Handle) String() string {
 }
 
 func (h *Handle) Commit(ctx context.Context, sess *session.Session, waitTime *int32) error {
```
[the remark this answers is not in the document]
cgtexmex: updated...

pkg/trace/operation.go
cgtexmex: `op` could never == nil in this case.  I guess it could be the zero value of the operation struct, so I've added a check to ensure there's an `op.id` -- so if we have a zero value we will return a new operation otherwise an operation based on the one passed via `context.Value`

lib/portlayer/exec/container.go
```diff
@@ -516,21 +518,28 @@ func (c *Container) LogReader(ctx context.Context, tail int, follow bool, since
...
+		hosts, _ := c.vm.Datastore.AttachedHosts(op)
 		if len(hosts) > 1 {
 			// In this case, we need download from the VM host as it owns the file lock
 			// #nosec: Errors unhandled.
-			h, _ := c.vm.HostSystem(ctx)
+			h, _ := c.vm.HostSystem(op)
 			if h != nil {
-				ctx = c.vm.Datastore.HostContext(ctx, h)
+				// get a context that embeds the host as a value
+				ctx := c.vm.Datastore.HostContext(op, h)
```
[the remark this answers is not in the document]
cgtexmex: you're correct -- the values in context are preserved.  There was an issue with `FromContext` that was preventing the reuse of the operation -- that has been fixed and will be in the next commit.

matthewavery:
I am really excited about this, it looks great!

lib/portlayer/event/manager.go
```diff
@@ -18,11 +18,11 @@ import (
 	"fmt"
 	"sync"
 
-	log "github.com/Sirupsen/logrus"
-
 	"github.com/vmware/vic/lib/portlayer/event/collector"
 	"github.com/vmware/vic/lib/portlayer/event/events"
 	"github.com/vmware/vic/pkg/trace"
+
+	log "github.com/Sirupsen/logrus"
```
emlin: just curious, why this import is moved?

pkg/trace/operation.go
```diff
@@ -71,9 +77,9 @@ func newOperation(ctx context.Context, id string, skip int, msg string) Operatio
 
 // Creates a header string to be printed.
 func (o *Operation) header() string {
-	if Logger.Level >= logrus.DebugLevel {
-		return fmt.Sprintf("op=%s (delta:%s)", o.id, o.t[0].delta())
-	}
+	// if Logger.Level >= logrus.DebugLevel {
```
emlin: is this still needed?

lib/portlayer/event/manager.go
cgtexmex: during an iteration of the development it wasn't needed, but had to revert back to using it...apparently I changed the location...

pkg/trace/operation.go
cgtexmex: nope - thanks!

PR #6694, opened 2017-11-04
Implement API handler for VCH deletion [specific ci=Group23-VIC-Machine-Service]

### LINKED ISSUE #6137: Implement an API handler for deletion of a VCH

### LINKED ISSUE #6139: Implement end-to-end tests for the API handler for deletion of a VCH

zjs:
Introduce a pair of handlers for deleting VCHs within a vSphere target or datacenter. By default, deletion includes the VCH and powered off containers. Deletion may optionally include powered on containers and/or volume stores as well.

If any containers remain, the VCH is not deleted. If the VCH is not deleted, the response includes a non-2xx status code.

Define a suite of end-to-end tests which verify the intended deletion behavior. End-to-end tests do not attempt to verify the behavior of concurrent operations.

Fixes #6137
Fixes #6139

### DIFF
--- a/lib/apiservers/service/restapi/handlers/vch_delete.go
+++ b/lib/apiservers/service/restapi/handlers/vch_delete.go
@@ -1,0 +1,127 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package handlers
+
+import (
+	"fmt"
+	"net/http"
+
+	"github.com/go-openapi/runtime/middleware"
+
+	"github.com/vmware/vic/lib/apiservers/service/models"
+	"github.com/vmware/vic/lib/apiservers/service/restapi/handlers/util"
+	"github.com/vmware/vic/lib/apiservers/service/restapi/operations"
+	"github.com/vmware/vic/lib/install/data"
+	"github.com/vmware/vic/lib/install/management"
+	"github.com/vmware/vic/lib/install/validate"
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/version"
+)
+
+// VCHDelete is the handler for deleting a VCH
+type VCHDelete struct {
+}
+
+// VCHDatacenterDelete is the handler for deleting a VCH within a Datacenter
+type VCHDatacenterDelete struct {
+}
+
+func (h *VCHDelete) Handle(params operations.DeleteTargetTargetVchVchIDParams, principal interface{}) middleware.Responder {
+	op := trace.NewOperation(params.HTTPRequest.Context(), "VCHDelete: %s", params.VchID)
+
+	b := buildDataParams{
+		target:     params.Target,
+		thumbprint: params.Thumbprint,
+	}
+
+	d, err := buildData(op, b, principal)
+	if err != nil {
+		return operations.NewDeleteTargetTargetVchVchIDDefault(util.StatusCode(err)).WithPayload(&models.Error{Message: err.Error()})
+	}
+
+	d.ID = params.VchID
+
+	err = deleteVCH(op, d, params.DeletionSpecification)
+	if err != nil {
+		return operations.NewDeleteTargetTargetVchVchIDDefault(util.StatusCode(err)).WithPayload(&models.Error{Message: err.Error()})
+	}
+
+	return operations.NewDeleteTargetTargetVchVchIDAccepted()
+}
+
+func (h *VCHDatacenterDelete) Handle(params operations.DeleteTargetTargetDatacenterDatacenterVchVchIDParams, principal interface{}) middleware.Responder {
+	op := trace.NewOperation(params.HTTPRequest.Context(), "VCHDelete: %s", params.VchID)
+
+	b := buildDataParams{
+		target:     params.Target,
+		thumbprint: params.Thumbprint,
+		datacenter: &params.Datacenter,
+	}
+
+	d, err := buildData(op, b, principal)
+	if err != nil {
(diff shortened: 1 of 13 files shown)


zjs:
CI was successful prior to rebase: https://ci.vcna.io/vmware/vic/14524

lib/apiservers/service/restapi/handlers/vch_delete.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	d, err := buildData(op, b, principal)
+	if err != nil {
+		return operations.NewDeleteTargetTargetVchVchIDDefault(util.StatusCode(err)).WithPayload(&models.Error{Message: err.Error()})
+	}
+
+	d.ID = params.VchID
+
+	err = deleteVCH(op, d, params.DeletionSpecification)
```
anchal-agrawal: This can be collapsed to:

```go
if err = deleteVCH(op, d, params.DeletionSpecification); err != nil {
    ...
}
```

lib/apiservers/service/restapi/handlers/vch_delete.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	d, err := buildData(op, b, principal)
+	if err != nil {
+		return operations.NewDeleteTargetTargetDatacenterDatacenterVchVchIDDefault(util.StatusCode(err)).WithPayload(&models.Error{Message: err.Error()})
+	}
+
+	d.ID = params.VchID
+
+	err = deleteVCH(op, d, params.DeletionSpecification)
```
anchal-agrawal: This can be collapsed to:

```go
if err = deleteVCH(op, d, params.DeletionSpecification); err != nil {
    ...
}
```

lib/apiservers/service/restapi/handlers/vch_delete.go
```diff
@@ -0,0 +1,127 @@
...
+		specification = &models.DeletionSpecification{}
+	}
+
+	executor := management.NewDispatcher(validator.Context, validator.Session, nil, false)
+	vch, err := executor.NewVCHFromID(d.ID)
+	if err != nil {
+		return util.NewError(http.StatusNotFound, fmt.Sprintf("Failed to find VCH: %s", err))
+	}
+
+	err = validate.SetDataFromVM(validator.Context, validator.Session.Finder, vch, d)
```
anchal-agrawal: This error block can be collapsed too.

lib/apiservers/service/restapi/handlers/vch_delete.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+
+	// compare vch version and vic-machine version
+	installerBuild := version.GetBuild()
+	if vchConfig.Version == nil || !installerBuild.Equal(vchConfig.Version) {
+		err = fmt.Errorf("VCH version %q is different than API version %s", vchConfig.Version.ShortVersion(), installerBuild.ShortVersion())
+		return util.WrapError(http.StatusBadRequest, err)
+	}
+
+	err = executor.DeleteVCH(vchConfig, specification.Containers, specification.VolumeStores)
```
anchal-agrawal: This error block can be collapsed too.

lib/install/management/delete.go
```diff
@@ -188,9 +190,12 @@ func (d *Dispatcher) detachAttachedDisks(v *vm.VirtualMachine) error {
 	return err
 }
 
-func (d *Dispatcher) DeleteVCHInstances(vmm *vm.VirtualMachine, conf *config.VirtualContainerHostConfigSpec) error {
+func (d *Dispatcher) DeleteVCHInstances(vmm *vm.VirtualMachine, conf *config.VirtualContainerHostConfigSpec, containers *string) error {
 	defer trace.End(trace.Begin(conf.Name))
 
+	deletePoweredOnContainers := d.force || (containers != nil && *containers == "all")
```
anchal-agrawal: Where is `"all"` coming from? We should consider making it a constant since we're using it in two places here and it may be referenced in other places in the future.

lib/apiservers/service/restapi/handlers/vch_delete.go
zjs: It can be collapsed, but I don't think doing so improves readability.

With the current structure, `err` is not as tightly scoped as it could be, but the overall flow is clear. By collapsing the assignments into initialization statements of the conditional, I think it becomes harder to read the overall flow of the function.

I think it's best to limit our use of optional initialization statements to calls where the sole purpose of the statement is the conditional. Typically, these are calls which are free of side-effects.

This allows someone skimming the code to easily see what's happening. In the following example, it's clear that the overall flow is to check something, do something (and handle any errors), validate something, call a method, and then do something else (and handle any errors). 
```
if total := a+b+c; total <= 0 {
   ...
}

d, err := doThingOne(a, b, c)
if err != nil {
   ...
}

if err = validateResult(d); err != nil {
   ...
}

d.callThisMethod()

err = doThingThree(e)
if err != nil {
    ...
}
```

We can re-write this, collapsing more aggressively, and have the same behavior.
```
if total := a+b+c; total <= 0 {
   ...
}

d, err := doThingOne(a, b, c)
if err != nil {
   ...
}

if err = validateResult(d); err != nil {
   ...
}

d.callThisMethod()

if err = doThingThree(e); err != nil {
    ...
}
```
However, this structure emphasizes the conditional nature of `doThingThree` (i.e., that it may result in an error) instead of emphasizing that we're doing something (and then handling an error).

The difference is subtle, but I think it improves readability.

Are there advantages to the collapsed version I may be overlooking?

lib/install/management/delete.go
zjs: Good catch! `"all"` is defined in Swagger, and is captured in a generated constant.

I could use that here, but that seems conceptually wrong (since this code shouldn't "know about" swagger).

I'll look for a better solution.

zjs: What do you think of the approach in b5b48d08? I don't love it, but it avoids "magic constants", provides type safety, avoids coupling the management layer to swagger, and is easily extended if we introduce more granular deletion specifications.

lib/apiservers/service/restapi/handlers/vch_delete.go
anchal-agrawal: Sounds good! I brought this up purely for idiomatic purposes - if variables assigned in an `if` block aren't used outside of it and an error is being checked in the said block, we can perform the assignment and the check on the same line. However, if it makes more sense to separate them for readability, we can do that.

lib/install/management/delete.go
anchal-agrawal: Looks great - thanks!

Issue #3897, opened 2017-02-10
CI system sometimes pull older code than what's in the PR


sflxn:
As a VIC engineer, I should expect the code pulled in a CI build matches the PR that kicked off the build.

Acceptance criteria:  CI no longer pulls in old code for PR builds.  Perhaps even display the commit sha in the build.

----

I have seen this multiple times.

Here is an example from the CI logs:

```

Full Name:	Test-Cases.Group1-Docker-Commands.1-29-Docker-Checkpoint.Docker checkpoint ls
Start / End / Elapsed:	20170209 18:36:17.888 / 20170209 18:36:17.912 / 00:00:00.024
Status:	FAIL (critical)
Message:	'docker: 'checkpoint' is not a docker command.
See 'docker --help'.' does not contain 'vSphere Integrated Containers does not yet implement checkpointing'
00:00:00.021KEYWORD ${rc}, ${output} = OperatingSystem . Run And Return Rc And Output docker %{VCH-PARAMS} checkpoint ls test-busybox
00:00:00.000KEYWORD BuiltIn . Should Be Equal As Integers ${rc}, 1
00:00:00.001KEYWORD BuiltIn . Should Contain ${output}, vSphere Integrated Containers does not yet implement checkpointing
Documentation:	
Fails if container does not contain item one or more times.
Start / End / Elapsed:	20170209 18:36:17.911 / 20170209 18:36:17.912 / 00:00:00.001
18:36:17.912	FAIL	'docker: 'checkpoint' is not a docker command.
See 'docker --help'.' does not contain 'vSphere Integrated Containers does not yet implement checkpointing'	
```

Notice, it attempts to call 'docker %{VCH-PARAMS} pull busybox'.  This was not the latest code in the PR.  This is the code in the PR:

```
Docker checkpoint ls
    ${rc}  ${output}=  Run And Return Rc And Output  docker1.13 %{VCH-PARAMS} checkpoint ls test-busybox
    Should Be Equal As Integers  ${rc}  1
    Should Contain  ${output}  vSphere Integrated Containers does not yet implement checkpointing
```

Notice the code in the PR calls 'docker1.13'.  Resubmitting the same build pulled in the latest code.  I have seen this issue before.  In the past, it may require more than one build resubmission.

mhagen-vmware:
Why is this in To Do? It also needs to have a priority assigned.

sflxn:
@mhagen-vmware I added acceptance criteria above.

[...]

PR #6696, opened 2017-11-05
Add initial validation step to drone pipeline

zjs:
Add a step near the beginning of the drone pipeline to print details about the actual and expected commits to allow for easier detection and debugging when drone does not test the expected commit.

Because testing the wrong commit is worse than not testing at all (as it can lead to false confidence that something is working or wasted time debugging something that is no longer broken), this will fail the build if a mismatch is encountered.

Partially addresses #3897.


zjs:
> actually, if we are detecting the mismatch successfully here? We can just fix it instead of giving up right?

What would fixing it look like? For whatever reason, the merge commit we're attempting to check out isn't available.

mhagen-vmware:
Based on your description it sounded more like a race, so where you are going git logs, just do a git checkout of the correct commit hash.

.drone.yml
```diff
@@ -17,6 +17,18 @@ pipeline:
     # dont clone submodules
     recursive: false
 
+  display-status:
+    image: harbor.ci.drone.local/library/vic-integration-test:1.36
```
mdharamadas1: You may want to use some other light weight image for this.

zjs:
> Based on your description it sounded more like a race, so where you are going git logs, just do a git checkout of the correct commit hash with a retry.

It doesn't seem to be that simple; I'm not sure if the issue is with drone, GitHub, or some part of our infrastructure, but I had cases over the weekend where I seemed to be getting the wrong commit for more than an hour.

The goal of this PR is to "fail fast" when we're testing the wrong commit. That should avoid cases where we push code based on a false-positive (which I've done) or waste time debugging a false-negative (which I've also done).

If we see a lot of failures, we know it's worth investing resources to investigate and explore options for actually fixing the issue.

.drone.yml
zjs: I considered using some other image, but almost the entire pipeline seems to use this one. Homogeneity seems valuable, and the whole `display-status` step takes less than a second of a 10+ minute process so I'm not sure what using something lighter weight would gain us.

mdharamadas1: That timestamp probably doesn't include time to download that image which is `1.95GB` size. It doesn't matter much in passing case but mostly in failure case. Even though we are not gaining much in terms of time savings I think its always better to use lightest possible image ( docker best practice :-) )

mhagen-vmware:
absolutely, that is why I have LGTM'ed with my comment, I would much rather fail fast than not.  Was just hoping we could fix it when we detect it, but as you say it can take up to an hour then just detecting is probably best and we can address later a way to fix it.  But overall it looks like they might actually have it fixed as well based on the drone bug, so we probably just need to add this to the long list of reasons why we really need to upgrade to latest drone ASAP.

.drone.yml
mhagen-vmware: that image is hosted locally on a gigabit network and cached.  We almost never download the full image on a build.

zjs: The entire drone build takes ~9 seconds when this check fails (see, e.g., https://ci.vcna.io/vmware/vic/14467 or https://ci.vcna.io/vmware/vic/14468). Reducing that isn't worth the added complexity of introducing a different image into the process.

zjs: >  I think its always better to use lightest possible image ( docker best practice :-) )

My understanding is that this is a best practice because it reduces complexity; fewer things in an image mean fewer things that can go wrong.

In this case, using a different image _increases_ complexity in a different way.

PR #6703, opened 2017-11-06
VCH Creation API: Stop writing key files to disk [full ci]

### LINKED ISSUE #6650: VCH creation API leaves certificate files on disk
This can lead to an error about "cname option doesn't match existing server certificate" when re-creating a VCH with the same name as a previously-created VCH.

cc @cristianfalcone

zjs:
When creating a VCH via the API, we should not write key/certificate files to the API server's disk.

This change introduces the behavior as a flag so that it could be exposed as an option for the CLI at some point in the future, but does not expose it at this time.

Resolves #6650

### DIFF
--- a/cmd/vic-machine/common/certificate.go
+++ b/cmd/vic-machine/common/certificate.go
@@ -55,6 +55,7 @@ type CertFactory struct {
 	NoTLSverify  bool
 	KeyPEM       []byte
 	CertPEM      []byte
+	NoSaveToDisk bool
 }
 
 func (c *CertFactory) CertFlags() []cli.Flag {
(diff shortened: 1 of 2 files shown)


zjs:
Full CI was successful: https://ci.vcna.io/vmware/vic/14535

cmd/vic-machine/common/certificate.go
```diff
@@ -55,6 +55,7 @@ type CertFactory struct {
 	NoTLSverify  bool
 	KeyPEM       []byte
 	CertPEM      []byte
+	NoSaveToDisk bool
```
anchal-agrawal: We should consider calling this `SaveToDisk` since `if c.SaveToDisk`is easier to understand than `if !c.NoSaveToDisk` for all the checks below. Also, since booleans are false by default `c.Certs.NoSaveToDisk = true` wouldn't be needed either.

zjs: Normally, I would prefer to avoid "negatives" in boolean names. In this case, I think we want to default to saving the files to disk (as the CLI is still the primary way of interacting with this package) and we have exactly one code path where we don't want to save to disk (the create API handler).

I _think_ it's helpful to word the name such that the zero value is the one we want by default, even if it means we have to read double negatives in a few places.

anchal-agrawal: Makes sense 👍

mdharamadas1:
@zjs https://ci.vcna.io/vmware/vic/14535 seems like it didn't run full CI but only regression.

zjs:
> https://ci.vcna.io/vmware/vic/14535 seems like it didn't run full CI but only regression.

Good catch. I should have looked more closely at the results.

Issue #6502, opened 2017-10-04
Modify CI to run with restricted ops-user ( --ops-user)


lcastellano:
**User Statement:**
The VCH should be configured to use a ops-user ( --ops-user ) in our CI enviroment. Currently the CI runs with vCenter administrator credentials.

**Details:**
The CI should be modified to use the ops-user ( --ops-user ) instead of the vCenter administrator credentials. 

**Acceptance Criteria:**
CI runs creates VCH and runs test using the ops-user.

anchal-agrawal:
@lcastellano Please add a priority and team label(s), thanks.

lcastellano:
Manually closing this. PR references: #6709, #6774

PR #6709, opened 2017-11-07
Add support for new "vic-machine" flag to init Roles/Permission for ops-user, not exposed

### LINKED ISSUE #6505: VCH-Update: implementation of create/validate/repair of vSphere Permissions for ops-user and VCH resources

lcastellano:
In addition to the new flag, various bug fixes including the correct set of minimum permissions to run a container.

Fixes #6505 addendum

Issue: #6502 tracks the addition of the Nightly/Integration Test

### DIFF
--- a/cmd/vic-machine/common/ops_credentials.go
+++ b/cmd/vic-machine/common/ops_credentials.go
@@ -31,6 +31,7 @@ type OpsCredentials struct {
 	OpsUser     *string `cmd:"ops-user"`
 	OpsPassword *string
 	IsSet       bool
+	GrantPerms  bool
 }
 
 func (o *OpsCredentials) Flags(hidden bool) []cli.Flag {
@@ -66,13 +67,17 @@ func (o *OpsCredentials) ProcessOpsCredentials(isCreateOp bool, adminUser string
 			if adminPassword == nil {
 				return errors.New("Unable to use nil password from administrative user for operations user")
 			}
-
 			o.OpsPassword = adminPassword
+			// If using the admin user don't init roles and Permissions
+			o.GrantPerms = false
 			return nil
 		}
 	} else {
 		if o.OpsUser != nil {
 			o.IsSet = true
+		} else {
+			// If using the admin user don't init roles and Permissions
+			o.GrantPerms = false
 		}
 	}
 
--- a/lib/config/virtual_container_host.go
+++ b/lib/config/virtual_container_host.go
@@ -55,6 +55,8 @@ const (
 	GeneralHTTPSProxy  = "HTTPS_PROXY"
 	VICAdminHTTPProxy  = "VICADMIN_HTTP_PROXY"
 	VICAdminHTTPSProxy = "VICADMIN_HTTPS_PROXY"
+
+	AddPerms = "ADD"
 )
 
 // Can we just treat the VCH appliance as a containerVM booting off a specific bootstrap image
@@ -95,6 +97,9 @@ type VirtualContainerHostConfigSpec struct {
 
 	// configuration for vic-machine
 	CreateBridgeNetwork bool `vic:"0.1" scope:"read-only" key:"create_bridge_network"`
+
+	// grant ops-user permissions, string instead of bool for future enhancements
+	GrantPermsLevel string `vic:"0.1" scope:"read-only" key:"grant_permissions"`
 
 	// vic-machine create options used to create or reconfigure the VCH
 	VicMachineCreateOptions []string `vic:"0.1" scope:"read-only" key:"vic_machine_create_options"`
@@ -246,6 +251,14 @@ func (t *VirtualContainerHostConfigSpec) SetIsCreating(creating bool) {
 // IsCreating is checking if this configuration is for one creating VCH VM
 func (t *VirtualContainerHostConfigSpec) IsCreating() bool {
 	return t.ExecutorConfig.ID == CreatingVCH
+}
+
+func (t *VirtualContainerHostConfigSpec) SetGrantPerms() {
+	t.GrantPermsLevel = AddPerms
+}
+
+func (t *VirtualContainerHostConfigSpec) ShouldGrantPerms() bool {
+	return t.GrantPermsLevel == AddPerms
 }
 
 // AddNetwork adds a network that will be configured on the appliance VM
--- a/lib/install/management/configure.go
+++ b/lib/install/management/configure.go
@@ -111,6 +111,12 @@ func (d *Dispatcher) Configure(vch *vm.VirtualMachine, conf *config.VirtualConta
 	}
 
 	if err = d.update(conf, settings, isConfigureOp); err == nil {
+		if conf.ShouldGrantPerms() {
+			err = GrantOpsUserPerms(d.ctx, d.session.Vim25(), conf)
+			if err != nil {
+				return errors.Errorf("Cannot init ops-user permissions, failure: %s. Exiting...", err)
+			}
+		}
 		// compatible with old version's upgrade snapshot name
 		if oldSnapshot != nil && (vm.IsConfigureSnapshot(oldSnapshot, ConfigurePrefix) || vm.IsConfigureSnapshot(oldSnapshot, UpgradePrefix)) {
 			d.retryDeleteSnapshotByRef(&oldSnapshot.Snapshot, oldSnapshot.Name, conf.Name)
--- a/lib/install/management/rbac.go
+++ b/lib/install/management/rbac.go
@@ -72,111 +72,6 @@ type RBACResourcePermission struct {
 	permission types.Permission
 }
 
-var rolePrefix = "vic-vch-"
-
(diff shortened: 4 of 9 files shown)


cmd/vic-machine/common/ops_credentials.go
```diff
@@ -47,6 +48,12 @@ func (o *OpsCredentials) Flags(hidden bool) []cli.Flag {
 			Usage:  "Password or token for the operations user. Defaults to the credential supplied with target",
 			Hidden: hidden,
 		},
+		cli.BoolFlag{
+			Name:        "ops-init-perms",
+			Usage:       "Initialize Roles and Permissions for ops-user",
```
anchal-agrawal: `Initialize roles and permissions for ops-user` would be more consistent with the other usage texts above, what do you think?

lib/install/management/rbac.go
```diff
@@ -281,7 +176,7 @@ func (am *AuthzManager) setupPermissions(ctx context.Context) ([]RBACResourcePer
 	}
 
 	// Find bridged networks
-	bridgeNet, ok := am.configSpec.Networks["bridge"]
+	bridgeNet, ok := am.configSpec.Network.ContainerNetworks["bridge"]
```
anchal-agrawal: What's the purpose of this change?

lcastellano: This is a bug, fix. Taking the bridgeNet from the Networks map works with the Sim but bot with real VC.

lcastellano:
The integration test for this feature has to be done from scratch and requires creating users in VC which has not been done so far (no support in govmomi). This may require a substantial investigation/work. That's the reason why I broke it off in a different Issue (#6502) which is scheduled for the current sprint.

cmd/vic-machine/common/ops_credentials.go
```diff
@@ -47,6 +48,12 @@ func (o *OpsCredentials) Flags(hidden bool) []cli.Flag {
 			Usage:  "Password or token for the operations user. Defaults to the credential supplied with target",
 			Hidden: hidden,
 		},
+		cli.BoolFlag{
+			Name:        "ops-init-perms",
```
anchal-agrawal: Does `vic-machine configure` apply to this option? Ping @lcastellano @andrewtchin @mdubya66 - we made the call that for any new `vic-machine` option, `configure` support should be implemented simultaneously if applicable.

lcastellano: Working with George on vic-machine-configure..

lib/config/virtual_container_host.go
```diff
@@ -248,6 +253,14 @@ func (t *VirtualContainerHostConfigSpec) IsCreating() bool {
 	return t.ExecutorConfig.ID == CreatingVCH
 }
 
+func (t *VirtualContainerHostConfigSpec) SetGrantPerms() {
+	t.GrantPermsLevel = AddPerms
+}
+
+func (t *VirtualContainerHostConfigSpec) ShouldGrantPerms() bool {
+	return t.GrantPermsLevel != ""
```
anchal-agrawal: Based on my understanding, this will always return true with this change, which would mean that `GrantOpsUserPerms` runs for every `configure` operation. Is that valid? I'd also like to see the `vic-machine configure` integration test passing with this change - please run specific ci.

lcastellano: In the latest change I removed the flag from being available in vic-machine (both create and configure). I have done that for two reasons: need a better way to handle Boolean flags, lack of Integration tests. 

Eventually it will work this way:

Requesting the grant option in either create or configure, will do the granting and save the value of the option in the config for the VCH. This will be automatically be applied on each of the following configure request unless you explicitly specify --ops-grant-perms=false. Not specifying the flag will leave the situation as it was in the last create/configure operation. This is done to allow the user to omit the option in the following configure operations. Making it somewhat persistent. I have discussed with George at length about the approach and that is what we think is the best choice for now.

lcastellano:
Since the integration test is not ready yet. I have removed the option from vic-machine, the rest of this PR is in preparation of the adding the new option + bug fixes + some refactoring.

lcastellano:
Since the option is now removed from vic-machine this is definitely not the final merge.

lib/config/virtual_container_host.go
anchal-agrawal: > Not specifying the flag will leave the situation as it was in the last create/configure operation.

Could you point me to the code where this is handled? Thanks.

Also, if we're not enabling this feature in `vic-machine` yet, would it make sense to set `AddPerms` to `""` so `GrantOpsUserPerms` isn't invoked at all?

mdubya66:
@mhagen-vmware with the vic-machine changes removed is this clear to merge with unit tests? @lcastellano you may want to change the name of the PR since it is no longer adding the vic-machine flag.

lib/config/virtual_container_host.go
lcastellano: The code that handles the different flavor of the Bool Flag is not in this commit. That is why I removed its external visibility. To support the three variants:
- the option is not set,
- the option is set,
- the option is set to false,
We need to implement a new version of the Bool flag, that's why I have removed it from now.
Regarding AddPerms I am not sure what you mean. GrantOpsUserPerms() is called only when GrantPermLevel is set to a non empty string. In the current version GrantPermLevel is always the empty string.

anchal-agrawal: > In the current version GrantPermLevel is always the empty string.

We're saying `AddPerms = "ADD"` and `t.GrantPermsLevel = AddPerms`, so `ShouldGrantPerms()` will return true. Are we setting `AddPerms = ""` somewhere else?

mhagen-vmware:
I don't see any unit tests?

lcastellano:
I have added some unit tests. Some of the changes will only be testable when the option is exposed in vic-machine.

anchal-agrawal:
Code LGTM. I'd suggest renaming the commit header to `Add support to init ops-user perms` - this indicates that while this change adds some functionality, it's not enabled in vic-machine (as there's no reference to vic-machine or its flags in the commit message)

lib/config/virtual_container_host.go
hickeng: I'm not sure if this is retreading the same ground as above. Was expecting this line to be:
```go
   return t.GrantPermsLevel == AddPerms
```

lib/install/management/configure.go
```diff
@@ -111,6 +111,11 @@ func (d *Dispatcher) Configure(vch *vm.VirtualMachine, conf *config.VirtualConta
 	}
 
 	if err = d.update(conf, settings, isConfigureOp); err == nil {
+		if conf.ShouldGrantPerms() {
+			if err = GrantOpsUserPerms(d.ctx, d.session.Vim25(), conf); err != nil {
```
hickeng: Not going to block the PR for it, but I find the uncollapsed form of this is significantly easier to visually parse for important operations; the collapsed form needs extra effort to tell if there's a significant side effect or if it's just a test for existing state.
Only time I think it's really necessary to use the collapsed form is when scoping of the return is important:
```go
err := GrantOpsUserPerms(d.ctx, d.session.Vim25(), conf)
if err != nil {
	return errors.Errorf("Cannot init ops-user permissions, failure: %s. Exiting...", err)
}
```

lib/install/validate/validator_test_sim_util.go
```diff
@@ -209,6 +209,9 @@ func (v *Validator) validateForSim(ctx context.Context, input *data.Data) (*conf
 	vmRef := vm.Reference()
 	conf.SetMoref(&vmRef)
 
+	// Remove bridge network from main section of config, to better emulate real VC behaviour
+	conf.Networks["bridge"] = nil
```
hickeng: I'm not sure why this needs doing - what was the bug that you mentioned that relates to this?

lcastellano: There are two changes related to this problem. The first is in rbac.go line 179. The original code was looking for the bridged network mo inside ExecutorConfig.Networks. That worked fine in the Simulator but failed when running against my VC test bed. For that to work I needed to change the rbac code to look at Network.ContainerNetworks. In case of the simulator the bridged network mo is found in both places. In case of my VC test bed it is only found in Network.ContainerNetworks. This modification tries to make the simulator behavior match the VC behavior, where ExecutorConfig.Networks can be nil.

mdubya66:
@hickeng please review for config group

PR #6718, opened 2017-11-08
Extract constants for magic values [specific ci=Group23-VIC-Machine-Service]

zjs:
Extract constants for several "magic values" identified during review of the merge commit for the VCH creation API feature branch.

See also: #6665


hickeng:
These MUST go into `lib/constants/` - they MUST be used by both sides. In this case this means an update to `cmd/vicadmin` to reference that constant as MUST has been optional in the past.

zjs:
> These MUST go into `lib/constants/` ...

Also fix:
 * https://github.com/vmware/vic/pull/6665#discussion_r149548721

Issue #6700, opened 2017-11-06
unable to wait for process launch status: container VM has unexpectedly powered off


matthewavery:
Seen on 11/5/2017
VC version: 6.5
VIC BUILD: 14432
suite: 5-4-High-Availability
Test: "Test"

Looks like a CDE from the CommitHandler: 

```
CommitHandler error on handle(3487c0ee5c39c5f12bdbdaadf139a2d7) for eaeb57361728956b10fee6994050e14ef1be3b9ffe77d63d48a8aae06b372307: unable to wait for process launch status: container VM has unexpectedly powered off
```

Some more looking into the tether does not have anything that readily jumps out as bad. Looks like the container may have actually finished running(needs more eyes on the tether logs). 

[5-4-High-Availability.zip](https://github.com/vmware/vic/files/1447236/5-4-High-Availability.zip)

matthewavery:
Marked as high as it is a nightly failure. We need to atleast understand what happened here and see what we can do about it.

hickeng:
The command run in the container was `/bin/ls` which we would expect to exit almost instantly, causing the container to shut down. That's completely legitimate.

The Commit logic should be allowing for the cVM to cleanly complete before the wait completes - it'll be hard to test without vcsim as this is one of the archetypal timing issues that are hard to elicit without being able to inject delays.

hickeng:
I've created a PR that's speculative for addressing option (1) in the previous comment.
It is not marked as fixing this issue as we do not have a recreation - only approach I can see is a stress load of very short-lived containers.

hickeng:
This reoccurred in a manual nimbus run of the HA tests. A stress load of `run -d busybox ls` against VC seems necessary to progress this any further.

AngieCris:
Seen again on 12/20/2017
VC version: 6.0
VIC BUILD: vic 1.3.0 rc4
suite: 5-22-NFS-Volume
Test step: `Docker inspect mount data after reboot`

After VCH reboot, docker info exists with error code 1. 
In portlayer log, the handshake client detects closed network connection:
```
Dec 20 2017 03:35:32.283Z DEBUG connection read: EOF
Dec 20 2017 03:35:32.283Z INFO  Closing the connection
Dec 20 2017 03:35:32.283Z DEBUG connection loop write-side closed
Dec 20 2017 03:35:32.283Z DEBUG data handler closed
Dec 20 2017 03:35:32.283Z INFO  telnet connection closed
[...]

PR #6724, opened 2017-11-08
Allow out-of-order power/property changesets

hickeng:
This allows for the possibility of a changeset not being time ordered.
This is speculative but may well occur if pulling from vCenter and the
power and property change via host sync lose the original ordering on the
host.

Towards: #6700 
Requires: #6742


matthewavery:
lgtm

Issue #6697, opened 2017-11-06
Longevity failure: container name in use


mhagen-vmware:
It appears that we randomly tried to pick the same name we already had randomly given to a container:
```
Nov  5 2017 22:56:06.890Z ERROR Handler for POST /v1.23/containers/create returned error: conflict. The name "happy_hoover" is already in use by container 70af98fc862598113ed501952be1f14675abcf5ba2539f5a5b5da9665699364d. You have to remove (or rename) that container to be able to re use that name.
```

[container-logs.zip](https://github.com/vmware/vic/files/1446910/container-logs.zip)

anchal-agrawal:
This is likely due to https://github.com/vmware/vic/blob/cb687c09be5b88934d20f8d10e230a35b8e5248b/lib/apiservers/engine/backends/container.go#L2012

The `GetRandomName` func has a retry parameter which, if non-zero, causes the name to be appended by a random number b/w 0 and 10. We can also have an ad-hoc retry mechanism:

* get name, retry if it exists in the persona container cache, retry again with the retry parameter if needed
* get a bunch of names at once and take the first one that doesn't exist

Both approaches aren't foolproof theoretically but offer a best-effort strategy that will work in most use cases (given that the original error is rare as it is). For true randomness, we could consider appending the name with a UUID.

mhagen-vmware:
Do we really need randomness?  We own all of the container VMs within a resource pool.. can we not just ensure we don't use an already used name?

anchal-agrawal:
See https://github.com/vmware/vic/pull/6742.

hickeng:
@sgairo I ended up creating a PR for this as it failed in a PR build for a different issue of mine this morning. Didn't realize there was an existing issue.

Issue #334, opened 2016-03-29
docker personality needs reservation feature for concurrent operations


caglar10ur:
so that we don't end up using the same id or name for different containers.

hickeng:
Hit this in https://ci.vmware.run/vmware/vic/6087 #2820 for the name collision

PR #6742, opened 2017-11-09
Reserve generated names as well as user supplied

hickeng:
#6561 ensured that user supplied names were reserved during creation
but did not do the same for generated names. 
Build https://ci.vcna.io/vmware/vic/14611 failed due to a generated
name collision.

Followup #6261 #6561
Fixes: #6697 
Towards: #334 (still requires id reservation)

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -63,6 +63,7 @@ import (
 	"github.com/vmware/vic/lib/archive"
 	"github.com/vmware/vic/lib/constants"
 	"github.com/vmware/vic/lib/metadata"
+	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/retry"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/uid"
@@ -132,6 +133,9 @@ var (
 	containerByPort map[string]string // port:containerID
 
 	ctx = context.TODO()
+
+	// allow mocking
+	randomName = namesgenerator.GetRandomName
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -592,8 +592,18 @@ func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (contain
...
-		return containertypes.ContainerCreateCreatedBody{}, derr.NewRequestConflictError(err)
+	if config.Name != "" {
+		err := cache.ContainerCache().ReserveName(container, config.Name)
+		if err != nil {
+			return containertypes.ContainerCreateCreatedBody{}, derr.NewRequestConflictError(err)
+		}
+	} else {
+		for {
+			config.Name = namesgenerator.GetRandomName(0)
+			if cache.ContainerCache().ReserveName(container, config.Name) == nil {
```
anchal-agrawal: We should add a test case in the `1-04-Docker-Create` suite for this. There's already a test at https://github.com/vmware/vic/blob/master/tests/test-cases/Group1-Docker-Commands/1-04-Docker-Create.robot#L194 - we can add this case there.

hickeng: I am unaware of a means to manipulate urandom to force a collision - I will look at adding a unit test and mocking `GetRandomName`.

anchal-agrawal: Oh right, my bad!

hickeng:
@anchal-agrawal I've updated it for a limited retry and using the numeric suffix mechanism.

sflxn:
lgtm

Issue #6559, opened 2017-10-12
Tests: VCH multi NIC configurations


hickeng:
**Story**
As a VIC developer I want to know that the various multi-NIC deployments are functional
As a viadmin I expect clear documentation on what NIC configurations are permitted or not permitted

**Detail**
The 3 NIC limitation described in #2802 has been addressed so we need to expand the testing configurations that use multiple port groups. This should encompass subnet segregation between vCenter and ESX hosts to reflect best practice deployment guidelines.

@pdaigle @emeirell can you comment on use cases for more complex scenarios that we don't currently support and raise issues for support if necessary:
- link aggregation style (multiple NICs on same network, different roles)
- asymmetric setup (e.g. forwarded ports should be directly served to client network as well as public, but not visa versa)

**Acceptance**
- [ ] confirm that test env has subnet segregation between vCenter and ESX
- [ ] test each of the follow deployments and assert connectivity/lack of from the networks:
  - [ ] public/client/management shared
  - [ ] public/client shared (2)
  - [ ] no sharing (3)

I'd suggest a mix of (2) and (3) as the normal test deployment basis as those are the common customer scenarios.

Assigned to @gigawhitlocks for decomposition into actual work required against current state of VC test env.

mhagen-vmware:
This is best handled in nimbus, nightly tests.  If I am understanding the scenarios correctly:
1. public/client/management all using the same DPG
2. public/client using same DPG and management using different DPG
3. all three interfaces all using different DPGs.

This should be easily accomplished via our existing patterns and govc commands to setup the networking as desired.

mhagen-vmware:
Removing from the Epic, as if agreed on my previous comment, then this is unrelated to our VC CI efforts.

hickeng:
@mhagen-vmware If CI is deployed following VMware best practice we will have most of this in place already (i.e. separate management network and VC/ESX subnet segregation).
I see no reason not to be able to exercise our recommended deployment config for VIC in CI as it doesn't seem to require significant additional configuration complexity (2 additional DPGs with VLAN assignment).

Using static IPs is something we should likely confine to nimbus for now as we'd otherwise need an IPAM mechanism for reserving/releasing addresses indirectly (e.g. out-of-band DHCP?).

PR #6743, opened 2017-11-10
 Handle routing destinations and static addresses in API like the CLI [specific ci=Group23-VIC-Machine-Service]

### LINKED ISSUE #6715: VCH Creation API: Decide what to do if RoutingDestinations is empty
Currently, the `fromGateway` method used by the VCH creation API handlers is unclear about the behavior if the supplied `RoutingDestinations` are empty. We should determine what work, if any, is required here.

zjs:
The `vic-machine` CLI has differing requirements for gateway routing information, depending on the type of network.

According to the CLI help:
 - a client gateway must specify one or more routing destinations
 - a public gateway must not specify any routing destinations
 - a management gateway must specify one or more routing destinations
 - a container gateway must specify exactly one routing destination

This does not seem to be enforced in code, and may simply be more of a suggestion about how these gateways should be used than a requirement.

Update the parsing for client, public, and management to support all zero or more routing destinations in all cases; defer to the existing `ProcessNetwork` code to ensure consistent validation behavior now and in the future.

Additionally update the parsing for container to provide a clear error message if the expected routing destination is not supplied.

Fixes #6715

---

The vic-machine CLI requires that static addresses are specified as a CIDR, which allows the static address and subnet mask to be supplied in a compact way on the command line. This pattern does not allow for static addresses to be expressed in terms of a hostname.

Update the API to match this convention. We may wish to allow for more flexibility in the API in the future, but there's value in at least starting with consistent behavior.

### DIFF
--- a/lib/apiservers/service/restapi/handlers/vch_create.go
+++ b/lib/apiservers/service/restapi/handlers/vch_create.go
@@ -233,7 +233,7 @@ func buildCreate(op trace.Operation, d *data.Data, finder *find.Finder, vch *mod
 				}
 				c.ClientNetworkName = path
 				c.ClientNetworkGateway = fromGateway(vch.Network.Client.Gateway)
-				c.ClientNetworkIP = fromNetworkAddress(vch.Network.Client.Static)
+				c.ClientNetworkIP = fromCIDR(&vch.Network.Client.Static)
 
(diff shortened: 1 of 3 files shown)


zjs:
Note that until #6559 is addressed, it is not feasible to write a comprehensive positive test case for this functionality.

PR #6745, opened 2017-11-10
Don't try to autocomplete the password field

gigawhitlocks:
Got a request from Pietro for a small change to the password field in vicadmin. This is a mostly-cosmetic change that shouldn't affect functionality at all, but will make for happier "scanners and customers" I am told. :)

Not attached to an issue -- this is super minor.


gigawhitlocks:
yeah @anchal-agrawal I can do that when I merge

zjs:
> that shouldn't affect functionality at all

Doesn't this have the functional change of preventing (many) password managers and browsers from auto-filling passwords?

hickeng:
Interestingly, vIDM allows for autocomplete (frustratingly when asking for an RSA token).

gigawhitlocks:
@zjs theoretically yes, but notably Chrome ignores this attribute!

PR #6747, opened 2017-11-10
Handle IP ranges in the API like in the CLI [specific ci=Group23-VIC-Machine-Service]

zjs:
The `vic-machine` CLI allows IP ranges to be expressed in CIDR notation or as simple ranges in some places, but requires that CIDR notation be used in others.

Initially there was a hope of making the API behave more consistently, but that requires changes to underlying logic (some of which is not well covered by existing tests).

For now, be consistent with the CLI.

### DIFF
--- a/lib/apiservers/service/swagger.json
+++ b/lib/apiservers/service/swagger.json
@@ -544,9 +544,14 @@
     },
     "CIDR": {
       "type": "string",
-      "pattern": "^(([1-9]?[0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\\.?){4}(\/([0-9]|[1-2][0-9]|3[0-2]))?$"
+      "description": "A range of IP addresses in CIDR notation (like 192.0.2.0/24).",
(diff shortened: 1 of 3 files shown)


lib/apiservers/service/swagger.json
```diff
@@ -544,33 +544,25 @@
     },
     "CIDR": {
       "type": "string",
-      "pattern": "^(([1-9]?[0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\\.?){4}(\/([0-9]|[1-2][0-9]|3[0-2]))?$"
+      "description": "A range of IP addresses in CIDR notation (like 192.0.2.0/24).",
+      "pattern": "^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/([0-9]|[1-2][0-9]|3[0-2]))$"
```
andrewtchin: lol

PR #6755, opened 2017-11-13
[skip ci] Grab NFSD logs to debug potential problems

mhagen-vmware:


sgairo:
Is it okay to not stub out the password?

mhagen-vmware:
I see 103,000 results with this password leaked... I think we are probably alright:
https://www.google.com/search?q=ca%24hc0w&oq=ca%24hc0w&aqs=chrome.0.69i59.2541j0j7&client=ubuntu&sourceid=chrome&ie=UTF-8

And this is just the password to a temp server that isn't publicly accessible as well.

hickeng:
It's a solid defense about leaking a specific datum, but still not good practice. Is there an issue with making it an env variable that's in the secrets file?

PR #6758, opened 2017-11-13
Fix logging to op logging

matthewavery:
Noticed some missed `op.log` style changes in the exec handlers. Threw this together really fast to address it.


matthewavery:
nah not really. @sgairo This was something that @hickeng noticed while we were looking through the exec code.

PR #6764, opened 2017-11-14
Add unit tests for VCH create

### LINKED ISSUE #6018: Implement unit tests for the API handler for creating a VCH
Implement unit tests for the API handler for creating a VCH (#6017)

jzt:
This main purpose of this change is to test the `buildCreate` function in vch_create.go, to verify that the `Create` struct that is generated matches the same struct that would be generated using the CLI.

There are some tests around help functions as well.

Fixes #6018

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -31,6 +31,7 @@ import (
 	"github.com/vmware/vic/lib/install/data"
 	"github.com/vmware/vic/lib/install/management"
 	"github.com/vmware/vic/lib/install/validate"
+
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/trace"
 )
@@ -65,7 +66,7 @@ type Create struct {
 	*data.Data
 	Certs             common.CertFactory
 	containerNetworks common.CNetworks
-	registries        common.Registries
+	Registries        common.Registries
 
 	volumeStores common.VolumeStores
 
@@ -243,17 +244,17 @@ func (c *Create) Flags() []cli.Flag {
 		Hidden:      true,
 	})
 
-	registries := c.registries.Flags()
+	registries := c.Registries.Flags()
 	registries = append(registries,
 		cli.StringSliceFlag{
 			Name:  "insecure-registry, dir",
-			Value: &c.registries.InsecureRegistriesArg,
+			Value: &c.Registries.InsecureRegistriesArg,
 			Usage: "Specify a list of permitted insecure registry server addresses",
 		})
 	registries = append(registries,
 		cli.StringSliceFlag{
 			Name:  "whitelist-registry, wr",
-			Value: &c.registries.WhitelistRegistriesArg,
+			Value: &c.Registries.WhitelistRegistriesArg,
 			Usage: "Specify a list of permitted whitelist registry server addresses (insecure addresses still require the --insecure-registry option in addition)",
 		})
 
@@ -317,7 +318,7 @@ func (c *Create) Flags() []cli.Flag {
 	return flags
 }
 
-func (c *Create) processParams() error {
+func (c *Create) ProcessParams() error {
 	defer trace.End(trace.Begin(""))
 
 	if err := c.HasCredentials(); err != nil {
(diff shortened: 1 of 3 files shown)


lib/apiservers/service/restapi/handlers/vch_create_test.go
```diff
@@ -0,0 +1,296 @@
...
+	actual, err = fromManagedObject(op, mf, "t", m)
+	assert.NoError(t, err, "Expected nil error, got %#v", err)
+	assert.Equal(t, expected, actual, "Expected %s, got %s", expected, actual)
+
+	m.ID = "testID"
+
+	expected = m.ID
+	actual, err = fromManagedObject(op, mf, "t", m)
+	assert.NoError(t, err, "Expected nil error, got %#v", err)
+	assert.Equal(t, expected, actual, "Expected %s, got %s", expected, actual)
```
zjs: Curious (since I'm new to unit testing in go): what's the goal of explicitly passing `Expected %s, got %s", expected, actual`? Shouldn't the default behavior be to print something along those lines?

lib/apiservers/service/restapi/handlers/vch_create_test.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+
+	mf := mockFinder{}
+
+	expected = m.Name
+	actual, err = fromManagedObject(op, mf, "t", m)
+	assert.NoError(t, err, "Expected nil error, got %#v", err)
+	assert.Equal(t, expected, actual, "Expected %s, got %s", expected, actual)
+
+	m.ID = "testID"
```
zjs: This doesn't clear the `Name`. Should there be a separate test case for having an ID and no name?

lib/apiservers/service/restapi/handlers/vch_create_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+	"github.com/vmware/govmomi/list"
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/cmd/vic-machine/common"
+	"github.com/vmware/vic/cmd/vic-machine/create"
+	"github.com/vmware/vic/lib/apiservers/service/models"
+	"github.com/vmware/vic/pkg/trace"
+)
+
+type mockFinder struct {
```
zjs: Would it be valuable to make this embed a testify `mock.Mock` so that we can verify that the `Element` method is called with the expected arguments?

jzt: Sure, I will look into this.

lib/apiservers/service/restapi/handlers/vch_create_test.go
jzt: That's actually a good point - currently, the func will return `Name` if `ID` is `nil`. So we could update this to test expect an empty string in that case, if that's the intended behavior.

lib/apiservers/service/restapi/handlers/vch_create_test.go
jzt: I've been using this pattern since I started because of some other test I referenced in the beginning. :)

But, upon investigating:
```go
func TestTest(t *testing.T) {                
        assert.Equal(t, "expected", "actual")
}
```
```
--- FAIL: TestTest (0.00s)
  Error Trace:  vch_create_test.go:47
  Error:    Not equal: "expected" (expected)
              != "actual" (actual)
```

jzt: Fixed

lib/apiservers/service/restapi/handlers/vch_create_test.go
jzt: Will add this at a later time as per our slack convo.

lib/apiservers/service/restapi/handlers/vch_create_test.go
zjs: Good enough for now :+1:

cmd/vic-machine/create/create.go
```diff
@@ -31,6 +31,7 @@ import (
 	"github.com/vmware/vic/lib/install/data"
 	"github.com/vmware/vic/lib/install/management"
 	"github.com/vmware/vic/lib/install/validate"
+
```
zjs: nit: I don't think this line should be here

jzt: Agreed, fixed

PR #6769, opened 2017-11-15
Move slack notif step to the end in CI

mdharamadas1:
Moving all slack notification steps in CI to the end to avoid situations like https://ci.vcna.io/vmware/vic/14675 build which posted success msg on slack but build actually failed due to downstream trigger failure.


mdharamadas1:
@mhagen-vmware I cannot verify this fully until merge to `master`, let me know if I'am missing anything here.

mdharamadas1:
Sure @mhagen-vmware

mdharamadas1:
@mdubya66 This will fail builds even if all the primary steps like build, integration steps, publishing, etc are passed and if any other steps like downstream trigger or code coverage steps fail. Just want to make sure you are OK with this?

mdubya66:
i defer to mike hagen

mdharamadas1:
@mhagen-vmware I know you approved this but look at my question to Matt above, which is now deferred to you.. let me know what you think?

mhagen-vmware:
It is fine, the only issue we had in the past was failure to publish because of coverage server, this doesn't cause any problems like that

mdharamadas1:
Thanks @mhagen-vmware !

PR #6770, opened 2017-11-15
[skip ci]Add initial future test

mhagen-vmware:


mhagen-vmware:
From George:
Hey Mike,
The test confirms API function which is core. The portions liable to fail if we are impacted:
    install of VCH >90 days after authoring ISO (this is what we were hit by in 1.0 iirc)
    enabling SSH via vic-machine debugvic-machine debug should invoke chage in order to allow reset of password https://github.com/vmware/vic/blob/76df62cf139c4389f2cd830a77f9f4775921074a/cmd/vic-init/toolbox.go#L235 but that should be confirmed.

PR #6774, opened 2017-11-15
Enable --ops-grant-perms in vic-machine. Add nightly and unit tests

### LINKED ISSUE #6691: Integrate vic-machine create command with ops-user permission setting
**User Statement:**
Add an option to vic-machine create to trigger the code that automatically sets permissions for the ops-user.

**Details:**
The user of vic-machine create will have the option of selecting a previously created ops-user. The minimal permission for that ops-user will be set automatically.

**Acceptance Criteria:**
- New option for vic-machine create
- Integration test

lcastellano:
Fixes #6691

During create the --ops-grant-perms flag must be used in combination with the argument --ops-user. Otherwise the flag is left undefined and the granting is not performed. During configure, priority goes to the value stored in the SpecConfig, unless there is a specific request from the command line. The flag --ops-grant-perms=false can be used to turn it off while --ops-grant-perms can be used to turn it on.

### DIFF
--- a/cmd/vic-machine/common/ops_credentials.go
+++ b/cmd/vic-machine/common/ops_credentials.go
@@ -30,8 +30,8 @@ import (
 type OpsCredentials struct {
 	OpsUser     *string `cmd:"ops-user"`
 	OpsPassword *string
+	GrantPerms  *bool
 	IsSet       bool
-	GrantPerms  bool
 }
 
 func (o *OpsCredentials) Flags(hidden bool) []cli.Flag {
@@ -48,6 +48,12 @@ func (o *OpsCredentials) Flags(hidden bool) []cli.Flag {
 			Usage:  "Password or token for the operations user. Defaults to the credential supplied with target",
 			Hidden: hidden,
 		},
+		cli.GenericFlag{
+			Name:   "ops-grant-perms",
+			Value:  flags.NewOptionalBool(&o.GrantPerms),
+			Usage:  "Create roles and grant required permissions to the specified ops-use",
+			Hidden: hidden,
+		},
 	}
 }
 
@@ -62,22 +68,26 @@ func (o *OpsCredentials) ProcessOpsCredentials(isCreateOp bool, adminUser string
 
 	if isCreateOp {
 		if o.OpsUser == nil {
+			// Check if there was a request to setup ops-user Roles and Permissions
+			if o.GrantPerms != nil {
+				// If true return error
+				if *o.GrantPerms {
+					return errors.Errorf("Invalid use of flag: --ops-grant-perms. Cannot setup Roles and Permissions for administrative user.")
+				}
+				// If false ignore
+				o.GrantPerms = nil
+			}
 			log.Warn("Using administrative user for VCH operation - use --ops-user to improve security (see -x for advanced help)")
 			o.OpsUser = &adminUser
 			if adminPassword == nil {
 				return errors.New("Unable to use nil password from administrative user for operations user")
 			}
 			o.OpsPassword = adminPassword
-			// If using the admin user don't init roles and Permissions
-			o.GrantPerms = false
 			return nil
 		}
 	} else {
 		if o.OpsUser != nil {
 			o.IsSet = true
-		} else {
-			// If using the admin user don't init roles and Permissions
-			o.GrantPerms = false
 		}
 	}
 
--- a/cmd/vic-machine/common/ops_credentials_test.go
+++ b/cmd/vic-machine/common/ops_credentials_test.go
@@ -60,4 +60,45 @@ func TestProcessOpsCredentials(t *testing.T) {
 	configureOps.OpsUser = nil
 	err = configureOps.ProcessOpsCredentials(isCreateOp, "", nil)
 	assert.NotNil(t, err)
+
+	// Test correct grant permissions
+	createOps = &OpsCredentials{}
+	createOps.OpsUser = &opsUser
+	createOps.OpsPassword = &opsPassword
+	grantPerms := true
+	createOps.GrantPerms = &grantPerms
+	err = createOps.ProcessOpsCredentials(true, adminUser, &adminPassword)
+	assert.NoError(t, err)
+	assert.Equal(t, *createOps.OpsUser, opsUser)
+	assert.Equal(t, *createOps.OpsPassword, opsPassword)
+	assert.True(t, *createOps.GrantPerms)
+
+	// Create Negative test: grantPerms is set to true but there is no ops-user,
+	// grantPerms should be reset to false
+	createOps = &OpsCredentials{}
+	createOps.OpsUser = nil
+	createOps.OpsPassword = nil
+	grantPerms = true
+	createOps.GrantPerms = &grantPerms
+	err = createOps.ProcessOpsCredentials(true, adminUser, &adminPassword)
+	assert.Error(t, err)
+
+	// Create Negative test: grantPerms is set to true but there is no ops-user,
+	// grantPerms should be reset to false
+	createOps = &OpsCredentials{}
+	createOps.OpsUser = nil
+	createOps.OpsPassword = nil
+	grantPerms = false
+	createOps.GrantPerms = &grantPerms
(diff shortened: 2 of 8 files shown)


cmd/vic-machine/common/ops_credentials.go
```diff
@@ -48,6 +48,12 @@ func (o *OpsCredentials) Flags(hidden bool) []cli.Flag {
 			Usage:  "Password or token for the operations user. Defaults to the credential supplied with target",
 			Hidden: hidden,
 		},
+		cli.GenericFlag{
+			Name:   "ops-grant-perms",
+			Value:  flags.NewOptionalBool(&o.GrantPerms),
+			Usage:  "Password or token for the operations user. Defaults to the credential supplied with target",
```
zjs: Copy & paste error

lcastellano: Yes.. thanks

cmd/vic-machine/common/ops_credentials_test.go
```diff
@@ -60,4 +60,34 @@ func TestProcessOpsCredentials(t *testing.T) {
...
+
+	// Create Negative test: grantPerms is set to true but there is no ops-user,
+	// grantPerms should be reset to false
+	createOps = &OpsCredentials{}
+	createOps.OpsUser = nil
+	createOps.OpsPassword = nil
+	grantPerms = true
+	createOps.GrantPerms = &grantPerms
+	err = createOps.ProcessOpsCredentials(true, adminUser, &adminPassword)
+	assert.NoError(t, err)
```
hickeng: I'm inclined to have this be an error instead of silently ignored.

lib/install/management/configure.go
```diff
@@ -110,13 +110,18 @@ func (d *Dispatcher) Configure(vch *vm.VirtualMachine, conf *config.VirtualConta
...
+	// If successful try to grant permissions to the ops-user
+	if err == nil && conf.ShouldGrantPerms() {
+		err = GrantOpsUserPerms(d.ctx, d.session.Vim25(), conf)
+		if err != nil {
+			// Fall through to roll back
+			err = errors.Errorf("Cannot init ops-user permissions, failure: %s", err)
 		}
+	}
+
+	if err == nil {
```
hickeng: There's some weird indenting going on here making it very unclear what the clean exit path is below. Could just be the PR rendering.

lib/install/validate/validator_test.go
```diff
@@ -175,6 +174,127 @@ func TestMain(t *testing.T) {
...
+
+		// Test 1: current value in conf is the empty string,
+		// input is true, after validation
+		// the new GrantPerms value in conf should be
+		// "config.AddPerms"
+		conf.GrantPermsLevel = ""
+		input.OpsCredentials.GrantPerms = &opsGrantPermissions
+		validator, err := NewValidator(ctx, input)
+		if err != nil {
+			t.Errorf("Failed to new validator: %s", err)
```
hickeng: Failed to create validator.
There is also another testify package, `require`. You could say the following without the `if` clause and it would abort the test immediately. IIRC `t.Errorf` will fail the test but it will still continue to run past that point. I think you'd need `t.Fatalf` which will use `t.FailNow` instead of `t.Fail`:
```
require.NoError(t, err, "Failed to create validator")
```

tests/manual-test-cases/Group5-Functional-Tests/5-25-OPS-User-Grant.robot
```diff
@@ -0,0 +1,60 @@
...
+    Log To Console  Setting up ops-user: ${ops_user_name}
+    ${rc}  ${output}=  Run And Return Rc And Output  sshpass -p vmware ssh -o StrictHostKeyChecking=no root@${vc-ip} /usr/lib/vmware-vmafd/bin/dir-cli user create --account ${ops_user_base_name} --user-password ${ops_user_password} --first-name ${ops_user_base_name} --last-name ${ops_user_domain} --password ${vc_admin_password}
+    Log  User Create ${ops_user_name}, rc: ${rc}, output: ${output}
+    Should Be Equal As Integers  ${rc}  0
+    ${rc}  ${output}=  Run And Return Rc And Output  sshpass -p vmware ssh -o StrictHostKeyChecking=no root@${vc-ip} /usr/lib/vmware-vmafd/bin/dir-cli user find-by-name --account ${ops_user_base_name} --password ${vc_admin_password}
+    Log  User Find ${ops_user_base_name}, rc: ${rc}, output: ${output}
+    Should Be Equal As Integers  ${rc}  0
+
+    Log To Console  Running govc to set permissions ..
+    ${out}=  Run  govc permissions.set -principal ${ops_user_name} -role ReadOnly -propagate=false /
```
hickeng: A comment to note why this is necessary - it should end up in our docs (and we should ensure we have a doc issue for that) - but worth noting here too.

tests/manual-test-cases/Group5-Functional-Tests/5-25-OPS-User-Grant.robot
```diff
@@ -0,0 +1,60 @@
...
+    Should Be Empty  ${out}
+    ${out}=  Run  govc role.usage
+    Log  Output, govc role.usage: ${out}
+
+*** Test Cases ***
+Test
+    Log To Console  \nStarting test...
+    Install VIC Appliance To Test Server  additional-args=--ops-user ${ops_user_name} --ops-password ${ops_user_password} --ops-grant-perms
+
+    Run Regression Tests
```
hickeng: Can we add a test for the restrictive nature of the ops user? For example a govc command that should fail because we're trying to inspect a pool/network/datastore or similar that's not used by the VCH.
We're trying to catch the case where we accidentally make the grant way too broad.

cmd/vic-machine/common/ops_credentials.go
```diff
@@ -48,6 +48,12 @@ func (o *OpsCredentials) Flags(hidden bool) []cli.Flag {
 			Usage:  "Password or token for the operations user. Defaults to the credential supplied with target",
 			Hidden: hidden,
 		},
+		cli.GenericFlag{
+			Name:   "ops-grant-perms",
+			Value:  flags.NewOptionalBool(&o.GrantPerms),
+			Usage:  "Initialize roles and permissions for ops-user",
```
hickeng: "Create roles and add required permissions to the specified ops-user"

tests/manual-test-cases/Group5-Functional-Tests/5-25-OPS-User-Grant.robot
lcastellano: Yes. I will open an issue to update the docs. We also need to update the manual settings since they are different when using Resource Pool.

cmd/vic-machine/common/ops_credentials_test.go
```diff
@@ -60,4 +60,45 @@ func TestProcessOpsCredentials(t *testing.T) {
...
+	grantPerms := true
+	createOps.GrantPerms = &grantPerms
+	err = createOps.ProcessOpsCredentials(true, adminUser, &adminPassword)
+	assert.NoError(t, err)
+	assert.Equal(t, *createOps.OpsUser, opsUser)
+	assert.Equal(t, *createOps.OpsPassword, opsPassword)
+	assert.True(t, *createOps.GrantPerms)
+
+	// Create Negative test: grantPerms is set to true but there is no ops-user,
+	// grantPerms should be reset to false
```
hickeng: comment update. should return error.

cmd/vic-machine/common/ops_credentials_test.go
```diff
@@ -60,4 +60,45 @@ func TestProcessOpsCredentials(t *testing.T) {
...
+	createOps = &OpsCredentials{}
+	createOps.OpsUser = nil
+	createOps.OpsPassword = nil
+	grantPerms = true
+	createOps.GrantPerms = &grantPerms
+	err = createOps.ProcessOpsCredentials(true, adminUser, &adminPassword)
+	assert.Error(t, err)
+
+	// Create Negative test: grantPerms is set to true but there is no ops-user,
```
hickeng: comment update - grantPerms is set to false

Issue #6612, opened 2017-10-24
Ensure VCH creation log streaming handles concurrent creation operations


zjs:
`vchlogger.go` currently uses process-global variables (e.g., `pipe`, `signalChan`). This works for command-line use, where at most one creation occurs in each process. However, the API needs to be able to handle multiple, concurrent creation requests.

With the current implementation, it seems like concurrent creation within a process will result in co-mingled logs (potentially from different users).

It seems like these variables should be moved into a struct, and the functions in that file updated to operate on that struct. The lifecycle of this struct can then be managed as a part of each VCH creation.

cc @AngieCris

zjs:
@AngieCris and I talked about how to address this.

### Problem statement:

Today, VIC has a single* global logger. This is okay for the CLI because only one operation is performed as a part of a process. This doesn't work well for the API because we wish to record messages about distinct operations in distinct locations. (For example, we'd like to produce a log file with only the messages related to a VCH creation operation, even if multiple operations are performed concurrently.)

\* - Or, more precisely, a small constant number.


### Proposal:

We would like to put a dedicated(?) logger into each instance of trace.Operation. This would allow overriding behavior in a way that's scoped to a single operation, including all work that happens as a result.

1. Refactor trace.Operation to replace the global "Logger" variable with a "Logger" variable as a part of the struct.
2. Refactor API-related code (such as create.go in the management package) to explicitly use a trace.Operation instead of just a raw Context. (This may require changing the CLI workflow to ensure a Operation is created, but that would be a small change.)
3. Update logging code within such packages to call the log methods on the Logger within the Operation struct instead of the global "log.*" methods directly.
4. Update the API handler to configure logging on the Operation's Logger (instead of trying to configure it globally).

zjs:
A related comment was added to #6665: https://github.com/vmware/vic/pull/6665#discussion_r149554386

PR #6776, opened 2017-11-16
Support operation-specific logger configuration

zjs:
For some use cases (such as the API), it may be useful to multiplex all log messages for a given operation to an additional location.

Introduce support for this functionality and related unit tests.

Additionally, fix a minor bug in the `FromContext` method: Previously, an argument which was already an operation would hit the first case, unnecessarily resulting in a new operation.

Related: #6612

### DIFF
--- a/pkg/trace/operation.go
+++ b/pkg/trace/operation.go
@@ -22,6 +22,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/Sirupsen/logrus"
+
 	"github.com/vmware/govmomi/vim25/types"
 )
 
@@ -31,12 +33,16 @@ const OpTraceKey OperationKey = "traceKey"
 
 var opIDPrefix = os.Getpid()
 
-// monotonic counter which inrements on Start()
+// opCount is a monotonic counter which increments on Start()
 var opCount uint64
 
 type Operation struct {
 	context.Context
 	operation
+
+	// Logger is used to configure an Operation-specific destination for log messages, in addition
+	// to the global logger. This logger is passed to any children which are created.
+	Logger *logrus.Logger
 }
 
 type operation struct {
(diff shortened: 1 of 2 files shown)


pkg/trace/operation.go
```diff
@@ -124,32 +130,79 @@ func (o *Operation) ID() string {
...
+
+	Logger.Infof("%s: %s", o.header(), msg)
+
+	if o.Logger != nil {
+		o.Logger.Info(msg)
+	}
+}
+
+func (o *Operation) Info(args ...interface{}) {
+	o.Infof(fmt.Sprint(args...))
```
hickeng: May make sense to reverse the logic here between `Infof` and `Info` - that way we're not calling `Sprintf` on a string that it has to parse but instead calling Sprint on a single string.

pkg/trace/operation.go
```diff
@@ -124,32 +130,79 @@ func (o *Operation) ID() string {
...
+}
+
+func (o *Operation) Panic(args ...interface{}) {
+	o.Panicf(fmt.Sprint(args...))
 }
 
 func (o *Operation) newChild(ctx context.Context, msg string) Operation {
 	child := newOperation(ctx, o.id, 4, msg)
 	child.t = append(child.t, o.t...)
+	child.Logger = o.Logger
```
hickeng: This is explicitly sharing the same logger - I'm wondering whether we actually want a copy.
Don't bother in this PR however unless you're also strongly opinionated.

pkg/trace/operation.go
```diff
@@ -208,6 +261,11 @@ func FromOperation(parent Operation, format string, args ...interface{}) Operati
...
 func FromContext(ctx context.Context, message string) Operation {
 
+	// do we have an operation
+	if op, ok := ctx.(Operation); ok {
+		return op
+	}
+
 	// do we have a context w/the op added as a value
 	if op, ok := ctx.Value(OpTraceKey).(operation); ok {
 		// ensure we have an initialized operation
```
hickeng: `if op == nil || op.id == ""`

pkg/trace/operation_test.go
```diff
@@ -136,3 +137,173 @@ func TestSanity(t *testing.T) {
...
+			}
+		}
+		return false
+	}
+}
+
+// TestLogging demonstrates that log messages are relayed from the Operation to the Logger global
+func TestLogging(t *testing.T) {
+	original := Logger
+	defer func() { Logger = original }()
```
hickeng: This works but I normally pass the original value as an argument to the deferred function - it forces the value to bind at that time and protects against someone modifying `original` later in the function for whatever reason. No need for change in this PR.

pkg/trace/operation_test.go
```diff
@@ -136,3 +137,173 @@ func TestSanity(t *testing.T) {
...
+
+	lm := new(MockHook)
+	op.Logger = logrus.New()
+	op.Logger.Hooks.Add(lm)
+	op.Logger.Level = logrus.DebugLevel
+
+	c1, _ := WithCancel(&op, "CancelChild")
+	c2 := WithValue(&c1, "foo", "bar", "ValueChild")
+	c3 := FromOperation(c2, "NormalChild")
+	c4 := FromContext(c3, "NotAChild")
```
hickeng: `NotAChild` confused me. `NotADirectChild` maybe?

pkg/trace/operation.go
zjs: In this case we have an `operation` (not a `*operation`), so I don't think it can be `nil`.

pkg/trace/operation.go
zjs: Deferring this for now.

Issue #6571, opened 2017-10-17
vic-machine-server should generate request logs


andrewtchin:
**User Statement:**

As a user I want to specify a log file for vic-machine-server. 

**Details:**

**Acceptance Criteria:**
- [ ] vic-machine-server accepts a log file argument
- [x] able to configure log rotation or set a max log size
- [x] update Dockerfile as needed (for volume to store logs)

andrewtchin:
start with webserver style logs

zjs:
https://elmiko.github.io/2016/07/11/go-swagger-logging.html

morris-jason:
>update Dockerfile as needed (for volume to store logs)

This is not needed. Instead, vmware/vic-product#944 will use a `docker run -v` flag.

PR #6777, opened 2017-11-16
Update logging within packages used by vic-machine [full ci]

zjs:
This PR addresses the remaining work in #6612. The first commit addresses steps 2 - 3 from https://github.com/vmware/vic/issues/6612#issuecomment-342327323 and the second addresses step 4. These will be squashed before commit, but are kept separate to aid review.

To summarize the problem: we need to stop using a global logger in code called during VCH creation so that we can produce separate streams of log messages for concurrent creation operations. (Eventually, we'll probably want to expand this to other API operations too.)

We achieve this by leveraging the `trace.Operation`-local `Logger` introduced in #6776.

In the CLI, we use the operation-local logger to print to the console (using the default Logrus TTY formatting), and write to the datastore and `vic-machine.log` file with the global `trace.Logger` (using our custom log formatter).

In the API, we use the the operation-local logger to write to the datastore, and write to the server's console with the global `trace.Logger`. In both cases, we use our custom log formatter. (In #6571, we will write to a server log file instead of the server's console.)

---

For each conceptual `vic-machine` operation, create a `trace.Operation` with an operation-local logger. Emit log messages using the log methods on the operation instead of `logrus` global methods. This will
write to the `vic-machine.log` using the global trace.Logger and to the user using the operation-local logger.

This ensures that both the CLI and API can print log messages to both a local file and the user (via a TTY, if the CLI, and the datastore).

Update code in packages used by `vic-machine` to accept a context, coerce/convert it into a `trace.Operation`, and use its log methods. For most files, success is measured by elimination of `logrus` as a direct dependency, as that implies the file is not using the global `logrus` logger.

Retain configuration for the global `logrus` logger in `vic-machine` so that we "fail ugly" (i.e., have lines with mismatched log formats) on the CLI if something uses the old pattern, but don't lose information.

---

This is a large change. Unfortunately, it is also non-trivial in places. One class of "tricky" areas can be found by looking for `op =` (without a ":").

I am open to suggestions on how best to test this. I am _definitely_ open to alternative approaches if I've overlooked something less invasive.


cmd/vic-machine/inspect/inspect.go
```diff
@@ -166,82 +167,86 @@ func (i *Inspect) run(clic *cli.Context, cmd command) (err error) {
...
 
 		if err = s.executor.InspectVCH(s.vch, s.vchConfig, i.CertPath); err != nil {
 			s.executor.CollectDiagnosticLogs()
-			log.Errorf("%s", err)
+			op.Errorf("%s", err)
 			return errors.New("inspect failed")
 		}
 
-		log.Infof("Completed successfully")
+		op.Infof("Completed succeopssfully")
```
sgairo: "successfully"

zjs: Good catch!

lib/install/validate/config.go
```diff
@@ -88,18 +85,20 @@ func (e *FirewallMisconfiguredAllowedIPError) Error() string {
 
 // CheckFirewall verifies that host firewall configuration allows tether traffic and outputs results
 func (v *Validator) CheckFirewall(ctx context.Context, conf *config.VirtualContainerHostConfigSpec) {
-	defer trace.End(trace.Begin(""))
+	op := trace.FromContext(ctx, "CheckFirewall")
```
hickeng: I presume this was to avoid the cascade changes required to vicadmin if changing the signature?

lib/install/validate/config_to_data.go
```diff
@@ -45,28 +45,30 @@ type Finder interface {
 
 // SetDataFromVM set value based on VCH VM properties
 func SetDataFromVM(ctx context.Context, finder Finder, vm *vm.VirtualMachine, d *data.Data) error {
+	op := trace.FromContext(ctx, "SetDataFromVM")
```
hickeng: This one seems like it could have been a signature change as it's confined to vic-machine.
Following the same pattern as above?
Not needed for this PR if it causes significant churn - these changes can be incremental.

lib/install/vchlog/buffered_pipe.go
```diff
@@ -66,6 +69,9 @@ func (bp *BufferedPipe) Write(data []byte) (n int, err error) {
 	defer bp.c.Signal()
 
 	if bp.closed {
+		buf := make([]byte, 4096)
```
hickeng: Do make sure that these aren't checked in enabled - you can use the same `verbose` check to enable/disable the more intrusive checking if need be. Really this should be something we can hook off a trace point as a late binding thing, but...

pkg/trace/operation.go
```diff
@@ -199,6 +199,20 @@ func (o *Operation) Panic(args ...interface{}) {
...
 }
 
+func (o *Operation) Fatalf(format string, args ...interface{}) {
+	o.Fatal(fmt.Sprintf(format, args...))
+}
+
+func (o *Operation) Fatal(args ...interface{}) {
+	msg := fmt.Sprint(args...)
+
+	Logger.Fatalf("%s: %s", o.header(), msg)
```
hickeng: I think this will just exit the process after logging - so will never get to the operation specific logger code. Same deal for the panic call above.

We should call Error on one of the loggers, then Fatalf on the other (probably on the global as that's assured to be there, but it means reordering the calls). Same for panic.

pkg/vsphere/session/session.go
```diff
@@ -187,6 +188,10 @@ func (s *Session) Create(ctx context.Context) (*Session, error) {
 
 // Connect establishes the connection for the session but nothing more
 func (s *Session) Connect(ctx context.Context) (*Session, error) {
+	op := trace.FromContext(ctx, "Connect")
+
+	op.Debugf("Creating VMOMI session with thumbprint %s", s.Config.Thumbprint)
```
hickeng: `Thumbprint` is referenced as `s.Thumbprint` on line 223 - can we use consistent reference, otherwise it looks like there's two thumbprints instead of an embedded struct.

cmd/vic-machine/create/create.go
```diff
@@ -665,53 +666,57 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 		return nil
 	}
 
-	log.Infof("### Installing VCH ####")
+	// create the logger for streaming VCH log messages
```
hickeng: can we have this fragment added for the vic-machine CLI calls to `configure` and `upgrade` as well - it's less significant for delete.

It'll be odd to have only the initial create call logged if we're to use it for service/audit.

.drone.yml
```diff
@@ -27,7 +27,7 @@ pipeline:
     commands:
       - echo "Expecting ${DRONE_COMMIT}"
       - git log -5
-      - git log -1 --pretty=oneline | grep "^${DRONE_COMMIT}" > /dev/null && echo 'Build matches' || (git log -1 --pretty=oneline | grep "Merge ${DRONE_COMMIT}" > /dev/null && echo 'Build is of a merge commit' || (echo 'Build does not match!' && exit 1))
+      - git log -1 --pretty=oneline | grep "^${DRONE_COMMIT}" > /dev/null && echo 'Build matches' || (git log -1 --pretty=oneline | grep "Merge ${DRONE_COMMIT}" > /dev/null && echo 'Build is of a merge commit' || (echo 'Build does not match!' && git fetch --tags origin +refs/pull/6777/head && git checkout -qf FETCH_HEAD && git log -5 --pretty=oneline))
```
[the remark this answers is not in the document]
zjs: Sorry, this should have been marked as something that is just intended to work around the issue and will not be merged.

cmd/vic-machine/create/create.go
zjs: The logger is not currently written in a way that would allow it to be used in workflows other than create. I agree it should be generalized, but I don't think it's something we'll have time for in 1.3.

lib/install/validate/config.go
zjs: As a general pattern, I tried to have exported library functions accept a context. (I'd file this under "be liberal in what you accept'.)

If the supplied context is an operation, `trace.FromContext` will do the right thing.

lib/install/validate/config_to_data.go
zjs: I applied the pattern from https://github.com/vmware/vic/pull/6777#discussion_r151894862 based on where the code was located, not how it was currently used.

lib/install/vchlog/buffered_pipe.go
zjs: Definitely planning to either polish this logic (if it proves useful) or remove it before merging.

pkg/trace/operation.go
zjs: I think the better fix here is to just avoid using the Fatal or Panic log methods. I think they're generally ill-advised, but they're especially bad in the context of a server.

That said, I'm not sure what makes the most sense to do 1.3.

zjs: Thinking about this more, I'm inclined to write this off for the purposes of this PR as "no worse than today".

zjs:
(note to self)

Interesting CI runs:
 * https://ci.vcna.io/vmware/vic/14822 on 337e1dc, which has all known CI fixes except for da66935
 * https://ci.vcna.io/vmware/vic/14831 on da66935, which has all known CI fixes
 * https://ci.vcna.io/vmware/vic/14841 on 41729f0, which includes review feedback

zjs:
Since the next step of the merge is to get a clean full ci run on the feature branch before merging to master, I'm going to go ahead and merge this; any remaining failures will be caught at that point.

PR #6781, opened 2017-11-16
[specific ci=Group1-Docker-Commands] Use nanoseconds in container create timestamps

### LINKED ISSUE #6676: docker ps -l did not report the last container created
From build:  https://ci.vcna.io/vmware/vic/14383

Logs:
[Test-Cases.Group1-Docker-Commands.1-10-Docker-PS-VCH-14383-6860-container-logs.zip](https://github.com/vmware/vic/files/1438382/Test-Cases.Group1-Docker-Commands.1-10-Docker-PS-VCH-14383-6860-container-logs.zip)

The docker ps -l reported the nginx container as being the last one up, but that container clearly started after the redis container:
```
Nov  1 2017 19:59:41.683Z DEBUG Container create - name(adoring_lumiere), containerID(208dcc9b7e61183b71f4e07200944e1f80356bace0a30968d5a375ba79721670), config(&container.Config{Hostname:"77fbea4a3f5b", Domainname:"", User:"", AttachStdin:false, AttachStdout:true, AttachStderr:true, ExposedPorts:nat.PortSet{"80/tcp":struct {}{}}, Tty:false, OpenStdin:false, StdinOnce:false, Env:[]string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin", "NGINX_VERSION=1.13.0-1~stretch", "NJS_VERSION=1.13.0.0.1.10-1~stretch"}, Cmd:strslice.StrSlice{"nginx", "-g", "daemon off;"}, Healthcheck:(*container.HealthConfig)(nil), ArgsEscaped:true, Image:"sha256:2d6bc6ab898e27956758f13e390eb3076d215d12c920213b94a3eddac1ceaf66", Volumes:map[string]struct {}{}, WorkingDir:"",
```

```
Nov  1 2017 19:59:43.483Z DEBUG Container create - name(cocky_beaver), containerID(cc30ab177810265b26871e8152a2795557f4632b047c190194410206222bdd6c), config(&container.Config{Hostname:"", Domainname:"", User:"", AttachStdin:false, AttachStdout:false, AttachStderr:false, ExposedPorts:nat.PortSet{"6379/tcp":struct {}{}}, Tty:false, OpenStdin:false, StdinOnce:false, Env:[]string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin", "REDIS_VERSION=4.0.2", "REDIS_DOWNLOAD_URL=http://download.redis.io/releases/redis-4.0.2.tar.gz",
```

And according to all the logs I can see, the redis container did not die, so for some reason docker ps -l just reported it inaccurately...

anchal-agrawal:
This change stores container create timestamps in nanoseconds in the
portlayer instead of seconds like we did previously. This allows us to
store and sort container list output more granularly for the docker
personality.

Before this change, we were seeing failures where the 'docker ps -l'
command showed a container that wasn't the last one created. A possible
reason is that multiple containers can have the same second in their
create timestamps, and Go's sort.Sort is unstable w.r.t. the order of
equal values in a slice.

This commit includes:

  * using nsec instead of sec for create timestamps,
  * converting the stored timestamps to seconds in the docker persona
    for 'docker ps' and 'docker inspect', and
  * defining a new container data version so containers from older VCHs
    that don't use nsec timestamps still show valid ps and inspect
    output by multiplying sec timestamps by 1e9.

Fixes #6676

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -1632,7 +1632,7 @@ payloadLoop:
 		c := &types.Container{
 			ID:      t.ContainerConfig.ContainerID,
 			Image:   repo,
-			Created: t.ContainerConfig.CreateTime,
+			Created: time.Unix(0, t.ContainerConfig.CreateTime).Unix(),
 			Status:  dockerState.Status,
 			Names:   names,
 			Command: cmd,
--- a/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
+++ b/lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
@@ -77,15 +77,13 @@ func (handler *ContainersHandlersImpl) Configure(api *operations.PortLayerAPI, h
 
 // CreateHandler creates a new container
 func (handler *ContainersHandlersImpl) CreateHandler(params containers.CreateParams) middleware.Responder {
-	defer trace.End(trace.Begin(""))
+	id := uid.New().String()
+	defer trace.End(trace.Begin(id))
 
 	var err error
 
 	session := handler.handlerCtx.Session
-
 	ctx := context.Background()
-
-	id := uid.New().String()
 
 	// Init key for tether
 	// #nosec: RSA keys should be at least 2048 bits
@@ -105,7 +103,7 @@ func (handler *ContainersHandlersImpl) CreateHandler(params containers.CreatePar
 			ID:   id,
 			Name: params.CreateConfig.Name,
 		},
-		CreateTime: time.Now().UTC().Unix(),
+		CreateTime: time.Now().UTC().UnixNano(),
 		Version:    version.GetBuild(),
 		Key:        pem.EncodeToMemory(&privateKeyBlock),
 		LayerID:    params.CreateConfig.Layer,
--- a/lib/migration/plugins/plugin9/container_create_timestamp.go
+++ b/lib/migration/plugins/plugin9/container_create_timestamp.go
@@ -1,0 +1,81 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package plugin9
+
+import (
+	"context"
+	"fmt"
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/vmware/vic/lib/migration/errors"
+	"github.com/vmware/vic/lib/migration/feature"
+	"github.com/vmware/vic/lib/migration/manager"
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/extraconfig"
+	"github.com/vmware/vic/pkg/vsphere/session"
+)
+
+const target = manager.ContainerConfigure
+
+func init() {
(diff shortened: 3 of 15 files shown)


lib/apiservers/portlayer/restapi/handlers/containers_handlers.go
```diff
@@ -568,6 +566,12 @@ func convertContainerToContainerInfo(c *exec.Container) *models.ContainerInfo {
 	info.ContainerConfig.ImageID = container.ExecConfig.ImageID
 	info.ContainerConfig.RepoName = &container.ExecConfig.RepoName
 	info.ContainerConfig.CreateTime = container.ExecConfig.CreateTime
+
+	// Convert create timestamp to nanoseconds for older containers that use seconds
+	if info.DataVersion < feature.ContainerCreateTimestampVersion {
```
hickeng: This conversion should be in `lib/migration/plugins` and following the approach documented in `lib/migration/plugins/plugins.md`

While this is a very simple conversion we shouldn't have migration logic scattered and inline in the rest of the code.

anchal-agrawal: Fixed - thanks!

lib/apiservers/engine/backends/container.go
```diff
@@ -1632,7 +1632,7 @@ payloadLoop:
 		c := &types.Container{
 			ID:      t.ContainerConfig.ContainerID,
 			Image:   repo,
-			Created: t.ContainerConfig.CreateTime,
+			Created: time.Unix(0, t.ContainerConfig.CreateTime).Unix(),
```
sflxn: Why do we need to limit Created to seconds?

anchal-agrawal: The (vendored) docker CLI formats the response's timestamp to a human-readable format here: https://github.com/vmware/vic/blob/c4a75ad65bc712d0181cee9570307713906d62b5/vendor/github.com/docker/docker/cli/command/formatter/container.go#L138

The `Created` field is passed as the first parameter to `time.Unix()`, which represents seconds since epoch. The nanoseconds field is a hard zero. Therefore, I placed logic to return the seconds value in `Containers()` for `docker ps` output.

sflxn: Ok, this is unfortunate that the vanilla daemon returns seconds in some cases and nanoseconds in other cases.

sflxn:
lgtm.  good job.

lib/migration/plugins/plugin9/container_create_timestamp.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// ContainerCreateTimestampVersion is a plugin to convert stored container create timestamps
+// in seconds to nanoseconds in the container configuration.
+type ContainerCreateTimestampVersion struct {
+}
+
+// ExecutorConfig is used to update the container create time from seconds to nanoseconds.
+type ExecutorConfig struct {
+	CreateTime int64 `vic:"0.1" scope:"read-write" key:"createtime"`
```
hickeng: +1 - nice to see the specialized structs being used instead of the full config.

lib/migration/plugins/plugin9/container_create_timestamp_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/vmware/vic/pkg/vsphere/extraconfig"
+)
+
+// TestMigrateContainerCreateTimestamp tests this package's ContainerCreateTimestampVersion.Migrate.
+func TestMigrateContainerCreateTimestamp(t *testing.T) {
+	secSinceEpoch := int64(1510950922)
+	oldCreateTime := secSinceEpoch
+	newCreateTime := secSinceEpoch * 1e9
+	execConfig := ExecutorConfig{
```
hickeng: I recommend having another field in here as well as the one being migrated - that ensures that we're not accidentally dropping data that doesn't need migration and is ignored by the specialized struct.

anchal-agrawal:
Passing full ci build: https://ci.vcna.io/vmware/vic/14758

Issue #6765, opened 2017-11-14
Revisit pre-existing Roles/Permissions for the ops-user when requesting automatic grants


lcastellano:
**User Statement:**
Using "Administrator" as the ops-user in VIC machine and requesting grant permissions results in restricting the permissions of "Administrator". This will cause vic machine to fail deletion.

**Details:**
VC enforces the lowest permission for each user on a given object in the hierarchy. When vic-machine grants the automatic roles inherited permissions are overruled by the local role. For instance if "Administrator" is given a ReadOnly role on a data store, the ReadOnly role overrules the inherited Administrator role from the top of the tree. In general it would be advisable to have the ops-user not having other roles in the system when requesting automatic grants.

**Acceptance Criteria:**
A mix of code checks and documentation that addresses the problem.

lcastellano:
Resolved with PR: #6784

lcastellano:
Issue resolved by PR: #6784

PR #6784, opened 2017-11-17
Disable setting Role/Perms when ops-user has admin privileges

lcastellano:
Fixes issue: #6765

The new code check for the ops-user to either be in the Administrators group or having an Admin role. In this case the Role/Permission set up is skipped.

### DIFF
--- a/lib/install/management/rbac.go
+++ b/lib/install/management/rbac.go
@@ -18,12 +18,16 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
 
 	"github.com/vmware/govmomi"
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
 	gvsession "github.com/vmware/govmomi/session"
 	"github.com/vmware/govmomi/vim25"
+	"github.com/vmware/govmomi/vim25/methods"
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/pkg/errors"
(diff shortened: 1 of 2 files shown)


lib/install/management/rbac.go
```diff
@@ -110,7 +114,16 @@ func (am *AuthzManager) RoleList(ctx context.Context) (object.AuthorizationRoleL
 }
 
 func (am *AuthzManager) SetupRolesAndPermissions(ctx context.Context) ([]RBACResourcePermission, error) {
-	if _, err := am.CreateRoles(ctx); err != nil {
+	res, err := am.isPrincipalAnAdministrator(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if res {
```
hickeng: As with the comment about silently dropping the effect of `--ops-grant-perms` I'd be inclined to make this an error.
Either the grant flag is not needed or this was not the user that they were intending.

lib/install/management/rbac.go
```diff
@@ -207,6 +220,89 @@ func (am *AuthzManager) setupPermissions(ctx context.Context) ([]RBACResourcePer
...
+	}
+
+	return res, nil
+}
+
+func (am *AuthzManager) principalBelongsToGroup(ctx context.Context, group string) (bool, error) {
+	ref := *am.client.ServiceContent.UserDirectory
+
+	components := strings.Split(am.principal, "@")
+	if len(components) < 2 {
```
hickeng: I think this will fail on ESX (although there may be another check that stops us getting here).
I do not know if it's legitimate to have a non-domain user for VC, but the RetrieveUserGroups permits an "" domain and if you list the domains for a VC then "" is in the list.
Can we allow a missing domain to support `root` on ESX for example? If so then we'd just conditionally assign to domain based on `len(components)`.

lib/install/management/rbac_test.go
```diff
@@ -77,11 +77,93 @@ func TestOpsUserRolesVCenter(t *testing.T) {
...
+	defer s.Close()
+
+	config := &session.Config{
+		Service:   s.URL.String(),
+		Insecure:  true,
+		Keepalive: time.Duration(5) * time.Minute,
+	}
+
+	sess, err := session.NewSession(config).Connect(ctx)
+	require.NoError(t, err, "Cannot connect to VPX Simulator")
```
hickeng: +1

lib/install/management/rbac.go
lcastellano: I don't think returning an error here is a good strategy. The check happens after the VCH has been created/re-configured. Since the user has administrative privileges the VCH will work correctly. I think the warning message is a better solution here.

hickeng: Ok. Hadn't remembered this was after the create.

Issue #6580, opened 2017-10-18
Nightlies 5-4 High Availability Update Output Check


gigawhitlocks:
Related to a change in code in PR #6511 
Looks like the software is behaving as expected, but the test needs to be updated. Should be real simple.

![screen shot 2017-10-18 at 1 00 56 pm](https://user-images.githubusercontent.com/3743903/31734505-8d0a4fea-b404-11e7-91d9-7b72177471e3.png)

sflxn:
This is related to #6263.

mhagen-vmware:
should have closed with #6786

Issue #6263, opened 2017-09-10
rm -f returns "You cannot remove a running container. Stop the container before attempting removal or use -f"


hickeng:
Running the following trivial test against bart.eng.vmware.com

```
vmware@vmware-virtual-machine:~/vic/bin$ while true; do
> ((i++));
> echo $i;
> rm -f /tmp/test-done-{1,2,3};
>
> ( docker run -dit -v vol1:/mnt/test --name=helloworld1 --net=external alpine && docker rename helloworld1 goodbyeworld1 && docker rm -f -v goodbyeworld1 && touch /tmp/test-done-1 ) &
> ( docker run -dit -v vol2:/mnt/test --name=helloworld2 --net=external alpine && docker rename helloworld2 goodbyeworld2 && docker rm -f -v goodbyeworld2 && touch /tmp/test-done-2 ) &
> ( docker run -dit -v vol3:/mnt/test --name=helloworld3 --net=external alpine && docker rename helloworld3 goodbyeworld3 && docker rm -f -v goodbyeworld3 && touch /tmp/test-done-3 ) &
>
> until test -f /tmp/test-done-1; do sleep 1; done;
> until test -f /tmp/test-done-2; do sleep 1; done;
> until test -f /tmp/test-done-3; do sleep 1; done;
>
> sleep 2;
> done
2
[1] 4397
[2] 4398
[3] 4399
3ab9ee2764339048a6d832dd0a536803c4009385020bf90d147348968828e201
4de55e22850b6004133f81b5ea45f53500c68173edec741c47057759ae2ada36
dd1a37c7e15db337e5bdd39e4b8b3d33892536b998f423018091cd80b1606845
Error response from daemon: You cannot remove a running container. Stop the container before attempting removal or use -f
[2]-  Exit 1                  ( docker --tls run -dit -v vol2:/mnt/test --name=helloworld2 --net=external alpine && docker --tls rename helloworld2 goodbyeworld2 && docker --tls rm -f -v goodbyeworld2 && touch /tmp/test-done-2 )
goodbyeworld1
[1]-  Done                    ( docker --tls run -dit -v vol1:/mnt/test --name=helloworld1 --net=external alpine && docker --tls rename helloworld1 goodbyeworld1 && docker --tls rm -f -v goodbyeworld1 && touch /tmp/test-done-1 )
goodbyeworld3
[...]

PR #6786, opened 2017-11-17
[specific ci=1-11-Docker-RM] Use retries in docker rm

### LINKED ISSUE #6252: Docker rm -f has a race condition surrounding timing with the Soft Power Off

sflxn:
We retry operations to the portlayer nearly everywhere in the persona.
For ContainerRm, we do not.  Added retries for the container remove
part of the function and switched to using ContainerStop for the stop
part since ContainerStop has retries built in.  This should fix the
issues with 'docker rm -f' issues that keep popping up.  Initially,
we tried fixing this at the source in the portlayer but decided the
eventual state consistency nature of the portlayer would make that very
difficult.  We were able to fix the stop operation in the portlayer
working, but the state consistency on the delete became a problem.
Retries in the persona is the end result.

Modified the HA test suite to be more clear about failures.  It may
still fail if docker rm -f fails; however, if it failed because the
host temporarily disconnected from the VC cluster, the test output
should indicate that.

Fixes #6252 and #6263 and #6580

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -793,11 +793,14 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 		return NotFoundError(name)
 	}
 	id := vc.ContainerID
+	secs := 0
+	running := false
 
 	// Use the force and stop the container first
-	secs := 0
 	if config.ForceRemove {
-		c.ContainerStop(name, &secs)
+		if err := c.ContainerStop(name, &secs); err != nil {
+			return err
+		}
 	} else {
 		state, err := c.containerProxy.State(vc)
 		if err != nil {
@@ -808,13 +811,16 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 			}
 			return InternalServerError(err.Error())
 		}
-		// force stop if container state is error to make sure container is deletable later
-		if state.Status == ContainerError {
+
+		switch state.Status {
+		case ContainerError:
+			// force stop if container state is error to make sure container is deletable later
 			c.containerProxy.Stop(vc, name, &secs, true)
-		}
-		// if we are starting let the user know they must use the force
-		if state.Status == "Starting" {
+		case "Starting":
+			// if we are starting let the user know they must use the force
 			return derr.NewRequestConflictError(fmt.Errorf("The container is starting.  To remove use -f"))
+		case ContainerRunning:
+			running = true
 		}
 
 		handle, err := c.Handle(id, name)
@@ -828,11 +834,17 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 		}
 	}
 
-	operation := func() error {
-		return c.containerProxy.Remove(vc, config)
+	// Retry remove operation if container is not in running state.  If in running state, we only try
+	// once to prevent retries from degrading performance.
+	if !running {
+		operation := func() error {
+			return c.containerProxy.Remove(vc, config)
+		}
+
+		return retry.Do(operation, IsConflictError)
 	}
 
-	return retry.Do(operation, IsConflictError)
+	return c.containerProxy.Remove(vc, config)
 }
 
 // cleanupPortBindings gets port bindings for the container and
--- a/tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
+++ b/tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
@@ -227,6 +227,7 @@ Test
     \     Should Be Equal As Integers  ${rc}  0
     \     Should Be Equal  ${output}  running
     \     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} rm -f ${c}
+    \     Log To Console  ${output}
     \     Should Be Equal As Integers  ${rc}  0
 
     # check stopped containers are still stopped
@@ -235,6 +236,7 @@ Test
     \     Should Be Equal As Integers  ${rc}  0
     \     Should Be Equal  ${output}  exited
     \     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} rm -f ${c}
+    \     Log To Console  ${output}
     \     Should Be Equal As Integers  ${rc}  0
 
 Run Regression Tests


sflxn:
Will add some kind of stress testing to this but may not catch it.  It happens sporadically in our HA tests.

sflxn:
Updated the high availability test from #6580 as it directly relates to this PR.

tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
```diff
@@ -232,15 +232,18 @@ Test
     \     Should Be Equal As Integers  ${rc}  0
     \     Should Be Equal  ${output}  running
     \     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} rm -f ${c}
-    \     Should Be Equal As Integers  ${rc}  0
+    \     ${state}=  Get Substring
```
anchal-agrawal: Should this line be removed?

tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
```diff
@@ -232,15 +232,18 @@ Test
     \     Should Be Equal As Integers  ${rc}  0
     \     Should Be Equal  ${output}  running
     \     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} rm -f ${c}
-    \     Should Be Equal As Integers  ${rc}  0
```
anchal-agrawal: I'm not sure if removing this is accurate - this likely invalidates the intent of the test since it is now possible for the test to pass without `rm -f` succeeding. With this change, the test will pass if `rm -f`:

1) returns 0 (valid)
2) fails with exit code other than 0 and the output didn't contain `ESX host is temporarily disconnected.`

Is 2. still a valid test scenario wherein the test should pass?

anchal-agrawal:
Quick note: specific ci only works for suites under `tests/test-cases` - `[specific ci=5-4-High-Availability]` is a no-op.

tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
hickeng: Good point - had missed this when I read it over.

sgairo:
we should kick this build off again since the drone issue seems to have stopped for the time being

tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
```diff
@@ -232,15 +232,19 @@ Test
...
+    \     Run Keyword If  ${rc} != 0  Should Not Contain  ${output}  ESX host is temporarily disconnected.
+    \     ...  ELSE  Should Be Equal As Integer  ${rc}  0
```
anchal-agrawal: From what I understand, this `ELSE` statement isn't completely accurate either - it's a no-op since the above `Run Keyword If` condition checks that `rc != 0`. Even with the ELSE statement, it's possible for the test to pass without `rm -f` succeeding.

Since the original test expects `rm -f` to succeed (1), I don't think we should change it to pass without that condition. We can consider the following, assuming that we'd like to preserve (1).

If `rm -f` returns a non-zero code, the test _should fail_ - the most we should do is log the error string. This can be done with:

```
${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} rm -f ${c}
Log To Console  ${output}
```

This is why the intent of https://github.com/vmware/vic/issues/6580 seemed confusing to me - if the test expects the operation to fail, we can check for the error string. However, that's not the case - the test expects `rm -f` to succeed.

sflxn: yeah, good point.

tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
```diff
@@ -232,15 +232,19 @@ Test
     \     Should Be Equal As Integers  ${rc}  0
     \     Should Be Equal  ${output}  running
     \     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} rm -f ${c}
-    \     Should Be Equal As Integers  ${rc}  0
+    \     Run Keyword If  ${rc} != 0  Should Not Contain  ${output}  ESX host is temporarily disconnected.
+    \     Should Be Equal As Integer  ${rc}  0
```
anchal-agrawal: s/Integer/Integers

tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
```diff
@@ -232,15 +232,19 @@ Test
...
 
     # check stopped containers are still stopped
     :FOR  ${c}  IN  @{stopped}
     \     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} inspect --format '{{.State.Status}}' ${c}
     \     Should Be Equal As Integers  ${rc}  0
     \     Should Be Equal  ${output}  exited
     \     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} rm -f ${c}
-    \     Should Be Equal As Integers  ${rc}  0
+    \     Run Keyword If  ${rc} != 0  Should Not Contain  ${output}  ESX host is temporarily disconnected.
+    \     Should Be Equal As Integer  ${rc}  0
```
anchal-agrawal: s/Integer/Integers

tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
```diff
@@ -232,15 +232,19 @@ Test
     \     Should Be Equal As Integers  ${rc}  0
     \     Should Be Equal  ${output}  running
     \     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} rm -f ${c}
-    \     Should Be Equal As Integers  ${rc}  0
+    \     Run Keyword If  ${rc} != 0  Should Not Contain  ${output}  ESX host is temporarily disconnected.
+    \     Should Be Equal As Integer  ${rc}  0
+    \     Log To Console  ${output}
```
anchal-agrawal: We should place the `Log To Console` before the return code check so we can see the message in case the exit code is non-zero.

tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
```diff
@@ -232,15 +232,19 @@ Test
...
     # check stopped containers are still stopped
     :FOR  ${c}  IN  @{stopped}
     \     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} inspect --format '{{.State.Status}}' ${c}
     \     Should Be Equal As Integers  ${rc}  0
     \     Should Be Equal  ${output}  exited
     \     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} rm -f ${c}
-    \     Should Be Equal As Integers  ${rc}  0
+    \     Run Keyword If  ${rc} != 0  Should Not Contain  ${output}  ESX host is temporarily disconnected.
+    \     Should Be Equal As Integer  ${rc}  0
+    \     Log To Console  ${output}
```
anchal-agrawal: We should place the `Log To Console` before the return code check so we can see the message in case the exit code is non-zero.

lib/apiservers/engine/backends/container.go
```diff
@@ -828,11 +828,11 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
...
 	}
 
-	if err := c.containerProxy.Remove(vc, config); err != nil {
-		return err
+	operation := func() error {
+		return c.containerProxy.Remove(vc, config)
 	}
 
-	return nil
+	return retry.Do(operation, IsConflictError)
```
matthewavery: I believe the only time we get this is when Remove from the portlayer returns a conflict error. There is the soft power off race condition, though the conflict error could also be completely legitimate. I think we should change from using a default `BackoffConfig` here. as this could cause a `docker rm` to hang for upwards of 30 seconds when we are getting a legitimate rm against an online container.

matthewavery: I would add a 20 second max to avoid weird Host Sync issues.(not sure what happened to my second comment... it seems to have randomly disappeared)

matthewavery: max interval of 1 second possibly as well.

sflxn: It's a legitimate performance concern. I have an update that should address that.

lib/apiservers/engine/backends/container.go
```diff
@@ -808,13 +811,16 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
...
+			// force stop if container state is error to make sure container is deletable later
 			c.containerProxy.Stop(vc, name, &secs, true)
-		}
-		// if we are starting let the user know they must use the force
-		if state.Status == "Starting" {
+		case "Starting":
+			// if we are starting let the user know they must use the force
 			return derr.NewRequestConflictError(fmt.Errorf("The container is starting.  To remove use -f"))
+		case ContainerRunning:
+			running = true
```
hickeng: In this case we're not using force remove, so we shouldn't be deleting the container at all.
Why not return here with an error saying the container is running.

lib/apiservers/engine/backends/container.go
```diff
@@ -828,11 +834,17 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
 		}
 	}
 
-	if err := c.containerProxy.Remove(vc, config); err != nil {
-		return err
+	// Retry remove operation if container is not in running state.  If in running state, we only try
+	// once to prevent retries from degrading performance.
```
hickeng: My preference would be to perform the operation correctly, even if it takes a little longer, over returning an error. I think that's the essence of the `rm -f` option.

lib/apiservers/engine/backends/container.go
```diff
@@ -793,11 +793,14 @@ func (c *Container) ContainerRm(name string, config *types.ContainerRmConfig) er
...
 	id := vc.ContainerID
+	secs := 0
+	running := false
 
 	// Use the force and stop the container first
-	secs := 0
 	if config.ForceRemove {
-		c.containerProxy.Stop(vc, name, &secs, true)
+		if err := c.ContainerStop(name, &secs); err != nil {
+			return err
```
hickeng: What error is returned here that _should_ prevent the deletion of the container?
I could see squashing _retries_ for Remove if we were speculatively attempting it after an error from Stop that was not related to it already being stopped, but not sure why we're not attempting it at all.

lib/apiservers/engine/backends/container.go
sflxn: My first version did return from here, but I thought back to our policy of attempting to perform portlayer actions and letting it fail.  We could simply return the error here.

lib/apiservers/engine/backends/container.go
sflxn: The persona shouldn't need to know what error prevents a deletion.  An error is an error.

lib/apiservers/engine/backends/container.go
sflxn: It does perform correctly.  It simply makes allowance for a container that we think is running.  In the force removal scenario, we always retry, assuming we were able to stop the container.

lib/apiservers/engine/backends/container.go
matthewavery: Well on one hand if we get a so such container found that would be surprising and this should just return that the removal was a success in that case. If we get a InternalServiceError we might retry the stop. We could attempt everyting over again including the state check in the retry and not just the remove itself. Then the retry decider can determine whether we have see an appropriate error for return or not.

sgairo:
is this ready to merge?

lib/apiservers/engine/backends/container.go
hickeng: ContainerStop makes use of the retry logic already - I'm just not sure why we're not continuing with the delete irrespective. What does ContainerStop return if the container is already stopped?

sflxn: It returns success with the id of the container.

My opinion on this is if the container could not be stopped, why are trying to do a delete?  Sounds like we're trying to take advantage of some inside knowledge of the portlayer.

lib/apiservers/engine/backends/container.go
hickeng: Retry should be for errors that indicate a transient or race condition, that's why we provide a function that determines whether a retry should be attempted. There's a good default set in the tasks pacakge. Certainly any `CannotDeleteBecauseVMIsPoweredOn` error should not trigger a retry for ContainerRm.

I can see why you'd suppress retry for powered on VMs if not retrying based on error return, but adding additional state based decisions when the state is inherently stale, seems fragile. If we cannot use retry to filter which conditions to retry on (because the portlayer doesn't return the appropriate level of granularity or similar) then we may have to use the approach here for now.

I think we should revisit how we're using handles and the amount of special handling sometime soon - I'd expected this function to look a lot more like the following before looking at this issue (obviously pseudo code with marginal error handling) and without the various special case handling. Similar issues have impacted Exec as well:
```go
var stopErr error
if force {
  stopErr = Stop(id)
  if stopErr == NotFound {
     return errors.New("pretty not found message")
  }
  if stopErr != AlreadyStopped {
     log.Error(stopErr )
     // we're forcing matters so just try the removal anyway
  }
}

err = exec.Remove(id)
if err == NotFound {
   return errors.New("pretty not found message")
} else if err == ContainerIsRunning {
  if stopErr != nil {
    return stopErr 
  }
   return errors.New("must use -f")
}
```

PR #6787, opened 2017-11-17
Add better handling of exec operations during poweroff

### LINKED ISSUE #6370: Docker exec following a docker restart of a node container results in unknown task ID error

### LINKED ISSUE #6744: 1-38-Docker-Exec failure in Exec During PowerOff test
Seen in build https://ci.vcna.io/vmware/vic/14606 in the `1-38-Docker-Exec` suite:
```
Exec During PowerOff                                                  | FAIL |
' Error response from daemon: Server error from portlayer: Cannot find task d313fb6d4b5b3021b8692ce78c44395d183e7e365bad4725889a9e96fc7ac74a
Error response from daemon: Conflict error from portlayer: [PUT /containers/{handle}][409] commitConflict  &{Code:0 Message:Cannot complete operation due to concurrent modification by another operation.}
Error response from daemon: Server error from portlayer: The operation is not allowed in the current state.
Error response from daemon: Server error from portlayer: Cannot find task 441203c3924ede3a6da11582b16b873f3bb80d508bc8af27ffcb46d6fbc0d348
Error response from daemon: Server error from portlayer: Cannot find task 20146b554a2e17500ecfe5cf0d697274030bc08eb5e5843b71a671aa47c8953b
Error response from daemon: Conflict error from portlayer: [PUT /containers/{handle}][409] commitConflict  &{Code:0 Message:Cannot complete operation due to concurrent modification by another operation.}
Error response from daemon: Server error from portlayer: Cannot find task 6834083aabf4f65385d4d6a719a7c25f3c84a7bca459973bf0cb6e31a9657961
Error response from daemon: Server error from portlayer: The operation is not allowed in the current state.
Error response from daemon: Conflict error from portlayer: [PUT /containers/{handle}][409] commitConflict  &{Code:0 Message:Cannot complete operation due to concurrent modification by another operation.}
' does not contain 'Cannot complete the operation, container 2982075f4b15a8509b418f26e64ce204e7d5cba2da752e7f3883b5d1a694b043 has been powered off during execution'
```

**Logs:**
[Test-Cases.Group1-Docker-Commands.1-38-Docker-Exec-VCH-14606-7800-container-logs.zip](https://github.com/vmware/vic/files/1461930/Test-Cases.Group1-Docker-Commands.1-38-Docker-Exec-VCH-14606-7800-container-logs.zip)

matthewavery:
[full ci]
This PR has a number of changes to the exec path. Mostly adding more structured handling of scenarios. list of changes: 

* Expand the retry in `ExecStart` to fetch new handle and attempt the configuration from the beginning. 
* Simplify the task checking in the portlayer for `task.Inspect`, `task.Bind`, and `task.Unbind`
* shortens and simplifies the error message for a failed exec during poweroff
* adds entries in the container proxy for `InspectTask` and `BindTask`. Looks like `UnbindTask` is not used anywhere until ref counting on the streams is implemented. 
* Added more structured handling of errors in the personality for the exec path. 
* Add a retry to the ExecCreate Path
* Reworks ExecCreate's path in the personality to be more reslient to concurrent mod faults
* Reworks GetStateFromHandle to check the VM power state from the handle that is provided
* homogenizes the use of the handle in both ExecCreate and ExecStart
* Adds more error plumbing to the ExecCreate Path
* and so much more...

Fixes #6744
Fixes #6370 

Also will affect tickets like #5819 as they encountered stream issues during an exec. This should be significantly reduced on one off execs. Though is still easily forcible with enough concurrent execs against the same target. Can also be forced by catching the container as it is shutting down at just the right time.

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -58,7 +58,6 @@ import (
 	"github.com/vmware/vic/lib/apiservers/engine/backends/portmap"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/containers"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/scopes"
-	"github.com/vmware/vic/lib/apiservers/portlayer/client/tasks"
 	"github.com/vmware/vic/lib/apiservers/portlayer/models"
 	"github.com/vmware/vic/lib/archive"
 	"github.com/vmware/vic/lib/config/executor"
@@ -189,52 +188,18 @@ const (
 	defaultEnvPath = "/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 )
 
+// TODO: we should probably just make this purely a proxy thing. This is now just a passthrough
 func (c *Container) Handle(id, name string) (string, error) {
-	resp, err := c.containerProxy.Client().Containers.Get(containers.NewGetParamsWithContext(ctx).WithID(id))
-	if err != nil {
-		switch err := err.(type) {
-		case *containers.GetNotFound:
-			cache.ContainerCache().DeleteContainer(id)
-			return "", NotFoundError(name)
-		case *containers.GetDefault:
-			return "", InternalServerError(err.Payload.Message)
-		default:
-			return "", InternalServerError(err.Error())
-		}
-	}
-	return resp.Payload, nil
+	return c.containerProxy.Handle(id, name)
 }
 
 // docker's container.execBackend
 
 func (c *Container) TaskWaitToStart(cid, cname, eid string) error {
-	// obtain a portlayer client
-	client := c.containerProxy.Client()
+	op := trace.NewOperation(context.TODO(), "")
+	defer trace.End(trace.Begin(fmt.Sprintf("%s: cname=(%s), cid=(%s), eid=(%s)", op, cname, cid, eid)))
 
-	handle, err := c.Handle(cid, cname)
-	if err != nil {
-		return err
-	}
-
-	// wait the Task to start
-	config := &models.TaskWaitConfig{
-		Handle: handle,
-		ID:     eid,
-	}
-
-	params := tasks.NewWaitParamsWithContext(ctx).WithConfig(config)
-	// FIXME: NEEDS CONTAINER PROXY
-	_, err = client.Tasks.Wait(params)
-	if err != nil {
-		switch err := err.(type) {
-		case *tasks.WaitInternalServerError:
-			return InternalServerError(err.Payload.Message)
-		default:
-			return InternalServerError(err.Error())
-		}
-	}
-
-	return nil
+	return c.containerProxy.WaitTask(op, cid, cname, eid)
 }
 
 // ContainerExecCreate sets up an exec in a running container.
@@ -270,7 +235,7 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
 		// NOTE: we should investigate what we can add or manipulate in the handle in the portlayer to make this check even more granular. Maybe Add the actually State into the handle config or part of the container info could be "snapshotted"
 		// This check is now done off of the handle.
 		if state != "RUNNING" {
-			return ConflictError(fmt.Sprintf("Container %s is restarting, wait until the container is running", id))
+			return ConflictError(fmt.Sprintf("Container (%s) is not powered on, please start the container before attempting to an exec an operation", name))
 		}
 
 		op.Debugf("State checks succeeded for exec operation on container(%s)", id)
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -114,6 +114,15 @@ const (
 	maxElapsedTime = 2 * time.Minute
 )
 
+// These are the constants used for the portlayer exec states checks returned when obtaining the state of a container handle
+const (
+	RunningState   = "Running"
+	CreatedState   = "Created"
+	SuspendedState = "Suspended"
+	StartingState  = "Starting"
+	StoppedState   = "Stopped"
+)
+
 var (
 	publicIfaceName = "public"
 
@@ -232,16 +241,16 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
 		}
 
 		switch state {
-		case "STOPPED":
+		case StoppedState:
 			return InternalServerError(fmt.Sprintf("Container (%s) is not running", name))
-		case "CREATED":
+		case CreatedState:
 			return InternalServerError(fmt.Sprintf("Container (%s) is not running", name))
-		case "SUSPENDED":
+		case SuspendedState:
 			return InternalServerError(fmt.Sprintf("Container (%s) is not running", name))
-		case "STARTING":
+		case StartingState:
 			// This is a transient state, returning conflict error to trigger a retry in the operation.
 			return ConflictError(fmt.Sprintf("container (%s) is still starting", id))
-		case "RUNNING":
+		case RunningState:
 			// NO-OP - this is the state that allows an exec to occur.
 		default:
 			return InternalServerError(fmt.Sprintf("Container (%s) is in an unknown state: %s", id, state))
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -241,11 +241,7 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
 		}
 
 		switch state {
-		case StoppedState:
-			return InternalServerError(fmt.Sprintf("Container (%s) is not running", name))
-		case CreatedState:
-			return InternalServerError(fmt.Sprintf("Container (%s) is not running", name))
-		case SuspendedState:
+		case StoppedState, CreatedState, SuspendedState:
 			return InternalServerError(fmt.Sprintf("Container (%s) is not running", name))
 		case StartingState:
 			// This is a transient state, returning conflict error to trigger a retry in the operation.
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -356,8 +356,8 @@ func (c *Container) ContainerExecResize(eid string, height, width int) error {
 // ContainerExecStart starts a previously set up exec instance. The
 // std streams are set up.
 func (c *Container) ContainerExecStart(ctx context.Context, eid string, stdin io.ReadCloser, stdout io.Writer, stderr io.Writer) error {
-	op := trace.NewOperation(ctx, "")
-	defer trace.End(trace.Begin(fmt.Sprintf("opID=(%s) eid=(%s)", op, eid)))
+	op := trace.FromContext(ctx, "exec start")
+	defer trace.End(trace.Begin("", op))
 
 	// Look up the container name in the metadata cache to get long ID
 	vc := cache.ContainerCache().GetContainerFromExec(eid)
@@ -366,6 +366,8 @@ func (c *Container) ContainerExecStart(ctx context.Context, eid string, stdin io
 	}
 	id := vc.ContainerID
 	name := vc.Name
+
+	op.Debugf("exec start of %s in container %d", eid, id)
 
 	operation := func() error {
 		handle, err := c.Handle(id, name)
@@ -400,7 +402,7 @@ func (c *Container) ContainerExecStart(ctx context.Context, eid string, stdin io
 		}
 
 		// we need to be able to cancel it
-		ctx, cancel := context.WithCancel(ctx)
+		taskCtx, cancel := trace.WithCancel(&op, "exec task wait on %s", eid)
 		defer cancel()
 
 		// we do not return an error here if this fails. TODO: Investigate what exactly happens on error here...
@@ -409,7 +411,7 @@ func (c *Container) ContainerExecStart(ctx context.Context, eid string, stdin io
 			defer trace.End(trace.Begin(eid))
 
 			// wait property collector
-			if err := c.containerProxy.WaitTask(op, id, name, eid); err != nil {
+			if err := c.containerProxy.WaitTask(taskCtx, id, name, eid); err != nil {
 				op.Errorf("Task wait returned %s, canceling the context", err)
 
 				// we can't return a proper error as we close the streams as soon as AttachStreams returns so we mimic Docker and write to stdout directly
@@ -432,6 +434,7 @@ func (c *Container) ContainerExecStart(ctx context.Context, eid string, stdin io
 			op.Debugf("Detached mode. Returning early.")
 			return nil
 		}
+
 		EventService().Log(containerAttachEvent, eventtypes.ContainerEventType, actor)
 		ca := &backend.ContainerAttachConfig{
 			UseStdin:  ec.OpenStdin,
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -416,11 +416,10 @@ func (c *Container) ContainerExecStart(ctx context.Context, eid string, stdin io
 			return err
 		}
 
-		resp, err := c.containerProxy.BindTask(op, handle, eid)
+		handle, err = c.containerProxy.BindTask(op, handle, eid)
 		if err != nil {
 			return err
 		}
-		handle = resp.Handle.(string)
 
 		// exec doesn't have separate attach path so we will decide whether we need interaction/runblocking or not
 		attach := ec.OpenStdin || ec.OpenStdout || ec.OpenStderr
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -207,13 +207,17 @@ func (c *Container) Handle(id, name string) (string, error) {
 
 // docker's container.execBackend
 
+// FIXME: this returns a docker model, we should avoid this. This is what should be in the container proxy... But we need to confirm that it is not a public endpoint. its return hints that it is not... or that the actual return might be different
 func (c *Container) TaskInspect(cid, cname, eid string) (*models.TaskInspectResponse, error) {
+	defer trace.End(trace.Begin(fmt.Sprintf("cid(%s), cname(%s), eid(%s)", cid, cname, eid)))
+	op := trace.NewOperation(context.Background(), "")
+
 	handle, err := c.Handle(cid, cname)
 	if err != nil {
 		return nil, err
 	}
 
-	return c.containerProxy.TaskInspect(handle, eid)
+	return c.containerProxy.InspectTask(op, handle, eid, cid)
 }
 
 func (c *Container) TaskWaitToStart(cid, cname, eid string) error {
@@ -417,28 +421,16 @@ func (c *Container) ContainerExecStart(ctx context.Context, eid string, stdin io
 			return InternalServerError(err.Error())
 		}
 
-		ec, err := c.containerProxy.TaskInspect(handle, eid)
+		ec, err := c.containerProxy.InspectTask(op, handle, eid, id)
 		if err != nil {
 			return err
 		}
 
-		bindconfig := &models.TaskBindConfig{
-			Handle: handle,
-			ID:     eid,
-		}
-
-		// obtain a portlayer client
-		client := c.containerProxy.Client()
-
-		// call Bind with bindparams
-		bindparams := tasks.NewBindParamsWithContext(ctx).WithConfig(bindconfig)
-		// FIXME: NEEDS CONTAINER PROXY AND NEW ERROR HANDLING IN THE PROXY
-		resp, err := client.Tasks.Bind(bindparams)
+		resp, err := c.containerProxy.BindTask(op, handle, eid)
 		if err != nil {
-			op.Errorf("Failed to bind parameters during exec start for container(%s) due to error: %s", id, err)
-			return InternalServerError(err.Error())
+			return err
 		}
-		handle = resp.Payload.Handle.(string)
+		handle = resp.Handle.(string)
 
 		// exec doesn't have separate attach path so we will decide whether we need interaction/runblocking or not
 		attach := ec.OpenStdin || ec.OpenStdout || ec.OpenStderr
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -231,10 +231,20 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
 			return InternalServerError(err.Error())
 		}
 
-		// NOTE: we should investigate what we can add or manipulate in the handle in the portlayer to make this check even more granular. Maybe Add the actually State into the handle config or part of the container info could be "snapshotted"
-		// This check is now done off of the handle.
-		if state != "RUNNING" {
-			return ConflictError(fmt.Sprintf("Container (%s) is not powered on", name))
+		switch state {
+		case "STOPPED":
+			return InternalServerError(fmt.Sprintf("Container (%s) is not running", name))
+		case "CREATED":
+			return InternalServerError(fmt.Sprintf("Container (%s) is not running", name))
+		case "SUSPENDED":
+			return InternalServerError(fmt.Sprintf("Container (%s) is not running", name))
+		case "STARTING":
+			// This is a transient state, returning conflict error to trigger a retry in the operation.
+			return ConflictError(fmt.Sprintf("container (%s) is still starting", id))
+		case "RUNNING":
+			// NO-OP - this is the state that allows an exec to occur.
+		default:
+			return InternalServerError(fmt.Sprintf("Container (%s) is in an unknown state: %s", id, state))
 		}
 
 		handle, eid, err = c.containerProxy.CreateExecTask(handle, config)
@@ -252,7 +262,6 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
 		return nil
 	}
 
-	// FIXME: We cannot necessarily check for IsConflictError here yet since the state check also returns conflict error
 	if err := retry.Do(operation, IsConflictError); err != nil {
 		op.Errorf("Failed to start Exec task for container(%s) due to error (%s)", id, err)
 		return "", err
@@ -289,7 +298,7 @@ func (c *Container) ContainerExecInspect(eid string) (*backend.ExecInspect, erro
 	// Look up the container name in the metadata cache to get long ID
 	vc := cache.ContainerCache().GetContainerFromExec(eid)
 	if vc == nil {
-		return nil, NotFoundError(eid)
+		return nil, InternalServerError(fmt.Sprintf("No container was found with exec id: %s", eid))
 	}
 	id := vc.ContainerID
 	name := vc.Name
@@ -362,7 +371,7 @@ func (c *Container) ContainerExecStart(ctx context.Context, eid string, stdin io
 	// Look up the container name in the metadata cache to get long ID
 	vc := cache.ContainerCache().GetContainerFromExec(eid)
 	if vc == nil {
-		return NotFoundError(eid)
+		return InternalServerError(fmt.Sprintf("No container was found with exec id: %s", eid))
 	}
 	id := vc.ContainerID
 	name := vc.Name
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -234,7 +234,7 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
 		// NOTE: we should investigate what we can add or manipulate in the handle in the portlayer to make this check even more granular. Maybe Add the actually State into the handle config or part of the container info could be "snapshotted"
 		// This check is now done off of the handle.
 		if state != "RUNNING" {
-			return ConflictError(fmt.Sprintf("Container (%s) is not powered on, please start the container before attempting to an exec an operation", name))
+			return ConflictError(fmt.Sprintf("Container (%s) is not powered on", name))
 		}
 
 		handle, eid, err = c.containerProxy.CreateExecTask(handle, config)
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -207,19 +207,6 @@ func (c *Container) Handle(id, name string) (string, error) {
 
 // docker's container.execBackend
 
-// FIXME: this returns a docker model, we should avoid this. This is what should be in the container proxy... But we need to confirm that it is not a public endpoint. its return hints that it is not... or that the actual return might be different
-func (c *Container) TaskInspect(cid, cname, eid string) (*models.TaskInspectResponse, error) {
-	defer trace.End(trace.Begin(fmt.Sprintf("cid(%s), cname(%s), eid(%s)", cid, cname, eid)))
-	op := trace.NewOperation(context.Background(), "")
-
-	handle, err := c.Handle(cid, cname)
-	if err != nil {
-		return nil, err
-	}
-
-	return c.containerProxy.InspectTask(op, handle, eid, cid)
-}
-
 func (c *Container) TaskWaitToStart(cid, cname, eid string) error {
 	// obtain a portlayer client
 	client := c.containerProxy.Client()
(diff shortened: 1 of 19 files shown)


lib/apiservers/engine/backends/container.go
```diff
@@ -272,56 +214,64 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
...
+			return err
+		}
+
+		return nil
 	}
 
-	err = c.containerProxy.CommitContainerHandle(handleprime, id, 0)
-	if err != nil {
-		op.Errorf("Failed to commit exec handle for container(%s) due to error(%s)", id, err)
+	// FIXME: We cannot necessarily check for IsConflictError here yet since the state check also returns conflict error
```
matthewavery: This may no longer be true. I think I fixed this, I will be checking whenever time permits from the current issues I am working on.

lib/portlayer/task/common.go
```diff
@@ -31,30 +31,38 @@ func toggleActive(op *trace.Operation, h interface{}, id string, active bool) (i
...
+
+	var task *executor.SessionConfig
+	if okS {
+		task = taskS
 	}
 
-	task := taskS
-	if handle.Runtime != nil && handle.Runtime.PowerState != types.VirtualMachinePowerStatePoweredOff {
-		op.Debugf("Task bind configuration applies to ephemeral set")
+	if okE {
```
vburenin: Does it mean exec task gets precedence over Session task?
Why would not you then construct the logic like this:
```
var task *executor.SessionConfig
if taskE, okE := handle.ExecConfig.Execs[id]; okE {
  task = taskE
} else if taskS, okS := handle.ExecConfig.Sessions[id]; okS
  task = taskS
}
```

lib/portlayer/task/common.go
```diff
@@ -31,30 +31,38 @@ func toggleActive(op *trace.Operation, h interface{}, id string, active bool) (i
...
 			return nil, err
 		}
+
+		// NOTE: this was the previous error, before merging we need to decide which one to use.
+		// return nil, fmt.Errorf("Cannot modify task %s in current state", id)
+		return nil, TaskNotFoundError{msg: fmt.Sprintf("Cannot find task %s", id)}
 	}
 
-	// if no task has been joined that can be manipulated in the container's current state
 	if task == nil {
```
vburenin: Seems like this code makes no sense anymore.

matthewavery: Your absolutely right. I was focused on the above case and forgot to remove this one. There still reamains a questing there for @hickeng as to whether or not we keep the old error message or we take the new one.

matthewavery: good catche on the empty if statement, I am surprised that `govet` did not complain at all

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -101,6 +107,7 @@ type VicContainerProxy interface {
 
 	Stop(vc *viccontainer.VicContainer, name string, seconds *int, unbound bool) error
 	State(vc *viccontainer.VicContainer) (*types.ContainerState, error)
+	GetStateFromHandle(op trace.Operation, handle string) (string, string, error)
```
hickeng: We might want to have this reflect the same state that is returned by either GetContainerInfoHandler or in the GetContainerListHandler - those all pull from Runtime and Handle state, it's just that it's pulled from the exec cache instead of from the provided handle.

lib/apiservers/engine/backends/container.go
```diff
@@ -189,75 +188,18 @@ const (
...
-			cache.ContainerCache().DeleteContainer(id)
-			return "", NotFoundError(name)
-		case *containers.GetDefault:
-			return "", InternalServerError(err.Payload.Message)
-		default:
-			return "", InternalServerError(err.Error())
-		}
-	}
-	return resp.Payload, nil
+	return c.containerProxy.Handle(id, name)
```
hickeng: I've got a strong preference *not* to move the cache management (the NotFound case) into proxy - it's supposed to be purely for the client management, marshalling, unmarshalling, and handling of error type translation.
The cache is a specific behaviour that the personality layers on top so should remain out of the proxy.

That said I'm not sure this needs to be exported from Container (captial H) - I don't think this is implementing an interface that has this as part of it's signature. May be wrong.

lib/apiservers/engine/backends/container.go
```diff
@@ -189,75 +188,18 @@ const (
...
-	_, err = client.Tasks.Wait(params)
-	if err != nil {
-		switch err := err.(type) {
-		case *tasks.WaitInternalServerError:
-			return InternalServerError(err.Payload.Message)
-		default:
-			return InternalServerError(err.Error())
-		}
-	}
+	op := trace.NewOperation(context.TODO(), "")
```
hickeng: This function isn't part of the backend container interface that we're implementing - don't know that it needs to be exported. It's also shallow enough that it could live only in the proxy.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -317,6 +324,104 @@ func (c *ContainerProxy) CreateExecTask(handle string, config *types.ExecConfig)
...
+
+	// call Bind with bindparams
+	resp, err := c.client.Tasks.Bind(bindparams)
+	if err != nil {
+		switch err := err.(type) {
+		case *tasks.BindNotFound:
+			op.Errorf("received TaskNotFound error during task bind: %s", err.Payload.Message)
+			return "", NotFoundError("container (%s) has been poweredoff")
+		case *tasks.BindInternalServerError:
+
```
hickeng: misplaced newline

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -317,6 +324,104 @@ func (c *ContainerProxy) CreateExecTask(handle string, config *types.ExecConfig)
...
+	// FIXME: right now we are only using this path for exec targets. But later the error messages may need to be changed
+	// to be more accurate.
+	params := tasks.NewInspectParamsWithContext(ctx).WithConfig(config)
+	resp, err := c.client.Tasks.Inspect(params)
+	if err != nil {
+		switch err := err.(type) {
+		case *tasks.InspectNotFound:
+			// These error types may need to be expanded. NotFoundError does not fit here.
+			op.Errorf("received a TaskNotFound error during task inspect: %s", err.Payload.Message)
+			return nil, ConflictError(fmt.Sprintf("container (%s) has been poweredoff", cid))
```
hickeng: I'm not sure turning this into a conflict error is the right thing to do in the proxy. While it could be used to force a handle refresh it could also lead to looping behaviour if the task remains not found.

If we are to return a conflict error from here I think it should be the port layer that generates it - and does so by checking the handle provided against it's cache state. If there's a conflict in changeVersion then a conflict error is legitimate. If there isn't then it's just a mistake in the eid/cid that's used.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -317,6 +324,104 @@ func (c *ContainerProxy) CreateExecTask(handle string, config *types.ExecConfig)
...
+		Handle: handle,
+		ID:     eid,
+	}
+
+	params := tasks.NewWaitParamsWithContext(ctx).WithConfig(config)
+	_, err = c.client.Tasks.Wait(params)
+	if err != nil {
+		switch err := err.(type) {
+		case *tasks.WaitNotFound:
+			return InternalServerError("the Container(%s) has been shutdown during execution ofthe exec operation")
```
hickeng: incorrectly capitalized `Container` and a missing space (ofthe) - you also don't need to pack the ID into the error - it's going to be returned to the client as the response from an operation on a specific container.

I'd also recommend saying "exited" instead - it's a more imperative phrasing and leaks less of the back end implementation: `the container exited during the exec operation`

Please run `make misspell` against this code.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -317,6 +324,104 @@ func (c *ContainerProxy) CreateExecTask(handle string, config *types.ExecConfig)
...
+		case *tasks.WaitPreconditionRequired:
+			return InternalServerError("container(%s) must be powered on in order perform the desired exec operation")
```
hickeng: `container must be running to perform an exec` - short is sweet, we've got way too many words adding no value in a lot of these messages.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -317,6 +324,104 @@ func (c *ContainerProxy) CreateExecTask(handle string, config *types.ExecConfig)
...
+	if !ok {
+		op.Errorf("Unable to marshal string object from BindTask response for handle(%s) on eid(%s)", handle, eid)
+		// TODO: perhaps a better error message here?
+		return "", InternalServerError("An unknown error occurred during the handling of this request")
+	}
+
+	return respHandle, nil
+}
+
+func (c *ContainerProxy) WaitTask(op trace.Operation, cid string, cname string, eid string) error {
```
hickeng: Is there a good reason to have this take a `cid` string instead of a Handle? Consistency suggests using a handle.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -317,6 +324,104 @@ func (c *ContainerProxy) CreateExecTask(handle string, config *types.ExecConfig)
 	return handleprime, eid, nil
 }
 
+func (c *ContainerProxy) InspectTask(op trace.Operation, handle string, eid string, cid string) (*models.TaskInspectResponse, error) {
```
hickeng: I don't see `cid` being used except for debug logging. If we're using op logging then this should be unnecessary as the calling function can log the cid.

See my other comments about dropping the parameterized error messages where it doesn't add value (it _does_ make parsing the errors in scripts harder).

lib/apiservers/engine/backends/container.go
```diff
@@ -481,8 +406,11 @@ func (c *Container) ContainerExecStart(ctx context.Context, eid string, stdin io
 		defer cancel()
 
 		// we do not return an error here if this fails. TODO: Investigate what exactly happens on error here...
+		// FIXME: This being in a retry could lead to multiple writes to stdout(?)
 		go func() {
 			defer trace.End(trace.Begin(eid))
+
+			// FIXME: NEEDS CONTAINER PROXY
```
hickeng: I think this now has a container proxy entry - should be updated to use it.

lib/apiservers/engine/backends/container.go
```diff
@@ -481,8 +406,11 @@ func (c *Container) ContainerExecStart(ctx context.Context, eid string, stdin io
 		defer cancel()
 
 		// we do not return an error here if this fails. TODO: Investigate what exactly happens on error here...
+		// FIXME: This being in a retry could lead to multiple writes to stdout(?)
```
hickeng: This should not be writing an error in the goroutine - this needs properly integrating with the retry logic and context cancellation.

I'm unsure why this is in a goroutine actually. We should check to see if an `exec_start` event is generated when trying to exec an non-existent binary in regular docker - if it reflects the container behaviour it will not generate that event on failure.

Please figure out if this needs to be in a goroutine and comment why in the code or adjust it appropriately.

lib/apiservers/engine/backends/container.go
```diff
@@ -272,56 +214,64 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
...
+		handle, state, err := c.containerProxy.GetStateFromHandle(op, handle)
+		if err != nil {
+			return InternalServerError(err.Error())
+		}
 
-	handleprime, eid, err := c.containerProxy.CreateExecTask(handle, config)
-	if err != nil {
-		op.Errorf("Failed to create exec task for container(%s) due to error(%s)", id, err)
-		return "", InternalServerError(err.Error())
+		// NOTE: we should investigate what we can add or manipulate in the handle in the portlayer to make this check even more granular. Maybe Add the actually State into the handle config or part of the container info could be "snapshotted"
```
hickeng: See my comment about `GetStateFromHandle` returning state much the same as `GetContainerInfo` or the truncated form from `GetContainerList` - this could then passed through the `convert.State()` function to get the granular state I think you're talking about here.

lib/apiservers/engine/backends/container.go
```diff
@@ -272,56 +214,64 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
...
+		// This check is now done off of the handle.
+		if state != "RUNNING" {
+			return ConflictError(fmt.Sprintf("Container (%s) is not powered on, please start the container before attempting to an exec an operation", name))
```
hickeng: This error is actually too coarse - the container could still be `Starting` or `Stopping` or a bunch of transient states. The transient states would ideally result in a retry given they're inherently unstable. Starting is the state where this is most likely to occur for us as it's the state that doesn't really exist in regular docker where a process is either running or it's not.
Same comment applies about verbosity - and we already have this message in string form elsewhere suggesting a common state check to error function mapping could be useful: `container is not running, unable to perform an exec`

lib/apiservers/engine/backends/container.go
```diff
@@ -272,56 +214,64 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
...
+		}
+
+		op.Debugf("State checks succeeded for exec operation on container(%s)", id)
```
hickeng: I think we're going to have a log message from the CreateExecTask entry so this debug message adds no value.

lib/apiservers/engine/backends/container.go
matthewavery: It is not, but it does make a call via swagger which is why it has been moved there. I can pull the cache portion out, but  I still think the `GetHandle` style endpoint should be move to the proxy. I kept that function as a passthrough since it is used in several other functions and I was making an effort to not add additional bulk to this PR.

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.robot
```diff
@@ -125,24 +125,24 @@ Exec NonExisting
...
-     #:FOR  ${idx}  IN RANGE  1  10
-     #\   ${result}=  Wait For Process  exec-%{VCH-NAME}-${idx}  timeout=2 mins
-     #\   ${combinedOutput}=  Catenate  ${combinedOutput}  ${result.stderr}${\n}
-     #
-     #Should Contain  ${combinedOutput}  Cannot complete the operation, container ${id} has been powered off during execution
+     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull ${busybox}
+     Should Be Equal As Integers  ${rc}  0
+     Should Not Contain  ${output}  Error
+     ${suffix}=  Evaluate  '%{DRONE_BUILD_NUMBER}-' + str(random.randint(1000,9999))  modules=random
+     Set Test Variable  ${ExecPowerOffContainer}  I-Have-Two-Anonymous-Volumes-${suffix}
```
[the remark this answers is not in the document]
matthewavery: it does not, this was an accidental naming collision when I was writing some code for the volume deletion issue I handled during vic-week sorry about that.

lib/portlayer/task/common.go
matthewavery: I actually have been fairly unsure if crownding a branching statement with the var assignments is good practice with golang. I have made  the change and the code definitely looks more concise, I avoided this orginally since I was worried about readability.

lib/apiservers/engine/backends/container.go
matthewavery: Upon rereading this with some sleep, I think that is exactly what you were asking for. :) I will be moving the cache handling of the not found case back into the container.go file rather than dumping the entire function handling into the proxy.

lib/apiservers/engine/backends/container.go
matthewavery: totally agreed, this will also tie the `opID` to the appropriate start operation. I am not sure as to whether or not that function is directly called by a docker client endpoint hit. I will check the configuration of the handlers to see if it is.

lib/apiservers/engine/backends/container_proxy.go
matthewavery: In this case it is also used in error message reporting. I can remove the logging, but we are still reporting this error message:
 `return nil, ConflictError(fmt.Sprintf("container (%s) has been poweredoff", cid))`

Which is then handed directly back to the docker client. I believe that I added the logging and I will remove it, but the `cid` is still needed to report the shutdown. 

I guess my other question to this is, do we need to include the container ID in the error message to begin with? The caller must know the target of `docker exec` to perform the operation. I am mainly thinking of automated scripts here that may have paramaterized inputs. Having the ID in the error message in this case would make it much easier for people to debug problems in their scripts.

lib/apiservers/engine/backends/container_proxy.go
matthewavery: Agreed, I will likely look at adding a different error type that is translated to a valid error in `container.go` since these error messages are adhering to the docker client return codes.

lib/apiservers/engine/backends/container_proxy.go
matthewavery: I wondered about this as well. When I did my first pass at refactoring I wanted to avoid changing behavior too much until I saw a `full ci` passing. As far as I can tell this is always called during an `ExecStart` which has already obtained a handle. In light of that it makes absolute since that we would pass in a handle instead retrieving another. This should also improve the performance of exec some as we avoid another http call.

lib/apiservers/engine/backends/container_proxy.go
matthewavery: hmmm my misspell seems happy with that odd... Also the CI run of `misspell` has passed this too. I will fix that immediately. How odd...

lib/apiservers/engine/backends/container_test.go
```diff
@@ -344,6 +345,17 @@ func (m *MockContainerProxy) State(vc *viccontainer.VicContainer) (*types.Contai
 	return state, nil
 }
 
+func (m *MockContainerProxy) GetStateFromHandle(op trace.Operation, handle string) (string, string, error) {
+	return "", "", nil
+}
+
+func (m *MockContainerProxy) InspectTask(op trace.Operation, handle string, eid string, cid string) (*models.TaskInspectResponse, error) {
```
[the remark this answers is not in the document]
matthewavery: Yeah, this really should not have a model being returned at all(per my comment in the code as well). To be honest, I looked in the git time capsule from `magit` this code was originally written at the same time that the container proxy was being added. So likely it was just a hold over from when we were making the switch and never got addressed. I will address it in this PR. I should have already, but I wanted to confirm that everything was functioning properly before I started to refactor more. In an effort to keep the meat of this PR reviewable. :)

lib/apiservers/portlayer/restapi/handlers/task_handlers.go
```diff
@@ -257,14 +274,24 @@ func (handler *TaskHandlersImpl) WaitHandler(params tasks.WaitParams) middleware
...
+	// wait task to set started field to something
 	// wait task to set started field to something
 	err := task.Wait(&op, handle, params.Config.ID)
 	if err != nil {
-		op.Errorf("%s", err.Error())
-
-		return tasks.NewWaitInternalServerError().WithPayload(
-			&models.Error{Message: err.Error()},
-		)
+		switch err := err.(type) {
```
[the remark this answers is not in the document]
matthewavery: I will mark them at the `Error` level since they are indications of an Error which should also be visible on standard deployments.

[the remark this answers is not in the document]
matthewavery: and work on making them more concise. I wanted the logs to have an indicator as to what error case was seen in the handlers. Especially since the handler is the entrance to the portlayer.

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.robot
```diff
@@ -125,24 +125,24 @@ Exec NonExisting
...
+     ${rc}  ${id}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -itd --name ${ExecPoweroffContainer} ${busybox} /bin/top
+     Should Be Equal As Integers  ${rc}  0
+
+     :FOR  ${idx}  IN RANGE  1  20
+     \   Start Process  docker %{VCH-PARAMS} exec ${id} /bin/top  alias=exec-%{VCH-NAME}-${idx}  shell=true
```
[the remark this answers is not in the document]
matthewavery: I will make a separate test for this. Then I can confirm how it performs.

lib/portlayer/task/inspect.go
```diff
@@ -35,42 +34,33 @@ func Inspect(op *trace.Operation, h interface{}, id string) (*executor.SessionCo
...
-	op.Debugf("exec tasks during inspect: %s", etasks)
 
-	if !okS && !okE {
-		if handle.Runtime.PowerState == types.VirtualMachinePowerStatePoweredOff {
-			powerStateError := TaskPowerStateError{
-				msg: fmt.Sprintf("the operation cannot be completed, container(%s) has been shut down during the operations execution.", handle.ExecConfig.ID),
-			}
-			return nil, powerStateError
-		}
+	// print all of them, otherwise we will have to assemble the id list regardless of
```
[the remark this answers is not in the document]
matthewavery: Looking at this more, I think I was trying to avoid looking through the lists. But the lists come from the handle so they will all be associated with the correct call(or should be barring some other failure). I am going to remove this comment entirely.

lib/apiservers/engine/backends/container.go
anchal-agrawal: +1 on the terse error message.

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.md
```diff
@@ -23,5 +23,41 @@ This test requires that a vSphere server is running and available
 * Step 2-6 should echo the ID given
 * Step 7 should return an error
 
+## Exec
```
anchal-agrawal: Looks like a stray heading.

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.md
```diff
@@ -23,5 +23,41 @@ This test requires that a vSphere server is running and available
...
+
 # Possible Problems:
 None
+
+# Exec Power Off test for long running Process
```
anchal-agrawal: I'm a fan of this test doc approach as it makes it easier to follow along and understand what the test is accomplishing, as opposed to simply listing commands in the markdown file. cc @mhagen-vmware

Overhauling all our test docs would be way too tedious at this point, I propose we adopt this approach for all new tests and fix up any old ones we can along the way.

lib/apiservers/engine/backends/container.go
```diff
@@ -272,56 +213,63 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
...
+			return InternalServerError(err.Error())
+		}
 
-	handleprime, eid, err := c.containerProxy.CreateExecTask(handle, config)
-	if err != nil {
-		op.Errorf("Failed to create exec task for container(%s) due to error(%s)", id, err)
-		return "", InternalServerError(err.Error())
+		// NOTE: we should investigate what we can add or manipulate in the handle in the portlayer to make this check even more granular. Maybe Add the actually State into the handle config or part of the container info could be "snapshotted"
+		// This check is now done off of the handle.
+		if state != "RUNNING" {
```
anchal-agrawal: Could we use a constant here? I see `state.Running` was used previously here.

matthewavery: Yes, I will check for those constants now. I was confirming that this logic works and forgot to switch to the comments earlier on.

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.md
matthewavery: +1 to this @mhagen-vmware thoughts?

lib/apiservers/engine/backends/container.go
matthewavery: So `state.Running` was a bool from the docker `ContainerState` struct previously. these new returns are really string representations of the internal state kept in the portlayer. So the portlayer is relaying the snap shotted state. I will make some constants for them, but I will not be planning to move to using the `ContainerState` struct again as it does not easily map for us. We do have a `covert.state` function. but it actually takes the `VicContainer` struct from the persona cache as the direct input. We are using a handle in order to check the state in the snapshot of time where we began manipulating aspects of the container.

lib/apiservers/engine/backends/container.go
```diff
@@ -231,10 +231,20 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
...
+		switch state {
+		case "STOPPED":
+			return InternalServerError(fmt.Sprintf("Container (%s) is not running", name))
+		case "CREATED":
+			return InternalServerError(fmt.Sprintf("Container (%s) is not running", name))
+		case "SUSPENDED":
+			return InternalServerError(fmt.Sprintf("Container (%s) is not running", name))
+		case "STARTING":
+			// This is a transient state, returning conflict error to trigger a retry in the operation.
+			return ConflictError(fmt.Sprintf("container (%s) is still starting", id))
```
hickeng: 👍 
Now _this_ is how it's supposed to be used!

lib/apiservers/engine/backends/container.go
```diff
@@ -289,7 +298,7 @@ func (c *Container) ContainerExecInspect(eid string) (*backend.ExecInspect, erro
 	// Look up the container name in the metadata cache to get long ID
 	vc := cache.ContainerCache().GetContainerFromExec(eid)
 	if vc == nil {
-		return nil, NotFoundError(eid)
+		return nil, InternalServerError(fmt.Sprintf("No container was found with exec id: %s", eid))
```
hickeng: Why not use the `NotFoundError`? This does seem like a 404 - presumably an incorrect `eid` was supplied.

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.robot
```diff
@@ -155,13 +155,13 @@ Exec During Poweroff Of A Container Performing A Short Running Task
 
      ${suffix}=  Evaluate  '%{DRONE_BUILD_NUMBER}-' + str(random.randint(1000,9999))  modules=random
      Set Test Variable  ${ExecPoweroffContainerShort}  Exec-Poweroff-${suffix}
-     ${rc}  ${id}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -itd --name ${ExecPoweroffContainerShort} ${busybox} sleep 5
+     ${rc}  ${id}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -itd --name ${ExecPoweroffContainerShort} ${busybox} /bin/ls /; sleep 20;
```
hickeng: The chained commands here make me uncomfortable as this would not work as expected when run in a shell.
Does it still work if you use `"/bin/ls; sleep 20"`?

Also, why are we making this sleep longer? If exec is running fast then we could easily have completed all of the execs before the sleep completes. I thought the essence of this test was to issue and exec _as_ the container is stopping?

lib/apiservers/engine/backends/container.go
```diff
@@ -232,16 +241,16 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
 		}
 
 		switch state {
-		case "STOPPED":
+		case StoppedState:
```
hickeng: you can use a comma separated list in a case statement:
```
switch state {
case CreatedState, StoppedState, SuspendedState:
   return InternalServerError(fmt.Sprintf("Container (%s) is not running", name))
case StartingState:
   return ConflictError(fmt.Sprintf("container (%s) is still starting", id))
case RunningState:
   // noop
default:
   return InternalServerError(fmt.Sprintf("Container (%s) is in an unknown state: %s", id, state))
}

File: lib/apiservers/engine/backends/container.go (hunk 27 above)
Comment (matthewavery) [reply lib/apiservers/engine/backends/container.go]: When we look into the `NotFoundError` function it is actually prefacing it with "Container Not Found". What I will do is create my own custom `TaskNotFoundError` function instead. and place it in the same place as the `NotFoundError` function.

File: lib/apiservers/engine/backends/container.go (hunk 29 above)
Comment (matthewavery) [reply lib/apiservers/engine/backends/container.go]: Good point, will fix this. I knew I was doing something silly there when I had so many wasted lines.

File: tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.robot (hunk 28 above)
Comment (matthewavery) [reply tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.robot]: Exec is running faster in some cases. But I am unsure if the execs will all work here. we can change it back to 5. I changed it 20 since I saw some exec's succeeding sometimes. I will try your additional suggestion above and lower it to 5 seconds. then run it and see if any of the execs worked. they were not before which was odd.

File: tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.robot [hunk 30]
```diff
@@ -124,25 +124,128 @@ Exec NonExisting
...
+
+     :FOR  ${idx}  IN RANGE  1  10
+     \   ${result}=  Wait For Process  exec-%{VCH-NAME}-${idx}  timeout=2 mins
+     \   ${combinedErr}=  Catenate  ${combinedErr}  ${result.stderr}${\n}
+     \   ${combinedOut}=  Catenate  ${combinedOut}  ${result.stdout}${\n}
+
+     Should Contain  ${combinedErr}  Container (${id}) is not running
+
+     # We should get atleast one successful exec...
+     Should Contain  ${combinedOut}  bin
```
hickeng: pull this out into a separate keyword - it's significant duplication in just these two tests.

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.robot
```diff
@@ -124,25 +124,128 @@ Exec NonExisting
...
-     #
-     #Should Contain  ${combinedOutput}  Cannot complete the operation, container ${id} has been powered off during execution
+Concurrent Simple Exec
+     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull ${busybox}
+     Should Be Equal As Integers  ${rc}  0
+     Should Not Contain  ${output}  Error
+
+     ${suffix}=  Evaluate  '%{DRONE_BUILD_NUMBER}-' + str(random.randint(1000,9999))  modules=random
+     Set Test Variable  ${ExecSimpleContainer}  Exec-simple-${suffix}
+     ${rc}  ${id}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -itd --name ${ExecSimpleContainer} ${busybox} sleep 30
```
hickeng: I think you meant this to be indefinite, not `sleep 30`. It looks as though this is testing concurrent exec, _without_ the risk of container stopping underneath us.

matthewavery: One was expecting full success and the other not. Once i had written them and reworked some of them they ended up being similar. I believe the short running task should be shifted to having a long running exec I think that would be a valuable use of that test. The simple case is to say we have a container and we expect all execs to succeed. So I will change that to `/bin/top` as it should definitely be indefinite(pun not intended).

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.robot
matthewavery: For this I would expect at least one success. So the keyword itself should probably check for both 1 and 2. in combination. `Simple Exec` is not meant to fail. If we abstract it out to a keyword then I can add the OR logic there in an effort to reduce the success checking in general. but having an OR on success and failure may mean that test never fails.

mhagen-vmware: agreed on the keyword breakout. agreed on the OR. by my reading of this test it is completely valid to get returned back 10 valid errors and never see the directory output - especially if the first exec takes more than 10s to happen..

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.md
```diff
@@ -25,3 +25,66 @@ This test requires that a vSphere server is running and available
...
+
+# Exec Power Off test for short Running Process
+## Purpose:
+This test is designed to start a container and then initiate a long running exec operation(/bin/top). The expectation is that the exec will succeed in execution and run for the length of the containers life. Once the container exits the exec operation should also exit cleanly without error.
+
+## Test Steps
+1. Pull an image that contains `sleep`. Busybox suffices here.
+2. Create a container running `sleep` to simulate a short running process.
+3. Run the container and detach from it.
+4. Start 20 simple exec operations in parallel against the detached container.
```
mhagen-vmware: this is not the steps you have implemented below

matthewavery: ah yeah I need to clean that up since I changed the test. :( fixing that now!

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.md
```diff
@@ -46,12 +50,15 @@ None
...
+This test is designed to test running exec's against a container running a process which would not naturally exit. Then after running many execs concurrently we explicitly stop the container. We should then handle the the execs in their varying states properly.
+
 ## Test Steps
 1. Pull an image that contains `/bin/top`. Busybox suffices here.
-2. Create a container running `/bin/top` to simulate a long running process.
+2. Create a container running `/bin/top` to simulate a long running process that should not exit on it's own.
 3. Run the container and detach from it.
 4. Start 20 simple exec operations in parallel against the detached container.
-5. Stop the container while the execs are still running in order to trigger the exec power off errors.
+5. Explicitly stop the container while the execs are still running in order to trigger the exec power off errors.
```
hickeng: I think there is a conceptual race between (4) and (5). You are attempting to ensure that some of the execs have been dispatched before stop takes effect, and that some are issued _during the stop process_.
It's ok to note that this test is stochastic and not expected to behave in the same manner every time.

lib/apiservers/engine/backends/container.go
```diff
@@ -267,7 +267,12 @@ func (c *Container) ContainerExecCreate(name string, config *types.ExecConfig) (
 		return nil
 	}
 
-	if err := retry.Do(operation, IsConflictError); err != nil {
+	// configure custom exec back off configure
+	backoffConf := retry.NewBackoffConfig()
```
hickeng: you have the same config duplicated twice - consider consolidating if they are _supposed_ to be the same.
Also note _why_ those times have been chosen; if lacking concrete numbers it's okay to express what you're attempting to accomplish with these timings. At the moment it's left to the imagination of the reader.

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.robot
```diff
@@ -171,19 +171,41 @@ Exec During Poweroff Of A Container Performing A Long Running Task
      ${rc}  ${id}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -itd --name ${ExecPoweroffContainerLong} ${busybox} /bin/top
```
hickeng: You have `sleep` documented in the test markdown

lib/portlayer/task/wait.go
```diff
@@ -33,7 +33,7 @@ func Wait(op *trace.Operation, h interface{}, id string) error {
 	}
 
 	if handle.Runtime != nil && handle.Runtime.PowerState != types.VirtualMachinePowerStatePoweredOn {
-		err := fmt.Errorf("Unable to wait for task when container %s is not running", id)
+		err := fmt.Errorf("Unable to wait for task when container %s is not running", handle.ExecConfig.ID)
```
hickeng: incorrect capitalization of error message

tests/test-cases/Group1-Docker-Commands/1-38-Docker-Exec.robot
```diff
@@ -19,6 +19,41 @@ Suite Setup  Conditional Install VIC Appliance To Test Server
 Suite Teardown  Cleanup VIC Appliance On Test Server
 Test Timeout  20 minutes
 
+*** Keywords ***
+Verify Poweroff During Exec Error Message
+       [Arguments]  ${error}  ${containerID}  ${containerName}
+       Set Test Variable  ${msg1}  Container (${containerName}) is not running
+       Set Test Variable  ${msg2}  container (${containerID}) has been poweredoff
```
hickeng: `poweredoff` -> `powered off`, or even `stopped`

PR #6791, opened 2017-11-19
Configure managedBy info for plugin and OVA

jzt:
This change adds functionality to the VIC ui installer that allows it to configure both the extension and the OVA that the extension manages. This gets us away from having the asynchronous dependency on having the fileserver waiting to configure the OVA until the user decides to install the plugin.

Fixes https://github.com/vmware/vic-ui/issues/26
Fixes https://github.com/vmware/vic-ui/issues/26.

### DIFF
--- a/lib/install/ova/configure.go
+++ b/lib/install/ova/configure.go
@@ -1,0 +1,135 @@
+// Copyright 2016 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ova
+
+import (
+	"context"
+	"net"
+	"net/url"
+
+	log "github.com/Sirupsen/logrus"
+
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/lib/config/dynamic/admiral"
+	"github.com/vmware/vic/pkg/errors"
+	"github.com/vmware/vic/pkg/vsphere/session"
+	"github.com/vmware/vic/pkg/vsphere/tasks"
+	"github.com/vmware/vic/pkg/vsphere/vm"
+)
+
+const (
+	// ManagedByKey defines the extension key to use in the ManagedByInfo of the OVA
+	ManagedByKey = "com.vmware.vic"
+	// ManagedByType defines the type to use in the ManagedByInfo of the OVA
+	ManagedByType = "VicApplianceVM"
+)
+
+// ConfigureManagedByInfo takes sets the ManagedBy field for the VM specified by ovaURL
+func ConfigureManagedByInfo(ctx context.Context, config *session.Config, ovaURL string) error {
+	sess := session.NewSession(config)
+	sess, err := sess.Connect(ctx)
+	if err != nil {
+		return err
+	}
+
+	v, err := getOvaVMByTag(ctx, sess, ovaURL)
+	if err != nil {
+		return err
+	}
+
+	log.Infof("Attempting to configure ManagedByInfo")
+	err = configureManagedByInfo(ctx, sess, v)
+	if err != nil {
+		return err
+	}
+
+	log.Infof("Successfully configured ManagedByInfo")
+	return nil
+}
+
+func configureManagedByInfo(ctx context.Context, sess *session.Session, v *vm.VirtualMachine) error {
+	spec := types.VirtualMachineConfigSpec{
+		ManagedBy: &types.ManagedByInfo{
+			ExtensionKey: ManagedByKey,
+			Type:         ManagedByType,
+		},
+	}
+
+	info, err := v.WaitForResult(ctx, func(ctx context.Context) (tasks.Task, error) {
+		return v.Reconfigure(ctx, spec)
+	})
+
+	if err != nil {
+		log.Errorf("Error while setting ManagedByInfo: %s", err)
+		return err
+	}
+
+	if info.State != types.TaskInfoStateSuccess {
+		log.Errorf("Setting ManagedByInfo reported: %s", info.Error.LocalizedMessage)
+		return err
+	}
+
+	return nil
+}
+
+func getOvaVMByTag(ctx context.Context, sess *session.Session, u string) (*vm.VirtualMachine, error) {
+	ovaURL, err := url.Parse(u)
+	if err != nil {
+		return nil, err
+	}
+
+	host := ovaURL.Hostname()
+
+	log.Debugf("Looking up host %s", host)
+	ips, err := net.LookupIP(host)
+	if err != nil {
+		return nil, errors.Errorf("IP lookup failed: %s", err)
+	}
+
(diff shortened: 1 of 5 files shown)


lib/install/ova/configure.go
```diff
@@ -0,0 +1,117 @@
...
+	return nil
+}
+
+func getVM(ctx context.Context, client *vim25.Client, u string) (*object.VirtualMachine, error) {
+	ovaURL, err := url.Parse(u)
+	if err != nil {
+		return nil, err
+	}
+
+	ip := ovaURL.Hostname()
```
hickeng: You may want to perform a lookup on the Hostname element in case it's phrased as an FQDN rather than an IP.
This would be necessary if using dyndns and DHCP as a way of providing a stable reference to the appliance - in that case the IP could not be in the certificate, but the FQDN would be.

lib/install/ova/configure.go
```diff
@@ -0,0 +1,117 @@
...
+	ref, err := searchIndex.FindByIp(ctx, dc, ip, true)
+	if err != nil {
+		return nil, errors.Errorf("failed to find vm by IP: %s", err)
+	}
+
+	if ref == nil {
+		return nil, errors.Errorf("failed to find vm with IP %s", ip)
+	}
+
+	ref, err = finder.ObjectReference(ctx, ref.Reference())
```
hickeng: I suggest checking that the VM has the guestinfo keys we expect - this would protect against accidentally identifying the wrong VM (for example if multiple VMs have the same IP which is perfectly acceptable and perhaps even expected with NSX/VXLAN configurations).
Although it looks like `FindByIp` only returns a single match...

lib/install/ova/configure.go
```diff
@@ -0,0 +1,117 @@
...
+const (
+	// ManagedByKey defines the extension key to use in the ManagedByInfo of the OVA
+	ManagedByKey = "com.vmware.vic"
+	// ManagedByType defines the type to use in the ManagedByInfo of the OVA
+	ManagedByType = "VicApplianceVM"
+)
+
+func configureManagedByInfo(ctx context.Context, sess *session.Session, v *object.VirtualMachine) error {
+	vm := virtualmachine.NewVirtualMachineFromVM(ctx, sess, v)
+	spec := types.VirtualMachineConfigSpec{
```
hickeng: I'd also suggest applying this when the OVA is initially given the Admin credentials during Getting Started. This is when the OVA tags itself so that it can be located by the VCHs - it makes sense to also enable discovery by the plugin at the same time.

Performing this during Getting Started will also mean that multiple OVAs can easily be marked as managed by the extension without an extra user step and/or having to reinstall the plugin.

jzt: @hickeng yeah that's our next step. I wanted to do that this sprint but we didn't have the time.

lib/install/ova/configure.go
jzt: Fixed, let me know if you have feedback on my approach!

lib/install/ova/configure.go
```diff
@@ -0,0 +1,123 @@
...
+
+	for i, v := range vms {
+		vmIP, err := v.WaitForIP(ctx)
+		if err != nil && i == len(vms)-1 {
+			return nil, errors.Errorf("failed to get VM IP: %s", err)
+		}
+
+		// verify the tagged vm has the IP we expect
+		if vmIP == ip {
+			log.Debugf("Found OVA with matching IP: %s", ip)
```
zjs: I wonder if we might want some more debug logging like this. (Otherwise, I could imagine this being hard to troubleshoot in a production environment.)

Maybe we could try logging the hostname that we're looking up, the IPs we find for it, the tagged vms admiral returns, and _then_ information about the one we found (maybe including some other information from `v`, like its name or moref, since we know IP isn't a unique identifier?)

jzt: Fixed!

jzt:
@mhagen-vmware I've been testing with a unit test while iterating, but unless we already have an ova deployed in CI somewhere, an integration test isn't going to happen.

mhagen-vmware:
just unit tests esp for the new functions in configure.go, I don't see a unit test here? Where is the one you have been using?

jzt:
```go
package ova

import (
        "context"
        "crypto/tls"
        "fmt"
        "net/url"
        "testing"

        "github.com/vmware/govmomi/object"
)

const (
        username = "administrator@vsphere.local"
        password = "Admin!23"
        vcURL    = "https://10.161.237.9/sdk"
        vcDNS    = "https://sc-rdops-vm16-dhcp-237-9.eng.vmware.com/sdk"
        ovaURL   = "https://10.160.31.195:9443/path/to/something/illegal"
)

func TestGetOvaVMByTag(t *testing.T) {
        ctx := context.Background()

        vc, err := url.Parse(vcDNS)
        if err != nil {
                fmt.Printf("Failed to parse VC url: %s", err)
                t.FailNow()
        }

        vc.User = url.UserPassword(username, password)

        var cert object.HostCertificateInfo
        if err = cert.FromURL(vc, new(tls.Config)); err != nil {
                log.Error(err.Error())
                t.FailNow()
        }

        if cert.Err != nil {
                log.Errorf("Failed to verify certificate for target=%s (thumbprint=%s)", vc.Host, cert.ThumbprintSHA1)
                log.Error(cert.Err.Error())
        }

        tp := cert.ThumbprintSHA1
        log.Infof("Accepting host %q thumbprint %s", vc.Host, tp)

        sessionConfig := &session.Config{
                Thumbprint:     tp,
                Service:        vc.String(),
                DatacenterPath: "/ha-datacenter",
                DatastorePath:  "datastore1",
                User:           vc.User,
                Insecure:       true,
        }

        s := session.NewSession(sessionConfig)
        sess, err := s.Connect(ctx)
        if err != nil {
                log.Errorf("Error connecting: %s", err.Error())
        }
        defer sess.Logout(ctx)

        sess, err = sess.Populate(ctx)
        if err != nil {
                log.Errorf("Error populating: %s", err.Error())
        }

        vm, err := getOvaVMByTag(ctx, sess, ovaURL)
        if err != nil {
                log.Errorf("Error getting OVA by tag: %s", err.Error())
        }
        log.Infof("%s", vm.String())
}
```
I can't check it in without a VC + OVA to point it to.

jzt:
Added the test I showed you (but set it to skip in the absence of required info), and one to test the url error case you wanted covered.

lib/install/ova/configure.go
```diff
@@ -0,0 +1,129 @@
...
+	err = configureManagedByInfo(ctx, sess, v)
+	if err != nil {
+		return err
+	}
+
+	log.Infof("Successfully configured ManagedByInfo")
+	return nil
+}
+
+func getOvaVMByTag(ctx context.Context, sess *session.Session, u string) (*vm.VirtualMachine, error) {
```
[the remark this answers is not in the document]
jzt: The entrypoint into this code is via the ui install script. The user supplies username/password, and a new session is created in the `vic-ui` binary. That session is passed along into this function, whose primary purpose is to verify that the host/IP of the VM we found via the tag is the same host/IP as specified in the OVA url that we passed in (the OVA url is auto-populated into the installer script on the backend when the "Getting Started" step is kicked off).

So no, this doesn't refactor ova-by-tag code in the engine, but rather uses `admiral.DefaultDiscovery.Discover()` to find an ova by tag, and it uses the ova URL to compare the two IP addresses. If that checks out, then the `ManagedBy` info is set on the OVA VM so that the plugin and the OVA can play nice together.

Eventually this will all move into the fileserver code and will run after the "Getting Started" step is complete. The user will not need to re-enter credentials (as they do now by having to run the script), as they will already be known at the time of execution since they are entered at the beginning of "Getting Started". All of this logic will be removed from the vic-ui binary and it will return to a more plugin-agnostic universal installer.

PR #6794, opened 2017-11-19
Check vCenter options in vic-machine

### LINKED ISSUE #6473: vic-machine: set vpxd.SerialPort.PersistNetworkBacking to true
**Story**
As a viadmin I expect to be able to reboot vCenter without damaging containerVM function

**Details**
@matthewavery please sanity check me - I seem to have omitted when this occurred from my notes so going off vague recollections from internal storage...
This occurs when rebooting a vCenter and containerVMs are configured with a vSPC connection. After reboot the VM cannot be configured to reconnect to the vSPC on the endpointVM, manually or otherwise. Newly created VMs are fine.
The 'damaged' containerVMs show vSPC as as Unsupported and attempts to reconfigure them with correct settings fail.

The fix is to set the following `config.vpxd.SerialPort.PersistNetworkBacking` advanced vCenter option - this only applies to newly created VMs after that point but functions as desired.

The always awesome @lamw has [a perl script](https://github.com/lamw/vghetto-scripts/blob/master/perl/vCenterAdvSettingMgmt.pl#L60) which has led me to the [service content->setting](http://pubs.vmware.com/vsphere-6-0/index.jsp#com.vmware.wssdk.apiref.doc/vim.ServiceInstanceContent.html) field as the probable target for automatically updating this setting.

**Acceptance**
I've broken this out into two distinct behaviours because I have little knowledge about programmatic setting of VC advanced config - only the notes above:
- [x] vic-machine checks that this value is set appropriately
- [x] vic-machine sets the value if not set (optional)
- [ ] docker attach/exec function is confirmed after vCenter restart in VC restart tests

@malikkal

andrewtchin:
Fixes #6473 
Tests are WIP

If the value is set correctly:
```
⇒  ./bin/vic-machine-linux create --target 'administrator@vsphere.local:Admin!23@10.160.96.27' --name d --no-tlsverify --compute-resource /ha-datacenter/host/cluster1 --image-store datastore1 --bridge-network bridge --thumbprint=93:17:0D:D6:C5:C7:58:EC:33:66:69:2C:13:E5:CD:BA:4C:30:34:29
Nov 19 2017 12:33:11.342-08:00 INFO  ### Installing VCH ####
Nov 19 2017 12:33:11.342-08:00 WARN  Using administrative user for VCH operation - use --ops-user to improve security (see -x for advanced help)
Nov 19 2017 12:33:11.344-08:00 INFO  Generating self-signed certificate/key pair - private key in d/server-key.pem
Nov 19 2017 12:33:11.469-08:00 WARN  Configuring without TLS verify - certificate-based authentication disabled
Nov 19 2017 12:33:15.418-08:00 INFO  Validating supplied configuration
...
Nov 19 2017 12:33:21.955-08:00 INFO  vCenter settings check OK
...
```

If the value is unset or not set to the correct value it is set to "true":
```
⇒  ./bin/vic-machine-linux create --target 'administrator@vsphere.local:Admin!23@10.160.96.27' --name e --no-tlsverify --compute-resource /ha-datacenter/host/cluster1 --image-store datastore1 --bridge-network bridge --thumbprint=93:17:0D:D6:C5:C7:58:EC:33:66:69:2C:13:E5:CD:BA:4C:30:34:29
Nov 19 2017 12:35:27.268-08:00 INFO  ### Installing VCH ####
Nov 19 2017 12:35:27.268-08:00 WARN  Using administrative user for VCH operation - use --ops-user to improve security (see -x for advanced help)
Nov 19 2017 12:35:27.268-08:00 INFO  Generating self-signed certificate/key pair - private key in e/server-key.pem
Nov 19 2017 12:35:27.355-08:00 WARN  Configuring without TLS verify - certificate-based authentication disabled
Nov 19 2017 12:35:30.885-08:00 INFO  Validating supplied configuration
Nov 19 2017 12:35:35.487-08:00 INFO  vDS configuration OK on "bridge"
...
Nov 19 2017 12:35:37.714-08:00 INFO  Set vCenter settings for containerVMs
Nov 19 2017 12:35:37.714-08:00 INFO  vCenter settings check OK
...

```

### DIFF
--- a/lib/install/validate/config.go
+++ b/lib/install/validate/config.go
@@ -29,7 +29,10 @@ import (
 	"github.com/vmware/vic/lib/constants"
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/optmanager"
 )
+
+const persistNetworkBackingKey = "config.vpxd.SerialPort.PersistNetworkBacking"
 
 type FirewallStatus struct {
 	Rule                          types.HostFirewallRule
@@ -561,4 +564,35 @@ func (v *Validator) CheckDrs(ctx context.Context) {
 	}
 	op.Info("DRS check OK on:")
 	op.Infof("  %q", v.Session.Cluster.InventoryPath)
+}
+
+// check that PersistNetworkBacking is set
+func (v *Validator) CheckPersistNetworkBacking(ctx context.Context, quiet bool) bool {
+	op := trace.FromContext(ctx, "Check vCenter serial port backing")
+	defer trace.End(trace.Begin("", op))
+
+	errMsg := "vCenter settings check SKIPPED"
+	if !v.sessionValid(op, errMsg) {
+		return false
+	}
+	if !v.IsVC() {
+		op.Info(errMsg)
+		return true
+	}
+
+	val, err := optmanager.QueryOptionValue(ctx, v.Session, persistNetworkBackingKey)
+	if err != nil {
+		// the key is not set
+		val = "false"
+	}
+	if val != "true" {
+		if !quiet {
+			op.Errorf("vCenter settings check FAILED")
+			msg := fmt.Sprintf("vCenter advanced option %s=true must be set", persistNetworkBackingKey)
+			v.NoteIssue(errors.New(msg))
+		}
+		return false
+	}
+	op.Infof("vCenter settings check OK")
+	return true
 }
--- a/lib/install/validate/validator.go
+++ b/lib/install/validate/validator.go
@@ -44,6 +44,7 @@ import (
 	"github.com/vmware/vic/pkg/registry"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/version"
+	"github.com/vmware/vic/pkg/vsphere/optmanager"
 	"github.com/vmware/vic/pkg/vsphere/session"
 )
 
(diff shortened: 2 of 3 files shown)


lib/install/validate/config.go
```diff
@@ -557,3 +560,52 @@ func (v *Validator) CheckDrs(ctx context.Context) {
...
+		if !ok {
+			log.Errorf("vCenter settings check FAILED")
+			return
+		}
+	}
+	log.Infof("vCenter settings check OK")
+}
+
+// set PersistNetworkBacking key to "true"
+func (v *Validator) SetPersistNetworkBacking(ctx context.Context) bool {
```
hickeng: This doesn't belong in the validator.

If we're setting this automatically when deploying a VCH then we should have a Dispatcher call for ConfigureVCenter or similar (my preference - automatic where viable).
If we're setting this via the `update` command then it's an explicit operation and lives with that code.

lib/install/validate/config.go
```diff
@@ -557,3 +560,52 @@ func (v *Validator) CheckDrs(ctx context.Context) {
...
+	val, err := optmanager.QueryOptionValue(ctx, v.Session, persistNetworkBackingKey)
+	if err != nil {
+		// the key is not set, set it now
+		val = "false"
+	}
+	var ok bool
+	if val != "true" {
+		ok = v.SetPersistNetworkBacking(ctx)
+		if !ok {
+			log.Errorf("vCenter settings check FAILED")
```
hickeng: This should be `v.NoteIssue` and should note user remediation, e.g. `requires vCenter advanced option vpxd.SerialPort.PersistNetworkBacking=true`.

lib/install/validate/config.go
```diff
@@ -557,3 +560,52 @@ func (v *Validator) CheckDrs(ctx context.Context) {
...
 }
+
+// check that PersistNetworkBacking is set
+func (v *Validator) CheckPersistNetworkBacking(ctx context.Context) {
+	errMsg := "vCenter settings check SKIPPED"
+	if !v.sessionValid(errMsg) {
+		return
+	}
+	if !v.IsVC() {
+		log.Info(errMsg)
```
hickeng: use `op := trace.FromContext(ctx, "Check vCenter serial port backing")` or similar and then log using `op.Infof`.
This will ease integration with the UI feature branch

lib/install/validate/config.go
andrewtchin: I will move this. I haven't included work to set this through `vic-machine update`. Is it ok to open an issue to add that feature? I documented the manual steps to set this through the mob

lib/install/validate/config.go
andrewtchin: fixed

lib/install/validate/config.go
andrewtchin: fixed

andrewtchin:
Moved to #6815 so the build can actually build

PR #6795, opened 2017-11-20
High-availability test reliability

hickeng:
This adds tolerance to conflict error in ContainerRm (duplicates work
for #6252). Having looked at the ContainerRm logic we should have
a general pass over the personality code as the use of portlayer
Handles is not really as designed; these methods should almost all
be following the pattern below, however currently they are invoking
separate calls that are not batched within a single commit:
1. new handle
2. apply changes as necessary to handle
3. commit handle
4. if conflict error, redo from (1)

In the test side this explicitly checks for VCH initialization instead of
sleeping for set period.
This still requires logic to preserve the test environment if one of the
rare and non-reproducible errors occurs.

#6786 contains the same fix for ContainerRm but it's unclear what
the intent of the HA test change is - if we need an exception of that
kind it really should apply to all operations that could hit it, not a
single test. That may well require significant rework of the tests to
allow for common condition checking - the same requirement as
for preserving test envs for non-repo errors.

Towards #4666

### DIFF
--- a/tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
+++ b/tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
@@ -183,10 +183,7 @@ Test
     # Speculating that this is here to add some stability if VMs vmotion immediately after power on
     # If so it should be replaced with a check for active tasks running against the VM or a DRS rule to avoid it completely
     Sleep  2 minutes
-
-    ${output}=  Run  govc vm.info %{VCH-NAME}
-    @{output}=  Split To Lines  ${output}
-    ${curHost}=  Fetch From Right  @{output}[-1]  ${SPACE}
+    ${curHost}=  Get VM Host Name  %{VCH-NAME}
 
     ${info}=  Run  govc vm.info \\*
     Log  ${info}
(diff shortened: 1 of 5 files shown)


tests/manual-test-cases/Group5-Functional-Tests/5-4-High-Availability.robot
```diff
@@ -204,17 +206,15 @@ Test
...
 
     ${info}=  Run  govc vm.info \\*
     Log  ${info}
 
-    # Really not sure what better to do here?  Otherwise, vic-machine-inspect returns the old IP address... maybe some sort of power monitoring? Can I pull uptime of the system?
-    Sleep  4 minutes
-    Run VIC Machine Inspect Command
-    Wait Until Keyword Succeeds  20x  5 seconds  Run Docker Info  %{VCH-PARAMS}
+    # Wait for the VCH to come back up fully - if it's not completely reinitialized it will still report the old IP address 
+    Wait For VCH Initialization
```
mhagen-vmware: without the 4 minute sleep above... I think that this could potentially return before the system has powered off and did it's HA event right? sometimes the power off of the ESXi host takes a couple of minutes.  I think in that case, we could see the vic machine inspect and docker info return before the HA event even occurs.

hickeng: Good to know. Will add some logic to confirm that host has changed, probably by using govc events to wait for the HA event.

sgairo:
failed on build does not match, restarting as it seems to be ready for merge otherwise

PR #6798, opened 2017-11-20
ops-grant-perms exposed on vic-machine create API [specific ci=Group23-VIC-Machine-Service]

### LINKED ISSUE #6632: Expose "grant permissions" option for the operations credentials in the VCH creation API
Epic #6464 includes work to optionally have the required permissions be automatically granted to the operations user when the VCH is created. That boolean setting must be exposed through the VCH creation API (likely as a write-only field on `operations_credentials`).

AngieCris:
Fixes #6632 

`grant_perms` is added to vic-machine create API as a write-only field
correspondent to the `--ops-grant-perms` flag in CLI #6774

### DIFF
--- a/lib/apiservers/service/restapi/handlers/vch_create.go
+++ b/lib/apiservers/service/restapi/handlers/vch_create.go
@@ -401,6 +401,7 @@ func buildCreate(op trace.Operation, d *data.Data, finder finder, vch *models.VC
 				c.OpsCredentials = common.OpsCredentials{
 					OpsUser:     &vch.Endpoint.OperationsCredentials.User,
 					OpsPassword: &opsPassword,
+					GrantPerms:  &vch.Endpoint.OperationsCredentials.GrantPermissions,
 				}
 			}
 		}
--- a/lib/apiservers/service/swagger.json
+++ b/lib/apiservers/service/swagger.json
@@ -759,6 +759,9 @@
                 },
                 "user": {
                   "type": "string"
+                },
+                "grant_permissions": {
+                  "type": "boolean"
                 }
               }
             }


lib/apiservers/service/swagger.json
```diff
@@ -759,6 +759,9 @@
                 },
                 "user": {
                   "type": "string"
+                },
+                "grant_perms": {
```
zjs: Let's just spell out "grant_permissions"; unlike the CLI, there's no worry about being too verbose, and we've generally managed to avoid abbreviating things in the API so far.

lib/apiservers/service/restapi/handlers/vch_create.go
```diff
@@ -401,6 +401,7 @@ func buildCreate(op trace.Operation, d *data.Data, finder finder, vch *models.VC
 				c.OpsCredentials = common.OpsCredentials{
```
zjs: Can we add some error handling here to reject requests if `vch.Endpoint.OperationsCredentials.GrantPerms` is set, but `vch.Endpoint.OperationsCredentials.User` is not?

This would match the CLI validation from 31358560bd91c7f2267f7f14dfab866e9e4f7934.

zjs:
Could you add @lcastellano to the review, since he implemented the relevant functionality?

lib/apiservers/service/restapi/handlers/vch_create.go
lcastellano: In the CLI path, we make sure not do enable granting if no ops-user was specified (the CLI defaults to the admin user). Does that need to be done here?

AngieCris: @lcastellano @zjs This is handled later in the function call to `c.OpsCredentials.ProcessOpsCredentials`. Do we want to handle this before it reaches the function call?

Issue #6677, opened 2017-11-02
Nightly Failure 11/2/2017 NFS-Volume: Portlayer fails to power on after reboot due to a hung NFS mount attempt via the go-nfs-client


matthewavery:
seen 11/2/2017
on: VC 6.0
Engine Build: 14352
Test Failure: obtain docker mount data after reboot
logs here: 
[5-22-NFS-Volume.zip](https://github.com/vmware/vic/files/1438520/5-22-NFS-Volume.zip)

Further important log information and analysis.

* Personality fails to present due to portlayer inability to initialize.

* Portlayer fails to initialize due to hung nfs mount attempt 
seen here

```
Nov  2 2017 07:20:37.646Z INFO  op=298.6 (delta:1.416013754s): Getting metadata volumes/2fcc0ea2-bf9e-11e7-a054-000c29e4a3c3/imageMetadata/DockerMetaData
Nov  2 2017 07:20:37.695Z INFO  Volumestore: Found vol 2fcc0ea2-bf9e-11e7-a054-000c29e4a3c3 on store http:///storage/volumes/default.
Nov  2 2017 07:20:37.695Z INFO  op=298.6 (delta:1.465297984s): Registering importer: ds://datastore1/%5Bdatastore1%5D%20VCH-0-9472-VOL => *vsphere.VolumeStore
Nov  2 2017 07:20:37.695Z INFO  op=298.6 (delta:1.465311048s): Registering exporter: ds://datastore1/%5Bdatastore1%5D%20VCH-0-9472-VOL => *vsphere.VolumeStore
Nov  2 2017 07:20:37.695Z INFO  op=298.6 (delta:1.465319168s): Registering importer: http:///storage/volumes/default => *vsphere.VolumeStore
Nov  2 2017 07:20:37.695Z INFO  op=298.6 (delta:1.465323557s): Registering exporter: http:///storage/volumes/default => *vsphere.VolumeStore
Nov  2 2017 07:20:37.696Z INFO  op=298.6 (delta:1.465334581s): Creating nfs volumestore nfsVolumeStore on nfs://10.161.242.84/store?uid=0&gid=0
Nov  2 2017 07:20:37.696Z DEBUG op=298.6 (delta:1.465339094s): Mounting nfs://10.161.242.84/store?uid=0&gid=0
```
These are the last lines in the portlayer log and the `docker info` command is reattempted 20 times in the test. 

suspected related ticket for a potential fix: https://github.com/vmware/vic/issues/6620

[...]

Issue #6616, opened 2017-10-24
NFS Nightly Failures - Appears that Mount Target Hangs During Container Start Time In Tether


gigawhitlocks:
Seen Oct 23, 21, 20 in Nightlies against 6.0 and 6.5:

10/23, 5-22-NFS-Volume against 6.5:
https://storage.cloud.google.com/vic-ci-logs/functional_logs_10_23_06_09_vic_14150.tar.gz.zip?authuser=1

10/21 against 6.0:
https://storage.cloud.google.com/vic-ci-logs/functional_logs_10_21_06_09_vic_14136.tar.gz.zip?authuser=1

10/20 against 6.5:
https://storage.cloud.google.com/vic-ci-logs/functional_logs_10_20_06_09_vic_14108.tar.gz.zip?authuser=1

matthewavery:
```
Notes from investigation of the NFS failures on 10/21/17 for container ---> keen_wing-3e980cb6d3a12b20048dc9c0e3f55b37b324eeb0f3518c143d785710b274d430

### `personality.log` errors: 

* Oct 22 2017 04:19:55.790Z ERROR Handler for POST /v1.25/containers/3e980cb6d3a12b20048dc9c0e3f55b37b324eeb0f3518c143d785710b274d430/start returned error: Server error from portlayer: unable to wait for process launch status: Post https://10.192.56.31/sdk: context deadline exceeded

### `portlayer.log` errors :

* Oct 22 2017 04:19:55.693Z ERROR CommitHandler error on handle(e798615b9f1f2e75753b6ce092a5102a) for 3e980cb6d3a12b20048dc9c0e3f55b37b324eeb0f3518c143d785710b274d430: unable to wait for process launch status: Post https://10.192.56.31/sdk: context deadline exceeded

* Oct 22 2017 04:19:55.694Z DEBUG Event manager calling back to PLE-829f8190f851b955d7274422e3b9736bb52de1b2c78ef90cddf12480f81ae84f for 0 - events.ContainerEvent: Container 3e980cb6d3a12b20048dc9c0e3f55b37b324eeb0f3518c143d785710b274d430 Failed

### `tether.log` errors :

* Oct 22 2017 04:17:04.470Z ERROR failed to unmount /etc/hostname: invalid argument

* Oct 22 2017 04:17:04.614Z ERROR failed to unmount /etc/resolv.conf: invalid argument

[...]

PR #6802, opened 2017-11-20
Put nfs nightly test infra deployment on the same pod

matthewavery:
This is a PR that has been made in response to recent NFS test failures. Putting all of the Nimbus infra on the same POD eliminates another potential failure situation with the tests connectivity issues. It is not guaranteed to fix all of the problems we are seeing in the NFS nightly tests, but may go a long way toward fixing them and making other kinds of errors more predictable. 

this is in reference to: #6616 and #6677

### DIFF
--- a/tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
+++ b/tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
@@ -37,11 +37,14 @@ Setup ESX And NFS Suite
     Run Keyword And Ignore Error  Nimbus Cleanup  ${list}  ${false}
     Log To Console  \nStarting test...
 
-    ${esx1}  ${esx1_ip}=  Deploy Nimbus ESXi Server  %{NIMBUS_USER}  %{NIMBUS_PASSWORD}
+    ${esx_name}=  Evaluate  'ESX-' + str(time.clock())  modules=time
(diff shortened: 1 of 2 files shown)


matthewavery:
IMPORTANT NOTE: This is subject to POD placement of the esx. Which means that it may fail mid way through due to a POD being marked as busy. We have a retry around the `nimbus` setup. So this test should reasonably be restarted several times and get a good placement for all infra during that period. Until we get the chance to write a correct and well formed testbed this is what we should do. the result of this fact is that the test make take longer than before during high nimbus usage periods(or edge case placement of the of esx)

tests/manual-test-cases/Group5-Functional-Tests/5-22-NFS-Volume.robot
```diff
@@ -38,10 +37,14 @@ Setup ESX And NFS Suite
     Log To Console  \nStarting test...
 
     ${esx1}  ${esx1_ip}=  Deploy Nimbus ESXi Server  %{NIMBUS_USER}  %{NIMBUS_PASSWORD}
+    Open Connection  %{NIMBUS_GW}
+    Wait Until Keyword Succeeds  2 min  30 sec  Login  %{NIMBUS_USER}  %{NIMBUS_PASSWORD}
+    ${POD}=  Fetch POD  ${esx1}
+    Log To Console  ${POD}
```
mhagen-vmware: lgtm, just add a close connection here to cleanup ssh connections.

PR #6808, opened 2017-11-20
[specific ci=Group1-Docker-Commands] Fix tar symlink bug

### LINKED ISSUE #6863: Fix symlink handling for Docker CP offline
Resolve symlink handling in our tar archive code as described in bug 1994971.

gigawhitlocks:
addresses #6863  by performing the unpack operation in docker cp offline inside of a chroot

Fixes #6863


lib/archive/unpack_unix.go
```diff
@@ -0,0 +1,336 @@
...
+			}
+		default:
+			// TODO: add support for special file types - otherwise we will do absurd things such as read infinitely from /dev/random
+		}
+		op.Debugf("Finished writing to: %s", rebased)
+	}
+	return nil
+}
+
+func simpleCopy(src, dst string) error {
```
gigawhitlocks: ignore this helper func, it's unused

lib/archive/unpack_unix.go
```diff
@@ -0,0 +1,336 @@
...
+
+// unpack will unpack the given tarstream(if it is a tar stream) on the local filesystem based on the specified root
+// combined with any rebase from the path spec
+//
+// the pathSpec will include the following elements
+// - include : any tar entry that has a path below(after stripping) the include path will be written
+// - strip : The strip string will indicate the
+// - exlude : marks paths that are to be excluded from the write operation
+// - rebase : marks the the write path that will be tacked onto (appended or prepended? TODO improve this comment) the "root". e.g /tmp/unpack + /my/target/path = /tmp/unpack/my/target/path
+func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec) error {
```
gigawhitlocks: it looks like this whole file is new, because I moved it to unpack_unix.go re: what I was talking about on the phone, but the changes are only:
Unpack renamed to InvokeUnpack
new "Unpack" all new code
a few routines in InvokeUnpack are commented out

isos/bootstrap.sh
```diff
@@ -82,6 +82,7 @@ fi
 
 # copy in our components
 cp ${BIN}/tether-linux $(rootfs_dir $PKGDIR)/bin/tether
+cp ${BIN}/unpack $(rootfs_dir $PKGDIR)/bin/unpack
```
gigawhitlocks: this is also unnecessary; I added it to bootstrap while chasing a bug and need to remove it

isos/appliance.sh
```diff
@@ -124,6 +124,7 @@ chroot $(rootfs_dir $PKGDIR) chmod 0660 /var/run/lock/logrotate_run.lock
 cp ${BIN}/vic-init $(rootfs_dir $PKGDIR)/sbin/vic-init
 
 cp ${BIN}/{docker-engine-server,port-layer-server,vicadmin} $(rootfs_dir $PKGDIR)/sbin/
+cp ${BIN}/unpack $(rootfs_dir $PKGDIR)/bin/
```
gigawhitlocks: will probably move this back to line 126 and put it in /sbin with everything else now that I have the execution worked out

Makefile
```diff
@@ -365,12 +370,12 @@ $(appliance-staging): isos/appliance-staging.sh $(iso-base)
...
 
 # main appliance target - depends on all top level component targets
-$(appliance): isos/appliance.sh isos/appliance/* isos/vicadmin/** $(vicadmin) $(vic-init) $(portlayerapi) $(docker-engine-api) $(appliance-staging)
+$(appliance): isos/appliance.sh isos/appliance/* isos/vicadmin/** $(vicadmin) $(vic-init) $(portlayerapi) $(docker-engine-api) $(appliance-staging) $(archive)
 	@echo building VCH appliance ISO
 	@$(TIME) $< -p $(appliance-staging) -b $(BIN)
 
 # main bootstrap target
-$(bootstrap): isos/bootstrap.sh $(tether-linux) $(bootstrap-staging) isos/bootstrap/*
+$(bootstrap): isos/bootstrap.sh $(archive) $(tether-linux) $(bootstrap-staging) isos/bootstrap/*
```
gigawhitlocks: along with the change to bootstrap.sh i'll strip this build target, which shouldn't be here

lib/archive/unpack_unix.go
```diff
@@ -0,0 +1,200 @@
...
+			// TODO: add support for special file types - otherwise we will do absurd things such as read infinitely from /dev/random
+		}
+		op.Debugf("Finished writing to: %s", absPath)
+	}
+	return nil
+}
+
+func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root string) error {
+	// execute the unpack binary
+	// cmd := exec.Cmd{
```
matthewavery: Is this here just for reference from a previous code path?

gigawhitlocks: ah nope! that's stuff that should've been removed, good eye

lib/archive/unpack_unix.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	encodedFilter, err := EncodeFilterSpec(op, filter)
+	if err != nil {
+		op.Error(err)
+		return err
+	}
+
+	// #nosec: Subprocess launching with variable. -- neither variable is user input & both are bounded inputs so this is fine
+	cmd := exec.Command("/bin/unpack", op.ID(), root, *encodedFilter)
```
matthewavery: Maybe a comment reference to the expected unpack binary we prepare and a short explanation for people coming to this code with no knowledge of why this is needed.

gigawhitlocks: roger that, I'll write some more comments in general

cmd/archive/unpack.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+func main() {
+	ctx := context.Background()
+	op := trace.NewOperation(ctx, "Unpack") // TODO op ID?
+	op.Debugf("New unpack operation created")
+
+	if len(os.Args) != 4 {
+		op.Errorf("Wrong number of arguments passed to unpack binary")
+		os.Exit(5)
```
sflxn: I would give these exit codes constants, defined above.  That will allow others to easily understand the return results.

const (
    InvalidArguments = iota
    InvalidFilterSpec
    ...
)

Make sure to capitalize the first letter so other code can refer to it.

cmd/archive/unpack.go
```diff
@@ -0,0 +1,85 @@
...
+	fi, err := os.Stat(root)
+	if err != nil {
+		// the target unpack path does not exist. We should not get here.
+		op.Errorf("tar unpack target does not exist: %s", root)
+		os.Exit(9)
+	}
+
+	if !fi.IsDir() {
+		err := fmt.Errorf("unpack root target is not a directory: %s", root)
+		op.Error(err)
```
sflxn: you should use op.Errorf() instead of fmt.Errorf() to be consistent with the rest.

gigawhitlocks: ah yeah can do

cmd/archive/unpack.go
gigawhitlocks: sure thing

lib/archive/unpack_unix.go
```diff
@@ -0,0 +1,198 @@
...
+// - strip : The strip string will indicate the
+// - exlude : marks paths that are to be excluded from the write
+// - rebase : marks the the write path that will be tacked onto (appended or prepended? TODO improve this comment) the "root". e.g /tmp/unpack + /my/target/path = /tmp/unpack/my/target/path
+func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root string) error {
+	op.Debugf("unpacking archive to root: %s, filter: %+v", root, filter)
+
+	// Online datasource is sending a tar reader instead of an io reader.
+	// Type check here to see if we actually need to create a tar reader.
+	var tr *tar.Reader
+	if trCheck, ok := tarStream.(*tar.Reader); ok {
```
sflxn: you don't need to use trCheck.  You can just use tr.

if tr, ok := tarStream.(*tar.Reader); !ok {
     tr = tar.NewReader(tarStream)
}

lib/archive/unpack_unix.go
```diff
@@ -0,0 +1,198 @@
...
+	fi, err := os.Stat(root)
+	if err != nil {
+		// the target unpack path does not exist. We should not get here.
+		op.Errorf("tar unpack target does not exist: %s", root)
+		return err
+	}
+
+	if !fi.IsDir() {
+		err := fmt.Errorf("unpack root target is not a directory: %s", root)
+		op.Error(err)
```
sflxn: use op.Errorf() for consistency since you do this everywhere else.

lib/archive/unpack_unix.go
```diff
@@ -0,0 +1,198 @@
...
```
sflxn: op.Errorf()

lib/archive/unpack_unix.go
```diff
@@ -0,0 +1,198 @@
...
+	stdin, err := cmd.StdinPipe()
+
+	if err != nil {
+		op.Error(err)
+		return err
+	}
+
+	if stdin == nil {
+		err = errors.New("stdin was nil")
+		op.Error(err)
```
sflxn: op.Errorf()

lib/archive/unpack_unix.go
```diff
@@ -0,0 +1,198 @@
...
+		return err
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
```
sflxn: go func() error {
   ...
   return err
}

sflxn: Instead of waitgroup, use a channel.

done := make(chan error)
go func() {
     ...
     done <- err
}()

lib/archive/unpack_unix.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+	}()
+
+	// copy the tarStream to the binary via stdin; the binary will stream it to InvokeUnpack unchanged
+	if _, err := io.Copy(stdin, tarStream); err != nil {
+		op.Errorf("Error copying tarStream: %s", err.Error())
+		return err
+	}
+
+	wg.Wait()
```
sflxn: If you're waiting, then you're expecting some status from the go function, but you're not getting it.  If you use the channel as I described above, you can do,

err = <-done

This will wait and allow you to retrieve the result of the command you launched.

lib/archive/unpack_unix.go
sflxn: Well, the comments I made above requires that ok be already defined... so you can either predefine ok or just stick with the current version.

gigawhitlocks: I'm going to leave this as-is because this is actually the old unpack code

lib/archive/unpack_unix.go
gigawhitlocks: ah ok cool :+1:

gigawhitlocks:
Yes that's an idiomatic error that stems from my relative inexperience w/ channels -- I will fix this and I will note the idiom that I didn't know for future usage as well.

gigawhitlocks:
#mentored

lib/archive/unpack_unix.go
gigawhitlocks: I want to return an error here, which is why I use `fmt.Errorf` to create a new error

lib/archive/unpack_unix.go
gigawhitlocks: same here, I don't just want to print the error to the op log, but also return it. Hence creating a new error, since one does not exist

lib/archive/unpack_unix.go
gigawhitlocks: Everywhere else, there is already an error to return.

sflxn:
Please update the resolved issue to #6863 as I've converted the original issue to an epic.

lib/archive/unpack_unix.go
```diff
@@ -0,0 +1,196 @@
...
+			op.Debugf("%s", string(out))
+		}
+		if err != nil {
+			op.Errorf("Command returned error %s", err.Error())
+		}
+		done <- err
+	}()
+
+	// copy the tarStream to the binary via stdin; the binary will stream it to InvokeUnpack unchanged
+	if _, err := io.Copy(stdin, tarStream); err != nil {
```
hickeng: I think you may want to invert which part of this is backgrounded and have `io.Copy` run in the goroutine instead of `cmd.CombinedOutput`.

It's possible for this function to return and process to exit before the unpack command has completed if there's an error from tarStream. I'm unsure what that means for the process i/o buffers for `/bin/unpack` but pretty sure it will become orphaned.

We should be closing stdin irrespective of `io.Copy` return - `io.Copy` does not close the destination when it receives EOF from the source. This might not be an issue with a well formed archive because I suspect that unpack will exit when it sees the nil record terminator, however if there is a malformed header or missing file data after header then unpack will likely block indefinitely.

gigawhitlocks: >It's possible for this function to return and process to exit before the unpack command has completed if there's an error from tarStream.

Yeah I see that now.. you're right, I'll address this as suggested

lib/archive/unpack_unix.go
gigawhitlocks: also creating an error here where one does not exist, same as above. need `fmt.Errorf` to create the error as `op.Errorf` does not return anything

lib/archive/unpack_unix.go
```diff
@@ -0,0 +1,194 @@
...
+		op.Error(err)
+		return err
+	}
+	done := make(chan error)
+	go func() {
+		// copy the tarStream to the binary via stdin; the binary will stream it to InvokeUnpack unchanged
+		if _, err := io.Copy(stdin, tarStream); err != nil {
+			op.Errorf("Error copying tarStream: %s", err.Error())
+		}
+		done <- err
```
hickeng: I think we need a `stdin.Close()` call - if we have a malformed header that's expecting more data than is present then this may well block indefinitely if `unpack` binary is left waiting for data that will never come.

gigawhitlocks:
Just waiting on @hickeng at this point (I think his comments are addressed) but I'm not nagging him since this is missing 1.3

mdubya66:
where are we with this @gigawhitlocks. let's get it into master asap

lib/archive/unpack_unix.go
```diff
@@ -0,0 +1,195 @@
...
+		return err
+	}
+	done := make(chan error)
+	go func() {
+		// copy the tarStream to the binary via stdin; the binary will stream it to InvokeUnpack unchanged
+		defer stdin.Close()
+		if _, err := io.Copy(stdin, tarStream); err != nil {
+			op.Errorf("Error copying tarStream: %s", err.Error())
+		}
+		done <- err
```
hickeng: This `err` is the value of err from _outside_ of the go routine. I don't know if that's the intent, but if it is then line 193 is effectively:
```
 err = err
```

I suspect you actually want to return the `err` from CombinedOutput as that will be the one with NotExist errors or similar, but the way it's currently written is unclear. If I'm correct then you could simply use the following in the goroutine:
```
close(done)
```

gigawhitlocks:
@mhagen-vmware

gigawhitlocks:
Need approval from `build-ci-infra`

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.robot
```diff
@@ -57,6 +57,28 @@ Clean up test files and VIC appliance to test server
...
 *** Test Cases ***
+Try To Exploit VCH With Offline Copy of Malicious Tarball
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull ${busybox}
+    Should Be Equal As Integers  ${rc}  0
+
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} create --name exploitme ${busybox}
+    Should Be Equal As Integers  ${rc}  0
+
+    ${rc}  ${output}=  Run And Return Rc And Output  ls ${CURDIR}/*.tar.gz
+    Should Be Equal As Integers  ${rc}  0
```
mhagen-vmware: don't really need this it is presumably handled by the No such file check below

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.robot
```diff
@@ -57,6 +57,28 @@ Clean up test files and VIC appliance to test server
...
+
+    ${rc}  ${output}=  Run And Return Rc And Output  cat ${CURDIR}/archive.tar.gz | docker %{VCH-PARAMS} cp - exploitme:/
```
mhagen-vmware: please move this to resources folder instead of within the Docker1 group

tests/test-cases/Group1-Docker-Commands/1-43-Docker-CP-Offline.robot
gigawhitlocks: oh, yeah, good catch this is leftover from testing my test (making sure I had put the archive in the right place and that the tests could see it etc)

gigawhitlocks:
just waiting on master to be unlocked 😿

PR #6809, opened 2017-11-20
Always update software in vic-machine-server image

zjs:
To ensure we always ship the most up-to-date version of packages with the vic-machine-server image, perform a distro-sync as a part of the image build.

To reduce bloat, clean up up temporary files, data, and metadata.

In between, print information about installed packages for tracking.


morris-jason:
I'd prefer to see the 'RUN set -eux' pattern here: https://github.com/vmware/vic-product/blob/master/installer/build/Dockerfile

PR #6810, opened 2017-11-20
Add retry logic to image store creation

### LINKED ISSUE #4858: Nightly 6.5 HA test fails during rmi busybox

cgtexmex:
Retry logic added to image store creation.  This was needed
as we were seeing failures during and HA event that were the
result of host communication issues.  The image store will
retry when the host it's communicating with is unvailable.

This should reduce one of the more common failure modes
seen while working on #4666. We _should_ confirm with
vSphere what is expected in Host HA/failover scenarios.

Fixes #4858

### DIFF
--- a/lib/portlayer/storage/image_cache.go
+++ b/lib/portlayer/storage/image_cache.go
@@ -21,12 +21,15 @@ import (
 	"net/url"
 	"os"
 	"sync"
+	"time"
 
 	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/vic/lib/archive"
 	"github.com/vmware/vic/lib/portlayer/util"
 	"github.com/vmware/vic/pkg/index"
+	"github.com/vmware/vic/pkg/retry"
 	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/tasks"
 	"github.com/vmware/vic/pkg/vsphere/vm"
 )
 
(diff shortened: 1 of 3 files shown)


lib/portlayer/storage/image_cache.go
```diff
@@ -153,15 +156,35 @@ func (c *NameLookupCache) CreateImageStore(op trace.Operation, storeName string)
...
 		return nil, os.ErrExist
 	}
-
-	if !os.IsNotExist(err) {
-		op.Warnf("Error getting image store %s: %s", storeName, err)
+	// log and return any error besides NotExist -- if the image store
+	// doesn't exist it will be created below
+	if err != nil && err != os.ErrNotExist {
+		op.Errorf("Error getting image store %s: %s", storeName, err)
+		return nil, err
```
hickeng: If we return here then we will not have the chance to trigger the (sort of) cleanup on line 202. I also seem to recall there being logic to remove inconsistent or incomplete images in either the `GetImageStore` or `CreateImageStore` paths.

Can we confirm that we're not skipping repair logic by returning here and what the behaviour of the portlayer is if we do error out at this point.

cgtexmex: I've changed it so that it will continue on the following two errors:
- ErrNotExist
- ErrCorruptImageStore
Those two will attempt to be corrected by the rest of the function.

cgtexmex:
@hickeng I've been unable to find any self correcting behavior in the image store creation.  Specifically if we hit the `ErrCorruptImageStore` I've not seen the portLayer be able to heal itself and recover.  I create the scenario by manually removing the `scratch` folder from the `datastore` and then restart the VCH.  The `ErrCorruptImageStore` is encountered, but the VCH is dead.

AFAIK that's been the behavior in that scenario all along...possibly I don't understand the scenario that brought about the creation of `ErrCorruptImageStore`.  Thoughts?

hickeng:
@cgtexmex I'll have to dig through code for specifics, but in the interim and iirc - thought there was logic that would delete a partially there image. So if manifest is missing for example but the vmdk is there it would clean up.
That logic _might_ be in `WriteImage` or `GetImage` rather than at the store level.

PR #6812, opened 2017-11-20
[specific ci=Group23-VIC-Machine-Service] VCH creation logging flushes all data before exiting

### LINKED ISSUE #6732: Ensure VCH creation log streaming flushes all data before exiting
See https://github.com/vmware/vic/pull/6665#discussion_r149552286:
> Do we also want a Flush call that waits until all data written has been read? Currently I think it's possible for `vic-machine` to exit without having pushed all of the log data to the datastore, depending on timing.
> I would not be surprised to see truncation of the last informational portions of the output.

And https://github.com/vmware/vic/pull/6665#discussion_r149551715:
> The way this is written there's nothing preventing multiple concurrent calls to `Read` which would leave all but one thread blocked indefinitely.
> 
> I think both `Write` and `Close` should be using cond.Broadcast(). If we're being good then Read() should also have a deferred Broadcast so that one Read completing will allow another to continue - this model also allows for addition of a Flush semantic.

And https://github.com/vmware/vic/pull/6665#discussion_r149553610:
> There is a race here - we could have called `bp.Close()` before the `bp.Read` takes place. I'm not sure it can be resolved and still have the test pass reliably without changing Close to a flushing semantic, i.e. Closed to writing, but allows read to continue to drain existing data.

AngieCris:
Fixes #6732 

Upon calling `bp.Close()` on the logger pipe, write end closes first. All subsequent writes fail. If there's left-over data in the pipe, `bp.Close()` blocks and waits until all data is drained.


lib/install/vchlog/buffered_pipe_test.go
```diff
@@ -205,6 +205,41 @@ func TestPipeWriteClose(t *testing.T) {
...
+	bp := NewBufferedPipe()
+	writeDone := make(chan int)
+
+	go write(t, bp, testByte, len(testByte), writeDone)
+	<-writeDone
+	go bp.Close()
+
+	// wait for write end to close
+	// closing of read end is expected to block until data is drained from the pipe
+	time.Sleep(1 * time.Millisecond)
```
hickeng: Both here and below should really be `until` logic - for simplicity the global test timeout can be used to avoid something here (I did not see a handy testify until mechanism).
As it stands it will probably work, until run on a heavily loaded system and 1ms is not long enough for the other goroutine to be scheduled - for example if running unit tests in parallel.

AngieCris: Thanks for pointing that out. I think there's a better way around to fix this: instead of waiting or timing out, it could continuously write to the pipe until the write fails, while the close() is running in the background in a go routine. Then, after the write has failed, begin reading on the left-over data until exhausted. I can keep a count of the number of bytes written before failing and check if the read fails after reading the exact number of bytes. Fixes coming up.

lib/install/vchlog/buffered_pipe.go
```diff
@@ -37,24 +41,28 @@ func NewBufferedPipe() *BufferedPipe {
...
 	bp.c.L.Lock()
 	defer bp.c.L.Unlock()
+	defer bp.c.Broadcast()
 
-	// pipe closed, drop all left-over data
-	if bp.closed {
-		return 0, io.EOF
-	}
-	for bp.buffer.Len() == 0 && !bp.closed {
+	bp.readerReady = true // the reader is ready to read
```
hickeng: I'd update the comment to something like: `// We have a valid consumer so switch to flushing in Close`

lib/install/vchlog/buffered_pipe.go
```diff
@@ -23,11 +23,15 @@ import (
 // BufferedPipe struct implements a pipe readwriter with buffer
 // buffer: the internal buffer to hold data
 // c: the sync locker to manage concurrent reads and writes
-// closed: boolean indicating if the stream is closed
+// readClosed: boolean indicating if read end of the pipe is closed
+// writeClosed: boolean indicating if write end of the pipe is closed
+// readerReady: boolean indicating if the reader is ready
```
jzt: Minor: normally these descriptive comments go above each field in the struct (but aren't necessary if the field isn't exported).

lib/install/vchlog/buffered_pipe.go
```diff
@@ -37,24 +41,28 @@ func NewBufferedPipe() *BufferedPipe {
 	var m sync.Mutex
 	c := sync.NewCond(&m)
 	return &BufferedPipe{
-		buffer: bytes.NewBuffer(nil),
-		c:      c,
-		closed: false,
+		buffer:      bytes.NewBuffer(nil),
+		c:           c,
+		readClosed:  false,
```
jzt: The zero-value for booleans is `false`, so if you leave `readClosed`, `writeClosed` and `readerReader` out of this instantiation, they will init to `false` automatically. Feel free to leave it like this if you are doing it to be explicit/verbose!

lib/install/vchlog/buffered_pipe.go
```diff
@@ -37,24 +41,28 @@ func NewBufferedPipe() *BufferedPipe {
...
 	bp.c.L.Lock()
 	defer bp.c.L.Unlock()
+	defer bp.c.Broadcast()
 
-	// pipe closed, drop all left-over data
-	if bp.closed {
-		return 0, io.EOF
-	}
-	for bp.buffer.Len() == 0 && !bp.closed {
+	bp.readerReady = true // now we have a valid consumer to switch to flushing in Close
```
zjs: What happens if we had a reader at one point, but no longer have one?

This shouldn't happen with the way this code is currently used, but in theory I think something like this could happen:
```
bp = NewBufferedPipe()
bp.Write([]byte("Hello World"))
b = make([]byte, 5)
bp.Read(b) // Read "Hello", caused readerReady = true
bp.Close() // Waits forever for someone to read " World"
```

Maybe this isn't a case we can reasonably handle? (Or maybe we handle this with some sort of timeout on the flush?

AngieCris: Yes, it would probably be a case we can't reasonably handle. The read is from the http connection established by govc with the datastore, so the early abortion of read only happens if the datastore connection is cut down when consuming the logs. In this case, `datastore.Upload` will return an error. However I'm not sure how to treat this connection failure reasonably, it could be the problem with VCH, or the problem from client side.
A possible way to handle this is that, if upload returns an error in `vchlogger.go`, we send a signal to the underlying pipe telling it to drop any leftover messages.

lib/install/vchlog/buffered_pipe.go
AngieCris: Thanks! I just feel like it's more explicit this way, to let code reader know that the pipe is default open and we're waiting for a reader to ready

lib/install/vchlog/buffered_pipe.go
AngieCris: In general usage, it's hard to tell if someone is still calling read() continuously. If someone calls read() at one point and stop, the pipe won't know about this (pipe isn't keeping track of the status of caller). Adding a timeout would be a possible workaround

hickeng: Ironically adding a `WriteTo` method would probably allow us to resolve this, but in a manner that's not consistent with the pure Reader behaviour.
WriteTo shouldn't return at all until the copy has completely finished, and you would get an error from the writer that can be used to purge any remaining data.

lib/install/vchlog/buffered_pipe.go
```diff
@@ -76,7 +87,26 @@ func (bp *BufferedPipe) Write(data []byte) (n int, err error) {
...
+
+	if bp.readerReady { // flush only if there is a valid consumer
+		done := make(chan bool)
+		for bp.buffer.Len() > 0 {
+			go func() {
+				bp.c.Wait()
+				done <- true
+			}()
+			select {
+			case <-time.After(time.Minute * 30): // timeout if consumer inactive for 30 minutes
```
hickeng: I don't like timeouts - they just cause pain, but....
30 min seems like a long time given the amount of data we're trying to move and the expected duration of the operation. This blocks Close from returning - does that prevent the API request from completing?

I'm wondering if we should chose an equally horrific approach of creating the pipe with a context, saving that into the structure and then waiting for that to time out or be cancelled. At least then we're exposed to cancellation by way of the client http connection dropping.

AngieCris: Yes, the timeout does prevent the API request from returning. The VCH will still be created, but the close hangs so the API hangs as well. Also I'm not sure what would that mean for the vic-machine service and other processes running on the service

I'm not sure if it makes sense for a pipe stream to have its caller context. Maybe it's vchlogger who needs to have the context and keep track of context timeouts?

AngieCris:
With lots of help from @hickeng , the new proposed solution is:
1. Instead of blocking on `buffered_pipe.Close()` waiting for consumer on remaining data, leverage the context timeout on CLI. Have the vic-machine create CLI wait for the log streaming to complete, or time out if context deadline is exceeded.
2. `buffered_pipe.Read()` only quits with EOF if the buffer is empty and the write end is closed. On API side, after the VCH creation ends and pipe is closed, log streaming will still run in the background reading from the pipe, if there's remaining data. This read will not be blocked.

cmd/vic-machine/create/create.go
```diff
@@ -767,5 +767,16 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 	executor.ShowVCH(vchConfig, c.Certs.Ckey, c.Certs.Ccert, c.Certs.Cacert, c.Certs.EnvFile, c.Certs.CertPath)
 	op.Info("Installer completed successfully")
 
+	// wait on the logger to finish streaming
+	go func() {
+		select {
+		case <-time.After(time.Second * 3):
```
hickeng: Just realized that we don't need the `select` statement here at all:
```
go func() {
   // let the user know why they're waiting if it causes a noticeable delay
   time.Sleep(3 * time.Second)
   op.Infof("Waiting for log upload to complete")
}

Comment (AngieCris) [reply cmd/vic-machine/create/create.go]: good point, thanks

File: cmd/vic-machine/create/create.go [hunk 8]
```diff
@@ -767,5 +767,16 @@ func (c *Create) Run(clic *cli.Context) (err error) {
 	executor.ShowVCH(vchConfig, c.Certs.Ckey, c.Certs.Ccert, c.Certs.Cacert, c.Certs.EnvFile, c.Certs.CertPath)
 	op.Info("Installer completed successfully")
 
+	// wait on the logger to finish streaming
+	go func() {
```
zjs: What happens to this goroutine if we return before this is reached?

Should this be taking `op` as an argument and have a `case <-op.Done():` so that this gets cleaned up if `op` is canceled (which I _think_ happens just before we exit)?

Issue #6803, opened 2017-11-20
vic-machine Dispatcher/Validator tech debt


andrewtchin:
**User Statement:**

As a person I want to address vic-machine tech debt in the Dispatcher/Validator implementation

**Details:**
WIP

**Acceptance Criteria:**
WIP

PR #6815, opened 2017-11-21
Check vCenter options in vic-machine

### LINKED ISSUE #6473: vic-machine: set vpxd.SerialPort.PersistNetworkBacking to true

andrewtchin:
Fixes #6473
Replacement PR off of upstream instead

If the value is set correctly:
```
⇒  ./bin/vic-machine-linux create --target 'administrator@vsphere.local:Admin!23@10.160.96.27' --name d --no-tlsverify --compute-resource /ha-datacenter/host/cluster1 --image-store datastore1 --bridge-network bridge --thumbprint=93:17:0D:D6:C5:C7:58:EC:33:66:69:2C:13:E5:CD:BA:4C:30:34:29
Nov 19 2017 12:33:11.342-08:00 INFO  ### Installing VCH ####
Nov 19 2017 12:33:11.342-08:00 WARN  Using administrative user for VCH operation - use --ops-user to improve security (see -x for advanced help)
Nov 19 2017 12:33:11.344-08:00 INFO  Generating self-signed certificate/key pair - private key in d/server-key.pem
Nov 19 2017 12:33:11.469-08:00 WARN  Configuring without TLS verify - certificate-based authentication disabled
Nov 19 2017 12:33:15.418-08:00 INFO  Validating supplied configuration
...
Nov 19 2017 12:33:21.955-08:00 INFO  vCenter settings check OK
...
```
If the value is unset or not set to the correct value it is set to "true":
```
⇒  ./bin/vic-machine-linux create --target 'administrator@vsphere.local:Admin!23@10.160.96.27' --name e --no-tlsverify --compute-resource /ha-datacenter/host/cluster1 --image-store datastore1 --bridge-network bridge --thumbprint=93:17:0D:D6:C5:C7:58:EC:33:66:69:2C:13:E5:CD:BA:4C:30:34:29
Nov 19 2017 12:35:27.268-08:00 INFO  ### Installing VCH ####
Nov 19 2017 12:35:27.268-08:00 WARN  Using administrative user for VCH operation - use --ops-user to improve security (see -x for advanced help)
Nov 19 2017 12:35:27.268-08:00 INFO  Generating self-signed certificate/key pair - private key in e/server-key.pem
Nov 19 2017 12:35:27.355-08:00 WARN  Configuring without TLS verify - certificate-based authentication disabled
Nov 19 2017 12:35:30.885-08:00 INFO  Validating supplied configuration
Nov 19 2017 12:35:35.487-08:00 INFO  vDS configuration OK on "bridge"
...
Nov 19 2017 12:35:37.714-08:00 INFO  Set vCenter settings for containerVMs
Nov 19 2017 12:35:37.714-08:00 INFO  vCenter settings check OK
...

```

### DIFF
--- a/lib/install/validate/validator.go
+++ b/lib/install/validate/validator.go
@@ -927,7 +927,7 @@ func (v *Validator) ConfigureVCenter(ctx context.Context) error {
 		return nil
 	}
 	if !v.IsVC() {
-		op.Info(errMsg)
+		op.Debug(errMsg)
 		return nil
 	}
 
--- a/lib/install/validate/validator.go
+++ b/lib/install/validate/validator.go
@@ -44,6 +44,7 @@ import (
 	"github.com/vmware/vic/pkg/registry"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/version"
+	"github.com/vmware/vic/pkg/vsphere/optmanager"
 	"github.com/vmware/vic/pkg/vsphere/session"
 )
 
@@ -309,7 +310,18 @@ func (v *Validator) Validate(ctx context.Context, input *data.Data) (*config.Vir
 	v.compute(op, input, conf)
 	v.storage(op, input, conf)
 	v.network(op, input, conf)
+	// FIXME ATC DEBT setting this value needs to be moved to Dispatcher
+	// https://github.com/vmware/vic/issues/6803
+	ok := v.CheckPersistNetworkBacking(op, true)
+	if !ok {
+		err := v.ConfigureVCenter(op)
+		if err != nil {
+			op.Errorf("%s", err)
+			op.Errorf("vCenter settings update FAILED")
+		}
+	}
 	v.CheckFirewall(op, conf)
+	v.CheckPersistNetworkBacking(op, false)
 	v.CheckLicense(op)
 	v.CheckDrs(op)
 
@@ -901,4 +913,29 @@ func (v *Validator) syslog(op trace.Operation, conf *config.VirtualContainerHost
 		Network: network,
 		RAddr:   host,
 	}
+}
+
+// FIXME ATC DEBT setting this value needs to be moved to Dispatcher
+// https://github.com/vmware/vic/issues/6803
+// set PersistNetworkBacking key to "true"
+func (v *Validator) ConfigureVCenter(ctx context.Context) error {
+	op := trace.FromContext(ctx, "Set vCenter serial port backing")
+	defer trace.End(trace.Begin("", op))
+
+	errMsg := "Set vCenter settings SKIPPED"
+	if !v.sessionValid(op, errMsg) {
+		return nil
+	}
+	if !v.IsVC() {
+		op.Info(errMsg)
+		return nil
+	}
+
+	err := optmanager.UpdateOptionValue(ctx, v.Session, persistNetworkBackingKey, "true")
+	if err != nil {
+		msg := fmt.Sprintf("Failed to set required value \"true\" for %s: %s", persistNetworkBackingKey, err)
+		return errors.New(msg)
+	}
+	op.Infof("Set vCenter settings OK")
+	return nil
 }
--- a/lib/install/validate/config.go
+++ b/lib/install/validate/config.go
@@ -29,7 +29,10 @@ import (
 	"github.com/vmware/vic/lib/constants"
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/optmanager"
 )
+
+const persistNetworkBackingKey = "config.vpxd.SerialPort.PersistNetworkBacking"
 
 type FirewallStatus struct {
 	Rule                          types.HostFirewallRule
@@ -561,4 +564,35 @@ func (v *Validator) CheckDrs(ctx context.Context) {
 	}
 	op.Info("DRS check OK on:")
 	op.Infof("  %q", v.Session.Cluster.InventoryPath)
+}
+
+// check that PersistNetworkBacking is set
+func (v *Validator) CheckPersistNetworkBacking(ctx context.Context, quiet bool) bool {
+	op := trace.FromContext(ctx, "Check vCenter serial port backing")
+	defer trace.End(trace.Begin("", op))
+
+	errMsg := "vCenter settings check SKIPPED"
+	if !v.sessionValid(op, errMsg) {
+		return false
+	}
+	if !v.IsVC() {
+		op.Info(errMsg)
+		return true
+	}
+
+	val, err := optmanager.QueryOptionValue(ctx, v.Session, persistNetworkBackingKey)
+	if err != nil {
+		// the key is not set
+		val = "false"
+	}
+	if val != "true" {
+		if !quiet {
+			op.Errorf("vCenter settings check FAILED")
+			msg := fmt.Sprintf("vCenter advanced option %s=true must be set", persistNetworkBackingKey)
+			v.NoteIssue(errors.New(msg))
+		}
+		return false
+	}
+	op.Infof("vCenter settings check OK")
+	return true
 }
--- a/lib/install/validate/config.go
+++ b/lib/install/validate/config.go
@@ -576,7 +576,7 @@ func (v *Validator) CheckPersistNetworkBacking(ctx context.Context, quiet bool) 
 		return false
 	}
 	if !v.IsVC() {
-		op.Info(errMsg)
+		op.Debug(errMsg)
 		return true
 	}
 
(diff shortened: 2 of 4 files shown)


lib/install/validate/validator.go
```diff
@@ -902,3 +914,28 @@ func (v *Validator) syslog(op trace.Operation, conf *config.VirtualContainerHost
...
+func (v *Validator) ConfigureVCenter(ctx context.Context) error {
+	op := trace.FromContext(ctx, "Set vCenter serial port backing")
+	defer trace.End(trace.Begin("", op))
+
+	errMsg := "Set vCenter settings SKIPPED"
+	if !v.sessionValid(op, errMsg) {
+		return nil
+	}
+	if !v.IsVC() {
+		op.Info(errMsg)
```
hickeng: I don't think we should even bother telling the user about things we didn't do that we shouldn't have done.
Same for earlier Info log in the same condition. Fine for it to be Debug.

lib/install/validate/validator.go
```diff
@@ -902,3 +914,28 @@ func (v *Validator) syslog(op trace.Operation, conf *config.VirtualContainerHost
...
+		return nil
+	}
+
+	err := optmanager.UpdateOptionValue(ctx, v.Session, persistNetworkBackingKey, "true")
```
hickeng: I thought that we found we needed to omit `config` from the key for it to be set, and that prefix magically got appended, or was it something else that was inhibiting it from being updated?

lib/install/validate/validator.go
andrewtchin: fixed

lib/install/validate/validator.go
andrewtchin: that was reading the key when it was unset both with and without `config` would give an error so now we assume that if there's an error it's unset and go ahead and set the value

dougm:
We can add UpdateOptions support to vcsim, the patch below will work in the meantime (see also https://godoc.org/github.com/vmware/govmomi/simulator#example-package)

```diff
diff --git a/lib/install/validate/validator_test.go b/lib/install/validate/validator_test.go
index 6a56308c..0557328c 100644
--- a/lib/install/validate/validator_test.go
+++ b/lib/install/validate/validator_test.go
@@ -34,6 +34,7 @@ import (
 	"github.com/vmware/govmomi/simulator/esx"
 	"github.com/vmware/govmomi/vim25/methods"
 	"github.com/vmware/govmomi/vim25/mo"
+	"github.com/vmware/govmomi/vim25/soap"
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/lib/install/data"
@@ -584,6 +585,19 @@ func testStorage(v *Validator, input *data.Data, conf *config.VirtualContainerHo
 	}
 }
 
+// TODO: vcsim should support UpdateOptions
+type testOptionManager struct {
+	*simulator.OptionManager
+}
+
+func (m *testOptionManager) UpdateOptions(req *types.UpdateOptions) soap.HasFault {
+	m.Setting = append(req.ChangedValue, m.Setting...)
+
+	return &methods.UpdateOptionsBody{
+		Res: new(types.UpdateOptionsResponse),
+	}
+}
+
 func TestValidateWithFolders(t *testing.T) {
 	log.SetLevel(log.InfoLevel)
 	op := trace.NewOperation(context.Background(), "TestValidateWithFolders")
@@ -606,6 +620,9 @@ func TestValidateWithFolders(t *testing.T) {
 	s := m.Service.NewServer()
 	defer s.Close()
 
+	om := simulator.Map.Get(*m.ServiceContent.Setting).(*simulator.OptionManager)
+	simulator.Map.Put(&testOptionManager{om})
+
 	input := data.NewData()
 	input.URL = &url.URL{
 		Scheme: s.URL.Scheme,
```

lib/install/validate/validator.go
```diff
@@ -902,3 +914,28 @@ func (v *Validator) syslog(op trace.Operation, conf *config.VirtualContainerHost
...
+// set PersistNetworkBacking key to "true"
+func (v *Validator) ConfigureVCenter(ctx context.Context) error {
+	op := trace.FromContext(ctx, "Set vCenter serial port backing")
+	defer trace.End(trace.Begin("", op))
+
+	errMsg := "Set vCenter settings SKIPPED"
+	if !v.sessionValid(op, errMsg) {
+		return nil
+	}
+	if !v.IsVC() {
```
anchal-agrawal: Minor: can this check be moved above the session valid check? If the target isn't vCenter we can exit quickly that way.

lib/install/validate/validator.go
```diff
@@ -902,3 +914,28 @@ func (v *Validator) syslog(op trace.Operation, conf *config.VirtualContainerHost
...
+// FIXME ATC DEBT setting this value needs to be moved to Dispatcher
+// https://github.com/vmware/vic/issues/6803
+// set PersistNetworkBacking key to "true"
+func (v *Validator) ConfigureVCenter(ctx context.Context) error {
+	op := trace.FromContext(ctx, "Set vCenter serial port backing")
+	defer trace.End(trace.Begin("", op))
+
+	errMsg := "Set vCenter settings SKIPPED"
+	if !v.sessionValid(op, errMsg) {
+		return nil
```
anchal-agrawal: Should we still return nil here?

lib/install/validate/validator.go
```diff
@@ -902,3 +914,28 @@ func (v *Validator) syslog(op trace.Operation, conf *config.VirtualContainerHost
 		RAddr:   host,
 	}
 }
+
+// FIXME ATC DEBT setting this value needs to be moved to Dispatcher
+// https://github.com/vmware/vic/issues/6803
+// set PersistNetworkBacking key to "true"
+func (v *Validator) ConfigureVCenter(ctx context.Context) error {
```
anchal-agrawal: `ConfigureVCenter` can be left unexported - are we planning on using it somewhere else?

lib/install/validate/validator.go
andrewtchin: yes its just skipped and sessionValid handles the error message and noteError

lib/install/validate/validator.go
hickeng: If the session isn't valid I don't think we can get that information.

lib/install/validate/validator.go
```diff
@@ -309,7 +310,18 @@ func (v *Validator) Validate(ctx context.Context, input *data.Data) (*config.Vir
 	v.compute(op, input, conf)
 	v.storage(op, input, conf)
 	v.network(op, input, conf)
+	// FIXME ATC DEBT setting this value needs to be moved to Dispatcher
+	// https://github.com/vmware/vic/issues/6803
+	ok := v.CheckPersistNetworkBacking(op, true)
+	if !ok {
+		err := v.ConfigureVCenter(op)
```
hickeng: should this be:
```
err := v.ConfigureVCenter(op)
v.NoteIssue(err)
```
That would trigger a failure in the case that the update fails: https://vmware.slack.com/archives/D0G98MP88/p1511121034000100

andrewtchin: that fails to set then we check again on 324 with quiet=false which will note issue here
https://github.com/vmware/vic/pull/6815/files/c333627948db8ef82ed9b895ab2502d95ecff37c#diff-1e16518ca9e1f0254fdf02fd7b7bd502R591

hickeng: ok - non-obvious for me, but we'll be moving this out of validator soon enough anyway.
Can you open a follow up issue for the move to Dispatcher and then this can be merged.

andrewtchin: 👍 https://github.com/vmware/vic/issues/6803

PR #6816, opened 2017-11-21
Prevent vic-machine from deleting non cVMs

### LINKED ISSUE #6603: vic-machine delete can result in non-containerVMs being deleted

andrewtchin:
Fixes #6603 
Replacement PR for #6679 to get the build to build things

### DIFF
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -91,6 +91,24 @@ func (d *Dispatcher) isVCH(vm *vm.VirtualMachine) (bool, error) {
 	return false, nil
 }
 
+func (d *Dispatcher) isContainerVM(vm *vm.VirtualMachine) (bool, error) {
+	if vm == nil {
+		return false, errors.New("nil parameter")
+	}
+	defer trace.End(trace.Begin(vm.InventoryPath))
+	var cspec executor.ExecutorConfig
+	info, err := vm.FetchExtraConfig(d.op)
+	if err != nil {
+		err = errors.Errorf("Failed to fetch guest info of appliance vm: %s", err)
+		return false, err
+	}
+	extraconfig.Decode(extraconfig.MapSource(info), &cspec)
+	if cspec.Version == nil {
+		return false, nil
+	}
+	return true, nil
+}
+
 func (d *Dispatcher) checkExistence(conf *config.VirtualContainerHostConfigSpec, settings *data.InstallerData) error {
 	defer trace.End(trace.Begin(""))
 
--- a/tests/test-cases/Group6-VIC-Machine/6-03-Delete.robot
+++ b/tests/test-cases/Group6-VIC-Machine/6-03-Delete.robot
@@ -16,7 +16,7 @@
 Documentation  Test 6-03 - Verify delete clean up all resources
 Resource  ../../resources/Util.robot
 Test Setup  Install VIC Appliance To Test Server
-Test Teardown  Run Keyword If Test Failed  Cleanup VIC Appliance On Test Server
+Test Teardown  Run Keyword If Test Failed  Cleanup Delete Tests
 Test Timeout  20 minutes
 
 *** Keywords ***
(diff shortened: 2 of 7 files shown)


lib/install/management/appliance.go
```diff
@@ -91,6 +91,24 @@ func (d *Dispatcher) isVCH(vm *vm.VirtualMachine) (bool, error) {
 	return false, nil
 }
 
+func (d *Dispatcher) isContainerVM(vm *vm.VirtualMachine) (bool, error) {
+	if vm == nil {
```
vburenin: Is it possible that anybody pass a nil value here? Asking, because of excessive check is actually bad for readability and maintainability.

tests/test-cases/Group6-VIC-Machine/6-03-Delete.robot
```diff
@@ -95,3 +104,125 @@ Attach Disks and Delete VCH
...
+    Should Be Equal As Integers  ${rc}  0
+
+    # Delete with force
+    ${moid}=  Get VM Moid  %{VCH-NAME}
+    ${ret}  ${output}=  Run And Return Rc And Output  bin/vic-machine-linux delete --target %{TEST_URL} --user %{TEST_USERNAME} --password=%{TEST_PASSWORD} --compute-resource=%{TEST_RESOURCE} --id ${moid} --force
+    Log  ${output}
+    Should Contain  ${output}  Completed successfully
+
+    # Verify VM exists
+    ${rc}  ${output}=  Run And Return Rc And Output  govc ls /ha-datacenter/vm/${dummyvm}
```
mhagen-vmware: that is going to break on vc... I think. I think you don't need ha-datacenter for the ls call, just let it assume path and do govc ls vm/${...}

tests/test-cases/Group6-VIC-Machine/6-03-Delete.robot
```diff
@@ -95,3 +104,125 @@ Attach Disks and Delete VCH
...
+    Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Pass Execution  Test skipped on ESX due to unable to move into RP
+    ${rand}=  Generate Random String  15
+    ${dummyvm}=  Set Variable  anothervm-${rand}
+    Set Suite Variable  ${tempvm}  ${dummyvm}
+    Log To Console  Create VM ${dummyvm} in %{TEST_RESOURCE}/%{VCH-NAME} net %{PUBLIC_NETWORK}
+    ${rc}  ${output}=  Run And Return Rc And Output  govc vm.create -pool="%{TEST_RESOURCE}" -net=%{PUBLIC_NETWORK} -on=false ${dummyvm}
+    Should Be Equal As Integers  ${rc}  0
+
+    # Verify VM exists
+    ${rc}  ${output}=  Run And Return Rc And Output  govc ls /ha-datacenter/vm/${dummyvm}
```
mhagen-vmware: same as below

tests/test-cases/Group6-VIC-Machine/6-03-Delete.robot
mhagen-vmware: otherwise, we should use some combination of %{TEST_DATACENTER} instead

andrewtchin:
Group6-VIC-Machine/6-03-Delete.robot passed locally on ESX and nimbus VC
https://ci.vcna.io/vmware/vic/14966 Group6 passed except for a flaky test

PR #6819, opened 2017-11-21
Add a route to the default gateway network

### LINKED ISSUE #6647: Traffic from containers to hosts in VCH subnet is sent to default gateway
**Description:**
All traffic from containers to external hosts is sent out from the VCH via the default gateway of the public interface, regardless of destination IP. Traffic destined to a host in the same subnet as the public interface of the VCH should be sent directly to the destination host, bypassing the gateway.

**VIC version:**
1.2.0 / master

**Deployment details:**

```vic-machine-linux create --name vch --image-store VMFS --compute-resource R610/Resources/VCH --public-network admin --public-network-ip XXX.YYY.235.1/22 --public-network-gateway XXX.YYY.235.254 --dns-server XXX.YYY.24.20 --dns-server XXX.YYY.24.10```

**Steps to reproduce:**
1. Create VCH as described above
2. Run ```docker run --rm busybox traceroute XXX.YYY.235.2``` where XXX.YYY.235.2 is an IP address in the same subnet as the VCH public interface

**Actual behavior:**
```
traceroute to XXX.YYY.235.2 (XXX.YYY.235.2), 30 hops max, 46 byte packets
 1  172.16.0.1 (172.16.0.1)  1.675 ms  1.249 ms  0.708 ms
 2  XXX.YYY.235.254 (XXX.YYY.235.254)  0.690 ms  1.053 ms  0.739 ms
 3  XXX.YYY.235.2 (XXX.YYY.235.2)  0.843 ms  1.104 ms  0.740 ms
```

**Expected behavior:**
```
traceroute to XXX.YYY.235.2 (XXX.YYY.235.2), 30 hops max, 46 byte packets
 1  172.16.0.1 (172.16.0.1)  1.215 ms  0.126 ms  32.674 ms
 2  XXX.YYY.235.2 (XXX.YYY.235.2)  1.011 ms  0.293 ms  1.437 ms
```

What seems to be happening is the lines
```
# add a routing table for traffic from bridge that is not destined
# for a local address
echo 201 bridge.out >> /etc/iproute2/rt_tables
ip rule add fwmark 1 table bridge.out
```
in /etc/systemd/scripts/nat-setup on the VCH is creating a new routing table named bridge.out, but not adding any routes. I believe the table inherits the system default route, but nothing else. 

```
root@ [ /etc/systemd/scripts ]# ip route show table bridge.out
default via XXX.YYY.235.254 dev public
```

Since there is only a single entry in this table, all packets get forwarded to the gateway, even those that are in the same subnet as the public IP (XXX.YYY.235.1/22 in the example above).

**Workaround:**
On the VCH run
```
ip route add XXX.YYY.232.0/22 dev public table bridge.out
```
to add a route for the local subnet. At this point traffic to the local subnet is now sent directly to the destination host, bypassing the gateway, as expected.

### PR DESCRIPTION
**Pull request by vburenin**
Fixes #6647 

bridge.out routing table on VCH is used to route a traffic from a containers subnet. That table initially is populated by default gateway route only to route a traffic to a public network, however, network where gateway exists is not added to the routing table, thus any traffic that is directed to the gateway network is going through gateway unnecessary saturating network bandwidth. This fix adds a route to the gateway network.

Before:
```
VCH:
root@ [ ~ ]# ip route show table bridge.out
default via 10.0.1.1 dev client 

Container:
/ # traceroute 10.0.1.96
traceroute to 10.0.1.96 (10.0.1.96), 30 hops max, 46 byte packets
 1  172.16.0.1 (172.16.0.1)  3.949 ms  1.344 ms  1.244 ms
 2  10.0.1.1 (10.0.1.1)  4.752 ms  1.232 ms  2.116 ms
 3  10.0.1.96 (10.0.1.96)  3.783 ms  2.376 ms  2.132 ms
```

After:
```
VCH:
root@ [ ~ ]# ip route show table bridge.out
default via 10.0.1.1 dev client 
10.0.1.0/24 dev client 

Container:
/ # traceroute 10.0.1.96
traceroute to 10.0.1.96 (10.0.1.96), 30 hops max, 46 byte packets
 1  172.16.0.1 (172.16.0.1)  0.038 ms  2.481 ms  2.080 ms
 2  10.0.1.96 (10.0.1.96)  3.492 ms  2.332 ms  2.112 ms
```

### CODE REVIEW

**Comment by sgairo** [comment]
Thanks! I also think a small test addition should be made here.

**Comment by vburenin** [comment]
@sgairo There is no easy way to test that out for multiple reasons. Unit test environment is not present for this logic. Writing functional test requires predictable network environment that need to be set up. I would write it if I have couple days left, but not by noon. This logic runs every time you boot VCH, so every test we run will "test" it.

File: lib/tether/ops_linux.go [hunk 1]
```diff
@@ -33,6 +33,7 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/d2g/dhcp4"
 	"github.com/docker/docker/pkg/archive"
+
```
anchal-agrawal: 👍 Thank you for doing this!

lib/tether/ops_linux.go
```diff
@@ -490,6 +492,24 @@ func updateDefaultRoute(t Netlink, link netlink.Link, endpoint *NetworkEndpoint)
 	}
 
 	if bTablePresent {
+		// Gateway IP may not container network mask, so it is taken from the assigned interface configuration
```
hickeng: `s/container/contain/`

lib/tether/ops_linux.go
```diff
@@ -490,6 +492,24 @@ func updateDefaultRoute(t Netlink, link netlink.Link, endpoint *NetworkEndpoint)
...
+			IP:   gw.IP.Mask(newIP.Mask),
+			Mask: newIP.Mask,
+		}
+
+		log.Debugf("Adding route to default gateway network: %s/%s", gwNet.IP, gwNet.Mask)
+
+		route = &netlink.Route{LinkIndex: link.Attrs().Index, Dst: gwNet, Table: bridgeTableNumber}
+		if err := t.RouteAdd(route); err != nil {
+			if errno, ok := err.(syscall.Errno); !ok || errno != syscall.EEXIST {
+				return fmt.Errorf(
```
hickeng: let's not wrap this - it makes the logic flow less obvious and doesn't gain anything in modern environments.

vburenin: It actually does in case of DHCP, if IP address changes within the same subnet it will case and error.

vburenin:
@hickeng To add a route to a network, you need a network mask - otherwise you can only use class masks which are almost irrelevant. That route was simply never added before.

Issue #6827, opened 2017-11-21
Automatically set ReadOnly role for ops-user


lcastellano:
**User Statement:**
When using --ops-grant-perms the ops-user must have a ReadOnly Role on / or the DC. Otherwise login fails during the vic-machine validation. A ReadOnly Role on the DC should be created automatically.

**Details:**
The current implementation requires the ops-user to have ReadOnly role on / or the DC when using --ops-grant-perms. This issue tracks the changes in documentation. However it would be a lot easier to use if the ReadOnly Permission was automatically granted when running vic-machine or the wizard.

**Acceptance Criteria:**
Code + Unit and integration test

PR #6842, opened 2017-11-26
Implement automatic setting of read-only top permission + refactoring

lcastellano:
Fixes: #6827


lib/install/validate/validator.go
```diff
@@ -470,6 +471,22 @@ func (v *Validator) credentials(op trace.Operation, input *data.Data, conf *conf
 		}
 	}
 
+	// If Grant Perms is set trying adding ReadOnly role to the Datacenter
+	// for the ops-user. This is necessary since the Login operation below
+	// fails if the ops-user has no permissions.
+	//
+	// FIXME ATC DEBT.
```
andrewtchin: Is this copied from the other place I put this? If so, can you add the issue link here and add a link to this line in the comments of the issue? And remove my initials from this one haha

lcastellano: LOL, I thought it was an acronym for a particular type of TECH DEBT

lib/install/opsuser/opsuser.go
```diff
@@ -0,0 +1,243 @@
...
+
+	finder.SetDatacenter(datacenter)
+
+	cluster, err := rp.GetCluster(ctx)
+	if err != nil {
+		return nil, errors.Errorf("Ops-User: AuthzManager, Unable to find Cluster: %s", err.Error())
+	}
+	resourceDescs = append(resourceDescs, ResourceDesc{rbac.Cluster, cluster.Reference()})
+
+	// Find image datastore
```
hickeng: `// Find image and volume datastores`

PR #6848, opened 2017-11-27
Fail earlier in install process when the host is down

mhagen-vmware:


tests/resources/VCH-Util.robot
```diff
@@ -48,7 +48,8 @@ Set Test Environment Variables
     ${server_date}=  Run  govc host.date.info
     Log To Console  \nTest_Server_Date=\n${server_date}\n
 
-    ${host}=  Run  govc ls host
+    ${rc}  ${host}=  Run And Return Rc And Output  govc ls host
```
hickeng: I'm surprised we don't failure earlier in the `govc about.cert` (passes because of the pipe?) or the `govc host.date.info` call. It seems that if any of those fail we should abort.

Should we look at having some assert/require keywords, e.g. `${output}= Require With Output  govc ls host`? as we do in the unit tests. Require is immediate failure and exit whereas Assert is fail the test but keep running past the statement.

mhagen-vmware: Yep, the about.cert gets masked by the pipe and the host.date.info doesn't have any assertions around it at all, just prints whatever gets returned.

PR #6915, opened 2017-12-05
CreateImageStore returns vSphere error allowing retry

### LINKED ISSUE #4858: Nightly 6.5 HA test fails during rmi busybox

cgtexmex:
- Propagate vSphere error allowing retry
- Log when a new operation is created
- Additional logging during ImageStore creation

Fixes #4858

### DIFF
--- a/lib/portlayer/storage/image_cache.go
+++ b/lib/portlayer/storage/image_cache.go
@@ -60,9 +60,20 @@ func NewLookupCache(ds ImageStorer) *NameLookupCache {
 	}
 }
 
+// isRetry will check the error for retryability - if so reset the cache
+func (c *NameLookupCache) isRetry(op trace.Operation, err error) bool {
+	if tasks.IsRetryError(op, err) {
+		op.Debugf("%s is retryable, resetting store cache", err)
+		c.storeCache = make(map[url.URL]*index.Index)
+		return true
+	}
+	return false
+}
+
 // GetImageStore checks to see if a named image store exists and returns the
 // URL to it if so or error.
 func (c *NameLookupCache) GetImageStore(op trace.Operation, storeName string) (*url.URL, error) {
+	defer trace.End(trace.Begin(fmt.Sprintf("StoreName: %s", storeName), op))
 	store, err := util.ImageStoreNameToURL(storeName)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


pkg/trace/operation.go
```diff
@@ -226,7 +226,11 @@ func opID(opNum uint64) string {
 
 // NewOperation will return a new operation with operationID added as a value to the context
 func NewOperation(ctx context.Context, format string, args ...interface{}) Operation {
-	return newOperation(ctx, opID(atomic.AddUint64(&opCount, 1)), 3, fmt.Sprintf(format, args...))
+	o := newOperation(ctx, opID(atomic.AddUint64(&opCount, 1)), 3, fmt.Sprintf(format, args...))
+
+	frame := o.t[0]
+	o.Debugf("[NewOperation] %s [%s:%d]", o.header(), frame.funcName, frame.lineNo)
```
hickeng: Are you intentionally adding this back in? I seem to recall it was removed during the big operation trace PR.

lib/portlayer/storage/image_cache.go
```diff
@@ -60,9 +60,20 @@ func NewLookupCache(ds ImageStorer) *NameLookupCache {
 	}
 }
 
+// isRetry will check the error for retryability - if so reset the cache
+func (c *NameLookupCache) isRetry(op trace.Operation, err error) bool {
+	if tasks.IsRetryError(op, err) {
+		op.Debugf("%s is retryable, resetting store cache", err)
+		c.storeCache = make(map[url.URL]*index.Index)
```
hickeng: Am I correct in assuming that there's no single place where we can purge the cache before regenerating it, e.g at the start of GetImageStore, and that's why it's done in the error type check?

cgtexmex: I didn't see a good place to do it -- originally I had it as part of the retry function, but it didn't quite work....so went with this approach...open to suggestions.

pkg/trace/operation.go
cgtexmex: Yes...this is intentional.  If you recall when we were looking into the `fixVM` issue I lamented that I removed it -- so I'm returning it.

PR #6919, opened 2017-12-05
Allow using only Harbor with content trust enabled

### LINKED ISSUE #6778: Only harbor registry should be used once content trust is enabled
**Story**
As a cloudadmin when I check the _content trust_ checkbox I expect only signed images to be allowed to run.

**Details**
With the addition of docker hub to the defined registry set in Admiral we need to be more prescriptive about which registries we pull from the config to use as the VCH whitelist.

We currently use the following query:
```
/config/registries?expand=true
```

We need to switch to
```
/config/registries/default-vic-registry
```
or
```
/config/registries?$filter=documentSelfLink eq '/config/registries/default-vic-registry'&expand=all
```

Note: the following would give _all but the default Docker Hub instance_:
/config/registries?$filter=documentSelfLink ne '/config/registries/default-registry'&expand=all

Queries provided by Sergio.

Estimate is for the change and should include an update to the integration test plan.

anchal-agrawal:
This commit modifies the whitelist registry logic for when content
trust is enabled through Admiral. Before this change, it was possible
to use Harbor and Docker Hub when content trust was enabled, however,
users should be able to use only signed images on Harbor.

Therefore, this change returns only the Harbor instance as the
whitelisted registry when content trust is enabled.

Test updates are in https://github.com/vmware/vic-product/pull/1227.

Fixes #6778

### DIFF
--- a/lib/config/dynamic/admiral/source.go
+++ b/lib/config/dynamic/admiral/source.go
@@ -51,6 +51,9 @@ const (
 	admiralEndpointKey = "guestinfo.vicova.admiral.endpoint"
 
 	clusterFilter = "(address eq '%s' and customProperties.__containerHostType eq 'VCH')"
+
+	// defaultVicRegistry is the query string to fetch the default VIC registry from Admiral
+	defaultVicRegistry = "default-vic-registry"
 )
 
 // #nosec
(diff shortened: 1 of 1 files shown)


lib/config/dynamic/admiral/source.go
```diff
@@ -352,22 +355,18 @@ func (a *source) whitelist(ctx context.Context, c *client.Admiral, hostProjs []s
...
-		m := &models.ComVmwareAdmiralServiceCommonRegistryServiceRegistryState{}
-		if err := mapstructure.Decode(r, m); err != nil {
-			log.Warnf("skipping registry: %s", err)
-			continue
-		}
-
-		wl = append(wl, m.Address)
+	if regs.Payload == nil {
+		// No default VIC registry configured.
+		return nil, nil
```
hickeng: can we update the calling method to record at Warn level that no default registry was found despite content trust being configured.

anchal-agrawal: There's another place in this func where we return `nil, nil` when there are no Admiral projects with a VCH and content trust enabled (https://github.com/vmware/vic/pull/6919/files#diff-1a00736b9f8c89eb8f1862400f7c8a0fR355) - so I've added a warn-level log to this function itself. The caller logs the received whitelist at debug level.

Issue #6851, opened 2017-11-27
Error while powering on containers - "Server error from portlayer: could not find IP range"


andrewtchin:
@lgayatri commented on [Tue Nov 21 2017](https://github.com/vmware/vic-product/issues/1123)

@andrewtchin @cgtexmex @mhagen-vmware 

![image](https://user-images.githubusercontent.com/25006749/33111526-9e35474e-cf74-11e7-9b8b-950373bfd907.png)

![image](https://user-images.githubusercontent.com/25006749/33111538-ad43dc28-cf74-11e7-8ac2-9b850f0c4886.png)

Service https://10.197.37.143:2376/v1.24/containers/c8b910a14b383e2d473ac18befbc220d0e072829e9b44424cf8554714f21ac49/start returned error 500 for POST. id 284135304 message Server error from portlayer: could not find IP range; Reason: {"message":"Server error from portlayer: could not find IP range"}
docker-personality.log says:
```
Nov 22 2017 05:09:54.751Z INFO  volumes found:
Nov 22 2017 05:09:55.326Z WARN  error getting config from source: config not modified
Nov 22 2017 05:14:47.145Z ERROR Operation failed with errors.apiError{error:(*errors.errorString)(0xc420a15240), statusCode:500}
Nov 22 2017 05:14:47.145Z ERROR Handler for POST /v1.24/containers/c8b910a14b383e2d473ac18befbc220d0e072829e9b44424cf8554714f21ac49/start returned error: Server error from portlayer: could not find IP range
Nov 22 2017 05:14:48.231Z ERROR Operation failed with errors.apiError{error:(*errors.errorString)(0xc4202c9ad0), statusCode:500}
Nov 22 2017 05:14:48.231Z ERROR Handler for POST /v1.24/containers/c8b910a14b383e2d473ac18befbc220d0e072829e9b44424cf8554714f21ac49/start returned error: Server error from portlayer: could not find IP range
Nov 22 2017 05:14:48.379Z ERROR Operation failed with errors.apiError{error:(*errors.errorString)(0xc420573f30), statusCode:500}
Nov 22 2017 05:14:48.379Z ERROR Handler for POST /v1.24/containers/c8b910a14b383e2d473ac18befbc220d0e072829e9b44424cf8554714f21ac49/start returned error: Server error from portlayer: could not find IP range
Nov 22 2017 05:14:48.661Z ERROR Operation failed with errors.apiError{error:(*errors.errorString)(0xc42048b360), statusCode:500}
[...]

Issue #6384, opened 2017-09-18
9/50 Containers are not exiting after completing the desired command


lgayatri:
vic-version : v1.2.0-rc5
I started command sleep 5d on photon container. The 5d period completed 43hours back but still I am seeing few containers running and up since 6 days.

Pasting the output:

`root@Ubuntu-VIc:/home/vmlib/VIC/VIC-rc4/vic# docker -H 10.197.37.133:2376 --tls ps -a
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS                    PORTS               NAMES
667d166740a7        photon              "sleep 5d"          6 days ago          Exited (0) 43 hours ago                       blissful_wilson
dce93563ccf3        photon              "sleep 5d"          6 days ago          Exited (0) 43 hours ago                       unruffled_chandrasekhar
ca1acc17386e        photon              "sleep 5d"          6 days ago          Exited (0) 43 hours ago                       reverent_edison
6bec4cdaede3        photon              "sleep 5d"          6 days ago          Exited (0) 43 hours ago                       fervent_jepsen
05470c6ac631        photon              "sleep 5d"          6 days ago          Exited (0) 43 hours ago                       frosty_blackwell
ce73cc359d79        photon              "sleep 5d"          6 days ago          Exited (0) 43 hours ago                       flamboyant_gates
90d8293133fa        photon              "sleep 5d"          6 days ago          Exited (0) 43 hours ago                       wonderful_fermi
33a2d80ec52f        photon              "sleep 5d"          6 days ago          Exited (0) 43 hours ago                       laughing_edison
8b88cc0ff3fa        photon              "sleep 5d"          6 days ago          Exited (0) 43 hours ago                       hopeful_banach
0782a88abdd4        photon              "sleep 5d"          6 days ago          Exited (0) 43 hours ago                       lucid_borg
4e88a9870699        photon              "sleep 5d"          6 days ago          Exited (0) 43 hours ago                       festive_einstein
032f4f9f94c1        photon              "sleep 5d"          6 days ago          Exited (0) 43 hours ago                       jovial_bassi
6f3c14f583ca        photon              "sleep 5d"          6 days ago          Exited (0) 43 hours ago                       condescending_tesla
870e1cfcd026        photon              "sleep 5d"          6 days ago          Up 6 days                                     hopeful_leavitt
5a108bb4a63e        photon              "sleep 5d"          6 days ago          Exited (0) 44 hours ago                       inspiring_sammet
[...]

PR #6937, opened 2017-12-08
Increase event page size to 1000

cgtexmex:
Increasing the event page size on the portlayer
to help with missing events.

Helps with #6384, #6851


hickeng:
Could you run test `Group0-Bugs/5343` against this to confirm sane recovery when the session is dropped.

cgtexmex:
@hickeng ran that test locally and am concerned by the amount of _event spew_ on VCH startup -- the event collector is created and the last 1K events on that host / cluster before it's powerOn will be seen by the VCH.  That could be helpful in avoiding missed events in the case of an outage, but unhelpful when it's a new VCH.  In the case of a new VCH we'll log and callback for many events that have nothing to do with the VCH.  It added about 70KB to the portLayer log...

PR #6945, opened 2017-12-12
Write all vic-machine log files at debug log level

zjs:
Historically, the log level of output written to the TTY has been tied to the log level of messages written to the `vic-machine.log` file.

This tight coupling no longer exists. Now, separate loggers are used for writing to files (`vic-machine.log` and the datastore log file) and standard out, allowing for separate log level configuration.

Leverage this separation of concerns to always write log files at the debug log level. This may simplify debugging of VCH creation issues; we will no longer ever have to ask a user to re-try the operation at
a more verbose log level.

If useful, this pattern could be applied to other CLI operations; perhaps all log files should be written at the debug log level.

### DIFF
--- a/cmd/vic-machine/create/create.go
+++ b/cmd/vic-machine/create/create.go
@@ -25,7 +25,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/Sirupsen/logrus"
 	"gopkg.in/urfave/cli.v1"
 
 	"github.com/vmware/vic/cmd/vic-machine/common"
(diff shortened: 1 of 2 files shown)


mhagen-vmware:
The original issue, didn't really give the reason behind splitting them - but this PR is the exact reason why we wanted to split them

zjs:
I've pushed a second commit which up-levels this logic to apply to _all_ `vic-machine` commands instead of just `create`.

This may be especially helpful for debugging commands like `upgrade`, which do not have a `debug` flag.

cc @hickeng

PR #6949, opened 2017-12-12
Cherry-pick: Ensure that only a single session is constructed for each VCH Management API request [specific ci=Group23-VIC-Machine-Service]

### LINKED ISSUE #6947: Datacenter-scoped VCH management APIs log in twice, which does not work with session tickets
**VIC version:**

RC3

**Steps to reproduce:**

Use a session ticket to authenticate to a datacenter-scoped API.

For example (ugly command):
```
curl --insecure --silent --request GET https://$OVA_IP:8443/container/target/$VC_IP/datacenter/datacenter-41/vch?thumbprint=$THUMBPRINT -H "X-VMWARE-TICKET: $(govc vm.console virtual-container-host | cut -d "/" -f 3 | cut -d ":" -f 2 | cut -d "@" -f 1)"
```

**Actual behavior:**

The request fails with:
```
{
  "message": "Validation Error: Failed to log in to 10.161.241.196: ServerFaultCode: Cannot complete login due to an incorrect user name or password."
}
```

**Expected behavior:**

The request succeeds.

The root cause of this issue is the "double validation" pattern we use for datacenter-scoped sessions.

zjs:
The VCH Management API supports two authentication mechanisms:
1. Basic authentication with a username and password.
2. Ticket-based authentication using a session ticket.

One limitation of ticket-based authentication, which is used by the vSphere H5 Client Plugin, is that a session ticket may only be used once.

Prior to this series, we created two sessions as a part of handling datacenter-scoped API requests: one un-scoped session to resolve and validate the supplied datacenter, and one datacenter-scoped session to actually handle the request.

This pattern worked fine for basic authentication, but not for ticket-based authentication.

To correct this mostly involves refactoring of handler code, followed by a commit to eliminate the use of two validators (f274cde).

Two test cases are introduced for session-based authentication to un-scoped and datacenter-scoped requests (749058a). Additionally, the datacenter-scoped list and create APIs were manually exercised using ticket-based authentication.

Fixes #6947 

---

Note: I have broken this fix into a series of small (atomic?) commits. It may be easier to understand by reading commit-by-commit instead of attempting to understand the entire change at once.

### DIFF
--- a/lib/apiservers/service/restapi/handlers/common.go
+++ b/lib/apiservers/service/restapi/handlers/common.go
@@ -40,71 +40,88 @@ type buildDataParams struct {
 	thumbprint      *string
 	datacenter      *string
 	computeResource *string
+	vchID           *string
 }
 
-func buildData(op trace.Operation, params buildDataParams, principal interface{}) (*data.Data, error) {
-	d := data.Data{
+func buildDataAndValidateTarget(op trace.Operation, params buildDataParams, principal interface{}) (*data.Data, *validate.Validator, error) {
+	data := &data.Data{
 		Target: &common.Target{
 			URL: &url.URL{Host: params.target},
 		},
 	}
 
 	if c, ok := principal.(Credentials); ok {
-		d.Target.User = c.user
-		d.Target.Password = &c.pass
+		data.Target.User = c.user
+		data.Target.Password = &c.pass
 	} else {
-		d.Target.CloneTicket = principal.(Session).ticket
+		data.Target.CloneTicket = principal.(Session).ticket
+	}
+
+	if err := data.HasCredentials(op); err != nil {
+		return data, nil, util.NewError(http.StatusUnauthorized, "Invalid Credentials: %s", err)
 	}
 
 	if params.thumbprint != nil {
-		d.Thumbprint = *params.thumbprint
+		data.Thumbprint = *params.thumbprint
 	}
 
+	if params.computeResource != nil {
+		data.ComputeResourcePath = *params.computeResource
+	}
+
+	if params.vchID != nil {
+		data.ID = *params.vchID
+	}
+
+	// TODO (#6032): clean this up
+	var validator *validate.Validator
 	if params.datacenter != nil {
-		validator, err := validateTarget(op, &d)
+		v, err := validate.NewValidator(op, data)
 		if err != nil {
-			return nil, util.WrapError(http.StatusInternalServerError, err)
+			return data, nil, util.NewError(http.StatusBadRequest, "Validation Error: %s", err)
 		}
 
 		datacenterManagedObjectReference := types.ManagedObjectReference{Type: "Datacenter", Value: *params.datacenter}
 
-		datacenterObject, err := validator.Session.Finder.ObjectReference(op, datacenterManagedObjectReference)
+		datacenterObject, err := v.Session.Finder.ObjectReference(op, datacenterManagedObjectReference)
 		if err != nil {
-			return nil, util.WrapError(http.StatusNotFound, err)
+			return nil, nil, util.WrapError(http.StatusNotFound, err)
 		}
 
-		d.Target.URL.Path = datacenterObject.(*object.Datacenter).InventoryPath
+		dc, ok := datacenterObject.(*object.Datacenter)
+		if !ok {
+			return data, nil, util.NewError(http.StatusBadRequest, "Validation Error: datacenter parameter is not a datacenter moref")
+		}
+
+		data.Target.URL.Path = dc.InventoryPath
+
+		// Do what NewValidator would have done if a Datacenter had been set
+		v.DatacenterPath = dc.Name()
+		v.Session.DatastorePath = v.DatacenterPath
+		v.Session.Finder.SetDatacenter(dc)
+
+		validator = v
+	} else {
+		v, err := validate.NewValidator(op, data)
+		if err != nil {
+			return data, nil, util.NewError(http.StatusBadRequest, "Validation Error: %s", err)
+		}
+
+		// If dc is not set, and multiple datacenters are available, operate on all datacenters.
+		v.AllowEmptyDC()
+
+		validator = v
 	}
 
-	if params.computeResource != nil {
-		d.ComputeResourcePath = *params.computeResource
+	if _, err := validator.ValidateTarget(op, data); err != nil {
+		return data, nil, util.NewError(http.StatusBadRequest, "Target validation failed: %s", err)
 	}
 
-	return &d, nil
-}
-
-func validateTarget(op trace.Operation, d *data.Data) (*validate.Validator, error) {
-	if err := d.HasCredentials(op); err != nil {
-		return nil, fmt.Errorf("Invalid Credentials: %s", err)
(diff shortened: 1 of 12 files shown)


lib/apiservers/service/restapi/handlers/util/error.go
```diff
@@ -14,7 +14,11 @@
 
 package util
 
-import "net/http"
+import (
+	"fmt"
+
```
jzt: Can we delete this line?

andrewtchin:
should this be targeted to master then cherry picked to 1.3?

lib/apiservers/service/restapi/handlers/util/error.go
zjs: Yes! Done.

zjs:
> should this be targeted to master then cherry picked to 1.3?

Given the nature of this change, it seemed safer to develop and test against 1.3 and then cherry-pick back to `master`.

lib/apiservers/service/restapi/handlers/common.go
```diff
@@ -40,71 +41,82 @@ type buildDataParams struct {
...
 		}
 
-		d.Target.URL.Path = datacenterObject.(*object.Datacenter).InventoryPath
-	}
+		data.Target.URL.Path = datacenterObject.(*object.Datacenter).InventoryPath
 
-	if params.computeResource != nil {
-		d.ComputeResourcePath = *params.computeResource
-	}
+		// Copied from NewValidator
```
hickeng: I really don't like this - I can see you're trying to limit the changes to not touch anything outside of the server code, but it's really fragile.

I think our neater options are:
1. Update NewValidator to support a datacenter moref 
2. Add a utility method to construct a session from `data.Data`, resolve the datacenter with that session, then use it to construct a Target url with /datacenter
3. Proper refactoring of validator 

However... if confining the changes to this package then I think this needs the following:
```
dc, ok := datacenterObject.(*object.Datacenter)
if !ok {
   return data, nil, util.NewError(http.StatusBadRequest, "Validation Error: datacenter parameter is not a datacenter moref")

v.Session.Finder.SetDatacenter(dc)
v.DatacenterPath = dc.Name()
```

The `finder.SetDatacenter` operation is what sets the scoping for all subsequent moref/name resolution queries.

lib/apiservers/service/restapi/handlers/common.go
```diff
@@ -40,71 +41,82 @@ type buildDataParams struct {
...
+		if err != nil {
+			return data, nil, util.NewError(http.StatusBadRequest, "Validation Error: %s", err)
+		}
 
-func validateTarget(op trace.Operation, d *data.Data) (*validate.Validator, error) {
-	if err := d.HasCredentials(op); err != nil {
-		return nil, fmt.Errorf("Invalid Credentials: %s", err)
-	}
+		// If dc is not set, and multiple datacenters are available, operate on all datacenters.
+		v.AllowEmptyDC()
```
hickeng: IIRC this was added solely for the operations that don't care about a specific VCH, so `ls`, and `update`, of which only `ls` is relevant for us.

I think it may be better to constrain this path based on the rest operation - I don't have confidence that we've tested what happens if you try other operations without a datacenter defined.

lib/apiservers/service/restapi/handlers/common.go
zjs: > I really don't like this - I can see you're trying to limit the changes to not touch anything outside of the server code, but it's really fragile.

This is definitely something I'm torn about.

However, I think confining the changes to server code for 1.3 greatly reduces the overall risk of the change. As soon as `master` is open for post-1.3 changes, I'd like to explore option 1 or 3. (I'm not sure 2 works without 3, since to achieve 2 we'd need to refactor validator enough that it can accept an existing session.)

I'm also not sure any of these options remove the ugly aspects of this change. For example, achieving option 1 still requires creating a session, using it to resolve a datacenter moref into a path, and updating the session to use that datacenter path. That is, the fragile code for updating a session which has already been connected would still exist ― just in a different place. (I think that different place is a _better_ place in the long run, but I don't think it's worth the risk for 1.3.)

> However... if confining the changes to this package then I think this needs the following:

Is this code something you wrote for this review, or was it borrowed from somewhere?

> `dc, ok := datacenterObject.(*object.Datacenter)`

I think it would be good to be defensive like this, but it does not seem like `!ok` could ever happy with the way `object.NewReference` is currently implemented. Is that correct?

(Happy to add this; just trying to make sure I haven't misunderstood something.)

> `v.Session.Finder.SetDatacenter(dc)`

Good catch!

> `v.DatacenterPath = dc.Name()`

This seems slightly different from `strings.TrimPrefix(data.Target.URL.Path, "/")`, which is what's used in `validator`:

https://github.com/vmware/vic/blob/d860ced9c3e1362ef04447798a28025cef29c94a/lib/install/validate/validator.go#L136

Might there be cases where `.Name()` doesn't produce the right value here?

lib/apiservers/service/restapi/handlers/common.go
zjs: The call to `.AllowEmptyDC()` is present in the `validateTarget` method today, so it's been present for all of the testing of the API/UI.

If there are not known issues with having it here, I'd be tempted to leave it for now to avoid unknown issues with removing it (since I think we've done more testing _with_ it than we would do _without_ it between now and when we ship).

zjs:
> this needs to go into master first though

Why? Forward-porting and back-porting are both valid strategies for fixing an issue on two branches, and each has its own benefits.

> You can develop it on release/1.3.0 but not sure why you would do that... 

`master` and `release/1.3.0` aren't _exactly_ the same. I don't know that they're different in a way that's meaningful to this change, but I'd much rather play it safe and develop a change like this against the release branch so that reviewers can verify that nothing was lost or unintentionally changed during any of the refactoring steps.

One key difference is that when I post a PR for the cherry-pick, it will just be one big commit which makes it harder to catch those sorts of mistakes. (If I introduce a subtle issue as a result of cherry-picking to master, we have lots of time to fix it. If I introduce a subtle issue as a result of cherry-picking to the release branch, I'm not confident we'd notice it before shipping.)

> also, is this ultimately to fix vmware/vic-ui#208?

Yes. This dependency relationship should be captured in #6947.

lib/apiservers/service/restapi/handlers/common.go
hickeng: a. You're correct that the cast will always succeed as written, but that's dependent on code from a different repo so I tend towards defensive.

b. It is slightly different because the source is coming from a different location - it may be called a path but if you look at `validator.go:164` we actually ensure there's no path separators because it was a common error to add cluster, etc, as well. Finder will cope with a path, and may in fact do better with it, but I don't know if anything else will break if we allow a DatacenterPath with multiple segments.

mhagen-vmware:
We have a well defined flow of commits and cherry picks from master to release branches that provides us a historical look and full testing in one direction.

mhagen-vmware:
https://github.com/vmware/vic/blob/master/doc/processes/BRANCHING.md

zjs:
Successful CI run on 652c9a68: https://ci.vcna.io/vmware/vic/15317 (two failures were due to #6958 and #5621).

Issue #6886, opened 2017-11-30
vSphere SDK endpoint becomes unresponsive/eventually login fails completely


mhagen-vmware:
I think it is pretty much this:
https://github.com/vmware/vic/issues/5621

But recently we have seen this behavior with only a single iteration of our CI not parallel executions. Logs and more history can be found here:
https://bugzilla.eng.vmware.com/show_bug.cgi?id=2005701

PR #6951, opened 2017-12-13
Reduce datastore searches during non-vSAN delete operations

hickeng:
Noticed a LOT of datastore searches and work for what looked like image store delete.
I remember needing this on vSAN because it wasn't performing recursive delete cleanly, however the fact it has been applied to other datastore types as well is either an accident/miscommunication.

This was added to see if reducing the raw number of operations reduced the hostd instability issue (#6886) - while it did not it's still worth adding to reduce unnecessary control plane load.


lib/install/management/store_files.go
```diff
@@ -158,23 +158,26 @@ func (d *Dispatcher) isVSAN(ds *object.Datastore) bool {
...
-	// Get sorted result to make sure children files listed ahead of folder. Then we can empty folder before delete it
-	// This function specifically designed for vSan, as vSan sometimes will throw error to delete folder is the folder is not empty
-	res, err := d.getSortedChildren(ds, dsPath)
-	if err != nil {
-		if !types.IsFileNotFound(err) {
-			err = errors.Errorf("Failed to browse sub folders %q: %s", dsPath, err)
-			return err
+	if d.isVSAN(ds) {
+		// Get sorted result to make sure children files listed ahead of folder. Then we can empty folder before delete it
+		// This function specifically designed for vSan, as vSan sometimes will throw error to delete folder is the folder is not empty
```
AngieCris: teeny-weeny comment typo: `if the folder is not empty`?

PR #6943, opened 2017-12-12
Guest shutdown support in endpointVM

hickeng:
This adds neat shutdown support in the endpointVM:
* cleans up Views, Managers and sessions in portlayer
* cleans up session in personality
* cleans up vicadmin sessions on shutdown

Unifies diskmanager usage on a single instance for the storage
component. This is unrelated to the rest of the PR but is the correct usage
given locking assumptions and will reduce retries of vmomi operations.

Updates vic-machine to use guest shutdown for polite shutdown first.

Fixes tether unit tests to run as non-root (again).

Notes:
Need to find an effective way to test this. I was planning on adding the name of the VCH to the client but that ran into the immovable object that is `session.Session` usage. I still think this would be a useful UX enhancement for both us and customers, and `session.Session` _definitely_ needs attention.

Currently there is still occasional dangling sessions from the personality admiral client. In this case client.Logout gets called but it does not occur. In the last full-ci run this happened twice:
```
Deleting the VCH appliance VCH-17937-4337
[ WARN ] Dangling sessions found: 52558178-8cbb-2dd6-fe91-75f1f64a7128  VSPHERE.LOCAL\Administrator                                        2018-04-02 23:01  3m34s  10.158.214.80   vic-dynamic-config/1.4.0-dev  
```

This is ~~**not ready to merge**~~ until the hardcoded filtering for 1.4.0-dev sessions is improved.
 - updated to filter for `-dev`, skipping the upgrade leaked sessions. This can be removed once old versions are not expected to leak sessions.

```
[parallel jobs=4]
```
Todo:
- [x] remove hardcoded version filtering
  - now filters for `-dev` to omit sessions from GA code. This means it will not currently match RC builds but given we're only using it for warning currently I'm ok with that.
- [x] perform neat exit on `could not initialize port layer` errors rather than log.Fatal

PR #6955, opened 2017-12-13
Vch guest operations ec update [full ci]

hickeng:
Variant of #6943 

Updates govmomi to delete event history collector on session deletion.


hickeng:
#6956 was created as this PR seemed stalled for picking up changes in CI.

PR #6960, opened 2017-12-14
Add per container API locking for ContainerStart and improve logging

### LINKED ISSUE #6851: Error while powering on containers - "Server error from portlayer: could not find IP range"

sflxn:
In load testing, we've found overlapping request to ContainerStart
for the same container, causing IP reservation issues.  We are
adding back the per container locking we had before to prevent this.
Currently, this locking is being utilized for operations around
ContainerStart.  We need to add this back in for other docker API
access in the future.

Also, we removed useless logging and added more logging around
ContainerStart in the persona and around IP reservation/release in
the portlayer.

Resolves #6851

### DIFF
--- a/lib/apiservers/engine/backends/backends.go
+++ b/lib/apiservers/engine/backends/backends.go
@@ -57,6 +57,7 @@ const (
 	// RetryTimeSeconds defines how many seconds to wait between retries
 	RetryTimeSeconds        = 2
 	defaultSessionKeepAlive = 20 * time.Second
+	APITimeout              = constants.PropertyCollectorTimeout + 3*time.Second
 )
 
 var (
--- a/lib/apiservers/engine/backends/cache/repo_cache.go
+++ b/lib/apiservers/engine/backends/cache/repo_cache.go
@@ -23,7 +23,6 @@ import (
 
 	"github.com/vmware/vic/lib/apiservers/engine/backends/kv"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client"
-	"github.com/vmware/vic/pkg/trace"
 
 	"github.com/docker/distribution/digest"
 	"github.com/docker/docker/reference"
@@ -128,8 +127,6 @@ func init() {
 // NewRespositoryCache will create a new repoCache or rehydrate
 // an existing repoCache from the portlayer k/v store
 func NewRepositoryCache(client *client.PortLayer) error {
-	defer trace.End(trace.Begin(""))
-
 	rCache.client = client
 
 	val, err := kv.Get(client, repoKey)
@@ -165,8 +162,6 @@ func NewRepositoryCache(client *client.PortLayer) error {
 // Save will persist the repository cache to the
 // portlayer k/v
 func (store *repoCache) Save() error {
-	defer trace.End(trace.Begin(""))
-
 	b, err := json.Marshal(store)
 	if err != nil {
 		log.Errorf("Unable to marshal repository cache: %s", err.Error())
@@ -183,7 +178,6 @@ func (store *repoCache) Save() error {
 }
 
 func (store *repoCache) AddReference(ref reference.Named, imageID string, force bool, layerID string, save bool) error {
-	defer trace.End(trace.Begin(""))
 	if ref.Name() == string(digest.Canonical) {
 		return errors.New("refusing to create an ambiguous tag using digest algorithm as name")
 	}
@@ -253,7 +247,6 @@ func (store *repoCache) AddReference(ref reference.Named, imageID string, force 
 // Tags: busybox:1.25.1
 // Digest: nginx@sha256:7281cf7c854b0dfc7c68a6a4de9a785a973a14f1481bc028e2022bcd6a8d9f64
 func (store *repoCache) Remove(ref string, save bool) (string, error) {
-	defer trace.End(trace.Begin(""))
 	n, err := reference.ParseNamed(ref)
 	if err != nil {
 		return "", err
@@ -270,7 +263,6 @@ func (store *repoCache) Remove(ref string, save bool) (string, error) {
 // Delete deletes a reference from the store. It returns true if a deletion
 // happened, or false otherwise.
 func (store *repoCache) Delete(ref reference.Named, save bool) (bool, error) {
-	defer trace.End(trace.Begin(""))
 	ref = reference.WithDefaultTag(ref)
 
 	store.mu.Lock()
@@ -315,7 +307,6 @@ func (store *repoCache) Delete(ref reference.Named, save bool) (bool, error) {
 // GetImageID will return the imageID associated with the
 // specified layerID
 func (store *repoCache) GetImageID(layerID string) string {
-	defer trace.End(trace.Begin(layerID))
 	var imageID string
 	store.mu.RLock()
 	defer store.mu.RUnlock()
@@ -327,7 +318,6 @@ func (store *repoCache) GetImageID(layerID string) string {
 
 // Get returns the imageID for a parsed reference
 func (store *repoCache) Get(ref reference.Named) (string, error) {
-	defer trace.End(trace.Begin(""))
 	ref = reference.WithDefaultTag(ref)
 
 	store.mu.RLock()
@@ -347,8 +337,6 @@ func (store *repoCache) Get(ref reference.Named) (string, error) {
 
 // Tags returns a slice of tags for the specified imageID
 func (store *repoCache) Tags(imageID string) []string {
-	defer trace.End(trace.Begin(""))
-
 	store.mu.RLock()
 	defer store.mu.RUnlock()
 	var tags []string
@@ -362,8 +350,6 @@ func (store *repoCache) Tags(imageID string) []string {
 
 // Digests returns a slice of digests for the specified imageID
 func (store *repoCache) Digests(imageID string) []string {
-	defer trace.End(trace.Begin(""))
-
 	store.mu.RLock()
 	defer store.mu.RUnlock()
 	var digests []string
@@ -378,8 +364,6 @@ func (store *repoCache) Digests(imageID string) []string {
 // References returns a slice of references to the given imageID. The slice
 // will be nil if there are no references to this imageID.
 func (store *repoCache) References(imageID string) []reference.Named {
-	defer trace.End(trace.Begin(""))
-
 	store.mu.RLock()
 	defer store.mu.RUnlock()
 
@@ -401,8 +385,6 @@ func (store *repoCache) References(imageID string) []reference.Named {
 // If there are no references known for this repository name,
 // ReferencesByName returns nil.
 func (store *repoCache) ReferencesByName(ref reference.Named) []Association {
-	defer trace.End(trace.Begin(""))
-
 	store.mu.RLock()
 	defer store.mu.RUnlock()
 
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -752,7 +752,8 @@ func (c *Container) ContainerResize(name string, height, width int) error {
 // stop. Returns an error if the container cannot be found, or if
 // there is an underlying error at any stage of the restart.
 func (c *Container) ContainerRestart(name string, seconds *int) error {
-	defer trace.End(trace.Begin(name))
+	op := trace.NewOperation(context.Background(), "ContainerRestart - %s", name)
+	defer trace.End(trace.Begin(name, op))
 
 	// Look up the container name in the metadata cache ot get long ID
 	vc := cache.ContainerCache().GetContainer(name)
@@ -768,7 +769,7 @@ func (c *Container) ContainerRestart(name string, seconds *int) error {
 	}
 
 	operation = func() error {
-		return c.containerStart(name, nil, true)
+		return c.containerStart(op, name, nil, true)
 	}
 	if err := retry.Do(operation, IsConflictError); err != nil {
 		return InternalServerError(fmt.Sprintf("Start failed with: %s", err))
@@ -902,19 +903,21 @@ func (c *Container) cleanupPortBindings(vc *viccontainer.VicContainer) error {
 
 // ContainerStart starts a container.
 func (c *Container) ContainerStart(name string, hostConfig *containertypes.HostConfig, checkpoint string, checkpointDir string) error {
-	defer trace.End(trace.Begin(name))
+	op := trace.NewOperation(context.Background(), "ContainerStart - %s", name)
+	defer trace.End(trace.Begin(name, op))
 
 	operation := func() error {
-		return c.containerStart(name, hostConfig, true)
+		return c.containerStart(op, name, hostConfig, true)
 	}
 	if err := retry.Do(operation, IsConflictError); err != nil {
+		op.Debugf("Container start failed due to error - %s", err.Error())
 		return err
 	}
 
 	return nil
 }
 
-func (c *Container) containerStart(name string, hostConfig *containertypes.HostConfig, bind bool) error {
+func (c *Container) containerStart(op trace.Operation, name string, hostConfig *containertypes.HostConfig, bind bool) error {
 	var err error
 
 	// Get an API client to the portlayer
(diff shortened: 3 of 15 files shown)


sflxn:
This PR has small code changes, some useless logging removed, a lot of useful logging added around IP reservation.

sgairo:
Do you think we should create an issue to add testing of the API locking?

lib/apiservers/engine/backends/cache/repo_cache.go
```diff
@@ -128,8 +127,6 @@ func init() {
 // NewRespositoryCache will create a new repoCache or rehydrate
 // an existing repoCache from the portlayer k/v store
 func NewRepositoryCache(client *client.PortLayer) error {
-	defer trace.End(trace.Begin(""))
```
vburenin: I like you removed this relatively annoying debug traces!

lib/apiservers/engine/backends/container.go
```diff
@@ -752,7 +752,8 @@ func (c *Container) ContainerResize(name string, height, width int) error {
 // stop. Returns an error if the container cannot be found, or if
 // there is an underlying error at any stage of the restart.
 func (c *Container) ContainerRestart(name string, seconds *int) error {
-	defer trace.End(trace.Begin(name))
+	op := trace.NewOperation(context.Background(), "ContainerRestart - %s", name)
+	defer trace.End(trace.Begin(op.String()))
```
cgtexmex: FYI - you can pass the `op` to trace as a second parameter and it will include the opID in it's tracing exercise...
[see](https://github.com/vmware/vic/blob/master/lib/apiservers/portlayer/restapi/handlers/storage_handlers.go#L167) as an example

lib/apiservers/engine/backends/backends.go
```diff
@@ -57,6 +57,7 @@ const (
 	// RetryTimeSeconds defines how many seconds to wait between retries
 	RetryTimeSeconds        = 2
 	defaultSessionKeepAlive = 20 * time.Second
+	APITimeout              = 5 * time.Second
```
vburenin: Why 5 seconds? You need to add at least a comment to justify it for other people.
5 seconds seems very small and not particularly clear.

lib/apiservers/engine/backends/container/viccontainer.go
```diff
@@ -17,6 +17,8 @@ package container
 import (
 	"sync"
```
vburenin: not needed ws

lib/apiservers/engine/backends/container.go
```diff
@@ -925,7 +928,12 @@ func (c *Container) containerStart(name string, hostConfig *containertypes.HostC
 	if vc == nil {
 		return NotFoundError(name)
 	}
+	if !vc.TryLock(APITimeout) {
+		return ConcurrentAPIError(name, "ContainerStart")
```
cgtexmex: Is there no way to provide a useful message here?  i.e. the container is already being _fill in the blank_ -- essentially it's not a timeout, but something else is happening can't we tell the user what's happening?

lib/apiservers/engine/backends/container/viccontainer.go
```diff
@@ -65,3 +71,21 @@ func (v *VicContainer) List() []string {
...
+		return false
+	case <-v.lockChan:
+		return true
+
+	}
+}
+
+// Unlocks the container
+func (v *VicContainer) Unlock() {
+	v.lockChan <- true
```
vburenin: What happens if you accidentally call Unlock twice?

cgtexmex:
Are we adding lock complexity due to an order of operation issue? Why wouldn't we facilitate the bind in the portLayer as part of the commit -- possibly that's just pushing the _problem_ there, but in many cases we'll be waiting on the infrastructure to provide us an IP, etc...

lib/apiservers/engine/backends/container/viccontainer.go
sflxn: That would be a problem.  This is a super simple lock with timeout.  Let me rethink this.

sflxn:
@cgtexmex We can take this discussion offline if the debate goes further.  Pertaining to where we should solve the problem, I think pushing the bind and unbind on vsphere concurrent access to commit seems like adding a whole lot of more complexity to the portlayer and overloading what might should be primitive operations provided by the portlayer.

lib/apiservers/engine/backends/container/viccontainer.go
vburenin: You could use select to read from the channel and panic if default works.
My point is that unexpected lock is hard to debug, if it panics here, it would be a lot easier.
Standard mutex doesn't allow you to unlock things twice, it panics.

lib/apiservers/engine/backends/container.go
```diff
@@ -902,19 +903,21 @@ func (c *Container) cleanupPortBindings(vc *viccontainer.VicContainer) error {
 
 // ContainerStart starts a container.
 func (c *Container) ContainerStart(name string, hostConfig *containertypes.HostConfig, checkpoint string, checkpointDir string) error {
-	defer trace.End(trace.Begin(name))
+	op := trace.NewOperation(context.Background(), "ContainerStart - %s", name)
+	defer trace.End(trace.Begin(op.String()))
```
matthewavery: op should be suppliable as a parameter to this as well. Per Clint's comment above.

lib/apiservers/engine/backends/container.go
sflxn: I'm not sure what you're saying.  We are supplying the API caller an error message.
```
func ConcurrentAPIError(name, request string) error {
	return derr.NewRequestConflictError(fmt.Errorf("A concurrent %s request is already in progress for container '%s'.  Please try again later.", request, name))
}
```
This is the error you will see if you're able to make overlapping ContainerStart request on the same container on a loaded system.  It is also what you see in the logs.

```
Dec 14 2017 01:56:40.373Z DEBUG op=281.2143: Container start failed due to error - A concurrent ContainerStart request is already in progress for container '9771b7879ec6c570a15a20c07cd8d6198da9d76e5eaff7231ecaefd8f4342b2b'.  Please try again later.
```

lib/apiservers/engine/backends/backends.go
sflxn: Completely arbitrary.  Not sure what a good timeout should be.  I'm open to suggestions.

lib/apiservers/engine/backends/cache/repo_cache.go
sflxn: Yes, there were so many annoying useless debug traces clogging up the logs while I was debugging this issue.

lib/apiservers/engine/backends/container.go
cgtexmex: You're right..I missed the fact that you were providing the request -- ie. `start`-- I originally read it as just a vanilla error when no request detail.

lib/portlayer/network/context.go
```diff
@@ -698,14 +698,20 @@ func (c *Context) bindContainer(h *exec.Handle) ([]*Endpoint, error) {
...
+				// for VCH restart, we need to reserve
+				// the IP of the running container
+				//
+				// this may be a DHCP assigned IP, however, the
+				// addContainer call below will ignore reserving
+				// an IP if the scope is "dynamic"
+				log.Debugf("BindContainer found container %s with dynamic endpoint and assigned IP %s", con.ID(), ne.Assigned.IP.String())
+				eip = &ne.Assigned.IP
+			} else {
+				log.Debugf("BindContainer found container %s with dynamic endpoint and unassigned IP", con.ID())
```
hickeng: Really should be using `op.Debug` - otherwise we're just reverting to the pain of not being able to easily correlate logs during concurrent operations.
This would also make it consistent with the Unbind path

sflxn:
@sgairo It's really difficult to add test for this.  You need a heavily loaded system to make the concurrency occur.  I tried a hack to allow me to test it quicker yesterday, but it can't actually be used in CI.  It requires changing the code to artificially make the occurrence happen.

lib/portlayer/network/ipam.go
```diff
@@ -281,13 +283,25 @@ func (s *AddressSpace) ReserveIP4Range(firstIP net.IP, lastIP net.IP) (*AddressS
...
 		subSpace := NewAddressSpaceFromRange(firstIP, lastIP)
 		subSpace.Parent = s
 		return subSpace, nil
 	}
 
-	return nil, fmt.Errorf("could not find IP range")
+	var err error
+	if compareIP4(firstIP, s.Pool.FirstIP) > 0 && compareIP4(lastIP, s.Pool.LastIP) < 0 {
+		// IP range is within the pool but not found available
+		err = fmt.Errorf("Cannot reserve IP range [%s, %s].  Already in use", firstIP.String(), lastIP.String())
```
hickeng: error strings should not be capitalized, or contain sentence fragments. The following is much more concise (acronyms remain capitalized as its not contextual): 
```
IP range already in use: %s - %s
```

lib/portlayer/network/ipam.go
```diff
@@ -281,13 +283,25 @@ func (s *AddressSpace) ReserveIP4Range(firstIP net.IP, lastIP net.IP) (*AddressS
 		}
 
 		// found range
+		log.Infof("Reserving IP range [%s, %s]", firstIP.String(), lastIP.String())
```
hickeng: Should also be using op trace as there's no way to correlate at all - don't even have container ID in here.

lib/portlayer/network/ipam.go
sflxn: This is what I had originally.  I want to keep symmetry with the release call, which explicitly states it's releasing.

lib/portlayer/network/ipam.go
sflxn: In order for op to be useful here, I would need to weave op throughout the code path.  Given the time left, I didn't do it.

lib/apiservers/engine/backends/errors.go
```diff
@@ -111,6 +111,10 @@ func StreamFormatNotRecognized() error {
 	return derr.NewRequestConflictError(fmt.Errorf("Stream format not recognized"))
 }
 
+func ConcurrentAPIError(name, request string) error {
+	return derr.NewRequestConflictError(fmt.Errorf("A concurrent %s request is already in progress for container '%s'.  Please try again later.", request, name))
```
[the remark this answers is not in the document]
sflxn: Are we breaking the standard we have in errors.go just for this one PR?  Is this lower case an arbitrary request?

sflxn:
Using the PropertyCollectorTimeout for the API call timeout is too long.  3 mins.  We don't even allow that length of time for attach.  I bet Xenon (Admiral) will not wait that long to time out the requests to the persona.

lib/apiservers/engine/backends/errors.go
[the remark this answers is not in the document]
hickeng: Of course it's not an arbitrary request - I'm referring to the contributor guidelines and standard Go idioms. If we've already violated those in a systemic manner then consistency trumps it.

hickeng:
3 min was used based on observed behavior in heavily loaded environments - even then it still times out if VMs get vmotioned during start. If you use a timeout of less than you are likely giving up before the in flight start call has any chance to complete.

Note that start will block until the process reports success - in a vCenter environment that's impacted by the host sync interval, so anything under 20s + time to boot is almost assured of timing out early.

There also exists the strong possibility with short running containers that by the time this lock releases the container will have completed it's work and shut down meaning this call will this start will then run it a second time. I don't know if that's the desired behavior or not.

lib/apiservers/engine/backends/backends.go
vburenin: Please add a comment then that it is your best guess and it may need to be tuned in the future. Also it would be good to add why use have chosen 5. Why not 15 or 30?

lib/apiservers/engine/backends/container/viccontainer.go
```diff
@@ -65,3 +70,24 @@ func (v *VicContainer) List() []string {
...
+		return true
+	}
+}
+
+// Unlocks the container
+func (v *VicContainer) Unlock() {
+	select {
+	case <-v.lockChan:
+	default:
+		panic()
```
vburenin: Could you panic with comment? like panic("Attempt to unlock VicContainer more than once.")

sflxn:
ping

PR #6961, opened 2017-12-14
Log the sessions and hostd memory usage of the test system

mhagen-vmware:


hickeng:
Can this use keys instead? If it's using a password it'll fail completely if/when hostd gets into it's terminal state.
Unless that's what we're after an we make this a FATAL error to end the test?

mhagen-vmware:
I think it is a moot point at that point.  Once the host is in the state where login fails, then we already know it is in that state and memory has been exceeded. This is more to show while the system is healthy where the trends are going.

PR #6966, opened 2017-12-14
[specific ci=Group3-Docker-Compose] Revert govc change - Just Test

mdharamadas1:
Just to see if this is impacting `govc` issues on master.


hickeng:
#6998 addressed this

Issue #5668, opened 2017-07-06
Update project's README.md with new features


anchal-agrawal:
Our current README (particularly, the`Project Status` section) has not been updated in quite a while. We should add updates from our latest release(s) to it.

Estimated as a 2 since the work involved spans the entire project and will need input from all component owners.

**Acceptance Criteria:**
* [x] The `Project Status` section is updated with major features
* [x] Other sections are checked for validity and corrected if necessary

corrieb:
Agreed. I would recommend that project status be its own section where each release has its major features called out. This is where we can link to upcoming features too.

anchal-agrawal:
#6974 covers this - closing.

PR #6974, opened 2017-12-18
[skip ci] Update readme and contributing doc

anchal-agrawal:
This commit polishes the VIC Engine readme and contributing markdown
files and highlights some new key features in the readme's Project
Status section.

Towards #5668

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,12 +2,14 @@
 
 ## Community
 
-Contributors and users are encouraged to collaborate using the following resources in addition to the GitHub issue
-tracker:
+In addition to using the GitHub issue tracker, contributors and users are encouraged to collaborate using the following
+resources:
 
-- [Slack](https://vmwarecode.slack.com/messages/vic-engine): This is the primary community channel. If you don't have an @vmware.com or @emc.com email, please sign up at https://code.vmware.com/slack to get a Slack invite.
+- [Slack](https://vmwarecode.slack.com/messages/vic-engine): This is the primary community channel. **If you don't have
+an @vmware.com or @emc.com email, please sign up at https://code.vmware.com/join to get a Slack invite.**
 
-- [Gitter](https://gitter.im/vmware/vic): Gitter is monitored but go to Slack if you need a response quickly.
+- [Gitter](https://gitter.im/vmware/vic): Gitter is monitored, but please use the Slack channel if you need a response
+quickly.
 
 ## Getting started
 
@@ -51,7 +53,7 @@ git push $USER my-new-feature
 
 ### Stay in sync with upstream
 
-When your branch gets out of sync with the vmware/master branch, use the following to update:
+When your branch gets out of sync with the vmware/master branch, use the following to update it:
 
 ``` shell
 git checkout my-new-feature
@@ -88,7 +90,7 @@ notification when you git push.
 
 ### Code style
 
-The coding style suggested by the Golang community is used in VIC. See the
+VIC Engine uses the coding style suggested by the Golang community. See the
 [style doc](https://github.com/golang/go/wiki/CodeReviewComments) for details.
 
 Try to limit column width to 120 characters for both code and markdown documents such as this one.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -6,69 +6,63 @@ vSphere Integrated Containers Engine (VIC Engine) is a container runtime for vSp
...
+[wizard]:https://vmware.github.io/vic-product/assets/files/html/1.3/vic_vsphere_admin/deploy_vch_client.html
+[dch]:https://vmware.github.io/vic-product/assets/files/html/1.3/vic_vsphere_admin/deploy_vch_dchphoton.html
+
 VIC Engine now provides:
-* basic function for most of the core lifecycle operations: pull, create, start, attach, run, kill, stop, wait, rm, logs.
-* early vCenter support, leveraging DRS for initial placement.
-* volume support [--volume-store](doc/user/usage.md#configuring-volumes-in-a-virtual-container-host) - SIOC is not integrated but can be set as normal.
+* support for most of the Docker commands for core container, image, volume and network lifecycle operations. Several `docker compose` commands are also supported. See the complete list of supported commands [here][commands].
+* vCenter support, leveraging DRS for initial placement. vMotion is also supported.
+* volume support for vSphere datastores (including vSAN) and NFS shares. See [--volume-store](doc/user/usage.md#configuring-volumes-in-a-virtual-container-host) - SIOC is not integrated but can be set as normal.
```
anchal-agrawal: Ping @matthewavery for reviewing this part.

CONTRIBUTING.md
```diff
@@ -191,72 +193,75 @@ priority/p3 - Bugs that SHOULD be fixed by a given release, time permitting.
 
 priority/p4 - Bugs that SHOULD be fixed in a future (to be determined) release.
 
-### Backlog
+### Not Ready
```
anchal-agrawal: Ping @mhagen-vmware for reviewing these sections - these changes add details and guidelines about our pipeline lanes. The recent priority changes are unaffected.

CONTRIBUTING.md
```diff
@@ -191,72 +193,75 @@ priority/p3 - Bugs that SHOULD be fixed by a given release, time permitting.
...
+4. no assignee (assignees are set when the issue is selected to work on)
+
+Story Points | Story Size
+------------ | -------------------------------------------------------
+1            | Less than 1 day of effort
+2            | 2 - 3 days of effort
+3            | 3 - 5 days of effort
+5            | 5 - 10 days of effort, consider splitting this if it's 7 - 10 days
+8            | > 10 days, anything of this size should be split before moving into In Progress
+
```
mhagen-vmware: This is fundamentally not a good idea for several reasons, we should stick with our existing definitions of story points or we are not using story points at all and need to move to time based estimates ( which aren't good, thus Agile uses story points )

anchal-agrawal: I see. My change doesn't add this table, it's already there: https://github.com/vmware/vic/blob/master/CONTRIBUTING.md#in-progress. I've just moved it from the In Progress section to the Backlog section.

anchal-agrawal: Should we remove this table altogether?

mhagen-vmware: I would like to see it corrected to our agreed upon estimation process:
```
Story points
● We have not been consistent in story point estimation
● Proposal:
– Stick with fibonacci
– All bugs are a 2 unless we know it is significantly more or less work than the
average bug
– 1 is easier than the average bug
– 3 is slightly more work than the average bug and probably should be about
an average feature work for an easy feature (which includes design doc,
implementation, testing, review)
– 5 is about 2x more work than the average bug and the highest single issue
value we want
– >5 can be on EPICs, but single issues should be decomposed further if you
think they are more than a 5
```

anchal-agrawal: Fixed: https://github.com/anchal-agrawal/vic/blob/860c9b267b8c59d9dc6057613bfc982213a787e7/CONTRIBUTING.md#story-point-estimates

README.md
matthewavery: is the [here][commands] mean to be a hyperlink? if so It might be missing the link :)

matthewavery: `vsphere datastores` may not be quite correct. we do support standard vmfs backed datastores. We do support vsan datastores, NFS datastores, and ISCSI datastores. However, things like VVOL datastores and any other kind of specialty datastores are not supported. :)

anchal-agrawal: The `[commands]` hyperlink is defined here: https://github.com/vmware/vic/pull/6974/files#diff-04c6e90faac2675aa89e2176d2eec7d8R11

PR #6979, opened 2017-12-19
[skip ci] Move all jenkins logic into a trackable script and upload the results to GCP

mhagen-vmware:


tests/longevity-tests/jenkins-job.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+cd ~/vic
+
+git clean -fd
+git fetch https://github.com/vmware/vic master
+git pull
+
+cp ~/secrets .
+tests/longevity-tests/run-longevity.bash 6.0
```
anchal-agrawal: If I'm understanding correctly: should we also include the command for 6.5 in this script?

mhagen-vmware: yep, sorry meant to parameterize, should be good now

PR #6980, opened 2017-12-19
[skip ci] Make it a fatal error if vic machine fails to cleanup a VCH

### LINKED ISSUE #7040: Add a square bracket fast failure tag
Depending on what we are trying to test in a PR, we might care whether a single test fails or all possible failures.  When we only care whether a single test fails it would be useful to have a [fast fail] build time directive that will halt all other test immediately so that we can rapidly get output from failure we care about.

mhagen-vmware:
fixes #7040 
Also needs #7060 to be merged to complete

### DIFF
--- a/tests/resources/VCH-Util.robot
+++ b/tests/resources/VCH-Util.robot
@@ -345,7 +345,8 @@ Run Secret VIC Machine Inspect Command
 Run VIC Machine Delete Command
     ${rc}  ${output}=  Run Secret VIC Machine Delete Command  %{VCH-NAME}
     Log  ${output}
-    Wait Until Keyword Succeeds  6x  5s  Check Delete Success  %{VCH-NAME}
+    ${status}=  Run Keyword And Return Status  Wait Until Keyword Succeeds  6x  5s  Check Delete Success  %{VCH-NAME}
+    Assert  ${status}  vic-machine delete failed to remove %{VCH-NAME}
     Should Be Equal As Integers  ${rc}  0
     Should Contain  ${output}  Completed successfully
     ${output}=  Run  rm -rf %{VCH-NAME}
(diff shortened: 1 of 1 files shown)


tests/resources/VCH-Util.robot
```diff
@@ -322,7 +322,8 @@ Run Secret VIC Machine Inspect Command
 
 Run VIC Machine Delete Command
     ${rc}  ${output}=  Run Secret VIC Machine Delete Command  %{VCH-NAME}
-    Wait Until Keyword Succeeds  6x  5s  Check Delete Success  %{VCH-NAME}
+    ${status}=  Run Keyword And Return Status  Wait Until Keyword Succeeds  6x  5s  Check Delete Success  %{VCH-NAME}
+    Run Keyword Unless  ${status}  Fatal Error  vic-machine delete has failed to remove the VCH: %{VCH-NAME} as expected
```
hickeng: Is there any way we could have a suite/group cleanup check?
In both the vic-machine-service tests and the delete-resource-pool tests we're leaking because we don't call vic-machine delete, not because it fails.

tests/resources/VCH-Util.robot
```diff
@@ -322,7 +322,8 @@ Run Secret VIC Machine Inspect Command
 
 Run VIC Machine Delete Command
     ${rc}  ${output}=  Run Secret VIC Machine Delete Command  %{VCH-NAME}
-    Wait Until Keyword Succeeds  6x  5s  Check Delete Success  %{VCH-NAME}
+    ${status}=  Run Keyword And Return Status  Wait Until Keyword Succeeds  6x  5s  Check Delete Success  %{VCH-NAME}
+    Run Keyword Unless  ${status}  Conditional Fatal Error  vic-machine delete has failed to remove the VCH: %{VCH-NAME} as expected
```
hickeng: This might be a request for a different Keyword, but I would more naturally phrase this with the same imperatives used in unit tests:
* assert
* require

Assert is a failure, require is a fatal failure. In this case we're changing the behaviour of assert to control whether it become fatal, so it wraps both the check of status and the appropriate handling on failure into a single keyword:
```
${status}=  Run Keyword And Return Status  Wait Until Keyword Succeeds  6x  5s  Check Delete Success  %{VCH-NAME}
Assert  ${status}  "vic-machine delete failed to remove %{VCH-NAME}"
```

sgairo:
Should the [skip ci] be moved to the body?

mhagen-vmware:
@sgairo [skip ci] needs to be in the title still:
https://github.com/vmware/vic/blob/master/CONTRIBUTING.md#automated-testing

Issue #6581, opened 2017-10-18
Container missing after vMotion


gigawhitlocks:
Seen in 13-2 vMotion-Container on 17 October 17
![screen shot 2017-10-18 at 1 07 29 pm](https://user-images.githubusercontent.com/3743903/31734806-7b895bca-b405-11e7-9f6b-71e723bfb810.png)

Full logs are in the 6.0 failure section of the log bundle here:
https://storage.cloud.google.com/vic-ci-logs/functional_logs_10_16_06_09_vic_14037.tar.gz.zip?authuser=1

matthewavery:
seen during nightly VC 6.5 13-2-vMotion failure on 11/4/2017:
VIC Engine Build 14394

[13-2-vMotion-Container.zip](https://github.com/vmware/vic/files/1447044/13-2-vMotion-Container.zip)

sgairo:
Seen in both: 
`Oct 17 2017 17:58:18.380Z ERROR Handler for POST /v1.25/containers/75b90a578eb15501985726963ee6ac70aebbab3eee25ee611f9ee5e991bd5b1d/stop returned error: No such container: 75b90a578eb15501985726963ee6ac70aebbab3eee25ee611f9ee5e991bd5b1d`

sgairo:
There is nothing particularly alarming from the logs, no signs of issue or abnormality in persona or portlayer until the actual failure. The second container that is tested after a vMotion was started ~prior~ to the vMotion and couldn't be found to stop afterwards. Also used the same image as the first container which was started, stopped, and removed ~after~ the vMotion, but before the second container was tested.

sgairo:
Am re-running and trying to reproduce.

sgairo:
We will have to wait for another occurrence of this

mlh78750:
If this continues to no-repo, let's close and reopen if it does show again.

PR #6989, opened 2017-12-20
[WIP] Add concurrent access error to network

sflxn:
Add concurrent access error as a return result from
network reservation.  This should allow the persona
to retry on 'IP already in use' error instead of
just returning the error to the API caller.

Also removed more useless logging clogging up the logs.

Towards #6581

### DIFF
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -1470,13 +1470,11 @@ func processVolumeFields(volumes []string) (map[string]volumeFields, error) {
 }
 
 func finalizeVolumeList(specifiedVolumes, anonymousVolumes []string) ([]volumeFields, error) {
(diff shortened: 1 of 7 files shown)


lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1470,13 +1470,11 @@ func processVolumeFields(volumes []string) (map[string]volumeFields, error) {
 }
 
 func finalizeVolumeList(specifiedVolumes, anonymousVolumes []string) ([]volumeFields, error) {
-	log.Infof("Specified Volumes : %#v", specifiedVolumes)
```
matthewavery: I would rather see this logging changed to debug instead of removed as it indicates the volume requirements requested of the personality during a create call.

zjs:
Closing in favor of #7128.

PR #6994, opened 2017-12-20
[specific ci=Group23-VIC-Machine-Service] Remove path from target URL in vic-machine-service API

### LINKED ISSUE #6978: VCH created by wizard docker endpoint and admin portal unavailable: docker persona hangs
**Details**
Create a VCH with the wizard, wait after docker endpoint and admin portal to come up. Once they're available, connecting to docker endpoint via `docker -H VCH_IP:2376 --tls info` gives:
```Cannot connect to the Docker daemon at tcp://10.192.124.47:2375. Is the docker daemon running?```

Also, the admin portal cannot authenticate.

Deploy again with the generated CLI command, it succeeds and the VCH is up and running (docker endpoint and admin portal both work alright).

The generated CLI command:
```
./vic-machine-linux create --name virtual-container-host --debug 1 --compute-resource /RegionA01/host/RegionA01-COMP01 --image-store RegionA01-ISCSI01-COMP01 --base-image-size 8GB --bridge-network Bridge01-RegionA01-vDS-COMP --bridge-network-range 172.16.0.0/12 --public-network VM-RegionA01-vDS-COMP --no-tlsverify --user Administrator@CORP.LOCAL --thumbprint F7:68:F0:93:F4:EC:B7:FE:C1:02:3F:F3:AB:62:1A:50:E8:9A:0E:85 --target vcsa-01a.corp.local/RegionA01 --ops-user administrator@corp.local
```

**Logs**
From the docker-persona logs, portlayer never comes up, so the persona hangs waiting for portlayer:
```
time="2017-12-18T20:25:08Z" level=info msg="Launching docker personality pprof server on 127.0.0.1:6062" 
Dec 18 2017 20:25:09.331Z DEBUG log cfg: {Formatter:0xc4200730a0 Level:debug Syslog:<nil>}
Dec 18 2017 20:25:09.346Z DEBUG 296.1: [OperationFromContext] [vic/pkg/vsphere/session.(*Session).Connect:191]
Dec 18 2017 20:25:09.346Z DEBUG op=296.1: Creating VMOMI session with thumbprint F7:68:F0:93:F4:EC:B7:FE:C1:02:3F:F3:AB:62:1A:50:E8:9A:0E:85
Dec 18 2017 20:25:09.843Z DEBUG Session Environment Info: 
Dec 18 2017 20:25:09.896Z DEBUG Loaded 180 CAs for registries from system CA bundle
Dec 18 2017 20:25:09.901Z ERROR Unable to load CAs for registry access in config
Dec 18 2017 20:25:09.909Z INFO  Waiting for portlayer to come up
```

Portlayer gives bad request trying to connect:
```
Dec 18 2017 21:53:38.248Z DEBUG GuestInfoSource: key: guestinfo.vice./grant_permissions, value: "", error: %!s(<nil>)
Dec 18 2017 21:53:38.248Z DEBUG GuestInfoSource: key: guestinfo.vice./vic_machine_create_options, value: "", error: key not found
Dec 18 2017 21:53:38.248Z DEBUG GuestInfoSource: key: guestinfo.vice./vic_machine_create_options~, value: "", error: key not found
Dec 18 2017 21:53:38.248Z DEBUG op=2108.1: Creating VMOMI session with thumbprint F7:68:F0:93:F4:EC:B7:FE:C1:02:3F:F3:AB:62:1A:50:E8:9A:0E:85
Dec 18 2017 21:53:38.969Z FATAL configure_port_layer ERROR: Failed to connect to vcsa-01a.corp.local: 400 Bad Request
```

The log bundle: 
[vch_logs_20171218.zip](https://github.com/vmware/vic/files/1569660/vch_logs_20171218.zip)

Since the CLI works, there's probably something in the API that caused this error, might be related to network configuration. 
Cc: @zjs

### PR DESCRIPTION
**Pull request by AngieCris**
Fixes #6978 , also hopefully fixes https://github.com/vmware/vic-ui/issues/257

The root causes of the issues are:
1. path is not stripped from the target URL when creating a VCH
2. datacenter in validator session config is never set
3. typo on assigning datacenter path to datastore path

So the fixes are: 
1. remove the path from target URL
2. set datacenter in validator session to the management object
3. fix the typo

Also a simple VCH initialization check (docker connectivity check) is added to the API create robot test

### DIFF
--- a/lib/apiservers/service/restapi/handlers/common.go
+++ b/lib/apiservers/service/restapi/handlers/common.go
@@ -93,11 +93,10 @@ func buildDataAndValidateTarget(op trace.Operation, params buildDataParams, prin
 			return data, nil, util.NewError(http.StatusBadRequest, "Validation Error: datacenter parameter is not a datacenter moref")
 		}
 
-		data.Target.URL.Path = dc.InventoryPath
-
 		// Do what NewValidator would have done if a Datacenter had been set
 		v.DatacenterPath = dc.Name()
-		v.Session.DatastorePath = v.DatacenterPath
+		v.Session.DatacenterPath = v.DatacenterPath
+		v.Session.Datacenter = dc
 		v.Session.Finder.SetDatacenter(dc)
 
 		validator = v
--- a/lib/apiservers/service/restapi/handlers/common.go
+++ b/lib/apiservers/service/restapi/handlers/common.go
@@ -93,7 +93,7 @@ func buildDataAndValidateTarget(op trace.Operation, params buildDataParams, prin
 			return data, nil, util.NewError(http.StatusBadRequest, "Validation Error: datacenter parameter is not a datacenter moref")
 		}
 
-		// Do what NewValidator would have done if a Datacenter had been set
+		// Set validator datacenter path and correspondent validator session config
 		v.DatacenterPath = dc.Name()
 		v.Session.DatacenterPath = v.DatacenterPath
 		v.Session.Datacenter = dc
--- a/tests/test-cases/Group23-VIC-Machine-Service/23-03-VCH-Create.robot
+++ b/tests/test-cases/Group23-VIC-Machine-Service/23-03-VCH-Create.robot
@@ -280,3 +280,11 @@ Fail to create VCH without network
 
     Output Should Contain    network
 
+
+Fail to create VCH with gateway without static address
+    Create VCH    '{"name":"%{VCH-NAME}-api-bad-gateway","compute":{"resource":{"name":"%{TEST_RESOURCE}"}},"storage":{"image_stores":["ds://%{TEST_DATASTORE}"]},"network":{"bridge":{"ip_range":"172.16.0.0/12","port_group":{"name":"%{BRIDGE_NETWORK}"}},"public":{"port_group":{"name":"${PUBLIC_NETWORK}"},"gateway":{"address":"127.0.0.1","routing_destinations":[]}}},"auth":{"server":{"generate":{"cname":"vch.example.com","organization":["VMware, Inc."],"size":{"value":2048,"units":"bits"}}},"client":{"no_tls_verify": true}}}'
+
+    Verify Return Code
+    Verify Status Bad Request
+
+    Output Should Contain    static
--- a/tests/test-cases/Group23-VIC-Machine-Service/23-03-VCH-Create.robot
+++ b/tests/test-cases/Group23-VIC-Machine-Service/23-03-VCH-Create.robot
@@ -78,7 +78,6 @@ Verify VCH Initialization ${name}
     Wait Until Keyword Succeeds  12x  10 seconds  Get VCH Docker Info  ${name}
 
 
-
 *** Test Cases ***
 Create minimal VCH
     Create VCH    '{"name":"%{VCH-NAME}-api-test-minimal","compute":{"resource":{"name":"%{TEST_RESOURCE}"}},"storage":{"image_stores":["ds://%{TEST_DATASTORE}"]},"network":{"bridge":{"ip_range":"172.16.0.0/12","port_group":{"name":"%{BRIDGE_NETWORK}"}},"public":{"port_group":{"name":"${PUBLIC_NETWORK}"}}},"auth":{"server":{"generate":{"cname":"vch.example.com","organization":["VMware, Inc."],"size":{"value":2048,"units":"bits"}}},"client":{"no_tls_verify": true}}}'
--- a/tests/test-cases/Group23-VIC-Machine-Service/23-03-VCH-Create.robot
+++ b/tests/test-cases/Group23-VIC-Machine-Service/23-03-VCH-Create.robot
@@ -51,31 +51,11 @@ Inspect VCH Config ${name}
     Set Test Variable    ${OUTPUT}
 
 
-Inspect VCH
-    [Arguments]  ${name}  ${expected}
-    ${rc}    ${output}=    Run And Return Rc And Output  bin/vic-machine-linux inspect --name=${name} --target=%{TEST_URL} --thumbprint=%{TEST_THUMBPRINT} --user=%{TEST_USERNAME} --password=%{TEST_PASSWORD} --compute-resource=%{TEST_RESOURCE}
-    Should Be Equal As Integers  ${rc}  0
-    Should Contain  ${output}  ${expected}
-    [Return]  ${output}
-
-
 Get VCH ${name}
     Get Path Under Target    vch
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by mhagen-vmware** [comment]
While nimbus would be nice I feel like this can just be testing using the default ha-datacenter within the CI, as long as you explicitly specify it in the test.

**Comment by AngieCris** [comment]
> I would like to know why the existing `Create minimal VCH within datacenter` did not show this issue - I'm presuming it's because the Session logic selected a default datacenter given there's only one?

@hickeng the issue is about docker connectivity - the VCH will be successfully created, but we will only know that the VCH is broken until we use docker. API create doesn't do connectivity check like CLI. So the test passes every time, since it only creates a VCH and doesn't do any docker connectivity check.

From rc3 to rc4, on the UI side we decide to use datacenter specific endpoint no matter what. Before rc4, if no datacenter is specified, we just use non-datacenter specific endpoint. After rc4, we just select a default datacenter if there's none. 
The bug on API is only in the datacenter specific endpoint (for non-datacenter specific endpoint, the handler sets the `validator.AllowEmptyDC` to true). Since the UI always uses datacenter specific endpoint, the docker connectivity issue happens every time. 
@anchal-agrawal I was misled at first when I thought the issue only happened in multi-dc. It happens in single-dc as well. Whether it's multi-dc or single-dc, the user selects a dc they want to deploy their VCH on, and the request gets sent to the dc specific endpoint on API, and the VCH broke. 
So whether it's multi-dc or single-dc, the problem remains the same.

For the test, I'm thinking about adding a docker connectivity check with `docker info` to the API create integration test. @hickeng What do you think? Can you clarify more on nimbus multi-dc path? 

Sidenote:
This bug would have been caught if the test checked docker connectivity. I tried to enable docker regression test several months back, but the current docker regression test sets docker env params by parsing CLI command line output. This sounds super weird to me and I couldn't get it working. So we lowered that priority.
If it's worth it, I'll open up a ticket tracking enabling docker regression test on API integration test.

**Comment by AngieCris** [comment]
@hickeng @mhagen-vmware @anchal-agrawal 
I added docker connectivity check in API create integration test. Can you guys take a look? Thanks!

File: tests/test-cases/Group23-VIC-Machine-Service/23-03-VCH-Create.robot [hunk 1]
```diff
@@ -134,7 +158,7 @@ Create complex VCH
     Verify Status Created
 
 
-    Inspect VCH %{VCH-NAME}-api-test-complex
+    Inspect VCH Config %{VCH-NAME}-api-test-complex
```
anchal-agrawal: Should we run `Verify VCH Initialization` for the `Create complex VCH` case as well?

AngieCris: The purpose of `Create Complex VCH` is to check if all the VCH create params goes into VCH config, so some input params are fake just for testing purposes, like the `https-proxy` and `container-network` etc. So the VCH created by this test probably can't connect to docker endpoint...
Testing docker connectivity for the test case `Create minimal VCH within datacenter` should be sufficient, since the problem was with datacenter specification

anchal-agrawal: Thanks! It wasn't clear from the test whether this VCH would be functional.

AngieCris: Yea...on API side we reused a lot of code and logic in CLI, so we assumed that if all the VCH create params on API successfully goes into VCH config, the VCH should be identically functional as created by CLI. That's why the API create integration test is not complete (for ex, no docker regression test)

emlin: @AngieCris is it better to add some comment in the test.md file?

lib/apiservers/service/restapi/handlers/common.go
```diff
@@ -93,11 +93,10 @@ func buildDataAndValidateTarget(op trace.Operation, params buildDataParams, prin
 			return data, nil, util.NewError(http.StatusBadRequest, "Validation Error: datacenter parameter is not a datacenter moref")
 		}
 
-		data.Target.URL.Path = dc.InventoryPath
-
 		// Do what NewValidator would have done if a Datacenter had been set
 		v.DatacenterPath = dc.Name()
-		v.Session.DatastorePath = v.DatacenterPath
+		v.Session.DatacenterPath = v.DatacenterPath
```
anchal-agrawal: Since this is a nuanced fix, could we add a comment explaining why setting these two fields is critical?

tests/test-cases/Group23-VIC-Machine-Service/23-03-VCH-Create.robot
AngieCris: @emlin Good point. Thanks!

lib/apiservers/service/restapi/handlers/common.go
AngieCris: Good catch. Thanks!

mdubya66:
Lgtm

tests/test-cases/Group23-VIC-Machine-Service/23-03-VCH-Create.robot
```diff
@@ -45,19 +45,39 @@ Create VCH Within Datacenter
...
+
+
 Get VCH ${name}
     Get Path Under Target    vch
     ${id}=    Run    echo '${OUTPUT}' | jq -r '.vchs[] | select(.name=="${name}").id'
 
     Get Path Under Target    vch/${id}
 
 
+Get VCH Docker Info
```
hickeng: These should really live in the VCH-Util library, and unified with the unparameterized versions (we can use the default values mechanism to preserve existing behaviour). Thanks for the modification!

AngieCris: Yes I think they should be moved too. I can move both the `Get VCH Docker Info` and `Inspect VCH` to util library, so they can co-exist with the unparameterized ones to preserve existing behavior.

Having two keywords of essentially the same functionality seems to be the only way to go for now. Re-using the keywords from util library has been hard because:
1. In the API create robot test, the `VCH-NAME` test env var is not reset for every test, so we can't use the unparameterized (using env vars) versions from util library 
2. The `VCH Docker Info` provided in util library calls inspect without specifying `compute-resource`, and it fails in our case with VCH created by API.

mhagen-vmware: Should be able to just default to %{TEST_RESOURCE}

AngieCris: Do we want to run a full CI to check or we can just assume that existing behaviors are preserved

PR #7000, opened 2017-12-21
[specific ci=1-05-Docker-Start] Use event types to filter events

dougm:
An alternative to increasing the collector page size.  It will reduce the throughput to the event collector and hence reduce event misses.

See issues #6937 and #6998


hickeng:
I would like to see this run against `1-05-Docker-Start` - I'm seeing reliable failures of that test in #6956 but without any changes I can easily identify that would trigger it.
That PR does have a govmomi vendor update however so I'd like to confirm that's not the proximate cause.

dougm:
Merged via #7001 and #7009

PR #7001, opened 2017-12-22
Reduce hostd mem usage via event type filter

hickeng:
This bumps eventPageSize back to 200 to determine if the event type filter has significant effect on memory footprint of task history.
Having run full CI against this branch we saw a maximum of `124880` which is directly inline with the numbers found for 1.2.1 full CI and an unfiltered pagesize of 25.
The fact these two numbers are almost identical implies to me that we're not saturating the 25 pagesize in CI and in fact the number is dominated at this time by the taskRetention interval in hostd.

https://ci.vcna.io/vmware/vic/15511 is the CI build used for reference. There is a failure in the log test which looks like it's related to either a local stall or clock skew - will confirm once logs are uploaded.

@dougm has tests for the filtering in vmware/govmomi and I've confirmed that we're getting only the expected events delivered. The `docker ps` and out-of-band tests already exercise the function that events support.

This is based off #7000 and should be merged instead of that, or after having dropped the govmomi portion of this PR.

Follow up for : #6886 and #6384


mdubya66:
LGTM

mdubya66:
Merging to close the issue.

PR #7004, opened 2017-12-25
Enable docker -D info integration test

### LINKED ISSUE #780: docker -D info command not supported
**VIC Version:**

master

**Code**

```
docker -H <VIC> info
docker -H <VIC> -D info
```

**Expected behavior:**
It is expected that additional debug information will be provided when passing the -D flag to info command.

**Actual behavior:**
The flag is ignored and prints the same output as it would without the flag.

anchal-agrawal:
This commit enables the 'docker -D info' integration test in the
1-01-Docker-Info suite. The -D flag enables debug mode for the docker
client - it is not passed to the info command.

Fixes #780


mdubya66:
Merged for Anchal as Pull approve was acting up

Issue #7018, opened 2018-01-02
Fix report-coverage CI step


mdharamadas1:
Step `report-coverage` is failing with below error and needs to be fixed. Until that it will be disabled on CI.

https://ci.vcna.io/vmware/vic/15526
`This is an outdated version of Coverage and is expired. Please upgrade.Unable to authenticate user. Error fetching user. GET https://api.github.com/user: 401 Bad credentials []`

mhagen-vmware:
https://github.com/vmware/vic/issues/4960

PR #7019, opened 2018-01-02
Disable report coverage CI step

mdharamadas1:
Disables the `report-coverage` step which will be fixed in #7018.


matthewavery:
lgtm

PR #7042, opened 2018-01-05
Add copyright notice to new file

mhagen-vmware:


anchal-agrawal:
Looks like we'll need to update https://github.com/vmware/vic/blob/master/infra/scripts/header-check.sh as well.

matthewavery:
the `jenkins-job.sh` as well

mhagen-vmware:
I just updated the template as well for new files

PR #7057, opened 2018-01-09
Volume mount permission fix

### LINKED ISSUE #6881: Directory permissions for volume mounts
**Story**
As a VCH user I expect to be able to use volumes as a non-root user without issue

**Details**
We currently have the following behavioural deviations:
1. `MkdirAll("/volume/mount/point")` should be creating any missing directories with `0755` instead of `0644`
2. The permissions of the mount point should be preserved once the volume is mounted. Currently if we have `/mnt/vol` with `0775` then it gets changed to `0755` when the volume is mounted.

I'm unclear on the exact semantics of (2) with regular docker - it looks like `chown/chmod` are used every time the volume is mounted (even if in use else where) and the permissions are updated in some conditional fashion.

Sample - vo6 is a new volume. `voltest` is from the following Dockerfile:
```
FROM busybox
RUN mkdir -p /volumes/user && chmod 775 /volumes/user && chown 2001 /volumes/user
```
result:
```
vagrant@devbox:/tmp/tmp$ docker run  -v vo6:/volumes/user -it busybox
/ # ls -l /volumes
total 4
drwxr-xr-x    2 root     root          4096 Nov 30 19:04 user
/ # vagrant@devbox:/tmp/tmp$ #detached
vagrant@devbox:/tmp/tmp$ docker run  -v vo6:/volumes/user -it voltest
/ # ls -l /volumes
total 4
drwxrwxr-x    2 2001     root          4096 Nov 30 19:04 user
/ # exit
vagrant@devbox:/tmp/tmp$ docker ps
CONTAINER ID        IMAGE                                                   COMMAND                  CREATED             STATUS              PORTS                    NAMES
24a67b033af1        busybox                                                 "sh"                     32 seconds ago      Up 31 seconds                                gracious_mestorf
vagrant@devbox:/tmp/tmp$ docker attach gracious_mestorf
/ # ls -l /volumes
total 4
drwxrwxr-x    2 2001     root          4096 Nov 30 19:04 user
/ #
```
as can be see the owner of the volume was changed to `2001` for all containers, however it's not reset to `root` when using `docker run  -v vo6:/volumes/user -it busybox` after the above.

The necessary code changes are likely simple, but we need to confirm the appropriate conditional syntax for applying permissions a user/group ownership.

vburenin:
Bunch of permissions which were set for some directories before simply make no sense due to the missed eXecutable permission. This patch addresses that.

As soon as volume is mounted the original target directory permissions are applied after mount is done.

Note: If you are familiar with any of that, please let me know if this change can impact anything.

Fixes #6881


lib/tether/ops_linux.go
```diff
@@ -778,7 +778,7 @@ func (t *BaseOperations) MountLabel(ctx context.Context, label, target string) e
 	fi, err := os.Stat(target)
 	if err != nil {
 		// #nosec
-		if err := os.MkdirAll(target, 0744); err != nil {
+		if err := os.MkdirAll(target, 0755); err != nil {
```
matthewavery: if we are changing all of these to `0755` could you make a constant for that value.

vburenin: I thought about that. I am not sure it is a constant type of value. It is a specific permission that I better see what it is instead of decoding a constant name.

PR #7059, opened 2018-01-09
Include vic-machine server log file in integration test log bundle [specific ci=Group23-VIC-Machine-Service]

### LINKED ISSUE #7051: Failure in Group23-VIC-Machine-Service.23-01-Metadata
looks like the sleep for 1 second for the service to come up wasn't enough, but I can't really find any logs for the vic-machine service to debug further, but shortly after not responding to get version it did respond to get hello

As part of whatever fix we might implement here, we need to make sure we are capturing the log within our log bundle or at least within the test.

Here is the build:
https://ci.vcna.io/vmware/vic/15572

Here is our log bundle:
https://console.cloud.google.com/m/cloudstorage/b/vic-ci-logs/o/integration_logs_15572_4c17da525ab5a5422c018ddd6f505769ddd71f1c.zip?authuser=1

### LINKED ISSUE #7052: Include vic-machine-server log in integration test log bundle
When running vic-machine-service, the log file is located in server machine at /var/log/vic-machine-server/vic-machine-server.log, but it's not included in the integration test bundle. It's hard to debug from integration test result if there is a failure during Group 23 vic-machine-service tests.

AngieCris:
Fixes #7052 
Fixes #7051 

Changes:
1. Include the vic-machine-server log file in the integration log bundle
2. Add retries in checking version endpoint in 23-01-Metadata integration test

### DIFF
--- a/tests/integration-test.sh
+++ b/tests/integration-test.sh
@@ -47,12 +47,7 @@ rc="$?"
 timestamp=$(date +%s)
 outfile="integration_logs_"$DRONE_BUILD_NUMBER"_"$DRONE_COMMIT".zip"
 
-if [ -f "/var/log/vic-machine-server/vic-machine-server.log" ]
-then
-  zip -9 $outfile output.xml log.html report.html package.list *container-logs.zip *.log /var/log/vic-machine-server/vic-machine-server.log 
-else
-  zip -9 $outfile output.xml log.html report.html package.list *container-logs.zip *.log
-fi
+zip -9 $outfile output.xml log.html report.html package.list *container-logs.zip *.log /var/log/vic-machine-server/vic-machine-server.log 
 
 # GC credentials
 keyfile="/root/vic-ci-logs.key"
--- a/tests/integration-test.sh
+++ b/tests/integration-test.sh
@@ -47,7 +47,12 @@ rc="$?"
 timestamp=$(date +%s)
 outfile="integration_logs_"$DRONE_BUILD_NUMBER"_"$DRONE_COMMIT".zip"
 
-zip -9 $outfile output.xml log.html report.html package.list *container-logs.zip *.log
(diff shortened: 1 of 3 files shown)


tests/integration-test.sh
```diff
@@ -47,7 +47,12 @@ rc="$?"
 timestamp=$(date +%s)
 outfile="integration_logs_"$DRONE_BUILD_NUMBER"_"$DRONE_COMMIT".zip"
 
-zip -9 $outfile output.xml log.html report.html package.list *container-logs.zip *.log
+if [ -f "/var/log/vic-machine-server/vic-machine-server.log" ]
+then
+  zip -9 $outfile output.xml log.html report.html package.list *container-logs.zip *.log /var/log/vic-machine-server/vic-machine-server.log 
+else
+  zip -9 $outfile output.xml log.html report.html package.list *container-logs.zip *.log
+fi
```
mhagen-vmware: zip actually just works on whatever file exists, so I don't think you need this new if/then block:
```
anon@anon-Latitude-E7450:~$ zip -9 test.zip test.robot doesntexit.log
	zip warning: name not matched: doesntexit.log
  adding: test.robot (deflated 59%)
anon@anon-Latitude-E7450:~$ ls test.zip
test.zip
```

AngieCris: Awesome thanks!

AngieCris:
@mhagen-vmware 
The vic-machine server log file is under directory `/var/log/vic-machine-server`, without `-j` flag zip would zip the whole path (the zipped files would include directories `var` and `log` as well, just feel like it doesn't look very good that way)
`Get Version` keyword only posts path under a URL target using curl, it doesn't check if the URL target is reachable. That means if the service isn't up, curl will return error code 7 and the keyword still counts as success. The `Verify Return Code` checks availability by verifying if curl returns 0.

PR #7060, opened 2018-01-09
Move build time directives from title to body of PR

### LINKED ISSUE #7038: Change square bracket tagging behavior to work in description and not title of PRs
PRs with full ci and specific ci still in the title after merge are polluting our git log history so we need to change the behavior to look in the body instead of the title of PRs so that people will no longer include it in the title.

mhagen-vmware:
fixes #7038

### DIFF
--- a/tests/integration-test.sh
+++ b/tests/integration-test.sh
@@ -22,6 +22,14 @@ dpkg -l > package.list
 
 
 buildinfo=$(drone build info vmware/vic $DRONE_BUILD_NUMBER)
+prNumber=$(drone build info --format "{{ .Ref }}" vmware/vic $DRONE_BUILD_NUMBER | cut -f 3 -d'/')
+prBody=$(curl https://api.github.com/repos/vmware/vic/pulls/$prNumber | jq -r ".body")
+
+if (echo $prBody | grep -q "\[fast fail\]"); then
+    export FAST_FAILURE=1
+else
+    export FAST_FAILURE=0
+fi
 
 if [[ $DRONE_BRANCH == "master" || $DRONE_BRANCH == "releases/"* ]] && [[ $DRONE_REPO == "vmware/vic" ]] && [[ $DRONE_BUILD_EVENT == "push" ]]; then
     echo "Running full CI for $DRONE_BUILD_EVENT on $DRONE_BRANCH"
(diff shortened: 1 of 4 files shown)


tests/integration-test.sh
```diff
@@ -22,19 +22,21 @@ dpkg -l > package.list
...
     pybot --removekeywords TAG:secret --suite Group11-Upgrade --suite 7-01-Regression tests/test-cases
-elif grep -q "\[full ci\]" <(drone build info vmware/vic $DRONE_BUILD_NUMBER); then
+elif grep -q "\[full ci\]" <$prBody; then
     echo "Running full CI as per commit message"
     pybot --removekeywords TAG:secret --exclude skip tests/test-cases
-elif (echo $buildinfo | grep -q "\[specific ci="); then
+elif (echo $prBody | grep -q "\[specific ci="); then
     echo "Running specific CI as per commit message"
-    buildtype=$(echo $buildinfo | grep "\[specific ci=")
+    buildtype=$(echo $prBody | grep "\[specific ci=")
```
mdharamadas1: I think it would be good if this can also run multiple specific groups if someone wants to run more than one groups.

tests/integration-test.sh
```diff
@@ -22,19 +22,21 @@ dpkg -l > package.list
...
+prBody=$(curl https://api.github.com/repos/vmware/vic/pulls/$prNumber | jq -r ".body")
 
 if [[ $DRONE_BRANCH == "master" || $DRONE_BRANCH == "releases/"* ]] && [[ $DRONE_REPO == "vmware/vic" ]] && [[ $DRONE_BUILD_EVENT == "push" ]]; then
     echo "Running full CI for $DRONE_BUILD_EVENT on $DRONE_BRANCH"
     pybot --removekeywords TAG:secret --exclude skip tests/test-cases
 elif [[ $DRONE_BRANCH == *"refs/tags"* ]] && [[ $DRONE_REPO == "vmware/vic" ]] && [[ $DRONE_BUILD_EVENT == "tag" ]]; then
     echo "Running only Group11-Upgrade and 7-01-Regression for $DRONE_BUILD_EVENT on $DRONE_BRANCH"
     pybot --removekeywords TAG:secret --suite Group11-Upgrade --suite 7-01-Regression tests/test-cases
-elif grep -q "\[full ci\]" <(drone build info vmware/vic $DRONE_BUILD_NUMBER); then
+elif grep -q "\[full ci\]" <$prBody; then
```
zjs: I don't think this is valid bash syntax. I think you'll want to rewrite it as `(echo $prBody | grep -q "\[full ci\]")`.

gigawhitlocks:
:ship: :it: LTGM

tests/integration-test.sh
matthewavery: @anchal-agrawal has commented on this in the past. This actually gets passed in as a raw string, so if commas or the pybot delimiter is put in properly it already can. correct me if I am wrong @anchal-agrawal :)

anchal-agrawal: Correct! Multiple groups/suites can be specified by saying, for instance, `[specific ci=foo bar]`.

tests/integration-test.sh
```diff
@@ -22,19 +22,27 @@ dpkg -l > package.list
...
+fi
 
 if [[ $DRONE_BRANCH == "master" || $DRONE_BRANCH == "releases/"* ]] && [[ $DRONE_REPO == "vmware/vic" ]] && [[ $DRONE_BUILD_EVENT == "push" ]]; then
     echo "Running full CI for $DRONE_BUILD_EVENT on $DRONE_BRANCH"
     pybot --removekeywords TAG:secret --exclude skip tests/test-cases
 elif [[ $DRONE_BRANCH == *"refs/tags"* ]] && [[ $DRONE_REPO == "vmware/vic" ]] && [[ $DRONE_BUILD_EVENT == "tag" ]]; then
     echo "Running only Group11-Upgrade and 7-01-Regression for $DRONE_BUILD_EVENT on $DRONE_BRANCH"
     pybot --removekeywords TAG:secret --suite Group11-Upgrade --suite 7-01-Regression tests/test-cases
-elif grep -q "\[full ci\]" <(drone build info vmware/vic $DRONE_BUILD_NUMBER); then
+elif (echo $prBody | grep -q "\[full ci\]"); then
```
hickeng: Any chance we could have these match for the following for example:
- [x] run full CI
- [x] run unit tests
- [x] run group23
```
- [x] <!-- CI-DIRECTIVE:full ci -->run full CI
- [x] <!-- CI-DIRECTIVE:unit tests -->run unit tests
- [x] <!-- CI-DIRECTIVE:Group23-VIC-Machine-Service -->run group23
```
This would make it quick and easy to switch between multiple groups or focused groups/tests.
Admittedly most PRs won't really benefit that much from it, but I would have found it really useful for my most recent few.

I'm envisaging some staple ones added to the template, and then provision for customization

hickeng:
@mhagen-vmware think we should just merge this if it's functional, and enhancements to come later.

PR #7062, opened 2018-01-09
Simplify vch log API handler

### LINKED ISSUE #6733: VCH Creation Log APIs: simplify getAllLogFilePaths
See https://github.com/vmware/vic/pull/6665#discussion_r149573903:
> If you look at the implementation of `helper.Ls` it would be trivial to change it so that you can actually provide a pattern instead of a static name. There is already a MatchPattern defined as part of the search spec.
> That would basically eliminate the need for this function.

AngieCris:
Fixes #6733 

`[specific ci=Group23-VIC-Machine-Service]`

Modified `pkg/vsphere` datastore LS function to take a search pattern as optional argument. The log handler can provide a search pattern instead, so no need to grab all files.


lib/apiservers/service/restapi/handlers/vch_log_get.go
```diff
@@ -139,35 +128,27 @@ func getDatastoreHelper(op trace.Operation, d *data.Data, validator *validate.Va
 	return helper, nil
 }
 
-// getAllLogFilePaths returns a list of all log file paths under datastore folder, errors out when no log file found
-func getAllLogFilePaths(op trace.Operation, helper *datastore.Helper) ([]string, error) {
-	res, err := helper.Ls(op, "")
+// getAllLogs downloads all log files in datastore and concatenates the content
+func getAllLogs(op trace.Operation, helper *datastore.Helper) (string, error) {
+	res, err := helper.Ls(op, "", []string{vchlog.LogFilePrefix + "*" + vchlog.LogFileSuffix}...)
```
zjs: I think this could be written more simply:

`res, err := helper.Ls(op, "", vchlog.LogFilePrefix + "*" + vchlog.LogFileSuffix)`

PR #7088, opened 2018-01-12
[specific ci=Group6-VIC-Machine] Cleanup trailing empty resource pool if VCH creation fails

### LINKED ISSUE #6975: VCH create leaves dangling vApp or Resource Pool
**Deployment Details**
I tried to create a VCH named `angie-test` using create wizard, and it failed with 500 internal server error due to misconfigured network setting:
`{"message":"Failed to create VCH: Creating the appliance failed with The resource vim.dvs.DistributedVirtualPort is not available in vim.dvs.DistributedVirtualPortgroup DVUplinks-RegionA01-vDS-COMP.. Exiting..."}`

After it failed, I switched to a different network and tried to create again with the same name `angie-test`

**Actual Behavior**
The create failed with error `There is already a VirtualApp or ResourcePool that exists with the same name`:
screen shot 2017-12-18 at 10 55 59 am

This is not because of already created VCH, since it does not show up on the VCH inventory list:
screen shot 2017-12-18 at 10 56 19 am

It might be that on the API side, the failed create leaves a dangling vApp or Resource Pool.

**Expected Behavior**
The create should succeed or at least failed with a different error not concerning existing vApp or resource pool

Cc: @zjs

AngieCris:
Fixes #6975 

Added a cleanup step for trailing empty VCH resource pool and bridge networks created if VCH creation fails in creating appliance. 

An unit test is added in management package. Didn't figure out a way to write an integration test on it. Making the VCH creation fail after validation would require some interventions, and there's a big timing element of that in integration test, so it's easier to do in unit test. I haven't found an existing integration test case of related scenario.

Sidenote:
For the unit test added, I haven't found a good way to check in the vcsim that the VCH pool is successfully removed. For some reason, the existing unit test cases do not check if an VM is actually created, or a volume store is actually deleted for a negative volume store test case, etc. 
Ping people for advice: @hickeng @zjs


lib/install/management/create.go
```diff
@@ -53,10 +53,12 @@ func (d *Dispatcher) CreateVCH(conf *config.VirtualContainerHostConfigSpec, sett
 	}
 
 	if err = d.createBridgeNetwork(conf); err != nil {
+		d.cleanupAfterCreationFailed(conf, false)
 		return err
 	}
 
 	if err = d.createAppliance(conf, settings); err != nil {
+		d.cleanupAfterCreationFailed(conf, true)
```
zjs: Can/should this be registered as a defer which checks for success to ensure it's run in all cases? (It seems like there are disadvantages to this approach. I'm not sure what the right answer is.)

lib/install/management/create.go
```diff
@@ -211,3 +213,69 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
 	}
 	return nil
 }
+
+// cleanupAfterCreationFailed cleans up the dangling resource pool for the failed VCH and any bridge network if there is any.
+// The function will not abort and early terminate upon any error during cleanup process. Error details are logged.
+func (d *Dispatcher) cleanupAfterCreationFailed(conf *config.VirtualContainerHostConfigSpec, cleanupNetwork bool) {
+	defer trace.End(trace.Begin("", d.op))
```
zjs: I think we should give the trace a name?

lib/install/management/create.go
```diff
@@ -211,3 +213,69 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
...
+			d.op.Errorf("Clean up dangling bridge network returns error: %s", err)
+		}
+
+		d.op.Info("Successfully cleaned up dangling bridge network.")
+	}
+}
+
+// cleanupEmptyPool cleans up any dangling empty VCH resource pool when creating this VCH. no-op when VCH pool is nonempty.
+func (d *Dispatcher) cleanupEmptyPool(conf *config.VirtualContainerHostConfigSpec) error {
+	defer trace.End(trace.Begin("", d.op))
```
zjs: I think we should give the trace a name?

lib/install/management/create.go
```diff
@@ -211,3 +213,69 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
...
+	var err error
+
+	// clean up dangling empty resource pool
+	d.op.Info("Clean up dangling VCH resource pool after VCH creation failed...")
+
+	if err = d.cleanupEmptyPool(conf); err != nil {
+		d.op.Errorf("Failed to clean up dangling VCH resource pool: %s", err)
+	}
+
+	d.op.Info("Successfully cleaned up dangling resource pool.")
```
zjs: If `err != nil` above, it seems like we still hit this line. (And if `err != nil`, I don't think it was successful.)

lib/install/management/create.go
```diff
@@ -211,3 +213,69 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
...
 }
+
+// cleanupAfterCreationFailed cleans up the dangling resource pool for the failed VCH and any bridge network if there is any.
+// The function will not abort and early terminate upon any error during cleanup process. Error details are logged.
+func (d *Dispatcher) cleanupAfterCreationFailed(conf *config.VirtualContainerHostConfigSpec, cleanupNetwork bool) {
+	defer trace.End(trace.Begin("", d.op))
+	var err error
+
+	// clean up dangling empty resource pool
+	d.op.Info("Clean up dangling VCH resource pool after VCH creation failed...")
```
zjs: Is this going to be confusing for users? It seems like we may be cleaning up the dangling resource pool between when the failure occurs and when we tell them that the failure occurred.

lib/install/management/create.go
```diff
@@ -211,3 +213,69 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
...
+
+	if err = d.cleanupEmptyPool(conf); err != nil {
+		d.op.Errorf("Failed to clean up dangling VCH resource pool: %s", err)
+	}
+
+	d.op.Info("Successfully cleaned up dangling resource pool.")
+
+	// only clean up bridge network created if told to
+	if cleanupNetwork {
```
zjs: If we go with the defer pattern suggested above we could define a _separate_ function and make a deferred call to it once the network has been created.

lib/install/management/create.go
```diff
@@ -211,3 +213,69 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
...
+		d.op.Info("Cleaning up dangling bridge networks created after VCH creation failed...")
+
+		if err = d.cleanupBridgeNetwork(conf); err != nil {
+			d.op.Errorf("Clean up dangling bridge network returns error: %s", err)
+		}
+
+		d.op.Info("Successfully cleaned up dangling bridge network.")
```
zjs: As above, we'll print this even if `err != nil`.

lib/install/management/create.go
```diff
@@ -211,3 +213,69 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
...
+// cleanupAfterCreationFailed cleans up the dangling resource pool for the failed VCH and any bridge network if there is any.
+// The function will not abort and early terminate upon any error during cleanup process. Error details are logged.
+func (d *Dispatcher) cleanupAfterCreationFailed(conf *config.VirtualContainerHostConfigSpec, cleanupNetwork bool) {
+	defer trace.End(trace.Begin("", d.op))
+	var err error
+
+	// clean up dangling empty resource pool
+	d.op.Info("Clean up dangling VCH resource pool after VCH creation failed...")
+
+	if err = d.cleanupEmptyPool(conf); err != nil {
```
zjs: I'd suggest splitting this into two lines. This syntax is useful to initialize a variable used in a conditional, but can be less clear in cases like this where the effects of the code "hidden in" the conditional are the primary point of the method.

lib/install/management/create.go
```diff
@@ -211,3 +213,69 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
...
+		return err
+	}
+
+	defaultrp, err := d.session.Cluster.ResourcePool(d.op)
+	if err != nil {
+		return err
+	}
+
+	if d.parentResourcepool != nil && d.parentResourcepool.Reference() == defaultrp.Reference() {
+		d.op.Warnf("VCH resource pool is cluster default pool - skipping cleanup")
```
zjs: Should this actually be a warning? It seems like essentially what we're saying is "Nevermind; there's no cleanup to do."

lib/install/management/create.go
```diff
@@ -211,3 +213,69 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
...
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// cleanupBridgeNetwork cleans up any bridge networks created when creating this VCH. no-op for VCenter environment.
+func (d *Dispatcher) cleanupBridgeNetwork(conf *config.VirtualContainerHostConfigSpec) error {
+	defer trace.End(trace.Begin("", d.op))
```
zjs: I think we should give the trace a name?

lib/install/management/create.go
```diff
@@ -211,3 +213,69 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
...
+
+	if err := d.removeNetwork(conf); err != nil {
```
zjs: As above; I think this makes more sense written as two lines.

lib/install/management/create.go
```diff
@@ -211,3 +213,69 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
...
+		d.op.Errorf("Failed to clean up dangling VCH resource pool: %s", err)
+	}
+
+	d.op.Info("Successfully cleaned up dangling resource pool.")
+
+	// only clean up bridge network created if told to
+	if cleanupNetwork {
+		d.op.Info("Cleaning up dangling bridge networks created after VCH creation failed...")
+
+		if err = d.cleanupBridgeNetwork(conf); err != nil {
```
zjs: As above; I think this makes more sense split across two lines.

lib/install/management/create.go
AngieCris: I'm not very sure about that. The cleanup step only applies to two creation steps between pool creation and appliance VM creation. For the steps before creating the pool, there's nothing to cleanup; for the steps after the appliance VM is created, it's more involved since the VM is created, if it fails anywhere during these steps the VM might be broken as well.

lib/install/management/create.go
AngieCris: Thanks, good point. Most of the times in the creation steps, the error message is logged before the creation exists and the message `Clean up dangling VCH resource...` will get logged after the failed error message, so the user knows what's going on. But we can't guarantee that it's always the case.
I'll add logging before we cleanup the dangling resource telling user what's failed.

lib/install/management/create.go
AngieCris: These lines are copied from delete (https://github.com/vmware/vic/blob/master/lib/install/management/delete.go#L114). Not sure what's the reasoning behind printing a warning there, but yes here `d.op.Info` makes more sense.

lib/install/management/create.go
AngieCris: Ah thanks good catch

lib/install/management/create.go
```diff
@@ -211,3 +213,72 @@ func (d *Dispatcher) uploadImages(files map[string]string) error {
...
 	return nil
 }
+
+// cleanupAfterCreationFailed cleans up the dangling resource pool for the failed VCH and any bridge network if there is any.
+// The function will not abort and early terminate upon any error during cleanup process. Error details are logged.
+func (d *Dispatcher) cleanupAfterCreationFailed(conf *config.VirtualContainerHostConfigSpec, cleanupNetwork bool) {
+	defer trace.End(trace.Begin(conf.Name, d.op))
+	var err error
+
+	d.op.Info("VCH creation has failed in previous steps, start cleaning up dangling resources...")
```
hickeng: General message structure comments:
* use of `...` is misleading unless you're going to append `done`, `failed`, `complete` or similar to the same line once the operation is completed given traditional console usage.
* it's ok to be imperative and concise, e.g. `Cleaning up after failure`
* multi-line output in this manner should only be used if it adds significant value as it's prone to tearing.

I don't think any of this needs to be at Info level, debug is fine - it will not take significant time under normal conditions and it's not something that requires user intervention. At most one info message noting cleanup is occurring but even that I consider optional.

lib/install/management/create_test.go
```diff
@@ -83,7 +83,18 @@ func TestMain(t *testing.T) {
 		errConf.VolumeLocations = make(map[string]*url.URL)
 		errConf.VolumeLocations["volume-store"], _ = url.Parse("ds://store_not_exist/volumes/test")
 		testCreateVolumeStores(op, validator.Session, errConf, true, t)
+
+		// FIXME: (pull vic/7088) have to make another VCH config from validator for negative test case and cleanup test
+		// If we re-use the previous validator like we did in the earlier test (*errConf = *conf), it's not a deep copy of conf
+		// This conf will get modified by appliance creation and cleanup test, and we can't test create appliance in positive case
+		// The other way around, if we test positive case first, the VCH data and session data are modified, so we are not able to test the negative case
+		conf2, err := validator.Validate(op, input)
+		conf2.ImageStores[0].Host = "http://non-exist"
```
hickeng: I think this will cause a validator failure meaning we'll never try creating the resource pool or appliance.
Please confirm that you see the expected output when running the test.

AngieCris: This `conf2` produced from validation is identical to the `conf` used in existing tests. For some reason, in producing `conf` in existing tests, the validation failed from the output from running the unit test, but a config is successfully created, and I do see the expected log messages from the unit test output when running it. Also checked with manual testing.

PR #7089, opened 2018-01-12
Collect vmx and vmsd files in log bundle

jzt:
This change will enable the vicadmin log bundler to also collect the `vmx` and `vmsd` files for VMs associated with the VCH. Useful for debugging.

### DIFF
--- a/cmd/vicadmin/vicadm.go
+++ b/cmd/vicadmin/vicadm.go
@@ -472,6 +472,26 @@ func findDatastoreLogs(c *session.Session) (map[string]entryReader, error) {
 
 			log.Debugf("Added log file for collection: %s", logfile.URL.String())
 		}
+
+		f := logfile.VMName
+		if !c.IsVC() {
+			f = logfile.URL.Path
+		}
+
+		// get the vmx and vmsd files
+		for _, ext := range []string{"vmx", "vmsd"} {
+			rpath := fmt.Sprintf("%s/%s.%s", logfile.URL.Path, f, ext)
+			wpath := fmt.Sprintf("%s/%s.%s", logfile.VMName, f, ext)
+			log.Infof("Processed File read Path : %s", rpath)
+			log.Infof("Processed File write Path : %s", wpath)
+			readers[wpath] = datastoreReader{
+				ds:   ds,
+				path: rpath,
+				ctx:  hCtx,
+			}
+
+			log.Debugf("Added log file for collection: %s", logfile.URL.String())
+		}
 	}
 
 	return readers, nil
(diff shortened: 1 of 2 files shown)


cmd/vicadmin/vicadm.go
```diff
@@ -472,6 +472,26 @@ func findDatastoreLogs(c *session.Session) (map[string]entryReader, error) {
...
 		}
+
+		f := logfile.VMName
+		if !c.IsVC() {
+			f = logfile.URL.Path
+		}
+
+		// get the vmx and vmsd files
+		for _, ext := range []string{"vmx", "vmsd"} {
+			rpath := fmt.Sprintf("%s/%s.%s", logfile.URL.Path, f, ext)
```
vburenin: for file path purposes use filepath.Join

cmd/vicadmin/vicadm.go
```diff
@@ -472,6 +472,26 @@ func findDatastoreLogs(c *session.Session) (map[string]entryReader, error) {
...
+			wpath := fmt.Sprintf("%s/%s.%s", logfile.VMName, f, ext)
+			log.Infof("Processed File read Path : %s", rpath)
+			log.Infof("Processed File write Path : %s", wpath)
+			readers[wpath] = datastoreReader{
+				ds:   ds,
+				path: rpath,
+				ctx:  hCtx,
+			}
+
+			log.Debugf("Added log file for collection: %s", logfile.URL.String())
```
vburenin: No need to call String. string formatter checks for Stringer interface on every %s occurrence, if it matches that interface it calls String() method.

cmd/vicadmin/vicadm.go
```diff
@@ -472,6 +472,26 @@ func findDatastoreLogs(c *session.Session) (map[string]entryReader, error) {
 
 			log.Debugf("Added log file for collection: %s", logfile.URL.String())
 		}
+
+		f := logfile.VMName
```
vburenin: could you use vmName as a variable name?

cmd/vicadmin/vicadm.go
```diff
@@ -472,6 +472,26 @@ func findDatastoreLogs(c *session.Session) (map[string]entryReader, error) {
...
+		if !c.IsVC() {
+			f = logfile.URL.Path
+		}
+
+		// get the vmx and vmsd files
+		for _, ext := range []string{"vmx", "vmsd"} {
```
vburenin: Just sharing my thoughts, this comment might is arguable and doesn't really worth any effort in this particular case to address it: From testability standpoint I would use a separate function and would call this function twice with different file extensions. From simplicity standpoint, I would probably avoid using loop and created two reader objects, but that would be questionable if needed to be extended later.

Issue #6642, opened 2017-10-27
Failed blueprint provision leaves template network in an un-removable state


michaelschlies:
**VIC version:**

1.2.1

**Deployment details:**

What was the vic-machine create command used to deploy the VCH?

```
./vic-machine-linux create --target [vsphere ip] --user [username] \
 --thumbprint [thumbprint] --client-network Inside --no-tls --public-network ContainerRoutable \
 --bridge-network Containers --image-store vSphereDS --name vch-app-host --volume-store \
vSphereDS:default --container-network ContainerRoutable:public --container-network-firewall \
ContainerRoutable:open --client-network-ip 192.168.1.231/23 --dns-server 192.168.0.1
```

**Steps to reproduce:**
I created a multi-container application template for Ghost including _/ghost and _/mysql, a connection for ghost -> public, ghost -> "db-network" (network added in VIC designer), and mysql -> "db-network". 

```yaml  
---
name: "Ghost"
components:
  ghost:
    type: "App.Container"
    data:
      name: "ghost"
      image: "registry.hub.docker.com/library/ghost"
      _cluster: 1
      volumes:
      - "ghost_data:/var/lib/ghost"
      hostname: "ghost"
      networks:
      - name: "db-network"
      - name: "public"
      publish_all: false
      restart_policy: "no"
  public:
    type: "App.Network"
    data:
      name: "public"
      external: true
  db-network:
    type: "App.Network"
    data:
      name: "db-network"
      external: false
  mysql:
    type: "App.Container"
    data:
      name: "mysql"
      image: "registry.hub.docker.com/library/mysql"
      _cluster: 1
      volumes:
      - "mysql_data:/var/lib/mysql"
      hostname: "mysql"
      networks:
      - name: "db-network"
      publish_all: true
      health_config:
        protocol: "TCP"
        port: 3306
        autoredeploy: true
        timeout_millis: 2000
        ignore_on_provision: true
      restart_policy: "no"
  ghost_data:
    type: "App.Volume"
    data:
      id: "69ca19bf-f4b0-4117-8acd-6f8db378e3d8"
      name: "ghost_data"
      tenantLinks:
      - "/projects/default-project"
      driver: "local"
  mysql_data:
    type: "App.Volume"
    data:
      id: "dfbe7c16-2e7d-438e-80d9-16ab46edae42"
      name: "mysql_data"
      tenantLinks:
      - "/projects/default-project"
      driver: "local"

```
**Actual behavior:**
[...]

PR #7090, opened 2018-01-15
[specific ci=1-17-Docker-Network-Connect] Add concurrency support for docker network connect

anchal-agrawal:
This commit allows for concurrent 'docker network connect' operations on
the same container. Earlier, this would cause only one such operation to
succeed, while the others would fail with a conflict error during the
portlayer's Commit call.

This change wraps a retry for the portlayer operations involved. If the
portlayer returns a CommitConflict error, the operations are retried for
a maximum of two minutes.

Towards #6642

### DIFF
--- a/lib/apiservers/engine/backends/network.go
+++ b/lib/apiservers/engine/backends/network.go
@@ -38,6 +38,7 @@ import (
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/containers"
 	"github.com/vmware/vic/lib/apiservers/portlayer/client/scopes"
 	"github.com/vmware/vic/lib/apiservers/portlayer/models"
+	"github.com/vmware/vic/pkg/retry"
 )
 
 type Network struct {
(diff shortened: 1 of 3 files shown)


lib/apiservers/engine/backends/network.go
```diff
@@ -182,12 +183,19 @@ func (n *Network) CreateNetwork(nc types.NetworkCreateRequest) (*types.NetworkCr
...
 }
 
-func (n *Network) ConnectContainerToNetwork(containerName, networkName string, endpointConfig *apinet.EndpointSettings) error {
-	vc := cache.ContainerCache().GetContainer(containerName)
-	if vc != nil {
-		containerName = vc.ContainerID
+// isCommitConflictError returns true if err is a conflict error from the portlayer's
+// handle commit operation, and false otherwise.
+func isCommitConflictError(err error) bool {
+	if err == nil {
```
vburenin: https://play.golang.org/p/mTaIL8391mJ

lib/apiservers/engine/backends/network.go
```diff
@@ -280,8 +288,26 @@ func (n *Network) ConnectContainerToNetwork(containerName, networkName string, e
...
+}
+
+// ConnectContainerToNetwork connects a container to a container network. It wraps the portlayer operations
+// in a retry for when there's a conflict error received, such as one during a similar concurrent operation.
+func (n *Network) ConnectContainerToNetwork(containerName, networkName string, endpointConfig *apinet.EndpointSettings) error {
+	vc := cache.ContainerCache().GetContainer(containerName)
+	if vc != nil {
+		containerName = vc.ContainerID
+	}
+
```
gigawhitlocks: What happens if `vc` is `nil`?

lib/apiservers/engine/backends/network.go
anchal-agrawal: Thanks, fixed!

Issue #7084, opened 2018-01-12
Issues after upgrade to engine v1.3.0-15556-473375a


malikkal:
hickeng: The following is the original issue description which is a facet of #7117 and #7091. I have prompted this to an epic for the two distinct upgrade issues that have been identified:
1. containers attached to Open container networks cause various issues after VCH reboot
2. intermittent failure to delete images by ID that were pulled pre-upgrade

Original details:
1. Docker start/stop does not work and results in
`Error response from daemon: all ports are allocated.` 

However, start/stop does work for existing containers that were in 'Created' state.

2. Docker exec does not work and results in
`Error response from daemon: Server error from portlayer: unknown task ID: 168f7b4dd0129d38fec73a032ced341c848eaec9213f8c728f2d3c4373df617c`

3. Cannot delete existing containers, as they would complain as running and stop wouldn’t work either. Force removal would through an error as:
`Error response from daemon: all ports are allocated`.

Interestingly, old cVMs that were up and running as part of docker-compose could be removed using docker-compose down -v. Also, a docker kill stopped an old container, which I could start using docker start later. Subsequent docker stop/start continued **not** to work though..

anchal-agrawal:
Hi @malikkal! For the `Error response from daemon: all ports are allocated.` issue:

* What do this container's port-mappings look like?
* How was this container created/run?

Thanks!

malikkal:
Regardless of the use of -p during run, the error would always appear, when it's on external network. You could notice the difference in output for a ps -a when a container is attached to external network between current and previous versions.

```
PORTS
10.x.y.z:0->0/*
```
This would lead to verbose inspect output #7091. Earlier versions used to be empty.

```
PORTS
10.x.y.z:8080->80/tcp
```
Using NAT, where inspect is okay in new version. Obviously port bindings are different; see below.

```
"PortBindings": {
                "80/tcp": [
                    {
                        "HostIp": "",
                        "HostPort": "8080"
                    }
                ]
            }
```
Hope this helps.

hickeng:
I've promoted this to an epic and added related issues as they all derive from the same root cause (assumption here for upgrade, but a reasonable one until verified).

Comment duplicated from https://github.com/vmware/vic/issues/7117#issuecomment-358797679
[...]

PR #6641, opened 2017-10-27
Correct ps output and restart/nat behaviour

hickeng:
PR #6557 addressing IP output in docker ps listings was incorrect
if connecting to a container network with network connect, after having
created container with a bridge network.

Also updates the robot utility function `Get Container IP` to use container
inspect which is populated with DHCP addresses, where the network inspect
entry for the same endpoint is not. Prior to this change that function was
returning an empty string and the tests using it were written such that they
pass if `Get Container IP` fails in that manner.

Example output:
```
$ docker run -d -p 8080 --net=bridge --name=bridge_only tomcat:alpine
8ad3acab10fea51921ac7a01e0dfa8021b0fe2610d6aff2b8a4351aead32bf2d
$ docker run -d -p 8080 --net=external --name=published_only tomcat:alpine
60ce369b115d16220b9428741c20e3229be42c83fa28407b051d209540a0e17f
$ docker run -d -p 80:8080 --net=external --name=published_redirect tomcat:alpine
b3c9be5bc6560924b384da125dd334618a96d228d45eaf09f193be8b498db839
$ docker run -d -p 8080 --net=open --name=open_only tomcat:alpine
4ec055fc5389f435d8640a9a352bbf23116f21f9c46cd33f31e36e7abb43297a
$ docker run -d -p 80:8080 --net=open --name=open_redirect tomcat:alpine
187f1a94af8b7610a319a9a7370f4214c9bfa902415933173cbc11d7ee40ac49
$ docker create -p 80:8080 --net=bridge --name=bridge_first tomcat:alpine
2e63a2e59fe8f21cac26ff45260e5d452a2bf6afd2b1ea78d35d7c4c4b7c9531
$ docker network connect external bridge_first
$ docker start bridge_first
bridge_first
$ docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS                                                NAMES
2e63a2e59fe8        tomcat:alpine       "catalina.sh run"   5 minutes ago       Up 20 seconds       192.168.78.127:80->8080/tcp                          bridge_first
187f1a94af8b        tomcat:alpine       "catalina.sh run"   5 minutes ago       Up About a minute   192.168.78.200:0->0/*, 192.168.78.200:80->8080/tcp   open_redirect
4ec055fc5389        tomcat:alpine       "catalina.sh run"   6 minutes ago       Up 2 minutes        192.168.78.203:0->0/*                                open_only
[...]

PR #7105, opened 2018-01-17
Add integration test on open container upon VCH reboot [specific ci=Group10-VCH-Restart]

AngieCris:
Currently we have a bug that if we configure container network with open firewall policy, after VCH reboot, all ports are being forwarded. docker persona takes a long time to start, `docker ps` and `docker inspect` gives huge list of all ports. 
A test is needed on VCH reboot with open container network. 

Related to: https://github.com/vmware/vic/issues/7084

Cc: @hickeng

### DIFF
--- a/tests/test-cases/Group10-VCH-Restart/10-01-VCH-Restart.robot
+++ b/tests/test-cases/Group10-VCH-Restart/10-01-VCH-Restart.robot
@@ -74,14 +74,6 @@ Created Network And Images Persists As Well As Containers Are Discovered With Co
     ${rc}=  Run And Return Rc  docker %{VCH-PARAMS} exec bar-c2 ping -c3 bar-c1
     Should Be Equal As Integers  ${rc}  0
 
-    # ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} create -it -p 10000:80 -p 10001:80 --name webserver ${nginx}
-    # Should Be Equal As Integers  ${rc}  0
-    # Should Not Contain  ${output}  Error
-    # ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} start webserver
-    # Should Be Equal As Integers  ${rc}  0
-    # Should Not Contain  ${output}  Error
-    # Wait Until Keyword Succeeds  20x  5 seconds  Hit Nginx Endpoint  %{VCH-IP}  10000
-    # Wait Until Keyword Succeeds  20x  5 seconds  Hit Nginx Endpoint  %{VCH-IP}  10001
     Launch Container With Port Forwarding  ${name}=webserver  ${port1}=10000  ${port2}=10001
 
     Reboot VM  %{VCH-NAME}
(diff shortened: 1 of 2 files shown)


tests/test-cases/Group10-VCH-Restart/10-01-VCH-Restart.robot
```diff
@@ -145,7 +149,58 @@ Created Network And Images Persists As Well As Containers Are Discovered With Co
...
+    ${open-running}=  Launch Container  vch-restart-open-running  open-net
+    ${open-exited}=  Launch Container  vch-restart-open-exited  open-net  ls
+
+    # Create nginx on the open network and check port forwarding
+    Launch Container With Port Forwarding  ${name}=webserver  ${port1}=10000  ${port2}=10001  ${network}=open-net
+
+    # Reboot VCH
+    Reboot VM  %{VCH-NAME}
+    Log To Console  Getting VCH IP ...
+    ${new-vch-ip}=  Get VM IP  %{VCH-NAME}
```
hickeng: This is failure prone in a DHCP environment as the IP address we get is likely to be from the prior boot. Keyword `Wait For VCH Initialization` specifically to avoid this class of error (added for Nimbus HA testing).
Assuming the standard variable names are used and are therefore updated by this keyword, we would have :
```
Reboot VM  %{VCH-NAME}
Wait For VCH Initialization  20x  5 seconds

# Check if the open container is persisted
```

tests/test-cases/Group10-VCH-Restart/10-01-VCH-Restart.robot
```diff
@@ -15,8 +15,6 @@
 *** Settings ***
 Documentation  Test 10-01 - VCH Restart
 Resource  ../../resources/Util.robot
-Suite Setup  Install VIC Appliance To Test Server
```
hickeng: I would prefer to keep these here and override them for the specific tests using the following:
```
[Setup]  Install VIC Appliance ...
...
[Teardown]  Cleanup VIC Appliance On Test Server
```

This would mean that `[Teardown]` doesn't need to be used as it's common for all tests, and it's only the open network test that needs to override `[Setup]`

tests/test-cases/Group10-VCH-Restart/10-01-VCH-Restart.robot
```diff
@@ -145,7 +149,58 @@ Created Network And Images Persists As Well As Containers Are Discovered With Co
     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull ${alpine}
     Should Be Equal As Integers  ${rc}  0
 
+    Cleanup VIC Appliance On Test Server
+
+Container on Open Network And Port Forwarding Persist After Reboot
+    Set Test Environment Variables
```
hickeng: Use `[Setup]  NONE` to suppress the normal appliance install instead of changing all the tests. Longer term we should provide Keywords for deploying VCHs of various standard configurations for easier implementation of these more involved setups.

AngieCris:
Close this as pulled in as part of #6641

PR #7125, opened 2018-01-18
Add new tag to skip unit tests

### LINKED ISSUE #7039: Add a square bracket tag for skipping unit tests
When trying to debug failures, sometimes we want to skip unit tests in order to improve speed of our tests so we need to add a [skip unit] tag to our build time directives.

mhagen-vmware:
[skip unit]
fixes #7039


tests/unit-test-check.sh
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+prNumber=$(drone build info --format "{{ .Ref }}" vmware/vic $DRONE_BUILD_NUMBER | cut -f 3 -d'/')
+prBody=$(curl https://api.github.com/repos/vmware/vic/pulls/$prNumber | jq -r ".body")
+
+if ! (echo $prBody | grep -q "\[skip unit\]"); then
```
hickeng: If you're after a slight UI refinement this could even leverage the markdown and check for:
```
- [x] [skip unit]
```
which will render as the following (the `x` will be missing if the box is unchecked)
- [x] [skip unit]

Issue #7115, opened 2018-01-17
Rebooting VCH overwrites log files


mhagen-vmware:
10-01 test failed in this build: https://ci.vcna.io/vmware/vic/15691

But in part of the test, the VCH is rebooted.  After the reboot the logs are collected but neither port-layer nor personality logs have the failure anymore.
I expected that maybe vic-admin didn't collect logs.1 or older logs, but enabling SSH on a VCH that I rebooted, there are no logs at all just new overwritten logs:
```
root@ [ /var/log/vic ]# ls
docker-personality.log  init.log  port-layer.log  vicadmin.log
```

We need to maintain those logs for our log bundles otherwise we can do much for debugging.  Esp at customer site, if they reboot then it is all gone!

matthewavery:
I second this, right now the VCH is completely stateless. But this makes finding some errors in the logs impossible to diagnose. It is hard to recommend a VCH restart to mitigate a problem when we will lose the logs that let to a scenario to begin with.

hickeng:
It doesn't overwrite them, it discards them because they are in memory. The appliance is deliberately non-persistent from a configuration/binaries perspective but was always intended to have a disk for log persistence and caching. #2079 was to address this, but was hurried for initial 0.8 release and only applies to debug configurations.

We had an issue for adding a disk to persist logs but I cannot find it right now - it may have been closed or may be part of the log rotation work. Regardless this is certainly something that needs adding, along with using a scratch disk to cache image layers while downloading instead of tmpfs.

hickeng:
https://ci.vcna.io/vmware/vic/15691 10-01-VCH-Restart is also impacted by this.
There is a failure to ping between two containers on a bridge network:
https://github.com/vmware/vic/blob/dd3325314d0372b02a7fbc0dd0f85760976140c1/tests/test-cases/Group10-VCH-Restart/10-01-VCH-Restart.robot#L63
and the only data we have is from the bar-c2 tether.debug:
```
Jan 17 2018 03:56:49.435Z DEBUG [0xc420375620] write "ping: bad address 'bar-c1'\n" to 1 writers (err: <nil>)
```
unfortunately we do not have portlayer logs for that time period which means we cannot determine if the name resolution occurred an this is another instance of [broken busybox ping](https://github.com/vmware/vic/issues/2669#issuecomment-253953729) or whether it was an actual failure to resolve the container names correctly.

anchal-agrawal:
[...]

PR #7133, opened 2018-01-19
Collect logs before reboot in VCH restart test

anchal-agrawal:
This commit adds log collection before every reboot in the
10-01-VCH-Restart test so we can retain logs for every test operation.

See #7115


sgairo:
@anchal-agrawal build fail was build does not match, want to restart?

PR #7136, opened 2018-01-19
Workaround for issue: 7124, ops-user Administrator group check

lcastellano:
Fixes # 7124, 7054

This PR temporary disables the check that is meant to guarantee that the ops-user does not belong to the Administrators group. The test is only skipped if an incorrect NotFound error is returned from the PSC. In the future we will validate the ops-user privileges independently from its Role or Group membership.

### DIFF
--- a/pkg/vsphere/rbac/rbac.go
+++ b/pkg/vsphere/rbac/rbac.go
@@ -17,13 +17,17 @@ package rbac
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/govmomi/vim25"
 	"github.com/vmware/govmomi/vim25/methods"
+	"github.com/vmware/govmomi/vim25/soap"
 	"github.com/vmware/govmomi/vim25/types"
+
 	"github.com/vmware/vic/pkg/errors"
+	"github.com/vmware/vic/pkg/trace"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


lcastellano:
It seems to me to, but unfortunately that's what vCenter does. Before I added the check I mistakenly used admin for ops-user and ended up spending 30 minutes removing the permissions from all the different resources in the tree, to make my admin account working again. That was actually the reason why I added those checks. The correct approach is to perform a smarter validation before adding permissions to resources. But that is big change.

mdubya66:
do we need to add tests for this?

lcastellano:
For this to be exercised in an integration test we need Patrick's environment. An Active Directory local server must be added to the Identity Sources in PSC/VC. As for unit tests, I have used the Simulator, but the Simulator behaves correctly in this case. The actual PSC is the problem.

pkg/vsphere/rbac/rbac.go
```diff
@@ -376,3 +392,19 @@ func (am *AuthzManager) getRoleName(resource *Resource) string {
...
+
+func isNotFoundError(ctx context.Context, err error) bool {
+	op := trace.FromContext(ctx, "isNotFoundError")
+
+	if soap.IsSoapFault(err) {
+		vimFault := soap.ToSoapFault(err).VimFault()
+		op.Debugf("Error type: %s", reflect.TypeOf(vimFault))
+
+		_, ok := soap.ToSoapFault(err).VimFault().(types.NotFound)
+		if ok == true {
```
hickeng: just return `ok` ;)

Issue #7118, opened 2018-01-18
docker rmi behaviour changed


malikkal:
VIC Engine v1.3.0-15556-473375a

User statement: As a customer of VIC, I would like to have upgrade maintain backward compatibility (if possible) so that we could handle consistent development.

Details: In the past docker rmi <image id> used to work; but, it now returns an error as below:

```
docker rmi 6f146fe8049d
Error response from daemon: No such image: 6f146fe8049d:latest
```
However, removing the image by referencing as registry/repository:tag succeeds.

malikkal:
Tried on another VCH that was upgraded too and it works well. Looks like its inconsistent behavior. I am not sure how to reproduce. :-(

anchal-agrawal:
Looks like images that have been pulled pre-upgrade (1.2.1) cannot be deleted by image ID after upgrade (1.3.0). Images pulled after upgrading can be removed by name and ID. It would be very helpful If you could confirm this on your setup, @malikkal.

The attached log bundle is from a 1.2.1 VCH that was upgraded to 1.3.0. I pulled `busybox` and `redis:alpine` before upgrading, and `debian` and `nginx` after upgrading. I was able to `rmi` using the image IDs of `debian` and `nginx` but not `busybox` and `redis:alpine`. After upgrading, I deleted `busybox` by name and re-pulled it, and was then able to delete it by image ID. The log bundle has all these operations.

[7118-logs.zip](https://github.com/vmware/vic/files/1644779/7118-logs.zip)

hickeng:
@anchal-agrawal @malikkal I've confirmed that this error occurs with a restart, not just an upgrade. Deleting using the full ID works, just not the shortID.
```
$ docker rmi c3528103517d
Error response from daemon: No such image: c3528103517d:latest
$ docker rmi c3528103517df45826e18b195cc7688b5f101758225f02e986bd2aa43399ffd1
Deleted: a3ed95caeb02ffe68cdd9fd84406680ae93d633cb16422d00e8a7c22955b46d4
<snip>
Untagged: tomcat:alpine
Untagged: tomcat@sha256:55def4ece5aaf838bf626dc665274be80c749f8f31e920dc942596ea1efa97bb
```

malikkal:
> ERROR Handler for DELETE /v1.25/images/21a2450d623e returned error: No such image: 21a2450d623e:latest

[...]

PR #7139, opened 2018-01-20
[specific ci=Group1-Docker-Commands] Fix persona image cache init upon VCH restart

### LINKED ISSUE #6076: Cannot Delete Images Using Image ID After VCH Restart
As a user I should be able to do `docker images` and then use an Image ID from that output to delete images. Likewise, doing `docker rmi -f $(docker images -aq)` should successfully remove all images present on the VCH.

Current behavior:
1) Install VCH
2) Pull some images (examples below)
3) Restart VCH
4) Follow below:
```bash
~/go/src/github.com/vmware/vic fix-local-build-issues*
❯ docker images
ubuntu              latest              ccc7a11d65b1        11 days ago         120MB
tomcat              latest              72d2be374029        12 days ago         292MB
nginx               latest              b8efb18f159b        3 weeks ago         107MB
golang              latest              6ce094895555        3 weeks ago         699MB
python              latest              968120d8cbe8        4 weeks ago         684MB
ruby                latest              3630c02d3d1b        4 weeks ago         679MB
fedora              latest              49236bc2f0da        4 weeks ago         231MB
alpine              latest              7328f6f8b418        7 weeks ago         3.96MB

~/go/src/github.com/vmware/vic fix-local-build-issues*
❯ docker rmi ccc7a11d65b1
Error response from daemon: No such image: ccc7a11d65b1:latest
```
Desired behavior:
```
[vagrant@devbox:/opt/go/src/github.com/vmware/vic(reload)] docker rmi ccc7a11d65b1
Untagged: ubuntu:latest
Untagged: ubuntu@sha256:34471448724419596ca4e890496d375801de21b0e67b81a77fd6155ce001edad
Deleted: sha256:ccc7a11d65b1b5874b65adb4b2387034582d08d65ac1817ebc5fb9be1baa5f88
Deleted: sha256:cb5450c7bb149c39829e9ae4a83540c701196754746e547d9439d9cc59afe798
Deleted: sha256:364dc483ed8e64e16064dc1ecf3c4a8de82fe7f8ed757978f8b0f9df125d67b3
Deleted: sha256:4f10a8fd56139304ad81be75a6ac056b526236496f8c06b494566010942d8d32
Deleted: sha256:508ceb742ac26b43bdda819674a5f1d33f7b64c1708e123a33e066cb147e2841
Deleted: sha256:8aa4fcad5eeb286fe9696898d988dc85503c6392d1a2bd9023911fb0d6d27081
[vagrant@devbox:/opt/go/src/github.com/vmware/vic(reload)]
```

Workaround available: delete the image with the human-readable name, e.g. `ubuntu:latest`

Acceptance Criteria:
Fix above behavior *and add an integration test* as it seems we currently only test deleting `human-readable-name:tag`and this is likely a regression.

anchal-agrawal:
This commit fixes the initialization logic of the docker personality's
image cache. Before this change, images pulled before a VCH restart
would fail to be removed later by their short ID (their long ID would
work). This was because when the personality server was restarted, the
ID index in the image cache was populated with the incorrect key format.

When an image is pulled, the ID index in the image cache is keyed by the
full ID of the image, while the name cache is keyed by the full ID
prefixed with "sha256:". In the cache init logic, the ID index was
previously populated with the name cache's keys, and this would cause
image lookups (docker rmi, inspect) to fail when the image's short ID
was used.

To fix this issue, this change trims the "sha256:" prefix from the name
cache's keys before using them as keys for the ID index.

Fixes: #6076 #7118

### DIFF
--- a/lib/apiservers/engine/backends/cache/image_cache.go
+++ b/lib/apiservers/engine/backends/cache/image_cache.go
@@ -1,4 +1,4 @@
-// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
+// Copyright 2016-2018 VMware, Inc. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -105,6 +105,13 @@ func InitializeImageCache(client *client.PortLayer) error {
 
 		// populate the trie with IDs
 		for k := range i.CacheByID {
+			// Separate out the hash prefix from the CacheByID key before indexing iDIndex
+			// as it is keyed by the full image ID without the hash prefix.
+			fields := strings.SplitN(k, ":", 2)
+			if len(fields) == 2 {
+				k = fields[1]
+			}
+
 			imageCache.iDIndex.Add(k)
 		}
 
@@ -201,14 +208,14 @@ func (ic *ICache) getImageByNamed(named reference.Named) *metadata.ImageConfig {
 	return copyImageConfig(ic.cacheByID[prefixImageID(id)])
 }
 
-// Add the "sha256:" prefix to the image ID if missing.
-// Don't assume the image id in image has "sha256:<id> as format.  We store it in
-// this format to make it easier to lookup by digest
+// Add the default "sha256:" prefix to the image ID if it doesn't include a hash
+// prefix. Don't assume the image ID has "<hash>:<id> as format (e.g. "sha256:<id>").
+// We store it in this format to make it easier to lookup by digest.
 func prefixImageID(imageID string) string {
-	if strings.HasPrefix(imageID, "sha256:") {
+	if strings.Contains(imageID, ":") {
 		return imageID
 	}
-	return "sha256:" + imageID
+	return digest.Canonical.String() + ":" + imageID
 }
 
 // Add adds an image to the image cache
--- a/tests/test-cases/Group1-Docker-Commands/1-12-Docker-RMI.robot
+++ b/tests/test-cases/Group1-Docker-Commands/1-12-Docker-RMI.robot
@@ -1,4 +1,4 @@
-# Copyright 2016-2017 VMware, Inc. All Rights Reserved.
+# Copyright 2016-2018 VMware, Inc. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -105,6 +106,9 @@ func InitializeImageCache(client *client.PortLayer) error {
 
 		// populate the trie with IDs
 		for k := range i.CacheByID {
+			// Trim the "sha256:" prefix from the CacheByID key before indexing iDIndex
+			// as it is keyed by the full image ID without the "sha256:" prefix.
+			k = strings.TrimPrefix(k, sha256Prefix)
```
hickeng: We are currently exclusively using `sha256` but the only reason for there to be a prefix is to allow for other hashes to be used.
I'd suggest `strings.SplitN(k, ":", 2)` or similar and use the last element.

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -205,10 +209,10 @@ func (ic *ICache) getImageByNamed(named reference.Named) *metadata.ImageConfig {
 // Don't assume the image id in image has "sha256:<id> as format.  We store it in
 // this format to make it easier to lookup by digest
 func prefixImageID(imageID string) string {
-	if strings.HasPrefix(imageID, "sha256:") {
+	if strings.HasPrefix(imageID, sha256Prefix) {
```
hickeng: If you make the change I suggest to avoid the `sha256` assumption, then this would change to `strings.Contains(k, ":")` instead of the prefix check.
In this case sha256 because the _default_ hash.

tests/test-cases/Group1-Docker-Commands/1-12-Docker-RMI.robot
```diff
@@ -26,15 +26,10 @@ Basic docker pull, restart, and remove image
     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} pull ${alpine}
     Should Be Equal As Integers  ${rc}  0
 
+    # Gather logs before rebooting
+    Run Keyword And Continue On Failure  Gather Logs From Test Server  -before-reboot-1
```
hickeng: Perhaps we should move this into a keyword `Reboot VCH` that combines the log collection with the `Reboot VM` call - if we always collect logs before rebooting then it will avoid a lot of unnecessary duplication.

anchal-agrawal: I'm inclined to defer this change as there are several other tests that use this keyword, and we should make minimal unrelated changes for release purposes. I'll open an issue for this - we can also capture the need to develop a better log bundle naming mechanism that doesn't rely on manual naming for uniqueness across log bundle names.

anchal-agrawal: Done: https://github.com/vmware/vic/issues/7143

lib/apiservers/engine/backends/cache/image_cache.go
anchal-agrawal: Fixed!

lib/apiservers/engine/backends/cache/image_cache.go
anchal-agrawal: Fixed!

lib/apiservers/engine/backends/cache/image_cache.go
```diff
@@ -49,7 +49,8 @@ type ICache struct {
 }
 
 const (
-	imageCacheKey = "images"
+	imageCacheKey     = "images"
+	defaultHashPrefix = "sha256:"
```
jzt: Minor: you could use docker's [`digest.Canonical`](https://github.com/vmware/vic/blob/a5383fd8fdbfbe118b780e8e8952622a787ffb32/vendor/github.com/docker/distribution/digest/digester.go#L17) here.

anchal-agrawal: Good call, will fix. I'll also update the reference in `backends/image.go`.

PR #7155, opened 2018-01-23
[specific-ci=1-04-Docker-Create] Add logical handling for superseding image volumes with named volumes

### LINKED ISSUE #7138: Named volumes incorrectly detected as anonymous, getting deleted when container is removed

**VIC version:**

1.3.0 / Commit ID# 473375a

**Deployment details:**

vic-machine-windows --target=https://X.X.X.X --thumbprint=<thumbprint> --name=dev-vch --compute-resource=/AWN-DC-H5/host/AWN Production/Resources/Servers --ops-user=administrator@vs
phere.local --image-store=ds://Servers-AFA --volume-store=ds://Servers-AFA/docker-volumes:default --dns-server=X.X.X.X --bridge-network=vic-bridge --bridge-network-range=172.32.0.0/12 --public-network=Manage
ment --public-network-gateway=X.X.X.X --public-network-ip=X.X.X.X/24 --container-network=WWW:vic-www --container-network=Management:vic-mgmt --container-network-gateway=Management:X.X.X.X/24 --container-
network-gateway=WWW:X.X.X.X/24 --container-network-ip-range=WWW:X.X.X.X/25 --container-network-ip-range=Management:X.X.X.X/29 --container-network-dns=Management:X.X.X.X --container-network-dns=
WWW:X.X.X.X --memory-shares=500 --cpu-shares=500

**Steps to reproduce:**

_**The image being used for testing is sebp/elk:611**_

These are the steps I have been performing when I experience the bug:

1. Follow procedure for backing up/restoring VMDK's for docker volumes (vic 1.3 documentation)
2. restart VCH to detect new volumes 
3. Run container using the following: docker -H vch.awninc.local:2376 --tls run --name elk-node2 -p 9200 -p 9300 -it --rm -m 4G -e LOGSTASH_START=0 -e KIBANA_START=0 -e MAX_MAP_COUNT=262144 --memory-swap -1 --network vic-mgmt -v elk-node2-config:/etc/elasticsearch/ -v elk-node2-data:/var/lib/elasticsearch vic.awninc.local/utilities/elk:611 bash 
4. Stop container by exiting bash, using 'exit'

**Actual behavior:**
Container is stopped and removed from VCH; 2nd named volume gets deleted (in this case, elk-node2-data)
**Expected behavior:**
Container stops and is removed, all named volumes remain intact.
**Logs:**

[logs (1).zip](https://github.com/vmware/vic/files/1648043/logs.1.zip)

Named volumes in question that are being attached to the containers were moved from another datastore.

matthewavery:
Fixes #7138 

This fix will now remove the anonymous volume metadata associated with a container config if a named volume supersedes the anonymous volume. Since Image Volumes are by default going to show up as anonymous volumes, there is a need to "swap out" the metadata used to identify a volume as anonymous or named. `docker rm -v` and `docker run --rm` both delete the anonymous volumes associated with a container. We now correctly associate the metadata in the container correctly for the deployment that was specified.

### DIFF
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -565,11 +565,10 @@ func (c *Container) ExecExists(eid string) (bool, error) {
 	return true, nil
 }
 
-// docker's container.stateBackend
-
 // ContainerCreate creates a container.
 func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (containertypes.ContainerCreateCreatedBody, error) {
 	defer trace.End(trace.Begin(""))
+	// TODO: Add Op logging to this operation. This is the start of container create operation and op logging here is very valuable.
 
 	var err error
 
--- a/lib/apiservers/engine/backends/container.go
+++ b/lib/apiservers/engine/backends/container.go
@@ -568,7 +568,6 @@ func (c *Container) ExecExists(eid string) (bool, error) {
 // ContainerCreate creates a container.
 func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (containertypes.ContainerCreateCreatedBody, error) {
 	defer trace.End(trace.Begin(""))
-	// TODO: Add Op logging to this operation. This is the start of container create operation and op logging here is very valuable.
 
 	var err error
 
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -1269,7 +1269,7 @@ func removeAnonContainerVols(pl *client.PortLayer, cID string, vc *viccontainer.
 		volFields := strings.SplitN(vol, ":", 3)
 
 		// NOTE(mavery): this check will start to fail when we fix our metadata correctness issues
-		if len(volFields) != 3 {
+		if len(volFields) >= 3 {
 			log.Debugf("Invalid entry in the volumes metadata section for container %s: %s", cID, vol)
 			continue
 		}
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -1253,6 +1253,7 @@ func removeAnonContainerVols(pl *client.PortLayer, cID string, vc *viccontainer.
 		fields := strings.SplitN(entry, ":", 2)
 		if len(fields) >= 2 {
 			log.Errorf("Invalid entry in the HostConfig.Binds metadata section for container %s: %s", cID, entry)
+			continue
 		}
 		destPath := fields[1]
 		namedMaskList[destPath] = struct{}{}
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -1247,20 +1247,20 @@ func removeAnonContainerVols(pl *client.PortLayer, cID string, vc *viccontainer.
 	// NOTE: these strings come in the form of <volume id>:<destination path>
 	namedVolumes := vc.HostConfig.Binds
 
-	// assemble white list of volume paths before processing binds. MUST be paths, as we want to move to honoring the proper metadata in the "volumes" section in the future.
-	whitelist := make(map[string]struct{}, 0)
+	// assemble a mask of volume paths before processing binds. MUST be paths, as we want to move to honoring the proper metadata in the "volumes" section in the future.
+	namedMaskList := make(map[string]struct{}, 0)
 	for _, entry := range namedVolumes {
 		fields := strings.SplitN(entry, ":", 2)
-		if len(fields) != 2 {
-			log.Debugf("Invalid entry found in the HostConfig.Binds metadata for container %s: entry---> %s", cID, entry)
+		if len(fields) >= 2 {
+			log.Errorf("Invalid entry in the HostConfig.Binds metadata section for container %s: %s", cID, entry)
 		}
 		destPath := fields[1]
-		whitelist[destPath] = struct{}{}
+		namedMaskList[destPath] = struct{}{}
 	}
 
 	joinedVols, err := fetchJoinedVolumes()
 	if err != nil {
-		log.Warnf("Unable to obtain joined volumes from portlayer, skipping removing anonymous volumes for %s: %s", cID, err.Error())
+		log.Errorf("Unable to obtain joined volumes from portlayer, skipping removing anonymous volumes for %s: %s", cID, err.Error())
 		return
 	}
 
@@ -1268,15 +1268,15 @@ func removeAnonContainerVols(pl *client.PortLayer, cID string, vc *viccontainer.
 		// Extract the volume ID from the full mount path, which is of form "id:mountpath:flags" - see getMountString().
 		volFields := strings.SplitN(vol, ":", 3)
 
-		// XXX: this check will start to fail when we fix our metadata correctness issues
+		// NOTE(mavery): this check will start to fail when we fix our metadata correctness issues
 		if len(volFields) != 3 {
-			log.Debugf("Invalid Entry found in the Volumes metadata seciont for container %s: entry---> %s", cID, vol)
+			log.Debugf("Invalid entry in the volumes metadata section for container %s: %s", cID, vol)
 			continue
 		}
 		volName := volFields[0]
 		volPath := volFields[1]
 
-		_, isNamed := whitelist[volPath]
+		_, isNamed := namedMaskList[volPath]
 		if _, joined := joinedVols[volName]; !joined && !isNamed {
 			_, err := pl.Storage.RemoveVolume(storage.NewRemoveVolumeParamsWithContext(ctx).WithName(volName))
 			if err != nil {
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -394,6 +394,11 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
 	// remove overlapping specified volume entries from the config.Config.Volumes map since they are superseding the anonymous behavior of image volumes.
 	for _, v := range config.HostConfig.Binds {
 		fields := strings.SplitN(v, ":", 2)
+		if len(fields) != 2 {
+			log.Error("HostConfig.Binds has an incorrect entry: %s should have two distinct entries separate by the ':' separator", v)
+			return nil, fmt.Errorf("Container HostConfig.Bind entry had an invalid entry: %s please check your docker client version and docker api version", v)
+		}
+
 		destination := fields[1]
 		delete(config.Config.Volumes, destination)
 	}
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -1261,7 +1261,7 @@ func removeAnonContainerVols(pl *client.PortLayer, cID string, vc *viccontainer.
 
 	joinedVols, err := fetchJoinedVolumes()
 	if err != nil {
-		log.Errorf("Unable to obtain joined volumes from portlayer, skipping removing anonymous volumes for %s: %s", cID, err.Error())
+		log.Errorf("Unable to obtain joined volumes from portlayer, skipping removal of anonymous volumes for %s: %s", cID, err.Error())
 		return
 	}
 
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -391,6 +391,13 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
 	}
 	log.Infof("Finalized volume list: %#v", volList)
 
+	// remove overlapping specified volume entries from the config.Config.Volumes map since they are superseding the anonymous behavior of image volumes.
+	for _, v := range config.HostConfig.Binds {
+		fields := strings.SplitN(v, ":", 2)
+		destination := fields[1]
+		delete(config.Config.Volumes, destination)
+	}
+
 	if len(config.Config.Volumes) > 0 {
 		// override anonymous volume list with generated volume id
 		for _, vol := range volList {
@@ -1253,10 +1260,12 @@ func removeAnonContainerVols(pl *client.PortLayer, cID string, volumes map[strin
 		volFields := strings.SplitN(vol, ":", 2)
 		volName := volFields[0]
 		if _, joined := joinedVols[volName]; !joined {
+			log.Debugf("Attempting to remove Anonymous volume %s from container %s", volName, cID)
 			_, err := pl.Storage.RemoveVolume(storage.NewRemoveVolumeParamsWithContext(ctx).WithName(volName))
 			if err != nil {
 				log.Debugf("Unable to remove anonymous volume %s in container %s: %s", volName, cID, err.Error())
 			}
+			log.Debugf("Successfully removed anonmyous volume %s", volName)
 		}
 	}
 }
--- a/lib/apiservers/engine/backends/container_proxy.go
+++ b/lib/apiservers/engine/backends/container_proxy.go
@@ -1278,12 +1278,14 @@ func removeAnonContainerVols(pl *client.PortLayer, cID string, vc *viccontainer.
 		volPath := volFields[1]
 
 		_, isNamed := namedMaskList[volPath]
-		if _, joined := joinedVols[volName]; !joined && !isNamed {
+		_, joined := joinedVols[volName]
+		if !joined && !isNamed {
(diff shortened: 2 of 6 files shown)


lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -391,6 +391,13 @@ func (c *ContainerProxy) AddVolumesToContainer(handle string, config types.Conta
 	}
 	log.Infof("Finalized volume list: %#v", volList)
 
+	// remove overlapping specified volume entries from the config.Config.Volumes map since they are superseding the anonymous behavior of image volumes.
+	for _, v := range config.HostConfig.Binds {
+		fields := strings.SplitN(v, ":", 2)
+		destination := fields[1]
```
anchal-agrawal: Should we do a len check for `fields` here for defensive purposes?

matthewavery: Yes we can, in general this field should honestly always have this split input wise. But defensively we should definitely log if the lengths are wrong and return an error potentially. The `HostConfig.Binds` indicates that a call such as `docker -v my-volume:/goes/right/here` another way for docker to populate this field is `docker -v /this/is/my/data:/it/goes/here` in both cases to end up in the binds metadata you should absolutely meet the criteria for this split. I think it would likely indicate a regression upstream from here if the length check failed. Some thought into the logging message should be considered. 

From a docker perspective, a `docker create` option should rarely fail and we should work our hardest to not fail out. in this case though, I would find it hard for us to continue since it would be a case of incorrect metadata passed to us. I am not really sure how we would excercise this check outside of a unit test potentially. I will look at a way to ensure this does not regress, but I am not aware of a way to force that situation from docker client perspective.

matthewavery:
I have looked at the behavior of vanilla docker in this regard. when a named volume supersedes an image volume it still shows up

lib/apiservers/engine/backends/container.go
```diff
@@ -565,11 +565,10 @@ func (c *Container) ExecExists(eid string) (bool, error) {
 	return true, nil
 }
 
-// docker's container.stateBackend
-
 // ContainerCreate creates a container.
 func (c *Container) ContainerCreate(config types.ContainerCreateConfig) (containertypes.ContainerCreateCreatedBody, error) {
 	defer trace.End(trace.Begin(""))
+	// TODO: Add Op logging to this operation. This is the start of container create operation and op logging here is very valuable.
```
hickeng: This is a general statement for all of the direct rest endpoint if only from an audit logging perspective.
Either add the logging or drop the comment.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1250,13 +1266,23 @@ func removeAnonContainerVols(pl *client.PortLayer, cID string, volumes map[strin
 
 	for vol := range volumes {
 		// Extract the volume ID from the full mount path, which is of form "id:mountpath:flags" - see getMountString().
-		volFields := strings.SplitN(vol, ":", 2)
+		volFields := strings.SplitN(vol, ":", 3)
+
+		// XXX: this check will start to fail when we fix our metadata correctness issues
```
anchal-agrawal: Instead of an `XXX:` we should perhaps use a `NOTE(<name>):` so it's easier to address in the future (I'm not sure what an `XXX` denotes here).

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1241,7 +1241,23 @@ func (c *ContainerProxy) Remove(vc *viccontainer.VicContainer, config *types.Con
...
+	volumes := vc.Config.Volumes
+	// NOTE: these strings come in the form of <volume id>:<destination path>
+	namedVolumes := vc.HostConfig.Binds
+
+	// assemble white list of volume paths before processing binds. MUST be paths, as we want to move to honoring the proper metadata in the "volumes" section in the future.
+	whitelist := make(map[string]struct{}, 0)
+	for _, entry := range namedVolumes {
+		fields := strings.SplitN(entry, ":", 2)
+		if len(fields) != 2 {
+			log.Debugf("Invalid entry found in the HostConfig.Binds metadata for container %s: entry---> %s", cID, entry)
```
hickeng: This should be a warning or error unless we expect it to occur

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1250,13 +1266,23 @@ func removeAnonContainerVols(pl *client.PortLayer, cID string, volumes map[strin
 
 	for vol := range volumes {
 		// Extract the volume ID from the full mount path, which is of form "id:mountpath:flags" - see getMountString().
-		volFields := strings.SplitN(vol, ":", 2)
+		volFields := strings.SplitN(vol, ":", 3)
+
+		// XXX: this check will start to fail when we fix our metadata correctness issues
+		if len(volFields) != 3 {
+			log.Debugf("Invalid Entry found in the Volumes metadata seciont for container %s: entry---> %s", cID, vol)
```
anchal-agrawal: Formatting suggestions (capitalization, sentence structure and typos):

```go
log.Debugf("Invalid entry in the volumes metadata section for container %s: %s", cID, vol)
```

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1241,7 +1241,23 @@ func (c *ContainerProxy) Remove(vc *viccontainer.VicContainer, config *types.Con
...
+	namedVolumes := vc.HostConfig.Binds
+
+	// assemble white list of volume paths before processing binds. MUST be paths, as we want to move to honoring the proper metadata in the "volumes" section in the future.
+	whitelist := make(map[string]struct{}, 0)
+	for _, entry := range namedVolumes {
+		fields := strings.SplitN(entry, ":", 2)
+		if len(fields) != 2 {
+			log.Debugf("Invalid entry found in the HostConfig.Binds metadata for container %s: entry---> %s", cID, entry)
+		}
+		destPath := fields[1]
```
hickeng: assert that `len(fields) >=2` before accessing the field given we're coding for possible invalid entries.
Will be particularly relevant if we have an `""` entry in the map.

tests/test-cases/Group1-Docker-Commands/1-06-Docker-Run.robot
```diff
@@ -159,6 +159,25 @@ Docker run and auto remove
...
+Docker run and auto remove with anonymous volumes and named volumes
+       ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} ps -a
+       Should Be Equal As Integers  ${rc}  0
+       ${output}=  Split To Lines  ${output}
+       ${count}=  Get Length  ${output}
+       ${suffix}=  Evaluate  '%{DRONE_BUILD_NUMBER}-' + str(random.randint(1000,9999))  modules=random
+       Set Test Variable  ${namedImageVol}  non-anonymous-image-volume-${suffix}
+       Should Be Equal As Integers  ${rc}  0
+       Set Test Variable  ${imageVolumeContainer}  I-Have-Two-Anonymous-Volumes-${suffix}
+       ${rc}  ${c5}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run --rm --name ${imageVolumeContainer} -v ${namedImageVol}:/data/db -v /I/AM/ANONYMOOOOSE mongo bash
```
hickeng: could do with a check of the `mongo` image to ensure the `/data/db` path remains a defined volume. Unlikely to check and not a blocker for this PR.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1241,7 +1241,23 @@ func (c *ContainerProxy) Remove(vc *viccontainer.VicContainer, config *types.Con
...
 // anonymous volumes if they are dangling. Errors, if any, are only logged.
-func removeAnonContainerVols(pl *client.PortLayer, cID string, volumes map[string]struct{}) {
+func removeAnonContainerVols(pl *client.PortLayer, cID string, vc *viccontainer.VicContainer) {
+	// NOTE: these strings come in the form of <volume id>:<destination>:<volume options>
+	volumes := vc.Config.Volumes
+	// NOTE: these strings come in the form of <volume id>:<destination path>
+	namedVolumes := vc.HostConfig.Binds
+
+	// assemble white list of volume paths before processing binds. MUST be paths, as we want to move to honoring the proper metadata in the "volumes" section in the future.
+	whitelist := make(map[string]struct{}, 0)
```
hickeng: given this is `removeAnonContainerVols` a _whitelist_ to me implies those that will be deleted.
`maskList` or `preserve` or similar would be a more explicit description of it's purpose.

lib/apiservers/engine/backends/container.go
matthewavery: I will not be adding the logging, I will remove the comment. Adding op logging is defintely badly needed, but will require a decently sized refactor.

lib/apiservers/engine/backends/container_proxy.go
matthewavery: yes it should be an Error, this in general should not happen. Good catch! I thought I had made them all Error already.

lib/apiservers/engine/backends/container_proxy.go
matthewavery: We are using `SplitN` with a value of n==2. so in the case of an entry of `""` should only yeiled a single entry and fail the check. I will move it to n>=2. The statements are equivalent and will future proof us from the format changing 👍

tests/test-cases/Group1-Docker-Commands/1-06-Docker-Run.robot
matthewavery: @hickeng we should actually be covered by this case in my `1-04-Docker-Create` test modification. That test is designed to ensure that the metadata after creation indicates that a image volume was present and that we are binding a named volume over it. It checks that a `volume` path is present for `/data/db` which indicates that it was either requested anonymously or is a image volume. and it checks for the `HostConfig.Binds` entry which indicates that we mounted a named volume for that path`

matthewavery: Thinking about this further we will actually see my test fail whenever we achieve more metadata correctness. Forcing us to improve the test to also check that the specific path is there rather than the named volume name. That is a potential flaw in this test, but it may be important since it will force us to understand what has changed with the increased metadata correctness.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1241,22 +1241,48 @@ func (c *ContainerProxy) Remove(vc *viccontainer.VicContainer, config *types.Con
...
+	namedVolumes := vc.HostConfig.Binds
+
+	// assemble a mask of volume paths before processing binds. MUST be paths, as we want to move to honoring the proper metadata in the "volumes" section in the future.
+	namedMaskList := make(map[string]struct{}, 0)
+	for _, entry := range namedVolumes {
+		fields := strings.SplitN(entry, ":", 2)
+		if len(fields) >= 2 {
+			log.Errorf("Invalid entry in the HostConfig.Binds metadata section for container %s: %s", cID, entry)
+		}
+		destPath := fields[1]
```
hickeng: this can still cause a panic if `len(fields) == 1`. SplitN doesn't guarantee you that number of fields, it's a maximum.
https://play.golang.org/p/_Pw8WCCmtaA

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1241,22 +1241,48 @@ func (c *ContainerProxy) Remove(vc *viccontainer.VicContainer, config *types.Con
...
+	// NOTE: these strings come in the form of <volume id>:<destination>:<volume options>
+	volumes := vc.Config.Volumes
+	// NOTE: these strings come in the form of <volume id>:<destination path>
+	namedVolumes := vc.HostConfig.Binds
+
+	// assemble a mask of volume paths before processing binds. MUST be paths, as we want to move to honoring the proper metadata in the "volumes" section in the future.
+	namedMaskList := make(map[string]struct{}, 0)
+	for _, entry := range namedVolumes {
+		fields := strings.SplitN(entry, ":", 2)
+		if len(fields) >= 2 {
```
jzt: Are we looking for `/path/to/volume` format here (vs. `name:/path/to/volume`)?  I'm curious as to why we error, but I confess I don't know the expected format of "proper metadata".

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1241,22 +1241,48 @@ func (c *ContainerProxy) Remove(vc *viccontainer.VicContainer, config *types.Con
...
+			log.Errorf("Invalid entry in the HostConfig.Binds metadata section for container %s: %s", cID, entry)
+		}
+		destPath := fields[1]
+		namedMaskList[destPath] = struct{}{}
+	}
+
 	joinedVols, err := fetchJoinedVolumes()
 	if err != nil {
-		log.Warnf("Unable to obtain joined volumes from portlayer, skipping removing anonymous volumes for %s: %s", cID, err.Error())
+		log.Errorf("Unable to obtain joined volumes from portlayer, skipping removing anonymous volumes for %s: %s", cID, err.Error())
```
jzt: Minor nit: `skipping removal of anonymous volumes` reads better to me than `skipping removing...`

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1241,22 +1241,48 @@ func (c *ContainerProxy) Remove(vc *viccontainer.VicContainer, config *types.Con
...
-		if _, joined := joinedVols[volName]; !joined {
+		volPath := volFields[1]
+
+		_, isNamed := namedMaskList[volPath]
+		if _, joined := joinedVols[volName]; !joined && !isNamed {
 			_, err := pl.Storage.RemoveVolume(storage.NewRemoveVolumeParamsWithContext(ctx).WithName(volName))
 			if err != nil {
 				log.Debugf("Unable to remove anonymous volume %s in container %s: %s", volName, cID, err.Error())
 			}
+			log.Debugf("Successfully removed anonymous volume %s", volName)
```
jzt:
```go
_, err := pl.Storage.RemoveVolume(storage.NewRemoveVolumeParamsWithContext(ctx).WithName(volName))
if err != nil {
        log.Debugf("Unable to remove anonymous volume %s in container %s: %s", volName, cID, err.Error())
}
log.Debugf("Successfully removed anonymous volume %s", volName)
```
I think that last `log.Debugf` should be in an `else` block.

lib/apiservers/engine/backends/container_proxy.go
anchal-agrawal: +1, this log message came from me originally. Thanks @matthewavery!

lib/apiservers/engine/backends/container_proxy.go
anchal-agrawal: We may also want to log the container ID like we did in the `if` block's log statement.

lib/apiservers/engine/backends/container_proxy.go
matthewavery: Expected format is above in comments for the `volumes` and `namedVolumes` vars. This particular section should always be in the `<name>:<path>` format. this is the correct docker prescribed format for their metadata.

lib/apiservers/engine/backends/container.go
```diff
@@ -565,8 +565,6 @@ func (c *Container) ExecExists(eid string) (bool, error) {
 	return true, nil
 }
 
-// docker's container.stateBackend
```
anchal-agrawal: 👍

lib/apiservers/engine/backends/container_proxy.go
matthewavery: haha alright I will update that as well.

lib/apiservers/engine/backends/container_proxy.go
jzt: @matthewavery then shouldn't it be `if len(fields) > 2` if we expect it to be 2? Currently you log an error on the `len(fields) == 2` case.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -1241,22 +1241,51 @@ func (c *ContainerProxy) Remove(vc *viccontainer.VicContainer, config *types.Con
...
+
+	// assemble a mask of volume paths before processing binds. MUST be paths, as we want to move to honoring the proper metadata in the "volumes" section in the future.
+	namedMaskList := make(map[string]struct{}, 0)
+	for _, entry := range namedVolumes {
+		fields := strings.SplitN(entry, ":", 2)
+		if len(fields) >= 2 {
+			log.Errorf("Invalid entry in the HostConfig.Binds metadata section for container %s: %s", cID, entry)
+			continue
+		}
+		destPath := fields[1]
```
sflxn: How do we ever get here?  If len(fields) == 2, we continue (from above).  Fields[1] seems like out of bound.

matthewavery: yeah I may have misread a comment from George on this one. it was originally `!=2`

PR #7161, opened 2018-01-25
Add unit tests to archives util_test.go

matthewavery:
This is a starter change to improve a small hole in testing for `/lib/archive/util.go`

### DIFF
--- a/lib/archive/util_test.go
+++ b/lib/archive/util_test.go
@@ -21,6 +21,25 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+// INPUT VALIDATION TESTS
+
+func TestNilSpecForAddInclusionExclusion(t *testing.T) {
+	var emptyMounts []string
+	err := AddMountInclusionsExclusions("", nil, emptyMounts, "")
+	assert.Error(t, err)
+
+	emptySpec := FilterSpec{
+		RebasePath: "",
+		StripPath:  "",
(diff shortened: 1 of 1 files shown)


lib/archive/util_test.go
```diff
@@ -21,6 +21,25 @@ import (
...
+	emptySpec := FilterSpec{
+		RebasePath: "",
+		StripPath:  "",
+		Exclusions: nil,
+		Inclusions: nil,
+	}
+
+	err = AddMountInclusionsExclusions("", &emptySpec, emptyMounts, "")
+	assert.Error(t, err)
+
```
jzt: whitespace

lib/archive/util_test.go
```diff
@@ -21,6 +21,24 @@ import (
...
+	emptySpec := FilterSpec{
+		RebasePath: "",
+		StripPath:  "",
+		Exclusions: nil,
+		Inclusions: nil,
+	}
+
+	err = AddMountInclusionsExclusions("", &emptySpec, emptyMounts, "")
```
anchal-agrawal: Is there a way to test a no-error case?

matthewavery: yes you can do an `assert.NoError(...`

matthewavery: But in this case we expect there to be an error. I am unsure of what you are asking here :)

PR #7164, opened 2018-01-25
Allow `make push` to replace and relaunch Appliance components

gigawhitlocks:
Hackathon task

### DIFF
--- a/infra/scripts/replace-running-components.sh
+++ b/infra/scripts/replace-running-components.sh
@@ -67,7 +67,6 @@ password=$(govc env | grep GOVC_PASSWORD | cut -d= -f2)
 target=$(govc env | grep GOVC_URL | cut -d= -f2)
 
 
-echo "is the vch name ambiguous?"
 if [[ ! -v VIC_ID ]] && [[ $(vch-name-is-ambiguous) ]]; then
     echo "The provided VIC name is ambiguous; please choose the correct VCH ID from the output below and assign it to the environment variable VIC_ID, e.g., export VIC_ID=12"
     $VIC_DIR/bin/vic-machine-linux ls --target $GOVC_URL --user $GOVC_USER --password=$GOVC_PASSWORD --thumbprint=$(get-thumbprint)
@@ -75,11 +74,9 @@ if [[ ! -v VIC_ID ]] && [[ $(vch-name-is-ambiguous) ]]; then
 fi
 
 if [[ ! -v VIC_ID ]]; then
-    echo "Getting VCH ID"
     VIC_ID=$($VIC_DIR/bin/vic-machine-linux ls --target=$target --user=$username --password=$password --thumbprint=$(get-thumbprint) | grep $VIC_NAME | awk '{print $1}')
 fi
 
-echo "Enabling SSH access on your VCH"
 $VIC_DIR/bin/vic-machine-linux debug --target=$target --id=$VIC_ID --user=$username --password=$password --thumbprint=$(get-thumbprint)
 
 VCH_IP="$(govc vm.ip $VIC_NAME)"


Makefile
```diff
@@ -272,6 +272,20 @@ install-govmomi:
 
 test: install-govmomi portlayerapi $(TEST_JOBS)
 
+push:
+	./infra/scripts/replace-running-components.sh
```
hickeng: Use `$(BASE_DIR)//infra/scripts/replace-running-components.sh` - I know it's not consistent throughout the Makefile currently but it should be.

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,67 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+for x in $(echo GOVC_USERNAME GOVC_PASSWORD VCH_NAME GOVC_URL GOPATH GOVC_INSECURE); do
```
hickeng: Use `govc env` to dump out the necessary fields and check them. This doesn't care how the user/password are supplied so is tolerant to more configurations.

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,67 @@
...
+    if [[ ! -v $x ]]; then
+        echo "Insufficient inputs. Please set $x environment variable and re-execute this script.";
+        echo "GOVC_USERNAME: username on ESX/vCenter target"
+        echo "GOVC_PASSWORD: password on ESX/vCenter target"
+        echo "VCH_NAME: name of VCH; matches --name argument for vic-machine"
+        echo "GOVC_URL: IP or FQDN of your vCenter/ESX target"
+        echo "GOPATH: your GOPATH, obviously"
```
hickeng: Don't use GOPATH - it's not actually needed in what we're doing.
You can use the following to determine the directory of the script and then go relative from there:
```
BASE_DIR=$(dirname $(readlink -f "$BASH_SOURCE"))
```
https://github.com/vmware/vic/blob/b82d95f35626f3f46f60b78e87c5d490acb7a7a5/isos/base/utils.sh#L19

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,67 @@
...
+        echo "GOVC_INSECURE: set to 1 to disable tls verify when using govc to talk to ESX/vC"
+        exit 1
+    fi;
+done
+
+function get-thumbprint () {
+    openssl s_client -connect $GOVC_URL:443 </dev/null 2>/dev/null \
```
hickeng: check out `govc about.cert --help`, particularly `govc about.cert --thumbprint`
It removes the additional dependency on having openssl available (I know it's very common but I'd prefer to assume `govc`.

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,67 @@
...
+        | openssl x509 -fingerprint -noout \
+        | cut -d= -f2
+}
+
+$GOPATH/src/github.com/vmware/vic/bin/vic-machine-linux debug --target=$GOVC_URL --name=$VCH_NAME --user=$GOVC_USERNAME --password=$GOVC_PASSWORD --thumbprint=$(get-thumbprint)
```
hickeng: I recommend either explicitly specifying the compute-resource or using the `--id` parameter instead. This has the potential for name conflicts if in VC

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,67 @@
...
+
+VCH_IP="$(govc vm.ip $VCH_NAME)"
+
+function replace-component() {
+    sshpass -p 'password' scp -oUserKnownHostsFile=/dev/null -oStrictHostKeyChecking=no $GOPATH/src/github.com/vmware/vic/bin/$1 root@$VCH_IP:/tmp/$1
+    on-vch mv /sbin/$1 /tmp/old-$1
+    on-vch mv /tmp/$1 /sbin/$1
+    on-vch chmod 755 /sbin/$1
+    pid=$(on-vch ps -e --format='pid,args' | grep $1 | grep -v grep | awk '{print $1}')
+    on-vch kill -9 $pid
```
hickeng: It's not in yet, but I expect to merge https://github.com/vmware/vic/pull/6943 soon. That moves to providing for a clean shutdown via SIGTERM. Just a note that we should update this in whichever merges second.

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,67 @@
...
+
+if [[ $1 == "" ]]; then
+    for x in port-layer-server docker-engine-server vicadmin vic-init; do
+        replace-component $x
+    done
+elif [[ $1 == "vic-init" ]]; then
+    replace-component vic-init
+else
+    replace-component $1
+    replace-component vic-init
```
hickeng: This should not be necessary - I tried it on the flight for other reasons. It takes a couple of seconds due to the throttling but it happily relaunched the docker-personality for me.
We should dig into what's actually occurring and not mask it with this.

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,67 @@
...
+fi
+
+on-vch vic-init &
```
hickeng: Likewise the systemd unit should be configured for restart if it's not currently which would negate the need for an explicit launch.

gigawhitlocks:
Thanks for the feedback @hickeng I'll clean this up

hackathon code: _all the shortcuts_ LOL

gigawhitlocks:
OK @hickeng that should address all of your feedback except the systemd unit (which I'll take care of in the morning when I'm warming up before I get back on the tar issue again) and also I did a bit of cleanup/organization/polish, took some liberties with allowing the user to provide the VCH name instead of the ID and as long as it comes back unique it will figure out the ID and continue, but if the ID is provided it'll just use that instead, and adjusted the error handling to be a bit better.

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,99 @@
...
+    $VIC_DIR/bin/vic-machine-linux ls --target $GOVC_URL --user $GOVC_USER --password=$GOVC_PASSWORD --thumbprint=$(get-thumbprint)
+    exit 1
+fi
+
+if [[ ! -v VIC_ID ]]; then
+    echo "Getting VCH ID"
+    VIC_ID=$($VIC_DIR/bin/vic-machine-linux ls --target=$target --user=$username --password=$password --thumbprint=$(get-thumbprint) | grep $VIC_NAME | awk '{print $1}')
+fi
+
+echo "Enabling SSH access on your VCH"
```
gigawhitlocks: feck

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,96 @@
...
+
+# Check for our one required argument
+if [[ ! -v VIC_NAME ]] && [[ ! -v VIC_ID ]]; then
+    echo "Please set one of the following environment variables to specify the VCH which you would like to reconfigure:"
+    echo "VIC_NAME: name of VCH; matches --name argument for vic-machine"
+    echo "VIC_ID: ID of VCH, as displayed in output of vic-machine ls"
+    exit 1
+fi
+
+username=$(govc env | grep GOVC_USERNAME | cut -d= -f2)
```
sflxn: This will be problem for those of us who define GOVC_URL="user:pass@targeturl".  We could try to grab user/pass from GOVC_URL if GOVC_USERNAME and GOVC_PASSWORD are not defined, but I see you do echo usage info explaining this.

matthewavery: `govc env` should break those out already if you only assign `GOVC_URL` so this should be safe from this concern. By using `govc env` it should always have all of the required fields for GOVC populated for consumption, which should avoid a "something isn't set right" style of error.

sflxn: oh, interesting.  I didn't know about this option.

gigawhitlocks:
ahahaha i messed up my repo and need to close and re-open this, sorry

PR #7165, opened 2018-01-25
Speed up compilation time.

vburenin:
Removed unnecessary dependency on a portlayer. Reduces scope of the the generated admiral client code.

Fixes #


pkg/vsphere/datastore/datastore.go
```diff
@@ -36,6 +36,15 @@ import (
 	"github.com/vmware/vic/pkg/vsphere/tasks"
 )
 
+// All paths on the datastore for images are relative to <datastore>/VIC/
+var StorageParentDir = "VIC"
```
hickeng: There is an explicit `lib/constants` package for sharing this kind of shared constant.
The specifics like this don't really belong in the supposedly generic `pkg` implementations.

Can you move the constants under `lib/constants` and update the references please.

vburenin: sounds good to me.

Makefile
```diff
@@ -343,7 +343,14 @@ PORTLAYER_DEPS ?= lib/apiservers/portlayer/swagger.json \
 
 $(admiralapi-client): lib/config/dynamic/admiral/swagger.json $(SWAGGER)
 	@echo regenerating swagger models and operations for Admiral API client...
-	@$(SWAGGER) generate client -A Admiral --target lib/config/dynamic/admiral -f lib/config/dynamic/admiral/swagger.json --tags /projects --tags /resources/compute --tags /config/registries 2>>swagger-gen.log
```
gigawhitlocks: this makes me happy based on the reformat alone

Makefile
```diff
@@ -343,7 +343,14 @@ PORTLAYER_DEPS ?= lib/apiservers/portlayer/swagger.json \
...
+	@$(SWAGGER) generate client -A Admiral --target lib/config/dynamic/admiral -f lib/config/dynamic/admiral/swagger.json \
+	    --tags /projects --tags /resources/compute --tags /config/registries \
+	    -O GetResourcesCompute -O GetProjects -O GetConfigRegistriesID \
+	    -M "com:vmware:photon:controller:model:resources:ComputeService:ComputeState" \
+	    -M "com:vmware:xenon:common:ServiceDocumentQueryResult" \
+	    -M "com:vmware:admiral:service:common:RegistryService:RegistryState" \
+	    -M "com:vmware:xenon:common:ServiceDocumentQueryResult:ContinuousResult" \
+	    -M "com:vmware:xenon:common:ServiceErrorResponse" 2>>swagger-gen.log
```
zjs: nit: Can we put `2>>swagger-gen.log` on its own line and sort the `-M` lines to make it easy for readers to see what's included?

vburenin: I updated it to be even better than that.

PR #7170, opened 2018-01-25
Added local ci testing target to makefile

### LINKED ISSUE #7162: Add ability to run local CI tests locally using make
As a VIC developer, I should be able to run CI tests locally using make.
As a VIC developer, I should be able to reuse an existing deployed VCH using the make target.
As a VIC developer, when I run a test that fails and the test breaks, I should have a method to get the output.xml to debug the failure.

---

This is one of the hackathon event during vic week 2018.

sflxn:
Make testing locally as friction-free as possible by

1. Adding a makefile target 'local-ci-test'
2. Using TARGET_VCH added in VIC 1.3 to use an existing VCH
3. Using a custom script that doesn't utilize drone so that if
   the test fails and craters, we can still access the logs

Resolves #7162

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ endif
 # Caches dependencies to speed repeated calls
 define godeps
 	$(call assert,$(call gmsl_compatible,1 1 7), Wrong GMSL version) \
-	$(if $(filter-out push push-portlayer push-docker push-vic-init push-vicadmin focused-test test check clean distclean mrrobot mark sincemark .DEFAULT,$(MAKECMDGOALS)), \
+	$(if $(filter-out push push-portlayer push-docker push-vic-init push-vicadmin focused-test test check clean distclean mrrobot mark sincemark local-ci-test .DEFAULT,$(MAKECMDGOALS)), \
 		$(if $(call defined,dep_cache,$(dir $1)),,$(info Generating dependency set for $(dir $1))) \
 		$(or \
 			$(if $(call defined,dep_cache,$(dir $1)), $(debug Using cached Go dependencies) $(wildcard $1) $(call get,dep_cache,$(dir $1))),
@@ -287,6 +287,10 @@ push-vic-init:
 push-vicadmin:
 	$(BASE_DIR)/infra/scripts/replace-running-components.sh vicadmin
 
+local-ci-test:
+	@echo running CI tests locally...
+	infra/scripts/local-ci.sh
+
 focused-test:
 # test only those packages that have changes
 	infra/scripts/focused-test.sh $(REMOTE)
@@ -336,7 +340,7 @@ $(imagec): $(call godeps,cmd/imagec/*.go) $(portlayerapi-client)
 
 $(docker-engine-api): $(portlayerapi-client) $(admiralapi-client) $$(call godeps,cmd/docker/*.go)
 ifeq ($(OS),linux)
-	@echo Building docker-engine-api server...
+	@echo building docker-engine-api server...
 	@$(TIME) $(GO) build $(RACE) -ldflags "$(LDFLAGS)" -o $@ ./cmd/docker
 else
 	@echo skipping docker-engine-api server, cannot build on non-linux
@@ -344,11 +348,12 @@ endif
 
 $(docker-engine-api-test): $$(call godeps,cmd/docker/*.go) $(portlayerapi-client)
 ifeq ($(OS),linux)
-	@echo Building docker-engine-api server for test...
+	@echo building docker-engine-api server for test...
 	@$(TIME) $(GO) test -c -coverpkg github.com/vmware/vic/lib/...,github.com/vmware/vic/pkg/... -outputdir /tmp -coverprofile docker-engine-api.cov -o $@ ./cmd/docker
 else
 	@echo skipping docker-engine-api server for test, cannot build on non-linux
 endif
+
 
 # Common portlayer dependencies between client and server
 PORTLAYER_DEPS ?= lib/apiservers/portlayer/swagger.json \
--- a/infra/scripts/local-ci.sh
+++ b/infra/scripts/local-ci.sh
@@ -1,0 +1,215 @@
+#!/bin/bash
+# Copyright 2018 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+##### defaults
+secretsfile=""
+docker_test="Group1-Docker-Commands"
+target_vch=""
+odir="ci-results"
+ci_container="gcr.io/eminent-nation-87317/vic-integration-test:1.39"
+github_api_key=""
+test_url=""
+test_datastore=""
+test_username=""
+test_password=""
+BASE_DIR=$(dirname $(readlink -f "$BASH_SOURCE"))
+vic_dir=${BASE_DIR}/../../
+
+##### utility functions
+function usage() {
+  echo "Usage: $0 [options]" 1>&2
+  echo
+  echo "  Options can be provided by commandline argument, environment variable, or a secrets yaml file containing the"
+  echo "  variables.  If a secrets file is not provided, this script will attempt to retrieve some info from govc, such as"
+  echo "  TEST_URL, TEST_USERNAME, and TEST_PASSWORD."
+  echo
+  echo "  options:"
+  echo "    -t DOCKER_TEST (or env var)"
+  echo "    -v TARGET_VCH (or env var) name of VCH"
+  echo "    -f SECRETS_FILE (or env var)"
+  echo "    -g GITHUB_API_KEY (or env var)"
+  echo "    -u TEST_URL (or env var or in secretsfile)"
+  echo "    -s TEST_DATASTORE (or env var or in secretsfile)"
+  echo "    -n TEST_USERNAME (or env var or in secretsfile)"
+  echo "    -p TEST_PASSWORD (or env var or in secretsfile)"
+  echo "    -d debug dumps out all the inputs and results of the resulting options"
+  echo
+  echo "  example:"
+  echo "    $0 -t Group1-Docker-Commands -v my_vch -s test.secrets.nested -g xxxxxx"
+  echo "    $0 -t 1-01-Docker-Info.robot -v my_vch -s test.secrets.nested -g xxxxxx"
+  echo 
+  echo "    DOCKER_TEST=Group1-Docker-Commands/1-01-Docker-Info.robot TARGET_VCH=my_vch SECRETS_FILE=test.secrets.nested $0"
+  echo
+  echo "    $0 -s test.secrets.nested (all params defined in secrets file)"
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -272,6 +272,13 @@ install-govmomi:
 
 test: install-govmomi portlayerapi $(TEST_JOBS)
 
+local-ci-test:
+ifndef TARGET_VCH
+	@echo To run ci tests, you must set TARGET_VCH to an already deployed VCH
+else
+	@echo running ci-tests locally...
+	infra/scripts/local-ci.sh ./test.secrets.nested
```
hickeng: Use `$(BASE_DIR)/infra/scripts/` as it allows calling from directories other than the build

infra/scripts/local-ci.sh
```diff
@@ -0,0 +1,68 @@
...
+
+# usage() {
+#   echo "Usage: $0 [-d DATACENTER] [-c CLUSTER] VCSA_IP ESX_IP..." 1>&2
+#   exit 1
+# }
+
+username='nloc'
+secretfile="$1"
+odir="ci-results"
+
```
hickeng: The following allows you to get the directory of the script so that you don't need to assume the callers directory:
```
BASE_DIR=$(dirname $(readlink -f "$BASH_SOURCE"))
```

infra/scripts/local-ci.sh
```diff
@@ -0,0 +1,68 @@
...
+echo "Running tests"
+github_api_key="$(grep 'GITHUB_AUTOMATION_API_KEY' ${secretfile} | awk '{ print $2 }')"
+test_url="$(grep 'TEST_URL_ARRAY' ${secretfile} | awk '{ print $2 }')"
+test_datastore="$(grep -E '\s+TEST_DATASTORE' ${secretfile} | awk '{ print $2 }')"
+test_username="$(grep 'TEST_USERNAME' ${secretfile} | awk '{ print $2 }')"
+test_password="$(grep 'TEST_PASSWORD' ${secretfile} | awk '{ print $2 }')"
+suite="$(grep -E '\s+TESTS' ${secretfile} | awk '{ print $2 }')"
+target_vch=$TARGET_VCH
+vic_dir=$(pwd)
```
hickeng: `vic_dir=$(BASE_DIR)/../../`

infra/scripts/local-ci.sh
```diff
@@ -0,0 +1,68 @@
...
+
+username='nloc'
+secretfile="$1"
+odir="ci-results"
+
+mkdir -p $odir
+
+echo "Running tests"
+github_api_key="$(grep 'GITHUB_AUTOMATION_API_KEY' ${secretfile} | awk '{ print $2 }')"
+test_url="$(grep 'TEST_URL_ARRAY' ${secretfile} | awk '{ print $2 }')"
```
hickeng: For all bar the github token you may well be able to use the output from `govc env` - we heavily leverage those environment variables already so likely their set.
They are being used in #7164 and the current `bash-helpers.sh` script sets them as part of the profiles.

hickeng:
Need to add support for TLS configuration:
1. check the VCH config for TLS config: no-tls, no-tlsverify, or tlsverify 
2. if no-tlsverify ensure `unset DOCKER_TLS_VERIFY`
3. if tlsverify ensure that `export DOCKER_TLS_VERIFY=set_by_local_ci` and that `DOCKER_CERT_PATH` is set

We may be able to add some better logic to vic-machine CLI to determine `DOCKER_CERT_PATH` during inspect and should update `infra/scripts/bash-helpers.sh` to allow configuring environment for a specific deployed VCH; we have `vic-inspect` but need to add `vic-select`.

tests/README.md
```diff
@@ -78,6 +78,41 @@ Use ./local-integration-test.sh
   ./tests/robot.sh tests/test-cases/Group6-VIC-Machine/6-04-Create-Basic.robot
   ```
 
+## Run Docker command tests via makefile target
+
+There exist a makefile target for developers to run the docker command tests locally (not in CI environment) against a pre-deployed VCH.  This is a fast way for contributors to test their potential PR submissions against the CI tests locally before pushing the commit.  There is also another benefit gained from using the makefile target.  With the way it is setup, logs from the run are written out to vic/ci-results, even if the tests fails.  The method described above to run the tests locally with drone has the weakness that a failure in the test can sometimes result in no written logs to help debug the failure.
```
mdubya66: There exists a makefile target for developers to run the docker command tests locally (not in CI environment) against a pre-deployed VCH.  This is a fast way for contributors to test their potential code chages, against the CI tests locally, before pushing the commit.  There is another benefit gained from using the makefile target, the way it is setup, logs from the run are written out to vic/ci-results. Even if the tests fail.  The method described above, to run the tests locally with drone, has the weakness that a failure in the test can sometimes result in no written logs to help debug the failure.

tests/README.md
```diff
@@ -78,6 +78,41 @@ Use ./local-integration-test.sh
...
+
+There are a few requirements before using this makefile target.
+
+1. A VCH must be pre-deployed before calling this makefile target
+2. The makefile target relies on a script that looks for a few more secrets variable, TEST_URL_ARRAY, TEST_USERNAME, TEST_PASSWORD, and TEST_DATASTORE
```
mdubya66: The makefile target relies on a script that looks for a few more secrets variables

tests/README.md
```diff
@@ -78,6 +78,41 @@ Use ./local-integration-test.sh
...
+```
+To run these tests using the makefile target,
+
+```
+make local-ci-test secrets=test.secrets.nested dockertest=Group1-Docker-Commands/1-01-Docker-Info.robot vch=my_vch
+
+make local-ci-test secrets=test.secrets.nested dockertest=Group1-Docker-Commands vch=my_vch
+```
+
+Currently, only the Group1 tests are setup to use an existing VCH so this makefile target only works on the group 1 tests.  In the above, the first line runs a specific docker command test.  The second line runs the entire group 1 tests.  Notice, we define extra params to the *local-ci-test* target by specifying the param's name.  There are three parameters, *secrets*, *dockertest*, and *vch*.  The *vch* parameter must target an existing pre-deployed VCH.
```
mdubya66: Currently only the Group1 tests are setup to use an existing VCH, so this makefile target only works on the Group1 tests.  In the above, the first line runs a specific docker command test.  The second line runs the entire Group1 tests.  Notice, we define extra params to the *local-ci-test* target by specifying the param's name.  There are three parameters, *secrets*, *dockertest*, and *vch*.  The *vch* parameter must target an existing pre-deployed VCH.

tests/README.md
```diff
@@ -78,6 +78,41 @@ Use ./local-integration-test.sh
...
+
+A helpful tip is to create different secrets file for different environments.  For instance, test.secrets.esx and test.secrets.vc for an ESX host and VC cluster, respectively.
```
mdubya66: A helpful tip is to create different secrets files for different environments.  For instance, test.secrets.esx and test.secrets.vc for an ESX host and VC cluster, respectively.

infra/scripts/local-ci.sh
```diff
@@ -0,0 +1,198 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+##### defaults
+secretsfile=""
+docker_test="Group1-Docker-Commands"
+target_vch=""
+odir="ci-results"
+ci_container="gcr.io/eminent-nation-87317/vic-integration-test:1.39"
```
hickeng: @mhagen-vmware Are we tagging this with `latest` as well as versions? For the build/test portions of the system I'm all for explicit versioning, but for the dev scripts my preference is to use latest available versions without requiring update.

infra/scripts/local-ci.sh
```diff
@@ -0,0 +1,198 @@
...
+  secretsfile=${SECRETS_FILE}
+  GetParamsFromSecrets
+else
+  GetParamsFromSecrets
+fi
+
+if [[ -z "${TARGET_VCH}" ]] && [[ -z $target_vch ]] && [[ -z $secrets_target_vch ]]; then 
+  usage
+elif [[ -z $target_vch ]] ; then
+  target_vch=${TARGET_VCH}
```
hickeng: I think this needs to be:
```
target_vch=${TARGET_VCH:-$secrets_target_vch}
```
and the whole may be easier to write as:
```
target_vch=${TARGET_VCH:-$secrets_target_vch}
if [[ -z "${target_vch}" ]]; then
   usage
fi
```

same conciseness comment applies to the rest of the tests.

infra/scripts/local-ci.sh
```diff
@@ -0,0 +1,198 @@
...
+docker_test="Group1-Docker-Commands"
+target_vch=""
+odir="ci-results"
+ci_container="gcr.io/eminent-nation-87317/vic-integration-test:1.39"
+github_api_key=""
+test_url=""
+test_datastore=""
+test_username=""
+test_password=""
+vic_dir=$(pwd)
```
hickeng: This is used to locate test sources so be explicit by basing of path of the script in the repo:
```
BASE_DIR=$(dirname $(readlink -f "$BASH_SOURCE"))
vic_dir=${BASE_DIR}/../../
```

infra/scripts/local-ci.sh
```diff
@@ -0,0 +1,198 @@
...
+                        -e GOVC_INSECURE=1 \
+                        -e GITHUB_AUTOMATION_API_KEY=${github_api_key}\
+                        -e TEST_URL_ARRAY=${test_url}\
+                        -e TEST_DATASTORE=${test_datastore}\
+                        -e TEST_USERNAME=${test_username}\
+                        -e TEST_PASSWORD=${test_password}\
+                        -e TARGET_VCH=${target_vch}\
+                        -e DEBUG_VCH=1\
+                        ${ci_container}\
+                        bash -c "pybot -d /vic/${odir} /vic/tests/test-cases/"$docker_test";\
```
hickeng: Looks like `odir` is bound to `/tmp` and not to `/vic/odir`.

hickeng: Using double quoted string there's no need to unquote to get shell variable replacement for `$docker_test`

infra/scripts/local-ci.sh
```diff
@@ -0,0 +1,198 @@
...
+                 mv *-container-logs.zip /tmp/ 2>/dev/null; \
+                 mv VCH-*-vmware.log /tmp/ 2>/dev/null; \
+                 mv vic-machine.log /tmp/ 2>/dev/null; \
+                 mv index.html* /tmp/ 2>/dev/null; \
+                 mv VCH-* /tmp/ 2>/dev/null")
```
hickeng: surprised to see these here - suggests to me that either I don't understand what the `pybot -d` option does, or it's not taken proper effect.

infra/scripts/local-ci.sh
```diff
@@ -0,0 +1,198 @@
...
+}
+
+##### Get command line arguments
+while getopts s:t:v:g:u:d:n:p: flag
+do
+  case $flag in
+    s)
+      secretsfile=$OPTARG
+      ;;
+    t)
```
hickeng: I think you can provide `pybot -t` to run only a specific test in a suite. We may want to use a different option here so adding pass-through for that later doesn't result in option name collision.

infra/scripts/local-ci.sh
```diff
@@ -0,0 +1,217 @@
...
+      usage
+      ;;
+  esac
+done
+
+##### Preconditions...
+
+# There is a priority in the preconditions.  First, environment variable.  Second, secrets file.  Third, command line argument.
+
+if [[ -z "${SECRETS_FILE}" ]] && [[ -z $secretsfile ]] ; then
```
hickeng:
```
if [[ -z "${secretsfile:-$SECRETS_FILE}" ]]; then
  GetGovcParamsFromEnv
else
  GetParamsFromSecrets
fi
```

PR #7172, opened 2018-01-26
Enforce container naming convention on all paths [specific ci=6-18-Container-Name-Convention]

### LINKED ISSUE #7159: docker rename breaks CNC
VIC Engine - 1.3.x

User statement: As a customer of VIC, I would like to have the [Container Naming Convention](https://github.com/vmware/vic/issues/6267) to be enforced at all times. However, a docker rename breaks this.

Steps to reproduce.

1. Deploy a VCH with --cnc cVM_xxx_{id}
2. Deploy any container and check vSphere UI; cVM appears as cVM_xxx_e09e04e5a20b
3. docker rename e09e04e5a20b Test
4. container friendly name changes to Test on docker ps
4. However, vSphere UI changes to Test-e09e04e5a20b, where it should have remained as-is.

hickeng:
This ensures we always call the name convention logic during rename paths.
Additionally this enforces maximum VM name length limits on the resulting names.
If the convention template as a whole somehow exceeds the maximum length we will revert to using default.
ID is replaced first, then Name. If truncation occurs it will be applied to the ID or Name insertions and not the template.

Adds:
* rename assertions to robot tests
* unit tests for convention transformations

Fixes #7159

### DIFF
--- a/lib/portlayer/util/util.go
+++ b/lib/portlayer/util/util.go
@@ -23,6 +23,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 
+	"github.com/vmware/vic/lib/config"
 	"github.com/vmware/vic/lib/constants"
 	"github.com/vmware/vic/lib/spec"
 )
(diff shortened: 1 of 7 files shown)


lib/portlayer/util/util.go
```diff
@@ -61,30 +62,49 @@ func ServiceURL(serviceName string) *url.URL {
...
-		if len(prettyName) > nameMaxLen-1 {
-			name = fmt.Sprintf("%s-%s", prettyName[:nameMaxLen-1], shortID)
-		} else {
-			name = fmt.Sprintf("%s-%s", prettyName, shortID)
+
+		name = strings.Replace(name, config.IDToken.String(), trunc, -1)
+		availableLen -= len(trunc)
+	}
+
+	// check for presence in template not working result so we don't have issues with recursive templates
```
zjs: Is this actually sufficient to avoid issues with recursive templates? It seems like we'd need to check for (1) presence in template and (2) no additionally presence in working result.

lib/portlayer/util/util.go
```diff
@@ -61,30 +62,49 @@ func ServiceURL(serviceName string) *url.URL {
...
+	if strings.Contains(namingConvention, config.NameToken.String()) {
```
zjs: Can this block be deduplicated with the above (both perform the same operation on a working result given a template and a value to attempt to substitute)?

lib/portlayer/util/util.go
```diff
@@ -61,30 +62,49 @@ func ServiceURL(serviceName string) *url.URL {
...
+	if namingConvention == "" {
+		namingConvention = config.DefaultNamePattern
+	}
+	name := namingConvention
+
+	// determine length of template without tokens
+	templateLen := len(strings.Replace(strings.Replace(namingConvention, config.NameToken.String(), "", -1), config.IDToken.String(), "", -1))
+	availableLen := constants.MaxVMNameLength - templateLen
+
+	if availableLen < 0 {
```
zjs: Should we (additionally) have validation of the length of the template when the CNC is set? If the user enters a template that will hit this case, we should at least warn then. (And in the case of the UI we should probably be reflecting this limit and the truncation logic in the generated examples.)

lib/portlayer/util/util.go
hickeng: It has the possibility of not replacing a token entry that shows up in one of the values but the tokens have illegal characters for names so _shouldnt_ hit that...

hickeng:
@zjs comments are addressed.

PR #7181, opened 2018-01-29
[specific ci=11-02-Upgrade-Exec] Fix Get Container IP Robot util keyword

anchal-agrawal:
This commit changes the 'Get Container IP' keyword's default value for
the network parameter to "bridge" instead of "default" as the keyword
was updated in commit 29eeb81 (#6641) to use container inspect instead
of network inspect. It also changes the 11-02-Upgrade-Exec test to
launch containers on the bridge network explicitly so the bridge network
is propagated to the 'Get Container IP' keyword.

Cherry picks df22cfd451aa6ca14881e260488301bb50a1d865 from #7152


anchal-agrawal:
Closing in favor of #7182

Issue #6516, opened 2017-10-05
VCH IPAM calculating the container network range wrongly.


malikkal:
VIC Engine - vic_1.2.1-rc2, vic_1.2.1

User statement: As a customer of VIC, I would like to use DHCP, where multiple VCH may share same CNW subnet; IP addresses being managed by DHCP server. 

Details: Please note #6515 is causing issues that forces continued use of CNR.

Issue Description: However, any CNR range specified with x-**127** (as below) would result in a /25 subnet as the IPRange inside VCH IPAM without regard to the starting octet specified.

```
--cnr 10.92.159.0-24_1159_uat_vic_cnw:10.92.159.5-127 OR
--cnr 10.92.159.0-24_1159_uat_vic_cnw:10.92.159.16-10.92.159.127
```

result; refer screenshot.
![image](https://user-images.githubusercontent.com/16984768/31213914-776e182a-a9db-11e7-907b-9f775b37edb7.png)

@hickeng

hickeng:
This has been on my list of items to look at after addressing #6515 (and a piece of follow on work in #6641 for returning useful IP addresses from the `docker port` command), however having checked just now I see exactly what I'd expect from a naive verification:

Network related options:
```
vic-machine-linux create ... 
--container-network=VMExternal:external
--container-network-firewall=VMExternal:Published
--container-network=VMExternal2:open
--container-network-firewall=VMExternal2:Open
--cnr=VMExternal2:192.168.78.10-192.168.78.20
--cng=VMExternal2:192.168.78.2/24
--container-network=VMExternal3:outbound
--container-network-firewall=VMExternal3:Outbound
--public-network-ip=192.168.78.127/24
--public-network-gateway=192.168.78.2
```

```
docker network inspect open
[
    {
        "Name": "open",
        "Id": "7fed3a71b6598733799935407eb82245bc492be11d772a10ca674bb4b19be373",
        "Created": "2017-11-01T00:09:52.492993271Z",
        "Scope": "",
        "Driver": "external",
        "EnableIPv6": false,
        "IPAM": {
            "Driver": "",
            "Options": {},
            "Config": [
                {
                    "Subnet": "192.168.78.0/24",
                    "IPRange": "192.168.78.10-192.168.78.20",
                    "Gateway": "192.168.78.2"
                }
            ]
        },
        "Internal": false,
        "Attachable": false,
        "Containers": {
[...]

PR #7183, opened 2018-01-29
[Needs Rebase] Preserve the starting offset within a CIDR range

hickeng:
This updates the IP range parsing to preserve the starting offset.
This is a behavioural delta from Go's normal semantic for ParseCIDR which masks the corresponding IP with the specified netmask.

This issue was exposed by the dual parsing of the range specification; on the first parse, because it's an `A-B` range it doesn't get processed by the standard `net.ParseCIDR`, however the result is that it's collapsed into an address/mask pair if it can be expressed that way; i.e. the last address aligns on a CIDR boundary. 
The second parse, when the portlayer loads the configuration, then uses the standard libraries and drops the final octet of the IP address.

There is follow up work because the last address in the range cannot be allocated. Can see below that the space is terminated at `.126`
```
Jan 29 2018 02:07:55.726Z DEBUG checking range: 192.168.68.125-192.168.68.126
Jan 29 2018 02:07:55.726Z DEBUG first address slected: 192.168.68.125
Jan 29 2018 02:07:55.726Z INFO  Reserving IP range [192.168.68.125, 192.168.68.125]
Jan 29 2018 02:07:55.726Z DEBUG adding alias nas:hungry_bassi -> hungry_bassi
Jan 29 2018 02:07:55.726Z DEBUG adding alias nas:80141f807be5 -> hungry_bassi
```

We can see that `.127` is pulled out of the config, but is explicitly reserved in a separate range. I'm unclear why at this time, or what the ERROR logging is about (hypothesis is that the `.0` and `.1` are related to the gateway - could `.127` be reserved for broadcast? @hmahmood ?):
```
Jan 29 2018 02:04:39.238Z DEBUG [BEGIN]  [vic/lib/portlayer/network.(*Context).reserveSubnet:422]
Jan 29 2018 02:04:39.238Z ERROR Cannot reserve IP range 192.168.68.0 - 192.168.68.127.  Not within pool's range 172.16.0.0 - 172.31.255.255
Jan 29 2018 02:04:39.238Z DEBUG [ END ]  [vic/lib/portlayer/network.(*Context).reserveSubnet:422] [8.743µs] 
Jan 29 2018 02:04:39.238Z DEBUG [BEGIN]  [vic/lib/portlayer/network.reservePools:451]
Jan 29 2018 02:04:39.238Z INFO  Reserving IP range [192.168.68.125, 192.168.68.127]
Jan 29 2018 02:04:39.238Z DEBUG [ END ]  [vic/lib/portlayer/network.reservePools:451] [5.658µs] 
Jan 29 2018 02:04:39.238Z INFO  Reserving IP range [192.168.68.127, 192.168.68.127]
Jan 29 2018 02:04:39.238Z ERROR Cannot reserve IP range 192.168.68.0 - 192.168.68.0.  Not within pool's range 192.168.68.125 - 192.168.68.127
```

- [ ] may need to confirm that the broadcast address aligns with the gateway mask, and not based on the CIDR size of the ip range. (it's unclear to me whether the broadcast address should be based on the netmask that includes the gateway, or the possibly smaller netmask that dictates the size of the available allocatable range). 
- [ ] ensure that either:
  - the first address in the CIDR range is configured as gateway (both in containers and on the endpoint), or
  - there is a direct rule created in containers to allow communication with the gateway that is configured (e.g. routing to 172.16.0.1 directly via the interfacec, even if CIDR is 127.16.0.128/30).

Towards #6516


hickeng:
Sample output with slightly modified output for better error message and debugging. Trying to determine how to provide a useful indicator as to which network is out of addresses if the container were connected to multiple:
```
vagrant@devbox:~/vicsmb/src/github.com/vmware/vic$ docker run -dit --net=nas alpine /bin/ash
Unable to find image 'alpine:latest' locally
latest: Pulling from library/alpine
ff3a5c916c92: Pull complete
a3ed95caeb02: Pull complete
Digest: sha256:8c03bb07a531c53ad7d0f6e7041b64d81f99c6e493cb39abba56d956b40eacbc
Status: Downloaded newer image for library/alpine:latest
99e5ed0645a68e27a687acf42aa55551c8637babf7addd2580fc059d1339a018
vagrant@devbox:~/vicsmb/src/github.com/vmware/vic$ docker run -dit --net=nas alpine /bin/ash
11630b883ce54e5cd19f18be83b5590a68ba26ee9beb005e918adead3542952a
vagrant@devbox:~/vicsmb/src/github.com/vmware/vic$ docker run -dit --net=nas alpine /bin/ash
2e9f1018f4ae3e4b7b1e941868c77c0527fb4dcb5740b63279affbc451f6ab77
vagrant@devbox:~/vicsmb/src/github.com/vmware/vic$ docker run -dit --net=nas alpine /bin/ash
6aeb9b976542f04f9b1465ac6be3607406415427b7bf9c7d8adc33becd2bbe1c
docker: Error response from daemon: Server error from portlayer: no IP addresses available in space &{Parent:<nil> Network:192.168.68.0/25 Pool:192.168.68.0/25 availableRanges:[0xc42011dda0 0xc42011ddd0]} 192.168.68.124-192.168.68.126 <nil> [].
vagrant@devbox:~/vicsmb/src/github.com/vmware/vic$ docker network inspect nas
[
    {
        "Name": "nas",
        "Id": "d7c988a4c05a6832b08e0e1bb37a685666891265a5d77446c986952a712cc021",
        "Created": "2018-01-29T05:15:03.819790798Z",
        "Scope": "",
        "Driver": "external",
        "EnableIPv6": false,
        "IPAM": {
            "Driver": "",
            "Options": {},
            "Config": [
                {
                    "Subnet": "192.168.68.0/25",
                    "IPRange": "192.168.68.124-192.168.68.126",
                    "Gateway": "192.168.68.1"
                }
            ]
        },
        "Internal": false,
        "Attachable": false,
        "Containers": {
            "11630b883ce54e5cd19f18be83b5590a68ba26ee9beb005e918adead3542952a": {
                "Name": "trusting_bell",
                "EndpointID": "11630b883ce54e5cd19f18be83b5590a68ba26ee9beb005e918adead3542952a",
                "MacAddress": "",
                "IPv4Address": "192.168.68.125/25",
                "IPv6Address": ""
            },
            "2e9f1018f4ae3e4b7b1e941868c77c0527fb4dcb5740b63279affbc451f6ab77": {
                "Name": "elastic_hopper",
                "EndpointID": "2e9f1018f4ae3e4b7b1e941868c77c0527fb4dcb5740b63279affbc451f6ab77",
                "MacAddress": "",
                "IPv4Address": "192.168.68.126/25",
                "IPv6Address": ""
            },
            "99e5ed0645a68e27a687acf42aa55551c8637babf7addd2580fc059d1339a018": {
                "Name": "happy_northcutt",
                "EndpointID": "99e5ed0645a68e27a687acf42aa55551c8637babf7addd2580fc059d1339a018",
                "MacAddress": "",
                "IPv4Address": "192.168.68.124/25",
                "IPv6Address": ""
            }
        },
        "Options": {},
        "Labels": {}
    }
]
```

pkg/ip/ip_test.go
```diff
@@ -157,3 +158,35 @@ func TestRangeNetwork(t *testing.T) {
...
+		{"10.10.10.128-10.10.10.129", "10.10.10.128/31"},
+		// test the boundaries of full subnets where the range is not the full subnet
+		{"10.10.10.125-10.10.10.127", "10.10.10.125/30"},
+		{"10.10.10.5-10.10.10.127", "10.10.10.5/25"},
+		{"10.10.10.0-10.10.10.127", "10.10.10.0/25"},
+		{"10.10.10.128-10.10.10.254", symmetric},
+		{"10.10.10.252-10.10.10.255", "10.10.10.252/30"},
+		// NOTE: I'm unsure why a single address translates into a range and not a /32
+		{"10.10.10.1-10.10.10.1", symmetric},
+		// {"10.10.10.2/32", symmetric},
```
anchal-agrawal: Is this an unneeded test case?

hmahmood:
@hmahmood ... you guessed right :-) The range is pared down since only certain addresses are routable and so can be handed out to hosts: https://www.cisco.com/c/en/us/support/docs/ip/dynamic-address-allocation-resolution/13711-40.html#subnetzero

pkg/ip/ip_test.go
hickeng: It's there because I would expect, given the other ranges, to have the range collapsed to `/32`. It's not that it's needed but that it's illustrating something that's not occurring. I don't want to enshrine it as a negative test case because I think it should be allowed as it's consistent with the other range behaviours that end on an aligned address.

hickeng:
@hmahmood Do you know why `.127` is getting reserved? Is it just because it's aligned at and the end of the range? It's not one that cannot be assigned.

hmahmood:
@hickeng http://jodies.de/ipcalc?host=192.168.68.0&mask1=25&mask2=

Its the broadcast address for `192.168.0/25`

hickeng:
- [ ] need to confirm that the broadcast address aligns with the gateway mask, and not based on the CIDR size of the ip range.

zjs:
It seems like this needs to be rebased before it can be effectively reviewed. Adding `[Needs Rebase]` to summary.

yuyangbj:
I will get started to learn this PR and continue to fix it.

yuyangbj:
Adding new PR https://github.com/vmware/vic/pull/8471 to track this fix

PR #7158, opened 2018-01-24
Fix user selectable log level for Longevity, addendum to PR 7077

lcastellano:
A couple of code review observation that came in after the original PR: 7077 was checked in.

PR #7185, opened 2018-01-29
Revert "Fix user selectable log level for Longevity, addendum to PR 7077"

mhagen-vmware:
Reverts vmware/vic#7158


mdubya66:
Unable to start longevity as it fails immediately with 

```sh: 0: Illegal option --```

PR #7201, opened 2018-01-30
Upgrade golang to 1.10

mhagen-vmware:
[full ci]
~Add pabot for nightlies and future potential CI parallelism~ and upgrade integration image to ~1.9.3~ 1.10 golang
openjdk needed to be upgraded, as what we were using is no longer in the repo and wget didn't need to be manually installed anymore as the newer one is installed by default


vburenin:
I guess we should already consider 1.9.4

vburenin:
This is probably too late already. We should switch over to Go 1.10.
Go 1.10 has some added benefits such as faster compilation using already pre-build packages.
Thus, if we start using a separate volume and point GOCACHE variable to it, every time build happens it will reuse part of already built binaries, so that we can significantly benefit from a quick CI runs for short builds.

PR #7214, opened 2018-01-31
Cosmetic readability improvements. Does not change workflow.

vburenin:
Cosmetic readability improvements. Does not change workflow.


lib/install/management/store_files.go
```diff
@@ -350,8 +350,10 @@ func (d *Dispatcher) deleteVolumeStoreIfForced(conf *config.VirtualContainerHost
 			dsURL.Path = path.Join(dsURL.Path, constants.VolumesDir)
 		}
 
-		d.op.Debugf("Provided datastore URL: %q\nParsed volume store path: %q", url.Path, dsURL.Path)
-		d.op.Infof("Deleting volume store %q on Datastore %q at path %q", label, dsURL.Host, dsURL.Path)
+		d.op.Debugf("Provided datastore URL: %q", url.Path)
+		d.op.Debugf("Parsed volume store path: %q", dsURL.Path)
+		d.op.Infof("Deleting volume store %q on Datastore %q at path %q",
```
hickeng: Please don't be this aggressive about line wrapping - it's not necessary with modern monitors and environments and makes it much harder to quickly scan for logic flow.

pkg/vsphere/vm/vm.go
```diff
@@ -519,20 +519,20 @@ func (vm *VirtualMachine) Properties(ctx context.Context, r types.ManagedObjectR
 	defer trace.End(trace.Begin(fmt.Sprintf("VM(%s) Properties(%s)", r, ps), op))
 
 	contains := false
-	for i := range ps {
-		if ps[i] == "summary" || ps[i] == "summary.runtime" {
+	for _, v := range ps {
+		if strings.HasPrefix(v, "summary") {
```
hickeng: This would also match `summary.config` - and in that circumstance we would not have the Runtime structure populated. I *think* this would result in a panic, but might be worse and just give you a zero initialized structure.
This could have been correctly phrased as:
```
strings.HasPrefix("summary.runtime") || strings.Equals("summary")
```

PR #7216, opened 2018-01-31
Tether messages moved from cmd/tether to lib/tether

vburenin:
Tether messages moved from cmd/tether to lib/tether


vburenin:
@hickeng cmd/tether is not the place for almost any dependency that used across the project. cmd should only be used to contain a bootstrap for binary that is expected to be built. The only exception can be CLI parameters. Or something that is really tightly belong to that bootstrap logic and not used anywhere else.

PR #7229, opened 2018-01-31
Checking for explicit string values instead of just a prefix

vburenin:
Checking for explicit string values instead of just a prefix. Using prefix happened to be not a safe option since summary.config key is also can be present. This is basically a rollback if to a previous "if" condition.


vburenin:
@anchal-agrawal Thanks Anchal. Squash and merge allows to edit commit message before merge is actually happens.

PR #7233, opened 2018-02-01
[skip-ci] Allow `make push` to replace and relaunch Appliance components

### LINKED ISSUE #7167: Add ability to easily replace binaries on Appliance with make
We would like to have some make targets for replacing binaries on the appliance without doing an ISO remaster or reboot of the appliance VM for incremental development.

gigawhitlocks:
Continuing from #7164 sorry, comments from that PR should be addressed at this time, except for a small modification to systemd in the appliance, which I'll have to modify and test. Resolves #7167


sgairo:
I think its [skip ci] for future reference

hickeng:
@gigawhitlocks I think this comment may have been meant for this PR: https://github.com/vmware/vic/pull/7170#issuecomment-361001565

If it's relevant, please note. If not I've not idea who I was writing that for.

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,96 @@
...
+# limitations under the License.
+
+set -e
+
+BASE_DIR=$(dirname $(readlink -f "$BASH_SOURCE"))
+VIC_DIR=$(dirname $(readlink -f $BASE_DIR/..))
+
+# Grab vSphere's thumbprint for calling vic-machine
+function get-thumbprint () {
+    govc about.cert | grep "SHA-1 Thumbprint" | awk '{print $NF}'
```
hickeng: `govc about.cert -k -thumbprint` gives you the thumbprint undercoated so you don't need the grep/awk.

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,96 @@
...
+}
+
+# Run the command given on the VCH instead of locally
+function on-vch() {
+    sshpass -ppassword ssh -oUserKnownHostsFile=/dev/null -oStrictHostKeyChecking=no root@$VCH_IP -C $@ 2>/dev/null
```
hickeng: could we use the `vic-machine debug --authorized-key` option?
It's a path I don't think we test sufficiently but would also avoid hard coding dev passwords into the scripts.

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,96 @@
...
+    exit 1
+fi
+
+username=$(govc env | grep GOVC_USERNAME | cut -d= -f2)
+password=$(govc env | grep GOVC_PASSWORD | cut -d= -f2)
+target=$(govc env | grep GOVC_URL | cut -d= -f2)
+
+
+if [[ ! -v VIC_ID ]] && [[ $(vch-name-is-ambiguous) ]]; then
+    echo "The provided VIC name is ambiguous; please choose the correct VCH ID from the output below and assign it to the environment variable VIC_ID, e.g., export VIC_ID=12"
```
hickeng: Check if the [script is interactive](https://www.gnu.org/software/bash/manual/html_node/Is-this-Shell-Interactive_003f.html) - if it is then use `read -p "Enter VCH ID: " VIC_ID` to get the variable from the user.

From a consistency perspective I suggest renaming that variable to be `VCH_ID` instead of `VIC_ID`

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,96 @@
...
+    for x in port-layer-server docker-engine-server vicadmin vic-init; do
+        replace-component $x
+    done
+else
+    for x in $@; do # replace provided list
+        replace-component $x
+    done
+fi
+
+on-vch vic-init &
```
hickeng: This really shouldn't be necessary unless configured at a debug level that causes vic-init to exit instead of relaunching components (think that's debug>=3)

I'm not sure what will happen here, but there is a risk of multiple vic-init's running - I don't know that there's anything that prevents that.

I'd suggest using `systemctl restart vic-init.service` to perform a restart if necessary.
I'd also ask you to follow up on why the restart is needed.

infra/scripts/replace-running-components.sh
gigawhitlocks: oh nice!

gigawhitlocks: still prints the IP so I do need awk, but not sed

gigawhitlocks:
Apologies @hickeng I moved everything around and made a bunch of improvements, in addition to addressing your concerns

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,173 @@
...
+# Run the command given on the VCH instead of locally
+function on-vch() {
+    ssh -oUserKnownHostsFile=/dev/null -oStrictHostKeyChecking=no -i $VIC_KEY root@$VCH_IP -C $@ 2>/dev/null
+}
+
+function get-thumbprint() {
+    govc about.cert -k -thumbprint | awk '{print $NF}'
+}
+
+# SCPs the component in $1 to the VCH, plops it in place, and brutally kills the previous running process
```
hickeng: This this comment is intended for `replace-component`

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,173 @@
...
+                                   --authorized-key=$VIC_KEY \
+                                   --thumbprint=$(get-thumbprint) \
+                 | grep -A1 "Published ports" | tail -n1 | awk '{print $NF}')
+}
+
+function replace-component() {
+    scp -oUserKnownHostsFile=/dev/null -oStrictHostKeyChecking=no $VIC_DIR/bin/$1 root@$VCH_IP:/tmp/$1 2>/dev/null
+    on-vch mv /sbin/$1 /tmp/old-$1
+    on-vch mv /tmp/$1 /sbin/$1
+    on-vch chmod 755 /sbin/$1
```
hickeng: suggest doing this first before the `mv /sbin/$1 /tmp/old-$1` - that way there's no window in which vic-init will try and fail to execute it.

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,173 @@
...
+    on-vch rm -f /tmp/old-$1
+}
+
+function replace-components () {
+    echo "Killing vic-init (and not taking no for an answer)..."
+    on-vch systemctl kill vic-init
```
hickeng: still wanting an answer as to why this is necessary - vic-init _should_ be restarting dead components.
If you still observe that not occurring with debug <= 1 please raise an issue for it with /var/log/vic/init.log attached that showcases the failure to restart.

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,173 @@
...
+BASE_DIR=$(dirname $(readlink -f "$BASH_SOURCE"))
+VIC_DIR=$(dirname $(readlink -f $BASE_DIR/..))
+
+# Run the command given on the VCH instead of locally
+function on-vch() {
+    ssh -oUserKnownHostsFile=/dev/null -oStrictHostKeyChecking=no -i $VIC_KEY root@$VCH_IP -C $@ 2>/dev/null
+}
+
+function get-thumbprint() {
+    govc about.cert -k -thumbprint | awk '{print $NF}'
```
hickeng: might want `pipefail` and or $? check for govc.

```
$ govc about.cert -k -thumbprint | awk '{print $NF}'
govc: dial tcp 192.168.78.128:443: getsockopt: no route to host
```

infra/scripts/replace-running-components.sh
gigawhitlocks: We don't want it to restart the dead components until they're replaced, or the replacement part of the script might fail. Hence we stop vic-init (SIGKILL so it doesn't reboot the VCH), replace necessary components, and then restart it

infra/scripts/replace-running-components.sh
gigawhitlocks: strictly speaking `govc ls`  is called first, but I think pipefail should just be on in the whole script, I was already thinking about that

gigawhitlocks:
@mhagen-vmware apparently you have dominion over `infra/scripts`

infra/scripts/replace-running-components.sh
```diff
@@ -137,37 +137,29 @@ function enable-debug () {
...
     fi
-    echo "Starting vic-init.."
-    on-vch systemctl start vic-init
+
+    for x in $services; do
+        echo "Replacing component $x..."
+        replace-component $x
+    done
+
+    # George swears you don't need to do this but if you don't, the new binaries are not launched
```
hickeng: George still swears you don't need to do this and wants logs from it not happening so we can track down WHY.
As written in this PR it's likely because the component processes haven't been killed after the binary is replaced.

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,169 @@
...
+function replace-component() {
+    scp -oUserKnownHostsFile=/dev/null -oStrictHostKeyChecking=no $VIC_DIR/bin/$1 root@$VCH_IP:/tmp/$1 2>/dev/null
+    pid=$(on-vch ps -e --format='pid,args' \
+                 | grep $1 | grep -v grep | awk '{print $1}')
+    on-vch chmod 755 /tmp/$1
+    on-vch mv /tmp/$1 /sbin/$1
+    if [[ $1 == "vic-init" ]]; then
+        on-vch systemctl restart vic-init
+    else
+        on-vch kill -9 $pid
```
hickeng: When #6943 can you remind me that it should update this to use `kill -TERM`. That will cause vSphere sessions to be cleaned up properly

infra/scripts/replace-running-components.sh
```diff
@@ -0,0 +1,169 @@
...
+
+# SCPs the component in $1 to the VCH, plops it in place, and brutally kills the previous running process
+function replace-component() {
+    scp -oUserKnownHostsFile=/dev/null -oStrictHostKeyChecking=no $VIC_DIR/bin/$1 root@$VCH_IP:/tmp/$1 2>/dev/null
+    pid=$(on-vch ps -e --format='pid,args' \
+                 | grep $1 | grep -v grep | awk '{print $1}')
+    on-vch chmod 755 /tmp/$1
+    on-vch mv /tmp/$1 /sbin/$1
+    if [[ $1 == "vic-init" ]]; then
+        on-vch systemctl restart vic-init
```
hickeng: We might need to look at the effect of this with #6943 - it's possible that this would cause all components to be shutdown as well which is not what we're after.

infra/scripts/replace-running-components.sh
gigawhitlocks: roger that

PR #7234, opened 2018-02-01
Quiet output on successful header check

gigawhitlocks:
`echo` is very slow. So is scrolling past a million lines of "OK". So is trying to find the one file you forgot to update that says FAIL in the middle of a million OKs. How about instead, we only print the failures, and make our builds a little faster and legible?


gigawhitlocks:
@mhagen-vmware you're the owner of this component apparently

PR #7236, opened 2018-02-02
Setup correct Syslog hooks for trace Logger

lcastellano:
Syslog hooks are currently not properly setup for the trace Logger. That causes the operation syslog messages to go lost.

Fixes 7169

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -70,22 +70,28 @@ func Init(cfg *LoggingConfig) error {
 
 			logrus.Debugf("log cfg: %+v", *cfg)
 
-			if cfg.Syslog != nil {
-				logrus.Debugf("syslog cfg: %+v", *cfg.Syslog)
-				hook, err := syslog.NewHook(
-					cfg.Syslog.Network,
-					cfg.Syslog.RAddr,
-					cfg.Syslog.Priority,
-					cfg.Syslog.Tag,
-				)
-				if err != nil {
-					// not a fatal error, so just log a warning
-					logrus.Warnf("error trying to initialize syslog: %s", err)
-				} else {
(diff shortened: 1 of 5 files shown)


pkg/trace/trace.go
```diff
@@ -66,6 +66,15 @@ func (t *Message) delta() time.Duration {
 	return time.Now().Sub(t.startTime)
 }
 
+// Add Syslog hook
+func InitLogger(cfg *log.LoggingConfig) error {
+	hook, err := log.GetSyslogHook(cfg)
+	if err == nil && hook != nil {
+		Logger.Hooks.Add(hook)
```
hickeng: I suspect we'll get away with this but if I recall correctly `Hooks.Add` is not thread safe.
Please confirm I'm correctly, that there's no risk of calling this method concurrently and note thread safety in the function comment.

mhagen-vmware:
this really should have some unit tests, but I know that we have an integration test already for this - thus the lgtm

pkg/log/log.go
```diff
@@ -70,22 +70,28 @@ func Init(cfg *LoggingConfig) error {
...
+			hook, err := GetSyslogHook(cfg)
+			if err == nil && hook != nil {
+				logrus.AddHook(hook)
 			}
 		})
 
 	return initializer.err
 }
+
+func GetSyslogHook(cfg *LoggingConfig) (logrus.Hook, error) {
```
hmahmood: Nit: `CreateSyslogHook` would be closer to what this function is doing.

anchal-agrawal:
@lcastellano The commit message itself still contains `...: issue 7169`. I've noticed that Github often uses the commit message (not the PR title) while merging. It'd be great if we updated the commit too.

Issue #6895, opened 2017-12-01
Fix archive unpack symlink handling - docker cp (online)


gigawhitlocks:
like #6863 but online copy
given the following scenario

running container -> shared volume A
stopped container -> shared volume A

run docker cp against the _stopped_ container which shares a volume with the running container using the same kind of symlink in the archive as in #6863 and observe that the file gets copied to the _running_ container

anchal-agrawal:
Not sure if this should be high or medium priority. Moving to Not Ready for now and adding assignees.

Issue #5933, opened 2017-08-07
Migrate Toolbox Archive from lib/tether to cmd/tether


morris-jason:
**Details:**
This story addresses toolbox debt from the merge #5865 for Docker CP.
 - Migrate Toolbox Archive Handler Overrides from lib/tether to cmd/tether. Adding them to lib/tether means that they override the default handlers in the applianceVM, but the current archive overrides are containerVM specific. 
 - On a similar note, migrate `toolbox.go:mountDiskLabel()` to ops_linux.go.

morris-jason:
Needs refactor of https://github.com/vmware/vic/pull/5934

PR #7262, opened 2018-02-05
Tar bug cp online case

gigawhitlocks:
[specific ci=1-44-Docker-CP-Online]
Resolves the online case of the `docker cp` bug as described in #6895 

At the time of posting the PR I need to write tests, but in my ad-hoc testing it looks like the Offline case was unaffected by my changes (YAY!) and the functionality is fully in place.

Looking forward to getting some comments and getting this issue resolved once and for all!

### DIFF
--- a/isos/bootstrap/bootstrap
+++ b/isos/bootstrap/bootstrap
@@ -41,7 +41,7 @@ if timeout --signal=KILL 2m mount -t ext4 /dev/disk/by-label/containerfs ${MOUNT
     # WARNING WARNING WARNING WARNING WARNING
     # if the tmpfs is not large enough odd hangs can occur and the ESX event log will
     # report the guest disabling the CPU
-    mount -t tmpfs -o size=64m tmpfs ${MOUNTPOINT}/.tether/
+    mount -t tmpfs -o size=128m tmpfs ${MOUNTPOINT}/.tether/
 
     # enable full system functionality in the container
     ln -s lib64 ${MOUNTPOINT}/.tether/lib
@@ -57,6 +57,7 @@ if timeout --signal=KILL 2m mount -t ext4 /dev/disk/by-label/containerfs ${MOUNT
     echo "prepping for switch to container filesystem"
 
     cp /bin/tether ${MOUNTPOINT}/.tether/tether
+    cp /bin/unpack ${MOUNTPOINT}/.tether/unpack
 
     echo 'tether tmpfs size before copying libraries: '
     df -k ${MOUNTPOINT}/.tether
--- a/isos/bootstrap/bootstrap
+++ b/isos/bootstrap/bootstrap
@@ -41,7 +41,7 @@ if timeout --signal=KILL 2m mount -t ext4 /dev/disk/by-label/containerfs ${MOUNT
     # WARNING WARNING WARNING WARNING WARNING
     # if the tmpfs is not large enough odd hangs can occur and the ESX event log will
     # report the guest disabling the CPU
-    mount -t tmpfs -o size=128m tmpfs ${MOUNTPOINT}/.tether/
+    mount -t tmpfs -o size=80m tmpfs ${MOUNTPOINT}/.tether/
 
     # enable full system functionality in the container
     ln -s lib64 ${MOUNTPOINT}/.tether/lib
--- a/lib/archive/unpack_unix.go
+++ b/lib/archive/unpack_unix.go
@@ -131,26 +131,50 @@ func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, r
 	return nil
 }
 
+func OfflineUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root string, binPath string) error {
+	d := &DoneChannel{Done: make(chan error)}
+
+	var cmd *exec.Cmd
+	var err error
+	if cmd, err = d.Unpack(op, tarStream, filter, root, binPath); err != nil {
+		return err
+	}
+
+	if err = cmd.Wait(); err != nil {
+		return err
+	}
+
+	if err = <-d.Done; err != nil {
+		return err
+	}
+
+	return nil
+}
+
+type DoneChannel struct {
+	Done chan (error)
+}
+
 // Unpack hooks into a binary present in the appliance vm called unpack in order to execute InvokeUnpack inside of a chroot. This method works identically to InvokeUnpack, except that it will not function in areas where the binary is not present at /bin/unpack
-func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root string, binPath string) error {
+func (d *DoneChannel) Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root string, binPath string) (*exec.Cmd, error) {
 
 	fi, err := os.Stat(root)
 	if err != nil {
 		// the target unpack path does not exist. We should not get here.
 		op.Errorf("tar unpack target does not exist: %s", root)
-		return err
+		return nil, err
 	}
 
 	if !fi.IsDir() {
 		err := fmt.Errorf("unpack root target is not a directory: %s", root)
 		op.Error(err)
-		return err
+		return nil, err
 	}
 
 	encodedFilter, err := EncodeFilterSpec(op, filter)
 	if err != nil {
 		op.Error(err)
-		return err
+		return nil, err
 	}
 
 	// Prepare to launch the binary, which will create a chroot at root and then invoke InvokeUnpack
@@ -164,17 +188,20 @@ func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root st
 
 	if err != nil {
 		op.Error(err)
-		return err
+		return cmd, err
 	}
 
 	if stdin == nil {
 		err = errors.New("stdin was nil")
 		op.Error(err)
-		return err
+		return cmd, err
 	}
-	done := make(chan error)
+
+	if d.Done == nil {
+		return nil, errors.New("initialize the Done channel in the receiver before calling Unpack")
+	}
+
 	go func() {
-		defer stdin.Close()
 		if tr, ok := tarStream.(*tar.Reader); ok {
 			tw := tar.NewWriter(stdin)
 			var th *tar.Header
@@ -183,19 +210,20 @@ func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root st
 				if err == io.EOF {
 					err = nil // this just signals the end of the stream, so we don't want to pass it out to the parent process, as it doesn't signal a problem
 					tw.Close()
-					break
+					d.Done <- err
+					return
 				}
 				if err != nil {
 					op.Errorf("error reading tar header %s", err)
-					done <- err
-					break
+					d.Done <- err
+					return
 				}
 				op.Debugf("processing tar header: asset(%s), size(%d)", th.Name, th.Size)
 				err = tw.WriteHeader(th)
 				if err != nil {
 					op.Errorf("error writing tar header %s", err)
-					done <- err
-					break
+					d.Done <- err
+					return
 				}
 				switch th.Typeflag {
 				case tar.TypeReg:
@@ -204,37 +232,22 @@ func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root st
 					op.Debugf("wrote %d bytes", k)
 					if err != nil {
 						op.Errorf("error writing file body bytes to stdin %s", err)
-						done <- err
-						break
+						d.Done <- err
+						return
 					}
 				}
 			}
-			done <- err
-			return
 		}
 		// if we're passed a stream that doesn't cast to a tar.Reader copy the tarStream to the binary via stdin; the binary will stream it to InvokeUnpack unchanged
+
 		if _, err := io.Copy(stdin, tarStream); err != nil {
 			op.Errorf("Error copying tarStream: %s", err.Error())
 		}
-		done <- err
+		d.Done <- err
 	}()
 
-	out, err := cmd.CombinedOutput()
-	if len(out) == 0 {
-		op.Debug("No output from command")
-	} else {
-		// output should just be trace messages
-		op.Debugf("%s", string(out))
+	if err = cmd.Start(); err != nil {
+		return nil, err
 	}
-
-	if err != nil {
-		stdin.Close()
-		op.Errorf("Command returned error %s", err.Error())
-		return err
-	}
-
-	// This error gets logged by the goroutine if it is non-nil.
-	// This receive is just functioning as a wait
-	err = <-done
-	return err
+	return cmd, err
 }
--- a/lib/archive/unpack_unix.go
+++ b/lib/archive/unpack_unix.go
@@ -132,7 +132,7 @@ func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, r
 }
 
 // Unpack hooks into a binary present in the appliance vm called unpack in order to execute InvokeUnpack inside of a chroot. This method works identically to InvokeUnpack, except that it will not function in areas where the binary is not present at /bin/unpack
-func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root string) error {
+func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root string, binPath string) error {
 
 	fi, err := os.Stat(root)
 	if err != nil {
@@ -155,7 +155,9 @@ func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root st
 
 	// Prepare to launch the binary, which will create a chroot at root and then invoke InvokeUnpack
 	// #nosec: Subprocess launching with variable. -- neither variable is user input & both are bounded inputs so this is fine
-	cmd := exec.Command("/bin/unpack", op.ID(), root, *encodedFilter)
+	// "/bin/unpack" on appliance
+	// "/.tether/unpack" inside container
+	cmd := exec.Command(binPath, op.ID(), root, *encodedFilter)
 
 	//stdin
 	stdin, err := cmd.StdinPipe()
--- a/lib/archive/unpack_unix.go
+++ b/lib/archive/unpack_unix.go
@@ -37,6 +37,12 @@ const (
 	// O_TRUNC = truncate file to 0 length if it does exist(overwrite the file)
 	// O_WRONLY = We use this since we do not intend to read, we only need to write.
 	fileWriteFlags = os.O_CREATE | os.O_TRUNC | os.O_WRONLY
+
+	// Location of the unpack binary inside of containers
+	ContainerBinaryPath = "/.tether/unpack"
+
+	// Location of the unpack binary inside the endpoint VM
+	ApplianceBinaryPath = "/bin/unpack"
 )
 
 // InvokeUnpack will unpack the given tarstream(if it is a tar stream) on the local filesystem based on the specified root
@@ -131,11 +137,12 @@ func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, r
 	return nil
 }
 
-func OfflineUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root string, binPath string) error {
+// OfflineUnpack wraps Unpack for usage in contexts without a childReaper.
+func OfflineUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root string) error {
 
 	var cmd *exec.Cmd
 	var err error
-	if cmd, err = Unpack(op, tarStream, filter, root, binPath); err != nil {
+	if cmd, err = Unpack(op, tarStream, filter, root, ApplianceBinaryPath); err != nil {
 		return err
 	}
 
@@ -146,7 +153,7 @@ func OfflineUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, 
 	return nil
 }
 
-// Unpack hooks into a binary present in the appliance vm called unpack in order to execute InvokeUnpack inside of a chroot. This method works identically to InvokeUnpack, except that it will not function in areas where the binary is not present at /bin/unpack
+// Unpack runs the binary compiled in cmd/unpack.go which creates a chroot at `root` and passes `op`, `tarStream`, and `filter` to InvokeUnpack for extraction of the tar on the filesystem. `binPath` should be either ApplianceBinaryPath or ContainerBinaryPath. Unpack returns a `Cmd` to allow use in conjunction with the tether's `LaunchUtility`, so it is necessary to call `cmd.Wait` after `Unpack` exits e.g. OfflineUnpack, if not being used in conjunction with LaunchUtility and the childReaper.
 func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root string, binPath string) (*exec.Cmd, error) {
 
 	fi, err := os.Stat(root)
--- a/lib/archive/unpack_unix.go
+++ b/lib/archive/unpack_unix.go
@@ -190,15 +190,12 @@ func streamCopy(op trace.Operation, stdin io.WriteCloser, tarStream io.Reader) {
 			op.Errorf("error writing tar header %s", err)
 			return
 		}
-		switch th.Typeflag {
-		case tar.TypeReg:
-			var k int64
-			k, err = io.Copy(tw, tr)
-			op.Debugf("wrote %d bytes", k)
-			if err != nil {
-				op.Errorf("error writing file body bytes to stdin %s", err)
-				return
-			}
+		var k int64
+		k, err = io.Copy(tw, tr)
+		op.Debugf("wrote %d bytes", k)
+		if err != nil {
+			op.Errorf("error writing file body bytes to stdin %s", err)
+			return
 		}
 	}
 }
--- a/lib/archive/unpack_unix.go
+++ b/lib/archive/unpack_unix.go
@@ -174,8 +174,34 @@ func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root st
 	}
 	done := make(chan error)
 	go func() {
-		// copy the tarStream to the binary via stdin; the binary will stream it to InvokeUnpack unchanged
 		defer stdin.Close()
+		if tr, ok := tarStream.(*tar.Reader); ok {
+			op.Errorf("XXX tarstream is a tar.reader, not an io.reader")
+			tw := tar.NewWriter(stdin)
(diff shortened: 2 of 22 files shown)


lib/tether/toolbox.go
```diff
@@ -293,7 +293,9 @@ func toolboxOverrideArchiveRead(u *url.URL, tr *tar.Reader) error {
 
 		// no need to join on u.Path here. u.Path == spec.Rebase, but
 		// Unpack will rebase tar headers for us. :thumbsup:
-		err = archive.InvokeUnpack(op, tr, spec, diskPath)
+		op.Debugf("Unpacking tar archive to %s", diskPath)
+		// err = archive.InvokeUnpack(op, tr, spec, diskPath)
```
gigawhitlocks: note to self: delete commented code

lib/archive/unpack_unix.go
```diff
@@ -172,8 +174,45 @@ func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root st
...
-		// copy the tarStream to the binary via stdin; the binary will stream it to InvokeUnpack unchanged
 		defer stdin.Close()
+		if tr, ok := tarStream.(*tar.Reader); ok {
+			tw := tar.NewWriter(stdin)
+			var th *tar.Header
+			for {
+				th, err = tr.Next()
+				if err == io.EOF {
+					err = nil // this just signals the end of the stream, so we don't want to pass it out to the parent process, as it doesn't signal a problem
+					tw.Close()
```
gigawhitlocks: also I probably want to `defer` this call on line 180

isos/bootstrap/bootstrap
```diff
@@ -39,7 +39,7 @@ if mount -t ext4 /dev/disk/by-label/containerfs ${MOUNTPOINT}; then
     # WARNING WARNING WARNING WARNING WARNING
     # if the tmpfs is not large enough odd hangs can occur and the ESX event log will
     # report the guest disabling the CPU
-    mount -t tmpfs -o size=64m tmpfs ${MOUNTPOINT}/.tether/
+    mount -t tmpfs -o size=128m tmpfs ${MOUNTPOINT}/.tether/
```
hickeng: can you check to see what size you actually need this to be, then set it to that plus maybe 10% headroom?
Use `df -h` in a containerVM and you can see the consumption

lib/tether/toolbox.go
```diff
@@ -293,7 +293,9 @@ func toolboxOverrideArchiveRead(u *url.URL, tr *tar.Reader) error {
...
+		err = archive.Unpack(op, tr, spec, diskPath, "/.tether/unpack")
```
hickeng: This should be in `cmd/tether` - the `lib/tether` code is built into `vic-init` as well.
This _might_ require a re-org of toolbox code in general - we never really defined how guest copy operations are expected to work on the endpointVM.

The only use-case I can think of for it currently is collecting the logs if vicadmin cannot be used for some reason. I cannot think of a solid use-case for putting files onto the endpointVM.

lib/archive/unpack_unix.go
```diff
@@ -172,8 +174,45 @@ func Unpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, root st
 	}
 	done := make(chan error)
 	go func() {
-		// copy the tarStream to the binary via stdin; the binary will stream it to InvokeUnpack unchanged
 		defer stdin.Close()
+		if tr, ok := tarStream.(*tar.Reader); ok {
```
hickeng: By what path do we get here? I suspect we're going `bytestream->tarstream->bytestream->tarstream` and we may have another option that is to suppress the first `bytestream->tarstream` translation.

gigawhitlocks: The following only applies to the online case.

The flow is tough to follow but the short answer is: we only actually copy the tarstream once, and it's tarstream->tarstream, over stdin.

The long answer is that the PortLayer sends the data to vSphere as SOAP data one way or the other (I didn't dig too deeply into how *that* works), and the govmomi code pulls it out of vSphere (again, didn't dig in because I didn't want to modify the toolbox) and expects a handling function that takes a `tar.Reader`, so by the time that my code gets the data in the toolbox, in `toolboxOverrideArchiveRead` down below, the data is already a tarstream. It's passed by reference to `Unpack`, so no copy there. Then it's streamed to the binary over stdin -- the binary calls `InvokeUnpack` with its `stdin` as the reference that's passed directly to `InvokeUnpack`, so you can think of it as streaming from `Unpack` to `InvokeUnpack` with a single copy and no data transformation to a generic bytestream for the Online case. `InvokeUnpack` is written to expect either a `tar.Reader` or a generic `io.Reader` because the Offline case produces the latter.

lib/tether/toolbox.go
gigawhitlocks: I can move this file over to `cmd`. I see the issue that you're pointing out.

isos/bootstrap/bootstrap
gigawhitlocks: Absolutely.

lib/tether/toolbox.go
gigawhitlocks: OK, looking into this, there are a lot of tangled dependencies, and maybe the refactor you're requesting (which is to move `toolboxOverrideArchiveRead` to `cmd`, I believe) will fully obfuscate this PR, because we at least need to move all of the container-related functions.. maybe open an issue so I can assign it to myself and handle that in a different PR @hickeng ?

gigawhitlocks:
Looking into why this change is causing the concurrent online copy test to fail

lib/archive/unpack_unix.go
gigawhitlocks: I DONT WANT TO DO THIS IT BREAKS ERRITHANG

isos/bootstrap/bootstrap
```diff
@@ -41,7 +41,7 @@ if timeout --signal=KILL 2m mount -t ext4 /dev/disk/by-label/containerfs ${MOUNT
     # WARNING WARNING WARNING WARNING WARNING
     # if the tmpfs is not large enough odd hangs can occur and the ESX event log will
     # report the guest disabling the CPU
-    mount -t tmpfs -o size=64m tmpfs ${MOUNTPOINT}/.tether/
+    mount -t tmpfs -o size=80m tmpfs ${MOUNTPOINT}/.tether/
```
vburenin: Is there any way to put something into the logs so devs can at least see a warning message?

lib/archive/unpack_unix.go
```diff
@@ -131,71 +137,104 @@ func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, r
...
+	// "/.tether/unpack" inside container
+	cmd := exec.Command(binPath, op.ID(), root, *encodedFilter)
 
-	//stdin
 	stdin, err := cmd.StdinPipe()
 
 	if err != nil {
 		op.Error(err)
-		return err
+		return cmd, err
```
vburenin: Does it make sense to return cmd here if you've got an error?

lib/archive/unpack_unix.go
```diff
@@ -37,6 +37,12 @@ const (
 	// O_TRUNC = truncate file to 0 length if it does exist(overwrite the file)
 	// O_WRONLY = We use this since we do not intend to read, we only need to write.
 	fileWriteFlags = os.O_CREATE | os.O_TRUNC | os.O_WRONLY
+
+	// Location of the unpack binary inside of containers
+	ContainerBinaryPath = "/.tether/unpack"
```
hmahmood: Do these need to be exported?

lib/archive/unpack_unix.go
```diff
@@ -131,71 +137,104 @@ func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, r
...
 	if err != nil {
 		op.Error(err)
-		return err
+		return nil, err
 	}
 
 	// Prepare to launch the binary, which will create a chroot at root and then invoke InvokeUnpack
 	// #nosec: Subprocess launching with variable. -- neither variable is user input & both are bounded inputs so this is fine
-	cmd := exec.Command("/bin/unpack", op.ID(), root, *encodedFilter)
+	// "/bin/unpack" on appliance
```
hmahmood: If it is only ever going to be these two paths, you should check if `binPath` is indeed one of these.

lib/archive/unpack_unix.go
```diff
@@ -131,71 +137,104 @@ func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, r
...
-		return err
+		return cmd, err
 	}
-	done := make(chan error)
+
 	go func() {
-		// copy the tarStream to the binary via stdin; the binary will stream it to InvokeUnpack unchanged
-		defer stdin.Close()
+		if tr, ok := tarStream.(*tar.Reader); ok {
+			tw := tar.NewWriter(stdin)
```
vburenin: This part of IF close should be a separate function for testability and readability purposes.

lib/archive/unpack_unix.go
```diff
@@ -131,71 +137,104 @@ func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, r
...
+			defer stdin.Close()
```
vburenin: Do you really want to close stdin only if it is a tar.Reader?

lib/archive/unpack_unix.go
```diff
@@ -131,71 +137,104 @@ func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, r
...
+			var th *tar.Header
+			for {
+				th, err = tr.Next()
+				if err == io.EOF {
+					err = nil // this just signals the end of the stream, so we don't want to pass it out to the parent process, as it doesn't signal a problem
```
vburenin: not sure how this nil assignment helps.

lib/archive/unpack_unix.go
```diff
@@ -131,71 +137,104 @@ func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, r
...
+				err = tw.WriteHeader(th)
+				if err != nil {
+					op.Errorf("error writing tar header %s", err)
+					return
+				}
+				switch th.Typeflag {
+				case tar.TypeReg:
+					var k int64
+					k, err = io.Copy(tw, tr)
+					op.Debugf("wrote %d bytes", k)
```
vburenin: are you sure you want to log this? it can be really excessive logging in case of a large content.

vburenin: nm, it is not Write call as I though about Copy

lib/archive/unpack_unix.go
```diff
@@ -131,71 +137,104 @@ func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, r
...
 		op.Error(err)
-		return err
+		return cmd, err
 	}
-	done := make(chan error)
+
 	go func() {
-		// copy the tarStream to the binary via stdin; the binary will stream it to InvokeUnpack unchanged
-		defer stdin.Close()
+		if tr, ok := tarStream.(*tar.Reader); ok {
```
vburenin: you definitely may want swap if/else end exit earlier, so code gets more readable here.
Something like:
```
if tr, ok := tarStream.(*tar.Reader); !ok {
    if _, err := io.Copy(stdin, tarStream); err != nil {
        op.Errorf("Error copying tarStream: %s", err.Error())
    }
  return
}
all other logic here.

File: lib/archive/unpack_unix.go [hunk 15]
```diff
@@ -131,71 +137,104 @@ func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, r
...
+					k, err = io.Copy(tw, tr)
+					op.Debugf("wrote %d bytes", k)
+					if err != nil {
+						op.Errorf("error writing file body bytes to stdin %s", err)
+						return
+					}
+				}
+			}
+		}
+		// if we're passed a stream that doesn't cast to a tar.Reader copy the tarStream to the binary via stdin; the binary will stream it to InvokeUnpack unchanged
```
vburenin: please make it two line comment

isos/bootstrap/bootstrap
gigawhitlocks: A warning message about what? the warning from the comment above this change?

lib/archive/unpack_unix.go
gigawhitlocks: Ah, this is vestigial. It is a no-op now and I will remove it.

lib/archive/unpack_unix.go
hmahmood: Why do we need special handling for tar files?

lib/archive/unpack_unix.go
```diff
@@ -131,71 +137,104 @@ func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, r
...
+					return
+				}
+				op.Debugf("processing tar header: asset(%s), size(%d)", th.Name, th.Size)
+				err = tw.WriteHeader(th)
+				if err != nil {
+					op.Errorf("error writing tar header %s", err)
+					return
+				}
+				switch th.Typeflag {
+				case tar.TypeReg:
```
hmahmood: What about other types?

lib/archive/unpack_unix.go
gigawhitlocks: Yeah that's a good point. The structure here is a reflection of the history of the function rather than that of sane design decisions. I'll flip these around, as well as break stuff out into its own function as you suggested in another comment

lib/archive/unpack_unix.go
gigawhitlocks: Your editor doesn't smart-wrap long lines ;) ?

lib/archive/unpack_unix.go
gigawhitlocks: Probably not, I wasn't entirely sure when I wrote this part. I'll defer to your expertise and return `nil, err`.

lib/archive/unpack_unix.go
gigawhitlocks: it gets explicitly closed when we see an EOF, further down (there's a `tw.Close()`), so sometimes this call will be redundant and throw an (uncaptured) error, but this is just to make sure that the stream closes regardless as to whether the tar routine below succeeds. That said, I should probably `defer tw.Close()` instead of `stdin.Close()`. The difference in this context, though, I believe to be fairly minor. I don't believe there's an effect on functionality, between the two.

lib/archive/unpack_unix.go
gigawhitlocks: Yeah

gigawhitlocks:
Excellent feedback thank you @vburenin

lib/archive/unpack_unix.go
gigawhitlocks: They don't need to be handled specially in this context

lib/archive/unpack_unix.go
gigawhitlocks: `io.Copy` doesn't work on `tar.Reader`s because there are EOFs inside the stream and `io.Copy` will hang forever.

lib/archive/unpack_unix.go
gigawhitlocks: Ah, as the code exists currently `ContainerBinaryPath` does, but I can think of a short refactor to address that.

lib/archive/unpack_unix.go
gigawhitlocks: I'm not going to add a runtime check for these because they're constants, so if these are incorrect all of the cp tests will fail. However, to prevent that, I can define an unexported type that will result in a compile-time error if any other value is used and also help guide developers to the appropriate options should they need to know these values.

lib/tether/interfaces.go
```diff
@@ -56,6 +56,23 @@ type Operations interface {
...
+// System is a very preliminary interface that provides extensions with a means by which to
+// perform system style operations.
+// This should be designed with intent to provide sufficient abstraction that vic-init, tether
+// can be run by vcsim. Odds are that a significant portion of the methods on this interface
+// will be defined in vcsim and this interface will simply embed the vcsim interface.
+// There may well be overlap with shared.Sys that needs to be resolved
+//
+// For now its sole purpose is to allow extensions access to the specific portions of
+// BaseOperations that they actively use.
+type System interface {
```
hmahmood: There is a System interface in lib/system, you likely can modify that for your use.

lib/archive/unpack_unix.go
```diff
@@ -131,71 +139,113 @@ func InvokeUnpack(op trace.Operation, tarStream io.Reader, filter *FilterSpec, r
...
+			return
+		}
+		op.Debugf("processing tar header: asset(%s), size(%d)", th.Name, th.Size)
+		err = tw.WriteHeader(th)
+		if err != nil {
+			op.Errorf("error writing tar header %s", err)
+			return
+		}
+		switch th.Typeflag {
+		case tar.TypeReg:
```
morris-jason: What about `TypeRegA`?

lib/archive/unpack_unix.go
hmahmood: The docs say that the other types _may_ not have data. Why not just copy over without checking for a type?

lib/tether/interfaces.go
gigawhitlocks: @hmahmood , @hickeng wrote this commit to expose this object from the tether package to tether extensions to give me access to the pidMap/childReaper so I'll let him comment as necessary but I suspect that whatever is in `lib/system` isn't what I want here.

lib/archive/unpack_unix.go
gigawhitlocks: Discussed on Slack, appears to be deprecated.

lib/archive/unpack_unix.go
hmahmood: [`Read`](https://golang.org/pkg/archive/tar/#Reader.Read) actually will return `(0, io.EOF)` for the other types, so I am not sure if any special handling is required here for `TypeReg`.

lib/archive/unpack_unix.go
matthewavery: Covering the case should not hurt. and might cover an edge case. I am curious if you have tried this by piping a tar file with directory entries and such directly into a docker cp? so `cat tar_with_dirs_and_files.tar | docker cp - <cid>:/some/path` since this looks to only be writing the TypeReg. It is true that the TypeDir ad such should not have a body that needs to be written so this is likely good. I just want to make sure if possible.

isos/bootstrap/bootstrap
hickeng: We could echo a line but we cannot determine _which_ command will hang so we cannot target the logging to the point of failure. If we cannot rely on devs to at least glance over the code comments close to where the problem occurs then I doubt a not-proximate log message is going to help.
Having a build time script that _calculates_ the necessary footprint would be my preference, then we can right size it.

lib/tether/interfaces.go
hickeng: @hmahmood yeah - I suspect there'll be some consolidation onto that eventually, but this is explicitly a subset of `Operations`. I've not removed these methods from `Operations` and used composition with this structure because I think this is all liable to change. As noted in the comment there's conceptual overlap with `shared.Sys` which is the `lib/system` that you mention. 

Basically I don't like the PR I submitted for @gigawhitlocks to expose LaunchUtility in a manner that's useful to him - it's layered on top of some primordial tech debt around extensions and how they interact with the environment. I want to revisit this but don't want to invest much time in polishing the current state.

lib/archive/unpack_unix.go
matthewavery: update: worked with @gigawhitlocks on this out of band. I am convinced that things are working as they should.

lib/tether/toolbox.go
hickeng: ok

lib/archive/unpack_unix.go
hickeng: @gigawhitlocks did the slack discussion cover @hmahmood 's point about just copying data regardless of type and leaning on `io.Copy` to return `0, io.EOF`?

gigawhitlocks: @matthewavery was under the impression that the case that @hmahmood was concerned about could be exercised by copying a tarball that contains a directory w/ files in it. I copied him this shell transcript of myself exercising that case (it is also exercised in the tests, although the tests do `docker cp foo.tar.gz container:/dest` instead of `cat foo.tar.gz | docker cp - container:/dest`

```
✘ ian@trantor  ~/g/s/g/v/vic   pull-tar-bug ⁝ S ? 
 ❯ docker -H sc-rdops-vm07-dhcp-206-183.eng.vmware.com:2376 --tlsverify --tlscacert="ian-vch/ca.pem" --tlscert="ian-vch/cert.pem" --tlskey="ian-vch/key.pem" volume create vol1                                                                                                           [14:20:55]
vol1

 ian@trantor  ~/g/s/g/v/vic   pull-tar-bug ⁝ S ? 
 ❯ docker -H sc-rdops-vm07-dhcp-206-183.eng.vmware.com:2376 --tlsverify --tlscacert="ian-vch/ca.pem" --tlscert="ian-vch/cert.pem" --tlskey="ian-vch/key.pem" run --name on -v vol1:/vol1 -itd busybox                                                                                     [14:21:10]
a4d8aadad482c07edc21d544c6aedc39d00b4cfe119e0e494bb05e4d9b02155e

>>> elapsed time 46s                                                                                                                                                                                                                                                                                 

 ian@trantor  ~/g/s/g/v/vic   pull-tar-bug ⁝ S ? 
 ❯                                                                                                                                                                                                                                                                                        [14:21:57]
>>> elapsed time 47s                                                                                                                                                                                                                                                                                 

 ian@trantor  ~/g/s/g/v/vic   pull-tar-bug ⁝ S ? 
 ❯ docker -H sc-rdops-vm07-dhcp-206-183.eng.vmware.com:2376 --tlsverify --tlscacert="ian-vch/ca.pem" --tlscert="ian-vch/cert.pem" --tlskey="ian-vch/key.pem" create --name off -v vol1:/vol1 -it busybox                                                                                  [14:22:19]
71aa580a25f6911ee5220e86d463cd436a2936fee36d854bd896f4ffe6ff6d0c
>>> elapsed time 18s                                                                                                                                                                                                                                                                                 

 ian@trantor  ~/g/s/g/v/vic   pull-tar-bug ⁝ S ? 
 ❯ cat good.tar.gz | docker -H sc-rdops-vm07-dhcp-206-183.eng.vmware.com:2376 --tlsverify --tlscacert="ian-vch/ca.pem" --tlscert="ian-vch/cert.pem" --tlskey="ian-vch/key.pem" cp - off:/vol1                                                                                             [14:23:00]
>>> elapsed time 33s                                                                                                                                                                                                                                                                                 

 ian@trantor  ~/g/s/g/v/vic   pull-tar-bug ⁝ S ? 
 ❯ docker -H sc-rdops-vm07-dhcp-206-183.eng.vmware.com:2376 --tlsverify --tlscacert="ian-vch/ca.pem" --tlscert="ian-vch/cert.pem" --tlskey="ian-vch/key.pem" exec on ls /vol1                                                                                                             [14:23:56]
good

 ian@trantor  ~/g/s/g/v/vic   pull-tar-bug ⁝ S ? 
 ❯ docker -H sc-rdops-vm07-dhcp-206-183.eng.vmware.com:2376 --tlsverify --tlscacert="ian-vch/ca.pem" --tlscert="ian-vch/cert.pem" --tlskey="ian-vch/key.pem" exec on ls /vol1/good                                                                                                        [14:24:01]
foo
good.tar.gz
hostd.log
secrets.yml
service-swagger-gen.log
swagger-gen.log
vic-machine.log

 ian@trantor  ~/g/s/g/v/vic   pull-tar-bug ⁝ S ? 
 ❯ docker -H sc-rdops-vm07-dhcp-206-183.eng.vmware.com:2376 --tlsverify --tlscacert="ian-vch/ca.pem" --tlscert="ian-vch/cert.pem" --tlskey="ian-vch/key.pem" exec on ls /vol1/good/foo                                                                                                    [14:24:10]
bar

 ian@trantor  ~/g/s/g/v/vic   pull-tar-bug ⁝ S ? 
 ❯ tar xvf good.tar.gz                                                                                                                                                                                                                                                                    [14:24:21]
good/
good/good.tar.gz
good/service-swagger-gen.log
good/secrets.yml
good/foo/
good/foo/bar
good/vic-machine.log
good/hostd.log
good/swagger-gen.log

 ian@trantor  ~/g/s/g/v/vic   pull-tar-bug ⁝ S ? 
 ❯ cat evil-image/archive.tar.gz | docker -H sc-rdops-vm07-dhcp-206-183.eng.vmware.com:2376 --tlsverify --tlscacert="ian-vch/ca.pem" --tlscert="ian-vch/cert.pem" --tlskey="ian-vch/key.pem" cp - off:/vol1                                                                               [14:24:43]
Error response from daemon: Server error from portlayer: error writing files to device vol1
```

gigawhitlocks: @hmahmood does this satisfy your concern about `TypeRegA`?

gigawhitlocks:
OK now just gotta wait for CI to be back and for @mhagen-vmware to approve the test and we're good to go!

gigawhitlocks:
https://github.com/vmware/vic/issues/5933 is the follow-on issue re: tech debt @hickeng ; already exists.

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
```diff
@@ -296,3 +296,18 @@ Sub volumes: copy from an offline container to host, src includes a shared vol w
     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} rm -f subVol_on
     Should Be Equal As Integers  ${rc}  0
     Should Not Contain  ${output}  Error
+
+Malformed tarball doesn't extract to wrong container from shared volume
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} run -itd --name one -v vol1:/vol1 -v vol2:/vol2 busybox /bin/sh
```
mhagen-vmware: ${busybox}

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
```diff
@@ -296,3 +296,18 @@ Sub volumes: copy from an offline container to host, src includes a shared vol w
...
+    Should Be Equal As Integers  ${rc}  0
+    Should Not Contain  ${output}  Error
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} create --name two -it -v vol2:/vol2 -v vol1:/vol1 busybox
```
mhagen-vmware: ${busybox}

gigawhitlocks: right right

Issue #7221, opened 2018-01-31
Refactor image proxy code from the persona


sflxn:
Refactor the image proxy out of the persona code to allow other portlayer clients to access the portlayer code without dealing with swagger house keeping.

Acceptance Criteria:

- [ ] All the swagger code refactored out into "proxy" code and moved out of lib/apiservers/engine/backends to lib/apiservers/enging/proxy
- [ ] Add ops tracing logging to all the proxy code
- [ ] Add unit tests for image proxy

PR #7268, opened 2018-02-07
Implement proxy backend for the image module.

lcastellano:
Only one method calls directly the portlayer: DeleteImage.
Also implemented unit test using generated mock for the proxy.
Since the unit test does not have a real port layer client, various types
of caches have been changed not to save their values in the KV Store
when running as part of a test.

Issue: #7221


lib/apiservers/engine/backends/image.go
```diff
@@ -41,6 +43,7 @@ import (
 	"github.com/vmware/vic/lib/metadata"
 	"github.com/vmware/vic/lib/portlayer/util"
 	"github.com/vmware/vic/pkg/trace"
+	optrace "github.com/vmware/vic/pkg/trace"
```
sflxn: You already included trace above.  You're including a new one and aliasing it to optrace.

sflxn:
So a comment about adding the ephermeral to the cache.  I think a more standard way to do this is not to add new data to the cache to let it know that it's not acting in "real" cache mode.  A better way would be to create interfaces for the cache.  Then pass in the interfaces to whatever needs to use them.  Then in the unit test code, we create a mock cache that follows the interface but doesn't really do any real caching.

lib/apiservers/engine/backends/image.go
```diff
@@ -24,6 +24,8 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 
+	"github.com/vmware/vic/lib/apiservers/engine/proxy"
```
vburenin: keep all imports vmware/vic related in one block.

lib/apiservers/engine/backends/image.go
```diff
@@ -74,6 +80,11 @@ func (i *Image) Exists(containerName string) bool {
 func (i *Image) ImageDelete(imageRef string, force, prune bool) ([]types.ImageDelete, error) {
 	defer trace.End(trace.Begin(imageRef))
 
+	// This API is called by the docker front end, create
+	// operation and child context here
+	op := optrace.NewOperation(context.Background(), "ImageDelete: %s", imageRef)
+	lCtx := context.WithValue(context.Background(), trace.OpTraceKey, op)
```
vburenin: name it just ctx please.

lib/apiservers/engine/backends/image_test.go
```diff
@@ -20,7 +20,19 @@ import (
 	"time"
 
 	"github.com/docker/docker/api/types/container"
+	"github.com/docker/docker/daemon/events"
 	v1 "github.com/docker/docker/image"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/require"
+
+	"github.com/vmware/vic/lib/imagec"
```
vburenin: regroup imports pls

lib/apiservers/engine/proxy/mocks/image_mock.go
```diff
@@ -0,0 +1,63 @@
...
+// Source: image.go
+
+// Package mocks is a generated GoMock package.
+package mocks
+
+import (
+	context "context"
+	reflect "reflect"
+
+	gomock "github.com/golang/mock/gomock"
```
vburenin: I am extremely happy to see gomock in VIC project, but I do not see it in vendor on my side. Did you add it locally?

lib/apiservers/engine/proxy/mocks/image_mock.go
```diff
@@ -0,0 +1,63 @@
+// Copyright 2017 VMware, Inc. All Rights Reserved.
```
vburenin: it is 2018 now ;)

lib/apiservers/engine/proxy/mocks/image_mock.go
vburenin: btw, since this is a generated code, which means you should provide a way to regenerate it.

lib/imagec/layer_cache.go
```diff
@@ -35,6 +35,9 @@ type LCache struct {
 
 	client *client.PortLayer
 	dirty  bool
+
+	// In memory only cache, used in unit tests
+	ephemeral bool
```
vburenin: same thing, keep it as isolated as possible.

PR #7290, opened 2018-02-08
Add basic integration tests for VCH inspect API

### LINKED ISSUE #6031: Implement end-to-end tests for the API handler for inspecting a VCH
Implement end-to-end tests for the API handler for inspecting a VCH (#6029)

AngieCris:
`[specific ci=23-04-VCH-Inspect --suite 23-02-VCH-List]`

Fixes #6031 

some basic robot tests added for API inspect.

### DIFF
--- a/tests/resources/Group23-VIC-Machine-Service-Util.robot
+++ b/tests/resources/Group23-VIC-Machine-Service-Util.robot
@@ -146,6 +146,12 @@ Property Should Not Be Empty
     ${actual}=  Run    echo '${OUTPUT}' | jq -r '${jq}'
     Should Not Be Empty    ${actual}
 
+Property Length Should Be
+    [Arguments]    ${jq}    ${expected}
+
+    ${actual}=  Run    echo '${OUTPUT}' | jq -r '${jq} | length'
+    Should Be Equal    ${actual}    ${expected}
+
 Verify VCH List Empty
     ${vchs}=  Run  echo '${OUTPUT}' | jq -r '.vchs[]'
     Log  ${vchs}
--- a/tests/resources/Group23-VIC-Machine-Service-Util.robot
+++ b/tests/resources/Group23-VIC-Machine-Service-Util.robot
@@ -152,6 +152,14 @@ Property Length Should Be
     ${actual}=  Run    echo '${OUTPUT}' | jq -r '${jq} | length'
     Should Be Equal    ${actual}    ${expected}
 
+
+Get Service Version String
+    ${rc}  ${output}=    Run And Return Rc And Output    curl -s -w "\n\%{http_code}\n" -X GET "http://127.0.0.1:${HTTP_PORT}/container/version"
+    ${output}    ${status}=    Split String From Right    ${output}    \n    1
+    Should Be Equal As Integers    ${rc}    0
+    Should Be Equal As Integers    ${status}    200
+    [Return]  ${output}
+
 Verify VCH List Empty
     ${vchs}=  Run  echo '${OUTPUT}' | jq -r '.vchs[]'
     Log  ${vchs}
--- a/tests/test-cases/Group23-VIC-Machine-Service/23-04-VCH-Inspect.robot
+++ b/tests/test-cases/Group23-VIC-Machine-Service/23-04-VCH-Inspect.robot
@@ -1,0 +1,153 @@
+# Copyright 2018 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#	http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License
+
+*** Settings ***
+Documentation     Test 23-04 - VCH Inspect
+Resource          ../../resources/Util.robot
+Resource          ../../resources/Group23-VIC-Machine-Service-Util.robot
+Suite Setup       Setup
+Suite Teardown    Teardown
+
+Default Tags
+
+*** Keywords ***
+# TODO [AngieCris]: Uncomment it after ops user is enabled on CI (after https://github.com/vmware/vic/pull/7892 merges)
+#Install VIC Appliance With Ops Credentials
+#    Install VIC Appliance To Test Server  certs=${true}  additional-args=--ops-user=%{VCH_OPS_USERNAME} --ops-password=%{VCH_OPS_PASSWORD} --ops-grant-perms --debug 1
+
+Setup
+    Start VIC Machine Server
+#    Install VIC Appliance With Ops Credentials
+    Install VIC Appliance To Test Server
+
+    ${id}=  Get VCH ID  %{VCH-NAME}
+    ${dc-id}=  Get Datacenter ID
+
+    Set Suite Variable  ${VCH-ID}  ${id}
+    Set Suite Variable  ${DC-ID}  ${dc-id}
+
+
+Teardown
+    Cleanup VIC Appliance On Test Server
+    Terminate All Processes    kill=True
+
+Inspect VCH
+    Get Path Under Target  vch/${VCH-ID}
+
+Inspect VCH Using Session
+    Get Path Under Target Using Session  vch/${VCH-ID}
+
+Inspect VCH Within Datacenter
+    Get Path Under Target    datacenter/${DC-ID}/vch/${VCH-ID}
+
+Inspect VCH Within Datacenter Using Session
+    Get Path Under Target Using Session    datacenter/${DC-ID}/vch/${VCH-ID}
+
+Verify VCH Inspect Output
+    # basic
+    Property Should Be Equal        .debug                                                1
+    Property Should Be Equal        .name                                                 %{VCH-NAME}
+
+    # networks
+    Property Should Be Equal        .network.bridge.ip_range                              172.16.0.0/12
+    Property Should Not Be Empty    .network.bridge.port_group
+    Property Should Be Equal        .network.public.nameservers[0]                        null
+    Property Should Not Be Empty    .network.public.port_group
+    Property Should Be Equal        .network.container[0].alias                           public
+    Property Should Not Be Empty    .network.container[0].port_grou
+
+    # cert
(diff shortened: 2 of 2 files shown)


tests/test-cases/Group23-VIC-Machine-Service/23-04-VCH-Inspect.robot
```diff
@@ -0,0 +1,145 @@
...
+
+Inspect VCH Within Datacenter
+    Get Path Under Target    datacenter/${DC-ID}/vch/${VCH-ID}
+
+Inspect VCH Within Datacenter Using Session
+    Get Path Under Target Using Session    datacenter/${DC-ID}/vch/${VCH-ID}
+
+Verify VCH Inspect Output
+    # debug
+    Property Should Be Equal        .debug  1
```
zjs: nit: this might be slightly more readable if you insert extra spaces between `.debug` and `1` so that `1` is vertically aligned with the other values, below

tests/test-cases/Group23-VIC-Machine-Service/23-04-VCH-Inspect.robot
```diff
@@ -0,0 +1,145 @@
...
+
+    # connection
+    Property Should Be Equal        .runtime.docker_host                        %{DOCKER_HOST}
+    Property Should Be Equal        .runtime.admin_portal                       %{VIC-ADMIN}
+    Property Should Be Equal        .runtime.power_state                        poweredOn
+    Property Should Contain         .runtime.upgrade_status                     Up to date
+
+    # version
+    ${version}=  Get Version String
+    Property Should Be Equal  .version  ${version}
```
zjs: nit: as above, I think this would be more readable if `.version` were vertically aligned with the other properties and `${version}` were vertically aligned with the other values

tests/test-cases/Group23-VIC-Machine-Service/23-04-VCH-Inspect.robot
```diff
@@ -0,0 +1,145 @@
...
+    # compute
+    Property Should Not Be Empty    .compute.resource.id
+
+    # storage
+    Property Should Be Equal        .storage.base_image_size.value              8000000
+    Property Should Be Equal        .storage.base_image_size.units              KB
+
+    Property Should Contain         .storage.image_stores[0]                    %{TEST_DATASTORE}
+    Property Should Contain         .storage.volume_stores[0].datastore         %{TEST_DATASTORE}/%{VCH-NAME}-VOL
+    Property Should Be Equal        .storage.volume_stores[0].label             default
```
zjs: Should we also be verifying that `.image_stores` and `.volume_stores` contain only a single element? (We verify that the first element of each is what we're expecting, but the test would still pass if there were other bogus values showing up in the response.)

Solving this elegantly may require adding a new `Property Should ...` keyword (perhaps along the lines of `Property Should Be Unary`, which we could use for each array we expect to have a single element, or perhaps along the lines of `Property Length Should Be`, which we could use for each array we expect to have a specific length).

tests/test-cases/Group23-VIC-Machine-Service/23-04-VCH-Inspect.robot
```diff
@@ -0,0 +1,145 @@
...
+    Get Path Under Target Using Session  vch/${VCH-ID}
+
+Inspect VCH Within Datacenter
+    Get Path Under Target    datacenter/${DC-ID}/vch/${VCH-ID}
+
+Inspect VCH Within Datacenter Using Session
+    Get Path Under Target Using Session    datacenter/${DC-ID}/vch/${VCH-ID}
+
+Verify VCH Inspect Output
+    # debug
```
zjs: Could we also verify the other top-level properties, like `.name`?

AngieCris:
@zjs Yes you're right, it's better to put the delete bug fixes in a separate pr. Idk what's the best way to do it but I could always manually create a new branch, cherry-pick commits to that branch, and manually revert back this branch...not sure if it will work...let me try and see

tests/test-cases/Group23-VIC-Machine-Service/23-04-VCH-Inspect.robot
AngieCris: ah, missed this one

hickeng:
@AngieCris 
> Re-installing VCH with existing volume stores accidentally does a volume store clean-up step. Disabled that by passing a cleanup=${false} to VIC appliance install

I recall us testing that volumes in existing volume stores still exist after a reinstall as part of the delete tests. This behaviour should have been caught there if it's a product code issues, or we need a keyword that _expects_ existing volume stores during VCH install.

The flip side of that is that the regular VCH install keyword should confirm that volume stores don't exist and error if they do - that way we're being explicit about our state expectations for the test.

AngieCris:
@hickeng 
> I recall us testing that volumes in existing volume stores still exist after a reinstall as part of the delete tests. 
Yes, we have a test case that re-installs VCH using existing volume stores, and then check after installation completes that the volume stores exist. But the re-install keyword does a cleanup on test server first, which could accidentally delete the existing volume stores. 
> This behaviour should have been caught there if it's a product code issues, or we need a keyword that expects existing volume stores during VCH install.
The reason why this behavior hasn't been caught on CI is that because CI checks drone status and skips cleanup if the test is running (but when I ran locally, it failed, and that's how I found out this issue).
But conceptually in the test code if we expect existing volume store, we should not do any cleanup step before re-installing.

But yes passing `cleanup=${false}` to the re-install keyword probably isn't the most optimal way to do it...it disables all the cleanup steps including networks, volume stores, etc.
We probably need another keyword in VCH-util that expects volume stores before installation, and have the old keyword errors out if there's existing volume store. 
The `cleanup=${false}` kind of gives similar functionality, but it's not used anywhere in the code. I wonder what's our original purpose of having this argument.

AngieCris:
Kind of blocked by this ticket: https://github.com/vmware/vic/issues/7321
Found a bug on API inspect code when working on this PR - inspect API doesn't check if all allowed IP according to the configured certificate. 
Need to figure out if we want to wait on the test after the bug is fixed and checked in...

zjs:
> Kind of blocked by this ticket: #7321

If that issue blocks this PR, we should go ahead and fix it. (It may only be P2 on its own merits, but if it's blocking a P1 we should treat it as a P1.)

mhagen-vmware:
cleanup=${false} was added for nightly tests, because some tests don't have VCH installed anymore at the end and resulted in a 5 minute hang trying to gather logs from a non-existent VCH.

tests/resources/Group23-VIC-Machine-Service-Util.robot
```diff
@@ -78,6 +78,11 @@ Delete Path Under Target
     Set Test Variable    ${OUTPUT}
     Set Test Variable    ${STATUS}
 
+Get Version String
```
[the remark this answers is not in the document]
zjs: The original intent was to put this resource file in the group directory, as that was the closest pattern I could find to namespacing.

I was asked to move the file to the resources directory, so I put the group name in the file name instead.

Generally, I'd strongly prefer to find a pattern to allow scoping of keywords over attempting to name each keyword verbosely, for all of the reasons we would avoid defining _everything_ at the global scope in the product code.

tests/test-cases/Group23-VIC-Machine-Service/23-04-VCH-Inspect.robot
```diff
@@ -0,0 +1,145 @@
...
+Resource          ../../resources/Group23-VIC-Machine-Service-Util.robot
+Suite Setup       Setup
+Suite Teardown    Teardown
+
+Default Tags
+
+*** Keywords ***
+Setup
+    Start VIC Machine Server
+    Install VIC Appliance To Test Server  certs=${true}  additional-args=--ops-user=%{TEST_USERNAME} --ops-password=%{TEST_PASSWORD} --debug 1
```
zjs: We may need to tag this keyword as secret (or move this one call into a separate "Secret Setup" keyword tagged secret) because we're using `%{TEST_PASSWORD}` (and don't want that logged).

tests/test-cases/Group23-VIC-Machine-Service/23-04-VCH-Inspect.robot
```diff
@@ -0,0 +1,149 @@
...
+Resource          ../../resources/Group23-VIC-Machine-Service-Util.robot
+Suite Setup       Setup
+Suite Teardown    Teardown
+
+Default Tags
+
+*** Keywords ***
+Secret Install VIC Appliance With Ops Credentials
+    [Tags]  secret
+    Install VIC Appliance To Test Server  certs=${true}  additional-args=--ops-user=%{TEST_USERNAME} --ops-password=%{TEST_PASSWORD} --debug 1
```
mhagen-vmware: this is just setting the ops user to administrator@vsphere.local... not sure that really is what we want to do, also wrapping this entire thing in [tags] secret is going to be a real problem when it comes to debugging.

AngieCris: > this is just setting the ops user to administrator@vsphere.local... not sure that really is what we want to do
in order to not duplicate with CLI's tests, here it only checks that the ops user configuration goes into VCH config (inspect output matches up with what we put in install parameter). So it doesn't really matter what ops user we set it to. 

> wrapping this entire thing in [tags] secret is going to be a real problem when it comes to debugging
Since we're setting ops user, I guess we don't want to expose `%{TEST_USERNAME}` and `%{TEST_PASSWORD}` in the log?

mhagen-vmware: @jzt is adding an ops user to our CI already, just use that and remove the tags secret from here.

tests/test-cases/Group23-VIC-Machine-Service/23-02-VCH-List.robot
```diff
@@ -120,3 +122,15 @@ Get VCH List Within Invalid Datacenter and Compute Resource
...
+    Cleanup VIC Appliance On Test Server
+
+    Get VCH List
+
+    Verify Return Code
+    Verify Status Ok
+
+    Verify VCH List Empty
+
+    [Teardown]  NONE
```
mhagen-vmware: I just removed this test here:
https://github.com/vmware/vic/pull/7879/files

Please do not add it back in.  It inherently cannot be run in our CI, it needs to be moved to manual tests

AngieCris: oh sorry I didn't see it. It got automatically added by git merge. will remove it now.

AngieCris: just removed

PR #7292, opened 2018-02-08
updating priority definitions to include features [skip ci]

mdubya66:
Clarifying the priority definitions for all features and non-bug issues. 

For 1.4.0 - The two Anchor Features are ROBO and the Quality Work (OKR). They are the only non-bug issues (based on the definition above) that should be p1. Wizard work would be considered p2. Photon OS 2.0 is p3. The wolfpack work should be considered p4 as it is not 1.4.0 at all. 

Only P0 and P1's (Both bug and feature) should be added to the 1.4.0 All releases tracking.


CONTRIBUTING.md
```diff
@@ -181,17 +181,23 @@ discussion.
...
-priority/p1 - Bugs that NEED to be fixed by the assigned phase of the current release.
+##### Features and non-bug issues
+Indicate the VMware priority of a feature or non-bug issue. Priority drives when a feature or non-bug issue will be resolved in a release. This field must be filled out by the issue owner and in most cases verified with the product owner.
 
-priority/p2 - Bugs that SHOULD be fixed by the assigned phase of the current release, time permitting.
-
-priority/p3 - Bugs that SHOULD be fixed by a given release, time permitting.
-
-priority/p4 - Bugs that SHOULD be fixed in a future (to be determined) release.
+- priority/p0 - An issue that is not a bug and is blocking meaningful testing. No feature should be p0. eg. builds are failing because the syslog server is
```
anchal-agrawal: We're missing something at the end of `because the syslog server is`

CONTRIBUTING.md
```diff
@@ -181,17 +181,23 @@ discussion.
...
+- priority/p1 - An issue or feature that is required for the next release, typically an anchor feature; a large feature that is the focus for the release and drives the release date.
+- prority/p2 - An issue or feature that is desired for the next release, typically a pebble;, a feature that has been approved for inclusion but is not considered the anchor feature.
```
anchal-agrawal: Stray comma in `pebble;,`

mdubya66:
For 1.4.0 - The two Anchor Features are ROBO and the Quality Work (OKR). They are the only non-bug issues (based on the definition above) that should be p1. Wizard work would be considered p2. Photon OS 2.0 is p3. The vKublet work should be considered p4 as it is not 1.4.0 at all. 

Only P0 and P1's (Both bug and feature) should be added to the 1.4.0 All releases tracking.

mdubya66:
Anchor's drive the release. pebbles do not. Anchor's should not have extraneous work in them. Same for pebbles. I'll consider a table but given the defintions for bugs and featuers are different enough. I'll keep it like this for now.

mdubya66:
I think it would be ok to have pebbles use 2, 3 and 4 priorities however based on critical path and am open to cleaning up that verbage. But Anchors should be P1.

mdubya66:
/me goes and looks at how to do tables in .md

mdubya66:
@zjs let me know if i've addressed your comments

PR #7298, opened 2018-02-09
Setup updated for Longevity Tests

lgayatri:
Fixes #
Updated Dockerfile.6.5 and Dockerfile.foundation to reflect new test environment.


tests/longevity-tests/Dockerfile.6.5
```diff
@@ -15,4 +15,4 @@
 FROM longevity-base
 ENV GOVC_URL=blinky.eng.vmware.com
 ENV TEST_URL_ARRAY=blinky.eng.vmware.com
```
mdubya66: Don't we want this off of blinky and onto some other testbed?

tests/longevity-tests/Dockerfile.6.5
```diff
@@ -15,4 +15,4 @@
...
-ENV STATIC_VCH_OPTIONS="--insecure-registry=vic-executor1.vcna.io --public-network-ip 10.17.109.7/24 --public-network-gateway 10.17.109.253 --dns-server 10.118.81.1"
+ENV STATIC_VCH_OPTIONS="--insecure-registry=vic-executor1.vcna.io --public-network-ip 10.197.37.210/23 --public-network-gateway 10.197.37.253 --dns-server 10.142.7.21"
```
mdubya66: I thought we wanted the executors to be on onecloud? Does this move them off?

tests/longevity-tests/Dockerfile.foundation
```diff
@@ -18,12 +18,12 @@ COPY secrets /go/src/github.com/vmware/vic
...
 ENV TEST_USERNAME=administrator@vsphere.local
 ENV TEST_TIMEOUT=3m
-ENV TEST_RESOURCE=cls
+ENV TEST_RESOURCE=Cluster
 ENV DOMAIN=
-ENV BRIDGE_NETWORK=bridge
-ENV PUBLIC_NETWORK=vm-network
+ENV BRIDGE_NETWORK=vxw-dvs-3444-virtualwire-1-sid-5000-VCH-LS-1
+ENV PUBLIC_NETWORK=DPortGroup
 ENV DOCKER_API_VERSION=1.23
```
mdubya66: Odd, not sure what this does but VIC 1.3.1 supports DOCKER_API_VERSION=1.25. https://vmware.github.io/vic-product/assets/files/html/1.3/vic_app_dev/container_operations.html

mdubya66: Also, can we consider getting log publication to GCP added to Longevity. That could be in a separate PR.

Final note, no need to run [Full CI] on this, this code is not executed at all in the CI. It could save some time :)

tests/longevity-tests/Dockerfile.6.5
mhagen-vmware: no, that is the static IP address that the VCH gets assigned nothing to do with executors.

tests/longevity-tests/Dockerfile.foundation
mhagen-vmware: yea, that has just been set that way for a while. We need a follow up issue to update that, I don't want to try and conflate multiple changes into a single PR.

mhagen-vmware: yea, different PR as you said.  I will get it done with the DOCKER_API update

tests/longevity-tests/Dockerfile.6.5
mhagen-vmware: yea, that needs to be changed in both 6.5 and 6.0 otherwise we are not using your infra at all @lgayatri   these URLs need to point to your 6.5 and 6.0 VC address

PR #7308, opened 2018-02-09
Add virtual-kubelet to the appliance ISO

lcastellano:
1) The **_Makefile_** has been changed to add a new target appliance-vkube.iso, the path of the virtual-kubelet binary must be specified in an environment variable otherwise it defaults to: ./virtual-kubelet. The environment variable is called: VIRTUAL_KUBELET_PATH. The content of that variable is passed down to the script **_isos/appliance.sh_**
2) the file **_iso/appliance.sh_** has been modified to retrieve the appropriate binary for virtual-kubelet. 

The changes to the iso/appliance.sh script have been developed to allow the specification of different binary files to the ISO, not specifically the virtual-kubelet binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,12 @@ GOVC ?= $(GOPATH)/bin/govc$(BIN_ARCH)
 GAS ?= $(GOPATH)/bin/gas$(BIN_ARCH)
 MISSPELL ?= $(GOPATH)/bin/misspell$(BIN_ARCH)
 
+ifeq ($(VIRTUAL_KUBELET_PATH),)
+	VIRTUAL_KUBELET := virtual-kubelet
+else
+	VIRTUAL_KUBELET := $(VIRTUAL_KUBELET_PATH)
+endif
+
 .PHONY: all tools clean test check distro \
 	goversion goimports gopath govet gofmt misspell gas golint \
 	isos tethers apiservers copyright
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -420,6 +428,12 @@ $(appliance): isos/appliance.sh isos/appliance/* isos/vicadmin/** $(vicadmin) $(
 	@echo building VCH appliance ISO
 	@$(TIME) $< -p $(appliance-staging) -b $(BIN)
 
+# main appliance target + kubelet - depends on all top level component targets
+$(appliance-vkube): isos/appliance.sh isos/appliance/* isos/vicadmin/** $(vicadmin) $(vic-init) $(portlayerapi) $(docker-engine-api) $(appliance-staging) $(archive)
+	@echo building VCH appliance ISO
+	@$(TIME) $< -p $(appliance-staging) -b $(BIN) -x $(VIRTUAL_KUBELET) -o appliance.vkube
```
hickeng: Use `$@` instead of `appliance.vkube` - that way you'll actually get conditional builds if tweaking names.

isos/appliance.sh
```diff
@@ -126,6 +138,39 @@ cp ${BIN}/vic-init $(rootfs_dir $PKGDIR)/sbin/vic-init
...
+	APPLIANCE_NAME=appliance.iso
+fi
+
+echo Appliance name: $APPLIANCE_NAME
+
+if [ -n "${XTRABINS}" ]; then
+    for XBIN in $XTRABINS; do
+        GS=$(echo ${XBIN} | grep '^gs://' | cat)
+        if [ -n "$GS" ]; then
+            XBIN_LATEST_BUILD="$(gsutil ls -l ${XBIN} | grep -v TOTAL | sort -k2 -r | (trap ' ' PIPE; head -1))"
```
hickeng: unless this is expected to change build by build it should be in staging along with all the other remote package dependencies.

Honestly it looks like you want to have a separate (or additional) staging step, and a separate final build step instead of trying to inline it into the current appliance.sh script.

lcastellano:
Moving to a feature branch

PR #7310, opened 2018-02-12
Change default fellows for gandalf

sgairo:
This makes default fellows @mhagen-vmware, @mdubya66 , and @hickeng .

### DIFF
--- a/cmd/gandalf/main.go
+++ b/cmd/gandalf/main.go
@@ -46,7 +46,7 @@ const (
 	DefaultRepo   = "vic"
 	DefaultBranch = "master"
 
-	DefaultFellows = "caglar10ur,mhagen,mwilliamson,fabio"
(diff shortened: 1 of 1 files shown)


cmd/gandalf/main.go
```diff
@@ -46,7 +46,7 @@ const (
 	DefaultRepo   = "vic"
 	DefaultBranch = "master"
 
-	DefaultFellows = "caglar10ur,mhagen,mwilliamson,fabio"
+	DefaultFellows = "mhagen,mwilliamson,ghicken"
```
anchal-agrawal: @mdharamadas1 We'll need to change this in the binary that runs in CI.

mdharamadas1: Yeah, thanks for the heads-up! I will update it after this is merged.

PR #7314, opened 2018-02-13
Add virtual-kubelet binary to VIC ISO

lcastellano:
1) The **_Makefile_** has been changed to add a new target appliance-vkube.iso, the path of the virtual-kubelet binary must be specified in an environment variable otherwise it defaults to: ./virtual-kubelet. The environment variable is called: VIRTUAL_KUBELET_PATH. The content of that variable is passed down to the script **_isos/appliance-xtra.sh_**
2) the file **_iso/appliance-xtra.sh_** has been added to retrieve the appropriate binary for virtual-kubelet. 

The new **iso/appliance-xtra.sh** script has been developed to allow the addition of different binary files to the ISO, not specifically the virtual-kubelet binary.


lcastellano:
PR contains one too many files, inherited from master.

PR #7315, opened 2018-02-13
Add virtual-kubelet binary to VIC ISO

lcastellano:
1) The **_Makefile_** has been changed to add a new target appliance-vkube.iso, the path of the virtual-kubelet binary must be specified in an environment variable otherwise it defaults to: ./virtual-kubelet. The environment variable is called: VIRTUAL_KUBELET_PATH. The content of that variable is passed down to the script **_isos/appliance-extra.sh_**
2) the file **_iso/appliance-extra.sh_** has been added to retrieve the appropriate binary for virtual-kubelet. 

The new **iso/appliance-xtra.sh** script has been developed to allow the addition of different binary files to the ISO, not specifically the virtual-kubelet binary.

### DIFF
--- a/isos/appliance-extra.sh
+++ b/isos/appliance-extra.sh
@@ -1,0 +1,171 @@
+#!/bin/bash
+# Copyright 2018 VMware, Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This file should eventually be merged back into the main VIC appliance
+# build process. Common code between the regular appliance build and the
+# the extra-binary appliance build should be identified before merging.
+
+# Build the appliance filesystem ontop of the base
+
+# exit on failure and configure debug, include util functions
+set -e && [ -n "$DEBUG" ] && set -x
+DIR=$(dirname $(readlink -f "$0"))
+. $DIR/base/utils.sh
+
+
+function usage() {
+echo "Usage: $0 -p staged-package(tgz) -b binary-dir -x binary-source -f binary-filename (inside the ISO) -o appliance-output-name" 1>&2
+exit 1
+}
+
+while getopts "p:b:x:f:o:" flag
+do
+    case $flag in
+
+        p)
+            # Required. Package name
+            PACKAGE="$OPTARG"
+            ;;
+
+        b)
+            # Required. Target for iso and source for components
+            BIN="$OPTARG"
+            ;;
+
+        x)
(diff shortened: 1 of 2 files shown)


isos/appliance-xtra.sh
```diff
@@ -0,0 +1,185 @@
...
+
+#################################################################
+# Above: arg parsing and setup
+# Below: the image authoring
+#################################################################
+
+echo RootFsDir $(rootfs_dir $PKGDIR)
+
+# sysctl
+cp ${DIR}/appliance/sysctl.conf $(rootfs_dir $PKGDIR)/etc/
```
zjs: Starting from here, this duplicates substantial non-trivial image authoring code from `appliance.sh`. We should have a single copy of this code so that we don't wind up with inconsistencies if they drift out of sync.

isos/appliance-xtra.sh
```diff
@@ -0,0 +1,185 @@
...
+
+# unpackage base package
+unpack $PACKAGE $PKGDIR
+
+#################################################################
+# Above: arg parsing and setup
+# Below: the image authoring
+#################################################################
+
+echo RootFsDir $(rootfs_dir $PKGDIR)
```
zjs: This also seems like debugging logic which should either be removed or cleaned up.

isos/appliance-xtra.sh
```diff
@@ -0,0 +1,185 @@
...
+cp ${BIN}/{docker-engine-server,port-layer-server,vicadmin} $(rootfs_dir $PKGDIR)/sbin/
+cp ${BIN}/unpack $(rootfs_dir $PKGDIR)/bin/
+
+if [ -n "${APPLIANCE_OUTNAME}" ]; then
+    APPLIANCE_NAME=$(basename ${APPLIANCE_OUTNAME})
+else
+    APPLIANCE_NAME=appliance-xtra.iso
+fi
+
+echo Appliance name: $APPLIANCE_NAME
```
zjs: More debugging logic.

isos/appliance-xtra.sh
```diff
@@ -0,0 +1,185 @@
...
+
+if [ -n "${XTRABIN}" ]; then
```
zjs: We've already checked that this is set.

(However, one way to deduplicate the logic here with the logic in `appliance.sh` is to leave this check here, remove the check above, change the default name to `appliance.iso`, and rename this whole file to `appliance.sh` so that there's a single script that optionally includes an extra binary.)

sflxn:
Why couldn't the new code in appliance_extra.sh have been included in appliance.sh instead?  Why create a new script file?  Maybe I didn't read it carefully, but it seems like that file is the same as appliance.sh with a few new arguments and code.

isos/appliance-extra.sh
```diff
@@ -0,0 +1,171 @@
...
+done
+
+shift $((OPTIND-1))
+
+# check there were no extra args and the required ones are set
+if [ ! -z "$*" -o -z "$PACKAGE" -o -z "${BIN}" ]; then
+    usage
+fi
+
+if [ -z "${EXTRABIN}" ] || [ -z "${EXTRABIN_FILENAME}" ] || [ -z "${APPLIANCE_OUTNAME}" ]; then
```
zjs: Is there a reason this is separate from the conditional above (and using a different style)? (Line 71)

PR #7324, opened 2018-02-14
Debug and dev workflow improvements

hickeng:
Adds basic vic-select shell function to set environment variables 
appropriately for the active bash-helpers.sh profile.
Allows for access to the pprof ports for vic-init and vicadmin remotely:
* 6060 - vic-init
* 6061 - vicadmin
* 6062 - personality
* 6063 - portlayer

These rules should be changed to be dynamic based on debug but not going to do that here, #7355 

`[specific ci=1-*-Docker-Network-*]`


vburenin:
nitpick in bash-helpers.sh script if directory ~/.vic doesn't exist at line 160 it may cause an error:
```
vmware/vic [master●] » . infra/scripts/bash-helpers.sh
infra/scripts/bash-helpers.sh:.:160: no such file or directory: /home/vlad/.vic
```

hickeng:
@vburenin your nitpick sent me down something of a longer path, but I think the resulting profiles are much easier to create.
They're not that well tested yet as I've only exercised the options have have available at home, which doesn't include the 3 different external networks (with non-conflicting IP ranges).
Has resulted in me raising #7337

.gitignore
```diff
@@ -21,6 +21,7 @@ package.list
 *certs.zip
 integration_logs*
 vmware-ovftool
+.vic.profiles
```
vburenin: if it is not very disrupting, could you also add ".godeps_cache"

isos/appliance/nat-setup
```diff
@@ -81,6 +81,8 @@ COMMIT
 -A INPUT -p tcp -m tcp --dport 2378 -j ACCEPT
 -A INPUT -p tcp -m tcp --dport 2380 -j ACCEPT
 
+-A INPUT -p tcp -m tcp --dport 6060 -j ACCEPT
```
vburenin: it actually would be awesome to have a comment for each line to explain for which service that port is being opened.

infra/scripts/bash-helpers.sh
```diff
@@ -13,32 +13,62 @@
...
 }
 
 vic-create () {
+    vicProfileTranscode
+
     base=$(pwd)
     (
         cd "$(vic-path)"/bin || return
-        "$(vic-path)"/bin/vic-machine-"$OS" create --target="$GOVC_URL" "${OPS_CREDS[@]}" --image-store="$IMAGE_STORE" --compute-resource="$COMPUTE" "${TLS[@]}" ${TLS_OPTS} --name="${VIC_NAME:-${USER}test}" "${MAPPED_NETWORKS[@]}" "${VOLUME_STORES[@]}" "${NETWORKS[@]}" ${IPADDR} ${TIMEOUT} --thumbprint="$THUMBPRINT" "$@"
+        "$(vic-path)"/bin/vic-machine-"$OS" create --target="$TARGET_URL" "${OPS_CREDS[@]}" --image-store="$IMAGE_STORE" --compute-resource="$COMPUTE" "${TLS[@]}" ${TLS_OPTS} --name="${VIC_NAME:-${USER}test}" "${MAPPED_NETWORKS[@]}" "${VOLUME_STORES[@]}" "${NETWORKS[@]}" ${IPADDR} ${TIMEOUT} --thumbprint="$THUMBPRINT" "$@"
```
vburenin: Would it make sense to set a debug level by default?

infra/scripts/bash-helpers.sh
```diff
@@ -49,44 +79,52 @@ vic-create () {
...
 
     echo "SSH to ${host}"
     sshpass -ppassword ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@"${host}"
 }
 
 vic-admin () {
-    out=$("$(vic-path)"/bin/vic-machine-"$OS" debug --target="$GOVC_URL" --compute-resource="$COMPUTE" --name="${VIC_NAME:-${USER}test}" --enable-ssh "$keyarg" --rootpw=password --thumbprint="$THUMBPRINT" "$@")
+    vicProfileTranscode
+
+    out=$("$(vic-path)"/bin/vic-machine-"$OS" debug --target="$TARGET_URL" --compute-resource="$COMPUTE" --name="${VIC_NAME:-${USER}test}" --enable-ssh "$keyarg" --rootpw=password --thumbprint="$THUMBPRINT" "$@")
```
vburenin: I see that you removed double quotation at some places and at some didn't. It is on purpose?

infra/scripts/bash-helpers.sh
```diff
@@ -97,53 +135,146 @@ addr-from-dockerhost () {
...
 
-    out=$("$(vic-path)"/bin/vic-machine-"$OS" debug --target="$GOVC_URL" --compute-resource="$COMPUTE" --name="${VIC_NAME:-${USER}test}" --enable-ssh "$keyarg" --rootpw=password --thumbprint="$THUMBPRINT" "$@")
+    out=$("$(vic-path)"/bin/vic-machine-"$OS" debug --target="$TARGET_URL" --compute-resource="$COMPUTE" --name="${VIC_NAME:-${USER}test}" --enable-ssh "$keyarg" --rootpw=password --thumbprint="$THUMBPRINT" "$@")
     host=$(echo "$out" | grep DOCKER_HOST | awk -F"DOCKER_HOST=" '{print $2}' | cut -d ":" -f1 | cut -d "=" -f2)
 
     sshpass -ppassword ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@"${host}" tail -f /var/log/vic/docker-personality.log
 }
 
+# turns the configuration variables into the ones expected by the vic-x functions
+vicProfileTranscode() {
```
vburenin: Could you rename it to: "initVicProfileVariables" or something better that could be more clear what it does by its name?
It was not initially clear to me what "vicProfileTranscode" does, so I was looking for this function specifically to understand what is going on.

isos/appliance/nat-setup
hickeng: Agreed - I don't think it can be done inline, but might be wrong. Will add it inline if I can, out side the block if not.

infra/scripts/bash-helpers.sh
hickeng: I don't - it's very simple to add refinement onto the end of the function, e.g:
```
vic-create --debug=2`
```

vburenin:
@hickeng I am am glad it went that way. It is a lot more human friendly now.

infra/scripts/bash-helpers.sh
hickeng: I don't see any quotes changed - that said I _do_ see that `$keyarg` is incorrectly quoted meaning you cannot have an empty authorized_keys file.

infra/scripts/bash-helpers.sh
hickeng: I'll update the comment to explain more - it's transcoding the variables from the profile definition to the ones the helper expects so that existing profiles in the old form will still work.

infra/scripts/bash-helpers.sh
hickeng: Ah, I see you mean the other functions. Yeah, none of them should have keyarg quoted.

infra/scripts/bash-helpers.sh
```diff
@@ -13,32 +13,62 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.!/bin/bash
```
gigawhitlocks: well first of all `!/bin/bash` is appended to the license there lol

gigawhitlocks:
One issue with these helpers is that they're dependent upon the user actually using Bash as their shell. When I source the script in my shell, `zsh`, I get errors because a few of the methods as Bash-specific:

```
 ian@trantor  ~/g/s/g/v/v/i/scripts   pprof-and-bash-helpers ⁝ ✱ ? 
 ❯ . ./bash-helpers.sh                                                                       [12:31:40]
dirname: missing operand
Try 'dirname --help' for more information.
./bash-helpers.sh:pwd:18: too many arguments
Loading profiles - a new profile selection must be made
```

Not sure how you want to approach this. What comes to mind is making the script more self-contained and less dependent upon writing to the environment, or I can just live with this and write a wrapper for those of us who use zsh. Let me know.

infra/scripts/sample-helper.profiles
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+# example entry, actived by typing "example"
```
gigawhitlocks: This comment is confusing because it does not match the example entry, which would be activated by running `my-vch`. Left me a little confused as to how to select a profile.

gigawhitlocks:
I also think there's a need for a 'getting started' style help output or comment or document to accompany these helpers, as otherwise you basically need to read the script to figure out how to use it, and I think that's hurting adoption among the team.

hickeng:
@gigawhitlocks I'm fine with generalizing them for other shells, but they were first written explicitly for bash (as is seen in the name) because I didn't want to get bogged down in basically lowest common denominator `sh`. If you can tweak them for `zsh`, I'll review your PR ;)

I've added really basic description and index of function to the README.md in that directory.

Issue #7251, opened 2018-02-02
5-25- OPS-User-Grant Nightly Test Failure: Failure to install VCH


sgairo:
2/1 vic engine nightly test failure on 6.0:
[Suites.Group5-Functional-Tests.5-25-OPS-User-Grant.zip](https://github.com/vmware/vic/files/1690853/Suites.Group5-Functional-Tests.5-25-OPS-User-Grant.zip)

```
time="2018-02-01T22:34:30Z" level=debug msg="VCH DNS servers: []" 
time="2018-02-01T22:34:30Z" level=debug msg="Unable to locate existing server certificate in cert path" 
time="2018-02-01T22:34:30Z" level=info msg="Generating CA certificate/key pair - private key in VCH-0-2630/ca-key.pem" 
time="2018-02-01T22:34:30Z" level=info msg="Generating server certificate/key pair - private key in VCH-0-2630/server-key.pem" 
time="2018-02-01T22:34:31Z" level=info msg="Generating client certificate/key pair - private key in VCH-0-2630/key.pem" 
time="2018-02-01T22:34:31Z" level=info msg="Generated browser friendly PFX client certificate - certificate in VCH-0-2630/cert.pfx" 
time="2018-02-01T22:34:31Z" level=debug msg="--target=https://10.160.194.207/datacenter1" 
    [ Message content over the limit has been removed. ]
time="2018-02-01T22:34:34Z" level=info msg="DRS check OK on:" 
time="2018-02-01T22:34:34Z" level=info msg="  \"/datacenter1/host/cls1\"" 
Feb  1 2018 22:34:34.200Z DEBUG URL: https://harbor.ci.drone.local/v2/
Feb  1 2018 22:34:34.333Z DEBUG URL: http://harbor.ci.drone.local/v2/
time="2018-02-01T22:34:34Z" level=warning msg="Unable to confirm insecure registry harbor.ci.drone.local is a valid registry at this time." 
time="2018-02-01T22:34:34Z" level=info msg="Insecure registries = harbor.ci.drone.local" 
time="2018-02-01T22:34:34Z" level=warning msg="Only one host can access all of the image/container/volume datastores. This may be a point of contention/performance degradation and HA/DRS may not work as intended." 
time="2018-02-01T22:34:34Z" level=error msg=-------------------- 
time="2018-02-01T22:34:34Z" level=error msg="Failed to validate operations credentials: ServerFaultCode: The operation is not supported on the object." 
[...]

PR #7328, opened 2018-02-14
Ignore not-supported result from RetrieveUserGroups in VC 6.0

lcastellano:
On VC 6.0, the API RetrieveUserGroups is not implemented. Return success and print warning message when running against 6.0. This PR also add the collection of the VC logs when the test 5-25-OPS-User-Grant test fails.

Fixes: #7251

### DIFF
--- a/pkg/vsphere/rbac/rbac.go
+++ b/pkg/vsphere/rbac/rbac.go
@@ -146,11 +146,12 @@ func (am *AuthzManager) PrincipalBelongsToGroup(ctx context.Context, group strin
 	}
 
 	results, err := methods.RetrieveUserGroups(ctx, am.client, &req)
+
 	// This is to work around a bug in vSphere, when AD is added to
 	// the identity source list, the API returns Object Not Found,
 	// In this case, we ignore the error and return false (BUG: 2037706)
-	if err != nil && isNotFoundError(ctx, err) {
-		op.Debugf("Received Not Found Error from PrincipalBelongsToGroup(), could not verify user %s is not a member of the Administrators group", am.Principal)
+	if err != nil && (isNotSupportedError(ctx, err) || isNotFoundError(ctx, err)) {
+		op.Debugf("Received Error (%s) from PrincipalBelongsToGroup(), could not verify user %s is not a member of the Administrators group", err.Error(), am.Principal)
 		op.Warnf("If ops-user (%s) belongs to the Administrators group, permissions on some resources might have been restricted", am.Principal)
 		return false, nil
 	}
@@ -391,6 +392,20 @@ func (am *AuthzManager) getRoleName(resource *Resource) string {
 	default:
 		return am.RolePrefix + resource.Role.Name
 	}
+}
+
+func isNotSupportedError(ctx context.Context, err error) bool {
+	op := trace.FromContext(ctx, "isNotSupportedError")
+
+	if soap.IsSoapFault(err) {
+		vimFault := soap.ToSoapFault(err).VimFault()
+		op.Debugf("Error type: %s", reflect.TypeOf(vimFault))
+
+		_, ok := soap.ToSoapFault(err).VimFault().(types.NotSupported)
+		return ok
+	}
+
+	return false
 }
 
 func isNotFoundError(ctx context.Context, err error) bool {
--- a/tests/manual-test-cases/Group5-Functional-Tests/5-25-OPS-User-Grant.robot
+++ b/tests/manual-test-cases/Group5-Functional-Tests/5-25-OPS-User-Grant.robot
@@ -17,8 +17,15 @@ Documentation  Test 5-25 - OPS-User-Grant
 Resource  ../../resources/Util.robot
 Suite Setup  Wait Until Keyword Succeeds  10x  10m  Ops User Create
 Suite Teardown  Run Keyword And Ignore Error  Nimbus Cleanup  ${list}
+Test Teardown  Run Keyword If Test Failed  Gather VC Logs
 
 *** Keywords ***
+
+Gather VC Logs
+    Log To Console  Collecting VC logs ..
+    Run Keyword And Ignore Error  Gather Logs From ESX Server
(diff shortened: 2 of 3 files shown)


pkg/vsphere/rbac/rbac.go
```diff
@@ -393,6 +394,20 @@ func (am *AuthzManager) getRoleName(resource *Resource) string {
 	}
 }
 
+func isNotSupportedError(ctx context.Context, err error) bool {
```
hickeng: Feels like this and isNotFound should roll into base govmomi somehow. @dougm?
Not needed for this PR unless already available in govmomi.

tests/manual-test-cases/Group5-Functional-Tests/5-25-OPS-User-Grant.robot
```diff
@@ -17,8 +17,16 @@ Documentation  Test 5-25 - OPS-User-Grant
 Resource  ../../resources/Util.robot
 Suite Setup  Wait Until Keyword Succeeds  10x  10m  Ops User Create
 Suite Teardown  Run Keyword And Ignore Error  Nimbus Cleanup  ${list}
+Test Teardown  Run Keyword And Ignore Error  Gather VC Logs
 
 *** Keywords ***
+
+Gather VC Logs
```
hickeng: This should go into one of the utility libraries.
I'm also unsure whether it will download the ESX log bundles or just VC - a keyword comment noting which would be good.

tests/manual-test-cases/Group5-Functional-Tests/5-25-OPS-User-Grant.robot
```diff
@@ -17,8 +17,16 @@ Documentation  Test 5-25 - OPS-User-Grant
...
+    Log To Console  Collecting VC logs ..
+    Run Keyword And Ignore Error  Run  govc logs.download
+    Run Keyword And Ignore Error  Run  mv vcsupport*.tgz vcsupport.tgz.log
```
hickeng: presume this is so that the bundle gets gathered by the test harness? We already gather the container-logs bundle from the run (which seem to be .zip) - is it as simple as updating a pattern somewhere to grab .tgz as well? @mhagen-vmware

tests/manual-test-cases/Group5-Functional-Tests/5-25-OPS-User-Grant.robot
```diff
@@ -17,8 +17,16 @@ Documentation  Test 5-25 - OPS-User-Grant
 Resource  ../../resources/Util.robot
 Suite Setup  Wait Until Keyword Succeeds  10x  10m  Ops User Create
 Suite Teardown  Run Keyword And Ignore Error  Nimbus Cleanup  ${list}
+Test Teardown  Run Keyword And Ignore Error  Gather VC Logs
```
hickeng: presume this is going to be removed after we have a bundle? Could we have it as a separate commit so it can just be reverted.

tests/manual-test-cases/Group5-Functional-Tests/5-25-OPS-User-Grant.robot
mhagen-vmware: Yep, the place for this would be https://github.com/vmware/vic/blob/master/tests/resources/VCH-Util.robot#L367
With an additional: `Run Keyword If  '%{HOST_TYPE}' == 'VC'  gather what logs you want`, however as you can see in that keyword already it pulls *specific* logs.  Otherwise, this govc logs.download would just add 5-15 minutes to every single test case which we probably don't want - esp as we are moving CI to VC which would hit that case every time as well. As we discussed in slack the current bundle inclusion pattern is .zip .log .debug... zip is very specific because that is the only archive extension that github allows attaching so if you are pulling the whole govc logs.download you really should just dump that archive into a zip archive I guess.

pkg/vsphere/rbac/rbac.go
dougm: govmomi probably should have some of these error helpers, but it doesn't other than `types.IsFileNotFound` that VIC already makes use of.

tests/manual-test-cases/Group5-Functional-Tests/5-25-OPS-User-Grant.robot
mhagen-vmware: If you just wanted to gather all the logs, we have that already here - although could be better named as it will work on either esx or vc:
https://github.com/vmware/vic/blob/master/tests/resources/Vsphere-Util.robot#L193

tests/manual-test-cases/Group5-Functional-Tests/5-25-OPS-User-Grant.robot
```diff
@@ -17,8 +17,16 @@ Documentation  Test 5-25 - OPS-User-Grant
...
 
 *** Keywords ***
+
+Gather VC Logs
+    Log To Console  Collecting VC logs ..
+    Run Keyword And Ignore Error  Gather Logs From ESX Server
+    Run Keyword And Ignore Error  Run  mv vcsupport*.tgz vcsupport.tgz.log
```
mhagen-vmware: Don't just move this to a different extension, please zip the file into a zip archive.

tests/manual-test-cases/Group5-Functional-Tests/5-25-OPS-User-Grant.robot
mhagen-vmware: Test Teardown  Run Keyword If Test Failed  Gather VC Logs

tests/manual-test-cases/Group5-Functional-Tests/5-25-OPS-User-Grant.robot
hickeng: @mhagen-vmware can we add .tgz as a recognized archive type? Seems a little ott to have a `.tar.gz.zip`

Issue #7249, opened 2018-02-02
21-1 Whitelist Nightly Test Failure: Cannot Connect to Docker daemon


sgairo:
2/1 Nightly VIC Engine Test failure:
[Suites.Group21-Registries.21-1-Whitelist.zip](https://github.com/vmware/vic/files/1690832/Suites.Group21-Registries.21-1-Whitelist.zip)

`Cannot connect to the Docker daemon at tcp://sc-rdops-vm01-dhcp-30-126.eng.vmware.com:2376. Is the docker daemon running?`

sgairo:
Feeling this is a fluke, will investigate further to confirm.

sgairo:
Potentially seen in 2/6 nightly 6.0 test run. No logs, but console output says every test failed as "Harbor doesn't come up properly!"

sgairo:
This is essentially a slow test, worth seeing ways to speed it up, otherwise will close.

zjs:
## Historical Data
```
* [PASS] VIC Engine Nightly Build - 6.5 - 2018-02-15-00-15-26 
* [PASS] VIC Engine Nightly Build - 6.0 - 2018-02-14-20-32-00 
* [PASS] VIC Engine Nightly Build - 6.5 - 2018-02-14-00-48-22 
* [PASS] VIC Engine Nightly Build - 6.0 - 2018-02-13-20-32-00 
* [FAIL] VIC Engine Nightly Build - 6.5 - 2018-02-13-00-52-33
   * No logs for this suite included in overall results .tar.gz, but all three tests failed with "Harbor failed to come up properly!"
   * The tests failed due to a suite setup failure. No details are available because the failing keyword, `Install Harbor To Test Server`, is tagged secret.
* [PASS] VIC Engine Nightly Build - 6.0 - 2018-02-12-20-32-00 
* [PASS] VIC Engine Nightly Build - 6.5 - 2018-02-12-00-41-35 
* [PASS] VIC Engine Nightly Build - 6.0 - 2018-02-11-20-32-00 
* [PASS] VIC Engine Nightly Build - 6.5 - 2018-02-11-00-10-15 
* [PASS] VIC Engine Nightly Build - 6.0 - 2018-02-10-20-32-00 
* [PASS] VIC Engine Nightly Build - 6.5 - 2018-02-09-23-21-06 
* [PASS] VIC Engine Nightly Build - 6.0 - 2018-02-09-20-32-00 
[...]

PR #7335, opened 2018-02-15
Refactor Install Harbor To Test Server keyword

zjs:
The secret tag on the `Install Harbor To Test Server` makes it difficult to investigate failures when they occur.

Only one out of 30+ lines actually uses secret information.

Refactor the keyword to extract the secret information into its own keyword, allowing the tag to be applied in a more focused way. This is similar to the pattern used by keywords in `VCH-Util`.

Towards #7249

### DIFF
--- a/tests/resources/Harbor-Util.robot
+++ b/tests/resources/Harbor-Util.robot
@@ -26,9 +26,14 @@ ${harbor_cert}  getcert
 ${ova_harbor_admin_password}  harbor-admin-passwd
 
 *** Keywords ***
-Install Harbor To Test Server
+Secret Install Harbor To Test Server
     [Tags]  secret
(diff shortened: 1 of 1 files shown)


tests/resources/Harbor-Util.robot
```diff
@@ -42,7 +47,7 @@ Install Harbor To Test Server
     ${rc}  ${output}=  Run Keyword If  '%{HOST_TYPE}' == 'VC'  Set Suite Variable  ${host}  @{URLs}[${IDX}]%{TEST_DATACENTER}/host/%{TEST_RESOURCE}
 
     Log To Console  \nDeploying ova...
-    ${out}=  Run  ovftool --noSSLVerify --acceptAllEulas --datastore=${datastore} --name=${name} --net:"Network 1"='${network}' --diskMode=thin --powerOn --X:waitForIp --X:injectOvfEnv --X:enableHiddenProperties --prop:root_pwd=${password} --prop:harbor_admin_password=${password} --prop:db_password=${db_password} --prop:auth_mode=db_auth --prop:verify_remote_cert=${verify} --prop:protocol=${protocol} ${HARBOR_VERSION}.ova 'vi://${user}:${password}@${host}'
+    ${out}=  Secret Install Harbor To Test Server  ${name}  ${protocol}  ${verify}  ${host}  ${datastore}  ${network}
```
mhagen-vmware: In order to actually get this output for debugging purposes, I would `Log  ${out}` before you hit the first should contain

PR #7340, opened 2018-02-19
Add ability to cache generated dependency.

### LINKED ISSUE #7339: Speed up build time for small changes
**User Statement:**

At the moment to rebuild a binary of VIC components can take longer time than it really should due to two things:

1. Dependency generation take around 8 seconds
2. Rebuilding binary from scratch also incur full binary rebuild which is costly.

On my laptop "make vic-machine" call takes 54 seconds from which 8 seconds is a cache generation and the rest is attempt to build 3 different vic-machine binaries. 

**Details:**
Most of the required time can be drastically reduced if we cache dependency generation and utilize Go 1.10 build cache that caches compiled packages and rebuild only those which have changed.

**Acceptance Criteria:**
Build time should be shorter at least 5 times. Such result is possible to achieve:

```
root@0f8631ba6c89:/go/src/github.com/vmware/vic# time make vic-machine
Generating dependency set for cmd/imagec/
Generating dependency set for cmd/vic-machine/
Generating dependency set for cmd/vicadmin/
Generating dependency set for cmd/port-layer-server/
Generating dependency set for cmd/vic-dns/
Generating dependency set for cmd/tether/
Generating dependency set for cmd/vic-ui/
Generating dependency set for cmd/vic-machine-server/
Generating dependency set for cmd/docker/
Generating dependency set for cmd/vic-init/
Generating dependency set for cmd/archive/
Generating dependency set for cmd/rpctool/
Generating dependency set for cmd/gandalf/
building vic-machine linux...
building vic-machine windows...
building vic-machine darwin...

real	0m5.449s
user	0m5.191s
sys	0m2.234s
```

vburenin:
This PR adds ability to cache discovered Go dependencies that are used by 'make' tool to build changed targets. So that, if you are not changing project structure and only working withing existing set of packages regeneration such dependencies is redundant and adds 7-30 seconds (depends on IO speed) each time you are trying to run 'make <target>'.

With this PR you can cut 7-30 seconds for each 'make <target>' call. To make that possible just define an environment variable as the following:
```
export VIC_CACHE_DEPS=1
```

At some point if you add an additional package and need to regenerate dependencies, just run "make cleandeps", so next "make" run will regenerate all dependencies from scratch.

Fixes #7339

### DIFF
--- a/infra/scripts/go-deps.sh
+++ b/infra/scripts/go-deps.sh
@@ -21,19 +21,32 @@
 #     pkg       This is github.com/vmware/vic/cmd/imagec for example
 #
 
+cache_dir=.godeps_cache
+
 pkg=$1
 flags=$2
+cachedname=`echo .$1.godeps_cache | sed 's/\//_/g'`
 
 if [ -d "$pkg" ]; then
-    if [[ "$flags" == *d* ]]
-    then
+
+    if [[ "$flags" == *d* ]]; then
         # Only output if make is given the '-d' flag
         echo "Generating deps for $pkg" >&2
     fi
 
-    go list -f '{{join .Deps "\n"}}' github.com/vmware/vic/"$pkg" 2>/dev/null | \
-        xargs go list -f '{{if not .Standard}}{{.ImportPath}}{{end}}' 2>/dev/null | \
-        sed -e 's:github.com/vmware/vic/\(.*\)$:\1/*:'
+    if [ -n "$VIC_CACHE_DEPS" ]; then
+        mkdir -p $cache_dir
+        if [ ! -f $cache_dir/$cachedname ]; then
+            go list -f '{{join .Deps "\n"}}' github.com/vmware/vic/"$pkg" 2>/dev/null | \
+                xargs go list -f '{{if not .Standard}}{{.ImportPath}}{{end}}' 2>/dev/null | \
+                sed -e 's:github.com/vmware/vic/\(.*\)$:\1/*:' > "$cache_dir/$cachedname"
+        fi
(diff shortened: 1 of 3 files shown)


infra/scripts/go-deps.sh
```diff
@@ -21,19 +21,32 @@
...
+
+    if [[ "$flags" == *d* ]]; then
         # Only output if make is given the '-d' flag
         echo "Generating deps for $pkg" >&2
     fi
 
-    go list -f '{{join .Deps "\n"}}' github.com/vmware/vic/"$pkg" 2>/dev/null | \
-        xargs go list -f '{{if not .Standard}}{{.ImportPath}}{{end}}' 2>/dev/null | \
-        sed -e 's:github.com/vmware/vic/\(.*\)$:\1/*:'
+    if [ -n "$VIC_CACHE_DEPS" ]; then
```
anchal-agrawal: Can we note the `export VIC_CACHE_DEPS=1` requirement somewhere in this file? It's only mentioned in the commit message, so having it noted here will help.

gigawhitlocks:
Could we get some docs? <3

vburenin:
@gigawhitlocks I think it is very important to have docs updates As well as I am urging to move forward to Go1.10 that should significantly speed up components re-build time. 2-5 seconds instead of 20-45 seconds.

gigawhitlocks:
LGTM

I mostly just mean please copy the description of the feature that you put in the PR into, maybe README.md or something like that so that the rest of us don't forget the name of the flag we need to set and never get to take advantage of your feature!

vburenin:
@gigawhitlocks done

Issue #7337, opened 2018-02-17
ops-grant-perms fails if nfs datastore specified


hickeng:
Installing with the following options combined results in an error:
```
--ops-grant-perms
--volume-store=nfs://192.168.68.3/vic-volumes:nas'
```
Result:
```
Cannot init ops-user permissions, failure: Ops-User: RBACManager, Unable to find Datastores: datastore '192.168.68.3' not found. Exiting...
```

@lcastellano

PR #7346, opened 2018-02-21
Skip non vSphere managed datastores when granting perms

lcastellano:
Fix: #7337

Avoid locating datastores when they are not managed by vSphere.


lib/install/opsuser/opsuser.go
```diff
@@ -217,6 +217,10 @@ func (mgr *RBACManager) collectDatastores(ctx context.Context, finder *find.Find
 	}
 	volumeLocations := make([]url.URL, 0, len(mgr.configSpec.Storage.VolumeLocations))
 	for _, volumeLocation := range mgr.configSpec.Storage.VolumeLocations {
+		// Only apply changes to datastores managed by vSphere
+		if volumeLocation.Scheme != "ds" {
```
hickeng: We should move the "ds" scheme into govmomi as a constant and transition vic as a whole to using the govmomi DatastorePath structure.
Not needed for this PR however.

PR #7383, opened 2018-02-23
WIP:Move version

vburenin:
Move version related constants to the common place.

Fixes: #7384

### DIFF
--- a/cmd/tether/attach.go
+++ b/cmd/tether/attach.go
@@ -585,7 +585,7 @@ func (t *attachServerSSH) globalMux(in <-chan *ssh.Request, cleanup func()) {
 			payload = msg.Marshal()
 		case msgs.VersionReq:
 			msg := msgs.VersionMsg{
-				Version: feature.MaxPluginVersion - 1,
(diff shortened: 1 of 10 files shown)


vburenin:
@anchal-agrawal That would be true if version package was not referencing lib/migration/feature. So, it is global. Version package should not depend on anything else outside it, since its only purpose to keep a version of the software.

hickeng:
@vburenin I agree with you that versions responsibility is to keep the version of the software. The fact that it's referencing `lib/migration` is an issue, but I don't think the correct solution is to move the migration logic into the version package.

Can this be apporached the other way around and remove the reference to migration from the version package? I hate that fact that build number is used at all in the migration logic - this was viable when we were committed to a single branch model but having been forced off that by corporate the build number is no longer a definitive reference for "direction" when it comes to structure alterations.

zjs:
> Look at removing coupling between build version/build number and migration instead of combining them.

It seems like this requires essentially restarting this work. Closing this PR for now as it will still be discoverable via the issue.

PR #7386, opened 2018-02-23
Update plugin secrets

mdharamadas1:


.drone.yml
```diff
@@ -287,15 +287,14 @@ pipeline:
...
     files:
       - .cover/cover.out
 
   notify-slack-on-fail:
     image: plugins/slack
     secrets:
-      - slack_url
-    webhook: $SLACK_URL
+      - source: slack_url
+        target: slack_webhook
```
mhagen-vmware: this needs leading dash?

mdharamadas1: Nope

PR #7389, opened 2018-02-26
make push uses pki identity file

hickeng:
Updates the make push logic so that it can use an identity file if
one is available and will not prompt if a sane default is available.

That said it may make more sense to actively generate a keypair
every time then delete them when done - this negates the reliance
on existing keys and/or the use of a key agent.

`[skip ci]`


infra/scripts/replace-running-components.sh
```diff
@@ -99,7 +99,7 @@ function check-name-isnt-ambiguous ()  {
 # Translates VCH name to ID if necessary
 function get-vic-id () {
     if [[ -z $VIC_ID ]]; then
-        export VIC_ID="$($VIC_DIR/bin/vic-machine-linux ls --target=$target --user=$username --password=$password --thumbprint=$(get-thumbprint) | grep $VIC_NAME | awk '{print $1}')"
```
gigawhitlocks: thank you so much for all the places where you put quotes around variables like on this line -- this is something I should've done and know better etc.

gigawhitlocks:
These changes still LGTM after additional commits :) :+1:

Issue #7392, opened 2018-02-26
Organise storage portlayer files for consistent structure and naming


hickeng:
**Story**
As a VIC developer I want to be able to quickly and easily locate a file/function implementation across interface boundaries

**Details**
The files in `lib/portlayer/storage` do not have an organisation consistent with the rest of the portlayer and are only somewhat consistent within the component. This results in a navigation challenge when not intimately familiar with the component as the layered interfaces prevent easy navigation via `guru`.

There are additional refactor items that should be addressed but this is purely file/function location changes.

Issue #4776, opened 2017-04-13
Consistent Join/Bind portlayer API for image disks


hickeng:
**Story**
As a VIC developer I want a consistent means and usage for composing the elements of a container, meaning the Join, Bind, Unbind, Remove flow.

**Detail**
Currently the image dist for a containerVM is a special case - it's added by logic in the specs/guests packages based on various assumptions. This makes it significantly difficult to do things such as:
* read/only root filesystem
* non-persistent root filesystem
* multiple image stores

and any other facets that deal with how the image VMDK is used and where it's found.

**Acceptance**
storage portlayer supplies the Join, Bind, Unbind, Remove calls for images in the same style as all the other elements.

renmaosheng:
let's triage with James on this Epic for VIC 1.5.

PR #7397, opened 2018-02-26
Organise portlayer/storage and add Image.Join semantic

hickeng:
**For reviewers**
I recommend starting at [this part of the commit](https://github.com/vmware/vic/pull/7397/files#diff-5ef7d3491c894b4d1078ddad94017772R15) because that's package documentation for the resulting package structure.
This code is used primarily by `docker cp`, `docker diff`, and `docker export`.

**Details**
This combines two major pieces of refactoring:

1. file level organisation in the `lib/portlayer/storage` packages to make them more consistent
2. normalizes the image logic to use the standard Join pattern
    * this still has Join (make available) and Bind (make active) into the Join call for now to limit the scope of the change.

This work was done for various reasons:

1. To bring order to the force - I was finding myself unable to make reasoned guesses about where to look for specific storage code. This does not address everything but significantly improves the high level structure for the various `store` (e.g. container store, image store, volume store) implementations.
2. As necessary foundation work for the podVM prototype in the `feature/wolfpack` branch. This branch required being able to Join multiple images to a containerVM. The podVM prototype also changes the way in which the images are identified in the VM (switching to disk UUID for class of disk and label for identifier) but that portion was dropped from this PR because it's not currently required and causes upgrade complications
3. Addresses some of the most egregious  tech debt about how image disk is merged into the containerVM VirtualMachineConfigSpec (#7393).

This _does not_ address some of the large swathes of tech debt that it otherwise touches on:
* spec generation - really, really need another OS in order to clarify spec usage and interface
* storage portlayer
  * the use of three different Path formats that are not easily convertible, differentiated, or inter-operable makes changing this component horrific. The addition of Image.DatastorePath field is a stop-gap because reworking the entire component to use semantically consistent return values ballooned way past what is feasible here.
  * WriteImage - this is the means by which images are written into the image store. It's not clear whether we should try to rephrase this in terms of an Importer, but it is very close to the current implementations. It would need custom logic to handle the unpacking of image layers with their masking mechanic.
  * CRUD lifecycle of store entities - these are all custom per implementation currently.

Towards #4776
Fixes: #7392 #7393

TODO:
- [x] separate the join/bind semantic from the uuid update so we can have the organisational refactor without introducing upgrade implications.
- [x] test "image not found" errors - these have moved from containerCreate to imageJoin call.
```
[full ci]
[parallel jobs=4]
```


lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -350,6 +354,37 @@ func (h *StorageHandlersImpl) WriteImage(params storage.WriteImageParams) middle
...
 }
 
+//ImageJoin modifies the config spec of a container to include the specified image
+func (h *StorageHandlersImpl) ImageJoin(params storage.ImageJoinParams) middleware.Responder {
+	op := trace.NewOperation(context.Background(), "ImageJoin %s", params.ID)
+	defer trace.End(trace.Begin("", op))
+
+	handle := epl.HandleFromInterface(params.Config.Handle)
+	if handle == nil {
+		err := &models.Error{Message: "Failed to get the Handle"}
```
vburenin: Maybe better to phrase it like "Handle not found"?

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -350,6 +354,37 @@ func (h *StorageHandlersImpl) WriteImage(params storage.WriteImageParams) middle
...
+	}
+
+	storeURL, _ := util.ImageStoreNameToURL(params.StoreName)
+	img, err := h.imageCache.GetImage(op, storeURL, params.ID)
+	if err != nil {
+		op.Errorf("Volumes: StorageHandler : %#v", err)
+		return storage.NewImageJoinNotFound().WithPayload(&models.Error{Code: http.StatusNotFound, Message: err.Error()})
+	}
+
+	handleprime, err := image.Join(op, handle, params.Config.DeltaID, params.Config.ImageID, params.Config.RepoName, img)
```
vburenin: for readability purposes maybe better to do this?:
```
cfg := params.Config
handleprime, err := image.Join(op, handle, cfg.DeltaID, cfg.ImageID, cfg.RepoName, img)
```

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
hickeng: I've left it consistent with the `task.Join` call (task_handles.go:53) - not going to debate that the message could be better, but trying to limit the scope of the change.

pkg/fs/ext4.go
```diff
@@ -72,8 +75,13 @@ func (e *Ext4) Unmount(op trace.Operation, path string) error {
 func (e *Ext4) SetLabel(op trace.Operation, devPath, labelName string) error {
 	defer trace.End(trace.Begin(devPath))
 
+	// Warn if truncating label
```
vburenin: Can it potentially lead to a collision? Can we enforce a specific length?

lib/portlayer/storage/volume/vsphere/store.go
```diff
@@ -25,13 +25,26 @@ import (
...
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/datastore"
 	"github.com/vmware/vic/pkg/vsphere/disk"
 	"github.com/vmware/vic/pkg/vsphere/session"
 )
 
+const (
+	// TODO: this was shared with image store hence the disjoint naming. Should be updated
+	// but migration/upgrade implications are unclear
+	metaDataDir = "imageMetadata"
```
vburenin: it seems like it might be a useful constant. Should not it be moved to lib/constant?

lib/portlayer/storage/volume/cache/cache.go
```diff
@@ -35,17 +37,17 @@ type VolumeLookupCache struct {
...
 	// Maps the service url of the volume store to the underlying data storage implementation
-	volumeStores map[string]VolumeStorer
+	volumeStores map[string]volume.VolumeStorer
 }
 
 func NewVolumeLookupCache(op trace.Operation) *VolumeLookupCache {
 	v := &VolumeLookupCache{
-		vlc:          make(map[string]Volume),
-		volumeStores: make(map[string]VolumeStorer),
+		vlc:          make(map[string]volume.Volume),
```
vburenin: Is there a reason why we use Volume as a value, not as a pointer? Volume is a relatively massive data structure, copying it each time is not really desirable.

lib/portlayer/storage/image/vsphere/store.go
```diff
@@ -68,8 +71,10 @@ var (
 )
 
 const (
-	StorageImageDir     = "images"
-	defaultDiskLabel    = "containerfs"
+	StorageImageDir = "images"
```
vburenin: lib/constants?

lib/portlayer/storage/container/store.go
```diff
@@ -0,0 +1,90 @@
...
+func NewContainerStore(op trace.Operation, s *session.Session, imageResolver storage.Resolver) (*ContainerStore, error) {
+	dm, err := disk.NewDiskManager(op, s, storage.Config.ContainerView)
+	if err != nil {
+		return nil, err
+	}
+
+	cs := &ContainerStore{
+		Vmdk: disk.Vmdk{
+			Manager: dm,
+			//ds: ds,
```
vburenin: is something is missing here?

lib/apiservers/portlayer/swagger.json
```diff
@@ -593,6 +593,68 @@
...
 			},
+			"post": {
+				"description": "Join an image to a container",
+				"operationId": "ImageJoin",
+				"tags": [
+					"storage"
+				],
+				"consumes": [
+					"application/json",
+					"application/octet-stream"
```
sflxn: you should remove application/octet-stream

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -243,6 +244,41 @@ func (c *ContainerProxy) CreateContainerHandle(vc *viccontainer.VicContainer, co
 	return id, h, nil
 }
 
+// AddImageToContainer adds the specified image to a container, referenced by handle.
+// If an error is return, the returned handle should not be used.
+//
+// returns:
+//	modified handle
+func (c *ContainerProxy) AddImageToContainer(handle, deltaID, layerID, imageID string, config types.ContainerCreateConfig) (string, error) {
```
sflxn: Can we document this "deltaID" above.  I had to cross reference several location to figure out what this deltaID was.  Is there a reason we don't just use id as we've done in the other functions? I know id is not very precise, but now we have this one-off that refers to "deltaID".

lib/apiservers/portlayer/swagger.json
```diff
@@ -593,6 +593,68 @@
...
+				],
+				"parameters": [
+					{
+						"name": "store_name",
+						"type": "string",
+						"in": "path",
+						"required": true
+					},
+					{
+						"name": "id",
```
sflxn: should we rename this to layerid?  I know it's very specific to docker, but we've been following a loose convention of setting id to the object that we're acting on.  Here, we doing ImageJoin, but the id is weirdly layerid.  If we follow our loose convention, id would be something like the container or handle.  Since it's not one of those, maybe we should rename this from id to layerID or shove it into the config parameter.

lib/apiservers/engine/backends/container_proxy.go
```diff
@@ -243,6 +244,41 @@ func (c *ContainerProxy) CreateContainerHandle(vc *viccontainer.VicContainer, co
...
+	if c.client == nil {
+		return "", InternalServerError("ContainerProxy.AddImageToContainer failed to get the portlayer client")
+	}
+
+	host, err := sys.UUID()
+	if err != nil {
+		return "", InternalServerError("ContainerProxy.AddImageToContainer got unexpected error getting VCH UUID")
+	}
+
+	response, err := c.client.Storage.ImageJoin(storage.NewImageJoinParamsWithContext(ctx).WithStoreName(host).WithID(layerID).
```
sflxn: This reads funny.  We're joining an image to a layer?  Should WithID() be given the deltaID, instead?

lib/portlayer/storage/image/join.go
```diff
@@ -0,0 +1,133 @@
...
+func Join(op trace.Operation, handle *exec.Handle, id, imgID, repoName string, img *Image) (*exec.Handle, error) {
+	defer trace.End(trace.Begin(img.ID, op))
+
+	// if _, ok := handle.ExecConfig.Mounts[volume.ID]; ok {
+	// 	return nil, fmt.Errorf("Volume with ID %s is already in container %s's mountspec config", volume.ID, handle.ExecConfig.ID)
+	// }
+
+	// //constuct MountSpec for the tether
+	// mountSpec := createMountSpec(volume, mountPath, diskOpts)
+	// //append a device addition spec change to the container config
```
sflxn: Remove commented out code?

lib/apiservers/engine/backends/container_proxy.go
hickeng: We're joining an image from a given image store and identified by it's ID.
The ID of the image to join is held in layerID (the leaf of the image tree).
The ID to use for creating a read/write layer based on the image we're joining is in deltaID.

Basically we've got lots of IDs so there's going to be some nomenclature confusion

lib/apiservers/portlayer/swagger.json
hickeng: The layerID _is_ the object we're acting on - that's the ID of the image that we're joining to the container.

lib/apiservers/portlayer/restapi/handlers/storage_handlers.go
```diff
@@ -350,6 +357,38 @@ func (h *StorageHandlersImpl) WriteImage(params storage.WriteImageParams) middle
...
+	}
+
+	cfg := params.Config
+	handleprime, err := vsimage.Join(op, handle, cfg.DeltaID, cfg.ImageID, cfg.RepoName, img)
+	if err != nil {
+		op.Errorf("join image failed: %#v", err)
+		return storage.NewImageJoinInternalServerError().WithPayload(&models.Error{Message: err.Error()})
+	}
+
+	op.Debugf("image %s has been joined to %s as %s", params.ID, handle.Spec.ID(), cfg.DeltaID)
```
matthewavery: This could be useful at the `Info` level, it is an important part of the container create cycle.

lib/portlayer/storage/container/import.go
```diff
@@ -0,0 +1,120 @@
...
+func (c *ContainerStore) Import(op trace.Operation, id string, spec *archive.FilterSpec, tarstream io.ReadCloser) error {
+	l, err := c.NewDataSink(op, id)
+	if err != nil {
+		return err
+	}
+
+	return l.Import(op, spec, tarstream)
+}
+
+// NewDataSink creates and returns an DataSink associated with container storage
```
matthewavery: s/an/a

lib/portlayer/storage/image/vsphere/join.go
```diff
@@ -0,0 +1,70 @@
...
+	// NOTE: this is a POOR assumption - I'm not clear on how it's functioning on vSAN at all in shipping code given the assumption that
+	// "[ds] id/id.vmdk" is a legitimate path. Some vsphere magic path adjustment?
+	rwlayer := fmt.Sprintf("%s/%s.vmdk", path.Dir(handle.Spec.VMPathName()), id)
+
+	disk := handle.Guest.NewDisk()
+	moref := handle.Spec.Datastore.Reference()
+
+	// NOTE: this spec construction really should be captured in one place down in the disk layer. That code is currently biased towards
+	// the appliance disk flows so couples spec creation with disk creation/attach.
+	// TODO: we absolutely shouldn't be mixing the handle.Spec.Datastore (wtf does this come from) and the DatastorePath for the disk
```
matthewavery: Perhaps some of this work/fixup can come with the batching logic for the disk manager and volume cache?

sflxn:
lgtm

matthewavery:
lgtm

matthewavery:
lgtm

matthewavery:
lgtm

lib/portlayer/storage/container/import.go
```diff
@@ -0,0 +1,120 @@
...
+		// sanity check to see if we are the owner - this should catch transitions
+		// from container running to diff or commit for example between the offline attempt and here
+		uuid, err := o.UUID(op)
+		if err == nil {
+			// check if the vm is appliance VM if we can successfully get its UUID
+			// #nosec: Errors unhandled.
+			self, _ := guest.IsSelf(op, uuid)
+			if self && offlineAttempt < 2 {
+				op.Infof("Appliance is owner of online vmdk - retrying offline source path")
+				goto offline
```
jzt: Curious: do we want to retry without any sort of time delay/backoff?

lib/portlayer/storage/image/cache/cache.go
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
```
jzt: 2017-2018

lib/portlayer/storage/image/image_test.go
```diff
@@ -12,13 +12,14 @@
 // See the License for the specific language governing permissions and
```
jzt: 2016-2018

lib/portlayer/storage/image/vsphere/export.go
```diff
@@ -0,0 +1,111 @@
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
```
jzt: 2018

lib/portlayer/storage/image/vsphere/import.go
```diff
@@ -0,0 +1,65 @@
+// Copyright 2016-2017 VMware, Inc. All Rights Reserved.
```
jzt: I know its getting old, but 2018 :)

lib/portlayer/storage/volume/vsphere/join.go
```diff
@@ -22,61 +22,49 @@ import (
...
-	//append a device addition spec change to the container config
-	diskDevice := createVolumeVirtualDisk(volume)
-	config := createDeviceConfigSpec(diskDevice)
-	handle.Spec.DeviceChange = append(handle.Spec.DeviceChange, config)
-
 	if handle.ExecConfig.Mounts == nil {
 		handle.ExecConfig.Mounts = make(map[string]executor.MountSpec)
 	}
+
+	//constuct MountSpec for the tether
```
jzt: s/constuct/construct

lib/portlayer/storage/container/import.go
hickeng: Yes - this code has determined that the disk in question is already connected to the appliance and that's why we couldn't attach it. That means we should be able to fall back to the offline path immediately.
There's no repeated retry so no delay/backoff logic.

zjs:
It seems like this needs to be rebased before it can be effectively reviewed. Adding `[Needs Rebase]` to summary.

lib/portlayer/storage/storage.go
```diff
@@ -12,6 +12,74 @@
...
+// * importer
+// * data source
+// * data sink
+//
+// The odd one out for these concepts is the `store`. This is because we did not want to be prescriptive about
+// what a store supports, but it may well distill into an interface at a later date. Currently the various
+// store implementations consist of a composition of `resolver`, `importer`, and `exporter` interfaces.
+//
+// The `data source` and `data sink` interfaces are present to allow abstracted read and write access respectively
+// _to a specific storage mechanism. The data source and data sink instances should be considered single use and not
```
zjs: nit: is the underscore at the beginning of this line intentional?

lib/portlayer/storage/storage.go
```diff
@@ -12,6 +12,74 @@
...
+// a. VMDKs on a vSphere datastore
+// b. NFS shares
+//
+// We have more than two sub-packages however as we use distinct packages to provide different lifecycle and
+// usage semantics for types of data using those mechanisms. Examples include:
+// * application logic (images)
+// * application transient data (image read/write layer)
+// * application persistent data (volumes)
+//
+// As such there are a set of concepts captured as interfaces:
```
zjs: nit: it's a little confusing to say that this is a list of interfaces, include store, and then explain in the following paragraph that store isn't an interface

lib/portlayer/storage/storage.go
```diff
@@ -12,6 +12,74 @@
...
+//
+// The data source and sink interfaces also specify a basic accessor method for the underlying mechanism. This was
+// added to allow for some future-proofing, with the primary extension expected to be a generic FileWalker interface
+// so that directory walk and listings can be done without needing to care about the access mechanism (e.g. common
+// between a local filesystem mount, an XDR NFS client, or a VM toolbox client).
+//
+// The general structure of a specific storage implementation is as follows and is imposed so that it is easier to
+// locate a specific functional path for a given usage/storage type combination:
+// ```
+// storage/ - this package
```
zjs: I like the inclusion of a file tree here.

Some thoughts:
1. It'd be good to include `storage.go`, at least to explain when something should go in `storage.go` vs. one of the other top-level files — something simple like "interfaces, `RegisterImporter`, and `RegisterExporter`" may encourage people to put other stuff in a more specific file.
2. It seems like `errors.go` is a consistent convention for the existing code, and is probably worth including.
3. It seems like `store.go` is a consistent convention for the existing code, and is probably worth including.
4. I found it a little hard to differentiate between "variables" (e.g., `implementation`) and "literals" (e.g., `cache`).

In order words, how about something like:

```
// storage/                  - this package
// ├── [type]/               - the high level type of use, implies semantics (e.g., volume)
// │   ├── [type].go         - the common store implementation aspects across implementations
// │   ├── errors.go         - ???
// │   ├── cache             - a `store` compliant cache implementation, if needed, with appropriate semantics for the type of use
// │   │   └── cache.go
// │   └── [implementation]/ - a specific implementation type (e.g., vsphere)
// │       ├── bind.go       - modifies a portlayer handle to configure active use of a specific `joined` storage
// │       ├── export.go     - implements the `read` side of the interfaces (Exporter and DataSource)
// │       ├── import.go     - a specific implementation type
// │       ├── join.go       - modifies a portlayer handle to configure basic access to a specific instance of the storage type
// │       └── store.go      - ???
// ├── storage.go            - ???
// └── [purpose].go          - common functions used by the various type/implementation combinations
```

isos/bootstrap/bootstrap
```diff
@@ -114,7 +114,7 @@ if timeout --signal=KILL 2m mount -t ext4 /dev/disk/by-label/containerfs ${MOUNT
     # TEMP: https://github.com/vmware/vic/issues/6279
     echo 262144 > /proc/sys/vm/max_map_count
 
-    until [[ $(ls -1 /dev/disk/by-label | wc -l) -eq $(ls -1 /sys/block | wc -l) ]]; do sleep 0.1;done
+    until [[ $(ls -1 /dev/disk/by-label | wc -l) -eq $(ls -1 --ignore='loop*' --ignore='ram*' /sys/block | wc -l) ]]; do sleep 0.1;done
```
zjs: What's the significance of this change?

lib/apiservers/engine/proxy/storage_proxy.go
```diff
@@ -239,6 +241,60 @@ func (s *StorageProxy) Remove(ctx context.Context, name string) error {
 	return nil
 }
 
+// AddImageToContainer adds the specified image to a container, referenced by handle.
+// If an error is return, the returned handle should not be used.
```
zjs: nit: "is return**ed**"

lib/apiservers/portlayer/restapi/handlers/storage_handlers_test.go
```diff
@@ -17,23 +17,23 @@ package handlers
...
 
-	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/vic/lib/apiservers/portlayer/models"
 	"github.com/vmware/vic/lib/apiservers/portlayer/restapi/operations/storage"
-	"github.com/vmware/vic/lib/archive"
 	"github.com/vmware/vic/lib/constants"
-	spl "github.com/vmware/vic/lib/portlayer/storage"
+	"github.com/vmware/vic/lib/portlayer/storage/image"
+	icache "github.com/vmware/vic/lib/portlayer/storage/image/cache"
+	imagemock "github.com/vmware/vic/lib/portlayer/storage/image/mock"
```
zjs: nit: `icache` and `imagemock` is... incongruous. It results in code like `icache.NewLookupCache(imagemock.NewMockDataStore(nil))`

Any reason not to be consistent? (`icache.NewLookupCache(imock.NewMockDataStore(nil)),` or `imagecache.NewLookupCache(imagemock.NewMockDataStore(nil)),`)

lib/portlayer/storage/storage.go
```diff
@@ -12,6 +12,74 @@
...
+//       import.go - implements the `write` side of the interfaces (Importer and DataSink)
+//       export.go - implements the `read` side of the interfaces (Exporter and DataSource)
+//       join.go - modifies a portlayer handle to configure basic access to a specific instance of the storage type
+//       bind.go - modifies a portlayer handle to configure active use of a specific `joined`` storage
+//     type.go - the common store implementation aspects across implementations
+//   [xxx.go] - common functions used by the various type/implementation combinations
+// ```
+//
+// This structure is not completely consistent at the time of writing. Most notable is that the bind.go functions
+// have all been rolled into the Join calls (the portlayer uses the following common verbs across components, again
```
zjs: nit: mismatched parenthesis

lib/apiservers/portlayer/swagger.json
```diff
@@ -637,6 +637,67 @@
 					}
 				}
 			},
+			"post": {
+				"description": "Join an image to a container",
```
zjs: nit: Can we describe this more... usefully? (If someone doesn't know what "ImageJoin" does, "Join an image" probably isn't going to clarify things.) Perhaps you could steal a different phrasing from you package documentation?

lib/apiservers/portlayer/swagger.json
```diff
@@ -4112,4 +4196,4 @@
 			}
 		}
 	}
-}
+}
```
zjs: nit: this seems like an accidental change

tests/concurrent/concurrent.go
```diff
@@ -256,7 +256,7 @@ func main() {
 			Name:       name,
 			VMFullName: name,
 
-			ParentImageID: parent,
+			// ParentImageID: parent,
```
zjs: What's the purpose of this change?

hickeng: `ParentImageID` has been deleted as a field - it was being used to pass the parent ID through to the hacky code that was constructing the image disk reference. Now that that is done in a structured manner the field is no longer needed.

PR #7399, opened 2018-02-26
Add ESX host metrics library

### LINKED ISSUE #7327: Host Metrics
In support of #7279 vic will need to gather host performance / usage metrics.  These metrics will be used to help make the host placement recommendation.

- [x] document design including metric identification
- [x] Host metric collection
- [x] vcsim testing

anchal-agrawal:
This change adds the HostMetrics() library to pkg/vsphere/performance.
HostMetrics takes a populated session as input, collects all hosts in
the cluster and feeds them to the govmomi metrics helpers. The returned
artifact is a map of HostSystem objects to their assembled metric value.

We currently collect CPU - active usage percentage - and memory -
active, consumed, overhead and total (all in KB). Non-aggregate metric
values returned by govmomi are not processed.

Fixes #7327 

---

Sample usage:

```diff
--- a/lib/portlayer/exec/commit.go
+++ b/lib/portlayer/exec/commit.go
@@ -24,6 +24,7 @@ import (
        "github.com/vmware/vic/lib/portlayer/event/events"
        "github.com/vmware/vic/pkg/retry"
        "github.com/vmware/vic/pkg/trace"
+       "github.com/vmware/vic/pkg/vsphere/performance"
        "github.com/vmware/vic/pkg/vsphere/session"
        "github.com/vmware/vic/pkg/vsphere/tasks"
        "github.com/vmware/vic/pkg/vsphere/vm"
@@ -56,6 +57,24 @@ func Commit(op trace.Operation, sess *session.Session, h *Handle, waitTime *int3
 
                var res *types.TaskInfo
                var err error
+
+               op.Infof("checking host metrics")
+               metrics, err := performance.HostMetrics(op, sess)
+               if err != nil {
+                       op.Errorf("error: %s", err)
+               } else {
+                       for host, metric := range metrics {
+                               op.Infof("host: %s", host.String())
+                               op.Infof("cpu usage: %f", metric.CPU.UsagePercent)
+                               op.Infof("mem active: %d", metric.Memory.ActiveKB)
+                               op.Infof("mem consumed: %d", metric.Memory.ConsumedKB)
+                               op.Infof("mem overhead: %d", metric.Memory.OverheadKB)
+                               op.Infof("mem total: %d", metric.Memory.TotalKB)
+                       }
+               }
```

### DIFF
--- a/pkg/vsphere/performance/host_metrics.go
+++ b/pkg/vsphere/performance/host_metrics.go
@@ -1,0 +1,179 @@
+// Copyright 2018 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package performance
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/performance"
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/session"
+)
+
+const (
+	// cpuUsage measures the actively used CPU of the host, as a percentage of the total available CPU.
+	cpuUsage = "cpu.usage.average"
+
+	// memActive measures the sum of all active metrics for all powered-on VMs plus vSphere services on the host.
+	memActive = "mem.active.average"
+
+	// memConsumed measures the amount of machine memory used on the host, including vSphere services, VMkernel,
+	// the service console and the total consumed memory metrics for all running VMs.
+	memConsumed = "mem.consumed.average"
+
+	// memTotalCapacity measures the total amount of memory reservation used by and available for powered-on VMs
+	// and vSphere services on the host.
+	memTotalCapacity = "mem.totalCapacity.average"
+
+	// memOverhead measures the total of all overhead metrics for powered-on VMs, plus the overhead of running
+	// vSphere services on the host.
+	memOverhead = "mem.overhead.average"
+)
+
+// HostMemory stores an ESXi host's memory metrics.
+type HostMemory struct {
+	ActiveKB   int64
+	ConsumedKB int64
+	OverheadKB int64
+	TotalKB    int64
+}
+
+// HostCPU stores an ESXi host's CPU metrics.
+type HostCPU struct {
+	UsagePercent float64
+}
+
+// HostMetricsInfo stores an ESXi host's memory and CPU metrics.
+type HostMetricsInfo struct {
+	Memory HostMemory
+	CPU    HostCPU
+}
+
+// HostMetrics returns CPU and memory metrics for all ESXi hosts in the input session's cluster.
+func HostMetrics(op trace.Operation, session *session.Session) (map[*object.HostSystem]*HostMetricsInfo, error) {
+	if session == nil {
+		return nil, fmt.Errorf("session not set")
+	}
+
+	// Gather hosts from the session cluster and then obtain their morefs.
+	hosts, err := gatherHosts(op.Context, session)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/vsphere/performance/host_metrics.go
```diff
@@ -0,0 +1,178 @@
...
+}
+
+// HostMetricsInfo stores an ESXi host's memory and CPU metrics.
+type HostMetricsInfo struct {
+	Memory HostMemory
+	CPU    HostCPU
+}
+
+// HostMetrics returns CPU and memory metrics for all ESXi hosts in the input session's cluster.
+func HostMetrics(session *session.Session) (map[*object.HostSystem]*HostMetricsInfo, error) {
```
vburenin: Did you consider passing trace.Operation as a function parameter?

anchal-agrawal: Sure. I can see how that would be helpful; the caller can supply their own operation and context (and therefore, the deadline/cleanup steps). Are there other benefits?

pkg/vsphere/performance/host_metrics.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	// Gather hosts from the session cluster and then obtain their morefs.
+	hosts, err := gatherHosts(op.Context, session)
+	if err != nil {
+		return nil, fmt.Errorf("unable to obtain host morefs from session: %s", err)
+	}
+	morefToHost := make(map[types.ManagedObjectReference]*object.HostSystem)
+	morefs := make([]types.ManagedObjectReference, len(hosts))
+	for i := range hosts {
```
vburenin: I think since you are interested in index and values it might make sense to do this like
```
for i, h := range hosts {
	moref := h.Reference()
	morefToHost[moref] = h
	morefs[i] = moref
}
```

When for iterates over array it doesn't make a copy of an array element, so technically it should be even faster than accessing by index.

pkg/vsphere/performance/host_metrics.go
```diff
@@ -0,0 +1,178 @@
...
+		// Process each value and assign it directly to the corresponding metric
+		// field since there is only one sample.
+		for _, v := range res.Value {
+
+			// We don't need to collect non-aggregate (non-empty Instance) metrics.
+			if v.Instance != "" {
+				continue
+			}
+
+			switch v.Name {
```
vburenin: I would check v.Value for the length, just in case. since accessing it by [0] is kinda spooky.

pkg/vsphere/performance/host_metrics.go
anchal-agrawal: Interesting, per my understanding, there will be a copy here unless we iterate over the pointer to the array. https://garbagecollected.org/2017/02/22/go-range-loop-internals/

pkg/vsphere/performance/host_metrics.go
anchal-agrawal: Sounds good - will do.

pkg/vsphere/performance/host_metrics.go
cgtexmex: consistency of OpID logging -- which is passed down to vSphere, so we'd have better traceability -- so I agree we should have op as a param

pkg/vsphere/performance/host_metrics.go
cgtexmex: The i from `for i := range results` could be used in place of the zero

pkg/vsphere/performance/host_metrics.go
vburenin: What does it print? https://play.golang.org/p/4eFbdl1b1Sg

pkg/vsphere/performance/host_metrics.go
anchal-agrawal: Fixed, thanks @vburenin @cgtexmex!

pkg/vsphere/performance/host_metrics.go
anchal-agrawal: Fixed, thanks!

vburenin: After refreshing some internal details, the value is still going to be copied, but for/range reuses the same memory address for the value.

pkg/vsphere/performance/host_metrics.go
anchal-agrawal: Not sure if the index from `range results` will work. Each `result` holds a slice of metrics for a particular moref, and the metrics slice has a `Values` field with only one value since we're fetching one sample.

I've added the length check though.

anchal-agrawal:
Verified clean build on the old CI infra: https://ci.vcna.io/vmware/vic/16556. Merging.

PR #7408, opened 2018-02-27
Fix `docker pull` variant of tar unpacking bug

### LINKED ISSUE #6865: Fix symlink handling for Docker Pull
Fix the issue with malicious tar during docker pull, as described in bug 1973536.

gigawhitlocks:
I have manually replicated the bug as described in bz 1973536 and have confirmed that this change does indeed resolve the issue. 

Resolves #6865 and bz 1973536
`[specific ci=1-02-Docker-Pull]`


cmd/archive/unpack.go
```diff
@@ -36,26 +38,44 @@ const (
...
+	if _, err := docker.ApplyLayer("/", os.Stdin); err != nil {
+		op.Errorf("Error applying layer: %s", err)
+		os.Exit(FailedApplyLayer)
+	}
+
+	os.Exit(Success)
+
+}
+
+func setupChroot() trace.Operation {
```
vburenin: This function has little to do with the returned trace.Operation.
It even would make unit testing a lot more complicated if log messages are the part of the unit test checks.
Please pass trace.Operation as a function parameter initializing it before setupChroot call.

lib/archive/unpack_unix.go
```diff
@@ -202,6 +202,56 @@ func streamCopy(op trace.Operation, stdin io.WriteCloser, tarStream io.Reader) e
 	}
 }
 
+func DockerUnpack(op trace.Operation, root string, tarStream io.Reader) (int64, error) {
+	fi, err := os.Stat(root)
+	if err != nil {
+		// the target unpack path does not exist. We should not get here.
+		op.Errorf("tar unpack target does not exist: %s", root)
```
vburenin: You may get this error not only because file doesn't not exist, there permission reasons and other potential error. You may want to check for NotExists error specifically. before you write a log message.

lib/archive/unpack_unix.go
```diff
@@ -202,6 +202,56 @@ func streamCopy(op trace.Operation, stdin io.WriteCloser, tarStream io.Reader) e
...
+		return 0, err
+	}
+
+	if !fi.IsDir() {
+		err := fmt.Errorf("unpack root target is not a directory: %s", root)
```
vburenin: Does it even make sense to log messages in this function? Would not it be better to just return a specific error? 
```
return 0, fmt.Errorf("unpack root target is not a directory: %s", root)
```

lib/archive/unpack_unix.go
```diff
@@ -202,6 +202,56 @@ func streamCopy(op trace.Operation, stdin io.WriteCloser, tarStream io.Reader) e
...
+	}
+
+	if stdin == nil {
+		err = errors.New("stdin was nil")
+		op.Error(err)
+		return 0, err
+	}
+
+	var n int64
+	go func(n int64) {
```
vburenin: You just shadowed n here. so returned n is always 0.

lib/archive/unpack_unix.go
```diff
@@ -202,6 +202,56 @@ func streamCopy(op trace.Operation, stdin io.WriteCloser, tarStream io.Reader) e
...
+	}(n)
+
+	if err = cmd.Start(); err != nil {
+		return 0, err
+	}
+
+	if err = cmd.Wait(); err != nil {
+		return 0, err
+	}
+	return n, nil
```
vburenin: you are also potentially racing here for than n value when cmd.Wait completes if cmd.Wait internals are not synchronized.

lib/archive/unpack_unix.go
```diff
@@ -202,6 +202,56 @@ func streamCopy(op trace.Operation, stdin io.WriteCloser, tarStream io.Reader) e
...
+	go func(n int64) {
+
+		defer stdin.Close()
+		if n, err = io.Copy(stdin, tarStream); err != nil {
+			op.Errorf("Error copying tarStream: %s", err.Error())
+		}
+
+	}(n)
+
+	if err = cmd.Start(); err != nil {
```
vburenin: You are starting a goroutine before you call start. Does that goroutine complete its work if cmd.Start fails?

cmd/archive/unpack.go
```diff
@@ -88,10 +108,16 @@ func main() {
...
-	if err = archive.UnpackNoChroot(op, os.Stdin, filterSpec, "/"); err != nil {
-		op.Error(err)
-		os.Exit(FailedInvokeUnpack)
-	}
+	return op
+}
 
-	os.Exit(Success)
+func main() {
+	switch len(os.Args) {
```
vburenin: I think all these magic numbers need some docs.

lib/archive/unpack_unix.go
gigawhitlocks: no but it blocks while waiting to read in `io.Copy`

cmd/archive/unpack.go
gigawhitlocks: yeah i just need a better argument structure entirely, this is all implicit and not great

gigawhitlocks: at the very least i'll add docs

lib/archive/unpack_unix.go
gigawhitlocks: yeah, if we get an error here, we should just report the error regardless of what kind it is -- I'll update this. I'll also update the other parts of the file this same error reporting routine is used, because I totally copied this part from old code, ha.

lib/archive/unpack_unix.go
gigawhitlocks: Hmm, probably. I must've had a reason for doing this, but it doesn't jump out at me.

lib/archive/unpack_unix.go
gigawhitlocks: ah, right, shouldn't be an argument, aw heck

cmd/archive/unpack.go
gigawhitlocks: from a conceptual standpoint i see where you're coming from, so i'll refactor this a bit, but 

> It even would make unit testing a lot more complicated if log messages are the part of the unit test checks.

how is unit testing any of this even possible in the first place? it's all disk operations & system calls.

vburenin: Better abstraction and design make things more testable. It is hard to test, not because it works with files and syscalls, it is hard to test because it is designed so. Frequently asking a question "how do I test it?" lead to a better design in the future, that is SOLID and testable.

lib/archive/unpack_unix.go
gigawhitlocks: no, this should go after start and before wait, i realize. it also doesn't complete its work if EOF is never seen, same as w/ `unpack`.. @hickeng and I were trying to think of a way to avoid that problem, but weren't able to think of anything sufficiently reliable and uncomplicated. If you have insight into that @vburenin it would be much appreciated feedback.

lib/archive/unpack_unix.go
gigawhitlocks: yeah, i'll sync w/ a channel

cmd/archive/unpack.go
gigawhitlocks: Sure, usually that's correct, but in this case, if I redesign everything against an interface I'd have to mock out all of the function calls, and then I'm just testing a mock. What's the point? The only thing this function does is call system operations & handle errors. I guess I could test the error checking, but if `if err != nil` is a fragile operation, we're in bad shape.

cmd/archive/unpack.go
```diff
@@ -36,26 +38,47 @@ const (
...
-	op.Debugf("New unpack operation created")
-
-	if len(os.Args) != 4 {
-		op.Errorf("Wrong number of arguments passed to unpack binary")
-		os.Exit(WrongArgumentCount)
-	}
-
-	root := os.Args[2]
+// Calls our custom unpack method inside of the chroot
+func customUnpack(op trace.Operation) {
```
vburenin: Not sure it makes sense to do that in this PR since it is not your change, but referring os.Args[3] inside this function is just not even close to ok. I am only making this comment to point that out, not to change it.

cmd/archive/unpack.go
```diff
@@ -36,26 +38,47 @@ const (
...
+		op.Errorf("Error applying layer: %s", err)
+		os.Exit(FailedApplyLayer)
+	}
+
+	os.Exit(Success)
+
+}
+
+// Performs sanity checking (makes sure the unpack directory exists and has permissions, etc), changes directory to the unpack directory, and calls chroot on that directory. Will exit the executable if an error occurs.
+func setupChroot(op trace.Operation) {
```
vburenin: same here. os.Arg[2] has to be a function parameter.

gigawhitlocks:
RFC on the integration test from @mhagen-vmware

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
```diff
@@ -159,3 +166,60 @@ Verify image manifest digest against vanilla docker
     Log  ${output}
     Should Be Equal As Integers  ${rc}  0
     Should Contain  ${output}  sha256:be3c11fdba7cfe299214e46edc642e09514dbb9bbefcd0d3836c05a1e0cd0642
+
+Attempt docker pull mitm
+    Wait Until Keyword Succeeds  5x  15 seconds  Pull image  gigawhitlocks/docker-layer-injection-proxy:latest
```
jzt: Consider the hit by a bus rule: worth putting this image in vmware's bintray or dockerhub account?

gigawhitlocks: I was thinking of adding the folder I used to build them, with the 2 dockerfiles, to the repo? I wasn't sure and was hoping @mhagen-vmware could give me some feedback as to the best approach for that. The registry one, in particular, is a fork of this repo https://github.com/docker/distribution-library-image but they don't have a license on the repo? So I'm not sure if I can check that in?

gigawhitlocks: I guess yeah I could just put the images somewhere, although if I get hit by a bus my dockerhub will live on. You'd just have to pull them down in the case dockerhub was being EOL'd.

gigawhitlocks:
It's .. not really a CLI tool .. it's a binary .. I mean you _could_ use it on the CLI but it doesn't provide any niceties or expect the user to do so.. @vburenin which included library implementation details do you not like? If it's something in `unpack_unix.go` I'd like to take your feedback on that, because other than `UnpackNoChroot` that code is all from my last 2 PRs.

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
mdubya66: I really don't like this. we have it already with some of our images though. it should go under vmware IMO. @mhagen-vmware

mhagen-vmware: We have a docker hub account already:
https://hub.docker.com/r/victest/

The password is in our secrets

mhagen-vmware: and yea, the dockerfiles should be added to the repo

gigawhitlocks: Until I know what to do with it for sure, I posted the additional code related to this test (what's running in those two images) here https://github.com/gigawhitlocks/mitmproxy-docker-registry

cmd/archive/unpack.go
```diff
@@ -36,62 +38,91 @@ const (
...
+		op.Errorf("Error applying layer: %s", err)
+		os.Exit(FailedApplyLayer)
 	}
 
-	fi, err := os.Stat(root)
+	os.Exit(Success)
+
+}
+
+// Performs sanity checking (makes sure the unpack directory exists and has permissions, etc), changes directory to the unpack directory, and calls chroot on that directory. Will exit the executable if an error occurs.
```
jzt: Minor: can you break this comment into multiple lines?

cmd/archive/unpack.go
```diff
@@ -36,62 +38,91 @@ const (
...
-	if err = archive.UnpackNoChroot(op, os.Stdin, filterSpec, "/"); err != nil {
-		op.Error(err)
-		os.Exit(FailedInvokeUnpack)
+func main() {
+	ctx := context.Background()
+	op := trace.NewOperation(ctx, "Unpack")
+
+	switch len(os.Args) {
+	case 3:
+		// When performing an unpack via docker cp, we use a custom unpack routine, which requires a FilterSpec. Thus, if we have 3 arguments (command, unpack location, filterspec), we should perform a custom unpack
```
jzt: This one and the one on 123 as well

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
gigawhitlocks: kk

cmd/archive/unpack.go
gigawhitlocks: yes

but also turn on line wrapping in emacs

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
gigawhitlocks: also this "we have it already with some of our images though" is the only reason I did it this way initially, lol

gigawhitlocks: OK I committed the necessary bits to build those two images. Third party includes are `mitmdump` (MIT License https://github.com/mitmproxy/mitmproxy/blob/master/LICENSE) and the docker registry image Dockerfile (edited) which is Apache licensed https://github.com/docker/distribution/blob/master/LICENSE

gigawhitlocks: We don't compile against those so I don't exactly know if the LICENSE matters, but Apache & MIT should be OK regardless.

tests/test-cases/Group1-Docker-Commands/1-44-Docker-CP-Online.robot
```diff
@@ -309,5 +309,5 @@ Malformed tarball doesn't extract to wrong container from shared volume
     Should Be Equal As Integers  ${rc}  0
     ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} start two
     Should Be Equal As Integers  ${rc}  0
-    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} exec two ls /tmp
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} exec one ls /tmp
```
gigawhitlocks: This is otherwise unrelated to this change. I just noticed that this test was technically incorrect while preparing my demo.

tests/resources/dockerfiles/docker-pull-mitm-proxy/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+FROM python:3
+COPY archive.tar /
+COPY docker-mitm.py /
+COPY mitmdump /
+RUN chmod +x /mitmdump
```
mhagen-vmware: just check it in to the repo as executable?

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
```diff
@@ -16,9 +16,85 @@
...
+    # We're just going to eschew helpers and install this VCH manually to avoid mutating hidden environmental state which is difficult to debug
+    ${rc}  ${output}=  Run And Return Rc And Output  bin/vic-machine-linux create --name=VCH-XPLT --target=%{TEST_URL}%{TEST_DATACENTER} --user=%{TEST_USERNAME} --password=%{TEST_PASSWORD} --force=true --compute-resource=%{TEST_RESOURCE} --timeout %{TEST_TIMEOUT} --bridge-network=%{BRIDGE_NETWORK} --container-network=%{PUBLIC_NETWORK}:public --public-network=%{PUBLIC_NETWORK} %{VICMACHINETLS} --image-store=%{TEST_DATASTORE} --insecure-registry=http://${registry} --http-proxy http://${mitm}
+    Log  ${output}
+
+    Should Be Equal As Integers  ${rc}  0
+    ${rc}  ${vch2-params}=  Run And Return Rc And Output  echo '${output}' | grep -A1 "Connect to docker" | tail -n1 | cut -d' ' -f4- | sed 's/ info" $//g'
+
+    # this comment fixes syntax highlighting "
+
+    Log  ${output}
```
mhagen-vmware: duplicate

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
```diff
@@ -16,9 +16,85 @@
...
+    Should Be Equal As Integers  ${rc}  0
+
+    ${rc}  ${vch2-IP}=  Run And Return Rc And Output  echo '${output}' | grep -A1 "Published ports" | tail -n1 | awk '{print $NF}' | cut -d= -f2
+    Log  ${output}
```
mhagen-vmware: duplicate, not understanding if you are intentionally doing this for some reason?  The var hasn't changed since the first call?

tests/resources/VCH-Util.robot
```diff
@@ -72,6 +72,7 @@ Set Test Environment Variables
     ${domain}=  Get Environment Variable  DOMAIN  ''
     Run Keyword If  $domain == ''  Set Suite Variable  ${vicmachinetls}  --no-tlsverify
     Run Keyword If  $domain != ''  Set Suite Variable  ${vicmachinetls}  --tls-cname=*.${domain}
+    Set Environment Variable  VICMACHINETLS  ${vicmachinetls}
```
mhagen-vmware: not sure why you are doing this? it is set as a suite variable so should be accessible throughout your test already.  I would have preferred env var from the beginning but I don't think now is the time to change this, and it creates a confusing mixed config as well.

tests/resources/dockerfiles/docker-pull-mitm-proxy/README.md
```diff
@@ -0,0 +1,24 @@
...
+$ docker push localhost:5000/busybox
+$ docker commit registry
+$ docker tag registry victest/registry-busybox:latest
+$ docker push victest/registry-busybox:latest
+```
+You may need to add the local registry to your insecure registry list in `/etc/docker/daemon.json` before you can push to the registry.
+`docker build .` in either directory to recreate the image. The registry image is a standard docker registry image, except that the `VOLUME` 
+
+### MITMproxy Image
+This is a custom image that performs a MITM on the image being pulled from the registry container when Docker is configured to use it as an HTTP proxy. `docker build .` should be sufficient for this image and then just `docker push victest/docker-layer-injection-proxy:latest`
```
mhagen-vmware: Can you document at least in here or somewhere else appropriate, where the below binary blobs came from?

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
```diff
@@ -16,9 +16,85 @@
 Documentation  Test 1-02 - Docker Pull
 Resource  ../../resources/Util.robot
 Suite Setup  Conditional Install VIC Appliance To Test Server
-Suite Teardown  Cleanup VIC Appliance On Test Server
+Suite Teardown  Local Teardown
```
mhagen-vmware: don't do that, it is not appropriate for any test but the new one.  Just add a [Teardown] at the end of the specific test

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
```diff
@@ -16,9 +16,85 @@
...
 Test Timeout  20 minutes
 
+*** Keywords ***
+Local Teardown
+    Run Keyword  Cleanup VIC Appliance On Test Server
+    Run Keyword  Cleanup VCH Bridge Network  %{BRIDGE_NETWORK_2}
```
mhagen-vmware: just move this to a test [Teardown] and keep the original suite teardown.  Then eliminate this whole keyword.

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
```diff
@@ -16,9 +16,85 @@
...
+    [Arguments]  ${container-name}  ${docker-ps}
+    ${rc}  ${container}=  Run And Return Rc And Output  echo '${docker-ps}' | grep ${container-name} | awk '{print $(NF-1)}' | cut -d- -f1
+    Should Be Equal As Integers  ${rc}  0
+    Log  ${container}
+    [Return]  ${container}
+
+Pull And MITM Prepared Image
+    [Arguments]  ${vch2-params}  ${registry}
+    ${rc}  ${output}=  Run And Return Rc And Output  docker ${vch2-params} pull ${registry}/busybox
+    Log  ${output}
```
mhagen-vmware: is this complete?  Seems cut-off in the middle?

mhagen-vmware: or at least not doing what the name of the keyword implies it is intended to do..

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
```diff
@@ -16,9 +16,85 @@
...
+Run Docker Ps
+    ${rc}  ${output}=  Run And Return Rc And Output  docker %{VCH-PARAMS} ps
+    Should Be Equal As Integers  ${rc}  0
+    [Return]  ${output}
+
+Get Container Address
+    [Arguments]  ${container-name}  ${docker-ps}
+    ${rc}  ${container}=  Run And Return Rc And Output  echo '${docker-ps}' | grep ${container-name} | awk '{print $(NF-1)}' | cut -d- -f1
+    Should Be Equal As Integers  ${rc}  0
+    Log  ${container}
```
mhagen-vmware: very little reason to log this if you are returning it anyways

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
```diff
@@ -16,9 +16,85 @@
...
+    ${rc}  ${output}=  Run And Return Rc And Output  docker ${vch2-params} pull ${registry}/busybox
+    Log  ${output}
+
+Enable SSH On MITMed VCH
+    ${rc}  ${thumbprint}=  Run And Return Rc And Output  govc about.cert -k -json | jq -r .ThumbprintSHA1
+    Log  ${thumbprint}
+    Should Be Equal As Integers  ${rc}  0
+    ${rc}  ${output}=  Run And Return Rc And Output  bin/vic-machine-linux debug --rootpw password --target %{TEST_URL}%{TEST_DATACENTER} --password %{TEST_PASSWORD} --name VCH-XPLT --user %{TEST_USERNAME} --compute-resource %{TEST_RESOURCE} --enable-ssh --thumbprint=${thumbprint}
+    Should Be Equal As Integers  ${rc}  0
+    Log  ${output}
```
mhagen-vmware: for this to be beneficial in future debugging, should be above the rc check

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
```diff
@@ -159,3 +235,14 @@ Verify image manifest digest against vanilla docker
...
+Attempt docker pull mitm
+    Get And Run MITMProxy Container
+    Get And Run Prepared Registry
+    ${ps}=  Run Docker Ps
+    ${mitm}=  Get Container Address  mitm  ${ps}
+    ${registry}=  Get Container Address  registry  ${ps}
+    ${vch2-IP}  ${vch2-params}=  Deploy Proxified VCH  ${registry}  ${mitm}
+    Pull And MITM Prepared Image  ${vch2-params}  ${registry}
+    Enable SSH on MITMed VCH
+    Check For Injected Binary  ${vch2-IP}
```
mhagen-vmware: It looks like you have leaked the second VCH, you need to clean that up.  Your Local Cleanup (not really a clear name) - can probably take care of that and make sure that you run a manual vic-machine delete --name VCH-XPLT

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
gigawhitlocks: Didn't know about `Teardown`

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
gigawhitlocks: It does do what the keyword implies it is intended to do. VCH-2 is deployed to use MITMproxy as a proxy and the prepared registry. The `pull` exercises the bug, just like in the bug report, and in the real world.

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
gigawhitlocks: Hmm, I was having some trouble getting it to remove w/ a keyword & Cleanup Dangling VCHes always catches it on the next test but I'll go ahead and do it manually.

tests/resources/dockerfiles/docker-pull-mitm-proxy/README.md
gigawhitlocks: I thought it was clear that it's MITMProxy but I can provide a link to the project.

tests/resources/dockerfiles/docker-pull-mitm-proxy/README.md
```diff
@@ -0,0 +1,24 @@
...
+$ docker pull busybox:latest
+$ docker run -itd --net=host registry
+$ docker tag busybox localhost:5000/busybox
+$ docker push localhost:5000/busybox
+$ docker commit registry
+$ docker tag registry victest/registry-busybox:latest
+$ docker push victest/registry-busybox:latest
+```
+You may need to add the local registry to your insecure registry list in `/etc/docker/daemon.json` before you can push to the registry.
+`docker build .` in either directory to recreate the image. The registry image is a standard docker registry image, except that the `VOLUME`
```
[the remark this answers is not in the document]
gigawhitlocks: ah thanks, must've hit the wrong button in the editor

tests/resources/dockerfiles/docker-pull-mitm-proxy/Dockerfile
gigawhitlocks: Sure

tests/resources/VCH-Util.robot
gigawhitlocks: It is set as a suite variable? I did not think I was able to access `${vicmachinetls}` where I was using the variable but perhaps I am mistaken. If I am unable to use that variable, I'll get back to you on this. I didn't make this change arbitrarily.

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
gigawhitlocks: I had a lot of Log statements because this was a tough test to get working. Then I did a big refactor so that it was actually readable, because it became clear to me that the whole test was pretty opaque. A side effect of this, and that I didn't remove the Log statements because generally I think they're valuable, is that we got a couple dupes and a couple that aren't in places that make sense. Thanks for finding them in review.

tests/test-cases/Group1-Docker-Commands/1-02-Docker-Pull.robot
```diff
@@ -19,6 +19,85 @@ Suite Setup  Conditional Install VIC Appliance To Test Server
 Suite Teardown  Cleanup VIC Appliance On Test Server
 Test Timeout  20 minutes
 
+*** Keywords ***
+Cleanup MITM Test Detritus
+    Run Keyword  Destroy Proxified VCH
```
mhagen-vmware: that doesn't really do anything, don't need both Destroy Proxified VCH as well as this keyword...

gigawhitlocks: ahahaha yeah I had that thought while I was typing it out.. why does `Run Keyword` even exist as a keyword if it's never necessary?

mhagen-vmware: to be able to run a keyword as a variable,
i.e. Run Keyword  ${keyword1}  or ${keyword2}

PR #7409, opened 2018-02-27
Refactor VM delete process: re-enable destroy on VM, add retry pattern to concurrent access error

### LINKED ISSUE #6079: vic-machine delete failure: ServerFaultCode: The method is disabled by 'VIC' then failed to unregister VM: ServerFaultCode: vim.fault.ConcurrentAccess
8 out of 50 VCH - Delete VCH failed with -
```
Aug 22 2017 05:39:20.844-04:00 INFO  ### Removing VCH ####
Aug 22 2017 05:39:21.790-04:00 INFO  Validating target
Aug 22 2017 05:39:22.881-04:00 INFO
Aug 22 2017 05:39:22.881-04:00 INFO  VCH ID: VirtualMachine:vm-15275
Aug 22 2017 05:39:23.032-04:00 INFO  Removing VMs
Aug 22 2017 05:40:42.373-04:00 INFO  Unregistered VM to cleanup after failed destroy: "VirtualMachine:vm-15940"
Aug 22 2017 05:40:42.962-04:00 INFO  Unregistered VM to cleanup after failed destroy: "VirtualMachine:vm-17869"
...
Aug 22 2017 05:42:59.980-04:00 INFO  Specify --force to force delete
Aug 22 2017 05:42:59.980-04:00 INFO  Collecting da2ca126-f532-4bc0-9b6d-b7cc879b569e vpxd.log
Aug 22 2017 05:43:00.111-04:00 ERROR Failed to destroy VM "VirtualMachine:vm-15647": ServerFaultCode: The method is disabled by 'VIC' then failed to unregister VM: ServerFaultCode: vim.fault.ConcurrentAccess
Aug 22 2017 05:43:00.117-04:00 ERROR --------------------
Aug 22 2017 05:43:00.117-04:00 ERROR vic-machine-linux delete failed: delete failed
[vic-machine.zip](https://github.com/vmware/vic/files/1241611/vic-machine.zip)
```
Attached is the vic-machine log

AngieCris:
`[specific ci=6-03-Delete --suite 1-11-Docker-RM]`

Fixes #6079 

Two problems:
1. We disabled destroy method on VCH appliance VM and container VM on create, but we never re-enable the destroy method
2. Because of 1, we have to manually unregister the VM every time, and it's a potential resource leak
3. Either 1 or 2 could fail with `ConcurrentAccess` error. If that happens, retry is needed

Fixes:
1. Re-enabled destroy method on VM in `deleteExceptDisks` to avoid unnecessary VM unregister
2. Added retry pattern for `ConcurrentAccess` failure in VM delete


lib/install/management/appliance.go
```diff
@@ -190,33 +191,39 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
...
+	}
+
+	if err = retry.Do(operation, isConcurrentAccessError); err != nil {
+		d.op.Errorf("Delete VM %s fails with error: %s", vm.Reference(), err)
+		return err
 	}
 
 	if _, err = d.deleteDatastoreFiles(d.session.Datastore, folder, true); err != nil {
-		d.op.Warnf("Failed to remove datastore files for VM path %q: %s", folder, err)
+		d.op.Warnf("Remove datastore files for VM %s with folder path %s failed with error: %s", vm.Reference(), folder, err)
```
andrewtchin: I'd make this `"Failed to remove datastore files for VM %s with folder path %s: %s"`

lib/install/management/appliance.go
```diff
@@ -190,33 +191,39 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
...
-		if f, ok := err.(types.HasFault); ok {
-			switch f.Fault().(type) {
-			case *types.ConcurrentAccess:
-				d.op.Warn("DeleteExceptDisks failed with ConcurrentAccess error. Ignoring it.")
-				ignore = true
-			}
+	operation := func() error {
+		_, err := vm.DeleteExceptDisks(d.op)
+
+		if err != nil {
```
hickeng: Unclear.

If this is retrying the `vm.Unregister` then the `vm.DeleteExceptDisks` should be outside of the `operation` function.
If this is retrying `vm.DeleteExceptDisks` then the error check should be after the `retry.Do` or we will always unregister even for errors which trigger a retry.

lib/install/management/appliance.go
```diff
@@ -190,33 +191,39 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
...
-			if err2 != nil {
-				return errors.Errorf("%s then failed to unregister VM: %s", err, err2)
-			}
-			d.op.Infof("Unregistered VM to cleanup after failed destroy: %q", vm.Reference())
+
+		return nil
+	}
+
+	// Retry destroy or unregister VM on ConcurrentAccess error
+	isConcurrentAccessError := func(err error) bool {
```
hickeng: We have checks for concurrent access error in enough places that we should add `IsConcurrentAccessError` as a function into the tasks package for common reference.

pkg/vsphere/vm/vm.go
```diff
@@ -281,7 +284,37 @@ func (vm *VirtualMachine) DeleteExceptDisks(ctx context.Context) (*object.Task,
...
+
+	return nil, err
+}
+
+func isMethodDisabledError(ctx context.Context, err error) bool {
+	op := trace.FromContext(ctx, "isMethodDisabledError")
+
+	if soap.IsSoapFault(err) {
+		vimFault := soap.ToSoapFault(err).VimFault()
+		op.Debugf("Error type: %s", reflect.TypeOf(vimFault))
```
hickeng: I don't think this method needs the logging and it's not consistent with the `isConcurrentAccess` check. We report the error in the caller of `DeleteExceptDisks` if it's not a method disabled and we should be logging in `EnableDestroy`.
Not a strong preference on my part.

pkg/vsphere/vm/vm.go
```diff
@@ -281,7 +284,37 @@ func (vm *VirtualMachine) DeleteExceptDisks(ctx context.Context) (*object.Task,
...
-	return vm.Destroy(op)
+	info, err := vm.WaitForResult(op, func(ctx context.Context) (tasks.Task, error) {
+		return vm.Destroy(ctx)
+	})
+
+	if err == nil {
+		return info, nil
+	}
+
+	if isMethodDisabledError(ctx, err) {
```
hickeng: reverse the test - it's clearer to return early:
```
if !isMethodDisabledError(ctx, err) {
  return nil, err
}
```
you may even be able to combine it with the test above:
```
if err == nil || !isMethodDisabledError(ctx, err) {
   return info, err
}
```

pkg/vsphere/vm/vm.go
```diff
@@ -281,7 +284,37 @@ func (vm *VirtualMachine) DeleteExceptDisks(ctx context.Context) (*object.Task,
...
+		vm.EnableDestroy(ctx)
```
hickeng: If we fail calling `EnableDestroy` then we will not be able to delete the VM below and it's worthy of a custom error message and Warning logging - it requires user intervention to clean up at this point.

lib/install/management/appliance.go
AngieCris: You're right, if `DeleteExceptDisks` failed with `ConcurrentAccess` we should retry this separately before unregister as well. Fixing it.

lib/install/management/appliance.go
AngieCris: If `EnableDestroy` failed, a warning message is logged by `EnableDestroy` function, but we still attempt to destroy the VM after. We should not attempt to destroy the VM if `EnableDestroy` failed, and we should return the error back to the caller for retry. Fixing it.

pkg/vsphere/vm/vm.go
AngieCris: I will move all the error type checks to tasks package and make them all consistent. 
If it's not method disabled error, we report it back to DeleteExceptDisks, and the latter report back to the caller for retry. And then the caller logs the error message if retry fails.

pkg/vsphere/vm/vm.go
AngieCris: If EnableDestroy failed, a warning message is already logged by EnableDestroy function, but we still attempt to destroy the VM after. We should not attempt to destroy the VM if EnableDestroy failed, and we should return the error back to the caller for retry. Fixing it.

pkg/vsphere/tasks/waiter.go
```diff
@@ -185,13 +185,89 @@ func IsRetryError(op trace.Operation, err error) bool {
...
+	if soap.IsVimFault(err) {
+		_, ok := soap.ToVimFault(err).(*types.ConcurrentAccess)
+
+		return ok
+	}
+
+	if soap.IsSoapFault(err) {
+		_, ok := soap.ToSoapFault(err).VimFault().(types.ConcurrentAccess)
+
+		// sometimes we get the correct fault but wrong type
```
hickeng: Do you have some examples of this? We should raise a vSphere bug if they're returning the wrong or inconsistent types. If it's happening during govmomi deserialization then we should definitely let @dougm know.

hickeng: Ah - looks like this was a comment by @caglar10ur so we'll just let @dougm know for follow up if it's something he's unaware of.

hickeng:
Updating `pkg/vsphere/test/env` to construct a vcsim environment may be the simplest way to bump coverage here.

pkg/vsphere/vm/vm.go
```diff
@@ -281,7 +282,27 @@ func (vm *VirtualMachine) DeleteExceptDisks(ctx context.Context) (*object.Task,
...
+		return vm.Destroy(ctx)
+	})
+
+	if err == nil || !tasks.IsMethodDisabledError(err) {
+		return info, err
+	}
+
+	// If destroy method is disabled on this VM, re-enable it and retry
+	if tasks.IsMethodDisabledError(err) {
+		err = vm.EnableDestroy(ctx)
```
zjs: Should this follow the `tasks.IsConcurrentAccessError` retry pattern?

pkg/vsphere/tasks/waiter.go
AngieCris: Yes it's an old comment made by @caglar10ur a while back. I didn't see the same behavior he described in the comment and I'm not sure under what conditions the behavior is likely to occur...

pkg/vsphere/vm/vm.go
AngieCris: Different callers of `DeleteExceptDisks` treat `ConcurrentAccess` failure differently - when portlayer calls it to remove cVM, it ignores the `ConcurrentAccess` failure because it shouldn't occur given that we're not setting ChangeVersion. 
https://github.com/vmware/vic/blob/master/lib/portlayer/exec/container.go#L639
The retry on `ConcurrentAccess` error is probably not necessary but since we're removing appliance VM and all of the cVMs, we probably want to make sure everything is gone

zjs: > not enough permission to enable the operation due to some ops-user creds configuration stuff, corrupted VM, etc.

It seems like these examples are all be different errors (not Concurrent Access Error).

If there's just a concurrent access error, why wouldn't we want to just retry?

AngieCris: If `ConcurrentAccess` error is returned from `vm.Destroy`, it makes more sense for the caller to decide what to do (like portlayer ignores it because it's not setting `ChangeVersion`, and vic-machine retries on it)
If `ConcurrentAccess` is returned from `vm.EnableDestroy`, then it probably needs a retry, since we should at least grant us the permission to destroy it first
Fixing it, thanks

PR #7413, opened 2018-02-27
Create a new, unique bridge network on a random VLAN for VC

### LINKED ISSUE #7428: Link alias name resolution failed.
**User Statement:**

Created link alias to a running container was not resolved from container.
```
docker run -ti busybox ping cont7alias 
ping: bad address 'cont7alias'
```

**Details:**
I was not able to reproduce it on my VC setup. Everything worked as expected.

image

[Test-Cases.Group1-Docker-Commands.1-36-Docker-Rename-VCH-17099-4617-container-logs.zip](https://github.com/vmware/vic/files/1769082/Test-Cases.Group1-Docker-Commands.1-36-Docker-Rename-VCH-17099-4617-container-logs.zip)

mhagen-vmware:
[full ci]
Seeing DNS query collisions when multiple VCHs are installed so adding the same logic we had for ESXi for VC.

fixes #7428


tests/resources/VCH-Util.robot
```diff
@@ -76,11 +76,16 @@ Set Test Environment Variables
...
+
+Create Unique Bridge Network
     # Set a unique bridge network for each VCH that has a random VLAN ID
     ${vlan}=  Evaluate  str(random.randint(1, 4093))  modules=random
-    ${out}=  Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Run  govc host.portgroup.add -vlan=${vlan} -vswitch vSwitchLAN %{VCH-NAME}-bridge
-    ${out}=  Run Keyword If  '%{HOST_TYPE}' == 'VC'  Run  govc dvs.portgroup.add -vlan=${vlan} -dvs test-ds %{VCH-NAME}-bridge
-    Set Environment Variable  BRIDGE_NETWORK  %{VCH-NAME}-bridge
+    ${rc}  ${output}=  Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Run And Return Rc And Output  govc host.portgroup.add -vlan=${vlan} -vswitch vSwitchLAN %{VCH-NAME}-${vlan}
+    Run Keyword If  '%{HOST_TYPE}' == 'ESXi'  Should Be Equal As Integers  ${rc}  0
+    ${rc}  ${output}=  Run Keyword If  '%{HOST_TYPE}' == 'VC'  Run And Return Rc And Output  govc dvs.portgroup.add -vlan=${vlan} -dvs test-ds %{VCH-NAME}-${vlan}
```
hickeng: DPG should **not** have the VCH name in it. If it does then we do not benefit from any name based reservation of the VLAN id.

mhagen-vmware: Sure I can continue to tweak this, but you are seriously over engineering a non-existent problem.  The *only* case in which this PR logic kicks in is if within a single build we fail to cleanup previous VCHs and bridges on a previous test *and then* we happen to get the same random vlan number as one of those previously not cleaned up bridges. 

Changing to an indefinite until keyword succeeds will just mask that further until we consume all 4096 vlans.  And we cannot remove the VCH name from the bridge name otherwise we have *no* simple way to find and cleanup leaked bridge networks.

tests/resources/VCH-Util.robot
```diff
@@ -76,10 +76,16 @@ Set Test Environment Variables
     Set Test VCH Name
     # cleanup any potential old certs directories
     Remove Directory  %{VCH-NAME}  recursive=${true}
+    Wait Until Keyword Succeeds  5x  1s  Create Unique Bridge Network
```
hickeng: I'd give this a longer timeout. We're expecting an active failure during the collision path so timeouts are a different matter.

tests/resources/VCH-Util.robot
```diff
@@ -434,10 +440,10 @@ Cleanup VIC Appliance On Test Server
     ${memory}=  Run Keyword And Ignore Error  Get Hostd Memory Consumption
     Log  ${memory}
     Log To Console  Gathering logs from the test server %{VCH-NAME}
-    Gather Logs From Test Server
-    Wait Until Keyword Succeeds  3x  5 seconds  Remove All Containers
+    Run Keyword And Continue On Failure  Gather Logs From Test Server
     # Exit from Cleanup if VCH-NAME is currently in exception list
     ${exclude}=  Check If VCH Is In Exception  vch=%{VCH-NAME}
+    Run Keyword If  ${exclude}  Wait Until Keyword Succeeds  3x  5 seconds  Remove All Containers
```
hickeng: Might want more time here - if containers are still running then we use staged stop (stop signal/sigkill/power off), with the default timeout being 10s between the stages.

Issue #7365, opened 2018-02-22
Portlayer panic github.com/vmware/vic/lib/portlayer/network.(*Endpoint).copy on VC


mhagen-vmware:
The test container is hung on this command:
```
root         1  0.0  0.0   4344   636 ?        Ss   16:01   0:00 /bin/sh -c echo $CI_SCRIPT | base64 -d | /bin/sh -e
root         9  0.0  0.0   4344   688 ?        S    16:01   0:00 /bin/sh -e
root        51  0.0  0.0  11684  2500 ?        S    16:01   0:00 /bin/bash tests/integration-test.sh
root       182  0.2  1.0 142580 82132 ?        S    16:01   0:31 /usr/bin/python /usr/local/bin/pybot --removekeywords TAG:secret --exclude skip tests/test-cases
root      4563  0.0  0.0      0     0 ?        Zs   16:35   0:00 [sh] <defunct>
root     11039  0.0  0.0   4344   732 ?        S    17:15   0:00 sh -c docker -H 10.158.214.77:2376 --tls run --name verify-anon-1 jakedsouza/group-1-19-docker-verify-volume-files:1.0 ls /etc/example 2>&1
root     11040  0.0  0.1 200516 16340 ?        Sl   17:15   0:00 docker -H 10.158.214.77:2376 --tls run --name verify-anon-1 jakedsouza/group-1-19-docker-verify-volume-files:1.0 ls /etc/example
root     11047  0.0  0.0  20304  3308 pts/0    Ss   19:27   0:00 /bin/bash
root     11057  0.0  0.0  17508  2076 pts/0    R+   19:28   0:00 ps aux
```

The container id of concern:

`074b7ed656e57633397df644b879ef4ca317f79a1c907644d663d687854c0326`

Failure - personality:
```
Feb 22 2018 17:18:08.160Z INFO  container attach: stdout (074b7ed656e57633397df644b879ef4ca317f79a1c907644d663d687854c0326) done
Feb 22 2018 17:18:08.160Z INFO  container attach: stderr (074b7ed656e57633397df644b879ef4ca317f79a1c907644d663d687854c0326) done
Feb 22 2018 17:18:08.160Z INFO  cleaned up connections to 074b7ed656e57633397df644b879ef4ca317f79a1c907644d663d687854c0326. Checking errors
Feb 22 2018 17:18:08.160Z INFO  No error found. Returning nil...
[...]

PR #7420, opened 2018-02-28
"Handle event resubscription to PL when it panics" or something that pertains to the fix.

gigawhitlocks:
`[full ci]`

@hickeng @hmahmood 

Related to #7365 but does not resolve it.

### DIFF
--- a/lib/apiservers/engine/backends/eventmonitor.go
+++ b/lib/apiservers/engine/backends/eventmonitor.go
@@ -74,6 +74,7 @@ type PortlayerEventMonitor struct {
 	stop      chan struct{}
 	proxy     eventproxy
 	publisher eventpublisher
+	noRestart bool
 }
 
 // StreamEvents() handles all swagger interaction to the Portlayer's event manager
@@ -115,6 +116,8 @@ func NewPortlayerEventMonitor(proxy eventproxy, publisher eventpublisher) *Portl
 func (m *PortlayerEventMonitor) Start() error {
 	defer trace.End(trace.Begin(""))
 
+	m.noRestart = false
+
 	if m.stop != nil {
 		return fmt.Errorf("Portlayer event monitor: Already started")
 	}
@@ -123,10 +126,13 @@ func (m *PortlayerEventMonitor) Start() error {
 	go func() {
 		var err error
 		for {
-			if err = m.monitor(); err != nil {
-				log.Errorf("XXX %s", err)
+			if m.noRestart {
+				break
 			}
-			log.Debug("XXX Restarting Portlayer event monitor")
+			if err = m.monitor(); err != nil {
+				log.Error(err)
+			}
+			log.Debug("Restarting Portlayer event monitor")
 			m.stop = make(chan struct{})
 		}
 	}()
@@ -137,6 +143,7 @@ func (m *PortlayerEventMonitor) Start() error {
 func (m *PortlayerEventMonitor) Stop() {
 	defer trace.End(trace.Begin(""))
 
+	m.noRestart = true
 	if m.stop != nil {
 		close(m.stop)
 	}
--- a/lib/apiservers/engine/backends/eventmonitor.go
+++ b/lib/apiservers/engine/backends/eventmonitor.go
@@ -130,8 +130,7 @@ func (m *PortlayerEventMonitor) Start() error {
 				if err = m.monitor(); err != nil {
 					log.Error(err)
 				}
-				log.Debug("Restarting Portlayer event monitor")
-				m.stop = make(chan struct{})
+				log.Warn("Restarting Portlayer event monitor")
 			}
 		}
 	}()
--- a/lib/apiservers/engine/backends/eventmonitor.go
+++ b/lib/apiservers/engine/backends/eventmonitor.go
@@ -125,12 +125,12 @@ func (m *PortlayerEventMonitor) Start() error {
 		for {
 			select {
 			case <-m.stop:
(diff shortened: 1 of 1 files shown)


matthewavery:
I am not well versed on this yet, but why would the endpoints be missing? it is odd, but I have not tried this on my local VC. Otherwise I like the change for sure!

matthewavery:
Upon reading the ticket I think  there are other issues surrounding the run failure. I think this will prevent the PL from crashing and restarting which will keep us from dropping the stream handler. But the other issues that @hickeng has detailed could still be an issue. That being said, I think this is a solid improvement from the current behavior and should likely help to keep things from breaking in complex ways.

LGTM - for the current changes

lib/portlayer/network/context.go
```diff
@@ -940,7 +940,13 @@ func (c *Context) RemoveIDFromScopes(op trace.Operation, id string) ([]*Endpoint
 		s := ne.scope
 
 		// save the endpoint info
-		e := con.Endpoint(s).copy()
+		e := con.Endpoint(s)
```
hmahmood: I think there is something else going on here. This should not be `nil`. I would say we need to dig deeper; this just works around a bug in the code somewhere else.

gigawhitlocks: Well it is possible for `Endpoint(foo)` to return `nil`, which means we do have to handle that case, and we're not doing that here. I think you're right that there is likely more going on, but this would at least allow the PL to not panic & return a sane error along this code path. I'm still digging into it -- I think it's likely that this is not the whole fix, but I do think it's part of the fix at least.

hmahmood: I mean I am fine with you putting this in to debug the problem, but I don't want to merge it like this. In this code path, it looks like this should not be nil. There is some race condition or some other assumption being made here that may or may not be correct. I just want us to figure that out first.

hmahmood: Just looking at this code, I see that we are iterating over con.endpoints without acquiring a lock on con. con.Endpoint does acquire the lock. So we should at least fix that.

lib/portlayer/network/context.go
```diff
@@ -940,7 +940,13 @@ func (c *Context) RemoveIDFromScopes(op trace.Operation, id string) ([]*Endpoint
...
+
+		if e == nil {
```
hickeng: panic'ing in this circumstance is the correct thing to do. It's indicative of a violation of our core assumptions.

There are two things that should occur given we're seeing cascade failures from this:
1. fix the root cause
2. fix the fact that the personality doesn't handle portlayer restarts

To address part of (2) we can/should just have the personality re-establish the event stream if it dies. That's partial.
The more complex part is re-validating cache state, etc which should occur (I hate multi-level caches without true invalidation/revalidation schemes)

gigawhitlocks:
Just closing this because I don't believe my approach(es) here are correct

lib/apiservers/engine/backends/eventmonitor.go
```diff
@@ -120,8 +120,21 @@ func (m *PortlayerEventMonitor) Start() error {
...
+		var err error
+		for {
+			select {
+			case <-m.stop:
+				break
+			default:
+				if err = m.monitor(); err != nil {
+					log.Error(err)
+				}
+				log.Debug("Restarting Portlayer event monitor")
```
hickeng: I'd actually make this a `Warn` - it's not something that should happen under normal circumstances and is an indication that we may have missed significant state transitions.

lib/apiservers/engine/backends/eventmonitor.go
```diff
@@ -120,8 +120,21 @@ func (m *PortlayerEventMonitor) Start() error {
...
+				m.stop = make(chan struct{})
```
hickeng: no - this races with `m.Stop()` closing the channel that was assigned to this previously.
The m.stop channel should still be happily open when `m.monitor` exits.

lib/apiservers/engine/backends/eventmonitor.go
```diff
@@ -120,8 +120,20 @@ func (m *PortlayerEventMonitor) Start() error {
...
+		var err error
+		for {
+			select {
+			case <-m.stop:
+				break
+			default:
+				if err = m.monitor(); err != nil {
+					log.Error(err)
+				}
+				log.Warn("Restarting Portlayer event monitor")
```
hickeng: Will be logged even if we're not going to restart. Perhaps we should just combine this with the error logging.

lib/apiservers/engine/backends/eventmonitor.go
```diff
@@ -120,8 +120,20 @@ func (m *PortlayerEventMonitor) Start() error {
...
 
 	m.stop = make(chan struct{})
-	go m.monitor()
-
+	go func() {
+		var err error
+		for {
+			select {
+			case <-m.stop:
+				break
```
hmahmood: Should an info log here for exiting cleanly.

lib/apiservers/engine/backends/eventmonitor.go
gigawhitlocks: oh right, didn't think about it after we restructured the loop.

sflxn:
You should retitle this PR and also follow the PR comment standards.  Resolution number in separate line at the bottom.  Retitle this because this doesn't "prevent PL panic".  "Handle event resubscription to PL when it panics" or something that pertains to the fix.

PR #7423, opened 2018-02-28
Delete datastore files recursively for all datastore types

### LINKED ISSUE #7359: vic-machine-{os} delete fails to cleanup properly if there are pulled images
vic-machine-{os} delete fails to cleanup properly if there are pulled images.

The container VMs and volumes are removed, but the images are not deleted and the respective folder on the datastore remains. All files related to the VCH deployment are also not cleaned up.
If the VCH is created and deleted without pulling any images, everything is cleaned up properly.
The delete commad output is normal, but running it with the  **--debug 3** option produces:
```
DEBU[0013] Failed to delete "[datastore0] virtual-container-host/VIC": Cannot delete file [datastore0] virtual-container-host/VIC
DEBU[0017] Image store parent directory not empty, leaving in place.
```

This happens if the VCH is created on a vCenter server, i have not tried it on just an ESX.
Trying to delete the folder from the vSphere web client manually fails with:
"Cannot delete file [datastore0] virtual-container-host"
All other files related to the VCH can be deleted, but the image files cannot.
SSH-ing to the ESX and deleting the folder manually succeeds

The vic-machine version is: v1.4.0-dev-0-8d92ff7

Steps to reproduce:
1. Create a VCH in a vCenter server. I used the following command:
```
./vic-machine-linux create --target {vcenter-ip} --user {username} --password {password} --bridge-network vch-bridge --image-store datastore0 --volume-store datastore0/vol
umes:default --no-tlsverify --thumbprint {vcenter-cert-thumbprint} --name virtual-container-host
```

2. With a docker client execute:
```
docker --tls -H {VCH-IP}:2376 pull alpine
```

3. Delete the VCH. I used the following command:
```
./vic-machine-linux delete --target {vcenter-ip} --user {username} --password {password} --force --name virtual-container-host
```

AngieCris:
`[specific ci=6-03-Delete --suite 1-02-Docker-Pull]`

Fixes #7359 

For non-vSAN type datastores, sometimes it complains when we try to delete non-empty datastore folders with child image files, causing trailing VCH datastore folders after vic-machine delete finishes.

Revert of PR: https://github.com/vmware/vic/pull/6951

### DIFF
--- a/lib/install/management/store_files.go
+++ b/lib/install/management/store_files.go
@@ -43,6 +43,7 @@ func (d *Dispatcher) deleteImages(conf *config.VirtualContainerHostConfigSpec) e
 	d.op.Info("Removing image stores")
 
 	for _, imageDir := range conf.ImageStores {
+
 		imageDSes, err := d.session.Finder.DatastoreList(d.op, imageDir.Host)
 		if err != nil {
 			errs = append(errs, err.Error())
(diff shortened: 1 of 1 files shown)


hickeng:
@AngieCris I suggest you drop this PR and create a new one using `git revert`. That will generate a standard commit message and makes it obvious that it's a revert of a prior commit.

PR #7424, opened 2018-02-28
Build non-stripped binaries when env VIC_DEBUG_BUILD is set

lcastellano:
When VIC_DEBUG_BUILD is set, the generated binaries are non-stripped and contain DWARF debug information.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,11 @@ define godeps
 endef
 endif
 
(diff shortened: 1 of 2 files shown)


gigawhitlocks:
Will this allow us to attach a debugger to the various components?

lcastellano:
Yes, the idea is to attach dlv headless to our servers running inside the VCH VM. I am working on a couple of shell script that use ssh to do actions on the VCH. The first one copies the environment needed for dlv to run. The second one remotely lunches dlv headless to attach the desired process (persona, port-layer). You can then attach from the goland to the running process inside the VCH.

PR #7429, opened 2018-02-28
Fix group 6 delete tests

### LINKED ISSUE #7427: CI fail after moving to VC: 6-03-Delete fails in govc vm.create: resource pool not found
Build: https://ci-vic.vmware.com/vmware/vic/17109/8
and also build: https://ci-vic.vmware.com/vmware/vic/17117/8
test log: https://console.cloud.google.com/m/cloudstorage/b/vic-ci-logs/o/integration_logs_17109_af224d97e2de01a92dd8cd461b87df080a4ab300.zip?authuser=1

Test suite: 6-03-Delete
Test cases:
1. Delete VCH with non-cVM in same RP
2. Delete VCH moved from its RP

command: `govc vm.create -pool="/dc1/host/cls/VCH-17109-2932" -net=vm-network -on=false anothervm-I3qf59MEGyuxTjy 2>&1`
output: `govc: resource pool '/dc1/host/cls/VCH-17109-2932' not found`
expected: vm.create should succeed

This test succeeds on my nimbus VC but fails on CI.

mdharamadas1:
[specific ci=6-03-Delete]

fixes #7427

### DIFF
--- a/.drone.local.yml
+++ b/.drone.local.yml
@@ -40,5 +40,6 @@ pipeline:
       - 'export GOVC_DATASTORE=${TEST_DATASTORE}'
       - 'export DOMAIN=${DOMAIN}'
       - tests/integration-test.sh
+      # - pybot -t "Delete VCH with non-cVM in same RP" tests/test-cases/Group6-VIC-Machine/6-03-Delete.robot
     volumes:
       - /tmp
--- a/tests/test-cases/Group6-VIC-Machine/6-03-Delete.robot
+++ b/tests/test-cases/Group6-VIC-Machine/6-03-Delete.robot
@@ -153,13 +153,13 @@ Delete VCH moved from its RP
 
     Install VIC Appliance To Test Server
 
-    Set Test Variable  ${pool}  "%{TEST_RESOURCE}/Resources/%{VCH-NAME}"
+    Set Test Variable  ${test-resource}  "%{TEST_RESOURCE}/Resources"
 
     ${rand}=  Generate Random String  15
     ${dummyvm}=  Set Variable  anothervm-${rand}
     Set Suite Variable  ${tempvm}  ${dummyvm}
-    Log To Console  Create VM ${dummyvm} in ${pool} net %{PUBLIC_NETWORK}
-    ${rc}  ${output}=  Run And Return Rc And Output  govc vm.create -pool=${pool} -net=%{PUBLIC_NETWORK} -on=false ${dummyvm}
+    Log To Console  Create VM ${dummyvm} in ${test-resource}/%{VCH-NAME} net %{PUBLIC_NETWORK}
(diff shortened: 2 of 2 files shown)


.drone.local.yml
```diff
@@ -40,5 +40,6 @@ pipeline:
       - 'export GOVC_DATASTORE=${TEST_DATASTORE}'
       - 'export DOMAIN=${DOMAIN}'
       - tests/integration-test.sh
+      # - pybot -t "Delete VCH with non-cVM in same RP" tests/test-cases/Group6-VIC-Machine/6-03-Delete.robot
```
mdharamadas1: Intentionally left here to help with running single test locally.

anchal-agrawal:
The format for using specific ci here would be `[specific ci=Group6-VIC-Machine]`. For running just `6-03-Delete.robot` we can do `[specific ci=6-03-Delete]`. After changing the PR description, another push to this branch is required to trigger a new build.

hickeng:
@anchal-agrawal Doesn't need to be in the header any more! you can have it in the PR body, and then you don't need a new push.

tests/test-cases/Group6-VIC-Machine/6-03-Delete.robot
```diff
@@ -152,8 +156,8 @@ Delete VCH moved from its RP
     ${rand}=  Generate Random String  15
     ${dummyvm}=  Set Variable  anothervm-${rand}
     Set Suite Variable  ${tempvm}  ${dummyvm}
-    Log To Console  Create VM ${dummyvm} in %{TEST_RESOURCE}/%{VCH-NAME} net %{PUBLIC_NETWORK}
-    ${rc}  ${output}=  Run And Return Rc And Output  govc vm.create -pool="%{TEST_RESOURCE}/%{VCH-NAME}" -net=%{PUBLIC_NETWORK} -on=false ${dummyvm}
+    Log To Console  Create VM ${dummyvm} in %{TEST_RESOURCE}/Resources/%{VCH-NAME} net %{PUBLIC_NETWORK}
```
hickeng: should we be using the `${pool}` variable here instead of the explicit path? That said, the `Resources` pool exists on ESXs as well, so another possibility is setting `TEST_RESOURCE` so that consistent path construction can be used irrespective of ESX or VC.

mdharamadas1: Yes, I will add `${pool}` var but `TEST_RESOURCE` is a global var, so if I change / update it , won't it impact existing tests which are passing? @mhagen-vmware ?

mhagen-vmware: for now, let's just do the pool variable and please create a follow up issue to define the TEST_RESOURCE properly throughout all the code.

mdharamadas1: Cool, I added https://github.com/vmware/vic/issues/7434 to take of it later.

Issue #7279, opened 2018-02-07
VM placement algorithm


cgtexmex:
When DRS is not available in a vSphere cluster vic will need to specify the host for VM placement.  That host recommendation will _not_ be attempting to identify the best host, but simply to avoid picking a _wrong_ host.  

An algorithm will be created that will make the recommendation based on identified inputs.  This will require tight collaboration with #7327  

- [x] document design
- [x] gather host metrics for the cluster
- [x] sort hosts based on weighted metrics
- [x] identify least _loaded_ host based on weighted metrics
- [x] unit tests

hickeng:
@dougm FYI vcsim

jzt:
Pretty much done (in verify), but will likely be tweaked over time as we learn during testing.

PR #7433, opened 2018-03-01
Initial rudimentary implementation of vm placement library

jzt:
This change introduces VM placement engine, and the `MetricsProvider` interface. Existing host metrics code has been refactored to implement this interface. 

The design is such that a `MetricsProvider` can be used as the source of metrics data used by the placement engine in order to make a decision as to the host on which a newly created VM should be placed. This is a stepping-stone change that does not yet implement decision logic based on metric data, but instead randomly chooses a host from the compute resource and returns it. Also included are some rudimentary first steps toward host scoring/ranking. The purpose of pushing this placement library skeleton code is so that development that depends on this interface will not be blocked.

Further changes will include host ranking, pre-emptive fast-path checking to see whether or not the host that the VM is created on is adequate for power-on, vm migration in the event that the initial host fails this check, and plumbing together with the container create/start paths.

Towards #7279.

### DIFF
--- a/pkg/vsphere/compute/placement/placement.go
+++ b/pkg/vsphere/compute/placement/placement.go
@@ -1,0 +1,107 @@
+// Copyright 2018 VMware, Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package placement
+
+import (
+	"fmt"
+	"math/rand"
+	"sort"
+
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/compute"
+	"github.com/vmware/vic/pkg/vsphere/performance"
+	"github.com/vmware/vic/pkg/vsphere/vm"
+)
+
+const (
+	memUnconsumedWeight = 0.7 // available memory (total - consumed)
+	memInactiveWeight   = 0.3 // active memory on the host
+)
+
+type rankedHost struct {
+	*object.HostSystem
+	*performance.HostMetricsInfo
+	score float64
+}
+
+func (r *rankedHost) String() string {
+	return fmt.Sprintf("%s:%f", r.HostSystem.InventoryPath, r.score)
+}
+
+type rankedHosts []rankedHost
+
+func (r rankedHosts) Len() int           { return len(r) }
+func (r rankedHosts) Swap(i, j int)      { r[i], r[j] = r[j], r[i] }
+func (r rankedHosts) Less(i, j int) bool { return r[i].score > r[j].score }
+
+// Engine uses data from a MetricsProvider to decide on which host to powerOn a VM.
+type Engine struct {
+	source performance.MetricsProvider
+}
+
+// NewEngine returns an Engine instance using the supplied MetricsProvider.
+func NewEngine(s performance.MetricsProvider) *Engine {
+	return &Engine{source: s}
+}
+
+// CheckHost returns true if the host has adequate capacity to power on the VM, false otherwise.
+func (e *Engine) CheckHost(op trace.Operation, host *object.HostSystem) bool {
+	// TODO(jzt): return false until we have host checking logic decided
+	return false
+}
+
+// RecommendHost recommends an ideal host on which to place a newly created VM.
+func (e *Engine) RecommendHost(op trace.Operation, vm *vm.VirtualMachine) (*object.HostSystem, error) {
+	// TODO(jzt): randomize placement initially to allow usage of this interface for development
+	// towards other ROBO-related issues.
+	r, err := vm.ResourcePool(op)
+	if err != nil {
+		return nil, err
+	}
+
+	rp := compute.NewResourcePool(op, vm.Session, r.Reference())
+
+	cls, err := rp.GetCluster(op)
+	if err != nil {
+		return nil, err
+	}
+
+	hosts, err := cls.Hosts(op)
+	if err != nil {
+		return nil, err
+	}
+
+	return hosts[rand.Intn(len(hosts))], nil
+}
+
+func rankHosts(op trace.Operation, hm map[*object.HostSystem]*performance.HostMetricsInfo) []rankedHost {
+	ranking := []rankedHost{}
+	for h, m := range hm {
+		rh := rankedHost{
+			HostSystem:      h,
+			HostMetricsInfo: m,
+			score:           rankMemory(m) * (1 - m.CPU.UsagePercent),
+		}
+		ranking = append(ranking, rh)
+	}
+	sort.Sort(rankedHosts(ranking))
+	return ranking
+}
+
+func rankMemory(hm *performance.HostMetricsInfo) float64 {
+	free := float64(hm.Memory.TotalKB-hm.Memory.ConsumedKB) / 1024.0
+	return free * memUnconsumedWeight
+}
--- a/pkg/vsphere/compute/placement/placement.go
+++ b/pkg/vsphere/compute/placement/placement.go
@@ -17,21 +17,8 @@ package placement
 import (
 	"github.com/vmware/govmomi/object"
 	"github.com/vmware/vic/pkg/trace"
-	"github.com/vmware/vic/pkg/vsphere/performance"
 	"github.com/vmware/vic/pkg/vsphere/vm"
 )
-
-type rankedHost struct {
-	HostReference string
-	*performance.HostMetricsInfo
-	score float64
-}
-
-type rankedHosts []rankedHost
-
-func (r rankedHosts) Len() int           { return len(r) }
-func (r rankedHosts) Swap(i, j int)      { r[i], r[j] = r[j], r[i] }
-func (r rankedHosts) Less(i, j int) bool { return r[i].score > r[j].score }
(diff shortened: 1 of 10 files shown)


pkg/vsphere/performance/host_metrics.go
```diff
@@ -64,17 +63,41 @@ type HostMetricsInfo struct {
...
+
+	if hosts == nil {
+		return nil, fmt.Errorf("no hosts found in compute resource")
+	}
+
+	return h.GetMetricsForHosts(op, hosts)
+}
+
+// GetMetricsForHosts returns metrics pertaining to supplied ESX hosts.
+func (h *HostMetricsProvider) GetMetricsForHosts(op trace.Operation, hosts []*object.HostSystem) (map[*object.HostSystem]*HostMetricsInfo, error) {
```
anchal-agrawal: Since this func is exported and `hosts` is now a parameter, we should add a `len(hosts) == 0` check and bail early if true.

pkg/vsphere/performance/host_metrics.go
```diff
@@ -64,17 +63,41 @@ type HostMetricsInfo struct {
 	CPU    HostCPU
 }
 
-// HostMetrics returns CPU and memory metrics for all ESXi hosts in the input session's cluster.
-func HostMetrics(op trace.Operation, session *session.Session) (map[*object.HostSystem]*HostMetricsInfo, error) {
-	if session == nil {
+// HostMetrics returns CPU and memory metrics for all ESXi hosts in a cluster
+// via implementation of the MetricsProvider interface.
+type HostMetricsProvider struct {
+	Session *session.Session
```
anchal-agrawal: Should we export `Session` if the caller can supply it in  `NewHostMetricsProvider()`?

pkg/vsphere/performance/host_metrics.go
jzt: 👍 fixed

pkg/vsphere/performance/host_metrics.go
jzt: No, we shouldn't. Good catch, that was left over from a previous iteration where I was accessing the session from outside of the package before I had the design figured out. Fixed!

pkg/vsphere/performance/metrics_provider.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/vic/pkg/trace"
+)
+
+// MetricsProvider defines the interface for providing metrics.
+type MetricsProvider interface {
+	// GetMetricsForComputeResource returns metrics for a particular compute resource
+	GetMetricsForComputeResource(trace.Operation, *object.ComputeResource) (map[*object.HostSystem]*HostMetricsInfo, error)
```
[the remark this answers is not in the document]
jzt: Fixed, I guess it will be up to the caller to convert back from `host.Reference()` to `*object.HostSystem`?

pkg/vsphere/compute/placement/placement.go
```diff
@@ -0,0 +1,107 @@
...
+	source performance.MetricsProvider
+}
+
+// NewEngine returns an Engine instance using the supplied MetricsProvider.
+func NewEngine(s performance.MetricsProvider) *Engine {
+	return &Engine{source: s}
+}
+
+// CheckHost returns true if the host has adequate capacity to power on the VM, false otherwise.
+func (e *Engine) CheckHost(op trace.Operation, host *object.HostSystem) bool {
```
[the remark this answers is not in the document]
jzt: @hickeng I like the idea of using input types from the same package, but if we switch to `*object.VirtualMachine`, we lose access to the `session.Session` which is required to find the compute resource. Any way around that?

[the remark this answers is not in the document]
jzt: I've gone ahead and defined the `HostPlacementPolicy` interface, and implemented it in the form of `RandomHostPolicy`. I've left the input types as they are until a better design is discovered around how we obtain the compute resource via the `Session`.

After a discussion with @hickeng, we've decided to merge this as is, and further work addressing the weighting configuration and ranking algorithm, as well as tests using the VC simulator, will come in a followup PR.

pkg/vsphere/performance/metrics_provider.go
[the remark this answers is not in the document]
hickeng: The other option is it's a map that's only used for iteration, with no intent of simple access via a known key. Either could work - but moref is likely to result in less confusion.

pkg/vsphere/compute/placement/placement.go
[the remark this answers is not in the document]
hickeng: I don't really like the placement logic needing to hold a `Session` just in order to get the cluster either. Correct answer is probably to move that call into govmomi so it can use the object.VirtualMachine.client. Keep it as `vm.VirtualMachine` for now.

pkg/vsphere/performance/host_metrics.go
```diff
@@ -64,22 +63,51 @@ type HostMetricsInfo struct {
...
+type HostMetricsProvider struct {
+	session *session.Session
+}
+
+// NewHostMetricsProvider returns a new instance of HostMetricsProvider.
+func NewHostMetricsProvider(s *session.Session) *HostMetricsProvider {
+	return &HostMetricsProvider{session: s}
+}
+
+// GetMetricsForComputeResource gathers host metrics from the supplied compute resource.
```
hickeng: Comment should include what the map is keyed by.

pkg/vsphere/compute/placement/placement_policy.go
```diff
@@ -0,0 +1,44 @@
...
+
+type rankedHost struct {
+	HostReference string
+	*performance.HostMetricsInfo
+	score float64
+}
+
+type rankedHosts []rankedHost
+
+func (r rankedHosts) Len() int           { return len(r) }
```
hickeng: Why are these exported? Doesn't look like it's necessary.

pkg/vsphere/compute/placement/placement_test.go
```diff
@@ -0,0 +1,163 @@
+// Copyright 2018 VMware, Inc. All Rights Reserved.
```
hickeng: Think this file should be named `ranked_test.go`

pkg/vsphere/compute/placement/placement_policy.go
```diff
@@ -0,0 +1,44 @@
...
+func (r rankedHosts) Swap(i, j int)      { r[i], r[j] = r[j], r[i] }
+func (r rankedHosts) Less(i, j int) bool { return r[i].score > r[j].score }
+
+// HostPlacementPolicy defines the interface for using metrics to decide the appropriate host
+// for a VM based on host metrics and VM provisioned resources.
+type HostPlacementPolicy interface {
```
hickeng: Consider splitting this interface into a file called `placement.go` and the ranked definitions into a `ranked.go` along with the ranking logic.

pkg/vsphere/compute/placement/random.go
```diff
@@ -0,0 +1,90 @@
...
+
+	hosts, err := cls.Hosts(op)
+	if err != nil {
+		return nil, err
+	}
+
+	return hosts[rand.Intn(len(hosts))], nil
+}
+
+func rankHosts(op trace.Operation, hm map[string]*performance.HostMetricsInfo) []rankedHost {
```
hickeng: These rank functions don't belong in `random.go`

pkg/vsphere/compute/placement/placement_policy.go
jzt: I think for `sort.Sort`?

pkg/vsphere/compute/placement/placement_policy.go
jzt: You read my mind. :) I've created `random.go`, `ranked.go` and `ranked_test.go`, and renamed `placement_policy.go` -> `placement.go`.

pkg/vsphere/compute/placement/random.go
jzt: Caught me before I could push a cleanup commit. Fixed!

jzt:
Added a vcsim test for `RandomHostPolicy`.

pkg/vsphere/compute/placement/ranked_test.go
```diff
@@ -0,0 +1,141 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/govmomi/vim25/types"
+	"github.com/vmware/vic/pkg/trace"
+	"github.com/vmware/vic/pkg/vsphere/performance"
+
+	units "github.com/docker/go-units"
```
anchal-agrawal: Minor: we should combine this import with the group for `"github.com/stretchr/testify/assert"`

pkg/vsphere/compute/placement/random_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/vmware/vic/pkg/vsphere/extraconfig"
+	"github.com/vmware/vic/pkg/vsphere/session"
+	"github.com/vmware/vic/pkg/vsphere/tasks"
+	"github.com/vmware/vic/pkg/vsphere/test"
+	"github.com/vmware/vic/pkg/vsphere/vm"
+)
+
+// vpxModelSetup creates a VPX model, starts its server and populates the session. The caller must
+// clean up the model and the server once it is done using them.
+func vpxModelSetup(ctx context.Context, t *testing.T) (*simulator.Model, *simulator.Server, *session.Session) {
```
anchal-agrawal: We could consider moving this func to `pkg/vsphere/test` since it's present in several test files now.

jzt: Yep, that's on my TODO list actually! I'll be adding more vcsim tests for the other policies, once I do I'll move it in there.

pkg/vsphere/compute/placement/ranked_test.go
jzt: Hmm..wonder why gofmt didn't complain. Fixed.

Issue #7275, opened 2018-02-07
vic-machine to provide license and feature check


cgtexmex:
vic-machine create facilitates a license check to ensure VCH compliance.  This will need to be augmented to include the feature checks that will be identified in #7278.  

- [x] vic-machine create ensures license and feature compliance
- [ ] CI Test Plan
- [ ] CI Tests

hickeng:
@dougm Any chance of a "ROBO" license for vcsim?

dougm:
You can call `LicenseManager.AddLicense` to add one, see the `EvalLicense` for an example structure: https://github.com/vmware/govmomi/blob/81743157fb5ccf548d6bd5088c25ee6156a359ee/simulator/license_manager.go#L40-L61

Tests can also modify `simulator.EvalLicense` directly, to change the `EditionKey`, `Properties`, etc.  I've never seen a ROBO license before, so not sure how exactly it would look.

cgtexmex:
If DRS is disabled then there are no resource pools and the following flags must be ignored:
```
   --memory value, --mem value                      VCH resource pool memory limit in MB (unlimited=0) (default: <nil>)
   --memory-reservation value, --memr value         VCH resource pool memory reservation in MB (default: <nil>)
   --memory-shares value, --mems value              VCH resource pool memory shares in level or share number, e.g. high, normal, low, or 163840 (default: <nil>)
   
   --cpu value                                      VCH resource pool vCPUs limit in MHz (unlimited=0) (default: <nil>)
   --cpu-reservation value, --cpur value            VCH resource pool reservation in MHz (default: <nil>)
   --cpu-shares value, --cpus value                 VCH VCH resource pool vCPUs shares, in level or share number, e.g. high, normal, low, or 4000 (default: <nil>)
```
We will warn the user that they are not applicable to the current configuration and ignore them.

Issue #7032, opened 2018-01-04
vic-machine create / delete supports folders


matthewavery:
**User Statement:**
See Parent Issue.

**Details:**
When creating a VCH we should create a Folder (that will display in the VMs & Folders view), named for the VCH by default as with the resource pool. The endpointVM and all containerVMs it creates should be placed in this folder.

Decision needed - how is the Folder specified in CLI arguments?
1. as part of the Path element of the `--target` URL
 2. as a separate `--folder` option

Irrespective of this choice the semantics should be the same as for `--compute-resource` in that a folder will be created _under_ the specified folder with a name corresponding to the VCH.

**Acceptance Criteria:**
- [x] A folder is created, named as the VCH, _under_ the folder specified
- [x] The endpointVM is present in the folder
- [x] All containerVMs are removed from the folder as part of delete

hickeng:
I'm in favour of a separate argument. `target` was limited to only specifying the datacenter and not the compute resource path specifically because the datacenter is the top level namespace mechanism that controls the scoping of all other arguments.
Inventory folders have no impact on whether a value for other options is valid or not so shouldn't be part of the target.

cgtexmex:
In a ROBO environment the `folder` flag cannot be optional -- we _must_ place the VCH and resulting containerVMs in a folder.  Do we want a default value?  Or do we simply make it required for a ROBO environment?

cc @pdaigle

cgtexmex:
For this initial release we will **not** provide a `vic-machine` flag for the folder creation -- instead we will create the folder in the `vm folders` directory of VC with the same name as the VCH.  Then as containerVMs are created we will place those VMs in the same folder.  In the future we can iterate on this implementation and provide more flexibility.

cc @matthewavery @pdaigle

PR #7440, opened 2018-03-02
Basic vic-machine ROBO support

cgtexmex:
Provides vic-machine create/delete functionality for clusters that have DRS disabled.  In a future refinement support will be limited to robo advanced licenses.  This unblocks other work dependent on ROBO support.  Changes will be required for inventory folder support and to complete the license and feature check.   

Toward #7275

### DIFF
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -113,7 +113,6 @@ func (d *Dispatcher) checkExistence(conf *config.VirtualContainerHostConfigSpec,
 	defer trace.End(trace.Begin(""))
 
 	var err error
-	d.vchPoolPath = path.Join(settings.ResourcePoolPath, conf.Name)
 	var orp *object.ResourcePool
 	if orp, err = d.findResourcePool(d.vchPoolPath); err != nil {
 		return err
--- a/lib/install/management/create.go
+++ b/lib/install/management/create.go
@@ -45,6 +45,15 @@ func (d *Dispatcher) CreateVCH(conf *config.VirtualContainerHostConfigSpec, sett
 
 	var err error
 
+	// resource pool path determined based on DRS setting.  If enabled then
+	// append the appliance name to the path and a pool will be created.
+	// disabled then no resource pools are supported, so use the provided compute path.
+	if d.session.DRSEnabled != nil && *d.session.DRSEnabled {
+		d.vchPoolPath = path.Join(settings.ResourcePoolPath, conf.Name)
+	} else {
+		d.vchPoolPath = settings.ResourcePoolPath
+	}
+
 	if err = d.checkExistence(conf, settings); err != nil {
 		return err
 	}
--- a/lib/install/management/resource_pool.go
+++ b/lib/install/management/resource_pool.go
@@ -17,7 +17,6 @@ package management
 import (
 	"context"
 	"fmt"
-	"path"
 
 	"github.com/vmware/govmomi/find"
 	"github.com/vmware/govmomi/object"
@@ -33,8 +32,6 @@ import (
 
 func (d *Dispatcher) createResourcePool(conf *config.VirtualContainerHostConfigSpec, settings *data.InstallerData) (*object.ResourcePool, error) {
 	defer trace.End(trace.Begin("", d.op))
-
-	d.vchPoolPath = path.Join(settings.ResourcePoolPath, conf.Name)
 
 	rp, err := d.session.Finder.ResourcePool(d.op, d.vchPoolPath)
 	if err != nil {
--- a/lib/install/validate/config.go
+++ b/lib/install/validate/config.go
@@ -22,7 +22,6 @@ import (
 	"github.com/vmware/govmomi/govc/host/esxcli"
 	"github.com/vmware/govmomi/license"
 	"github.com/vmware/govmomi/object"
-	"github.com/vmware/govmomi/vim25/mo"
 	"github.com/vmware/govmomi/vim25/soap"
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/lib/config"
@@ -517,22 +516,9 @@ func (v *Validator) checkEvalLicense(op trace.Operation, licenses []types.Licens
 	}
 }
 
-// isStandaloneHost checks if host is ESX or vCenter with single host
-func (v *Validator) isStandaloneHost() bool {
-	cl := v.Session.Cluster.Reference()
-
-	if cl.Type != "ClusterComputeResource" {
-		return true
-	}
-	return false
-}
-
-// drs checks that DRS is enabled
-func (v *Validator) CheckDrs(ctx context.Context) {
-	if v.DisableDRSCheck {
-		return
-	}
-	op := trace.FromContext(ctx, "CheckDrs")
+// CheckDRS will validate DRS settings
+func (v *Validator) CheckDRS(ctx context.Context) {
+	op := trace.FromContext(ctx, "CheckDRS")
 	defer trace.End(trace.Begin("", op))
 
 	errMsg := "DRS check SKIPPED"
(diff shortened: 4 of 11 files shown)


lib/install/management/appliance.go
```diff
@@ -113,7 +113,12 @@ func (d *Dispatcher) checkExistence(conf *config.VirtualContainerHostConfigSpec,
 	defer trace.End(trace.Begin(""))
 
 	var err error
-	d.vchPoolPath = path.Join(settings.ResourcePoolPath, conf.Name)
+	// TODO: Remove when inventory folder support complete #7032
```
matthewavery: Duly noted. If you merge before me then I will make a note to do this.

lib/install/management/resource_pool.go
```diff
@@ -34,8 +34,12 @@ import (
 func (d *Dispatcher) createResourcePool(conf *config.VirtualContainerHostConfigSpec, settings *data.InstallerData) (*object.ResourcePool, error) {
 	defer trace.End(trace.Begin("", d.op))
 
-	d.vchPoolPath = path.Join(settings.ResourcePoolPath, conf.Name)
-
+	// TODO: Remove when inventory folder support complete #7032
```
matthewavery: ditto

anchal-agrawal:
> So the only way to delete via a cluster without DRS enabled is via the VM id -- (i.e. moref).

As I understand we will (for now) be unable to delete a VCH by name on a DRS-disabled setup. Will we make `vic-machine delete` changes in a future PR towards #7275, or is there a separate issue for it?

cgtexmex:
@matthewavery actually I don't know that those `rp` changes will need to be removed -- even w/inventory folders we'll still need `rp` info....to create in some cases and simply reference in others.  Regardless, those changes will need attention...

cgtexmex:
@anchal-agrawal yes..for now no delete by name if DRS is disabled.  I think the changes either fall in #7275 or #7032 -- since we are effectively saying we won't be creating / removing RPs for ROBO clusters I'm leaning toward #7275 ... it has little to do with Inventory Folders and more to do w/no RPs in DRS disabled clusters...

lib/install/validate/config.go
```diff
@@ -524,11 +524,8 @@ func (v *Validator) isStandaloneHost() bool {
 	return false
 }
 
-// drs checks that DRS is enabled
-func (v *Validator) CheckDrs(ctx context.Context) {
-	if v.DisableDRSCheck {
-		return
-	}
+// CheckDrs will validate DRS settings
+func (v *Validator) CheckDrs(ctx context.Context, conf *config.VirtualContainerHostConfigSpec) {
```
anchal-agrawal: s/CheckDrs/CheckDRS

hickeng:
There is a `skipDRSCheck` iirc which should be used for the delete path. In fact, most of the validation is irrelevant for the delete path.

lib/install/validate/config.go
```diff
@@ -524,11 +524,8 @@ func (v *Validator) isStandaloneHost() bool {
 	return false
 }
 
-// drs checks that DRS is enabled
-func (v *Validator) CheckDrs(ctx context.Context) {
-	if v.DisableDRSCheck {
-		return
-	}
+// CheckDrs will validate DRS settings
```
anchal-agrawal: We should mention here that this func modifies `conf` if DRS is disabled. This would be helpful since the func name `CheckDRS` doesn't convey that it modifies an input parameter.

lib/config/virtual_container_host.go
```diff
@@ -124,6 +124,8 @@ type Container struct {
 	ContainerNameConvention string
 	// Permitted datastore URLs for container storage for this virtual container host
 	ContainerStores []url.URL `vic:"0.1" scope:"read-only" recurse:"depth=0"`
+	// HostSelctionRequired vic to select host if DRS is disabled & ROBO Adv. License
+	HostSelectionRequired bool `vic:"0.1" scope:"read-only"`
```
hickeng: Thought we were going to determine this at boot time in the endpoint?

lib/install/management/appliance.go
```diff
@@ -113,7 +113,12 @@ func (d *Dispatcher) checkExistence(conf *config.VirtualContainerHostConfigSpec,
...
+	if conf.HostSelectionRequired {
```
hickeng: Suggest just switching this based on DRS enabled state - it will still need to be done when the folder support is in place as we still want to support resource pools for non-ROBO installs.

lib/config/virtual_container_host.go
anchal-agrawal: I was wondering whether adding this field would require a future change to account for `vic-machine upgrade` in the code path(s) where we apply VCH/container configuration (similar to what https://github.com/vmware/vic/pull/6284 did). However, we'll most likely run feature/license checks and set `HostSelectionRequired` separately every time we run `vic-machine upgrade`, so I don't think the said code change would be required.

lib/install/management/appliance.go
cgtexmex: Agreed

lib/config/virtual_container_host.go
cgtexmex: We discussed it and think that's the right thing to do longer term, but for now I was approaching the same way we handle license / feature eval currently...which is only during install.

cgtexmex:
@hickeng regarding the `skipDRSCheck` -- there was a data element ([`DisableDRSCheck`](https://github.com/vmware/vic/blob/master/lib/install/validate/config.go#L529)) for that and it was evaluated in the `CheckDrs` func, but I didn't see anywhere that it was set -- so it would always be false.

cgtexmex:
PR updated with following changes:

- create supports ResourcePools as needed and isn't temporary code as before
- delete by name now works

lib/install/management/create.go
```diff
@@ -45,6 +45,17 @@ func (d *Dispatcher) CreateVCH(conf *config.VirtualContainerHostConfigSpec, sett
...
+	// resource pool path determined based on DRS setting.  If DRS is disabled then
+	// no resource pools are supported, so use the provided compute path.  If enabled then
+	// append the appliance name to the path and a pool will be created.
+	if conf.DRSDisabled {
+		d.vchPoolPath = settings.ResourcePoolPath
+	} else {
+		d.vchPoolPath = path.Join(settings.ResourcePoolPath, conf.Name)
+	}
+
+	// TODO:  This check's existence via resourcePool -- with the intro of inventory
```
matthewavery: As long as we get the correct resource pool we should still be able to determine if one of the VM's is the VCH right? So this should still be good I think. The inventory entry for the vm will be moved, but should still be in the same resource pool with my changes.

lib/install/validate/config.go
```diff
@@ -517,54 +516,44 @@ func (v *Validator) checkEvalLicense(op trace.Operation, licenses []types.Licens
...
 	if !v.sessionValid(op, errMsg) {
 		return
 	}
 
-	cl := v.Session.Cluster
-	ref := cl.Reference()
-
-	if v.isStandaloneHost() {
+	// TODO:  Cluster should only every be a cluster
+	if v.Session.Cluster.Reference().Type != "ClusterComputeResource" {
```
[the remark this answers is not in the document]
cgtexmex: I hear you..but will leave for now as it's going to be refactored and I can address that concern...which will just hopefully go away...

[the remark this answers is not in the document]
cgtexmex: not the concern..the hard coded string...

pkg/vsphere/compute/cluster.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"github.com/vmware/govmomi/object"
+	"github.com/vmware/vic/pkg/trace"
+)
+
+type Cluster struct {
+	*object.ClusterComputeResource
+}
+
+func NewCluster(compute object.ComputeResource) *Cluster {
```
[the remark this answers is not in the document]
cgtexmex: Now you're just trying to make my life difficult ;-)

pkg/vsphere/session/session.go
```diff
@@ -327,6 +330,14 @@ func (s *Session) Populate(ctx context.Context) (*Session, error) {
 		errs = append(errs, fmt.Sprintf("Failure finding cluster (%s): %s", s.ClusterPath, err.Error()))
 	} else {
 		op.Debugf("Cached cluster: %s", s.ClusterPath)
+		// if we have a cluster lets get DRS Status
+		if s.Cluster != nil && s.Cluster.Reference().Type == "ClusterComputeResource" {
```
matthewavery: constant for the magic value here. Looks like it lines up with the one above. I think we have a constants file somewhere :/

matthewavery: Maybe just add that constant to the new Cluster package that you created. :)

pkg/vsphere/compute/cluster.go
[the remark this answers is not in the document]
cgtexmex: The assignment of `ComputeResource` isn't a pointer -- `ClusterComputeResource` is the pointer...so not sure it really needs to be a pointer..but fine either way...currently not used

pkg/vsphere/session/session.go
cgtexmex: I'm not using the cluster addition because of go import madness...I think we go constant crazy...anyhoo...will rectify when I clean up the DRS check (which requires Cluster)

PR #7441, opened 2018-03-02
Delete datastore files iteratively for all datastore types

### LINKED ISSUE #7359: vic-machine-{os} delete fails to cleanup properly if there are pulled images

AngieCris:
`[specific ci=Group6-VIC-Machine]`

Fixes #7359 

Revert of commit https://github.com/vmware/vic/commit/1fd1e7fa361da3623a6e629759687315d0f323fe PR #6951 

For non-vSAN type datastores, sometimes it complains when we try to delete non-empty datastore folders with child image files, causing trailing VCH datastore folders after vic-machine delete finishes.
So the new approach is, we attempt to delete a datastore folder first. If it fails, we delete children files one by one, and then finally empty the folder.


emlin:
https://github.com/vmware/vic/pull/2652 is the PR to delete all kind of datastore files recursively, for vc6.5 requires that.

emlin:
lgtm

lib/install/management/store_files.go
```diff
@@ -158,26 +158,23 @@ func (d *Dispatcher) isVSAN(ds *object.Datastore) bool {
...
-		// This behaviour is specifically for vSan, as vSan sometimes throws an error when deleting a folder that is not empty.
-		res, err := d.getSortedChildren(ds, dsPath)
-		if err != nil {
-			if !types.IsFileNotFound(err) {
-				err = errors.Errorf("Failed to browse sub folders %q: %s", dsPath, err)
-				return err
-			}
-			d.op.Debugf("Folder %q is not found", dsPath)
-			return nil
+	// Get sorted result to make sure children files listed ahead of folder. Then we can empty folder before delete it
```
anchal-agrawal: s/children/child

anchal-agrawal: s/delete/deleting

lib/install/management/store_files.go
```diff
@@ -158,26 +158,23 @@ func (d *Dispatcher) isVSAN(ds *object.Datastore) bool {
...
+	// This function specifically designed for vSan, as vSan sometimes will throw error to delete folder is the folder is not empty
```
anchal-agrawal: Can we remove this comment if it's not relevant anymore?

lib/install/management/store_files.go
```diff
@@ -158,26 +158,23 @@ func (d *Dispatcher) isVSAN(ds *object.Datastore) bool {
 func (d *Dispatcher) deleteFilesIteratively(m *object.DatastoreFileManager, ds *object.Datastore, dsPath string) error {
 	defer trace.End(trace.Begin(dsPath, d.op))
```
hickeng: sorry for the whiplash Angie - given this isn't a direct revert if addressing anchal's comments, could we add a recursive delete _attempt_ as the first thing, then fail over into the current path.
So adding the following at the top:
```go
// don't check errors given detail process of any remaining files will occur after
d.deleteVMFSFiles(m, ds, dsPath)
```

AngieCris:
@anchal-agrawal 
`6-03-Delete` test suite already includes a test case called `Attach Disks and Delete VCH` that pulls an image, attaches the disk files, then delete the VCH, and finally check if the image folder is deleted. This covers the case where it could potentially fail.

The reason why all group 6 tests didn't fail on CI before was because old CI ran on ESX, and deleting a non-empty datastore folder with child image files works on ESX. But the operation won't go thru on newer version of VC.

lib/install/management/store_files.go
AngieCris: that's alright. unnecessary recursive delete doesn't sound very neat tbh....
it seems like ESX and older versions of VC doesn't have the issue but vSAN and new VC version do have that, in most cases it has to go thru iterative delete

if we add `d.deleteVMFSFiles(m, ds, dsPath)` at the top, then it would probably print failure log messages complaining not being able to delete non-empty datastore folders. Since we ignore the error coming out of this, and recursively clean stuff up later, if the folder did get cleaned up in the end, looking at the log it would still feel like that there's a failure deleting the folder.
sorry about the messy description, but just wondering will it cause bad user experience?

AngieCris: It looks like this from a test run I did:
```
INFO[0003] Removing image stores 
DEBU[0004] Failed to delete "[datastore1] vch-test-4/VIC": Cannot delete file [datastore1] vch-test-4/VIC 
DEBU[0013] Folder "[datastore1] vch-test-4/kvStores" is not found 
DEBU[0013] Image store parent directory not empty, leaving in place.
......
DEBU[0016] Folder "[datastore1] vch-test-4" is not found 
INFO[0016] Removing Resource Pool "vch-test-4" 
```
The later `Folder "[datastore1] vch-test-4" is not found` is from `vm.Unregsiter`, it shows that the folder is already been deleted
but still we see ` Failed to delete "[datastore1] vch-test-4/VIC": Cannot delete file [datastore1] vch-test-4/VIC` on the top, even tho it's actually deleted successfully right after

hickeng: I'm okay with them showing up as debug messages - we could add an additional debug message saying "Attempting recursive delete" then if that's not clean a message saying "Cleaning up recusive deletion failures via iterative delete" or similar.

We may want to return a different error type rather than re-use NotFound if it's causing misleading output.

AngieCris:
@hickeng 
Just thought of that since our new CI runs on VC right now, this bug might cause datastore files failed to delete for every one of the VCHs created on CI......
We might want to get this PR in asap...

lib/install/management/store_files.go
AngieCris: Got it. Just experimented and here's the log output:
```
DEBU[0007] No disks found attached to VM                
INFO[0007] Removing image stores                        
DEBU[0009] Failed to delete "[datastore1] vch-2/VIC": Cannot delete file [datastore1] vch-2/VIC 
DEBU[0009] Attempt to delete top level folder "[datastore1] vch-2/VIC" failed. Removing the children files instead. 
DEBU[0027] Image store parent directory not empty, leaving in place. 
INFO[0027] Removing volume stores 
```
Does this look okay? I added a log saying `Attempt to delete top level folder failed. Removing the children files instead`

PR #7444, opened 2018-03-03
Make NFS mount target clean path

hickeng:
Fixes a "permission denied" response from a qnap NAS and possibly an open SR
relating to a NetApp share.

@matthewavery  please open an issue for the SR you mentioned if this seems likely
to be the problem and add a link to that issue.

@mhagen-vmware this does suggest that we need a much more strict NFS server to
test against as the default linux config we have seems _incredibly_ forgiving.


hickeng:
As a note on how I tracked this down, it was a comparison between two wireshark NFS protocol traces:
failed:
```
Mount Service
    [Program Version: 3]
    [V3 Procedure: MNT (1)]
    Path: //vic-volumes/volumes/custom-iso
        length: 32
        contents: //vic-volumes/volumes/custom-iso
```
working:
```
Mount Service
    [Program Version: 3]
    [V3 Procedure: MNT (1)]
    Path: /vic-volumes/volumes/custom-iso
        length: 31
        contents: /vic-volumes/volumes/custom-iso
        fill bytes: opaque data
```

lib/tether/ops_linux.go
```diff
@@ -893,10 +894,18 @@ func (t *BaseOperations) MountTarget(ctx context.Context, source url.URL, target
 		return fmt.Errorf("unable to create mount point %s: %s", target, err)
 	}
 
-	rawSource := source.Hostname() + ":/" + source.Path
+	var rawSource bytes.Buffer
```
vburenin: purely for readability I would do the following:

src := filepath.Join(source.Hostname() + ":", source.Path)

filepath.Join controls number of leading slashes,
https://play.golang.org/p/cx4yphPIN3h

hickeng: It's also significantly slower and with extra allocations; neither of which will be significant in this case.
I mostly did not want to rely on `Clean` (which Join calls) because I don't want to impose actual path content restrictions on people.

PR #7446, opened 2018-03-05
Provide more time for slower VC operations

mhagen-vmware:
The VM ended up powering off in 24s during the last test.


vburenin:
@mhagen-vmware I would consider to increase this interval up to 60 seconds. 30 seconds vs 24 seconds is a very tight gap. In case of overloaded environment you can easily go over 30 seconds. A good rule would be setting 2x large value than expected to handle unexpected load.

PR #7450, opened 2018-03-05
Issue gratuitous ARP request on network update

### LINKED ISSUE #2932: Containers connected to non-bridge networks should issue gratuitous ARP broadcast
**VIC Version:**

build 6299

**Expected behavior:**

Containers connected to non-bridge are accessible over the network as soon as they start up.

**Actual behavior:**

tl;dr: No connection to the container until ARP cache on the gateway expires.

Explanation in detail:

Let's take a container that's configured to connect to a real external or any other non-bridge network using static IP assignment. Its default gateway will likely be a proper network device such as a router or a firewall. Very often, these devices have very long ARP cache expiration timeouts (for example, Cisco default is 4 hours). Now, let's take a situation where that container is wiped and recreated, for whatever reason. vSphere will assign it a new MAC address, however, the gateway device will not be aware of that. There is no traffic coming from the container VM, ARP cache is still valid, so the gateway may not issue any ARP requests to the container VM for quite some time. All this time, all attempts to connect to the container from outside its network through the gateway will fail. 

This could also affect clients on the same subnet that have similar ARP behavior (that is, long ARP cache expirations and no attempts to issue ARP requests if their SYNs are not answered).

The solution to this is to issue a gratuitous ARP broadcast once tether finishes configuring the container's network stack.

hickeng:
This is a lazy implementation that issues a gratuitous ARP request any time the
interface ApplyEndpoint logic is called, irrespective of whether changes occurred.

```
[skip unit]
[specific ci=10-01-VCH-Restart,6-07-Create-Network]
```

Fixes: #2932 
IP collision detection follow up: #7485


lib/tether/ops_linux.go
```diff
@@ -698,6 +731,13 @@ func ApplyEndpoint(nl Netlink, t *BaseOperations, endpoint *NetworkEndpoint) err
 		}
 	}
 
+	// notify our peers of the assigned address
+	err = gratuitous(t, link, endpoint)
```
morris-jason: Should we put this in the deffered function on line 622? Due to the return on line 677, I'm curious to know if broadcast would occur in the following case (or if it's not needed here):

1) User deploys a VCH with dchp issuing it an address or x.x.x.1.
2) User upgrades & reconfigures or otherwise deploys a vch with **static** ip x.x.x.1, where this function exists on L677.

Will bridged containers will be unaware of the mac address change?

lib/tether/ops_linux.go
```diff
@@ -326,6 +328,37 @@ func renameLink(t Netlink, link netlink.Link, slot int32, endpoint *NetworkEndpo
 	return link, nil
 }
 
+func gratuitous(t Netlink, link netlink.Link, endpoint *NetworkEndpoint) error {
+	if ip.IsUnspecifiedIP(endpoint.Assigned.IP) {
+		// refusing to broadcast unspecified IP
+		return nil
+	}
+
+	intf, err := net.InterfaceByIndex(link.Attrs().Index)
```
vburenin: Is link.Attrs().Index actually the same as slot in here? `nl.LinkBySlot(int32(slot))` Just curious.

lib/tether/ops_linux.go
hickeng: I've moved the call to immediately after the endpoint is updated with the new address so that we make a good best effort to update ARP tables even if subsequent updates such as routing tables fails.

We do not need to care about the early return at 677 - that's a scenario where we are not assigning an IP address because the link is not managed by tether.

lib/tether/ops_linux.go
hickeng: No. `LinkBySlot` is getting a link based on it's PCI slot number, whereas `Index` is the internal Linux indexing for the interfaces.

lib/tether/ops_linux.go
```diff
@@ -326,6 +328,37 @@ func renameLink(t Netlink, link netlink.Link, slot int32, endpoint *NetworkEndpo
...
+		return err
+	}
+	defer func() {
+		client.Close()
+	}()
+
+	srcHW := link.Attrs().HardwareAddr
+	srcPA := endpoint.Assigned.IP
+
+	gratuitousPacket, err := arp.NewPacket(arp.OperationRequest, srcHW, srcPA, ethernet.Broadcast, srcPA)
```
vburenin: What if you get a response for this ARP request? Which of course means IP address conflict.
I think we may want to send this ARP request a few times just in case of packet loss or some congestion happens and packet is accidentally was dropped.

I would run a goroutine that sends the same packet twice after that later.

lib/tether/ops_linux.go
morris-jason: Thanks for clarification!

sgairo:
This is a potential contributor to https://github.com/vmware/vic/issues/5819

lib/tether/ops_linux.go
hickeng: @vburenin I'm going to stick with just the one which is in line with the linux networking stack when it issues them. In a distributed switch vSphere will be doing ARP suppression and processing anyway  to avoid ARP cascades. This is primarily an issue for us in the bridge networks as that's where we get very rapid reuse of the same IP addresses and in that case it's the DVS/vSphere level caches we care about.

PR #7454, opened 2018-03-05
Vch inventory folder support

matthewavery:
[specific ci=Group6-VIC-Machine]

This PR should provide a default behavior for `vic-machine create` and `vic-machine delete` with regards to inventory folders. 

The default behavior for `vic-machine create` will be to mimic the compute resource structure and add a `VCH/VCH.vm` namespacing on the end. This should allow for multiple VCHs with the same name but different compute resources. 

The default behavior for `vic-machine delete` will be to remove folders working backwards from the VCH folder until we find a folder that is not empty or we find the `VMFolder`.

note: tests are not added yet.

### DIFF
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -44,6 +44,7 @@ import (
 	"github.com/vmware/vic/lib/spec"
 	"github.com/vmware/vic/pkg/errors"
 	"github.com/vmware/vic/pkg/ip"
+	"github.com/vmware/vic/pkg/retry"
 	"github.com/vmware/vic/pkg/trace"
 	"github.com/vmware/vic/pkg/vsphere/compute"
 	"github.com/vmware/vic/pkg/vsphere/diag"
@@ -201,88 +202,6 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
 		return err
 	}, tasks.IsConcurrentAccessError)
 
-	//removes the vm from vsphere, but detaches the disks first
-	_, err = vm.WaitForResult(d.op, func(ctx context.Context) (tasks.Task, error) {
-		return vm.DeleteExceptDisks(ctx)
-	})
-
-	// We are getting ConcurrentAccess errors from DeleteExceptDisks - even though we don't set ChangeVersion in that path
-	// We are ignoring the error because in reality the operation finishes successfully.
-	ignore := false
-	if err != nil {
-		if f, ok := err.(types.HasFault); ok {
-			switch f.Fault().(type) {
-			case *types.ConcurrentAccess:
-				d.op.Warn("DeleteExceptDisks failed with ConcurrentAccess error. Ignoring it.")
-				ignore = true
-			}
-		}
-		if !ignore {
-			err = errors.Errorf("Failed to destroy VM %q: %s", vm.Reference(), err)
-			err2 := vm.Unregister(d.op)
-			if err2 != nil {
-				return errors.Errorf("%s then failed to unregister VM: %s", err, err2)
-			}
-			d.op.Infof("Unregistered VM to cleanup after failed destroy: %q", vm.Reference())
-		}
-	}
-
-	if _, err = d.deleteDatastoreFiles(d.session.Datastore, folder, true); err != nil {
-		d.op.Warnf("Failed to remove datastore files for VM path %q: %s", folder, err)
-	}
-
-	// at this point the VCH should be gone and we need to cleanup the inventory
-	err = d.deleteVCHInventoryFolders()
-	if err != nil {
-		// if this fails manual cleanup is necessary
-		return err
-	}
-
-	return nil
-}
-
-func (d *Dispatcher) deleteVCHInventoryFolders() error {
-	parentFolderPath := path.Dir(d.appliance.InventoryPath)
-	// Now that the VCH is gone we will need to delete the inventory folder structure.
-	d.op.Debugf("Attempting to remove inventory folder: %s", parentFolderPath)
-
-	// grab the folder ref of the vm's direct parent
-	folderRef, err := d.session.Finder.Folder(d.op, parentFolderPath)
-	if err != nil {
-		d.op.Debugf("failed to find folder: %s", parentFolderPath)
-		return err
-	}
-
-	// we need to see if the folder is empty.
-	folderContents, err := folderRef.Children(d.op)
-	if err != nil {
-		return err
-	}
-
-	for len(folderContents) == 0 && folderRef.Reference() != d.session.VMFolder.Reference() {
-		// NOTE: Destroy on Inventory Folders is RECURSIVE, start from the leaf most target and check folder children to avoid undesired deletions.
-		err = d.removeFolder(folderRef)
-		if err != nil {
-			return err
-		}
-		d.op.Debugf("Successfully deleted folder at path : %s", parentFolderPath)
-
-		parentFolderPath = path.Dir(parentFolderPath)
-		folderRef, err = d.session.Finder.Folder(d.op, parentFolderPath)
-		if err != nil {
-			// at this point we have already partially cleaned up. So we may leave artifacts around when we bale.
-			return err
-		}
-
-		folderContents, err = folderRef.Children(d.op)
-		if err != nil {
-			return err
-		}
-	}
-
-	// We are getting ConcurrentAccess errors from DeleteExceptDisks - even though we don't set ChangeVersion in that path
-	// We are ignoring the error because in reality the operation finishes successfully.
-	ignore := false
 	if err != nil {
 		d.op.Warnf("Destroy VM %s failed with %s, unregister the VM instead", vm.Reference(), err)
 
@@ -301,20 +220,6 @@ func (d *Dispatcher) deleteVCHInventoryFolders() error {
 		d.op.Warnf("Failed to remove datastore files for VM %s with folder path %s: %s", vm.Reference(), folder, err)
 	}
 
-	return nil
-}
-
-// removeFolder will initiate a destroy task against the target folder and wait for the result.
-func (d *Dispatcher) removeFolder(folder *object.Folder) error {
-	// NOTE: Destroy on Inventory Folders is RECURSIVE, start from the leaf most target and check folder children to avoid undesired deletions.
-	folderRemoveFunction := func(ctx context.Context) (tasks.Task, error) {
-		return folder.Destroy(d.op)
-	}
-
-	_, err := tasks.WaitForResult(d.op, folderRemoveFunction)
-	if err != nil {
-		return err
-	}
 	return nil
 }
 
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -498,7 +498,40 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 	} else {
 		// if vapp is not created, fall back to create VM under default resource pool
 		info, err = tasks.WaitForResult(d.op, func(ctx context.Context) (tasks.Task, error) {
-			return d.session.VMFolder.CreateVM(ctx, *spec, d.vchPool, d.session.Host)
+			// Default behavior is to check if the VCH folder already exists in the VMFolder. If so reject.
+			// Otherwise make a folder with the VCH name and create the VCH VM there.
+
+			vmFolderPath := d.session.VMFolder.InventoryPath
+			vchName := spec.Name
+			vchFolderPath := fmt.Sprintf("%s/%s", vmFolderPath, vchName)
+			_, err := d.session.Finder.Folder(ctx, vchFolderPath)
+			if err == nil {
+				// This should mean that the folder existed or that we could find it at the very least.
+				// in this case we should reject since the folder is already there? It should be cleaned up on a delete normally...
+				// we should check for the govmomi define NotFoundError here. see reference of `folderList` function within the `Folder` function
+				return nil, fmt.Errorf("the path %s already exists for an existing VCH", vchFolderPath)
+			}
+
+			vchVMPath := fmt.Sprintf("%s/%s", vchFolderPath, vchName)
+			vchVM, err := d.session.Finder.VirtualMachine(ctx, vchVMPath)
+			if vchVM != nil {
+				// We have found another VCH at the target path that we intended to install to.
+			}
+
+			vchFolder, err := d.session.VMFolder.CreateFolder(ctx, vchName)
+			if err != nil {
+				// POSSIBLE FAULTS HERE:
+				// - InvalidName
+				// - DuplicateName
+				// - Runtime
+				return nil, fmt.Errorf("failed to create vch folder under the VMFolder in vsphere: %s", err)
+			}
+
+			// we should also be storing this path somewhere. rather than just looking at the VMFolder.
+			// This should be stored in GuestInfo right now since we will need it in guest info eventually and it will
+			// provide that functionality now.
+
+			return vchFolder.CreateVM(ctx, *spec, d.vchPool, d.session.Host)
 		})
 	}
 
--- a/lib/install/management/appliance.go
+++ b/lib/install/management/appliance.go
@@ -35,7 +35,6 @@ import (
 	"github.com/docker/docker/opts"
 
 	"github.com/vmware/govmomi/object"
-	"github.com/vmware/govmomi/task"
 	"github.com/vmware/govmomi/vim25/soap"
 	"github.com/vmware/govmomi/vim25/types"
 	"github.com/vmware/vic/lib/config"
@@ -603,21 +602,27 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 
 	var info *types.TaskInfo
 
-	vchParentFolder, err := d.createFolders(spec.Name)
+	d.op.Info("Creating the VCH inventory folder")
+
+	vchFolder := d.session.VMFolder
+	createFolders := func() error {
+		vchFolder, err = d.createFolders(spec.Name)
+		if err != nil {
+			return err
+		}
+		return nil
+	}
+
+	err = retry.Do(createFolders, isConcurrentAcces)
 	if err != nil {
+		d.op.Debugf("Encountered a failure during creation of the inventory folders : %s", err)
+
+		// TODO: we should likely do some processing on this error. Mainly for customer readability. log something friendly to the console and return a human error rather than a raw fault.
 		return err
 	}
 
-	// if vapp is not created, fall back to create VM under default resource pool
-	intendedVCHPath := fmt.Sprintf("%s/%s", vchParentFolder.InventoryPath, spec.Name)
 	info, err = tasks.WaitForResult(d.op, func(ctx context.Context) (tasks.Task, error) {
-		vchVM, err := d.session.Finder.VirtualMachine(ctx, intendedVCHPath)
-		if vchVM != nil {
-			// We have found another VCH at the target path that we intended to install to.
-			return nil, fmt.Errorf("Could not create VCH because one already exists with the same name and compute at path(%s) : %s", intendedVCHPath, err)
-		}
-
-		return vchParentFolder.CreateVM(ctx, *spec, d.vchPool, d.session.Host)
+		return vchFolder.CreateVM(ctx, *spec, d.vchPool, d.session.Host)
 	})
 
 	if err != nil {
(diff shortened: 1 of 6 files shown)


lib/portlayer/exec/config.go
```diff
@@ -53,4 +53,7 @@ type Configuration struct {
 
 	// Datastore URLs for image stores - the top layer is [0], the bottom layer is [len-1]
 	ImageStores []url.URL `vic:"0.1" scope:"read-only" key:"storage/image_stores"`
+
+	// Inventory Folder Path for creating cVMs
+	VCHFolderPath string `vic:"0.1" scope:"read-only" key:"container/VCHFolderPath"`
```
matthewavery: Removing this immediately. It was a bad idea...

lib/install/management/appliance.go
```diff
@@ -219,6 +222,69 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
...
+		err = d.removeFolder(folderRef)
+		if err != nil {
+			return err
+		}
+		d.op.Debugf("Successfully deleted folder at path : %s", parentFolderPath)
+
+		parentFolderPath = path.Dir(parentFolderPath)
+		folderRef, err = d.session.Finder.Folder(d.op, parentFolderPath)
+		if err != nil {
+			// at this point we have already partially cleaned up. So we may leave artifacts around when we bale.
```
sgairo: "bail" I believe

lib/install/management/delete.go
```diff
@@ -82,6 +82,13 @@ func (d *Dispatcher) DeleteVCH(conf *config.VirtualContainerHostConfigSpec, cont
 		}
 	}
 
+	element, err := d.session.Finder.Element(d.op, vmm.Reference())
```
matthewavery: Will need some eyes on this. Before I did a new Element lookup the inventory path was being incorrectly filled. I can likely just refactor the other call. I did this to get around the InventoryPath mysteriously being `""`.

lib/install/management/appliance.go
matthewavery: ha yeah I was thinking of hay whenever I typed that... #CountryRoots

lib/install/management/appliance.go
```diff
@@ -35,6 +35,7 @@ import (
 	"github.com/docker/docker/opts"
```
sgairo: copyright

lib/install/management/appliance.go
```diff
@@ -219,6 +222,69 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
...
+	err = d.deleteVCHInventoryFolders()
+	if err != nil {
+		// if this fails manual cleanup is necessary
+		return err
+	}
+
+	return nil
+}
+
+func (d *Dispatcher) deleteVCHInventoryFolders() error {
```
matthewavery: This could be moved to `lib/install/management/delete.go` instead of being here. Thoughts?

lib/install/management/appliance.go
```diff
@@ -219,6 +222,69 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
...
+
+	for len(folderContents) == 0 && folderRef.Reference() != d.session.VMFolder.Reference() {
+		// NOTE: Destroy on Inventory Folders is RECURSIVE, start from the leaf most target and check folder children to avoid undesired deletions.
+		err = d.removeFolder(folderRef)
+		if err != nil {
+			return err
+		}
+		d.op.Debugf("Successfully deleted folder at path : %s", parentFolderPath)
+
+		parentFolderPath = path.Dir(parentFolderPath)
```
matthewavery: Add comment about parsing the Parent InventoryFolder path.

lib/install/management/appliance.go
```diff
@@ -503,7 +569,20 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 	} else {
 		// if vapp is not created, fall back to create VM under default resource pool
 		info, err = tasks.WaitForResult(d.op, func(ctx context.Context) (tasks.Task, error) {
-			return d.session.VMFolder.CreateVM(ctx, *spec, d.vchPool, d.session.Host)
+
+			vchParentFolder, err := d.createVCHInventoryFolders(spec)
```
matthewavery: Should not be in this WaitForResult. Move it out and above.

lib/install/management/appliance.go
```diff
@@ -660,6 +739,132 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 	return nil
 }
 
+// createVCHInventoryFolders creates the default structure of the inventory for the VCH and returns the object reference to the last folder created
+func (d *Dispatcher) createVCHInventoryFolders(spec *types.VirtualMachineConfigSpec) (*object.Folder, error) {
+	d.op.Info("Creating VCH inventory folders")
+	vchName := spec.Name
+	clusterName := path.Base(d.session.ClusterPath)
```
matthewavery: just append this, do consolidate it. it is wasted effort.

lib/install/management/appliance.go
```diff
@@ -660,6 +739,132 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
...
+		f, err := existingFolder.CreateFolder(d.op, fname)
+		ok := false
+		if err != nil {
+			// check for already exists fault
+
+			// Most likely the case for "AlreadyExists"
+			if soap.IsSoapFault(err) {
+
+				switch fault := soap.ToSoapFault(err).VimFault().(type) {
+				case types.AlreadyExists:
```
matthewavery: This is not a success case. We must reject here.

lib/install/management/appliance.go
```diff
@@ -660,6 +739,132 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
...
+		d.op.Debugf("Attempting to create folder : %s", folderCreateTarget)
+
+		// attempt to create the folder
+		f, err := existingFolder.CreateFolder(d.op, fname)
+		ok := false
+		if err != nil {
+			// check for already exists fault
+
+			// Most likely the case for "AlreadyExists"
+			if soap.IsSoapFault(err) {
```
matthewavery: Break this out into a certification function.

lib/install/management/appliance.go
```diff
@@ -660,6 +739,132 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
...
+						d.op.Errorf("Encountered unexpected fault : %#v ", fault)
+						return nil, fmt.Errorf("unexpected fault when attempting to create the inventory folder %s please see vic-machine.log for more information", folderCreateTarget)
+					}
+				default:
+					d.op.Errorf("Encountered unexpected error : %#v ", unknownError)
+					return nil, fmt.Errorf("unexpected error when attempting to create the inventory folder %s please see vic-machine.log for more information", folderCreateTarget)
+				}
+
+				if ok {
+					// the folder already exists. Get the ref.
```
matthewavery: change this comment, `AlreadyExists` style fault is actually bad. Something else like "the folder is already in the inventory".

lib/install/management/appliance.go
```diff
@@ -660,6 +739,132 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
...
+	consolidated := clusterName + "/" + strings.TrimLeft(d.vchPoolPath, d.session.ClusterPath)
+
+	// assemble creation list
+	folders := strings.Split(consolidated, "/")
+	folders = append(folders, vchName)
+	d.op.Debugf("Determined inventory targets for creation: %s", folders)
+
+	// now that we have the list of needed constructions we need to assemble it.
+	existingFolder := d.session.VMFolder
+	for _, fname := range folders {
```
matthewavery: Find all the targets that you can first. Then create the rest in order to avoid needless govmomi calls.

lib/install/management/appliance.go
```diff
@@ -219,6 +222,69 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
 		d.op.Warnf("Failed to remove datastore files for VM path %q: %s", folder, err)
 	}
 
+	// at this point the VCH should be gone and we need to cleanup the inventory
+	err = d.deleteVCHInventoryFolders()
```
matthewavery: Move this logic to below line `109` so that it occurs after the `deleteVM` call

lib/install/management/appliance.go
```diff
@@ -219,6 +222,69 @@ func (d *Dispatcher) deleteVM(vm *vm.VirtualMachine, force bool) error {
...
+
+	// we need to see if the folder is empty.
+	folderContents, err := folderRef.Children(d.op)
+	if err != nil {
+		return err
+	}
+
+	for len(folderContents) == 0 && folderRef.Reference() != d.session.VMFolder.Reference() {
+		// NOTE: Destroy on Inventory Folders is RECURSIVE, start from the leaf most target and check folder children to avoid undesired deletions.
+		err = d.removeFolder(folderRef)
```
matthewavery: Don't delete here, keep track of the last one to delete. Since it is recursive. This will avoid needless vsphere calls.

lib/install/management/appliance.go
matthewavery: Needs to be. Separation of concerns.

lib/install/management/delete.go
matthewavery: Moved this in to the findApplianceByID function. Should fill the inventory Path correctly for all callers now.

lib/install/management/appliance.go
matthewavery: Arguably this function should not be in this function at all. it should be in the `CreateVCH` path

matthewavery: Kept it in the same function, but moved it out of the WaitForResult.

lib/install/management/appliance.go
matthewavery: removed comment altogether.

lib/install/management/appliance.go
matthewavery: This eliminates the need for us to determine if a folder already exists or not when advancing the pointer.

lib/install/management/appliance.go
```diff
@@ -660,6 +691,96 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
...
+	for _, fname := range folders {
+		folderCreateTarget := fmt.Sprintf("%s/%s", createdFolder.InventoryPath, fname)
+		d.op.Debugf("Attempting to create folder : %s", folderCreateTarget)
+
+		// attempt to create the folder
+		f, err := createdFolder.CreateFolder(d.op, fname)
+		err = processInventoryCreationError(d.op, err, fname, vchName)
+
+		// creation failed, we must cleanup and bail
+		if err != nil {
```
matthewavery: I want to note for discussion sake, we do not check the error for a logical action here. we log it and then attempt to cleanup. There is a case here that we will likely be concerned with. if a `DuplicateName` occurs it is possible that a VM was here instead of a folder. in which case we should likely not call `remove` on the base folder. I do not currently see this as an issue as we have attempted a `Finder.Folder` read on the target and we did not get a folder back. which means we will try to make a folder. At that point we should fail and the cleanup should fail as well. Additionally if the folder did exist we would not be trying to make it. So I do not think the lack of check for a `DuplicateName` issue would be a problem here, though I could be missing some edge case that I have not though of. Feel free to add addtional thoughts and comments!

lib/install/management/appliance.go
```diff
@@ -397,6 +409,12 @@ func (d *Dispatcher) findApplianceByID(conf *config.VirtualContainerHostConfigSp
 		return nil, err
 	}
 	vmm = vm.NewVirtualMachine(d.op, d.session, ovm.Reference())
+
+	element, err := d.session.Finder.Element(d.op, vmm.Reference())
```
matthewavery: I am thinking we will want this to become a function from the dispatcher which creates a singleton private var. so if private var == nil get pointer to VMFolder and set private var... etc...

matthewavery:
Move Folder functionality to a `pkg`

lib/install/management/appliance.go
```diff
@@ -660,6 +683,29 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 	return nil
 }
 
+// createFolders creates the default structure of the inventory for the VCH and returns the object reference to the last folder created
+func (d *Dispatcher) createFolders(vchName string) (*object.Folder, error) {
```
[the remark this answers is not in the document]
matthewavery: I am keeping it as a function since it will simplify the retry on concurrent access approach(push coming soon, just wrapping up other things).

lib/install/management/appliance.go
```diff
@@ -660,6 +683,29 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
...
+	if !d.isVC {
+		return d.session.VMFolder, nil
+	}
+
+	d.op.Info("Creating VCH inventory folder")
+	vchFolder, err := d.session.VMFolder.CreateFolder(d.op, vchName)
+	if err != nil {
+		found, err := d.processInventoryCreationError(err, vchName)
```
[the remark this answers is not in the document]
matthewavery: I think we should keep part of this. It provides a one stop shop for us to "prettify" the error. Then we can log the raw error on `DEBUG` and return the human readable error(w/ explanations) directly to the user.

lib/install/management/appliance.go
```diff
@@ -1184,3 +1230,89 @@ func (d *Dispatcher) CheckServiceReady(ctx context.Context, conf *config.Virtual
 	}
 	return nil
 }
+
+// processInventoryCreationError will check the given error and return whether it is a inventory folder creation error and return a human readable error if so
+func (d *Dispatcher) processInventoryCreationError(err error, name string) (*object.Folder, error) {
```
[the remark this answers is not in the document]
matthewavery: ^

lib/install/management/appliance.go
```diff
@@ -495,18 +512,24 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
...
-			return d.session.VMFolder.CreateVM(ctx, *spec, d.vchPool, d.session.Host)
-		})
+
+	vchParentFolder, err := d.createFolders(spec.Name)
+	if err != nil {
+		return err
 	}
 
+	// if vapp is not created, fall back to create VM under default resource pool
+	intendedVCHPath := fmt.Sprintf("%s/%s", vchParentFolder.InventoryPath, spec.Name)
```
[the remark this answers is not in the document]
matthewavery: this was intended to check for the existence of a VCH already in the inventory target. we might want to move this to `CheckExistence` as we will want to confirm that our target is no already there before we attempt to create the vch altogether.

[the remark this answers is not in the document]
matthewavery: never mind :) cannot create the folder if the vm is already there.

lib/install/management/delete.go
```diff
@@ -342,3 +357,62 @@ func (d *Dispatcher) networkDevices(vmm *vm.VirtualMachine) ([]types.BaseVirtual
...
+	VMFolder := dcFolder.VmFolder
+	if folderRef.Reference() == VMFolder.Reference() {
+		// cannot delete the vm folder
+		return nil
+	}
+
+	// we need to see if the folder is empty.
+	folderContents, err := folderRef.Children(d.op)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
matthewavery: Oddly I have that second part locally but I am not seeing it in the branch. I must have missed an additional code addition that I had already written :)

lib/install/management/appliance.go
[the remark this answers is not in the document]
matthewavery: I plan to modify it such that it will provide some useful debug information and also take a raw error and return a useful error.

lib/install/management/delete.go
```diff
@@ -342,3 +357,62 @@ func (d *Dispatcher) networkDevices(vmm *vm.VirtualMachine) ([]types.BaseVirtual
...
+
+func (d *Dispatcher) deleteFolders() error {
+	if !d.isVC {
+		return nil
+	}
+
+	d.op.Info("Removing VCH Inventory Folder")
+	vchFolderPath := path.Dir(d.appliance.InventoryPath)
+	folderRef, err := d.session.Finder.Folder(d.op, vchFolderPath)
+	if err != nil {
```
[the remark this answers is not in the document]
matthewavery: will fix this.

lib/install/management/appliance.go
```diff
@@ -660,6 +681,19 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
 	return nil
 }
 
+// TODO this should be moved to somewhere common. I did not seen a common function used in vic-machine.(I did not really see this pattern in vic-machine)
+func isConcurrentAccess(err error) bool {
```
cgtexmex: No longer needed

lib/install/management/appliance.go
```diff
@@ -1184,3 +1218,27 @@ func (d *Dispatcher) CheckServiceReady(ctx context.Context, conf *config.Virtual
 	}
 	return nil
 }
+
+// TODO: we need a general place for this kind of handling in vic-machine.
+// processInventoryCreationError will check the given error and return whether it is a inventory folder creation error and return a human readable error if so
+func processInventoryCreationError(op trace.Operation, err error, name string) error {
+	if err != nil {
+		if f, ok := err.(types.HasFault); ok {
```
cgtexmex: Do we need the If?  Won't it just go to default if it can't be cast?

lib/install/management/delete.go
```diff
@@ -342,3 +354,53 @@ func (d *Dispatcher) networkDevices(vmm *vm.VirtualMachine) ([]types.BaseVirtual
...
+	}
+
+	// NOTE: Destroy on Inventory Folders is RECURSIVE, start from the leaf most target and check folder children to avoid undesired deletions.
+	folderRemoveFunction := func(ctx context.Context) (tasks.Task, error) {
+		return folderRef.Destroy(d.op)
+	}
+
+	_, err = tasks.WaitForResult(d.op, folderRemoveFunction)
+	if err != nil {
+		d.op.Debugf("Received error when attempting to delete the vch inventory folder: %s", err)
```
cgtexmex: remove inventory

lib/install/management/appliance.go
```diff
@@ -63,6 +63,15 @@ const (
 	volumeStoresID = "VolumeStores"
 )
 
+//VCH Inventory Constants
+const (
+	alreadyExistsError            = "An object already exists on the inventory path for vch (%s) that is not an inventory folder"
+	invalidNameError              = "An invalid name was specified for the VCH: %s does not meet the naming criteria for a vch"
+	unexpectedInventoryFaultError = "unexpected fault when attempting to create the inventory folder %s please see vic-machine.log for more information"
```
cgtexmex: remove inventory

lib/install/management/appliance.go
```diff
@@ -63,6 +63,15 @@ const (
...
+	unexpectedInventoryError      = "unexpected error when attempting to create the inventory folder %s please see vic-machine.log for more information"
```
cgtexmex: ditto

lib/install/management/appliance.go
```diff
@@ -495,18 +511,23 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
...
-			return d.session.VMFolder.CreateVM(ctx, *spec, d.vchPool, d.session.Host)
-		})
+
+	// Create the VCH inventory folder
+	vchFolder := d.session.VMFolder
+	if d.isVC {
+		d.op.Info("Creating the VCH inventory folder")
+		vchFolder, err = d.session.VMFolder.CreateFolder(d.op, spec.Name)
+		if err != nil {
+			d.op.Debugf("Encountered a failure during creation of the inventory folders : %s", err)
```
cgtexmex: possibly wordsmith to "VCH Folder creation failed: %s" -- don't need inventory and it's singular

lib/install/management/appliance.go
matthewavery: This is only used when folder creation fails.

lib/install/management/appliance.go
```diff
@@ -495,18 +507,37 @@ func (d *Dispatcher) createAppliance(conf *config.VirtualContainerHostConfigSpec
...
-	} else {
-		// if vapp is not created, fall back to create VM under default resource pool
-		info, err = tasks.WaitForResult(d.op, func(ctx context.Context) (tasks.Task, error) {
-			return d.session.VMFolder.CreateVM(ctx, *spec, d.vchPool, d.session.Host)
-		})
+
+	// Create the VCH inventory folder
+	vchFolder := d.session.VMFolder
+	if d.isVC {
+		d.op.Info("Creating the VCH inventory folder")
```
cgtexmex: prefer we remove the word inventory from all messages here

lib/install/management/delete.go
```diff
@@ -342,3 +354,53 @@ func (d *Dispatcher) networkDevices(vmm *vm.VirtualMachine) ([]types.BaseVirtual
...
+		return
+	}
+
+	// we need to see if the folder is empty.
+	folderContents, err := folderRef.Children(d.op)
+	if err != nil || len(folderContents) != 0 {
+		d.op.Debugf("Could not remove VCH inventory folder, %s has existing contents in it. Manual cleanup required.", vchFolderPath)
+	}
+
+	// NOTE: Destroy on Inventory Folders is RECURSIVE, start from the leaf most target and check folder children to avoid undesired deletions.
```
cgtexmex: Not sure this comment is needed since we only have a single folder

lib/install/management/delete.go
```diff
@@ -342,3 +354,53 @@ func (d *Dispatcher) networkDevices(vmm *vm.VirtualMachine) ([]types.BaseVirtual
...
+	folderRemoveFunction := func(ctx context.Context) (tasks.Task, error) {
+		return folderRef.Destroy(d.op)
+	}
+
+	_, err = tasks.WaitForResult(d.op, folderRemoveFunction)
```
cgtexmex: I prefer this approach:
```
info, err = tasks.WaitForResult(d.op, func(ctx context.Context) (tasks.Task, error) {
		return vchFolder.CreateVM(ctx, *spec, d.vchPool, d.session.Host)
	})
``` 
is that not applicable here?  I'm not seeing the reason for the func var...

